PR #320, opened 2020-02-12
Rework zng integer types

aswan:
- Implement types from the zng spec: byte, int16, uint16, int32, uint32, int64, uint64
- internal references to the old "count" type are replaced by uint64 and the name count in zeek logs is mapped to zng uint64
- zeek log reader creates an alias for "port" to uint16
- the zng enum type is removed

Testing for the new types is still quite thin, will add before merging, but looking for early feedback.


aswan:
Requirements here were not well thought out so this is getting a pretty major revision.

PR #322, opened 2020-02-13
create release assets for any v* tag

alfred-landrum:
When any tag matching a "v*" pattern is pushed to the repo, including
    when a Github release is created, this action will create per-platform
    zip files containing the binaries built from cmd/... . The zip files
    will be attached to the Github release as release assets.

    If the tag push was not due to a Release creation, this will create a
    Github release and attach the zip files as assets. I'm not using an
    action to explicitly create a release, as I didn't find anything yet
    that would automatically pull the release changes text from the
    changelog file; we can add that later.

The content of the zip files is a top level directory with the version name & platform OS and architecture:
```
$ unzip -t dist/release/zq-v0.1.0-3-gd3cda09-dirty.darwin-amd64.zip
Archive:  dist/release/zq-v0.1.0-3-gd3cda09-dirty.darwin-amd64.zip
    testing: zq-v0.1.0-3-gd3cda09-dirty.darwin-amd64/   OK
    testing: zq-v0.1.0-3-gd3cda09-dirty.darwin-amd64/zq   OK
    testing: zq-v0.1.0-3-gd3cda09-dirty.darwin-amd64/zqd   OK
```

Here's an example (and temporary) release from this action:

https://github.com/brimsec/zq/releases/tag/v0-asset-test

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,12 +11,9 @@ jobs:
       with:
         go-version: '1.13'
     - uses: actions/checkout@v1
-    - name: run unit tests
-      run: make test-unit
-    - name: run system tests
-      run: make test-system
-    - name: create release assets
-      run: make create-release-assets
+    - run: make test-unit
+    - run: make test-system
+    - run: make create-release-assets
     - name: upload release assets
       uses: svenstaro/upload-release-action@1.1.0
       with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,27 @@
+name: release assets
+on:
+  push:
+    tags:
+      - v*
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/setup-go@v1
+      with:
+        go-version: '1.13'
+    - uses: actions/checkout@v1
+    - name: run unit tests
+      run: make test-unit
+    - name: run system tests
+      run: make test-system
+    - name: create release assets
+      run: make create-release-assets
+    - name: upload release assets
+      uses: svenstaro/upload-release-action@1.1.0
+      with:
+        repo_token: ${{ secrets.GITHUB_TOKEN }}
+        tag: ${{ github.ref }}
+        file: dist/release/*
+        file_glob: true
+        overwrite: true
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,19 @@ build:
 install:
 	@go install -ldflags='$(LDFLAGS)' ./cmd/...
 
+create-release-assets:
+	for os in darwin linux windows; do \
+		zqdir=zq-$(VERSION).$${os}-amd64 ; \
+		rm -rf dist/$${zqdir} ; \
+		mkdir -p dist/$${zqdir} ; \
+		GOOS=$${os} GOARCH=amd64 go build -ldflags='$(LDFLAGS)' -o dist/$${zqdir} ./cmd/... ; \
+	done
+	rm -rf dist/release && mkdir -p dist/release
+	cd dist && for d in zq-$(VERSION)* ; do \
+		zip -r release/$${d}.zip $${d} ; \
+	done
+
 clean:
 	@rm -rf dist
 
-.PHONY: vet test-unit test-system clean build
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -26,6 +26,21 @@ build:
...
+	for os in darwin linux windows; do \
+		zqdir=zq-$(VERSION).$${os}-amd64 ; \
+		rm -rf dist/$${zqdir} ; \
+		mkdir -p dist/$${zqdir} ; \
+		GOOS=$${os} GOARCH=amd64 go build -ldflags='$(LDFLAGS)' -o dist/$${zqdir} ./cmd/... ; \
+	done
+	rm -rf dist/release && mkdir -p dist/release
+	cd dist && for d in zq-$(VERSION)* ; do \
+		zip -r release/$${d}.zip $${d} ; \
+	done
```
mikesbrown: This isn't a blocker; it's a commentary on the zq Makefile and the direction it's going.

I understand that Go tooling obviates some of the traditional reasons to use make, but it seems like where we can, we could still make traditional use of Makefiles: real, non-phony targets, and dependency chains. If I run `make create-release-artifacts` twice, the work is done twice. In most make user experiences this is not expected.

There's also a bit of `go build` duplication happening here with `make build`.

The way I could see this working with deps conceptually is something like:
```
non-phony-arch-specifc-build:
        go build instructions to build arch-specific-build-target

build:
       phony shortcut for a single arch-specifc-build for the calling arch; possibly creates a symlink from dist/zq to dist/arch/zq

non-phony-arch-release-zip: non-phony-arch-specifc-build
        zip action against the dep to produce the target

all-non-phony-release-zips: individual arch-release zips

create-release-artifacts: all-non-phony-release-zips

publish-release-artifacts: all-non-phony-release-zips
        leave as is, basically

.PHONY: build create-release-artifacts publish-release-artifacts
```
One way around fewer phony targets is to emit stamps via `touch(1)` when the target's work is done. That might be appropriate for `all-non-phony-release-zips` in this example.

This lets you remove the `rm -rf` setup you need (that should be part of `make clean`). A second `make` run then won't reduplicate work it doesn't need to. The `go build` duplication is gone. Finally, it removes a lot of shell from Makefile and looks more Makefile-y.

Makefile
```diff
@@ -26,6 +26,21 @@ build:
...
+
+publish-release-artifacts:
```
mikesbrown: If leaving as is, I suggest you add create-release-artifacts and publish-release-artifacts to the .PHONY list below.

alfred-landrum:
@nwt : as suggested offline, I've changed from using google storage to Github Release assets. I've update the first PR comment to reflect the change, PTAL.

Makefile
alfred-landrum: In the update to use github assets, I did update the phony line.

Makefile
alfred-landrum: @mikesbrown : thanks for the comments, and the +1 . As you say, since there's a great deal of optimization within go tooling for building packages & caching the results, there's not as much of a need for exposing the source to target dependencies within make.

We have some other Makefile related changes we should do - like incoporating the zql/Makefile work into this top level one. I'll go ahead with this PR for now, but we'll plan on making larger changes & establishing dependency changes in the future.

alfred-landrum:
I've downloaded the release assets and verified them:
* mac: will not work immediately; you have to allow zq to run via the System Preferences security tab. Not a surprise as we're not yet signing for Darwin.
* windows: zq.exe runs as expected
* linux: zq runs as expected.

.github/workflows/release.yml
```diff
@@ -0,0 +1,27 @@
...
+        go-version: '1.13'
+    - uses: actions/checkout@v1
+    - name: run unit tests
+      run: make test-unit
+    - name: run system tests
+      run: make test-system
+    - name: create release assets
+      run: make create-release-assets
+    - name: upload release assets
+      uses: svenstaro/upload-release-action@1.1.0
```
nwt: Maybe use https://github.com/actions/upload-release-asset?

alfred-landrum: Unfortunately, it only supports uploading a single asset at a time, and doesn't support file patterns or globs when selecting the assets to upload. I started by trying to use it, then shifted over to this one.

PR #323, opened 2020-02-14
Accept negative integers in filter expressions

aswan:


alfred-landrum:
I'm not very familiar with the peg format, so apologies for maybe an obvious question : was it the use of `sinteger` in `searchValue` (w/o these changes) that was preventing the use of negative integers?

aswan:
> I'm not very familiar with the peg format, so apologies for maybe an obvious question : was it the use of `sinteger` in `searchValue` (w/o these changes) that was preventing the use of negative integers?

Yeah, sinteger is defined here: https://github.com/brimsec/zq/blob/250507027e943eaee4f6e0d2aba1c8428af73639/zql/zql.peg#L518-L521
That definition allows any sequence of digits, but no leading sign.

Slightly confusingly, the leading s means "string" (i.e., the "result" of that rule is the string representation of the matching characters rather than the actual integer).  This patch changes "integer" and "sinteger" to be signed and adds "unsignedInteger" and "suint" for the unsigned equivalents.

PR #324, opened 2020-02-14
Avoid fetching zql parser build tools unnecessarily

aswan:
Check the versions of pegjs, pigeon, and goimports rather than calling
"go get" and "npm install" every time we run make in the zql directory.

### DIFF
--- a/zql/Makefile
+++ b/zql/Makefile
@@ -4,12 +4,13 @@ PEGJS = $(deps)/bin/pegjs
 PIGEON = $(deps)/bin/pigeon
 GOIMPORTS = $(deps)/bin/goimports
 
-PEGJS_VERSION = $(shell test -x $(PEGJS) && $(PEGJS) --version)
-PEGJS_EXPECTED = 0.10.0
-PIGEON_VERSION = $(shell test -d $(deps) && cd $(deps); go list -m github.com/mna/pigeon || echo "")
-PIGEON_EXPECTED = v1.0.1-20190520151103-9fec3898cef8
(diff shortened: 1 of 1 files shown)


zql/Makefile
```diff
@@ -1,21 +1,53 @@
...
-zql.go:
+$(PEGJS):
+ifeq (,$(findstring $(PEGJS_EXPECTED),$(PEGJS_VERSION)))
+	mkdir -p $(deps)
+	npm install --global --prefix $(deps) pegjs@0.10.0
+endif
+
+GODEPS :=
+ifeq (,$(findstring $(PIGEON_EXPECTED),$(PIGEON_VERSION)))
+  GODEPS += github.com/mna/pigeon@v1.0.1-0.20190520151103-9fec3898cef8
```
mikesbrown: Oh, maybe this could be the reason. It's maddening that the substring after `@` doesn't match `PIGEON_EXPECTED`.

zql/Makefile
```diff
@@ -1,21 +1,53 @@
...
 
 run: all
 	go run ./main
 
-.PHONY: zql.go
-zql.go:
+$(PEGJS):
+ifeq (,$(findstring $(PEGJS_EXPECTED),$(PEGJS_VERSION)))
+	mkdir -p $(deps)
+	npm install --global --prefix $(deps) pegjs@0.10.0
```
mikesbrown: Any reason not to replace `0.10.0` with `$(PEGJS_EXPECTED)`?

aswan: We can do it here, though as you noted we can't easily do it for go modules.
I expect we'll change versions quite infrequently (if at all!) so changing in two places if/when we do is okay and not worth optimizing down to one place...

PR #276, opened 2020-01-31
Implement Alias Type

mattnibs:
Done according to the zng spec. Includes end-to-end reading and
writing for bzng, zng and zjson. Support for aliases added to
sort, reducer, and groupby procs.

TODO (in this pr):
- [x] zjson support

PR #325, opened 2020-02-16
Add integer types from the zng spec

aswan:


zx/coerce.go
```diff
@@ -51,53 +56,145 @@ func CoerceToFloat64(in zng.Value) (float64, bool) {
...
-func CoerceToInt(in zng.Value) (int64, bool) {
-	var out int64
+// values.
+func CoerceToInt(in zng.Value, typ zng.Type) (zng.Value, bool) {
+	var i int64
 	var err error
-	body := in.Bytes
-	switch in.Type.ID() {
+
+	switch in.Type.(type) {
```
mattnibs: Doing this is going to break aliases.

zx/coerce.go
```diff
@@ -51,53 +56,145 @@ func CoerceToFloat64(in zng.Value) (float64, bool) {
...
-		if v > math.MaxInt64 {
-			return 0, false
+		// can't be cast to integer
+		return zng.Value{}, false
+	}
+
+	switch typ.(type) {
+	case *zng.TypeOfInt16:
+		if i < math.MinInt16 || i > math.MaxInt16 {
+			return zng.Value{}, false
```
mattnibs: Also going to break aliases.

zx/coerce.go
aswan: can you elaborate?

mattnibs: What if you pass an aliased type into here?

mattnibs: You have to either unwrap it or use `in.Type.ID()`

filter/filter_test.go
```diff
@@ -212,6 +215,51 @@ func TestFilters(t *testing.T) {
...
+		{"u64 < 1", records[17], true},
+		{"u64 > 1", records[17], false},
+
+		{"b = 255", records[18], true},
+		{"i16 = 32767", records[18], true},
+		{"u16 = 65535", records[18], true},
+		{"i32 = 2147483647", records[18], true},
+		{"u32 = 4294967295", records[18], true},
+		{"i64 = 9223372036854775807", records[18], true},
+		// can't represent large unsigned 64 bit values in zql...
```
alfred-landrum: For my own info, not a comment on the PR: what prevents zql from handling large uint64's?

zng/type.go
```diff
@@ -67,18 +67,25 @@ type Type interface {
...
-	TypePort     = &TypeOfPort{}
 	TypeIP       = &TypeOfIP{}
+	TypePort     = &TypeOfPort{}
 	TypeNet      = &TypeOfNet{}
-	TypeEnum     = &TypeOfEnum{}
+	TypeTime     = &TypeOfTime{}
+	TypeDuration = &TypeOfDuration{}
 	TypeNull     = &TypeOfNull{}
+
+	// XXX kill me:
```
alfred-landrum: Please don't add snarky comments to the repo.

zx/coerce.go
```diff
@@ -148,9 +245,9 @@ func CoerceToPort(in zng.Value) (uint32, bool) {
 		out = uint32(v)
 	case zng.IdUint64:
 		var v uint64
-		v, err = zng.DecodeCount(body)
+		v, err = zng.DecodeUint(body)
 		// check for overflow
-		if v > math.MaxInt64 {
+		if v > math.MaxInt16 {
```
alfred-landrum: Would you explain why MaxInt16 is the limit? Also, why this uint64 is cast down to a uint32 before returning from the function as a uint64?

zx/coerce.go
aswan: > What if you pass an aliased type into here?

I beileve it is broken, whether we use `.(type)` or `.ID()`.  There's a separate PR that addresses this.

filter/filter_test.go
aswan: > what prevents zql from handling large uint64's?
We don't have any syntax to indicate that an integer literal should be unsigned.  This is described in an issue on Jira, we could either automatically turn values that are too big to fit into int64 (but that do fit into uint64) or we could add new syntax (e.g., C style "1234U" or something)

zx/coerce.go
mattnibs: In master, calls to `.ID()` on an alias will recurse to the original primitive type of the alias (see https://github.com/brimsec/zq/blob/master/zng/alias.go#L22). This does the same thing as the function you created in `AliasedType` in #326 .

zx/coerce.go
aswan: Whoops, should be MaxUint16 for a port, will fix.  I don't understand the second part of the question, the return value is a uint32 (which appears wrong to me, it should be a uint16 but I don't think that needs to be addressed in this PR)

zx/coerce.go
mattnibs: I had switched these coercion functions in the original alias pr to use id's instead of types for this very purpose.

aswan: > I had switched these coercion functions in the original alias pr to use id's instead of types for this very purpose.

Great, it would have been nice to have handled compare.go as well and to have added tests to prevent future regressions.

mattnibs: Yep, I thought I had handled all cases, but forgot about compare.go.

mattnibs: FWIW I did add a test for groupby (https://github.com/brimsec/zq/pull/276/files#diff-9fcad515342ab12c2e5bf1a1d096caa5). Had I considered compare as well (which I didn't) I would have add a test there.

zx/compare.go
```diff
@@ -143,6 +151,46 @@ func CompareInt64(op string, pattern int64) (Predicate, error) {
...
+	CompareUint, ok1 := compareUint[op]
+	CompareFloat, ok2 := compareFloat[op]
+	if !ok1 || !ok2 {
+		return nil, fmt.Errorf("unknown int comparator: %s", op)
+	}
+	// many different zeek data types can be compared with integers
+	return func(val zng.Value) bool {
+		zv := val.Bytes
+		switch val.Type.(type) {
+		case *zng.TypeOfInt16, *zng.TypeOfInt32, *zng.TypeOfInt64:
```
mattnibs: You should also switch on `val.Type.ID()` here as well to avoid the same alias issue.

zx/compare.go
```diff
@@ -105,20 +110,23 @@ func CompareInt64(op string, pattern int64) (Predicate, error) {
 	return func(val zng.Value) bool {
 		zv := val.Bytes
 		switch val.Type.(type) {
-		case *zng.TypeOfInt:
-			// we can parse counts and ports as an integer
```
mattnibs: These should also switch on `val.Type.ID()`.

zx/compare.go
aswan: Given that comparisons aren't currently working in master, do you mind if I make that change in #326?  (which I still need to update to use `ID()` instead of `.type()`)

mattnibs: Sounds good!

PR #326, opened 2020-02-18
Allow comparisons with aliased types

aswan:


zng/alias.go
```diff
@@ -45,3 +45,12 @@ func (t *TypeAlias) StringOf(zv zcode.Bytes, out OutFmt, b bool) string {
 func (t *TypeAlias) Marshal(zv zcode.Bytes) (interface{}, error) {
 	return t.Type.Marshal(zv)
 }
+
+func AliasedType(typ Type) Type {
```
alfred-landrum: Would you consider naming this `Unalias`? I find that would be a little clearer, and would stutter less in its usage:
`zng.Unalias(val.Type).(type)` 
`zng.Unalias(v.Type) != zng.TypeFoo`

zx/compare.go
```diff
@@ -199,7 +199,7 @@ func CompareFloat64(op string, pattern float64) (Predicate, error) {
 	}
 	return func(val zng.Value) bool {
 		zv := val.Bytes
-		switch val.Type.(type) {
+		switch zng.AliasedType(val.Type).(type) {
```
mattnibs: You should be using `val.Type.ID()` instead of this (at least this is the convention that @mccanne wanted).

aswan: I don't have any objection to changing but:
1. We have tons of places in the code that use the `(type)` technique, if we want to make this change, that seems like something orthogonal that should be done through the tree in a separate PR
2. What's the rationale for changing?

zng/alias.go
alfred-landrum: This may be moot with the discussion of the `val.Type.ID()` usage below, but fyi, using `Unalias` came to mind partly by how go errors are unwrapped, both in the popular `pkg/errors` package from Dave Cheney, and now via the go standard library:
https://godoc.org/github.com/pkg/errors#Unwrap
https://golang.org/pkg/errors/#Unwrap

zx/compare.go
mattnibs: 1. All the alias changes that I made use `Type.ID()` so if we should decide to go this route we should change those as well (in another pr of course).
2. Probably better @mccanne to explain, but I'll take my best shot. I personally find it confusing that primitive types are instances of `TypeOf*` whereas records/sets/arrays don't follow this pattern (for good reasons). A better reason: With aliases functionality, now every time you want to access a value's type you'll have to unwrap it first; I imagine that it would be easy to forget about why this needs to be done. If you forget to unwrap the functionality would appear to generally work except that one time you come across an alias. The convention to always do type switching using `Type.ID()`, seems less error prone and requiring less institutional knowledge from  contributors/reviewers.

PR #328, opened 2020-02-18
zqd: log listener address to stdout

mattnibs:


cmd/zqd/listen/command.go
```diff
@@ -34,5 +37,30 @@ func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
...
+	return http.Serve(ln, handler)
+}
+
+func newLogger() *zap.Logger {
+	encoder := zap.NewProductionEncoderConfig()
+	encoder.CallerKey = ""
+	c := zap.Config{
+		Level:            zap.NewAtomicLevelAt(zap.DebugLevel),
+		Development:      false,
+		Encoding:         "json",
```
aswan: Why json by default?

mattnibs: Structured logs are and easily readable by jq! But also for the brim app I was going to have the os grab a random open port to use. I was going to parse the output from stdout so the client app can know which port was selected.

cmd/zqd/listen/command.go
```diff
@@ -34,5 +37,30 @@ func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
...
+		EncoderConfig:    encoder,
+		OutputPaths:      []string{"stdout"},
+		ErrorOutputPaths: []string{"stderr"},
+	}
```
nwt: Maybe shorten this since it's the production config with a few tweaks?
```go
c := zap.NewProductionConfig()
c.Level = zap.NewAtomicLevelAt(zap.DebugLevel)
c.Sampling = nil
c.EncoderConfig.CallerKey = ""
```

mattnibs: They're both ugly but ok.

PR #330, opened 2020-02-19
support truncated output in markdown examples

mikesbrown:
Extend the markdown info string for `zq-output` to include an optional `head:N` directive, where only the first N lines of output of a `zq-command` are verified against its `zq-output`.

To find these we must extend our AST walk function. Rename `ZQExamplePair` to `ZQExampleInfo`; add a new struct member to hold the number of lines to compare. Do the same for `ZQExampleTest`.

Move the responsibility of actually running the `zq-command` to a `ZQExampleTest` `Run()` method. This also handles any truncation.

Update all remaining examples in `zql/docs`.


mikesbrown:
#329 could come first if there is a desire to make `make test-system` not run these.

tests/zq_example.go
```diff
@@ -102,28 +89,75 @@ import (
...
 
-// ZQExamplePair holds a ZQ example as found in markdown.
-type ZQExamplePair struct {
-	command *ast.FencedCodeBlock
-	output  *ast.FencedCodeBlock
+// ZQExampleInfo holds a ZQ example as found in markdown.
+type ZQExampleInfo struct {
+	command         *ast.FencedCodeBlock
+	output          *ast.FencedCodeBlock
+	outputLineCount uint8
```
alfred-landrum: No need to be as specific as `uint8`, `uint` is acceptable.

tests/zq_example.go
```diff
@@ -102,28 +89,75 @@ import (
...
-// from a ZQExamplePair.
+// from a ZQExampleInfo.
 type ZQExampleTest struct {
-	Name     string
-	Command  []string
-	Expected string
+	Name            string
+	Command         []string
+	Expected        string
+	OutputLineCount uint8
```
alfred-landrum: No need to be as specific as `uint8`, `uint` is acceptable.

mikesbrown:
Rebase. Incidentally some of the examples' outputs have changed.

PR #331, opened 2020-02-19
Remove zng enum type

aswan:
Since changes that are likey to be incompatible with the current
implementation are planned for this type, remove it for now.  Note
that this does not affect the "enum" type in legacy Zeek logs which
is handled separately.


expr/sort.go
```diff
@@ -180,7 +180,7 @@ func lookupSorter(typ zng.Type) comparefn {
 			return -1
 		}
 
-	case zng.TypeString, zng.TypeEnum:
+	case zng.TypeString:
```
mattnibs: Tangential to this pr but does this work for aliases? Are the values unwrapped before they get here?

aswan: I didn't follow the initial implementation of aliases closely.  I thought the desired behavior was that an aliased type is sorted separately?  From a quick glance, that looks to me like what `sort_test.go` is testing?  (though I may have mis-read it)

PR #334, opened 2020-02-19
Perf comparison script should do a shallow clone of zq-sample-data

philrz:
@alfred-landrum made me hip to the fact that we were doing the `git clone --depth=1` in @mikesbrown's docs automation to deal with the bulky history there. That made me realize I should do the same in other contexts, such as this script that can be used to check current `zq` perf when working with the various formats.


philrz:
@mikesbrown: Yes, seems like a good idea. I'll take a shot at it in another PR.

PR #335, opened 2020-02-19
Perf comparison script referenced from Makefile

philrz:
@mikesbrown correctly pointed out in https://github.com/brimsec/zq/pull/334 that the automated doc example checker and my perf comparison script were each cloning their own copy of zq-sample-data. Here I've taken a shot at removing the redundancy by making the the perf comparison script a Makefile target and having it work off the same single copy of the sample data.

While I was in there, I noticed that the perf comparison script hadn't been touched in the time since we started supporting GZIP'ed input files, so I've updated that to simplify.

I did retest this works on both macOS and Linux, but there's a catch: There's a known open bug right now with BZNG input format that makes this script fail to run to completion. I've made a note-to-self to re-run this when that bug is fixed, at which point I'll put up a PR with updated perf numbers.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,4 +54,4 @@ create-release-assets:
 clean:
 	@rm -rf dist
 
-.PHONY: vet fmt sampledata test-unit test-system test-heavy build install create-release-assets clean
+.PHONY: vet fmt sampledata test-unit test-system test-heavy perf-compare build install create-release-assets clean
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,9 @@ test-system: build
 test-heavy: build $(SAMPLEDATA)
 	@go test -v -tags=heavy ./tests
 
+perf-compare: build $(SAMPLEDATA)
+	scripts/comparison-test.sh
+
 build:
 	@mkdir -p dist
 	@go build -ldflags='$(LDFLAGS)' -o dist ./cmd/...
--- a/scripts/comparison-test.sh
+++ b/scripts/comparison-test.sh
@@ -6,14 +6,7 @@ set -eo pipefail
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 cd "$SCRIPT_DIR" || exit 1
 
-DATA="zq-sample-data"
-DATA_REPO="https://github.com/brimsec/zq-sample-data.git"
-if [ -d "$DATA" ]; then
-  (cd "$DATA"&& git pull) || exit 1
-else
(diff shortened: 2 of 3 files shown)


scripts/comparison-test.sh
```diff
@@ -6,14 +6,7 @@ set -eo pipefail
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 cd "$SCRIPT_DIR" || exit 1
 
-DATA="zq-sample-data"
-DATA_REPO="https://github.com/brimsec/zq-sample-data.git"
-if [ -d "$DATA" ]; then
-  (cd "$DATA"&& git pull) || exit 1
-else
-  git clone --depth=1 "$DATA_REPO"
-fi
```
mikesbrown: It would be fine with me to replace line 9 as you did and retain lines 10-15. I agree line 16 can be deleted.

scripts/comparison-test.sh
```diff
@@ -6,14 +6,7 @@ set -eo pipefail
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 cd "$SCRIPT_DIR" || exit 1
 
-DATA="zq-sample-data"
-DATA_REPO="https://github.com/brimsec/zq-sample-data.git"
-if [ -d "$DATA" ]; then
-  (cd "$DATA"&& git pull) || exit 1
```
mikesbrown: If you do retain this, nit: use spacing, as in
```bash
  (cd "$DATA" && git pull) || exit 1
```

Makefile
```diff
@@ -29,6 +29,9 @@ test-system: build
 test-heavy: build $(SAMPLEDATA)
 	@go test -v -tags=heavy ./tests
 
+perf-compare: build $(SAMPLEDATA)
```
mikesbrown: The `.PHONY` list is getting long but it seems correct to make this phony, too. https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html

scripts/comparison-test.sh
philrz: If you don't care either way, I'll leave it as-is with the excess lines removed and hence encourage running it via `make`.

Makefile
philrz: Thanks for the education. I've added it to the phony list and re-tested.

scripts/comparison-test.sh
philrz: Since I'm opting to not retain, this nit should not apply.

PR #336, opened 2020-02-20
Accept negative floats in the zql grammar

aswan:


aswan:
> Would be nice to have a test for this.

Exercised in #338

PR #337, opened 2020-02-20
Remove CompareUint64()

aswan:
This was erroneously introduced in #325.  The CompareFoo methods are
called on the type of the literal that appears on the RHS of a filter
with a comparison operator but there's no way to create an unsigned
literal in zql so this code is all unreachable.


aswan:
This came in via #338

PR #340, opened 2020-02-20
Support absolute file paths for spaces

mattnibs:
Augment the space info and space packet endpoints to accept url
path encoded (https://golang.org/pkg/net/url/#PathEscape) paths.

Also:
- Add some testing to search and space info endpoints.
- Remove space list endpoint.


zqd/api/api.go
```diff
@@ -2,11 +2,26 @@ package api
...
+	"net/http"
+	"net/url"
 
 	"github.com/brimsec/zq/pkg/nano"
 	"github.com/brimsec/zq/zio/zjsonio"
+	"github.com/gorilla/mux"
 )
 
+// ExtractSpace returns the unescaped space from the path of a request.
+// XXX is this in the right package?
```
alfred-landrum: In the existing package layout, I think this function should live in the space package.

zqd/search/endpoint_test.go
```diff
@@ -0,0 +1,59 @@
+package search_test
```
alfred-landrum: I don't see a reason for this to be a different package from the other files in this directory.

zqd/space/endpoint_test.go
```diff
@@ -0,0 +1,46 @@
+package space_test
```
alfred-landrum: I don't see a reason for this to be a different package from the other files in this directory.

zqd/zqdtest/zqdtest.go
```diff
@@ -0,0 +1,58 @@
+package zqdtest
```
alfred-landrum: I'm not sure this set of functions really worthy of a separate package. The logic in `NewRequest` and `ExecRequest` seem too small to be separated into functions.

For CreateSpace/WriteSpace: Consider folding the content of WriteSpace into CreateSpace, and then making CreateSpace *not* take a testing.T, and just return an error. That new CreateSpace function could live inside the api package, with a minimal set of (non-test specific) added imports.

zqd/search/endpoint_test.go
mattnibs: I was follow the go convention of placing tests in a separate `*_test` package. I like doing this especially when I'm building libs because it allows me to use my package as an end user would, and also so I avoid testing internal functionality.

zqd/space/endpoint_test.go
mattnibs: See answer above.

zqd/api/api.go
mattnibs: This can't go in package `space` because it's also needed in package `packet` and thus would cause a circular dependency. As `zqd` continues to grow I imagine we'll have some package dedicated to `http` helper functions (maybe extending mux.Router or something).

mattnibs: But now that I'm thinking about this maybe the handler for `/space/{space}/packet` should live in `space` pkg, and deal with decoding the space path, and then call out the packet library with everything parsed.

zqd/search/endpoint.go
```diff
@@ -84,17 +89,15 @@ func httpError(w http.ResponseWriter, msg string, code int) {
...
-	}
 	query, err := parseSearchRequest(r.Body)
 	if err != nil {
 		httpError(w, err.Error(), http.StatusBadRequest)
 		return
 	}
-	//logger.Debug("parseSearchRequest", zap.Stringer("query", query))
-	dataPath := filepath.Join(".", query.Space, "all.bzng") //XXX need root dir param
+	//XXX We do not support root dir param right now. If query.Space is an
+	// absolute path we will ignore the root dir param. For now relative paths
```
aswan: Do we have a good reason to accept absolute paths?  Given that we're listening on INADDR_ANY, accepting unauthenticated requests to read an arbitrary file on the filesystem strikes me as a bad idea.

mattnibs: > Given that we're listening on INADDR_ANY

I'm going to change this to localhost. Does that resolve your concerns?

aswan: > I'm going to change this to localhost. Does that resolve your concerns?

It lessens them a little bit, but unauthenticated requests to read any file still seems like a bad idea.  Can we invert this discussion: what's the product requirement here?  from there maybe we can figure out a solution that wouldn't be so trivially exploitable.

mattnibs:
I hear your concerns. We’re doing this right now because it’s the simplest
way to get this working w the requirements for brim: we need to be able to
create a space next to wherever in the file system the user selected a pcap

zqd/search/endpoint.go
mattnibs: I hear your concerns. We’re doing this right now because it’s the simplest
way to get this working w the requirements for brim: we need to be able to
create a space next to wherever in the file system the user selected a pcap

mattnibs: > Given that we're listening on INADDR_ANY

I should say I'm going to change that in the brim app, the default functionality here will remain the same.

aswan:
> we need to be able to create a space next to wherever in the file system the user selected a pcap

I don't understand this sentence.  Is the requirement that all the temporary files we create (e.g., `all.bzng`, extracted pcaps, etc) be put in the same directory as the pcap file they are dervied from?  That seems like a strange requirement as it means the app wouldn't be able to handle a file in a read-only directory.  ..which seems like a scenario that is quite likely to occur in practice.

zqd/search/endpoint_test.go
alfred-landrum: I don't think having a separate *package* ending in `_test` is a go convention. I see that pattern in the zq repo a couple of times, once in the `proc` package, to avoid a dependency, but that's not a common practice I see in well known go systems.

zqd/search/endpoint.go
mattnibs: > I don't understand this sentence. Is the requirement that all the temporary files we create (e.g., all.bzng, extracted pcaps, etc) be put in the same directory as the pcap file they are dervied from?

Yes

> That seems like a strange requirement as it means the app wouldn't be able to handle a file in a read-only directory. ..which seems like a scenario that is quite likely to occur in practice.

This was a product decision. I'll defer to @alfred-landrum and @siskojr

aswan: Okay, if that's the requirement, the user can only work with one file at a time right?  Given that we have to run zeek and then convert to bzng, restarting zqd with a different root would not be noticeably slower, would not require changing zqd, and would not entail having zqd read arbitrary files.

zqd/search/endpoint_test.go
mattnibs: I'm seeing it here in the go standard library: https://github.com/golang/go/blob/master/src/sort/search_test.go

mattnibs: Its also nice because you don't populate the package namespace with any test-related internal functions.

zqd/search/endpoint.go
mattnibs: > the user can only work with one file at a time right?

I believe you can have multiple tabs and multiple windows that point at different pcaps.

aswan: Well, I think we're in agreement that this is not a good long-term solution?  If the same-directory requirement is firm, one instance of zqd per pcap (or per unique directory) seems like a reasonable long-term solution?
If that's too much work in the short term, I understand, but please file a follow-up issue to address this, I really think that shipping with such a wide open unauthenticated endpoint is a big mistake.

zqd/search/endpoint_test.go
mattnibs: Also mentioned in the [documentation](https://golang.org/cmd/go/#hdr-Testing_flags) for the `go test` command:

> Test files that declare a package with the suffix "_test" will be compiled as a separate package, and then linked and run with the main test binary.

zqd/search/endpoint.go
mattnibs: Not opposed to this, but yes this does seem like a lot more work on the client-side. I'll discuss with the brim team.

zqd/search/endpoint_test.go
alfred-landrum: I just verified how frequently the `zq` repo uses this pattern, and see it is used in many of the packages here. I don't want to pollute your PR with comments about this, and see this is consistent with the repo as-is. I do think we should change, but I'll move that discussion off of your PR.

zqd/api/api.go
mattnibs: Alright I resolved this by moving the all endpoints into package `zqd` and made `ExtractSpace` an internal function of `zqd`

mccanne:
I would vote for keeping the space API we had and augmenting it with a way to upload a pcap.    The advantage of this is our existing tools and code just works and we don't make unnecessary work for us.  

Also, the existing work we did on the APIs was fairly well thought out and, in the very near future, when we want brim to reach over the network to a remote zqd, we'll start putting back in then what we took out now.

It seems like we have convolved ux design of a desktop-scale open-brim experience with a re-do of the REST API.  I don't think we need to make changes to the existing API;  it just needs augmentation.  e.g., GET /space went away and broke tip of brim as well as my scripts. 

Here are a few suggestions:
* keep the local spaces directory in ~/Library/Brim (or equivalent for non mac)
* put back the GET /space endpoint (the space names need not be exposed in the UI, but probably later)
* add POST /space/{space}/pcaps which takes an absolute path of the pcap on the local filesystem as a parameter in the JSON body (support just a single post for now).  (this could also post a binary pcap over the network instead of using the file path... but later) 
* add GET /space/{space}/pcaps to return a list of the absolute file names of pcaps that have been ingested into that space

On startup, the app could list the spaces and get each absolute path to allow the user to open preloaded pcaps.

On pcap open, the app could generate a space name from the file name (or auto generate one if they aren't exposed to the user, space1, space2, etc) and create the space, then post the pcap file.  (We could stream back json status updates on the POST request, later.)

If we want a leave-no-traces ux (probably not given the zeek bottleneck), the app could do DELETE space/* on exit.

This approach accomplishes all the ux we've talked about while maintaining compatibility with our existing network-services-oriented API and results in fewer new moving parts.

Also, this sets us up to later support:
* posting multiple pcaps at different times to the same space
* post pcaps over the net
* do everything we can do locally over the net to a zqd running on a server
* support multiple connections to different data sources with tabbed searches running to different locations
* etc

mattnibs:
> GET /space went away and broke tip of brim as well as my scripts.

Out of curiosity what were your scripts doing? Were you running brim against this branch?

mattnibs:
> add POST /space/{space}/pcaps which takes an absolute path of the pcap on the local filesystem as a parameter in the JSON body (support just a single post for now). (this could also post a binary pcap over the network instead of using the file path... but later)

Does this endpoint just the path of that pcap in the space?

mccanne:
> > GET /space went away and broke tip of brim as well as my scripts.
> 
> Out of curiosity what were your scripts doing? Were you running brim against this branch?

I have aliases that I run; script was the wrong word.  Also, we have existing code that we may want to cut and paste into an open-source repo that uses the logic of the existent API (fetch a list of spaces then do something with them).

mccanne:
> > add POST /space/{space}/pcaps which takes an absolute path of the pcap on the local filesystem as a parameter in the JSON body (support just a single post for now). (this could also post a binary pcap over the network instead of using the file path... but later)
> 
> Does this endpoint just the path of that pcap in the space?

I just spoke to @alfred-landrum and I realized I didn't finish my stream of consciousness here... the ux of putting the data next to the pcap can be accomplished for the pcap(s)-in-the-same-dir approach by having each space directory have a json config file that points to the .brim dir next to the pcap, essentially keeping the existing state management on the services side (I have a list of places I put pcaps and zeek logs and meta-info associated with them) instead of the app side (I keep a list of .brim dirs in localstorage).  The app can query the space list just as easily as the loading the list of recent pcaps from local storage.    This lets us keep the general model while giving the desktop ux experience so we'll be teed up to do zqd over the network before we know it.

mattnibs:
Closing whilst I rework this for new changed scope.

PR #342, opened 2020-02-20
convert most tests in tests/ to YAML ztests

nwt:
* Add package ztest to run formulaic tests defined in YAML files that
  look like this:
```yaml
    zql: count()

    input: |
      #0:record[i:int64]
      0:[1;]
      0:[2;]

    output: |
      #0:record[count:uint64]
      0:[2;]
```
* Convert tests in tests/formats and most of those in tests/suite to
  YAML.  As a side effect, both "make test-unit" and "make test-system"
  now run all of the converted tests.  (test-unit uses
  driver.(*Driver).Run, while test-system uses the dist/zq executable).

***

## Future Work

* Move YAML ztests into the packages they're intended to test. (I put them in the same directory as the tests they replaces to make this change easier to digest.)
* Convert most of the tests in proc/ and some of those in filter/ and zio/* to YAML.
* Convert the remaining tests in tests/suite/errors/. (This requires adding something like an "error expected" field to ztest.Ztest.)
* Pick a directory in which to put regression tests.


aswan:
I'm not very familiar with yaml, is `!!binary` handled by the parser?  For bzng data, it would be great if we could write the data out in hex rather than base64.  If somebody is looking at a particular test (or diffs to a particular test) it is tedious to be sure, but feasible to read through the byte-by-byte representation of binary data.  This isn't really practical with base64.

ztest/ztest.go
```diff
@@ -0,0 +1,219 @@
...
+		})
+	}
+}
+
+// ZTest defines a ztest.
+type ZTest struct {
+	ZQL          string `yaml:"zql"`
+	Input        string `yaml:"input"`
+	OutputFormat string `yaml:"output-format,omitempty"`
+	Output       string `yaml:"output`
```
mikesbrown: Missing trailing double-quote for after `"output`.

tests/ztest_test.go
```diff
@@ -0,0 +1,30 @@
+package tests
```
mikesbrown: ztests are also being run during `make test-heavy`. https://circleci.com/gh/brimsec/zq/2970 . I understand we're in a changing world.

nwt:
@aswan: `!!binary` is handled by the parser. It is, unfortunately, the only means YAML defines for encoding non-UTF-8 byte sequences.

We could always define our own binary encoding above YAML. Do you think that's worth doing?

aswan:
> We could always define our own binary encoding above YAML. Do you think that's worth doing?

Maybe?  That could happen in a later follow-up for sure if it seems like it would be worthwhile.

tests/ztest_test.go
```diff
@@ -0,0 +1,30 @@
...
+		return err
+	})
+	require.NoError(t, err)
+	for d := range dirs {
+		d := d
+		t.Run(d, func(t *testing.T) {
+			t.Parallel()
+			ztest.Run(t, d)
+		})
+	}
```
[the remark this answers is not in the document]
nwt: I'd rather remove it entirely. 😄

My vision is to
* move all of the YAML ztests found by this logic into testdata/ztest/ subdirectories of the packages that they're intended to test,
* put a ztest_test.go containing `func TestZTest(t *testing.T) { ztest.Run(t, "testdata/ztest") }` in each of those packages,
* and then delete this file.

ztest/ztest.go
```diff
@@ -0,0 +1,219 @@
...
+			t.Fatalf("bad ZTEST_ZQ value %s: %s%s", zq, err, out)
+		}
+	}
+	fileinfos, err := ioutil.ReadDir(dirname)
+	if err != nil {
+		t.Fatal(err)
+	}
+	for _, fi := range fileinfos {
+		filename := fi.Name()
+		const dotyaml = ".yaml"
```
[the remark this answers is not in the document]
nwt: I agree that the current situation isn't ideal. My vision is that this will be the only code specifying the types of files to inspect.

ztest/ztest.go
```diff
@@ -0,0 +1,219 @@
...
+				t.Fatalf("%s: expected and actual outputs differ:\n%s", filename, diff)
+			}
+		})
+	}
+}
+
+// ZTest defines a ztest.
+type ZTest struct {
+	ZQL          string `yaml:"zql"`
+	Input        string `yaml:"input"`
```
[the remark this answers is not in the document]
nwt: I envision a similar future, though I'm not sure if we do all of it here with a single YAML schema or if we limit this to the current style of test and define a separate schema that adds these features (one that maps to pkg/test.ShellTest, perhaps).

ztest/ztest.go
nwt: 👍

ztest/ztest.go
```diff
@@ -0,0 +1,219 @@
...
+//      0:[2;]
+//
+//    output-format: table
+//
+//    output: |
+//      COUNT
+//      2
+//
+// Ztest YAML files for a package should reside in a subdirectory named
+// testdata/ztest.  The package should also contain a file named ztest_test.go.
```
alfred-landrum: Why is the recommendation to have a separate `ztest_test.go` file just to call `ztest.Run`?

The `filter` package has two files currently, `filter.go` and `filter_test.go`. Why shouldn't we add the example `func TestZTest(...)` to the existing `filter_test.go` file?

nwt: Good question. It highlights that a package contains ztests since the YAML files themselves are somewhat hidden in testdata/ztest/. I’m not certain it’s worth the overhead, though. If you agree, I’ll replace it with guidance that TestZTest for pkg should go into pkg_test.go.

alfred-landrum: 👍

PR #345, opened 2020-02-22
add pcap indexing and slicing

mccanne:
This commit adds packages to do indexing and slicing of pcap files.
It currently supports only legacy pcap format (i.e., no support
yet for pcap-ng though it is designed so pcap-ng can be dropped in
without too much difficulty).

We also added a "pcap" command in cmd/pcap.  "pcap index" will create
a JSON index for use by "pcap slice".  "pcap slice" takes this
index, a time range, and a flow filter as arguments and produces
an extracted pcap file.  All of these arguments are optioonal.
If the index is provided, then the packets that fall outside of
the indexed time range are skipped, which dramatically speeds up the
slicing when exctracting a small range out of a large pcap.
If the time range is specified, it is used by the index and only
packets that fall within the time range are scanned.  (If the time
range is given but no index is provided, then all packets are scanned
but only packets that fall within the time range are matched.)
If a flow filter is specified in the format "ip:port ip:port",
along with a protocol ("tcp" or "udp"), then only packets from that
flow are matched.

The pcap packages are also used direclty by the pcap endpoint in zqd.
For now, the stiching is in package zqd/pcapdemo but this will change
in a subsequent PR when the new file handling model is in place.


mccanne:
While this is under review, I'm going to put together some pcap test files and add tests for this functionality.  It's a bit tricky.

pcap/pcap.go
```diff
@@ -0,0 +1,234 @@
...
+	"github.com/brimsec/zq/pkg/peeker"
+	"github.com/google/gopacket"
+	"github.com/google/gopacket/layers"
+)
+
+// Reader wraps an underlying io.Reader to read packet data in PCAP
+// format.  See http://wiki.wireshark.org/Development/LibpcapFileFormat
+// for information on the file format.
+//
+// We currenty read v2.4 file format with nanosecond and microsecdond
```
mattnibs:
```suggestion
// We currenty read v2.4 file format with nanosecond and microsecond
```

mattnibs: Also I thought Libpcap classic (non ng) doesn't support nanosecond time resolution.

pcap/search.go
```diff
@@ -0,0 +1,172 @@
...
+	var n int
+	//XXX the .LayerType() method is returning Unknown for some reason
+	//outerLayer := pcap.LinkType().LayerType()
+	outerLayer := layers.LayerTypeEthernet
+	opts := gopacket.DecodeOptions{Lazy: true, NoCopy: true}
+	for {
+		block, err := pcap.ReadBlock(s.span)
+		if err != nil {
+			if err == io.EOF {
+				err = nil
```
mattnibs: You should do a break here right? If no packets are found you want to return `ErrNoPacketsFound`.

pcap/search.go
```diff
@@ -0,0 +1,172 @@
...
+
+// Search describes the parameters for a packet search over a pcap file.
+type Search struct {
+	span    nano.Span
+	matcher Predicate
+	id      string
+}
+
+// NewSearch creates a new search object.
+func NewSearch(span nano.Span, proto string, srcHost string, srcPort *uint16, dstHost string, dstPort *uint16) (*Search, error) {
```
mattnibs: You should drop the port pointers and just return an error `ports == 0` (if the proto isn't icmp).

pkg/ranger/envelope.go
```diff
@@ -0,0 +1,105 @@
+package ranger
```
mattnibs: nit: Would be nice to have a brief package comment stating what this package does (it isn't obvious).

pcap/pcap.go
```diff
@@ -0,0 +1,234 @@
...
+	r.linkType = layers.LinkType(r.byteOrder.Uint32(hdr[20:24]))
+	return nil
+}
+
+func (r *Reader) ReadBlock(span nano.Span) ([]byte, error) {
+	header := r.header
+	if header != nil {
+		r.header = nil
+		return header, nil
+	}
```
mattnibs: Seems odd that `ReadBlock` that takes a `nano.Span` as an argument would return a non-packet block. Why not have this not return the header?

pcap/slicer.go
```diff
@@ -0,0 +1,105 @@
...
+	return 0, io.EOF
+}
+
+func (s Slice) Overlaps(x Slice) bool {
+	return x.Offset >= s.Offset && x.Offset < s.Offset+x.Length
+}
+
+func (s Slice) NewReader(r io.ReaderAt) *io.SectionReader {
+	return io.NewSectionReader(r, int64(s.Offset), int64(s.Length))
+}
```
mattnibs: I would keep these grouped with the `Slice` type declaration, probably moving the type declaration here.

pcap/search.go
```diff
@@ -0,0 +1,172 @@
...
+	switch proto {
+	//case "icmp", "tcp", "udp":
```
mattnibs: drop this?

pcap/pcap.go
```diff
@@ -0,0 +1,234 @@
...
+//  - https://github.com/wireshark/wireshark/blob/bfd51199e707c1d5c28732be34b44a9ee8a91cd8/wiretap/pcap-common.c#L723-L742
+//    - https://github.com/wireshark/wireshark/blob/f07fb6cdfc0904905627707b88450054e921f092/wiretap/libpcap.c#L592-L598
+//    - https://github.com/wireshark/wireshark/blob/f07fb6cdfc0904905627707b88450054e921f092/wiretap/libpcap.c#L714-L727
+//  - https://github.com/the-tcpdump-group/tcpdump/commit/d033c1bc381c76d13e4aface97a4f4ec8c3beca2
+//  - https://github.com/the-tcpdump-group/tcpdump/blob/88e87cb2cb74c5f939792171379acd9e0efd8b9a/netdissect.h#L263-L290
+func (r *Reader) SetSnaplen(newSnaplen uint32) {
+	r.snaplen = newSnaplen
+}
+
+// Reader formater
```
mattnibs:
```suggestion
// Reader formatter
```

pcap/pcap.go
```diff
@@ -0,0 +1,234 @@
+// Copyright 2014 Damjan Cvetko. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file in the same directory
+// containing this source file.
```
mattnibs: Is this supposed to be here?

cmd/pcap/index/command.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/mccanne/charm"
+)
+
+var Index = &charm.Spec{
+	Name:  "index",
+	Usage: "index [options]",
+	Short: "creates time index files for pcaps for use by pcap slice",
+	Long: `
+The index sub-command creates a time index for a pcap file.  The pcap file is not
+modified or copy.
```
mattnibs:
```suggestion
modified or copied.
```

pcap/pcap.go
mccanne: Ok, that typo is in the copied-in code.  Will fix.

pcap/search.go
mccanne: yes

pcap/search.go
mccanne: We can fix this in a subsequent PR.  I was using the API that your past pcap storage code implemented.

pcap/pcap.go
mccanne: In a subsequent PR.  When I had the ng-reader I will probably introduced an interface and clean this up.

pcap/slicer.go
mccanne: ok

pcap/search.go
mccanne: It's a reminder that we need to add support for icmp and other filter schemes.  I updated the comment.

pcap/pcap.go
mccanne: ok, fixed Cvetko's typo

pcap/pcap.go
mccanne: Yes, we should leave people's copyrights in place... I updated the comment.  In the old days, we put copyrights at the top of every source file.  Now, the convention is the LICENSE file in the root of the repo.  But we should leave this copyright here since I copied the code and conform with the LICENSE, which we do.

PR #349, opened 2020-02-24
Fix combine for multiple bzng inputs

mattnibs:
Have bzng readers keep their own internal context and use a
resolve.Mapper to map return Records to the shared context.

Also:
- Augment package `ztest` to support array arguments for
the input field.

### DIFF
--- a/zio/bzngio/reader.go
+++ b/zio/bzngio/reader.go
@@ -20,25 +20,36 @@ const (
 type Reader struct {
 	peeker *peeker.Reader
 	zctx   *resolver.Context
+	mapper *resolver.Mapper
 }
 
 func NewReader(reader io.Reader, zctx *resolver.Context) *Reader {
 	return &Reader{
 		peeker: peeker.NewReader(reader, ReadSize, MaxSize),
-		zctx:   zctx,
+		zctx:   resolver.NewContext(),
+		mapper: resolver.NewMapper(zctx),
 	}
 }
 
 func (r *Reader) Read() (*zng.Record, error) {
 	for {
-		r, b, err := r.ReadPayload()
+		rec, b, err := r.ReadPayload()
 		if b != nil {
 			if err != nil {
 				return nil, err
 			}
 			continue
 		}
-		return r, err
+		if rec == nil {
+			return nil, err
+		}
+		id := rec.Type.ID()
+		sharedType := r.mapper.Map(id)
+		if sharedType == nil {
+			sharedType = r.mapper.Enter(id, rec.Type)
+		}
+		rec.Type = sharedType
+		return rec, err
 	}
 }
 
--- a/zio/bzngio/types.go
+++ b/zio/bzngio/types.go
@@ -96,6 +96,7 @@ func (t *TypeFile) Dirty() bool {
 
 func ReadTypeContext(r io.Reader, zctx *resolver.Context) error {
 	reader := NewReader(r, zctx)
+	reader.zctx = zctx
 	for {
 		rec, err := reader.Read()
 		if err != nil {
--- a/ztest/ztest.go
+++ b/ztest/ztest.go
@@ -70,6 +70,7 @@ import (
 	"github.com/brimsec/zq/driver"
 	"github.com/brimsec/zq/emitter"
 	"github.com/brimsec/zq/scanner"
+	"github.com/brimsec/zq/zbuf"
 	"github.com/brimsec/zq/zio/detector"
 	"github.com/brimsec/zq/zng/resolver"
 	"github.com/brimsec/zq/zql"
@@ -83,7 +84,7 @@ import (
 func Run(t *testing.T, dirname string) {
 	zq := os.Getenv("ZTEST_ZQ")
 	if zq != "" {
-		if out, err := run(zq, "*", "", "null"); err != nil {
+		if out, err := run(zq, "*", ""); err != nil {
 			if out != "" {
 				out = fmt.Sprintf(" with output %q", out)
 			}
@@ -111,7 +112,7 @@ func Run(t *testing.T, dirname string) {
 			if err != nil {
 				t.Fatalf("%s: %s", filename, err)
 			}
-			out, err := run(zq, zt.ZQL, zt.Input, zt.OutputFormat)
+			out, err := run(zq, zt.ZQL, zt.OutputFormat, zt.Input...)
 			if err != nil {
 				if out != "" {
 					out = "\noutput:\n" + out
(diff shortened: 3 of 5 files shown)


mccanne:
Looks good.

ztest/ztest.go
```diff
@@ -135,11 +136,28 @@ func Run(t *testing.T, dirname string) {
...
 	ZQL          string `yaml:"zql"`
-	Input        string `yaml:"input"`
+	Input        Inputs `yaml:"input"`
 	OutputFormat string `yaml:"output-format,omitempty"`
 	Output       string `yaml:"output"`
 }
 
+type Inputs []string
+
+func (i *Inputs) UnmarshalYAML(value *yaml.Node) error {
```
mikesbrown: Nit: Here and elsewhere, please add doc strings to the methods and functions you introduce. Since the rest of the file already has them, let's keep the convention going. Refer to https://golang.org/doc/effective_go.html#commentary as a reminder.

zio/bzngio/reader.go
```diff
@@ -20,25 +20,36 @@ const (
...
 	peeker *peeker.Reader
 	zctx   *resolver.Context
+	mapper *resolver.Mapper
 }
 
 func NewReader(reader io.Reader, zctx *resolver.Context) *Reader {
 	return &Reader{
 		peeker: peeker.NewReader(reader, ReadSize, MaxSize),
-		zctx:   zctx,
+		zctx:   resolver.NewContext(),
```
mikesbrown: As I am less familiar with this code, I black box-tested these changes with using multiple bzng files. Unfortunately, I backed this patch out without backing out the ztest you added. I ran the following commands, and the test failed.

```
$ go test -count=1 -short ./...
$ ZTEST_ZQ=$(pwd)/dist/zq go test -v -count=1 -tags=system ./tests -args PATH=$(pwd)/dist
```

This makes me nervous that the test isn't going to catch a regression in this area again.

ztest/ztest.go
```diff
@@ -135,11 +136,28 @@ func Run(t *testing.T, dirname string) {
...
+		err := value.Decode(&inputs)
+		*i = inputs
+		return err
+	}
+	var input string
+	if err := value.Decode(&input); err != nil {
+		return err
+	}
+	*i = append(*i, input)
+	return nil
```
mikesbrown: If `value.Kind` is a `yaml.SequenceNode`, the code sets `i` regardless of the value of `err` from `value.Decode()`. But if `value.Kind` is something else, `i` is only modified if `err` is `nil`. Any particular reason for this difference?

ztest/ztest.go
```diff
@@ -167,10 +185,15 @@ func FromYAMLFile(filename string) (*ZTest, error) {
 // auto" and maybe be gzip-compressed.  outputFormat may be any string accepted
 // by "zq -f".  If zq is empty, the query runs in the current process.  If zq is
 // not empty, it specifies a zq executable that will be used to run the query.
-func run(zq, ZQL, input, outputFormat string) (string, error) {
+func run(zq, ZQL, outputFormat string, inputs ...string) (string, error) {
```
mikesbrown: Please adjust the docstring to account for `input` -> `inputs` transition.

zio/bzngio/reader.go
mattnibs: So you backed out all the bzng changes except for my `mulitple` ztest and the ztest changes that allow one to specify multiple inputs in a test?

ztest/ztest.go
mattnibs: I would imagine that `value.Decode` would throw an error if the underlying yaml isn't compatible with the provided interface (i.e. "string").

ztest/ztest.go
mattnibs: I'm going to try to make this private- it's not really something an end user would/should care about.

zio/bzngio/reader.go
mikesbrown:
```bash
$ git reset --hard 8e4faf0 && git diff a83ca90..HEAD -- scanner/combiner.go | git apply -R
$ make build
$ go test -count=1 -short ./...
$ ZTEST_ZQ=$(pwd)/dist/zq go test -v -count=1 -tags=system ./tests -args PATH=$(pwd)/dist
```

mikesbrown: If the procedure for fixing a regression is:
1. Write a failing test
2. Apply a fix to make the test pass

Then I'd expect the above to result in a failed test. They don't; the tests pass. This makes me suspicious that the test is going to catch future regressions. After all, if the test passes when the bug is present, how is it going to fail if the bug is back?

mattnibs: Did you back out the changes to bzngio/reader.go as well?

mattnibs: Sure, but your change script doesn't back out the changes to `bzngio.Reader`- where the real change happens. The changes to `scanner/combiner.go` only get rid of a `resolver.Mapper` that is now functionally a nop. I ran the inputs in the test against zq master and produced the same bug that was originally reported.

ztest/ztest.go
mattnibs: ok

zio/bzngio/reader.go
mikesbrown: 🤦‍♂ So I didn't even *notice* this was taking place in `zio/bzingio/reader.go`. I thought all this was in `scanner/combiner.go`.

mikesbrown: Yep, withdrawn on all this. My apologies. Backing out of *both* those files and rerunning the test causes the test to fail. So sorry for the mistake.

zio/bzngio/types.go
```diff
@@ -96,6 +96,7 @@ func (t *TypeFile) Dirty() bool {
 
 func ReadTypeContext(r io.Reader, zctx *resolver.Context) error {
 	reader := NewReader(r, zctx)
+	reader.zctx = zctx
```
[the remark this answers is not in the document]
mattnibs: No, this is kind of hacky because the passed-in context gets binding to a `resolver.Mapper`. This would work but if the passed in reader doesn't have any records, the passed in Context doesn't get hydrated with the mapped types.

ztest/ztest.go
```diff
@@ -135,11 +136,30 @@ func Run(t *testing.T, dirname string) {
...
+		err := value.Decode(&inputs)
+		*i = inputs
+		return err
+	}
+	var input string
+	if err := value.Decode(&input); err != nil {
+		return err
+	}
+	*i = append(*i, input)
+	return nil
```
nwt: Either approach to error handling here is fine, but let's pick one for consistency.

mattnibs: If I followed the same approach below as I do on top, an error would decode to an array with a single empty string value. I don't it should do that.

nwt: 👍

PR #352, opened 2020-02-26
Perf updates & current run

philrz:
There was a bug recently that caused the BZNG-centric parts of the perf script to fail, but now that https://github.com/brimsec/zq/pull/349 is merged, the script can run again, so I figured I'd check in with some up-to-date numbers.

While I was at it, I found that my other recent changes to the script to open the GZIP-compressed sample data logs hadn't covered `zeek-cut` and `jq` so I fixed those in the process.

Given our other deadlines, I'm not advocating we start obsessing over perf deltas at this time. For the record, though, I will point out that there's an apparent significant perf drop in handling BZNG format in general and BZNG output in particular. I've not yet done any kind of binary search to see what commits may have brought it on. For instance, the numbers show that doing `zq -f bzng *` went from <2 seconds to 18+ seconds. Granted, I know we've had to do a lot lately to fix bugs and/or support more sophisticated parts of the spec, so I don't know what's a reasonable resource budget to achieve this. But if this surprises folks, I could open a to-do item for us to circle back and revisit this perf.

### DIFF
--- a/performance/README.md
+++ b/performance/README.md
@@ -19,111 +19,111 @@ As there are many results to sift through, here's a few key summary take-aways:
 
 |**<br>Tool**|**<br>Arguments**|**Input<br>Format**|**Output<br>Format**|**<br>Real**|**<br>User**|**<br>Sys**|
 |:----------:|:---------------:|:-----------------:|:------------------:|-----------:|-----------:|----------:|
-|`zq`|`*`|zeek|zeek|18.25|35.86|1.72|
-|`zq`|`*`|zeek|bzng|5.08|6.85|0.66|
-|`zq`|`*`|zeek|zng|12.73|23.44|1.02|
-|`zq`|`*`|zeek|ndjson|64.13|88.97|4.47|
-|`zq`|`*`|bzng|zeek|16.05|21.72|0.70|
-|`zq`|`*`|bzng|bzng|1.67|2.30|0.28|
-|`zq`|`*`|bzng|zng|11.45|15.10|0.56|
-|`zq`|`*`|bzng|ndjson|58.98|64.94|1.08|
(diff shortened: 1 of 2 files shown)


philrz:
FYI, I've done the binary search to find a spot with a significant performance drop. Things were faster at `zq` commit `3d4a9ff`, where turning all the Zeek TSV sample data logs into BZNG was taking ~4 seconds.

```
~/work/zq-sample-data/zeek-default$ zq -version
Version: v0.0.1-296-g3d4a9ff

~/work/zq-sample-data/zeek-default$ time zq -f bzng * > /dev/null
real	0m4.019s
user	0m4.824s
sys	0m0.346s
```

Then @aswan's https://github.com/brimsec/zq/pull/314 briefly broke things, as it was not even possible to do a `make install` of the commit `aebabc3` associated with that PR. However, when @aswan got things working again shortly after with his https://github.com/brimsec/zq/pull/316, things were now much slower:

```
~/work/zq-sample-data/zeek-default$ zq -version
Version: v0.1.0

~/work/zq-sample-data/zeek-default$ time zq -f bzng * > /dev/null

real	0m16.868s
user	0m32.511s
sys	0m1.803s
```

Once again, I'm not banging the table saying we need to chase this down right now. First and most importantly I just wanted a record of where it dropped. Beyond that, if @aswan or anyone else wants to give a quick look at what changed and confirm whether the enhancement justifies the resource use that led to the perf drop, we could always make yet another change and re-run the perf scripts before I commit this PR.

philrz:
Now that https://github.com/brimsec/zq/pull/377 is merged, I've re-run the perf comparison script and the results show the BZNG results are closer to when we were at our best.

Separately, I notice that the permutations that read NDJSON are slower than when we were at our best, but I'm less concerned about this right now. We can always circle back and optimize that more later.

PR #353, opened 2020-02-26
Add simple tests for zqd handlers

mattnibs:
Add rudimentary testing for the space info and search handlers,
using package httptest.

Formerly endpoint specific logic was spread across several
different packages. The downside of this approach was that it
required all these packages to be aware of http protocol related
stuff. I reformatted zqd so all http-protocol related parsing
happens in the zqd package.

This pr also introduces gorilla to take of path routing.


cmd/zqd/listen/command.go
```diff
@@ -37,8 +38,12 @@ func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
 }
 
 func (c *Command) Run(args []string) error {
+	cwd, err := os.Getwd()
```
alfred-landrum: We'll want a cli flag to set this, in a later PR.

zqd/handlers.go
```diff
@@ -0,0 +1,217 @@
...
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	spaces := []string{}
+	for _, subdir := range info {
+		if !subdir.IsDir() {
+			continue
+		}
+		s, err := space.Open(root, subdir.Name())
+		if err != nil || s.HasFile("all.bzng") {
```
alfred-landrum: Should this be `!s.HasFile`?

zqd/handlers_test.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/brimsec/zq/zng/resolver"
+	"github.com/brimsec/zq/zqd"
+	"github.com/brimsec/zq/zqd/api"
+	"github.com/brimsec/zq/zqd/space"
+	"github.com/brimsec/zq/zql"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSimpleSearch(t *testing.T) {
+	src := `#0:record[_path:string,ts:time,uid:bstring]
```
alfred-landrum: Would you please start the characters for this raw string on the next line? It's a little easier to read.

zqd/handlers_test.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/brimsec/zq/zio/bzngio"
+	"github.com/brimsec/zq/zio/zngio"
+	"github.com/brimsec/zq/zng/resolver"
+	"github.com/brimsec/zq/zqd"
+	"github.com/brimsec/zq/zqd/api"
+	"github.com/brimsec/zq/zqd/space"
+	"github.com/brimsec/zq/zql"
+	"github.com/stretchr/testify/require"
+)
+
```
alfred-landrum: Would you please add a test for the space listing api?

zqd/handlers_test.go
mattnibs: sure!

zqd/handlers_test.go
mattnibs: Sure!

cmd/zqd/listen/command.go
mattnibs: Yep, was trying to keep to the scope creep down on a pr where it admittedly is a little out of hand 😄 .

mattnibs: I'll add a comment though.

zqd/handlers.go
mattnibs: Whoops. I suppose if I had a test of space list this would have been caught...

mattnibs: Actually I'm going to get rid of the check for `all.bzng`. As space still exists if it can open (i.e. has a valid `config.json`) but has no data.

PR #355, opened 2020-02-26
remove trailing slashes from routes, and 404 them if called

alfred-landrum:


alfred-landrum:
I verified this worked with the current head of brimsec/brim; I was able to see a list of spaces & search in a space with this change in place.

PR #356, opened 2020-02-27
Fix pcap slicer

mccanne:
The logic to read slices from an underlying pcap file was wrong as
io.SectionReader didn't work as I expected.  This commit fixes
the Slicer to simply use seek on the underlying file to implement
the Read interface for the sections desired.

### DIFF
--- a/pcap/slicer.go
+++ b/pcap/slicer.go
@@ -16,8 +16,9 @@ import (
 // modifying or copying the file.
 type Slicer struct {
 	slices []Slice
+	slice  Slice
 	file   *os.File
-	reader *io.SectionReader
+	eof    bool
 }
 
 func NewSlicer(file *os.File, index *Index, span nano.Span) (*Slicer, error) {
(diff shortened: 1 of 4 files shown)


pcap/slicer.go
```diff
@@ -16,43 +16,53 @@ import (
...
-			return 0, err
+	if s.eof {
+		return 0, io.EOF
+	}
+	p := b
+	if uint64(len(p)) > s.slice.Length {
+		p = p[:s.slice.Length]
+	}
+	n, err := s.file.Read(p)
+	if n != 0 {
```
mattnibs: Per the godocs, file.Read will always return 0, io.EOF. Unless you want to return `io.EOF` you should check for it before this if statement.

mccanne: I think the logic is correct here.  From go docs on io.Reader
```
When Read encounters an error or end-of-file condition after successfully reading n > 0 bytes, it returns the number of bytes read. It may return the (non-nil) error from the same call or return the error (and n == 0) from a subsequent call. 
```

PR #359, opened 2020-02-27
Add space POST endpoint

mattnibs:
Add the ability to create a new empty space via the space post
endpoint. The space post endpoint requires either a space name or
a space data dir (additionally both can be provided).

If a space data dir is specified, with no space name, a new space
is created given the name of data dir's base directory. If the name
collides with an exisiting space, a unique name is generated and
returned in the request response.

### DIFF
--- a/pkg/fs/unique.go
+++ b/pkg/fs/unique.go
@@ -1,4 +1,4 @@
-package file
+package fs
 
 import (
 	"fmt"
--- a/pkg/fs/unique.go
+++ b/pkg/fs/unique.go
@@ -2,32 +2,33 @@ package fs
 
 import (
 	"fmt"
+	"os"
 	"path/filepath"
-	"sort"
 	"strings"
 )
 
-// UniquePath determines a unique path given the desired name and parent
-// directory. For instance if the desired path is /usr/mypath but such a path
-// already exists, the path /usr/mypath_01 is returned (and so on).
-// File extensions are respected and the unique number is appended before any
-// extensions (e.g. mypath.txt -> mypath_01.txt).
-func UniquePath(parent, name string) (string, error) {
-	ext := filepath.Ext(name)
+// UniqueDir creates a unique dir given the desired name and parent directory.
+// For instance if the desired path is /usr/mydir but such a path already
+// exists, the dir /usr/mydir_01 is returned (and so on).  Path extensions are
+// respected and the unique number is appended before any extensions
+// (e.g. mydir.brim -> mydir_01.txt).
+func UniqueDir(parent, name string) (string, error) {
+	var ext string
+	if n := strings.Index(name, "."); n != -1 {
+		ext = name[n:]
+	}
 	base := strings.TrimSuffix(name, ext)
-	matches, err := filepath.Glob(filepath.Join(parent, base+"*"+ext))
-	if err != nil {
-		return "", err
-	}
-	if len(matches) > 0 {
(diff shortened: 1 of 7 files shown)


pkg/fs/unique.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"fmt"
+	"path/filepath"
+	"sort"
+	"strings"
+)
+
+// UniquePath determines a unique path given the desired name and parent
+// directory. For instanace of the desired path in /usr/mypath but such a path
```
nwt:
```suggestion
// directory. For instance if the desired path is /usr/mypath but such a path
```

pkg/fs/unique.go
```diff
@@ -0,0 +1,33 @@
...
+	matches, err := filepath.Glob(filepath.Join(parent, base+"*"+ext))
+	if err != nil {
+		return "", err
+	}
+	if len(matches) > 0 {
+		slice := sort.StringSlice(matches)
+		slice.Sort()
+		for i := 1; true; i++ {
+			name = fmt.Sprintf("%s_%02d%s", base, i, ext)
+			if n := slice.Search(base); n == slice.Len() {
```
nwt:
```suggestion
			if slice.Search(base) == slice.Len() {
```

pkg/fs/unique.go
```diff
@@ -0,0 +1,33 @@
...
+func UniquePath(parent, name string) (string, error) {
+	ext := filepath.Ext(name)
+	base := strings.TrimSuffix(name, ext)
+	matches, err := filepath.Glob(filepath.Join(parent, base+"*"+ext))
+	if err != nil {
+		return "", err
+	}
+	if len(matches) > 0 {
+		slice := sort.StringSlice(matches)
+		slice.Sort()
```
nwt:
```suggestion
		sort.Strings(matches)
```

zqd/api/api.go
```diff
@@ -80,6 +80,16 @@ type StatusResponse struct {
...
+type SpacePostRequest struct {
+	Name    string `json:"name"`
+	DataDir string `json:"data_dir"`
+}
+
+type SpacePostResponse struct {
+	Name    string `json:"name"`
+	DataDir string `json:"data_dir"`
+}
+
```
nwt:
```suggestion
type SpacePostResponse SpacePostRequest
```

pkg/fs/unique.go
```diff
@@ -0,0 +1,33 @@
...
+
+// UniquePath determines a unique path given the desired name and parent
+// directory. For instanace of the desired path in /usr/mypath but such a path
+// already exists, the path /usr/mypath_01 is returned (and so on).
+// File extensions are respected and the unique number is appended before any
+// extensions (e.g. mypath.txt -> mypath_01.txt).
+func UniquePath(parent, name string) (string, error) {
+	ext := filepath.Ext(name)
+	base := strings.TrimSuffix(name, ext)
+	matches, err := filepath.Glob(filepath.Join(parent, base+"*"+ext))
```
nwt: I don't think this does the right thing if `name` doesn't exist but `base+"x"+ext` does. I suggest doing something similar to what ioutil.TempDir does.

mattnibs: ok

mattnibs:
> It would be nice if the file path were optional and set to the path of $root/name if absent.

@mccanne this is what it does if `data_dir` is not specified in the post request.

PR #365, opened 2020-03-01
add support for ICMP pcap search

mccanne:
This commit adds support for searching for ICMP flows in pcaps.
It also adds hooks to "pcap slice" and the zqd packet end point
for filtering on icmp.

The commit also cleans up the pcap filtering API and internals by
creating more modular closures for packet filtering  At some
point, we may want to create a Go DSL for creating BPF programs
and port the BPF virtual machine and data flow optimizer to Go.
This would allow for much more flexible filtering, but is not
a priority right now.

### DIFF
--- a/pcap/search.go
+++ b/pcap/search.go
@@ -16,50 +16,39 @@ var (
 	ErrNoPacketsFound = errors.New("no packets found")
 )
 
-//XXX bool true = TCP, else UDP, change this
-type Predicate func(bool, Socket, Socket) bool
+type PacketFilter func(gopacket.Packet) bool
 
 // Search describes the parameters for a packet search over a pcap file.
 type Search struct {
-	span    nano.Span
-	matcher Predicate
-	id      string
+	span   nano.Span
+	filter PacketFilter
+	id     string
 }
 
-// NewSearch creates a new search object.
-func NewSearch(span nano.Span, proto string, srcHost string, srcPort *uint16, dstHost string, dstPort *uint16) (*Search, error) {
-	switch proto {
-	// XXX TBD add support for icmp and other packet filters
-	//case "icmp", "tcp", "udp":
-	case "tcp", "udp":
-	default:
-		return nil, fmt.Errorf("unsupported proto type: %s", proto)
-	}
-	// convert ips
-	src := net.ParseIP(srcHost)
-	if src == nil {
-		return nil, fmt.Errorf("invalid ip: %s", srcHost)
-	}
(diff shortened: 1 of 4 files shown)


philrz:
I tested out the Brim app at commit `be53d3b` with this branch at commit `0e39433` and it worked well. I was able to open all the ICMP flows I tried in Wireshark, some of which had null `duration` and some >0. 👍

pcap/search.go
```diff
@@ -75,21 +64,74 @@ func (s Search) Span() nano.Span {
...
+
+func genFlowFilter(flow Flow) func(Socket, Socket) bool {
+	return func(s0, s1 Socket) bool {
 		return s0.IP.Equal(flow.S0.IP) && s1.IP.Equal(flow.S1.IP) && s0.Port == flow.S0.Port && s1.Port == flow.S1.Port
 	}
-	return func(tcp bool, src, dst Socket) bool {
-		if tcp != tcp_ {
+}
+
+func genTCPFilter(flow Flow) func(gopacket.Packet) bool {
```
alfred-landrum: nit: these gen*Filter functions could have return arguments of `PacketFilter`

pcap/search.go
```diff
@@ -75,21 +64,74 @@ func (s Search) Span() nano.Span {
...
+	match := genFlowFilter(flow)
+	return func(packet gopacket.Packet) bool {
+		srcIP, dstIP := matchIP(packet)
+		if srcIP == nil {
```
alfred-landrum: I spent a few seconds wondering why there's no `dstIP == nil` check here. Would you consider making matchIP return arguments look like `(src net.IP, dst net.IP, ok bool)`?

pcap/search.go
```diff
@@ -75,21 +64,74 @@ func (s Search) Span() nano.Span {
...
-	return s.id
+func genUDPFilter(flow Flow) func(gopacket.Packet) bool {
+	match := genFlowFilter(flow)
+	return func(packet gopacket.Packet) bool {
+		srcIP, dstIP := matchIP(packet)
+		if srcIP == nil {
+			return false
+		}
+		transport := packet.TransportLayer()
+		tcp, ok := transport.(*layers.UDP)
```
alfred-landrum: Consider using `udp` for the udp layer var instead of tcp.

pcap/search.go
mccanne: ok

pcap/search.go
mccanne: ha!  of course

PR #368, opened 2020-03-02
add github actions ci

alfred-landrum:
Github Actions equivalent of the CircleCI tests. Example run here:
https://github.com/brimsec/zq/actions/runs/47950707


alfred-landrum:
I'll have a follow-up PR to remove the CircleCI config after we've seen this run a few times.

PR #375, opened 2020-03-03
Use zql source comparator in comparison error messages

henridf:
Previously we were using the AST encoding (e.g. 'gt' for '>'), which was a little obscure.


philrz:
Seems good!

PR #376, opened 2020-03-03
remove circleci

alfred-landrum:
Remove CircleCI since the Github Actions CI has been running ok for the last
day.


henridf:
does some config need to be tweaked in circleCI or this repo to avoid the build error?

alfred-landrum:
@henridf : yes, I've turned off the circleci zq builds now.

PR #377, opened 2020-03-03
Fix ugly bug in bzng encoder

aswan:
There was a logic error that was causing us to not call Encoder.enter()
on typedefs.  This was only noticed indirectly via the perf regression
reported in #352.

### DIFF
--- a/zng/resolver/encoder.go
+++ b/zng/resolver/encoder.go
@@ -50,7 +50,7 @@ func (e *Encoder) isEncoded(id int) bool {
 // the new type into the buffer provided.
 func (e *Encoder) Encode(dst []byte, external zng.Type) ([]byte, zng.Type, error) {
 	dst, typ, err := e.encodeType(dst, external)
-	if err != nil {
+	if err == nil {
 		e.enter(external.ID(), typ)
 	}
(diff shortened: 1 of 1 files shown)


zng/resolver/encoder.go
```diff
@@ -50,7 +50,7 @@ func (e *Encoder) isEncoded(id int) bool {
 // the new type into the buffer provided.
 func (e *Encoder) Encode(dst []byte, external zng.Type) ([]byte, zng.Type, error) {
 	dst, typ, err := e.encodeType(dst, external)
-	if err != nil {
+	if err == nil {
```
alfred-landrum: Would you please flip the logic around so this looks like the usual `if err != nil {return err}`?

aswan: That would like like this which seems unnecessarily redundant:
```
if err != nil {
  return dst, typ, err
}
e.enter(...)
return dst, typ, err
```

alfred-landrum: The error return can supply nils:

```
if err != nil {
  return nil, nil, err
}
```

Which is conventional, and will work for the only caller of Encode (in bzngio.Writer.Write).

PR #378, opened 2020-03-03
Packet Post Endpoint: Stream status updates

mattnibs:
Posting large pcap files to a space can often take a long time as zeek slowly chugs away. When a pcap is posted to a space via the `POST /space/:space/packet` endpoint api now provides streaming updates in order to provide clients with an idea of how far along the ingest process is:
- When a the request is initiated, a zeek subprocess is spawned, and the contents of the posted pcap is piped sequentially into zeek stdin and a pcap index writer. Assuming this all starts with no errors, a 200 ok status messaged is returned once the first zeek log files in temp zeek log file directory have been written to disk. A `TaskStart` response is transmitted over the response stream. What data that has been written is also transformed to sorted bzng. At this point the space is queryable.
- On a hardwired two second interval, the server streams back piped json status update payloads that contain: The start timestamp of ingest, the update timestamp, the total size of pcap file and the size of bytes read from the pcap file. From this information a client should be able to approximate the percentage completion of the ingest process (this won't be entirely accurate, however, it doesn't account for the time it will take transform the zeek logs in to time sorted bzng at the end).
- A `TaskEnd` message is sent at the completion of ingest. If `TaskEnd.Error` is not null users will know that an error occurred during ingest. If an error occurs during ingest, the process is aborted and the space is reset.

Also:
- Add api.Stream/JSONPipe functionality to make it easier to read/write piped json files over the wire.


zqd/handlers.go
```diff
@@ -247,7 +247,7 @@ func handleSpacePost(root string, w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		status := http.StatusInternalServerError
 		if err == space.ErrSpaceExists {
-			status = http.StatusBadRequest
+			status = http.StatusConflict
```
mattnibs: I hope you will allow me this one out-of-scope change.

zqd/handlers.go
```diff
@@ -272,87 +272,53 @@ func handlePacketPost(root string, w http.ResponseWriter, r *http.Request) {
...
 		return
 	}
-	idx, err := pcap.CreateIndex(pcapfile, 10000)
-	if err != nil {
-		http.Error(w, err.Error(), http.StatusInternalServerError)
-		return
+	ticker := time.NewTicker(time.Second * 2)
+	var done bool
+	var status api.PacketPostStatus
+	for {
```
alfred-landrum: I think the overall channel select for the api call should be here instead of inside `IngestProcess`. The select would check ctx.Done(), a done channel provided by IngestProcess, and the ticker. When the ticker fires, it would call into IngestProcess to get the current read size, and create a status message to send.

That would separate any rest-api specific concerns in here in the zqd handler, and nothing rest-api specific in IngestHandler.

mattnibs: Not a bad idea 👍

mattnibs: Though you don't need to check ctx.Done(): If `ctx` is cancelled this will trigger IngestProcess (or at least the resource intensive part) to exit.

mccanne:
This looks good!

mattnibs:
> This looks good!

@mccanne is that an approval I hear?

zqd/api/jsonpipescanner.go
```diff
@@ -0,0 +1,33 @@
+package api
```
alfred-landrum: Why is this in package api, when it's only used during package zqd testing?

zqd/api/stream.go
```diff
@@ -0,0 +1,47 @@
+package api
```
alfred-landrum: similar to above, why have this code here if its used only in zqd test?

alfred-landrum:
I’ve asked nibs to verify this works with existing Brim before merging

zqd/packet/packet.go
```diff
@@ -0,0 +1,236 @@
...
+		return nil
+	}
+	if err := bzngfile.Close(); err != nil {
+		return err
+	}
+	return os.Rename(bzngfile.Name(), p.space.DataPath("all.bzng"))
+}
+
+func (p *IngestProcess) startZeek(ctx context.Context, dir string) (*exec.Cmd, io.WriteCloser, error) {
+	disable := `event zeek_init() { Log::disable_stream(PacketFilter::LOG); Log::disable_stream(LoadedScripts::LOG); }`
```
nwt:
```suggestion
	const disable = `event zeek_init() { Log::disable_stream(PacketFilter::LOG); Log::disable_stream(LoadedScripts::LOG); }`
```

jameskerr:
This looks great to me. It can be merged now without breaking the app, since the app just waits for the request to finish before showing the table. I'll follow soon with the UI bits.

mattnibs:
> This looks great to me. It can be merged now without breaking the app, since the app just waits for the request to finish before showing the table. I'll follow soon with the UI bits.

Thanks @jameskerr, much appreciated.

zqd/api/jsonpipescanner.go
mattnibs: When I add back the `zqdcli` pr that I was working on, the client functionality will live in `api` and use this. This is the humble beginnings of the client api.

henridf:
@mattnibs I'm late to this PR, but I see the ingest zql has changed  from 

`sort -limit 1000000000 -r ts`
to 
`sort -limit 10000000 ts`

is that change intentional? (just checking as it seems unrelated to the PR, and isn't easily apparent in the diff)

zqd/packet/packet.go
```diff
@@ -0,0 +1,236 @@
...
+		return err
+	}
+	zw := bzngio.NewWriter(bzngfile)
+	const program = "sort -limit 10000000 ts"
+	if err := search.Copy(ctx, zw, zr, program); err != nil {
+		// If an error occurs here close and remove tmp bzngfile, lest we start
+		// leaking files and file descriptors.
+		bzngfile.Close()
+		os.Remove(bzngfile.Name())
+		return nil
```
henridf: @mattnibs shouldn't this return the error?

mattnibs: Whoops, yeah. There was one other error I made in this pr. I'll file a follow up error that fixes that and this.

mattnibs:
> @mattnibs I'm late to this PR, but I see the ingest zql has changed from
> 
> `sort -limit 1000000000 -r ts`
> to
> `sort -limit 10000000 ts`
> 
> is that change intentional? (just checking as it seems unrelated to the PR, and isn't easily apparent in the diff)

Another good catch. I'll fix this in my follow up pr.

henridf:
@mattnibs `1000000000` (the original limit) might be too high? at 100 bytes per record, that would be 100GB in mem.

mattnibs:
ok

PR #379, opened 2020-03-03
Add space delete endpoint.

mattnibs:
The directory created for the space in the zqd's datadir is deleted.
If the space contains a separate data dir, this will also be deleted.

### DIFF
--- a/zqd/space/space.go
+++ b/zqd/space/space.go
@@ -102,6 +102,15 @@ func (s Space) PacketPath() string {
 	return s.conf.PacketPath
 }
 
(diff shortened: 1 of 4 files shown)


zqd/space/space.go
```diff
@@ -102,6 +102,15 @@ func (s Space) PacketPath() string {
 	return s.conf.PacketPath
 }
 
+// Delete removes the space's path and data dir (should the data dir be
+// different then the space's path).
```
alfred-landrum: 👍

PR #384, opened 2020-03-04
Fix test expectation in zqd.PacketPostSuite

henridf:
The test was expecting 2 records, but there should be (and are) 4:

```
(after manually running the test pcap through zeek)
$ cat *log | zq " _path!=packet_filter and _path!=loaded_scripts | count() by _path" -
#0:record[_path:string,count:uint64]
0:[conn;1;]
0:[http;1;]
0:[stats;1;]
0:[capture_loss;1;]
```

### DIFF
--- a/zqd/handlers_zeek_test.go
+++ b/zqd/handlers_zeek_test.go
@@ -30,7 +30,7 @@ type PacketPostSuite struct {
 func (s *PacketPostSuite) TestCount() {
 	expected := `
 #0:record[count:uint64]
-0:[2;]`
+0:[4;]`
 	res := execSearch(s.T(), s.root, s.space, "count()")
 	s.Equal(test.Trim(expected), res)
 }


mattnibs:
@henridf this was failing on your machine (and on your particular zeek setup)?

henridf:
@mattnibs yes, I ran `zeek -C -r test.pcap local` which is what zeek.Logify runs. Do you get something different ?

zqd/handlers_zeek_test.go
```diff
@@ -51,7 +51,7 @@ func (s *PacketPostSuite) TestSpaceInfo() {
 	s.True(info.PacketSupport)
 }
 
-func (s *PacketPostSuite) SetupTest() {
+func (s *PacketPostSuite) SetupTestSuite() {
```
mattnibs: This didn't cause an error? I believe `TearDownTest` runs after each test and deletes the root.

henridf: actually it did! i got tripped up by the `// build zeek`tag. 
never mind, removing this commit and merging...

mattnibs: I actually don't think this is bad idea to run this once for entire test suite (with an accompanying call to SuiteTearDown), being all tests don't change the spaces underlying state... it'll run faster. That said we can also add this back later.

PR #385, opened 2020-03-04
Remove defer from proc.(*Sort).Pull loop

nwt:
Each loop iteration in proc.(*Sort).Pull defers a call to batch.Unref.
The accumulated runtime.newdefer allocations consume memory proportional
to the number of batches produced by the parent proc.  Remove the defer
to eliminate those allocations.


mattnibs:
Good catch.

PR #387, opened 2020-03-05
make pkg/test.Exec work with windows

alfred-landrum:
Remove the reference to bash and call the command directly. Make sure the path arg to exec.Command is a path, not just a command name, otherwise we'll trigger the usage of exec.LookPath on windows, which expects to use the PATH in the current process environment, and can't be overridden without an os.Setenv.


alfred-landrum:
closing; @mattnibs will start up a new PR.

PR #388, opened 2020-03-05
Cache min and max times

henridf:
Space min and max times are now computed at ingest time and written into the space in a file named `info.json`. 

An alternate approach (which I initially started on) would have been to continue computing min/max upon receiving the first `GET /space/..` for that space, and writing it at that point, but that would lead to a first-time slow response which this approach avoids. (Also, it seems like extracting min/maxes upon ingest and updating the space with them will be a better model when we start to support adding data to a space).

### DIFF
--- a/driver/driver_test.go
+++ b/driver/driver_test.go
@@ -51,7 +51,7 @@ func TestMuxDriver(t *testing.T) {
 		err = d.Run(flowgraph)
 		assert.NoError(t, err)
 		assert.Equal(t, 1, cs[0].(*counter).n)
-		assert.Equal(t, 1, cs[0].(*counter).n)
+		assert.Equal(t, 1, cs[1].(*counter).n)
 	})
 
 	t.Run("Mismatched channels and writer counts", func(t *testing.T) {
--- a/driver/driver_test.go
+++ b/driver/driver_test.go
@@ -47,7 +47,7 @@ func TestMuxDriver(t *testing.T) {
 		flowgraph, err := Compile(query, scanner.NewScanner(reader, nil))
 		assert.NoError(t, err)
 		cs := []zbuf.Writer{&counter{}, &counter{}}
-		d := NewDemuxed(cs)
+		d := New(cs...)
 		err = d.Run(flowgraph)
 		assert.NoError(t, err)
 		assert.Equal(t, 1, cs[0].(*counter).n)
@@ -58,7 +58,7 @@ func TestMuxDriver(t *testing.T) {
 		flowgraph, err := Compile(query, nil)
 		assert.NoError(t, err)
 		cs := []zbuf.Writer{&counter{}, &counter{}, &counter{}}
-		d := NewDemuxed(cs)
+		d := New(cs...)
 		err = d.Run(flowgraph)
 		assert.Error(t, err)
 	})
--- a/driver/driver_test.go
+++ b/driver/driver_test.go
@@ -54,7 +54,7 @@ func TestMuxDriver(t *testing.T) {
 		assert.Equal(t, 1, cs[1].(*counter).n)
 	})
 
-	t.Run("Mismatched channels and writer counts", func(t *testing.T) {
+	t.Run("mismatched channels and writer counts", func(t *testing.T) {
 		flowgraph, err := Compile(query, nil)
 		assert.NoError(t, err)
 		cs := []zbuf.Writer{&counter{}, &counter{}, &counter{}}
--- a/driver/driver_test.go
+++ b/driver/driver_test.go
@@ -1,0 +1,64 @@
+package driver
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/brimsec/zq/scanner"
+	"github.com/brimsec/zq/zbuf"
+	"github.com/brimsec/zq/zio/zngio"
+	"github.com/brimsec/zq/zng"
+	"github.com/brimsec/zq/zng/resolver"
+	"github.com/brimsec/zq/zql"
+	"github.com/stretchr/testify/assert"
+)
+
+type counter struct {
+	n int
+}
+
+func (c *counter) Write(*zng.Record) error {
+	c.n++
+	return nil
+}
+
+func TestMuxDriver(t *testing.T) {
+	input := `#0:record[_path:string,ts:time]
+0:[conn;1425565514.419939;]`
+
+	zctx := resolver.NewContext()
+	query, err := zql.ParseProc("(tail 1; tail 1)")
+	assert.NoError(t, err)
+
+	t.Run("muxed into one writer", func(t *testing.T) {
+		reader := zngio.NewReader(strings.NewReader(input), zctx)
+		flowgraph, err := Compile(query, scanner.NewScanner(reader, nil))
+		assert.NoError(t, err)
+		c := counter{}
+		d := New(&c)
+		err = d.Run(flowgraph)
+		assert.NoError(t, err)
+		assert.Equal(t, 2, c.n)
+	})
+
+	t.Run("muxed into individual writers", func(t *testing.T) {
+		reader := zngio.NewReader(strings.NewReader(input), zctx)
+		flowgraph, err := Compile(query, scanner.NewScanner(reader, nil))
+		assert.NoError(t, err)
+		cs := []zbuf.Writer{&counter{}, &counter{}}
+		d := NewDemuxed(cs)
+		err = d.Run(flowgraph)
+		assert.NoError(t, err)
+		assert.Equal(t, 1, cs[0].(*counter).n)
+		assert.Equal(t, 1, cs[0].(*counter).n)
+	})
+
+	t.Run("Mismatched channels and writer counts", func(t *testing.T) {
+		flowgraph, err := Compile(query, nil)
+		assert.NoError(t, err)
+		cs := []zbuf.Writer{&counter{}, &counter{}, &counter{}}
+		d := NewDemuxed(cs)
+		err = d.Run(flowgraph)
+		assert.Error(t, err)
+	})
+}
--- a/driver/driver_test.go
+++ b/driver/driver_test.go
@@ -23,7 +23,8 @@ func (c *counter) Write(*zng.Record) error {
 }
 
 func TestMuxDriver(t *testing.T) {
-	input := `#0:record[_path:string,ts:time]
+	input := `
+#0:record[_path:string,ts:time]
 0:[conn;1425565514.419939;]`
 
 	zctx := resolver.NewContext()
--- a/driver/run.go
+++ b/driver/run.go
@@ -10,13 +10,21 @@ import (
 )
 
 type Driver struct {
-	writer   zbuf.Writer
+	writers  []zbuf.Writer
 	warnings io.Writer
+	demux    bool
 }
 
 func New(w zbuf.Writer) *Driver {
 	return &Driver{
-		writer: w,
+		writers: []zbuf.Writer{w},
+	}
+}
+
+func NewDemuxed(ws []zbuf.Writer) *Driver {
+	return &Driver{
+		writers: ws,
+		demux:   true,
 	}
 }
 
@@ -25,8 +33,11 @@ func (d *Driver) SetWarningsWriter(w io.Writer) {
 }
 
 func (d *Driver) Write(cid int, arr zbuf.Batch) error {
+	if !d.demux {
+		cid = 0
+	}
 	for _, r := range arr.Records() {
-		if err := d.writer.Write(r); err != nil {
+		if err := d.writers[cid].Write(r); err != nil {
 			return err
 		}
 	}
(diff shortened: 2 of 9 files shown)


zqd/space/space.go
```diff
@@ -8,9 +8,12 @@ import (
 	"path/filepath"
 
 	"github.com/brimsec/zq/pkg/fs"
+	"github.com/brimsec/zq/pkg/nano"
+	"github.com/brimsec/zq/zqd/api"
 )
 
 const configFile = "config.json"
+const infoFile = "info.json"
```
alfred-landrum: I'm a little concerned about spreading space related information across files. What's your thoughts on using a single file (with a name other than 'config') to hold configuration & metadata?

driver/driver_test.go
```diff
@@ -0,0 +1,64 @@
...
+	n int
+}
+
+func (c *counter) Write(*zng.Record) error {
+	c.n++
+	return nil
+}
+
+func TestMuxDriver(t *testing.T) {
+	input := `#0:record[_path:string,ts:time]
```
alfred-landrum: nit: would you start the string literal on the next line so it reads a little easier down the page?:
```
  input := `
#0:record ...
0:...
```

zqd/packet/packet.go
```diff
@@ -70,6 +72,7 @@ func IngestFile(ctx context.Context, s *space.Space, pcap string) (*IngestProces
 }
 
 func (p *IngestProcess) run(ctx context.Context) error {
+	var minTs, maxTs nano.Ts
```
nwt: Move this closer to where these are used.

zqd/packet/packet.go
```diff
@@ -193,29 +208,36 @@ func (p *IngestProcess) writeData(ctx context.Context) error {
...
 	if err != nil {
-		return err
+		return nano.Ts(0), nano.Ts(0), err
 	}
 	zw := bzngio.NewWriter(bzngfile)
-	const program = "sort -limit 10000000 ts"
-	if err := search.Copy(ctx, zw, zr, program); err != nil {
+	const program = "sort -limit 10000000 ts | (filter *; head 1; tail 1)"
+	headW := recWriter{}
+	tailW := recWriter{}
```
nwt: Use var.

zqd/space/space.go
```diff
@@ -116,10 +147,28 @@ type config struct {
 	PacketPath string `json:"packet_path"`
 }
 
+type info struct {
+	MinTime nano.Ts `json:"min_time"`
+	MaxTime nano.Ts `json:"max_time"`
+}
+
+func (s Space) SetTimes(minTs, maxTs nano.Ts) error {
+	cur, err := loadInfo(s.path)
```
mattnibs: This should be in the data path, not the space path.

zqd/space/space.go
```diff
@@ -69,6 +72,34 @@ func (s Space) Name() string {
 	return filepath.Base(s.path)
 }
 
+func (s Space) Info() (api.SpaceInfo, error) {
```
mattnibs: I've been trying to be disciplined about not bringing `api` into this package. `space`, `packet`, and `search` shouldn't be aware of protocol level stuff.

zqd/space/space.go
```diff
@@ -116,10 +147,28 @@ type config struct {
...
+	cur.MaxTime = nano.Max(cur.MaxTime, maxTs)
+	return cur.save(s.path)
+}
+
 // loadConfig loads the contents of config.json in a space's path.
-func loadConfig(name string) (config, error) {
+func loadConfig(spacePath string) (config, error) {
 	var c config
-	b, err := ioutil.ReadFile(filepath.Join(name, configFile))
+	b, err := ioutil.ReadFile(filepath.Join(spacePath, configFile))
```
mattnibs: Fine, but outside the scope of this pr.

zqd/packet/packet.go
```diff
@@ -183,7 +189,16 @@ func (p *IngestProcess) Done() <-chan struct{} {
...
+type recWriter struct {
+	r *zng.Record
+}
+
+func (rw *recWriter) Write(r *zng.Record) error {
+	rw.r = r
+	return nil
+}
+
+func (p *IngestProcess) writeData(ctx context.Context) (nano.Ts, nano.Ts, error) {
```
mattnibs: My preference would be generate the min/max cache when the `/space/:space/info` endpoint is hit and no space cache exists- its really the only place that cares about this information. That said I can see an argument for your approach, my approach would make the first call to space/info slow.

zqd/packet/packet.go
```diff
@@ -193,29 +208,36 @@ func (p *IngestProcess) writeData(ctx context.Context) error {
...
 	// write bzng to a temp file and rename on successful conversion.
 	bzngfile, err := p.space.CreateFile("all.bzng.tmp")
 	if err != nil {
-		return err
+		return nano.Ts(0), nano.Ts(0), err
 	}
 	zw := bzngio.NewWriter(bzngfile)
-	const program = "sort -limit 10000000 ts"
-	if err := search.Copy(ctx, zw, zr, program); err != nil {
+	const program = "sort -limit 10000000 ts | (filter *; head 1; tail 1)"
```
mattnibs: I would suggest writing the program like this: `sort -limit 10000000 ts; min(ts) max(ts)` and have the second output to second file called `all.spancache.bzng` adjacent to`all.bzng`. We wouldn't need to have the logic in `space` to create the `info.json` file.

zqd/packet/packet.go
henridf: @nwt can't do that due to gotos.

zqd/packet/packet.go
henridf: Yes, this is discussed in the PR description.

zqd/space/space.go
henridf: Fair enough, but since I added a similar function with the more appropriate argument name ("spacePath" vs "name"), I thought it was reasonable to adjust the pre-existing function for symmetry.

That said, is this a request to remove it?

mattnibs: Nah, just pointing it out.

zqd/packet/packet.go
henridf: Vars used.

zqd/space/space.go
henridf: That's a fair point and I could go either way on that one - if you think this is preferable I can adjust the PR. That said, there's feedback below saying this stuff should go in an altogether different place, so I'll try to resolve that question offline first...

driver/driver_test.go
```diff
@@ -0,0 +1,65 @@
...
+	t.Run("muxed into individual writers", func(t *testing.T) {
+		reader := zngio.NewReader(strings.NewReader(input), zctx)
+		flowgraph, err := Compile(query, scanner.NewScanner(reader, nil))
+		assert.NoError(t, err)
+		cs := []zbuf.Writer{&counter{}, &counter{}}
+		d := NewDemuxed(cs)
+		err = d.Run(flowgraph)
+		assert.NoError(t, err)
+		assert.Equal(t, 1, cs[0].(*counter).n)
+		assert.Equal(t, 1, cs[0].(*counter).n)
```
nwt:
```suggestion
		assert.Equal(t, 1, cs[0].(*counter).n)
```

driver/driver_test.go
```diff
@@ -0,0 +1,65 @@
...
+	})
+
+	t.Run("Mismatched channels and writer counts", func(t *testing.T) {
```
nwt:
```suggestion
	t.Run("mismatched channels and writer counts", func(t *testing.T) {
```

driver/run.go
```diff
@@ -10,13 +10,21 @@ import (
...
+		writers: []zbuf.Writer{w},
+	}
+}
+
+func NewDemuxed(ws []zbuf.Writer) *Driver {
+	return &Driver{
+		writers: ws,
+		demux:   true,
 	}
 }
```
nwt: Could we simplify this to
```go
type Driver struct {
	writers  []zbuf.Writer
	warnings io.Writer
}

func New(w ...zbuf.Writer) *Driver {
	return &Driver{
		writers: w,
	}
}
```
and use `len(d.writers) > 1` in place of `d.demux`?

driver/driver_test.go
henridf: oops! Actually the second line should be `assert.Equal(t, 1, cs[1].(*counter).n)`

henridf:
```suggestion
		assert.Equal(t, 1, cs[1].(*counter).n)
```

driver/run.go
henridf: @nwt much better. Thanks!

zqd/space/space.go
henridf: 👍

zqd/space/space.go
henridf: Offline resolution was that info should move to the data dir, which would prevent colocating it with config.

zqd/space/space.go
henridf: Moved.

zqd/packet/packet.go
henridf: After offline discussion, we agreed to go with json.

zqd/space/space.go
henridf: I'm not seeing what this rule gets us given that the "protocol level stuff" in this instance is merely a struct populated with space fields and that is exclusively the space's concern.

But it's cosmetic and small, so adopted it in the interest of wrapping this PR up.

driver/run.go
```diff
@@ -25,8 +25,11 @@ func (d *Driver) SetWarningsWriter(w io.Writer) {
...
 
 func (d *Driver) Write(cid int, arr zbuf.Batch) error {
+	if len(d.writers) == 1 {
```
alfred-landrum: nit: is this check & cid enforcement necessary?

henridf: This check (and the `cid = 0` below it) serve to maintain the default pre-existing behavior, which is to multiplex all flowgraph leaves into a single output.

mccanne:
I don't understand the new complexity here.  All we need to do is keep a cache of the stats and compare the time of the cache to the modtime of the zng file.  If you want to avoid first hit is slow, then you could trigger a cache check anytime the zng file is changed.  This would also allow trivial support for incremental updates of all.bzng as zeek is crunching the pcap and generating the logs.

henridf:
@mccanne I'm not really seeing the new complexity, but yes, we could have taken the approach you suggest. 

But triggering a cache check any time the zng file is changed means re-reading the whole thing? In previous work we went to great lengths to avoid reading the same data twice, so I was probably influenced by that.

zqd/search/search.go
```diff
@@ -75,7 +75,11 @@ func Copy(ctx context.Context, w zbuf.Writer, r zbuf.Reader, prog string) error
 	if err != nil {
 		return err
 	}
-	d := zdriver.New(w)
+	if len(w) == 1 {
+		d := zdriver.New(w[0])
+		return d.Run(mux)
+	}
+	d := zdriver.New(w...)
```
mattnibs: I don't think you need the length check here. Shorten to:
```suggestion
	d := zdriver.New(w...)
```

mccanne:
> @mccanne I'm not really seeing the new complexity, but yes, we could have taken the approach you suggest.
> 
> But triggering a cache check any time the zng file is changed means re-reading the whole thing? In previous work we went to great lengths to avoid reading the same data twice, so I was probably influenced by that.

Ok, I was thrown off a bit here because I thought it was baked into the plan that progressive updates to all.bzng would occur while zeek runs.  This is now a stretch goal.  Toward this end, the logic in IngestProcess.writeData() will soon run periodically and asynchronously to zeek/slurp so there should be tight consistency between all.bzng and info.json:  I would put the creation of info.json at the same place as os.Rename() of the tmp bzng file to all.bzng rather than returning the min/max timestamps.  (This should be atomic but we can live on the edge for now given it's a desktop app for now.)

zqd/packet/packet.go
```diff
@@ -80,7 +83,10 @@ func (p *IngestProcess) run(ctx context.Context) error {
 	if err = p.space.SetPacketPath(p.pcapPath); err != nil {
 		goto abort
 	}
-	if err = p.writeData(ctx); err != nil {
+	if minTs, maxTs, err = p.writeData(ctx); err != nil {
+		goto abort
+	}
+	if err = p.space.SetTimes(minTs, maxTs); err != nil {
```
mccanne: Move this into writeData since the logic there will eventually do progressive updates to all.bnzg and we want a tight coupling between the zng file and the meta-data about the zng file.

zqd/packet/packet.go
```diff
@@ -193,29 +208,35 @@ func (p *IngestProcess) writeData(ctx context.Context) error {
...
+
+	minTs := headW.r.Ts
+	maxTs := tailW.r.Ts
+
 	if err := bzngfile.Close(); err != nil {
-		return err
+		return nano.Ts(0), nano.Ts(0), err
 	}
-	return os.Rename(bzngfile.Name(), p.space.DataPath("all.bzng"))
+	return minTs, maxTs, os.Rename(bzngfile.Name(), p.space.DataPath("all.bzng"))
```
mccanne: Update info.json here.

zqd/search/search.go
henridf: oops... leftover from 2f55965dede9c372808b3a9e7dda869741f43ff1

zqd/packet/packet.go
henridf: done

zqd/packet/packet.go
henridf: done

PR #393, opened 2020-03-05
Small AST cleanups

aswan:
Make Literal into a full fledged ast Node (which means that in json it
gains an Op property).  Also convert the node formerly known as
BooleanLiteral to MatchAll to better reflect its meaning and avoid
confusion with actual Literal nodes.

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -42,6 +42,7 @@ type FieldExpr interface {
 // the native Go types) and value is a string representation of that value that
 // must conform to the provided type.
 type Literal struct {
+	Node
 	Type  string `json:"type"`
 	Value string `json:"value"`
 }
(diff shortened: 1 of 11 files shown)


henridf:
I realize this is driven by your work on expressions, and I don't have that context, but does a literal need an "Op" property?

aswan:
> I realize this is driven by your work on expressions, and I don't have that context, but does a literal need an "Op" property?

Yeah, we'll end up with expression nodes for things like arithmetic operations that have additional expressions as children.  For those children to be able to be literals, they need to identify themselves as such!
I can move that part into a different patch if you think it doesn't belong here...

henridf:
Sounds good!  no need for a different patch.

PR #394, opened 2020-03-06
Make zeek executable path configurable

mattnibs:
Previously ingest pcap endpoint was using any available in path.
Provide added flexibility to by allowing users to specify a specific
zeek executable to run against. If no executable is provided zeek
if found use $PATH.

### DIFF
--- a/cmd/zqd/listen/command.go
+++ b/cmd/zqd/listen/command.go
@@ -47,7 +47,7 @@ func (c *Command) Run(args []string) error {
 		return err
 	}
 	logger := newLogger()
-	core := zqd.Core{Root: dataDir}
+	core := zqd.Core{Root: dataDir, ZeekExec: c.zeekExec}
 	handler := zqd.NewHandler(core)
 	ln, err := net.Listen("tcp", c.listenAddr)
 	if err != nil {
--- a/cmd/zqd/listen/command.go
+++ b/cmd/zqd/listen/command.go
@@ -47,7 +47,7 @@ func (c *Command) Run(args []string) error {
 		return err
 	}
 	logger := newLogger()
-	core := zqd.Core{Root: dataDir, ZeekExec: c.zeekExec}
+	core := &zqd.Core{Root: dataDir, ZeekExec: c.zeekExec}
 	handler := zqd.NewHandler(core)
 	ln, err := net.Listen("tcp", c.listenAddr)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


mattnibs:
This is part the work to adding more testing for the packet post endpoint. I thought it would be easier on reviewers to split this into its own commit, rather then bundle it all together.

The `Core` thing struct idea I got from Vault's http handler: https://github.com/hashicorp/vault/blob/master/http/handler.go

cmd/zqd/listen/command.go
```diff
@@ -30,12 +30,14 @@ type Command struct {
...
 	listenAddr string
 	dataDir    string
+	zeekExec   string
 }
 
 func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
 	c := &Command{Command: parent.(*root.Command)}
 	f.StringVar(&c.listenAddr, "l", ":9867", "[addr]:port to listen on")
 	f.StringVar(&c.dataDir, "datadir", ".", "data directory")
+	f.StringVar(&c.zeekExec, "zeekpath", "", "path to the zeek executeable to use (defaults to zeek in $PATH)")
```
alfred-landrum: misspelling: `executable`

cmd/zqd/listen/command.go
```diff
@@ -45,7 +47,8 @@ func (c *Command) Run(args []string) error {
 		return err
 	}
 	logger := newLogger()
-	handler := zqd.NewHandler(dataDir)
+	core := zqd.Core{Root: dataDir, ZeekExec: c.zeekExec}
+	handler := zqd.NewHandler(core)
```
alfred-landrum: I like the idea of a `Core` struct. I think it should be passed as a pointer-to-struct parameter instead of a struct parameter in the all of the places it's used.

zqd/server.go
```diff
@@ -7,6 +7,13 @@ import (
 	"github.com/gorilla/mux"
 )
 
+type Core struct {
+	Root string
+	// The exact path of the zeek executable. This is needed in the
```
alfred-landrum: Would you alter the comment to point out if ZeekExec is blank, we'll use $PATH?

cmd/zqd/listen/command.go
mattnibs: Good idea.

PR #395, opened 2020-03-06
space get endpoint: handle spaces w no data

mattnibs:
Currently querying GET /space/:space results in a 500 error:
all.bzng does not exist. Make endpoint more robust and return
a 200 with a space info response with the "size" field set to 0.

### DIFF
--- a/zqd/handlers_test.go
+++ b/zqd/handlers_test.go
@@ -86,6 +86,23 @@ func TestSpaceInfo(t *testing.T) {
 	require.Equal(t, expected, info)
 }
 
+func TestSpaceInfoNoData(t *testing.T) {
+	space := "test"
+	root := createTempDir(t)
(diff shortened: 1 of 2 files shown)


mattnibs:
I'll wait to merge this until #388 gets in... but either way the added test should exist.

zqd/handlers_test.go
```diff
@@ -86,6 +86,23 @@ func TestSpaceInfo(t *testing.T) {
 	require.Equal(t, expected, info)
 }
 
+func TestSpaceInfoNoData(t *testing.T) {
+	space := "test"
```
nwt:
```suggestion
	const space = "test"
```

henridf:
> I'll wait to merge this until #388 gets in... but either way the added test should exist.

Ugh. Would have been nice to wait, since you'd thought of it.

PR #397, opened 2020-03-06
WIP on expressions

aswan:
Very much still in progress.  The grammar is mostly complete with the
exception of function calls and field/array dereferences (which are
going to require additional AST node types).  I've also begun to adopt
the convention in many other PEG grammars of using CamelCase rule names.
This means the grammar is in an inconsistent style right now but I
propose we go with this and clean up the rest of the grammar in a
separate pass when the rest of the work on expressions is finished.

Compilation is a different story, only addition of integers works right
now.  Making this work has required a tremendous amount of repetitive
code.  I haven't come up with any good ideas to simplify this yet,
definitely looking for ideas here...


aswan:
This isn't ready to merge yet, but looking for feedback, particularly on how to handle the largish matrix of operand types when compiling expressions.
@mccanne this seems like something you'll have an opinion on :)

ast/ast.go
```diff
@@ -121,6 +125,20 @@ type (
...
+type BinaryExpression struct {
+	Node
+	Operator string     `json:"operator"`
+	LHS      Expression `json:"lhs"`
+	RHS      Expression `json:"rhs"`
+}
+
+func (*BinaryExpression) exprNode() {}
+func (*Literal) exprNode()          {}
+func (*FieldRead) exprNode()        {}
```
henridf: Shouldn't FieldCall also get an `exprNode()`?

expr/expr_test.go
```diff
@@ -0,0 +1,358 @@
...
+			exp = fmt.Sprintf("%sb < %sa", t1, t2)
+			testSuccessful(t, exp, record, zbool(false))
+
+			// a <= comparison that is true (becuase <)
+			exp = fmt.Sprintf("%sa <= %sb", t1, t2)
+			testSuccessful(t, exp, record, zbool(true))
+
+			// a <= comparison that is false (because >)
+			exp = fmt.Sprintf("%sb <= %sa", record, zbool(false))
+
```
henridf: There's no test on `exp` for this one and the one below

expr/expr.go
```diff
@@ -0,0 +1,580 @@
...
+		case "AND":
+			if !lv {
+				return lhs, nil
+			}
+		case "OR":
+			if lv {
+				return lhs, nil
+			}
+		default:
+			panic("bad operator")
```
henridf: Don't panic! 🆘

expr/expr.go
```diff
@@ -0,0 +1,580 @@
...
+			return nil, fmt.Errorf("invalid binary operator %s", n.Operator)
+		}
+
+	default:
+		return nil, fmt.Errorf("invalid expression type %T", node)
+	}
+}
+
+func compileLogical(node ast.BinaryExpression) (NativeEvaluator, error) {
+	lhsFunc, err := compileNative(node.LHS)
```
henridf: nit: I was a little surprised to see `compileNative()` inside this (and in compileRelative and compileEqual)... it could be pulled out of each of these and done in compileNative, and passed down. The signature would then be `func compileLogical(lhs, rhs NativeEvaluator, op string)`.  I think the recursion would then better reflect the AST, but this is probably just cosmetic, and ymmv, so... nit!

expr/expr.go
henridf: Oh, nvm, when I wrote this I hadn't realized that operators are checked before getting here (in `compileNative()`)... so disregard the above.

henridf:
This looks good to me! Re feedback on how to handle the matrix of operand types, I'm not coming up with any better ideas...

ast/ast.go
aswan: At least for now, FieldCall is part of the filter syntax but not part of the expression syntax.  I think that the easiest short-term thing is to keep the two separate and then, once function calls are part of the expression syntax we can look at either combining related AST nodes if that makes sense or renaming them to clarify their roles if they remain separate...

aswan:
This patch now covers logical, comparison, and arithmetic operators.
Function calls, array/field references, and the `put` proc will come in subsequent PRs but this is already gigantic.

ast/ast.go
```diff
@@ -36,6 +36,10 @@ type FieldExpr interface {
 	Copy() FieldExpr
 }
 
+type Expression interface {
+	exprNode()
```
henridf: This is missing `Copy()` ... I think we should either remove it everywhere, or add it here. 

I think it's useful to have, but maybe we should switch the implementation to use some deep copy library (or roll our own with reflect).

aswan: Since we're not actually using `Copy()` anywhere, I'm loathe to spend time on this (and associated tests).  Since we support json serialization and deserialization, how about if somebody suddenly has an urgent need to make a deep copy of an ast, they just make a round trip through json?

henridf: That seems like a good approach too. The point remains I think that we should either have `Copy()` on all nodes, or on none...

aswan: @nwt you were the other one who objected to removing `Copy()` are you okay with using json if/when we ever need to copy asts?
FWIW, the immediate problem with adding it here is that FieldRead wants to be both a FieldExpr and an Expression so those two interfaces can't both have `Copy()` methods with different signatures...

henridf: > FWIW, the immediate problem with adding it here is that FieldRead wants to be both a FieldExpr and an Expression so those two interfaces can't both have Copy() methods with different signatures...

Ugh. `Copy()` should go! :)

nwt: Don't care.

PR #398, opened 2020-03-06
create pcapio and generic Reader interface

mccanne:
This commit refactors the pcap package to get it ready to drop in
support for pcap-ng.  We moved the pcap reader implementatin to
a new pcapio package that exports a generic interface for reading
pcap-ng blocks (section headers, packet blocks, inteface blocks)
in a way that is backward compatible with legacy pcap (a single
section header representing the pcap file header followed by
packet blocks).

All of the pcap parsing code imported from pcapgo will live in
the pcapio package, named with by their original source file names.

We also simplified the Index type since the struct contained
just a slice of Sections so now pcap.Index should be passed around
instead of *pcap.Index.  This also means that the marshaling of
this data structure changes the format of the on-disk index
file so existing spaces need a new run of "pcap index".

This commit is the second step toward getting the pcap subsytem
read for pcap-ng.  The next PR in this sequence will have the
actual support for pcap-ng.


pcap/pcapio/read.go
```diff
@@ -33,7 +33,7 @@ import (
 //
 // If the PCAP data is gzip compressed it is transparently uncompressed
 // by wrapping the given io.Reader with a gzip.Reader.
-type Reader struct {
+type PcapReader struct {
```
nwt: Nit: Update the doc comment to reflect the new name.

pcap/pcapio/read.go
```diff
@@ -69,8 +69,8 @@ const magicGzip2 = 0x8b
 //  defer f.Close()
 //  r, err := NewReader(f)
 //  data, ci, err := r.ReadPacketData()
-func NewReader(r io.Reader) (*Reader, error) {
-	reader := &Reader{
+func NewPcapReader(r io.Reader) (*PcapReader, error) {
```
nwt: Nit: Update the doc comment to reflect the new name.

pcap/pcapio/read.go
```diff
@@ -119,47 +119,46 @@ func (r *Reader) readHeader() error {
...
-	if !span.ContainsClosed(ts) {
-		goto again
+		return nil, Info{}, err
 	}
-	return pkt, nil
+	r.offset += uint64(caplen + packetHeaderLen)
+	return pkt, Info{Type: TypePacket, Link: r.linkType, Ts: ts}, nil
 }
 
 // ReadPacketData reads next packet from file.
```
nwt:
```suggestion
// readPacketData reads next packet from file.
```

mccanne: Actually, this code is  unused and subsumed by Read() so I will delete it.

PR #399, opened 2020-03-06
More testing for packet post endpoint

mattnibs:
First pass of tests for the POST packet endpoint. Includes tests for the success case (including
various subtests verifying correct space state around this) and also failing cases:
- invalid pcap
- zeek failing on start
- zeek failing after start and initial write

Issues that should be addressed in future prs:
- For some reason the zeek failing after start and initial write test causes the space's config.json to get
deleted- this centainly needs to be fixed.
- Better error messages around zeek exiting with an error code (right now we just get a nebulous "exit code 1" error.
- Post invalid status code returns a 500 error when it should return status 400- this is caught by the zeek process when it should be caught by `pcap.CreateIndex`. This also needs better error messaging (for instance is this an unsupported version of pcap/ngpcap or is the file nonsense).
- Test to verify that cancelling a request terminates the quest and properly disposes of the zeek subprocess.


Makefile
```diff
@@ -24,6 +39,13 @@ $(SAMPLEDATA):
 
 sampledata: $(SAMPLEDATA)
 
+bin/$(ZEEKPATH):
+	@mkdir -p bin
+	@curl -L -o bin/$(ZEEKPATH).zip \
+		https://github.com/brimsec/zeek/releases/download/$(ZEEKTAG)/zeek-$(ZEEKTAG).$(PLATFORM)-amd64.zip
```
nwt:
```suggestion
		https://github.com/brimsec/zeek/releases/download/$(ZEEKTAG)/zeek-$(ZEEKTAG).$$(go env GOOS)-$$(go env GOARCH).zip
```

This will let you remove the PLATFORM stuff up above.

nwt:
```suggestion
		https://github.com/brimsec/zeek/releases/download/$(ZEEKTAG)/zeek-$(ZEEKTAG).$$(go env GOOS)-$$(go env GOARCH).zip
```

This will let you remove the PLATFORM stuff up above.

.gitignore
```diff
@@ -1,4 +1,5 @@
 dist/
+bin/
```
nwt:
```suggestion
bin/
dist/
```

Makefile
```diff
@@ -33,6 +55,9 @@ test-system: build
 test-heavy: build $(SAMPLEDATA)
 	@go test -v -tags=heavy ./tests
 
+test-zeek: bin/$(ZEEKPATH)
+	@ZEEKEXEC=$(CURDIR)/bin/$(ZEEKPATH)/zeek go test -v -run=Zeek -tags=zeek ./zqd
```
nwt: Nit: Call this ZEEK instead of ZEEKEXEC.

Makefile
mattnibs: Beautiful

zqd/handlers_zeek_test.go
```diff
@@ -11,88 +12,142 @@ import (
...
+	ZeekExec       string
+	Pcapfile       string
+	ExpectedStatus int
+	Test           func(packetPostResult)
 }
 
-func (s *PacketPostSuite) TestCount() {
-	expected := `
+func TestZeekPacketPost(t *testing.T) {
+	var cases = []packetPostTest{
```
alfred-landrum: I like the breakout of different tests here. I think this array form of test cases is better suited for table-driven tests ( https://github.com/golang/go/wiki/TableDrivenTests ), where the test logic is the same for each case. So, consider if these tests were structured like:
```go
t.Run("SearchCount", func(t *testing.T) {
  pt := setupPacketPostTest(t)
  defer cleanupPacketPostTest(pt)
  ...
})
```

zqd/handlers_zeek_test.go
```diff
@@ -11,88 +12,142 @@ import (
...
-	info, err := os.Stat(s.pcapfile)
-	s.NoError(err)
-	s.Len(s.payloads, 3)
-	status := s.payloads[1].(*api.PacketPostStatus)
-	s.Equal(status.Type, "PacketPostStatus")
-	s.Equal(status.PacketSize, info.Size())
-	s.Equal(status.PacketReadSize, info.Size())
+func (p packetPostTest) Run(t *testing.T) {
+	if p.ZeekExec == "" {
+		p.ZeekExec = zeekpath
```
alfred-landrum: Is the use case for this to have "zeek" binaries that return specific failures?

zqd/handlers_zeek_test.go
mattnibs: Thanks for the post to the wiki! I agree, will change.

zqd/handlers_zeek_test.go
mattnibs: Yes.

zqd/handlers_zeek_test.go
```diff
@@ -16,98 +16,179 @@ import (
...
+		u := fmt.Sprintf("http://localhost:9867/space/%s", p.space)
+		var info api.SpaceInfo
+		httpJSONSuccess(t, zqd.NewHandler(p.core), "GET", u, nil, &info)
+		assert.Equal(t, p.space, info.Name)
+		assert.Equal(t, nano.Unix(1501770877, 471635000), *info.MinTime)
+		assert.Equal(t, nano.Unix(1501770880, 988247000), *info.MaxTime)
+		assert.InDelta(t, 1437, info.Size, 2)
+		assert.True(t, info.PacketSupport)
+		assert.Equal(t, p.pcapfile, info.PacketPath)
+		// XXX Cannot test with full struct because zeek's randomly generate
```
alfred-landrum: The test verifies most of the fields in .info. Is there something specific that's not tested due to the changes between zeek runs? If so, could you slim this comment down & just state that?

zqd/handlers_zeek_test.go
```diff
@@ -16,98 +16,179 @@ import (
...
 
-func (s *PacketPostSuite) TestNoPacketFilter() {
-	expected := ""
-	res := execSearch(s.T(), s.core, s.space, "_path = packet_filter | count()")
-	s.Equal(expected, res)
+func TestPacketPostInvalidPcap(t *testing.T) {
+	p := packetPost(t, zeekpath, "./testdata/invalid.pcap", 500)
+	defer p.cleanup()
+	t.Run("ErrorMessage", func(t *testing.T) {
+		// XXX Better error message here.
```
alfred-landrum: I think there's a `require.Regexp`? So perhaps `require.Regexp(t, "Unknown magic.*")`?

zqd/handlers_zeek_test.go
```diff
@@ -16,98 +16,179 @@ import (
...
+		require.Contains(t, p.payloads, expected)
+	})
 }
 
-func (s *PacketPostSuite) TestWritesIndexFile() {
-	stat, err := os.Stat(filepath.Join(s.core.Root, s.space, packet.IndexFile))
-	s.NoError(err)
-	s.NotNil(stat)
+func TestPacketPostZeekFailAfterWrite(t *testing.T) {
+	exec := filepath.Join(cwd, "testdata", "zeekwritefail.sh")
```
alfred-landrum: Was there something you wanted to verify related to the logs that `zeekwritefail.sh` wrote?

zqd/handlers.go
```diff
@@ -275,7 +275,7 @@ func handlePacketPost(c *Core, w http.ResponseWriter, r *http.Request) {
 		}
 	}
 	taskEnd := api.TaskEnd{Type: "TaskEnd", TaskID: taskId}
-	if err != nil {
+	if err := proc.Err(); err != nil {
```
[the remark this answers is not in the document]
mattnibs: Yes it did!

zqd/handlers_zeek_test.go
mattnibs: Sure.

zqd/handlers_zeek_test.go
```diff
@@ -16,98 +16,169 @@ import (
...
-type PacketPostSuite struct {
-	suite.Suite
-	core     *zqd.Core
-	space    string
-	pcapfile string
-	payloads []interface{}
-}
+var (
+	zeekpath = os.Getenv("ZEEK")
+	cwd      = getwd()
```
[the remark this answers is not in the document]
mattnibs: This is how I originally had it but `exec` requires an absolute path.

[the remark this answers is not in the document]
alfred-landrum: How about using `filepath.Abs` as needed then? Or wrapping it in test utility function: `abspath(t *testing.T, ...paths)` that will do a filepath join & abs as needed?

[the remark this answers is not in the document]
mattnibs: sure

zqd/handlers_zeek_test.go
mattnibs: No there was not. It was more relevant when the api was waiting on zeek logfiles to return status code 2xx... I still think it's an interesting test and it also ensures that no data remains in the space (with the subsequent test that calls to spaceInfo).

zqd/handlers_zeek_test.go
mattnibs: No, the test was just more readable when it test against the whole struct. I had to add the InDelta on `info.Size` to account for the change. Yeah I'll just comment on that.

PR #400, opened 2020-03-07
Sort in reverse on ts when creating all.bzng

philrz:
In https://github.com/brimsec/zq/pull/378, we accidentally lost the reverse sorting on the `ts` field when generating the `all.bzng`, which results in an incorrect histogram view. E.g. for one of our sample data sets:

image

### DIFF
--- a/zqd/packet/packet.go
+++ b/zqd/packet/packet.go
@@ -239,8 +239,8 @@ func (p *IngestProcess) writeData(ctx context.Context) error {
 		return err
 	}
 
-	minTs := headW.r.Ts
-	maxTs := tailW.r.Ts
+	minTs := tailW.r.Ts
+	maxTs := headW.r.Ts
 
 	if err := bzngfile.Close(); err != nil {
 		return err
--- a/zqd/packet/packet.go
+++ b/zqd/packet/packet.go
@@ -228,7 +228,7 @@ func (p *IngestProcess) writeData(ctx context.Context) error {
 		return err
 	}
 	zw := bzngio.NewWriter(bzngfile)
-	const program = "sort -limit 10000000 ts | (filter *; head 1; tail 1)"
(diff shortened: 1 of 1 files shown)


philrz:
My first commit to just restore the `-r` yielded another problem, where now the app refused to render the histogram. As sniffed out of Wireshark, the error was:

```
POST /search?format=zjson HTTP/1.1
Host: localhost:9867
Connection: keep-alive
Content-Length: 238
sec-ch-ua: 
Authorization: Basic Og==
Origin: file://
Sec-Fetch-Dest: empty
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_3) AppleWebKit/537.36 (KHTML, like Gecko) Brim/0.3.0 Chrome/78.0.3904.130 Electron/7.1.9 Safari/537.36
Content-Type: text/plain;charset=UTF-8
Accept: */*
Sec-Fetch-Site: cross-site
Sec-Fetch-Mode: cors
Accept-Encoding: gzip, deflate, br
Accept-Language: en-US

{"proc":{"op":"SequentialProc","procs":[{"op":"FilterProc","filter":{"op":"MatchAll"}},{"op":"HeadProc","count":500}]},"space":"all.pcap.brim","dir":-1,"span":{"ts":{"sec":1428917684,"ns":732000000},"dur":{"sec":-3352172,"ns":212000000}}}[1 bytes missing in capture file].HTTP/1.1 400 Bad Request
Content-Type: text/plain; charset=utf-8
X-Content-Type-Options: nosniff
Date: Sat, 07 Mar 2020 19:55:19 GMT
Content-Length: 42

time span must have non-negative duration
```

I confirmed with @henridf that this was a side-effect of some of the recent space caching changes, so he's pushed an additional commit to this branch to address that also.

philrz:
With the benefit of @henridf's latest commit, now the histogram is back to what we're accustomed to seeing.

image

PR #401, opened 2020-03-08
add support for pcap-ng

mccanne:
This commit adds support for pcap-ng.  The pcapio.Reader interface
was tweaked a bit to accommodate the changes so that a Reader can
be either a pcap-ng reader or a legacy pcap reader.  pcapio.Reader
uses the same trick as zio to autodetect whether input is pcap-ng
or pcap legacy.

The pcap-ng parsing code is derived from
https://github.com/google/gopacket/.

We added a simple test for pcap-ng along the lines of the pcap test
that was already there.  Testing is still limited for pcap and
needs to be addressed in a future PR.


henridf:
I noticed this earlier and wanted to dig in to the PR and `ranger` to understand things but didn't get around to it, so... some very black-boxy feedback:

It looks like result of `pcap index` on a (legacy) pcap file have changed with this PR.  Is this expected?

Here's master:
```json
    "Index": [
      {
        "X": 24,
        "Y0": 0,
        "Y1": 1521832548719396000
      },
      {
        "X": 390608,
        "Y0": 1521832548719400000,
        "Y1": 1521832548739203000
      },
      {
        "X": 799324,
        "Y0": 1521832548739205000,
        "Y1": 1521832548755799000
      },
      {
        "X": 1114496,
        "Y0": 1521832548755801000,
        "Y1": 1521832548765437000
      },
      {
```

and on this branch:
```json
    "Index": [
      {
        "X": 24,
        "Y0": 1521832548699327000,
        "Y1": 1521832548719400000
      },
      {
        "X": 390684,
        "Y0": 1521832548719488000,
        "Y1": 1521832548739205000
      },
      {
        "X": 799400,
        "Y0": 1521832548739251000,
        "Y1": 1521832548755801000
      },
   ```

**Comment by mccanne** [comment]
> It looks like result of `pcap index` on a (legacy) pcap file have changed with this PR. Is this expected?

I didn't expect that but I'm not surprised, so I'll have a look.  It looks like it still is correct, but I will double check.  Probably won't get to this for a few hours.

I changed a lot and found and fixed several bugs so I'll try to do some more comprehensive QAing...

File: pcap/pcapio/ngread.go [hunk 1]
```diff
@@ -0,0 +1,317 @@
...
+			ifno := int(r.getUint32(block[8:12]))
+			if ifno >= len(r.ifaces) {
+				return nil, 0, fmt.Errorf("Interface id %d not present in section (have only %d interfaces)", r.ci.InterfaceIndex, len(r.ifaces))
+			}
+			caplen := int(r.getUint32(block[20:24]))
+			//length := int(r.getUint32(block[24:28]))
+			packet := block[28:]
+			if len(packet) < caplen {
+				return nil, 0, ErrCorruptPcap
+			}
```
alfred-landrum: There's a lot of commonality between this and `Packet()`; would it be possible to share the logic?

pcap/pcapio/ngread.go
```diff
@@ -0,0 +1,317 @@
...
+}
+
+// NInterfaces returns the current number of interfaces.
+func (r *NgReader) NInterfaces() int {
+	return len(r.ifaces)
+}
+
+// Resolution returns the timestamp resolution of acquired timestamps before scaling to NanosecondTimestampResolution.
+func (r *NgReader) Resolution() gopacket.TimestampResolution {
+	//XXX
```
alfred-landrum: If this isn't used, would you remove the method?

mccanne:
> > It looks like result of `pcap index` on a (legacy) pcap file have changed with this PR. Is this expected?
> 
> I didn't expect that but I'm not surprised, so I'll have a look. It looks like it still is correct, but I will double check. Probably won't get to this for a few hours.
> 
> I changed a lot and found and fixed several bugs so I'll try to do some more comprehensive QAing...

Ok @henridf, I looked at this a bit closer and I think the newer code handles the first point correctly with the re-write to handle pcap-ng.  The first index entry should be X=24 (the size of the pcap legacy header and the Y0/Y1 range should be the timestamp range, which should *no*t start from 0 as with the old code in master.

pcap/pcapio/ngread.go
```diff
@@ -0,0 +1,317 @@
...
+	}
+	if err := ret.parseSectionHeader(block); err != nil {
+		return nil, err
+	}
+	ret.first = block
+	return ret, nil
+}
+
+func (r *NgReader) Packet(block []byte) ([]byte, nano.Ts, layers.LinkType) {
+	//XXX support only enhance packet blocks right now
```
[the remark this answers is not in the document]
mccanne: will add clarifying comment for the Packet() method

pcap/pcapio/ngread.go
```diff
@@ -0,0 +1,317 @@
...
+	"github.com/brimsec/zq/pkg/nano"
+	"github.com/brimsec/zq/pkg/peeker"
+	"github.com/google/gopacket"
+	"github.com/google/gopacket/layers"
+)
+
+var ErrCorruptPcap = errors.New("bad pcap file")
+
+// NgReaderOptions holds options for reading a pcapng file
+type NgReaderOptions struct {
```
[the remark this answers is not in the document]
mccanne: ok, that was the gopacket code.

pcap/pcapio/ngread.go
```diff
@@ -0,0 +1,317 @@
...
+
+	"github.com/brimsec/zq/pkg/nano"
+	"github.com/brimsec/zq/pkg/peeker"
+	"github.com/google/gopacket"
+	"github.com/google/gopacket/layers"
+)
+
+var ErrCorruptPcap = errors.New("bad pcap file")
+
+// NgReaderOptions holds options for reading a pcapng file
```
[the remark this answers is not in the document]
mccanne: ok, gopacket code

pcap/pcapio/ngread.go
```diff
@@ -0,0 +1,317 @@
...
+type NgReaderOptions struct {
+	// ErrorOnMismatchingLinkType enables returning an error if a packet with a link type not matching the first interface is encountered and WantMixedLinkType == false.
```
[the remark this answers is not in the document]
mccanne: now that I'm making these edits, we doon't have a use for these options so I will just delete all thsi stuff.

pcap/pcapio/ngread.go
```diff
@@ -0,0 +1,317 @@
...
+func (r *NgReader) Packet(block []byte) ([]byte, nano.Ts, layers.LinkType) {
+	//XXX support only enhance packet blocks right now
+	if len(block) < 28 {
+		return nil, 0, 0
+	}
+	ifno := int(r.getUint32(block[8:12]))
+	if ifno >= len(r.ifaces) {
+		return nil, 0, 0
+	}
+	t := time.Unix(r.convertTime(ifno, uint64(r.getUint32(block[12:16]))<<32|uint64(r.getUint32(block[16:20])))).UTC()
```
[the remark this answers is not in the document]
mccanne: Ok, will do.  I think this can be simplified but didn't want to break the existing pcapgo code.

pcap/pcapio/ngread.go
```diff
@@ -0,0 +1,317 @@
...
+		r.bigEndian = true
+	} else if binary.LittleEndian.Uint32(b[8:12]) == ngByteOrderMagic {
+		r.bigEndian = false
+	} else {
+		return errors.New("Wrong byte order value in Section Header")
+	}
+	return nil
+}
+
+// readOption reads a single arbitrary option (type and value). If there is no space left for options and end of options is missing, it is faked.
```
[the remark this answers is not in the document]
mccanne: will fix

pcap/pcapio/ngread.go
```diff
@@ -0,0 +1,317 @@
...
+			}
+			intf.TimestampResolution = NgResolution(body[0])
+		}
+	}
+	if intf.TimestampResolution == 0 {
+		intf.TimestampResolution = 6
+	}
+
+	//parse options
+	if intf.TimestampResolution.Binary() {
```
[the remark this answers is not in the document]
mccanne: I'd rather just leave it for now since it's the pcapgo code.

pcap/pcapio/ngread.go
mccanne: ok

pcap/pcapio/ngread.go
mccanne: good idea

pcap/pcapio/ngread.go
```diff
@@ -0,0 +1,310 @@
...
+	caplen := int(r.getUint32(block[20:24]))
+	packet := block[PacketBlockHeaderLen:]
+	if len(packet) < caplen {
+		return nil, 0
+	}
+	return packet[:caplen], ifno
+}
+
+// Packet returns the captured portion of a packet from an enhanced packet
+// block return by Read() (i.e., with BlockTYpe equal to TypePacket) beginning
```
mattnibs:
```suggestion
// block return by Read() (i.e., with BlockType equal to TypePacket) beginning
```

pcap/pcapio/ngread.go
```diff
@@ -0,0 +1,310 @@
...
+// with the link-layer header.  It also extracts the capture timestamp and link
+// layer type and returns those values along with the packet.  If an error is
+// encountered, zero values are returned for the three values.  PCAP-NG simple
+// packet types aren't supported yet (we presume thiis type of trace is rare and
```
mattnibs:
```suggestion
// packet types aren't supported yet (we presume this type of trace is rare and
```

pcap/pcapio/ngread.go
```diff
@@ -0,0 +1,310 @@
...
+// does not fit our use case here as these traces do not include capture timestamps
+// annd the point here is to pull our ranges of packets from a large pcap based
```
mattnibs:
```suggestion
// and the point here is to pull our ranges of packets from a large pcap based
```

pcap/index.go
```diff
@@ -21,40 +21,72 @@ type Section struct {
...
 
-// CreateIndex creates an index for a pcap file.  If the file isn't
-// a pcap file, an error is returned.
-func CreateIndex(r io.Reader, limit int) (Index, error) {
-	reader, err := pcapio.NewPcapReader(r) // XXX TBD: lookup the right reader
-	if err != nil {
-		return nil, err
-	}
+// CreateIndex creates an index for a pcap file.
+func CreateIndex(reader pcapio.Reader, limit int) (Index, error) {
```
mattnibs: Unless necessary I would be obliged to keep this interface the same- more boiler plate for clients.

pcap/pcapio/ngread.go
```diff
@@ -0,0 +1,310 @@
...
+}
+
+func (r *NgReader) parseSectionHeader(b []byte) error {
+	if len(b) < 12 {
+		return ErrCorruptPcap
+	}
+	if binary.BigEndian.Uint32(b[8:12]) == ngByteOrderMagic {
+		r.bigEndian = true
+	} else if binary.LittleEndian.Uint32(b[8:12]) == ngByteOrderMagic {
+		r.bigEndian = false
```
mattnibs: nit: Instead of this why not replace `bigEndian bool` with `byteOrder binary.ByteOrder` and do:
```golang
if binary.BigEndian.Uint32(b[8:12]) == ngByteOrderMagic {
	r.byteOrder = binary.BigEndian
} else if binary.LittleEndian.Uint32(b[8:12]) == ngByteOrderMagic {
	r.byteOrder = binary.LittleEndian
```
This way you can get rid of your `getUint64/32/16` helpers.

pcap/pcapio/ngread.go
```diff
@@ -0,0 +1,310 @@
...
+	} else {
+		return errors.New("Wrong byte order value in Section Header")
+	}
```
mattnibs: nit: It would be nice for this to check Major Version and throw an error if != 1 (or at least a comment about that).

pcap/pcapio/ngread.go
```diff
@@ -0,0 +1,310 @@
...
+
+// NewNgReader initializes a new writer, reads the first section header,
+// and if necessary according to the options the first interface.
+func NewNgReader(r io.Reader) (*NgReader, error) {
+	ret := &NgReader{
+		Reader: peeker.NewReader(r, 32*1024, 1024*1024),
+	}
+
+	//pcapng _must_ start with a section header
+	typ, block, err := ret.readBlock()
```
mattnibs: If this pcap was recorded with written with bigEndian byte order (the default is littleEndian) I believe this going to get return an error: "pcap-ng trailer length mismatch" (Same goes for `Read()`. I believe you're going to need to read the block type bits, and if its a section header run this through `parseSectionHeader`, before  calling `readBlock`.

zqd/handlers.go
```diff
@@ -85,6 +86,12 @@ func handlePacketSearch(c *Core, w http.ResponseWriter, r *http.Request) {
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
+	pcapReader, err := pcapio.NewReader(slicer)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
```
mattnibs:
```suggestion
		http.Error(w, err.Error(), http.StatusInternalServerError)
```

pcap/pcapio/reader.go
```diff
@@ -27,6 +24,23 @@ type Info struct {
...
 }
+
+// NewReader returns a Reader by trying both pcap and pcap-ng formats.
+func NewReader(r io.Reader) (Reader, error) {
+	recorder := detector.NewRecorder(r)
+	track := detector.NewTrack(recorder)
+	_, err := NewPcapReader(track)
+	if err == nil {
+		return NewPcapReader(recorder)
+	}
```
mattnibs: Can fixed in a follow up pr but if this returns an error because it's an unsupported pcap version, that error should get returned.

pcap/pcapio/ngread.go
```diff
@@ -0,0 +1,317 @@
...
+// If the type is a section header, endianess is also read.
+func (r *NgReader) readBlock() (ngBlockType, []byte, error) {
+	hdr, err := r.Peek(8)
+	if err != nil {
+		return 0, nil, err
+	}
+	typ := ngBlockType(r.getUint32(hdr[:4]))
+	length := r.getUint32(hdr[4:8])
+	if length == 0 {
+		// avoid infinite loop for bad input
```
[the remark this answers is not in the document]
mccanne: You know, I like to point this one out because its a common flaw in protocol design and protocol implementation and thus I think it's worth it to include the comment.   May I reminisce?  I put up a bug fix at Riverbed in the CIFS protocol accelerator module because a length==0 check like this was missing and Kartik closed it with no fix explaining that a sender shouldn't put 0 in the length field.  Ugh.  Van Jacobson taught me years ago that length fields should never include the length of the fixed size header contained within referenced payload so that: (1) you don't have to check for length==0 to avoid an infinite loop (i.e., you can't make this mistake even if you tried) and (2) it lets the payload length be a little longer for the bit budget you have for the length field (which in the old days of TCP/IP header design made some difference but nowadays now big deal).  More to the point, the previous code from gopacket has not quite this bug, but the corollary, where it is computing a length incorrectly and causing an unsigned integer underflow for length value < 8.  The moral of the story is that the  block length field in the pcap-ng spec should have been defined to be the actual length less 12 (to account for the fixed size of the 4-byte block type, block length, and block length trailer) rather than defined as encoding the overall length.  It would be fun to create a pcap-ng file that is fully compliant except for a block that has a length of 0 and feed it to various pcap-related programs and see which ones lock up.  That all said, this discussion made me realize there is a different bug here, which is panic if the length is 1, 2, or 3, but according to Kartik, a file should never have those values here.  Ha!  I will fix this.

pcap/pcapio/ngread.go
```diff
@@ -0,0 +1,317 @@
...
+		return 0, nil, errors.New("zero-length pcap-ng block")
+	}
+	// The next part is a bit messed up since a section header could change the endianess...
+	// So first read then length just into a buffer, check if its a section header and then do the endianess part...
+	//XXX move this above
```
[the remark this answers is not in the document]
mccanne: This is more gopacket vestige.  I will have a look.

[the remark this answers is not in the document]
mccanne: oops, removed.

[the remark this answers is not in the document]
mccanne: btw, there was profanity here from gopacket that I already edited to "messed up"

cmd/pcap/cut/command.go
```diff
@@ -130,19 +130,24 @@ func (c *Command) Run(args []string) error {
 		out = w
 	}
 
-	// XXX assumes legacy pcap format, will fix in next PR
-	reader, err := pcapio.NewPcapReader(in)
+	reader, err := pcapio.NewReader(in)
```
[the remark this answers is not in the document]
mccanne: I will give it a go.

pcap/pcapio/ngread.go
```diff
@@ -0,0 +1,317 @@
...
+		}
+	}
+	intf.scaleDown = 1
+	intf.scaleUp = 1
+	if intf.secondMask < 1e9 {
+		intf.scaleUp = 1e9 / intf.secondMask
+	} else {
+		intf.scaleDown = intf.secondMask / 1e9
+	}
+	//XXX this list should be per section no?
```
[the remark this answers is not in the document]
mccanne: good catch on the note to self.  I needed to reset the interface list in parseSectionHeader()... fixed.

pcap/index.go
mccanne: excellent point.  will fix.

pcap/pcapio/ngread.go
mccanne: How about later?  This is all gopacket logic that I was just hoping would work for us.

pcap/pcapio/ngread.go
mccanne: How about later? This is all gopacket logic that I was just hoping would work for us.

pcap/pcapio/ngread.go
mattnibs: ok

pcap/pcapio/ngread.go
mattnibs: ok

zqd/handlers.go
mccanne: I didn't think that was right but I copied it from the one above, which should also change right?

mattnibs: Yes.

pcap/pcapio/reader.go
mccanne: NewPcapReader() returns the error as the second arg

mattnibs: If NewPcapReader returns a "this version of legacy pcap is invalid err" from a legacy pcap, the error will be over-written by whatever error NewNgReader finds.

pcap/pcapio/ngread.go
mccanne: Will fix it!

pcap/pcapio/ngread.go
```diff
@@ -0,0 +1,317 @@
...
+	ifaces        []NgInterface
+	currentOption ngOption
+	first         []byte
+	ci            gopacket.CaptureInfo
+	bigEndian     bool
+	offset        uint64
+}
+
+// NewNgReader initializes a new writer, reads the first section header, and if necessary according to the options the first interface.
+func NewNgReader(r io.Reader) (*NgReader, error) {
```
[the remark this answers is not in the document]
mccanne: Yes good point, NgReaderOptions is gone now

pcap/pcapio/ngread.go
mccanne: good catch

pcap/pcapio/reader.go
mccanne: No, if they both fail, then it's ErrCorruptPcap

mccanne: I think this is ok.
GSD

PR #402, opened 2020-03-09
Make space available for querying during ingest

henridf:
This PR adjusts the ingest process following @mccanne's design to make data available for querying as it is processed, rather than waiting for everything to land before allowing anything to be queried.

The main changes are:
   - Indexing and zeek processing of the pcap now happen sequentially. Indexing is first; a HTTP 2xx is returned upon successful index creation.
  - Instead of waiting for all zeek logs to be ready, and then running a one-time bzng conversion job after that, we now take "snapshots" of the zeek logs as they are being generated, and make these snapshots
available for regular querying while zeek processing continues. A final snapshot is done after all logs have been processed.

@mccanne, I did not implement the "stable" scheme in this PR, as I had a question that raised offline (on the issue). I can add it here once that's resolved or in a different PR. 

This change needs tests. I'm waiting for #399 to get in and will add those after that.

### DIFF
--- a/zqd/handlers.go
+++ b/zqd/handlers.go
@@ -252,11 +252,10 @@ func handlePacketPost(c *Core, w http.ResponseWriter, r *http.Request) {
 		case <-ticker.C:
 		}
 
-		var minTs, maxTs nano.Time
+		var minTs, maxTs *nano.Ts
 		if minTs, maxTs, err = s.GetTimes(); err != nil {
 			break
 		}
-
 		status := api.PacketPostStatus{
 			Type:           "PacketPostStatus",
 			StartTime:      proc.StartTime,
--- a/zqd/handlers.go
+++ b/zqd/handlers.go
@@ -248,14 +248,23 @@ func handlePacketPost(c *Core, w http.ResponseWriter, r *http.Request) {
 		select {
 		case <-proc.Done():
 			done = true
+		case <-proc.Snap():
 		case <-ticker.C:
 		}
+
+		var minTs, maxTs nano.Time
+		if minTs, maxTs, err = s.GetTimes(); err != nil {
+			break
+		}
+
 		status := api.PacketPostStatus{
 			Type:           "PacketPostStatus",
 			StartTime:      proc.StartTime,
 			UpdateTime:     nano.Now(),
 			PacketSize:     proc.PcapSize,
 			PacketReadSize: proc.PcapReadSize(),
+			MinTime:        minTs,
+			MaxTime:        maxTs,
 		}
 		if err := pipe.Send(status); err != nil {
 			// XXX This should be zap instead.
--- a/zqd/handlers.go
+++ b/zqd/handlers.go
@@ -231,7 +231,7 @@ func handlePacketPost(c *Core, w http.ResponseWriter, r *http.Request) {
 	}
 	w.Header().Set("Content-Type", "application/ndjson")
 	w.Header().Set("Transfer-Encoding", "chunked")
-	w.WriteHeader(http.StatusCreated)
+	w.WriteHeader(http.StatusAccepted)
 	pipe := api.NewJSONPipe(w)
 	taskId := atomic.AddInt64(&taskCount, 1)
 	taskStart := api.TaskStart{Type: "TaskStart", TaskID: taskId}
--- a/zqd/packet/packet.go
+++ b/zqd/packet/packet.go
@@ -36,7 +36,7 @@ type IngestProcess struct {
 	pcapPath     string
 	pcapReadSize int64
 	logdir       string
-	done         chan struct{}
+	done, snap   chan struct{}
 	err          error
 	zeekExec     string
 }
@@ -72,6 +72,7 @@ func IngestFile(ctx context.Context, s *space.Space, pcap, zeekExec string) (*In
 		pcapPath:  pcap,
 		logdir:    logdir,
 		done:      make(chan struct{}),
+		snap:      make(chan struct{}),
 		zeekExec:  zeekExec,
 	}
 	if err = p.indexPcap(ctx); err != nil {
@@ -85,6 +86,7 @@ func IngestFile(ctx context.Context, s *space.Space, pcap, zeekExec string) (*In
 	go func() {
 		p.err = p.run(ctx)
 		close(p.done)
+		close(p.snap)
 	}()
 	return p, nil
 }
@@ -117,6 +119,10 @@ outer:
 				if err := p.writeData(ctx); err != nil {
 					abort()
 					return err
+				}
+				select {
+				case p.snap <- struct{}{}:
+				default:
 				}
 				next = 2 * next
 			}
@@ -207,6 +213,12 @@ func (p *IngestProcess) Err() error {
 // Done returns a chan that emits when the ingest process is complete.
 func (p *IngestProcess) Done() <-chan struct{} {
 	return p.done
+}
+
+// Snap returns a chan that emits every time a snapshot is made. It
+// should no longer be read from after Done() has emitted.
+func (p *IngestProcess) Snap() <-chan struct{} {
+	return p.snap
 }
 
 type recWriter struct {
--- a/zqd/packet/packet.go
+++ b/zqd/packet/packet.go
@@ -9,6 +9,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"sync/atomic"
+	"time"
 
 	"github.com/brimsec/zq/pcap"
 	"github.com/brimsec/zq/pkg/nano"
@@ -77,6 +78,10 @@ func IngestFile(ctx context.Context, s *space.Space, pcap, zeekExec string) (*In
 		os.Remove(p.space.DataPath(IndexFile))
 		return nil, err
 	}
+	if err = p.space.SetPacketPath(p.pcapPath); err != nil {
+		os.Remove(p.space.DataPath(IndexFile))
+		return nil, err
+	}
 	go func() {
 		p.err = p.run(ctx)
 		close(p.done)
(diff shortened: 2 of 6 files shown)


henridf:
@jameskerr with this PR, as soon as you get a first `PacketPostStatus`, the space is available for querying (partially). When you get the TaskEnd message, you know that all data is ingested.

zqd/packet/packet.go
```diff
@@ -74,64 +74,83 @@ func IngestFile(ctx context.Context, s *space.Space, pcap, zeekExec string) (*In
...
+	start := time.Now()
+	next := time.Second
+outer:
+	for {
+		select {
+		case <-slurpDone:
+			break outer
+		case t := <-ticker.C:
+			if t.After(start.Add(next)) {
+				if err := p.writeData(ctx); err != nil {
```
mccanne: writeData() is a vague name.  How about createSnapshot()?

zqd/packet/packet.go
```diff
@@ -74,64 +74,83 @@ func IngestFile(ctx context.Context, s *space.Space, pcap, zeekExec string) (*In
...
+	defer ticker.Stop()
+	start := time.Now()
+	next := time.Second
+outer:
+	for {
+		select {
+		case <-slurpDone:
+			break outer
+		case t := <-ticker.C:
+			if t.After(start.Add(next)) {
```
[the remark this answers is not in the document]
henridf: I thought of this issue too, but then I read `go doc time.NewTicker` which says:

"It adjusts the intervals or drops ticks to make up for slow receivers."

zqd/packet/packet.go
henridf: I agree and there's a bunch of other names around here that I think deserve adjusting, but I was going to keep this for a separate PR... I'll adjust writeData in this one.

mccanne: ok, leave it

henridf:
> To do this, Done() channel on the "proc" instead could be a status channel that multiplexes updates from the zeek writer and the snapshotter and gets closed when ingest is finished.

@mccanne totally agree and I started out wanting to do streaming updates as you describe, but that would required the `packet` package to assemble api messages (or bits of them), which it appears is no good.

However your comment makes me realize we could have a separate snapshot channel on which the ingest process sends a `struct{}` at every snapshot; the handler can then use notifiction to send a "snapshot done" API message.

jameskerr:
@henridf Awesome, thanks! I'll adjust the app to be ready to search after I receive the first one. @mccanne I don't quite follow how you'd like to stream back results from ingest. Maybe we can talk over zoom later.

zqd/packet/packet.go
```diff
@@ -74,64 +74,83 @@ func IngestFile(ctx context.Context, s *space.Space, pcap, zeekExec string) (*In
...
-	w, err := fsnotify.NewWatcher()
+func (p *IngestProcess) indexPcap(ctx context.Context) error {
+	pcapfile, err := os.Open(p.pcapPath)
 	if err != nil {
 		return err
 	}
-	defer w.Close()
-	if err := w.Add(p.logdir); err != nil {
+	defer pcapfile.Close()
+	iw := pcap.NewIndexWriter(10000)
```
mattnibs: this is more a nit, but: I would get rid of the pcap.IndexWriter and just call pcap.CreateIndex- not needed now that we're not piping it all. pcap.IndexWriter could be deleted.

zqd/packet/packet.go
```diff
@@ -74,64 +74,83 @@ func IngestFile(ctx context.Context, s *space.Space, pcap, zeekExec string) (*In
...
-	os.Remove(p.space.DataPath(IndexFile))
-	p.space.SetPacketPath("")
-	return err
 }
 
-// awaitZeekLogs waits for the first zeek logs to hit the file system. Should
-// an error occur before this happens, the error will be returned.
-func (p *IngestProcess) awaitZeekLogs() error {
-	w, err := fsnotify.NewWatcher()
+func (p *IngestProcess) indexPcap(ctx context.Context) error {
```
mattnibs:
```suggestion
func (p *IngestProcess) indexPcap() error {
```
ctx doesn't appear to be used here. This can be removed.

zqd/packet/packet.go
henridf: Cool, I hadn't noticed that IndexWriter was only needed in the previous setup. I'll delete it in this PR.

zqd/packet/packet.go
mattnibs: I might be nice at some point to have `ctx` stop the copy, but I don't think this is important- the time it will take to create the index will be negligible compared to the rest of this procedure.

zqd/packet/packet.go
```diff
@@ -74,64 +74,83 @@ func IngestFile(ctx context.Context, s *space.Space, pcap, zeekExec string) (*In
...
+	}
+	if err = p.space.SetPacketPath(p.pcapPath); err != nil {
+		os.Remove(p.space.DataPath(IndexFile))
+		return nil, err
+	}
 	go func() {
 		p.err = p.run(ctx)
 		close(p.done)
 	}()
-	return p, p.awaitZeekLogs()
```
mattnibs: My ideal would be that the inner `writeData` loops waits to kick off until after log files... that said I think for now it's ok to get rid of this functionality as well `fsnotify`. If you are doing this however, I think `writeData` needs to be adjusted so it doesn't return an error in the event that no long files exist.

henridf:
> we could have a separate snapshot channel on which the ingest process sends a struct{} at every snapshot; the handler can then use notifiction to send a "snapshot done" API message.

@mccanne (and/or anyone else) I did this in ff088397743ae59596daf16860e15b97f88ae85e , please take a look

Rather than create a separate snapshot message, i just added the minTs/maxTs fields to PostStatusUpdate. This way @jameskerr just gets these messages and updates the mints/maxts every time (it's possible that two subsequent messages have the same values in these fields, but that's fine).

zqd/packet/packet.go
mattnibs: My intention was that once the client got a `200` response code, it would mean there was data available for the client to query. The `status` api will need to change to somehow notify clients when data is available in the space to query.

zqd/handlers.go
```diff
@@ -248,14 +248,22 @@ func handlePacketPost(c *Core, w http.ResponseWriter, r *http.Request) {
 		select {
 		case <-proc.Done():
 			done = true
+		case <-proc.Snap():
 		case <-ticker.C:
```
mattnibs: Hmm, it would be nice for the `ticker` to be reset here, though I don't see away to do this with the existing `time.Ticker` api.

mattnibs:
> Indexing and zeek processing of the pcap now happen sequentially. Indexing is first; a HTTP 2xx is returned upon successful index creation.

The index isn't really helpful unless there is data pointing to pcaps. I would consider either here or in a separate pr waiting for some data to be "snapshotted" until the 2xx response header is returned.

alfred-landrum:
> The index isn't really helpful unless there is data pointing to pcaps. I would consider either here or in a separate pr waiting for some data to be "snapshotted" until the 2xx response header is returned.

This is useful in the case of a bad pcap file, since we'll fail the request faster & before we start up zeek & the rest of the ingest actions.

zqd/packet/packet.go
henridf: > If you are doing this however, I think writeData needs to be adjusted so it doesn't return an error in the event that no long files exist. 

Good point, I'll do that

henridf: > My intention was that once the client got a 200 response code, it would mean there was data available for the client to query. The status api will need to change to somehow notify clients when data is available in the space to query.

This seems reasonable, but I'm not not sure 200 conventionally means "some data has been ingested and we'll stream you updates about the rest". I think a HTTP 202 accepted would be the right response, signaling that the job has been accepted for processing. And as written else where, as soon as you get a status update that tells you something is available... then you know something is available.

(As a side note, on the whole this PR tries to follow @mccanne's design referred to in the description; discussion/revisiting of that design should probably involve him).

henridf:
> The index isn't really helpful unless there is data pointing to pcaps. I would consider either here or in a separate pr waiting for some data to be "snapshotted" until the 2xx response header is returned.

@mattnibs just saw this comment after responding above; same response as https://github.com/brimsec/zq/pull/402#discussion_r389853935

mattnibs:
> This is useful in the case of a bad pcap file, since we'll fail the request faster & before we start up zeek & the rest of the ingest actions.

My comment isn't suggesting that we shouldn't return 4xx immediately in the event of a bad pcap.

zqd/handlers.go
henridf: Yeah. But worst case we'll send a handful of redundant status messages.

zqd/packet/packet.go
henridf: Agreed - as I wrote offline `pcap index` is 💥 fast

zqd/packet/packet.go
mattnibs: In a future pr we should consider adding back `fsnotify` to kick off the snapshot loop... It would be nice to have the client not wait an arbitrary amount of time on every request to begin querying spaces. That said for now I think this is fine.

mccanne:
> Rather than create a separate snapshot message, i just added the minTs/maxTs fields to PostStatusUpdate. This way @jameskerr just gets these messages and updates the app's minTs/maxTs every time (it's possible that two subsequent messages have the same values in these fields, but that's fine).

This looks good.  App can compare its min/maxTs to the update and update when different.  And since trigger an update from the snapshot, the first update will be nice and tight.

zqd/packet/packet.go
henridf: changed

zqd/packet/packet.go
henridf: It's the time of the pcap index, which is very fast. Given that we've gone from "time of full zeek processing" to "time of pcap index", I agree this that further optimization can wait.

henridf:
As noted above, tests will follow in a separate PR.

PR #405, opened 2020-03-10
Provide better errors on zeek related failures

mattnibs:
Include any output from stderr in error message along with exit status
code and the fact that the error came from the zeek subprocess.

If zeek cannot be found, or the user does not have execute permissions 
the request will return with a 500 level errors. All other errors will occur after
the 2xx request headers are sent and thus will be streamed back in the json
TaskEnd payload.

### DIFF
--- a/zqd/packet/packet.go
+++ b/zqd/packet/packet.go
@@ -1,13 +1,16 @@
 package packet
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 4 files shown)


zqd/packet/packet.go
```diff
@@ -281,6 +293,8 @@ func (p *IngestProcess) startZeek(ctx context.Context, dir string) (*exec.Cmd, i
 	if err != nil {
 		return nil, nil, err
 	}
+	// capture stderr
```
mattnibs:
```suggestion
	// Capture stderr for error reporting.
```

zqd/packet/packet.go
```diff
@@ -187,6 +193,12 @@ func (p *IngestProcess) slurp(ctx context.Context) error {
 		return err
 	}
 	if err := zeekproc.Wait(); err != nil {
+		var exitErr *exec.ExitError
+		if errors.As(err, &exitErr) {
```
alfred-landrum: 🔥

PR #406, opened 2020-03-10
A little more packet post testing

henridf:
Add a little coverage of #402 (and fix one failure handling issue I found in the process).

### DIFF
--- a/zqd/handlers_zeek_test.go
+++ b/zqd/handlers_zeek_test.go
@@ -54,7 +54,10 @@ func TestPacketPostSuccess(t *testing.T) {
 	t.Run("PacketIndexExists", func(t *testing.T) {
 		require.FileExists(t, filepath.Join(p.core.Root, p.space, packet.IndexFile))
 	})
-	t.Run("ResponseMessages", func(t *testing.T) {
+	t.Run("TaskStartMessage", func(t *testing.T) {
+		status := p.payloads[0].(*api.TaskStart)
+		assert.Equal(t, status.Type, "TaskStart")
+	})
 	t.Run("StatusMessage", func(t *testing.T) {
 		info, err := os.Stat(p.pcapfile)
 		require.NoError(t, err)
@@ -64,6 +67,11 @@ func TestPacketPostSuccess(t *testing.T) {
 		assert.Equal(t, status.PacketReadSize, info.Size())
 		assert.Equal(t, nano.Unix(1501770877, 471635000), *status.MinTime)
 		assert.Equal(t, nano.Unix(1501770880, 988247000), *status.MaxTime)
+	})
+	t.Run("TaskEndMessage", func(t *testing.T) {
+		status := p.payloads[len(p.payloads)-1].(*api.TaskEnd)
+		assert.Equal(t, status.Type, "TaskEnd")
+		assert.Nil(t, status.Error)
 	})
 }
 
--- a/zqd/handlers_zeek_test.go
+++ b/zqd/handlers_zeek_test.go
@@ -55,12 +55,15 @@ func TestPacketPostSuccess(t *testing.T) {
 		require.FileExists(t, filepath.Join(p.core.Root, p.space, packet.IndexFile))
 	})
 	t.Run("ResponseMessages", func(t *testing.T) {
+	t.Run("StatusMessage", func(t *testing.T) {
 		info, err := os.Stat(p.pcapfile)
 		require.NoError(t, err)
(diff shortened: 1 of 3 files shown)


zqd/handlers_zeek_test.go
```diff
@@ -92,30 +103,39 @@ func TestPacketPostZeekFailImmediate(t *testing.T) {
...
-		var info api.SpaceInfo
-		httpJSONSuccess(t, zqd.NewHandler(p.core), "GET", u, nil, &info)
-		expected := api.SpaceInfo{
-			Name: p.space,
-		}
-		require.Equal(t, expected, info)
-	})
+	var tests = []struct{ name, script string }{
+		{"Fail immediately", "zeekwritefail.sh"},
+		{"Fail after snapshot", "zeekwritefailslow.sh"},
```
mattnibs: Given that the trigger for snapshotting the data is now time based, I would assume just getting rid of the "zeekwritefail" script and going with the timeout test.

mattnibs: Also more a matter of tastes but: I find this table based approach more difficult to read as it doesn't follow the convention of the rest of test cases in the file. In my earlier approach I did something similar and @alfred-landrum made a comment that I believe is applicable here: https://github.com/brimsec/zq/pull/399#discussion_r389391872 

I'd rather just create functions `testTaskEndError` and `testEmptySpaceInfo`.

zqd/testdata/zeekwritefailslow.sh
```diff
@@ -0,0 +1,26 @@
...
+#types	time	string	addr	port	addr	port	enum	string	interval	count	count	string	bool	bool	count	string	count	count	count	count	set[string]
+1521911721.255387	C8Tful1TvM3Zf5x8fl	10.164.94.120	39681	10.47.3.155	3389	tcp	-	0.004266	97	19	RSTR	-	-	0	ShADTdtr	10	730	6	342	-
+1521911721.411148	CXWfTK3LRdiuQxBbM6	10.47.25.80	50817	10.128.0.218	23189	tcp	-	0.000486	0	0	REJ	-	-	0	Sr	2	104	2	80	-
+1521911721.926018	CM59GGQhNEoKONb5i	10.47.25.80	50817	10.128.0.218	23189	tcp	-	0.000538	0	0	REJ	-	-	0	Sr	2	104	2	80	-
+1521911722.690601	CuKFds250kxFgkhh8f	10.47.25.80	50813	10.128.0.218	27765	tcp	-	0.000546	0	0	REJ	-	-	0	Sr	2	104	2	80	-
+1521911723.205187	CBrzd94qfowOqJwCHa	10.47.25.80	50813	10.128.0.218	27765	tcp	-	0.000605	0	0	REJ	-	-	0	Sr	2	104	2	80	-
+1521911724.896854	CFzn9A3l9ppbMBVin3	10.164.94.120	40659	10.47.8.208	3389	tcp	-	0.011922	147	19	RSTR	-	-	0	ShADTdtr	10	830	6	342	-
+EOF
+
+sleep 2.5
```
mattnibs: I typically like to avoid hardcoded timeouts in test (tend to be error prone and make testing slow) but I get doing something like this so we have coverage in the meantime.

zqd/packet/packet.go
```diff
@@ -108,6 +108,7 @@ func (p *IngestProcess) run(ctx context.Context) error {
 	abort := func() {
 		os.RemoveAll(p.logdir)
 		os.Remove(p.space.DataPath(IndexFile))
+		os.Remove(p.space.DataPath("all.bzng"))
```
mattnibs: If there is a "slurp"/zeek error (bad name on my part) `run` will not wait for an ongoing snapshot routine to complete before it aborts. Thus there is a race and an all.bzng could be written after the fact. I don't think this needs to be addressed here, but it should be addressed at some point.

zqd/testdata/zeekwritefailslow.sh
henridf: I have a similar dislike, and took this route for expediency...

zqd/handlers_zeek_test.go
henridf: I lack strong feelings on these kind of aesthetic points, and would have adjusted if that's implicit convention... but your point about the two tests being redundant is good. I've removed the zeekwritefail script, and we're back to one test!

zqd/packet/packet.go
henridf: Good point. There's also crash recovery that could leave an all.bzng around and needs to be addressed (we have an issue for that).

PR #408, opened 2020-03-10
packet post sort limit reached error

mattnibs:
Ensure that if the sort limit is reached in the packet post
endpoint that a sensible error message is returned.

For testing purpose add a SortLimit field to zqd.Core
allowing test cases to test on custom sort limits.

### DIFF
--- a/zqd/api/api.go
+++ b/zqd/api/api.go
@@ -104,6 +104,12 @@ type SpacePostResponse SpacePostRequest
 
 type PacketPostRequest struct {
 	Path string `json:"path"`
+	// SortLimit specifies the limit of logs in posted pcap to sort. Its
+	// existence in the request is only as a hook for testing (though users
+	// running zqd in high memory environments may want to push this higher).
+	// Eventually zqd will sort an unlimited amount of logs and this can be
+	// taken out.
+	SortLimit int `json:"sort_limit,omitempty"`
 }
 
 type PacketPostStatus struct {
--- a/zqd/api/api.go
+++ b/zqd/api/api.go
@@ -104,12 +104,6 @@ type SpacePostResponse SpacePostRequest
 
 type PacketPostRequest struct {
 	Path string `json:"path"`
-	// SortLimit specifies the limit of logs in posted pcap to sort. Its
-	// existence in the request is only as a hook for testing (though users
-	// running zqd in high memory environments may want to push this higher).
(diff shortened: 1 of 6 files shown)


zqd/api/api.go
```diff
@@ -104,6 +104,12 @@ type SpacePostResponse SpacePostRequest
 
 type PacketPostRequest struct {
 	Path string `json:"path"`
+	// SortLimit specifies the limit of logs in posted pcap to sort. Its
```
henridf: It would be nice to be able to avoid broadening the API for testing support where it can be avoided... did you consider making this a constant in package `packet` and just modifying it for the test?

mattnibs: Normally I would be one the making this comment, however I believe @alfred-landrum is of the school of its okay to add hooks specifically for testing. That said, I could totally change package.DefaultSortLimit to be a `var` instead of a `const`- the down side is we would have to make sure this wasn't running in parallel.

henridf: It's not parallel right now, but hopefully will be, and I agree we should be parallel-proof. Which would mean this would have to be added into `Core` or something which is also intrusive. Well, if we're ok having testing fields in the public API, then I guess let's!

alfred-landrum: I don't think we should expose the ability to override the limit in the REST api, since that's exposed to end users of the service. I think there should be a way to setup the test server so that when it calls IngestFile, it calls it with a non-default.

mattnibs: Ok, I'll change this.

mattnibs: Alright changed. I put the setting zqd.Core.

zqd/handlers_zeek_test.go
```diff
@@ -75,18 +75,39 @@ func TestPacketPostSuccess(t *testing.T) {
...
+
 func TestPacketPostInvalidPcap(t *testing.T) {
-	p := packetPost(t, zeekpath, "./testdata/invalid.pcap", 400)
+	p := packetPost(t, "./testdata/invalid.pcap", 400)
 	defer p.cleanup()
 	t.Run("ErrorMessage", func(t *testing.T) {
 		// XXX Better error message here.
 		require.Regexp(t, "^bad pcap file*", string(p.body))
 	})
+	t.Run("EmptySpaceInfo", func(t *testing.T) {
```
alfred-landrum: 👍 to adding this

PR #409, opened 2020-03-10
add pprof routes to zqd

mattnibs:


mattnibs:
> Would you please add a cli flag to enable this, and have the feature off by default?

As you wish.

PR #410, opened 2020-03-11
/search: return empty results if no all.bzng

henridf:
Previously, /search was returning a "no such space" error when all.bzng couldn't be opened. Now, if all.bzng is missing, we return empty results. I ran into this issue while using Brim against zqd master. This was part of @mccanne 's design and should have been done in #402 .

### DIFF
--- a/zqd/handlers_test.go
+++ b/zqd/handlers_test.go
@@ -39,6 +39,14 @@ func TestSimpleSearch(t *testing.T) {
 	require.Equal(t, test.Trim(src), execSearch(t, c, space, "*"))
 }
 
+func TestSearchEmptySpace(t *testing.T) {
+	space := "test"
+	c := newCore(t)
+	defer os.RemoveAll(c.Root)
+	createSpace(t, c, space, "")
+	require.Equal(t, "", execSearch(t, c, space, "*"))
+}
+
 func TestSpaceList(t *testing.T) {
 	c := newCore(t)
 	defer os.RemoveAll(c.Root)
--- a/zqd/search/search.go
+++ b/zqd/search/search.go
@@ -5,6 +5,10 @@ package search
 import (
 	"context"
 	"errors"
+	"io"
+	"io/ioutil"
+	"os"
+	"strings"
 	"time"
 
 	"github.com/brimsec/zq/ast"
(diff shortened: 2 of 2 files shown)


henridf:
This PR needs a test. However, unless I'm missing something and there's a simpler way, I think testing this requires adding hooks into the ingest methods allowing the test to control (and pause) the timing of each step. In this case, we specifcially need to be able to run a /search just after the pcap has been indexed but before "slurp" has run.  It also requires creating an "async" version of the packetPost helper, which runs concurrently with the test func and sends the status messages it receives on a channel that the test func can use.

(A quicker but ugly hack would be to do the async packetPost and post a pcap that's large enough that we can be confident that zeek ingest will still be running when we post the `/search` right after receiving a PacketPostStatus..).

@alfred-landrum @mattnibs  what do you think: get this fix in and the test later? or do one of the test variants above (or some other) on this PR?

mattnibs:
@henridf re: testing: Couldn't you just do a POST to create a space, then run a search on the space?

alfred-landrum:
> @alfred-landrum @mattnibs what do you think: get this fix in and the test later? or do one of the test variants above (or some other) on this PR?

Let's get this in for now. I think we can improve how spaces are handled in a way that would make it easier to test them. In this case, for example, it'd be convenient if you could create something that has the interface of a Space, and accepts & indexes pcaps, but would never import actual records, then make sure search's always succeed.

henridf:
> @henridf re: testing: Couldn't you just do a POST to create a space, then run a search on the space?

@mattnibs that is an excellent idea! it's not _exactly_ testing the sequence as would happen with the issue I ran into via  app-based ingest, but for this purpose it should be just fine, is quick to write and is way better than no test.

I'll add this and merge.

mattnibs:
I'm ok with merging this, but I would like to know if this case is any different than if I created an empty space and queried it? If this is the case, though the test doesn't need to be created in this pr, it should exit (and is probably 4 lines of code).

zqd/handlers_test.go
```diff
@@ -39,6 +39,14 @@ func TestSimpleSearch(t *testing.T) {
 	require.Equal(t, test.Trim(src), execSearch(t, c, space, "*"))
 }
 
+func TestSearchEmptySpace(t *testing.T) {
+	space := "test"
+	c := newCore(t)
+	defer os.RemoveAll(c.Root)
+	createSpace(t, c, space, "")
+	require.Equal(t, "", execSearch(t, c, space, "*"))
```
mattnibs: 🔥

PR #429, opened 2020-03-17
Function calls

aswan:


expr/functions.go
```diff
@@ -0,0 +1,63 @@
...
+	case zng.IdInt16, zng.IdInt32, zng.IdInt64:
+		x = float64(args[0].value.(int64))
+	case zng.IdByte, zng.IdUint16, zng.IdUint32, zng.IdUint64:
+		x = float64(args[0].value.(uint64))
+	default:
+		return NativeValue{}, fmt.Errorf("Math.sqrt: %w", ErrBadArgument)
+	}
+
+	r := math.Sqrt(x)
+	if math.IsNaN(r) {
```
alfred-landrum: for the other reviewers: I asked @aswan about why we wanted to return errors for NaN's. That's not something we need to do forever, but we're not ready yet to declare NaN-like values in the language. So I think he'll add a comment here, and we'll file an issue to consider how we'll handle NaN-like values.

expr/functions.go
```diff
@@ -0,0 +1,63 @@
...
+	fn   Function
+}{
+	{"Math.sqrt", mathSqrt},
+}
+
+var allFnsMap map[string]Function
+var fnsInited = false
+
+func lookupFunction(name string) *Function {
+	if !fnsInited {
```
henridf: Why not just declar allFnsMap as a map literal instead of allFns and skip this initialization?

aswan: derp, no good reason.  updated.

expr/functions.go
```diff
@@ -0,0 +1,54 @@
...
+	"Math.sqrt": mathSqrt,
+}
+
+func lookupFunction(name string) *Function {
+	fn, ok := allFns[name]
+	if ok {
+		return &fn
+	}
+	return nil
+}
```
nwt: The Function zero value is nil, so you can simplify this to
```suggestion
func lookupFunction(name string) Function {
	return allFns[name]
}
```
or drop it entirely and just say `fn := allFns[node.Function]` in compileFunctionCall.

expr/function_test.go
```diff
@@ -0,0 +1,27 @@
+package expr_test
```
nwt: Nit: Maybe call this functions_test.go to match functions.go.

aswan: The compiler doesn't like that:
```
proc/cut.go:8:2: found packages expr (expr.go) and function (function_test.go) in /Users/aswan/src/zq/expr
```

nwt: Sorry I wasn't clearer here. I meant that since the new function code is in a file named functions.go, you might want to change the name of this corresponding test file to functions_test.go for symmetry.

zql/zql.peg
```diff
@@ -497,36 +497,23 @@ NotExpression
...
-    rest:("." n:FunctionNameComponent { RETURN(n) })* {
-        RETURN(XXX)
-    }
-
-FunctionNameComponent
   = FunctionNameStart FunctionNameRest* { RETURN(TEXT) }
 
 FunctionNameStart = [A-Za-z]
-FunctionNameRest = FunctionNameStart / [0-9]
+FunctionNameRest = FunctionNameStart / [.0-9]
```
henridf: I forgot to add this comment yesterday, and this can be sorted out later, but do we want function calls like `Math..sqrt(x)` to be syntactically valid, correct? (I'd think not).

aswan: Good question, I had started to write more complicated grammar rules earlier that would have prevented this earlier (https://github.com/brimsec/zq/blob/2d8a04c3d6aae61eb1d53c29fc093d64edd15c91/zql/zql.peg#L508-L515) but I wanted to keep the grammar simple.  If we never create functions with such names, then this would be a compilation error (as opposed to a parse error).  Do you think that's adequate?

henridf: Yep, I think this is totally adequate for now. Given that our syntax errors aren't currently very helpful, it's actually more useful to let `Math..sqrt()` all the way through to the function lookup check. Maybe we can revisite this when we get around to improving syntax errors with locations etc...

PR #432, opened 2020-03-18
Moving zeek process into own package

mattnibs:
The zeek pkg introduces the zeek.Launcher func and the zeek.Process
interface. These interfaces allows for test that aren't dependent on
having the zeek binary on the system running the tests.
Additionally passing the pcap reader directly as the zeek process
stdin ensures that we will not get the unfortunate "writing to pipe"
errors that occur when the zeek process is killed during ingest.

For the zqd listen command, if zeek is not available in the system
running the test, disable the packet ingest endpoint and signfy in
the logger output that zeek was not found.

### DIFF
--- a/zqd/handlers_test.go
+++ b/zqd/handlers_test.go
@@ -2,6 +2,7 @@ package zqd_test
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -23,6 +24,7 @@ import (
 	"github.com/brimsec/zq/zqd"
 	"github.com/brimsec/zq/zqd/api"
 	"github.com/brimsec/zq/zqd/space"
+	"github.com/brimsec/zq/zqd/zeek"
 	"github.com/brimsec/zq/zql"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -156,10 +158,7 @@ func TestSpaceDelete(t *testing.T) {
 	run := func(name string, cb func(*testing.T, string, *zqd.Core)) {
 		tmp := createTempDir(t)
 		defer os.RemoveAll(tmp)
-		c := zqd.NewCore(zqd.Config{
-			Root:   filepath.Join(tmp, "spaces"),
-			Logger: zaptest.NewLogger(t),
-		})
+		c := newCoreAtDir(t, filepath.Join(tmp, "spaces"))
 		require.NoError(t, os.Mkdir(c.Root, 0755))
 		t.Run(name, func(t *testing.T) {
 			cb(t, tmp, c)
@@ -189,7 +188,7 @@ func TestURLEncodingSupport(t *testing.T) {
 	c := newCore(t)
 	defer os.RemoveAll(c.Root)
 
-	rawSpace := "raw %<>space.brim"
+	rawSpace := "raw %space.brim"
 	encodedSpaceURL := fmt.Sprintf("http://localhost:9867/space/%s", url.PathEscape(rawSpace))
 
 	createSpace(t, c, rawSpace, "")
@@ -235,6 +234,12 @@ func TestRequestID(t *testing.T) {
 	})
 }
 
+func nopZeekLauncher() zeek.Launcher {
+	return func(context.Context, io.Reader, string) (zeek.Process, error) {
+		return nil, nil
+	}
+}
+
 func execSearch(t *testing.T, c *zqd.Core, space, prog string) string {
 	parsed, err := zql.ParseProc(prog)
 	require.NoError(t, err)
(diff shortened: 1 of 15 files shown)


zqd/handlers_test.go
```diff
@@ -189,7 +188,7 @@ func TestURLEncodingSupport(t *testing.T) {
 	c := newCore(t)
 	defer os.RemoveAll(c.Root)
 
-	rawSpace := "raw %<>space.brim"
+	rawSpace := "raw %space.brim"
```
mattnibs: Characters `<>` are not allowed for windows filenames.

mattnibs:
> With this change, it would be impossible to run zqd w/o a zeek executable available in path or specified via a cli option. I think we should have some way of doing so, even if that puts out a warning.

Whoops. That was not my intention. If `exec.Lookup` yielded a not found error I was planning to have it start anyways. It appears I made a mistake there.

mattnibs:
> I don't think it's a good idea generally to have api routes that are only enabled in certain scenarios; the errors seen when hitting these will be 404's, which aren't as useful in bug reports as "ingest failed: zeek not found".

That's fine. I'll just have it report a 500 error "zeek not found". I don't think it should even attempt ingest.

alfred-landrum:
> Whoops. That was not my intention. If `exec.Lookup` yielded a not found error I was planning to have it start anyways. It appears I made a mistake there.

oh - I may have misread the code, sorry if so. It sounds like we agree on the intention to let zqd run w/o a zeek present.

mattnibs:
> oh - I may have misread the code, sorry if so. It sounds like we agree on the intention to let zqd run w/o a zeek present.

I have a bug in the code I need to fix.

siskojr:
Thanks, great effort to try get zeek times in Brim close to CLI ones.

zqd/handlers_zeek_test.go
```diff
@@ -1,11 +1,11 @@
-// +build zeek
-
```
nwt: With the changes here, we're no longer testing zeek, right? That seems like a big step backward to me.

zqd/zeek/zeek.go
```diff
@@ -0,0 +1,91 @@
...
+type Process interface {
+	// Wait waits for a running process to exit, returning any errors that
+	// occur.
+	Wait() error
+}
+
+// Launcher is a function when fed a context, pcap reader stream, and a zeek
+// log output dir, will return a running zeek process. If there is an error
+// starting the Process, that error will be returned.
+type Launcher func(context.Context, io.Reader, string) (Process, error)
```
nwt: Nit: No need for the Process interface if you change `type process` to `type Process` and change Launcher to return `(*Process, error)`.

zqd/zeek/zeek.go
```diff
@@ -0,0 +1,91 @@
...
+
+// LauncherFromPath returns a Launcher instance that will initiate start zeek
```
nwt:
```suggestion
// LauncherFromPath returns a Launcher instance that will start a zeek
```

zqd/zeek/zeek.go
```diff
@@ -0,0 +1,91 @@
...
+	// Capture stderr for error reporting.
+	cmd.Stderr = bytes.NewBuffer(nil)
+	p := &process{cmd: cmd}
+	return p
+}
+
+func (p *process) wrapError(err error) error {
+	var exitErr *exec.ExitError
+	if errors.As(err, &exitErr) {
+		stderr := p.cmd.Stderr.(*bytes.Buffer).String()
```
nwt: Nit: Instead of a type assertion, add a `stderrBuf bytes.Buffer` field to `type process` and use it here.

zqd/zeek/zeek.go
```diff
@@ -0,0 +1,91 @@
...
+		p := newExecProcess(ctx, r, zeekpath, dir)
+		return p, p.start()
+	}, nil
+}
+
+type process struct {
+	cmd *exec.Cmd
+}
+
+func newExecProcess(ctx context.Context, pcap io.Reader, zeekpath, outdir string) *process {
```
nwt:
```suggestion
func newProcess(ctx context.Context, pcap io.Reader, zeekpath, outdir string) *process {
```
since this returns a process, not an execProcess.

zqd/zeek/zeek.go
mattnibs: It would not allow this to be testable then and would defeat the purpose of this pr.

zqd/handlers_zeek_test.go
mattnibs: We are testing the functionality surrounding zeek by passing a `testProcess` into `core` that simulates what a zeek process would do. This is better for unit tests, which these endpoint tests are. I agree that we will need some kind of integration/system test w/ live zeek, but even still its hard to replicate some of the error cases we can do here with a mocked zeek process.

zqd/zeek/zeek.go
mattnibs: I'm down with that.

alfred-landrum:
I discussed with @mattnibs ; he's going to add back the zeek binary download, and we'll retain at least one test that uses the actual zeek binary to verify the behavior here.

FYI, we do have integration tests in Brim that verify the full round trip of using packet ingest, with a real zeek binary, via the Brim UI, so we do have integration test coverage there as well.

PR #434, opened 2020-03-19
Add github action workflow for windows testing

mattnibs:
Make pkg/test.Exec work with windows. Remove the reference to bash
and call the command directly. Make sure the path arg to exec.Command
is a path, not just a command name, otherwise we'll trigger the usage
of exec.LookPath on windows, which expects to use the PATH in the
current process environment, and can't be overridden without an os.Setenv.

Make it so shell script system tests only build/run on linux and mac
hosts.


pkg/test/exec.go
```diff
@@ -14,11 +15,16 @@ type Exec struct {
...
-		src = "PATH=" + path + ":$PATH " + src
+		abspath, err := filepath.Abs(path)
+		if err != nil {
+			return "", err
+		}
+		args[0] = filepath.Join(abspath, args[0])
 	}
-	cmd := exec.Command("/bin/bash", "-c", src)
+
+	cmd := exec.Command(args[0], args[1:]...)
```
nwt: If you'd like to stay closer to the original behavior, you could do this.
```go
	if path != "" {
		savedpath := os.Getenv["PATH"]
		os.Setenv("PATH", path+os.PathListSeparator+savedpath)
		defer os.Setenv("PATH", savedpath)
	}
	args := strings.Fields(e.Command)
	cmd := exec.Command(args[0], args[1:]...)
```

tests/system_test.go
```diff
@@ -41,36 +39,3 @@ func TestCommands(t *testing.T) {
 		})
 	}
 }
-
-func TestScripts(t *testing.T) {
-	t.Parallel()
```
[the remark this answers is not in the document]
mattnibs: i like this better. I also like typing in the word "GOOS".

.github/workflows/windows.yaml
```diff
@@ -0,0 +1,19 @@
+name: "windows-testing"
+
+on:
+  push:
+    branches:
+      - "windows-testing"
```
alfred-landrum: When you're ready to merge, would you make sure the `on` triggers match those in the existing ci.yaml?

nwt: If that's the goal, should this be another job in ci.yaml rather than a separate workflow?

alfred-landrum: 👍 good idea!

mattnibs: ok

mattnibs: I won't merge this pr until #432 goes in.

pkg/test/exec.go
mattnibs: I kind of like using an explicit path in the arg over relying on $PATH. As mccanne might say "Chocolate vs Vanilla".

nwt: 👍

alfred-landrum:
oops: in an attempt to get the zq release out this am, I hit the 'squash & merge' button here, w/o realizing it was going to merge to `zeek-process` and not `master`.

PR #436, opened 2020-03-19
Allow expressions to evaluate to complex types

aswan:
toZngValue() wasn't handling non-primitive types which means that
an expression that evaluates to an array, set, or record would fail.
Expressions can't yet create these types, but they can read existing
values so this should work.


aswan:
> Handle zng.TypeUnion too?

Unions don't currently work anywhere inside expressions :(  There's a separate issue open to handle them, I'd prefer to handle them all at once in that issue...

PR #437, opened 2020-03-19
Put proc

aswan:
Not yet ready for merging but ready for review

### DIFF
--- a/proc/put.go
+++ b/proc/put.go
@@ -111,8 +111,12 @@ func (p *Put) put(in *zng.Record) *zng.Record {
 
 	out, err := zng.NewRecord(info.outType, bytes)
 	if err != nil {
-		// ??
-		return in
+		// NewRecord fails if the descriptor has a ts field but
+		// the value can't be extracted or parsed.  Since the input
+		// record had to be valid for us to get into this proc, this
+		// would only happen if a bug in the logic above produced
+		// an invalid record representation.
+		panic(err)
 	}
 	return out
 }
--- a/proc/put.go
+++ b/proc/put.go
@@ -1,0 +1,136 @@
+package proc
+
+import (
+	"github.com/brimsec/zq/ast"
+	"github.com/brimsec/zq/expr"
+	"github.com/brimsec/zq/zbuf"
+	"github.com/brimsec/zq/zcode"
+	"github.com/brimsec/zq/zng"
+)
+
+// The cached information we keep for generating an output record.
+// For a given input descriptor + computed type for the put expression
+// includes the descriptor for output records (outType) plus information
+// about where (position) and how (container) to write the computed value
+// into the output record.
+type descinfo struct {
+	outType   *zng.TypeRecord
+	valType   zng.Type
+	position  int
+	container bool
+}
+
+type Put struct {
+	Base
+	target string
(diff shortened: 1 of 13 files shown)


aswan:
Note about testing, the new tests added here are meant to exercise the different code paths in the put proc but not the full range of zql expressions, those are covered in `expr/expr_test.go`

zql/docs/processors/README.md
```diff
@@ -130,6 +130,41 @@ conn  1521911720.607695 CpjMvj2Cvj048u6bF1 10.164.94.120 39169     10.47.3.200 8
 
 ---
 
+## `put`
+
+|                           |                                                 |
+| ------------------------- | ----------------------------------------------- |
+| **Description**           | Add/update fields based on the results of a computed expression |
+| **Syntax**                | `put <field> = <expression>`                    |
+| **Required arguments**    | `<field>` Field into which the computed value will be stored.<br>`<expression>`A valid ZQL expression (XXX citation needed) |
```
philrz:
```suggestion
| **Required arguments**    | `<field>` Field into which the computed value will be stored.<br>`<expression>` A valid ZQL expression (XXX citation needed) |
```
Micro-nit: Add a space char. 😄

proc/put.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+func (p *Put) put(in *zng.Record) *zng.Record {
+	val, err := p.eval(in)
+	if err != nil {
+		return in
+	}
+
+	// Figure out the output descriptor.  We cache it in outmap
+	// but if we haven't seen this input descriptor before or if
```
henridf: What's an example of where the computed type of the expression changes for a given input descriptor? I would have thought that for a given input descriptor, we always get the same output descriptor...

proc/put.go
```diff
@@ -0,0 +1,136 @@
...
+				bytes = zcode.AppendContainer(bytes, item)
+			} else {
+				bytes = zcode.AppendPrimitive(bytes, item)
+			}
+		}
+	}
+
+	out, err := zng.NewRecord(info.outType, bytes)
+	if err != nil {
+		// ??
```
henridf: this comment is a bit cryptic :)

proc/put.go
aswan: A union would be one example -- unions don't work right now, but when they do I don't expect expressions to propagate the union-ness so the expression `x+1` would evaluate to an integer if the concrete value inside `x` is an integer and similarly for float.  Or when we have conditional expressions, `a ? b : c` or `if(a, b, c)` we can't determine at compile time if `b` and `c` are the same type (since that depends on the types in the underlying record).

proc/put.go
aswan: Oh whoops, will update.

proc/put.go
henridf: Ah ok. I guess when that time comes we'll probably want to keep those different output types cached somehow rather than recompute them (for unions, there would be one output type per union concrete type). Which of course isn't for now...

aswan: Yeah.  fwiw, I think the examples above should be uncommon cases so it is reasonable that they aren't handled as efficiently as possible, its more important that the results are correct even for a strange query.

henridf: True...

PR #439, opened 2020-03-19
Improve pcap ingest sort

nwt:
Take advantage of the fact that Zeek logs are mostly sorted by increasing timestamp to sort fewer records.  Do that by
1. splitting each Zeek log file into two files — one for records already sorted by increasing timestamp (and can be written without sorting) and one for records that must be sorted before writing,
2. creating readers that read these new, sorted files in reverse, and
3. combining the records from these readers into all.bzng.


nwt:
I don't think this is what we're looking for.

PR #440, opened 2020-03-19
Derive space time span from pcap index

mattnibs:
Add a Span() func to pcap.Index that computes the total span of the
index. Use this functionality to report the min/max time of a space
ingesting a packet.

### DIFF
--- a/zqd/packet/packet.go
+++ b/zqd/packet/packet.go
@@ -19,7 +19,6 @@ import (
 	"github.com/brimsec/zq/scanner"
 	"github.com/brimsec/zq/zbuf"
 	"github.com/brimsec/zq/zio/bzngio"
-	"github.com/brimsec/zq/zng"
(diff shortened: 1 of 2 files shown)


mattnibs:
Ran this against brim master and it looks great.

zqd/packet/packet.go
```diff
@@ -271,10 +276,8 @@ func (p *IngestProcess) createSnapshot(ctx context.Context) error {
 		return err
 	}
 	zw := bzngio.NewWriter(bzngfile)
-	program := fmt.Sprintf("sort -limit %d -r ts | (filter *; head 1; tail 1)", p.sortLimit)
-	var headW, tailW recWriter
```
henridf: remove recWriter?

mattnibs: Oh yeah, good point. Thanks!

PR #441, opened 2020-03-20
Typed NDJSON reader

henridf:
This commit introduces a typed ndjson reader. This reader uses side-information describing the zng types of different ndjson records in order to decode ndjson into rich zng types rather than into the subset of zng types that can be inferred from json.

This type side-information is stored in a json file that is presented to the zq tool using the `-j` flag. For example:
```sh
zq -j types.json conn.ndjson
```
If no `-j` flag is present, ndjson input is read via the existing mechanism that projects json types onto the subset of corresponding zng types.

### DIFF
--- a/cmd/zq/zq.go
+++ b/cmd/zq/zq.go
@@ -1,8 +1,11 @@
 package main
 
 import (
+	"encoding/json"
+	"errors"
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"os"
 
 	"github.com/brimsec/zq/ast"
@@ -14,6 +17,7 @@ import (
 	"github.com/brimsec/zq/zbuf"
 	"github.com/brimsec/zq/zio"
 	"github.com/brimsec/zq/zio/detector"
+	"github.com/brimsec/zq/zio/ndjsonio"
 	"github.com/brimsec/zq/zng/resolver"
 	"github.com/brimsec/zq/zql"
 	"github.com/mccanne/charm"
@@ -71,16 +75,17 @@ func init() {
 }
 
 type Command struct {
-	zctx        *resolver.Context
-	ifmt        string
-	ofmt        string
-	dir         string
-	path        string
-	outputFile  string
-	verbose     bool
-	stats       bool
-	quiet       bool
-	showVersion bool
+	zctx         *resolver.Context
+	ifmt         string
+	ofmt         string
+	dir          string
+	path         string
+	jsonTypePath string
+	outputFile   string
+	verbose      bool
+	stats        bool
+	quiet        bool
+	showVersion  bool
 	zio.Flags
 }
 
@@ -92,6 +97,7 @@ func New(f *flag.FlagSet) (charm.Command, error) {
 	f.StringVar(&c.path, "p", cwd, "path for input")
 	f.StringVar(&c.dir, "d", "", "directory for output data files")
 	f.StringVar(&c.outputFile, "o", "", "write data to output file")
+	f.StringVar(&c.jsonTypePath, "j", "", "path to json types file")
 	f.BoolVar(&c.verbose, "v", false, "show verbose details")
 	f.BoolVar(&c.stats, "S", false, "display search stats on stderr")
 	f.BoolVar(&c.quiet, "q", false, "don't display zql warnings")
@@ -159,6 +165,19 @@ func (c *Command) printVersion() error {
 	return nil
 }
 
+func (c *Command) loadJsonTypes() (*ndjsonio.TypeConfig, error) {
+	data, err := ioutil.ReadFile(c.jsonTypePath)
+	if err != nil {
+		return nil, err
+	}
+	var tc ndjsonio.TypeConfig
+	err = json.Unmarshal(data, &tc)
+	if err != nil {
+		return nil, fmt.Errorf("%s: unmarshaling error: %s", c.jsonTypePath, err)
+	}
+	return &tc, nil
+}
+
 func (c *Command) Run(args []string) error {
 	if c.showVersion {
 		return c.printVersion()
@@ -193,9 +212,30 @@ func (c *Command) Run(args []string) error {
 		c.ofmt = "null"
 		defer logger.Close()
 	}
-	var reader zbuf.Reader
-	if reader, err = c.loadFiles(paths); err != nil {
+	var tc *ndjsonio.TypeConfig
+	if c.jsonTypePath != "" {
+		if tc, err = c.loadJsonTypes(); err != nil {
+			return err
+		}
+	}
+
+	var readers []zbuf.Reader
+	if readers, err = c.inputReaders(paths); err != nil {
 		return err
+	}
+	for _, r := range readers {
+		if ndjr, ok := r.(*ndjsonio.Reader); ok && (tc != nil) {
+			if err = ndjr.SetTypeConfig(*tc); err != nil {
+				return err
+			}
+		}
+	}
+
+	var reader zbuf.Reader
+	if len(readers) == 1 {
+		reader = readers[0]
+	} else {
+		reader = scanner.NewCombiner(readers)
 	}
 	writer, err := c.openOutput()
 	if err != nil {
@@ -217,34 +257,41 @@ func (c *Command) errorf(format string, args ...interface{}) {
 	_, _ = fmt.Fprintf(os.Stderr, format, args...)
 }
 
-func (c *Command) loadFiles(paths []string) (zbuf.Reader, error) {
+func (c *Command) inputReaders(paths []string) ([]zbuf.Reader, error) {
 	var readers []zbuf.Reader
 	for _, path := range paths {
 		var zr zbuf.Reader
+		var f *os.File
 		if path == "-" {
-			r := detector.GzipReader(os.Stdin)
+			f = os.Stdin
+		} else {
 			var err error
-			if c.ifmt == "auto" {
-				zr, err = detector.NewReader(r, c.zctx)
-			} else {
-				zr, err = detector.LookupReader(c.ifmt, r, c.zctx)
-			}
+			info, err := os.Stat(path)
 			if err != nil {
 				return nil, err
 			}
-		} else {
-			var err error
-			zr, err = scanner.OpenFile(c.zctx, path, c.ifmt)
+			if info.IsDir() {
+				return nil, errors.New("is a directory")
+			}
+			f, err = os.Open(path)
 			if err != nil {
 				return nil, err
 			}
 		}
+		r := detector.GzipReader(f)
+		var err error
+		if c.ifmt == "auto" {
+			zr, err = detector.NewReader(r, c.zctx)
+		} else {
+			zr, err = detector.LookupReader(c.ifmt, r, c.zctx)
+		}
+		if err != nil {
+			return nil, err
+		}
+
 		readers = append(readers, zr)
 	}
-	if len(readers) == 1 {
-		return readers[0], nil
-	}
-	return scanner.NewCombiner(readers), nil
+	return readers, nil
 }
 
 func (c *Command) openOutput() (zbuf.WriteCloser, error) {
--- a/cmd/zq/zq.go
+++ b/cmd/zq/zq.go
@@ -212,21 +212,21 @@ func (c *Command) Run(args []string) error {
 		c.ofmt = "null"
 		defer logger.Close()
 	}
-	var tc *ndjsonio.TypeConfig
-	if c.jsonTypePath != "" {
-		if tc, err = c.loadJsonTypes(); err != nil {
-			return err
-		}
-	}
 
-	var readers []zbuf.Reader
-	if readers, err = c.inputReaders(paths); err != nil {
+	readers, err := c.inputReaders(paths)
+	if err != nil {
 		return err
 	}
-	for _, r := range readers {
-		if ndjr, ok := r.(*ndjsonio.Reader); ok && (tc != nil) {
-			if err = ndjr.SetTypeConfig(*tc); err != nil {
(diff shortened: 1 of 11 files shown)


henridf:
(More tests are going to be added to this PR)

zio/ndjsonio/reader.go
```diff
@@ -6,28 +10,124 @@ import (
...
+		tr.descriptors[key] = recType
+	}
+	r.typ = &typeParser{zctx: r.zctx, tr: tr, stats: r.stats.typeStats}
+	return nil
+}
+
+// Parse returns a zng.Encoding slice as well as an inferred zng.Type
+// from the provided JSON input. The function expects the input json to be an
+// object, otherwise an error is returned.
+func (r *Reader) Parse(b []byte) (zcode.Bytes, zng.Type, error) {
```
aswan: nit: signature here would be a little simpler if this returned a zng.Value

zio/ndjsonio/typeparser.go
```diff
@@ -0,0 +1,326 @@
...
+		if len(r.Match) != 1 {
+			// This will turn into a panic when we do validation upon loading json types
+			return nil, fmt.Errorf("Rule %v with length %d", r, len(r.Match))
+		}
+
+		var name, val string
+		for n, v := range r.Match {
+			name = n
+			val = v
+		}
```
aswan: Can we make the Rule struct just have Name and Value members instead of the Match map?
The types.json structure could be changed similarly or it could be kept as it currently is and the conversion could be done in an UnmarshalJSON() method on Rule.
(this could happen in a follow-up as it overlaps with validation which you've explicitly mentioned is planned for a follow-up in the comment here)

zio/ndjsonio/typeparser.go
```diff
@@ -0,0 +1,326 @@
...
+func lookupTypeInfo(zctx *resolver.Context, desc *zng.TypeRecord, path string) *typeInfo {
+	flatCols := zeekio.FlattenColumns(desc.Columns)
+	flatDesc := zctx.LookupTypeRecord(flatCols)
+	info := typeInfo{desc, flatDesc, []byte(path)}
+	return &info
+}
+
+// findTypeInfo returns the typeInfo struct matching an input json
+// object.  If no match is found, an error is returned. If defaultPath
+// is non-zero, it is used as the default _path if the object has no
```
aswan:
```suggestion
// is not empty, it is used as the default _path if the object has no
```

zio/ndjsonio/typeparser.go
```diff
@@ -0,0 +1,326 @@
...
+				// "b"}', "a") returns "b". This is ok because
+				// these errors will later be caught by ObjectEach.
+				fieldVal, err = jsonparser.GetUnsafeString(jobj, name)
+			}
+			if err != nil {
+				continue
+			}
+		}
+		if val == fieldVal {
+			return lookupTypeInfo(zctx, tr.descriptors[r.Descriptor], path), nil
```
aswan: Can we cache this info in the descriptor?
Ah, I see there's a comment about this being costly below.  Follow-up optimization?

zio/ndjsonio/typeparser.go
```diff
@@ -0,0 +1,326 @@
...
+// the underlying JSON values.  This slice follows the order of the flattened
+// columns.  Second, it builds the full encoded value and building nested
+// records as necessary.
+func (info *typeInfo) newRawFromJSON(data []byte) (zcode.Bytes, int, error) {
+	var droppedFields int
+	type jsonVal struct {
+		val []byte
+		typ jsonparser.ValueType
+	}
+	jsonVals := make([]jsonVal, 32) // Fixed size for stack allocation.
```
aswan: How about just allocating a fixed array inside the typeinfo?

zio/ndjsonio/typeparser.go
henridf: Good idea. I'll just do it on this PR.

zio/ndjsonio/typeparser.go
henridf: Sure but what's the benefit? I think we'd still have to have the `len(..) > N` check here.

aswan: When typeinfo is created we know how many total columns there are and could create a single array of the needed size.

zio/ndjsonio/typeparser.go
henridf: done

zio/ndjsonio/reader.go
henridf: this is pre-existing but agreed; will change

henridf: done

zio/ndjsonio/typeparser.go
henridf: oh of course! will do

henridf: done

zio/ndjsonio/typeparser.go
henridf: oops, no i left the comment but wanted to do it in this PR... coming

zio/ndjsonio/ndjson_test.go
```diff
@@ -154,3 +156,47 @@ func NDJSONEq(t *testing.T, expected string, actual string) {
 		require.JSONEq(t, expectedLines[i], actualLines[i])
 	}
 }
+func TestNewRawFromJSON(t *testing.T) {
```
nwt:
```suggestion

func TestNewRawFromJSON(t *testing.T) {
```

cmd/zq/zq.go
```diff
@@ -193,10 +212,31 @@ func (c *Command) Run(args []string) error {
...
+	if readers, err = c.inputReaders(paths); err != nil {
 		return err
 	}
+	for _, r := range readers {
+		if ndjr, ok := r.(*ndjsonio.Reader); ok && (tc != nil) {
+			if err = ndjr.SetTypeConfig(*tc); err != nil {
+				return err
+			}
+		}
+	}
```
nwt: Nit: This might be clearer.
```suggestion
	readers, err := c.inputReaders(paths)
	if err != nil {
		return err
	}
	if c.jsonTypePath != "" {
		tc, err := c.loadJsonTypes()
		if err != nil {
			return err
		}
		for _, r := range readers {
			if ndjr, ok := r.(*ndjsonio.Reader); ok {
				if err := ndjr.SetTypeConfig(*tc); err != nil {
					return err
				}
			}
		}
	}
```

zio/ndjsonio/reader.go
```diff
@@ -6,28 +10,124 @@ import (
...
+			return zng.Value{}, err
+		}
+		return zv, nil
+	}
+
+	zv, err := r.inf.parseObject(val)
+	if err != nil {
+		return zng.Value{}, err
 	}
+	return zv, nil
```
nwt: Nit: Shorten to
```suggestion
	if r.typ != nil {
		return r.typ.parseObject(val)
	}
	return r.inf.parseObject(val)
```

zio/ndjsonio/typeparser.go
```diff
@@ -0,0 +1,315 @@
...
+	switch typ {
+	case zng.TypeTime:
+		ts, err := parseJSONTimestamp(value)
+		if err != nil {
+			return nil, err
+		}
+		return zng.EncodeTime(ts), nil
+	case zng.TypeDuration:
+		// cannot use nano.Parse because javascript floats values can have
+		// greater precision than 10e-9.
```
nwt: 10e-9 or 1e-9?

zio/ndjsonio/typeparser.go
```diff
@@ -0,0 +1,315 @@
...
+}
+
+func parseSimpleType(value []byte, typ zng.Type) ([]byte, error) {
+	switch typ {
+	case zng.TypeTime:
+		ts, err := parseJSONTimestamp(value)
+		if err != nil {
+			return nil, err
+		}
+		return zng.EncodeTime(ts), nil
```
nwt: Nit: You can shorten all of these cases by leaving out the if.
```suggestion
		ts, err := parseJSONTimestamp(value)
		return zng.EncodeTime(ts), nil
```

zio/ndjsonio/typeparser.go
henridf: done

cmd/zq/zq.go
henridf: it sure is. thanks!

zio/ndjsonio/reader.go
henridf: ha!

zio/ndjsonio/typeparser.go
henridf: 1e-9 i think

zio/ndjsonio/typeparser.go
henridf: @nwt what am i missing here... the suggested change ignores the err returned by parseJSONTimestamp?

nwt: Doh! It sure does. Here's what I meant to suggest.
```suggestion
		ts, err := parseJSONTimestamp(value)
		return zng.EncodeTime(ts), err
```

henridf: Ah ok. I like the reduction in lines, and it would work just fine, but I think it's a bit clearer the current way, as opposed to calling a subsequent function including with an invalid value.

zio/ndjsonio/reader.go
```diff
@@ -6,28 +10,122 @@ import (
...
+	}
+	if typ != jsonparser.Object {
+		return zng.Value{}, fmt.Errorf("expected JSON type to be Object but got %s", typ)
+	}
+	if r.typ != nil {
+		if zv, err := r.typ.parseObject(val); err == nil {
+			return zv, nil
+		}
 	}
+	return r.inf.parseObject(val)
```
aswan: Is quietly falling back to inference the right default?  If there's a typo or something in types.json, seems like it would be tricky to debug that.  Though in practice, I think its unlikely that anybody will use anything other than the default zeek types.json any time soon so this is probably fine for now.

zio/ndjsonio/typeparser.go
```diff
@@ -0,0 +1,322 @@
...
+		// This is always a KeyPathNotFoundError, including if the json was invalid.
+		if defaultValue == "" {
+			return "", jsonparser.KeyPathNotFoundError
+		}
+		return defaultValue, nil
+	}
+	return val, nil
+}
+
+func lookupTypeInfo(zctx *resolver.Context, desc *zng.TypeRecord, path string) *typeInfo {
```
aswan: nit: this is doing more than a lookup, maybe call it makeTypeInfo or newTypeInfo?

zio/ndjsonio/typeparser.go
```diff
@@ -0,0 +1,322 @@
...
+	colno := 0
+	nestedColno := 0
+	nestedColumns := info.descriptor.Columns
+	flatColumns := info.flatDesc.Columns
+	for i := range flatColumns {
+		val := info.jsonVals[i].val
+		if i == info.descriptor.TsCol {
+			if info.jsonVals[i].typ != jsonparser.String && info.jsonVals[i].typ != jsonparser.Number {
+				return nil, 0, fmt.Errorf("invalid json type for ts: %s", info.jsonVals[i].typ)
+			}
```
aswan: nit: seems like this should be inside the function parseJSONTimestamp

zio/ndjsonio/typeparser.go
```diff
@@ -0,0 +1,322 @@
...
+		}
+	}
+	if path == "" {
+		return nil, ErrMissingPath
+	}
+	return nil, ErrDescriptorNotFound
+}
+
+func (p *typeParser) parseObject(b []byte) (zng.Value, error) {
+
```
aswan: unnecessary blank line

zio/ndjsonio/reader.go
henridf: I don't know, and I think we'll only learn with experience. When in doubt I've leaned towards @philrz so if he thinks we should flip this around right away I'll do it. @philrz ?

zio/ndjsonio/typeparser.go
henridf: 👍

zio/ndjsonio/typeparser.go
henridf: meaning it would take a `jsonVal` rather than a `[]byte`? I'm not sure, I like the idea of parseJSONTimestamp being agnostic to the json typing stuff (even if it's only used here right now)

aswan: Feels to me like it splits the logic of "what is a valid representation of a timestamp in json" between here and parseJSONTimestamp but if you prefer it this way, I don't feel strongly.

zio/ndjsonio/reader.go
henridf: Update: after conversing with @philrz I'm going to flip this around as per @aswan 's suggestion. Not the final word, but it's what we had in boom and seems reasonable. Thanks @aswan for raising this.

philrz: Here's additional color. Indeed, in our previous attempt at this in other software, we'd been assuming we'd want to guide the user toward a conservative bootstrapping phase. If we provide them a schema for a modern out-of-the-box Zeek environment and it works for 100% of their logs, great! If they've added customizations or additional plugins/scripts that produce additional fields/types our schema is not prepared for, we figure it's best to push them toward enduring the pain up-front to get the schema correct and "lock it in", then be notified if some later customization causes it to fail again. This is better than if some IP address or ISO timestamp gets inferred to a string and weeks later they're tearing their hair out wondering why the queries aren't matching.

PR #442, opened 2020-03-20
catch and log panics during api requests

alfred-landrum:
I don't see an easy way to test this at the moment without introducing more substantial changes, so would like to do that after the release.

### DIFF
--- a/zqd/middleware.go
+++ b/zqd/middleware.go
@@ -2,6 +2,7 @@ package zqd
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"strconv"
 	"sync/atomic"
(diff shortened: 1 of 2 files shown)


zqd/middleware.go
```diff
@@ -67,6 +68,25 @@ func accessLogMiddleware(logger *zap.Logger) mux.MiddlewareFunc {
...
+	logger = logger.Named("zqd")
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			defer func() {
+				r := recover()
+				if r == nil {
+					return
+				}
+				rstr := fmt.Sprint(r)
+				logger.Error("panic", zap.String("error", rstr))
```
mattnibs: I would do a `logger.DPanic` here.

zqd/middleware.go
```diff
@@ -67,6 +68,25 @@ func accessLogMiddleware(logger *zap.Logger) mux.MiddlewareFunc {
 	}
 }
 
+func panicCatchMiddleware(logger *zap.Logger) mux.MiddlewareFunc {
+	logger = logger.Named("zqd")
```
mattnibs: We need better facilities for this, but this logger should also use the `request-ID` value attached to context.

PR #447, opened 2020-03-20
Zx cleanup

aswan:
This is a baby step toward unifying filters and expressions.  Move existing code out of zx/ into more appropriate directories and clean up the type coercion interface for integers.

### DIFF
--- a/expr/expr.go
+++ b/expr/expr.go
@@ -10,6 +10,7 @@ import (
 	"github.com/brimsec/zq/pkg/nano"
 	"github.com/brimsec/zq/zcode"
 	"github.com/brimsec/zq/zng"
+	"github.com/brimsec/zq/zngnative"
 )
 
 type ExpressionEvaluator func(*zng.Record) (zng.Value, error)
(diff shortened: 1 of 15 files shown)


nwt:
@aswan: The zng package already has a lot a surface area. I think that was @mccanne's motivation for creating the zx package. How about keeping it and moving NativeType there?

aswan:
> @aswan: The zng package already has a lot a surface area. I think that was @mccanne's motivation for creating the zx package. How about keeping it and moving NativeType there?

Why is surface area a problem per se?  The code that handles native types and type coercion is intimately coupled to the details of the zng type system and is unusable without it.  Its not clear to me that two tightly coupled packages is somehow better than a single package with a wider public interface.

nwt:
> Why is surface area a problem per se?

It isn't a problem, just a matter of taste. My take is that there's already too much in the zng package, making it harder to comprehend than it should be.

> The code that handles native types and type coercion is intimately coupled to the details of the zng type system and is unusable without it.

The same is true of much of the code in this repository.

> Its not clear to me that two tightly coupled packages is somehow better than a single package with a wider public interface.

That is, I think, very much a matter of taste. I think the repository's a slightly better place with this stuff in zx, as presumably did @mccanne when he moved it from its original home in zng (19d8be4f357fcc640a31a130a98330e34f4c16c1). If you remain unswayed, that's fine, and you should feel very free to merge this. I will not be cranky about it. 😃

aswan:
Well, this (small number of big packages or large number of small packages) seems like something we should have group consensus on.  My 2c is that favorinng small packages leads to the nonsense that is npm (how in the world is https://www.npmjs.com/package/is-odd on version 3.0.1).  Realistically, I don't think we would do anything that absurd but I think the advantage of being able to absorb and understand the full contents of a small package ends up being offset between the need to know about a larger number of packages and the relationships between them, and potentially running into circular import issues.

That's just my 2c though, I'm happy going with the group's preference if it is different (though if we like small packages, I'll want to rename zx to something more descriptive)

@henridf, @mattnibs, @mccanne, @alfred-landrum, any opinions?

aswan:
Here's a revised version in which zx is renamed to zngnative and NativeType moves there

henridf:
lgtm (as discussed offline i have some reservations about zngnative but no slam dunk alternative)

PR #451, opened 2020-03-23
make space deletion cancel and await other operations

alfred-landrum:
When a space delete API call is made, signal other active operations to
cancel and wait for them to do so, preventing further operations
until the deletion is done.

This uses zqd.Core as the holder of the necessary state. Eventually, I
think this would better live in a dedicated space manager, but that's a
much larger change, and I think this code will move there easily.

This is a part of addressing ingest cancellation during application
shutdown on Windows, as we currently see "file in use" errors there when
a delete is called concurrently with ingest.

### DIFF
--- a/cmd/pcap/slice/command.go
+++ b/cmd/pcap/slice/command.go
@@ -2,6 +2,7 @@ package slice
 
 import (
 	"bufio"
+	"context"
 	"errors"
 	"flag"
 	"fmt"
@@ -166,5 +167,5 @@ func (c *Command) Run(args []string) error {
 	} else {
 		search = pcap.NewRangeSearch(span)
 	}
-	return search.Run(out, pcapReader)
+	return search.Run(context.TODO(), out, pcapReader)
 }
--- a/zqd/core.go
+++ b/zqd/core.go
@@ -1,7 +1,9 @@
 package zqd
 
 import (
+	"context"
 	"net/http"
+	"sync"
 	"sync/atomic"
 
 	"github.com/brimsec/zq/zqd/zeek"
@@ -36,6 +38,19 @@ type Core struct {
 	SortLimit int
 	taskCount int64
 	logger    *zap.Logger
+
+	// ingestLock protects the ingests map and the deletePending
+	// field inside the ingestWaitState's.
+	ingestLock sync.Mutex
+	ingests    map[string]*ingestWaitState
+}
+
+type ingestWaitState struct {
+	deletePending int
+	// WaitGroup for active ingests
+	wg sync.WaitGroup
+	// closed to signal active ingest should terminate
+	cancelChan chan struct{}
 }
 
 func NewCore(conf Config) *Core {
@@ -48,6 +63,7 @@ func NewCore(conf Config) *Core {
 		ZeekLauncher: conf.ZeekLauncher,
 		SortLimit:    conf.SortLimit,
 		logger:       logger,
+		ingests:      make(map[string]*ingestWaitState),
 	}
 }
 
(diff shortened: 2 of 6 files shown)


mattnibs:
I imagine you will also run into "file in use" errors if a delete is issued during a search or even a space info request. That said, I'm not against adding this for now pending the larger space.Manager refactor.

cmd/pcap/slice/command.go
```diff
@@ -166,5 +167,5 @@ func (c *Command) Run(args []string) error {
 	} else {
 		search = pcap.NewRangeSearch(span)
 	}
-	return search.Run(out, pcapReader)
+	return search.Run(context.TODO(), out, pcapReader)
```
mattnibs, line 170: Note to self: Good usage of `context.TODO()` here.

zqd/core.go
```diff
@@ -36,6 +38,20 @@ type Core struct {
 	SortLimit int
 	taskCount int64
 	logger    *zap.Logger
+
+	// spaceOpsLock protects the spaceOps map and the active and
+	// deletePending fields inside the spaceOpsState's.
+	spaceOpsLock sync.Mutex
```
nwt, line 44: Nit: Call this spaceOpsMu for consistency with existing sync.Mutex/RWMutex names.

zqd/handlers_test.go
```diff
@@ -315,6 +319,65 @@ func TestPostInvalidLogs(t *testing.T) {
...
+	}
+	go func() {
+		packetPostErr <- doPost()
+	}()
+
+	wg.Wait()
+	err = client.SpaceDelete(context.Background(), spaceName)
+	require.NoError(t, err)
+
+	require.NoError(t, <-packetPostErr)
```
mattnibs, line 378: I think something like this would be make this test more clear:
```suggestion
		if taskEnd != nil  {
			return taskEnd.Error
		}
		return nil
	}
	go func() {
		packetPostErr <- doPost()
	}()

	wg.Wait()
	err = client.SpaceDelete(context.Background(), spaceName)
	require.NoError(t, err)

	require.Error(t, <-packetPostErr, "context canceled")
```

zqd/core.go
```diff
@@ -62,3 +79,89 @@ func (c *Core) requestLogger(r *http.Request) *zap.Logger {
 func (c *Core) getTaskID() int64 {
 	return atomic.AddInt64(&c.taskCount, 1)
 }
+
+// startSpaceOp registers that an operation on a space is in progress.
+// If the space is pending deletion, the bool parameter returns false.
+// Otherwise, this returns a new context, and a done function that must
+// be called when the operation completes.
+func (c *Core) startSpaceOp(ctx context.Context, space string) (context.Context, func(), bool) {
```
nwt, line 87:
```suggestion
func (c *Core) startSpaceOp(ctx context.Context, space string) (context.Context, context.CancelFunc, bool) {
```

zqd/handlers.go
```diff
@@ -35,6 +35,14 @@ func handleSearch(c *Core, w http.ResponseWriter, r *http.Request) {
 		http.Error(w, err.Error(), status)
 		return
 	}
+
+	ctx, opDone, ok := c.startSpaceOp(r.Context(), s.Name())
```
nwt, line 39: Nit: opDone is really a context.CancelFunc, so call it cancel.

zqd/core.go
nwt: I take that back. It's more than a CancelFunc, so leave this as it is.

zqd/handlers.go
nwt: I take that back. It's more than a CancelFunc, so leave this as it is.

PR #452, opened 2020-03-23
Add validation for TypeConfig

henridf:


zio/ndjsonio/config.go
```diff
@@ -0,0 +1,61 @@
...
+	for name, desc := range conf.Descriptors {
+		col, ok := desc[0].(map[string]interface{})
+		if !ok {
+			return fmt.Errorf("descriptor %s has invalid structure", name)
+		}
+		if col["name"] != "_path" {
+			return fmt.Errorf("descriptor %s does not have _path as first column", name)
+		}
+		for _, d := range desc {
+			col := d.(map[string]interface{})
```
aswan: Is it possible that this cast fails and panics?

zio/ndjsonio/config.go
```diff
@@ -0,0 +1,61 @@
...
+			if col["name"] == "ts" && col["type"] != "time" {
+				return fmt.Errorf("descriptor %s has field time with wrong type %s", name, col["type"])
```
aswan:
```suggestion
				return fmt.Errorf("descriptor %s has field ts with wrong type %s", name, col["type"])
```

zio/ndjsonio/config.go
henridf: oops... fixed and added tests for this

PR #453, opened 2020-03-23
Add Math.min and Math.max functions

aswan:


expr/functions.go
```diff
@@ -15,9 +15,93 @@ var ErrWrongArgc = errors.New("wrong number of arguments")
...
+	switch args[0].Type.ID() {
+	case zng.IdInt16, zng.IdInt32, zng.IdInt64:
+		ret := args[0].Value.(int64)
+		for _, val := range args[1:] {
+			v, ok := zngnative.CoerceNativeToInt(val)
+			if ok && v > ret {
+				ret = v
+			}
+		}
+		return zngnative.NativeValue{zng.TypeInt64, ret}, nil
```
henridf: (not feedback on this PR, but seeing this makes me think that `zngnative.Value` might work better...)

expr/functions.go
```diff
@@ -15,9 +15,93 @@ var ErrWrongArgc = errors.New("wrong number of arguments")
 var ErrBadArgument = errors.New("bad argument")
 
 var allFns = map[string]Function{
+	"Math.max":  mathMax,
+	"Math.min":  mathMin,
 	"Math.sqrt": mathSqrt,
 }
 
+func mathMax(args []zngnative.NativeValue) (zngnative.NativeValue, error) {
+	if len(args) == 0 {
```
henridf: Could we encode arity elsewhere (for example in `allFns`) and check this at compile time?

zngnative/coerce.go
```diff
@@ -11,48 +11,26 @@ import (
 // resulting coerced value is written to out, and true is returned. If
 // the value cannot be coerced, then false is returned.
 func CoerceToFloat64(in zng.Value) (float64, bool) {
-	var out float64
-	var err error
-	switch in.Type.ID() {
-	default:
+	native, err := ToNativeValue(in)
```
henridf: Are the changes to this file separate cleanup?

expr/functions.go
aswan: that's a great idea, I'll do that in a separate PR

expr/functions.go
aswan: Yes, I had actually done that in a draft of the first functions PR and then decided to switch it this way so that all the argument validation (count, types, etc.) remains in one place.  I'm happy to revisit it but at the moment, don't see a huge advantage for either approach.

zngnative/coerce.go
aswan: Not entirely, the CoerceNativeTo... functions are new, then there's some shuffling of code to implement the existing CoerceTo... functions in terms of their CoerceNativeTo... equivalents.  While I was there I made the float coercion more consistent with integer (eg removed the bool->float coercion which wasn't listed on the type coercion wiki page)

expr/functions.go
henridf: > don't see a huge advantage for either approach

I might have been making an incorrect assumption here: with the way things are currently done, doesn't `fmt.Errorf("Math.max: %w", ErrWrongArgc)` end up as a runtime warning ? If that's the case, then it seems to me like this should fail at compile time since the error is a property of the program regardless of the data. If that's not the case, then I agree there's no major advantage to either approach...

aswan: Oh yes, good point.  I think we still want the runtime warnings to be useful and prominent here but in the short term, I think you're right.  Will update.

henridf: 👍 

> we still want the runtime warnings to be useful and prominent

Out of curiosity, is this in anticipation of variadic args (e.g `Math.max(some_array...)`)?  That was the one situation I could think of where compile-time wouldn't suffice...

aswan: Nothing so fancy.  Just that if you mistype a field name or apply a function to a value of the wrong type, that can't be detected until runtime and it would be a pain to debug such a query without warnings to point toward the problem.

henridf: Ah I see, agree that those should continue to be runtime warnings (I was thinking specifically about arity-related issues).

PR #454, opened 2020-03-23
packet post endpoint: add snapshot counter

mattnibs:
Track the number of snapshots created during packet ingest into a space. 
Every time a snapshot is produced increment the counter. The current snapshot 
count is returned in the api.PacketPostStatus payload.

### DIFF
--- a/zqd/api/api.go
+++ b/zqd/api/api.go
@@ -112,6 +112,7 @@ type PacketPostStatus struct {
 	UpdateTime     nano.Ts  `json:"update_time"`
 	PacketSize     int64    `json:"packet_total_size" unit:"bytes"`
 	PacketReadSize int64    `json:"packet_read_size" unit:"bytes"`
+	SnapshotCount  int      `json:"snapshot_count"`
 	MinTime        *nano.Ts `json:"min_time,omitempty"`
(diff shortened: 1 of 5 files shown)


mccanne:
It would be simpler to keep the counter as a local variable in handlePacketPost.  Bump it at
```
		case <-proc.Snap():
```

mattnibs:
@mccanne this is where I originally had it, but its possible for zeek to finish before the snapshot channel has fired.

mccanne:
> @mccanne this is where I originally had it, but its possible for zeek to finish before the snapshot channel has fired.

What's the problem with that?

mattnibs:
> What's the problem with that?

For starters it means the number isn't necessarily accurate. More importantly this made it so I could easily have test coverage on this feature.

mccanne:
> > What's the problem with that?
> 
> For starters it means the number isn't necessarily accurate. More importantly this made it so I could easily have test coverage on this feature: Even on a very small pcap the snapshot count should be at least one.

Ok, I knew there was a reason why I didn't like the Done/Snap channel design and here is is.  If there were a single channel that sent updates from the ingester to the api handler that was closed to signal done, this would all be a bit simpler and you wouldn't have this problem.

Anyway, not suggesting you change it at this point but it would certainly clean this up.

PR #455, opened 2020-03-24
Wire up _path inference for use with json types

henridf:


henridf:
@philrz given the selectivity of the regex (https://github.com/brimsec/zq/pull/455/commits/6a0710b9cdf18e2dd5ed511e168a9b795aae4df0#diff-3bca4778f483572fe8a11b91416a038fR100-R101), and the fact that any `_path` present in a record overrides the default inferred  from a file name, I'm thinking we may want to enable this by default rather than having to pass `-inferpath`. What do you think?

cmd/zq/zq.go
```diff
@@ -75,29 +77,38 @@ func init() {
...
+	// x509_20191101_14:00:00-15:00:00+0000.log.gz (corelight)
+	c.jsonPathRegexp = `([a-zA-Z0-9_]+)(?:\.|_\d{8}_)\d\d:\d\d:\d\d\-\d\d:\d\d:\d\d(?:[+\-]\d{4})?\.log(?:$|\.gz)`
+
 	f.StringVar(&c.ifmt, "i", "auto", "format of input data [auto,bzng,ndjson,zeek,zjson,zng]")
 	f.StringVar(&c.ofmt, "f", "zng", "format for output data [bzng,ndjson,table,text,types,zeek,zjson,zng]")
 	f.StringVar(&c.path, "p", cwd, "path for input")
 	f.StringVar(&c.dir, "d", "", "directory for output data files")
 	f.StringVar(&c.outputFile, "o", "", "write data to output file")
 	f.StringVar(&c.jsonTypePath, "j", "", "path to json types file")
+	f.BoolVar(&c.jsonInferPath, "inferpath", false, "infer zeek _path from file name (for ndjson objects without _path)")
```
mattnibs: I'm not the end user, so take my input here with a large dalop of salt, but... I would just assume get rid of the `jsonInferPath` option and if a json file does not have appear to have `_path` in its records it tries to get them using the filename.

cmd/zq/zq.go
```diff
@@ -222,9 +236,18 @@ func (c *Command) Run(args []string) error {
...
+				var path string
+				if c.jsonInferPath {
+					re := regexp.MustCompile(c.jsonPathRegexp)
+					match := re.FindStringSubmatch(paths[i])
+					if len(match) != 2 {
+						return fmt.Errorf("could not infer path from filename %s", paths[i])
+					}
+					path = match[1]
+				}
+				if err := ndjr.ConfigureTypes(*tc, path); err != nil {
```
mattnibs: In an effort to keep the commands `Run` function simple I would move this logic into its own function (perhaps something like `configureJSONReader` and do the `ndjsonio.Reader` type coercion in the loop of the function `inputReaders`.

cmd/zq/zq.go
henridf: That sounds reasonable... @philrz what do you say?

philrz: It has been a while since I've paged in my strong opinions about JSON typing, but, yes, that sounds reasonable to me too.

henridf: @philrz 👍 this addresses my question above (https://github.com/brimsec/zq/pull/455#issuecomment-603265371) too.

henridf: (done)

cmd/zq/zq.go
henridf: 👍

cmd/zq/zq.go
```diff
@@ -75,29 +77,38 @@ func init() {
...
 	zio.Flags
 }
 
 func New(f *flag.FlagSet) (charm.Command, error) {
 	cwd, _ := os.Getwd()
 	c := &Command{zctx: resolver.NewContext()}
+
+	// x509.14:00:00-15:00:00.log.gz (open source zeek)
+	// x509_20191101_14:00:00-15:00:00+0000.log.gz (corelight)
+	c.jsonPathRegexp = `([a-zA-Z0-9_]+)(?:\.|_\d{8}_)\d\d:\d\d:\d\d\-\d\d:\d\d:\d\d(?:[+\-]\d{4})?\.log(?:$|\.gz)`
```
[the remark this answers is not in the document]
henridf: done

PR #458, opened 2020-03-24
go api client

mattnibs:
Add go api client to package api. Switch zqd endpoint tests to use
the go api client.

### DIFF
--- a/zqd/api/bzng.go
+++ b/zqd/api/bzng.go
@@ -1,0 +1,64 @@
+package api
+
+import (
+	"bytes"
+	"io"
+
+	"github.com/brimsec/zq/zio/bzngio"
+	"github.com/brimsec/zq/zng"
+	"github.com/brimsec/zq/zng/resolver"
+)
+
+type BzngSearch struct {
+	reader *bzngio.Reader
+	ctrl   interface{}
+	err    error
+	onctrl func(interface{})
+}
+
+func NewBzngSearch(body io.Reader) *BzngSearch {
+	return &BzngSearch{
+		reader: bzngio.NewReader(body, resolver.NewContext()),
+	}
+}
+
+// SetOnCtrl registers a callback function that will be fired when a control
+// payload is found in the search stream. Not safe for concurrent use, this
+// should be set before the first read is called.
+func (r *BzngSearch) SetOnCtrl(cb func(interface{})) {
+	r.onctrl = cb
+}
+
+func (r *BzngSearch) Read() (*zng.Record, error) {
+	for {
+		rec, b, err := r.reader.ReadPayload()
+		if err != nil {
+			return nil, err
+		}
+		if rec == nil && b == nil {
+			return nil, nil
+		}
+		if b != nil {
+			if !bytes.HasPrefix(b, []byte("json:")) {
+				// We expect only json control payloads.
+				// XXX should log error if something else,
+				// but just skip for now.
+				continue
+			}
+			ctrl, err := unpack(b[5:])
+			if err != nil {
+				return nil, err
+			}
+			if r.onctrl != nil {
+				r.onctrl(ctrl)
+			}
+			if end, ok := ctrl.(*TaskEnd); ok {
+				if end.Error != nil {
+					return nil, end.Error
+				}
+			}
+			continue
+		}
+		return rec, nil
+	}
+}
--- a/zqd/api/bzng.go
+++ b/zqd/api/bzng.go
@@ -11,8 +11,6 @@ import (
 
 type BzngSearch struct {
 	reader *bzngio.Reader
-	ctrl   interface{}
-	err    error
 	onctrl func(interface{})
 }
 
--- a/zqd/api/bzng.go
+++ b/zqd/api/bzng.go
@@ -30,33 +30,24 @@ func (r *BzngSearch) SetOnCtrl(cb func(interface{})) {
 func (r *BzngSearch) Read() (*zng.Record, error) {
 	for {
 		rec, b, err := r.reader.ReadPayload()
+		if err != nil || b == nil {
+			return rec, err
+		}
+		if !bytes.HasPrefix(b, []byte("json:")) {
+			// We expect only json control payloads.
+			// XXX should log error if something else,
+			// but just skip for now.
+			continue
+		}
+		ctrl, err := unpack(b[5:])
 		if err != nil {
 			return nil, err
 		}
-		if rec == nil && b == nil {
-			return nil, nil
+		if r.onctrl != nil {
+			r.onctrl(ctrl)
 		}
-		if b != nil {
-			if !bytes.HasPrefix(b, []byte("json:")) {
-				// We expect only json control payloads.
-				// XXX should log error if something else,
-				// but just skip for now.
-				continue
-			}
-			ctrl, err := unpack(b[5:])
-			if err != nil {
-				return nil, err
-			}
-			if r.onctrl != nil {
-				r.onctrl(ctrl)
-			}
-			if end, ok := ctrl.(*TaskEnd); ok {
-				if end.Error != nil {
-					return nil, end.Error
-				}
-			}
-			continue
+		if end, ok := ctrl.(*TaskEnd); ok && end.Error != nil {
+			return nil, end.Error
 		}
-		return rec, nil
 	}
 }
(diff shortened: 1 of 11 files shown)


zqd/api/connection.go
```diff
@@ -0,0 +1,364 @@
...
+	ErrSpaceNotFound = errors.New("space not found")
+	// ErrSpaceExists returns when specified the space already exists.
+	ErrSpaceExists = errors.New("space exists")
+)
+
+type httpClient interface {
+	Do(req *http.Request) (*http.Response, error)
+}
+
+type Connection struct {
```
alfred-landrum: For the lower-level connection setup, transport options, and http roundtrip, would you look into using a request library, like resty ( https://github.com/go-resty/resty )? It supports the features we need now, and is tested & documented, so it should be easy for us to add support for other features (like authentication options). It also allows getting raw access to the response reader, so it will be possible to handle our ndjson streaming responses with it.

zqd/api/connection.go
```diff
@@ -0,0 +1,364 @@
...
+	return e.Err
+}
+
+func (e *ErrorResponse) Error() string {
+	return fmt.Sprintf("status code %d: %v", e.StatusCode, e.Err)
+}
+
+// Ping checks to see if the server and measure the time it takes to
+// get back the response.
+func (c *Connection) Ping() (time.Duration, error) {
```
alfred-landrum: I think all of the higher level api calls (Ping, SpaceInfo, SpaceGet, etc) should take a context.Context.

zqd/api/connection.go
```diff
@@ -0,0 +1,364 @@
...
+
+// Search sends a search task to the server and returns a Search interface
+// that the caller uses to stream back results via the Pull method.
+func (c *Connection) Search(req SearchRequest, format string, params url.Values) (Search, error) {
+	return c.SearchWithContext(context.Background(), req, format, params)
+}
+
+// SearchWithContext is like PostSearch, except that it takes a
+// context that will be used in the underlying http request.
+func (c *Connection) SearchWithContext(ctx context.Context, req SearchRequest, format string, params url.Values) (Search, error) {
```
alfred-landrum: I would just make this call `Search` and drop the other function that doesn't take a context.

zqd/api/bzng.go
```diff
@@ -0,0 +1,84 @@
...
+				// We expect only json control payloads.
+				// XXX should log error if something else,
+				// but just skip for now.
+				continue
+			}
+			ctrl, err := unpack(b[5:])
+			if err != nil {
+				return nil, nil, err
+			}
+			if end, ok := v.(*TaskEnd); ok {
```
alfred-landrum: It looks to me like `v` will always be nil here because of the check at the beginning of the function?

zqd/api/bzng.go
```diff
@@ -0,0 +1,84 @@
...
+func NewBzngSearch(body io.Reader) *BzngSearch {
+	return &BzngSearch{
+		reader: bzngio.NewReader(body, resolver.NewContext()),
+	}
+}
+
+//XXX
+const mtu = 100
+
+func (r *BzngSearch) Pull() (zbuf.Batch, interface{}, error) {
```
alfred-landrum: How would we replace the bare `interface{}` return type here with a more meaningful type?

zqd/api/connection.go
mattnibs: ok

mattnibs: done

zqd/api/bzng.go
mattnibs: Still working on this... A switch is going to be unavoidable but I'm thinking for wrapping this is in a `SearchCtrl` struct.

zqd/api/connection.go
mattnibs: done

zqd/api/connection.go
mattnibs: I reimplemented this using resty as you suggested. Having gotten pretty familiar with the resty api by now here's my takeaway:

It works great for simple our simple request/response w/ json payloads. OTH on more complicated requests it more often than not seems to get in the way. For instance: For our streaming responses I have to disable response parsing to prevent resty from trying to read the entire response into memory. This has the side effect of disabling all middleware, which means I had to reimplement this functionality on streaming requests (i.e. search and packet post). I think we should still use resty, however, I imagine in the future we're going to want to implement our own version of resty that is easier to work with streaming data. The vault api client does something similar to this.

alfred-landrum: Do you mean we would implement all of resty's functionality, or that we would try to add ndjson streaming support to our own fork of resty?

zqd/api/connection.go
```diff
@@ -0,0 +1,261 @@
...
+	client.SetError(Error{})
+	client.OnAfterResponse(checkError)
+	return &Connection{
+		client: client,
+	}
+}
+
+// NewConnection creates a new connection with the given useragent string
+// and a base URL set up to talk to http://localhost:defaultport
+func NewConnection(useragent string) *Connection {
```
alfred-landrum: Since setting a non-default user agent will be rare, how about removing it from these constructors and provide a setter?

I'd also suggest that the constructors just take a url string, and not constructors that differ only on url-string vs *url.URL; http.NewRequest takes a `url string` parameter.

zqd/api/connection.go
```diff
@@ -0,0 +1,261 @@
...
+func (c *Connection) SetURL(u string) {
+	c.client.SetHostURL(u)
+}
+
+type ErrorResponse struct {
+	*resty.Response
+	Err error
+}
+
+func (e *ErrorResponse) Unwrap() error {
```
alfred-landrum: 👍

zqd/api/connection.go
```diff
@@ -0,0 +1,261 @@
...
+
+func (c *Connection) URL() string {
+	return c.client.HostURL
+}
+
+func (c *Connection) SetURL(u string) {
+	c.client.SetHostURL(u)
+}
+
+type ErrorResponse struct {
```
nwt: Nit: Maybe move this and its methods to the bottom of the file so it isn't in the middle of the Connection methods.

zqd/api/connection.go
mattnibs: Good idea.

zqd/api/bzng.go
mattnibs: Yep, rewrote this to just implement the `zbuf.Reader` interface which I think is a lot more straightforward.

zqd/api/connection.go
mattnibs: I suppose I meant reimplementing the resty functionality that we use. But really this is tbd

mattnibs:
@alfred-landrum et al. I'm happy with where this is at for now. This is ready for another round of review.

zqd/api/bzng.go
```diff
@@ -0,0 +1,64 @@
...
+	"io"
+
+	"github.com/brimsec/zq/zio/bzngio"
+	"github.com/brimsec/zq/zng"
+	"github.com/brimsec/zq/zng/resolver"
+)
+
+type BzngSearch struct {
+	reader *bzngio.Reader
+	ctrl   interface{}
```
henridf: `ctrl` and `err` are unused.

mattnibs: danke

zqd/api/bzng.go
```diff
@@ -0,0 +1,62 @@
...
+			}
+			if end, ok := ctrl.(*TaskEnd); ok {
+				if end.Error != nil {
+					return nil, end.Error
+				}
+			}
+			continue
+		}
+		return rec, nil
+	}
```
nwt: Nit: Control flow might be a little clearer this way:
```suggestion
	for {
		rec, b, err := r.reader.ReadPayload()
		if err != nil || b == nil {
			return rec, err
		}
		if !bytes.HasPrefix(b, []byte("json:")) {
			// We expect only json control payloads.
			// XXX should log error if something else,
			// but just skip for now.
			continue
		}
		ctrl, err := unpack(b[5:])
		if err != nil {
			return nil, err
		}
		if r.onctrl != nil {
			r.onctrl(ctrl)
		}
		if end, ok := ctrl.(*TaskEnd); ok && end.Error != nil {
			return nil, end.Error
		}
	}
```

mattnibs: beauts

zqd/api/connection.go
mattnibs: sure

PR #459, opened 2020-03-24
Add additional Math.* functions

aswan:


aswan:
> math.Mod is looking a bit lonely as the only function without a test :)

whoops, oversight, will add.

PR #461, opened 2020-03-25
Add String.parse* functions

aswan:


expr/functions.go
```diff
@@ -239,3 +245,48 @@ func mathSqrt(args []zngnative.Value) (zngnative.Value, error) {
 
 	return zngnative.Value{zng.TypeFloat64, r}, nil
 }
+
+func stringParseInt(args []zngnative.Value) (zngnative.Value, error) {
+	switch args[0].Type.ID() {
+	case zng.IdString, zng.IdBstring:
+		i, perr := strconv.ParseInt(args[0].Value.(string), 10, 64)
```
henridf:
```suggestion
		i, perr := strconv.Atoi(args[0].Value.(string))
```

henridf: nitty

henridf:
Not specific to this PR, but these functions will get added to the docs somewhere right?

aswan:
> Not specific to this PR, but these functions will get added to the docs somewhere right?

Yes, but we haven't documented expression syntax at all yet, in my head I have documenting functions lumped in with that...

PR #470, opened 2020-03-26
Improve scanner.Combiner error messages

henridf:
`scanner.Combiner` error strings previously included a string representation of
the `zbuf.Reader` go struct from which the error-causing line
originated. Remove this, and prefix the remaining (useful) error
message with the name of the file that had the error.

### DIFF
--- a/cmd/zq/zq.go
+++ b/cmd/zq/zq.go
@@ -282,6 +282,15 @@ func (c *Command) errorf(format string, args ...interface{}) {
 	_, _ = fmt.Fprintf(os.Stderr, format, args...)
 }
 
+type namedReader struct {
+	zbuf.Reader
+	name string
+}
+
+func (r namedReader) String() string {
+	return r.name
+}
+
 func (c *Command) inputReaders(paths []string) ([]zbuf.Reader, error) {
 	var readers []zbuf.Reader
 	for _, path := range paths {
@@ -319,7 +328,7 @@ func (c *Command) inputReaders(paths []string) ([]zbuf.Reader, error) {
 				return nil, err
 			}
 		}
-		readers = append(readers, zr)
+		readers = append(readers, namedReader{zr, path})
 	}
 	return readers, nil
 }
--- a/scanner/combiner.go
+++ b/scanner/combiner.go
@@ -49,7 +49,7 @@ func (c *Combiner) Read() (*zng.Record, error) {
 			if tup == nil {
 				c.done[k] = true
 				if err := c.closeReader(l); err != nil {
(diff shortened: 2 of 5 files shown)


aswan:
Instead of having the parallel arrays in Combiner, what about adding a new method to zbuf.Reader (or just using String() for this purpose) to return a useful representation, ie the file path?

henridf:
I considered that but insofar as the `zbuf.Reader` interface really just represents a stream of records (mirroring `io.Reader`), it felt mis-placed to add "Name" method to it. 

But I'm entirely ok with making that change, if others are ok with it. @mccanne @alfred-landrum @nwt @mattnibs any thoughts on adding a name method to `zbuf.Reader` ?

ztest/ztest.go
```diff
@@ -232,17 +233,19 @@ func run(zq, ZQL, outputFormat string, inputs ...string) (string, error) {
...
 	var readers []zbuf.Reader
-	for _, input := range inputs {
+	var names []string
+	for i, input := range inputs {
 		zr, err := detector.NewReader(detector.GzipReader(strings.NewReader(input)), zctx)
 		if err != nil {
 			return nil, err
 		}
 		readers = append(readers, zr)
+		names = append(names, strconv.Itoa(i))
```
mattnibs: Instead of changing the `NewCombiner` api to accept this extra arg, I would suggest wrapping `zr` with something like this:
```golang
type namedReader struct {
    zbuf.Reader
    name string
}

func (r *namedReader) String() string {
    return r.name
}
```

cmd/zq/zq.go
```diff
@@ -247,7 +247,7 @@ func (c *Command) Run(args []string) error {
 	if len(readers) == 1 {
 		reader = readers[0]
 	} else {
-		reader = scanner.NewCombiner(readers)
+		reader = scanner.NewCombiner(readers, paths)
```
mattnibs: I can't comment on `loadInputs` so I'll comment here. If you load files with `scanner.OpenFile` you don't need to do this- expect for the case of stdin, in which case I recommend wrapping with the `namedReader` type I referenced below.

ztest/ztest.go
henridf: That sounds good

henridf: Done

cmd/zq/zq.go
henridf: I originally had to change `loadInputs` due to the NewCombiner interface change, but that's no longer needed with your suggestion to pass a Stringer.

PR #471, opened 2020-03-26
Implement ast unpacking for expressions and PutProc

aswan:


philrz:
FYI, I've tried this out by moving the `zq` pointer in my Brim's `package.json` to hash `093f506` and copying that `zqd` version into `zdeps/`, and it works.

image

PR #475, opened 2020-03-27
Add more string functions

aswan:


expr/functions.go
```diff
@@ -39,6 +40,10 @@ var allFns = map[string]struct {
 	"String.parseFloat":  {1, 1, stringParseFloat},
 	"String.parseInt":    {1, 1, stringParseInt},
 	"String.parseIp":     {1, 1, stringParseIp},
+	"String.replace":     {3, 3, stringReplace},
+	"String.tolower":     {1, 1, stringToLower},
```
henridf: it looks like the above are doing camelCase... should these be toLower and toUpper?

aswan: oops, will fix

PR #482, opened 2020-03-30
Crude prototype for zng archive (zar)

mccanne:
This commit adds a zar command for createing and searching
bzng index files.  The index files are organized as sorted
string tables and use the sst package.

The current protoype is very early and and currently only
indexes IP addresses as a hard-wired configuration.  The code
however is set up to do more flexible indexing soon.

The sst package uses a simple binary format for framing
keys and values as arbitrary byte sequences but we realized
it could be a very powerful foundation to store the sst
files as bzng.  We will implement this change in a future PR.

### DIFF
--- a/archive/finder.go
+++ b/archive/finder.go
@@ -1,7 +1,6 @@
 package archive
 
 import (
-	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -10,7 +9,6 @@ import (
 )
 
 func Find(dir string, pattern []byte) ([]string, error) {
-	nerr := 0
 	//XXX this should be parallelized with some locking presuming a little
 	// parallelism won't mess up the file system assumptions
 	var hits []string
@@ -31,11 +29,6 @@ func Find(dir string, pattern []byte) ([]string, error) {
 			hit, err := SearchFile(path, pattern)
 			if err != nil {
 				fmt.Printf("%s\n", err)
-				nerr++
-				if nerr > 10 {
-					//XXX
-					return errors.New("stopping after too many errors...")
-				}
 			}
 			if hit {
 				hits = append(hits, path)
--- a/archive/finder.go
+++ b/archive/finder.go
@@ -9,7 +9,6 @@ import (
 	"github.com/brimsec/zq/pkg/sst"
 )
 
-//XXX this is a test searches for IP addresses
 func Find(dir string, pattern []byte) ([]string, error) {
 	nerr := 0
 	//XXX this should be parallelized with some locking presuming a little
@@ -47,6 +46,7 @@ func Find(dir string, pattern []byte) ([]string, error) {
 	return hits, err
 }
 
+//XXX for now we hard code search for IP address.
 func SearchFile(path string, pattern []byte) (bool, error) {
 	subdir := path + zarExt
 	sstName := "sst:type:ip" //XXX
--- a/archive/finder.go
+++ b/archive/finder.go
@@ -1,0 +1,68 @@
+package archive
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/brimsec/zq/pkg/sst"
+)
+
+//XXX this is a test searches for IP addresses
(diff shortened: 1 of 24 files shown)


archive/finder.go
```diff
@@ -0,0 +1,68 @@
...
+				return filepath.SkipDir
+			}
+			// descend...
+			return nil
+		}
+		if filepath.Ext(name) == ".bzng" {
+			hit, err := SearchFile(path, pattern)
+			if err != nil {
+				fmt.Printf("%s\n", err)
+				nerr++
```
alfred-landrum, line 34: Even in this prototype, why skip over any errors?

archive/type.go
```diff
@@ -0,0 +1,216 @@
...
+			if colType == t.Type {
+				t.value(body)
+			}
+		}
+	}
+	return nil
+}
+
+func (t *TypeIndexer) value(body zcode.Bytes) {
+	t.Table.Enter(string(body), nil)
```
mattnibs, line 215: So the index simply says whether or not a value exists to its attached file. Isn't this the same as a perfectly accurate bloom filter that trades accuracy of results for space? Why not just use a bloom filter? What's the benefit?

It feels like if you are going the approach of creating indexes then it would make sense to produce a single index for a multiple files and the value of the index points to "hits". Otherwise you incur the cost of space for redundant key storage- something I image you would want to reduce.

pkg/sst/sst.go
```diff
@@ -0,0 +1,251 @@
...
+// and so forth.  The index files can always be recreated from the base table.
+//
+// A file constists of a sequence of frames, where each frame consists of a
+// frame header, indicating its length (on disk) and compression type, followed
+// by a possibly compressed body, where the body costs of a sequence of key, value
+// pairs.  Each key and each value is encoded as a 4-byte length and sequence of bytes
+// representing the key's string or the value's byte slice.  The body is compressed
+// according to the compression type.
+//
+// When an SST file has values of all the same length, the length is indicated
```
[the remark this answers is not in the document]
mccanne: The benefit from this old design was that you didn't need to encode the length of every key if you the keys were all fixed length as with the index-of-index files.  The plan is to move to zng representation then this won't matter.

archive/indexer.go
```diff
@@ -0,0 +1,113 @@
...
+	if err != nil {
+		return err
+	}
+	table, err := indexTypeIP(reader)
+	if err != nil {
+		return err
+	}
+	if table.Size() == 0 {
+		//XXX
+		return errors.New("nothing to index")
```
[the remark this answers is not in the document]
mccanne: Yes, this would be a better approach but the current logic doesn't handle the empty table.  I would prefer to handle this later when we soon switch over to zng.

archive/finder.go
mccanne: It's not skipping errors.  It was bailing after too many errors.  I'll take it out.

archive/type.go
mccanne: For low cardinality value sets, the overhead of redundant storage is far smaller than the size of the source data and the robustness and simplicity of having a separate index per archive file is very powerful.  For high cardinality value sets, the amount of overlap between files is usually going to be small (think of uids or file hashes) so the savings you would get from aggregation is small.  If it's not small, in worst case you are doubling the size of the data, but I think that will be rare in practice.  Bloom filters can definitely be useful here and we can get to that soon.

mccanne: Another consideration here regarding bloom filters is the ux for archive search involves returning the set of matched files so bloom-filter false positives could be counterintuitive.

alfred-landrum: One advantage of separate indices is time-based retention: if you want to delete old data on a periodic basis, it's easy to drop "old data" by removing the index and zng for that old range, and not touch the rest of the data set.

mccanne:
> Should the commands have test coverage so we know they continue to work?

Totally agree.  I'm going to merge this PR now as a WIP since it doesn't impact zq/zqd at all and we can then port SST to zng, refine the command syntax, and add tests as you suggest in subsequent PRs.

PR #483, opened 2020-03-31
Remove unused test helpers

henridf:
no longer needed as of #458


henridf:
Trapped by `// +build zeek` once again ... this stuff is needed.

PR #485, opened 2020-03-31
Revise len() for strings/bstrings

aswan:
Since len() is ambiguous when operating on bstrings, change it to only
work on arrays and sets, and introduce String.byteLen() and String.runeLen()
which return the length of a string in either raw bytes or UTF8 encoded runes.

### DIFF
--- a/expr/functions.go
+++ b/expr/functions.go
@@ -7,6 +7,7 @@ import (
 	"net"
 	"strconv"
 	"strings"
+	"unicode/utf8"
 
 	"github.com/brimsec/zq/zng"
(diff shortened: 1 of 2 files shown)


henridf:
Is there a downside to keeping `len()` on the string types, returning whichever of the two kinds of length we think is most commonly used ? (I'd think that's byte length)

aswan:
> Is there a downside to keeping len() on the string types, returning whichever of the two kinds of length we think is most commonly used ? 

Just the ambiguity and potential confusion that results from it.  I don't really have a strong opinion one way or the other.

PR #487, opened 2020-03-31
log post endpoint

henridf:
This PR adds a `POST /space/{space}/log` endpoint to post a list of log files that are to be ingested into a space. Files can be in any format supported by `zq`. 

Like for the pcap endpoint, each ingest operation truncates the space and concurrent ingest operations into a space are rejected.

Unlike for the pcap endpoint, streaming stats are not sent during ingest, but we could add them as/when they become necessary. In fact a likely first step might be to send a single stats update before the final `TaskEnd`, which would be useful for clients; but this can come in follow-on work.

This PR has some very basic test coverage in `zqd/handlers_test.go`;  more complete tests will come later.

### DIFF
--- a/zqd/api/connection.go
+++ b/zqd/api/connection.go
@@ -194,10 +194,22 @@ func (c *Connection) Search(ctx context.Context, search SearchRequest) (Search, 
 
 func (c *Connection) PostPacket(ctx context.Context, space string, payload PacketPostRequest) (*Stream, error) {
 	req := c.Request(ctx).
-		SetBody(payload).
-		SetHeader("format", "bzng")
+		SetBody(payload)
 	req.Method = http.MethodPost
 	req.URL = path.Join("/space", url.PathEscape(space), "packet")
+	r, err := c.stream(req)
+	if err != nil {
+		return nil, err
+	}
+	jsonpipe := NewJSONPipeScanner(r)
+	return NewStream(jsonpipe), nil
+}
+
+func (c *Connection) PostLogs(ctx context.Context, space string, payload LogPostRequest) (*Stream, error) {
+	req := c.Request(ctx).
+		SetBody(payload)
+	req.Method = http.MethodPost
+	req.URL = path.Join("/space", url.PathEscape(space), "log")
 	r, err := c.stream(req)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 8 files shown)


zqd/api/connection.go
```diff
@@ -206,6 +205,19 @@ func (c *Connection) PostPacket(ctx context.Context, space string, payload Packe
...
+	req.Method = http.MethodPost
+	req.URL = path.Join("/space", url.PathEscape(space), "log")
+	r, err := c.stream(req)
+	if err != nil {
+		return nil, err
+	}
+	jsonpipe := NewJSONPipeScanner(r)
+	return NewStream(jsonpipe), nil
+}
+
```
nwt: Nits: Maybe factor out code shared with PostPacket?

zqd/api/connection.go
```diff
@@ -206,6 +205,19 @@ func (c *Connection) PostPacket(ctx context.Context, space string, payload Packe
 	return NewStream(jsonpipe), nil
 }
 
+func (c *Connection) PostLogs(ctx context.Context, space string, payload LogPostRequest) (*Stream, error) {
```
nwt: Nit: Maybe call this PostLog for consistency with LogPostRequest and /space/{name}/log? (Or better yet, since the endpoint handles multiple logs, switch to LogsPostRequest and /space/{name}/logs.)

mattnibs: For consistency I would call this func `LogPost`.

zqd/handlers.go
```diff
@@ -307,6 +307,51 @@ func handlePacketPost(c *Core, w http.ResponseWriter, r *http.Request) {
...
+	if len(req.Paths) == 0 {
+		http.Error(w, "empty paths", http.StatusBadRequest)
+		return
+	}
+	var ingestErr error
+	done := make(chan struct{})
+	go func() {
+		ingestErr = ingest.Logs(r.Context(), s, req.Paths, c.SortLimit)
+		close(done)
+	}()
```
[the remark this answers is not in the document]
henridf: la classe!

zqd/api/connection.go
henridf: > (The rest of this file follows Noun Verb)

@mattnibs not the immediately preceding function (`PostPacket`) just above... I'll fix that one too.

PR #492, opened 2020-04-01
Add waterfall logger

mattnibs:
The waterfall logger distributes logs to the underlying loggers
in a waterfall pattern; for each log entry the logger will iterate through
each of its children, in order, stopping when it finds a logger that will
accept the log entry.

The waterfall logger will allow us to send named loggers to different
destinations (e.g. The noisy http.access logger) without having to
write complicated ! rules on the default logger.

Also:
- Make the namespace of the default logger ""

### DIFF
--- a/cmd/zqd/logger/logger.go
+++ b/cmd/zqd/logger/logger.go
@@ -1,8 +1,19 @@
 package logger
 
 import (
+	"fmt"
+	"io/ioutil"
+
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
+)
+
+type LoggerType string
+
+const (
+	TypeSink      LoggerType = "sink"
+	TypeWaterfall            = "waterfall"
+	TypeTee                  = "tee"
 )
 
 type Config struct {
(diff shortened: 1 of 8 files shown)


cmd/zqd/logger/logger.go
```diff
@@ -1,20 +1,56 @@
...
+	"io/ioutil"
+
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 )
 
+type LoggerType string
+
+const (
+	TypeSink      LoggerType = "sink"
```
nwt, line 14: Nit: Sink looks like it's a synonym for file. If that's the case, I think file is clearer terminology.

cmd/zqd/logger/logger.go
```diff
@@ -1,20 +1,56 @@
...
+	// Specifies underlying children loggers. Only applicable when Type is
+	// TypeWaterfall or TypeTee.
+	Children []Config `yaml:"children,omitempty"`
+}
+
+func New(conf Config) (zapcore.Core, error) {
+	return newCore(conf)
+}
+
+func newCore(conf Config) (zapcore.Core, error) {
```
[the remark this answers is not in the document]
mattnibs: Yep.

mattnibs:
@nwt great comments, thank you.

cmd/zqd/logger/logger.go
```diff
@@ -1,35 +1,83 @@
...
+)
+
 type Config struct {
 	Path string `yaml:"path"`
 	// If Path is a file, Mode will determine how the log file is managed.
 	// FileModeAppend is the default if value is undefined.
 	Mode  FileMode      `yaml:"mode,omitempty"`
 	Name  string        `yaml:"name"`
 	Level zapcore.Level `yaml:"level"`
+	// Type specifies how log entries are handled. If Type is Waterfall or Tee
```
alfred-landrum, line 26: I see that this looks designed to make the existing Brim default logger config continue to work:
https://github.com/brimsec/brim/blob/master/src/js/zqd/zqd.js#L35

Did you try it out to make sure its ok? If so, once we cut a zqd release with this PR's change & update Brim to use it, should we change that default config in Brim?

zqd/middleware.go
```diff
@@ -78,7 +77,7 @@ func panicCatchMiddleware(logger *zap.Logger) mux.MiddlewareFunc {
 					return
 				}
 				rstr := fmt.Sprint(r)
-				logger.Error("panic", zap.String("error", rstr))
+				logger.DPanic("Panic", zap.Any("error", rstr))
```
alfred-landrum, line 80: Does `zap.Any` do something special? `rstr` is a string, so wondering what the output difference will be.

mattnibs: It was a comment from @nwt. But I think he meant I should get rid of the `fmt.Sprint(r)` and do `zap.Any`

cmd/zqd/logger/logger.go
mattnibs: The config file in brim will have to change. It should look like this instead:
```yaml
logger:
  type: waterfall
  children:
  - name: http.access
     level: info
     path: ${accessLogFile}
     mode: rotate
  - level: info
     path: ${zqdLogFile}
     mode: rotate
```

zqd/middleware.go
nwt: @mattnibs: That's what I had in mind.

mattnibs: But actually the fact that we also return the string in the below `http.Error` says to me that we should keep it as-is.

nwt: Right you are. Sorry I missed that.

cmd/zqd/logger/logger.go
mattnibs: I have not checked if this works with Brim. Doing that now.

mattnibs: Verified that if you make the above changes, this will work as desired with brim.

mattnibs:
Accompanying brim pr: https://github.com/brimsec/brim/pull/540

PR #500, opened 2020-04-01
Graceful http shutdown

mattnibs:
Add pkg httpd. httpd.Server is an http Server that provides user-friendly
graceful http shutdown. The Server instance's Run and Start method both take
a context that will be used as the base context for all requests issued
initiated by the server. If the passed context is cancelled,  the Server
closes the active listener and waits for all active connections to become
idle. Because the passed context is used base context for all requests,
when the context is cancelled all active requests in turn receive a
cancellation signal. If the wait for active connections to idle exceeds
the duration set in httpd.ShutdownTimeout the Server forcefully closes all
active connections.

It is up to the implmentor of each handler route to determine whether to
listen to a request's ctx.Done() signal. It is in this pr author's opinion
that short-lived requests (e.g. space info, space list, status, etc) should
ignore ctx.Done() and all long running requests (e.g. searches and data posts
to spaces) should start to exit on ctx.Done(). This will the effect that during
graceful shutdown short-lived requests will be allowed to finish to completion,
while long running requests will terminate and prevent a forced shutdown.

Additional changes:
- A pkg catcher: a user friendly interface for catching signals in
a user friend manner.
- Have zqd listen utilize pkg catcher and httpd to trigger a graceful
shutdown when an interrupt signal is received.


mattnibs:
FYI to reviews. Based on waterfall-logger. I will wait till that branch (#492) is merged before rebasing this onto master and merging.

pkg/catcher/catcher.go
```diff
@@ -0,0 +1,56 @@
...
+func (s *SignalCatcher) SetOnCaught(cb func(os.Signal)) {
+	s.oncaught = cb
+}
+
+func (s *SignalCatcher) Catch(ctx context.Context) context.Context {
+	//XXX this could be made more efficient by having the signal handler
+	// be a parent of each context created and keeping just one signal
+	// channel around.  then when we get the signal, we cancel the parent
+	// context and all the children get canceled.
+	// but for now, this is ok
```
alfred-landrum, line 33: Please remove; I don't think it's helping to explain, as its describing how this is essentially being used now.

pkg/catcher/catcher.go
```diff
@@ -0,0 +1,56 @@
...
+	signals := make(chan os.Signal, 1)
+	signal.Notify(signals, s.sigs...)
+	childctx, cancel := context.WithCancel(ctx)
+	go func() {
+		// The for loop waits until we get a signal or a done.
+		// When we're done, we return and this goroutine exits.
+		// When we get a signal, we call cancel and continue the
+		// loop and wait for the done to finally make us leave.
```
alfred-landrum, line 41: This comment doesn't seem necessary, as it's just describing the straightforward code below.

pkg/httpd/server.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+var ShutdownTimeout = time.Second * 5
+
+type Server struct {
+	addr   string
+	lnAddr string
+	logger *zap.Logger
+	srv    *http.Server
+	done   chan struct{}
```
alfred-landrum, line 20: done as used below can be a sync.WaitGroup.

pkg/httpd/server.go
```diff
@@ -0,0 +1,95 @@
...
+			errCh <- err
+		}
+		close(errCh)
+	}()
+	var srvErr error
+	select {
+	case srvErr = <-errCh:
+	case <-ctx.Done():
+	}
+	timeout, cancel := context.WithTimeout(context.Background(), ShutdownTimeout)
```
alfred-landrum, line 78: Can you add a log here (and a blank line too) to identify the shift from  'waiting & serving' phase from the 'shutting down'  phase?

pkg/httpd/server.go
```diff
@@ -0,0 +1,95 @@
...
+	"context"
+	"errors"
+	"net"
+	"net/http"
+	"time"
+
+	"go.uber.org/zap"
+)
+
+var ShutdownTimeout = time.Second * 5
```
alfred-landrum, line 13: This can be a const.

pkg/httpd/server.go
```diff
@@ -0,0 +1,95 @@
...
+	var srvErr error
+	select {
+	case srvErr = <-errCh:
+	case <-ctx.Done():
+	}
+	timeout, cancel := context.WithTimeout(context.Background(), ShutdownTimeout)
+	defer cancel()
+	err := s.srv.Shutdown(timeout)
+	// This means Server was already closed before Shutdown was called. The
+	// error from  is most interesting here.
```
alfred-landrum, line 82: This comment is incomplete & probably unnecessary.

pkg/httpd/server.go
```diff
@@ -0,0 +1,95 @@
...
+	if errors.Is(err, http.ErrServerClosed) {
+		err = srvErr
+	} else if errors.Is(err, context.DeadlineExceeded) {
+		s.logger.Warn("Server shutdown deadline exceeded", zap.Duration("duration", ShutdownTimeout))
+		// Return context.DeadlineExceeded unless srv.Close() materilizes an
+		// error.
```
alfred-landrum, line 88: I think this is an unnecessary comment.

pkg/httpd/server.go
mattnibs: On a test I rely on being able to change this. Generally I don't think it's bad to have a knob to change this.

pkg/httpd/server.go
mattnibs: fair

cmd/zqd/listen/command.go
```diff
@@ -85,12 +89,10 @@ func (c *Command) Run(args []string) error {
 	if c.pprof {
 		h = pprofHandlers(h)
 	}
-	ln, err := net.Listen("tcp", c.listenAddr)
-	if err != nil {
-		return err
-	}
-	c.logger.Info("Listening", zap.Stringer("addr", ln.Addr()))
-	return http.Serve(ln, h)
+	ctx := c.catcher.Catch(context.Background())
```
nwt, line 92: I think it would be clearer to ditch catcher and do this:
```suggestion
	ctx, cancel := context.WithCancel(context.Background())
	ch := make(chan os.Signal, 1)
	signal.Notify(ch, os.Interrupt)
	go func() {
		sig := <- ch
		c.logger.Info("Signal received", zap.Stringer("signal", sig))
		cancel()
	}()
```

alfred-landrum: I agree with @nwt : on the whole, I think its simpler to just handle the signals & cancellation in these few lines of code directly where its needed.

pkg/httpd/server.go
alfred-landrum: 👍 I didn't notice the test case usage before now.

pkg/httpd/server.go
alfred-landrum: just wanted to bump the above comment

cmd/zqd/listen/command.go
mattnibs: ok

pkg/httpd/server.go
mattnibs: done.

cmd/zqd/listen/command.go
mattnibs: For the record, I still think `catcher` looks prettier than this.

pkg/httpd/server.go
alfred-landrum: @mattnibs : I don't see the change?

mattnibs: For some reason it's not rendering here. If you look at the files directly the change is there: https://github.com/brimsec/zq/pull/500/files#diff-a2c896ceeab22af556c3fb6d1c953cc1R77

mattnibs: Now I understand. I added a log line for "Shutting down" along with the reason for shutdown.

pkg/httpd/server.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+func (s *Server) serve(ctx context.Context, ln net.Listener) {
+	defer s.done.Done()
+	errCh := make(chan error)
+	go func() {
+		if err := s.srv.Serve(ln); err != nil && !errors.Is(err, http.ErrServerClosed) {
+			errCh <- err
+		}
+		close(errCh)
```
nwt, line 69:
```suggestion
		errCh <- s.srv.Serve(ln)
```

mattnibs: nice

PR #506, opened 2020-04-02
Send LogPostStatus after succesful log ingest

henridf:


zqd/api/api.go
```diff
@@ -117,6 +117,13 @@ type PacketPostStatus struct {
 	MaxTime        *nano.Ts `json:"max_time,omitempty"`
 }
 
+type LogPostStatus struct {
+	Type    string   `json:"type"`
+	MinTime *nano.Ts `json:"min_time,omitempty"`
+	MaxTime *nano.Ts `json:"max_time,omitempty"`
+	Size    int64    `json:"size" unit:"bytes"`
+}
+
```
nwt, line 126: Nit: Put this after LogPostRequest.

zqd/handlers_test.go
```diff
@@ -260,9 +260,20 @@ func TestPostLogs(t *testing.T) {
...
+			MinTime: &min,
+			MaxTime: &max,
+			Size:    80,
+		}
+		require.Equal(t, expected, status)
+
+		taskend := payloads[len(payloads)-1].(*api.TaskEnd)
+		assert.Equal(t, taskend.Type, "TaskEnd")
+		assert.Nil(t, taskend.Error)
+
```
nwt, line 276:
```suggestion
```

mattnibs:
Can we hold off on merging for a sec? Just want to finish my review.

zqd/api/api.go
henridf: done

mattnibs:
Why have this response when the user could just issue a space/info request upon completion? I'm struggling to see a value add that justifies the added part.

henridf:
Fair question @mattnibs, here's the missing PR description:

I added this out of symmetry with the pcap endpoint and with the expectation that it would evolve into periodic updates for large ingest jobs, in support of a progress bar. Similarly to how the pcap endpoint evolved.

mattnibs:
@henridf thanks for the explanation. Sounds good to me then.

PR #507, opened 2020-04-02
Bzng write test

aswan:


tests/formats/bzng/write.yaml
```diff
@@ -0,0 +1,49 @@
+zql: '*'
+
+input: |
+  #0:record[_path:string,ts:time,d:float64]
+  0:[a;10;1;]
+  0:[xyz;20;1.5;]
+
+output-format: bzng
+
+output: |
```
alfred-landrum, line 10: I think the format for encoding binary here is really neat. Would you consider using a different tag, instead of the `HEX` marker? For example, `outputHex` or something similar; it would be an error to use both `output` and `outputHex`.

ztest/ztest.go
```diff
@@ -182,6 +196,47 @@ func (i *Inputs) UnmarshalYAML(value *yaml.Node) error {
 	return nil
 }
 
+// Try to decode a yaml-friendly way of representing binary data in hex:
+// each lines is either a comment explaining the contents (denoted with
```
alfred-landrum, line 200:
```suggestion
// each line is either a comment explaining the contents (denoted with
```

Issue #499, opened 2020-04-01
Add type context reset control code to spec and zq


philrz:

aswan:
This was done in #508

PR #508, opened 2020-04-03
Bzng framing

aswan:
(this is based on the bzng write test branch, but is intended for merging to master after that branch)
Addresses #499

### DIFF
--- a/zio/bzngio/writer.go
+++ b/zio/bzngio/writer.go
@@ -4,25 +4,50 @@ import (
 	"io"
 
 	"github.com/brimsec/zq/zcode"
+	"github.com/brimsec/zq/zio"
 	"github.com/brimsec/zq/zng"
 	"github.com/brimsec/zq/zng/resolver"
 )
 
 type Writer struct {
 	io.Writer
-	encoder *resolver.Encoder
-	buffer  []byte
+	zio.Flags
+	encoder      *resolver.Encoder
+	buffer       []byte
+	frameRecords int
+	frameBytes   int
 }
 
-func NewWriter(w io.Writer) *Writer {
+func NewWriter(w io.Writer, flags zio.Flags) *Writer {
 	return &Writer{
-		Writer:  w,
-		encoder: resolver.NewEncoder(),
-		buffer:  make([]byte, 0, 128),
+		Writer:       w,
+		Flags:        flags,
+		encoder:      resolver.NewEncoder(),
+		buffer:       make([]byte, 0, 128),
+		frameRecords: -1,
+		frameBytes:   0,
 	}
 }
 
+func (w *Writer) NewFrame() error {
+	w.encoder.Reset()
+
+	var marker []byte
+	marker = append(marker, zng.FrameMarker)
+	marker = zcode.AppendUvarint(marker, uint64(w.frameBytes))
+	n, err := w.Writer.Write(marker)
+
+	w.frameRecords = 0
+	w.frameBytes = n
+
+	return err
+}
+
 func (w *Writer) Write(r *zng.Record) error {
+	if w.FrameSize > 0 && (w.frameRecords < 0 || w.frameRecords >= w.FrameSize) {
+		w.NewFrame()
+	}
+
 	// First send any typedefs for unsent types.
 	typ := w.encoder.Lookup(r.Type)
 	if typ == nil {
@@ -33,10 +58,11 @@ func (w *Writer) Write(r *zng.Record) error {
 			return err
 		}
 		w.buffer = b
-		_, err = w.Writer.Write(b)
+		n, err := w.Writer.Write(b)
 		if err != nil {
 			return err
 		}
+		w.frameBytes += n
 	}
 	dst := w.buffer[:0]
 	id := typ.ID()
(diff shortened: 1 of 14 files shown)


zng/docs/spec.md
```diff
@@ -110,6 +110,16 @@ Note that the value encoding need not refer to the field names and types as that
...
+Since type definitions are carried in-band in a ZNG stream,
+the entire stream must generally be processed in-order to maintain
+accurate and complete information on types.
+To facilitate random access into large stored ZNG streams, a stream
+may also be optionally organized into a sequence of "frames",
+each of which has an independent type context and can thus be
+processed correctly without first processing the entire preceding stream.
+This benefit comes at the cost of some additional overhead --
+the space consumed by frame boundary markers and repeated type definitions.
+
```
nwt, line 122: Are we adding framing to the ZNG text format? I see no reason we should, and if we don't, then framing isn't part of the ZNG data model, and this paragraph belongs in the next section.

zng/docs/spec.md
```diff
@@ -335,6 +346,46 @@ subsequent key for values that have previous keys of equal value.
 It is an error for any such values to appear that contradicts the most
 recent ordering directives.
 
+### 2.1.3 Frame Markers
+
+A frame marker divides a ZNG stream into a sequence of smaller frames
+which can be decoded indepenently.  It is left up to implementations to
+decide how frequently to use these markers.
```
henridf, line 353: `s/decide how frequently/decide if and how frequently/` ?

zng/docs/spec.md
```diff
@@ -110,6 +110,16 @@ Note that the value encoding need not refer to the field names and types as that
 completely captured by the type ID.  Values merely encode the value
 information consistent with the referenced type ID.
 
+Since type definitions are carried in-band in a ZNG stream,
+the entire stream must generally be processed in-order to maintain
+accurate and complete information on types.
+To facilitate random access into large stored ZNG streams, a stream
```
henridf, line 116: s/To facilitate random access/to enable bi-directional reading and facilitate random access/?

zng/docs/spec.md
```diff
@@ -335,6 +346,46 @@ subsequent key for values that have previous keys of equal value.
...
+
+Implementations are encouraged to include the previous frame length
+as it allows a BZNG processor to read a large stored BZNG stream backward.
+To do this, a frame marker must be placed at the very end of the
+stream so that a reader may seek to the end of the stream and read the
+final marker and then use `<prevlen>` to find the last frame that holds
+records.  The contents of the each frame must be read in the order they
+appear in the stream.
+
+To process every individual record in a stream in reverse order,
```
henridf, line 382: It might be worth clarifying that the "reverse order" here is unrelated to the orthogonal to the order of "ordering hints" from the preceding section.

zng/docs/spec.md
```diff
@@ -335,6 +346,46 @@ subsequent key for values that have previous keys of equal value.
 It is an error for any such values to appear that contradicts the most
 recent ordering directives.
 
+### 2.1.3 Frame Markers
```
henridf, line 349: This comment probably belongs in section 2.1.2 but I can't comment there: are ordering hints scoped to their frame or global?

aswan:
Trying to merge master after #507 merged was a catastrophe, I'm going to rebase and force-push this, apologies in advance for any lost context on comments.

zng/docs/spec.md
aswan: Good question.  One of the things nagging at me about this effort is that the goal here is to be able to build an external index, but we don't have any real principled reasoning about what goes in-band and what is kept externally.
The idea of the external index is that it only works if the bzng file is sorted and contains frame markers.  If the presence of an index means that the thing that created the index verified that these things are true, then what's the point of adding an additional in-band hint?  If it is just for some hypothetical future consumer, then we don't have enough information to answer your question :/

zio/zio.go
```diff
@@ -1,16 +1,28 @@
...
 	EpochDates bool
+	FrameSize  int
+}
+
+func (f *Flags) SetFlags(fs *flag.FlagSet) {
+	fs.BoolVar(&f.ShowTypes, "T", false, "display field types in text output")
+	fs.BoolVar(&f.ShowFields, "F", false, "display field names in text output")
+	fs.BoolVar(&f.EpochDates, "E", false, "display epoch timestamps in text output")
+	fs.BoolVar(&f.UTF8, "U", false, "display zeek strings as UTF-8")
+	fs.IntVar(&f.FrameSize, "b", 0, "BZNG frame size (0 for no frames)")
```
nwt, line 25: This stuff should stay in cmd/zq/zq.go.

aswan: Some context would be helpful -- I moved it to re-use it from ztest.go.  Do you prefer duplicating it there?

nwt: Doh! I forgot that cmd/zq is main package, making my suggestion impossible. Sorry about that.

zio/bzngio/writer.go
```diff
@@ -4,25 +4,43 @@ import (
...
-		buffer:  make([]byte, 0, 128),
+		Writer:       w,
+		Flags:        flags,
+		encoder:      resolver.NewEncoder(),
+		buffer:       make([]byte, 0, 128),
+		frameRecords: -1,
 	}
 }
 
+func (w *Writer) NewFrame() error {
```
nwt, line 30: Nit: Maybe call this StartFrame or StartNewFrame since NewFrame reads like a constructor.

zio/zio.go
```diff
@@ -1,16 +1,28 @@
...
 
 import (
+	"flag"
 	"io"
 
 	"github.com/brimsec/zq/zbuf"
 )
 
+// This is clumsy, it has the union of the flags accepted by all
+// the different Writer classes.
```
nwt, line 11:
```suggestion
// Flags has the union of the flags accepted by all the different Writer implementations.
```

zng/docs/spec.md
```diff
@@ -126,7 +126,7 @@ or a value message (0).
 ### 2.1 Control Messages
 
 The lower 7 bits of a control header byte define the control code.
-Control codes 0 through 5 are reserved for BZNG:
+Control codes 0 through 6 are reserved for BZNG:
```
nwt, line 129: Shouldn't this still be 5?

zio/zio.go
nwt, line 25: Usage message should identify the units. Also, they're records now, but would bytes be more useful?

aswan: > would bytes be more useful?

I don't know, using bytes would be complicated by compression if/when it is added.  It appears to me that this is simpler and meets our immediate needs for indexing...

nwt: 👍

Issue #505, opened 2020-04-02
Add support for typed NDJSON to the /space/{space}/log endpoint


henridf:
The `LogPostRequest` that is passed when POSTing to this endpoint should be expanded to accommodate passing a `types.json` struct; this will have the same effect as passing `-j types.json` to `zq`.

PR #514, opened 2020-04-04
Fill in `record.Ts` when creating a record in the ndjson reader

henridf:
Discovered this bug while working on #505 .

### DIFF
--- a/zio/ndjsonio/reader.go
+++ b/zio/ndjsonio/reader.go
@@ -135,6 +135,5 @@ again:
 	if err := record.TypeCheck(); err != nil {
 		return nil, err
 	}
-	// Ignore error, which just means the point doesn't have a time-typed ts field
 	return record, nil
 }
--- a/zio/ndjsonio/reader.go
+++ b/zio/ndjsonio/reader.go
@@ -128,5 +128,14 @@ again:
 		return nil, fmt.Errorf("line %d: %w", r.scanner.Stats.Lines, err)
 	}
 	outType := r.zctx.LookupTypeRecord(zv.Type.(*zng.TypeRecord).Columns)
-	return zng.NewRecordCheck(outType, 0, zv.Bytes)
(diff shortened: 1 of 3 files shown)


zio/ndjsonio/reader.go
```diff
@@ -128,5 +128,14 @@ again:
...
 	outType := r.zctx.LookupTypeRecord(zv.Type.(*zng.TypeRecord).Columns)
-	return zng.NewRecordCheck(outType, 0, zv.Bytes)
+	record, err := zng.NewRecordCheck(outType, 0, zv.Bytes)
+	if err != nil {
+		return nil, err
+	}
+	ts, err := record.AccessTime("ts")
+	if err == nil {
+		record.Ts = ts
+	}
```
nwt, line 138: Use zng.NewRecord to avoid a map access by taking advantage of TypeRecord.TsCol.
```suggestion
	record, err := zng.NewRecord(outType, zv.Bytes)
	if err != nil {
		return nil, err
	}
	if err := record.TypeCheck(); err != nil {
		return nil, err
	}
```

henridf: 👍

zio/ndjsonio/reader.go
```diff
@@ -128,5 +128,13 @@ again:
...
 	outType := r.zctx.LookupTypeRecord(zv.Type.(*zng.TypeRecord).Columns)
-	return zng.NewRecordCheck(outType, 0, zv.Bytes)
+	record, err := zng.NewRecord(outType, zv.Bytes)
+	if err != nil {
+		return nil, err
+	}
+	if err := record.TypeCheck(); err != nil {
+		return nil, err
+	}
+	// Ignore error, which just means the point doesn't have a time-typed ts field
```
nwt, line 138: Just realized this comment no longer applies.
```suggestion
```

PR #516, opened 2020-04-06
Extract zbuf.ReadBatch from scanner.(*Scanner).Pull

nwt:


zbuf/batch.go
```diff
@@ -0,0 +1,50 @@
+package zbuf
+
+import (
+	"github.com/brimsec/zq/pkg/nano"
+	"github.com/brimsec/zq/zng"
+)
+
+// Batch is an inteface to a bundle of records.
```
henridf, line 8:
```suggestion
// Batch is an interface to a bundle of records.
```

nwt: 👍

PR #517, opened 2020-04-06
Support typed NDJSON in log ingest

henridf:
#505 

This needs more tests for error conditions, but how/what errors are returned is going to change soon so I'll add those tests with that change.

### DIFF
--- a/zqd/handlers_test.go
+++ b/zqd/handlers_test.go
@@ -338,7 +338,7 @@ func TestPostNDJSONLogs(t *testing.T) {
 	})
 	t.Run("Search", func(t *testing.T) {
 		res := zngSearch(t, client, spaceName, "*")
-		expected := "#0:record[_path:string,ts:time,uid:bstring]\n0:[http;2;CXY9a54W2dLZwzPXf1;]\n0:[http;1;CXY9a54W2dLZwzPXf1;]"
+		const expected = "#0:record[_path:string,ts:time,uid:bstring]\n0:[http;2;CXY9a54W2dLZwzPXf1;]\n0:[http;1;CXY9a54W2dLZwzPXf1;]"
 		require.Equal(t, expected, strings.TrimSpace(res))
 	})
 	t.Run("SpaceInfo", func(t *testing.T) {
--- a/zqd/handlers_test.go
+++ b/zqd/handlers_test.go
@@ -18,6 +18,7 @@ import (
 	"github.com/brimsec/zq/pkg/test"
 	"github.com/brimsec/zq/zbuf"
 	"github.com/brimsec/zq/zio/detector"
+	"github.com/brimsec/zq/zio/ndjsonio"
 	"github.com/brimsec/zq/zio/zngio"
 	"github.com/brimsec/zq/zqd"
 	"github.com/brimsec/zq/zqd/api"
@@ -39,7 +40,7 @@ func TestSearch(t *testing.T) {
 	defer done()
 	sp, err := client.SpacePost(context.Background(), api.SpacePostRequest{Name: "test"})
 	require.NoError(t, err)
-	_ = postSpaceLogs(t, client, sp.Name, src)
+	_ = postSpaceLogs(t, client, sp.Name, nil, src)
 	res := zngSearch(t, client, sp.Name, "*")
 	require.Equal(t, test.Trim(src), res)
 }
@@ -89,7 +90,7 @@ func TestSpaceInfo(t *testing.T) {
 	defer done()
 	sp, err := client.SpacePost(ctx, api.SpacePostRequest{Name: "test"})
 	require.NoError(t, err)
-	_ = postSpaceLogs(t, client, sp.Name, src)
+	_ = postSpaceLogs(t, client, sp.Name, nil, src)
 	min, max := nano.Ts(1e9), nano.Ts(2e9)
 	expected := &api.SpaceInfo{
 		MinTime:       &min,
@@ -246,7 +247,7 @@ func TestRequestID(t *testing.T) {
 	})
 }
 
-func TestPostLogs(t *testing.T) {
+func TestPostZngLogs(t *testing.T) {
 	src1 := []string{
 		"#0:record[_path:string,ts:time,uid:bstring]",
 		"0:[conn;1;CBrzd94qfowOqJwCHa;]",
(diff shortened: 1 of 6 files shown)


zqd/ingest/log.go
```diff
@@ -48,20 +50,50 @@ func (rw *recWriter) Write(r *zng.Record) error {
...
+			return err
+		}
+		jr, ok := sf.Reader.(*ndjsonio.Reader)
+		if ok && tc != nil {
+			if err = configureJSONTypeReader(jr, *tc, path); err != nil {
+				return err
+			}
+		}
+		readers = append(readers, sf)
+	}
```
nwt, line 85: Nit: We should close the open readers if we encounter an error in this loop.

zqd/handlers_test.go
```diff
@@ -298,11 +299,67 @@ func TestPostLogs(t *testing.T) {
...
+	})
+	t.Run("Post", func(t *testing.T) {
+		payloads := postSpaceLogs(t, client, spaceName, &tc, src1)
+		last := payloads[len(payloads)-1].(*api.TaskEnd)
+		assert.Equal(t, last.Type, "TaskEnd")
+		assert.Nil(t, last.Error)
+	})
+	t.Run("Search", func(t *testing.T) {
+		res := zngSearch(t, client, spaceName, "*")
+		expected := "#0:record[_path:string,ts:time,uid:bstring]\n0:[http;2;CXY9a54W2dLZwzPXf1;]\n0:[http;1;CXY9a54W2dLZwzPXf1;]"
```
nwt, line 341:
```suggestion
		const expected = "#0:record[_path:string,ts:time,uid:bstring]\n0:[http;2;CXY9a54W2dLZwzPXf1;]\n0:[http;1;CXY9a54W2dLZwzPXf1;]"
```

zqd/ingest/log.go
```diff
@@ -48,20 +50,50 @@ func (rw *recWriter) Write(r *zng.Record) error {
 	return nil
 }
 
-func ingestLogs(ctx context.Context, s *space.Space, paths []string, sortLimit int) error {
-	zr, err := scanner.OpenFiles(resolver.NewContext(), paths...)
-	if err != nil {
+// x509.14:00:00-15:00:00.log.gz (open source zeek)
+// x509_20191101_14:00:00-15:00:00+0000.log.gz (corelight)
+const DefaultJSONPathRegexp = `([a-zA-Z0-9_]+)(?:\.|_\d{8}_)\d\d:\d\d:\d\d\-\d\d:\d\d:\d\d(?:[+\-]\d{4})?\.log(?:$|\.gz)`
```
alfred-landrum, line 55:
```suggestion
var DefaultJSONPathRegexp = regexp.MustCompile(`([a-zA-Z0-9_]+)(?:\.|_\d{8}_)\d\d:\d\d:\d\d\-\d\d:\d\d:\d\d(?:[+\-]\d{4})?\.log(?:$|\.gz)`)
```

zqd/ingest/log.go
```diff
@@ -48,20 +50,50 @@ func (rw *recWriter) Write(r *zng.Record) error {
...
+
+func configureJSONTypeReader(ndjr *ndjsonio.Reader, tc ndjsonio.TypeConfig, filename string) error {
+	var path string
+	re := regexp.MustCompile(DefaultJSONPathRegexp)
+	match := re.FindStringSubmatch(filename)
```
alfred-landrum, line 60:
```suggestion
	match := DefaultJSONPathRegexp.FindStringSubmatch(filename)
```

zqd/ingest/log.go
henridf: how come?

henridf:
> Should we plan on trying to commonize the many files/auto-detect/per-type setup that is needed both here & in zq?

@alfred-landrum I'm happy to do that; what held me back so far was i couldn't figure out the place where that should live. If there's a good home that I didn't see, I can do that in this PR.

zqd/handlers_test.go
```diff
@@ -298,11 +299,67 @@ func TestPostLogs(t *testing.T) {
 	})
 }
 
+func TestPostNDJSONLogs(t *testing.T) {
+	src1 := `{"ts":"1000","uid":"CXY9a54W2dLZwzPXf1","_path":"http"}
```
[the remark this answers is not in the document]
henridf: I'm happy to make the change but would also like to learn, as it seems like we usually don't `const`-ify immutable things inside functions - do you have a rough guiding rule for when to const or not to const?

zqd/ingest/log.go
henridf: oh never mind @alfred-landrum - glancing at the change I missed the MustCompile addition, just saw the const->var change. 👍

zqd/handlers_test.go
```diff
@@ -298,11 +299,67 @@ func TestPostLogs(t *testing.T) {
...
+			},
+		},
+		Rules: []ndjsonio.Rule{
+			ndjsonio.Rule{"_path", "http", "http_log"},
+		},
+	}
+	_, client, done := newCore(t)
+	defer done()
+	spaceName := "test"
+	t.Run("Space", func(t *testing.T) {
```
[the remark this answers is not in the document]
henridf: Hmm, good point. This pattern is sprinkled throughout this file, so I think I'll come back to this separately.

zqd/ingest/log.go
```diff
@@ -48,20 +50,50 @@ func (rw *recWriter) Write(r *zng.Record) error {
...
+	if len(match) == 2 {
+		path = match[1]
+	}
+	if err := ndjr.ConfigureTypes(tc, path); err != nil {
 		return err
 	}
-	defer zr.Close()
+	return nil
```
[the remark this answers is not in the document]
henridf: 👍

zqd/ingest/log.go
henridf: hmm, this change conflicts with the string type of `jsonPathRegexp` https://github.com/brimsec/zq/blob/6abf592dc5ea792f67d51948e9913074f1c77df0/cmd/zq/zq.go#L90 which is a string because it is a CLI-settable. I'd rather not add another field to the `main.Command struct`, so have reverted the change.

zqd/handlers_test.go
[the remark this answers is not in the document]
nwt: My guiding rule: If a string won't be modified, make it const.

zqd/handlers_test.go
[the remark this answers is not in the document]
nwt: 👍

zqd/ingest/log.go
henridf: good catch, and sounds like more than a nit to me! fixed.

alfred-landrum:
> @alfred-landrum I'm happy to do that; what held me back so far was i couldn't figure out the place where that should live. If there's a good home that I didn't see, I can do that in this PR.

Ok, let's not block this PR since it's ready; would you file an issue to try to figure out how to make them common?

zqd/handlers_test.go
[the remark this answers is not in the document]
henridf: >  I'll come back to this separately.

done in #583

Issue #525, opened 2020-04-06
Improve sorting used during ingest


philrz:
Brim v0 ensures records are sorted by timestamp by invoking the `sort` processor using a limit:

```
sort -r -limit 10000000 ts
```
We should use a different technique that's efficient & doesn't have a hardcoded maximum record cap.

philrz:
Verified in Brim commit `e9b0840` talking to `zqd` commit `48ed30a`.

It took a bit of waiting, but I was able to import the wrccdc "year1" data set, which is 12 GB of uncompressed Zeek logs, which was turned into 8.8 GB of uncompressed `all.zng`. I was no longer blocked by the 10-million record `sort` limit that used to be in place. Once I was presented with the splash of newest events, I did a `count()` and confirmed that I saw the same count as when running `zq` to count the unsorted data set.

![image](https://user-images.githubusercontent.com/5934157/80731528-e7c42700-8abf-11ea-9702-a2aa16776986.png)

```
~/Downloads/sampledata/wrccdc-year1/zeek-logs$ zq -t "count()" *
#0:record[count:uint64]
0:[96217189;]
```

Thanks @nwt!

PR #527, opened 2020-04-07
Do external sort for large inputs

nwt:
* Modify the sort proc to do an external merge sort when its input size
  (total number of zng.Record.Raw bytes) exceeds a limit of 128 MB.  The
  limit also governs run size.

* Repurpose the existing sort limit setting to specify an alternative
  memory limit in megabytes instead of a record count.

* Remove zqd's sort limit setting since it's no longer needed.

Addresses #525.

I need to add a test that does an external sort before merging, but this is ready for a first look.


zql/docs/processors/README.md
```diff
@@ -171,9 +171,9 @@ ID.ORIG_H     ID.ORIG_P ID.RESP_H       ID.RESP_P ORIG_BYTES RESP_BYTES TOTAL_BY
 |                           |                                                                           |
 | ------------------------- | ------------------------------------------------------------------------- |
 | **Description**           | Sort events based on the order of values in the specified named field(s). | 
-| **Syntax**                | `sort [-r] [-limit N] [-nulls first\|last] [field-list]`                   |
+| **Syntax**                | `sort [-r] [-limit MB] [-nulls first\|last] [field-list]`                 |
```
alfred-landrum, line 174: I think we should remove the limit parameter entirely. In this form, as a user supplied size of memory, it's both dangerous, because a nefarious user could use a very large size to OOM the service, and it's hard to use, since a user won't know how much is too little or too much. 

I think a configurable cutoff for when to shift from memory-only to disk usage does make sense as a config parameter for zqd/zq, though, and that can be memory based, since the operator would know more about the machine where zqd/zq is running.

nwt: 👍

nwt: @alfred-landrum: I replaced the limit parameter with, proc.SortMemMaxBytes, a global variable set by zq in response to `-sortmem` and set by zqd in response to `sortMemMaxBytes` in the config file.

@alfred-landrum and @mattnibs: I'm a little uncomfortable adding a global variable. The only reasonable alternative I see is passing a configuration struct or a configuration function to proc.Compile and proc.CompileProc, but that's a little more complicated than the global variable. Let me know if you think it's worth doing.

cmd/zqd/listen/command.go
```diff
@@ -136,20 +138,28 @@ func pprofHandlers(h http.Handler) http.Handler {
...
 	if err != nil {
 		return err
 	}
 	err = yaml.Unmarshal(b, conf)
 	c.loggerConf = &conf.Logger
+	if v := conf.SortMemMaxBytes; v != nil {
+		if *v <= 0 {
+			return fmt.Errorf("%s: sortMemMaxBytes value must be greater than zero", c.configfile)
+		}
+		proc.SortMemMaxBytes = *v
```
mattnibs, line 161: I suppose this is ok for now, but ideally this would be passed into the zqd's core.Config. I always worry about these global w.r.t running unit tests in parallel.

proc/sort.go
```diff
@@ -1,55 +1,189 @@
...
+}
+
+func (s *Sort) sortLoop() {
+	defer close(s.resultCh)
+	firstRunRecs, eof, err := s.recordsForOneRun()
+	if err != nil || len(firstRunRecs) == 0 {
+		s.sendResult(nil, err)
+		return
+	}
+	s.setSortFn(firstRunRecs[0])
```
mattnibs, line 69: Rather than have to do separate logic for the first in memory "run", I think it would be cool to create something like `zbuf.Peeker` and do something like:
```golang
rec, err := s.peeker.Peek(1)
if err != nil {
    return err
}
s.setSortFn(rec)
```

proc/sort.go
```diff
@@ -93,85 +227,194 @@ func guessSortField(rec *zng.Record) string {
...
+type runManager struct {
+	runs       []*runFile
+	runIndices map[*runFile]int
+	sortFn     expr.SortFn
+	tempDir    string
+	zctx       *resolver.Context
+}
+
+// newRunManager creates a temporary directory.  Call cleanup to remove it.
+func newRunManager(sortFn expr.SortFn) (*runManager, error) {
```
mattnibs, line 240: "Neither do people light a lamp and put it under a bowl. Instead they put it on its stand, and it gives light to everyone in the house." Why hide this function? It would be cool to have this return a `zbuf.Reader`- perhaps this function could even live in zbuf. I imagine the signature could some like:
```golang
func FileMergeSort(ctx context.Context, sortFn expr.SortFn, r zbuf.Reader, maxSize int64) (zbuf.Reader, error)
```

The function would always keep the last record set in memory so if the amount of total records is less then maxSize, no records would ever hit disk.

nwt:
I've rebased this with an eye toward merging it as is (and saving for another day the test, the elimination of the proc.SortMemMaxBytes global variable, and the refactoring suggested by @mattnibs).

henridf:
> I need to add a test that does an external sort before merging, but this is ready for a first look.

Just to be sure, please add this before merging.

nwt:
@henridf: Will do. As @alfred-landrum has shown, a test is definitely needed.

alfred-landrum:
I asked @nwt if the above errors could be related to different record types being split across different runfiles, then merged together by the runManager. He said that seemed likely, and that he'd seen and addressed an earlier version of a similar issue while creating the PR, so he's got a next step for the PR (and an idea for test cases).

nwt:
@alfred-landrum: I think 890c86bb14507a6d628d3fcf472eab939682fffe fixes the bug you encountered. I'm now able to sort an 8.9 GB conn.log.

@henridf: I've added a test.

proc/sort_test.go
```diff
@@ -164,3 +169,37 @@ func TestSort(t *testing.T) {
...
+	var ss []string
+	for n <= 2*proc.SortMemMaxBytes {
+		s := fmt.Sprintf("%016x", rand.Uint64())
+		n += len(s)
+		ss = append(ss, s)
+	}
+	input := makeTzng(ss)
+	sort.Strings(ss)
+	output := makeTzng(ss)
+	(&ztest.ZTest{
```
henridf, line 200: 😮 TIL

proc/sort.go
nwt: This is a good idea, and I'm going to do it in another PR.

cmd/zqd/listen/command.go
nwt: Agreed. There has to be a better way. I'm just not yet sure what that is yet.

proc/sort.go
nwt: @mattnibs: Is part of the idea here that the logic for keeping the first run in memory would live in FileMergeSort, so we'd need another way to get at the first record?

PR #536, opened 2020-04-08
Make searches case-insensitive by default

### LINKED ISSUE #509: Case-insensitive searches
The `zq` engine (and hence the searches performed in the Brim app that rely on it) has up until now treated all searches as case sensitive. So for example:

```
$ cat hostnames.ndjson 
{"hostname": "Facebook.com"}
{"hostname": "facebook.com"}

$ zq -f table "facebook" hostnames.ndjson 
HOSTNAME
facebook.com
```

We recognize that most search tools users are already familiar with are case-insensitive by default, so this issue captures our intent to change our behavior to match that expectation.

Once we flip the default behavior,  we should likely offer a way that users can specify if they want some portion of their search to still be matched in a case-sensitive way.

Whatever approach we offer users to do that should be designed to accommodate a related enhancement we may add in the future (doesn't have to be now, as the demand for it has not been as strong): Case-insensitive aggregations. Revisiting the example above:

```
$ zq -f table "count() by hostname" hostnames.ndjson 
HOSTNAME     COUNT
Facebook.com 1
facebook.com 1
```

We may want to similarly flip the default behavior to get a count of 2 here, and hopefully the syntax we come up to invoke case-sensitivity in the search context could be repurposed here when that becomes a priority.

aswan:
Fixes #509

### DIFF
--- a/filter/compare.go
+++ b/filter/compare.go
@@ -257,6 +257,27 @@ func CompareFloat64(op string, pattern float64) (Predicate, error) {
 	}, nil
 }
 
+// stringSearch is like strings.Contains() but with case-insensitive
+// comparison.
+func stringSearch(a, b string) bool {
+	alen := len(a)
+	blen := len(b)
+
+	if blen > alen {
+		return false
+	}
+
+	end := alen - blen + 1
+	i := 0
+	for i < end {
+		if strings.EqualFold(a[i:blen], b) {
+			return true
+		}
+		i++
+	}
+	return false
+}
+
 var compareString = map[string]func(string, string) bool{
 	"eql":  func(a, b string) bool { return a == b },
 	"neql": func(a, b string) bool { return a != b },
@@ -265,7 +286,7 @@ var compareString = map[string]func(string, string) bool{
 	"lt":   func(a, b string) bool { return a < b },
 	"lte":  func(a, b string) bool { return a <= b },
 	//XXX this doesn't belong here.  primitive comparison vs container operator
-	"search": func(a, b string) bool { return strings.Contains(a, b) },
+	"search": stringSearch,
 }
 
 func CompareBstring(op string, pattern zng.Bstring) (Predicate, error) {
--- a/filter/compare.go
+++ b/filter/compare.go
@@ -270,7 +270,7 @@ func stringSearch(a, b string) bool {
 	end := alen - blen + 1
 	i := 0
 	for i < end {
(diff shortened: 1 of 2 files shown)


aswan:
The implementation here is not at all optimized (though I think there's going to inevitably be some performance hit here).
@philrz I was going to ask to re-run the performance nubmers on this branch, but it looks like we don't cover string search there so I wouldn't expect to see any difference.  I'll try to generate some ad hoc before-after perf numbers for string searching on some of our sample data sets.

aswan:
I'm not seeing a significant difference here in practice on an arbitrary chosen query.

On this branch:
```
(~/src/zq)$ time ./dist/zq "windows | count()" ~/zeek-sample-data/corelight/http.log 
#0:record[count:uint64]
0:[27857;]

real	0m0.476s
user	0m0.532s
sys	0m0.032s
```

On the master revision this branch is based on (df94d6e1723e3e86c8e8d8a55eff8f7ffc09d865):
```
(~/src/zq)$ time ./dist/zq "windows | count()" ~/zeek-sample-data/corelight/http.log 
#0:record[count:uint64]
0:[27857;]

real	0m0.465s
user	0m0.523s
sys	0m0.030s
```

aswan:
As expected, there's a modest perf hit on a query that searches.
On the master rev from which this is based (df94d6e1723e3e86c8e8d8a55eff8f7ffc09d865):
```
time ./dist/zq "windows | count()" ~/zeek-sample-data/corelight/http.log 
#0:record[count:uint64]
0:[2684;]

real	0m0.349s
user	0m0.409s
sys	0m0.028s
```

And on this branch:
```
(~/src/zq)$ time ./dist/zq "windows | count()" ~/zeek-sample-data/corelight/http.log  
#0:record[count:uint64]
0:[27857;]

real	0m0.469s
user	0m0.527s
sys	0m0.031s
```

aswan:
I suspect there's some room to optimize this a bit by having `stringSearch` operate on byte slices instead of taking repeated string slices, but I haven't explored it...

filter/compare.go
```diff
@@ -257,6 +257,27 @@ func CompareFloat64(op string, pattern float64) (Predicate, error) {
...
+
+	end := alen - blen + 1
+	i := 0
+	for i < end {
+		if strings.EqualFold(a[i:i+blen], b) {
+			return true
+		}
+		i++
+	}
+	return false
```
nwt, line 278: Yikes, but let's fix this later.

Issue #493, opened 2020-04-01
Commonize ZQL compilation code


philrz:
The code to parse, compile, and drive a zql query to completion can be found in a few places in slightly different forms. It's most obvious that it's copied between the zq command the zqd search API endpoint, but it's also used in a few test calls as well. We should refactor as needed so that we can put the code in a common place.

PR #540, opened 2020-04-09
Centralize flowgraph compile

henridf:
Move compilation and parsing to a single place, used by zq, tests, and zqd.

That single place is `driver/`, which seems like the best home for it for now (I also considered `proc/` but it seemed slightly off-topic there, and a new `compile/` dir but I don't think this deserves it's own top-level package).

Part of #493 (addresses the compile & parse pieces... refactoring the driver piece will come separately).

### DIFF
--- a/driver/compile.go
+++ b/driver/compile.go
@@ -13,11 +13,17 @@ import (
 	"go.uber.org/zap"
 )
 
-// Compile takes an AST, an input proc, and configuration parameters,
+// Compile takes an AST, an input reader, and configuration parameters,
 // and compiles it into a runnable flowgraph, returning a
 // proc.MuxOutput that which brings together all of the flowgraphs
 // tails, and is ready to be Pull()'d from.
-func Compile(ctx context.Context, program ast.Proc, input proc.Proc, reverse bool, logger *zap.Logger) (*proc.MuxOutput, error) {
+func Compile(ctx context.Context, program ast.Proc, reader zbuf.Reader, reverse bool, span nano.Span, logger *zap.Logger) (*proc.MuxOutput, error) {
+
+	filterAst, program := liftFilter(program)
+	input, err := inputProc(reader, filterAst, span)
+	if err != nil {
+		return nil, err
+	}
 	pctx := &proc.Context{
 		Context:     ctx,
 		TypeContext: resolver.NewContext(),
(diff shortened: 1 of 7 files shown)


driver/compile.go
```diff
@@ -4,21 +4,73 @@ import (
...
 )
 
-func Compile(program ast.Proc, input proc.Proc) (*proc.MuxOutput, error) {
-	ctx := &proc.Context{
-		Context:     context.Background(),
+// Compile takes an AST, an input reader, and configuration parameters,
+// and compiles it into a runnable flowgraph, returning a
+// proc.MuxOutput that which brings together all of the flowgraphs
+// tails, and is ready to be Pull()'d from.
+func Compile(ctx context.Context, program ast.Proc, reader zbuf.Reader, reverse bool, span nano.Span, logger *zap.Logger) (*proc.MuxOutput, error) {
```
henridf, line 20: This function signature is a bit big... it reflects the fact that a bunch of stuff goes into wiring up a flowgraph. 

We could also pass in the `proc.Context` instead of the last three parameters, but that would then mean more prepwork at the call sites.

scanner/scanner.go
```diff
@@ -39,7 +44,7 @@ func (s *Scanner) Read() (*zng.Record, error) {
 		if err != nil || rec == nil {
 			return nil, err
 		}
-		if s.span.Dur != 0 && !s.span.Contains(rec.Ts) ||
+		if s.span != nano.MaxSpan && !s.span.Contains(rec.Ts) ||
```
nwt, line 47:
```suggestion
		if !s.span.Contains(rec.Ts) ||
```

scanner/scanner.go
```diff
@@ -8,21 +8,26 @@ import (
...
 	return &Scanner{
 		reader: reader,
-		filter: f,
+		span:   nano.MaxSpan,
 	}
 }
 
-func (s *Scanner) SetSpan(span nano.Span) {
-	s.span = span
+func NewFilteredScanner(reader zbuf.Reader, f filter.Filter, s nano.Span) *Scanner {
```
nwt, line 25: Nit: I think Scanner should have just have one constructor:
```go
func NewScanner(reader zbuf.Reader, f filter.Filter, s nano.Span) *Scanner {
```

henridf: I had that at one point, then thought it wasn't nice to have all the call sites with nil arguments... but was pretty ambivalent abuot which was better, so I'll switch back to single constructor.

nwt: 👍

Issue #528, opened 2020-04-07
Log ingest error/warning handling


henridf:
Log ingest (the `/space/{space}/log` POST endpoint) needs to be changed to:
- if an input file can't be read/autodetected, send a warning but keep going for the others
- send warnings back from the per-file readers

This is driven by Brim app requirements. More design details will be added to this issue as it comes together.

PR #542, opened 2020-04-09
Move space info logic into space.Space struct

mattnibs:
This is the first of a series of prs aimed simplifying the space api
and moving logic out of the http handlers. Handlers should only be
concerned with translation between the http protocol and the underlying
services and facilities that make up zqd.

### DIFF
--- a/zqd/ingest/pcap.go
+++ b/zqd/ingest/pcap.go
@@ -109,6 +109,7 @@ func (p *Process) run(ctx context.Context) error {
 		os.Remove(p.space.DataPath(PcapIndexFile))
 		os.Remove(p.space.DataPath(space.AllBzngFile))
 		p.space.SetPacketPath("")
+		p.space.UnsetTimes()
 	}
 
 	ticker := time.NewTicker(time.Second)
--- a/zqd/space/space.go
+++ b/zqd/space/space.go
@@ -108,11 +108,7 @@ func (s Space) LogSize() (int64, error) {
 
 // PacketSize returns the size in bytes of the packet capture in the space.
 func (s Space) PacketSize() (int64, error) {
-	path := s.PacketPath()
-	if path == "" {
-		return 0, nil
-	}
-	return sizeof(path)
+	return sizeof(s.PacketPath())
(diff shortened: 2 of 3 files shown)


zqd/ingest/pcap.go
```diff
@@ -109,6 +109,7 @@ func (p *Process) run(ctx context.Context) error {
 		os.Remove(p.space.DataPath(PcapIndexFile))
 		os.Remove(p.space.DataPath(space.AllBzngFile))
 		p.space.SetPacketPath("")
+		p.space.UnsetTimes()
```
mattnibs, line 112: Because this pr added some logic missing in the existing `space.Info`, this had to be added in order to fix unit tests.

henridf:
> This is the first of a series

Sounds good... and fyi that i'm doing some of the same in the context of #493 and #528... should arrive in the next day or so.

zqd/space/space.go
```diff
@@ -76,32 +76,53 @@ func (s Space) Name() string {
...
+	}
+	return sizeof(path)
+}
+
+func sizeof(path string) (int64, error) {
+	f, err := os.Stat(path)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return 0, nil
+		}
```
nwt, line 123:
```suggestion
		}
		return 0, err
```

zqd/space/space.go
```diff
@@ -76,32 +76,53 @@ func (s Space) Name() string {
...
+}
 
-	return spaceInfo, nil
+// PacketSize returns the size in bytes of the packet capture in the space.
+func (s Space) PacketSize() (int64, error) {
+	path := s.PacketPath()
+	if path == "" {
+		return 0, nil
+	}
+	return sizeof(path)
```
nwt, line 115: If you want to, you can remove this check for "" since you check for os.ErrNotExist is sizeof.

zqd/space/space.go
mattnibs: whoops, thanks

PR #545, opened 2020-04-09
Move pcap index search logic into space pkg

mattnibs:
This is part of a series of prs aimed simplifying the space api
and moving logic out of the http handlers. Handlers should only be
concerned with translation between the http protocol and the underlying
services and facilities that make up zqd.

Also:
- Refactor pcap search to make it pull based instead of push based. This was necessary in order to have `space.PcapSearch` return a byte stream that could be copied to the `http.ResponseWriter`

### DIFF
--- a/pcap/search.go
+++ b/pcap/search.go
@@ -138,8 +138,7 @@ func genICMPFilter(src, dst net.IP) PacketFilter {
 
 // XXX need to handle searching over multiple pcap files
 func (s *Search) Run(ctx context.Context, w io.Writer, r pcapio.Reader) error {
-	reader := s.Reader(r)
-	_, err := reader.WriteToWithContext(ctx, w)
+	_, err := s.Reader(r).WriteToWithContext(ctx, w)
 	return err
 }
 
--- a/pcap/search.go
+++ b/pcap/search.go
@@ -151,7 +151,7 @@ type SearchReader struct {
 	buf    []byte
 }
 
-func (s *Search) Reader(r pcapio.Reader) io.Reader {
+func (s *Search) Reader(r pcapio.Reader) *SearchReader {
 	opts := gopacket.DecodeOptions{Lazy: true, NoCopy: true}
 	return &SearchReader{Search: s, reader: r, opts: opts}
 }
@@ -160,7 +160,7 @@ func (s *SearchReader) Read(p []byte) (n int, err error) {
 	if len(s.buf) == 0 {
 		s.buf, err = s.next()
 		if err != nil {
-			return n, err
+			return 0, err
 		}
 		if len(s.buf) == 0 {
 			return 0, io.EOF
--- a/pcap/search.go
+++ b/pcap/search.go
@@ -136,22 +136,63 @@ func genICMPFilter(src, dst net.IP) PacketFilter {
 	}
 }
 
-// XXX currently assumes legacy pcap is produced by the input reader
 // XXX need to handle searching over multiple pcap files
 func (s *Search) Run(ctx context.Context, w io.Writer, r pcapio.Reader) error {
+	reader := s.Reader(r)
+	_, err := reader.WriteToWithContext(ctx, w)
+	return err
+}
+
+type SearchReader struct {
+	*Search
+	reader pcapio.Reader
+	opts   gopacket.DecodeOptions
+	npkt   int
+}
+
+func (s *Search) Reader(r pcapio.Reader) *SearchReader {
 	opts := gopacket.DecodeOptions{Lazy: true, NoCopy: true}
-	var npkt int
+	return &SearchReader{Search: s, reader: r, opts: opts}
+}
+
+func (s *SearchReader) WriteToWithContext(ctx context.Context, w io.Writer) (n int64, err error) {
 	for {
 		if err := ctx.Err(); err != nil {
+			return n, err
+		}
+		block, err := s.next()
+		if err != nil {
+			return n, err
+		}
+		if block == nil {
+			return n, nil
+		}
+		if err := writeFull(w, block); err != nil {
+			return n, err
+		}
+		n += int64(len(block))
+	}
+}
+
+func writeFull(w io.Writer, p []byte) error {
+	for len(p) > 0 {
+		n, err := w.Write(p)
+		if err != nil {
 			return err
 		}
+		p = p[n:]
+	}
+	return nil
+}
 
-		block, typ, err := r.Read()
+func (s *SearchReader) next() ([]byte, error) {
+	for {
+		block, typ, err := s.reader.Read()
 		if err != nil {
 			if err == io.EOF {
 				break
 			}
-			return err
+			return nil, err
 		}
 		if block == nil {
 			break
@@ -164,29 +205,24 @@ func (s *Search) Run(ctx context.Context, w io.Writer, r pcapio.Reader) error {
 		// until we get to the first packet and never writing the blocksa
 		// for sections that have no packets.
 		if typ != pcapio.TypePacket {
-			if _, err = w.Write(block); err != nil {
-				return err
-			}
-			continue
+			return block, nil
 		}
-		pktBuf, ts, linkType := r.Packet(block)
+		pktBuf, ts, linkType := s.reader.Packet(block)
 		if pktBuf == nil {
-			return pcapio.ErrCorruptPcap
+			return nil, pcapio.ErrCorruptPcap
 		}
 		if !s.span.ContainsClosed(ts) {
 			continue
 		}
-		packet := gopacket.NewPacket(pktBuf, linkType, opts)
+		packet := gopacket.NewPacket(pktBuf, linkType, s.opts)
 		if s.filter != nil && !s.filter(packet) {
 			continue
 		}
-		if _, err = w.Write(block); err != nil {
-			return err
-		}
-		npkt++
+		s.npkt++
+		return block, nil
 	}
-	if npkt == 0 {
-		return ErrNoPacketsFound
+	if s.npkt == 0 {
+		return nil, ErrNoPacketsFound
 	}
-	return nil
+	return nil, nil
 }
--- a/pcap/search.go
+++ b/pcap/search.go
@@ -8,6 +8,7 @@ import (
 	"net"
 
 	"github.com/brimsec/zq/pcap/pcapio"
+	"github.com/brimsec/zq/pkg/ctxio"
 	"github.com/brimsec/zq/pkg/nano"
 	"github.com/google/gopacket"
 	"github.com/google/gopacket/layers"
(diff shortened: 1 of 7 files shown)


pcap/search.go
```diff
@@ -136,22 +136,63 @@ func genICMPFilter(src, dst net.IP) PacketFilter {
 	}
 }
 
-// XXX currently assumes legacy pcap is produced by the input reader
 // XXX need to handle searching over multiple pcap files
 func (s *Search) Run(ctx context.Context, w io.Writer, r pcapio.Reader) error {
+	reader := s.Reader(r)
+	_, err := reader.WriteToWithContext(ctx, w)
```
nwt, line 142: Maybe shorten to `_, err := s.Reader(r).WriteToWithContext(ctx, w)`?

zqd/api/api.go
```diff
@@ -200,6 +199,11 @@ func (ps *PacketSearch) FromQuery(v url.Values) error {
 	}
 	ps.Span = span
 	ps.Proto = v.Get("proto")
+	switch ps.Proto {
+	case "tcp", "udp", "icmp":
+	default:
```
nwt, line 204: Nit: `if p := ps.Proto; p != "tcp" && p != "udp" && p != "icmp" {` is one line instead of three.

zqd/handlers.go
```diff
@@ -78,59 +77,25 @@ func handlePacketSearch(c *Core, w http.ResponseWriter, r *http.Request) {
 	}
 	defer cancel()
 
-	req := &api.PacketSearch{}
+	req := api.PacketSearch{}
```
nwt, line 80: Nit: `var req api.PacketSearch` for a zero value.

zqd/space/space.go
```diff
@@ -3,24 +3,30 @@ package space
...
+	infoFile      = "info.json"
+	AllBzngFile   = "all.bzng"
 )
 
 var (
-	ErrSpaceNotExist = errors.New("space does not exist")
-	ErrSpaceExists   = errors.New("space exists")
+	ErrSpaceNotExist       = errors.New("space does not exist")
+	ErrSpaceExists         = errors.New("space exists")
+	ErrPcapOpsNotSupported = errors.New("space does not support pcap operations")
```
nwt, line 29: Nit: Sort me.

zqd/space/space.go
```diff
@@ -3,24 +3,30 @@ package space
...
 )
 
 const (
-	configFile  = "config.json"
-	infoFile    = "info.json"
-	AllBzngFile = "all.bzng"
+	PcapIndexFile = "packets.idx.json"
+	configFile    = "config.json"
+	infoFile      = "info.json"
+	AllBzngFile   = "all.bzng"
```
nwt, line 23: Nit: Sort me.

zqd/space/space.go
```diff
@@ -101,6 +107,47 @@ func (s Space) Info() (api.SpaceInfo, error) {
 	return spaceInfo, nil
 }
 
+// PcapSearch returns a *pcap.SearchReader that streams all the packets meeting
+// the provided search request. If pcaps are not supported in this Space
```
nwt, line 111:
```suggestion
// the provided search request. If pcaps are not supported in this Space,
```

pcap/search.go
```diff
@@ -136,22 +136,63 @@ func genICMPFilter(src, dst net.IP) PacketFilter {
...
+	npkt   int
+}
+
+func (s *Search) Reader(r pcapio.Reader) *SearchReader {
 	opts := gopacket.DecodeOptions{Lazy: true, NoCopy: true}
-	var npkt int
+	return &SearchReader{Search: s, reader: r, opts: opts}
+}
+
+func (s *SearchReader) WriteToWithContext(ctx context.Context, w io.Writer) (n int64, err error) {
```
alfred-landrum, line 158: The naming here doesn't align anymore: SearchReader should become "SearchFilter" or "SearchWriter" to match its Write-like functionality.

You could also pass in the context when a SearchFilter (or whatever) is created, then change this method to match the usual Writer interface.

mattnibs: It actually is more like a `Reader` as it attempts to follow the `WriterTo` interface (see the godocs for `io.Copy`: https://golang.org/pkg/io/#Copy). I would've liked to make SearchReader an `io.Reader` but pkg `io` doesn't have an `io.CopyWithContext` func.

zqd/api/api.go
mattnibs: They're both not great.

pcap/search.go
mattnibs: Sure.

pcap/search.go
nwt: I'm not sure this is an improvement, but you could make this WriterTo and use io.Copy if you wrap the writer with something like this:
```go
type ContextWriter struct {
	ctx context.Context
	w   io.Writer
}

func NewContextWriter(ctx context.Context, w io.Writer) *ContextWriter {
	return &ContextWriter{ctx: ctx, w: w}
}

func (c *ContextWriter) Write(buf []byte) (int, error) {
	if err := c.ctx.Err(); err != nil {
		return 0, err
	}
	return c.w.Write(buf)
}
```

pcap/search.go
```diff
@@ -136,22 +136,62 @@ func genICMPFilter(src, dst net.IP) PacketFilter {
...
+		block, err := s.next()
+		if err != nil {
+			return n, err
+		}
+		if block == nil {
+			return n, nil
+		}
+		if err := writeFull(w, block); err != nil {
+			return n, err
+		}
```
nwt, line 171: Nit: This isn't strictly necessary since the [io.Writer documentation](https://pkg.go.dev/io?tab=doc#Writer) says you'll get a non-nil error from a short Write.

mattnibs: Oh great. Thanks @nwt. Great to get rid of this.

pcap/search.go
mattnibs: @nwt also an interesting approach.

zqd/handlers.go
```diff
@@ -78,59 +77,25 @@ func handlePacketSearch(c *Core, w http.ResponseWriter, r *http.Request) {
...
-	w.Header().Set("Content-Disposition", fmt.Sprintf("inline; filename=%s.pcap", search.ID()))
-	err = search.Run(ctx, w, pcapReader)
+	w.Header().Set("Content-Disposition", fmt.Sprintf("inline; filename=%s.pcap", reader.ID()))
+	_, err = reader.WriteToWithContext(ctx, w)
 	if err != nil {
 		if err == pcap.ErrNoPacketsFound {
 			http.Error(w, err.Error(), http.StatusNotFound)
 		} else {
-			http.Error(w, err.Error(), http.StatusBadRequest)
+			http.Error(w, err.Error(), http.StatusInternalServerError)
```
henridf, line 98: This call to `http.Error()` will be a no-op if the `reader.WriteToWithContext()` call above resulted in any `Write` to the http.ResponseWriter. 

There's a similar issue in the `/search` handler, for which I have a branch that I haven't yet turned into a PR (https://github.com/brimsec/zq/tree/fix-search-error).

The pattern I've used there is that the handler does request validation, returning an HTTP error upon failure. If validation is ok, then the handler calls out to the relevant function (in this case WriteToWithContext) and the HTTP request gets a 2xx back, and  subsequent api protocol messages. If tehre is an error during processing, it is carried in the final `TaskEnd` message.

mattnibs: Yep that's a good point @henridf 

You can't really use `EndTask` here because the content-type has already been set to pcap. Hmm.

We also have a problem with the `pcap.ErrNoPacketsFound` because again the content headers have been set... I think. Also `pcap.Search` will still write pcap section headers even if no results are found. Ugh.

Let me suggest that I file these concerns as a separate ticket and follow up on a separate pr? I think we do need to address this.

henridf: > You can't really use EndTask here because the content-type has already been set to pcap.

Oh right. Addressing this is more complicated than for the search endpoint... so yes please file a separate ticket and it can be done separately.

pcap/search.go
```diff
@@ -136,22 +137,48 @@ func genICMPFilter(src, dst net.IP) PacketFilter {
...
-		if err := ctx.Err(); err != nil {
-			return err
+	return &SearchReader{Search: s, reader: r, opts: opts}
+}
+
+func (s *SearchReader) Read(p []byte) (n int, err error) {
+	if len(s.buf) == 0 {
+		s.buf, err = s.next()
+		if err != nil {
+			return n, err
```
nwt, line 163: Nit: I think `return 0, err` would be clearer.

zqd/api/api.go
nwt: True!

zqd/handlers.go
nwt: Excellent insight, @henridf!

pcap/search.go
```diff
@@ -147,41 +148,27 @@ type SearchReader struct {
 	reader pcapio.Reader
 	opts   gopacket.DecodeOptions
 	npkt   int
+	buf    []byte
 }
 
-func (s *Search) Reader(r pcapio.Reader) *SearchReader {
+func (s *Search) Reader(r pcapio.Reader) io.Reader {
```
alfred-landrum, line 154: This can still return a SearchReader; there's no reason to limit its method set, since SearchReader is exported.

```suggestion
func (s *Search) Reader(r pcapio.Reader) *SearchReader {
```

zqd/space/space.go
```diff
@@ -101,6 +107,51 @@ func (s Space) Info() (api.SpaceInfo, error) {
...
+	pcapReader, err := pcapio.NewReader(slicer)
+	if err != nil {
+		f.Close()
+		return nil, "", err
+	}
+	r := search.Reader(pcapReader)
+	return struct {
+		io.Reader
+		io.Closer
+	}{r, f}, search.ID(), nil
```
nwt, line 152:
```go
struct {
	io.Reader
	io.Closer
}{r, f}
```
Nice!

zqd/space/space.go
```diff
@@ -129,23 +129,27 @@ func (s Space) PcapSearch(req api.PacketSearch) (*pcap.SearchReader, io.Closer,
...
 		f.Close()
-		return nil, nil, err
+		return nil, "", err
 	}
-	return search.Reader(pcapReader), f, nil
+	r := search.Reader(pcapReader)
+	return struct {
+		io.Reader
+		io.Closer
+	}{r, f}, search.ID(), nil
```
alfred-landrum, line 152: I like this approach, would suggest you have it return a 'SearchReadCloser' or something that looks like:
```go
type SearchReadCloser {
  *pcap.SearchReader
   f *os.File
}

func (c *SearchReadCloser) Close() error {
  return c.f.Close()
}
```

With above change, you'll be able to pass just this instance and an error; the code in handlers.go can call `.ID()` on it directly.

pcap/search.go
mattnibs: ok

zqd/space/space.go
mattnibs: sure

zqd/handlers.go
mattnibs: Filed #550 and #551

mattnibs:
> (looks like you've come around to the use of api in space !)

I'll admit I was on the wrong side of history for that argument.

PR #548, opened 2020-04-10
Time index wip

aswan:
This is a rebase to recent master of @mccanne's simple-time-index-rebase branch plus a bit of hacking on it.
It needs testing and a zqd config switch to enable/disable it before merging, but here it is for early feedback.

### DIFF
--- a/pkg/peeker/reader.go
+++ b/pkg/peeker/reader.go
@@ -11,6 +11,7 @@ type Reader struct {
 	buffer []byte
 	cursor []byte
 	eof    bool
+	nread  uint64
 }
 
 var (
@@ -60,7 +61,12 @@ func (r *Reader) fill(min int) error {
 	}
 	r.buffer = r.buffer[:clen]
 	r.cursor = r.buffer
+	r.nread += uint64(clen)
 	return nil
+}
+
+func (r *Reader) Position() uint64 {
+	return r.nread
 }
 
 func (r *Reader) Peek(n int) ([]byte, error) {
--- a/pkg/peeker/reader.go
+++ b/pkg/peeker/reader.go
@@ -28,6 +28,11 @@ func NewReader(reader io.Reader, size, max int) *Reader {
 	}
 }
 
+func (r *Reader) Reset() {
+	r.cursor = r.buffer[:0]
+	r.eof = false
+}
+
 func (r *Reader) fill(min int) error {
 	if min > r.limit {
 		return ErrBufferOverflow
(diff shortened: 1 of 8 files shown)


alfred-landrum:
How about if the index creation happens inside space.OpenFile(all.bzng), and gets cleared in space.CreateFile? 

OpenFile(all.bzng) would become "OpenZng" that returns a zng reader. Inside OpenZng, use the Space struct to see if we have index offsets, and if not, read the whole file there, store the index offsets as a member of the Space struct. CreateFile(all.bzng) becomes "CreateZng" that clears the index.

We're working towards keeping more state in the `Space` struct, so keeping the index there is probably the best place.

aswan:
> How about if the index creation happens inside space.OpenFile(all.bzng), and gets cleared in space.CreateFile?
> 
> OpenFile(all.bzng) would become "OpenZng" that returns a zng reader. Inside OpenZng, use the Space struct to see if we have index offsets, and if not, read the whole file there, store the index offsets as a member of the Space struct. CreateFile(all.bzng) becomes "CreateZng" that clears the index.

Maybe I'm misunderstanding the suggestion but the way this works right now, the first reader gets no indexing benefits so it has to scan the whole file, and it builds the index while doing this.  But there's no extra pass over the file for the purpose of constructing the index.  I think that's a useful optimization but it would be more awkward to manage with the code to manage indexes outside of index.go.

alfred-landrum:
> Maybe I'm misunderstanding the suggestion but the way this works right now, the first reader gets no indexing benefits so it has to scan the whole file, and it builds the index while doing this. But there's no extra pass over the file for the purpose of constructing the index. I think that's a useful optimization but it would be more awkward to manage with the code to manage indexes outside of index.go.

Ok, putting aside the extra pass: I do think the index should be stored as a field on the `Space` struct, and not as in a package level map in bzngio. That's where it will eventually be managed, when we want to persist it along with other indices on the space.

aswan:
Rebased onto @mccane's bzng stream io branch.  Still has some obsolete names that need to be cleaned up.  Also, this touches zqd space and search in a way that might conflict with ongoing cleanup that @henridf and/or @mattnibs are doing, I can separate some of the refactoring there if it looks like we're headed for a collision...

mccane:
> Rebased onto @mccane's bzng stream io branch. Still has some obsolete names that need to be cleaned up. Also, this touches zqd space and search in a way that might conflict with ongoing cleanup that @henridf and/or @mattnibs are doing, I can separate some of the refactoring there if it looks like we're headed for a collision...

@aswan I think you meant @mccanne, not @mccane ...

PR #549, opened 2020-04-10
Driver refactor

### LINKED ISSUE #493: Commonize ZQL compilation code

henridf:
This change introduces a new `driver.Driver` interface consisting of methods needed to handle the output of running a flowgraph. It is derived from the existing driver.Driver struct, which is now `driver.CLI`. In addition, there is now a single  func to run a flowgraph (`driver.Run(..)`), instead of the two slightly different ones we had before.

All places that execute flowgraphs are ported to use a `driver.Driver` and execute the flowgraph via `driver.Run()`. This includes search, ingest, zq, and tests.

closes #493


henridf:
PS: When writing the description, I realized that `driver.Driver` doesn't actually drive a flowgraph, it's `Run()` that does. This terminology predates this PR but it's easy to change here. I'd propose `driver.Driver` -> `driver.Output`.

alfred-landrum:
I like this direction, and a name change for `driver.Driver` makes sense. For a separate PR, I wonder if`MuxOutput` should move into the driver package, as it doesn't appear to be used except to create via Compile and then passed to Run.

henridf:
> I wonder if MuxOutput should move into the driver package

I think that would be a good move; it's also a little out of place in `proc/` as it doesn't implement the `proc.Proc` interface

zqd/ingest/driver.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+// driver implements driver.Driver
+type driver struct {
+	pipe      *api.JSONPipe
+	startTime nano.Ts
+	writers   []zbuf.Writer
+}
+
+func (d *driver) Write(cid int, arr zbuf.Batch) error {
```
nwt, line 16: Nit: batch would be a better name than arr.

zqd/ingest/driver.go
```diff
@@ -0,0 +1,41 @@
...
+}
+func (d *driver) Stats(stats api.ScannerStats) error {
+	return nil
+}
+
+func (d *driver) ChannelEnd(cid int, stats api.ScannerStats) error {
+	if err := d.Stats(stats); err != nil {
+		return err
+	}
+	return nil
```
nwt, line 40:
```suggestion
	return d.Stats(stats)
```

zqd/ingest/log.go
```diff
@@ -36,11 +41,14 @@ func Logs(ctx context.Context, s *space.Space, paths []string, tc *ndjsonio.Type
 	if sortLimit == 0 {
 		sortLimit = DefaultSortLimit
 	}
-	if err := ingestLogs(ctx, s, paths, tc, sortLimit); err != nil {
+
+	pipe.Send(&api.TaskStart{"TaskStart", 0})
```
nwt, line 45: Handle errors.

zqd/ingest/log.go
```diff
@@ -111,5 +120,36 @@ func ingestLogs(ctx context.Context, s *space.Space, paths []string, tc *ndjsoni
 			return err
 		}
 	}
-	return os.Rename(bzngfile.Name(), s.DataPath(space.AllBzngFile))
+	if err = os.Rename(bzngfile.Name(), s.DataPath(space.AllBzngFile)); err != nil {
```
nwt, line 123:
```suggestion
	if err := os.Rename(bzngfile.Name(), s.DataPath(space.AllBzngFile)); err != nil {
```

driver/driver.go
```diff
@@ -0,0 +1,104 @@
...
+				err := d.sendStats(out.Stats())
+				if err != nil {
+					return d.abort(0, err)
+				}
+				*/
+				continue
+			}
+			if chunk.Err == context.Canceled {
+				out.Drain()
+				return errors.New("canceled")
```
nwt, line 41:
```suggestion
```

driver/driver.go
```diff
@@ -0,0 +1,104 @@
...
+
+	"github.com/brimsec/zq/proc"
+	"github.com/brimsec/zq/zbuf"
+	"github.com/brimsec/zq/zqd/api"
+)
+
+type Driver interface {
+	Warn(msg string) error
+	Write(int, zbuf.Batch) error
+	ChannelEnd(int, api.ScannerStats) error
```
nwt, line 18:
```suggestion
	Write(channelID int, batch zbuf.Batch) error
	ChannelEnd(channelID int, batch api.ScannerStats) error
```
For documentation.

zqd/search/search.go
```diff
@@ -21,14 +21,15 @@ import (
...
 	"github.com/brimsec/zq/zqd/space"
-	"github.com/brimsec/zq/zql"
 	"go.uber.org/zap"
 )
 
 // This mtu is pretty small but it keeps the JSON object size below 64kb or so
 // so the recevier can do reasonable, interactive streaming updates.
 const DefaultMTU = 100
 
+const DefaultStatsInterval = time.Millisecond * 500
```
nwt, line 31:
```suggestion
const StatsInterval = time.Millisecond * 500
```
Since this is an actual value, not a default value.

zqd/search/search.go
```diff
@@ -129,18 +127,19 @@ func (d *driver) abort(id int64, err error) error {
...
+func (d *driver) Warn(warning string) error {
 	v := api.SearchWarnings{
 		Type:     "SearchWarnings",
 		Warnings: []string{warning},
 	}
 	return d.output.SendControl(v)
 }
 
-func (d *driver) sendStats(stats api.ScannerStats) error {
+func (d *driver) Write(cid int, arr zbuf.Batch) error {
```
nwt, line 138: Nit: batch would be a better name than arr.

driver/driver.go
```diff
@@ -0,0 +1,104 @@
...
+	return &CLI{
+		writers: w,
+	}
+}
+
+func (d *CLI) SetWarningsWriter(w io.Writer) {
+	d.warnings = w
+}
+
+func (d *CLI) Write(cid int, arr zbuf.Batch) error {
```
nwt, line 81: Nit: batch would be a better name than arr.

zqd/ingest/log.go
```diff
@@ -8,22 +8,27 @@ import (
 	"path/filepath"
 	"regexp"
 
+	zdriver "github.com/brimsec/zq/driver"
```
nwt, line 11:
```suggestion
	"github.com/brimsec/zq/driver"
```

zqd/ingest/pcap.go
```diff
@@ -12,6 +12,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	zdriver "github.com/brimsec/zq/driver"
```
nwt, line 15:
```suggestion
	driver "github.com/brimsec/zq/driver"
```

zqd/search/search.go
```diff
@@ -65,20 +66,16 @@ func Search(ctx context.Context, s *space.Space, req api.SearchRequest, out Outp
...
-	if err != nil {
-		return err
+	d := &driver{
+		output:    out,
+		startTime: nano.Now(),
 	}
-	mux, err := zdriver.Compile(ctx, p, r, false, nano.MaxSpan, zap.NewNop())
-	if err != nil {
+	d.start(0)
+	if err := zdriver.Run(mux, d, DefaultStatsInterval); err != nil {
```
nwt, line 74:
```suggestion
	if err := driver.Run(mux, d, DefaultStatsInterval); err != nil {
```
(And undo the package rename to zdriver in the imports.)

zqd/ingest/log.go
```diff
@@ -111,5 +120,36 @@ func ingestLogs(ctx context.Context, s *space.Space, paths []string, tc *ndjsoni
...
+		return err
+	}
+	status := api.LogPostStatus{
+		Type:    "LogPostStatus",
+		MinTime: info.MinTime,
+		MaxTime: info.MaxTime,
+		Size:    info.Size,
+	}
+	pipe.Send(status)
+	return nil
```
nwt, line 137:
```suggestion
	return pipe.Send(status)
```

zqd/ingest/pcap.go
```diff
@@ -262,6 +265,21 @@ func (p *Process) createSnapshot(ctx context.Context) error {
...
+	mux, err := zdriver.Compile(ctx, proc, r, false, nano.MaxSpan, zap.NewNop())
+	if err != nil {
+		return err
+	}
+	d := &driver{
+		startTime: nano.Now(),
+		writers:   []zbuf.Writer{w},
+	}
+	return zdriver.Run(mux, d, search.DefaultStatsInterval)
+}
```
mattnibs, line 282:
```suggestion
}

```

zqd/ingest/log.go
```diff
@@ -98,7 +106,8 @@ func ingestLogs(ctx context.Context, s *space.Space, paths []string, tc *ndjsoni
 	zw := bzngio.NewWriter(bzngfile, zio.Flags{})
 	program := fmt.Sprintf("sort -limit %d -r ts | (filter *; head 1; tail 1)", sortLimit)
 	var headW, tailW recWriter
-	if err := search.Copy(ctx, []zbuf.Writer{zw, &headW, &tailW}, reader, program); err != nil {
+	err = runLogIngest(ctx, s, reader, program, pipe, []zbuf.Writer{zw, &headW, &tailW}...)
```
mattnibs, line 109:
```suggestion
	err = runLogIngest(ctx, s, reader, program, pipe, zw, &headW, &tailW)
```

driver/driver.go
henridf: good point

zqd/ingest/log.go
henridf: 👍

zqd/ingest/log.go
henridf: done

zqd/search/search.go
henridf: 👍

zqd/search/search.go
henridf: 👍 (renamed the driver struct to searchdriver)

zqd/ingest/pcap.go
henridf: not sure how zdriver got here... gone

zqd/ingest/log.go
henridf: renamed driver struct to logdriver

zqd/ingest/log.go
henridf: 👍

driver/driver.go
```diff
@@ -0,0 +1,102 @@
...
+
+	"github.com/brimsec/zq/proc"
+	"github.com/brimsec/zq/zbuf"
+	"github.com/brimsec/zq/zqd/api"
+)
+
+type Driver interface {
+	Warn(msg string) error
+	Write(channelID int, batch zbuf.Batch) error
+	ChannelEnd(channelID int, stats api.ScannerStats) error
```
mattnibs, line 17: I don't think adding the `api.ScannerStats` is necessary here. They aren't/won't be specific to that channel anyways.

driver/driver.go
```diff
@@ -0,0 +1,102 @@
...
+	Stats(api.ScannerStats) error
+}
+
+func Run(out *proc.MuxOutput, d Driver, statsInterval time.Duration) error {
```
mattnibs, line 21: nit: This would be more work, but it would be nice if when `statsInterval` is set to 0 a Noop Ticker is used. I think this would make clearer a lot of places in the code where you set the interval to an arbitrarily long interval to serve as functionally the same thing.

driver/driver.go
henridf: This is just a transcription of existing search behavior. I agree with your point (and wondered why this is this way) but this PR isn't for changing the search protocol... we should change this in a follow-up. (And in general, stats/warnings/feedback is going to get a bunch of updates with the forthcoming work on #528).

driver/driver.go
henridf: Yes, i didn't like passing this arbitrary interval around. I'll change so that 0 means no tick.

henridf: Actually we need a ticker to pass to `Pull()`, we can't not tick. I've changed instead so that if `0` is passed, we switch to an (arbitrary) value of 10 seconds. 
This is now the same behavior as before this PR. This can probably be improved (in a separate PR).

nwt: Maybe pass a `<-chan time.Time` instead of a `time.Duration`? Then callers that don't want a `time.Ticker` can pass nil, and those that do can create one and pass its channel.

henridf: @nwt That seems like a good approach but could you explain what the benefit of passing `nil` as a zero value vs `0` would be? Happy to do it but hoping to learn something here...

nwt: It just strikes me as more intuitive. It's clear that a nil `statsTicks <-chan time.Time` never ticks, but the meaning of a zero `statsInterval time.Duration` isn't obvious.

driver/driver.go
```diff
@@ -0,0 +1,105 @@
...
+		} else {
+			if err := d.Write(chunk.ID, chunk.Batch); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// CLI implements Driver
```
nwt, line 66:
```suggestion
// CLI implements Driver.
```

zqd/ingest/driver.go
```diff
@@ -0,0 +1,38 @@
+package ingest
+
+import (
+	"github.com/brimsec/zq/pkg/nano"
+	"github.com/brimsec/zq/zbuf"
+	"github.com/brimsec/zq/zqd/api"
+)
+
+// logdriver implements driver.Driver
```
nwt, line 9:
```suggestion
// logdriver implements driver.Driver.
```

zqd/search/json.go
```diff
@@ -9,6 +9,7 @@ import (
 	"github.com/brimsec/zq/zqd/api"
 )
 
+// JSON implements the Output interface
```
nwt, line 12:
```suggestion
// JSON implements the Output interface.
```

zqd/search/search.go
```diff
@@ -111,36 +108,38 @@ func UnpackQuery(req api.SearchRequest) (*Query, error) {
 	}, nil
 }
 
-type driver struct {
+// searchdriver implements driver.Driver
```
nwt, line 111:
```suggestion
// searchdriver implements driver.Driver.
```

driver/driver.go
```diff
@@ -0,0 +1,105 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+func (d *CLI) Warn(msg string) error {
+	if d.warnings != nil {
+		_, err := fmt.Fprintln(d.warnings, msg)
+		if err != nil {
```
nwt, line 97:
```suggestion
		if _, err := fmt.Fprintln(d.warnings, msg); err != nil {
```

driver/driver.go
henridf: It's obvious that a nil `<- chan time.Time` never ticks, but upon getting such a nil the `Run()` func would create a `time.NewTicker(10 * time.Second)` under the hood, and use those ticks... which is why  `nil` and `0` seem equally non-obvious to me. (This is all just pre-existing behavior, which we should probably change, but I'm not trying to do that as part of this PR).

Hope this clarifies things. Either way, I can change to this if you'd like; just let me know.

nwt: I was assuming `driver.Run` would always use `statsTicks` directly and would never create its own `timer.Ticker`.

nwt: And for the record, I'm fine with either `time.Duration` or `<-chan time.Time`.

henridf: Passing a nil ticker to `(*MuxOutput).Pull(..)` would be a change to the existing behavior of zq and tests: https://github.com/brimsec/zq/blob/d7c70735c855523588207db22a286f6bbde380bb/driver/run.go#L55

(Not saying this change shouldn't be made! but not here, especially as it breaks some tests)

henridf: > And for the record, I'm fine with either time.Duration or <-chan time.Time.

@nwt 👍. I filed #554 to revisit passing a nil ticker to `(*MuxOutput).Pull(..)`.

driver/driver.go
henridf: Filed https://github.com/brimsec/zq/issues/555 to track this.

alfred-landrum:
> > I wonder if MuxOutput should move into the driver package
> 
> I think that would be a good move; MuxOutput is also a little out of place in `proc/` as it doesn't implement the `proc.Proc` interface

Filed https://github.com/brimsec/zq/issues/559 to track.

PR #553, opened 2020-04-11
zqe errors package

alfred-landrum:
This is a proposal for an errors package for our zq repo, that makes error creation ergnomic and makes converting to api response bodies and status codes straightforward. It's based on the upspin errors package:
https://upspin.googlesource.com/upspin/+/master/errors/errors.go

An error is created via the `E()` function, which takes a mix of defined error Kinds (to facilitate translation to status codes), existing errors, or string formatting directives:

```go
err0 := zqe.E("something bad happened")
err1 := zqe.E("bad thing: %v", foo)
err2 := zqe.E(zqe.Invalid, existingError)
err3 := zqe.E(zqe.Conflict)
err4 := zqe.E(zqe.NotFound, "space %s not found", space)
```
The created error supports errors.Is and errors.As functions. String formatting is done via `fmt.Errorf`, to support the `%w` verb.


errors/errors.go
```diff
@@ -0,0 +1,99 @@
+package zqe
```
mattnibs, line 1: I've never seen this before- the pkg name differing from the folder name. Does this mean that goimports will automatically do the right thing?

mattnibs:
I didn't deep into a review but I like this approach. One thing I'm wondering is if this package should live as a subpackage to `zqd`? Would you want packages no directly related to `zqd` use this? Except for `pcap` all other packages appear to be in `zqd`. As an aside I am wondering if `pcap` does not also belong inside `zqd` since `zq` does not use it.

alfred-landrum:
> I didn't deep into a review but I like this approach. One thing I'm wondering is if this package should live as a subpackage to `zqd`? Would you want packages no directly related to `zqd` use this? Except for `pcap` all other packages appear to be in `zqd`. As an aside I am wondering if `pcap` does not also belong inside `zqd` since `zq` does not use it.

I'd like to see it used throughout the zq repo, as any tool we build using our packages (like zqd or the zq cli today, or future tools) could look at a received error, and check if it's a zqe.Error, in order to understand how to present an informational error to the user.

A feature that isn't present in this PR, but would be trivial to add, is collecting a stack trace during the call to E(), perhaps only enabled via a dev/debug flag.

zqe/error.go
```diff
@@ -0,0 +1,114 @@
...
+// these into a domain specific error representation; for example, an http
+// handler can convert these to http specific status codes.
+type Kind int
+
+const (
+	Other Kind = iota
+	Invalid
+	NotFound
+	Exists
+	Conflict
```
nwt, line 21:
```suggestion
	Conflict
	Exists
	Invalid
	NotFound
```

zqe/error.go
```diff
@@ -0,0 +1,114 @@
...
+
+import (
+	"bytes"
+	"fmt"
+	"runtime"
+)
+
+// A Kind represents a class of error. API layers will typically convert
+// these into a domain specific error representation; for example, an http
+// handler can convert these to http specific status codes.
```
nwt, line 13:
```suggestion
// these into a domain specific error representation; for example, an HTTP
// handler can convert these to HTTP status codes.
```

zqe/error.go
```diff
@@ -0,0 +1,114 @@
...
+	case Other:
+		return "other error"
+	case Invalid:
+		return "invalid operation"
+	case NotFound:
+		return "item does not exist"
+	case Exists:
+		return "item already exists"
+	case Conflict:
+		return "conflict with pending operation"
```
nwt, line 35:
```suggestion
	case Other:
		return "other error"
	case Conflict:
		return "conflict with pending operation"
	case Invalid:
		return "invalid operation"
	case Exists:
		return "item already exists"
	case NotFound:
		return "item does not exist"
```

zqe/error.go
```diff
@@ -0,0 +1,114 @@
...
+
+func pad(b *bytes.Buffer, s string) {
+	if b.Len() == 0 {
+		return
+	}
+	b.WriteString(s)
+}
+
+func (e *Error) Error() string {
+	b := &bytes.Buffer{}
```
nwt, line 53: Nits: Might as well use strings.Builder. And there's no reason to make this a pointer.

zqe/error.go
```diff
@@ -0,0 +1,114 @@
...
+		case Kind:
+			e.Kind = arg
+		case error:
+			e.Err = arg
+		case string:
+			e.Err = fmt.Errorf(arg, args[i+1:]...)
+			return e
+		default:
+			_, file, line, _ := runtime.Caller(1)
+			return fmt.Errorf("unknown type %T value %v in errors.E call at %v:%v", arg, arg, file, line)
```
nwt, line 109:
```suggestion
			return fmt.Errorf("unknown type %T value %v in zqe.E call at %v:%v", arg, arg, file, line)
```

zqd/handlers_zeek_test.go
```diff
@@ -229,6 +229,7 @@ func (r *packetPostResult) postPcap(t *testing.T, file string) {
 	if r.err == nil {
 		r.readPayloads(t, stream)
 	}
+
```
mattnibs, line 232: Was this on purpose?

errors/errors.go
```diff
@@ -0,0 +1,99 @@
...
+	if b.Len() == 0 {
+		return
+	}
+	b.WriteString(s)
+}
+
+func (e *Error) Error() string {
+	b := &bytes.Buffer{}
+	if e.Kind != Other {
+		pad(b, ": ")
```
henridf: Is this `pad()` needed?

errors/errors.go
```diff
@@ -0,0 +1,99 @@
...
+		return "item does not exist"
+	case Exists:
+		return "item already exists"
+	case Conflict:
+		return "conflict with pending operation"
+	}
+	return "unknown error kind"
+}
+
+type Error struct {
```
henridf: Would you consider also adding a `Is()` method (https://godoc.org/errors#Is) to `Error`?

henridf: Which I haven't thought about in depth, but maybe it would look `Kind`s first if available, and otherwise call the `Err.Is()`, if available. 

Or maybe instead clients who want this should just call `errors.Is(e.Err)` directly?

zqd/handlers_zeek_test.go
alfred-landrum: nope, leftover from debugging, will remove

errors/errors.go
alfred-landrum: pad() checks if b.len() is zero, so it's to keep in.

errors/errors.go
alfred-landrum: I put in an `Is(target error) bool` method that only checks if the target is also a zqe.Error & if their Kind's are the same. I didn't go beyond that for now, as that seemed like it would fit a common pattern of:

```go
  err := doSomething()
  if errors.Is(err, zqe.E(zqe.Exists)) {
    // already there
  }
```

and if we wanted to check additional cases, you could add more errors.Is calls:

```go
  err := doSomething()  // returns zqe.E(zqe.NotExist, os.ErrClosed)
  if errors.Is(err, zqe.E(zqe.Exists)) {
    // ...
  } else if errors.Is(err, os.ErrClosed) {
   // this branch will executed
  }
```

PR #561, opened 2020-04-13
Fix error handling in search endpoint handler

### LINKED ISSUE #556: zqd: http error handling issues in /search endpoint
The search handler calls `http.Error()` after having possibly written to its responsewriter, which triggers sending a 200. In that case, the subsequent `http.Error()` is a no-op and no error is sent to the client.

This could be addressed by having the handler do upfront request validation, returning an HTTP error upon failure. If the request  is valid, then the handler launches the search and the HTTP request gets a 2xx back, after which api protocol messages can be sent to the client. If there is an error during search processing, it is sent back to the client in the final TaskEnd message.

henridf:
Previously, the search endpoint could call http.Error() after
having sent data. This commit fixes that by breaking the search action
into a setup phase and a run phase.

If the setup phase fails, an HTTP error is returned. If the setup
phase succeeds, a 200 is returned, and  any error encountered during
the execution of the search will result in a TaskEnd message with the
error being sent to the client.

This PR should not change existing behavior, but it removes a possible no-op call to `http.Error()`, and makes the handler code better reflect the above.

closes #556

### DIFF
--- a/zqd/handlers.go
+++ b/zqd/handlers.go
@@ -43,6 +43,16 @@ func handleSearch(c *Core, w http.ResponseWriter, r *http.Request) {
 	}
 	defer cancel()
 
+	// XXX This always returns bad request but should return status codes
+	// that reflect the nature of the returned error.
+	w.Header().Set("Content-Type", "application/ndjson")
+	srch, err := search.NewSearch(ctx, s, req)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+	defer srch.Close()
+
 	var out search.Output
 	format := r.URL.Query().Get("format")
 	switch format {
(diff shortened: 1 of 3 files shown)


zqd/handlers.go
```diff
@@ -43,6 +43,16 @@ func handleSearch(c *Core, w http.ResponseWriter, r *http.Request) {
 	}
 	defer cancel()
 
+	// XXX This always returns bad request but should return status codes
+	// that reflect the nature of the returned error.
+	w.Header().Set("Content-Type", "application/ndjson")
```
nwt, line 48: Shouldn't this happen only if http.Error isn't called?

henridf: good catch, thanks!

PR #564, opened 2020-04-13
fix typo causing master breakage

mccanne:


mccanne:
fix typo causing master breakage

PR #569, opened 2020-04-13
bzngio API adjustments for stream boundaries

mccanne:
This PR adds hooks the bzngio API to allow for querying and seeking to stream boundaries.  This will be used by zdx and the bzngio time index.

### DIFF
--- a/zio/bzngio/seeker.go
+++ b/zio/bzngio/seeker.go
@@ -11,7 +11,11 @@ type Seeker struct {
 	seeker io.ReadSeeker
 }
 
-func NewSeeker(s io.ReadSeeker, zctx *resolver.Context, framesize int) *Seeker {
+func NewSeeker(s io.ReadSeeker, zctx *resolver.Context) *Seeker {
+	return NewSeekerWithSize(s, zctx, ReadSize)
+}
+
+func NewSeekerWithSize(s io.ReadSeeker, zctx *resolver.Context, framesize int) *Seeker {
 	return &Seeker{
 		Reader: *NewReaderWithSize(s, zctx, framesize),
 		seeker: s,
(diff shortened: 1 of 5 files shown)


zio/bzngio/seeker.go
```diff
@@ -0,0 +1,27 @@
...
+	return &Seeker{
+		Reader: *NewReaderWithSize(s, zctx, framesize),
+		seeker: s,
+	}
+}
+
+func (s *Seeker) Seek(offset int64) (int64, error) {
+	s.peeker.Reset()
+	s.zctx.Reset()
+	n, err := s.seeker.Seek(offset, 0)
```
aswan, line 24: i think this should be `io.SeekStart` for clarity

zio/bzngio/seeker.go
```diff
@@ -0,0 +1,27 @@
...
+
+	"github.com/brimsec/zq/zng/resolver"
+)
+
+type Seeker struct {
+	Reader
+	seeker io.ReadSeeker
+}
+
+func NewSeeker(s io.ReadSeeker, zctx *resolver.Context, framesize int) *Seeker {
```
aswan, line 14: why does the caller need to pass a buffer size here?  that's a change from how existing readers work.

zio/bzngio/writer.go
```diff
@@ -26,12 +27,22 @@ func NewWriter(w io.Writer, flags zio.Flags) *Writer {
 	}
 }
 
+func (w *Writer) write(b []byte) (int, error) {
```
aswan, line 30: it looks like every single caller ignores the first return value, this can return just the error?

zio/bzngio/seeker.go
mccanne: cool

mccanne:
Working on a test right now...

zio/bzngio/seeker.go
mccanne: I will add NewSeekerWithSize().   Zdx wants to control the read buffer size.

PR #570, opened 2020-04-13
Improve string search matching on field names

### LINKED ISSUE #511: Bare searches should match on field names
Currently bare searches match on field values, but not field names. For example:

```
$ echo '{"foo": "bar"}' | zq "bar" -
#0:record[foo:string]
0:[bar;]

$ echo '{"foo": "bar"}' | zq "foo" -
[no output]
```

Other search tools that users are accustomed to using would often match in the second case as well. To meet user expectations, we should do the same.

It seems like once we do this, we already have a syntax for users to express a desire to match _only_ on field values. In this example it would be:

```
$ echo '{"foo": "bar"}' | zq "*=bar" -
#0:record[foo:string]
0:[bar;]
```

aswan:
- Match sub-strings in field names
- Match fully qualified field names (e.g., id.resp_h)
- Do case insensitive matching

(Re-)fixes #511

### DIFF
--- a/filter/filter_test.go
+++ b/filter/filter_test.go
@@ -299,8 +299,16 @@ func TestFilters(t *testing.T) {
 		{"ts<2.0", true},
 		{"ts2=1578411532", true},
 		{"ts3=1578411533", false},
-		// The ASCII value of 'T' (0x54) is present inside the binary
-		// encoding of 1.001.  But naked string search should not match.
+	})
+
+	// Test that string search doesn't match non-string types:
+	// The ASCII value of 'T' (0x54) is present inside the binary
+	// encoding of 1.001.  But naked string search should not match.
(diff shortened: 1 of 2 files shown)


filter/filter_test.go
```diff
@@ -441,11 +449,14 @@ func TestFilters(t *testing.T) {
...
-#0:record[foo:string]
-0:[bleah;]`)
+#0:record[foo:string,rec:record[sub:string]]
+0:[bleah;[meh;]]`)
 	require.NoError(t, err)
 	runCases(t, record, []testcase{
 		{"foo", true},
+		{"FOO", true},
+		{"sub", true},
+		{"rec.sub", true},
```
henridf, line 459: might be nice to add `{"c.s", true}` to the list of tests

PR #579, opened 2020-04-14
Time indexing

aswan:
This includes both the implementation of an in-memory zng index, and its integration into zqd.  These can be split into separate PRs if that's desired.

### DIFF
--- a/zio/bzngio/index.go
+++ b/zio/bzngio/index.go
@@ -1,0 +1,183 @@
+package bzngio
+
+import (
+	"errors"
+	"io"
+	"os"
+
+	"github.com/brimsec/zq/pkg/nano"
+	"github.com/brimsec/zq/zbuf"
+	"github.com/brimsec/zq/zng"
+	"github.com/brimsec/zq/zng/resolver"
+)
+
+type Index struct {
+	index      []mark
+	indexReady bool
+}
+
+type IndexReader interface {
+	zbuf.ReadCloser
+	// count of how many records were read from disk, currently just
+	// used for testing.
+	Reads() uint64
+}
+
+type mark struct {
+	Ts     nano.Ts
+	Offset int64
+}
+
+// NewIndex creates a new Index object, which is the container that holds
+// the in-memory index for a (b)zng file.  The first call to NewReader()
+// will return a reader that scans the entire file, building a time-based
+// index in the process, subsequent readers can use this index to read
+// only the relevant zng streams from the underlying file.
+func NewIndex() Index {
+	return Index{}
+}
+
+func (i *Index) NewReader(f *os.File, zctx *resolver.Context, span nano.Span) (IndexReader, error) {
+	if i.indexReady {
+		return newRangeReader(f, zctx, i.index, span)
+	} else {
+		return &indexReader{
+			Reader: *NewReader(f, zctx),
+			closer: f,
+			start:  span.Ts,
+			end:    span.End(),
+			parent: i,
+		}, nil
+	}
+}
+
+// indexReader is a zbuf.Reader that also builds an index as it reads.
+type indexReader struct {
+	Reader
+	closer  io.Closer
+	start   nano.Ts
+	end     nano.Ts
+	parent  *Index
+	marks   []mark
+	lastSOS int64
+	lastTs  nano.Ts
+}
+
+func (i *indexReader) Read() (*zng.Record, error) {
+	for {
+		rec, err := i.readOne()
+		if rec == nil {
+			i.parent.index = i.marks
+			i.parent.indexReady = true
+			return nil, nil
+		}
+
+		if err != nil {
+			return nil, err
+		}
+
+		if rec.Ts < i.start {
+			continue
+		}
+		if rec.Ts <= i.end {
+			return rec, nil
+		}
+
+		// This record falls after the end of the requested time
+		// span.  Spin through this loop until we hit EOF anyway
+		// to finish building the index.
+		// XXX this will be wasteful if small ranges near the
+		// start of the file is all that is ever read.  revisit this...
+	}
+}
+
+func (i *indexReader) readOne() (*zng.Record, error) {
+	rec, err := i.Reader.Read()
+	if err != nil || rec == nil {
+		return nil, err
+	}
+
+	sos := i.Reader.LastSOS()
+	if sos != i.lastSOS {
+		i.lastSOS = sos
+		ts := rec.Ts
+		if ts > i.lastTs {
+			i.lastTs = ts
+			i.marks = append(i.marks, mark{ts, sos})
+		}
+	}
+
+	return rec, nil
+}
+
+func (i *indexReader) Close() error {
+	return i.closer.Close()
+}
+
+// this is only used for testing and only called on rangeReader
+func (i *indexReader) Reads() uint64 {
+	return 0
+}
+
+// rangeReader is a wrapper around bzngio.Reader that uses an in-memory
+// index to reduce the I/O needed to get matching records when reading a
+// large bzng file that includes sub-streams and a nano.Span that refers
+// to a smaller time range within the file.
+type rangeReader struct {
+	Reader
+	closer io.Closer
+	start  nano.Ts
+	end    nano.Ts
+	nread  uint64
+}
+
+func newRangeReader(f *os.File, zctx *resolver.Context, index []mark, span nano.Span) (*rangeReader, error) {
+	var off int64
+	// XXX binary search
+	for _, mark := range index {
+		if mark.Ts > span.Ts {
+			break
+		}
+		off = mark.Offset
+	}
+	if off > 0 {
+		newoff, err := f.Seek(off, 0)
+		if err != nil {
+			return nil, err
+		}
+		if newoff != int64(off) {
+			return nil, errors.New("file truncated") //XXX
+		}
+	}
+	return &rangeReader{
+		Reader: *NewReader(f, zctx),
+		closer: f,
+		start:  span.Ts,
+		end:    span.End(),
+	}, nil
+}
+
+func (r *rangeReader) Read() (*zng.Record, error) {
+	for {
+		rec, err := r.Reader.Read()
+		if err != nil {
+			return nil, err
+		}
+		r.nread++
+		if rec.Ts < r.start {
+			continue
+		}
+		if rec != nil && rec.Ts > r.end {
+			rec = nil
+		}
+		return rec, nil
+	}
+}
+
+func (r *rangeReader) Close() error {
+	return r.closer.Close()
+}
+
+func (r *rangeReader) Reads() uint64 {
+	return r.nread
+}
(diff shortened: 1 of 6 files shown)


zio/bzngio/index_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+func TestBzngIndex(t *testing.T) {
+	// get a scratch directory
+	dir, err := ioutil.TempDir("", "")
+	require.NoError(t, err)
+	defer os.RemoveAll(dir)
+
+	// create a test bzng file
+	reader := zngio.NewReader(strings.NewReader(bzng), resolver.NewContext())
+	fname := fmt.Sprintf("%s%c%s", dir, os.PathSeparator, "test.zng")
```
nwt, line 77:
```suggestion
	fname := filepath.Join(dir, "test.zng")
```

zio/bzngio/index_test.go
```diff
@@ -0,0 +1,126 @@
...
+	fp, err := os.OpenFile(fname, os.O_WRONLY|os.O_CREATE, 0666)
```
nwt, line 78:
```suggestion
	f, err := os.Create(fname)
```

zio/bzngio/index_test.go
```diff
@@ -0,0 +1,126 @@
...
+// file with the contents above and a time span delimited by startTime
+// and endTime as defined above.  First verifies that calling Read()
+// repeatedly gives just the records that fall within the requested time
+// span.  Then, if checkReads is true, verify that the total records read
+// from disk is just enough to cover the time span, and not the entire
+// file (with streams of 2 records each and parts of 3 streams being
+// inside the time span, a total of 6 records should be read).
+func checkReader(t *testing.T, r bzngio.IndexReader, checkReads bool) {
+	expect := 3
+	for expect <= 6 {
```
nwt, line 49:
```suggestion
for expect := 3; expect <= 6; expect++ {
```

zio/bzngio/index_test.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/brimsec/zq/pkg/nano"
+	"github.com/brimsec/zq/zio"
+	"github.com/brimsec/zq/zio/bzngio"
+	"github.com/brimsec/zq/zio/zngio"
+	"github.com/brimsec/zq/zng/resolver"
+	"github.com/stretchr/testify/require"
+)
+
+const bzng = `
+#0:record[ts:time,value:int32]
```
nwt, line 19: This isn't bzng.

zio/bzngio/index.go
```diff
@@ -0,0 +1,183 @@
...
+	for {
+		rec, err := r.Reader.Read()
+		if err != nil {
+			return nil, err
+		}
+		r.nread++
+		if rec.Ts < r.start {
+			continue
+		}
+		if rec != nil && rec.Ts > r.end {
```
alfred-landrum, line 170: This check for rec != nil happens after the `rec.Ts < r.start` check?

zio/bzngio/index.go
```diff
@@ -0,0 +1,183 @@
...
+	var off int64
+	// XXX binary search
+	for _, mark := range index {
+		if mark.Ts > span.Ts {
+			break
+		}
+		off = mark.Offset
+	}
+	if off > 0 {
+		newoff, err := f.Seek(off, 0)
```
alfred-landrum, line 144:
```suggestion
		newoff, err := f.Seek(off, io.SeekStart)
```

zio/bzngio/index.go
aswan: whoops good catch

zio/bzngio/index.go
```diff
@@ -0,0 +1,183 @@
...
+		rec, err := i.readOne()
+		if rec == nil {
+			i.parent.index = i.marks
+			i.parent.indexReady = true
+			return nil, nil
+		}
+
+		if err != nil {
+			return nil, err
+		}
```
nwt, line 77: This should probably happen before the `if rec == nil` statement.

aswan: Should it?  The contract for Read() isn't spelled out anywhere but if it returns (nil, io.EOF) then this check needs to come first (or move inside the err check)

nwt: It should. See #587.

zio/bzngio/index.go
```diff
@@ -0,0 +1,183 @@
...
+		if rec != nil && rec.Ts > r.end {
+			rec = nil
+		}
+		return rec, nil
+	}
+}
+
+func (r *rangeReader) Close() error {
+	return r.closer.Close()
+}
```
nwt, line 179: Embed the io.Closer (instead of making it a field) so you don't need this forwarding method.

zio/bzngio/index.go
```diff
@@ -0,0 +1,175 @@
...
+type Index struct {
+	index      []mark
+	indexReady bool
+}
+
+type IndexReader interface {
+	zbuf.ReadCloser
+	// count of how many records were read from disk, currently just
+	// used for testing.
+	Reads() uint64
```
henridf, line 23: if this is only used for testing, how about removing it from the public interface and accessing it via a type assertion in the test?

zio/bzngio/index.go
```diff
@@ -0,0 +1,175 @@
...
+// index in the process, subsequent readers can use this index to read
+// only the relevant zng streams from the underlying file.
+func NewIndex() Index {
+	return Index{}
+}
+
+func (i *Index) NewReader(f *os.File, zctx *resolver.Context, span nano.Span) (IndexReader, error) {
+	if i.indexReady {
+		return newRangeReader(f, zctx, i.index, span)
+	} else {
```
henridf, line 43: else not needed

zio/bzngio/index.go
```diff
@@ -0,0 +1,175 @@
...
+	"io"
+	"os"
+
+	"github.com/brimsec/zq/pkg/nano"
+	"github.com/brimsec/zq/zbuf"
+	"github.com/brimsec/zq/zng"
+	"github.com/brimsec/zq/zng/resolver"
+)
+
+type Index struct {
```
henridf, line 14: Maybe a comment indicating that this is a time index? Or call the struct TimeIndex? I guess the high-level bit is given the other kinds of indexing we have (or will have) elsewhere, it seems like the time-specific nature of this should be highlighted somehow.

zio/bzngio/index.go
```diff
@@ -0,0 +1,175 @@
...
+	Reader
+	io.Closer
+	start nano.Ts
+	end   nano.Ts
+	nread uint64
+}
+
+func newRangeReader(f *os.File, zctx *resolver.Context, index []mark, span nano.Span) (*rangeReader, error) {
+	var off int64
+	// XXX binary search
```
henridf, line 132: This comment is a bit cryptic.. ❔

zio/bzngio/index.go
```diff
@@ -0,0 +1,175 @@
...
+// NewIndex creates a new Index object, which is the container that holds
+// the in-memory index for a (b)zng file.  The first call to NewReader()
+// will return a reader that scans the entire file, building a time-based
+// index in the process, subsequent readers can use this index to read
+// only the relevant zng streams from the underlying file.
+func NewIndex() Index {
+	return Index{}
+}
+
+func (i *Index) NewReader(f *os.File, zctx *resolver.Context, span nano.Span) (IndexReader, error) {
```
henridf, line 40: This probably deserves a comment explaining what `span` means...

zio/bzngio/index.go
```diff
@@ -0,0 +1,175 @@
...
+		}
+		if rec.Ts <= i.end {
+			return rec, nil
+		}
+
+		// This record falls after the end of the requested time
+		// span.  Spin through this loop until we hit EOF anyway
+		// to finish building the index.
+		// XXX this will be wasteful if small ranges near the
+		// start of the file is all that is ever read.  revisit this...
```
henridf, line 90: (fwiw, the current seems just right to me for a volatile index. if/when we persist it, then the index could be created at ingest time, where we read everything anway...)

zio/bzngio/index.go
henridf: something like 

```go
if readser, ok := v.(interface{ reads() uint64 }); ok {
  reads := readser.reads()
 // ...
}
```

zio/bzngio/index.go
```diff
@@ -11,16 +11,13 @@ import (
 	"github.com/brimsec/zq/zng/resolver"
 )
 
-type Index struct {
+type TimeIndex struct {
 	index      []mark
 	indexReady bool
 }
 
 type IndexReader interface {
```
henridf, line 19: Is this interface needed at all any more?

zio/bzngio/index.go
alfred-landrum: This is an example where using separate test packages (`package bzngio_test`) isn't helpful. If the test where in `package bzngio`, NewReader could return a zbuf.ReadCloser, and the unit test can type assert directly to `indexReader`.

@aswan : if you just wanted to use `package bzngio` in index_test.go, that would seem ok to me.

aswan: I don't have any strong opinions here, do you want me to make this change or merge as-is?

nwt: Make the change!

alfred-landrum: It's ok as-is; apologies for the drive-by comment here. I'll talk to folks about the "_test" package pattern separately.

zio/bzngio/index.go
```diff
@@ -0,0 +1,168 @@
...
+		parent: ti,
+	}, nil
+}
+
+// indexReader is a zbuf.Reader that also builds an index as it reads.
+type indexReader struct {
+	Reader
+	io.Closer
+	start   nano.Ts
+	end     nano.Ts
```
nwt, line 55: Nit: Replace these with `span nano.Span`.

zqd/space/space.go
```diff
@@ -48,7 +50,7 @@ func Open(root, name string) (*Space, error) {
 		}
 		return nil, err
 	}
-	return &Space{path, c}, nil
+	return &Space{path, c, bzngio.NewTimeIndex()}, nil
```
nwt, line 53: It looks like we call space.Open for every search right now. Doesn't that mean the index will be recreated on every search?

zio/bzngio/index.go
aswan: This was deliberate since we reference end over and over and the only way to get the end time from a span is to compute it.  Maybe I'm optimizing prematurely or maybe the go compiler is clever enough to figure out that there's no aliasing and so that expression is constant and not recompute it every time, but I opted for the obviously-not-unnecessarily-inefficient route.

zqd/space/space.go
aswan: Ugh.  How about if I leave out the integration with zqd for now since fixing this entails bigger changes to space.go

zio/bzngio/index.go
nwt: Maybe do `end := i.span.End()` before entering the for loop inside Read, then?

zqd/space/space.go
nwt: Sounds good to me.

zio/bzngio/index.go
aswan: Sure, but Read() gets called repeatedly without the span ever changing.

PR #580, opened 2020-04-15
add zng.Builder to easily build record values

mccanne:
This PR introducers a Builder for package zng that constructs zng.Record.
This is a bare bones first cut and could be a very useful building block 
if fleshed out to handle lots of different record-building use cases.
The point here is you can create a builder and keep re-using it in the inner loop
without anything escaping to GC a la the builder design patter in Go.

For now,  the zdx port to zng will use this bare bones version.

### DIFF
--- a/zng/builder.go
+++ b/zng/builder.go
@@ -1,0 +1,42 @@
+package zng
+
+import (
+	"github.com/brimsec/zq/zcode"
+)
+
+// Builder provides a way of easily and efficiently building records
+// of the same type.
+type Builder struct {
+	zcode.Builder
+	Type *TypeRecord
+}
+
+func NewBuilder(typ *TypeRecord) *Builder {
+	return &Builder{Type: typ}
+}
+
+// Build encodes the top-level zcode.Bytes values as the Raw field
+// of a record and sets that field and the Type field of the passed-in record.
+// XXX This currently only works for zvals that are properly formatted for
+// the top-level scan of the record, e.g., if a field is record[id:[record:[orig_h:ip]]
+// then the zval passed in here for that field must have the proper encooding...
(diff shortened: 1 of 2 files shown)


zng/builder_test.go
```diff
@@ -0,0 +1,68 @@
...
+#0:record[key:ip]
+#1:record[a:int64,b:int64,c:int64]
+#2:record[a:int64,r:record[x:int64]]
+0:[1.2.3.4;]
+1:[1;2;3;]
+2:[7;[3;]]`
+
+func TestBuilder(t *testing.T) {
+	in := []byte(strings.TrimSpace(builder) + "\n")
+	r := zngio.NewReader(bytes.NewReader(in), resolver.NewContext())
```
nwt, line 26: zngio.Reader is nicely flexible about leading and trailing newlines, so you can simplify if you like.
```suggestion
	r := zngio.NewReader(strings.NewReader(builder), resolver.NewContext())
```

zng/builder.go
```diff
@@ -0,0 +1,42 @@
...
+
+// Build encodes the top-level zcode.Bytes values as the Raw field
+// of a record and sets that field and the Type field of the passed-in record.
+// XXX This currently only works for zvals that are properly formatted for
+// the top-level scan of the record, e.g., if a field is record[id:[record:[orig_h:ip]]
+// then the zval passed in here for that field must have the proper encooding...
+// this works fine when values are extracted and inserted from the proper level
+// but when a leaf values are inserted we should have another method to handle this,
+// e.g., by encoding the dfs traversal of the record type with info about
+// primitive vs container insertions.  This could be the start of a whole pacakge
```
nwt, line 26:
```suggestion
// primitive vs container insertions.  This could be the start of a whole package
```

PR #582, opened 2020-04-15
Error-tolerant log ingest

### LINKED ISSUE #528: Log ingest error/warning handling

henridf:
Switch log ingest to use the "error-tolerant" mode by default, where
upon encountering an error, the error'd file is closed, a warning sent
to client, and ingest continues reading from any other readers.

The "stop everything upon error" behavior can be selected by the
stop_err flag in the log post request.

closes #528 

~~(based on #577 ; will rebase to master when that one merges)~~ now this PR has been approved, I've re-based to master as i wanted to already resolve conflicts from the merge of #538.

### DIFF
--- a/zqd/api/api.go
+++ b/zqd/api/api.go
@@ -112,7 +112,13 @@ type PacketPostStatus struct {
 
 type LogPostRequest struct {
 	Paths          []string             `json:"paths"`
+	StopErr        bool                 `json:"stop_err"`
 	JSONTypeConfig *ndjsonio.TypeConfig `json:"json_type_config"`
+}
+
+type LogPostWarning struct {
+	Type string `json:"type"`
+	Msg  string `json:"msg"`
 }
 
 type LogPostStatus struct {
--- a/zqd/api/api.go
+++ b/zqd/api/api.go
@@ -116,9 +116,9 @@ type LogPostRequest struct {
 	JSONTypeConfig *ndjsonio.TypeConfig `json:"json_type_config"`
 }
 
-type LogPostWarning struct {
-	Type string `json:"type"`
-	Msg  string `json:"msg"`
+type LogPostWarnings struct {
+	Type     string   `json:"type"`
+	Warnings []string `json:"warnings"`
 }
 
 type LogPostStatus struct {
--- a/zqd/api/unpack.go
+++ b/zqd/api/unpack.go
@@ -34,8 +34,8 @@ func unpack(b []byte) (interface{}, error) {
 		out = &PacketPostStatus{}
 	case "LogPostStatus":
 		out = &LogPostStatus{}
-	case "LogPostWarning":
-		out = &LogPostWarning{}
+	case "LogPostWarnings":
+		out = &LogPostWarnings{}
 	case "":
 		return nil, fmt.Errorf("no type field in search result: %s", string(b))
 	default:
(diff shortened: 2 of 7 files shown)


zqd/api/api.go
```diff
@@ -112,9 +112,15 @@ type PacketPostStatus struct {
 
 type LogPostRequest struct {
 	Paths          []string             `json:"paths"`
+	StopErr        bool                 `json:"stop_err"`
 	JSONTypeConfig *ndjsonio.TypeConfig `json:"json_type_config"`
 }
 
+type LogPostWarning struct {
+	Type string `json:"type"`
+	Msg  string `json:"msg"`
```
alfred-landrum, line 121: This would make it consistent with `SearchWarnings`:
```suggestion
	Warnings  []string `json:"warnings"`
```
though I'm not sure what the rationale is to have a slice of string instead of a single string.

henridf: I noticed that, but also not sure of the rationale for slice in `SearchWarnings`. Currently I don't think there's a way for search to send multiple warnings in one message. Still I'll adjust as per your suggestion as it gives us the option to send multiple in the future.

zqd/api/api.go
```diff
@@ -112,9 +112,15 @@ type PacketPostStatus struct {
 
 type LogPostRequest struct {
 	Paths          []string             `json:"paths"`
+	StopErr        bool                 `json:"stop_err"`
```
nwt, line 115: I feel like this deserves a more intention-revealing name, though I'm not sure what. FailOnLogError? FailOnInputError? SomethingElse?

zqd/api/api.go
```diff
@@ -112,9 +112,15 @@ type PacketPostStatus struct {
...
 	JSONTypeConfig *ndjsonio.TypeConfig `json:"json_type_config"`
 }
 
+type LogPostWarnings struct {
+	Type     string   `json:"type"`
+	Warnings []string `json:"warnings"`
```
nwt, line 121: Any reason not to simplify this to carry just a single warning?
```suggestion
type LogPostWarning struct {
	Type    string   `json:"type"`
	Warning string `json:"warning"`
```

alfred-landrum: @nwt - see my comment above; this is to make it consistent with the existing search warnings response. Separate from the PR, we can modify the Brim client to just expect a single string.

zqd/api/api.go
henridf: Brevity vs explicitness tradeoff... I used StopErr here to be consistent with the `zq` command line flag. I think we should change both jointly if we want to use something else.

nwt: Let's change 'em later, then.

zqd/api/api.go
henridf: @nwt i went with @alfred-landrum's to be consistent with search warnings. https://github.com/brimsec/zq/pull/582#discussion_r408943752

we are both still uncertain about the need for a slice so maybe we should change this and search warning

nwt: Let's move to a single string later, then.

Issue #555, opened 2020-04-11
zqd: send scanner stats when running a flowgraph


henridf:
`zqd` currently sends zero-valued `api.ScannerStats` when running a flowgraph. We should switch it to sending actual scanner stats at some point. TBD if these should also be sent during ingest or only during search (ingest also does run scanner(s) on log files). this would also be a good time to revisit the question raised in this thread: https://github.com/brimsec/zq/pull/549#discussion_r406919672.

PR #592, opened 2020-04-16
JSON: error on incomplete descriptors

henridf:
Previously, incomplete descriptors did not lead to any user-visible
notification. (A stats counter is increased, but flowgraph stats are
not currently hooked up to anything, see #555).

This PR turns incomplete descriptors into errors (like other json type
errors), which will lead to their notification in the Brim app and on
the CLI.

### DIFF
--- a/zqd/handlers_test.go
+++ b/zqd/handlers_test.go
@@ -409,7 +409,7 @@ func TestPostNDJSONLogWarning(t *testing.T) {
 	const src1 = `{"ts":"1000","_path":"nosuchpath"}
 {"ts":"2000","_path":"http"}`
 	const src2 = `{"ts":"1000","_path":"http"}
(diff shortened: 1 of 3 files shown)


zqd/handlers_test.go
```diff
@@ -409,7 +409,7 @@ func TestPostNDJSONLogWarning(t *testing.T) {
 	const src1 = `{"ts":"1000","_path":"nosuchpath"}
 {"ts":"2000","_path":"http"}`
 	const src2 = `{"ts":"1000","_path":"http"}
-{"ts":"2000","_path":"nosuchpath"}`
+{"ts":"1000","_path":"http","extra": "foo"}`
```
nwt, line 412:
```suggestion
{"ts":"1000","_path":"http","extra":"foo"}`
```

PR #600, opened 2020-04-17
adapt zdx to use zng instead of ad hoc sorted tables

mccanne:
This commit adapts zar and zdx to use zng files as the underlying
representation of sorted index keys.  Now that we have stream
boundaries within zng files, we can use these boundaries to create
hierarchical indexes.

This commit also generalizes the zar command funcationality to handle
arbitrary indexing patterns of types or fields.  Once the zng indexing
machinery was in place, this change was relatively straightforward.


archive/field.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/brimsec/zq/zng/resolver"
+)
+
+// FieldIndexer builds (as a zbuf.Writer) and generates (as a zbuf.Reader)
+// a sorted set of zng.Records where each record is has one column called
+// "key" whose values are the unique values found in the zbuf.Writer stream
+// of records that contain the indicated field.  XXX It is currently an error
+// to try to index a field name that appears as different types.  This could
+// be pretty easily fixed by creating a new FieldIndexer for each type that is
+// found and and encodging each type name (with the field name) in the zdx
```
henridf, line 16:
```suggestion
// found and and encoding each type name (with the field name) in the zdx
```

cmd/zar/index/command.go
```diff
@@ -11,15 +11,23 @@ import (
...
+	Short: "create zdx index bundles for bzng files",
 	Long: `
-zar find descends the directory argument looking for bzng files and creates an index
-file for IP addresses for each bzng file encountered.  An index is written to
-a sub-directory of the directory containing each encountered bzng file, where the
-name of the sub-directory is a concatenation of the bzng file name and the suffix
-".zar".
-The current version supports only IP address, but this will soon change.
+zar index descends the directory argument starting at dir and looks
+for bzng files.  Each bzng file fund is indexed according to the one or
```
henridf, line 18:
```suggestion
for bzng files.  Each bzng file found is indexed according to the one or
```

archive/field.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/brimsec/zq/expr"
+	"github.com/brimsec/zq/zdx"
+	"github.com/brimsec/zq/zng"
+	"github.com/brimsec/zq/zng/resolver"
+)
+
+// FieldIndexer builds (as a zbuf.Writer) and generates (as a zbuf.Reader)
+// a sorted set of zng.Records where each record is has one column called
+// "key" whose values are the unique values found in the zbuf.Writer stream
+// of records that contain the indicated field.  XXX It is currently an error
```
henridf, line 13:
```suggestion
// FieldIndexer implements the Indexer interface, building an index where each record is has one column called
// "key" whose values are the unique values found in the zbuf.Writer stream
// of records that contain the indicated field.  XXX It is currently an error
```

archive/type.go
```diff
@@ -6,17 +6,37 @@ import (
...
 	"github.com/brimsec/zq/zng"
+	"github.com/brimsec/zq/zng/resolver"
 )
 
 var ErrSyntax = errors.New("syntax/format error encountered parsing zng data")
 
+// TypeIndexer builds (as a zbuf.Writer) and generates (as a zbuf.Reader)
+// a sorted set of zng.Records where each record has one column called
+// "key" whose values are the unique values found in the zbuf.Writer stream
+// from any fields of the incoming records that have the indicated type.
```
henridf, line 17:
```suggestion
// TypeIndexer implements the Indexer interface, building an index where each record has one column called
// "key" whose values are the unique values found in the zbuf.Writer stream
// from any fields of the incoming records that have the indicated type.
```

archive/indexer.go
```diff
@@ -54,60 +83,57 @@ func CreateIndexes(dir string) error {
...
-	// make subdirectory for index if it doesn't exist
-	if err := os.Mkdir(subdir, 0755); err != nil {
-		if !os.IsExist(err) {
-			return err
-		}
+	var indexers []Indexer
+	for _, rule := range rules {
+		indexer := rule.NewIndexer(subdir)
+		indexers = append(indexers, indexer)
+		fmt.Printf("%s: creating index %s\n", path, indexer.Path())
```
henridf, line 95: debugging leftover?

mccanne: Good pint.  This was intended as output of command but should be elsewhere so this can be restified.

We need to chanelize this as well as the hits ini zar find so I think I will put this in an issue and leave for a later PR.

archive/type.go
```diff
@@ -6,17 +6,38 @@ import (
...
+	"github.com/brimsec/zq/zng/resolver"
 )
 
 var ErrSyntax = errors.New("syntax/format error encountered parsing zng data")
 
+// TypeIndexer implements the Indexer interface, building an index
+// where each record has one column called "key" whose values
+// are the unique values found in the zbuf.Writer stream
+// from any fields of the incoming records that have the indicated type.
+
```
henridf, line 18:
```suggestion
// from any fields of the incoming records that have the indicated type.
```

looks like my edit inserted a newline

archive/indexer.go
henridf: 👍

henridf:
This looks great and now I want to be able to do something like `zar index -d . :record[orig_h:ip,orig_p:port,resp_h:ip,resp_p:port]` .... to be able to index on the conn tuple... we're pretty closed right?

cmd/zar/find/command.go
```diff
@@ -13,13 +13,22 @@ import (
...
+or ":type=value" (for type searches).  For example, if type "ip" has been
+indexed then the IP 10.0.1.2 can be searched by saying
+
+	zar find -d /path/to/logs :ip=10.0.1.2
+
+Or if the field "uri" has been indexed, you might say
+
+	zar find -d /path/to/logs uri=/x/y/z
+
+The path of each bnzng file that matches the pattern is printed.
```
henridf, line 31:
```suggestion
The path of each bzng file that matches the pattern is printed.
```

cmd/zar/find/command.go
```diff
@@ -13,13 +13,22 @@ import (
...
-a corresponding zar index and if that index contains the <ip> argument,
-then the path of the zng file is printed.
-The current version supports only IP address, but this will soon change.
+"zar find" descends the directory given by the -d option looking for bzng files
+that have a corresponding zar index conforming to the indicated <pattern>.
+The <pattern> argument has the form "field=value" (for field searches)
+or ":type=value" (for type searches).  For example, if type "ip" has been
+indexed then the IP 10.0.1.2 can be searched by saying
+
+	zar find -d /path/to/logs :ip=10.0.1.2
```
henridf, line 25: Probably not for this PR, but I think for user-friendliness we should try to have the zar find expression syntax here be contained within the zql expression syntax... which would mean either adding support for `:type=value` to zql, or changing zar find to require `*=10.0.1.2`.

zdx/mem.go
```diff
@@ -1,59 +1,109 @@
...
+// MemTable implements an in-memory table to build a zdx.
+// It implements zbuf.Reader and will generate a stream of zng.Records that
+// are either single column ("key") or a two-column ("key", "value") where the
+// types of the columns depend upon the zng.Values entered into the table.
 type MemTable struct {
-	table  map[string][]byte
-	keys   []string
-	offset int
+	table   map[string]zcode.Bytes
+	keys    []string
```
aswan, line 18: why string and not []byte ?

zdx/finder.go
```diff
@@ -45,71 +79,111 @@ func (f *Finder) Close() error {
...
+// of each record in the stream.  The second column in the record must be an
+// integer and that value is returned for the largest said key.
+func lookupOffset(reader zbuf.Reader, key zcode.Bytes) (int64, error) {
 	var lastOff int64 = -1
-	for len(buf) > 10 {
-		k, off, n := DecodeIndex(buf)
-		if k == nil {
-			return -1, ErrCorruptFile
+	for {
+		rec, err := reader.Read()
```
henridf, line 89: this `err` is not handled

zdx/mem.go
```diff
@@ -1,59 +1,109 @@
...
+	offset  int
+	zctx    *resolver.Context
+	builder *zng.Builder
+	recBuf  zng.Record
```
henridf, line 22: looks like `recBuf` isn't used ?

archive/common.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+// Common implements the shared function across the different indexers.
+// Right now, this is pretty lean but once the indexers handle LSM-like
+// merging of sorted zng files, this will probably go here.
+type Common struct {
+	path string
+	// zdx.MemTable is embedded to provide the zbuf.Reader implementation
+	// for the Indexer interface.
+	*zdx.MemTable
```
nwt, line 14: Nit: Put embedded fields before explicitly-named fields.

archive/finder.go
```diff
@@ -39,23 +44,28 @@ func Find(dir string, pattern []byte) ([]string, error) {
...
 		} else {
-			err = fmt.Errorf("%s: %s", zdxPath, err)
+			err = fmt.Errorf("%s: %s", finder.Path(), err)
 		}
 		return false, err
 	}
-	v, err := finder.Lookup(pattern)
+	keyBytes, err := keyType.Parse([]byte(pattern))
+	if err != nil {
+		return false, fmt.Errorf("%s: %s", finder.Path(), err.Error())
```
nwt, line 64:
```suggestion
		return false, fmt.Errorf("%s: %w", finder.Path(), err)
```

archive/finder.go
```diff
@@ -39,23 +44,28 @@ func Find(dir string, pattern []byte) ([]string, error) {
...
+		return false, err
+	}
+	finder := rule.NewFinder(subdir)
+	keyType, err := finder.Open()
 	if err != nil {
 		if err == os.ErrNotExist {
 			err = nil
 		} else {
-			err = fmt.Errorf("%s: %s", zdxPath, err)
+			err = fmt.Errorf("%s: %s", finder.Path(), err)
```
nwt, line 58:
```suggestion
			err = fmt.Errorf("%s: %w", finder.Path(), err)
```

archive/finder.go
```diff
@@ -39,23 +44,28 @@ func Find(dir string, pattern []byte) ([]string, error) {
...
 	}
-	v, err := finder.Lookup(pattern)
+	keyBytes, err := keyType.Parse([]byte(pattern))
+	if err != nil {
+		return false, fmt.Errorf("%s: %s", finder.Path(), err.Error())
+	}
+	rec, err := finder.Lookup(keyBytes)
 	if err != nil {
-		err = fmt.Errorf("%s: %s", zdxPath, err.Error())
+		err = fmt.Errorf("%s: %s", finder.Path(), err.Error())
```
nwt, line 68:
```suggestion
		err = fmt.Errorf("%s: %w", finder.Path(), err)
```

archive/indexer.go
```diff
@@ -8,21 +8,50 @@ import (
...
+		if !os.IsExist(err) {
+			return "", err
+		}
+	}
+	return subdir, nil
+}
+
+// Indexer provides a means to index a zng file.  First, a stream of zng.Records
+// is written to the Indexer via zbuf.Writer, then the indexed records are read
+// as a stream via zbuf.Reader.   The index is managed as a zdx bundle.
```
nwt, line 44:
```suggestion
// as a stream via zbuf.Reader.  The index is managed as a zdx bundle.
```

archive/indexer.go
```diff
@@ -8,21 +8,50 @@ import (
...
+// XXX currently we are supporting just in-memory indexing but it would be
+// straightforward to extend this to spill in-memory tables then merge them
+// on close a la LSM.
 type Indexer interface {
-	Create(*zdx.Writer, zbuf.Reader)
-	//Search([]byte) bool
+	Path() string
+	zbuf.Writer
+	zbuf.Reader
```
nwt, line 51: Nit: Put embedded interfaces first.

archive/rule.go
```diff
@@ -0,0 +1,85 @@
...
+	"fmt"
+	"path/filepath"
+
+	"github.com/brimsec/zq/expr"
+	"github.com/brimsec/zq/zdx"
+	"github.com/brimsec/zq/zng"
+	"github.com/brimsec/zq/zng/resolver"
+)
+
+// Rule is a interface for creating pattern-specific indexers and finders
```
nwt, line 13:
```suggestion
// Rule is an interface for creating pattern-specific indexers and finders
```

archive/rule.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	return &TypeRule{
+		Type: typ,
+	}, nil
+}
+
+func (t *TypeRule) NewIndexer(dir string) Indexer {
+	zdxPath := filepath.Join(dir, typeZdxName(t.Type))
+	// XXX DANGER, remove without warning, should we have a force flag?
+	zdx.Remove(zdxPath)
```
nwt, line 47: Handle errors?

archive/indexer.go
```diff
@@ -54,60 +83,57 @@ func CreateIndexes(dir string) error {
...
-			return nil, err
+			return err
 		}
 		if rec == nil {
-			return table, nil
+			break
+		}
+		for _, indexer := range indexers {
+			err := indexer.Write(rec)
+			if err != nil {
```
nwt, line 116:
```suggestion
			if err := indexer.Write(rec); err != nil {
```

archive/rule.go
```diff
@@ -0,0 +1,85 @@
...
+	return &FieldRule{
+		field:    field,
+		accessor: accessor,
+	}, nil
+}
+
+func (f *FieldRule) NewIndexer(dir string) Indexer {
+	zdxPath := filepath.Join(dir, fieldZdxName(f.field))
+	// XXX DANGER, remove without warning, should we have a force flag?
+	zdx.Remove(zdxPath)
```
nwt, line 78: Handle errors?

cmd/zdx/create/command.go
```diff
@@ -30,34 +47,109 @@ type CreateCommand struct {
...
+	f.StringVar(&c.outputFile, "o", "zdx", "output zdx bundle name")
+	f.StringVar(&c.keyField, "k", "", "field name of keys ")
+	f.StringVar(&c.valField, "v", "", "field name of values ")
+	f.BoolVar(&c.skip, "S", false, "skip all records except for the first of each stream")
 	return c, nil
 }
 
 func (c *CreateCommand) Run(args []string) error {
+	if c.keyField == "" {
+		return errors.New("must specify a key field with -key")
```
nwt, line 69:
```suggestion
		return errors.New("must specify a key field with -k")
```

cmd/zdx/lookup/command.go
```diff
@@ -1,23 +1,26 @@
...
 var Lookup = &charm.Spec{
 	Name:  "lookup",
-	Usage: "lookup -k <key> <file>",
-	Short: "lookup a key in an zdx file and print value as hex bytes",
+	Usage: "lookup -k <key> <bundle>",
+	Short: "lookup a key in an zdx file and print value as zng record",
 	Long: `
-The lookup command uses the index files of an zdx hierarchy to locate the
-specified key in the base zdx file and displays the value as bytes.`,
+The lookup command locates the specified <key> in the base file oof the
```
nwt, line 20:
```suggestion
The lookup command locates the specified <key> in the base file of the
```

cmd/zdx/lookup/command.go
```diff
@@ -1,23 +1,26 @@
...
+zdx <bundle> and displays the result as a zng record.
+The key argument specifies a value to lookup in the table and must be parseable
```
nwt, line 22:
```suggestion
The key argument specifies a value to look up in the table and must be parseable
```

cmd/zdx/create/command.go
```diff
@@ -30,34 +47,109 @@ type CreateCommand struct {
...
+		in = f
 	}
-	writer, err := zdx.NewWriter(c.outputFile, c.framesize, 0)
+	zctx := resolver.NewContext()
+	reader := bzngio.NewReader(in, zctx)
+	writer, err := zdx.NewWriter(c.outputFile, c.framesize)
 	if err != nil {
 		return err
 	}
 	defer writer.Close()
```
nwt, line 97: Handle writer.Close errors.

cmd/zdx/lookup/command.go
```diff
@@ -44,23 +51,32 @@ func (c *LookupCommand) Run(args []string) error {
...
+	if rec == nil {
+		return nil
+	}
+	writer, err := emitter.NewFile(c.outputFile, c.ofmt, &zio.Flags{})
+	if err != nil {
+		return err
 	}
-	return nil
+	defer writer.Close()
+	return writer.Write(rec)
```
nwt, line 81: Handle writer.Close errors.

cmd/zdx/merge/command.go
```diff
@@ -53,15 +57,20 @@ func (c *MergeCommand) Run(args []string) error {
...
 	combiner := zdx.NewCombiner(files, combine)
 	defer combiner.Close()
-	writer, err := zdx.NewWriter(c.oflag, c.framesize, 0)
+	writer, err := zdx.NewWriter(c.oflag, c.framesize)
 	if err != nil {
 		return err
 	}
-	return zdx.Copy(writer, combiner)
+	defer writer.Close()
+	return zbuf.Copy(writer, combiner)
```
nwt, line 75: Handle writer.Close errors.

zdx/zdx_test.go
```diff
@@ -6,64 +6,47 @@ import (
...
-	stream := &entryStream{entries: entries}
-	writer, err := zdx.NewWriter(path, 32*1024, 0)
+	reader := newTextReader(zngText)
+	writer, err := zdx.NewWriter(path, 32*1024)
 	if err != nil {
 		t.Error(err)
 	}
 	defer writer.Close()
-	if err := zdx.Copy(writer, stream); err != nil {
+	if err := zbuf.Copy(writer, reader); err != nil {
```
nwt, line 49: Handle writer.Close errors.

zdx/zdx_test.go
```diff
@@ -132,30 +116,23 @@ func TestZdx(t *testing.T) {
...
-	if err := zdx.Copy(writer, stream); err != nil {
-		t.Error(err)
-	}
+	stream, err := newReader(N)
+	require.NoError(t, err)
+	writer, err := zdx.NewWriter(path, 32*1024)
+	require.NoError(t, err)
+	err = zbuf.Copy(writer, stream)
+	require.NoError(t, err)
 	writer.Close()
```
nwt, line 125: Handle writer.Close errors.

mccanne:
> This looks great and now I want to be able to do something like `zar index -d . :record[orig_h:ip,orig_p:port,resp_h:ip,resp_p:port]` .... to be able to index on the conn tuple... we're pretty close right?

Yes, this is very close.  There's no way right now to parse standalone, typed record values.

cmd/zar/index/command.go
```diff
@@ -11,15 +11,23 @@ import (
...
+for bzng files.  Each bzng file found is indexed according to the one or
+more indexing rules provided.
+
+A pattern is either a field name or a ":" followed by a zng type name.
+For example, to index the all fields of type ports and the field id.orig_h,
+you would run
+
+	zar index -d /path/to/logs id.orig_h :port
+
+Each pattern results a separate zdx bundle for each bzng file found.  The zdx
```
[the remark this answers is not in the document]
mccanne: Agreed on "bundle".  I was discussing with @alfred-landrum .  He had a really good idea for putting all the index bzng files into one big bzng file now that our stream model permits this.  Why don't we fix the terminology when we make that change and then we can just refer to the big index file as an "index file".

archive/indexer.go
```diff
@@ -8,21 +8,50 @@ import (
...
+func typeZdxName(t zng.Type) string {
+	return "zdx:type:" + t.String()
+}
+
+func fieldZdxName(fieldname string) string {
+	return "zdx:field:" + fieldname
+}
+
+func archiveDir(path string) (string, error) {
+	//XXX for now the index directory is the name of the zng file
```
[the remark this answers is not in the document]
mccanne: That would be much better.  I didn't put much thought on this and @alfred-landrum has been working on the "big picture" here for how everything looks in the file system.

cmd/zar/find/command.go
mccanne: I think this is a different discussion from search vs filter.  We're deliberately exposing index search as a foundational aspect of our architecture.  Coming up with the right index search syntax is still TBD.  This command-line syntax is a bit of a placeholder that maps directly onto the present indexing model.  I agree with your philosophy that it should be zql or maybe a subset of zql.

archive/common.go
```diff
@@ -0,0 +1,19 @@
+package archive
+
+import (
+	"github.com/brimsec/zq/zdx"
+)
+
+// Common implements the shared function across the different indexers.
+// Right now, this is pretty lean but once the indexers handle LSM-like
+// merging of sorted zng files, this will probably go here.
+type Common struct {
```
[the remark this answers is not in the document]
mccanne: Sounds good.  I'm guessing this stuff gets refactored and split out at some point, but IndexerCommon is a better name.

archive/finder.go
```diff
@@ -4,11 +4,16 @@ import (
...
 	"os"
 	"path/filepath"
-
-	"github.com/brimsec/zq/zdx"
 )
 
-func Find(dir string, pattern []byte) ([]string, error) {
+// Find descends a directory hierarchy looking for the provided pattern
+// that is used by the rule to determine whether the pattern exists
+// in each index found that conforms with the rule.
```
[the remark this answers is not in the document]
mccanne: Yeah, I remember writing that thinking the same thing as you.  I'll rewrite it.

zdx/finder.go
```diff
@@ -45,71 +79,111 @@ func (f *Finder) Close() error {
...
+		if rec == nil {
+			break
+		}
+		// Since we know that each record is a key in column 0 and
+		// and an int64 offset in column 1, we can pull the zcode.Bytes
+		// encoding out directly with Slice.
```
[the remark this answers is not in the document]
mccanne: lookupOffset() is called only on the non-base index files that have {key,offset} as the record type.  See the "for level > 0" loop in Finder.search()

zdx/finder.go
```diff
@@ -45,71 +79,111 @@ func (f *Finder) Close() error {
...
-func lookupValue(key, buf []byte) ([]byte, error) {
-	for len(buf) > 8 {
-		pair, n := DecodePair(buf)
-		if pair.Key == nil {
-			return nil, ErrCorruptFile
+// lookupExact searches for an exact match of the given key compared to the
+// first columns of the records read from the reader.  If a match is found,
+// the returned boolean is true and, if a second column in the matched record
+// exists, the zng.Value of that column is returned.  If there is no match found,
+// then the boolean return value is false and the zng.Value is the zero value.
```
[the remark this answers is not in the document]
mccanne: yes, out of date.  I will fix.

zdx/mem.go
```diff
@@ -1,59 +1,109 @@
...
 )
 
-// MemTable implements a simple stub for an in-memory SST Reader.  This is
-// useful for testing and reference.  An SST client will typically implement its
-// own in-memory data structure that provides a Reader implementation to convert
-// custom key/value pairs into string/byte-slice pairs.
+// MemTable implements an in-memory table to build a zdx.
+// It implements zbuf.Reader and will generate a stream of zng.Records that
+// are either single column ("key") or a two-column ("key", "value") where the
+// types of the columns depend upon the zng.Values entered into the table.
```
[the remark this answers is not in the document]
mccanne: Correct.  It didn't seem necessary to create separate table types since nil means no value for the key and the two cases are easily handled together.  At some point, we will do more interesting things with the value field (e.g., store the results of a multi-column, aggregatable reducers to do fast stats for a given key across multiple indexes).

zdx/mem.go
mccanne: byte slices cannot be keys in the hash table so when you pull the keys back out they are already strings.  Converting to byte slices would create unnecessary copies.  That said, think things are still escaping to GC in the inner loop but we can fix this later with "unsafe" etc.

zdx/mem.go
mccanne: yeah, that got moved into zng.Builder.  Will delete.

archive/indexer.go
mccanne: I still do 2 spaces after period, but yeah, 3 is too much.

cmd/zdx/lookup/command.go
mccanne: You always get me on lookup

archive/indexer.go
nwt: I'm all about two spaces between sentences.

archive/indexer.go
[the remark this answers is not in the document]
henridf: Cool. One day not too far off I'm guessing this will also want to be all be http-capable so that this can work with local indexes and cloud data, or everything in the cloud.

cmd/zar/find/command.go
henridf: Sounds good

archive/indexer.go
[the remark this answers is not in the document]
mccanne: Yes, @alfred-landrum has been focused on this future proofing.

zdx/mem.go
mccanne: I missed your point!  Sorting doesn't work right and we need to do comparisons of zng.Values.  Will fix.  We should address the GC escapes at some point.  I think it's ok to make the pointers all unsafe while the sort and write happens.

archive/indexer.go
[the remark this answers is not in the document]
henridf: 👍

tests/suite/zdx/babble.tzng
```diff
@@ -0,0 +1,1001 @@
+#0:record[ts:time,s:string,v:int64]
+0:[1587508830.06852324;Potamogalidae-precommissure;51;]
```
henridf, line 2: now i'm curious... what generated these?

expr/temp.go
```diff
@@ -0,0 +1,53 @@
...
+				// accessing a field that is not a record
+				return zng.Value{}
+			}
+			col, ok := recType.ColumnOfField(field)
+			if !ok {
+				return zng.Value{}
+			}
+			body = slice(body, col)
+			if body == nil {
+				return zng.Value{}
```
henridf, line 32: shouldn't this return `zng.Value{recType.Columns[col].Type, nil}` ?

archive/finder.go
```diff
@@ -5,13 +5,21 @@ import (
...
+// Find descends a directory hierarchy looking for index files that can
+// resolve the provided rule.  For each such index found, the rule is used to
+// see if the pattern is in the index.  If the pattern is in the index,
+// then the path name of the file corresponding to that index is included in
+// the slice of strings comprising the return value.
+// XXX We currently allow only one pattern at a time though it might
+// be more efficient at large scale to allow multipe patterns that
+// are effectively OR-ed together so that there is locality of
+// access to the zdx files.
+func Find(dir string, rule Rule, pattern string, hits chan string) error {
```
nwt, line 20:
```suggestion
func Find(dir string, rule Rule, pattern string, hits chan<- string) error {
```

archive/indexer.go
```diff
@@ -54,59 +83,61 @@ func CreateIndexes(dir string) error {
...
-	file, err := os.Open(path)
-	if err != nil {
-		return err
-	}
-	defer file.Close()
-	// XXX should be able to index any kind of logs... find should take
-	// a regexp to match files and then use auto-detect or -i for input format
-	reader := bzngio.NewReader(file, resolver.NewContext())
-	table, err := indexTypeIP(reader)
+func Run(path string, rules []Rule, progress chan string) error {
```
nwt, line 86:
```suggestion
func Run(path string, rules []Rule, progress chan<- string) error {
```

archive/indexer.go
```diff
@@ -15,14 +15,43 @@ import (
...
-	Create(*zdx.Writer, zbuf.Reader)
-	//Search([]byte) bool
+	zbuf.Reader
+	zbuf.Writer
+	Path() string
 }
 
-//XXX this is a test stub that creates simple indexes of IP addresses
-func CreateIndexes(dir string) error {
+func IndexDirTree(dir string, rules []Rule, progress chan string) error {
```
nwt, line 54:
```suggestion
func IndexDirTree(dir string, rules []Rule, progress chan<- string) error {
```

archive/indexer.go
```diff
@@ -15,14 +15,43 @@ import (
...
+func archiveDir(path string) (string, error) {
+	//XXX for now the index directory is the name of the zng file
+	// with the ".zar" extension
+	subdir := path + zarExt
+	// make subdirectory for index if it doesn't exist
+	if err := os.Mkdir(subdir, 0755); err != nil {
+		if !os.IsExist(err) {
+			return "", err
+		}
+	}
```
nwt, line 38:
```suggestion
	if err := os.Mkdir(subdir, 0755); err != nil && !os.IsExist(err) {
		return "", err
	}
```

archive/field.go
```diff
@@ -0,0 +1,40 @@
...
+type FieldIndexer struct {
+	IndexerCommon
+	accessor expr.FieldExprResolver
+}
+
+func NewFieldIndexer(path, field string, accessor expr.FieldExprResolver) *FieldIndexer {
+	return &FieldIndexer{
+		IndexerCommon: IndexerCommon{
+			path:     path,
+			MemTable: zdx.NewMemTable(resolver.NewContext()),
```
nwt, line 27:
```suggestion
			MemTable: zdx.NewMemTable(resolver.NewContext()),
			path:     path,
```
To match declaration order.

archive/type.go
```diff
@@ -6,17 +6,38 @@ import (
...
+	IndexerCommon
+	Type zng.Type
+}
+
+func NewTypeIndexer(path string, refType zng.Type) *TypeIndexer {
+	zctx := resolver.NewContext()
+	return &TypeIndexer{
+		IndexerCommon: IndexerCommon{
+			path:     path,
+			MemTable: zdx.NewMemTable(zctx),
```
nwt, line 29:
```suggestion
			MemTable: zdx.NewMemTable(zctx),
			path:     path,
```
To match declaration order.

cmd/zdx/create/command.go
```diff
@@ -28,36 +40,125 @@ func init() {
...
 func newCreateCommand(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
-	c := &CreateCommand{Command: parent.(*root.Command)}
+	c := &CreateCommand{
+		Command: parent.(*root.Command),
+	}
 	f.IntVar(&c.framesize, "f", 32*1024, "minimum frame size used in zdx file")
-	f.StringVar(&c.outputFile, "o", "zdx", "output file name")
+	f.StringVar(&c.outputFile, "o", "zdx", "output zdx bundle name")
+	f.StringVar(&c.keyField, "k", "", "field name of keys ")
+	f.StringVar(&c.valField, "v", "", "field name of values ")
```
nwt, line 58:
```suggestion
	f.StringVar(&c.keyField, "k", "", "field name of keys")
	f.StringVar(&c.valField, "v", "", "field name of values")
```

tests/suite/zdx/babble.tzng
mccanne: https://github.com/tjarratt/babble

mccanne: https://gist.github.com/mccanne/73b59cea0d71f729e749919ba9ae1d71

archive/indexer.go
nwt: Also, maybe rename to run since this looks like an internal function?

zdx/reader.go
```diff
@@ -1,241 +1,43 @@
...
 
-type reader struct {
-	filename string
-	file     *os.File
-	frame    []byte
-	valsize  int
-}
+const (
+	FrameSize = 32 * 1024
+)
```
nwt, line 12: Do you affirmatively want this to be public?

archive/indexer.go
```diff
@@ -54,59 +83,61 @@ func CreateIndexes(dir string) error {
...
+		}
+		for _, indexer := range indexers {
+			if err := indexer.Write(rec); err != nil {
+				return err
+			}
+		}
+	}
+	// we make the framesize here larger than the writer framesize
+	// since the writer always writes a bit past the threshold
+	const framesize = 32 * 1024 * 2
```
nwt, line 127: Does this comment still make sense? (I'm not sure what "the writer framesize" refers to.)

cmd/zar/index/command.go
```diff
@@ -30,17 +39,40 @@ func init() {
...
+		progress = make(chan string)
+		go func() {
+			for line := range progress {
+				fmt.Println(line)
+			}
+		}()
 	}
-	dir := args[0]
-	return archive.CreateIndexes(dir)
+	return archive.IndexDirTree(c.dir, rules, progress)
```
nwt, line 77: Nit: While unlikely, it's possible for archive.IndexDirTree to return (and for the process to exit) before the progress goroutine print the final message. Adding a sync.WaitGroup would fix that.

(This also applies to the hits goroutine in cmd/zar/find/command.go.)

cmd/zdx/create/command.go
```diff
@@ -28,36 +40,125 @@ func init() {
...
+			// with a reducer to coalesce values that land
+			// on the same key.  right now, a new value
+			// will clobber the old one.
+			if err := table.EnterKeyVal(k, v); err != nil {
+				return err
+			}
+		}
+		if err != nil {
+			return err
+		}
```
nwt, line 157:
```suggestion
```

cmd/zdx/create/command.go
```diff
@@ -28,36 +40,125 @@ func init() {
...
+	}
+	close = false
```
nwt, line 159: I think this might belong after zbuf.Copy.

tests/suite/zdx/babble.tzng
henridf: cool!

expr/temp.go
henridf: (I'm already using this function for something else and ran into a nil type error for nil values)

mccanne: Good point but slice() here should distinguish between nil value and error.  I'll fix.

zdx/reader.go
mccanne: I will leave it for now.  We probably want to make this stuff configurable soon and this will become the default.

archive/indexer.go
mccanne: I'd like to leave this for now.  We should have an issue to fix this up.  The indexer buffer should be  larger than threshold used by the writer because it's a threshold rather than a limit.

cmd/zar/index/command.go
mccanne: goroutines 101

mccanne: Fixed same bug in "zar find" too.

PR #602, opened 2020-04-17
/search: Don't ignore query.Dir

### LINKED ISSUE #599: driver doesn't handle reverse option
A reverse option is passed into here https://github.com/brimsec/zq/blob/master/driver/compile.go#L25 and then ignored.
It should be copied into the Reverse property of the proc.Context.

Looks like this was introduced in the process of refactoring the old duplicated code that turned into driver.  But ideally we should have zqd tests that would have caught this.

henridf:
close #599

### DIFF
--- a/zqd/handlers_test.go
+++ b/zqd/handlers_test.go
@@ -45,7 +45,7 @@ func TestSearch(t *testing.T) {
 	require.Equal(t, test.Trim(src), res)
 }
 
-func TestSearchReverse(t *testing.T) {
+func TestGroupByReverse(t *testing.T) {
 	src := `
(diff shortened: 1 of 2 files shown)


zqd/handlers_test.go
```diff
@@ -45,6 +45,27 @@ func TestSearch(t *testing.T) {
 	require.Equal(t, test.Trim(src), res)
 }
 
+func TestSearchReverse(t *testing.T) {
```
aswan, line 48: nit: name it AggregateReverse or GroupbyReverse or something?
Not necessary though, feel free to merge as-is if this has gone on long enough :)

henridf: done

PR #604, opened 2020-04-17
Introduce =~ operator in filters

aswan:
This new operator means roughly "does the thing on the left hand side
match the pattern on the right hand side".  It is used for glob and
regular expression matching and for matching addresses against subnets.
So, the filter "foo = abc*" now matches the literal string "abc*", glob
matching is expressed as "foo =~ abc*".  A "naked" search for a glob
expression still works as before.
Filters that previously used the = operator for matching are no longer
valid, various tests and docs examples were updated to reflect the new
requirement.

### DIFF
--- a/filter/compare.go
+++ b/filter/compare.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/brimsec/zq/ast"
 	"github.com/brimsec/zq/pkg/byteconv"
+	"github.com/brimsec/zq/reglob"
 	"github.com/brimsec/zq/zng"
 )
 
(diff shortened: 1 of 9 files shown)


filter/compare.go
```diff
@@ -250,6 +255,7 @@ func CompareFloat64(op string, pattern float64) (Predicate, error) {
 var compareString = map[string]func(string, string) bool{
 	"=":  func(a, b string) bool { return a == b },
 	"!=": func(a, b string) bool { return a != b },
+	"=~": func(a, b string) bool { return false },
```
henridf, line 258: hmm actually for string don't we want to return `a == b` ?

aswan: This is actually unreachable due to https://github.com/brimsec/zq/blob/pattern-match-operator/filter/compare.go#L445-L448
I had to put that check outside CompareBstring because we need to do the glob-regexp conversion before unescaping bstrings (ugh).  This could probably stand to be refactored to be clearer, but that snowballs into the whole topic of merging filters with expressions which is on hold indefinitely.

PR #607, opened 2020-04-19
extend ztest to handle arbitrary shell scripts

mccanne:
This commit extends the yaml-based ztest framework to handle
arbitrary scripts.  The details are documented in the ztest package
comment.  The input/output specs have been generalized a bit
and we may want to refactor this all a bit more so that the
newer bits of funcionality can be used by the existing zq-direct
commands and in-process runs.

Only a couple of the old go tests under tests/suite have been
ported over in this commit.  Once this commit is reviewed and
merged, we can port the rest over in a subsequent PR.


ztest/ztest.go
```diff
@@ -271,12 +367,173 @@ func FromYAMLFile(filename string) (*ZTest, error) {
...
+
+func runsh(testname, bindir, dirname string, zt *ZTest) error {
+	dir, err := NewDir(testname, dirname)
+	if err != nil {
+		return err
+	}
+	defer dir.RemoveAll()
+	for _, f := range zt.Inputs {
+		if f.Name == "stdin" {
+			return errors.New("cannot use stdin in a script test")
```
alfred-landrum, line 483: The example in the package documentation above uses stdin; if it's not supported, the doc example should change.

ztest/ztest.go
```diff
@@ -121,66 +169,114 @@ func Run(t *testing.T, dirname string) {
...
+		})
+	}
+}
 
-				if !utf8.ValidString(a) {
-					a = encodeHex(a)
-					b = encodeHex(b)
-				}
+type File struct {
+	// Name is the name of the file in which the test shell code runs.
```
alfred-landrum, line 178:
```suggestion
	// Name is the base file name for this data.
```

ztest/ztest.go
```diff
@@ -121,66 +169,114 @@ func Run(t *testing.T, dirname string) {
...
+	Name string `yaml:"name"`
+	// Data, Hex, and Source represents the different ways file data can
+	// be defined for this file.  Data is a string turned into the contents
+	// of the file, Hex is hex decded, and Source is a string representing
+	// the pathname of a file the repo that is read to comprise the data.
+	Data   string `yaml:"data,omitempty"`
+	Hex    string `yaml:"hex,omitempty"`
+	Source string `yaml:"source,omitempty"`
```
alfred-landrum, line 186: What's an example use case for needing `Source` where `Name` isn't sufficient?

ztest/ztest.go
```diff
@@ -31,6 +31,50 @@
 //      COUNT
 //      2
 //
+// Alternatively, tests can be configured to run as shell scripts.
```
alfred-landrum, line 34: We should add a note at the top of the comments (before the first description) saying that the yaml files are used for two different cases.

ztest/ztest.go
```diff
@@ -121,66 +169,114 @@ func Run(t *testing.T, dirname string) {
...
-				if !utf8.ValidString(a) {
-					a = encodeHex(a)
-					b = encodeHex(b)
-				}
+type File struct {
+	// Name is the name of the file in which the test shell code runs.
+	Name string `yaml:"name"`
+	// Data, Hex, and Source represents the different ways file data can
+	// be defined for this file.  Data is a string turned into the contents
+	// of the file, Hex is hex decded, and Source is a string representing
```
alfred-landrum, line 182:
```suggestion
	// of the file, Hex is hex decoded, and Source is a string representing
```

ztest/ztest.go
```diff
@@ -121,66 +169,114 @@ func Run(t *testing.T, dirname string) {
...
+	if f.Source != "" {
+		cnt++
+	}
+	if cnt > 1 {
+		return fmt.Errorf("%s: must at most one of data, hex, or source", f.Name)
+	}
+	return nil
+}
+
+type Regexp struct {
```
alfred-landrum, line 209: The source text for the regex is available via Regexp.String():
https://golang.org/pkg/regexp/#Regexp.String
so the wrapping struct here isn't necessary.

tests/ztest_test.go
```diff
@@ -1,3 +1,5 @@
+// +build system
+
```
nwt, line 2: I believe the goal of this is to prevent `go test ./...` from running the new script ztests.

It has the side effect of preventing `go test ./...` from running the "in-process" ZQL ztests. I don't think we want that.

One way to achieve the goal without the side effect is to replace the ZTEST_ZQ environment variable with, say, ZTEST_PATH and condition running the script ztests on its presence.

ztest/shell.go
```diff
@@ -0,0 +1,22 @@
...
+func RunShell(dir *Dir, binpath, script string) (string, string, error) {
+	const shfile = "_run.sh"
+	cmd := exec.Command("/bin/bash", dir.Join(shfile))
+	cmd.Env = []string{"PATH=" + binpath}
+	src := "cd " + dir.Path() + "\n" + script
+	if err := dir.Write(shfile, []byte(src)); err != nil {
+		return "", "", err
+	}
+	var stdout bytes.Buffer
+	var stderr bytes.Buffer
```
nwt, line 17:
```suggestion
	var stdout, stderr bytes.Buffer
```

ztest/shell.go
```diff
@@ -0,0 +1,22 @@
...
+)
+
+func RunShell(dir *Dir, binpath, script string) (string, string, error) {
+	const shfile = "_run.sh"
+	cmd := exec.Command("/bin/bash", dir.Join(shfile))
+	cmd.Env = []string{"PATH=" + binpath}
+	src := "cd " + dir.Path() + "\n" + script
+	if err := dir.Write(shfile, []byte(src)); err != nil {
+		return "", "", err
+	}
```
nwt, line 15:
```suggestion
func RunShell(dir, path, script string) (string, string, error) {
	var cmd *exec.Cmd
	if runtime.GOOS == "windows" {
		cmd = exec.Command("cmd.exe", "/c", script)
	} else {
		cmd = exec.Command("bash", "-c", script)
	}	
	cmd.Env = []string{"PATH=" + path}
	cmd.Dir = dir.Path()
```

ztest/ztest.go
mccanne: Oops, just pushed a fix and test for stdin.

ztest/ztest.go
mccanne: I added more clarification because it's a bit overloaded.

ztest/ztest.go
mccanne: Shared test inputs, e.g., ../../test-data/in1.zng

Maybe overkill but it's in and it works.

ztest/ztest.go
mccanne: Sounds good.

ztest/ztest.go
mccanne: sweet

tests/ztest_test.go
mccanne: this is perfect, thanks.

ztest/shell.go
mccanne: nice!

ztest/ztest.go
```diff
@@ -271,15 +374,180 @@ func FromYAMLFile(filename string) (*ZTest, error) {
...
+		}
+		if !bytes.Equal(expected, actual) {
+			return diffErr(string(expected), string(actual))
+		}
+	}
+	return nil
+}
+
+func runsh(testname, bindir, dirname string, zt *ZTest) error {
+	dir, err := NewDir(testname, dirname)
```
nwt, line 483: This creates a temporary directory in dirname, which is inside the source tree. Is that deliberate? (Our other tests do ioutil.TempDir("", pattern) to use the default location for temporary directories.)

mccanne: fix has been pushed (this was from old logic that would leave the testdir around if the test failed for debugging)

Issue #520, opened 2020-04-06
Factor out  multi-file/auto-detect/per-type setup 


henridf:
The zq and ingest packages both have similar setup to turn a list of paths into a combiner. This issue serves as a reminder to revisit and see how to clear this up.

(It was pointed out by @alfred-landrum during review: https://github.com/brimsec/zq/pull/517#pullrequestreview-388503287)

PR #608, opened 2020-04-20
refactor file reading/writing so other tools can leverage

mccanne:
This commit splits zio.Flags into ReaderFlags and WriterFlags and refactors the file opening code so it lives in one place.  This will make it easier to re-use file logic with command options by other tools in the zq package.

The next step is to move much of the json typings stuff into zio and come up with some new names to reflect the fact that this is solving the schema mapping problem.


zio/detector/file.go
```diff
@@ -10,16 +10,22 @@ import (
...
-	}
-	if info.IsDir() {
-		return nil, errors.New("is a directory")
-	}
-	f, err := os.Open(path)
-	if err != nil {
-		return nil, err
+	var f *os.File
+	var err error
+	if path == "-" {
```
henridf, line 15: Shouldn't the "translate '-' to /dev/stdin" step stay in zq?  I don't think we want zq to try reading from stdin if passed a file named "-". 

(I've wanted to do this little bit of factoring for a while and this is what prevented me from doing it.)

henridf: One option would be to have this func take a `[]os.File` or `[]io.Reader` instead.

henridf:
looks like this will close #520

zio/detector/lookup.go
```diff
@@ -22,11 +23,18 @@ func (*nullWriter) Write(*zng.Record) error {
 	return nil
 }
 
-func LookupWriter(format string, w io.WriteCloser, optionalFlags *zio.Flags) *zio.Writer {
-	var flags zio.Flags
+func LookupWriter(w io.WriteCloser, optionalFlags *zio.WriterFlags) *zio.Writer {
+	if optionalFlags == nil {
+		panic("FLAGS SHOULDN'T BE OPTIONAL")
+	}
```
alfred-landrum, line 29:
```suggestion
func LookupWriter(w io.WriteCloser, wrFlags *zio.WriterFlags) *zio.Writer {
	if inFlags == nil {
		panic("required wrFlags not set")
	}
```

and adjust below accordingly

zio/detector/file.go
mccanne: Hmm, I'm not worried about a file named "-" and it's pretty convenient to have this here.  What if we used path="" to mean stdin and the caller could check for "-"?

zio/detector/lookup.go
mccanne: Oops, I wasn't done.  I was looking for calls with nil and will just take out the panic and require the flags arg.

PR #614, opened 2020-04-21
add config param to detector.OpenFile

alfred-landrum:


zio/detector/file.go
```diff
@@ -3,19 +3,26 @@ package detector
 import (
 	"errors"
 	"os"
+	"regexp"
 
 	"github.com/brimsec/zq/zbuf"
-	"github.com/brimsec/zq/zio"
+	"github.com/brimsec/zq/zio/ndjsonio"
```
nwt, line 9: Pulling in ndjsonio here doesn't feel quite right. I'd like to spend a little time on Tuesday thinking about another approach before we merge this.

henridf: @nwt in that case heads-up for when you think about this that another requirement here is for detector to configure it's ndjson trial-reader with the same config as the regular ndjson reader. 

Right now, the ndjson reader used for detection doesn't get json types, which it should.

alfred-landrum: @nwt : note that the `detector` package already has a dependency on ndjsonio; see detector/lookup.go. If the concern is that the detector package shouldn't be doing this kind of configuration itself, then we would need to pass the ndjson config (and any other eventual format config) into their respective `NewReader` functions, similarly to how most NewWriter's take a WriterFlags.

alfred-landrum: pinged @nwt about this offline; he's going with this as-is for now, since it at least moves the json types configuration to one place instead of needing to be spread out.

PR #615, opened 2020-04-21
Fix single remaining case of string api error response

### LINKED ISSUE #496: HTTP errors should return json payloads
For zqd, if an error ours on a request the response should always be a json response.

henridf:
closes #496

### DIFF
--- a/zqd/api/api.go
+++ b/zqd/api/api.go
@@ -13,9 +13,10 @@ import (
 )
 
 type Error struct {
-	Type    string      `json:"type"`
-	Message string      `json:"error"`
-	Info    interface{} `json:"info,omitempty"`
+	Type      string      `json:"type"`
+	ErrorType string      `json:"error_type"`
+	Message   string      `json:"error"`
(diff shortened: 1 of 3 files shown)


zqd/api/api.go
```diff
@@ -13,9 +13,10 @@ import (
 )
 
 type Error struct {
-	Type    string      `json:"type"`
-	Message string      `json:"error"`
-	Info    interface{} `json:"info,omitempty"`
+	Type      string      `json:"type"`
+	ErrorType string      `json:"error_type"`
```
alfred-landrum, line 17: This could be useful, but I don't see it utilized in the PR?

henridf: ooops... that's an omission that i've fixed in a follow-on commit. the purpose of this field is to hold the "Kind" string which was previously being written into Type. (This field could also be called Kind instead of ErrorType btw... happy to change if any strong preference).

alfred-landrum: Ok! My naming preference is `Kind`, then.

PR #616, opened 2020-04-21
More info in autodetect error

### LINKED ISSUE #494: Improve auto-detection errors
Autodetection is working as intended in terms of autodetection. However, it isn't providing very helpful feedback upon failures. The reason for this is that it tries each format in turn and returns a "malformed input" if none succeed.

Here's an example of how autodetection occludes the useful error message:

```
19:03 ~/work/looky/zq(no-dupe-record-fields)
$ cat !$
cat ~/tmp/stuff.zson
#0:record[foo:record[bar:string,bar:string]]
0:[["1";"2";]]
19:03 ~/work/looky/zq(no-dupe-record-fields)
$ zq ~/tmp/stuff.zson 
malformed input
19:03 ~/work/looky/zq(no-dupe-record-fields)
$ zq -i zng ~/tmp/stuff.zson 
line 1: duplicate fields in record type
```

A possible solution here would be:

* Each reader needs to distinguish between syntax errors (signaled via the SyntaxError err) and higher-level semantic errors such as duplicate fields (above), or a value not matching its record type, etc. We may already be doing this perfectly but should check as part of this issue.
* The autodetector should keep each reader's error, and upon failure of all readers, if any error is a non-Syntax error, that one should be used.

Other options:
* Collect all errors, and output a summary (“ndjson error: x, zeek error: y, …”)
* Use the file extension as a hint of which candidate reader’s error to return. For example, if the filename is foo.zng, then use the zng reader’s error message. This works better for some formats than others, for example .json could be ndjson or zjson; .log could be ndjson or zeek.

These are non-exclusive: a good solution might combine elements of more than one of these. And there might of course be other ways to improve this.

henridf:
This commit changes the detector error message to include each attempted format's error.

closes #494

### DIFF
--- a/zio/detector/reader.go
+++ b/zio/detector/reader.go
@@ -18,9 +18,7 @@ func NewReader(r io.Reader, zctx *resolver.Context) (zbuf.Reader, error) {
 	recorder := NewRecorder(r)
 	track := NewTrack(recorder)
 
-	var zngErr, zeekErr, ndjsonErr, zjsonErr, bzngErr error
-
-	zngErr = match(zngio.NewReader(track, resolver.NewContext()), "tzng")
+	zngErr := match(zngio.NewReader(track, resolver.NewContext()), "tzng")
 	if zngErr == nil {
 		return zngio.NewReader(recorder, zctx), nil
 	}
@@ -30,14 +28,14 @@ func NewReader(r io.Reader, zctx *resolver.Context) (zbuf.Reader, error) {
 	if err != nil {
 		return nil, err
 	}
-	zeekErr = match(zr, "zeek")
+	zeekErr := match(zr, "zeek")
 	if zeekErr == nil {
 		return zeekio.NewReader(recorder, zctx)
 	}
 	track.Reset()
 
 	// zjson must come before ndjson since zjson is a subset of ndjson
-	zjsonErr = match(zjsonio.NewReader(track, resolver.NewContext()), "zjson")
+	zjsonErr := match(zjsonio.NewReader(track, resolver.NewContext()), "zjson")
 	if zjsonErr == nil {
 		return zjsonio.NewReader(recorder, zctx), nil
 	}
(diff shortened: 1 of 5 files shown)


zio/zeekio/reader.go
```diff
@@ -34,11 +34,18 @@ func NewReader(reader io.Reader, zctx *resolver.Context) (*Reader, error) {
 }
 
 func (r *Reader) Read() (*zng.Record, error) {
+	e := func(err error) error {
```
alfred-landrum, line 37: 👍

zio/detector/reader.go
```diff
@@ -11,50 +11,68 @@ import (
...
 )
 
-var ErrUnknown = errors.New("malformed input")
-
 func NewReader(r io.Reader, zctx *resolver.Context) (zbuf.Reader, error) {
 	recorder := NewRecorder(r)
 	track := NewTrack(recorder)
-	if match(zngio.NewReader(track, resolver.NewContext())) {
+
+	var zngErr, zeekErr, ndjsonErr, zjsonErr, bzngErr error
```
alfred-landrum, line 21: You can use short variable declaration for these in their respective `match` calls.

zio/detector/reader.go
```diff
@@ -11,50 +11,68 @@ import (
...
+	return nil, joinErrs([]error{zngErr, zeekErr, ndjsonErr, zjsonErr, bzngErr})
 }
 
-func match(r zbuf.Reader) bool {
+func joinErrs(errs []error) error {
+	s := "format detection error"
+	for _, e := range errs {
+		s += "\n\t" + e.Error()
+	}
+	//	fmt.Println(s)
```
alfred-landrum, line 69: nit: remove

alfred-landrum:
I was curious what this looks like, so in case anyone else wonders:

```
$ ./dist/zq ./cmd/zq/main.go
./cmd/zq/main.go: format detection error
	tzng: line 1: bad format
	ndjson: line 1: Unknown value type
	zjson: line 1: invalid character 'p' looking for beginning of value
	zng: zng descriptor out of range
```

zio/detector/reader.go
henridf: oops

PR #617, opened 2020-04-21
Use bzng time index in zqd

### LINKED ISSUE #595: Use (b)zng time index in zqd
An time index on bzng files was added in #579 but it is not yet used inside zqd, partly because there's currently no persistent data structure for spaces where the index should be kept.
This issue is to do whatever plumbing is necessary to be able to use these indexes in zqd spaces.

aswan:
Closes #595

### DIFF
--- a/zqd/space/space.go
+++ b/zqd/space/space.go
@@ -39,6 +39,8 @@ type Space struct {
 	conf config
 }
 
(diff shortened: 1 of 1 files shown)


nwt:
Locking? Space deletion?

PR #618, opened 2020-04-21
Add gzipped log post test

henridf:


zqd/handlers_test.go
```diff
@@ -398,32 +399,46 @@ func TestPostNDJSONLogs(t *testing.T) {
...
-	require.NoError(t, err)
-	require.Equal(t, &api.SpaceInfo{
-		MinTime:       &min,
-		MaxTime:       &max,
-		Name:          spaceName,
-		Size:          80,
-		PacketSupport: false,
-	}, info)
+		res := zngSearch(t, client, spaceName, "*")
+		const expected = "#0:record[_path:string,ts:time,uid:bstring]\n0:[http;2;CXY9a54W2dLZwzPXf1;]\n0:[http;1;CXY9a54W2dLZwzPXf1;]"
```
aswan, line 417: nit: maybe hoist this to the top next to `src` so its easy for people reading the test to compare the two?

zqd/handlers_test.go
```diff
@@ -398,32 +399,46 @@ func TestPostNDJSONLogs(t *testing.T) {
 			ndjsonio.Rule{"_path", "http", "http_log"},
 		},
 	}
-	_, client, done := newCore(t)
-	defer done()
-	const spaceName = "test"
 
-	_, err := client.SpacePost(context.Background(), api.SpacePostRequest{Name: spaceName})
-	require.NoError(t, err)
+	test := func(t *testing.T, input string) {
```
aswan, line 403: nit: Since `t` is in the enclosing scope, you don't need to pass it again here...

zqd/handlers_test.go
henridf: 👍

PR #620, opened 2020-04-21
Support =~ operator in zql expressions

aswan:


expr/expr.go
```diff
@@ -100,6 +102,8 @@ func compileNative(node ast.Expression) (NativeEvaluator, error) {
 			return compileIn(lhsFunc, rhsFunc)
 		case "=", "!=":
 			return compileCompareEquality(lhsFunc, rhsFunc, n.Operator)
+		case "=~":
```
henridf, line 105: Are we considering `!~` too at some point?

aswan: I hadn't really thought about it.  We can always write `! (foo =~ pattern)`, but a new operator would be easy enough to add...

PR #628, opened 2020-04-22
Make bzng time index handle descending timestamps

### LINKED ISSUE #625: Time index doesn't handle files sorted by descending timestamps
The bzng time index introduced in #548 assumes that the file being indexed has ascending timestamps.  But of course zqd stores all.bzng with descending timestamps.

aswan:
Closes #625

### DIFF
--- a/zio/zngio/index.go
+++ b/zio/zngio/index.go
@@ -163,10 +163,10 @@ func newRangeReader(f *os.File, zctx *resolver.Context, order Ordering, index []
 			if order == OrderAscending && mark.Ts > span.Ts {
 				break
 			}
-			off = mark.Offset
-			if order == OrderDescending && mark.Ts > span.Ts {
+			if order == OrderDescending && mark.Ts < span.End() {
 				break
 			}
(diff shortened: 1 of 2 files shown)


zio/zngio/index.go
```diff
@@ -51,12 +61,14 @@ func (ti *TimeIndex) NewReader(f *os.File, zctx *resolver.Context, span nano.Spa
...
-	marks   []mark
-	lastSOS int64
-	lastTs  nano.Ts
+	start         nano.Ts
+	end           nano.Ts
+	parent        *TimeIndex
+	order         Ordering
+	marks         []mark
+	lastSOS       int64
+	lastTs        nano.Ts
```
henridf, line 70: nit: it might be more descriptive to call this `lastSoSTs` instead of `lastTs`

aswan: But that's not what this is -- it is the timestamp of the last read record of any sort.

henridf: oh nvm

PR #631, opened 2020-04-22
Add space.Manager

### LINKED ISSUE #624: implement manager for spaces
We need a manager of sorts in the `space` package for zqd, that centralizes the acts of creating, acquiring, and deleting `Space`'s, as well as listing the available spaces. A strawman as to what it's api would look like would match:

```go
type Manager interface {
   Create(ctx, options) (SpaceInfo, error)
   Get(ctx, name) *Space
   Delete(ctx, name) error
   List(ctx) []SpaceInfo
}
``` 

It would get created at zqd startup, and new'd during the Core setup. The `spaceOps` state that's currently in zqd/core can move into the space manager.

aswan:
This new struct is a centralized place to keep track of existing spaces.
The public api surface from Space is still quite large and can probably
be shrunk, but this can happen in follow-up patches.

Closes #624


zqd/space/space.go
```diff
@@ -34,59 +35,199 @@ var (
...
+
+	wg sync.WaitGroup
+	// closed to signal non-delete ops should terminate
+	cancelChan chan struct{}
 }
 
-func Open(root, name string) (*Space, error) {
-	path := filepath.Join(root, name)
-	c, err := loadConfig(path)
+func NewManager(root string) *Manager {
```
alfred-landrum, line 57: I suggest NewManager takes a zap.Logger instance that is stored on the Manager (and used at the "log somewhere" comments), and that NewManager returns `(*Manager, error)`.

zqd/space/space.go
```diff
@@ -34,59 +35,199 @@ var (
...
-	// XXX this should be validated before reaching here.
+func (m *Manager) Create(name, dataPath string) (api.SpacePostResponse, error) {
+	_, exists := m.spaces.Load(name)
+	if exists {
+		return api.SpacePostResponse{}, ErrSpaceExists
+	}
+
 	if name == "" && dataPath == "" {
-		return nil, errors.New("must supply non-empty name or dataPath")
+		return api.SpacePostResponse{}, errors.New("must supply non-empty name or dataPath")
```
alfred-landrum, line 96:
```suggestion
		return api.SpacePostResponse{}, zqe.E(zqe.Invalid, "must supply non-empty name or dataPath")
```
(I know the error was existing code)

zqd/space/space.go
```diff
@@ -34,59 +35,199 @@ var (
...
+		mgr.spaces.Store(space.Name(), &space)
 	}
-	return &Space{path, c}, nil
+
+	return mgr
 }
 
-func Create(root, name, dataPath string) (*Space, error) {
-	// XXX this should be validated before reaching here.
+func (m *Manager) Create(name, dataPath string) (api.SpacePostResponse, error) {
```
alfred-landrum, line 89:
```suggestion
func (m *Manager) Create(name, dataPath string) (*api.SpacePostResponse, error) {
```

zqd/space/space.go
```diff
@@ -34,59 +35,199 @@ var (
...
+}
+
+func (m *Manager) Delete(name string) error {
+	space, err := m.Get(name)
+	if err != nil {
+		return err
+	}
+
+	err = space.delete()
+	m.spaces.Delete(name)
```
alfred-landrum, line 155: I think we should only delete it from the name map if the delete succeeds. It looks like we check for `os.IsExist` in the right places in Create, but if there's some partial delete error, I'd rather a user complain that they have a space they can't delete, than the more mysterious failure that they can't create a space with a pre-existing space name.

```suggestion
	if err != nil {
	  m.spaces.Delete(name)
	}
```

zqd/space/space.go
```diff
@@ -34,59 +36,207 @@ var (
...
+	mgr := &Manager{
+		rootPath: root,
+		logger:   logger,
+	}
+
+	dirs, err := ioutil.ReadDir(root)
 	if err != nil {
-		if os.IsNotExist(err) {
-			return nil, ErrSpaceNotExist
+
```
alfred-landrum, line 67:
```suggestion
```

zqd/space/space.go
```diff
@@ -34,59 +36,207 @@ var (
...
+		path:       path,
+		conf:       c,
+		cancelChan: make(chan struct{}, 0),
+	}
+	m.spaces.Store(name, &space)
+	return &api.SpacePostResponse{
+		Name:    name,
+		DataDir: dataPath,
+	}, nil
+
```
alfred-landrum, line 137:
```suggestion
```

zqd/space/space.go
```diff
@@ -34,59 +36,207 @@ var (
 	ErrSpaceNotExist       = zqe.E(zqe.NotFound, "space does not exist")
 )
 
+type Manager struct {
+	rootPath string
+	spaces   sync.Map
```
nwt, line 41: sync.Map is not the right tool for this job. As https://pkg.go.dev/sync?tab=doc#Map says:

> The Map type is specialized. Most code should use a plain Go map instead, with separate locking or coordination, for better type safety and to make it easier to maintain other invariants along with the map content.

That's borne out by the race in Create.

PR #632, opened 2020-04-23
open output files with truncation

mccanne:
This PR fixes a bug where the zng.Writer output files
were opened without truncation.  This would cause corrupt
files to be created when overlaying new but shorter output
on an existing file.


alfred-landrum:
Searching for usages of `os.OpenFile`, I see some other cases of this. One close by in emitter/types.go, and in pcap/ts/command.go & pcap/ts/command.go; would you adjust those as well please?

PR #637, opened 2020-04-23
port last go tests under tests/suite to yaml

mccanne:


tests/suite.go
```diff
@@ -1,25 +0,0 @@
...
-
-import (
-	"github.com/brimsec/zq/pkg/test"
-	"github.com/brimsec/zq/tests/suite/jsontype"
-	"github.com/brimsec/zq/tests/suite/zeek"
-)
-
-var RootDir = "./test-root"
-
-var internals = []test.Internal{
```
henridf, line 11: All these names UnionIncompat, TestInferPath, ... are at least somewhat descriptive... can we use those in the new yaml file names instead of test1.yaml, test2.yaml, etc?

PR #639, opened 2020-04-23
Cut -c

### LINKED ISSUE #497: Add "complement" option to cut processor
`zeek-cut` has this option:

```
-n       Print all fields except those specified.
```

Also available in GNU `cut` (as `--complement`). 

ZQL `cut` should have it too.

henridf:
Add support for the `-c` ("complement") option to cut, where all record fields _not_ in the provided field list are to be passed through.

closes #497

### DIFF
--- a/proc/cut.go
+++ b/proc/cut.go
@@ -31,7 +31,11 @@ func CompileCutProc(c *Context, parent Proc, node *ast.CutProc) (*Cut, error) {
 	if err != nil {
 		return nil, fmt.Errorf("compiling cut: %w", err)
 	}
-	builder, err := NewColumnBuilder(c.TypeContext, node.Fields)
+	var fields []string
+	for _, field := range node.Fields {
+		fields = append(fields, expr.FieldExprToString(field))
+	}
+	builder, err := NewColumnBuilder(c.TypeContext, fields)
 	if err != nil {
 		return nil, fmt.Errorf("compiling cut: %w", err)
 	}
--- a/proc/cut.go
+++ b/proc/cut.go
@@ -38,6 +38,16 @@ func CompileCutProc(c *Context, parent Proc, node *ast.CutProc) (*Cut, error) {
 	for _, field := range node.Fields {
 		fields = append(fields, expr.FieldExprToString(field))
 	}
+
+	_, err := expr.CompileFieldExprs(node.Fields)
+	if err != nil {
+		return nil, fmt.Errorf("compiling cut: %w", err)
+	}
+	_, err = NewColumnBuilder(c.TypeContext, fields)
+	if err != nil {
+		return nil, fmt.Errorf("compiling cut: %w", err)
+	}
+
 	return &Cut{
 		Base:        Base{Context: c, Parent: parent},
 		complement:  node.Complement,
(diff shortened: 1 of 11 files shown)


proc/cut.go
```diff
@@ -27,96 +34,163 @@ type Cut struct {
 // do this now since it might confuse users who expect to see output
 // fields in the order they specified.
 func CompileCutProc(c *Context, parent Proc, node *ast.CutProc) (*Cut, error) {
-	resolvers, err := expr.CompileFieldExprs(node.Fields)
+	var fields []string
+	for _, field := range node.Fields {
+		fields = append(fields, expr.FieldExprToString(field))
+	}
```
aswan, line 40: Seems like between here and groupby, we have enough cases where procs operate on dotted field names that it would be nice to have the parser just put the full dotted field name into the ast rather than working backward from a tree of FieldExprs here.
That could easily be done in a follow-up though.

proc/cut.go
```diff
@@ -27,96 +34,163 @@ type Cut struct {
...
+
+	_, err := expr.CompileFieldExprs(node.Fields)
 	if err != nil {
 		return nil, fmt.Errorf("compiling cut: %w", err)
 	}
-	builder, err := NewColumnBuilder(c.TypeContext, node.Fields)
+	_, err = NewColumnBuilder(c.TypeContext, fields)
```
aswan, line 46: Can we do this only if complement is false?  And add a comment explaining that this is for the purpose of error checking?

proc/cut_test.go
```diff
@@ -58,6 +65,21 @@ func TestCut(t *testing.T) {
 	proc.TestOneProc(t, fooAndBar, fooAndBar, "cut foo,bar")
 }
 
+func TestCutComplement(t *testing.T) {
```
aswan, line 68: how about a test of `cut -c` on a record?

zql/zql.peg
```diff
@@ -387,8 +387,10 @@ top
 procLimitArg
   = _ "-limit" _ limit:unsignedInteger { RETURN(limit) }
 
+cutArg = (_ "-c" { RETURN(makeArg("c", NULL)) })*
```
aswan, line 390: why the * ?

proc/cut_test.go
henridf: added to TestCutNested

zql/zql.peg
henridf: This  is done in other args... for example `sort -r -r -r` is gramatically valid but rejected in the parser-support builders. I'm just following the pattern and if we think this should be changed i don't mind but wholesale...

proc/cut.go
henridf: Agreed. I think this should be done outside of this PR though. Will file an issue.

zql/zql.peg
aswan: Right, this was because sort accepts a few different arguments.  The parser used to have explicit rules for the different orderings of the different args, but there's a combinatorial issue there as new arguments are added.
Anyway, agreed that we can just keep the pattern here...

henridf: yeah, i think anything that avoids our inscrutable parser errors is a win. (separately, i think `-x -x -x` should be handled just like `-x` and shouldn't be an error... but getting off topic).

proc/cut.go
henridf: https://github.com/brimsec/zq/issues/643

PR #647, opened 2020-04-23
Enabling time indexing in zqd

aswan:


nwt:
We should make zngio.TimeIndex safe for concurrent use before merging this.

aswan:
> We should make zngio.TimeIndex safe for concurrent use before merging this.

can you elaborate on what's unsafe?

nwt:
TimeIndex.indexReady is a shared variable and needs synchronization.

nwt:
Didn't mean to close this!

aswan:
> TimeIndex.indexReady is a shared variable and needs synchronization.

It is a boolean that starts out as false and exactly once in its life switches to true.  By design, everything works (sup-optimally but still correctly) if due to a race, a reader reads the old (false) value while a writer is in the process of writing the value true.
Maybe its a bit sloppy but I don't think there's a correctness issue.

alfred-landrum:
> TimeIndex.indexReady is a shared variable and needs synchronization.

I think more than just sync'ing on indexReady itself, we could ensure that there's only a single indexReader active at any given time.

aswan:
> > TimeIndex.indexReady is a shared variable and needs synchronization.
> 
> I think more than just sync'ing on indexReady itself, we could ensure that there's only a single indexReader active at any given time.

We could, its not clear if that's a good idea.  Would you make a new reader wait for the existing indexReader to complete?  If the new reader is going to read from at or near the beginning of the file, then its not going to get much benefit.  And if the existing reader is slow, it might have to wait for some time.

I viewed this as a minimal quick-and-dirty implementation -- if we want to optimize this I think that persisting the index to disk would give us more bang for the buck.

alfred-landrum:
> We could, its not clear if that's a good idea. Would you make a new reader wait for the existing indexReader to complete? If the new reader is going to read from at or near the beginning of the file, then its not going to get much benefit. And if the existing reader is slow, it might have to wait for some time.
good points!

> I viewed this as a minimal quick-and-dirty implementation -- if we want to optimize this I think that persisting the index to disk would give us more bang for the buck.

agreed. in that case, would you just make the simpler change of adding some simple sync to indexReady?

aswan:
> would you just make the simpler change of adding some simple sync to indexReady?

what problem are we trying to guard against exactly?

alfred-landrum:
In Go, there is no guarantee that a goroutine will observe a write to a
variable made by another goroutine without explicit synchronization:
https://golang.org/ref/mem#tmp_10

Not having any synchronization on indexReady means that we could
conceivably keep recreating the index, over and over, because every api
request happens in a different goroutine. Yes, it's unlikely, but I would
prefer not to have to worry about it. So, please add the explicit
synchronization.

aswan:
> In Go, there is no guarantee that a goroutine will observe a write to a variable made by another goroutine without explicit synchronization: https://golang.org/ref/mem#tmp_10 Not having any synchronization on indexReady means that we could conceivably keep recreating the index, over and over, because every api request happens in a different goroutine. Yes, it's unlikely, but I would prefer not to have to worry about it. So, please add the explicit synchronization.

Okay, your concern is different from Noah's original concern.  This seems like massive overkill to me, but whatever...

nwt:
> Okay, your concern is different from Noah's original concern. This seems like massive overkill to me, but whatever...
It's the same concern.

aswan:
The only issue I see here is the possibility that a new reader is created at the same moment the original index reader is finishing and the compiler re-orders the writes to indexReady/order/index and the new reader manages to see indexReady=true but not see the actual index.  I'll add handling for this.
Maybe I misunderstood, but it seems like Al is describing a different concern.

nwt:
I think we have the same fundamental concern: unsynchronized access to shared variables is a bug.

Yes, the zngio.TimeIndex code will probably always work as intended in spite of missing synchronization. But adding synchronization will guarantee it always works as intended.

aswan:
> unsynchronized access to shared variables is a bug.

can we bit a little more precise?  there are two distinct things here:
1. Different goroutines are reading and writing the indexReady variable simultaneously.  We discussed this previously, I don't believe there's a bug here.
2. Re-ordering of reads and writes (which would specifically manifest as I described in my previous comment).  The go "memory model" docs are awfully vague here, its not clear to me that adding a lock actually addresses this.

zio/zngio/index.go
```diff
@@ -21,6 +22,7 @@ const (
 )
 
 type TimeIndex struct {
+	lock       sync.Mutex
```
nwt, line 25: Nit: Our traditional name for a sync.Mutex or sync.RWMutex is mu.

aswan: 🐄

mattnibs: Took me a second to get this, but nice.

henridf: I mean, we could go all in and use μ, which is a valid name in Go...

PR #651, opened 2020-04-24
tzng: don't panic on unterminated containers

### LINKED ISSUE #638: tzng parser panics on malformed record
```sh
10:45 ~/work/brim/zq(master)
$ git rev-parse --short HEAD
6a05953
10:45 ~/work/brim/zq(master)
$ cat blar.zng
#0:record[blar:string]
0:[blar1]
10:45 ~/work/brim/zq(master)
$ zq blar.zng
panic: runtime error: slice bounds out of range [7:6]
```

henridf:
closes #638

### DIFF
--- a/tests/suite/errors/nosemi.yaml
+++ b/tests/suite/errors/nosemi.yaml
@@ -1,12 +1,7 @@
-script: zq in.tzng
+zql: filter *
 
-inputs:
-  - name: in.tzng
-    data: |
-        #0:record[a:string]
-        0:[1]
+input: |
+  #0:record[x:int32]
+  0:[1]
 
-outputs:
-  - name: stderr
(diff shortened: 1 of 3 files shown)


tests/suite/errors/nosemi.yaml
```diff
@@ -0,0 +1,12 @@
+script: zq in.tzng
```
aswan, line 1: nit: it looks like this test can be written without script and with errorRE like this one: https://github.com/brimsec/zq/blob/master/tests/suite/put/put-6.yaml which would be a little more concise

henridf: thanks, i'll do that, i'm not keeping up with our test capabilities...

PR #654, opened 2020-04-24
make -f zng the default for zng output

mccanne:
This commit makes (binary) zng the default for zq and zdx lookup.
It also adds a check to see if binary data is being written to
a terminal and exists with an error. -B overrides this check and
-t was added as a shortcut for "-f tzng" to save typing when
you want the text output.

There is code replicated between the two commands.   We should 
refactor this so there is a common package ("tools" or "zq") used 
across all the related zq tools to do common arg parsing, zng 
file reader/writer setup, schema-map config set, and so forth.

### DIFF
--- a/cmd/zdx/lookup/command.go
+++ b/cmd/zdx/lookup/command.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"os"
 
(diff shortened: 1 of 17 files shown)


philrz:
@mccanne: A plea for your list of things to include in the effort to have shared "common" stuff across all the CLI tools: The `-version` option (if possible). Thanks!

PR #657, opened 2020-04-25
Consistent span usage in api

### LINKED ISSUE #541: Change space.Info Min/Max Time to Span
In the search request we represent a time range as a `nano.Span`. For consistency sake, `api.SpaceInfo` should do the same.

mattnibs:
Make the api consistent by leveraging nano.Span when expressing time
ranges.

Note: I didn't change the info file to a span. This would require
a migration for older installations. I figure this isn't worth the
effort as at some point we'll probably change the way determine the
time span of a space.

closes #541

### DIFF
--- a/zqd/ingest/log.go
+++ b/zqd/ingest/log.go
@@ -125,7 +125,7 @@ func ingestLogs(ctx context.Context, pipe *api.JSONPipe, s *space.Space, req api
 	if tailW.r != nil {
 		min := nano.Min(tailW.r.Ts, headW.r.Ts)
 		max := nano.Max(tailW.r.Ts, headW.r.Ts)
-		if err = s.SetSpan(nano.NewSpanTs(min, max)); err != nil {
+		if err = s.SetSpan(nano.NewSpanTs(min, max+1)); err != nil {
 			return err
 		}
 	}
--- a/zqd/ingest/log.go
+++ b/zqd/ingest/log.go
@@ -123,7 +123,9 @@ func ingestLogs(ctx context.Context, pipe *api.JSONPipe, s *space.Space, req api
 		return err
 	}
 	if tailW.r != nil {
-		if err = s.SetTimes(tailW.r.Ts, headW.r.Ts); err != nil {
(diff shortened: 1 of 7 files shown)


zqd/space/space.go
```diff
@@ -422,6 +422,10 @@ type info struct {
 	MaxTime nano.Ts `json:"max_time"`
 }
 
+func (i info) Span() nano.Span {
+	return nano.NewSpanTs(i.MinTime, i.MaxTime)
+}
+
 // UnsetTimes nils out the cached time range value for the space.
 // XXX For right now this simply deletes the info file as nothing else is stored
 // there. When we get to brimsec/zq#541 the time range should be represented as
```
henridf, line 431: Looks like this comment should be updated if this isn't being done in #541

zqd/ingest/log.go
```diff
@@ -123,7 +123,9 @@ func ingestLogs(ctx context.Context, pipe *api.JSONPipe, s *space.Space, req api
 		return err
 	}
 	if tailW.r != nil {
-		if err = s.SetTimes(tailW.r.Ts, headW.r.Ts); err != nil {
+		min := nano.Min(tailW.r.Ts, headW.r.Ts)
+		max := nano.Max(tailW.r.Ts, headW.r.Ts)
+		if err = s.SetSpan(nano.NewSpanTs(min, max)); err != nil {
```
henridf, line 128: This span is incorrect as spans are open on the right... I think you need to make it `NewSpanTs(min, max+1)` ?

zqd/space/space.go
mattnibs: Yep

zqd/ingest/log.go
mattnibs: I think you are right.

mattnibs, line 128:
```suggestion
		if err = s.SetSpan(nano.NewSpanTs(min, max+1)); err != nil {
```

henridf:
Does this have a companion issue over on the brim repo?

mattnibs:
> Does this have a companion issue over on the brim repo?

I will do this

PR #660, opened 2020-04-25
Changelog for v0.12.0

philrz:
https://github.com/brimsec/zq/compare/v0.11.1...a2de891

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## v0.12.0
+* zql: Introduce `=~` and `!~` operators in filters for globs, regexps, and matching addresses against subnets (#604, #620)
+* zq: When input auto-detect fails, include each attempted format's error (#616)
+* zng: Binary format is now called "ZNG" and text format is called "TZNG" ("BZNG" has been retired) (#621, #630, #656)
+* zql: `cut` now has a `-c` option to show all fields _not_ in the provided list (#639, #655)
+* zq: Make `-f zng` (binary ZNG) the default `zq` output format, and introduce `-t` as shorthand for `-f tzng` (#654)
+
 ## v0.11.1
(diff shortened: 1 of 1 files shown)


henridf:
Just curious, are we waiting for features to show up in Brim before mentioning them in the Changelog? (question prompted by the absence of the recently-added zqd log post endpoint).

philrz:
@henridf: Yes, that has pretty much been my approach. I've been crafting the Release Notes as if for an audience of end users of the Brim app & `zq`, respectively, as opposed to outside developers. I figure developers are more likely to expect that the comprehensive record of what changed is in the `git log` and the code itself.

I've not really gone out of my way to describe `zqd` in terms of the services it exposes. Since the Brim app is currently the only client that relies on it, its services have generally been exposed via the Brim app release notes about the enabled features. If we'd documented the `zqd` REST API and had a known audience of developers actively coding against it I might be more inclined to expose `zqd` & API changes as they arrive only at the backend.

All that said, this is an evolving process and we could certainly discuss changing it.

henridf:
Sounds good. Thanks for the clarification. Part of the reason I asked was the presence of low-level zqd API entries for example in v0.11.1 (and other zqd entries) that made it appear that zqd was being represented.

> All that said, this is an evolving process and we could certainly discuss changing it.

My 2c: I think the usual expectation of a changelog is that it has coverage of all components in a repo, whether directly exposed or used as a library/supporting tool by other things.  At least that's my expectation when I go changelog-fishing for some historical information (date, PR numbers, overall sequencing) in a repo I'm unfamiliar with.

If there's a reason for some components to be excluded, then I'd suggest flagging those explicitly somewhere,  maybe at top the changelog. Otherwise, with implicit, selective coverage, it's a less useful history of the repo. 

I do realize that overall this topic is less straightforward in zq with its multiple tools (and growing) than in a more single-purpose repo!

PR #661, opened 2020-04-26
add zar mkdirs/rmdirs

mccanne:
This commit makes the creation of zar directories in an archive
explicit and adds the command "zar mkdirs" to create directories
and "zar rmdirs" to remove them.

We also added /bin and /usr/bin to the PATH in ztest script commands
so ls, find, etc now work.

### DIFF
--- a/archive/finder.go
+++ b/archive/finder.go
@@ -39,7 +39,7 @@ func Search(zardir string, rule Rule, pattern string, skipMissing bool) (bool, e
 			// No index for this rule.  Skip it if the skip boolean
 			// says it's ok.  Otherwise, we return ErrNotExist since
 			// the client was looking for something that wasn't indexed,
-			// and they probabl want to know.  T
+			// and they would probably want to know.
 			err = nil
 		} else {
 			err = fmt.Errorf("%s: %w", finder.Path(), err)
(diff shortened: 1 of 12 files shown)


archive/finder.go
```diff
@@ -17,46 +16,30 @@ import (
...
+func Search(zardir string, rule Rule, pattern string, skipMissing bool) (bool, error) {
+	finder := rule.NewFinder(zardir)
 	keyType, err := finder.Open()
 	if err != nil {
-		if err == os.ErrNotExist {
+		if err == os.ErrNotExist && skipMissing {
+			// No index for this rule.  Skip it if the skip boolean
+			// says it's ok.  Otherwise, we return ErrNotExist since
+			// the client was looking for something that wasn't indexed,
+			// and they probabl want to know.  T
```
mattnibs, line 42:
```suggestion
			// and they probably want to know.
```

cmd/zar/rmdirs/command.go
```diff
@@ -0,0 +1,43 @@
...
+	"flag"
+
+	"github.com/brimsec/zq/archive"
+	"github.com/brimsec/zq/cmd/zar/root"
+	"github.com/mccanne/charm"
+)
+
+var RmDirs = &charm.Spec{
+	Name:  "rmdirs",
+	Usage: "rmdirs [-d <dir>]",
```
alfred-landrum, line 14:
```suggestion
	Usage: "rmdirs <dir>",
```

PR #663, opened 2020-04-27
driver.Run(): take a <-chan time.Time rather than a time.Duration

### LINKED ISSUE #554: zq: Pass null ticker to (*MuxOutput).Pull(..)
We currently pass a 10s ticker to `(*MuxOutput).Pull(..)` in zq and tests. This doesn't seem necessary, but changing that to a nil ticker breaks some tests. Investigate and fix.

henridf:
Following @nwt and @mattnibs suggestion in https://github.com/brimsec/zq/pull/549#pullrequestreview-391636896

closes #554

### DIFF
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -17,16 +17,11 @@ type Driver interface {
 	Stats(api.ScannerStats) error
 }
 
-func Run(out *MuxOutput, d Driver, statsInterval time.Duration) error {
+func Run(out *MuxOutput, d Driver, statsTicker <-chan time.Time) error {
 	//stats are zero at this point.
 	var stats api.ScannerStats
-	if statsInterval == 0 {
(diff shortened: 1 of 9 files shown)


driver/driver.go
```diff
@@ -17,16 +17,11 @@ type Driver interface {
 	Stats(api.ScannerStats) error
 }
 
-func Run(out *MuxOutput, d Driver, statsInterval time.Duration) error {
+func Run(out *MuxOutput, d Driver, statsTicker <-chan time.Time) error {
```
nwt, line 20: Nit: This name reads like it's a time.Ticker, which it isn't. I think statsTicks or statsTickCh would be a better name.

henridf: https://github.com/brimsec/zq/pull/669

PR #667, opened 2020-04-27
Remove time-sorting from MergeProc

### LINKED ISSUE #648: Add and use a new mux proc
During post-discussion of the deadlock fixing work in https://github.com/brimsec/zq/pull/611 , we said that we take a different approach: The existing `proc.Merge` was created to simultaneously a) multiplex parent streams while b) ensuring the output records are in sorted order by timestamp. `proc.Merge` makes more assumptions that we need for a general case multiplexer, so we should create a new `proc.Mux` that more simply returns whatever data is available, w/o caring about ordering or availability of data from parents. `proc.Mux` should then replace the current usage of `proc.Merge`.

aswan:
To avoid confusion with MuxOutput, I've retained the name merge here,
this just removes all the time-sorting logic from and simplifies the
merge proc.

Closes #648

### DIFF
--- a/proc/merge.go
+++ b/proc/merge.go
@@ -16,6 +16,7 @@ type Merge struct {
 	Base
 	once     sync.Once
 	ch       <-chan Result
+	doneCh   chan struct{}
 	parents  []*runnerProc
 	nparents int
 }
@@ -23,14 +24,14 @@ type Merge struct {
 type runnerProc struct {
 	Base
 	ch     chan<- Result
-	doneCh chan struct{}
+	doneCh <-chan struct{}
 }
 
-func newrunnerProc(c *Context, parent Proc, ch chan<- Result) *runnerProc {
+func newrunnerProc(c *Context, parent Proc, ch chan<- Result, doneCh <-chan struct{}) *runnerProc {
 	return &runnerProc{
 		Base:   Base{Context: c, Parent: parent},
 		ch:     ch,
-		doneCh: make(chan struct{}, 1),
+		doneCh: doneCh,
 	}
 }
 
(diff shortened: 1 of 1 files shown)


proc/merge.go
```diff
@@ -17,46 +15,40 @@ import (
...
+		Base: Base{Context: c, Parent: parent},
+		ch:   ch,
 	}
 }
 
 func (r *runnerProc) run() {
 	for {
 		batch, err := r.Get()
 		r.ch <- Result{batch, err}
-		_, ok := <-r.proceed
```
alfred-landrum, line 43: Without the proceed channel or some equivalent, these runnerProc goroutines will leak.

proc/merge.go
```diff
@@ -35,6 +35,9 @@ func (r *runnerProc) run() {
 	for {
 		batch, err := r.Get()
 		r.ch <- Result{batch, err}
+		if batch == nil && err == nil {
```
alfred-landrum, line 38: Should this be `EOS(batch, err)`?

proc/merge.go
```diff
@@ -68,85 +63,17 @@ func (m *Merge) Parents() []Proc {
...
-	res := m.bufs[pick]
-	m.reload(pick)
-	if m.err != nil {
-		m.Done()
-		return nil, m.err
-	}
-	return res, nil
-}
-
-func (m *Merge) Done() {
```
alfred-landrum, line 143: From yesterday's slack discussion, I think I learned that Merge still needs a `Done()` call that will signal the runnerProc's to call Done() on their parent.

aswan:
I've added Done(), and also additional cases that weren't handled properly before such as one branch producing an error.

Ideally this would have tests, but this is going to be difficult since we've made the behavior of merge non-deterministic -- results from different branches are merged as they arrive but they are processed by separate goroutines and we can't make any assumptions about the order results are generated and hence the final output.

proc/merge.go
```diff
@@ -16,47 +14,55 @@ import (
...
 
-func newrunnerProc(c *Context, parent Proc) *runnerProc {
+func newrunnerProc(c *Context, parent Proc, ch chan<- Result) *runnerProc {
 	return &runnerProc{
-		Base:    Base{Context: c, Parent: parent},
-		ch:      make(chan Result),
-		proceed: make(chan struct{}),
+		Base:   Base{Context: c, Parent: parent},
+		ch:     ch,
+		doneCh: make(chan struct{}, 1),
```
nwt, line 33: There's no need for a separate done channel per runnerProc. Just add `doneCh chan struct{}` to Merge, pass it to newrunnerProc where you then select on <-doneCh, and do `close(m.doneCh)` in (*Merge).Done.

aswan: That's much nicer, thanks

PR #668, opened 2020-04-27
zqd api renaming: s/Packet/Pcap

### LINKED ISSUE #562: Rename packet endpoints to pcap
With the decision to refer to packet captures as "pcaps", these name of these two endpoints need to change:
- `GET /space/:space/packet` -> `/space/:space/pcap`
- `POST /space/:space/packet` -> `/space/:space/pcap`

There's also some api payloads that need to change:
- `SpaceInfo.packet_path`
- `SpaceInfo.packet_size`
- `SpaceInfo.packet_support`
- `PacketPostStatus.packet_total_size`
- `PacketPostStatus.packet_read_size`

Should also use this opportunity to change the name of any additional struct/var/function names that refer to Packet when it should be pcap.

The api changes will break the brim api client, changes will need to be made there when updating to the latest version of zqd.

mattnibs:
Based on the decision to refer to packet captures as pcaps, replace
the parts of the zqd api that incorrectly refer to packets with pcaps.

closes #562

### DIFF
--- a/pcap/search.go
+++ b/pcap/search.go
@@ -15,8 +15,8 @@ import (
 )
 
 var (
-	// ErrNoPacketsFound is an error indicating no packets have been found.
-	ErrNoPacketsFound = zqe.E(zqe.NotFound, "no packets found")
+	// ErrNoPcapsFound is an error indicating no packets have been found.
+	ErrNoPcapsFound = zqe.E(zqe.NotFound, "no packets found")
 )
 
 type PacketFilter func(gopacket.Packet) bool
@@ -163,7 +163,7 @@ func (s *Search) Reader(ctx context.Context, r pcapio.Reader) (*SearchReader, er
 		return nil, err
 	}
 	if len(reader.window) == 0 {
-		return nil, ErrNoPacketsFound
+		return nil, ErrNoPcapsFound
 	}
 	return reader, nil
 }
--- a/zqd/space/space.go
+++ b/zqd/space/space.go
@@ -24,9 +24,10 @@ import (
 )
 
 const (
-	AllZngFile        = "all.zng"
-	configFile        = "config.json"
-	infoFile          = "info.json"
+	AllZngFile = "all.zng"
+	configFile = "config.json"
+	infoFile   = "info.json"
+	// XXX This should be named pcaps.idx.json. Once we have a system for migrations this should be done.
 	PcapIndexFile     = "packets.idx.json"
 	defaultStreamSize = 5000
 )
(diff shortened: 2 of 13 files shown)


mattnibs:
Should this get approved I will create a brim issue to update the client library over there.

zqd/space/space.go
```diff
@@ -420,8 +421,10 @@ func (s *Space) delete() error {
 }
 
 type config struct {
-	DataPath      string `json:"data_path"`
-	PacketPath    string `json:"packet_path"`
+	DataPath string `json:"data_path"`
+	// XXX PcapPath should be named pcap_path in json land. To avoid having to
+	//do a migration we'll keep this as-is for now.
```
mattnibs, line 426:
```suggestion
	// do a migration we'll keep this as-is for now.
```

zqd/space/space.go
```diff
@@ -24,9 +24,10 @@ import (
 )
 
 const (
-	AllZngFile        = "all.zng"
-	configFile        = "config.json"
-	infoFile          = "info.json"
+	AllZngFile = "all.zng"
+	configFile = "config.json"
+	infoFile   = "info.json"
+	// XXX This should be named pcaps.idx.json
```
mattnibs, line 30:
```suggestion
	// XXX This should be named pcaps.idx.json. Once we have a system for migrations this should be done.
```

pcap/search.go
```diff
@@ -15,8 +15,8 @@ import (
 )
 
 var (
-	// ErrNoPacketsFound is an error indicating no packets have been found.
-	ErrNoPacketsFound = zqe.E(zqe.NotFound, "no packets found")
+	// ErrNoPcapsFound is an error indicating no packets have been found.
+	ErrNoPcapsFound = zqe.E(zqe.NotFound, "no pcaps found")
```
henridf, line 19: "no pcaps found" sounds like we haven't found any packet captures, as opposed to having found packet captures but no matching packets inside them. Is this what we want?

mattnibs: So what I hear you proposing is the error message should remain: "no packets found"?

henridf: I don't know the the API. 

If the error is that there is no pcap, then no pcaps found. If the error is "no matching packets inside the pcaps", then yes "no packets found" (or "no matching packets found", if that makes sense).

mattnibs: Yeah I agree with the "no packets found" here.

mattnibs:
Going to merge this after the next release is cut.

mattnibs:
accompanying brim pr: https://github.com/brimsec/brim/pull/752

PR #674, opened 2020-04-28
add "zar zq"

mccanne:
This command adds support for the "zq" sub-command of zar, which
lets you run zq-style commands over the log files in a zar archive.
In this case, the file "_" is the log file in question and any
files referred to live within the zar directory associated with
the log.  The zq command is run for each log file found.

### DIFF
--- a/cmd/zar/zq/command.go
+++ b/cmd/zar/zq/command.go
@@ -30,8 +30,8 @@ import (
 
 var Zq = &charm.Spec{
 	Name:  "zq",
-	Usage: "zq [-R dir] [options]",
-	Short: "walk an archive and run zq logic",
+	Usage: "zq [-R dir] [options] [zql] file [file...]",
+	Short: "walk an archive and run zql queries",
 	Long: `
 "zar zq" descends the directory given by the -R option (or ZAR_ROOT env) looking for
 logs with zar directories and for each such directory found, it runs
@@ -77,11 +77,9 @@ func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
 	c := &Command{Command: parent.(*root.Command)}
 	f.StringVar(&c.root, "R", os.Getenv("ZAR_ROOT"), "root directory of zar archive to walk")
 	f.BoolVar(&c.quiet, "q", false, "don't display zql warnings")
-	// XXX f.StringVar(&c.dir, "d", "", "directory for output data files")
 	f.StringVar(&c.outputFile, "o", "", "write data to output file")
 	f.StringVar(&c.jsonTypePath, "j", "", "path to json types file")
 	f.StringVar(&c.jsonPathRegexp, "pathregexp", c.jsonPathRegexp, "regexp for extracting _path from json log name (when -inferpath=true)")
-	// XXX f.BoolVar(&c.quiet, "q", false, "don't display zql warnings")
 	f.BoolVar(&c.stopErr, "e", true, "stop upon input errors")
 
 	// Flags added for writers are -f, -T, -F, -E, -U, and -b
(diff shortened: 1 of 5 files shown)


mccanne:
I will add tests in the morning...

cmd/zar/zq/command.go
```diff
@@ -0,0 +1,238 @@
...
+			}
+			inputs = inputs[1:]
+		}
+		localPaths := archive.Localize(zardir, inputs)
+		readers, err := c.inputReaders(resolver.NewContext(), localPaths)
+		if err != nil {
+			return err
+		}
+		if len(readers) == 0 {
+			// XXX silently skip if no file found
```
alfred-landrum, line 161: Skipping & proceeding seems like the right thing to do; do you think we should log here?

cmd/zar/zq/command.go
```diff
@@ -0,0 +1,238 @@
...
+			}
+			err = fmt.Errorf("%s: %w", path, err)
+			if c.stopErr {
+				return nil, err
+			}
+			fmt.Fprintf(os.Stderr, "%s\n", err)
+			continue
+		}
+		// wrap in a named reader so the reader implements Stringer
+		readers = append(readers, namedReader{file, path})
```
alfred-landrum, line 213: I don't see the String() method used, is that for debugging?

cmd/zar/zq/command.go
```diff
@@ -0,0 +1,238 @@
...
+	"github.com/brimsec/zq/zng/resolver"
+	"github.com/brimsec/zq/zql"
+	"github.com/mccanne/charm"
+	"go.uber.org/zap"
+)
+
+var Zq = &charm.Spec{
+	Name:  "zq",
+	Usage: "zq [-R dir] [options]",
+	Short: "walk an archive and run zq logic",
```
henridf, line 34: what does "logic" mean here ... could this be more simply expressed as "walk an archive and run zq over it"? (similar comment for the `Long:` section below)

cmd/zar/zq/command.go
```diff
@@ -0,0 +1,238 @@
...
+		Format:         c.ReaderFlags.Format,
+		DashStdin:      true,
+		JSONTypeConfig: c.jsonTypeConfig,
+		JSONPathRegex:  c.jsonPathRegexp,
+	}
+	var readers []zbuf.Reader
+	for _, path := range paths {
+		file, err := detector.OpenFile(zctx, path, cfg)
+		if err != nil {
+			// XXX for zar zq siliently conotinue if file doesn't exist
```
henridf, line 201: siliently conotinue

cmd/zar/zq/command.go
```diff
@@ -0,0 +1,238 @@
...
+	// Flags added for writers are -f, -T, -F, -E, -U, and -b
+	c.WriterFlags.SetFlags(f)
+
+	// Flags added for readers are -i XXX json
+	c.ReaderFlags.SetFlags(f)
+
+	return c, nil
+}
+
+func (c *Command) loadJsonTypes() (*ndjsonio.TypeConfig, error) {
```
henridf, line 96: can we get rid of all this json stuff here? or do that as part of the "xxx refactor into a tools package" comment?

cmd/zar/zq/command.go
mccanne: changing to warn to stderr in no -q

cmd/zar/zq/command.go
mccanne: scanner.Combiner.Read() calls *printf with the reader interface pointer as a %s.   I think Noah came up with this.

mccanne: I will clarify the comment

cmd/zar/zq/command.go
mccanne: I'll tweak it.  It doesn't "run zq".  It runs zql queries similar to how the standalone zql command works.

cmd/zar/zq/command.go
mccanne: will put a warnring for no -q

cmd/zar/zq/command.go
mccanne: I would rather leave it as a placeholder so we remember to adjust this call site.  The log files here referred to with _ can be json in need of schema maps.

PR #675, opened 2020-04-28
Send scanner stats back in search

### LINKED ISSUE #555: zqd: send scanner stats when running a flowgraph

henridf:
This commit tracks scanner stats in scanner.Scanner, and wires things up so that they get sent back in searches (via the existing driver.Driver interface).

closes #555

### DIFF
--- a/driver/compile.go
+++ b/driver/compile.go
@@ -25,7 +25,7 @@ func Compile(ctx context.Context, program ast.Proc, reader zbuf.Reader, reverse 
 func CompileWarningsCh(ctx context.Context, program ast.Proc, reader zbuf.Reader, reverse bool, span nano.Span, logger *zap.Logger, ch chan string) (*MuxOutput, error) {
 
 	filterAst, program := liftFilter(program)
-	input, err := inputProc(reader, filterAst, span)
+	scanner, err := inputProc(reader, filterAst, span)
 	if err != nil {
 		return nil, err
 	}
@@ -36,11 +36,11 @@ func CompileWarningsCh(ctx context.Context, program ast.Proc, reader zbuf.Reader
 		Reverse:     reverse,
 		Warnings:    ch,
 	}
-	leaves, err := proc.CompileProc(nil, program, pctx, input)
+	leaves, err := proc.CompileProc(nil, program, pctx, scanner)
 	if err != nil {
 		return nil, err
 	}
-	return NewMuxOutput(pctx, leaves), nil
+	return NewMuxOutput(pctx, leaves, scanner), nil
 }
 
 // liftFilter removes the filter at the head of the flowgraph AST, if
@@ -70,7 +70,7 @@ func liftFilter(p ast.Proc) (*ast.FilterProc, ast.Proc) {
 // inputProc takes a Reader, optional Filter AST, and timespan, and
 // constructs an input proc that can be used as the head of a
 // flowgraph.
-func inputProc(reader zbuf.Reader, fltast *ast.FilterProc, span nano.Span) (proc.Proc, error) {
+func inputProc(reader zbuf.Reader, fltast *ast.FilterProc, span nano.Span) (*scanner.Scanner, error) {
 	var f filter.Filter
 	if fltast != nil {
 		var err error
--- a/driver/compile.go
+++ b/driver/compile.go
@@ -25,7 +25,7 @@ func Compile(ctx context.Context, program ast.Proc, reader zbuf.Reader, reverse 
 func CompileWarningsCh(ctx context.Context, program ast.Proc, reader zbuf.Reader, reverse bool, span nano.Span, logger *zap.Logger, ch chan string) (*MuxOutput, error) {
 
 	filterAst, program := liftFilter(program)
-	scanner, err := inputProc(reader, filterAst, span)
+	scanner, err := newScanner(reader, filterAst, span)
 	if err != nil {
 		return nil, err
 	}
@@ -67,10 +67,10 @@ func liftFilter(p ast.Proc) (*ast.FilterProc, ast.Proc) {
 	return nil, p
 }
 
-// inputProc takes a Reader, optional Filter AST, and timespan, and
-// constructs an input proc that can be used as the head of a
+// newScanner takes a Reader, optional Filter AST, and timespan, and
+// constructs a scanner that can be used as the head of a
 // flowgraph.
-func inputProc(reader zbuf.Reader, fltast *ast.FilterProc, span nano.Span) (*scanner.Scanner, error) {
+func newScanner(reader zbuf.Reader, fltast *ast.FilterProc, span nano.Span) (*scanner.Scanner, error) {
 	var f filter.Filter
 	if fltast != nil {
 		var err error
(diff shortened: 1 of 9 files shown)


driver/mux.go
```diff
@@ -63,16 +66,23 @@ func (m *Mux) run() {
 	}
 }
 
-func NewMuxOutput(ctx *proc.Context, parents []proc.Proc) *MuxOutput {
+func NewMuxOutput(ctx *proc.Context, parents []proc.Proc, scanner *scanner.Scanner) *MuxOutput {
```
henridf, line 69: Rather than add a `*scanner.Scanner` to MuxOutput, we could just say it takes a `interface{Stats() api.ScannerStats}`, if this is preferred. And/or rename MuxOutput to something like Flowgraph or Cradle to enshrine it as the thing that holds a running flowgraph.

scanner/scanner.go
```diff
@@ -1,18 +1,28 @@
...
+	"github.com/brimsec/zq/zqd/api"
 )
 
 // Scanner implements the proc.Proc interface.
 type Scanner struct {
 	reader zbuf.Reader
 	filter filter.Filter
 	span   nano.Span
+	stats  struct {
+		currentTs      int64
```
mattnibs, line 20: Doesn't look like this field is being used.

zqd/handlers_test.go
```diff
@@ -576,7 +605,10 @@ func TestDeleteDuringPacketPost(t *testing.T) {
 	require.Error(t, <-packetPostErr, "context canceled")
 }
 
-func zngSearch(t *testing.T, client *api.Connection, space, prog string) string {
+// zngSearch runs the provided zql program as a search on the provided
+// space, returning a slice of control messages along with a strong
+// containing the tzng results.
+func zngSearch(t *testing.T, client *api.Connection, space, prog string) ([]interface{}, string) {
```
mattnibs, line 611: Rather than add this extra return argument that non of the other tests use, I would suggest creating a different func for retrieving stats *or* just calling the api Connection directly in the one test.

scanner/scanner.go
henridf: good catch

zqd/handlers_test.go
henridf: I think it's reasonable that a test helper running a search would return the control messages alongside the search results, even if most tests use just one or the other. Also, I think we have an indexing test in our near future that need to look at both. That's the context, but if you prefer I can make this suggested change.

zqd/handlers_test.go
```diff
@@ -576,7 +605,10 @@ func TestDeleteDuringPacketPost(t *testing.T) {
 	require.Error(t, <-packetPostErr, "context canceled")
 }
 
-func zngSearch(t *testing.T, client *api.Connection, space, prog string) string {
+// zngSearch runs the provided zql program as a search on the provided
+// space, returning a slice of control messages along with a strong
```
alfred-landrum, line 609:
```suggestion
// space, returning a slice of control messages along with a string
```

zqd/handlers_test.go
alfred-landrum: fwiw, I'm also ok with zngSearch returning both the control & results in one call.

mattnibs: I would just prefer that you use `api.Connection` directly. No longer can I do things like:
```
require.Equal(t, zngSearch(t, "test", "* | count"), expected)
```

mattnibs: Antidotally in the invalid the SearchRequest test will call the Connection directly. I think this is the right move, but why shouldn't we also now return `error` so we can use the same func?

mattnibs: I would prefer that you just make a new function `search` that is the new `zngSearch` and make `zngSearch` do:
```
func zngSearch(args) string {
   str, _ := search(args)
   return str
}
```

driver/compile.go
```diff
@@ -70,7 +70,7 @@ func liftFilter(p ast.Proc) (*ast.FilterProc, ast.Proc) {
 // inputProc takes a Reader, optional Filter AST, and timespan, and
 // constructs an input proc that can be used as the head of a
 // flowgraph.
-func inputProc(reader zbuf.Reader, fltast *ast.FilterProc, span nano.Span) (proc.Proc, error) {
+func inputProc(reader zbuf.Reader, fltast *ast.FilterProc, span nano.Span) (*scanner.Scanner, error) {
```
nwt, line 73: Nit: Maybe rename to newScanner?

zqd/handlers_test.go
```diff
@@ -42,10 +42,39 @@ func TestSearch(t *testing.T) {
 	sp, err := client.SpacePost(context.Background(), api.SpacePostRequest{Name: "test"})
 	require.NoError(t, err)
 	_ = postSpaceLogs(t, client, sp.Name, nil, false, src)
-	res := zngSearch(t, client, sp.Name, "*")
+	_, res := zngSearch(t, client, sp.Name, "*")
 	require.Equal(t, test.Trim(src), res)
 }
 
+func TestSearchStats(t *testing.T) {
+	src := `
```
nwt, line 50: Nit: const!

zqd/handlers_test.go
nwt: I don't have strong feelings one way or the other, but I do think a function returning both a tzng string and a []interface{} of control messages should return them as `string, []interface{}` since the string is almost always more interesting.

nwt: Having thought about it a little more, I'm now of the opinion that two functions (maybe search, returning two values, and searchTzng, returning one) would be nice.

henridf: Thanks for all the thoughts... updated to the suggestion in the most recent comment.

> Antidotally in the invalid the SearchRequest test will call the Connection directly. I think this is the right move, but why shouldn't we also now return error so we can use the same func?

@mattnibs I couldn't parse the above question... not sure if it still needs to be answered.

driver/compile.go
henridf: 👍

zqd/handlers_test.go
henridf: We have `src := <string>`  throughout this file.

zqd/handlers_test.go
```diff
@@ -591,8 +623,17 @@ func zngSearch(t *testing.T, client *api.Connection, space, prog string) string
...
+	var msgs []interface{}
+	r.SetOnCtrl(func(i interface{}) {
+		msgs = append(msgs, i)
+	})
 	require.NoError(t, zbuf.Copy(w, r))
-	return buf.String()
+	return buf.String(), msgs
+}
+
+func searchTzng(t *testing.T, client *api.Connection, space, prog string) string {
```
mattnibs, line 634: 💥  gracias

PR #676, opened 2020-04-28
Prevent creating record types with duplicate field names

### LINKED ISSUE #652: Don't allow duplicate field names in reducer/groupby
This simple query produces invalid output (two identically named columns)
```
(~/src/zq)$ go run ./cmd/zq 'count(uid), count(id.orig_h)' conn.log 
#0:record[count:uint64,count:uint64]
0:[92;92;]
```

A similar issue was fixed for cut in #165, but the logic there is inside ColumnBuilder which isn't used for reducer.  groupby uses ColumnBuilder but only to build the keys portion of output records.  We should fix this for the reducer and groupby procs...

aswan:
The substantive change in LookupTypeRecord() is relatively small,
most of the change here involves adding error handling to previously
infallible code paths.

Closes #652

### DIFF
--- a/tests/suite/reducer/duplicate.yaml
+++ b/tests/suite/reducer/duplicate.yaml
@@ -1,0 +1,8 @@
+zql: count(), count()
+
+input: |
+  #0:record[x:int32]
+  0:[1;]
+  0:[2;]
+
+errorRE: duplicate column count
--- a/tests/suite/reducer/duplicate.yaml
+++ b/tests/suite/reducer/duplicate.yaml
@@ -5,4 +5,4 @@ input: |
   0:[1;]
   0:[2;]
 
-errorRE: duplicate column count
+errorRE: duplicate field count
--- a/zdx/mem.go
+++ b/zdx/mem.go
@@ -77,7 +77,8 @@ func (t *MemTable) open() {
 	if t.valType != nil {
 		cols = append(cols, zng.Column{"value", t.valType})
 	}
-	t.builder = zng.NewBuilder(t.zctx.LookupTypeRecord(cols))
+	typ, _ := t.zctx.LookupTypeRecord(cols)
+	t.builder = zng.NewBuilder(typ)
 }
 
 func checkType(which string, before *zng.Type, after zng.Type) error {
--- a/zdx/mem.go
+++ b/zdx/mem.go
@@ -77,7 +77,7 @@ func (t *MemTable) open() {
 	if t.valType != nil {
 		cols = append(cols, zng.Column{"value", t.valType})
 	}
-	typ, _ := t.zctx.LookupTypeRecord(cols)
+	typ := t.zctx.MustLookupTypeRecord(cols)
 	t.builder = zng.NewBuilder(typ)
 }
 
--- a/zng/resolver/context.go
+++ b/zng/resolver/context.go
@@ -226,6 +226,14 @@ func (c *Context) LookupTypeRecord(columns []zng.Column) (*zng.TypeRecord, error
 	return typ, nil
 }
 
+func (c *Context) MustLookupTypeRecord(columns []zng.Column) *zng.TypeRecord {
+	r, err := c.LookupTypeRecord(columns)
+	if err != nil {
+		panic(err)
+	}
+	return r
+}
+
 func (c *Context) LookupTypeSet(inner zng.Type) *zng.TypeSet {
 	key := setKey(inner)
 	c.mu.Lock()
--- a/zng/resolver/context.go
+++ b/zng/resolver/context.go
@@ -209,7 +209,7 @@ func (c *Context) LookupTypeRecord(columns []zng.Column) (*zng.TypeRecord, error
 	for _, col := range columns {
 		_, exists := names[col.Name]
 		if exists {
-			return nil, fmt.Errorf("duplicate column %s", col.Name)
+			return nil, fmt.Errorf("duplicate field %s", col.Name)
 		}
 		names[col.Name] = val
 	}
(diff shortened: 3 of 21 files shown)


zdx/mem.go
```diff
@@ -77,7 +77,8 @@ func (t *MemTable) open() {
 	if t.valType != nil {
 		cols = append(cols, zng.Column{"value", t.valType})
 	}
-	t.builder = zng.NewBuilder(t.zctx.LookupTypeRecord(cols))
+	typ, _ := t.zctx.LookupTypeRecord(cols)
```
henridf, line 80: I've seen functions that have a "Must" variant for cases that are (allegedly) certain not to fail, and that panic instead of returning an error. (For example https://godoc.org/regexp#Compile vs https://godoc.org/regexp#MustCompile). 

Maybe we should use this here? Curious to hear what others think too.

tests/suite/reducer/duplicate.yaml
```diff
@@ -0,0 +1,8 @@
+zql: count(), count()
```
henridf, line 1: We could also detect this statically and make it a compile-time error (or alternatively, don't make it an error, but automatically rename the second field to something like `count_1`, though I'm not sure what the use case would be...).

aswan: Given that we can (in theory) manually write `count2 = count()` etc, I have a slight preference for making this an error and having users manually rewrite names if that's what they want.
Re detecting statically, that does seem more user-friendly, but I think that the changes here are a good safety net -- they'll also catch file inputs that repeat column names, other procs that we haven't thought through to add static checking to, etc.  I propose merging this and adding static checking in reducer/groupby in a separate patch?

henridf: That sounds good - I agree that we need this change anyway, since it's a bug that LookupTypeRecord doesn't check for duplicate names.

zdx/mem.go
aswan: I followed this suggestion and I like it.
@nwt any opinion on this?

zng/resolver/context.go
```diff
@@ -448,10 +470,14 @@ func (c *Context) parseRecordTypeBody(in string) (string, zng.Type, error) {
 			continue
 		}
 		rest, ok = match(rest, "]")
-		if ok {
-			return rest, c.LookupTypeRecord(columns), nil
+		if !ok {
+			return "", nil, zng.ErrTypeSyntax
 		}
-		return "", nil, zng.ErrTypeSyntax
+		typ, err := c.LookupTypeRecord(columns)
```
henridf, line 476: Just a few lines above (can't add a comment there), I think you can remove this?

```go
for _, c := range columns {
	if col.Name == c.Name {
              return "", nil, zng.ErrDuplicateFields
        }
}
```

henridf: and probably you can also get rid of zng.ErrDuplicateFields

zng/resolver/context.go
```diff
@@ -202,17 +202,36 @@ func (c *Context) AddType(t zng.Type) zng.Type {
...
 // use TranslateTypeRecord.
-func (c *Context) LookupTypeRecord(columns []zng.Column) *zng.TypeRecord {
+func (c *Context) LookupTypeRecord(columns []zng.Column) (*zng.TypeRecord, error) {
+	// First check for duplicate columns
+	names := make(map[string]struct{})
+	var val struct{}
+	for _, col := range columns {
+		_, exists := names[col.Name]
+		if exists {
+			return nil, fmt.Errorf("duplicate column %s", col.Name)
```
henridf, line 212: if the terminology is column, could you also update the duplicate fields error message in proc/colbuilder.go to refer to column rather than fields?

aswan: Oops, i think i like field better, i changed it here.  I don't feel strongly though, we can switch to column everywhere if people prefer that.

zdx/mem.go
nwt: I think it's an excellent idea!

zng/resolver/context.go
```diff
@@ -577,7 +598,12 @@ func (c *Context) TranslateTypeRecord(ext *zng.TypeRecord) *zng.TypeRecord {
 		child := c.TranslateType(col.Type)
 		columns = append(columns, zng.NewColumn(col.Name, child))
 	}
-	return c.LookupTypeRecord(columns)
+
+	// LookupTypeRecord() fails if there are duplicate columns but
+	// since this comes from an already validated record, it should
+	// not fail
+	typ, _ := c.LookupTypeRecord(columns)
+	return typ
```
nwt, line 606:
```suggestion
	return c.MustLookupTypeRecord(columns)
```

zng/resolver/context.go
```diff
@@ -202,17 +202,36 @@ func (c *Context) AddType(t zng.Type) zng.Type {
...
 // this type context.  If you want to use columns from a different type context,
 // use TranslateTypeRecord.
-func (c *Context) LookupTypeRecord(columns []zng.Column) *zng.TypeRecord {
+func (c *Context) LookupTypeRecord(columns []zng.Column) (*zng.TypeRecord, error) {
+	// First check for duplicate columns
+	names := make(map[string]struct{})
+	var val struct{}
+	for _, col := range columns {
+		_, exists := names[col.Name]
+		if exists {
```
nwt, line 211:
```suggestion
		if _, ok := names[col.Name]; ok  {
```
ok is the conventional name for this boolean.

aswan: Hm, I chose that because I found ok misleading -- things are not okay if that name already exists.  But shrug, I'll change it.

PR #685, opened 2020-04-28
zqd search endpoint: only allow backwards searches

### LINKED ISSUE #503: Handle search Dir parameters properly
The zqd search api takes a "Dir" parameter which is meant to specify whether results should be from oldest-to-newest or from newest-to-oldest.  But the only value that works right now is -1 to specify newest-to-oldest.  Passing 1 (to request oldest-to-newest) just silently breaks some queries (such as those that include time-binned groupby procs).

At the very least we should validate Dir and return an error if a value other than -1 is passed.  We could also just remove this parameter, it would not be difficult to add it back if/when we implement the ability to generate results in either order.  (see also #501)

mattnibs:
Also, use zqe when returning errors for invalid dir values.

closes #503

### DIFF
--- a/zqd/handlers_test.go
+++ b/zqd/handlers_test.go
@@ -78,7 +78,7 @@ func TestSearchEmptySpace(t *testing.T) {
 	require.Equal(t, "", res)
 }
 
-func TestSearchInvalidRequest(t *testing.T) {
+func TestSearchError(t *testing.T) {
 	src := `
 #0:record[_path:string,ts:time,uid:bstring]
 0:[conn;1521911723.205187;CBrzd94qfowOqJwCHa;]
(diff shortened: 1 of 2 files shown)


zqd/search/search.go
```diff
@@ -37,9 +38,13 @@ func NewSearch(ctx context.Context, s *space.Space, req api.SearchRequest) (*Sea
 	if req.Span.Dur < 0 {
 		return nil, errors.New("time span must have non-negative duration")
 	}
-	// XXX allow either direction even through we do forward only right now
-	if req.Dir != 1 && req.Dir != -1 {
-		return nil, errors.New("time direction must be 1 or -1")
+	// XXX zqd only supports backwards searches, remove once this has been
+	// fixed.
+	if req.Dir == 1 {
+		return nil, zqe.E("forward facing searches not supported", zqe.Invalid)
```
alfred-landrum, line 44: When giving arguments to `zqe.E`, a string (and any formatting verbs) need to be last, so that it can mimic fmt.Errorf's variadiac arguments.

```suggestion
		return nil, zqe.E(zqe.Invalid, "forward facing searches not supported")
```

zqd/search/search.go
```diff
@@ -37,9 +38,13 @@ func NewSearch(ctx context.Context, s *space.Space, req api.SearchRequest) (*Sea
...
+	}
+	if req.Dir != -1 {
+		return nil, zqe.E("time direction must be 1 or -1", zqe.Invalid)
```
alfred-landrum, line 47:
```suggestion
		return nil, zqe.E(zqe.Invalid, "time direction must be 1 or -1")
```

zqd/handlers_test.go
```diff
@@ -78,7 +78,7 @@ func TestSearchEmptySpace(t *testing.T) {
 	require.Equal(t, "", res)
 }
 
-func TestSearchInvalidRequest(t *testing.T) {
+func TestSearchError(t *testing.T) {
```
alfred-landrum, line 81: Thanks for using the zqe.E to create the errors! You should be to add assertions for 400 (bad request) status codes since you specified `zqe.Invalid` in the zqe.E calls.

mattnibs: Will do

zqd/search/search.go
```diff
@@ -37,9 +38,13 @@ func NewSearch(ctx context.Context, s *space.Space, req api.SearchRequest) (*Sea
 	if req.Span.Dur < 0 {
 		return nil, errors.New("time span must have non-negative duration")
 	}
-	// XXX allow either direction even through we do forward only right now
-	if req.Dir != 1 && req.Dir != -1 {
-		return nil, errors.New("time direction must be 1 or -1")
+	// XXX zqd only supports backwards searches, remove once this has been
+	// fixed.
+	if req.Dir == 1 {
+		return nil, zqe.E(zqe.Invalid, "forward facing searches not supported")
```
henridf, line 44:
```suggestion
		return nil, zqe.E(zqe.Invalid, "forward searches not yet supported")
```

PR #690, opened 2020-04-28
Log 5xx level errors to core logger

### LINKED ISSUE #640: log 500 class errors to zqd-core
If we return an internal server error (status code 500), we should log the error message to zqd-core (not just to the zqd-access log), so that we can more easily notice & review the error.

mattnibs:
closes #640

### DIFF
--- a/zqd/handlers.go
+++ b/zqd/handlers.go
@@ -55,6 +55,9 @@ func respond(c *Core, w http.ResponseWriter, r *http.Request, status int, body i
 
 func respondError(c *Core, w http.ResponseWriter, r *http.Request, err error) {
 	status, ae := errorResponse(err)
+	if status >= 500 {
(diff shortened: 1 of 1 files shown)


zqd/handlers.go
```diff
@@ -55,6 +55,9 @@ func respond(c *Core, w http.ResponseWriter, r *http.Request, status int, body i
 
 func respondError(c *Core, w http.ResponseWriter, r *http.Request, err error) {
 	status, ae := errorResponse(err)
+	if status >= 500 {
+		c.requestLogger(r).Warn("Error", zap.Int("status", status), zap.Error(err))
```
alfred-landrum, line 59: nit: suggest we keep error messages lower case:

```suggestion
		c.requestLogger(r).Warn("error", zap.Int("status", status), zap.Error(err))
```

PR #691, opened 2020-04-28
add "zar zdx" command

mccanne:
This commit adds a zdx subcommand to zar, which lets you walk an
archive and run zdx on a given named zng file, which must be
a zng file with sorted keys.

### DIFF
--- a/cmd/zar/zdx/command.go
+++ b/cmd/zar/zdx/command.go
@@ -1,0 +1,120 @@
+package cmdzdx
+
+import (
+	"errors"
+	"flag"
+	"fmt"
+	"os"
+
+	"github.com/brimsec/zq/archive"
+	"github.com/brimsec/zq/cmd/zar/root"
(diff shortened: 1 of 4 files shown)


cmd/zar/zdx/command.go
```diff
@@ -0,0 +1,120 @@
...
+		writer, err := zdx.NewWriter(outputPath[0], c.framesize)
+		if err != nil {
+			return err
+		}
+		close := true
+		defer func() {
+			if close {
+				writer.Close()
+			}
+		}()
```
alfred-landrum, line 103:
```suggestion
		defer writer.Close()
```

mccanne: That's how I had it originally elsewherer but Noah wanted the return error from writer.Close() checked.

mccanne: hmm somewhere the error check went away

PR #693, opened 2020-04-28
Make zar ls better

mccanne:


cmd/zar/ls/command.go
```diff
@@ -29,25 +32,74 @@ func init() {
...
+		pattern = args[0]
+	}
+	return archive.Walk(c.root, func(zardir string) error {
+		printDir(zardir, pattern, c.lflag)
+		return nil
+	})
+}
+
+func fileExists(path string) bool {
+	if path == "-" {
```
alfred-landrum, line 64: I don't think you need to check for "-" in this particular case of fileExists. :)

tests/suite/zar/ls.yaml
```diff
@@ -0,0 +1,37 @@
...
+  mkdir logs
+  zar chop -q -b 2500 -R ./logs babble.tzng
+  echo ===
+  zar ls -R ./logs
+  echo ===
+  zar mkdirs ./logs
+  zar ls -R ./logs
+  echo ===
+  touch logs/20200421/1587508830.06852324.zng.zar/foo
+  zar ls -R ./logs foo
```
alfred-landrum, line 11: I don't see a usage of the `-l` long form flag.

PR #695, opened 2020-04-29
fix subnet searches

mccanne:
subnet searches stopped working at some point.
this is a workaround for now.


aswan:
What's the bug exactly?  On master, a subnet expression appears to parse correctly but the search doesn't match anything.  The fix for the latter would be in compileSearch, not in the grammar.

mccanne:
I think @aswan is right here.  I'm a little confused but will close this for now.

PR #697, opened 2020-04-29
Allow multiple fields to be written from put

aswan:


henridf:
Should the result of one clause be reflected in the subsequent clause? (without having thought about it much, my immediate inclination would be "yes")

zql/docs/processors/README.md
```diff
@@ -152,7 +152,7 @@ conn  1521911720.607695 CpjMvj2Cvj048u6bF1 10.164.94.120 39169     10.47.3.200 8
 |                           |                                                 |
 | ------------------------- | ----------------------------------------------- |
 | **Description**           | Add/update fields based on the results of a computed expression |
-| **Syntax**                | `put <field> = <expression>`                    |
+| **Syntax**                | `put <field> = <expression> [, <field> = <expression> ...]`     |
```
henridf, line 155: explain the evaluation model, in particular  that clauses do not observe the results of other clauses?

henridf:
Also, it looks like if one of the assignment expressions fails, all fail. This seems a little inconsistent with the above, but fine with me, however should probably also be mentioned in the doc?

```sh
 echo "  #0:record[a:int32,b:int32]
  0:[1;2;]
  0:[3;4;]" | zq -t "put a=b,b=a,c=d" -
field is not present
#0:record[a:int32,b:int32]
0:[1;2;]
0:[3;4;]
```

PR #708, opened 2020-05-01
support progress updates during log ingest

alfred-landrum:
Send the log bytes read & total in LogPostStatus messages.

This is the zqd side of https://github.com/brimsec/brim/issues/567 .

### DIFF
--- a/zqd/api/api.go
+++ b/zqd/api/api.go
@@ -119,11 +119,9 @@ type LogPostWarning struct {
 }
 
 type LogPostStatus struct {
-	Type         string     `json:"type"`
-	Span         *nano.Span `json:"span,omitempty"`
-	Size         int64      `json:"size" unit:"bytes"`
-	LogTotalSize int64      `json:"log_total_size" unit:"bytes"`
-	LogReadSize  int64      `json:"log_read_size" unit:"bytes"`
+	Type         string `json:"type"`
+	LogTotalSize int64  `json:"log_total_size" unit:"bytes"`
+	LogReadSize  int64  `json:"log_read_size" unit:"bytes"`
 }
 
 // PacketSearch are the query string args to the packet endpoint when searching
--- a/zqd/api/api.go
+++ b/zqd/api/api.go
@@ -119,9 +119,11 @@ type LogPostWarning struct {
 }
 
 type LogPostStatus struct {
-	Type string     `json:"type"`
-	Span *nano.Span `json:"span,omitempty"`
-	Size int64      `json:"size" unit:"bytes"`
+	Type         string     `json:"type"`
+	Span         *nano.Span `json:"span,omitempty"`
+	Size         int64      `json:"size" unit:"bytes"`
+	LogTotalSize int64      `json:"log_total_size" unit:"bytes"`
+	LogReadSize  int64      `json:"log_read_size" unit:"bytes"`
 }
 
 // PacketSearch are the query string args to the packet endpoint when searching
--- a/zqd/handlers_test.go
+++ b/zqd/handlers_test.go
@@ -372,9 +372,8 @@ func TestPostZngLogs(t *testing.T) {
 	span := &nano.Span{Ts: 1e9, Dur: 1e9 + 1}
 	require.Equal(t, &api.LogPostStatus{
 		Type:         "LogPostStatus",
-		Span:         span,
-		Size:         80,
 		LogTotalSize: 148,
+		LogReadSize:  58,
 	}, status)
 
 	taskend := payloads[len(payloads)-1].(*api.TaskEnd)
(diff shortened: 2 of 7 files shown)


zqd/ingest/log.go
```diff
@@ -127,22 +132,26 @@ func ingestLogs(ctx context.Context, pipe *api.JSONPipe, s *space.Space, req api
 		min := nano.Min(tailW.r.Ts, headW.r.Ts)
 		max := nano.Max(tailW.r.Ts, headW.r.Ts)
 		if err = s.SetSpan(nano.NewSpanTs(min, max+1)); err != nil {
+			os.Remove(zngfile.Name())
 			return err
 		}
 	}
-	if err := os.Rename(zngfile.Name(), s.DataPath(space.AllZngFile)); err != nil {
+	if err = os.Rename(zngfile.Name(), s.DataPath(space.AllZngFile)); err != nil {
```
henridf, line 139: nit

```suggestion
	if err := os.Rename(zngfile.Name(), s.DataPath(space.AllZngFile)); err != nil {
```

zqd/api/api.go
```diff
@@ -119,9 +119,11 @@ type LogPostWarning struct {
 }
 
 type LogPostStatus struct {
-	Type string     `json:"type"`
-	Span *nano.Span `json:"span,omitempty"`
-	Size int64      `json:"size" unit:"bytes"`
+	Type         string     `json:"type"`
+	Span         *nano.Span `json:"span,omitempty"`
+	Size         int64      `json:"size" unit:"bytes"`
```
mattnibs, line 124: This field is kind of confusing next to `Log*Size`. The brim client doesn't use it and it can easily find this by calling `GET /space/:space`. I think we should drop it.

zqd/handlers_test.go
```diff
@@ -371,9 +371,10 @@ func TestPostZngLogs(t *testing.T) {
 	status := payloads[len(payloads)-2].(*api.LogPostStatus)
 	span := &nano.Span{Ts: 1e9, Dur: 1e9 + 1}
 	require.Equal(t, &api.LogPostStatus{
-		Type: "LogPostStatus",
-		Span: span,
-		Size: 80,
+		Type:         "LogPostStatus",
+		Span:         span,
+		Size:         80,
+		LogTotalSize: 148,
```
mattnibs, line 377: Should we also have `LogReadSize: 148` here as well?

zqd/api/api.go
mattnibs: And the `Size` field doesn't appear to be updated until the very end.

zqd/handlers_test.go
alfred-landrum: I looked at doing this, and thought it would be complex, since the value for LogReadSize comes from the driver.Ouput stat call, but I just realized how to do it.

zqd/ingest/driver.go
```diff
@@ -32,10 +34,42 @@ func (d *logdriver) Warn(warning string) error {
...
+		}
+	}
+	batch.Unref()
+	return nil
+}
+
+func (s *simpledriver) Warn(warning string) error {
+	return nil
+}
+func (s *simpledriver) Stats(stats api.ScannerStats) error {
```
nwt, line 69:
```suggestion
}

func (s *simpledriver) Stats(stats api.ScannerStats) error {
```

zqd/ingest/log.go
```diff
@@ -127,22 +132,25 @@ func ingestLogs(ctx context.Context, pipe *api.JSONPipe, s *space.Space, req api
 		min := nano.Min(tailW.r.Ts, headW.r.Ts)
 		max := nano.Max(tailW.r.Ts, headW.r.Ts)
 		if err = s.SetSpan(nano.NewSpanTs(min, max+1)); err != nil {
+			os.Remove(zngfile.Name())
```
nwt, line 135: Should we also add one of these a few lines above, like this?
```go
	if err := zngfile.Close(); err != nil {
		os.Remove(zngfile.Name())
		return err
	}

**Comment by alfred-landrum** [comment]
I made all the changes suggested (thanks!), but in the process, discovered a problem: As you can see from the adjusted tests in handlers_test.go, at the end of ingest, LogTotalSize != LogReadSize :| . That's because LogTotalSize is the total file size of the incoming logs, and LogReadSize is the total of len(record.Raw) of the records read from the logs. So for formats like ndjson, where the size diff between json and zng is large, the progress reported is going to look very slow.

Trying to plumb the actual read count from the zbuf.Readers seems complex, so open to any other suggestions.

File: zqd/ingest/log.go [hunk 6]
```diff
@@ -79,6 +79,7 @@ func ingestLogs(ctx context.Context, pipe *api.JSONPipe, s *space.Space, req api
 		cfg.JSONTypeConfig = req.JSONTypeConfig
 		cfg.JSONPathRegex = DefaultJSONPathRegexp
 	}
+	var totalSize int64
 	for _, path := range req.Paths {
 		sf, err := detector.OpenFile(zctx, path, cfg)
```
mattnibs, line 84: @alfred-landrum regarding your question about getting bytes read from the reader. I would suggest opening the file manually here and wrapping the file in some `io.Reader` struct that increments some counter based off the bytes pulled off from `Read([]byte)`.

mattnibs: Then just doing `detector.NewReader` instead of `detector.OpenFile`.

alfred-landrum: I've adjusted this to get the actual bytes read; it was more complex than calling NewReader, as the I wanted to leave the json configuration in the detect package. PTAL.

PR #709, opened 2020-05-01
Have scanner listen and stop on context.Cancel

### LINKED ISSUE #707: Can't quit Brim during Zeek log ingest
Repro is with Brim commit `f2b4948` talking to `zqd` commit `233e930`.

In the attached video I drag the 12 GB "year1" subset of wrccdc logs into the app, then try to quit. I get the "Are you sure...?" warning and click that I'm ok with that. However, unlike with pcap ingest where the app deletes the partial data and quits immediately, here it hangs for several minutes and never seems to quit (notice the timer in the upper-right hand corner of the video.)

[Repro.mp4.zip](https://github.com/brimsec/brim/files/4561663/Repro.mp4.zip)

mattnibs:
The lack of context listening to the cancel signal was causing a
a long running ingest handler to not exit when the context of the
request was cancelled. Add context.Err check to scanner which in
turn emits EOS and lets the the flowgraph elegantly cease operations.

Also there was a deadlock occurring in the split proc it recieves an
EOS and the split proc has receieved no data.
The was occurring because the child SplitChannel procs were themselves
listening for context.Done() and exiting before the SplitProc could
tell them itself and the parent SplitProc was then deadlocked trying to
push an EOS to children that were not listening. Tell the SplitChannel to calm
down and trust the process.

closes #707

### DIFF
--- a/proc/split.go
+++ b/proc/split.go
@@ -54,8 +54,11 @@ func (s *Split) run() {
 	// the upstream path and a reference-counted batch is transmitted to
 	// each requesting entity.
 	strip := make([]chan<- Result, 0, s.n)
-	for s.n > 0 {
+	for {
 		flight := s.gather(strip)
+		if s.n == 0 {
+			break
+		}
 		batch, err := s.Get()
 		s.send(flight, Result{batch, err})
 		if batch != nil {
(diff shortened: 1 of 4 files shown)


proc/split.go
```diff
@@ -110,19 +113,12 @@ func (s *SplitChannel) Pull() (zbuf.Batch, error) {
 	// which will cause it to exit when it sees that all of us SplitChannels
 	// are gone.  We don't want both SplitProc and SplitChannel listening
 	// on context canceled as that could lead to deadlock.
```
nwt, line 115: I think this comment needs an update.

mattnibs: Good call.

PR #711, opened 2020-05-01
create proc.Cutter from proc.Cut so it can be reused by zdx

mccanne:


henridf:
Are there any substantive changes to the proc.Cutter or is just just a move/rename? (hard to tell from the single commit)

mccanne:
> Are there any substantive changes to the proc.Cutter or is just just a move/rename? (hard to tell from the single commit)

No change at all.  Just a re-factor so cutter.Cut() can be called an a sequence of zng.Records without the rest of the proc plumbing.  I might have fixed a Go nit or two but no semantics change.

henridf:
> No change at all. Just a re-factor so cutter.Cut() can be called an a sequence of zng.Records without the rest of the proc plumbing. I might have fixed a Go nit or two but no semantics change.

That makes things easier!

mccanne:
> > No change at all. Just a re-factor so cutter.Cut() can be called an a sequence of zng.Records without the rest of the proc plumbing. I might have fixed a Go nit or two but no semantics change.
> 
> That makes things easier!

Thanks, I should I have pointed this out in the first place...

henridf:
> Thanks, I should I have pointed this out in the first place...

Or let the commit sequence do the talking and have a one commit that does any changes in place, and one commit that does the move. Like the two latter commits of #715 🤓 😎

Issue #679, opened 2020-04-28
fix zq -version when zq is built with no .git dir present


mikesbrown:
zq versioning needs to work when zq is npm installed into Brim. Right now the method of using `git describe --tags --dirty` only makes sense from a zq checkout. When zq is npm installed from an arbitrary hash, `git describe --tags --dirty` finds a .git in the parent directory, and the returned result isn't a sensible zq version at all.

Getting into scope creep here, but worth pointing out... this also means zq tarball or zip release assets can't be built. We might be able to fix both of these in one go.

This is need for brimsec/brim#634.

Issue #678, opened 2020-04-28
zq package.json should have build instructions when npm installing


mikesbrown:
When Brim attempts to install zq at an arbitrary commit, it calls `make build`. Brim shouldn't do this. For one, zq knows how it should be built. For two, this won't work on Windows (thanks to @alfred-landrum for pointing this out before I discovered what should be obvious).

Instead, the instructions for building zq/zqd should be contained in zq's package.json. When CI or a Brim developer wants to install zq/zqd from an arbitrary zq commit hash, those instructions will run.

This is needed to solve brimsec/brim#634

PR #712, opened 2020-05-01
Npm installable zq

### LINKED ISSUE #677: make zq npm-installable from arbitrary hashes
It's not possible to npm install zq from a commit hash, in the form like:
```
$ npm install "brimsec/zq#master"
npm ERR! Can't install github:brimsec/zq#0e6240b88cfd921bd73a9a5b1477df3a68d3ddf3: Missing package version
```
We need to fix this for  brimsec/brim#634

mikesbrown:
Here is one approach for being able to have Brim point to arbitrary zq commit hashes. When you `npm install github:brimsec/zq#hash`, this will clone zq.git, run scripts.prepack, and put dist/ and zq's javascript files into node_modules/zq. Everything from dist/ also goes into node_modules/.bin so that npx zq and similar usage is possible. Brim's scripts/download-zdeps will copy zq and zqd from node_modules/zq/dist into zdeps.

Companion Brim-side PR: https://github.com/brimsec/brim/pull/730

There are a lot of words in the ordered list below. This is one of those changes where a lot happens under the hood even though the diff line count is small, and I want to show that earlier realizations or conditions inform later design choices. Not having managed a Node project before, I might have missed some NPM capability. Please let me know if something is obviously wrong.

1. Local development changes or CI changes to Brim's package.json to modify zq SHOULD be performed via `npm install github:brimsec/zq#commitish`. Other methods of changing package.json are not as nearly attractive. This choice has consequences, as you'll see below.
1. We do not want to produce and track zq binary artifacts after all zq.git CI runs. This means anything that wants to npm install Brim MUST be able to build zq, so anything that wants to npm install Brim MUST also have Go. This means CI and Brim developers MUST have Go. Because release packaging includes zq/zqd, users of release artifacts either MAY or MAY NOT have Go.
1. To be able to `npm install github:brimsec/zq#commitish`, zq's package.json MUST contain a version.
1. A npm-installed zq -version MUST contain a commit hash of zq.
1. A commit hash of zq MUST be derived from a clone of zq, not an export. The .git directory must exist. Without this, 4 is violated.
1. npm install of a Github project first clones the repository and then copies stuff from that clone directory into node_modules, with optional intermediate steps along the way. When npm installing a Github project, it is not possible to carry over that project's clone's .git directory into node_modules. [npm-packlist](https://github.com/npm/npm-packlist/blob/master/index.js#L111) explicitly filters the .git directory, even if you try to unignore it in .npmignore. Note this violates 5 above.
1. This means zq's package.json MUST build zq from its pre-npm packed git clone *before* being installed into node_modules.
1. zq's version derivation mechanism is git describe --tags --dirty.
1. When a Github project in package.json has an arbitrary hash that is also pointed to by a ref in the source repository (e.g., Brim's package.json specifies zq as brimsec/zq#d4cba020b77b16957aba25ecd41da44ad5028c9b, and zq.git's refs/heads/master points to d4cba020b77b16957aba25ecd41da44ad5028c9b), [pacote](https://github.com/npm/pacote/blob/v9.5.11/lib/util/git.js#L147) performs a shallow clone of the project as part of the npm install. In those cases, git describe --tags --dirty will fail, because there are no tags fetched. This violates 8 above. Changing this to git describe --tags --dirty --always will produce a short hash if the clone is shallow.  In cases where pacote has to perform a full clone, git describe --tags --dirty --always will produce identical output to git describe --tags --dirty.
1. You might notice that the version in zq's package.json is largely arbitrary and never gets surfaced. It only satisfies brim-side `npm install zq` but serves no other purpose.
1. ~The instructions for building zq should be in zq.git, but one problem here is that Makefile is the source of truth, and make isn't available on Windows. TODO: zq's package.json's scripts.prepack should call something that can set VERSION and LDFLAGS that works on Windows as well as UNIX-based platforms. zq's Makefile should use the same thing. How to make that platform- and tool-independent is still TBD. Note that without VERSION, zq produced from npm install has a version of "unknown".~ While this is a valid concern, I think it's also "perfect as enemy of the good". I have implemented the build recipe in a parallel node script. More elaboration is below.
1. ~dist is kept around for now as a persistent, empty directory, because it makes the prepack script Just Work for demonstration purposes. When 10 above is working, this can probably go away.~ Obsolete after 2494187
1. Because zq's scripts.prepack is responsible for building zq (as it should be), when Brim wants to npm install zq#v1.2.3 (something that looks like a semver/release, not a ash), npm will clone zq#v1.2.3, build zq, put the built assets into node_modules/zq as usual. But then scripts/download-zdeps will download the zq v1.2.3 release assets from Github anyway. I'm not sure how to make npm install of zq in brim behave differently in this case. This is getting into some nastiness: trying to use npm to deal with a Go project. This also comes all the way back to the beginning, in that the choice to use npm install of a Go project is a little awkward. **Update after 7fa865c:** I have left this behavior as written. I think it's convenient to have zq's dist in node_modles/zq/dist. If folks don't want that confusion for tagged releases, that's something we should talk about.

Fixes #677 #678 #679

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 bin/
-dist/
+dist/*
+!dist/.gitignore
 tests/test-root
 zq-sample-data
 zql/deps/
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 bin/
-dist/*
-!dist/.gitignore
+dist/
 tests/test-root
 zq-sample-data
 zql/deps/
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 export GO111MODULE=on
 
-VERSION = $(shell git describe --tags --dirty)
+VERSION = $(shell git describe --tags --dirty --always)
 LDFLAGS = -s -X main.version=$(VERSION)
 ZEEKTAG = v3.0.2-brim1
 ZEEKPATH = zeek-$(ZEEKTAG)
--- a/Makefile
+++ b/Makefile
@@ -73,7 +73,7 @@ create-release-assets:
 test-ci: fmt tidy vet test-unit test-system test-zeek test-heavy
 
 clean:
-	@rm -rf dist
+	@rm -rf dist/*
 
 .PHONY: fmt tidy vet test-unit test-system test-heavy sampledata test-ci
 .PHONY: perf-compare build install create-release-assets clean
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
 export GO111MODULE=on
 
+# If VERSION or LDFLAGS change, please also change
+# npm/build.
 VERSION = $(shell git describe --tags --dirty --always)
 LDFLAGS = -s -X main.version=$(VERSION)
 ZEEKTAG = v3.0.2-brim1
@@ -49,6 +51,7 @@ test-zeek: bin/$(ZEEKPATH)
 perf-compare: build $(SAMPLEDATA)
 	scripts/comparison-test.sh
 
+# If the build recipe changes, please also change npm/build.
 build:
 	@mkdir -p dist
 	@go build -ldflags='$(LDFLAGS)' -o dist ./cmd/...
--- a/Makefile
+++ b/Makefile
@@ -73,7 +73,7 @@ create-release-assets:
 test-ci: fmt tidy vet test-unit test-system test-zeek test-heavy
 
 clean:
-	@rm -rf dist/*
+	@rm -rf dist
 
 .PHONY: fmt tidy vet test-unit test-system test-heavy sampledata test-ci
 .PHONY: perf-compare build install create-release-assets clean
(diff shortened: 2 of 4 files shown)


mikesbrown:
tl;dr: The big question I still need to muddle on, and would appreciate any opinions on, is how to make `scripts.prepack` run `make build` without `make` in Windows, while also making `make build` in the Makefile go through same code path, on both Windows and *nix.

mikesbrown:
I ended up implementing the build recipe in a node script. Doing anything else seemed like "perfect being the enemy of the good". Doing this in node means I don't want to worry about 3 different shells on Windows (powershell, bash, or cmd), and it works on *nix of course. Note that I did look for Makefile parsing packages, but none of them were adequate.

* https://github.com/mklabs/make - looks like make but only processes some Makefile-looking files
* https://github.com/kba/makefile-parser - gives you a faux-parsed Makefile, but doesn't evaluate variables before doing so.
* https://github.com/codejamninja/node-gnumake - downloads random binaries off the Internet

Another alternative was calling node in Makefile but uh, that wasn't going to go over well, and I wanted zq to still be Go-only for most folks.

That left this proposed solution.

mikesbrown:
Offline discussion is that we want the package.json version to be latest-dev, not presumed-next-dev.

PR #713, opened 2020-05-01
zqd: Ensure posted space name is unique

mattnibs:
Previously we were returning an error if a user tried to create
a space with name that already existed. Allow a duplicate name to
be posted and use fs.UniqueDir to make the name unique.

### DIFF
--- a/zqd/space/space.go
+++ b/zqd/space/space.go
@@ -120,7 +120,6 @@ func (m *Manager) Create(name, dataPath string) (*api.SpacePostResponse, error) 
 	}
 
 	if _, exists := m.spaces[name]; exists {
-		m.mapLock.Unlock()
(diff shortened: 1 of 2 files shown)


zqd/space/space.go
```diff
@@ -130,6 +119,12 @@ func (m *Manager) Create(name, dataPath string) (*api.SpacePostResponse, error)
 		return nil, err
 	}
 
+	if _, exists := m.spaces[name]; exists {
+		m.mapLock.Unlock()
```
alfred-landrum, line 123: the defer will already unlock the mutex, so this isn't needed

mattnibs: whoops

mattnibs, line 123:
```suggestion
```

PR #717, opened 2020-05-03
spawn zeek on windows with new hidden subcommand

alfred-landrum:
This adds a hidden zqd subcommand "winexec", that executes a command
using the same job object based spawned process termination we are
already using in the windows zeek-launcher utility. Use this subcommand
when we launch zeek on windows. This is a first step to treating zeek as
more of a black box from inside zqd, so that users can run their own
zeek setup.

### DIFF
--- a/cmd/zqd/winexec/command.go
+++ b/cmd/zqd/winexec/command.go
@@ -1,0 +1,35 @@
+package winexec
+
+import (
+	"flag"
+
+	"github.com/brimsec/zq/cmd/zqd/root"
+	"github.com/mccanne/charm"
+)
+
+func init() {
+	root.Zqd.Add(spec)
+}
+
+var spec = &charm.Spec{
+	Hidden: true,
+	Name:   "winexec",
+	Usage:  "winexec <command> <command-options...>",
+	Short:  "exec helper for windows",
+	Long: `
+Executes the given command and arguments, terminating all spawned processes on exit.
+`,
+	New: newWindowsExecutor,
+}
+
+func newWindowsExecutor(_ charm.Command, _ *flag.FlagSet) (charm.Command, error) {
+	c := &windowsExecutor{}
+	return c, nil
+}
+
+type windowsExecutor struct {
+}
+
+func (w *windowsExecutor) Run(args []string) error {
+	return winexec(args)
+}
--- a/cmd/zqd/winexec/command.go
+++ b/cmd/zqd/winexec/command.go
@@ -15,16 +15,15 @@ var spec = &charm.Spec{
 	Hidden: true,
 	Name:   "winexec",
 	Usage:  "winexec <command> <command-options...>",
-	Short:  "exec helper for windows",
+	Short:  "exec helper for Windows",
 	Long: `
-Executes the given command and arguments, terminating all spawned processes on exit.
+Executes the given command, terminating all spawned processes on exit.
 `,
 	New: newWindowsExecutor,
 }
 
 func newWindowsExecutor(_ charm.Command, _ *flag.FlagSet) (charm.Command, error) {
-	c := &windowsExecutor{}
-	return c, nil
+	return &windowsExecutor{}, nil
 }
 
 type windowsExecutor struct {
--- a/cmd/zqd/winexec/exec.go
+++ b/cmd/zqd/winexec/exec.go
@@ -1,0 +1,9 @@
+// +build !windows
+
+package winexec
+
+import "fmt"
+
+func winexec(_ []string) error {
+	return fmt.Errorf("winexec only for windows platforms")
+}
--- a/cmd/zqd/winexec/exec.go
+++ b/cmd/zqd/winexec/exec.go
@@ -2,8 +2,8 @@
 
 package winexec
 
-import "fmt"
+import "errors"
 
 func winexec(_ []string) error {
-	return fmt.Errorf("winexec only for windows platforms")
+	return errors.New("winexec is only available on Windows")
 }
--- a/cmd/zqd/winexec/exec_windows.go
+++ b/cmd/zqd/winexec/exec_windows.go
@@ -1,7 +1,7 @@
 package winexec
 
 import (
-	"fmt"
+	"errors"
 	"os"
 	"os/exec"
 	"unsafe"
@@ -28,8 +28,7 @@ func ensureSpawnedProcessTermination() error {
 
 	// We add ourselves so that any process we launch will automatically be
 	// added to the job.
-	err = jo.AddCurrentProcess()
-	if err != nil {
+	if err := jo.AddCurrentProcess(); err != nil {
 		return err
 	}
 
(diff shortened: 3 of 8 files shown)


cmd/zqd/winexec/exec_windows.go
```diff
@@ -0,0 +1,68 @@
...
+	"fmt"
+	"os"
+	"os/exec"
+	"unsafe"
+
+	"github.com/alexbrainman/ps"
+	"github.com/alexbrainman/ps/winapi"
+)
+
+// ensureSpawnedProcessTermination ensures that when this Go process terminates, the
```
alfred-landrum, line 13: fyi, this is all taken from the launcher utility already in use on windows:
https://github.com/brimsec/zeek/blob/master/brim/windows/launcher/zeek-launcher.go#L72

nwt:
Rather than use a subcommand like this, it might be a little cleaner and clearer to
1. start the Zeek process with CREATE_SUSPENDED,
2. create and configure the job object (including adding the Zeek process to it) in zqd,
3. use DuplicateHandle to give the Zeek process a handle to the job object,
4. close the job object handle in zqd, and finally
5. call ResumeThread on the Zeek process's thread.

alfred-landrum:
> Rather than use a subcommand like this, it might be a little cleaner and clearer to

So, funny story, I tried that first. :) And though it worked, I didn't like the solution, because of what it took to get the thread to resume:

If you create the process with CREATE_SUSPENDED, when you want to resume it, you have to resume the processes main (only) initial thread, via ResumeThread. To call ResumeThread, you need a handle to the thread, and for a handle, you need the thread id. 

To get the thread id, the only option I found is to use the win32 api `CreateToolhelp32Snapshot`, and then iterate through all threads in the system (!), looking for the one with the correct parent process. Even if you specify only the process id you're interested in, the `CreateToolhelp32Snapshot` will always take a snapshot of the state of every system thread, and give you a handle to that snapshot, which you then need to iterate on with other calls.

I implemented all of this myself, both the create suspended & resuming the thread, and got it working. I later found that there's a test utility in the go source code that implements the resume handling:
https://github.com/golang/go/blob/master/src/runtime/syscall_windows_test.go#L828

If it were possible to get the created processes thread id through the exec.Cmd calls, this would be much easier. But that information is thrown away in `syscall.StartProcess`, which is near the bottom of the chain of `exec.Cmd.Start()`.

Overall, I thought that needing to have all the extra syscall definitions, and then creating a snapshot of all system threads & iterating on it, felt pretty complex to have around. So that led me to this PR's approach.

So with more knowledge about the implementation, let me know what you think between the two approaches.

cmd/zqd/winexec/command.go
```diff
@@ -0,0 +1,35 @@
...
+	Short:  "exec helper for windows",
+	Long: `
+Executes the given command and arguments, terminating all spawned processes on exit.
+`,
+	New: newWindowsExecutor,
+}
+
+func newWindowsExecutor(_ charm.Command, _ *flag.FlagSet) (charm.Command, error) {
+	c := &windowsExecutor{}
+	return c, nil
```
nwt, line 27:
```suggestion
	return &windowsExecutor{}, nil
```

cmd/zqd/winexec/exec.go
```diff
@@ -0,0 +1,9 @@
+// +build !windows
+
+package winexec
+
+import "fmt"
+
+func winexec(_ []string) error {
+	return fmt.Errorf("winexec only for windows platforms")
```
nwt, line 8:
```suggestion
import "errors"

func winexec(_ []string) error {
	return errors.New("winexec is only available on Windows")
```

cmd/zqd/winexec/command.go
```diff
@@ -0,0 +1,35 @@
...
+
+func init() {
+	root.Zqd.Add(spec)
+}
+
+var spec = &charm.Spec{
+	Hidden: true,
+	Name:   "winexec",
+	Usage:  "winexec <command> <command-options...>",
+	Short:  "exec helper for windows",
```
nwt, line 18:
```suggestion
	Short:  "exec helper for Windows",
```

cmd/zqd/winexec/command.go
```diff
@@ -0,0 +1,35 @@
...
+	Long: `
+Executes the given command and arguments, terminating all spawned processes on exit.
```
nwt, line 20:
```suggestion
Executes the given command, terminating all spawned processes on exit.
```

cmd/zqd/winexec/exec_windows.go
```diff
@@ -0,0 +1,68 @@
...
+	// process needs to find or interact with it.
+	jo, err := ps.CreateJobObject("")
+	if err != nil {
+		return err
+	}
+
+	// We add ourselves so that any process we launch will automatically be
+	// added to the job.
+	err = jo.AddCurrentProcess()
+	if err != nil {
```
nwt, line 32:
```suggestion
	if err := jo.AddCurrentProcess(); err != nil {
```

cmd/zqd/winexec/exec_windows.go
```diff
@@ -0,0 +1,68 @@
...
+			LimitFlags: winapi.JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE,
+		},
+	}
+	return winapi.SetInformationJobObject(jo.Handle, winapi.JobObjectExtendedLimitInformation,
+		uintptr(unsafe.Pointer(&limitInfo)), uint32(unsafe.Sizeof(limitInfo)))
+}
+
+func winexec(args []string) error {
+	if len(args) == 0 {
+		return fmt.Errorf("expected command to execute")
```
nwt, line 55:
```suggestion
		return errors.New("command required")
```

cmd/zqd/winexec/exec_windows.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	err := ensureSpawnedProcessTermination()
+	if err != nil {
+		return err
+	}
+
```
nwt, line 62:
```suggestion
	if err := ensureSpawnedProcessTermination(); err != nil {
		return err
	}
```

cmd/zqd/winexec/exec_windows.go
```diff
@@ -0,0 +1,68 @@
+package winexec
+
+import (
+	"fmt"
```
nwt, line 4:
```suggestion
	"errors"
```

zqd/zeek/zeek.go
```diff
@@ -63,8 +77,8 @@ type process struct {
 	stderrBuf *bytes.Buffer
 }
 
-func newProcess(ctx context.Context, pcap io.Reader, zeekpath, outdir string) *process {
-	cmd := exec.CommandContext(ctx, zeekpath, "-C", "-r", "-", "--exec", ExecScript, "local")
+func newProcess(ctx context.Context, pcap io.Reader, args []string, outdir string) *process {
```
nwt, line 80: Args doesn't feel like the right name for this since it includes the command name. Maybe call it command? Or change the signature to resemble os/exec.Command, like this?
```go
func newProcess(ctx context.Context, pcap io.Reader, outdir string, name string, args []string) *process {
```

PR #718, opened 2020-05-04
change zqd option to specify zeek runner

alfred-landrum:
This is the brimsec/zq repo side of brimsec/brim#731.

The cli options used when launching zeek are removed, and will be managed in the zeek repo ( https://github.com/brimsec/zeek/pull/23 ). 

The zqd cli option to specify a zeek runner is `-zeekrunner`, which should either be a full path to a command or the name of a command available in PATH.


Makefile
```diff
@@ -4,7 +4,7 @@ export GO111MODULE=on
 # npm/build.
 VERSION = $(shell git describe --tags --dirty --always)
 LDFLAGS = -s -X main.version=$(VERSION)
-ZEEKTAG = v3.0.2-brim1
+ZEEKTAG = dev-zeek-runner
```
alfred-landrum, line 7: Will update this once https://github.com/brimsec/zeek/pull/23 is merged & we have a new zeek release that has the zeekrunner.

zqd/zeek/zeek.go
```diff
@@ -12,16 +12,6 @@ import (
...
-// ExecScript will be fed into a launched zeek process as the --exec option. The
-// default script disables the packet_filter and loaded scripts logs. These logs
-// are disabled because the emit either timeless logs or logs with timestamp
-// set to execution time rather than time of capture.
-var ExecScript = `
-event zeek_init() {
-	Log::disable_stream(PacketFilter::LOG);
-	Log::disable_stream(LoadedScripts::LOG);
-}`
-
```
mattnibs, line 24: Given that we're getting rid of this, is the api able to handle timeless points? Do we have a test for this?

alfred-landrum: We're still using this setup, but it's moved (like the zeek command line arguments) into the zeekrunner script; see https://github.com/brimsec/zeek/pull/23.

Issue #686, opened 2020-04-28
zqd support for space ids


alfred-landrum:
Add support for space ids, uuid-like immutable identifiers that will be used instead of space names as the resource id in api calls. The id should be created by zqd when the space is returned, and passed back in the create response struct. 

To make the transition easy from the existing name-based calls, I suggest that we continue to treat the directory name under the `spaces` directory as the space's id. So before this work goes in, if there was a directory `spaces/foo`, because space foo was created, we treat `foo` as the id for that space going forward. After this work goes in, though, when a new space is created with name `bar`, the fs would look like:
`space/ae123428d33.../`
`space/ae123428d33.../config.json`

where `config.json` now has an entry `name: "bar"`.

philrz:
Verified in Brim commit `22604c1` talking to `zqd` commit `f9a90eb`.

I triggered the creation of a couple Spaces, one via pcap import and one via Zeek log import. I can see the opaque Space names now below my `data/` directory.

```
~/Library/Application Support/Brim/data/spaces$ tree
.
├── sp_1c8HwI2VOUQWlgvIoELq51XrFZG
│   ├── all.zng
│   ├── config.json
│   ├── info.json
│   └── packets.idx.json
└── sp_1c8Hz0OM81fitoUQagFvW0uLKrk
    ├── all.zng
    ├── config.json
    └── info.json

2 directories, 7 files
```

Thanks @mattnibs!

PR #721, opened 2020-05-04
Space IDs

### LINKED ISSUE #687: space list api should return more info
The api that currently returns a list of space names (`GET /space`) should return an array of structs that include at least the space's name and id. We'll add more to this struct over time as needed by Brim.

mattnibs:
Update space api to support immutable space IDs. The space uses
segmentio/ksuid to generate a date sortable uuid. The space id is
also prefixed by the first 5 alphanumeric characters of the name
or post data path base name.

The space.ID is the name of the space's directory in the zqd root.

There is now no constraints on the characters that can be in
space.Name. Additionally space.Name does not have to be unique
among other spaces.

In order to support backwards compatibility the names of all
previously existing spaces are now the ids of the space.

Also:
- Change spaces list api to return a list of api.SpaceDesc structs
instead of just a list of space names as strings.
- For consistency sake change api.SpaceInfo.DataDir/DataPath

closes #687 #686

### DIFF
--- a/zqd/api/api.go
+++ b/zqd/api/api.go
@@ -35,7 +35,7 @@ type TaskEnd struct {
 }
 
 type SearchRequest struct {
-	Space string          `json:"space" validate:"required"`
+	Space SpaceID         `json:"space" validate:"required"`
 	Proc  json.RawMessage `json:"proc" validate:"required"`
 	Span  nano.Span       `json:"span"`
 	Dir   int             `json:"dir" validate:"required"`
@@ -72,8 +72,12 @@ type ScannerStats struct {
 	RecordsMatched int64 `json:"records_matched"`
 }
 
+type SpaceID string
+
 type SpaceInfo struct {
+	ID          SpaceID    `json:"id"`
 	Name        string     `json:"name"`
+	DataPath    string     `json:"data_path"`
 	Span        *nano.Span `json:"span,omitempty"`
 	Size        int64      `json:"size" unit:"bytes"`
 	PcapSupport bool       `json:"pcap_support"`
@@ -87,11 +91,13 @@ type StatusResponse struct {
 }
 
 type SpacePostRequest struct {
-	Name    string `json:"name"`
-	DataDir string `json:"data_dir"`
+	Name     string `json:"name"`
+	DataPath string `json:"data_path"`
 }
 
-type SpacePostResponse SpacePostRequest
+type SpacePutRequest struct {
+	Name string `json:"name"`
+}
 
 type PcapPostRequest struct {
 	Path string `json:"path"`
--- a/zqd/space/manager.go
+++ b/zqd/space/manager.go
@@ -1,0 +1,140 @@
+package space
+
+import (
+	"errors"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"sync"
+
+	"github.com/brimsec/zq/zqd/api"
+	"github.com/brimsec/zq/zqe"
+	"go.uber.org/zap"
+)
+
+type Manager struct {
+	rootPath string
+	mapLock  sync.Mutex
+	spaces   map[api.SpaceID]*Space
+	logger   *zap.Logger
+}
+
+func NewManager(root string, logger *zap.Logger) (*Manager, error) {
+	mgr := &Manager{
+		rootPath: root,
+		spaces:   make(map[api.SpaceID]*Space),
+		logger:   logger,
+	}
+
+	dirs, err := ioutil.ReadDir(root)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, dir := range dirs {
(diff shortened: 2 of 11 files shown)


zqd/space/manager.go
```diff
@@ -0,0 +1,151 @@
...
+		}
+
+		space := newSpace(path, config)
+		mgr.spaces[space.ID()] = space
+	}
+
+	return mgr
+}
+
+func newSpaceID(prefix string) string {
```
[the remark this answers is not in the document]
mattnibs: Ok. Will there be any other folders in the root dir in the future that would be a space?

zqd/space/manager.go
```diff
@@ -0,0 +1,151 @@
...
+	"path/filepath"
+	"regexp"
+	"sync"
+
+	"github.com/brimsec/zq/zqd/api"
+	"github.com/brimsec/zq/zqe"
+	"github.com/segmentio/ksuid"
+	"go.uber.org/zap"
+)
+
```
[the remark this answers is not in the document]
mattnibs: I am struggling to see the value add of doing this. Are getting at that in the future you could add functionality that would parse the ksuid and get the date meta data stored in it?

zqd/space/manager.go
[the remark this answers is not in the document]
mattnibs: I put `sp` (no underscore) in front of the space id.

zqd/api/api.go
```diff
@@ -72,7 +72,14 @@ type ScannerStats struct {
...
 }
 
+type SpaceDesc struct {
+	ID      string `json:"id"`
+	Name    string `json:"name"`
+	DataDir string `json:"data_dir"`
+}
+
 type SpaceInfo struct {
+	ID            string     `json:"id"`
```
nwt, line 82: Having both SpaceDesc and SpaceInfo feels a little funny, perhaps because when you should expect the API to use one or the other ins't really clear from the names. Would it make sense to drop SpaceDesc, add the DataDir field to SpaceInfo, and then use SpaceInfo everywhere?

mattnibs: This is really more about how api responses are rendered in json. When you as the api to list all the spaces for zqd, I would imagine that sending back all the metadata associated with the space would be too much information. `SpaceDesc` should really be embedded in `SpaceInfo` though I'm pretty sure encoding/json would output something undesirable.

zqd/space/manager.go
```diff
@@ -0,0 +1,147 @@
...
+var idregexp = regexp.MustCompile("[^a-zA-Z0-9]+")
```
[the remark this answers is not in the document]
mattnibs: whoops

zqd/space/manager.go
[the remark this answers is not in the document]
alfred-landrum: No, but having a specific type for the id's will identify them in function arguments, make it easier to track down users of ids, and other general benefits of having a unique type different from just string.

zqd/space/space.go
```diff
@@ -37,11 +36,9 @@ var (
...
 )
 
-type Manager struct {
-	rootPath string
-	mapLock  sync.Mutex
-	spaces   map[string]*Space
-	logger   *zap.Logger
+func newSpaceID() api.SpaceID {
+	id := ksuid.New()
+	return api.SpaceID(fmt.Sprintf("sp%s", id.String()))
```
alfred-landrum, line 41: Would you please add back the "_", so the prefix is "sp_" ?

zqd/api/api.go
```diff
@@ -72,7 +72,16 @@ type ScannerStats struct {
 	RecordsMatched int64 `json:"records_matched"`
 }
 
+type SpaceID string
+
+type SpaceDesc struct {
```
alfred-landrum, line 77: I had similar questions as noah about SpaceDesc vs SpaceInfo. At the moment, I see Brim making a lot of requests to get the full space info, as they're trying to watch for updates to the span & size during ingest, so there's no expectation that SpaceInfo is "a lot" of info to gather & return. Once we build in support for spaces backed by non-local files, we'll have a better idea of what's cheap for a SpaceList call vs not, so maybe we can hold off on this split until then.

mattnibs: ok

mattnibs: I'll do this but I do want to say this- I don't think this is about making SpaceList a cheaper call. As config options grow (they most certainly will) we are going to be adding more and more information to space info. It feels like there are circumstances where this is TMI. For instance the response in a SpacePost as well as SpaceList- where I mostly just care about the ID of the space and the name.

(All smaller issue to consider) This also adds additional complexity to tests where the expected response now has to account for larger range of variables.

zqd/space/space.go
mattnibs: Done.

PR #723, opened 2020-05-05
zngio: don't swallow errors when reading

henridf:
The zngio.Reader was not returning errors from the underlying io.Reader. This PR fixes that.

### DIFF
--- a/zio/zngio/reader.go
+++ b/zio/zngio/reader.go
@@ -95,8 +95,11 @@ func (r *Reader) LastSOS() int64 {
 func (r *Reader) ReadPayload() (*zng.Record, []byte, error) {
 again:
 	b, err := r.read(1)
-	if err == io.EOF || len(b) == 0 {
+	if err == io.EOF || err == peeker.ErrTruncated {
 		return nil, nil, nil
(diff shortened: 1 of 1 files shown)


zio/zngio/reader.go
```diff
@@ -95,9 +95,12 @@ func (r *Reader) LastSOS() int64 {
 func (r *Reader) ReadPayload() (*zng.Record, []byte, error) {
 again:
 	b, err := r.read(1)
-	if err == io.EOF || len(b) == 0 {
+	if err == io.EOF || err == peeker.ErrTruncated {
 		return nil, nil, nil
 	}
+	if err != nil {
+		return nil, nil, err
+	}
```
nwt, line 103: How about nesting the first if inside the second? And why does it make sense to treat peeker.ErrTruncated this way?

henridf: ErrTruncated means that there is nothing left to read (after doing a peek(1) specifically.

henridf: (in other words, it is like an io.EOF  in this case. i'll add a comment).

PR #725, opened 2020-05-06
fix zng reader to properly handle stream reset

### LINKED ISSUE #724: uid correlation query failing: slice bounds out of range
Found on Brim tagged `v0.9.0` talking to `zqd` tagged `0.13.0`. It doesn't seem to trigger with all pcaps, but I can reliably repro with https://archive.wrccdc.org/pcaps/2018/wrccdc.2018-03-23.010014000000000.pcap.gz. I first saw this when smoke testing on Linux, but just reproduced on macOS as well.

Once uncompressed and imported into Brim, the Wireshark button does not activate for any events I click on. When I sniff in Wireshark, I see an error coming back from `zqd` which I can repro with `curl`:

```
$ curl "http://localhost:9867/search?format=zjson" -H "Content-Type: text/plain;charset=UTF-8" -d '{"proc":{"op":"SequentialProc","procs":[{"op":"FilterProc","filter":{"op":"LogicalOr","left":{"op":"LogicalOr","left":{"op":"LogicalOr","left":{"op":"CompareField","comparator":"=","field":{"op":"FieldRead","field":"uid"},"value":{"op":"Literal","type":"string","value":"CeAR584pP6sHvgnrLe"}},"right":{"op":"CompareField","comparator":"in","field":{"op":"FieldRead","field":"conn_uids"},"value":{"op":"Literal","type":"string","value":"CeAR584pP6sHvgnrLe"}}},"right":{"op":"CompareField","comparator":"in","field":{"op":"FieldRead","field":"uids"},"value":{"op":"Literal","type":"string","value":"CeAR584pP6sHvgnrLe"}}},"right":{"op":"CompareField","comparator":"=","field":{"op":"FieldCall","fn":"RecordFieldRead","field":{"op":"FieldRead","field":"referenced_file"},"param":"uid"},"value":{"op":"Literal","type":"string","value":"CeAR584pP6sHvgnrLe"}}}},{"op":"HeadProc","count":100}]},"space":"wrccdc.2018-03-23.010014000000000.pcap.brim","dir":-1,"span":{"ts":{"sec":1521835102,"ns":636000000},"dur":{"sec":111,"ns":696000000}}}'
{"type":"TaskStart","task_id":0}


{"type":"TaskEnd","task_id":0,"error":{"type":"INTERNAL","kind":"","error":"panic: runtime error: slice bounds out of range [44:2]"}}
```

mccanne:
This commit fixes a bug in the zng reader where we were not properly
tracking the internal type context separately from the passed-in shared context.
The fix is to keep an internal context along with the external context
and teardown the mapper that maps between them whenever the context
is reset.

We also added a test that exercises this case by encoding two different
values with a max-stream size of 1 record so the two records will be
different types but have the same type ID in the mapper table.

Resolves #724


mccanne:
> Could you change the test file name to something meaningful, such as maybe `zctx-reset.yaml` ?

Sure.  I was copying the pattern Andrew created for the stream test.

henridf:
> Sure. 

👍 it gets really hard to figure out what tests are doing what when you do a `find ./tests` and get a wall of names like test1.yaml, test2.yaml, ...

PR #726, opened 2020-05-06
add "test" target to Makefile for easy isolated test runs

mccanne:
This change to the Makefile makes it much easier to run isolated
tests that fail by cutting and pasting the go test path that
you want to run and assinging to a var called TEST on the make
command line, e.g.,

	make test TEST=TestZTest/formats/zng/streams-3

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,9 @@ test-unit:
 test-system: build
 	@ZTEST_BINDIR=$(CURDIR)/dist go test -v ./tests
 
+test: build
+	@ZTEST_BINDIR=$(CURDIR)/dist go test -v -tags=system ./tests -run $(TEST) -args PATH=$(shell pwd)/dist
+
 test-heavy: build $(SAMPLEDATA)
 	@go test -v -tags=heavy ./tests
 
--- a/Makefile
+++ b/Makefile
@@ -7,9 +7,8 @@ LDFLAGS = -s -X main.version=$(VERSION)
 ZEEKTAG = v3.0.2-brim1
 ZEEKPATH = zeek-$(ZEEKTAG)
 
-# By putting this optional target first, conditiional on the TEST env var,
-# you can run just one test using by simiply typing "make TEST=go-test-path",
-# e.g., "make TEST=TestZTest/suite/cut/cut"
+# This enables a shortcut to run a single test from the ./tests suite, e.g.:
+# make TEST=TestZTest/suite/cut/cut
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -42,6 +42,9 @@ test-unit:
 test-system: build
 	@ZTEST_BINDIR=$(CURDIR)/dist go test -v -tags=system ./tests -args PATH=$(shell pwd)/dist
 
+test: build
```
henridf, line 45: `test` seems like prime real estate for this special target... how about `test-run` or something?
also i think the `-tags=system` and `PATH=...` are unnecessary (independently ran into this and put up #728)

mccanne:
> (directly using `go test -run` seems more useful to me but ymmv!)

You have to remember to do a "make install" whenever you make a source code change then in order to get the script tests to run you have to have env  ZTEST_BINDIR.  Am I missing something?

henridf:
> You have to remember to do a "make install" whenever you make a source code change then in order to get the script tests to run you have to have env ZTEST_BINDIR. Am I missing something?

No I think thats right! I was probably doing `go run` for non-script tests, and I agree that for script tests, it isn't convenient.

Makefile
mccanne: > `test` seems like prime real estate for this special target... how about `test-run` or something?
> also i think the `-tags=system` and `PATH=...` are unnecessary (independently ran into this and put up #728)

I just pushed a change inspired by your comment here... now you can just say 
```
make TEST=TestZTest/suite/zar/ls 
```
using Makefile if-logic.  And I took out system tag and PATH.  What do you think?

henridf: looks good. but i merged #728 so you have a confict now 😬

Makefile
```diff
@@ -7,6 +7,13 @@ LDFLAGS = -s -X main.version=$(VERSION)
 ZEEKTAG = v3.0.2-brim1
 ZEEKPATH = zeek-$(ZEEKTAG)
 
+# By putting this optional target first, conditiional on the TEST env var,
+# you can run just one test using by simiply typing "make TEST=go-test-path",
+# e.g., "make TEST=TestZTest/suite/cut/cut"
```
alfred-landrum, line 12:
```suggestion
# This enables a shortcut to run a single test from the ./tests suite, e.g.:
# make TEST=TestZTest/suite/cut/cut
```

mccanne: ok

PR #730, opened 2020-05-06
add zng.Builder.Parse()

mccanne:
This commit adds a Parse method to the zng builder so that records
can be built from a DFS traversal of the string representation of
the leaf values.  A future PR will define a syntax for record values
and do a proper recursive descent here, but for now, this will be
used by the upcoming zar/zdx PR for multi-value keys.

### DIFF
--- a/zng/builder.go
+++ b/zng/builder.go
@@ -1,8 +1,14 @@
 package zng
 
 import (
+	"errors"
+	"fmt"
+	"strings"
+
 	"github.com/brimsec/zq/zcode"
 )
+
+var ErrIncomplete = errors.New("not enough values supplied to complete record")
 
 // Builder provides a way of easily and efficiently building records
 // of the same type.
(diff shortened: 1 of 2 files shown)


zng/builder.go
```diff
@@ -48,3 +54,168 @@ func (b *Builder) Build(zvs ...zcode.Bytes) *Record {
 	}
 	return &b.rec
 }
+
+// Parse creates a record from the zng string representation of each leaf field
```
mattnibs, line 58: by "zng string" do you mean "tzng"?

zng/builder.go
```diff
@@ -48,3 +54,168 @@ func (b *Builder) Build(zvs ...zcode.Bytes) *Record {
...
+	in = in[1 : len(in)-1]
+	if len(in) == 0 {
+		// empty set
+		b.appendUnset(inner)
+		return nil
+	}
+	if IsContainerType(inner) {
+		return &RecordTypeError{Name: "<set>", Type: typ.String(), Err: ErrNotPrimitive}
+	}
+	//XXX for now just use simple comma rule, which means comman
```
alfred-landrum, line 193: I have a question: A comma is different than the `;` separator we use to separate items in tzng, why not use a semicolon?

zng/builder.go
```diff
@@ -48,3 +54,168 @@ func (b *Builder) Build(zvs ...zcode.Bytes) *Record {
...
 }
+
+// Parse creates a record from the zng string representation of each leaf field
+// in the DFS traversal of the record type.  If there aren't enough inputs values
+// to occupy every leaf value, then those values are left unset, in which case
+// a valid record is returned along with ErrIncomplete.
+// XXX This currently does not work for embedded sets, unions, and so forth.
+// We need to define a record value syntax and do proper recusive descent
+// parsing here.
+func (b *Builder) Parse(vals ...string) (*Record, error) {
```
nwt, line 65: Nit: I think BuildFromStrings or something along that line would be a better name for this.

zng/builder.go
mccanne: I don't think tzng syntax is the right thing.  The question is what is the zql syntax for literals of the container types?

PR #732, opened 2020-05-06
include stack trace on panic errors

alfred-landrum:
With this change, if a panic occurs either inside one of the mux goroutines, or inside the chain of an api handler, we will:
* log the panic message and stack trace to the zq-core.log file
* send the error message back to Brim, which will display the error & stack trace in the "yellow error" field, where it can be copied, pasted, and reported.

### DIFF
--- a/zqd/middleware.go
+++ b/zqd/middleware.go
@@ -2,12 +2,12 @@ package zqd
 
 import (
 	"context"
-	"fmt"
 	"net/http"
 	"strconv"
 	"sync/atomic"
 	"time"
 
+	"github.com/brimsec/zq/zqe"
 	"github.com/gorilla/mux"
 	"go.uber.org/zap"
 )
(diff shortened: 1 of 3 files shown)


zqd/middleware.go
```diff
@@ -76,9 +76,9 @@ func panicCatchMiddleware(logger *zap.Logger) mux.MiddlewareFunc {
 				if r == nil {
 					return
 				}
-				rstr := fmt.Sprint(r)
-				logger.DPanic("Panic", zap.String("error", rstr))
-				http.Error(w, rstr, http.StatusInternalServerError)
+				err := zqe.RecoverError(r)
+				logger.DPanic("panic", zap.Error(err))
+				http.Error(w, err.Error(), http.StatusInternalServerError)
```
nwt, line 81:
```suggestion
				logger.DPanic("Panic", zap.String("error", fmt.Sprint(r)))
				err := zqe.RecoverError(r)
				http.Error(w, err.Error(), http.StatusInternalServerError)
```
since
1. we usually capitalize the first word of the log message and
2. zap already includes the stack trace at DpanicLevel.

Issue #613, opened 2020-04-21
occasional file i/o "access is denied" errors during pcap ingest in windows CI with brim


mikesbrown:
Several tests for Brim / zq integration involve uploading a pcap through Brim and then doing something with it. In [one run](https://github.com/brimsec/brim/runs/596826483), there were i/o errors that prevented ingest from succeeding.

One was
```
CreateFile
C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\
scoped_dir596_1117112332\\upload596_235124938\\
sample.pcap.brim\\.tmp.ingest\\capture_loss.log:
Access is denied.
```

The other was
```
rename C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\
scoped_dir940_1959449935\\upload940_1015111360\\
sample.pcap.brim\\all.bzng.tmp 

C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\
scoped_dir940_1959449935\\upload940_1015111360\\
sample.pcap.brim\\all.bzng: Access is denied
```

The bug is filed here as the error is happening in `zqd` and being propagated to Brim.

The file being ingested is [sample.pcap](https://github.com/brimsec/brim/blob/1890b07c91cf63f8d64fdb5c764229046ee6bc8e/itest/sample.pcap) from brim.git.

The logs of the run are available [here](https://github.com/brimsec/brim/suites/607484043/artifacts/4585797).

brim was at https://github.com/brimsec/brim/commit/1890b07c91cf63f8d64fdb5c764229046ee6bc8e . zq was at tag v0.11.1. Note that windows integration testing is quite new; we don't have many past data points. But so far this is the only known instance.

mikesbrown:
Happened here https://github.com/brimsec/brim/runs/637035303?check_suite_focus=true . Confirmed with screen shot. The zq hash is e7b866d which is an experimental branch. No changes were made to zq itself on this branch. The parent hash from master is e40934a

alfred-landrum:
thanks for pointer @mikesbrown - I did pull the archive for that test run. The logs have very little data to help with the issue, I may need to add some to help triage.

mikesbrown:
https://github.com/brimsec/brim/actions/runs/95611919

Looks like this can now manifest with a user-facing message like this:
```
Detail: D:\var\tmp\brimsec\itest\histogram.test.js\1\data\spaces\sample.pcap.brim\.tmp.ingest\conn.log: line 1: bad format
```
[...]

PR #741, opened 2020-05-08
open files on windows with FILE_SHARE_DELETE

alfred-landrum:
Address the occasional rename related failures on Windows by using the FILE_SHARE_DELETE flag when opening files on Windows. As noted in https://github.com/brimsec/zq/issues/613, this unfortunately requires copying substantial code from the os package, as there's no way to use that flag with the existing os.Open api.

Fixes https://github.com/brimsec/zq/issues/613


alfred-landrum:
I've created a Brim branch that uses this zq, and it shows many clean CI runs on Windows:
https://github.com/brimsec/brim/pull/763

pkg/fs/open_windows.go
```diff
@@ -0,0 +1,258 @@
+package fs
```
henridf, line 1: We should add the original copyright notice here. Also for future reference, the go version that these funcs taken from. (Optionally, since these appear to come from different files, mention which ones too?)

henridf:
The zql `os.Open`s  are still there, how about porting those over too?

alfred-landrum:
> The zql/ `os.Open`s are still there, how about porting those over too?

So zql/zql.go is a generated file from pigeon, which is why I had not updated it. I can make update the 2 use cases in zql_test.go, though.

aswan:
A parser's `OpenFile()` method is boilerplate that pigeon creates but we never use that method...

henridf:
> So zql/zql.go is a generated file from pigeon, which is why I had not updated it. I can make update the 2 use cases in zql_test.go, though.

Ah, had missed that. My goal was simply to make it easy to `grep -R os.Open` and be sure there were none.

PR #743, opened 2020-05-08
Add `/search?noctrl=true` query param to disable control messages

### LINKED ISSUE #644: api to support zng file save

henridf:
closes #644

### DIFF
--- a/zqd/handlers.go
+++ b/zqd/handlers.go
@@ -98,14 +98,18 @@ func handleSearch(c *Core, w http.ResponseWriter, r *http.Request) {
 	defer srch.Close()
 
 	var out search.Output
+	ctrl := true
+	if r.URL.Query().Get("noctrl") != "" {
+		ctrl = false
+	}
 	format := r.URL.Query().Get("format")
 	switch format {
 	case "zjson", "json":
 		// XXX Should write appropriate ndjson content header.
-		out = search.NewJSONOutput(w, search.DefaultMTU)
+		out = search.NewJSONOutput(w, search.DefaultMTU, ctrl)
 	case "zng":
 		// XXX Should write appropriate zng content header.
-		out = search.NewZngOutput(w)
+		out = search.NewZngOutput(w, ctrl)
 	default:
 		respondError(c, w, r, zqe.E(zqe.Invalid, "unsupported format: %s", format))
 		return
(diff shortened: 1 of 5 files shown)


zqd/handlers_test.go
```diff
@@ -47,6 +47,41 @@ func TestSearch(t *testing.T) {
...
+	require.NoError(t, err)
+	proc, err := json.Marshal(parsed)
+	require.NoError(t, err)
+	req := api.SearchRequest{
+		Space: "test",
+		Proc:  proc,
+		Span:  nano.MaxSpan,
+		Dir:   -1,
+	}
+	r, err := client.Search(context.Background(), req, map[string]string{"noctrl": "true"})
```
alfred-landrum, line 72: I was confused about how this was sending back zng without a format query param; the client always specifies zng, we'll have to change that in https://github.com/brimsec/zq/issues/744 .

zqd/handlers.go
```diff
@@ -98,14 +98,18 @@ func handleSearch(c *Core, w http.ResponseWriter, r *http.Request) {
...
 	format := r.URL.Query().Get("format")
 	switch format {
 	case "zjson", "json":
 		// XXX Should write appropriate ndjson content header.
-		out = search.NewJSONOutput(w, search.DefaultMTU)
+		out = search.NewJSONOutput(w, search.DefaultMTU, ctrl)
 	case "zng":
 		// XXX Should write appropriate zng content header.
-		out = search.NewZngOutput(w)
+		out = search.NewZngOutput(w, ctrl)
```
alfred-landrum, line 112: If ctrl == false, we could truthfully set the content-header to specify that pure zng is returned.

henridf: I think it would be truthful to send that content header even if `ctrl == true`, because the in line control messages are part of the zng spec, which specifies that "Any such application-specific payloads not known by a ZNG data receiver shall be ignored.".

(Yes, this means that it should be ok for the client to get data without even passing this new `noctrl` query param... but i it's still a bit nicer not to have stats and such messages lying around in downloaded zngs)

zqd/handlers_test.go
henridf: 👍 agree.

zqd/handlers.go
henridf: (And re the original comment, I'll file an issue to fix the content-headers in a follow-up.)

henridf: > I'll file an issue 

#745

PR #749, opened 2020-05-08
support multi-key indexes and clean up zar/zdx

mccanne:
This commit adds support for multi-key indexes and refactors
significant pieces of the zar/zdx machinery.

The zar index files are still stored as separate files but
a header has been added to the base layer to provide magic,
version, key fields, and customization of the btree child
field (since it is stored alongside the key fields in the
index and there can be a name collision).

We also reworked index searches to return result over a
channel so that results from a dir walk are easily managed.

A bug is fixed where null values would end up in the index.
It does not make sense to index non-existent values.

We separated the zdx create command into create and convert,
where the latter converts an existing sorted-key zng file
into a zdx and the former creates a sorted-key table using
an in-memory map (this should be replaced eventually with
zql proc code when we have a spill-capable group-by).

There is a new "zdx seek" command, which indexes just
first records after the start-of-stream (sos), where the
value alongside each key is the seek offset of the sos.

We also made the behavior of zar find more sensible with the
-z option, when selecting between file-path output and zng output.

To make multi-key search work properly, we added a new comparator
function, expr.KeyCompareFn, which stores the key in the
closure and returns a comparison result for an arbitrary record
with the stored key.  In this case, the key and the record
can be of different types and the respective key fields in
no consistent order between the types.


archive/indexer.go
```diff
@@ -56,15 +43,16 @@ func IndexDirTree(dir string, rules []Rule, progress chan<- string) error {
...
+		w, err := rule.NewIndexer(zardir)
 		if err != nil {
 			return err
 		}
-		indexers = append(indexers, indexer)
+		writers = append(writers, w)
 		if progress != nil {
-			progress <- fmt.Sprintf("%s: creating index %s", logPath, indexer.Path())
+			//XXX
+			progress <- fmt.Sprintf("%s: creating index", logPath)
```
nwt, line 55: Why the XXX?

cmd/zar/find/command.go
```diff
@@ -18,25 +17,40 @@ import (
...
-that have a corresponding zar index conforming to the indicated <pattern>.
-The "pattern" argument has the form "field=value" (for field searches)
+"zar find" descends the directory given by the -R option (or ZAR_ROOT)
+looking for zng files that have been indexed and performs a search on
+each such index file in accordance withe the specified search pattern.
+An index may either be a standard index file
+ceated by "zar index" or a custom index (-x) file created from an abitrary zng file
+that has been formatted and wrapped by the "zdx convert" or "zar zdx" commands.
+
+The standard indexes, "pattern" argument has the form "field=value" (for field searches)
```
nwt, line 30:
```suggestion
For standard indexes, "pattern" argument has the form "field=value" (for field searches)
```

cmd/zar/find/command.go
```diff
@@ -68,77 +86,63 @@ func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
...
-			for hit := range hits {
-				fmt.Println(hit)
+	}()
+	for hit := range hits {
+		var err error
+		if writer != nil {
+			err = writer.Write(hit)
+		} else {
+			var path string
+			path, err := hit.AccessString(c.pathField)
```
nwt, line 134:
```suggestion
			path, err = hit.AccessString(c.pathField)
```

cmd/zdx/convert/command.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/brimsec/zq/zio"
+	"github.com/brimsec/zq/zio/detector"
+	"github.com/brimsec/zq/zng/resolver"
+	"github.com/mccanne/charm"
+)
+
+var Convert = &charm.Spec{
+	Name:  "convert",
+	Usage: "convert [-f framesize] [ -o file ] -k field[,field,...] file",
+	Short: "generate an zdx file from one or more zng files",
```
nwt, line 20:
```suggestion
	Short: "generate a zdx file from one or more zng files",
```

cmd/zdx/convert/command.go
```diff
@@ -0,0 +1,93 @@
...
+	Long: `
+The convert command generates a zdx index containing keys and optional values
+from the input file.  The required flag -k specifies one or more zng record
+field names that comprise the index search keys, in precedence order.
+The keys must be pre-sorted in ascending order with
+respect to the stream of zng records; otherwise the index will not work correctly.
+The input records are all copied to the base layer of the output index, as is,
+so any information stored alongside the keys (e.g., pre-computed aggregations).
```
nwt, line 28:
```suggestion
so any additional information is stored alongside the keys (e.g., pre-computed aggregations).
```

cmd/zdx/convert/command.go
```diff
@@ -0,0 +1,93 @@
...
+	defer func() {
+		if close {
+			writer.Close()
+		}
+	}()
+	if err := zbuf.Copy(writer, zbuf.Reader(file)); err != nil {
+		return err
+	}
+	close = false
+	return writer.Close()
```
nwt, line 92: Maybe shorten?
```suggestion
	if err := zbuf.Copy(writer, zbuf.Reader(file)); err != nil {
			writer.Close()
			return err
	}
	return writer.Close()
```

cmd/zdx/convert/command.go
```diff
@@ -0,0 +1,93 @@
...
+		//JSONPathRegex:  c.jsonPathRegexp,
+	}
+	file, err := detector.OpenFile(zctx, path, cfg)
+	if err != nil {
+		return err
+	}
+	keys := strings.Split(c.keys, ",")
+	writer, err := zdx.NewWriter(zctx, c.outputFile, keys, c.framesize)
+	if err != nil {
+		return err
```
nwt, line 80:
```suggestion
		file.Close()
		return err
```

cmd/zdx/create/command.go
```diff
@@ -3,73 +3,67 @@ package create
...
 
-func (c *CreateCommand) Run(args []string) error {
-	if !c.inputReady && c.keyField == "" {
-		return errors.New("must specify a key field with -k")
+func (c *Command) Run(args []string) error {
+	if c.keyField == "" {
+		return errors.New("must specify at least one key field with -k")
 	}
+	//XXX no reason to limit this... we will fix this when we refactor
+	// the code here to use zql/proc instead fo the hash table (after we
```
nwt, line 63:
```suggestion
	// the code here to use zql/proc instead of the hash table (after we
```

cmd/zdx/lookup/command.go
```diff
@@ -1,29 +1,35 @@
...
-The lookup command locates the specified <key> in the base file of the
-zdx <bundle> and displays the result as a zng record.
-The key argument specifies a value to look up in the table and must be parseable
-as a zng type of the key that was originally indexed.`,
+The lookup command locates the specified key(s) in the base layer of the
+zdx index and displays the result as a zng record.
+If the index has multiple keys, then multiple records may be returned for
+all the records that match the supplied keys.
+Each key argument specifies a value to look up in the table and must be parseable
+as a zng type of the key that was originally indexed where the keys refer to the leaf
```
nwt, line 30:
```suggestion
as the zng type of the key that was originally indexed where the keys refer to the leaf
```

cmd/zdx/lookup/command.go
```diff
@@ -1,29 +1,35 @@
...
+values in left-to-awrite order of the keys represented as a record, inclusive
```
nwt, line 31:
```suggestion
values in left-to-right order of the keys represented as a record, inclusive
```

archive/indexer.go
mccanne: It used to print the name of the index file too but the writer doesn't have a String()

archive/finder.go
```diff
@@ -1,56 +1,79 @@
...
+		logPath := ZarDirToLog(zardir)
+		for hit := range searchHits {
+			if pathCol != nil {
+				val := zng.Value{pathCol[0].Type, zng.EncodeString(logPath)}
+				hit, err = zctx.AddColumns(hit, pathCol, []zng.Value{val})
+				if err != nil {
+					cancel()
+					for _ = range searchHits {
+						// let search unravel
+					}
```
alfred-landrum, line 64: (similar question as other file, could return err here & let deferred cancel stop the search)

cmd/zdx/lookup/command.go
```diff
@@ -66,43 +72,46 @@ func (c *LookupCommand) Run(args []string) error {
...
+	}()
 	writer, err := emitter.NewFile(c.outputFile, &c.WriterFlags)
 	if err != nil {
 		return err
 	}
-	if err := writer.Write(rec); err != nil {
-		return err
+	for hit := range hits {
+		if err = writer.Write(hit); err != nil {
+			return err
```
alfred-landrum, line 109: this error return leaves writer unclosed

expr/sort.go
```diff
@@ -81,52 +81,54 @@ func NewSortFn(nullsMax bool, fields ...FieldExprResolver) SortFn {
...
+		name, val, err := it.Next()
+		if err != nil {
+			return nil, err
 		}
-		if nullA {
-			if nullsMax {
-				return 1
-			} else {
+		// we got to unset... all remaining values inthe key
+		// must be unset
```
alfred-landrum, line 95:
```suggestion
		// once an unset value is seen, all remaining values are also unset
```

cmd/zdx/seek/command.go
```diff
@@ -0,0 +1,130 @@
...
+
+func (c *Command) Run(args []string) error {
+	//XXX no reason to limit this... fix later
+	if len(args) != 1 {
+		return errors.New("must specify a single zng input file containing keys and optional values")
+	}
+	file := os.Stdin
+	path := args[0]
+	if path != "-" {
+		file, err := os.Open(path)
```
alfred-landrum, line 70: suggest the package fs `fs.Open` function:

```suggestion
		file, err := fs.Open(path)
```

archive/finder.go
```diff
@@ -1,56 +1,79 @@
...
+		var err error
+		go func() {
+			err = search(ctx, zctx, searchHits, path, pattern, skipMissing)
+			close(searchHits)
+		}()
+		logPath := ZarDirToLog(zardir)
+		for hit := range searchHits {
+			if pathCol != nil {
+				val := zng.Value{pathCol[0].Type, zng.EncodeString(logPath)}
+				hit, err = zctx.AddColumns(hit, pathCol, []zng.Value{val})
```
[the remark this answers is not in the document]
mccanne: Oops, yeah, I was using the channel close to provide sync and didn't mean to share err that way.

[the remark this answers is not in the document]
mccanne: Good catch.

archive/indexer.go
mccanne: It felt right to add back a Path() method to Rule so this addresses it.

cmd/zar/find/command.go
mccanne: ugh!

cmd/zdx/convert/command.go
mccanne: Yeah, that pattern looks really silly now that it's not wrapping a bunch of error conditions...

cmd/zdx/convert/command.go
mccanne: I think we want defer file.Close() after open.

archive/finder.go
```diff
@@ -1,56 +1,79 @@
...
-			hit.Keep()
+		pathCol = []zng.Column{{pathField, typ}}
+	}
+	ctx, cancel := context.WithCancel(ctx)
+	return Walk(rootDir, func(zardir string) error {
+		path := filepath.Join(zardir, indexName)
+		searchHits := make(chan *zng.Record)
+		var err error
+		go func() {
+			err = search(ctx, zctx, searchHits, path, pattern, skipMissing)
```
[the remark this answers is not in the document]
mccanne: yes better

zdx/finder.go
```diff
@@ -169,32 +136,90 @@ func (f *Finder) search(key zng.Value) error {
...
+		// columns are "don't care", keep reading records and return
+		// them via the channel.
+		rec, err := lookup(f.files[0], compare, true)
+		if err != nil {
+			return err
+		}
+		if rec == nil {
+			return nil
+		}
+		hits <- rec
```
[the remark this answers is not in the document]
mccanne: yes, much better.  This pattern emphasizes the deadlock avoidance of writing to hits but only if you haven't been canceled.

zdx/header.go
```diff
@@ -0,0 +1,67 @@
...
+const (
+	Magic      = "zdx"
+	Version    = "0.2"
+	ChildField = "_btree_child"
+)
+
+var ErrNotIndex = errors.New("not a zdx index")
+
+func ParseHeader(rec *zng.Record) (string, *zng.TypeRecord, error) {
+	magic, err := rec.AccessString("magic")
```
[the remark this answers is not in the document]
mccanne: ha!  I knew that was coming.

archive/indexer.go
```diff
@@ -56,15 +43,16 @@ func IndexDirTree(dir string, rules []Rule, progress chan<- string) error {
...
-		indexer, err := rule.NewIndexer(zardir)
+		w, err := rule.NewIndexer(zardir)
 		if err != nil {
 			return err
 		}
-		indexers = append(indexers, indexer)
+		writers = append(writers, w)
 		if progress != nil {
-			progress <- fmt.Sprintf("%s: creating index %s", logPath, indexer.Path())
+			//XXX
```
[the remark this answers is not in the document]
mccanne: fixed

cmd/zar/find/command.go
```diff
@@ -68,77 +86,63 @@ func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
...
-	wg.Wait()
-	if err == nil {
-		err = searchErr
+		}
+		if err != nil {
+			cancel()
+			for _ = range hits {
+				// let search goroutine unwind after cancel
+			}
+			return err
```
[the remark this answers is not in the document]
mccanne: I think there is a race between the cancel being detected down below and it getting blocked writing to hits when this guy returns.  I tried too convince myself this race didn't exist but got confused.  If you drain to make sure the close from the other goroutine is sync'd with the close, for sure there is no deadlock.

archive/common.go
```diff
@@ -12,8 +14,25 @@ type IndexerCommon struct {
...
+	zctx *resolver.Context
 }
 
 func (c *IndexerCommon) Path() string {
 	return c.path
 }
+
+// we make the framesize here larger than the writer framesize
+// since the writer always writes a bit past the threshold
+const framesize = 32 * 1024 * 2
```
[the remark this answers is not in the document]
mccanne: Yes, we should file an issue.

tests/suite/zdx/multikey.yaml
```diff
@@ -0,0 +1,33 @@
+script: |
+  # create a sorted-keys base layer with primary key "sum" and secondary
+  # key "s".  Note that the keys do not need to appear in precedence order
+  # in the base zng index.
+  zq -o sorted.zng "sum(v) by s | sort sum,s"  babble.tzng
+  # -x says input keys already sorted and don't create new base records
```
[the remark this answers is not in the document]
mccanne: Yeah, it used to be "create -x" instead of convert.  Will fix the comment.

archive/finder.go
mccanne: see above

expr/sort.go
mccanne: I took Noah's suggestion; he beat you to it.

Issue #747, opened 2020-05-08
Inferred json reader crashes on multiple nested fields


henridf:
I noticed this while working on #714.

```
$ echo '{"a.b.c": 1, "a.b.d": 2}' | zq -t -
panic: runtime error: index out of range [1] with length 1

goroutine 1 [running]:
github.com/brimsec/zq/zio/ndjsonio.(*inferParser).parseObject(0xc000032608, 0xc000342000, 0x18, 0x10000, 0x0, 0x0, 0x0, 0x0, 0x0, 0x3, ...)
        /Users/henridf/work/brim/zq/zio/ndjsonio/inferparser.go:62 +0xe2c
github.com/brimsec/zq/zio/ndjsonio.(*Reader).Parse(0xc000032600, 0xc000342000, 0x18, 0x10000, 0x0, 0x0, 0x0, 0x0, 0x0, 0xc00003c090, ...)
        /Users/henridf/work/brim/zq/zio/ndjsonio/reader.go:112 +0x184
github.com/brimsec/zq/zio/ndjsonio.(*Reader).Read(0xc000032600, 0x13d9f42, 0x18f8ce0, 0xc00003c090)
        /Users/henridf/work/brim/zq/zio/ndjsonio/reader.go:126 +0xac
github.com/brimsec/zq/zio/detector.match(0x1b903c0, 0xc000032600, 0x195f8a4, 0x6, 0x0, 0x0)
        /Users/henridf/work/brim/zq/zio/detector/reader.go:70 +0x38
github.com/brimsec/zq/zio/detector.NewReader(0x1b90340, 0xc0002fb200, 0xc0002e2540, 0xc0002fb200, 0xc0000be000, 0x0, 0x1)
        /Users/henridf/work/brim/zq/zio/detector/reader.go:49 +0x588
github.com/brimsec/zq/zio/detector.OpenFromNamedReadCloser(0xc0002e2540, 0x1b99b20, 0xc0000be000, 0x7ffeefbffa86, 0x1, 0x195d3d9, 0x4, 0x1, 0x0, 0x1998ef3, ...)
        /Users/henridf/work/brim/zq/zio/detector/file.go:131 +0xaf
github.com/brimsec/zq/zio/detector.OpenFile(0xc0002e2540, 0x7ffeefbffa86, 0x1, 0x195d3d9, 0x4, 0x1, 0x0, 0x1998ef3, 0x59, 0x0, ...)
[...]

PR #752, opened 2020-05-10
Use json typing info during format detection

henridf:
Previously, the throwaway ndjsonio reader used during detection was not configured any with json typing info that may have been provided (via the zqd API or zq -j).

This meant that format detection ndjsonio did not fully reflect actual reader behavior. 

This commit fixes that. It also removes an egregious manifestation of #747, which is that even post #748 , reading a typed json file with nesting would crash, because detection was not passing the typing scheme to the ndjsonio reader, and so that reader was using inference, which ran into #747.

Now type detection works with nested ndjson, as the change to nesting.yaml shows. #747 remains unfixed of course but is less visible pending a complete fix.

### DIFF
--- a/zio/detector/file.go
+++ b/zio/detector/file.go
@@ -129,7 +129,7 @@ func OpenFromNamedReadCloser(zctx *resolver.Context, rc io.ReadCloser, path stri
 	r := GzipReader(rc)
 	var zr zbuf.Reader
 	if cfg.Format == "" || cfg.Format == "auto" {
-		zr, err = NewReader(r, zctx, path, cfg)
+		zr, err = NewReaderWithConfig(r, zctx, path, cfg)
 	} else {
 		zr, err = LookupReader(r, zctx, cfg.Format)
 	}
--- a/zio/detector/file.go
+++ b/zio/detector/file.go
@@ -129,7 +129,7 @@ func OpenFromNamedReadCloser(zctx *resolver.Context, rc io.ReadCloser, path stri
 	r := GzipReader(rc)
 	var zr zbuf.Reader
 	if cfg.Format == "" || cfg.Format == "auto" {
-		zr, err = NewReader(r, zctx)
+		zr, err = NewReader(r, zctx, path, cfg)
 	} else {
 		zr, err = LookupReader(r, zctx, cfg.Format)
 	}
--- a/zio/detector/file.go
+++ b/zio/detector/file.go
@@ -5,7 +5,6 @@ import (
 	"io"
 	"net/url"
 	"os"
-	"regexp"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
@@ -131,16 +130,10 @@ func OpenFromNamedReadCloser(zctx *resolver.Context, rc io.ReadCloser, path stri
 	if cfg.Format == "" || cfg.Format == "auto" {
 		zr, err = NewReaderWithConfig(r, zctx, path, cfg)
 	} else {
-		zr, err = LookupReader(r, zctx, cfg.Format)
+		zr, err = lookupReader(r, zctx, path, cfg)
 	}
 	if err != nil {
 		return nil, err
-	}
-
-	if jr, ok := zr.(*ndjsonio.Reader); ok && cfg.JSONTypeConfig != nil {
-		if err = jsonConfig(cfg, jr, path); err != nil {
-			return nil, err
-		}
 	}
 
 	return zbuf.NewFile(zr, rc, path), nil
@@ -156,14 +149,4 @@ func OpenFiles(zctx *resolver.Context, paths ...string) (*zbuf.Combiner, error) 
 		readers = append(readers, reader)
 	}
 	return zbuf.NewCombiner(readers), nil
-}
-
-func jsonConfig(cfg OpenConfig, jr *ndjsonio.Reader, filename string) error {
-	var path string
-	re := regexp.MustCompile(cfg.JSONPathRegex)
-	match := re.FindStringSubmatch(filename)
-	if len(match) == 2 {
-		path = match[1]
-	}
-	return jr.ConfigureTypes(*cfg.JSONTypeConfig, path)
 }
--- a/zio/detector/reader.go
+++ b/zio/detector/reader.go
@@ -47,7 +47,7 @@ func NewReaderWithConfig(r io.Reader, zctx *resolver.Context, path string, cfg O
 		return nil, err
 	}
 	if cfg.JSONTypeConfig != nil {
-		if err = jsonConfig(cfg, nr, path); err != nil {
+		if err := jsonConfig(cfg, nr, path); err != nil {
 			return nil, err
 		}
 	}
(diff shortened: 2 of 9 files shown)


zio/detector/file.go
```diff
@@ -129,7 +129,7 @@ func OpenFromNamedReadCloser(zctx *resolver.Context, rc io.ReadCloser, path stri
 	r := GzipReader(rc)
 	var zr zbuf.Reader
 	if cfg.Format == "" || cfg.Format == "auto" {
-		zr, err = NewReader(r, zctx)
+		zr, err = NewReader(r, zctx, path, cfg)
```
mattnibs, line 132: I'm confused why this is needed when you do the same check that this does on line 140?

zio/detector/reader.go
```diff
@@ -14,7 +14,7 @@ import (
 	"github.com/brimsec/zq/zqe"
 )
 
-func NewReader(r io.Reader, zctx *resolver.Context) (zbuf.Reader, error) {
+func NewReader(r io.Reader, zctx *resolver.Context, path string, cfg OpenConfig) (zbuf.Reader, error) {
```
mattnibs, line 17: I would prefer not to add more arguments to this simple interface. If this is necessary I would prefer just adding a new function: `NewReaderWithConfig`.

Separately what if OpenConfig.Format is set to `zng`. Does it make sense for it to return a json reader here?

zqd/handlers_test.go
```diff
@@ -574,14 +574,14 @@ func TestPostNDJSONLogWarning(t *testing.T) {
...
+	assert.Regexp(t, ": line 1: descriptor not found", warn1.Warning)
 	assert.Regexp(t, ": line 2: incomplete descriptor", warn2.Warning)
 
 	status := payloads[len(payloads)-2].(*api.LogPostStatus)
 	expected := &api.LogPostStatus{
 		Type:         "LogPostStatus",
-		LogTotalSize: 134,
-		LogReadSize:  134,
+		LogTotalSize: 71,
+		LogReadSize:  71,
```
mattnibs, line 584: I'm curious why the size changed here and not on the other ndjson tests?

zio/detector/reader.go
mattnibs: Even better than adding `NewReadingWithConfig` would be to just do type inference in whatever method invokes this:

```
if ndr, ok := r.(*ndjsonio.Reader); ok {
    ndr.SetConfig(path, cfg)
}
```

The scope of package `detector` should really limited to detecting the type of unknown streams. We've been adding a lot of bloat to this (I'm certainly guilty of this).

henridf: > If this is necessary I would prefer just adding a new function: NewReaderWithConfig.

I was expecting that... sure I can use this color if you prefer.

henridf: > Even better than adding NewReadingWithConfig would be to just do type inference in whatever method invokes this:
> `...`

I don't understand. What you're suggesting seems like exactly what `OpenFromNamedReadCloser` is already doing, right after calling NewReader. If I'm misunderstanding this, could you please clarify?

zio/detector/file.go
henridf: The reader of line 140 is not the same as the throwaway reader that is created inside NewReader (as per the PR description). 

That said, maybe the `jsonConfig()` call of line 140 should also be done inside NewReader (or NewReaderWithConfig) rather than one here and one there.

zqd/handlers_test.go
henridf: Before, the bad record was read once in the throwaway ndjson reader using inference (which succeeded), and those bytes were counted. This was incorrect, and is no longer the case after this fix.

zio/detector/reader.go
henridf: > Separately what if OpenConfig.Format is set to zng. Does it make sense for it to return a json reader here?

It would not. But it would make even less sense to call this function with OpenConfig.Format that is set to a format. (Such an misplaced call to detector.NewReader() has always been possible, btw).

I could add a check for this, or change the interface to take a `*JSONTypeConfig` rather than the entire OpenConfig, if you like that better.

zio/detector/file.go
mattnibs: I see.

zqd/handlers_test.go
mattnibs: Ugh, I see.

zio/detector/reader.go
```diff
@@ -46,6 +46,11 @@ func NewReader(r io.Reader, zctx *resolver.Context) (zbuf.Reader, error) {
 	if err != nil {
 		return nil, err
 	}
+	if cfg.JSONTypeConfig != nil {
+		if err = jsonConfig(cfg, nr, path); err != nil {
```
nwt, line 50:
```suggestion
		if err := jsonConfig(cfg, nr, path); err != nil {
```

mattnibs:
@henridf forgive the naive question but I'm a little out of the loop with state-of-the-art json type inference functionality. Could the problem be stated like this: There are some ndjson streams that are valid json that the detector is say are NOT valid streams?

alfred-landrum:
My suggestions:
* change stringReader/internal.Run to more closely mimic the zq command "inputReaders" , so that it calls detetector.OpenFile . You can then unexport  detector.LookupReader.
* change ndjsonio.NewReader to take the path & type config parameters, and modify the now un-exported lookupReader to take the path & cfg arguments, like NewReaderWithConfig does.

henridf:
> Could the problem be stated like this: There are some ndjson streams that are valid json that the detector is say are NOT valid streams?

Not quite - the problem is that if the detector uses a differently-configured (throwaway) reader for trying out a format that the actual reader, then that mismatch can lead to invalid detections. 

In this specific case, there's a record that is valid "untyped ndjson" but is invalid "typed ndjson".

But a simpler formulation is just that

mattnibs:
> In this specific case, there's a record that is valid "untyped ndjson" but is invalid "typed ndjson".

Thanks. 

It seems like this in this case of you really have a warning for a bad record whereas the rest of the stream *could* be valid.

henridf:
> It seems like this in this case of you really have a warning for a bad record whereas the rest of the stream could be valid.

@mattnibs absolutely. In that regard it's identical to all the other text-based readers.
(if we could distinguish between syntax/parse errors vs "semantically incorrect" errors, then we could make format detection based on the former category only, and avoid the situation you describe).

henridf:
@alfred-landrum I implemented your suggestion from https://github.com/brimsec/zq/pull/752#issuecomment-626838140 in 353f72b.

One side-observation: seeing `ndjsonio.NewReader(..)` with its additional parameters (which are not always used), it seems like this change is going in the opposite direction from the one requested by @mattnibs in https://github.com/brimsec/zq/pull/752#discussion_r422679690.

zio/ndjsonio/reader.go
```diff
@@ -34,20 +35,29 @@ type Reader struct {
...
+	r := &Reader{
 		scanner: scanner,
 		stats:   ReadStats{Stats: &scanner.Stats, typeStats: &typeStats{}},
 		inf:     inferParser{zctx},
 		zctx:    zctx,
-	}, nil
+	}
+	if tc != nil {
+		var path string
+		re := regexp.MustCompile(JSONPathRegex)
```
alfred-landrum, line 53: This doesn't need to be a `MustCompile` anymore.

PR #754, opened 2020-05-11
Parquet reader improvements

aswan:
- Organize parquet->zng transcoding around the parquet types instead of
  zng types
- Skip over un-typed repeated elements for now
- Handle parquet timestamp times (not yet tested)
- Use original parquet column names for zng field names instead of
  names that have been mangled for go structures.


zio/parquetio/reader.go
```diff
@@ -42,91 +44,126 @@ func NewReader(f source.ParquetFile, zctx *resolver.Context) (*Reader, error) {
...
+	TimestampMilliseconds
+	TimestampMicroseconds
+	TimestampNanoseconds
+
+	// XXX INTERVAL
+	// XXX INT_*, UINT_* types
+
+	// composite types
+	List
+)
```
nwt, line 77: Don't export any of this.

aswan:
> I did some manual testing with this branch & reported a few issues where were addressed. There should be some unit tests for those cases.

For posterity, the bugs to be tested:
- zng array types created from the reader weren't getting type ids assigned.  A test for that entails reading a parquet file with a LIST in the schema.
- OpenParquet() wasn't passing the resolver.Context into the reader.  This can be tested by reading two parquet files with different schemas.

zio/parquetio/reader.go
```diff
@@ -197,7 +309,8 @@ func convertNestedElement(els []*parquet.SchemaElement, i int) (int, *parquetCol
 		return convertListType(els, i)
 	}
 
-	return 1, nil, fmt.Errorf("Cannot handle non-LIST nested element %s", el.Name)
+	return countChildren(els, i), nil, nil
+	// return 1, nil, fmt.Errorf("Cannot handle non-LIST nested element %s", el.Name)
```
alfred-landrum, line 313: comment still needed?

zio/parquetio/reader.go
```diff
@@ -173,6 +218,10 @@ func convertSchema(schema []*parquet.SchemaElement) ([]parquetColumn, error) {
 			continue
 		}
 
+		// XXX translate the column name
```
alfred-landrum, line 221: The below lines are performing an operation on the column name, would you describe what more needs to be done?

PR #761, opened 2020-05-11
Space Storage

mattnibs:
Make package Storage, responsible for managing all log data for a space.

Eventually there should be a central interface for Storage so that
a space can support multiple storage backends. For now this is limited
to a single ZngFile storage backend where:
- Data is populated by passing a zbuf.Reader into it.
- Data can only be rewritten. It cannot be updated.
- Data is housed in a single all.zng file.
- Concurrent reads are allowed but only one active write is allowed.
- Maintains an active cache of the total time span of all records stored.

Additional supporting changes:
- Reworked ingest.Log to instead return a LogIngestOp struct that exports
go channels for "Warnings" and "Done" notifications. Also includes
a Stats() function that provides a thread-safe snapshot of current ingest
stats.
- Reworked ingest.Pcap to use space.Storage for generating log snapshots.


zio/zngio/sort.go
```diff
@@ -0,0 +1,170 @@
+package zngio
```
alfred-landrum, line 1: Like we discussed in slack, this sort implementation will get removed, and would need to be in a separate PR where it would be considered to replace the existing logic in the sort proc.

zbuf/batch.go
```diff
@@ -48,3 +48,52 @@ func ReadBatch(zr Reader, n int) (Batch, error) {
 	}
 	return NewArray(recs, nano.NewSpanTs(minTs, maxTs)), nil
 }
+
+func ReadBatchSize(zr Reader, n int64) (Batch, error) {
```
alfred-landrum, line 52: This can be removed with the sort code.

zio/detector/reader.go
```diff
@@ -59,6 +60,52 @@ func NewReader(r io.Reader, zctx *resolver.Context) (zbuf.Reader, error) {
 	return nil, joinErrs([]error{tzngErr, zeekErr, ndjsonErr, zjsonErr, zngErr})
 }
 
+func Format(r io.Reader) (string, error) {
```
alfred-landrum, line 63: I don't see why you've duplicated all of the detection logic. This is only called from logOp.openFile; there should be a way for it to call detector.OpenFile.

zbuf/zng.go
```diff
@@ -97,3 +97,70 @@ func CopyWithContext(ctx context.Context, dst WriteFlusher, src Reader) error {
...
+	return nil, nil
+}
+
+func MultiReader(readers ...Reader) Reader {
+	r := make([]Reader, len(readers))
+	copy(r, readers)
+	return &multiReader{readers: r}
+}
+
+type namedReader struct {
```
alfred-landrum, line 155: I don't think this is worth putting in the zbuf package. It's called once from LogOp.openFile. I don't understand how it can be used since NamedReader returns a Reader interface which doesn't expose String()?

zqd/ingest/log.go
```diff
@@ -5,207 +5,138 @@ import (
...
-	}
-	return pipe.Send(&api.LogPostStatus{
-		Type:         "LogPostStatus",
-		LogTotalSize: d.totalSize,
-		LogReadSize:  logBytesRead(),
-	})
+	// Wrap zbuf.Reader in NamedReader (to make file easier to
+	// identify for errors) and WarningReader (so Read warnings propogate to
+	// warning channel).
+	zr = zbuf.NamedReader(zr, path)
```
alfred-landrum, line 97: I don't see how NamedReader's String() method could be called.

zqd/ingest/log.go
```diff
@@ -5,207 +5,138 @@ import (
...
-			n += readCounters[i].nread
-		}
-		return n
-	}
+func (p *LogOp) openWarning(path string, err error) {
+	p.warnings = append(p.warnings, fmt.Sprintf("%s: %s", path, err))
+}
 
-	zngfile, err := s.CreateFile(filepath.Join(tmpIngestDir, allZngTmpFile))
+func (p *LogOp) openFile(zctx *resolver.Context, path string, cfg detector.OpenConfig) error {
```
alfred-landrum, line 62: This function does too much, and all of the reader/writer plumbing that's going on is hard to follow. I think the former setup was easier to follow: loop over req.path, and do 3 separate things: open the file, calling the detector package to get a zbuf.Reader from it, then add the reader & info to the readers state.

zqd/storage/zng.go
```diff
@@ -0,0 +1,148 @@
...
+	if first {
+		w.span = s
+	} else {
+		w.span = w.span.Union(s)
+	}
+	return nil
+}
+
+// XXX only one writeto allowed at a time
+func (s *ZngStorage) WriteTo(ctx context.Context, zr zbuf.Reader) error {
```
alfred-landrum, line 84: As I suggested previously, renaming this to "Replace" would better communicate what it currently does.

zqd/space/space.go
```diff
@@ -44,26 +40,25 @@ func newSpaceID() api.SpaceID {
 }
 
 type Space struct {
+	Storage *storage.ZngStorage
+
 	path string
 	conf config
-
-	index *zngio.TimeIndex
+	mu   sync.RWMutex
```
alfred-landrum, line 47: Please remove the opMutex, deletePending, and other associated changes, if needed, those can happen in a separate PR.

zio/detector/reader.go
mattnibs: This is the essence of detector should be about. Aside from this its a little complicated why I did this: In log ingest endopoint, if a file was getting rejected by detector.NewReader, the bytes read would get counted towards the total of bytes read in the log ingest op. To me the simplest solution was to use this function to detetect the stream type then set the underlying file seek to `SeekStart` and pass the reset file into `LookupReader`.

zqd/ingest/log.go
mattnibs: If it's the last reader wrapped then it works. Warning reader should happen before this.

zqd/storage/zng.go
mattnibs: Yep.

zqd/space/space.go
mattnibs: Sure.

zqd/ingest/log.go
mattnibs: I agree that this function could be broken up. OTH IMHO the way it was was harder to follow. Instead you had the `ingestLogs` function that was doing too much.

zqd/ingest/log.go
mattnibs: I was trying to model the wonder flexibility of io package- the ability to wrap and compose interfaces.

zqd/space/space.go
mattnibs: Not needed, but simpler with less moving parts.

zio/detector/reader.go
mattnibs: Dropped it and went back to readCounter.

zqd/ingest/log.go
mattnibs: Got rid of it.

zqd/ingest/log.go
mattnibs: Reverted.

zbuf/zng.go
mattnibs: Removed.

mattnibs:
The operations on ZngStorage.Span() are all not thread safe and they should be. I do not think they were beforehand. In the interest of blocking other work I thought I'd put this up, but am willing to fix that here if people think I should.

zqd/ingest/log_test.go
```diff
@@ -33,6 +34,7 @@ func writeTempFile(t *testing.T, data string) string {
 }
 
 func TestLogsErrInFlight(t *testing.T) {
+	t.Skip("Flaky, racy tests. Redo this.")
```
mattnibs, line 37: While this was passing every time on my macbook pro, on github actions this reliably fails. This used to normally pass but in this change we don't know about concurrent writes until inget.LogIngestOp calls storage.Replace.

Writing a reliable test case for this is going to require more plumbing.

zqd/api/api.go
```diff
@@ -109,7 +109,6 @@ type PcapPostStatus struct {
 
 type LogPostRequest struct {
 	Paths          []string             `json:"paths"`
-	StopErr        bool                 `json:"stop_err"`
```
alfred-landrum, line 112: Removing this option from the api can happen in separate PR.

zqd/ingest/log.go
```diff
@@ -3,79 +3,86 @@ package ingest
...
 // Like ingest.Pcap, this overwrites any existing data in the space.
-func Logs(ctx context.Context, pipe *api.JSONPipe, s *space.Space, req api.LogPostRequest) error {
-	ingestDir := s.DataPath(tmpIngestDir)
-	if err := os.Mkdir(ingestDir, 0700); err != nil {
-		// could be in use by pcap or log ingest
-		if os.IsExist(err) {
-			return ErrIngestProcessInFlight
-		}
-		return err
+func NewLogOp(ctx context.Context, store *storage.ZngStorage, req api.LogPostRequest) (*LogOp, error) {
```
alfred-landrum, line 35: Interactions with LogOp have 3 different ways to give feedback/warnings/errors:
- an error from NewLogOp
- warnings sent back via channel
- final error sent when done channel closed

Since all the warnings are generated at setup time in NewLogOp, consider changing NewLogOp to:
```go
func NewLogOp(ctx context.Context, store *storage.ZngStorage, req api.LogPostRequest) (op *LogOp, warnings []error, err error)
```
and send all the warnings immediately after the call; that will remove the warnings channel & associated code.

You can also change the done channel to a `chan error`, so the select on it can be:
```go
case err := <-Done()
```

zqd/storage/zng.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+	defer s.wsem.Release(1)
+	if err := os.Remove(s.join(allZngFile)); err != nil && !os.IsNotExist(err) {
+		return err
+	}
+	return s.UnsetSpan()
+}
+
+// XXX This is not thread safe and it should be.
+func (s *ZngStorage) SetSpan(span nano.Span) error {
```
alfred-landrum, line 163: The behavior for this function doesn't match the name, and in pcap ingest, and this behavior requires the `UnsetSpan` , `SetSpan` sequence during pcap ingest. Can the unioning be removed so this function just sets the space's span?

zqd/storage/zng.go
```diff
@@ -0,0 +1,246 @@
...
+		f, err = os.Open(s.join(bzngFile))
+		if err != nil {
+			if !os.IsNotExist(err) {
+				return nil, err
+			}
+			r := zngio.NewReader(strings.NewReader(""), zctx)
+			return zbuf.NopReadCloser(r), nil
+		}
+	}
+	// XXX plz keep track of how long this takes
```
alfred-landrum, line 80: You can remove this comment.

zqd/api/api.go
mattnibs: ok

zqd/ingest/log.go
mattnibs: @alfred-landrum warnings can also be generated if a reader comes across an invalid record while reading, or runs into a spontaneous error: see line 58

zqd/storage/zng.go
mattnibs: The unioning is needed because otherwise spans from the pcap ingests snapshots would overwrite the the full span size we get from generating the pcap index. I think the best long term solution would be to have space.Span() take the union of PcapStorage.Span and LogStorage.Span, but I put this in for now to arrive at feature parity. I should add a comment mentioning this here.

zqd/ingest/log_test.go
```diff
@@ -42,30 +43,18 @@ func TestLogsErrInFlight(t *testing.T) {
...
-		return
-	}
-	assert.Fail(t, "expected only one error")
+	op1, err1 := ingest.NewLogOp(ctx, s.Storage, api.LogPostRequest{Paths: []string{f}})
+	op2, err2 := ingest.NewLogOp(ctx, s.Storage, api.LogPostRequest{Paths: []string{f}})
+	require.NoError(t, err1)
+	require.NoError(t, err2)
+	assert.NoError(t, op1.Error())
+	require.Error(t, op2.Error())
+	assert.Contains(t, op2.Error().Error(), storage.ErrWriteInProgress.Error())
```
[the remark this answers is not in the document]
mattnibs: LogOp doesn't close all goroutines when done- I don't believe this is necessary either.

zqd/ingest/log.go
mattnibs: w.r.t making Done and `chan error`, the down side to this approach is that only one subscriber will know if an error occurred. If I make the channel a buffered channel there won't even need to be a subscriber in order for the error to be received by consumers.

I suspect this is why the authors of package context took this approach.

zqd/ingest/log_test.go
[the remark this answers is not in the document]
mattnibs: That said perhaps closing the goroutines is the most hygienic thing to do (I'm a little fuzzy on the idiomatic go approach here)? Either way I've done away with this test and put a similar test in zqd/storage- can add a test for your case if you want.

zqd/ingest/log.go
alfred-landrum: ok, how about removing the error return from NewLogOp entirely, and having it use a single channel to return warnings on, that when closed, signals the operation is done? So inside log:

```go
  op := NewLogOp(ctx, store, req)
  ...
  select {
  case warning, ok := <- op.Status()
  if !ok {
     // op has finished, get final error if any
     err := op.Error()
     ...
     return
  }
  // Otherwise, send warning:
  err := pipe.Send(api.LogPostWarning{...})
  ...
  }
```

That would reduce the op to use a single channel, and remove the extra code to handle storing & draining the warnings.

zqd/storage/zng_test.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+type waitReader time.Duration
+
+func (w waitReader) Read() (*zng.Record, error) {
+	time.Sleep(time.Duration(w))
+	return nil, errors.New("time out")
+}
+
+func TestFailOnConcurrentWrites(t *testing.T) {
```
alfred-landrum, line 23: 👍

zqd/storage/zng.go
alfred-landrum: Then for now, I suggest you change the name from `SetSpan` to `UnionSpan`, as the behavior is surprising given the name.

zqd/ingest/log.go
mattnibs: Yep, I really like that idea.

zqd/storage/zng.go
mattnibs: Agreed, much clearer.

mattnibs:
> Would you please create a draft Brim PR using this branch's head, so that we can verify that the Brim integration tests pass with this change?

@alfred-landrum https://github.com/brimsec/brim/pull/772

PR #762, opened 2020-05-12
Introduce record visitor and use it in zar TypeIndexer and zng TypeCheck

### LINKED ISSUE #641: archive TypeIndexer should use a record visitor
Like the one added in:
https://github.com/brimsec/zq/pull/609

henridf:
This change adds a method `(*zng.Record).Walk(v RecordVisitor)` that can be used to visit all fields in a record in DFS order. Two places that were previously rolling their own traversal have been moved over: zar's TypeIndexer and zng's TypeCheck. (If there's any others I've missed please let me know and I can move them in this PR).

The relevant comment on how visiting works works is here:
```go
// A RecordVisitor is called for each value in a record encountered by
// Walk. Upon visiting a container value, the boolean returned by the
// RecordVisitor determines whether the values of that
// container value will be individually visited.
// If the visitor returns an error, the walk stops and that error will be
// returned to the caller of Walk().
type RecordVisitor func(typ Type, body zcode.Bytes) (bool, error)
```

closes #641


henridf:
(I will add some unit tests before merging, but wanted to get a review pulse first).

henridf:
@mccanne If I'm not missing anything, I think this change should set things up to support indexing on container types per the discussion in https://github.com/brimsec/zq/pull/600#issuecomment-616512917 ...

zng/walk.go
```diff
@@ -0,0 +1,288 @@
...
+			enter, err := rv(t, body)
+			if err != nil {
+				return err
+			}
+			if !enter {
+				continue
+			}
+			if err := walkUnion(t, body, rv); err != nil {
+				return err
+			}
```
nwt, line 84: These four cases are identical except for the function called, so you can probably refactor this so you're not repeating yourself.

henridf: Great point. In fact there's a some repetition throughout these record traversal funcs, as there was in the TypeCheck and TypeIndexer that this ressembles. The good news is this repetition now only occurs once.

To this specific instance, the issue is that each function called has a different signature. So you can't pass it to a factored out case function, as far as I can tell. Specifically, I'd like to write a handler like:

```go
process := func(t Type, walker func(Type, zcode.Bytes, RecordVisitor) error) (bool, error) {
..
}
```

and then call it as:
```go
		case *TypeRecord:
			if err := process(t, walkRecord); err != nil {
				return err
			}
```

but this is a type error. Am I missing a better way to refactor ? Fwiw, the above refactor would save about 15 lines, if it worked.

mattnibs:
I personally prefer the SkipErr pattern for walk seen in stdlib filepath.Walk (and also adherence to convention in stdlib): https://golang.org/pkg/path/filepath/#WalkFunc

That said I believe @mccanne might say chocolate vs vanilla.

henridf:
@mattnibs that seems like a good idiom.... I can switch to it.

PR #768, opened 2020-05-13
Lenient parsing of json numbers

### LINKED ISSUE #766: Support scientific notation for integer types
Consider the following NDJSON object:

```
$ cat int64-sci-notation.ndjson
{"_path":"dns","datetime":1.521835103E9}
```

With default type inference, `zq` (commit `75fe8b9`) treat `datetime` as a `float64`.

```
$ zq -t int64-sci-notation.ndjson 
#0:record[_path:string,datetime:float64]
0:[dns;1521835103;]
```

However, it fails if I try to apply a JSON typing schema that tries to to treat it as an `int64`.

```
$ cat int64-sci-notation-types.json
{
  "descriptors": {
    "dns_log": [
      {
        "name": "_path",
        "type": "string"
      },
      {
        "name": "datetime",
        "type": "int64"
      }
    ]
  },
  "rules": [
    {
      "descriptor": "dns_log",
      "name": "_path",
      "value": "dns"
    }
  ]
}

$ zq -t -j int64-sci-notation-types.json int64-sci-notation.ndjson
int64-sci-notation.ndjson: line 1: bad format
```

However, if you look closely at the value, you can see that if you expand the scientific notation, it's a whole number.

FWIW, it was in Nifi that I bumped into values represented this way.

henridf:
Json has a single numeric type, so we should be a bit more lenient when parsing numbers into the multitude of zng numeric types.

closes #766

### DIFF
--- a/zio/ndjsonio/ndjson_test.go
+++ b/zio/ndjsonio/ndjson_test.go
@@ -200,6 +200,20 @@ func TestNewRawFromJSON(t *testing.T) {
 			json:        `{"_path": "test", "ts":1573860644637}`,
 			defaultPath: "inferred",
 		},
+		{
+			name: "uint64 in scientific notation",
+			tzng: `#0:record[_path:string,datetime:uint64]
+0:[test;1521835103;]`,
+			json:        `{"_path": "test", "datetime":1.521835103E9}`,
+			defaultPath: "inferred",
+		},
+		{
+			name: "int64 in scientific notation",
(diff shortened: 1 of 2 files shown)


henridf:
> I don't think it makes a difference, but the zng float64 type's Parse uses byteconv.ParseFloat64 (from pkg/byteconf) to wrap strconv.ParseFloat.

Actually it does! As byteconv uses copyless/unsafe string conversion. Switched to that.

PR #774, opened 2020-05-13
Add parquet to help for -i arg

### LINKED ISSUE #769: Mention Parquet among supported input formats
`zq` as of commit `64d5e4a` includes support for reading input files of Parquet format, but it's not yet mentioned in the supported list of formats in `zq help`.

```
$ zq help 2>&1 | grep "input data"
    -i format of input data [auto,zng,ndjson,zeek,zjson,tzng] (default "auto")
```

I checked with @aswan and he confirmed this was an oversight, so opening this as a reminder to circle back and add it.

aswan:
Closes #769

### DIFF
--- a/zio/detector/file.go
+++ b/zio/detector/file.go
@@ -17,7 +17,7 @@ import (
 	"github.com/brimsec/zq/zio/parquetio"
 	"github.com/brimsec/zq/zng/resolver"
 
-	_ "github.com/xitongsys/parquet-go-source/local"
+	"github.com/xitongsys/parquet-go-source/local"
 )
 
 type OpenConfig struct {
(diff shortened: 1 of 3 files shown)


alfred-landrum:
I think we should add support for auto-detecting parquet like we do for the other input formats before we expose it in the help text.

aswan:
> I think we should add support for auto-detecting parquet like we do for the other input formats before we expose it in the help text.

Auto-detecting parquet will be tricky, the current auto-detection code is built around reading files sequentially from the beginning.  But parquet needs to do random-access reads.  In particular, the main file metadata block is at the very end of the file.  This is all doable but I think it would basically entail a rewrite of the detection logic.

alfred-landrum:
> Auto-detecting parquet will be tricky, the current auto-detection code is built around reading files sequentially from the beginning. But parquet needs to do random-access reads. In particular, the main file metadata block is at the very end of the file. This is all doable but I think it would basically entail a rewrite of the detection logic.

I slacked with andrew, and suggested adding a change so that `parquet: auto-detection not supported` shows up in the list of errors when auto-detection is used & fails.

PR #775, opened 2020-05-13
Translate "\r\n" line endings to "\n" in skim.Scanner

nwt:
This is motivated by the "\r\n" line endings in the logs currently generated by MinGW Zeek. That behavior might change, but this is probably a good idea regardless.

### DIFF
--- a/pkg/skim/skim_test.go
+++ b/pkg/skim/skim_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 
 	"github.com/brimsec/zq/pkg/skim"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/skim/skim_test.go
```diff
@@ -71,3 +72,14 @@ func TestSkimNoNewLine(t *testing.T) {
 	}
 	require.Equal(t, []byte(nil), line)
 }
+
+func TestSkimCarriageReturn(t *testing.T) {
+	data := []byte("line1\n\r\nline2\r\n\r\n\rline3\rline3\nline4\r")
+	buf := make([]byte, ReadSize)
+	scanner := skim.NewScanner(bytes.NewReader(data), buf, MaxLineSize)
+	for _, s := range []string{"line1\n", "line2\n", "\rline3\rline3\n", "line4\r"} {
```
alfred-landrum, line 80: 👍

nwt:
I forgot to mention that I added zio/zeekio.TestReaderCRLF to make anyone think twice before removing the skim.Scanner change.

PR #776, opened 2020-05-14
ndjsonio: get rid of hostile "bad format" error message

### LINKED ISSUE #760: More helpful ndjsonio parsing errors
@philrz ran into an issue with typed ndjson where a field that was declared as `uint64` in types.json, but appeared in a json record with a non-integer value of `1.521835102647908E12`. 

This failed with a quasi-hostile error message of `line 1: bad format`. This needs to be better.

henridf:
Error messages related to invalid user input (data, queries, config) should be helpful. This is one small change towards that goal.

closes #760

### DIFF
--- a/zio/ndjsonio/typeparser.go
+++ b/zio/ndjsonio/typeparser.go
@@ -35,7 +35,6 @@ type typeParser struct {
 var (
 	ErrDescriptorNotFound   = errors.New("descriptor not found")
 	ErrMissingPath          = errors.New("missing path field")
-	ErrBadFormat            = errors.New("bad format")
 	ErrIncompleteDescriptor = errors.New("incomplete descriptor")
 )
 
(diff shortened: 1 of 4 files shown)


zio/ndjsonio/typeparser.go
```diff
@@ -118,16 +117,16 @@ func (info *typeInfo) makeViews(data []byte) (int, error) {
...
 		case jsonparser.Array:
 			builder.BeginContainer()
-			ztyp := zng.InnerType(typ)
+			ztyp := zng.InnerType(col.Type)
 			var iterErr error
 			callback := func(v []byte, typ jsonparser.ValueType, offset int, _ error) {
 				zv, err := parseSimpleType(v, ztyp)
 				if err != nil {
-					iterErr = err
+					iterErr = fmt.Errorf("field \"%s\" (type %s): %w", col.Name, typ, err)
```
mattnibs, line 129: Since this error happens twice what do you think about making an exported error? Something like this:
```
type FieldError struct {
   Field string
   Type zng.Type
   Err error
}
```
Might be nice to start doing stuff like this so we could one day separate the warnings from the actual errors using `package` errors reflection...

henridf: I'm all for structured errors and agree we have a lot to do there, but I'd prefer to do this when we actually get to using this warning vs error distinction somewhere (or btw also distinguishing different types of errors, such as syntax/parse error vs semantic-y errors, which is a relevant distinction for detection).

PR #777, opened 2020-05-14
Stdin error filename

henridf:
It's a little odd when zq produces error messages starting with `-:` like:

```
$ cat bad.tzng | zq -t - 
-: format detection error
   [...]
```

This PR fixes things so that those messages now start with `stdin:`, like they would with other file names.

### DIFF
--- a/cmd/zq/zq.go
+++ b/cmd/zq/zq.go
@@ -259,6 +259,9 @@ func (c *Command) inputReaders(paths []string) ([]zbuf.Reader, error) {
 	var readers []zbuf.Reader
 	for _, path := range paths {
 		file, err := detector.OpenFile(c.zctx, path, cfg)
+		if path == "-" {
+			path = "stdin"
+		}
 		if err != nil {
 			err = fmt.Errorf("%s: %w", path, err)
 			if c.stopErr {
--- a/cmd/zq/zq.go
+++ b/cmd/zq/zq.go
@@ -252,16 +252,15 @@ func (c *Command) errorf(format string, args ...interface{}) {
 func (c *Command) inputReaders(paths []string) ([]zbuf.Reader, error) {
 	cfg := detector.OpenConfig{
 		Format:         c.ReaderFlags.Format,
-		DashStdin:      true,
 		JSONTypeConfig: c.jsonTypeConfig,
 		JSONPathRegex:  c.jsonPathRegexp,
 	}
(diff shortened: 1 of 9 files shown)


cmd/zq/zq.go
```diff
@@ -268,6 +259,9 @@ func (c *Command) inputReaders(paths []string) ([]zbuf.Reader, error) {
 	var readers []zbuf.Reader
 	for _, path := range paths {
 		file, err := detector.OpenFile(c.zctx, path, cfg)
+		if path == "-" {
+			path = "stdin"
```
henridf, line 263: this is a bit clunky and can go away with #778

zio/detector/file.go
```diff
@@ -51,6 +51,7 @@ func OpenFile(zctx *resolver.Context, path string, cfg OpenConfig) (*zbuf.File,
 	var f *os.File
 	if cfg.DashStdin && path == "-" {
 		f = os.Stdin
+		path = "stdin"
```
henridf, line 54: this is a bit clunky and can go away with #778

cmd/zq/zq.go
henridf: #778 has been merged into this PR

zio/detector/file.go
henridf: #778 has been merged into this PR

zio/detector/file.go
```diff
@@ -49,7 +48,7 @@ func OpenFile(zctx *resolver.Context, path string, cfg OpenConfig) (*zbuf.File,
 	}
 
 	var f *os.File
-	if cfg.DashStdin && path == "-" {
+	if path == "/dev/stdin" {
```
mattnibs, line 51: Zap just does "stdin" for this case- I don't think `/dev/stdin` makes sense for windows right?

henridf: This is just an internal string. (I'll define it as a constant somewhere). The only requirement is that it shouldn't conflict with an actual file name. Are you concerned that we could have a file called "/dev/stdin" on Windows?

mattnibs: No, this is fine.

mattnibs: I might consider adding this as an export const `StdinPath` or something and using this in `zq/zq/cmd.go`

henridf: indeed - done

PR #780, opened 2020-05-14
Add support for reading parquets from S3

henridf:
(minio test coming; put this up so @philrz can take a look)


philrz:
Seems to work well. I've successfully read all the sample Parquet data I have on hand from an S3 bucket using the `zq` on this branch.

go.sum
```diff
@@ -1,8 +1,28 @@
 cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
 cloud.google.com/go v0.34.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
+cloud.google.com/go v0.38.0/go.mod h1:990N+gfupTy94rShfmMCWGDn0LpTmnzTp2qbd1dvSRU=
```
mattnibs, line 3: 178+ lines of dependencies... jeez. This is starting to get npm-like.

henridf: image

zio/detector/file.go
```diff
@@ -40,14 +43,14 @@ func IsS3Path(path string) bool {
 // URL. If the path is neither of these or can't otherwise be opened,
 // an error is returned.
 func OpenFile(zctx *resolver.Context, path string, cfg OpenConfig) (*zbuf.File, error) {
-	if IsS3Path(path) {
-		return OpenS3File(zctx, path, cfg)
-	}
-
 	if cfg.Format == "parquet" {
```
mattnibs, line 46: Consider adding a comment here. Something like:
```suggestion
        // Parquest is special and needs its own reader for s3 sources- therefore this must go before
        // the IsS3Path check.
	if cfg.Format == "parquet" {
```

go.sum
mattnibs: indeed

mccanne:
looks great

PR #781, opened 2020-05-14
Fix race in storage concurrent write tests

mattnibs:
Move WaitGroup into reader to insure that thread is running the
Write operation before issuing the second write operation.

Fixes failure seen here: https://github.com/brimsec/zq/runs/673503168#step:4:526

### DIFF
--- a/zqd/storage/zng_test.go
+++ b/zqd/storage/zng_test.go
@@ -13,10 +13,14 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-type waitReader time.Duration
+type waitReader struct {
+	sync.WaitGroup
+	dur time.Duration
+}
 
-func (w waitReader) Read() (*zng.Record, error) {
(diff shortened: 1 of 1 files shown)


zqd/storage/zng_test.go
```diff
@@ -13,10 +13,14 @@ import (
...
+type waitReader struct {
+	sync.WaitGroup
+	dur time.Duration
+}
 
-func (w waitReader) Read() (*zng.Record, error) {
-	time.Sleep(time.Duration(w))
+func (w *waitReader) Read() (*zng.Record, error) {
+	w.Done()
+	time.Sleep(time.Duration(w.dur))
```
nwt, line 23:
```suggestion
	time.Sleep(w.dur)
```

mattnibs: 👍

PR #782, opened 2020-05-14
Batch parquet reads

aswan:
The parquet-go library has a bunch of overhead to dispatch tasks to
worker goroutines and collect the results.  When we read records one
at a time, this was a major performance bottleneck.  By reading 1000
records at a time, this overhead gets amortized across multiple rows
and we get a roughly 4x speedup in read performance.

### DIFF
--- a/zio/parquetio/reader.go
+++ b/zio/parquetio/reader.go
@@ -14,11 +14,14 @@ import (
 	"github.com/xitongsys/parquet-go/source"
 )
 
+const BUFSIZE = 1000
+
 type Reader struct {
 	pr      *reader.ParquetReader
 	typ     *zng.TypeRecord
 	columns []parquetColumn
 	record  int
+	buf     []interface{}
 }
 
 func NewReader(f source.ParquetFile, zctx *resolver.Context) (*Reader, error) {
@@ -41,7 +44,7 @@ func NewReader(f source.ParquetFile, zctx *resolver.Context) (*Reader, error) {
 		return nil, err
 	}
 
-	return &Reader{pr, typ, cols, 0}, nil
+	return &Reader{pr, typ, cols, 0, nil}, nil
 }
 
 type HandledType int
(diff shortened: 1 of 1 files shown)


zio/parquetio/reader.go
```diff
@@ -360,18 +363,28 @@ func convertListType(els []*parquet.SchemaElement, i int) (int, *parquetColumn,
 }
 
 func (r *Reader) Read() (*zng.Record, error) {
-	if r.record == int(r.pr.GetNumRows()) {
-		return nil, nil
-	}
-	r.record++
+	var err error
+	if r.buf == nil || len(r.buf) == 0 {
```
nwt, line 367:
```suggestion
	if len(r.buf) == 0 {
```
Because `len([]interface{}(nil)) == 0`.

zio/parquetio/reader.go
```diff
@@ -14,11 +14,14 @@ import (
 	"github.com/xitongsys/parquet-go/source"
 )
 
+const BUFSIZE = 1000
```
nwt, line 17:
```suggestion
const bufsize = 1000
```
No need to export this, but if you do, call it Bufsize or BufSize.

zio/parquetio/reader.go
```diff
@@ -360,18 +363,28 @@ func convertListType(els []*parquet.SchemaElement, i int) (int, *parquetColumn,
 }
 
 func (r *Reader) Read() (*zng.Record, error) {
-	if r.record == int(r.pr.GetNumRows()) {
-		return nil, nil
-	}
-	r.record++
+	var err error
```
nwt, line 366: Nit: Move down so it's adjacent to `r.buf, err = r.pr.ReadByNumber(toread)`.

aswan: this also gets used down on line 403, we can create another local there?  i assume they compile to the same thing, neither one seems dramatically better to me, do you prefer the two separate instances?

nwt: Doh! Missed the other use. The generated code will almost certainly be identical. I like the two locals approach, but what's here is also totally fine.

PR #787, opened 2020-05-15
zar archives are created with zar import

alfred-landrum:
Rename `zar chop` to `zar import`, and make it create a `zar.json` configuration file for the archive. Modify the zar commands to load that into a `struct Archive`.

### DIFF
--- a/archive/schema.go
+++ b/archive/schema.go
@@ -23,25 +23,34 @@ type Config struct {
 	LogSizeThreshold int `json:"log_size_threshold"`
 }
 
-func (c *Config) Write(path string) error {
-	b, err := json.Marshal(c)
+func writeTempFile(dir, pattern string, b []byte) (name string, err error) {
+	f, err := ioutil.TempFile(dir, pattern)
 	if err != nil {
-		return err
-	}
-	f, err := ioutil.TempFile(filepath.Dir(path), "."+configName+".*")
-	if err != nil {
-		return err
+		return "", err
 	}
 	_, err = f.Write(b)
 	if err != nil {
 		f.Close()
-		return err
+		os.Remove(f.Name())
+		return "", err
 	}
-	err = f.Close()
+	return f.Name(), f.Close()
+}
+
+func (c *Config) Write(path string) (err error) {
+	b, err := json.Marshal(c)
 	if err != nil {
 		return err
 	}
-	return os.Rename(f.Name(), path)
+	tmp, err := writeTempFile(filepath.Dir(path), "."+configName+".*", b)
+	if err != nil {
+		return err
+	}
+	err = os.Rename(tmp, path)
+	if err != nil {
+		os.Remove(tmp)
+	}
+	return err
 }
 
 func ConfigRead(path string) (*Config, error) {
@@ -50,16 +59,8 @@ func ConfigRead(path string) (*Config, error) {
 		return nil, err
 	}
 	defer f.Close()
-	b, err := ioutil.ReadAll(f)
-	if err != nil {
-		return nil, err
-	}
 	var c Config
-	err = json.Unmarshal(b, &c)
-	if err != nil {
-		return nil, err
-	}
-	return &c, nil
+	return &c, json.NewDecoder(f).Decode(&c)
 }
 
 type CreateFlags struct {
@@ -91,21 +92,12 @@ type Archive struct {
 
 func OpenArchive(path string) (*Archive, error) {
 	if path == "" {
-		return nil, errors.New("no ZAR_ROOT directory specified")
+		return nil, errors.New("no archive directory specified")
 	}
-	info, err := os.Stat(path)
-	if err != nil {
-		return nil, err
-	}
-	if !info.IsDir() {
-		return nil, errors.New("ZAR_ROOT not a directory")
-	}
-
 	c, err := ConfigRead(filepath.Join(path, configName))
 	if err != nil {
 		return nil, err
 	}
-
 	return &Archive{
 		Config: c,
 		Root:   path,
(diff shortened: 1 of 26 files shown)


archive/schema.go
```diff
@@ -0,0 +1,140 @@
...
+	if err != nil {
+		return err
+	}
+	f, err := ioutil.TempFile(filepath.Dir(path), "."+configName+".*")
+	if err != nil {
+		return err
+	}
+	_, err = f.Write(b)
+	if err != nil {
+		f.Close()
```
mattnibs, line 37: Being not quite sure myself how os' handle temp files, do you not need to delete the tempfile as well if there is an error?

archive/schema.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+type Archive struct {
+	Config *Config
+	Root   string
+}
+
+func OpenArchive(path string) (*Archive, error) {
+	if path == "" {
+		return nil, errors.New("no ZAR_ROOT directory specified")
```
mattnibs, line 94: Feels like this error should just say "no root directory specified"- the ZAR_ROOT is the domain of the cmd that calls this library.

archive/schema.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	info, err := os.Stat(path)
+	if err != nil {
+		return nil, err
+	}
+	if !info.IsDir() {
+		return nil, errors.New("ZAR_ROOT not a directory")
```
mattnibs, line 101: See above comment about ZAR_ROOT.

archive/schema.go
```diff
@@ -0,0 +1,140 @@
...
+func (c *Config) Write(path string) error {
+	b, err := json.Marshal(c)
+	if err != nil {
+		return err
+	}
+	f, err := ioutil.TempFile(filepath.Dir(path), "."+configName+".*")
+	if err != nil {
+		return err
+	}
+	_, err = f.Write(b)
```
mattnibs, line 35: Though not likely isn't this possible that the file doesn't write the whole buffer to the file. Consider doing `json.NewEncoder(f).Encode(c)`.

archive/schema.go
```diff
@@ -0,0 +1,140 @@
...
+		return err
+	}
+	err = f.Close()
+	if err != nil {
+		return err
+	}
+	return os.Rename(f.Name(), path)
+}
+
+func ConfigRead(path string) (*Config, error) {
```
mattnibs, line 47: We do this a lot. At some point (not here) we should just create a package- json file (I think we had this in boom).

archive/schema.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	return &c, nil
+}
+
+type CreateFlags struct {
+	megaThresh int
+	byteThresh int
+}
+
+func (f *CreateFlags) SetFlags(fs *flag.FlagSet) {
```
mattnibs, line 70: I am skeptical of this pattern of specifying flags in library packages. It is nice to look and the command definition and see all flags specified whereas this pattern obscures them. Additionally it seems like there is a high likelihood of flag name collisions.

archive/schema.go
nwt: https://pkg.go.dev/io?tab=doc#Writer says you'll get an error if the write is short, but regardless @mattnibs's suggestion is shorter than json.Marshal plus f.Write.

archive/schema.go
nwt: @mattnibs is right; we should close and then remove the temporary file on any error.

archive/schema.go
```diff
@@ -0,0 +1,140 @@
...
+	if err != nil {
+		return err
+	}
+	_, err = f.Write(b)
+	if err != nil {
+		f.Close()
+		return err
+	}
+	err = f.Close()
+	if err != nil {
```
nwt, line 41:
```suggestion
	if err := f.Close(); err != nil {
```

archive/schema.go
```diff
@@ -0,0 +1,140 @@
...
+	b, err := ioutil.ReadAll(f)
+	if err != nil {
+		return nil, err
+	}
+	var c Config
+	err = json.Unmarshal(b, &c)
+	if err != nil {
+		return nil, err
+	}
+	return &c, nil
```
nwt, line 62:
```suggestion
	var c Config
	return &c, json.NewDecoder(f).Decode(&c)
```

archive/schema.go
nwt: I'm not a fan either.

archive/schema.go
```diff
@@ -0,0 +1,140 @@
...
+		}
+	} else if !info.IsDir() {
+		return nil, errors.New("ZAR_ROOT not a directory")
+	}
+
+	cfgpath := filepath.Join(path, configName)
+	if _, err := os.Stat(cfgpath); err != nil {
+		if os.IsNotExist(err) {
+			err = f.Config().Write(cfgpath)
+			if err != nil {
```
nwt, line 134:
```suggestion
			if err := f.Config().Write(cfgpath); err != nil {
```

archive/schema.go
```diff
@@ -0,0 +1,140 @@
...
+	if info, err := os.Stat(path); err != nil {
+		if os.IsNotExist(err) {
+			err = os.MkdirAll(path, 0700)
+			if err != nil {
+				return nil, err
+			}
+		}
+	} else if !info.IsDir() {
+		return nil, errors.New("ZAR_ROOT not a directory")
+	}
```
nwt, line 128:
```suggestion
	if err := os.MkdirAll(path, 0700); err != nil {
		return nil, err
	}
```
Because the first thing os.MkdirAll does is call os.Stat and then os.FileInfo.IsDir. You can also move this inside the body of the os.IsNotExist if statement below.

archive/schema.go
alfred-landrum: Ok - I started with this by having a 'CreateOptions' struct:
```go
type CreateOptions struct {
   LogSizeThreshold *int
}
```
Where individual values, if set, specify overrides to the default. Do yall find that more palatable?

archive/schema.go
alfred-landrum: 👍

alfred-landrum:
@nwt @mattnibs : I've adjusted the passing of creation options, not using flags anymore.

cmd/zar/import/command.go
```diff
@@ -35,24 +37,24 @@ timestamp of the first zng record in that chunk.
...
+	empty       bool
 	ReaderFlags zio.ReaderFlags
 }
 
 func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
 	c := &Command{Command: parent.(*root.Command)}
 	f.StringVar(&c.root, "R", os.Getenv("ZAR_ROOT"), "root directory of zar archive for chopped files")
-	f.IntVar(&c.megaThresh, "s", 500, "target size of chopped files in MB")
-	f.IntVar(&c.byteThresh, "b", 0, "target size of chopped files in bytes (overrides -s)")
+	f.StringVar(&c.thresh, "s", units.Base2Bytes(archive.DefaultConfig.LogSizeThreshold).String(), "target size of chopped files, as '10MB' or '4GiB', etc.")
```
mattnibs, line 55: cool 👍

PR #790, opened 2020-05-15
Serialize zero spans as a json null

mattnibs:
A span with Dur == 0 is an impossible time span- why not use this
as the nil value of span?

Change json serialization of `nano.Span`s so that if span.Dur == 0
the span gets a null value in json. Like was json null values set a
span in a struct to `nano.Span{0, 0}`.

### DIFF
--- a/pkg/nano/span.go
+++ b/pkg/nano/span.go
@@ -1,6 +1,7 @@
 package nano
 
 import (
+	"bytes"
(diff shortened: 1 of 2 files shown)


mattnibs:
Wanted to gather thoughts on this. It would be nice to drop pointers to `nano.Span`'s in package api.

PR #793, opened 2020-05-18
zjson: handle aliases on record types

### LINKED ISSUE #758: Alias handling bug in zjson writer
The zjson writer handles aliases to primitive types ok, but not aliases to container types. (The zjson reader likely has a similar issue).

```sh
$ cat alias.tzng 
#0:record[host:ip]
0:[127.0.0.2;]
#myrec=record[host:ip]
#1:record[foo:myrec]
1:[[127.0.0.2;]]
19:21 ~/work/brim/zq(zeekstrings-to-zeekio)
$ zq -f zjson alias.tzng 
{"id":23,"type":[{"name":"host","type":"ip"}],"values":["127.0.0.2"]}
malformed zng value
```

henridf:
closes #758

### DIFF
--- a/zio/zjsonio/reader.go
+++ b/zio/zjsonio/reader.go
@@ -250,17 +250,16 @@ func decodeContainer(builder *zcode.Builder, typ zng.Type, body []interface{}) e
 			if k >= len(columns) {
 				return &zng.RecordTypeError{Name: "<record>", Type: typ.String(), Err: zng.ErrExtraField}
 			}
-			childType = columns[k].Type
+			childType = zng.AliasedType(columns[k].Type)
 		}
-		realType := zng.AliasedType(childType)
 		s, ok := column.(string)
 		if ok {
-			if err := decodeField(builder, realType, s); err != nil {
+			if err := decodeField(builder, childType, s); err != nil {
 				return err
 			}
 			continue
 		}
-		if utyp, ok := realType.(*zng.TypeUnion); ok {
+		if utyp, ok := childType.(*zng.TypeUnion); ok {
 			if err := decodeUnion(builder, utyp, column); err != nil {
 				return err
 			}
@@ -270,7 +269,7 @@ func decodeContainer(builder *zcode.Builder, typ zng.Type, body []interface{}) e
 		if !ok {
 			return errors.New("bad json for zjson value")
 		}
-		if err := decodeContainer(builder, realType, children); err != nil {
+		if err := decodeContainer(builder, childType, children); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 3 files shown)


zio/zjsonio/reader.go
```diff
@@ -239,7 +239,7 @@ func decodeContainer(builder *zcode.Builder, typ zng.Type, body []interface{}) e
 		// each column either a string value or an array of string values
 		if column == nil {
 			// this is an unset column
-			if zng.IsContainerType(childType) || zng.IsContainerType(columns[k].Type) {
+			if zng.IsContainerType(zng.AliasedType(childType)) || zng.IsContainerType(zng.AliasedType(columns[k].Type)) {
```
mattnibs, line 242: being that we don't seem to care about the alias types here I would do this:
- line 273: `childType = zng.AliasedType(childType)`
- line 253: `childType = zng.AliasedType(childType)`
- drop changes on line 242 and 255 and all references to `realType`.
The variable name `realType` had me confused with whether it is referring to the container Type or the child Type.

zio/zjsonio/stream.go
```diff
@@ -108,7 +108,8 @@ func encodeContainer(typ zng.Type, val []byte) (interface{}, error) {
 				childType = columns[k].Type
```
mattnibs, line 108: Being that we don't care about aliases at this point I would just on line 94 do `childType = zng.AliasedType(childType) and here: 
```suggestion
				childType = zng.AliasedType(columns[k].Type)
```

zio/zjsonio/reader.go
henridf: > being that we don't seem to care about the alias types here I would do this:

Why don't we need it here? I thought we needed to undo aliases wherever we handled types (as we do elsewhere, for example in tzngio). The whole alias scheme remains a bit fuzzy to me though, so I'd be happy to learn here.

> The variable name realType

Ok, I borrowed this name from tzngio, but agree it's not highly descriptive. I'll change the name before merging.

zio/zjsonio/stream.go
henridf: I don't think I understand this comment as calling AliasedType in the two places you suggest is functionally identical to calling it in the single place that it is currently called?

zio/zjsonio/reader.go
henridf: > > being that we don't seem to care about the alias types here I would do this:

> Why don't we need it here? I thought we needed to undo aliases wherever we handled types (as we do elsewhere, for example in tzngio). The whole alias scheme remains a bit fuzzy to me though, so I'd be happy to learn here.


@mattnibs I added a test in 0f41a8b2ce22c82c81fed686327dd36464c243c5 that answers this question - without the changes suggested in https://github.com/brimsec/zq/pull/793#discussion_r426793717, this test fails.

zio/zjsonio/stream.go
mattnibs: I didn't like the name realType- with your change below this isn't needed.

PR #795, opened 2020-05-18
archive metadata with ordered import info

alfred-landrum:
(on top of https://github.com/brimsec/zq/pull/787)

These changes cause import to enforce an ordering on incoming data, and have it remember the mapping of spans to zng files in the zar.json metadata file.  The 'mkdirs' command is removed, since the zar directories will be created as needed during archive walks.


archive/import.go
```diff
@@ -0,0 +1,126 @@
...
+	batch.Unref()
+	return nil
+}
+
+func (d *importDriver) ChannelEnd(cid int) error {
+	return d.close()
+}
+
+func (d *importDriver) Warn(warning string) error          { return nil }
+func (d *importDriver) Stats(stats api.ScannerStats) error { return nil }
```
mattnibs, line 99: Don't have to do this in the pr, but writing these is getting tired. We could do something in driver similar to what stdlib package `http` does with an `http.ResponseWriter` and `http.Flusher`. That is have add on interfaces like this:
```golang
type Warner interface {
    Warn(warning string) error
}
```
If the passed driver implements this is `driver.Run`, the warning is sent to the driver.

archive/import.go
```diff
@@ -0,0 +1,126 @@
...
+		if err := d.writeOne(batch.Index(i)); err != nil {
+			return err
+		}
+	}
+	batch.Unref()
+	return nil
+}
+
+func (d *importDriver) ChannelEnd(cid int) error {
+	return d.close()
```
mattnibs, line 95: Although this is slick, this is kind of a misuse of the `ChannelEnd` callback- if a forked flowgraph was added this would result in the scanner getting closed preemptively. I would leave this nil (like Warn and Stats) and call d.close after driver.Run finishes.

archive/import.go
```diff
@@ -0,0 +1,126 @@
...
+	zw  zbuf.Writer
+	n   int64
+
+	span    nano.Span
+	relpath string
+	spans   []SpanInfo
+}
+
+func (d *importDriver) writeOne(rec *zng.Record) error {
+	recspan := nano.NewSpanTs(rec.Ts, rec.Ts+1)
```
mattnibs, line 37: Definite nit:
```suggestion
	nano.Span{Ts: rec.Ts, Dur: 1}
```

archive/import.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+}
+
+func Import(ark *Archive, r zbuf.Reader) error {
+	proc, err := zql.ParseProc(importProc(ark))
+	if err != nil {
+		return err
+	}
+
+	fg, err := driver.Compile(context.TODO(), proc, r, false, nano.MaxSpan, zap.NewNop())
```
mattnibs, line 115: I would change this function signature to this and have the command pass in `context.TODO()`- just so the TODO where the real context should be constructed.
```suggestion
func Import(ctx context.Context, ark *Archive, r zbuf.Reader) error {
	proc, err := zql.ParseProc(importProc(ark))
	if err != nil {
		return err
	}

	fg, err := driver.Compile(ctx, proc, r, false, nano.MaxSpan, zap.NewNop())
```

archive/import.go
alfred-landrum: I added panic checks if a non-zero channel id is used; this driver is specifically for the import action below that either uses "sort ts" or "sort -r ts", so it shouldn't be called for multiple channels. PTAL!

archive/schema.go
```diff
@@ -6,20 +6,30 @@ import (
...
 	"os"
 	"path/filepath"
+	"sort"
 
 	"github.com/brimsec/zq/pkg/fs"
+	"github.com/brimsec/zq/pkg/nano"
 )
 
-const configName = "zar.json"
+const metaDataFilename = "zar.json"
```
nwt, line 15: @alfred-landrum: Metadata is one word, not two, so this should be metadataFilename.

PR #797, opened 2020-05-19
honor space datapath parameter for space storage

alfred-landrum:


zqd/space/space.go
```diff
@@ -54,7 +54,7 @@ type Space struct {
 }
 
 func newSpace(path string, conf config) (*Space, error) {
-	s, err := storage.OpenZng(path, conf.ZngStreamSize)
+	s, err := storage.OpenZng(conf.DataPath, conf.ZngStreamSize)
```
mattnibs, line 57: As an aside, this space path / dataPath is confusing even to me... Feeling like a name change is in order to distinguish them. Perhaps changing space.path to space.root?

alfred-landrum: I agree the current naming is confusing. It will lessen some in the future, when we add support for Storage that could be backed on a remote store (like S3), and we can change 'DataPath' to something else ('StorageLocation'? 'StorageURL'?).

PR #811, opened 2020-05-21
make span.Union handle zero valued spans

alfred-landrum:


pkg/nano/span.go
```diff
@@ -117,10 +121,15 @@ func (s Span) Intersect(b Span) Span {
 	return NewSpanTs(start, end)
 }
 
-// Union merges two spans returning a new span where start equals
-// min(a.start, b.start) and end equals max(a.end, b.end). Assumes the two spans
-// overlap.
+// Union returns a new span that includes the time range of this span
+// and the input span.
```
nwt, line 125: This no longer gives any indication of the behavior on spans that don't intersect. And it doesn't mention that zero spans are treated specially.

pkg/nano/span.go
```diff
@@ -21,6 +21,10 @@ type jsonSpan struct {
 	Dur Ts `json:"dur"`
 }
 
+func (s Span) IsZero() bool {
+	return s.Ts == 0 && s.Dur == 0
```
nwt, line 25:
```suggestion
	return s == Span{}
```

pkg/nano/span.go
nwt: The more I think about this, the odder it seems to imbue the zero span with special powers, and to do so here but not in other methods. This doesn't feel right.

alfred-landrum: Can you clarify, do you mean that A) it's strange to have a span function have special case handling for a zero span, or B) that the special casing ought to be done in more than just the `Union` call - or both A & B?

I think there's a case for case A: if I think of a span has a set of nanosecond-sized durations, and the various span functions as operator on these sets, then it makes sense that the functions would have rules that define what happens when the empty set (a zero span) is used.

mattnibs: @nwt What would be the result of unioning a zero duration span with another span?

mattnibs: I'm not sure whether a ZeroSpan should care about the value of `Ts` (hence why in my json null value pr I defined IsZero as dur == 0). Either way I would argue that the way Union is current working for spans with duration == 0 is incorrect.

nwt: @alfred-landrum: A and B. And C, which is that special handling for zero spans ought to be mentioned in the doc comment. And D, which is that Span{0, 0} and Span{1, 0} probably shouldn't be treated differently, though I'm not sure about that.

There's certainly a case for A, but I think the package suffers if we do just A.

pkg/nano/span.go
nwt: Or should it be `return s.Dur == 0`?

alfred-landrum:
closing this PR, filed #819 to revisit taking into account the discussion above

PR #813, opened 2020-05-22
Page iterator

aswan:
This is definitely a first cut but the desire on Slack was to create a PR for it now.  I have not had a chance to clean this up, some specific things that need cleaning are:
- error handling is sloppy and inconsistent, it many cases we just panic when encountering some parquet feature or variant that is not yet handled
- various bits of duplicated code
- little attention has been given to comments, names of functions/variables, ordering of functions in the source filed, etc.
- the wide reader interfaces that have methods for every parquet primitive type make for very verbose code.  @mccanne 's suggestion for this is to push the zng encoding all the way into the iterators which would also improve performance.

I think the last item would be best done in a separate follow-up.  As for the rest, I'm aware of all the cleanup that needs to happen, a review with detailed comments about those issues is probably not a good use of anybody's time at this stage...

### DIFF
--- a/zio/parquetio/iterator.go
+++ b/zio/parquetio/iterator.go
@@ -1,0 +1,528 @@
+package parquetio
+
+import (
+	"encoding/binary"
+	"fmt"
+	"io"
+	"math"
+
+	"github.com/apache/thrift/lib/go/thrift"
+	"github.com/xitongsys/parquet-go/common"
+	"github.com/xitongsys/parquet-go/compress"
+	"github.com/xitongsys/parquet-go/layout"
+	"github.com/xitongsys/parquet-go/parquet"
+	"github.com/xitongsys/parquet-go/reader"
+	"github.com/xitongsys/parquet-go/source"
+)
+
+type valueReader interface {
+	nextBoolean() bool
+	nextInt32() int32
+	nextInt64() int64
+	nextFloat() float64
+	nextDouble() float64
+	nextByteArray() []byte
+}
+
+type columnIterator struct {
+	pr   *reader.ParquetReader
+	name string
+
+	maxRL int32
+	maxDL int32
+
+	rowGroup int
+
+	groupRead    int
+	groupTotal   int
+	thriftReader *thrift.TBufferedTransport
+	colMetadata  *parquet.ColumnMetaData
+
+	pageRead  int
+	pageTotal int
+
+	rlReader  *hybridReader
+	dlReader  *hybridReader
+	valReader valueReader
+
+	// only used for PageType_DICTIONARY_PAGE
+	byteArrayDict [][]byte
+	// XXX need other types
+}
+
+func newColumnIterator(pr *reader.ParquetReader, name string, maxRL, maxDL int32) *columnIterator {
+	return &columnIterator{
+		pr:    pr,
+		name:  name,
+		maxRL: maxRL,
+		maxDL: maxDL,
+	}
+}
+
+func (i *columnIterator) clearDictionaries() {
+	i.byteArrayDict = nil
+	// XXX other types
+}
+
+func (i *columnIterator) loadOnePage() (*parquet.PageHeader, []byte, error) {
+	if i.groupRead == i.groupTotal {
+		if i.rowGroup >= len(i.pr.Footer.RowGroups) {
+			return nil, nil, io.EOF
+		}
+		rg := i.pr.Footer.RowGroups[i.rowGroup]
+		i.rowGroup++
+
+		i.groupRead = 0
+		i.groupTotal = int(rg.NumRows)
+
+		var col *parquet.ColumnChunk
+		for _, c := range rg.Columns {
+			if c.MetaData.PathInSchema[0] == i.name {
+				col = c
+				break
+			}
+		}
+		if col == nil {
+			panic("cannot find column")
+		}
+
+		if col.FilePath != nil {
+			panic("ColumnChunk refers to external file")
+		}
+
+		i.colMetadata = col.MetaData
+		offset := i.colMetadata.DataPageOffset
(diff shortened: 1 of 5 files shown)


aswan:
As discussed in person, we'd like to merge this as-is (pending review) and then clean up other things (consistent error handling in particular) in a follow-up PR.

zio/parquetio/iterator.go
```diff
@@ -0,0 +1,726 @@
...
+	valReader valueReader
+
+	// only used for PageType_DICTIONARY_PAGE
+	int64Dict     []int64
+	floatDict     []float64
+	byteArrayDict [][]byte
+	// XXX need other types
+}
+
+func newColumnIterator(pr *reader.ParquetReader, name string, maxRL, maxDL int32) *columnIterator {
```
mattnibs, line 68: In parquet-go do a `*source.ParquetFile.Open("")` to read each separate column. I think we might want to do this as well.

zio/parquetio/iterator.go
```diff
@@ -0,0 +1,726 @@
...
+		offset := i.colMetadata.DataPageOffset
+		if i.colMetadata.DictionaryPageOffset != nil {
+			offset = *i.colMetadata.DictionaryPageOffset
+		}
+		size := i.colMetadata.TotalCompressedSize
+
+		// XXX
+		i.thriftReader = source.ConvertToThriftReader(i.pr.PFile, offset, size)
+
+		i.clearDictionaries()
```
mattnibs, line 124: Shouldn't this get called whenever `loadDictionaryPage` is called?

zio/parquetio/iterator.go
```diff
@@ -0,0 +1,726 @@
...
+	panic("cannot read FLOAT from INT64 dictionary reader")
+}
+
+func (r *dictionaryInt64Reader) nextDouble() float64 {
+	panic("cannot read DOUBLE from INT64 dictionary reader")
+}
+
+func (r *dictionaryInt64Reader) nextByteArray() []byte {
+	panic("cannot read BYTE_ARRAY from INT64 dictionary reader")
+}
```
mattnibs, line 642: Instead of doing all this panics it might be nice to have a baseReader that these all inherit from. E.g:
```
type baseReader struct {
   name string
}

func (r *baseReader) nextFloat() float64 {
	panic(fmt.Sprintf("cannot read FLOAT from %s reader", r.name)
}
```

zio/parquetio/iterator.go
```diff
@@ -0,0 +1,726 @@
...
+	header, err := layout.ReadPageHeader(i.thriftReader)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	compressedLen := header.GetCompressedPageSize()
+
+	raw := make([]byte, compressedLen)
+	_, err = i.thriftReader.Read(raw)
+	if err != nil {
```
nwt, line 139:
```suggestion
	raw := make([]byte, header.GetCompressedPageSize())
	if _, err = i.thriftReader.Read(raw); err != nil {
```

zio/parquetio/iterator.go
aswan: That would be necessary if we had different goroutines reading different columns concurrently, but this code is all single-threaded.  I think an even better fix would be to pass in a go `io.ReaderAt` and then wrap that with something that we can give to `thrift.NewTBufferedTransport` so that we're not using a bunch of unneeded file descriptors...

zio/parquetio/iterator.go
aswan: I don't think it matters, loadDictionaryPage overwrites any existing dictionary.  And any given column can't have dictionaries of different types.  This would be a lot nicer/clearer if we had generics rather than putting dictionaries of all the primitive types into this class...

zio/parquetio/iterator.go
```diff
@@ -0,0 +1,726 @@
...
+// columnIterator reads and emits all the values from an individual
+// column inside a parquet file.  This implementation is not yet complete,
+// it only handles PLAIN and PLAIN_DICTIONARY encodings for a few
+// primitive data types.
+type columnIterator struct {
+	pr   *reader.ParquetReader
+	name string
+
+	maxRL int32
+	maxDL int32
```
nwt, line 45:
```suggestion
	maxRepetitionLevel int32
	maxDefinitionLevel int32
```

zio/parquetio/reader.go
```diff
@@ -353,57 +315,203 @@ func convertListType(els []*parquet.SchemaElement, i int) (int, *parquetColumn,
...
+	if r.record == r.total {
+		return nil, nil
+	}
+	r.record++
 
-		err := c.append(builder, fv)
+	r.builder.Reset()
+	for _, c := range r.columns {
+		err := c.append(r.builder)
 		if err != nil {
```
nwt, line 512: Nit: `if err := c.append(r.builder); err != nil {`

zio/parquetio/iterator.go
```diff
@@ -0,0 +1,723 @@
...
+	// panic("implement plain boolean reader")
+	return false
+}
+
+func (r *plainReader) nextInt32() int32 {
+	if len(r.buf) < 4 {
+		panic("underflow in PLAIN INT32")
+	}
+
+	ret := binary.LittleEndian.Uint32(r.buf[:4])
```
nwt, line 552: The runtime already does bounds checking, so there's no need to do it yourself in any of these functions.

aswan: These were placeholders to eventually turn into real error handling that propagates to the caller instead of panics but I can remove them if you prefer.

alfred-landrum:
I let @aswan know that I hit a panic with an internal test file:
https://gist.github.com/alfred-landrum/cab2c5c67f906f94aa2430a7c9e96366

He's taking a look.

PR #814, opened 2020-05-22
Add args to generate cpu/mem profile

mattnibs:
Also:
- Add cleanup function. The same as testing.Cleanup, runs when
the command has finished.

### DIFF
--- a/cmd/zq/zq.go
+++ b/cmd/zq/zq.go
@@ -10,6 +10,7 @@ import (
 	"log"
 	"os"
 	"regexp"
+	"runtime"
 	"runtime/pprof"
 
 	"github.com/brimsec/zq/ast"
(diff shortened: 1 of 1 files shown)


cmd/zq/zq.go
```diff
@@ -34,7 +36,7 @@ var Zq = &charm.Spec{
 	Name:        "zq",
 	Usage:       "zq [ options ] [ zql ] file [ file ... ]",
 	Short:       "command line logs processor",
-	HiddenFlags: "pathregexp",
+	HiddenFlags: "pathregexp,cpuprofile,memprofile",
```
nwt, line 39: Nit: Sort these.

cmd/zq/zq.go
```diff
@@ -288,3 +302,31 @@ func (c *Command) openOutput() (zbuf.WriteCloser, error) {
...
+	pprof.StartCPUProfile(f)
+	c.cleanup(pprof.StopCPUProfile)
+}
+
+func (c *Command) runMemProfile() {
+	f, err := os.Create(c.memprofile)
+	if err != nil {
+		log.Fatal(err)
+	}
+	pprof.WriteHeapProfile(f)
```
nwt, line 330: Maybe add a call to runtime.GC as shown at https://pkg.go.dev/runtime/pprof?tab=doc#hdr-Profiling_a_Go_program?

mattnibs: Yep, saw this 👍

PR #816, opened 2020-05-25
zar index: use procs

### LINKED ISSUE #798: zar index: use a zql flowgraph
splitting this out from https://github.com/brimsec/zq/issues/646#issuecomment-628899965, where @mccanne said:

> I think we should approach this one more holistically. If you look through the zar README, the key steps are here:
> 
> zar zq -o keys.zng "_path=conn | sum(resp_bytes) as resp_bytes by id.resp_h,id.orig_h | sort id.resp_h,id.orig_h" _
> zar zdx -o custom2 -k id.resp_h,id.orig_h keys.zng
> It would be nice to just re-use this zql machinery in the zar index path and get rid of zdx.MemTable completely. If the zql group-by could spill, then we don't need to re-create zdx.Memtable spillling. The "zar zdx" pass could take input from a scanner/combiner on the spilled files.

henridf:
Switch zar indexing over to use a zql flowgraph. The in-memory table from `zdx/mem.go` is no longer used in zar indexing and has been replaced by the groupby proc. 

CLI commands for custom index creation have also been simplified. A custom index that was previously created via the two-step:
```
zar zq -o groupby.zng "count() by _path, id.orig_h | sort id.orig_h" _
zar zdx -o custom -k id.orig_h groupby.zng
```
is now created via a zar index command, like field and type indexes:

```
zar index -q -o custom -k id.orig_h -z "count() by _path, id.orig_h | sort id.orig_h"
```

closes #798


henridf:
Some performance measurements on dubset year 1:

```sh
$ time zq ~/Downloads/sampledata/wrccdc-year1/zeek-logs/* | zar import -s 25MB  -

real    13m48.303s
user    20m28.394s
sys     2m31.501s

$ time zar index :ip
master:
real    2m47.777s
user    2m56.512s
sys     0m8.209s

this branch:
real    3m44.164s
user    5m6.517s
sys     0m12.322s

$ time zar index id.orig_h
master:
real    1m44.465s
user    1m54.889s
sys     0m6.547s

this branch:
real    2m21.282s
user    3m22.979s
sys     0m10.012s
```

archive/proc.go
```diff
@@ -12,6 +14,82 @@ import (
 
 var ErrSyntax = errors.New("syntax/format error encountered parsing zng data")
 
+// A FieldCutter is a custom proc that, given an input record and a
+// field name, outputs one record per input record containing the
+// field, with the field as only column.  It is used for field-based
+// indexing.
+type FieldCutter struct {
```
alfred-landrum, line 21: Was it impractical to alter the cut proc to add the type tracking?

Would you add to the comment here, reflecting your commit comment, that this was put in because the cut proc couldn't track type changes?

cmd/zar/index/command.go
```diff
@@ -30,6 +31,12 @@ you would run
 
 Each pattern results a separate zdx index file for each log file found.
 
+For custom indexes, zql flowgraph can be used instead of a
```
alfred-landrum, line 34:
```suggestion
For custom indexes, zql can be used instead of a
```

cmd/zar/index/command.go
```diff
@@ -30,6 +31,12 @@ you would run
...
+pattern. This requires specifying the key and output file name. For
+example
```
alfred-landrum, line 36:
```suggestion
example:
```

cmd/zar/index/command.go
```diff
@@ -42,19 +49,27 @@ func init() {
...
 }
 
 func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
 	c := &Command{Command: parent.(*root.Command)}
 	f.StringVar(&c.root, "R", os.Getenv("ZAR_ROOT"), "root directory of zar archive to walk")
+	f.StringVar(&c.keys, "k", "key", "one or more comma-separated key fields")
+	f.IntVar(&c.framesize, "f", 32*1024, "minimum frame size used in zdx file")
+	f.StringVar(&c.outputFile, "o", "zdx", "output index name (for custom indexes)")
 	f.BoolVar(&c.quiet, "q", false, "don't print progress on stdout")
+	f.StringVar(&c.zql, "z", "", "zql flowgraph for custom indexes")
```
alfred-landrum, line 67:
```suggestion
	f.StringVar(&c.zql, "z", "", "zql for custom indexes")
```

archive/proc.go
```diff
@@ -0,0 +1,161 @@
...
+
+	"github.com/brimsec/zq/ast"
+	"github.com/brimsec/zq/expr"
+	"github.com/brimsec/zq/proc"
+	"github.com/brimsec/zq/zbuf"
+	"github.com/brimsec/zq/zcode"
+	"github.com/brimsec/zq/zng"
+)
+
+var ErrSyntax = errors.New("syntax/format error encountered parsing zng data")
```
alfred-landrum, line 15: This is unused.

archive/proc.go
henridf: I don't think it would have been impractical, but I didn't want to add a "error on type changes" mode to the cut proc as it didn't seem to belong there and wouldn't be exposed from zql. And FieldCutter is way simpler/shorter than cut proc (the cut proc actually does a whole bunch of work to enable cutting of different types).

PR #818, opened 2020-05-26
Simplify zar zq

### LINKED ISSUE #757: Factor out common option handling between zq and zar zq
`zq` and `zar zq` share a number of command-line options. Currently, the related handling code is duplicated. We should unify this. As a code/refactor task this should be straightforward, but this task has the prerequisite of settling "what options should `zar zq` support" first.

### LINKED ISSUE #759: Cleaner `zar zq` output on stdout with text formats
The zar README brings the reader to this `zar zq` query:

```sh
$ zar zq "count()" _ | zq -f table -
COUNT
222617
223815
218575
...
```

at which point, one might recall that zar zq supportsd different output formats, and directly try:

```sh
$ zar zq -f table "count()" _ 
COUNT
222617
COUNT
223815
COUNT
218575
...
```

which gives correct results, but with extra headers that are a little surprising. This is not a bug, but would likely be unexpected to someone without deep familiarity with the inner workings of zq and zar (as a README reader might be).

This can be addressed as suggested by @mccanne:

> A fix could be for stdout to merge the zng sub-streams that are created during the walk into one type context  and have a single writer instance for stdout.

henridf:
This PR has two simplifying changes to `zar zq`. The change itself is straightforward but has implications on zar in two ways:

- The first commit removes the `-f` flag, such that `zar zq` always emits zng. This is the way it is used throughout the README, and for other formats, one can pipe `zar zq` into `zq`, as in the README.  (And as it says, "we love pipes"!). This closes #759.

- The second commit removes all input reader flags. Given that zar archive contents are stored in zng, there's no need for json-related flags.  This closes #757.

### DIFF
--- a/cmd/zar/zq/command.go
+++ b/cmd/zar/zq/command.go
@@ -34,9 +34,9 @@ var Zq = &charm.Spec{
 	Long: `
 "zar zq" descends the directory given by the -R option (or ZAR_ROOT env) looking for
 logs with zar directories and for each such directory found, it runs
-the zq logic relative to that directory.  The file names here are relative to that
-directory and the special name "_" refers to the actual log file
-in the parent of the zar directory.
+the zq logic relative to that directory and emits the results in zng format.
+The file names here are relative to that directory and the special name "_" refers
+to the actual log file in the parent of the zar directory.
 
 If the root directory is not specified by either the ZAR_ROOT environemnt
 variable or the -R option, then the current directory is assumed.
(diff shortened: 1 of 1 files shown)


cmd/zar/zq/command.go
```diff
@@ -224,7 +179,8 @@ func (c *Command) openOutput(zardir, filename string) (zbuf.WriteCloser, error)
 	if path != "" {
 		path = filepath.Join(zardir, filename)
 	}
-	w, err := emitter.NewFile(path, &c.WriterFlags)
+	flags := zio.WriterFlags{Format: "zng"}
+	w, err := emitter.NewFile(path, &flags)
```
nwt, line 183:
```suggestion
	w, err := emitter.NewFile(path, &zio.WriterFlags{Format: "zng"})
```

PR #823, opened 2020-05-26
Generic Spill to Disk SortReader

mattnibs:
An implementation of a generic disk spill sort reader than could be
used as a basis for the sort and group by proc.

Add package zio/transform for placement zio readers that transform.

Also:
- Adjust zbuf.Combiner so it takes an expr.SortFn as an arugment and
can there for be used anywhere where streams need to be merged in a
sorted fashion.


mattnibs:
I could see that. Ok, closing.

PR #825, opened 2020-05-27
Stop using ParquetReader completely

aswan:
The main thing we were still using the ParquetReader class for was its
SchemaHandler.  But that was clumsy as it still had vestiges of the
renaming-columns-to-valid-go-public-field-names gunk that was needed for
its reflect-based api.

In this update, things that care about the parquet schema read the parquet
file footer directly, bypassing parqet-go's SchemaHandler class.  We're
still using the ParquetFile interface and low-level handlers for thrift,
compression, etc. from parquet-go, but we're now totally free from its
reader package.


aswan:
> Would it make sense to still have a struct that contains the source.ParquetFile and *parquet.FileMetaData values together, since they're typically passed together?

Yes, I think it makes sense to just use the Reader struct for this instead of creating something new.  Cleaned up to follow this pattern...

aswan:
Ugh, this passes our ztests but its broken with another test file that I have here.  I'll see if I can debug this...

PR #830, opened 2020-05-28
Add options to ParquetReader

aswan:
The column handling can now be customized -- columns with unknown structure
that the reader cannot handle can either be a fatal error (the default) or
ignored.  Also, a specific list of column names can be passed in in which
case the reader will not do the I/O and decoding work for other columns.
None of these options are yet wired up to zq command line arguments.

### DIFF
--- a/zio/parquetio/reader.go
+++ b/zio/parquetio/reader.go
@@ -126,8 +126,8 @@ func simpleParquetTypeToZngType(typ HandledType) zng.Type {
 }
 
 type ReaderOpts struct {
-	columns                []string
-	ignoreUnhandledColumns bool
+	Columns                []string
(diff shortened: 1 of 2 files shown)


zio/parquetio/reader.go
```diff
@@ -125,6 +125,11 @@ func simpleParquetTypeToZngType(typ HandledType) zng.Type {
 	panic(fmt.Sprintf("unhandled type %d", typ))
 }
 
+type ReaderOpts struct {
+	columns                []string
+	ignoreUnhandledColumns bool
```
nwt, line 130: We should export these so they're available outside this package.

aswan: whoops indeed we should.

PR #831, opened 2020-05-28
reduce archive zng files in create archive test

alfred-landrum:


alfred-landrum:
This is just a workaround for @aswan who's seeing a 'too many files' open error, a real fix for the general issue will come later, filed https://github.com/brimsec/zq/issues/832 .

henridf:
(fwiw a good old `ulimit -n` has also let me run these tests)

PR #833, opened 2020-05-29
initial archive index search api

alfred-landrum:
This adds an api endpoint for index queries of the form:
```
POST /space/{space_id}/indexsearch
{
  "index_name": "<name>"
  "patterns": ["<pattern1>",...]
}
```

Valid requests are similar to the options for `zar find`: For "standard" indices, requests have a blank index_name, and a single pattern of forms like `:ip=10.0.0.1` or `uri=/hello`.  Custom indices require specifying an index name with one or more patterns; there's not yet an api to list the available indices for an archive space.

Note: As a side effect of making this work on Windows, the file names for zar indices have changed, replacing colons with dashes, eg: `zdx:type:ip.zng` becomes `zdx-type-ip.zng`. Colons are not valid filename characters for Windows.


archive/rule.go
```diff
@@ -1,34 +1,53 @@
...
 	"strings"
 
 	"github.com/brimsec/zq/ast"
 	"github.com/brimsec/zq/zng/resolver"
+	"github.com/brimsec/zq/zqe"
 	"github.com/brimsec/zq/zql"
 )
 
-func ParsePattern(in string) (string, string, error) {
+type IndexQuery struct {
```
henridf, line 14: nittish: do IndexName and Patterns need to be exported?

alfred-landrum: they definitely don't right now, so I'll un-export them.

archive/schema.go
```diff
@@ -28,9 +28,15 @@ type MetaData struct {
...
+
+func (l LogID) Path(ark *Archive) string {
+	return filepath.Join(ark.Root, filepath.ToSlash(string(l)))
+}
+
 type SpanInfo struct {
-	Span nano.Span `json:"span"`
-	Part string    `json:"part"`
+	Span  nano.Span `json:"span"`
+	LogID LogID     `json:"log_id"`
```
alfred-landrum, line 39: FYI, I'm following the pattern I see in the api package to use underscores for json field names.

zqd/api/connection.go
```diff
@@ -209,6 +209,23 @@ func (c *Connection) Search(ctx context.Context, search SearchRequest, params ma
...
 }
 
+func (c *Connection) IndexSearch(ctx context.Context, space SpaceID, search IndexSearchRequest, params map[string]string) (Search, error) {
+	req := c.Request(ctx).
+		SetBody(search).
+		SetQueryParam("format", "zng")
+
+	for n, v := range params {
+		req.SetQueryParam(n, v)
+	}
```
mattnibs, line 219: SetQueryParams does the same thing: https://github.com/go-resty/resty/blob/v2.3.0/client.go#L258
```suggestion
		SetQueryParam("format", "zng").
		SetQueryParams(params)
```

mattnibs: (we also at some point need to stop passing accept-type as a query parameter)

zqd/search/zng.go
```diff
@@ -69,3 +69,7 @@ func (r *ZngOutput) SendControl(ctrl interface{}) error {
 	r.flush()
 	return nil
 }
+
+func (r *ZngOutput) ContentType() string {
+	return "application/x-zng"
```
mattnibs, line 74: This should be a `const` in api so the client also has access to this and can respond appropriately.

zqd/storage/archivestore/archivestore.go
```diff
@@ -62,17 +64,56 @@ func (s *Storage) Open(ctx context.Context, span nano.Span) (zbuf.ReadCloser, er
...
+func (s *Storage) IndexSearch(ctx context.Context, query archive.IndexQuery) (zbuf.ReadCloser, error) {
+	ctx, cancel := context.WithCancel(ctx)
+	is := &indexSearch{
+		ctx:    ctx,
+		cancel: cancel,
+		hits:   make(chan *zng.Record),
+	}
+	go func() {
+		defer close(is.hits)
+		is.err = archive.Find(ctx, s.ark, query, is.hits, archive.AddPath(archive.DefAddPathField, false))
```
mattnibs, line 116:
```suggestion
		is.err = archive.Find(ctx, s.ark, query, is.hits, archive.AddPath(archive.DefAddPathField, false))
		close(is.hits)
```

zqd/search/indexsearch.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+	rc, err := arkstore.IndexSearch(ctx, query)
+	if err != nil {
+		return nil, err
+	}
+	return &IndexSearch{rc}, nil
+}
+
+func (s *IndexSearch) Run(out Output) (err error) {
+	id := int64(0)
```
mattnibs, line 36: I would just drop this id here.

archive/finder.go
```diff
@@ -11,83 +11,112 @@ import (
...
-	// a single stream
-	zctx := resolver.NewContext()
-	var pathCol []zng.Column
-	if pathField != "" {
-		// Create a type alias called "zfile" that the client will
-		// understand as the pathname of a zng file, e.g., available
-		// as a new space via zqd.
-		typ, err := zctx.LookupTypeAlias("zfile", zng.TypeString)
-		if err != nil {
+func Find(ctx context.Context, ark *Archive, query IndexQuery, hits chan<- *zng.Record, opts ...FindOption) error {
```
alfred-landrum, line 74: This is the first usage of functional options in the zq repo, so an fyi on my rationale here:
- addPath needs to happen in the search loop so it has access to the originating log file of the record.
- addPath is almost just a transform on the output records, but it needs a small amount of setup (the `LookupTypeAlias` call), and it needs the `zctx` at both setup time & while it operates.
- addPath takes two parameters that are only relevant to itself, not the overall find/search operation.
- addPath isn't used for every find call.

The above led me to consider how to make addPath a kind of optional transformation that gets called within Find, and the zctx usage for its setup means it needed two parts. Overall, that seemed like a good match for the functional option technique.

(fyi, I think the `driver.Compile` related functions & arguments would also work well as functional options.)

archive/finder.go
```diff
@@ -11,83 +11,112 @@ import (
 	"github.com/brimsec/zq/zng/resolver"
 )
 
+type findOptions struct {
+	skipMissing bool
+	zctx        *resolver.Context
+	addPath     func(ark *Archive, si SpanInfo, rec *zng.Record) (*zng.Record, error)
+}
+
+type FindOption func(*findOptions) error
```
mattnibs, line 20: You are exporting a function with an un-exported  argument.

mattnibs: I think you are perhaps looking to do something similar to what zap does with options? https://github.com/uber-go/zap/blob/master/options.go#L30

alfred-landrum: That's intended: it allows users outside of the package write code like:
```go
	var findOptions []archive.FindOption
	if c.pathField != "" {
		findOptions = append(findOptions, archive.AddPath(c.pathField, true))
	}
```
to create slices of `FindOption`'s. However, they don't need access to the internal workings of the `findOptions` type.

alfred-landrum: (I pointed @mattnibs to dave cheney's blog post on functional options in go: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis )

archive/finder.go
```diff
@@ -11,83 +11,112 @@ import (
...
+
+func SkipMissing() FindOption {
+	return func(opt *findOptions) error {
+		opt.skipMissing = true
+		return nil
+	}
+}
+
+const DefAddPathField = "_log"
```
nwt, line 29:
```suggestion
const DefaultAddPathField = "_log"
```

archive/finder.go
```diff
@@ -11,83 +11,112 @@ import (
...
-		// as a new space via zqd.
-		typ, err := zctx.LookupTypeAlias("zfile", zng.TypeString)
-		if err != nil {
+func Find(ctx context.Context, ark *Archive, query IndexQuery, hits chan<- *zng.Record, opts ...FindOption) error {
+	opt := findOptions{
+		zctx:    resolver.NewContext(),
+		addPath: func(_ *Archive, _ SpanInfo, rec *zng.Record) (*zng.Record, error) { return rec, nil },
+	}
+	for i := range opts {
+		if err := opts[i](&opt); err != nil {
```
nwt, line 80:
```suggestion
	for _, o := range opts {
		if err := o(&opt); err != nil {
```

archive/finder.go
```diff
@@ -11,83 +11,112 @@ import (
...
-				var err error
-				hit, err = zctx.AddColumns(hit, pathCol, []zng.Value{val})
-				if err != nil {
-					cancel()
-					for _ = range searchHits {
-						// let search unravel
-					}
-					return err
-				}
+			xhit, err := opt.addPath(ark, si, hit)
```
nwt, line 99: Nit: xhit is a weird name. Maybe just h or hit2?

archive/import.go
```diff
@@ -28,24 +30,28 @@ type importDriver struct {
...
+		if err := os.MkdirAll(dpath, 0755); err != nil {
 			return err
 		}
-		d.relpath = filepath.Join(tsDir(rec.Ts), rec.Ts.StringFloat()+".zng")
-		path := filepath.Join(d.ark.Root, d.relpath)
+
 		//XXX for now just truncate any existing file.
 		// a future PR will do a split/merge.
-		out, err := fs.OpenFile(path, os.O_TRUNC|os.O_WRONLY|os.O_CREATE, 0644)
+		fpath := filepath.Join(d.ark.Root, dname, fname)
```
nwt, line 53:
```suggestion
		fpath := filepath.Join(dpath, fname)
```

archive/rule.go
```diff
@@ -1,34 +1,53 @@
...
+}
+
+func ParseIndexQuery(indexName string, patterns []string) (iq IndexQuery, err error) {
+	if len(patterns) == 0 {
+		return iq, zqe.E(zqe.Invalid, "no search patterns")
+	}
+	if indexName != "" {
+		iq.indexName = indexName
+		iq.patterns = patterns
+		return
```
nwt, line 26: Nit: There's no reason to use a naked return here or name the result parameters, so don't.

archive/schema.go
```diff
@@ -28,9 +28,15 @@ type MetaData struct {
 	Spans             []SpanInfo     `json:"spans"`
 }
 
+type LogID string
+
+func (l LogID) Path(ark *Archive) string {
+	return filepath.Join(ark.Root, filepath.ToSlash(string(l)))
```
nwt, line 34: Passing the result of filepath.ToSlash to filepath.Join doesn't seem right. Did you mean path.Join?

zqd/storage/archivestore/archivestore.go
```diff
@@ -40,8 +41,9 @@ func (s *Storage) Open(ctx context.Context, span nano.Span) (zbuf.ReadCloser, er
 		}
 	}()
 	zctx := resolver.NewContext()
-	err = archive.SpanWalk(s.ark, func(sp nano.Span, zngpath string) error {
-		if !span.Overlaps(sp) {
+	err = archive.SpanWalk(s.ark, func(si archive.SpanInfo, zardir string) error {
+		zngpath := archive.ZarDirToLog(zardir)
```
nwt, line 45: Nit: Move zngpath's declaration adjacent to its use. Or better yet, drop it and just say `f, err := fs.Open(archive.ZarDirToLog(zardir)`.

zqd/handlers.go
```diff
@@ -404,6 +407,42 @@ loop:
...
+	defer srch.Close()
+
+	out, err := getSearchOutput(w, r)
+	if err != nil {
+		respondError(c, w, r, err)
+		return
+	}
+
+	w.Header().Set("Content-Type", out.ContentType())
+	if err = srch.Run(out); err != nil {
```
nwt, line 441:
```suggestion
	if err := srch.Run(out); err != nil {
```

zqd/handlers.go
```diff
@@ -97,28 +97,31 @@ func handleSearch(c *Core, w http.ResponseWriter, r *http.Request) {
...
 
-	var out search.Output
+	out, err := getSearchOutput(w, r)
+	if err != nil {
+		respondError(c, w, r, err)
+		return
+	}
+
+	w.Header().Set("Content-Type", out.ContentType())
+	if err = srch.Run(out); err != nil {
```
nwt, line 107:
```suggestion
	if err := srch.Run(out); err != nil {
```

zqd/handler.go
```diff
@@ -35,6 +35,7 @@ func NewHandler(core *Core, logger *zap.Logger) http.Handler {
 	h.Handle("/space/{space}/pcap", handlePcapSearch).Methods("GET")
 	h.Handle("/space/{space}/pcap", handlePcapPost).Methods("POST")
 	h.Handle("/space/{space}/log", handleLogPost).Methods("POST")
+	h.Handle("/space/{space}/index_search", handleIndexSearch).Methods("POST")
```
nwt, line 38: Nit: I vote for indexsearch over index_search.

zqd/api/connection.go
```diff
@@ -209,6 +207,21 @@ func (c *Connection) Search(ctx context.Context, search SearchRequest, params ma
 	return NewZngSearch(r), nil
 }
 
+func (c *Connection) IndexSearch(ctx context.Context, space SpaceID, search IndexSearchRequest, params map[string]string) (Search, error) {
+	req := c.Request(ctx).
+		SetBody(search).
+		SetQueryParam("format", "zng")
+
```
nwt, line 214:
```suggestion
```

cmd/zar/find/command.go
```diff
@@ -122,8 +119,8 @@ func (c *Command) Run(args []string) error {
 	hits := make(chan *zng.Record)
 	var searchErr error
 	go func() {
-		searchErr = archive.Find(ctx, ark, indexFile, patterns, hits, c.pathField, c.skipMissing)
-		close(hits)
+		defer close(hits)
+		searchErr = archive.Find(ctx, ark, query, hits, findOptions...)
```
nwt, line 123: No need for defer here.

zqd/search/search.go
```diff
@@ -25,6 +25,11 @@ const DefaultMTU = 100
 
 const StatsInterval = time.Millisecond * 500
 
+const (
+	MimeTypeNDJSON = "application/ndjson"
```
nwt, line 29: https://github.com/ndjson/ndjson-spec#33-mediatype-and-file-extensions says
```suggestion
	MimeTypeNDJSON = "application/x-ndjson"
```

mattnibs: Actually @nwt  the `x-` convention has been deprecated. Elastic also just does `application/ndjson`. We've had this discussion on another pr.

nwt: @mattnibs: I don't see application/ndjson in https://github.com/elastic/elasticsearch, but I do see application/x-ndjson.

mattnibs: Looks like they were going to do it but decided it would be too big of hassle: https://github.com/elastic/elasticsearch/issues/25718

alfred-landrum:
Ok, so we'll use "application/x-ndjson" and "application/x-zng".

zqd/search/search.go
mattnibs: Whatever we do we should remain consistent. Because in other places we are consistently doing `application/ndjson` we should go with that, then consider if we want to switch to `x-ndjson` in another pr.

alfred-landrum: I looked in the Brim & zq repo, and didn't see any other references for ndjson, where else do we use `application/ndjson`?

PR #837, opened 2020-05-29
Support grouping by expressions

### LINKED ISSUE #535: Let groupby use arbitrary expressions
The `by ...` clause in a groupby proc is currently a list of field names.  We should generalize this to allow an arbitrary expression.  This would enable lots of things that can't be easily expressed right now, such as
```
count() by len(someset)
count() by String.tolower(something)
count() by 1000*Math.round(bytes/1000)
```

henridf:
I still need to add tests.

closes #535

### DIFF
--- a/archive/rule.go
+++ b/archive/rule.go
@@ -44,6 +44,11 @@ const framesize = 32 * 1024 * 2
 
 const keyName = "key"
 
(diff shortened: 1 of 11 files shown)


henridf:
This needs more work around the handling of ErrUnset when evaluating the key expression.

PR #838, opened 2020-05-29
fix package colw

mattnibs:
Was not working at all. Fix to work like columns in the bash ls
command.


mattnibs:
> Would writing tests for this package be worth the effort? (Honest question.)

@nwt It's so dead simple I'd rather spare the energy.

PR #840, opened 2020-06-01
Support grouping by computed keys

### LINKED ISSUE #535: Let groupby use arbitrary expressions

henridf:
closes #535

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -176,6 +176,7 @@ func (*FunctionCall) exprNode()          {}
 func (*CastExpression) exprNode()        {}
 func (*Literal) exprNode()               {}
 func (*FieldRead) exprNode()             {}
+func (*FieldCall) exprNode()             {}
 
 // ----------------------------------------------------------------------------
 // Procs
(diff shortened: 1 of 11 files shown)


proc/groupby.go
```diff
@@ -15,12 +15,13 @@ import (
 	"github.com/brimsec/zq/zcode"
 	"github.com/brimsec/zq/zng"
 	"github.com/brimsec/zq/zng/resolver"
+	"github.com/pkg/errors"
```
alfred-landrum, line 18: Was this intentional, or did you mean just `"errors"`?

henridf: oops, it was certainly not! i should pay closer attention to what goimports is doing for me :)

alfred-landrum: I figured, the same has happened to me :)

ast/ast.go
```diff
@@ -283,11 +284,11 @@ type (
 
 	PutProc struct {
 		Node
-		Clauses []PutClause `json:"clauses"`
+		Clauses []Assignment `json:"clauses"`
```
nwt, line 287: If you're going to rename this type, please finish the job by renaming the fields and variables of the type, either in this PR or in another. Leaving the clause terminology in place is confusing.

henridf: Ok. How far would you like this to go: all the way including renaming things like  https://github.com/brimsec/zq/blob/master/proc/put.go#L39 ?

henridf: (I'll do this in a separate PR upon answer to the above question).

PR #843, opened 2020-06-01
zapi: Add get command

### LINKED ISSUE #526: CLI tool for interfacing with a zqd instance
We have code from a previous CLI tool that could be ported over for chatting with a `zqd` instance.

mattnibs:
Adds searching capability to zapi via the get command.

Also:
- Add repl mode which launches when no commands are provided to zapi.

closes #526

### DIFF
--- a/cmd/zapi/cmd/get/get.go
+++ b/cmd/zapi/cmd/get/get.go
@@ -41,7 +41,6 @@ type Command struct {
 	protocol   string
 	dir        string
 	outputFile string
-	reverse    bool
 	stats      bool
 	warnings   bool
 	wire       bool
@@ -50,11 +49,10 @@ type Command struct {
 
 func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
 	c := &Command{Command: parent.(*cmd.Command)}
-	f.StringVar(&c.Format, "f", "text", "format for output data [ndjson,text,table,tzng,zeek,zng]")
+	f.StringVar(&c.Format, "f", "text", "format for output data [zng,ndjson,table,text,types,zeek,zjson,tzng]")
 	f.StringVar(&c.protocol, "p", "zng", "protocol to use for search request [ndjson,zjson,zng]")
 	f.StringVar(&c.dir, "d", "", "directory for output data files")
 	f.StringVar(&c.outputFile, "o", "", "write data to output file")
-	f.BoolVar(&c.reverse, "R", false, "reverse search order (from oldest to newest)")
 	f.BoolVar(&c.stats, "S", false, "display search stats on stderr")
 	f.BoolVar(&c.warnings, "W", false, "display warnings on stderr")
 	f.BoolVar(&c.ShowTypes, "T", false, "display field types in text output")
@@ -80,7 +78,7 @@ func (c *Command) Run(args []string) error {
 	if err != nil {
 		return err
 	}
-	req, err := parseExpr(id, expr, c.reverse)
+	req, err := parseExpr(id, expr)
 	if err != nil {
 		return fmt.Errorf("parse error: %s", err)
 	}
@@ -119,7 +117,7 @@ func (c *Command) Run(args []string) error {
 }
 
 // parseExpr creates an api.SearchRequest to be used with the client.
-func parseExpr(spaceID api.SpaceID, expr string, reverse bool) (*api.SearchRequest, error) {
+func parseExpr(spaceID api.SpaceID, expr string) (*api.SearchRequest, error) {
 	search, err := zql.ParseProc(expr)
 	if err != nil {
 		return nil, err
@@ -131,7 +129,6 @@ func parseExpr(spaceID api.SpaceID, expr string, reverse bool) (*api.SearchReque
 	return &api.SearchRequest{
 		Space: spaceID,
 		Proc:  proc,
-		Dir:   map[bool]int{true: 1, false: -1}[reverse],
 	}, nil
 }
 
--- a/cmd/zapi/cmd/get/get.go
+++ b/cmd/zapi/cmd/get/get.go
@@ -53,7 +53,7 @@ func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
 	f.StringVar(&c.dir, "d", "", "directory for output data files")
 	f.StringVar(&c.outputFile, "o", "", "write data to output file")
 	f.BoolVar(&c.stats, "S", false, "display search stats on stderr")
-	f.BoolVar(&c.warnings, "W", false, "display warnings on stderr")
+	f.BoolVar(&c.warnings, "W", true, "display warnings on stderr")
 	f.BoolVar(&c.ShowTypes, "T", false, "display field types in text output")
 	f.BoolVar(&c.ShowFields, "F", false, "display field names in text output")
 	f.BoolVar(&c.EpochDates, "E", false, "display epoch timestamps in text output")
@@ -102,13 +102,12 @@ func (c *Command) Run(args []string) error {
 	if err != nil {
 		return err
 	}
-	stream := api.NewZngSearch(r) // XXX fix this to match specified format
+	stream := api.NewZngSearch(r)
 	stream.SetOnCtrl(c.handleControl)
 	if err := zbuf.Copy(zbuf.NopFlusher(writer), stream); err != nil {
 		writer.Close()
 		if c.Context().Err() != nil {
-			fmt.Fprintln(os.Stderr, "search aborted")
-			return nil
+			return errors.New("search aborted")
 		}
 		return err
 	}
--- a/cmd/zapi/cmd/get/get.go
+++ b/cmd/zapi/cmd/get/get.go
@@ -37,7 +37,6 @@ func init() {
 type Command struct {
 	*cmd.Command
 	zio.WriterFlags
-	// format     string
 	protocol   string
 	dir        string
 	outputFile string
--- a/cmd/zapi/cmd/get/get.go
+++ b/cmd/zapi/cmd/get/get.go
@@ -41,7 +41,6 @@ type Command struct {
 	protocol   string
 	dir        string
 	outputFile string
-	verbose    bool
 	reverse    bool
 	stats      bool
 	warnings   bool
@@ -55,7 +54,6 @@ func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
 	f.StringVar(&c.protocol, "p", "zng", "protocol to use for search request [ndjson,zjson,zng]")
 	f.StringVar(&c.dir, "d", "", "directory for output data files")
 	f.StringVar(&c.outputFile, "o", "", "write data to output file")
-	f.BoolVar(&c.verbose, "v", false, "show verbose details")
 	f.BoolVar(&c.reverse, "R", false, "reverse search order (from oldest to newest)")
 	f.BoolVar(&c.stats, "S", false, "display search stats on stderr")
 	f.BoolVar(&c.warnings, "W", false, "display warnings on stderr")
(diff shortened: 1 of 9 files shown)


cmd/zapi/cmd/get/get.go
```diff
@@ -0,0 +1,188 @@
...
+
+func init() {
+	cmd.CLI.Add(Get)
+	cmd.Get = Get
+}
+
+type Command struct {
+	*cmd.Command
+	zio.WriterFlags
+	// format     string
```
mattnibs, line 40:
```suggestion
```

nwt:
The REPL deserves its own PR.

mattnibs:
> The REPL deserves its own PR.

I don't disagree. That said this pr series has been pretty drawn out, especially for porting existing code to this repo. Unless there's a serious objection I would appreciate a review as-is.

cmd/zapi/cmd/get/get.go
```diff
@@ -0,0 +1,188 @@
...
+	dir        string
+	outputFile string
+	reverse    bool
+	stats      bool
+	warnings   bool
+	wire       bool
+	final      *api.SearchStats
+}
+
+func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
```
alfred-landrum, line 51: Will you please remove all of the ununsed options?

mattnibs: I thought I got them all, let me take another look.

cmd/zapi/cmd/get/get.go
```diff
@@ -0,0 +1,188 @@
...
+		return nil, err
+	}
+	proc, err := json.Marshal(search)
+	if err != nil {
+		return nil, err
+	}
+	return &api.SearchRequest{
+		Space: spaceID,
+		Proc:  proc,
+		Dir:   map[bool]int{true: 1, false: -1}[reverse],
```
alfred-landrum, line 134: As noted above, we can drop the reverse option & always specify -1 for the search direction until we support forward searches in zqd.

cmd/zapi/cmd/get/get.go
```diff
@@ -0,0 +1,188 @@
...
+	warnings   bool
+	wire       bool
+	final      *api.SearchStats
+}
+
+func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
+	c := &Command{Command: parent.(*cmd.Command)}
+	f.StringVar(&c.Format, "f", "text", "format for output data [ndjson,text,table,tzng,zeek,zng]")
+	f.StringVar(&c.protocol, "p", "zng", "protocol to use for search request [ndjson,zjson,zng]")
+	f.StringVar(&c.dir, "d", "", "directory for output data files")
```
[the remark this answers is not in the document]
mattnibs: You need this to output zeek log files. Should we remove this option?

cmd/zapi/cmd/get/get.go
```diff
@@ -0,0 +1,188 @@
...
+	stats      bool
+	warnings   bool
+	wire       bool
+	final      *api.SearchStats
+}
+
+func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
+	c := &Command{Command: parent.(*cmd.Command)}
+	f.StringVar(&c.Format, "f", "text", "format for output data [ndjson,text,table,tzng,zeek,zng]")
+	f.StringVar(&c.protocol, "p", "zng", "protocol to use for search request [ndjson,zjson,zng]")
```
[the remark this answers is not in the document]
mattnibs: I would prefer to keep this, but we could make the option hidden. This is helpful for debugging bugs with the wire format.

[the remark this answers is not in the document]
mattnibs: For instance the brim client uses zjson not zng.

cmd/zapi/cmd/get/get.go
```diff
@@ -0,0 +1,188 @@
...
+	f.StringVar(&c.dir, "d", "", "directory for output data files")
+	f.StringVar(&c.outputFile, "o", "", "write data to output file")
+	f.BoolVar(&c.reverse, "R", false, "reverse search order (from oldest to newest)")
```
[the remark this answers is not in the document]
mattnibs: Sure.

cmd/zapi/cmd/get/get.go
mattnibs: Looks like they are all connected. The zio.WriterFlags obscure some of this.

cmd/zapi/cmd/get/get.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+	defer r.Close()
+	if c.wire {
+		return c.runWireSearch(r)
+	}
+	writer, err := openOutput(c.dir, c.outputFile, c.WriterFlags)
+	if err != nil {
+		return err
+	}
+	stream := api.NewZngSearch(r) // XXX fix this to match specified format
```
alfred-landrum, line 105: the comment is referring to the wire protocol, not the output format, right?

cmd/zapi/cmd/get/get.go
```diff
@@ -0,0 +1,184 @@
...
+}
+
+func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
+	c := &Command{Command: parent.(*cmd.Command)}
+	f.StringVar(&c.Format, "f", "text", "format for output data [zng,ndjson,table,text,types,zeek,zjson,tzng]")
+	f.StringVar(&c.protocol, "p", "zng", "protocol to use for search request [ndjson,zjson,zng]")
+	f.StringVar(&c.dir, "d", "", "directory for output data files")
+	f.StringVar(&c.outputFile, "o", "", "write data to output file")
+	f.BoolVar(&c.stats, "S", false, "display search stats on stderr")
+	f.BoolVar(&c.warnings, "W", false, "display warnings on stderr")
```
alfred-landrum, line 56: I think this be true by default.

cmd/zapi/cmd/get/get.go
```diff
@@ -0,0 +1,184 @@
...
+	if err != nil {
+		return err
+	}
+	stream := api.NewZngSearch(r) // XXX fix this to match specified format
+	stream.SetOnCtrl(c.handleControl)
+	if err := zbuf.Copy(zbuf.NopFlusher(writer), stream); err != nil {
+		writer.Close()
+		if c.Context().Err() != nil {
+			fmt.Fprintln(os.Stderr, "search aborted")
+			return nil
```
alfred-landrum, line 111: If Err() != nil, I think that means the search was canceled due to a signal. Even in that case, the command should return a non-zero exit status, which I think means returning something other than `nil` here?

cmd/zapi/cmd/get/get.go
```diff
@@ -0,0 +1,184 @@
...
+		}
+		defer file.Close()
+		w = file
+	}
+	_, err := io.Copy(w, r)
+	return err
+}
+
+func fromTo(ctx context.Context, client *api.Connection, id api.SpaceID) (nano.Span, error) {
+	//XXX for now we run the search over the entire space
```
alfred-landrum, line 166: I think the ability to specify the time range should go in this PR.

zqd/handlers.go
```diff
@@ -116,7 +116,7 @@ func getSearchOutput(w http.ResponseWriter, r *http.Request) (search.Output, err
 	}
 	format := r.URL.Query().Get("format")
 	switch format {
-	case "zjson", "json":
+	case "zjson", "ndjson":
```
alfred-landrum, line 119: 👍

pkg/repl/repl.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func NewREPL(c Consumer) *REPL {
+	return &REPL{consumer: c}
+}
+
+// Run executes the REPL.
+func (r *REPL) Run() error {
+	l := liner.NewLiner()
+	defer l.Close() //nolint:errcheck
```
alfred-landrum, line 27: I don't think this nolint command is looked at by our build tools?

pkg/repl/repl.go
```diff
@@ -0,0 +1,71 @@
...
+	l.SetMultiLineMode(true)
+	for {
+		line, e := l.Prompt(r.consumer.Prompt())
+		if e == io.EOF {
+			return io.EOF
+		} else if e != nil {
+			// Ignore this error; the prior is more interesting.
```
alfred-landrum, line 34: I don't understand this comment.

pkg/repl/repl.go
```diff
@@ -0,0 +1,71 @@
...
+	return &REPL{consumer: c}
+}
+
+// Run executes the REPL.
+func (r *REPL) Run() error {
+	l := liner.NewLiner()
+	defer l.Close() //nolint:errcheck
+	l.SetMultiLineMode(true)
+	for {
+		line, e := l.Prompt(r.consumer.Prompt())
```
alfred-landrum, line 30:
```suggestion
		line, err := l.Prompt(r.consumer.Prompt())
```

cmd/zapi/cmd/get/get.go
mattnibs: Yes it does. I wasn't clear if an abort should be a non-zero exit status. Will change.

cmd/zapi/cmd/signal.go
```diff
@@ -8,29 +8,37 @@ import (
...
-	return
+	s := &signalCtx{signals: signals}
+	s.Reset()
+	return s
+}
+
+func (s *signalCtx) listen() {
+	select {
+	case sig := <-s.signals:
+		s.caught = sig
```
nwt, line 27:
```suggestion
	case s.caught = <-s.signals:
```

pkg/repl/repl.go
```diff
@@ -0,0 +1,71 @@
...
+type REPL struct {
+	consumer Consumer
+}
+
+func NewREPL(c Consumer) *REPL {
+	return &REPL{consumer: c}
+}
+
+// Run executes the REPL.
+func (r *REPL) Run() error {
```
nwt, line 25: Since the only thing you can do with a REPL is call its Run method, you could drop the type and NewREPL and just have a Run function:
```suggestion
func Run(c Consumer) error {
```

cmd/zapi/cmd/get/get.go
mattnibs: I can do that. How would you imagine this working? Just simple `-from` / `-to` args that default to min/max time if not set? Do we want to support relative time (e.g `-last 5mins`)  How crazy do we want to get with parsing human readable time formats (I suppose I could search for some suitable library that does this)?

pkg/repl/repl.go
mattnibs: Nope copy pasta.

cmd/zapi/cmd/get/get.go
mattnibs: Yes. Right now we're only doing zng results over the wire for non-wire dump searches so this comment is unnecessary (hope that made sense!)

pkg/repl/repl.go
mattnibs: Excellent call.

pkg/repl/repl.go
mattnibs: I don't get it either. In fact this whole stanza could be simplified.

cmd/zapi/cmd/get/get.go
alfred-landrum: My suggestion would be to start small: `-from` and/or `-to` that are come from calling `time.Parse` with `RFC3339Nano`.

cmd/zapi/cmd/get/get.go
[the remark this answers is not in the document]
alfred-landrum: I understand now this is required by Emitter, so never mind.

cmd/zapi/cmd/get/get.go
mattnibs: Sounds good

PR #844, opened 2020-06-01
report detailed pcap errors

### LINKED ISSUE #490: Improve "bad pcap file" error message
While we're certain to encounter truly corrupt pcap files in the wild, we've also had situations where we failed to extract flows successfully due to our own bugs or lack of full support for all pcap/pcapng variations. It'd therefore probably be appropriate to adjust the error message to something a little less definite, e.g. "unable to extract from pcap file" or whatever is most accurate based on the thing we're trying & failing, rather than pronouncing judgement on the quality of the file.

While I was looking at this, I noticed that in the code the `ErrCorruptPcap` that results in `bad pcap file` is used in several places. Would it make sense to break these out into separate, more granular errors that would help us narrow down the root cause if a user reports one of these problems to us?

mattnibs:
Previously package pcap was hiding detailed pcap error information
by returning a generic ErrCorruptPcap error.

This pr restructure the way package pcap reports errors. package pcapio
now returns unwrapped errors with specifics as to why a pcap function
failed.

package pcap wraps errors returned from pcapio with ErrInvalidPcap.
As a result the detailed error reporting is preserved.

closes #490

### DIFF
--- a/pcap/index.go
+++ b/pcap/index.go
@@ -43,7 +43,7 @@ type Section struct {
 func CreateIndex(r io.Reader, size int) (Index, error) {
 	reader, err := pcapio.NewReader(r)
 	if err != nil {
-		return nil, err
+		return nil, errInvalidPcap(err)
 	}
 	var offsets []ranger.Point
 	var index Index
@@ -55,7 +55,7 @@ func CreateIndex(r io.Reader, size int) (Index, error) {
 			if err == io.EOF {
 				break
 			}
-			return nil, err
+			return nil, errInvalidPcap(err)
 		}
 		if block == nil {
 			break
@@ -63,7 +63,8 @@ func CreateIndex(r io.Reader, size int) (Index, error) {
 		switch typ {
 		default:
 			if section == nil {
-				return nil, errors.New("missing section header")
+				err = errors.New("missing section header")
+				return nil, errInvalidPcap(err)
 			}
 			slice := slicer.Slice{
 				Offset: off,
@@ -72,9 +73,9 @@ func CreateIndex(r io.Reader, size int) (Index, error) {
 			section.Blocks = append(section.Blocks, slice)
 
 		case pcapio.TypePacket:
-			pkt, ts, _ := reader.Packet(block)
+			pkt, ts, _, err := reader.Packet(block)
 			if pkt == nil {
-				return nil, pcapio.ErrCorruptPcap
+				return nil, errInvalidPcap(err)
 			}
 			y := uint64(ts)
 			offsets = append(offsets, ranger.Point{X: off, Y: y})
@@ -82,7 +83,8 @@ func CreateIndex(r io.Reader, size int) (Index, error) {
 		case pcapio.TypeSection:
 			// end previous section and start a new one
 			if section == nil && offsets != nil {
-				return nil, errors.New("packets found without section header")
+				err = errors.New("missing section header")
+				return nil, errInvalidPcap(err)
 			}
 			if section != nil && offsets != nil {
 				section.Index = ranger.NewEnvelope(offsets, size)
--- a/pcap/index.go
+++ b/pcap/index.go
@@ -43,7 +43,7 @@ type Section struct {
 func CreateIndex(r io.Reader, size int) (Index, error) {
 	reader, err := pcapio.NewReader(r)
 	if err != nil {
-		return nil, errInvalidPcap(err)
+		return nil, err
 	}
 	var offsets []ranger.Point
 	var index Index
@@ -55,7 +55,7 @@ func CreateIndex(r io.Reader, size int) (Index, error) {
 			if err == io.EOF {
 				break
 			}
-			return nil, errInvalidPcap(err)
+			return nil, err
 		}
 		if block == nil {
 			break
@@ -64,7 +64,7 @@ func CreateIndex(r io.Reader, size int) (Index, error) {
 		default:
 			if section == nil {
 				err = errors.New("missing section header")
-				return nil, errInvalidPcap(err)
+				return nil, pcapio.NewErrInvalidPcap(err)
 			}
 			slice := slicer.Slice{
 				Offset: off,
@@ -75,7 +75,7 @@ func CreateIndex(r io.Reader, size int) (Index, error) {
 		case pcapio.TypePacket:
 			pkt, ts, _, err := reader.Packet(block)
 			if pkt == nil {
-				return nil, errInvalidPcap(err)
+				return nil, err
 			}
 			y := uint64(ts)
 			offsets = append(offsets, ranger.Point{X: off, Y: y})
@@ -84,7 +84,7 @@ func CreateIndex(r io.Reader, size int) (Index, error) {
 			// end previous section and start a new one
 			if section == nil && offsets != nil {
 				err = errors.New("missing section header")
-				return nil, errInvalidPcap(err)
+				return nil, pcapio.NewErrInvalidPcap(err)
 			}
 			if section != nil && offsets != nil {
 				section.Index = ranger.NewEnvelope(offsets, size)
--- a/pcap/index.go
+++ b/pcap/index.go
@@ -63,7 +63,7 @@ func CreateIndex(r io.Reader, size int) (Index, error) {
 		switch typ {
 		default:
 			if section == nil {
-				err = errors.New("missing section header")
+				err := errors.New("missing section header")
 				return nil, pcapio.NewErrInvalidPcap(err)
 			}
 			slice := slicer.Slice{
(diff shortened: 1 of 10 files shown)


alfred-landrum:
At quick glance, I think that `package pcap` wraps errors with 'invalid pcap' that should have been added by `package pcapio`.  Just a single example: the errors from NgReader parsePacket could be 'invalid pcap's.

mattnibs:
> At quick glance, I think that `package pcap` wraps errors with 'invalid pcap' that should have been added by `package pcapio`. Just a single example: the errors from NgReader parsePacket could be 'invalid pcap's.

@alfred-landrum my thinking around this is that if pcapio returns an error it will always be around an invalid pcap. You would also then have to wrap every error you return there with invalidPcap which would get cumbersome. You also have to worry about doing double even triple error wraps- for instance on the pcapio.NewReader detector there are three levels of errors that could be wrapped with invalidPcap. It becomes unwieldy.

When you go into package pcap errors are not necessarily having to do with invalid pcaps and the distinction is more important here.

alfred-landrum:
pcapio can return non-invalid errors too, like general io failure, that
should be seen as an 'invalid pcap'

I count only 11 uses of 'errors.New' in package pcapio, and  2 uses of the
existing ErrCorruptPcap, so there aren't that many error sites. And maybe
I'm missing something, but I don't understand why you need to then wrap
errors as you say.

mattnibs:
> like general io failure, that should be seen as an 'invalid pcap'

This is true. Let me take a shot and see if I can make this work.

mattnibs:
@alfred-landrum I took another look and ended up liking your way better. PTAL.

pcap/pcapio/ngread.go
```diff
@@ -44,32 +40,44 @@ func NewNgReader(r io.Reader) (*NgReader, error) {
...
+	}
+	if err := ret.parseSectionMagic(hdr[8:12]); err != nil {
+		return nil, err
+	}
+	typ, block, err := ret.readBlock()
+	if err != nil {
+		return nil, err
+	}
+	if typ != ngBlockTypeSectionHeader {
+		return nil, errInvalidf("Unknown magic %x", typ)
```
nwt, line 61:
```suggestion
		return nil, errInvalidf("unknown magic %x", typ)
```

pcap/pcapio/error.go
```diff
@@ -0,0 +1,43 @@
...
+	"fmt"
+
+	"github.com/brimsec/zq/zqe"
+)
+
+type detectorErr struct {
+	pcap   error
+	ngpcap error
+}
+
```
nwt, line 14:
```suggestion
```

pcap/index.go
```diff
@@ -63,7 +63,8 @@ func CreateIndex(r io.Reader, size int) (Index, error) {
 		switch typ {
 		default:
 			if section == nil {
-				return nil, errors.New("missing section header")
+				err = errors.New("missing section header")
```
nwt, line 66:
```suggestion
				err := errors.New("missing section header")
```

pcap/index.go
```diff
@@ -72,17 +73,18 @@ func CreateIndex(r io.Reader, size int) (Index, error) {
...
+				return nil, err
 			}
 			y := uint64(ts)
 			offsets = append(offsets, ranger.Point{X: off, Y: y})
 
 		case pcapio.TypeSection:
 			// end previous section and start a new one
 			if section == nil && offsets != nil {
-				return nil, errors.New("packets found without section header")
+				err = errors.New("missing section header")
```
nwt, line 86:
```suggestion
				err := errors.New("missing section header")
```

pcap/pcapio/ngread.go
```diff
@@ -44,32 +40,44 @@ func NewNgReader(r io.Reader) (*NgReader, error) {
...
 	}
 	ret.first = block
 	return ret, nil
 }
 
-func (r *NgReader) parsePacket(block []byte) ([]byte, int) {
+func (r *NgReader) parsePacket(block []byte) ([]byte, int, error) {
 	if len(block) < PacketBlockHeaderLen {
-		return nil, 0
+		return nil, 0, errInvalid("packet buffer length less then minimum packet size")
```
nwt, line 69:
```suggestion
		return nil, 0, errInvalid("packet buffer length less than minimum packet size")
```

pcap/pcapio/read.go
```diff
@@ -83,17 +83,17 @@ func NewPcapReader(r io.Reader) (*PcapReader, error) {
 	return reader, nil
 }
 
-func (r *PcapReader) Packet(block []byte) ([]byte, nano.Ts, layers.LinkType) {
+func (r *PcapReader) Packet(block []byte) ([]byte, nano.Ts, layers.LinkType, error) {
 	if len(block) <= packetHeaderLen {
-		return nil, 0, 0
+		return nil, 0, 0, errInvalid("packet buffer length less then minimum packet size")
```
nwt, line 88:
```suggestion
		return nil, 0, 0, errInvalid("packet buffer length less than minimum packet size")
```

pcap/pcapio/reader.go
```diff
@@ -25,22 +28,29 @@ const (
...
-	if err == nil {
+	_, err2 := NewNgReader(track)
+	if err2 == nil {
 		return NewNgReader(recorder)
 	}
-	return nil, ErrCorruptPcap
+	var pcaperr, ngerr *ErrInvalidPcap
+	if errors.As(err1, &pcaperr) && errors.As(err2, &ngerr) {
+		err1 = fmt.Errorf("pcap: %w", pcaperr.err)
+		err2 = fmt.Errorf("ngpcap: %w", ngerr.err)
```
nwt, line 51:
```suggestion
		err2 = fmt.Errorf("pcapng: %w", ngerr.err)
```

zqd/handlers_zeek_test.go
```diff
@@ -152,8 +152,7 @@ func TestPcapPostInvalidPcap(t *testing.T) {
 			t.Fatalf("expected error to be for type *api.ErrorResponse, got %T", p.err)
 		}
 		assert.Equal(t, http.StatusBadRequest, reserr.StatusCode())
-		// XXX Better error message here.
-		require.Regexp(t, "bad pcap file", reserr.Err.Error())
+		require.Regexp(t, "invalid pcap: pcap: Unknown magic 73696874; ngpcap: first block type not a section header: 1936287860", reserr.Err.Error())
```
nwt, line 155:
```suggestion
		require.Regexp(t, "invalid pcap: pcap: unknown magic 73696874; pcapng: first block type not a section header: 1936287860", reserr.Err.Error())
```

mattnibs: whoops, thanks

pcap/pcapio/read.go
```diff
@@ -116,10 +116,10 @@ func (r *PcapReader) readHeader() error {
 		r.byteOrder = binary.BigEndian
 		r.nanoSecsFactor = 1000
 	} else {
-		return fmt.Errorf("Unknown magic %x", magic)
+		return errInvalidf("Unknown magic %x", magic)
```
nwt, line 119:
```suggestion
		return errInvalidf("unknown magic %x", magic)
```

pcap/pcapio/read.go
```diff
@@ -116,10 +116,10 @@ func (r *PcapReader) readHeader() error {
...
 	}
 	if r.versionMajor = r.byteOrder.Uint16(hdr[4:6]); r.versionMajor != versionMajor {
-		return fmt.Errorf("Unknown major version %d", r.versionMajor)
+		return errInvalidf("Unknown major version %d", r.versionMajor)
```
nwt, line 122:
```suggestion
		return errInvalidf("unknown major version %d", r.versionMajor)
```

pcap/pcapio/read.go
```diff
@@ -116,10 +116,10 @@ func (r *PcapReader) readHeader() error {
...
 	}
 	if r.versionMinor = r.byteOrder.Uint16(hdr[6:8]); r.versionMinor != versionMinor {
 		return fmt.Errorf("Unknown minor version %d", r.versionMinor)
```
nwt, line 125: s/Unknown/unknown/

pcap/pcapio/error.go
```diff
@@ -0,0 +1,38 @@
...
+type ErrInvalidPcap struct {
+	err error
+}
+
+func NewErrInvalidPcap(err error) error {
+	e := &ErrInvalidPcap{err: err}
+	return zqe.E(zqe.Invalid, e)
+}
+
+func errInvalid(text string) error {
```
alfred-landrum, line 19: Nit: consider removing `errInvalid(...)` and use `errInvalidf(...)` everywhere.

mattnibs: sure!

mattnibs:
> Is this PR's title what you want it to be?

Nope. Changed.

PR #848, opened 2020-06-02
Remove special case time-binning support from groupby proc

### LINKED ISSUE #828: groupby: remove special-cased time bin handling
Once #535 is complete, we can get rid of the special-case time bin handling in the groupby proc by adding zql sugar that turns `-every X` into a groupby key `ts=Time.trunc(ts, X)`. 

This includes generalizing the "early timebin output" feature that is present in every-groupby, which outputs time-bins as soon as input record time has advanced past a bin. This early-emit will be controlled by a new `-sorted` flag that will tell groupby that the input is sorted on the primary groupby key, and that it is therefore safe to do early-emit.

### LINKED ISSUE #847: groupby "every" assumes forward sorted data
If records presented to a `every X` groupby are not in forward order, then that groupby will emit multiple groups for a given time bin. For example:

```
$ zq -t "every 1h count()" ~/Downloads/sampledata/wrccdc-year1/zng-logs/conn.zng | head -n 10
0:[1490382000;88795;]
0:[1490382000;100;]
0:[1490382000;100;]
0:[1490382000;100;]
0:[1490382000;92;]
0:[1490382000;98;]
0:[1490382000;100;]
0:[1490382000;92;]
0:[1490382000;100;]
```

henridf:
(currently based on groupby-computed-keys branch, and includes the commit of the time-trunc branch... will rebase as those get merged)

Now that zql can group by computed keys, `every X reducer() by field` can be computed as `reducer by ts=Time.trunc("ts", X), field`.

This PR makes that change by converting the Duration AST node (resulting from `every` in zql) into a new groupby key representing the time trunc expression.

There are still some behavioral differences brought by this PR:
1. Results for a time bin are no longer emitted as soon as record timestamps have advanced past that. Instead, they are emitted all at once on eof. This fixes #847.
2. Results are not sorted by ts. This is because they are now handled as regular keys and so fall under the "Records and spans generated by time-binning are partially ordered by timestamp coincident with search direction." groupby contract.

This second point needs some discussion and is the reason I've put this up as a draft PR. Options I see:

- Go with this new behavior and like for other groupby keys, require an explicit `sort ts` after a time-binned groupby where needed.
- Implicitly insert a `sort ts` proc after a time-binned groupby to maintain prior behavior.

I'm in favor of the first option. (I believe this is what @nwt has been advocating).
 
Also, further out, we may want to do "early emitting" of completed bins in a more general way, for _any_ key that is known to be sorted at the input point. That might be a good point to revisit this question and possibly auto-sort on timebinned groupby outputs on time-sorted streams. 

closes #828

### DIFF
--- a/proc/groupby.go
+++ b/proc/groupby.go
@@ -2,6 +2,7 @@ package proc
 
 import (
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"sort"
 	"time"
@@ -15,7 +16,6 @@ import (
 	"github.com/brimsec/zq/zcode"
 	"github.com/brimsec/zq/zng"
 	"github.com/brimsec/zq/zng/resolver"
-	"github.com/pkg/errors"
 	"go.uber.org/zap"
 )
 
--- a/proc/groupby.go
+++ b/proc/groupby.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"sort"
 	"strconv"
-	"time"
 
 	"github.com/brimsec/zq/ast"
 	"github.com/brimsec/zq/expr"
@@ -26,7 +25,6 @@ type GroupByKey struct {
 }
 
 type GroupByParams struct {
-	duration ast.Duration
 	limit    int
 	keys     []GroupByKey
 	reducers []compile.CompiledReducer
@@ -63,7 +61,6 @@ func CompileGroupBy(node *ast.GroupByProc, zctx *resolver.Context) (*GroupByPara
 			},
 		}
 		astKeys = append([]ast.Assignment{everyKey}, astKeys...)
-		node.Duration.Seconds = 0
 	}
 
 	for _, astKey := range astKeys {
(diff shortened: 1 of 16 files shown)


proc/groupby.go
```diff
@@ -135,59 +147,44 @@ type GroupByAggregator struct {
...
-		logger:          c.Logger,
-		limit:           limit,
+		keys:        params.keys,
+		zctx:        c.TypeContext,
+		kctx:        resolver.NewContext(),
+		reducerDefs: params.reducers,
+		builder:     params.builder,
+		keyRows:     make(map[int]keyRow),
+		table:       make(map[string]*GroupByRow),
+		reverse:     c.Reverse,
```
alfred-landrum, line 175: Do we still need a reverse flag with these changes?

henridf: The reverse flag wasn't properly handled (#847) and per the comment my proposal (to be debated) is to let groupby output time bins in "unsorted  but deterministic" order like for other keys...

henridf:
coming soon for real

PR #854, opened 2020-06-04
support archive based subspaces

### LINKED ISSUE #590: adhoc space creation for index matching chunks

alfred-landrum:
Add apis to support archive backed spaces and subspaces. 

The space creation api is extended to explicitly support archive backed spaces. A new space can point to an existing archive root, or can specify the creation of a new archive. However, there's not yet support for an archive space to reflect changes made via a separate `zar import` to the same root, but that's easy and will come shortly.

The api to create a subspace of an archive store backed space is:
```
POST /space/{parent_space_id}/subspace
{
"name": "<name>",
"open_options": {
  "log_filter": ["logid1", ...]
  },
}
```
The response is the same as the existing space post request. The subspace supports index and zql searches, and changing its name, via the existing apis.

Closes #590


alfred-landrum:
FYI that I've succesfully run the Brim integration tests locally against these changes.

cmd/zapi/cmd/new/new.go
```diff
@@ -16,20 +17,27 @@ var New = &charm.Spec{
...
+type NewCommand struct {
 	*cmd.Command
+	kind     string
+	datapath string
 }
 
-func (c *Command) Run(args []string) error {
+func NewFn(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
+	c := &NewCommand{Command: parent.(*cmd.Command)}
+	f.StringVar(&c.kind, "k", string(storage.FileStore), "kind of storage for this space")
```
mattnibs, line 35: Consider having `storage.Kind` implement `flag.Value` so you can return an error if an invalid option is specified.

zqd/storage/filestore/filestore.go
```diff
@@ -180,9 +179,8 @@ func (s *Storage) UnionSpan(span nano.Span) error {
 
 }
 
-// XXX This is not thread safe and it should be.
```
mattnibs, line 183: This comment is still relevant.

zqd/storage/filestore/filestore.go
```diff
@@ -165,11 +165,10 @@ func (s *Storage) Clear(ctx context.Context) error {
 	if err := os.Remove(s.join(allZngFile)); err != nil && !os.IsNotExist(err) {
 		return err
 	}
-	return s.UnsetSpan()
+	return s.SetSpan(nano.Span{})
 }
 
-// XXX This is not thread safe and it should be.
```
mattnibs, line 171: This comment is still relevant.

zqd/space/space.go
```diff
@@ -32,232 +29,180 @@ var (
...
+		case <-g.cancelChan:
 			cancel()
 		}
 	}()
 
 	done := func() {
-		s.opMutex.Lock()
-		defer s.opMutex.Unlock()
+		g.opMutex.Lock()
+		defer g.opMutex.Unlock()
```
mattnibs, line 91: I don't believe this mutex is necessary.

zqd/handlers.go
```diff
@@ -274,7 +311,17 @@ func handlePcapPost(c *Core, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	op, err := ingest.NewPcapOp(ctx, s, req.Path, c.ZeekLauncher)
+	pspace, ok := s.(ingest.PcapSpace)
+	if !ok {
+		respondError(c, w, r, zqe.E(zqe.Invalid, "space does not support pcap import"))
```
mattnibs, line 316: Can we change line 255 to have the same error message?

zqd/handlers.go
```diff
@@ -274,7 +311,17 @@ func handlePcapPost(c *Core, w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	pstore, ok := s.Storage().(ingest.PcapStore)
+	if !ok {
+		respondError(c, w, r, zqe.E(zqe.Invalid, "storage does not support pcap import"))
+		return
+	}
```
mattnibs, line 323: From first glance and without digging deeper it feels odd that PcapSpace wouldn't have store that implements PcapStore and therefore render this check unnecessary.

archive/schema.go
```diff
@@ -120,7 +128,11 @@ func (ark *Archive) AppendSpans(spans []SpanInfo) error {
 	return ark.Meta.Write(filepath.Join(ark.Root, metaDataFilename))
 }
 
-func OpenArchive(path string) (*Archive, error) {
+type OpenOptions struct {
+	LogFilter []string
```
mattnibs, line 132: Shouldn't this be?
```suggestion
	LogFilter []LogID
```

zqd/space/archivespace.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+	return os.RemoveAll(s.conf.DataPath)
+}
+
+func (s *archiveSpace) CreateSubspace(req api.SubspacePostRequest) (*archiveSubspace, error) {
+	s.confMutex.Lock()
+	defer s.confMutex.Unlock()
+
+	substore, err := archivestore.Load(s.conf.DataPath, &storage.ArchiveConfig{
+		OpenOptions: &req.OpenOptions,
```
mattnibs, line 70: Should we return an error invalid request if OpenOptions is nil?

zqd/space/archivespace.go
```diff
@@ -0,0 +1,158 @@
...
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	subcfg := subspaceConfig{
+		ID:          newSpaceID(),
+		Name:        req.Name,
```
mattnibs, line 78: Should subspaces be allowed to have empty Name values?

mattnibs: Looks like we do a check for empty name values in Update... we should do one here as well.

zqd/handlers.go
mattnibs: Oh I guess you do because a log filespace would pass as a PcapSpace but fail as a PcapStorage.

zqd/search/pcapsearch.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/brimsec/zq/pcap"
+	"github.com/brimsec/zq/pcap/pcapio"
+	"github.com/brimsec/zq/pkg/fs"
+	"github.com/brimsec/zq/zqd/api"
+)
+
+type PcapSpace interface {
+	PcapIndexPath() string
+	PcapPath() string
+}
```
mattnibs, line 17: It would be great to consolidate these interfaces in one location- it would make this clearer for me. That said I'm not quite sure where that one location should be. A lot is changing around this I suppose it would be best to wait and see how this shakes out.

zqd/space/filespace.go
```diff
@@ -0,0 +1,84 @@
...
+	"context"
+	"errors"
+	"os"
+	"path/filepath"
+
+	"github.com/brimsec/zq/zqd/api"
+	"github.com/brimsec/zq/zqe"
+)
+
+const PcapIndexFile = "packets.idx.json"
```
mattnibs, line 13: This can be made private now.

zqd/space/space.go
```diff
@@ -32,232 +29,180 @@ var (
...
+			}
+			ret = append(ret, sub)
+		}
+		return ret, nil
 
-func (c *SearchReadCloser) Close() error {
-	return c.f.Close()
+	default:
+		return nil, zqe.E(zqe.Invalid, "loadSpace: unknown storage kind: %s", conf.Storage.Kind)
+	}
```
mattnibs, line 164: We're going to need a migration for old spaces here as they will run into this error on load- their config will have a zero value for StorageConfig and thus an Unknown value for Storage kind.

zqd/space/manager.go
```diff
@@ -44,37 +45,49 @@ func NewManager(root string, logger *zap.Logger) (*Manager, error) {
 			continue
 		}
 
-		space, err := loadSpace(path, config)
+		spaces, err := loadSpaces(path, config)
```
mattnibs, line 48: Because subspaces do not have configs stored in their path you will get a logger error "Error loading config" for every subspace on manager open. We could change `loadConfig` to not log an error if config file DoesNotExist, that said I think my preference would be for a subspace to have its config in its path and for a subspace config to include a reference to its parentId. One of the nice things about this solution is if I needed to debug a space and all I had was the file system I could at least cat a space's config file and have some idea what it does.

Did not quite know where to put this comment: The fact that a sub-spaces configs are stored in its parent is a little strange to me. With this setup there will be

zqd/space/space.go
alfred-landrum: thanks, I believe this was left over from when there was an 'active' integer counter to track the number of outstanding operations, will remove.

zqd/handlers.go
alfred-landrum: The line 255 I see is for the warning about zeek not being present; I'd rather keep that in place since it describes a different scenario (zqd not having a configured zeek vs an individual space not capable of handling pcaps).

mattnibs: ok

zqd/handlers.go
alfred-landrum: In this PR, I did not try to move all of the pcap specific operations from what was a 'Space' to the filestore, hence this split. It seems the me the only complication to doing that is that we track the location of the pcapfile in the space's configuration file currently, and we probably wouldn't want the store to try to write to that.

zqd/handlers_test.go
```diff
@@ -634,7 +634,7 @@ func TestDeleteDuringPcapPost(t *testing.T) {
 	err = client.SpaceDelete(context.Background(), sp.ID)
 	require.NoError(t, err)
 
-	require.Error(t, <-pcapPostErr, "context canceled")
+	require.Regexp(t, "context canceled", (<-pcapPostErr).Error())
```
nwt, line 637: More specific and clearer to do this:
```suggestion
	require.EqualError(t, <-pcapPostErr, "context canceled")
```

archive/find_test.go
```diff
@@ -0,0 +1,97 @@
...
+	thresh := int64(1000)
+	createArchiveSpace(t, datapath, "../tests/suite/zdx/babble.tzng", &CreateOptions{
+		LogSizeThreshold: &thresh,
+	})
+
+	_, err = OpenArchive(datapath, &OpenOptions{
+		LogFilter: []string{"foo"},
+	})
+	require.Error(t, err)
+	require.Regexp(t, "no spans", err.Error())
```
nwt, line 68:
```suggestion
	require.EqualError(t, err, "no spans")
```

zqd/space/space.go
```diff
@@ -32,232 +29,180 @@ var (
...
+	ID() api.SpaceID
+	Storage() storage.Storage
+	Info(context.Context) (api.SpaceInfo, error)
+
+	// StartOp is called to register an operation is in progress; the
+	// returned cancel function must be called when the operation is done.
+	StartOp(context.Context) (context.Context, context.CancelFunc, error)
+
+	// Delete cancels any outstanding operations, then removes the space's path
+	// and data dir (should the data dir be different then the space's path).
```
nwt, line 42:
```suggestion
	// and data dir (should the data dir be different than the space's path).
```

zqd/space/space.go
alfred-landrum: We won't; loadConfig handles this by putting `storage.FileStore` into the Kind field if it's empty.  I also verified that I was able to create spaces with zqd on master & then load them with this PR.

zqd/space/manager.go
alfred-landrum: You're correct that a does subspace in this PR does not have a config.json file for itself. Note that they also don't have a separate directory under zqd's root, so the error you mention won't occur.

The reason I went this route is to avoid needing to update file-based configuration in two places when you create a subspace, and to help prevent a parent space from being deleted when there are subspaces that refer to it.

zqd/space/filespace.go
alfred-landrum: turns out it's still referenced by a test in handlers_zeek_test.go

archive/find_test.go
alfred-landrum: the actual error text here is longer, so would just keep the regex

archive/schema.go
alfred-landrum: I'd like to use archive.LogID here, but with our current package layout, that would lead to a circular dependency.

zqd/space/archivespace.go
alfred-landrum: req.OpenOptions isn't a pointer, and the archivestore.Load function will fail if the contents aren't valid.

zqd/search/pcapsearch.go
alfred-landrum: Yes - in another comment, I mentioned that we should be able to move the pcap handling functionality off of the space & onto the store.

zqd/space/archivespace.go
mattnibs: I suppose what I meant to say is can you create a subspace with an empty array for LogFilter?

zqd/space/manager.go
mattnibs: > Note that they also don't have a separate directory under zqd's root, so the error you mention won't occur.

I was missing this part. This sounds good then.

zqd/ingest/pcap.go
```diff
@@ -49,17 +51,9 @@ type PcapOp struct {
...
-		return nil, ErrNoPcapImport
-	}
-	logdir := s.DataPath(tmpIngestDir)
-	if err := os.Mkdir(logdir, 0700); err != nil {
-		if os.IsExist(err) {
-			// could be in use by pcap or log ingest
-			return nil, ErrIngestProcessInFlight
-		}
+func NewPcapOp(ctx context.Context, pspace PcapSpace, pstore PcapStore, pcap string, zlauncher zeek.Launcher) (*PcapOp, error) {
+	logdir, err := ioutil.TempDir("", "zqd-pcap-import-")
```
nwt, line 55:
```suggestion
	logdir, err := ioutil.TempDir("", "zqd-pcap-ingest-")
```

zqd/search/indexsearch.go
```diff
@@ -6,26 +6,22 @@ import (
...
 }
 
-func NewIndexSearch(ctx context.Context, s *space.Space, req api.IndexSearchRequest) (*IndexSearch, error) {
-	arkstore, ok := s.Storage.(*archivestore.Storage)
-	if !ok {
-		return nil, zqe.E(zqe.Invalid, "index search only supported on archive spaces")
-	}
+type IndexedArchive interface {
+	IndexSearch(context.Context, archive.IndexQuery) (zbuf.ReadCloser, error)
+}
```
nwt, line 17: Nits: I think IndexSearcher would be a more descriptive name for this interface. Also, declare it before IndexSearch so IndexSearch remains adjacent to its constructor, NewIndexSearch.

zqd/search/pcapsearch.go
```diff
@@ -0,0 +1,69 @@
...
+		f.Close()
+		return nil, err
+	}
+	return &PcapSearchReadCloser{r, f}, nil
+}
+
+type PcapSearchReadCloser struct {
+	*pcap.SearchReader
+	f *os.File
+}
```
nwt, line 65: Nits: PcapSearchReadCloser reads like an interface name. I think PcapSearch would be a better name (even if it does collide with api.PcapSearch). Also, declare this before its constructor, NewPcapSearch.

zqd/search/search.go
```diff
@@ -35,7 +34,11 @@ type Search struct {
 	io.Closer
 }
 
-func NewSearch(ctx context.Context, s *space.Space, req api.SearchRequest) (*Search, error) {
+type SearchStore interface {
+	Open(ctx context.Context, span nano.Span) (zbuf.ReadCloser, error)
+}
```
nwt, line 39: Nit: Declare this before Search so Search remains adjacent to its constructor.

zqd/space/archivespace.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/brimsec/zq/zqd/storage"
+	"github.com/brimsec/zq/zqd/storage/archivestore"
+	"github.com/brimsec/zq/zqe"
+)
+
+type archiveSpace struct {
+	spaceBase
+
+	// confMutex protects changes to configuration changes.
+	confMutex sync.Mutex
```
nwt, line 20: Nit: Our conventional name for a sync.Mutex or sync.RWMutex field is mu or whateverMu, so call this confMu. You might also consider grouping it separately with the conf field to emphasize that it protects conf but not path.

zqd/space/archivespace.go
```diff
@@ -0,0 +1,162 @@
...
+	s.conf.Name = req.Name
+	return s.conf.save(s.path)
+}
+
+func (s *archiveSpace) delete() error {
+	s.confMutex.Lock()
+	defer s.confMutex.Unlock()
+
+	if len(s.conf.Subspaces) != 0 {
+		return zqe.E(zqe.Conflict, "unable to delete space with subspaces")
```
nwt, line 53: For consistency with other error message in this PR, which begin with "cannot":
```suggestion
		return zqe.E(zqe.Conflict, "cannot delete space with subspaces")
```

zqd/space/archivespace.go
```diff
@@ -0,0 +1,162 @@
+package space
```
nwt, line 1: Nit: I think archive.go would be a better file name since this file defines both archiveSpace and archiveSubspace.

zqd/space/archivespace.go
```diff
@@ -0,0 +1,162 @@
...
+	return &archiveSubspace{
+		spaceBase: spaceBase{subcfg.ID, substore, newGuard()},
+		parent:    s,
+	}, nil
+}
+
+type archiveSubspace struct {
+	spaceBase
+	parent *archiveSpace
+}
```
nwt, line 99: Nit: I think you should either put this type and its methods in archivesubspace.go (my preference) or rename this file to archive.go.

zqd/space/archivespace.go
```diff
@@ -0,0 +1,162 @@
...
+	s.parent.confMutex.Lock()
+	defer s.parent.confMutex.Unlock()
+
+	for i := range s.parent.conf.Subspaces {
+		if s.parent.conf.Subspaces[i].ID == s.id {
+			return fn(i)
+		}
+	}
+	// should not happen
+	return fmt.Errorf("subspace %s cant find conf in parent %s", s.id, s.parent.id)
```
nwt, line 161:
```suggestion
	return fmt.Errorf("subspace %s cannot find conf in parent %s", s.id, s.parent.id)
```

zqd/space/manager.go
```diff
@@ -86,15 +99,34 @@ func (m *Manager) Create(name, dataPath string) (*Space, error) {
...
-	return sp, nil
+	if len(spaces) != 1 {
+		panic("multiple spaces created during space create")
+	}
+	m.spaces[id] = spaces[0]
+	return spaces[0], nil
 }
 
-func (m *Manager) Get(id api.SpaceID) (*Space, error) {
+func (m *Manager) SubspaceCreate(parent Space, req api.SubspacePostRequest) (Space, error) {
```
nwt, line 113: Nit: Call this CreateSubspace for consistency with (*archiveSpace).CreateSubspace.

zqd/search/indexsearch.go
alfred-landrum: as discussed 1x1, going to use IndexSearcher for the interface name & IndexSearchOp for the struct.

PR #863, opened 2020-06-05
detector.MultiFileReader

### LINKED ISSUE #832: archive search should open zng files on demand
Opening all the matching zng files for a search is convenient for use with a Combiner, but will cause all archive files to be opened at search start, which won't scale.

mattnibs:
MultiReader returns a zbuf.ReadCloser that's the logical concatenation of the
provided input paths. They're read sequentially. Once a stream has reached EOF,
Close is called on the file. Once all inputs have reached end of stream,
Read will return end of stream. If any of the readers return a non-nil error,
Read will return that error.

The primary use case is for reading large amounts of files but
limiting the amount of open file descriptors.

Use MultiFileReader when doing zar searches in zqd as well as
performing searches via the zar command.

closes #832

### DIFF
--- a/zio/detector/file.go
+++ b/zio/detector/file.go
@@ -15,6 +15,7 @@ import (
 	"github.com/brimsec/zq/zbuf"
 	"github.com/brimsec/zq/zio/ndjsonio"
 	"github.com/brimsec/zq/zio/parquetio"
+	"github.com/brimsec/zq/zng"
 	"github.com/brimsec/zq/zng/resolver"
 
 	"github.com/xitongsys/parquet-go-source/local"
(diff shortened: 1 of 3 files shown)


zio/detector/file.go
```diff
@@ -168,3 +169,55 @@ func OpenFiles(zctx *resolver.Context, dir zbuf.RecordCmpFn, paths ...string) (*
...
+type inOrderFileReader struct {
+	reader *zbuf.File
+	zctx   *resolver.Context
+	paths  []string
+	cfg    OpenConfig
+}
+
+// InOrderFileReader is a zbuf.ReadCloser compliant reader
+// that takes an array paths and in-order reads all records in each
+// file.
```
henridf, line 182:
```suggestion
// SequentialFileReader returns a zbuf.ReadCloser that reads the records from the provided slice of files sequentially.
```

(and along with this suggestion, rename InOrderFileReader to SequentialFileReader)

mattnibs: sounds good

nwt: Given the behavioral similarity to io.MultiReader, I think we should call this MultiFileReader.

mattnibs: Oooh kind of nice. @henridf what do you think?

mattnibs: I made zbuf.MultiReader at one point but got rid of it.

henridf: Fancy!

mattnibs: In this thread it just keeps on getting better!

zio/detector/file.go
```diff
@@ -168,3 +169,54 @@ func OpenFiles(zctx *resolver.Context, dir zbuf.RecordCmpFn, paths ...string) (*
...
+	reader *zbuf.File
+	zctx   *resolver.Context
+	paths  []string
+	cfg    OpenConfig
+}
+
+// SequentialFileReader returns a zbuf.ReadCloser that reads the records from
+// the provided slice of files sequentially.
+// The primary use case is for reading large amounts of files while
+// limiting the amount of open file descriptors.
```
alfred-landrum, line 183:
```suggestion
// SequentialFileReader returns a zbuf.ReadCloser that reads records from
// the provided slice of files sequentially.
```

zqd/storage/archivestore/archivestore.go
```diff
@@ -32,32 +31,18 @@ func (s *Storage) NativeDirection() zbuf.Direction {
...
-		readers = append(readers, zbuf.NewFile(r, f, f.Name()))
 		return nil
 	})
 	if err != nil {
 		return nil, err
 	}
-	combiner := zbuf.NewCombiner(readers, zbuf.RecordCompare(s.NativeDirection()))
-	return combiner, nil
+	zctx := resolver.NewContext()
+	return detector.SequentialFileReader(zctx, paths, detector.OpenConfig{}), nil
```
alfred-landrum, line 45:
```suggestion
	return detector.SequentialFileReader(zctx, paths, detector.OpenConfig{Format: "zng"}), nil
```

alfred-landrum:
I think the final name suggestion was `MultiFileReader`, when you make that change in the code would you also edit the PR title & first comment?

zio/detector/file.go
```diff
@@ -168,3 +169,54 @@ func OpenFiles(zctx *resolver.Context, dir zbuf.RecordCmpFn, paths ...string) (*
...
 }
+
+type multiFileReader struct {
+	reader *zbuf.File
+	zctx   *resolver.Context
+	paths  []string
+	cfg    OpenConfig
+}
+
+// MultiReader returns a zbuf.ReadCloser that's the logical concatenation of the
```
alfred-landrum, line 180: 👍

nwt, line 180: s/MultiReader/MultiFileReader/

mattnibs: Darn, will file a followup fix.

Issue #855, opened 2020-06-04
zar import should update zqd archive space results


alfred-landrum:
Following up on https://github.com/brimsec/zq/pull/854 : if zqd has an archive-backed space, we should be able to run `zar import` against the archive root, and new searches/request should reflect any newly imported data.

Issue #862, opened 2020-06-05
zapi and zqd should support unix sockets


alfred-landrum:
We'd like to write script-style tests that have a per-test zqd instance, and then interact with it via zapi. However, with the current capabilities, we'd need to handle port rendezvous as well; just using our default port of 9867 will likely interfere with a dev running zqd/Brim on their machine.

We could get around this by supporting unix sockets. We can alter zapi to take a url for a server location, and it should handle urls in the form of `unix:///a/b/c/foo.sock` . zqd would then need to be able to listen & accept connections from a specified socket file.

FYI that resty (the http library we use), supports unix sockets:
https://github.com/go-resty/resty#unix-socket

mccanne:
There are better ways to handle port collision.

PR #864, opened 2020-06-05
zqd archive spaces reflect new data added via zar import

alfred-landrum:
On top of https://github.com/brimsec/zq/pull/854 : With these changes, I can run a local demo where I can run Brim from master, then use zapi to create an archive space, import data into the archive space with zar, and then see the resulting records in Brim. I can then use the zapi newsubspace command from https://github.com/brimsec/zq/pull/895 to create a subspace that is viewable inside Brim as well.

When https://github.com/brimsec/zq/issues/862 is complete, we should be able to use script style tests that use zapi, zar, and zqd to perform those steps.


alfred-landrum:
Closes #855

archive/archive_test.go
```diff
@@ -0,0 +1,166 @@
...
+0:[336;20200422/1587517412.06741443.zng;]
+0:[336;20200421/1587508871.06471174.zng;]
+`
+	out := indexQuery(t, ark1, query, AddPath(DefaultAddPathField, false))
+	require.Equal(t, test.Trim(exp), out)
+
+	ark2, err := OpenArchive(datapath, &OpenOptions{
+		LogFilter: []string{"20200422/1587517412.06741443.zng"},
+	})
+
```
henridf, line 96:
```suggestion
require.NoError(t, err)
```

archive/archive_test.go
```diff
@@ -0,0 +1,166 @@
...
+	require.NoError(t, err)
+
+	importTestFile(t, ark2, "testdata/td2.zng")
+
+	// Verify that the data appears to the earlier opened handle
+	var postSpans []SpanInfo
+	err = SpanWalk(ark1, func(si SpanInfo, _ string) error {
+		postSpans = append(postSpans, si)
+		return nil
+	})
```
henridf, line 152:
```suggestion
	})
	require.NoError(t, err)
```

archive/schema.go
```diff
@@ -107,25 +123,97 @@ func (c *CreateOptions) toMetadata() *Metadata {
 }
 
 type Archive struct {
-	Meta *Metadata
-	Root string
+	Root              string
+	DataSortDirection zbuf.Direction
```
henridf, line 127: I think this is a good thing to do regardless, but is replacing `Archive.Meta` by  `Archive{DataSortDirection,LogSizeThreshold}`

alfred-landrum: On an earlier version of this change, I had a bug where I used the wrong spans slice -  `Archive.Spans` (for filtered spans) vs `Archive.Meta.Spans` (for all the available spans). So I wanted to remove the metadata pointer from `Archive` to prevent tripping up on that again.

PR #869, opened 2020-06-07
zql: remove special-case timebin support

### LINKED ISSUE #828: groupby: remove special-cased time bin handling

### LINKED ISSUE #847: groupby "every" assumes forward sorted data

henridf:
This PR removes the special-case "timebinning" support in the groupby proc, using the new computed groupby key facility introduced in #860. The zql `every` syntax remains such that `every x count() by y` is effectively translated to `count() by ts=time.Trunc(ts, x), y`.

The prior timebinning assumed that input was always forward-sorted in time, and it would output time-bin keys as soon as time (as observed on input records) had advanced past a bin. This assumption is a remnant of a prior system (when we believed that "time was on our side") where records _were_ guaranteed to be time-sorted, but the guarantee is not present in zq or zqd, and so the assumption could lead to silently incorrect results.

Nonetheless, the early emit facility is useful, and has been reimplemented, in a general way that is no specific to time bins. It is enabled by passing the `-sorted <dir>` argument to groupby (where `dir` should be `1` or `-1` to indicate the input sort order), which signals that the input is sorted in the primary key.

For example: `every 1h count() by _path -sorted 1` could be used to tell groupby that input records are forward-ordered by time and can be early-emitted.

closes #828 
closes #847

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -256,15 +256,15 @@ type (
 	// A GroupByProc node represents a proc that consumes all the records
 	// in its input, partitions the records into groups based on the values
 	// of the fields specified in the keys field (where the first key is the
-	// primary key), and applies reducers (if any) to each group. If the
-	// duration field is non-zero, then the groups further partioned by time
-	// into bins of the duration. In this case time is the primary key.
+	// primary grouping key), and applies reducers (if any) to each group. If the
+	// Duration field is non-zero, then the groups are further partioned by time
+	// into bins of the duration. In this case, the primary grouping key is ts.
 	// The Limit field specifies the number of different groups that can be
 	// aggregated over. When absent, the runtime defaults to an
 	// appropriate value.
 	// The InputSortDir field indicates that input is sorted (with
-	// direction indicated by the sign of the field) in the
-	// primary key. In this case, the proc outputs the reducer
+	// direction indicated by the sign of the field) in the primary
+	// grouping key. In this case, the proc outputs the reducer
 	// results from each key as they complete so that large inputs
 	// are processed and streamed efficiently.
 	GroupByProc struct {
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -255,23 +255,25 @@ type (
 	}
 	// A GroupByProc node represents a proc that consumes all the records
 	// in its input, partitions the records into groups based on the values
-	// of the fields specified in the keys paramater, and applies one or
-	// more reducers to each group.  If the duration parameter is non-zero,
-	// then the groups further partioned by time according to the time interval
-	// equal to the duration.  In this case, the proc transmits to its output
-	// the reducer results from each time interval as they complete so that
-	// large time ranges are processed and streamed efficiently.
-	// The limit parameter specifies the number of different groups that can
-	// be aggregated over. When absent, the runtime defaults to an appropriate value.
-	//
-	// xxx comment sorted
+	// of the fields specified in the keys field (where the first key is the
+	// primary key), and applies reducers (if any) to each group. If the
+	// duration field is non-zero, then the groups further partioned by time
+	// into bins of the duration. In this case time is the primary key.
+	// The Limit field specifies the number of different groups that can be
+	// aggregated over. When absent, the runtime defaults to an
+	// appropriate value.
+	// The InputSortDir field indicates that input is sorted (with
+	// direction indicated by the sign of the field) in the
+	// primary key. In this case, the proc outputs the reducer
+	// results from each key as they complete so that large inputs
+	// are processed and streamed efficiently.
 	GroupByProc struct {
 		Node
-		Duration Duration     `json:"duration"`
-		Limit    int          `json:"limit,omitempty"`
-		Keys     []Assignment `json:"keys"`
-		Reducers []Reducer    `json:"reducers"`
-		Sorted   int          `json:"sorted"`
+		Duration     Duration     `json:"duration"`
+		Limit        int          `json:"limit,omitempty"`
+		Keys         []Assignment `json:"keys"`
+		Reducers     []Reducer    `json:"reducers"`
+		InputSortDir int          `json:"input_sort_dir"`
 	}
 	// TopProc is similar to proc.SortProc with a few key differences:
 	// - It only sorts in descending order.
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -263,12 +263,15 @@ type (
 	// large time ranges are processed and streamed efficiently.
 	// The limit parameter specifies the number of different groups that can
 	// be aggregated over. When absent, the runtime defaults to an appropriate value.
+	//
+	// xxx comment sorted
 	GroupByProc struct {
 		Node
 		Duration Duration     `json:"duration"`
 		Limit    int          `json:"limit,omitempty"`
 		Keys     []Assignment `json:"keys"`
 		Reducers []Reducer    `json:"reducers"`
+		Sorted   int          `json:"sorted"`
 	}
 	// TopProc is similar to proc.SortProc with a few key differences:
 	// - It only sorts in descending order.
--- a/proc/groupby.go
+++ b/proc/groupby.go
@@ -50,7 +50,7 @@ func CompileGroupBy(node *ast.GroupByProc, zctx *resolver.Context) (*GroupByPara
 
 	if node.InputSortDir != 0 && node.Duration.Seconds == 0 && len(astKeys) > 0 {
 		if _, ok := astKeys[0].Expr.(ast.FieldExpr); !ok {
-			return nil, fmt.Errorf("compiling groupby: cannot use -sorted in conjunction with a computed primary key")
+			return nil, fmt.Errorf("compiling groupby: cannot use -sorted in conjunction with a computed primary grouping key")
 		}
 	}
 	if node.Duration.Seconds > 0 {
@@ -159,7 +159,7 @@ type GroupByAggregator struct {
 	reverse      bool
 	logger       *zap.Logger
 	limit        int
-	valueSortFn  expr.ValueSortFn // to compare primary keys for early key output
+	valueSortFn  expr.ValueSortFn // to compare primary group keys for early key output
 	recordSortFn expr.SortFn
 	maxKey       *zng.Value
 	inputSortDir int
@@ -168,7 +168,7 @@ type GroupByAggregator struct {
 type GroupByRow struct {
 	keycols  []zng.Column
 	keyvals  zcode.Bytes
-	primary  *zng.Value // for sorting when input sorted
+	groupval *zng.Value // for sorting when input sorted
 	reducers compile.Row
 }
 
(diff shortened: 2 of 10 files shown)


henridf:
This still needs a companion Brim PR before merging... but should be otherwise ready for review.

henridf:
The companion brim PR is https://github.com/brimsec/brim/pull/864

ast/ast.go
```diff
@@ -263,12 +263,15 @@ type (
 	// large time ranges are processed and streamed efficiently.
 	// The limit parameter specifies the number of different groups that can
 	// be aggregated over. When absent, the runtime defaults to an appropriate value.
+	//
+	// xxx comment sorted
```
nwt, line 267: Fix me!

ast/ast.go
```diff
@@ -263,12 +263,15 @@ type (
...
 	GroupByProc struct {
 		Node
 		Duration Duration     `json:"duration"`
 		Limit    int          `json:"limit,omitempty"`
 		Keys     []Assignment `json:"keys"`
 		Reducers []Reducer    `json:"reducers"`
+		Sorted   int          `json:"sorted"`
```
nwt, line 274:
```suggestion
		Sorted   int          `json:"sorted,omitempty"`
```

proc/groupby.go
```diff
@@ -381,26 +396,29 @@ func (g *GroupByAggregator) Results(eof bool, minTs nano.Ts, maxTs nano.Ts) (zbu
...
 	for _, k := range keys {
-		row := table[k]
-		var zv zcode.Bytes
-		if g.TimeBinDuration > 0 {
-			zv = zcode.AppendPrimitive(zv, zng.EncodeTime(row.ts))
+		row := g.table[k]
+		if !eof {
+			if g.valueSortFn(*row.primary, *g.maxKey) >= 0 {
+				continue
+			}
```
nwt, line 418:
```suggestion
		if !eof && g.valueSortFn(*row.primary, *g.maxKey) >= 0 {
			continue
```

proc/groupby.go
```diff
@@ -316,62 +353,40 @@ func (g *GroupByAggregator) Consume(r *zng.Record) error {
...
-// times; all completed time bins at the time of the invocation are
-// returned. A final call with eof=true should be made to get the
-// final (possibly incomplete) time bin.
-// If this is not a time-binned aggregation, a single call (with
-// eof=true) should be made after all records have been Consumed()'d.
-func (g *GroupByAggregator) Results(eof bool, minTs nano.Ts, maxTs nano.Ts) (zbuf.Batch, error) {
-	var bins []nano.Ts
-	for b := range g.tables {
-		bins = append(bins, b)
+func (g *GroupByAggregator) updateMax(v zng.Value) {
```
nwt, line 368: Nit: updateMaxKey might be a better name.

proc/groupby.go
```diff
@@ -207,26 +237,27 @@ func (g *GroupBy) Pull() (zbuf.Batch, error) {
 		}
 	}
 	batch.Unref()
-	if g.timeBinned {
-		f, err := g.agg.Results(false, g.MinTs, g.MaxTs)
+	if g.agg.inputSortDir != 0 {
+		arr, err := g.agg.Results(false)
```
nwt, line 241: Nit: Call this batch.

ast/ast.go
nwt: Also, Sorted is a pretty vague name, so I think we should call this InputSortDir/input_sort_dir.

proc/groupby.go
```diff
@@ -25,11 +25,11 @@ type GroupByKey struct {
...
-	duration ast.Duration
-	limit    int
-	keys     []GroupByKey
-	reducers []compile.CompiledReducer
-	builder  *ColumnBuilder
+	limit        int
+	keys         []GroupByKey
+	reducers     []compile.CompiledReducer
+	builder      *ColumnBuilder
+	inputSortDir int
```
nwt, line 32: Keep the configuration fields together and (mostly) sorted.
```suggestion
	inputSortDir int
	limit        int
	keys         []GroupByKey
	reducers     []compile.CompiledReducer
	builder      *ColumnBuilder
```

proc/groupby.go
```diff
@@ -47,8 +47,32 @@ const defaultGroupByLimit = 1000000
...
+					&ast.Literal{ast.Node{"Literal"}, "int64", strconv.Itoa(node.Duration.Seconds)},
+				},
+			},
+		}
+		astKeys = append([]ast.Assignment{everyKey}, astKeys...)
+	}
+	if len(astKeys) == 0 {
+		return nil, fmt.Errorf("compiling groupby: must have at least one key (or -every)")
+	}
+	for _, astKey := range astKeys {
```
nwt, line 75: Please don't add 25 lines of unrelated code between the declarations of keys and targetNames and their uses.

ast/ast.go
henridf: 👍

henridf: More importantly, do you think the zql flag `-sorted` should also be renamed?

ast/ast.go
henridf: Fixed

proc/groupby.go
henridf: oops

proc/utils.go
```diff
@@ -197,7 +197,7 @@ func (p *ProcTest) Finish() error {
 	}
 }
 
-func parse(zctx *resolver.Context, src string) (*zbuf.Array, error) {
+func ParseTestZng(zctx *resolver.Context, src string) (*zbuf.Array, error) {
```
nwt, line 200: Nit: This parses tzng, not zng, so ParseTzng might be a better name.

ast/ast.go
nwt: I don't think we should expose this through ZQL.

Instead, I think we should add a `sortKey string` parameter to driver.Compile*, set it appropriately when calling those functions (to `"ts"` in zqd; to `""` in zq for now, ultimately wiring it to a command line flag), and then use it as input to an analysis that will set ast.GroupByProc.InputSortDir to the correct value before calling proc.CompileProc.

That's obviously more work, but we don't need to do a thorough job of it now. We just need enough to catch the Brim histogram AST, which seems pretty straightforward.

ast/ast.go
```diff
@@ -255,20 +255,25 @@ type (
...
+	// duration field is non-zero, then the groups further partioned by time
+	// into bins of the duration. In this case time is the primary key.
+	// The Limit field specifies the number of different groups that can be
+	// aggregated over. When absent, the runtime defaults to an
+	// appropriate value.
+	// The InputSortDir field indicates that input is sorted (with
+	// direction indicated by the sign of the field) in the
+	// primary key. In this case, the proc outputs the reducer
+	// results from each key as they complete so that large inputs
+	// are processed and streamed efficiently.
```
nwt, line 269:
```suggestion
	// of the fields specified in the keys field (where the first key is the
	// grouping key), and applies reducers (if any) to each group.  If the
	// Duration field is non-zero, then the groups are further partitioned by time
	// into bins of the duration. In this case, the grouping key is always ts.
	// The Limit field specifies the number of different groups that can be
	// aggregated over. When absent, the runtime defaults to an
	// appropriate value.
	// The InputSortDir field indicates that input is sorted (with
	// direction indicated by the sign of the field) in the
	// grouping key. In this case, the proc outputs the reducer
	// results from each key as they complete so that large inputs
	// are processed and streamed efficiently.
```

proc/groupby.go
```diff
@@ -46,9 +46,33 @@ func IsErrTooBig(err error) bool {
 const defaultGroupByLimit = 1000000
 
 func CompileGroupBy(node *ast.GroupByProc, zctx *resolver.Context) (*GroupByParams, error) {
+	astKeys := node.Keys
+
+	if node.InputSortDir != 0 && node.Duration.Seconds == 0 && len(astKeys) > 0 {
+		if _, ok := astKeys[0].Expr.(ast.FieldExpr); !ok {
+			return nil, fmt.Errorf("compiling groupby: cannot use -sorted in conjunction with a computed primary key")
```
nwt, line 53:
```suggestion
			return nil, fmt.Errorf("compiling groupby: cannot use -sorted in conjunction with a computed grouping key")
```

proc/groupby.go
```diff
@@ -131,63 +151,73 @@ type GroupByAggregator struct {
...
-	limit           int
+	kctx         *resolver.Context
+	keys         []GroupByKey
+	reducerDefs  []compile.CompiledReducer
+	builder      *ColumnBuilder
+	table        map[string]*GroupByRow
+	reverse      bool
+	logger       *zap.Logger
+	limit        int
+	valueSortFn  expr.ValueSortFn // to compare primary keys for early key output
```
nwt, line 162:
```suggestion
	valueSortFn  expr.ValueSortFn // to compare grouping keys for early key output
```

proc/groupby.go
```diff
@@ -316,62 +353,40 @@ func (g *GroupByAggregator) Consume(r *zng.Record) error {
...
-		newRecs, err := g.recordsForTable(g.tables[b])
-		if err != nil {
-			return nil, err
-		}
-		recs = append(recs, newRecs...)
-		delete(g.tables, b)
+}
+
+// Results returns a batch of aggregation result records. If the input
+// is sorted in the primary key, this can be called multiple times;
```
nwt, line 379:
```suggestion
// is sorted in the grouping key, this can be called multiple times;
```

proc/groupby.go
```diff
@@ -131,63 +151,73 @@ type GroupByAggregator struct {
...
+	recordSortFn expr.SortFn
+	maxKey       *zng.Value
+	inputSortDir int
 }
 
 type GroupByRow struct {
 	keycols  []zng.Column
 	keyvals  zcode.Bytes
-	ts       nano.Ts
+	primary  *zng.Value // for sorting when input sorted
```
nwt, line 171:
```suggestion
	group  *zng.Value // for sorting when input sorted
```
Or groupval or groupVal or something else.

proc/groupby.go
```diff
@@ -46,9 +46,33 @@ func IsErrTooBig(err error) bool {
...
+					&ast.FieldRead{ast.Node{"FieldRead"}, "ts"},
+					&ast.Literal{ast.Node{"Literal"}, "int64", strconv.Itoa(node.Duration.Seconds)},
+				},
+			},
+		}
+		astKeys = append([]ast.Assignment{everyKey}, astKeys...)
+	}
+	if len(astKeys) == 0 {
+		return nil, fmt.Errorf("compiling groupby: must have at least one key (or -every)")
+	}
```
nwt, line 72: I think it would be simpler to do this transformation in zql.peg instead of here. That'd let us drop ast.GroupByProc.Duration, too.

ast/ast.go
henridf: Applied but defined first key as "primary grouping key" rather than "grouping key" because all keys are grouping keys.

proc/groupby.go
henridf: (You mean in parser-support.go and parser-support.js, right?)

I had tried that early on but it wasn't simpler. More importantly, I don't think that parser code should be aware of specific zql functions like "Time.trunc", or doing this kind of transformation.  Ideally it would go in some intermediate compilation pass, but we just have one for the time being... so I put it there.

henridf: (separately, and independently of this PR, Duration should be called Every).

PR #873, opened 2020-06-08
Rename archive.MetaData to Metadata

nwt:
Metadata is one word.


archive/schema.go
```diff
@@ -15,13 +15,13 @@ import (
 
 const metaDataFilename = "zar.json"
```
henridf, line 16: what about metaDataFilename?

nwt: Will fix. Thank you!

PR #877, opened 2020-06-08
Remove zbuf.Batch.Span

### LINKED ISSUE #872: Remove batch.Span()
batch.Span is a holdover from previous work. While all procs (re-)compute & pass the span on, except for groupby's 'reverse' flag that just flips the span's min/max, no proc that consumes the span info & makes decisions based off it.

nwt:
Closes #872.


henridf:
That got de-iced quick! 👍 but this will conflict with #869 so consider letting it go in first.

PR #880, opened 2020-06-09
Rename ReducerProc->ReduceProc

henridf:
None of the other proc names use the "-er" action-noun suffix form.

(Why now? I ran into this when I wanted to move some reducer-stuff into proc/reducer.go.)

### DIFF
--- a/zql/zql.peg
+++ b/zql/zql.peg
@@ -208,7 +208,7 @@ parallelChain
 
 proc
   = simpleProc
-  / reducerProc
+  / reduceProc
   / "(" _? proc:procList _? ")" {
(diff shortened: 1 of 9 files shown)


zql/zql.peg
```diff
@@ -333,7 +333,7 @@ reducerProc
       RETURN(makeGroupByProc(every, limit, keys, reducers))
     }
 
-    RETURN(makeReducerProc(reducers))
+    RETURN(makeReduceProc(reducers))
```
nwt, line 336: Don't forget to rename the rule containing this from reducerProc to reduceProc.

PR #895, opened 2020-06-11
add zapi newsubspace command

alfred-landrum:
Carving this out of the work in https://github.com/brimsec/zq/pull/864 to enable a zar import / zqd archive demo.

### DIFF
--- a/cmd/zapi/cmd/new/new.go
+++ b/cmd/zapi/cmd/new/new.go
@@ -5,6 +5,8 @@ import (
 	"flag"
 	"fmt"
 
+	"github.com/alecthomas/units"
+	"github.com/brimsec/zq/archive"
 	"github.com/brimsec/zq/cmd/zapi/cmd"
 	"github.com/brimsec/zq/zqd/api"
 	"github.com/brimsec/zq/zqd/storage"
@@ -28,6 +30,7 @@ type Command struct {
 	*cmd.Command
 	kind     storage.Kind
 	datapath string
+	thresh   string
 }
 
 func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
@@ -37,6 +40,7 @@ func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
 	}
 	f.Var(&c.kind, "k", "kind of storage for this space")
 	f.StringVar(&c.datapath, "d", "", "specific directory for storage data")
+	f.StringVar(&c.thresh, "log-size-threshold", units.Base2Bytes(archive.DefaultLogSizeThreshold).String(), "target size of chopped files, as '10MB' or '4GiB', etc.")
 	return c, nil
 }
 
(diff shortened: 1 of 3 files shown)


cmd/zapi/cmd/newsubspace/newsubspace.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+	client := c.Client()
+	req := api.SubspacePostRequest{
+		Name: c.name,
+		OpenOptions: storage.ArchiveOpenOptions{
+			LogFilter: args,
+		},
+	}
+	if _, err := client.SubspacePost(c.Context(), api.SpaceID(c.parentID), req); err != nil {
+		return fmt.Errorf("couldn't create subspace %s: %v", c.name, err)
```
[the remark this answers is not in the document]
nwt: (This is largely cosmetic, but I do think it's good to be consistent about using %w for errors.)

cmd/zapi/cmd/new/new.go
```diff
@@ -28,6 +30,7 @@ type Command struct {
 	*cmd.Command
 	kind     storage.Kind
 	datapath string
+	thresh   string
```
nwt, line 33: Maybe make this a bytesValue.
```go
// bytesValue implements flag.Value.
type bytesValue int64

func (b *bytesValue) Set(s string) error {
	bytes, err := units.ParseStrictBytes(s)
	*b = bytesValue(bytes)
	return err
}

func (b bytesValue) String() string {
	return strconv.FormatInt(int64(b), 10)
}
```
You'll have to add it witth flag.(*FlagSet).Var instead of StringVar below, but it'll let you get rid of storageCfg and just say
```go
	req := api.SpacePostRequest{
		Name:     name,
		DataPath: c.datapath,
		Storage: &storage.Config{
			Kind: storage.Kind(c.kind),
			Archive: &storage.ArchiveConfig{
				CreateOptions: &storage.ArchiveCreateOptions{
					LogSizeThreshold: c.thresh.String(),
				},
			},
		},
	}
```

henridf:
fwiw: I hadn't closely tracked all the work around zqd and archive spaces, and caught up with while reviewing this and other PRs today. As a newcomer the term "subspace" initially confused me into thinking it was a subset of a space, perhaps by analogy to subdirectory. It seems like it's more akin to a view in sql lingo, or maybe a virtual space. (If this has already been discussed and I'm late to the party, then disregard...).

cmd/zapi/cmd/new/new.go
```diff
@@ -47,14 +51,28 @@ func (c *Command) Run(args []string) error {
...
+	if c.thresh != "" {
+		if thresh, err := units.ParseStrictBytes(c.thresh); err != nil {
+			return fmt.Errorf("invalid log-size-threshold: %w", err)
+		} else {
+			storageCfg.Archive = &storage.ArchiveConfig{
+				CreateOptions: &storage.ArchiveCreateOptions{
+					LogSizeThreshold: &thresh,
+				},
+			}
+		}
```
nwt, line 67: If you keep this, I think it reads better if you don't wrap the success path inside an else block.
```suggestion
		thresh, err := units.ParseStrictBytes(c.thresh)
		if err != nil {
			return fmt.Errorf("invalid log-size-threshold: %w", err)
		}
		storageCfg.Archive = &storage.ArchiveConfig{
			CreateOptions: &storage.ArchiveCreateOptions{
				LogSizeThreshold: &thresh,
			},
		}
```

PR #897, opened 2020-06-11
ZNG spec cleanup

philrz:
The ZNG spec has held still for a little while, so I've made a pass through it to clean up some formatting, grammar, and make other readability tweaks so I'll feel better about linking to it from other docs.

While I expect most of the changes are not controversial, I've placed comments in spots where I'd appreciate explicit sign-off that my attempts at better wording have not corrupted the intent of the original text.

A high-level summary of things I've touched here:

1. I've added a table-of-contents (TOC) to make it easier for readers to bounce to relevant parts of the doc. Since Markdown doesn't have easy ways to automatically keep a TOC updated based on the headings, this does mean we're now responsible for keeping the TOC in sync with the headers, should we add/change section numbers & titles. However, it's my hope that ZNG is now stable enough that the benefits of having the TOC outweigh the hassles of maintaining it.

2. I noticed that our use of section headings was not wholly consistent (e.g. a number heading three-deep like `x.y.z` did not always have a consistent Markdown header size like `###`). While I've made this _consistent_, at the same time I noticed that if we followed a hard rule of always moving down a header size for every additional level of number depth, our four-deep number headings would look very small like the one below. Therefore I've maintained a pattern that was there before where number depths `x.y` and `x.y.z` both have the same size headers (`###`), but the TOC shows them at different indent depths.

---

##### 3.1.1.4 This is a small header. Too small, I would say!

---

3. I ended up changing the section numbers in order to assign a number to the top-level introduction and make the "Examples" section a sub-section of the TZNG section, since it's framed as being TZNG-specific detail.

### DIFF
--- a/zng/docs/spec.md
+++ b/zng/docs/spec.md
@@ -22,14 +22,14 @@
       - [3.1.1.4 Union Typedef](#3114-union-typedef)
       - [3.1.1.5 Alias Typedef](#3115-alias-typedef)
     - [3.1.2 End-of-Stream Markers](#312-end-of-stream-markers)
-  + [3.2 ZNG Value Messages](#32-zng-value-messages)
+  + [3.2 Value Messages](#32-value-messages)
 * [4. ZNG Text Format (TZNG)](#4-zng-text-format-tzng)
-  + [4.1 ZNG Control Messages](#41-zng-control-messages)
+  + [4.1 Control Messages](#41-control-messages)
     - [4.1.1 Type Binding](#411-type-binding)
     - [4.1.2 Type Alias](#412-type-alias)
     - [4.1.3 Application-Specific Payload](#413-application-specific-payload)
   + [4.2 Type Grammar](#42-type-grammar)
-  + [4.3 ZNG Values](#43-zng-values)
+  + [4.3 Values](#43-values)
     - [4.3.1 Character Escape Rules](#431-character-escape-rules)
     - [4.3.2 Value Syntax](#432-value-syntax)
   + [4.4 Examples](#44-examples)
@@ -332,6 +332,10 @@ existing type ID ``<type-id>``.  ``<type-id>`` is encoded as a `uvarint` and `<n
 is encoded as a `uvarint` representing the length of the name in bytes,
 followed by that many bytes of UTF-8 string.
 
+It is an error to define an alias that has the same name as a primitive type.
+It is also an error to redefine a previously defined alias with a
+type that differs from the original definition.
+
 ### 3.1.2 End-of-Stream Markers
 
 A ZNG stream must be terminated by an end-of-stream marker.
@@ -368,7 +372,7 @@ previously defined type, the appropriate typedef control code must
 be re-emitted
 (and note that the typedef may now be assigned a different ID).
 
-### 3.2 ZNG Value Messages
+### 3.2 Value Messages
 
 Following a header byte with bit 7 zero is a `typed value`
 with a `uvarint7` encoding its length.
@@ -391,6 +395,9 @@ length in bytes of the type ID.
 |uvarint7|type-id|value|
 ------------------------
 ```
+
+It is an error for a value to reference a type ID that has not been previously
+defined by a typedef scoped to the stream in which the value appears.
 
 A typed value with a `value` of length `N` and the type indicated
 is interpreted as follows:
@@ -480,10 +487,10 @@ lexicographically greater than that of the preceding element.
 ## 4. ZNG Text Format (TZNG)
 
 The ZNG text format is a human-readable form that follows directly from the ZNG
-binary format.  A ZNG file/stream is encoded with UTF-8.
+binary format.  A TZNG file/stream is encoded with UTF-8.
 All subsequent references to characters and strings in this section refer to
 the Unicode code points that result when the stream is decoded.
-If a ZNG stream includes data that is not valid UTF-8, the stream is invalid.
+If a TZNG stream includes data that is not valid UTF-8, the stream is invalid.
 
 A stream of control messages and values messages is represented
 as a sequence of lines each terminated by a newline.
@@ -493,18 +500,18 @@ i.e., via `\u{0a}` or `\x0a`.
 A line that begins with `#` is a control message and all other lines
 are values.
 
-### 4.1 ZNG Control Messages
+### 4.1 Control Messages
 
-ZNG control messages have one of four forms defined below.
+TZNG control messages have one of four forms defined below.
 
 Any line beginning with `#` that does not conform with the syntax described here
 is an error.
-When errors are encountered parsing ZNG, an implementation should return a
-corresponding error and allow ZNG parsing to proceed if desired.
+When errors are encountered parsing TZNG, an implementation should return a
+corresponding error and allow TZNG parsing to proceed if desired.
 
 ### 4.1.1 Type Binding
 
-A type binding has the following form:
+A TZNG type binding has the following form:
 ```
 #<type-tag>:<type-string>
 ```
@@ -514,7 +521,7 @@ a binding between the indicated tag and the indicated type.
 
 ### 4.1.2 Type Alias
 
-A type alias has the following form:
+A TZNG type alias has the following form:
 ```
 #<type-name>:<type-string>
 ```
@@ -524,17 +531,14 @@ binding between the indicated tag and the indicated type.
 This form defines an alias mapping the identifier to the indicated type.
 `<type-name>` is an identifier with semantics as defined in [Section 3.1.1.5](#3115-alias-typedef).
 
-It is an error to define an alias that has the same name as a primitive type.
-It is also an error to redefine a previously defined alias with a
-type that differs from the original definition.
 
 ### 4.1.3 Application-Specific Payload
 
-An application-specific payload has the following form:
+A TZNG application-specific payload has the following form:
 ```
-#!<control code>:<payload>
+#!<control-code>:<payload>
 ```
-Here, `<control code>` is a decimal integer in the range 6-127 and `<payload>`
+Here, `<control-code>` is a decimal integer in the range 6-127 and `<payload>`
 is any UTF-8 string with escaped newlines.
 
 ### 4.2 Type Grammar
@@ -564,19 +568,19 @@ XXX - I don't see ctype referenced anywhere else. Shouldn't it be?
 
 <alias-name> := <id>
 
-<id> := <id_start> <id_continue>*
+<id> := <id-start> <id-continue>*
 
-<id_start> := [A-Za-z_$]
+<id-start> := [A-Za-z_$]
 
-<id_continue> := <id_start> | [0-9]
+<id-continue> := <id-start> | [0-9]
 ```
 
 A reference implementation of this type system is embedded in
 [zq/zng](../).
 
-### 4.3 ZNG Values
+### 4.3 Values
 
-A ZNG value is encoded on a line as a typed value, which is encoded as
+A TZNG value is encoded on a line as a typed value, which is encoded as
 an integer type code followed by `:`, which is in turn followed
 by a value encoding.
 
@@ -614,9 +618,6 @@ if the parsed value does not match the indicated type in terms of number and
 sub-structure of value elements present and their interpretation as a valid
 string of the specified type.
 
-It is an error for a value to reference a type ID that has not been previously
-defined by a typedef scoped to the stream in which the value appears.
-
 ### 4.3.1 Character Escape Rules
 
 Any Unicode code point may be represented in a `string` value using
@@ -633,7 +634,7 @@ is not a valid escape sequence.
 The behavior of an implementation that encounters such
 invalid sequences in a `string` type is undefined.
 
-Any character in a `bstring` value may be escaped from the ZNG formatting rules
+Any character in a `bstring` value may be escaped from the TZNG formatting rules
 using the hex escape syntax, i.e., `\xhh` where `h` is a hexadecimal digit.
 This allows binary data that does not conform to a valid UTF-8 character encoding
 to be embedded in the `bstring` data type.
@@ -708,7 +709,7 @@ an earlier-defined type alias:
 #SET:set[string]
 #99:record[v:REC,s:SET,r:REC,s2:SET]
 ```
-This ZNG defines a tag for the primitive string type and defines a record
+This TZNG defines a tag for the primitive string type and defines a record
 and references the types accordingly in three values;
 ```
 #0:string
@@ -726,7 +727,7 @@ string("this is a semicolon: ;")
 Note that the tag integers occupy their own numeric space indepedent of
 any underlying ZNG type IDs.
 
-The semicolon terminator is important.  Consider this ZNG depicting
+The semicolon terminator is important.  Consider this TZNG depicting
 sets of strings:
 ```
 #0:set[string]
--- a/zng/docs/spec.md
+++ b/zng/docs/spec.md
@@ -522,7 +522,7 @@ A TZNG type binding has the following form:
 #<type-tag>:<type-string>
 ```
 Here, `<type-tag>` is a string decimal integer and `<type-string>`
-is a string defining a type according to the [TZNG type grammar`](#42-type-grammar). They create
+is a string defining a record type (`<rtype>`) according to the [TZNG type grammar`](#42-type-grammar). They create
 a binding between the indicated tag and the indicated type.
 
 ### 4.1.2 Type Alias
@@ -556,13 +556,16 @@ grammar describing the textual type encodings is:
          | string | bytes | bstring | enum | ip | port | net | time | duration | null
          | <alias-name>
 
-XXX - I don't see ctype referenced anywhere else. Shouldn't it be?
 <ctype> :=  array [ <stype> ]
-          | record [ <columns> ]
-          | record [ ]
           | union [ <stype-list> ]
           | set [ <stype> ]
+          | <rtype>
 
+<rtype> :=  record [ <columns> ]
+          | record [ ]
+
+
+<type> := <stype> | <ctype>
 
 <stype-list> :=    <stype>
                  | <stype-list> , <stype>
@@ -570,7 +573,7 @@ XXX - I don't see ctype referenced anywhere else. Shouldn't it be?
 <columns> :=      <column>
                 | <columns> , <column>
 
-<column> := <id> : <stype>
+<column> :=  <id> : <type>
 
 <alias-name> := <id>
 
--- a/zng/docs/spec.md
+++ b/zng/docs/spec.md
@@ -522,7 +522,7 @@ A TZNG type binding has the following form:
 #<type-tag>:<type-string>
 ```
 Here, `<type-tag>` is a string decimal integer and `<type-string>`
-is a string defining a type according to the ZNG type syntax. They create
+is a string defining a type according to the [TZNG type grammar`](#42-type-grammar). They create
 a binding between the indicated tag and the indicated type.
 
 ### 4.1.2 Type Alias
@@ -532,7 +532,7 @@ A TZNG type alias has the following form:
 #<type-name>:<type-string>
 ```
 Here, `<type-name>` is an identifier and `<type-string>`
-is a string defining a type according to the ZNG type syntax. They create a
+is a string defining a type according to the [TZNG type grammar](#42-type-grammar). They create a
 binding between the indicated tag and the indicated type.
 This form defines an alias mapping the identifier to the indicated type.
 `<type-name>` is an identifier with semantics as defined in [Section 3.1.1.5](#3115-alias-typedef).
--- a/zng/docs/spec.md
+++ b/zng/docs/spec.md
@@ -510,7 +510,7 @@ A type binding has the following form:
 #<type-tag>:<type-string>
 ```
 Here, `<type-tag>` is a string decimal integer and `<type-string>`
-is a string defining a type according to the ZNG type syntax, which creates
+is a string defining a type according to the ZNG type syntax. They create
 a binding between the indicated tag and the indicated type.
 
 ### 4.1.2 Type Alias
--- a/zng/docs/spec.md
+++ b/zng/docs/spec.md
@@ -80,6 +80,12 @@ As far as we know, the Zeek log format pioneered the concept of
 embedding the schema of the log lines within the log file itself using
 meta-records, and ZNG merely modernizes this original approach.
 
+The [`zq`](https://github.com/brimsec/zq) command-line tool provides a
+reference implementation of ZNG as it's described here, including the type
+system, error handling, etc., barring the exceptions
+described in the [alpha notice](http://0.0.0.0:6419/#zng-specification)
+at the top of this specification.
+
 ## 2. The ZNG data model
 
 ZNG encodes a sequence of one or more typed data values to comprise a stream.
@@ -574,9 +580,6 @@ XXX - I don't see ctype referenced anywhere else. Shouldn't it be?
 
 <id-continue> := <id-start> | [0-9]
 ```
-
-A reference implementation of this type system is embedded in
-[zq/zng](../).
 
 ### 4.3 Values
 
--- a/zng/docs/spec.md
+++ b/zng/docs/spec.md
@@ -12,6 +12,32 @@
 > Also, we are contemplating reducing the number of primitive types, e.g.,
 > the number of variations in integer types.
 
+* [Summary](#summary)
+* [1. The ZNG data model](#1-the-zng-data-model)
+* [2. ZNG Binary Format (ZNG)](#2-zng-binary-format--zng-)
+  + [2.1 Control Messages](#21-control-messages)
+  + [2.1.1 Typedefs](#211-typedefs)
+    - [2.1.1.1 Record Typedef](#2111-record-typedef)
+    - [2.1.1.2 Array Typedef](#2112-array-typedef)
+    - [2.1.1.3 Set Typedef](#2113-set-typedef)
+    - [2.1.1.4 Union Typedef](#2114-union-typedef)
+    - [2.1.1.5 Alias Typedef](#2115-alias-typedef)
+  + [2.1.2 End-of-stream Markers](#212-end-of-stream-markers)
+  + [2.2 ZNG Value Messages](#22-zng-value-messages)
+* [3. ZNG Text Format (TZNG)](#3-zng-text-format--tzng-)
+  + [3.1 ZNG Control Messages](#31-zng-control-messages)
+  + [3.1.1 Type Binding](#311-type-binding)
+  + [3.1.2 Type Alias](#312-type-alias)
+  + [3.1.3 Application-specific Payload](#313-application-specific-payload)
+  + [Type Grammar](#type-grammar)
+  + [3.2 ZNG Values](#32-zng-values)
+    - [3.2.1 Character Escape Rules](#321-character-escape-rules)
+    - [3.2.2 Value Syntax](#322-value-syntax)
+* [4. Examples](#4-examples)
+* [5. Related Links](#5-related-links)
+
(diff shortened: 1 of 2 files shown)


zng/docs/spec.md
```diff
@@ -386,6 +412,7 @@ is interpreted as follows:
 | `bstring`  | variable |  UTF-8 byte sequence with `\x` escapes           |
 | `enum `    | variable |  UTF-8 bytes of enum string                      |
 | `ip`       | 4 or 16  |  4 or 16 bytes of IP address                     |
+| `port`     | XXX      |  XXX - to be filled in during PR review          |
```
philrz, line 415: I noticed there's an entry for the `port` type in the section 3.1.1 Typedefs table, and it seems to be in the code, but it was't mentioned in this table here. I'm hoping someone who knows the code could plug in the correct suggestion. @alfred-landrum sounded like he's game to dig this up.

zng/docs/spec.md
```diff
@@ -511,15 +538,16 @@ An application-specific payload has the following form:
...
 
-### Type Grammar
+### 4.2 Type Grammar
 
 Given the above textual definitions and the underlying ZNG specification, a
 grammar describing the textual type encodings is:
 ```
 <stype> := bool | byte | int16 | uint16 | int32 | uint32 | int64 | uint64 | float64
-         | string | bytes | bstring | enum | ip | net | time | duration | null
+         | string | bytes | bstring | enum | ip | port | net | time | duration | null
```
philrz, line 547: I noticed there's an entry for the `port` type in the section 3.1.1 Typedefs table, and it seems to be in the code, but it was't mentioned in this table here. I'm hoping someone who knows the code could plug in the correct suggestion. @alfred-landrum sounded like he's game to dig this up.

zng/docs/spec.md
```diff
@@ -511,15 +538,16 @@ An application-specific payload has the following form:
 Here, `<control code>` is a decimal integer in the range 6-127 and `<payload>`
 is any UTF-8 string with escaped newlines.
 
-### Type Grammar
+### 4.2 Type Grammar
```
philrz, line 541: I don't know if it was intentional or an oversight that this section didn't have a number before. However, once there's a TOC, it's usually populated with the headings that make up the sections, and once we start numbering sections it feels odds to have non-numbered, heading-labeled sections, so it seemed appropriate to add one here.

zng/docs/spec.md
```diff
@@ -511,15 +538,16 @@ An application-specific payload has the following form:
...
 
 Given the above textual definitions and the underlying ZNG specification, a
 grammar describing the textual type encodings is:
 ```
 <stype> := bool | byte | int16 | uint16 | int32 | uint32 | int64 | uint64 | float64
-         | string | bytes | bstring | enum | ip | net | time | duration | null
+         | string | bytes | bstring | enum | ip | port | net | time | duration | null
          | <alias-name>
 
+XXX - I don't see ctype referenced anywhere else. Shouldn't it be?
```
philrz, line 550: I'm not a pro at these pseudo-grammars, but just following the pattern of what I see in the ones in this spec, it seems like we only define something in `<brackets>` if it's actually referenced elsewhere, and this one isn't referenced elsewhere, so I want to check if something needs to be corrected here.

zng/docs/spec.md
```diff
@@ -652,12 +680,13 @@ Type | Format
 `bstring` | a UTF-8 string with `\x` escapes of non-UTF binary data
 `enum` | a string representing an enumeration value defined outside the scope of ZNG
 `ip` | a string representing an IP address in [IPv4 or IPv6 format](https://tools.ietf.org/html/draft-main-ipaddr-text-rep-02#section-3)
+`port` | XXX - to be filled in during PR review          |
```
philrz, line 683: I noticed there's an entry for the `port` type in the section 3.1.1 Typedefs table, and it seems to be in the code, but it wasn't mentioned in this table here. I'm hoping someone who knows the code could plug in the correct suggestion. @alfred-landrum sounded like he's game to dig this up.

zng/docs/spec.md
```diff
@@ -476,33 +503,33 @@ is an error.
...
+### 4.1.1 Type Binding
 
 A type binding has the following form:
 ```
 #<type-tag>:<type-string>
 ```
 Here, `<type-tag>` is a string decimal integer and `<type-string>`
-is a string defining a type according to the ZNG type syntax creates a binding
-between the indicated tag and the indicated type.
+is a string defining a type according to the ZNG type syntax, which creates
```
philrz, line 513: The original wording was grammatically odd. After my adjustment, is this text still clearly communicating the original intent?

zng/docs/spec.md
```diff
@@ -476,33 +503,33 @@ is an error.
...
+### 4.1.2 Type Alias
 
 A type alias has the following form:
 ```
 #<type-name>:<type-string>
 ```
 Here, `<type-name>` is an identifier and `<type-string>`
-is a string defining a type according to the ZNG type syntax creates a binding
-between the indicated tag and the indicated type.
+is a string defining a type according to the ZNG type syntax, which creates a
```
philrz, line 523: Same as prior comment: Does my `, which` negatively impact what was being communicated here?

zng/docs/spec.md
alfred-landrum, line 415:
```suggestion
| `port`     | 2      |  2 byte unsigned int          |
```

zng/docs/spec.md
alfred-landrum, line 683:
```suggestion
`port` | a string representing an 16-bit network port |
```

zng/docs/spec.md
```diff
@@ -230,9 +256,9 @@ The field names in a record must be unique.
 
 The `<nfields>` is encoded as a `uvarint`.
 
-The field name is encoded as a UTF-8 string defining a "ZNG identifier"
+The field name is encoded as a UTF-8 string defining a "ZNG identifier".
 The UTF-8 string
-is further encoded as a "counted string", which is `uvarint` encoding
+is further encoded as a "counted string", which is a `uvarint` encoding
```
nwt, line 261:
```suggestion
is further encoded as a "counted string", which is the `uvarint` encoding
```

zng/docs/spec.md
nwt: Nit: `s/2 byte unsigned int/unsigned int of length 2/` for consistency with other rows.

zng/docs/spec.md
```diff
@@ -12,6 +12,32 @@
...
+    - [4.1.2 Type Alias](#412-type-alias)
+    - [4.1.3 Application-specific Payload](#413-application-specific-payload)
+  + [4.2 Type Grammar](#42-type-grammar)
+  + [4.3 ZNG Values](#43-zng-values)
+    - [4.3.1 Character Escape Rules](#431-character-escape-rules)
+    - [4.3.2 Value Syntax](#432-value-syntax)
+  + [4.4 Examples](#44-examples)
+* [5. Related Links](#5-related-links)
+
+## 1. Summary
```
nwt, line 39: s/Summary/Introduction/

zng/docs/spec.md
nwt: Yes, but it still reads oddly to me. How about this?
```suggestion
is a string defining a type according to the ZNG type syntax. They create
```

zng/docs/spec.md
nwt: No, but I think it reads oddly, so how about this?
```suggestion
is a string defining a type according to the ZNG type syntax. They create a
```

zng/docs/spec.md
```diff
@@ -386,6 +412,7 @@ is interpreted as follows:
 | `bstring`  | variable |  UTF-8 byte sequence with `\x` escapes           |
 | `enum `    | variable |  UTF-8 bytes of enum string                      |
 | `ip`       | 4 or 16  |  4 or 16 bytes of IP address                     |
+| `port`     | 2      |  2 byte unsigned int          |
```
nwt, line 415: Nits:
1. `s/2 byte unsigned int/unsigned int of length 2/` for consistency with the other integer types.
2. Add spaces to keep the table pretty.

zng/docs/spec.md
```diff
@@ -652,12 +680,13 @@ Type | Format
 `bstring` | a UTF-8 string with `\x` escapes of non-UTF binary data
 `enum` | a string representing an enumeration value defined outside the scope of ZNG
 `ip` | a string representing an IP address in [IPv4 or IPv6 format](https://tools.ietf.org/html/draft-main-ipaddr-text-rep-02#section-3)
+`port` | a string representing an 16-bit network port |
```
nwt, line 683: Nit: `s/decimal string representation of an unsigned, 16-bit integer/` for consistency with other integer types.

zng/docs/spec.md
philrz: 👍

zng/docs/spec.md
philrz: 👍

zng/docs/spec.md
```diff
@@ -75,7 +101,7 @@ looks like this:
 ```
 Here, the type ID is the integer "9" representing the string type
 (defined in [Typedefs](#typedefs)) and the data value "hello, world"
-is an instance of string.
+is an instance of a string.
```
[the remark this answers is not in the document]
philrz: 👍

zng/docs/spec.md
philrz: 👍

zng/docs/spec.md
philrz: 🤷‍♂️  Works for me. I actually did waffle between "Summary" and "Introduction" and didn't feel strongly either way, so I flipped the coin and went with "Summary". Since I have at least one tiebreaker vote now, I've flipped it to "Introduction" per your suggestion.

zng/docs/spec.md
philrz: 1. Strictly speaking, the `...of length...` notation seems to be used for the ones that list `variable` in the `N` column, whereas most of the ones that have a static number in the `N` column lead off with a reference to that static number like @alfred-landrum's suggestion was doing, albeit with a slightly different wording, i.e. if I was trying to follow that pattern I might have landed on `2 bytes of unsigned int`. But that reads a little odd to me, so I've just gone with your suggestion.

2. You're certainly right about the spaces, so I've fixed that to.

zng/docs/spec.md
philrz: Fine point. While I was thinking about consistency, I noticed another entry in the same table led with `a decimal` so I shortened that one to just `decimal` to be like the others. Then I noticed the whole table was lacking in spaces to make it pretty at the Markdown level, so I tidied that up a bit while I was at it.

zng/docs/spec.md
```diff
@@ -12,6 +12,32 @@
 > Also, we are contemplating reducing the number of primitive types, e.g.,
 > the number of variations in integer types.
 
+* [1. Summary](#1-summary)
```
henridf: 👍

henridf: (for the addition of the ToC)

zng/docs/spec.md
```diff
@@ -467,7 +494,7 @@ i.e., via `\u{0a}` or `\x0a`.
 A line that begins with `#` is a control message and all other lines
 are values.
 
-### 3.1 ZNG Control Messages
+### 4.1 ZNG Control Messages
```
henridf, line 497:
```suggestion
### 4.1 TZNG Control Messages
```

zng/docs/spec.md
```diff
@@ -511,15 +538,16 @@ An application-specific payload has the following form:
 Here, `<control code>` is a decimal integer in the range 6-127 and `<payload>`
```
henridf, line 538: Suggestion: Replace "Here, `<control code>`" with "`Where, `<control code>`"

zng/docs/spec.md
```diff
@@ -547,9 +575,9 @@ grammar describing the textual type encodings is:
 A reference implementation of this type system is embedded in
```
henridf, line 575: This comment seems a little out of place just after the TZNG pseudo-grammar. Maybe it should go at the end of section 3?

zng/docs/spec.md
```diff
@@ -12,6 +12,32 @@
...
+      - [3.1.1.3 Set Typedef](#3113-set-typedef)
+      - [3.1.1.4 Union Typedef](#3114-union-typedef)
+      - [3.1.1.5 Alias Typedef](#3115-alias-typedef)
+    - [3.1.2 End-of-stream Markers](#312-end-of-stream-markers)
+  + [3.2 ZNG Value Messages](#32-zng-value-messages)
+* [4. ZNG Text Format (TZNG)](#4-zng-text-format-tzng)
+  + [4.1 ZNG Control Messages](#41-zng-control-messages)
+    - [4.1.1 Type Binding](#411-type-binding)
+    - [4.1.2 Type Alias](#412-type-alias)
+    - [4.1.3 Application-specific Payload](#413-application-specific-payload)
```
nwt, line 31:
```suggestion
    - [4.1.3 Application-Specific Payload](#413-application-specific-payload)
```

zng/docs/spec.md
```diff
@@ -12,6 +12,32 @@
...
+* [2. The ZNG data model](#2-the-zng-data-model)
+* [3. ZNG Binary Format (ZNG)](#3-zng-binary-format-zng)
+  + [3.1 Control Messages](#31-control-messages)
+    - [3.1.1 Typedefs](#311-typedefs)
+      - [3.1.1.1 Record Typedef](#3111-record-typedef)
+      - [3.1.1.2 Array Typedef](#3112-array-typedef)
+      - [3.1.1.3 Set Typedef](#3113-set-typedef)
+      - [3.1.1.4 Union Typedef](#3114-union-typedef)
+      - [3.1.1.5 Alias Typedef](#3115-alias-typedef)
+    - [3.1.2 End-of-stream Markers](#312-end-of-stream-markers)
```
nwt, line 25:
```suggestion
    - [3.1.2 End-Of-Stream Markers](#312-end-of-stream-markers)
```

zng/docs/spec.md
```diff
@@ -307,7 +333,7 @@ existing type ID ``<type-id>``.  ``<type-id>`` is encoded as a `uvarint` and `<n
 is encoded as a `uvarint` representing the length of the name in bytes,
 followed by that many bytes of UTF-8 string.
 
-### 2.1.2 End-of-stream Markers
+### 3.1.2 End-of-stream Markers
```
nwt, line 336:
```suggestion
### 3.1.2 End-Of-Stream Markers
```

zng/docs/spec.md
```diff
@@ -476,33 +503,33 @@ is an error.
...
 This form defines an alias mapping the identifier to the indicated type.
-`<type-name>` is an identifier with semantics as defined in Section 2.1.1.4.
+`<type-name>` is an identifier with semantics as defined in [Section 3.1.1.5](#3115-alias-typedef).
 
 It is an error to define an alias that has the same name as a primitive type.
 It is also an error to redefine a previously defined alias with a
 type that differs from the original definition.
 
-### 3.1.3 Application-specific Payload
+### 4.1.3 Application-specific Payload
```
nwt, line 532:
```suggestion
### 4.1.3 Application-Specific Payload
```

zng/docs/spec.md
philrz: Actually, if I made that change, I'd think we should drop the comma as well. @nwt is usually good for a tiebreaker in nit situations. Any opinion @nwt?

nwt: "Where A is B and C is D." doesn't make sense to me.

henridf: Oops I meant without comma.

philrz: @nwt: Is that your way of saying you prefer the original wording?

nwt: @philrz: It is, but only because the suggested change appears to turn the sentence into a sentence fragment.

zng/docs/spec.md
philrz: @nwt: According to https://www.grammarcheck.net/capitalization-in-titles-101/:

> Never capitalize prepositions and conjunctions of four or fewer letters

So I've committed a change to capitalize "Stream" but not "of".

nwt: @philrz: Good catch!

zng/docs/spec.md
philrz: @nwt: Agreed. Based on that I think I'll decline @henridf's suggestion here. Thanks to both for weighing in. 😄

zng/docs/spec.md
philrz: @henridf: Good catch. I agree it seems out of place. I'm not sure if it belongs at the end of section 3 either. Here's some thinking out loud...

I looked back at the history of this. As I suspected, the line has been there since [the first version of the ZSON spec](https://github.com/brimsec/zq/blob/5fc47f71c7f9b130c3bcd4a2d377735349fd39a9/pkg/zson/docs/spec.md) when there was no binary format yet and it was trying to be backward compatible with Zeek, and hence was pointing at what was then `zq/pkg/zeek` as the reference implementation, since at the time we were implementing the Zeek data types in their entirety. It seems as if the line has lived on and morphed we moved packages around and cut the ties on 100% backward compatibility with Zeek.

Now, even without me fully understanding the code, no doubt that TZNG psuedo-grammar is implemented _somewhere_ in zq. What I can't say is precisely where, or if it's spread out, if the line is helpful as currently written, as it seems to hand-wave at a point in the code tree that implements lots of other things too.

Up-leveling a bit, wouldn't it be fair to say that zq acts as a reference implementation for _everything_ in the ZNG spec? That is, not just the implementation of the type systems in both TZNG and binary ZNG, but also the reader behaviors, error handling, etc. The opening disclaimer at the top of the spec effectively already says that zq is this reference implementation. It seems one thing we could do is remove the line from this spot below the TZNG pseudo-grammar, make it more generic, and move it to the bottom of the Introduction. After all, it seems a fitting way to close that paragraph would be to remind the reader that the spec is backed up by running code.

... or... if we think it's helpful to the reader to give _precise_ pointers to places within the zq source code to where certain parts of the spec are implemented, we could do that, starting by perhaps refining this one to point specifically to where the TZNG grammar is implemented, then for consistency, adding the same at other places in the spec. This feels like a lot of work that I'd need to lean on you Dev guys for, though. Considering it remains to be seen how many people are scrutinizing the spec at that level of detail, I'm _not_ personally advocating that we make that investment. If people start coming to us with such questions on Slack we could answer them a la carte until it becomes such a nuisance that we add this to the spec.

I'm open to other suggestions as well.

zng/docs/spec.md
```diff
@@ -6,12 +6,37 @@
 > zq attempts to implement everything herein excepting:
 >
 > * the bytes type is not yet implemented,
-> * ordering hints are not generated or taken advantage of,
```
philrz, line 9: I've realized late in the PR review process that this was a dangling reference, as mention of ordering hints was removed from the spec in April, 2020 as part of #508. There's plenty of chatter on Slack around April 3, 2020 among the Dev team all agreeing that it was appropriate to remove ordering hints from the spec since it had never been implemented and there were apparently no plans to implement it, so it seems the dangling reference should go as well.

zng/docs/spec.md
philrz: @henridf (and cc: @nwt since he gave a 👍  to this comment): Good catch. Upon closer reading though, it seems like there may be more to do here than one surgical change. Throughout the TZNG section, there's places where the spec could probably stand to be explicit in clarifying where we're talking specifically about TZNG vs. general characteristics of ZNG that apply to both types of presentation. This comes up early in the intro text of Section 4 on TZNG: _"A ZNG file/stream is encoded with UTF-8"_ should probably say _"TZNG"_, for instance. If you guys agree, I'll take a shot at making the correct changes throughout that section and ask for a re-review on it.

zng/docs/spec.md
henridf: I agree with making the sentence more generic and moving it to the intro. And also that doing detailed references to specific parts of the implementation isn't worthwhile for the time being.

siskojr:
The TOC helps navigate faster to specific areas after spec is read for the first time.
The examples in section 4.4 help a lot to understand the outputs. The more the better. 

I like the size of the numbering-section titles. I don't see a change versus current version of spec.

Thanks

zng/docs/spec.md
```diff
@@ -343,7 +372,7 @@ previously defined type, the appropriate typedef control code must
 be re-emitted
 (and note that the typedef may now be assigned a different ID).
 
-### 2.2 ZNG Value Messages
+### 3.2 Value Messages
```
philrz, line 375: Maybe I went a little crazy with the scissors here, but while I was cleaning up the TZNG section, I realized we were being inconsistent with the headers, i.e. redundant references to "ZNG" in headers of sub-sections that had already been described in a higher-level header as being about ZNG. Rather than have lots of redundant ZNG refs in the ZNG headers & redundant TZNG refs in the TZNG section, I opted for consistent brevity.

zng/docs/spec.md
```diff
@@ -367,7 +396,10 @@ length in bytes of the type ID.
 ------------------------
 ```
 
-A typed value with a `value` of length N and the type indicated
+It is an error for a value to reference a type ID that has not been previously
+defined by a typedef scoped to the stream in which the value appears.
```
philrz, line 400: I moved this up from the TZNG section, as it reads as if it's about ZNG in general.

zng/docs/spec.md
```diff
@@ -467,59 +500,57 @@ i.e., via `\u{0a}` or `\x0a`.
...
 
-A type binding has the following form:
+A TZNG type binding has the following form:
 ```
 #<type-tag>:<type-string>
 ```
 Here, `<type-tag>` is a string decimal integer and `<type-string>`
-is a string defining a type according to the ZNG type syntax creates a binding
-between the indicated tag and the indicated type.
+is a string defining a type according to the ZNG type syntax. They create
```
philrz, line 519: Can someone help point to the right place in the spec for the "ZNG type syntax" it's referencing here? I should surely put a hyperlink here.

zng/docs/spec.md
```diff
@@ -467,59 +500,57 @@ i.e., via `\u{0a}` or `\x0a`.
...
 
-A type alias has the following form:
+A TZNG type alias has the following form:
 ```
 #<type-name>:<type-string>
 ```
 Here, `<type-name>` is an identifier and `<type-string>`
-is a string defining a type according to the ZNG type syntax creates a binding
-between the indicated tag and the indicated type.
+is a string defining a type according to the ZNG type syntax. They create a
```
philrz, line 529: Same as previous comment: Can someone help point to the right place in the spec for the "ZNG type syntax" it's referencing here? I should surely put a hyperlink here.

zng/docs/spec.md
```diff
@@ -307,7 +332,11 @@ existing type ID ``<type-id>``.  ``<type-id>`` is encoded as a `uvarint` and `<n
 is encoded as a `uvarint` representing the length of the name in bytes,
 followed by that many bytes of UTF-8 string.
 
-### 2.1.2 End-of-stream Markers
+It is an error to define an alias that has the same name as a primitive type.
+It is also an error to redefine a previously defined alias with a
+type that differs from the original definition.
```
philrz, line 337: I moved this up from the TZNG section, as it reads as if it's about ZNG in general.

zng/docs/spec.md
```diff
@@ -467,59 +500,57 @@ i.e., via `\u{0a}` or `\x0a`.
...
+### 4.1.3 Application-Specific Payload
 
-An application-specific payload has the following form:
+A TZNG application-specific payload has the following form:
 ```
-#!<control code>:<payload>
+#!<control-code>:<payload>
 ```
-Here, `<control code>` is a decimal integer in the range 6-127 and `<payload>`
+Here, `<control-code>` is a decimal integer in the range 6-127 and `<payload>`
```
philrz, line 541: I noticed that the majority of our other `<things-in-angle-brackets>` use hyphens in place of spaces, so I've made that consistent here.

zng/docs/spec.md
```diff
@@ -537,19 +568,19 @@ grammar describing the textual type encodings is:
...
 <alias-name> := <id>
 
-<id> := <id_start> <id_continue>*
+<id> := <id-start> <id-continue>*
 
-<id_start> := [A-Za-z_$]
+<id-start> := [A-Za-z_$]
 
-<id_continue> := <id_start> | [0-9]
+<id-continue> := <id-start> | [0-9]
```
philrz, line 575: I noticed that the majority of our other `<things-in-angle-brackets>` use hyphens instead of underscores, so I've made that consistent here.

zng/docs/spec.md
philrz: @henridf: I've made the changes to clean up the TZNG section, so PTAL. Not only did I mention TZNG explicitly now in several places, it also led me to make some of the headers more consistent and notice a couple bits that probably belonged in the ZNG section, so I moved those up. I've added new PR comments to call attention to where I did that. I also noticed a couple references to "ZNG type syntax" that I realized should be a link to somewhere, but it's not obvious where, so I'm hoping someone can give me a tip there. That someone could be you. 😄

zng/docs/spec.md
philrz: @henridf: I've made it generic and moved it to the bottom of the intro. PTAL and let me know what you think. Thanks!

henridf: 👍

henridf: 👍

zng/docs/spec.md
henridf: This is referring to section 4.2 below.

zng/docs/spec.md
henridf: (https://github.com/brimsec/zq/pull/897/files/6d6554fc93539db4e8ee8b4ad87cfc144b465c16#r439974371)

zng/docs/spec.md
henridf: Good catch. We could just add:

```
<type-string> := <ctype>
```

to say that a type definition can be any container type. The caveat is that this isn't currently matched by the tzng implementation, which requires that a type definition be a record type. 

This could be ok based on a similar precedent of the zng spec being broader than the current implementation. If however we want the grammar to match the implementation, then we could have `<type-string> := <rtype>` and 

```
<rtype> :=  record [ <columns> ]
               | record [ ]

<ctype> :=  array [ <stype> ]
          | union [ <stype-list> ]
          | set [ <stype> ]
          | <rtype>
```

zng/docs/spec.md
philrz: Thanks @henridf. Based on that, I've not only added a hyperlink but set the link text to "TZNG type grammar" so that way a reader familiar with a spec will know what it's referencing without even having to click on the link.

zng/docs/spec.md
henridf: I pushed a commit to do the above, and also updates `<column>` to consist of both simple and container types. (having realized as I did it that my comment above was incomplete)

zng/docs/spec.md
```diff
@@ -522,7 +522,7 @@ A TZNG type binding has the following form:
 #<type-tag>:<type-string>
 ```
 Here, `<type-tag>` is a string decimal integer and `<type-string>`
-is a string defining a type according to the [TZNG type grammar`](#42-type-grammar). They create
```
henridf: Looks like there's a typo here with a stray backtick after "TZNG type grammar"

zng/docs/spec.md
philrz: Cool, thanks @henridf! While looking at your commit, I noticed for the first time some excess space characters that were in there from before, so I've pushed yet another commit to improve that.

zng/docs/spec.md
philrz: Eek. Thanks for noticing. Fixed.

PR #898, opened 2020-06-12
zq: support directory output to s3

mattnibs:
Make emitter.Dir support s3 paths.

Actually a lot simpler than originally thought. In emitter.Dir if
dir path is a valid s3 path:
1. Do not attempt os.MkdirAll to create the directory.
2. Use path.Join instead of filepath.Join to create leaf paths.

### DIFF
--- a/pkg/iosource/file.go
+++ b/pkg/iosource/file.go
@@ -22,8 +22,7 @@ func (f *FileSource) NewWriter(path string) (io.WriteCloser, error) {
 }
 
 func filePath(path string) string {
-	u, _ := url.Parse(path)
-	if u != nil {
+	if u, _ := url.Parse(path); u != nil {
 		return u.Path
 	}
 	return path
--- a/pkg/iosource/file.go
+++ b/pkg/iosource/file.go
@@ -4,6 +4,8 @@ import (
 	"io"
 	"net/url"
 	"os"
+
+	"github.com/brimsec/zq/pkg/fs"
 )
 
 var DefaultFileSource = &FileSource{Perm: 0666}
@@ -13,12 +15,12 @@ type FileSource struct {
 }
 
 func (f *FileSource) NewReader(path string) (io.ReadCloser, error) {
-	return os.Open(filePath(path))
+	return fs.Open(filePath(path))
 }
 
 func (f *FileSource) NewWriter(path string) (io.WriteCloser, error) {
 	path = filePath(path)
-	return os.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_TRUNC, f.Perm)
+	return fs.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_TRUNC, f.Perm)
 }
 
 func filePath(path string) string {
--- a/pkg/iosource/file.go
+++ b/pkg/iosource/file.go
@@ -1,0 +1,30 @@
+package iosource
+
+import (
+	"io"
+	"net/url"
+	"os"
+)
+
+var DefaultFileSource = &FileSource{Perm: 0666}
+
+type FileSource struct {
+	Perm os.FileMode
+}
+
+func (f *FileSource) NewReader(path string) (io.ReadCloser, error) {
+	return os.Open(filePath(path))
+}
+
+func (f *FileSource) NewWriter(path string) (io.WriteCloser, error) {
+	path = filePath(path)
+	return os.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_TRUNC, f.Perm)
+}
+
+func filePath(path string) string {
+	u, _ := url.Parse(path)
+	if u != nil {
+		return u.Path
+	}
+	return path
+}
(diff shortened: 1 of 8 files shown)


mattnibs:
> The changes look ok, how about a test to verify?

Good idea, will do.

mattnibs:
@alfred-landrum testing the emitter.Dir was a bit tricky. I ended up implemented a package `iosource` for loading reader/writers for paths based on the schema found in path. I based this off of what zap does for [mapping paths to sink factories](https://godoc.org/go.uber.org/zap#RegisterSink)

I had been thinking of doing something like this anyways- the need to test emitter.Dir provided the impetus for this. PTAL!

pkg/iosource/file.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+func (f *FileSource) NewWriter(path string) (io.WriteCloser, error) {
+	path = filePath(path)
+	return os.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_TRUNC, f.Perm)
+}
+
+func filePath(path string) string {
+	u, _ := url.Parse(path)
+	if u != nil {
```
nwt, line 26:
```suggestion
	if u, _ := url.Parse(path); u != nil {
```

pkg/iosource/file.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+var DefaultFileSource = &FileSource{Perm: 0666}
+
+type FileSource struct {
+	Perm os.FileMode
+}
+
+func (f *FileSource) NewReader(path string) (io.ReadCloser, error) {
+	return os.Open(filePath(path))
```
nwt, line 16: I think you'll need to use pkg/fs.Open here.

pkg/iosource/file.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+func (f *FileSource) NewWriter(path string) (io.WriteCloser, error) {
+	path = filePath(path)
+	return os.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_TRUNC, f.Perm)
```
nwt, line 21: I think you need to use pkg/fs.OpenFile here.

pkg/iosource/iosource.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func getScheme(path string) string {
+	u, _ := url.Parse(path)
+	if u == nil {
+		return FileScheme
+	}
+	if u.Scheme == "" {
+		return FileScheme
+	}
```
nwt, line 98:
```suggestion
	if u == nil || u.Scheme == "" {
		return FileScheme
	}
```

pkg/iosource/iosource.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+func init() {
+	resetDefaultRegistry()
+}
+
+func resetDefaultRegistry() {
+	DefaultRegistry = &Registry{}
+	DefaultRegistry.Add(FileScheme, DefaultFileSource)
+}
```
nwt, line 23:
```suggestion
var DefaultRegistry = &Registry{
	schemes: map[string]Source{"file": DefaultFileSource},
}
```
And I think this file will read better if you put this right before the functions that use it (namely, immediately before Register).

pkg/iosource/iosource.go
```diff
@@ -0,0 +1,100 @@
...
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+	_, ok := r.schemes[scheme]
+	return scheme, ok
+}
+
+type Source interface {
+	NewReader(path string) (io.ReadCloser, error)
+	NewWriter(path string) (io.WriteCloser, error)
+}
```
nwt, line 77: Nit: I think this file will read better if you put this adjacent to the declaration of Registry.

pkg/s3io/iosource.go
```diff
@@ -0,0 +1,24 @@
...
+
+type Source struct {
+	Config *aws.Config
+}
+
+func (l *Source) NewWriter(path string) (io.WriteCloser, error) {
+	return NewWriter(path, l.Config)
+}
+
+func (l *Source) NewReader(path string) (io.ReadCloser, error) {
```
nwt, line 22: Nit: `l` is an odd name for the receiver.
```suggestion
func (s *Source) NewWriter(path string) (io.WriteCloser, error) {
	return NewWriter(path, s.Config)
}

func (s *Source) NewReader(path string) (io.ReadCloser, error) {
```

pkg/iosource/file.go
mattnibs: Yep, good call.

pkg/iosource/iosource.go
mattnibs: word.

pkg/iosource/iosource.go
```diff
@@ -0,0 +1,88 @@
...
+	return loader, nil
+}
+
+func (r *Registry) GetScheme(path string) (string, bool) {
+	scheme := getScheme(path)
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+	_, ok := r.schemes[scheme]
+	return scheme, ok
+}
```
nwt, line 68: I believe `(&Registry{}).Source("")` and `(&Registry{}).GetScheme("")` will both panic because `r.schemes` will be nil for an empty registry. Not a deal breaker, but perhaps worth fixing.

mattnibs: good point

PR #899, opened 2020-06-12
Change ZQL cut to emit any matching fields

### LINKED ISSUE #852: Return as many named fields as possible from records that may not contain some of them
New processor functionality is needed to allow the Brim "Export to ZNG" option to work as the user would expect when the Columns chooser has been invoked. The need for this was first identified when verifying [brim/814](https://github.com/brimsec/brim/issues/814#issuecomment-636130351).

Consider the following query result in Brim `v0.10.0` after having imported [this 2018 wrccdc pcap](https://archive.wrccdc.org/pcaps/2018/wrccdc.2018-03-23.010014000000000.pcap.gz).

Starting from this diverse mix of `http` and `dns` events, let's say I click the Columns chooser and check off a mix of fields that appear in both event types (`ts`, `uid`) and fields that appear in only one event type (`uri` from the `http` events, and `query` from the `dns` events).

If I click Export, the `results.zng` that's produced contains _zero_ events. This is a side-effect of how the app currently implements the Export function: The `cut` processor is used to ask `zqd` to trim the query response to contain only the named columns. However, the way `cut` is currently implemented, a record is only returned if _all_ the named fields are present. This can be seen by dropping to the command line and executing equivalent `zq` by hand on the `all.zng` that was generated by the import.

```
$ zq -t "cut ts,uid,uri,query" all.zng 
Cut fields ts,uid,uri,query not present together in input
```

So the high-level requirement is that `zqd` needs to provide some variation on the current `cut` behavior that returns ZNG that could be re-imported to produce the same results the user was looking at when they clicked Export. It seems like this could be done via a new option on `cut` (which makes sense to me, but I'm open to suggestion) or a whole new processor. I'm not advocating for a change in the default behavior of `cut` (though I'd not necessarily rule it out).

While writing this up, I'm reminded that FWIW, `zeek-cut` is more forgiving. Turning my query result into Zeek TSV and doing the equivalent cut operation, I get all the diversity:

```
$ zq -f zeek "ocsp.pki.goog" all.zng > all.zeek
$ cat all.zeek | /usr/local/zeek-3.1.3/bin/zeek-cut -C ts uid uri query
#separator \x09
#set_separator	,
#empty_field	(empty)
#unset_field	-
#path	http
#fields	ts	uid	uri	
#types	time	string	string	
1521835139.718680	CUkXuv2aZs5NpuHng1	/GTSGIAG3	
#path	dns
#fields	ts	uid		query
#types	time	string		string
1521835139.673223	CkiLMN1OGJBFlZdZZc		ocsp.pki.goog
1521835139.673219	CkiLMN1OGJBFlZdZZc		ocsp.pki.goog
#path	http
#fields	ts	uid	uri	
#types	time	string	string	
1521835105.567982	CMt4qp2hgQEMhqmmj7	/GTSGIAG3	
1521835105.177261	C5YqLp10QwzPgYH5Aa	/GTSGIAG3	
1521835104.054822	CCZseg1hl12M0nc7p6	/GTSGIAG3	
1521835103.652957	C5YqLp10QwzPgYH5Aa	/GTSGIAG3	
1521835103.607992	CCZseg1hl12M0nc7p6	/GTSGIAG3	
#path	dns
#fields	ts	uid		query
#types	time	string		string
1521835103.596102	CVLaYjq85ar0vTxPa		ocsp.pki.goog
1521835103.596097	CVLaYjq85ar0vTxPa		ocsp.pki.goog
```

This speaks to another implementation detail I've pre-discussed with @henridf: How would would the ZNG handle the fields not present? For instance, would the descriptor for the `dns` event contain a `uri` field that gets populated with a `null` value, and likewise would the descriptor for the `http` event contain a `query` field that's populated with `null`? If the implementation can handle it, I would argue the answer to this is "no". Like we just saw with `zeek-cut` it would seem ideal if this new cutting functionality could effectively trim the descriptor of each record to cover only the set of named fields that it contains.

nwt:
ZQL cut emits a record for each input record containing all of the
specified field.  Instead, emit a record if the input contains any of
the specified fields.

Should close #852.

### DIFF
--- a/proc/cutter.go
+++ b/proc/cutter.go
@@ -43,8 +43,11 @@ type Cutter struct {
 	complement  bool
 	cutBuilders map[int]*cutBuilder
 	fieldnames  []string
+	strict      bool
 }
 
+// NewCutter returns a Cutter whose Cut method returns a record if its input
+// record contains any field in fieldnames.
 func NewCutter(zctx *resolver.Context, complement bool, fieldnames []string) *Cutter {
 	return &Cutter{
 		zctx:        zctx,
@@ -52,6 +55,14 @@ func NewCutter(zctx *resolver.Context, complement bool, fieldnames []string) *Cu
 		fieldnames:  fieldnames,
 		cutBuilders: make(map[int]*cutBuilder),
 	}
+}
+
+// NewCutter returns a Cutter whose Cut method returns a record if its input
+// record contains every field in fieldnames.
+func NewStrictCutter(zctx *resolver.Context, complement bool, fieldnames []string) *Cutter {
+	c := NewCutter(zctx, complement, fieldnames)
+	c.strict = true
+	return c
 }
 
 func (c *Cutter) FoundCut() bool {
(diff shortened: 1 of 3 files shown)


proc/cutter.go
```diff
@@ -43,8 +43,11 @@ type Cutter struct {
 	complement  bool
 	cutBuilders map[int]*cutBuilder
 	fieldnames  []string
+	strict      bool
 }
 
+// NewCutter returns a Cutter whose Cut method returns a record if its input
+// record contains any field in fieldnames.
```
henridf, line 50: This comment doesn't work for `complement=true`. How about something like

```suggestion
// NewCutter returns a Cutter whose Cut method returns a record if its input
// record contains any field in fieldname if complement is false. If complement 
// is true, the returned Cutter's Cut method returns a record for every input record.
```

proc/cutter.go
```diff
@@ -54,6 +57,14 @@ func NewCutter(zctx *resolver.Context, complement bool, fieldnames []string) *Cu
 	}
 }
 
+// NewCutter returns a Cutter whose Cut method returns a record if its input
```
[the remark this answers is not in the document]
nwt: Fixed!

proc/cutter.go
nwt: Fixed, I hope!

nwt:
@alfred-landrum and @henridf: I've adjusted the comments. Please take another look at them.

PR #902, opened 2020-06-12
zq: return error of writer.Close

### LINKED ISSUE #900: Failed attempt to write to S3 bucket looks like success
Found while verifying https://github.com/brimsec/zq/issues/771. With `zq` commit `4e6680f`, if I attempt to send the output to an S3 bucket that doesn't exist, `zq` behaves as if it worked.

```
$ echo '{"foo": "bar"}' | zq -t -
#0:record[foo:string]
0:[bar;]

$ echo '{"foo": "bar"}' | zq -t -o s3://this-bucket-does-not-exist/foo.zng -
$ echo $?
0
```

mattnibs:
Because of the way s3manager is written, we may only find out about
an s3 error until after Writer.Close is called. The zq command was
ignoring writer.Close errors. Do not do this.

Closes #900

### DIFF
--- a/cmd/zq/zq.go
+++ b/cmd/zq/zq.go
@@ -259,6 +259,7 @@ func (c *Command) Run(args []string) error {
 		d.SetWarningsWriter(os.Stderr)
 	}
 	if err := driver.Run(mux, d, nil); err != nil {
+		writer.Close()
 		return err
 	}
 	return writer.Close()
--- a/cmd/zq/zq.go
+++ b/cmd/zq/zq.go
@@ -249,16 +249,22 @@ func (c *Command) Run(args []string) error {
 	if err != nil {
 		return err
 	}
-	defer writer.Close()
 	mux, err := driver.CompileWarningsCh(context.Background(), query, reader, false, nano.MaxSpan, zap.NewNop(), wch)
 	if err != nil {
+		writer.Close()
 		return err
 	}
 	d := driver.NewCLI(writer)
 	if !c.quiet {
 		d.SetWarningsWriter(os.Stderr)
 	}
-	return driver.Run(mux, d, nil)
+	if err := driver.Run(mux, d, nil); err != nil {
+		return err
+	}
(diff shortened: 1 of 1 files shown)


cmd/zq/zq.go
```diff
@@ -249,16 +249,22 @@ func (c *Command) Run(args []string) error {
...
 	if !c.quiet {
 		d.SetWarningsWriter(os.Stderr)
 	}
-	return driver.Run(mux, d, nil)
+	if err := driver.Run(mux, d, nil); err != nil {
+		return err
+	}
+	// If writer is an s3 error it could have important information as to
+	// whether the request was written successfully. Cannot do a simple defer
+	// writer.Close() here.
```
alfred-landrum, line 266: I think you can leave out the comment: for any writecloser, it's good practice to not discard the close error. @nwt called this recently on a separate PR of mine, and I think took the code from here; I should have come back to update this as well.

mattnibs: Sounds good.

mattnibs, line 266:
```suggestion
```

mattnibs: Done w/ pr description updated.

cmd/zq/zq.go
```diff
@@ -249,16 +249,19 @@ func (c *Command) Run(args []string) error {
...
+		writer.Close()
 		return err
 	}
 	d := driver.NewCLI(writer)
 	if !c.quiet {
 		d.SetWarningsWriter(os.Stderr)
 	}
-	return driver.Run(mux, d, nil)
+	if err := driver.Run(mux, d, nil); err != nil {
+		return err
```
alfred-landrum, line 262: Missing a writer.Close here.

mattnibs: Whoops.

mattnibs, line 262:
```suggestion
                writer.Close()
		return err
```

PR #908, opened 2020-06-15
Rename zio/tzingio.ReadStats.BadMetaData to BadMetadata

nwt:
Metadata is one word.


nwt:
@mattnibs: No, we're not using this anywhere.

PR #910, opened 2020-06-15
Consolidate Data Type tables

philrz:
_"The ZNG beatings shall continue until ZNG morale improves..."_

This is a try-it-on-for-size PR. I _think_ it's an improvement, but I'm definitely prepared to yield to majority rule here.

I've been mildly bothered by the three tables in the spec that enumerated each of the 19 primitive types, each time to reveal one trait about them. I respect it's somewhat helpful to have the reference info in-line at the appropriate place in the spec. That said, I've also identified the following shortcomings:

1. The top table in [3.1.1 Typdefs](https://github.com/brimsec/zq/blob/master/zng/docs/spec.md#311-typedefs) has always been a bother. When originally formatted with two columns in native Markdown, I found it confusing ([example](https://github.com/brimsec/zq/blob/dd4bce07688bbae9d24346ca22d4bd99ef8a0644/pkg/zng/docs/spec.md#211-typedefs)). I later updated it to use some minimal HTML to improve its presentation, but it's still pretty ugly, and the presence of HTML has confused some editors ([Slack](https://brim-labs.slack.com/archives/CP3BMAEEP/p1579885990052100)). And while a straight one-column format would arguably be the least confusing, it would take up a lot of vertical space to communicate limited info.

2. There have been incidents where the tables got out of sync because someone added an entry in one table, not realizing the others needed the same treatment. Having the info consolidated in one place hopefully eliminates this exposure.

3. I'll soon be wanting to write a ZQL doc that talks about data types, and it would be helpful to be able to link to one place.

I've tried to minimize the cross-reference hassle for the reader by using bullets before the consolidated table to restate how to interpret it. I'm picturing that a reader might just keep the reference table open in a separate window while they're reading the spec.

I also got to thinking about whether this table actually should be in an appendix. One reference online says:

> Appendices are used when the incorporation of material in the body of the work would make it poorly structured or too long and detailed. Appendices may be used for helpful, supporting or essential material that would otherwise clutter, break up or be distracting to the text.

Per my earlier argument, I did feel like the long/repeated tables bordered on "distracting", but they're part of what defines the spec, so I've opted to turn it into its own numbered section. That said, at the same time I _did_ become convinced that the "Related Links" belongs in an appendix, so I've proposed that change here as well.

### DIFF
--- a/zng/docs/spec.md
+++ b/zng/docs/spec.md
@@ -33,7 +33,7 @@
     - [4.3.1 Character Escape Rules](#431-character-escape-rules)
     - [4.3.2 Value Syntax](#432-value-syntax)
   + [4.4 Examples](#44-examples)
-* [5. Data Type Reference](#5-data-type-reference)
+* [5. Primitive Types](#5-primitive-types)
 * [Appendix A. Related Links](#appendix-a-related-links)
 
 ## 1. Introduction
@@ -191,7 +191,7 @@ begin at the value 23 and increase by one for each typedef. These bindings
 are scoped to the stream in which the typedef occurs.
 
 Type IDs for the "primitive types" need not be defined with typedefs and
-are predefined with the IDs shown in the [Data Type Reference](#5-data-type-reference).
+are predefined with the IDs shown in the [Primitive Types](#5-primitive-types) table.
 
 A typedef is encoded as a single byte indicating the container type ID followed by
 the type encoding.  This creates a binding between the implied type ID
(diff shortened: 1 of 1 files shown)


zng/docs/spec.md
```diff
@@ -763,7 +687,36 @@ An unset value indicates a field of a `record` that wasn't set by the encoder:
 ```
 e.g., the North Pole has a latitude but no meaningful longitude.
 
-## 5. Related Links
+## 5. Data Type Reference
```
nwt, line 690: This section deals exclusively with primitive types, and it doesn't just serve as a reference—it defines them—so I'd title it "Primitive Types".

philrz: @nwt: That's a good suggestion. I've pushed a commit to make that adjustment in the section header as well as the TOC and the three places in the spec that link to here.

PR #911, opened 2020-06-16
create brim python package

alfred-landrum:
This change creates a brim python package, and adds supporting scripts to build a wheel for distribution. For Linux usage, there's support to create manylinux wheels via the PEP513 standard docker build process. 

On a mac, you should be able to run `make build-python-wheel`, and then `pip3 install dist/brim*.whl`. That should be enough to run the single example script, python/examples/zql.py, to apply a zql expression to a file.

For linux, you can perform the same process, although that won't produce a manylinux tagged wheel. To create a manylinux wheel, you'll need to have docker available, then run python/manylinux/docker.sh.

There's not yet support to tag the wheels with a particular version number, or to generate the wheels via CI.


Makefile
```diff
@@ -80,16 +80,23 @@ create-release-assets:
 		zip -r release/$${d}.zip $${d} ; \
 	done
 
-build-pyzq:
-	@mkdir -p dist/pybuild
-	@go build -buildmode=c-shared -o dist/pybuild/zq.so ./python/zq.go
+build-python-wheel: build-python-lib
+	pip3 wheel --no-deps -w dist ./python
+
+build-python-lib:
```
alfred-landrum, line 86: Note: I could not find a way to have the archive live under `dist/...` and still have it usable by the pip wheel / setup.py bdist_wheel process, hence the archive is created under `python/`.

henridf:
I tried this out on my macbook, and get:
```sh
$ pip install dist/brim-0.0.1-cp37-cp37m-macosx_10_15_x86_64.whl 
brim-0.0.1-cp37-cp37m-macosx_10_15_x86_64.whl is not a supported wheel on this platform.
```
happy to provide any debugging information of course.

alfred-landrum:
> I tried this out on my macbook, and get:
I asked @henridf to try `pip3 install ...`, and that worked. I have the makefile using pip3 to produce the wheel by default; I haven't tried (and am not planning) to do any testing with Python v2; all the dev has been towards Python v3 versions.

python/setup.py
```diff
@@ -0,0 +1,17 @@
+from setuptools import setup
+
+setup(
+    name = 'brim',
+    version = '0.0.1',
+    description = 'Brim tooling for ZNG data',
```
nwt, line 6:
```suggestion
    description = 'Brim tools for ZNG data',
```

python/manylinux/build.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+set -o pipefail -ex
+
+# Prepare Go
+GOTAR=go1.13.12.linux-amd64.tar.gz
+cd /
+curl -O https://storage.googleapis.com/golang/$GOTAR
+tar zxvf $GOTAR
```
nwt, line 9:
```suggestion
curl -sS https://storage.googleapis.com/golang/go1.13.12.linux-amd64.tar.gz | tar xvzf - -C /
```

python/manylinux/build.sh
```diff
@@ -0,0 +1,20 @@
...
+export PATH=$PATH:/go/bin
+
+# Build go zqext
+cd /io
+make build-python-lib
```
nwt, line 14:
```suggestion
cd /io
make -C /io build-python-lib
```

python/examples/zql.py
```diff
@@ -0,0 +1,22 @@
...
+    sys.exit(parser.print_help())
+
+kwargs = {
+    'inquery': args.args[0],
+    'infile': args.args[1],
+    'informat': args.i,
+    'outfile': args.args[2],
+    'outformat': args.f,
+}
+brim.zql_file_eval(**kwargs)
```
nwt, line 22:
```suggestion
brim.zql_file_eval(
    inquery=args.args[0],
    infile=args.args[1],
    informat=args.i,
    outfile=args.args[2],
    outformat=args.f,
)
```

PR #913, opened 2020-06-16
groupby: run main loop in separate goroutine

henridf:
This is primarily prepwork for spillable groupby, but it also cleans up existing groupby's operation a little, by no longer emitting empty batches (as can be seen from the test expectation change).

### DIFF
--- a/proc/groupby.go
+++ b/proc/groupby.go
@@ -4,6 +4,7 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
+	"sync"
 
 	"github.com/brimsec/zq/ast"
 	"github.com/brimsec/zq/expr"
@@ -93,7 +94,9 @@ func compileKeyExpr(ex ast.Expression) (expr.ExpressionEvaluator, error) {
 // GroupBy computes aggregations using a GroupByAggregator.
 type GroupBy struct {
 	Base
-	agg *GroupByAggregator
+	agg      *GroupByAggregator
+	once     sync.Once
+	resultCh chan Result
 }
 
 // A keyRow holds information about the key column types that result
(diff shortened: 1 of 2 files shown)


proc/groupby.go
```diff
@@ -187,38 +190,60 @@ func NewGroupBy(c *Context, parent Proc, params GroupByParams) *GroupBy {
...
-			return nil, err
+		if batch == nil {
+			b, err := g.agg.Results(true)
+			g.sendResult(b, err)
+			g.sendResult(nil, nil)
+			return
+		}
+		for k := 0; k < batch.Length(); k++ {
+			err := g.agg.Consume(batch.Index(k))
+			if err != nil {
```
nwt, line 221:
```suggestion
			if err := g.agg.Consume(batch.Index(k)); err != nil {
```

proc/groupby.go
```diff
@@ -187,38 +190,60 @@ func NewGroupBy(c *Context, parent Proc, params GroupByParams) *GroupBy {
...
 		}
-	}
-	batch.Unref()
-	if g.agg.inputSortDir != 0 {
-		batch, err := g.agg.Results(false)
-		if err != nil {
-			return nil, err
+		if batch == nil {
+			b, err := g.agg.Results(true)
+			g.sendResult(b, err)
```
nwt, line 215:
```suggestion
			 g.sendResult(g.agg.Results(true))
```

proc/groupby.go
```diff
@@ -187,38 +190,60 @@ func NewGroupBy(c *Context, parent Proc, params GroupByParams) *GroupBy {
...
-		return g.agg.Results(true)
-	}
-	for k := 0; k < batch.Length(); k++ {
-		err := g.agg.Consume(batch.Index(k))
+	g.once.Do(func() { go g.runLoop(g.resultCh) })
+	r := <-g.resultCh
+	return r.Batch, r.Err
+}
+
+func (g *GroupBy) runLoop(ch chan Result) {
```
nwt, line 205: Nit: The corresponding method on both proc.runnerProc and proc.Split is called run, so maybe follow that convention.

proc/groupby.go
henridf: huh!

proc/groupby.go
```diff
@@ -186,38 +189,58 @@ func NewGroupBy(c *Context, parent Proc, params GroupByParams) *GroupBy {
...
-	}
-	for k := 0; k < batch.Length(); k++ {
-		err := g.agg.Consume(batch.Index(k))
+	g.once.Do(func() { go g.run(g.resultCh) })
+	r := <-g.resultCh
+	return r.Batch, r.Err
+}
+
+func (g *GroupBy) run(ch chan Result) {
+	defer close(ch)
```
nwt, line 205:
```suggestion
	g.once.Do(func() { go g.run() })
	r := <-g.resultCh
	return r.Batch, r.Err
}

func (g *GroupBy) run() {
	defer close(g.resultCh)
```
```

henridf: nicer 👍

PR #918, opened 2020-06-18
Change zng.Record.Ts from field to method

nwt:
That zng.Record.Ts is a field feels odd, and setting it doesn't make sense if it won't be used. Make it a method instead.

Master (87f55e847a91b8c8740938c403f42525591a0e05) on my laptop:
```
$ for i in 1 2 3; do /usr/bin/time ./zq -o /dev/null 'count()' zq-sample-data/zng/conn.zng.gz; done
        1.32 real         1.78 user         0.13 sys
        1.30 real         1.76 user         0.12 sys
        1.32 real         1.77 user         0.12 sys
$ for i in 1 2 3; do /usr/bin/time ./zq -o /dev/null 'max(ts)' zq-sample-data/zng/conn.zng.gz; done
        1.43 real         1.92 user         0.13 sys
        1.42 real         1.90 user         0.13 sys
        1.41 real         1.89 user         0.13 sys
```

This branch (855053c7823ab117e86c948a670207b083909255) on my laptop:
```
$ for i in 1 2 3; do /usr/bin/time ./zq -o /dev/null 'count()' zq-sample-data/zng/conn.zng.gz; done
        1.24 real         1.73 user         0.13 sys
        1.22 real         1.72 user         0.13 sys
        1.24 real         1.73 user         0.13 sys
$ for i in 1 2 3; do /usr/bin/time ./zq -o /dev/null 'max(ts)' zq-sample-data/zng/conn.zng.gz; done
        1.35 real         1.87 user         0.14 sys
        1.35 real         1.87 user         0.13 sys
        1.33 real         1.85 user         0.13 sys
```


henridf:
Nice. Can we get rid of `TypeRecord.TsCol` as part of this?

zng/recordval.go
```diff
@@ -335,6 +320,16 @@ func (r *Record) AccessTimeByColumn(colno int) (nano.Ts, error) {
 	return DecodeTime(zv)
 }
 
+// Ts returns the value of the receiver's "ts" field.  If the field is absent,
+// is null, or has a type other than TypeOfTime, Ts returns nano.MinTs.
+func (r *Record) Ts() nano.Ts {
+	if !r.tsValid {
+		r.ts, _ = r.AccessTime("ts")
```
mattnibs, line 327: I love getting errors when we do not need them but... how do we know there will be nothing wrong with the underlying buffer when we try to access time here?

nwt:
> Nice. Can we get rid of TypeRecord.TsCol as part of this?

@henridf: Done in 8693d83e9fa523df3630a8cbf4a62956bcf47ed2. Any about the zio/ndjsonio.(*typeInfo).newRawFromJSON change at
https://github.com/brimsec/zq/pull/918/commits/8693d83e9fa523df3630a8cbf4a62956bcf47ed2#diff-70a77b62be99c19c798241f5ad35a582?

zng/recordval.go
nwt: @mattnibs: In general, we don't. In practice, if a record survives (*Record).TypeCheck, we do, and all existing zio reader implementations call that method (either directly or via NewRecordCheck).

Are you thinking we shouldn't discard the error here?

mattnibs: maybe at least a panic?

nwt: Like this?
```go
	if !r.tsValid {
		var err error
		r.ts, err = r.AccessTime("ts")
		if err != ErrNoSuchField && err != ErrTypeMismatch {
			panic(err)
		}
		r.tsValid = true
	}
```

PR #919, opened 2020-06-19
Zeek compatibility overhaul

philrz:
As of the opening of this PR, the version of this doc in `master` is mostly remnants from the ZSON days when we were attempting to maintain true backward compatibility with the Zeek TSV format. Therefore it was due for an overhaul to explain current behaviors.

This represents my attempt to rationalize what I observe as the black box treatment of Zeek data as ZNG and how it's written back out again as Zeek by `zq`. Reviewers who know the code are encouraged to contribute better rationalizations, as well as pointing out cases I might have forgotten to include.

Also, if you feel any of the behaviors documented here are undesirable to the point where we wish they did not occur, I'd be open to pausing to revise the ZNG spec and `zq` implementation such that the Zeek compatibility story reads better. The special treatment of `_path` as `string` might be one such area, but personally, I'm also fine leaving the behavior as-is and letting the doc cover the topic.

Finally, https://github.com/brimsec/zq/pull/917 should ideally be merged before this one, since we refer to `enum` here, but the spec is currently slightly inaccurate in terms of how it describes the existing (non-) support for a true ZNG `enum`.

### DIFF
--- a/zng/docs/zeek-compat.md
+++ b/zng/docs/zeek-compat.md
@@ -92,7 +92,9 @@ ZNG implementations (should they exist) may handle these differently.
 ### `double`
 
 As they do not affect accuracy, "trailing zero" decimal digits on Zeek `double`
-values will _not_ be preserved (e.g. `123.4560` becomes `123.456`).
+values will _not_ be preserved when they are formatted into a string, such as
+via the TZNG/Zeek/table output options in `zq` (e.g. `123.4560` becomes
+`123.456`).
 
 ### `set`
 
--- a/zng/docs/zeek-compat.md
+++ b/zng/docs/zeek-compat.md
@@ -139,7 +139,7 @@ _not_ intended to be read or presented as such. Meanwhile, another Zeek
 UTF-8. These details are currently only captured within the Zeek source code
 itself that defines how these values are generated.
 
-ZNG includes a [primtive type](spec.md#5-primitive-types) called `bytes` that's
+ZNG includes a [primitive type](spec.md#5-primitive-types) called `bytes` that's
 suited to storing the former "always binary" case and a `string` type for the
 latter "always printable" case. However, Zeek logs do not currently communicate
 detail that would allow a ZNG implementation to know which Zeek `string` fields
--- a/zng/docs/zeek-compat.md
+++ b/zng/docs/zeek-compat.md
@@ -50,6 +50,14 @@ applicable to handling certain types.
 | [`vector`](https://docs.zeek.org/en/current/script-reference/types.html#type-vector)     | [`array`](spec.md#3112-array-typedef)   | |
 | [`record`](https://docs.zeek.org/en/current/script-reference/types.html#type-record)     | [`record`](spec.md#3111-record-typedef) | See [`record` details](#record) |
 
+* **Note**: The [Zeek data type](https://docs.zeek.org/en/current/script-reference/types.html)
+page describes the types in the context of the
+[Zeek scripting language](https://docs.zeek.org/en/current/examples/scripting/).
+The Zeek types available in scripting are a superset of the data types that may
+appear in Zeek log files. The encodings of the types also differ in some ways
+between the two contexts. However, we link to this reference because there is
+no authoritative specification of the Zeek log format.
+
 ## Example
 
 The following example shows an input log that includes each Zeek data type,
@@ -105,8 +113,7 @@ maintain the order of `set` elements as they originally appeared in a Zeek log.
 
 As they're encountered in common programming languages, enum variables
 typically hold one of a set of predefined constants. While this is
-how Zeek's `enum` type behaves inside the
-[Zeek scripting language](https://docs.zeek.org/en/current/examples/scripting/),
+how Zeek's `enum` type behaves inside the Zeek scripting language,
 when the `enum` type is output in a Zeek log, the log does not communicate
 any such set of "allowed" constants as they were originally defined. Therefore,
 when `zq` reads a Zeek `enum` into ZNG, it defines a
--- a/zng/docs/zeek-compat.md
+++ b/zng/docs/zeek-compat.md
@@ -85,9 +85,9 @@ T	123	456	123.456	1592502151.123456	123.456	smile\xf0\x9f\x98\x81smile	\x09\x07\
 ## Type-Specific Details
 
 As `zq` acts as a reference implementation for ZNG, it's helpful to understand
-the following subtleties of how it reads the following Zeek data types into
-ZNG equivalents and writes them back out again in Zeek log format. Other
-ZNG implementations (should they exist) may handle these differently.
+how it reads the following Zeek data types into ZNG equivalents and writes
+them back out again in Zeek log format. Other ZNG implementations (should they
+exist) may handle these differently.
 
 ### `double`
 
--- a/zng/docs/zeek-compat.md
+++ b/zng/docs/zeek-compat.md
@@ -112,10 +112,10 @@ maintain the order of `set` elements as they originally appeared in a Zeek log.
 ### `enum`
 
 As they're encountered in common programming languages, enum variables
-typically hold one of a set of predefined constants. While this is
+typically hold one of a set of predefined values. While this is
 how Zeek's `enum` type behaves inside the Zeek scripting language,
 when the `enum` type is output in a Zeek log, the log does not communicate
-any such set of "allowed" constants as they were originally defined. Therefore,
+any such set of "allowed" values as they were originally defined. Therefore,
 when `zq` reads a Zeek `enum` into ZNG, it defines a
 [type alias](spec.md#412-type-alias) called `zenum` to use for such a field,
 ultimately treating the value as if it were of the ZNG `string` type. The use
@@ -125,7 +125,7 @@ if/when the field may be later output again in Zeek log format. However, when
 working with the value in ZQL, only `string`-type operations will be possible.
 
 As explained in the [alpha notice in the ZNG specification](spec.md), a true
-ZNG `enum` type with predefined constants has not yet been defined in the spec
+ZNG `enum` type with predefined values has not yet been defined in the spec
 nor implemented in `zq`. Once available in ZNG, Zeek could potentially
 offer direct log output in ZNG format that communicates the full definition of
 an `enum`, including the set of allowed values.
--- a/zng/docs/zeek-compat.md
+++ b/zng/docs/zeek-compat.md
@@ -63,7 +63,7 @@ $ cat zeek_types.log
 #set_separator	,
 #empty_field	(empty)
 #unset_field	-
-#fields	my_bool	my_count	my_int	my_double	my_time	my_interval	my_printable_string	my_bytes_string	my_port	my_addr	my_subnet	my_enummy_set	my_vector	my_record.name	my_record.age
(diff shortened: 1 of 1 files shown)


zng/docs/zeek-compat.md
```diff
@@ -1,138 +1,188 @@
...
+
+As `zq` acts as a reference implementation for ZNG, it's helpful to understand
+the following subtleties of how it reads the following Zeek data types into
+ZNG equivalents and writes them back out again in Zeek log format. Other
+ZNG implementations (should they exist) may handle these differently.
+
+### `double`
+
+As they do not affect accuracy, "trailing zero" decimal digits on Zeek `double`
+values will _not_ be preserved (e.g. `123.4560` becomes `123.456`).
```
henridf, line 95:
```suggestion
values will _not_ be preserved when they are formatted into a string (e.g. `123.4560` becomes `123.456`).
```

this isn't really about doubles but about their string representation, so i'm not entirely sure it belongs next to the other points below.

zng/docs/zeek-compat.md
```diff
@@ -1,138 +1,188 @@
...
+
+### `set`
+
+Because order within sets is not significant, no attempt is made by `zq` to
+maintain the order of `set` elements as they originally appeared in a Zeek log.
+
+### `enum`
+
+As they're encountered in common programming languages, enum variables
+typically hold one of a set of predefined constants. While this is
```
henridf, line 105:
```suggestion
typically hold one of a set of predefined values. While this is
```

zng/docs/zeek-compat.md
```diff
@@ -1,138 +1,188 @@
...
+multiple uses leaves out important detail about the intended interpretation and
+presentation of the bytes that make up the value. For instance, one Zeek
+`string` field may hold arbitrary network data that _coincidentally_ sometimes
+form byte sequences that could be interpreted as prinable UTF-8, but they are
+_not_ intended to be read or presented as such. Meanwhile, another Zeek
+`string` field may be populated such that it will _only_ ever contain printable
+UTF-8. These details are currently only captured within the Zeek source code
+itself that defines how these values are generated.
+
+ZNG includes a [primtive type](spec.md#5-primitive-types) called `bytes` that's
```
alfred-landrum, line 140:
```suggestion
ZNG includes a [primitive type](spec.md#5-primitive-types) called `bytes` that's
```

zng/docs/zeek-compat.md
```diff
@@ -1,138 +1,188 @@
...
-character must be escaped.
-(Such lines can only exist as legacy values and not regular values
-as regular values always begin with an integer descriptor.)
 
+## Type-Specific Details
+
+As `zq` acts as a reference implementation for ZNG, it's helpful to understand
+the following subtleties of how it reads the following Zeek data types into
+ZNG equivalents and writes them back out again in Zeek log format. Other
+ZNG implementations (should they exist) may handle these differently.
```
alfred-landrum, line 90:
```suggestion
As `zq` acts as a reference implementation for ZNG, it's helpful to understand
how it reads the following Zeek data types into
ZNG equivalents and writes them back out again in Zeek log format. Other
ZNG implementations (should they exist) may handle these differently.
```

zng/docs/zeek-compat.md
```diff
@@ -1,138 +1,188 @@
...
+imported. Therefore, knowing the equivalent types will prove useful when
+performing [ZQL](../../zql/README.md) operations such as
+[type casting](../../zql/docs/data-types#example) or looking at the
+data when output as [TZNG](spec.md#4-zng-text-format-tzng).
+
+## Equivalent Types
+
+The following table summarizes which ZNG data type corresponds to each
+[Zeek data type](https://docs.zeek.org/en/current/script-reference/types.html)
+that may appear in a Zeek log. While most types have a simple 1-to-1 mapping
```
nwt, line 31: I think it's worth warning that the linked page describes the Zeek scripting language types, which are a superset of the Zeek log file types with encodings that differ in some places, and that we refer to it because there is no authoritative specification of the Zeek log format.

zng/docs/zeek-compat.md
```diff
@@ -1,138 +1,188 @@
...
-RELATIONSHIP BETWEEN ZEEK AND ZNG FORMAT.
+- [Introduction](#introduction)
+- [Equivalent Types](#equivalent-types)
+- [Example](#example)
+- [Type-Specific Details](#type-specific-details)
+  * [`double`](#double)
+  * [`set`](#set)
+  * [`enum`](#enum)
+  * [`string`](#string)
+  * [`record`](#record)
```
nwt, line 11: Nit: I think alphabetical order is nicer than arbitrary order.

zng/docs/zeek-compat.md
```diff
@@ -1,138 +1,188 @@
...
+suited to storing the former "always binary" case and a `string` type for the
+latter "always printable" case. However, Zeek logs do not currently communicate
+detail that would allow a ZNG implementation to know which Zeek `string` fields
+to store as which of these two ZNG data types. Therefore the single ZNG
+`bstring` type is typically used to hold values that are read from Zeek
+`string`-type fields.
+
+One exception to this is Zeek's `_path` field. As it's a standard field that's
+known to be populated by Zeek's logging system (or populated by `zq` during
+while reading some variations of
```
nwt, line 150:
```suggestion
known to be populated by Zeek's logging system (or populated by `zq` when reading some
```

zng/docs/zeek-compat.md
```diff
@@ -1,138 +1,188 @@
...
+would create an opportunity to assign the appropriate ZNG `bytes` or `string`
+type at the point of origin, depending on what's known about how the field's
+value is intended to be populated and used.
+
+### `record`
+
+Zeek's `record` type is unique in that every Zeek log line effectively _is_ a
+record, with its schema defined via the `#fields` and `#types` directives in
+the headers of each log file. Unlike what we saw in the
+[example TZNG output](#example), the word "record" is never called out
```
nwt, line 164:
```suggestion
[example TZNG output](#example), the word "record" never appears
```

zng/docs/zeek-compat.md
philrz: @henridf:  I'm inclined to take your suggestion and leave it rather than dropping the section as you seem to be implying you might prefer. Due to my obsessive nature to try and please everybody, lemme try out my logic on you anyway about why I think it's still appropriate. 😄 

1. I agree that it's about string representation. However, as Zeek TSV effectively _is_ string representation, and this doc is about compatibility with that, it seems like it becomes appropriate to address. With other generic UNIX pipeline tools or domain-specific ones like `zeek-cut`, it's usually reasonable for a user to assume that the data they pass through some in/out pipeline is going to be undisturbed unless they invoke some option that's explicitly intended to change the data. Thinking along these lines, I took the role here of a user assuming "I should be able to send my Zeek data into `zq`, no-op it, and get the same Zeek data back, byte-for-byte". If this is seen as something valid to expect, then I'd expect that for any areas where we _don't_ meet that assumption, we should be able to defend our behaviors as beneficial or harmless. That's what I'm trying to do here for both the "cosmetic" cases I cite here for `double` and `set`: We feel it's mathematically defensible what we're doing from an _accuracy_ perspective, so, as a user, they should just expect it. If someone was super offended by it, they could raise it as a concern or pursue their own ZNG implementation that behaves differently.

2. In terms of mathematical arguments, I have flashbacks to high school chemistry class and having the concept of "significant digits" pounded into our heads. That is, we were taught that there truly _is_ a difference between `123.4560` and `123.456` because the former clearly communicates a measurement was taken out to the 4th decimal place and the it happened to fall perfectly on a 0, as opposed to the latter which _might_ have been the same measurement or it might have been subject to rounding to the 3rd decimal place. An obsessive person might defend Zeek's extra zeros on these grounds and say that, while we're free to represent them internally however is necessary to maintain mathematical accuracy, we'd have a responsibility to maintain record of the extra zeros in the event that we're asked to print it back out again in string representation undisturbed. And, of course, I feel we can defensibly counter that the compute and complexity to achieve that would not be worth the effort, especially when we also consider that binary representations can save storage.

Rather than get into all that, I felt like it's worthwhile to just tell the user to expect it so that way they don't waste time digging into it if they ever do a `diff` before/after, and assume nobody will ever complain.

zng/docs/zeek-compat.md
philrz: @alfred-landrum: Thanks for catching that! I'd like the record to show that I _did_ send this page through the W3C Spell Checker at https://www.w3.org/2002/01/spellchecker and it did _not_ spot this error, so, that's the last time I'll use that turd of a tool. 😉

zng/docs/zeek-compat.md
philrz: 👍

zng/docs/zeek-compat.md
philrz: @nwt: Hmm. FWIW, the order I chose wasn't wholly arbitrary, though I would accept it wasn't at all intuitive. [ 😄  ] I was effectively getting the easy/cosmetic ones out of the way first, then building my way up to the hairier ones. I doubt that was benefitting the reader in any way, so I'm open to changing it.

I'm not sure if this was obvious, but the _top_ table was intentionally put in the same order as the [Zeek Data Types](https://docs.zeek.org/en/current/script-reference/types.html) I was referencing. If that's seen as a reasonable thing to do, it seems I could extend that logic and make my "Additional Detail" sections follow the same order. Or... alphabetical, as you suggested.

Since you framed it as a nit, I know I'm probably already spending more time thinking about this as necessary. But you've opened my mind to changing the order, so now I'd be curious if you feel strongly enough about alphabetical that you'd be in favor of changing the order of the _top_ table as well, or if you'd think it's a-ok for me to change the Additional Details section to mirror the order of the top table.

zng/docs/zeek-compat.md
philrz: @henridf: I actually used the word "constants" in this way three times in the paragraph but you only flagged the first one. I assume you'd advocate making the change in all places?

FWIW, I only chose this lingo because the top Google hit on "enum type" spoke about it this way (admittedly it was a Java ref), but I'm perfectly happy to flip the wording. I just wanted to make certain there wasn't subtle reason why _this_ ref should change from "constants" to "values" but the other two were still seen as appropriate for some reason.

zng/docs/zeek-compat.md
philrz: @nwt: Thanks for speaking up about this! It did cross my mind, so your comment pushed me over into adding some text. Feel free to give it a read and suggest adjustments. I intentionally opted to put the clarification _below_ the table so as to not interrupt flow.

zng/docs/zeek-compat.md
henridf: There is no reason I just suggested changing this one and not the others - I probably just missed the others. And if you've found other wording that uses "constant", it's entirely ok with me to drop this suggestion.

philrz: Thanks @henridf. I went ahead and standardized on "constants", then. You got me to thinking that "constant" is a word that comes up more in programming language contexts, and in this case we're closer to a log/query context, so "values" does, indeed, feel more appropriate.

zng/docs/zeek-compat.md
philrz: I've opted for re-order the entries in the "Additional Detail" section to match the order they appear in the top table. Happy to revisit in a future pass if you'd like.

zng/docs/zeek-compat.md
philrz: I'm going ahead and merging as-is. Happy to revisit in a future pass if you have any feedback later.

PR #920, opened 2020-06-19
Rename SortFn to CompareFn

henridf:
These functions are comparison functions and don't sort anything.

### DIFF
--- a/expr/sort.go
+++ b/expr/sort.go
@@ -30,12 +30,12 @@ func isNull(val zng.Value) bool {
 // a record with a null value is considered larger than a record with any
 // other value, and vice versa.
 func NewCompareFn(nullsMax bool, fields ...FieldExprResolver) CompareFn {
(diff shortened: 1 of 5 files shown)


expr/sort.go
```diff
@@ -29,7 +29,7 @@ func isNull(val zng.Value) bool {
 // is governed by the nullsMax parameter.  If this parameter is true,
 // a record with a null value is considered larger than a record with any
 // other value, and vice versa.
-func NewSortFn(nullsMax bool, fields ...FieldExprResolver) SortFn {
+func NewCompareFn(nullsMax bool, fields ...FieldExprResolver) CompareFn {
```
alfred-landrum, line 32: Nit: could also change the usage of `sorters`, `sorter`, `sf` , and other var names that are derived from 'sort' in this file as well.

henridf: 👍

Issue #791, opened 2020-05-15
Enforce unique constraints on space names


mattnibs:
Currently we allow spaces to:
1. Have duplicate names (all spaces have unique ids)
2. Contain any unicode character

In light of the `zapi` cli we have seen that it useful to have a human readable, unique handle for spaceIDs. As such change space name so that:
1. They are unique among all spaces in a zqd instances
2. Containing only characters `a-zA-Z0-9_`

This will also have to include some undocumented endpoint for migrating old space name/ids to this new format (i.e. old brim spaces).

alfred-landrum:
Reopening this as there's a need to discuss the name character restriction impact on the Brim UI.

philrz:
Verified in Brim commit `cd25120` talking to `zqd` commit `ac2ea5e`. See https://github.com/brimsec/zq/issues/937#issuecomment-650311707.

Thanks @mattnibs and @alfred-landrum!

PR #926, opened 2020-06-21
add dot and underscore to allowed space chars

alfred-landrum:
Add "." and "_"  to the allowed characters for space names to allow Brim integration tests to pass. I've verified the Brim integration tests pass locally, will put up a draft Brim PR showing tests passing in CI.

Related to https://github.com/brimsec/zq/issues/791 ; we may further change the space name restriction after this as well.


alfred-landrum:
Brim CI tests passed (with the exception of a windows failure due to zq#613), so merging.

PR #928, opened 2020-06-21
retry handling zar import for zqd archive spaces

alfred-landrum:
https://github.com/brimsec/zq/pull/864 was reverted due to a CI failure that showed up after the merge. After many passing CI runs with debugging info, I saw a single failure, with surprising behavior: the metadata json file for the archive didn't appear to have an updated modification time immediately after replacing it with a new version with added logs. (The sequence of operations was: create temp file, write temp file, close temp file, rename temp to final, stat final to get mtime. And yet, the mtime on final appeared to be the same as before this sequence.)

Since that failure was hard to repro, I can't be sure that this resolves the problem. However, it uses the original reviewed commit from #864, and then modifies it such that:
* it uses fs.MarshalJSONFile, which uses fs.ReplaceFile, which issues a Sync on the file descriptor before closing it.
* it simplifies the metadata writing code, so that mtime tracking isn't needed there.

Closes https://github.com/brimsec/brim/issues/883


archive/schema.go
```diff
@@ -107,25 +81,104 @@ func (c *CreateOptions) toMetadata() *Metadata {
...
+func (ark *Archive) mdPath() string {
+	return filepath.Join(ark.Root, metadataFilename)
+}
+
+// UpdateCheck looks at the archive's metadata file to see if it
+// has been written to since last read; if so, it is read and the
+// available spans are updated. A counter is returned, starting
+// from 1, which is incremented every time this Archive has re-read
+// the metadata file, and possibly updated the available spans.
+func (ark *Archive) UpdateCheck() (int, error) {
```
mattnibs, line 146: Why not just return mdModTime? What does having the counter (and thus an extra piece of state) give us?

mattnibs: Oh, I'm guessing this for testing purposes?

archive/walk.go
```diff
@@ -38,7 +38,14 @@ func Walk(ark *Archive, visit Visitor) error {
 type SpanVisitor func(si SpanInfo, zardir string) error
 
 func SpanWalk(ark *Archive, v SpanVisitor) error {
-	for _, s := range ark.Spans {
+	if _, err := ark.UpdateCheck(); err != nil {
+		return err
+	}
+
+	ark.mu.RLock()
+	defer ark.mu.RUnlock()
```
mattnibs, line 46: There's a race here where the spans could be updated in between the `UpdateCheck` and `RLock`. Probably not the end of the world if the SpanWalk is a little stale/out of date, just wanted to make sure this was the intended level of consistency.

alfred-landrum: `... eventually consistent` - yep, exactly

archive/schema.go
alfred-landrum: I did want a counter to aid with testing. I also thought it was a better api to return an integer instead of a date, as the integer was simpler to interpret ('if its not the same as last time, revoke your cache').

PR #930, opened 2020-06-23
Simplify reducer/compile.Row

henridf:


reducer/compile/row.go
```diff
@@ -10,36 +10,19 @@ import (
...
-		if r.Reducers[k] != nil {
-			continue
-		}
-		red := r.Defs[k].Instantiate()
-		r.Reducers[k] = red
-		r.n++
+func NewRow(defs []CompiledReducer) Row {
+	reducers := make([]reducer.Interface, len(defs))
+	for k := range defs {
+		reducers[k] = defs[k].Instantiate()
```
alfred-landrum, line 18: truly a nit:
```suggestion
	for i := range defs {
		reducers[i] = defs[i].Instantiate()
```

PR #932, opened 2020-06-23
Spillable groupby

### LINKED ISSUE #800: Spillable groupby
Groupby should be able to spill its table to disk.

henridf:
When the groupby table grows to a certain size (defined by the existing `-limit` parameter), if reducers are decomposable, spill to disk rather than giving up. Spills are sorted and then merged when reading. 

closes #800


proc/combiner.go
```diff
@@ -0,0 +1,125 @@
+package proc
```
henridf, line 1: I put this in proc for now, but in a follow-on PR I think we could replace zbuf.Combiner with it. This could also be used by sort.

proc/combiner.go
```diff
@@ -0,0 +1,125 @@
...
+	return minrec, mink, nil
+}
+
+func (c *Combiner) Read() (*zng.Record, error) {
+	minrec, mink, err := c.peek()
+	if minrec == nil {
+		return nil, err
+	}
+	recs := []*zng.Record{minrec}
+	// Assemble a slice with all head-of-line records that are
```
henridf, line 94: In a follow-on PR we could use a heap (like the sort proc does).

reducer/last.go
```diff
@@ -18,8 +18,16 @@ func (lp *LastProto) Instantiate() Interface {
...
+func (lp *LastProto) TargetResolver() expr.FieldExprResolver {
+	return lp.tresolver
+}
+
+func NewLastProto(target string, tresolver, resolver expr.FieldExprResolver) *LastProto {
+	return &LastProto{
+		target:    target,
+		tresolver: tresolver,
+		resolver:  resolver,
+	}
```
nwt, line 30: We usually put constructors immediately after the type definition, not after its methods.

reducer/first.go
```diff
@@ -18,8 +18,16 @@ func (fp *FirstProto) Instantiate() Interface {
...
+func (fp *FirstProto) TargetResolver() expr.FieldExprResolver {
+	return fp.tresolver
+}
+
+func NewFirstProto(target string, tresolver, resolver expr.FieldExprResolver) *FirstProto {
+	return &FirstProto{
+		target:    target,
+		tresolver: tresolver,
+		resolver:  resolver,
+	}
```
nwt, line 30: We usually put constructors immediately after the type definition, not after its methods.

reducer/field/field.go
```diff
@@ -13,21 +13,30 @@ type Streamfn interface {
...
 
-func NewFieldProto(target string, resolver expr.FieldExprResolver, op string) *FieldProto {
-	return &FieldProto{target, op, resolver}
+func NewFieldProto(target string, tresolver, resolver expr.FieldExprResolver, op string) *FieldProto {
+	return &FieldProto{
+		target:    target,
+		tresolver: tresolver,
+		resolver:  resolver,
+		op:        op,
+	}
```
nwt, line 39: We usually put constructors immediately after the type definition, not after its methods.

reducer/countdistinct.go
```diff
@@ -22,8 +22,16 @@ func (cdp *CountDistinctProto) Instantiate() Interface {
...
+func (cdp *CountDistinctProto) TargetResolver() expr.FieldExprResolver {
+	return cdp.tresolver
+}
+
+func NewCountDistinctProto(target string, tresolver, resolver expr.FieldExprResolver) *CountDistinctProto {
+	return &CountDistinctProto{
+		target:    target,
+		tresolver: tresolver,
+		resolver:  resolver,
+	}
```
nwt, line 34: We usually put constructors immediately after the type definition, not after its methods.

reducer/count.go
```diff
@@ -7,20 +7,28 @@ import (
...
 }
 
-func NewCountProto(target string, resolver expr.FieldExprResolver) *CountProto {
-	return &CountProto{target, resolver}
+func NewCountProto(target string, tresolver, resolver expr.FieldExprResolver) *CountProto {
+	return &CountProto{
+		target:    target,
+		resolver:  resolver,
+		tresolver: tresolver,
+	}
```
nwt, line 31: We usually put constructors immediately after the type definition, not after its methods.

reducer/compile/row.go
```diff
@@ -1,12 +1,16 @@
...
 import (
+	"errors"
+
 	"github.com/brimsec/zq/reducer"
 	"github.com/brimsec/zq/zcode"
 	"github.com/brimsec/zq/zng"
 	"github.com/brimsec/zq/zng/resolver"
 )
 
+var ErrDecompose = errors.New("reducer row doesn't decompose")
```
nwt, line 12: Does adding this to the package's public interface meet a need or yield a benefit? If not, just do `return errors.New("reducer row doesn't decompose")` below.

reducer/compile/row.go
```diff
@@ -26,6 +30,21 @@ func (r *Row) Consume(rec *zng.Record) {
...
 
+func (r *Row) ConsumePart(rec *zng.Record) error {
+	for i, red := range r.Reducers {
+		dec, ok := red.(reducer.Decomposable)
+		if !ok {
+			return ErrDecompose
+		}
+		resolver := r.Defs[i].TargetResolver()
+		err := dec.ConsumePart(resolver(rec))
+		if err != nil {
```
nwt, line 41:
```suggestion
		if err := dec.ConsumePart(resolver(rec)); err != nil {
```

reducer/avg.go
```diff
@@ -9,20 +9,28 @@ import (
...
 }
 
-func NewAvgProto(target string, field expr.FieldExprResolver) *AvgProto {
-	return &AvgProto{target, field}
+func NewAvgProto(target string, tresolver, resolver expr.FieldExprResolver) *AvgProto {
+	return &AvgProto{
+		target:    target,
+		tresolver: tresolver,
+		resolver:  resolver,
+	}
```
nwt, line 33: We usually put constructors immediately after the type definition, not after its methods.

proc/groupby.go
```diff
@@ -344,7 +475,13 @@ func (g *GroupByAggregator) Consume(r *zng.Record) error {
 	row, ok := g.table[string(keyBytes)]
 	if !ok {
 		if len(g.table) >= g.limit {
-			return errTooBig(g.limit)
+			if !g.decomposable || g.spillManager == nil {
+				return errTooBig(g.limit)
+			}
+			err := g.spillTable()
+			if err != nil {
```
nwt, line 482:
```suggestion
			if err := g.spillTable(); err != nil {
```

proc/combiner.go
```diff
@@ -0,0 +1,125 @@
...
+		readers: readers,
+		hol:     make([]*zng.Record, len(readers)),
+		done:    make([]bool, len(readers)),
+	}
+}
+
+// Readers can be dynamically added during combiner operation. It is
+// the client's responsibility to ensure that the new reader's records
+// are all greater than or equal to the biggest record of the existing
+// readers.
```
mattnibs, line 50: I would also add a note of how this is not threadsafe with Read operations.

proc/combiner.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/brimsec/zq/expr"
+	"github.com/brimsec/zq/zbuf"
+	"github.com/brimsec/zq/zng"
+)
+
+type MergeFunc func(*zng.Record, ...*zng.Record) (*zng.Record, error)
+
+// Combiner reads from two or more sorted inputs, implementing a new
+// zbuf.Reader from the inputs. It uses a comparison function to
+// merge-sort its outputs. If inputs are all sorted according to ths
```
mattnibs, line 15:
```suggestion
// merge-sort its outputs. If inputs are all sorted according to the
```

proc/combiner.go
```diff
@@ -0,0 +1,125 @@
...
+			c.hol[k] = nil
+			continue
+		}
+		if c.hol[k] != nil && c.compare(recs[0], c.hol[k]) == 0 {
+			recs = append(recs, c.hol[k])
+			c.hol[k] = nil
+		}
+	}
+	return c.merge(recs[0], recs[1:]...)
+}
```
mattnibs, line 108: If the output of Merge changes the record, won't this result in different return values for Peek() vs Read()? Can't you just write this as a regular zbuf.Reader and wrap this with a zbuf.Peeker if you want Peek functionality as well?

henridf: I did not know about zbuf.Peeker! Will switch to that.

proc/groupby.go
```diff
@@ -125,16 +134,54 @@ type GroupByAggregator struct {
...
+	spillManager *spillManager
+	combiner     *Combiner
+}
+
+type spillManager struct {
+	tempDir string
+	n       int
+}
+
+func newSpillManager() (*spillManager, error) {
```
alfred-landrum, line 158: I don't see a call to remove the temporary directory created here at the end of the groupby operation.

reducer/avg.go
henridf: This layout was here long before these changes (I guess that's why you wrote "usually" 🙂), but I'll fix it in this PR.

(Same response for all the other identical comments.)

proc/groupby.go
henridf: Oops, I had this (and other cleanup) in a stash that i forgot to apply before putting up the PR.  c1a14ba

proc/combiner.go
henridf: Actually the issue here is that `Peek()` is a bad name. The purpose of this function is to find out what the current minimum record (under the combiner's comparison function) is. In the general case this minimum record is **not** equal to the record that will be returned by the next `Read()`, and so this is not a Peek() in the usual sense. I've renamed it to PeekMin() instead.

Thanks for flagging this!

proc/combiner.go
mattnibs: Withdrawn.

proc/combiner.go
```diff
@@ -0,0 +1,127 @@
...
+type Combiner struct {
+	merge   MergeFunc
+	compare expr.CompareFn
+	readers []zbuf.Reader
+	done    []bool
+	hol     []*zng.Record
+}
+
+// NewCombiner returns a new combiner
+func NewCombiner(readers []zbuf.Reader, c expr.CompareFn, m MergeFunc) *Combiner {
```
mattnibs, line 34: Instead of a creating another version of Combiner why not use sort's `runManager` and make a separate zbuf.Reader that merges adjacent equal records? Otherwise feels like we're adding a lot of code that already exists elsewhere.

proc/groupby.go
```diff
@@ -161,26 +213,99 @@ func NewGroupByAggregator(c *Context, params GroupByParams) *GroupByAggregator {
...
+	rs := expr.NewCompareFn(true, resolvers...)
+	if params.inputSortDir < 0 {
+		keysCompare = func(a, b *zng.Record) int { return rs(b, a) }
+	} else {
+		keysCompare = rs
+	}
+	sm, err := newSpillManager()
+	if err != nil {
+		c.Logger.Warn("groupby: could not create spill manager", zap.Error(err))
+	}
```
mattnibs, line 234: This call creates a tempDir. I don't we should be creating a temp dir until a groupby needs to spill to disk.

proc/combiner.go
henridf: (Nit: this isn't another version of Combiner as it replaces the zdx one. Also, as noted above, I think this one can be unified with the single other one, which is currently in `zbuf`).

Good question. One notable difference is that `runManager` always waits for all records to be written before it starts to read. Here, we need to be able to start reading before all records are written, which leads to `combiner.PeekMin()` and `combiner.AddReader()`. I don't think those can be easily added to runManager.

With that said, your point about having merging take place outside of the combiner is a good one. I thought of that along the way but went with the path of least resistance which was to do it the way zdx.Combiner did. But we can change that if it's nicer.

proc/groupby.go
henridf: good point, will fix

proc/combiner.go
henridf: > With that said, your point about having merging take place outside of the combiner is a good one.

I've done this in eb1d845d9d0e2978aabbd6d01a2f034bb77854a6.

proc/groupby.go
```diff
@@ -353,56 +502,151 @@ func (g *GroupByAggregator) Consume(r *zng.Record) error {
...
 // Results returns a batch of aggregation result records. If the input
-// is sorted in the primary grouping key, this can be called multiple times;
-// all completed keys at the time of the invocation are returned (but
-// not necessarily in their input sort order). A final call with
-// eof=true should be made to get the final keys.
-//
-// If the input is not sorted, a single call (with eof=true) should be
-// made after all records have been Consumed()'d.
+// is sorted in the primary key, only keys that are completed are
+// returned (but not necessarily in their input sort order). A final
```
alfred-landrum, line 558: `... not necessarily in their input sort order`: I see that g.keycompare is created based on the input sort direction, and it's used to sort the output per-batch records, so it seems like this comment isn't accurate? Or am I misunderstanding?

henridf: say you're doing `every 1s count() by ip`, and the input is sorted in `ts`, then what i was trying to express here is that the output will also be sorted in `ts`, but the output order of the ips may not match the input order. For example, with this input:

```
ts:1, ip:1.1.1.1
ts:1, ip:1.1.1.2
ts:2, ip:1.1.1.1
```

you might get this output where 1.1.1.2 is ahead of 1.1.1.1:

```
ts:1, count:1, ip:1.1.1.2
ts:1, count:1, ip:1.1.1.1
ts:2, count:1, ip:1.1.1.1
```

Maybe I should just drop the `(but not necessarily in their input sort order)` from the comment as I now I think of it I don't expect one would have any sortedness expectations here?

alfred-landrum: ah, thanks for the clarification; yes, I think you can just drop that portion.

proc/groupby.go
```diff
@@ -353,56 +502,151 @@ func (g *GroupByAggregator) Consume(r *zng.Record) error {
 	return nil
 }
 
-func (g *GroupByAggregator) updateMaxKey(v zng.Value) {
-	if g.maxKey == nil {
-		g.maxKey = &v
+func (g *GroupByAggregator) spillTable(eof bool) error {
+	parts, err := g.readTable(true, true)
```
alfred-landrum, line 506:
```suggestion
	batch, err := g.readTable(true, true)
```

proc/groupby.go
```diff
@@ -353,56 +502,151 @@ func (g *GroupByAggregator) Consume(r *zng.Record) error {
...
+	if err != nil {
+		return err
+	}
+	if parts == nil {
```
alfred-landrum, line 510:
```suggestion
	if batch == nil {
```

proc/groupby.go
```diff
@@ -353,56 +502,151 @@ func (g *GroupByAggregator) Consume(r *zng.Record) error {
...
+		return nil
+	}
+	recs := parts.Records()
```
alfred-landrum, line 513:
```suggestion
	recs := batch.Records()
```

proc/groupby.go
```diff
@@ -353,56 +502,151 @@ func (g *GroupByAggregator) Consume(r *zng.Record) error {
...
+
+	if !eof && g.inputSortDir != 0 {
+		keyVal, err := g.keys[0].expr(recs[len(recs)-1])
+		if err != nil && !errors.Is(err, zng.ErrUnset) {
+			return err
+		}
+		g.updateMaxSpillKey(keyVal)
+	}
+
+	spill, err := g.spillManager.writeSpill(g.zctx, parts)
```
alfred-landrum, line 524:
```suggestion
	spill, err := g.spillManager.writeSpill(g.zctx, batch)
```

proc/groupby.go
```diff
@@ -413,10 +657,13 @@ func (g *GroupByAggregator) records(eof bool) ([]*zng.Record, error) {
...
 	}
-	return recs, nil
+	if len(recs) == 0 {
+		return nil, nil
+	}
+	return zbuf.NewArray(recs), nil
 }
 
-func (g *GroupByAggregator) lookupRowType(row *GroupByRow) (*zng.TypeRecord, error) {
+func (g *GroupByAggregator) lookupRowType(row *GroupByRow, part bool) (*zng.TypeRecord, error) {
```
alfred-landrum, line 666:
```suggestion
func (g *GroupByAggregator) lookupRowType(row *GroupByRow, decompose bool) (*zng.TypeRecord, error) {
```

proc/groupby.go
```diff
@@ -430,7 +677,16 @@ func (g *GroupByAggregator) lookupRowType(row *GroupByRow) (*zng.TypeRecord, err
 	}
 	cols = append(cols, g.builder.TypedColumns(types)...)
 	for k, red := range row.reducers.Reducers {
-		z := red.Result()
+		var z zng.Value
+		if part {
```
alfred-landrum, line 681:
```suggestion
		if decompose {
```

PR #935, opened 2020-06-24
More ZNG spec fixes

philrz:
After polishing the [Zeek compatibility](https://github.com/brimsec/zq/blob/master/zng/docs/zeek-compat.md) doc, I came back to the ZNG spec yet again with intent to see if more de-Zeek-ification was now possible, and I ended up finding more fundamental things to scrub, including removal of a couple remaining relics from ZSON days that don't seem easy to resuscitate. Other than fixes for correctness, the most significant changes are in the spirit of trying to have the spec be in sync with the current `zq` implementation... _or_ calling out explicitly where the two are known to not be in sync.


zng/docs/spec.md
```diff
@@ -1,12 +1,15 @@
 # ZNG Specification
 
-> Note: This specification is ALPHA and a work in progress.
-> Zq's implementation of ZNG is tracking this spec and as it changes,
+> ### Note: This specification is ALPHA and a work in progress.
```
philrz, line 3: By using `###` to make this into a header, I can link directly to it from within this doc or the "Zeek compatibility" one. I'd previously been linking to the "ZNG Specification" header since the alpha notice was already effectively at the top, but I figured it doesn't hurt to be explicit.

zng/docs/spec.md
```diff
@@ -1,12 +1,15 @@
...
+> [Zq](https://github.com/brimsec/zq/blob/master/README.md)'s
+> implementation of ZNG is tracking this spec and as it changes,
 > the zq output format is subject to change.  In this branch,
 > zq attempts to implement everything herein excepting:
 >
 > * the `bytes` type is not yet implemented,
-> * the `enum` type is not yet implemented.
+> * the `enum` type is not yet implemented,
+> * only streams of `record` types (which may consist of any combination of
+>   other implemented types) may currently be expressed in value messages.
```
philrz, line 12: I'm interested in feedback on if I'm capturing this correctly. I've confirmed via conversations & black box testing that `zq` doesn't support several of the examples here that are _not_ streams of records, e.g. a super simple example:

```
$ cat not-record.tzng 
#0:string
0:hello, world;

$ zq -t -i tzng not-record.tzng 
not-record.tzng: line 1: zng typedef not a record while parsing: "string"
```

Does this additional disclaimer covers the territory?

zng/docs/spec.md
```diff
@@ -43,17 +46,15 @@ of heterogeneously typed records, e.g., structured logs, where filtering and
...
+ZNG has a binary form called "ZNG" as well as text form called "TZNG" that's
+comprised of a sequence of newline-delimited UTF-8 strings.
 
 ZNG is richly typed and thinner on the wire than JSON.
-Like [newline-delimited JSON (NDJSON)](http://ndjson.org/),
-the ZNG text format represents a sequence of data objects
+ZNG strikes a balance between the narrowly typed but flexible
+[newline-delimited JSON (NDJSON)](http://ndjson.org/) format and
+a more structured approach like [Apache Avro](https://avro.apache.org).
+Like NDJSON, the TZNG text format represents a sequence of data objects
```
philrz, line 56: I've re-ordered the text here to match the flow of the overall doc: We lead with binary ZNG because it's the more compact/efficient representation & the default output of `zq`.

zng/docs/spec.md
```diff
@@ -74,8 +75,7 @@ ZNG is more expressive than JSON in that any JSON input
 can be mapped onto ZNG and recovered by decoding
 that ZNG back into JSON, but the converse is not true.
 
-The ZNG design [was motivated by](./zeek-compat.md)
-and is compatible with the
+The ZNG design was motivated by and [is compatible with](./zeek-compat.md) the
```
philrz, line 78: I'm just attaching the link to better words here: The "Zeek compatibility" doc is more about compatibility than motivation.

zng/docs/spec.md
```diff
@@ -133,8 +133,8 @@ needed to define the schema for each distinct record.  To define a new
 type, the "#" syntax is used.  For example,
 logs from the open-source Zeek system might look like this
 ```
-#24:record[_path:string,ts:time,uid:string,id:record[orig_h:ip,orig_p:port,resp_h:ip,resp_p:port]...
-#25:record[_path:string,ts:time,fuid:string,tx_hosts:set[ip]...
+#24:record[_path:string,ts:time,uid:bstring,id:record[orig_h:ip,orig_p:port,resp_h:ip,resp_p:port]...
+#25:record[_path:string,ts:time,fuid:bstring,tx_hosts:set[ip]...
```
philrz, line 137: This adjustment was just for correctness, since for some time we've been representing every Zeek `string` field other than `_path` as a `bstring`.

zng/docs/spec.md
```diff
@@ -180,8 +180,7 @@ These payloads are guaranteed to be preserved
 in order within the stream and presented to higher layer components through
 any ZNG streaming API.  In this way, senders and receivers of ZNG can embed
 protocol directives as ZNG control payloads rather than defining additional
-encapsulating protocols.  See the
-[zng-over-http](zng-over-http.md) protocol for an example.
+encapsulating protocols.
```
philrz, line 183: The "ZNG over HTTP" doc as it stands today is yet another artifact from a "ZSON" time when things were all text-based and we were attempting genuine backward-compatibility with Zeek TSV. While it's surely helpful to point to a reference implementation for anything described in the spec, since the [zeek-tsv-http-plugin](https://github.com/brimsec/zeek-tsv-http-plugin) is not ZNG in the modern sense of the word, it seems like we should probably do away with it. Opinions?

zng/docs/spec.md
```diff
@@ -454,7 +453,7 @@ are values.
 
 ### 4.1 Control Messages
 
-TZNG control messages have one of four forms defined below.
+TZNG control messages have one of three forms defined below.
```
philrz, line 456: The fourth (when it existed) was "Ordering Hints" which are now gone.

zng/docs/spec.md
```diff
@@ -468,17 +467,17 @@ A TZNG type binding has the following form:
 #<type-tag>:<type-string>
 ```
 Here, `<type-tag>` is a string decimal integer and `<type-string>`
-is a string defining a record type (`<rtype>`) according to the [TZNG type grammar](#42-type-grammar). They create
+is a string defining a type (`<type>`) according to the [TZNG type grammar](#42-type-grammar). They create
```
philrz, line 470: @henridf: It was only in this review pass that I realized the full implications of what you'd helpfully added in https://github.com/brimsec/zq/pull/897/commits/0a39ecc53ff175df586705dcbcb8897b0d04d43a. While the reference to `<rtype>` (which I was in favor of at the time 😄 ) got the spec in sync with the `zq` implementation, being fully in sync between the two would also imply purging most of the examples from Section 4.4. After discussing with Steve & thinking about it myself I now recognize that would be going a little too far. So it still seems like your enhancements to the pseudo-grammar are very much helpful, but I _think_ relaxing this to just `<type>` makes the examples still legit (though still not implemented in `zq`). Am I correct about that?

zng/docs/spec.md
```diff
@@ -468,17 +467,17 @@ A TZNG type binding has the following form:
...
 a binding between the indicated tag and the indicated type.
 
 ### 4.1.2 Type Alias
 
 A TZNG type alias has the following form:
 ```
-#<type-name>:<type-string>
+#<type-name>=<type-string>
```
philrz, line 477: I'm shocked I let this one elude me for so long, but it looks to me like the spec was dead wrong before:

```
$ zq -t conn.log.gz | head -2
#zenum=string
#0:record[_path:string,ts:time,uid:bstring,id:record[orig_h:ip,orig_p:port,resp_h:ip,resp_p:port],proto:zenum,service:bstring,duration:duration,orig_bytes:uint64,resp_bytes:uint64,conn_state:bstring,local_orig:bool,local_resp:bool,missed_bytes:uint64,history:bstring,orig_pkts:uint64,orig_ip_bytes:uint64,resp_pkts:uint64,resp_ip_bytes:uint64,tunnel_parents:set[bstring]]
```

zng/docs/spec.md
```diff
@@ -648,7 +647,8 @@ and references the types accordingly in three values;
 1:[hello;world;]
 0:this is a semicolon: \x3b;
 ```
-which represents a stream of the following three values:
+which represents a stream of the following three values (expressed as
+pseudocode):
```
philrz, line 651: It's always bothered me a little that what's shown below looks like structured syntax but we never really say what "it" is. If we wanted to reference some kind of lingua franca instead, perhaps we could write the values as JSON & say that's what it is. Opinions?

zng/docs/spec.md
henridf: Yes I think it does!

zng/docs/spec.md
henridf: I agree we should do away with it.

zng/docs/spec.md
henridf: 👍 Yup.

zng/docs/spec.md
henridf: I would just drop the entire section 4.4, seeing as much  it is not currently implemented. If anything I'd say the zng section needs examples more than the tzng section, as tzng is easier to understand from the spec and to manually experiment with.

zng/docs/spec.md
henridf: One addition: the tzng grammar currently allows an empty record type (`record [ ]`). The implementation doesn't. We could call it out as a delta, or get rid of it altogether. I'd opt for the latter.

zng/docs/spec.md
philrz: Thanks for the confirmation @henridf. I've now removed the doc itself and another link to it in a follow-on commit.

zng/docs/spec.md
philrz: In the binary ZNG context in [3.1.1.1 Record Typedef](https://github.com/brimsec/zq/blob/master/zng/docs/spec.md#3111-record-typedef), a closing remark also says:

> A record may not contain zero columns.

...which is effectively another way of saying "no empty records allowed", right? That would mean that as of this moment the TZNG grammar is out of sync with the implementation _and_ the binary ZNG spec, which seems to further validate your instincts that we should just remove `record [ ]` from the TZNG grammar, which I'll do in a commit now.

zng/docs/spec.md
```diff
@@ -507,7 +507,6 @@ grammar describing the textual type encodings is:
          | <rtype>
 
 <rtype> := record [ <columns> ]
```
henridf, line 509:
```suggestion
         | record [ <columns> ]
```

zng/docs/spec.md
```diff
@@ -84,65 +83,10 @@ meta-records, and ZNG merely modernizes this original approach.
...
-#25:record[_path:string,ts:time,fuid:string,tx_hosts:set[ip]...
-24:[conn;1425565514.419939;CogZFI3py5JsFZGik;[192.168.1.1:;80/tcp;192.168.1.2;8080;]...
-25:[files;1425565514.419987;Fj8sRF1gdneMHN700d;[52.218.49.89;52.218.48.169;]...
-```
-Note that the value encoding need not refer to the field names and types as both are 
-completely captured by the type ID.  Values merely encode the value
-information consistent with the referenced type ID.
-
-## 3. ZNG Binary Format (ZNG)
+## 2. ZNG Binary Format (ZNG)
```
philrz, line 89: I expect this may be controversial (cc: @mccanne, as I'm sure this text started with you in the very beginning) but @henridf and I have spoken offline a fair amount about it and I'm ready to advocate for removing Section 2 entirely. Read with fresh eyes, it looks like yet another relic from the ZSON days when things were text-only, so it was fair game to lead off with simple human-readable examples in hopes of pulling the reader along. However, in the time since then, not only do we now lead with binary ZNG as the more important piece, the examples that were in Section 2 that look like TZNG aren't even valid TZNG anymore, i.e. the `9` and `4` shown below are _not_ actually type IDs. To provide an accurate example we'd need to get into the details of tag integers vs. underlying type IDs as we do later in the doc where that feels more justified. If left as-is, we'd effectively be leading off with examples in pseudocode, and I don't think that serves the reader well.Since the other text in this section is already redundant with things covered elsewhere in the spec (Intro, TZNG Examples, etc.), I'm in favor of ripping off the band-aid and removing it.

Since records are the one "value message" that the `zq` reference implementation supports right now (as mentioned in the alpha notice up top) I _did_ put a Zeek example in Section 3.4 to take the place of the one that had been here.

zng/docs/spec.md
philrz: I've talked to @henridf offline about this one. We're both in agreement that it feels a little hazardous to leave so many non-working examples. The line I added to the alpha notice is probably not enough for a first-time reader to know how to stay inside the lines if they start experimenting with `zq` while they're making their way through the spec. I've tried to add a touch more guidance via the Zeek example I moved into this section from the now-deleted "ZNG Data Model" section" and describing it as a "common case", but that still may not be enough. At the risk of being too heavy with disclaimers, maybe one more alpha-style notice would be justified at the start of this Examples section just to call attention to the lag in the implementation and call their attention to the Zeek example as the most representative one. Anyone have opinions?

zng/docs/spec.md
```diff
@@ -648,15 +588,27 @@ and references the types accordingly in three values;
...
-which represents a stream of the following three values:
+which represents a stream of the three values, that could be expressed in JSON
+as:
 ```
-string("hello, world")
-record(a:"hello",b:"world")
-string("this is a semicolon: ;")
+"hello, world"
+{"a": "hello", "b": "world"}
+"this is a semicolon: ;"
```
philrz, line 596: I've gone ahead and shown these as JSON values and called them out as such. When making a comparison I feel like it's slightly more helpful to speak a lingua franca rather than use a pseudocode syntax.

zng/docs/spec.md
nwt: I think it's useful to describe the data model before defining the protocols that implement it, so I think we should keep this section.

philrz: @nwt: Do you feel strongly enough that you're arguing for keeping the pseudocode examples as-is, or are you speaking up more in defense of the descriptive text outside the examples?

My first thoughts were to try and update the section so it could be kept, but as mentioned above, I found that difficult to do that while keeping the not-really-ZNG-and-not-really-TZNG examples. And once I started dropping examples, the remaining text didn't read as well by itself.

One approach I considered was to show the examples in a table (i.e. so there's no mistaking it for TZNG anymore) and be explicit about the fact that we're trying to present a human-readable representation of the soon-to-be-revealed binary format. After keeping the section's first two paragraphs intact as-is, I might take it in this direction:

---
...

For example, the following illustrates a ZNG value message for the single string "hello world" (he actual binary encoding is described in detail in the following section):

| Type ID | Value          |
|---------|------------|
| `9`        | hello, world |

Here, the type ID is the integer `9` representing the string type (defined in Typedefs) and the data value "hello, world" is an instance of the string type.

ZNG gets more interesting when different data types are interleaved in the stream. For example,

| Type ID | Value          |
|---------|------------|
| `9`        | hello, world |
| `4`        | 42               |
| `9`        | there's a fly in my soup! |
| `9`        | no, there's isn't. |
| `4`        | 3 |

where type ID `4` represents a 32-bit signed integer. This encoding represents the sequence of values that could be expressed in JSON as:

```
"hello, world"
42
"there's a fly in my soup!"
"no, there isn't."
3
```

ZNG streams are often comprised as a sequence of records, which works well to provide an efficient representation of structured logs. In this case, a new type defines the schema for each distinct record. For example, the following shows type bindings and values in TZNG for open source Zeek `weird` and `ftp` events:

```
#24:record[_path:string,ts:time,uid:bstring,id:record[orig_h:ip,orig_p:port,resp_h:ip,resp_p:port],name:bstring,addl:bstring,notice:bool,peer:bstring]
24:[weird;1521911720.600843;C1zOivgBT6dBmknqk;[10.47.1.152;49562;23.217.103.245;80;]TCP_ack_underflow_or_misorder;-;F;zeek;]
#25:record[_path:string,ts:time,uid:bstring,id:record[orig_h:ip,orig_p:port,resp_h:ip,resp_p:port],user:bstring,password:bstring,command:bstring,arg:bstring,mime_type:bstring,file_size:uint64,reply_code:uint64,reply_msg:bstring,data_channel:record[passive:bool,orig_h:ip,resp_h:ip,resp_p:port],fuid:bstring]
25:[ftp;1521911724.699488;ChkumY1k35TmZFL0V3;[10.164.94.120;45905;10.47.27.80;21;]anonymous;nessus@nessus.org;PASV;-;-;-;227;Entering Passive Mode (172,20,0,80,200,63).;[T;10.164.94.120;172.20.0.80;51263;]-;]
```

Note that the value encoding need not refer to the field names and types as both are completely captured by the type definition. Values merely encode the value information consistent with the referenced type.

---

That is, I'd also move the Zeek example back into this section, but be explicit in presenting it as TZNG, and it's valid TZNG in this case, so that feels right. In so doing I've also replaced the references to "type ID" since those are not accurate for this TZNG, instead speaking about "type definition" and "referenced type" more in the abstract.

Do you think this is an improvement?

nwt: I'm speaking only in defense of the descriptive text and think the section is fine with or without examples. It's certainly dry without them, but I think it's OK for a specification to be dry.

If you do want to keep the examples, the tables are fine, but I think straight TZNG with minimal explanatory text would better. The table approach effectively introduces a third notation for the data model, and that's something I think we should avoid.

philrz: @nwt: Yes, a "third notation" is precisely what I was trying to get away from, since what was there before was just that: It looked like TZNG, but it wasn't valid TZNG. You're correct that my new "third notation" wasn't great either. But I was nervous that I'd have difficulty crafting "TZNG with minimal explanatory text" since there's some devil in the details about TZNG vs. ZNG that only become fully apparent (to me, at least) after both formats are described in detail (e.g. "Type Bindings" in TZNG vs. "Typedefs" in binary ZNG).

I've pushed yet another commit that shows the examples as correct TZNG while using accurate wording & says just a bit about binary ZNG to nod toward the subtleties that will be explained later in the spec. Let me know if you think it's an improvement. I'd also love to hear from @henridf since he's been along for this ride. Thanks.

zng/docs/spec.md
```diff
@@ -100,47 +100,55 @@ The ZNG type system comprises the standard set of primitive types like integers,
...
+may look like this:
 ```
-9:hello, world
+#35:string
+35:hello, world
 ```
-Here, the type ID is the integer "9" representing the string type
-(defined in [Typedefs](#typedefs)) and the data value "hello, world"
-is an instance of the string type.
+Here, the first line binds a tag `35` to the ZNG `string` data type
```
henridf, line 109: I don't quite follow this paragraph... it seems to imply that in TZNG, you'd _have_ to have a new type definition to use an existing type, which I don't think is the case - it should be possible to just write `9:hello, world` (as per the previous rev).

Now to further complicate things, none of this works at all today since we require records as top level values.... which brings me to my second comment, which is that it seems potentially confusing to lead with things that aren't implemented, as the subset of readers who try off these leading examples only to get error messages might be 😕

henridf: > I don't quite follow this paragraph... it seems to imply that in TZNG, you'd have to have a new type definition to use an existing type, which I don't think is the case - it should be possible to just write 9:hello, world (as per the previous rev).

Ooops. This comment was based on my mixing up tzng with zng, which @philrz helpfully clarified offline. Based on the spec, you cannot just write `9:hello world` in tzng (while the equivalent in zng should be ok, which is what led to my comment).

zng/docs/spec.md
```diff
@@ -620,23 +624,23 @@ type descriptor of the line using the formats shown in the
 
 Here are some simple examples to get the gist of the ZNG text format.
 
-Primitive types look like this and do not need typedefs:
+Primitive types look like this:
```
philrz, line 627: Late-breaking change: Now that I've brought @henridf around to the subtleties of invoking TZNG in the "ZNG Data Model" section [😄  ] I'm realizing the TZNG Examples seems to need one last adjustment.

1. "Typedefs" aren't really "a thing" in TZNG
2. While "Type Bindings" _are_ a thing in TZNG, it's not like I could just do a 1-for-1 swap of the terms, since as we show in the examples, you _would_ need to do a Type Binding first if you wanted to express a standalone Value of a primitive type (this being something not even possible in `zq` today, admittedly)

Therefore, I've opted to just remove the wording that references "typedefs" in this section and otherwise leave the examples as-is. Just showing that the primitive types are written out as-is maybe doesn't communicate much, but it still seems like a legit way to build up to when we start referencing them in the example just below that starts building on type aliases, etc.

henridf: LGTM

zng/docs/spec.md
```diff
@@ -100,47 +100,55 @@ The ZNG type system comprises the standard set of primitive types like integers,
...
+type. TZNG is being shown here for human readability. In the equivalent
+binary ZNG encoding, a single [Value Message](#32-value-messages) can represent
+the string by referencing the type ID of `9` (per the
+[Primitive Types](#5-primitive-types) table) with no leading
+[Typedef](#311-typedefs) required.
 
-ZNG gets more interesting when different data types are interleaved in the stream.
-For example,
+ZNG gets more interesting when different data types are interleaved in the
+stream.  For example, in the following TZNG:
```
nwt, line 118:
```suggestion
stream.  For example, consider this TZNG stream:
```
I think the existing sentence fragment reads a little roughly.

zng/docs/spec.md
```diff
@@ -100,47 +100,55 @@ The ZNG type system comprises the standard set of primitive types like integers,
...
+#36:int64
+36:42
+35:there's a fly in my soup!
+35:no, there isn't.
+36:3
 ```
-where type ID 4 represents an integer.  This encoding represents the sequence of
-values:
+Here the tag `36` now binds to one of ZNG's integer types. This encoding
+represents the sequence of values that could be expressed in JSON as:
```
nwt, line 129:
```suggestion
represents the sequence of values that could be expressed in JSON as
```
A colon doesn't make sense to me here.

zng/docs/spec.md
nwt: > I've pushed yet another commit that shows the examples as correct TZNG while using accurate wording & says just a bit about binary ZNG to nod toward the subtleties that will be explained later in the spec.

I think this section better serves its informational purpose if we omit mention of differences between ZNG and TZNG.

zng/docs/spec.md
```diff
@@ -100,47 +100,55 @@ The ZNG type system comprises the standard set of primitive types like integers,
...
 ```
-ZNG streams are often comprised as a sequence of records, which works well to provide
-an efficient representation of structured logs.  In this case, a new type ID is
-needed to define the schema for each distinct record.  To define a new
-type, the "#" syntax is used.  For example,
-logs from the open-source Zeek system might look like this
+ZNG streams are often comprised as a sequence of records, which works well to
+provide an efficient representation of structured logs. In this case, a new
+type defines the schema for each distinct record. For example, the following
+shows type bindings and values in TZNG for open source `weird` and `ftp`
```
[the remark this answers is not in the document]
philrz: Duh. Yes. Thanks. 😄

zng/docs/spec.md
philrz: Agreed.

zng/docs/spec.md
philrz: Hmm, ok. I flipped through the rest of the spec and in _most_ places we use some kind of punctuation before a code block that shows example text, but I see your point in that sometimes the sentence fragments seem to justify it and sometimes they maybe don't. I spotted one other place where we went straight from an "as" to a block without punctuation, so I'm happy with your suggestion. FWIW there's another place in the Section 4 "Examples" where we do nearly the same lead-in to a bit of comparative JSON, so I'll make the same change there.

zng/docs/spec.md
philrz: @nwt: Agreed. The "differences" text I had there before was my attempt to handle the resistance I was sensing to the removing the standalone `9:hello, world` that used to be there. Since it seems everyone is now on-board why I've needed to include type bindings in order to use human-readable TZNG in this section, it looks like I can remove those two sentences, which I've done by accepting your suggestion. I think I can finally merge!

PR #942, opened 2020-06-25
log when zqd can't bind to its port

### LINKED ISSUE #941: log when zqd cannot bind and exits
It would be convenient for test triage to know that a particular `zqd`
1. couldn't bind to its port
1. is exiting

If you want to know a case for how this would be helpful, https://github.com/brimsec/brim/issues/897 is a long, complex issue. But know that there was effectively an "orphaned" `zqd` process that the Brim instances were talking to, and `zqd` for tests couldn't start.

There are a bunch of `zqd-core` logs only containing a single message like this:
```
{"level":"info","ts":1593101809.98069,"msg":"Starting","datadir":"/var/tmp/brimsec/itest/histogram.test.js/1/data/spaces","pprof_routes":false,"zeek_supported":true}
```
There's no error in here about not being able to bind to the default port. There's no mention in here about `zqd` exiting. Triage would have been faster had these messages been present.

mikesbrown:
fixes #941

### DIFF
--- a/pkg/httpd/server.go
+++ b/pkg/httpd/server.go
@@ -39,6 +39,7 @@ func (s *Server) Start(ctx context.Context) error {
 	s.srv.BaseContext = func(l net.Listener) context.Context { return ctx }
 	ln, err := net.Listen("tcp", s.addr)
 	if err != nil {
+		s.logger.Error("error", zap.Error(err))
 		return err
 	}
 	s.lnAddr = ln.Addr().String()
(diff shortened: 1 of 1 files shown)


pkg/httpd/server.go
```diff
@@ -39,6 +39,7 @@ func (s *Server) Start(ctx context.Context) error {
 	s.srv.BaseContext = func(l net.Listener) context.Context { return ctx }
 	ln, err := net.Listen("tcp", s.addr)
 	if err != nil {
+		s.logger.Error("error", zap.Error(err))
```
alfred-landrum, line 42:
```suggestion
		s.logger.Error("Listen error:", zap.Error(err))
```

mikesbrown:
```
{"level":"error","ts":1593125842.093623,"logger":"httpd","msg":"Listen error","error":"listen tcp :9867: bind: address already in use"}
```

PR #950, opened 2020-06-29
Fix reducer assignment syntax

### LINKED ISSUE #593: ZQL "c=count()" syntax no longer works
Start with the following input data:

```
cat bug.zng
#0:record[a:int]
0:[1;]
0:[2;]
```

As of `zq` commit `1759030`, the following syntax was working:

```
# zq "c=count()" bug.zng
#0:record[c:count]
0:[2;]
```

Then since the following commit `91e062c` ("Support field expressions in reducers (#260)" from @mattnibs) it fails to parse.

```
# zq "c=count()" bug.zng
parse error: 1:10 (9): rule reducerExpr: interface conversion: interface {} is *ast.FieldRead, not string
```

Note that the following alternate syntax has always worked (though to run on current `master` the input data would need to use a different type such as `int64`):

```
# zq "count() as c" bug.zng
#0:record[c:count]
0:[2;]
```

@mccanne cited this regression as unintentional, so I'm opening this as a bug.

FWIW, with similar language topics in the past, we've sometimes debated whether we should have multiple syntaxes to achieve the same thing or standardize on just one. We might want to have that debate as we're addressing this, or just address it so both work as they used and move forward.

henridf:
in: c=count()
out: count() as c

closes #593


alfred-landrum:
There are a few usages I know of in the zar readme, would you please update those as well?

henridf:
@alfred-landrum oops, done.

alfred-landrum:
and one more thing, sorry: would you run the Brim integration tests against this change?  I'm good for it to go in as long as Brim is happy.

philrz:
Another PR https://github.com/brimsec/zq/pull/954 already has syntax in it that depends on this, so since this one is already tested & approved I'm going to merge it.

PR #951, opened 2020-06-29
zar: initial support of s3

mattnibs:
Support s3 data_paths for `zar import` and `zar zq` commands.

In this first pass, the metadata file for an archive is stored locally with
the ability to store data files either in the local filesystem or s3 bucket.

Use pkg iosrc when opening/creating files within zar, allowing zar
directories to be accessed via origins other than the local filesystem.

Also:
- iosrc.URI: A struct for unifying file paths and url paths so that
operations like AppendPath will work for all supported path types.


alfred-landrum:
A few initial thoughts:
- As a phasing strategy, let's first support storing metadata for an archive as a local file, and not in s3. That will prevent AtomicWrite & Rename on the io sources, and we can figure out if/how we want to store & version metadata in s3.
- During import of data, let's add the size of the individual zng files into the metadata info. This will prevent needing to Stat() (or s3 equivalent) the log files, and makes sense for us anyway, since these files will be immutable.

mattnibs:
@alfred-landrum 
> As a phasing strategy, let's first support storing metadata for an archive as a local file, and not in s3. That will prevent AtomicWrite & Rename on the io sources, and we can figure out if/how we want to store & version metadata in s3.

This works be we would need to add a zar import option that would specify placing data files in a different location.

mattnibs:
@alfred-landrum et al, this is ready for review. I still need to add some kind of s3 related zar and zar zqd tests. Working on that.

pkg/iosrc/uri_unix.go
```diff
@@ -0,0 +1,23 @@
+// +build aix darwin dragonfly freebsd js,wasm linux netbsd openbsd solaris
```
nwt, line 1:
```suggestion
// +build !windows
```

nwt: And maybe change the file name to uri_notwindows.go.

mattnibs: Yep, thanks

cmd/zar/main.go
```diff
@@ -12,11 +12,17 @@ import (
...
+	"github.com/brimsec/zq/pkg/s3io"
 )
 
 // Version is set via the Go linker.
 var version = "unknown"
 
+func init() {
+	iosrc.Register("s3", s3io.DefaultSource)
+}
+
```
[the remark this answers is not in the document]
mattnibs: I can do that.

pkg/iosrc/uri.go
```diff
@@ -0,0 +1,87 @@
...
+		return stdin
+	case "stderr":
+		return stderr
+	}
+}
+
+func (p URI) AppendPath(elem ...string) URI {
+	for _, el := range elem {
+		p.Path = p.Path + "/" + el
+	}
```
nwt, line 53:
```suggestion
	p.Path = path.Join(append([]string{p.Path}, elem...))
```

pkg/iosrc/uri.go
```diff
@@ -0,0 +1,87 @@
+package iosrc
+
+import (
+	"errors"
+	"net/url"
+)
+
+type URI struct {
+	url.URL
+}
```
nwt, line 10: Maybe just `type URI url.URL`?

pkg/iosrc/uri.go
```diff
@@ -0,0 +1,87 @@
...
+	stdin  = "stdio:///stdin"
+	stdout = "stdio:///stdout"
+	stderr = "stdio:///stderr"
+)
+
+// ParseURI parses the path using `url.Parse`. If the provided uri does not
+// contain a scheme, the scheme will set to file. Relative paths will be
+// treated as files and resolved as absolute paths using filepath.Abs.
+// path is an empty, Scheme is set to file.
+func ParseURI(path string) (URI, error) {
```
nwt, line 22: Maybe return a *URI? (url.Parse returns a *url.URL.)

pkg/iosrc/uri.go
```diff
@@ -0,0 +1,87 @@
...
+func stdio(path string) string {
+	switch path {
+	default:
+		return path
+	case "stdout":
+		return stdout
+	case "stdin":
+		return stdin
+	case "stderr":
+		return stderr
```
nwt, line 46:
```suggestion
	case "stdin":
		return stdin
	case "stdout":
		return stdout
	case "stderr":
		return stderr
	default:
		return path
```

zdx/zdx.go
```diff
@@ -27,44 +27,19 @@ package zdx
...
-		if err != nil {
-			if os.IsNotExist(err) {
-				err = nil
-			}
-			return err
-		}
-		level++
+		uri.Path = fmt.Sprintf("%s.zng", uri.Path)
+	} else {
+		uri.Path = fmt.Sprintf("%s.%d.zng", uri.Path, level)
```
nwt, line 42:
```suggestion
		uri.Path += ".zng"
	} else {
		uri.Path += fmt.Sprintf(".%d.zng", level)
```

zdx/writer.go
```diff
@@ -34,7 +33,7 @@ import (
 // XXX TBD: the single-file implementation will arrive in a subsequent PR.
 type Writer struct {
 	zctx        *resolver.Context
-	path        string
+	path        iosrc.URI
```
nwt, line 36: Nit: Call this uri.

zdx/reader.go
```diff
@@ -24,21 +25,29 @@ type Reader struct {
...
-	f, err := fs.Open(filename(path, level))
+func newReader(zctx *resolver.Context, uri iosrc.URI, level int) (*Reader, error) {
+	r, err := iosrc.NewReader(filename(uri, level))
 	if err != nil {
 		return nil, err
 	}
-	seeker := zngio.NewSeekerWithSize(f, zctx, FrameSize)
+	rs, ok := r.(io.ReadSeeker)
+	if !ok {
+		return nil, fmt.Errorf("underyling iosrc.NewReader did not return a io.ReadSeeker")
```
nwt, line 42:
```suggestion
		return nil, fmt.Errorf("underyling iosrc.NewReader did not return an io.ReadSeeker")
```

tests/suite/zar/basic.yaml
```diff
@@ -16,11 +15,9 @@ inputs:
...
       ===
-      logs/20200422/1587518620.0622373.zng
-      logs/20200422/1587514053.06464653.zng
-      logs/20200421/1587513216.06789438.zng
+      20200422/1587518620.0622373.zng
+      20200422/1587514053.06464653.zng
+      20200421/1587513216.06789438.zng
```
nwt, line 21: Nit: All these test changes appear to be cosmetic. If that's the case, don't include them in this PR.

pkg/s3io/source.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+var DefaultSource = &Source{}
+var _ iosrc.Source = DefaultSource
+
+type Source struct {
+	Config *aws.Config
+}
+
+func (s *Source) wrapErr(uri iosrc.URI, err error) error {
```
nwt, line 22:
```suggestion
func wrapErr(uri iosrc.URI, err error) error {
```

pkg/iosrc/uri.go
mattnibs: Not a fan. I want this to be immutable like time.Time.

tests/suite/zar/basic.yaml
mattnibs: It is not (I def would not have gone through the trouble I know an other way). The uri change requires that paths get stored as absolute paths not relative paths. Thus left un-changed this would show the absolute path of a tempdir which would make the test fail everytime.

pkg/s3io/source.go
mattnibs: why not

zdx/writer.go
mattnibs: ok

pkg/iosrc/uri.go
mattnibs: path.Join also does a call to path.Clean that was messing things up for me. Don't want that.

cmd/zar/zq/command.go
```diff
@@ -90,12 +78,12 @@ func (c *Command) Run(args []string) error {
...
 	// XXX this is parallelizable except for writing to stdout when
 	// concatenating results
-	return archive.Walk(ark, func(zardir string) error {
+	return archive.Walk(ark, func(zardir iosrc.URI) error {
 		inputs := args
 		var query ast.Proc
 		var err error
 		first := archive.Localize(zardir, inputs[0])
-		if first != "" && fileExists(first) {
+		if ok, _ := iosrc.Exists(first); ok {
```
alfred-landrum, line 86: any reason to ignore the error here?

cmd/zar/zq/command.go
```diff
@@ -149,37 +133,11 @@ func (c *Command) Run(args []string) error {
 	})
 }
 
-func (c *Command) verifyPaths(paths []string) ([]string, error) {
```
alfred-landrum, line 152: Would you explain why you dropped the `verifyPaths` logic?

emitter/file.go
```diff
@@ -19,21 +19,29 @@ func (*noClose) Close() error {
...
 func NewFile(path string, flags *zio.WriterFlags) (*zio.Writer, error) {
-	return NewFileWithSource(path, flags, iosource.DefaultRegistry)
+	if path == "" {
+		path = "stdout"
+	}
+	uri, err := iosrc.ParseURI(path)
+	if err != nil {
+		return nil, err
+	}
+	src, err := iosrc.DefaultRegistry.Source(uri)
```
alfred-landrum, line 29: Can this be `iosrc.GetSource`?

pkg/iosrc/error.go
```diff
@@ -0,0 +1,25 @@
+package iosrc
+
+import (
+	"errors"
+	"fmt"
+)
+
```
alfred-landrum, line 7: Instead of defining a new error type & utilities, could you use the `zqe` package? 

So `NewPathError(uri, err)` would become `zqe.E(zqe.NotFound, uri.String())`?

pkg/iosrc/iosrc.go
```diff
@@ -0,0 +1,134 @@
...
+	"errors"
+	"io"
+	"os"
+	"sync"
+	"time"
+)
+
+const FileScheme = "file"
+
+var DefaultRegistry = &Registry{
```
alfred-landrum, line 15: I suggested to @mattnibs off-PR that in a potential follow-up PR, we could remove the ability to add Sources to registries at runtime, he said he'd investigate.

pkg/iosrc/uri.go
mattnibs: Ok uri would have to be hidden in the struct for it to be totally immutable but I like the fact that I can pass the default return value here into a function and know it will not me mutated.

pkg/iosrc/uri.go
mattnibs: My gut is tell me that I wrote it this way for a particular reason, but now that I'm looking at it I can't see why. I'll do as you suggest.

pkg/iosrc/uri_windows.go
```diff
@@ -0,0 +1,32 @@
...
+	"path/filepath"
+	"regexp"
+)
+
+var winVolumeRe = regexp.MustCompile("^[a-zA-Z]:")
+
+func normalizeFilepaths(path string) (string, error) {
+	if !winVolumeRe.MatchString(path) {
+		if scheme, err := getscheme(path); err != nil || scheme != "" {
+			return path, err
```
alfred-landrum, line 13: This will leave relative paths on windows as relative, is that right & intended?

pkg/iosrc/uri_windows_test.go
```diff
@@ -0,0 +1,16 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestURIWin(t *testing.T) {
+	expected := `c:\test\folder`
+	uri, err := ParseURI(expected)
+	require.NoError(t, err)
+	assert.Equal(t, "file:///c:/test/folder", uri.String())
+	assert.Equal(t, expected, uri.Filepath())
+}
```
alfred-landrum, line 16: Can you add a test that parses `a\b\..\c` and make sure it gives the expected uri?

archive/archive_test.go
```diff
@@ -176,3 +180,32 @@ func TestImportWhileOpen(t *testing.T) {
...
+		LogSizeThreshold: &thresh,
+	}
+	ark, err := CreateOrOpenArchive(root, co, &OpenOptions{DataSource: src})
+	require.NoError(t, err)
+	importTestFile(t, ark, "testdata/td1.zng")
+
+	assert.Equal(t, "s3://test-bucket/test-key/20200422/1587514063.06854538.zng", recvuri.String())
+}
+
+type nopCloser struct{ io.Writer }
```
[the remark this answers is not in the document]
mattnibs: Good point.

pkg/iosrc/uri_windows_test.go
mattnibs: Good idea.

pkg/iosrc/error.go
mattnibs: sure.

Makefile
```diff
@@ -41,6 +41,12 @@ bin/$(ZEEKPATH):
 	@unzip -q bin/$(ZEEKPATH).zip -d bin \
 		&& mv bin/zeek bin/$(ZEEKPATH)
 
+generate:
+	@GOBIN=$(CURDIR)/bin go install github.com/golang/mock/mockgen
+	@PATH=$(CURDIR)/bin:$(PATH) go generate ./...
```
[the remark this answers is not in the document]
mattnibs: In my dist is short for distribution (or some variation) was for binaries that came from this project. `bin` where we also place the `zeek` binary artifact seems like the place to put binary dependencies.

pkg/iosrc/uri.go
nwt: In that case, maybe `p.Path = strings.Join(append([]string{p.Path}, elem...), "/")`. Or maybe not.

cmd/zar/zq/command.go
mattnibs: First off it seemed excessive to do an extra network hop when the call to read the data would give us the same information. Secondly looking at the code below it looks like we are wrapping the created reader in a warningsReader meaning that we'll get the same error message anyways.

cmd/zar/zq/command.go
mattnibs: Actually, no. Will fix.

emitter/file.go
mattnibs: Yep should be.

pkg/iosrc/uri_windows.go
mattnibs: No, this just makes sure that already scheme-ified paths will not get mutated.

tests/suite/zar/basic.yaml
alfred-landrum: to clarify: it's the new `data_path` field in zar.json that now contains an absolute path, not the individual entries for each log.

pkg/s3io/source.go
```diff
@@ -0,0 +1,67 @@
...
+	r, err := NewReader(uri.String(), s.Config)
+	return r, wrapErr(uri, err)
+}
+
+// XXX TODO
+func (s *Source) Remove(uri iosrc.URI) error {
+	return errors.New("method unsupported")
+}
+
+// XXX TODO
```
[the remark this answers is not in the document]
mattnibs: Actually s3io.RemoveAll should be in my next pr (only trying to support zar import/zq right now).

pkg/s3io/source.go
```diff
@@ -0,0 +1,67 @@
...
+	w, err := NewWriter(uri.String(), s.Config)
+	return w, wrapErr(uri, err)
+}
+
+func (s *Source) NewReader(uri iosrc.URI) (io.ReadCloser, error) {
+	r, err := NewReader(uri.String(), s.Config)
+	return r, wrapErr(uri, err)
+}
+
+// XXX TODO
```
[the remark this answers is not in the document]
mattnibs: Actually s3io.RemoveAll should be in my next pr (only trying to support zar import/zq right now).

tests/suite/zar/basic.yaml
mattnibs: @alfred-landrum is correct

PR #956, opened 2020-06-30
force mtime update on metadata file update

alfred-landrum:
After some research and experimentation, it seems that the simplest & most reliable way to ensure the mtime is always updated is to directly set it via `os.Chtimes`. I tried an approach of stat'ing the file before the update, then stat'ing it afterwards, but it seems more efficient to just always do so.

### DIFF
--- a/archive/archive_test.go
+++ b/archive/archive_test.go
@@ -110,7 +110,6 @@ func TestOpenOptions(t *testing.T) {
 }
 
 func TestImportWhileOpen(t *testing.T) {
-	t.Skip("skipping due to test failures, see brim#883")
 	// Create an archive with initial data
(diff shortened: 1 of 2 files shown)


alfred-landrum:
Here's a CI run where I show that the os.Chtimes code works if the condition is hit:
https://github.com/brimsec/zq/runs/880125697
In that that test run (with the code at 8e55d83), the code stats the metadata file, then rewrites the file, then stats again, showing the mtime didn't change. Then the os.Chtimes code is called & forces the mtime to update. No other test failures (except the assert to show that the condition had been hit) were triggered, so the separate archive handle did recognize that the file had been updated & the expected new data was found.

PR #961, opened 2020-06-30
Use runManager in proc.GroupByAggregator

nwt:


henridf:
@nwt could you comment on the benefits of this change? I like the Combiner abstraction for reading from multiple sorted files. I think either way is fine, but I suspect you've done some thinking on this and I'd be curious to know more.

(Btw, I was planning to merge detector.Combiner and proc.Combiner into a single combiner at some point but post #932).

mattnibs:
I think the `runManager` could use the combiner. That said just looking at the line count diff says to me that this is a step forward.

henridf:
> That said just looking at the line count diff says to me that this is a step forward.

Fwiw I think we'd get a similar line count reduction by merging proc.Combiner and detector.Combiner. We're at a fork in the road where both paths lead to LoC reduction 🙂

nwt:
> could you comment on the benefits of this change?

1. Replaces the O(n) list operation on each file read with an O(log n) binary heap operation (where n is the number of files).
2. Reduces the O(n) limit on storage required to merge partial results to O(1) (where n is the number of files).
3. Removes two files, five types, and 234 lines.

> I think the runManager could use the combiner.

It could, but then you're back to the O(n) limit on storage required to merge partial results.

> Fwiw I think we'd get a similar line count reduction by merging proc.Combiner and detector.Combiner.

There's only one way to know for sure! 😄

nwt:
I'm going to see what it looks like if I replace the zbuf.Combiner implementation with a runManager-style heap and then rework runManager to use zbuf.Combiner. I'll open a separate PR if that works out.

henridf:
> I'm going to see what it looks like if I replace the zbuf.Combiner implementation with a runManager-style heap and then rework runManager to use zbuf.Combiner. I'll open a separate PR if that works out.

This is what I had intended to do as noted in the spillable groupby PR (along with unifying the two Combiners as I noted elsewhere). But I'm in no hurry and if you feel like doing that now, go for it!

PR #969, opened 2020-07-03
Support field renaming in cut

### LINKED ISSUE #665: support field rename with cut proc
Add support to the cut proc to rename the fields it allows. So an expression like:
` ... | cut x=a, b ` would have columns x & b.

henridf:
closes #665

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -269,11 +269,11 @@ type (
 	// appropriate value.
 	GroupByProc struct {
 		Node
-		Duration     Duration     `json:"duration"`
-		InputSortDir int          `json:"input_sort_dir,omitempty"`
-		Limit        int          `json:"limit,omitempty"`
-		Keys         []Assignment `json:"keys"`
-		Reducers     []Reducer    `json:"reducers"`
+		Duration     Duration               `json:"duration"`
+		InputSortDir int                    `json:"input_sort_dir,omitempty"`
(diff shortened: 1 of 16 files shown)


ast/ast.go
```diff
@@ -289,15 +289,20 @@ type (
...
 )
 
-type Assignment struct {
+type ExpressionAssignment struct {
 	Target string     `json:"target"`
 	Expr   Expression `json:"expression"`
 }
 
+type FieldAssignment struct {
+	Target string `json:"target"`
```
mattnibs, line 302: Out of curiosity, and I know this change was done a while ago, but why the move away using a `string` here instead of `FieldExpr`?

henridf: It was done here: https://github.com/brimsec/zq/pull/699/

proc/cutter.go
```diff
@@ -33,20 +33,22 @@ func (cb *cutBuilder) cut(in *zng.Record) *zng.Record {
...
 type Cutter struct {
-	zctx        *resolver.Context
-	complement  bool
-	cutBuilders map[int]*cutBuilder
-	fieldnames  []string
-	strict      bool
+	zctx                *resolver.Context
+	complement          bool
+	cutBuilders         map[int]*cutBuilder
+	targets, fieldnames []string
```
nwt, line 39:
```suggestion
	targets []string
	fieldnames []string
```

PR #978, opened 2020-07-06
Publish ES Module ZQL

jameskerr:
In addition to publishing a commonjs version of zql.js for use with Node runtimes, also publish an ES module version. This will make it easy to import into the Deno zealot client, or directly into a modern browser.

This conversion is done using the rollup node module.

### DIFF
--- a/zql/Makefile
+++ b/zql/Makefile
@@ -59,4 +59,4 @@ zql.js: $(PEGJS)
 
 .PHONY: zql.es.js
 zql.es.js: zql.js
-	$(ROLLUP) --format es --plugin commonjs -o zql.es.js zql.js
+	$(ROLLUP) --format es --plugin commonjs -o $@ zql.js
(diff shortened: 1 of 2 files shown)


zql/Makefile
```diff
@@ -55,3 +56,7 @@ zql.go: $(PIGEON) $(GOIMPORTS)
 .PHONY: zql.js
 zql.js: $(PEGJS)
 	cpp -E -P zql.peg | $(PEGJS) $(PEGJS_ARGS) -o $@
+
+.PHONY: zql.es.js
+zql.es.js: zql.js
+	$(ROLLUP) --format es --plugin commonjs -o zql.es.js zql.js
```
nwt, line 62: End with a newline, please.

nwt, line 62:
```suggestion
	$(ROLLUP) --format es --plugin commonjs -o $@ zql.js
```

PR #981, opened 2020-07-08
simplify compilation calls

alfred-landrum:


driver/compile.go
```diff
@@ -14,45 +14,54 @@ import (
...
-func CompileCustom(ctx context.Context, zctx *resolver.Context, custom proc.Compiler, program ast.Proc, reader zbuf.Reader, reverse bool, span nano.Span, logger *zap.Logger) (*MuxOutput, error) {
-	ch := make(chan string, 5)
-	return CompileWarningsChCustom(ctx, zctx, custom, program, reader, "", reverse, span, logger, ch)
-}
-
-func CompileWarningsCh(ctx context.Context, zctx *resolver.Context, program ast.Proc, reader zbuf.Reader, reverse bool, span nano.Span, logger *zap.Logger, ch chan string) (*MuxOutput, error) {
-	return CompileWarningsChCustom(ctx, zctx, nil, program, reader, "", reverse, span, logger, ch)
-}
+func Compile(ctx context.Context, program ast.Proc, reader zbuf.Reader, cfg Config) (*MuxOutput, error) {
+	if cfg.TypeContext == nil {
```
nwt, line 32: Failure to provide the a pointer to reader’s resolver.Context is always a bug, so zctx should remain a required argument.

driver/compile.go
```diff
@@ -14,45 +14,54 @@ import (
 	"go.uber.org/zap"
 )
 
+type Config struct {
```
nwt, line 17: Nit: Sort the field names unless the current order is significant.

driver/compile.go
```diff
@@ -14,45 +14,54 @@ import (
...
+	TypeContext *resolver.Context
+	SortKey     string
```
nwt, line 19: Nit: It’s unclear that this is a property of the reader, so I think you should document that or call it ReaderSortKey.

driver/compile.go
```diff
@@ -14,45 +14,50 @@ import (
...
+func Compile(ctx context.Context, program ast.Proc, zctx *resolver.Context, reader zbuf.Reader, cfg Config) (*MuxOutput, error) {
+	if cfg.Span.Dur == 0 {
+		cfg.Span = nano.MaxSpan
+	}
+	if cfg.Warnings == nil {
+		cfg.Warnings = make(chan string, 5)
+	}
+	if cfg.Logger == nil {
+		cfg.Logger = zap.NewNop()
+	}
```
nwt, line 39:
```suggestion
	if cfg.Logger == nil {
		cfg.Logger = zap.NewNop()
	}
	if cfg.Span.Dur == 0 {
		cfg.Span = nano.MaxSpan
	}
	if cfg.Warnings == nil {
		cfg.Warnings = make(chan string, 5)
	}
```
Nit: Follow the field declaration order.

Issue #973, opened 2020-07-06
zar s3 system tests


mattnibs:
Use minio and ztest to write a series of system tests for all the newly s3-ified zar commands.

henridf:
Are we ok with adding minio to ztest now? I think it makes sense, but there had been opposition in the past.

alfred-landrum:
I'm ok with it: we're using more ztest style tests, and we're increasingly going to focus on S3 backed archives, so it seems like a useful fit going forward. @mattnibs had brought it'd be useful if the test framework could make it easier to launch & teardown the minio service; I'd suggested we should wait until we have a few tests in place to see how that could work.

Issue #972, opened 2020-07-06
s3 support for zar find/index/ls/rm/rmdirs/stat


mattnibs:

philrz:
Verified in `zq` commit `59b4bcc` with some bugs found/opened.

Here's the "happy path" of [`zar` README](https://github.com/brimsec/zq/blob/master/cmd/zar/README.md) steps performed with the storage landing on an S3 bucket.

```
$ export ZAR_ROOT="$(pwd)"

$ zq ~/work/zq-sample-data/zng/*.gz | zar import -data s3://zq-972 -s 25MB -

$ zar index :ip
s3://zq-972/20180324/1521912990.158766.zng: creating index s3://zq-972/20180324/1521912990.158766.zng.zar/zdx-type-ip
s3://zq-972/20180324/1521912792.328806.zng: creating index s3://zq-972/20180324/1521912792.328806.zng.zar/zdx-type-ip
s3://zq-972/20180324/1521912549.366398.zng: creating index s3://zq-972/20180324/1521912549.366398.zng.zar/zdx-type-ip
s3://zq-972/20180324/1521912335.72784.zng: creating index s3://zq-972/20180324/1521912335.72784.zng.zar/zdx-type-ip
s3://zq-972/20180324/1521912152.518493.zng: creating index s3://zq-972/20180324/1521912152.518493.zng.zar/zdx-type-ip
s3://zq-972/20180324/1521911975.777469.zng: creating index s3://zq-972/20180324/1521911975.777469.zng.zar/zdx-type-ip
s3://zq-972/20180324/1521911841.543641.zng: creating index s3://zq-972/20180324/1521911841.543641.zng.zar/zdx-type-ip

$ zar index uri
s3://zq-972/20180324/1521912990.158766.zng: creating index s3://zq-972/20180324/1521912990.158766.zng.zar/zdx-field-uri
s3://zq-972/20180324/1521912792.328806.zng: creating index s3://zq-972/20180324/1521912792.328806.zng.zar/zdx-field-uri
s3://zq-972/20180324/1521912549.366398.zng: creating index s3://zq-972/20180324/1521912549.366398.zng.zar/zdx-field-uri
s3://zq-972/20180324/1521912335.72784.zng: creating index s3://zq-972/20180324/1521912335.72784.zng.zar/zdx-field-uri
s3://zq-972/20180324/1521912152.518493.zng: creating index s3://zq-972/20180324/1521912152.518493.zng.zar/zdx-field-uri
[...]

PR #986, opened 2020-07-08
zar ls: add s3 support

mattnibs:
Add support for s3 data paths for zar ls command.

Use minio to create simple zar import/ls system test for s3
datapaths.

Supporting changes:
- Rename /s/ZTEST_BINDIR/ZTEST_PATH and support ":" separated path
lists for this environment variables. This change was needed in order
have the minio command available for s3 tests.

Part of #972 & #973


pkg/s3io/s3io.go
```diff
@@ -119,14 +118,35 @@ func Stat(path string, cfg *aws.Config) (*s3.HeadObjectOutput, error) {
...
-	sess := session.Must(session.NewSession(cfg))
-	client := s3.New(sess)
-	return client.HeadObject(&s3.HeadObjectInput{
+	return newClient(cfg).HeadObject(&s3.HeadObjectInput{
 		Bucket: aws.String(bucket),
 		Key:    aws.String(key),
 	})
 }
 
+func Ls(path string, cfg *aws.Config) ([]string, error) {
```
alfred-landrum, line 127:
```suggestion
func ListObjects(path string, cfg *aws.Config) ([]string, error) {
```

tests/suite/zar/s3/import.yaml
```diff
@@ -0,0 +1,51 @@
...
+  mkdir -p data/bucket
+  minio server --quiet --address "localhost:9000" ./data &> /dev/null &
+  pid=$!
+  until nc -vzw 2 localhost 9000; do sleep 2; done
+  zar import -s 2500B -R ./root -data s3://bucket/zartest babble.tzng
+  echo ===
+  zar ls -R ./root
+  echo ===
+  zar ls -relative -R ./root
+  kill $pid &> /dev/null
```
alfred-landrum, line 13:
```suggestion
  kill -KILL $pid &> /dev/null
```

tests/suite/zar/s3/import.yaml
```diff
@@ -0,0 +1,51 @@
+script: |
+  export AWS_REGION=us-east-2 AWS_ACCESS_KEY_ID=minioadmin \
+        AWS_SECRET_ACCESS_KEY=minioadmin AWS_S3_ENDPOINT=http://localhost:9000
+  mkdir -p data/bucket
+  minio server --quiet --address "localhost:9000" ./data &> /dev/null &
+  pid=$!
+  until nc -vzw 2 localhost 9000; do sleep 2; done
+  zar import -s 2500B -R ./root -data s3://bucket/zartest babble.tzng
```
alfred-landrum, line 8: Consider using `-s 20KiB`, as it will reduce the number of zng logs to just 2.

pkg/s3io/s3io.go
mattnibs: good idea

tests/suite/zar/s3/import.yaml
mattnibs: good idea.

tests/suite/zar/s3/import.yaml
```diff
@@ -0,0 +1,27 @@
+script: |
+  export AWS_REGION=us-east-2 AWS_ACCESS_KEY_ID=minioadmin \
+        AWS_SECRET_ACCESS_KEY=minioadmin AWS_S3_ENDPOINT=http://localhost:9000
+  mkdir -p data/bucket
+  minio server --quiet --address "localhost:9000" ./data &> /dev/null &
+  pid=$!
+  until nc -vzw 2 localhost 9000; do sleep 2; done
```
nwt, line 7: If minio fails to start, this test is going to fail without a clear indication of the problem. To improve on that, I think you'll need to stop redirecting the server's output to /dev/null and replace the until command with one that won't loop forever.

tests/suite/zar/s3/import.yaml
```diff
@@ -0,0 +1,27 @@
...
+  zar import -s 20KiB -R ./root -data s3://bucket/zartest babble.tzng
+  echo ===
+  zar ls -R ./root
+  echo ===
+  zar ls -relative -R ./root
+  kill -KILL $pid &> /dev/null
```
nwt, line 13: `trap "kill -9 $!" EXIT` immediately after the background command instead?

ztest/ztest.go
```diff
@@ -146,13 +147,13 @@ import (
 // subtest named f.  bindir is a path to the executables that the script-mode
 // tests will run.
 func Run(t *testing.T, dirname string) {
-	bindir := os.Getenv("ZTEST_BINDIR")
+	bindir := os.Getenv("ZTEST_PATH")
```
nwt, line 150: You should rename all the bindir variables in this file, too.

Makefile
```diff
@@ -105,4 +108,4 @@ clean: clean-python
 
 .PHONY: fmt tidy vet test-unit test-system test-heavy sampledata test-ci
 .PHONY: perf-compare build install create-release-assets clean clean-python
-.PHONY: build-python-wheel generate test-generate
+.PHONY: build-python-wheel generate test-generate bin/minio
```
nwt, line 111: Do you really want to declare this to be phony?

tests/suite/zar/s3/import.yaml
mattnibs: Sure

tests/suite/zar/s3/import.yaml
mattnibs: This would be nice I'll try this out.

Makefile
mattnibs: Will this update if go.mod changes to a newer version of bin/minio?

tests/suite/zar/s3/import.yaml
mattnibs: I'm not quite clear on what the "stop redirecting the server's output to /dev/null" will help with.

mattnibs: Whoops, nevermind

mattnibs: @nwt re-did this... PTAL!

tests/suite/zar/s3/minio.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+mkdir -p data/bucket
+minio server --writeportfile="./port" --quiet --address ":0" ./data &
```
alfred-landrum, line 4: I know its not necessary for the ztest usage (since it runs tests in a temp dir), but would you use `mktemp -d` or similar technique to create a unique port file? That would make this script useful if I wanted to run zar on a fake S3 during development.

tests/suite/zar/s3/minio.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
alfred-landrum, line 1: 🎉

tests/suite/zar/s3/minio.sh
alfred-landrum, line 4: Is it possible to use 'localhost:0' as the minio address, to avoid system popup warnings about listening to network ports?

mattnibs: sure

mattnibs: Sure @alfred-landrum. Do I need to worry about delete this tempdir when the test is done?

mattnibs, line 4:
```suggestion
minio server --writeportfile="./port" --quiet --address "localhost:0" ./data &
```

alfred-landrum: yes, should just be another exit trap

ztest/ztest.go
```diff
@@ -143,16 +144,16 @@ import (
 
 // Run runs the ztests in the directory named dirname.  For each file f.yaml in
 // the directory, Run calls FromYAMLFile to load a ztest and then runs it in
-// subtest named f.  bindir is a path to the executables that the script-mode
-// tests will run.
+// subtest named f. path is an os-specific delimited list of paths like the
```
nwt, line 147:
```suggestion
// subtest named f.  path is a command search path like the
```

ztest/ztest.go
```diff
@@ -143,16 +144,16 @@ import (
...
+// $PATH environment variable.
```
nwt, line 148:
```suggestion
// PATH environment variable.
```

ztest/ztest.go
```diff
@@ -546,14 +547,17 @@ func runsh(testname, bindir, dirname string, zt *ZTest) error {
 // runzq runs the query in ZQL over inputs and returns the output formatted
 // according to outputFormat. inputs may be in any format recognized by "zq -i
 // auto" and maybe be gzip-compressed.  outputFormat may be any string accepted
-// by "zq -f".  If bindir is empty, the query runs in the current process.
-// If bindir is not empty, it specifies a zq path that will be used to run
+// by "zq -f".  If path is empty, the query runs in the current process.
+// If path is not empty, it specifies a zq path that will be used to run
 // the query.
```
nwt, line 552:
```suggestion
// If path is not empty, it specifies a command search path used to find
// a zq executable to run the query.
```

tests/suite/zar/s3/minio.sh
```diff
@@ -0,0 +1,25 @@
...
+
+mkdir -p data/bucket
+portdir=$(mktemp -d)
+trap "rm -rf $portdir" EXIT
+minio server --writeportfile="$portdir/port" --quiet --address "localhost:0" ./data &
+trap "kill -9 $!" EXIT
```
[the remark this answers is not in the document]
mattnibs: jeez, thanks

go.mod
```diff
@@ -27,10 +27,11 @@ require (
...
+	golang.org/x/crypto v0.0.0-20200709230013-948cd5f35899
 	golang.org/x/sync v0.0.0-20190911185100-cd5d95a43a6e
-	golang.org/x/sys v0.0.0-20200513112337-417ce2331b5c // indirect
-	golang.org/x/text v0.3.2
+	golang.org/x/text v0.3.3
 	gopkg.in/natefinch/lumberjack.v2 v2.0.0
 	gopkg.in/yaml.v3 v3.0.0-20200121175148-a6ecf24a6d71
 )
+
+replace github.com/minio/minio => github.com/mattnibs/minio v0.0.0-20200716173056-375b93385c07
```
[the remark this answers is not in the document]
mattnibs: 1. Sure
2. The package uses canonical imports so builds fail if I require directly from github.com/mattnibs/minio https://golang.org/doc/go1.4#canonicalimports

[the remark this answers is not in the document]
mattnibs: Could be fixed by changing import name definition in minio/main.go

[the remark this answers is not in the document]
nwt: Makes sense. I think the replace directive is fine.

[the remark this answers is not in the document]
alfred-landrum: let's not do that (import name def change), to minimize the minio changes.

PR #993, opened 2020-07-09
Add zar zq s3 system test

mattnibs:
Also fixes a bug where files were objects were not getting opened
from package s3io (and thus not responding to the AWS_S3_ENDPOINT
env variable).

Part of #973

### DIFF
--- a/zio/detector/file.go
+++ b/zio/detector/file.go
@@ -8,9 +8,6 @@ import (
 	"os"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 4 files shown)


zio/detector/file.go
```diff
@@ -76,41 +77,6 @@ func (pw *pipeWriterAt) WriteAt(p []byte, _ int64) (n int, err error) {
 	return pw.Write(p)
 }
```
nwt, line 78: Delete pipeWriterAt too, please.

mattnibs: 👍

PR #998, opened 2020-07-10
Rename proc

### LINKED ISSUE #856: "rename" processor
https://github.com/brimsec/zq/issues/665 first proposed the idea of renaming fields as a side-effect of using the `cut` processor. In a [comment](https://github.com/brimsec/zq/issues/665#issuecomment-637668926), @henridf made the valid point that as a first step we should have an explicit `rename` processor, then after that exists we could start to offer renaming as a side-effect in other processors. In a group discussion, we agreed that this makes sense. Among other things, a user seeking to rename fields might look at the [ZQL processors doc](https://github.com/brimsec/zq/blob/master/zql/docs/processors/README.md) and see the word "rename" prominently in the table of contents and be off to the races.

The proposed syntax is a simple `... | rename source_ip=id.orig_h,dest_ip=id.resp_h` that would take a list of several `to=from` pairings. Data types (including aliases) would be preserved. We also agreed that this processor would be kept super simple with no expressions permitted, e.g. `rename bar=String.toLower(foo)` would _not_ be available. That's what `put` is for. When we're updating the docs, we should mention this so that way if users were expecting to do more than simple renaming, they know where to go.

henridf:
closes #856


proc/rename.go
```diff
@@ -0,0 +1,116 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/brimsec/zq/ast"
+	"github.com/brimsec/zq/zbuf"
+	"github.com/brimsec/zq/zng"
+)
+
+// Rename renames one or more fields in a record. A field can only be
+// renamed within it's own record. For example id.orig_h can be
```
henridf, line 13:
```suggestion
// renamed within its own record. For example id.orig_h can be
```

henridf:
@alfred-landrum doc added

zql/docs/processors/README.md
```diff
@@ -231,6 +232,38 @@ ID.ORIG_H     ID.ORIG_P ID.RESP_H       ID.RESP_P ORIG_BYTES RESP_BYTES TOTAL_BY
...
+
+## `rename`
+
+|                           |                                                 |
+| ------------------------- | ----------------------------------------------- |
+| **Description**           | Rename fields in a record. |
+| **Syntax**                | `rename <newname> = <oldname> [, <newname> = <oldname> ...]`     |
+| **Required arguments**    | One or more field assignment expressions. Renames are applied left to right; each rename observes the effect of all renames that preceded it. |
+| **Optional arguments**    | None |
+| **Limitations**           | A field can only be renamed within its own record. For example id.orig_h can be renamed to id.src, but it cannot be renamed to src. |
```
philrz, line 245:
```suggestion
| **Limitations**           | A field can only be renamed within its own record. For example `id.orig_h` can be renamed to `id.src`, but it cannot be renamed to `src`. |
```

I've been trying to stick to a convention in the docs of using fixed width when referencing field names.

zql/docs/processors/README.md
```diff
@@ -231,6 +232,38 @@ ID.ORIG_H     ID.ORIG_P ID.RESP_H       ID.RESP_P ORIG_BYTES RESP_BYTES TOTAL_BY
...
+conn  1521911721.255387 C8Tful1TvM3Zf5x8fl 10.164.94.120  39681            10.47.3.155      3389             tcp   -        0.004266 97         19         RSTR       -          -          0            ShADTdtr    10        730           6         342           -
+conn  1521911721.411148 CXWfTK3LRdiuQxBbM6 10.47.25.80    50817            10.128.0.218     23189            tcp   -        0.000486 0          0          REJ        -          -          0            Sr          2         104           2         80            -
+conn  1521911721.926018 CM59GGQhNEoKONb5i  10.47.25.80    50817            10.128.0.218     23189            tcp   -        0.000538 0          0          REJ        -          -          0            Sr          2         104           2         80            -
+conn  1521911722.690601 CuKFds250kxFgkhh8f 10.47.25.80    50813            10.128.0.218     27765            tcp   -        0.000546 0          0          REJ        -          -          0            Sr          2         104           2         80            -
+```
```
philrz, line 264:
```suggestion
...
```
```

I've been trying to include the "..." when we're showing just an excerpt of output.

Comment (henridf) [reply zql/docs/processors/README.md]: oops should have read this one more closely before applying 🙂

File: tests/suite/rename/rename-error-dupe.yaml [hunk 4]
```diff
@@ -0,0 +1,12 @@
+script: zq -t "rename foo=goo" -
```
nwt, line 1: This should be a ZQL test, not a script test.

tests/suite/rename/rename-error-move.yaml
```diff
@@ -0,0 +1,11 @@
+script: zq -t "rename dst=id.resp_h" -
```
nwt, line 1: This should be a ZQL test, not a script test.

proc/rename.go
```diff
@@ -0,0 +1,113 @@
...
+		typeMap:    make(map[int]*zng.TypeRecord),
+	}, nil
+}
+
+func (r *Rename) renamedType(typ *zng.TypeRecord, fields []string, target string) (*zng.TypeRecord, error) {
+	c, ok := typ.ColumnOfField(fields[0])
+	if !ok {
+		return typ, nil
+	}
+	var innerTyp zng.Type
```
nwt, line 53:
```suggestion
	var innerType zng.Type
```

tests/suite/rename/rename-error-dupe.yaml
henridf: I made these script tests because they look at stderr, and a zql test fails due to exit status 1. Is there an example of a zql test that tests errors that I'm missing?

nwt: There are.

```
$ git grep -l errorRE -- *.yaml
tests/suite/errors/nosemi.yaml
tests/suite/put/put-6.yaml
tests/suite/reducer/duplicate.yaml
```

zql/parser-support.go
```diff
@@ -281,6 +281,15 @@ func makePutProc(first, rest interface{}) *ast.PutProc {
 	return &ast.PutProc{ast.Node{"PutProc"}, clauses}
 }
 
+func makeRenameProc(first, rest interface{}) (*ast.RenameProc, error) {
+	fields := []ast.FieldAssignment{first.(ast.FieldAssignment)}
+	for _, c := range rest.([]interface{}) {
+		fields = append(fields, c.(ast.FieldAssignment))
+	}
+
```
nwt, line 289:
```suggestion
```

proc/rename.go
```diff
@@ -0,0 +1,113 @@
...
+				r.typeMap[id] = typ
+			}
+			in.Type = r.typeMap[id]
+			recs = append(recs, in)
+		}
+		batch.Unref()
+		if len(recs) > 0 {
+			return zbuf.NewArray(recs), nil
+		}
+	}
```
nwt, line 112: While it makes sense to check for zero-length batches in procs that remove records, there's no need to do so here.
```suggestion
		batch, err := r.Get()
		if EOS(batch, err) {
			return nil, err
		}
		recs := make([]*zng.Record, 0, batch.Length())
		for k := 0; k < batch.Length(); k++ {
			in := batch.Index(k)
			id := in.Type.ID()
			if _, ok := r.typeMap[id]; !ok {
				typ, err := r.computeType(in.Type)
				if err != nil {
					return nil, fmt.Errorf("rename: %w", err)
				}
				r.typeMap[id] = typ
			}
			in.Type = r.typeMap[id]
			recs = append(recs, in)
		}
		batch.Unref()
		return zbuf.NewArray(recs), nil
```

proc/rename.go
```diff
@@ -0,0 +1,113 @@
...
+// Rename renames one or more fields in a record. A field can only be
+// renamed within its own record. For example id.orig_h can be
+// renamed to id.src, but it cannot be renamed to src. Renames are
+// applied left to right; each rename observes the effect of all
+// renames that preceded it.
+type Rename struct {
+	Base
+	fieldnames []string
+	targets    []string
+	typeMap    map[int]*zng.TypeRecord
```
nwt, line 21: Is there a real need for type IDs here? If not, just make this `map[zng.Type]*zng.TypeRecord`.

tests/suite/rename/rename-error-dupe.yaml
henridf: thanks!

proc/rename.go
henridf: Could you explain the rationale please?

nwt: Using the types themselves is, I think, clearer than using their IDs. Specifically, the role played here by a `map[zng.Type]*zng.TypeRecord` is more obvious than that played by a `map[int]*zng.TypeRecord`.

proc/rename.go
```diff
@@ -0,0 +1,109 @@
...
+		in := batch.Index(k)
+		id := in.Type.ID()
+		if _, ok := r.typeMap[id]; !ok {
+			typ, err := r.computeType(in.Type)
+			if err != nil {
+				return nil, fmt.Errorf("rename: %w", err)
+			}
+			r.typeMap[id] = typ
+		}
+		in.Type = r.typeMap[id]
```
nwt, line 104: `in` is a `*zng.Record`, so this write will be visible to other goroutines that receive the same zbuf.Batch. We should make a copy before changing the type (and use `in` without modification if we don't need to change the type).

proc/rename.go
alfred-landrum: We have many places where there's a `map[int]something` where the key is the return of `zng.Type.ID()`. Would you suggest we have `Type.ID()` return something other than `int` & use that?

proc/rename.go
henridf: good catch!

proc/rename.go
henridf: I see. Seems reasonable though I think it should be `map[zng.TypeRecord]*zng.TypeRecord`. 

Though if this is all about type descriptiveness, then an alternative would be to declare `type TypeID int` in `zng/type.go` (as we have done elsewhere) and switch to that type in place of ints, including using a  `map[zng.TypeID]*zng.TypeRecord` right here.

nwt: > We have many places where there's a map[int]something where the key is the return of zng.Type.ID(). Would you suggest we have Type.ID() return something other than int & use that?

@alfred-landrum: No, I'd instead suggest that we use `map[*zng.RecordType]something` or `map[zng.Type]something` instead of `map[int]something` whenever practical because I think they're clearer. (They're also marginally faster.)

tests/suite/rename/rename-error-dupe.yaml
henridf: now i know why i had the recollection that zql tests didn't fully work for error checks: https://github.com/brimsec/zq/pull/1001

proc/rename.go
henridf: 1f72da07d1983cd8a41085680bfec627f801096e

(I started off using `record.Keep()`, but that was a no-op since records are non-volatile. Maybe split should mark all records as volatile).

proc/rename.go
nwt: > Though if this is all about type descriptiveness, then an alternative would be to declare `type TypeID int` in `zng/type.go` (as we have done elsewhere) and switch to that type in place of ints, including using a `map[zng.TypeID]*zng.TypeRecord` right here.

@henridf: After some reflection, the ultimate issue for me here is my sense that type IDs shouldn't be part of the zng.Type interface. That's clearly a topic for another venue, and as @alfred-landrum pointed out, there's plenty of precedent for what you've done here, so I think you should leave it as is.

Sorry it took me so long to figure that out!

proc/rename.go
```diff
@@ -0,0 +1,111 @@
...
+		if _, ok := r.typeMap[id]; !ok {
+			typ, err := r.computeType(in.Type)
+			if err != nil {
+				return nil, fmt.Errorf("rename: %w", err)
+			}
+			r.typeMap[id] = typ
+		}
+		if id != r.typeMap[id].ID() {
+			in = zng.NewRecord(r.typeMap[id], in.Raw)
+		}
```
nwt, line 106: @henridf: I'm sorry I didn't realize this when I made my previous comment here, but it now occurs to me that the batch.Unref call below means we can't reuse records unless we know they're nonvolatile. Since we don't have a zng.Record method for checking that, we could do this.
```suggestion
		in := batch.Index(k).Keep()
		id := in.Type.ID()
		if _, ok := r.typeMap[id]; !ok {
			typ, err := r.computeType(in.Type)
			if err != nil {
				return nil, fmt.Errorf("rename: %w", err)
			}
			r.typeMap[id] = typ
		}
		in.Type = r.typeMap[id]
```

Alternatively, you could add zng.(*Record).IsNonvolatile and then replace lines 104 and 105 with this.
```go
		if id != r.typeMap[id].ID() || !in.IsNonvolatile() {
			in = in.Keep()
			in.Type = r.typeMap[id]
```

henridf: @nwt  this doesn't work, specifically it fails the test I added in https://github.com/brimsec/zq/commit/1f72da07d1983cd8a41085680bfec627f801096e. The reason for that is noted in https://github.com/brimsec/zq/pull/998#discussion_r454375765 : split does not mark records as nonvolatile.

nwt: @henridf: Doh! I forgot that zng.(*Record).Keep returns its receiver if the receiver is nonvolatile. This is a little clumsy but I think it's finally correct. Tests do pass with it.
```suggestion
		in := batch.Index(k)
		id := in.Type.ID()
		if _, ok := r.typeMap[id]; !ok {
			typ, err := r.computeType(in.Type)
			if err != nil {
				return nil, fmt.Errorf("rename: %w", err)
			}
			r.typeMap[id] = typ
		}
		out := in.Keep()
		if id != r.typeMap[id].ID() {
			if out != in {
				out.Type = r.typeMap[id]
			} else {
				out = zng.NewRecord(r.typeMap[id], out.Raw)
			}
		}
		recs = append(recs, out)
```

PR #1001, opened 2020-07-14
ztest: return compilation error as string

henridf:
Previously, compilation errors weren't returned as the "warnOrError"
return argument of ztest.runzq(), and so a zql-style test couldn't
check for those. This change allows zql-style ztests to check for
compilation errors.


henridf:
Are we ok with merging things with the broken link check?

alfred-landrum:
@henridf : yes, it's ok to ignore the link check failure

Issue #989, opened 2020-07-09
record type and field indexes in archive metadata


alfred-landrum:
When a type or field index (ie, a non-custom index) is created, we should record its existence in the archive's metadata file, and when a user issues a non-custom index search, we should use the entries in the metadata file to choose what index to use for lookup.

philrz:
Verified in `zq` commit `314d783`. After adding the type & field indexes first shown in the [zar README](https://github.com/brimsec/zq/blob/master/cmd/zar/README.md), here's what the additional info in the archive metadata looks like:

```
...
$ cat zar.json | jq .
...
  "indexes": {
    "zdx-field-uri": {
      "type": "field",
      "path": "zdx-field-uri"
    },
    "zdx-type-ip": {
      "type": "type",
      "path": "zdx-type-ip"
    }
  }
...
```

Thanks @henridf!

Issue #990, opened 2020-07-09
archive stats should include index information


alfred-landrum:
Follow on from https://github.com/brimsec/zq/issues/989 : The archive stats information should include information about the non-custom indexes (type or field indexes). There should be a record for each index file, and it should include a field that has the log id the index covers.

philrz:
Verified in `zq` commit `cdb8ca6`.

After creating the type, field, and custom indexes following the steps in the [`zar` README](https://github.com/brimsec/zq/blob/master/cmd/zar/README.md), now a `zar stat` includes the additional "index" lines.

```
$ zar stat
TYPE  LOG_ID                         START             DURATION      SIZE
chunk 20180324/1521912990.158766.zng 1521912792.331503 197.827263001 25483642
TYPE  LOG_ID                         INDEX_ID INDEX_TYPE SIZE KEYS.ID
index 20180324/1521912990.158766.zng custom   zql        4731 record[orig_h:ip]
TYPE  LOG_ID                         INDEX_ID      INDEX_TYPE SIZE  KEYS.KEY
index 20180324/1521912990.158766.zng zdx-field-uri field      29842 bstring
index 20180324/1521912990.158766.zng zdx-type-ip   type       22936 ip
TYPE  LOG_ID                         START             DURATION      SIZE
chunk 20180324/1521912792.328806.zng 1521912549.366782 242.962024001 25478311
TYPE  LOG_ID                         INDEX_ID INDEX_TYPE SIZE KEYS.ID
index 20180324/1521912792.328806.zng custom   zql        5947 record[orig_h:ip]
TYPE  LOG_ID                         INDEX_ID      INDEX_TYPE SIZE  KEYS.KEY
index 20180324/1521912792.328806.zng zdx-field-uri field      52462 bstring
index 20180324/1521912792.328806.zng zdx-type-ip   type       10827 ip
TYPE  LOG_ID                         START             DURATION      SIZE
chunk 20180324/1521912549.366398.zng 1521912335.728195 213.638203001 25499352
TYPE  LOG_ID                         INDEX_ID INDEX_TYPE SIZE  KEYS.ID
index 20180324/1521912549.366398.zng custom   zql        10319 record[orig_h:ip]
TYPE  LOG_ID                         INDEX_ID      INDEX_TYPE SIZE  KEYS.KEY
index 20180324/1521912549.366398.zng zdx-field-uri field      41855 bstring
index 20180324/1521912549.366398.zng zdx-type-ip   type       79054 ip
[...]

PR #1003, opened 2020-07-15
Index metadata

henridf:
This fulfills the requirements of #989 (I think) in a very simple way, but putting this up as a draft to see if this is what @alfred-landrum had in mind. 

In particular, the the spec says "use the entries in the metadata file to choose what index to use for lookup." which could probably be interpreted in other ways. Very open to revising this if needed!

### DIFF
--- a/archive/finder.go
+++ b/archive/finder.go
@@ -10,6 +10,7 @@ import (
 	"github.com/brimsec/zq/zdx"
 	"github.com/brimsec/zq/zng"
 	"github.com/brimsec/zq/zng/resolver"
+	"github.com/brimsec/zq/zqe"
 )
 
 type findOptions struct {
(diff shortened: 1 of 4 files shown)


alfred-landrum:
thanks @henridf - this is what I had in mind!

henridf:
I haven't added a test so far because this doesn't change externally-visible behavior. I'm reluctant to add a ztest script test that checks the contents of `zar.json`...

alfred-landrum:
> I haven't added a test so far because this doesn't change externally-visible behavior. I'm reluctant to add a ztest script test that checks the contents of `zar.json`...

When I work on https://github.com/brimsec/zq/issues/990 , there will be tests that look for index stats, that would only succeed if the indexes are known, so should be covered then.

PR #1005, opened 2020-07-15
zql examples tests: inspect all markdown files

mikesbrown:
Inspect all markdown files in the tree for zql examples instead of requiring maintainers to maintain a list.

Testing:
- the same number of markdown example tests ran
- I created a dummy markdown file with a failing example, and the test failed

### DIFF
--- a/tests/zq_example.go
+++ b/tests/zq_example.go
@@ -80,6 +80,7 @@ import (
 	"bytes"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


mikesbrown:
Just realized I want to skim the documentation not shown in the diff to make sure it's still relevant. Expect an update patch in a second.

mikesbrown:
> Just realized I want to skim the documentation not shown in the diff to make sure it's still relevant. Expect an update patch in a second.

Done

PR #1011, opened 2020-07-17
Make s3io.Reader an io.Seeker

mattnibs:
This functionality is needed for the zar find command to work.

Add system test for zar index and zar find commands.

Part of #972 & #973

### DIFF
--- a/pkg/s3io/reader.go
+++ b/pkg/s3io/reader.go
@@ -58,6 +58,16 @@ func (r *Reader) bytesRange(num int) string {
 	return fmt.Sprintf("bytes=%d-%d", r.offset, r.offset+int64(num)-1)
 }
 
+type writeAtBuf []byte
+
+func (w writeAtBuf) WriteAt(p []byte, off int64) (int, error) {
+	n := copy(w[off:], p)
+	if n < len(p) {
+		return n, errors.New("s3io: short write")
+	}
+	return n, nil
+}
+
 func (r *Reader) Read(p []byte) (int, error) {
 	if r.offset >= r.size {
 		return 0, io.EOF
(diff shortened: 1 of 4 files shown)


pkg/s3io/reader.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	wab := aws.NewWriteAtBuffer(p)
+	bytesDownloaded, err := r.downloader.Download(wab, getObj)
+	if err != nil {
+		return 0, err
+	}
+
+	buf := wab.Bytes()
+	if len(buf) > n {
```
nwt, line 85: Nit: I think `p == wab.Bytes()` would be a clearer condition.

pkg/s3io/reader.go
```diff
@@ -0,0 +1,97 @@
...
+	case io.SeekStart:
+	case io.SeekCurrent:
+		offset += r.offset
+	case io.SeekEnd:
+		offset += r.size
+	default:
+		return 0, fmt.Errorf("%d: unknown whence value", whence)
+	}
+	if offset < 0 {
+		return 0, errInvalidOffset
```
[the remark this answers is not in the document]
mattnibs: Awesome

pkg/s3io/reader.go
mattnibs: You can only compare []byte to `nil`.

pkg/s3io/reader.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		return 0, err
+	}
+
+	buf := wab.Bytes()
+	if len(buf) > len(p) {
+		// backing buffer reassigned, copy over some of the data
+		copy(p, buf)
+		bytesDownloaded = int64(len(p))
+	}
```
nwt, line 84: Nit: The io.WriterAt documentation says WriteAt should return an error on a short write, but using aws.WriteAtBuffer like this effectively ignores a short write. Maybe use this instead.
```go
type writeAtBuf []byte

func (w writeAtBuf) WriteAt(p []byte, off int64) (int, error) {
	n := copy(w[off:], p)
	if n < len(p) {
		return n, errors.New("s3io: short write")
	}
	return n, nil
}
```

mattnibs: #attentionToDetail

PR #1012, opened 2020-07-17
zar s3: support rm and rmdirs commands

mattnibs:
Part of #972 & #973

### DIFF
--- a/pkg/s3io/s3io.go
+++ b/pkg/s3io/s3io.go
@@ -97,6 +97,41 @@ func (w *Writer) Close() error {
 	return w.err
 }
 
+func RemoveAll(path string, cfg *aws.Config) error {
+	bucket, key, err := parsePath(path)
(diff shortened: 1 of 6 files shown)


pkg/s3io/s3io.go
```diff
@@ -112,6 +112,41 @@ func NewReader(path string, cfg *aws.Config) (io.ReadCloser, error) {
...
 }
 
+func RemoveAll(path string, cfg *aws.Config) error {
+	bucket, key, err := parsePath(path)
+	if err != nil {
+		return err
+	}
+	client := newClient(cfg)
+	deleter := s3manager.NewBatchDeleteWithClient(client)
+	itr := s3manager.NewDeleteListIterator(client, &s3.ListObjectsInput{
```
nwt, line 122: Nit: We use it for iterators everywhere else.

PR #1015, opened 2020-07-18
add index information to zar stat

### LINKED ISSUE #990: archive stats should include index information

alfred-landrum:
Adds index information to the results of the `zar stat` command and the `/space/{spaceid}/archivestat` api.
 
Closes #990

### DIFF
--- a/archive/stat.go
+++ b/archive/stat.go
@@ -14,7 +14,7 @@ import (
 	"github.com/brimsec/zq/zqe"
 )
 
-// statReadCloser implements zbuf.ReadCloser
+// statReadCloser implements zbuf.ReadCloser.
 type statReadCloser struct {
 	ctx    context.Context
 	cancel context.CancelFunc
@@ -89,15 +89,15 @@ func (s *statReadCloser) indexRecord(si SpanInfo, zardir iosrc.URI, indexPath st
 	}
 	if s.indexBuilders[indexPath] == nil {
 		keycols := make([]zng.Column, len(info.Keys))
-		for i := range info.Keys {
+		for i, k := range info.Keys {
 			keycols[i] = zng.Column{
-				Name: info.Keys[i].Name,
+				Name: k.Name,
 				Type: zng.TypeString,
 			}
 		}
 		keyrec := s.zctx.MustLookupTypeRecord(keycols)
 
-		indexBuilder := zng.NewBuilder(s.zctx.MustLookupTypeRecord([]zng.Column{
+		s.indexBuilders[indexPath] = zng.NewBuilder(s.zctx.MustLookupTypeRecord([]zng.Column{
 			zng.NewColumn("type", zng.TypeString),
 			zng.NewColumn("log_id", zng.TypeString),
 			zng.NewColumn("index_id", zng.TypeString),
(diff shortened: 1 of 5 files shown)


archive/stat.go
```diff
@@ -0,0 +1,170 @@
...
+		}))
+	}
+
+	rec := s.chunkBuilder.Build(
+		zng.EncodeString("chunk"),
+		zng.EncodeString(string(si.LogID)),
+		zng.EncodeTime(si.Span.Ts),
+		zng.EncodeDuration(si.Span.Dur),
+		zng.EncodeUint(uint64(fi.Size())),
+	).Keep()
```
nwt, line 69: Calling Keep does an extra copy of both the record and its bytes. (zng.Builder.Build always starts with a fresh (nil, really) buffer because it calls zcode.Builder.Reset before using its buffer.)

archive/stat.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	rec := s.indexBuilders[indexPath].Build(
+		zng.EncodeString("index"),
+		zng.EncodeString(string(si.LogID)),
+		zng.EncodeString(indexPath),
+		zng.EncodeString(s.ark.indexes[indexPath].Type),
+		zng.EncodeUint(uint64(info.Size)),
+		keybytes,
+	).Keep()
```
nwt, line 126: As above, calling Keep does an extra work.

archive/stat.go
```diff
@@ -0,0 +1,170 @@
...
+	if s.indexBuilders == nil {
+		s.indexBuilders = make(map[string]*zng.Builder)
+	}
+	if s.indexBuilders[indexPath] == nil {
+		keycols := make([]zng.Column, len(info.Keys))
+		for i := range info.Keys {
+			keycols[i] = zng.Column{
+				Name: info.Keys[i].Name,
+				Type: zng.TypeString,
+			}
```
nwt, line 96:
```suggestion
		for i, k := range info.Keys {
			keycols[i] = zng.Column{
				Name: k.Name,
				Type: zng.TypeString,
			}
```

archive/stat.go
```diff
@@ -0,0 +1,170 @@
...
+
+		indexBuilder := zng.NewBuilder(s.zctx.MustLookupTypeRecord([]zng.Column{
+			zng.NewColumn("type", zng.TypeString),
+			zng.NewColumn("log_id", zng.TypeString),
+			zng.NewColumn("index_id", zng.TypeString),
+			zng.NewColumn("index_type", zng.TypeString),
+			zng.NewColumn("size", zng.TypeUint64),
+			zng.NewColumn("keys", keyrec),
+		}))
+		s.indexBuilders[indexPath] = indexBuilder
```
nwt, line 108:
```suggestion
		s.indexBuilders[indexPath] = zng.NewBuilder(s.zctx.MustLookupTypeRecord([]zng.Column{
			zng.NewColumn("type", zng.TypeString),
			zng.NewColumn("log_id", zng.TypeString),
			zng.NewColumn("index_id", zng.TypeString),
			zng.NewColumn("index_type", zng.TypeString),
			zng.NewColumn("size", zng.TypeUint64),
			zng.NewColumn("keys", keyrec),
		}))
```

archive/stat.go
alfred-landrum: Without the call to Keep, I saw test failures due to data races; `Build` returns a pointer to its own `rec zng.Record` field. I could just drop using a zng.Builder entirely - I really just want the loop of building the zcode.Bytes over the columns.

nwt: Ah, I missed that. Keep the Keep, then.

PR #1020, opened 2020-07-21
Add changes to zql.es.js

jameskerr:
Soon the app will only rely on the `zql.es.js` version of the parser. It appears some changes went in to the zql javascript parser that did not make it into the zql.es.js file. 

@henridf, in the RenameProc PR #998 , when you ran "make", did the zql.es.js file not get output? I tried to make it part of the default make tasks but I may have not accounted for all environments.

### DIFF
--- a/zql/zql.es.js
+++ b/zql/zql.es.js
@@ -457,57 +457,49 @@ function peg$parse(input, options) {
             return result
         },
       peg$c95 = function(base, refs) { return text() },
-      peg$c96 = function(first, ref) { return ref },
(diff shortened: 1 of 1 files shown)


henridf:
> @henridf, in the RenameProc PR #998 , when you ran "make", did the zql.es.js file not get output? I tried to make it part of the default make tasks but I may have not accounted for all environments.

Yes, it did but I was missing some dependency and commented out the dependency during dev and then forgot about it before putting up the PR :)

PR #1044, opened 2020-07-24
Add s3 source to iosrc by default

### LINKED ISSUE #992: Add s3 registry to iosrc by default
Currently for the zq/zar commands the s3 Source has to be added to the iosrc.Registry so that s3 uri will work. Let's rearrange things the s3 Source is added to the registry by default and the statement in main() is not needed.

mattnibs:
Move s3io.Source to package iosrc and add s3 as a default Source to
the scheme registry.

Remove the now unnecessary AddSource calls in the main of zar, zqd, zq.

Remove the now unused iosrc.Registry struct.

closes #992


pkg/s3io/s3io.go
```diff
@@ -30,7 +30,7 @@ func IsS3Path(path string) bool {
 }
 
 func parsePath(path string) (bucket, key string, err error) {
-	u, err := iosrc.ParseURI(path)
+	u, err := iosrc.Parse(path)
```
nwt, line 33: Can you leave URI in pkg/iosrc if you use net/url.Parse here?

mattnibs: It could... I don't think its bad uri is in its own package (kind of makes sense imho).

mattnibs: How was this not cause compilation errors?!

mattnibs: Not opposed going your way though.

mattnibs: After having the weekend away from this I'm going to get rid of the move URI to its own package. If I want to do this, this should go in another pr.

pkg/iosrc/iosrc.go
```diff
@@ -44,95 +42,49 @@ type ReplacerAble interface {
...
+		return false, err
 	}
 	return source.Exists(uri)
 }
 
 func Remove(uri URI) error {
-	source, err := DefaultRegistry.Source(uri)
+	source, err := GetSource(uri)
 	if err != nil {
 		return nil
```
nwt, line 72: Should this return err?

pkg/iosrc/iosrc.go
```diff
@@ -44,95 +42,49 @@ type ReplacerAble interface {
...
-	r.schemes[scheme] = loader
-}
-
-func (r *Registry) NewReader(uri URI) (io.ReadCloser, error) {
-	s, err := r.Source(uri)
+func NewReader(uri URI) (io.ReadCloser, error) {
+	source, err := GetSource(uri)
 	if err != nil {
-		return nil, err
+		return nil, nil
```
nwt, line 48: Should this return err?

pkg/iosrc/iosrc.go
```diff
@@ -44,95 +42,49 @@ type ReplacerAble interface {
...
+		return nil, fmt.Errorf("unknown scheme: %q", scheme)
+	}
+	return source, nil
 }
 
 func Stat(uri URI) (Info, error) {
-	source, err := DefaultRegistry.Source(uri)
+	source, err := GetSource(uri)
 	if err != nil {
 		return nil, nil
```
nwt, line 89: Should this return err?

pkg/iosrc/iosrc.go
mattnibs: whoops

PR #1048, opened 2020-07-28
Fix tests that do not remove temporary directories

nwt:


zdx/zdx_test.go
```diff
@@ -40,6 +40,7 @@ func buildTestTable(t *testing.T, zngText string) string {
 	if err != nil {
 		t.Error(err)
 	}
+	t.Cleanup(func() { os.RemoveAll(dir) })
```
mattnibs, line 43: You can remove the RemoveAll on line 97

nwt: Indeed. Done in cb37b5ff045a005793a087237e760f0e8c8f9d39.

PR #1054, opened 2020-07-31
peg go parser: return go maps

### LINKED ISSUE #996: peg go parser should not build ast structs
The output from the ZQL peg parser is constructed differently for Go and Javascript: For Javascript, a Javascript object is created. For Go, structs from package `ast` are constructed & returned. As evidenced by https://github.com/brimsec/zq/issues/985 , errors in the struct construction can cause different output AST's that can't be caught without thorough testing.

A previous implementation of our parser handling for Go generated `map[string]interface{}`'s; after some discussion today, we'd like to revert back to that approach.

Note to whoever works on this: The https://github.com/mitchellh/mapstructure/ Go library may be useful.

mattnibs:
Make the go peg parser behave like the javascript peg parser and
return map[string]interface{}'s instead of golang structs from
package ast.

Closes #996

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,8 @@ require (
 	github.com/kballard/go-shellquote v0.0.0-20180428030007-95032a82bc51
 	github.com/klauspost/compress v1.10.3 // indirect
 	github.com/mccanne/charm v0.0.3-0.20191224190439-b05e1b7b1be3
-	github.com/mccanne/joe v0.0.0-20181124064909-25770742c256
+	github.com/mccanne/joe v0.0.0-20200731213236-7c7845acf98b
+	github.com/mitchellh/mapstructure v1.3.3
 	github.com/peterh/liner v1.1.0
 	github.com/pierrec/lz4/v4 v4.0.1
 	github.com/pmezard/go-difflib v1.0.0
--- a/zql/zql.peg
+++ b/zql/zql.peg
@@ -6,8 +6,8 @@
     package zql
 }
 #define RETURN(x) return x, nil
+#define VAR(x) var x
 #define NULL nil
-#define INIT_ASSIGN_VAR(x, var) x := var
 #define ARRAY(...) []interface{}{__VA_ARGS__}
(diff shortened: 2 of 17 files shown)


go.mod
```diff
@@ -41,3 +42,5 @@ require (
 )
 
 replace github.com/minio/minio => github.com/brimsec/minio v0.0.0-20200716214025-90d56627f750
+
+replace github.com/mccanne/joe => github.com/mattnibs/joe v0.0.0-20200731205736-7a14568ac179
```
mattnibs, line 46: See: https://github.com/mccanne/joe/pull/3

mccanne: I merge your PR on mccanne/joe

mccanne:
How did you find mapstructure?  That was kind of what we were trying to do with jab but never finished :-)

mattnibs:
> How did you find mapstructure? That was kind of what we were trying to do with jab but never finished :-)

It was @alfred-landrum 's clutch suggestion

alfred-landrum:
@mattnibs : if you haven't already done so, please create a Brim branch so we can have the Brim integration tests run against this change.

mattnibs:
@al

> @mattnibs : if you haven't already done so, please create a Brim branch so we can have the Brim integration tests run against this change.

@alfred-landrum good idea, will do

zql/zql.peg
```diff
@@ -37,25 +41,29 @@
...
 
 
 start = _? ast:query _? EOF { RETURN(ast) }
 
 query
   = procs:procChain {
-      INIT_ASSIGN_VAR(filt, makeFilterProc(makeMatchAll()))
-      RETURN(makeSequentialProc(PREPEND(filt, procs)))
+      VAR(filt) = MAP("op": "FilterProc", "filter": MAP("op": "MatchAll"))
+	  RETURN(MAP("op": "SequentialProc", "procs": PREPEND(filt, procs)))
```
nwt, line 56: The indentation here is inconsistent, as it is in a number of other places in this file.

mattnibs: quite right

Issue #1040, opened 2020-07-23
Pass paths to scanners


henridf:
Assuming a flowgraph that has been split per #1022  and has `N` sources, these sources need to know which archive files to read from. This can be done by having all of the sources read from a queue of paths into which all the archive files are written (in their sort order). This issue consists of doing that.

(Implementation note: a first step could be to implement this on top of the existing system, pre-#1022/#1036 with a single source reading from the queue, but we can also go straight to implementing it with `N` sources post-#1022/#1036).

PR #1061, opened 2020-08-04
dont buffer terminal output

alfred-landrum:
While working on #1040 , I was running a locally modified `zar zq` that concurrently executes on archive files, and was surprised that I waited until the entire archive had been searched before seeing results. It turns out that was due to output buffering. This change disables buffering for terminal output, so that results are seen quickly.

### DIFF
--- a/emitter/file.go
+++ b/emitter/file.go
@@ -53,12 +53,10 @@ func NewFileWithSource(path iosrc.URI, flags *zio.WriterFlags, source iosrc.Sour
 	if path.Scheme == "stdio" {
 		// Don't close stdio in case we live inside something
 		// that has multiple stdio users.
-		nc := &noClose{f}
-		if isTerminal(f) {
+		wc = &noClose{f}
+		if !isTerminal(f) {
 			// Don't buffer terminal output.
-			wc = &noClose{f}
-		} else {
-			wc = bufwriter.New(nc)
+			wc = bufwriter.New(wc)
 		}
 	} else {
 		wc = bufwriter.New(f)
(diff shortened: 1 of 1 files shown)


emitter/file.go
```diff
@@ -33,20 +34,39 @@ func NewFile(path string, flags *zio.WriterFlags) (*zio.Writer, error) {
...
 	if path.Scheme == "stdio" {
-		// Don't close stdout in case we live inside something
-		// here that runs multiple instances of this to stdout.
-		f = &noClose{os.Stdout}
+		// Don't close stdio in case we live inside something
+		// that has multiple stdio users.
+		nc := &noClose{f}
+		if isTerminal(f) {
+			// Don't buffer terminal output.
+			wc = &noClose{f}
```
mattnibs, line 59: nit:
```suggestion
			wc = nc
```

emitter/file.go
```diff
@@ -33,20 +34,39 @@ func NewFile(path string, flags *zio.WriterFlags) (*zio.Writer, error) {
...
+		} else {
+			wc = bufwriter.New(nc)
+		}
```
mattnibs, line 62:
```suggestion
		wc = &noClose{f}
		if !isTerminal(f) {
			// Don't buffer terminal output.
			wc = bufwriter.New(nc)
		}
```

nwt: Maybe do this for any terminal?
```go
	f, err := source.NewWriter(path)
	if err != nil {
		return nil, err
	}
	wc := io.WriteCloser(f)
	if !isTerminal(f) {
		wc = bufwriter.New(f)
	}
	if path.Scheme == "stdio" {
		// Don't close stdio in case we live inside something
		// that has multiple stdio users.
		wc = &noClose{wc}
	}
```

Issue #681, opened 2020-04-28
boyer-moore design


alfred-landrum:
Document the design & tasks needed for https://github.com/brimsec/zq/issues/682 .

nwt:
Here's how I see this going. Questions and suggestions are welcome!

1. Add these interfaces to package driver. (NewScanner has `filter ast.BooleanExpr` because we'll need it in step 6.
```go
// Scanner is the interface implemented by a record source that also provides
// statistics.
type Scanner interface {
	Pull() (zbuf.Batch, error)
	Stats() api.ScannerStats
}

// ScannerAble is the interface implemented by zbuf.Readers that provide an
// optimized implementation of the Scanner interface.  (Those that do not can be
// wrapped with scanner.NewScanner to provide the Scanner interface.)
type ScannerAble interface {
	NewScanner(ctx context.Context, filter ast.BooleanExpr, s nano.Span) Scanner
}
```

2. Make this change to driver.CompileWarningsChCustom.
```diff
 	filterAst, program := liftFilter(program)
-	scanner, err := newScanner(ctx, reader, filterAst, span)
-	if err != nil {
-		return nil, err
+	var scanner Scanner
+	if sa, ok := reader.(ScannerAble); ok {
+		var filter ast.BooleanExpr
+		if filterAst != nil {
+			filter = filterAst.Filter
+		}
+		scanner = sa.NewScanner(ctx, filter, span)
+	} else {
+		var err error
+		scanner, err = newScanner(ctx, reader, filterAst, span)
+		if err != nil {
+			return nil, err
+		}
 	}
```

3. Modify zngio.Reader to implement driver.ScannerAble.

4. Modify zngio.Reader to incorporate framing and compression from boom/common/tuple/{reader,writer}.go and boom/common/compression/lz4.go.  This has to live in harmony with the existing end-of-stream markers and zngio.Seeker. An alternative is the `<comp-type-code><comp-kind><uvarint-len><len bytes of compressed zng data>` approach from #521. I favor the first approach primarily because I have a better feeling for how it will look (owing to familiarity with the boom implementation), but if I'm open to the second approach if anyone prefers it.

5. Modify zio/zngio.Reader to incorporate buffer management from boom/common/tuple/{buffer,chunk}.go.

[...]

PR #1063, opened 2020-08-05
Add a zngio scanner

nwt:
This implements steps 1 through 3 of #681.

* Change the ZNG specification to allow only value messages in a
  compressed message block.

* Add scanner.ScannerAble and scanner.Scanner interfaces.

* Modify zngio.Reader to implement scanner.ScannerAble with efficient
  handling for compressed value message blocks.

* Add scanner.NewMultiScanner with knowledge of scanner.ScannerAble.

* Change driver.Compile to accept a []zbuf.Reader parameter instead of a
  single zbuf.Reader and to use scanner.NewMultiScanner.


nwt:
System tests are broken, but I wanted to get this out before fixing them.

philrz:
I read through the spec changes FWIW and they look good to me. 👍

scanner/scanner.go
```diff
@@ -4,16 +4,53 @@ import (
...
+// api.ScannerStats.
+type ScannerStats struct {
+	BytesRead      int64
+	BytesMatched   int64
+	RecordsRead    int64
+	RecordsMatched int64
+}
+
+// Accumulate updates its receiver by adding to it the values in ss.
+func (s *ScannerStats) Accumulate(ss *ScannerStats) {
```
mattnibs, line 36: Nice!

driver/compile.go
```diff
@@ -28,7 +27,7 @@ type Config struct {
 // and compiles it into a runnable flowgraph, returning a
 // proc.MuxOutput that which brings together all of the flowgraphs
 // tails, and is ready to be Pull()'d from.
-func Compile(ctx context.Context, program ast.Proc, zctx *resolver.Context, reader zbuf.Reader, cfg Config) (*MuxOutput, error) {
+func Compile(ctx context.Context, program ast.Proc, zctx *resolver.Context, readers []zbuf.Reader, cfg Config) (*MuxOutput, error) {
```
mattnibs, line 30: I'd prefer to do this:
```suggestion
func Compile(ctx context.Context, program ast.Proc, zctx *resolver.Context, cfg Config, readers ...zbuf.Reader) (*MuxOutput, error) {
```

scanner/multiscanner.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/brimsec/zq/pkg/nano"
+	"github.com/brimsec/zq/zbuf"
+)
+
+// NewMultiScanner returns a Scanner for the logical concatenation of the
+// provided readers that filters records by filterExpr and span. Readers are
+// read sequentially, and when all records are consumed, Pull will return a nil
+// batch and nil error. If any reader returns a non-nil error, Pull will return
+// that error.
+func NewMultiScanner(ctx context.Context, readers []zbuf.Reader, filterExpr ast.BooleanExpr, span nano.Span) (Scanner, error) {
```
mattnibs, line 19: We don't have to do this now but at some point it we should consider adopting the Options pattern here:
```suggestion
func NewMultiScanner(ctx context.Context, readers []zbuf.Reader, opts ...Options) (Scanner, error) {
```

nwt:
The driver.Compile changes here (particularly use of scanner.NewMultiScanner) conflict with work @alfred-landrum is doing for #1040 and will probably be discarded. In light of that, I've converted this PR to a draft. I'll mark it ready for review after I've updated it to align with the #1040 work.

driver/compile.go
nwt: I've removed this change to the signature of driver.Compile.

scanner/multiscanner.go
nwt: I've removed this function.

zbuf/combiner.go
```diff
@@ -32,27 +32,33 @@ func CmpTimeReverse(a, b *zng.Record) bool {
...
-// NewCombiner returns a Combiner that merges zbuf.Readers into
+// NewCombiner returns a ReaderCloser that merges zbuf.Readers into
 // a single Reader. If the ordering of the input readers matches
 // the direction specified, the output records will maintain
 // that order.
-func NewCombiner(readers []Reader, cmp RecordCmpFn) *Combiner {
-	return &Combiner{
+func NewCombiner(readers []Reader, cmp RecordCmpFn) ReadCloser {
+	if len(readers) == 1 {
+		// XXX
```
mattnibs, line 41: Seems like a reasonable check, can this comment be removed?

nwt: It certainly can. Done in 79d7a50f8d77978dcfb75a3beb8f1a7383403385.

zng/docs/spec.md
```diff
@@ -351,27 +351,26 @@ previously defined type, the appropriate typedef control code must
...
-### 3.1.3 Compressed Message Block
+### 3.1.3 Compressed Value Message Block
 
-Following a header byte of 0x86 is a compressed message block.  Such a
-block comprises a compressed sequence of control and value messages.
-The sequence can include any message except for end-of-stream markers
-and compressed message blocks.
+Following a header byte of 0x86 is a compressed value message block.
+Such a block comprises a compressed sequence of value messages.  The
+sequence must not include control messages.
```
nwt, line 358: I made this ZNG specification change in anticipation of adding Boyer-Moore string search inside zio/zngio.zngScanner.scanUncompressed. Excluding control messages like this means we can immediately discard a block if the search fails, without looking for control messages.

It accounts for the the changes in ZNG outputs and the zng-output-check failure.

zio/zngio/reader.go
```diff
@@ -126,29 +141,24 @@ again:
...
-			}
 			r.reset()
 		case zng.CtrlCompressed:
-			if r.uncompressed != nil {
-				return nil, nil, errors.New("zngio: CtrlCompressed in compressed data")
+			if err := r.readCompressed(); err != nil {
+				return 0, nil, err
 			}
-			err = r.readCompressed()
+			return -zng.CtrlCompressed, r.uncompressed.Bytes(), nil
```
alfred-landrum, line 149: This is being exposed like this in preparation for the boyer-moore work, right? (instead of just reading the compressed data & then falling out to the `goto again`, like the `CtrlEOS` handling)

alfred-landrum: never mind, I read down to `zngScanner.Pull`

PR #1065, opened 2020-08-05
CI test for ZNG output check

philrz:
See the comments in the script in this PR for the background/motivation.

An example of the test passing with current `zq` and [zq-sample-data](https://github.com/brimsec/zq-sample-data):

https://github.com/brimsec/zq/actions/runs/196891811

I also simulated a failure by creating a branch on which I reverted the commit `6b19b3b` where ZNG compression was enabled, and let this script run against the ZNG data currently in zq-sample-data. Since the current ZNG data in zq-sample-data is compressed but the failure branch was creating uncompressed output, the test failed as shown here:

https://github.com/brimsec/zq/actions/runs/196887459

Before I started writing this test, I'd thought that maybe I'd only let this Action run post-merge since we expect it to fail rarely and it's not a catastrophe if the zq-sample-data is briefly out of date. However, once I saw how it could run as a parallel job, I figured I'd take a shot at letting it run parallel alongside the existing CI. The overall runtime for the CI seems roughly in line with where it was before, and I figure developers might appreciate knowing this info sooner if they introduced some kind of ZNG output bug that wouldn't otherwise have been caught. However, if anyone wants to hold me to my initial sentiment, I won't fight it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,6 +73,9 @@ test-zeek: bin/$(ZEEKPATH)
 perf-compare: build $(SAMPLEDATA)
 	scripts/comparison-test.sh
 
+zng-output-check: build $(SAMPLEDATA)
+	scripts/zng-output-check.sh
+
 # If the build recipe changes, please also change npm/build.
 build:
 	@mkdir -p dist
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - master
+      - zng-output-check
 
 jobs:
   test:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -46,6 +46,6 @@ jobs:
       - uses: actions/setup-go@v2
         with:
           go-version: '1.14'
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -73,6 +73,9 @@ test-zeek: bin/$(ZEEKPATH)
 perf-compare: build $(SAMPLEDATA)
 	scripts/comparison-test.sh
 
+zng-output-check: build $(SAMPLEDATA)
+	scripts/zng-output-check.sh
```
nwt, line 77: Not sure how others feel, but I'd prefer this over a separate script.
```mk
zng-output-check: build $(SAMPLEDATA)
	PATH=$(CURDIR)/dist:$PATH zq-sample-data/scripts/check_md5sums.sh tzng
	PATH=$(CURDIR)/dist:$PATH zq-sample-data/scripts/check_md5sums.sh zng
	PATH=$(CURDIR)/dist:$PATH zq-sample-data/scripts/check_md5sums.sh zng-uncompressed
```

alfred-landrum: I like the script since it gives more documentation space that would clutter the makefile.

PR #1070, opened 2020-08-07
Simplify zql/Makefile

nwt:
I think this is clearer. Behavior should be unchanged.

### DIFF
--- a/zql/Makefile
+++ b/zql/Makefile
@@ -1,62 +1,37 @@
-deps = $(CURDIR)/deps
-
-PEGJS = $(deps)/bin/pegjs
-ROLLUP = $(deps)/bin/rollup
-PIGEON = $(deps)/bin/pigeon
-GOIMPORTS = $(deps)/bin/goimports
-
+GOIMPORTS_VERSION = v0.0.0-20200204192400-7124308813f3
 PEGJS_VERSION = 0.10.0
 PIGEON_VERSION = v1.0.1-0.20190520151103-9fec3898cef8
-GOIMPORTS_VERSION = v0.0.0-20200204192400-7124308813f3
 
-PEGJS_INSTALLED = $(shell test -x $(PEGJS) && $(PEGJS) --version)
-PIGEON_INSTALLED = $(shell test -d $(deps) && (cd $(deps); go list -m -f {{.Version}} github.com/mna/pigeon || echo ""))
-GOIMPORTS_INSTALLED = $(shell test -d $(deps) && (cd $(deps); go list -m -f {{.Version}} golang.org/x/tools || echo ""))
-
-PEGJS_ARGS = --allowed-start-rules start,Expression
+deps = $(CURDIR)/deps
+npm = npm --global --prefix $(deps)
 
 all: zql.go zql.js zql.es.js
 
 run: zql.go zql.js
 	go run ./main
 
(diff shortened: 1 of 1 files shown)


zql/Makefile
```diff
@@ -1,62 +1,37 @@
...
 PIGEON_VERSION = v1.0.1-0.20190520151103-9fec3898cef8
-GOIMPORTS_VERSION = v0.0.0-20200204192400-7124308813f3
-
-PEGJS_INSTALLED = $(shell test -x $(PEGJS) && $(PEGJS) --version)
-PIGEON_INSTALLED = $(shell test -d $(deps) && (cd $(deps); go list -m -f {{.Version}} github.com/mna/pigeon || echo ""))
-GOIMPORTS_INSTALLED = $(shell test -d $(deps) && (cd $(deps); go list -m -f {{.Version}} golang.org/x/tools || echo ""))
 
-PEGJS_ARGS = --allowed-start-rules start,Expression
+deps = $(CURDIR)/deps
+npm = npm --global --prefix $(deps)
```
mattnibs, line 6: I know this was already here but I don't think we should be using the --global flag here.

nwt: I'm open to removing it. For context, I originally added it so all of the dependencies would land in $(deps)/bin.

What's your thinking on why we shouldn't use it?

mattnibs: I don't think a Makefile should globally install dependencies into my system.

nwt: I don't disagree. The presence of --prefix $(deps) prevents npm from touching anything outside of $(deps), even with --global.

If you'd like to get rid of it anyway, I'm happy to, like this.

```diff
$ g diff
diff --git a/zql/Makefile b/zql/Makefile
index 67e3cea..ff2b202 100644
--- a/zql/Makefile
+++ b/zql/Makefile
@@ -3,7 +3,7 @@ PEGJS_VERSION = 0.10.0
 PIGEON_VERSION = v1.0.1-0.20190520151103-9fec3898cef8

 deps = $(CURDIR)/deps
-npm = npm --global --prefix $(deps)
+npm = npm --prefix $(deps)

 all: zql.go zql.js zql.es.js

@@ -27,11 +27,11 @@ endif

 .PHONY: zql.js
 zql.js:
-ifeq "$(shell $(deps)/bin/pegjs --version 2>&1 | fgrep $(PEGJS_VERSION))" ""
+ifeq "$(shell $(deps)/node_modules/.bin/pegjs --version 2>&1 | fgrep $(PEGJS_VERSION))" ""
        $(npm) install pegjs@$(PEGJS_VERSION) rollup @rollup/plugin-commonjs
 endif
-       cpp -E -P zql.peg | $(deps)/bin/pegjs --allowed-start-rules start,Expression -o $@
+       cpp -E -P zql.peg | $(deps)/node_modules/.bin/pegjs --allowed-start-rules start,Expression -o $@

 .PHONY: zql.es.js
 zql.es.js: zql.js
-       $(deps)/bin/rollup --format es --plugin commonjs $^ -o $@
+       $(deps)/node_modules/.bin/rollup --format es --plugin commonjs $^ -o $@
```

mattnibs: Ugh. Yeah let's just keep it global for now.

PR #1072, opened 2020-08-07
zqd listen: s3 datapath

### LINKED ISSUE #1051: zqd support for S3 datadir
zqd expects it's 'datadir' parameter to point to a filesystem directory, where it creates a directory per space to hold the spaces config.json and possibly the space for the data. If zqd could support a datadir pointing to an S3 location, we would be able to run a zqd instance where S3 is the only persistent storage needed, as long as the only types of space storage utilized is the `archivestore`, and data is imported via `zar import`. Note that the current import API assumes that the requester & zqd have a shared filesystem of some kind, so that api would not be expected to work without additional implementation changes (to be filed & scheduled as a separate issue).

mattnibs:
Support starting zqd with the datapath set to an s3 path. For now zqd
instances with s3 path's only allow creation of archive spaces. For s3
backed spaces return a 400 error on attempts to create a filestore space.

Closes #1051

### DIFF
--- a/cmd/zqd/listen/command.go
+++ b/cmd/zqd/listen/command.go
@@ -12,7 +12,6 @@ import (
 	"os"
 	"os/exec"
 	"os/signal"
-	"path/filepath"
 
 	"github.com/brimsec/zq/cmd/zqd/logger"
 	"github.com/brimsec/zq/cmd/zqd/root"
@@ -75,7 +74,7 @@ type Command struct {
 func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
 	c := &Command{Command: parent.(*root.Command)}
 	f.StringVar(&c.listenAddr, "l", ":9867", "[addr]:port to listen on")
-	f.StringVar(&c.conf.Root, "datadir", ".", "data directory")
+	f.StringVar(&c.conf.Root, "data", "", "data location")
 	f.StringVar(&c.zeekRunnerPath, "zeekrunner", "", "path to command that generates zeek logs from pcap data")
 	f.BoolVar(&c.pprof, "pprof", false, "add pprof routes to api")
 	f.BoolVar(&c.prom, "prometheus", false, "add prometheus metrics routes to api")
@@ -128,11 +127,6 @@ func (c *Command) Run(args []string) error {
 }
 
 func (c *Command) init() error {
-	var err error
-	c.conf.Root, err = filepath.Abs(c.conf.Root)
-	if err != nil {
-		return err
-	}
 	if err := c.loadConfigFile(); err != nil {
 		return err
 	}
--- a/cmd/zqd/listen/command.go
+++ b/cmd/zqd/listen/command.go
@@ -74,7 +74,7 @@ type Command struct {
 func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
 	c := &Command{Command: parent.(*root.Command)}
 	f.StringVar(&c.listenAddr, "l", ":9867", "[addr]:port to listen on")
-	f.StringVar(&c.conf.Root, "data", "", "data location")
+	f.StringVar(&c.conf.Root, "data", ".", "data location")
 	f.StringVar(&c.zeekRunnerPath, "zeekrunner", "", "path to command that generates zeek logs from pcap data")
 	f.BoolVar(&c.pprof, "pprof", false, "add pprof routes to api")
 	f.BoolVar(&c.prom, "prometheus", false, "add prometheus metrics routes to api")
--- a/pkg/iosrc/uri.go
+++ b/pkg/iosrc/uri.go
@@ -19,6 +19,9 @@ const (
 // treated as files and resolved as absolute paths using filepath.Abs.
 // path is an empty, Scheme is set to file.
 func ParseURI(path string) (URI, error) {
+	if path == "" {
+		return URI{}, nil
+	}
 	// First resolve stdio keywords in to fully-formed uri.
 	path = stdio(path)
 	var err error
@@ -53,9 +56,9 @@ func (p URI) AppendPath(elem ...string) URI {
 	return p
 }
 
-func (p URI) String() string {
-	u := url.URL(p)
-	return (&u).String()
+func (u URI) String() string {
+	url := url.URL(u)
+	return url.String()
 }
 
 func (u URI) RelPath(target URI) string {
@@ -63,6 +66,23 @@ func (u URI) RelPath(target URI) string {
 		u.Path += "/"
 	}
 	return strings.TrimPrefix(target.Path, u.Path)
+}
+
+func (u URI) IsZero() bool {
+	return u == URI{}
+}
+
+func (u URI) MarshalText() ([]byte, error) {
+	return []byte(u.String()), nil
+}
+
+func (u *URI) UnmarshalText(b []byte) error {
+	uri, err := ParseURI(string(b))
+	if err != nil {
+		return err
+	}
+	*u = uri
+	return nil
 }
 
 // Maybe rawurl is of the form scheme:path.
--- a/tests/suite/zqd/s3/info.yaml
+++ b/tests/suite/zqd/s3/info.yaml
@@ -4,7 +4,7 @@ script: |
   echo ===
   zapi -h $ZQD_HOST new -d ./root -k archivestore testsp
   echo ===
-  zapi -h $ZQD_HOST -s testsp get -f tzng "s=harefoot-raucous"
+  zapi -h $ZQD_HOST info testsp | egrep -v 'data_path|id|pcap_path'
 
 inputs:
   - name: babble.tzng
@@ -18,5 +18,10 @@ outputs:
       ===
       testsp: space created
       ===
-      #0:record[ts:time,s:string,v:int64]
-      0:[1587508881.0613914;harefoot-raucous;137;]
+      testsp
+        name:         testsp
+        storage_kind: archivestore
+        span:         2020-04-21T22:40:30Z+2h43m9.993714061s
+        size:         34.34KB
+        pcap_support: false
+        pcap_size:    0B
--- a/tests/suite/zqd/services.sh
+++ b/tests/suite/zqd/services.sh
@@ -33,7 +33,7 @@ export AWS_ACCESS_KEY_ID=minioadmin
 export AWS_SECRET_ACCESS_KEY=minioadmin
 export AWS_S3_ENDPOINT=http://localhost:$(cat $portdir/minio)
 
-zqd listen -l=localhost:0 -portfile="$portdir/zqd" -data=$zqdroot 2> zqd.error &
+zqd listen -l=localhost:0 -portfile="$portdir/zqd" -data="$zqdroot" 2> zqd.error &
 zqdpid=$!
 trap "rm -rf $portdir; kill -9 $miniopid $zqdpid" EXIT
 
--- a/tests/suite/zqd/services.sh
+++ b/tests/suite/zqd/services.sh
@@ -13,8 +13,13 @@ function awaitfile {
   done
 }
 
+zqdroot=$1
+if [ -z "$zqdroot" ]; then
+  zqdroot=zqdroot
+  mkdir -p zqdroot
+fi
+
 mkdir -p s3/bucket
-mkdir -p zqdroot
 portdir=$(mktemp -d)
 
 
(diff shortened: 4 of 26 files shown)


tests/suite/zqd/s3/info.yaml
```diff
@@ -1,10 +1,10 @@
 script: |
-  source services.sh
+  source services.sh
```
mattnibs, line 2:
```suggestion
  source services.sh
```

zqd/api/api.go
```diff
@@ -98,7 +99,7 @@ func (s *SpaceID) Set(str string) error {
 type SpaceInfo struct {
 	ID          SpaceID      `json:"id"`
 	Name        string       `json:"name"`
-	DataPath    string       `json:"data_path"`
+	DataPath    iosrc.URI    `json:"data_path"`
```
mattnibs, line 102: I think this should also change to data_uri but didn't want to force an api migration.

zqd/handlers_test.go
```diff
@@ -206,29 +206,29 @@ func TestSpaceList(t *testing.T) {
...
+				DataPath:    c.Root.AppendPath(string(sp.ID)),
 				StorageKind: storage.FileStore,
 			})
 		}
 
 		list, err := client.SpaceList(ctx)
 		require.NoError(t, err)
 		sort.Slice(list, func(i, j int) bool { return list[i].Name < list[j].Name })
-		require.Equal(t, expected, list)
+		require.Equal(t, list[0], expected[0])
```
mattnibs, line 217:
```suggestion
		require.Equal(t, expected, list)
```

zqd/handlers_test.go
```diff
@@ -219,16 +219,16 @@ func TestSpaceList(t *testing.T) {
...
 
-	c, client, done = newCoreAtDir(t, c.Root)
+	_, client, done = newCoreAtDir(t, c.Root.Filepath())
 	defer done()
 
 	list, err := client.SpaceList(ctx)
 	require.NoError(t, err)
 	sort.Slice(list, func(i, j int) bool { return list[i].Name < list[j].Name })
-	require.Equal(t, expected, list)
+	require.Equal(t, expected, list, "failedhere")
```
mattnibs, line 231:
```suggestion
	require.Equal(t, expected, list)
```

zqd/space/config.go
```diff
@@ -91,24 +217,25 @@ func validateName(names map[string]api.SpaceID, name string) error {
 }
 
 // safeName converts the proposed name to a name that adheres to the constraints
-// placed on a space's name (i.e. follows the name regex and is unique). In
-// order to ensure the generated name is unique this should be called with the
-// manager lock held.
-func safeName(names map[string]api.SpaceID, proposed string) string {
+// placed on a space's name (i.e. follows the name regex.
```
mattnibs, line 220:
```suggestion
// placed on a space's name (i.e. follows the name regex).
```

mattnibs:
brimsec/brim#976 passing

zqd/space/config.go
```diff
@@ -55,26 +74,133 @@ func (c config) subspaceIndex(id api.SpaceID) int {
...
+	}
+	if err := r.Close(); err != nil {
+		return c, err
+	}
+	var vc versionCheck
+	if err := json.Unmarshal(data, &vc); err != nil {
+		return c, err
+	}
+	if vc.Version > configVersion {
+		return c, fmt.Errorf("Space config version '%d' ahead of binary version '%d'", vc.Version, configVersion)
```
nwt, line 97:
```suggestion
		return c, fmt.Errorf("space config version %d ahead of binary version %d", vc.Version, configVersion)
```

zqd/space/config.go
```diff
@@ -55,26 +74,133 @@ func (c config) subspaceIndex(id api.SpaceID) int {
...
+	var m migrator
+	var err error
+	for version < configVersion {
+		switch version {
+		case 0:
+			m = migrateConfigV1
+		case 1:
+			m = migrateConfigV2
+		default:
+			return c, fmt.Errorf("Unsupported config migration '%d'", version)
```
nwt, line 121:
```suggestion
			return c, fmt.Errorf("unsupported config migration %d", version)
```

zqd/space/config.go
```diff
@@ -55,26 +74,133 @@ func (c config) subspaceIndex(id api.SpaceID) int {
...
+		return c, err
+	}
+	data, err := ioutil.ReadAll(r)
+	if err != nil {
+		r.Close()
+		return c, err
+	}
+	if err := r.Close(); err != nil {
+		return c, err
+	}
```
nwt, line 91: It's fine to ignore a Close error from a reader.
```suggestion
	defer r.Close()
	data, err := ioutil.ReadAll(r)
	if err != nil {
		return c, err
	}
```

pkg/iosrc/uri.go
```diff
@@ -53,9 +56,12 @@ func (p URI) AppendPath(elem ...string) URI {
...
 
-func (p URI) String() string {
-	u := url.URL(p)
-	return (&u).String()
+func (u URI) String() string {
+	if u.IsZero() {
+		return ""
+	}
+	url := url.URL(u)
+	return (&url).String()
```
nwt, line 64: The if statement isn't necessary. (Tests pass with the original implementation.)

tests/suite/zqd/services.sh
```diff
@@ -28,7 +33,7 @@ export AWS_ACCESS_KEY_ID=minioadmin
 export AWS_SECRET_ACCESS_KEY=minioadmin
 export AWS_S3_ENDPOINT=http://localhost:$(cat $portdir/minio)
 
-zqd listen -l=localhost:0 -portfile="$portdir/zqd" -datadir=./zqdroot &
+zqd listen -l=localhost:0 -portfile="$portdir/zqd" -datadir=$zqdroot 2> zqd.error &
```
nwt, line 36:
```suggestion
zqd listen -l=localhost:0 -portfile="$portdir/zqd" -datadir="$zqdroot" 2> zqd.error &
```
And `2> zqd.error` feels like it's hiding errors in a non-obvious place. Is it necessary?

cmd/zqd/listen/command.go
```diff
@@ -75,7 +74,7 @@ type Command struct {
 func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
 	c := &Command{Command: parent.(*root.Command)}
 	f.StringVar(&c.listenAddr, "l", ":9867", "[addr]:port to listen on")
-	f.StringVar(&c.conf.Root, "datadir", ".", "data directory")
+	f.StringVar(&c.conf.Root, "datadir", "", "data directory")
```
nwt, line 77:
```suggestion
	f.StringVar(&c.conf.Root, "datauri", "", "data location")
```
Since this is no longer has to be a directory.

tests/suite/zqd/s3/info.yaml
```diff
@@ -4,7 +4,7 @@ script: |
   echo ===
   zapi -h $ZQD_HOST new -d ./root -k archivestore testsp
   echo ===
-  zapi -h $ZQD_HOST info testsp | awk '!/id|pcap_path/{print }'
+  zapi -h $ZQD_HOST info testsp | awk '!/id|pcap_path|data_path/{print }'
```
nwt, line 7:
```suggestion
  zapi -h $ZQD_HOST info testsp | egrep -v 'data_path|id|pcap_path'
```

zqd/space/config.go
```diff
@@ -55,26 +74,133 @@ func (c config) subspaceIndex(id api.SpaceID) int {
...
+	if vc.Version > configVersion {
+		return c, fmt.Errorf("Space config version '%d' ahead of binary version '%d'", vc.Version, configVersion)
+	}
+	if vc.Version < configVersion {
+		return migrateConfig(vc.Version, data, spaceURI)
+	}
+	if err := json.Unmarshal(data, &c); err != nil {
 		return c, err
 	}
+	return c, nil
```
nwt, line 105: Nit: Shorten to
```go
err = json.Unmarshal(data, &c)
return c, err
```

zqd/space/config.go
```diff
@@ -55,26 +74,133 @@ func (c config) subspaceIndex(id api.SpaceID) int {
...
+}
+
+type migrator func([]byte, iosrc.URI) (int, []byte, error)
+
+func migrateConfig(version int, data []byte, spaceURI iosrc.URI) (config, error) {
+	var c config
+	var m migrator
+	var err error
```
nwt, line 113: Nit: Declare c and err closer to their uses.

zqd/space/config.go
```diff
@@ -55,26 +74,133 @@ func (c config) subspaceIndex(id api.SpaceID) int {
...
+		w, err = replacer.NewReplacer(uri)
+		if err != nil {
+			return err
+		}
+	} else {
+		w, err = src.NewWriter(uri)
+		if err != nil {
+			return err
+		}
+	}
```
nwt, line 198:
```suggestion
	if replacer, ok := src.(iosrc.ReplacerAble); ok {
		w, err = replacer.NewReplacer(uri)
	} else {
		w, err = src.NewWriter(uri)
	}
	if err != nil {
		return err
	}
```

mattnibs: v nice

zqd/space/manager.go
```diff
@@ -2,63 +2,57 @@ package space
...
 		names:    make(map[string]api.SpaceID),
 		logger:   logger,
 	}
-
-	dirs, err := ioutil.ReadDir(root)
-	if err != nil {
-		return nil, err
+	var err error
+	var dirs []iosrc.URI
+	switch root.Scheme {
```
alfred-landrum, line 35: What about `iosrc` supporting a way to list or walk through the objects of a directory-like uri?

cmd/zqd/listen/command.go
alfred-landrum: `dataloc`?

mattnibs: I'd go with `datauri` but feels like the verbiage might confuse users who go the simple route of selecting a location on the file system. Perhaps just the option `data` with @nwt change for the help text "data location". That or `dataloc`.

zqd/space/manager.go
mattnibs: That's an interesting idea @alfred-landrum . Let me look into that.

mattnibs: We already have the code somewhere in zar/zdx that does this.

tests/suite/zqd/services.sh
mattnibs: I put this in because zqd by default writes logs to stderr which messes up some of my tests for checking the output of stderr. I suppose I could change the zqd logger.Level to "warn"? Thoughts?

nwt: I think changing the logging level is the way to go, and Warn strikes me as appropriate.

cmd/zqd/listen/command.go
nwt: I like -data because it's shorter, but -dataloc also seem fine.

pkg/iosrc/uri.go
```diff
@@ -53,9 +56,9 @@ func (p URI) AppendPath(elem ...string) URI {
 	return p
 }
 
-func (p URI) String() string {
-	u := url.URL(p)
-	return (&u).String()
+func (u URI) String() string {
+	url := url.URL(u)
+	return (&url).String()
```
nwt, line 61:
```suggestion
	return url.String()
```

zqd/space/manager.go
mattnibs: @alfred-landrum if it's alright with you I'm going to hold off on this for now. I agree that there should be some simple iosrc function here but I'm not quite sure what this should look like. Proposing we merge and revisit.

tests/suite/zqd/services.sh
```diff
@@ -28,7 +33,7 @@ export AWS_ACCESS_KEY_ID=minioadmin
 export AWS_SECRET_ACCESS_KEY=minioadmin
 export AWS_S3_ENDPOINT=http://localhost:$(cat $portdir/minio)
 
-zqd listen -l=localhost:0 -portfile="$portdir/zqd" -datadir=./zqdroot &
+zqd listen -l=localhost:0 -portfile="$portdir/zqd" -data=$zqdroot 2> zqd.error &
```
nwt, line 36: $zqdroot might contain spaces.
```suggestion
zqd listen -l=localhost:0 -portfile="$portdir/zqd" -data="$zqdroot" 2> zqd.error &
```
And I'm still in favor of changing the zqd logging level to Warn so we can drop `2> zqd.error`.

mattnibs: The best way to do this is to have a `-loglevel` flag for zqd which we don't have. I'm going to do this in a follow up pr.

mattnibs:
Running one more system test on the brim side for good measure, then merging.

Issue #1022, opened 2020-07-21
AST transformation for distributed query execution


henridf:
Write an AST transformer that walks a flowgraph from a root and splits into `N` branches (where `N` is an input), putting the split point as downstream as possible. Here's an example represented in ZQL text rather than AST:

`<filter> | put <assignments> | cut <fields> | count() by <keys> | sort count` can be split into

```
<filter> | put <assignments> | cut <fields> | count() -p by <keys>   \
<filter> | put <assignments> | cut <fields> | count() -p by <keys>    \
..                                                                                                                   |  count() -c by <keys> | sort count
..                                                                                                                  /
<filter> | put <assignments> | cut <fields> | count() -p by <keys>    /
```

The `-p` flags indicate that the left-hand aggregators are to emit partial results, while `-c` flag indicates that the right hand aggregator is to consume partial results (in both cases using the `reducer.Decomposable` interface). These will be added to the `ast.GroupByProc` node as part of this issue.

part of #1004

PR #1074, opened 2020-08-08
add parallel scanning

### LINKED ISSUE #1040: Pass paths to scanners

alfred-landrum:
This adds code into the driver package to support parallel scanning and reading of sources, like zar archives, whose data is partitioned into multiple locations, possibly with some ordering defined on the partitions.

I've modified the existing `driver.Compile` to call the new code, but since `driver.Compile` expects a single `Reader`, that won't introduce any real parallelism. In a separate PR, I've updated `zar zq` to use the new `driver.MultiCompile` call, but that was large & raised some cli UX questions worth a separate discussion, see #1075 .

There's some overlap between this & the current state of https://github.com/brimsec/zq/pull/1063 ; @nwt and I were going to get our drafts up & then discuss how to land them both. I've already copied in some bits of that PR and expect I'll need more edits, as I think @nwt 's PR should land first.

Closes #1040


scanner/scanner.go
```diff
@@ -14,7 +13,7 @@ import (
 // ScannerAble is implemented by zbuf.Readers that provide an optimized
 // implementation of the Scanner interface.
 type ScannerAble interface {
-	NewScanner(ctx context.Context, filterExpr ast.BooleanExpr, s nano.Span) (Scanner, error)
+	NewScanner(ctx context.Context, filter filter.Filter, span nano.Span) Scanner
```
nwt, line 16: I'm planning to extract the Boyer-Moore search pattern from this ast.BooleanExpr in zngio.Reader.NewScanner, so I think this needs to stay as it was.

driver/multi.go
```diff
@@ -0,0 +1,169 @@
...
+		mcfg.Logger = zap.NewNop()
+	}
+	if mcfg.Span.Dur == 0 {
+		mcfg.Span = nano.MaxSpan
+	}
+	if mcfg.Warnings == nil {
+		mcfg.Warnings = make(chan string, 5)
+	}
+	if mcfg.Parallelism == 0 {
+		mcfg.Parallelism = runtime.NumCPU()
```
nwt, line 81:
```suggestion
		mcfg.Parallelism = runtime.GOMAXPROX(0)
```

scanner/scanner.go
alfred-landrum: In e47a3366a3e5078c82e9ee1cae89f20e038717d8, I adjusted this to take both the compiled filter & the filter expression. I already had both in the `driver.SearchFilter` struct, as we'll eventually want to use the filter expression to prune data in the `SendSources` call. Let me know what you think.

nwt: I want to keep the second return parameter so I can return an error after examining the ast.BooleanExpr.

And given that, I think the interface is nicer with just the ast.BooleanExpr. For example, with just that you can't accidentally pass mismatched parameters.

zbuf/combiner.go
```diff
@@ -55,9 +56,12 @@ type combiner struct {
 	hol     []*zng.Record
 	done    []bool
 	less    RecordCmpFn
+	mu      sync.Mutex
```
alfred-landrum, line 59: A Combiner reads from its done slice during reads, and writes to its done slice during close. That can lead to races in a use case like:
```go
rc, err := NewCombine(...)
defer rc.Close()

err = driver.Compile(...)
err = driver.Run(...)
```
Because `driver.Run` now uses the parallel scanner, the `Read()` on the Combiner will happen in a different goroutine than the calling goroutine, where `Close` is called in the defer. This is called out by the race checker when running unit tests.

Combiner is seemingly the only zbuf.ReadCloser that has this property, and using the mutex here seems like the smallest option, but would be glad to hear other suggestions.

scanner/scanner.go
alfred-landrum: How about if we create a new struct that has both the filter expression & the filter, like:
```go
struct CompiledFilter {
Expr ast.BooleanExp
Matcher filter.Filter
}
```

and then `NewScanner` would take that? I'd like to avoid needing to re-compile the same filter expression for every file in an archive.

nwt: I prefer separate arguments to a new structure.

scanner.ScannerAble.NewScanner and scanner.NewScanner don't need to have the same signature, and the latter's the one that matters for this PR, so how about leaving scanner.ScannerAble.NewScanner unchanged and changing scanner.NewScanner to this (with a second return parameter for an error)?
```go
func NewScanner(ctx context.Context, r zbuf.Reader, filter filter.Filter, filterExpr ast.BooleanExpr, s nano.Span) (Scanner, error) {
```

nwt: Oh, idea that doesn't make any sense.

nwt: My first preference is to leave scanner.ScannerAble.NewScanner and scanner.NewScanner as they were. The interfaces are simpler, compiling the filter doesn't strike me as too expensive to do once per file, and if proves wrong we can always add the filter.Filter parameter later.

My second preference is to change both to have this signature.
```go
NewScanner(ctx context.Context, r zbuf.Reader, f filter.Filter, filterExpr ast.BooleanExpr, s nano.Span) (Scanner, error)
```

driver/multi.go
```diff
@@ -0,0 +1,165 @@
...
+		scanner = &namedScanner{scanner, stringer.String()}
+	}
+	f := func() (ScannerCloser, error) {
+		return &noCloseScanner{scanner}, nil
+	}
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case c <- f:
+	}
```
nwt, line 132:
```suggestion
	case c <- f:
	case <-ctx.Done():
		return ctx.Err()
	}
```

driver/multi.go
```diff
@@ -0,0 +1,165 @@
...
+	if mcfg.Parallelism == 0 {
+		mcfg.Parallelism = runtime.GOMAXPROCS(0)
+	}
+
+	pctx := &proc.Context{
+		Context:     ctx,
+		TypeContext: zctx,
+		Logger:      mcfg.Logger,
+		Warnings:    mcfg.Warnings,
+	}
```
nwt, line 89: Nit: Declare this closer to its use.

driver/parallel_test.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	return nil
+}
+
+func trim(s string) string {
+	return strings.TrimSpace(s) + "\n"
+}
+
+func TestParallelOrder(t *testing.T) {
+	zctx := resolver.NewContext()
```
nwt, line 72: NIt: Declare this closer to its use.

driver/parallel_test.go
```diff
@@ -0,0 +1,97 @@
...
+	// Use `v!=3` to trigger & verify empty rank handling in orderedWaiter.
+	query, err := zql.ParseProc("v!=3")
+	require.NoError(t, err)
+
+	var buf bytes.Buffer
+	w := tzngio.NewWriter(&buf)
+	d := NewCLI(w)
```
nwt, line 79: NIt: Declare this closer to its use.

driver/parallel_test.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	for i := len(parallelTestInputs) - 1; i >= 0; i-- {
+		close(releaseChs[i])
+	}
+	return nil
+}
+
+func trim(s string) string {
+	return strings.TrimSpace(s) + "\n"
+}
```
nwt, line 69: This is pkg/test.Trim.

driver/parallel_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+func TestParallelOrder(t *testing.T) {
```
nwt, line 71: Maybe call t.Parallel?

zbuf/combiner.go
nwt, line 59: Nits: Put the mutex immediately before done, and either add a `// protects done` comment on the same line or change the name to doneMu.

driver/parallel.go
```diff
@@ -0,0 +1,483 @@
...
+	var err error
+	for i := 0; i < cap(errors); i++ {
+		e := <-errors
+		if err == nil {
+			err = e
+		}
+	}
+
+	pg.scanErr = err
+	close(pg.scanCh)
```
nwt, line 77: Maybe use golang.org/x/sync/errgroup?
```go
	rsChan := make(chan rankSourceOpener)
	var g errgroup.Group
	g.Go(func() error {
		defer close(rsChan)
		return pg.recvSources(rsChan)
	})
	for i := range pg.heads {
		i := i
		g.Go(func() error {
			return pg.scanSources(i, rsChan)
		})
	}
	pg.scanErr = g.Wait()
	close(pg.scanCh)
```

driver/heap.go
```diff
@@ -0,0 +1,33 @@
...
+// https://github.com/golang/go/blob/go1.14.5/LICENSE
+//
+// Copyright 2012 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package driver
+
+// An IntHeap is a min-heap of ints.
+type IntHeap []int
```
mattnibs, line 15: Consider making this private? I'm not sure this makes sense as a public type in package driver.

driver/compile.go
```diff
@@ -1,7 +1,6 @@
 package driver
```
henridf, line 1: This file no longer has any compile()'s - maybe it should be renamed to something like transforms.go

driver/parallel.go
```diff
@@ -0,0 +1,477 @@
...
+
+type scanSplitResult struct {
+	filterExpr ast.BooleanExpr
+	pre        ast.Proc
+	post       ast.Proc
+}
+
+func scanSplit(program ast.Proc) (scanSplitResult, error) {
+	f, p := liftFilter(program)
+	// todo
```
henridf, line 410: Is my understanding correct that the "todo" here roughly consists of doing what is done by https://github.com/brimsec/zq/commit/7fec0dffdd9c1d224af4f336bf538a7bd2d14490#diff-3b3ea8373a3aa8fc06a84639b6f90779R414 ?

If so, I thought that rather than intermingle this with scanner creation, we would have an AST transform step that, given a parameter N (threads) and information on the source ordering, transforms a linear flowgraph into a parallel one (where possible). This is similar to what we did for spillable groupby. 

(I may not be aware of some constraints and am still digesting this PR..).

driver/parallel_test.go
alfred-landrum: The `test` package depends on `driver`, so it can't be imported here.

driver/parallel_test.go
alfred-landrum: That'd be ok since this test would work in parallel with any others, but was there a specific reason you thought it should?

scanner/scanner.go
alfred-landrum: added the new signature in 17f84c5

driver/parallel.go
alfred-landrum: That's correct about the todo, and how that works & the setup for the parallels lines is something that can change in a subsequent PR. In an earlier revision of this branch, I did create AST nodes for the scanning & merging nodes. (Those have turned into `parallelHead` and `sgWait` in the current revision). That was while I was still working out how to handle the ordering dependencies, and to make it a little easier on myself, I collapsed everything into `createParallelScanner`.

driver/compile.go
alfred-landrum: I put the `compile` function back here in c191813 ; having a separate multisource compile function made more sense before the `Run`/`Compile` combining.

driver/parallel.go
henridf: This sounds good. I think that I may indeed propose additional changes in a subsequent PR, but given the various threads of work that need to land, I agree that merging this as-is is the practical way to go.

driver/multi.go
```diff
@@ -0,0 +1,143 @@
...
+)
+
+// A MultiSource is a set of one or more ZNG record sources, which could be
+// a zar archive, a local directory, a collection of remote objects, etc.
+//
+// If the sources are ordered, and that order is expected to be maintained
+// on order-dependent query operations, it must report that ordering
+// via the OrderInfo call, and send SourceOpeners in order.
+type MultiSource interface {
+	// OrderInfo reports if the records for individual sources have an
```
henridf, line 26: I think it would be helpful to future readers to distinguish between inter-source vs intra-source ordering, and which are required (both I believe) to be present in order to return a non-blank `field`.

driver/multi.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/brimsec/zq/scanner"
+	"github.com/brimsec/zq/zbuf"
+	"github.com/brimsec/zq/zng/resolver"
+	"go.uber.org/zap"
+)
+
+// A MultiSource is a set of one or more ZNG record sources, which could be
+// a zar archive, a local directory, a collection of remote objects, etc.
+//
+// If the sources are ordered, and that order is expected to be maintained
```
henridf, line 22: > and that order is expected to be maintained on order-dependent query operations

Just going by the comment here (not the implementation), I'm not entirely sure I get why what OrderInfo returns should be affected by the query. In other words, it seems like `OrderInfo()`'s results are a property of the sources, irrespective of the query.

driver/parallel_test.go
nwt: More goroutines might expose concurrency bugs unseen with fewer.

scanner/scanner.go
```diff
@@ -14,7 +14,7 @@ import (
 // ScannerAble is implemented by zbuf.Readers that provide an optimized
 // implementation of the Scanner interface.
 type ScannerAble interface {
-	NewScanner(ctx context.Context, filterExpr ast.BooleanExpr, s nano.Span) (Scanner, error)
+	NewScanner(ctx context.Context, f filter.Filter, expr ast.BooleanExpr, s nano.Span) (Scanner, error)
```
nwt, line 17: s/expr/filterExpr/ everywhere because it isn't obvious that expr ast.BooleanExpr is the filter.

zbuf/combiner.go
```diff
@@ -51,13 +52,16 @@ func NewCombiner(readers []Reader, cmp RecordCmpFn) ReadCloser {
 }
 
 type combiner struct {
-	readers []Reader
-	hol     []*zng.Record
 	done    []bool
+	doneMu  sync.Mutex // protects done
```
nwt, line 56: Nit: Declare the mutex before the field it protects.

alfred-landrum:
@henridf & I had a good discussion this morning about the PR; I'm going to attempt some changes that should both simplify & make this easier for his upcoming AST work.

alfred-landrum:
@henridf and I discussed, and think the structure is at a decent point to go in. I added comments in parallel.go that reflect the manual setup of the parallel scanner is temporary, and will be replaced by @henridf 's upcoming work for https://github.com/brimsec/zq/issues/1022 . This is good enough now that we can start adjusting `zar zq` and zqd to take advantage of the multisource & parallel scanning.

Note that this now depends on the ordered merge in https://github.com/brimsec/zq/pull/1104.

scanner/scanner.go
```diff
@@ -14,13 +14,18 @@ import (
 // ScannerAble is implemented by zbuf.Readers that provide an optimized
 // implementation of the Scanner interface.
 type ScannerAble interface {
-	NewScanner(ctx context.Context, filterExpr ast.BooleanExpr, s nano.Span) (Scanner, error)
+	NewScanner(ctx context.Context, f filter.Filter, filterExpr ast.BooleanExpr, s nano.Span) (Scanner, error)
+}
+
+// A ScannerStatsAble generates scanner statistics.
+type ScannerStatsAble interface {
+	Stats() *ScannerStats
```
nwt, line 22: This is an odd name. Statter is more conventional.

driver/heap.go
```diff
@@ -0,0 +1,33 @@
+// This file contains code from these Go 1.14.5 source files:
+// - container/heap/example_intheap_test.go
+//
+// That code is covered by the following copyright notice,
+// and the license file mentioned is available at:
+// https://github.com/golang/go/blob/go1.14.5/LICENSE
```
nwt, line 6: Is a reference to the license file sufficient? The file itself says:
>Redistributions of source code must retain the above copyright
>notice, this list of conditions and the following disclaimer.
And referring doesn't seem like quite the same thing as retaining.

scanner/scanner.go
nwt: Bah. The method is Stats, not Stat, so make that Statser.

alfred-landrum: step by step, you are working out how I came to 'ScannerStatsAble'. :)

henridf:
> @henridf and I discussed, and think the structure is at a decent point to go in. 

For context for reviewers, here are the two main points that we  agreed were worth pursuing in follow-on PRs (that I will work on after this goes in):

-  In `createParallelGroup()`, the parameters that are used to inform merging order (sortField, reversed) need to come from the analysis pass rather than from `msrc.OrderInfo()`. One way would be to have `pscanAnalyze()` store them in `pgSetup{}`, which is then passed to `createParallelGroup()`.
Another more declarative way would be to have this sort info added by distributed transform AST pass to the `ast.ParallelProc{}` node. Then, `proc.CompileProc()` can wire up a merge point differently depending on this info. If the sort info is absent, then use our existing un-ordered merger. If the sort info is present, then use the ordered merge proc.
I think this latter way is better because it keeps AST analysis and flowgraph compilation separate from  scanner setup. Also, I suspect we'll eventually have other forms of analysis/transform that also care about deriving/indicating sortedness at merge points, and this enables that to be done and "just work".

- I think one of the things preventing further simplification is that `proc.CompileProc()` takes a single parent Proc argument. In regular single-scanner setup, `driver.Compile()` passed it's single source proc into that argument. Now that we have multiple parallelHeads, that doesn't work. But if instead of a single source, `driver.Compile()` could pass a slice `parents []Proc` into `proc.CompileProc()`, then it just needs to create N parallelHeads and pass them into compilation, like in the current model.

driver/parallel.go
```diff
@@ -0,0 +1,204 @@
...
+		cmp, err := newCompareFn(sortField, reversed)
+		if err != nil {
+			return nil, nil, err
+		}
+		merger = proc.NewOrderedMerge(pctx, chains, cmp)
+	} else {
+		merger = proc.NewMerge(pctx, chains)
+	}
+
+	return merger, pg, nil
```
nwt, line 203: Maybe shorten?
```suggestion
	if sortField == "" {
		return proc.NewMerge(pctx, chains), pg, nil
	}
	cmp, err := newCompareFn(sortField, reversed)
	if err != nil {
		return nil, nil, err
	}
	return proc.NewOrderedMerge(pctx, chains, cmp), pg, nil
```

driver/heap.go
```diff
@@ -0,0 +1,32 @@
...
+// LICENSE file in the root directory of this repository.
+
+// Copyright (c) 2009 The Go Authors. All rights reserved.
+// See acknowledgments.txt for full license text from:
+// https://github.com/golang/go/blob/master/LICENSE
+
+package driver
+
+// An intHeap is a min-heap of ints.
+type intHeap []int
```
nwt, line 14: I don't see any use of this type. Can the file go away?

driver/parallel.go
```diff
@@ -0,0 +1,222 @@
...
+func (ph *parallelHead) closeOnDone() {
+	select {
+	case <-ph.Context.Done():
+		ph.mu.Lock()
+		defer ph.mu.Unlock()
+		if ph.sc != nil {
+			ph.sc.Close()
+			ph.sc = nil
+		}
+	}
```
nwt, line 33: Nits: No need for select or defer.
```suggestion
	<-ph.Context.Done()
	ph.mu.Lock()
	if ph.sc != nil {
		ph.sc.Close()
		ph.sc = nil
	}
	ph.mu.Unlock()
```

driver/parallel.go
```diff
@@ -0,0 +1,222 @@
...
+
+func (ph *parallelHead) Pull() (zbuf.Batch, error) {
+	// Trigger the parallel group to read from the multisource.
+	ph.pg.once.Do(func() {
+		go ph.pg.run()
+	})
+	// Ensure open scanners are closed when flowgraph execution stops.
+	ph.once.Do(func() {
+		go ph.closeOnDone()
+	})
```
nwt, line 44: Nit: No need for parallelGroup.once.
```suggestion
	ph.once.Do(func() {
		go ph.closeOnDone()
		go ph.pg.run()
	})
```

driver/parallel.go
```diff
@@ -0,0 +1,222 @@
...
+	s := pg.stats
+	for _, sc := range pg.scanners {
+		s.Accumulate(sc.Stats())
+	}
+	return &s
+}
+
+func (pg *parallelGroup) run() {
+	defer close(pg.sourceChan)
+	pg.sourceErr = pg.msrc.SendSources(pg.pctx, pg.pctx.TypeContext, pg.filter, pg.sourceChan)
```
nwt, line 135: Nit: No need for defer.
```suggestion
	pg.sourceErr = pg.msrc.SendSources(pg.pctx, pg.pctx.TypeContext, pg.filter, pg.sourceChan)
	close(pg.sourceChan)
```

driver/parallel.go
```diff
@@ -0,0 +1,222 @@
...
+	filter     SourceFilter
+	msrc       MultiSource
+	once       sync.Once
+	pctx       *proc.Context
+	sourceChan chan SourceOpener
+	sourceErr  error
+
+	mu       sync.Mutex // protects below
+	stats    scanner.ScannerStats
+	scanners []scanner.Scanner
```
nwt, line 83: Nit: You could simplify the methods a little if you changed the type of scanners to map[scanner.Scanner]struct{}.

alfred-landrum: oh, that's a good suggestion!

driver/parallel.go
alfred-landrum: pg.once exists because pg.run needs to run whenever the first (of possibly many) parallel heads get their Pull.

PR #1075, opened 2020-08-08
make zar zq use the parallel scanner

alfred-landrum:
Adjust `zar zq` to use the parallel scanner from https://github.com/brimsec/zq/pull/1074 . This took a `zar zq` operation on my mac to find a single record in a 4.5GB zar archive from ~70 seconds to ~14 seconds.

However, this raised an interesting question about the intended usage of `zar zq`. In its pre-existing form, it expects to run sequentially on each zar directory, processing the zng file for that zar directory, and sending its output to either 1: an aggregate (like stdout), or 2: to a new file inside the zar directory of the file it just processed. 

The parallel scanning (as implemented in https://github.com/brimsec/zq/pull/1074 ) only makes sense to use with behavior 1. I've keyed the different behaviors off the `-o` output flag in the current state, but that was intended as a proof of concept. I think it makes sense to have different zar subcommands for these behaviors, and was thinking of keeping behavior 1 as `zar zq`, and behavior 2 as `zar map`.

### DIFF
--- a/archive/walk.go
+++ b/archive/walk.go
@@ -1,9 +1,16 @@
 package archive
 
 import (
+	"context"
+	"io"
 	"strings"
 
+	"github.com/brimsec/zq/driver"
 	"github.com/brimsec/zq/pkg/iosrc"
+	"github.com/brimsec/zq/scanner"
+	"github.com/brimsec/zq/zbuf"
+	"github.com/brimsec/zq/zio/detector"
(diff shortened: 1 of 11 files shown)


alfred-landrum:
Closing this draft; work moved (with less history noise) to https://github.com/brimsec/zq/pull/1138 .

PR #1080, opened 2020-08-11
Add Boyer-Moore filter optimization to ZNG scanner

nwt:
This implements steps 6 and 7 of #681.

Attempt to improve filter performance for the ZNG scanner performance
with a Boyer-Moore string search over compressed value message blocks.

The Boyer-Moore string search algorithm itself is implemented in
zio/zngio/search.go, a copy of Go's src/strings/search.go with minimal
modifications.

### DIFF
--- a/zio/zngio/scanner.go
+++ b/zio/zngio/scanner.go
@@ -7,7 +7,9 @@ import (
 	"io"
 	"sync/atomic"
 
+	"github.com/brimsec/zq/ast"
 	"github.com/brimsec/zq/filter"
+	"github.com/brimsec/zq/pkg/byteconv"
 	"github.com/brimsec/zq/pkg/nano"
 	"github.com/brimsec/zq/scanner"
 	"github.com/brimsec/zq/zbuf"
@@ -16,11 +18,83 @@ import (
 
 // zngScanner implements scanner.Scanner.
 type zngScanner struct {
-	ctx    context.Context
-	reader *Reader
-	filter filter.Filter
-	span   nano.Span
-	stats  scanner.ScannerStats
+	ctx          context.Context
+	reader       *Reader
+	filter       filter.Filter
+	span         nano.Span
+	stats        scanner.ScannerStats
+	stringFinder *stringFinder
+}
+
+var _ scanner.ScannerAble = (*Reader)(nil)
+
+func (r *Reader) NewScanner(ctx context.Context, filterExpr ast.BooleanExpr, s nano.Span) (scanner.Scanner, error) {
+	var f filter.Filter
+	var sf *stringFinder
+	if filterExpr != nil {
+		var err error
+		if f, err = filter.Compile(filterExpr); err != nil {
+			return nil, err
+		}
+		p, err := requiredPattern(filterExpr)
+		if err != nil {
+			return nil, err
+		}
+		if len(p) > 1 {
+			// Profitable if pattern is more than one bytes.
+			sf = makeStringFinder(p)
+		}
+	}
+	return &zngScanner{ctx: ctx, reader: r, filter: f, span: s, stringFinder: sf}, nil
+}
+
+// requiredPattern searches e for a required pattern of maximum length. A
+// required pattern for e is a byte sequence that must be present in any buffer
+// that contains the ZNG encoding of a record matching e.
+func requiredPattern(e ast.BooleanExpr) (string, error) {
+	switch e := e.(type) {
+	case *ast.LogicalAnd:
+		left, err := requiredPattern(e.Left)
+		if err != nil {
+			return "", err
+		}
+		right, err := requiredPattern(e.Right)
+		if err != nil {
+			return "", err
+		}
+		if len(right) > len(left) {
+			return right, nil
+		}
+		return left, nil
+	case *ast.CompareAny:
+		return requiredPatternForCompare(e.Comparator, e.Value)
+	case *ast.CompareField:
+		return requiredPatternForCompare(e.Comparator, e.Value)
+	default:
+		return "", nil
+	}
+}
+
+// requiredPatternForCompare tries to determine a required pattern for an
+// ast.CompareAny or ast.CompareField with comparator and value. If it cannot,
+// it returns the empty string.
+func requiredPatternForCompare(comparator string, value ast.Literal) (string, error) {
+	if comparator != "=" && comparator != "in" {
+		return "", nil
+	}
+	if value.Type == "regexp" {
+		return "", nil
+	}
+	if value.Type == "string" {
+		value = ast.Literal{Node: ast.Node{Op: "Literal"}, Type: "bstring", Value: value.Value} // xxx
+	}
+	v, err := zng.Parse(value)
+	if err != nil {
+		return "", err
+	}
+	// We're looking for a complete ZNG value, so we can lengthen the
+	// pattern by calling Encode to add a tag.
+	return string(v.Encode(nil)), err
 }
 
 // Pull implements scanner.Scanner.Pull.
@@ -59,6 +133,16 @@ func (s *zngScanner) Pull() (zbuf.Batch, error) {
 }
 
 func (s *zngScanner) scanUncompressed() ([]*zng.Record, error) {
+	if s.stringFinder != nil {
+		buf := s.reader.uncompressed.Bytes()
+		if s.stringFinder.next(byteconv.UnsafeString(buf)) == -1 {
+			// We know s.reader.uncompressed cannot contain any
+			// records matching s.filter.
+			s.reader.uncompressed = nil
+			// xxx stats
+			return nil, nil
+		}
+	}
 	var recs []*zng.Record
 	for uncompressed := s.reader.uncompressed; uncompressed.Len() > 0; {
 		id, err := readUvarint7(uncompressed)
--- a/zio/zngio/reader.go
+++ b/zio/zngio/reader.go
@@ -2,17 +2,12 @@ package zngio
 
 import (
 	"bytes"
-	"context"
 	"encoding/binary"
 	"errors"
 	"fmt"
 	"io"
 
-	"github.com/brimsec/zq/ast"
-	"github.com/brimsec/zq/filter"
-	"github.com/brimsec/zq/pkg/nano"
 	"github.com/brimsec/zq/pkg/peeker"
-	"github.com/brimsec/zq/scanner"
 	"github.com/brimsec/zq/zng"
 	"github.com/brimsec/zq/zng/resolver"
 	"github.com/pierrec/lz4/v4"
@@ -403,17 +398,4 @@ func (r *Reader) parseValue(id int, b []byte) (*zng.Record, error) {
 		return nil, err
 	}
 	return rec, nil
-}
-
-var _ scanner.ScannerAble = (*Reader)(nil)
-
-func (r *Reader) NewScanner(ctx context.Context, filterExpr ast.BooleanExpr, s nano.Span) (scanner.Scanner, error) {
-	var f filter.Filter
-	if filterExpr != nil {
-		var err error
-		if f, err = filter.Compile(filterExpr); err != nil {
-			return nil, err
-		}
-	}
-	return &zngScanner{ctx: ctx, reader: r, filter: f, span: s}, nil
 }
--- a/zio/zngio/search.go
+++ b/zio/zngio/search.go
@@ -1,0 +1,126 @@
+// Copyright 2012 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package zngio
+
+import "strings"
+
+// stringFinder efficiently finds strings in a source text. It's implemented
+// using the Boyer-Moore string search algorithm:
+// https://en.wikipedia.org/wiki/Boyer-Moore_string_search_algorithm
+// https://www.cs.utexas.edu/~moore/publications/fstrpos.pdf (note: this aged
+// document uses 1-based indexing)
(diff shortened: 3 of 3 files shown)


zio/zngio/scanner.go
```diff
@@ -16,11 +18,83 @@ import (
...
+			sf = makeStringFinder(p)
+		}
+	}
+	return &zngScanner{ctx: ctx, reader: r, filter: f, span: s, stringFinder: sf}, nil
+}
+
+// requiredPattern searches e for a required pattern of maximum length. A
+// required pattern for e is a byte sequence that must be present in any buffer
+// that contains the ZNG encoding of a record matching e.
+func requiredPattern(e ast.BooleanExpr) (string, error) {
```
nwt, line 54: I'm working on a test for this.

zio/zngio/scanner.go
```diff
@@ -59,6 +133,16 @@ func (s *zngScanner) Pull() (zbuf.Batch, error) {
 }
 
 func (s *zngScanner) scanUncompressed() ([]*zng.Record, error) {
+	if s.stringFinder != nil {
+		buf := s.reader.uncompressed.Bytes()
+		if s.stringFinder.next(byteconv.UnsafeString(buf)) == -1 {
+			// We know s.reader.uncompressed cannot contain any
+			// records matching s.filter.
+			s.reader.uncompressed = nil
+			// xxx stats
```
nwt, line 142: Efficiently updating scanner.ScannerStats.RecordsRead probably requires that we change the ZNG spec so that a compressed value message block stores the number of records it contains. Do we need to do that?

zio/zngio/scanner.go
```diff
@@ -16,11 +18,83 @@ import (
...
+func requiredPatternForCompare(comparator string, value ast.Literal) (string, error) {
+	if comparator != "=" && comparator != "in" {
+		return "", nil
+	}
+	if value.Type == "regexp" {
+		return "", nil
+	}
+	if value.Type == "string" {
+		value = ast.Literal{Node: ast.Node{Op: "Literal"}, Type: "bstring", Value: value.Value} // xxx
+	}
```
nwt, line 90: This was copied from zng.ParseLiteral. I don't like it but don't see an easy way to get us down to one copy or, better yet, get rid of it entirely.

zio/zngio/scanner.go
```diff
@@ -16,11 +18,83 @@ import (
...
+			return right, nil
+		}
+		return left, nil
+	case *ast.CompareAny:
+		return requiredPatternForCompare(e.Comparator, e.Value)
+	case *ast.CompareField:
+		return requiredPatternForCompare(e.Comparator, e.Value)
+	default:
+		return "", nil
+	}
```
nwt, line 75: ast.Search semantics make it tough to handle. We probably need to allow for two patterns, only one of which must match. And how to handle the case folding in filter.searchString is unclear.

zio/zngio/scanner.go
```diff
@@ -16,11 +18,83 @@ import (
...
-	stats  scanner.ScannerStats
+	ctx          context.Context
+	reader       *Reader
+	filter       filter.Filter
+	span         nano.Span
+	stats        scanner.ScannerStats
+	stringFinder *stringFinder
+}
+
+var _ scanner.ScannerAble = (*Reader)(nil)
```
henridf, line 29: 👍 I've wondered why we don't use  more of this pattern as opposed to writing "x implements the y interface".  Type checking >> comments.

zio/zngio/scanner.go
henridf: This seems worthy of being added as comment in the .go file.

zio/zngio/scanner.go
```diff
@@ -16,11 +18,83 @@ import (
...
+	case *ast.LogicalAnd:
+		left, err := requiredPattern(e.Left)
+		if err != nil {
+			return "", err
+		}
+		right, err := requiredPattern(e.Right)
+		if err != nil {
+			return "", err
+		}
+		if len(right) > len(left) {
```
henridf, line 65: Why not search for both `left` and `right` when both are "profitable" per your definition above (len > 1)?

zio/zngio/scanner.go
alfred-landrum: Since we BytesRead & RecordsRead, we could just interpret RecordsRead as "instantiated records needed for this operation". In which case, we wouldn't need to update RecordsRead here.

nwt: Works for me.

nwt, line 142: Another issue here is that the obvious way to update scanner.ScannerStats.BytesRead is to add len(buf) to it, but that will include value message header bytes (the uvarint7 and uvarint containing the type ID and the value length), which aren't included by scanner.scanner or by zngScanner.scanOne.

In other words, scanner.ScannerStats.BytesRead can vary with the filter. That feels wrong, but so does the obvious fix (changing the ZNG spec so that a compressed value mesasge block stores the number of non-header bytes it contains).

zio/zngio/scanner.go
nwt: It was more work, and I wanted to create this PR before doing it. 😃

zio/zngio/scanner.go
alfred-landrum: This is the header you're referring to, correct?
https://github.com/brimsec/zq/blob/a1a817e0a3d2e1781807df51533bd7794d97d8f7/zio/zngio/writer.go#L158

If so, I think it's valid to include that in BytesRead, since it reflects bytes that needed to be read to pull data.

nwt: No, I'm referring to the header on each value message. It's added here.
https://github.com/brimsec/zq/blob/a1a817e0a3d2e1781807df51533bd7794d97d8f7/zio/zngio/writer.go#L82-L90

alfred-landrum:
thanks. In that case, should we perhaps add header value length to
BytesRead for zngScanner.scanOne calls, and do so here as well? Would that
make scanner stats at least consistent between any zng based sources?

zio/zngio/scanner.go
henridf: Makes sense!

zio/zngio/scanner.go
mattnibs: I am probably missing something, but why can't BytesRead just be derived from a count of bytes Read by the underlying io.Reader (probably by wrapping it with a counter)?

mattnibs: I think we are going to want to have records read reflect the number of records read from disk. The number of instantiated records is a gray area between records read and records match. I am skeptical of its utility at that point. This would mean I am in favor of storing the number of records in a compressed block.

Could be done down the road, but for now I would just assume drop RecordsRead from scanner stats for the time being if not done in this pr.

zio/zngio/scanner.go
nwt: Done in 2167375125364dfb59629b9df425e0bbd84d7cc6.

filter/bufferfilter.go
```diff
@@ -0,0 +1,147 @@
...
+			// TODO Match the behavior of searchRecordString.
+			return nil, nil
+		}
+		for _, r := range e.Text {
+			// TODO Make stringFinder insensitive to case. Until
+			// then, bail if e.Text contains a letter.
+			if unicode.IsLetter(r) {
+				return nil, nil
+			}
+		}
```
nwt, line 88: I'll create an issue for this work.

nwt: #1128 tracks it.

zio/zngio/scanner.go
nwt: > why can't BytesRead just be derived from a count of bytes Read by the underlying io.Reader (probably by wrapping it with a counter)?

It certainly could. The catch is that that would change behavior, though I don't know if anyone cares. (Right now, a given query against a given set of records will produce the same stats regardless of whether those records are stored as Zeek TSV, JSON, TZNG, or ZNG. That would change.)

zio/zngio/scanner.go
nwt: Done. #1128 tracks the work.

zio/zngio/scanner.go
nwt: I'd like to move forward with what's here (a skipped compressed message value block doesn't affect RecordsRead) and revisit the question for #1010.

nwt: I'd like to move forward with what's here (BytesRead includes value message headers for skipped compressed message value blocks only) and revisit the question for #1010.

filter/bufferfilter.go
```diff
@@ -0,0 +1,147 @@
...
+	default:
+		panic(fmt.Sprintf("BufferFilter: unknown type %T", e))
+	}
+}
+
+func newBufferFilterForLiteral(l ast.Literal) (*BufferFilter, error) {
+	switch l.Type {
+	case "bool", "byte", "int16", "uint16", "int32", "uint32", "int64", "uint64", "float64", "time", "duration":
+		// These are all comparable, so they can require up to three
+		// patterns: float, varint, and uvarint.
```
nwt, line 107: @alfred-landrum: Is this worth an issue?

filter/bufferfilter.go
```diff
@@ -0,0 +1,147 @@
...
+			return nil, err
+		}
+		return &BufferFilter{op: opOr, left: left, right: right}, nil
+	default:
+		panic(fmt.Sprintf("BufferFilter: unknown type %T", e))
+	}
+}
+
+func newBufferFilterForLiteral(l ast.Literal) (*BufferFilter, error) {
+	switch l.Type {
```
mattnibs, line 104: Do we want a `default` in this switch?

zio/zngio/scanner.go
henridf: nice. (and i agree that bufferFilter is a better name).

filter/bufferfilter.go
```diff
@@ -0,0 +1,147 @@
...
+			op:           opStringFinder,
+			stringFinder: makeStringFinder(e.Text),
+		}
+		right, err := newBufferFilterForLiteral(e.Value)
+		if right == nil || err != nil {
+			return nil, err
+		}
+		return &BufferFilter{op: opOr, left: left, right: right}, nil
+	default:
+		panic(fmt.Sprintf("BufferFilter: unknown type %T", e))
```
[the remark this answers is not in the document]
henridf: (I'm not @nwt but i think a panic makes sense as it would indicate a new expression node was added to the AST and needs to be case'd here.)

henridf (approved):
👍

filter/bufferfilter.go
nwt: @mattnibs: I don't think so since it would be empty, but perhaps you're seeing something I'm not.

filter/bufferfilter.go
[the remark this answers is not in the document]
nwt: filter.Compile returns an error in this situation, so make this consistent with that.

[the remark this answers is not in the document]
nwt: Done in 4c56cd13ae5ea8acfecc2e6f24d9ae8a9af75421.

filter/filter_test.go
```diff
@@ -92,6 +93,21 @@ func runCases(t *testing.T, record *zng.Record, cases []testcase) {
...
 			err := runTest(tt.filter, record, tt.expectedResult)
 			require.NoError(t, err)
+
+			proc, err := zql.ParseProc(tt.filter)
+			require.NoError(t, err, "filter: %q", tt.filter)
+
+			filterExpr := proc.(*ast.FilterProc).Filter
+
+			bf, err := filter.NewBufferFilter(filterExpr)
+			if assert.NoError(t, err, "filter: %q", tt.filter) &&
```
alfred-landrum, line 103: Is it ok for filter.NewBufferFilter to ever return an error here? I can't tell from the if statement.

nwt: It isn't. The construct is equivalent to this:
```go
assert.NoError(t, err, "filter: %q", tt.filter)
if err == nil && ... {
```
I'm happy to change it to that if you'd like.

zio/zngio/scanner.go
```diff
@@ -58,8 +61,17 @@ func (s *zngScanner) Pull() (zbuf.Batch, error) {
 }
 
 func (s *zngScanner) scanUncompressed() ([]*zng.Record, error) {
+	uncompressed := s.reader.uncompressed
+	if bf := s.bufferFilter; bf != nil && !bf.Eval(uncompressed.Bytes()) {
```
alfred-landrum, line 65:
```suggestion
	if s.bufferFilter != nil && !s.bufferFilter.Eval(uncompressed.Bytes()) {
```

filter/bufferfilter.go
alfred-landrum: yes please!

zio/zngio/scanner.go
nwt: Done in 8a59d3a0818607efc8b3ca16635d66e6ad08e5a4.

filter/bufferfilter.go
nwt: Created #1140 for this.

filter/filter_test.go
alfred-landrum: Would you please? It's just a nit, but I think it's easier to read.

nwt:
> Will #1128 add direct testing of `BufferFilter` results?

@alfred-landrum: I think so.

filter/filter_test.go
nwt: Done in b28d0b9db7672f402703fcf49259aaaa9dd82c27.

PR #1081, opened 2020-08-11
merge driver.Compile into driver.Run

alfred-landrum:
Combine the existing driver.Compile and driver.Run calls into driver.Run, and stop exporting the mux related names. I did this mostly because I wanted to add the extra `context.WithCancel` call into the new driver.Run, so that any resources needed for executing the ZQL (like the parallel scanner setup in #1074) would be released as soon as possible.

There was only one user of MuxOutput, the search api handler in zqd. It sent the stats after a successful search query. I maintained that behavior (and verified that Brim tests still work) by sending the stats at the end of a successful run, if a stats timer had been configured.

### DIFF
--- a/driver/compile.go
+++ b/driver/compile.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"strconv"
+	"time"
 
 	"github.com/brimsec/zq/ast"
 	"github.com/brimsec/zq/expr"
@@ -22,13 +23,10 @@ type Config struct {
 	ReaderSortReverse bool
 	Span              nano.Span
 	Warnings          chan string
+	StatsTick         <-chan time.Time
 }
 
-// Compile takes an AST, an input reader, and configuration parameters,
-// and compiles it into a runnable flowgraph, returning a
-// proc.MuxOutput that which brings together all of the flowgraphs
-// tails, and is ready to be Pull()'d from.
-func Compile(ctx context.Context, program ast.Proc, zctx *resolver.Context, reader zbuf.Reader, cfg Config) (*MuxOutput, error) {
+func compileMux(ctx context.Context, program ast.Proc, zctx *resolver.Context, reader zbuf.Reader, cfg Config) (*MuxOutput, error) {
 	if cfg.Logger == nil {
 		cfg.Logger = zap.NewNop()
 	}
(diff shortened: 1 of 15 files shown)


driver/compile.go
```diff
@@ -22,13 +23,10 @@ type Config struct {
 	ReaderSortReverse bool
 	Span              nano.Span
 	Warnings          chan string
+	StatsTick         <-chan time.Time
```
nwt, line 26: Nit: Put this before Warnings so these remain alphabetized.

driver/compile.go
```diff
@@ -22,13 +23,10 @@ type Config struct {
...
 }
 
-// Compile takes an AST, an input reader, and configuration parameters,
-// and compiles it into a runnable flowgraph, returning a
-// proc.MuxOutput that which brings together all of the flowgraphs
-// tails, and is ready to be Pull()'d from.
-func Compile(ctx context.Context, program ast.Proc, zctx *resolver.Context, reader zbuf.Reader, cfg Config) (*MuxOutput, error) {
+func compileMux(ctx context.Context, program ast.Proc, zctx *resolver.Context, reader zbuf.Reader, cfg Config) (*muxOutput, error) {
```
nwt, line 29: Nit: Call this compile. compileMux reads like this compiles a mux, but it doesn't.

alfred-landrum:
> The one knock I see on this is that it makes it impossible for zq to print any ZQL error before creating the output file. To my surprise, zq doesn't currently do that, but it probably should. (To be clear, right now it creates the output file first, then prints the ZQL error.)

In zq, I see a call to `ParseProc` that happens before hitting the driver code, and that triggers for me locally if I give a bad ZQL expression, without creating an output file.

python/src/zqext.go
```diff
@@ -74,13 +74,8 @@ func doZqlFileEval(inquery, inpath, informat, outpath, outformat string) (err er
...
 	}()
 
-	fg, err := driver.Compile(context.Background(), query, zctx, rc, driver.Config{})
-	if err != nil {
-		return err
-	}
 	d := driver.NewCLI(w)
-
-	return driver.Run(fg, d, nil)
+	return driver.Run(context.Background(), d, query, zctx, rc, driver.Config{})
```
nwt, line 78: You can get rid of the defer up above if you want to.

PR #1088, opened 2020-08-11
zqd listen: add -loglevel flag

mattnibs:
Add -loglevel flag to set the level of printed logs. If logging
is specified in the passed zqd config, this flag will be ignored.

### DIFF
--- a/cmd/zqd/listen/command.go
+++ b/cmd/zqd/listen/command.go
@@ -24,6 +24,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 	"gopkg.in/yaml.v3"
 )
 
@@ -66,6 +67,7 @@ type Command struct {
 	zeekRunnerPath string
 	configfile     string
 	loggerConf     *logger.Config
+	logLevel       zapcore.Level
 	logger         *zap.Logger
 	devMode        bool
 	portFile       string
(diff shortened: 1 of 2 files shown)


cmd/zqd/listen/command.go
```diff
@@ -79,6 +81,7 @@ func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
 	f.BoolVar(&c.pprof, "pprof", false, "add pprof routes to api")
 	f.BoolVar(&c.prom, "prometheus", false, "add prometheus metrics routes to api")
 	f.StringVar(&c.configfile, "config", "", "path to a zqd config file")
+	f.Var(&c.logLevel, "loglevel", "level for log output (defaults to info)")
```
nwt, line 84:
```suggestion
	f.Var(&c.logLevel, "loglevel", "level for log output")
```
because
```
$ go run ./cmd/zqd help listen |& grep loglevel
    -loglevel level for log output (defaults to info) (default "info")
```

cmd/zqd/listen/command.go
```diff
@@ -211,6 +214,9 @@ func (c *Command) initZeek() error {
 func (c *Command) initLogger() error {
 	if c.loggerConf == nil {
 		c.loggerConf = defaultLogger
+		for i := range c.loggerConf.Children {
+			c.loggerConf.Children[i].Level = c.logLevel
+		}
```
nwt, line 219: Instead of adding this loop, it might be clearer to change defaultLogger to a method.
```go
func (c *Command) defaultLoggerConfig() &logger.Config {
	return &logger.Config{
		Type: logger.TypeWaterfall,
		Children: []logger.Config{
			{
				Name:  "http.access",
				Path:  "/dev/null",
				Level: c.logLevel,
			},
			{
				Path:  "stderr",
				Level: c.logLevel,
			},
		},
	}
}
```

cmd/zqd/listen/command.go
mattnibs: cool!

cmd/zqd/listen/command.go
mattnibs: yeah, let's do this

PR #1090, opened 2020-08-12
Return nonvolatile records from zngio.zngScanner.Pull

### LINKED ISSUE #1087: zq change causing Brim error messages during pcap import
Brim commit `8c496e0` represents the advance of its `zq` dependency to a `zq` commit that includes the change for https://github.com/brimsec/zq/pull/1081. Per the attached video, importing example pcap [wrccdc.2018-03-23.010014000000000.pcap.gz](https://archive.wrccdc.org/pcaps/2018/wrccdc.2018-03-23.010014000000000.pcap.gz) now causes multiple error messages to be triggered, including "Unexpected token p in JSON at position 31" and "Malformed zng value". These error messages were not present when importing the same pcap at the Brim commit `9da5f05` that came just prior.

[Repro.zip](https://github.com/brimsec/zq/files/5059353/Repro.zip)

I also see error messages on the NPM console during import:

```
zqd2020/08/11 13:36:33 http: superfluous response.WriteHeader call from github.com/brimsec/zq/zqd.(*recordingResponseWriter).WriteHeader (middleware.go:118)
```

nwt:
zngio.zngScanner.Pull returns volatile records, but driver.mux.run sends
them to another goroutine to be read, presenting a potential data race.
Fix that by returning nonvolatile records from zngio.zngScanner.Pull.

Fixes #1087.

### DIFF
--- a/zio/zngio/scanner.go
+++ b/zio/zngio/scanner.go
@@ -78,6 +78,7 @@ func (s *zngScanner) scanUncompressed() ([]*zng.Record, error) {
 			return nil, err
 		}
 		if rec != nil {
+			rec.CopyBody()
 			recs = append(recs, rec)
 		}
 	}
--- a/zio/zngio/scanner.go
+++ b/zio/zngio/scanner.go
@@ -52,7 +52,6 @@ func (s *zngScanner) Pull() (zbuf.Batch, error) {
 			return nil, err
 		}
 		if rec != nil {
-			rec.CopyBody()
 			return zbuf.NewArray([]*zng.Record{rec}), nil
 		}
 	}
@@ -78,7 +77,6 @@ func (s *zngScanner) scanUncompressed() ([]*zng.Record, error) {
 			return nil, err
 		}
 		if rec != nil {
-			rec.CopyBody()
 			recs = append(recs, rec)
 		}
 	}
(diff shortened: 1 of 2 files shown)


zio/zngio/scanner.go
```diff
@@ -78,6 +78,7 @@ func (s *zngScanner) scanUncompressed() ([]*zng.Record, error) {
 			return nil, err
 		}
 		if rec != nil {
+			rec.CopyBody()
```
alfred-landrum, line 81: Both calls to scanOne are now followed by this call, can it move inside scanOne?

alfred-landrum:
Confirming that this fixes my cli repro case: https://github.com/brimsec/zq/issues/1087#issuecomment-672299522

alfred-landrum:
@nwt : would you please a ztest for this?

philrz:
I've also confirmed that a current `master` of Brim (commit `def9cbb`) when pointed at a `zqd` at the commit `958b7ed` (this branch) no longer has errors during ingest as originally reported. 👍 

Since this seems to prove @alfred-landrum's theory that the root cause was back when #1063 was merged, it's still a strange mystery to me why the ingest errors can't be reproduced on Brim commit `ce37016` which was when Brim's `zq` pointer was first advanced to include the change in #1063. But since the app breakage is fixed I guess I'm content not knowing. 🤷‍♂️

zio/zngio/scanner.go
nwt: Done in 0c0ffd0d07f19518a61585c514577572c269d786.

nwt:
> @nwt : would you please a ztest for this?

Done in 0c0ffd0d07f19518a61585c514577572c269d786.

PR #1094, opened 2020-08-13
update help text for zar commands

### LINKED ISSUE #1031: Provide explicit mention of S3 in "zar import" help text
When verifying #881, I mistakenly thought I was supposed to set my `$ZAR_ROOT` to an actual S3 URL, but this is incorrect. As of `zq` commit `59b4bcc`, the `zar import` help text mentions the `-data` location I was supposed to be using for my S3 URL, but the interplay with S3 was not clear from the help text alone.

```
$ zar help import
NAME
    import - import log files into pieces

USAGE
    import [options] file

OPTIONS
    -R root directory of zar archive to walk (default "/Users/phil/logs")
    -data location for storing data files (defaults to root directory)
    -empty create an archive without initial data (default "false")
    -i format of input data [auto,zng,ndjson,zeek,zjson,tzng,parquet] (default "auto")
    -s target size of chopped files, as '10MB' or '4GiB', etc. (default "500MiB")

DESCRIPTION
    The import command provides a way to create a new zar archive with
    data. It takes as input zng data and cuts the stream into chunks where
    each chunk is created when the size threshold is exceeded, either in
    bytes (-b) or megabytes (-s).  The path of each chunk is a subdirectory
    in the specified directory (-R or ZAR_ROOT) where the subdirectory name
    is derived from the timestamp of the first zng record in that chunk.
```

It seems like an additional paragraph would be justified that explains the concept of separating the archive metadata from the data storage, and calls out S3 as an example where this separation _must_ be done.

alfred-landrum:
We've added support for S3 root's since #1031 was filed, so I thought I'd add some text both for `zar import` and the higher level `zar help` that mentioned S3. I also tried to make a few other improvements, like consistent language about a "chunk" as the main log file, and the use of a "root" location instead of just directory for the `-R`/`$ZAR_ROOT` configs.

Closes #1031

### DIFF
--- a/cmd/zar/import/command.go
+++ b/cmd/zar/import/command.go
@@ -18,14 +18,16 @@ import (
 
 var Import = &charm.Spec{
 	Name:  "import",
-	Usage: "import [options] file",
+	Usage: "import [-R root] [options] [file|S3-object|-]",
 	Short: "import log files into pieces",
 	Long: `
-The import command provides a way to create a new zar archive with data.
-It takes as input zng data and cuts the stream
-into chunks where each chunk is created when the size threshold is exceeded,
-either in bytes (-b) or megabytes (-s).  The path of each chunk is a subdirectory
-in the specified directory (-R or ZAR_ROOT) where the subdirectory name is derived from the
+The import command provides a way to create a new zar archive with ZNG data
+from an existing file, S3 location, or stdin.
+
+The input data is sorted and paritioned by time into chunks. Each chunk is
+created when the size threshold is exceeded, either in bytes (-b) or megabytes
+(-s).  The path of each chunk is a subdirectory in the specified root
+(-R or ZAR_ROOT) where the subdirectory name is derived from the
 timestamp of the first zng record in that chunk.
 `,
 	New: New,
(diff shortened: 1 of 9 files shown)


cmd/zar/import/command.go
```diff
@@ -18,14 +18,16 @@ import (
...
-It takes as input zng data and cuts the stream
-into chunks where each chunk is created when the size threshold is exceeded,
-either in bytes (-b) or megabytes (-s).  The path of each chunk is a subdirectory
-in the specified directory (-R or ZAR_ROOT) where the subdirectory name is derived from the
+The import command provides a way to create a new zar archive with ZNG data
+from an existing file, S3 location, or stdin.
+
+The input data is sorted and paritioned by time into chunks. Each chunk is
+created when the size threshold is exceeded, either in bytes (-b) or megabytes
+(-s).  The path of each chunk is a subdirectory in the specified root
```
nwt, line 29: There's no -b flag, and -s uses units.Base2Bytes, not megabytes.

alfred-landrum:
We do need the `-R` option for all of the zar commands. My ergonomics opinion is that I don't like needing to remember if I put an option after global command or the subcommand, so am happy with how -R works now, but can take that up with everyone separately.

Issue #970, opened 2020-07-04
spilled records not cleaned up if zq is interrupted


mikesbrown:
If `zq` spills to disk on a sort or group-by, the records in `TempDir("", "zq-sort-")` aren't cleaned up if zq is interrupted.

I was able to simulate this by doing something like `zq -t 'sort ts' zq-sample-data/zeek-default/*.log.gz` and hitting Ctrl+C once `/tmp/zq-sort-*` appeared.

nwt:
Should be fixed by the combination of #1093 and #1099.

philrz:
Verified in `zq` commit `8bdbdb0`.

The first attached video captures the repro as of `zq` commit `9310cd0` that came just before the change. We do a `watch -n1 'ls $TMPDIR | grep zq-sort'` while attempting a `zq -t "sort -r ts" *` on a massive data set. As you can see, the `zq-sort-*` directory appears when the spill-to-disk kicks in, and when I Control-C the `zq`, the `zq-sort-*` directory is left behind.

[Repro.zip](https://github.com/brimsec/zq/files/5076974/Repro.zip)

Then repeating the same in `zq` commit `8bdbdb0`, the `zq-sort-*` directory is cleaned up automatically when I Control-C my `zq`.

[Verify.zip](https://github.com/brimsec/zq/files/5076984/Verify.zip)

Thanks @nwt!

PR #1099, opened 2020-08-13
Add signal-canceled context to commands that call driver.Run

nwt:
proc.GroupBy and proc.Sort rely on context cancelation for cleanup of
temporary files created by proc.runManager.  Add a signal-canceled
context in pkg/signalctx and use it in commands that call driver.Run
(and thus might create a proc.GroupBy or proc.Sort) to ensure that
cleanup happens.

This in combination with #1093, this should address #970.

### DIFF
--- a/cmd/zq/zq.go
+++ b/cmd/zq/zq.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"context"
 	"encoding/json"
 	"errors"
 	"flag"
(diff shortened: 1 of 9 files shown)


cmd/zq/zq.go
```diff
@@ -250,7 +250,9 @@ func (c *Command) Run(args []string) error {
 	if !c.quiet {
 		d.SetWarningsWriter(os.Stderr)
 	}
-	if err := driver.Run(context.Background(), d, query, c.zctx, reader, driver.Config{
+	ctx, cancel := signalctx.New(os.Interrupt)
+	defer cancel()
+	if err := driver.Run(ctx, d, query, c.zctx, reader, driver.Config{
```
mattnibs, line 255: I can't believe this wasn't doing this. Good stuff.

PR #1103, opened 2020-08-14
fix merge behavior on eos

alfred-landrum:
Found while working on #1074  : merge would block forever after all of its upstreams reported eos, it returns a nil batch from Pull, and then a subsequent Pull occurred again. That branch will use Merge, and this issue was visible with many tests locking up, hence I'm not adding explicit tests here.

### DIFF
--- a/proc/merge.go
+++ b/proc/merge.go
@@ -84,8 +84,10 @@ func (m *Merge) Pull() (zbuf.Batch, error) {
 			go m.run()
 		}
 	})
-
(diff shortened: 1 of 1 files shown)


henridf:
> returns a nil batch from Pull, and then a subsequent Pull occurred again

@alfred-landrum: am I reading correctly that some downstream proc was calling `Pull()` after getting a `nil` batch?  I think that would be a bug...

Issue #1036, opened 2020-07-22
Configurable merging


henridf:
#1022 will automatically create split flowgraphs that have a merge point. As described in that issue, in some cases, records can be read in arbitrary order at the merge point, while in other cases the merge point needs to read them in order of a configurable field.

This issue consists of adding support for either behaviors. Currently the merge proc is implicitly added during compilation rather than being represented in the AST; it is tbd whether that should remain the case. Maybe we can have an implicit "unordered merge", and an explicit "order on X merge" that is represented in the AST.

alfred-landrum:
We'll need an "order on X merge" when we distribute queries without aggregations, to maintain the currently expected reverse timestamp order of query results.

PR #1104, opened 2020-08-14
add ordered merge proc

alfred-landrum:
I'm splitting this out of some ongoing work for https://github.com/brimsec/zq/pull/1074 . This is also the Proc part of the work for https://github.com/brimsec/zq/issues/1036 .

### DIFF
--- a/proc/orderedmerge.go
+++ b/proc/orderedmerge.go
@@ -1,0 +1,113 @@
+package proc
+
+import (
+	"sync"
+
+	"github.com/brimsec/zq/zbuf"
+	"github.com/brimsec/zq/zng"
+)
+
+// An OrderedMerge proc merges multiple upstream inputs into one output.
+// If the input streams are ordered according to the configured comparison,
+// the output of OrderedMerge will have the same order.
+type OrderedMerge struct {
+	Base
+	recCmp  zbuf.RecordCmpFn
+	once    sync.Once
+	parents []mergeParent
+}
+
+type mergeParent struct {
+	batch    zbuf.Batch
+	recs     []*zng.Record
+	recIdx   int
+	done     bool
+	proc     Proc
+	resultCh chan Result
+}
+
+func NewOrderedMerge(c *Context, parents []Proc, cmp zbuf.RecordCmpFn) *OrderedMerge {
+	m := &OrderedMerge{
+		Base:    Base{Context: c},
+		recCmp:  cmp,
+		parents: make([]mergeParent, len(parents)),
+	}
+	for i := range parents {
+		m.parents[i].proc = parents[i]
+		m.parents[i].resultCh = make(chan Result)
+	}
+	return m
+}
+
+func (m *OrderedMerge) run() {
+	for i := range m.parents {
+		p := &m.parents[i]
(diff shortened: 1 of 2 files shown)


proc/orderedmerge.go
```diff
@@ -0,0 +1,89 @@
...
+				if res.Batch == nil {
+					m.done[i] = true
+					continue
+				}
+				m.batches[i] = res.Batch
+			case <-m.Context.Done():
+				return nil, m.Context.Err()
+			}
+		}
+		if idx == -1 || m.compare(m.batches[i], m.batches[idx]) {
```
henridf, line 79: Unless I'm missing something, I don't think we can always use batch-by-batch comparison here. I think we need to do record by record (like `zbuf.Combiner` does, and like sort/groupby do when reading from spills). 

(Batch-by-batch could be an optimization for cases where we know that batches are non-overlapping in the ordering key, but I'd suggest we add that at a later stage if necessary).

alfred-landrum: Thanks, I'll adjust. For the use case with archives, I was working on the assumption that there is no overlap between the batches we use for the parallel scanner, but that's a very narrow use case that doesn't work well in this general proc.

alfred-landrum: I've added the per-record changes & a test, PTAL!

proc/orderedmerge.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/brimsec/zq/zbuf"
+	"github.com/brimsec/zq/zng"
+)
+
+// An OrderedMerge proc merges multiple upstream inputs into one output.
+// If the input streams are ordered according to the configured comparison,
+// the output of OrderedMerge will have the same order.
+type OrderedMerge struct {
+	Base
+	recCmp  zbuf.RecordCmpFn
```
henridf, line 15: This seems like it should be `proc.CompareFn`

proc/orderedmerge_test.go
```diff
@@ -0,0 +1,134 @@
...
+		for _, r := range b.Records() {
+			if err := w.Write(r); err != nil {
+				return "", err
+			}
+		}
+		b.Unref()
+	}
+}
+
+func fieldReadCompare(field string, reversed bool) (zbuf.RecordCmpFn, error) {
```
henridf, line 69: nit: `reversed` isn't used in this func

henridf: (also we could use expr.NewCompareFn instead of this).

proc/orderedmerge_test.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/brimsec/zq/proc"
+	"github.com/brimsec/zq/zbuf"
+	"github.com/brimsec/zq/zio/tzngio"
+	"github.com/brimsec/zq/zng"
+	"github.com/brimsec/zq/zng/resolver"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+var omTestInputs = []string{
```
henridf, line 22: nit: place this closer to the test outputs (`cases`)

Issue #927, opened 2020-06-21
Unclear error message in response to bad addr


philrz:
Repro is with `zq` commit `ee18903`.

If I make a typo in an IP address:

```
$ zq -t "id.resp_h=10.150.0.999" *
bad addr value field [49 48 46 49 53 48 46 48 46 57 57 57]
```

marktwallace:
Notes:
To repro this, I used sample data from conn.log. For commit ee18903 I got the same error as above:

$ zq -t "id.resp_h=10.150.0.999" conn.log
bad addr value field [49 48 46 49 53 48 46 48 46 57 57 57]

In the branch I pulled from master to fix it, I get:

$ zq -t "id.resp_h=10.150.0.999" cmd/zq/conn.log
bad addr value field [49 48 46 49 53 48 46 48 46 57 57 57]

So the bug is still there, and I will go find it now.

philrz:
Verified in `zq` commit `04f2c11`. The error message is significantly improved:

```
$ zq -t "id.resp_h=10.150.0.999" *
net.ParseIP: parsing "10.150.0.999": syntax error
```

This also led me to think more about the nature of such error messages, and I said more about this in [a late PR comment](https://github.com/brimsec/zq/pull/1106#discussion_r473137566). However, we've definitely addressed the tactical issue at hand.

Thanks @marktwallace!

PR #1106, opened 2020-08-14
Provide better error message for bad IP address

marktwallace:


pkg/byteconv/byteconv.go
```diff
@@ -21,9 +22,17 @@ func ParseBool(b []byte) (bool, error) {
...
 func ParseIP(b []byte) (net.IP, error) {
-	ip := net.ParseIP(UnsafeString(b))
+	unsafeStr := UnsafeString(b)
+	ip := net.ParseIP(unsafeStr)
 	if ip == nil {
-		return nil, fmt.Errorf("bad addr value field %v", b)
+		var err error
+		if utf8.Valid(b) {
+			// Print a better error message if b is a utf8 string
+			err = fmt.Errorf("Cannot parse %s as an IP address", unsafeStr)
```
alfred-landrum, line 31:
```suggestion
			err = fmt.Errorf("Cannot parse %s as an IP address", string(b))
```

pkg/byteconv/byteconv.go
```diff
@@ -21,9 +22,17 @@ func ParseBool(b []byte) (bool, error) {
 }
 
 func ParseIP(b []byte) (net.IP, error) {
-	ip := net.ParseIP(UnsafeString(b))
+	unsafeStr := UnsafeString(b)
+	ip := net.ParseIP(unsafeStr)
```
alfred-landrum, line 26:
```suggestion
	ip := net.ParseIP(UnsafeString(b))
```

alfred-landrum: My suggestions are to keep the UnsafeString usage contained to the one spot where its seemingly safe & provides a potential efficiency gain.

pkg/byteconv/byteconv.go
nwt: How about just
```go
return nil, fmt.Errorf("bad addr value field %q", b)
```
so we get the same representation for b used in errors from strconv?

marktwallace: The issue raised by Phil (#927) was that "bad addr value field" followed by a byte array was confusing for end users who had supplied an IP address as part of a query. Another approach to fixing that would be to leave this function alone, as you suggest, and create another code path to validate IP addresses provided as query strings. I will defer to your preference as to whether to change this code, or add a separate validation function for user provided IP addr.

nwt: My original comment wasn't very clear. I'm sorry about that.

I'm not suggesting we leave this function alone. It's definitely the right place to fix #927.

What I meant to suggest is that we change the %v verb in the current format specifier to %q, like this.
```diff
- return nil, fmt.Errorf("bad addr value field %v", b)
+ return nil, fmt.Errorf("bad addr value field %q", b)
```

%q ultimately invokes strconv.AppendQuote (https://github.com/golang/go/blob/go1.15/src/fmt/format.go#L444-L459), and that would give us the same representation used in errors from the strconv.Parse* functions (https://github.com/golang/go/blob/go1.15/src/fmt/format.go#L444-L459) that implement everything else in this file.

If we also want to change the "bad addr value field" string, I suggest something like "net.ParseIP: parsing %q: syntax error" for symmetry with strconv.NumError.Error.

Whatever we do decide, [avoid capitalizing the first word of the error string](https://github.com/golang/go/wiki/CodeReviewComments#error-strings).

marktwallace:
I'm changing the error message to be the same format as the others from strconv, as Noah suggested. I also removed the check for invalid utf8 in the error string, since I saw that the other functions in this file do not do that check.

pkg/byteconv/byteconv.go
```diff
@@ -23,7 +23,7 @@ func ParseBool(b []byte) (bool, error) {
 func ParseIP(b []byte) (net.IP, error) {
 	ip := net.ParseIP(UnsafeString(b))
 	if ip == nil {
-		return nil, fmt.Errorf("bad addr value field %v", b)
+		return nil, fmt.Errorf("net.ParseIP: parsing %q: syntax error", UnsafeString(b))
```
nwt, line 26: Nit: %q treats a byte slice identically to a string, so UnsafeString is effectively a NOP here.
```suggestion
		return nil, fmt.Errorf("net.ParseIP: parsing %q: syntax error", b)
```

pkg/byteconv/byteconv.go
philrz: I'm late to this one because I don't manage to follow every dev PR. But I just wanted to say that while I respect the spirit of standardizing that @nwt has advocated here, let's not lose sight of the fact that users actually see these errors. Showing the errors with names of Go functions or following the Go standards of how error messages are formatted in Go context is surely helpful to us as Go developers, and also to community Go developers who start working on the project. On the other hand, security analysts that only know the Brim app may have zero coding/scripting experience and may benefit from error messages that are closer to natural language. Perhaps the ultimate solution would be to have alternate wordings for errors and a way to automatically deliver the appropriate one based on audience, e.g. whether they're an app user or a zq user, or whether they've got some preference set for "developer mode" etc. In any case, I'm not advocating we dig into any of that now, but figured I'd mention it while the topic is fresh in everyone's minds.

Issue #904, opened 2020-06-12
Improve guidance to user when there's missing AWS config


philrz:
As a user trying out our AWS features, I stumbled into at least one behavior that appeared at first like it could be a bug on our part. Closer inspection of the docs for the AWS SDK for Go indicates it's actually a quirk of that SDK and hence not technically "our problem". However, as we rely on the SDK, we're indirectly responsible for its imperfect behaviors, so we might stand to make improvements here. It probably makes sense to prioritize the effort relative to the user uptake on these AWS features.

Let's say I start out on a Linux VM that has no existing AWS tools installed. Unsurprisingly, my first attempt to use an AWS feature in `zq` (commit `8820a88` at the moment) comes back with an error message.

```
$ echo '{"foo": "bar"}' | zq -o s3://zq-771/foo.zng -
MissingRegion: could not find region configuration
```

For starters, it might have helped here if this error (and any other errors we may bubble up from the AWS SDK) were somehow labeled as being specific to AWS. That said, any user trying to use an AWS feature is certain to be familiar with the importance of the Region setting and hence will know what this error is about, so this is not a crisis.

Now let's say the error has made me hip to the absence of base AWS config on my VM. I install and configure the familiar AWS CLI tools.

```
$ sudo apt install awscli
...
The following NEW packages will be installed:
  awscli docutils-common python3-botocore python3-colorama python3-docutils
  python3-jmespath python3-pyasn1 python3-pygments python3-roman python3-rsa
  python3-s3transfer sgml-base xml-core
...
Setting up awscli (1.14.44-1ubuntu1) ...

$ aws configure
AWS Access Key ID [None]: XXXXXXXXXXXXX
AWS Secret Access Key [None]: XXXXXXXXXXXXX
Default region name [None]: us-west-1
Default output format [None]: 

$ aws s3 ls s3://zq-771
2020-06-12 13:06:01         14 hello.txt
```

My base AWS CLI config was successful, as evidenced by the fact the `ls` on my bucket succeeded. `zq help` doesn't mention anything about AWS config, so as a user I'm led to assume that, like many other AWS tools, `zq` will likely derive the config from the contents of my `$HOME/.aws` directory. Alas, my `zq` command line still does not work.

```
$ echo '{"foo": "bar"}' | zq -o s3://zq-771/foo.zng -
MissingRegion: could not find region configuration
```

[...]

PR #1109, opened 2020-08-15
enable shared config for aws sessions

alfred-landrum:
Unless a user has explicitly set an environment variable to control whether config from their shared config file (`~/.aws/config`) is loaded, do so automatically, so that users don't need the AWS_SDK_LOAD_CONFIG=1 setting in their setup.

Related to #904. This won't close that issue, as it doesn't improve the error if a user still doesn't have their config in place, but should help avoid users running into it the first time they try to use S3 objects.

### DIFF
--- a/pkg/s3io/s3io.go
+++ b/pkg/s3io/s3io.go
@@ -220,6 +220,17 @@ func newClient(cfg *aws.Config) *s3.S3 {
 		cfg.Endpoint = aws.String(endpoint)
 		cfg.S3ForcePathStyle = aws.Bool(true) // https://github.com/minio/minio/tree/master/docs/config#domain
 	}
-	sess := session.Must(session.NewSession(cfg))
+
+	// Unless the user has a environment setting for shared config, enable it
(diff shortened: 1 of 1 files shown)


pkg/s3io/s3io.go
```diff
@@ -220,6 +220,17 @@ func newClient(cfg *aws.Config) *s3.S3 {
 		cfg.Endpoint = aws.String(endpoint)
 		cfg.S3ForcePathStyle = aws.Bool(true) // https://github.com/minio/minio/tree/master/docs/config#domain
 	}
-	sess := session.Must(session.NewSession(cfg))
+
+	// Unless the user has a environment setting for shared config, enable it
+	// so that region & other info is automatically picked up from the
+	// .aws/config file.
+	scs := session.SharedConfigEnable
+	if v := os.Getenv("AWS_SDK_LOAD_CONFIG"); v != "" {
```
nwt, line 228:
```suggestion
	if os.Getenv("AWS_SDK_LOAD_CONFIG") != "" {
```

PR #1110, opened 2020-08-16
rework zdx as single-file microindex

mccanne:
This commit reworks things so microindexes are now a single file.


cmd/zdx/section/command.go
```diff
@@ -0,0 +1,106 @@
...
+	Name:  "section",
+	Usage: "section [flags] path",
+	Short: "extract a section of a zdx file",
+	Long: `
+The section command extracts a section from a microindex file and
+writes it to the output.  The -trailer option writes
+the microindex trailer to the output in addition to the section if the section
+number was specified.
+
+See the zdx command help for a description of a microindex.`,
```
henridf, line 28: `zdx help` refers to boom sst files and doesn't say "microindex"... time to update that one I think 🙂

zdx/writer.go
```diff
@@ -28,102 +31,192 @@ import (
...
+	buffer     *bufwriter.Writer
+	zng        *zngio.Writer
+	frameStart int64
+	frameEnd   int64
+	frameKey   *zng.Record
+	headerSize int
+}
+
+// NewWriter returns a Writer ready to write a microindex or it returns
+// an error.  The microindex is written to the URL given while temporary
```
henridf, line 60:
```suggestion
// an error.  The microindex is written to the URL provided in the path argument while temporary
```

zdx/writer.go
```diff
@@ -28,102 +31,192 @@ import (
...
+// file are written locally.  Calls to Write must
 // provide keys in increasing lexicographic order.  Duplicate keys are not
 // allowed but will not be detected.  Close() must be called when done writing.
-func NewWriter(zctx *resolver.Context, path string, keyFields []string, framesize int) (*Writer, error) {
+func NewWriter(zctx *resolver.Context, path string, keyFields []string, frameThresh int) (*Writer, error) {
```
henridf, line 64: If frameThresh is a better name (I'm still catching up with all of this...), should we change the various places that still refer to `framesize`? I'm primarily wondering about CLI docs...

zdx/writer.go
```diff
@@ -28,102 +31,192 @@ import (
...
-	return w.writer.Flush()
+	return w.writer.write(rec)
 }
 
 func (w *Writer) Close() error {
-	// Make sure to pass up framekeys to parent trees, even though frames aren't
-	// full.
-	if w.parent != nil && w.frameKey != nil {
-		if err := w.endFrame(); err != nil {
+	// Delete the temp files comprising the index hiearchy.
```
henridf, line 111:
```suggestion
	// Delete the temp files comprising the index hierarchy.
```

zdx/writer.go
```diff
@@ -28,102 +31,192 @@ import (
...
+	defer os.RemoveAll(w.tmpdir)
+	if w.writer == nil {
+		// If the writer hasn't been created because no records were
+		// encounted, just close the underlying storage object and return.
+		return w.iow.Close()
+	}
+	// First, close the parent if it exist (which will recursively close
```
henridf, line 118:
```suggestion
	// First, close the parent if it exists (which will recursively close
```

zdx/writer.go
mccanne: Yes, I'll update.

zdx/trailer.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	return nil, 0, errors.New("trailer not found")
+}
+
+func readOne(reader *bytes.Reader) *zng.Record {
+	r := zngio.NewReader(reader, resolver.NewContext())
+	rec, err := r.Read()
+	if err != nil && err != io.EOF {
+		rec = nil
+	}
```
nwt, line 114: A zbuf.Reader.Read implementation should never io.EOF and should never return both a non-nil record and a non-nil error.
```suggestion
	rec, _ := r.Read()
```

zdx/writer.go
```diff
@@ -28,102 +31,192 @@ import (
...
-func newWriter(zctx *resolver.Context, path iosrc.URI, keyFields []string, framesize, level int, hdr *zng.Record) (*Writer, error) {
-	if level > 5 {
-		panic("something wrong")
+	writer, err := iosrc.NewWriter(uri)
+	if err != nil {
+		return nil, err
 	}
-	name := filename(path, level)
-	w, err := iosrc.NewWriter(name)
+	tmp, err := ioutil.TempDir("", "*")
```
nwt, line 79: Nit: Give the directory prefix suggestive of its origin.
```suggestion
	tmp, err := ioutil.TempDir("", "zdx-")
```

zdx/writer.go
```diff
@@ -28,102 +31,192 @@ import (
...
+	if err != nil {
 		return err
 	}
-	if w.parent != nil {
-		return w.parent.Close()
+	if err := base.zng.Write(rec); err != nil {
+		return err
 	}
+	base.zng.EndStream()
 	return nil
```
nwt, line 185:
```suggestion
	return base.zng.EndStream()
```

zdx/writer.go
```diff
@@ -28,102 +31,192 @@ import (
...
+		base:     base,
+		buffer:   writer,
+		name:     name,
+		zng:      zngio.NewWriter(writer, zio.WriterFlags{}),
+		frameEnd: int64(base.frameThresh),
+	}, nil
+}
+
+func (w *indexWriter) newParent() (*indexWriter, error) {
+	file, err := ioutil.TempFile(w.base.tmpdir, "*")
```
nwt, line 200: NIt: We conventionally use "", which is equivalent in this context to "*".
```suggestion
	file, err := ioutil.TempFile(w.base.tmpdir, "")
```

zdx/zdx.go
```diff
@@ -1,23 +1,14 @@
...
+// microindexes.
 //
-// A table on disk consists of the base table with zero or more b-tree files.
-// These files are all named with the same path prefix, e.g., "zdx.zng", where the
-// base table is zdx.zng and the b-tree index files, if any, are zdx.1.zng,
-// zdx.2.zng, and so forth.  The index files can always be reconstructed from
-// the base table.
+// A microindex comprises a base index section followed by zero or more parent
+// section indexes followed by a trailer.  The sections are organized into a
+// btree-like data structure so keys can be looked up efficiently without
```
nwt, line 6: Nit: This name is usually styled as B-tree.
```suggestion
// B-tree-like data structure so keys can be looked up efficiently without
```

zdx/zdx.go
```diff
@@ -1,23 +1,14 @@
...
-//
-// The zdx base file and index files are all organized as a sequence of
-// zng streams.
-//
-// Each b-tree index file contains a key,value pair for each stream in the file
-// below in the hiearchy where the key is the first key found in that stream and
-// the value is the offset or the stream in the file below.
+// The trailer provides meta information about the microindex, e.g., indicating
+// the sizes of each section (so section boundaries can be found), the keys
+// that were indexed, the frame threshold used in build the btree hierarchy, etc.
```
nwt, line 11:
```suggestion
// that were indexed, the frame threshold used in build the B-tree hierarchy, etc.
```

zdx/trailer.go
nwt: (If you've observed either behavior from zngio.Reader.Read, you've found a bug that I want to fix.)

mccanne: Cool, thanks.  As documented in the interface declaration :-)

PR #1120, opened 2020-08-18
add version check to zdx module

### LINKED ISSUE #1116: add version check for microindex reader
The microindex reader needs to do a version check in the trailer field and fail under version mismatch.

mccanne:
This commit changes the microindex version type from a string to
an integer and adds a version check when opening a microindex
that currently fails if the version numbers do not match exactly.

Fixes #1116

### DIFF
--- a/zdx/trailer.go
+++ b/zdx/trailer.go
@@ -91,8 +91,8 @@ func readTrailer(r io.ReadSeeker, n int64) (*Trailer, int, error) {
 		// look for end of stream followed by an array[int64] typedef then
 		// a record typedef indicating the possible presence of the trailer,
 		// which we then try to decode.
-		if buf[off] == 0x81 && buf[off+1] == 0x06 && buf[off+2] == 0x80 {
-			if off > 0 && buf[off-1] != 0x85 {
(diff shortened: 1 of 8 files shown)


zdx/trailer.go
```diff
@@ -87,36 +87,56 @@ func readTrailer(r io.ReadSeeker, n int64) (*Trailer, int, error) {
...
-	for off := int64(n) - 4; off >= 0; off-- {
+	for off := int(n) - 3; off >= 0; off-- {
 		// look for end of stream followed by an array[int64] typedef then
 		// a record typedef indicating the possible presence of the trailer,
 		// which we then try to decode.
-		if buf[off] == 0x85 && buf[off+1] == 0x81 && buf[off+2] == 0x06 && buf[off+3] == 0x80 {
-			attempt := buf[off+1 : n]
-			r := bytes.NewReader(attempt)
+		if buf[off] == 0x81 && buf[off+1] == 0x06 && buf[off+2] == 0x80 {
+			if off > 0 && buf[off-1] != 0x85 {
```
nwt, line 95:
```suggestion
		if bytes.Equal(buf[off:3], []byte{zng.TypeDefArray, zng.IdUint64, zng.TypeDefRecord}) {
			if off > 0 && buf[off-1] != zng.CtrlEOS {
```

mccanne: sweet!

PR #1121, opened 2020-08-19
Configurable merge

### LINKED ISSUE #1036: Configurable merging

henridf:
Add two fields to ParallelProc to specify ordering behavior when merging branches. 

If `ParallelProc.MergeOrderField` is zero-valued, then the existing (unordered) `proc.Merge` proc is used. Otherwise, a `proc.OrderedMerge` proc is used, configured with the ordering field name and direction indicated in the new AST fields.

These new fields are not currently exposed in ZQL syntax but could be at a later point if needed.

closes #1036

### DIFF
--- a/proc/proc.go
+++ b/proc/proc.go
@@ -186,6 +186,11 @@ func CompileProc(custom Compiler, node ast.Proc, c *Context, parent Proc) ([]Pro
 			// in which case the output layer will mux
 			// into channels.
 			if len(parents) > 1 && k < n-1 {
+				p := v.Procs[k].(*ast.ParallelProc)
+				if p.MergeOrderField != "" {
+					parent = NewOrderedMerge(c, parents, p.MergeOrderField, p.MergeOrderReverse)
+					return []Proc{parent}, nil
+				}
 				parent = NewMerge(c, parents)
 			} else {
 				parent = parents[0]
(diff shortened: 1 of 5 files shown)


henridf:
Testing note: there is some existing test coverage in `proc/orderedmerge_test.go` and `driver/parallel_test.go`. More extensive testing will come along with #1022

proc/proc.go
```diff
@@ -186,6 +186,11 @@ func CompileProc(custom Compiler, node ast.Proc, c *Context, parent Proc) ([]Pro
 			// in which case the output layer will mux
 			// into channels.
 			if len(parents) > 1 && k < n-1 {
+				p := v.Procs[k].(*ast.ParallelProc)
+				if p.MergeOrderField != "" {
+					parent = NewOrderedMerge(c, parents, p.MergeOrderField, p.MergeOrderReverse)
+					return []Proc{parent}, nil
```
alfred-landrum, line 192: This early return will leave v.Procs[k+1:n] uncompiled.

henridf: d'oh!

Issue #1023, opened 2020-07-21
groupby: add support for emitting and consuming partials


henridf:
#1022 will add two new flags to `ast.GroupByProc`; one to indicate that the aggregator should output partial results, and one to indicate that partials are being fed to its input. 

This issue consists of adding that behavior to the groupby proc. Note that support for all of this is already present via the `reducer.Decomposable` interface and used by internally when spilling, so this issue should primarily consist of a little wiring and maybe some refactoring.

henridf:
part of #1004

Issue #1004, opened 2020-07-15
Parallel query execution


alfred-landrum:
In the near term, we want to execute a ZQL query in parallel over the ZNG files in an archive. Doing so safely & efficiently requires analyzing the query's AST, moving portions of the query to separate execution cores.

For this epic, we want to:
* create some initial framework code to support AST analysis/transformation passes
* use that framework to write a pass enabling parallel execution on a ZAR archive, for at least some simple cases, like a search-only query, or a search immediately followed by an aggregation.

Here's a small slide set we used for an internal discussion:
https://docs.google.com/presentation/d/1fQBD12BELuIaHL3faS-94HB5iGEdkBK5TPy-IU4TMzA/edit?usp=sharing

PR #1122, opened 2020-08-19
Remove reduce proc

henridf:
We've discussed in the past how we don't need a separately implemented reduce proc; this PR finally removes it.

(This was prompted by realizing that #1023 needed to be mirrored for the reduce proc, and figuring that instead this would be a good time to remove it).


proc/proc.go
```diff
@@ -91,18 +90,8 @@ func CompileProc(custom Compiler, node ast.Proc, c *Context, parent Proc) ([]Pro
...
-			if err != nil {
-				return nil, err
-			}
-			reducers = append(reducers, compiled)
-		}
-		params := ReduceParams{
-			reducers: reducers,
-		}
-		return []Proc{NewReduce(c, parent, params)}, nil
+		g := &ast.GroupByProc{Reducers: v.Reducers}
```
henridf, line 93: (The endgame is probably to change zql.peg to just emit an ast.GroupByProc and remove ast.ReduceProc altogether; I'd rather do that in a separate step right after wrapping up #1004 )

alfred-landrum:
> The PR title is misleading given that you're leaving ast.ReduceProc in place. Please make it more specific (e.g., "Remove proc.Reduce"), and please also update the commit message when you merge.

fwiw: When I see/hear "proc" used, I'm assuming proc.Proc, but when I think about the AST, I think of "nodes". So I understood the title to mean what the code does.

henridf:
Fwiw, for me "reduce proc" also brings to mind the runtime processor rather than the AST node.

PR #1124, opened 2020-08-19
make MultiFileReader ScannerAble

alfred-landrum:
Make a MultiFileReader implement the scanner.ScannerAble interface, so that we can use the scanner speedups on the individual files in a MultiFileReader. While working on the test, I also found a need to replicate some code to copy data from a Pull() like object to a Writer, so I added a `zbuf.Puller` interface and a copy utility.

Needed for https://github.com/brimsec/zq/issues/1075 .

### DIFF
--- a/scanner/scanner.go
+++ b/scanner/scanner.go
@@ -25,7 +25,7 @@ type Statser interface {
 // A Scanner is a zbuf.Batch source that also provides statistics.
 type Scanner interface {
 	Statser
-	Pull() (zbuf.Batch, error)
+	zbuf.Puller
 }
 
 // ScannerStats holds Scanner statistics. It should be identical to
--- a/scanner/scanner.go
+++ b/scanner/scanner.go
@@ -25,7 +25,7 @@ type Statser interface {
 // A Scanner is a zbuf.Batch source that also provides statistics.
 type Scanner interface {
 	Statser
-	zbuf.Puller
+	Pull() (zbuf.Batch, error)
 }
 
 // ScannerStats holds Scanner statistics. It should be identical to
--- a/zio/detector/file.go
+++ b/zio/detector/file.go
@@ -5,10 +5,15 @@ import (
 	"io"
 	"net/url"
 	"os"
+	"sync"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/brimsec/zq/ast"
+	"github.com/brimsec/zq/filter"
 	"github.com/brimsec/zq/pkg/iosrc"
+	"github.com/brimsec/zq/pkg/nano"
 	"github.com/brimsec/zq/pkg/s3io"
+	"github.com/brimsec/zq/scanner"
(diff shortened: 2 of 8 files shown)


zio/detector/file.go
```diff
@@ -108,53 +113,123 @@ func OpenFiles(zctx *resolver.Context, dir zbuf.RecordCmpFn, paths ...string) (z
 	return zbuf.NewCombiner(readers, dir), nil
 }
 
-type multiFileReader struct {
+type MultiFileReader struct {
```
nwt, line 116: Does this need to be public?

scanner/scanner.go
```diff
@@ -25,7 +25,7 @@ type Statser interface {
 // A Scanner is a zbuf.Batch source that also provides statistics.
 type Scanner interface {
 	Statser
-	Pull() (zbuf.Batch, error)
+	zbuf.Puller
```
nwt, line 28: If we're going to do this, we should probably (a) do the same thing to proc.Proc and (b) move the verbiage in the proc.Proc doc comment regarding Pull over to zbuf.Puller.

zio/detector/file.go
alfred-landrum: It's not required.

scanner/scanner.go
alfred-landrum: I can put zbuf.Puller into proc.Proc, but I think the proc.Proc comment reads fine as is.

alfred-landrum: > I can put zbuf.Puller into proc.Proc, but I think the proc.Proc comment reads fine as is.

Actually, I take that back: I'll just undo the change to scanner.Scanner.

nwt: What I'm getting at is that zbuf.Puller and proc.Proc each have a Pull method with identical semantics, that we should document those semantics once instead of twice, and that zbuf.Puller seems like the place to do that.

nwt: I'm happy with leaving scanner.Scanner as is.

zio/detector/file.go
```diff
@@ -158,3 +175,61 @@ func (r *multiFileReader) Close() (err error) {
...
+}
+
+type multiFileScanner struct {
+	*multiFileReader
+	ctx        context.Context
+	filter     filter.Filter
+	filterExpr ast.BooleanExpr
+	span       nano.Span
+
+	mu      sync.Mutex // protects below
```
nwt, line 196: It just occurred to me to ask: Do you expect to have multiple goroutines active in here?

nwt: That's the wrong question. The right one: Do you expect to have multiple goroutines fiddling with the scanner field?

alfred-landrum: The mutex is to ensure that Stats() calls from the mux:
https://github.com/brimsec/zq/blob/master/driver/driver.go#L43
are synchronized wrt to the scanner change.

PR #1127, opened 2020-08-19
fix reverse OrderedMerge comparison

alfred-landrum:
Using the reverse ordering led to stack overflow; fix & add a test for the reverse case.


henridf (approved):
Thanks! 
(I was sure Go would bind `cmp` to the earlier definition rather than the LHS... the actual behavior seems surprising, but my bad for not having added a test).

alfred-landrum:
> Thanks!
> (I was sure Go would bind `cmp` to the earlier definition rather than the LHS... the actual behavior seems surprising, but my bad for not having added a test).

I missed this in my earlier review, and was surprised when I this with some local changes!

PR #1138, opened 2020-08-20
add zar map and archive multisource support

### LINKED ISSUE #1092: create zar map subcommand
After discussing the two modes of operation that `zar zq` supports:
https://github.com/brimsec/zq/pull/1075#issue-464902803

we decided that we would split out the per-file output operation (behavior 2 in the link above) to a new subcommand, `zar map`. `zar zq` will generate an aggregate result by operating all of the chunk files, or some other per-zar directory file, in the archive.

alfred-landrum:
This split out the "for each file" like operations of `zar zq` into a new subcommand `zar map`, adds multisource support for archive.Archive, and uses the multisource support in the remaining uses of `zar zq`.

Closes #1092

### DIFF
--- a/archive/walk.go
+++ b/archive/walk.go
@@ -1,9 +1,16 @@
 package archive
 
 import (
+	"context"
+	"io"
 	"strings"
 
+	"github.com/brimsec/zq/driver"
 	"github.com/brimsec/zq/pkg/iosrc"
+	"github.com/brimsec/zq/scanner"
+	"github.com/brimsec/zq/zbuf"
+	"github.com/brimsec/zq/zio/detector"
+	"github.com/brimsec/zq/zng/resolver"
 )
 
 func ZarDirToLog(uri iosrc.URI) iosrc.URI {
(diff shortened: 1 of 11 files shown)


alfred-landrum:
I started editing the zar README for this work, then passed on an early draft with zar map / zar zq changes (plus fixes due to zar directory paths due to compression) to @philrz , who'll be adding to my changes & putting up a separate PR.

archive/walk.go
```diff
@@ -66,3 +73,73 @@ func SpanWalk(ark *Archive, v SpanVisitor) error {
...
+// chunk file itself), with no defined ordering.
+func NewMultiSource(ark *Archive, paths []string) driver.MultiSource {
+	if len(paths) == 0 {
+		paths = []string{"_"}
+	}
+	ms := &multiSource{
+		ark:   ark,
+		paths: paths,
+	}
+	return ms
```
nwt, line 95:
```suggestion
	return &multiSource{
		ark:   ark,
		paths: paths,
	}
```

cmd/zar/map/command.go
```diff
@@ -0,0 +1,162 @@
...
+		reader := zbuf.Reader(rc)
+		wch := make(chan string, 5)
+		if !c.stopErr {
+			reader = zbuf.NewWarningReader(reader, wch)
+		}
+		writer, err := c.openOutput(zardir)
+		if err != nil {
+			return err
+		}
+		defer writer.Close()
```
nwt, line 139: Check for an error from writer.Close.

cmd/zar/map/command.go
```diff
@@ -0,0 +1,162 @@
...
+		return driver.Run(ctx, d, query, zctx, reader, driver.Config{
+			ReaderSortKey:     "ts",
+			ReaderSortReverse: ark.DataSortDirection == zbuf.DirTimeReverse,
+			Warnings:          wch,
+		})
+	})
+}
+
+func (c *Command) openOutput(zardir iosrc.URI) (zbuf.WriteCloser, error) {
+	path := ""
```
nwt, line 153: Nit: Use var for the zero value.
```suggestion
	var path string
```

archive/walk.go
```diff
@@ -66,3 +73,73 @@ func SpanWalk(ark *Archive, v SpanVisitor) error {
...
+	ark   *Archive
+	paths []string
+}
+
+// NewMultiSource returns a driver.MultiSource for an Archive. If no paths are
+// specified, the MultiSource will send a source for each chunk file, and
+// report the same ordering as the archive. Otherwise, the sources come from
+// localizing the given paths to each chunk's directory (recognizing "_" as the
+// chunk file itself), with no defined ordering.
+func NewMultiSource(ark *Archive, paths []string) driver.MultiSource {
```
henridf, line 87: Now I see driver.MultiSource used here, it seems like an inversion to have a core library use and return types defined in an outer package like `driver`. Maybe driver.MultiSource could move to another core library like scanner/. Either way, this doesn't have to be addressed here.

cmd/zar/zq/command.go
```diff
@@ -6,28 +6,26 @@ import (
...
 	Name:  "zq",
-	Usage: "zq [-R root] [options] [zql] file [file...]",
-	Short: "walk an archive and run zql queries",
+	Usage: "zq [-R root] [options] zql [file...]",
+	Short: "execute ZQL against all archive directories",
 	Long: `
-"zar zq" executes a ZQL query against each chunk or associated file in an 
-archive. The special name "_" refers to chunk file itelf, and other names
-are interpreted relative to the chunk's associated file directory.
+"zar zq" executes a ZQL query against one or more files from all the directories
```
henridf, line 24: Not for now, as this isn't new to this PR, but reading this help text makes me think we'll want to document these sub-commands in more user-visible terms, without it being a description of mechanics invoking the directory layout etc. Easier said than done but we'll get there 🙂. cc @philrz.

cmd/zar/zq/command.go
```diff
@@ -6,28 +6,26 @@ import (
...
+of an archive, generating a single result. By default, the chunk file in each
```
henridf, line 25:
```suggestion
of an archive, generating a single result stream. By default, the chunk file in each
```

PR #1141, opened 2020-08-20
zqd: Enable archive store api ingest

### LINKED ISSUE #1085: zed backend use lakes for default storage
We should transition to use the Zed lake (aka "archivestore") space storage by default in zqd. We may keep the `filestore` kind available for compatibility with existing Brim spaces, but it could transition to read-only.

mattnibs:
Closes #1085

### DIFF
--- a/archive/schema.go
+++ b/archive/schema.go
@@ -240,6 +240,7 @@ func (ark *Archive) UpdateCheck() (int, error) {
 	}
 
 	ark.spans = md.Spans
+	ark.indexes = md.Indexes
 	ark.mdModTime = mtime
 	ark.mdUpdateCount++
 	return ark.mdUpdateCount, nil
--- a/zqd/api/stream.go
+++ b/zqd/api/stream.go
@@ -27,3 +27,24 @@ func (s *Stream) Next() (interface{}, error) {
 	}
 	return nil, nil
 }
+
+func ReadStream(s *Stream) ([]interface{}, error) {
+	var payloads []interface{}
+	for {
+		v, err := s.Next()
+		if err != nil {
+			if err == io.EOF {
+				return payloads, nil
+			}
+			return nil, err
+		}
+		payloads = append(payloads, v)
+		switch payload := v.(type) {
+		case *TaskEnd:
+			if payload.Error != nil {
+				return nil, payload.Error
+			}
+			return payloads, nil
+		}
+	}
+}
--- a/zqd/storage/archivestore/archivestore.go
+++ b/zqd/storage/archivestore/archivestore.go
@@ -113,10 +113,26 @@ func (s *Storage) Summary(_ context.Context) (storage.Summary, error) {
 	return sum, nil
 }
 
+func (s *Storage) Write(ctx context.Context, zctx *resolver.Context, zr zbuf.Reader) error {
+	_, err := s.ark.UpdateCheck()
+	if err != nil {
+		return err
+	}
+	return archive.Import(ctx, s.ark, zctx, zr)
+}
+
 func (s *Storage) IndexSearch(ctx context.Context, zctx *resolver.Context, query archive.IndexQuery) (zbuf.ReadCloser, error) {
+	_, err := s.ark.UpdateCheck()
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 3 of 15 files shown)


archive/schema.go
```diff
@@ -240,6 +240,7 @@ func (ark *Archive) UpdateCheck() (int, error) {
 	}
 
 	ark.spans = md.Spans
+	ark.indexes = md.Indexes
```
alfred-landrum, line 243: 👍

zqd/storage/archivestore/archivestore.go
```diff
@@ -113,10 +113,26 @@ func (s *Storage) Summary(_ context.Context) (storage.Summary, error) {
...
 func (s *Storage) IndexSearch(ctx context.Context, zctx *resolver.Context, query archive.IndexQuery) (zbuf.ReadCloser, error) {
+	_, err := s.ark.UpdateCheck()
+	if err != nil {
+		return nil, err
+	}
 	return archive.FindReadCloser(ctx, zctx, s.ark, query, archive.AddPath(archive.DefaultAddPathField, false))
 }
 
 func (s *Storage) ArchiveStat(ctx context.Context, zctx *resolver.Context) (zbuf.ReadCloser, error) {
+	_, err := s.ark.UpdateCheck()
```
alfred-landrum, line 133: (similar to above, the SpanWalk inside archive.Stat should do this already)

zqd/storage/archivestore/archivestore.go
```diff
@@ -113,10 +113,26 @@ func (s *Storage) Summary(_ context.Context) (storage.Summary, error) {
...
+func (s *Storage) Write(ctx context.Context, zctx *resolver.Context, zr zbuf.Reader) error {
+	_, err := s.ark.UpdateCheck()
+	if err != nil {
+		return err
+	}
+	return archive.Import(ctx, s.ark, zctx, zr)
+}
+
 func (s *Storage) IndexSearch(ctx context.Context, zctx *resolver.Context, query archive.IndexQuery) (zbuf.ReadCloser, error) {
+	_, err := s.ark.UpdateCheck()
```
alfred-landrum, line 125: I don't think should be needed, as the archive.SpanWalk that happens inside FindReadCloser does this.

zqd/storage/archivestore/archivestore.go
```diff
@@ -113,10 +113,26 @@ func (s *Storage) Summary(_ context.Context) (storage.Summary, error) {
 	return sum, nil
 }
 
+func (s *Storage) Write(ctx context.Context, zctx *resolver.Context, zr zbuf.Reader) error {
+	_, err := s.ark.UpdateCheck()
```
alfred-landrum, line 117: Instead of an UpdateCheck here, I suspect adjustments for locking need to be made to archive.Import and/or archive.Archive.AppendSpans.

zqd/storage/archivestore/archivestore.go
mattnibs: Doesn't look like it is. Tests on ArchiveStat and ArchiveIndexSearch fail when I remove these. I'll see about adding something like this in SpanWalk.

mattnibs: Huh, it is being called. Digging...

zqd/api/stream.go
```diff
@@ -27,3 +27,24 @@ func (s *Stream) Next() (interface{}, error) {
 	}
 	return nil, nil
 }
+
+func ReadStream(s *Stream) ([]interface{}, error) {
```
nwt, line 31: Nit: Make this a method.

zqd/api/stream.go
```diff
@@ -27,3 +27,24 @@ func (s *Stream) Next() (interface{}, error) {
...
+		v, err := s.Next()
+		if err != nil {
+			if err == io.EOF {
+				return payloads, nil
+			}
+			return nil, err
+		}
+		payloads = append(payloads, v)
+		switch payload := v.(type) {
+		case *TaskEnd:
```
nwt, line 43: Nit: Shorten to
```suggestion
		if end, ok := v.(*TaskEnd); ok {
```

PR #1145, opened 2020-08-21
AST transform for parallelizing a flowgraph

### LINKED ISSUE #1022: AST transformation for distributed query execution

henridf:
Add `driver.parallelizeFlowgraph()`, a function that tries to transform a sequential flowgraph into a parallel one. 

Wiring this into compilation will come in a subsequent PR (I have that working locally).

closes #1022

### DIFF
--- a/driver/compile.go
+++ b/driver/compile.go
@@ -409,11 +409,11 @@ func decomposable(rs []ast.Reducer) bool {
 	return true
 }
 
-// distributeFlowgraph takes a sequential proc AST and tries to
-// distribute it by splitting as much as possible of the sequence into
+// parallelizeFlowgraph takes a sequential proc AST and tries to
+// parallelize it by splitting as much as possible of the sequence into
 // N parallel branches. The boolean return argument indicates whether
 // distribution succeeded.
-func distributeFlowgraph(seq *ast.SequentialProc, N int, inputSortField string, inputSortReversed bool) (*ast.SequentialProc, bool) {
+func parallelizeFlowgraph(seq *ast.SequentialProc, N int, inputSortField string, inputSortReversed bool) (*ast.SequentialProc, bool) {
 	for i := range seq.Procs {
 		switch p := seq.Procs[i].(type) {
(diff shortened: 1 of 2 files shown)


driver/compile.go
```diff
@@ -357,3 +360,118 @@ func computeColumnsR(p ast.Proc, colset map[string]struct{}) (map[string]struct{
...
+// parallelize it by splitting as much as possible of the sequence into
+// N parallel branches. The boolean return argument indicates whether
+// distribution succeeded.
+func parallelizeFlowgraph(seq *ast.SequentialProc, N int, inputSortField string, inputSortReversed bool) (*ast.SequentialProc, bool) {
+	for i := range seq.Procs {
+		switch p := seq.Procs[i].(type) {
+		case *ast.CutProc, *ast.FilterProc, *ast.PassProc, *ast.PutProc, *ast.RenameProc:
+			// Stateless procs: continue until we reach one of the procs below at
+			// which point we'll either split the flowgraph or see we can't and return it as-is.
+			continue
```
nwt, line 422: @henridf, @alfred-landrum: I don't think this does the right thing if cut, put, or rename alters the sort field. Is that handled elsewhere?

henridf: Good catch @nwt. Addressed in #1171

PR #1158, opened 2020-08-22
fix bug in put proc and refactor

### LINKED ISSUE #1152: "slice bounds out of range" during attempted type conversion
Found while assisting a community user. Repro is with `zq` commit `41a4641`.

Here I'm attempting to perform lots of type conversion, or as we've been calling it (perhaps incorrectly), "casting".

This works:

```
$ echo '{"stime":"1598051907.12345678", "saddr":"1.2.3.4" }' | zq -t "put stime=String.parseFloat(stime):time,saddr=saddr:ip" -
#0:record[saddr:ip,stime:time]
0:[1.2.3.4;1598051907.123456716;]
```

This also works:

```
$ echo '{"stime":"1598051907.12345678", "ltime":"1498051907.87654321" }' | zq -t "put stime=String.parseFloat(stime):time,ltime=String.parseFloat(ltime):time" -
#0:record[ltime:time,stime:time]
0:[1498051907.876543284;1598051907.123456716;]
```

However if I attempt to convert all three at once, now I get a panic:

```
$ echo '{"stime":"1598051907.12345678", "ltime":"1498051907.87654321", "saddr":"1.2.3.4" }' | zq -t "put stime=String.parseFloat(stime):time,ltime=String.parseFloat(ltime):time,saddr=saddr:ip" -
#0:record[ltime:time,saddr:ip,stime:time]
0:[1498051907.876543284;<ZNG-ERR type ip [%!s(PANIC=String method: runtime error: slice bounds out of range [:24] with capacity 22)]: failure trying to decode IP address that is not 4 or 16 bytes long>;1598051907.123456716;<record> (record[ltime:time,saddr:ip,stime:time]): record with extra field
```

mccanne:
The put proc had a bug where it attempted to append a clause result
after all of the clauses had already replaced fields in the original
input.  Put.put() was doing many different things with some confusing
variable and type names so a refactor was warranted.

This commit fixes the bug and refactors the logic into bite-sized chunks
with hopefully more helpful names.  The resulting code is smaller.

Fixes #1152


tests/suite/put/put-multi-5.yaml
```diff
@@ -0,0 +1,13 @@
+# Exercises the bug in zq issue 1152
```
nwt, line 1: Let's give this file a more meaningful name. Maybe put-overwrite-c-a-b.yaml, put-overwrite-3rd-1st-2nd.yaml, or put-overwrite-last-field-first.yaml?

proc/put.go
```diff
@@ -70,153 +80,133 @@ func (p *Put) maybeWarn(err error) {
...
 	}
-
-	// any remaining fields are to be appended
-	for i < len(descInfo.fields) {
-		vali := descInfo.order[i]
-		i++
-		if descInfo.fields[vali].container {
-			bytes = zcode.AppendContainer(bytes, p.vals[vali].Bytes)
+	// Finish building the output by appending the remaining clauses if any.
+	for clause := range rule.append {
```
nwt, line 194: It's wild that existing tests all pass with this loop, but
```yaml
zql: put a=3,b=2,c=1

input: |
  #0:record[a:int32,b:int32]
  0:[1;2;]

output: |
  #0:record[a:int64,b:int64,c:int64]
  0:[3;2;1;]
```
fails with
```diff
        expected and actual outputs differ:
        --- expected
        +++ actual
        @@ -1,3 +1,3 @@
         #0:record[a:int64,b:int64,c:int64]
        -0:[3;2;1;]
        +0:[3;2;3;]
```

It passes with this.
```suggestion
	for _, clause := range rule.append {
```

mccanne: Golly

tests/suite/put/put-multi-5.yaml
mccanne: I actually think that ztest should handle multiple zql tests in one yaml file and all of these put-multi-* tests should be in one file...

mccanne: ok

proc/put.go
```diff
@@ -70,153 +80,133 @@ func (p *Put) maybeWarn(err error) {
...
 		}
+	}
+	return true
+}
 
-		p.outmap[in.Type.ID()] = newinfo
-		descInfo = newinfo
+func (p *Put) lookupRule(inType *zng.TypeRecord, vals []zng.Value) (*putRule, error) {
+	rule := p.rules[inType.ID()]
+	if rule != nil && clauseTypesMatch(rule.clauseTypes, vals) {
```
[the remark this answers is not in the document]
mccanne: nils can be put in the table on error.  If I make this change, then I'll need to add a check to for error that wasn't otherwise needed.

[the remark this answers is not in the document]
mccanne: I find `rule ! =nil` clearer in this case anyway because the code requires rule not nil and it doesn't matter if nils are allowed in the table or not for this code to work correctly.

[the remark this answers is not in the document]
nwt: Doh. I didn't realize that nil could be in the map.

PR #1160, opened 2020-08-22
allow ztest yaml tests to live anywhere

mccanne:
This PR proposes an extension to the system tests to allow yaml ztests to be put anywhere in the source tree.  This would allow us to use ztest to more easily write unit-like tests throughout the code.

For example, we can now move a test like this, which is really a unit test for the put proc,
```
tests/suite/put/put-1.yaml
```
to
```
proc/put/ztests/put-1.yaml
```
The logic currently looks for any yaml files that are a descendent of a directory called `ztests` or any yaml files under the top-level `tests` directory.  I was thinking we could move `tests/suite` and `tests/formats` to a new top-level directory `ztests` and get rid of the `tests` special case maybe do this in a subsequent PR?


mccanne:
After talking to @nwt , it seems simplest to just move the system test driver from `zq/tests/ztest_test.go` to `zq/zq_test.go`

mccanne:
The latest heuristic is to find tests that end in .yaml that are a descendent of a directory called `ztests` or are under `zq/tests`.  This avoids the conflict in `zq/charts/.../tests/*.yaml`

mccanne:
One more move would be to move `zq/tests/suite` and `zq/tests/formats` to a new top-level dir `zq/ztests` and get rid of the search for yaml ztests in `zq/tests`

zq_test.go
```diff
@@ -1,21 +1,23 @@
...
 	"github.com/brimsec/zq/ztest"
 	"github.com/stretchr/testify/require"
 )
 
-func TestZTest(t *testing.T) {
+func TestZq(t *testing.T) {
 	t.Parallel()
-	dirs := map[string]bool{}
+	dirs := map[string]struct{}{}
+	re, _ := regexp.Compile(".*ztests/.*\\.yaml$")
```
alfred-landrum, line 17:
```suggestion
	re := regexp.MustCompile(`.*ztests/.*\.yaml$`)
```

alfred-landrum (approved):
thanks  for this @mccanne !

My preference would be to move the zq tests to a `ztest` folder.

zq_test.go
```diff
@@ -1,21 +1,23 @@
...
+func TestZq(t *testing.T) {
 	t.Parallel()
-	dirs := map[string]bool{}
+	dirs := map[string]struct{}{}
+	re := regexp.MustCompile(`.*ztests/.*\.yaml$`)
 	err := filepath.Walk(".", func(path string, info os.FileInfo, err error) error {
-		if strings.HasSuffix(path, ".yaml") {
-			dirs[filepath.Dir(path)] = true
+		if !info.IsDir() && strings.HasSuffix(path, ".yaml") && re.MatchString(path) {
+			dirs[filepath.Dir(path)] = struct{}{}
```
nwt, line 20: re will never match on Windows because the path separator there is \. You could fix a few ways, but I think the simplest is to assume interest in any directory named ztests. Then you could just do this.
```go
	 var dirs []string
	 err := filepath.Walk(".", func(path string, info os.FileInfo, err error) error {
		if info.IsDir() && filepath.FileName(path) == "ztests" {
			dirs = append(dirs, path)
		}
		return err
	}
```

mccanne: Just pushed a fix to match both backslash or forward slash.

nwt (approved):
:shipit:

PR #1168, opened 2020-08-24
Mention Go version in README

philrz:
A couple times lately, I've advised community users to compile `master` of `zq` so they could work around a bug before the next GA release. The steps in the README for an install seem simple enough, but the fact that we don't mention the Go version might set them up for trouble if, say, they already have Go installed on their Linux distro and assume that whatever version is there will be adequate. I know we mention the "Go 1.14 or later" requirement in the CONTRIBUTING doc, but since ordinary users just compiling code may not look at that doc, I figure we should be redundant and mention this in the README as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,8 @@ To install `zq` or any other tool from this repo, you can either clone the repo
  [release](https://github.com/brimsec/zq/releases), available for Windows, macOS, and Linux.
 
 If you don't have Go installed, download and install it from the
-[Go downloads page](https://golang.org/dl/).
+[Go downloads page](https://golang.org/dl/). Go version 1.14 or later is
+required.
 
 To install the binaries in `$GOPATH/bin`, clone this repo and
 execute `make install`:


README.md
```diff
@@ -22,7 +22,8 @@ To install `zq` or any other tool from this repo, you can either clone the repo
  [release](https://github.com/brimsec/zq/releases), available for Windows, macOS, and Linux.
 
 If you don't have Go installed, download and install it from the
-[Go downloads page](https://golang.org/dl/).
+[Go downloads page](https://golang.org/dl/). Go version 1.14 or later is
+required.
```
henridf, line 26: I'm not sure if we should say "1.14 or later" or just "1.14" (and then change update to 1.15 when that happens, etc). @alfred-landrum ?

alfred-landrum: I'm good with "1.14 or later".

philrz: Cool. Yeah, i was just mimicking what I saw in CONTRIBUTING. :)

Issue #1034, opened 2020-07-22
AST transform for parallelizing flowgraphs (step 2)


henridf:
#1022 is the first step towards splitting a flowgraph for parallel execution. It handles the case of splitting at a decomposable aggregation that comes after stateless procs. 

This is a placeholder for more general splitting. I'll just provide a few examples of what could be done:

1. Splitting at `sort x`, with arbitrary input sortedness

`<stateless procs> ... | sort x | <other procs>` can be transformed into

```
(
  <stateless procs> ... | sort x;
  ...                                                  
  <stateless procs> ... | sort x
) 
| merge x | <other procs>
```
(where `merge x` means "merge in the order of field x").

2. Splitting with `head` and `tail`, with input sorted in x:

`<stateless procs> ... | head N | <stateless procs> | <other procs>` can be transformed into

```
(
  <stateless procs> ... | head N | <other procs>;
  ...                                                                                     
  <stateless procs> ... | head N | <other procs> 
) 
| merge x | head N | <other procs>
```
(this one seems less useful in practice given that for reasonable N, things should be fast without distribution... but tbd).

3. Undoubtedly more!

alfred-landrum:
A comment on case 1 above: the merge of sorted streams can support a form of 'early release' of data, the same way the groupby code does (that @henridf wrote :) ). When we schedule the scanning of input partitions, we'll have the info needed to schedule them in sorted order. They may not finish perfectly in order, of course, but doing so could help reduce the amount of memory the merge needs on average.

henridf:
The two suggestions from the issue description were done in #1022 and #1182 respectively.

PR #1171, opened 2020-08-25
Fix bug in AST parallelization with cut/put/rename

henridf:
Fix issue spotted by @nwt here: https://github.com/brimsec/zq/pull/1145/files/e65c62a62b9cf59e5acff3cde111b23c0029fe3b#r475911797

### DIFF
--- a/driver/compile_test.go
+++ b/driver/compile_test.go
@@ -194,15 +194,27 @@ func TestParallelizeFlowgraph(t *testing.T) {
 			"ts",
 		},
 		{
-			"* | cut x | uniq",
+			"* | cut ts, foo=x | uniq",
 			"ts",
-			"(filter * | cut x; filter * | cut x) | uniq",
+			"(filter * | cut ts, foo=x; filter * | cut ts, foo=x) | uniq",
 			"ts",
 		},
 		{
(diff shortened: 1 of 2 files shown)


henridf:
Note what while the previous version was in a sense eager to parallelize, this one is overly cautious. For example, 

`cut -c ts | count()` will be parallelized into `(filter * ; filter *) | cut -c ts | count()`, whereas it could be parallelized into `(filter *  | cut -c ts | count(); filter * | cut -c ts | count()) | count()`. This seems like candidate material to address in #1034.

driver/compile_test.go
```diff
@@ -270,10 +282,11 @@ func TestParallelizeFlowgraph(t *testing.T) {
 	}
 	// This needs a standalone test due to the presence of a pass
 	// proc in the transformed AST.
-	t.Run("* | cut x | put x=y | rename foo=boo", func(t *testing.T) {
+	t.Run("* | cut ts, y, z | put x=y | rename y=z", func(t *testing.T) {
+		t.Skip()
```
alfred-landrum, line 286: Was this meant to be skipped?

henridf: oops, no! unskipped.

PR #1173, opened 2020-08-25
K8s obs

marktwallace:
This updates the procedures and config files for EKS deployment of zqd.
It also adds tools for observability, including an example Grafana dashboard for zqd.
The PR does not yet expose EKS endpoints, so kubectl port-forward must be used for testing.


k8s/README.md
```diff
@@ -242,143 +263,74 @@ aws ecr create-repository \
...
 ```
-curl -v http://localhost:9867/status
+~/go/bin/zar import -R s3://zqd-demo-1/mark/zeek-logs/dns s3://brim-sampledata/wrccdc/zeek-logs/dns.log.gz
 ```
-You should get an 'ok' response.
 
+## [Trouble shooting](trouble-shooting.md)
 
-# Tearing down the environment
+## Tearing down the environment
```
alfred-landrum, line 331: Would you please move this section up near the other local cluster work, so that its not separated by the AWS/EKS instructions?

k8s/README.md
```diff
@@ -242,143 +263,74 @@ aws ecr create-repository \
...
-```
-Now use zapi to make the zar data available to your running zqd instance. Note that zapi defaults to `localhost:9867` for connecting to zqd.
-```
-zapi new -k archivestore -d s3://brim-scratch/mark/sample-http-zng http-space-1
-```
-The `-d` parameter provides the same s3 dir that we used in the `-R` parameter of the zar command above. This command creates a new space for zqd called `http-space-1`. If we again list the s3 directories with `aws s3 ls brim-scratch/mark --recursive` we will see that zqd has now created a new object under its datauri, a directory name that is a base64 string prefixed with "sp_1g0".
+Work in progess...
 
-Now that you have a zqd running with a space, and you have made an archive from zar data, you can use zapi commands to query the sample data. Example:
+For now just use:
```
alfred-landrum, line 296: FWIW, I was able to spin up a zqd pod, but I wasn't able to get any data back from it, using either Brim or zapi, after the port forwarding. I'll ping you off the PR with the logs, though they generally refer connection reset & timeouts, so perhaps there was some issue related to linkerd?

k8s/README.md
alfred-landrum: Also, we should eksctl teardown instructions.

k8s/trouble-shooting.md
```diff
@@ -0,0 +1,241 @@
+# Troubleshooting
```
alfred-landrum, line 1: For consistency, could you please use `troubleshooting` (one word) for the name of this file & in the line below?

k8s/README.md
```diff
@@ -159,32 +159,54 @@ linkerd dashboard &
...
+https://github.com/google/cadvisor/blob/master/docs/storage/prometheus.md
+
+## Delete the local cluster
+When you no longer need the local cluster, you can remove it with:
+```
+kind delete cluster --name zq-local
+```
 
-NOTE: this EKS procedure is not yet working end-to-end in the k8s branch!
+## Deploying the ZQ daemon to an AWS EKS cluster
```
philrz, line 193: It's kind of a nit that I didn't get to in the previous review pass, but there's places in the readme where all caps like "ZQ" and "ZQD" is used and other places it's lowercase like "zq" and "zqd". I've generally seen us try to standardize on the lowercase representations even in titles, just because they're commands that are typed.

k8s/README.md
```diff
@@ -193,45 +209,50 @@ eksctl version
...
---nodegroup-name standard-workers \
---node-type t3.medium \
---nodes 1 \
---nodes-min 1 \
---nodes-max 3 \
---ssh-access \
---ssh-public-key zq-eks-test.pub \
---managed \
---asg-access
+eksctl create cluster -f k8s/cluster.yaml
```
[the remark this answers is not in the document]
marktwallace: I think editing the file is easiest.

k8s/README.md
marktwallace: Yes on both

k8s/README.md
marktwallace: I think this was cleared up on a subsequent run. Al had more than one occurrence of a non-fully functional EKS cluster being created.

k8s/trouble-shooting.md
marktwallace: Yes

k8s/README.md
marktwallace: Yes, I changed to lower case for all zq and zqd

marktwallace:
Thanks for finding all the types and problems with conventions! I fixed the problems that Al and Phil pointed out and few more that I found on the way. I like Phil's suggestion of adding some environment variables for repeated URIs. I will try that the next time I add new content for K8s.

PR #1175, opened 2020-08-25
unify log/pcap post logic and add warnings to pcap post

mccanne:
This commits reworks the logic on the post-path for pcap files to
look more like the post-path for log files.  They now both use
the the closing of the warnings channel to indicate that the
posting goroutine is done.  This also extends pcap path to have
a warnings channel though it is not used yet (except for indicating
"done").  It will be used in a subsequent PR to send pcap-ng
parse warnings to the zqd client.


mccanne:
Oopsies, I'll fix.

mccanne:
I'm abandoning this approach because there is too much to change here and I'd rather just hack in warnings for pcap-ng.

PR #1176, opened 2020-08-25
remove an unncessary waitgroup in log ingester

mccanne:
This commit removes an unncessary waitgroup from the log ingester
since synchronous was already being provided by the closure of
the warnings channel.  We also added comments for clarification.
This follows more closely the pattern of context.Context.

### DIFF
--- a/zqd/ingest/log.go
+++ b/zqd/ingest/log.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"sync"
 	"sync/atomic"
 
 	"github.com/brimsec/zq/pkg/fs"
@@ -33,7 +32,6 @@ type LogOp struct {
 	err          error
 
 	warningCh chan string
-	wg        sync.WaitGroup
 	zctx      *resolver.Context
 }
 
@@ -70,7 +68,6 @@ func NewLogOp(ctx context.Context, store storage.Storage, req api.LogPostRequest
 		p.readCounters = append(p.readCounters, rc)
 		p.readers = append(p.readers, zr)
 	}
-	p.wg.Add(1)
 	go p.start(ctx, store)
 	return p, nil
 }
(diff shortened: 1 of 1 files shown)


mccanne:
I think a little helper Type that pcap-post and log-post could use for warnings would be nice, emulating the context.Context pattern, e.g., <-Status(), Err(), etc.  It could also handle the warnings-at-open buffer that is then drained by Status().

zqd/ingest/log.go
```diff
@@ -158,7 +154,8 @@ func (p *LogOp) Status() <-chan string {
 	return p.warningCh
 }
 
+// Error indicates what if any error occurred during import, after the
+// Status channel is closed.
```
nwt, line 158: NIt: Maybe state explicitly that this returns an undefined value if the Status channel is still open?

zqd/ingest/log.go
```diff
@@ -158,7 +154,8 @@ func (p *LogOp) Status() <-chan string {
...
 func (p *LogOp) Error() error {
-	p.wg.Wait()
```
mattnibs, line 162: Correct me if I'm wrong but doesn't this make this call no longer thread safe?

mattnibs:
> I think a little helper Type that pcap-post and log-post could use for warnings would be nice, emulating the context.Context pattern, e.g., <-Status(), Err(), etc. It could also handle the warnings-at-open buffer that is then drained by Status(). In a later PR.

Sounds cool, but I'm not quite sure I follow. Is this like a general interface that both ops would implement?

zqd/ingest/log.go
nwt: @mattnibs: It's safe after the Status channel is closed (but not before).

mccanne:
> > I think a little helper Type that pcap-post and log-post could use for warnings would be nice, emulating the context.Context pattern, e.g., <-Status(), Err(), etc. It could also handle the warnings-at-open buffer that is then drained by Status(). In a later PR.
> 
> Sounds cool, but I'm not quite sure I follow. Is this like a general interface that both ops would implement?

I'll put up a PR for review at some point...

PR #1178, opened 2020-08-26
add warnings channel to pcap-ng reader and report through zqd api

### LINKED ISSUE #1117: "truncated input" error on pcap slice operation
A Brim community user reported an issue where they were able to import a pcapng successfully into Brim but not able to extract any flows from it via the Packets button.

Suspecting a problem with the packet indexing/extraction, we debugged at the command line. Per a [Slack thread](https://brimsec.slack.com/archives/C010DR0HHMF/p1597642269030000), we baselined a "happy path" of functionality equivalent to what `zqd` normally does to extract packets. Using a known good sample [hello.pcapng.zip](https://github.com/brimsec/zq/files/5091162/hello.pcapng.zip), to build the packet index:

```
$ ./pcap index -r hello.pcapng -x hello.idx
```

To look at some sample packet timestamps in the pcapng:

```
$ ./pcap ts -r hello.pcapng 
1583768523.826851
1583768523.826857
1583768523.826968
...
```

To extract a "slice" by using some of those timestamps against the index:

```
$ ./pcap slice -x hello.idx -r hello.pcapng -from 1583768523.826851 -to 1583774873.620491 > tmp.pcapng
```

That resulting `tmp.pcapng` is similar to what the daemon should be creating behind the Brim app when you click on the Packets button in Brim, and this slice from the example `hello.pcapng` can, indeed, be opened in Wireshark.

The user completed the equivalent steps with their sample pcap and at the `pcap slice` step saw the failure message `truncated input: truncated input`.

The user's pcapng file unfortunately cannot be shared due to NDA issues. We may need to provide a debug build to gain further insight into why we're not able to work with this pcapng file.

While we're at it, we should better understand if it would have been possible to surface this error in the app, since going into CLI mode to debug is pretty painful. https://github.com/brimsec/brim/issues/1003 is already tracking the app's role in this, but I imagine the backend might have something to add here as well.

mccanne:
This commit works around a bug in certain pcap-ng files where extra
bytes show up at the end of the file that are not a compliant pcap-ng
block.  The fix is to log a warning rather than throw an error.
    
To add warnings, we hacked in some hooks to the pcap-post endpoint.
This should be cleaned up and normalized with log-post endpoint
in a later commit.

The code here should look more like the way the warnings channel is
handled in the log-post endpoint but the tests were fussy and the
fact that all of the handlers are in one source file has promoted
largish handler functions instead of handlers decomposed into
bite sized chunks.  I think we should we fix both of these issues in 
later PRs.

We also need to update the brim app to handle the pcap-post warning messages.

Fixes #1117


pcap/pcapio/reader.go
```diff
@@ -43,7 +51,9 @@ func NewReader(r io.Reader) (Reader, error) {
 	track.Reset()
 	_, err2 := NewNgReader(track)
 	if err2 == nil {
-		return NewNgReader(recorder)
+		r, _ := NewNgReader(recorder)
+		r.SetWarningChan(warningCh)
+		return r, nil
```
nwt, line 56: Nit: Even though err should always be nil, I don't think there's a compelling reason not to capture and return it.
```suggestion
		r, err := NewNgReader(recorder)
		r.SetWarningChan(warningCh)
		return r, err
```

zqd/handlers_zeek_test.go
```diff
@@ -145,6 +145,16 @@ func TestPcapSearchNotFound(t *testing.T) {
 	defer p.cleanup()
 }
 
+func TestPcapPostPcapNgWithExtraBytes(t *testing.T) {
+	p := pcapPost(t, "./testdata/extra.pcapng", testZeekLauncher(nil, nil))
+	defer p.cleanup()
+	t.Run("PcapNgExtra", func(t *testing.T) {
+		require.NoError(t, p.err)
+		warning := p.payloads[1].(*api.PcapPostWarning)
+		assert.Equal(t, warning.Warning, "pcap-ng has extra bytes at eof: 20")
```
nwt, line 154: Testify's protocol puts the actual value after the expected value.
```suggestion
		assert.Equal(t, "pcap-ng has extra bytes at eof: 20", warning.Warning)
```

cmd/pcap/index/command.go
```diff
@@ -65,8 +66,16 @@ func (c *Command) Run(args []string) error {
 	if err != nil {
 		return err
 	}
+	warn := make(chan string)
 	defer f.Close()
```
nwt, line 70: Nit: Don't separate the defer from the constructor it services.
```suggestion
	defer f.Close()
	warn := make(chan string)
```

pcap/pcapio/reader.go
```diff
@@ -34,6 +34,14 @@ type Reader interface {
...
 // NewReader returns a Reader by trying both pcap and pcap-ng formats.
 func NewReader(r io.Reader) (Reader, error) {
+	return NewReaderWithWarnings(r, nil)
+}
+
+// NewReaderWithWarnings returns a Reader by trying both pcap and pcap-ng formats
+// and arranges for warning messages to be sent over the given channel.  Different
+// pcap implementations can have out-of-spec peculiarities that can be tolerated
+// so we send warnings and try to okeep going.
+func NewReaderWithWarnings(r io.Reader, warningCh chan<- string) (Reader, error) {
```
mattnibs, line 44: +1 for creating a new method rather than just tacking `warningCh` as an additional argument to `NewReader`

zqd/api/api.go
```diff
@@ -151,6 +156,7 @@ type PcapPostStatus struct {
 	PcapReadSize  int64      `json:"pcap_read_size" unit:"bytes"`
 	SnapshotCount int        `json:"snapshot_count"`
 	Span          *nano.Span `json:"span,omitempty"`
+	Warning       string     `json:"warning,omitempty"`
```
mattnibs, line 159: I do not believe this is being used, nor do I believe this should go in `PcapPostStatus`.

pcap/pcapio/reader.go
```diff
@@ -34,6 +34,14 @@ type Reader interface {
 
 // NewReader returns a Reader by trying both pcap and pcap-ng formats.
 func NewReader(r io.Reader) (Reader, error) {
+	return NewReaderWithWarnings(r, nil)
+}
+
+// NewReaderWithWarnings returns a Reader by trying both pcap and pcap-ng formats
+// and arranges for warning messages to be sent over the given channel.  Different
+// pcap implementations can have out-of-spec peculiarities that can be tolerated
+// so we send warnings and try to okeep going.
```
nwt, line 43:
```suggestion
// so we send warnings and try to keep going.
```

tests/suite/pcap/extra.yaml
```diff
@@ -0,0 +1,13 @@
+# make sure we can read packets with a buggy caplen greater than the
+# original packet size.  bad-caplen.pcap was exdtracted fromo the pcap
+# in brim/issue https://github.com/brimsec/brim/issues/615
```
nwt, line 3:
```suggestion
# original packet size.  bad-caplen.pcap was extracted from the pcap
# in https://github.com/brimsec/brim/issues/615.
```

zqd/api/api.go
mccanne: Oops, changed the approach and missed that, thanks

PR #1180, opened 2020-08-26
Add Done() handling to ordered merge

### LINKED ISSUE #1170: orderedmerge should call Done on parent when done

henridf:
closes #1170


proc/orderedmerge/orderedmerge.go
```diff
@@ -55,6 +59,9 @@ func (p *Proc) run() {
 					if proc.EOS(batch, err) {
 						return
 					}
+				case _ = <-p.doneCh:
```
nwt, line 62:
```suggestion
				case <-p.doneCh:
```

PR #1187, opened 2020-08-28
Zar import sortmem

marktwallace:
This adds -sortmem flag to zar import.
  
It works as expected in the code, but so far, it only make sense to use the flag for sorting zeek log files of less than 100 MB. For larger files, setting $TMPDIR to spill to disk (Ramdisk or SDD) gives better performance. For zeek files of more than a few GB, the -sortmem option can cause the zar process to run out of memory.

There are a series of PR comments below that I am leaving in to capture the results of multiple tests of this feature.


marktwallace:
It is also worth mentioning that the '-sortmem' approach may not be very resource efficient:

When I tried running on a lower memory instance (a c5d.large with 4GB RAM rather the the 8 GB Arm on the m5d.large) the process ran out of memory. 

So, the -sortmem=2000000000 

Which would use at least 2GB, actually chewed up enough RAM to exceed 4GB. The process completed on the 8GB instance, but I did not measure max process memory.

marktwallace:
I will leave comments on measurements in this PR.

I checked to see it the "chuck size", that is the '-s' flag has an effect on IO wait. It does not appear to be a significant factor. Here are two run with chuck side of 25MB and 90 MB:

[ec2-user@ip-172-31-19-39 ~]$ time ~/go/bin/zar import -s 25MB -R s3://zqd-demo-1/mark/zeek-logs/dns5 s3://brim-sampledata/wrccdc/zeek-logs/dns.log.gz
Doing an external merge sort that spills to disk, MemMaxBytes=134217728

real	17m3.258s
user	2m11.579s
sys	0m17.645s
[ec2-user@ip-172-31-19-39 ~]$ time ~/go/bin/zar import -s 90MB -R s3://zqd-demo-1/mark/zeek-logs/dns5 s3://brim-sampledata/wrccdc/zeek-logs/dns.log.gz
Doing an external merge sort that spills to disk, MemMaxBytes=134217728

real	16m45.278s
user	2m10.630s
sys	0m15.954s
[ec2-user@ip-172-31-19-39 ~]$

marktwallace:
To isolate the IO problem, I ran the same import in stages, using a local ramdisk. This eliminated all the IO wait, and the overall process ran as fast as could be expected given the CPU time required and the 10 Gigabit/sec transfer rate to S3.
This is strong evidence that the IO wait is coming from the S3 access in zar import.

1) Copy the 264 MB S3 object to a local ramdisk:
$ sudo mkdir /mnt/ramdisk
$ sudo mount -t tmpfs -o size=1024m tmpfs /mnt/ramdisk
$ time aws s3 cp s3://brim-sampledata/wrccdc/zeek-logs/dns.log.gz /mnt/ramdisk/dns.log.gz
download: s3://brim-sampledata/wrccdc/zeek-logs/dns.log.gz to ../../mnt/ramdisk/dns.log.gz

real	0m2.391s
user	0m1.772s
sys	0m1.126s

Note that the time to copy is just 2.4 seconds -- this is perfectly in line with the expected performance on this EC2 instance of a 10 Gigabit/s transfer rate from S3.

2) Run the zar import on dns.log.gz, sending the output back to the ramdisk. We use the -sortmem flag to make sure we do not spill to local disk.

$ time ~/go/bin/zar import -s 25MB -sortmem 2000000000 -R /mnt/ramdisk/dns-dir /mnt/ramdisk/dns.log.gz
Doing a single in-memory sort, MemMaxBytes=2000000000

real	1m5.028s
user	1m11.120s
sys	0m2.105s

The real time is slightly less than the CPU time, since some small portion of the user CPU time can run in parallel.

3) Copy the ZNG file back to S3, using the sync command. We omit most of the output here to get straight to the timing:

$ time aws s3 sync /mnt/ramdisk/dns-dir s3://zqd-demo-1/mark/zeek-logs/dns-dir
upload: ../../mnt/ramdisk/dns-dir/20170324/1490386867.002293.zng to s3://zqd-demo-1/mark/zeek-logs/dns-dir/20170324/1490386867.002293.zng
...

real	0m6.087s
user	0m3.221s
sys	0m1.767s

The sync takes longer than the download from s3, which is expected, since the ZNG files are about 50% larger and are divided into multiple directories.

marktwallace:
Isolating this further, based on Al's suggestion, the following run reads from local ramdisk and writes to S3:

$ time ~/go/bin/zar import -s 25MB -sortmem 2000000000 -R s3://zqd-demo-1/mark/zeek-logs/dns-dir2 /mnt/ramdisk/dns.log.gz
Doing a single in-memory sort, MemMaxBytes=2000000000

real	1m11.241s
user	1m16.770s
sys	0m2.763s

It is faster than steps (2) and (3) of the previous run combined.

So the problem seems to be just with the way zar import reads from S3. The write pattern seems fine.

I I had to guess, I would say the our use of the "Range" parameter on the s3.GetObject is problematic.

https://github.com/brimsec/zq/blob/cae3cca70957b7d656bf6a28a1113cc1468af1d1/pkg/s3io/reader.go#L82

	getObj := &s3.GetObjectInput{
		Bucket: aws.String(r.bucket),
		Key:    aws.String(r.key),
		Range:  aws.String(bytesRange(off, len(p))),
	}

We should see how many separate calls we are doing for S3 with a Range parameter. I will investigate that now.

marktwallace:
I added a counter and println to s3io.ReadAt.

Running zar on a small log (about 5 MB), we did nearly 1500 reads from S3:

$ time ~/go/bin/zar import -R s3://zqd-demo-1/mark/sample-http-zng s3://zq-sample-data/zng/http.zng.gz
...
1495 s3io.ReadAt s3.GetObjectInput: zng/http.zng.gz
Doing a single in-memory sort, MemMaxBytes=134217728

real	0m21.333s
user	0m1.655s
sys	0m0.263s

Next I'm going to try running a zq query on the same file to see how often it reads.

marktwallace:
I won't use this branch, but this closed PR will be a record of the testing.

Issue #1091, opened 2020-08-12
Add record buffer management


alfred-landrum:
In a previous system to zq/zqd, we used memory pools for records & batches of records, to minimize garbage collection pressure from the zq inner loop. We should bring that work into the current codebase.

PR #1190, opened 2020-08-28
Fix some zbuf.Batch reference counting bugs

nwt:
A zbuf.Batch implementation that includes memory reuse will panic on tests without this.

Part of #1091.


henridf (approved):
👍

PR #1192, opened 2020-08-30
Parsimonious use of zng.Record.TypeCheck()

### LINKED ISSUE #1181: Consider skipping zng.(*Record).TypeCheck()
Calling `TypeCheck()` on every record is costly and leads to a nearly ~3x slowdown on a simple processing task.  These measurements were made on a zar archive where the dubset year1  conn.log had been imported:

```
$ GOMAXPROCS=1 time zar zq -t "* | tail 1" _
[..]
       95.87 real        92.31 user         2.12 sys
```

Now with the [call to `TypeCheck()`](https://github.com/brimsec/zq/blob/41416357a63be577356f27c7cd1e1cd956a5ca4c/zio/zngio/reader.go#L402) in zngio.Reader commented out:
```
$ GOMAXPROCS=1 time zar zq -t "* | tail 1" _                                                                                                                                                 
[...]
       32.19 real        29.14 user         1.74 sys
```

henridf:
We currently run TypeCheck() on every record read by a zio reader,
which for some workloads adds significant overhead (see #1181). Yet
for data that we've imported and validated along the
way (specifically, zar and zqd), this check is unneeded.

This commit addresses the above by making the TypeCheck() configurable
via a flag in detector.OpenConfig.

zar import and zqd log ingest enable it. zq enables it and also
exposes the knob through a new `-check` CLI option. It is disabled in
other uses, such as `zar zq` or zqd queries.

closes #1181


zio/detector/file.go
```diff
@@ -61,6 +62,25 @@ func OpenFile(zctx *resolver.Context, path string, cfg OpenConfig) (*zbuf.File,
...
+	if rec == nil || err != nil {
+		return rec, err
+	}
+	if err := rec.TypeCheck(); err != nil {
+		return nil, err
+	}
+	return rec, nil
+}
+
+func newCheckReader(r zbuf.Reader) zbuf.Reader {
```
alfred-landrum, line 80: This function isn't used, and I think it's fine to create a checkReader with just `&checkReader{r}`, so suggest deleting it.

zio/detector/file.go
```diff
@@ -61,6 +62,25 @@ func OpenFile(zctx *resolver.Context, path string, cfg OpenConfig) (*zbuf.File,
 	return OpenFromNamedReadCloser(zctx, f, path, cfg)
 }
 
+type checkReader struct {
```
alfred-landrum, line 65: With these changes, we will not be able to perform fast scanning on checkReaders. (We could; a `checkReader` could implement `scanner.ScannerAble` by calling `scanner.NewScanner` on its reader.)

I think that not allowing fast scanning matches what we're trying to accomplish with the check parameter: there's some doubt on the validity of the data source, and if there's no question about that, then you get better performance by asserting so with `-check=false`. I just wanted to raise the issue in case there were other opinions.

zio/detector/file.go
henridf: Oops it's a leftover. Will remove.

zio/detector/file.go
henridf: I hadn't considered that ramification. (Go's type system isn't very helpful in that respect).

I agree with @alfred-landrum's  assessment but am happy to add the scanner.ScannerAble plumbing as outlined above if anyone feels like we need it.

nwt:
I think this out to be a field in zio.ReaderFlags called ZngCheck that applies only to zngio.Reader and .zngScanner. It's of little use when applied to the other readers.

henridf:
@nwt that's what I thought too until I realized that it _is_ of use to other scanners. See for example the tzng test that's added to this PR.

nwt:
I mean to say that I don't think it should be possible to create a non-validating reader for anything except ZNG. What's the scenario in which that would be valuable?

henridf:
Ah, you're saying TypeCheck should remain always-on for other readers. That seems reasonable, I'll make the change.

henridf:
Updated but now this is part of ReaderFlags and exposed in other tools, I'm not sure we want the same default for `zar import` than for `zq`. It seems like for `zar import` we want the default to be "check" while for `zq` (and `zar zq`) we want the default to be "don't check". Any thoughts?

alfred-landrum:
> Updated but now this is part of ReaderFlags and exposed in other tools, I'm not sure we want the same default for `zar import` than for `zq`. It seems like for `zar import` we want the default to be "check" while for `zq` (and `zar zq`) we want the default to be "don't check". Any thoughts?

I think our desired outcome is:
* always check data originating from a non-ZNG format (from @nwt 's suggestion above)
* don't check ZNG data whose provenance is a file inside a zar archive
* check all other ZNG data by default

So that would imply:
* `zar import` & `zq` default to `ZngCheck=True`
* `zar zq` and `zar map` default to `ZngCheck=False`

WDYT?

henridf:
> WDYT?

ILI! (I Like It)

I'll work on that. (Not sure if the different defaults will mean taking the flag out of ReaderFlags, but will figure something out).

zio/zngio/reader.go
```diff
@@ -399,6 +409,9 @@ func (r *Reader) parseValue(id int, b []byte) (*zng.Record, error) {
 		}
 	}
 	rec := zng.NewVolatileRecord(sharedType, b)
+	if !r.check {
+		return rec, nil
+	}
 	if err := rec.TypeCheck(); err != nil {
 		return nil, err
 	}
```
nwt, line 417:
```suggestion
	if r.check {
		if err := rec.TypeCheck(); err != nil {
			return nil, err
		}
	}
```

zio/zngio/reader.go
```diff
@@ -35,18 +35,28 @@ type Reader struct {
 	// mapper to map internal to shared type contexts
 	mapper *resolver.Mapper
 	sos    int64
+	check  bool
+}
+
+type ReaderOpts struct {
+	Size  int
+	Check bool
```
nwt, line 43:
```suggestion
	Check bool
	Size  int
```

zio/detector/reader.go
```diff
@@ -55,7 +55,7 @@ func NewReaderWithConfig(r io.Reader, zctx *resolver.Context, path string, cfg O
 
 	zngErr := match(zngio.NewReader(track, resolver.NewContext()), "zng")
```
nwt, line 56:
```suggestion
	zngErr := match(zngio.NewReaderWithOpts(track, resolver.NewContext(), zngio.ReaderOpts{Check: true}), "zng")
```

henridf: I wondered about that one. On the whole I don't think type checking should be done in format detection - you can correctly identify a format even if a record doesn't typecheck. That said, since the other formats have type checking always-on, we may as well have it on here for consistency.

henridf:
I'm realizing I somewhat mis-framed the question in https://github.com/brimsec/zq/pull/1192#issuecomment-684864208. Right now, we have: 
- `zq` and `zar import` have the zng check flag, defaulting to "true"
- `zar map` and `zar zq` have checking turned off and don't expose the zngcheck at all. 

So for the latter two, do we need to expose `-zngcheck` at all? I'm thinking we don't, but I could be biased by a desire to avoid plumbing readerflags into readers created by chose commands :) 

@alfred-landrum WDYT?

alfred-landrum:
> * `zq` and `zar import` have the zng check flag, defaulting to "true"
> * `zar map` and `zar zq` have checking turned off and don't expose the zngcheck at all.

That seems good to me. If for dev/debug purposes we wanted to run a TypeCheck on a ZNG file inside the archive, we'd just do it manually with `zq` alone.

PR #1193, opened 2020-08-30
improve done and context handling in merge

alfred-landrum:
Ensure that a merge runner won't call Pull on its parent after receiving
a Done call, and use context cancellation to reap runner goroutines.

### DIFF
--- a/proc/merge/merge.go
+++ b/proc/merge/merge.go
@@ -9,6 +9,7 @@ import (
 
 // A Merge proc merges multiple upstream inputs into one output.
 type Proc struct {
+	ctx      *proc.Context
 	once     sync.Once
 	ch       <-chan proc.Result
 	doneCh   chan struct{}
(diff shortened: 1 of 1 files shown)


henridf (approved):

proc/merge/merge.go
```diff
@@ -9,6 +9,7 @@ import (
 
 // A Merge proc merges multiple upstream inputs into one output.
 type Proc struct {
+	ctx      *proc.Context
```
nwt, line 12: @alfred-landrum: Can we change the name of this to pctx? ctx means context.Context everywhere else in the repo.

alfred-landrum:
Other procs like cut, rename, uniq, and others have instances of `ctx
*proc.Context`. It'd be fine by me to make `pctx` is the canonical name for
a `*proc.Context`, but we should update those as well.

nwt:
Ah, #1161 was merged before I could carp about that. I'll open a PR to use pctx everywhere.

nwt:
I opened #1197 to use pctx for *proc.Context consistently.

PR #1203, opened 2020-09-01
Add -sortmem flag to zar import, and a yaml test

marktwallace:
This adds the same -sortmem flag that is used in zq to the zar export command. It writes an error to stderr if the sort space is smaller the 1 MB, which I determined to be a reasonable lower limit. There is a  ztest yaml test with a positive and negative test case for -sortmem.


cmd/zar/import/command.go
```diff
@@ -62,6 +65,13 @@ func (c *Command) Run(args []string) error {
 		return errors.New("zar import: exactly one input file must be specified (- for stdin)")
 	}
 
+	if c.sortMemMaxBytes <= 1048576 {
```
henridf, line 68: I'm not sure this check is worth having given the limit is arbitrary. The number of open files also depends on input size (size / sortmem), so if we really want to prevent "too many open files" we'd have to take into account both quantities.

marktwallace: I considered that, because for zq the condition is >0. However, if the user then chooses "-sortmem 1", the program will terminate in an error, "too many open files". I chose 1 MB empirically -- I tried zar import on a small log (less than 10 MB) and increased the number until it did not crash. The default for -sortmem is 128 MB, so this is not a substantial constraint.

In the case of zar import, we know there are constraining factors. I think giving the user some guidance is better than not. A user may not be able to figure out that making the -sortmem parameter too small would cause the "too many open files" error. Ideally, when zar terminates because it runs out of a resource (RAM, file handles, etc.) the program would provide guidance about how to vary the input parameters to avoid the problem. Reasonable limits on values for command line flags is a step in that direction.

nwt: I'm with @henridf; I don't think this check is worth having.

alfred-landrum: In a discussion with  @marktwallace , he mentioned that one way to address this is to have the cli option take a >0 number of megabytes. That seems like a way forward.

And fyi: https://github.com/brimsec/zq/pull/1210 will impact the "too many open files" errors.

marktwallace: I like the idea of changing the unit to MB. If we do that, I think we should change it for zq also. Are there any objections to that?

henridf: None from me.

nwt: None from me.

marktwallace: I changed the units for sortmem in zq and zar to MiB, and added test for zq.

This is ready for review again.

nwt (approved):
:shipit:

PR #1206, opened 2020-09-02
Only parallelize flowgraphs with aggregations

### LINKED ISSUE #1204: Don't parallelize flowgraph in ways that rely on stable sorting
Ordered merge does not have sufficient information to sort stably: when multiple records with same sort field are present at different upstreams, it does not know what their original order (*) was.

This issue consists of removing flowgraph parallelizations that rely on stable ordered merge for deterministic output.

(*) And in some possible future situations, such as with overlapping chunks, there simply _isn't_ any "original order"... but that will come later.

henridf:
closes #1204

### DIFF
--- a/driver/compile.go
+++ b/driver/compile.go
@@ -53,7 +53,7 @@ func compile(ctx context.Context, program ast.Proc, zctx *resolver.Context, msrc
 
 	var isParallel bool
 	if mcfg.Parallelism > 1 {
-		program, isParallel = parallelizeFlowgraph(ensureSequentialProc(program), mcfg.Parallelism, sortKey, sortReversed)
+		program, isParallel = parallelizeFlowgraph(ensureSequentialProc(program), mcfg.Parallelism)
 	}
 	if !isParallel {
 		mcfg.Parallelism = 1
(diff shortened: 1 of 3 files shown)


driver/compile.go
```diff
@@ -439,69 +439,18 @@ func decomposable(rs []ast.Reducer) bool {
...
-			}
-			for _, c := range p.Clauses {
-				if c.Target == inputSortField {
-					return buildSplitFlowgraph(seq.Procs[0:i], seq.Procs[i:], inputSortField, inputSortReversed, N), true
-				}
-			}
+		case *ast.UniqProc:
+			return seq, false
+		case *ast.SequentialProc, *ast.ParallelProc:
+			return seq, false
```
nwt, line 448:
```suggestion
		case *ast.ParallelProc, *ast.SequentialProc, *ast.UniqProc:
			return seq, false
```

nwt (approved):

henridf:
After an internal discussion we decided not to do this due to the performance implications. We will instead leave these parallelizations in place, and at some point we'll do #1215 to ensure deterministic outputs.

Issue #1167, opened 2020-08-24
archive.Import: too many open files


philrz:
Repro was on macOS with `zar` & `zq` commit `b7023d5`.

I happened to try importing the full 70 GB wrccdc set of Zeek logs. The failure message:

```
$ zq * | zar import -s 25MB -
archive.Import: run failed: open /var/folders/yn/jbkxxkpd4vg142pc3_bd_krc0000gn/T/zq-sort-293329507/248: too many open files
```

This class of problem has resurfaced multiple times in the iterations of software we've written in the past couple years. It's known that the default `ulimit -n` on macOS (or, at least, on whatever my Mac laptop happens to be running) is small compared to what we're accustomed to seeing on Linux systems:

```
$ ulimit -n
256
```

In these situations, the fallback plan has always been that we could just require users to up this setting before they can expect our software to work. However, in the past we've always successfully managed to make our software behave in a way that it doesn't _require_ as many open files to function/perform, and hence has a higher chance of "just working" out-of-the-box on all of our target platforms. Therefore I've opened this issue to see if we can repeat the same yet again.

alfred-landrum:
From the error string, this happens during the time sorting of data that archive.Import uses.

philrz:
Verified in `zq` commit `b2c8356`.

Following the original repro steps, my `zq * | zar import -s 25MB -` of that 70 GB test data set did now run to completion without error. As the import was happening, I watched in another window as the temp space for the disk-based merge sort grew in size and eventually flushed out. During several repros of the failure case, the `too many open files` error would consistently show up right as the temp directory got to 31 GB in size.

```
$ while true; do du -sh zq-sort-607514208/ ; sleep 60; done
394M	zq-sort-607514208/
1.1G	zq-sort-607514208/
1.8G	zq-sort-607514208/
2.4G	zq-sort-607514208/
3.2G	zq-sort-607514208/
4.1G	zq-sort-607514208/
5.0G	zq-sort-607514208/
5.6G	zq-sort-607514208/
6.1G	zq-sort-607514208/
6.8G	zq-sort-607514208/
7.5G	zq-sort-607514208/
8.3G	zq-sort-607514208/
8.9G	zq-sort-607514208/
9.6G	zq-sort-607514208/
 10G	zq-sort-607514208/
 11G	zq-sort-607514208/
 12G	zq-sort-607514208/
 12G	zq-sort-607514208/
 13G	zq-sort-607514208/
 13G	zq-sort-607514208/
[...]

PR #1210, opened 2020-09-03
Raise open files soft limit to hard limit in ZQL commands

nwt:
Addresses #1167.


henridf (approved):
Nice. (I've seen this code somewhere 🙂).

Issue #1217, opened 2020-09-04
Read non-adjacent record fields in Zeek TSV


philrz:
A community user was attempting to read a Zeek TSV log that had been generated by a Corelight Sensor. Sample contents:

```
#separator \x09
#set_separator	,
#empty_field	(empty)
#unset_field	-
#path	conn
#open	2020-08-24-00-25-01
#fields	ts	uid	id.orig_h	id.orig_p	id.resp_h	id.resp_p	proto	service	duration	orig_bytes	resp_bytes	conn_state	local_orig	local_resp	missed_bytes	history	orig_pkts	orig_ip_bytes	resp_pkts	resp_ip_bytes	tunnel_parents	orig_cc	resp_cc	id.orig_h_name.src	id.orig_h_name.vals	id.resp_h_name.src	id.resp_h_name.vals
#types	time	string	addr	port	addr	port	enum	string	interval	count	count	string	bool	bool	count	string	count	count	count	count	set[string]	string	string	string	set[string]	string	set[string]
1598243094.015046	CWjxkd3jpmxuvN21uj	10.124.2.117	61927	10.70.70.70	8080	tcp	-	0.002716	0	77	SF	F	F	0	FdfR	3	120	2	157	-	-	-	-	-	SSL_SNI	oneclient.sfx.ms,bats.video.yahoo.com,ctldl.windowsupdate.com,tapestry.tapad.com,www.gstatic.com,www.google.com:443,c.clicktale.net,eb2.3lift.com:443,13-237-209-96.expertcity.com:443,pr-bh.ybp.yahoo.com:443,clientservices.googleapis.com:443,js-sec.indexww.com:443,collect.tealiumiq.com,www.pacast.com,oneclient.sfx.ms:443,clientservices.googleapis.com,bats.video.yahoo.com:443,www.youtube.com
```

As of `zq` commit `16d510a`, this triggers the following error:

```
$ zq -t -i zeek sample.tsv 
sample.tsv: line 9: duplicate field id
```

The root cause involves zq's internal reconstruction of the Zeek `record` data type into the hierarchical format as it existed when the data originated inside Zeek. One of the assumptions built into the current functionality is that all sub-fields of a record are expected to be adjacent, such as is traditionally seen with `id.orig_h`, `d.orig_p`, `id.resp_h`, and `id.resp_p`. The zq reader is therefore not expecting to see the fields starting rightward from `id.orig_h_name.src` separated by columns representing other fields.

[...]

PR #1218, opened 2020-09-04
Refactor zeekio record construction

nwt:
~~This is groundwork for a #1217 fix.~~

Bugbear from work on #1217


mccanne (commented):
This looks like an improvement.  I'm not too concerned about the performance of this code path as the bulk of deployments will be zeek json or our zng plugin.

mccanne:
And glad we're hopping on this one!

henridf (approved):

PR #1221, opened 2020-09-06
create skeleton concept for package z

mccanne:
Any thoughts on this draft?  I always pull my hair out creating zng.Records from native Go types.  Even programmers don't want to create their schemas ahead of time!

This is inspired by zapper and if we finished it up, it would make it easy to do a zapzng core extension to zapper.

Also, this could be helpful when we get to zng.Marshal(interface{}).

### DIFF
--- a/z/z.go
+++ b/z/z.go
@@ -6,7 +6,7 @@ import (
 	"github.com/brimsec/zq/zng/resolver"
 )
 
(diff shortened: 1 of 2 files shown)


mccanne:
Putting in the icebox for now.  See issue #1347

PR #1227, opened 2020-09-08
cut -c: use descriptor to derive fields

### LINKED ISSUE #1226: "bad uvarint" with cut -c
```
$ cat input.tzng 
#0:record[id:record[ip:ip],id2:record[ip:ip]]
  0:[-;[1.1.1.1;]]
$ zq -t "cut -c id" input.tzng 
bad uvarint: 0
```

This is due to `*fieldIter.Next()`trying to step into an unset record. That's easy to fix, but what `*fieldIter.Next()` should return here is less obvious, since it currently returns leaves only (bug #1223 notwithstanding).

henridf:
Previously, cut -c used a zng.fieldIter on the first record of a type
to compute which fields to extract. This leads to problems with unset
values: if an inner record is unset, fieldIter doesn't visit leaves,
and the cutter gets confused, as observed in https://github.com/brimsec/zq/issues/1226.

Instead, use the record's descriptor to figure out which fields to cut.

closes #1226

### DIFF
--- a/proc/cut/cutter.go
+++ b/proc/cut/cutter.go
@@ -1,8 +1,6 @@
 package cut
 
 import (
-	"strings"
-
 	"github.com/brimsec/zq/expr"
 	"github.com/brimsec/zq/proc"
 	"github.com/brimsec/zq/zng"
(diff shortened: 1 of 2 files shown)


proc/cut/cutter.go
```diff
@@ -78,48 +76,58 @@ func (c *Cutter) FoundCut() bool {
...
-	for _, setel := range set {
-		for j := range splits {
-			prefix := strings.Join(splits[:j+1], ".")
-			if prefix == setel {
-				return true
-			}
+// complementFields returns the slice of fields and associated types
+// that make up the complemente of the set of fields passed as first
+// argument.
+func complementFields(drops []string, prefix string, typ *zng.TypeRecord) ([]string, []zng.Type) {
```
nwt, line 108:
```suggestion
// complementFields returns the slice of fields and associated types
// that make up the complement of the set of fields in drops.
func complementFields(drops []string, prefix string, typ *zng.TypeRecord) ([]string, []zng.Type) {
```

proc/cut/cutter.go
```diff
@@ -78,48 +76,58 @@ func (c *Cutter) FoundCut() bool {
...
 	// fields, then there is no output for this input type.
-	if len(outColTypes) == 0 {
+	if len(fieldTypes) == 0 {
 		return nil, nil
 	}
+
+	for _, f := range fields {
+		resolver := expr.CompileFieldAccess(f)
+		resolvers = append(resolvers, resolver)
+	}
```
nwt, line 91:
```suggestion
	var resolvers []expr.FieldExprResolver
	for _, f := range fields {
		resolvers = append(resolvers, expr.CompileFieldAccess(f))
	}
```

nwt (approved):
:shipit:

PR #1228, opened 2020-09-08
Add support for empty zng records and empty ndjson objects

### LINKED ISSUE #1069: Handling empty JSON objects
I came across this problem while working with [Suricata EVE JSON](https://suricata.readthedocs.io/en/suricata-5.0.3/output/eve/eve-json-output.html) data. Repro is with `zq` commit `07fc77b`.

Consider the following NDJSON:

```
$ cat empty_object.ndjson 
{"ja3s":{}}
```

If read & re-written as NDJSON via `zq`, the original representation is not preserved.

```
$ cat empty_object.ndjson | zq -f ndjson -
{"ja3s":null}
```

Also, if I convert to binary ZNG as an intermediate step, a downstream `zq` sees it as unreadable.

```
$ cat empty_object.ndjson | zq - | zq -i zng -f ndjson -
/dev/stdin: type record: zero columns not allowed
```

On this last point, I suspected this might be expected behavior, since the [Section 3.1.1.1 of the ZNG spec](https://github.com/brimsec/zq/blob/master/zng/docs/spec.md#3111-record-typedef) says:

> A record may not contain zero columns.

However, we'd like to be able to represent any valid NDJSON in ZNG, so it seems we'll want to address this.

FWIW, other JSON tools such as `jq` do not complain.

```
$ cat empty_object.ndjson | jq .
{
  "ja3s": {}
}
```

henridf:
Closes #1069

### DIFF
--- a/tests/suite/ndjson/empty-objects.yaml
+++ b/tests/suite/ndjson/empty-objects.yaml
@@ -1,12 +1,10 @@
 zql: '*'
 
 input: |
-      {}
-      {"ja3s":{}}
+  {}
+  {"ja3s":{}}
 
 output-format: ndjson
-output:
-  |
+output: |
     {}
     {"ja3s":{}}
-
--- a/tests/suite/ndjson/empty-objects.yaml
+++ b/tests/suite/ndjson/empty-objects.yaml
@@ -1,0 +1,12 @@
+zql: '*'
+
+input: |
+      {}
+      {"ja3s":{}}
+
+output-format: ndjson
+output:
+  |
+    {}
+    {"ja3s":{}}
+
--- a/tests/suite/ndjson/empty-objects.yaml
+++ b/tests/suite/ndjson/empty-objects.yaml
@@ -6,5 +6,5 @@ input: |
 
 output-format: ndjson
 output: |
-    {}
-    {"ja3s":{}}
+  {}
+  {"ja3s":{}}
--- a/tests/suite/zjson/empty-records.yaml
+++ b/tests/suite/zjson/empty-records.yaml
@@ -5,8 +5,7 @@ input: |
   {"id":24,"type":[{"name":"ja3s","type":[]}],"values":[[]]}
 
 output-format: zjson
-output:
-  |
+output: |
     {"id":23,"type":[],"values":[]}
     {"id":24,"type":[{"name":"ja3s","type":[]}],"values":[[]]}
 
--- a/tests/suite/zjson/empty-records.yaml
+++ b/tests/suite/zjson/empty-records.yaml
@@ -1,0 +1,12 @@
+zql: '*'
+
+input: |
+  {"id":23,"type":[],"values":[]}
+  {"id":24,"type":[{"name":"ja3s","type":[]}],"values":[[]]}
+
+output-format: zjson
+output:
+  |
+    {"id":23,"type":[],"values":[]}
+    {"id":24,"type":[{"name":"ja3s","type":[]}],"values":[[]]}
+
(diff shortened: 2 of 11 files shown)


tests/suite/zjson/empty-records.yaml
```diff
@@ -0,0 +1,12 @@
...
+
+input: |
+  {"id":23,"type":[],"values":[]}
+  {"id":24,"type":[{"name":"ja3s","type":[]}],"values":[[]]}
+
+output-format: zjson
+output:
+  |
+    {"id":23,"type":[],"values":[]}
+    {"id":24,"type":[{"name":"ja3s","type":[]}],"values":[[]]}
```
nwt, line 11:
```suggestion
output: |
  {"id":23,"type":[],"values":[]}
  {"id":24,"type":[{"name":"ja3s","type":[]}],"values":[[]]}
```

zng/resolver/context.go
```diff
@@ -429,6 +429,15 @@ func (c *Context) parseRecordTypeBody(in string) (string, zng.Type, error) {
...
 	var columns []zng.Column
+	in, ok = match(in, "]")
+	if ok {
+		typ, err := c.LookupTypeRecord([]zng.Column{})
+		if err != nil {
+			return "", nil, err
+		}
+		return in, typ, nil
+	}
+
```
nwt, line 440: Please don't insert code that doesn't use `columns` between its declaration and the code that does use it.

zng/resolver/context.go
```diff
@@ -580,7 +589,7 @@ func (c *Context) TranslateType(ext zng.Type) (zng.Type, error) {
 }
 
 func (c *Context) TranslateTypeRecord(ext *zng.TypeRecord) (*zng.TypeRecord, error) {
-	var columns []zng.Column
+	columns := []zng.Column{}
```
nwt, line 592: This is unnecessary.

zio/zngio/reader.go
```diff
@@ -286,10 +286,7 @@ func (r *Reader) readTypeRecord() error {
 	if err != nil {
 		return zng.ErrBadFormat
 	}
-	if ncol == 0 {
-		return errors.New("type record: zero columns not allowed")
-	}
-	var columns []zng.Column
+	columns := []zng.Column{}
```
nwt, line 289: This is unnecessary.

tests/suite/ndjson/empty-objects.yaml
```diff
@@ -0,0 +1,10 @@
+zql: '*'
+
+input: |
+  {}
+  {"ja3s":{}}
+
+output-format: ndjson
+output: |
+    {}
+    {"ja3s":{}}
```
nwt, line 10: Nit: Indent this block scalar by two spaces as we do elsewhere.
```suggestion
output: |
  {}
  {"ja3s":{}}
```

tests/suite/zjson/empty-records.yaml
```diff
@@ -0,0 +1,11 @@
...
+
+input: |
+  {"id":23,"type":[],"values":[]}
+  {"id":24,"type":[{"name":"ja3s","type":[]}],"values":[[]]}
+
+output-format: zjson
+output: |
+    {"id":23,"type":[],"values":[]}
+    {"id":24,"type":[{"name":"ja3s","type":[]}],"values":[[]]}
+
```
nwt, line 11: Nits: Ditch the extra blank line. Indent this block scalar by two spaces as we do elsewhere.
```suggestion
  {"id":23,"type":[],"values":[]}
  {"id":24,"type":[{"name":"ja3s","type":[]}],"values":[[]]}
```

nwt (approved):

PR #1236, opened 2020-09-10
replace *zio.Writer with zbuf.WriteCloser

### LINKED ISSUE #1235: rework zio design pattern so it's easier to add zst

mccanne:
This commit cleans up the zio design pattern to allow generic
writers to be closed instead of doing a flush followed by reaching
around the writer to close the underlyinig object.

This will make integrating the zst writer easier because it needs
to be closed to finish up writing metadata and a trailer.  I was
fighting with the existing pattern to wire it all up and this change
will make it all straightforward.

The previous flush-based approach had merits but this seems
easier to understand and work with, albeit we had to add Close
stubs to all of the zio writers.

We also mooved zbuf/zng.go to zbuf/zbuf.go as the previous
name was historical and no longer accurate, and deleted
zbuf.WriteFlusher as it is no longer used.

Finally, this refactoring exposed a problem in the fs.ReplaceFile
call in storage filestore, which was doing a storage copy without
any write buffering, so this is fixed.

Fixed #1235

### DIFF
--- a/zio/tableio/writer.go
+++ b/zio/tableio/writer.go
@@ -12,8 +12,8 @@ import (
 	"github.com/brimsec/zq/zng/resolver"
 )
 
-type Table struct {
-	io.Writer
+type Writer struct {
+	writer    io.WriteCloser
 	flattener *zeekio.Flattener
 	table     *tabwriter.Writer
 	typ       *zng.TypeRecord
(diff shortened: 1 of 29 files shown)


zio/tableio/writer.go
```diff
@@ -23,66 +23,74 @@ type Table struct {
...
+func (w *Writer) flush() error {
+	return w.table.Flush()
+}
+
+func (w *Writer) Close() error {
+	firstErr := w.flush()
+	if err := w.writer.Close(); err != nil && firstErr == nil {
+		firstErr = err
+	}
+	return firstErr
```
alfred-landrum, line 95:
```suggestion
	err := w.flush()
	if closeErr := w.writer.Close(); closeErr != nil && err == nil {
		err = closeErr
	}
	return err
```

mccanne: This the same but with different names.  I was using Noah's pattern "firstErr" but I have a slight preference for this since the idiomatic "err" stays in the outer scope.

zio/zngio/writer.go
```diff
@@ -35,6 +37,14 @@ func NewWriter(w io.Writer, flags zio.WriterFlags) *Writer {
 	}
 }
 
+func (w *Writer) Close() error {
+	firstErr := w.flush()
+	if err := w.closer.Close(); err != nil && firstErr == nil {
+		return firstErr
+	}
+	return firstErr
```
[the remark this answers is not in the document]
mccanne: Yes, it was supposed to be 
```
firstErr = err
```
instead of return, but II will change to your pattern (see comment below)

zio/tableio/writer.go
mccanne: btw, microindex.Writer uses the firstErr pattern so we should probably change it to be consistent.  I'll put up another PR.

nwt: Not that you asked, but I like
```go
	err := w.flush()
	if err2 := w.writer.Close(); err == nil {
		err = err2
	}
	return err

```
for this situation.

mccanne: That's where we landed but do you want me to change the var name `closeErr` to `err2`?

mccanne: https://github.com/brimsec/zq/pull/1236/commits/b803091c1ae2f6295d24547cec707eb33804e901

nwt: > That's where we landed but do you want me to change the var name closeErr to err2?

Go with your preference.

zqd/storage/filestore/filestore.go
```diff
@@ -119,15 +120,15 @@ func (s *Storage) Write(ctx context.Context, zctx *resolver.Context, zr zbuf.Rea
 
 	spanWriter := &spanWriter{}
 	if err := fs.ReplaceFile(s.join(allZngFile), 0600, func(w io.Writer) error {
-		fileWriter := zngio.NewWriter(w, zio.WriterFlags{
+		fileWriter := zngio.NewWriter(bufwriter.New(&zio.NoCloser{w}), zio.WriterFlags{
```
alfred-landrum, line 123: 👍

alfred-landrum (approved):
LGTM, would wait to see if @nwt has further comments

zio/zio.go
```diff
@@ -87,3 +64,7 @@ func Extension(format string) string {
 		return ""
 	}
 }
+
+type NoCloser struct{ io.Writer }
+
+func (_ *NoCloser) Close() error { return nil }
```
nwt, line 70: Nit: I think this would be a little nicer if exposed as a function rather than a type, like io/ioutil.NopCloser, but we can do that later.
```suggestion
// NoCloser returns an io.WriteCloser with a no-op Close method wrapping w.
func NoCloser(w io.Writer) io.WriteCloser { return &noCloser{w} }

type noCloser struct{ io.Writer }

func (_ *noCloser) Close() error { return nil }
```

nwt (approved):

zio/zio.go
mccanne: Thanks, I'll make the change before I merge and follow the go pattern by changing the to NopCloser.

PR #1238, opened 2020-09-10
handle file paths with uri escape codes

### LINKED ISSUE #1196: iosrc.URI: Preserve file names with url encoded characters
Steps to reproduce:

1. call iosrc.ParseURI on a file with url encoded characters `c:\data\file%20with%20spaces`
2. On the resultant uri, call Filepath()
3. Observe that the printed value is different than the file name `c:\data\file with spaces`

This will result in all parts of zq/zqd/zar that rely on iosrc.URI being unable to find the file name with url encoded characters in them.

This should be fixed so the actual file name is preserved.

alfred-landrum:
Closes #1196

### DIFF
--- a/pkg/iosrc/uri.go
+++ b/pkg/iosrc/uri.go
@@ -22,12 +22,11 @@ func ParseURI(path string) (URI, error) {
 	if path == "" {
 		return URI{}, nil
 	}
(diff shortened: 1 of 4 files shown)


mccanne (approved):
nice!

PR #1241, opened 2020-09-11
Add test for nanosecond pcap

philrz:
Since https://github.com/brimsec/brim/pull/1069 seeks to make nanosecond pcap one of the supported formats we let in via the app, I figured we should have a test in the backend that does a basic smoke test with the format.


mccanne (approved):

philrz:
@mccanne: As heads-up since you already approved the PR. I've circled back and made the `pcap` command line more specific to more closely match the logic of how the Brim app requests flows. The test as of when you approved it:

```
pcap slice -r pings.pcapnano -x pings.index -from 1599787801.385053844 -to 1599787801.385053844
```

The revised test:

```
pcap slice -r pings.pcapnano -x pings.index -from 1599787801.385053844 -to 1599787801.385053844 -p icmp 10.138.0.44:8 192.168.1.2:0
```

The revised test is also passing in CI, I'll go ahead and merge now.

PR #1243, opened 2020-09-11
workround pcap filtering problem with zeek's truncated timestamps

### LINKED ISSUE #1240: Nanosecond timestamps are preventing flow extraction
They're a bit of a rarity, but as one community user helped us realize, pcapng files with nanosecond timestamps are out there. Consider the attached one containing four ping packets:

[pings.pcapng.zip](https://github.com/brimsec/zq/files/5209599/pings.pcapng.zip)

When viewed in Wireshark, the timestamps have granularity out to 9 digits after the decimal place.

image

However, when run through Zeek, only 6 digits are preserved.

```
$ /usr/local/zeek-3.2.1/bin/zeek -C -r pings.pcapng local

$ cat conn.log | /usr/local/zeek-3.2.1/bin/zeek-cut ts
1599787799.006763
1599787801.385053
1599787804.008506
1599787806.320433
```

This creates a problem because the packet index maintains the full precision, but since the values from the `conn` record are used to seek into that index, they don't match up. For instance, if that capture is imported in to Brim commit `73d51c1` that talks to `zqd` commit `e322b38` and the Packets button is hit to attempt to extract a flow, they see:

image

The REST call being made to `zqd` here:

```
GET /space/sp_1hMvoDrKySuA5bKvmh3RJsybMD5/pcap?ts_sec=1599787801&ts_ns=385053000&duration_sec=0&duration_ns=0&proto=icmp&src_host=10.138.0.44&src_port=8&dst_host=192.168.1.2&dst_port=0 HTTP/1.1
```

The equivalent attempted from the command line:

```
$ pcap slice -r pings.pcapng -x pings.idx -from 1599787801.385053000 -to 1599787801.385053000 10.138.0.44:8 192.168.1.2:0
truncated input; invalid operation: invalid pcap: pcap-ng file is too small to be valid

$ pcap ts -r pings.pcapng 
1599787799.006763706
1599787801.385053844
1599787804.008506334
1599787806.320433812
```

I see the error response shown here was added as part of https://github.com/brimsec/zq/pull/1178. The change in that PR was motivated by symptoms reported from the same community user's pcapng, but now we have a better understanding, maybe the change needs to be revisited.

I did file a Zeek issue https://github.com/zeek/zeek/issues/1163 and got some initial buy-in from Vern that Zeek could surely benefit from offering a way to preserve this granularity. In the meantime, I assume we could work around the problem by effectively treating the timestamps in the packet index as if they're truncated after 6 digits so they'll match the ones we get from Zeek (not rounded: you can see that the timestamp `1599787801.385053844` in Wireshark becomes`1599787801.385053` in Zeek).

mccanne:
Zeek truncates packet timestamps to microseconds so packets
that have nanosecond-precison timestamps can fall outside of
the range of the the implied microsecond-precision time span
of a conn record.  This commit puts a hack in the pcap search
endpoint to round adjust the requested span to account for
the truncation of both the conn timestamp and its duration.

Fixes #1240

### DIFF
--- a/zqd/pcapstorage/pcapstorage.go
+++ b/zqd/pcapstorage/pcapstorage.go
@@ -152,15 +152,21 @@ func (s *Store) NewSearch(ctx context.Context, req api.PcapSearch) (*Search, err
 	s.metaMu.Lock()
 	defer s.metaMu.Unlock()
 	var search *pcap.Search
+	// We add two microseconds to the end of the span as fudge to deal with the
+	// fact that zeek truncates timestamps to microseconds where pcap-ng
+	// timestamps have nanosecond precision.  We need two microseconds because
+	// both the base timestamp of a conn record as well as the durtion time
+	// can be truncated downward.
+	span := nano.NewSpanTs(req.Span.Ts, req.Span.End()+2000)
 	switch req.Proto {
 	case "tcp":
 		flow := pcap.NewFlow(req.SrcHost, int(req.SrcPort), req.DstHost, int(req.DstPort))
-		search = pcap.NewTCPSearch(req.Span, flow)
+		search = pcap.NewTCPSearch(span, flow)
 	case "udp":
 		flow := pcap.NewFlow(req.SrcHost, int(req.SrcPort), req.DstHost, int(req.DstPort))
-		search = pcap.NewUDPSearch(req.Span, flow)
+		search = pcap.NewUDPSearch(span, flow)
 	case "icmp":
-		search = pcap.NewICMPSearch(req.Span, req.SrcHost, req.DstHost)
+		search = pcap.NewICMPSearch(span, req.SrcHost, req.DstHost)
 	default:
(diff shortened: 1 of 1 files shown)


zqd/pcapstorage/pcapstorage.go
```diff
@@ -152,23 +152,29 @@ func (s *Store) NewSearch(ctx context.Context, req api.PcapSearch) (*Search, err
 	s.metaMu.Lock()
 	defer s.metaMu.Unlock()
 	var search *pcap.Search
+	// We add two microseconds to the end of the span as fudge to deal with the
+	// fact that zeek truncates timestamps to microseconds where pcap-ng
+	// timestamps have nanosecond precision.  We need two microseconds because
+	// both the base timestamp of a conn record as well as the durtion time
```
nwt, line 158:
```suggestion
	// both the base timestamp of a conn record as well as the duration time
```

nwt (approved):

PR #1244, opened 2020-09-11
rename reducer/compile as reducer/compiler

### LINKED ISSUE #1162: reducer/compile and proc/compiler should be consistent
This is a simply rename task.  @mccanne votes to change reducer/compile to reducer/compiler.

mccanne:
This commit renames reducer/compile as "compiler" to have consistent
naming convention for compiler abstractons in the zq code base.

Fixes #1162


mccanne:
I just re-read @henridf comment in the issue #1162 and agree we should do this with a top-level directory that merges the reducer/proc compilers.

PR #1245, opened 2020-09-12
change union implementation to conform with zng spec

mccanne:
I noticed this problem when debugging the union column reader for zst.

Union values were being incorrectly encoded by double encoding the
length of the selector integer.  This commit changes things to
conform.

We also noticed a bug in the ndjson reader where the union array
values were being encoded as a sequence selector, value, selector,
value instead of sequence of containers of selector/value pairs.
This has been fixed.


henridf (approved):
Oops

mccanne:
> Oops

I recall you asked me about it and I gave you bad advice.

PR #1248, opened 2020-09-12
simplify proc compiler

### LINKED ISSUE #1169: simplify proc/compiler slice of parents arg to Compile

mccanne:
This commit simplifies the proc compiler so the nodes that can
take multiple parents (sequential and parallel) are both factored out
into separate methods.

Fixes #1169

### DIFF
--- a/proc/compiler/compiler.go
+++ b/proc/compiler/compiler.go
@@ -1,6 +1,7 @@
 package compiler
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/brimsec/zq/ast"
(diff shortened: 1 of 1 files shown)


henridf (approved):
Nice! (All the  `[]proc.Interface{p}` were bothering me as I did this... ). 

Extra credit for making sequential compilation recursive. For loops are so imperative 🙂

mccanne:
> Nice! (All the `[]proc.Interface{p}` were bothering me as I did this... ).
> 
> Extra credit for making sequential compilation recursive. For loops are so imperative 🙂

Ha!  Thanks.  I figured the recursion was easier to understand and the compiler would optimize out the tail recursion.

henridf:
(Fwiw not that it's needed here, but from last I looked it doesn't seem like the Go compiler does tail call optimization).

PR #1249, opened 2020-09-12
cleanly separate flags and options in zio package

### LINKED ISSUE #1250: refactor zio reader/writer options and flags and improve CLI tools re-use of flags

mccanne:
This commit eliminates significant technical debt for how we
manage command flags and configuration options for the zio
readers and writers.  There were multiple patterns fighting
agaist each other.

This commit cleanly separates command flags from zio options
and refactors the CLI tools to re-use many common flags.
For example, you can now set the sortmem parameter and run
profiling from most all of the tools.

This also cleans up the way json type mappings are read and
configured into the reader options.

In the new approach, each zio reader/writer is responsible for
creating an options struct and package zio/options glues them
together.  The only package under zio that knows about command
flags is zio/flags, and its job is to relate flags to configuration
options.

We also deleted the "types" output format rather than make it work
here since it was hokey (my bad) and I think a better approach
is to make proc that helps collect up all the types.

Whether and when we switch over to cobra (or something else), these
changes are orthogonal and still very helpful.

Fixes #1250


philrz:
@mccanne: As long as you're making common flags work across the tools, is now the time to ask about getting a `-version` flag in all of them? 😄

mccanne:
> @mccanne: As long as you're making common flags work across the tools, is now the time to ask about getting a `-version` flag in all of them? 😄

Piece of cake!

mccanne:
@philrz I just pushed commit to make -version common so it's now in all the tools.

cmd/cli/cli.go
```diff
@@ -0,0 +1,155 @@
+package cli
```
nwt, line 1: A non-main package in cmd/ is a bit of a surprise. I think this should go elsewhere.

zio/flags/flags.go
```diff
@@ -0,0 +1,85 @@
...
+	"regexp"
+
+	"github.com/brimsec/zq/zio/ndjsonio"
+	"github.com/brimsec/zq/zio/options"
+	"github.com/brimsec/zq/zio/zngio"
+)
+
+// Reader has the union of all the flags accepted by the different
+// Reader implementations.
+type Reader struct {
```
nwt, line 17: Nit: As in zio/options, calling these types Reader and Writer feels off. I'd call them ReaderFlags and WriterFlags.

cmd/cli/cli.go
```diff
@@ -0,0 +1,155 @@
...
+func IsTerminal(f *os.File) bool {
+	return terminal.IsTerminal(int(f.Fd()))
+}
+
+func runCPUProfile(path string) {
+	f, err := os.Create(path)
+	if err != nil {
+		log.Fatal(err)
+	}
+	pprof.StartCPUProfile(f)
```
nwt, line 144: We never close f, but I think that's fine.

nwt (approved):
My comments are all nits, so I think this is fine to merge.

zio/options/options.go
```diff
@@ -0,0 +1,22 @@
...
+	JSON   ndjsonio.ReaderOpts
+	AwsCfg *aws.Config
+}
+
+type Writer struct {
+	Format string
+	UTF8   bool
+	Text   textio.WriterOpts
+	Zng    zngio.WriterOpts
+}
```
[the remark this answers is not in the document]
mccanne: They can't go in zio because of circular dependency.

cmd/cli/cli.go
mccanne: I'll put it in brimsec/zq/cli

zio/options/options.go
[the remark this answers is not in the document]
mccanne: I am not happy with alternative naming schemes I'm coming up with right now.  I'll merge it this way and make an issue to revisit zio/flags and zio/options.  I guess I don't like zio/zioflags, but maybe I would get used to it.  We can bikeshed a bit and fix it in a later PR.

zio/flags/flags.go
```diff
@@ -0,0 +1,85 @@
+package flags
```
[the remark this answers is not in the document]
mccanne: Agreed.  See issue #1253

cmd/cli/cli.go
mccanne: Good point.  Probably not a problem, but I'll fix.  I think it was that way before.

zio/flags/flags.go
mccanne: Agreed.  See issue #1253

PR #1251, opened 2020-09-13
simplify groupby proc names

mccanne:
Now that the groupby proc lives in its own package, we can simplify
its naming conventions a bit.

### DIFF
--- a/proc/groupby/groupby.go
+++ b/proc/groupby/groupby.go
@@ -18,15 +18,15 @@ import (
 	"github.com/brimsec/zq/zng/resolver"
 )
 
-type GroupByKey struct {
+type Key struct {
 	target string
 	expr   expr.ExpressionEvaluator
 }
 
-type GroupByParams struct {
+type Params struct {
 	inputSortDir int
 	limit        int
-	keys         []GroupByKey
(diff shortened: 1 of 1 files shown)


proc/groupby/groupby.go
```diff
@@ -635,10 +635,10 @@ func (a *Aggregator) readTable(flush, decompose bool) (zbuf.Batch, error) {
 	return zbuf.NewArray(recs), nil
 }
 
-func (a *Aggregator) lookupRowType(row *GroupByRow, decompose bool) (*zng.TypeRecord, error) {
+func (a *Aggregator) lookupRowType(row *Row, decompose bool) (*zng.TypeRecord, error) {
 	// This is only done once per row at output time so generally not a
 	// bottleneck, but this could be optimized by keeping a cache of the
-	// descriptor since it is rare for there to be multiple descriptors
+	// descriptor since it is rare for there to be multiple descriptoras
```
nwt, line 641:
```suggestion
	// descriptor since it is rare for there to be multiple descriptors
```

nwt (approved):

henridf (approved):

PR #1252, opened 2020-09-13
move zng flattener from zeekio to zng/flattener

mccanne:
This commit moves the flattening logic out of the zeekio package
and into its own package under zng since this operation is a
generic zng transformation and is not specific to zio.

This will be helpful also to the csv writer when we add it.

### DIFF
--- a/zng/flattener/flattener.go
+++ b/zng/flattener/flattener.go
@@ -1,4 +1,4 @@
-package zeekio
+package flattener
 
 import (
 	"fmt"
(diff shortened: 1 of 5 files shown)


zng/flattener/flattener.go
```diff
@@ -17,7 +17,7 @@ type Flattener struct {
 // records where the type context of the received records must match the
 // zctx parameter provided here.  Any new type descriptors that are created
 // to flatten types also use zctx.
-func NewFlattener(zctx *resolver.Context) *Flattener {
+func New(zctx *resolver.Context) *Flattener {
```
nwt, line 20: Don't forget to update the doc comment.

nwt (approved):

PR #1256, opened 2020-09-13
fix zqd version breakage

### LINKED ISSUE #1255: /version endpoint stopped working at commit ca0ef53
The changes associated with #1249 seem to have broken the `/version` endpoint on `zqd`.

As of commit `059c003` that came right before it:

```
$ zqd -version
Version: v0.21.0-4-g059c003

$ zqd listen &

{"boomd":"v0.21.0-4-g059c003","zq":"v0.21.0-4-g059c003"}
```

Then at commit `ca0ef53` that's associated with #1249:

```
$ zqd -version
Version: v0.21.0-5-gca0ef53

$ zqd listen &

{"boomd":"","zq":""}
```

mccanne:
This commit fixes a bug introduced by the recent refactoring of
cli flags and zio options.  While we're at it, we added a zapi
version command and added a test to make sure it's working.

We also simplified the format of the version response message.

The brim app doesn't seem to use the /version endpoint so the 
change shouldn't break anything.

Fixes #1255

### DIFF
--- a/tests/suite/zqd/version.yaml
+++ b/tests/suite/zqd/version.yaml
@@ -1,0 +1,12 @@
+script: |
+  source services.sh
+  zapi -h $ZQD_HOST version
+
+inputs:
+  - name: services.sh
+    source: services.sh
+
+outputs:
+  - name: stdout
+    regexp: |
+      v[0-9]*\.[0-9]*.*
--- a/zqd/api/connection.go
+++ b/zqd/api/connection.go
@@ -137,6 +137,29 @@ func (c *Connection) Ping(ctx context.Context) (time.Duration, error) {
 	return resp.Time(), nil
 }
 
+// backward compatibility with old versions of zqd
+type VersionCompat struct {
+	VersionResponse
+	Zqd string `json:"boomd"`
+	Zq  string `json:"zq"`
+}
+
+// Version retrieves the version string from the service.
+func (c *Connection) Version(ctx context.Context) (string, error) {
+	resp, err := c.Request(ctx).
+		SetResult(&VersionCompat{}).
+		Get("/version")
+	if err != nil {
+		return "", err
+	}
+	msg := resp.Result().(*VersionCompat)
+	version := msg.Version
+	if version == "" {
+		version = msg.Zqd
+	}
+	return version, nil
+}
+
 // SpaceInfo retrieves information about the specified space.
 func (c *Connection) SpaceInfo(ctx context.Context, id SpaceID) (*SpaceInfo, error) {
 	path := path.Join("/space", url.PathEscape(string(id)))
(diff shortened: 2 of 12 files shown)


tests/suite/zqd/version.yaml
```diff
@@ -0,0 +1,12 @@
...
+  zapi -h $ZQD_HOST version
+
+inputs:
+  - name: services.sh
+    source: services.sh
+
+outputs:
+  - name: stdout
+    regexp: |
+      v[0-9]*\.[0-9]*.*
```
mattnibs, line 12: I did not know this existed! This is great.

zqd/core.go
```diff
@@ -33,7 +25,10 @@ type Core struct {
 	logger       *zap.Logger
 }
 
-func NewCore(conf Config) (*Core, error) {
+var Version string
+
+func NewCore(conf Config, version string) (*Core, error) {
+	Version = version
```
mattnibs, line 31: Won't this cause weird memory race issues, notably in highly concurrent unit tests? Consider adding version as a value in zqd.Config.

zqd/api/connection.go
```diff
@@ -137,6 +137,29 @@ func (c *Connection) Ping(ctx context.Context) (time.Duration, error) {
 	return resp.Time(), nil
 }
 
+// backward compatibility with old versions of zqd
+type VersionCompat struct {
+	VersionResponse
+	Zqd string `json:"boomd"`
+	Zq  string `json:"zq"`
+}
```
mattnibs, line 145: If brim isn't using the `/version` endpoint can we just get rid of this?

mattnibs (commented):
Just to be sure I would run the brim system tests on this branch. I can help you with this.

zqd/api/connection.go
mccanne: Yes, I wasn't sure what used it so out this here.  We can take it out.  I grepped for /version in the brim source and didn't find it.

zqd/core.go
mccanne: Oops.  Yes, of course.  Will do.

mattnibs (approved):

zqd/handler.go
```diff
@@ -40,8 +41,9 @@ func NewHandler(core *Core, logger *zap.Logger) http.Handler {
 	h.Handle("/space/{space}/subspace", handleSubspacePost).Methods("POST")
 	h.Handle("/search", handleSearch).Methods("POST")
 	h.HandleFunc("/version", func(w http.ResponseWriter, r *http.Request) {
+		version := api.VersionResponse{Version: core.Version}
 		w.Header().Add("Content-Type", "application/json")
-		json.NewEncoder(w).Encode(&Version)
+		json.NewEncoder(w).Encode(&version)
```
nwt, line 46: Nit: Define version closer to its use. Or say `json.NewEncocder(w).Encode(&api.VersionResponse{Version: core.Version})`.

nwt (approved):

PR #79, opened 2019-11-29
add semantic checking of record type with value

mccanne:
This commit adds type checking to the reader path to make sure
the container structure of an incoming value matches its type descriptor.

I commented out some sort tests that have incorrect zson.  I think we should merge
this then fix the commented out tests.

PR #1258, opened 2020-09-14
simplify zng visitor

mccanne:
I was working on updating zng sets and didn't understand why I had 
to change so many places in zng/walk.go so I simplified it.

This commit simplifies the code in zng/walk.go to use a common
function for descending into container types.  The error messages
are a little different and a union record value gets visited now
instead of omitted, but I think that was a bug/oversight.

We also changed zng.RecordVisitor to zng.Visitor since you can now 
called zng.Walk on any type.

### DIFF
--- a/zng/walk.go
+++ b/zng/walk.go
@@ -17,8 +17,8 @@ type Visitor func(typ Type, body zcode.Bytes) error
 // not returned as an error by any function.
 var SkipContainer = errors.New("skip this container")
 
-func Walk(typ Type, body zcode.Bytes, rv Visitor) error {
-	if err := rv(typ, body); err != nil {
+func Walk(typ Type, body zcode.Bytes, visit Visitor) error {
+	if err := visit(typ, body); err != nil {
 		if err == SkipContainer {
 			return nil
 		}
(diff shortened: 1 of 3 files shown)


zng/walk.go
```diff
@@ -195,85 +125,31 @@ func walkUnion(typ *TypeUnion, body zcode.Bytes, rv RecordVisitor) error {
...
+	return Walk(inner, body, visit)
 }
 
-func walkSet(typ *TypeSet, body zcode.Bytes, rv RecordVisitor) error {
+func walkSet(typ *TypeSet, body zcode.Bytes, visit Visitor) error {
 	if body == nil {
 		return nil
 	}
 	inner := AliasedType(InnerType(typ))
-	if IsContainerType(inner) {
```
henridf, line 264: This is a behavior change... can you keep it for the subsequent work where you are lifting the primitive set type restriction, so that this commit remains refactor-only?

henridf (approved):
Tight! (This had evolved from #79 and I remember trying to simplify and not seeing the way 🙂 ).

zng/walk.go
mccanne: Yes, good catch.  This was my intent and missed it here...

PR #1268, opened 2020-09-14
create zst format

### LINKED ISSUE #1219: design (and implement prototype) columnar zng, aka zst

mccanne:
This commit introduces the zst columnar format to zq.  Package zst
provides the core new functionality.  Package zstio is a zio wrapper
for zst to wire it into the zq pipeline.  A new CLI command "zst"
provides a tool for test, debug, and demo.

There is much more work to be done for columnar format but this
should serve as a good foundation to start from.

Fixes #1219

### DIFF
--- a/tests/suite/zst/cut.yaml
+++ b/tests/suite/zst/cut.yaml
@@ -1,0 +1,41 @@
+script: |
+  #zq -i zst -t out.zst
+  #zq  -t in.tzng
+  zst create -o out.zst in.tzng
+  zst cut -t -k a out.zst
+  echo ===
+  zst cut -t -k b out.zst
+  echo ===
+  zst cut -t -k b.c out.zst
+
+inputs:
+  - name: in.tzng
+    data: |
+      #0:record[a:string,b:record[c:int32,d:int32]]
+      0:[hello;[1;2;]]
+      0:[world;[3;4;]]
+      #1:record[junk:string]
+      1:[trick;]
+      0:[goodnight;[5;6;]]
+      0:[gracie;[7;8;]]
+
+outputs:
+  - name: stdout
+    data: |
+      #0:record[a:string]
+      0:[hello;]
+      0:[world;]
+      0:[goodnight;]
+      0:[gracie;]
+      ===
+      #0:record[b:record[c:int32,d:int32]]
+      0:[[1;2;]]
+      0:[[3;4;]]
+      0:[[5;6;]]
+      0:[[7;8;]]
+      ===
+      #0:record[b:record[c:int32]]
+      0:[[1;]]
+      0:[[3;]]
+      0:[[5;]]
+      0:[[7;]]
--- a/tests/suite/zst/cut.yaml
+++ b/tests/suite/zst/cut.yaml
@@ -1,6 +1,4 @@
 script: |
-  #zq -i zst -t out.zst
-  #zq  -t in.tzng
   zst create -o out.zst in.tzng
   zst cut -t -k a out.zst
   echo ===
--- a/tests/suite/zst/empty-file.yaml
+++ b/tests/suite/zst/empty-file.yaml
@@ -1,6 +1,5 @@
 script: |
-  echo -n "" > empty.zst
-  zq -t -i zst empty.zst
+  zq -t -i zst /dev/null
 
 outputs:
   - name: stderr
--- a/tests/suite/zst/empty-file.yaml
+++ b/tests/suite/zst/empty-file.yaml
@@ -1,0 +1,8 @@
+script: |
+  echo -n "" > empty.zst
+  zq -t -i zst empty.zst
+
+outputs:
+  - name: stderr
+    regexp: |
+      .*: zst trailer not found
--- a/zio/detector/file.go
+++ b/zio/detector/file.go
@@ -83,7 +83,10 @@ func OpenParquet(zctx *resolver.Context, path string, opts zio.ReaderOpts) (*zbu
 
 func OpenFromNamedReadCloser(zctx *resolver.Context, rc io.ReadCloser, path string, opts zio.ReaderOpts) (*zbuf.File, error) {
 	var err error
-	r := GzipReader(rc)
+	r := io.Reader(rc)
+	if opts.Format != "zst" {
+		r = GzipReader(rc)
+	}
 	var zr zbuf.Reader
 	if opts.Format == "" || opts.Format == "auto" {
 		zr, err = NewReaderWithOpts(r, zctx, path, opts)
--- a/zio/zstio/reader.go
+++ b/zio/zstio/reader.go
@@ -1,0 +1,25 @@
+package zstio
+
+import (
+	"errors"
+	"io"
+
+	"github.com/brimsec/zq/zng/resolver"
+	"github.com/brimsec/zq/zst"
+)
+
+type Reader struct {
+	zst.Reader
+}
+
+func NewReader(r io.Reader, zctx *resolver.Context) (*Reader, error) {
+	seeker, ok := r.(zst.Seeker)
+	if !ok {
+		return nil, errors.New("zst must be used with a seekable input")
+	}
+	reader, err := zst.NewReaderFromSeeker(zctx, seeker)
+	if err != nil {
+		return nil, err
+	}
+	return &Reader{*reader}, nil
+}
--- a/zio/zstio/reader.go
+++ b/zio/zstio/reader.go
@@ -8,18 +8,10 @@ import (
 	"github.com/brimsec/zq/zst"
 )
 
-type Reader struct {
-	zst.Reader
-}
-
-func NewReader(r io.Reader, zctx *resolver.Context) (*Reader, error) {
+func NewReader(r io.Reader, zctx *resolver.Context) (*zst.Reader, error) {
 	seeker, ok := r.(zst.Seeker)
 	if !ok {
 		return nil, errors.New("zst must be used with a seekable input")
 	}
-	reader, err := zst.NewReaderFromSeeker(zctx, seeker)
-	if err != nil {
-		return nil, err
-	}
-	return &Reader{*reader}, nil
+	return zst.NewReaderFromSeeker(zctx, seeker)
 }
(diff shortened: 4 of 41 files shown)


zst/README.md
```diff
@@ -0,0 +1,552 @@
...
+> additional passes or by writing the output to multiple files then then
+> merging them together (or even leaving the zst object as separate files
+> that can be collectively read and scanned by a zst implementation).
+
+### The Data Section
+
+The data section contains raw data values organized into "segments",
+where a segment is simply a seek offset and byte length relative to the
+data section.  Each segment contains a sequence of
+[primitive-type zng values](https://github.com/brimsec/zq/blob/zst/zng/docs/spec.md#5-primitive-types),
```
henridf, line 147: We'd discussed relaxing this so that segments could have value columns, are you still 👍 on that?

mccanne: This turned out a biit more difficult than I thought.  Devil's in the details.  You need some way to configure the writer where to stop the recursive descent then you need to capture that information in the reassembly records, which then topologically diverge from the origin zng.TypeRecord.  All do-able but not trivial so I thought we could create an issue and get to this later after we have more experience with the library and what we want to do with it.

tests/suite/zst/cut.yaml
```diff
@@ -0,0 +1,41 @@
+script: |
+  #zq -i zst -t out.zst
+  #zq  -t in.tzng
```
nwt, line 3:
```suggestion
```

tests/suite/zst/empty-file.yaml
```diff
@@ -0,0 +1,8 @@
+script: |
+  echo -n "" > empty.zst
+  zq -t -i zst empty.zst
```
nwt, line 3:
```suggestion
  zq -t -i zst /dev/null
```

zio/detector/file.go
```diff
@@ -79,7 +79,10 @@ func OpenParquet(zctx *resolver.Context, path string, opts zio.ReaderOpts) (*zbu
 
 func OpenFromNamedReadCloser(zctx *resolver.Context, rc io.ReadCloser, path string, opts zio.ReaderOpts) (*zbuf.File, error) {
 	var err error
-	r := GzipReader(rc)
+	r := io.Reader(rc)
+	if opts.Format != "zst" {
+		r = GzipReader(rc)
+	}
```
nwt, line 85: Why special-case this?

zio/zstio/reader.go
```diff
@@ -0,0 +1,25 @@
...
+
+type Reader struct {
+	zst.Reader
+}
+
+func NewReader(r io.Reader, zctx *resolver.Context) (*Reader, error) {
+	seeker, ok := r.(zst.Seeker)
+	if !ok {
+		return nil, errors.New("zst must be used with a seekable input")
+	}
```
nwt, line 19: Nit: This feels wrong. Why not make the first parameter a zst.Seeker and leave the check to the caller?

zio/zstio/reader.go
```diff
@@ -0,0 +1,25 @@
...
+	reader, err := zst.NewReaderFromSeeker(zctx, seeker)
+	if err != nil {
+		return nil, err
+	}
+	return &Reader{*reader}, nil
```
nwt, line 24: Nit: This dereference is a little odd. Is there any reason the embedded zst.Reader field shouldn't just be a *zst.Reader?

zio/zstio/writer.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+func NewWriter(w io.WriteCloser, opts WriterOpts) *Writer {
+	skewthresh := MibToBytes(opts.SkewThresh)
+	colthresh := MibToBytes(opts.ColumnThresh)
+	//XXX should handle error, but zio API doesn't have this yet...
+	// this is just checking bounds on the threshholds so maybe we
+	// do this from above?
+	writer, _ := zst.NewWriter(w, skewthresh, colthresh)
+	return &Writer{*writer}
```
nwt, line 34: Nit: This dereference is a little odd. Is there any reason the embedded zst.Reader field shouldn't just be a *zst.Reader?

zst/README.md
```diff
@@ -0,0 +1,552 @@
...
+together as a single zst entity.  While the zst format provides much flexibility
+for how data is laid out, it's up to implementations to layout data
+in intelligent ways for efficient sequential read access in spite of occasional
+seeks.
+
+## The "Column Stream" Abstraction
+
+The zst data abstraction is built around a collection of "column streams".
+
+There is one column stream for each field of zng record type definition
```
mattnibs, line 100:
```suggestion
There is one column stream for each field of a zng record type definition
```

zst/README.md
```diff
@@ -0,0 +1,552 @@
...
+that appears collectively throughout the zng input data including records
+embedded within records.
+
+For example, a field that is an array of records containing other fields would
+have a column stream for the top-level array field as well as column a stream
```
mattnibs, line 105:
```suggestion
have a column stream for the top-level array field as well as a column stream
```

zst/README.md
henridf: 👍 I was still hoping it might "just work" but if not deferring makes sense

zst/README.md
```diff
@@ -0,0 +1,552 @@
...
+(via [Thrift](https://thrift.apache.org/)) to describe
+analogous data structures, we simply reuse zng here.
+
+> So, we are using zng to encode zng in column format.  Yo dawg.
+
+#### The Schema Definitions
+
+This reassembly zng stream encodes 2*N+1 zng records, where N is equal to the number
+of top-level zng record types that are present in the encoded input.
+To simply terminology, we call a top-level zng record type a "schema",
```
mattnibs, line 218:
```suggestion
To simplify terminology, we call a top-level zng record type a "schema",
```

zio/detector/file.go
mccanne: The zst reader uses type inference to pull the seeker interface out of the reader and wrapping the gzip reader hides the underlying interface.  It would be clearer to maybe have a method `zio.MustBeSeekable(opts.Format)` and wrap the gzip reader only when this is false.

zio/zstio/reader.go
mccanne: I had it the other way initially and it felt wrong that way :-). There are a few places that create a zstio.NewReader and lifting the code out to those places seemed wrong.  This zst.NewReader takes a Seeker, but it seems like this was ok to do in zstio.NewReader.  I think to get this right, we need to refactor detector a bit.  Maybe fix in a subsequent PR?

zio/zstio/reader.go
mccanne: Because it wraps the struct rather than the interface... it makes sense to me.  Why have an extra pointer when you don't need?

zio/zstio/writer.go
mccanne: Same as abovoe

zio/zstio/reader.go
mccanne: The difference is that the zst reader implementation lives in zq/zst.  Perhaps it should move to zstio then there is only one reader and zst.Object/Assembler live in zq/zst?

mccanne: Hmm, the dissonance here is that I'm wrapping the struct when I don't have to and can just return a zst.Reader (and zst.Writer) since they implement zbuf.Reader (and zbuf.WriteCloser).  I'll fix.

zio/zstio/writer.go
mccanne: WIll return zst.Reader instead of zstio.Reader and get rid of latter

zio/flags.go
```diff
@@ -81,4 +81,6 @@ func (w *WriterFlags) SetFlags(fs *flag.FlagSet) {
 	fs.IntVar(&w.Zng.StreamRecordsMax, "b", 0, "limit for number of records in each ZNG stream (0 for no limit)")
 	fs.IntVar(&w.Zng.LZ4BlockSize, "znglz4blocksize", zngio.DefaultLZ4BlockSize,
 		"LZ4 block size in bytes for ZNG compression (nonpositive to disable)")
+	fs.Float64Var(&w.Zst.ColumnThresh, "coltresh", 5, "minimum frame size (MiB) used for zst columns")
+	fs.Float64Var(&w.Zst.SkewThresh, "skewtresh", 25, "minimum skew size (MiB) used to group zst columns")
```
mattnibs, line 85: In `zar import` @alfred-landrum introduced `github.com/alecthomas/units` which allows us to do things like this:
```golang
	f.StringVar(&c.thresh, "s", units.Base2Bytes(archive.DefaultLogSizeThreshold).String(), "target size of chunk files, as '10MB' or '4GiB', etc.")
```
I would suggest using this here. Also helps to have consistency with how we specify data size with all our command line tools.

philrz: 👍  to this idea! I'd recently made the dumb mistake of typing `-sortmem 100000` when I thought I'd typed `-sortmem 10000` and all the while of course I was _trying_ to ask for 10 GiB, and hilarity ensued. I doubt I would have made the same mistake if I'd been able to explicitly request 10 GiB.

mccanne: Ok, I'll put up a PR that addresses this (units.Base2Bytes should involved a flag.Var) as well as #1279.  The flags stuff needs another refactor to kill two birds here.

mccanne: See PR #1323

zio/zstio/writer.go
```diff
@@ -0,0 +1,27 @@
...
+)
+
+const (
+	DefaultColumnThresh = 5 * 1024 * 1024
+	DefaultSkewThresh   = 25 * 1024 * 1024
+)
+
+type WriterOpts struct {
+	ColumnThresh float64
+	SkewThresh   float64
```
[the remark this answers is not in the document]
mccanne: Ok, this is now units.Bytes

zio/flags.go
mccanne: Done

alfred-landrum (approved):
:shipit: A great start to a lot of future work :)

Issue #882, opened 2020-06-09
zapi command for archive index search


alfred-landrum:
zapi needs a subcommand that can perform an archive index search, using the api introduced in https://github.com/brimsec/zq/pull/833 .

philrz:
Verified in `zq` commit `7f00817`. I can now run through the steps for creating & using indexes as shown in the [`zar` README](https://github.com/brimsec/zq/blob/master/cmd/zar/README.md), but now using `zapi` to talk to a `zqd` that was started by my Brim app.

```
$ zq zng/*.gz | zar import -s 25MB -
$ zapi new -k archivestore -d $ZAR_ROOT MyArchive
MyArchive: space created

$ zapi -s MyArchive index create :ip
$ zapi -s MyArchive index find :ip=10.10.23.2 | zq -t -
#zfile=string
#0:record[key:ip,count:uint64,_log:zfile]
0:[10.10.23.2;1;20180324/1521911772.980384.zng;]

$ zapi -s MyArchive index create uri
$ zapi -s MyArchive index find uri=/file | zq -t -
#zfile=string
#0:record[key:bstring,count:uint64,_log:zfile]
0:[/file;1;20180324/1521912507.399929.zng;]
0:[/file;2;20180324/1521912075.114273.zng;]
0:[/file;1;20180324/1521911772.980384.zng;]

$ zapi -s MyArchive index create -o custom.zng -k id.orig_h -z "count() by _path, id.orig_h | sort id.orig_h"
$ zapi -s MyArchive index find -x custom.zng 10.164.94.120 | zq -f table "count=sum(count) by _path" -
_PATH       COUNT
rdp         4116
notice      35
dce_rpc     2
smb_mapping 65
dpd         24
smb_files   1
dns         8
http        13485
ssl         9538
ftp         93
conn        26726
rfb         3
smtp        1178
ntlm        80
weird       316
ssh         1
```

Here's what the archive tree looks like populated with the microindexes that were created via the `zapi` commands:

```
$ tree -s logs
logs
├── [        320]  20180324
│   ├── [    4425007]  1521911772.980384.zng
│   ├── [        160]  1521911772.980384.zng.zar
│   │   ├── [       6347]  custom.zng
│   │   ├── [      97012]  microindex-field-uri.zng
│   │   └── [      29766]  microindex-type-ip.zng
│   ├── [   25001925]  1521912075.114273.zng
│   ├── [        160]  1521912075.114273.zng.zar
[...]

PR #1272, opened 2020-09-14
zqd: Add index post endpoint

mattnibs:
Part of #882

### DIFF
--- a/zqd/handlers.go
+++ b/zqd/handlers.go
@@ -17,6 +17,7 @@ import (
 	"github.com/brimsec/zq/zqd/ingest"
 	"github.com/brimsec/zq/zqd/search"
 	"github.com/brimsec/zq/zqd/space"
+	"github.com/brimsec/zq/zqd/storage/archivestore"
 	"github.com/brimsec/zq/zqe"
 	"github.com/gorilla/mux"
 	"go.uber.org/zap"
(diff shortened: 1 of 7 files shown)


mccanne:
What about some tests and zapi client-side support?  The ztest stuff you set up is really easy and it would help understand the addition and play around with it.   We should work on zql syntax for index searches and having it all wire-able with zapi will help internalize it all.

mattnibs:
> What about some tests and zapi client-side support? The ztest stuff you set up is really easy and it would help understand the addition and play around with it. We should work on zql syntax for index searches and having it all wire-able with zapi will help internalize it all.

Definitely. I am working on a follow up issue to this that will include zapi commands for index find and create.

zqd/handlers.go
```diff
@@ -473,6 +474,34 @@ loop:
...
+
+	store, ok := s.Storage().(*archivestore.Storage)
+	if !ok {
+		respondError(c, w, r, zqe.E(zqe.Invalid, "space storage does not support creating indexes"))
+		return
+	}
+	if err := store.IndexCreate(ctx, req); err != nil {
+		respondError(c, w, r, err)
+		return
+	}
```
alfred-landrum, line 502:
```suggestion
	}
	w.WriteHeader(http.StatusNoContent)
```

alfred-landrum (approved):

zqd/storage/archivestore/archivestore.go
```diff
@@ -101,6 +104,34 @@ func (s *Storage) Write(ctx context.Context, zctx *resolver.Context, zr zbuf.Rea
 	return archive.Import(ctx, s.ark, zctx, zr)
 }
 
+func (s *Storage) IndexCreate(ctx context.Context, req api.IndexPostRequest) error {
+	var rules []archive.Rule
+	if req.AST != nil {
+		proc, err := ast.UnpackMap(nil, req.AST)
```
mattnibs, line 110:
```suggestion
		proc, err := ast.UnpackJSON(nil, req.AST)
```

PR #1280, opened 2020-09-15
add csv option to search endpoint

mccanne:
This commit adds a csv option to the search endpoint enabled by
setting the "format" query parameter on the search to "csv".

We also fixed a small problem in "zapi get" so that "-e csv"
will work properly.

### DIFF
--- a/zio/csvio/writer.go
+++ b/zio/csvio/writer.go
@@ -43,7 +43,7 @@ func (w *Writer) Close() error {
 
 func (w *Writer) Flush() error {
 	w.encoder.Flush()
-	return nil
+	return w.encoder.Error()
 }
 
 func (w *Writer) Write(rec *zng.Record) error {
--- a/zio/csvio/writer.go
+++ b/zio/csvio/writer.go
@@ -41,6 +41,11 @@ func (w *Writer) Close() error {
 	return w.writer.Close()
 }
 
+func (w *Writer) Flush() error {
+	w.encoder.Flush()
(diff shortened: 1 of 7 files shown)


zio/csvio/writer.go
```diff
@@ -41,6 +41,11 @@ func (w *Writer) Close() error {
 	return w.writer.Close()
 }
 
+func (w *Writer) Flush() error {
+	w.encoder.Flush()
+	return nil
```
nwt, line 46:
```suggestion
	return w.encoder.Error()
```

zqd/search/csv.go
```diff
@@ -0,0 +1,62 @@
...
+	response http.ResponseWriter
+	writer   *csvio.Writer
+}
+
+func NewCSVOutput(response http.ResponseWriter, ctrl bool) *CSVOutput {
+	o := &CSVOutput{
+		response: response,
+		writer:   csvio.NewWriter(zio.NopCloser(response), true, false),
+	}
+	return o
```
nwt, line 24:
```suggestion
	return &CSVOutput{
		response: response,
		writer:   csvio.NewWriter(zio.NopCloser(response), true, false),
	}
```

nwt (approved):

zio/csvio/writer.go
mccanne: Have to call csv.Writer.Error().  Will do.

PR #1283, opened 2020-09-16
add ndjson option to search endpoint

### LINKED ISSUE #1275: add vanilla ndjson to search endpoint
Add an ndjson writer path to the search endpoint that is used when the format in the API request is "ndjson".  The client currently requests "zjson" so this won't break anything.

Change the response mime type for in the current zjson responder to be application/x-szjson (i.e,. steaming zjson).

Given the clean design of zqd/search, this task should be trivial by following the pattern of the zng responder using an ndjsonio.Writer and dropping control messages.

mccanne:
This commit adds an ndjson option to the search endpoint enabled by
setting the "format" query parameter on the search to "ndjson".

Fixes #1275

### DIFF
--- a/zqd/search/ndjson.go
+++ b/zqd/search/ndjson.go
@@ -11,7 +11,7 @@ import (
 )
 
 // NDJSONOutput implements the Output inteface and writes NDJSON encoded-output
-// directly to the client as text/NDJSON.
+// directly to the client as application/x-ndjson..
 type NDJSONOutput struct {
 	response http.ResponseWriter
 	writer   *ndjsonio.Writer
--- a/zqd/search/ndjson.go
+++ b/zqd/search/ndjson.go
@@ -1,0 +1,65 @@
+package search
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/brimsec/zq/pkg/bufwriter"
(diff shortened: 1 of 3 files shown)


zqd/search/ndjson.go
```diff
@@ -0,0 +1,65 @@
...
+	"net/http"
+
+	"github.com/brimsec/zq/pkg/bufwriter"
+	"github.com/brimsec/zq/zbuf"
+	"github.com/brimsec/zq/zio"
+	"github.com/brimsec/zq/zio/ndjsonio"
+)
+
+// NDJSONOutput implements the Output inteface and writes NDJSON encoded-output
+// directly to the client as text/NDJSON.
```
henridf, line 14: is `text/NDJSON` what Go figures out via `http.DetectContentType`And should we explicitly set the content type as `application/x-ndjson`?

henridf: Oops never mind I hadn't seen the `ContentType()` method below

henridf, line 14:
```suggestion
// directly to the client as application/x-ndjson.
```

henridf (approved):

PR #1285, opened 2020-09-16
add json option to search endpoint

### LINKED ISSUE #1277: add non-streaming zjson search endpoint
Add an option to the search point to return the results of a search as a single (non-chunked) http response.

All but the the final stats updates should be ignored. 

The mime type of the response should be application/json.

Given the clean design of zqd/search, this task should be trivial by collecting up all the data objects into a go slice then calling the encoding the slice as json onto the http response.

mccanne:
This commit renames the old json search endpoint "zjson" and adds
a new json format option that returns a valid, one-shot json
representation of query results.  This is enabled by
setting the "format" query parameter on the search to "json".

Fixes #1277

### DIFF
--- a/zqd/search/search.go
+++ b/zqd/search/search.go
@@ -28,7 +28,7 @@ const StatsInterval = time.Millisecond * 500
 
 const (
 	MimeTypeCSV    = "text/csv"
-	MimeTypeJSON   = "application/x-ndjson"
+	MimeTypeJSON   = "application/json"
 	MimeTypeNDJSON = "application/x-ndjson"
 	MimeTypeZJSON  = "application/x-zjson"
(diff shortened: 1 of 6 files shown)


zqd/search/search.go
```diff
@@ -28,7 +28,9 @@ const StatsInterval = time.Millisecond * 500
 
 const (
 	MimeTypeCSV    = "text/csv"
+	MimeTypeJSON   = "application/x-ndjson"
```
henridf, line 31: Should this be be `application/json` ?

mccanne: That's what was in my head but not in my typing fingers.  :-)

henridf (approved):

PR #1289, opened 2020-09-16
zapi index create|find commands

### LINKED ISSUE #882: zapi command for archive index search

mattnibs:
Add zar index related commands, allowing users to create and query
indices in archive backed spaces.

Closes #882

### DIFF
--- a/cmd/zapi/cmd/index/create.go
+++ b/cmd/zapi/cmd/index/create.go
@@ -24,7 +24,7 @@ A pattern is either a field name or a ":" followed by a zng type name.
 For example, to index the all fields of type port and the field id.orig_h,
 you would run:
 
-	zapi index create id.orig_h:port
+	zapi index create id.orig_h :port
 
 Each pattern results in a separate microindex file for each log file found.
 
--- a/cmd/zapi/cmd/index/create.go
+++ b/cmd/zapi/cmd/index/create.go
@@ -26,7 +26,7 @@ you would run:
 
 	zapi index create id.orig_h :port
 
-Each pattern results in a separate microindex file for each log file found.
+Each pattern results in a separate, single-key microindex file for each log file found.
 
 For custom indexes, zql can be used instead of a pattern. This
 requires specifying the key and output file name. For example:
--- a/cmd/zapi/cmd/index/create.go
+++ b/cmd/zapi/cmd/index/create.go
@@ -1,0 +1,97 @@
+package idx
+
+import (
+	"encoding/json"
+	"errors"
+	"flag"
+	"strings"
+
+	"github.com/brimsec/zq/cmd/zapi/cmd"
+	"github.com/brimsec/zq/zqd/api"
+	"github.com/brimsec/zq/zql"
+	"github.com/mccanne/charm"
+)
+
+var Create = &charm.Spec{
+	Name:  "create",
+	Usage: "create [options] [-z zql] [ pattern [ pattern ...]]",
+	Short: "create index on a space",
+	Long: `
+"zapi index create" creates index files in a zar archive using one or more indexing
+rules.
+
+A pattern is either a field name or a ":" followed by a zng type name.
+For example, to index the all fields of type port and the field id.orig_h,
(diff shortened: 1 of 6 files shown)


cmd/zapi/cmd/index/create.go
```diff
@@ -0,0 +1,97 @@
...
+	Short: "create index on a space",
+	Long: `
+"zapi index create" creates index files in a zar archive using one or more indexing
+rules.
+
+A pattern is either a field name or a ":" followed by a zng type name.
+For example, to index the all fields of type port and the field id.orig_h,
+you would run:
+
+	zapi index create id.orig_h:port
```
mccanne, line 27:
```suggestion
	zapi index create id.orig_h :port
```

cmd/zapi/cmd/index/create.go
```diff
@@ -0,0 +1,97 @@
...
+
+Each pattern results in a separate microindex file for each log file found.
```
mccanne, line 29:
```suggestion
Each pattern results in a separate, single-key microindex file for each log file found.
```

cmd/zapi/cmd/index/create.go
```diff
@@ -0,0 +1,97 @@
...
+
+For custom indexes, zql can be used instead of a pattern. This
+requires specifying the key and output file name. For example:
+
+    zapi index create -k id.orig_h -o custom -z "count() by _path, id.orig_h | sort id.orig_h"
+`,
```
mccanne, line 35:
```suggestion
`,
Multiple keys may be specified with multiple -k arguments, in which case the first key is the primary search key, the second key is the secondary search key, and so forth.  For example,

zapi index create -k id.orig_h -k count -o custom -z "count() by _path, id.orig_h | sort id.orig_h,count"

```

cmd/zapi/cmd/index/create.go
```diff
@@ -0,0 +1,97 @@
...
+	inputFile  string
+	outputFile string
+	keys       arrayFlag
+	zql        string
+}
+
+func NewCreate(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
+	c := &CreateCmd{Command: parent.(*IndexCmd).Command}
+	f.Var(&c.keys, "k", "key fields (can be specified multiple times)")
+	f.StringVar(&c.inputFile, "i", "", "input file relative to each zar directory ('' means archive log file in the parent of the zar directory)")
```
mccanne, line 55: I'm not a fan of "" for the parent file.  If we don't like "_", maybe we can come up with something better?  "."?  "THIS"?  Unix find uses "{}" inside of exec.  I'm gonna noodle on this.

cmd/zapi/cmd/index/create.go
```diff
@@ -0,0 +1,97 @@
...
+	if err != nil {
+		return err
+	}
+	req := api.IndexPostRequest{
+		Patterns:   args,
+		InputFile:  c.inputFile,
+		OutputFile: c.outputFile,
+	}
+	if c.zql != "" {
+		proc, err := zql.ParseProc(c.zql)
```
mccanne, line 75: cool, so easy!

mccanne (approved):

cmd/zapi/cmd/index/create.go
mattnibs: I have found it confusing that you need to know about using `_` to do the default behavior you would expect. That said, if we're going to stick with some special character we should probably add a constant somewhere to the code base to make this easier to see/understand.

cmd/zapi/cmd/index/create.go
mattnibs: 💥

cmd/zapi/cmd/index/create.go
mattnibs: Also should we expect the `_` be sent over the api call as well?

PR #1290, opened 2020-09-16
add a convenience function to apply zql and copy to single writer

alfred-landrum:


mccanne (approved):

henridf:
Nice! I've been wanting to have something like this for a while. (I'd thought of a slightly different approach: add func to compile a zql and zbuf.Reader into a new zbuf.Reader. Then we could do `zbuf.Copy(dst, flowgraph)` without having to set up and connect all the driver/mux machinery. But this works too).

Just a name thing - Transform seems generic. Maybe `Pipe()` or `Copy()`?

alfred-landrum:
I'd be fine with Copy; I nearly called it that, will change.

FYI, I added this because I need to do another sort in the upcoming zar
change work.

henridf (approved):

PR #1292, opened 2020-09-16
Add context.Context args interface iosrc.Source

### LINKED ISSUE #1108: add context to iosrc calls
We need to plumb in context.Context into package iosrc calls.

mattnibs:
Currently this change only applies to the s3 Source.

Adjust the various public methods of s3io to require a Context in
the first argument.

Closes #1108

### DIFF
--- a/zio/detector/file.go
+++ b/zio/detector/file.go
@@ -31,6 +31,10 @@ const StdinPath = "/dev/stdin"
 // URL. If the path is neither of these or can't otherwise be opened,
 // an error is returned.
 func OpenFile(zctx *resolver.Context, path string, opts zio.ReaderOpts) (*zbuf.File, error) {
+	return OpenFileWithContext(context.Background(), zctx, path, opts)
+}
+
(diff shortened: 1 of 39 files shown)


zio/detector/file.go
```diff
@@ -31,6 +31,10 @@ const StdinPath = "/dev/stdin"
 // URL. If the path is neither of these or can't otherwise be opened,
 // an error is returned.
 func OpenFile(zctx *resolver.Context, path string, opts zio.ReaderOpts) (*zbuf.File, error) {
+	return OpenFileWithContext(context.Background(), zctx, path, opts)
+}
+
+func OpenFileWithContext(ctx context.Context, zctx *resolver.Context, path string, opts zio.ReaderOpts) (*zbuf.File, error) {
```
nwt, line 37: Nit: Maybe add the context parameter to OpenFile and other existing constructors instead of adding WithContext variants?

nwt (approved):

zio/detector/file.go
mattnibs: I went back and forth on this. I suppose it's a matter of taste, but all the `context.Background()` calls for simple command line tool and tests got to me.

nwt: I get it.

PR #1294, opened 2020-09-16
Suricata launcher

### LINKED ISSUE #1214: zqd: run suricata and ingest output

henridf:
This PR adds support for running Suricata as part of the pcap ingest process and importing the resulting eve.json alongside the zeek logs. 

While reviewing, please keep in mind that this is just the first step of the overall feature. Notably still missing are (a) packaged suricata binaries that can be installed via the makefile like Zeek and (b) proper conversion of eve.json including types and renaming the timestamp field. But all that said, this does work and add Suricata events, which can be verified by running zqd manually and posting a pcap with zapi on a system with suricata. 

In Linux CI, pcap ingest tests exercise this code path as of this commit. This still of course deserves specific test coverage, but I would prefer to add that after the eve.json processing mentioned above has settled.

close #1214


.github/workflows/ci.yaml
```diff
@@ -17,6 +17,11 @@ jobs:
       - run: go mod download
       - run: make fmt
       - run: make tidy
+      - run: |
```
henridf, line 20: If/when we move to hosted `.zip` binaries like we do for Zeek, we can remove this and install into `zq/bin` via the Makefile (like for Zeek).

.gitignore
```diff
@@ -1,4 +1,6 @@
-bin/
+!bin/
+bin/*
+!bin/suricata/
```
henridf, line 3: Temporary pending our suricata package. Until then, suricatarunner is part of this repo and so can't be .gitignored.

henridf:
brim integration test: https://github.com/brimsec/brim/pull/1088

zqd/ingest/pcap.go
```diff
@@ -116,7 +140,7 @@ outer:
 			break outer
 		case t := <-ticker.C:
 			if t.After(start.Add(next)) {
-				if err := p.createSnapshot(ctx); err != nil {
+				if err := p.createSnapshot(ctx, p.zeekFiles()); err != nil {
```
alfred-landrum, line 143: Would you add a comment stating why the suricata files aren't included in the periodic snapshots?

zqd/ingest/pcap.go
```diff
@@ -210,6 +267,28 @@ func (p *PcapOp) createSnapshot(ctx context.Context) error {
 	return nil
 }
 
+func (p *PcapOp) convertSuricataLog(ctx context.Context) error {
+	// For now, this just converts from json to
+	// zng. Soon it will apply a typing config and rename at least
+	// the timestamp field.
+	path := filepath.Join(p.logdir, "eve.json")
+	zctx := resolver.NewContext()
+	wc, err := fs.NewFileReplacer(filepath.Join(p.logdir, "eve.zng"), os.FileMode(0666))
```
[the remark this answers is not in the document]
henridf: 👍 thanks

zqd/suricata/suricata.go
```diff
@@ -0,0 +1,41 @@
+package suricata
```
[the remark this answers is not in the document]
henridf: Yes! Done.

zqd/ingest/pcap.go
henridf: Great question 🙂. This separation is an artefact of how the PR evolved, and as it currently stands there's no reason not to include them. So I've changed this to do that.

zqd/core.go
```diff
@@ -5,26 +5,29 @@ import (
...
 	"github.com/brimsec/zq/pkg/iosrc"
+	"github.com/brimsec/zq/zqd/process"
 	"github.com/brimsec/zq/zqd/space"
-	"github.com/brimsec/zq/zqd/zeek"
 	"go.uber.org/zap"
 )
 
+type Launchers struct {
+	Suricata, Zeek process.Launcher
+}
```
mattnibs, line 15: Rather than create a Launchers struct why not just add these as separate values in the Config struct?

zqd/suricata/suricata.go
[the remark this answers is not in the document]
mattnibs: Was going to suggest the same thing. Could we also rename the `process` package? `process` makes me think this is a general package for running system processes, which is only have the story. We use both `zeek` and `suricata` to perform analysis on pcaps and generate logs. I am struggling with come up with a great name, but what about something like `pcapanalyzer`?

zqd/ingest/pcap.go
```diff
@@ -210,6 +268,29 @@ func (p *PcapOp) createSnapshot(ctx context.Context) error {
...
+	if err != nil {
+		return err
+	}
+	defer wc.Close()
+	zw := zngio.NewWriter(wc, zngio.WriterOpts{})
+	if err := zbuf.CopyWithContext(ctx, zw, zr); err != nil {
+		return err
+	}
+	return zw.Close()
```
alfred-landrum, line 291: Oops, I should have remembered last night that there's a wrapper which will handle all of the file closes internally:
```suggestion
	return fs.ReplaceFile(filepath.Join(p.logdir, "eve.zng"), os.FileMode(0666), func(w io.Writer) error {
		zw := zngio.NewWriter(zio.NopCloser(w), zngio.WriterOpts{})
		return zbuf.CopyWithContext(ctx, zw, zr)
	})
```

alfred-landrum (approved):

zqd/suricata/suricata.go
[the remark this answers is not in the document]
henridf: 👍 i agree about `process` but likewise couldn't come up with a better name. I'll switch to `pcapanalyzer`

mattnibs (approved):

mattnibs:
This is an awesome feature 🚀

Issue #1271, opened 2020-09-14
Invoke fuse to ensure output of heterogeneous csv


mccanne:
The CSV writer should be able to write output zng data that comes from different record types. Including the new `fuse` processor at the end of the ZQL pipeline ensures this is possible today. However, `fuse` requires making two passes through the data, which has a performance cost and delays the immediate stream of output. Power users that are confident their data already conforms to a single record definition may want to avoid this penalty.

As a group we discussed adding a flag to determine this behavior when CSV output format is requested. In one mode it would always implicitly add `fuse` to the pipeline even if the user didn't request it, ensuring successful CSV output no matter what. There was consensus that this behavior would be invoked by the Brim app for CSV export. The other mode would follow the current behavior where only a single pass is made through the data and output stops as soon as a record is encountered in the stream that doesn't match the schema for the header already printed, at which point the user would see a message that effectively tells them to rework their query or explicitly add `fuse`. There still seemed to be some room for debate on whether `zq` at the command line should also default to the "always `fuse`" behavior planned for the Brim app or if the `zq` default should flip to this more "power user" mode.

philrz:
Verified in `zq` commit `1e501e85`.

Circling back to the previous behavior, in the last GA `zq` release tagged `v0.26.0`, invoking `-f csv` with heterogeneous data halted the output. Using the [zq-sample-data](https://github.com/brimsec/zq-sample-data):

```
$ zq -version
Version: v0.26.0

$ zq -f csv pe.log.gz stats.log.gz 
_path,ts,peer,mem,pkts_proc,bytes_recv,pkts_dropped,pkts_link,pkt_lag,events_proc,events_queued,active_tcp_conns,active_udp_conns,active_icmp_conns,tcp_conns,udp_conns,icmp_conns,timers,active_timers,files,active_files,dns_requests,active_dns_requests,reassem_tcp_size,reassem_file_size,reassem_frag_size,reassem_unknown_size
stats,2018-03-24T17:15:20.600725Z,zeek,74,26,29375,,,,404,11,1,0,0,1,0,0,36,32,0,0,0,0,1528,0,0,0
csv output requires uniform records but different types encountered
```

Now with the benefit of the enhancement, the output proceeds to completion.

```
$ zq -version
Version: v0.26.0-25-g1e501e85

$ zq -f csv pe.log.gz stats.log.gz 
[...]

Issue #1237, opened 2020-09-10
CSV writer


philrz:
A community user asked:

>  With some non-Zeek NDJSON data sources we turn into ZNG and store, we also may want to write to a SQL db (JSON works but prefer CSV) to populate visualizations of the data in say like an Apache Superset (which I actually have plans to integrate for Zeek data, probably back out in Zeek JSON from ZNG).

alfred-landrum:
I think the CSV writer should handle heterogenous records, though I assume that'd require writing to a temporary file to build the union of all the record types. That may be more than what the community user asked for, if their target was a sql db import.

mccanne:
@alfred-landrum, agreed.  I will write up a new issue for this.  There are a couple ways to do it.

mccanne:
We also need to change the search endpoint with csv output (as part of the larger refactoring there) so the front-end doesn't have to have its own csv writer and record unflattener in javascript.

philrz:
Verified in `zq` commit `4bce00d`.

Output as CSV currently works for any data that can be represented by a single "descriptor" (i.e. they all have the same record type). For instance, the contents of a single Zeek log:

```
$ zq -f csv stats.log.gz 
_path,ts,peer,mem,pkts_proc,bytes_recv,pkts_dropped,pkts_link,pkt_lag,events_proc,events_queued,active_tcp_conns,active_udp_conns,active_icmp_conns,tcp_conns,udp_conns,icmp_conns,timers,active_timers,files,active_files,dns_requests,active_dns_requests,reassem_tcp_size,reassem_file_size,reassem_frag_size,reassem_unknown_size
stats,2018-03-24T17:15:20.600725Z,zeek,74,26,29375,-,-,-,404,11,1,0,0,1,0,0,36,32,0,0,0,0,1528,0,0,0
stats,2018-03-24T17:20:20.60102Z,zeek,281,6435390,3052165142,-,-,-,2575313,2575317,5818,187,1070,338620,5126,9490,1301330,33086,48528,27,61,0,77448,53168,0,0
stats,2018-03-24T17:25:20.601054Z,zeek,282,4556854,1920350471,-,-,-,1958516,1958507,4401,110,1010,211123,4600,2354,1006374,27351,41930,10,5,0,90752,0,1672,0
[...]

PR #1300, opened 2020-09-17
Output format changes: Add "csv", remove "types"

philrz:
While verifying #1237, I noticed that CSV is not yet listed among the output formats. I wondered if maybe we were intentionally holding off on revealing it until we address #1271, but it seems useful enough in its present form that I'm proposing here that we reveal it now.

I'd also recalled seeing @mccanne mention recently that `types` was removed as an output format. Indeed, as of `zq` commit `4bce00d`:

```
$ zq -version
Version: v0.21.0-27-g4bce00d
```

Therefore I'm also taking that out while I'm at it.

### DIFF
--- a/zio/flags.go
+++ b/zio/flags.go
@@ -73,7 +73,7 @@ func (w *WriterFlags) Options() WriterOpts {
 }
 
 func (w *WriterFlags) SetFlags(fs *flag.FlagSet) {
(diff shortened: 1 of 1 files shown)


mccanne (approved):
Good catch.  I had this in at some point but lost it somehow along the way.

PR #1307, opened 2020-09-18
simplify zbuf.Array

mccanne:
No need to allocate a struct to hold a pointer when the zbuf.Array
type can be a slice and implement the zbuf.Batch interface.

### DIFF
--- a/proc/proctest/utils.go
+++ b/proc/proctest/utils.go
@@ -35,7 +35,7 @@ func (r *RecordPuller) Pull() (zbuf.Batch, error) {
 		if rec == nil || err != nil {
 			return nil, err
 		}
-		return zbuf.Array([]*zng.Record{rec}), nil
+		return zbuf.Array{rec}, nil
 	}
 }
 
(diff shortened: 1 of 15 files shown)


proc/proctest/utils.go
```diff
@@ -35,7 +35,7 @@ func (r *RecordPuller) Pull() (zbuf.Batch, error) {
 		if rec == nil || err != nil {
 			return nil, err
 		}
-		return zbuf.NewArray([]*zng.Record{rec}), nil
+		return zbuf.Array([]*zng.Record{rec}), nil
```
nwt, line 38:
```suggestion
		return zbuf.Array{rec}, nil
```

proc/uniq/uniq.go
```diff
@@ -67,14 +67,14 @@ func (p *Proc) Pull() (zbuf.Batch, error) {
 		}
 		t := p.wrap(p.last)
 		p.last = nil
-		return zbuf.NewArray([]*zng.Record{t}), nil
+		return zbuf.Array([]*zng.Record{t}), nil
```
nwt, line 70:
```suggestion
		return zbuf.Array{t}, nil
```

nwt (approved):

proc/proctest/utils.go
mccanne: nice

PR #1310, opened 2020-09-18
add fuse/dataframe proc

### LINKED ISSUE #1269: dataframe proc
zql should have a dataframe proc to turn a zng stream into a dataframe comprised of one type.  This would entail creating an "uber schema" that is the union of all the record types and rewriting each value against the uber schema.  This clearly needs two passes.

For an initial implementation, we can buffer all the records in memory and return an error if we hit a limit.  This is what will be done here. 

As a second step, we can add spilling when the memory limit is exceeded.  This will be put in a separate issue.

mccanne:
This commit adds a dataframe proc "fuse" which transforms the input
zng stream into an output stream that all has the same uber schema
by taking the union of all of the record types in the input and
coding each record to its position in the uber schema.

Because fuse needs to know all the types to build the uber schema,
it must buffer all of its output before it can generate the first
output record, just like sort  Thus, there is a memory-limit
circuit breaker.

The next step here is to add logic to spill to disk when the breaker
is tripped.  This is easier than sort because you don't need multiple
runs in different files and a merge.  It can all be done with a
single file.

Closes #1269

### DIFF
--- a/driver/compile.go
+++ b/driver/compile.go
@@ -536,11 +536,7 @@ func parallelizeFlowgraph(seq *ast.SequentialProc, N int, inputSortField string,
 			}
 			// put one head/tail on each parallel branch and one after the merge.
 			return buildSplitFlowgraph(seq.Procs[0:i+1], seq.Procs[i:], inputSortField, inputSortReversed, N), true
-		case *ast.FuseProc:
-			// The df proc could conceivably be parallelized but there are better ways
-			// to optimize it with columnar form when we get to that...
-			return seq, false
-		case *ast.UniqProc:
+		case *ast.UniqProc, *ast.FuseProc:
 			if inputSortField == "" {
 				// Unknown order: we can't parallelize because we can't maintain this unknown order at the merge point.
 				return seq, false
--- a/driver/compile.go
+++ b/driver/compile.go
@@ -200,7 +200,7 @@ func setGroupByProcInputSortDir(p ast.Proc, inputSortField string, inputSortDir 
 			}
 		}
 		return true
-	case *ast.FilterProc, *ast.HeadProc, *ast.PassProc, *ast.UniqProc, *ast.TailProc:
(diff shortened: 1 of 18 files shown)


proc/df/schema.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+func (s *schema) mixin(typ *zng.TypeRecord) []int {
+	var positions []int
+	for _, c := range typ.Columns {
+		name := c.Name
+		key := fmt.Sprintf("%s%d", name, c.Type.ID())
+		uberPosition, ok := s.position[key]
+		if !ok {
+			cnt := s.touch(c.Name)
```
henridf, line 37:
```suggestion
			cnt := s.touch(name)
```

henridf (commented):
`ast.DfProc` needs be added to `driver.Compile.computeColumnsR` and `driver.Compile.parallelizeFlowgraph`. 

In both cases, I think it can be handled in the same case as `UniqProc`.

henridf (approved):

driver/compile.go
```diff
@@ -536,6 +536,10 @@ func parallelizeFlowgraph(seq *ast.SequentialProc, N int, inputSortField string,
 			}
 			// put one head/tail on each parallel branch and one after the merge.
 			return buildSplitFlowgraph(seq.Procs[0:i+1], seq.Procs[i:], inputSortField, inputSortReversed, N), true
+		case *ast.DfProc:
```
henridf, line 539: Why not handle this like UniqProc below? So that `search | df` can be parallelized. Pending all this getting subsumed by columnar, of course.

alfred-landrum:
I don't think we should use `df` as the name of this proc; it's not a verb, and doesn't readily communicate its action, like our other proc names. I don't have a solid counter - `combine` first comes to mind, but its problematic as well.

henridf:
> I don't think we should use df as the name of this proc

I did have the same thought but was wary of the shed. Now the discussion is opened, another candidate could be `assemble`. I thought of that because this is essentially doing the same thing as "Assembler" over in columnar land, but one layer above. That said I'm not sure that the word readily communicates the action either. 😬. `table`?

Issue #1262, opened 2020-09-14
Use data file names to hold span information


alfred-landrum:
Per #1183 , encode a data files time span in its name, use that info when selecting files to read for a query . As part of this work, we should also stop storing span/index information in the zar.json metadata file.

philrz:
Verified in `zar` commit `d8f8855`.

In terms of what a user would see differently, the key difference is that ZNG files created by the import process will have unique names for every import action. Using the import example from the [`zar` README](https://github.com/brimsec/zq/blob/master/cmd/zar/README.md):

```
$ rm -rf $ZAR_ROOT ; zq zng/*.gz | zar import -s 25MB -
$ tree -s logs
logs
├── [         88]  zar.json
└── [         96]  zd
    └── [        320]  20180324
        ├── [   25002258]  d-1hyC2rr6TwJIwflUMUAt0AZnXAy.zng
        ├── [   25005739]  d-1hyC2sIVetRsvVQ66nDukNc6R6o.zng
        ├── [   25004149]  d-1hyC2tb584tszcEhLjMO2rGFdmO.zng
        ├── [    5047321]  d-1hyC2whHpj6XZ4z3jf5qPvNsLCZ.zng
        ├── [       1493]  ts-1hyC2rr6TwJIwflUMUAt0AZnXAy-443503-1521912512587283000-1521912080565725000.zng
        ├── [       1554]  ts-1hyC2sIVetRsvVQ66nDukNc6R6o-456922-1521912990158766000-1521912512587864000.zng
        ├── [       1588]  ts-1hyC2tb584tszcEhLjMO2rGFdmO-470614-1521912080565723000-1521911778227216000.zng
        └── [        386]  ts-1hyC2whHpj6XZ4z3jf5qPvNsLCZ-91039-1521911778225955000-1521911720600725000.zng

2 directories, 9 files

$ rm -rf $ZAR_ROOT ; zq zng/*.gz | zar import -s 25MB -
$ tree -s logs
logs
├── [         88]  zar.json
└── [         96]  zd
    └── [        320]  20180324
        ├── [   25005739]  d-1hyCAXRGJbEdQiLoRjn0ogb17dx.zng
        ├── [   25002258]  d-1hyCAZLsx9BnpbohERv8nTeKI1J.zng
        ├── [   25004149]  d-1hyCAc9fsvmLa1tSQawVoO8G9GP.zng
[...]

PR #1311, opened 2020-09-18
add readdir equivalent to iosrc

alfred-landrum:
This adds a ReadDir call to iosrc with implementations for file & s3 schemas. This will be used (and tested) in the zar changes coming in #1262 .

### DIFF
--- a/zqd/space/manager.go
+++ b/zqd/space/manager.go
@@ -70,14 +70,14 @@ func NewManagerWithContext(ctx context.Context, root iosrc.URI, logger *zap.Logg
 }
 
 func s3spaces(ctx context.Context, root iosrc.URI) ([]iosrc.URI, error) {
-	prefixes, err := s3io.ListCommonPrefixes(ctx, root.String(), nil)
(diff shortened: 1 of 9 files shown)


zqd/space/manager.go
```diff
@@ -70,14 +70,14 @@ func NewManagerWithContext(ctx context.Context, root iosrc.URI, logger *zap.Logg
...
+	entries, err := s3io.List(ctx, root.String(), nil)
 	if err != nil {
 		return nil, err
 	}
 	var uris []iosrc.URI
-	for _, p := range prefixes {
+	for _, e := range entries {
 		u := root
-		u.Path = p
+		u.Path = e.Name
```
mattnibs, line 80: I think with these changes we can now merge the s3spaces and filespaces calls, no?

mattnibs (approved):

zqd/space/manager.go
alfred-landrum: 👍 done!

PR #1323, opened 2020-09-19
refactor command flags one more time

### LINKED ISSUE #1279: Remove "-sortmem" from tools that don't sort
When looking at https://github.com/brimsec/zq/pull/1249, I saw how `-sortmem` is now present in the `help` text of tools like `pcap` and `zapi` that I don't think of as doing sorting. I asked @mccanne if this was intentional and he said it was not and could be fixed in a planned refactor. We also agreed this is a low priority, but I've opened this issue as a reminder to circle back to it.

mccanne:
This commit refactors command flags so all the flag processing
is handled under "zq/flags" and different flag "bundles" properly
partition functionality (e.g., so -sortmem doesn't show up
in the pcap command and so anything that imports cli doesn't
necessarily import all of zio).

Along the way, we made the 3rd-party "units" package a bit more
ergonomic for use with the go "flag" package using a wrapper
package in "zq/pkg/units".

This commit addresses the issue regarding bytes syntax for command 
flags raised on PR #1268 

Fixes #1279


mattnibs (approved):
lgtm

cmd/zar/map/command.go
```diff
@@ -152,10 +141,11 @@ func (c *Command) Run(args []string) error {
 
 func (c *Command) openOutput(zardir iosrc.URI) (zbuf.WriteCloser, error) {
 	var path string
-	if c.outputFile != "" {
-		path = zardir.AppendPath(c.outputFile).String()
+	filename := c.outputFlags.FileName()
+	if filename != "" {
+		path = zardir.AppendPath(filename).String()
```
alfred-landrum, line 146:
```suggestion
	if filename := c.outputFlags.FileName(); filename != "" {
		path = zardir.AppendPath(filename).String()
```

alfred-landrum (approved):

PR #1324, opened 2020-09-19
ZQL doc for fuse processor

philrz:


zql/docs/processors/README.md
```diff
@@ -154,6 +155,58 @@ ssl   1521912240.189735 CSbGJs3jOeB6glWLJj 10.47.7.154 27137     52.85.83.215 44
 
 ---
 
+## `fuse`
+
+|                           |                                                   |
+| ------------------------- | ------------------------------------------------- |
+| **Description**           | Return results under a single record definition that unifies the field and type information across all records in the query result. |
```
mccanne, line 162: I think "return" is the wrong terminology for a dataflow pipeline.  How about "Transforms input records to output records that unify... "

zql/docs/processors/README.md
```diff
@@ -154,6 +155,58 @@ ssl   1521912240.189735 CSbGJs3jOeB6glWLJj 10.47.7.154 27137     52.85.83.215 44
...
+_PATH TS                UID                ID.ORIG_H   ID.ORIG_P ID.RESP_H      ID.RESP_P NAME                             ADDL NOTICE PEER
+weird 1521911720.600843 C1zOivgBT6dBmknqk  10.47.1.152 49562     23.217.103.245 80        TCP_ack_underflow_or_misorder    -    F      zeek
+weird 1521911720.608108 -                  -           -         -              -         truncated_header                 -    F      zeek
+weird 1521911720.610033 C45Ff03lESjMQQQej1 10.47.5.155 40712     91.189.91.23   80        above_hole_data_without_any_acks -    F      zeek
+weird 1521911720.742818 Cs7J9j2xFQcazrg7Nc 10.47.8.100 5900      10.129.53.65   58485     connection_originator_SYN_ack    -    F      zeek
+```
+
+Here a `stats` event was the first record type to be printed in the results stream, so the preceding header row describes the names of its fields. Then a `weird` event came next in the results stream, so a header row describing its fields was printed. This presentation accurately conveys the heterogeneous nature of the data, but the frequent interruption of data rows with different headers is distracting and may make the output difficult to process in downstream tools.
```
mccanne, line 188: I'm not sure "distracting" is the right motivation.  Why not say that certain tooling like SQL inserts or csv output cannot handle heterogeneous record types so fuse allows you to combine them into a unified schema.

mccanne (approved):

zql/docs/processors/README.md
philrz: @mccanne: Fair point. Your suggestion looks good to me, so I've committed that change.

zql/docs/processors/README.md
philrz: I was on the fence about covering much CSV here because I know the output is due to change when we address #1326, so I was sticking with tables, which I admit is pretty banal. But your feedback is certainly valid so I've committed a change that shows CSV and mentions SQL, and the automation will remind whoever addresses #1326 to fix the example output.

(A closing note on this change in case someone circles back to this PR at a later date. The `zq` command line that causes the `csv output requires uniform records but different types encountered` message is intentionally _not_ run through the output checker. This is because the automation that checks these examples appears to fail any test that returns a non-zero exit code. Ideally we'd enhance the automation to allow the marking of a `zq` command where we _expect_ this to happen, but we want to check that the output and error message that make it through before the exit don't change. However I didn't wanna hold this up by enlisting a Go developer to write that enhancement, so I'm pushing forward with what we've got.)

PR #1327, opened 2020-09-20
zng marshal

mccanne:
This commit adds a basic marshaler to generate zng streams
from native go values.  It could be extended to handle zng
sets, enums, etc by adding the struct tag syntax.  This can
come later.

Also, we haven't done unmarshal yet as we want the marshal
side asap for a demo.


alfred-landrum (approved):
I'd be good with having this go in & iterate. I was just touching the code for `zar stat`, and I can try using Marshal there; that will require adding support for marshalling time types.

PR #1328, opened 2020-09-20
align zapi with rest of tools that print help with no args

### LINKED ISSUE #1306: Make "zapi" by itself print help text
While verifying #1250, I noticed that for most(*) of our CLI tools, just executing the command name by itself (e.g. `zq`) gave the same help text as if it had been explicitly requested (e.g. `zq help`). One exception to this is `zapi`:

```
$ zapi
either space name (-s) or id (-id) must be specified
```

@mccanne explained this is because minimal `zapi` normally drops into a REPL for entering ZQL.

```
$ zapi new foo
foo: space created

$ zapi -s foo
foo> 
```

However, since the user needs to type _something_ regardless to get into the REPL, we agreed that it would be reasonable to match the behavior of the other commands and print the help text by default, and have an explicit command that drops into the REPL mode if that's functionality we want to keep.

---
(*) - Technically `mockbrim` has the same problem, but that one doesn't even have a `mockbrim help`. It's apparently a tool that's just used for debug/test, and @mccanne explained that it doesn't even use Charm like the others, so it'd not be an easy change. I'm happy to leave that one out-of-scope for this exercise.

mccanne:
This commit changes "zapi" to print help when run with no args.
The repl is now run with a new subcommand "zapi repl".

Fixes #1306

### DIFF
--- a/cmd/zapi/cmd/cli.go
+++ b/cmd/zapi/cmd/cli.go
@@ -5,12 +5,10 @@ import (
 	"errors"
 	"flag"
 	"fmt"
-	"io"
 	"os"
 	"syscall"
 
 	"github.com/brimsec/zq/cli"
-	"github.com/brimsec/zq/pkg/repl"
 	"github.com/brimsec/zq/zqd/api"
 	"github.com/kballard/go-shellquote"
 	"github.com/mccanne/charm"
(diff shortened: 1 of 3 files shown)


mccanne:
This will need to be rebased to #1334.  We should merge that first.

mattnibs (approved):

Issue #1183, opened 2020-08-27
design zar support for overlapping data


alfred-landrum:
Ingest & reading of data from an archive needs to handle the case when the time range for chunk files overlap, as well as dealing with the eventual consistency model of S3. I put up a proposal to handle this:

[edit: see later comment with updated proposal]

I'll break this out into a few issues.

alfred-landrum:
Closing, as the layout changes to handle overlapping time ranges (including compaction) are in place. Further overall design work that's left & needed for zar archives is how we want to handle named mapping files, which is needed for #1265 .

PR #1330, opened 2020-09-20
change zar data file layout for future overlap support

### LINKED ISSUE #1262: Use data file names to hold span information

alfred-landrum:
This PR changes the name & layout of data files in a zar archive, following the design in #1183.

- On import, two files are created: a data file, and an index file, whose key/values are record timestamps & stream seek offsets. The index file content is not yet used during reads, but its name is used to encode the time span of records within the data file.
- zar import now ensures the top level time partition directories ("tsDirs" in the code) only have data for their span, splitting incoming data into more data files if necessary. This is one reason why the "-s 20KiB" flag for importing the "babble" went away; as that was done mostly to cause the babble import to create multiple data files. That now happens w/o extra flags, as babble's data is split between different days.
- Nearly all zar tests needed to change, since they relied on the file paths of an archive to be predictable. With this change, uuids are used in the names of ingested data files, as foundational work for handling S3's consistency model.
- To help account for the test changes, I added a "-ranges" command to a few zar operations, which causes the output to use the time range of records per log file, instead of the log files zardir's path (which would include the data file's uuid).

Closes #1262


archive/walk.go
```diff
@@ -2,18 +2,235 @@ package archive
...
+
+// A uuid is a unique identifier created with the ksuid package. These ids
+// are always 27 characters long, and combine 32 bits of time plus 128
+// random bits. The ids are roughly sortable by generation time, which can
+// be useful for debugging & investigation.
+type uuid string
+
+func newUUID() uuid {
+	return uuid(ksuid.New().String())
+}
```
nwt, line 48: Why not just use github.com/segmentio/ksuid.KSUID for this?

archive/walk.go
```diff
@@ -2,18 +2,235 @@ package archive
...
+func parseTsDirName(name string) (tsDir, bool) {
+	t, err := time.Parse("20060102", name)
+	if err != nil {
+		return tsDir{}, false
+	}
+	return tsDirFor(nano.TimeToTs(t)), true
+}
+
+func (t tsDir) name() string {
+	return path.Join(t.Ts.Time().Format("20060102"))
```
nwt, line 144:
```suggestion
	return t.Ts.Time().Format("20060102")
```

archive/walk.go
```diff
@@ -2,18 +2,235 @@ package archive
...
+	}
+	id := uuid(match[1])
+	return dataFile{id, ingestDataKind}, true
+}
+
+// A seekIndexFile is a microindex whose keys are record timestamps, and whose
+// values are seek offsets into the data file with the same uuid. The name
+// of a seekIndexFile encodes the number of records, and the first & last
+// record timestamps of the corresponding data file. The order of the
+// first & last records matches the archives data order.
```
nwt, line 80:
```suggestion
// first & last records matches the archive's data order.
```

archive/walk.go
alfred-landrum: 👍 I thought about using something other than jus ksuid.String(), like adding some prefix (like we do with space id's). But file names will have a prefix already (like "d-..." for data files) that will help explain what the id represents.

archive/walk.go
```diff
@@ -2,18 +2,232 @@ package archive
...
+		tsdirs = append(tsdirs, tsd)
+	}
+	sort.Slice(tsdirs, func(i, j int) bool {
+		if ark.DataSortDirection == zbuf.DirTimeForward {
+			return tsdirs[i].Ts < tsdirs[j].Ts
+		}
+		return tsdirs[j].Ts < tsdirs[i].Ts
+	})
+	for _, d := range tsdirs {
+		dirents, err := iosrc.ReadDir(ctx, zdDir.AppendPath(d.name()))
```
mattnibs, line 173: This is not going to scale well for dense datasets. At some point we'll want to consider either some callback approach that returns chunks of file infos or generate some kind of file tree that has a max number of leaves before a node is split.

archive/walk.go
```diff
@@ -2,18 +2,232 @@ package archive
...
+)
+
+// A fileKind is the first part of a file name, used to differentiate files
+// when they are listed from the archive's backing store.
+type fileKind string
+
+const (
+	unknownFileKind fileKind = ""
+	ingestDataKind           = "d"
+	seekIndexKind            = "ts"
```
mattnibs, line 38: I think these should start with `fileKind`. Will read easier.
```suggestion
	fileKindUnkown fileKind = ""
	fileKindData                    = "d"
	fileKindSeekIndex          = "ts"
```

alfred-landrum:
Yes; one easy way is to keep the tsDir hierarchy static, and just increase
the granularity (like, from a day to an hour). I wouldn't yet want to try
to recreate the dynamic splitting that was used in the boom system.

mattnibs (commented):
Looking good 👍

mattnibs (approved):

archive/import.go
```diff
@@ -75,17 +121,48 @@ func (d *importDriver) writeOne(rec *zng.Record) error {
...
+		idxWriter.Abort()
+		return err
+	}
+	if err := idxWriter.Close(); err != nil {
+		return err
+	}
+	// TODO: zq#1264
+	// Add an entry to the update log for S3 backed stores containing the
+	// location of the just added data & index file.
 	return nil
```
nwt, line 166:
```suggestion
	// TODO: zq#1264
	// Add an entry to the update log for S3 backed stores containing the
	// location of the just added data & index file.
	return idxWriter.Close()
```

archive/schema.go
```diff
@@ -3,49 +3,22 @@ package archive
...
-	RecordCount int       `json:"record_count"`
-}
-
-type IndexInfo struct {
-	Type string `json:"type"`
-	Path string `json:"path"`
+	Version           int            `json:"version"`
+	DataPath          string         `json:"data_path"`
+	LogSizeThreshold  int64          `json:"log_size_threshold"`
+	DataSortDirection zbuf.Direction `json:"data_sort_direction"`
```
nwt, line 21: Nit: Sort these or, if you prefer to have Version first, sort the subsequent fields (i.e., Version, DataPath, DataSortDirection, LogSizeThreshold).

archive/schema.go
```diff
@@ -297,22 +176,13 @@ func openArchive(ctx context.Context, root iosrc.URI, oo *OpenOptions) (*Archive
 		}
 	}
 	if oo != nil && len(oo.LogFilter) != 0 {
```
nwt, line 178: Nit: No need for ` && len(oo.LogFilter) != 0` here.

archive/schema.go
```diff
@@ -342,6 +212,9 @@ func CreateOrOpenArchiveWithContext(ctx context.Context, rpath string, co *Creat
 			if err := dm.MkdirAll(root, 0700); err != nil {
 				return nil, err
 			}
+			if err := dm.MkdirAll(root.AppendPath(dataDirname), 0700); err != nil {
+				return nil, err
+			}
```
nwt, line 217: NIt: A single dm.MkdirAll should suffice here.

archive/walk.go
```diff
@@ -2,18 +2,232 @@ package archive
...
+}
+
+var dataFileNameRegex = regexp.MustCompile(`d-([0-9A-Za-z]{27}).zng$`)
+
+func dataFileNameMatch(s string) (f dataFile, ok bool) {
+	match := dataFileNameRegex.FindSubmatch([]byte(s))
+	if match == nil {
+		return
+	}
+	id, err := ksuid.Parse(byteconv.UnsafeString(match[1]))
```
nwt, line 63: Nit: No need for []byte(s) or byteconv.UnsafeString if you use dataFileNameRegex.FindStringSubmatch.

archive/walk.go
```diff
@@ -2,18 +2,232 @@ package archive
...
+	}
+	recordCount, err := strconv.ParseInt(byteconv.UnsafeString(match[2]), 10, 64)
+	if err != nil {
+		return
+	}
+	firstTs, err := strconv.ParseInt(byteconv.UnsafeString(match[3]), 10, 64)
+	if err != nil {
+		return
+	}
+	lastTs, err := strconv.ParseInt(byteconv.UnsafeString(match[4]), 10, 64)
```
nwt, line 109: Nit: As above, no need for []byte(s) or byteconv.UnsafeString if you use seekIndexNameRegex.FindStringSubmatch.

archive/walk.go
```diff
@@ -2,18 +2,232 @@ package archive
...
+}
+
+// A tsDir is a directory found in the "<DataPath>/zd" directory of the archive,
+// and holds all of the data & associated files for a span of time, currently
+// fixed to a single day.
+type tsDir struct {
+	nano.Span
+}
+
+func tsDirFor(ts nano.Ts) tsDir {
```
nwt, line 128: NIt: I'd call this newTsDir.

archive/walk.go
```diff
@@ -2,18 +2,232 @@ package archive
...
+			return err
+		}
+		if err := visitor(d, tsDirEntriesToSpans(ark, filterSpan, dirents)); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func tsDirEntriesToSpans(ark *Archive, filterSpan nano.Span, entries []iosrc.Info) []SpanInfo {
```
nwt, line 184: Nit: I'd call this tsDirEntriesToSpanInfos.

zio/zngio/writer.go
```diff
@@ -9,8 +9,10 @@ import (
 	"github.com/pierrec/lz4/v4"
 )
 
-// DefaultLZ4BlockSize is a reasonable default for the WriterOpts.LZ4BlockSize.
-const DefaultLZ4BlockSize = 16 * 1024
+const (
+	DefaultLZ4BlockSize     = 16 * 1024 // a reasonable default for WriterOpts.LZ4BlockSize
+	DefaultStreamRecordsMax = 5000      // a reasonable default for WriterOpts.StreamRecordsMax
```
nwt, line 14: Nit: Make these doc comments.
```suggestion
	// DefaultLZ4BlockSize is a reasonable default for WriterOpts.LZ4BlockSize.
	DefaultLZ4BlockSize     = 16 * 1024
	// DefaultStreamRecordsMax is a reasonable default for WriterOpts.StreamRecordsMax.
	DefaultStreamRecordsMax = 5000
```

zio/zngio/writer.go
```diff
@@ -21,6 +23,7 @@ type Writer struct {
 	buffer           []byte
 	streamRecords    int
 	streamRecordsMax int
+	lastsos          int64
```
nwt, line 26: Nit: Call this lastSOS and put it between buffer and streamRecords.

zqd/handlers_test.go
```diff
@@ -1081,6 +1084,28 @@ func searchTzng(t *testing.T, client *api.Connection, space api.SpaceID, prog st
...
+	zctx := resolver.NewContext()
+	r := tzngio.NewReader(bytes.NewReader([]byte(in)), zctx)
+	buf := bytes.NewBuffer(nil)
+	w := textio.NewWriter(zio.NopCloser(buf), false, textio.WriterOpts{}, false)
+	p := zql.MustParseProc(prog)
+	err := driver.Copy(context.Background(), w, p, zctx, r, driver.Config{})
+	require.NoError(t, err)
+	return buf.String()
+}
+
```
nwt, line 1108: How about this?
```go
func tzngCopy(t *testing.T, prog string, in string, outputFormat string) string {
	zctx := resolver.NewContext()
	r := tzngio.NewReader(strings.NewReader(in), zctx)
	var buf bytes.Buffer
	w := detector.LookupWriter(zio.NopCloser(&buf), zio.WriterOpts{Format: outputFormat})
	p := zql.MustParseProc(prog)
	err := driver.Copy(context.Background(), w, p, zctx, r, driver.Config{})
	require.NoError(t, err)
	return buf.String()
}
```

nwt (approved):

archive/walk.go
alfred-landrum: I'm leaving this as is - I'm going to change the type name from 'SpanInfo' to something else in the overlap PR.

archive/walk.go
alfred-landrum: I didn't want to use the regex string functions since we don't need to create the strings; I did change this to use the `byteconv.ParseInt64` function, so it's clearer.

nwt: Both FindSubmatch and FindStringSubmatch return a slice of slices of their argument, so their allocation behavior is the same.

archive/walk.go
nwt: 👍

archive/walk.go
alfred-landrum: ah, thanks!

PR #1334, opened 2020-09-20
one more time fixing CLI flags refactor

### LINKED ISSUE #1333: "zq -version" broke at commit 7f00817
WIth the arrival of commit `7f00817` that corresponds to PR #1323, `zq -version` stopped working. It just prints the help text now.

```
$ zq -version
NAME
    zq - command line logs processor

USAGE
    zq [ options ] [ zql ] file [ file ... ]
...
```

mccanne:
The last refactor broke "zq -version".  The previous logic was a bit
fragile because cli.Flags.Init() would return a flag to stop
executing if -version was given.  But now, Init() can fail for
other reasons, e.g., writing binary data to stdout.  This would
block the help screen if Init was before the help check, but if we
put it after, then help would be displayed with -version.

The fix is to simplify the logic by cli.Init calling os.Exit if
"-version" is given and putting the error check after the subsequent
check for help.  This only applies to zq and not the other tools
because the other tools use subcommands and the issue does not
arise with the root command that does not have outputFlags checks.

We also added procFlags to zqd since they were missing in the last
refactor.  This gives "-sortmem" back to zqd.

Fixes #1333

### DIFF
--- a/cmd/microindex/root/command.go
+++ b/cmd/microindex/root/command.go
@@ -29,7 +29,7 @@ func (c *Command) Cleanup() {
 	c.cli.Cleanup()
 }
 
-func (c *Command) Init(all ...cli.Initializer) (bool, error) {
+func (c *Command) Init(all ...cli.Initializer) error {
 	return c.cli.Init(all...)
 }
 
@@ -41,11 +41,12 @@ func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
 
 func (c *Command) Run(args []string) error {
 	defer c.cli.Cleanup()
-	if ok, err := c.cli.Init(); !ok {
-		return err
-	}
+	err := c.cli.Init()
 	if len(args) == 0 {
 		return MicroIndex.Exec(c, []string{"help"})
+	}
+	if err != nil {
+		return err
 	}
 	return charm.ErrNoRun
 }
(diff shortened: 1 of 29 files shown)


cmd/microindex/root/command.go
```diff
@@ -41,11 +41,12 @@ func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
...
 func (c *Command) Run(args []string) error {
 	defer c.cli.Cleanup()
-	if ok, err := c.cli.Init(); !ok {
-		return err
-	}
+	err := c.cli.Init()
 	if len(args) == 0 {
 		return MicroIndex.Exec(c, []string{"help"})
 	}
+	if err != nil {
```
mattnibs, line 48: drop line 44 and do
```suggestion
	if err := c.cli.Init(); err != nil {
```

mattnibs (approved):

alfred-landrum:
Can you add a test for the `zq -version` call?

cmd/microindex/root/command.go
mccanne: If you do this, then you'll get the help screen when you run `microindex -version`.

mccanne: It's a little fragile.  I think I'll leave it as is until we come up with something better.

PR #1336, opened 2020-09-21
zqd: Multipart log ingest

### LINKED ISSUE #1084: Support network based multiple log import
The existing log import api sends filenames between Brim & zqd, as it initially only needed to support the Brim setup where both are co-located. We should change the api to send log data over the socket, including supporting multiple logs in a single api call.

UPDATE- We want to include the new multipart-form upload but keep the old path based endpoint. The old path based endpoint is useful in that s3 logs can be posted directly to the server.

mattnibs:
Rearrange zqd log post endpoint so log files are transmitted over
the network instead of opened from a locally available file.

The files are submitted as an http multipart request. JSON typing
can be submitted as part of the request- it should be given the
form value of "json_type_config".

For http 1.1 the server cannot send responses back while the
requester is sending data. As a result remove status updates and
transmit any wires in the server responses. Clients that want to
know the amount of data sent while the request is in flight will
have to keep track the bytes written over the wire themselves.

Closes #1084


mattnibs:
note: This will break brim master. I am working on an update there.

zqd/ingest/logreader.go
```diff
@@ -0,0 +1,124 @@
...
+read:
+	if m.done {
+		return nil, nil
+	}
+	if m.zreader == nil {
+		zr, err := m.next()
+		if err == io.EOF {
+			m.done = true
+			return nil, nil
+		}
```
henridf, line 53:
```suggestion
	if m.zreader == nil {
		zr, err := m.next()
		if err == io.EOF {
			return nil, nil
		}
```
(and if you apply this, also remove `done` from `MultipartLogReader`.

zqd/ingest/logreader.go
```diff
@@ -0,0 +1,124 @@
...
+	return rec, err
+}
+
+func (m *MultipartLogReader) next() (*zbuf.File, error) {
+next:
+	part, err := m.mr.NextPart()
+	if err != nil {
+		return nil, err
+	}
+	if part.FormName() == "json_type_config" {
```
henridf, line 81: The zapi client sends `json_type_config` first, then files, which seems like what we'd want every client to do. How about checking that here, e.g. if a `json_type_config` is sent after logs, that is an error?

mattnibs: It seems like maybe it could be a feature? For instance if you have two separate typing configs for two different sets of json logs. No one will ever do this, but it is there. Perhaps @philrz is crazy enough to try.

zqd/ingest/logreader.go
mattnibs: You know I had this but for some reason I was getting an extra Read call after EOS. I was too lazy to debug it, but I'll take a second 👀

mattnibs: We need keep this. `ReadBatch` for scanner requires receiving two EOS Reads from the underlying reader. One two prematurely close the `Batch` and another to return `EOS` on the next call. For some reason the golang multipart reader returns a different, unwrapped `EOF` after the first `EOF` is returned. (see https://golang.org/src/mime/multipart/multipart.go?s=8945:8987#L339). Frustrating. I will add a comment for clarity.

zqd/ingest/logreader.go
henridf: Feature or opportunity for confusion is the question 🙂 ... we can leave it as is if you think this is useful.

henridf:
General question... with this API it's now possible to post one or more files with `curl -F`?  😎

henridf (approved):
lgtm

mattnibs:
> General question... with this API it's now possible to post one or more files with `curl -F`? 😎

Yep! Good thinking from @alfred-landrum 😎

mattnibs:
Will hold off on merging until we have accompanying changes on brim.

zqd/ingest/logreader.go
philrz: There definitely will be a future where different JSON sources with different typing needs will want to be ingested, though I see it as entirely possible we'll overhaul the approach to JSON typing at the same time we make that a first class operation in the app/tooling.

zqd/ingest/logreader.go
mattnibs: Put a fix for this here: https://github.com/golang/go/pull/41560

cmd/zapi/cmd/post/post.go
```diff
@@ -73,43 +74,22 @@ func (c *LogCommand) Run(args []string) (err error) {
...
 		fmt.Println("post aborted")
 		os.Exit(1)
 		return nil
 	}
-	if err == nil {
-		read := atomic.LoadInt64(&c.bytesRead)
-		fmt.Printf("posted %s in %v\n", format.Bytes(read), time.Since(c.start))
+	if err != nil {
+		return err
+	}
```
nwt, line 85: Nit: Just test for err != nil once.
```go
	if err != nil {
		if c.Context().Err() != nil {
			fmt.Println("post aborted")
			os.Exit(1)
		}
		return err
	}
```

zqd/api/logwriter.go
```diff
@@ -0,0 +1,115 @@
...
+		if err != nil {
+			return nil, err
+		}
+		lw.BytesTotal += info.Size()
+		lw.URIs = append(lw.URIs, u)
+	}
+	return lw, nil
+}
+
+func (l *LogWriter) SetJSONTyping(config *ndjsonio.TypeConfig) {
```
nwt, line 47: Nit: I think we should drop the "typing" terminology entirely and call this SetJSONTypeConfig or SetJSONConfig.

zqd/api/logwriter.go
```diff
@@ -0,0 +1,115 @@
...
+	w, err := l.form.CreateFormFile("", u.String())
+	if err != nil {
+		return err
+	}
+	counter := &partCounter{reader: bufio.NewReader(r), nread: &l.bytesRead}
+	_, err = io.Copy(w, counter)
+	return err
+}
+
+func (l *LogWriter) sendJSONTyping() error {
```
nwt, line 95: Nit: Again, I think we should drop the "typing" terminology entirely and call this sendJSONTypeConfig or sendJSONConfig.

zqd/api/logwriter.go
```diff
@@ -0,0 +1,115 @@
...
+	if err != nil {
+		return err
+	}
+	return json.NewEncoder(w).Encode(l.json)
+}
+
+type partCounter struct {
+	reader io.Reader
+	nread  *int64
+}
```
nwt, line 109: Nit: The name partCounter describes how this is used, but I think a name that describes what it does, like countingReader, would be better.

zqd/handlers.go
```diff
@@ -424,60 +424,30 @@ func handleLogPost(c *Core, w http.ResponseWriter, r *http.Request) {
...
-				return
-			}
-		}
-	}
-	// send final status
-	if err := pipe.Send(op.Stats()); err != nil {
-		logger.Warn("error sending payload", zap.Error(err))
-		return
+
+	store := s.Storage()
```
nwt, line 433: NIt: Declare this closer to its use or, better yet, not at all (by saying `s.Storage().Write` below).

zqd/ingest/logreader.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/brimsec/zq/zio"
+	"github.com/brimsec/zq/zio/detector"
+	"github.com/brimsec/zq/zio/ndjsonio"
+	"github.com/brimsec/zq/zio/zngio"
+	"github.com/brimsec/zq/zng"
+	"github.com/brimsec/zq/zng/resolver"
+	"github.com/brimsec/zq/zqe"
+)
+
+type MultipartLogReader struct {
```
nwt, line 21: Nit: I'd just call this LogReader.

mattnibs:
@brimsec/zq this is ready for another review. Now that both multipart and the former post path methods exist, this can go in without any brim dependent changes.

ppl/zqd/ingest/multipartreader.go
```diff
@@ -0,0 +1,129 @@
...
+	m.stopErr = true
+}
+
+func (m *MulitpartLogReader) Read() (*zng.Record, error) {
+read:
+	if m.done {
+		return nil, nil
+	}
+	if m.zreader == nil {
+		zr, err := m.next()
```
alfred-landrum, line 50: A suggestion on handling the eof's: make the contract for `m.next()` that it always returns nil, nil when there's nothing else to read. Inside `MulitpartLogReader.next`, have it check if m.mr is nil, and if not, calls m.mr.NextPart(). If that errors with EOF, set m.mr to nil.

That will move all of the multipart logic handling inside next(), and remove the need for `done`.

ztest/shell.go
```diff
@@ -21,5 +22,7 @@ func RunShell(dir *Dir, bindir, script string, stdin io.Reader) (string, string,
 	cmd.Stdout = &stdout
 	cmd.Stderr = &stderr
 	err := cmd.Run()
+	fmt.Println("stderr", stderr.String())
+	// fmt.Println("stdout", stderr.String())
```
alfred-landrum, line 26:
```suggestion
```

ppl/zqd/ingest/multipartreader.go
mattnibs: Much better, thanks

api/api.go
```diff
@@ -181,6 +181,12 @@ type LogPostStatus struct {
 	LogReadSize  int64  `json:"log_read_size" unit:"bytes"`
 }
 
+type LogPostResponse struct {
+	Type      string `json:"type"`
+	BytesRead int64  `json:"bytes_read" unit:"bytes"`
+	Warnings  []string
```
nwt, line 187: `json:"warnings"`?

nwt (commented):
Delete the empty files named file and file2.

api/client/connection.go
```diff
@@ -341,16 +364,42 @@ func (c *Connection) LogPostStream(ctx context.Context, space api.SpaceID, paylo
...
 	}
-	return payloads.Error()
+	return c.LogPostWriter(ctx, space, opts, w)
+}
+
+func (c *Connection) LogPostWriter(ctx context.Context, space api.SpaceID, opts *LogPostOpts, writer *MultipartWriter) (api.LogPostResponse, error) {
+	req := c.Request(ctx).
+		SetBody(writer).
+		SetResult(&api.LogPostResponse{})
+	req.Header.Add("Content-Type", writer.ContentType())
```
nwt, line 387:
```suggestion
		SetResult(&api.LogPostResponse{}).
		SetHeader("Content-Type", writer.ContentType())
```

ppl/zqd/ingest/multipartreader.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/brimsec/zq/zio"
+	"github.com/brimsec/zq/zio/detector"
+	"github.com/brimsec/zq/zio/ndjsonio"
+	"github.com/brimsec/zq/zio/zngio"
+	"github.com/brimsec/zq/zng"
+	"github.com/brimsec/zq/zng/resolver"
+	"github.com/brimsec/zq/zqe"
+)
+
+type MulitpartLogReader struct {
```
nwt, line 20: Nit: Use same name for file and type.

api/client/multipartwriter.go
```diff
@@ -0,0 +1,139 @@
...
+
+func MultipartDataWriter(readers ...io.Reader) (*MultipartWriter, error) {
+	pr, pw := io.Pipe()
+	form := multipart.NewWriter(pw)
+	lw := &MultipartWriter{form: form, pr: pr, pw: pw}
+	lw.readers = readers
+	return lw, nil
+}
+
+func (l *MultipartWriter) SetJSONConfig(config *ndjsonio.TypeConfig) {
```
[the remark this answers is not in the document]
mattnibs: I can't believe I forgot to do this.

cmd/zapi/cmd/post/path.go
```diff
@@ -10,56 +10,53 @@ import (
...
-	Usage: "post [options] path...",
-	Short: "post log file(s) to a space",
-	New:   NewLogPost,
+var LogPath = &charm.Spec{
+	Name:  "postpath",
+	Usage: "postpath [options] path...",
+	Short: "post log paths to a space",
+	Long: `Post log paths to a space. ZQD will open the specifed paths and
+writing the data into the space. Posted paths must therefore be accessible by
+the specified zqd endpoint. Paths can be s3 URIs.`,
```
nwt, line 26:
```suggestion
	Long: `Post log paths to a space. Zqd will open the paths and
write the data into the space, so paths must be accessible by
zqd. Paths can be S3 URIs.`,
```

cmd/zapi/cmd/post/flags.go
```diff
@@ -0,0 +1,45 @@
...
+		return err
+	}
+	if sp != nil {
+		c.SetSpaceID(sp.ID)
+	} else {
+		// space already exists, fetch space ID
+		_, err := c.SpaceID()
+		return err
+	}
+	return nil
```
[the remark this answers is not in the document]
mattnibs: Nice!

cmd/zapi/cmd/post/pcap.go
```diff
@@ -12,63 +12,50 @@ import (
...
 	"github.com/brimsec/zq/cmd/zapi/format"
 	"github.com/brimsec/zq/pkg/display"
 	"github.com/mccanne/charm"
 )
 
-var PcapPost = &charm.Spec{
-	Name:  "pcappost",
-	Usage: "pcappost [options] path",
+var Pcap = &charm.Spec{
+	Name:  "postpcap",
```
nwt, line 22: Nit: s/Pcap/PostPcap/ so the variable name matches the command.

mattnibs: Good point, I'm going to change the rest of them for consistency as well.

alfred-landrum (approved):

Issue #861, opened 2020-06-05
ZQL expressions v2


henridf:
We've added recently expressions to ZQL (currently exposed in the put proc, and soon also in the group by proc). In the interest of expediency, some open design questions were intentionally put aside to be revisited later. This epic tracks those open design questions, as well as a few other issues that have come up since.

PR #59, opened 2019-11-25
Nested records

aswan:
I think this is ready for review.  Flattened field names in ndjson input are not yet working but this patch set is already quite large, I'd like to do that in a follow-up before this gets stale.

PR #1337, opened 2020-09-21
Fix wildcard field name example by introducing **

philrz:
I was re-reading the docs with an eye to making improvements, but I stumbled onto this place where we regressed. The tl;dr is that this doc example stopped working with the changes in #59, but no one caught it because it pre-dated Michael Brown's automation that ensured docs examples stayed in sync with `zq` outputs against out sample data. Then in #330 where Michael introduced that automation, he pasted in updated `zq` output that made the test pass, but the output no longer made sense with the accompanying text. It's been that way ever since.

Figuring out how to fix it was a trip down memory lane. I'd forgotten how around the time of #59, Andrew changed the behavior of `*=` such that it only matched against _top-level_ fields, and then introduced `**=` that also targeted fields inside of nested records. I had a long 1-on-1 Slack thread going with Andrew at the time about how I didn't love this change, but I guess conditions were such that I never argued the prior behavior back into existence nor did `**=` end up being documented. So I do my best to reconcile all that here.

I admit that it's not the easiest thing to read, but I feel like that's a reflection of the underlying complexity of the operators. If looked at as a likely `jq` substitute, I respect that having operators that target "levels" of nesting is sometimes desirable. OTOH for the security-centric use cases that the Brim app has in mind (and these ZQL docs were written for an audience of those security users, which is why Zeek data is used in the examples) I have difficulty imagining real world examples where it matters. All that said, the text & example before this PR were confusing & wrong, so it seems I'd have to make a choice between fixing it or not mentioning `*=` / `**=` at all, and I've chosen to attempt the former. If there's a silver lining, no _users_ have complained about this doc, which says to me that this may be power user functionality that users aren't reaching for yet, so we probably don't need to overthink it regardless.

I know we have a high-level to-do to polish more corner cases in ZQL (#861), so perhaps this is one to consider during that exercise.

### DIFF
--- a/zql/docs/search-syntax/README.md
+++ b/zql/docs/search-syntax/README.md
@@ -372,7 +372,7 @@ dns   1521912892.637238 CN9X7Y36SH6faoh8t 10.47.8.10 58340     10.0.0.100 53    
 
 ### Wildcard Field Names
 
-It's possible to search for the typed value across _any_ field by entering a wildcard in place of the field name. Two wildcard operators are available depending on how broad you want your search to be. The `*` operator matches all top-level fields of the value's type, and the `**` operator additionally matches such values when they appear nested within records.
+It's possible to search across _all_ fields of the value's data type by entering a wildcard in place of the field name. Two wildcard operators are available depending on how broad you want your search to be. The `*` operator matches all top-level fields of the value's type, and the `**` operator additionally matches such values when they appear nested within records.
 
 For example, the following search matches many `ssl` and `conn` events that contain the value `10.150.0.85` in `addr`-type fields of the `id` record, such as `id.resp_h`. It also matches `notice` events where it appears in `id.resp_h` and also `dst`, a top-level field also of the `addr` type. Compare this with our [bare word](#bare-word) example where we also matched as a substring of the `string`-type field named `certificate.subject`.
 
--- a/zql/docs/search-syntax/README.md
+++ b/zql/docs/search-syntax/README.md
@@ -80,7 +80,7 @@ The search result can be narrowed to include only events that contain certain va
 
 ### Bare Word
 
-The simplest form of such a search is a "bare" word (not wrapped in quotes), which will match against any field value that contains the word as a substring.
+The simplest form of such a search is a "bare" word (not wrapped in quotes), which will match against any field that contains the word, whether it's an exact match to the data type and value of a field or the word appears as a substring in a field.
 
 For example, searching across all our logs for `10.150.0.85` matches against events that contain `addr`-type fields containing this precise value (fields such as `tx_hosts` and `id.resp_h` in our sample data) and also where it appears within `string`-type fields (such as the field `certificate.subject` in `x509` events.)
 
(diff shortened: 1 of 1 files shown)


zql/docs/search-syntax/README.md
```diff
@@ -357,9 +357,29 @@ dns   1521912892.637238 CN9X7Y36SH6faoh8t 10.47.8.10 58340     10.0.0.100 53
 
 ### Wildcard Field Names
 
-Since the data type of the value is considered in field/value matches, it's possible to search for the value across any fields of the value's type by entering a wildcard (`*`) in place of the field name.
+Since the data type of the value is considered in field/value matches, it's possible to search for the value across any fields of the value's type by entering a wildcard in place of the field name. Two wildcard operators are available depending on how broad you want your search to be. The `*` operator matches all top-level fields of the value's type, and the `**` operator additionally matches such values when they appear nested within records.
```
henridf, line 360:
```suggestion
It's possible to search for the typed value across _any_ field by entering a wildcard in place of the field name. Two wildcard operators are available depending on how broad you want your search to be. The `*` operator matches all top-level fields of the value's type, and the `**` operator additionally matches such values when they appear nested within records.
```

zql/docs/search-syntax/README.md
```diff
@@ -357,9 +357,29 @@ dns   1521912892.637238 CN9X7Y36SH6faoh8t 10.47.8.10 58340     10.0.0.100 53
...
 
-For example, the following search finds many `ssl` and `conn` events that contain the value `10.150.0.85` in an `addr`-type field such as `id.resp_h`. Compare this with our [bare word](#bare-word) example where it also matched as a substring of the `string`-type field named `certificate.subject`.
+For example, the following search matches many `ssl` and `conn` events that contain the value `10.150.0.85` in `addr`-type fields of the `id` record, such as `id.resp_h`. It also matches `notice` events where it appears in `id.resp_h` and also `dst`, a top-level field also of the `addr` type. Compare this with our [bare word](#bare-word) example where we also matched as a substring of the `string`-type field named `certificate.subject`.
```
henridf, line 362:
```suggestion
For example, the following search matches many `ssl` and `conn` events that contain the value `10.150.0.85` in `addr`-type fields of the `id` record, such as `id.resp_h`. It also matches `notice` events where it appears in `id.resp_h` and also `dst`, a top-level field also of the `addr` type. Compare this with our [bare word](#bare-word) example where we also matched as a substring of the `string`-type field named `certificate.subject`. This highlights how search matches both on typed values and their string representation, whereas field/value match is stricter, and considers typed values only.
```

I think making the difference explicit is helpful (but perhaps with improved phrasing).

henridf:
I can't put this as a comment as it's outside of the diff, but I think this also needs adjustment:

> The simplest form of such a search is a "bare" word (not wrapped in quotes), which will match against any field value that contains the word as a substring.

Since a bare search will match against both the string value and the typed value. (Of course if the typed value is a string, those are the same).

henridf:
Another one I noticed above: 

> To match successfully, the value entered must be comparable to the data type of the named field. For instance, since id.resp_h is typically an addr-type field, an attempted field/value match id.resp_h=10.150.0.999 will return an error, since this is not valid IP address syntax.

`id.resp_h=10.150.0.999` is just a ZQL syntax error regardless of the type found in the LHS of the expression, be it `id.resp_h` or `_path` or `*`.

zql/docs/search-syntax/README.md
nwt: s/across any field/across all fields/

nwt (approved):

zql/docs/search-syntax/README.md
```diff
@@ -80,7 +80,7 @@ The search result can be narrowed to include only events that contain certain va
 
 ### Bare Word
 
-The simplest form of such a search is a "bare" word (not wrapped in quotes), which will match against any field value that contains the word as a substring.
+The simplest form of such a search is a "bare" word (not wrapped in quotes), which will match against any field that contains the word, whether it's an exact match to the data type and value of a field or the word appears as a substring in a field.
```
philrz, line 83: @henridf: I made this change after seeing your PR review comment. Please give it a look.

zql/docs/search-syntax/README.md
```diff
@@ -247,9 +247,24 @@ conn  1521912990.158539 ChhAfsfyuz4n2hFMe 10.239.34.35 56602     10.47.6.51 873
...
+   #### Output:
+   ```zq-output head:4
+   #0:record[host:bstring,count:uint64]
+   0:[0988253c66242502070643933dd49e88.clo.footprintdns.com;1;]
+   0:[10.47.21.1;1;]
+   0:[10.47.21.80/..;1;]
+   ...
+   ```
+
+   An attempted field/value match `host=10.47.21.1` would not match the event counted in the middle row of this table, since ZQL recognizes the bare value `10.47.21.1` as an IP address before comparing it to all the fields named `host` that it sees in the input stream. However, `host="10.47.21.1"` would match, since the quotes cause ZQL to treat the value as a string.
```
philrz, line 267: @henridf: I made this change after seeing your PR review comment. Please give it another look.

zql/docs/search-syntax/README.md
philrz: @henridf @nwt: I've incorporated your feedback, but made a small additional adjustment. I respect that the leading "Since the data type..." text was arguably redundant, so I understand the instinct to remove it. At the same time (and maybe I'm not giving our readers enough credit here) I'm concerned that the shorthand "typed value" could be misinterpreted as "the value I typed into the Search box in Brim" as opposed to "the value of the data type that the ZQL parser interpreted the value to be". I don't expect confusion from users who do programming, but I also expect some non-programmer users coming to Brim/ZQL after using incumbent search tools may need more of a helping hand. Anyway, please do give it another look and let me know what you think.

philrz:
@henridf @nwt: Well, I had an approval from @nwt, but I made enough changes that I feel like a re-review is justified. I'd appreciate if you could each give it a final look. Thanks!

zql/docs/search-syntax/README.md
henridf: 👍

zql/docs/search-syntax/README.md
henridf: Reading this I'm thinking of `literal`  in place of `bare value`. But your call, I'm guessing you might find that it's too progamming language-y for the goal of this doc.

henridf (approved):

zql/docs/search-syntax/README.md
philrz: Yes, I'm inclined to leave it as-is. I'm realizing that at some point we'll surely write a "Zesty Guide For Developers" where these same concepts are explained in programmer-friendly terms that use different terminology and use "foo/bar" examples rather than detailed Zeek ones that would make more sense to the security users. For now, I've got buy-in from the team that with our limited docs cycles we'll continue to make things as accessible as we can to security users who may not have a programming background. Thanks for helping me think it through.

Issue #1263, opened 2020-09-14
Read from multiple files to support time overlaps


alfred-landrum:
When multiple data files cover the same span of time, a reader should read from all of them, producing records by filtering, merging, and sorting the records from the required span.

PR #1340, opened 2020-09-22
add a scanner that combines multiple readers

alfred-landrum:
This is a subset of work for #1263 ; this will be used when scanning/reading from overlapping files in a zar archive.

### DIFF
--- a/scanner/combiner.go
+++ b/scanner/combiner.go
@@ -1,0 +1,54 @@
+package scanner
+
+import (
+	"context"
+
+	"github.com/brimsec/zq/ast"
+	"github.com/brimsec/zq/filter"
+	"github.com/brimsec/zq/pkg/nano"
+	"github.com/brimsec/zq/zbuf"
+	"github.com/brimsec/zq/zng"
+)
+
+type combinerScanner struct {
+	combiner zbuf.Reader
+	scanners []Scanner
+}
+
+// NewCombinerScanner returns a Scanner that combines the records scanned from
+// a set of filtered readers.
+func NewCombinerScanner(ctx context.Context, readers []zbuf.Reader, spans []nano.Span, cmp zbuf.RecordCmpFn, f filter.Filter, filterExpr ast.BooleanExpr) (Scanner, error) {
+	if len(readers) != len(spans) {
+		panic("length mismatch between readers and spans")
+	}
+	scanners := make([]Scanner, len(readers))
+	scanReaders := make([]zbuf.Reader, len(readers))
+	for i, r := range readers {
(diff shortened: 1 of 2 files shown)


scanner/combiner.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+type combinerScanner struct {
+	combiner zbuf.Reader
+	scanners []Scanner
+}
+
+// NewCombinerScanner returns a Scanner that combines the records scanned from
+// a set of filtered readers.
+func NewCombinerScanner(ctx context.Context, readers []zbuf.Reader, spans []nano.Span, cmp zbuf.RecordCmpFn, f filter.Filter, filterExpr ast.BooleanExpr) (Scanner, error) {
```
nwt, line 20: Nit: This imports as scanner.NewCombinerScanner, so including Scanner in the name is redundant. How about just calling it Combiner (so it imports as scanner.Combiner)?

scanner/combiner.go
```diff
@@ -0,0 +1,54 @@
...
+	}, nil
+}
+
+func (c *combinerScanner) Pull() (zbuf.Batch, error) {
+	return zbuf.ReadBatch(c, BatchSize)
+}
+
+func (c *combinerScanner) Read() (*zng.Record, error) {
+	return c.combiner.Read()
+}
```
nwt, line 46:
```suggestion
func (c *combinerScanner) Pull() (zbuf.Batch, error) {
	return zbuf.ReadBatch(c.combiner, BatchSize)
}
```

zbuf/batch.go
```diff
@@ -72,3 +72,32 @@ func CopyPuller(w Writer, p Puller) error {
...
+	return &pullerReader{p: p}
+}
+
+type pullerReader struct {
+	p     Puller
+	batch Batch
+	idx   int
+}
+
+var _ Reader = (*pullerReader)(nil)
```
nwt, line 86: Nit: PullerReader returns a Reader, so this is redundant.

zbuf/batch.go
```diff
@@ -72,3 +72,32 @@ func CopyPuller(w Writer, p Puller) error {
...
+
+func (r *pullerReader) Read() (*zng.Record, error) {
+	if r.batch == nil {
+		batch, err := r.p.Pull()
+		if err != nil || batch == nil {
+			return nil, err
+		}
+		r.batch = batch
+		r.idx = 0
+	}
```
nwt, line 96: Although we try to avoid zero-length batches, they do sometimes occur, so you might want to handle them.

nwt (approved):

PR #1341, opened 2020-09-22
pcapng: fix bad parse interface bug

### LINKED ISSUE #1057: invalid pcap: bad option length
I stumbled onto the attached pcap. Wireshark, `tcpdump`, and Zeek all read it without complaint, but when our `pcap` tool tries to index it, it returns the error message:

```
$ pcap index -r full.pcapng 
invalid operation: invalid pcap: bad option length
```

The above is reproducible as of `pcap` commit `e0bc881`.

Perhaps it truly _is_ bad and we're just the only tool with code sophisticated enough to notice it. However, since the other tools don't seem to have a problem with it, I figure it's worth checking if this isn't maybe a bug on our part... or, if it is truly "bad", if it's something where we could stand to be more tolerant like the other tools appear to be.

[full.pcapng.zip](https://github.com/brimsec/zq/files/5012519/full.pcapng.zip)

mattnibs:
The parseInterfaceDescriptor was not advancing the read buffer
before reading through the list of options for the block. This
was causing a red-herring bad-option-length error; the reader was
not actually reading the option list block.

closes #1057


nwt (approved):
I hate to be that person but should we add a test?

mattnibs:
> I hate to be that person but should we add a test?

This is a bit of a pain generating a test case, but I will look into it.

mattnibs:
@nwt I added a test case. Need to be done.

PR #1345, opened 2020-09-22
create spill package

### LINKED ISSUE #1346: Refactor spill code in sort so it can be reused generically and specifically by fuse

mccanne:
This commit refactors the spill code used by sort and groupby so
its interdependencies are cleanly delineated and it can be reused
by other packages without reaching into the sort package.

This also separates spill.File into a file piece and a peeker piece so
single-file spilling can be implemented without necessarily needing
to sort each chunk (e.g., as will be used by the fuse proc).

Resolves #1346


nwt (commented):
Take this with a grain of salt given my relationship to the runManager stuff, but I'm afraid I find the new names and structure less clear than the current form. But if others like what's here, I don't object.

mccanne:
> Take this with a grain of salt given my relationship to the runManager stuff, but I'm afraid I find the new names and structure less clear than the current form. But if others like what's here, I don't object.

 I changed the names back.  Does this address your concern or is it more about how I refactored the stuff inside the new package so I can use it with fuse?  I understand where you came from and how you like thinking in terms of "sorting runs" but I am wondering if the name changes I was proposing is more in line with how open systems like spark and hadoop talk about their spilling patterns?

mattnibs:
Respectfully, I was never a fan of the name `runManager`. I like Spiller better.

proc/spill/file.go
```diff
@@ -0,0 +1,59 @@
...
+	if _, err := f.file.Seek(0, 0); err != nil {
+		f.closeAndRemove()
+		return err
+	}
+	f.Reader = zngio.NewReader(bufio.NewReader(f.file), zctx)
+	return nil
+}
+
+// closeAndRemove closes and removes the underlying file.
+// XXX errors are ignored.
```
henridf:
```suggestion
// Errors are ignored.
```

proc/spill/merge.go
```diff
@@ -12,42 +12,46 @@ import (
...
+func (r *Merger) Cleanup() {
 	for _, run := range r.runs {
 		run.closeAndRemove()
 	}
 	os.RemoveAll(r.tempDir)
 }
 
-// CreateRun creates a new run containing the records in recs.
-func (r *RunManager) CreateRun(recs []*zng.Record) error {
+// Spiil spills a new run of records to a file in the Merger's temp directory.
```
henridf:
```suggestion
// Spill spills a new run of records to a file in the Merger's temp directory.
```

proc/spill/peeker.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+	first, err := f.Read()
+	if err != nil {
+		f.closeAndRemove()
+		return nil, err
+	}
+	return &peeker{f, first}, nil
+}
+
+// read is like Read but returns eof at the last record so manager can
```
henridf:
```suggestion
// read is like Read but returns eof at the last record so Merger can
```

henridf (approved):
Fwiw I prefer the new names from the first commit.

nwt (approved):
The people have spoken for 17a7c19de254d202005f549fb23e709f28837a29!

Issue #1215, opened 2020-09-03
Deterministic ordered merge


henridf:
As noted in #1204:

> Ordered merge does not have sufficient information to sort stably: when multiple records with same sort field are present at different upstreams, it does not know what their original order (*) was.

We need to make ordered merge deterministic. Note "deterministic" and not "stable", because in a case such as multiple input files that overlap in the sort key, there is no stable ordering in the data. This ordering must be imposed externally.

alfred-landrum:
In recent discussions, @henridf suggested if the sort field value is the same, we can compare the entire zcode.Bytes of the record, which has the desirable quality of not requiring any external information to be present to provide the ordering.

Issue #1132, opened 2020-08-20
zqd: Support archive space pcap ingest support


mattnibs:

alfred-landrum:
We ran an into issue when discussing how to handle this in the "drag a pcap" import of Brim: the "long conn" problem of Zeek. That is: if you feed a pcap to the zeek runner, and occasionally take a snapshot of the zeek logs, you will occasionally see timestamps prior to a previous snapshot, since the timestamp for a zeek conn record is the connection start, but the record is emitted at connection end. (@mattnibs : please make sure I'm describing this accurately).

We want to feed snapshots of the transformed zeek logs into the archive, so that a Brim user can examine & query the available data without waiting for the entire pcap to be read. We can do that by taking a snapshot of the logs, converting them  to zng, then import the delta of records from the last snapshots zng file (so that we don't import duplicate records). The new records may have timestamps that overlap with existing imported data - which we don't yet handle. This led to the discussion & design in #1183 .

philrz:
Verified with Brim commit `480f4ca` that uses `zq` tooling at commit `51a0a44`.

As of the last GA combination of Brim tagged `v0.18.0` with `zq` tagged `v0.22.0`, importing a pcap to an archive Space was not possible. With my Brim app running and hence the `zqd` also started:

```
$ zapi -version
Version: v0.22.0

$ zapi new -k archivestore mypcap
mypcap: space created

$ zapi -s mypcap pcappost hello.pcapng
posting...
status code 400: space does not support pcap import
```

Now with Brim commit `480f4ca` with `zq` tooling at commit `51a0a44`:

```
$ zapi -version
Version: v0.22.0-53-g51a0a44

$ zapi new -k archivestore mypcap
mypcap: space created

$ zapi -s mypcap pcappost hello.pcapng
100.0% 12.40KB/12.40KB
/Users/phil/pcap/hello.pcapng: pcap posted
```

At this point, as shown in the attached video, I can "Reload" in my Brim app and see the archive Space I just created, and successfully extract a flow from the pcap.

[Verify.zip](https://github.com/brimsec/zq/files/5385657/Verify.zip)

There's still an open issue #1488 that's necessary to get progress updates working in the app with archive Spaces. @alfred-landrum discusses a branch-based test in https://github.com/brimsec/zq/pull/1450#pullrequestreview-508419232 that validates that we're on the right track with the app, modulo that #1488 enhancement. I'll plan to verify the full app experience when that lands.

Thanks @mattnibs!

PR #1348, opened 2020-09-23
support overlapping data files in an archive

### LINKED ISSUE #1263: Read from multiple files to support time overlaps

alfred-landrum:
When the data files for a query have overlapping time spans, scan & read the overlapping files, merging their records in the archive's order.

On top of #1330
Closes #1263


archive/multisource.go
```diff
@@ -0,0 +1,181 @@
...
+	"github.com/brimsec/zq/pkg/nano"
+	"github.com/brimsec/zq/scanner"
+	"github.com/brimsec/zq/zbuf"
+	"github.com/brimsec/zq/zio"
+	"github.com/brimsec/zq/zio/detector"
+	"github.com/brimsec/zq/zio/zngio"
+	"github.com/brimsec/zq/zng/resolver"
+)
+
+// A SpanInfo is a logical view of the records within a time span, stored
```
alfred-landrum, line 20: I don't think 'SpanInfo' is a good name for this type anymore, but before I change it & the relevant comments, wanted to get buy in: I was thinking of calling this a 'segment'. (It doesn't need to be exported anymore.)

archive/import.go
```diff
@@ -132,7 +131,6 @@ func (d *importDriver) close() error {
 		return err
 	}
 	d.dataFileWriter = nil
-
```
mattnibs, line 135: I can't comment below but with these changes can't we get rid of the `sort` proc we use for import now? In fact can't we just not use the `driver` and make the `importDriver` into a regular ole' `zbuf.Reader`?

alfred-landrum: We need the records in each chunk file to be sorted according to the archive's sort direction. Perhaps you're wondering if we could just take N records at a time from a reader, sort them, and write them, instead of sorting the entire input? If so, we'd potentially be writing overlapping chunks that we wouldn't have if we sorted all of the input at once.

mattnibs (commented):
Seems like in future work we should have a version of `scanner.Combiner` that produces memory pooled `Batches` to pass into the flowgraph.

archive/import.go
mattnibs: > Perhaps you're wondering if we could just take N records at a time from a reader, sort them, and write them, instead of sorting the entire input?

Yes this exactly.

 > If so, we'd potentially be writing overlapping chunks that we wouldn't have if we sorted all of the input at once.

Sure but we wouldn't have to incur the cost of writing the entire stream onto some temp location on disk, then read and write them back to their final destination. Ingest would be a lot faster- it's the beauty of lsm. 

Also, if we go this route, I cannot see how we would allow query during ingest for brim pcap post. I suppose we could arbitrarily break up the stream and do a new import, but in the end this would be the same thing as what I'm suggesting.

archive/overlap.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+// spanToFirstLast returns the timestamps that whose closed interval
+// is represented by span s. It assumes s.Dur is greater than zero.
+func spanToFirstLast(dir zbuf.Direction, s nano.Span) (first, last nano.Ts) {
+	a := s.Ts
+	b := s.Ts + nano.Ts(s.Dur-1)
+	if a > b {
+		a, b = b, a
+	}
```
mattnibs, line 82: I `s.Dur` is greater than zero (as it should) you do not need this check.

archive/multisource.go
```diff
@@ -0,0 +1,181 @@
...
+// in one or more Chunks.
+type SpanInfo struct {
+	First nano.Ts // timestamp of first record in this span
+	Last  nano.Ts // timestamp of last record in this span
```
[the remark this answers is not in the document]
alfred-landrum: I could: Chunk has 'First,Last' because it felt natural to use the first and last record's timestamps as part of a file's name, and because that also communicated the ordering of the records in the file. But that's not true for SpanInfo, its time is always just used as a filter.

archive/overlap.go
```diff
@@ -0,0 +1,142 @@
...
+			}
+		}
+		dst = append(dst, src[i])
+	}
+	return
+}
+
+// closedSpan returns a span for the closed interval of [x,y].
+func closedSpan(x, y nano.Ts) nano.Span {
+	return nano.Span{Ts: x, Dur: 1}.Union(nano.Span{Ts: y, Dur: 1})
```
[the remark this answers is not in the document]
alfred-landrum: I used `Union` here because it works regardless of whether x > y. For example, a Chunk could have a First,Last pair of "10,5" because its data is in reverse timestamp order.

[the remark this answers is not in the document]
mattnibs: Oh, I see. Cool.

archive/import.go
alfred-landrum: Let's discuss this on #1132 ; this is good topic, but I don't think we're going to want to add more changes to this PR.

archive/multisource.go
alfred-landrum: ok, I unexported, so it's just `spanInfo` for now.

mattnibs:
I've read through this pr sufficiently to understand to what this is doing. There is some complicated logic in determining boundaries for SpanInfos that are then used to create a spanScanners that generate sorted batches of records. Why do we need this delineation?

Why not do something like this:
```golang
type ChunkReader struct {
    peeker zbuf.Peeker
    span nano.Span
    ...
}

func (c *Chunk) NextTs() (nano.Ts, error) {
    if c.peeker == nil {
        return span.Ts, nil // or end depending on direction
    }
    rec, err := c.peeker.Peek()
    if rec == nil || err != nil {
        c.peeker.Close()
        return 0, err
    }
    return rec.Ts, nil
}

func (c *Chunk) Read() (*zng.Record, error) {
    if (c.peeker == nil) {
        r, err := iosrc.NewReader(c.URI)
        if err == nil {
            return err
        }
        c.peeker = zbuf.NewPeeker(r)
    }
    return c.peeker.Read()
}
```
Then create something similar to `zscanner.Combiner` that takes a sorted []Chunks and uses chunk.NextTs to read from the next right chunk?

alfred-landrum:
> Then create something similar to `zscanner.Combiner` that takes a sorted []Chunks and uses chunk.NextTs to read from the next right chunk?

I may not understand the scope of what you're proposing. I think you're probably not saying we should implement `zbuf.Combiner` with `zbuf.Peeker`; I think you're suggesting that we open all of the Chunks for a query, so we can peek to see which one to read next? If so, the issue is that you need to open all the chunks.

A rationale for the span/boundary calculation is that it a) determines the minimum set of Chunks needed for the full range of query and b) for overlap occurrences within the full query, a SpanInfo contains the minimum set of Chunks to simultaneously scan/read/merge their data.

archive/finder.go
```diff
@@ -46,21 +46,21 @@ func AddPath(pathField string, absolutePath bool) FindOption {
 			{"first", zng.TypeTime},
 			{"last", zng.TypeTime},
 		}
-		opt.addPath = func(ark *Archive, si SpanInfo, rec *zng.Record) (*zng.Record, error) {
+		opt.addPath = func(ark *Archive, chunk Chunk, rec *zng.Record) (*zng.Record, error) {
 			var path string
 			if absolutePath {
 				if ark.DataPath.Scheme == "file" {
-					path = si.LogID.Path(ark).Filepath()
+					path = chunk.Path(ark).Filepath()
```
nwt, line 53: Did you consider adding an `ark *Archive` field to Chunk? Then you wouldn't have to pass ark to the Chunk methods.

archive/multisource.go
```diff
@@ -0,0 +1,174 @@
...
+	if len(altPaths) == 1 && altPaths[0] == "_" {
+		altPaths = nil
+	}
+	return &multiSource{
+		ark:      ark,
+		altPaths: altPaths,
+	}
+}
+
+func (ams *multiSource) OrderInfo() (string, bool) {
```
nwt, line 125: Nit: Maybe change the receiver name int the multiSource methods to m?

archive/overlap.go
```diff
@@ -0,0 +1,135 @@
...
+				continue outer
+			}
+		}
+		dst = append(dst, src[i])
+	}
+	return
+}
+
+// closedSpan returns a span for the closed interval of [x,y].
+func closedSpan(x, y nano.Ts) nano.Span {
```
nwt, line 67: Span aren't closed or open—they're always half-open—so this doesn't seem like the best name for this. How about firstLastToSpan for symmetry with spanToFirstLast?

archive/walk.go
```diff
@@ -273,110 +224,91 @@ func (l LogID) Path(ark *Archive) iosrc.URI {
...
-		paths: paths,
-	}
+func (c Chunk) Range(ark *Archive) string {
+	return fmt.Sprintf("[%d-%d]", c.First, c.Last)
 }
 
-func (ams *multiSource) OrderInfo() (string, bool) {
-	if len(ams.paths) == 1 && ams.paths[0] == "_" {
-		return "ts", ams.ark.DataSortDirection == zbuf.DirTimeReverse
+func chunkTsCompare(dir zbuf.Direction, iTs nano.Ts, iKid ksuid.KSUID, jTs nano.Ts, jKid ksuid.KSUID) bool {
```
nwt, line 267: Nit: Compare functions usually return an int (like bytes.Compare), but this is a less function (like sort.Interface.Less), so I think chunkTsLess would be a better name.

archive/walk.go
```diff
@@ -273,110 +227,84 @@ func (l LogID) Path(ark *Archive) iosrc.URI {
...
-				return err
-			}
-		}
-		return nil
-	})
+// ZarDir returns a URI for a directory specific to this data file, expected
+// to hold microindexes or other files associated with this chunk's data.
+func (c Chunk) ZarDir(ark *Archive) iosrc.URI {
+	return ark.DataPath.AppendPath(string(c.LogID() + zarExt))
 }
```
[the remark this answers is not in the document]
alfred-landrum: There's always an `*Archive` available in the context where they are used.

archive/multisource.go
```diff
@@ -0,0 +1,174 @@
...
+	readers := make([]zbuf.Reader, len(si.chunks))
+	for i, chunk := range si.chunks {
+		rc, err := iosrc.NewReader(ctx, chunk.Path(ark))
+		if err != nil {
+			return nil, err
+		}
+		closers[i] = rc
+		readers[i] = zngio.NewReader(rc, zctx)
+	}
+	sn, err := scanner.NewCombiner(ctx, readers, zbuf.RecordCompare(ark.DataSortDirection), f, filterExpr, si.span)
```
mattnibs, line 93: This is a bit of an edge case, but if want raw, in-order searches to be deterministic shouldn't the `filterExpr` passed into the Combiner here take into account the chunk id, so it can consistently sort equal timestamps from separate chunks?

mattnibs:
> I think you're suggesting that we open all of the Chunks for a query, so we can peek to see which one to read next? If so, the issue is that you need to open all the chunks.

Perhaps I should have called `Chunk` in my example above `ChunkReader`, but I think you are getting the general idea.

You wouldn't have to open all files instantly- my example leverages the span info we have for each chunk so it only opens a chunk when the combiner reaches the chunk's time window. When all of a chunk's data has been read, the chunk's reader is closed. This way you only have open the files that you need. 

This would also mean that you would have to write a separate `Combiner` specific to archive that determines what `zbuf.Reader` to read from next by calling `NextTs()` instead of `zbuf.Peeker.Peek`.

archive/multisource.go
alfred-landrum: You're right; this another instance of this issue: https://github.com/brimsec/zq/issues/1215#issuecomment-698356017 . As mentioned there, a useful technique is to use the zcode.Bytes of the record itself to break sort field ties.

mattnibs: Even better, this wouldn't include the complication of having to some how include chunkid in the combiner.

mattnibs (approved):
Sweet!

archive/walk.go
```diff
@@ -273,110 +224,91 @@ func (l LogID) Path(ark *Archive) iosrc.URI {
 	return ark.DataPath.AppendPath(string(l))
 }
 
-type SpanInfo struct {
-	First       nano.Ts // timestamp of first record in this span
-	Last        nano.Ts // timestamp of last record in this span
-	LogID       LogID
-	RecordCount int64
+type Chunk struct {
```
nwt, line 227: This can wait for another PR, but a Chunk represents a data file, right? So why not call it DataFile?

archive/multisource.go
```diff
@@ -0,0 +1,174 @@
...
+			rc.Close()
+			return nil, err
+		}
+		return &scannerCloser{sn, rc}, nil
+	}
+	closers := make([]io.Closer, len(si.chunks))
+	defer func() {
+		if err != nil {
+			for _, c := range closers {
+				c.Close()
```
nwt, line 80: This will do the equivalent of io.Closer(nil).Close() if an error occurs inside the si.chunks loop.

archive/overlap.go
```diff
@@ -0,0 +1,125 @@
...
+
+	"github.com/brimsec/zq/pkg/nano"
+	"github.com/brimsec/zq/zbuf"
+)
+
+// mergeChunksToSpans takes an unordered set of Chunks with possibly overlapping
+// spans, and returns an ordered list of SpanInfos, whose spans will be bounded
+// by filter, and where each spanInfo contains one or more Chunks whose data
+// falls into the spanInfo's span.
+func mergeChunksToSpans(chunks []Chunk, dir zbuf.Direction, filter nano.Span) []spanInfo {
```
nwt, line 14: Nit: To avoid confusion with filter.Filter, maybe rename filter to span?

archive/multisource.go
alfred-landrum: ah, thanks for the catch, I messed that up when I preallocated the slice length.

archive/overlap.go
alfred-landrum: I've gone back & forth on this: I'd like to keep it as filter, otherwise the comments above become really hard to read. ('the spaninfo's span is bounded by span').

archive/overlap.go
```diff
@@ -0,0 +1,125 @@
...
+
+import (
+	"sort"
+
+	"github.com/brimsec/zq/pkg/nano"
+	"github.com/brimsec/zq/zbuf"
+)
+
+// mergeChunksToSpans takes an unordered set of Chunks with possibly overlapping
+// spans, and returns an ordered list of SpanInfos, whose spans will be bounded
```
nwt, line 11:
```suggestion
// spans, and returns an ordered list of spanInfos, whose spans will be bounded
```

archive/overlap.go
nwt: "... ordered list of spanInfos, each bounded by span and containing one or more ..."?

nwt: But this is just a nit, so definitely feel free to leave it as is.

archive/overlap_test.go
```diff
@@ -0,0 +1,241 @@
...
+func kid(s string) ksuid.KSUID {
+	var b [20]byte
+	copy(b[:], []byte(s))
+	k, _ := ksuid.FromBytes(b[:])
+	return k
+}
+
+func importTzng(t *testing.T, ark *Archive, s string) {
+	zctx := resolver.NewContext()
+	reader := tzngio.NewReader(bytes.NewReader([]byte(s)), zctx)
```
nwt, line 29:
```suggestion
	reader := tzngio.NewReader(strings.NewReader(s), zctx)
```

nwt (approved):
:shipit:

archive/walk.go
alfred-landrum: Interesting question. A Chunk represents both the data & seek index file, would also hold information about the files it 'masks' in the future. I'll leave as-is for this PR, but can discuss more.

PR #1350, opened 2020-09-23
zqd: Add helpful index page

### LINKED ISSUE #1209: Provide guidance on http://localhost:9867 instead of a 404
In a [thread on Slack](https://brimsec.slack.com/archives/C010DR0HHMF/p1599084438005600), a new community user reported having installed/started Brim on Linux, saw a process listening on `localhost:9867`, and therefore tried to connect to `http://localhost:9867` from a browser on the assumption that Brim was accessed this way rather than through the Electron-based client. Here's what such a user sees today:

image

With remote/standalone `zqd` likely to become a more common config in the days ahead, the exposure for this kind of confusion seems likely to increase. Therefore it would seem a good idea to respond in this situation with more helpful guidance. Perhaps something like the following (including the hyperlinks):

> A [zqd](https://github.com/brimsec/zq/tree/master/cmd/zqd) daemon is listening on this host/port.
>
> If you're a [Brim](https://www.brimsecurity.com/) user, connect to this host/port from the [Brim application](https://github.com/brimsec/brim) in the graphical desktop interface in your operating system (not a web browser).
>
> If your goal is to perform command line operations against this zqd, use the [zapi](https://github.com/brimsec/zq/tree/master/cmd/zapi) client

mattnibs:
closes #1209

### DIFF
--- a/zqd/handler.go
+++ b/zqd/handler.go
@@ -3,6 +3,7 @@ package zqd
 import (
 	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
 
 	"github.com/brimsec/zq/zqd/api"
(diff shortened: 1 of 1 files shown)


mattnibs:
Screen Shot 2020-09-23 at 2 56 19 PM

philrz:
I dig it! I tried out on the branch and all the links work. 👍

henridf (approved):

PR #1355, opened 2020-09-24
add spilling to fuse proc

### LINKED ISSUE #1270: Allow fuse to spill-to-disk
When the `fuse` processor hits its memory limit, it should spill to disk.

mccanne:
This commit adds spill logic to the fuse proc triggered when its memory
footprint exceeds a threshold.  We changed the fuse proc implementation
to use a goroutine so it can cleanly teardown the spill file when
the proc context is canceled in addition to providing parallelism
between scanning and spilling.

The -fusemem proc flag option is analogous to -sortmem and controls
how much memory fuse will use before spilling.

We also added a Read method in zbuf.Array to implement zbuf.Reader,
which is handy for turning a slice of zng records into a reader.

Closes #1270

### DIFF
--- a/proc/fuse/fuse.go
+++ b/proc/fuse/fuse.go
@@ -2,8 +2,10 @@ package fuse
 
 import (
 	"errors"
+	"sync"
 
 	"github.com/brimsec/zq/proc"
+	"github.com/brimsec/zq/proc/spill"
 	"github.com/brimsec/zq/zbuf"
 	"github.com/brimsec/zq/zcode"
 	"github.com/brimsec/zq/zng"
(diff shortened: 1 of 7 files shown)


proc/fuse/fuse.go
```diff
@@ -2,65 +2,155 @@ package fuse
...
+			return p.finish()
+		}
+		if err := p.writeBatch(batch); err != nil {
+			return err
+		}
+	}
+}
+
+func (p *Proc) writeBatch(batch zbuf.Batch) error {
+	l := batch.Length()
```
mattnibs, line 75: Should we add a `defer batch.Unref()` here?

mccanne: Good point.  This entailed adding a bit of logic to stash to do the Keep when stashing and not spilling.

proc/fuse/fuse.go
```diff
@@ -69,46 +164,61 @@ type slot struct {
...
 
-func (p *Proc) finish(types resolver.Slice, recs []*zng.Record) (zbuf.Batch, error) {
+func (p *Proc) finish() error {
 	uber := newSchema()
 	// positionsByID provides a map from a type ID to a slice of integers
 	// that represent the column position in the uber schema for each column
 	// of the input record type.
-	positionsByID := make([][]int, len(types))
-	for _, typ := range types {
+	p.slotByID = make([][]int, len(p.types))
```
alfred-landrum, line 172: The comment above this refers to the previous variable, `positionsByID`.

alfred-landrum (approved):

mccanne:
This is now waiting on #1391. I will rebase to that before merging this PR.

PR #1368, opened 2020-09-25
temporarily disable suricata during system tests

alfred-landrum:


philrz:
Closed since #1369 seems to do the trick instead.

PR #1370, opened 2020-09-27
JSON types: remove "_path must be first column" check

henridf:
This commit removes the validation check that ensured that _path was
present as the first field of descriptors. This check is a reflection of our
zeek-onlyness back when this feature was first introduced. Now with the
impending arrival of Suricata and other data types, this hardcoded
check no longer belongs.

If we find ourselves needing such a validation, we can re-introduce
something more general once we have a few more ingest types under our
belt.

### DIFF
--- a/zio/ndjsonio/config.go
+++ b/zio/ndjsonio/config.go
@@ -52,10 +52,6 @@ func (conf TypeConfig) Validate() error {
 				return fmt.Errorf("descriptor %s has field ts with wrong type %s", name, col["type"])
 			}
 		}
-		col := desc[0].(map[string]interface{})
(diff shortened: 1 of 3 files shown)


mattnibs (commented):
lgtm, but wanted to check if this is going to cause issues for the brim client when ingesting pcap logs?

henridf:
@mattnibs not unless I'm missing something: when ingesting pcap logs, we'll use a json type schema for suricata thta has no `_path` (the motivation for this PR), and the zeek logs will continue to go from zeek to zng without any json involved.

mccanne (approved):
Zeek-only is so 2020.  Onward and upward!

PR #1374, opened 2020-09-27
Allow for singular "week" like we do for other time units

philrz:
While documenting aggregate functions, I noticed that the singular `week` without a leading number was not available, though it is for all the other supported time units. I couldn't see a reason for leaving this out, so I've followed the existing pattern to make this consistent.


henridf (approved):
Change looks good. Please re-build parsers by running `make` from inside `zql/` before merging.

philrz:
Thanks for the tip @henridf. Parsers are now committed and tests are passing. Merging.

PR #1375, opened 2020-09-27
refactor package expr

mccanne:
My profuse apologies up front for this massive PR.  I really thought long 
and hard about breaking this into smaller chunks and made some failed 
attempts, but the interdependencies between the refactoring changes 
were just too much so I went for the whole shebang.

THIS COMMIT BREAKS BACKWARD COMPAT WITH THE ZNG FILE FORMAT.

 Now that we're addressing changes to the ZNG spec to bring it to beta status,
I thought this PR would be a good time to introduce breaking changes, shortly 
after our previous release on Friday.  I think we have a handle on the breaking changes
we  want to make and I think I Ican get them all in by the next release, so after this,
hopefully we won't have anymore backward-compat changes (though we will have a few
more additions that don't break compat as we go from beta to stable of the
ZNG format).

===

This commit refactors the expr package to be more extensible and
performant.  The zngnative package has been eliminated.  It had
good intentions but every expression resulted in the allocation
of an empty interface to hold a go native value resulting in lots
of traffic to the GC.  The implementation keeps scratch buffers
around and reuses them and aside from strings and IPs, does not
allocate any memory to evaluate expressions.  There is a "keep"
flag passed to expr.CompileExpr() that says whether you want
values that get overwritten on subsequent calls to Eval()
(e.g., ok for filtering and aggregations) or whether you want to
hold onto the values (like groupby and sort do with field keys).

We also changed the field implementation and cleaned up the AST.
There was overlap between the old FieldCall stuff and the generic
expression code, which is now eliminated.  This also makes the
grammar a bit easier to reason about as we improve and extend it.
There is now one inteface type, expr.Evaluator, to take over for
all the function closures we previously had.  This makes it
easier to cookie-cutter and add new expression functionality.
Also, the field expressions are now just an instance of
expr.Evaluator so everything is now clean and uniform.

While we were changnig the coercion logic around and getting
rid of zngnative, we laid out the ZNG type codes a bit differently
and upudated the spec.  Since we were making incompatible changes,
we changed the "byte" type to "uint8" and added "int8" so all
the integer types are now orthogonal.  Also, the integers are
all encoded using the same varint logic now, which greatly
simplifies value encoding and decoding in the coercion engine.

Most of the tests that were updated simply accommdate the change
to the zng binary format.


mccanne:
FYI the failing test is because the zng output format changed.  I don't know if we want to bypass this for a week or so while I finish making breaking changes.  Then, once the dust settles, hopefully by the next release, we can re-enable this check?

@philrz ?

philrz:
Whatever you're doing here is already having positive knock-on effects. While writing aggregate function docs, I saw this crash with GA `zq` tagged `v0.22.0` when run against the full 70 GB wrccdc data set:

```
$ zq -f table 'every 1 week count() | sort -r ts' *.log.gz
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x13efe36]

goroutine 42 [running]:
github.com/brimsec/zq/zngnative.CoerceNativeToTime(0x0, 0x0, 0x0, 0x0, 0x0, 0x289a7d0)
	/Users/phil/work/zq/zngnative/coerce.go:202 +0x26
github.com/brimsec/zq/expr.timeTrunc(0xc000abbc00, 0x2, 0x2, 0x1724f80, 0xc0002f8028, 0x0, 0x0, 0x203000, 0x1)
	/Users/phil/work/zq/expr/functions.go:454 +0x6f
github.com/brimsec/zq/expr.compileFunctionCall.func1(0xc000a41640, 0xc0008a92c0, 0x24d9200, 0xc0008ba948, 0xc000680000, 0x289a7d0, 0x0)
	/Users/phil/work/zq/expr/expr.go:938 +0x250
github.com/brimsec/zq/expr.CompileExpr.func1(0xc000a41640, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
	/Users/phil/work/zq/expr/expr.go:48 +0x63
github.com/brimsec/zq/proc/groupby.(*Aggregator).Consume(0xc000354000, 0xc000a41640, 0xc000a41640, 0x0)
	/Users/phil/work/zq/proc/groupby/groupby.go:396 +0x1c9
github.com/brimsec/zq/proc/groupby.(*Proc).run(0xc000090200)
	/Users/phil/work/zq/proc/groupby/groupby.go:269 +0xa4
created by github.com/brimsec/zq/proc/groupby.(*Proc).Pull.func1
	/Users/phil/work/zq/proc/groupby/groupby.go:240 +0x3e
```

Since I'd just seen your reference to eliminating the zngnative package, and zngnative appeared in this error, I tried with commit `1ddde2b` for this branch and can see it's already fixed.

```
$ zq -f table 'every 1 week count() | sort -r ts' *.log.gz
TS                COUNT
1541635200.000000 3131
1521676800.000000 475593430
1490227200.000000 96242990
0.000000          26
```

👍

mccanne:
Hmm, I just noticed a problem with integer 0 turning into unset in expressions.  Will fix...

mccanne:
> Hmm, I just noticed a problem with integer 0 turning into unset in expressions. Will fix and adjust the tests...

Fixed.

philrz:
Regarding the failing ZNG output check, I've put up https://github.com/brimsec/zq-sample-data/pull/16 that has the test data changes necessary to make the automation pass, along with draft PR https://github.com/brimsec/zq/pull/1383 that shows it passing "green" in a scratch branch. There's comments in https://github.com/brimsec/zq-sample-data/pull/16 about the options for the order in which the changes should land. Overall, though, it's just a small loose end we can resolve right as we're merging _this_ PR. The long pole in this effort remains the actual review of the code. 😜

expr/coerce.go
```diff
@@ -0,0 +1,268 @@
...
+// so the same buffers can be reused on each call without zcode.Bytes buffers
+// escaping to GC.  This method uses the zng.AppendInt(), zng.AppendUint(),
+// etc to encode zcode.Bytes as an in-place slice instead of allocating
+// new slice buffers for every value created.
+type Coercion struct {
+	// a and b point to inputs that can't change
+	a zcode.Bytes
+	b zcode.Bytes
+	// buf is a scratch buf that stays around between calls and the
+	// landng place for either the a or b value if one of them needs to
```
mattnibs, line 28:
```suggestion
	// landing place for either the a or b value if one of them needs to
```

expr/expr.go
```diff
@@ -24,209 +25,237 @@ var ErrNoSuchFunction = errors.New("no such function")
...
+// optimized code for the now-static types within that record type.
+//
+// The keep flag, if true, says to return zng.Value that are safe to stash
+// and will be garbage collected.  Otherwise, the expression will produce
+// values into temporary buffers may be modified on subsequent calls to Eval.
+// This is intended to minimize the garbage collection needs of the inner loop
+// by not allocating memory on a per-Eval basis.  For uses like filtering and
+// aggregations, where the results are immediately use, this is desirable and
+// "keep" should be false in these cases. For use cases like storing the results
+// as groupby keys, the results must bae nonvolatile and "keep" should be true.
```
mattnibs, line 51:
```suggestion
// as groupby keys, the results must be nonvolatile and "keep" should be true.
```

expr/scratch.go
```diff
@@ -0,0 +1,60 @@
...
+
+func (s *scratch) Int(v int64) zcode.Bytes {
+	if s.keep {
+		return zng.EncodeInt(v)
+	}
+	b := s.buf
+	if b != nil {
+		b = b[:0]
+	}
+	s.buf = zng.AppendInt(b, v)
```
mattnibs, line 22: You do not need all these nil checks in all these functions:
```suggestion
	s.buf = s.buf[:0]
	s.buf = zng.AppendInt(s.buf, v)
```

mattnibs, line 22: I don't believe you need these nil checks in all these functions:
```suggestion
	s.buf = zng.AppendInt(s.buf[:0], v)
```

mccanne: Cool!  I did not know `[]byte(nil)[:0]` was legit.

zql/zql.peg
```diff
@@ -482,7 +481,7 @@ ExpressionAssignment
     }
 
 FieldAssignment
-  = l:fieldRefDotOnly __ "=" __ r:fieldRefDotOnly {
+  = l:DotExprText __ "=" __ r:DotExpr {
```
henridf, line 484: (This made me realize that put doesn't support dots on the LHS which is strange... unrelated to this PR so no need to address that here).

henridf:
Would this be a good opportunity to rename `expr/temp.go`?

expr/expr.go
```diff
@@ -24,209 +25,237 @@ var ErrNoSuchFunction = errors.New("no such function")
...
-		nv, err := ne(rec)
-		if err != nil {
-			return zng.Value{}, err
-		}
-
-		return nv.ToZngValue()
-	}, nil
+// This is the "intepreted slow path" of the analytics engine.
+// It is completely adaptive to the dynamic type system of zng so there
+// are a lot of types checks and run-tme decisions made according to type.
```
alfred-landrum, line 34:
```suggestion
// This is the "interpreted slow path" of the analytics engine.
// It adapts to the dynamic type system of zng, so many
// checks and run-time decisions are made according to type.
```

expr/expr.go
```diff
@@ -24,209 +25,237 @@ var ErrNoSuchFunction = errors.New("no such function")
...
-				return zngnative.Value{zng.TypeBool, true}, nil
-			}
+		if i.vals.equal() {
+			return zng.True, nil
 		}
-	}, nil
+	}
 }
 
 func floatToInt64(f float64) (int64, bool) {
```
alfred-landrum, line 261: The `floatToInt64` and `floatToUint64` functions remaining in the file are unused & can be deleted.

expr/expr.go
```diff
@@ -24,209 +25,237 @@ var ErrNoSuchFunction = errors.New("no such function")
 var ErrNotContainer = errors.New("cannot apply in to a non-container")
 var ErrBadCast = errors.New("bad cast")
 
-type NativeEvaluator func(*zng.Record) (zngnative.Value, error)
-
-// CompileExpr tries to compile the given Expression into a function
-// that evalutes the expression against a provided Record.  Returns an
+// CompileExpr compiles the given Expression into an object
+// that evalutes the expression against a provided Record.  It returns an
```
alfred-landrum, line 29:
```suggestion
// CompileExpr compiles the given Expression into an object
// that evaluates the expression against a provided Record.  It returns an
```

zng/docs/spec.md
```diff
@@ -753,3 +761,56 @@ For each ZNG primitive type, the following table describes:
...
+and casts or conversion functions should be used.
+
+Also,
+* numeric constants should be int64 or float64 unless cast, which means
+comparisons with constants will generally be coerced to these types and results
+of mathematical operations with constants will be promoted as well;
+* times and durations may be added, resulting in a time;
+* times may be subtracted, resulting in a duration; and,
+* a "plus" operator applied to two strings, implies concatenation,
+but a "plus" applied to a string and is a type mismatch and casts
```
henridf, line 815:
```suggestion
but a "plus" applied to a string and another type is a type mismatch and casts
```

expr/expr.go
```diff
@@ -245,542 +274,371 @@ func floatToUint64(f float64) (uint64, bool) {
...
+			return c.result(result), nil
+		}
+		return zng.False, err
+	}
+	var result int
+	if !c.vals.equal() {
+		if zng.IsFloat(id) {
+			v1, _ := zng.DecodeFloat64(c.vals.a)
+			v2, _ := zng.DecodeFloat64(c.vals.b)
+			if v1 < v2 {
```
mattnibs, line 458: nit: Maybe keep this section as a `switch` statement:
```golang
		switch {
		case zng.IsFloat(id):
			v1, _ := zng.DecodeFloat64(c.vals.a)
			v2, _ := zng.DecodeFloat64(c.vals.b)
			if v1 < v2 {
				result = -1
			} else {
				result = 1
			}
		case zng.IsSigned(id):
			v1, _ := zng.DecodeInt(c.vals.a)
			v2, _ := zng.DecodeInt(c.vals.b)
			if v1 < v2 {
				result = -1
			} else {
				result = 1
			}
		case zng.IsNumber(id):
			v1, _ := zng.DecodeUint(c.vals.a)
			v2, _ := zng.DecodeUint(c.vals.b)
			if v1 < v2 {
				result = -1
			} else {
				result = 1
			}
...
```

expr/expr.go
```diff
@@ -24,209 +25,237 @@ var ErrNoSuchFunction = errors.New("no such function")
...
+//
+// Eventually, we will optimize this by adding a "fast path" that dynamically
+// generates byte codes (which an in turn be JIT assembled into machine code)
+// for each zng TypeRecord encountered.  Once you know the type record,
+// you can generate code using strong typing just as an OLAP system does
+// due to its schemas defined up-front in its relational tables.  Here,
+// each record type is like a schema and as we encounter them, we can compile
+// optimized code for the now-static types within that record type.
+//
+// The keep flag, if true, says to return zng.Value that are safe to stash
```
alfred-landrum, line 44: Looking at the uses in the PR, it looks like nearly all calls CompileExpr uses keep = false. What do you think about either of these different approaches:
opt 1: Make all calls to CompileExpr/Eval act like "keep = false". Change the Evaluator interface to use a method name like "SharedValEval" (or some other name that clarifies a shared result is being returned). Add a convenience copy utility that takes a value and returns a copy of it for the uses where we'd otherwise use 'keep = true'.
opt 2: Make zng.Value similar to zng.Record: it has a 'nonvolatile' flag, and `Keep` and `CopyBody` functions. Remove the keep parameter from CompileExpr, and have all Eval's return volatile zng.Value.

I like option 2 for the similarity with zng.Record handling, so devs have a common set of idioms to use.

expr/functions.go
```diff
@@ -58,407 +70,486 @@ var allFns = map[string]struct {
...
 		if err != nil {
-			return zngnative.Value{}, err
+			return zng.Value{}, err
 		}
-		l, err := v.ContainerLength()
-		if err != nil {
-			return zngnative.Value{}, err
-		}
-		return zngnative.Value{zng.TypeInt64, int64(l)}, nil
+		return zng.Value{zng.TypeInt64, args.Int(int64(len))}, nil
```
mattnibs, line 87: I see we're using the scratch for the generated `zcode.Bytes`, are we worried about GC overhead for all these `zng.Value` structs?

expr/functions.go
```diff
@@ -58,407 +70,486 @@ var allFns = map[string]struct {
...
+func stringFormatFloat(args *Args) (zng.Value, error) {
+	zv := args.vals[0]
+	if zv.Type.ID() != zng.IdFloat64 {
 		return err("string.floatToString", ErrBadArgument)
 	}
-	s := strconv.FormatFloat(args[0].Value.(float64), 'g', -1, 64)
-	return zngnative.Value{zng.TypeString, s}, nil
+	f, _ := zng.DecodeFloat64(zv.Bytes)
+	s := strconv.FormatFloat(f, 'g', -1, 64)
+	return zng.Value{zng.TypeString, zng.EncodeString(s)}, nil
```
mattnibs, line 339: Should scratch have a `String(str string) zcode.Bytes` function?

proc/cut/cutter.go
```diff
@@ -20,7 +20,7 @@ type cutBuilder struct {
 func (cb *cutBuilder) cut(in *zng.Record) *zng.Record {
 	cb.builder.Reset()
 	for _, resolver := range cb.resolvers {
-		val := resolver(in)
+		val, _ := resolver.Eval(in)
```
mattnibs, line 23: Should we at least add a panic on the error here?

zql/zql.peg
mccanne: Indeed.  We're getting closer though... we need a universal cut/column builder that can handle this.  None of these builds support reaching into a sub-field as an lvalue to modify an existing field inside a nested column though the column builder can create entirely new nested columns.

expr/expr.go
mccanne: This changes the meaning a bit but I'll take a stab at a rewrite of the confusing sentence...

expr/expr.go
mccanne: I think this is a relic zngnative and reflects a change that I made... that 1.5:int will work but previously failed with error.  I will take them out, but are we okay with this change?  It general, I like an approach where zql does things that work and are robust but when surprising things happen there should be an easy figure things out (e.g., such coercion events could work rather than fail and be debounced and sent to the warnings channel).
```
% cat t.tzng
#0:record[x:int32,f:float64]
0:[1;1.5;]
% oldzq -t "put z=f:int64" t.tzng
bad cast
#0:record[x:int32,f:float64]
0:[1;1.5;]
% zq -t "put z=f:int64" t.tzng
#0:record[x:int32,f:float64,z:int64]
0:[1;1.5;1;]
```

zng/docs/spec.md
```diff
@@ -753,3 +761,56 @@ For each ZNG primitive type, the following table describes:
 * [Protocol Buffers](https://developers.google.com/protocol-buffers)
 * [MessagePack](https://msgpack.org/index.html)
 * [gNMI](https://github.com/openconfig/reference/tree/master/rpc/gnmi)
+
+## Appendix B. Recommended Type Coercion rules
+
+> TBD: it might be better to put this in the ZQL docs
```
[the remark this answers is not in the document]
mccanne: I'll take it out and put it in an issue.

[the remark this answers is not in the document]
mccanne: See Issue #1408

zcode/counted.go
```diff
@@ -30,6 +30,16 @@ func EncodeCountedUvarint(dst []byte, u64 uint64) uint {
 	return n
 }
 
+func AppendCountedUvarint(dst []byte, u64 uint64) []byte {
+	var n uint
+	for u64 != 0 {
+		dst = append(dst, byte(u64))
+		u64 >>= 8
+		n++
```
mattnibs, line 38: `n` is doing anything here.

expr/fieldexpr.go
```diff
@@ -1,176 +1,101 @@
...
-	if nodes != nil {
-		resolvers = make([]FieldExprResolver, 0, len(nodes))
-		for _, exp := range nodes {
-			res, err := CompileFieldExpr(exp)
-			if err != nil {
-				return nil, err
-			}
-			resolvers = append(resolvers, res)
+func (f *FieldExpr) Eval(rec *zng.Record) (zng.Value, error) {
+	// XXX make test for 'put x=a[selector]' where selector is another field
```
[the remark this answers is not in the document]
mccanne: Thanks for flagging this.  This is relevant to the index operator not the dot operator and doesn't work yet.  I was planning to add support for this in the zng-beta PR since this is a change to the spec allowing any string to reference a field, e.g., `rec["@f_.oo"]`.  For this PR, I will move this comment to the index operator code path and it will show up in my zng-beta rebase.

zng/docs/spec.md
```diff
@@ -753,3 +761,56 @@ For each ZNG primitive type, the following table describes:
...
+* For float64 and any integer type, the integer is converted to float64
+and any loss of precision causes no error.
+* For integers of same signed-ness but different widths, the smaller width
+type is converted to the wider type.
+* For any signed and unsigned integers smaller than 64 bits, the unsigned value
+is converted to the corresponding signed type if possible, and otherwise,
+both are converted to the widest signed type that will allow conversion of
+the unsigned value unless the unsigned value cannot be converted,
+in which case an overflow error occurs.   e.g., uint8(255) and int8(-1), are
+converted to int16(255) and int16(-1), but uint64(2^32) and any signed value
```
mattnibs, line 801:
```suggestion
converted to int16(255) and int16(-1), but uint64(1<<64-1) and any signed value
```

zng/port.go
```diff
@@ -11,33 +10,15 @@ import (
...
-		return 0, errors.New("port encoding must be 2 bytes")
-
-	}
-	return uint32(zv[0])<<8 | uint32(zv[1]), nil
+	return Value{TypePort, EncodeUint(uint64(p))}
 }
 
 func (t *TypeOfPort) Parse(in []byte) (zcode.Bytes, error) {
-	i, err := byteconv.ParseUint32(in)
+	i, err := byteconv.ParseUint64(in)
```
mattnibs, line 17: Do we not need to check for i <= math.MaxUint16?

expr/expr.go
mccanne: Good idea.  More readable as it exposes the structure more clearly.  I changed these if-else's throughout.

expr/expr.go
mccanne: I'm averse to option 2 because it adds overhead to zng.Value which is passed around extensively in the inner loop.  How about option 3?  Change Eval to always return a potentially volatile expression but introduce `expr.Keep(zng.Value)` to make copies of the byte slices in a Value as package expr knows which thingy-s are volatile.

mattnibs (commented):
Finished with my review. Pretty nifty. A couple of small comments but otherwise looking good.

expr/expr.go
alfred-landrum: > I'm averse to option 2 because it adds overhead to zng.Value which is passed around extensively in the inner loop. How about option 3? Change Eval to always return a potentially volatile expression but introduce `expr.Keep(zng.Value)` to make copies of the byte slices in a Value as package expr knows which thingy-s are volatile. Basically option 1 but with a new function rather than a method on Evaluator.

+1 - that's what's I was trying to describe as option 1; that sounds good to me.

expr/functions.go
mccanne: No problems here as discussed.

proc/cut/cutter.go
mccanne: It was this way before my PR.  I will create a new issue for cut to send up errors.  There is another case that should go up.  It could also insert error values.  This is probably the best approach.

mccanne: Issue #1409

zng/docs/spec.md
mccanne: I moved this to #1408

henridf:
> > Would this be a good opportunity to rename expr/temp.go?

> I'm confused... expr/temp.go is gone as I've reworked things to remove the presumption that you can do strings.Split(field, ".") to get the path of keys as this doesn't work with arbitrary strings as key names (which zng now and must support to be a superset of JSON). There are methtods to turn ast dot expressions and expr.FieldExpr's into human-readable form with "." but the other direction doesn't work.

@mccanne I'm the one who's confused. once i got the gist of the changes on the PR i figured temp.go would be gone and saw it but on another branch. sorry for the noise.

henridf (approved):
No more comments from me - I like the regularization of the AST, and I look forward to experiencing the 🚤 effect.

zql/zql.peg
mccanne: Issue #1410

zng/docs/spec.md
mccanne: This is now in in #1408 but the intent here was to talk aboubt uint64(maxint32+1) not maxuint64.

zng/port.go
mccanne: That discussion goes in #1408.  My preference for some of this stuff (like this one) is to just let the overflow happen.

mccanne:
> Update: I hadn't noticed the test failures. :shipit: after they pass

The tests are all passing.  The red is from the zng format change against zq-sample-data because hte format has changed.  I proposed to check this in red and update the sample data for the zng-beta merge, which will entail further zng format changes.  Right @philrz ?

philrz:
@mccanne: There's a corresponding `eval` branch of the zq-sample-data repo that has ZNG data based on your `eval` zq ready to go. It's in https://github.com/brimsec/zq-sample-data/pull/16 ready to be merged whenever we want. I just now made a new zq branch off your zq `eval` branch and added one modification to have it point to that updated sample data branch, and you can see in draft PR #1411 that the "zng-output-check" automation now passes "green". So if you're ready to merge on the zq side, we can merge https://github.com/brimsec/zq-sample-data/pull/16 and re-run your Actions one last time to have the satisfaction of seeing it run totally green before we merge.

That said, I do see in both your most recent Actions run and the one in #1411 that's based on your branch that there's still some failures related to the docs examples. For instance, some of the expressions math is creating some crazy high numbers:

```
$ zq -f table 'duration > 100 | put total_bytes=orig_bytes+resp_bytes | cut orig_bytes,resp_bytes,total_bytes' conn.log.gz
ORIG_BYTES RESP_BYTES TOTAL_BYTES
32         0          32
32         0          8224
406        1720       139337760
32         31         270722277408
32         31         69539954827296
32         31         17802463487598624
32         0          2323645472701292576
32         0          2323645472701292576
1          63         2323645472701292576
...
```

mccanne:
> @mccanne: There's a corresponding `eval` branch of the zq-sample-data repo that has ZNG data based on your `eval` zq ready to go. It's in [brimsec/zq-sample-data#16]

Cool.  We can rinse and repeat for the zng-beta branch.  I'm putting the last touches on @alfred-landrum's suggestion about "keep" and will be ready to merge soon.

expr/functions.go
mccanne: Excellent point.  I think this might be non-trivial so I created #1412 to do the work in a subsequent PR.

expr/expr.go
mccanne: This was a really good idea.  It forced me to consider exactly which calls sites needed to make a copy of the zng.Value.  Strings and IPs will be allocated twice by the logic here, but rather than check if the copied type s one of these, we'll just copy everything (and fortunately these copies don't happen in the inner loop) and I made issue #1412 to fix the string/IP issue.

mccanne: Just pushed the change for this...

mccanne:
> That said, I do see in both your most recent Actions run and the one in #1411 that's based on your branch that there's still some failures related to the docs examples. For instance, some of the expressions math is creating some crazy high numbers:

That was a coercion bug for duration types.  Lost in translation from zngnative to eval.  I will add a system test for this since it was only the markdown tests that caught it.

PR #1385, opened 2020-09-29
Aggregate function & grouping docs

philrz:
Here's the long overdue ZQL docs on aggregate functions & related grouping concepts. The docs and examples are based on my own black box testing and attempts to interpret the grammar. There might be subtleties that I've not captured, so please speak up if you think something is missing.

I also recognize that some of this might be subject to change as @mccanne continues his overhaul of expressions in particular and ZQL in general. However, as I've had community members recently cite the existence of these parts of the language and ask for guidance on use, I figure it's better to capture it & iterate rather than continuing to kick the can down the road.


alfred-landrum (approved):
👍 Read through a couple of times, looks good!

zql/docs/aggregate-functions/README.md
```diff
@@ -1,11 +1,111 @@
...
+* [`sum`](#sum)
+
+**Note**: Per ZQL [search syntax](../search-syntax/README.md), many examples
+below use shorthand that leaves off the explicit leading `* |`, matching all
+events before invoking the first element in a pipeline.
+
+# General Usage
+
+All aggregate functions may be invoked with one or more
+[grouping](../grouping/README.md) options that define the batches of events on
```
henridf, line 28:
```suggestion
[grouping](../grouping/README.md) options that define the groups of events on
```

zql/docs/grouping/README.md
```diff
@@ -0,0 +1,181 @@
+# Grouping
+
+All [aggregate functions](../aggregate-functions/README.md) may be invoked with
+one or more _grouping_ options that define the batches of events on which they
+operate. If explicit grouping is not used, an aggregate function will operate
```
henridf, line 5:
```suggestion
one or more _grouping_ options that partition the input stream into batches that are aggregated separately. If explicit grouping is not used, an aggregate function will operate
```

zql/docs/grouping/README.md
```diff
@@ -0,0 +1,181 @@
...
+| `day`     | `days`, `d`                       |
+| `week`    | `weeks`, `wks`, `wk`, `w`         |
+
+#### Example #1:
+
+To see the total number of bytes originated across all connections during each
+minute:
+
+```zq-command
+zq -f table 'every minute sum(orig_bytes) | sort -r ts' conn.log.gz
```
henridf, line 38: I think it would be helpful to explain that `every X` is just syntactic sugar for `by Time.trunc(ts, X)`.

zql/docs/grouping/README.md
```diff
@@ -0,0 +1,181 @@
...
+bad_HTTP_request                            11777
+line_terminated_with_single_CR              11734
+unknown_HTTP_method                         140
+above_hole_data_without_any_acks            107
+...
+```
+
+#### Example #2:
+
+By specifying multiple comma-separated field names, batches are formed for each
```
henridf, line 97:
```suggestion
By specifying multiple comma-separated field names, one batch is formed for each
```

henridf (approved):

zql/docs/grouping/README.md
philrz: @henridf and I discussed this suggestion offline and I'm inclined to leave it out for now. Concepts like syntactic sugar (whether we go with that phrasing or try to come up with alternate words) feels a bit deeper than we're trying to go for the audience I have in mind for the current ZQL docs. I expect at some point we may have alternate/supplemental docs for users that come from a programming or data science background, and I suspect that's when it'll become helpful to expose things like this as well as other implementation details. I'll leave this comment "unresolved" in the event that users do happen to come across this closed PR, though.

PR #1386, opened 2020-09-29
backward compat for reading the alpha version of zng files

mccanne:
THIS IS A SHORT COMMIT.  DON'T LET THE LINE COUNT FOOL YOU.
THERE ARE TWO COMMITS.  YOU ONLY NEED TO REVIEW THE SECOND ONE.

The commit is only big because we're copying a snapshot of the current 
zng package and couple other things to retain backward compat.  We will remove 
this later when zng-alpha is fully deprecated.  Please review the second commit 
in the PR.   The first commit is the snapshot of the zng package and the code 
is completely unmodified except for some of the import paths to have them
refer to zq/alpha/{zng,zio} instead of zq/{zng,zio}.

===

This commit adds backward compatibility for reading alpha-version
zng files.  Right now, the tip of the tree is still the alpha version
but we will soon introduce changes so we are putting this backward
compat in now, in front of coming changes.

This commit introduces a copy of package zng and other bits we need
under the zq/alpha tree.  This can be removed down the road when we
fully deprecate the alpha zng format.

Since the tzng format hasn't changed except for "byte" to "uint8",
which isn't currently easy to get into a zng file without hand editing,


henridf:
This looks like it does the job, but it did make me wonder if we could use the go module system to import the old version we want instead of copying everything into `alpha`. It looks like it's possible to import multiple versions of a module (https://www.percona.com/blog/2020/03/09/using-different-versions-of-a-package-in-an-application-via-go-modules/) but I don't know if that extends to re-importing oneself under a different version... 🤷

ztests/suite/format/hello.azng
```diff
@@ -0,0 +1 @@
+�a	b	�a	bhelloworldgoodnight�
```
alfred-landrum, line 1: Can you mark this file as binary for git? It didn't seem to detect it automatically.

zio/azngio/reader.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+// Reader provides a way to read the old alpha format of zng files and return
+// them as an updated zng stream.  We do this by referring to the old alpha
+// code in zq/alpha, creating an old zng reader that writes to an old tzng writer,
+// and a new tzng reader that reads the old tzng using an io.Pipe.  Since tzng
+// hasn't changed (except for "byte" changing to "uint8" and additions like
+// end-of-stream) and since byte is not really used in practice (at least yet),
+// it serves as the neutral translation format from alpha zng to beta zng.
+type Reader struct {
```
alfred-landrum, line 24: 👍 very cool approach!

zio/azngio/reader.go
```diff
@@ -0,0 +1,90 @@
...
+			writer.Close()
+			w.CloseWithError(err)
+			return
+		}
+		if rec == nil {
+			if err := writer.Close(); err != nil {
+				w.CloseWithError(err)
+			} else {
+				w.Close()
+			}
```
alfred-landrum, line 80:
```suggestion
			err = writer.Close()
			w.CloseWithError(err)
```

henridf (approved):

mccanne:
> This looks like it does the job, but it did make me wonder if we could use the go module system to import the old version we want instead of copying everything into `alpha`. It looks like it's possible to import multiple versions of a module (https://www.percona.com/blog/2020/03/09/using-different-versions-of-a-package-in-an-application-via-go-modules/) but I don't know if that extends to re-importing oneself under a different version... 🤷

Sweet.  That totally worked.

ztests/suite/format/hello.azng
mccanne: Leaving this "as is" as we agreed...

PR #1390, opened 2020-09-29
use space migration to convert alpha zng files

### LINKED ISSUE #1376: brim/zqd handling for breaking zng changes
Some of the zng changes will be backwards incompatible, so the new zq code won't be able to handle older zng files. For the next release of Brim/zqd, we can bundle the v0.22.0 version of `zq` (before the zng changes) and have a space migration that  uses it to read the old zng data, pipe it out as tzng, and then have the tzng written with the new zng code.

alfred-landrum:
This uses the in-tree alpha zng reader to convert the all.zng file from alpha zng binary to post-alpha zng. In a separate branch, I combined this with the `eval` branch in #1375 to test. That uncovered an issue with the alpha zng reader, which @mccanne is working on. That also verified the replacer code changes for aborting the replacement operation and leaving the original file intact.

I'll file this as a draft, as this should go in after #1375 .

closes #1376

### DIFF
--- a/ppl/zqd/pcapstorage/pcapstorage.go
+++ b/ppl/zqd/pcapstorage/pcapstorage.go
@@ -175,7 +175,7 @@ func (s *Store) NewSearch(ctx context.Context, req api.PcapSearch) (*Search, err
 
 const metaFileV0 = "packets.idx.json"
 
-func MigrateV3(u iosrc.URI, pcapuri iosrc.URI) error {
+func MigrateV3(ctx context.Context, u iosrc.URI, pcapuri iosrc.URI) error {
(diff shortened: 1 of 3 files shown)


alfred-landrum:
Updated this to recent master, and is now on top of https://github.com/brimsec/zq/pull/1390 .

mccanne:
Looks good to me!

alfred-landrum:
Closing this in favor of https://github.com/brimsec/zq/pull/1574 and an additional PR that will do the migrations in the background, so that Brim users don't need to wait for all of their spaces to be migrated before zqd is ready.

PR #1392, opened 2020-09-29
fix yo-dawg problem with zng-alpha backward compat

mccanne:
This commit fixes the zng-alpha commit.  Importing an old version of zq 
inside of zq because the import paths of the old version brought in the current
code instead of the old code.

The fix is to copy the needed zng package and zio pieces into zq/alpha.

This PR has two commits: the first is the big copy to zq/alpha and the second
are the changes to point to zq/alpha instead of zq/0220.


alfred-landrum:
This is working for me: I ran into this working on #1390 . With these changes, I was able to take the alpha zng file from a current Brim release, and use the alpha zng reader to convert it to post-alpha zng. I compared the `tzng` for each - using the alpha zq code to generate its tzng, and the post-alpha zq code to generate its tzng - and they match.

alfred-landrum (approved):

Issue #1365, opened 2020-09-25
first-class zng types


mccanne:
@henridf suggested having type IDs available to zql (e.g., to do `count() by id(fld)`), but since they are arbitrary integers this would an odd semantics.   So, instead we could have first-class types in zng where typeof(fld) is a type value that images the same as a zng type string but is a bona fide value in the zq runtime.

This could be encoded as a type ID in the zng binary format.  

The zng spec needs to be updated along with the addition of the zng type type to package zng and handling for the type in the zq runtime.

henridf:
>  handling for the type in the zq runtime.

This might be a separate task, but along with this we should get a way to refer to the top-level record in ZQL. So that we can not only refer to a field `id(fld)` but to the entire record. Maybe just `id()` for that?

mccanne:
Yes, I like `id()` to give the type value of the top record.  Maybe we could call it `typeof()` instead of `id()`?

mccanne:
This needs to be a typedef.  We will add the typedef when we renumber them as part of the final-spec review, which breaks the zng-sensitive tests so best to do all the renumbering at the same time.

henridf:
Sounds good, but in `typeof(fld)` how do you distinguish between asking for the type of the literal `fld` vs the contents of field `fld`? 
(Of course, we've been through this at length in a previous project...)

mccanne:
Right now, `fld` isn't a literal.  It's always a field reference.  So `typeof(fld)` is the type of the field.  We might want to think through this some more.

philrz:
Verified in `zq` commit `032d9c5`.

There's now a `typeof()` function that returns TZNG-format data type details.

For example, we know how the `version` field in Zeek data appears with different data types in different Zeek log types. This would show the data types and counts for each:

```
$ zq -t 'count() by _path,typeof(version)' *
#0:record[_path:string,typeof:type,count:uint64]
0:[http;bstring;144034;]
0:[ssl;bstring;35493;]
0:[ntp;uint64;904;]
0:[snmp;bstring;65;]
0:[ssh;uint64;22;]
```

`.` can also be used to refer to the type definition for the entire record being processed at that point in the ZQL pipeline. Therefore to see the TZNG-format schemas for every Zeek log type in the sample data:

```
#0:record[_path:string,typeof:type,count:uint64]
[...]

Issue #1317, opened 2020-09-18
define and implement a zng table/map type


mccanne:
A table/map type would be like set but the contents are key value pairs where only keys need to be unique and the canonical order is based on the key order.

The type definition would be go-like and look like this:
```
map[<any>]<any>
```
Given the zcode model, this could be used as an associative array for *any* zng data type.

This wouldn't be useful for very large arbitrary key,value maps *within* a zng stream as in hadoop because it would need to fit comfortably in memory given the sorted-key contraint and zng.Record encoding.  That said, this type could be a very useful gadget for manipulating record values in various ways at small scale.

On the other hand, large-scale variants of this datatype could be useful in the runtime referred to by procs that comes from elsewhere, e.g., to be a first-class external object available to a zql query but populated externally, e.g., to join threat-intel datq to a zng stream.

e.g., something like this:

```
live-search | put intel=config.intel.map[id.orig_h] | filter intel.badguy=true | alert "${$id.orig_h}: ${intel.msg}"
```
where config.intel.map is a map of type `map[ip]record[badguy:bool,info:string]` and config refers to some external configuration for how this map appears in the runtime.  This data structure could also be a cache of database lookups by key, e.g., retrieving the intel data from an online service and caching each result in the runtime table a la a DNS lookup cache.  Of course, the database or map could be stored locally if performance is an issue (though the lookups could be easily parallelized).

mccanne:
A table/map type would be like set but the contents are key value pairs where only keys need to be unique and the canonical order is based on the key order.

The type definition would be go-like and look like this:
```
map[<any>]<any>
```
Given the zcode model, this could be used as an associative array for *any* zng data type.

This wouldn't be useful for very large arbitrary key,value steams as in hadoop because it would need to fit comfortably in memory given the sorted-key contraint and zng.Record encoding.  That said, this type could be a very useful gadget for manipulating record values in various ways as you could be a datatype in the runtime referred to by procs that comes from elsewhere, e.g., to be a first-class external object   available to a zql query but populated externally, e.g., to join threat-intel datq to a zng stream.

e.g., something like this:

```
search | put intel=join(id.orig_h,config.intel.map) | filter intel.badguy=true | alert "${$id.orig_h}: ${intel.msg}"
```
where config is a map of type `map[ip]record[badguy:bool,info:string]`

philrz:
Verified in `zq` commit `18045ab`.

[...]

Issue #1315, opened 2020-09-18
add bytes type


mccanne:
Add bytes type to zq. Allow index operator that works on arrays to apply to the bytes type and return a byte value of the result.

mccanne:
Add bytes type to zq.  Allow index operator that works on arrays to apply to the bytes type and return a byte value of the result.

philrz:
Verified in `zq` commit `18045ab`.

The `bytes` type can now be read/written using `zq`. Since it represents a sequence of bytes encoded as base64, there's also Base64 helper functions to help get these in & out of strings.

Example:

```
$ echo -n 'hello' | base64
aGVsbG8=

$ cat bytes.tzng 
#0:record[encoded_greeting:bytes]
0:[aGVsbG8=;]

$ zq -t 'put greeting_as_string=encoded_greeting:string | put b64_in_string=toBase64(greeting_as_string) | put b64_in_bytes=toBase64(greeting_as_string):bytes | put another_decoded=fromBase64(b64_in_string):string' bytes.tzng 
#0:record[encoded_greeting:bytes,greeting_as_string:string,b64_in_string:string,b64_in_bytes:bytes,another_decoded:string]
0:[aGVsbG8=;hello;aGVsbG8=;YUdWc2JHOD0=;hello;]
```

As for this part of the opening description:

>Allow index operator that works on arrays to apply to the bytes type and return a byte value of the result.

This has yet to be done and is tracked in #1421.

Thanks @mccanne!

Issue #1314, opened 2020-09-18
update zng type enum


mccanne:
Enums are currently represented by strings but they should be coded as ints (i.e., the positional index of the enum string in the typedef) so that end systems using zng with enum can recover the full typedef of the enum. This will also serve as a hint to optimization that an enum column has low cardinality.

mccanne:
Enums are currently represented by strings but they should be coded as ints (i.e., the positional index of the enum string in the typedef) so that end systems using zng with enum can recover the full typedef of the enum.  This will also serve as a hint to optimization that an enum column has low cardinality./

mccanne:
Need to add as an enum type when we re-number the typedefs.

philrz:
I've verified current functionality in `zq` commit `18045ab`, though there's more yet to come.

A summary of what's changed & current state:

1. In the ZNG spec `enum` was removed from the [Primitive Types](https://github.com/brimsec/zq/blob/master/zng/docs/spec.md#5-primitive-types) and instead there's now an [enum typedef](https://github.com/brimsec/zq/blob/master/zng/docs/spec.md#3115-enum-typedef).
2. At the moment (`zq` commit `18045ab`), references to `enum` field names in expressions resolve to their underlying value, such as shown in one of the tests @mccanne added in the linked PR:

```
$ zq -version
Version: v0.22.0-57-g18045ab

$ cat enum.tzng 
#0:record[e:enum[int32,foo:[1],bar:[2],baz:[4]]]
0:[0;]
0:[1;]
0:[2;]

$ zq -t "put s=e:string, v=e+1" enum.tzng 
#0:record[e:enum[int32,foo:[1],bar:[2],baz:[4]],s:string,v:int64]
0:[0;foo;2;]
0:[1;bar;3;]
0:[2;baz;5;]
```

Follow-on work that remains:
- Support for `enum` in ZST (#1469)
- Add more ZQL syntax for working with `enum` (#1498)

PR #1394, opened 2020-09-30
Zng beta

### LINKED ISSUE #1291: allow strings as zng field names

mccanne:
This commit updates the ZNG spec based on deployment experience,
further comparisons with other approaches, and team feedback.
We hope to make no more backward incompatible changes and have
the ZNG format finalized by spring 2021.

Various encoding changes were made to improve the ergonomcis
of the format and simplify the implementation.

New types haven been added: bytes, enum, map, type, and error.
Float32 is coming soon in a subsequent PR.

New functions are: typeOf(), isErr(), fromBase64(), and toBase64().

The index operator (i.e., []) now works on records and maps,
in addition to arrays, and record types can now handle arbitrary
strings as keys, e.g., rec[".@_FOO"].

The alpha-zng autodetector now works.

Closes #1291, #1314, #1315, #1317, #1365, #1220 

OLD DRAFT COMMENT

Here is a draft for commentary of the zng-beta branch as a draft merge off of the eval branch.  This will hopefully get merged shortly after the big eval PR is merged... 

Please provide any thoughts or feedback on the beta version of zng/docs/spec.md.  The intention is to have no more backward incompatible changes and hopefully also no more forward-incompatible changes but I'm less concerned about the latter.

Since putting up this PR, I updated the zngio reader/writert o conform with the new control codes.  Now that this is in, the alpha-zng autodetector should work.


zng/docs/spec.md
```diff
@@ -9,10 +13,13 @@
...
 > * only streams of `record` types (which may consist of any combination of
->   other implemented types) may currently be expressed in value messages.
+>   other implemented types) are supported by zq even though a stream of
+>   any types may currently be expressed in value messages.
 >
-> Also, we are contemplating reducing the number of [primitive types](#5-primitive-types), e.g.,
-> the number of variations in integer types.
+> TBD: should we add maps? It would be a complex type comprising
+> a list of key/values where the key type and value type is defined in the
+> map typedef.  The contrarian view is that maps are easily represented
```
henridf, line 21: Note that the contrarian view does not preclude having maps in ZQL at some point in the future. There could be a operator `map(list)` that turns a list of `record[key:keyType,value:valueType]`  into a map that can then be operated in by map-savvy procs and reducers in ZQL. Or for more flexibility, maybe `map()` takes the key/value field names, so you could say `map(list, key, value)`. Maps would have to be converted back into a list representation before exiting the flowgraph. 

There might also be a way to create maps from a stream of a records, for example `count() by _path | mapStream(_path, count)` would transform the stream of records into a histogram-like map of `string->int` which would be more suitable e.g. for statistical operations than separate records.

Anyway, spitballing here and not suggesting we need to do this... but it shows how the space of co-designing a data format and data language is novel broad. We're just scratching the outer surface here. 😎

zng/docs/spec.md
```diff
@@ -95,11 +109,15 @@ The stream of values is interleaved with control messages
...
 
 The ZNG type system comprises the standard set of primitive types like integers,
-floating point, strings, byte arrays, etc. as well as container types
-like records, arrays, and sets arranged from the primitive types.
+floating point, strings, byte arrays, etc. as well as complex types
+that are built from other types like records, arrays, and sets arranged
+from the primitive types.
+
+ZNG also includes first-class types, where a value can be of type "type".
+This is useful for introspection and grouping aggregations by type.
```
henridf, line 120:
```suggestion
This is useful for data exploration and grouping aggregations by type.
```

(maybe?)

zng/docs/spec.md
```diff
@@ -288,7 +319,26 @@ The `<ntypes>` and the type IDs are all encoded as `uvarint`.
...
+--------------------------------------------------------
+```
+`<type-id>` and `<nelem>` are encoded as `uvarint`.
+The names have the same UTF-8 format as a record and are encoded
+as tag-encoded primitive strings.  Each value is encoded as
+a tag-encoded value in accordance with the type indicated
+by ``<typd-id>`.
+
+Enums are typically integers but this representatiions, like the rust
+programming language, allows for arbitrarily-valued enumerations.
```
henridf, line 339:
```suggestion
Enums are typically integers but this representation, allows for arbitrarily-valued enumerations.
```

These enums are not like Rust enums

zng/docs/spec.md
```diff
@@ -350,122 +453,116 @@ previously defined type, the appropriate typedef control code must
...
 
-### 3.2 Value Messages
+Each value is prefixed with a "tag" that defines:
+* whether it is a primitive or complex value,
+* whether it is a nil value, and
+* if non-nil, its encoded length in bytes.
 
-Following a header byte with bit 7 zero is a `typed value`
-with a `uvarint7` encoding its length.
+The list collection of sub-values comprising a value of a complex type
```
henridf, line 486:
```suggestion
The collection of sub-values comprising a value of a complex type
```

zng/docs/spec.md
```diff
@@ -350,122 +453,116 @@ previously defined type, the appropriate typedef control code must
...
+is called a "container".
```
henridf, line 487: I'm not sure I get this definition... it sounds like it's saying that "container" is the contents ?

(Maybe "container" really refers to the encoding of a complex typed value?)

zng/docs/spec.md
```diff
@@ -514,28 +611,42 @@ This form defines an alias mapping the identifier to the indicated type.
...
-#!<control-code>:<payload>
+#!<control-code>:<message>
 ```
-Here, `<control-code>` is a decimal integer in the range 6-127 and `<payload>`
+Here, `<control-code>` is a decimal integer in the range 248-254 and `<message>`
 is any UTF-8 string with escaped newlines.
 
+### 4.1.4 End of Stream
+
+A TZNG end-of-stream market has the following form:
```
henridf, line 625:
```suggestion
A TZNG end-of-stream marker has the following form:
```

zng/docs/spec.md
```diff
@@ -514,28 +611,42 @@ This form defines an alias mapping the identifier to the indicated type.
...
+> TBD: implement TZNG EOS in zq
+
 ### 4.2 Type Grammar
 
 Given the above textual definitions and the underlying ZNG specification, a
 grammar describing the textual type encodings is:
 ```
-<stype> := bool | byte | int16 | uint16 | int32 | uint32 | int64 | uint64 | float64
-         | string | bytes | bstring | enum | ip | port | net | time | duration | null
+<stype> := bool | init8 | uinit8 | int16 | uint16 | int32 | uint32 | int64 | uint64 | float64
```
henridf, line 641:
```suggestion
<stype> := bool | int8 | uint8 | int16 | uint16 | int32 | uint32 | int64 | uint64 | float64
```

zng/docs/spec.md
```diff
@@ -741,15 +861,19 @@ For each ZNG primitive type, the following table describes:
...
-| `error`    | 21 | 8 or 32  | 8 or 32 bytes of IP prefix and subnet mask     | a string in CIDR notation representing an IP address and prefix length as defined in RFC 4632 and RFC 4291. |
-| `null`     | 22 |    0     | No value, always represents an undefined value | must be the literal value `-`                                 |
-
-> TBD: Types "enum" and "type" will actually be complex types not primitives.  We will
-> address this clarification in a subsequent PR.  There goes our magic constant 23.
+| `ip`       | 17 | 4 or 16  | 4 or 16 bytes of IP address                    | a string representing an IP address in [IPv4 or IPv6 format](https://tools.ietf.org/html/draft-main-ipaddr-text-rep-02#section-3) |
+| `net`      | 18 | 8 or 32  | 8 or 32 bytes of IP prefix and subnet mask     | a string in CIDR notation representing an IP address and prefix length as defined in RFC 4632 and RFC 4291. |
+| `error`    | 19 | variable | UTF-8 byte sequence of string of error message | a string in CIDR notation representing an IP address and prefix length as defined in RFC 4632 and RFC 4291. |
```
henridf, line 866:
```suggestion
| `error`    | 19 | variable | UTF-8 byte sequence of string of error message | a UTF-8 string. |
```

zng/docs/spec.md
```diff
@@ -741,15 +861,19 @@ For each ZNG primitive type, the following table describes:
...
+| `type`     | 20 | variable | unsigned int of length N representing type ID  | a string in CIDR notation representing an IP address and prefix length as defined in RFC 4632 and RFC 4291. |
```
henridf, line 867:
```suggestion
| `type`     | 20 | variable | UTF-8 byte sequence of string representing type in the [TZNG type grammar](#42-type-grammar)   | a UTF-8 string  |
```

(right?)

philrz (commented):
I found one link worth adding, but otherwise @henridf seems to have already spotted the majority of typos I would have spoken up about. 👍

zng/docs/spec.md
mccanne: This reads funny to me when I just swap the terms.  I landed here
```
This is useful for grouping aggregations by type to do introspective data exploration.
```

zng/docs/spec.md
mccanne: I don't find it confusing so it would be helpful to get more clarity here.  I changed it to this: `The collection of sub-values comprising a complex-type value is called a "container".`  The low-level encoding represents a container of things, which is why I like the term for the spec and the code (not for user facing docs).  As a point of contrast, avro has complex values but they are not serialized with containers.  You decode the values in depth-first order and you need the complete type definition to make sense of the encoding.  With zng, containers let you decode the values without knowing the type.

zng/docs/spec.md
mccanne: The market for end-of-streams is huge!

zng/docs/spec.md
mccanne: yes

zng/docs/spec.md
henridf: The rewording is clear to me, thanks.

zng/resolver/context.go
```diff
@@ -580,6 +631,104 @@ func (c *Context) parseArrayTypeBody(in string) (string, *zng.TypeArray, error)
 	return rest, c.LookupTypeArray(inner), nil
 }
 
+// parse an array body type of the form "[type]"
```
alfred-landrum, line 634:
```suggestion
// parse an enum body type of the form "[type,element1,element2,...]"
```

zng/enum.go
```diff
@@ -0,0 +1,61 @@
...
+	Type     Type
+	Elements []Element
+}
+
+type Element struct {
+	Name  string
+	Value zcode.Bytes
+}
+
+func NewTypeEnum(id int, typ Type, elements []Element) *TypeEnum {
```
alfred-landrum, line 21: Would you please add a ztest that uses an enum? I didn't see one.

expr/functions.go
```diff
@@ -67,6 +68,11 @@ var allFns = map[string]struct {
 	"Time.fromMicroseconds": {1, 1, timeFromUsec},
 	"Time.fromNanoseconds":  {1, 1, timeFromNsec},
 	"Time.trunc":            {2, 2, timeTrunc},
+
+	"typeof":     {1, 1, typeOf},
+	"iserr":      {1, 1, isErr},
```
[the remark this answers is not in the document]
mccanne: @philrz and I talked about function names and I think we need a group consensus around naming conventions.  The current name all are package-style but we don't have packages so that seems wrong.  I'm not sure about lower camelCase.  It seems a little arbitrary.  Also for a interactive query language where a user often type lots of stuff rapid fire, getting capitalization right seems like a mistake.  I guess I am arguing for snake case where underscores are used only for long names that are less often types.  How about I leave this for now and we do a PR soon on whatever we decide.   Issue #1426

zio/ndjsonio/reader.go
```diff
@@ -49,6 +49,10 @@ func NewReader(reader io.Reader, zctx *resolver.Context, opts ReaderOpts, filepa
 	if err != nil {
 		return nil, err
 	}
+	_, err = zctx.LookupTypeAlias("port", zng.TypeUint16)
```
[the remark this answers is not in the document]
mccanne: There don't seem to be any tests that exercise this.  If I comment them out, all tests pass.  But I'm pretty sure the typings stuff needs this and you are totally correct that the typings schema policy should define what zenum and port are rather than being hardcoded here.  I will add the comment you suggest.  I can't find an issue where to hang this so I will create a new one but it will probably be redundant with the larger task of imrpoving schema maps.  #1427

zio/tzngio/writer.go
```diff
@@ -56,6 +56,7 @@ func (w *Writer) Write(r *zng.Record) error {
 	if err != nil {
 		return nil
 	}
+	// XXX why can't we call StringOf on r.Type and make this easier?
```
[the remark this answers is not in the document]
mccanne: Yes, #1417 though I needed to clarify it.  I will put a comment in the code here that refers to #1417.  This is a quick PR after this is merged to clean this up and do proper error checks.

zio/tzngio/writer.go
```diff
@@ -115,12 +116,17 @@ func (w *Writer) writeContainer(parent zng.Value) error {
 	if _, ok := realType.(*zng.TypeUnion); ok {
 		return w.writeUnion(parent)
 	}
+	if typ, ok := realType.(*zng.TypeMap); ok {
+		s := typ.StringOf(parent.Bytes, zng.OutFormatZNG, true)
+		//XXX catch error
```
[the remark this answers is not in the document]
mccanne: I fixed the comment to refer to #1417

expr/expr_test.go
```diff
@@ -389,8 +391,8 @@ func TestCompareNonNumbers(t *testing.T) {
 	// port
 	testSuccessful(t, "p = 443", record, zbool(true))
 	testSuccessful(t, "p != 443", record, zbool(false))
-	testSuccessful(t, "p = :443", record, zbool(true))
-	testSuccessful(t, "p != :443", record, zbool(false))
+	//testSuccessful(t, "p = :443", record, zbool(true))
```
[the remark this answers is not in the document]
mccanne: I'll remove the comment.  Issue #1428

zng/enum.go
mccanne: Oops, good catch.  I found the enum test I thought I had committed and it reminded me I had another loose end to take care of... all done.

mccanne: BTW, there's some hokiness to the enum values in the type syntax...
```
#0:record[e:enum[int32,foo:[1],bar:[2],baz:[4]]]
```
This is the first time value syntax appeared in the type syntax.
I wrapped the values here in [] since this is escaped by the tzng value syntax and commas aren't. It's a bit ugly.  I had one thought here: we could change the tzng value encoding to use comma separation instead of semi-colon termination (and otherwise keep brackets) so that the type syntax and value syntax is consistent (@philrz suggested this a year ago).  Now would be the time to change this since it will break tzng format (it also messes up the azng updater).  The good news is only two small places in the code would need to change (but many tests would need to be updated).

zng/docs/spec.md
```diff
@@ -350,121 +470,126 @@ previously defined type, the appropriate typedef control code must
...
-> arbitrary binary data or a valid UTF-8 string.
+#### 3.2.2 Tag-Encoded Body of Complex Values
 
-A union value is encoded as a container with two elements. The first
-element is the `uvarint` encoding of the index determining the type of
-the value in reference to the union type, and the second element is
-the value encoded according to that type.
+The body of a length-N container comprises zero or more tag-encoded values.
 
 Array, set, and record types are variable length and are encoded
```
henridf, line 554:
```suggestion
Array, set, and record values are variable length and are encoded
```

zng/docs/spec.md
```diff
@@ -350,121 +470,126 @@ previously defined type, the appropriate typedef control code must
...
-as a sequence of elements:
+as a sequence of elements and an enum as a small integer:
```
henridf, line 555:
```suggestion
as a sequence of elements and enum values as a small integer:
```

The above sentence doesn't mention `map` or `union`.  Rather than adding them, maybe drop the sentence as it's basically paraphrasing the table?

zng/docs/spec.md
```diff
@@ -350,121 +470,126 @@ previously defined type, the appropriate typedef control code must
...
+A map value is encoded as a container as a sequence of alternating tag-encoded
+key and value. The concatenation of elements must be normalized so that the
+sequence of bytes encoding each tag-counted key (of the key/value pair) is
+lexicographically greater than that of the preceding key (of the preceding
+key/value pair).
 
-For sets, the concatenation of elements must be normalized so that the
-sequence of bytes encoding each element's tag-counted value is
-lexicographically greater than that of the preceding element.
+> XXX it's not clear that we should require sorted keys and give up
```
henridf, line 590: I'm not understanding this sentence... is the deletion of the paragraph describing set normalization suggesting that we remove that requirement ? (I don't see how we can, hence part of my confusion here).

zng/docs/spec.md
```diff
@@ -219,11 +250,11 @@ A record typedef creates a new type ID equal to the next stream type ID
 with the following structure:
 ```
 ----------------------------------------------------------
-|0x80|<nfields>|<field1><type-id-1><field2><type-id-2>...|
+|0xf0|<nfields>|<field1><type-id-1><field2><type-id-2>...|
```
henridf, line 253: should `nfields` be `ncolumns`? and `field1` be `fieldname` ?

zng/docs/spec.md
```diff
@@ -288,7 +317,35 @@ The `<ntypes>` and the type IDs are all encoded as `uvarint`.
...
+An enum type is encoded as the type code of the enum values as a `uvarint`
+followed by a count of the number of elements in the enum, and in turn,
+followed by the names and values of each element.
+```
+--------------------------------------------------------
+|0xf4|<type-id>|<nelem>|<name1><val-1><name2><val-2>...|
+--------------------------------------------------------
+```
+`<type-id>` and `<nelem>` are encoded as `uvarint`.
+The names have the same UTF-8 format as a record and are encoded
```
henridf, line 331:
```suggestion
The names have the same UTF-8 format as record field names and are encoded
```

zng/docs/spec.md
```diff
@@ -288,7 +317,35 @@ The `<ntypes>` and the type IDs are all encoded as `uvarint`.
...
+as tag-encoded primitive strings.  Each value is encoded as
+a tag-encoded value in accordance with the type indicated
+by ``<typd-id>`.
```
henridf, line 334:
```suggestion
by ``<type-id>`.
```

zng/docs/spec.md
```diff
@@ -350,121 +470,126 @@ previously defined type, the appropriate typedef control code must
...
-|----------|--------------------------------------|
-| `array`  | concatenation of elements            |
-| `set`    | normalized concatenation of elements |
-| `record` | concatenation of elements            |
+| Type     |          Value                          |
+|----------|-----------------------------------------|
+| `array`  | concatenation of elements               |
+| `set`    | normalized concatenation of elements    |
+| `record` | concatenation of elements               |
+| `union`  | concatenation of selector and value     |
```
henridf, line 562:
```suggestion
| `union`  | concatenation of type index and value     |
```

zng/docs/spec.md
```diff
@@ -350,121 +470,126 @@ previously defined type, the appropriate typedef control code must
...
+| `enum`   | selector of id and value in enum type   |
```
henridf, line 563:
```suggestion
| `enum`   | index of name and value in enum type   |
```

zng/docs/spec.md
```diff
@@ -350,121 +470,126 @@ previously defined type, the appropriate typedef control code must
...
+whereby the inner loop need not consult and interpret the type ID of each element.
 
-### 3.2 Value Messages
+#### 3.2.1 Tag-Encoding of Values
 
-Following a header byte with bit 7 zero is a `typed value`
-with a `uvarint7` encoding its length.
+Each value is prefixed with a "tag" that defines:
+* whether it is a primitive or complex value,
+* whether it is a nil value, and
```
henridf, line 500: I think some discussion of what `nil` means would be helpful. Likewise for `null`.

zng/docs/spec.md
```diff
@@ -9,10 +13,13 @@
...
->   other implemented types) may currently be expressed in value messages.
+>   other implemented types) are supported by zq even though a stream of
+>   any types may currently be expressed in value messages.
 >
-> Also, we are contemplating reducing the number of [primitive types](#5-primitive-types), e.g.,
-> the number of variations in integer types.
+> TBD: should we add maps? It would be a complex type comprising
+> a list of key/values where the key type and value type is defined in the
+> map typedef.  The contrarian view is that maps are easily represented
+> as a sequence of record[key:keyType,value:valueType].
```
henridf, line 22:
```suggestion
```

zng/docs/spec.md
```diff
@@ -860,7 +864,7 @@ This scheme allows complex types to be embedded in other complex types, e.g., a
 #26:record[city:string,lat:LL,long:LL]
 26:[NYC;[N;40.7128;][W;74.0060;]]
 ```
-An unset value indicates a field of a `record` that wasn't set by the encoder:
+An null value indicates a field of a `record` that wasn't set by the encoder:
```
henridf, line 867:
```suggestion
A null value indicates a field of a `record` that wasn't set by the encoder:
```

zng/docs/spec.md
```diff
@@ -786,7 +790,7 @@ of a value:
 [ ]
 ```
 In addition, `-` must be escaped if representing the single ASCII byte equal
-to `-` as opposed to representing an unset value.
+to `-` as opposed to representing an null value.
```
henridf, line 793:
```suggestion
to `-` as opposed to representing a null value.
```

zng/ztests/enum-err.yaml
```diff
@@ -0,0 +1,9 @@
+zql: "put s=e:string"
+
+input: |
+  #0:record[e:enum[int32,foo:[1],bar:[2],baz:[4]]]
+  0:[100;]
```
henridf, line 5: shouldn't this just be invalid tzng/zng rather than an error that is discovered when you try to use it?

zng/ztests/enum-err.yaml
```diff
@@ -0,0 +1,9 @@
...
+
+output: |
+  #0:record[e:enum[int32,foo:[1],bar:[2],baz:[4]],s:error]
```
henridf, line 8: I know we've talked about this on another PR, but seeing this confirms my impression that it will be hard to work with. I guess we can let it play out and get some mileage on it (this doesn' tneed to hold up this PR as it's not zng).

henridf:
Should the error type also be an alias ?

zng/ztests/enum.yaml
```diff
@@ -0,0 +1,13 @@
+zql: "put s=e:string, v=e+1"
```
henridf, line 1: So it looks like for an enum-valued `e`, to get the name you have to coerce to string, whereas `e` itself is the typed value ? This would be quite different both from  c-like enums and from sum types. 
I'm thinking it might be best to have c-like enums with a name and an (implicit) integer index, where only the name is visible in ZQL.

zng/docs/spec.md
mccanne: This was moved to line 573.  The XXX refers to map values.  Maybe we can discuss at sync-up?

zng/docs/spec.md
mccanne: Ok, I did `ncolumns` and `name1`...

zng/docs/spec.md
mccanne: This reads like a "type index" is a thing.  I like the term selector but how about "position"?

zng/docs/spec.md
mccanne: I used `position` as aboove and added further clarification in the text.

henridf: 👍

zng/docs/spec.md
henridf: Ah hadn't spotted that the paragraph moved. This makes sense.

zng/docs/spec.md
mccanne: Ok, adding more on null value and type null.  Getting rid of "nil".  We did that bikeshed :-)

zng/ztests/enum-err.yaml
mccanne: We don't have a semantic checker yet in the zq system but yes, if the index is a constant (it doesn't have to be), then we should check for this at compile time.  e.g., this current works...
```
"head 1 | put x=e[e]" enum.tzng

Comment (henridf) [reply zng/ztests/enum-err.yaml:5]: Hmm, I was just saying that 
```
#0:record[e:enum[int32,foo:[1],bar:[2],baz:[4]]]
  0:[100;]
```
could be determined as invalid zng when reading the record, similar to how `Record.Typecheck()` catches some invalid encodings.  Maybe I'm misunderstanding something here?

File: zng/ztests/enum.yaml (hunk 31 above)
Comment (mccanne) [reply zng/ztests/enum.yaml:1]: C enums aren't necessarily a positional and can take initializers.  I was trying to capture the enum type as a place to interface with languages that have enums in them, which is I why I made them more general.  (We talked about rust enum previously, but I think that can be captured as a zng union). 

Here, a goal is to be able to put enums in and get enums out, for most style of enums out there.  If you google around for enum relating to spark, parquet, avro, mongo, OLAP, etc they do come up but perhaps the use cases are niche.

The avro definitioin of enum follows your proposal.  Parquet and mongo do not have enums.

I made cast work to get the name so you could do introspection, e.g., so you could do 
```
count() by e:string
```
and get something reasonable.  Normally the name of the enum is the useful thing and the value is useful for the code.  So if we just assume int, you'd still want to get at the name. 
I do realize e:string is a little odd if the value is also string.  Maybe we could define a function to give the name?  As I mentioned on today's call, I don't see this as something we would use in the near term but this fell in the scope of converging to a final zng spec.

Comment (mccanne) [reply zng/ztests/enum.yaml:1]: MySql enums follow your proposal as well.  SQL Server and Snowflake, as examples, don't seem to support enum.

File: expr/functions.go [hunk 32]
```diff
@@ -552,3 +558,41 @@ func timeTrunc(args *Args) (zng.Value, error) {
...
+	return zng.Value{zng.TypeType, zng.EncodeType(zv.Type.String())}, nil
+}
+
+func isErr(args *Args) (zng.Value, error) {
+	zv := args.vals[0]
+	result := zng.True
+	if zv.Type != zng.TypeError {
+		result = zng.False
+	}
+	return result, nil
```
nwt, line 573:
```suggestion
	if zv.Type == zng.TypeError {
		return zng.True, nil
	}
	return zng.False, nil
```

zio/zeekio/reader.go
```diff
@@ -25,7 +25,10 @@ func NewReader(reader io.Reader, zctx *resolver.Context) (*Reader, error) {
 	if err != nil {
 		return nil, err
 	}
-
+	_, err = zctx.LookupTypeAlias("port", zng.TypeUint16)
+	if err != nil {
```
nwt, line 29:
```suggestion
	if _, err = zctx.LookupTypeAlias("port", zng.TypeUint16); err != nil {
```

zng/ztests/bytes.yaml
```diff
@@ -0,0 +1,12 @@
+# make sure we can read packets with a buggy caplen greater than the
+# original packet size.  bad-caplen.pcap was extracted from the pcap
+# in brim/issue https://github.com/brimsec/brim/issues/615.
```
nwt, line 3: Delete me.

zng/ztests/enum-err.yaml
```diff
@@ -0,0 +1,9 @@
+zql: "put s=e:string"
```
nwt, line 1:
```suggestion
zql: put s=e:string
```
Most of the new ztests also quote the zql field's value unnecessarily.

zql/zql.peg
```diff
@@ -170,21 +169,16 @@ RegexpLiteral
...
     }
 
-PortLiteral
-  = v:port {
-      RETURN(MAP("op": "Literal", "type": "port", "value": v))
-    }
-
 SubnetLiteral
-  = v:ip6subnet {
+  =  v:ip6subnet !fieldNameRest {
```
nwt, line 173:
```suggestion
  = v:ip6subnet !fieldNameRest {
```

zql/zql.peg
```diff
@@ -170,21 +169,16 @@ RegexpLiteral
...
       RETURN(MAP("op": "Literal", "type": "net", "value": v))
     }
   / v:subnet {
       RETURN(MAP("op": "Literal", "type": "net", "value": v))
     }
 
 AddressLiteral
-  = v:ip6addr {
+  =  v:ip6addr !fieldNameRest {
```
nwt, line 181:
```suggestion
  = v:ip6addr !fieldNameRest {
```

zql/zql.peg
```diff
@@ -567,18 +560,19 @@ NotExpression
...
+      RETURN(MAP("op": "CastExpr", "expr": e, "type": ct))
   }
+  / CallExpression
 
-ZngType
- = ("uint8" / "int8" / "int16" / "uint16" / "int32" / "uint32"
- / "int64" / "uint64" / "float64" / "string" / "bstring"
- / "ip" / "net" / "time" / "duration") { RETURN(TEXT) }
+PrimitiveType
+ = (   "bytes" / "uint8" / "uint16" / "uint32" / "uint64"
```
nwt, line 569:
```suggestion
  = (  "bytes" / "uint8" / "uint16" / "uint32" / "uint64"
      / "int8" / "int16" / "int32" / "int64"
      / "duration" / "time"
      / "float64"
      / "bool" / "bytes" / "string" / "bstring"
      / "ip" / "net"
      / "type" / "error" / "null" ) { RETURN(TEXT) }
```
We indent by two spaces everywhere else.

zng/docs/spec.md
```diff
@@ -288,7 +329,35 @@ The `<ntypes>` and the type IDs are all encoded as `uvarint`.
...
+An enum type is encoded as the type code of the enum values as a `uvarint`
+followed by a count of the number of elements in the enum, and in turn,
+followed by the names and values of each element.
+```
+--------------------------------------------------------
+|0xf4|<type-id>|<nelem>|<name1><val-1><name2><val-2>...|
+--------------------------------------------------------
+```
+`<type-id>` and `<nelem>` are encoded as `uvarint`.
+The names have the same UTF-8 format a record field names and are encoded
```
philrz, line 343:
```suggestion
The names have the same UTF-8 format as record field names and are encoded
```

zng/docs/spec.md
```diff
@@ -288,7 +329,35 @@ The `<ntypes>` and the type IDs are all encoded as `uvarint`.
...
+as tag-encoded primitive strings.  Each value is encoded as
+a tag-encoded value in accordance with the type indicated
+by ``<type-id>`.
```
philrz, line 346:
```suggestion
by `<type-id>`.
```

zng/docs/spec.md
```diff
@@ -314,7 +387,66 @@ It is an error to define an alias that has the same name as a primitive type.
...
+
+> The reason control messages are not allowed in compressed blocks is to
+> allow for optimizations that discard entire buffers of data based on
+> heuristics to know a filtering predicate can't be true based on a
+> quick scan of the data (e.g., using the Boyer-Moore algorithm to determine
+> that a comparison with a string constant would not work for any record
+> value in the buffer).  Since blocks may be dropped without parsing using
+> such an optimization, any typedefs should be lifted out into the zng data
+> stream in front of the compressed blocks (i.e., the stream is rearranged
+// but its always safe to move typedefs earlier in the stream as long as
```
philrz, line 404:
```suggestion
// but it's always safe to move typedefs earlier in the stream as long as
```

Also, the `//` in there makes this paragraph look like it's still under construction. Not sure if it could be replaced with a simple comma or if this was some "XXX"-style placeholder to remind you to come back to it.

zng/docs/spec.md
```diff
@@ -514,39 +654,62 @@ This form defines an alias mapping the identifier to the indicated type.
...
+A TZNG end-of-stream marker has the following form:
+```
+#!255
+```
+A TZNG stream or file should always be terminated with end-of-stream.
+This clears all of the previous type tag bindings and aliases, allowing
+multiple TZNG files with overlapping tags to be concatenated without the
+tags colliding.
+
+> TBD: implement TZNG EOS in zq
```
philrz, line 677: Should this just be moved to the top of the doc with the other zq "not implemented" exceptions? Or is it here because you intend to implement it really soon?

philrz (commented):
I also see some broken links to anchors within the doc that need to be adjusted, since some sections got moved around. Unfortunately it won't let me do suggestions on areas not close to where you've already made changes. Perhaps I could do a quick links-only pass after this PR gets merged.

zng/ztests/enum.yaml
henridf: > So if we just assume int, you'd still want to get at the name.

I think this is the heart of the matter: if we just assume int, do you ever want anything other than the name? I'm assuming not, which is why i'm thinking `e` evaluates to the name (or label).


> We talked about rust enum previously, but I think that can be captured as a zng union

(I agree that zng unions are in spirit comparable to rust enums aka sum types, but they're not as expressive. I'm assuming that's ok...  if we want to have a data type system that cleanly maps onto an algebraic type system, that's a whole new can of worms).

zng/ztests/enum-err.yaml
mccanne: Sorry, you're right.  Will address.

zng/ztests/enum-err.yaml
mccanne: Agreed.  Let's get this in the commit history in this general way then we can change or delete.

zng/ztests/enum.yaml
mccanne: Agreed. Let's get this version of eunm in the commit history in this general way then we can change or delete.

mccanne:
> Should the error type also be an alias ?

Let's discuss this live.  I suggest committing error as is and changing it to alias perhaps soon.

zng/ztests/enum-err.yaml
mccanne: I changed this to a script test now that it looks for a stderr result per @henridf's feedback.

zng/docs/spec.md
mccanne: Yes, also am changing this to `#eos`.

zng/ztests/enum.yaml
henridf: 👍

henridf (approved):
🥳

mccanne:
Thanks, @henridf.  @nwt let me know he is looking over `spec.md` so I will wait for his always-excellent edits before hitting merge.

zng/docs/spec.md
```diff
@@ -95,14 +105,30 @@ The stream of values is interleaved with control messages
...
+a null value in a field column means an optional value was not present, though
+the notion of optionality is not explicitly defined by ZNG.
+
+Null values can be typed or untyped.  For example, a null value in a field of
+a record has the type of the field's type.  A null can be "untyped" by creating
+a null value of type null.  The only valid value of type null is null.
+
+### Data Model Examples
+
+Here are a few examples of the ZNG data model expressed iin the TZNG text format.
```
nwt, line 129:
```suggestion
Here are a few examples of the ZNG data model expressed in the TZNG format.
```

zng/docs/spec.md
```diff
@@ -298,23 +362,86 @@ By encoding the aliases in the format, there is no need to configure mapping
...
+>
+> Since ZNG streams typically consist of a very sparse
+> set of typedefs with very long runs of data, these constraints are not
+> a barrier to performance in practice.
+
+A compressed value message block is encoded as follows:
+```
+------------------------------------------------------------------------------
+|0xf6|<format><uncompressed-length>|<compressed-length>|<compressed-messages>|
+------------------------------------------------------------------------------
```
nwt, line 412:
```suggestion
-------------------------------------------------------------------------------
|0xf6|<format>|<uncompressed-length>|<compressed-length>|<compressed-messages>|
-------------------------------------------------------------------------------
```
I botched this back when I added it.

nwt (approved):
:shipit:

PR #1396, opened 2020-09-30
add initial zng unmarshal support

alfred-landrum:
Adds support for handling the same types as the marshaling code, with todo's for types that will come soon with the zng changes.


mccanne (approved):
Looks good.  I was thinking for duration that package nano should define nano.Duration to int64 then the marshaler could act accordingly...

PR #1050, opened 2020-07-30
Add ZNG compression

nwt:
* Add a new "compressed message block" control message to the ZNG
  specification with initial support for LZ4.

* Add support for the new control message in zio/zngio.Reader and
  .Writer under control of a new zio.WriterFlags field called ~~ZngCompress~~
  ZngLZ4BlockSize.

* Enable compression for all commands by setting ~~ZngCompress to true~~ ZngLZ4BlockSize to a
  default value of 16384.

* Add a new ~~-zngcompress~~ -znglz4blocksize flag to zq ~~to disable compression~~.

Closes #521.

~~I'm still fixing some failing tests.~~

PR #1397, opened 2020-09-30
archive.Import: implement buffered streaming

mattnibs:
Previously archive.Import would read the entirety of the provided
stream, spilling to disk if necessary, before reading the stream
back in the order desired.

The change takes advantage of new overlapping chunk functionality.
Streams are now read into a zng buffer that is then sorted and written
to the archive once the buffer has:
1. Reached maxsize.
3. The provided zng stream has ended.

Part of #1132

### DIFF
--- a/archive/archive_test.go
+++ b/archive/archive_test.go
@@ -114,9 +114,9 @@ func TestOpenOptions(t *testing.T) {
 		return nil
 	})
 	require.NoError(t, err)
-	// if logid1 == "" || logid2 == "" {
-	// t.Fatalf("expected data files not found")
-	// }
+	if logid1 == "" || logid2 == "" {
+		t.Fatalf("expected data files not found")
+	}
 
 	out := indexQuery(t, ark1, query, AddPath(DefaultAddPathField, false))
 	require.Equal(t, test.Trim(fmt.Sprintf(expFormat, logid1, logid2)), out)
--- a/archive/archive_test.go
+++ b/archive/archive_test.go
@@ -97,8 +97,6 @@ func TestOpenOptions(t *testing.T) {
 `
 	first1 := nano.Ts(1587518620062237300)
 	first2 := nano.Ts(1587513578068769300)
-	// first1 := nano.Ts(1587518620062237300)
-	// first2 := nano.Ts(1587509168067598390)
 	var logid1, logid2 LogID
 	err = filepath.Walk(datapath, func(p string, fi os.FileInfo, err error) error {
 		if err != nil {
--- a/archive/archive_test.go
+++ b/archive/archive_test.go
@@ -92,11 +92,13 @@ func TestOpenOptions(t *testing.T) {
 	expFormat := `
 #zfile=string
 #0:record[key:int64,count:uint64,_log:zfile,first:time,last:time]
-0:[336;1;%s;1587518620.0622373;1587513894.06692143;]
-0:[336;1;%s;1587509168.06759839;1587508830.06852324;]
+0:[336;1;%s;1587518620.0622373;1587513611.06391469;]
+0:[336;1;%s;1587513578.0687693;1587508830.06852324;]
 `
 	first1 := nano.Ts(1587518620062237300)
-	first2 := nano.Ts(1587509168067598390)
+	first2 := nano.Ts(1587513578068769300)
+	// first1 := nano.Ts(1587518620062237300)
+	// first2 := nano.Ts(1587509168067598390)
 	var logid1, logid2 LogID
 	err = filepath.Walk(datapath, func(p string, fi os.FileInfo, err error) error {
 		if err != nil {
@@ -128,7 +130,7 @@ func TestOpenOptions(t *testing.T) {
 	expFormat = `
 #zfile=string
 #0:record[key:int64,count:uint64,_log:zfile,first:time,last:time]
-0:[336;1;%s;1587518620.0622373;1587513894.06692143;]
+0:[336;1;%s;1587518620.0622373;1587513611.06391469;]
 `
 	out = indexQuery(t, ark2, query, AddPath(DefaultAddPathField, false))
 	require.Equal(t, test.Trim(fmt.Sprintf(expFormat, logid1)), out)
--- a/archive/archive_test.go
+++ b/archive/archive_test.go
@@ -92,11 +92,12 @@ func TestOpenOptions(t *testing.T) {
 	expFormat := `
 #zfile=string
 #0:record[key:int64,count:uint64,_log:zfile,first:time,last:time]
-0:[336;1;%s;1587518620.0622373;1587513611.06391469;]
-0:[336;1;%s;1587513578.0687693;1587508830.06852324;]
+0:[336;1;%s;1587517405.06665591;1587517149.06304407;]
+0:[336;1;%s;1587509168.06759839;1587508830.06852324;]
 `
-	first1 := nano.Ts(1587518620062237300)
-	first2 := nano.Ts(1587513578068769300)
+	first1 := nano.Ts(1587517405066655910)
+	first2 := nano.Ts(1587509168067598390)
+
 	var logid1, logid2 LogID
 	err = filepath.Walk(datapath, func(p string, fi os.FileInfo, err error) error {
 		if err != nil {
@@ -113,9 +114,9 @@ func TestOpenOptions(t *testing.T) {
 		return nil
 	})
 	require.NoError(t, err)
-	if logid1 == "" || logid2 == "" {
-		t.Fatalf("expected data files not found")
-	}
+	// if logid1 == "" || logid2 == "" {
+	// t.Fatalf("expected data files not found")
+	// }
 
 	out := indexQuery(t, ark1, query, AddPath(DefaultAddPathField, false))
 	require.Equal(t, test.Trim(fmt.Sprintf(expFormat, logid1, logid2)), out)
@@ -128,7 +129,7 @@ func TestOpenOptions(t *testing.T) {
 	expFormat = `
 #zfile=string
 #0:record[key:int64,count:uint64,_log:zfile,first:time,last:time]
-0:[336;1;%s;1587518620.0622373;1587513611.06391469;]
+0:[336;1;%s;1587517405.06665591;1587517149.06304407;]
 `
 	out = indexQuery(t, ark2, query, AddPath(DefaultAddPathField, false))
 	require.Equal(t, test.Trim(fmt.Sprintf(expFormat, logid1)), out)
--- a/archive/archive_test.go
+++ b/archive/archive_test.go
@@ -97,7 +97,6 @@ func TestOpenOptions(t *testing.T) {
 `
 	first1 := nano.Ts(1587517405066655910)
 	first2 := nano.Ts(1587509168067598390)
-
 	var logid1, logid2 LogID
 	err = filepath.Walk(datapath, func(p string, fi os.FileInfo, err error) error {
 		if err != nil {
--- a/archive/import.go
+++ b/archive/import.go
@@ -52,7 +52,7 @@ type importWriter struct {
 	ctx     context.Context
 	writers map[tsDir]*tsDirWriter
 
-	bufSize int64
+	memBuffered int64
 }
 
 func newImportWriter(ctx context.Context, ark *Archive) *importWriter {
@@ -77,7 +77,7 @@ func (iw *importWriter) Write(rec *zng.Record) error {
 	if err := dw.writeOne(rec); err != nil {
 		return err
 	}
-	for iw.bufSize > ImportBufSize {
+	for iw.memBuffered > ImportBufSize {
 		if err := iw.spillLargestBuffer(); err != nil {
 			return err
 		}
@@ -144,7 +144,7 @@ func newTsDirWriter(iw *importWriter, tsDir tsDir) (*tsDirWriter, error) {
 func (dw *tsDirWriter) addBufSize(delta int64) {
 	old := dw.bufSize()
 	dw.size += delta
-	dw.importWriter.bufSize += dw.bufSize() - old
+	dw.importWriter.memBuffered += dw.bufSize() - old
 }
 
 // bufSize returns the actually buffer size as a multiple of importLZ4BlockSize.
--- a/archive/import.go
+++ b/archive/import.go
@@ -142,9 +142,9 @@ func newTsDirWriter(iw *importWriter, midnight nano.Ts) (*tsDirWriter, error) {
 	return d, nil
 }
 
-func (w *tsDirWriter) addBufSize(add int64) {
+func (w *tsDirWriter) addBufSize(delta int64) {
 	old := w.bufSize()
-	w.size += add
+	w.size += delta
 	w.importWriter.bufSize += w.bufSize() - old
 }
 
--- a/archive/import.go
+++ b/archive/import.go
@@ -334,10 +334,11 @@ func (cw *chunkWriter) close(ctx context.Context) error {
 		tf.Close()
 		os.Remove(tf.Name())
 	}()
-	tfr := zngio.NewReader(tf, resolver.NewContext())
+	zctx := resolver.NewContext()
+	tfr := zngio.NewReader(tf, zctx)
 	sf := seekIndexFile{id: cw.dataFile.id, recordCount: cw.rcount, first: cw.firstTs, last: cw.lastTs}
 	idxURI := cw.uri.AppendPath(sf.name())
-	idxWriter, err := microindex.NewWriter(resolver.NewContext(), idxURI.String(), []string{"ts"}, framesize)
+	idxWriter, err := microindex.NewWriter(zctx, idxURI.String(), []string{"ts"}, framesize)
 	if err != nil {
 		return err
 	}
--- a/archive/import.go
+++ b/archive/import.go
@@ -2,23 +2,27 @@ package archive
 
 import (
 	"context"
-	"fmt"
 	"io/ioutil"
 	"os"
 
-	"github.com/brimsec/zq/driver"
+	"github.com/brimsec/zq/expr"
 	"github.com/brimsec/zq/microindex"
 	"github.com/brimsec/zq/pkg/bufwriter"
 	"github.com/brimsec/zq/pkg/fs"
 	"github.com/brimsec/zq/pkg/iosrc"
 	"github.com/brimsec/zq/pkg/nano"
+	"github.com/brimsec/zq/proc/sort"
+	"github.com/brimsec/zq/proc/spill"
 	"github.com/brimsec/zq/zbuf"
 	"github.com/brimsec/zq/zio/zngio"
 	"github.com/brimsec/zq/zng"
 	"github.com/brimsec/zq/zng/resolver"
-	"github.com/brimsec/zq/zqd/api"
-	"github.com/brimsec/zq/zql"
+	"go.uber.org/multierr"
 )
+
+// ImportBufSize specifies the max size of the records buffered during import
+// before they are flushed to disk.
+var ImportBufSize = int64(sort.MemMaxBytes)
 
 // The below are vars for unit testing.
 var (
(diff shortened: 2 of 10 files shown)


zbuf/bufsort.go
```diff
@@ -0,0 +1,61 @@
+package zbuf
+
+import (
+	"context"
+
+	"github.com/brimsec/zq/expr"
+)
+
+type BufferedSort struct {
```
henridf, line 9: A comment explaining what BufferedSort does would be helpful.

zbuf/bufsort.go
```diff
@@ -0,0 +1,61 @@
...
+
+func NewBufferedSort(reader Reader, compareFn expr.CompareFn, maxsize int64) *BufferedSort {
+	return &BufferedSort{
+		maxsize:   maxsize,
+		peeker:    NewPeeker(reader),
+		compareFn: compareFn,
+	}
+}
+
+// Batch reads the next Batch into memory from the provided reader returns the
```
henridf, line 23:
```suggestion
// Batch reads the next Batch into memory from the provided reader and returns the
```

zbuf/bufsort.go
```diff
@@ -0,0 +1,61 @@
...
+// next Batch when either the reader reaches EOS, the size of the batch reaches
+// maxsize, or the context errors out. If the provided context has an error of
+// context.DeadlineExceeded, Batch will return the accumlated batch with no
+// errors.
+func (b *BufferedSort) Batch(ctx context.Context) (Batch, error) {
+	var size int64
+	// XXX Just create a new array at this point. At some point we should use
+	// a memory pool to reduce allocations.
```
henridf, line 31:
```suggestion
	// At some point we should use a memory pool to reduce allocations.
```

zbuf/bufsort.go
```diff
@@ -0,0 +1,60 @@
+package zbuf
+
+import (
+	"context"
+
+	"github.com/brimsec/zq/expr"
+)
+
+type BufferedSort struct {
```
henridf, line 9:
```suggestion
// BufferedSort reads from its reader, and accumulates records into a batch until either it reaches a maximum batch size, a timeout expires, or the stream ends. At that point, the batch is sorted and returned.
type BufferedSort struct {
```

Ok I realized that `Batch()` has the documentation  and since this a single-method struct, that's probably ok.

henridf (approved):
This looks good to me. I think part of my comment about adding a BufferedSort comment was that I didn't quite understand the name. I'm thinking BatchSort (or maybe BoundedSort) would be more descriptive, what do you think?

I also wonder if this belongs in zbuf but the only alternative I see is in archive.

alfred-landrum (changes requested):
I like the ideas this PR introduces, but think that more work is needed:
- I think its a good idea for import operations to land data in the archive sooner, so trying to land  data before reaching the archive threshold size (`-s` option) is ok, but then what is the threshold size for?
- If we want to enforce a time limit on keeping data buffered before writing, I think it'd be wise to have that explicitly enforced with a mechanism other than via context deadlines. As it is, if we actually wanted to enforce timeouts on connections to read data from, we'd need to add additional checking in archive.Import, since the bufferwriter code will assume DeadlineExceeded means the buffer timeout fired.

mattnibs:
> This looks good to me. I think part of my comment about adding a BufferedSort comment was that I didn't quite understand the name. I'm thinking BatchSort (or maybe BoundedSort) would be more descriptive, what do you think?

> I also wonder if this belongs in zbuf but the only alternative I see is in archive.

@henridf agree with the naming. I'm going to move it to be a private struct inside of package archive. This means the name is of less importance but I will cogitate on a better name.

mattnibs:
I discussed this pr with @alfred-landrum. I'm going to make the following changes:
- [ ] make bufferedSort a private struct in archive/import.go
- [ ] remove the context timeout check (will be addressed down the road)
- [ ] get rid of the peeker in bufferedSort and let the values exceed maxsize before flushing
- [ ] add an archive.Import option to specify maxsize and add the ability to toggle this in the zqd endpoint
- [ ] set maxsize to archive.threshold if maxsize is less than archive.threshold

mattnibs:
@alfred-landrum this is ready for another look.

archive/import.go
```diff
@@ -161,48 +161,101 @@ func (d *importDriver) close() error {
...
+}
+
+// Batch reads the next Batch into memory from the provided reader and returns the
+// next Batch when either the reader reaches EOS, the size of the batch reaches
+// maxsize, or the context errors out.
+func (b *boundedSort) Batch(ctx context.Context) (zbuf.Batch, error) {
+	var size int64
+	// XXX At some point we should use a memory pool to reduce allocations.
+	array := zbuf.Array{}
+	for {
```
alfred-landrum, line 237: If you move this loop into the existing loop of `Import`, you could:
- use an `var arr []*zng.Record` variable directly instead of a zbuf.Batch, and reset it (`arr = arr[:0]`) between writes
- you could loop over the array and call `writeOne` directly, or just adjust the writer to take an `[]*zng.Record` parameter.

archive/import.go
```diff
@@ -161,48 +161,101 @@ func (d *importDriver) close() error {
 	return idxWriter.Close()
 }
 
-func (d *importDriver) Write(cid int, batch zbuf.Batch) error {
-	if cid != 0 {
-		panic("importDriver write to non-zero channel")
-	}
+func (d *batchWriter) Write(batch zbuf.Batch) error {
```
alfred-landrum, line 164: This doesn't need to be exported now.

zqd/storage/storage.go
```diff
@@ -63,5 +63,5 @@ type Summary struct {
 type Storage interface {
 	NativeDirection() zbuf.Direction
 	Summary(ctx context.Context) (Summary, error)
-	Write(ctx context.Context, zctx *resolver.Context, zr zbuf.Reader) error
+	Write(ctx context.Context, zctx *resolver.Context, zr zbuf.Reader, maxBufSize int64) error
```
alfred-landrum, line 66: I would suggest leaving the interface as-is, so removing this change.

zqd/storage/archivestore/archivestore.go
```diff
@@ -73,8 +73,8 @@ func (s *Storage) Summary(ctx context.Context) (storage.Summary, error) {
 	return sum, err
 }
 
-func (s *Storage) Write(ctx context.Context, zctx *resolver.Context, zr zbuf.Reader) error {
-	return archive.Import(ctx, s.ark, zctx, zr)
+func (s *Storage) Write(ctx context.Context, zctx *resolver.Context, zr zbuf.Reader, maxBufSize int64) error {
+	return archive.Import(ctx, s.ark, zctx, zr, 0)
```
alfred-landrum, line 77: I would use the package variable here and/or allow a zqd setting to adjust the value used (either with a config variable, or using the same technique we use for the sort mem max).

```suggestion
func (s *Storage) Write(ctx context.Context, zctx *resolver.Context, zr zbuf.Reader) error {
		return archive.Import(ctx, s.ark, zctx, zr, archive.DefaultImportBufSize)
```

mattnibs: I get this, I didn't really like have to change the storage.Storage interface. That said my understanding from our conversation about this yesterday was that we wanted to have the ability to adjust maxBufSize on a per request basis.

> using the same technique we use for the sort mem max
The cli package does this by changing the package level var proc.MemMaxSize. We cannot really do this with zqd because of the long lived highly concurrent nature of a zqd instance.

> either with a config variable
Certainly doable if we decide to not make ImportBufSize not tunable at the request level.

If do want to make this configurable at the request level we could follow the `Options` pattern and change the storage.Storage.Write interface to `Write(context.Context, *resolver.Context, zbuf.Reader, ...storage.WriteOpts)`.

As an aside, I think we can get rid of the `*resolver.Context` as an argument in `Storage.Write`, no?

alfred-landrum: > I get this, I didn't really like have to change the storage.Storage interface. That said my understanding from our conversation about this yesterday was that we wanted to have the ability to adjust maxBufSize on a per request basis.

Ah, I'm sorry, I should have made the distinction between two operations:
* a cli user who runs `zar import` manually: we should be able to use a cli flag to control the memory used, since the user is directly interacting with the archive & is in control of the machine running the ingest operation.
* a log import request coming over the network: the remote requester may not & should not need to know the memory characteristics of whatever instance handles their request.

So its only the cli `zar import` case that I think should be able to specify the memory size.

> > using the same technique we use for the sort mem max
> 
> The cli package does this by changing the package level var proc.MemMaxSize. We cannot really do this with zqd because of the long lived highly concurrent nature of a zqd instance.
Agreed: if we only allow the change to happen via the cli for `zar import`, like we do with zq, we're only changing the variable during the execution of that command.

> As an aside, I think we can get rid of the `*resolver.Context` as an argument in `Storage.Write`, no?
I think so!

mattnibs: I'll add that I think you should be able to specify import memory size in the zqd.Config, but we can add this later.

archive/import.go
```diff
@@ -161,48 +161,60 @@ func (d *importDriver) close() error {
...
 	}
-	return "sort -r ts"
 }
 
 func Import(ctx context.Context, ark *Archive, zctx *resolver.Context, r zbuf.Reader) error {
-	proc, err := zql.ParseProc(importProc(ark))
-	if err != nil {
-		return err
+	maxBufSize := ImportBufSize
+	if maxBufSize < ark.LogSizeThreshold {
```
alfred-landrum, line 190: I think it should be reversed: ark.LogSizeThreshold could be much greater (gigabytes) than the desired memory consumption for import:
```suggestion
	if maxBufSize > ark.LogSizeThreshold {
```

alfred-landrum (approved):

archive/import.go
mattnibs: @alfred-landrum I'm realizing this is causing test to fail that set the `ark.LogSizeThreshold` small. I'm inclined to think we should just get rid of this check all together for the time being.

mattnibs:
@alfred-landrum et al. This is ready for another review. Per our conversation I reformatted the write process to utilize package `spill` on a per tsDir basis- this will allow a tsDir to keep records around until it is able to write a chunk file of sorted records as close to ark.LogSizeThreshold as we can get.

There's a couple of optimization that we'll probably want to add on subsequent prs:
- Add a MaxTempFsSize parameter that will keep archive.Import from exceeded the size of a provisioned tmpfs disk.
- Speed this up with concurrency. This is all currently written with no go routines. Seems like it would be prudent to run the chunkWriter functionality (and possibly also the spill operations) in separate goroutines to keep from holding up import while these expensive operations are carried out.

archive/import.go
```diff
@@ -2,156 +2,344 @@ package archive
...
+			d.records = d.records[:0]
+			d.addBufSize(d.size * -1)
+		}()
+	}
+	w, err := newChunkWriter(d.ctx, d.ark, d.tsDir)
+	if err != nil {
+		return err
+	}
+	if err := zbuf.CopyWithContext(d.ctx, w, r); err != nil {
+		w.close(d.ctx)
```
alfred-landrum, line 229: This `w.close(.)` call should be something like `w.abort()`, where chunkWriter.Abort just closes the dataFileWriter & the indexTempWriter, but doesn't try to create the seek file & push that up.

cmd/zar/import/command.go
```diff
@@ -42,6 +42,7 @@ type Command struct {
 	root       string
 	dataPath   string
 	thresh     units.Bytes
+	maxBufSize units.Bytes
```
alfred-landrum, line 45:
```suggestion
	importBufSize units.Bytes
```

archive/import.go
```diff
@@ -2,156 +2,344 @@ package archive
...
 	}
 	defer func() {
 		tf.Close()
 		os.Remove(tf.Name())
 	}()
-	tfr := zngio.NewReader(tf, d.zctx)
-	sf := seekIndexFile{id: d.dataFile.id, recordCount: d.rcount, first: d.firstTs, last: d.lastTs}
-	idxURI := d.tsDir.AppendPath(sf.name())
-	idxWriter, err := microindex.NewWriter(d.zctx, idxURI.String(), []string{"ts"}, framesize)
+	tfr := zngio.NewReader(tf, resolver.NewContext())
```
alfred-landrum, line 332:
```suggestion
        zctx := resolver.NewContext()
	tfr := zngio.NewReader(tf, resolver.NewContext())
```

archive/import.go
```diff
@@ -2,156 +2,344 @@ package archive
...
+	sf := seekIndexFile{id: c.dataFile.id, recordCount: c.rcount, first: c.firstTs, last: c.lastTs}
+	idxURI := c.tsDir.AppendPath(sf.name())
+	idxWriter, err := microindex.NewWriter(resolver.NewContext(), idxURI.String(), []string{"ts"}, framesize)
```
alfred-landrum, line 335:
```suggestion
	idxWriter, err := microindex.NewWriter(zctx, idxURI.String(), []string{"ts"}, framesize)
```

archive/import.go
```diff
@@ -2,156 +2,344 @@ package archive
...
+	}
+	if err := w.close(d.ctx); err != nil {
+		return err
 	}
+	return nil
+}
 
+// chunkWriter is a zbuf.Writer that writes a stream of sorted records into an
+// archive chunk file. chunkWriter is created and written to  by tsDirWriter
+// when it recieves the tsDirWriter.flush() call.
```
alfred-landrum, line 240:
```suggestion
// chunkWriter is a zbuf.Writer that writes a stream of sorted records into an
// archive chunk file.
```

archive/import.go
```diff
@@ -2,156 +2,344 @@ package archive
...
+type importWriter struct {
+	ark     *Archive
+	ctx     context.Context
+	writers map[nano.Ts]*tsDirWriter
 
-	// Create the data file writer
-	d.tsDir = d.ark.DataPath.AppendPath(dataDirname, newTsDir(d.firstTs).name())
-	if dirmkr, ok := d.ark.dataSrc.(iosrc.DirMaker); ok {
-		if err := dirmkr.MkdirAll(d.tsDir, 0755); err != nil {
+	bufSize int64
```
alfred-landrum, line 56: how about `memBuffered` for this?

archive/import.go
```diff
@@ -2,156 +2,344 @@ package archive
...
+		}
+		delete(i.writers, ts)
+	}
+	return merr
+}
+
+// tsDirWriter accumlates to records to be written to a particular tsDir-
+// currently segmented by day. When a tsdir receives enough records to exceed
+// ark.LogSizeThreshold, the underlying records are written to a chunk file in
+// the archive.
```
alfred-landrum, line 117:
```suggestion
// tsDirWriter accumulates records for one tsDir.
// When the expected size of writing the records is greater than
// ark.LogSizeThreshold, they are written to a chunk file in
// the archive.
```

archive/import.go
```diff
@@ -2,156 +2,344 @@ package archive
...
+		}
+	}
+	return d, nil
+}
+
+func (w *tsDirWriter) addBufSize(add int64) {
+	old := w.bufSize()
+	w.size += add
+	w.importWriter.bufSize += w.bufSize() - old
+}
```
alfred-landrum, line 149:
```suggestion
func (w *tsDirWriter) addBufSize(delta int64) {
	old := w.bufSize()
	w.size += delta
	w.importWriter.bufSize += w.bufSize() - old
}
```

archive/import.go
```diff
@@ -2,156 +2,344 @@ package archive
...
+
+func newImportWriter(ctx context.Context, ark *Archive) *importWriter {
+	return &importWriter{
+		ark:     ark,
+		ctx:     ctx,
+		writers: make(map[nano.Ts]*tsDirWriter),
+	}
+}
+
+func (w *importWriter) Write(rec *zng.Record) error {
```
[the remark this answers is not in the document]
mattnibs: Definitely

archive/archive_test.go
```diff
@@ -92,11 +92,13 @@ func TestOpenOptions(t *testing.T) {
...
-0:[336;1;%s;1587518620.0622373;1587513894.06692143;]
-0:[336;1;%s;1587509168.06759839;1587508830.06852324;]
+0:[336;1;%s;1587518620.0622373;1587513611.06391469;]
+0:[336;1;%s;1587513578.0687693;1587508830.06852324;]
 `
 	first1 := nano.Ts(1587518620062237300)
-	first2 := nano.Ts(1587509168067598390)
+	first2 := nano.Ts(1587513578068769300)
+	// first1 := nano.Ts(1587518620062237300)
+	// first2 := nano.Ts(1587509168067598390)
```
mattnibs, line 101:
```suggestion
```

archive/import.go
mattnibs: Good point.

mattnibs: Is/will the archive read path be resilient if there's a chunk with no accompanying index?

archive/import.go
mattnibs: Out of curiosity, why do they need the same context here?

archive/import.go
mattnibs: Sure

archive/import.go
```diff
@@ -2,156 +2,350 @@ package archive
...
+	var dw *tsDirWriter
+	for _, w := range iw.writers {
+		if dw == nil || w.size > dw.size {
+			dw = w
+		}
+	}
+	return dw.spill()
+}
+
+func (i *importWriter) close() error {
```
alfred-landrum, line 102:
```suggestion
func (iw *importWriter) close() error {
```

archive/import.go
```diff
@@ -2,156 +2,350 @@ package archive
...
+	if closeErr := w.close(); err == nil {
+		err = closeErr
+	}
+	return err
 }
 
-func (d *importDriver) newWriter(rec *zng.Record) error {
-	d.firstTs = rec.Ts()
-	d.rcount = 0
+// importWriter is a zbuf.Writer the partitions records by day into the
```
alfred-landrum, line 42:
```suggestion
// importWriter is a zbuf.Writer that partitions records by day into the
```

archive/import.go
```diff
@@ -2,156 +2,350 @@ package archive
...
+// appropriate tsDirWriter. importWriter keeps track of the overall memory
+// footprint of the collection of tsDirWriter and instructs the tsDirWriter
+// with the largest footprint to spill its records to a temporary file on disk.
+//
+// XXX importWriter does not currently keep track of size of records written
```
alfred-landrum, line 47: How about add a note saying there's no time based flushing yet, either?

archive/import.go
mattnibs: Whoops

archive/import.go
```diff
@@ -2,156 +2,350 @@ package archive
...
+func (dw *tsDirWriter) addBufSize(delta int64) {
+	old := dw.bufSize()
+	dw.size += delta
+	dw.importWriter.memBuffered += dw.bufSize() - old
+}
+
+// bufSize returns the actually buffer size as a multiple of importLZ4BlockSize.
+// This is done to ensure that whenever spill the generated spill is compressed
+// nicely onto disk.
+func (dw *tsDirWriter) bufSize() int64 {
```
alfred-landrum, line 153: I don't think taking compression into account here will make a significant difference.

archive/import.go
mattnibs: Should I file an issue for both of these?

archive/import.go
```diff
@@ -2,156 +2,350 @@ package archive
...
+	if dw.spiller == nil {
+		var err error
+		dw.spiller, err = spill.NewMergeSort(importCompareFn(dw.ark))
+		if err != nil {
 			return err
 		}
+		dw.spiller.SetWriterOpts(zngio.WriterOpts{
+			LZ4BlockSize:     importLZ4BlockSize,
+			StreamRecordsMax: importStreamRecordsMax,
+		})
```
nwt, line 191: Based on https://github.com/brimsec/zq/pull/1050#discussion_r463670238, we probably don't want to compress these files, and there's no benefit to giving them a nonzero StreamRecordsMax, so this and the zngio.WriterOpts additions in proc/spill can go away.

mattnibs: Sure, I'll remove this.

archive/import.go
mattnibs: Issues created.

archive/archive_test.go
```diff
@@ -92,11 +92,12 @@ func TestOpenOptions(t *testing.T) {
...
 #zfile=string
 #0:record[key:int64,count:uint64,_log:zfile,first:time,last:time]
-0:[336;1;%s;1587518620.0622373;1587513894.06692143;]
+0:[336;1;%s;1587517405.06665591;1587517149.06304407;]
 0:[336;1;%s;1587509168.06759839;1587508830.06852324;]
 `
-	first1 := nano.Ts(1587518620062237300)
+	first1 := nano.Ts(1587517405066655910)
 	first2 := nano.Ts(1587509168067598390)
+
```
mattnibs, line 100:
```suggestion
```

alfred-landrum (approved):
Thank you for filing & noting the issues for followup areas!

cmd/zar/import/command.go
```diff
@@ -53,6 +54,8 @@ func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
 	f.StringVar(&c.dataPath, "data", "", "location for storing data files (defaults to root)")
 	c.thresh = archive.DefaultLogSizeThreshold
 	f.Var(&c.thresh, "s", "target size of chunk files, as '10MB' or '4GiB', etc.")
+	c.importBufSize = units.Bytes(archive.ImportBufSize)
+	f.Var(&c.importBufSize, "bufsize", "maximum size of data read into memory before flushing to disk '99MB', '4GiB', etc.")
```
nwt, line 58:
```suggestion
	f.Var(&c.importBufSize, "bufsize", "maximum size of data read into memory before flushing to disk, as '99MB', '4GiB', etc.")
```
For symmetry with the c.thresh usage string.

PR #1389, opened 2020-09-29
Add support for timestamps with signed tz offsets

henridf:
Suricata outputs timestamp like `2015-04-13T11:32:45.143323-0200`,
whereas Zeek does `2015-04-13T11:32:45.143323Z02:00`. This commit lets
us parse both.

PR #1400, opened 2020-09-30
Convert eve.json upon ingest

### LINKED ISSUE #1213: Suricata eve.json format conversion
Design and implement the means by which suricata's json alerts from the eve.json output file will be converted into zng. The solution will likely involve something à la types.json, but if some other approach turns out to be simpler/faster, let's not exclude it off the bat.

henridf:
This PR adds a json type config that is suited for "alert" events output by the Brim suricata package. The config is used when transforming the Suricata-produced "eve.json" file. In addition, the suricata `timestamp` field is renamed to `ts`.

This PR is based on #1389 so if you come here first, please try to make the detour.

Closes #1213


suricata/types.json
```diff
@@ -0,0 +1,147 @@
+{
```
henridf, line 1: It feels a bit rich to have a top-level directory just for this file. I did it to follow what we did for Zeek, but I'd be game to move this file to the https://github.com/brimsec/build-suricata repo if people want it. Preferably in a follow on PR.

henridf: Actually (simpler to do) it could go in zqd/ingest alongside the go generated suricata.go.

pkg/jsontyper/generator.go
```diff
@@ -0,0 +1,49 @@
...
+	fileName := flag.Arg(0)
+	tc, err := inputflags.LoadJSONConfig(fileName)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "error parsing json config: %s\n", err)
+		return
+	}
+
+	contents := fmt.Sprintf(`package %s
+
+import "github.com/brimsec/zq/zio/ndjsonio"
```
mattnibs, line 35: Packages in `pkg` shouldn't have any concept of packages outside. Maybe call the package `typegenerator` and put it in `zqd/ingest` or as a sub-package of `zio/ndjsonio`?

suricata/types.json
mattnibs: My vote would be to put this in `zqd/ingest`. *Or* we could have a top level folder called `jsontypes` and put the type configs for suricata and zeek json.

mattnibs: *Or* we could do what we do with log ingest and pcap ingest a form post and allow the suricata type config to be a value in the form.

pkg/jsontyper/generator.go
henridf: good point, i'll put it in `zio/ndjsonio`

suricata/types.json
henridf: i can put it in `zqd/ingest`. I'd rather not change the pcap API for now since (unlike for log ingest) this isn't currently intended to be user configurable.

mattnibs (approved):

PR #1402, opened 2020-09-30
Remove fuse spill-to-disk limitation from ZQL docs

philrz:
Now that https://github.com/brimsec/zq/pull/1355 is merged, we can remove the caution in the ZQL doc about the limitation & error message.

While I'm at it, I also noticed some apparent copypasta in the help text from when `-fusemem` was created based on `-sortmem`, so I've also corrected that here.


mccanne (approved):
Maybe we should say somewhere that spilling to disk happens when the fusemax (and sortmax for sort) is exceeded.

philrz:
Thanks @mccanne. Yes, I'll cover the `-*mem` options in another pass.

PR #1416, opened 2020-10-03
add ast command

mccanne:
This commit adds a new command called "ast", which prints parsed zql as its 
underlying JSON object.  This is useful for test and dev and debugging zql queries.
As we add more sophisticated ast transformations as various optimizations are
implemented, this command could prove very useful for test, e.g., writing ztest-style 
tests for the various ast transformations.  It's also useful for debugging other parts
of the system to dig into how the ast is being transformed for distributed processing.
In this way, it's a bit like sql "EXPLAIN".

This command currently lives in zq/cmd, but we might consider packaging up test and dev 
tools in a different subdirectory that isn't installed by default via "make install".  Perhaps 
there would be a "make tools".


mattnibs:
This is the same thing as `zql/main/main.go`. Would be nice to merge these two, as this has a couple features that this lacks:
- Ability to run in repl mode
- Ability to compare javascript and golang compiler side-by-side.

I also agree with having a tools directory apart from cmd. Command line utils are starting to pile up.

mccanne:
> This is the same thing as `zql/main/main.go`. Would be nice to merge these two, as this has a couple features that this lacks:
> 
> * Ability to run in repl mode
> * Ability to compare javascript and golang compiler side-by-side.
> 
> I also agree with having a tools directory apart from cmd. Command line utils are starting to pile up.

Good idea.  I just pushed a commit along these lies. @mattnibs what do you think?

Instead of doing this during peg grammar dev:
```
cd zql
make run
```
you say 
```
make peg
```
in the top-level directory and you can run the ast command any time, from anywhere, against the currently compiled grammar.

mattnibs (approved):
Love it!

zql/README.md
```diff
@@ -31,17 +31,16 @@ pigeon and pegjs to create the two parsers.
...
 ## Testing
 
-[main](main) contains a simple test program that takes zql queries typed
-on the command line, parses them with both parsers, and prints out both
-results so you can compare for accuracy.  Type this to run the test
-program:
-
-`go run ./main`
+The [ast command](../cmd/ast) can be used for easiliy testing the output of
+the zql parser.
```
nwt, line 35:
```suggestion
the zql parser.
```

zql/main/main.go
```diff
@@ -1,80 +0,0 @@
-package main
```
nwt, line 1: Remove the run goal in zql/Makefile, too.

Makefile
```diff
@@ -151,6 +151,14 @@ build-python-lib:
...
 	@rm -rf python/build
 
+zql/zql.go: zql/zql.peg
+	(cd zql ; make)
+
+peg: zql/zql.go FORCE
+	go run ./cmd/ast -repl
+
+FORCE:
+
```
nwt, line 161:
```suggestion
peg:
	$(MAKE) -C zql
	go run ./cmd/ast -repl
```
`zql/zql.go: zql/zql.peg` and `peg: zql/zql.go` don't accurately reflect the dependencies.

Also, since this runs the ast command, I think ast would be a better name for this goal than peg.

mccanne: The move to `-C` is good but I don't understand your comments here.  Your change will cause pigeon and pegjs to run every time, even when zql/zql.peg has't been modified.  I believe my rules are correct.  They say that zq/zql.go depends on zql/zql.peg and the peg rule says that if I want to run the ast command on the current version of zql/zql.peg, then it needs to be updated if it's stale, but otherwise it can run.  The output from the rules is very helpful because if you're editing along and it is run everytime, but you happen to be editing the wrong zql.peg, then the feedback from make on triggering the build is very helpful.  And when you just want to type `!make` and it re-runs pigeon and pegjs when it doesn't have to, well I find that to be an unnecessary delay in my workflow.  It might be that I've completely missed your point though, so let me know what you think of my comments here...  p.s. I like `make peg` better than `make ast` because it's meant to be used in an edit-compile-debug cycle and you are building the peg grammar, not the ast command.

nwt: zql.go depends on zql.peg, parser-support.go, and the installed versions of goimports and pigeon.

I understand your point about unnecessary delay. At the same time, I'm wary of hiding a surprise for others inside a makefile that doesn't reflect accurately reflect dependencies. Let's see if we can come up with something fast and accurate.

nwt: Replacing `$(MAKE) -C zql` with `$(MAKE) -C zql zql.go` trims that step from 4.3 seconds to 3.2 seconds for me, but that's still slow.

mccanne: zql.go does not depend on parser-support.go in the build sense but you have a good point about the dependency on version info in zql/Makefile.  I just pushed another change that should address this.  What do you think now?

nwt: > zql.go does not depend on parser-support.go in the build sense

Right you are. I'd forgotten that peg.js includes only parser-support.js.

Makefile
```diff
@@ -151,6 +151,19 @@ build-python-lib:
 clean-python:
 	@rm -rf python/build
 
+PEG_GEN = zql/zql.go zql/zql.js zql/zql.es.js
+$(PEG_GEN): zql/zql.peg zql/Makefile
```
nwt, line 155:
```suggestion
$(PEG_GEN): zql/Makefile zql/parser-support.js zql/zql.peg
```

Makefile
```diff
@@ -151,6 +151,19 @@ build-python-lib:
...
+	$(MAKE) -C zql
+
+# This rule is best for edit-compile-debug cycle of peg development.  It should
+# properly trigger rebuilds of peg-generated code, but best to run "make" in the
+# zql subdirectory if you are changing versions of pigeon, pegjs, or javascript
+# dependencies.
+peg: $(PEG_GEN) FORCE
+	go run ./cmd/ast -repl
+
+FORCE:
```
nwt, line 165:
```suggestion
.PHONY: peg
peg: $(PEG_GEN)
	go run ./cmd/ast -repl
```

nwt (approved):

PR #1424, opened 2020-10-05
zqd worker process

marktwallace:
This adds a new API to zqd: /worker
This API allows the zqd to be invoked as worker process for a manager zqd process.

It adds a hidden flag, "chunk" to zapi, so the zqd /worker API can tested independently 
of a manager zqd. The test is in ztests/suite/zqd/worker-count.yaml.

The PR does NOT add features for a manager zqd process. I will split that into a separate PR to avoid having either one get too big.


marktwallace:
I have some questions about the code in multisource.spanWalk, and how to create a special case where it just returns scannerCloser for a single S3 object when the zqd instance is running a "/worker" request rather than a "/search" request.

cmd/zapi/cmd/get/get.go
```diff
@@ -81,6 +84,7 @@ func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
 	f.BoolVar(&c.debug, "debug", false, "dump raw HTTP response straight to output")
 	f.Var(&c.from, "from", "search from timestamp in RFC3339Nano format (e.g. 2006-01-02T15:04:05.999999999Z07:00)")
 	f.Var(&c.to, "to", "search to timestamp in RFC3339Nano format (e.g. 2006-01-02T15:04:05.999999999Z07:00)")
+	f.StringVar(&c.chunkInfo, "chunk", "", "comma separated list of ksuid,first_ts,last_ts,dataFileKind")
```
alfred-landrum, line 87:
```suggestion
	f.StringVar(&c.chunkInfo, "chunk", "", "dash separated list of ksuid,first_ts,last_ts,dataFileKind")
```

cmd/zapi/cmd/get/get.go
```diff
@@ -153,6 +173,55 @@ func parseExpr(spaceID api.SpaceID, expr string) (*api.SearchRequest, error) {
...
+	// This is only for testing usng the -chunk flag
+	search, err := zql.ParseProc(expr)
+	if err != nil {
+		return nil, err
+	}
+	proc, err := json.Marshal(search)
+	if err != nil {
+		return nil, err
+	}
+	chunkInfoArr := strings.Split(chunkInfo, "-")
```
alfred-landrum, line 187: FYI that I just filed https://github.com/brimsec/zq/issues/1448 : during that work, we can export the data file name parsing function from the archive package & use that single function here, instead of manually needing to parse out the individual chunk fields.

ztests/suite/zqd/worker-count.yaml
```diff
@@ -0,0 +1,73 @@
+script: |
+  source services.sh
+  mkdir spacedir
+  #
```
alfred-landrum, line 4: thank you for the comments explaining the flow here 👍

alfred-landrum (approved):

cmd/zapi/cmd/get/get.go
```diff
@@ -153,6 +173,55 @@ func parseExpr(spaceID api.SpaceID, expr string) (*api.SearchRequest, error) {
 	}, nil
 }
 
+// parseExprPlusChunk creates an api.WorkerRequest to be used with the client.
+func parseExprPlusChunk(spaceID api.SpaceID, expr string, chunkInfo string) (*api.WorkerRequest, error) {
```
nwt, line 177: s/Plus/With/ for a name that's in keeping with our existing convention.

cmd/zapi/cmd/get/get.go
```diff
@@ -153,6 +173,55 @@ func parseExpr(spaceID api.SpaceID, expr string) (*api.SearchRequest, error) {
...
+	// This is only for testing usng the -chunk flag
+	search, err := zql.ParseProc(expr)
+	if err != nil {
+		return nil, err
+	}
+	proc, err := json.Marshal(search)
+	if err != nil {
+		return nil, err
+	}
```
nwt, line 186:
```suggestion
	searchRequest, err := parseExpr(spaceID, expr)
	if err != nil {
		return nil, err
	}
```

k8s/dist_test.md
```diff
@@ -0,0 +1,39 @@
+# Testing procedures for distributed zqd
```
nwt, line 1: We use - instead of _ in all other markdown file names, so call this file dist-test.md.

zqd/api/api.go
```diff
@@ -45,6 +45,19 @@ type SearchRequest struct {
 	Dir   int             `json:"dir" validate:"required"`
 }
 
+type WorkerRequest struct {
+	SearchRequest
+	Chunks []ChunkEntry `json:"chunks"`
+}
+
+type ChunkEntry struct {
```
nwt, line 53: I think we should just call this Chunk.

zqd/api/api.go
```diff
@@ -45,6 +45,19 @@ type SearchRequest struct {
...
+	Id           string  `json:"ksuid" validate:"required"`
+	First        nano.Ts `json:"first" validate:"required"`
+	Last         nano.Ts `json:"last" validate:"required"`
+	DataFileKind string  `json:"filekind" validate:"required"`
+	RecordCount  int     `json:"recordcount" validate:"required"`
```
nwt, line 58:
```suggestion
	Id           string  `json:"id" validate:"required"`
	First        nano.Ts `json:"first" validate:"required"`
	Last         nano.Ts `json:"last" validate:"required"`
	FileKind string  `json:"data_file_kind" validate:"required"`
	RecordCount  int     `json:"record_count" validate:"required"`
```
For JSON field names we use the snake-case version of the Go field name.

zqd/search/worker.go
```diff
@@ -0,0 +1,93 @@
+// Package search provides an implementation for launching zq searches and performing
+// analytics on zng files stored in the server's root directory.
```
nwt, line 2:
```suggestion
```
search.go already has a package comment.

archive/staticsource.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"context"
+
+	"github.com/brimsec/zq/driver"
+	"github.com/brimsec/zq/zbuf"
+	"github.com/brimsec/zq/zng/resolver"
+)
+
+type staticSource struct {
```
mattnibs, line 11: Would appreciate a comment here describing what a static source is.

cmd/zapi/cmd/get/get.go
```diff
@@ -153,6 +173,55 @@ func parseExpr(spaceID api.SpaceID, expr string) (*api.SearchRequest, error) {
 	}, nil
 }
 
+// parseExprPlusChunk creates an api.WorkerRequest to be used with the client.
+func parseExprPlusChunk(spaceID api.SpaceID, expr string, chunkInfo string) (*api.WorkerRequest, error) {
+	// This is only for testing usng the -chunk flag
```
mattnibs, line 178:
```suggestion
	// This is only for testing using the -chunk flag
```

zqd/search/worker.go
```diff
@@ -0,0 +1,93 @@
...
+func NewWorkerOp(req api.WorkerRequest) (*WorkerOp, error) {
+	// XXX zqd only supports backwards searches, remove once this has been
+	// fixed.
+	if req.Dir == 1 {
+		return nil, zqe.E(zqe.Invalid, "forward searches not yet supported")
+	}
+	if req.Dir != -1 {
+		return nil, zqe.E(zqe.Invalid, "time direction must be 1 or -1")
+	}
+	//query, err := UnpackQuery(req)
```
mattnibs, line 38:
```suggestion
```

zqd/handlers.go
```diff
@@ -112,6 +112,47 @@ func handleSearch(c *Core, w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		respondError(c, w, httpReq, err)
+		return
+	}
+	defer cancel()
+
+	srch, err := search.NewWorkerOp(req)
+	if err != nil {
+		// XXX This always returns bad request but should return status codes
+		// that reflect the nature of the returned error.
```
mattnibs, line 137: I don't see any errors in `search.NewWorkerOp` that shouldn't be a 400 errors. Also, if there is an error with `ast.UnpackJSON` I believe the status would be 500 (the returned error is unwrapped).

mattnibs:
I'm kind of behind the times with the distributed query work, so I hope you'll forgive a naive question: In order to handle overlaps shouldn't the worker be given spans to operate on as opposed to chunks?

cmd/zapi/cmd/get/get.go
marktwallace: Great, I will update this in my next PR.

marktwallace:
> I'm kind of behind the times with the distributed query work, so I hope you'll forgive a naive question: In order to handle overlaps shouldn't the worker be given spans to operate on as opposed to chunks?

The /worker API for ZQD does accept a span which can include multiple chunks. The zapi -chunk flag takes one chunk with the dash-separated format from the file name, and builds a span that includes one chunk in order to call zqd /worker.

PR #1444, opened 2020-10-08
Use zcode.Bytes.Iter everywhere

nwt:
zb.Iter() is more concise than zcode.Iter(zb).

### DIFF
--- a/zng/value.go
+++ b/zng/value.go
@@ -36,7 +36,6 @@ func Parse(v ast.Literal) (Value, error) {
 	return Value{t, zv}, nil
 }
 
-//XXX b should be zcode.Bytes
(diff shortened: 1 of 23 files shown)


zng/value.go
```diff
@@ -37,11 +37,11 @@ func Parse(v ast.Literal) (Value, error) {
 }
 
 //XXX b should be zcode.Bytes
```
mccanne, line 39: update comment

mccanne (approved):
Good move!

zng/value.go
nwt: Done in 511b3f7a9126cda1b047449e7e8068aae111f3a8.

PR #1449, opened 2020-10-09
consolidate microindex Open into NewFinder

### LINKED ISSUE #1419: SEGV when attempting to query non-existent index
Repro is with `zar` commit `ccbfade`.

If I seed the archive as shown in the  [`zar` README](https://github.com/brimsec/zq/blob/master/cmd/zar/README.md), but then attempt to query the index before creating it, I get as SEGV rather than an error message.

```
$ zq zng/*.gz | zar import -s 25MB -
$ zar find :ip=10.10.23.2
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x16c65b6]

goroutine 55 [running]:
github.com/brimsec/zq/microindex.(*Reader).Path(...)
	/Users/phil/work/zq/microindex/reader.go:134
github.com/brimsec/zq/archive.search(0x1ac9b60, 0xc0000b43c0, 0xc00009a320, 0xc00010e720, 0xc0000364e0, 0x4, 0x0, 0x0, 0x0, 0xc0000364e7, ...)
	/Users/phil/work/zq/archive/finder.go:129 +0x1a6
github.com/brimsec/zq/archive.Find.func2.1(0xc00010e720, 0xab5c41b8751c080c, 0x33d7f41a86710e79, 0xcdeaa9ef, 0x151eebcd15773fb0, 0x151eeb5da84c7438, 0x188e245, 0x1, 0x6fa8a, 0xc00015a8c0, ...)
	/Users/phil/work/zq/archive/finder.go:102 +0x1f6
created by github.com/brimsec/zq/archive.Find.func2
	/Users/phil/work/zq/archive/finder.go:99 +0x1aa
```

alfred-landrum:
This also fixes a panic that would be raised when an error occurred
opening the microindex that left some Reader fields nil, and uses the
correct error check for the "-Q" option to zar find.

Fixes #1419


microindex/finder.go
```diff
@@ -23,22 +23,20 @@ type Finder struct {
 }
 
 // NewFinder returns an object that is used to lookup keys in a microindex.
-func NewFinder(zctx *resolver.Context, uri iosrc.URI) *Finder {
-	return &Finder{
-		zctx: zctx,
-		uri:  uri,
+// It opens the file and reads the trailer, returning errors if the file is
+// corrupt, doesn't exist, or if its microindex trailer is invalid.  If the
```
nwt, line 27:
```suggestion
// It opens the file and reads the trailer, returning errors if the file is
// corrupt, doesn't exist, or has an invalid microindex.  If the
```
For parallel structure.

ztests/suite/zar/index-missing.yaml
```diff
@@ -0,0 +1,21 @@
+script: |
+  mkdir logs
+  zar import -R ./logs babble.tzng
+  echo ===
+  zar find -R ./logs v=10 # no such index
+  echo ===
+  zar find -R ./logs -Q v=10 && echo "success" # no such index
```
nwt, line 7:
```suggestion
  zar import -R logs babble.tzng
  echo ===
  zar find -R logs v=10 # no such index
  echo ===
  zar find -R logs -Q v=10 && echo success # no such index
```

nwt (approved):

ztests/suite/zar/index-missing.yaml
alfred-landrum: I'm leaving the `./logs` in as it matches the other zar yaml tests, and can change all of them at once.

Issue #1448, opened 2020-10-09
move chunk metadata from seek to data file name


alfred-landrum:
With the recent changes in archive.Import, when we create a new data file, we'll know the first & last record timestamps, as well as the record count, for the data file. So we no longer need to encode that information in the seek file name, which will let us simplify the walk code that parses file names & identifies chunks.

PR #1450, opened 2020-10-09
zqd: Support pcap ingest for archive stores

### LINKED ISSUE #1132: zqd: Support archive space pcap ingest support

mattnibs:
Because of differences in how an archive ingests data implement
new ingest.PcapOp (ingest.archivePcapOp) for archive-backed spaces.

In order to take advantage of the streaming nature of ingest, abandon
the brute force snapshort approach taken with file stores. Instead,
temporary log files created by each pcap analyzer is are tailed and
written to the archive store as new data arrives (see fs.TFile and
Fs.DirWatcher).  As a result for pcap ingest on archive stores, new
data should constantly arriving.

In order to continue pcap ingest support for soon-to-be deprecated
filestore spaces, keep around existing ingest.PcapOp (renamed to
ingest.legacyPcapOp).

closes #1132

### DIFF
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/brimsec/zq/ast"
 	"github.com/brimsec/zq/zbuf"
+	"github.com/brimsec/zq/zng"
 	"github.com/brimsec/zq/zng/resolver"
 	"github.com/brimsec/zq/zqd/api"
 )
@@ -146,4 +147,49 @@ func (d *transformDriver) ChannelEnd(cid int) error           { return nil }
 func Copy(ctx context.Context, w zbuf.Writer, prog ast.Proc, zctx *resolver.Context, r zbuf.Reader, cfg Config) error {
 	d := &transformDriver{w: w}
 	return Run(ctx, d, prog, zctx, r, cfg)
+}
+
+type muxReader struct {
+	*muxOutput
+	batch       zbuf.Batch
+	index       int
+	statsTickCh <-chan time.Time
+}
+
+func (mr *muxReader) Read() (*zng.Record, error) {
+read:
+	if mr.batch == nil {
+		chunk := mr.Pull(mr.statsTickCh)
+		if chunk.ID != 0 {
+			return nil, errors.New("transform proc with multiple tails")
+		}
+		if chunk.Batch != nil {
+			mr.batch = chunk.Batch
+		} else if chunk.Err != nil {
+			return nil, chunk.Err
+		} else if chunk.Warning != "" {
+			goto read
+		} else {
+			return nil, nil
+		}
+	}
+	if mr.index >= mr.batch.Length() {
+		mr.batch.Unref()
+		mr.batch, mr.index = nil, 0
+		goto read
+	}
+	rec := mr.batch.Index(mr.index)
+	mr.index++
+	return rec, nil
+}
+
+func NewReader(ctx context.Context, program ast.Proc, zctx *resolver.Context, reader zbuf.Reader) (zbuf.Reader, error) {
+	mux, err := compileSingle(ctx, program, zctx, reader, Config{})
+	if err != nil {
+		return nil, err
+	}
+	return &muxReader{
+		muxOutput:   mux,
+		statsTickCh: make(chan time.Time),
+	}, nil
 }
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -152,6 +152,7 @@ func Copy(ctx context.Context, w zbuf.Writer, prog ast.Proc, zctx *resolver.Cont
 type muxReader struct {
 	*muxOutput
 	batch       zbuf.Batch
+	cancel      context.CancelFunc
 	index       int
 	statsTickCh <-chan time.Time
 }
@@ -183,12 +184,20 @@ read:
 	return rec, nil
 }
 
-func NewReader(ctx context.Context, program ast.Proc, zctx *resolver.Context, reader zbuf.Reader) (zbuf.Reader, error) {
+func (mr *muxReader) Close() error {
+	mr.cancel()
+	return nil
+}
+
+func NewReader(ctx context.Context, program ast.Proc, zctx *resolver.Context, reader zbuf.Reader) (zbuf.ReadCloser, error) {
+	ctx, cancel := context.WithCancel(ctx)
 	mux, err := compileSingle(ctx, program, zctx, reader, Config{})
 	if err != nil {
+		cancel()
 		return nil, err
 	}
 	return &muxReader{
+		cancel:      cancel,
 		muxOutput:   mux,
 		statsTickCh: make(chan time.Time),
 	}, nil
--- a/pkg/fs/tail.go
+++ b/pkg/fs/tail.go
@@ -6,7 +6,6 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
-	"sync"
 
 	"github.com/fsnotify/fsnotify"
 )
@@ -109,10 +108,10 @@ type FileEvent struct {
 // or removed. When open for the first time this will emit an event for
 // every existing file.
 type DirWatcher struct {
+	Events chan FileEvent
+
 	dir     string
-	events  chan FileEvent
-	once    sync.Once
-	err     error
+	watched map[string]struct{}
 	watcher *fsnotify.Watcher
 }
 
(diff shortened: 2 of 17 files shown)


zqd/ingest/legacypcap.go
```diff
@@ -0,0 +1,290 @@
+package ingest
+
```
alfred-landrum, line 2: @mattnibs : is this file just a copy of the pre-existing pcap ingest code? The diffs make it challenging to know what to review here & in the new archive specific pcap work.

mattnibs: For the most part. I'll keep the rename legacypcap.go -> pcap.go and give the current pcap.go another name, to make this pr more reviewable. After this merges I can file a followup pr to change the names back.

pkg/fs/tail.go
```diff
@@ -0,0 +1,175 @@
...
+	if !info.IsDir() {
+		return nil, errors.New("provided path must be a directory")
+	}
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return nil, err
+	}
+	return &DirWatcher{
+		dir:     dir,
+		events:  make(chan FileEvent, 5),
```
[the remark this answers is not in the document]
mattnibs: Good point.

pkg/fs/tail.go
```diff
@@ -0,0 +1,175 @@
...
+	return &TFile{f, watcher}, nil
+}
+
+func (t *TFile) Read(b []byte) (int, error) {
+read:
+	n, err := t.f.Read(b)
+	if errors.Is(err, os.ErrClosed) {
+		err = io.EOF
+	}
+	if n == 0 && err == io.EOF {
```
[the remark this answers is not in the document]
mattnibs: Good catch, thanks!

zqd/ingest/legacypcap.go
mattnibs: Done.

mattnibs:
I'm noticing a race with `fs.DirWatcher` that appears to show up in linux tests not mac: DirWatcher.Stop() can be called when fsnotify has not yet got the event for a created file. I'm going to add an additional poll call once Stop() is called to scan and emit any existing files that were not caught.

driver/driver.go
```diff
@@ -147,3 +148,48 @@ func Copy(ctx context.Context, w zbuf.Writer, prog ast.Proc, zctx *resolver.Cont
...
+		mr.batch.Unref()
+		mr.batch, mr.index = nil, 0
+		goto read
+	}
+	rec := mr.batch.Index(mr.index)
+	mr.index++
+	return rec, nil
+}
+
+func NewReader(ctx context.Context, program ast.Proc, zctx *resolver.Context, reader zbuf.Reader) (zbuf.Reader, error) {
```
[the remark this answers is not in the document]
mattnibs: Did not know this. Certainly, I'll make this a ReadCloser. In the next iteration of typing configs would be nice to have rename functionality for fields- then we can get rid of needing a flowgraph for ingest.

zqd/ingest/archivepcap.go
```diff
@@ -0,0 +1,247 @@
...
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	pcapfile, err := iosrc.NewReader(ctx, p.pcapuri)
+	if err != nil {
+		return err
+	}
+	defer pcapfile.Close()
+	// Keeps track of bytes read from pcapfile.
+	r := io.TeeReader(pcapfile, p)
```
alfred-landrum, line 97: Would you consider moving the io.Writer satisfying code off of the archivePcapOp struct, and to a separate struct that is used by archivePcapOp?

zqd/ingest/archivepcap.go
```diff
@@ -0,0 +1,247 @@
...
+	p.recordCounter.reader = combiner
+
+	if err := p.store.Write(ctx, p.zctx, p.recordCounter); err != nil {
+		return err
+	}
+	return group.Wait()
+}
+
+func (p *archivePcapOp) runAnalyzers(ctx context.Context, pcapstream io.Reader) (*errgroup.Group, []zbuf.Reader, error) {
+	group, ctx := errgroup.WithContext(ctx)
```
alfred-landrum, line 115: Consider moving the group creation up to `run()`, and passing it here, so that `run()` creates the cancellation context & the group that encompass the lifecycle of the operation.

zqd/ingest/archivepcap.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+	dr, err := newLogTailer(p.zctx, logdir, zio.ReaderOpts{
+		JSON: ndjsonio.ReaderOpts{TypeConfig: suricataTC},
+	})
+	if err != nil {
+		return nil, nil, err
+	}
+	group.Go(func() error {
+		err := waiter.Wait()
+		// Analyzer has either encountered an error or recieved an EOF from the
```
alfred-landrum, line 174:
```suggestion
		// Analyzer has either encountered an error or received an EOF from the
```

zqd/ingest/archivepcap.go
```diff
@@ -0,0 +1,247 @@
...
+
+func (r *recordCounter) Bytes() int64 {
+	return atomic.LoadInt64(&r.bytesRead)
+}
+
+func (r *recordCounter) Records() int64 {
+	return atomic.LoadInt64(&r.recordsRead)
+}
+
+// Snap for archivePcapOp is functionally useless. It is only here to satisfy
```
alfred-landrum, line 220: When we can run Brim against this branch, I'm curious how this impacts the live updating. I thought Brim triggered updating its view during pcap by looking for new snapshots, but I haven't looked at that code in while.

zqd/ingest/logtailer_test.go
```diff
@@ -0,0 +1,155 @@
...
+`
+
+type logTailerTSuite struct {
+	suite.Suite
+	dir  string
+	zctx *resolver.Context
+	dr   *logTailer
+}
+
+func TestLogTailer(t *testing.T) {
```
alfred-landrum, line 52: Just want to register that I'm +1 on using testify/suite.

alfred-landrum (commented):
Before this gets merged, I'd like to try this out with Brim, by temporarily making archive store the default.

zqd/ingest/archivepcap.go
mattnibs: Ugh, I always get this wrong. Thanks!

zqd/ingest/logtailer_test.go
mattnibs: It is definitely very useful for certain circumstances.

zqd/ingest/archivepcap.go
mattnibs: Yes, I was trying to be fancy but I think this just makes it confusing.

zqd/ingest/archivepcap.go
mattnibs: Sounds good.

zqd/ingest/archivepcap.go
mattnibs: As discussed offline, it is. Without an incrementing snapshot brim will not know new data is available to query. As discussed, this should really be keeping track of record count (and bytes) written to the archive, return it in the status updates, and Brim should use this to establish if new data is available.

pkg/fs/tail.go
```diff
@@ -0,0 +1,215 @@
...
+	// no data is added to the file, the next Read call will initiate the wait.
+	if n > 0 && err == io.EOF {
+		err = nil
+	}
+	if n == 0 && err == io.EOF {
+		if err := t.waitWrite(); err != nil {
+			return 0, err
+		}
+		goto read
+	}
```
nwt, line 61: Nits: Wrap the io.EOF handling in a single if statement. And save a trip through t.waitWrite if t.f.Read returns 0 and os.ErrClosed by handling io.EOF before os.ErrClosed.
```suggestion
	if err == io.EOF {
		if n > 0 {
			return n, nil
		}
		if err := t.waitWrite(); err != nil {
			return 0, err
		}
		goto read
	}
	if errors.Is(err, os.ErrClosed) {
		err = io.EOF
	}
```

zqd/ingest/logtailer.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/brimsec/zq/zng/resolver"
+)
+
+type result struct {
+	rec *zng.Record
+	err error
+}
+
+// logTailer is a zbuf.Reader that watches a specified directory and starts
+// tailing existing and new created files in the directory for new logs. Newly
```
nwt, line 20:
```suggestion
// tailing existing and newly created files in the directory for new logs. Newly
```

nwt (approved):

pkg/fs/tail.go
mattnibs: genius!

zqd/ingest/archivepcap.go
mattnibs: #1488

alfred-landrum (approved):
I pulled this branch, merged to master, altered zqd to choose archive store by default, and then ran Brim against it. I then uploaded the 'all.pcap' file & watched the archive tsdir's get created during import. (as noted, the live updates in Brim aren't working yet, but that will happen soon via #1488  ). I was able to verify that the resulting archive had the same count as the doing the process with the filestore. So 👍 from me.

mattnibs:
Thanks for the thorough review @alfred-landrum 

> (as noted, the live updates in Brim aren't working yet, but that will happen soon via #1448 ).

You mean #1488, correct?

alfred-landrum:
> Thanks for the thorough review @alfred-landrum
> 
> > (as noted, the live updates in Brim aren't working yet, but that will happen soon via #1448 ).
> 
> You mean #1488, correct?

oops, yes, fixed the comment!

PR #1452, opened 2020-10-09
separate compilation for driver.Run calls

alfred-landrum:
Instead of creating a fake multisource for calls to driver.Run, break
out the compilation for single reader vs multisource. This makes
upcoming changes to multisource easier to manage.

### DIFF
--- a/driver/compile.go
+++ b/driver/compile.go
@@ -31,21 +31,20 @@ type Config struct {
 	Warnings          chan string
 }
 
-func programPrep(program ast.Proc, sortKey string, sortReversed bool) (ast.Proc, ast.BooleanExpr, filter.Filter, error) {
+func programPrep(program ast.Proc, sortKey string, sortReversed bool) (ast.Proc, filter.Filter, ast.BooleanExpr, error) {
 	ReplaceGroupByProcDurationWithKey(program)
 	if sortKey != "" {
 		setGroupByProcInputSortDir(program, sortKey, zbufDirInt(sortReversed))
 	}
-	var filterExpr ast.BooleanExpr
 	var filt filter.Filter
-	filterExpr, program = liftFilter(program)
+	filterExpr, program := liftFilter(program)
 	if filterExpr != nil {
 		var err error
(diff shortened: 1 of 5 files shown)


driver/compile.go
```diff
@@ -27,7 +31,80 @@ type Config struct {
...
 
-func compile(ctx context.Context, program ast.Proc, zctx *resolver.Context, msrc MultiSource, mcfg MultiConfig) (*muxOutput, error) {
+func programPrep(program ast.Proc, sortKey string, sortReversed bool) (ast.Proc, ast.BooleanExpr, filter.Filter, error) {
+	ReplaceGroupByProcDurationWithKey(program)
+	if sortKey != "" {
+		setGroupByProcInputSortDir(program, sortKey, zbufDirInt(sortReversed))
+	}
+	var filterExpr ast.BooleanExpr
+	var filt filter.Filter
+	filterExpr, program = liftFilter(program)
```
nwt, line 41:
```suggestion
	var filt filter.Filter
	filterExpr, program := liftFilter(program)
```

driver/compile.go
```diff
@@ -27,7 +31,80 @@ type Config struct {
...
+	if cfg.Warnings == nil {
+		cfg.Warnings = make(chan string, 5)
+	}
+
+	program, filterExpr, filt, err := programPrep(program, cfg.ReaderSortKey, cfg.ReaderSortReverse)
+	if err != nil {
+		return nil, err
+	}
+
+	sn, err := scanner.NewScanner(ctx, r, filt, filterExpr, cfg.Span)
```
nwt, line 86:
```suggestion
	program, fil, filterExpr, err := programPrep(program, cfg.ReaderSortKey, cfg.ReaderSortReverse)
	if err != nil {
		return nil, err
	}

	sn, err := scanner.NewScanner(ctx, r, filt, filterExpr, cfg.Span)
```
Maybe return values in the same order they'll be passed to scanner.NewScanner?

nwt (approved):

Issue #1417, opened 2020-10-04
bikeshed topic zng.Type.StringOf() -> Format()


mccanne:
`<bikeshed>`

I find the `StringOf()` method name in the `zng.Type` interface confusing.  Ask someone what the difference between a `String()` and a `StringOf()` is?  I think `Format()` would be more aligned with the semantics.  After all, the method even takes an output format type as a parameter!

`</bikeshed>`

mccanne:
Also, it should return an error so errors can be bubbled up.  The whole tzng parsnig thing can be simplified by having each type do its own parse.

mccanne:
My latest thoughts are to replace `StringOf()` with `FormatZNG(strings.Builder) error` and `FormatCSV(strings.Builder) error` and get rid of `zng.OutFormat*` and not support the "debug" output format.

nwt:
@mccanne: FormatZNG and FormatCSV methods would definitely be an improvement over StringOf. But I think we should move all this formatting code out of methods in the zng package and into something like `WriteValue(w io.Writer, v zng.Value) error` in zio/tzngio and zio/zeekio. (strings.Builder implements io.Writer.)

(And then we should do something similar with the zng package's parsing code.)

philrz:
@mccanne reports that this got reworked and can be closed. #2473 is open to track one remaining piece of tech debt related to this, and #1566 also remains to track the wider topic of eventually morphing TZNG into an "assembler"-style, write-only representation of ZNG and hence retire TZNG as we knew it.

PR #1459, opened 2020-10-09
Remove zng.OutFormatDebug

nwt:
As suggested at https://github.com/brimsec/zq/issues/1417#issuecomment-704707245.


henridf (approved):

expr/fieldexpr.go
```diff
@@ -90,7 +90,10 @@ func FieldExprToString(e Evaluator) string {
 		rhs := FieldExprToString(e.field)
 		return lhs + rhs
 	case *Literal:
-		return e.zv.String()
+		if e.zv.Type != zng.TypeString {
+			panic("expr: unexpected type " + e.zv.Type.String())
```
mccanne, line 94: This changed in the dot-record merge and has an error return.  I don't think panic should be called here because then the caller needs to make sure no weird stuff crept through.  Better to get an error (which you will see when you rebase though your check isn;t there)

mccanne (approved):
This looks good.  I think we want to make one more move on this front.  See Issue #1417.  Meanwhile we should merge this.

expr/fieldexpr.go
nwt: The *Literal case is no longer needed, so it goes away in #1487.

PR #1461, opened 2020-10-09
move chunk metadata from seek to data file name

### LINKED ISSUE #1448: move chunk metadata from seek to data file name

alfred-landrum:
With the recent changes in archive.Import, when we create a new data file, we know the first & last record timestamps & record count. So we no longer need to encode that information in the seek file name, which simplifies the walk code that parses file names & identifies chunks.

Closes #1448

### DIFF
--- a/archive/import.go
+++ b/archive/import.go
@@ -18,6 +18,7 @@ import (
 	"github.com/brimsec/zq/zio/zngio"
 	"github.com/brimsec/zq/zng"
 	"github.com/brimsec/zq/zng/resolver"
+	"github.com/segmentio/ksuid"
 	"go.uber.org/multierr"
 )
 
@@ -122,13 +123,15 @@ func (iw *importWriter) close() error {
 // the archive.
 type tsDirWriter struct {
 	ark          *Archive
+	bufSize      int64
+	count        int64
 	ctx          context.Context
 	importWriter *importWriter
+	maxTs        nano.Ts
+	minTs        nano.Ts
 	records      []*zng.Record
-	bufSize      int64
 	spiller      *spill.MergeSort
 	tsDir        tsDir
-	uri          iosrc.URI
 }
 
 func newTsDirWriter(iw *importWriter, tsDir tsDir) (*tsDirWriter, error) {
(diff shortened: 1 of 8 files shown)


archive/walk.go
```diff
@@ -214,30 +120,72 @@ func tsDirEntriesToChunks(ark *Archive, filterSpan nano.Span, entries []iosrc.In
...
-}
-
 // Path returns the local filesystem path for the log file, using the
 // platforms file separator.
 func (l LogID) Path(ark *Archive) iosrc.URI {
 	return ark.DataPath.AppendPath(string(l))
 }
 
+// A Chunk is a file that holds records ordered according to the archive's
+// data order. The name of the file encodes the number of records it contains,
```
nwt, line 130:
```suggestion
// data order. The name of the file encodes the number of records it contains
```

nwt (approved):

archive/import.go
```diff
@@ -208,12 +220,19 @@ func (dw *tsDirWriter) flush() error {
 		}
 		expr.SortStable(dw.records, importCompareFn(dw.ark))
 		r = zbuf.Array(dw.records).NewReader()
-		defer func() {
-			dw.records = dw.records[:0]
-			dw.addBufSize(dw.bufSize * -1)
-		}()
 	}
-	w, err := newChunkWriter(dw.ctx, dw.ark, dw.uri)
+	first, last := dw.minTs, dw.maxTs
```
mattnibs, line 224: I know this is kind of late to be making this comment, but: Why not just make this min and max and not worry about having to switch this up based off of sorted direction?

alfred-landrum: You're right, we could do that, and then later in the walk code we could look at the archive's data sort direction to know how to construct the span for the chunk. I like doing it here, though, as it makes the file name reflect the actual order of the records inside.

PR #1468, opened 2020-10-10
fix field model and dot expression syntax

mccanne:
This commit fixes problems with the design patterns underlying
fields, dotted accesses, and the data structures in packages
ast and expr supporting this.  Dots can now be in field names
and the field representation is no longer a string, but rather
a slice of strings (as defined in the new field package in zq/field).

For example, you can now say .["@."] to refer to the field named "@."
in the root record refered to with '.'.  'foo' is shorthand for '.foo'
and 'foo["bar"]' is, the same as 'foo.bar', but 'foo[bar]' means to
look for a field in record '.foo' whose name is the value of '.bar'.

You can also say things like 'count() by ." or "count() by typeof(version)"

Reducer and assignment expressions within the ast have been simplified
and follow more conventional compiler design patterns.  Also,
field accesses now explicitly start from the root record and there
is now an ast.RootRecord that represents this.

This refactoring allowed several other improvements:
- You can sort by expressions, e.g., "sort String.toLower(s)"
- Put expressions are more flexible, e.g., "put typeof(.),String.toLower(s)"
- Cut expressions are more flexible, e.g., "cut x=y+1,typeof(s)"

Finally, the PEG grammar was simplified and made more orthogonal
so it will be easier to extend and modify moving forward.


proc/rename/rename.go
```diff
@@ -16,71 +17,77 @@ import (
...
+		// We call CompileLval on the RHS because renames are
+		// restricted to dotted field name expressions.
+		src, err := expr.CompileLval(fa.RHS)
+		if err != nil {
+			return nil, err
+		}
+		if len(dst) != len(src) {
+			return nil, fmt.Errorf("cannot rename %s to %s", src, dst)
+		}
+		for i := len(src) - 2; i >= 0; i-- {
```
nwt, line 45: Would it be better to iterate forward, thus showing the first path component that differs rather than the last?

zql/zql.peg
```diff
@@ -287,41 +298,34 @@ fieldExprList
...
+  = "sum"
+  / "avg"
+  / "stdev"
+  / "var"
+  / "entropy"
+  / "min"
+  / "max"
+  / "first"
+  / "last"
+  / "countdistinct"
```
nwt, line 328: Nit: Sort these.

zql/zql.peg
```diff
@@ -218,25 +217,29 @@ procList
...
-      RETURN(PREPEND(first, rest))
-    }
-
+  = "by"i _ columns:FlexAssignments { RETURN(columns) }
 
-groupByKey
-  = ExpressionAssignment
-  / field:fieldExpr { RETURN(MAP("op": "ExpressionAssignment", "target": TEXT, "expression": field)) }
+// A FlexAssignment is like an Assignment but it can optionally omit the lhs,
+// in which case, the semantic pass will infer a name from the rhs, e.g., for
```
nwt, line 233:
```suggestion
// in which case the semantic pass will infer a name from the rhs. E.g., for
```

nwt (approved):
:shipit:

filter/filter.go
```diff
@@ -193,15 +198,37 @@ func searchRecordString(term string) Filter {
 		}
 	}
 	searchContainer := Contains(search)
-
+	fieldNameCheck := make(map[int]bool)
 	return func(r *zng.Record) bool {
+		// Memoize the result of a search across the names in the
+		// record columns for each unique record type.
+		id := r.Type.ID()
+		match, ok := fieldNameCheck[id]
```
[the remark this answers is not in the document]
mccanne: The scanner will only work if the typeIDs in the value buffer match the typeIDs in its context.  If that gets messed up then nothing works, so I don't understand what you mean by unlikely Reset().  If you are envisioning some sort of race during parallel decoding, then that would be a bug.  What am I missing here?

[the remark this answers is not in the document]
mccanne: Good catch.  This context is different from the flow graph because it's in the scanner..  If we adopted your fix, then unused type pointers would hang in the table during the life of the scan.  That seems ok.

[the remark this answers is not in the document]
mccanne: Issue #1478 for "another forum" topic.

zql/zql.peg
mccanne: They are in the ordered by related functionality.   The procs are laid out this way too, grouped by like functionality rather than alphabetic.  I want to make this list go away in a later PR anyway (i.e., so anything of the form `| agg() ...` so we can add reducers without changing the grammarrs).

zql/zql.peg
mccanne: Hmm I always put that comment in.  Need to break the habit.

PR #1472, opened 2020-10-12
adjust zar ranges option to show span and chunk ranges

alfred-landrum:
This adjusts the `-ranges` option on zar ls & rm, used by ztests to verify layout of the chunks, to show the time range of the SpanInfo where the chunk is located. This will be useful for upcoming tests that have more overlapping chunks; only the overlap.yaml ztest needed real adjustment.


archive/overlap.go
```diff
@@ -69,6 +69,13 @@ func firstLastToSpan(x, y nano.Ts) nano.Span {
 	return nano.Span{Ts: x, Dur: 1}.Union(nano.Span{Ts: y, Dur: 1})
 }
 
+func spanToFirstLast(dir zbuf.Direction, span nano.Span) (nano.Ts, nano.Ts) {
+	if dir == zbuf.DirTimeForward {
+		return span.Ts, nano.Ts(int64(span.Ts) + span.Dur - 1)
+	}
+	return nano.Ts(int64(span.Ts) + span.Dur - 1), span.Ts
```
nwt, line 76:
```suggestion
		return span.Ts, span.End() - 1
	}
	return span.End() - 1, span.Ts
```

archive/walk.go
```diff
@@ -312,3 +312,43 @@ func RmDirs(ctx context.Context, ark *Archive) error {
 		return ark.dataSrc.RemoveAll(ctx, chunk.ZarDir(ark))
 	})
 }
+
+// A SpanInfo is a logical view of the records within a time span, stored
+// in one or more Chunks.
+type SpanInfo struct {
+	Span nano.Span
+
+	// chunks are the data files that contain records within this SpanInfo.
```
nwt, line 321:
```suggestion
	// Chunks are the data files that contain records within this SpanInfo.
```

archive/walk.go
```diff
@@ -312,3 +312,43 @@ func RmDirs(ctx context.Context, ark *Archive) error {
...
+	// The Chunks may have spans that extend beyond this SpanInfo, so any
+	// records from these Chunks should be limited to those that fall within a
+	// closed span constructed from First & Last.
```
nwt, line 324:
```suggestion
	// records from these Chunks should be limited to those that fall within
	// this SpanInfo's Span.
```
And note that spans are always half-open on the right, so "closed span" isn't well-defined..

nwt (approved):

PR #1483, opened 2020-10-13
NFM: support converting a multisource source to a worker request

alfred-landrum:
Changes the type that's used in the MultiSource source channel from a closure to an interface, and adds calls that can be used to convert the source to an api.WorkerRequest.

### DIFF
--- a/driver/compile.go
+++ b/driver/compile.go
@@ -33,6 +33,13 @@ type Config struct {
 	Warnings          chan string
 }
 
+func zbufDirInt(reversed bool) int {
+	if reversed {
+		return -1
+	}
+	return 1
+}
+
 func programPrep(program ast.Proc, sortKey field.Static, sortReversed bool) (ast.Proc, filter.Filter, ast.BooleanExpr, error) {
 	ReplaceGroupByProcDurationWithKey(program)
 	if sortKey != nil {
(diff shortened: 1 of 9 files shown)


driver/parallel.go
```diff
@@ -121,6 +123,26 @@ func (pg *parallelGroup) doneSource(sc ScannerCloser) {
 	delete(pg.scanners, sc)
 }
 
+// XXX Note for mark: This function isn't yet used, but is an example of how
```
alfred-landrum, line 126: fyi @marktwallace

zqd/search/worker.go
```diff
@@ -33,34 +30,24 @@ func NewWorkerOp(req api.WorkerRequest) (*WorkerOp, error) {
...
-		chunks[i].RecordCount = chunk.RecordCount
+	store, ok := st.(*archivestore.Storage)
+	if !ok {
+		return nil, zqe.ErrInvalid("unhandled storage type for WorkerOp: %T", store)
 	}
-
-	si := archive.SpanInfo{
-		Span:   req.Span,
-		Chunks: chunks,
+	src, err := store.MultiSource().SourceFromRequest(&req)
```
nwt, line 37: Why is SourceFromRequest a method of driver.MultiSource instead of archivestore.Storage?

driver/compile.go
```diff
@@ -33,6 +33,13 @@ type Config struct {
 	Warnings          chan string
 }
 
+func zbufDirInt(reversed bool) int {
+	if reversed {
+		return -1
+	}
+	return 1
+}
+
```
mattnibs, line 42: Just a note: I think we have several different ways we specify sort direction. We should probably consolidate these at some point.

zqd/search/worker.go
alfred-landrum: fyi, @nwt & I discussed some of the different packaging & layout options, and both want some time to think about how this should be structured. @nwt did make an good point that for making requests to the workers, we could use the api client instead of constructing the http requests.

driver/compile.go
mattnibs: Actually, should we change driver.MultiSource.OrderInfo() to return zbuf.Direction instead of bool?

alfred-landrum:
Closing, this will get incorporated into #1502

PR #1486, opened 2020-10-13
merge spaninfos with the same largest chunk

alfred-landrum:
This reduces the number of SpanInfo's needed to read for a query when contiguous SpanInfo's share the same largest chunk, measured by record count.

An example: imagine chunk1's span is 0, 20, and it has 100 records; chunk2's span is 10, 20 and it has 10 records. Without this change, we would generate 3 SpanInfos: 0-10 (chunk1),10-20 (chunk1,chunk2), and 20-30 (chunk2). With this change, we'll generate 2 SpanInfo's: 0-20 (chunk1, chunk2), and 10-20 (chunk2). 

This should help reduce the overall impact when a query includes a mix of large & small chunks, as might happen due to  longconn records.


archive/overlap.go
```diff
@@ -130,3 +140,63 @@ func boundaries(chunks []Chunk, dir zbuf.Direction, fn func(ts nano.Ts, firstChu
...
+	for i := 1; i < len(spans); i++ {
+		largest := largestChunk(spans[i])
+		if largest.Id != runLargest.Id {
+			res = append(res, mergeSpanInfos(run, dir))
+			run = []SpanInfo{spans[i]}
+			runLargest = largest
+		} else {
+			run = append(run, spans[i])
+		}
+	}
```
nwt, line 199:
```suggestion
	for _, s := range spans[1:] {
		largest := largestChunk(s)
		if largest.Id != runLargest.Id {
			res = append(res, mergeSpanInfos(run, dir))
			run = []SpanInfo{s}
			runLargest = largest
		} else {
			run = append(run, s)
		}
	}
```

archive/overlap.go
```diff
@@ -130,3 +140,63 @@ func boundaries(chunks []Chunk, dir zbuf.Direction, fn func(ts nano.Ts, firstChu
...
+	res = append(res, mergeSpanInfos(run, dir))
+	return res
```
nwt, line 201:
```suggestion
	return append(res, mergeSpanInfos(run, dir))
```

archive/overlap.go
```diff
@@ -5,13 +5,23 @@ import (
...
 )
 
 // mergeChunksToSpans takes an unordered set of Chunks with possibly overlapping
 // spans, and returns an ordered list of spanInfos, whose spans will be bounded
 // by filter, and where each SpanInfo contains one or more Chunks whose data
 // falls into the SpanInfo's span.
 func mergeChunksToSpans(chunks []Chunk, dir zbuf.Direction, filter nano.Span) []SpanInfo {
+	sinfos := alignChunksToSpans(chunks, dir, filter)
+	sinfos = mergeLargestChunkSpanInfos(sinfos, dir)
+	return sinfos
```
nwt, line 18:
```suggestion
	return mergeLargestChunkSpanInfos(sinfos, dir)
```

archive/overlap.go
```diff
@@ -5,13 +5,22 @@ import (
...
 // spans, and returns an ordered list of spanInfos, whose spans will be bounded
 // by filter, and where each SpanInfo contains one or more Chunks whose data
 // falls into the SpanInfo's span.
 func mergeChunksToSpans(chunks []Chunk, dir zbuf.Direction, filter nano.Span) []SpanInfo {
+	sinfos := alignChunksToSpans(chunks, dir, filter)
+	return mergeLargestChunkSpanInfos(sinfos, dir)
+}
+
+// alignChunksToSpans creates an ordered slice of SpanInfo's whose boundaries
+// match either the boundaries of a chunk, or in the case of overlapping chunks,
```
nwt, line 21:
```suggestion
// match either the boundaries of a chunk or, in the case of overlapping chunks,
```

archive/overlap.go
```diff
@@ -130,3 +139,62 @@ func boundaries(chunks []Chunk, dir zbuf.Direction, fn func(ts nano.Ts, firstChu
...
+				res.Chunks = append(res.Chunks, c)
+			}
+		}
+	}
+	chunksSort(dir, res.Chunks)
+	return res
+}
+
+// mergeLargestChunkSpanInfos merges contiguous SpanInfo's whose largest
+// Chunks by RecordCount is the same in the input slice.
```
nwt, line 181:
```suggestion
// Chunk by RecordCount is the same.
```

nwt (approved):

PR #1487, opened 2020-10-13
Remove *Literal case in expr.DotExprToField

nwt:
It's an unused and unwanted remnant of expr.DotExprToField's previous
life as expr.FieldExprToString.


mccanne (changes requested):
It seems okay to turn foo["bar"] into [ "foo", "bar" ] but I don't think we'd hit this unless someone typed a field access that way in a command-line option.  If you get rid of Literal, you should get rid of Index too.

nwt:
I didn't realize that foo["bar"] relies this. Let's keep it.

PR #1490, opened 2020-10-15
add where clause to reducers

mccanne:
This commit adds a "where" clause to reducers allowing you to
apply filtering to individual aggregations in a multi-column
group-by.  e.g.,

  lg=sum(foo) where x>100, sm=sum(foo) where x<100 by bar

This is needed in the beacon project.

This was also the right time to simplify the reducer syntax in the
peg grammar so that there is no longer an enumerated list of the available
aggregation functions.  Instead, the compiler pass makes sure the
the named reducer exists making it easier from now on to add reducers
without having to modify the grammar.

### DIFF
--- a/proc/groupby/ztests/where.yaml
+++ b/proc/groupby/ztests/where.yaml
@@ -1,0 +1,16 @@
+zql: "sum(x) where s='a', s2=sum(x) by key=animal | sort ."
+
+input: |
+  #0:record[animal:string,s:string,x:int32]
+  0:[cat;a;1;]
+  0:[dog;b;1;]
+  0:[cat;a;1;]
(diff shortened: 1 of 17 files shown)


mccanne:
There was some discussion that "where" should be used in other places but this needs specifics and consensus and belongs in a future PR.

proc/groupby/ztests/where.yaml
```diff
@@ -0,0 +1,16 @@
+zql: "sum(x) where s='a', s2=sum(x) by key=animal | sort ."
```
henridf, line 1: `sort .` 👍

henridf (approved):

PR #1493, opened 2020-10-15
add union aggregator

mccanne:
This commit adds a union aggregator that takes union over all of its
input values and produces a value of type set of this computed union.
The type of the input values must be uniform.

### DIFF
--- a/proc/groupby/row.go
+++ b/proc/groupby/row.go
@@ -19,11 +19,11 @@ type valCol struct {
 
 type valRow []valCol
 
-func newValRow(makers []reducerMaker) valRow {
+func newValRow(zctx *resolver.Context, makers []reducerMaker) valRow {
 	cols := make([]valCol, 0, len(makers))
 	for _, maker := range makers {
 		e := expr.NewDotExpr(maker.name)
-		cols = append(cols, valCol{maker.name, e, maker.create()})
+		cols = append(cols, valCol{maker.name, e, maker.create(zctx)})
 	}
 	return cols
 }
@@ -40,7 +40,8 @@ func (v valRow) ConsumePart(rec *zng.Record) error {
 		if !ok {
 			return errors.New("reducer row doesn't decompose")
 		}
-		v, err := col.nameExpr.Eval(rec)
+		resolver := col.nameExpr
+		v, err := resolver.Eval(rec)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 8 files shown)


reducer/reducer.go
```diff
@@ -19,7 +19,7 @@ var (
 	ErrFieldRequired = errors.New("field parameter required")
 )
 
-type Maker func() Interface
+type Maker func(*resolver.Context) Interface
```
henridf, line 22: This isn't a git conflict but runMany() in reducer_test.go also needs updating... else the build will fail after merging this.

mccanne: Fixed

henridf (commented):
I think you mentioned a size limit / circuit breaker on the forthcoming collect reducer; doesn't this need one too? Or maybe in a later PR?

mccanne:
> I think you mentioned a size limit / circuit breaker on the forthcoming collect reducer; doesn't this need one too? Or maybe in a later PR?

Yes, thanks!  I had the same thought and already added it to the upcoming branch.  I will bring it forward to this PR.

mccanne:
I think this is ready now.

proc/groupby/row.go
```diff
@@ -40,7 +40,8 @@ func (v valRow) ConsumePart(rec *zng.Record) error {
 		if !ok {
 			return errors.New("reducer row doesn't decompose")
 		}
-		v, err := col.nameExpr.Eval(rec)
+		resolver := col.nameExpr
+		v, err := resolver.Eval(rec)
```
nwt, line 44:
```suggestion
		v, err := col.nameExpr.Eval(rec)
```

reducer/union.go
```diff
@@ -0,0 +1,91 @@
...
+		for u.size > MaxValueSize {
+			u.deleteOne()
+			u.MemExceeded++
+		}
+	}
+}
+
+func (u *Union) deleteOne() {
+	for key := range u.val {
+		u.size -= len([]byte(key))
```
nwt, line 58:
```suggestion
		u.size -= len(key)
```

reducer/union.go
```diff
@@ -0,0 +1,91 @@
...
+		delete(u.val, key)
+		return
+	}
+}
+
+func (u *Union) Result() zng.Value {
+	b := zcode.NewBuilder()
```
nwt, line 65:
```suggestion
	var b zcode.Builder
```

nwt (approved):

proc/groupby/row.go
mccanne: that looks familiar

reducer/union.go
mccanne: Ok, I was thinking go did rune length on strings but I see it doesn't.  Thanks!

PR #1496, opened 2020-10-15
add collect aggregator

mccanne:
This commit adds an aggregator called "collect" that assembles its input 
into an array.

This is needed in the beacon work.

### DIFF
--- a/reducer/collect.go
+++ b/reducer/collect.go
@@ -1,0 +1,72 @@
+package reducer
+
+import (
+	"github.com/brimsec/zq/expr"
+	"github.com/brimsec/zq/zcode"
+	"github.com/brimsec/zq/zng"
+	"github.com/brimsec/zq/zng/resolver"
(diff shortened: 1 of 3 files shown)


reducer/collect.go
```diff
@@ -0,0 +1,72 @@
...
+	c.val = append(c.val, b)
+	c.size += len(b)
+	for c.size > MaxValueSize {
+		c.size -= len(c.val[0])
+		c.val = c.val[1:]
+	}
+}
+
+func (c *Collect) Result() zng.Value {
+	b := zcode.NewBuilder()
```
nwt, line 46:
```suggestion
	var b zcode.Builder
```

nwt (approved):

reducer/collect.go
mccanne: That looks familiar :-)

PR #1497, opened 2020-10-15
add "and" and "or" aggregators

mccanne:


henridf (approved):
I would suggest adding a test for `and`/`or` to https://github.com/brimsec/zq/blob/master/reducer/reducer_test.go#L73 which exercises the `xPart()` calls in multiple configurations (and likewise for collect/union but I missed those trains...)

mccanne:
> I would suggest adding a test for `and`/`or` to https://github.com/brimsec/zq/blob/master/reducer/reducer_test.go#L73 which exercises the `xPart()` calls in multiple configurations (and likewise for collect/union but I missed those trains...)

Will do but I prefer it as a ztest... coming up!

mccanne:
Just realize I had the spill trigger wrong in the union PR.  I'll fix and make sure both spill tests work exercise the partials path of union/logical/collect.

henridf:
> Will do but I prefer it as a ztest... coming up!

if a ztest can cover the conditions that reducer_test.go, I will be pleased to eat my hat.

mccanne:
> > Will do but I prefer it as a ztest... coming up!
> 
> if a ztest can cover the conditions that reducer_test.go exercises, I will be pleased to eat my hat.

My bad for not writing tests that covered the partials path on union and collect.  See PR #1503.  Do you take sauce on your hat?

henridf:
@mccanne: no hats have been served yet... this doesn't exercise all the conditions covered in reducer_test.go.

reducer/logical.go
```diff
@@ -0,0 +1,57 @@
...
+	} else {
+		l.val = l.val || val
+	}
+}
+
+func (l *Logical) Result() zng.Value {
+	if l.val {
+		return zng.True
+	}
+	return zng.False
```
henridf, line 43:
```suggestion
	return l.val
```

henridf (approved):
The test locks in `-limit` which is currently undocumented and I believe unused. Pointing it out to be sure but if that's ok and intentional, 👍

henridf:
> @henridf: RE: "undocumented"... is this is a reference to user-facing ZQL docs?

@philrz yes that's what I meant by "undocumented". I wouldn't suggest putting it in the docs at this point since I don't see a use for it. I think it's more of an "advanced tuning" option if anything.

philrz:
Thanks @henridf. That's what I guessed, but wanted to be sure. I know we discussed `-sortmem` recently and reached a similar conclusion. I figure once we have some experience building our own setups that tune all these settings to work at scale we could work on an advanced config guide with best practices for users that actually want to try to run this stuff themselves at scale with proper tunings.

reducer/ztests/logical-spill.yaml
```diff
@@ -0,0 +1,18 @@
+script: |
+  zq -t "t1=or(x>2),t3=and(x>0) by key with -limit 1 | sort ." in.tzng
```
nwt, line 2: Maybe make this a zql test instead of a script test?

nwt (approved):

reducer/logical.go
mccanne: l.val is a go bool and zng.True and zng.False are zng.Values

PR #1501, opened 2020-10-15
fix bug with groupby ast

mccanne:
This commit fixes a bug in the groupby ast where the limit arg wasn't
being passed through without a keys argument (arguably the limit isn't
needed without keys, but that will no longer be the case for large-valued
reducers using union or collect).

We also simplify the groupby ast assembly logic in the peg rule by
following the pattern that any optional args with leading space
have space in the child rule not the parent.

Also, we added several omitempty tags to make the go ast json more
in line with the javascript json.

### DIFF
--- a/zql/zql.peg
+++ b/zql/zql.peg
@@ -231,7 +231,7 @@ proc
     }
 
 groupByKeys
-  = "by"i _ columns:FlexAssignments { RETURN(columns) }
+  = _ "by"i _ columns:FlexAssignments { RETURN(columns) }
 
 // A FlexAssignment is like an Assignment but it can optionally omit the lhs,
 // in which case the semantic pass will infer a name from the rhs, e.g., for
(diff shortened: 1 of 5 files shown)


mccanne:
I wanted to use ztest for exercises the partials code path in the reducer tests, but the `-limit` option was being ignored by the groupby peg rule if there were no keys.  This PR will fix.

zql/zql.peg
```diff
@@ -314,21 +314,18 @@ ExprList
...
+    VAR(p) = MAP("op": "GroupByProc", "reducers": reducers)
+    if ISNOTNULL(keys) {
+      p["keys"] = keys
+    }
+    if ISNOTNULL(every) {
+      p["duration"] = every
+    }
+    if ISNOTNULL(limit) {
+      p["limit"] = limit
     }
```
nwt, line 327: Nit: Follow the order of the expressions in the rule.
```suggestion
    if ISNOTNULL(every) {
      p["duration"] = every
    }
    if ISNOTNULL(limit) {
      p["limit"] = limit
    }
    if ISNOTNULL(keys) {
      p["keys"] = keys
    }
```

nwt (approved):

PR #1502, opened 2020-10-16
zqd root process

marktwallace:
This implements distributed queries with a zqd root process that will parallelize the work in a FlowGraph by sending it to multiple zqd worker processes. This feature is intended for testing the scalability of distributed zqd and to try out K8s deployment. It is not yet ready for production.

A zqd root process must be provided with a list of the ports for its zqd worker with the command line argument -workers. Each zqd worker process must have access to the same spaces as the zqd root process.

Example:
```
# Start 3 zqd/workers
zqd listen -l :9871 -data spaces &> zqd-w1.log &
zqd listen -l :9872 -data spaces &> zqd-w2.log &
zqd listen -l :9873 -data spaces &> zqd-w3.log &
# Start the root zqd
zqd listen -l :9867 -data spaces -workers :9871,:9872,:9873  &> zqd-root.log &
```

### DIFF
--- a/archive/multisource.go
+++ b/archive/multisource.go
@@ -198,6 +198,7 @@ func (s *chunkSource) Open(ctx context.Context, zctx *resolver.Context, sf drive
 	rc := detector.MultiFileReader(zctx, paths, zio.ReaderOpts{Format: "zng"})
 	sn, err := scanner.NewScanner(ctx, rc, sf.Filter, sf.FilterExpr, sf.Span)
 	if err != nil {
+		rc.Close()
 		return nil, err
 	}
 	return &scannerCloser{Scanner: sn, Closer: rc}, nil
--- a/archive/multisource.go
+++ b/archive/multisource.go
@@ -97,9 +97,7 @@ func NewMultiSource(ark *Archive, altPaths []string) driver.MultiSource {
 			altPaths: altPaths,
 		}
 	}
-	return &spanMultiSource{
-		ark: ark,
-	}
+	return &spanMultiSource{ark}
 }
 
 type spanMultiSource struct {
--- a/archive/multisource.go
+++ b/archive/multisource.go
@@ -151,6 +151,7 @@ func (s *spanSource) Open(ctx context.Context, zctx *resolver.Context, sf driver
 
 func (s *spanSource) ToRequest(req *api.WorkerRequest) error {
 	req.Span = s.spanInfo.Span
+	req.DataPath = s.ark.DataPath.String()
 	for _, c := range s.spanInfo.Chunks {
 		req.ChunkPaths = append(req.ChunkPaths, c.RelativePath())
 	}
--- a/archive/multisource.go
+++ b/archive/multisource.go
@@ -2,6 +2,7 @@ package archive
 
 import (
 	"context"
+	"errors"
 	"io"
 
 	"github.com/brimsec/zq/ast"
@@ -9,12 +10,15 @@ import (
 	"github.com/brimsec/zq/field"
 	"github.com/brimsec/zq/filter"
 	"github.com/brimsec/zq/pkg/iosrc"
+	"github.com/brimsec/zq/pkg/nano"
 	"github.com/brimsec/zq/scanner"
 	"github.com/brimsec/zq/zbuf"
 	"github.com/brimsec/zq/zio"
 	"github.com/brimsec/zq/zio/detector"
 	"github.com/brimsec/zq/zio/zngio"
 	"github.com/brimsec/zq/zng/resolver"
+	"github.com/brimsec/zq/zqd/api"
+	"github.com/brimsec/zq/zqe"
 )
 
 type multiCloser struct {
@@ -76,11 +80,6 @@ func newSpanScanner(ctx context.Context, ark *Archive, zctx *resolver.Context, f
 	}, nil
 }
 
-type multiSource struct {
-	ark      *Archive
-	altPaths []string
-}
-
 // NewMultiSource returns a driver.MultiSource for an Archive. If no alternative
 // paths are specified, the MultiSource will send a source for each span in the
 // driver.SourceFilter span, and report the same ordering as the archive.
@@ -92,26 +91,88 @@ func NewMultiSource(ark *Archive, altPaths []string) driver.MultiSource {
 	if len(altPaths) == 1 && altPaths[0] == "_" {
 		altPaths = nil
 	}
-	return &multiSource{
-		ark:      ark,
-		altPaths: altPaths,
+	if len(altPaths) != 0 {
+		return &chunkMultiSource{
+			ark:      ark,
+			altPaths: altPaths,
+		}
+	}
+	return &spanMultiSource{
+		ark: ark,
 	}
 }
 
-func (m *multiSource) OrderInfo() (field.Static, bool) {
-	if len(m.altPaths) == 0 {
-		return field.New("ts"), m.ark.DataOrder == zbuf.OrderDesc
+type spanMultiSource struct {
+	ark *Archive
+}
+
+func (m *spanMultiSource) OrderInfo() (field.Static, bool) {
+	return field.New("ts"), m.ark.DataOrder == zbuf.OrderDesc
+}
+
+func (m *spanMultiSource) SendSources(ctx context.Context, span nano.Span, srcChan chan driver.Source) error {
+	return SpanWalk(ctx, m.ark, span, func(si SpanInfo) error {
+		select {
+		case srcChan <- &spanSource{ark: m.ark, spanInfo: si}:
+			return nil
+		case <-ctx.Done():
+			return ctx.Err()
+		}
+	})
+}
+
+func (m *spanMultiSource) SourceFromRequest(ctx context.Context, req *api.WorkerRequest) (driver.Source, error) {
+	si := SpanInfo{Span: req.Span}
+	for _, p := range req.ChunkPaths {
+		tsd, _, id, ok := parseChunkRelativePath(p)
+		if !ok {
+			return nil, zqe.E(zqe.Invalid, "invalid chunk path: %v", p)
+		}
+		md, err := readChunkMetadata(ctx, chunkMetadataPath(m.ark, tsd, id))
+		if err != nil {
+			return nil, err
+		}
+		si.Chunks = append(si.Chunks, md.Chunk(id))
 	}
+	return &spanSource{
+		ark:      m.ark,
+		spanInfo: si,
+	}, nil
+}
+
+type spanSource struct {
+	ark      *Archive
+	spanInfo SpanInfo
+}
+
+func (s *spanSource) Open(ctx context.Context, zctx *resolver.Context, sf driver.SourceFilter) (driver.ScannerCloser, error) {
+	return newSpanScanner(ctx, s.ark, zctx, sf.Filter, sf.FilterExpr, s.spanInfo)
+}
+
+func (s *spanSource) ToRequest(req *api.WorkerRequest) error {
+	req.Span = s.spanInfo.Span
+	for _, c := range s.spanInfo.Chunks {
+		req.ChunkPaths = append(req.ChunkPaths, c.RelativePath())
+	}
+	return nil
+}
+
+// A chunkMultiSource uses the archive.Walk call to provide a driver.Source
+// for each chunk in the archive, possibly combining its data with files named
+// by altPaths located in the chunk's zar directory.
+type chunkMultiSource struct {
+	ark      *Archive
+	altPaths []string
+}
+
+func (cms *chunkMultiSource) OrderInfo() (field.Static, bool) {
 	return nil, false
 }
 
-func (m *multiSource) spanWalk(ctx context.Context, zctx *resolver.Context, sf driver.SourceFilter, srcChan chan<- driver.SourceOpener) error {
-	return SpanWalk(ctx, m.ark, sf.Span, func(si SpanInfo) error {
-		so := func() (driver.ScannerCloser, error) {
-			return newSpanScanner(ctx, m.ark, zctx, sf.Filter, sf.FilterExpr, si)
-		}
+func (cms *chunkMultiSource) SendSources(ctx context.Context, span nano.Span, srcChan chan driver.Source) error {
+	return Walk(ctx, cms.ark, func(chunk Chunk) error {
 		select {
-		case srcChan <- so:
+		case srcChan <- &chunkSource{cms: cms, chunk: chunk}:
 			return nil
 		case <-ctx.Done():
 			return ctx.Err()
@@ -119,32 +180,30 @@ func (m *multiSource) spanWalk(ctx context.Context, zctx *resolver.Context, sf d
 	})
 }
 
-func (m *multiSource) chunkWalk(ctx context.Context, zctx *resolver.Context, sf driver.SourceFilter, srcChan chan<- driver.SourceOpener) error {
-	return Walk(ctx, m.ark, func(chunk Chunk) error {
-		so := func() (driver.ScannerCloser, error) {
-			paths := make([]string, len(m.altPaths))
-			for i, input := range m.altPaths {
-				paths[i] = chunk.Localize(m.ark, input).String()
-			}
-			rc := detector.MultiFileReader(zctx, paths, zio.ReaderOpts{Format: "zng"})
-			sn, err := scanner.NewScanner(ctx, rc, sf.Filter, sf.FilterExpr, sf.Span)
-			if err != nil {
-				return nil, err
-			}
-			return &scannerCloser{Scanner: sn, Closer: rc}, nil
-		}
-		select {
-		case srcChan <- so:
-			return nil
-		case <-ctx.Done():
-			return ctx.Err()
-		}
-	})
+var errReqForChunk = errors.New("no request support for chunk sources")
+
+func (cms *chunkMultiSource) SourceFromRequest(context.Context, *api.WorkerRequest) (driver.Source, error) {
+	return nil, errReqForChunk
 }
 
-func (m *multiSource) SendSources(ctx context.Context, zctx *resolver.Context, sf driver.SourceFilter, srcChan chan driver.SourceOpener) error {
-	if len(m.altPaths) == 0 {
-		return m.spanWalk(ctx, zctx, sf, srcChan)
+type chunkSource struct {
+	cms   *chunkMultiSource
+	chunk Chunk
+}
+
+func (s *chunkSource) Open(ctx context.Context, zctx *resolver.Context, sf driver.SourceFilter) (driver.ScannerCloser, error) {
+	paths := make([]string, len(s.cms.altPaths))
+	for i, input := range s.cms.altPaths {
+		paths[i] = s.chunk.Localize(s.cms.ark, input).String()
 	}
-	return m.chunkWalk(ctx, zctx, sf, srcChan)
+	rc := detector.MultiFileReader(zctx, paths, zio.ReaderOpts{Format: "zng"})
+	sn, err := scanner.NewScanner(ctx, rc, sf.Filter, sf.FilterExpr, sf.Span)
+	if err != nil {
+		return nil, err
+	}
+	return &scannerCloser{Scanner: sn, Closer: rc}, nil
+}
+
+func (s *chunkSource) ToRequest(*api.WorkerRequest) error {
+	return errReqForChunk
 }
--- a/cmd/zapi/cmd/get/get.go
+++ b/cmd/zapi/cmd/get/get.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
@@ -98,25 +99,26 @@ func (c *Command) Run(args []string) error {
 		expr = strings.Join(args, " ")
 	}
 	client := c.Client()
-	id, err := c.SpaceID()
-	if err != nil {
-		return err
-	}
 
 	var r io.ReadCloser
 	if c.chunkInfo == "" {
+		id, err := c.SpaceID()
+		if err != nil {
+			return err
+		}
 		req, err := parseExpr(id, expr)
 		if err != nil {
 			return fmt.Errorf("parse error: %s", err)
 		}
 		req.Span = nano.NewSpanTs(nano.Ts(c.from), nano.Ts(c.to))
+
 		params := map[string]string{"format": c.encoding}
 		r, err = client.SearchRaw(c.Context(), *req, params)
 		if err != nil {
 			return fmt.Errorf("search error: %w", err)
 		}
 	} else {
-		req, err := parseExprWithChunk(id, expr, c.chunkInfo)
+		req, err := parseExprWithChunk(expr, c.chunkInfo)
 		req.Span = nano.NewSpanTs(nano.Ts(c.from), nano.Ts(c.to))
 		params := map[string]string{"format": c.encoding}
 		if err != nil {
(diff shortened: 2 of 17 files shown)


marktwallace:
This branch implements distributed zqd queries, but it has a bug:
The order of records is being reversed on distributed queries. Try:

make TEST=TestZq/ztests/suite/zqd/worker-dist

This test "passes", but only because I am comparing the output of "zq tail" with "zqd head"

I still can't tell why the record order is reversed. There are some old comments in the code that say that zqd only supports time-reversed queries (Dir=-1) but I'm not sure that is correct, since both the head and tail queries are supported. There have been changes in the logic for ordering recently, that I merged from master into this branch.

marktwallace:
This problem was due to the order of smtp.log.gz being reversed when the archive was created, so the reversed order was expected.  Fixed by adding "sort -r ts" to the zq query.

> This branch implements distributed zqd queries, but it has a bug:
> The order of records is being reversed on distributed queries. Try:
> 
> make TEST=TestZq/ztests/suite/zqd/worker-dist
> 
> This test "passes", but only because I am comparing the output of "zq tail" with "zqd head"
> 
> I still can't tell why the record order is reversed. There are some old comments in the code that say that zqd only supports time-reversed queries (Dir=-1) but I'm not sure that is correct, since both the head and tail queries are supported. There have been changes in the logic for ordering recently, that I merged from master into this branch.

alfred-landrum (commented):
When I ran this locally using a root & only a single worker, I got a surprising result where it seemed like the operation (just a count() in my case) was off. Would you please try that?

cmd/zqd/listen/command.go
```diff
@@ -258,6 +265,24 @@ func (c *Command) initSuricata() error {
 	return nil
 }
 
+func (c *Command) initWorkers() error {
+	// This is for local testing only, at this point.
+	// Workers will be available through a K8s service in a prod deployment.
+	if c.workers != "" {
+		workerArr := strings.Split(c.workers, ",")
+		for i, addr := range workerArr {
+			// default host to 127.0.0.1
```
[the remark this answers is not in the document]
marktwallace: Great suggestion, I made that change.

driver/compile.go
```diff
@@ -130,8 +133,21 @@ func compileMulti(ctx context.Context, program ast.Proc, zctx *resolver.Context,
 	if mcfg.Warnings == nil {
 		mcfg.Warnings = make(chan string, 5)
 	}
+
 	if mcfg.Parallelism == 0 {
-		mcfg.Parallelism = runtime.GOMAXPROCS(0)
+		// If mcfg.Parallelism has not been set by external configuration,
+		// then it will be zero here.
+		if len(WorkerURLs) >= 2 {
```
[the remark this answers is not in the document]
marktwallace: Just 1 worker is a degenerate case that I do not think we should support. The root process should just run the proc directly rather that have a remote process do the work. Probably the parameter validation for -workers should reject a list of less than length 2. If we end up ever unhiding the -worker parameter, I will add the validation.

driver/compile.go
```diff
@@ -113,8 +114,10 @@ func compileSingle(ctx context.Context, program ast.Proc, zctx *resolver.Context
 
 type MultiConfig struct {
 	Custom      compiler.Hook
+	Dir         int
 	Logger      *zap.Logger
 	Parallelism int
+	SpaceID     api.SpaceID
```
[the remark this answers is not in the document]
marktwallace: Yes, I removed the reference from zqd/search.go also.

marktwallace:
> When I ran this locally using a root & only a single worker, I got a surprising result where it seemed like the operation (just a count() in my case) was off. Would you please try that?

The system test:
make TEST=TestZq/ztests/suite/zqd/worker-dist

tests the count() function, and passes. Since you have found a case for count() that does not pass, I want to fix it and then add it to the system tests. Please send me the details of the case where count() failed for you.

cmd/zapi/cmd/get/get.go
```diff
@@ -98,25 +99,26 @@ func (c *Command) Run(args []string) error {
...
+		id, err := c.SpaceID()
+		if err != nil {
+			return err
+		}
 		req, err := parseExpr(id, expr)
 		if err != nil {
 			return fmt.Errorf("parse error: %s", err)
 		}
 		req.Span = nano.NewSpanTs(nano.Ts(c.from), nano.Ts(c.to))
+
```
nwt, line 114:
```suggestion
```

cmd/zapi/cmd/get/get.go
```diff
@@ -173,15 +175,25 @@ func parseExpr(spaceID api.SpaceID, expr string) (*api.SearchRequest, error) {
...
+	searchRequest := &api.SearchRequest{
+		Proc: proc,
+		Dir:  -1,
+	}
+	dataPath := filepath.Join(filepath.Dir(chunkPath), "../..")
 	return &api.WorkerRequest{
 		SearchRequest: *searchRequest,
-		ChunkPaths:    []string{chunkInfo},
+		DataPath:      dataPath,
+		ChunkPaths:    []string{chunkPath},
```
nwt, line 196: Maybe shorten?
```go
	return &api.WorkerRequest{
		SearchRequest: api.SearchRequest{
			Proc: proc,
			Dir:  -1,
		},
		DataPath:      filepath.Join(filepath.Dir(chunkPath), "../.."),
		ChunkPaths:    []string{chunkPath},
```

driver/compile.go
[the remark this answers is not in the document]
alfred-landrum: I think we should support the single worker case; it will be helpful for testing & demos to support a single worker.

cmd/zqd/listen/command.go
```diff
@@ -152,6 +156,9 @@ func (c *Command) init() error {
 	if err := c.initZeek(); err != nil {
 		return err
 	}
+	if err := c.initWorkers(); err != nil {
+		return err
+	}
```
nwt, line 161: Put this either before initZeek and initSurcata or after them, not between them.

cmd/zqd/listen/command.go
```diff
@@ -258,6 +265,24 @@ func (c *Command) initSuricata() error {
 	return nil
 }
 
+func (c *Command) initWorkers() error {
+	// This is for local testing only, at this point.
+	// Workers will be available through a K8s service in a prod deployment.
+	if c.workers != "" {
+		workerArr := strings.Split(c.workers, ",")
```
nwt, line 272: This is a slice, not an array, so workerArr isn't the best name. I'd just call it workers.

ztests/suite/zqd/services.sh
```diff
@@ -37,10 +36,28 @@ export AWS_ACCESS_KEY_ID=minioadmin
...
+  zqd listen -l=localhost:0 -portfile="$portdir/zqd-w1" -data="$zqdroot" -loglevel=warn &> zqd-w1.log &
+  zqdw1pid=$!
+  zqd listen -l=localhost:0 -portfile="$portdir/zqd-w2" -data="$zqdroot" -loglevel=warn &> zqd-w2.log &
+  zqdw2pid=$!
+
+  awaitfile $portdir/zqd-w1
+  portw1=$(cat $portdir/zqd-w1 | tr -d '[:space:]')
+
+  awaitfile $portdir/zqd-w2
+  portw2=$(cat $portdir/zqd-w2 | tr -d '[:space:]')
```
nwt, line 50: Do we actually need tr here?

cmd/zqd/listen/command.go
```diff
@@ -61,6 +63,7 @@ type Command struct {
 	logger             *zap.Logger
 	devMode            bool
 	portFile           string
+	workers            string
```
nwt, line 66: Put this after brimfd to keep the hidden flags together.

cmd/zqd/listen/command.go
```diff
@@ -79,6 +82,7 @@ func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
 	f.Var(&c.logLevel, "loglevel", "level for log output (defaults to info)")
 	f.BoolVar(&c.devMode, "dev", false, "runs zqd in development mode")
 	f.StringVar(&c.portFile, "portfile", "", "write port of http listener to file")
+	f.StringVar(&c.workers, "workers", "", "comma separated list of [addr]:port for zqd/workers")
```
nwt, line 85: Put this after brimfd to keep the hidden flags together.

nwt: And how about getting rid of initWorkers entirely by doing this?
```go
f.StringVar(&c.workers, "workers", "", "comma-separated list of worker URLs")
```

driver/compile.go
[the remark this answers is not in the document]
marktwallace: I changed this to a > 0 condition, since it must match a similar test for WorkerURLs in parallel.go

cmd/zapi/cmd/get/get.go
marktwallace: Yes

ztests/suite/zqd/services.sh
marktwallace: No, we did not

zqd/search/search.go
```diff
@@ -78,11 +78,13 @@ func (s *SearchOp) Run(ctx context.Context, store storage.Storage, output Output
 	defer statsTicker.Stop()
 	zctx := resolver.NewContext()
 
+	store := spc.Storage()
 	switch st := store.(type) {
```
nwt, line 82:
```suggestion
	switch st := spc.Storage().(type) {
```

driver/parallel.go
```diff
@@ -48,9 +56,20 @@ func (ph *parallelHead) Pull() (zbuf.Batch, error) {
...
+				if sc == nil || err != nil {
+					return nil, err
+				}
+			} else {
+				// Worker process parallelism uses nextSourceForConn
+				sc, err = ph.pg.nextSourceForConn(ph.workerConn)
+				if sc == nil || err != nil {
+					return nil, err
+				}
 			}
```
nwt, line 73:
```suggestion
			if ph.workerConn == nil {
				// Thread (goroutine) parallelism uses nextSource
				sc, err = ph.pg.nextSource()
	
			} else {
				// Worker process parallelism uses nextSourceForConn
				sc, err = ph.pg.nextSourceForConn(ph.workerConn)
			}
			if sc == nil || err != nil {
				return nil, err
			}
```

driver/parallel.go
```diff
@@ -81,8 +100,9 @@ type parallelGroup struct {
 	pctx       *proc.Context
 	filter     SourceFilter
 	msrc       MultiSource
+	mcfg       MultiConfig // supports a hack -MTW
```
nwt, line 103:
```suggestion
	mcfg       MultiConfig
```

driver/parallel.go
```diff
@@ -114,13 +134,76 @@ func (pg *parallelGroup) nextSource() (ScannerCloser, error) {
...
+			if !ok {
+				return nil, pg.sourceErr
+			}
+
+			req, err := pg.sourceToRequest(src)
+			if err != nil {
+				return nil, err
+			}
+			if req == nil {
+				continue
```
nwt, line 154: sourcetoRequest only returns a nil request when it returns a non-nil err, so this continue is unreachable. Drop it and the for enclosing for loop.

driver/parallel.go
```diff
@@ -114,13 +134,76 @@ func (pg *parallelGroup) nextSource() (ScannerCloser, error) {
 	}
 }
 
+// nextSourceForConn is similar to nextSource, but instead of returning the scannerCloser
+// for an open file (i.e. the stream for the open file),
+// nextSourceForConn sends an zapi api request to a remote zqd worker process, and returns
```
nwt, line 139:
```suggestion
// nextSourceForConn sends a request to a remote zqd worker process, and returns
```

driver/parallel.go
```diff
@@ -114,13 +134,76 @@ func (pg *parallelGroup) nextSource() (ScannerCloser, error) {
...
+// the scannerCloser (i.e.output stream) for the remote zqd worker.
```
nwt, line 140:
```suggestion
// the ScannerCloser (i.e.output stream) for the remote zqd worker.
```

driver/parallel.go
```diff
@@ -164,13 +247,30 @@ func createParallelGroup(pctx *proc.Context, filt filter.Filter, filterExpr ast.
...
-	sources := make([]proc.Interface, mcfg.Parallelism)
-	for i := range sources {
-		sources[i] = &parallelHead{pctx: pctx, parent: nil, pg: pg}
+	var sources []proc.Interface
+	// Two type of parallelGroups:
+	if len(workerURLs) > 0 {
+		// If -worker URLs are passed in zqd listen command,
+		// and driver.compile has determined that execution should be parallel
+		// then the sources are parallelHead procs that hold connections to
+		// remote zqd workers.
```
nwt, line 261: Rephrase this so it doesn't refer to the zqd listen command or its flags.

driver/parallel.go
```diff
@@ -164,13 +247,30 @@ func createParallelGroup(pctx *proc.Context, filt filter.Filter, filterExpr ast.
...
+		sources = make([]proc.Interface, len(workerURLs))
+		for i := range sources {
+			conn := api.NewConnectionTo(workerURLs[i])
+			sources[i] = &parallelHead{pctx: pctx, parent: nil, pg: pg, workerConn: conn}
+		}
```
nwt, line 266:
```suggestion
		range _, w := range workerURLs {
			sources = append(sources, &parallelHead{pctx: pctx, pg: pg, workerConn: api.NewConnectionTo(w)}
		}
```

driver/parallel.go
```diff
@@ -164,13 +247,30 @@ func createParallelGroup(pctx *proc.Context, filt filter.Filter, filterExpr ast.
...
+	} else {
+		// Normal: the sources are regular parallelHead procs
+		// and Parallelism is determined by mcfg
+		sources = make([]proc.Interface, mcfg.Parallelism)
+		for i := range sources {
+			sources[i] = &parallelHead{pctx: pctx, parent: nil, pg: pg}
```
nwt, line 272:
```suggestion
			sources[i] = &parallelHead{pctx: pctx, pg: pg}
```

proc/compiler/compiler.go
```diff
@@ -160,6 +160,9 @@ func compileParallel(custom Hook, pp *ast.ParallelProc, c *proc.Context, parents
 		}
 	}
 	if len(parents) != n {
+		// TODO: This should not be able to happen.
+		// Consider modifying the flow control before this so parents and pp.Procs are
+		// created in the same loop. -MTW
```
nwt, line 165:
```suggestion
```

zqd/api/api.go
```diff
@@ -48,6 +48,7 @@ type SearchRequest struct {
 type WorkerRequest struct {
 	SearchRequest
 	ChunkPaths []string `json:"chunk_paths"`
+	DataPath   string   `json:"datapath"`
```
nwt, line 51:
```suggestion
	DataPath   string   `json:"data_path"`
```

driver/compile.go
```diff
@@ -104,6 +118,16 @@ func compileSingle(ctx context.Context, program ast.Proc, zctx *resolver.Context
 	return newMuxOutput(pctx, leaves, sn), nil
 }
 
+type MultiConfig struct {
+	Custom      compiler.Hook
+	Dir         int
```
mattnibs, line 123: Can you use zbuf.Order here and if you want an int just call `order.Int()`?
```suggestion
	Order         zbuf.Order
```

cmd/zapi/cmd/get/get.go
```diff
@@ -173,15 +174,23 @@ func parseExpr(spaceID api.SpaceID, expr string) (*api.SearchRequest, error) {
...
 		return nil, err
 	}
 	return &api.WorkerRequest{
-		SearchRequest: *searchRequest,
-		ChunkPaths:    []string{chunkInfo},
+		SearchRequest: api.SearchRequest{
+			Proc: proc,
+			Dir:  -1,
+		},
+		DataPath:   filepath.Join(filepath.Dir(chunkPath), "../.."),
```
mattnibs, line 192: If the path for a chunk is an s3 uri, this will fail for windows. I would suggest using iosrc.URI for this.

cmd/zqd/listen/command.go
```diff
@@ -258,6 +265,27 @@ func (c *Command) initSuricata() error {
...
+			host, port, err := net.SplitHostPort(addr)
+			if err != nil {
+				return err
+			}
+			if host == "" {
+				host = "127.0.0.1"
+			}
+			workers[i] = "http://" + host + ":" + port
+		}
+		driver.WorkerURLs = workers
```
mattnibs, line 284: nit: IMHO, this really should be a config option in zqd.Config. In general I like to avoid using globals because they give you less flexibility for writing in process unit tests.

archive/multisource.go
```diff
@@ -92,59 +91,120 @@ func NewMultiSource(ark *Archive, altPaths []string) driver.MultiSource {
...
+type spanSource struct {
+	ark      *Archive
+	spanInfo SpanInfo
+}
+
+func (s *spanSource) Open(ctx context.Context, zctx *resolver.Context, sf driver.SourceFilter) (driver.ScannerCloser, error) {
+	return newSpanScanner(ctx, s.ark, zctx, sf.Filter, sf.FilterExpr, s.spanInfo)
+}
+
+func (s *spanSource) ToRequest(req *api.WorkerRequest) error {
```
mattnibs, line 152: (This is more of a comment for further discussion, I don't think this should be addressed in this pr): I do not think archive (or driver for that matter) should import packages from zqd. packages outside of zqd should be lower level than zqd. At some point we should consider moving all logic having to do with running queries through distributed zqd instances into the zqd package. The apis in archive and driver should have hooks to make this possible.

cmd/zapi/cmd/get/get.go
mattnibs: Also, is there any reason why the server can't be the one to derive this from the provided chunk paths? Can we get rid of this field?

archive/multisource.go
```diff
@@ -92,59 +91,120 @@ func NewMultiSource(ark *Archive, altPaths []string) driver.MultiSource {
...
-		altPaths: altPaths,
+	if len(altPaths) != 0 {
+		return &chunkMultiSource{
+			ark:      ark,
+			altPaths: altPaths,
+		}
+	}
+	return &spanMultiSource{
+		ark: ark,
 	}
```
mattnibs, line 102: nit:
```suggestion
	return &spanMultiSource{ark: ark}
```

driver/compile.go
```diff
@@ -21,6 +21,13 @@ import (
 	"go.uber.org/zap"
 )
 
+// Global variable for driver package
+// which determines whether this go process will
+// (1) implement parallelism by engaging multiple
+// remote zqd /worker processes, or
+// (2) implement parallelism with local goroutines
```
nwt, line 28: Make this a proper doc comment.
```suggestion
// WorkerURLs, if not empty, causes this process to
// implement parallelism using worker processes
// instead of goroutines.
```

driver/parallel.go
```diff
@@ -114,13 +132,71 @@ func (pg *parallelGroup) nextSource() (ScannerCloser, error) {
...
 }
 
+// sourceToRequest takes a Source and converts it into a WorkerRequest
+func (pg *parallelGroup) sourceToRequest(src Source) (*api.WorkerRequest, error) {
+	var req api.WorkerRequest
+	if err := src.ToRequest(&req); err != nil {
+		return nil, err
+	}
+	filterExpr := pg.filter.FilterExpr
+	if filterExpr != nil {
```
nwt, line 189:
```suggestion
	if filterExpr := pg.filter.FilterExpr; filterExpr != nil {
```

archive/multisource.go
alfred-landrum: agreed; this was part of a commit I wrote to unblock the distributed demo work, and not intended to be permanent. @nwt , @marktwallace and I had a few discussions about how to structure this, but haven't yet decided on the best way, and thought this was ok for the moment.

cmd/zapi/cmd/get/get.go
alfred-landrum: FYI the `ChunkPaths` field above are relative chunk paths, relative to the data path of the archive, so they're not redundant with the DataPath field.

driver/compile.go
marktwallace: What happens here is that the Dir come from the inbound /search request, get copied into the MultiConfig, and then is used to create the outbound /worker request. In the inbound request and outbound request, Dir is an integer. What is the advantage of converting the inbound /search Dir into a zbuf.Order and then converting that back into an int for the outbound /worker request? 

This would impact some function signatures, like this one in search.go:
func (s *SearchOp) Run(ctx context.Context, dir int, spc space.Space, output Output) ...

I would like these function signatures to use whatever our most common conventions are.

cmd/zapi/cmd/get/get.go
marktwallace: I might be missing something about iosrc, but I do not see a way to do the equivalent of filepath.Join with iosrc. I was modeling this code roughly on the pattern in archive/walk.go, so I thought using relative file paths was OK in this context.

BTW, I tried this:

	chunkURI, err := iosrc.ParseURI(chunkPath)
...
	datapath := chunkURI.AppendPath("../..").String()

To get the relative path using iosrc, but it didn't work.

mattnibs: Actually given Al's comment that chunks will be relative, I think you should just do path.Join(path.Dir(chunkPath), "../..")

marktwallace: ok

archive/multisource.go
```diff
@@ -92,59 +91,120 @@ func NewMultiSource(ark *Archive, altPaths []string) driver.MultiSource {
...
+func (s *chunkSource) Open(ctx context.Context, zctx *resolver.Context, sf driver.SourceFilter) (driver.ScannerCloser, error) {
+	paths := make([]string, len(s.cms.altPaths))
+	for i, input := range s.cms.altPaths {
+		paths[i] = s.chunk.Localize(s.cms.ark, input).String()
 	}
-	return m.chunkWalk(ctx, zctx, sf, srcChan)
+	rc := detector.MultiFileReader(zctx, paths, zio.ReaderOpts{Format: "zng"})
+	sn, err := scanner.NewScanner(ctx, rc, sf.Filter, sf.FilterExpr, sf.Span)
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
marktwallace: good catch to add the close

cmd/zqd/listen/command.go
marktwallace: I agree that we should not keep the driver.WorkerURLs global. I'm not sure about using zqd.Config for this, either. I would prefer to give it some thought and change it in a later PR.

mattnibs: Sure

driver/compile.go
mattnibs: > I would like these function signatures to use whatever our most common conventions are.

Certainly, and this is the point of my comment. I'm trying to establish one way of specifying Order in the code base (see #1505) as we currently have several different ways of doing this. I've seen the translate bool to int function in other packages.

 > What is the advantage of converting the inbound /search Dir into a zbuf.Order and then converting that back into an int for the outbound /worker request?

MultiConfig is not just used w.r.t zqd, its also used in zar zq. This layer should not be concerned with translating api requests off of the wire. Eventually the zqd package will be concerned with this translation. Making this zbuf.Order move us to consistency.

> This would impact some function signatures, like this one in search.go:
func (s *SearchOp) Run(ctx context.Context, dir int, spc space.Space, output Output) ...

Yeah this is an example of an inconsistency around how we specify order that I would like to work towards correcting.

marktwallace: OK, I understand the goal now. I changed Dir to Order everywhere but the zapi api calls.

alfred-landrum (approved):

PR #1503, opened 2020-10-16
add tests and fix bugs with collect/union partials

mccanne:


reducer/collect.go
```diff
@@ -57,8 +64,15 @@ func (c *Collect) Result() zng.Value {
 }
 
 func (c *Collect) ConsumePart(zv zng.Value) error {
+	if c.typ == nil {
+		typ, ok := zv.Type.(*zng.TypeArray)
+		if !ok {
+			return errors.New("partial not an array type")
+		}
+		c.typ = typ.Type
```
henridf, line 72: If the incoming array is empty, I don't think we should set `c.typ` here.  If we do, this means that an upstream reducer sending an empty partial will lead to the downstream reducer discarding all subsequent non-empty partials from other upstreams.

(This edge case is covered in reducer_test.go 🤓)

henridf (approved):

mccanne:
Change approach in PR #1508.

PR #1508, opened 2020-10-16
switch to ztest for reducer tests

mccanne:
(This subsumes PR #1503, which is closing without merge...)

This commit switches the reducer package from using reducer_test.go over
to ztest.

We designed the ztests so they would exercise the partials path and also
give empty partials to the partial consumer, which needs to be handled
correctly.

Along the way we fixed bugs in collect/union for handling empty partials
and noticed first/last should not be decomposable and thus should not
implement the reducer.Decomposable interface.


henridf (approved):

philrz:
🎩

PR #1510, opened 2020-10-16
Move link checker to separate Workflow

philrz:
(This is equivalent to what just merged in https://github.com/brimsec/brim/pull/1153.)

The automation for the link checker has been doing its job insofar as it's noticed when external links go bad or act flaky. However, it's also been correctly noted that having it run as a Job in the same Action with the rest of ci.yml can be troublesome, since it makes it more difficult to debug when there's link problems at the same time as legit code breakages, or CI infra instability.

To address this, here we break out the link checker to its own separate Actions Workflow.

### DIFF
--- a/.github/workflows/linkcheck.yml
+++ b/.github/workflows/linkcheck.yml
@@ -1,0 +1,14 @@
+name: Link checker
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+
+jobs:
+  markdown-link-check:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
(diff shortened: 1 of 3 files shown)


.github/workflows/linkcheck.yml
```diff
@@ -0,0 +1,14 @@
+name: Link checker
```
nwt, line 1:
```suggestion
name: Markdown link check
```
And name the file markdown-link-check.yaml.

nwt (approved):

.github/workflows/linkcheck.yml
philrz: @nwt: Ok, changes made.

PR #1521, opened 2020-10-19
Fix intermittent zqd system test failure

mattnibs:
Intermittent bug came from an issue in the brimsec minio fork where
the portfile would be written before the minio service was ready to
receive requests. This has been solved here in https://github.com/brimsec/minio/pull/2,
bump minio to current version.

Also:
- Echo zqd/minio logs on wait file fail.

### DIFF
--- a/ztests/suite/zqd/services.sh
+++ b/ztests/suite/zqd/services.sh
@@ -6,7 +6,11 @@ function awaitfile {
   until [ -f $file ]; do
     let i+=1
     if [ $i -gt 5 ]; then
-      echo "timed out waiting for file \"$file\" to appear"
+      echo "timed out waiting for file \"$file\" to appear\n"
+      echo "minio log:\n"
+      cat minio.log
+      echo "zqd log:\n"
+      cat zqd.log
       exit 1
     fi
     sleep 1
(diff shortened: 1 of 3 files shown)


ztests/suite/zqd/services.sh
```diff
@@ -6,7 +6,11 @@ function awaitfile {
   until [ -f $file ]; do
     let i+=1
     if [ $i -gt 5 ]; then
-      echo "timed out waiting for file \"$file\" to appear"
+      echo "timed out waiting for file \"$file\" to appear\n"
+      echo "minio log:\n"
+      cat minio.log
+      echo "zqd log:\n"
+      cat zqd.log
```
nwt, line 13: The echo command already emits one newline. Are you sure you want to add another one inside these strings?

mattnibs: whoops

mattnibs, line 13:
```suggestion
      echo "timed out waiting for file \"$file\" to appear"
      echo "minio log:"
      cat minio.log
      echo "zqd log:"
      cat zqd.log
```

alfred-landrum (approved):

PR #1523, opened 2020-10-19
Add more beta notice items and adjust TZNG EOS text

philrz:
@alfred-landrum @mccanne: Per our discussion this morning, here I've made some adjustments to the ZNG spec to line up with loose ends we're holding off on implementing in zq for now:

* TZNG EOS
* ZQL syntax for `enum`
* `float16`, `float32`, and `decimal`

I also adjusted the TZNG EOS text to emphasize that it's recommended, but optional.

I've included references to open issues in case users stumble onto these bullet items and want to express their use cases for adding them sooner.

### DIFF
--- a/zng/docs/spec.md
+++ b/zng/docs/spec.md
@@ -3,18 +3,18 @@
 > ### Note: This specification is in BETA development.
 > We hope that no backward incompatible changes will be made during
 > the BETA phase.  We plan to
-> declare the specification stable and finalized in spring 2021.
+> declare the specification stable and finalized in Spring, 2021.
 >
 > [Zq](https://github.com/brimsec/zq/blob/master/README.md)'s
 > implementation of ZNG is tracking this spec and as it changes,
 > the zq output format is subject to change.  In this branch,
 > zq attempts to implement everything herein excepting:
 >
-> * only streams of `record` types (which may consist of any combination of
+> * Only streams of `record` types (which may consist of any combination of
 >   other implemented types) are supported by zq even though a stream of
 >   any types may currently be expressed in value messages.
->
-> TBD: implement TZNG EOS in zq
+> * [TZNG End of Stream](#414-end-of-stream) markers are not yet implemented. ([zq/1364](https://github.com/brimsec/zq/issues/1364))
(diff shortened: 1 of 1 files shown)


zng/docs/spec.md
```diff
@@ -3,18 +3,19 @@
 > ### Note: This specification is in BETA development.
 > We hope that no backward incompatible changes will be made during
 > the BETA phase.  We plan to
-> declare the specification stable and finalized in spring 2021.
+> declare the specification stable and finalized in Spring, 2021.
```
nwt, line 6: Neither capitalization nor comma is conventional here.

zng/docs/spec.md
```diff
@@ -663,10 +664,11 @@ A TZNG end-of-stream marker has the following form:
 ```
 #eos
 ```
-A TZNG stream or file should always be terminated with end-of-stream.
-This clears all of the previous type tag bindings and aliases, allowing
-multiple TZNG files with overlapping tags to be concatenated without the
-tags colliding.
+
+It is recommended that a TZNG stream or file be terminated with an
```
nwt, line 668: Nit: I think the active voice of "A TZNG stream or file should" reads better.

zng/docs/spec.md
```diff
@@ -663,10 +664,11 @@ A TZNG end-of-stream marker has the following form:
...
+end-of-stream marker. This explicitly clears all of the previous type tag
+bindings and aliases, allowing multiple TZNG files with overlapping tags to be concatenated without the tags colliding. However, a TZNG reader should not
```
nwt, line 670: Nit: Wrap this.
```suggestion
bindings and aliases, allowing multiple TZNG files with overlapping tags to be
concatenated without the tags colliding. However, a TZNG reader should not
```

nwt (approved):

zng/docs/spec.md
philrz: Ah, right you are. I mistakenly thought it was a similar caps/comma rule as when writing month/day/year such as "October 21, 2020". But I see now I was wrong about that and have fixed.

PR #1526, opened 2020-10-20
microindex: support descending indexes

### LINKED ISSUE #1329: microindex find support for descending keys
zng files in a zar archive are stored in descending timestamp order. We want to create microindexes for these files whose key/values are timestamps & seek offfset of the start of a zng stream. The existing microindex code doesn't support find operations unless the keys are in ascending order, though.

I think it could be done by writing the index keys in descending order, and only adding a flag to the finder operation to control the comparison operation; that appears to work in a local branch, at least.

mattnibs:
Add support for microindexes created with a sorted flow of records 
in descending order.

Additionally this pr adds ClosetGTE and ClosetLTE methods to
microindex.Finder: Functionality that will be needed for finding the
correct offsets for archive chunk seek indicies.

Closes #1329


microindex/finder.go
```diff
@@ -40,27 +40,63 @@ func NewFinder(ctx context.Context, zctx *resolver.Context, uri iosrc.URI) (*Fin
...
 // lookup searches for a match of the given key compared to the
-// key values in the records read from the reader.  If the boolean argument
-// "exact" is true, then only exact matches are returned.  Otherwise, the
-// record with the lagest key smaller than the key argument is returned.
-func lookup(reader zbuf.Reader, compare expr.KeyCompareFn, exact bool) (*zng.Record, error) {
+// key values in the records read from the reader.  If the op argument is "="
+// then only exact matches are returned.  Otherwise, the record with the
+// largest key smaller (or larger) than the key argument is returned.
+func lookup(reader zbuf.Reader, compare expr.KeyCompareFn, dir zbuf.Order, op string) (*zng.Record, error) {
+	if dir == zbuf.OrderAsc {
```
nwt, line 47:
```suggestion
func lookup(reader zbuf.Reader, compare expr.KeyCompareFn, order zbuf.Order, op string) (*zng.Record, error) {
	if order == zbuf.OrderAsc {
```

microindex/finder.go
```diff
@@ -40,27 +40,63 @@ func NewFinder(ctx context.Context, zctx *resolver.Context, uri iosrc.URI) (*Fin
...
 
 // lookup searches for a match of the given key compared to the
-// key values in the records read from the reader.  If the boolean argument
-// "exact" is true, then only exact matches are returned.  Otherwise, the
-// record with the lagest key smaller than the key argument is returned.
-func lookup(reader zbuf.Reader, compare expr.KeyCompareFn, exact bool) (*zng.Record, error) {
+// key values in the records read from the reader.  If the op argument is "="
+// then only exact matches are returned.  Otherwise, the record with the
+// largest key smaller (or larger) than the key argument is returned.
+func lookup(reader zbuf.Reader, compare expr.KeyCompareFn, dir zbuf.Order, op string) (*zng.Record, error) {
```
nwt, line 46: Maybe define integer constants (like eq, gte, and lte) for op instead of using a string?

microindex/finder.go
```diff
@@ -146,7 +189,26 @@ func (f *Finder) LookupAll(ctx context.Context, hits chan<- *zng.Record, keys *z
...
+	reader, err := f.search(compare)
+	if err != nil {
+		return nil, err
+	}
+	return lookup(reader, compare, f.trailer.Order, ">=")
+}
+
+// ClosestLTE returns the closets record that is less than or equal to the
+// provided key values.
+func (f *Finder) ClosestLTE(keys *zng.Record) (*zng.Record, error) {
```
nwt, line 211:
```suggestion
// ClosestGTE returns the closets record that is greater than or equal to the
// provided key values.
func (f *Finder) ClosestGTE(keys *zng.Record) (*zng.Record, error) {
	return closest(keys, ">=")
}

// ClosestLTE returns the closets record that is less than or equal to the
// provided key values.
func (f *Finder) ClosestLTE(keys *zng.Record) (*zng.Record, error) {
	return closest(keys, "<=")
}
		
func (f *Finder) closest(keys *zng.Record, op string) (*zng.Record, error) {
```

microindex/finder.go
```diff
@@ -158,7 +220,7 @@ func (f *Finder) LookupClosest(keys *zng.Record) (*zng.Record, error) {
 	if err != nil {
 		return nil, err
 	}
-	return lookup(reader, compare, false)
+	return lookup(reader, compare, f.trailer.Order, "<=")
```
nwt, line 223:
```suggestion
	return lookup(reader, compare, f.trailer.Order, op)
```

microindex/writer.go
```diff
@@ -78,45 +79,76 @@ type indexWriter struct {
...
 	return w, nil
 }
 
+type Option interface {
+	apply(*Writer)
+}
+
+type optionFunc func(*Writer)
+
+func (f optionFunc) apply(w *Writer) { f(w) }
```
nwt, line 124: Any reason not to just say `type Option func(*Writer)` and do away with optionFunc?

microindex/finder.go
mattnibs: Yeah probably should.

microindex/writer.go
mattnibs: In the grand scheme of things probably no. When  I look at it through golang docs I like the way it looks better. Also I like that it obscures the internals, so consumers of it don't try to make their own options. That said, the stakes are low.

cmd/microindex/convert/command.go
```diff
@@ -49,6 +50,7 @@ func newCommand(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
 	f.IntVar(&c.frameThresh, "f", 32*1024, "minimum frame size used in microindex file")
 	f.StringVar(&c.outputFile, "o", "index.zng", "name of microindex output file")
 	f.StringVar(&c.keys, "k", "", "comma-separated list of field names for keys")
+	f.BoolVar(&c.desc, "desc", false, "specify data is in descending order")
```
nwt, line 53: Nit: Put this between frameThresh and outputFile so these are in the same order as the fields of Command.

microindex/finder.go
```diff
@@ -39,28 +39,72 @@ func NewFinder(ctx context.Context, zctx *resolver.Context, uri iosrc.URI) (*Fin
...
+	gte
+	lte
+)
+
 // lookup searches for a match of the given key compared to the
-// key values in the records read from the reader.  If the boolean argument
-// "exact" is true, then only exact matches are returned.  Otherwise, the
-// record with the lagest key smaller than the key argument is returned.
-func lookup(reader zbuf.Reader, compare expr.KeyCompareFn, exact bool) (*zng.Record, error) {
+// key values in the records read from the reader.  If the op argument is "="
```
nwt, line 51:
```suggestion
// key values in the records read from the reader.  If the op argument is eql
```

microindex/finder.go
```diff
@@ -146,7 +197,19 @@ func (f *Finder) LookupAll(ctx context.Context, hits chan<- *zng.Record, keys *z
 	}
 }
 
-func (f *Finder) LookupClosest(keys *zng.Record) (*zng.Record, error) {
+// ClosestGTE returns the closets record that is greater than or equal to the
```
nwt, line 200:
```suggestion
// ClosestGTE returns the closest record that is greater than or equal to the
```

microindex/finder.go
```diff
@@ -146,7 +197,19 @@ func (f *Finder) LookupAll(ctx context.Context, hits chan<- *zng.Record, keys *z
...
+// provided key values.
+func (f *Finder) ClosestGTE(keys *zng.Record) (*zng.Record, error) {
+	return f.closest(keys, gte)
+}
+
+// ClosestLTE returns the closets record that is less than or equal to the
```
nwt, line 206:
```suggestion
// ClosestLTE returns the closest record that is less than or equal to the
```

microindex/microindex_test.go
```diff
@@ -148,6 +72,121 @@ func TestMicroIndex(t *testing.T) {
...
+	lines = append(lines, "#0:record[key:string,value:int32]")
+	for i := 0; i < size; i++ {
+		line := fmt.Sprintf("0:[port:port:%d;%d;]", i, i)
+		lines = append(lines, line)
+	}
+	return reader(strings.Join(lines, "\n")), nil
+}
+
+func tempDir(t *testing.T) string {
+	dir, err := ioutil.TempDir("", "table_test")
```
nwt, line 184:
```suggestion
	dir, err := ioutil.TempDir("", t.Name())
```

microindex/trailer.go
```diff
@@ -138,6 +143,13 @@ func recordToTrailer(rec *zng.Record) (*Trailer, error) {
 	if err != nil {
 		return nil, err
 	}
+	order, err := rec.AccessInt(OrderField)
+	if err != nil {
+		return nil, err
+	}
+	if order < 0 {
+		trailer.Order = zbuf.OrderDesc
+	}
```
nwt, line 152: Nit: I think this would be a little nicer as a boolean field named `DescendingField = "descending"`.
```go
	descending, err := rec.AccessBool(DescendingField)
	if err != nil {
		return nil, err
	}
	if descending {
		trailer.Order = zbuf.OrderDesc
	}

nwt (approved):

cmd/microindex/convert/command.go
mattnibs: yup

microindex/trailer.go
mattnibs: I like it

PR #1528, opened 2020-10-21
move chunk metadata to separate file

alfred-landrum:
This PR moves the first & last timestamp, and the record count of a chunk, from the name of the chunk file to a separate metadata file. While working on compaction for overlapping files, I ran into local filesystem (not S3) limitations on the maximum length of a file name, which implied I'd need to store the masked chunk ids in a separate file. Since we plan on eventually storing metadata about a chunk to use for query optimization, I think it makes sense to encode all of the existing meta info to a separate file now, and add more content to it later as needed.

A repercussion of storing the timestamps this way is that all of the metadata files in a tsdir must be consulted at query time to determine which chunks to include for an query's time span. However, the metadata is immutable, so it can easily & safely be cached keyed on the chunk id.

### DIFF
--- a/archive/staticsource.go
+++ b/archive/staticsource.go
@@ -5,22 +5,29 @@ import (
 
 	"github.com/brimsec/zq/driver"
 	"github.com/brimsec/zq/field"
+	"github.com/brimsec/zq/pkg/nano"
 	"github.com/brimsec/zq/zbuf"
 	"github.com/brimsec/zq/zng/resolver"
+	"github.com/brimsec/zq/zqe"
 )
 
-// staticSource is an implemetation of driver.MultiSource that provides
+// staticSource is an implementation of driver.MultiSource that provides
 // a single SpanInfo (with Chunks) to be processed by a zqd worker.
 // staticSource is used only for the zqd /worker call.
 type staticSource struct {
 	ark *Archive
-	si  SpanInfo
+	sis SpanInfoSource
 }
 
-func NewStaticSource(ark *Archive, si SpanInfo) driver.MultiSource {
+type SpanInfoSource struct {
+	Span               nano.Span
+	ChunkRelativePaths []string
+}
+
+func NewStaticSource(ark *Archive, sis SpanInfoSource) driver.MultiSource {
 	return &staticSource{
 		ark: ark,
-		si:  si,
+		sis: sis,
 	}
 }
 
@@ -29,8 +36,20 @@ func (s *staticSource) OrderInfo() (field.Static, bool) {
 }
 
 func (s *staticSource) SendSources(ctx context.Context, zctx *resolver.Context, sf driver.SourceFilter, srcChan chan driver.SourceOpener) error {
+	si := SpanInfo{Span: s.sis.Span}
+	for _, p := range s.sis.ChunkRelativePaths {
+		tsd, _, id, ok := parseChunkRelativePath(p)
+		if !ok {
+			return zqe.E(zqe.Invalid, "invalid chunk path: %v", p)
+		}
+		md, err := readChunkMetadata(ctx, chunkMetadataPath(s.ark, tsd, id))
+		if err != nil {
+			return err
+		}
+		si.Chunks = append(si.Chunks, md.Chunk(id))
+	}
 	so := func() (driver.ScannerCloser, error) {
-		return newSpanScanner(ctx, s.ark, zctx, sf.Filter, sf.FilterExpr, s.si)
+		return newSpanScanner(ctx, s.ark, zctx, sf.Filter, sf.FilterExpr, si)
 	}
 	select {
 	case srcChan <- so:
--- a/archive/staticsource.go
+++ b/archive/staticsource.go
@@ -20,8 +20,8 @@ type staticSource struct {
 }
 
 type SpanInfoSource struct {
-	Span               nano.Span
-	ChunkRelativePaths []string
+	Span       nano.Span
+	ChunkPaths []string
 }
 
 func NewStaticSource(ark *Archive, sis SpanInfoSource) driver.MultiSource {
@@ -37,7 +37,7 @@ func (s *staticSource) OrderInfo() (field.Static, bool) {
 
 func (s *staticSource) SendSources(ctx context.Context, zctx *resolver.Context, sf driver.SourceFilter, srcChan chan driver.SourceOpener) error {
 	si := SpanInfo{Span: s.sis.Span}
-	for _, p := range s.sis.ChunkRelativePaths {
+	for _, p := range s.sis.ChunkPaths {
 		tsd, _, id, ok := parseChunkRelativePath(p)
 		if !ok {
 			return zqe.E(zqe.Invalid, "invalid chunk path: %v", p)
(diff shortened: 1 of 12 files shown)


archive/staticsource.go
```diff
@@ -5,22 +5,29 @@ import (
...
 )
 
-// staticSource is an implemetation of driver.MultiSource that provides
+// staticSource is an implementation of driver.MultiSource that provides
 // a single SpanInfo (with Chunks) to be processed by a zqd worker.
 // staticSource is used only for the zqd /worker call.
 type staticSource struct {
 	ark *Archive
-	si  SpanInfo
+	sis SpanInfoSource
```
marktwallace, line 19: This will conflict with #1502, which does not use the archive and adds a driver.Source

archive/staticsource.go
```diff
@@ -29,8 +36,20 @@ func (s *staticSource) OrderInfo() (field.Static, bool) {
...
+		}
+		md, err := readChunkMetadata(ctx, chunkMetadataPath(s.ark, tsd, id))
+		if err != nil {
+			return err
+		}
+		si.Chunks = append(si.Chunks, md.Chunk(id))
+	}
 	so := func() (driver.ScannerCloser, error) {
-		return newSpanScanner(ctx, s.ark, zctx, sf.Filter, sf.FilterExpr, s.si)
+		return newSpanScanner(ctx, s.ark, zctx, sf.Filter, sf.FilterExpr, si)
```
marktwallace, line 52: Another 1502 conflict here, also with driver.Source
Are you thinking we should replace driver.Source with SpanInfoSource?

archive/walk.go
```diff
@@ -179,19 +194,38 @@ func (c Chunk) seekIndexPath(ark *Archive) iosrc.URI {
...
 
-func (c Chunk) LogID() LogID {
-	name := fmt.Sprintf("%s-%s-%d-%d-%d.zng", c.Kind, c.Id, c.RecordCount, c.First, c.Last)
-	return LogID(path.Join(dataDirname, newTsDir(c.First).name(), name))
+func (c Chunk) RelativePath() string {
+	return path.Join(dataDirname, c.tsDir().name(), fmt.Sprintf("%s-%s.zng", c.Kind, c.Id))
+}
+
+func parseChunkRelativePath(s string) (tsDir, FileKind, ksuid.KSUID, bool) {
+	ss := strings.Split(s, "/")
```
marktwallace, line 210: You could use filepath functions here to guard against weird stuff, like "//" and "../.." etc. But I'm not sure you would ever expect those strings here.

cmd/zapi/cmd/get/get.go
```diff
@@ -180,20 +179,9 @@ func parseExprWithChunk(spaceID api.SpaceID, expr string, chunkInfo string) (*ap
...
-		SearchRequest: *searchRequest,
-		Chunks: []api.Chunk{{
-			Id:          chunk.Id.String(),
-			First:       chunk.First,
-			Last:        chunk.Last,
-			Kind:        string(chunk.Kind),
-			RecordCount: chunk.RecordCount,
-		}},
+		SearchRequest:      *searchRequest,
+		ChunkRelativePaths: []string{chunkInfo},
```
marktwallace, line 184: Another conflict with 1502, which should be resolvable because 1502 just adds DataPath

zqd/search/worker.go
```diff
@@ -80,8 +64,7 @@ func (w *WorkerOp) Run(ctx context.Context, store storage.Storage, output Output
 
 	switch st := store.(type) {
 	case *archivestore.Storage:
-		return driver.MultiRun(ctx, d, w.proc, zctx, st.StaticSource(w.si), driver.MultiConfig{
-			Span:      w.si.Span,
+		return driver.MultiRun(ctx, d, w.proc, zctx, st.StaticSource(w.ss), driver.MultiConfig{
```
marktwallace, line 67: This changed too, since WorkerOp.Run no longer requires the store to be passed in 1502

ztests/suite/zqd/worker-count.yaml
```diff
@@ -15,14 +15,13 @@ script: |
...
   #
   ZNG_FILE_PATH=$(find . -name "d-*.zng" -print | head -1)
-  CHUNK_INFO=$(basename $ZNG_FILE_PATH)
   #
   # The count() from zq should be identical to 
   # the count() from zapi get -chunk
   #
   zq -t "count()" $ZNG_FILE_PATH > zqcount.tzng
   zapi -h $ZQD_HOST -s testsp get -t \
-    -chunk $CHUNK_INFO "count()" > zapicount.tzng
```
marktwallace, line 25: This change also happened in 1502, since I pass in the full path of the chunk rather than just the basename.

marktwallace: I have this weird feeling that last night we were doing "parallel programming" (in the same sense as "parallel play") in a lot of the same files. :-)

marktwallace (approved):
This looks good, and is a nice simplification of the approach to chunks.
We touched a lot of the same files yesterday, so we may want to pair up for a few minutes to go through the conflicts.

archive/walk.go
```diff
@@ -25,9 +29,10 @@ const (
 type FileKind string
 
 const (
-	FileKindUnknown FileKind = ""
-	FileKindData             = "d"
-	FileKindSeek             = "ts"
+	FileKindUnknown  FileKind = ""
+	FileKindData              = "d"
+	FileKindSeek              = "ts"
+	FileKindMetadata          = "m"
```
nwt, line 35:
```suggestion
	FileKindData              = "d"
	FileKindMetadata          = "m"
	FileKindSeek              = "ts"
```

archive/staticsource.go
```diff
@@ -5,22 +5,29 @@ import (
...
 type staticSource struct {
 	ark *Archive
-	si  SpanInfo
+	sis SpanInfoSource
 }
 
-func NewStaticSource(ark *Archive, si SpanInfo) driver.MultiSource {
+type SpanInfoSource struct {
+	Span               nano.Span
+	ChunkRelativePaths []string
```
nwt, line 24: I'd just call this ChunkPaths.

archive/walk.go
```diff
@@ -90,85 +95,95 @@ func tsDirVisit(ctx context.Context, ark *Archive, filterSpan nano.Span, visitor
...
+			}
+			s := m[id]
+			switch kind {
+			case FileKindData:
+				s.data = true
+				m[id] = s
+			case FileKindMetadata:
+				s.meta = true
+				m[id] = s
+			}
```
nwt, line 128:
```suggestion
			switch kind {
			case FileKindData:
				m[id].data = true
			case FileKindMetadata:
				m[id].meta = true
			}
```

nwt (approved):

alfred-landrum:
@marktwallace & I coordinated; we'll land this PR first, then I'll rebase #1502 on it.

archive/walk.go
alfred-landrum: you can't do that ;)

nwt: 🤦

PR #1529, opened 2020-10-22
Updates to Zeek compatibility doc to sync with ZNG beta

philrz:
Our story around `port` and `enum` has changed since the last time this doc was revised, so I bring those current here. I also broke out a separate section about aliases since we now handle more than one Zeek type via alias.

### DIFF
--- a/zng/docs/zeek-compat.md
+++ b/zng/docs/zeek-compat.md
@@ -3,6 +3,7 @@
 - [Introduction](#introduction)
 - [Equivalent Types](#equivalent-types)
 - [Example](#example)
+- [About Aliases](#about-aliases)
 - [Type-Specific Details](#type-specific-details)
   * [`double`](#double)
   * [`set`](#set)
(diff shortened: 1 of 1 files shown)


zng/docs/zeek-compat.md
```diff
@@ -90,6 +92,17 @@ $ zq -t zeek_types.log | zq -f zeek -
...
+
+Multiple Zeek types discussed below are represented via a
+[type alias](spec.md#3117-alias-typedef) to one of ZNG's
+[primitive types](spec.md#5-primitive-types). The use of the alias maintains
+the history of the field's original Zeek type such that `zq` may restore it
+if/when the field may be later output again in Zeek format. Knowledge of its
+original Zeek type may also enable special operations in ZQL that are unique to
+values known to have originated as a specific Zeek type, though no such
+operations are currently implemented in `zq`.
+
```
nwt, line 105: I think this would make a little more sense as a second paragraph in the next section (i.e., Type-Specific Details).

nwt (approved):

zng/docs/zeek-compat.md
philrz: Thanks for reviewing, @nwt. I took your suggestion of moving the paragraph and it looks good to me so I've committed it.

PR #1531, opened 2020-10-22
support compaction of overlapping chunks

### LINKED ISSUE #1266: Use file masking to support data rewrite
Per #1183 , support rewriting data in an archive, to support operations like merging small data files together, or to merge data from overlapping time spans.

alfred-landrum:
This change adds the `zar compact` command, with an optional `-purge` flag, that will look for runs of overlapping chunk files and write their data into a single chunk. As described in #1183, the newly written chunks "mask" the existing chunks, so that at query time, if both an original chunk and a "compacted" chunk have data for the same time span, only the data from the compacted chunk will be used. This implementation records any masked chunk ids in a chunk's metadata file.

Closes #1266.

(The high line count is due to the addition of test files.)


archive/import.go
```diff
@@ -287,16 +267,22 @@ func newChunkWriter(ctx context.Context, ark *Archive, tsDir tsDir, chunk Chunk)
...
+		masks:           masks,
 		needIndexWrite:  true,
-		tsDir:           tsDir,
+		tsd:             tsd,
 	}, nil
 }
 
+func (cw *chunkWriter) position() (int64, nano.Ts, nano.Ts) {
+	return cw.dataFileWriter.Position(), cw.firstTs, cw.lastTs
+}
```
marktwallace, line 284: What are the constraints on this triple? Will the .Position() be between cw.firstTs and cw.lastTs?

archive/import.go
```diff
@@ -326,51 +319,62 @@ func (cw *chunkWriter) abort() {
...
+		return Chunk{}, err
 	}
 	defer func() {
 		tf.Close()
 		os.Remove(tf.Name())
 	}()
 	zctx := resolver.NewContext()
 	tfr := zngio.NewReader(tf, zctx)
-	idxURI := cw.chunk.seekIndexPath(cw.ark)
+	chunk := chunkMd.Chunk(cw.id)
```
marktwallace, line 357: It looks like this close method writes one extra chunk based on the metadata, which will be an addition to the all the chunks written before calling close. Do I have this right, or am I missing something?

archive/overlap.go
```diff
@@ -32,9 +36,11 @@ func alignChunksToSpans(chunks []Chunk, order zbuf.Order, filter nano.Span) []Sp
 				// last timestamp was just before ts.
 				siSpan := firstLastToSpan(siFirst, prevTs(ts, order))
 				if filter.Overlaps(siSpan) {
+					chunks := copyChunks(siChunks, nil)
+					chunksSort(order, chunks)
```
marktwallace, line 40: Did you add the sort here for correctness? i.e. was the code doing the right thing prior to the sort being added?

archive/overlap.go
```diff
@@ -198,3 +209,199 @@ func mergeLargestChunkSpanInfos(spans []SpanInfo, order zbuf.Order) []SpanInfo {
 	}
 	return append(res, mergeSpanInfos(run, order))
 }
+
+// removeMaskedChunks performs an in-place edit of the spans input slice,
+// updating each SpanInfo to remove chunks that mask other chunks within the
```
marktwallace, line 214: By "masked" you mean a chunk completely contained within the span of one other chunk , right?

archive/import.go
alfred-landrum: The Position() return parameter is the current offset in the writer, and is used to cap the length of the chunk that's being written by the chunkWriter. The firstTs value is the timestamp of the first record written to the chunk, and the lastTs value is the timetamp of the most recent record. So over the course of calling the `Write(*zng.Record)` method on chunkWriter, firstTs will stay the same, but Position() and lastTs will reflect the most recent Write call.

archive/import.go
alfred-landrum: A single instance of a `chunkWriter` will only write a single chunk. The close() method returns the `Chunk` that it wrote. There's a lot of book keeping & other calls happening, due mostly to writing the seek index file (as a temporary, the flushed during close) at the same time as writing the chunk data file.

archive/overlap.go
alfred-landrum: Before this PR, and specifically the 214c3712b7cc06ad46c204218d80f13db1270909 commit, the `boundaries` function returned chunks in a predictable order, so this change ensures that is maintained. I don't think this is needed for correctness, since the code that actually uses the chunks for data operations (like newSpanScanner) works on them as a group, and performs ordering based on the records that come back. Having a predictable order on the chunks is useful for things likes tests, where we can rely on the ordering for comparisons when we're listing the chunks in the archive, though.

archive/overlap.go
alfred-landrum: That's correct; I used that term in the design writeup in #1183 . I can add a comment in the code defining that usage; I think it'd go best with the `Masks` field of the `Chunk` struct, what do you think?

archive/overlap.go
```diff
@@ -198,3 +209,199 @@ func mergeLargestChunkSpanInfos(spans []SpanInfo, order zbuf.Order) []SpanInfo {
...
+		}
+	}
+	var mc []Chunk
+	if trackMasked {
+	outer:
+		for id := range maskedChunks {
+			for _, si := range spans {
+				if spanInfoContainsChunk(si, id) {
+					continue outer
+				}
```
mattnibs, line 266: Just to clarify: This check is important because a chunk could be partially masked and we therefore need to keep this chunk around?

archive/overlap.go
```diff
@@ -198,3 +209,199 @@ func mergeLargestChunkSpanInfos(spans []SpanInfo, order zbuf.Order) []SpanInfo {
...
+			}
+			if !masked {
+				chunks = append(chunks, c)
+			}
+		}
+		// No need to sort chunks since we perform the mask removal in-order.
+		spans[i] = SpanInfo{
+			Span:   si.Span,
+			Chunks: chunks,
+		}
```
mattnibs, line 257: I would move this functionality to be a receiver function of `SpanInfo`. I think it would help readability. Something like: 
```golang
func (s *SpanInfo) RemoveMasked() (removed []Chunk)
```

archive/overlap.go
```diff
@@ -198,3 +209,199 @@ func mergeLargestChunkSpanInfos(spans []SpanInfo, order zbuf.Order) []SpanInfo {
...
+		res = append(res, mergeSpanInfos(run, order))
+		run = []SpanInfo{s}
+	}
+	return append(res, mergeSpanInfos(run, order))
+}
+
+func Compact(ctx context.Context, ark *Archive) error {
+	return tsDirVisit(ctx, ark, nano.MaxSpan, func(_ tsDir, chunks []Chunk) error {
+		spans := alignChunksToSpans(chunks, ark.DataOrder, nano.MaxSpan)
```
mattnibs, line 381: Why do we need this `alignChunksToSpans` call? Why not just do:
```golang
spanInfo := SpanInfo{tsDir.Span, chunks}
```

archive/overlap.go
```diff
@@ -198,3 +209,199 @@ func mergeLargestChunkSpanInfos(spans []SpanInfo, order zbuf.Order) []SpanInfo {
...
+				return err
+			}
+		}
+		return nil
+	})
+}
+
+func Purge(ctx context.Context, ark *Archive) error {
+	return tsDirVisit(ctx, ark, nano.MaxSpan, func(_ tsDir, chunks []Chunk) error {
+		spans := alignChunksToSpans(chunks, ark.DataOrder, nano.MaxSpan)
```
mattnibs, line 395: Same comment as above.

archive/walk.go
```diff
@@ -29,10 +29,11 @@ const (
 type FileKind string
 
 const (
-	FileKindUnknown  FileKind = ""
-	FileKindData              = "d"
-	FileKindMetadata          = "m"
-	FileKindSeek              = "ts"
+	FileKindUnknown       FileKind = ""
+	FileKindData                   = "d"
+	FileKindDataCompacted          = "dc"
```
mattnibs, line 34: Is it important to create a file kind for compacted vs not-compacted? Are they not both just data files?

alfred-landrum: Good question: It is no longer needed to differentiate them, and I'll make that change.

I started this work before #1528, and before that change, we needed to differentiate between the two cases so that if walk obtained a file with compacted data, it could look at the file type and know it must look at the metadata to get the masks list. Now that we are always looking at the metadata for the chunk, there's no need (I can think of) to differentiate between them.

archive/overlap.go
alfred-landrum: If we did that, then we would always compact all of the chunks in a tsdir, not just those that shared some overlapping time span.

archive/overlap.go
alfred-landrum: (response above)

archive/overlap.go
alfred-landrum: That is the intent of this check, and I can add a comment stating so. This is only used when `trackMasked` is true, which happens because of a call to Purge, looking to see what chunks within a whole tsDir are safe to remove. FYI, though, the `Compact` implementation in this PR should not create chunk files that only partially mask existing files, as far as I can imagine, but I think this extra safety check is a useful one to have.

archive/overlap.go
```diff
@@ -198,3 +209,199 @@ func mergeLargestChunkSpanInfos(spans []SpanInfo, order zbuf.Order) []SpanInfo {
...
+	tsd := newTsDir(s.Span.Ts)
+	var masks []ksuid.KSUID
+	for _, c := range s.Chunks {
+		masks = append(masks, c.Id)
+	}
+	mw := &compactWriter{
+		ark:   ark,
+		ctx:   ctx,
+		masks: masks,
+		tsd:   tsd,
```
nwt, line 346:
```suggestion
	var masks []ksuid.KSUID
	for _, c := range s.Chunks {
		masks = append(masks, c.Id)
	}
	mw := &compactWriter{
		ark:   ark,
		ctx:   ctx,
		masks: masks,
		tsd:   newTsDir(s.Span.Ts),
```

archive/walk.go
```diff
@@ -238,29 +245,63 @@ func (c Chunk) Localize(ark *Archive, pathname string) iosrc.URI {
...
+		return j.Last < i.Last
 	}
-	if jTs == iTs {
-		return ksuid.Compare(jKid, iKid) < 0
+	if j.RecordCount != i.RecordCount {
+		return j.RecordCount < i.RecordCount
 	}
-	return jTs < iTs
+	return ksuid.Compare(j.Id, i.Id) < 0
 }
```
nwt, line 300:
```go
func chunkLess(order zbuf.Order, a, b Chunk) bool {
	if order == zbuf.OrderDesc {
		a, b = b, a
	}
	switch {
	case a.First != b.First:
		return a.First < b.First
	case a.Last != b.Last:
		return a.Last < b.Last
	case a.RecordCount != b.RecordCount:
		return a.RecordCount < b.RecordCount
	}
	return ksuid.Compare(a.Id, b.Id) < 0
}
```

cmd/zar/compact/command.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/mccanne/charm"
+)
+
+var Compact = &charm.Spec{
+	Name:  "compact",
+	Usage: "compact [-R root]",
+	Short: "merge overlapping chunk files",
+	Long: `
+"zar compact" looks for chunk files whose time range overlaps, and writes
+new chunk files that combine their records.
```
nwt, line 19:
```suggestion
"zar compact" looks for chunk files whose time ranges overlap and writes
new chunk files that combine their records.
```

archive/overlap.go
alfred-landrum: 👍  will do

archive/overlap.go
alfred-landrum: I added a comment describing masks here: 50db6ed898636685b65ee25ffcde78436a96236f

mattnibs (approved):

archive/walk.go
```diff
@@ -370,6 +402,39 @@ func (s SpanInfo) ChunkRange(order zbuf.Order, chunkIdx int) string {
...
+	for _, c := range si.Chunks {
+		for _, mid := range c.Masks {
+			maskIds = append(maskIds, mid)
+		}
+	}
+	if len(maskIds) == 0 {
+		return nil
+	}
+	var chunks []Chunk
+	var removed []Chunk
```
nwt, line 419:
```suggestion
	var chunks, removed []Chunk
```

archive/walk.go
```diff
@@ -171,6 +171,10 @@ func chunkFileMatch(s string) (kind FileKind, id ksuid.KSUID, ok bool) {
 
 // A Chunk is a file that holds records ordered according to the archive's
 // data order.
+// To support reading chunks that contain records originally from one or
+// more other chunks, a chunk has a list of chunk ids it "masks". During a read
```
nwt, line 175:
```suggestion
// more other chunks, a chunk has a list of chunk IDs it "masks". During a read
```

archive/import.go
```diff
@@ -326,51 +317,56 @@ func (cw *chunkWriter) abort() {
...
-	})
-	if err != nil {
-		return err
+	chunkMd := chunkMetadata{
+		First:       firstTs,
+		Last:        lastTs,
+		RecordCount: cw.count,
+		Masks:       cw.masks,
+	}
+	if err := writeChunkMetadata(ctx, chunkMetadataPath(cw.ark, cw.tsd, cw.id), chunkMd); err != nil {
```
nwt, line 338:
```suggestion
	chunkMD := chunkMetadata{
		First:       firstTs,
		Last:        lastTs,
		RecordCount: cw.count,
		Masks:       cw.masks,
	}
	if err := writeChunkMetadata(ctx, chunkMetadataPath(cw.ark, cw.tsd, cw.id), chunkMD); err != nil {
```

nwt (approved):

PR #1532, opened 2020-10-23
Allow zapi post of s3 objects

marktwallace:
This change allows zapi push to be used with S3 data, e.g.

zapi -s sp1 post s3://mybucket/conn.log.gz

I wanted this feature for testing distributed zqd features, and I also think it is useful for end-users.

Three things needed to change:

(1) the zapi api needed to not change names prefixed by "s3://" into absolute paths

(2) zqd/ingest/log.go needed to switch from opening files to using ReadClosers provided by iosrc.
This was Al's idea, and it worked really well!

(3) zqd spills to /tmp when processing the post request, so I needed to add a /tmp dir in the K8s deployment template.

### DIFF
--- a/charts/zqd/templates/deployment.yaml
+++ b/charts/zqd/templates/deployment.yaml
@@ -84,7 +84,9 @@ spec:
             {{- toYaml .Values.resources | nindent 12 }}
           volumeMounts:
             - mountPath: /spaces
-              name: spaces-volume     
+              name: spaces-volume
+            - mountPath: /tmp
+              name: tmp-volume                   
       {{- with .Values.nodeSelector }}
       nodeSelector:
         {{- toYaml . | nindent 8 }}
@@ -99,4 +101,6 @@ spec:
       {{- end }}
       volumes:
         - name: spaces-volume
+          emptyDir: {}
+        - name: tmp-volume
           emptyDir: {}
--- a/cmd/zapi/cmd/post/post.go
+++ b/cmd/zapi/cmd/post/post.go
@@ -116,6 +116,19 @@ func abspaths(paths []string) ([]string, error) {
 	var err error
 	out := make([]string, len(paths))
 	for i, path := range paths {
+		// Special handling for paths starting with S3:
+		// we will not convert to absolute path -MTW
+		if path[0:4] == "s3:/" {
+			out[i] = path
+			continue
+		}
+		// TODO: This looks wrong to me,
+		// because it assumes that zapi and zqd
+		// are using the same filesystem. If they do not,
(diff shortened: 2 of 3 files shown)


cmd/zapi/cmd/post/post.go
```diff
@@ -116,6 +116,19 @@ func abspaths(paths []string) ([]string, error) {
 	var err error
 	out := make([]string, len(paths))
 	for i, path := range paths {
+		// Special handling for paths starting with S3:
```
alfred-landrum, line 119: How about calling `uri, err := ParseURI` (instead of the prefix check), and then setting `out[i] = uri.String()` of the result? That would pair well with using the ParseURI in zqd. For file paths, ParseURI will send the absolute path, so the filepath.Abs call below could be removed.

cmd/zapi/cmd/post/post.go
```diff
@@ -116,6 +116,19 @@ func abspaths(paths []string) ([]string, error) {
...
+		// we will not convert to absolute path -MTW
+		if path[0:5] == "s3://" {
+			out[i] = path
+			continue
+		}
+		// TODO: This looks wrong to me,
```
alfred-landrum, line 125: This is the current behavior because it was based around the Brim/zqd setup. The work in #1336 is a move towards not passing file paths at all to zqd.

cmd/zapi/cmd/post/post.go
marktwallace: Good suggestion, I will make that change

charts/zqd/templates/deployment.yaml
```diff
@@ -84,7 +84,9 @@ spec:
             {{- toYaml .Values.resources | nindent 12 }}
           volumeMounts:
             - mountPath: /spaces
-              name: spaces-volume     
+              name: spaces-volume
+            - mountPath: /tmp
```
alfred-landrum, line 88: Did you need to add this because there was no `/tmp` present on the filesystem? Or because you wanted to ensure the directory was empty on launch?

alfred-landrum (approved):

charts/zqd/templates/deployment.yaml
marktwallace: The /tmp directory was not present. Making /tmp a emptyDir{} will allow files written to /tmp to go to a ramdisk which will have the same lifetime as the container.

nwt: @marktwallace: A RAM disk isn't going to work very well for an S3 object larger than RAM.

PR #1533, opened 2020-10-23
Add "where" docs and reformat

philrz:
Here's my attempt to add ZQL docs for the recently-added `where` clause. If users try to get super sophisticated with this functionality right now they're likely to uncover some rough edges related to the differences between searches & expressions, which we have plans to address soon. But I figured it'd be good to get the docs out there with what's working.

I wondered if this belonged in the [Grouping](https://github.com/brimsec/zq/blob/master/zql/docs/grouping/README.md) page, but when comparing to how `by` and `every` work, this seems different enough that I've opted to include it with the other "General Usage" guidance. But I'm open to hearing other opinions on that.

I did some reformatting while I'm at it.

### DIFF
--- a/zql/docs/aggregate-functions/README.md
+++ b/zql/docs/aggregate-functions/README.md
@@ -4,30 +4,28 @@ A pipeline may contain one or more _aggregate functions_, which operate on
 batches of events to carry out a running computation over values contained in
 the events.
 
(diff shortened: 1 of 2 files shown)


henridf (approved):
Looks good! (and I think this is the right place; certainly compared to putting it under "Grouping" ).

PR #1534, opened 2020-10-23
ZQL docs for aggregate functions: and, or, collect, union

philrz:
This PR adds ZQL docs for the recently-added aggregate functions `and()`, `or()`, `collect()`, and `union()`. I cleaned up a touch of other formatting while I was in there.

I've also noticed that the stuff under https://pkg.go.dev/github.com/brimsec/zq that we link to from the Developer Docs entries only seems to get updated when we tag a new release. Therefore the Developer Docs links I've added to these new reducers are not "broken" per se but they don't land at meaningful content yet. Likewise, some of the existing links still point at valid content that's been recently moved as part of code re-org and are going to need to be adjusted when we bundle the next release. Rather than trying to be fancy about timing the merge of docs PRs at the precise moment we bundle the next release, I'm inclined to merge docs now for the benefit of bleeding edge users working off the new stuff `master`. I've made a note-to-self to check/update links when I assemble the next GA release.

### DIFF
--- a/zql/docs/aggregate-functions/README.md
+++ b/zql/docs/aggregate-functions/README.md
@@ -396,7 +396,7 @@ MIN
 
 |                           |                                                                |
 | ------------------------- | -------------------------------------------------------------- |
-| **Description**           | Returns the boolean value `true` if the provided expression evaluates to `true` for any input value. Contrast with [`and`](#and). |
+| **Description**           | Returns the boolean value `true` if the provided expression evaluates to `true` for one or more inputs. Contrast with [`and`](#and). |
 | **Syntax**                | `or(<expression>)`                                             |
 | **Required<br>arguments** | `<expression>`<br>A valid ZQL [expression](../expressions/README.md). |
 | **Optional<br>arguments** | None                                                           |
--- a/zql/docs/aggregate-functions/README.md
+++ b/zql/docs/aggregate-functions/README.md
@@ -105,7 +105,7 @@ TS                SHORT_RTT            SHORT_COUNT LONG_RTT             LONG_COU
 
 |                           |                                                                |
 | ------------------------- | -------------------------------------------------------------- |
-| **Description**           | Returns the boolean value `true` if the provided expression evaluates to `true` for all input values. Contrast with [`or`](#or). |
+| **Description**           | Returns the boolean value `true` if the provided expression evaluates to `true` for all inputs. Contrast with [`or`](#or). |
 | **Syntax**                | `and(<expression>)`                                            |
 | **Required<br>arguments** | `<expression>`<br>A valid ZQL [expression](../expressions/README.md). |
 | **Optional<br>arguments** | None                                                           |
--- a/zql/docs/aggregate-functions/README.md
+++ b/zql/docs/aggregate-functions/README.md
@@ -9,14 +9,18 @@ relevant to all aggregate functions, then the following
 [Available Aggregate Functions](#available-aggregate-functions) are
 documented in detail:
 
+* [`and`](#and)
 * [`avg`](#avg)
+* [`collect`](#collect)
 * [`count`](#count)
 * [`countdistinct`](#countdistinct)
 * [`first`](#first)
 * [`last`](#last)
 * [`max`](#max)
 * [`min`](#min)
+* [`or`](#or)
 * [`sum`](#sum)
+* [`union`](#union)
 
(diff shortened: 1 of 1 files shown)


zql/docs/aggregate-functions/README.md
```diff
@@ -66,6 +70,49 @@ QUICKEST LONGEST     TYPICAL
 
 # Available Aggregate Functions
 
+## `and`
+
+|                           |                                                                |
+| ------------------------- | -------------------------------------------------------------- |
+| **Description**           | Returns the boolean value `true` if the provided expression evaluates to `true` for all input values. Contrast with [`or`](#or). |
```
henridf, line 77:
```suggestion
| **Description**           | Returns the boolean value `true` if the provided expression evaluates to `true` for all input records. Contrast with [`or`](#or). |
```

zql/docs/aggregate-functions/README.md
```diff
@@ -66,6 +70,49 @@ QUICKEST LONGEST     TYPICAL
...
+unknown_HTTP_method                         140
+above_hole_data_without_any_acks            107
+...
+```
+
+To count the number of connections for which this was the _only_ category of
+`weird` event observed:
+
+```zq-command
+zq -f table 'only_bads=and(name="bad_HTTP_request") by uid | count() where only_bads=true' weird.log.gz
```
henridf, line 106: nice query!

henridf: is `where` already documented? I might be missing something but I didn't see it.

As a separate comment, this use of `where` could also be expressed with `filter` (unlike say `| .. c=count(), c2=count() where xyz | ...` which cannot).

henridf: > is where already documented? I might be missing something but I didn't see it.

Answering myself: this is #1533

henridf (approved):

zql/docs/aggregate-functions/README.md
philrz: @henridf: Is this proposed change of wording a reflection of personal preference, or is there a substantial correctness argument here that I need to better understand? My use of the word "values" came from the same place as how I worded the high-level definition of an aggregate function at the top of the same doc (and therefore I'm prepared to accept critiques of this as well):

> A pipeline may contain one or more aggregate functions, which operate on batches of events to carry out a running computation over values contained in the events.

So now here when documenting `and()`, I was using "values" as shorthand for something like "what the expression evaluates to based on the values it takes as inputs and the computations performed on them in that expression". To me, using "records" here instead implies the aggregate function is somehow operating on the entirety of an event rather than just the parts referenced in the expression.

Having written that all out, now I'm wondering if I should just say "...for all inputs" and leave it at that. What do you think?

zql/docs/aggregate-functions/README.md
philrz: @henridf: Indeed. I was planning on holding off merging _this_ PR until #1533 gets reviewed/merged because:

1. I reference `where` here (as you noticed), and,
2. There's formatting changes in #1533 that'll create merge conflicts, so I'll plan to resolve them in this branch before merging.

Your point about `filter` is a fine one. I could add the `c=count(),` and say in the description that we're including the total for comparison, which would cement the fact that `where` is _necessary_ rather than just convenient. However, I'm inclined to keep it simple as it is. The use of `where` in these simpler contexts reads slightly more intuitive to me than having separate pipeline elements for `filter` and `count()`, so I'm inclined to start referencing `where` so users become aware it's available.

zql/docs/aggregate-functions/README.md
henridf: > @henridf: Is this proposed change of wording a reflection of personal preference, or is there a substantial correctness argument here that I need to better understand?

Since the expression is evaluated over an input record (matching the sentence you quoted, `operate on batches of events`, modulo the use of event vs record), it seemed more accurate to phrase it this way.  

I like your proposed "for all inputs". It's shorter, at least as clear, and as you've pointed out elsewhere, this is trying to be a helpful doc rather than a formal spec. 👍

zql/docs/aggregate-functions/README.md
henridf: 👍

zql/docs/aggregate-functions/README.md
philrz: Thanks @henridf! "...for all inputs" it is.

PR #1542, opened 2020-10-27
Move default case to bottom of all switch statements

nwt:
A thing of beauty is a joy forever.

A thing of consistency is a joy for as long as it remains consistent.


mattnibs (approved):
I wonder if there is a lint rule to enforce this?

nwt:
@mattnibs: Didn't find one with a quick search.

PR #1545, opened 2020-10-28
Distributed zqd on K8s using HAProxy with "leastconn" load balancing  

marktwallace:
Moving forward with Noah's idea to use an out-of-the-box least-connection load balancing algorithm rather than implement custom job scheduling for now.


marktwallace:
I copied the changes I needed from this PR into #1695.

PR #1546, opened 2020-10-29
rearrange zqd and zar related code

alfred-landrum:
In preparation for the upcoming licensing change for the server related components, this rearranges the code for zqd/zar related packages, moving them under a new `ppl` directory. I also moved some code out of the api package that was either client or server specific, and not just describing api messages.

This entire change is rearrangement; there should be no implementation or api changes. I've run Brim locally against these changes w/o issues.


philrz:
FYI, there's a couple links to README files in the moved `cmd/` directories in the Brim repo's top-level README and Wiki articles in the `docs/` directory. So I guess we can brace ourselves for a need for an immediate Brim-side change once this gets merged.

api/api.go
```diff
@@ -277,3 +276,47 @@ type IndexPostRequest struct {
 	InputFile  string          `json:"input_file"`
 	OutputFile string          `json:"output_file"`
 }
+
+type StorageKind string
+
+const (
+	UnknownStore StorageKind = ""
+	FileStore    StorageKind = "filestore"
+	ArchiveStore StorageKind = "archivestore"
```
nwt, line 285:
```suggestion
	ArchiveStore StorageKind = "archivestore"
	FileStore    StorageKind = "filestore"
```

api/api.go
```diff
@@ -277,3 +276,47 @@ type IndexPostRequest struct {
...
+	case FileStore:
+		fallthrough
+	case ArchiveStore:
+		*k = kx
+		return nil
+
+	default:
+		return fmt.Errorf("unknown storage kind: %s", x)
+	}
+}
```
nwt, line 304:
```suggestion
func (k *StorageKind) Set(s string) error {
	switch s := StorageKind(s); s {
	case ArchiveStore, FileStore:
		*k = s
		return nil
	}
	return fmt.Errorf("unknown storage kind: %s", s)
}
```

cmd/zapi/cmd/cli.go
```diff
@@ -8,8 +8,9 @@ import (
 	"os"
 	"syscall"
 
+	"github.com/brimsec/zq/api"
 	"github.com/brimsec/zq/cli"
-	"github.com/brimsec/zq/zqd/api"
+	"github.com/brimsec/zq/cmd/zapi/connection"
```
nwt, line 13: I think we should keep all this stuff in the api package, but if we do move it, I think it ought to move to github.com/brimsec/zq/api/client.

alfred-landrum: I would like to move it out of `api`, to remove as many dependencies of that package as we can (for example, the resty client). I'm fine with moving it to `api/client`.

api/api.go
```diff
@@ -277,3 +276,42 @@ type IndexPostRequest struct {
...
+func (k *StorageKind) Set(s string) error {
+	switch s := StorageKind(s); s {
+	case ArchiveStore, FileStore:
+		*k = s
+		return nil
+	}
+	return fmt.Errorf("unknown storage kind: %s", s)
+}
+
+type Config struct {
```
nwt, line 301: Since this no longer imports as space.Config, I think it ought to be called StorageConfig.

cmd/zapi/cmd/cli.go
```diff
@@ -75,12 +76,12 @@ func (c *Command) Context() context.Context {
 	return c.ctx
 }
 
-// Client returns a central api.Connection instance.
-func (c *Command) Client() *api.Connection {
-	if c.client == nil {
-		c.client = api.NewConnectionTo("http://" + c.Host)
+// Connection returns a central api.Connection instance.
```
nwt, line 79:
```suggestion
// Connection returns a central client.Connection instance.
```

nwt (approved):

Issue #1519, opened 2020-10-19
python zqd api client


alfred-landrum:
As a first step towards experimenting with Jupyter notebooks, we need a Python module that can be used to issue ZQL queries against a zqd endpoint, obtaining responses in zjson and translating to native Python objects.

PR #1549, opened 2020-11-02
Add api.SearchRequest.ZQL as a .Proc alternative

nwt:
A zqd client must provide an AST in api.SearchRequest.Proc, requiring
the client to implement a ZQL parser if it is to allow searches with
arbitrary ZQL.  Add api.SearchRequest.ZQL to lift this requirement.
Either the Proc field or the ZQL field may be provided in requests to
zqd, but providing both is an error.

Part of #1519.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -37,9 +37,11 @@ type TaskEnd struct {
 	Error  *Error `json:"error,omitempty"`
 }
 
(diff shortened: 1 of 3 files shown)


alfred-landrum (approved):

henridf (approved):

nwt:
Closed in favor of #1563.

Issue #1002, opened 2020-07-14
Consistent timestamp output formatting


philrz:
[A user on public Slack observed](https://brimsec.slack.com/archives/CTSMAK6G7/p1594740349054600):

> is there a way to change the output timestamp format? decimal times don't really mean much as im looking at a table, i'd like to see it as iso8601

Indeed, things are a bit inconsistent in this area as of `zq` commit `85c31a3`. Here's the list of available output formats:

```
$ zq help 2>&1 | grep "format for output"
    -f format for output data [zng,ndjson,table,text,types,zeek,zjson,tzng] (default "zng")
```

Of these, `ndjson`, `table`, `text`, and the just-added `csv` all seem like formats where the user might want either format.

We see that `ndjson` and `table` currently default to epoch.

```
$ zq -f ndjson 'cut ts | head 1' stats.log.gz 
{"ts":"1521911720.600725"}

$ zq -f table 'cut ts | head 1' stats.log.gz 
TS
1521911720.600725
```

Whereas `text` and `csv` default to ISO8601.

```
$ zq -f text 'cut ts | head 1' stats.log.gz 
2018-03-24T17:15:20.600725Z

$ zq -f csv 'cut ts | head 1' stats.log.gz 
ts
2018-03-24T17:15:20.600725Z
```

However, there's also an option to change these to epoch.

```
$ zq help 2>&1 | grep "epoch"
    -E display epoch timestamps in csv and text output (default "false")

$ zq -E -f text 'cut ts | head 1' stats.log.gz 
1521911720.600725

$ zq -E -f csv 'cut ts | head 1' stats.log.gz 
ts
1521911720.600725
```

The proposal is to standardize on this latter behavior such that `ndjson` and `table` also output ISO8601 by default, but their current behavior could also be restored by the user via `-E`.

(Side note: The Brim app has its "Preferences" menu that lets the user set this formatting, but it's not currently possible in `zq`. In the past I've argued that issues like this are a reminder that we could at some point implement persistent preferences in `zq`. We're slowly accumulating command-line options that I could imagine users thinking "Yeah, I want it set like that all the time." But I respect that's out-of-scope of this specific issue.)

captainGeech42:
+1 for a config file or something for persistent options, that'd be great

mikesbrown:
Note as a workaround, `zq -f text` will give you ISO 8601.

philrz:
[...]

PR #1551, opened 2020-11-03
Change zng.TypeTime format to time.RFC3339Nano for "zq -f table"

nwt:
"zq -f table" formats zng.TypeTime values as fractional seconds since
the Unix epoch.  Change that to time.RFC3339Nano with -E restoring the
epoch format, as for "-f cvs" and "-f text".

Part of #1002.


zio/tableio/writer.go
```diff
@@ -68,16 +68,25 @@ func (w *Writer) Write(r *zng.Record) error {
 		w.writeHeader(w.typ)
 		w.nline = 0
 	}
-	//XXX only works for zeek-oriented records right now (won't work for NDJSON nested records)
-	ss, changePrecision, err := zeekio.ZeekStrings(r, w.precision, w.format)
```
henridf, line 72: Only one use of ZeekStrings remaining. 👍

henridf (approved):

mattnibs (approved):

PR #1557, opened 2020-11-03
Format zng.TypeTime values consistently in zq output

nwt:
For the csv, table, and text output formats, zq formats zng.TypeTime
values as fractional seconds since the Unix epoch when they appear in a
column not named ts.  Instead, format all zng.TypeTime values
consistently: with time.RFC3339Nano by default, and as factional seconds
since the Unix epoch if the -E flag is set.

Part of #1002.


henridf (approved):
🥳 one `"ts"` less

PR #1558, opened 2020-11-03
adjust Dockerfile for zqd move

alfred-landrum:
I missed this in https://github.com/brimsec/zq/pull/1546 .


marktwallace (approved):
This now matches the directory change in #1546

Issue #1559, opened 2020-11-03
change expr.Function to an inteface type


mccanne:
This will allow functions to refer to a type context (and possibly other contextual state) that is set at compile-time rather than passed-in.

PR #1560, opened 2020-11-04
change zng type error from native type to alias

### LINKED ISSUE #1524: Hook type context into expressions library package
There's a few things that would benefit from type context being accessible from the expression library package:

* Implementing the ZNG `error` type as an alias rather than a primitive type
* Proper casting of complex types
* Creating complex literals

mccanne:
This commit removes the "error" type from the zng spec and updates
the zq implementation to use an alias type for "error" (aliased
to string) instead of the native type.

This change requires passing the type context into the expression compiler.
Right now, we've changed the type signature of expr.Function type
to take the type context, but in a future commit, we will change
the function the be an interface type that can hold the type context
when needed (see issue #1559)


zng/docs/spec.md
```diff
@@ -911,8 +911,7 @@ representing machine words are serialized in little-endian format.
 | `ip`       | 18 | 4 or 16  | 4 or 16 bytes of IP address                    | a string representing an IP address in [IPv4 or IPv6 format](https://tools.ietf.org/html/draft-main-ipaddr-text-rep-02#section-3) |
 | `net`      | 19 | 8 or 32  | 8 or 32 bytes of IP prefix and subnet mask     | a string in CIDR notation representing an IP address and prefix length as defined in RFC 4632 and RFC 4291. |
 | `type`     | 20 | variable | UTF-8 byte sequence of string representing type in the [TZNG type grammar](#42-type-grammar)  |  a UTF-8 string |
-| `error`    | 21 | variable | UTF-8 byte sequence of string of error message | a UTF-8 string |
-| `null`     | 22 |    0     | No value, always represents an undefined value | the literal value `-`                                 |
+| `null`     | 21 |    0    | No value, always represents an undefined value | the literal value `-`                                 |
```
alfred-landrum, line 914: This doesn't match up with the `idReserved` or `idNull` entries in zng/types.go.

mccanne: I marked this as reserved for now.  Should we change "23" to "22" and get rid of the reserved slot? Would require one more rev to the sample-data repo.

philrz: @mccanne: Revising the sample data is no biggie, so, whatever you think is cleanest long-term.

zng/resolver/context.go
```diff
@@ -812,3 +813,19 @@ func (c *Context) TranslateTypeMap(ext *zng.TypeMap) (*zng.TypeMap, error) {
 	}
 	return c.LookupTypeMap(keyType, valType), nil
 }
+
+func (c *Context) NewError(format string, args ...interface{}) (zng.Value, error) {
+	msg := fmt.Sprintf(format, args...)
+	if c.typeErr == nil {
+		typ, err := c.LookupTypeAlias("error", zng.TypeString)
+		if err != nil {
+			return zng.Value{}, err
```
nwt, line 822: Allowing this to fail causes surprising behavior if `error` is already aliased to a type other than `string`.

```sh
$ go run ./cmd/zq -t 'put i=1/i' - <<EOF
#error=bstring
#0:record[i:int64]
0:[0;]
EOF
alias exists with different type
#0:record[i:int64]
0:[0;]
```

Maybe create the `error` alias in NewContext?

An underlying matter here (and one I wish I'd seen coming) is that `error` is no longer part of the ZNG spec but is still quasi reserved by the implementation. This doesn't feel like a great place to be. Should we reserve it in the spec?

PR #1564, opened 2020-11-05
Add Python zqd API client

### LINKED ISSUE #1519: python zqd api client

nwt:
Install this branch with `pip install 'git+https://github.com/brimsec/zq@python-zqd#subdirectory=python/zqd'`.

After merge, install master with `pip install 'git+https://github.com/brimsec/zq#subdirectory=python/zqd'`.

Closes #1519.

JupyterLab action shot:
image

### DIFF
--- a/python/zqd/zqd.py
+++ b/python/zqd/zqd.py
@@ -116,5 +116,5 @@ if __name__ == '__main__':
     args = parser.parse_args()
 
     c = Client(args.base_url)
-    for rec in c.search(args.space_name, args.zql):
-        pprint.pprint(rec)
+    for record in c.search(args.space_name, args.zql):
+        pprint.pprint(record)
--- a/python/zqd/zqd.py
+++ b/python/zqd/zqd.py
@@ -30,7 +30,8 @@ class Client():
             'space': self.spaces()[space_name]['id'],
         }
         params = {'format': 'zjson'}
-        r = self.session.post(self.base_url + "/search", json=body, params=params, stream=True)
+        r = self.session.post(self.base_url + "/search", json=body,
+                              params=params, stream=True)
         r.raise_for_status()
         for line in r.iter_lines():
             if line:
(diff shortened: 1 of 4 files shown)


mattnibs:
Can we get a test with this? I know it would be a pain, having said that it would be nice to know if an api change broke this.

nwt:
@mattnibs: Added a test in c47bdb9c4fca5f3f065b9753748197c2e9c9e068.

mccanne:
very slick!

python/zqd/zqd.py
```diff
@@ -0,0 +1,104 @@
...
+        r = self.session.post(self.base_url + "/ast", json={'zql': zql})
+        r.raise_for_status()
+        return r.json()
+
+    def search(self, space_name, zql):
+        aliases, schemas = {}, {}
+        for raw in self.search_raw(space_name, zql):
+            if raw['type'] == 'SearchRecords':
+                for rec in raw['records']:
+                    if 'aliases' in rec:
```
mccanne, line 27: It seems like the the process of parsing zqd search results into a stream of zjson JSON objects should be separate from the thing that parses a stream of zjson JSON objects into a stream of native python objects.  This way, for example you could have a python program that reads a file that has zjson in and parses the zjson into the python objects.  This should also help people reading the code to understand the model here more easily.

mccanne: This would also make tests easier to write as this subsystem evolves.

nwt: @mccanne: Is this what you had in mind?

```python
class Client():
    ...
    def search(self, space_name, zql):
        raw = self.search_raw(space_name, zql)
        zjson = take_zjson(raw)
        return decode_zjson(zjson)


def take_zjson(raw):
    for obj in raw:
        if obj['type'] == 'SearchRecords':
            for zjson in obj['records']:
                yield zjson


def decode_zjson(zjson):
    aliases, schemas = {}, {}
    for obj in zjson:
        if 'aliases' in obj:
            for a in obj['aliases']:
                aliases[a['name']] = a['type']
        if 'schema' in obj:
            schemas[obj['id']] = obj['schema']
        yield _to_native(aliases, schemas[obj['id']], obj['values'])
```

mattnibs (approved):
lgtm

python/zqd/zqd.py
mccanne: I think so.

nwt: Done in cf76c327ba99e196236e177c0f9a259c7e64c135.

alfred-landrum (approved):
I mentioned to noah on slack that we could add a note to the contributing file that we expect python3 to be available in the dev environment.

PR #1575, opened 2020-11-06
refactor expr.Function

### LINKED ISSUE #1524: Hook type context into expressions library package

mccanne:
This commit refactors expr.Function into a more modular approach,
changing the type from a function pointer to an interface whose
single method is Call().  This allows us to embed state in a function
that can be used to implement complex machinery, e.g., a cut()
function that behaves like a cut proc (caching lots of information
about types and builders) or a function to access an external
service like a key-value store (e.g., to do joins with thread intel
data feeds).

This change also allows us to clean up the temporary hack where
we passed the result buffer to a previously stateless function
in its Args parameters (Args is now replaced with []zng.Value).

The function names all now have a snake_case name in addition
to the package-style syntax.  The package syntax will be deprecated
in a subsequent commit.

We also changed the expression compiler to take a type context,
which will be used by functions and expression that need this,
which are coming soon.

Closes #1524.


expr/function/math.go
```diff
@@ -0,0 +1,219 @@
...
+func (a *abs) Call(args []zng.Value) (zng.Value, error) {
+	v := args[0]
+	id := v.Type.ID()
+	if zng.IsFloat(id) {
+		f, _ := zng.DecodeFloat64(v.Bytes)
+		f = math.Abs(f)
+		return zng.Value{zng.TypeFloat64, a.Float64(f)}, nil
+	}
+	if !zng.IsInteger(id) {
+		return badarg("Math.abs")
```
nwt, line 25: Is this the time to update errors to use the new names?

mccanne: Okay, I changed the ones that have new ones and left the ones that will be deleted or merged to a later PR.

nwt (approved):
:shipit:

PR #1574, opened 2020-11-06
check and convert alpha zng filestores on access

alfred-landrum:
When a filestore's data file is opened for queries, migrate it from alpha zng to beta zng if needed. This adds a field to the filestore's `info.json` file to record if the migration check is needed or not. A separate PR will add the ability to perform the check as a background routine in zqd, so that filestores can potentially be migrated before a user tries to access them.

As @philrz requested, I added log lines for migration start, success, and failure:
```
{"level":"info","ts":1604702663.6730611,"msg":"Alpha zng migration start","space_id":"sp_1jw5vTsJq5YfGZC5Z5Je2hT3J7A","file":"/Users/alfred/Library/Application Support/Brim/data/spaces/sp_1jw5vTsJq5YfGZC5Z5Je2hT3J7A/all.zng"}
{"level":"info","ts":1604702663.775091,"msg":"Alpha zng migration success","space_id":"sp_1jw5vTsJq5YfGZC5Z5Je2hT3J7A","file":"/Users/alfred/Library/Application Support/Brim/data/spaces/sp_1jw5vTsJq5YfGZC5Z5Je2hT3J7A/all.zng"}
```

PR #1576, opened 2020-11-07
convert alpha zng spaces in background

alfred-landrum:
When zqd starts, it will begin migrating any alpha zng filestores one at a time in the background. If a search request comes in for a space while its data is being migrated, the search will wait for the migration to complete.

Depends on https://github.com/brimsec/zq/pull/1574 .

### DIFF
--- a/ppl/zqd/space/space.go
+++ b/ppl/zqd/space/space.go
@@ -109,13 +109,13 @@ func (g *guard) acquireForDelete() error {
 	return nil
 }
 
-func loadSpaces(ctx context.Context, p iosrc.URI, conf config, logger *zap.Logger) ([]Space, error) {
+func (m *Manager) loadSpaces(ctx context.Context, p iosrc.URI, conf config) ([]Space, error) {
 	datapath := conf.DataURI
 	if datapath.IsZero() {
 		datapath = p
 	}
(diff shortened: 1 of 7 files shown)


mccanne (approved):
looks good

mattnibs (approved):

ppl/zqd/space/space.go
```diff
@@ -126,6 +126,7 @@ func loadSpaces(ctx context.Context, p iosrc.URI, conf config, logger *zap.Logge
 		if err != nil {
 			return nil, err
 		}
+		m.alphaFileMigrator.Add(store)
```
mattnibs, line 129: If we are going to migrations this way (which I think is a good idea), why not just get rid of the migration check in filestore.Open and have it run the search against alphazng? Sure the search will be slow but at least they'll be able to access their space while the migration happens.

alfred-landrum: I did consider this, but on balance decided against. Reading from the alpha zng files via the tzng bridge is quite slow, and the existing seek index mechanism won't work with them. 

With these & the changes in #1574, a Brim user who happens to run a search on a space will get the usual UI spinner, so it looks like their initial search is just taking a long time if they get to a space before the backgrounder gets to it.

PR #1577, opened 2020-11-08
move Cutter and ColumnBuilder out of proc

mccanne:
This commit moves the Cutter and ColumnBuilder out of the proc
package so they can be reused by expr, which will be needed by
the cut() function.

ColumnBuilder was moved to its own package under zng/builder since
this purely a zng function to construct zng record values.  The Cutter
type has logic to take a stream of zng records and create "cuts" of
those value so this more complicated machinery was moved to package
expr.

### DIFF
--- a/expr/cutter.go
+++ b/expr/cutter.go
@@ -1,18 +1,17 @@
-package cut
+package expr
 
 import (
-	"github.com/brimsec/zq/expr"
 	"github.com/brimsec/zq/field"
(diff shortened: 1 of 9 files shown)


mccanne:
If feels like we should unify zng/builder.go and zng/builder and revisit the name of ColumnBuilder (it is really a RecordBuilder) but this can wait till a later PR.

henridf (approved):

PR #1585, opened 2020-11-10
add cut function

mccanne:
This commit adds a cut function that behaves like the cut proc
but in the function context.  Like the proc, the cut function
ignores records that do not contain all the fields being cut.

### DIFF
--- a/expr/ztests/cut-union.yaml
+++ b/expr/ztests/cut-union.yaml
@@ -1,0 +1,17 @@
+zql: a=union(cut(x)),b=union(cut(x,s))
+
+input: |
+  #0:record[x:int32,s:string]
+  0:[1;a;]
+  0:[2;b;]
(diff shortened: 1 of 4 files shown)


expr/ztests/cut-union.yaml
```diff
@@ -0,0 +1,17 @@
+zql: a=union(cut(x)),b=union(cut(x,s))
```
nwt, line 1: I'm surprised this works without `put` in front of it. Should it?

nwt (approved):

nwt (approved):

expr/ztests/cut-union.yaml
mccanne: As we discussed, this is the aggregation rather than expression form.  More on this topic coming soon on the search/expr syntax PR.

PR #1586, opened 2020-11-10
zqd: run suricata-update at startup

### LINKED ISSUE #1552: Update Suricata rules
Suricata needs to be kept abreast of the latest threats. Typically these are kept up-to-date by running the `suricata-update` tool that comes with Suricata. When we have a bundled Suricata artifact with Brim, we'll surely want a way for this update to run periodically, hopefully with little/no effort on the user's part.

One possibility that comes to mind is to just run the update once whenever the app is launched, though if users use Brim so much that they keep it always open on the desktop, perhaps this would not be enough. Maybe some `cron`-like scheduler could do the update on a regular schedule. Some button/control in the app to trigger an immediate update might be yet another option. We might also want to think about a place in the app where we could provide minimally-disruptive informative messages saying the update has run.

henridf:
closes #1552


nwt (commented):
Why not do this in Brim where we can allow the user to trigger it, tell the user when it completes, and alert the user if it fails?

alfred-landrum:
In the future, we could let Brim trigger/request a rules update, obtain
info abou the current rules via zqd apis. Getting this auto-update in now
lays some groundwork for that, and ensures we have some mechanism to keep
the rules updated.

mattnibs:
@henridf, my preferred approach would be to have the suricata updater run by brim instead. Is the reason why zqd needs to do this because the update might be occurring while a pcap ingest is started, and this could cause something weird to happen?

alfred:
@henridf Can you make you double-check when you copy @'s from Slack to Github. Obviously this is an accident but just wanted you to be aware of it

ppl/zqd/pcapanalyzer/process.go
```diff
@@ -15,9 +15,9 @@ var ErrNotFound = errors.New("executable not found")
 
 // Process is an interface for interacting running with a running process.
 type ProcessWaiter interface {
-	// Wait waits for a running process to exit, returning any errors that
-	// occur.
-	Wait() error
+	// Wait waits for a running process to exit, returning the
+	// process' accumulated stdout and any errors that occur.
+	Wait() (string, error)
```
alfred-landrum, line 20: I'm somewhat nervous about grabbing stdout automatically. How about if we only store it if desired (like a flag to NewProcess), and add another method on ProcessWaiter that returns the captured stdout, or `""`?

henridf: Sounds good - are you nervous for memory reasons?

alfred-landrum: Yes; I don't want to accidentally read a large output to memory.

henridf: 88775ab

ppl/zqd/ingest/pcap.go
```diff
@@ -211,7 +211,8 @@ func (p *legacyPcapOp) runZeek(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	return zproc.Wait()
+	err = zproc.Wait()
+	return err
```
alfred-landrum, line 215:
```suggestion
	return zproc.Wait()
```

ppl/zqd/pcapanalyzer/process.go
```diff
@@ -28,27 +31,40 @@ func wrapError(err error, name, stderr string) error {
...
 func (p *Process) Start() error {
 	return wrapError(p.cmd.Start(), p.cmd.Path, p.stderrBuf.String())
 }
 
 func (p *Process) Wait() error {
 	return wrapError(p.cmd.Wait(), p.cmd.Path, p.stderrBuf.String())
 }
+
+func (p *Process) Stdout() string {
+	if p.stdoutBuf == nil {
```
alfred-landrum, line 66: Given the current setup in `NewProcess`, will this ever be true?

ppl/zqd/pcapanalyzer/process.go
```diff
@@ -28,27 +31,40 @@ func wrapError(err error, name, stderr string) error {
...
 
 func NewProcess(cmd *exec.Cmd) *Process {
-	p := &Process{cmd: cmd, stderrBuf: bytes.NewBuffer(nil)}
-	// Capture stderr for error reporting.
+	p := &Process{cmd: cmd, stderrBuf: bytes.NewBuffer(nil), stdoutBuf: bytes.NewBuffer(nil)}
 	cmd.Stderr = p.stderrBuf
 	return p
 }
 
+func NewProcessStdout(cmd *exec.Cmd) *Process {
```
alfred-landrum, line 51: How about a `stdout bool` parameter to `NewParameter`?

ppl/zqd/pcapanalyzer/process.go
henridf: oops

ppl/zqd/pcapanalyzer/process.go
henridf: Sure.

henridf: (I'm assuming you meant NewProcess).

alfred-landrum (approved):

PR #1588, opened 2020-11-11
Make zar import work with volatile records

### LINKED ISSUE #1579: zar import missing records
Using the ~4GB `wrccdc-year1` logs, I see a large discrepancy in records when I import it into zar:

```
$ zq -t "count()" ~/work/data/wrccdc-year1/zeek-logs-sort-r-ts.zng
#0:record[count:uint64]
0:[96217189;]
...
$ zar import -s 128MiB ~/work/data/wrccdc-year1/zeek-logs-sort-r-ts.zng
...
$ zar zq -t "count()"
#0:record[count:uint64]
0:[94839372;]
```

Additionally, if I try to manually verify against the data files in the resulting archive, I see this zq error:
```
$ find zarroot -name "d*.zng" | xargs zq -t "count()"
zarroot/zd/20170324/d-1k4BMtV8RQt4HzJxdVMFhkJv3DA.zng: _path (string): expected primitive type, got container
```

I wonder if this is related to the fact that some records have the same timestamp, as I saw a similar problem when I was working on the overlap support. (Though note the above runs don't perform any compaction).

mattnibs:
Fix issue where zar import would end up with corrupted records when
importing data from zng data sources.

Part of #1579

### DIFF
--- a/ppl/archive/import.go
+++ b/ppl/archive/import.go
@@ -192,7 +192,7 @@ func (dw *tsDirWriter) writeOne(rec *zng.Record) error {
 	// and slow down import. We should instead copy the raw record bytes into a
 	// recycled buffer and keep around an array of ts + byte-slice structs for
 	// sorting.
-	rec = rec.Keep()
+	rec.CopyBody()
 	dw.records = append(dw.records, rec)
 	dw.addBufSize(int64(len(rec.Raw)))
 	if dw.chunkSizeEstimate() > dw.ark.LogSizeThreshold {
--- a/ppl/archive/import.go
+++ b/ppl/archive/import.go
@@ -188,6 +188,11 @@ func (dw *tsDirWriter) chunkSizeEstimate() int64 {
 }
 
 func (dw *tsDirWriter) writeOne(rec *zng.Record) error {
+	// XXX This call leads to a ton of one-off allocations that burden the GC
+	// and slow down import. We should instead copy the raw record bytes into a
+	// recycled buffer and keep around an array of ts + byte-slice structs for
+	// sorting.
+	rec = rec.Keep()
 	dw.records = append(dw.records, rec)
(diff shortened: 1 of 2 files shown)


alfred-landrum:
I confirmed this fixed the zq error I saw in #1579, but this doesn't fix the record count discrepancy there. @mattnibs can you unlink the issues so this doesn't close #1579 when you merge this?

alfred-landrum (approved):

ppl/archive/import.go
```diff
@@ -188,6 +188,11 @@ func (dw *tsDirWriter) chunkSizeEstimate() int64 {
 }
 
 func (dw *tsDirWriter) writeOne(rec *zng.Record) error {
+	// XXX This call leads to a ton of one-off allocations that burden the GC
+	// and slow down import. We should instead copy the raw record bytes into a
+	// recycled buffer and keep around an array of ts + byte-slice structs for
+	// sorting.
+	rec = rec.Keep()
```
nwt, line 195:
```suggestion
	rec.CopyBody()
```

zio/zngio/reader.go
```diff
@@ -121,8 +121,8 @@ func (r *Reader) reset() {
 	r.sos = r.peekerOffset
 }
 
-var startBatch = errors.New("start of uncmompressed batch")
-var endBatch = errors.New("end of uncmmpressed batch encountered while parsing data")
+var startBatch = errors.New("start of uncompressed batch")
+var endBatch = errors.New("end of uncompressed batch encountered while parsing data")
```
nwt, line 125: Nit: Do unrelated fixes like this in a separate PR.

nwt (approved):

zio/zngio/reader.go
mattnibs: I certainly get the principle here.

PR #1593, opened 2020-11-12
rework merge/combiner

mccanne:
This commit renames the orderedmerge proc as merge and
the old merge proc as combiner.  Toward this end, we
moved the merge logic into zbuf and changed the merge
proc to use it.  We also moved the scanner into zbuf.

This new arrangement allows Scanners to be created and
merged into a single puller without creating an
unnecessary proc object.  When we change the zar commands
to take pullers instead of readers (see #1614), we can pass
a zbuf.Merger in as the puller and this will prevent an
extra copy of zng.Records to unbatch them through the
Read interface method.

The logic in zbuf.Merger could be improved if Batch.Records()
were changed to return a slice of zng.Record instead of
*zng.Record.  This is described in issue #1613.


zbuf/scanner.go
```diff
@@ -46,10 +45,34 @@ func (s *ScannerStats) Accumulate(ss *ScannerStats) {
...
+		s, err := NewScanner(ctx, reader, nil, nano.MaxSpan)
+		if err != nil {
+			return nil, err
+		}
+		scanners = append(scanners, s)
+	}
+	return scanners, nil
+}
+
+func ReadersToPullers(ctx context.Context, readers []Reader) ([]Puller, error) {
```
alfred-landrum, line 60: Is the main reason to use Scanners to perform this operation, instead of say, zbuf.ReadBatch, so that the improved memory handling of the zngio scanner is used?

mccanne: merge.Proc needs pullers as its parent so this provides the glue when you are directly wiring readers to a merge proc (e.g., file readers that do not make native Scanners like zngio).  This was the minimal change I wanted to make to get this wired up.  We could look into further refactoring.  The other thing I noticed is that stuff that takes a Reader (like zar commands) have to turn the Scanner into a Reader, which forces the copy of Record out of the scanner buffer.  If we changed zar commands to take Pullers instead of Readers then they could spin over the Batch and release it without copying the record out of the Scanner buffer.

mccanne: As suggested by @mattnibs, I'll see about moving the join function into expr to be used separately by zbuf and proc.

nwt:
The general direction looks good to me.

zbuf/scanner.go
mccanne: I ended up moving merger to zbuf instead of expr because it's not an expression and implements zbuf.Puller.

ppl/archive/multisource.go
```diff
@@ -54,21 +55,19 @@ func newSpanScanner(ctx context.Context, ark *Archive, zctx *resolver.Context, s
...
+		pullers = append(pullers, scanner)
 	}
-	return &scannerCloser{
-		Scanner: scn,
-		Closer:  closers,
+	return &pullerCloser{
+		Puller:     zbuf.MergeByTs(ctx, pullers, ark.DataOrder),
+		MultiStats: scanners,
+		Closer:     closers,
 	}, stats, nil
```
mattnibs, line 71: Cool!

zbuf/merger.go
```diff
@@ -0,0 +1,192 @@
...
+		m.pullers = append(m.pullers, &puller{Puller: p, ch: make(chan batch)})
+	}
+	return m
+}
+
+func (m *Merger) Cancel() {
+	if m.cancelCh != nil {
+		close(m.cancelCh)
+		m.cancelCh = nil
+	}
```
mattnibs, line 63: Feels like this should be safe for concurrent access. Why not replace the cancelCh with a context.CancelFunc derived from the passed context? Ala on line 47

```golang
func NewMerger(ctx context.Context, pullers []Puller, cmp expr.CompareFn) *Merger {
        ctx, cancel := context.WithCancel(ctx)
 	m := &Merger{
 		ctx:      ctx,
 		cmp:      cmp,
 		cancel: cancel,
 	}
```

zbuf/merger.go
```diff
@@ -0,0 +1,192 @@
...
+					m.pullers[k] = nil
+					continue
+				}
+				// We're keeping records owned by res.Batch so don't call Unref.
+				// XXX this means the batch won't be returned to
+				// the pool and instead will run through GC.
+				p.recs = b.Batch.Records()
+				p.batch = b.Batch
+			case <-m.ctx.Done():
+				return -1, m.ctx.Err()
```
mattnibs, line 151: If leaving in cancelCh, don't we need a case for it here as well?

zbuf/merger.go
mattnibs: Either above or do this:
```suggestion
	select {
	case <-m.cancelCh:
	default:
		close(m.cancelCh)
	}
```

zbuf/merger.go
```diff
@@ -0,0 +1,192 @@
...
+	pullers, err := ReadersToPullers(ctx, readers)
+	if err != nil {
+		return nil, err
+	}
+	return MergeByTs(ctx, pullers, order), nil
+}
+
+func MergeByTs(ctx context.Context, pullers []Puller, order Order) *Merger {
+	reversed := bool(order)
+	cmp := NewCompareFn(field.New("ts"), reversed)
```
mattnibs, line 83: Not to address here, but perhaps `expr` is a better place for `Order`?

zbuf/merger.go
```diff
@@ -0,0 +1,192 @@
...
+	p.batch = nil
+	return rec
+}
+
+func (m *Merger) findLeader() (int, error) {
+	leader := -1
+	for k, p := range m.pullers {
+		if p == nil {
+			continue
+		}
```
mattnibs, line 133: Perhaps this is where you'll tell me that computers are fast, but considering this is in the critical path wouldn't it be better to use a heap here to manage finding the leader? Like what is done in package `spill`?

zbuf/merger.go
```diff
@@ -0,0 +1,192 @@
...
+		m.Cancel()
+		return nil, err
+	}
+	if !m.overlaps(leader) {
+		b := m.pullers[leader].batch
+		m.pullers[leader].recs = nil
+		return b, nil
+	}
+	const batchLen = 100 // XXX
+	return ReadBatch(m, batchLen)
```
mattnibs, line 191: Slick

zbuf/merger.go
mccanne: I changed it to use cancel().

zbuf/merger.go
mccanne: Yeah, I think for typical fan-ins the for-loop will be faster than a heap but when fan-ins get large the heap will work better.  The reason spill uses a heap is because you can have really large sets of files to merge.  I would make the change after we do some perf analysis and the logic here could make a run-time decision for which to approach use.

proc/combine/combine.go
```diff
@@ -0,0 +1,90 @@
...
+	doneCh   chan struct{}
+	parents  []*runnerProc
+	nparents int
+}
+
+type runnerProc struct {
+	pctx   *proc.Context
+	parent proc.Interface
+	ch     chan<- proc.Result
+	doneCh <-chan struct{}
```
nwt, line 24: Nit: You could simplify a tiny bit if you replaced pctx and doneCh with a context.Context created by context.WithCancel.

nwt (approved):

mattnibs (approved):

filter/compare.go
```diff
@@ -106,12 +106,12 @@ func CompareInt64(op string, pattern int64) (Predicate, error) {
...
 			ts, err := zng.DecodeTime(zv)
 			if err == nil {
-				return CompareInt(int64(ts), pattern*1e9)
+				return CompareInt(int64(ts), pattern*1_000_000_000)
 			}
 		case zng.IdDuration:
 			v, err := zng.DecodeInt(zv)
 			if err == nil {
-				return CompareInt(int64(v), pattern*1e9)
+				return CompareInt(int64(v), pattern*1_000_000_000)
```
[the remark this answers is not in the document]
mccanne: Yes, for sure.  This one got by me.  I'll undo it and I created issue #1622.

proc/combine/combine.go
mccanne: This was the same change Nibs suggested in the previous orderedmerge code I moved to zbuf, so yes, good idea, I will change it here too (as this was an unmodified copy except for name from the old merge proc).

PR #1602, opened 2020-11-13
Fix seekindex lookups for descending order indices

### LINKED ISSUE #1579: zar import missing records

mattnibs:
Fix an off-by-one error one descending seek indexes that was causing missed
records in certain circumstances.

Additionally: The way the seek index was written, made it difficult to isolate
issues. As such I moved seekindex into its own package that includes both
builder and reader functionality.

Closes #1579

### DIFF
--- a/ppl/archive/import.go
+++ b/ppl/archive/import.go
@@ -267,10 +267,10 @@ type chunkWriter struct {
 	dataFileWriter *zngio.Writer
 	firstTs        nano.Ts
 	id             ksuid.KSUID
-	seekIndex      *seekindex.Builder
 	lastTs         nano.Ts
 	masks          []ksuid.KSUID
 	needIndexWrite bool
+	seekIndex      *seekindex.Builder
 	tsd            tsDir
 	wroteFirst     bool
 }
--- a/ppl/archive/import.go
+++ b/ppl/archive/import.go
@@ -3,16 +3,13 @@ package archive
 import (
 	"context"
 	"io"
-	"io/ioutil"
-	"os"
 	"sync/atomic"
 
 	"github.com/brimsec/zq/expr"
-	"github.com/brimsec/zq/microindex"
 	"github.com/brimsec/zq/pkg/bufwriter"
-	"github.com/brimsec/zq/pkg/fs"
 	"github.com/brimsec/zq/pkg/iosrc"
 	"github.com/brimsec/zq/pkg/nano"
+	"github.com/brimsec/zq/ppl/archive/seekindex"
 	"github.com/brimsec/zq/proc/sort"
 	"github.com/brimsec/zq/proc/spill"
 	"github.com/brimsec/zq/zbuf"
@@ -264,20 +261,18 @@ func (dw *tsDirWriter) flush() error {
 // chunkWriter is a zbuf.Writer that writes a stream of sorted records into an
 // archive chunk file.
 type chunkWriter struct {
-	ark             *Archive
-	byteCounter     *writeCounter
-	count           uint64
-	dataFileWriter  *zngio.Writer
-	firstTs         nano.Ts
-	id              ksuid.KSUID
-	indexBuilder    *zng.Builder
-	indexTempPath   string
-	indexTempWriter *zngio.Writer
-	lastTs          nano.Ts
-	masks           []ksuid.KSUID
-	needIndexWrite  bool
-	tsd             tsDir
-	wroteFirst      bool
+	ark            *Archive
+	byteCounter    *writeCounter
+	count          uint64
+	dataFileWriter *zngio.Writer
+	firstTs        nano.Ts
+	id             ksuid.KSUID
+	seekIndex      *seekindex.Builder
+	lastTs         nano.Ts
+	masks          []ksuid.KSUID
+	needIndexWrite bool
+	tsd            tsDir
+	wroteFirst     bool
 }
 
 func newChunkWriter(ctx context.Context, ark *Archive, tsd tsDir, masks []ksuid.KSUID) (*chunkWriter, error) {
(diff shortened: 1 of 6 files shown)


ppl/archive/import.go
```diff
@@ -377,38 +359,17 @@ func (cw *chunkWriter) closeWithTs(ctx context.Context, firstTs, lastTs nano.Ts)
...
-		microindex.FrameThresh(framesize),
-		microindex.Order(cw.ark.DataOrder),
-	)
-	if err != nil {
-		return Chunk{}, err
-	}
-	if err = zbuf.CopyWithContext(ctx, idxWriter, tfr); err != nil {
-		idxWriter.Abort()
+	if err := cw.seekIndex.Close(); err != nil {
+		cw.seekIndex.Abort()
```
alfred-landrum, line 366: I see that the reason to call an Abort after a failed Close is that it will try to remove the partially written index file in Abort(), which we would want to do, otherwise we need to add logic to ignore a failed index read (since a partially written index file could cause reads of its data file to fail).

For simplicity sake, how about moving the close error check & Abort call inside the index builder's Close call?

This makes me think that microindex.Writer should use an iosrc.Replacer instead of an iosrc.Writer, so that the index writer can call replacer.Abort() - on S3, that will cancel the multipart upload & ensure that there's no partially written index file.

ppl/archive/seekindex/builder.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/brimsec/zq/microindex"
+	"github.com/brimsec/zq/pkg/nano"
+	"github.com/brimsec/zq/zbuf"
+	"github.com/brimsec/zq/zng"
+	"github.com/brimsec/zq/zng/resolver"
+)
+
+type Builder struct {
+	builder *zng.Builder
+	path    string
```
alfred-landrum, line 15: I don't see path used anywhere?

ppl/archive/seekindex/seekindex_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/brimsec/zq/pkg/iosrc"
+	"github.com/brimsec/zq/pkg/nano"
+	"github.com/brimsec/zq/zbuf"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAscending(t *testing.T) {
+	var entries = []entry{
+		{1490469784771872000, 0},
```
alfred-landrum, line 20: Would you change the timestamps here so that they're easier to read at a glance, as in TestDescending?

ppl/archive/seekindex/seekindex_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+func (t *testSeekIndex) Lookup(span nano.Span, expected Range) {
+	rg, err := t.SeekIndex.Lookup(context.Background(), span)
+	require.NoError(t, err)
+	assert.Equal(t, expected, rg)
+}
+
+func newTestSeekIndex(t *testing.T, entries []entry) *testSeekIndex {
+	path := build(t, entries)
+	s, err := Open(context.Background(), iosrc.MustParseURI(path))
```
alfred-landrum, line 85: I don't see a paired Close() for this open call.

ppl/archive/seekindex/builder.go
mattnibs: Nope, removing it.

ppl/archive/import.go
mattnibs: Sure

ppl/archive/seekindex/seekindex_test.go
```diff
@@ -0,0 +1,104 @@
...
+		{500, 904528},
+		{400, 1139588},
+		{300, 1355498},
+		{200, 1564211},
+		{100, 1776965},
+	}
+	s := newTestSeekIndex(t, entries)
+	s.Lookup(nano.Span{Ts: 900, Dur: 1}, Range{0, 215367})
+	s.Lookup(nano.Span{Ts: 700, Dur: 1}, Range{438514, 680477})
+	s.Lookup(nano.Span{Ts: 750, Dur: 100}, Range{0, 438514})
```
alfred-landrum, line 53: Why isn't this `Range(215367, 438514)`?

ppl/archive/import.go
```diff
@@ -264,20 +261,18 @@ func (dw *tsDirWriter) flush() error {
...
-	needIndexWrite  bool
-	tsd             tsDir
-	wroteFirst      bool
+	ark            *Archive
+	byteCounter    *writeCounter
+	count          uint64
+	dataFileWriter *zngio.Writer
+	firstTs        nano.Ts
+	id             ksuid.KSUID
+	seekIndex      *seekindex.Builder
```
nwt, line 270: Nit: Put this between needIndexWrite and tsd (and likewise in newChunkWriter).

alfred-landrum (approved):

ppl/archive/seekindex/builder.go
```diff
@@ -0,0 +1,48 @@
...
+
+func NewBuilder(ctx context.Context, path string, order zbuf.Order) (*Builder, error) {
+	zctx := resolver.NewContext()
+	writer, err := microindex.NewWriterWithContext(ctx, zctx, path, microindex.Order(order), microindex.Keys("ts"))
+	if err != nil {
+		return nil, err
+	}
+	builder := zng.NewBuilder(zctx.MustLookupTypeRecord(Schema))
+	return &Builder{
+		builder: builder,
```
nwt, line 26:
```suggestion
	return &Builder{
		builder: zng.NewBuilder(zctx.MustLookupTypeRecord(Schema)),
```

ppl/archive/seekindex/builder.go
```diff
@@ -0,0 +1,48 @@
...
+	rec := b.builder.Build(zng.EncodeTime(ts), zng.EncodeInt(offset))
+	return b.writer.Write(rec)
+}
+
+func (b *Builder) Abort() error {
+	return b.writer.Abort()
+}
+
+// Close closes the underlying microindex.Writer. Should an error occur the
+// microindex will be deleted via a call to abort.
```
nwt, line 41:
```suggestion
// microindex will be deleted via a call to Abort.
```

ppl/archive/seekindex/seekindex.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"github.com/brimsec/zq/microindex"
+	"github.com/brimsec/zq/pkg/iosrc"
+	"github.com/brimsec/zq/pkg/nano"
+	"github.com/brimsec/zq/zbuf"
+	"github.com/brimsec/zq/zng"
+	"github.com/brimsec/zq/zng/resolver"
+)
+
+var Schema = []zng.Column{
```
nwt, line 15: Maybe declare this in builder.go since that's where it's used? And maybe make it private?

nwt (approved):

ppl/archive/seekindex/seekindex.go
mattnibs: Sure

Issue #682, opened 2020-04-28
boyer-moore string search


alfred-landrum:
We should be able to implement boyer moore string searching, optimized to run against the raw zbytes of a zng stream, so that we minimize the amount of record parsing and and io buffer allocation. There's a connection with implementing zng compression support, as a compressed batch of records gives an easy way to identify record boundaries.

philrz:
Note to self: When this Epic is closed, it's probably worth cutting another set of the zq "perf" numbers so we can quantify the observable benefits from this.

PR #1064, opened 2020-08-05
Update perf numbers to be current & include compressed ZNG

philrz:
After #1050 was merged, https://github.com/brimsec/zq-sample-data/pull/13 was then merged to add an uncompressed ZNG variation to the [zq-sample-data](https://github.com/brimsec/zq-sample-data). This PR takes it to the next step by adding uncompressed ZNG to the permutations that are tested via `make perf-compare`. The performance README is also updated to include current outputs run through current `zq` (commit `253c336` at the moment).

While the numbers in the README are based on a Google Cloud VM for consistency, I also did a local variation on my Mac where I had the script do a second run of each `zq` command line with its output piped to `wc -c` so I could confirm that the byte counts reflected the expected changes without/with compression. It looks to me like everything checks out, i.e. compressed ZNG coming in at ~2.2x compression ratio compared to uncompressed ZNG. A sampling:

|**<br>Tool**|**<br>Arguments**|**Input<br>Format**|**Output<br>Format**|**<br>Real**|**<br>User**|**<br>Sys**|
|:----------:|:---------------:|:-----------------:|:------------------:|-----------:|-----------:|----------:|
|`zq`|`*`|zeek|zeek|9.81|27.45|1.14|
Bytes output = 319881573
|`zq`|`*`|zeek|zng|6.36|11.32|0.60|
Bytes output = 78336560
|`zq`|`*`|zeek|zng-uncompressed|5.77|10.03|0.52|
Bytes output = 175948548
|`zq`|`*`|zeek|tzng|7.17|19.09|0.65|
Bytes output = 244134486
|`zq`|`*`|zeek|ndjson|44.28|74.41|2.74|
Bytes output = 678951472
|`zq`|`*`|zng|zeek|9.65|14.32|0.46|
Bytes output = 320006251
|`zq`|`*`|zng|zng|2.57|3.52|0.16|
Bytes output = 78336464
|`zq`|`*`|zng|zng-uncompressed|2.51|2.81|0.17|
Bytes output = 175948548
|`zq`|`*`|zng|tzng|7.18|9.57|0.30|
Bytes output = 244134488
|`zq`|`*`|zng|ndjson|41.18|49.44|0.91|
Bytes output = 678951472
|`zq`|`*`|zng-uncompressed|zeek|9.35|14.17|0.46|
Bytes output = 320006251
|`zq`|`*`|zng-uncompressed|zng|2.40|3.30|0.15|
Bytes output = 78336464
[...]

PR #1605, opened 2020-11-13
Performance update v0.23.0

philrz:
Upon the completion of #682, it seems appropriate to publish another point-in-time snapshot of `zq` performance. Overall the numbers have improved modestly since the last time we logged numbers.

### DIFF
--- a/performance/README.md
+++ b/performance/README.md
@@ -17,160 +17,161 @@ As there are many results to sift through, here's a few key summary take-aways:
 
 # Results
 
+The results below reflect performance as of `zq` release `v0.23.0`.
+
 ### Output all events unmodified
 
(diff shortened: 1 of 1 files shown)


alfred-landrum (approved):
I notice particular improvement for the ndjson input tests.

philrz:
@alfred-landrum: Indeed. This may be making up for lost ground. Back in #1064 we flagged that the "NDJSON read" performance had taken a significant hit. We're not back to the best perf we ever had in that area, but it's good to see it trending better.

Anyway, I'm not too concerned. The current active perf work being more of the "scale out" parallel variety seems like a good investment, and I'm confident that when we want to do single-node perf improvement, we'll find gains. No doubt when "generic ingest" has us pushing lots more people toward ingesting JSON it'll inspire us to take it on.

PR #1606, opened 2020-11-13
Do not return empty batches from filter.Proc.Pull

nwt:


mattnibs (approved):

proc/filter/filter.go
```diff
@@ -20,20 +20,24 @@ func New(parent proc.Interface, f filter.Filter) *Proc {
...
-	for k := 0; k < batch.Length(); k++ {
-		r := batch.Index(k)
-		if f.Filter(r) {
-			out = append(out, r.Keep())
+	for {
+		batch, err := f.parent.Pull()
+		if proc.EOS(batch, err) {
+			return nil, err
+		}
+		// Now we'll a new batch with the (sub)set of reords that match.
```
henridf, line 28:
```suggestion
		// Now we'll a build new batch with the (sub)set of reords that match.
```

PR #1609, opened 2020-11-13
deprecate package-style function names

### LINKED ISSUE #1379: drop package names in functions
Currently, the list of function names supported in expr/functions.go uses a package-style naming convention, i.e., `pkgname.func()`.  Until and unless we support a package import concept in zql this seems clunky and like overkill, and it's burdensome to have to type the package name.

Perhaps this is a bikeshed topic, but the proposal here is to drop the package name prefix from our library of function names, e.g., `Strings.tolower()` becomes `tolower()`, `Math.log()` becomes `log()` etc.

mccanne:
This commit deprecates package-style function names by raising
an error with a helpful message for the alternative use.
We also eliminated some of the string functions in favor
of a type cast style, and extended the semantics of
numeric casts to accept strings (and parse the string
to a numeric value).  We also got rid of several of the
time conversion functions in favor of casts and adopted
the convention that a numeric value cast to time is in
units of nanosecond and added helper functions to convert
sec, msec, and usec to nanoseconds.

Closes #1379

### DIFF
--- a/expr/function/function.go
+++ b/expr/function/function.go
@@ -162,7 +162,7 @@ func (l *lenFn) Call(args []zng.Value) (zng.Value, error) {
 		}
 		return zng.Value{zng.TypeInt64, l.Int(int64(len))}, nil
 	case *zng.TypeOfString, *zng.TypeOfBstring:
-		v := len(string(args[0].Bytes))
+		v := len(args[0].Bytes)
 		return zng.Value{zng.TypeInt64, l.Int(int64(v))}, nil
 	default:
 		return badarg("len")
--- a/expr/function/function.go
+++ b/expr/function/function.go
@@ -20,7 +20,50 @@ type Interface interface {
 	Call([]zng.Value) (zng.Value, error)
 }
 
+var deprecated = map[string]string{
+	"Math.abs":              "abs",
+	"Math.ceil":             "ceil",
+	"Math.floor":            "floor",
+	"Math.log":              "log",
+	"Math.max":              "max",
+	"Math.min":              "min",
+	"Math.mod":              "mod",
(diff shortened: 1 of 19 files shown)


expr/function/function.go
```diff
@@ -20,7 +20,50 @@ type Interface interface {
...
+	"Time.fromNanoseconds":  "type cast, e.g., <nsec-value>:time",
+	"Time.trunc":            "trunc",
+	"toBase64":              "to_base64",
+	"fromBase64":            "from_base64",
+}
+
+func isDeprecated(name string) error {
+	msg, ok := deprecated[name]
+	if ok {
+		return fmt.Errorf("function is deprecated: use %s", msg)
```
nwt, line 58:
```suggestion
		return fmt.Errorf("function %s is deprecated: use %s", name, msg)
```

expr/function/function.go
```diff
@@ -138,6 +161,9 @@ func (l *lenFn) Call(args []zng.Value) (zng.Value, error) {
 			return zng.Value{}, err
 		}
 		return zng.Value{zng.TypeInt64, l.Int(int64(len))}, nil
+	case *zng.TypeOfString, *zng.TypeOfBstring:
+		v := len(string(args[0].Bytes))
```
nwt, line 165:
```suggestion
		v := len(args[0].Bytes)
```

nwt (approved):

expr/function/function.go
mccanne: I had it this way to begin with but the caller prints the function name as part of its error message so the name appeared twice.  I suppose I could change the caller and the other error condition here.  Maybe I'll leave it for this merge and we can change it later if warranted.

PR #1615, opened 2020-11-14
add support for aggregation-less group by

### LINKED ISSUE #821: Add syntax for keys-only/reducer-less groupby to ZQL
There should be a way to express a "reducer-less" groupby in ZQL, that would compute a group-by table of all the unique values for its keys, without any reducer. (With identical results to `count() by <keys> | cut <keys>`).

Proposed syntax:

`... | group by <keys> | * `

Note that this is a grammar-only change; the runtime already supports this, and can already handle a reducer-less groupby AST (as used in #816) :

```
			&ast.GroupByProc{
				Keys: []string{keyName},
			},
```

mccanne:
This commit changes the peg grammar to allow group-by expressions
of the form "by key-expr [, key-expr]", e.g., group-bys without
any aggregations.  No changes were needed besides the grammar
change since it turned out that all the aggregation code worked
fine with zero or more aggregations even though we only ever used
it with one or more aggregations to date.

Closes #821

### DIFF
--- a/proc/groupby/ztests/by-only.yaml
+++ b/proc/groupby/ztests/by-only.yaml
@@ -1,0 +1,19 @@
+zql: 'by s | sort s'
+
+input: |
+  #0:record[x:int32,s:string]
+  0:[1;b;]
+  0:[2;f;]
+  0:[1;b;]
+  0:[1;a;]
+  0:[1;b;]
+  0:[2;c;]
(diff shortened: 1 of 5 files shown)


henridf (approved):
I'd used it without aggregations in the past in AST-only form (for zar indexes I think).

PR #1618, opened 2020-11-15
remove unneeded match-all filter at front of flow graph

### LINKED ISSUE #1620: rework AST so parallel proc may appear as first element in flowgraph
There is a bunch of code in zq to deal with the fact that the peg parser tacks on a "filter *" for any zql query that does not include a search.  This means one cannot differentiate between the user typing
```
filter * | ( p1 ; p2 ) | ...
```
 and 
```
(p1 ; p2) | ...
```
since the two ASTs are identical.  This convention should be removed so we can add the parallel input option to zq to wire a set of files straight into the first flowgraph without the heuristic of removing the filter *.  To do this, we need to fix some tests and code assumptions that this tacked-on filter proc is always there.

mccanne:
This commits changes the zql parser so it doesn't insert a filter proc
with a match-all op at the front of flowgraphs that do not have a filter.
We used to rely on this but that seems fragile and we only needed to
make a couple changes here to fix things.

One change was to add logic to the worker clustering requests so that
a filter with matchall is added.  A better fix would be to change the
search endpoint to handle a null Proc and create a scanner for the
given spans instead of panic.  The panic arose from a null Proc
causing a null pointer deref in ast.Unpack.  We fixed this with checks
that return an error instead of panic stacktrace.

This change is needed so that we can add a parallel input flag to zq,
feeding a parallel zql query as the inserted `"filter *"` gets in the way
and we need to know the difference between the user typing `"* | ..."`
vs `"..."`.  This parallel input flag will be in turn needed by the
soon-coming merge join proc.

Closes #1620

### DIFF
--- a/driver/parallel.go
+++ b/driver/parallel.go
@@ -187,11 +187,6 @@ func (pg *parallelGroup) sourceToRequest(src Source) (*api.WorkerRequest, error)
 			return nil, err
 		}
 		req.Proc = b
-	} else {
-		// If there's no filter, then send a match-all proc since the
-		// client can't handle a null req.Proc.  XXX It seems like a
-		// null proc should mean return everything in the span.
-		req.Proc = json.RawMessage(`{"op":"FilterProc","filter":{"op":"MatchAll"}}`)
 	}
 	req.Dir = pg.mcfg.Order.Int()
 	return &req, nil
--- a/driver/parallel.go
+++ b/driver/parallel.go
@@ -187,6 +187,11 @@ func (pg *parallelGroup) sourceToRequest(src Source) (*api.WorkerRequest, error)
 			return nil, err
 		}
 		req.Proc = b
+	} else {
+		// If there's no filter, then send a match-all proc since the
+		// client can't handle a null req.Proc.  XXX It seems like a
+		// null proc should mean return everything in the span.
+		req.Proc = json.RawMessage(`{"op":"FilterProc","filter":{"op":"MatchAll"}}`)
 	}
 	req.Dir = pg.mcfg.Order.Int()
 	return &req, nil
--- a/api/api.go
+++ b/api/api.go
@@ -43,7 +43,7 @@ type TaskEnd struct {
 
 type SearchRequest struct {
 	Space SpaceID         `json:"space" validate:"required"`
-	Proc  json.RawMessage `json:"proc" validate:"required"`
+	Proc  json.RawMessage `json:"proc,omitempty"`
 	Span  nano.Span       `json:"span"`
 	Dir   int             `json:"dir" validate:"required"`
 }
--- a/ast/unpack.go
+++ b/ast/unpack.go
@@ -14,6 +14,9 @@ type Unpacker interface {
 }
 
 func unpackProcs(custom Unpacker, node joe.Interface) ([]Proc, error) {
+	if node == nil {
+		return nil, errors.New("ast.unpackProcs: procs field is missing")
+	}
 	procList, err := node.Get("procs")
 	if err != nil {
 		return nil, fmt.Errorf("ast.unpackProcs: procs field is missing")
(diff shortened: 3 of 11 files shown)


driver/parallel.go
```diff
@@ -187,6 +187,11 @@ func (pg *parallelGroup) sourceToRequest(src Source) (*api.WorkerRequest, error)
 			return nil, err
 		}
 		req.Proc = b
+	} else {
+		// If there's no filter, then send a match-all proc since the
+		// client can't handle a null req.Proc.  XXX It seems like a
+		// null proc should mean return everything in the span.
+		req.Proc = json.RawMessage(`{"op":"FilterProc","filter":{"op":"MatchAll"}}`)
```
nwt, line 194: How about allowing ast.SearchRequest.Proc to be null or unset?
```diff
diff --git a/api/api.go b/api/api.go
index 458f88c5..74c06cd6 100644
--- a/api/api.go
+++ b/api/api.go
@@ -43,7 +43,7 @@ type TaskEnd struct {
 
 type SearchRequest struct {
 	Space SpaceID         `json:"space" validate:"required"`
-	Proc  json.RawMessage `json:"proc" validate:"required"`
+	Proc  json.RawMessage `json:"proc,omitempty"`
 	Span  nano.Span       `json:"span"`
 	Dir   int             `json:"dir" validate:"required"`
 }
diff --git a/ast/unpack.go b/ast/unpack.go
index 17197a5b..06198ed3 100644
--- a/ast/unpack.go
+++ b/ast/unpack.go
@@ -421,10 +421,16 @@ func UnpackMap(custom Unpacker, m interface{}) (Proc, error) {
 
 // UnpackJSON transforms a JSON representation of a proc into an ast.Proc.
 func UnpackJSON(custom Unpacker, buf []byte) (Proc, error) {
+	if len(buf) == 0 {
+		return nil, nil
+	}
 	obj, err := joe.Unmarshal(buf)
 	if err != nil {
 		return nil, err
 	}
+	if obj == nil {
+		return nil, nil
+	}
 	proc, err := unpackProc(custom, obj)
 	if err != nil {
 		return nil, err
diff --git a/driver/compile.go b/driver/compile.go
index 38e8eb12..a3381823 100644
--- a/driver/compile.go
+++ b/driver/compile.go
@@ -43,7 +43,12 @@ func zbufDirInt(reversed bool) int {
 	return 1
 }
 
+var passProc = &ast.PassProc{Node: ast.Node{"PassProc"}}
+
 func programPrep(program ast.Proc, sortKey field.Static, sortReversed bool) (ast.BooleanExpr, ast.Proc) {
+	if program == nil {
+		return nil, passProc
+	}
 	ReplaceGroupByProcDurationWithKey(program)
 	if sortKey != nil {
 		setGroupByProcInputSortDir(program, sortKey, zbufDirInt(sortReversed))
@@ -177,8 +182,6 @@ func ensureSequentialProc(p ast.Proc) *ast.SequentialProc {
 	}
 }
 
-var passProc = &ast.PassProc{Node: ast.Node{"PassProc"}}
-
 // liftFilter removes the filter at the head of the flowgraph AST, if
 // one is present, and returns its ast.BooleanExpr and the modified
 // flowgraph AST. If the flowgraph does not start with a filter, it
diff --git a/driver/parallel.go b/driver/parallel.go
index ad59a3db..01e7a7a8 100644
--- a/driver/parallel.go
+++ b/driver/parallel.go
@@ -188,11 +188,6 @@ func (pg *parallelGroup) sourceToRequest(src Source) (*api.WorkerRequest, error)
 			return nil, err
 		}
 		req.Proc = b
-	} else {
-		// If there's no filter, then send a match-all proc since the
-		// client can't handle a null req.Proc.  XXX It seems like a
-		// null proc should mean return everything in the span.
-		req.Proc = json.RawMessage(`{"op":"FilterProc","filter":{"op":"MatchAll"}}`)
 	}
 	req.Dir = pg.mcfg.Order.Int()
 	return &req, nil
```

mccanne: cool

nwt (approved):
:shipit:

PR #1625, opened 2020-11-16
zqd auto compaction

### LINKED ISSUE #1555: compact imported data
Now that we have overlap support & a first compaction implementation, we need to trigger & run compaction sometime after data has been imported. @mattnibs is already in progress thinking about the shorter term Brim/zqd desktop case, as well as the longer term for a zqd service.

mattnibs:
A simple first pass at providing auto-compaction functionality to zqd with
a simple hueristic: A compaction is run when an archivestore is first opened or
after data has been written to the archivestore. If new data is written to the
archive store while a compaction is running the current compaction task will be
canceled.

Add the task package- a simple interface for running background tasks in zqd.
The interface allows a space to initiate a SpaceOp for each background task and
provided the task with the op's context. This ensures that on zqd shutdown or
space deletion the background tasks will shutdown.

Closes #1555


ppl/archive/import.go
```diff
@@ -116,9 +117,17 @@ func (w *Writer) Close() error {
 		}
 		delete(w.writers, ts)
 	}
+	for _, fn := range w.onclose {
+		fn()
+	}
 	return merr
 }
 
+// OnClose passes a function that will be called when the writer is closed.
```
alfred-landrum, line 126: We don't use this pattern anywhere else, afaik. Instead of doing this, we would normally wrap a WriteCloser with another Closer. Is there a reason we can't use that instead of this approach?

alfred-landrum (commented):
I'd like to discuss a different approach to this: We will definitely need some way to run activities that don't happen within the context of a user request. I think a way that will require less explicit locking & cancellation is to base that around the different types of tasks we'll need. So in the case of compaction, we can have a single goroutine that gets a message anytime a write is complete, and it tracks the current compaction calls.

mattnibs:
> I'd like to discuss a different approach to this: We will definitely need some way to run activities that don't happen within the context of a user request. I think a way that will require less explicit locking & cancellation is to base that around the different types of tasks we'll need. So in the case of compaction, we can have a single goroutine that gets a message anytime a write is complete, and it tracks the current compaction calls.

Will do

ppl/archive/import.go
mattnibs: Sure

mattnibs:
@alfred-landrum ready for another look

ppl/zqd/space/compactor.go
```diff
@@ -0,0 +1,79 @@
...
+	ctx, cancel, err := sp.StartOp(ctx)
+	if err != nil {
+		logger.Info("could aquire space op", zap.Error(err))
+		return
+	}
+	defer cancel()
+
+	logger.Info("compaction started")
+	start := time.Now()
+	if err = store.Compact(ctx); err != nil {
```
nwt, line 64:
```suggestion
	if err := store.Compact(ctx); err != nil {
```

ppl/zqd/space/compactor.go
```diff
@@ -0,0 +1,79 @@
...
+		if err == context.Canceled {
+			logger.Info("compaction aborted")
+		} else {
+			logger.Warn("compaction error", zap.Error(err))
+		}
+		return
+	}
+	logger.Info("compaction completed", zap.Duration("duration", time.Since(start)))
```
nwt, line 72: Nit: We usually capitalize the first word of log strings.

ppl/zqd/handlers_test.go
```diff
@@ -758,7 +761,7 @@ func TestIndexSearch(t *testing.T) {
 		},
 	})
 	require.NoError(t, err)
-	_, err = conn.LogPost(context.Background(), sp.ID, nil, babble)
+	_, err = conn.LogPost(context.Background(), sp.ID, nil, babbleSorted)
```
nwt, line 764: Why change this?

mattnibs: Because babble isn't sorted and this was causing archive to create overlaps, which are now getting compacted, and when chunks are compacted their indexes so their indices are getting wiped out. We don't plan on doing rebuilds of indices just yet.

nwt: Instead of adding babbleSorted, how about doing something like this?
```go
	zcts := resolver.NewContext()
	zf, err := detector.OpenFile(zctx, babble, zio.ReaderOpts{})
	require.NoError(t, err)
	defer zf.Close()
	zrc, err := driver.NewReader(context.Background(), "sort ts", zctx)
	require.NoError(t, err)
	defer zrc.Close()
	_, err = conn.LogPost(context.Background(), sp.ID, nil, zrc)
```

ppl/zqd/space/archivespace.go
```diff
@@ -13,7 +13,8 @@ import (
 
 type archiveSpace struct {
 	spaceBase
-	path iosrc.URI
+	path      iosrc.URI
+	compactor *compactor
```
nwt, line 17:
```suggestion
	compactor *compactor
	path      iosrc.URI
```

ppl/zqd/space/space.go
```diff
@@ -120,33 +120,40 @@ func (m *Manager) loadSpaces(ctx context.Context, p iosrc.URI, conf config) ([]S
...
+		m.alphaFileMigrator.Add(s.store.(*filestore.Storage))
+		return []Space{s}, nil
+
+	case api.ArchiveStore:
 		parent := &archiveSpace{
-			spaceBase: spaceBase{id, store, pcapstore, newGuard(), logger},
+			spaceBase: base,
 			path:      p,
 			conf:      conf,
+			compactor: m.compactor,
```
nwt, line 148: Nit: Initialize fields in order of declaration.

ppl/zqd/storage/archivestore/archivestore.go
```diff
@@ -83,11 +91,32 @@ func (s *Storage) Summary(ctx context.Context) (storage.Summary, error) {
...
+	}
+	return err
+}
+
+// NewWriter returns an arcive.Writer so data can be written to the archive
+// store. When the Writer has been closed a task will be started to compact
+// the store.
+func (s *Storage) NewWriter(ctx context.Context) *Writer {
+	writer := archive.NewWriter(ctx, s.ark)
+	return &Writer{writer, s.notifier}
```
nwt, line 119:
```suggestion
	return &Writer{archive.NewWriter(ctx, s.ark), s.notifier}
```

ppl/zqd/storage/archivestore/archivestore.go
```diff
@@ -30,7 +30,14 @@ func Load(ctx context.Context, path iosrc.URI, cfg *api.ArchiveConfig) (*Storage
...
 	}
-	return &Storage{ark: ark}, nil
+	return &Storage{
+		ark:      ark,
+		notifier: notifier,
+	}, nil
+}
+
+type WriteNotifier interface {
+	WriterClosed()
```
nwt, line 40: Maybe call this WriteNotify for a more conventional name?

ppl/zqd/space/compactor.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func (c *compactor) launchCompact(ctx context.Context, id api.SpaceID) {
+	go func() {
+		if err := c.sem.Acquire(ctx, 1); err != nil {
+			return
+		}
+		defer c.sem.Release(1)
+		defer func() { c.compactDone <- id }()
+		c.compact(ctx, id)
```
nwt, line 51: Nit: No need to defer these here. But maybe move `c.compactDone <- id` into compactor.compact?

ppl/zqd/space/compactor.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func (c *compactor) run(ctx context.Context) {
+	c.done.Add(1)
+	active := make(map[api.SpaceID]bool)
+	for {
+		select {
+		case id := <-c.notify:
+			_, ok := active[id]
+			if !ok {
```
nwt, line 62:
```suggestion
			if _, ok := active[id]; !ok {
```

ppl/zqd/space/compactor.go
```diff
@@ -0,0 +1,120 @@
...
+			return
+		}
+		defer c.sem.Release(1)
+		defer func() { c.compactDone <- id }()
+		c.compact(ctx, id)
+	}()
+}
+
+func (c *compactor) run(ctx context.Context) {
+	c.done.Add(1)
```
nwt, line 56: This should happen before the goroutine starts. I suggest moving it and the Done call into newCompactor like this:
```
	c.done.Add(1)
	go func() {
		c.run()
		c.done.Done()
	}

File: ppl/zqd/handlers_test.go (hunk 4 above)
Comment (mattnibs) [reply ppl/zqd/handlers_test.go:764]: I've got to admit I find this unsavory.

File: ppl/zqd/space/compactor.go (hunk 9 above)
Comment (mattnibs) [reply ppl/zqd/space/compactor.go:51]: I'd rather not have a bunch of `c.compactDone <- id` in all the return points of `compact`.

File: ppl/zqd/handlers_test.go (hunk 4 above)
Comment (mattnibs) [reply ppl/zqd/handlers_test.go:764]: Alright I'll do it, this can go away anyways once #1656 is done.

Comment (nwt) [reply ppl/zqd/handlers_test.go:764]: No offense taken if you decide it's too unsavory to swallow!

File: ppl/zqd/space/compactor.go (hunk 9 above)
Comment (nwt) [reply ppl/zqd/space/compactor.go]: Agreed, so I should have suggested moving `defer func() { c.compactDone <- id }()` into compactor.compact.

File: ppl/zqd/core.go [hunk 12]
```diff
@@ -169,3 +169,7 @@ func (c *Core) nextTaskID() int64 {
 func (c *Core) requestLogger(r *http.Request) *zap.Logger {
 	return c.logger.With(zap.String("request_id", getRequestID(r.Context())))
 }
+
+func (c *Core) Shutdown() {
+	c.spaces.Shutdown()
+}
```
nwt, line 175: Nit: Put this after Root so it's with the other exported methods.

ppl/zqd/handlers_test.go
nwt, line 764: Maybe add a comment explaining why babbleSorted is used here?

ppl/zqd/handlers_test.go
```diff
@@ -767,7 +770,7 @@ func TestIndexSearch(t *testing.T) {
 
 	exp := `
 #0:record[key:int64,count:uint64,first:time,last:time]
-0:[257;1;1587518582.06699522;1587517999.06307953;]
+0:[257;1;1587518582.06699522;1587518014.06491752;]
```
nwt, line 773: Why does this change? Do babble and babbleSorted contain different records?

nwt (approved):

ppl/zqd/handlers_test.go
mattnibs: Sure

ppl/zqd/handlers_test.go
mattnibs: There's a slight overlap generated with babble while babbleSorted does not generate an overlap. This causes a shift in chunks start times.

ppl/zqd/space/compactor.go
```diff
@@ -0,0 +1,121 @@
...
+			return
+		}
+		c.compact(ctx, id)
+		c.sem.Release(1)
+		c.compactDone <- id
+	}()
+}
+
+func (c *compactor) run(ctx context.Context) {
+	active := make(map[api.SpaceID]bool)
```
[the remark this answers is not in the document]
mattnibs: I like active.

PR #1626, opened 2020-11-16
Don't run suricata-update in system tests

henridf:
This PR changes the `services.sh` test setup script to pass a nop suricata-updater to zqd. This should address (or at least alleviate) a system test slowdown observed by @alfred-landrum .


mccanne (approved):
yay

PR #1628, opened 2020-11-16
add -P flag to zq

### LINKED ISSUE #1616: add parallel input option to zq
To run zq with a zql join, you need a way to specify the left file and the right file and input those files as separate input to a parallel input zql graph.   We could add a "from" operator that refers generically to the source, but to keep the first steps simple, we will simply an option to zq to specify that the input files are to be used as inputs to a parallel zql query, e.g., 
```
zq -P '( sort ip.resp_h ; sort ip_key ) | join ip.resp_h=ip_key | ...' logs.zng intel.zng
```
This could be used generically beyond the scope of join, e.g., this would just intermix threat intel logs 
within an edge graph computed from the logs file:
```
zq -P '( count() by id.orig_h,id._resp_h; filter * ) | sort id.orig_h' logs.zng intel.zng
```
At most one of the input files may be stdin.  It is an error for the number of files to not match the number of parallel inputs to the zql query. 

The join proc will be added in a subsequent PR.

mccanne:
This commit adds a new flag (-P) to zq to interconnect two or
more inputs directly to a zql query that begins with a parallel
flow graph.  This is needed by the forthcoming join proc.

Closes #1616

### DIFF
--- a/driver/compile.go
+++ b/driver/compile.go
@@ -75,36 +75,7 @@ func (n *namedScanner) Pull() (zbuf.Batch, error) {
 }
 
 func compileSingle(ctx context.Context, program ast.Proc, zctx *resolver.Context, r zbuf.Reader, cfg Config) (*muxOutput, error) {
-	if cfg.Logger == nil {
-		cfg.Logger = zap.NewNop()
-	}
-	if cfg.Span.Dur == 0 {
-		cfg.Span = nano.MaxSpan
-	}
-	if cfg.Warnings == nil {
-		cfg.Warnings = make(chan string, 5)
-	}
-
-	filterExpr, program := programPrep(program, field.Dotted(cfg.ReaderSortKey), cfg.ReaderSortReverse)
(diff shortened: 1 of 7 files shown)


driver/compile.go
```diff
@@ -107,6 +107,45 @@ func compileSingle(ctx context.Context, program ast.Proc, zctx *resolver.Context
 	return newMuxOutput(pctx, leaves, sn), nil
 }
 
+func compileParallel(ctx context.Context, program ast.Proc, zctx *resolver.Context, readers []zbuf.Reader, cfg Config) (*muxOutput, error) {
```
nwt, line 110: Are 
```go
compileSingle(ctx, program, zctx, reader, cfg)
```
and
```go
compileParallel(ctx, program, zctx, []zbuf.Reader{reader}, cfg)
```
equivalent?

If so, I think we should just replace compileSingle with this.

nwt (approved):

PR #1632, opened 2020-11-17
add preliminary join proc

### LINKED ISSUE #1629: create initial join proc

mccanne:
This commit adds a join proc.  It works with the -P flag in zq
and supports just left joins.  Note that the semantics of a left
join are a bit different than sql in that when the expression for
the left key refers to a field that is not present, the record
is dropped from the join.  However, for right keys that are not
matched with the left key, the left record is included in the
result, as with expected left-join semantics.  There is no attempt
to put nulls in the columns for missing righthand keys (though
this could be added perhaps as an option to join).

It will be trivial to add right join and inner join support given
the current framework but I wanted to start with left join to
keep things simple at first while we gather experience and user
feedback.

The left and right inputs must be pre-sorted by the key expressions
or the results are undefined.  We may want to call this version of
join "mergejoin" as an explicit indicator to the query planner that
the operator knows the inputs will be sorted and its doesn't ahve to
do anything special.  Likewise, a vanilla version of join would be
smart enough to use information from the storage layer to determine
automatically which datapaths needed to be sorted and how.

Also, we'll want to add hash joins at some point, but merge joins
should get us pretty far, and we'll want to perform aggregations
while performing the join though this can be an under-the-hood
optimization where aggregations are pushed into the join operator.

Closes #1629

### DIFF
--- a/proc/compiler/compiler.go
+++ b/proc/compiler/compiler.go
@@ -129,18 +129,6 @@ func compileProc(custom Hook, node ast.Proc, pctx *proc.Context, parent proc.Int
 	}
 }
 
-func compileJoin(custom Hook, v *ast.JoinProc, rest []ast.Proc, pctx *proc.Context, left, right proc.Interface) ([]proc.Interface, error) {
-	join, err := join.New(pctx, left, right, v)
-	if err != nil {
-		return nil, err
-	}
-	parents := []proc.Interface{join}
-	if len(rest) == 0 {
-		return parents, nil
-	}
-	return compileSequential(custom, rest, pctx, parents)
-}
-
 func enteringJoin(nodes []ast.Proc) bool {
 	var ok bool
 	if len(nodes) > 0 {
@@ -151,12 +139,6 @@ func enteringJoin(nodes []ast.Proc) bool {
 
 func compileSequential(custom Hook, nodes []ast.Proc, pctx *proc.Context, parents []proc.Interface) ([]proc.Interface, error) {
 	node := nodes[0]
-	if join, ok := node.(*ast.JoinProc); ok {
-		if len(parents) != 2 {
-			return nil, ErrJoinParents
-		}
-		return compileJoin(custom, join, nodes[1:], pctx, parents[0], parents[1])
-	}
 	parents, err := Compile(custom, node, pctx, parents)
 	if err != nil {
 		return nil, err
@@ -223,6 +205,16 @@ func Compile(custom Hook, node ast.Proc, pctx *proc.Context, parents []proc.Inte
 
 	case *ast.ParallelProc:
 		return compileParallel(custom, node, pctx, parents)
+
+	case *ast.JoinProc:
+		if len(parents) != 2 {
+			return nil, ErrJoinParents
+		}
+		join, err := join.New(pctx, parents[0], parents[1], node)
+		if err != nil {
+			return nil, err
+		}
+		return []proc.Interface{join}, nil
 
 	default:
 		if len(parents) > 1 {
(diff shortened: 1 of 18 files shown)


proc/compiler/compiler.go
```diff
@@ -117,14 +120,43 @@ func compileProc(custom Hook, node ast.Proc, pctx *proc.Context, parent proc.Int
...
+}
+
 func compileSequential(custom Hook, nodes []ast.Proc, pctx *proc.Context, parents []proc.Interface) ([]proc.Interface, error) {
 	node := nodes[0]
+	if join, ok := node.(*ast.JoinProc); ok {
+		if len(parents) != 2 {
+			return nil, ErrJoinParents
+		}
+		return compileJoin(custom, join, nodes[1:], pctx, parents[0], parents[1])
+	}
```
nwt, line 159: This is surprising. Why not add a case for ast.JoinProc in Compile instead, as we have for ast.ParallelProc and ast.SequentialProc?

proc/join/puller.go
```diff
@@ -0,0 +1,76 @@
...
+func (p *puller) Pull() (zbuf.Batch, error) {
+	select {
+	case res := <-p.ch:
+		return res.Batch, res.Err
+	case <-p.ctx.Done():
+		return nil, p.ctx.Err()
+	}
+}
+
+func (p *puller) next() (*zng.Record, error) {
```
nwt, line 54: Nit: Methods with this signature and semantics are usually named Read.

proc/join/join.go
```diff
@@ -0,0 +1,272 @@
...
+func (p *Proc) peekRight() (*zng.Record, error) {
+	if p.peek == nil {
+		var err error
+		p.peek, err = p.right.next()
+		if err != nil {
+			return nil, err
+		}
+	}
+	return p.peek, nil
+}
```
nwt, line 142: Can you get the same effect with a zbuf.Peeker?

proc/put/put.go
```diff
@@ -53,21 +52,16 @@ type clauseType struct {
...
-		clauses[k].field, clauses[k].eval, err = expr.CompileAssignment(pctx.TypeContext, &cl)
+	clauses := make([]expr.Assignment, 0, len(node.Clauses))
+	for _, cl := range node.Clauses {
+		c, err := expr.CompileAssignment(pctx.TypeContext, &cl)
 		if err != nil {
 			return nil, err
 		}
-		if len(clauses[k].field) > 1 {
-			name := clauses[k].field.String()
+		clauses = append(clauses, c)
```
nwt, line 62: Nit: Put this after the following if statement.

proc/join/ztests/empty-inner.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+inputs:
+  - name: A.tzng
+    data: |
+      #0:record[a:int32,sa:string]
+      0:[10;a0;]
+      0:[20;a1;]
+      0:[30;a2;]
+      0:[40;a3;]
+  - name: B.tzng
```
nwt, line 12: Maybe call this c.tzng to reflect the column name?

proc/join/ztests/left.yaml
```diff
@@ -0,0 +1,41 @@
...
+      0:[25;a7;]
+      0:[15;a8;]
+  - name: B.tzng
+    data: |
+      #0:record[b:int32,sb:string]
+      0:[20;b20;]
+      0:[40;b40;]
+      0:[60;b60;]
+      0:[40;b40v2;]
+      0:[20;b20v2;]
```
nwt, line 24: Nit: Both ztests would be easier to read if the inputs were already sorted.

nwt (approved):

PR #1634, opened 2020-11-17
archive: fix seekindex for consecutive timestamps

mattnibs:
During archive import do not generate seek indexes if the current record's
timestamp equals the previous records timestamp.

The previous functionality had a bug where searches would drop records if an
entry record for the seekindex had records with equal timetamps before it.

zq#1594

### DIFF
--- a/ppl/archive/import.go
+++ b/ppl/archive/import.go
@@ -315,7 +315,8 @@ func (cw *chunkWriter) Write(rec *zng.Record) error {
 	if err := cw.dataFileWriter.Write(rec); err != nil {
 		return err
 	}
-	if cw.needIndexWrite {
+	ts := rec.Ts()
+	if !cw.wroteFirst || (cw.needIndexWrite && ts != cw.lastTs) {
 		if err := cw.seekIndex.Enter(rec.Ts(), sos); err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


ppl/archive/import.go
```diff
@@ -315,7 +315,8 @@ func (cw *chunkWriter) Write(rec *zng.Record) error {
 	if err := cw.dataFileWriter.Write(rec); err != nil {
 		return err
 	}
-	if cw.needIndexWrite {
+	ts := rec.Ts()
+	if !cw.wroteFirst || (cw.needIndexWrite && ts != cw.lastTs) {
```
alfred-landrum, line 319: You could replace the `rec.Ts()` call below this line with `ts`.

Also, `needIndexWrite` is initialized to true inside `newChunkWriter`; with this check on `wroteFirst`, I don't think it needs to be set there anymore.

nwt: Yeah, zap that.

nwt (approved):

ppl/archive/import.go
mattnibs: I don't quite follow the `wroteFirst` thing. You mean I don't need this here? The problem with that would be if the first Ts was 0, the first offset entry would not have 0 as the ts value.

nwt: @alfred-landrum observation is just that there's no need for `needIndexWrite: true,` in newChunkWriter since you now have `!cw.writeFirst` here, and it's always true on the first trip through this method.

mattnibs: ooooh, good observation!

PR #1636, opened 2020-11-17
fix bug with parents slice corruption in compiler

### LINKED ISSUE #1635: stack overflow with merge proc
In compiler.go, compileSequential calls `merge.New()` with a slice of proc.Interfaces, which `merge.New` does *not* copy, it just retains the slice. compileSequential later rearranges the slice such that the merge proc has a parent pointer to itself, eventually leading to a stack overflow.

mccanne:
This commit fixes a bug where the parents slice was overwritten
and reused so the parent linkage in the compiled flowgraph
becomes corrupt.  The fix is to make a new parents slice instead
of trying to reuse the buffer.

Fixes #1635

### DIFF
--- a/proc/compiler/compiler_test.go
+++ b/proc/compiler/compiler_test.go
@@ -7,6 +7,7 @@ import (
 	"testing"
 
 	"github.com/brimsec/zq/ast"
+	"github.com/brimsec/zq/field"
 	"github.com/brimsec/zq/pkg/test"
 	"github.com/brimsec/zq/proc"
 	"github.com/brimsec/zq/proc/compiler"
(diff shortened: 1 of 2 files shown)


proc/compiler/compiler_test.go
```diff
@@ -67,3 +68,39 @@ func TestCompileParents(t *testing.T) {
...
+0:[1;]
+0:[2;]
+0:[3;]
+0:[4;]
+`
+	zctx := resolver.NewContext()
+	pctx := &proc.Context{Context: context.Background(), TypeContext: zctx}
+	r := tzngio.NewReader(bytes.NewReader([]byte(input)), zctx)
+	src := &proctest.RecordPuller{R: r}
+	t.Run("merge done before scanners out of data", func(t *testing.T) {
```
alfred-landrum, line 88: I suggest removing the t.Run wrapper, and call the inner code directly.

alfred-landrum (approved):
I tested this change with the same setup where I discovered it, and this fixes the stack overflow.

PR #1640, opened 2020-11-18
Add zqd listen -personality flag

### LINKED ISSUE #1630: zqd personalities support
We'll use zqd in different ways in our cloud services; some instances of zqd will act as frontend API servers, authenticating & responding to requests about spaces, and some zqd instances will execute ZQL queries, and so on.  To support these different ways that zqd is expected to act, we want a way to launch zqd and specify which API groups it should support & which code modules to initialize. We're currently calling this a "personality". We'd want to allow specifying multiple simultaneous personalities to zqd, as we'll have use cases (like Brim's local zqd) where an instance may be expected to perform duties that would be split in our cloud service.

For this issue, we'd like a way to specify the personality via the zqd command line, and some initial patterns of how we configure & execute code related to one or more personalities.

nwt:
The zqd listen -personality flag controls the routes handled by zqd.  It
accepts these three values:

   * all: handle all routes
   * apiserver: handle all routes except /worker
   * worker: handle /worker

In addition, all personalities handle the /, /metrics, /status, and
/version routes.

Depends on #1637.

Closes #1630.


ppl/cmd/zqd/listen/command.go
```diff
@@ -113,15 +114,26 @@ func (c *Command) Run(args []string) error {
...
+	case "all":
+		core.AddAPIServerRoutes()
+		core.AddWorkerRoutes()
+	case "apiserver":
+		core.AddAPIServerRoutes()
+	case "worker":
+		core.AddWorkerRoutes()
+	default:
+		return fmt.Errorf("unknown personality %s", c.personality)
+	}
```
mattnibs, line 135: I think it would be better to not call these functions in order for core to server any routes. Ideally it would by default have personality "all". Perhaps make personality a variable in config and move this switch to NewCore?

nwt: @mattnibs: Like 863a2edab8eeab8743fe818d401ddf76eec15aa9?

mattnibs: Excellent

mattnibs (approved):

alfred-landrum (approved):
👍

PR #1642, opened 2020-11-18
fix join key expressions and add test coverage

mccanne:
This commit fixes a bug in the join-key expression grammar
and adds a test to cover it.  There was ambiguity that was not
properly tested in the previous commit.  The ambiguity is resolved
by using parenthesis to wrap an expression for the join key
compared to a direct field reference.  There may be a better
syntactic solution but it will need some creative thought and
we can live with this for now.

### DIFF
--- a/proc/join/join.go
+++ b/proc/join/join.go
@@ -40,11 +40,11 @@ func New(pctx *proc.Context, left, right proc.Interface, v *ast.JoinProc) (*Proc
 		lhs = append(lhs, c.LHS)
 		rhs = append(rhs, c.RHS)
 	}
-	leftKey, err := expr.CompileLval(v.LeftKey)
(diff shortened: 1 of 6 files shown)


proc/join/join.go
```diff
@@ -239,6 +239,14 @@ func (p *Proc) combinedType(left, right *zng.TypeRecord) (*zng.TypeRecord, error
 }
 
 func (p *Proc) splice(left, right *zng.Record) (*zng.Record, error) {
+	if right == nil {
+		// This happens on a simple join, i.e., "join key",
+		// where there are not cut expressions.  For left joins,
```
nwt, line 244:
```suggestion
		// where there are no cut expressions.  For left joins,
```

nwt (approved):

Issue #1540, opened 2020-10-26
Unexpected error message "sequential proc without procs"


philrz:
Repro is with `zq` commit `a1e5c01` and https://github.com/brimsec/zq-sample-data/blob/master/zeek-default/ntp.log.gz as sample data.

I can do the following with plain `zq`:

```
$ zq -t 'cut uid | head 1' ntp.log.gz 
#0:record[uid:bstring]
0:[CYsbY620656cf4K3Ig;]
```

However, if I import it into an archive and attempt the same, I see an error message.

```
$ rm -rf $ZAR_ROOT
$ zq ntp.log.gz | zar import -s 25MB -
$ zar zq -t 'cut uid | head 1'
sequential proc without procs
```

philrz:
Verified in `zq` commit `20329d11`. Repeating the original repro steps, we now see the expected output rather than an error message.

```
$ zq -version
Version: v0.23.0-45-g20329d11

$ rm -rf $ZAR_ROOT
$ zq ntp.log.gz | zar import -s 25MB -
$ zar zq -t 'cut uid | head 1'
#0:record[uid:bstring]
0:[CSApMbgcjo2YuOjXl;]
```

Thanks @alfred-landrum!

PR #1650, opened 2020-11-18
fix and test for parallelized time grouping

### LINKED ISSUE #1641: time grouping not working for zar archives
I was testing Brim with some archive spaces, and noticed that the histogram seemed wrong: a) it populated oddly, with values appearing through the entire time span sporadically, and b) the final result visually looked different when compared with importing data into a file-based store.

I created a small input file (the data I used in Brim was large), and was able to see the difference by using a local `zqd` and some `zapi` commands.

I used this tzng file:
https://gist.github.com/alfred-landrum/cb40d931697223e4a52c5f28cf251627

and ran a comparison like so:

```
$ zapi -s gtest-ark post -k archivestore -thresh 1KiB -f group-test.tzng
$ zapi -s gtest-file post -f group-test.tzng
$ zapi -s gtest-ark get -t 'every 3600s count() by g | count()'
#0:record[count:uint64]
0:[628;]
$ zapi -s gtest-file get -t 'every 3600s count() by g | count()'
#0:record[count:uint64]
0:[480;]
```

Looking at the output from the archive operation, the groupings are out-of-order.

alfred-landrum:
When we parallelize a query with a time grouping, ensure the input sort order direction is passed to the final merge operation. Also fixes a "sequential proc without procs" error that appeared if a bare reducer was passed into `zar zq`.

Fixes #1641


alfred-landrum:
Fyi these tests also verify the fix for #1635 .

alfred-landrum:
Fixes https://github.com/brimsec/zq/issues/1540 , the "sequential proc without procs" error.

ppl/archive/ztests/groupby.yaml
```diff
@@ -0,0 +1,63 @@
+script: |
```
nwt, line 1: Nit: Give this a more informative name like multiday-group-by-ts.yaml.

driver/compile.go
```diff
@@ -544,6 +544,9 @@ func copyProcs(ps []ast.Proc) []ast.Proc {
 }
 
 func buildSplitFlowgraph(branch, tail []ast.Proc, mergeField field.Static, reverse bool, N int) *ast.SequentialProc {
+	if len(branch) == 0 {
+		branch = []ast.Proc{passProc}
+	}
```
nwt, line 549: This doesn't seem right. Wrapping multiple pass procs in a parallel proc can only impair performance.

alfred-landrum: addressed in 6e52d7ec370fe0f40b7f554032f357f4f2930b4f . That uncovered a bug in the first & last reducers, which is addressed in 92ede89508fe82ef9b03a66e22fece961b40f4eb .

driver/compile.go
```diff
@@ -543,7 +543,13 @@ func copyProcs(ps []ast.Proc) []ast.Proc {
...
 }
 
-func buildSplitFlowgraph(branch, tail []ast.Proc, mergeField field.Static, reverse bool, N int) *ast.SequentialProc {
+func buildSplitFlowgraph(branch, tail []ast.Proc, mergeField field.Static, reverse bool, N int) (*ast.SequentialProc, bool) {
+	if len(branch) == 0 {
+		return &ast.SequentialProc{
+			Node:  ast.Node{"SequentialProc"},
+			Procs: tail,
+		}, false
+	}
```
nwt, line 552: That buildSplitFlowgraph now returns an unsplit flowgraph feels like a signal that we should refactor this, but let's save that for later.

nwt (approved):

PR #1652, opened 2020-11-19
archive import: Flush stale tsdirs

### LINKED ISSUE #1433: archive.Import: flush stale tsDirWriter functionality
If a tsDir never gets enough data to reach ark.LogSizeThreshold,
the data will sit in the tsDirWriter and remain unsearchable until the
provided read stream is closed. Add some kind of timeout functionality that
periodically flushes stale tsDirWriters.

mattnibs:
Add a background co-routine to periodically (configured at 1 second intervals)
checks for active tsDirWriters whose time of last modified time is less than
ark.ImportFlushThreshold, and commits its data to disk.

Closes #1433

### DIFF
--- a/ppl/archive/import.go
+++ b/ppl/archive/import.go
@@ -26,6 +26,10 @@ var (
 	// before they are flushed to disk.
 	ImportBufSize          = int64(sort.MemMaxBytes)
 	ImportStreamRecordsMax = zngio.DefaultStreamRecordsMax
+	// DefaultImportFlushTimeout is the default time that an archive will wait
+	// before flushing to disk and closing a tsdir that has not received any new
+	// data.
+	DefaultImportFlushTimeout = time.Second
 )
 
 // For unit testing.
@@ -48,34 +52,39 @@ func Import(ctx context.Context, ark *Archive, zctx *resolver.Context, r zbuf.Re
 // TODO zq#1432 Writer does not currently keep track of size of records
 // written to temporary files. At some point this should have a maxTempFileSize
 // to ensure the Writer does not exceed the size of a provisioned tmpfs.
-//
-// TODO zq#1433 If a tsDir never gets enough data to reach ark.LogSizeThreshold,
-// the data will sit in the tsDirWriter and remain unsearchable until the
-// provided read stream is closed. Add some kind of timeout functionality that
-// periodically flushes stale tsDirWriters.
 type Writer struct {
-	ark      *Archive
-	cancel   context.CancelFunc
-	ctx      context.Context
-	errgroup *errgroup.Group
-	mu       sync.RWMutex
-	once     sync.Once
-	writers  map[tsDir]*tsDirWriter
+	ark          *Archive
+	cancel       context.CancelFunc
+	ctx          context.Context
+	errgroup     *errgroup.Group
+	flushTimeout time.Duration
+	once         sync.Once
+	writers      map[tsDir]*tsDirWriter
+	writersMu    sync.RWMutex
 
 	memBuffered int64
 	stats       ImportStats
 }
 
+// NewWriter creates a zbuf.Writer compliant writer for writing data to an
+// archive.
 func NewWriter(ctx context.Context, ark *Archive) *Writer {
 	g, ctx := errgroup.WithContext(ctx)
 	ctx, cancel := context.WithCancel(ctx)
 	return &Writer{
-		ark:      ark,
-		cancel:   cancel,
-		ctx:      ctx,
-		errgroup: g,
-		writers:  make(map[tsDir]*tsDirWriter),
+		ark:          ark,
+		cancel:       cancel,
+		ctx:          ctx,
+		errgroup:     g,
+		flushTimeout: DefaultImportFlushTimeout,
+		writers:      make(map[tsDir]*tsDirWriter),
 	}
+}
+
+// SetFlushTimeout sets the interval at which the Writer will wait to flush
+// stale records to disk. This must be called before writing any records.
+func (w *Writer) SetFlushTimeout(timeout time.Duration) {
+	w.flushTimeout = timeout
 }
 
 func (w *Writer) Write(rec *zng.Record) error {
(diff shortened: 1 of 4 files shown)


ppl/archive/import.go
```diff
@@ -91,11 +113,50 @@ func (w *Writer) Write(rec *zng.Record) error {
 	return nil
 }
 
+// flusher is a background go-routine for an active Writer that periodically
```
alfred-landrum, line 114:
```suggestion
// flusher is a background goroutine for an active Writer that periodically
```

ppl/archive/import.go
```diff
@@ -91,11 +113,50 @@ func (w *Writer) Write(rec *zng.Record) error {
...
+// checks for tsdir writers that have not recieved data in a while. If such a
```
alfred-landrum, line 115:
```suggestion
// checks for tsdir writers that have not received data in a while. If such a
```

ppl/archive/import.go
```diff
@@ -91,11 +113,50 @@ func (w *Writer) Write(rec *zng.Record) error {
...
+		case <-w.ctx.Done():
+			return nil
+		}
+	}
+}
+
+func (w *Writer) flushStaleWriters() error {
+	w.mu.Lock()
+	var stale []*tsDirWriter
+	now := time.Now()
```
nwt, line 135: No need for the lock across time.Now.
```suggestion
	now := time.Now()
	var stale []*tsDirWriter
	w.mu.Lock()
```

ppl/archive/schema.go
```diff
@@ -18,9 +18,10 @@ const metadataFilename = "zar.json"
 type Metadata struct {
 	Version int `json:"version"`
 
-	DataPath         string     `json:"data_path"`
-	DataOrder        zbuf.Order `json:"data_order"`
-	LogSizeThreshold int64      `json:"log_size_threshold"`
+	DataPath           string        `json:"data_path"`
+	DataOrder          zbuf.Order    `json:"data_order"`
+	ImportFlushTimeout time.Duration `json:"import_flush_time"`
```
alfred-landrum, line 24: The import flush timeout doesn't feel like an archive-defining value the same way that the log size threshold does. Instead of storing it in the archive metadata, it could have a default that could be specified at the import call.

ppl/archive/import.go
```diff
@@ -53,34 +56,53 @@ func Import(ctx context.Context, ark *Archive, zctx *resolver.Context, r zbuf.Re
...
 // periodically flushes stale tsDirWriters.
 type Writer struct {
-	ark     *Archive
-	ctx     context.Context
-	writers map[tsDir]*tsDirWriter
+	ark      *Archive
+	cancel   context.CancelFunc
+	ctx      context.Context
+	errgroup *errgroup.Group
+	mu       sync.RWMutex
```
nwt, line 61: Maybe call this writersMu since it only protects the writers field?

ppl/archive/import.go
```diff
@@ -53,34 +56,53 @@ func Import(ctx context.Context, ark *Archive, zctx *resolver.Context, r zbuf.Re
...
+		writers:  make(map[tsDir]*tsDirWriter),
 	}
 }
 
 func (w *Writer) Write(rec *zng.Record) error {
+	w.once.Do(func() {
+		w.errgroup.Go(w.flusher)
+	})
+	if w.ctx.Err() != nil {
+		return w.errgroup.Wait()
```
nwt, line 86: Why not `return w.ctx.Err()`?

nwt (approved):

ppl/archive/import.go
alfred-landrum: It looks like you've got `mu` trying to just guard the `writers` map, but I think there are code paths where more locking is needed. `memBuffered` for example is incremented I think in both the Write() call and the flush path.

I suggest changing `mu` to a `sync.Mutex`, and calling lock/defer-unlock on it at the top of: `Write`, `Close`, and `flushStaleWriters`'. If you do that, I think you don't need `modts` to be an atomic type.

ppl/archive/schema.go
mattnibs: Agreed.

ppl/archive/import.go
mattnibs: @alfred-landrum good call on memBuffered. WIll fix.

ppl/archive/import.go
mattnibs: I wanted to propogate the errgroup err if the context cancellation is called by the errgroup. That said, I should probably return ctx.Err() if errgroup.Wait() is nil.

ppl/archive/import.go
mattnibs: @alfred-landrum I was concerned about doing too much mutexing here because this is in the critical path for import. That said perhaps doing as you suggest has a negligible impact.

mattnibs: @alfred-landrum I went ahead and followed @nwt 's suggestion. I couldn't get away from not wanting to put a mutex over the whole thing, worried this would slow things down. For myself, I also easier find I'm less worried about deadlocks if I only have to worry about a mutex on the writer (I followed @nwt's suggestion to name it `writersMu`).

I used an atomic to make memBuffered thread safe.

ppl/archive/import.go
```diff
@@ -15,17 +18,24 @@ import (
...
+	"golang.org/x/sync/errgroup"
 )
 
 var (
 	// ImportBufSize specifies the max size of the records buffered during import
 	// before they are flushed to disk.
 	ImportBufSize          = int64(sort.MemMaxBytes)
 	ImportStreamRecordsMax = zngio.DefaultStreamRecordsMax
+	// DefaultImportFlushTimeout is the default time that an archive will wait
+	// before flushing to disk and closing a tsdir has not received any new
```
nwt, line 30:
```suggestion
	// before flushing to disk and closing a tsdir that has not received any new
```

ppl/archive/import.go
```diff
@@ -15,17 +18,24 @@ import (
...
+	// data.
+	DefaultImportFlushTimeout = time.Second
 )
 
-// For unit testing.
-var importLZ4BlockSize = zngio.DefaultLZ4BlockSize
+var (
+	// For unit testing.
+	importLZ4BlockSize = zngio.DefaultLZ4BlockSize
+)
```
nwt, line 38: Nit: Don't use `var (...)` for a single variable.

ppl/archive/import.go
```diff
@@ -44,55 +54,125 @@ func Import(ctx context.Context, ark *Archive, zctx *resolver.Context, r zbuf.Re
...
+	writersMu    sync.RWMutex
 
 	memBuffered int64
 	stats       ImportStats
 }
 
+// NewWriter creates a zbuf.Writer compliant writer for writing data to an
+// archive. If the provided flushTimeout timeout is 0, the flushTimeout will be
+// set to the value of DefaultImportFlushTimeout.
 func NewWriter(ctx context.Context, ark *Archive) *Writer {
```
nwt, line 74: Don't forget to add flushTimeout!

mattnibs: You call `SetFlushTimeout` method on Writer.

ppl/archive/import.go
mattnibs: Yep

PR #1659, opened 2020-11-19
Recruiter balance

marktwallace:
A scheduling algorithm to spread zqd worker instances across the maximum number of available nodes.


ppl/zqd/recruiter/recruiter.go
```diff
@@ -0,0 +1,191 @@
...
+	return &WorkerPool{
+		freePool:     make(map[string]WorkerDetail),
+		nodePool:     make(map[string][]WorkerDetail),
+		reservedPool: make(map[string]WorkerDetail),
+	}
+}
+
+func (pool *WorkerPool) Register(addr string, nodename string) error {
+	if _, _, err := net.SplitHostPort(addr); err != nil {
+		return fmt.Errorf("Invalid address for Register: %w", err)
```
nwt, line 37: Error strings generally should not be begin with a capital letter. See https://github.com/golang/go/wiki/CodeReviewComments#error-strings for details.

ppl/zqd/recruiter/recruiter.go
```diff
@@ -0,0 +1,191 @@
...
+	if nodename == "" {
+		return fmt.Errorf("Node name required for Register")
+	}
+	wd := WorkerDetail{Addr: addr, NodeName: nodename}
+
+	pool.mu.Lock()
+	defer pool.mu.Unlock()
+
+	_, prs := pool.reservedPool[addr]
+	if prs {
```
nwt, line 48: Nits: We usually use the name ok for the second value produced by a map index expression. And we usually do this on one line.
```suggestion
	if _, ok := pool.reservedPool[addr]; ok {
```

ppl/zqd/recruiter/recruiter.go
```diff
@@ -0,0 +1,191 @@
...
+		return nil // ignore register for existing workers
+	}
+	pool.freePool[addr] = wd
+	_, prs = pool.nodePool[nodename]
+	if !prs {
+		pool.nodePool[nodename] = make([]WorkerDetail, 0)
```
nwt, line 54: Appending to a zero-valued slice is fine, so this is unnecessary.

ppl/zqd/recruiter/recruiter.go
```diff
@@ -0,0 +1,191 @@
...
+	SkipSpread   bool                      // option to test algorithm performance
+}
+
+type WorkerDetail struct {
+	Addr     string
+	NodeName string
+}
+
+func NewWorkerPool() *WorkerPool {
+	rand.Seed(time.Now().UnixNano()) // for shuffling nodePool keys
```
nwt, line 27: Other package can undo this. Avoid that by creating a math.Rand for this package.

ppl/zqd/recruiter/recruiter.go
```diff
@@ -0,0 +1,191 @@
...
+// Recruit attempts to spread the workers across nodes.
+func (pool *WorkerPool) Recruit(n int) ([]WorkerDetail, error) {
+	if n < 1 {
+		return nil, fmt.Errorf("Recruit must request one or more workers: n=%d", n)
+	}
+
+	pool.mu.Lock()
+	defer pool.mu.Unlock()
+
+	recruits := make([]WorkerDetail, 0)
```
alfred-landrum, line 114: It's ok to return a nil slice, and you don't need to make this before usage, so this can be:

```suggestion
	var recruits []WorkerDetail
```

ppl/zqd/recruiter/recruiter.go
```diff
@@ -0,0 +1,187 @@
...
+	s[i] = s[len(s)-1]
+	pool.nodePool[wd.NodeName] = s[:len(s)-1]
+}
+
+func (pool *WorkerPool) Deregister(addr string) {
+	pool.mu.Lock()
+	defer pool.mu.Unlock()
+
+	wd, prs := pool.freePool[addr]
+	if prs {
```
nwt, line 87:
```suggestion
	if wd, ok := pool.freePool[addr]; ok {
```

ppl/zqd/recruiter/recruiter_test.go
```diff
@@ -0,0 +1,293 @@
...
+)
+
+func Register1(t *testing.T, addr string, nodename string, fp int, np int, np1 int, rp int) *WorkerPool {
+	wp := NewWorkerPool()
+	wp.Register(addr, nodename)
+	AssertPoolLen(t, wp, nodename, fp, np, np1, rp)
+	return wp
+}
+
+func AssertPoolLen(t *testing.T, wp *WorkerPool, name string, fp int, np int, np1 int, rp int) {
```
alfred-landrum, line 18: The `np1` parameter is unused.

ppl/zqd/recruiter/recruiter_test.go
```diff
@@ -0,0 +1,293 @@
...
+		}
+	}
+	AssertPoolLen(t, wp, nodename, 10, 5, 2, 0)
+	s, err := wp.Recruit(5)
+	require.Nil(t, err)
+	assert.Len(t, s, 5)
+	AssertPoolLen(t, wp, nodename, 5, 5, 1, 5)
+}
+
+func InitRectangle(t *testing.T, wp *WorkerPool, width int, height int) {
```
alfred-landrum, line 94: 'rectangle' and 'triangle' don't really help me understand what these do - can you change these names?

ppl/zqd/recruiter/recruiter_test.go
```diff
@@ -0,0 +1,293 @@
...
+	return wp
+}
+
+func AssertPoolLen(t *testing.T, wp *WorkerPool, name string, fp int, np int, np1 int, rp int) {
+	assert.Equal(t, fp, wp.LenFreePool(), "FreePool len=%d", fp)
+	require.Equal(t, np, wp.LenNodePool(), "NodePool len=%d", np)
+	assert.Equal(t, rp, wp.LenReservedPool(), "ReservedPool len=%d", rp)
+}
+
+func TestRegister1(t *testing.T) {
```
alfred-landrum, line 24: Could you group the short tests here that verify bad or duplicate calls have no impact into a single top-level test (maybe "TestBadCalls")? I think even calling `t.Run` for them is probably overkill, and it will be clearer to have a single test case.

ppl/zqd/recruiter/recruiter.go
```diff
@@ -0,0 +1,187 @@
...
+		return nil, fmt.Errorf("recruit must request one or more workers: n=%d", n)
+	}
+
+	pool.mu.Lock()
+	defer pool.mu.Unlock()
+
+	var recruits []WorkerDetail
+	if len(pool.nodePool) < 1 {
+		return recruits, nil
+	}
```
nwt, line 112: Nit: Say `return nil, nil` here and then declare recruits after the if statement (so the declaration stays close to its use).

ppl/zqd/recruiter/recruiter.go
```diff
@@ -0,0 +1,186 @@
...
+
+	// Make a single pass through the nodes in the cluster that have
+	// available workers, and try to pick evenly from each node. If that pass
+	// fails to recruit enough workers, then we start to pick from the freePool, regardless
+	// of node, until we recruit enough workers, or all available workers.
+	if !pool.SkipSpread && n > 1 {
+		var keys []string
+		for k, _ := range pool.nodePool {
```
nwt, line 116:
```suggestion
		for k := range pool.nodePool {
```

ppl/zqd/recruiter/recruiter.go
```diff
@@ -0,0 +1,186 @@
...
+			}
+			if len(recruits) == n {
+				break
+			}
+		}
+
+		// Delete the recruits obtained in this pass from the freePool
+		for _, wd := range recruits {
+			_, prs := pool.freePool[wd.Addr]
+			if !prs {
```
nwt, line 139:
```suggestion
			if _, ok := pool.freePool[wd.Addr]; !ok {
```

ppl/zqd/recruiter/recruiter_test.go
marktwallace: OK. I will consolidate a few of the other simple tests too.

ppl/zqd/recruiter/recruiter_test.go
marktwallace: I revised the names. BTW, the triangle is a set of nodes where every node has a different number of available workers.

ppl/zqd/recruiter/recruiter_test.go
```diff
@@ -0,0 +1,289 @@
...
+)
+
+func Register1(t *testing.T, addr string, nodename string, fp int, np int, rp int) *WorkerPool {
+	wp := NewWorkerPool()
+	wp.Register(addr, nodename)
+	AssertPoolLen(t, wp, fp, np, rp)
+	return wp
+}
+
+func AssertPoolLen(t *testing.T, wp *WorkerPool, fp int, np int, rp int) {
```
nwt, line 18: Nit: No need to export these.

ppl/zqd/recruiter/recruiter_test.go
```diff
@@ -0,0 +1,289 @@
...
+		nodename = "n" + strconv.Itoa(i)
+		for j := 0; j < height; j++ {
+			addr = nodename + ".x:" + strconv.Itoa(5000+j)
+			err := wp.Register(addr, nodename)
+			require.Nil(t, err)
+		}
+	}
+}
+
+func InitNodesOfVaryingSize(t *testing.T, wp *WorkerPool, size int) {
```
nwt, line 101: Nit: No need to export these.

ppl/zqd/recruiter/recruiter_test.go
```diff
@@ -0,0 +1,289 @@
...
+	wp.Deregister("a.b:5000")
+	AssertPoolLen(t, wp, 0, 0, 0)
+}
+
+func TestRecruit1(t *testing.T) {
+	addr := "a.b:5000"
+	nodename := "n2"
+	wp := Register1(t, addr, nodename, 1, 1, 0)
+	s, err := wp.Recruit(1)
+	require.Nil(t, err)
```
nwt, line 51: This applies in lots of other locations as well.
```suggestion
	require.NoError(t, err)
```

alfred-landrum (approved):

ppl/zqd/recruiter/recruiter.go
```diff
@@ -0,0 +1,178 @@
...
+import (
+	"fmt"
+	"math/rand"
+	"net"
+	"sync"
+	"time"
+)
+
+// WorkerPool is the internal state of the recruiter system
+// The pools are exported only for use in unit tests
```
nwt, line 12: Doc comments should comprise complete sentences. See https://golang.org/doc/effective_go.html#commentary.

And this says the pools are exported, but they aren't.

marktwallace: Thanks! I will correct that in my next PR.

PR #1663, opened 2020-11-20
Deterministic ordered merge

### LINKED ISSUE #1215: Deterministic ordered merge

mattnibs:
In an ordered merge if two records have the compare keys, use the results of
bytes.Compare to settle the tie and get deterministic results.

Also make the sort on archive.Import deterministic.

Closes #1215


ppl/archive/import.go
```diff
@@ -390,6 +391,9 @@ func importCompareFn(ark *Archive) expr.CompareFn {
 		if d > 0 {
 			cmp = 1
 		}
+		if d == 0 {
+			return bytes.Compare(a.Raw, b.Raw)
```
alfred-landrum, line 395: Doesn't this ignore the ordering of the archive?

While you're making changes here, how about changing this function to match how `chunkLess` works: a check on the ordering to see if a & b need to be swapped, then compare timestamps if they aren't equal, and a final return statement that compares the raw bytes?

mattnibs: Allow me to push back a bit: Does it matter if sorted descending or ascending? It's an arbitrary check to give us what we want: deterministic results.

If we do this here then we should probably also do this in the merger changes, but I'm not quite sure the best way to do that given that a merger can receive a generic compare function. I suppose we could do this where expressions are compiled?

alfred-landrum: I agree that with or without depending on the order, the result will be deterministic. However, I think it will be more consistent & less surprising if it uses the ordering.

As for merge: you could update `merge.NewCompareFn`, as it takes a `reversed` parameter.

mattnibs: Understood. Just to be clear, this will not provide a deterministic merge on compareFn that do not use this. Just want to verify we are ok with this for now?

mattnibs: Actually I could add the deterministic check in expr.NewCompareFn.

ppl/archive/ztests/deterministic-merge.yaml
```diff
@@ -0,0 +1,28 @@
+script: |
+  zar import -R logs -s 32B -
+  zar zq -t -R logs "*" > 1.tzng
+  zar zq -t -R logs "*" > 2.tzng
+  diff 1.tzng 2.tzng && echo "success"
```
nwt, line 5:
```suggestion
  diff 1.tzng 2.tzng && echo success
```

zbuf/merger.go
```diff
@@ -186,3 +188,13 @@ func (m *Merger) Pull() (Batch, error) {
 func (m *Merger) Cancel() {
 	m.cancel()
 }
+
+func deterministicCompare(fn expr.CompareFn) expr.CompareFn {
```
nwt, line 192: Nits: Put this after NewCompareFn. And deterministicCompare doesn't really describe what this does (it isn't any more or less deterministic than fn), so maybe call it totalOrderCompare or something else that expresses the intent, which I'd say is to describe a total order base on fn.

nwt (approved):

PR #1665, opened 2020-11-20
unify the three variants of AST

### LINKED ISSUE #1638: go generated ast missing Assignment op value
When I generated the ast for `every 3600s count() by _path` via the `ast` command, I noticed a difference between the js and go versions; the go version was missing the `"op": Assignment` key/values that appeared in the js output:

```
$ ast 'every 3600s count() by _path'
{
  "op": "SequentialProc",
  "procs": [
    {
      "op": "GroupByProc",
      "duration": {
        "seconds": 3600
      },
      "keys": [
        {
          "rhs": {
            "op": "BinaryExpr",
            "operator": ".",
            "lhs": {
              "op": "RootRecord"
            },
            "rhs": {
              "op": "Identifier",
              "name": "_path"
            }
          }
        }
      ],
      "reducers": [
        {
          "rhs": {
            "op": "Reducer",
            "operator": "count"
          }
        }
      ]
    }
  ]
}
```

```
$ ast -js 'every 3600s count() by _path'
{
    "op": "SequentialProc",
    "procs": [
        {
            "op": "GroupByProc",
            "reducers": [
                {
                    "op": "Assignment",
                    "rhs": {
                        "op": "Reducer",
                        "operator": "count",
                        "where": null
                    }
                }
            ],
            "duration": {
                "type": "Duration",
                "seconds": 3600
            },
            "keys": [
                {
                    "op": "Assignment",
                    "rhs": {
                        "op": "BinaryExpr",
                        "operator": ".",
                        "lhs": {
                            "op": "RootRecord"
                        },
                        "rhs": {
                            "op": "Identifier",
                            "name": "_path"
                        }
                    }
                }
            ]
        }
    ]
}
```

mccanne:
This commit unifies the parsed ASTs so that whether an AST is
generated by pigeon, pegjs, or by marshaling an ast.Proc, the
resulting JSON is deep equal for the same zql query text.
Previously, different versions treated the zero and empty values
differently.

This update involves changes to the peg grammar actions
as well as the go ast package.  We also updated the ast command
so that it can display these three variations, as previously,
there was no way to display the pigeon version of the ast.

We updated the unit tests in zq/zql to compare the three variants
and added a test that looks through all the ztest yaml files to find
zqls to run and run the three-way comparison.

Fixes #1638

### DIFF
--- a/zql/zql_test.go
+++ b/zql/zql_test.go
@@ -1,15 +1,85 @@
-package zql
+package zql_test
 
 import (
 	"bufio"
+	"encoding/json"
 	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"regexp"
+	"strings"
 	"testing"
 
 	"github.com/brimsec/zq/ast"
 	"github.com/brimsec/zq/pkg/fs"
+	"github.com/brimsec/zq/zql"
+	"github.com/brimsec/zq/ztest"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
+
+func searchForZqls() ([]string, error) {
+	var zqls []string
+	pattern := fmt.Sprintf(`.*ztests\%c.*\.yaml$`, filepath.Separator)
+	re := regexp.MustCompile(pattern)
+	err := filepath.Walk("..", func(path string, info os.FileInfo, err error) error {
+		if !info.IsDir() && strings.HasSuffix(path, ".yaml") && re.MatchString(path) {
+			zt, err := ztest.FromYAMLFile(path)
+			if err != nil {
+				return err
+			}
+			z := zt.ZQL
+			if z == "" || z == "*" {
+				return nil
+			}
+			zqls = append(zqls, z)
+		}
+		return err
+	})
+	return zqls, err
+}
+
+func parsePEGjs(z string) ([]byte, error) {
+	cmd := exec.Command("node", "run.js", "-e", "start")
+	cmd.Stdin = strings.NewReader(z)
+	return cmd.Output()
+}
+
+func parseProc(z string) ([]byte, error) {
+	proc, err := zql.ParseProc(z)
+	if err != nil {
+		return nil, err
+	}
+	return json.Marshal(proc)
+}
+
+func parsePigeon(z string) ([]byte, error) {
+	ast, err := zql.Parse("", []byte(z))
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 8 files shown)


mccanne:
I ran `npm test` in brim with zq commit 8d4968999c42c067166ad7f35ab8c959503b35c1 and it passes.  So presuming I did that right, I think we're good to go.

zql/zql_test.go
```diff
@@ -1,24 +1,105 @@
...
+	pigeonJSON, err := parsePigeon(line)
+	assert.NoError(t, err, "parsePigeon: %q", line)
+
+	astJSON, err := parseProc(line)
+	assert.NoError(t, err, "parseProc: %q", line)
+
+	assert.JSONEq(t, string(pigeonJSON), string(astJSON), "pigeon and ast.Proc mismatch: %q", line)
+
+	if runtime.GOOS != "windows" {
+		var err error
```
alfred-landrum, line 81:
```suggestion
```

alfred-landrum (approved):

nwt (commented):
Making changes to zql.peg and ast.go so you can compare JSON seem weird. Why not compare ast.Procs?

mccanne:
> Making changes to zql.peg and ast.go so you can compare JSON seem weird. Why not compare ast.Procs?

The bug that Al filed indeed had the same ast.Proc because a missing field and a present but `null` field are treated the same by the go unmarshaler.   People have been surprised that the ASTs can be different so I was trying to adopt the convention that the JSON ASTs will always the same wrt to zero and null values no matter where they come from.  I argued in the past that this wasn't a big deal but it came up again so I relented and dove in changed things and added the tests.

mccanne:
Also, trying to the "omit if empty" design pattern in the PEG rules made for more complicated code compared to just allowing null.  This also encourages you to think just a little bit more about the zero-value case when adding new AST nodes since we won't assume "omit if empty" on the go side.

nwt:
> The bug that Al filed indeed had the same ast.Proc because a missing field and a present but null field are treated the same by the go unmarshaler.

I should been clearer and written "compare ast.Procs for deep equality" instead of just "compare ast.Procs". I think that's a better place to draw the line since that's what we really care about It also means adding an omitempty in ast.go won't break TestZtestZqls, which is surprising.

With that said, I feel heard, so feel free to merge this if you prefer it (in which case we ought to explicitly call out our JSON-equivalence policy somewhere in code, tests, or docs).

PR #1669, opened 2020-11-20
Time entire pipelines in scripts/comparison-test.sh

nwt:
Times reported by scripts/comparison-test.sh for jq and zeek-cut reflect
only the associated zcat command, yielding CPU times that are
particularly misleading.  Adjust the position of the time command so
reported times reflect entire pipelines, including both zcat and the
subsequent jq or zeek-cut command.


philrz (approved):
Thanks very much for catching this @nwt!

PR #1673, opened 2020-11-21
Fix NDJSON encoding of single-character string "-"

### LINKED ISSUE #1325: A lone hyphen in NDJSON is changed when read/written
Repro is with `zq` commit `2fe0e93`.

```
$ echo '{"Hyphen": "-"}' | zq -f ndjson -
{"Hyphen":"\\u002d"}
```

I don't see the problem if the hyphen is embedded in other text.

```
$ echo '{"Hyphen": "foo-bar"}' | zq -f ndjson -
{"Hyphen":"foo-bar"}
```

`jq` has no problem with this:

```
$ echo '{"Hyphen": "-"}' | jq -c .
{"Hyphen":"-"}
```

nwt:
zio/ndjsonio.Writer encodes "-" as "\\u002d".  Encode it as "-" instead.

Fixes #1325.

### DIFF
--- a/zng/string.go
+++ b/zng/string.go
@@ -53,7 +53,7 @@ func uescape(r rune) []byte {
 }
 
 func (t *TypeOfString) StringOf(zv zcode.Bytes, fmt OutFmt, inContainer bool) string {
-	if bytes.Equal(zv, []byte{'-'}) {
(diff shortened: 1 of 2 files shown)


mccanne (approved):
I didn't know about the &* yaml trick.  Cool.

PR #1674, opened 2020-11-22
Recruiter api

marktwallace:
zqd recruit manages a pool of zqd workers and provides a mechanism for zqd root processes to recruit zqd worker processes for query execution.

The zqd recruiter is started with: zqd --personality=recruiter

API is described in the README


ppl/zqd/core.go
```diff
@@ -136,6 +141,13 @@ func (c *Core) addWorkerRoutes() {
 	c.handle("/worker", handleWorker).Methods("POST")
 }
 
+func (c *Core) addRecruiterRoutes() {
+	c.handle("/deregister", handleDeregister).Methods("POST")
+	c.handle("/recruit", handleRecruit).Methods("POST")
+	c.handle("/register", handleRegister).Methods("POST")
+	c.handle("/unreserve", handleUnreserve).Methods("POST")
```
mattnibs, line 148: Instead of putting these commands in the root namespace, can prefix this commands with `/recruiter`?

mattnibs: Or maybe `/workers`?

ppl/zqd/rechandlers.go
```diff
@@ -0,0 +1,75 @@
...
+		return
+	}
+	err := c.workerPool.Register(req.Addr, req.NodeName)
+	if err != nil {
+		respondError(c, w, r, zqe.ErrInvalid(err))
+		return
+	}
+	respond(c, w, r, http.StatusOK, api.StatusResponse{
+		Status: "ok",
+	})
```
mattnibs, line 63: The `StatusResponse` is for the status endpoint. Either way I think it would be better to just return 201 no content and leave out the body.

ppl/zqd/rechandlers.go
```diff
@@ -0,0 +1,75 @@
...
+
+func handleUnreserve(c *Core, w http.ResponseWriter, r *http.Request) {
+	var req api.UnreserveRequest
+	if !request(c, w, r, &req) {
+		return
+	}
+	c.workerPool.Unreserve(req.Addr)
+	respond(c, w, r, http.StatusOK, api.StatusResponse{
+		Status: "ok",
+	})
```
mattnibs, line 74: Same as above, I would return 201 No Content.

ppl/zqd/rechandlers.go
```diff
@@ -0,0 +1,75 @@
...
+
+func handleDeregister(c *Core, w http.ResponseWriter, r *http.Request) {
+	var req api.DeregisterRequest
+	if !request(c, w, r, &req) {
+		return
+	}
+	c.workerPool.Deregister(req.Addr)
+	respond(c, w, r, http.StatusOK, api.StatusResponse{
+		Status: "ok",
+	})
```
mattnibs, line 29: As suggested below, I would return 201 No Content here.

ztests/suite/zqd/rec-curl.yaml
```diff
@@ -0,0 +1,84 @@
+script: |
+  source recserv.sh
+  echo ===
+  curl --header "Content-Type: application/json" -request POST \
+    --data '{"addr":"a.b.c:5000","node":"a.b"}' \
+    http://$ZQD_HOST/register 2> err
```
mattnibs, line 6: Though I certainly respect the simplicity of the api, all of our requests have supporting zapi commands. I think we should add recruiter commands to the connection api and add recruiter commands to zapi, then execute these commands using zapi. This would make these test more concise and readable.

ztests/suite/zqd/rec-curl.yaml
```diff
@@ -0,0 +1,84 @@
...
+  echo ===
+  curl --header "Content-Type: application/json" -request POST \
+    --data '{"N":1}' \
+    http://$ZQD_HOST/recruit 2> err
+  #
+  # Now register a node, then deregister it, and verify that
+  # it is not available to be recruited
+  #
```
mattnibs, line 44: I for one find these story based, multiple cases in one tests hard to isolate when I'm unfamiliar with the code base but need to fix a breakage. I would prefer this test case be broken down into smaller isolated test cases.

ppl/zqd/rechandlers.go
```diff
@@ -0,0 +1,75 @@
...
+		respondError(c, w, r, zqe.ErrInvalid(err))
+		return
+	}
+	workers := make([]api.Worker, len(ws))
+	for i, e := range ws {
+		workers[i] = api.Worker{WorkerAddr: api.WorkerAddr{Addr: e.Addr}, NodeName: e.NodeName}
+	}
+	respond(c, w, r, http.StatusOK, api.RecruitResponse{
+		Workers: workers,
+	})
```
mattnibs, line 48: I am wondering if these case where a request yields no workers should return an error response. This is probably something we don't want to occur regularly no? At the very least, from a dev-ops perspective if it returned an error response this would be visible in the http response log.

ppl/zqd/recruiter/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+{"addr":"*host:port for worker*"}
+
+/unreserve is called by a zqd worker process that becomes idle after having completed a /worker request. It is removed from the reserved pool. This is a noop if it was not in the reserved pool.
+
+/unreserve is also called by a zqd worker process when it starts up, in case the addr of the worker was reserved by a previous process that is no longer running.
+
+The /unreserve API, along with the reserved pool maintained by /recruit, avoids the potential race condition of a worker re-registering itself after it has been recruited but before it has been sent work by a zqd root process.
+
+## Analysis of failures
```
alfred-landrum, line 58: @marktwallace : The implementation described here differs from the 'long poll' idea we discussed (as a comment in the zqd recruit doc, and live). As written here, if a zqd root (or any other entity) halts after making a `/recruit` call, the workers would essentially need to be restarted to clear their reservations.

The goal of that 'long poll' idea was to connect the reservation state in the recruiter directly to the liveness of the zqd root: the root makes a connection to the recruiter, gets a reservation, and as long as it holds the connection open, the reservation is maintained. When it closes the connection, the reservation is released.  That simplifies the worker-recruiter needs as well: workers just ping the recruiter on startup, then periodically afterwards; the workers don't need to worry about their reservation state.

ppl/zqd/core.go
marktwallace: OK, I will make that change

ztests/suite/zqd/rec-curl.yaml
marktwallace: Yes, my current working branch implements the changes to the connection api. However, I think it is important that we test these commands with curl also, since there are likely to be situations in K8s deployment where a system other that our client code will need to call one or or more of these API's.

ppl/zqd/rechandlers.go
marktwallace: That's a good point. Either the recruiter or the client should log something whenever the number returned is less than the number requested.

ppl/zqd/recruiter/README.md
marktwallace: A zqd worker will also clear the reserved state when it first becomes "idle" after finishing a /worker request, so a restart would not be needed. The client side of this system, including the connection implementation, is in progress in the branch recruiter-worker.

ppl/zqd/core.go
```diff
@@ -136,6 +141,13 @@ func (c *Core) addWorkerRoutes() {
 	c.handle("/worker", handleWorker).Methods("POST")
 }
 
+func (c *Core) addRecruiterRoutes() {
```
nwt, line 144: Nits: Keep methods sorted by putting this between addAPIServerRoutes and addWorkerRoutes.

ppl/zqd/recruiter/recruiter.go
```diff
@@ -33,6 +34,10 @@ func NewWorkerPool() *WorkerPool {
 	}
 }
 
+// Register adds workers to both the freePool and to the nodePool
```
nwt, line 37:
```suggestion
// Register adds workers to both the freePool and to the nodePool.
```

marktwallace:
Thanks everyone for the comments. I will implement these suggestions.
Although this server API has independent tests, it seems like this PR will be more 
complete for review if I go ahead and add the client connection work that I have
in progress in the recruiter-worker branch. So, I will close this PR and the changes
for the comments above will be included in a PR that merges the recruiter-worker branch.

Issue #1680, opened 2020-11-22
zson writer


mccanne:
First stab at zson.  Get something experimental merged that people can play with so we can refine...

Issue #1678, opened 2020-11-22
write zson spec


mccanne:
Draft the zson spec and deprecate the tzng spec.  I think the zson spec should be in a another markdown doc separate from zng.  We can cross-reference the two of them and cite how they are joined at the hip.

PR #1681, opened 2020-11-22
add experimental zson writer

mccanne:
This commit adds a first cut at the zson writer.  It came together quite
nicely and is ready for people to try out and provide feedback.
In subsequent PRs, we'll add an update to the spec (issue #1678),
comprehensive tests, reader support (issue #1680), and misc other
finishes like enums and proper string escaping.


zio/zsonio/ztests/first-test.yaml
```diff
@@ -0,0 +1,118 @@
+zql: "*"
```
mattnibs, line 1: Can we get a better more descriptive test than `first-test.yaml`? Ideally shouldn't these be broken up by test case?

zio/zsonio/writer.go
```diff
@@ -0,0 +1,366 @@
...
+	"github.com/brimsec/zq/zng"
+	"github.com/brimsec/zq/zng/resolver"
+)
+
+type Writer struct {
+	writer      io.WriteCloser
+	zctx        *resolver.Context
+	mapper      *resolver.Mapper
+	tags        typemap
+	nrec        int
```
mattnibs, line 22: This field is unused.

zio/zsonio/ztests/first-test.yaml
```diff
@@ -0,0 +1,118 @@
...
+      _path: "conn",
+      ts: 2018-03-24T17:15:20.600725Z,
+      id: {
+          orig_h: 10.47.1.152,
+          orig_p: 49562 (uint16) (port),
+          resp_h: 23.217.103.245,
+          resp_port: 80 (port)
+      } (25) (flow),
+      proto: "tcp" (zenum)
+  } (24),
```
mattnibs, line 38: I'm sorry if I missed the explanation, but why do these need a typeid next to them?

zio/zsonio/ztests/first-test.yaml
nwt: first-test.yaml is an instant classic!

zio/zio.go
```diff
@@ -23,6 +24,7 @@ type WriterOpts struct {
 	EpochDates bool
 	Text       textio.WriterOpts
 	Zng        zngio.WriterOpts
+	ZSON       zsonio.WriterOpts
```
nwt, line 27: Nit: We should be consistent about capitalization of acronyms, either by changing this to Zson or (in another PR) changing Zng and Zst to ZNG and ZST.

nwt (approved):

zio/zsonio/ztests/first-test.yaml
mccanne: If the body of a record is fully type-defined, then ZSON doezs not required the extra type decorator but it's also ok to have it.  The writer doesn't have logic to determine if it's redundant so it always includes it for now.  The reader (coming soon) will handle both cases correctly.

zio/zsonio/writer.go
```diff
@@ -0,0 +1,366 @@
...
+				return err
+			}
+		}
+		return w.writeDecorator(decType)
+	}
+	var err error
+	switch t := typ.(type) {
+	//XXX Need enum support. See #1676.
+	default:
+		err = w.writePrimitive(indent, typ, bytes)
```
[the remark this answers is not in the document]
mccanne: I like putting default first when it's a class of things that is operated on then the following things are the finer-grained exceptions.  When the default case is error or null, I think it's better going last.  Maybe we could discuss and agree on the convention?   I'm of course happy to go with the consensus here.  I will leave this one as is for now.

zio/zio.go
mccanne: I would vote for the latter.

PR #1685, opened 2020-11-23
Move archive Chunk to its own package

mattnibs:
Overtime package archive is becoming more complex with more moving parts.
Moving Chunk into its own package is an attempt at simplifying things. A chunk
is only aware of the directory it is supposed to write itself in. The wide
archive that is made up of chunks is responsible for writing them to the
correct tsdir.

Having a simplified chunk will allow for more isolated testing on specific
chunk features.

### DIFF
--- a/ppl/archive/chunk/chunk.go
+++ b/ppl/archive/chunk/chunk.go
@@ -54,7 +54,7 @@ func FileMatch(s string) (kind FileKind, id ksuid.KSUID, ok bool) {
 // seekIndexPath returns the path of an associated microindex written at import
 // time, which can be used to lookup a nearby seek offset for a desired
 // timestamp.
-// metadataPath returns the path of an associated zng file that holds
+// MetadataPath returns the path of an associated zng file that holds
 // information about the records in the chunk, including the total number,
 // and the first and last record timestamps.
 type Chunk struct {
--- a/ppl/archive/chunk/chunk.go
+++ b/ppl/archive/chunk/chunk.go
@@ -1,0 +1,167 @@
+package chunk
+
+import (
+	"context"
+	"fmt"
+	"regexp"
+	"sort"
+
+	"github.com/brimsec/zq/pkg/iosrc"
+	"github.com/brimsec/zq/pkg/nano"
+	"github.com/brimsec/zq/zbuf"
+	"github.com/brimsec/zq/zqe"
+	"github.com/segmentio/ksuid"
+)
+
+// A FileKind is the first part of a file name, used to differentiate files
+// when they are listed from the archive's backing store.
+type FileKind string
+
+const (
+	FileKindUnknown  FileKind = ""
+	FileKindData              = "d"
+	FileKindMetadata          = "m"
+	FileKindSeek              = "ts"
+)
+
+var fileRegex = regexp.MustCompile(`(d|m)-([0-9A-Za-z]{27}).zng$`)
+
+func FileMatch(s string) (kind FileKind, id ksuid.KSUID, ok bool) {
+	match := fileRegex.FindStringSubmatch(s)
+	if match == nil {
+		return
+	}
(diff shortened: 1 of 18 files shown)


ppl/archive/chunk/reader.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+type Reader struct {
+	io.Reader
+	io.Closer
+	TotalSize int64
+	ReadSize  int64
+}
+
+// newChunkReader returns an io.ReadCloser for this chunk. If the chunk has a seek
```
alfred-landrum, line 21: Would you update the comment here, both for the method name & the return type it mentions?

ppl/archive/multisource.go
```diff
@@ -115,15 +117,17 @@ func (m *spanMultiSource) SendSources(ctx context.Context, span nano.Span, srcCh
 func (m *spanMultiSource) SourceFromRequest(ctx context.Context, req *api.WorkerRequest) (driver.Source, error) {
 	si := SpanInfo{Span: req.Span}
 	for _, p := range req.ChunkPaths {
-		tsd, _, id, ok := parseChunkRelativePath(p)
+		dir, file := path.Split(p)
```
alfred-landrum, line 120: 👍

ppl/archive/chunk/chunk.go
```diff
@@ -0,0 +1,167 @@
...
+// A Chunk is a file that holds records ordered according to the archive's
+// data order.
+// To support reading chunks that contain records originally from one or
+// more other chunks, a chunk has a list of chunk IDs it "masks". During a read
+// of data for time span T, if chunks X and Y both have data within span T, and
+// X masks Y, then only data from X should be used.
+// seekIndexPath returns the path of an associated microindex written at import
+// time, which can be used to lookup a nearby seek offset for a desired
+// timestamp.
+// metadataPath returns the path of an associated zng file that holds
```
alfred-landrum, line 57:
```suggestion
// MetadataPath returns the path of an associated zng file that holds
```

alfred-landrum (approved):
thanks for this!

ppl/archive/chunk/reader.go
mattnibs: Whoops, yeah I'll check through the rest of the comments as well.

ppl/archive/chunk/metadata.go
```diff
@@ -0,0 +1,74 @@
...
+		Masks:       md.Masks,
+		Size:        md.Size,
+	}
+}
+
+func MetadataPath(dir iosrc.URI, id ksuid.KSUID) iosrc.URI {
+	return dir.AppendPath(fmt.Sprintf("%s-%s.zng", FileKindMetadata, id))
+}
+
+func WriteMetadata(ctx context.Context, uri iosrc.URI, md Metadata) error {
```
nwt, line 39: Maybe make this a method?

ppl/archive/chunk/metadata.go
```diff
@@ -0,0 +1,74 @@
...
+type Metadata struct {
+	First       nano.Ts
+	Last        nano.Ts
+	RecordCount uint64
+	Masks       []ksuid.KSUID
+	Size        int64
+}
+
+func (md Metadata) Chunk(dir iosrc.URI, id ksuid.KSUID) Chunk {
+	return Chunk{
```
nwt, line 24: Because I love single-character receiver names and Metadata is on the large size for a value receiver.
```suggestion
func (m *Metadata) Chunk(dir iosrc.URI, id ksuid.KSUID) Chunk {
	return Chunk{
```

nwt (approved):

ppl/archive/chunk/metadata.go
mattnibs: Yes you do

PR #1689, opened 2020-11-24
Fix flaky zqd create archive test

mattnibs:
TestCreateArchiveSpace was flaky in ci, I'm guessing because a compaction gets
kicked off. Use the sorted babble data set that will not trigger a compaction.


mccanne (approved):
Woohoo!

PR #1693, opened 2020-11-24
refactor space manager

alfred-landrum:
This change moves & refactors the code responsible for space operations. Properties of spaces (name, storage location, etc) are now stored within a single file, which is treated similar to a database.

For archive based storage, we don't keep any state within zqd between interactions with the archive. For file based storage, as commented in apiserver.Manager, we do retain instances due to the non-persisted seek indexes we use to speed up queries for sub-portions of the space.

I've run the Brim tests against this code locally, btw, and verified they passed.


api/api.go
```diff
@@ -89,6 +92,11 @@ var spaceIDRegexp = regexp.MustCompile("^[a-zA-Z0-9_]+$")
 
 type SpaceID string
 
+func NewSpaceID() SpaceID {
+	id := ksuid.New()
+	return SpaceID(fmt.Sprintf("sp_%s", id.String()))
+}
+
```
nwt, line 99: I think this should live in ppl/zqd/apiserver since isn't really part of the API.

api/api.go
```diff
@@ -326,3 +334,24 @@ type ArchiveOpenOptions struct {
...
+func SafeName(proposed string) string {
+	var sb strings.Builder
+	for _, r := range proposed {
+		if invalidSpaceNameRune(r) {
+			r = '_'
+		}
+		sb.WriteRune(r)
+	}
+	return sb.String()
+}
```
nwt, line 357: I think this too should live in ppl/zqd/apiserver since isn't really part of the API.

ppl/zqd/apiserver/filedb.go
```diff
@@ -0,0 +1,201 @@
...
+	"encoding/json"
+	"io"
+	"sync"
+
+	"github.com/brimsec/zq/api"
+	"github.com/brimsec/zq/pkg/iosrc"
+	"github.com/brimsec/zq/zqe"
+)
+
+type FileDb struct {
```
nwt, line 14: s/Db/DB/ everywhere like database/sql.

ppl/zqd/apiserver/filedb.go
```diff
@@ -0,0 +1,201 @@
...
+	mu   sync.Mutex
+	path iosrc.URI
+}
+
+func CreateFileDb(ctx context.Context, path iosrc.URI, rows []SpaceRow) (*FileDb, error) {
+	ldb := &FileDb{path: path}
```
nwt, line 20: I'm guessing l is for lake, but I wouldn't use it in the names of things in this file that aren't lakes.

ppl/zqd/apiserver/filedb.go
```diff
@@ -0,0 +1,201 @@
...
+func (ldb *FileDb) load(ctx context.Context) ([]SpaceRow, error) {
+	b, err := iosrc.ReadFile(ctx, ldb.path)
+	if err != nil {
+		return nil, err
+	}
+	var lf dbdataV4
+	if err := json.Unmarshal(b, &lf); err != nil {
+		return nil, err
+	}
+	return lf.SpaceRows, nil
```
nwt, line 52: Check the version?

ppl/zqd/apiserver/filedb.go
alfred-landrum: ah, thanks. Will keep the naming change for a separate PR.

api/api.go
alfred-landrum: I moved these few functions to a separate package, as they are needed by both apiserver and the oldconfig pacakge.

ppl/zqd/apiserver/filedb.go
alfred-landrum: done in https://github.com/brimsec/zq/pull/1693/commits/5974423b48c726319a9414cb575a93e9f21c5785

ppl/zqd/handlers.go
```diff
@@ -94,40 +93,26 @@ func handleSearch(c *Core, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	s, err := c.spaces.Get(req.Space)
-	if err != nil {
-		respondError(c, w, r, err)
-		return
-	}
-
-	ctx, cancel, err := s.StartOp(r.Context())
```
mattnibs, line 103: I am likely missing, but it appears the StartOp guards are going away in this pr. What happens if a long running search is issued, a delete is issueed which removes directory structure that the search is currently running on? Wouldn't the search get a strange error back?

This problem will be more difficult with zqd as a service, but nevertheless still exists.

mattnibs (commented):
Definitely a step in the right direction. Good stuff!

ppl/zqd/handlers.go
alfred-landrum: It is possible that we'll see some "file not found" errors with this change if a delete happens concurrently with the search. I saw an example of that error propagated to Brim, and I also saw an instance where the Brim tab (where the search was running) smoothly stopped the search and flipped to another remaining space.

For the service, I think our flow for this should be that the delete request marks the space id as deleted, but doesn't directly execute the storage removal operation, since that could take an exceedingly long time. The actual removal of data would happen as a background process, scheduled sometime after the delete request, which would have the side effect of allowiong shorter queries/searches to continue unabated.

mattnibs: > It is possible that we'll see some "file not found" errors with this change if a delete happens concurrently with the search. I saw an example of that error propagated to Brim, and I also saw an instance where the Brim tab (where the search was running) smoothly stopped the search and flipped to another remaining space.

Seems like we should keep this functionality then, as this leads to unpredictable/inaccurate errors.

nwt (approved):

ppl/zqd/handlers.go
alfred-landrum: I retested, and don't think there's a big enough change to hold this back: if we cancel during import with the current Brim release, Brim can show an error that is effectively "path not found" in the current UI. To see the same during a search query, I have to first have a big enough data set that the search lasts long enough for me start the space delete & confirm it. 

I'll file an issue that tracks handling space deletion in a clustered environment, so we can discuss how that should work.

PR #1695, opened 2020-11-24
Recruiter API in zqd for server and client 

marktwallace:
The zqd recruiter manages a pool of zqd workers and provides a mechanism for zqd root processes to recruit zqd worker processes for query execution. This PR includes the server API provided by the recruiter process, and the client side of the same API that is used by the zqd worker processes.

The zqd recruiter is started with: zqd --personality=recruiter

The API is described in ppl/zqd/recruiter/README


ppl/zqd/recruiter/workerpool.go
```diff
@@ -33,26 +35,30 @@ func NewWorkerPool() *WorkerPool {
 	}
 }
 
-func (pool *WorkerPool) Register(addr string, nodename string) error {
+// Register adds workers to both the freePool and to the nodePool.
+// In the nodePool, they are added to the end of the slice,
+// and when they are recruited they are removed from the start of the slice.
+// So the []WorkerDetail slice for each node functions as a FIFO queue.
+func (pool *WorkerPool) Register(addr string, nodename string) (bool, error) {
```
nwt, line 42: Document the bool return parameter.

ppl/zqd/recruiter/recruiter_test.go
```diff
@@ -10,7 +10,9 @@ import (
 
 func register1(t *testing.T, addr string, nodename string, fp int, np int, rp int) *WorkerPool {
 	wp := NewWorkerPool()
-	wp.Register(addr, nodename)
+	b, err := wp.Register(addr, nodename)
```
nwt, line 13: b is usually a []byte. How about registered?

ppl/zqd/recruiter/recwork.go
```diff
@@ -0,0 +1,63 @@
...
+	"net"
+	"os"
+	"strings"
+
+	"github.com/brimsec/zq/api"
+	"github.com/brimsec/zq/api/client"
+	"github.com/brimsec/zq/proc"
+)
+
+func RecruitWorkers(pctx *proc.Context, workerCount int) ([]string, error) {
```
nwt, line 15: Why not `ctx context.Context`?

api/api.go
```diff
@@ -42,10 +42,11 @@ type TaskEnd struct {
...
 type SearchRequest struct {
-	Space SpaceID         `json:"space" validate:"required"`
-	Proc  json.RawMessage `json:"proc,omitempty"`
-	Span  nano.Span       `json:"span"`
-	Dir   int             `json:"dir" validate:"required"`
+	Space   SpaceID         `json:"space" validate:"required"`
+	Proc    json.RawMessage `json:"proc,omitempty"`
+	Span    nano.Span       `json:"span"`
+	Dir     int             `json:"dir" validate:"required"`
+	Workers int             `json:"workers"`
```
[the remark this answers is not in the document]
alfred-landrum: I agree that it shouldn't be part of the public search api. 

@marktwallace : for this PR, I suggest you define a const "maxWorkersPerRequest" somewhere with a largish number, and have the root send that to the recruiter? In the future, we'll work on limiting the number requested based on analyzing the query & the data set.

api/client/connection.go
```diff
@@ -402,6 +414,51 @@ func (c *Connection) LogPostWriter(ctx context.Context, space api.SpaceID, opts
...
+func (c *Connection) Unreserve(ctx context.Context, req api.UnreserveRequest) (*api.UnreserveResponse, error) {
+	resp, err := c.Request(ctx).SetBody(req).SetResult(&api.UnreserveResponse{}).Post("/recruiter/unreserve")
+	if err != nil {
+		return nil, err
+	}
+	return resp.Result().(*api.UnreserveResponse), nil
+}
+
+// WorkersStats wraps the rechandlers.go handleWorkersStats API
+func (c *Connection) WorkersStats(ctx context.Context) (*api.WorkersStatsResponse, error) {
```
alfred-landrum, line 454: Should this be called `RecruiterStats`?

ppl/zqd/search/search.go
```diff
@@ -80,9 +81,11 @@ func (s *SearchOp) Run(ctx context.Context, store storage.Storage, output Output
...
 	case *archivestore.Storage:
 		return driver.MultiRun(ctx, d, s.query.Proc, zctx, st.MultiSource(), driver.MultiConfig{
-			Span:      s.query.Span,
-			StatsTick: statsTicker.C,
-			Order:     zbuf.OrderDesc,
+			Span:            s.query.Span,
+			StatsTick:       statsTicker.C,
+			Order:           zbuf.OrderDesc,
+			Parallelism:     s.workers,
+			DistributedExec: (s.workers > 0),
```
alfred-landrum, line 88: If workers is removed from the api (as suggested above), then DistributedExec would be set based on the zqd personality.

ztests/suite/zqd/rec-curl.yaml
```diff
@@ -0,0 +1,44 @@
+script: |
```
alfred-landrum, line 1: For this file and the others under suite/zqd, would you please use `recruiter` instead of `rec` in their file names?

ztests/suite/zqd/recroot.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+#
+# For test scenario with:
+# (1) a zqd recruiter process, 
+# (2) multiple zqd worker processed that register with the recruiter
+# (3) a zqd root process that recruits workers and performs a distributed query.
+#
```
alfred-landrum, line 7:
```suggestion
# This script creates:
# (1) a zqd recruiter process
# (2) multiple zqd worker processes that register with the recruiter
# (3) a zqd root process that recruits from the workers and performs a distributed query
#
```

ppl/zqd/recruiter/README.md
```diff
@@ -0,0 +1,82 @@
+# zqd recruiter
+
```
alfred-landrum, line 2: @marktwallace : I filed https://github.com/brimsec/zq/issues/1755 to capture the changes we've discussed a couple of times on the protocol.

charts/zqd/templates/deployment.yaml
```diff
@@ -52,7 +62,23 @@ spec:
...
+            value: "{{ .Values.RecruiterAddr }}"
+          - name: ZQD_POD_IP
+            valueFrom:
+              fieldRef:
+                fieldPath: status.podIP
+          - name: ZQD_NODE_NAME
+            valueFrom:
+              fieldRef:
+                fieldPath: spec.nodeName
+          - name: ZQD_WORKER_ONCE
```
alfred-landrum, line 76: I don't see this used anywhere?

charts/zqd/templates/deployment.yaml
```diff
@@ -52,7 +62,23 @@ spec:
               path: /status
               port: http
           env:
-          ##- if eq .Values.serviceAccount.name "" ##
+          {{- if eq .Values.personality "worker"}}
+          - name: ZQD_REGISTER
```
alfred-landrum, line 66:
```suggestion
          - name: ZQD_RECRUITER_ADDR
```

ppl/zqd/handlers.go
```diff
@@ -132,7 +132,22 @@ func handleSearch(c *Core, w http.ResponseWriter, r *http.Request) {
...
+		// No point in responding with the error back to zqd root process,
+		// since this is happening on the cleanup after the search is finished.
+		c.logger.Warn(
+			"WorkerReleaseError",
+			zap.String("error", err.Error()),
+		)
+	}
+	c.logger.Info(
+		"WorkerReleaseWithRegister",
+	)
```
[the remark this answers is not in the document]
marktwallace: I added this because I want "info" level logging of the release request. I will probably pull it out later after the load balancing and autoscaling systems are stable and trouble-free.

api/api.go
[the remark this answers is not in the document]
marktwallace: I will add my reason for including it: ultimately, the resources committed to a query must be based on the end user's intent. At some point, the user must specify their desired resource commitment, e.g. signing up for a plan;  specifying preferences for a session. Information about desired parallelism will vary based on the origin of the query, so should be carried along with it. 
Queries to the same cluster will have reasons to have different levels of resource commitment.

ppl/cmd/zqd/listen/command.go
```diff
@@ -142,6 +138,13 @@ func (c *Command) Run(args []string) error {
 	if err := srv.Start(ctx); err != nil {
 		return err
 	}
+	// Workers should registerWithRecruiter as late as possible,
+	// just before writing Port file for tests.
+	if c.conf.Personality == "worker" {
+		if err := core.WorkerRegistration(ctx, srv.Addr()); err != nil {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
marktwallace: OK, I will move it to NewCore.

api/recapi.go
```diff
@@ -0,0 +1,44 @@
+package api
```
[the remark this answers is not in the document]
marktwallace: ok

ppl/zqd/rechandlers.go
```diff
@@ -0,0 +1,95 @@
+package zqd
```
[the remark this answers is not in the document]
marktwallace: ok

api/api.go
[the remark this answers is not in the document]
marktwallace: This is not intended to be part of the public API. My plan is that the API server will gather information about desired threads from something about the users profile (e.g. whether they use a free or a premium plan). Then the API server will pass that information to the zqd root process, which will use it to request workers from the                      recruiter. This actually need to be a request parameter, not a global. It just looks confusing now because the API server has not been developed and deployed yet. 

It might clarify the issue if I introduce a "/root/request" route in core and only allow the "Workers" parameter for that request.

ppl/cmd/zqd/listen/command.go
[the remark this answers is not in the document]
marktwallace: OK, I checked again, and now I remember why this cannot be moved to the NewCore function. It depends on srv.Addr() that is only available after the server has been been created.

api/client/connection.go
marktwallace: yes

ppl/zqd/search/search.go
marktwallace: It is OK to request that a zqd root worker not use distributed execution. Leaving workers set to 0 will ensure that the root process executes the query directly.

api/client/connection.go
```diff
@@ -61,6 +61,11 @@ func NewConnectionTo(hostURL string) *Connection {
 	return newConnection(client)
 }
 
+// ClientHostURL allows us to print the host in log messages and internal error messages
+func (c *Connection) ClientHostURL() string {
+	return c.client.HostURL
+}
+
```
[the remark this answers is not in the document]
marktwallace: I had inadventently deleted the logger call that used it. I added the logging back in.

ppl/zqd/handlers.go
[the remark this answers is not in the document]
marktwallace: removed now

api/api.go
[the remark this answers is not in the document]
marktwallace: I removed Workers from the public search API. The approach I took was to more fully develop the /worker API. Now there are two types of /worker requests:
/worker/chunksearch  is like the old /worker request (it includes chunk details)
/worker/rootsearch will be the internal call from the API server to the root worker. It includes number_to_recruit.
I expect more information will be added to /worker/rootsearch as we build out the cloud product experience.

api/worker.go
```diff
@@ -0,0 +1,12 @@
...
+
+type WorkerChunkRequest struct {
+	SearchRequest
+	ChunkPaths []string `json:"chunk_paths"`
+	DataPath   string   `json:"data_path"`
+}
+
+type WorkerRootRequest struct {
+	SearchRequest
+	NumberOfWorkers int `json:"number_of_workers"`
```
alfred-landrum, line 11:
```suggestion
	MaxWorkers int `json:"max_workers"`
```

charts/zqd/templates/deployment.yaml
```diff
@@ -52,7 +62,21 @@ spec:
               path: /status
               port: http
           env:
-          ##- if eq .Values.serviceAccount.name "" ##
+          {{- if eq .Values.personality "worker"}}
+          - name: ZQD_REGISTER_ADDR
```
alfred-landrum, line 66: How about delete this, and just rely on `ZQD_RECRUITER_ADDR` as configured below?

charts/zqd/templates/deployment.yaml
```diff
@@ -52,7 +62,21 @@ spec:
...
+          - name: ZQD_POD_IP
+            valueFrom:
+              fieldRef:
+                fieldPath: status.podIP
+          - name: ZQD_NODE_NAME
+            valueFrom:
+              fieldRef:
+                fieldPath: spec.nodeName
+          {{- else if eq .Values.personality "root" }}
+          - name: ZQD_RECRUIT
```
alfred-landrum, line 77:
```suggestion
          - name: ZQD_RECRUITER_ADDR
```

ppl/zqd/recruiter.go
```diff
@@ -0,0 +1,95 @@
+package zqd
+
+// system test with: make TEST=TestZq/ztests/suite/zqd/rec-curl
```
alfred-landrum, line 3: Delete or update

driver/parallel.go
```diff
@@ -175,9 +178,9 @@ func (pg *parallelGroup) doneSource(sc ScannerCloser) {
 	delete(pg.scanners, sc)
 }
 
-// sourceToRequest takes a Source and converts it into a WorkerRequest
-func (pg *parallelGroup) sourceToRequest(src Source) (*api.WorkerRequest, error) {
-	var req api.WorkerRequest
+// sourceToRequest takes a Source and converts it into a WorkerChunkRequest
```
nwt, line 181:
```suggestion
// sourceToRequest takes a Source and converts it into a WorkerChunkRequest.
```

ztests/suite/zqd/recruiter-root.sh
```diff
@@ -0,0 +1,56 @@
...
+# (3) start a zqd "root" process (default personality for now) that may recruit workers
+ZQD_RECRUITER_ADDR=localhost:$(cat $portdir/zqd-rec) \
+zqd listen -l=localhost:0 -portfile="$portdir/zqd-root" -data="$zqdroot" &> zqd-root.log &
+zqdrootpid=$!
+awaitfile $portdir/zqd-root
+export ZQD_HOST=localhost:$(cat $portdir/zqd-root)
+
+trap "rm -rf $portdir; kill -9 $zqdrpid $zqdw1pid $zqdw2pid $zqdrootpid" EXIT
+
+
```
nwt, line 56:
```suggestion
```

charts/zqd/values.yaml
```diff
@@ -3,12 +3,15 @@
 # Declare variables to be passed into your templates.
 
 AWSRegion: ""
+
+personality: "standalone"
```
nwt, line 7: We don't have a standalone personality, so should this be all?

ppl/cmd/zqd/listen/command.go
```diff
@@ -141,7 +144,13 @@ func (c *Command) Run(args []string) error {
 	// Workers should registerWithRecruiter as late as possible,
 	// just before writing Port file for tests.
 	if c.conf.Personality == "worker" {
-		if err := core.WorkerRegistration(ctx, srv.Addr()); err != nil {
+		if _, _, err := net.SplitHostPort(c.conf.Recruiter); err != nil {
```
alfred-landrum, line 147: Could you move these checks (the c.conf.Recruiter and c.conf.SpecNodeName) into WorkerRegistration?

ppl/zqd/core.go
```diff
@@ -32,10 +32,14 @@ const indexPage = `
...
 type Config struct {
-	Logger      *zap.Logger
-	Personality string
-	Root        string
-	Version     string
+	Logger       *zap.Logger
+	Personality  string
+	Recruiter    string
+	Root         string
+	SpecPodIP    string
```
alfred-landrum, line 39: The IP and Node are only used by workers, and I think the code would be easier to understand if their names reflect their usage, instead of their particular k8s sources.

How about creating WorkerConfig struct, with field names like "Address" and "Node", then making adding that into the overall Config struct?

charts/zqd/templates/deployment.yaml
```diff
@@ -62,13 +63,11 @@ spec:
               path: /status
               port: http
           env:
-          - name: ZQD_RECRUITER_ADDR
-            value: "{{ .Values.RecruiterAddr }}"
-          - name: ZQD_POD_IP
+          - name: SPEC_POD_IP
```
alfred-landrum, line 66: Suggest making this "STATUS_POD_IP".

alfred-landrum (commented):
thanks for making this change, I just have some suggestions on naming & where the config lives.

cmd/zapi/cmd/get/get.go
```diff
@@ -113,18 +115,33 @@ func (c *Command) Run(args []string) error {
...
+			}
+			r, err = conn.WorkerRootSearch(c.Context(), *rootWorkerReq, params)
+			if err != nil {
+				return fmt.Errorf("distributed search error: %w", err)
+			}
+		} else {
+			r, err = conn.SearchRaw(c.Context(), *req, params)
+			if err != nil {
+				return fmt.Errorf("search error: %w", err)
+			}
```
nwt, line 132:
```suggestion
		if c.workers > 1 {
			rootWorkerReq := &api.WorkerRootRequest{
				SearchRequest: *req,
				MaxWorkers:    c.workers,
			}
			r, err = conn.WorkerRootSearch(c.Context(), *rootWorkerReq, params)
		} else {
			r, err = conn.SearchRaw(c.Context(), *req, params)
		}
		if err != nil {
			return fmt.Errorf("search error: %w", err)
		}
```

ppl/cmd/zqd/listen/command.go
```diff
@@ -39,7 +37,7 @@ var Listen = &charm.Spec{
 	Long: `
 The listen command launches a process to listen on the provided interface and
 `,
-	HiddenFlags: "brimfd,workers",
+	HiddenFlags: "brimfd,workers,recruiter,podip,nodename",
```
nwt, line 40: Keep these sorted.

ppl/cmd/zqd/listen/command.go
```diff
@@ -81,12 +77,15 @@ func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
 	f.BoolVar(&c.devMode, "dev", false, "run in development mode")
 	f.StringVar(&c.listenAddr, "l", ":9867", "[addr]:port to listen on")
 	f.Var(&c.logLevel, "loglevel", "logging level")
-	f.StringVar(&c.conf.Personality, "personality", "all", "server personality (all, apiserver, or worker)")
+	f.StringVar(&c.conf.Personality, "personality", "all", "server personality (all, apiserver, recruiter, or worker)")
 	f.StringVar(&c.portFile, "portfile", "", "write listen port to file")
 	f.BoolVar(&c.pprof, "pprof", false, "add pprof routes to API")
+	f.StringVar(&c.conf.Recruiter, "recruiter", "", "Recruiter address for worker registration")
```
nwt, line 83:
```suggestion
	f.StringVar(&c.conf.Recruiter, "recruiter", "", "recruiter address for worker registration")
```

ppl/zqd/core.go
marktwallace: I will introduce a type WorkerConfig with {Host, Node}
We have used address for strings of the form host:port, but the SpecPodIP is just the host portion, in either IP format or as a dns address.
SpecNodeName will only be used an a logical identifier by us, even though it can have the form of a dns address, because it is not the in-cluster dns address.

nwt: While we're here, is plumbing these through worth the effort?

When a worker registers with the recruiter, its IP address is already available to the recruiter via http.Request.RemoteAddr, so it doesn't actually need to be include in the request body. And if we want to include it there anyway, we can get it from httpd.Server.LocalAddr.

As for the node name, that can always be looked up using the worker's IP address.

ppl/zqd/recruiter/workerreg.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+type WorkerReg struct {
+	conn          *client.Connection
+	recruiteraddr string
+	selfaddr      string
+	nodename      string
+}
+
+func NewWorkerReg(ctx context.Context, srvAddr string, recruiteraddr string, specPodIP string, specNodeName string) (*WorkerReg, error) {
```
nwt, line 20: specPodIP AND specNodName deserve better names.

But is either parameter necessary? I don't think there's any scenario in which srvAddr differs from the pod IP address, and the node name can be looked up using the pod IP.

driver/compile.go
```diff
@@ -92,12 +87,15 @@ func compile(ctx context.Context, program ast.Proc, zctx *resolver.Context, read
 
 type MultiConfig struct {
 	Custom      compiler.Hook
+	Distributed bool // true if remote request specified worker count
```
nwt, line 90: Do we need this, or could we just check for `Recruiter != "" || Workers != nil`?

ppl/zqd/recruiter/workerreg.go
nwt: Let's address this later.

driver/compile.go
nwt: Let's address this later.

nwt (approved):

PR #1697, opened 2020-11-24
Remove zeekio.ZeekStrings

nwt:
textio.Writer and zeekio.Writer are better without it.

### DIFF
--- a/zio/textio/writer.go
+++ b/zio/textio/writer.go
@@ -1,12 +1,10 @@
 package textio
 
 import (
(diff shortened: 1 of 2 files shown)


alfred-landrum (commented):
The code looks good to me. I did a search for `-f text` looking for textio ztests, and only found a single occurrence. Could you add a ztest that excercises some of the zng.OutFmt options?

nwt:
I'll reopen this when I have time to work on it.

PR #1703, opened 2020-11-29
ZQL docs example showing "by" grouping with non-present fields

philrz:
When doing Suricata testing I recently tripped myself up when I was trying to count all the Zeek+Suricata events in a single `by` grouping (`count() by _path,event_type`) and it produced no results. Once I thought about it and ran a few additional ZQL variations, I realized how that could be explained as expected behavior. But since it wasn't immediately obvious to me and I suspect it might trip up others (or I might forget before next time I attempt thios 😉 ) here I've added a docs example that attempts to explain it.

### DIFF
--- a/zql/docs/grouping/README.md
+++ b/zql/docs/grouping/README.md
@@ -197,8 +197,8 @@ zq -t 'count() by _path,typeof(.) | sort -r' http.log.gz dns.log.gz
 
 A way to achieve this would be to use the
 [`fuse` processor](../processors/README.md#fuse) to unite the `http` and `dns`
-events under a single schema. This has the effect of populating previously
-non-present fields with null values. Now that the named fields are present in
+events under a single schema. This has the effect of populating missing
+fields with null values. Now that the named fields are present in
 all records, the `by` grouping has the desired effect.
 
 ```zq-command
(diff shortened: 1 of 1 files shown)


zql/docs/grouping/README.md
```diff
@@ -138,6 +138,81 @@ ID.RESP_H       NUM_ANSWERS COUNT
...
+0:[dns;record[_path:string,ts:time,uid:bstring,id:record[orig_h:ip,orig_p:port,resp_h:ip,resp_p:port],proto:zenum,trans_id:uint64,rtt:duration,query:bstring,qclass:uint64,qclass_name:bstring,qtype:uint64,qtype_name:bstring,rcode:uint64,rcode_name:bstring,AA:bool,TC:bool,RD:bool,RA:bool,Z:uint64,answers:array[bstring],TTLs:array[duration],rejected:bool];53615;]
+```
+
+A way to achieve this would be to use the
+[`fuse` processor](../processors/README.md#fuse) to unite the `http` and `dns`
+events under a single schema. This has the effect of populating previously
+non-present fields with null values. Now that the named fields are present in
```
nwt, line 201:
```suggestion
events under a single schema. This has the effect of populating missing
fields with null values. Now that the named fields are present in
```

nwt (approved):

PR #1715, opened 2020-11-30
zson spec

### LINKED ISSUE #1662: pretty zng - aka ZSON
We went to great lengths in the zng architecture to create a human-readable text form (tzng) suitable for hand editing, tests, and demo.  This is especially important because understanding the binary zng format takes developer-level commitment to get into it and understand the details.

But for broader audiences (and hence broader acceptance), tzng doesn't have the greatest ergonomics.  The syntax is compact and cryptic and the type structure is separate from the DFS-ordered list of values so it's not the best format to illustrate the zng data model.  And zjson suffers from this too.

Also, when we first developed tzng, we were concerned about parsing performance.  We've now come to realize that tzng is never used in a performance-sensitive context so here we are now focused on ergonomics over perform-ity.

The  intent is to adopt a "pretty" version of zng that would have broader appeal and be easier to understand.  I will leave it open as to whether pretty-zng replaces or augments text-zng (though I am now leaning toward replacement).

We want it to have some overlap in the look and feel of JSON without its shortcomings.  And, of course, needs to represent the full gamut of zng binary capabilities.

## Proposal

Here is a proposal open for discussion.

### Whitespace

All whitespace is equivalent so arbitrary pretty printing is possible.  There is no notion of newline-terminated boundaries so values must be parsed to find their end.

### Typedefs

Types are similar to tzng but can have arbitrary names.  Types are optional as values can also be fully specified with embedded or implied types.  A typedef has the form:
```.
type <name>=<type>
```
where name is an identifier or an integer and `<type>` is a zng/zql style type definition.  (Type strings are not yet implemented in zql but they will be.). Type names can be redefined in a stream so that files can be concatenating and still be correct.

An alias has a similar form:
```.
alias <name>=<type>
```
The difference between an alias and a type is that aliases are encoded into the corresponding zng input/output as zng aliases, where type names are just a convenience tag in pretty-zng to refer to a native zng type, which are localized to the pretty-zng file.

### Implied-type Values

There are 11 "Implied-type" values:

* **string**  a double-quoted string literal
* **int64** as in an integer text string, e.g., 123, -1
* **float64** as in an float text string, e.g., 123.0 or -1e6
* **time** unquoted string that parses as an ISO time
* **bool** unquoted identifier `true` or `false`
* **ip** unquoted numeric IP address (i.e., format parseable by go library)
* **net** unquoted numeric IP subnet address in cidr form (i.e., format parseable by go library)
* **array** as in a javascript style array where the type is implied by the elements or if the types are mixed then an array of unions
* **record** as in javascript-style object ''{ ... }'  notation where the record type of the object is implied by the key names and types of the alements
* **map** syntax TBD
* **null** the unquoted identifier `null`

_Note: I believe all of the above values are unambiguously parseable like the zql parser understanding that `10.0.0.1` is an IP but `"10.0.0.1"` is a string._

So here are some example implied-type values with their implied types:
```
"hello, world" -> string
123 -> int64
192.168.1.1 -> ip
[ 10, 11, 12 ] -> array[int64]
[ 10, "hello", 1.0, 2 ] -> array[union[int64,string,float64]]
[ "hello", "world" ]  -> array[string]
{ foo:10, bar:"hello, word" } -> record[foo:int64,bar:string]
```

### Explicit-type Values

To be able to represent all of the types of zng, values can be explicitly typed using the zql type case syntax and a prefix syntax for records and arrays.

* **Primitive values** have the form `<value>:<type>` where values it the tzng value and type is the type, e.g., uint8, bytes, etc.
* **Enum** has the form `<value>:<type>`
* **Union** has the form `<value>:<type><selector>` where type is a value and selector the integer position of the type in the union
* **Array** has the form `[ elements... ]:<type>`
* **Record** has the form `{ elements... }:<type>`

#### Ex 1
```
#0:record[s:string]
0:[hello,world;]
```
is the same as
```
{s:"hello,world"}
```

#### Ex 2
```
#0:record[ts:time,addr:ip,msg:string]
0:[1;10.0.0.1;hello world;]
0:[2;192.168.1.1;here is a message;]
0:[3;10.0.0.2; here is another message;]
```
is the same as
```
{ ts:1970-01-01T00:00:00.000000001Z, addr:10.0.0.1, msg:"hello, world" }
{ ts:1970-01-01T00:00:00.000000002Z, addr:192.168.1.1, msg:"here is a message" }
{ ts:1970-01-01T00:00:00.000000003Z, addr:10.0.0.2, msg:"here is another message" }
```
is the same as
```
type log = record[ts:time,addr:ip,msg:string]
{ 1970-01-01T00:00:00.000000001Z, 10.0.0.1, "hello, world" }:log
{ 1970-01-01T00:00:00.000000002Z, 192.168.1.1, "here is a message" }:log
{ 1970-01-01T00:00:00.000000003Z, 10.0.0.2, "here is another message" }:log
```
is the same as
```
type log = record[ts:time,addr:ip,msg:string]
{ 1970-01-01T00:00:00.000000001Z, 10.0.0.1, "hello, world" }:log
{ ts:1970-01-01T00:00:00.000000002Z, addr:192.168.1.1, msg:"here is a message" }
{ 1970-01-01T00:00:00.000000003Z, 10.0.0.2, "here is another message" }:log
```

#### Ex 3

```
#port=uint16
#0:record[id:record[orig_h:ip,orig_p:port]]
0:[[10.0.0.1;8080]]
```
is the same as
```
alias port=uint16
{id: {orig_h:10.0.0.1, orig_p:8080:port}}
```
is the same as
```
alias port=uint16
type zeekID = record[orig_h:ip,orig_p:port]
{ id:{10.0.0.1,8080}:zeekID }
```

#### Ex 4

```
#vector=array[uint16]
#0:record[x:int32,point:vector]]
0:[1;[2;3;4;]]
0:[2;[23;18;]]
0:[3;[99;1024;0;1;-6;]]
```
is the same as
```
alias vector=array[uint16]
type pointVector=record[x:int32,point:vector]
{ 1, [ 2, 3, 4 ] }:pointVector
{ 2, [ 23, 18 ] }:pointVector
{ 3, [ 99, 1024, 0, 1, -6 ] }:pointVector
```

mccanne:
This commit adds a markdown description of the new ZSON format.
References to tzng have been removed from the ZNG spec and replaced
with references to the ZSON spec.  The language in the ZNG spec
has been further simplified a bit assuming knowledge of the ZSON
data model described in the new spec.

The zsonio writer has been improved to fix bugs and follow a few
changes that happend to the ZSON format.

We also made a first cut at updating the Zeek compat doc but we need
to have the ZSON reader before this can be finished.  We should also
consider explaining compat in terms of ZSON instead of ZNG.

Closes #1662 #1678


zio/zsonio/writer.go
```diff
@@ -180,7 +156,7 @@ func (w *Writer) writeRecord(indent int, typ *zng.TypeRecord, bytes zcode.Bytes)
 	return w.indent(indent-w.tab, "}")
 }
 
-func (w *Writer) writeVector(indent int, open, close string, inner zng.Type, zv zng.Value) error {
+func (w *Writer) writeVector(indent int, open, close string, inner zng.Type, zv zng.Value, known bool) error {
```
alfred-landrum, line 159: `known` is not used in this function, should it be?

zng/docs/zson.md
```diff
@@ -0,0 +1,629 @@
...
+| `int16`    | decimal string representation of any signed, 16-bit integer   |
+| `int32`    | decimal string representation of any signed, 32-bit integer   |
+| `int64`    | decimal string representation of any signed, 64-bit integer   |
+| `duration` | a _duration string_ representing signed 64-bit nanoseconds |
+| `time`     | an RFC-3339 UTC data/time string representing signed 64-bit nanoseconds from epoch |
+| `float16`  | a _point string_ representing an IEEE-754 binary16 value |
+| `float32`  | a _point string_ representing an IEEE-754 binary32 value |
+| `float64`  | a _point string_ representing an IEEE-754 binary64 value |
+| `decimal`  | a _point string_ representing an IEEE-754 decimal128 value |
+| `bool`     | the string `true` or `false` |
```
philrz, line 287: FYI, right now the implementation is putting `T` and `F` for these, so I guess one of them needs to change.

```
$ echo '{"foo": true}' | zq -f zson -
{
    foo: T
} (=23)
```

zng/docs/zson.md
```diff
@@ -0,0 +1,629 @@
...
+```
+is an enum value of the indicated enum and is not the boolean value `true`.
+
+#### Map Value
+
+A map value has the following syntax:
+```
+|{ {<key>, <value>}, {<key>, <value>}, ... }|
+```
+A type decorator applied to an map can either be one element, referring to a
```
philrz, line 474:
```suggestion
A type decorator applied to a map can either be one element, referring to a
```

zng/docs/spec.md
```diff
@@ -602,326 +510,47 @@ key/value pair).
...
-## 5. Primitive Types
+## 3. Primitive Types
 
 For each ZNG primitive type, the following table describes:
-* The predefined ID, which need not be defined in [ZNG Typedefs](#311-typedefs)
-* How a typed `value` of length `N` is interpreted in a [ZNG Value Message](#32-value-messages)
-* The format of a UTF-8 string representing a [TZNG Value](#432-value-syntax) of that type
+* The predefined ID, which need not be defined in [ZNG Typedefs](#211-typedefs)
+* How a typed `value` of length `N` is interpreted in a [ZNG Value Message](#22-value-messages)
+* The format of a UTF-8 string representing a [ZSON Value](#xxx-zson-tag) of that type
```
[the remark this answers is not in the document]
mccanne: Actually, I'll just take this out as it's no longer relevant (it was used by the TZNG description which is now deleted).

zng/docs/spec.md
```diff
@@ -602,326 +510,47 @@ key/value pair).
...
+| `ip`       | 18 | 4 or 16  | 4 or 16 bytes of IP address                    |
+| `net`      | 19 | 8 or 32  | 8 or 32 bytes of IP prefix and subnet mask     |
+| `type`     | 20 | variable | UTF-8 byte sequence ZSON type value [zson type value](#42-xxx)  |
+| `error`    | 21 | variable | UTF-8 byte sequence of string of error message |
+| `null`     | 22 |    0     | No value, always represents an undefined value |
+
+> TBD: For decimal, ZSON assumes decimal128 and there's not a way to specify
+> different widths.  This should be okay since a 128 will hold the others.
+> For duration, ZSON says RFC-3339 but there is no way to specify negative
+> durations in RDC-3339 but ZNG is and zq are perfectly happy with them.
```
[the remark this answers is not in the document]
mccanne: Taking this out since I changed ZSON to use Go-style durations (which can be negative) instead of RFC-3339 durations.

zng/docs/zeek-compat.md
```diff
@@ -75,11 +75,36 @@ $ cat zeek_types.log
 #types	bool	count	int	double	time	interval	string	string	port	addr	subnet	enum	set[string]	vector[string]	string	count
 T	123	456	123.4560	1592502151.123456	123.456	smile\xf0\x9f\x98\x81smile	\x09\x07\x04	80	127.0.0.1	10.0.0.0/8	tcp	things,in,a,set	order,is,important	Jeanne	122
 
-$ zq -t zeek_types.log 
-#port=uint16
-#zenum=string
-#0:record[my_bool:bool,my_count:uint64,my_int:int64,my_double:float64,my_time:time,my_interval:duration,my_printable_string:bstring,my_bytes_string:bstring,my_port:port,my_addr:ip,my_subnet:net,my_enum:zenum,my_set:set[bstring],my_vector:array[bstring],my_record:record[name:bstring,age:uint64]]
-0:[T;123;456;123.456;1592502151.123456;123.456;smile😁smile;\x09\x07\x04;80;127.0.0.1;10.0.0.0/8;tcp;[a;in;set;things;][order;is;important;][Jeanne;122;]]
+$ zq -f zson zeek_types.log
```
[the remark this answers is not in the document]
mccanne: Yes, sounds good.  But don't want to break this just yet... a million tests depend on it.

zng/docs/zson.md
```diff
@@ -0,0 +1,629 @@
...
+>
+> ZSON is intended as a literal superset of JSON syntax (not semantics).
+> If there is anything in this document that leads to this not being the
+> case, please let us know and we will address it.
+>
+> Regarding the state of the implementation of the ZSON format in zq:
+> * An input reader is not yet implemented,
+> * End-of-sequence markers are not yet conveyed in the writer.
+
+> TBD: add a ToC and section numbering
```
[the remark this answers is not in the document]
mccanne: Yes, I would prefer to merge with this TBD and add the ToC when it feels like the dust has settled.

zng/docs/zson.md
```diff
@@ -0,0 +1,629 @@
...
+
+The simplest ZSON text is a single value, perhaps a string like this:
+```
+"hello, world"
+```
+There's no need for a type decorator here.  It's explicitly a string.
+
+A structured SQL table might look like this:
+```
+{ city: "Berkeley", state: "CA", population:121643 (uint32) } (=schema)
```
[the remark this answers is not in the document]
mccanne: Ok, I'll add the space like the zson writer outputs.  Following the lead of jq here... at some point we could even add terminal color coding of the text :-).  I think for the type values, I will omit the space.  It's a different sort of thing, but we can change that later if we don't like it.

zng/docs/zson.md
```diff
@@ -0,0 +1,629 @@
...
+complex type has been defined previously, there is no need to specify the
+type decorators for each nested field since they are all determined by the
+top-level record type.
+
+The subsequent value defines a type called `access_list`.  In this case,
+the `nets` field is an array of networks and illustrates the helpful range of
+primitive types in ZSON.  Note that the syntax here implies
+the type of the array, as its inferred from the type of the elements.
+
+Finally, there are four more values that show the ZSON's efficacy for
```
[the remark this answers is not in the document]
mccanne: Do you get the Internet with the ZSON?  :-)

zng/docs/zson.md
```diff
@@ -0,0 +1,629 @@
...
+| `uint8`    | decimal string representation of any unsigned, 8-bit integer  |
+| `uint16`   | decimal string representation of any unsigned, 16-bit integer |
+| `uint32`   | decimal string representation of any unsigned, 32-bit integer |
+| `uint64`   | decimal string representation of any unsigned, 64-bit integer |
+| `int8`     | decimal string representation of any signed, 8-bit integer    |
+| `int16`    | decimal string representation of any signed, 16-bit integer   |
+| `int32`    | decimal string representation of any signed, 32-bit integer   |
+| `int64`    | decimal string representation of any signed, 64-bit integer   |
+| `duration` | a _duration string_ representing signed 64-bit nanoseconds |
+| `time`     | an RFC-3339 UTC data/time string representing signed 64-bit nanoseconds from epoch |
```
nwt, line 282: Space instead of dash per https://tools.ietf.org/html/rfc7322#section-3.5.
```suggestion
| `time`     | an RFC 3339 UTC data/time string representing signed 64-bit nanoseconds from epoch |
```

zng/docs/zson.md
```diff
@@ -0,0 +1,629 @@
...
+### Primitive Types
+
+Primitive types include signed and unsigned integers, IEEE floating point of
+several widths, IEEE decimal,
+string, bstring, byte sequence, boolean, IP address, and IP network.
+
+> Note: The `bstring` type is an unusual type representing a hybrid type
+> mixing a UTF-8 string with embedded binary data.  This type is
+> useful in systems that, for instance, pull data off the network
+> while expecting a string, but there can be embedded binary data due to
```
nwt, line 140:
```suggestion
> useful in systems that, for instance, pull data off the network
> while expecting a string but must tolerate embedded binary data due to
```

zng/docs/zson.md
```diff
@@ -0,0 +1,629 @@
...
+### The Null Type
+
+The _null_ type is a primitive type representing only a `null` value.  A `null`
+value can have any type.
+
+Any value in ZSON can take on a null representation.  It is up to an
+implementation to decide how external data structures map into and
+out of values with nulls.  Typically, a null value means either the
+zero value or in the case of record fields, it means the field is optional
+and the value is not present, though these semantics are not explicitly
```
nwt, line 173:
```suggestion
zero value or, in the case of record fields, an optional field whose
value is not present, though these semantics are not explicitly
```

zng/docs/zson.md
```diff
@@ -0,0 +1,629 @@
...
+defined by ZSON.
+
+### Type Definitions and Contexts
+
+Type definitions embedded in the data sequence associate a name with a type
+so that later values can refer to their type by name instead of explicitly
+enumerating the type of every element.  A collection of mappings from name to
+type is called a "type context."  As bindings are read from a sequence, they
```
nwt, line 181: I think this reads better with consistent use of "binding" terminology.
```suggestion
Type definitions embedded in the data sequence bind a name to a type
so that later values can refer to their type by name instead of explicitly
enumerating the type of every element.  A collection of bindings from name to
type is called a "type context."  As bindings are read from a sequence, they
```

zng/docs/zson.md
```diff
@@ -0,0 +1,629 @@
...
+create the sequence's type context.  The context may be reset at any point in
+the sequence, implying that a new binding must be defined from that point on
+for each unique use of a type.
+
+A type name is either internal or external.  Internal names are used exclusively
+to organize the types in the type context within the data sequence itself and have
+no meaning outside of that data.  External names are visible outside of
+data sequence as named types providing external systems a way to
+discover and refer to types by name where the type is defined within the data
+sequence.  The binding between an external name and its type must be reestablished
```
nwt, line 191:
```suggestion
no meaning outside of that data.  External names are visible outside of
a data sequence as named types, providing external systems a way to
discover and refer to types defined within the data
sequence.  The binding between an external name and its type must be reestablished
```

zng/docs/zson.md
```diff
@@ -0,0 +1,629 @@
...
+in each type context in which it used.  External names are analogous to the notion
+of "logical types" in Parquet and Avro.
+
+Internal type names are represented by integer names while external names are
```
nwt, line 195:
```suggestion
Internal type names are represented by integers while external names are
```

zng/docs/zson.md
```diff
@@ -0,0 +1,629 @@
...
+represented by identifiers.  When ZSON data is organized into a sequence
+comprised of two or more subsequences where each subsequence has its own
+type context, the internal names may be "reused" across type contexts to refer
+to different types but external names must have the same type value across
+subsequences.  It is a "type mismatch" error if an external name has different
+type values across subsequences.  It is also a "type mismatch" error if any
+any type binding
```
nwt, line 202: Finish me.

zng/docs/zson.md
```diff
@@ -0,0 +1,629 @@
...
+
+The subsequent value defines a type called `access_list`.  In this case,
+the `nets` field is an array of networks and illustrates the helpful range of
+primitive types in ZSON.  Note that the syntax here implies
+the type of the array, as it is inferred from the type of the elements.
+
+Finally, there are four more values that show ZSON's efficacy for
+representing metrics.  Here, there are no type decorators as all of the field
+types are implied by their syntax, and hence, the top-level record type is implied.
+For instance, the `ts` field is an RFC-3339 date/time string,
```
nwt, line 115:
```suggestion
For instance, the `ts` field is an RFC 3339 date/time string,
```

zng/docs/zson.md
```diff
@@ -0,0 +1,629 @@
...
+123 (int64)
+```
+is the same as `123`.
+
+Values that do not have implied types must include a type decorator to clarify
+its type or appear in a context for which its type is defined (i.e., as a field
+value in a record, as an element in an array, etc).
+
+> Note that `time` values correspond to 64-bit epoch nanoseconds and thus
+> not all possible RFC-3339 date/time strings are valid.  In addition,
```
nwt, line 322:
```suggestion
> not all possible RFC 3339 date/time strings are valid.  In addition,
```

zng/docs/zson.md
```diff
@@ -0,0 +1,629 @@
...
+> and the old time epoch and new time epoch can live side by side with
+> the old using a typedef for the new epoch time aliased to the old `time`.
+> An app that wants more than 64 bits of timestamp precision can always use
+> a typedef to a `bytes` type and do its own conversions to and from the
+> corresponding bytes values.
+
+#### String Escape Rules
+
+Double-quoted `string` syntax is the same as that of JSON as described
+[RFC-7159](https://tools.ietf.org/html/rfc7159), specifically:
```
nwt, line 334:
```suggestion
[RFC 7159](https://tools.ietf.org/html/rfc7159), specifically:
```

zng/docs/zson.md
mccanne: Fermat's Last Theorem.  I don't remember what I was about to say.  I will delete until it comes back to me.

zng/docs/zson.md
```diff
@@ -0,0 +1,628 @@
...
+    conn_state: "SF" (bstring),
+    local_orig: null (bool),
+    local_resp: null (bool),
+    missed_bytes: 0 (uint64),
+    history: "^dtAttttFf" (bstring),
+    orig_pkts: 57490 (uint64),
+    orig_ip_bytes: 2358856 (uint64),
+    resp_pkts: 123713 (uint64),
+    resp_ip_bytes: 185470730 (uint64),
+    tunnel_parents: null (|[bstring]| (=25))
```
henridf, line 566: I'm not following what `(|[bstring]| (=25))` means, but maybe I missed something above?

zng/docs/zson.md
```diff
@@ -0,0 +1,628 @@
...
+} (=26)
```
henridf, line 567: Oh, I guess the internal names are here for illustrative reasons

zng/docs/zson.md
```diff
@@ -0,0 +1,628 @@
...
+
+### union
+
+```
+{ u: 12 (int32, string, {a:int8,b:ip} (=27)), |[27]| } (=28) (=ex_union)
+{ u: "foo" } (=ex_union)
+```
+is the same as
+```
+{ u: 12 (int32, string, {a:int8,b:ip} (=27)), |[27]| } (=28) (=rec_with_union)
```
henridf, line 614: probably missing something but I'm not understanding  `|[27]|` in:

`{ u: 12 (int32, string, {a:int8,b:ip} (=27)), |[27]| } (=28) (=ex_union)`.

Looks like a set of `{a:int8,b:ip}`, but should't there be a field name before it?

zng/docs/zson.md
```diff
@@ -0,0 +1,630 @@
+# ZSON - ZNG Structured Object-record Notation
+
+> ### DRAFT 11/30/20
+> ### Note: This specification is in BETA development.
+> We plan to have a final form established by Spring 2021.
+>
+> ZSON is intended as a literal superset of JSON syntax (not semantics).
```
henridf, line 7: It's a superset of NDJSON 🤓

mccanne: It's a superset of both JSON and NDJSON.  I will add NDJSON.  (zq won't handle top-level values that aren't records but the marshaler will handle any value.)

zng/docs/zson.md
mccanne: Good point.  This says type `25` is a set of type `bstring` and `tunnel_parents` is a null value of that type.   I need to elaborate that definitions can be inside the decorator.  I'll make this change now and after I implement the reader, I will then have more experience and context to write a better description of the decorator syntax.

zng/docs/zson.md
henridf: Yep, JSON ⊂ NDJSON ⊂ ZSON

zng/docs/zson.md
mccanne: That is output from `-f zson` and when writing a stream you don't know ahead of time which types might be needed.  If you write ZSON by hand or had an automated analysis that looked at the entire text, you could optimize the output by omitting unneeded type defs.

zng/docs/zson.md
mccanne: Good catch.  I had a typo here and will fix... the set of records is supposed to be inside the union type.  In a future PR, I will add more examples and add discussion of the examples.

zng/docs/zson.md
henridf: Hmm, would be nice to be able to output concise zson... but later!

zng/docs/zson.md
```diff
@@ -0,0 +1,683 @@
...
+
+The simplest ZSON text is a single value, perhaps a string like this:
+```
+"hello, world"
+```
+There's no need for a type decorator here.  It's explicitly a string.
+
+A structured SQL table might look like this:
+```
+{ city: "Berkeley", state: "CA", population: 121643 (uint32) } (=schema)
```
alfred-landrum, line 85: I suggest using a different type name for this example, since `schema` could cause some confusion with readers that are thinking about 'schemas' from other data models. (`cityinfo`?)

zng/docs/zson.md
```diff
@@ -0,0 +1,683 @@
...
+* _map_ - a collection of zero or more key/value pairs where the keys are of a
+uniform type called the key type and the values are of a uniform type called
+the value type.
+
+### 2.3 The Type Type
+
+ZSON also includes first-class types, where a value can be of type `type`.
+The syntax for type values is given below.
+
+### 2.4 The Null Type
```
alfred-landrum, line 191:
```suggestion
### 2.4 The Null Type and Null Values
```

alfred-landrum (approved):

zng/docs/zson.md
mccanne: Good point.  I had thought about that too.  It is ying and yang because it is supposed to be the schema of the table but there's nothing special about the word "schema" here.   I'll add some clarifying text in this regard and change the name to `city_schema`.

zng/docs/zson.md
mccanne: This breaks the symmetry of the "data model" section describing the various data types.  The format section describes the values.  I will fix this by creating a Null Value section and breaking out the text here into that new section.

Issue #1701, opened 2020-11-26
type mismatch bug resulting in ZNG-ERR panics


mccanne:
Groupby, cut, and put all presume that the key/field expressions they use always evaluate to the same output type for a given input type and thus memoize certain data structures based on the input record type.  This is commonly true but with the addition of functions, you can have different output types for the same input type when the output type varies by input value (instead of type).  Notably this occurs when functions return errors and the error value is inserted into the field as its result.

Interestingly, groupby handles differently typed row values for a common key but not differently typed key values.

The manifestation of this bug is that a zng record is built from the bytes body presuming the output type in the memoization table (i.e., the result of Eval is not type-checked against the assumptiuon).  This causes ZNG-ERR panics when decoding the bogus zng values that don't match their stored type.

The fix is, in theory, easy.  Just keep track of the different possible output types for each input type presuming that the output types can change and handle the common case (i.e., that the output types don't change) as efficiently as the current implementation.

This solution can be implemented in a new package that is shared by groupby, cutter, and put.

p.s. it seems like we should catch and better report ZNG-ERR panics

mccanne:
I am repurposing this for groupby only and closing this with that PR.  I will create another issue for the cut bug. 
Put doesn't seem to have this problem, but maybe it should be refactored to use the common code here.

PR #1720, opened 2020-12-01
small refactor of groupby to get rid of params

mccanne:
This is a small change to the groupby proc, which moves ast- and
compiler-specific logic to proc/compiler and eliminates groupby.Params.
This is a baby step toward the compiler refactor (issue #1162)
and the bug fix for output type tracking (issue #1701).


proc/compiler/groupby.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	var where expr.Evaluator
+	if reducerAST.Where != nil {
+		where, err = expr.CompileExpr(zctx, reducerAST.Where)
+		if err != nil {
+			return nil, nil, err
+		}
+	}
+	f, err := reducer.NewMaker(reducerOp, arg, where)
+	return lhs, f, err
```
nwt, line 88: NIt: f is surprising name for a reducer.Maker.
```suggestion
	m, err := reducer.NewMaker(reducerOp, arg, where)
	return lhs, m, err
```

proc/compiler/groupby.go
```diff
@@ -0,0 +1,103 @@
...
+
+func compileReducers(assignments []ast.Assignment, zctx *resolver.Context) ([]field.Static, []reducer.Maker, error) {
+	names := make([]field.Static, 0, len(assignments))
+	reducers := make([]reducer.Maker, 0, len(assignments))
+	for _, assignment := range assignments {
+		name, f, err := compileReducer(zctx, assignment)
+		if err != nil {
+			return nil, nil, err
+		}
+		reducers = append(reducers, f)
```
nwt, line 48: NIt: f is surprising name for a reducer.Maker.
```suggestion
		name, maker, err := compileReducer(zctx, assignment)
		if err != nil {
			return nil, nil, err
		}
		reducers = append(reducers, maker)
```

proc/compiler/groupby.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+	}
+	f, err := reducer.NewMaker(reducerOp, arg, where)
+	return lhs, f, err
+}
+
+func IsDecomposable(assignments []ast.Assignment) bool {
+	zctx := resolver.NewContext()
+	for _, assignment := range assignments {
+		_, create, err := compileReducer(zctx, assignment)
```
nwt, line 94: Nit: I'd call this maker instead of create since it's a reducer.Maker.

proc/groupby/groupby.go
```diff
@@ -120,57 +119,70 @@ type Row struct {
...
 		keyRows:      make(map[int]keyRow),
 		table:        make(map[string]*Row),
 		keyCompare:   keyCompare,
 		keysCompare:  keysCompare,
 		valueCompare: valueCompare,
-		consumePart:  params.consumePart,
-		emitPart:     params.emitPart,
-	}
+		consumePart:  partialsIn,
+		emitPart:     partialsOut,
```
nwt, line 184: NIt: Use the same name for the parameter and field.

proc/groupby/groupby.go
```diff
@@ -120,57 +119,70 @@ type Row struct {
...
 	rs := expr.NewCompareFn(true, resolvers...)
-	if params.inputSortDir < 0 {
+	if inputSortDir < 0 {
 		keysCompare = func(a, b *zng.Record) int { return rs(b, a) }
 	} else {
 		keysCompare = rs
 	}
+	builder, err := builder.NewColumnBuilder(zctx, keyNames)
+	if err != nil {
+		return nil, fmt.Errorf("compiling groupby: %w", err)
```
nwt, line 166: Nit: "compiling groupby" doesn't feel like the right text here.

nwt (approved):

PR #1725, opened 2020-12-01
allow len() function to work on IPs

mccanne:
This commit extends the len function to return the length of IP
addresses and networks.  This is handy for differentiating between
IPv4 (len=4) and IPv6 (len=16).  Likewise, networks are len 8 and
len 32.

### DIFF
--- a/expr/ztests/ip-len.yaml
+++ b/expr/ztests/ip-len.yaml
@@ -1,0 +1,11 @@
+zql: cut len(addr)
+
+input: |
+  #0:record[addr:ip]
+  0:[10.0.0.1;]
+  0:[fe80::215:17ff:fe84:c13f;]
+
+output: |
+  #0:record[len:int64]
+  0:[4;]
+  0:[16;]
--- a/expr/ztests/ip-len.yaml
+++ b/expr/ztests/ip-len.yaml
@@ -1,11 +1,11 @@
-zql: cut len(addr)
+zql: cut n1=len(addr),n2=len(n)
 
 input: |
-  #0:record[addr:ip]
-  0:[10.0.0.1;]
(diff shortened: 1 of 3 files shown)


expr/ztests/ip-len.yaml
```diff
@@ -0,0 +1,11 @@
+zql: cut len(addr)
+
+input: |
+  #0:record[addr:ip]
```
nwt, line 4: I hIghly recommend adding test coverage for net as well.

nwt (approved):

expr/ztests/ip-len.yaml
mccanne: Good idea.  The exposed an old bug with ipv6 network values.

philrz:
I've verified the `ip` variant of this in `zq` commit `315c0c7d`.

Using the [zq-sample-data](https://github.com/brimsec/zq-sample-data):

```
$ zq -version
Version: v0.24.0-40-g315c0c7d

$ zq -f table 'uid=C8Tful1TvM3Zf5x8fl | put MyLen=len(id.orig_h) | cut uid,id.orig_h,MyLen' conn.log.gz 
UID                ID.ORIG_H     MYLEN
C8Tful1TvM3Zf5x8fl 10.164.94.120 4

$ zq -f table 'uid=CFNJ252UabsYUxNnXe | put MyLen=len(id.orig_h) | cut uid,id.orig_h,MyLen' conn.log.gz 
UID                ID.ORIG_H                MYLEN
CFNJ252UabsYUxNnXe fe80::215:17ff:fe84:c13f 16
```

I don't have test data handy with subnets in it, but this gives you the idea. The linked PR shows a subnet example.

Thanks @mccanne!

PR #1727, opened 2020-12-02
simplify the groupby proc

mccanne:
This commit simplifies the groupby proc by using data structures from
package expr and avoiding the creation of new aggregation evaluators
for each row since the evaluators depend only on the field names.

This is just "code simplification" and doesn't change any logic.

This will get groupby ready for the fix regarding issue #1701.

### DIFF
--- a/proc/groupby/groupby.go
+++ b/proc/groupby/groupby.go
@@ -522,8 +522,8 @@ func (a *Aggregator) nextResultFromSpills() (*zng.Record, error) {
 	// Build the result record.
 	a.builder.Reset()
 	var types []zng.Type
-	for _, res := range a.keyEvals {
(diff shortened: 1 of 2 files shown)


proc/groupby/groupby.go
```diff
@@ -534,7 +522,7 @@ func (a *Aggregator) nextResultFromSpills() (*zng.Record, error) {
 	// Build the result record.
 	a.builder.Reset()
 	var types []zng.Type
-	for _, res := range a.keyResolvers {
+	for _, res := range a.keyEvals {
```
nwt, line 525: Maybe change loop variable to e or eval?

nwt (approved):
:shipit:

PR #1728, opened 2020-12-02
use flat keys, add comments, and improve names in groupby proc

mccanne:
This commit takes a step torward the fix for #1701 by rearranging
the group-by hash table keys to be stored as flattened values.
This will let the output stage build the output values based on
possibly changing output types.

We also added a comment on the spill-merge step and one relating
to memory pressure management of the hash table.

Finally, we adopted the names valRefs and keyRefs for the expr.Evaluators
that refer to the key/value fields themselves as compared to the expressions
that compute the values that go in those fields.

### DIFF
--- a/proc/groupby/groupby.go
+++ b/proc/groupby/groupby.go
@@ -524,8 +524,6 @@ func (a *Aggregator) nextResultFromSpills() (*zng.Record, error) {
 	var types []zng.Type
 	for _, e := range a.keyRefs {
 		keyVal, _ := e.Eval(firstRec)
-		//XXX don't need this copy
-		keyVal = keyVal.Copy()
 		types = append(types, keyVal.Type)
 		a.builder.Append(keyVal.Bytes, keyVal.IsContainer())
 	}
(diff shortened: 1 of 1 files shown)


proc/groupby/groupby.go
```diff
@@ -573,8 +572,28 @@ func (a *Aggregator) readTable(flush, partialsOut bool) (zbuf.Batch, error) {
...
+		it := row.keyvals.Iter()
+		i := 0
+		for !it.Done() {
+			flatVal, _, err := it.Next()
+			if err != nil {
+				return nil, err
+			}
+			a.builder.Append(flatVal, zng.IsContainerType(row.keyTypes[i]))
+			i++
+		}
```
nwt, line 590:
```suggestion
		for _, typ := range row.keyTypes {
			flatVal, _, err := it.Next()
			if err != nil {
				return nil, err
			}
			a.builder.Append(flatVal, zng.IsContainerType(typ))
		}
```

nwt (approved):

Issue #1656, opened 2020-11-19
support in-archive index definitions


alfred-landrum:
This is a more fleshed out description of what's described in the design doc at #1183 . This work is specifically for index files; mappings will be handled in separate issue(s).

When chunks are added, either at import or during compaction, we should be able to create index files for the chunks. To do this, we'll need to store the index definitions inside the archive. This will also be needed when we want to start automatically using indexes to optimize queries.

Here's a layout scheme that accomplishes this:
- Use a new toplevel directory in the archive (`indexdefs`) to hold index definition files.
- An index definition file name has a structure like: `idxdef-<uuid>.zng`.
- The index definition file contains:
  - the information needed to create an index (likely some or all of the current `archive.Rule` struct)
  - optionally, a human readable convenience name if used at creation time (eg, "myindex"). Since these are a convenience, there's no restriction that these are unique among the index definition files.

At import/compaction, the writing process will:
- obtain the list of index definitions
- for each new chunk, create an index file for each of the index definitions, writing them into the `.zar` directory of the new chunk.
- use an index file name with a structure like `idx-<uuid>.zng`, where the uuid used corresponds to the uuid of the index definition file.

With this layout, a `zar find` operation that wants to use indexes will:
- allow the user to specify an index either by the indexes uuid, or by the convenience name, resolving it to an index uuid
- use the index uuid to read the correct index files for each chunk

An index can be removed by deleting the index definition file and all of the corresponding index files.

The per-chunk index file paths include the chunk's uuid since they are located in the chunk's zar directory, and the index definition file paths are unique as every index definition has its own uuid. So, their paths can be used as cache keys in the future, to speed up operations as we require.

PR #1736, opened 2020-12-03
archive index package

mattnibs:
The index package provides the foundation for building indexdef functionality
into archives. index.Defs are rules for creating indexes within a chunk's zar
dir. The pr includes functionality for creating an indexing Rule and Defintion,
then transforming these into microindices by applying the defs to a stream of zng
records.

Future prs will incorporate this package into the wider archive eco-system as
well as remove redundant existing funtionality from the archive package.

Part of #1656

### DIFF
--- a/ppl/archive/index/definition.go
+++ b/ppl/archive/index/definition.go
@@ -1,0 +1,97 @@
+package index
+
+import (
+	"context"
+	"fmt"
+	"path"
+	"regexp"
+
+	"github.com/brimsec/zq/ast"
+	"github.com/brimsec/zq/pkg/iosrc"
+	"github.com/segmentio/ksuid"
+)
+
+type Definition struct {
+	Rule
+	ID   ksuid.KSUID
+	Proc ast.Proc
+}
+
+func ReadDefinition(ctx context.Context, u iosrc.URI) (*Definition, error) {
+	id, err := parseDefFile(path.Base(u.Path))
+	if err != nil {
+		return nil, err
+	}
+	b, err := iosrc.ReadFile(ctx, u)
+	if err != nil {
+		return nil, err
+	}
+	def := &Definition{ID: id}
+	def.Rule, err = UnmarshalRule(b)
+	if err != nil {
+		return nil, err
+	}
+	def.Proc, err = def.Rule.Proc()
+	return def, err
+}
+
+func NewDefinition(r Rule) (*Definition, error) {
+	proc, err := r.Proc()
+	if err != nil {
+		return nil, err
+	}
+	return &Definition{ID: ksuid.New(), Rule: r, Proc: proc}, nil
+}
+
+func MustNewDefinition(r Rule) *Definition {
+	d, err := NewDefinition(r)
+	if err != nil {
+		panic(err)
+	}
+	return d
+}
+
+var defFileRegex = regexp.MustCompile(`idxdef-([0-9A-Za-z]{27}).zng$`)
+
+func parseDefFile(name string) (ksuid.KSUID, error) {
+	match := defFileRegex.FindStringSubmatch(name)
+	if match == nil {
+		return ksuid.Nil, fmt.Errorf("invalid definition file: %s", name)
+	}
+	return ksuid.Parse(match[1])
+}
+
+func (d *Definition) Filename() string {
+	return fmt.Sprintf("idxdef-%s.zng", d.ID)
+}
+
+func (d *Definition) Write(ctx context.Context, dir iosrc.URI) error {
+	b, err := d.Rule.Marshal()
+	if err != nil {
+		return err
+	}
+	return iosrc.WriteFile(ctx, dir.AppendPath(d.Filename()), b)
+}
+
+type DefinitionMap map[ksuid.KSUID]*Definition
+
+type Definitions []*Definition
+
+func (l Definitions) MapByInputPath() map[string][]*Definition {
+	m := make(map[string][]*Definition)
+	for _, d := range l {
+		m[d.Input] = append(m[d.Input], d)
+	}
+	return m
+}
+
+// StandardInputs returns the Defs from the list that have an empty InputPath.
+func (l Definitions) StandardInputs() []*Definition {
+	var defs []*Definition
+	for _, d := range l {
+		if d.Input == "" {
+			defs = append(defs, d)
+		}
+	}
+	return defs
+}
--- a/ppl/archive/index/proc.go
+++ b/ppl/archive/index/proc.go
@@ -1,0 +1,167 @@
+package index
+
+import (
+	"fmt"
+
+	"github.com/brimsec/zq/ast"
+	"github.com/brimsec/zq/expr"
+	"github.com/brimsec/zq/field"
+	"github.com/brimsec/zq/proc"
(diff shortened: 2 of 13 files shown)


ppl/archive/index/writer.go
```diff
@@ -0,0 +1,152 @@
...
+	if len(leaves) != 1 {
+		return nil, fmt.Errorf("flowgraph can only have 1 leaf, has %d", len(leaves))
+	}
+	d.leaf = leaves[0]
+	d.Go(d.run)
+	return d, nil
+}
+
+func (h *indexer) Done() {
+	// XXX not sure what I'm supposed to do here.
```
mattnibs, line 117: Please advise.

ppl/archive/index/proc.go
```diff
@@ -0,0 +1,167 @@
+package index
```
mattnibs, line 1: This code was copied from archive/proc.go. I'll remove archive/proc.go in a subsequent pr.

ppl/archive/index/dirwriter.go
```diff
@@ -0,0 +1,80 @@
...
+func (ws DirWriter) Close() (merr error) {
+	for _, w := range ws {
+		if err := w.Close(); err != nil {
+			merr = multierr.Append(merr, err)
+		}
+	}
+	// Drop all indices if there was an error closing.
+	if merr != nil {
+		ws.Abort()
+	}
```
alfred-landrum, line 62: This is probably not what we want, since this will end up calling Abort on already closed indices. I don't think we should do anything extra if there's a close error on the index writer.

```suggestion
```

ppl/archive/index/dirwriter.go
```diff
@@ -0,0 +1,80 @@
...
+	"context"
+	"path"
+
+	"github.com/brimsec/zq/zbuf"
+	"github.com/brimsec/zq/zng"
+	"github.com/brimsec/zq/zqe"
+	"go.uber.org/multierr"
+)
+
+type DirWriter []*Writer
```
alfred-landrum, line 13:
```suggestion
type MultiWriter []*Writer
```

ppl/archive/index/dirwriter.go
```diff
@@ -0,0 +1,80 @@
+package index
```
alfred-landrum, line 1: I suggest changing this file to `multiwriter.go`.

ppl/archive/index/def.go
```diff
@@ -0,0 +1,97 @@
+package index
```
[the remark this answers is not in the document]
mattnibs: Sure

ppl/archive/index/writer.go
```diff
@@ -0,0 +1,152 @@
...
+	return err
+}
+
+func (w *Writer) Abort() {
+	w.Close()
+	w.indexer.microindex.Abort()
+}
+
+type indexer struct {
+	tomb.Tomb
```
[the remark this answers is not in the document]
mattnibs: tomb is a package that makes it easier to manage the life span of a goroutine. It is similar to `errgroup` except it includes a dying chan, which signals to the managed goroutine that the parent wants it to exit. This was important in this use case because the Writer needs a way to tell the `indexer` that it has reached the end of the stream.

alfred-landrum (commented):
I talked with @mattnibs and made some other suggestions about reducing the functionality in `Dir`, consolidating the multiple-index writing code into multiwriter, and a few other ideas.

zio/tzngio/reader.go
```diff
@@ -210,3 +212,9 @@ func (r *Reader) parseValue(line []byte) (*zng.Record, error) {
 
 	return zng.NewRecordCheck(typ, raw)
 }
+
+func ReadAll(str string) (zbuf.Array, error) {
```
nwt, line 216: This feels odd, partly because it isn't really in keeping with ioutil.ReadAll. If you want something like it, I suggest adding zbuf.ReadAll and tzngio.NewReaderString.

ppl/archive/index/writer.go
```diff
@@ -0,0 +1,152 @@
...
+	if d.keyType == nil {
+		d.keyType = key.Type
+	}
+	if key.Type.ID() != d.keyType.ID() {
+		return fmt.Errorf("key type changed from %s to %s", d.keyType, key.Type)
+	}
+	if err := d.microindex.Write(rec); err != nil {
+		return err
+	}
+	return nil
```
nwt, line 151:
```suggestion
	return d.microindex.Write(rec)
```

ppl/archive/index/writer.go
```diff
@@ -0,0 +1,152 @@
...
+		indexer: indexer,
+	}
+	w.WriteCloser = zbuf.BatchRecordWriter(w)
+	return w, nil
+}
+
+type Writer struct {
+	Def *Def
+	URI iosrc.URI
+	zbuf.WriteCloser
```
nwt, line 37: Nit: Embedded fields are special, so put them first.

ppl/archive/index/rule.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+	return Rule{
+		Kind: RuleZQL,
+		ZQL:  prog,
+		Name: name,
+		Keys: keys,
+	}, nil
+}
+
+func (r Rule) Equals(r2 Rule) bool {
```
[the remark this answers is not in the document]
mattnibs: This is used for checking if an equivalent rule already exists, thus framesize does not apply and keys only applies to zql. I'll add a check for keys on zql. Also I'll change the name.

zio/tzngio/reader.go
mattnibs: I agree it should be in zbuf. Also should have a context. I still thinks its cumbersome to have to write `zbuf.ReadAll(tzngio.NewReader(resolver.NewContext(), strings.NewReader(str)))`, for something we end up doing a lot in tests.

Perhaps we could add this a function in some kind of test utility?

mattnibs:
@alfred-landrum / @nwt thanks for the reviews. This is ready for another look.

ppl/archive/index/rule.go
```diff
@@ -0,0 +1,190 @@
...
+	default:
+		return nil, fmt.Errorf("unknown rule kind: %s", r.Kind)
+	}
+}
+
+var keyName = field.New("key")
+
+var keyAst = ast.Assignment{
+	LHS: ast.NewDotExpr(field.New("key")),
+	RHS: ast.NewDotExpr(field.New("key")),
```
nwt, line 122:
```suggestion
	LHS: ast.NewDotExpr(keyName),
	RHS: ast.NewDotExpr(keyName),
```

ppl/archive/index/definition.go
```diff
@@ -0,0 +1,97 @@
...
+
+var defFileRegex = regexp.MustCompile(`idxdef-([0-9A-Za-z]{27}).zng$`)
+
+type Definition struct {
+	Rule
+	ID   ksuid.KSUID
+	Proc ast.Proc
+}
+
+func OpenDefinition(ctx context.Context, u iosrc.URI) (*Definition, error) {
```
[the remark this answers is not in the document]
mattnibs: Definitely.

alfred-landrum (approved):

PR #1738, opened 2020-12-03
eliminate zng.ErrUnset

mccanne:
Now that the zson spec more clearly defines the null concept,
zng.ErrUnset is no longer needed and is otherwise causing problems
with  the new groupby model to handle dynamic types.

This commit removes zng.ErrUnset and has each zng.Decode*()
function return the zero value of the type it decodes to.
If a callee of zng.Decode*() wants to treat null differently
from the zero value, then it should check for zv.Bytes==nil
before calling zng.Decode*().  This means that ErrUnset is no
longer bubbled up as an error condition, e.g., to groupby,
and further, that we should decide what each functions should
do with null (the reducers already handle nulls).  For example,
len(null) returns int64(0) but replace("foo", null, null) is
an error.  I made a first pass over the expr/function in this
regard, but we should decide and document it all. See #1737.

### DIFF
--- a/zng/bytes.go
+++ b/zng/bytes.go
@@ -18,7 +18,7 @@ func EncodeBytes(b []byte) zcode.Bytes {
 
 func DecodeBytes(zv zcode.Bytes) ([]byte, error) {
 	if zv == nil {
-		return nil, ErrUnset
+		return nil, nil
 	}
 	return []byte(zv), nil
 }
--- a/zng/bytes.go
+++ b/zng/bytes.go
@@ -17,9 +17,6 @@ func EncodeBytes(b []byte) zcode.Bytes {
 }
 
 func DecodeBytes(zv zcode.Bytes) ([]byte, error) {
-	if zv == nil {
-		return nil, nil
-	}
 	return []byte(zv), nil
 }
 
--- a/zng/string.go
+++ b/zng/string.go
@@ -22,9 +22,6 @@ func EncodeString(s string) zcode.Bytes {
 }
 
 func DecodeString(zv zcode.Bytes) (string, error) {
-	if zv == nil {
-		return "", nil
-	}
 	return string(zv), nil
 }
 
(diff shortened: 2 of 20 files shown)


zng/bytes.go
```diff
@@ -18,7 +18,7 @@ func EncodeBytes(b []byte) zcode.Bytes {
 
 func DecodeBytes(zv zcode.Bytes) ([]byte, error) {
 	if zv == nil {
-		return nil, ErrUnset
+		return nil, nil
 	}
 	return []byte(zv), nil
```
nwt, line 23: Nit: No need for the if statement; `return []byte(zv), nil` alone is sufficient.

zng/string.go
```diff
@@ -23,7 +23,7 @@ func EncodeString(s string) zcode.Bytes {
 
 func DecodeString(zv zcode.Bytes) (string, error) {
 	if zv == nil {
-		return "", ErrUnset
+		return "", nil
 	}
 	return string(zv), nil
```
nwt, line 28: Nit: No need for the if statement; `return string(zv), nil` alone is sufficient.

zng/typetype.go
```diff
@@ -16,7 +16,7 @@ func EncodeType(s string) zcode.Bytes {
 
 func DecodeType(zv zcode.Bytes) (string, error) {
 	if zv == nil {
-		return "", ErrUnset
+		return "", nil
 	}
 	return string(zv), nil
```
nwt, line 21: Nit: No need for the if statement; `return string(zv), nil` alone is sufficient.

nwt (approved):
This is good.

zng/typetype.go
mccanne: Hmm, just realized the zero-value of a type type should probably be "null".  I'll change it to that.  This stuff feels like it needs more clarification in the zson spec.

PR #1743, opened 2020-12-03
zng.Marshal: support marshaling to time

mattnibs:


mattnibs:
I'm not sure this if this is best approach for marshaling/unmarshaling fields with nano.Ts. Yet the problem remains.

mattnibs:
What we should probably do in the long run is have `pkg/nano` implement the resolver.(Un)Marshaler interface (this is what the go std does with Time and json). The problem with this is that nano.Ts is in `pkg` which means it shouldn't be dependent on anything outside it. Additionally I think package `zng/resolver` has `nano` as a dependency. There's some untangling to do here.

alfred-landrum:
> What we should probably do in the long run is have `pkg/nano` implement the resolver.(Un)Marshaler interface (this is what the go std does with Time and json). The problem with this is that nano.Ts is in `pkg` which means it shouldn't be dependent on anything outside it. Additionally I think package `zng/resolver` has `nano` as a dependency. There's some untangling to do here.

I agree we need to untangle. I think this PR is currently the best way to get the marshal support in. I think we should remove nano.Ts as a dependency of the zng & zng/resolver packages.

alfred-landrum (approved):

PR #1747, opened 2020-12-03
Quote vars in Makefile that might include space chars when expanded

philrz:
I had reason to run `make generate` for the first time ever. It failed with an error that took me a while to figure out. Turns out the root cause was because my `$PATH` happens to include an entry with a space character in it. Much like in shell scripting, it seems like quotes are in order.

Having seen & addressed that, I got to looking closer and realized the references to `CURDIR` would fail similarly if someone happened to have zq checked out in a location on the filesystem whose path had a space char in it. See for yourself:

```
14:06:37-phil@PhilMac:~/has space/zq$ make test-system
/bin/sh: space/zq/bin: No such file or directory
make: *** [bin/minio] Error 127
```

I've addressed those cases here as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,11 +60,11 @@ bin/minio:
 	@echo 'module deps' > bin/go.mod
 	@echo 'require github.com/minio/minio latest' >> bin/go.mod
 	@echo 'replace github.com/minio/minio => github.com/brimsec/minio v0.0.0-20201019191454-3c6f24527f6d' >> bin/go.mod
-	@cd bin && GOBIN="$(CURDIR)"/bin go install github.com/minio/minio
+	@cd bin && GOBIN="$(CURDIR)/bin" go install github.com/minio/minio
 
 generate:
-	@GOBIN="$(CURDIR)"/bin go install github.com/golang/mock/mockgen
-	@PATH="$(CURDIR)"/bin:"$(PATH)" go generate ./...
+	@GOBIN="$(CURDIR)/bin" go install github.com/golang/mock/mockgen
+	@PATH="$(CURDIR)/bin:$(PATH)" go generate ./...
 
 test-generate: generate
 	git diff --exit-code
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -60,11 +60,11 @@ bin/minio:
...
 	@echo 'require github.com/minio/minio latest' >> bin/go.mod
 	@echo 'replace github.com/minio/minio => github.com/brimsec/minio v0.0.0-20201019191454-3c6f24527f6d' >> bin/go.mod
-	@cd bin && GOBIN=$(CURDIR)/bin go install github.com/minio/minio
+	@cd bin && GOBIN="$(CURDIR)"/bin go install github.com/minio/minio
 
 generate:
-	@GOBIN=$(CURDIR)/bin go install github.com/golang/mock/mockgen
-	@PATH=$(CURDIR)/bin:$(PATH) go generate ./...
+	@GOBIN="$(CURDIR)"/bin go install github.com/golang/mock/mockgen
+	@PATH="$(CURDIR)"/bin:"$(PATH)" go generate ./...
```
nwt, line 67: For readability, I suggest tweaking all of these additions to use a single pair of quotes around the entire right-hand side.
```suggestion
	@GOBIN="$(CURDIR)/bin" go install github.com/golang/mock/mockgen
	@PATH="$(CURDIR)/bin:$(PATH)" go generate ./...
```

nwt (approved):

Makefile
philrz: @nwt: Thanks for the approval. Indeed, it crossed my mind to go that route, and I was confident that whichever direction I went, the PR feedback would suggest me to go the other direction. I've pushed a change to use quotes around the entire RHS as you suggested. I'll leave it open for a bit just in case someone wants to flip it back the other way again. It's that kinda day. 😉

nwt (approved):
So nice I approved it twice.

PR #1748, opened 2020-12-03
Fix microindex trailer read bug

### LINKED ISSUE #1730: zar zq failure: couldn't read trailer: expected 397 bytes but read 101
I get this on the master branch today, running latest on an ec2 instance.

To reproduce this, you can zar import data from s3://brim-sampledata .
You can substitute your own bucket for the -R parameter.

First:

$ zar import -s 100MB -R s3://zqd-demo-1/mark/zeek-logs/dns s3://brim-sampledata/wrccdc/zeek-logs/dns.log.gz
$ export ZAR_ROOT=s3://zqd-demo-1/mark/zeek-logs/dns
$ zar zq -t "count()"
s3://zqd-demo-1/mark/zeek-logs/dns/zd/20180324/ts-1l78gr080Ay6KnVt5ec8ZBLbne1.zng: couldn't read trailer: expected 427 bytes but read 101

The expected and read number change each time you run zar zq.

I'm pretty sure this is a regression, since it is based on test scripts I was running successfully in October.

mattnibs:
The read trailer part of microindex was relying on the call to the
io.Reader.Read to fill the entire buffer. For the aws sdk this often do not
happen.

According to the golang io.Reader.Read documentation:

If some data is available but not len(p) bytes, Read conventionally returns
what is available instead of waiting for more.

Use io.ReadFull to ensure the entire trailer is read.

Closes #1730

### DIFF
--- a/microindex/trailer.go
+++ b/microindex/trailer.go
@@ -87,14 +87,8 @@ func readTrailer(r io.ReadSeeker, n int64) (*Trailer, int, error) {
 		return nil, 0, err
 	}
 	buf := make([]byte, n)
-	cc, err := io.ReadFull(r, buf)
-	if err != nil {
+	if _, err := io.ReadFull(r, buf); err != nil {
 		return nil, 0, err
-	}
-	if int64(cc) != n {
-		// this shouldn't happen but maybe could occur under a corner case
-		// or I/O problems XXX
-		return nil, 0, fmt.Errorf("couldn't read trailer: expected %d bytes but read %d", n, cc)
 	}
 	for off := int(n) - 3; off >= 0; off-- {
 		// look for end of stream followed by an array[int64] typedef then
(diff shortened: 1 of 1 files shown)


mccanne (approved):

microindex/trailer.go
```diff
@@ -87,7 +87,7 @@ func readTrailer(r io.ReadSeeker, n int64) (*Trailer, int, error) {
 		return nil, 0, err
 	}
 	buf := make([]byte, n)
-	cc, err := r.Read(buf)
+	cc, err := io.ReadFull(r, buf)
```
nwt, line 90: Nit: When io.ReadFull returns a nil error, cc is guaranteed to be len(buf), so you don't need to check cc below.

nwt (approved):

PR #1749, opened 2020-12-03
Reorder columns for Suricata alerts

philrz:
In a recent group discussion it was noted that the current left-to-right order of the field columns for Suricata alerts is in some ways suboptimal. This PR proposes some changes.

Everyone seemed to agree the `event_type` field should move directly to the right of the `ts` field. This would make it fairly apples-to-apples with how we present Zeek events, since those also has `ts` as the leftmost column, followed by `_path` which is roughly equivalent to what `event_type` is for Suricata. This should make it easier for us to make a visually appealing colored tile/icon for alerts that will stand out nicely relative to Zeek events.

We also noted that some of the fields within the `alert` record are likely to be seen as more useful relative to others, such as `alert.severity`, `alert.category`, etc., so I've moved some of those leftward as well. Even though the `alert.metadata` sub-fields are empty much of the time, the way things work in zq today, they need to be kept adjacent to the contents of the `alert` record, so I've just kept those rightward. That means other fields like `app_proto`, `dest_ip`, etc. are quite far out to the right after the often-empty columns of `alert.metadata`. Thankfully the contents of those fields are quite redundant with the contents of the Zeek events we'll be linking to for context.

The only one I'm on the fence about is `community_id`. It was originally all the way rightward, which is where it happens to land today on Zeek `conn` events as well, so I was prepared to leave it there for symmetry. However, the Zeek events have their `uid` fields that are very leftward (directly to the right of `_path`) and `community_id` is the closest Suricata has to an equivalent concept, so I can see the case for bringing it leftward as I've done here (I think @orochford was in favor of this). If a user had other tools were cutting & pasting Community ID values into, having it prominently visible in a default view would be helpful. OTOH once we have the correlation in Log Detail that unites all Zeek events and Suricata alerts via the combo of `uid` and `community_id`, users may never have to _see_ the Community ID value to obtain its benefits, so that would be a reason to let it hide off to the right and keep the main screen real estate for more `alert` contents.

With the changes in this branch, here's an example of what a a full screen of alerts might look like with these leftward columns.

image

Since that doesn't all fit, here's a top-down view. The left-to-right scrolling order would be the same as the top-down order shown here.

image


nwt (approved):

alfred-landrum (approved):

henridf (approved):

orochford:
As you have probably noticed over the past months - the way that Suricata data is displayed is very ad-hoc across the industry.

I think we have a grand chance here at improving that with little effort. I'd like to make the user experience as close to the Zeek logs as we can - I don't think this scales with every log type, but this is a core differentiator. My suggestion would be to have the start of the schema look v. similar to a zeek conn log

ts, event_type (colored), src_ip, src_port, dest_ip, dest_prt, proto, app_proto, alert.severity, alert_signature, alert_category, alert_action, alert* , ect

I'd like to be able to scan the data visually - for that the columns need to be comprable.

philrz:
I've pushed a commit with @orochford's suggested change. As part of that, the `community_id` is now back where it was originally as the rightmost column, just like where it lands today for Zeek `conn`. Having thought it through, this seems fine to me, as I'm hopeful the Community ID value will be something that "works for the user" without having its opaque contents staring them in the face all the time (i.e. via the new visualizations we have planned to join the alerts with the corresponding Zeek events, which will show up in Log Detail without them even having to know the Community ID value made it possible). If a user should need to get at it (such as to cut & paste it into another tool that has data keyed by Community ID) its current location also means that it'll be at the bottom of the top-down view in Log Detail, which seems convenient.

Here's a re-do of the screenshots with this new column order so reviewers can try it on for size.

image

image

Since three people had already approved the previous take at this, I'll let this bake a little longer in case anyone wants to circle back and argue that they like this less. If I don't hear from anyone I'll plan to merge it by the end of the day. Explicit re-approvals are welcomed if people agree it's as-good/better. 😄

PR #1750, opened 2020-12-03
Add default import stale duration

mattnibs:
Currently if StaleDuration for an archive import was not getting set (as is the
case for the zar command and zqd post to archive store), the StaleDuration
would stay at 0. A StaleDuration of 0 means that pretty much all the data in
the archive would get flushed every 1 second (the interval at which the flusher
checks for new data to flush).

Set a default staleDuration to a more appropriate 5 seconds for an archive
Writer.

### DIFF
--- a/ppl/archive/import.go
+++ b/ppl/archive/import.go
@@ -26,13 +26,12 @@ var (
 	// before they are flushed to disk.
 	ImportBufSize          = int64(sort.MemMaxBytes)
 	ImportStreamRecordsMax = zngio.DefaultStreamRecordsMax
+
+	// For unit testing.
+	importLZ4BlockSize = zngio.DefaultLZ4BlockSize
 )
 
-// For unit testing.
-var (
-	importLZ4BlockSize         = zngio.DefaultLZ4BlockSize
-	importDefaultStaleDuration = time.Second * 5
-)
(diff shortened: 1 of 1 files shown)


ppl/archive/import.go
```diff
@@ -29,7 +29,10 @@ var (
 )
 
 // For unit testing.
-var importLZ4BlockSize = zngio.DefaultLZ4BlockSize
+var (
+	importLZ4BlockSize         = zngio.DefaultLZ4BlockSize
+	importDefaultStaleDuration = time.Second * 5
```
nwt, line 34: Nit: Make importDefaultStaleDuration const, or sort us if you don't.

nwt (approved):

ppl/archive/import.go
```diff
@@ -29,7 +29,10 @@ var (
 )
 
 // For unit testing.
-var importLZ4BlockSize = zngio.DefaultLZ4BlockSize
+var (
```
alfred-landrum, line 32: Can you combine this var block with the one above?

alfred-landrum (approved):

ppl/archive/import.go
mattnibs: const it is

Issue #1756, opened 2020-12-04
break out expr/eval.go


mccanne:
expr/eval.go should be broken out into logically separate source files, e.g., math.go, logical.go, cast.go, etc

PR #1757, opened 2020-12-04
finish refactoring compiler

### LINKED ISSUE #1162: reducer/compile and proc/compiler should be consistent

mccanne:
This is the final, heavy lift on the compiler refactor issue #1162.
Except for the creation of the filter.Program interface and its
implementation as compiler.Filter, everything here is just code
movement.  All the compiler related logic from packages filter and
expr have been moved into package compiler.  Also, we missed
the rename and join procs in the previous commit so they are taken
care of here.

The filter.Program interface needed to be created to break an import
cycle since the scanner was previously calling filter.Compile and
moving that into compiler created the cycle.

The functions in expr/eval.go should br broken out in a subsequent PR 
See issue #1756.

Closes #1162.

### DIFF
--- a/compiler/expr.go
+++ b/compiler/expr.go
@@ -9,7 +9,6 @@ import (
 	"github.com/brimsec/zq/expr/function"
 	"github.com/brimsec/zq/field"
 	"github.com/brimsec/zq/zng/resolver"
-	"github.com/brimsec/zq/zql"
 )
 
 // CompileExpr compiles the given Expression into an object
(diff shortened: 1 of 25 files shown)


filter/filter_test.go
```diff
@@ -504,7 +505,7 @@ func TestBadFilter(t *testing.T) {
 	t.Parallel()
 	proc, err := zql.ParseProc(`s =~ \xa8*`)
 	require.NoError(t, err)
-	_, err = filter.Compile(resolver.NewContext(), proc.(*ast.FilterProc).Filter)
+	_, err = compiler.CompileFilter(resolver.NewContext(), proc.(*ast.FilterProc).Filter)
```
nwt, line 508: Maybe make this
```go
	_, err = compiler.NewFilter(resolver.NewContext(), proc.(*ast.FilterProc).Filter).AsFilter()
```
so we don't have to export compiler.CompileFilter?

filter/filter.go
```diff
@@ -2,19 +2,22 @@ package filter
...
 	"github.com/brimsec/zq/zng"
-	"github.com/brimsec/zq/zng/resolver"
 )
 
 type Filter func(*zng.Record) bool
 
+type Program interface {
+	AsFilter() (Filter, error)
+	AsBufferFilter() (*BufferFilter, error)
+}
```
nwt, line 19: I think this ought to live in zbuf/scanner.go since its use there motivates its existence.

compiler/filter.go
```diff
@@ -0,0 +1,178 @@
...
+
+// AsFilter implements filter.Program
+func (f *Filter) AsFilter() (filter.Filter, error) {
+	if f == nil {
+		return nil, nil
+	}
+	return CompileFilter(f.zctx, f.ast)
+}
+
+// AsBufferFilter implements filter.Program
```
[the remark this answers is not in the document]
mccanne: Cool, I wondered why we had those elsewhere.

compiler/expr.go
```diff
@@ -0,0 +1,298 @@
...
+		return nil, err
+	}
+
+	node, ok := parsed.(ast.Expression)
+	if !ok {
+		return nil, errors.New("expected Expression")
+	}
+
+	return CompileExpr(resolver.NewContext(), node)
+}
```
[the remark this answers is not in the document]
mccanne: Will delete.

filter/filter.go
mccanne: It felt lonely in package filter.  Since it's in zbuf, I called it zbuf.Filter.

nwt (approved):

Issue #1760, opened 2020-12-05
zson identifiers can't be type names


mccanne:
We need to finalize the type value syntax and make sure it's unambiguous with the zson value syntax when a type value is embedded in a  complex value.

PR #1765, opened 2020-12-06
improve zson types and add zson methods to zng.Type

### LINKED ISSUE #1675: implement ZSON canonical type values
NEW DESCRIPTION (12/5/20):

The ZSON type system makes type values easier to reason about in zq.  This issue revisits the type type implementation and proposes to change things so TypeType store its value as a primitive type but is implemented as a complex type so that each instance of a TypeType can return its canonical type string.  Also, resolver.Context should have a method to use a hash table to efficiently look up a TypeType pointer by the string representing its canonical ZSON type value. 

OLD DESCRIPTION:

Type values need another look as properly encoding embedded references to aliases is a bit tricky.  If we encoded a type value as its type ID (in zng) then we would need to translate zcode.Bytes of anything that referred to the type value.  This would be relatively expensive when crossing type context boundaries.  But if we encode the value as the zson type string, then any references to aliases in the type string need to have a corresponding alias in the zng stream prior to the use of that value.  So, this approach also requires a parse of every type value to look for the aliases.  The problems are duals of one another.

However, we observe that the number of distinct type values in a given zng stream is bounded by the number of distinct schemas which doesn't get all that big (except for the pathological case that every ingested record introduces a new type, which can only happen if an entity generates record keys with ever changing values.  it is easy to argue that we don't support this workload and detect it and raise an error.  If some app does do this (the guys who looked at suricata schemas indicated that this can happen), then the ingest policy do something to avert the problem like map the changing keys to a value with a command field name).

Given that the number of distinct type values is bounded, we can allocate them as a method on resolver.Context and keep a table that maps the zcode.Bytes for the type value to the list of aliases that are in that type.  That way, whenever we need to move aliases across type context's due to type values reference aliases, we can simply reference this table and pull across the needed aliases.

The various writers will need to be updated to handle this and at some point the app should understand type values but it's not hugely important as the text strings will be good enough ux for now.

mccanne:
This commit improves the type system for zson.  It updates the spec
and improves the implementation.  We also added ZSON() and ZSONOf()
methods to zng.Type.  Support for enums is now implemented.

The TypeType implementation has been changed to use a complex-type approach
(even the type values are serialized as primitive values in zng)
so that zng.Type pointers may be stored for each unique value in the context,
making the implementation simpler and more efficient.  We also
added a hash lookup to resolver.Context to map zng.Type pointers to
their string encoding so they don't have to be reformatted each time
typeof() is called on a given type.

Closes #1675, #1760


zng/docs/zson.md
```diff
@@ -257,41 +276,45 @@ The syntax for a decorator is a parenthesized type, as in
...
-type may then be used as or within a type value or decorator.  The name
+This creates a new type whose name is given by the identifier and whose
+type is equivalent to the type of `<value>`.  This new
+type may then be used anywhere a type may appear.  The name
 of the type definition
 must not be equal to any of the primitive type names.
 
-The result of a type definition is a value with the new type.
-Thus, type definitions can be chained as in
+A type definition may also appear inside of decorator as in
```
henridf, line 298:
```suggestion
A type definition may also appear inside a decorator as in
```

zng/docs/zson.md
```diff
@@ -257,41 +276,45 @@ The syntax for a decorator is a parenthesized type, as in
...
 ```
- <value> (= <identifier> ) (= <identifer> )
+<identifier> (= <type> )
```
henridf, line 300: Is it deliberate that the word "alias" doesn't appear here?

zng/docs/zson.md
```diff
@@ -506,52 +536,76 @@ map type, or two elements referring to the type of the keys and type of the valu
 A map value may be empty.  An empty map value without a type decorator is
 presumed to be an empty map of type (`null`, `null`).
 
-### 3.5 Type Value
+#### 3.4.7 Type Value
+
+The type of a type value is `type` while its value is depicted by parenthesizing
+its type description (as defined in [Section 3.5](#35-type-syntax).  For example,
+a value of a record with a single field called `t` of type `type` might look
```
henridf, line 543:
```suggestion
a type value of a record with a single string field called `t` would look
```

zio/zsonio/writer.go
```diff
@@ -55,76 +55,97 @@ func (w *Writer) Write(rec *zng.Record) error {
...
+	if err == nil && decorate {
+		err = w.decorate(typ, parentKnown)
 	}
 	return err
 }
 
-func (w *Writer) writeDecorator(typ zng.Type, known, null bool) error {
-	if known || (!null && impliedPrimitive(typ)) {
+func (w *Writer) nextInternalType() string {
+	name := fmt.Sprintf("%d", w.nid)
```
nwt, line 117:
```suggestion
	name := strconv.Itoa(w.nid)
```

zng/docs/zson.md
```diff
@@ -506,52 +536,76 @@ map type, or two elements referring to the type of the keys and type of the valu
...
+like this:
+```
+{ t: (string) (type) }
+```
+Since type values have implied types, the `type` type decorator can be omitted:
+```
+{ t: (string) }
+```
+Now supposing we created a second field called `t2` whose type is
+computed by introspecting the type of `t`.  This result is
```
henridf, line 553:
```suggestion
Now suppose we added a second field called `t2` whose type is
computed by introspecting the type of `t`.  The resulting record type value is
```

zng/bstring.go
```diff
@@ -73,3 +73,17 @@ func (t *TypeOfBstring) StringOf(data zcode.Bytes, fmt OutFmt, inContainer bool)
...
+func (t *TypeOfBstring) ZSON() string {
+	return "bstring"
+}
+
+// Values of type bstring may contain a mix of valid UTF-8 and arbitrary
+// binary data.  These are represented in output using the same formatting
+// with "\x.." escapes as Zeek.
+// In general, valid UTF-8 code points are passed through unmodified,
+// though for the ZEEK_ASCII output format, all non-ascii bytes are
+// escaped for compatibility with older versions of Zeek.
```
nwt, line 86:
```suggestion
```

zng/docs/zson.md
henridf: (I think if I were reading this without knowing about the latest zql, I might now be stuck wondering "what do they mean by "introspecting" here?)

zng/type.go
```diff
@@ -275,8 +277,6 @@ func LookupPrimitiveById(id int) Type {
 		return TypeTime
 	case IdDuration:
 		return TypeDuration
-	case IdType:
```
nwt, line 278: Delete the IdType const as well?

mccanne: Ruh, roh.  We need to keep IdType because type values are stored as primitives in zng but managed as complex types in resolver.Context at runtime...  it's a bit of a hybrid approach but I feel it is the right design.  I will get this working soon and push an update.

zng/docs/zson.md
mccanne: Oops, that's supposed to be
```
<identifier> = ( <type> )
```
And yes I'm intentionally avoiding "alias" in favor of "type name" in the ZSON spec.  I will leave "alias" as the low-level concept in zng that implements user-defined ZSON type names.  Does this seem reasonable?

zng/docs/zson.md
mccanne: Fair enough, but "type introspection" is a well-known concept from the world of programming languages and compilers.

zio/zsonio/writer.go
```diff
@@ -55,76 +55,97 @@ func (w *Writer) Write(rec *zng.Record) error {
...
+
+func (w *Writer) isKnown(typ zng.Type, known bool) bool {
+	if !known {
+		known = w.typedefs.exists(typ)
+	}
+	return known
+}
+
+func (w *Writer) writeValue(indent int, typ zng.Type, bytes zcode.Bytes, parentKnown, decorate bool) error {
+	known := w.isKnown(typ, parentKnown)
```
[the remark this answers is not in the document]
mccanne: Sounds good.  An early version of the branch had use isKnown throughout but now it's used just here.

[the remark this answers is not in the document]
mccanne: And I hadn't looked at the logic that way... much clearer!

zng/type.go
mccanne: Ok, I changed my mind about using type context for the type/zson hash table.  I created package zson (which is where I put the zson parser in a branch under dev) and put the hash table in zson.

zng/docs/zson.md
henridf: Totally! I would be intrigued and wondering "how do i introspect this stuff" 🤓

zng/docs/zson.md
henridf: Yes, it sounds reasonable to me

zson/zson.go
```diff
@@ -0,0 +1,26 @@
...
+
+func Implied(typ zng.Type) bool {
+	switch typ.(type) {
+	case *zng.TypeOfInt64, *zng.TypeOfTime, *zng.TypeOfFloat64, *zng.TypeOfBool, *zng.TypeOfBytes, *zng.TypeOfString, *zng.TypeOfIP, *zng.TypeOfNet, *zng.TypeOfType:
+		return true
+	}
+	return false
+}
+
+func SelfDescribing(typ zng.Type) bool {
```
henridf, line 15:
```suggestion
// SelfDescribing(t) returns true if the type can be unambiguously recovered from literal values of this type.
func SelfDescribing(typ zng.Type) bool {
```

henridf: I tried to add a doc string to both funcs here to see if I could, but now I'm thinking `Implied` would have the same documentation. Is the difference that Implied is for simple types and SelfDescribing is for all type?

zng/docs/zson.md
```diff
@@ -506,52 +536,76 @@ map type, or two elements referring to the type of the keys and type of the valu
 A map value may be empty.  An empty map value without a type decorator is
 presumed to be an empty map of type (`null`, `null`).
 
-### 3.5 Type Value
+#### 3.4.7 Type Value
 
-The syntax of a type value or of a decorator expressed as a type value mirrors
-the value syntax.
+The type of a type value is `type` while its value is depicted by parenthesizing
+its type description (as defined in [Section 3.5](#35-type-syntax).  For example,
```
henridf, line 542: There's a yaml test elsewhere in this PR that has: 
```
  {
      typeof: ({x:int32,s:string}),
      count: 2 (uint64)
  } (=0)
```

When I first read this, trying to find a pattern and knowing that the parenthesized `(uint64)` , my first guess was that `({x:int32,s:string})` was also a type...  but that couldn't be since there was no value.

So I'm wondering what is the motivation for parentheizing type values... is there some syntactical ambiguity?

Anyway, more of a design quetion than code review... so don't let it hold up the PR.

henridf (approved):
🏄 🌊

zng/docs/zson.md
mccanne: Yes, the zson parser is simpler with type values beginning with open paren as no value begins with that character.  Otherwise, you need deep backtracking to differentiate between 
```
{ foo:int64 }
```
as a type value and 
```
{ foo:10 }
```
as a value.  I also thought using the decorator parens for type value parens unified the look and feel but I see that might be a bit confusing. 

With the open paren here, I think the parser needs just one rune of lookahead throughout but I just started working on the parser and we will see.  If something unexpected comes up, maybe we can revisit this?

zson/zson.go
mccanne: Great point.  I added this as a comment and included similar verbiage in the spec:
```
+// SelfDescribing returns true for types whose type name can be entirely derived
+// from its typed value, e.g., a record type can be derived from a record value
+// because all of the field names and type names are present in the value, but
+// an enum type cannot be derived from an enum value because not all the enumerated
+// names are present in the value.  In the former case, a decorated typedef can
+// use the abbreviated form "(= <name>)", while the letter case, a type def must use
+// the longer form "<value> (<name> = (<type>))".
 func SelfDescribing(typ zng.Type) bool {
```

PR #1771, opened 2020-12-08
bearer token middleware

alfred-landrum:
This adds initial server side support to authenticate requests by looking for valid bearer tokens in JWT format. This adds a single test that enables authentication & uses a test public/private key pair to verify. The JWT's are expected to have claims that include the issuer, audience, and Brim specific tenant and user id.

The auth related config is made available via environment variables, and I've used my 'fromenv' library to handle that configuration.


ppl/cmd/zqd/listen/command.go
```diff
@@ -151,6 +152,9 @@ func (c *Command) Run(args []string) error {
 }
 
 func (c *Command) init() error {
+	if err := fromenv.Unmarshal(&c.conf); err != nil {
```
mattnibs, line 155: Wouldn't this overwrite conf variables from the command line. IIRC the standard command line arg > env > config file. No?

mattnibs: I suppose this ok for now, since the auth parameters can be configured via cli arg. Something to think about going forward though.

ppl/zqd/auth.go
```diff
@@ -0,0 +1,165 @@
...
+	// authTokenContextValue is the string the jwtmiddleware library uses to
+	// store the validated & parsed JWT in a request's context.
+	authTokenContextValue = "zqd-core-auth-token"
+
+	tenantIDClaim = "brim_tenant_id"
+	userIDClaim   = "brim_user_id"
+)
+
+type AuthConfig struct {
+	Enabled  bool   `env:"ZQD_AUTH_ENABLED"`
```
mattnibs, line 28: I found it kind of confusing that the AuthConfig reference in core.Config didn't have `env:"ZQD"`. Should the fromenv be able to preprend children of a config var with `PARENTTAG_`?

zqe/error.go
```diff
@@ -35,6 +36,8 @@ func (k Kind) String() string {
 		return "item already exists"
 	case NotFound:
 		return "item does not exist"
+	case NoCredentials:
+		return "misssing authentication credentials"
```
mattnibs, line 40:
```suggestion
		return "missing authentication credentials"
```

ppl/zqd/handlers.go
```diff
@@ -606,3 +608,12 @@ func extractSpaceID(c *Core, w http.ResponseWriter, r *http.Request) (api.SpaceI
 	}
 	return api.SpaceID(id), true
 }
+
+func handleIdentityGet(c *Core, w http.ResponseWriter, r *http.Request) {
+	ident, ok := IdentifyFromContext(r.Context())
+	if !ok {
+		respondError(c, w, r, zqe.ErrNoCredentials())
```
mattnibs, line 615: Wouldn't it be impossible to get this error given that `handleIdentityGet` is wrapped in core.authhandle? Or perhaps I am missing something.

Also wouldn't `zqe.InValid("no credentials")` be a better response code? If they got here it means they are authorized, its just an invalid request.

alfred-landrum:
@nwt & I had a long discussion about using environment variables to configure parts of zqd, and came to the conclusion that it will be better to stick with cli flags whenever we can. (I had given him a heads up about using fromenv in here to see how it would work.) I'll update the PR to pass in auth configuration via cli flags.

zqe/error.go
```diff
@@ -21,6 +21,7 @@ const (
 	Exists
 	Invalid
 	NotFound
+	NoCredentials
```
mattnibs, line 24: Doesn't look like this is being used anywhere, consider dropping it?

mattnibs (approved):

ppl/zqd/auth.go
```diff
@@ -0,0 +1,173 @@
...
+	Enabled  bool
+	Audience string
+	Issuer   string
+	JWKSPath string
+}
+
+func (c *AuthConfig) SetFlags(fs *flag.FlagSet) {
+	fs.BoolVar(&c.Enabled, "auth.enabled", false, "enable authentication checks")
+	fs.StringVar(&c.Audience, "auth.audience", "", "expected JWT audience claim")
+	fs.StringVar(&c.Issuer, "auth.issuer", "", "expected JWT issuer claim")
```
nwt, line 38:
```suggestion
	fs.StringVar(&c.Audience, "auth.audience", "", "requried JWT audience claim")
	fs.StringVar(&c.Issuer, "auth.issuer", "", "required JWT issuer claim")
```

ppl/zqd/core.go
```diff
@@ -115,33 +127,42 @@ func NewCore(ctx context.Context, conf Config) (*Core, error) {
...
+	c.authhandle("/space/{space}/subspace", handleSubspacePost).Methods("POST")
 }
 
 func (c *Core) addWorkerRoutes() {
 	c.handle("/worker", handleWorker).Methods("POST")
 }
 
-func (c *Core) handle(path string, f func(*Core, http.ResponseWriter, *http.Request)) *mux.Route {
-	return c.router.HandleFunc(path, func(w http.ResponseWriter, r *http.Request) {
+func coreHandler(c *Core, f func(*Core, http.ResponseWriter, *http.Request)) http.Handler {
```
nwt, line 152: Maybe a method instead of a function?

ppl/cmd/zqd/listen/command.go
```diff
@@ -88,6 +88,8 @@ func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
 	f.StringVar(&c.suricataUpdaterPath, "suricataupdater", "", "command to update Suricata rules (run once at startup)")
 	f.StringVar(&c.workers, "workers", "", "workers as comma-separated [addr]:port list")
 	f.StringVar(&c.zeekRunnerPath, "zeekrunner", "", "command to generate Zeek logs from pcap data")
+
+	c.conf.Auth.SetFlags(f)
```
nwt, line 92: Nit: Move this up before `c.conf.Version = cli.Version` to consolidate c.conf twiddling.

ppl/zqd/auth_test.go
```diff
@@ -0,0 +1,82 @@
...
+func TestAuthentication(t *testing.T) {
+	authConfig := zqd.AuthConfig{
+		Enabled:  true,
+		Audience: "https://test.brimsecurity.com/",
+		Issuer:   "https://testauth.brimsecurity.com",
+		JWKSPath: "testdata/auth-public-jwks.json",
+	}
+	core, conn := newCoreWithConfig(t, zqd.Config{
+		Logger: zap.NewNop(),
+		Auth:   authConfig,
```
nwt, line 41:
```suggestion
		Auth:   authConfig,
		Logger: zap.NewNop(),
```

zqe/error.go
nwt: Lies. It's used in ppl/zqd/handlers.go.

ppl/zqd/core.go
```diff
@@ -36,11 +36,17 @@ type Config struct {
...
 	Suricata pcapanalyzer.Launcher
 	Zeek     pcapanalyzer.Launcher
 }
 
+type middleware interface {
+	Middleware(next http.Handler) http.Handler
+}
+
 type Core struct {
+	auth      middleware
```
nwt, line 49: I think this could be a little more straightforward if (a) the type of this field were mux.MiddlewareFunc and (b) the field was nil when AuthConfig.Enabled was false. Then we wouldn't need the middleware interface or noopAuthenticator.

ppl/zqd/core.go
```diff
@@ -36,11 +36,17 @@ type Config struct {
 	Root        string
 	Version     string
 
+	Auth     AuthConfig
```
nwt, line 39: Put this before Logger.

nwt (approved):

marktwallace (approved):
Just got a chance to look at this now. The middleware looks good, and I think it will only require small adjustments to to the k8s yaml to deploy it. It's nice the Auth0 uses standard JWT tokens.
I will need to make some minor changes to the routes in #1695 after you merge this. I think it should go in before #1695 if possible. The deployment.yaml changes for this should go in a later PR, maybe next week.

ppl/zqd/auth.go
```diff
@@ -0,0 +1,145 @@
...
+	Issuer   string
+	JWKSPath string
+}
+
+func (c *AuthConfig) SetFlags(fs *flag.FlagSet) {
+	fs.BoolVar(&c.Enabled, "auth.enabled", false, "enable authentication checks")
+	fs.StringVar(&c.Audience, "auth.audience", "", "required JWT audience claim")
+	fs.StringVar(&c.Issuer, "auth.issuer", "", "required JWT issuer claim")
+	fs.StringVar(&c.JWKSPath, "auth.jwkspath", "", "path to JSON Web Key Set file")
+}
```
marktwallace, line 41: OK, now that I see this, I can make changes in #1695 so the "worker" flags follow the same pattern.

PR #1773, opened 2020-12-08
Fix cutter type bug and refactor into cut, drop, and pick

### LINKED ISSUE #1739: fix output type bug for cut proc
This is the same bug as in issue #1701 but for cut instead of groupby.

mccanne:
This commit fixes the output type bug for cut that affected groupby
and cut	where the logic incorrectly assumed that the input record type
reliably determined the	output record type.  Since this doesn't
always hold, zng could be generated that corresponding to the wrong
type resulting in corrupt zng output.

While we refactored this, we realized the semantics of cut and "cut -c"
are really different and after discussion decided to adopt "drop"
as a replacement for "cut -c".	We also	debated	whether	cut semantics
should be like sort and	groupby	key expressions, which skip records
for which all input fields aren't present instead of cuts current
semantics, which allows "partial" cuts to occur when not all fields
are present.  We decided to keep the cut semantics and introduce
a new operator called "pick" that has the strict semantics.

Another	improvement here is that we created a generic "apply" proc
that allows any function to be implemented as a proc and the implementations
of cut and pick now use this pattern.

Fixes #1739


expr/cutter.go
```diff
@@ -1,199 +1,146 @@
...
+	if c.FoundCut() {
+		return ""
 	}
-	if cb == nil {
-		return nil, nil
+	list := fieldList(c.fieldExprs)
+	cmd := "pick"
+	if c.droppers != nil {
+		cmd = "cut"
 	}
```
nwt, line 144: This part feels dicey.

How about a wrapper in compiler/proc.go like
```go
type prefixedWarningFunction {
	proc.Function
	prefix string
}

func (p *prefixedWarningFunction) Warning() string {
	if s := p.Function.Warning(); s != "" {
		return fmt.Sprintf("%s: %s", prefix, s)
	}
}
```
so that compileProc can say
```go
		return proc.NewApplier(pctx, parent, prefixedWarningFunction{cutter, "cut"}), nil
```
and
```go
		return proc.NewApplier(pctx, parent, prefixedWarningFunction{cutter, "pick"}), nil
```
?

zql/docs/processors/README.md
```diff
@@ -113,6 +95,33 @@ TIME                        UID
...
+---
+
+## `drop`
+
+|                           |                                                             |
+| ------------------------- | ----------------------------------------------------------- |
+| **Description**           | Return the data only from all but the specified named fields.       |
+| **Syntax**                | `drop <field-list>`                   |
+| **Required<br>arguments** | `<field-list>`<br>One or more comma-separated field names or assignments.  |
+| **Developer Docs**        | https://pkg.go.dev/github.com/brimsec/zq/proc/cut            |
```
philrz, line 107:
```suggestion
| **Developer Docs**        | https://pkg.go.dev/github.com/brimsec/zq/proc            |
```

These links are a mild inconvenience: It'd return a 404 until we cut a GA release and tell pkg.go.dev to fetch the docs for that latest tag. For now we can link to a spot one level up and I've made a note-to-self to push an update for this link right after I post the next release.

philrz (commented):
Everything looks good to me from a functional perspective. `drop` needs an entry in the table of contents of `processors/README.md`, but the PR review interface won't let me post a comment on that because it's too far from the code actively being updated. In any case, once this is merged, I'll gladly circle back and put up a PR for documenting `pick` at which time I'll also add some mention in the `cut` docs that the user might want to look at `pick` if they're seeking a stricter alternative.

expr/cut.go
```diff
@@ -1,19 +1,25 @@
...
+	c, err := NewCutter(zctx, fieldRefs, fieldExprs)
+	if err != nil {
+		return nil, err
+	}
+	return &CutFunc{c}, nil
 }
 
-func (c *cutFunc) Eval(rec *zng.Record) (zng.Value, error) {
-	out, err := c.Cut(rec)
+func (c *CutFunc) Eval(rec *zng.Record) (zng.Value, error) {
```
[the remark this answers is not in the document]
mccanne: sweet!

expr/dropper.go
```diff
@@ -0,0 +1,125 @@
...
+	fields    []field.Static
+	resolvers []Evaluator
+	strict    bool
+}
+
+func NewDropper(zctx *resolver.Context, fields []field.Static) *Dropper {
+	return &Dropper{
+		zctx:     zctx,
+		fields:   fields,
+		droppers: make(map[int]*dropper),
```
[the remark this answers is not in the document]
mccanne: I'll put droppers at the end of the struct instead.

expr/cutter.go
mccanne: This is a less bad hack.  I added a warning prefix to proc.applier and created issue #1776.

expr/ztests/pick.yaml
```diff
@@ -0,0 +1,20 @@
+zql: pick s,x
```
nwt, line 1: This is the only mention of pick within package expr, so expr/ztests/ feels like an odd place for it, but I don't have any good ideas about where else it should go.

nwt (approved):
My outstanding comments are all at the nit level, so :shipit:.

compiler/parallelize.go
```diff
@@ -445,20 +467,28 @@ func Parallelize(p ast.Proc, N int, inputSortField field.Static, inputSortRevers
...
-					if eq(f.RHS, inputSortField) {
-						return buildSplitFlowgraph(seq.Procs[0:i], seq.Procs[i:], inputSortField, inputSortReversed, N)
-					}
+			for _, e := range p.Fields {
+				if eq(e, inputSortField) {
+					return buildSplitFlowgraph(seq.Procs[0:i], seq.Procs[i:], inputSortField, inputSortReversed, N)
 				}
+			}
+			continue
+		case *ast.CutProc, *ast.PickProc:
```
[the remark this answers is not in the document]
mccanne: Well it does both cuts and drops so I moved the drops into a new test called `TestDropNested`.

expr/cut_test.go
```diff
@@ -152,8 +152,8 @@ const nestedOut2Complement = `
 // Test cutting fields inside nested records.
 func TestCutNested(t *testing.T) {
```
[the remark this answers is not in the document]
mccanne: Well it does both cuts and drops so I moved the drops into a new test called TestDropNested.

expr/cutter.go
```diff
@@ -1,199 +1,152 @@
...
-		fields:      lhs,
-		resolvers:   rhs,
-		cutBuilders: make(map[int]*cutBuilder),
-	}
+		builder:     b,
+		fieldRefs:   fieldRefs,
+		fieldExprs:  fieldExprs,
+		typeCache:   make([]zng.Type, len(fieldRefs)),
+		outTypes:    typevector.NewTable(),
+		recordTypes: make(map[int]*zng.TypeRecord),
```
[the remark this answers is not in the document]
mccanne: moved order in struct

expr/ztests/pick.yaml
mccanne: It's entirely implemented in expr/cutter.go

nwt: You and I know that, but but someone new to the code won't. 😉 I'll add a comment later.

Issue #1778, opened 2020-12-08
remove storage details from space list api


alfred-landrum:
The current call to `GET /space`, which returns a list of `api.SpaceInfo`, includes the summary information for the space's storage, including the total amount of data present. For archive spaces, that requires listing & querying the space for each chunk.

We should alter this call so that it returns the list of spaces with only their quickly accessed properties, like the kind of info would come via a single database call in the future. The more detailed information would still be available per-space with the `GET /space/{space}` api.

philrz:
Verified in `zqd` commit `a18854f`.

As of `zqd` commit `8a084e50` that came right before this change, there was more detail returned in this API call. After having imported logs into two separate Spaces:

```
$ curl http://localhost:9867/space | jq .
[
  {
    "id": "sp_1lR1gYidmQXtTWdbuZuXAvlyo9r",
    "name": "foo",
    "data_path": "file:///Users/phil/foo/sp_1lR1gYidmQXtTWdbuZuXAvlyo9r",
    "storage_kind": "filestore",
    "span": {
      "ts": {
        "sec": 1521911720,
        "ns": 865716000
      },
      "dur": {
        "sec": 1269,
        "ns": 285521001
      }
    },
    "size": 1889955,
    "pcap_support": false,
    "pcap_size": 0,
    "pcap_path": ""
  },
  {
    "id": "sp_1lR1gtR0UmbbCHw9feO7QGl4Hkg",
    "name": "bar",
    "data_path": "file:///Users/phil/foo/sp_1lR1gtR0UmbbCHw9feO7QGl4Hkg",
    "storage_kind": "filestore",
    "span": {
      "ts": {
        "sec": 1521911720,
        "ns": 600725000
      },
      "dur": {
        "sec": 1269,
        "ns": 557814001
      }
    },
    "size": 42161039,
    "pcap_support": false,
    "pcap_size": 0,
    "pcap_path": ""
  }
]
```

Now at `zqd` commit `a18854f0`, it's more terse.

```
$ curl http://localhost:9867/space | jq .
[
  {
    "id": "sp_1lR1gYidmQXtTWdbuZuXAvlyo9r",
    "name": "foo",
    "data_path": "file:///Users/phil/foo/sp_1lR1gYidmQXtTWdbuZuXAvlyo9r",
    "storage_kind": "filestore"
  },
  {
    "id": "sp_1lR1gtR0UmbbCHw9feO7QGl4Hkg",
    "name": "bar",
[...]

PR #1779, opened 2020-12-09
improve space list performance

alfred-landrum:
This change removes the storage details, like total size of the space's data, from the space list api response. Fetching those details was significantly slowing down `zapi` operations, since zapi uses the space list api to translate space names to space id's.

Brim tests pass against this change, and I didn't notice any difference using Brim manually with it. I'm checking with the frontend folks to be sure, but it seems like Brim uses the 'get space' operation for a space's storage details, and doesn't rely on the space list response for them.

Closes https://github.com/brimsec/zq/issues/1778

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -105,27 +105,21 @@ func (s *SpaceID) Set(str string) error {
 	return nil
 }
 
-type SpaceInfo struct {
+type Space struct {
 	ID          SpaceID     `json:"id"`
+	ParentID    SpaceID     `json:"parent_id,omitempty"`
 	Name        string      `json:"name"`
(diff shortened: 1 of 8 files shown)


mattnibs (approved):

marktwallace (approved):
It is interesting that you did not need to change the implementation of the function SpaceInfo() at all. Also it looks like the type SpaceInfos was only used in one place. Nice refactor to remove some complexity!

alfred-landrum:
@jameskerr checked the Brim source & confirmed that it uses the 'space get' operation when it needs storage details, so that explains why my passing Brim tests with this change.

PR #1788, opened 2020-12-10
Add pick docs and cross-ref with cut

philrz:
This PR adds ZQL docs for the recently added `pick` processor. I've made sure to include an example that illustrates their fundamental difference and included links between the two so users who start at one but are seeking the stricter/looser behavior of the other can easily find it. I also swapped the order of two of the `cut` examples so the the numeric order matches to its comparable `pick` example.

I'm aware this introduces some redundant examples between `cut` and `pick`, but, so be it. I happen to prefer this approach to making a new user look in two places to get the full story. With what I'm doing here, I'm hoping if a user likes what they see in the first one they find, they're off to the races.

### DIFF
--- a/zql/docs/processors/README.md
+++ b/zql/docs/processors/README.md
@@ -79,7 +79,7 @@ zq -f table 'cut nothere,alsoabsent' weird.log.gz
 
 #### Output:
 ```zq-output
-cut: nothing found for nothere,alsoabsent
+cut: no record found with columns nothere,alsoabsent
 ```
 
 #### Example #4:
@@ -331,7 +331,7 @@ zq -f table 'pick nothere,alsoabsent' weird.log.gz
 
 #### Output:
 ```zq-output
-pick: nothing found for nothere,alsoabsent
+pick: no record found with columns nothere,alsoabsent
 ```
 
 #### Example #4:
--- a/zql/docs/processors/README.md
+++ b/zql/docs/processors/README.md
@@ -5,9 +5,11 @@ A pipeline may contain one or more _processors_ to filter or transform event dat
 The following available processors are documented in detail below:
 
 * [`cut`](#cut)
+* [`drop`](#drop)
 * [`filter`](#filter)
 * [`fuse`](#fuse)
 * [`head`](#head)
+* [`pick`](#pick)
 * [`put`](#put)
 * [`rename`](#rename)
 * [`sort`](#sort)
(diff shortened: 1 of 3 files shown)


zql/docs/processors/README.md
```diff
@@ -272,6 +277,66 @@ conn  2018-03-24T17:15:20.607695Z CpjMvj2Cvj048u6bF1 10.164.94.120 39169     10.
...
+TS                          UID
+2018-03-24T17:15:21.255387Z C8Tful1TvM3Zf5x8fl
+2018-03-24T17:15:21.411148Z CXWfTK3LRdiuQxBbM6
+2018-03-24T17:15:21.926018Z CM59GGQhNEoKONb5i
+...
+```
+
+#### Example #2:
+
+All of the named fields must be present in a record for `pick` to return a result for it. If we extend the field list from our previous example to include a field that's not present anywhere in our data, now a warning is shown and no data is returned for `ts` and `uid` fields.
```
henridf, line 308: As I read this, I wondered "is the warning emitted if _any_ record is lacking the pick fields, or only if _all_ are lacking? Easy enough to figure out on the command line, but I think it would be worth clarifying here.

zql/docs/processors/README.md
```diff
@@ -272,6 +277,66 @@ conn  2018-03-24T17:15:20.607695Z CpjMvj2Cvj048u6bF1 10.164.94.120 39169     10.
...
+
+```zq-command
+zq -f table 'pick ts,uid,nothere' conn.log.gz
+```
+
+#### Output:
+```zq-output
+pick: nothing found for ts,uid,nothere
```
henridf, line 316: I suggest changing the warning to "no record found with columns ts, uid, nothere". What do you think @philrz @mccanne ?

zql/docs/processors/README.md
philrz: Thanks for the feedback @henridf. That's a good point. I just pushed a commit that reworks the examples a bit in both `cut` and `pick` to try to address this and they feel a bit more "real" and apples-to-apples now. Please do give it another look and let me know what you think.

zql/docs/processors/README.md
philrz: @henridf: I'm fine with that suggestion. Really, I'm just happy we have a warning. 😄 

I'm waiting to hear from @mccanne on if he has a strong attachment to the current wording. If he's ok with changing it, it looks innocent enough of a change that I'd be fine pushing it as part of this branch. Stay tuned.

philrz: @henridf: @mccanne was fine with your proposed re-wording, so I've pushed a commit with the changes and the tests are all passing. I'd accept an Approval if you're willing. 😄

henridf (approved):

PR #1793, opened 2020-12-11
Makefile: force GOARCH to amd64

mattnibs:
Currently we only support builds for zeek and suricata for amd64. Until we have
builds for other architectures (e.g. arm64) just download the amd64 binary.

This will work for Mac silicon users who have Rosetta 2 enabled (current master
does not).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@ export GO111MODULE=on
 
 # If VERSION or LDFLAGS change, please also change
 # npm/build.
+GOARCH = "amd64"
 VERSION = $(shell git describe --tags --dirty --always)
 ECR_VERSION = $(VERSION)-$(ZQD_K8S_USER)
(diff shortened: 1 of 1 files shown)


marktwallace (approved):
Nice! I'm thinking of getting a Mac mini with the M1, too. :-)

Makefile
```diff
@@ -2,6 +2,7 @@ export GO111MODULE=on
 
 # If VERSION or LDFLAGS change, please also change
 # npm/build.
+GOARCH = "amd64"
```
alfred-landrum, line 5: I'd change the name of this (`ARCH`) since it doesn't impact any of the go build steps. This makefile variable is not be exported as an environment variable anywhere, and the one place we set a GOARCH env var is already fixed to amd64 (in create-release-assets).

alfred-landrum (approved):
Would you please change the create-release-assets target to use the new `ARCH` makefile variable before merging?

PR #1794, opened 2020-12-11
Archive: create indices on write/compact

### LINKED ISSUE #1656: support in-archive index definitions

mattnibs:
When an archive has index definitions  and either new data is imported 
or chunks are compacted, apply index rules to newly created chunks.

Also:

- Break zar index into base command with three subcommands:
    - add: same as what zar index was before.
    - ls: prints information and stats on index rules.
    - drop: removes index definitions.

Closes #1656

### DIFF
--- a/cli/outputflags/flags.go
+++ b/cli/outputflags/flags.go
@@ -26,15 +26,7 @@ func (f *Flags) Options() zio.WriterOpts {
 	return f.WriterOpts
 }
 
-func (f *Flags) formatFlags(fs *flag.FlagSet) {
-	fs.StringVar(&f.Format, "f", "zng", "format for output data [zng,zst,ndjson,table,text,csv,zeek,zjson,zson,tzng]")
-	fs.BoolVar(&f.textShortcut, "t", false, "use format tzng independent of -f option")
-	fs.BoolVar(&f.forceBinary, "B", false, "allow binary zng be sent to a terminal output")
-}
-
 func (f *Flags) setFlags(fs *flag.FlagSet) {
-	f.formatFlags(fs)
-
 	// zio stuff
 	fs.BoolVar(&f.UTF8, "U", false, "display zeek strings as UTF-8")
 	fs.BoolVar(&f.Text.ShowTypes, "T", false, "display field types in text output")
@@ -57,6 +49,7 @@ func (f *Flags) setFlags(fs *flag.FlagSet) {
 }
 
 func (f *Flags) SetFlags(fs *flag.FlagSet) {
+	f.SetFormatFlags(fs)
 	f.setFlags(fs)
 }
 
@@ -66,7 +59,9 @@ func (f *Flags) SetFlagsWithFormat(fs *flag.FlagSet, format string) {
 }
 
 func (f *Flags) SetFormatFlags(fs *flag.FlagSet) {
-	f.formatFlags(fs)
+	fs.StringVar(&f.Format, "f", "zng", "format for output data [zng,zst,ndjson,table,text,csv,zeek,zjson,zson,tzng]")
+	fs.BoolVar(&f.textShortcut, "t", false, "use format tzng independent of -f option")
+	fs.BoolVar(&f.forceBinary, "B", false, "allow binary zng be sent to a terminal output")
 }
 
 func (f *Flags) Init() error {
--- a/cli/outputflags/flags.go
+++ b/cli/outputflags/flags.go
@@ -26,7 +26,15 @@ func (f *Flags) Options() zio.WriterOpts {
 	return f.WriterOpts
 }
 
+func (f *Flags) formatFlags(fs *flag.FlagSet) {
+	fs.StringVar(&f.Format, "f", "zng", "format for output data [zng,zst,ndjson,table,text,csv,zeek,zjson,zson,tzng]")
+	fs.BoolVar(&f.textShortcut, "t", false, "use format tzng independent of -f option")
+	fs.BoolVar(&f.forceBinary, "B", false, "allow binary zng be sent to a terminal output")
+}
+
 func (f *Flags) setFlags(fs *flag.FlagSet) {
+	f.formatFlags(fs)
+
 	// zio stuff
 	fs.BoolVar(&f.UTF8, "U", false, "display zeek strings as UTF-8")
 	fs.BoolVar(&f.Text.ShowTypes, "T", false, "display field types in text output")
@@ -45,19 +53,20 @@ func (f *Flags) setFlags(fs *flag.FlagSet) {
 	// emitter stuff
 	fs.StringVar(&f.dir, "d", "", "directory for output data files")
 	fs.StringVar(&f.outputFile, "o", "", "write data to output file")
-	fs.BoolVar(&f.forceBinary, "B", false, "allow binary zng be sent to a terminal output")
 
 }
 
 func (f *Flags) SetFlags(fs *flag.FlagSet) {
 	f.setFlags(fs)
-	fs.StringVar(&f.Format, "f", "zng", "format for output data [zng,zst,ndjson,table,text,csv,zeek,zjson,zson,tzng]")
-	fs.BoolVar(&f.textShortcut, "t", false, "use format tzng independent of -f option")
 }
 
 func (f *Flags) SetFlagsWithFormat(fs *flag.FlagSet, format string) {
 	f.setFlags(fs)
 	f.Format = format
+}
+
+func (f *Flags) SetFormatFlags(fs *flag.FlagSet) {
+	f.formatFlags(fs)
 }
 
 func (f *Flags) Init() error {
--- a/ppl/archive/chunk/writer.go
+++ b/ppl/archive/chunk/writer.go
@@ -7,6 +7,7 @@ import (
 	"github.com/brimsec/zq/pkg/bufwriter"
 	"github.com/brimsec/zq/pkg/iosrc"
 	"github.com/brimsec/zq/pkg/nano"
+	"github.com/brimsec/zq/ppl/archive/index"
 	"github.com/brimsec/zq/ppl/archive/seekindex"
 	"github.com/brimsec/zq/zbuf"
 	"github.com/brimsec/zq/zio/zngio"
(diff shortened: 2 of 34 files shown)


ppl/cmd/zar/index/ls.go
```diff
@@ -0,0 +1,145 @@
...
+			input = def.Input
+		}
+		if def.ZQL != "" {
+			zql = def.ZQL
+		}
+
+		if c.stats {
+			stat := stats[def.ID]
+			// XXX Would be better to embed DefLine however resolver.Marshal
+			// does not support handling of anonymous embedded structs the way
```
mattnibs, line 109:
```suggestion
			// does not support handle untagged embedded structs the way
```

ppl/cmd/zar/index/ls.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	defs, err := ark.ReadDefinitions(context.TODO())
+	if err != nil {
+		return fmt.Errorf("zar index ls: %w", err)
+	}
+
+	w, err := c.output.Open()
+	if err != nil {
+		return fmt.Errorf("zar indexdefs: %w", err)
```
mattnibs, line 82:
```suggestion
		return fmt.Errorf("zar index ls: %w", err)
```

ppl/archive/chunk/writer.go
```diff
@@ -31,30 +34,59 @@ type Writer struct {
 	wroteFirst     bool
 }
 
-func NewWriter(ctx context.Context, dir iosrc.URI, order zbuf.Order, masks []ksuid.KSUID, opts zngio.WriterOpts) (*Writer, error) {
+type Options struct {
+	Order       zbuf.Order
+	Masks       []ksuid.KSUID
+	Definitions index.Definitions
+	ZngWriter   zngio.WriterOpts
```
nwt, line 41: Maybe sort? And maybe call ZngWriter just Zng for consistency with zio.WriterOpts.Zng?

ppl/archive/chunk/writer.go
```diff
@@ -131,6 +172,12 @@ func (cw *Writer) RecordsWritten() uint64 {
 	return cw.count
 }
 
+// Chunk returs the Chunk produced by the writer. Should only be called once
+// Close() has been called.
```
nwt, line 176:
```suggestion
// Chunk return s the Chunk produced by the writer. Should only be called after
// Close() has been called.
```

ppl/archive/chunk/writer_test.go
```diff
@@ -0,0 +1,59 @@
...
+	_, err = index.Find(context.Background(), zctx, chunk.ZarDir(), inputdef.ID, "100")
+	assert.Truef(t, zqe.IsNotFound(err), "expected err to be zqe.IsNotFound, got: %v", err)
+}
+
+func testWriteWithDef(t *testing.T, tzng string, defs ...*index.Definition) Chunk {
+	dir := iosrc.MustParseURI(t.TempDir())
+	w, err := NewWriter(context.Background(), dir, Options{Order: zbuf.OrderDesc, Definitions: defs})
+	require.NoError(t, err)
+	err = tzngio.WriteString(w, tzng)
+	require.NoError(t, err)
```
nwt, line 55:
```suggestion
	require.NoError(t, tzngio.WriteString(w, tzng))
```

ppl/archive/chunk/writer_test.go
```diff
@@ -0,0 +1,59 @@
...
+	ctx := context.Background()
+	require.NoError(t, w.Close(ctx))
```
nwt, line 57:
```suggestion
	require.NoError(t, w.Close(context.Background()))
```

ppl/archive/chunk/writer.go
```diff
@@ -31,30 +34,59 @@ type Writer struct {
 	wroteFirst     bool
 }
 
-func NewWriter(ctx context.Context, dir iosrc.URI, order zbuf.Order, masks []ksuid.KSUID, opts zngio.WriterOpts) (*Writer, error) {
+type Options struct {
```
nwt, line 37: Maybe call this WriterOpts for consistency with all the zio WriterOpts types?

ppl/archive/chunk/writer.go
```diff
@@ -31,30 +34,59 @@ type Writer struct {
...
+	seekIndex, err := seekindex.NewBuilder(ctx, seekIndexPath.String(), opts.Order)
 	if err != nil {
 		return nil, err
 	}
+	idxWriter := indexWriter(nopIndexWriter{})
+	// For a Chunk writer we only care about writing index defs whose input
+	// path is empty (i.e. references the chunk data itself).
+	if l := opts.Definitions.StandardInputs(); len(l) > 0 {
+		dir := chunkZarDir(dir, id)
+		idxWriter, err = index.NewMultiWriter(ctx, dir, l)
```
nwt, line 62: `l` is a surprising name for something with type `Definitions`.
```suggestion
	if defs := opts.Definitions.StandardInputs(); len(defs) > 0 {
		dir := chunkZarDir(dir, id)
		idxWriter, err = index.NewMultiWriter(ctx, dir, defs)
```

ppl/cmd/zar/index/ls.go
nwt, line 82:
```suggestion
		return fmt.Errorf("zar index ls: %w", err)
```

ppl/cmd/zar/index/ls.go
```diff
@@ -0,0 +1,145 @@
...
+		}
+		if def.ZQL != "" {
+			zql = def.ZQL
+		}
+
+		if c.stats {
+			stat := stats[def.ID]
+			// XXX Would be better to embed DefLine however resolver.Marshal
+			// does not support handle untagged embedded structs the way
+			// encoding/json does.
```
nwt, line 110:
```suggestion
			// does not support untagged embedded structs the way
			// encoding/json does.
```

zbuf/encoder.go
```diff
@@ -0,0 +1,46 @@
+package zbuf
+
+import (
+	"reflect"
+
+	"github.com/brimsec/zq/zng/resolver"
+)
+
+type Encoder struct {
```
alfred-landrum, line 9: I don't think we need this. It doesn't act like, say, a json.Encoder, because it has the internal loop over slice-like types, and the json.Encoder functionality for setting output options we can get by extending the resolver.MarshalContext.

ppl/cmd/zar/index/ls.go
mattnibs: Would be nice if charm did this automatically.

ppl/cmd/zar/index/add.go
```diff
@@ -0,0 +1,160 @@
...
+var Add = &charm.Spec{
+	Name:  "add",
+	Usage: "add [-R root] [options] [-z zql] [ pattern [ pattern ...]]",
+	Short: "add index rule for archive chunk files",
+	Long: `
+"zar index add" creates index files in a zar archive using one or more indexing
+rules.
+
+A pattern is either a field name or a ":" followed by a zng type name.
+For example, to index the all fields of type port and the field id.orig_h,
```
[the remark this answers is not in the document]
mattnibs: Are we planning to alias port types so this can be done?

ppl/cmd/zar/index/add.go
```diff
@@ -0,0 +1,160 @@
...
+	return c, nil
+}
+
+func (c *AddCommand) Run(args []string) error {
+	defer c.Cleanup()
+	if err := c.Init(&c.procFlags); err != nil {
+		return err
+	}
+	if len(args) == 0 && c.zql == "" && !c.ensure {
+		return errors.New("zar index: unless -ensure is specified, one or more indexing patterns must be specified")
```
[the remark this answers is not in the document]
mattnibs: I believe I was carrying on a pattern that was already present for `zar index`, happy to get rid of it though as it is cumbersome to carry on for every error case.

zbuf/encoder.go
mattnibs: I get the point about looping through slices as a user might want to have an array serialized (though this was convenient!). Will remove this feature.

As far as removing this, I still think this is a nice convenience utility that needs to exist in zbuf since resolver (a dependency of zbuf) has no notion of a zbuf.Reader/Writer. Without this (and adjusted to remove the slice reflection) the code in ls would look something like this:

```golang
rec, err := marshalcontext.MarshalRecord(v)
if err != nil {
    return err
}
if err := zwriter.Write(rec); err != nil {
    return err
}
```

Compared to:
```golang
if err := encoder.Encode(v); err != nil {
    return err
}
```

The go standard library saw it fit to include this convenience. Perhaps change the name to: `zbuf.MarshalWriter`?

Perhaps what I'm struggling with is that having Marshal in resolver strains some of the flexibility I'm looking for. Thoughts?

cli/outputflags/flags.go
```diff
@@ -25,7 +25,15 @@ func (f *Flags) Options() zio.WriterOpts {
 	return f.WriterOpts
 }
 
+func (f *Flags) simpleFlags(fs *flag.FlagSet) {
```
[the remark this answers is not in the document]
mattnibs: How about formatFlags?

zbuf/encoder.go
mattnibs: After I removed the reflection piece of Encoder the amount of code saved for having another struct in zbuf made it seem not worth keeping in light the objections stated. I did you as suggested.

Still think it would nice to have something that encoded a struct directly into a write stream, perhaps this is something we can keep in mind as the design of resolver continues to evolve.

cli/outputflags/flags.go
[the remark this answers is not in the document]
alfred-landrum: `formatFlags` sounds ok.

ppl/archive/index/definition.go
```diff
@@ -55,6 +55,10 @@ func ReadDefinition(ctx context.Context, u iosrc.URI) (*Definition, error) {
 	return def, err
 }
 
+func RemoveDefinition(ctx context.Context, dir iosrc.URI, id ksuid.KSUID) error {
+	return iosrc.RemoveAll(ctx, dir.AppendPath(defFilename(id)))
```
alfred-landrum, line 59:
```suggestion
	return iosrc.Remove(ctx, dir.AppendPath(defFilename(id)))
```

ppl/cmd/zar/index/add.go
```diff
@@ -0,0 +1,160 @@
+package index
```
alfred-landrum, line 1: I see you changed the subcommand to `create`; would you please also rename this file to create.go, and adjust the `Add`, `AddCommand`, and related refs to `Create`?

mattnibs: Shoot, my bad

cli/outputflags/flags.go
```diff
@@ -26,7 +26,15 @@ func (f *Flags) Options() zio.WriterOpts {
 	return f.WriterOpts
 }
 
+func (f *Flags) formatFlags(fs *flag.FlagSet) {
```
nwt, line 29: Call this `SetFormatFlags`.

cli/outputflags/flags.go
```diff
@@ -26,7 +26,15 @@ func (f *Flags) Options() zio.WriterOpts {
...
+	fs.StringVar(&f.Format, "f", "zng", "format for output data [zng,zst,ndjson,table,text,csv,zeek,zjson,zson,tzng]")
+	fs.BoolVar(&f.textShortcut, "t", false, "use format tzng independent of -f option")
+	fs.BoolVar(&f.forceBinary, "B", false, "allow binary zng be sent to a terminal output")
+}
+
 func (f *Flags) setFlags(fs *flag.FlagSet) {
+	f.formatFlags(fs)
```
nwt, line 36: This causes `SetFlagsWithFormat` to add these flags, and I don't think we want that. Moving this call to `SetFlags` would fix that.

mattnibs: That was the way this working before this pr. `SetFlagsWithFormat` was just setting a default value for Format.

mattnibs: Actually though, I can see what you are saying. I'll make the change.

alfred-landrum (approved):

PR #1799, opened 2020-12-12
Simplify recruiter API and convert to long-poll register

marktwallace:
The goal is to close the known "holes" in the recruiter system by only allowing a worker to be recruited when it is holding open an HTTP long-poll with the recruiter. This will allow us to scale the K8s worker deployment up and down without getting faulty registration data.


marktwallace:
Changing branches, so closing this

Issue #1789, opened 2020-12-10
zson parser needs to check for mixed type arrays


mccanne:
When the zson parser sees a mixed type array, it needs to create an array of union...

PR #1806, opened 2020-12-12
intial zson parser and reader

### LINKED ISSUE #1679: zson reader
Now that we have a zson writer, we need the read side.

The reader will be more flexible compared to what the writer generates, e.g., allowing ergonomic shorthand by dropping field names when you have type decorator on a record, and so forth.

mccanne:
This commit adds a parser and reader to package zson.  The spec
has been updated a bit reflecting lessons learned during the
implementation of the parser.

The type system in ZSON requires an interesting semantic analysis
of each ZSON value object.  The parser scans the input text and
parses it into an AST.  The AST is then traversed by the semantic
analysis pass to convert the AST into an intermediate value form.
This pass uses the distributed type information in the AST to
determine an exact type for each constituent value and typedef
in a nested value.  This fully-typed value is then passed to the
Builer and transformed into a zng represenation with a simple
traversal of the nested value.

The new ZSON ast nodes in ast/zson.go have been designed to be
zql-ready for the peg grammar to generate and pass to the search
API as zql queries with embedded ZSON constants of arbitrary
complexity and size.

A large suite of "boomerang" tests was created by traversing
the zq repo and finding any ztest yaml file that had tzng in it.
For each such hit, a file was created in zson/ztests with
generic boilerplate that converts the tzng to zson and the
zson is read as input and output again as zson making sure that
the two zson outputs are the same.

Closes #1679 #1760 #1789


mccanne:
@mattnibs since all these tests are auto-generated from other tests in the repo, my preference would be to just leave them as is and not worry about how they are named etc.  If/when we improve ztest further or change how we do the tests, we can remove them all then.  The tests will be really helpful to catch regressions in zson and zql without requiring a huge investment to maintain each one by hand.  I could change the naming scheme and put a comment in each autogenerate file so its clear that they are dynamically generated.  Would this address your concerns?

mccanne:
I'll move the autogenerated tests to ztests/auto since there are handcrafted tests to and it's hard to tell them apart.

zson/values.go
```diff
@@ -0,0 +1,566 @@
...
+	"errors"
+	"io"
+	"net"
+	"strconv"
+	"time"
+
+	"github.com/brimsec/zq/ast"
+)
+
+func (p *Parser) ParseValue() (ast.Value, error) {
```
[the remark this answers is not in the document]
henridf: Oh, I see, there's also the other half in `types.go`. Maybe this is saying that parser deserves to be its own package, but this is nit land, so feel free to disregard.

ast/zson.go
```diff
@@ -0,0 +1,180 @@
...
+		Op       string  `json:"op"`
+		Elements []Value `json:"elements"`
+	}
+	Set struct {
+		Op       string  `json:"op"`
+		Elements []Value `json:"elements"`
+	}
+	Union struct {
+		Op       string `json:"op"`
+		Selector Type   `json:"selector"`
```
henridf, line 74: This looks unused

henridf (commented):
First looky, still want to look some more. looks good!

zng/docs/zson.md
```diff
@@ -324,6 +324,13 @@ than its previous definition though multiple definitions of the same
 type are legal (thereby allowing for concatenation of otherwise
 independent sequences).
 
+One decorator is allowed per value except for nested type-union values, which
+may additional decorators to successively refine the union type for unions vals
+that live inside other union types. This allows an already-decorated value to be
+further decorated with its union type and provides as a means to distinguish
+a union value's precise member type when it is other ambiguous as described in
```
nwt, line 331:
```suggestion
further decorated with its union type and provides a means to distinguish
a union value's precise member type when it is otherwise ambiguous as described in
```

zng/docs/zson.md
```diff
@@ -500,6 +515,11 @@ Where there is ambiguity, a decorator may resolve the ambiguity as in
 ```
 123 (int8) (int32, int8)
 ```
+Where a union value is inside of nested union types, multiple decorators are
+needed to resolve the ambiguity, e.g.,
+```
+"hello, world" (int32, string) ((int32,string),[int32])
```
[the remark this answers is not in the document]
mccanne: Correct.  But only for unions.  One decorator for other types.  I don't think I check for this so 
```
(int8) (uint8) 
``` 
would currently be accepted and the int8  would override but the parent would think it's uint8.  Bug.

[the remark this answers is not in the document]
mccanne: Correct.  If the outer type isn't a union then you will get a type error with multiple conflicting decorators.

zson/zson_test.go
```diff
@@ -0,0 +1,184 @@
...
+	analyzer := zson.NewAnalyzer()
+	return analyzer.ConvertValue(zctx, val)
+}
+
+func TestZsonAnalyzer(t *testing.T) {
+	zctx := resolver.NewContext()
+	val, err := analyze(zctx, testFile)
+	require.NoError(t, err)
+	assert.NotNil(t, val)
+	//pretty.Println(val)
```
nwt, line 170:
```suggestion
```

zson/values.go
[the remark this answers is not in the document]
mccanne: when editing and searching it was a lot easier having the types and the values in separate files.  I meant to rename the files parser-types.go and parser-values.go.   I could consolidate them too now that most of the editing is done.  Unless there are strong opinions, I will do the former.

zson/builder.go
```diff
@@ -0,0 +1,241 @@
...
+		return nil
+	case *zng.TypeOfDuration:
+		v, err := time.ParseDuration(val.Text)
+		if err != nil {
+			return fmt.Errorf("invalid duration: %s", val.Text)
+		}
+		b.AppendPrimitive(zng.EncodeDuration(int64(v)))
+		return nil
+	case *zng.TypeOfTime:
+		t, err := time.Parse(time.RFC3339Nano, val.Text)
```
henridf, line 90: Maybe later there's an optimization where we can avoid parsing primitive values twice (I know, zng is for performance, but i think having zson be fast will be valuable).

zson/analyzer.go
```diff
@@ -0,0 +1,699 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/brimsec/zq/ast"
+	"github.com/brimsec/zq/zng"
+	"github.com/brimsec/zq/zng/resolver"
+)
+
+type Value interface {
+	TypeOf() zng.Type
```
henridf, line 13: it seems like these don't need to be exported?

henridf (approved):
:shipit:

zson/analyzer.go
```diff
@@ -0,0 +1,699 @@
...
+		}
+	} else if isFloaty(typ) {
+		if isFloaty(cast) {
+			return cast, nil
+		}
+	}
+	return nil, fmt.Errorf("type mismatch: '%s' cannot be used as '%s'", typ.ZSON(), cast.ZSON())
+}
+
+func isStringy(typ zng.Type) bool {
```
henridf, line 253: Can these is<Type> funcs be moved to zng/type.go alongside the Is<Type> funcs that are already present there?

ast/zson.go
```diff
@@ -0,0 +1,174 @@
...
+		Op      string  `json:"op"`
+		Entries []Entry `json:"entries"`
+	}
+	Entry struct {
+		Key   Value `json:"key"`
+		Value Value `json:"value"`
+	}
+	TypeValue struct {
+		Op   string `json:"op"`
+		Type Type   `json:"type"`
```
henridf, line 86: maybe: 

```suggestion
		Value Type   `json:"value"`
```

(I came to this after looking at some zson ast while working on reshape...)

ast/zson.go
```diff
@@ -0,0 +1,174 @@
+package ast
+
+type Value interface {
+	valueNode()
+}
+
+const (
+	ImpliedValueOp = "implied_value"
+	CastValueOp    = "cast_value"
+	DefValueOp     = "def_value"
```
nwt, line 10: Nit: We usually (but admittedly not always) have a common prefix on constants like these and those below (i.e., OpImpliedValue), and I favor that scheme.

zng/docs/zson.md
```diff
@@ -324,6 +324,13 @@ than its previous definition though multiple definitions of the same
 type are legal (thereby allowing for concatenation of otherwise
 independent sequences).
 
+One decorator is allowed per value except for nested type-union values, which
+may include additional decorators to successively refine the union type for union values
+that live inside other union types. This allows an already-decorated value to be
+further decorated with its union type and provides as a means to distinguish
```
nwt, line 330:
```suggestion
further decorated with its union type and provides a means to distinguish
```

zng/docs/zson.md
```diff
@@ -777,9 +797,13 @@ the defines their type.
 
 <eos> = .
 
-<value> = <primary> | <primary> <decorator>
+<value> = <any> | <any> <val-typedef> | <any> <decorators>
 
-<primary> = <primitive> | <record> | <array> | <set> |
+<val-typedef> = "(" "=" <type-name> ")"
+
+<decorators> = "(" <type> ")" | <decorators> "(" <type ")"
```
nwt, line 804:
```suggestion
<decorators> = "(" <type> ")" | <decorators> "(" <type> ")"
```

zng/docs/zson.md
```diff
@@ -808,19 +832,17 @@ the defines their type.
 
 <map> = "|{" <mlist> "}|"  |  "|{"  "}|"
```
nwt, line 833: @mccanne: Out of curiosity, could maps be made more like records rather than lists, like this?
```
|{ <key>: <value>, <key>: <value>, ... }|
```

zng/docs/zson.md
```diff
@@ -808,19 +832,17 @@ the defines their type.
...
-
-<decorator> = "(" <type> ")" | "(" "=" <type-name> ")"
+<type-value> = "(" <type> ")"
 
 <type> = <primitive-type> | <record-type> | <array-type> | <set-type> |
             <union-type> | <enum-type> | <map-type> | <type-type> |
             <type-def> | <type-name>
 
-<primitive-type> = uint8 | uint16 | etc as defined above excepting "type"
+<primitive-type> = uint8 | uint16 | etc as defined above including "type"
```
nwt, line 845:
```suggestion
<primitive-type> = uint8 | uint16 | etc. as defined above including "type"
```

zson/builder.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+	return fmt.Errorf("unknown ast type: %T", val)
+}
+
+func (b *Builder) buildPrimitive(val *Primitive) error {
+	switch zng.AliasedType(val.Type).(type) {
+	case *zng.TypeOfUint8, *zng.TypeOfUint16, *zng.TypeOfUint32, *zng.TypeOfUint64:
+		v, err := strconv.ParseUint(val.Text, 10, 64)
+		if err != nil {
+			return fmt.Errorf("invalid unsigned integer: %s", val.Text)
```
nwt, line 71: Should we quote val.Text, perhaps with %q, in all of these errors?

zson/builder.go
nwt: @henridf: Do you mean an optimization to go directly from val.Text to a zcode.Bytes value without a trip through the Go representation (like time.Time here)?

zson/builder.go
```diff
@@ -0,0 +1,241 @@
...
+			return fmt.Errorf("invalid bytes: %s", s)
+		}
+		bytes, err := hex.DecodeString(s[2:])
+		if err != nil {
+			return fmt.Errorf("invalid bytes: %s (%s)", s, err.Error())
+		}
+		b.AppendPrimitive(zcode.Bytes(bytes))
+		return nil
+	case *zng.TypeOfString, *zng.TypeOfError:
+		b.AppendPrimitive(zng.EncodeString(val.Text))
```
nwt, line 124: Maybe return an error if unicode/utf8.ValidString doesn't like val.Text?

zson/builder.go
```diff
@@ -0,0 +1,241 @@
...
+		return nil
+	case *zng.TypeOfNet:
+		_, net, err := net.ParseCIDR(val.Text)
+		if err != nil {
+			return fmt.Errorf("invalid network: %s (%s)", val.Text, err.Error())
+		}
+		b.AppendPrimitive(zng.EncodeNet(net))
+		return nil
+	case *zng.TypeOfNull:
+		b.AppendPrimitive(nil)
```
nwt, line 144: Error if `val.Text != ""`?

zson/builder.go
```diff
@@ -0,0 +1,241 @@
...
+		if err != nil {
+			return fmt.Errorf("invalid bytes: %s (%s)", s, err.Error())
+		}
+		b.AppendPrimitive(zcode.Bytes(bytes))
+		return nil
+	case *zng.TypeOfString, *zng.TypeOfError:
+		b.AppendPrimitive(zng.EncodeString(val.Text))
+		return nil
+	case *zng.TypeOfBstring:
+		b.AppendPrimitive(unescapeHex([]byte(val.Text)))
```
nwt, line 127: Maybe return an error if val.Text contains a malformed \ sequence?

zson/builder.go
henridf: There's a `time.Parse(..)` call in parser-values.go that's done just as a grammar test.  Then this one to actually build the `zcode.Bytes`. There's probably a few different ways to coalesce those into one, but as I noted above this can come later I think.

zson/parser-values.go
```diff
@@ -0,0 +1,566 @@
+package zson
```
nwt, line 1: parservalues.go would be a more conventional name.

zson/parser-types.go
```diff
@@ -0,0 +1,367 @@
+package zson
```
nwt, line 1: parsertypes.go would be a more conventional name.

zson/lexer.go
```diff
@@ -0,0 +1,299 @@
...
+				}
+				continue
+			}
+		}
+		return nil
+	}
+}
+
+func isNewline(r rune) bool {
+	// from stackoverflow
```
nwt, line 169: http://www.unicode.org/versions/Unicode13.0.0/ch05.pdf#G10213 is a better reference.

zson/builder.go
mccanne: Agreed.  I hadn't first worried about this because I wasn't think about this being on the critical path.  Also, the design of how constants imply their type is necessary heavy-weight.  There's probably some regexp trickery would could come up with here that would classify the type in analyzer then do the parsing work in builder.

zson/analyzer.go
mccanne: I suspect we will end up using zson.Value in the planner/optimizer/jit-compiler rather than ast (which will have unresolved types).  I think I'll leave it as rather than go through the typing exercise now and then again later...

zson/analyzer.go
mccanne: Good idea. Those existing functions already to what I need if I call ID() and add IdError to stringy.

ast/zson.go
mccanne: Yes!  We're getting the hang of this.  I had it right in zson.TypeValue.

ast/zson.go
mccanne: All of the ast nodes in package ast should have *Op constants so let's hold off on this rename until we do that.  There is code all over the puts a string in for the Op field and somethimes it's left foo so I like the pattern ...
```
return &Thing{
  Op: ThingOp,
  Rest: ...
}
```
We can bikeshed whether its Op* or *Op when we get to this :-)
p.s. Op here doesn't mean operator but rather Opcode for the JSON object type that is being coded.  When we switch this to use ZSON instead of JSON encoding, we can elimiate Op everywhere and use type inference.  But we can't do this until the (javascript) client can do a ZSON.Parse() instead of JSON.Parse() to recover what will have been previously encoded in the "op" key of the JSON encoding.

zson/zson_test.go
```diff
@@ -0,0 +1,182 @@
...
+	})
+	return inputs, err
+}
+
+func parse(path string) (ast.Value, error) {
+	file, err := fs.Open(path)
+	if err != nil {
+		return nil, err
+	}
+	parser, err := zson.NewParser(file)
```
nwt, line 138: Maybe define a constant containing the contents of test.zson, create a strings.Reader from it, and pass that to zson.NewParser?

If not, `defer file.Close()`.

zng/docs/zson.md
mccanne: That should work.  I'll give it a spin.

zson/parser-values.go
```diff
@@ -0,0 +1,566 @@
...
+
+func (p *Parser) ParseValue() (ast.Value, error) {
+	v, err := p.matchValue()
+	if err == io.EOF {
+		err = nil
+	}
+	return v, err
+}
+
+func clean(err error) error {
```
nwt, line 22: Maybe call this discardEOF or ignoreEOF?

zng/docs/zson.md
mccanne: This will involve changes to the spec, parser, and writer so I will merge this PR and put up a subsequent PR.  I think it looks better and is a good idea and I don't see why it won't work syntactically since it's parallel to a record and has differentiating grouping symbols.

mccanne: issue #1824

zson/builder.go
mccanne: This is checked in the analyzer.

zson/zson_test.go
mccanne: I used it to debug.  Very handy to edit the file then not check in the changes.

zson/parser-types.go
mccanne: See issue #1825

mccanne:
> @mattnibs since all these tests are auto-generated from other tests in the repo, my preference would be to just leave them as is and not worry about how they are named etc. If/when we improve ztest further or change how we do the tests, we can remove them all then. The tests will be really helpful to catch regressions in zson and zql without requiring a huge investment to maintain each one by hand. I could change the naming scheme and put a comment in each autogenerate file so its clear that they are dynamically generated. Would this address your concerns?

See #1814

PR #1808, opened 2020-12-12
fix bug with rows larger than max read buffer

### LINKED ISSUE #1803: "malformed zng record" while running large aggregation
Repro is with `zq` commit `d62f4b1b` and the attached test file:

[repro.zng.gz](https://github.com/brimsec/zq/files/5682298/repro.zng.gz)

```
$ zq 'collect(name) by value' repro.zng.gz > /dev/null
malformed zng record
```

The original repro was actually involving the much larger (23 GB) public data set at https://opendata.rapid7.com/sonar.fdns_v2/2020-11-27-1606435855-fdns_a.json.gz. However, I've narrowed it down to just what's in the attachment above.

For whatever reason, what's in the attachment is the magic amount. If I trim either the top or bottom record from the file, it doesn't repro. So something is making it tip over at that point.

```
$ zq -t 'count()' repro.zng.gz 
#0:record[count:uint64]
0:[1650832;]

$ zq 'head 1650831 | collect(name) by value' repro.zng.gz > /dev/null
[no error]

$ zq 'tail 1650831 | collect(name) by value' repro.zng.gz > /dev/null
[no error]
```

mccanne:
This commit adds a better error message to the zng reader with the
length of a value exceeds the max read size.   We also added flags
to set the max read buffer (and target read buffer) sizes.

Closes #1803

### DIFF
--- a/cli/inputflags/flags.go
+++ b/cli/inputflags/flags.go
@@ -21,10 +21,10 @@ import (
 
 type Flags struct {
 	zio.ReaderOpts
+	ReadMax  auto.Bytes
+	ReadSize auto.Bytes
 	// The JSON type config is loaded from the types filie when Init is called.
 	jsonTypesFile string
-	ReadSize      auto.Bytes
-	ReadMax       auto.Bytes
 }
 
 func (f *Flags) Options() zio.ReaderOpts {
@@ -62,7 +62,7 @@ func (f *Flags) Init() error {
 	}
 	f.Zng.Size = int(f.ReadSize.Bytes)
 	if f.Zng.Size < 0 {
-		return errors.New("max read buffer size must be greater than zero")
+		return errors.New("target read buffer size must be greater than zero")
 	}
 	return nil
 }
--- a/cli/inputflags/flags.go
+++ b/cli/inputflags/flags.go
@@ -2,17 +2,20 @@ package inputflags
 
 import (
 	"encoding/json"
+	"errors"
 	"flag"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"regexp"
 
+	"github.com/brimsec/zq/cli/auto"
 	"github.com/brimsec/zq/zbuf"
 	"github.com/brimsec/zq/zio"
 	"github.com/brimsec/zq/zio/azngio"
 	"github.com/brimsec/zq/zio/detector"
 	"github.com/brimsec/zq/zio/ndjsonio"
+	"github.com/brimsec/zq/zio/zngio"
 	"github.com/brimsec/zq/zng/resolver"
 )
 
@@ -20,6 +23,8 @@ type Flags struct {
 	zio.ReaderOpts
 	// The JSON type config is loaded from the types filie when Init is called.
 	jsonTypesFile string
+	ReadSize      auto.Bytes
+	ReadMax       auto.Bytes
 }
 
 func (f *Flags) Options() zio.ReaderOpts {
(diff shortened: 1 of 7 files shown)


philrz:
I tested out the original repro with the branch and the improved error message is definitely 👍 . For kicks I then tried fiddling with the new options to see if I could see the operation complete with the original 23 GB data set. I admittedly had not sought guidance as to how to best set these options.

FWIW increasing `readmax` alone didn't seem to affect the outcome.

```
$ time zq -readmax 1000MB 'collect(name) by value' 2020-11-27-1606435855-fdns_a.zng.gz > rapid7_collected.zng
large value of 21484506 bytes exceeds maximum read buffer (10485760 bytes)

real	193m53.868s
user	216m51.091s
sys	6m18.596s
```

At this point I figured I'd see what happens if I increased the other as well, and here I triggered an error.

```
$ time zq -readmax 1000MB -readsize 10MB 'collect(name) by value' repro.zng.gz > rapid7_collected.zng
repro.zng.gz: format detection error
	tzng: line 1: strconv.ParseUint: parsing "\xf6\x04\x04name\x10\ttimestamp\x10\x04type\x10\x05value\x10\xfd\x00\xda\u007f\xdd\x1c\xfc-\x17": invalid syntax
	ndjson: line 1: Unknown value type
	zjson: line 1: invalid character 'ö' looking for beginning of value
	zng: buffer exceeded max size trying to infer input format
	azng: line 1: zng descriptor out of range
	parquet: auto-detection not supported
	zst: auto-detection not supported
```

mccanne:
As we discussed @philrz, I think we should add a max-row-size parameter to limit the growth of rows in aggregators like union and collect (and similar future aggregators).  Rather than do this on this PR, I created #1813

zio/zngio/ztests/big-value.yaml
```diff
@@ -0,0 +1,30 @@
...
+  cat out.zng out.zng out.zng out.zng > out2.zng
+  mv out2.zng out.zng
+  cat out.zng out.zng out.zng out.zng > out2.zng
+  mv out2.zng out.zng
+  cat out.zng out.zng out.zng out.zng > out2.zng
+  mv out2.zng out.zng
+  cat out.zng out.zng out.zng out.zng > out2.zng
+  mv out2.zng out.zng
+  cat out.zng out.zng out.zng out.zng > out2.zng
+  mv out2.zng out.zng
```
nwt, line 16:
```suggestion
  for i in {1..7}; do
    cat out.zng out.zng out.zng out.zng > out2.zng
    mv out2.zng out.zng
  done
```

cli/inputflags/flags.go
```diff
@@ -47,6 +56,14 @@ func (f *Flags) Init() error {
 		}
 		f.JSON.TypeConfig = c
 	}
+	f.Zng.Max = int(f.ReadMax.Bytes)
+	if f.Zng.Max < 0 {
+		return errors.New("max read buffer size must be greater than zero")
+	}
+	f.Zng.Size = int(f.ReadSize.Bytes)
+	if f.Zng.Size < 0 {
+		return errors.New("max read buffer size must be greater than zero")
```
nwt, line 65:
```suggestion
		return errors.New("target read buffer size must be greater than zero")
```

cli/inputflags/flags.go
```diff
@@ -2,24 +2,29 @@ package inputflags
...
+	"github.com/brimsec/zq/zio/zngio"
 	"github.com/brimsec/zq/zng/resolver"
 )
 
 type Flags struct {
 	zio.ReaderOpts
 	// The JSON type config is loaded from the types filie when Init is called.
 	jsonTypesFile string
+	ReadSize      auto.Bytes
+	ReadMax       auto.Bytes
```
nwt, line 27: Nit: Keep exported fields at the top.
```suggestion
	zio.ReaderOpts
	ReadMax  auto.Bytes
	ReadSize auto.Bytes
	// The JSON type config is loaded from the types filie when Init is called.
	jsonTypesFile string
```

nwt (approved):

PR #1811, opened 2020-12-14
copy columns when looking up TypeRecord

mccanne:
This commit makes a copy of the zng.Columns when looking up
a TypeRecord so that the caller mutate its underlying zng.Column
slice without harming the type context.


henridf (approved):
(I remember some place where I'd run into this and had made a copy somewhere after a call to LookupTypeRecord... I can't find it though. Maybe it got rewritten. Anyway, two copies is better than none!).

Issue #1326, opened 2020-09-19
Restore text representation before printing as CSV


philrz:
Repro is with `zq` commit `2fe0e93`.

There seem to be several minor variations in the world when it comes to formatting CSV output. However, short of adhering to any one true spec, we're currently letting through some changed values that likely don't belong there. Here I'm starting my data as NDJSON and comparing our output with the top Google hit for "json to csv" (https://www.convertcsv.com/json-to-csv.htm).

Backslashes:

```
$ echo '{"BackSlash":"\\"}' | zq -f csv -
BackSlash
\\
```

The JSON-to-CSV tool gives:

```
BackSlash
\
```

A lone hyphen (maybe will be fixed by #1325?)

```
$ echo '{"Hyphen":"-"}' | zq -f csv -
Hyphen
\u002d
```

The JSON-to-CSV tool gives:

```
Hyphen
-
```

A null value:

```
$ echo '{"a": "a", "nothing":null, "z": "z"}' | zq -f csv -
a,nothing,z
a,-,z
```

The JSON-to-CSV tool gives:

```
a,nothing,z
a,,z
```

Overall I get the impression that we just need to undo some of our escaping and handling of special characters before doing the final output as CSV. I discussed this with @mccanne and he summarized:

> I'm using the same logic as the text output so we need to fix this. I don't know that we have the correct methods on the zng types to just print a text representation without the escape business.

nwt:
We currently escape backslashes to distinguish them from the \xhh escape sequence we use to represent bytes that don't fit within the current character encoding (i.e., non-ASCII bytes for `zq -U=false`, and bytes that aren't part of a valid UTF-8 sequence for `zq -U=true`.

Here are some options that would let us stop escaping backslashes.
1. 
1. Escape them only in strings that contain \xhh sequences.


We could simply stop escaping backslashes for CSV output, but that renders any \xhh sequence ambiguous

One way we could stop escaping backslashes is to replace the boolean -U flag with a -charset flag that could be set to ascii (does what -U=false does now), utf8 (does what -U=true does now), or none (no escaping ever).

nwt:
We escape backslashes to distinguish them from the `\xhh` escape sequences we use to represent bytes that don't fit within the current character encoding (i.e., non-ASCII bytes for `zq -U=false`, and bytes that aren't part of a valid UTF-8 sequence for `zq -U=true`, but we've got options if we want to limit or avoid this for CSV.

[...]

PR #1818, opened 2020-12-15
Add CSV tests for primitive types

nwt:
This is part of fixing #1326. Adding these separately from the fix will make the effects of the fix clearer.


zio/csvio/ztests/float.yaml
```diff
@@ -0,0 +1,22 @@
+zql: '*'
+
+input: |
+  #0:record[float64:float64]
+  0:[-;]
+  0:[0.0;]
+  0:[1.797693134862315708145274237317043567981e+308;]
+  0:[1.797693134862315708145274237317043567981e-324;]
```
mccanne, line 8: You're testing tzng parsing / overflow handling here rather than the csv output.

zio/csvio/ztests/error.yaml
```diff
@@ -0,0 +1,21 @@
...
+  0:[";]
+  0:[,;]
+  0:[\u002d;]
+  0:[\u005c;]
+
+output-flags: -f csv
+
+output: |
+  error
+  -
```
mccanne, line 16: I don't know that this has any meaning except maybe to a zeek person who knows zeek TSV.  I would be inclined to map nulls to csv empty string rather than minus.

mccanne: I guess the code was already written this way...

mccanne (approved):
csv formats are notoriously incompatible and there's no way to do a boomerang in and out of zng so I would just try to get the basics working here and not sweat the details right now.  If users run into problems importing our csv-output into specific scenarios, they'll know and it will be very easy to fix.

zio/csvio/ztests/error.yaml
nwt: This should indeed be an empty string, not "-", and is one of the problems identified by #1326. It'll be fixed in a subsequent PR.

zio/csvio/ztests/float.yaml
nwt: True, though my aim here is to fix the current CSV representation of these values, which we should change (because they're `179769313486231570000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000` and `0`)!

Issue #1785, opened 2020-12-09
Shaping functions


henridf:
Per https://github.com/brimsec/zq/issues/1783#issuecomment-751790854 and https://github.com/brimsec/zq/issues/1783#issuecomment-754087671

philrz:
For a walk through the shaping functions in action, the best place to look will be the configs that come out of #1871 and #2104 which will serve as reference examples, since the Brim app will ultimately become dependent on them.

henridf:
done in #1984

Issue #1619, opened 2020-11-15
type value operators


mccanne:
Type values should be more flexible in terms of usage in expressions.  You should be able to say `.foo` and get a new type that is the type of the field foo in type value `t`.  Also, you should be able to search type values as a sting as in the `filter` proc, e.g., this would cut the top-level field `metadata` but match type values that have nested fields with the string `metadata`.
```
cut -c metadata | count() by typeof(.) | filter metadata"
```

Also, to be efficient on this front, zng.TypeType should have a zng.Type value inside of it in addition to the string.  The creation of this type needs to be handled by resolver.Context as with records/arrays/etc but this is a bit different (because it's a primitive value not a type and we assume primitive values can be parsed without a type context) and the Type pointer could be filled in lazily when accessing the value.  That said, perhaps we can revisit some of these assumptions and make this all work better.

PR #1820, opened 2020-12-15
Baby step towards storing type associated with type value

henridf:
@mccanne this is a first step toward what you outlined in #1619 and which I also need for #1785 : 

> Also, to be efficient on this front, zng.TypeType should have a zng.Type value inside of it in addition to the string.

The `eq` func in typevector/type.go should probably be lifted to `zng/type.go`. If we go this route, I would also look for any other occurences of type comparisons that may need to use this. But I wanted to see first if this route makes sense?

(And, the next step would be to fill in the Type pointer lazily for values that originated from zson, and where the zng.Type is not available at value creation time... )

### DIFF
--- a/zio/zsonio/typemap.go
+++ b/zio/zsonio/typemap.go
@@ -15,7 +15,7 @@ func (t typemap) known(typ zng.Type) bool {
 	if _, ok := t[typ]; ok {
 		return true
 	}
-	if _, ok := typ.(*zng.TypeOfType); ok {
+	if _, ok := typ.(*zng.TypeType); ok {
 		return true
 	}
 	if _, ok := typ.(*zng.TypeAlias); ok {
--- a/zio/zsonio/writer.go
+++ b/zio/zsonio/writer.go
@@ -101,7 +101,7 @@ func (w *Writer) writeValue(indent int, typ zng.Type, bytes zcode.Bytes, parentK
 		null, err = w.writeMap(indent, t, bytes, known)
 	case *zng.TypeEnum:
 		err = w.write(t.ZSONOf(bytes))
-	case *zng.TypeOfType:
+	case *zng.TypeType:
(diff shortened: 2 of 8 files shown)


henridf:
(I wanted to open this as a draft  but didn't and apparently you can't switch back...)

mccanne:
@henridf I guess the zson-parser merge is making these diffs hard to follow.  Are you proposing moving TypeOfType back into reolver.Context?  I went back and forth on that and decided it was a square peg ini a round hole and settled on this approach instead...
https://github.com/brimsec/zq/blob/master/zson/table.go
This ill let you look up any efficient lookup any zng.Value TypeValue using the zng.Type you want the TypeValue for.  So no need to have a TypeTypeValue that implements zng.Type.  e.g., see how the typoeof() function does this... https://github.com/brimsec/zq/blob/88b821dbf97f67db06e31ce797049ebec687ba1d/expr/function/function.go#L180

henridf:
@mccanne I restored this branch to something readable after github's automatic base change... 

> This will let you look up any efficient lookup any zng.Value TypeValue using the zng.Type you want the TypeValue for. 

My goal here is to the reverse: retrieve the zng.Type given a type value zng.Value. This is to support reshape, but also things like dot-operator on type values as you outlined in #1619.

That said your comment made me realize that while I'd seen `zson.TypeTable`, I missed its `LookupType` method which means this PR shouldn't be needed. I guess `zson.(*TypeTable).LookupType()` will need a resolver.Context to create the `zng.Type` (and it should probably move to `zng/` at some point). Does this sound right?

henridf:
>  I guess zson.(*TypeTable).LookupType() will need a resolver.Context to create the zng.Type (and it should probably move to zng/ at some point). Does this sound right?

I'll go ahead and try this on another PR...

PR #1821, opened 2020-12-15
update helm-install commands in eks-test-deploy

marktwallace:
This fixes the eks-test-deploy failures in master

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -153,9 +153,9 @@ helm-install-worker:
 	--set useCredSecret=false
 
 make helm-uninstall:
-	helm uninstall worker || :
-	helm uninstall root || :
-	helm uninstall recruiter || :
+	-helm uninstall worker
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -153,9 +153,9 @@ helm-install-worker:
 	--set useCredSecret=false
 
 make helm-uninstall:
-	helm uninstall worker
-	helm uninstall root
-	helm uninstall recruiter
+	helm uninstall worker || :
+	helm uninstall root || :
+	helm uninstall recruiter || :
```
nwt, line 158: https://www.gnu.org/software/make/manual/html_node/Errors.html
```suggestion
	-helm uninstall worker
	-helm uninstall root
	-helm uninstall recruiter
```

nwt (approved):

PR #1822, opened 2020-12-15
zqd PostgresDB

### LINKED ISSUE #1752: zqd database support
When zqd runs in a service cluster, it needs to use a database for metadata & resource persistence. In https://github.com/brimsec/zq/pull/1693 , the operations to create & modify spaces were grouped into `filedb.go`, which identifies the interface & constraints we'll need to match when backed by a database. We should target Postgresql 11 as our backing database, to line up with our plans for Temporal.

For this task:
- zqd should support configuration that enables database usage and the database endpoint, port, and credentials.
- We should add a new make target that runs tests verifying operations against a database enabled zqd instance.
- We should add a Github Action job that brings up a postgres instance (using docker or docker-compose), and then runs tests using the new make target.

For dev/test, if we need to automatically apply migrations, we should use this Go migration library: https://github.com/golang-migrate/migrate , as a couple of us have good experience with it in the past. When we use a database in production services, migrations will be applied by operators, not automatically during code deploys.

mattnibs:
In addition to storing SpaceRows using a filedb, add support for storing
SpaceRows with postgres. In package apiserver create PostgresDB struct that
implements the DB interface.

A lot of plumbing in this pr to make this work with respect to tests; ran both
locally and in ci:

- Add pgctl command for running common tasks on a zqd postgresdb. Currently this
has only options for creating a destroying a testdb, but eventually will be
extended to run migrations and introspect a database.
- zqd listen: Add -postgres arg for specifying database. Arg -postgres.DB also
added to specify database in instance (this will mostly be used for tests).
- Use package golang-migrate/migrate to create migrations folder in
ppl/zqd/migrations
- Add startup.sh script: To be used for postgres+zqd ztests. Uses pgctl to
create and destroy a testdb then launches zqd listen with the correct
configuration
- ztest: Plumbing for specifing environment variabes for shell scripts.
- ztest: Add tag feature. ztest yaml files that use the tag field will only run
if the tag matches a ZTEST_TAG env variable. This was needed so make test-system
didn't end up running postgres tests.

Closes #1752


mattnibs:
There's definitely ways to improve the usability for working with postgres, for this pr and I just wanted to get something working with the idea that we can improve usability in future iterations.

cmd/pgctl/main.go
```diff
@@ -0,0 +1,18 @@
+package main
```
alfred-landrum, line 1: Would you please move the `pgctl` directory to `ppl/cmd`?

ppl/zqd/migrations/20201204143155_create_space_table.down.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE IF EXISTS space;
```
alfred-landrum, line 1: Would you move the `migrations` directory under `ppl/zqd/postgres`?

ppl/zqd/ztests/postgres/duplname.yaml
```diff
@@ -0,0 +1,15 @@
+tag: postgres
```
alfred-landrum, line 1: How about creating a new `ppl/zqd/postgres/ztest` directory for all of these tests?

ppl/zqd/apiserver/postgresdb.go
```diff
@@ -0,0 +1,80 @@
+package apiserver
```
alfred-landrum, line 1: In a subsequent PR, we could rearrange the manager & filedb code so this postgres implementation could live under `ppl/zqd/postgres`.

ppl/zqd/apiserver/postgresdb.go
```diff
@@ -0,0 +1,80 @@
...
+		return nil, err
+	}
+	return &PostgresDB{db}, nil
+}
+
+func (d *PostgresDB) CreateSpace(ctx context.Context, row SpaceRow) error {
+	if row.ID == "" {
+		return zqe.ErrInvalid("row must have an id")
+	}
+
```
alfred-landrum, line 29: Could you verify that `row.ParentID` isn't set? In the future, we could try to unify CreateSpace & CreateSubspace here & in the filedb.

ppl/zqd/ztests/postgres/startup.sh
```diff
@@ -0,0 +1,33 @@
...
+portdir=$(mktemp -d)
+mkdir -p data
+
+zqd listen -l=localhost:0 \
+  -portfile="$portdir/zqd" \
+  -postgres=$PG_TEST \
+  -postgres.db=$dbname \
+  -data=data \
+  -loglevel=warn \
+  -suricataupdater=true &> zqd.log &
```
alfred-landrum, line 27: I think you can remove the `-suricataupdater` flag.

Makefile
```diff
@@ -85,6 +85,20 @@ test-heavy: build $(SAMPLEDATA)
 test-pcapingest: bin/$(ZEEKPATH)
 	@ZEEK="$(CURDIR)/bin/$(ZEEKPATH)/zeekrunner" go test -v -run=PcapPost -tags=pcapingest ./ppl/zqd
 
+.PHONY: test-postgres
+test-postgres: build 
+	@echo "pg test: $(PG_TEST)"
+	@ZTEST_PATH="$(CURDIR)/dist:$(CURDIR)/bin" \
+		ZTEST_TAG=postgres \
+		ZTEST_ENV="MIGRATIONS_DIR=$(CURDIR)/ppl/zqd/migrations PG_TEST=$(PG_TEST)" \
```
alfred-landrum, line 93: I'm not sure about stuffing the test environment settings through this var, but don't have an immediate alternative to suggest. I guess I'm unsure that we really need to have the ztests run with a completely clear environment; we could just alter their `PATH` to ensure they pick up the test binaries first.

ppl/zqd/apiserver/postgresdb.go
mattnibs: Makes sense

ppl/zqd/apiserver/postgresdb.go
mattnibs: Good idea.

Makefile
mattnibs: FWIW I wasn't a huge fan of doing this either. I think I'm for setting the testenv to the host env- I do worry about carrying over $PATH. I can imagine a situation where for some reasons a build on a binary stops working (or gets placed somewhere else) and some test keeps failing because its picking up a binary that exists in your path, while you are not sure why your code changes are having no effect.

ppl/zqd/postgres/config.go
```diff
@@ -0,0 +1,115 @@
...
+	// Default is tcp.
+	Network string `json:"network,omitempty"`
+	// TCP host:port or Unix socket depending on Network.
+	Addr string `json:"addr"`
+
+	User      string `json:"user"`
+	Password  string `json:"password"`
+	Database  string `json:"database"`
+	TLSConfig *tls.Config
+}
```
nwt, line 23: Could this just be `type Config struct { pg.Options }`?

mattnibs: That's just crazy enough to work!

marktwallace:
I have a generic question about the advantages of a an ORM like go-pg vs. using something lighter weight. Was there a performance reason to use go-pg?

I have really enjoyed the flexibility of using go database/sql with sqlx for mapping structs to query results, the same way go does json.

It looks like go-pg generates SQL from go code, which is simple now, for this initial code, where you are not yet having to use the struct tags from go-pg. Later, after the database has evolved some complexity, I would worry that the go-pg struct tags to override default behavior will get more complex than mapping structs with sqlx.

If you have experience with go-pg and like it, then it is fine with me. At first look, I do not see where it simplifies things. You are still writing sql, but mapping the results of queries seems more indirect than database/sql and sqlx.

mattnibs:
@marktwallace my choice of go-pg was just speed of development time. I'm definitely not worried about performance at this stage in the game.

Makefile
mattnibs: > That said, I'm uneasy about using MIGRATIONS_DIR to access files outside a ztest's temporary directory. That doesn't feel right.

Do you have a solution for this? Perhaps inputs could have a symlink type that would symlink a directory- I don't think you would want to copy the entirety of the migrations dir to a test dir, especially as they get bigger.

mattnibs: I like the env idea, I'll add that.

mattnibs: Took a little work but I've officially killed all the ZTEST_ENV env variable!

.github/workflows/ci.yaml
```diff
@@ -44,4 +44,23 @@ jobs:
...
+      postgres:
+        image: postgres:11
+        env:
+          POSTGRES_PASSWORD: test
+          POSTGRES_USER: test
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
```
nwt, line 60:
```suggestion
          --health-retries 5
          --health-timeout 5s
```

ppl/cmd/zqd/listen/command.go
```diff
@@ -60,6 +62,8 @@ type Command struct {
 	listenAddr          string
 	logLevel            zapcore.Level
 	portFile            string
+	postgres            postgres.Config
```
nwt, line 65: Nit: Name this postgresConf for so it's like loggerConf.

ppl/zqd/apiserver/db.go
```diff
@@ -0,0 +1,40 @@
...
+
+const (
+	DBUnspecified DBKind = ""
+	DBFile        DBKind = "file"
+	DBPostgres    DBKind = "postgres"
+)
+
+type DBConfig struct {
+	Kind     DBKind          `json:"kind"`
+	Postgres postgres.Config `json:"postgres"`
```
nwt, line 30: Is this ever marshaled? If not, no need for the struct tags.

ppl/zqd/postgres/errors.go
```diff
@@ -0,0 +1,15 @@
+package postgres
+
+import "github.com/go-pg/pg/v10"
+
+func ErrIsUniqueViolation(err error) bool {
+	// Would normally use the errors.As facilities but the does not work with
+	// certain queries in pkg pg.
```
nwt, line 7: This comment applies to both functions, so maybe move it to file level?

ppl/zqd/postgres/errors.go
```diff
@@ -0,0 +1,15 @@
...
+	pgerr, ok := err.(pg.Error)
+	return ok && pgerr.Field('C') == "23505"
+}
+
+func ErrIsForeignKeyViolation(err error) bool {
```
nwt, line 12: Maybe call these IsUniqueViolation and IsForeignKeyViolation, like os.IsNotExist et al.?

ztest/shell.go
```diff
@@ -2,19 +2,28 @@ package ztest
...
 		cmd = exec.Command("cmd.exe", "/c", script)
 	} else {
 		cmd = exec.Command("bash", "-c", script)
 	}
-	cmd.Env = []string{"PATH=/bin:/usr/bin:" + bindir}
+
+	for _, env := range useenvs {
+		if v := os.Getenv(env); v != "" {
+			cmd.Env = append(cmd.Env, fmt.Sprintf("%s=%s", env, v))
+		}
```
nwt, line 23: Don't drop variables with empty values.
```suggestion
		if v, ok := os.LookupEnv(env); ok {
			cmd.Env = append(cmd.Env, fmt.Sprintf("%s=%s", env, v))
		}
```

scripts/zqd-services/docker-compose.yaml
```diff
@@ -0,0 +1,10 @@
+# docker-compose.yaml
```
nwt, line 1: I think this file belongs in ppl.

ppl/zqd/apiserver/db.go
mattnibs: I suppose not. Will remove.

ztest/ztest.go
```diff
@@ -198,7 +198,8 @@ type File struct {
 	Source string  `yaml:"source,omitempty"`
 	// Re is a regular expression describing the contents of the file,
 	// which is only applicable to output files.
-	Re string `yaml:"regexp,omitempty"`
+	Re      string `yaml:"regexp,omitempty"`
+	Symlink string `yaml:"symlink,omitempty"`
```
nwt, line 202: Add comment documenting this thing.

Makefile
```diff
@@ -85,6 +85,18 @@ test-heavy: build $(SAMPLEDATA)
...
+test-postgres: build 
+	@ZTEST_PATH="$(CURDIR)/dist:$(CURDIR)/bin" \
+		ZTEST_TAG=postgres \
+		go test -v -run TestZq/ppl/zqd/postgres/ztests .
+
+.PHONY: test-postgres-docker
+test-postgres-docker:
+	@docker-compose -f $(CURDIR)/ppl/zqd/scripts/dkc-services.yaml up -d
+	@-$(MAKE) test-postgres PG_TEST="postgres://test:test@localhost:5432/test?sslmode=disable"
+	@docker-compose -f $(CURDIR)/ppl/zqd/scripts/dkc-services.yaml down
```
nwt, line 98: Using `-` means this rule can succeed even if the tests fail. Gross as it is, I think this should be something that will fail if either command fails, like this:
```suggestion
	 $(MAKE) test-postgres; \
		status=$?; \
		docker-compose -f $(CURDIR)/ppl/zqd/scripts/dkc-services.yaml down || exit; \
		exit $status
```

ppl/zqd/postgres/config.go
```diff
@@ -0,0 +1,50 @@
...
+
+// postgresql://[user[:password]@][netloc][:port]
+
+func (c Config) String() string {
+	if c.IsEmpty() {
+		return ""
+	}
+	var str string
+	str += "postgres://"
```
nwt, line 32:
```suggestion
	str := "postgres://"
```

ppl/zqd/postgres/config.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+	str += c.Addr
+	if c.Database != "" {
+		str += "/" + c.Database
+	}
+	params := url.Values{}
+	if c.TLSConfig == nil {
+		params.Add("sslmode", "disable")
+		str += "?" + params.Encode()
+	}
```
nwt, line 48:
```suggestion
	if c.TLSConfig == nil {
		params := url.Values{"sslmode": {"disable"}}
		str += "?" + params.Encode()
	}
```

.github/workflows/ci.yaml
```diff
@@ -44,4 +44,23 @@ jobs:
...
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-retries 5
+          --health-timeout 5s
+    steps:
+      - uses: actions/checkout@v1
+      - run: make test-postgres
+        env:
+          PG_TEST: 'postgres://test:test@postgres:5432/test?sslmode=disable'
```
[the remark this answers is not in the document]
mattnibs: Happy to do this on the github side, but my instincts want to keep PG_TEST around: In case I don't want tests interfering with a local postgres instance that I already have running. Django follows a similar pattern.

alfred-landrum (commented):
I also add some discussion with @mattnibs on slack about suggestions on changing the config style a bit:

`zqd -db.kind=file` , or
`zqd -db.kind=postgres -db.postgres.url=...` parse the postgres url
`zqd -db.kind=postgres -db.postgres.address=.. -db.postgres.db=... ...` or allow specifying (or overriding) parts of config, like the dbname

To handle this, modify the  type DBConfig struct  and add a SetFlags method on it, similar to AuthConfig.SetFlags or WorkerConfig.SetFlags.

mattnibs:
@alfred-landrum Added db flags you suggested.

alfred-landrum (approved):

PR #1826, opened 2020-12-16
Simplified Recruiter API

marktwallace:
This is a substantial change to the recruiter design from last week. Read the new design doc: ppl/zqd/recruiter/README.md to understand the select's and for loops in ppl/zqd/recruiter/workerreg.go.

https://github.com/brimsec/zq/blob/recruiter-alt/ppl/zqd/recruiter/README.md

The following google doc describes what changed in the design:
https://docs.google.com/document/d/1Y1mM7bpkBnTvwnMPFNdACNu9ELAPInYtFQhBf1d-C1Y/edit?usp=sharing


ppl/zqd/recruiter.go
```diff
@@ -32,6 +23,8 @@ func handleRecruit(c *Core, w http.ResponseWriter, r *http.Request) {
 	}
 	workers := make([]api.Worker, len(ws))
 	for i, e := range ws {
+		// Note the receiver for this channel is the select in handleRegister below.
```
alfred-landrum, line 26: There's an issue here: the handleRegister timeout could fire just before a worker is selected for recruitment, which would leave this request stuck trying to write to a channel that has no listeners.

driver/parallel.go
```diff
@@ -160,12 +158,10 @@ func (pg *parallelGroup) nextSourceForConn(conn *client.Connection) (ScannerClos
 			zbuf.Scanner
 			io.Closer
 		}{s, rc}
-
 		pg.mu.Lock()
 		pg.scanners[sc] = struct{}{}
 		pg.mu.Unlock()
 		return sc, nil
-
```
mattnibs, line 168: I would just assume leaving these lines as is considering you're not really doing work over here and line spacing is a matter of preference.

marktwallace: I put the spaces in there to begin with, and I regretted it, because it was inconsistent with the style of the rest of the code in this source file.

ppl/zqd/recruiter.go
marktwallace: This is a really good point. I cannot do a blocking write here because it could cause the call to /recruiter/recruit to hang. In my latest commit, I converted this to a non-blocking write to the "recruited" receiver. My intent here is in the event that the /recruit and the registration timeout are nearly simultaneous, I want the registration timeout to win, and the recruit to return fewer workers. I added logic to do that by having a callback in the recruiter.WorkerDetail struct, and using that callback to do a non-blocking write to the "recruited" receiver. In addition, I "Stop()" the timer in the callback to narrow the window of this race condition. (I read the code for timer.Stop(), and it looks like it is safe to call from another thread.)

driver/compile.go
```diff
@@ -6,6 +6,8 @@ import (
 	"runtime"
 	"time"
 
+	"github.com/brimsec/zq/ppl/zqd/recruiter"
+
 	"github.com/brimsec/zq/ast"
```
nwt, line 11: This shouldn't be in a separate section.

ppl/zqd/recruiter/recruitworker.go
```diff
@@ -27,36 +30,34 @@ func RecruitWorkers(ctx *proc.Context, workerCount int, recruiter string, worker
...
-
 	var workers []string
 	for _, w := range resp.Workers {
 		workers = append(workers, w.Addr)
 	}
 	return workers, nil
 }
 
+// RecruitWorkers is used by the zqd root process to release workers after a distributed query.
 func ReleaseWorker(ctx context.Context, conn *client.Connection, logger *zap.Logger) error {
```
nwt, line 61: Fix or remove comment.

ppl/zqd/recruiter/recruitworker.go
```diff
@@ -12,10 +12,13 @@ import (
 	"go.uber.org/zap"
 )
 
-func RecruitWorkers(ctx *proc.Context, workerCount int, recruiter string, workerstr string) ([]string, error) {
-	if workerstr != "" {
-		// Special case: workerstr is used for ZTests
-		workers := strings.Split(workerstr, ",")
+// RecruitWorkers is used by the zqd root process to recruit workers for a distributed query.
+func RecruitWorkers(ctx *proc.Context, workerCount int, conf WorkerConfig, logger *zap.Logger) ([]string, error) {
+	logger.Info("recruit workers", zap.Int("count", workerCount))
```
nwt, line 17: Log strings should start with a capital letter.

ztests/suite/zqd/awaitfile.sh
```diff
@@ -0,0 +1,22 @@
...
+    fi
+    sleep 1
+  done
+}
+
+zqdroot=zqdroot
+mkdir -p $zqdroot
+
+mkdir -p s3/bucket
+portdir=$(mktemp -d)
```
nwt, line 22: These don't seem like they belong in awaitfile.sh. How giving this file a better name or finding a better home for them?

ppl/zqd/recruiter/recruitworker.go
marktwallace: Oops, I got the convention for error messages and log messages mixed up. I will capitalize all the log messages.

ppl/zqd/recruiter.go
```diff
@@ -44,32 +37,68 @@ func handleRegister(c *Core, w http.ResponseWriter, r *http.Request) {
...
 	}
-	respond(c, w, r, http.StatusOK, api.RegisterResponse{
-		Registered: registered,
-	})
-}
-
-func handleUnreserve(c *Core, w http.ResponseWriter, r *http.Request) {
-	var req api.UnreserveRequest
-	if !request(c, w, r, &req) {
+	timer := time.NewTimer(time.Duration(req.Timeout) * time.Millisecond)
```
alfred-landrum, line 44: The changes here are an improvement to address previous comments about recruitment & idle timeout. I think the code would be less complex if it was handled inside the Recruit call and what is now the 'Register' call. If you combine the Register/Deregister calls into a single call, like 'recruiter.WaitForRecruitment(ctx, req.Addr, req.NodeName, req.Timeout)'.

ppl/zqd/recruiter/workerreg.go
```diff
@@ -2,64 +2,170 @@ package recruiter
 
 import (
```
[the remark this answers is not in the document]
marktwallace: Yes, good suggestion.

ppl/zqd/recruiter.go
marktwallace: I like that idea. I will take a look at how to do it.

ppl/zqd/recruiter/workerreg.go
```diff
@@ -2,64 +2,170 @@ package recruiter
...
+			if retryWait < w.conf.MaxRetry {
+				retryWait = (retryWait * 3) / 2
+				// Note: doubling is too fast a backoff for this, so using 1.5 x.
+			} else {
+				retryWait = w.conf.MaxRetry
+			}
+			continue
+		}
+		retryWait = w.conf.MinRetry // Retry goes back to min after a success.
+		if resp.Directive == "reserved" {
```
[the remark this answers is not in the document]
marktwallace: Thanks!

ppl/zqd/recruiter.go
marktwallace: I removed a redundant call to workerPool.Deregister which made this code confusing. An important aspect of the design is that workerPool.Recruit automatically deregisters recruited workers. The explicit Deregister call is only needed in the case of a request timeout or context cancel. So, workerPool.Deregister does not always follow workerPool.Register, which was the way the previous code made it look.

The code in the workerPool methods is as minimal as possible to avoid holding the mutex lock longer than needed. I can introduce an intermediate layer between the request handlers in the zqd package and the workerPool methods if you think that will make the code more readable.

alfred-landrum: yes please; I think an intermediate layer would be helpful to separate the api handling from the register goroutine wakeup details.

marktwallace: I moved this code into recruiter/longpoll.go. It looks better to me; the code works the same, but less cluttered.
I like having the channel as a local variable, so it is easy to find the readers and writers.

driver/parallel.go
nwt: I'm with @mattnibs on this one. The substantive changes here are already pretty large, and adding cosmetic changes on top that increases the burden on reviewers. I always appreciate a smaller diff.

ztests/suite/zqd/recruiter-curl.yaml
```diff
@@ -1,32 +1,29 @@
...
+  source common.sh
   source recruiter-curl.sh
-  curl -X POST -d '{"addr":"a.b.c:5000","node_name":"a.b"}' \
-    http://$ZQD_HOST/recruiter/register 2> err
+  curl -X POST -d '{"addr":"a.b.c:5000","node_name":"a.b","timeout":1}' \
+     http://$ZQD_HOST/recruiter/register 2> err
+  curl -X POST -d '{"addr":"a.b.c:5000","node_name":"a.b","timeout":30000}' \
+    http://$ZQD_HOST/recruiter/register 2> err &
+  # This is intentionally backgrounded. The 100ms sleep allows the registration.
+  sleep 0.1
```
alfred-landrum, line 9: Since our CI machines tend to be slow, can bump this up to at least a second? (It runs fine for me locally)

alfred-landrum:
FYI, I pulled this branch & ran some local manual tests with a large number of workers, a recruiter, and a 'default'/root personality. I manually ctrl-c'ed a few workers & the recruiter, and verified that the remaining components continued to work, and I could add & remove workers, yay!

ppl/zqd/worker/registration.go
```diff
@@ -0,0 +1,172 @@
...
+		logger:      logger,
+		releaseChan: make(chan string),
+		selfaddr:    net.JoinHostPort(host, port),
+	}
+	return rs, nil
+}
+
+// RegisterWithRecruiter is used by personality=worker.
+func (rs *RegistrationState) RegisterWithRecruiter() {
+	ctx := context.Background()
```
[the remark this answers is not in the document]
marktwallace: Yep, added that.

ztests/suite/zqd/recruiter-curl.yaml
marktwallace: Yes, I think I saw recruiter-curl fail once on GitHub. Must need a little more time there.

ppl/zqd/worker/registration.go
```diff
@@ -0,0 +1,172 @@
...
+	"net"
+	"os"
+	"time"
+
+	"github.com/brimsec/zq/api"
+	"github.com/brimsec/zq/api/client"
+	"go.uber.org/zap"
+)
+
+type WorkerConfig struct {
```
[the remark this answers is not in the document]
marktwallace: done

marktwallace:
I'm glad it worked well in your path. So far, I have not been able to crash the cluster, either. I've been trying to think of ways to stress it. One thing we should look at is how a zqd root process recovers from an interrupted /worker/chunksearch. I'd like to set up a test for that, but the timing is tricky.

ppl/zqd/recruiter/recruitworker.go
```diff
@@ -27,29 +33,26 @@ func RecruitWorkers(ctx *proc.Context, workerCount int, recruiter string, worker
...
+		return nil, fmt.Errorf("error on recruit for recruiter at %s : %v", conf.Recruiter, err)
 	}
 	if workerCount > len(resp.Workers) {
-		// TODO: we should fail back to running the query with fewer workers if possible.
-		// Determining when that is possible is non-trivial. One issue is that the
-		// parallel procs have already been compiled into the flowgraph by the time we get here.
-		// An alternative is to wait and try to recruit more workers,
-		// which would reserve the idle zqd root process while waiting. -MTW
-		return nil, fmt.Errorf("distributed exec failure: requested workers %d greater than available workers %d",
+		err := fmt.Errorf("requested workers %d greater than available workers %d",
```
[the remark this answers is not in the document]
marktwallace: dons

alfred-landrum (approved):

PR #1828, opened 2020-12-16
Fix some CSV output problems

### LINKED ISSUE #1326: Restore text representation before printing as CSV

nwt:
* Print "\" as itself rather than "\\".
* Print "-" as itself rather than "\u002d".
* Print an unset value as "" rather than "-".

Fixes #1326.


zio/csvio/ztests/bstring.yaml
```diff
@@ -12,12 +12,5 @@ input: |
...
-  bstring
-  -
-  
-  """"
-  ","
-  \x2d
-  \\
-  \x00\xff
+output: !!binary |
+  YnN0cmluZwoKCiIiIiIKIiwiCi0KXAoA/wo=
```
marktwallace, line 16: The old output has a logical correspondence to the input, but the new output just checks that the encoding has not changed from build to build. How do we verify the !!binary string included in the output still has the same semantics if for some reason it changes?

marktwallace (approved):
This looks reasonable and the ZTests all work for me, too. It looks like the main impact is that the csv output of 0:[-;] is now an empty string rather than "-".  This PR does not change output for float64, but the fixed-point output for large exponents looks very surprising.

zio/csvio/ztests/bstring.yaml
nwt: @marktwallace: The logical correspondence remains, though to see it in the YAML, you do unfortunately have to decode the Base64 string. But a test failure will print a diff against the decoded bytes, so there's that.

However, absent a ztest feature that interprets escape sequences in `output`, `!!binary` seems like the best way to include arbitrary bytes in it.

Issue #1801, opened 2020-12-12
zng.Record type field should be zng.Type


mccanne:
Currently, zng.Record's must be of type zng.TypeRecord.  This makes sense obviously but means that top-level record types can not have aliased types.  This means this fine example from the ZSON spec is currently not possible:
```
{ city: "Berkeley", state: "CA", population: 121643 (uint32) } (=city_schema)
{ city: "Broad Cove", state: "ME", population: 806 } (city_schema)
{ city: "Baton Rouge", state: "LA", population: 221599 } (city_schema)
```
As I have explained the concepts to people with real-time demos, I realize the lack of this capability is definitely an awkward limitation.

This change could be done wrt to a plan for issue #1762.  As a first step, in thinking ahead, we could change 
```
type Record struct {
	Type        *TypeRecord
	nonvolatile bool
	Raw     zcode.Bytes
	ts      nano.Ts
	tsValid bool
}
```
to use an embedded/exported zng.Value...
```
type Record struct {
        zng.Value
	nonvolatile bool
	ts      nano.Ts
	tsValid bool
}
```
given record fields of `Type` and `Bytes` instead of `TypeRecord` and `Raw` (raw was never the right word anyway).

Then we could add 
```
func (r *Record) Type() *TypeRecord {
        return AliasedType(r.Type).(*TypeRecord)
}
```
and change all the refs of `rec.Type` to `rec.TypeRecord()`.  It's a lot of typing but no real changes and moves us closer to issue #1762 (and to a generic model of processing zng.Values everywhere instead of zng.Record as @henridf has suggewted).

PR #1829, opened 2020-12-16
allow outer records to be aliased

mccanne:
This commit is a first step toward #1801, which is to embed
zng.Value into zng.Record so that a record can take on any
alias type.  This step toward this goal adds an extra zng.Type
to record called Alias, which is either a zng.TypeRecord
(behavior as before) or a zng.TypeAlias (new behavior).

This allows the example ZSONs in the ZSON spec, which have
top-level type defs, to actually work.  Tests were adding
to cover these examples and make sure the typeof(.) operator
worked on the aliased outer records.

### DIFF
--- a/zng/recordval.go
+++ b/zng/recordval.go
@@ -44,6 +44,7 @@ func (r *RecordTypeError) Unwrap() error { return r.Err }
 // zng.Records by accessing data via the Record methods.
 type Record struct {
 	Type        *TypeRecord
+	Alias       Type
 	nonvolatile bool
 	// Raw is the serialization format for records.  A raw value comprises a
 	// sequence of zvals, one per descriptor column.  The descriptor is stored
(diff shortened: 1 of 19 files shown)


zng/recordval.go
```diff
@@ -56,6 +57,16 @@ type Record struct {
...
+	}
+}
+
+func NewRecordFromType(typ Type, raw zcode.Bytes) *Record {
+	return &Record{
+		Type:        AliasedType(typ).(*TypeRecord),
+		Alias:       typ,
 		nonvolatile: true,
 		Raw:         raw,
 	}
```
nwt, line 72: Instead of adding NewRecordFromType et al., maybe just change the `typ *TypeRecord` to `typ Type`?
```
func NewRecord(typ Type, raw zcode.Bytes) *Record {
	return &Record{
		Type:        AliasedType(typ).(*TypeRecord),
		Alias:       typ,
		nonvolatile: true,
		Raw:         raw,
	}
}

File: zng/resolver/slice.go [hunk 2]
```diff
@@ -6,18 +6,18 @@ import (
 
 // Slice is a table of descriptors respresented as a slice and grown
 // on demand as small-in type descriptors are entered into the table.
```
nwt, line 8: s/in/int/

nwt (approved):

zng/recordval.go
mccanne: I did this on purpose to document the different call sites with different semantics for when we refactor in the next step and don't want to miss the variations in use cases.  I was presuming all the NewRecord* functions will be cleaned up and normalized at that point.

PR #1830, opened 2020-12-16
zson lexer: fix cursor length check

henridf:
I noticed this while parsing a hand-written type value, which was was returning `nil, io.EOF` until i added a trailing newline.

### DIFF
--- a/zson/lexer.go
+++ b/zson/lexer.go
@@ -62,14 +62,14 @@ func (l *Lexer) match(b byte) (bool, error) {
 }
 
 func (l *Lexer) matchTight(b byte) (bool, error) {
-	if len(l.cursor) > 1 {
-		ok := b == l.cursor[0]
-		if ok {
-			l.skip(1)
-		}
-		return ok, nil
+	if len(l.cursor) == 0 {
+		return false, io.EOF
 	}
-	return false, io.EOF
+	ok := b == l.cursor[0]
+	if ok {
+		l.skip(1)
+	}
+	return ok, nil
 }
 
 func (l *Lexer) matchBytes(b []byte) (bool, error) {
--- a/zson/lexer.go
+++ b/zson/lexer.go
@@ -65,11 +65,11 @@ func (l *Lexer) matchTight(b byte) (bool, error) {
 	if len(l.cursor) == 0 {
 		return false, io.EOF
 	}
-	ok := b == l.cursor[0]
-	if ok {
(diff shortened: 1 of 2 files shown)


zson/ztests/no-trailing-newline.yaml
```diff
@@ -0,0 +1,18 @@
+script: |
+  zq -t -i zson "count()" in.zson
```
nwt, line 2: Does this need to be a script test?

zson/lexer.go
```diff
@@ -62,14 +62,14 @@ func (l *Lexer) match(b byte) (bool, error) {
...
-		return ok, nil
+	if len(l.cursor) == 0 {
+		return false, io.EOF
 	}
-	return false, io.EOF
+	ok := b == l.cursor[0]
+	if ok {
+		l.skip(1)
+	}
+	return ok, nil
```
nwt, line 72: Might be clearer this way.
```suggestion
	if b == l.cursor[0] {
		l.skip(1)
		return true, nil
	}
	return false, nil
```

nwt (approved):

zson/ztests/no-trailing-newline.yaml
nwt: Just realized it does because detector doesn't understand ZSON.

mccanne:
I don't think we should get rid of the io.EOF at the lower layer.  It's up to the parsing layers to decide if EOF is ok or not in the middle of parsing something.

henridf:
@mccanne I'm surely missing something. I understand that it's up to the parsing layer to decide if EOF is ok, but I'm not able to see how this change affects that. 

One way to look at the change is to compare 
`(*Lexer).matchTight()`   with
`(*Lexer).matchBytesTight()`.

Before this change, they behave differently: `matchTight(')')` requires at least two bytes remaining in the cursor, while `matchBytesTight([]byte{')'})` requires only one byte remaining. 

In other words, `matchTight` requires that there be at least one byte left _after_ the match, while `matchBytesTight` is ok if it matches all the way to the end.

With this change, `matchTight` behaves like `matchBytesTight`, and no longer requires additional bytes after the match. 

(I'd guess that at the least, if this fix is incorrect, the proper fix should also therefore include making matchBytesTight behave like matchTight, e.g. require at least one byte after the match).

mccanne:
My bad

PR #1832, opened 2020-12-17
Fill in (t *TypeTable).LookupType

henridf:
This method can't yet be exercised via the contact surface (though that should change soon with #1785 and #1619).

So I couldn't test it with a ztest. Instead, I added a temporary Go test in aab7b99, and then reverted it after it passed CI.


mccanne (approved):
Cool.  We might want a zson.Compiler bundle that can Compile() into zng and CompileValue() to an zson.Value, but that can come later.  This would prevent the compilations of the lexer regexp on every type table miss but that's not important since the misses are rarea.

zson/table.go
```diff
@@ -41,7 +46,24 @@ func (t *TypeTable) LookupType(zson string) (zng.Type, error) {
...
+			return nil, err
+		}
+		a := NewAnalyzer()
+		val, err := a.ConvertValue(t.zctx, ast)
+		if err != nil {
+			return nil, err
+		}
+		tv, ok := val.(*TypeValue)
+		if !ok {
+			return nil, fmt.Errorf("(*TypeTable).LookupType() internal error: value of type %T", val)
```
nwt, line 64: Nit: No other error strings use "(*Type).Method" notation, and it'd be nice to be consistent.

henridf: I can make the change, but what's the proper form - just start with `"LookupType internal error:"`?

nwt: Based on what I see in package zson, I'd vote for something like `"internal error: value of type %T in TypeTable.LookupType()"`.
```
$ git grep -i '(errors.New|fmt.Errorf)\("[^"]*internal'
microindex/writer.go:                   return fmt.Errorf("internal microindex error: index file size (%d) does not equal zng size (%d)", n, size)
zio/zjsonio/stream.go:          return nil, errors.New("internal error: zng record body must be a container")
zio/zngio/scanner.go:                   return nil, errors.New("internal error: batch end before zngio batch started")
zson/analyzer.go:       return nil, fmt.Errorf("internal error: unknown ast type in Analyzer.convertAny(): %T", val)
zson/analyzer.go:                       return nil, fmt.Errorf("internal error built type (%q) does not match semantic type (%q)", zv.Type.ZSON(), typ.ZSON())
```

henridf: Nicer! https://github.com/brimsec/zq/pull/1834

PR #1837, opened 2020-12-17
add unmarshaling of interface values

mccanne:
This commit extends the zng marshaler with custom type bindings and
the capability of using those bindings to unmarshal ZNG/ZSON-serialized
values into Go interface values, something json.Unmarshal cannot due
as JSON lacks the necessary type system.

### DIFF
--- a/zng/resolver/marshal.go
+++ b/zng/resolver/marshal.go
@@ -142,14 +142,14 @@ const (
 	TypeFull
 )
 
-// Decorate informs the marshaler to add type decorations the resulting ZNG
-// in the form of type aliases that are named int the sytle indicated, e.g.,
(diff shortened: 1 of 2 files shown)


zng/resolver/marshal.go
```diff
@@ -104,6 +122,98 @@ func fieldName(f reflect.StructField) string {
...
+func typeFull(name, path string) string {
+	return fmt.Sprintf("%s.%s", path, name)
+}
+
+const (
+	TypeNone = iota
+	TypeSimple
+	TypePackage
+	TypeFull
+)
```
nwt, line 143: I think there should be a type for these like `type Style int` or `type TypeStyle int`. Also, they read like types because of the Type prefix, so maybe change that to Style or TypeStyle.

nwt (approved):

PR #1838, opened 2020-12-18
Remove links to Developer Docs for now

philrz:
In the time since we first put up the ZQL docs, we've included "Developer Docs" links with each processor and aggregate function to a pkg.go.dev destination that provided a reasonable jumping-in point in case someone wanted to start looking at its implementation. However, with the release of zq v0.25.0, the code is now structured so `cut` & `pick` (though not `drop` yet) can be used as both a function as well as an processor. This doesn't lend itself currently to being revealed through simple hyperlinks, so after speaking to @mccanne and @nwt a bit about it, it seems like we'd be ok just living without these links for now. Perhaps at some point in the future we can invest some cycles specifically toward making the project more approachable to new contributors, such as by investing in the sorts of things we did for Brim like the "code base walkthrough" we made available in both [doc](https://github.com/brimsec/brim/wiki/Code-Base-Walkthrough) and [video](https://www.youtube.com/watch?v=CPel0iu1pig) form.


mccanne (approved):
Code walkthrough video is a great idea.

PR #1840, opened 2020-12-18
create zson.Formatter from most of zsonio.Writer

### LINKED ISSUE #1763: zson writer needs formatter
We need to refactor the zson writer to use an encoder/formatter.  Comment from `zio/zngio/writer.go`:

```
// XXX this needs a refactor.  The type formatter needs to be be able to format
// a value for enum but the Writer type presume values are "written" not "formatter".
// This hack works with a bytes.Buffer works around this.  Also, when we are writing
// enum values inside of a type, we don't want the pretty printing so the refactor
// will allow you to specify indent on a per-format basis.
```

mccanne:
This commit creates zson.Formatter from the logic that used to
live in zsonio.Writer.  With this refactor, a coming-soon
zson.Marshaler can use the formatter without creating an
import cycles.

This is mostly code movement and rename.  The one logic change
is the formatter now uses a stack of strings.Builder instead of
saving the writer pointer in the stack when re-entering.  Also,
instead of writing the formatted output to an io.Writer, the
output is built in memory using a strings.Bulder.

Closes #1763


zson/formatter.go
```diff
@@ -0,0 +1,401 @@
...
+	typeTab     int
+	typeNewline string
+	nid         int
+	types       *TypeTable
+	builder     strings.Builder
+	stack       []strings.Builder
+}
+
+func NewFormatter(pretty int) *Formatter {
+	newline := ""
```
nwt, line 27:
```suggestion
	var newline string
```

zson/formatter.go
```diff
@@ -0,0 +1,401 @@
...
+	return empty, nil
+}
+
+func (f *Formatter) indent(tab int, s string) {
+	n := len(f.whitespace)
+	if n < tab {
+		n = 2 * tab
+		f.whitespace = strings.Repeat(" ", n)
+	}
+	f.build(f.whitespace[0:tab])
```
nwt, line 252:
```suggestion
	for k := 0; k < tab; k++ {
		f.builder.WriteByte(' ')
	}
```

zson/formatter.go
```diff
@@ -0,0 +1,401 @@
...
+		f.build(zng.QuotedName(col.Name))
+		f.build(":")
+		f.formatType(col.Type)
+		sep = ","
+	}
+	f.build("}")
+}
+
+func (f *Formatter) formatTypeUnion(typ *zng.TypeUnion) {
+	sep := ""
```
nwt, line 356:
```suggestion
	var sep string
```

zson/formatter.go
```diff
@@ -0,0 +1,401 @@
...
+	for _, typ := range typ.Types {
+		f.build(sep)
+		f.formatType(typ)
+		sep = ","
+	}
+	f.build(")")
+}
+
+func (f *Formatter) formatTypeEnum(typ *zng.TypeEnum) error {
+	sep := ""
```
nwt, line 367:
```suggestion
	var sep string
```

zio/zsonio/writer.go
```diff
@@ -54,362 +28,13 @@ func (w *Writer) Close() error {
...
-	if err := w.writeValueAndDecorate(typ, rec.Raw); err != nil {
-		return err
-	}
-	return w.write("\n")
-}
-
-func (w *Writer) writeValueAndDecorate(typ zng.Type, bytes zcode.Bytes) error {
-	known := w.typedefs.exists(typ)
-	if err := w.writeValue(0, typ, bytes, known, false); err != nil {
+	if _, err := w.writer.Write([]byte(s)); err != nil {
```
nwt, line 35: Avoid copying s if w.writer implements io.StringWriter.
```suggestion
	if _, err := io.WriteString(w.writer.Write, s); err != nil {
```

zson/formatter.go
```diff
@@ -0,0 +1,401 @@
...
+		typedefs:   make(typemap),
+		tab:        pretty,
+		newline:    newline,
+		whitespace: strings.Repeat(" ", 80),
+	}
+}
+
+func (f *Formatter) push() {
+	f.stack = append(f.stack, f.builder)
+	f.builder = strings.Builder{}
```
nwt, line 41:
```suggestion
	f.builder.Reset()
```

zson/formatter.go
```diff
@@ -0,0 +1,401 @@
...
+	}
+	var err error
+	var null bool
+	switch t := typ.(type) {
+	default:
+		f.build(typ.ZSONOf(bytes))
+	case *zng.TypeAlias:
+		if err := f.formatValue(indent, t.Type, bytes, known, false); err != nil {
+			return err
+		}
```
nwt, line 82: No need to check err here.
```suggestion
		err = f.formatValue(indent, t.Type, bytes, known, false)
```

nwt (approved):

zson/formatter.go
mccanne: I actually first wrote the code this way then wanted to make sure it works, and while it does work, the docs for strings.Builder say to not make a copy without further elaboration.  I take this to mean that you definitely shouldn't write into the copy but Reset happens to work because it just zeros out the pointers, which is the same as what I'm doing here but this way conforms with the documentation.

PR #1842, opened 2020-12-18
json types: add support for inferring type of "extra" fields

henridf:
A new boolean field `inferextra`  is added to the json type config. If true, the json typed reader no longer returns an error upon encountering fields not present in the current descriptor, but instead applies plain old json type inference and adds them to the end of the record. If false, the reader follows the prior behavior of returning an error.

Fields are added to the end of the record, at the top level, with a "flattened" field name. For example, assuming the incoming json object has a top-level field `uid2` and a nested field `id2` that are not in the schema, such as:

```
...
{
...
"uid2": "asdfasdfasdf",
"id2": {"orig_h": "2.2.2.1"}
}
```
then those fields will show up as "uid" and "id2.orig_h", both at the record top-level. On the one hand, this is is not very ergonomic, but on the other, it is necessary to allow "pushing" the extra fields to the right. And perhaps more importantly, leaving the inferred fields "in their place", would require a substantially bigger change to the typed parser, which does not seem worthwhile given its expected TTL.

First part of #1841

PR #1847, opened 2020-12-19
zqd: send warning upon unknown ndjson fields in pcap ingest

### LINKED ISSUE #1841: Keep unexpected fields in Suricata EVE records
As of the release of the GA combo of zq v0.25.0 and Brim v0.21.0, the JSON typing config for Suricata alert records is rigid such that we only accept the fields we expect to receive from the embedded Suricata. However, soon after release, a Brim community user reported a problem that (based on the error message) appears to be caused by one or more additional fields having appeared in the alert records that were not defined in the typing config:

image

The proposal is to relax the rigidity such that for unknown fields, we will do the best we can to convert these to appropriate ZNG types. If records with such additional fields should appear, this would prevent them from being automatically shown in columns in Brim after doing ZQL such as `event_type=alert` (*). But consensus is that this is an improvement over blocking the read altogether until we add knowledge of the field to the typing config.

These additional fields are not expected to be common. We did significant research in advance before creating what we believed to be an adequately comprehensive schema, and we're still waiting to hear from the user about this specific field so we can track down why we missed it. Therefore, it would seem ideal if the implementation were able to warn the user of the unexpected occurrence and nudge them toward telling us about it. My proposal would be a warning like

> Unexpected additional field 'foobar' found in Suricata alert. Please email this error message to support@brimsecurity.com.

---

(*) - Adding `| fuse` would address this, though.

henridf:
closes #1841


henridf:
I verified that the warnings show up in the Brim application.

mccanne (approved):
This looks good.  Attaching the warnings to something called "types" sounded odd to me but when I looked I realized the TypeConfig isn't a configuration of types but rather is the shaper verb that transforms the JSON and thus understandably would generate the errors.  I would suggest a name change here but this is all temporary and will be replaced by your shaper work, correct?

philrz:
I took it for a spin in Brim as well. For others who might appreciate the visual, the pcap warnings don't currently "pop up", but are first communicated through a somewhat subtle icon in the lower-right of the screen.

image

As a note-to-self or others that may want to repro this in the future, the approach I used was to start the pcap ingest, then `cd $TMPDIR/zqd-pcap-ingest-*` to get to where the logs were being incrementally generated during ingest, then did:

```
$ cat ~/pcap/mix.ndjson >> eve.json
```

...using the test data that included the alert with the "bonus" field from #1842.

When clicked, the pop-up looks like:

image

henridf:
> As a note-to-self or others that may want to repro this in the future

And fwiw another way is to: 
- remove a field (or more) from `ppl/zqd/ingest/suricata-types.json`, 
- run `make generate` 
- run `make install` and then copy the newly built zqd to brim/zdeps

henridf:
> Attaching the warnings to something called "types" sounded odd to me but when I looked I realized the TypeConfig isn't a configuration of types but rather is the shaper verb that transforms the JSON and thus understandably would generate the errors.

@mccanne i'm not entirely sure i understood as the warnings are emitted by the "typeparser"... but generally I agree, this code could do with some deep cleaning. I would also think it will go away early next year...

philrz (approved):
Thanks for adding the contact info to the messages, @henridf!

image

PR #1850, opened 2020-12-19
more flexibe zson type names and aliased rec bug fix

mccanne:
This commit changes the definition of zson type names to include
the characters "." and "/".  This will permit go-style etc type
names of the form "github.com/acme/foo.Bar".  It will also be
useful to embed schema versions in type names for a new style of
decentralized and hierarchical schema management that zson
enables (e.g., shopping_cart_v1.0 or employee_record.2).

It also fixes a bug that happened when casting a union type
in alias form, which caused the alias to be lost and replaced
with the underlying union type.

Finally, we found and fixed a bug where r.field failed if r
was type alias of a record type.

### DIFF
--- a/zng/name.go
+++ b/zng/name.go
@@ -1,12 +1,17 @@
 package zng
 
 import (
+	"strconv"
 	"strings"
 	"unicode"
 )
 
 func IdChar(c rune) bool {
 	return unicode.IsLetter(c) || c == '_' || c == '$'
+}
+
+func TypeChar(c rune) bool {
+	return IdChar(c) || unicode.IsDigit(c) || c == '/' || c == '.'
 }
 
 func IsIdentifier(s string) bool {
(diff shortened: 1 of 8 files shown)


zng/name.go
```diff
@@ -20,6 +25,18 @@ func IsIdentifier(s string) bool {
...
 }
 
+// IsTypeName returns true iff s is a valid zson typedef name (exclusive
+// of integer names for locally-scoped typedefs).
+func IsTypeName(s string) bool {
+	for _, c := range s {
+		if !TypeChar(c) {
+			return false
+		}
+	}
```
marktwallace, line 35: Are we actually OK with names like "....." or "///" ?

nwt (approved):

zng/name.go
mccanne: The short answer is yes, the syntax allows for odd names just like the record-key syntax does as well (which is the same as JSON's record key syntax).

PR #1854, opened 2020-12-21
add zson marshaler and clean up zng marshaler

mccanne:
This commit adds a zson marshaler, which is basically a wrapper
around zng marshal that allows one to do marshaling strictly
with ZSON text format and not worry about zng or package zng/resolver.

This work also exposed some problems with the zng marshaler regarding
interfaces so several bug fixes there included here as well as
an extension of the zng marshaler to handle empty interface for
non-struct types.  We also made things work when records have
recursively invested interface types (which are resolvable with
marshaler bindings).

### DIFF
--- a/zng/resolver/marshal.go
+++ b/zng/resolver/marshal.go
@@ -144,7 +144,7 @@ func typeFull(name, path string) string {
 type TypeStyle int
 
 const (
-	StyleNone = iota
+	StyleNone TypeStyle = iota
 	StyleSimple
 	StylePackage
 	StyleFull
@@ -200,7 +200,7 @@ func (m *MarshalContext) encodeValue(v reflect.Value) (zng.Type, error) {
 	}
 	if m.decorator != nil || m.bindings != nil {
 		// Don't create aliases for interface types as this is just
-		// one value for that interface and its the underlying concrete
+		// one value for that interface and it's the underlying concrete
 		// types that implement the interface that we want to alias.
 		if !v.IsValid() || v.Kind() == reflect.Interface {
 			return typ, nil
@@ -496,7 +496,10 @@ func (u *UnmarshalContext) decodeAny(zv zng.Value, v reflect.Value) error {
 			return err
 		}
 		if typ == nil {
-			// ZNG type null values can only by value null.
+			// If typ is nil, then the value must be of ZNG type null
+			// and ZNG type values can only have value null.  So, we
+			// set it to null of the type given for the marshaled-into
+			// value and return.
 			v.Set(reflect.Zero(v.Type()))
 			return nil
 		}
@@ -769,17 +772,16 @@ func typeNameOfValue(value interface{}) (string, error) {
 func (u *UnmarshalContext) lookupType(typ zng.Type) (reflect.Type, error) {
 	switch typ := typ.(type) {
 	case *zng.TypeAlias:
-		template := u.binder.lookup(typ.Name)
-		if template == nil {
-			// Ignore aliases for which there are no bindings.
-			// If an interface type being marshaled into doesn't
-			// have a binding, then a type mismatch will be caught
-			// by reflect when the Set() method is called on the
-			// value and the concrete value doesn't implement the
-			// interface.
-			return u.lookupType(typ.Type)
+		if template := u.binder.lookup(typ.Name); template != nil {
+			return template, nil
 		}
-		return template, nil
+		// Ignore aliases for which there are no bindings.
+		// If an interface type being marshaled into doesn't
+		// have a binding, then a type mismatch will be caught
+		// by reflect when the Set() method is called on the
+		// value and the concrete value doesn't implement the
+		// interface.
+		return u.lookupType(typ.Type)
 	case *zng.TypeRecord:
 		return u.lookupTypeRecord(typ)
 	case *zng.TypeArray:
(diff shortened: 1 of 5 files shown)


zng/resolver/marshal.go
```diff
@@ -138,10 +144,10 @@ func typeFull(name, path string) string {
 type TypeStyle int
 
 const (
-	TypeStyleNone = iota
-	TypeStyleSimple
-	TypeStylePackage
-	TypeStyleFull
+	StyleNone = iota
```
nwt, line 147: Make these TypeStyles rather than just ints.
```suggestion
	StyleNone TypeStyle = iota
```

zng/resolver/marshal.go
```diff
@@ -192,9 +198,11 @@ func (m *MarshalContext) encodeValue(v reflect.Value) (zng.Type, error) {
 	if err != nil {
 		return nil, err
 	}
-
 	if m.decorator != nil || m.bindings != nil {
-		if !v.IsValid() {
+		// Don't create aliases for interface types as this is just
+		// one value for that interface and its the underlying concrete
```
nwt, line 203:
```suggestion
		// one value for that interface and it's the underlying concrete
```

zng/resolver/marshal.go
```diff
@@ -475,20 +491,29 @@ func (u *UnmarshalContext) decodeAny(zv zng.Value, v reflect.Value) error {
...
-			return fmt.Errorf("no zng decoration for interface: %v", v.Type().Name())
+		typ, err := u.lookupType(zv.Type)
+		if err != nil {
+			return err
 		}
-		alias, ok := zv.Type.(*zng.TypeAlias)
-		if !ok {
-			return fmt.Errorf("no zng decoration for interface %q", v.Type().Name())
+		if typ == nil {
+			// ZNG type null values can only by value null.
```
nwt, line 499: What's this trying to say?

zng/resolver/marshal.go
```diff
@@ -741,3 +765,100 @@ func typeNameOfValue(value interface{}) (string, error) {
...
+		if template == nil {
+			// Ignore aliases for which there are no bindings.
+			// If an interface type being marshaled into doesn't
+			// have a binding, then a type mismatch will be caught
+			// by reflect when the Set() method is called on the
+			// value and the concrete value doesn't implement the
+			// interface.
+			return u.lookupType(typ.Type)
+		}
+		return template, nil
```
nwt, line 782:
```suggestion
		if template := u.binder.lookup(typ.Name); template != nil {
			return template, nil
		}
		// Ignore aliases for which there are no bindings.
		// If an interface type being marshaled into doesn't
		// have a binding, then a type mismatch will be caught
		// by reflect when the Set() method is called on the
		// value and the concrete value doesn't implement the
		// interface.
		return u.lookupType(typ.Type)
```

zng/resolver/marshal.go
```diff
@@ -741,3 +765,100 @@ func typeNameOfValue(value interface{}) (string, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		return reflect.SliceOf(elemType), nil
+	case *zng.TypeSet:
+		elemType, err := u.lookupType(typ.Type)
+		if err != nil {
+			return nil, err
+		}
+		return reflect.SliceOf(elemType), nil
```
nwt, line 796:
```suggestion
	case *zng.TypeArray, *zng.TypeSet:
		elemType, err := u.lookupType(zng.InnerType(typ))
		if err != nil {
			return nil, err
		}
		return reflect.SliceOf(elemType), nil
```

zng/resolver/marshal.go
```diff
@@ -741,3 +765,100 @@ func typeNameOfValue(value interface{}) (string, error) {
...
+	case *zng.TypeMap:
+		KeyType, err := u.lookupType(typ.KeyType)
+		if err != nil {
+			return nil, err
+		}
+		valType, err := u.lookupType(typ.ValType)
+		if err != nil {
+			return nil, err
+		}
+		return reflect.MapOf(KeyType, valType), nil
```
nwt, line 812:
```suggestion
		keyType, err := u.lookupType(typ.KeyType)
		if err != nil {
			return nil, err
		}
		valType, err := u.lookupType(typ.ValType)
		if err != nil {
			return nil, err
		}
		return reflect.MapOf(keyType, valType), nil
```

zng/resolver/marshal.go
```diff
@@ -741,3 +765,100 @@ func typeNameOfValue(value interface{}) (string, error) {
...
+
+func (u *UnmarshalContext) lookupPrimitiveType(typ zng.Type) (reflect.Type, error) {
+	var v interface{}
+	switch typ := typ.(type) {
+	// XXX We should have counterparts for bstring, error, and type type.
+	// See issue #1853.
+	case *zng.TypeOfString, *zng.TypeOfBstring, *zng.TypeOfError, *zng.TypeOfType:
+		v = ""
+	case *zng.TypeOfBool:
+		v = true
```
nwt, line 830: Make this the zero value like the others.
```suggestion
		v = false
```

zson/marshal_test.go
```diff
@@ -0,0 +1,98 @@
...
+		return &Animal{color}
+	default:
+		return nil
+	}
+}
+
+func TestInterfaceMarshal(t *testing.T) {
+	rose := Make("plant", "red")
+	expectedRose := `{MyColor: "red"} (=Plant)`
+	flamingo := Make("animal", "pink")
```
nwt, line 46:
```suggestion

func TestInterfaceMarshal(t *testing.T) {
	rose := Thing(&Plant{"red"})
	expectedRose := `{MyColor: "red"} (=Plant)`
	flamingo := Thing(&Animal{"pink"})
```

nwt (approved):

zson/marshal_test.go
mccanne: This came from my article draft to be pedantic.  I will simplify the test here.

PR #1858, opened 2020-12-21
Add compiler.IsParallelizable

### LINKED ISSUE #1839: ZQD: separate detection of parallelism from the changes to flow graph
Currently the code that detects whether a flow graph is capable of parallelism also mutates the flow graph to add a fixed number of Proc's that will implement the parallelism. This is awkward for distributed queries, where we would like to know If parallel processing is needed before we do a remote call to determine how many workers we can recruit. The number of workers recruited will vary based on the state of cluster and customer preferences for resources, but in the current code the flow graph has already been mutated with a fixed number of process before recruitment. 

The relevant code for this starts in driver/compile.go compileMulti().

One approach would be to split compiler.Parallelize into two parts, one that detects the potential for parallelism, and another to mutate the flow graph. Then the part that mutates the flow graph could be moved to run after createParallelGroup. I looked at doing this but it is a big change and I did not thoroughly understand all the cases under the "switch p := seq.Procs[i].(type)" statement in compiler.Parallelize.

Another idea would be to implement a new type of proc that does remote parallel processing, rather than having to add a static number of proc's to the flow graph. This would also allow us to simplify the methods for parallelGroup.

nwt:
This implementation is inefficient, but that's OK for now.

Closes #1839.

### DIFF
--- a/compiler/compile_test.go
+++ b/compiler/compile_test.go
@@ -291,6 +291,10 @@ func TestParallelizeFlowgraph(t *testing.T) {
 		t.Run(tc.zql, func(t *testing.T) {
 			query, err := zql.ParseProc(tc.zql)
 			require.NoError(t, err)
+
+			ok := IsParallelizable(query.(*ast.SequentialProc), sf(tc.orderField), false)
+			require.Equal(t, ok, tc.zql != tc.expected)
+
 			parallelized, ok := Parallelize(query.(*ast.SequentialProc), 2, sf(tc.orderField), false)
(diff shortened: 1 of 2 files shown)


henridf:
> that's OK for now.

and forever!

henridf (approved):

marktwallace (approved):
This will work so I can fix the problem described in #1839.  I know it is inefficient, but it is really nice that you were able to do this without changing Parallelize() -- the "ok" value is basically what I need, and by using copyProcs, this does not mutate the AST yet. Thanks!

PR #1860, opened 2020-12-21
Replace stopErr with stopOnErr everywhere

nwt:
Various fields named stopErr are bools, but that name reads like an
error.  Replace it with stopOnErr, which reads like a bool.


henridf:
This leaves some stopErrs in `api/{api.go,connection.go}` and `ppl/zqd/ingest/log.go`... shouldn't those change too?

nwt:
I'll reopen this when I have time to work on it.

Issue #1833, opened 2020-12-17
Add facility to run Z scripts at ingest


henridf:
Add a facility to run Z scripts on ingested data, before it lands. This should also be applicable to the pcap ingest path.

philrz:
While expectations have been set in #1870 that the functionality is limited, I've verified what we've got thus far in `zqd` commit `4dff9e9`.

I took the approach of trying to "shape" an EVE JSON log from a standalone/separate Suricata v5.0.3. While the following Z is not a 100% match of the functionality currently being achieved through the legacy JSON typing system that's invoked during Brim pcap import, it gets fairly close:

```
$ cat shape-suricata.zs 
event_type="alert"
| put src_ip=src_ip:ip, dest_ip=dest_ip:ip
| put ts=iso(replace(timestamp, "-0700", "-07:00"))
| cut ts,event_type,src_ip,src_port,dest_ip,dest_port,vlan,proto,app_proto,alert,flow_id,pcap_cnt,tx_id,icmp_code,icmp_type,community_id
| fuse
```

I found a couple unrelated bugs while working on this:

1. Because of #1907, I manually removed a single appearance of `"ja3s":{}` from my `eve.json` before attempting to import it. I was not able to find another way to work around the issue.
2. The `replace()` call above works around #1905, which is nice validation of some of the flexibility that's possible with the new "shaper" approach.

I've put the modified EVE JSON file up at https://storage.googleapis.com/brimsec/issues/zq-1833/eve.json.gz in case it might be useful in future testing.

To import:

```
$ zapi -s eve postpath -z "$(cat shape-suricata.zs)" eve.json.gz 
100.0% 31.53MB/31.53MB
posted 31.53MB in 31.97698114s
```

Once imported, I was able to click **View > Reload** in my Brim app that had spawned this `zqd` and see the Space and its contents. We can see some of the benefits of the shaping in the form of the populated time picker.

![image](https://user-images.githubusercontent.com/5934157/103320610-e3ebfa00-49ea-11eb-9243-9c9577355d58.png)

The same is true of other functionality that relies on correct data typing, such as doing a CIDR match on `ip` types.

![image](https://user-images.githubusercontent.com/5934157/103320623-f49c7000-49ea-11eb-9f4b-f9041c945e4a.png)

[...]

PR #1865, opened 2020-12-22
Don't pass in flowgraph warnings channel from outside

henridf:
This seems like an improvement, prompted by getting confused by the client-supplied warnings channel while working on #1833 (I initially thought that the point of passing it in was that I could then read from it... but of course the driver reads from it). From the commit:

Previously, the driver package allowed clients to pass in a flowgraph
warnings channel. This was used by clients that created a channel for
use by a zbuf.WarningReader, and then passed that channel to a
flowgraph.

This commit changes zbuf.WarningReader to take a callback instead. It is
a slightly simpler API and when used in conjunction with the driver package,
the pattern is cleaner: the driver creates its warning channel
internally, and the neither WarningReader nor the client plumbing
things together needs to know about either one.

### DIFF
--- a/zbuf/warning.go
+++ b/zbuf/warning.go
@@ -6,21 +6,26 @@ import (
 	"github.com/brimsec/zq/zng"
 )
 
-type WarningReader struct {
-	zr Reader
-	ch chan string
+type Warner interface {
+	Warn(msg string) error
 }
 
-// WarningReader returns a Reader that reads from zr.  Any error encountered is
-// sent to ch, and then a nil *zng.Record and nil error are returned.
-func NewWarningReader(zr Reader, ch chan string) *WarningReader {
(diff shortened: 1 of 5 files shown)


nwt (approved):
Definitely an improvement!

zbuf/warning.go
```diff
@@ -6,21 +6,26 @@ import (
 	"github.com/brimsec/zq/zng"
 )
 
+type Warner interface {
+	Warn(msg string) error
```
alfred-landrum, line 10: Consider removing the error return value, as it's not used.

```suggestion
	Warn(msg string)
```

alfred-landrum (approved):

zbuf/warning.go
henridf: It's needed so that a driver.Driver implements the interface.

PR #1868, opened 2020-12-22
Recruit fewer workers than requested

marktwallace:
This follows up on PR #1858.

These changes allow a distributed query to run with fewer workers than requested.


driver/compile.go
```diff
@@ -134,6 +129,11 @@ func compileMulti(ctx context.Context, program ast.Proc, zctx *resolver.Context,
 	if err != nil {
 		return nil, err
 	}
+	mcfg.Parallelism = len(sources)
+	if mcfg.Parallelism > 1 {
+		program, isParallel = compiler.Parallelize(program, mcfg.Parallelism, sortKey, sortReversed)
+	}
+
```
nwt, line 136: Why move this down here when you could keep it next to the compiler.IsParallelizable call?

ppl/zqd/recruiter/longpoll.go
```diff
@@ -19,6 +19,10 @@ func RecruitWithEndWait(wp *WorkerPool, numberRequested int, loggingLabel string
 			workers = append(workers, api.Worker{Addr: w.Addr, NodeName: w.NodeName})
 		}
 	}
+	logger.Info("Recruit request",
```
nwt, line 22:
```suggestion
	logger.Debug("Recruit request",
```

driver/compile.go
nwt: Oh, it's because you need `len(sources)` for the compiler.Parallelize call. That's hard to see because you're storing it in `mcfg.Parallelism`, so don't.
```suggestion
	if len(sources) > 1 {
		program, _ = compiler.Parallelize(program, len(sources), sortKey, sortReversed)
	}
```

driver/compile.go
```diff
@@ -108,22 +108,17 @@ func compileMulti(ctx context.Context, program ast.Proc, zctx *resolver.Context,
...
 	sortKey, sortReversed := msrc.OrderInfo()
 	filterExpr, program := compiler.Optimize(zctx, program, sortKey, sortReversed)
 
-	var isParallel bool
-	if mcfg.Parallelism > 1 {
-		program, isParallel = compiler.Parallelize(program, mcfg.Parallelism, sortKey, sortReversed)
-	}
+	isParallel := compiler.IsParallelizable(program, sortKey, sortReversed)
 	if !isParallel {
 		mcfg.Parallelism = 1
```
nwt, line 120:
```suggestion
	if !compiler.IsParallelizable(program, sortKey, sortReversed) {
		mcfg.Parallelism = 1
```

nwt (changes requested):

alfred-landrum (approved):

ppl/zqd/recruiter/longpoll.go
nwt, line 22:
```suggestion
	logger.Debug("Recruit request",
```

nwt (approved):

PR #1870, opened 2020-12-22
Ingest shaper

### LINKED ISSUE #1833: Add facility to run Z scripts at ingest

henridf:
This commit adds support for a running Z "shaper" script on incoming logs. This facility is intended for ETL tasks such as transforming record layouts and field names, casting leaf types, or policing incoming types against a set of admissible schemas. (Z script does not quite yet have the support for much of this, but that is imminent).

Closes #1833

### DIFF
--- a/api/client/connection.go
+++ b/api/client/connection.go
@@ -358,7 +358,7 @@ type LogPostOpts struct {
 	Shaper    ast.Proc
 }
 
-func (c *Connection) LogPostPath(ctx context.Context, space api.SpaceID, opts LogPostOpts, paths ...string) error {
+func (c *Connection) LogPostPath(ctx context.Context, space api.SpaceID, opts *LogPostOpts, paths ...string) error {
 	stream, err := c.LogPostPathStream(ctx, space, opts, paths...)
 	if err != nil {
 		return err
@@ -370,13 +370,13 @@ func (c *Connection) LogPostPath(ctx context.Context, space api.SpaceID, opts Lo
 	return payloads.Error()
 }
 
-func (c *Connection) LogPostPathStream(ctx context.Context, space api.SpaceID, opts LogPostOpts, paths ...string) (*Stream, error) {
+func (c *Connection) LogPostPathStream(ctx context.Context, space api.SpaceID, opts *LogPostOpts, paths ...string) (*Stream, error) {
 	body := api.LogPostRequest{
 		Paths:          paths,
 		JSONTypeConfig: opts.JSON,
 		StopErr:        opts.StopError,
 	}
-	if opts.Shaper != nil {
+	if opts != nil && opts.Shaper != nil {
 		raw, err := json.Marshal(opts.Shaper)
 		if err != nil {
 			return nil, err
--- a/api/client/connection.go
+++ b/api/client/connection.go
@@ -417,6 +417,9 @@ func (c *Connection) LogPostWriter(ctx context.Context, space api.SpaceID, opts 
 		SetResult(&api.LogPostResponse{}).
 		SetHeader("Content-Type", writer.ContentType())
 	if opts != nil {
+		if opts.Shaper != nil {
+			writer.SetShaper(opts.Shaper)
+		}
 		if opts.StopError {
 			req.SetQueryParam("stop_err", "true")
 		}
--- a/api/client/connection.go
+++ b/api/client/connection.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	"github.com/brimsec/zq/api"
+	"github.com/brimsec/zq/ast"
 	"github.com/brimsec/zq/pcap/pcapio"
 	"github.com/brimsec/zq/zio/ndjsonio"
 	"github.com/go-resty/resty/v2"
@@ -354,9 +355,10 @@ type PcapReadCloser struct {
 type LogPostOpts struct {
 	JSON      *ndjsonio.TypeConfig
 	StopError bool
+	Shaper    ast.Proc
 }
 
-func (c *Connection) LogPostPath(ctx context.Context, space api.SpaceID, opts *LogPostOpts, paths ...string) error {
+func (c *Connection) LogPostPath(ctx context.Context, space api.SpaceID, opts LogPostOpts, paths ...string) error {
 	stream, err := c.LogPostPathStream(ctx, space, opts, paths...)
 	if err != nil {
 		return err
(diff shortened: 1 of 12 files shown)


api/client/multipartwriter.go
```diff
@@ -125,6 +135,17 @@ func (m *MultipartWriter) sendJSONConfig() error {
...
+	if m.shaper == nil {
+		return nil
+	}
+	w, err := m.form.CreateFormField("shaper_ast")
+	if err != nil {
+		return err
+	}
+	return json.NewEncoder(w).Encode(m.shaper)
+}
+
```
mattnibs, line 148: Can we drop the `sendJSONConfig` or would this be done later?

api/client/connection.go
```diff
@@ -368,11 +370,18 @@ func (c *Connection) LogPostPath(ctx context.Context, space api.SpaceID, opts *L
 	return payloads.Error()
 }
 
-func (c *Connection) LogPostPathStream(ctx context.Context, space api.SpaceID, opts *LogPostOpts, paths ...string) (*Stream, error) {
-	body := api.LogPostRequest{Paths: paths}
-	if opts != nil {
-		body.JSONTypeConfig = opts.JSON
-		body.StopErr = opts.StopError
+func (c *Connection) LogPostPathStream(ctx context.Context, space api.SpaceID, opts LogPostOpts, paths ...string) (*Stream, error) {
```
mattnibs, line 373: I would just assume keep this option a pointer so you don't have to do `api.LogPostOpts` in the vanilla post case.

api/client/connection.go
```diff
@@ -368,11 +370,18 @@ func (c *Connection) LogPostPath(ctx context.Context, space api.SpaceID, opts *L
...
+	}
+	if opts.Shaper != nil {
+		raw, err := json.Marshal(opts.Shaper)
+		if err != nil {
+			return nil, err
+		}
+		body.Shaper = raw
 	}
 	req := c.Request(ctx).
 		SetBody(body)
```
mattnibs, line 387: Shouldn't the methods `LogPost`, `LogPostReaders` and `LogPostWriter` also handle the `Shaper` option?

cmd/zapi/cmd/post/post.go
```diff
@@ -55,6 +55,9 @@ func (c *PostCommand) Run(args []string) (err error) {
 	if err != nil {
 		return err
 	}
+	if c.postFlags.shaperAST != nil {
+		c.logwriter.SetShaper(c.postFlags.shaperAST)
+	}
```
mattnibs, line 60: Perhaps in this case create a `LogPostOpts` with the ast and pass this into `LogPostWriter`?

ppl/zqd/ingest/multipartlogreader.go
```diff
@@ -46,6 +53,12 @@ read:
 		if zr == nil || err != nil {
 			return nil, err
 		}
+		if m.shaperAST != nil {
+			zr, err = driver.NewReader(context.Background(), m.shaperAST, m.zctx, zr)
+			if err != nil {
+				return nil, err
+			}
+		}
```
mattnibs, line 61: Consider moving this into the `next` function that creates the reader.

mattnibs (commented):
This great! Can't wait to use this on the timestamp transform for suricata!

mattnibs (approved):

api/client/multipartwriter.go
henridf: Hmm, I'm assuming no one is using `sendJSONConfig` on multi-upload endpoint? Happy to remove right away if that's the case.

ppl/zqd/ingest/multipartlogreader.go
```diff
@@ -57,14 +70,14 @@ read:
...
-			m.appendWarning(zr, err)
+			m.appendWarning(zr.(*zbuf.File), err)
 		}
 		goto read
 	}
 	return rec, err
 }
 
-func (m *MultipartLogReader) next() (*zbuf.File, error) {
+func (m *MultipartLogReader) next() (zbuf.ReadCloser, error) {
```
alfred-landrum, line 80: Can you explain why you needed to change this return type?

driver/driver.go
```diff
@@ -214,9 +216,14 @@ func NewReaderWithConfig(ctx context.Context, conf Config, program ast.Proc, zct
...
 	}
-	return &muxReader{
+	mr := &muxReader{
 		cancel:      cancel,
 		muxOutput:   mux,
 		statsTickCh: make(chan time.Time),
-	}, nil
+		zr:          ioutil.NopCloser(nil),
+	}
+	if _, ok := reader.(io.Closer); ok {
```
alfred-landrum, line 225: I'm not seeing why the changes required added this type assertion, can you briefly explain?

ppl/zqd/ingest/multipartlogreader.go
henridf: Because this might be passed to driver.NewReader which expects a zbuf.ReadCloser.

driver/driver.go
henridf: The reader might be a `multipart.Part` which needs to be closed.
One alternate implementation would be to change the signature to pass a `zbuf.ReadCloser` instead, and do away with the type assertion.

cmd/zapi/cmd/post/post.go
henridf: Absolutely! Not sure why I didn't do that in the first place.

api/client/multipartwriter.go
mattnibs: I would assume its being used (cc: @philrz). Isn't there some default jsonConfig we use for zeek logs and suricata logs? Perhaps wait until we have a shaper that replaces the legacy config?

api/client/connection.go
henridf: done in 085394f24905391eff81c5d9a26e0017284d9352

cmd/zapi/cmd/post/post.go
henridf: done in 085394f24905391eff81c5d9a26e0017284d9352

api/client/connection.go
henridf: Done in 9cfce89f151ce9b9497e4ffa92f5215b6c1c2f4e.

(I think passing a zero-valued struct is preferable to passing a nil, but that's best kept to a separate change)

api/client/multipartwriter.go
philrz: It's not clear from the code fragment alone what user-facing functionality we might be talking about dropping here. If we're talking about something the Brim app uses, yes, there's built-in JSON typing config for out-of-the-box Zeek v3.x that is applied whenever a user attempts to import NDJSON (in fact, I have PR https://github.com/brimsec/brim/pull/1307 up right now for review that proposes an adjustment for it so we're better prepared for Zeek v4.x). We put that in place way back when we weren't sure how many users would bring-their-own-Zeek-TSV-logs vs. bring-their-own-Zeek-NDJSON-logs vs. bring-their-own-pcaps. By now we know that pcaps are by far what's been most popular, and at this point I can't say I'm aware of any community user that actually imported their own Zeek NDJSON logs. That doesn't mean it hasn't happened... one of the negatives of something being built-in and "just working" is that users might try it and succeed and not think twice about it. That said, knowing how brittle it is. I'd be willing to guess it's not regularly used, if ever. The one community user I _do_ know that went down the route of creating custom NDJSON type configs ended up doing it with `zq` outside of Brim to turn their NDJSON into ZNG and then importing _that_ into Brim, for example.

If it's not hurting anything to leave it alone until the ingest shaper has reached feature parity with what was there before, that might be the safest option.

henridf: 👍 agreed

ppl/zqd/ingest/multipartlogreader.go
henridf: done in 6a76767624ade2687273a0b2d3dbeef14d2a7673

Issue #1702, opened 2020-11-29
Non-deterministic "fuse" behavior


philrz:
Should `fuse` behave in a deterministic manner? I've noticed that one of the `fuse` docs examples I previously created fails the automated tests occasionally because its output sometimes flips.

Using the [Zeek TSV zq-sample-data](https://github.com/brimsec/zq-sample-data/tree/master/zeek-default), as of `zq` commit `0759cc38` this base output seems to be generated deterministically:

```
$ zq -t 'ts < 1521911721' stats.log.gz weird.log.gz
#0:record[_path:string,ts:time,peer:bstring,mem:uint64,pkts_proc:uint64,bytes_recv:uint64,pkts_dropped:uint64,pkts_link:uint64,pkt_lag:duration,events_proc:uint64,events_queued:uint64,active_tcp_conns:uint64,active_udp_conns:uint64,active_icmp_conns:uint64,tcp_conns:uint64,udp_conns:uint64,icmp_conns:uint64,timers:uint64,active_timers:uint64,files:uint64,active_files:uint64,dns_requests:uint64,active_dns_requests:uint64,reassem_tcp_size:uint64,reassem_file_size:uint64,reassem_frag_size:uint64,reassem_unknown_size:uint64]
0:[stats;1521911720.600725;zeek;74;26;29375;-;-;-;404;11;1;0;0;1;0;0;36;32;0;0;0;0;1528;0;0;0;]
#port=uint16
#1:record[_path:string,ts:time,uid:bstring,id:record[orig_h:ip,orig_p:port,resp_h:ip,resp_p:port],name:bstring,addl:bstring,notice:bool,peer:bstring]
1:[weird;1521911720.600843;C1zOivgBT6dBmknqk;[10.47.1.152;49562;23.217.103.245;80;]TCP_ack_underflow_or_misorder;-;F;zeek;]
1:[weird;1521911720.608108;-;[-;-;-;-;]truncated_header;-;F;zeek;]
1:[weird;1521911720.610033;C45Ff03lESjMQQQej1;[10.47.5.155;40712;91.189.91.23;80;]above_hole_data_without_any_acks;-;F;zeek;]
1:[weird;1521911720.742818;Cs7J9j2xFQcazrg7Nc;[10.47.8.100;5900;10.129.53.65;58485;]connection_originator_SYN_ack;-;F;zeek;]
```

However, if I `fuse` it and turn it into CSV output, sometimes the first unique columns are the ones associated with the `weird` events:

```
[...]

PR #1880, opened 2020-12-23
Extract Fuser from proc/fuse.Proc

nwt:
Groundwork for #1271.

### DIFF
--- a/proc/fuse/fuse.go
+++ b/proc/fuse/fuse.go
@@ -6,7 +6,6 @@ import (
 	"github.com/brimsec/zq/proc"
 	"github.com/brimsec/zq/zbuf"
 	"github.com/brimsec/zq/zng"
-	"github.com/brimsec/zq/zng/fuser"
 )
 
 var MemMaxBytes = 128 * 1024 * 1024
@@ -16,7 +15,7 @@ type Proc struct {
 	pctx   *proc.Context
 	parent proc.Interface
 
-	fuser    *fuser.Fuser
+	fuser    *Fuser
 	once     sync.Once
 	resultCh chan proc.Result
 }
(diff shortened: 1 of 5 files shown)


zng/fuser/fuser.go
```diff
@@ -0,0 +1,173 @@
...
+)
+
+// Fuser buffers records written to it, assembling from them a unified schema of
+// fields and types.  Fuser then transforms those records to the unified schema
+// as they are read back from it.
+type Fuser struct {
+	zctx        *resolver.Context
+	memMaxBytes int
+
+	spiller  *spill.File
```
alfred-landrum, line 20: nit: sort this & below field names.

proc/fuse/fuse.go
```diff
@@ -18,22 +15,17 @@ var BatchSize = 100
...
-	spiller  *spill.File
-	slotByID [][]int
-	slots    []slot
-	uberType *zng.TypeRecord
 	once     sync.Once
 	resultCh chan proc.Result
-	nbytes   int
-	recs     []*zng.Record
-	types    resolver.Slice
+	fuser    *fuser.Fuser
```
alfred-landrum, line 20: nit: move above `once`.

alfred-landrum (commented):
The test failure looks this existing fuse issue: #1702 , not necessarily something in this PR's changes.

proc/fuse/fuse.go
nwt: Done in a97847d99de10b5b91f5840cebcb2cb492349374.

zng/fuser/fuser.go
nwt: Done in 7f31208919e2c0013640c91752be292e36a5502f.

zng/fuser/fuser.go
```diff
@@ -0,0 +1,173 @@
...
+	f.nbytes += len(rec.Raw)
+	if f.nbytes >= f.memMaxBytes {
+		var err error
+		f.spiller, err = spill.NewTempFile()
+		if err != nil {
+			return err
+		}
+		if err := f.spiller.Write(rec); err != nil {
+			return err
+		}
```
mattnibs, line 75: Shouldn't this go after the recs iteration to keep the order or does this not matter?

nwt: I think fuse should preserve order and created #1904 to track that.

nwt:
`zng/fuser` didn't feel like the right home for `Fuser`, so I moved it back to `proc/fuse`.

alfred-landrum (approved):

PR #1883, opened 2020-12-24
Add context param to emitter.New* and outputflag.Flags.Open

### LINKED ISSUE #1881: Add context.Context parameter to emitter.New*
For #1271, I want to wrap a `csvio.Writer` with a new `FuseWriter` that will send records through a `fuser.Fuser`. Since `Fuser` has to see its entire input before it can produce any output, `FuseWriter` will wait until it's closed before touching the underlying `csvio.Writer`, at which point it will copy the `Fuser`'s buffered records to the `csvio.Writer`.

We want the copy operation to observe `context.Context` cancelation. For the `zq` command, the straightforward way to do that appears to involve threading the `Context` from `signalctx.New` through `outputflags.Flags.Open` into `emitter.New*` and finally to `source.NewWriter`, which already accepts a `Context`.

nwt:
Closes #1881.

@mattnibs: Please be sure to take a look at this one.

### DIFF
--- a/ppl/cmd/zar/find/command.go
+++ b/ppl/cmd/zar/find/command.go
@@ -109,7 +109,7 @@ func (c *Command) Run(args []string) error {
 		findOptions = append(findOptions, archive.SkipMissing())
 	}
 
-	ctx, cancel := context.WithCancel(context.Background())
+	ctx, cancel := context.WithCancel(context.TODO())
 	defer cancel()
 	outputFile := c.outputFlags.FileName()
 	if outputFile == "-" {
--- a/ppl/cmd/zar/find/command.go
+++ b/ppl/cmd/zar/find/command.go
@@ -109,6 +109,8 @@ func (c *Command) Run(args []string) error {
 		findOptions = append(findOptions, archive.SkipMissing())
 	}
 
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
 	outputFile := c.outputFlags.FileName()
 	if outputFile == "-" {
 		outputFile = ""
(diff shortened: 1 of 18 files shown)


ppl/cmd/zar/find/command.go
```diff
@@ -109,21 +109,21 @@ func (c *Command) Run(args []string) error {
 		findOptions = append(findOptions, archive.SkipMissing())
 	}
 
+	ctx, cancel := context.WithCancel(context.Background())
```
alfred-landrum, line 112: nit: make this context.TODO()

alfred-landrum (approved):

ppl/cmd/zar/find/command.go
nwt: Done in 921f5b6c7c5619c422e75072cf1d8b0f0d78220a.

python/brim/src/zqext.go
```diff
@@ -61,7 +61,7 @@ func doZqlFileEval(inquery, inpath, informat, outpath, outformat string) (err er
 	}
 	defer rc.Close()
 
-	w, err := emitter.NewFile(outpath, zio.WriterOpts{
+	w, err := emitter.NewFile(context.Background(), outpath, zio.WriterOpts{
```
mattnibs, line 64: Shouldn't this be TODO?
```suggestion
	w, err := emitter.NewFile(context.TODO(), outpath, zio.WriterOpts{
```

mattnibs (approved):

python/brim/src/zqext.go
nwt: I used `context.Background` since that's what's used below for the call to `driver.Run`.

Issue #1892, opened 2020-12-28
implement partials method in dcount aggregator


mccanne:
The `dcount` aggregator currently does not implement partials.  This is easy to remedy as the hyperloglog++ library can merge its internal sketches so we need to merely serialize the sketch in ResultAsPartial() and merge them back together in ConsumeAsPartial().

mccanne:
When this is fixed, please update the caveat in zql/docs/aggregate-functions/README.md regarding countdistinct.

philrz:
Verified in Zed commit 7ea6c59.

Taking a step back, an app-level summary of the user experience of this issue is captured in https://github.com/brimdata/brim/issues/2447#issuecomment-1189520229.  In brief, the issue could be produced in Zed 834a633 by loading the [ZNG zed-sample-data](https://github.com/brimdata/zed-sample-data/tree/main/zng) into a pool and running a query:

```
$ zed query 'from foo | summarize allClientIPCount := dcount(id.orig_h)'
Post "http://localhost:9867/query?ctrl=T": EOF
```

With the panic on stderr for the crashed `zed serve` process:

```
panic: dcount: partials not yet implemented

goroutine 273 [running]:
github.com/brimdata/zed/runtime/expr/agg.(*DCount).ResultAsPartial(0xc0000d1e00, 0x22974a0?)
	/Users/phil/work/zed/runtime/expr/agg/dcount.go:44 +0x27
github.com/brimdata/zed/runtime/op/groupby.(*Aggregator).readTable(0xc0002f07e0, 0x1, 0x1, {0x1cf06d0, 0xc0000c83c0})
	/Users/phil/work/zed/runtime/op/groupby/groupby.go:538 +0x6ca
github.com/brimdata/zed/runtime/op/groupby.(*Aggregator).nextResult(0x0?, 0x0?, {0x1cf06d0?, 0xc0000c83c0?})
	/Users/phil/work/zed/runtime/op/groupby/groupby.go:399 +0x9d
github.com/brimdata/zed/runtime/op/groupby.(*Proc).run.func1(0xc00035a5a0)
	/Users/phil/work/zed/runtime/op/groupby/groupby.go:173 +0x39
github.com/brimdata/zed/runtime/op/groupby.(*Proc).run(0xc00035a5a0)
	/Users/phil/work/zed/runtime/op/groupby/groupby.go:195 +0x132
created by github.com/brimdata/zed/runtime/op/groupby.(*Proc).Pull.func1
[...]

PR #1893, opened 2020-12-28
simplify and move reducers to expr/agg

mccanne:
This commit simplifies reducers and group-by so that all aggregators are
now decomposable.  It also clarifies the new agg.Function interface
with respect to partials.  Since first() and last() are not decomposable,
we removed them in favor of an any() aggregator as in systems like
clickhouse.  If and when we re-introduce non-decomposable aggregators,
we should call them something else and implement them differently at
the group-by layer since they cannot be decomposed the same way for
distributed aggregation (e.g., they need to work more like join/sort
depending on their sensitivity to input order).

We also reduced the footprint of each group-by row by removing the stats
counters that weren't used and moving the "expr" and "where" interface pointers
out of group-by row and into the group-by template for each row since
these pointers are the same for every row in the table.

These changes allow us to use aggregators now in expression context,
which will be useful in upcoming PRs.  It also sets us up to do a further
refactor in group-by to better handle distributed group-bys and the merging
of partials across a cluster.


compiler/compile_test.go
```diff
@@ -265,8 +265,9 @@ func TestParallelizeFlowgraph(t *testing.T) {
 		{
 			"* | put x=y | countdistinct(x) by y | uniq",
 			"ts",
-			"(filter * | put x=y; filter * | put x=y) | countdistinct(x) by y | uniq",
-			"ts",
+			" (filter * | put x=y | countdistinct(x) by y  ; filter * | put x=y | countdistinct(x) by y) | countdistinct(x) by y | uniq",
+			// shouldn't the merge-field be y?
```
henridf, line 269: is the comment still needed? (if so, might be rewored to clarify that it also applies to cases below)

expr/agg.go
```diff
@@ -0,0 +1,63 @@
...
+	return a.pattern()
+}
+
+func (a *Aggregator) Apply(f agg.Function, rec *zng.Record) {
+	if a.filter(rec) {
+		return
+	}
+	zv, err := a.expr.Eval(rec)
+	if err != nil {
+		//XXX
```
henridf, line 48: Can this //XXX  be addressed before merging?

zql/docs/aggregate-functions/README.md
```diff
@@ -281,56 +280,34 @@ to perform this test, the ZQL using `countdistinct()` executed almost 3x faster.
...
-| **Syntax**                | `first(<field-name>)`                                          |
+| **Description**           | Return any value observed for a specified field |
+| **Syntax**                | `any(<field-name>)`                                          |
 | **Required<br>arguments** | `<field-name>`<br>The name of a field.                         |
 | **Optional<br>arguments** | None                                                           |
 
 #### Example:
 
-To see the `name` of the first Zeek `weird` event in our sample data:
+To see the `name` of the a Zeek `weird` event in our sample data:
```
henridf, line 294:
```suggestion
To see the `name` of a Zeek `weird` event in our sample data:
```

zql/docs/aggregate-functions/README.md
```diff
@@ -15,8 +15,7 @@ the events.
      + [`collect`](#collect)
      + [`count`](#count)
      + [`countdistinct`](#countdistinct)
```
henridf, line 17: Unless #1892 is going to be addressed real soon, the doc  for countdistinct should probably get a note explaining it  may not yet work in Brim? ("may" because parallelism is only used  for archive stores)

zql/docs/aggregate-functions/README.md
```diff
@@ -281,56 +280,34 @@ to perform this test, the ZQL using `countdistinct()` executed almost 3x faster.
...
 ---
 
-### `first`
+### `any`
 
 |                           |                                                                |
 | ------------------------- | -------------------------------------------------------------- |
-| **Description**           | Return the first value observed for a specified field, based on input order. |
-| **Syntax**                | `first(<field-name>)`                                          |
+| **Description**           | Return any value observed for a specified field |
```
henridf, line 287:
```suggestion
| **Description**           | Return any value observed for a specified field. |
```

henridf (approved):
Decompose all the things!

(Side note: I can see `any` one day growing up into a more general `sample(..)`...)

expr/agg.go
mccanne: Yeah good idea.  Bubbling up aggregator errors to group-by has been put off forever.  This refactor makes it easy now.  I'll fix it.

PR #1894, opened 2020-12-29
WarningReader: always return nil records after err

henridf:
While working on #1895 I ran into some unexpected zqd behavior (illustrated in the yaml test) when posting to the `log/paths` endpoint: records continue to be read after a bad line. This is not the case with the `/log` endpoint.

The bug is caused by an interaction between `zbuf.WarningReader` and `zbuf.ReadBatch`. WarningReader returns `nil, nil` after encountering an error, which means that [no records remain](https://github.com/brimsec/zq/blob/ad3def60a92b30a748d8c3f82a3831f7b45a8d52/zbuf/zbuf.go#L13). But if called subsequently, it does return more records. This trips up zbuf.ReadBatch, which needs to read two consecutive `nil` records [in order to return a nil batch](https://github.com/brimsec/zq/blob/ad3def60a92b30a748d8c3f82a3831f7b45a8d52/zbuf/batch.go#L41). 

(Arguably ReadBatch should also be adjusted ... but it doesn't seem entirely unreasonable to call an EOF'd reader a second time).

### DIFF
--- a/zbuf/warning.go
+++ b/zbuf/warning.go
@@ -11,8 +11,9 @@ type Warner interface {
 }
 
 type WarningReader struct {
-	zr Reader
-	wn Warner
+	zr    Reader
+	wn    Warner
+	erred bool
 }
 
 // NewWarningReader returns a Reader that reads from zr.  Any error
@@ -23,9 +24,13 @@ func NewWarningReader(zr Reader, w Warner) Reader {
 }
 
 func (w *WarningReader) Read() (*zng.Record, error) {
+	if w.erred {
+		return nil, nil
+	}
 	rec, err := w.zr.Read()
 	if err != nil {
 		w.wn.Warn(fmt.Sprintf("%s: %s", w.zr, err))
+		w.erred = true
 		return nil, nil
 	}
 	return rec, nil
--- a/ztests/suite/zqd/zapi-postpath-errors.yaml
+++ b/ztests/suite/zqd/zapi-postpath-errors.yaml
@@ -1,0 +1,22 @@
+script: |
+  source services.sh
+  zapi -h $ZQD_HOST new test
+  zapi -h $ZQD_HOST -s test postpath bad.tzng
+  zapi -h $ZQD_HOST -s test get -f zson "count()" > out.zson
+
+inputs:
+  - name: services.sh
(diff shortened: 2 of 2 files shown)


nwt (commented):
> (Arguably ReadBatch should also be adjusted ... but it doesn't seem entirely unreasonable to call an EOF'd reader a second time).

While that doesn't seem unreasonable at all, it does represent a change to the `zbuf.Reader` contract, and if we're going to change that, we should verify that other implementations conform. According to `git grep ' Read\(\) \(\*zng\.Record, error\) | wc -l'`, there are 36. Given that scope, changing `zbuf.ReadBatch` (or how we use it) might be a more desirable approach.

henridf:
> While that doesn't seem unreasonable at all, it does represent a change to the zbuf.Reader contract

@nwt there's no change to the zbuf.Reader contract in this PR. To help me understand, can you clarify what should be checked in the 36 lines found by git grep?

nwt:
I guess I was looking at the `zbuf.Reader` contract as making no guarantees about the behavior of `Read` after it returns `nil, nil` and at this PR as saying, "Actually, `Read` should continue to return `nil, nil`, and here's a change to one implementation to make it do that," so I was saying, "That feels like a change to the contract, so we should update the contract, and we should also make sure that all the implementation implement the contract."

henridf:
Ah! I see. 

> at this PR as saying, "Actually, Read should continue to return nil, nil,

Since ReadBatch has relied on this behavior for some time, I'd inferred that other `Read`s do continue to return `nil, nil`, or we'd have noticed this earlier. 

> That feels like a change to the contract

Hmm, I was thinking at this part of the [`zbuf.Reader` comment](https://github.com/brimsec/zq/blob/4dff9e90356230f77aff5b3c118caa80c09a95f4/zbuf/zbuf.go#L13):

>> a nil record and nil error to indicate that no records remain.

And thinking if you send something indicating that no records remain, it would be broken behavior to send a record after that. Otherwise what's the point of ever returning `nil, nil`?

Regardless, the fact that there are differing interpretations mean that this might be worthy of an after-discussion. In 2021 🎉 !

henridf:
#1962

PR #1903, opened 2020-12-29
Fix lost warnings in log/path endpoint

### LINKED ISSUE #1901: zqd /log/path endpoint drops warnings
(I ran into while working on #1895)

Given this input file with a bad line:
```
$ cat bad.tzng
#0:record[ip:string]
0:[1.1.1.1;]
0:[1.1.1.2;
```

The `/log` enpoint produces a warning for the bad line
```
$ zapi -s test post -f in-bad.tzng
warning: file:///Users/henridf/work/brim/zq/in-bad.tzng: line 3: tzng syntax error: unterminated container
posted 46B in 1.096294ms
```

But the `/log/path` endpoint does not:

```
$ zapi -s test postpath -f in-bad.tzng
100.0% 46B/46B
posted 46B in 1.204216ms
```

henridf:
close #1901 (and sets things up for #1895 )

### DIFF
--- a/ppl/zqd/ingest/log.go
+++ b/ppl/zqd/ingest/log.go
@@ -41,6 +41,7 @@ type LogOp struct {
 func NewLogOp(ctx context.Context, store storage.Storage, req api.LogPostRequest) (*LogOp, error) {
 	p := &LogOp{
 		warningCh: make(chan string, 5),
+		warnings:  make([]string, 0, 5),
 		zctx:      resolver.NewContext(),
 	}
 	opts := zio.ReaderOpts{Zng: zngio.ReaderOpts{Validate: true}}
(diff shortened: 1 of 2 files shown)


ppl/zqd/ingest/log.go
```diff
@@ -91,7 +92,13 @@ func NewLogOp(ctx context.Context, store storage.Storage, req api.LogPostRequest
 }
 
 func (p *LogOp) Warn(msg string) error {
-	p.warnings = append(p.warnings, msg)
+	// warnings received before we've started our goroutine are
+	// saved here and will be drained in start()
+	if p.warnings != nil {
+		p.warnings = append(p.warnings, msg)
```
marktwallace, line 98: I'm not clear on whether this could ever be called from a goroutine, but if it is, then writing to the channel will be thread-safe, but the append will not be thread-safe without a lock.

marktwallace (approved):
This looks fine as long as the same LogOp instance is never used in more than one goroutine.

ppl/zqd/ingest/log.go
henridf: Yes, the only goroutine that calls Warn is the one started by NewLogOp (`go p.start(ctx, store)`). This is pre-existing behavior. I'll add a comment to clarify.

philrz:
Also fixes #1895.

PR #1906, opened 2020-12-29
Fix flaky postpcap stats test

mattnibs:
Use regexp on the test which better accounts for the fluctuating size of record
bytes written (caused by zeek ids).

Also remove the hack to truncate the byte values needed to make this test
originally pass.


nwt (approved):
Great idea!

PR #1910, opened 2020-12-30
EKS smoke test

marktwallace:
This automates a very simple set of smoke tests to check if a zqd image can be deployed on EKS and works for basic distributed queries. An important feature of the smoke test is that it does not reserve exclusive resources, so it can run in parallel with itself.

The tests are simple, but the overall process is non-trivial. Since there is a lot to unpack here, I am interspersing comments in my own PR.

This adds two new jobs to what we have been running for PR's. The new jobs make extensive use of S3, so they will fail in ways that minio based testing does not.

The manual workflow dispatch dialog will be available only after this is pushed to master.


.github/workflows/cd.yaml
```diff
@@ -38,14 +37,98 @@ jobs:
...
-          ZQD_DATA_URI: s3://zqd-demo-1/mark/zqd-meta
-          ZQD_K8S_USER: ci-master
-          ZQD_TEST_CLUSTER: zq-test.us-east-2.eksctl.io
+          ZQD_DATA_URI: s3://zq-ci-master/${{ env.TODAY }}.${{ github.run_number }}
         run: |
-          aws eks --region us-east-2 update-kubeconfig --name zq-test
-          kubectl config set-context --current --namespace=ci-master
+          set -x
+          aws eks --region us-east-2 update-kubeconfig --name $CLUSTER
+          kubectl create ns $NAMESPACE || :
```
marktwallace, line 50: I switched to deploying to a temporary namespace in order to avoid collisions when to PR's start builds close together. The smoke tests run in the temporary namespace and then we remove it.

.github/workflows/cd.yaml
```diff
@@ -38,14 +37,98 @@ jobs:
...
+          kubectl scale --replicas=4 deployment/worker-zqd
+          kubectl port-forward svc/recruiter-zqd 8020:9867 &
+          kubectl port-forward svc/root-zqd 9867:9867 &
+          sleep 2 # wait for port-forwards to complete
+          zapi ls
+          zapi new -k archivestore -d $ZQD_DATA_URI $SPACE
+          # Post an archive of about 1.6 GB to run test queries.
+          zapi -s $SPACE post $SOURCE_LOG_1
+          # There should be 4 workers running now.
+          [[ 4 == $(curl http://localhost:8020/recruiter/listfree | jq -r ".workers" | jq length) ]]
```
marktwallace, line 94: The next 5 commands are the smoke test. A basic search, using different numbers of workers, followed by a count() and a group-by query. we are not trying for extensive test coverage here: we just want to fail early if the basic distributed functionality is not working on K8s. (Similar queries are also run locally with ZTest.)

.github/workflows/cd.yaml
```diff
@@ -38,14 +37,98 @@ jobs:
...
+          # Post an archive of about 1.6 GB to run test queries.
+          zapi -s $SPACE post $SOURCE_LOG_1
+          # There should be 4 workers running now.
+          [[ 4 == $(curl http://localhost:8020/recruiter/listfree | jq -r ".workers" | jq length) ]]     
+          time zapi -s $SPACE get -workers 1 -t '"FgzbkQ15TOoOENn1lb" | sort tx_hosts'
+          time zapi -s $SPACE get -workers 2 -t '"FgzbkQ15TOoOENn1lb" | sort tx_hosts'
+          time zapi -s $SPACE get -workers 4 -t '"FgzbkQ15TOoOENn1lb" | sort tx_hosts'
+          time zapi -s $SPACE get -workers 4 -t 'count()' > count.tzng
+          time zapi -s $SPACE get -workers 4 -t \
+            '* | count() by tx_hosts | sort -r count | head 5' > countbytxhost.tzng
```
marktwallace, line 100: The last two zapi queries are slow at the moment. We should fix that. Since they use 4 workers to do a full scan of the archive, they should be able to run at the same speed as the 4 worker search, about 10 seconds. But right now they take over 90 seconds. When we fix that these tests will finish sooner.

.github/workflows/cd.yaml
```diff
@@ -38,14 +37,98 @@ jobs:
...
+          diff count.tzng k8s/testdiff/count.tzng
+          diff countbytxhost.tzng k8s/testdiff/countbytxhost.tzng
```
marktwallace, line 102: I put in a small amount of diff-testing to ensure that the tests will fail when they need to. This is not intended to be a scalable pattern for diff testing -- if we want to have an extensive  test suite, then we should organize it similarly to ZTest. This pattern is OK for a few smoke tests, but it is not the way I would want to approach a longer running set of distributed tests.

.github/workflows/cd.yaml
```diff
@@ -38,14 +37,98 @@ jobs:
...
+      - name: Cleanup on success or failure
+        if: ${{ always() }}
+        env:
+          NAMESPACE: ci-${{ github.run_number }}
+        run: kubectl delete ns $NAMESPACE
```
marktwallace, line 107: This is important -- if the namespace is not removed, we will run out of resources in our EKS cluster. We will need to keep an eye out to make sure that namespaces like "ci-310" are not hanging around in our zq-test cluster.

.github/workflows/cd.yaml
```diff
@@ -38,14 +37,98 @@ jobs:
...
+  eks-master-deploy:
+    # Deploys this build to ci-master namespace so it will be available
```
marktwallace, line 109: The point of this second deploy to to make a tested version of zqd available in the ci-master namespace so we will have something relatively stable to use for dev testing like Brim UI development.

ppl/archive/multisource.go
```diff
@@ -128,9 +128,10 @@ func (m *spanMultiSource) SourceFromRequest(ctx context.Context, req *api.Worker
 			return nil, zqe.E(zqe.Invalid, "invalid chunk path: %v", p)
 		}
 		uri := tsdir.path(m.ark)
-		md, err := chunk.ReadMetadata(ctx, chunk.MetadataPath(uri, id), m.ark.DataOrder)
+		mdPath := chunk.MetadataPath(uri, id)
+		md, err := chunk.ReadMetadata(ctx, mdPath, m.ark.DataOrder)
 		if err != nil {
-			return nil, err
+			return nil, zqe.E("failed to read chunk metadata: %s %w", mdPath.String(), err)
```
marktwallace, line 134: An error got thrown here a couple of times during test development, so we added a better message so we can track it down if it happens again. I would like to be able to reproduce it.

.github/workflows/cd.yaml
```diff
@@ -38,14 +38,98 @@ jobs:
...
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+      - name: Set current date as env variable
+        run: echo "TODAY=$(date +'%y-%m-%d')" >> $GITHUB_ENV
+      - env:
+          CLUSTER: zq-test
+          NAMESPACE: ci-${{ github.run_number }}
+          SOURCE_LOG_1: s3://brim-sampledata/wrccdc/zeek-logs/files.log.gz
+          SPACE: files
+          ZQD_DATA_URI: s3://zq-ci-test/${{ env.TODAY }}.${{ github.run_number }}
```
marktwallace, line 80: Note the new bucket used here: zq-ci-test is bucket with a short retention policy (3 days right now). A new directory is created for every build, so simultaneous builds will not collide while creating and querying archives. This creates S3 garbage fast but it will get cleared out by the retention policy.

.github/workflows/cd.yaml
```diff
@@ -38,14 +38,98 @@ jobs:
...
+        run: |
+          set -x
+          make install
+          aws eks --region us-east-2 update-kubeconfig --name $CLUSTER
+          kubectl config set-context --current --namespace=$NAMESPACE
+          kubectl scale --replicas=4 deployment/worker-zqd
+          kubectl port-forward svc/recruiter-zqd 8020:9867 &
+          kubectl port-forward svc/root-zqd 9867:9867 &
+          sleep 2 # wait for port-forwards to complete
```
philrz, line 89: This was discussed on a Zoom call, but I'm documenting it here for the record. I'm conditioned to react to these kinds of hard-coded sleep timers. @marktwallace mentioned that this has been passing fine, so we could defensibly let it ride. I've just been burned many times by these sorts of timers working fine at first, then one day infrastructure or automation decides to run abnormally slow, and now someone who doesn't know the automation as well has to react to a failure that didn't involve their code. It would seem ideal if we could instead poll some kind of "health endpoint" that's only reachable via the forwarded port, with the assumption that the health endpoint would only respond affirmatively if the service is truly 100% ready.

.github/workflows/cd.yaml
```diff
@@ -38,14 +38,98 @@ jobs:
...
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-east-2
+      - uses: actions/checkout@v1
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+      - env:
+          CLUSTER: zq-test
+          NAMESPACE: ci-master
```
philrz, line 125: I know care was taken with the separate namespaces so concurrent runs shouldn't step on each other. However, at this point it seems to converge on a single namespace. Have we tested and/or know what will happen if two concurrent runs reach this step at the same time? I'm hopeful the underlying K8s operations are all atomic enough that one of them will consistently "win" and cancel the other out, as opposed to some kind of race causing deadlock or inconsistency, but figured I'd ask.

philrz (approved):
I expressed one preference for improvement and had one question. But overall I understand the essence of what's being done here, so I'm happy to see it go in. That said, I don't trust my ability to scrutinize all things K8s/AWS, so I'd recommend waiting on @alfred-landrum's feedback/approval before merging.

.github/workflows/cd.yaml
marktwallace: That is a good question: I think Helm will be alright with a second install starting during another install that is in-progress. We still should keep an eye on it, in case it can fail in some way I have not seen in my testing so far. Since it is our test environment, it is a good place to learn if and when collisions of this type can happen.

.github/workflows/cd.yaml
marktwallace: There are kubectl commands that can wait for objects to enter a "Ready" state, but being "ready" is not a guarantee that the port-forward will work, so the short sleep is a good idea in any case. A retry on the port-forward might be the easiest way to go.

.github/workflows/cd.yaml
```diff
@@ -4,9 +4,33 @@ on:
...
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-east-2
+      - name: Set current date as env variable $TODAY
+        run: echo "TODAY=$(date +'%y-%m-%d')" >> $GITHUB_ENV
+      - env:
+          CLUSTER: zq-test
+          NAMESPACE: ci-${{ github.run_number }}
+          ZQD_DATA_URI: s3://zq-ci-test/${{ env.TODAY }}.${{ github.run_number }}
+        run: |
+          set -x
```
alfred-landrum, line 27: For simplicity, if you need to use environment variables, like the `TODAY` and the `env` section, you could do so inside the run block:

```suggestion
        run: |
          set -x
          ZQD_DATA_URI=s3://zq-ci-test/$(date +'%y-%m-%d').${{ github.run_number }}
          ...
```

.github/workflows/cd.yaml
alfred-landrum: I filed #1934 to track follow up work to convert these to tagged ztests.

.github/workflows/cd.yaml
marktwallace: done

.github/workflows/cd.yaml
```diff
@@ -39,13 +59,93 @@ jobs:
         with:
           ref: ${{ github.event.pull_request.head.sha }}
       - env:
+          CLUSTER: zq-test
+          NAMESPACE: ci-${{ github.run_number }}
+          ZQD_ECR_HOST: ${{ secrets.AWS_ECR_HOST }}
+        run: |
+          set -x
+          export ZQD_DATA_URI=s3://zq-ci-test/$(date +'%y-%m-%d').${{ github.run_number }}
```
nwt, line 67:
```suggestion
          export ZQD_DATA_URI=s3://zq-ci-test/$(date +%y-%m-%d).${{ github.run_number }}
```

.github/workflows/cd.yaml
```diff
@@ -4,9 +4,31 @@ on:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-east-2
+      - env:
+          CLUSTER: zq-test
+          NAMESPACE: ci-${{ github.run_number }}
```
nwt, line 22: Nit: Set these in the `run` script. (Or not at all if they're only used once.)

.github/workflows/cd.yaml
```diff
@@ -39,13 +59,93 @@ jobs:
         with:
           ref: ${{ github.event.pull_request.head.sha }}
       - env:
+          CLUSTER: zq-test
+          NAMESPACE: ci-${{ github.run_number }}
+          ZQD_ECR_HOST: ${{ secrets.AWS_ECR_HOST }}
```
nwt, line 64: Nit: Set these in the `run` script. (Or not at all if they're only used once.)

.github/workflows/cd.yaml
```diff
@@ -39,13 +59,93 @@ jobs:
...
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-east-2
+      - uses: actions/checkout@v1
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+      - env:
+          CLUSTER: zq-test
+          NAMESPACE: ci-${{ github.run_number }}
+          SOURCE_LOG_1: s3://brim-sampledata/wrccdc/zeek-logs/files.log.gz
+          SPACE: files
```
nwt, line 95: Nit: Set these in the `run` script. (Or not at all if they're only used once.)

.github/workflows/cd.yaml
```diff
@@ -39,13 +59,93 @@ jobs:
...
+          kubectl scale --replicas=4 deployment/worker-zqd
+          kubectl port-forward svc/recruiter-zqd 8020:9867 &
+          kubectl port-forward svc/root-zqd 9867:9867 &
+          sleep 2 # wait for port-forwards to complete
+          time zapi ls
+          time zapi new -k archivestore -d $ZQD_DATA_URI $SPACE
+          # Post an archive of about 1.6 GB to run test queries.
+          time zapi -s $SPACE post $SOURCE_LOG_1
+          # There should be 4 workers running now.
+          [[ 4 == $(curl http://localhost:8020/recruiter/listfree | jq -r ".workers" | jq length) ]]
```
nwt, line 111:
```suggestion
          [[ 4 == $(curl http://localhost:8020/recruiter/listfree | jq -r ".workers | length") ]]     
```

.github/workflows/cd.yaml
```diff
@@ -39,13 +59,93 @@ jobs:
...
+          time zapi -s $SPACE get -workers 4 -t 'count()' > count.tzng
+          time zapi -s $SPACE get -workers 4 -t \
+            '* | count() by tx_hosts | sort -r count | head 5' > countbytxhost.tzng
+          diff count.tzng k8s/testdiff/count.tzng
+          diff countbytxhost.tzng k8s/testdiff/countbytxhost.tzng
+      - name: Cleanup on success or failure
+        if: ${{ always() }}
+        env:
+          NAMESPACE: ci-${{ github.run_number }}
+        run: kubectl delete ns $NAMESPACE
```
nwt, line 124:
```suggestion
        run: kubectl delete ns ci-${{ github.run_number }}
```

.github/workflows/cd.yaml
```diff
@@ -39,13 +59,93 @@ jobs:
...
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+      - env:
+          CLUSTER: zq-test
+          NAMESPACE: ci-master
           ZQD_ECR_HOST: ${{ secrets.AWS_ECR_HOST }}
-          ZQD_DATA_URI: s3://zqd-demo-1/mark/zqd-meta
-          ZQD_K8S_USER: ci-master
-          ZQD_TEST_CLUSTER: zq-test.us-east-2.eksctl.io
+          ZQD_DATA_URI: s3://zq-ci-master/manual-test
```
nwt, line 143: Nit: Set these in the run script. (Or not at all if they're only used once.)

.github/workflows/cd.yaml
```diff
@@ -21,14 +30,94 @@ jobs:
...
+          SPACE: files
+          ZQD_DATA_URI: s3://zq-ci-test/${{ env.TODAY }}.${{ github.run_number }}
+        run: |
+          set -x
+          make install
+          aws eks --region us-east-2 update-kubeconfig --name $CLUSTER
+          kubectl config set-context --current --namespace=$NAMESPACE
+          kubectl scale --replicas=4 deployment/worker-zqd
+          kubectl port-forward svc/recruiter-zqd 8020:9867 &
+          kubectl port-forward svc/root-zqd 9867:9867 &
```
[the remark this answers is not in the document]
marktwallace: kubectl will only do one forward per line, unfortunately.

.github/workflows/cd.yaml
```diff
@@ -4,9 +4,28 @@ on:
...
+          aws-region: us-east-2
+      - run: |
+          set -x
+          ZQD_DATA_URI=s3://zq-ci-test/$(date +%y-%m-%d).${{ github.run_number }}
+          aws eks --region us-east-2 update-kubeconfig --name zq-test
+          aws s3 rm --recursive $ZQD_DATA_URI || :
+          kubectl delete ns ci-${{ github.run_number }} || :
   ecr-image-push:
+    # Build and push a Docker image for zqd.
+    needs: pre-clean
```
nwt, line 28: This file would be shorter (and clearer) if you used a single job rather than a sequence of jobs linked with `needs`. For instance, you'd only need one `uses: aws-actions/configure-aws-credentials@v1` step instead of four.

marktwallace: It is important that we can clearly distinguish between the jobs that fail and succeed. Combining the jobs would make the "happy path" easier to read, but would give us information about critical failures more slowly, and would require us to dig into the output to understand what failed.

Each job has a name that corresponds to its independent goal: clean, image build, deploy, test, master deploy. This code is optimized for operational clarity, as it should be. Making the code shorter is less important that the clarity in the operation of each job.

nwt: > Combining the jobs would make the "happy path" easier to read, but would give us information about critical failures more slowly, and would require us to dig into the output to understand what failed.

I don't think that's true, but the argument's tough to make in words, so let's merge this as is and explore later.

mattnibs: > Combining the jobs would make the "happy path" easier to read, but would give us information about critical failures more slowly, and would require us to dig into the output to understand what failed.

By this reasoning, you could say that we should break up every step in the "test" job (ie make fmt, make tidy, make test-unit, make test-system etc) into its own job would make it easier to identify where the failure occurred. But if we did this we would have so many jobs the list would be huge.  Fwiw I haven't had an issue with finding the reason for a failure in "test" job.

If this was called "eks-smoke test" and it failed while everything else passed, I would immediately have in my mind the parts of my pr that could have caused the failure. Having a small number of logically grouped together test jobs makes it easier to have a clear picture of what your ci test infrastructure looks like. If you a bunch of narrowly defined tests, you place a higher burden on institutional knowledge.

I don't want to hold up this pr, but I would prefer this was structured as a single job and moreover to keep an eye on the number of jobs we are creating.

marktwallace: The reasoning is different from what you describe. This is not a single test. It is a collection of the most important external integration points with the deployment technologies we are using. If the Docker build fails, that is a much more significant failure than a unit test: it means our code is not be being built on our selected deployment platform. The deploy step failing means that our AWS infrastructure cannot accept what we built, and EKS may be misconfigured, or our K8s yaml is broken. This is a very different type of failure than a functional test. The test step in this will contain an extensive set of distributed tests, reflecting a subset of our local ZTests, and need not be broken into multiple jobs.

.github/workflows/cd.yaml
```diff
@@ -4,9 +4,28 @@ on:
   push:
     branches:
       - master
+      - eks-test-2
```
mattnibs, line 7: This should be removed no?

marktwallace: Yes, I should have taken it out. However, it does no harm, so not urgent.

PR #1916, opened 2021-01-02
remove unused data-flow analysis from the optimizer

mccanne:
This commit removes the unused data-flow analysis that is in the
compiler optimizer, namely computeColumns(), colset, and supporting
functions.  The current logic that parallelizes the flowgraph does not
rely upon this analysis.  We have spent much collective time maintaining
this code and the related tests as the flowgraph and expr/proc logic
has evolved.  We will want such analysis down the road (e.g., to support
predicate-pushdown in the column reader) and at that point,
the requirements will likely be different as the AST and flowgraph
capabilities mature, and we will likely want to use a different
design pattern based on use-def dataflow analysis from the compiler
literature.


henridf:
Change looks good but I'll defer to @alfred-landrum for approval as he had scheduled this work based on his roadmap plans.

mccanne:
> Change looks good but I'll defer to @alfred-landrum for approval as he had scheduled this work based on his roadmap plans.

Thanks, I'll talk to Al.  I had not realized this.

mccanne:
Ok, having talked to Al and Henri, the plan is that at some point soon we will integrate a cleaner abstraction around semantic analysis and will probably have semantic version of the parser-AST, which will be used for dataflow analysis (much as zson/analyzer.go has a parallel structure to ast/zson.go to do type-flow analytics of ZSON values).  At this point, we can bring the data-flow algorithms being removed here back into master.

alfred-landrum (approved):

Issue #1887, opened 2020-12-24
Make cloud auth work in zapi


philrz:
Once https://github.com/brimsec/zq/issues/1886 is implemented, `zapi` will need to take advantage of it. See https://github.com/brimsec/brim/issues/1309 for the equivalent concept in the Brim app.

In terms of approach, we've envisioned something similar to what we've seen with the `gcloud` command-line tools. That is, at the point when auth is necessary, `gcloud` provides a URL that can be opened in a browser, launching a "device authorization" flow that can be completed in the browser. @alfred-landrum has a prototype that already does this.

PR #1925, opened 2021-01-05
server side auth routes

alfred-landrum:
A few server side auth updates:
* Add an `/auth/method` route that returns the kind of authentication supported, and for the Auth0 method, the details a client needs to obtain an access token.
* Add client calls for the auth identity & method routes.
* Improve JWT validation & tests, after a reviewing what go-jwt does & does not handle internally. 

I have local changes to zapi that utilize these routes & are working with a development Auth0 tenant, will push those in a subsequent PR.

Part of #1887 .


api/api.go
```diff
@@ -315,3 +315,33 @@ type ArchiveOpenOptions struct {
 type ArchiveCreateOptions struct {
 	LogSizeThreshold *int64 `json:"log_size_threshold,omitempty"`
 }
+
+// AuthAPIAudience is the audience claim for any zqd api access token.
+// Though represented as a URL, it does not need to be a reachable location.
+const AuthAPIAudience = "https://app.brimsecurity.com"
```
mattnibs, line 321: Out of curiosity, do imagine this will always be hardcoded thusly? Would we want this to be in ppl?

api/api.go
```diff
@@ -315,3 +315,33 @@ type ArchiveOpenOptions struct {
...
+type AuthIdentityResponse struct {
+	TenantID string `json:"tenant_id"`
+	UserID   string `json:"user_id"`
+}
+
+type AuthMethod string
+
+const (
+	AuthMethodNone  AuthMethod = ""
+	AuthMethodAuth0            = "auth0"
```
mattnibs, line 332: I'm pretty sure this pattern only works with iota:
```suggestion
	AuthMethodAuth0 AuthMethod = "auth0"
```

ppl/zqd/auth.go
```diff
@@ -84,20 +120,57 @@ func newAuthenticator(ctx context.Context, logger *zap.Logger, registerer promet
...
+
+func (a *Auth0Authenticator) MethodResponse() api.AuthMethodResponse {
+	return a.methodResponse
+}
+
+func verifyAPIAudience(claims jwt.MapClaims) bool {
+	// Audience claim may either be a string, or a slice of interfaces that are
+	// strings.
+	// https://auth0.com/docs/tokens/access-tokens/get-access-tokens#multiple-audiences
+	if str, ok := claims["aud"].(string); ok {
```
mattnibs, line 153: Should `aud` be a `const` like TenetIDClaim and UserIDClaim?

mattnibs (commented):
Cool. Are we able to yet to determine if a user has access to a given space?

ppl/zqd/auth.go
```diff
@@ -7,43 +7,73 @@ import (
...
+	checker        *jwtmiddleware.JWTMiddleware
+	methodResponse api.AuthMethodResponse
+}
+
 // newAuthenticator returns a mux.MiddlewareFunc that checks for a JWT signed
 // by a key referenced in the JWKS file, has the required audience and issuer
 // claims, and contains values for a brim tenant and user id.
-func newAuthenticator(ctx context.Context, logger *zap.Logger, registerer prometheus.Registerer, config AuthConfig) (mux.MiddlewareFunc, error) {
+func newAuthenticator(ctx context.Context, logger *zap.Logger, registerer prometheus.Registerer, config AuthConfig) (*Auth0Authenticator, error) {
+	domainUrl, err := config.Validate()
```
nwt, line 70:
```suggestion
	domainURL, err := config.Validate()
```

ppl/zqd/auth.go
```diff
@@ -7,43 +7,73 @@ import (
...
+	}
+	return u, nil
+}
+
+type Auth0Authenticator struct {
+	checker        *jwtmiddleware.JWTMiddleware
+	methodResponse api.AuthMethodResponse
+}
+
 // newAuthenticator returns a mux.MiddlewareFunc that checks for a JWT signed
```
nwt, line 66: This no longer returns a mux.MiddlewareFunc.

ppl/zqd/auth.go
```diff
@@ -7,43 +7,73 @@ import (
...
 // by a key referenced in the JWKS file, has the required audience and issuer
 // claims, and contains values for a brim tenant and user id.
-func newAuthenticator(ctx context.Context, logger *zap.Logger, registerer prometheus.Registerer, config AuthConfig) (mux.MiddlewareFunc, error) {
+func newAuthenticator(ctx context.Context, logger *zap.Logger, registerer prometheus.Registerer, config AuthConfig) (*Auth0Authenticator, error) {
+	domainUrl, err := config.Validate()
+	if err != nil {
+		return nil, err
+	}
+	// Auth0 issuer is always the url domain plus trailing "/".
```
nwt, line 74:
```suggestion
	// Auth0 issuer is always the domain URL with trailing "/".
```

ppl/zqd/auth.go
```diff
@@ -84,20 +120,57 @@ func newAuthenticator(ctx context.Context, logger *zap.Logger, registerer promet
...
+				Domain:   config.Domain,
+				ClientID: config.ClientID,
+			},
+		},
 	}, nil
 }
 
+func (a *Auth0Authenticator) Middleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if err := a.checker.CheckJWT(w, r); err != nil {
```
nwt, line 137:
```suggestion
		if a.checker.CheckJWT(w, r) != nil {
```

api/api.go
alfred-landrum: You're correct, thanks: https://golang.org/ref/spec#Constant_declarations

ppl/zqd/auth.go
alfred-landrum: It's standard (ie, well known) claim for a JWT (like "GET" as an HTTP verb), unlike our custom claims for user & tenant id.

api/api.go
alfred-landrum: ( @nwt asked a similar question below). The audience needs to be known to the client, as it needs to use it when it asks for an access token from Auth0. I hadn't put that in the method response as it doesn't need to be different between different environments, but I think it'd be more consistent to do so.

api/api.go
```diff
@@ -315,3 +315,33 @@ type ArchiveOpenOptions struct {
 type ArchiveCreateOptions struct {
 	LogSizeThreshold *int64 `json:"log_size_threshold,omitempty"`
 }
+
+// AuthAPIAudience is the audience claim for any zqd api access token.
```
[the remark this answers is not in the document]
alfred-landrum: (see reply to matt's similar question above)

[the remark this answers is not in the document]
nwt: I mean in a new file, `api/auth.go`. 😃

alfred-landrum:
> Cool. Are we able to yet to determine if a user has access to a given space?

Not yet. In a subsequent PR, we'll have a check to compare the tenant owner of a space, and verify that only users that are a member of that tenant can read/write to it.

api/api.go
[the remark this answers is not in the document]
alfred-landrum: If you also meant the other structs (`AuthMethodResponse`, `AuthMethodResponse`) I would prefer they stay in the api package.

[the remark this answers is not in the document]
nwt: Not sure if this was coming through, so to be clear, I'm not suggesting moving this stuff to another package but rather moving it to an `auth.go` file here in the `api` package, similar to `worker.go` and `recruiter.go`. I'm also fine with keeping it as is.

mattnibs (approved):

api/api.go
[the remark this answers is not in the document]
alfred-landrum: ah, ok!

PR #1937, opened 2021-01-06
zqd: package db

### LINKED ISSUE #1936: zqd: package db reorganization
Currently the logic for working with the zqd database responsible for storing SpaceRows is in different locations. Let's centralize all db related functionality for zqd into a single package. The package layout should look like this:

```
/ppl/zqd/
  db/ [db config, db flags, loads a db based on config. imports postgresdb, filedb]
    postgresdb/ [postgres related logic, imports schema]
    filedb/ [file related logic, imports schema]
    schema/ [contains SpaceRow and other data mapped to a db table]
```

mattnibs:
Move all db related functionality to single db package in ppl/zqd/db

The structure looks like this:

db/ - db configuration, flags, DB interface, db.Open function.
db/filedb/ - file store implementation of DB interface
db/postgresdb/ - postgres config and implementation of DB interface
db/schema/ - data persisted in a DB instance

Closes #1936


ppl/zqd/db/db.go
```diff
@@ -0,0 +1,49 @@
...
+	var err error
+	switch conf.Kind {
+	case DBFile, DBUnspecified:
+		db, err = filedb.Open(ctx, logger, root)
+	case DBPostgres:
+		db, err = postgresdb.Open(ctx, conf.Postgres)
+	default:
+		return nil, fmt.Errorf("db.Open: unknown DBKind %q", conf.Kind)
+	}
+	return db, err
```
nwt, line 48: Maybe shorten?
```suggestion
	switch conf.Kind {
	case DBFile, DBUnspecified:
		return filedb.Open(ctx, logger, root)
	case DBPostgres:
		return postgresdb.Open(ctx, conf.Postgres)
	default:
		return nil, fmt.Errorf("db.Open: unknown DBKind %q", conf.Kind)
	}
```

ppl/zqd/db/filedb/filedb.go
```diff
@@ -9,23 +9,75 @@ import (
...
+	dburi := root.AppendPath(dbname)
+	exists, err := iosrc.Exists(ctx, dburi)
+	if err != nil {
+		return nil, err
+	}
+	if exists {
+		return open(ctx, dburi)
+	}
+
+	// If the dbfile doesn't exist, we check if we need to migrate the older
```
nwt, line 43:
```suggestion
	// Since the dbfile doesn't exist, we check if we need to migrate the older
```

ppl/zqd/db/flags.go
```diff
@@ -1,13 +1,10 @@
-package apiserver
+package db
```
nwt, line 1: Nit: I think the methods in here should live next to their associated type declarations in `db.go`.

nwt (approved):

ppl/zqd/db/flags.go
mattnibs: I hear you, this is going to change further with helm-postgres, so I'm going to keep this as-is right now.

PR #1938, opened 2021-01-06
Init apiserver.Manager only when needed

### LINKED ISSUE #1935: zqd: Recruiter/Workers should not initialize apiserver.Manager
As observed while @marktwallace was testing helm postgres draft pr #1917, recruiter and worker containers fail to initialize if they do not have access to the postgres database. They should not have access to the postgres database.

Have zqd instances with the recruiter and worker personalities not initialize the space manager which needs access to postgres (if db.kind is set to postgres).

mattnibs:
No reason to init apiserver.Manager when the personalities worker and recruiter
have no use for them. Only init apiserver.Manager for personalities that serve
the api server routes.

Closes #1935

### DIFF
--- a/ppl/zqd/core.go
+++ b/ppl/zqd/core.go
@@ -217,7 +217,9 @@ func (c *Core) Root() iosrc.URI {
 }
 
 func (c *Core) Shutdown() {
-	c.mgr.Shutdown()
+	if c.mgr != nil {
+		c.mgr.Shutdown()
+	}
 }
 
 func (c *Core) nextTaskID() int64 {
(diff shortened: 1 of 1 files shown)


ppl/zqd/core.go
```diff
@@ -147,7 +140,14 @@ func NewCore(ctx context.Context, conf Config) (*Core, error) {
...
 }
 
-func (c *Core) addAPIServerRoutes() {
+func (c *Core) addAPIServerRoutes(ctx context.Context, conf Config) (err error) {
+	c.root, err = iosrc.ParseURI(conf.Root)
+	if err != nil {
+		return
+	}
+	if c.mgr, err = apiserver.NewManager(ctx, conf.Logger, c.registry, c.root, conf.DB); err != nil {
+		return
```
nwt, line 149: Nit: I think adorned returns improve readability.

nwt (approved):

ppl/zqd/core.go
mattnibs: I can get behind this.

PR #1945, opened 2021-01-07
add hidden zapi auth subcommand

alfred-landrum:
This adds a few auth related commands for zapi, hidden for now since the authentication is not wired for any other commands than these, including:
* auth login: implements the 'Device Authorization' flow using the Auth0 information reported back from the `/auth/method` api call, launching a browser automatically (by default) for login, and saves the tokens for the zqd endppoint in a local file for later 
usage.
* auth logout: removes the credentials stored for the zqd endpoint
* auth verify: hits the `/auth/identity` api to verify the stored token
* auth method: reports the authentication information sent back from the server

There's still more work to come, including wiring auth for other operations, handling token refreshing, testing, etc. , but I thought I'd put this up now as-is since its a useful milestone.

Here's a video showing it all in action against a test Auth0 tenant:
https://user-images.githubusercontent.com/704945/103917728-c988df00-50c2-11eb-8c28-5ad4faae07c0.mov

### DIFF
--- a/cmd/zapi/cmd/auth/command.go
+++ b/cmd/zapi/cmd/auth/command.go
@@ -1,0 +1,40 @@
+package auth
+
+import (
+	"flag"
+
+	"github.com/brimsec/zq/cmd/zapi/cmd"
+	"github.com/mccanne/charm"
+)
+
+var Auth = &charm.Spec{
+	Name:  "auth",
+	Usage: "auth [subcommand]",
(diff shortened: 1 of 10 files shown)


mattnibs (approved):
This is great! In future revs would be great to have the other commands by default use the endpoint set in credentials. Also use the zapi UserConfigDir to store things like default space for a host.

PR #1948, opened 2021-01-07
Postgres helm

### LINKED ISSUE #1718: Add postgres 11 instance to helm chart
To support Temporal, we'll need a Postgresql instance. For permanent cloud deployments, we'll use a database service, most likely aws aurora. For local deployments, the helm chart should have a postgres instance. Let's use postgres 11, as that's the latest major version available in aws aurora.

mattnibs:
Add helm postgres app and configure the zqd root helm chart to work with it.

Related changes:

- Add pgtcl command line utility to the root docker container.

- Add pgctl migrate command and include it as an init container for the zqd
  root deployment. This will ensure migrations are up-to-date on new root
  upgrades.

- Add k8s/postgres-secret.sh to generate random passwords for postgres.

Closes #1718


charts/postgres/Chart.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: v2
+name: postgres
+description: A Helm chart for postgres
+
+# A chart can be either an 'application' or a 'library' chart.
```
alfred-landrum, line 5: Would you please remove the boilerplate comments in this file?

ppl/zqd/db/db.go
```diff
@@ -23,6 +24,24 @@ type DB interface {
 
 type DBKind string
 
+// Implement the flag.Value interface for type DBKind
```
alfred-landrum, line 27: I suggest either removing the comment, or moving & adjusting it for godoc formatting (`Set implements ...`).

ppl/cmd/pgctl/migrate/command.go
```diff
@@ -0,0 +1,79 @@
...
+	if c.migrations == "" {
+		return errors.New("argument '-m' (migrations directory) is required")
+	}
+
+	u, err := iosrc.ParseURI(c.migrations)
+	if err != nil {
+		return err
+	}
+
+	m, err := migrate.New(u.String(), c.postgres.String())
```
marktwallace, line 54: This is a high-level question. I'm including it here because this code, which all looks good, is the context for the question. 

Why do we have our own CLI for migrations (that uses the "New" call) rather that using the CLI implemented by the golang-migrate project?

E.g. migrate -source file://path/to/migrations -database postgres://localhost:5432/database up 2

Dockerfile
```diff
@@ -20,8 +20,11 @@ COPY . .
...
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags="${LDFLAGS}" -a -installsuffix cgo -o /go/bin/zqd ./ppl/cmd/zqd
 
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags="${LDFLAGS}" -a -installsuffix cgo -o /go/bin/pgctl ./ppl/cmd/pgctl
 
 FROM scratch
 WORKDIR /app
 COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
 COPY --from=build /go/bin/zqd /app
+COPY --from=build /go/bin/pgctl /app
+COPY --from=build /build/ppl/zqd/db/postgresdb/migrations /postgres/migrations
```
marktwallace, line 30: These changes to the Docker build are OK, but in the future we should not include the migration information except for in the containers that require it. So, for example, the containers for the workers and recruiter which do not access postgres would not have the migration code included. Doing that is more complicated, so I think this code is best for now.

charts/postgres/Chart.yaml
```diff
@@ -0,0 +1,29 @@
...
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+appVersion: 0.1.0
+
+dependencies:
+  - name: postgresql
+    repository: https://charts.bitnami.com/bitnami
+    version: 10.x.x
+
```
marktwallace, line 29: BTW, I will probably end up absorbing this chart as a subchart in the "umbrella" chart for all services that I am working on in a branch.

charts/zqd/templates/deployment.yaml
```diff
@@ -105,6 +113,30 @@ spec:
               name: spaces-volume
             - mountPath: /tmp
               name: tmp-volume                   
+            {{- if eq .Values.personality "root" }}
+            - mountPath: /creds/postgres
+              name: postgres-creds-volume
+            {{- end }}
+
+      {{- if eq .Values.personality "root" }}
```
marktwallace, line 121: How do we handle the ordering for which of these containers does its work "first"? Does it matter if the migration happens before, after, or during startup of the zqd root process?

marktwallace (approved):
This is a great start on Postgres support!

I had some comment and questions, but nothing that would prevent this from being merged as long as the distributed tests pass consistently.

The helm chart work I am doing will touch a lot of these same files, and I can build on this to include postgres as a subchart in a "master" deploy script.

Dockerfile
mattnibs: Agreed.

ppl/cmd/pgctl/migrate/command.go
mattnibs: We could do that, I was just thinking that since we already need pgctl to create/destroy test database instances it wouldn't be that difficult to just make another command vs having to wire up another binary into our build system. Having our commands does have some niceties- for instance we can use the same cli options for specifying the postgres connection details that we use in zqd listen. We might also want to include more custom logging, etc.

charts/zqd/templates/deployment.yaml
mattnibs: The migration is an initContainer of the root process (i.e. "initdb"), so the root process won't start until after the migration has successfully exited.

PR #1951, opened 2021-01-08
expr.LookupValueCaster(): change argument type and rename 

henridf:
This commit changes `expr.LookupValueCaster()` to take a zng.Type rather than the string representation of the type. This is needed so that the forthcoming `cast()` function (now mostly working on branch shaping-function-cast) can call with the various leaf zng.Type's found during record traversal.

Also, while I was making this change, I realized ValueCaster is misleading and changed the name to PrimitiveCaster.

### DIFF
--- a/expr/eval.go
+++ b/expr/eval.go
@@ -708,21 +708,22 @@ func (c *Call) Eval(rec *zng.Record) (zng.Value, error) {
 	return c.fn.Call(c.args)
 }
 
-func NewCast(expr Evaluator, typ string) (Evaluator, error) {
+func NewCast(expr Evaluator, styp string) (Evaluator, error) {
 	// XXX should handle alias casts... need type context.
 	// compile is going to need a local type context to create literals
 	// of complex types?
-	vc := LookupValueCaster(typ)
-	if vc == nil {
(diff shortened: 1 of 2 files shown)


expr/eval.go
```diff
@@ -708,21 +708,22 @@ func (c *Call) Eval(rec *zng.Record) (zng.Value, error) {
...
 }
 
-func NewCast(expr Evaluator, typ string) (Evaluator, error) {
+func NewCast(expr Evaluator, styp string) (Evaluator, error) {
 	// XXX should handle alias casts... need type context.
 	// compile is going to need a local type context to create literals
 	// of complex types?
-	vc := LookupValueCaster(typ)
-	if vc == nil {
+	typ := zng.LookupPrimitive(styp)
```
mattnibs, line 715: Should this do a check to that styp is not a container and return a separate error for this case? I imagine we won't be casting arrays, records, set here?

henridf: If `styp` is a container, we'll end up three lines below with `fmt.Errorf("cast to %s not implemeneted", styp)`, which seems like the outcome we want.

mattnibs (approved):

PR #1961, opened 2021-01-12
add new command zident for user operations in Auth0

alfred-landrum:
This adds a new command, zident, for user and tenant administration.
This initial checkin supports user operations, including creating a new
user, triggering a password reset for a user, and searching for users
with an email, user id, or tenant id.

### DIFF
--- a/ppl/cmd/zident/user/new.go
+++ b/ppl/cmd/zident/user/new.go
@@ -1,0 +1,91 @@
+package user
+
+import (
+	"context"
+	"errors"
+	"flag"
+	"fmt"
+
+	"github.com/brimsec/zq/ppl/cmd/zident/root"
+	"github.com/mccanne/charm"
+	"gopkg.in/auth0.v5"
+	"gopkg.in/auth0.v5/management"
+)
+
+var NewUser = &charm.Spec{
+	Name:  "new",
+	Usage: "zident user new",
+	Short: "create new user",
+	Long: `
+Creates a new user in Auth0.
+
+See 'zident help user' for required authentication.
+`,
+	New: NewNewCommand,
+}
+
+type NewCommand struct {
+	*root.Command
+	a0cfg         auth0ClientConfig
+	email         string
+	emailVerified bool
+	tenantID      string
+}
+
+func NewNewCommand(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
+	c := &NewCommand{Command: parent.(*Command).Command}
+	f.StringVar(&c.email, "email", "", "email address of new user")
+	f.StringVar(&c.tenantID, "tenantid", "", "tenant_id to set instead of creating one")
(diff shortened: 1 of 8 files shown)


ppl/cmd/zident/user/new.go
```diff
@@ -0,0 +1,91 @@
...
+		}
+		tenantID = c.tenantID
+	} else {
+		tenantID = newTenantID()
+	}
+
+	userID := newUserID()
+
+	m, err := management.New(c.a0cfg.domain.String(),
+		management.WithClientCredentials(c.a0cfg.clientId, c.a0cfg.clientSecret))
```
marktwallace, line 63: This is our vendor secret from Auth0, right? How does c.a0cfg, which contains the secrets, become available to this command?

ppl/cmd/zident/user/user.go
```diff
@@ -0,0 +1,232 @@
...
+		err = multierr.Append(err, errors.New("AUTH0_CLIENT_ID not set"))
+	}
+	if c.clientSecret = os.Getenv("AUTH0_CLIENT_SECRET"); c.clientSecret == "" {
+		err = multierr.Append(err, errors.New("AUTH0_CLIENT_SECRET not set"))
+	}
+	if c.connection = os.Getenv("AUTH0_CONNECTION"); c.connection == "" {
+		err = multierr.Append(err, errors.New("AUTH0_CONNECTION not set"))
+	}
+	domain := os.Getenv("AUTH0_DOMAIN")
+	if domain == "" {
```
marktwallace, line 43: I assume that Auth0 recommends the use of these env vars, right? So we will include the env vars in the deployment.yaml with references to secrets. This is all good and expected, but now our distributed configuration will include a mix of env vars and command line flags.

ppl/cmd/zident/user/user.go
```diff
@@ -0,0 +1,232 @@
...
+	}); err != nil {
+		return nil, err
+	}
+	if user == nil {
+		return nil, errors.New("no user found")
+	}
+	return user, nil
+}
+
+func streamUsers(ctx context.Context, m *management.Management, f searchFlags, allowMultiple bool, fn func(*management.User) error) error {
```
marktwallace, line 85: "management" is kind of a vague name for the main Auth0 interface. Now that I see it, I will expect it in this context, so NP.

marktwallace (approved):
This all looks straightforward and right. One piece of context I am missing is whether or not Auth0 has CLI tools that do similar things. This functionality is so generic that I would not be surprised if Auth0 has CLI in addition to the Golang API you are using for these basic user management operations.

mattnibs (approved):
I imagine at some point we'll want to move some of this functionality (e.g. the user lookup methods) into its own package that can be accessed by both this command and zqd alike?

PR #1964, opened 2021-01-12
Include ZSON in zng-output-check

philrz:
Now that ZSON is one of the zq-sample-data outputs (https://github.com/brimsec/zq-sample-data/pull/18), this PR ensures the script invoked by `make zng-output-check` (which runs in CI) covers it.


marktwallace (approved):

philrz:
As discussed at sync, I'm going to hold this one open pending a talk with @mccanne regarding whitespace changes that he expects are going to change the ZSON output slightly.

philrz:
The merge of https://github.com/brimsec/zq-sample-data/pull/19 caught us up such that this output check now runs clean. I cleared with @mccanne that he's also cool with us starting to run this automation now & going forward so we'll know if we unintentionally change the ZSON output again.

As a final change, I renamed the Makefile target and automation reference from `zng-output-check` to `z-output-check` since, strictly speaking, it now covers all our output permutations in the "Z data model" and not just ZNG.

PR #1972, opened 2021-01-14
zqd listen: Simplify logging

mattnibs:
Get rid of waterfall, tee logic and use simple single file based logging.
Change default logging to output all named loggers into stderr (lvl info and
above).


mattnibs:
Here's what the std output looks like:

```
{"level":"info","ts":1610644000.2016902,"msg":"Starting","datadir":".","open_files_limit":10240,"personality":"all","suricata_supported":false,"zeek_supported":false}
{"level":"info","ts":1610644000.2024791,"logger":"httpd","msg":"Listening","addr":"[::]:9867"}
{"level":"info","ts":1610644070.0561721,"logger":"http.access","msg":"Request completed","request_id":"1","host":"localhost:9867","method":"GET","proto":"HTTP/1.1","remote_addr":"[::1]:50848","request_content_length":0,"url":"/space","elapsed":0.000352834,"response_content_length":3,"status_code":200}
```

mattnibs:
Note this will need an accompanying pr brim side before merge.

marktwallace (approved):
I really like this simplification of logging. For the distributed testing, there are many container logs to watch, so keeping it simple and just using stderr is great.

alfred-landrum:
There's a `Example configfile` comment above `loadConfigFile` that needs to be updated/removed. 

The logging config still takes a `name` parameter; do you want to remove the 'name' portion of logging intact? When I try to use it with this branch, it's not practical, as you'd have to include every named logger in the system.

mattnibs:
> The logging config still takes a name parameter; do you want to remove the 'name' portion of logging intact? When I try to use it with this branch, it's not practical, as you'd have to include every named logger in the system.

Yeah its not practical. With the waterfall logger gone this should go away as well.

alfred-landrum (approved):

Issue #1971, opened 2021-01-14
zeek events only appear at pcap ingest completion for archivestore


alfred-landrum:
While running some experiments with the filestore readonly value set ( from https://github.com/brimsec/zq/pull/1952 ), I noticed that when I dragged in a large pcap, the suricata events would appear in Brim gradually as the ingest progressed, but the zeek events didn't appear until the very end.

I've narrowed it down to the usage of `zbuf.MergeReadersByTsAsReader` in archivepcap.go; if I replace that with a simple combiner that doesn't try to perform any ordering, I see both the zeek & suricata events appear as expected during ingest. We don't actually need any ts ordering at that part of ingest, as the archive import code handles it already.

PR #1973, opened 2021-01-14
remove redundant ordered merging during archive pcap ingest

alfred-landrum:
We sort data during the Write calls to archive.Writer, so an ordered
merge beforehand is unnecessary. Additionally, since the underlying
streams of records from the zeek and suricata analyzers are not
necessarily in order, it can cause records to not appear progressively
during pcap ingest (#1971).


mattnibs (approved):
I can't believe we were doing this.

mattnibs:
At some point we should add a tests that verifies the ability to query during ingest for archivestore.

PR #1976, opened 2021-01-15
Master chart for helm install with Postgres

marktwallace:
This reworks all the existing helm charts, converting them to subcharts of an umbrella chart called zservice. It simplifies and streamlines helm install and uninstall. It makes it easier to understand the installation and will allow us to more easily add additional components such as the API Server and Redis.


.github/workflows/cd.yaml
```diff
@@ -62,13 +63,11 @@ jobs:
...
           kubectl config set-context --current --namespace=$NAMESPACE
-          make helm-uninstall
+          helm uninstall z || :
           k8s/postgres-secret.sh
-          make helm-install-postgres PG_PERSIST=false
-          kubectl wait --for=condition=Ready --timeout=60s pods/postgres-postgresql-0
-          make helm-install-recruiter
-          make helm-install-root
-          make helm-install-worker
+          make helm-install PG_PERSIST=false
```
mattnibs, line 68: Works great!

charts/zservice/charts/worker/templates/deployment.yaml
```diff
@@ -0,0 +1 @@
+{{ include "zservice.deployment" . }}
```
mattnibs, line 1:
```suggestion
{{ include "zservice.deployment" . }}

```

charts/zservice/values.yaml
```diff
@@ -0,0 +1,16 @@
+recruiter:
+  replicaCount: 1
+root:
+  datauri: "file:///spaces"
+  recruiterAddr: z-recruiter:9867
```
mattnibs, line 5: Does the default value here require that the installed app be called "z"?

Makefile
```diff
@@ -140,42 +140,13 @@ kubectl-config:
...
-	--set image.repository=$(ZQD_ECR_HOST)/ \
-	--set image.tag=zqd:$(ECR_VERSION) \
-	--set personality=recruiter \
-	--set useCredSecret=false
-
-helm-install-root:
-	helm install root charts/zqd \
-	--set AWSRegion=us-east-2 \
+helm-install:
+	helm install z charts/zservice \
```
mattnibs, line 144: I'd prefer this name of the app be either "zsrv" or "zservice", but not a hill I willing to die on.

ppl/zqd/core.go
```diff
@@ -127,11 +127,11 @@ func NewCore(ctx context.Context, conf Config) (*Core, error) {
 
 	var startFields []zap.Field
 	switch personality {
-	case "all", "apiserver":
+	case "all", "apiserver", "root":
 		if err := c.addAPIServerRoutes(ctx, conf); err != nil {
 			return nil, err
 		}
-		if personality == "all" {
+		if personality == "all" || personality == "root" {
```
mattnibs, line 134: Should we drop the use of personality "all" then?

marktwallace: I think the personality "root" will be temporary, and we will remove it after we start using personality "apiserver" in our deploys. I think we should keep personality "all".

alfred-landrum (commented):
I pulled this branch, and after making sure I had run the `k8s/postgres-secret.sh` file, was able to run `make helm-install` and successfully interact with the deployed chart.

.github/workflows/cd.yaml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - umbrella-chart
```
mattnibs, line 7: Friendly reminder to drop this before merge

mattnibs (approved):

Issue #1932, opened 2021-01-06
putting to `.` results in empty-named fields


henridf:
This adds a field with value `1` but with an empty name. 

```
$ echo '{ msg: "hello" }' | zq -f zson -i zson "put .=1" -
{
    msg: "hello",
    : 1
} (=0)
```

This should probably be a compile-time error (although if/when we support top-level zng.Values, it might no longer be an error and just output the value `1`).

henridf:
> This should probably be a compile-time error

That said if the RHS is a record, it seems like you'd expect the record's fields to be lifted into the root? For example, on a zeek record, `put .=id` would result in `orig_h, orig_p, resp_h, resp_p` all appearing at the root.

henridf:
> For example, on a zeek record, put .=id would result in orig_h, orig_p, resp_h, resp_p all appearing at the root.

More specifically, they will be the only fields at the root. 
For example, `echo '{"a": 1, "r2": {"b": 1}}' | zq -i zson -t "put .=r2"` should produce `{"b": 1}`.  This is consistent with `put`ing to a non-root record field.

philrz:
Verified in `zq` commit `7d38048c`.

Revisiting the examples cited above, we see that the first one does produce an error as desired.

```
$ echo '{ msg: "hello" }' | zq -f zson -i zson "put .=1" -
put .=x: cannot put a non-record to .
{
    msg: "hello"
} (=0)
```

And for when the RHS is a record, its contents become the only fields at the root, as desired. Using the Zeek [zq-sample-data](https://github.com/brimsec/zq-sample-data):

```
$ zq -f zson 'head 1 | put .=id' conn.log.gz
{
    orig_h: 10.164.94.120,
    orig_p: 39681 (port=(uint16)),
    resp_h: 10.47.3.155,
    resp_p: 3389 (port)
} (=0)
```

Thanks @henridf!

Issue #1783, opened 2020-12-09
Ingest processing design


henridf:
This issue contains a design writeup of what we're doing for ingest processing. 

1. [Background: The existing "json types" system](https://github.com/brimsec/zq/issues/1783#issuecomment-751766323)
2. [Z for ingest processing](https://github.com/brimsec/zq/issues/1783#issuecomment-751766464)
3. [The ZSON shaping functions](https://github.com/brimsec/zq/issues/1783#issuecomment-751790854)
4. [`reshape` for the common case](https://github.com/brimsec/zq/issues/1783#issuecomment-754087671)
4. [`An example with Zeek`](https://github.com/brimsec/zq/issues/1783#issuecomment-754135719)

henridf:
placeholder

henridf:
another placeholder

henridf:
placeholder

henridf:
## `reshape` for the common case


A common use pattern will be to use all four verbs, as in:

`put . = trim(., zeek_conn_t) | put . = fill(., zeek_conn_t) | put . = cast(., zeek_conn_t) | put . = order(., zeek_conn_t)`

or equivalently

`put . = trim(fill(cast(order(., zeek_conn_t))))`.

(it is left as an exercise to the reader to note that all four functions commute with each other).

To avoid this cumbersome notation, we could introduce a `reshape` function that is the composition of `trim`, `fill`, `cast`, and `order`. 

So we would instead write `put . = reshape(., zeek_conn_t)`.

We _may_ also want an open variant that omits the `trim`. `reshape_open(., typ)` would carry out the composition of `fill`, `cast`, and `order`.

henridf:
```
const zeek_id_t = ({orig_h:ip, orig_p:port=(uint16), resp_h:ip, resp_p:port})
const zeek_conn_t = {_path: string, id: zeek_id_t, uid: string, proto: zenum, ...}
const zeek_http_t = {_path: string, id: zeek_id_t, uid: string, method: bstring, ...}

* | switch (
  _path=conn => reshape(zeek_conn_t) 
  _path=http => reshape(zeek_http_t)
  ..
) | ...
```

PR #1984, opened 2021-01-18
Add shaping functions

henridf:
This commit adds the shaping 'primitive' functions cast(), fill(),
crop(), and order() along with fit() (composition of fill and crop)
and shape (composition of all four primitive functions). 

A description of what these functions do is in #1783; the ztests of this commit also provide some illustration.

I will expand the test set after #1932 is done, since a lot of shaping assignments will use that.


expr/shaper.go
```diff
@@ -0,0 +1,418 @@
...
+	"github.com/brimsec/zq/zng/resolver"
+	"github.com/brimsec/zq/zson"
+)
+
+// A ShaperTransform represents one of the different transforms that a
+// shaper can apply.
+type ShaperTransform int
+
+const (
+	Cast = ShaperTransform(iota)
```
nwt, line 17: NIt: Use conventional form.
```suggestion
	Cast ShaperTransform = iota
```

expr/shaper.go
```diff
@@ -0,0 +1,418 @@
...
+	Fill
+	Crop
+	Order
+)
+
+type step int
+
+const (
+	copyPrimitive = step(iota) // copy field fromIndex from input record
```
nwt, line 26: NIt: Use conventional form.
```suggestion
	copyPrimitive step = iota // copy field fromIndex from input record
```

expr/shaper.go
```diff
@@ -0,0 +1,418 @@
...
+	copyContainer
+	castPrimitive // cast field fromIndex from input record
+	null          // write null
+	record        // record into record below us
+)
+
+// A shapeBuilder is a recursive data structure encoding a series of
+// copy/cast steps to be carried out over an input record.
+type shapeBuilder struct {
+	op        step
```
nwt, line 36: I think you should change the type name to `op`.

compiler/expr.go
```diff
@@ -258,6 +258,52 @@ func compileCutter(zctx *resolver.Context, node ast.FunctionCall) (*expr.Cutter,
...
+		return map[expr.ShaperTransform]struct{}{expr.Crop: struct{}{}}
+	case "order":
+		return map[expr.ShaperTransform]struct{}{expr.Order: struct{}{}}
+	case "shape":
+		return map[expr.ShaperTransform]struct{}{expr.Crop: struct{}{}, expr.Fill: struct{}{}, expr.Order: struct{}{}, expr.Cast: struct{}{}}
+	default:
+		return nil
+	}
+
+}
```
nwt, line 279: Maps feel heavy here. Maybe use flags?
```go
func shaperOps(name string) ShaperTransform {
	switch name {
	case "cast":
		return expr.Cast
	case "fill":
		return expr.Fill
	case "fit":
		return expr.Crop | expr.Fill
	case "crop":
		return expr.Crop
	case "order":
		return expr.Order
	case "shape":
		return expr.Crop | expr.Fill | expr.Orde | expr.Cast
	default:
		return 0
	}
}
```

compiler/expr.go
```diff
@@ -258,6 +258,52 @@ func compileCutter(zctx *resolver.Context, node ast.FunctionCall) (*expr.Cutter,
...
+
+func isShaperFunc(name string) bool {
+	switch name {
+	case "cast", "fill", "fit", "crop", "order", "shape":
+		return true
+	}
+	return false
```
nwt, line 286:
```suggestion
	return shaperOps(name) != nil
```

zng/type.go
```diff
@@ -337,6 +337,17 @@ func IsUnionType(typ Type) bool {
...
 
+func IsRecordType(typ Type) bool {
+	switch typ := typ.(type) {
+	case *TypeAlias:
+		return IsRecordType(typ.Type)
+	case *TypeRecord:
+		return true
+	default:
+		return false
+	}
```
nwt, line 348:
```suggestion
	_, ok := AliasedType(typ).(*TypeRecord)
	return ok
```

compiler/expr.go
henridf: Yes!

zng/type.go
henridf: 👍

compiler/expr.go
henridf: 👍

expr/shaper.go
henridf: done in 4cd9d91

nwt (approved):

PR #1986, opened 2021-01-18
Notify slack when master workflow fails

### LINKED ISSUE #1963: Report master ci failures to slack #test channel
In attempt to increase visibility for failures on master, send a notification to the #test slack channel should a test failure on master occur.

mattnibs:
Add 'Notify master failure' github action which is triggered whenever
a workflow_run on branch master is completed. If the workflow failed the action
will notify the BrimHQ slack of the failure.

Closes #1963

### DIFF
--- a/.github/workflows/notify-master-failure.yml
+++ b/.github/workflows/notify-master-failure.yml
@@ -14,7 +14,7 @@ jobs:
     if: ${{ github.event.workflow_run.conclusion == 'failure' }}
     runs-on: ubuntu-latest
     steps:
-      - name: Inform Slack users at Brim HQ of the new artifact
+      - name: Notify Brim HQ of failure on master
         uses: tiloio/slack-webhook-action@v1.1.2
         with:
           slack_web_hook_url: ${{ secrets.SLACK_WEBHOOK_BRIMLABS_TEST }}
--- a/.github/workflows/notify-master-failure.yml
+++ b/.github/workflows/notify-master-failure.yml
@@ -1,0 +1,25 @@
+name: 'Notify master failure'
+
+on:
+  workflow_run:
+    branches: 
+      - master
+    workflows:
+      - '**'
+    types:
+      - completed
+
+jobs:
+  slackNotify:
+    if: ${{ github.event.workflow_run.conclusion == 'failure' }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Inform Slack users at Brim HQ of the new artifact
+        uses: tiloio/slack-webhook-action@v1.1.2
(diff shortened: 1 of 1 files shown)


marktwallace:
The changes to the workflow look easier than I expected. I saw Slack under installed GitHub apps, but I do not see how to configure it. Is it admin only?

mattnibs:
Here's what the slack message will look like:

image

We cannot test this on this repository: webhook initiated events only trigger when they workflow is in the default branch. I did create a test repository that uses this exact yaml and it works: https://github.com/mattnibs/githubactions/blob/main/.github/workflows/workflow.yml

mattnibs:
> The changes to the workflow look easier than I expected. I saw Slack under installed GitHub apps, but I do not see how to configure it. Is it admin only?

@marktwallace the only configuration is the `SLACK_WEBHOOK_BRIMLABS_TEST` secret that @philrz was kind enough to setup for me. It contains a slack uri that posts messages to channel #test

.github/workflows/notify-master-failure.yml
```diff
@@ -0,0 +1,25 @@
...
+      - '**'
+    types:
+      - completed
+
+jobs:
+  slackNotify:
+    if: ${{ github.event.workflow_run.conclusion == 'failure' }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Inform Slack users at Brim HQ of the new artifact
```
philrz, line 17: Looks like copypasta. 😉

philrz (changes requested):

.github/workflows/notify-master-failure.yml
```diff
@@ -0,0 +1,25 @@
...
+        uses: tiloio/slack-webhook-action@v1.1.2
+        with:
+          slack_web_hook_url: ${{ secrets.SLACK_WEBHOOK_BRIMLABS_TEST }}
+          slack_json: |
+            {
+              "username": "github-actions",
+              "text": "Workflow \"${{ github.event.workflow_run.name }}\" failed on master.\n${{ github.event.workflow_run.html_url }}"
```
philrz, line 24: Does `html_url` point to the failed run? The docs at https://docs.github.com/en/developers/webhooks-and-events/webhook-events-and-payloads#workflow_run don't seem to describe this field.

.github/workflows/notify-master-failure.yml
mattnibs, line 17:
```suggestion
      - name: Notify Brim HQ of failure on master
```

philrz (approved):

Issue #1997, opened 2021-01-20
"Worker timed out" and restarts during repeated queries


philrz:
Repro is while accessing the cluster via `z-*` services that are currently showing commit `7d3f886c` via the `/version` endpoint. I'm using `zapi` commit `b928e539` on my local laptop and tunneling to access the cluster.

To repro, I post a ~200 MB data set and repeatedly execute a `count()` on it.

```
export ZQD_DATA_URI="s3://zq-ci-test/21-01-20.rug1"
export SPACE="corelight"
zapi rm $SPACE
zapi new -k archivestore -d $ZQD_DATA_URI $SPACE
zapi -s $SPACE postpath $(for file in $(aws s3 ls --recursive s3://brim-sampledata/corelight/zeek-logs/ | grep "\.log" | awk '{ print $4 }'); do   echo s3://brim-sampledata/$file; done)
while true; do   zapi -s $SPACE get -workers 4 -t 'count()';   sleep 1; done
```

As we can see in the output, sometimes the query succeeds, and sometimes we see a failure.

```
$ while true; do   zapi -s $SPACE get -workers 4 -t 'count()';   sleep 1; done
#0:record[count:uint64]
0:[1031683;]
#0:record[count:uint64]
0:[1031683;]
#0:record[count:uint64]
0:[1031683;]
Post "http://192.168.70.186:9867/worker/chunksearch?format=zng": dial tcp 192.168.70.186:9867: connect: connection refused
#0:record[count:uint64]
0:[1031683;]
#0:record[count:uint64]
0:[1031683;]
#0:record[count:uint64]
0:[1031683;]
#0:record[count:uint64]
0:[1031683;]
#0:record[count:uint64]
0:[1031683;]
#0:record[count:uint64]
0:[1031683;]
#0:record[count:uint64]
0:[1031683;]
#0:record[count:uint64]
0:[1031683;]
#0:record[count:uint64]
0:[1031683;]
#0:record[count:uint64]
0:[1031683;]
Post "http://192.168.28.60:9867/worker/chunksearch?format=zng": dial tcp 192.168.28.60:9867: connect: connection refused
#0:record[count:uint64]
0:[1031683;]
#0:record[count:uint64]
0:[1031683;]
#0:record[count:uint64]
0:[1031683;]
#0:record[count:uint64]
0:[1031683;]
#0:record[count:uint64]
0:[1031683;]
#0:record[count:uint64]
0:[1031683;]
#0:record[count:uint64]
[...]

PR #1999, opened 2021-01-22
increase worker.idletime

marktwallace:
See issue https://github.com/brimsec/zq/issues/1997 for an explanation of why this default value needs to be changed.


mattnibs (approved):

alfred-landrum (changes requested):
I'd like to continue discussion in the linked issue before merging.

alfred-landrum (approved):
As noted in the related issue, there's more we'll end up doing to help with this, but this will help workaround long waits for workers on the source channel, which is ultimately waiting on S3 directory operations and metadata reads.

PR #2003, opened 2021-01-22
Explicitly declare FileKind enum type

mattnibs:
The pattern where the go compiler infers the type for all members of a const
declaration group from the first declaration's type only works when using iota.
Explictly declare the type for all FileKind enums.


alfred-landrum (commented):
My bad :| , thanks for fixing!

mattnibs:
@alfred-landrum I think you forgot to hit approve!

alfred-landrum (approved):

PR #2004, opened 2021-01-22
Update deploy job to use ZTest

marktwallace:
This adds a new set of ZTests tagged with "cluster" to run the same tests that were previously run directly with zapi in cd.yaml.

A Makefile target, test-cluster, that runs the cluster tests has been introduced. The cluster tests can be more easily extended and maintained as ZTests.

In addition, it addresses two existing intermittent failures in the Deploy test in cd.yaml.
(1) On some test failures, the temporary namespace (e.g. ci-408) was not being removed after the test. This PR consolidates three previous jobs: pre-clean, eks-test-deploy, and eks-test-zapi-1, into a single job that shares the same cleanup code that removes the temporary namespace.
(2) There was a "sleep 2" after the port-forward commands. It is replaced by two loops that verify the port-forward with netcat before proceeding.


Makefile
```diff
@@ -100,6 +100,12 @@ test-postgres-docker:
 		docker-compose -f $(CURDIR)/ppl/zqd/scripts/dkc-services.yaml down || exit; \
 		exit $$status
 
+.PHONY: test-cluster
+test-cluster: build 
+	@ZTEST_PATH="$(CURDIR)/dist:$(CURDIR)/bin" \
+		ZTEST_TAG=cluster \
+        go test -v -run TestZq/ppl/zqd/ztests/cluster .
```
nwt, line 107:
```suggestion
		go test -v -run TestZq/ppl/zqd/ztests/cluster .
```

.github/workflows/cd.yaml
```diff
@@ -57,58 +43,57 @@ jobs:
...
+              break
+            fi
+            sleep 1
+          done
+          for i in {1..5} ; do
+            if nc -z localhost 9867 </dev/null ; then
+              break
+            fi
+            sleep 1
+          done
```
mattnibs, line 83: Can you use k8s/zqd-port.sh here? Looks pretty similar.

ppl/zqd/ztests/cluster/count.yaml
```diff
@@ -0,0 +1,9 @@
+tag: cluster
+script: |
+  # Test assumes a remote zqd instance with space "files" is available at 9867
+  time zapi -s files get -workers 4 -t 'count()' 2>&1
+outputs:
```
nwt, line 5: Nit: Put a blank line between top level keys like we do in all the other ztests.
```suggestion
tag: cluster

script: |
  # Test assumes a remote zqd instance with space "files" is available at 9867
  time zapi -s files get -workers 4 -t 'count()' 2>&1

outputs:
```

ppl/zqd/ztests/cluster/countbytxhost.yaml
```diff
@@ -0,0 +1,13 @@
+tag: cluster
+script: |
+  # Test assumes a remote zqd instance with space "files" is available at 9867
+  zapi -s files get -workers 4 -t '* | count() by tx_hosts | sort -r count | head 5' 2>&1
+outputs:
```
nwt, line 5:
```suggestion
tag: cluster

script: |
  # Test assumes a remote zqd instance with space "files" is available at 9867
  zapi -s files get -workers 4 -t 'count() by tx_hosts | sort -r count | head 5' 2>&1

outputs:
```

ppl/zqd/ztests/cluster/countbytxhost.yaml
```diff
@@ -0,0 +1,13 @@
+tag: cluster
+script: |
+  # Test assumes a remote zqd instance with space "files" is available at 9867
+  zapi -s files get -workers 4 -t '* | count() by tx_hosts | sort -r count | head 5' 2>&1
```
nwt, line 4:
```suggestion
  zapi -s files get -workers 4 -t 'count() by tx_hosts | sort -r count | head 5' 2>&1
```

ppl/zqd/ztests/cluster/count.yaml
nwt: (And do this for the other ztests you're adding here too.)

.github/workflows/cd.yaml
```diff
@@ -57,58 +43,57 @@ jobs:
...
+          #
+          # Scale cluster for tests and set up port-forwards,
+          # then wait for the forwarded ports to be available.
+          #
           kubectl scale --replicas=4 deployment/zsrv-worker
           kubectl port-forward svc/zsrv-recruiter 8020:9867 &
           kubectl port-forward svc/zsrv-root 9867:9867 &
-          sleep 2 # wait for port-forwards to complete
+          for i in {1..5} ; do
+            if nc -z localhost 8020 </dev/null ; then
```
nwt, line 73: You don't need
```suggestion
            if nc -z localhost 8020; then
```

.github/workflows/cd.yaml
```diff
@@ -57,58 +43,57 @@ jobs:
...
+          kubectl wait --for=condition=available --timeout=120s deployment/zsrv-worker
+          #
+          # Scale cluster for tests and set up port-forwards,
+          # then wait for the forwarded ports to be available.
+          #
           kubectl scale --replicas=4 deployment/zsrv-worker
           kubectl port-forward svc/zsrv-recruiter 8020:9867 &
           kubectl port-forward svc/zsrv-root 9867:9867 &
-          sleep 2 # wait for port-forwards to complete
+          for i in {1..5} ; do
```
nwt, line 72: Nit: We usually don't include a space before a semicolon in sh scripts.
```suggestion
          for i in {1..5}; do
```

.github/workflows/cd.yaml
```diff
@@ -57,58 +43,57 @@ jobs:
       - run: |
           set -x
           NAMESPACE=ci-${{ github.run_number }}
+          SPACE=files
```
nwt, line 46: Nit: Using all capitals for these makes them look like environment variables, so maybe use lowercase.

.github/workflows/cd.yaml
```diff
@@ -57,58 +43,57 @@ jobs:
...
           export ZQD_DATA_URI=s3://zq-ci-test/$(date +%y-%m-%d).${{ github.run_number }}
           export ZQD_ECR_HOST=${{ secrets.AWS_ECR_HOST }}
+          make install
+          mkdir testlog
```
nwt, line 50: Move this to the step where it's actually used.

.github/workflows/cd.yaml
```diff
@@ -57,58 +43,57 @@ jobs:
...
           NAMESPACE=ci-${{ github.run_number }}
+          SPACE=files
           export ZQD_DATA_URI=s3://zq-ci-test/$(date +%y-%m-%d).${{ github.run_number }}
           export ZQD_ECR_HOST=${{ secrets.AWS_ECR_HOST }}
+          make install
+          mkdir testlog         
+          aws s3 rm --recursive $ZQD_DATA_URI || : # clear data from previous run
           aws eks --region us-east-2 update-kubeconfig --name zq-test
+          #
+          # Create temporary namespace for tests and deploy zsvc
```
nwt, line 54:
```suggestion
          # Create temporary namespace for tests and deploy zsrv
```

.github/workflows/cd.yaml
```diff
@@ -57,58 +43,57 @@ jobs:
...
+            fi
+            sleep 1
+          done
+          #
+          # Create archive for tests
+          #
           time zapi ls
           time zapi new -k archivestore -d $ZQD_DATA_URI $SPACE
           # Post an archive of about 1.6 GB to run test queries.
           time zapi -s $SPACE postpath s3://brim-sampledata/wrccdc/zeek-logs/files.log.gz
```
mattnibs, line 90: I would move these to be commands to `make test-cluster`

.github/workflows/cd.yaml
marktwallace: Good idea, I will break that part out.

mattnibs (approved):

PR #2015, opened 2021-01-26
initial multitenancy support

alfred-landrum:
This changes adds initial multitenancy support to zqd.

The high level summary is that when requests are issued to zqd, those requests will have an associated identity, either specified by an authenticated JWT, or the "anonymous" identity. When a space is created, the tenant id of the creating identity is stored with it. Only users with the same tenant id will be allowed to interact with the space. When a list of spaces is requested, the tenant id is used as a filter in the returned spaces.

The token validation and context interaction parts of auth moved into a separate package from package zqd & api. I dropped the dependency on the auth0 jwt middleware, as it was just a thin wrapper around the go-jwt 'jwt.Parse' call. The validation code is better tested here with the new Go tests, and I wrote those guided by looking at test coverage results in my IDE.

zapi gets a new flag, `-configpath`, used to set where it looks for its configuration holding directory. It defaults to os.UserConfigDir, which is $HOME/.config for Linux systems. To aid this change, I made sure that the $HOME variable is set during ztests; it's set to the temporary directory where the ztest is run.

With this change, it's always possible to call `auth.IdentityFromContext`. When zqd is run without authentication enabled, it will always return an identity with the anonymous user and tenant values.

I added migrations to the file and postgres database implementations, adding a column for "tenant_id", which is set to the anonymous tenant value by default.

I added ztests that verify the tenant handling when zqd is backed by either a filedb or postgres. The ztests uses a new utility, "testtoken", to generate a signed JWT using the public/private test key pair in zqd/testdata.

### DIFF
--- a/cmd/zapi/cmd/auth/logout.go
+++ b/cmd/zapi/cmd/auth/logout.go
@@ -5,7 +5,6 @@ import (
 	"flag"
 	"fmt"
 
-	"github.com/brimsec/zq/cmd/zapi/cmd"
 	"github.com/mccanne/charm"
 )
 
@@ -35,18 +34,14 @@ func (c *LogoutCommand) Run(args []string) error {
 		return errors.New("logout command takes no arguments")
 	}
 
-	cpath, err := cmd.UserStdCredentialsPath()
-	if err != nil {
-		return err
-	}
-	svccreds, err := cmd.LoadCredentials(cpath)
+	creds, err := c.LocalConfig.LoadCredentials()
 	if err != nil {
 		return fmt.Errorf("failed to load credentials file: %w", err)
 	}
-	svccreds.RemoveTokens(c.Host)
-	if err := cmd.SaveCredentials(cpath, svccreds); err != nil {
+	creds.RemoveTokens(c.Host)
+	if err := c.LocalConfig.SaveCredentials(creds); err != nil {
 		return fmt.Errorf("failed to save credentials file: %w", err)
 	}
-	fmt.Printf("Logout successful, cleared credentials for %s in %s\n", c.Host, cpath)
+	fmt.Printf("Logout successful, cleared credentials for %s\n", c.Host)
 	return nil
 }
--- a/cmd/zapi/cmd/cli.go
+++ b/cmd/zapi/cmd/cli.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"net"
 	"os"
 	"syscall"
 
@@ -58,18 +59,20 @@ func New(f *flag.FlagSet) (charm.Command, error) {
 	f.Var(&c.spaceID, "id", "<space_id>")
 	f.BoolVar(&c.NoFancy, "nofancy", c.NoFancy, "disable fancy CLI output (true if stdout is not a tty)")
 	c.cli.SetFlags(f)
+	c.LocalConfig.SetFlags(f)
 
 	return c, nil
 }
 
 type Command struct {
-	conn      *client.Connection
-	Host      string
-	Spacename string
-	NoFancy   bool
-	ctx       *signalCtx
-	spaceID   api.SpaceID
-	cli       cli.Flags
+	Host        string
+	LocalConfig localConfigFlags
+	NoFancy     bool
+	Spacename   string
+	cli         cli.Flags
+	conn        *client.Connection
+	ctx         *signalCtx
+	spaceID     api.SpaceID
 }
 
 func (c *Command) Context() context.Context {
@@ -78,9 +81,6 @@ func (c *Command) Context() context.Context {
 
 // Connection returns a central client.Connection instance.
 func (c *Command) Connection() *client.Connection {
-	if c.conn == nil {
-		c.conn = client.NewConnectionTo("http://" + c.Host)
-	}
 	return c.conn
 }
 
@@ -103,6 +103,19 @@ func (c *Command) Cleanup() {
 }
 
 func (c *Command) Init(all ...cli.Initializer) error {
+	if _, _, err := net.SplitHostPort(c.Host); err == nil {
+		c.Host = "http://" + c.Host
+	}
+	c.conn = client.NewConnectionTo(c.Host)
+
+	creds, err := c.LocalConfig.LoadCredentials()
+	if err != nil {
+		return err
+	}
+	if tokens, ok := creds.ServiceTokens(c.Host); ok {
+		c.conn.SetAuthToken(tokens.Access)
+	}
+
 	return c.cli.Init(all...)
 }
 
--- a/cmd/zapi/cmd/cli.go
+++ b/cmd/zapi/cmd/cli.go
@@ -66,7 +66,7 @@ func New(f *flag.FlagSet) (charm.Command, error) {
 
 type Command struct {
 	Host        string
-	LocalConfig localConfigFlags
+	LocalConfig LocalConfigFlags
 	NoFancy     bool
 	Spacename   string
 	cli         cli.Flags
--- a/cmd/zapi/cmd/index/create.go
+++ b/cmd/zapi/cmd/index/create.go
@@ -60,6 +60,10 @@ func (c *CreateCmd) Run(args []string) error {
 	if len(args) == 0 && c.zql == "" {
 		return errors.New("zapi index create: one or more indexing patterns must be specified")
 	}
+	defer c.Cleanup()
+	if err := c.Init(); err != nil {
+		return err
+	}
 	id, err := c.SpaceID()
 	if err != nil {
 		return err
--- a/cmd/zapi/cmd/rename/rename.go
+++ b/cmd/zapi/cmd/rename/rename.go
@@ -12,7 +12,7 @@ import (
 
 var Rename = &charm.Spec{
 	Name:  "rename",
-	Usage: "rename [old_name] [new_name]",
+	Usage: "rename [old_name] new_name",
 	Short: "renames a space",
 	Long:  `Renames a space, given the current space name and a desired new name.`,
 	New: func(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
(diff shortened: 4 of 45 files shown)


cmd/zapi/cmd/auth/logout.go
```diff
@@ -35,18 +34,14 @@ func (c *LogoutCommand) Run(args []string) error {
...
-	}
-	svccreds, err := cmd.LoadCredentials(cpath)
+	creds, err := c.LocalConfig.LoadCredentials()
 	if err != nil {
 		return fmt.Errorf("failed to load credentials file: %w", err)
 	}
-	svccreds.RemoveTokens(c.Host)
-	if err := cmd.SaveCredentials(cpath, svccreds); err != nil {
+	creds.RemoveTokens(c.Host)
+	if err := c.LocalConfig.SaveCredentials(creds); err != nil {
```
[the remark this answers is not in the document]
alfred-landrum: That's right; `zapi auth login` adds token to the stored credentials file, and `zapi auth logout` removes them.

cmd/zapi/cmd/index/create.go
```diff
@@ -60,6 +60,10 @@ func (c *CreateCmd) Run(args []string) error {
 	if len(args) == 0 && c.zql == "" {
 		return errors.New("zapi index create: one or more indexing patterns must be specified")
 	}
+	defer c.Cleanup()
```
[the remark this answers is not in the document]
alfred-landrum: I tried to add the 'Cleanup ... Init' pattern to all of the subcommands, as client connection is now instantied inside the call to Init.

cmd/zapi/cmd/rename/rename.go
```diff
@@ -29,18 +29,34 @@ type Command struct {
...
-	if err != nil {
+	defer c.Cleanup()
+	if err := c.Init(); err != nil {
 		return err
 	}
+
+	var err error
+	var id api.SpaceID
+	var newname string
+	switch len(args) {
```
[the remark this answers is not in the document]
alfred-landrum: zapi supports global options `-s` and `-id` (ie, `zapi -s <space> <subcommand>`) that can be used to specify the target space of a command. So the 1 argument handling allows relying on those options to to specify the space to be renamed.  And it may not be clear due to the size of the PR, but I put this support in so I could write a ztest that verifies the 'UpdateSpaceName' api properly checks permissions.

ppl/zqd/apiserver/manager.go
```diff
@@ -306,16 +311,29 @@ func (m *Manager) rowToSpaceInfo(ctx context.Context, sr schema.SpaceRow) (api.S
...
 
-func (m *Manager) GetSpace(ctx context.Context, id api.SpaceID) (api.SpaceInfo, error) {
+func (m *Manager) getSpacePermCheck(ctx context.Context, id api.SpaceID) (schema.SpaceRow, error) {
 	sr, err := m.db.GetSpace(ctx, id)
+	if err != nil {
+		return schema.SpaceRow{}, err
+	}
+	ident := auth.IdentifyFromContext(ctx)
+	if sr.TenantID != ident.TenantID {
+		return schema.SpaceRow{}, zqe.ErrForbidden()
```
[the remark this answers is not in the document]
alfred-landrum: I don't understand what you mean about the different return paths?

If the `m.db.GetSpace` call fails, it's because of some database fetching error; there's no permissions checking that happens inside the db code. This function `getSpacePermCheck` is a convenience function to add permissions checking around the db GetSpace call, since it's needed in several places in Manager.

zqe/error.go
```diff
@@ -22,6 +22,7 @@ const (
 	Invalid
 	NotFound
 	NoCredentials
+	Forbidden
```
nwt, line 25: Nit: Sort me.

ppl/cmd/testtoken/testtoken.go
```diff
@@ -0,0 +1,65 @@
...
+	"time"
+
+	"github.com/brimsec/zq/ppl/zqd/auth"
+	"github.com/mccanne/charm"
+)
+
+var CLI = &charm.Spec{
+	Name:  "testtoken",
+	Usage: "testtoken",
+	Short: "generate access token to test zqd auth",
```
nwt, line 17: Nit: testtoken reads like it tests tokens rather than creates them. Maybe call this gentoken or maketoken?

nwt (approved):
:shipit:

ppl/zqd/apiserver/manager.go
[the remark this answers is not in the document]
marktwallace: OK, so the information leak is that returning from line 317 indicates that a space cannot be retrieved from the db (which may be because the space does not exist) while returning from line 321 indicates that the space exists and is owned by another tenant. It will leak information in the case that an unauthorized tenant acquires a valid space id that they do not own. Could you reverse the order of the m.db.GetSpace() and auth.IdentifyFromContext() calls and have it still work the way you want?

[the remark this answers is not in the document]
alfred-landrum: > OK, so the information leak is that returning from line 317 indicates that a space cannot be retrieved from the db (which may be because the space does not exist) while returning from line 321 indicates that the space exists and is owned by another tenant. It will leak information in the case that an unauthorized tenant acquires a valid space id that they do not own. Could you reverse the order of the m.db.GetSpace() and auth.IdentifyFromContext() calls and have it still work the way you want?

I don't see how - we don't know the tenant id of the space until we retrieve it from the database.

ppl/zqd/auth/auth.go
```diff
@@ -0,0 +1,43 @@
...
+	"context"
+)
+
+type TenantID string
+type UserID string
+
+const (
+	AnonymousTenantID TenantID = "tenant_000000000000000000000000001"
+	AnonymousUserID   UserID   = "user_000000000000000000000000001"
+)
```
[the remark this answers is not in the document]
alfred-landrum: Yes - inside a zqd that is running without authentication enabled, like the Brim bundled zqd, these are the values (inside an `auth.Identity`) that will be used as the tenant/user for any incoming requests.

ppl/zqd/auth/generator.go
```diff
@@ -0,0 +1,46 @@
+package auth
```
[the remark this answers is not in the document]
alfred-landrum: Before this change, we were using a Go package provided by Auth0, github.com/auth0/go-jwt-middleware. However, that merely provided some convenience functionality to extract the Authorization header, verify the signing algorithm, and add a validated `*jwt.Token` into a context.Context. It did not automatically verify any of the claims in the token, though, which is where the bulk of the line count comes from in auth/validator.go. 

`jwt-go` is a known package for JWT validation, and is what Auth0 was using inside their go-jwt-middleware package until rescently. I see Auth0 is now pointing to the `https://github.com/form3tech-oss/jwt-go`, jwt-go fork, that isn't heavily used. That was primarily to get around an audience checking bug in jwt, but (even before this PR) we were handling that particular bug with our own `verifyAPIAudience` call that followed the Auth0 guidance on audience validation (this url was already a comment in the code):
https://auth0.com/docs/tokens/access-tokens/get-access-tokens#multiple-audiences

cmd/zapi/cmd/cli.go
```diff
@@ -58,18 +59,20 @@ func New(f *flag.FlagSet) (charm.Command, error) {
...
 type Command struct {
-	conn      *client.Connection
-	Host      string
-	Spacename string
-	NoFancy   bool
-	ctx       *signalCtx
-	spaceID   api.SpaceID
-	cli       cli.Flags
+	Host        string
+	LocalConfig localConfigFlags
```
mattnibs, line 69: You shouldn't export a field with a private type.

ppl/zqd/db/postgresdb/ztests/tenant.yaml
```diff
@@ -0,0 +1,81 @@
+tag: postgres
```
mattnibs, line 1: Point of discussion: As an outsider looking in, I find these story tests difficult to grok. When failures occur it is often  hard to find what caused the failure. I would prefer this be broken down into a suite of tests. Perhaps something like:
- auth/ls
- auth/rm
- auth/info
- auth/rename

ppl/zqd/apiserver/manager.go
marktwallace, line 321:
```suggestion
	}
	ident := auth.IdentifyFromContext(ctx)
	if sr.TenantID != ident.TenantID {
		return schema.SpaceRow{}, zqe.ErrForbidden()
	sr, err := m.db.GetSpace(ctx, id)
	if err != nil {
		return schema.SpaceRow{}, err
```

marktwallace: Like this -- always do the auth check first, before there is any possibility of an error return value, because error return values "leak information" and can sometimes be exploited.

alfred-landrum: @marktwallace : the code you suggested will not compile, because the `sr` variable does not exist in your suggested line 316. As I said above, we don't know the tenant id of the space until we retrieve its record from the database.

marktwallace, line 321:
```suggestion
	sr, err := m.db.GetSpace(ctx, id)
	if err != nil {
		return schema.SpaceRow{}, zqe.ErrForbidden()
		// also log error here
	}
	ident := auth.IdentifyFromContext(ctx)
	if sr.TenantID != ident.TenantID {
		return schema.SpaceRow{}, zqe.ErrForbidden()
```

marktwallace: So this is what I mean specifically by the information leak -- the returned error should be "Forbidden" regardless of the reason, unless the auth can be validated. If that is inconvenient for the user, then do not do it, but we should be aware of the possibility of exploit.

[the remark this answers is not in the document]
marktwallace: Yes, that is true. So by returning an error from the db, the code reveals the existence, non-existence, or any other possible db error state prior to knowing whether the tenant issuing the query is the correct tenant. That is why secure code generally uses "unauthorized" responses until authorization had been proven, regardless of the reason.

ppl/cmd/testtoken/testtoken.go
alfred-landrum: 👍 , will use 'gentoken'

ppl/zqd/db/postgresdb/ztests/tenant.yaml
alfred-landrum: @mattnibs : if I split up the tests, there's a lot of boilerplate that would need to be duplicated between the separate test files - the auth zqd configuration, the zapi user setup, the creation of the spaces, and test input files. Since these operations are all to verify the space related apis, is there something else I could do to make the test easier to understand? Let me add some extra comments to one of the files, and see if that helps.

alfred-landrum: Take a look at 252ce57168f7aab60e99b8acbb7265cee3099caa and let me know if that helps.

mattnibs (approved):

ppl/zqd/apiserver/manager.go
alfred-landrum: On balance, I think it's more important to return a "not found" error to the authenticated & rightful owner of a space that's (say) been deleted, rather than try to guard against accidentally allowing the existing of a space id to be known to non-owners.

PR #2018, opened 2021-01-27
put: support assigning to . and to nested fields

### LINKED ISSUE #1517: zql: put with dot on LHS creates empty-named field
```
~/work/brim/zq(master)
$ cat in.tzng 
#0:record[a:int32,b:int32]
0:[1;2;]
~/work/brim/zq(master)
$ zq -t "put .=1" in.tzng 
#0:record[a:int32,b:int32,:int64]
0:[1;2;1;]
```

### LINKED ISSUE #1753: nested lval support for put
The put proc gives an error if you use a nested field reference as the target.  We should fix this as it's easy now.  We have the technology.

### LINKED ISSUE #1932: putting to `.` results in empty-named fields

henridf:
close #1932 
close #1753 
close #1517


proc/put/put.go
```diff
@@ -81,116 +84,284 @@ func (p *Proc) eval(in *zng.Record) ([]zng.Value, error) {
...
+const (
+	fromInput  step = iota // copy field from input record
+	fromClause             // copy field from put assignment
+	root                   // values are being copied to the root record (put .=)
+	record                 // recurse into record below us
+)
+
+func (s op) build(in zcode.Bytes, b *zcode.Builder, vals []zng.Value) error {
+	if s.op == root {
+		iter := zcode.Iter(vals[s.index].Bytes)
```
nwt, line 112:
```suggestion
		iter := vals[s.index].Bytes.Iter()
```

proc/put/put.go
```diff
@@ -81,116 +84,284 @@ func (p *Proc) eval(in *zng.Record) ([]zng.Value, error) {
...
+			bytes, c, err := iter.Next()
+			if err != nil {
+				return err
+			}
+			if c {
+				b.AppendContainer(bytes)
+			} else {
+				b.AppendPrimitive(bytes)
+			}
+		}
```
nwt, line 123: Instead of copying `vals[s.index].Byte` using `b` like this, maybe change the method signature so you can use `copy`?
```go
func (s op) build(in zcode.Bytes, b *zcode.Builder, vals []zng.Value) error {
	switch s.op {
	case root:
		bytes := make(zcode.Bytes, len(vals[s.index].Bytes))
		copy(bytes, vals[s.index].Bytes)
		return bytes, nil
	case record:
		b.Reset()
		if err := s.buildRecord(in, b, vals); err != nil {
			return nil, err
		}
		return b.Bytes(), nil
	default:
		// top-level op should be root or record
		panic(fmt.Sprintf("put: unexpected op %v", op))
}
```

proc/put/put.go
```diff
@@ -81,116 +84,284 @@ func (p *Proc) eval(in *zng.Record) ([]zng.Value, error) {
...
-	clauseTypes := make([]clauseType, len(p.clauses))
-	var nreplace int
-	replace := make([]int, n)
-	for k := range replace {
-		replace[k] = -1
+// A op is a recursive data structure encoding a series of steps to be
+// carried out to construct an output record from an input record and
+// a slice of evaluated clauses.
+type op struct {
+	op        step
```
nwt, line 91: Swapping these type names would make this so much clearer.
```suggestion
type step struct {
	op        op
```

proc/put/put.go
```diff
@@ -199,8 +370,13 @@ func (p *Proc) Pull() (zbuf.Batch, error) {
...
 	for k := 0; k < batch.Length(); k++ {
 		in := batch.Index(k)
+		rec, err := p.put(in)
+		if err != nil {
+			return nil, err
+		}
 		// Keep is necessary because put can return its argument.
-		recs = append(recs, p.put(in).Keep())
+		rec.Keep()
+		recs = append(recs, rec)
```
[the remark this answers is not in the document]
henridf: 🤕 thanks!

proc/put/put.go
henridf: 👍 (and switch is also more readable).

proc/put/put.go
henridf: Agreed, done in 00ad1f26035b85c814ebd512daaad2b39e14c7aa
(I went back and forth and had that at some point (as reflected by `func (s op)`) but got stuck in a local minima. )

field/field.go
```diff
@@ -35,6 +38,20 @@ func (f Static) Equal(to Static) bool {
 	return true
 }
 
+func (f Static) IsRoot() bool {
+	return len(f) == 0
+}
+
+func (f Static) IsParent(child Static) bool {
```
nwt, line 45: The meaning of "parent" in this context isn't obvious. I think the intent would be clearer if it were remade as HasPrefix, with behavior analogous to bytes.HasPrefix and strings.HasPrefix.
```go
func (f Static) HasPrefix(prefix Static) bool {
	return len(f) >= len(prefix) && prefix.Equal(f[:len(prefix)])
}

File: proc/put/put.go [hunk 6]
```diff
@@ -81,116 +84,279 @@ func (p *Proc) eval(in *zng.Record) ([]zng.Value, error) {
...
+			b.BeginContainer()
+			bytes, err := in, error(nil)
+			if step.index >= 0 {
+				bytes, err = ig.nth(step.index)
+				if err != nil {
+					return err
+				}
+			}
+			err = step.buildRecord(bytes, b, vals)
+			if err != nil {
```
nwt, line 159:
```suggestion
			if err := step.buildRecord(bytes, b, vals); err != nil {
```

proc/put/put.go
```diff
@@ -81,116 +84,279 @@ func (p *Proc) eval(in *zng.Record) ([]zng.Value, error) {
...
+			return step{}, nil, fmt.Errorf("put .=x: cannot put a non-record to .")
+		}
+		typ, err := p.pctx.TypeContext.LookupTypeRecord(recVal.Columns)
+		return step{op: root, index: 0}, typ, err
+	}
+	return p.deriveRecordRule(parentPath, inType.Columns, vals)
+}
+
+func (p *Proc) deriveRecordRule(parentPath field.Static, inCols []zng.Column, vals []zng.Value) (step, *zng.TypeRecord, error) {
+	o := step{op: record}
```
nwt, line 224: `s` seems like a better name for this than `o`.

proc/put/put.go
```diff
@@ -81,116 +84,279 @@ func (p *Proc) eval(in *zng.Record) ([]zng.Value, error) {
...
+	// special case: assign to root (put .=x)
+	if p.clauses[0].LHS.IsRoot() {
+		recVal, ok := vals[0].Type.(*zng.TypeRecord)
+		if !ok {
+			return step{}, nil, fmt.Errorf("put .=x: cannot put a non-record to .")
+		}
+		typ, err := p.pctx.TypeContext.LookupTypeRecord(recVal.Columns)
+		return step{op: root, index: 0}, typ, err
+	}
+	return p.deriveRecordRule(parentPath, inType.Columns, vals)
```
nwt, line 220: I'd get rid of the first argument to this method since it's always `field.NewRoot()`.
```suggestion
func (p *Proc) deriveRule(inType *zng.TypeRecord, vals []zng.Value) (step, *zng.TypeRecord, error) {
	// special case: assign to root (put .=x)
	if p.clauses[0].LHS.IsRoot() {
		recVal, ok := vals[0].Type.(*zng.TypeRecord)
		if !ok {
			return step{}, nil, fmt.Errorf("put .=x: cannot put a non-record to .")
		}
		typ, err := p.pctx.TypeContext.LookupTypeRecord(recVal.Columns)
		return step{op: root, index: 0}, typ, err
	}
	return p.deriveRecordRule(field.NewRoot(), inType.Columns, vals)
```

proc/put/put.go
```diff
@@ -81,116 +84,279 @@ func (p *Proc) eval(in *zng.Record) ([]zng.Value, error) {
...
+}
+
+func (p *Proc) deriveRecordRule(parentPath field.Static, inCols []zng.Column, vals []zng.Value) (step, *zng.TypeRecord, error) {
```
nwt, line 223: Neither `deriveRecordRule` nor `deriveRule` returns a `putRule`, making those names a little misleading.

proc/put/put.go
henridf: ✅

field/field.go
henridf: I made this change in af793d6fa5b91ff2265691ec82fcb32a68fb1841. The changed semantics make the client code a bit more complicated. What do you think about adding 

```go
func (f Static) HasStrictPrefix(prefix Static) bool {
	return len(f) > len(prefix) && prefix.Equal(f[:len(prefix)])
}
```

?

proc/put/put.go
henridf: it is.

proc/put/put.go
henridf: good catch! this was a relic of some earlier step.

proc/put/put.go
henridf: changed to deriveSteps and deriveRecordSteps in 2ae01362d60b1044293a88a4754d376fd0cef878

field/field.go
nwt: 👍

proc/put/put.go
```diff
@@ -34,28 +36,29 @@ type Proc struct {
...
+	clauseTypes []zng.Type
+	step        step
 }
 
 func New(pctx *proc.Context, parent proc.Interface, clauses []expr.Assignment) (proc.Interface, error) {
+	for i, p := range clauses {
+		for j, c := range clauses {
+			if i != j && p.LHS.Equal(c.LHS) {
+				return nil, fmt.Errorf("put: multiple assignments to %s", p.LHS)
+			}
```
nwt, line 48: Why do any checks when `i == j`?
```suggestion
			if i == j {
				continue
			}
			if p.LHS.Equal(c.LHS) {
				return nil, fmt.Errorf("put: multiple assignments to %s", p.LHS)
			}
```

proc/put/put.go
```diff
@@ -34,28 +36,29 @@ type Proc struct {
...
+			if c.LHS.HasPrefix(p.LHS) && !c.LHS.Equal(p.LHS) {
+				return nil, fmt.Errorf("put: conflicting nested assignments to %s and %s", p.LHS, c.LHS)
+			}
+
+		}
```
nwt, line 53:
```suggestion
			}
		}
```

nwt (approved):

proc/put/put.go
henridf: 👍

field/field.go
henridf: done

Issue #2017, opened 2021-01-26
reduce source wait time for archive queries


alfred-landrum:
A possible reason we hit #1997 was that we were waiting a longer than expected time for a source to be created and pushed onto the source channel in `spanMultiSource.SendSources`, due the time needed to perform the tsDir walking operation. That involves performing list operations against S3 and reading chunk metadata files to create the list of files to read for a query. We already have caching work in progress that will help with the chunk metadata file access, and we have plans to work on caching the directory lookup operations as well.

This issue tracks two additional ways we can reduce the wait time:
* we can use channel buffering or some similar technique to generate the sources ahead of time, before a worker needs to wait on the source channel. I think that just adding a buffer capacity to the `chan Source` is not sufficient, since a Source is a single file in a tsDir, and I think we should try to 'readahead' an entire tsDir's worth of Chunks.
* We should also be able to parallelize most of the work in `tsDirVisit`, most easily the part that reads the metadata files to create the Chunk structs, and with a little more care the reading of the tsDir entries, given the ordering guarantee.

alfred-landrum:
I added the buffering I mentioned above in #2019, and am handing over to @mattnibs to handle the tsDirVisit parallelization.

philrz:
Verified in `zar` commit `1ac7787f`.

As described in the linked PR, a decent test data set is the oft-cited "corelight" data set, which is first imported from my laptop into an archive on S3.

```
$ zq * | zar import -R s3://phil-scratch-us-east-2/archive -s 25MB -
```

Going back to `zq` commit `10c789a2` that comes right before the changes in the linked PR, we see it consistently takes 11+ seconds to complete a `zar stat` of the archive.

```
$ zar -version
Version: v0.29.0-53-g10c789a2

$ time zar stat -R s3://phil-scratch-us-east-2/archive > /dev/null
real	0m11.415s
user	0m0.454s
sys	0m0.312s

$ time zar stat -R s3://phil-scratch-us-east-2/archive > /dev/null
real	0m11.640s
user	0m0.363s
sys	0m0.244s

$ time zar stat -R s3://phil-scratch-us-east-2/archive > /dev/null
real	0m11.712s
user	0m0.377s
sys	0m0.240s
```

But at `zar` commit `1ac7787f`, the same operation is more than 2x faster.

```
$ zar -version
Version: v0.29.0-59-g1ac7787f

$ time zar stat -R s3://phil-scratch-us-east-2/archive > /dev/null
real	0m4.942s
[...]

PR #2019, opened 2021-01-27
add tsDir SpanInfo prefetch during archive SendSources

alfred-landrum:
Related to #2017 .

### DIFF
--- a/ppl/archive/multisource.go
+++ b/ppl/archive/multisource.go
@@ -125,23 +125,16 @@ func (m *spanMultiSource) SendSources(ctx context.Context, span nano.Span, srcCh
 		return err
 	})
 	g.Go(func() error {
-		for {
-			select {
-			case sinfos, ok := <-sinfosChan:
-				if !ok {
-					return nil
+		for sinfos := range sinfosChan {
+			for _, si := range sinfos {
+				select {
+				case srcChan <- &spanSource{ark: m.ark, spanInfo: si, stats: m.stats}:
+				case <-ctx.Done():
+					return ctx.Err()
 				}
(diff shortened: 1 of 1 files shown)


alfred-landrum:
I'm marking this as ready for review: in our sync meeting, @mattnibs suggested & I agreed that some kind of metric to show the difference with this would be useful. I realized though that there isn't a clear place to 'hang' the metric off of - in other parts of the code, we pass through a Registerer, but there's no place obvious to do so. That feels like something we should tackle, but I think I'd rather do that separate from this. I did verify that this makes a difference with my testing with extra code (not present in the PR) that totaled the wait time for workers.

ppl/archive/multisource.go
```diff
@@ -105,14 +106,44 @@ func (m *spanMultiSource) OrderInfo() (field.Static, bool) {
...
+				for _, si := range sinfos {
+					select {
+					case srcChan <- &spanSource{ark: m.ark, spanInfo: si, stats: m.stats}:
+					case <-ctx.Done():
+						return ctx.Err()
+					}
+				}
+			case <-ctx.Done():
+				return ctx.Err()
+			}
```
nwt, line 143: Since the other goroutine is guaranteed to close sinfosChan, you could shorten this.
```suggestion
		for sinfos := range sinfosChan {
			for _, si := range sinfos {
				select {
				case srcChan <- &spanSource{ark: m.ark, spanInfo: si, stats: m.stats}:
				case <-ctx.Done():
					return ctx.Err()
				}
			}
```

nwt (approved):

PR #2020, opened 2021-01-27
archive: Immutable local cache

### LINKED ISSUE #1898: Small file in-process caching
Add in-process small file caching for single node zqd deployments, as preparation for later Redis based caching. This should include & test Prometheus metrics that highlight cache hits and misses. The https://github.com/hashicorp/golang-lru package may be useful.

mattnibs:
Add ability to enable local caching of small immutable files (currently this is
only chunk metadata) in an archive. Add prometheus metrics for tracking cache
hits and misses on gets for small files.

Closes #1898

### DIFF
--- a/pkg/iosrc/mux.go
+++ b/pkg/iosrc/mux.go
@@ -1,0 +1,144 @@
+package iosrc
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+)
+
+// MuxSource is a Source that routes each function call to the correct Source
+// based off the provided URI's scheme.
+type MuxSource struct {
+	schemes map[string]Source
+}
+
+func NewMuxSource(m map[string]Source) *MuxSource {
+	return &MuxSource{m}
+}
+
+func (m *MuxSource) AddScheme(scheme string, source Source) {
+	m.schemes[scheme] = source
+}
+
+func getScheme(uri URI) string {
+	if uri.Scheme == "" {
+		return FileScheme
+	}
+	return uri.Scheme
+}
+
+func (m *MuxSource) GetSource(u URI) (Source, error) {
+	scheme := getScheme(u)
+	source, ok := m.schemes[scheme]
+	if !ok {
+		return nil, fmt.Errorf("unknown scheme: %q", scheme)
+	}
+	return source, nil
+}
+
+func (m *MuxSource) NewReader(ctx context.Context, uri URI) (Reader, error) {
+	source, err := m.GetSource(uri)
+	if err != nil {
+		return nil, err
+	}
+	return source.NewReader(ctx, uri)
+}
+
+func (m *MuxSource) NewWriter(ctx context.Context, uri URI) (io.WriteCloser, error) {
+	source, err := m.GetSource(uri)
+	if err != nil {
+		return nil, err
+	}
+	return source.NewWriter(ctx, uri)
+}
+
+func (m *MuxSource) ReadFile(ctx context.Context, uri URI) ([]byte, error) {
+	source, err := m.GetSource(uri)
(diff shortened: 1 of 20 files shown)


pkg/iosrc/mux.go
```diff
@@ -0,0 +1,144 @@
...
+func (m *MuxSource) ReadFile(ctx context.Context, uri URI) ([]byte, error) {
+	source, err := m.GetSource(uri)
+	if err != nil {
+		return nil, err
+	}
+	return source.ReadFile(ctx, uri)
+}
+
+func (m *MuxSource) WriteFile(ctx context.Context, uri URI, d []byte) error {
+	source, err := GetSource(uri)
```
nwt, line 66: Should this be m.GetSource?

alfred-landrum:
I don't see a way to enable this via the zqd command line - would you add that, please?

mattnibs:
> I don't see a way to enable this via the zqd command line - would you add that, please?

Sure

pkg/iosrc/mux.go
mattnibs: Whoops, yes.

mattnibs:
> I don't see a way to enable this via the zqd command line - would you add that, please?

Done. Adding a ztest for this however is a little tricky.

alfred-landrum:
> > I don't see a way to enable this via the zqd command line - would you add that, please?
> 
> Done. Adding a ztest for this however is a little tricky.

If it helps, see the usage of `$ZQD_EXTRA_FLAGS` in a few of the auth tests, as a way to still use a common `services.sh`.

alfred-landrum:
FYI, I pulled this branch, forced the default size of the cache to be non-zero & large, and deployed to k8s, and can see from both timing query responses & the prometheus metrics that the cache is working.

pkg/promtest/promtest.go
```diff
@@ -0,0 +1,42 @@
+package promtest
```
alfred-landrum, line 1: 👍 this is a nice addition!

ppl/archive/schema.go
```diff
@@ -95,6 +96,10 @@ type Archive struct {
 	DataOrder        zbuf.Order
 	LogSizeThreshold int64
 	LogFilter        []ksuid.KSUID
+
+	immfiles interface {
```
alfred-landrum, line 100: How about just type `immcache.ImmutableCache`, as in `OpenOptions.ImmutableCache`?

mattnibs: I'm hesitant about this because this is not necessarily a cache- caching could be disabled, the file could not exist in the cache and need to be fetched.

ppl/zqd/ztests/cache-local.yaml
```diff
@@ -0,0 +1,23 @@
+script: |
+  ZQD_EXTRA_FLAGS=-immcache.localsize=128 source services.sh
+  zapi -h $ZQD_HOST new -d ./root -k archivestore -thresh 20KiB testsp
+  zapi -h $ZQD_HOST -s testsp post babble.tzng >/dev/null
+  echo ===
+  zapi -h $ZQD_HOST -s testsp get -f text "count()"
+  echo ===
+  zapi -h $ZQD_HOST -s testsp get -f text "count()"
+
```
alfred-landrum, line 9: How about adding something like 'curl $ZQD_HOST/metrics | grep archive_cache_hits_total ...' here?

alfred-landrum (approved):

ppl/zqd/ztests/cache-local.yaml
mattnibs: Good idea, will do 👍🏻

PR #2023, opened 2021-01-27
print any log files when awaitfile fails

alfred-landrum:
When `awaitfile` fails, typically because of a failure of a zqd instance to start, make sure to output any .log files found. This will capture the log files for the zqd root, recruiter, and worker instances in some of the ztests that verify distributed queries.


marktwallace (approved):
This is a great idea. I had been adding the print manually, but it makes sense that it should cat the file on every timeout.

PR #2026, opened 2021-01-28
refactor package filter into expr and string-search components

mccanne:
This commit refactors the filter package into the pieces that
perform the runtime logic (moved to package expr) and the pieces
that do generic string scanning (moved to pkg/stringsearch).

filter.Predicate has been renamed expr.Boolean and filter.Filter
is now expr.Filter.  expr.Boolean applies a predicate to a
zng.Value while expr.Filter applies a predicate to a zng.Record.

These changes are strictly code movement and renaming.  There are
no semenantic changes to the code here.

This refactoring is needed by the upcoming PR that unifies the search 
and expression syntax and machinery.


pkg/stringsearch/fieldnamefinder.go
```diff
@@ -9,20 +9,20 @@ import (
 	"github.com/brimsec/zq/zng/resolver"
 )
 
-type fieldNameFinder struct {
+type FieldNameFinder struct {
```
nwt, line 12: This is fine for now, but package pkg/stringsearch feels like an odd home for something so zng-specific.

expr/bufferfilter.go
```diff
@@ -21,9 +22,9 @@ type BufferFilter struct {
 	op    int
 	left  *BufferFilter
 	right *BufferFilter
-	fnf   *fieldNameFinder
-	scf   *stringCaseFinder
-	sf    *stringFinder
+	fnf   *stringsearch.FieldNameFinder
+	scf   *stringsearch.CaseFinder
+	sf    *stringsearch.Finder
```
nwt, line 27: This is fine for now, but I'd eventually like to rename these to `cf` and `f` to keep them in sync with their types.

pkg/stringsearch/fieldnamefinder.go
```diff
@@ -9,20 +9,20 @@ import (
...
 	checkedIDs       big.Int
-	fieldNameIter    fieldNameIter
-	stringCaseFinder *stringCaseFinder
+	fieldNameIter    FieldNameIter
+	stringCaseFinder *CaseFinder
```
nwt, line 15: This is fine for now, but I'd eventually like to rename this to `caseFinder` to keep it in sync with its type.

nwt (approved):
:shipit:

pkg/stringsearch/fieldnamefinder.go
mccanne: Agreed.  We do this in a couple places.  I thought about putting it in zq/stringsearch but that didn't feel right either...?

expr/bufferfilter.go
mccanne: I'll change it.

pkg/stringsearch/fieldnamefinder.go
mccanne: I'll change it.

pkg/stringsearch/fieldnamefinder.go
nwt: I think we should merge this as is and move it later.

PR #2031, opened 2021-01-29
finish up zson scanner to read input incrementally

### LINKED ISSUE #1802: zson scanner should read input as stream
Currently, the zson scanner reads in the entire input before starting its work.  This was good to get the design off the ground but now that the dust has settled and the scanner's requirements are clear, this should be changed to read input as a stream.

mccanne:
This commit addresses the placeholder logic in the zson scanner
that reads in the entire input upon open to simplify the scanning
logic.  Now that the zson parser is debugged and complete, this
change fixes the lexer to incrementally read input from the io.Reader
as needed.

Fixes #1802

### DIFF
--- a/zson/lexer.go
+++ b/zson/lexer.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"errors"
 	"io"
-	"io/ioutil"
 	"regexp"
 	"strings"
 	"unicode"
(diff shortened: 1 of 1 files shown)


henridf (approved):

philrz:
Verified in `zq` commit `ca01b3e1`.

To repro the problem being solved here, we circle back to `zq` commit `15bcf3f` that pre-dates the changes in the linked PR. If we execute `zq -f zson conn.zng.gz | zq -i zson -f tzng -` against the [zq-sample-data](https://github.com/brimsec/zq-sample-data), we see that there's a pause of several seconds before the downstream `zq` is able to start outputting TZNG, reflecting what's described in the opening issue text about waiting for the entire input.

https://user-images.githubusercontent.com/5934157/106973321-7dec4400-6707-11eb-8e40-65eae0df03b3.mov

Then repeating the same in `zq` commit `ca01b3e1`, we see the downstream `zq` is able to start streaming output almost immediately, indicating that it was able to start scanning incoming ZSON much sooner.

https://user-images.githubusercontent.com/5934157/106973366-95c3c800-6707-11eb-8c7a-905aab8c6927.mov

Thanks @mccanne!

PR #2032, opened 2021-01-29
remove support for subspaces

### LINKED ISSUE #2028: Remove zq repo subspace code
Once the UI side is done removing any subspace related code: https://github.com/brimsec/brim/issues/1397 , at a minimum any tests that depend on subspaces, we can remove subspace support from the archive package, and any other references in the zq repo.

alfred-landrum:
Closes #2028 .

Successful Brim tests in https://github.com/brimsec/brim/pull/1398 .


ppl/archive/schema.go
```diff
@@ -135,9 +134,7 @@ func (ark *Archive) ReadDefinitions(ctx context.Context) (index.Definitions, err
 	return defs, err
 }
 
-type OpenOptions struct {
-	LogFilter []string
-}
+type OpenOptions struct{}
```
alfred-landrum, line 137: I'm leaving this here as @mattnibs is about to add a field to it in #2020 .

mattnibs: Much appreciated

ppl/zqd/db/filedb/migrate.go
```diff
@@ -140,3 +127,38 @@ func migrateV5(data []byte) (int, []byte, error) {
...
+
+func migrateV6(data []byte) (int, []byte, error) {
+	var dbv5 dbdataV5
+	if err := json.Unmarshal(data, &dbv5); err != nil {
+		return 0, nil, err
+	}
+	var db dbdata
+	db.Version = 6
+	db.SpaceRows = make([]schema.SpaceRow, 0, len(dbv5.SpaceRows))
+	for _, r := range dbv5.SpaceRows {
```
mattnibs, line 153: This probably doesn't matter since I doubt there are any zqd instances with a subspace in the wild... but should we drop subspaces here?

mattnibs (approved):
👏🏻

ppl/zqd/db/filedb/migrate.go
alfred-landrum: I convinced myself that it was ok. We never exposed it in Brim, except for some internal demos. Plus, the subspace was essentially just a filtered view to the parent space's data - so removing the subspace doesn't actually delete any data.

Issue #2016, opened 2021-01-26
intermittent system test failure: pcappost


mattnibs:
This failure [was observed on master](https://github.com/brimsec/zq/runs/1770356191?check_suite_focus=true) while running through the zqd pcap testing suite. The test that failed was pcappost-suricata, but I seem to recall other tests in this suite also failing intermittently:

```
 --- FAIL: TestZq/ppl/zqd/ztests/pcap (0.02s)
        --- PASS: TestZq/ppl/zqd/ztests/pcap/postpcap (9.09s)
        --- PASS: TestZq/ppl/zqd/ztests/pcap/pcap-deleted (9.15s)
        --- FAIL: TestZq/ppl/zqd/ztests/pcap/postpcap-suricata (2.58s)
        --- PASS: TestZq/ppl/zqd/ztests/pcap/postpcap-suricata-update (12.78s)
```

mattnibs:
Would be nice to have more logs available when a ztest fails.

mattnibs:
Another occurrence: https://github.com/brimsec/zq/runs/1793149222?check_suite_focus=true

mattnibs:
At least we got some log output on this last failure:

```
ztest.go:398: /home/runner/work/zq/zq/ppl/zqd/ztests/pcap/postpcap-suricata.yaml: expected and actual stdout differ:
        --- expected
        +++ actual
        @@ -1,9 +1,4 @@
        -#0:record[ts:time,count:uint64]
        -0:[1425568033;13;]
        -0:[1425567868;2;]
        -#0:record[count:uint64]
        -0:[15;]
         ===
         #0:record[count:uint64]
         0:[379;]
```

mattnibs:
Hmm so looks like the suricata events aren't showing up.

philrz:
pcap processing is now handled by the separate [Brimcap](https://github.com/brimdata/brimcap) tool, outside of Zed. It looks like this test that was once failing is now handled via the test laid out in [analyze-all.yaml](https://github.com/brimdata/brimcap/blob/main/cmd/brimcap/ztests/analyze-all.yaml). I believe this test now runs reliably, so i'm closing this issue.

PR #2034, opened 2021-01-29
pcappost suricata ztest: empty stderr

mattnibs:
Make ztest for pcappost-suricata assume the result for stderr is an empty
string. This will (hopefully) give us more information should (when) this flaky
test fails again.

Reference: #2016


ppl/zqd/ztests/pcap/postpcap-suricata.yaml
```diff
@@ -23,3 +23,6 @@ outputs:
       ===
       #0:record[count:uint64]
       0:[379;]
+
```
nwt, line 26:
```suggestion
```

nwt (approved):

PR #2036, opened 2021-01-29
Add prototype of Temporal-managed compaction

### LINKED ISSUE #1859: Prototype post import compaction with temporal
I've discussed this with @nwt as our first attempt using Temporal: When we import data to an archive, we want to run compaction over the time range of the imported data, though to reduce write amplification, it makes sense to wait in case other data arrives. We want to demonstrate how this would work with temporal, by having a test we can run locally with a makefile target, that uses something like docker compose to make a Temporal instance available.

nwt:
This is a prototype of how we might use Temporal to manage archive
compaction.

Closes #1859.


Makefile
```diff
@@ -100,6 +101,35 @@ test-postgres-docker:
...
+	ln -fs $(<F) $@
+
+bin/tctl-$(TEMPORAL_TAG):
+	mkdir -p $(@D)
+	echo 'module deps' > $@.mod
+	go get -d -modfile=$@.mod go.temporal.io/server/cmd/tools/cli@$(TEMPORAL_TAG)
+	go build -modfile=$@.mod -o $@ go.temporal.io/server/cmd/tools/cli
+
+.PHONY: test-temporal-docker
+test-temporal-docker: bin/docker-compose.temporal-$(TEMPORAL_TAG).yaml
```
mattnibs, line 121: Like I did in my [redis pr](https://github.com/brimsec/zq/pull/2040) I think you should merge this with test-postgres-docker to renamed test-services-docker and add the docker compose temporal stuff to ppl/zqd/scripts/dkc-services-yaml. I looked at the temporal docker-compose and it has stuff like temporal-admin-tools and temporal-web that we won't need for testing environments. Plus it would be nice to have temporal using the same postgres db that standup for postgres tests.

ppl/cmd/zqd/listen/command.go
```diff
@@ -147,7 +150,11 @@ func (c *Command) Run(args []string) error {
 			return err
 		}
 	}
-	return srv.Wait()
+	if core.IsTemporalWorker() {
+		g.Go(func() error { return core.RunTemporalWorker(ctx) })
+	}
+	g.Go(srv.Wait)
+	return g.Wait()
```
mattnibs, line 157: IMHO I think I'd rather have core.New launch the temporal go routine and have the call to core.Shutdown exit the go-routine.

Makefile
alfred-landrum: I like that general direction, but noticed @nwt 's comment about temporal sometimes taking 10s to come up.

ppl/zqd/temporal/spaceworkflow.go
```diff
@@ -0,0 +1,100 @@
...
+		var signalVal string
+		signalCh.Receive(ctx, &signalVal)
+
+		logger.Info("Workflow waiting for timer or signal")
+		s := workflow.NewSelector(ctx)
+		timerCtx, cancel := workflow.WithCancel(ctx)
+		timerFuture := workflow.NewTimer(timerCtx, config.SpaceCompactDelay)
+		var err error
+		var executedActivity bool
+		s.AddFuture(timerFuture, func(f workflow.Future) {
```
alfred-landrum, line 42:
```suggestion
		s.AddFuture(timerFuture, func(_ workflow.Future) {
```

ppl/zqd/temporal/spaceworkflow.go
```diff
@@ -0,0 +1,100 @@
...
+			logger.Info("Workflow timer expired")
+			ctx := workflow.WithActivityOptions(ctx, workflow.ActivityOptions{
+				StartToCloseTimeout: 12 * time.Hour,
```
alfred-landrum, line 45: The temporal go docs say that both `StartToCloseTimeout` (which is here) and `ScheduleToStartTimeout` are mandatory, but this is clearly working without ScheduleToStartTimeout. Can you see if its really mandatory, in case not having a value leads to unexplained errors later?

alfred-landrum (commented):
I discussed with @nwt , and he's going to introduce a change where the workflow is created when the space is created & deleted on space deletion, as that will prepare us a little better for having streaming ingest.

Makefile
nwt: Done in 07e423bb3547fe9a358c1fb14c5c09238d70305e.

ppl/zqd/scripts/dkc-services.yaml
```diff
@@ -12,3 +12,15 @@ services:
...
+  temporal:
+    depends_on:
+      - database
+    environment:
+      - DB=postgresql
+      - DB_PORT=5432
+      - POSTGRES_USER=test
+      - POSTGRES_PWD=test
+      - POSTGRES_SEEDS=database
+    image: temporalio/auto-setup:${TEMPORAL_VERSION:?}
```
nwt, line 24: @mattnibs: Any concerns about this variable?

ppl/cmd/zqd/listen/command.go
nwt: Done in 17c20124e8a4d5f97b496b34b532153f0ba75ffa.

ppl/zqd/temporal/spaceworkflow.go
```diff
@@ -0,0 +1,100 @@
...
+				logger.Info("Workflow completed")
+				return nil
+			}
+		}
+	}
+}
+
+func spaceCompactActivity(ctx context.Context, id api.SpaceID) error {
+	l := activity.GetLogger(ctx)
+	l.Info("Activity startexd")
```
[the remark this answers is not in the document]
nwt: Fixed in 88f6b510fbe579129218be6015ec8f3fdc57bbae.

ppl/zqd/temporal/notifier.go
```diff
@@ -0,0 +1,45 @@
...
+	return &Notifier{client, logger}, err
+}
+
+func (n *Notifier) Shutdown() {
+	n.client.Close()
+}
+
+func (n *Notifier) SpaceDeleted(ctx context.Context, id api.SpaceID) {
+	if err := n.client.CancelWorkflow(ctx, id.String(), ""); err != nil {
+		n.logger.Warn("Canceling workflow failed", zap.Error(err), zap.Stringer("space_id", id))
```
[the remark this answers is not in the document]
nwt: As we discussed, I added `Notifier.SpaceCreated` to create the workflow for a space and changed `spaceWorkflow` to run forever (using `go.temporal.io/sdk/workflow.NewContinueAsNewError`). In light of that, think a warning here is appropriate now, so I've left this as is.

ppl/zqd/temporal/spaceworkflow.go
nwt: Done in 3591e5cf43e6699497dfb8b7fb8bee1a207e8115.

ppl/zqd/temporal/spaceworkflow.go
nwt: I think this is OK. The Go SDK docs at https://pkg.go.dev/go.temporal.io/sdk@v1.4.1/ look like they're out of sync with the underlying implementation at https://pkg.go.dev/go.temporal.io/api@v1.4.0/command/v1#ScheduleActivityTaskCommandAttributes, which says that either `ScheduleToCloseTimeout` or `StartToCloseTimeout` must be set.

alfred-landrum (approved):

mattnibs (approved):

PR #2053, opened 2021-02-02
rename archive package to lake

alfred-landrum:
This change renames the archive package to lake, the `Archive` struct to `Lake`, and the canonical variable name for a `*Archive` from `ark` to `lk`. No other changes (api routes, metric names, etc) are changed.


alfred-landrum:
I see there are still some usages of 'ark' as a variable name, will continue to search & replace.

mccanne:
I like it!

nwt (approved):
This still leaves us with a grip of "archive" occurrences.
```
$ git grep -i archive | wc -l
     221
```
Do we plan to eliminate those separately?

alfred-landrum:
> This still leaves us with a grip of "archive" occurrences.
> 
> ```
> $ git grep -i archive | wc -l
>      221
> ```
> 
> Do we plan to eliminate those separately?

Yes - many of those left are references to the api structs, or the storage kind, which will take some Brim coordination & migration to manage.

PR #2055, opened 2021-02-02
Fix data parameter in Makefile

marktwallace:
This fixes a bug Matt noticed today where the -data parameter was not set correctly for the root zqd process.


mattnibs:
@marktwallace are you planning on fixing postgres in this pr or another pr?

charts/zservice/templates/_helpers.tpl
```diff
@@ -73,6 +73,11 @@ Create args that vary based on .Values.personality
 {{- end }}
 {{- if eq .Values.personality "root" }}
 {{- $args = append $args (print "-data=" .Values.datauri) }}
+{{- $args = append $args (print "db.kind=postgres") }}
+{{- $args = append $args (print "db.postgres.addr=" .Values.postgres.addr) }}
+{{- $args = append $args (print "db.postgres.database=" .Values.postgres.database) }}
+{{- $args = append $args (print "db.postgres.user=" .Values.postgres.username) }}
+{{- $args = append $args (print "db.postgres.passwordFile=/creds/postgres/password") }}
```
mattnibs, line 80: Missing dashes to make these args:
```suggestion
{{- $args = append $args (print "-db.kind=postgres") }}
{{- $args = append $args (print "-db.postgres.addr=" .Values.postgres.addr) }}
{{- $args = append $args (print "-db.postgres.database=" .Values.postgres.database) }}
{{- $args = append $args (print "-db.postgres.user=" .Values.postgres.username) }}
{{- $args = append $args (print "-db.postgres.passwordFile=/creds/postgres/password") }}
```

mattnibs:
I would also get rid of the `-d` arg in the `zapi new` command for the cd cluster/space setup.

Makefile
```diff
@@ -107,8 +107,8 @@ test-services-docker:
 .PHONY: test-cluster
 test-cluster: build install
 	-zapi rm files
-	-aws s3 rm --recursive $(ZQD_DATA_URI)
-	zapi new -k archivestore -d $(ZQD_DATA_URI) files
+	-aws s3 rm --recursive $(ZQD_DATA_URI)/zd
```
mattnibs, line 110: I would just do `-zapi rm files` instead of the direct aws s3 cmd

mattnibs (approved):

Makefile
marktwallace: The "s3 rm" is needed because the "zapi postpath" below will add more S3 files to that directory -- duplicate records from what was added with previous tests by "zapi postpath".

"zapi rm files" has no effect on S3, so both commands are required.

PR #2059, opened 2021-02-03
Fix fill() logic

henridf:


nwt (approved):

expr/shaper.go
```diff
@@ -336,19 +336,32 @@ func (c *Shaper) fillRecordType(input, spec *zng.TypeRecord) (*zng.TypeRecord, e
...
-		ind, _ = input.ColumnOfField(name)
-		inCol := input.Columns[ind]
+		if !specOk {
+			// Field not in spec: input passes through.
+			cols = append(cols, inCol)
+			continue
+		}
+
+		// Field is present both in input and spec: recurse if
+		// both records, or select appropriate type if not.
```
nwt, line 362: I do think this would be a bit simpler without `names`.
```go
	cols := make([]zng.Column, len(input.Columns), len(input.Columns)+len(spec.Columns))
	copy(cols, input.Columns)
	for _, specCol := range spec.Columns {
		if i, ok := input.ColumnOfField(specCol.name); ok {
			// Field is present both in input and spec: recurse if
			// both records, or select appropriate type if not.
			if specRecType, ok := specCol.Type.(*zng.TypeRecord); ok {
				if inRecType, ok := input.Columns[i].Type.(*zng.TypeRecord); ok {
					var err error
					cols.Columns[i], err = c.fillRecordType(inRecType, specRecType)
					if err != nil {
						return nil, err
					}
				} else {
					cols.Columns[i] = specCol
				}
			}
		} else {
			cols = append(cols, specCol)
		}
	}
```

henridf: oh yes! 😍

PR #2060, opened 2021-02-03
redis helm

### LINKED ISSUE #1899: Small file redis caching
Similar to how the initial Postgres work was performed, add support for using Redis as a small file cache, and create makefile targets that can be used to test against a pre-existing Redis endpoint, or docker compose to temporarily stand up a Redis instance. This include CI support, and Prometheus metrics for cache hits and misses.

mattnibs:
Add redis to umbrella zservice helm chart.

Closes #1899


charts/zservice/charts/redis/values.yaml
```diff
@@ -0,0 +1,18 @@
...
+    enabled: false
+
+  # set maxmemory + maxmemory-policy in order to use redis as an lru cache. See
+  # https://redis.io/topics/lru-cache
+  maxmemory: 1gb
+  maxmemoryPolicy: volatile-lfu
+
+  configmap: |
+    maxmemory {{ .Values.maxmemory }}
+    maxmemory-policy {{ .Values.maxmemoryPolicy }}
```
marktwallace, line 18: This is cool -- I did not know that it was possible to reference .Values from within values.yaml.

marktwallace (approved):
I pulled the branch and tested the deploy in my namespace. It all looks set up correctly and the pod and service configurations look right.

Good to go!

charts/zservice/charts/redis/values.yaml
mattnibs: Yeah its kind of crazy.

PR #2061, opened 2021-02-03
Windows zq: propery route - path to stdin

### LINKED ISSUE #2000: Windows error: /dev/stdin: item does not exist: file:///C:/dev/stdin
A community user spotted that the use of `-` for taking inputs causes a failure on Windows, which limits the ability to use `zq` in pipelines. Here's a simplified repro at `zq` commit `4de0485` in a Windows 10 Command Prompt.

```
C:\>type foo.ndjson
{"foo": "bar"}

C:\>zq -t foo.gz
#0:record[foo:string]
0:[bar;]

C:\>type foo.gz | zq -t -
/dev/stdin: item does not exist: file:///C:/dev/stdin
```

I couldn't find a quick answer via web search as to if this is some fundamental Windows limitation. The problem has a similar feel to what we saw in #1029, in that maybe we're seeing some difference between Windows and *nix in terms of where the the role of the shell ends and the command begins. However, it's a different symptom, and like we saw with #1029, there's other commands that don't appear to have the problem, so we may want to address it at some point.

Example:

```
C:\>type foo.ndjson | gzip - > foo.gz

C:\>dir foo.gz
 Volume in drive C has no label.
 Volume Serial Number is 0C07-A419

 Directory of C:\

01/21/2021  06:15 PM                34 foo.gz
               1 File(s)             34 bytes
               0 Dir(s)  12,282,478,592 bytes free
```

mattnibs:
The current logic in zq was translating the stdin path arg '-' in a zq command
to /dev/stdin. While this works in unix environments this does not work for
windows. Instead translate '-' to the iosrc.Stdin representation
'stdio://stdin' so the path is properly routed to stdin in windows.

Closes #2000

### DIFF
--- a/cli/inputflags/flags.go
+++ b/cli/inputflags/flags.go
@@ -10,6 +10,7 @@ import (
 	"regexp"
 
 	"github.com/brimsec/zq/cli/auto"
+	"github.com/brimsec/zq/pkg/iosrc"
 	"github.com/brimsec/zq/zbuf"
 	"github.com/brimsec/zq/zio"
 	"github.com/brimsec/zq/zio/azngio"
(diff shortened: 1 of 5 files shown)


mattnibs:
Unfortunately since we don't have system tests running for windows, we can't really add a test for this.

nwt (approved):

PR #2072, opened 2021-02-04
first step toward unifying search and expressions

mccanne:
This commit makes changes toward unifying search and expressions.
There is still much to be done, but this is a first step forward.
As part of this, we introduced a generator pattern as the low-level
construct for iterator wildcard field comparisons (e.g., "*=foo").
The first cut uses a "select" function to select and generate
sequences of values.  The syntax is a proposal to server as a
starting point for discussion.

We also eliminated "**" for now and the "=~" in favor of "in" for
CIDR matches.

There is now a scope abstraction in the compiler which sets us
up for named constants, variables, etc. as well as handling the
"$" reference for generated values.

Any aggregator can now be applied to a generated sequence.

Generated sequences can be modified or filtered with methods.
Currently there is support for just map() and filter().

The grammar has been refactored quite a bit in making this changes
and should be easier to extend and maintain.  There is one more big
move we want to make in unifying the standard-expression and
search-expression syntaxes.

In the runtime, the filter predicates are now all optimizations
of binary expressions and ahve been moved to package expr.

The where clause in group-by is now a search expression.

The match() function can be used in expression context to escape
the expression syntax into the search syntax.  Match() is boolean
valued.

The ast.BinaryExpr nodes now use lower case "and" and "or" for
the binary operator field.  This makes everything consistent now
and fixes a bug where case mismatch could occur in certain
situations.

The notes in zql/expr.md are rough and should be eventally
replaced with user docs and/or a design doc.

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -20,11 +20,6 @@ type Proc interface {
 	ProcNode()
 }
 
-// BooleanExpr is the interface implement by all AST boolean expression nodes.
-type BooleanExpr interface {
-	booleanExprNode()
-}
-
 // Identifier refers to a syntax element analogous to a programming language
 // identifier.  It is currently used exclusively as the RHS of a BinaryExpr "."
 // expression though it may have future uses (e.g., enum names or externally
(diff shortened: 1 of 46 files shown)


jameskerr:
I just wanted to record two possible sources of inspiration for the "what to call a sequence" discussion.

The elixir `stream` module : https://hexdocs.pm/elixir/Stream.html
The elixir `|>` syntax: https://elixirschool.com/en/lessons/basics/pipe-operator/

henridf (approved):
Looks good to me.

mccanne:
The brim-app tests fail on this commit so I will put a fix on brim before merging this.  The AST conventions for dotted record accesses changed so that an Identifier without a dot expression on RootRecord is okay and presumed to be this.

mccanne:
> The brim-app tests fail on this commit so I will put a fix on brim before merging this. The AST conventions for dotted record accesses changed so that an Identifier without a dot expression on RootRecord is okay and presumed to be this.

Ok, the `zq-search-expr` branch in the brim repo passes all tests against this PR.

PR #2073, opened 2021-02-05
Fix shaper bug with null input records

henridf:


expr/ztests/shape-fill.yaml
```diff
@@ -1,9 +1,17 @@
-zql: put t = fill(., "({a:{b:{a1:string}}})")
+zql: put . = fill(., "({a:{b:{c:string}}})")
```
henridf, line 1: while i was at it, i switch this test to use `.` on the LHS, now that it works.

nwt (approved):

Issue #1603, opened 2020-11-13
better string cast for time and complex types


mccanne:
expr.StringCast currently uses zng.Value.String() to format its result.  This works fine for numbers and IPs but not so good for time (which formats the result as a string integer nanoseconds) or complex types.  In the latter case, we need to create a well-defined syntax for complex literals.

mccanne:
This has been addressed...
```
echo '{a:[1,2,3]}' | zq -z "put b:=string(this),as:=string(a),ts:=string(time(0))" -
{a:[1,2,3],b:"{a:[1,2,3]}",as:"[1,2,3]",ts:"1970-01-01T00:00:00Z"}
```

PR #2077, opened 2021-02-05
handle null casts properly for all Z types

### LINKED ISSUE #2065: Unexpected output format of nulls casted to strings in ZSON
Repro is with `zq` commit `2294ca24`.

Based on my understanding of the ZSON spec, this representation of a `null` value of type `string` looks correct.

```
$ echo '{"foo": null}' | zq -f zson -
{
    foo: null (string)
} (=0)
```

However, if I use `put` to create an additional `null` field while using casting to assign `string` as its explicit type, it's rendered as a hyphen.

```
$ echo '{"foo": null}' | zq -f zson 'put bar=null:string' -
{
    foo: null (string),
    bar: "-"
} (=0)
```

If I grab the null value from the named field and apply the cast (which is unnecessary here, but let's say I wanted to be sure), it shows up as an empty string.

```
$ echo '{"foo": null}' | zq -f zson 'put bar=foo:string' -
{
    foo: null (string),
    bar: ""
} (=0)
```

mccanne:
The cast expressions are still a work in progress, but in particular,
the various casters aren't handling null properly.  Since Z allows
any value to be null, this is very easy handed as universal check
before calling the type-specific caster.  This commit adds this logic
and a simple test to cover it.

This is related to issue #1603 and fixes #2065

### DIFF
--- a/expr/eval.go
+++ b/expr/eval.go
@@ -718,18 +718,24 @@ func NewCast(expr Evaluator, styp string) (Evaluator, error) {
 		// XXX See issue #1572.   To implement aliascast here.
 		return nil, fmt.Errorf("cast to %s not implemented", styp)
 	}
-	return &evalCast{expr, c}, nil
+	return &evalCast{expr, c, typ}, nil
 }
 
 type evalCast struct {
 	expr   Evaluator
 	caster PrimitiveCaster
+	typ    zng.Type
 }
 
 func (c *evalCast) Eval(rec *zng.Record) (zng.Value, error) {
 	zv, err := c.expr.Eval(rec)
 	if err != nil {
 		return zng.Value{}, err
+	}
+	if zv.Bytes == nil {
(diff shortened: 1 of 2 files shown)


expr/eval.go
```diff
@@ -718,19 +718,25 @@ func NewCast(expr Evaluator, styp string) (Evaluator, error) {
...
 }
 
 func (c *evalCast) Eval(rec *zng.Record) (zng.Value, error) {
 	zv, err := c.expr.Eval(rec)
 	if err != nil {
 		return zng.Value{}, err
 	}
+	if zv.Bytes == nil {
+		// Take care of null here so the casters don't have to
+		// worry about it.  Anything type can be a null after all.
```
henridf, line 737:
```suggestion
		// worry about it.  Any type can be a null after all.
```

henridf (approved):

PR #2082, opened 2021-02-08
Fix panic in zqd.NewCore when returning an error

nwt:
I broke this in #2036.


alfred-landrum (approved):
This addresses a panic I saw locally when running `make test-services` without having the local postgres,redis,temporal services up.

PR #2084, opened 2021-02-08
Add duration to the implied types mentioned in the ZSON spec

philrz:
While putting together release notes and reading #2039, I noticed the ZSON spec was missed.


mccanne (approved):
oopsies

PR #2087, opened 2021-02-09
switch proc

### LINKED ISSUE #1790: Switched flowgraph
Design and implement a "switch" ZQL construct. A sketch: 

```
... | switch (
    <boolean expression> => ... | ...;
    <boolean expression> => ... | ...;
    ...
    ) | ...
```

Each incoming record is evaluated successively against the list of `<boolean expression>` and is sent into the first RHS flowgraph (after the `=>` arrow) whose expression evaluated to true. The record is dropped if all expressions are false. Output order is the same as input order.

The immediate motivation for this is generic ingest, but this will surely be a broadly useful construct.

henridf:
close #1790


compiler/proc.go
```diff
@@ -277,6 +286,35 @@ func compileParallel(custom Hook, pp *ast.ParallelProc, c *proc.Context, scope *
...
+			sc := switcher.NewProc(f)
+			parents = append(parents, sc)
+		}
+	}
+	if len(parents) != n {
+		return nil, fmt.Errorf("proc.compileSwitch: %d parents for switch proc with %d branches", len(parents), len(pp.Cases))
+	}
+	var procs []proc.Interface
+	for k := 0; k < n; k++ {
+		proc, err := Compile(custom, pp.Cases[k].Proc, c, scope, []proc.Interface{parents[k]})
```
nwt, line 309:
```suggestion
	for _, p := range parents {
		proc, err := Compile(custom, pp.Cases[k].Proc, c, scope, []proc.Interface{p})
```

proc/switchproc/switch.go
```diff
@@ -0,0 +1,166 @@
...
+	p.reqCh <- request{p.id, p.ch}
+	result := <-p.ch
+	if proc.EOS(result.Batch, result.Err) {
+		p.Done()
+	}
+	return result.Batch, result.Err
+}
+
+func (p *Proc) Done() {
+	// Signal to our parent SWitchProc that this path is done by
```
nwt, line 157:
```suggestion
	// Signal to our parent SwitchProc that this path is done by
```

proc/switchproc/switch.go
```diff
@@ -0,0 +1,166 @@
...
+	// sending a request with a nil channel object.  We go ahead
+	// and mark the Proc done by setting it's channel to nil in
```
nwt, line 159:
```suggestion
	// and mark the Proc done by setting its channel to nil in
```

proc/switchproc/switch.go
```diff
@@ -0,0 +1,166 @@
+package switchproc
```
nwt, line 1: Maybe call this switcher so it matches the name of the type returned by New?

mccanne: The convention in the proc packages has been 
```
package foo 

type Proc struct {
...
}

func New() *Proc {
}
```
So, foo.New() creates a *foo.Proc

henridf: @mccanne @nwt two things at play here:

- re the convention on package name, this package should be called `switch`, but I had to call it `switchproc` due to `switch` being a reserved word.
- re the `Switcher` type, this is modelled on the existing `split.go` file, which has a `type Splitter`. There is a `type Proc` below in this file, just like in `split.go`.

mccanne: my bad

nwt: I think it's confusing for `switchproc.New` to return a `Switcher`, and I think we should fix that, either by changing the package name to `switcher` or changing the constructor name to `NewSwitcher`, though I'm fine saving the change for another PR.

What's going on in split.go is also confusing, and I think we should also fix that in a separate PR.

henridf: @nwt if we're going down this path... I think naming-wise this should all derive from the AST names, where we have SwitchProc (like PArallelProc, SequentialProc, etc). Assuming those AST names are staying, then I would keep the package name `switchproc` and rename type Switcher to Proc (per @mccanne's point above). What do you think?

nwt: Sorry for my tardy response!

Renaming `Switcher` to `Proc` doesn't feel quite right because it doesn't implement `proc.Interface`, which is a characteristic of all the other types named `Proc`.

henridf: @nwt ok - done in b6c7e3bfe9fbb9db8f7508ee70ff6b757cbf5bf2

nwt (approved):

Issue #2090, opened 2021-02-09
Z function for splitting strings into an array


philrz:
The Z equivalent of the Python:

```
$ python
>>> 'www.brimsecurity.com'.split('.')
['www', 'brimsecurity', 'com']
```

philrz:
Verified in `zq` commit `1471b759`.

```
$ echo '{"query": "www.brimsecurity.com"}' | zq -f ndjson 'put parts=split(query, ".")' -
{"parts":["www","brimsecurity","com"],"query":"www.brimsecurity.com"}
```

PR #2098, opened 2021-02-10
add join/split string functions

### LINKED ISSUE #2089: Z function for joining array contents into strings
The Z equivalent of the Python:

```
$ python
>>> '.'.join(['www', 'brimsecurity', 'com'])
'www.brimsecurity.com'
```

mccanne:
This commit adds functions for join() and split() with respect to strings.
Since Z is dynamically typed, this functions may do more down the road 
based on the types of their arguments, but for now they are just string
functions.

Closes #2089, #2090

### DIFF
--- a/expr/function/string.go
+++ b/expr/function/string.go
@@ -297,7 +297,7 @@ func (j *join) Call(args []zng.Value) (zng.Value, error) {
 	if !ok {
 		return zng.NewErrorf("argument to join() is not an array"), nil
 	}
-	if !zng.IsStringy(zng.AliasedType(typ.Type).ID()) {
+	if !zng.IsStringy(typ.Type.ID()) {
 		return zng.NewErrorf("argument to join() is not a string array"), nil
 	}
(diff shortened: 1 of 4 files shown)


expr/function/string.go
```diff
@@ -245,3 +247,87 @@ func (*trim) Call(args []zng.Value) (zng.Value, error) {
...
+	builder strings.Builder
+}
+
+func (j *join) Call(args []zng.Value) (zng.Value, error) {
+	zsplits := args[0]
+	typ, ok := zng.AliasedType(zsplits.Type).(*zng.TypeArray)
+	if !ok {
+		return zng.NewErrorf("argument to join() is not an array"), nil
+	}
+	if !zng.IsStringy(zng.AliasedType(typ.Type).ID()) {
```
nwt, line 300: ID resolves aliases so you don't really need to call AliasedType.

nwt (approved):

PR #2100, opened 2021-02-10
add slice expression to Z language and runtime

### LINKED ISSUE #2099: make slice more efficient
This slice expression currently copies the zng.Values of the array into a second buffer.  This code could be a big more efficient by simply identify the slice boundary in the input zcode.Bytes and using the byte slice pointing into this buffer for the output value.

mccanne:
This commit adds slice expressions to Z.  It currently works only for 
arrays and will be enhanced in a future PR to work on other data types
like strings, bytes, sequences, etc.

Close #2099


expr/ztests/slice.yaml
```diff
@@ -0,0 +1,14 @@
+zql: |
+   cut a1=a[1:-1],a2=a[1:],a3=a[:1],a4=a[:-1],a5=a[:-100],a6=a[-1:]
```
henridf, line 2:
```suggestion
   cut a1=a[1:-1],a2=a[1:],a3=a[:1],a4=a[:-1],a5=a[:-100],a6=a[-1:], a7=[:a[0]]
```
(or some other additive expr in the brakcets)

henridf (approved):

expr/slice.go
```diff
@@ -0,0 +1,95 @@
...
+		if k < int(from) {
+			continue
+		}
+		if k >= int(to) {
+			break
+		}
+		b = zcode.AppendAs(b, container, bytes)
+	}
+	s.bytes = b
+	return zng.Value{elem.Type, b}, nil
```
nwt, line 94: Maybe take care of #2099 now?
```suggestion
	bytes := elem.Bytes
	it := bytes.Iter()
	for k := 0; k < to && !it.Done(); k++ {
		if k == from {
			bytes = zcode.Bytes(it)
		}
		if _, _, err := it.Next(); err != nil {
			return zng.Value{}, err
		}
	}
	return zng.Value{elem.Type, bytes[:len(bytes)-len(it)]}, nil
```

nwt (approved):

PR #2105, opened 2021-02-10
make connection.Search easier to use

alfred-landrum:
connection.Search is a convenience wrapper around connection.SearchRaw. This change makes it a little easier to use, and adds example usage. It roundtrips with the server to parse the search query, to be more resilient against AST changes until we add support for AST versioning.


mattnibs (approved):

mccanne:
Sweet!  Not at all a blocker but it would be nice to not have to pass in a context.Context for the casual experimenter like the pattern in golang for http where you do SomethingWithContext() when you want to pass in your own context and Something() doesn't need a context.

PR #2112, opened 2021-02-11
Fix IPv6 literal parsing in Z language

### LINKED ISSUE #2111: Error parsing IPv6 (correlated with added slice support in 65a58a4)
We've had recent failures in the Brim integration tests. After narrowing down the root cause of the first observed failure, it's correlated with the slice support added in #2100 (cc: @mccanne).

Here's a simplified repro:

```
$ cat scalar.zson 
{
    scalar: fe80::58d2:2d09:e8cb:a8ad
} (=0)

$ zq -version
Version: v0.29.0-12-g65a58a42

$ zq -f zson -i zson 'scalar=fe80::58d2:2d09:e8cb:a8ad' scalar.zson 
parse error: 1 error(s) decoding:

* 'procs[0].filter.rhs.value' expected type 'string', got unconvertible type '[]interface {}'
```

This failure did not occur in the `zq` commit that came right before.

```
$ zq -f zson -i zson 'scalar=fe80::58d2:2d09:e8cb:a8ad' scalar.zson 
parse error: 1 error(s) decoding:

* 'procs[0].filter.rhs.value' expected type 'string', got unconvertible type '[]interface {}'
```

nwt:
Fixes #2111.

### DIFF
--- a/zql/ztests/ipv6.yaml
+++ b/zql/ztests/ipv6.yaml
@@ -1,0 +1,7 @@
+zql: fe80::58d2:2d09:e8cb:a8ad
+
+input: &input |
+  #0:record[a:ip]
+  0:[fe80::58d2:2d09:e8cb:a8ad;]
+
+output: *input
--- a/zql/ztests/ipv6.yaml
+++ b/zql/ztests/ipv6.yaml
@@ -1,7 +1,10 @@
-zql: fe80::58d2:2d09:e8cb:a8ad
+zql: 'fe80::58d2:2d09:e8cb:a8ad OR ::58d2:2d09:e8cb:a8ad OR 2d09:e8cb:a8ad:: OR ::'
 
 input: &input |
   #0:record[a:ip]
   0:[fe80::58d2:2d09:e8cb:a8ad;]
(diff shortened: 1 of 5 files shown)


zql/ztests/ipv6.yaml
```diff
@@ -0,0 +1,7 @@
+zql: fe80::58d2:2d09:e8cb:a8ad
+
+input: &input |
+  #0:record[a:ip]
+  0:[fe80::58d2:2d09:e8cb:a8ad;]
```
alfred-landrum, line 5: How about a test case for `::`?

zql/ztests/ipv6.yaml
```diff
@@ -0,0 +1,7 @@
...
+
+output: *input
```
alfred-landrum, line 7: nit: newline

zql/ztests/ipv6.yaml
alfred-landrum: (I meant a zql search for `::`, not adding a record with `::`)

alfred-landrum (approved):
I was able to successfully run the Brim integration tests with this commit.

PR #2117, opened 2021-02-12
s3io: Use s3iface.S3API over *aws.Config

mattnibs:
Currently the s3io code path requires *aws.Config for all its calls and creates
a new client for each request from the passed configuration. This is not
a totally expense free operation. I recently did a cpu profile of lake walk
and saw that having AWS_SDK_LOAD_CONFIG=true results in a OpenFile syscall
everytime s3io.newClient is called; meaning we read from the file system
everytime a call to s3 is made (twice for calls that need head first).

Instead have all s3io functions accept a s3iface.S3API interface. For the
default iosrc.S3Source use the go init() hook to call s3io.NewClient once at
the start of a program.

### DIFF
--- a/pkg/iosrc/s3.go
+++ b/pkg/iosrc/s3.go
@@ -8,8 +8,8 @@ import (
 	"net/http"
 	"time"
 
-	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/s3/s3iface"
 	"github.com/brimsec/zq/pkg/s3io"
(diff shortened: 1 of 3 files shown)


pkg/iosrc/s3.go
```diff
@@ -18,17 +18,21 @@ var defaultS3Source = &s3Source{}
 var _ Source = defaultS3Source
 var _ ReplacerAble = defaultS3Source
 
+func init() {
+	defaultS3Source.Client = s3io.NewClient(nil)
```
alfred-landrum, line 22: I believe you can move this into the declaration of `defaultS3Source` and remove the init.

alfred-landrum (approved):

PR #2119, opened 2021-02-12
s3io.ReadFile

mattnibs:
Add s3io.ReadFile which uses s3manager.Downloader to download the Object. For
downloading entire files this has several advantages over s3io.NewReader:

1. s3manager.Downloader doesn't require HEAD api request to get the size of the
   entire object.

2. s3manager can use concurrency to download Objects faster.

### DIFF
--- a/pkg/iosrc/s3.go
+++ b/pkg/iosrc/s3.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"io"
-	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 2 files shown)


mattnibs:
I was able to run `zar stat` on the corelight data set imported into an s3 bucket about 3 seconds faster on this branch compared to master (10 seconds vs 7 seconds).

alfred-landrum (approved):

PR #2121, opened 2021-02-13
Tolerate missing column during cut to root field

### LINKED ISSUE #2120: Better handling of missing column in `cut .=id`
https://github.com/brimsec/zq/issues/1931#issuecomment-778487366 includes an example where `zq` commit `871df67c` quickly exited when confronted with a record that lacked the named field.

```
$ zq -t "cut .=id" all.zng
field is not present
```

We have open issues #1429 and #1982 that both track ways this kind of failure could be improved, but @nwt also pointed out that he'd want to have better handled the case of a missing column in his fix for #1931, hence this issue.

nwt:
Cutting to the root field (e.g, cut .=f) produces a fatal error when it
encounters a record missing the source column.  Ignore the record
instead to match the behavior when cutting to non-root fields.

Closes #2120.

### DIFF
--- a/expr/ztests/cut-to-root-warning.yaml
+++ b/expr/ztests/cut-to-root-warning.yaml
@@ -1,0 +1,10 @@
+zql: 'cut .=c'
+
+input: |
+  #0:record[a:record[b:int32]]
+  0:[-;]
+  0:[[-;]]
+  0:[[1;]]
+
+warnings: |
+  cut: no record found with columns c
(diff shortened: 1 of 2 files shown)


expr/ztests/cut-to-root-warning.yaml
```diff
@@ -0,0 +1,10 @@
+zql: 'cut .=c'
+
+input: |
+  #0:record[a:record[b:int32]]
+  0:[-;]
+  0:[[-;]]
+  0:[[1;]]
+
+warnings: |
+  cut: no record found with columns c
```
henridf, line 10: (I know it's pre-existing, but I'm not sure I would understand that message.)

henridf (approved):

Issue #2122, opened 2021-02-13
json-strict mode for zson


mccanne:
When zson is used to parse ndjson, a strict mode should be allowed that preserves json types like float64 (instead of assuming that numbers appearing as integers has implied-type int64 ).  This should also create a canonical order of json column order since the column order is important in zson but not json.  This could be controlled by a separate config knob from strict as its also reasonable to create different record types with different column orders though shaping gets messy then.

mccanne:
This goes with #1810 when we make `-i ndjson` use the zson reader.

henridf:
> different record types with different column orders though shaping gets messy then.

Did you have anything specific in mind wrt shaping getting messy in this situation?

henridf:
Btw, I would suggest calling this "JSON transparency" mode.

mccanne:
Raising priority on this as JSON use cases affected by this when you end up with unnecessarily complex unions.

mccanne:
Now that we're back to reading JSON natively instead of using the ZSON, this is lower priority.  This still might be a useful mode for the ZSON reader but not a high priority for now...

philrz:
@mccanne, @nwt and I discussed this one again in the context of the use case described in #4777. We agreed that probably the best way to improve on what's described there is to adopt this "strict mode" proposal when reading numbers in JSON input such that they're all treated as float even if they are formatted as integers in the input data. This would become the new default behavior when reading JSON input, with the current behavior then made available via a new flag.

PR #2123, opened 2021-02-13
add zson to auto-detector

### LINKED ISSUE #2029: add zson to auto-detector

mccanne:
This commit adds zson to the auto detector and arranges for zson
to parse ndjson when auto-detected, unless a legacy ndjson type
config is provided, in which case the ndjson parser is used.
This allows us to support legacy type configs while we update
the user base with the new Z-shaper approach to ingesting json.

One ramification is that the zson parser converts numbers that
are textually an integer to int64 instead of json's float64.
We plan to add a json-strict mode to the zson parser, which
would allow us to circumvent this though this would be an explicit
flag and not inferred by default in the autodetection path.
See issue #2122.

Another ramification is that the ndjson parser assume null:string
for unknown nulls whereas zson assumes null:null.  Now that we have
shaping and type algebra, this later approach is preferred because
the shaper can know from null:null that a type is indeterminate
compared to null:string, where the type was arbitrarily chosen
to be string.

Closes #2029


henridf (approved):

philrz:
I've only recently noticed a side-effect of something that was anticipated in the PR description above:

> ...the ndjson parser assumed null:string  
> for unknown nulls whereas zson assumes null:null. Now that we have  
> shaping and type algebra, this latter approach is preferred because  
> the shaper can know from null:null that a type is indeterminate  
> compared to null:string, where the type was arbitrarily chosen  
> to be string.

Therefore, the following was working as of Zed commit 99713cb right before the changes merged in this PR:

```
$ zq -version
Version: v0.29.0-25-g99713cbc

$ echo '{"foo": []}' | zq -f zeek -
#separator \x09
#set_separator	,
#empty_field	(empty)
#unset_field	-
#fields	foo
#types	vector[string]
(empty)
```

And as the comment indicates, that's because the internal representation after reading it used the Zed `string` type:

```
$ echo '{"foo": []}' | zq -f zson -
{
    foo: [] (0=([string]))
} (=1)
```

Whereas starting with Zed commit d89b612 (associated with this PR) and up to today, it's now a fatal error. The error message echoes the comment above.

```
$ zq -version
Version: v0.29.0-26-gd89b6124

$ echo '{"foo": []}' | zq -f zeek -
type null: type cannot be represented in zeek format

$ echo '{"foo": []}' | zq -f zson -
{
    foo: [] (0=([null]))
} (=1)
```

This was a simplified example, but this same problem prevents being able to read the entirety of unshaped [Zeek NDJSON zed-sample-data](https://github.com/brimdata/zed-sample-data/tree/main/zeek-ndjson) and write it out as Zeek TSV, since the `ssl` log happens to contain such empty arrays.

```
~/work/zed-sample-data/zeek-ndjson$ zq -f zeek ssl.ndjson.gz 
type null: type cannot be represented in zeek format
```

I talked this one through with @nwt to debate whether it would make sense to restore the assumed `string` type when writing out Zeek TSV in these situations, but for now he's convinced me to leave things as they are. Per [https://github.com/brimdata/zed/blob/main/zeek/Shaping-Zeek-NDJSON.md,](https://github.com/brimdata/zed/blob/main/zeek/Shaping-Zeek-NDJSON.md,) our main recommendation for Zeek NDJSON users is that they apply shapers wherever possible (such as our reference one shown there, which will hopefully be useful as is for most users, and workable for others with minimal modification) since indeed that would have addressed the problem with the `ssl` log since it would have assigned a specific non-null data type at read time. Also, I'm assuming this same issue will not come up for non-Zeek NDJSON users, since I doubt users would take other arbitrary data sources and decide to output it as Zeek TSV (though if this comes up a lot, maybe we'd want to rethink this). Finally, we did poll the community at one time about use of `-f zeek`, and on the whole seems to be minimally used. I'd hope that _most_ the users who _do_ make use of it would also be _reading_ Zeek TSV, which once again would not be exposed to this since the specific data types would appear at read time (see https://github.com/brimdata/zed/blob/main/zeek/Data-Type-Compatibility.md).

PR #2131, opened 2021-02-14
add "function proc", exists(), and fields() and drop parens from types

mccanne:
(Sorry for the mixture of things going on in this PR.  They all dovetail and sort of
go together logically and I need them all on a subsequent branch I'm working on 
so I left them together in this PR.)

This commit adds the notion of a "function proc" which is a FunctionCall
ast node appearing as a proc.  This is used to represent the ambiguity
between a simple reducer like "count()" used without an assignment and
a boolean function like "is('{s:string}')".  The ambiguity is resolved by
the compiler by looking up the name and choosing either a GroupByProc
or a FilterProc to form the appropriate representation.  To do this,
there is now a SemanticTransform() step in package compiler.  This replaces
compiler.ReplaceGroupByProcDurationWithKey() and does this key transform
along with any other semantic transformation we need to do post parsing.

We also added an option to expression functions to ask that the root
record be passed to them.  This is used by "is()" to do type matching
on "." or an explicit second argument.

We added the is(), fields(), and exists() functions.  is()
provides type comparisons and fields() returns an array of dotted-field
strings representing the fields in a type value (that is a record type)
or a record value.  This latter function is useful during data shape
exploration to layout the fields encountered as arrays strings and
cut down on the noise.

The exists() function returns true iff all of it's arguments evaluate
to a non-missing, non-error value.  e.g., exists(foo) is true if the
field foo exists and exists(foo / 0) is always false.  exists(a.b,a.c)
is true if both fields exist unless their value is an error value.

Parens are dropped from type expressions.  We're heading toward a model 
where parens are used to imply a type as an expression and clarify typedefs 
in ZSON but are not part of the Z  type syntax itself.

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -89,6 +89,13 @@ type ConditionalExpression struct {
 	Else      Expression `json:"else"`
 }
 
+// A FunctionCall represents different things dependending on its context.
+// As a proc, it is either a group-by with no group-by keys and no duration,
+// or a filter with a function that is boolean valued.  This is determined
+// by the compiler rather than the syntax tree based on the specific functions
+// and aggregators that are defined at compile time.  In expression context,
+// a function call has the standard semantics where it takes one or more arguments
+// and returns a result.
 type FunctionCall struct {
 	Op       string       `json:"op"`
 	Function string       `json:"function"`
@@ -315,13 +322,7 @@ func (*PutProc) ProcNode()        {}
 func (*RenameProc) ProcNode()     {}
 func (*FuseProc) ProcNode()       {}
 func (*JoinProc) ProcNode()       {}
-
-// A FunctionCall can also represent a proc that is either a group-by
-// with no group-by keys and no duration or a filter with a function
-// that is boolean valued.  This is determined by the compiler rather
-// than the syntax tree based on the specific functions and aggregators
-// tha are defined at compile time.
-func (*FunctionCall) ProcNode() {}
+func (*FunctionCall) ProcNode()   {}
 
 // A Reducer is an AST node that represents a reducer function.  The Operator
 // field indicates the aggregation method while the Expr field indicates
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -316,6 +316,13 @@ func (*RenameProc) ProcNode()     {}
 func (*FuseProc) ProcNode()       {}
 func (*JoinProc) ProcNode()       {}
 
+// A FunctionCall can also represent a proc that is either a group-by
+// with no group-by keys and no duration or a filter with a function
+// that is boolean valued.  This is determined by the compiler rather
+// than the syntax tree based on the specific functions and aggregators
+// tha are defined at compile time.
+func (*FunctionCall) ProcNode() {}
+
 // A Reducer is an AST node that represents a reducer function.  The Operator
 // field indicates the aggregation method while the Expr field indicates
 // an expression applied to the incoming records that is operated upon by them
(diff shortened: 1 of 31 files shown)


ast/ast.go
```diff
@@ -316,6 +316,13 @@ func (*RenameProc) ProcNode()     {}
 func (*FuseProc) ProcNode()       {}
 func (*JoinProc) ProcNode()       {}
 
+// A FunctionCall can also represent a proc that is either a group-by
```
henridf, line 319:
```suggestion
// A FunctionCall represents a proc that is either a group-by
```

ast/ast.go
```diff
@@ -316,6 +316,13 @@ func (*RenameProc) ProcNode()     {}
...
+// with no group-by keys and no duration or a filter with a function
+// that is boolean valued.  This is determined by the compiler rather
+// than the syntax tree based on the specific functions and aggregators
+// tha are defined at compile time.
```
henridf, line 323:
```suggestion
// that are defined at compile time.
```

henridf:
`is(x, y)` is just a shortcut for `typeof(x) = y`?

henridf:
> Parens are dropped from type expressions. We're heading toward a model
where parens are used to imply a type as an expression and clarify typedefs
in ZSON but are not part of the Z type syntax itself.

🎉 ⬆️ 


A few comments/questions on the new functions...

- It looks like `is(x, y)` is the same thing as `typeof(x) = y`? if so, do we need it?
- `exists()` conflates two distinct concepts (presence of a field in a type vs expression evaluation error). This seemed confusing to me when I looked at the yaml test to figure out what it does. Maybe I just need to get used to it.
- Unless I missed it there's no test for `fields()`...

ast/ast.go
mccanne: That comment was explaining why it was a method on procNode but I moved it to the site of the type definition and clarified the comment.

mccanne:
> * It looks like `is(x, y)` is the same thing as `typeof(x) = y`? if so, do we need it?

Yes, here is my logic: I found myself typing `typeof(.)=val` and like typing `is(val)` instead.  It also has a place to put caching state so you don't need to do a call to x.Type.ZSON() for every comparison... down the road we will have optimization that makes this move and that optimizer would create precisely this `is()` gadget in the runtime so we'll need it anyway and might as well expose it now.  And when you do a SQL FROM, the way we get the table is a type search so the `is()` gadget will be important here.  (This will also work really nicely with ZST, where the columnar data is arranged by top-level type value so "table scans" of a common scheme will work really well.)

mccanne:
> * `exists()` conflates two distinct concepts (presence of a field in a type vs expression evaluation error). This seemed confusing to me when I looked at the yaml test to figure out what it does. Maybe I just need to get used to it.

I started out by adding `has()` but poked around and found this...

https://docs.aws.amazon.com/qldb/latest/developerguide/ql-functions.exists.html

It returns false for zero values but I otherwise liked this.  We have `iserr()` already and I think this is a fine bike-shedding topic. 

In practice, I think the conflation is what you want because in Z you want to evaluate things that evaluate and skip things that don't evaluate and just one of the reasons things do evaluate is MISSING.

I propose we bikeshed on a small kernel of functions that is ergonomic and covers all these corner cases.  In the meantime, I suggest we merge `exists()` while we work this out (soon) and replace it.  (p.s. I'm a little bit blocked on this PR.)

henridf (approved):

mccanne:
I'm adding the test for `fields`...

henridf:
Didn't realize these were bikeshedding questions... approved.

mccanne:
> Didn't realize these were bikeshedding questions... approved.

I meant "bikeshedding" in a positive way :-).  Your feedback is really god and I agree this an important discussion.

PR #2140, opened 2021-02-16
Fix cast() function on null values

henridf:


expr/ztests/shape-null-cast.yaml
```diff
@@ -0,0 +1,11 @@
+zql: |
+   put . = shape(., "{a:string,b:string}")
+
+
```
nwt, line 4:
```suggestion

```

expr/shaper.go
```diff
@@ -87,6 +87,10 @@ func createOp(in, out *zng.TypeRecord) (op, error) {
 
 func (s op) castPrimitive(in zcode.Bytes, b *zcode.Builder) {
 	pc := LookupPrimitiveCaster(s.castTypes.to)
+	if in == nil {
+		b.AppendNull()
+		return
+	}
 	v, err := pc(zng.Value{s.castTypes.from, in})
```
nwt, line 94: NIt: Keep `pc`'s definition close to its use.
```suggestion
	if in == nil {
		b.AppendNull()
		return
	}
	pc := LookupPrimitiveCaster(s.castTypes.to)
	v, err := pc(zng.Value{s.castTypes.from, in})
```

nwt (approved):

expr/ztests/shape-null-cast.yaml
```diff
@@ -0,0 +1,9 @@
+zql: |
+   put . = shape(., "{a:string,b:string}")
+input: |
```
nwt, line 3:
```suggestion
   put . = shape(., "{a:string,b:string}")

input: |
```

mccanne (approved):
oops

PR #2142, opened 2021-02-16
zson: add support for parsing map types

### LINKED ISSUE #2141: zson parser can't parse map type
Encountered while working on #2038 

```
$ echo '{"a": null}' | zq -t 'put . = shape(., "{m:|{string,int32}|}")' -
rhs of assigment expression: shaper could not parse type value literal: unknown type in Analyzer.convertType: *ast.TypeMap
```

henridf:
fix #2141

### DIFF
--- a/zson/analyzer.go
+++ b/zson/analyzer.go
@@ -590,6 +590,8 @@ func (a Analyzer) convertType(zctx *resolver.Context, typ ast.Type) (zng.Type, e
 			return nil, err
 		}
 		return zctx.LookupTypeSet(typ), nil
+	case *ast.TypeMap:
+		return a.convertTypeMap(zctx, t)
 	case *ast.TypeUnion:
 		return a.convertTypeUnion(zctx, t)
 	case *ast.TypeEnum:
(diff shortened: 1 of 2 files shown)


zson/analyzer.go
```diff
@@ -617,6 +619,19 @@ func (a Analyzer) convertTypeRecord(zctx *resolver.Context, typ *ast.TypeRecord)
...
+	keyType, err := a.convertType(zctx, tmap.KeyType)
+	if err != nil {
+		return nil, err
+	}
+	valType, err := a.convertType(zctx, tmap.ValType)
+	if err != nil {
+		return nil, err
+	}
+
+	return zctx.LookupTypeMap(keyType, valType), nil
```
nwt, line 632:
```suggestion
	}
	return zctx.LookupTypeMap(keyType, valType), nil
```

nwt (approved):

zson/ztests/container-types.yaml
```diff
@@ -0,0 +1,22 @@
+script: |
+  zq -t -i zson "by typeof(.) | sort ." in.zson
```
nwt, line 2: Now that #2123 has been merged, can this be a zql test instead of a script test?

nwt (approved):

zson/ztests/container-types.yaml
henridf: it can, done. (looks like there's a number of pre-existing tests that can be similarly changed).

PR #2148, opened 2021-02-16
Parallelize chunk metadata fetch

mattnibs:
Fetching the chunk metadata (particularly when the data hasn't been catched)
can be time consuming when doing a walk operation on a ts dir. Speed things up
by using a separate go routine to fetch each chunk's metadata in a ts dir.

Part of #2017

### DIFF
--- a/ppl/lake/walk.go
+++ b/ppl/lake/walk.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"sort"
+	"sync"
 	"time"
 
 	"github.com/brimsec/zq/pkg/iosrc"
@@ -12,6 +13,7 @@ import (
 	"github.com/brimsec/zq/zbuf"
 	"github.com/brimsec/zq/zqe"
 	"github.com/segmentio/ksuid"
+	"golang.org/x/sync/errgroup"
 )
 
 const (
@@ -79,10 +81,11 @@ func tsDirVisit(ctx context.Context, lk *Lake, filterSpan nano.Span, visitor tsD
 		if err != nil {
 			return err
 		}
-		chunks, err := tsDirEntriesToChunks(ctx, lk, filterSpan, d, dirents)
+		chunks, err := tsDirEntriesToChunks(ctx, lk, d, dirents)
 		if err != nil {
 			return err
 		}
+		chunks = chunks.Overlapping(filterSpan)
 		if err := visitor(d, chunks); err != nil {
 			return err
 		}
@@ -90,7 +93,7 @@ func tsDirVisit(ctx context.Context, lk *Lake, filterSpan nano.Span, visitor tsD
 	return nil
 }
 
-func tsDirEntriesToChunks(ctx context.Context, lk *Lake, filterSpan nano.Span, tsDir tsDir, entries []iosrc.Info) ([]chunk.Chunk, error) {
+func tsDirEntriesToChunks(ctx context.Context, lk *Lake, tsDir tsDir, entries []iosrc.Info) (chunk.Chunks, error) {
 	type seen struct {
 		data bool
 		meta bool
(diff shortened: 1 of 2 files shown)


ppl/lake/walk.go
```diff
@@ -111,32 +114,40 @@ func tsDirEntriesToChunks(ctx context.Context, lk *Lake, filterSpan nano.Span, t
...
+				return err
+			}
 
-		md, err := chunk.UnmarshalMetadata(b, lk.DataOrder)
-		if err != nil {
-			if zqe.IsNotFound(err) {
-				continue
+			md, err := chunk.UnmarshalMetadata(b, lk.DataOrder)
+			if err != nil {
+				if zqe.IsNotFound(err) {
```
alfred-landrum, line 137: This 'is not found' check looks funny right here. I know that's not the intent of this PR, but I think it's supposed to happen after the ReadFile?

mattnibs: Yep! Will Fix.

mattnibs:
I did a quick perf run on this. Not as impressive as I would have hoped. I ingested the corelight dataset into s3 with a thresh of 1MiB. This resulted in the average tsdir having around 5 chunks per.

For a benchmark I tested the time to it took to run `zar stat` on the ingested lake.
- On my macbook with my slowish internet master ran at around 27 secs while this branch ran at 20 seconds
- On a t2.medium (2 virtual cpus) ec2 instance there was no distinguishable difference between the two branches
- On a t2.2xlarge (8 cps) ec2 instance master ran at ~ 5 seconds and this branch was ~ 4 seconds.

Not a huge difference. I'm guessing we'll need more chunks in a tsdir to see a greater benefit.

alfred-landrum (approved):

ppl/lake/walk.go
```diff
@@ -111,32 +114,40 @@ func tsDirEntriesToChunks(ctx context.Context, lk *Lake, filterSpan nano.Span, t
...
+			md, err := chunk.UnmarshalMetadata(b, lk.DataOrder)
+			if err != nil {
+				return err
 			}
-			return nil, fmt.Errorf("failed to read chunk metadata from %v: %w", mdPath, err)
-		}
-		chunk := md.Chunk(dir, id)
-		if !filterSpan.Overlaps(chunk.Span()) {
-			continue
-		}
```
nwt, line 136: Nit: I think checking for overlap here rather than in the tsDirEntriesToChunks caller seems like a better approach, mostly because it's less code.

ppl/lake/walk.go
```diff
@@ -111,32 +114,40 @@ func tsDirEntriesToChunks(ctx context.Context, lk *Lake, filterSpan nano.Span, t
...
+
+			mu.Lock()
+			chunks = append(chunks, md.Chunk(dir, id))
+			mu.Unlock()
+			return nil
+		})
 	}
-	return chunks, nil
+
+	return chunks, group.Wait()
```
nwt, line 150: This seems like a race since `chunks` will be evaluated before `group.Wait()`.
```suggestion
	err := group.Wait()
	return chunks, err
```

ppl/lake/walk.go
mattnibs: I think it's easier to reason about tsDirEntriesToChunks if it literally just returns the chunks in a tsdir. Passing in filterSpan felt like cramming too much functionality into this function. Plus I consider any functionality we can move out walk.go and into other packages a win.

ppl/lake/walk.go
mattnibs: Good point! Considering it is possible for errgroup to return Wait before all go routines finish execution (in the case of a go routine returning an error, or the context getting cancelled), it's probably best to do this:
```suggestion
        if err := group.Wait(); err != nil {
            return nil, err
        }
        return chunks, nil
```

ppl/lake/walk.go
nwt: Fair enough!

ppl/lake/walk.go
nwt: Either approach seems OK to me.

In general, I think it's fine to return a non-nil value along with a non-nil error under the assumption that callers will ignore the value unless the error is nil or the documentation otherwise (as with io.Reader.Read).

> Considering it is possible for errgroup to return Wait before all go routines finish execution (in the case of a go routine returning an error, or the context getting cancelled),

This isn't quite right. `group.Wait` won't return until all functions passed to `group.Go` have returned.

ppl/lake/walk.go
```diff
@@ -111,30 +114,43 @@ func tsDirEntriesToChunks(ctx context.Context, lk *Lake, filterSpan nano.Span, t
...
+
+			mu.Lock()
+			chunks = append(chunks, md.Chunk(dir, id))
+			mu.Unlock()
+			return nil
+		})
+	}
+
+	// Because group.Wait can return an error before all go routines finish,
+	// check for error before returning chunks to avoid a race condition.
```
nwt, line 151:
```suggestion
	// Wait for goroutines before accessing chunks to avoid a race condition.
```

nwt (approved):

ppl/lake/walk.go
mattnibs: Yep you are right. In that case i'll write it as you suggested.

PR #2154, opened 2021-02-17
clean up conventions for missing values

### LINKED ISSUE #2130: simplify and unify missing value implementation
ErrNoSuchField should be replaced everywhere with the missing value and expressions in general should return the missing error value when they fail (instead of failing the computation with an error which should only be used for fatal errors like system bugs, I/O errors, etc)

To do this we should create `zng.ErrMissing` and the missing value inside the package zng like this:
```
var ErrMissing = errors.New("missing value")
var Missing = NewError(ErrMissing)
```
Then we should change all references to ErrNoSuchField and make sure the handling of missing values is consistent through the expression and proc runtime.

mccanne:
This commit creates zng.ErrMissing and zng.Missing to work toward
a unified way of handling missing values (i.e., expressions that
refer to fields that don't exist).  ErrMissing is a Go error
that can be returned up the chain to signal this condition while
Missing is a value that can be returned in zng data when a
result calls for embedding a Missing value instead of ignoring
it all together.

Closes #2130

### DIFF
--- a/expr/cutter.go
+++ b/expr/cutter.go
@@ -3,8 +3,9 @@ package expr
 import (
 	"errors"
 	"fmt"
-	"github.com/brimsec/zq/zcode"
 	"strings"
+
+	"github.com/brimsec/zq/zcode"
 
 	"github.com/brimsec/zq/field"
 	"github.com/brimsec/zq/zng"
@@ -76,7 +77,7 @@ func (c *Cutter) Apply(in *zng.Record) (*zng.Record, error) {
 	if len(c.fieldRefs) == 1 && c.fieldRefs[0].IsRoot() {
 		zv, err := c.fieldExprs[0].Eval(in)
 		if err != nil {
-			if err == ErrNoSuchField {
+			if err == zng.ErrMissing {
 				return nil, nil
 			}
 			return nil, err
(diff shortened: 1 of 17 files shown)


expr/expr_test.go
```diff
@@ -143,7 +143,7 @@ func TestPrimitives(t *testing.T) {
 	testSuccessful(t, "s", record, zstring("hello"))
 
 	// Test bad field reference
-	testError(t, "doesnexist", record, expr.ErrNoSuchField, "referencing non-existent field")
+	testError(t, "doesnexist", record, zng.ErrMissing, "referencing non-existent field")
```
nwt, line 146:
```suggestion
	testError(t, "doesnexist", record, zng.ErrMissing, "referencing nonexistent field")
```

zng/error.go
```diff
@@ -7,6 +7,9 @@ import (
 	"github.com/brimsec/zq/zcode"
 )
 
+var ErrMissing = errors.New("missing")
+var Missing = NewError(ErrMissing)
```
nwt, line 11: Mabye add doc comments?

expr/cutter.go
```diff
@@ -3,9 +3,10 @@ package expr
 import (
 	"errors"
 	"fmt"
-	"github.com/brimsec/zq/zcode"
 	"strings"
 
+	"github.com/brimsec/zq/zcode"
+
 	"github.com/brimsec/zq/field"
```
nwt, line 10:
```suggestion
	"github.com/brimsec/zq/field"
	"github.com/brimsec/zq/zcode"
```

nwt (approved):

PR #2155, opened 2021-02-17
Fix ztest

henridf:
Due to a variable sharing bug, ztest was re-running the same test under the guise of all the different tests it was running. 

The first commit in this PR illustrates that with an invalid test that passes.
The second commit fixes ztest which fails the bad test. The third commit removes. the bad test.


henridf:
Oops, this has uncovered some failing tests that we thought were green! Will fix.

mattnibs (approved):
lgtm

ztest/ztest.go
```diff
@@ -199,7 +199,8 @@ func Run(t *testing.T, dirname string) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	for _, b := range bundles {
+	for _, bundle := range bundles {
+		b := bundle
```
nwt, line 203: Nit: The usual idiom for this is to redeclare the same variable instead of introducing a new one.
```go
	for _, b := range bundles {
		b := b
```

henridf: Ah, good to know. I'm not good at remembering these idioms.

PR #2157, opened 2021-02-17
Revert zio/csvio/ztests/type.yaml

nwt:
#2128 clobbered most of this test unnecessarily.

### DIFF
--- a/zio/csvio/ztests/type.yaml
+++ b/zio/csvio/ztests/type.yaml
@@ -3,9 +3,19 @@ zql: '*'
 input: |
   #0:record[type:type]
   0:[-;]
+  0:[;]
+  0:[";]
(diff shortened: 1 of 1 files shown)


henridf (approved):

henridf:
oops i guess the test failure might be why it got clobbered?

mccanne:
None of those values are legal Z type values.  If we want to test this, we need to add appropriate error checking and add tests for the error conditions.

nwt:
I think the right moves are to
1. put a legal Z type value in `zio/csvio/ztests/type.yaml` so it's a meaningful test and
1. check for legal Z type values in `zio/tzngio.Reader`.

I'll create a new PR for that.

PR #2161, opened 2021-02-18
Aurora helm

marktwallace:
This PR consists of several sub-projects that all needed to work before Aurora could be used and tested.

(1) An Aurora cluster has been created. The instructions in aurora.md can be used to reproduce the configuration, but this only needs to be done once per AWS account, so we should not do it again until we start a Prod account.

(2) The Aurora cluster was created in a separate VPC from EKS. This will allow us to reconfigure EKS or create additional clusters without needing to change our Aurora cluster. We set up VPC peering from the EKS VPC to the Aurora VPC. Peering is a tricky to test and time-consuming setup process, so we created a script (peering.sh) so the key steps can be reproduced when we need to do it again for our Prod cluster. As per best practices for security, the Aurora cluster is not publicly accessible. It can only be accessed from within our EKS cluster.

(3) The Helm K8s deployment scripts needed to be restructured to allow Postgres (and Redis) to be optional containers. We use Helm "tags" for the optional features.

(4) Makefile targets needed to be adjusted and a new target created for deployments that use Aurora.

(5) All development and test activities will share a single Aurora cluster. Like EKS, Aurora can become very expensive if each developer needs their own cluster, so we developed a set of conventions to share the Aurora cluster that corresponded to the way we share the EKS test cluster. Each developer or test environment will use a separate Aurora database within the cluster. This provides ideal isolation for testing, which is a good match for the namespace isolation we use in our EKS cluster. There are instructions in aurora.md for how to create Aurora test users and databases using kubectl and psql. It is a manual process, and I will be happy to do it for developers that do not want to deep dive into kubectl and psql.

(6) The cd.yaml tests have been updated to add tests for Aurora. The master deploy now runs on Aurora, which means that environment will be stateful and use db migrations to maintain consistent state as we develop. The master deploy now runs the same tests we run in the test namespace, but runs them with a migrated Aurora db rather than a new, empty database. As we add more tests, the tests for master may diverge from the smoke tests that run in a temporary namespace, because of the need to support migrations.

(7) After upgrading the EKS cluster to K8s 1.19, we started having failures in the automated test in cd.yaml. There were multiple causes for this:
  (a) The new node group created lacked permission to access S3. (It was hard to find this omission.) The eks-setup.md instructions have been updated to add procedures for upgrading the cluster and modifying node groups so we can avoid this in the future.
  (b) Linkerd did not work properly with EKS 1.19 -- the problem was subtle, and showed up when the cd.yaml test would hang indefinitely when trying to remove a temporary namespace (e.g. ci-588). We fixed this by removing Linkerd, which I do not think we will need going forward because we plan on using AWS AMP for storing metrics, and the native metrics for K8s 1.19 are thorough enough that Linkerd does not add any metrics we need.
  (c) In addition, the pattern for temporary namespaces has been changed in the cd.yaml tests. The new tests use no more than 10 temp namespaces and recycle the old ones. I think this is more prudent and will avoid possible future problems with namespace deletion.


.github/workflows/cd.yaml
```diff
@@ -43,7 +43,8 @@ jobs:
           ref: ${{ github.event.pull_request.head.sha }}
       - run: |
           set -x
-          namespace=ci-${{ github.run_number }}
+          runnum=${{ github.run_number }}
+          namespace=ci-temp-${runnum: -1}
```
nwt, line 47: Why do this?

nwt, line 47: Why do this?

ppl/lake/schema.go
```diff
@@ -196,6 +196,11 @@ func CreateOrOpenLakeWithContext(ctx context.Context, rpath string, co *CreateOp
 	mdPath := root.AppendPath(metadataFilename)
 	ok, err := iosrc.Exists(ctx, mdPath)
 	if err != nil {
+		// The error encountered here has been an S3 permission failure,
+		// and it is fatal, so a panic might be better.
+		// A log message would be good in any case. Providing access to
+		// a logger will require modifying mulitple layers of APIs, so we
+		// should discuss the best solution first. -Mark
```
nwt, line 203: @alfred-landrum created https://github.com/brimsec/zq/issues/2204 for this. If additional work is necessary, let's create another issue instead of adding notes like this.
```suggestion
		// The error encountered here has been an S3 permission failure,
		// and it is fatal, so a panic might be better.
		// A log message would be good in any case. Providing access to
		// a logger will require modifying mulitple layers of APIs, so we
		// should discuss the best solution first. -Mark
```

mattnibs: Agreed, or at least this should go in another pr.

.github/workflows/cd.yaml
marktwallace: The idea is to reduce the number of namespaces used in the zq-test cluster. In this case, we are only using namespaces to get isolation from one test run to another. Having 10 separate namespaces which we cycle through is enough to avoid collisions given that each test run is less than 15 minutes. Namespaces have the potential to leak resources, so I prefer to recycle them rather than have an arbitrary number. The bug we hit yesterday where a CRD (for the Linkerd system I subsequently removed) was unexpectedly interfering with namespace removal is an example of a type of problem we could have with future CRDs, where recycle namespaces would minimize the risk of disruptions.

charts/zservice/values.yaml
```diff
@@ -1,22 +1,60 @@
...
+
+root:
+  datauri: "file:///spaces"
+  recruiterAddr: zsrv-recruiter:9867
+
+worker:
+  recruiterAddr: zsrv-recruiter:9867
+
+tags:
+  deploy-postgres: true
```
mattnibs, line 59: I'm trying to see how setting this to false stops postgres from being installed?

ppl/lake/schema.go
marktwallace: This code location is significant. It will be source a failures in the future. The issue 2204 is generic and does not refer to this specific code.

.github/workflows/cd.yaml
```diff
@@ -95,10 +101,32 @@ jobs:
...
           set -x
+          export ZQD_AURORA_USER=cimaster
           export ZQD_DATA_URI=s3://zq-ci-master/manual-test
           export ZQD_ECR_HOST=${{ secrets.AWS_ECR_HOST }}
           aws eks --region us-east-2 update-kubeconfig --name zq-test
           kubectl create ns ci-master || :
           kubectl config set-context --current --namespace=ci-master
-          helm uninstall zsrv || :
-          make helm-install
+          make helm-install-with-aurora
```
mattnibs, line 110: Would still be nice to have this test the postgres kubernetes setup. Perhaps later on, on non-master builds we can do this?

.github/workflows/cd.yaml
```diff
@@ -95,10 +101,32 @@ jobs:
...
+          make test-cluster
+      - name: Get zqd logs on success or failure
+        if: ${{ always() }}
+        run: |
+          set -x
+          ZQD_DATA_URI=s3://zq-ci-test/$(date +%y-%m-%d).${{ github.run_number }}
+          mkdir testlogm
+          kubectl logs --tail 1000 -l app.kubernetes.io/name=recruiter > testlogm/recruiter.log
+          kubectl logs --all-containers --tail 1000 -l app.kubernetes.io/name=root > testlogm/root.log
+          kubectl logs --tail 1000 -l app.kubernetes.io/name=worker > testlogm/worker.log
```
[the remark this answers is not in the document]
marktwallace: That is a good point, so I will increase the number of lines to make is very unlikely that the logs will be truncated. Previously, at the -1 value, there was a chance we would exceed resource in GitHub and not get any logs at all when the logs get flooded.

charts/zservice/values.yaml
marktwallace: The tags feature in Helm is tricky to follow, the tag here is referenced in Chart.yaml:

  - name: postgresql
    repository: https://charts.bitnami.com/bitnami
    version: 10.x.x
    tags: 
      - deploy-postgres

.github/workflows/cd.yaml
marktwallace: Yes, I would be a fan of running these test on all commits. I will figure out a way to do that in a future PR.

ppl/lake/schema.go
marktwallace: I will create an issue that refers to this location and remove the comment.

.github/workflows/cd.yaml
alfred-landrum: I don't understand how this substitution limits the number of namespaces to 10?

nwt: @alfred-landrum: `${runnum: -1}` is [Substring Expansion](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html) with a negative index, yielding the last character of `runnum`.

charts/zservice/values.yaml
alfred-landrum: Should the `deploy-redis` section be under `tags` too, then?

marktwallace: Yes, tags.deploy-redis is defined in values.yaml and defaults to true.

Makefile
```diff
@@ -176,6 +176,20 @@ helm-install:
...
 
+helm-install-with-aurora:
+	helm upgrade -i zsrv charts/zservice \
+	--set root.datauri=$(ZQD_DATA_URI) \
+	--set global.AWSRegion=us-east-2 \
+	--set global.image.repository=$(ZQD_ECR_HOST)/ \
+	--set global.image.tag=zqd:$(ECR_VERSION) \
+	--set global.postgres.addr=$$(aws rds describe-db-cluster-endpoints \
+		--db-cluster-identifier zq-test-aurora \
+		| jq -r ".DBClusterEndpoints[] | select(.EndpointType==\"WRITER\") | .Endpoint"):5432 \
```
nwt, line 187:
```suggestion
		--output text --query "DBClusterEndpoints[?EndpointType=='WRITER'] | [0].Endpoint"):5432 \
```

.github/workflows/cd.yaml
nwt: Reusing namespaces feels like a bad idea to me precisely because they might contain leaked resources, but I'll defer to you.

Makefile
marktwallace: Wow! Thanks!  You have mastered *ALL* the query languages. :-)

mattnibs (approved):

.github/workflows/cd.yaml
mattnibs: I do kind of agree with @nwt on not wanting to reuse namespaces.

PR #2162, opened 2021-02-18
lake: Parallelize tsdir fetch

### LINKED ISSUE #2017: reduce source wait time for archive queries

mattnibs:
The walk of lake stored in s3 is slowed greated in that tsDirs are fetched
sequentially. The main thread sits by idly while the metadata for all chunks in
a tsdir are retrieved over the network. Speed this process up by running the
tsdir fetches each in a separate goroutine.

Because it is important that results be returned in order, save the retrieved
tsdirs in a min/max heap, and release results from the stream when the next
min/max is received.

Closes #2017

### DIFF
--- a/ppl/lake/tsdirstream.go
+++ b/ppl/lake/tsdirstream.go
@@ -1,7 +1,6 @@
 package lake
 
 import (
-	"container/heap"
 	"context"
 	"sort"
 
@@ -13,6 +12,11 @@ import (
 type tsDirStream struct {
 	ch  chan tsDirStreamResult
 	err error
+}
+
+type tsDirStreamResult struct {
+	tsDir  tsDir
+	chunks []chunk.Chunk
 }
 
 func newTsDirStream(ctx context.Context, lk *Lake, tsDirs []tsDir) *tsDirStream {
@@ -29,12 +33,15 @@ func newTsDirStream(ctx context.Context, lk *Lake, tsDirs []tsDir) *tsDirStream 
 }
 
 func (t *tsDirStream) run(ctx context.Context, lk *Lake, tsDirs []tsDir) {
-	ch := make(chan tsDirStreamResult)
-	results := &tsDirStreamResultHeap{order: lk.DataOrder}
+	// Use a channel of channels to ensure results are returned in the correct
+	// order.
+	tsDirChs := make(chan chan tsDirStreamResult, len(tsDirs))
 	g, ctx := errgroup.WithContext(ctx)
-
 	for _, tsDir := range tsDirs {
 		tsDir := tsDir
+		ch := make(chan tsDirStreamResult)
+		tsDirChs <- ch
+
 		g.Go(func() error {
 			chunks, err := tsDirChunks(ctx, tsDir, lk)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


mattnibs:
Some rough performance metrics:

On a 8 cpu aws box running on corelight dataset with threshold set to 1MiB (40 tsdirs w/ ~5 chunks per day)
- master: ~5 secs
- parallelize-fetch-tsdirs: ~2.5 secs

alfred-landrum (commented):
I sent matt some suggestions via slack about using errgroup to track errors & manage the calls to tsDirChunks.

mattnibs:
@alfred-landrum I incorporated your changes with some minor adjustments.

ppl/lake/tsdirstream.go
```diff
@@ -0,0 +1,120 @@
...
+	})
+
+	t := &tsDirStream{ch: make(chan tsDirStreamResult)}
+	go t.run(ctx, lk, tsDirs)
+	return t
+}
+
+func (t *tsDirStream) run(ctx context.Context, lk *Lake, tsDirs []tsDir) {
+	ch := make(chan tsDirStreamResult)
+	results := &tsDirStreamResultHeap{order: lk.DataOrder}
```
nwt, line 33: Simplify by using a channel of channels instead of `tsDirStreamResultHeap`?

mattnibs: Interesting.

mattnibs: @nwt is this what you are talking about: https://github.com/brimsec/zq/pull/2162/commits/8daf603c6883973472396131ddf83d1127a63f7d

ppl/lake/tsdirstream.go
```diff
@@ -0,0 +1,81 @@
...
+			}
+			return nil
+		})
+	}
+
+	close(tsDirChs)
+	g.Go(func() error {
+		for ch := range tsDirChs {
+			select {
+			case t.ch <- <-ch:
```
alfred-landrum, line 63: I think the `<-ch` statement here will be evaluated before & separate from the select statement, so would block forever if goroutine above errored on `tsDirChunks(...)`.

ppl/lake/tsdirstream.go
```diff
@@ -0,0 +1,81 @@
...
+
+	t := &tsDirStream{ch: make(chan tsDirStreamResult)}
+	go t.run(ctx, lk, tsDirs)
+	return t
+}
+
+func (t *tsDirStream) run(ctx context.Context, lk *Lake, tsDirs []tsDir) {
+	// Use a channel of channels to ensure results are returned in the correct
+	// order.
+	tsDirChs := make(chan chan tsDirStreamResult, len(tsDirs))
```
alfred-landrum, line 38: Could this be a `[]chan tsDirStreamResult` instead of a channel of channels?

mattnibs: Yes actually and I think this is better.

ppl/lake/tsdirstream.go
mattnibs: I was wondering this myself, let me do a quick test.

ppl/lake/tsdirstream.go
nwt: It's certainly possible but requires explicit synchronization. I think the channel of channels winds up being simpler as a result, but to be clear, I'm fine with either approach.

mattnibs: One benefit of using the chan of chans I suppose is that it could be used to limit the number of in flight tsdir fetches at a given time. That said we're not currently doing this.

ppl/lake/tsdirstream.go
mattnibs: Yep you are right this causes a deadlock if the first channel in the chain never emits or closes.

ppl/lake/tsdirstream.go
mattnibs: > It's certainly possible but requires explicit synchronization.

@nwt I think I am missing something. How so?

mattnibs: The list of tsDirChans would be fully populated in the main goroutine before being iterated on in the consumer goroutine.

nwt: Ah, in that case, no synchronization required.

alfred-landrum (approved):

PR #2170, opened 2021-02-18
break out compiler package into sub-packages

mccanne:
This commit refactors the compiler package so that it is now
comprised of semantic analysis (compiler/semantic) and runtime
generation of the semantic output to the Z runtime kernel
represenation (aka the proc.Interface).

No new code or logic is introduced here.  A subsequent PR will use
this refactoring to create a new flowgraph-based DSL that will
server as the intermediate form between the parser's AST and
the Z kernel runtime flowgraph.

### DIFF
--- a/compiler/kernel/proc.go
+++ b/compiler/kernel/proc.go
@@ -1,4 +1,4 @@
-package compiler
+package kernel
 
 import (
 	"errors"
@@ -6,7 +6,6 @@ import (
 
 	"github.com/brimsec/zq/compiler/ast"
 	"github.com/brimsec/zq/expr"
-	"github.com/brimsec/zq/expr/agg"
 	"github.com/brimsec/zq/field"
 	"github.com/brimsec/zq/proc"
 	"github.com/brimsec/zq/proc/combine"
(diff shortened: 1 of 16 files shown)


compiler/kernel/proc.go
```diff
@@ -185,11 +184,7 @@ func compileProc(custom Hook, node ast.Proc, pctx *proc.Context, scope *Scope, p
 		return fuse.New(pctx, parent)
 
 	case *ast.FunctionCall:
-		p, err := convertFunctionProc(v)
-		if err != nil {
-			return nil, err
-		}
-		return compileProc(custom, p, pctx, scope, parent)
+		return nil, errors.New("internal bug: semantic analyzer should have converted functioni in proc context to filter or group-by")
```
nwt, line 187:
```suggestion
		return nil, errors.New("internal error: semantic analyzer should have converted function in proc context to filter or group-by")
```

driver/compile.go
```diff
@@ -7,6 +7,8 @@ import (
 	"time"
 
 	"github.com/brimsec/zq/compiler"
+	"github.com/brimsec/zq/compiler/kernel"
+
 	// XXX replace this with flow DSL
 	"github.com/brimsec/zq/compiler/ast"
```
nwt, line 13:
```suggestion
	"github.com/brimsec/zq/compiler"
	// XXX replace this with flow DSL
	"github.com/brimsec/zq/compiler/ast"
	"github.com/brimsec/zq/compiler/kernel"
```

nwt (approved):
:shipit:

PR #2172, opened 2021-02-19
zqd: Fix pprof routes

mattnibs:
The zqd pprof routes were setup in a way assuming the gorilla mux router
behaves similarly to go http.ServeMux. It does not; this resulted in the routes
not behaving as expected. For instance the router was only routing
/debug/pprof/ to pprof.Index and the other sub routes of pprof.Index (e.g.
"/block", "/goroutine", etc) were getting 404'd.

Do things the gorilla mux router way so the pprof routes work correctly.


alfred-landrum (approved):
nice!

PR #2173, opened 2021-02-19
Add support for shaping arrays and sets

### LINKED ISSUE #2038: Z shaping: handle sets and arrays (for "deep json")
The shaper functions introduced in #1984 do not handle arrays and sets. This issue consists of adding that support. This is likely to be useful particular for "deep json" where arrays of objects are common.

henridf:
close #2038


expr/shaper.go
```diff
@@ -50,40 +54,90 @@ func (s *step) append(step step) {
...
+	if _, ok := t.(*zng.TypeArray); ok {
+		return true
+	}
+	if _, ok := t.(*zng.TypeSet); ok {
+		return true
+	}
+	return false
+}
+
+func innerType(t zng.Type) zng.Type {
```
mccanne, line 106: It seems like for type comparisons, you want referential transparency but for the outcome types you want to preserve any original Z typedefs.

expr/shaper.go
```diff
@@ -383,35 +510,51 @@ func (s *Shaper) castRecordType(input, spec *zng.TypeRecord) (*zng.TypeRecord, e
...
-			}
-			cols = append(cols, zng.Column{inCol.Name, specCol.Type})
-		default:
-			// 5. Non-castable type pair with at least one
-			// (non-record) container: output column is left
-			// unchanged.  Note that eventually, we should
-			// recognize and cast e.g. array[string] to array[ip].
-			// xxx before merge: file issue for this and mention
-			// it here.
+		if inCol.Type.ID() == specCol.Type.ID() {
```
mccanne, line 517: Typedefs?

mccanne (approved):
Cool.   Nice work!  We should make sure we have all the right behaviors with Z typedefs but we can merge this and fix it later.    This is isolated and won't break anything core right now...

henridf:
@mccanne yes, filed https://github.com/brimsec/zq/issues/2192 for the right typedef behaviors

expr/shaper.go
henridf: 👍 #2192

expr/shaper.go
henridf: 👍 #2192

PR #2180, opened 2021-02-20
change -z to -I and allow multiple files

mccanne:
This commit changes the flag (-z) for running zq with the Z query in
a file.  The flag is replaced with -I (for "include") and the behavior
is a bit different: -I can be used multiple times and it is combined
with the normal Z query on the command line.  These combinations can
create some confusing syntax errors so the Z text parsed is display
when a syntax error occurs.

We also changed the "ZQL" name convention to "Z" in the command help.

### DIFF
--- a/cmd/zq/zq.go
+++ b/cmd/zq/zq.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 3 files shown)


henridf (approved):
👍 but not sure why the comment test is failing?

PR #2188, opened 2021-02-22
simplify ast unpacker

mccanne:
This commit simplifies the ast unpacker by adding a generic package
for umarshaling interface types by keying on a configurable field
in the json.  The value of said key indicates which Go struct type
to unmarshal into via its type name (though this can be overriden
as evidence in our current ast where the reflector AddAs() method
is used).  Also, this requires that the Op field be filled in for
anything marshaled, which is good habit anyway because forthcoming code
that transforms ASTs and sends the transformed AST over the network
to workers must have all the Op fields present.


mccanne:
The line count is net positive but only because of all the tests...

compiler/ast/ast.go
```diff
@@ -49,19 +49,19 @@ type Expression interface {
...
 type Search struct {
-	Op    string  `json:"op"`
+	Op    string  `json:"op" unpack:""`
 	Text  string  `json:"text"`
 	Value Literal `json:"value"`
 }
 
 type UnaryExpression struct {
-	Op       string     `json:"op"`
+	Op       string     `json:"op" unpack:"UnaryExpr"`
```
nwt, line 64: Maybe align all the `ast` type names with their corresponding `op` values so all of these can be `unpack:""`?

compiler/ast/unpack.go
```diff
@@ -1,648 +1,104 @@
...
+	TypeRecord{},
+	TypeSet{},
+	TypeUnion{},
+	TypeValue{},
+	UniqProc{},
+	BinaryExpression{},
+	SelectExpression{},
+	UnaryExpression{},
+	CastExpression{},
+	ConditionalExpression{},
```
nwt, line 63: Nit: Sort me.

pkg/unpack/reflector.go
```diff
@@ -0,0 +1,404 @@
...
+func New() Reflector {
+	return make(map[string]map[string]reflect.Type)
+}
+
+func (r Reflector) Init(templates ...interface{}) Reflector {
+	for _, t := range templates {
+		r.Add(t)
+	}
+	return r
+}
```
nwt, line 24: Nit: I'd get rid of `Init` and allow `New` to accept templates.
```suggestion
func New(templates ...interface{}) Reflector {
	r := Reflector(make(map[string]map[string]reflect.Type))
	for _, t := range templates {
		r.Add(t)
	}
	return r
}
```

pkg/unpack/reflector.go
```diff
@@ -0,0 +1,404 @@
...
+	unpackKey, unpackVal, err := structToUnpackRule(typ)
+	if err != nil {
+		panic(err.Error())
+	}
+	if unpackKey == "" {
+		panic(fmt.Sprintf("unpack tag not found for Go type '%s'", typ.Name()))
+	}
+	types := r.get(unpackKey, true)
+	_, ok := types[unpackVal]
+	if ok {
```
nwt, line 37:
```suggestion
	if _, ok := types[unpackVal]; ok {
```

pkg/unpack/reflector.go
```diff
@@ -0,0 +1,404 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if rv, ok := skeleton.(reflect.Value); ok {
+		skeleton = rv.Interface()
+	}
+	c := &mapstructure.DecoderConfig{
+		TagName: "json",
+		Result:  skeleton,
+		//Squash:  true,
```
nwt, line 85:
```suggestion
```

pkg/unpack/reflector.go
```diff
@@ -0,0 +1,404 @@
...
+		if types == nil {
+			continue
+		}
+		unpackVal, ok := val.(string)
+		if !ok {
+			return zero, fmt.Errorf("unpack key in JSON field '%s' is not a string: '%T'", key, val)
+		}
+		hits++
+		template, ok := types[unpackVal]
+		if ok {
```
nwt, line 107:
```suggestion
		if template, ok := types[unpackVal]; ok {
```

pkg/unpack/reflector.go
```diff
@@ -0,0 +1,404 @@
...
+			// Return the reflect.Value struct pointer as interface{}
+			// so that the callee can pull out the reflect.Value and
+			// either install it as a field of another reflect.Value
+			// or at the root of the descent, convert it back to an
+			// empty inteface pointing a conrete instance of the
+			// converted struct to be fully decoded by mapstructure.
+			return template, nil
+		}
+		return converted, nil
+
```
nwt, line 149:
```suggestion
```

pkg/unpack/tag.go
```diff
@@ -0,0 +1,112 @@
...
+	return s, ok
+}
+
+// structToRule looks through the tags in a struct to (1) make sure there
+// are no duplicate tags since package json doesn't do this and it results
+// in hard-to-debug behaviors, and (2) it finds an "unpack" tag which indicates
+// which field of the struct to use as the key and what value to match for
+// unmarshling into this concrete type.  If there is not exactly one unpack
+// tag, then an error is returned.
+// unpack="key,skip" unpack="key", unpack="", or unpack="skip"
```
nwt, line 49: I think the syntax ought to be `unpack=",skip"` just as `encoding/json` uses `json=",omitempty"`.

pkg/unpack/tag.go
```diff
@@ -0,0 +1,112 @@
...
+		val, skip, err := parseUnpackOpts(unpackOpt, opts, typ.Name())
+		if err != nil {
+			return "", "", err
+		}
+		if skip {
+			if unpackSkip {
+				return "", "", ErrSkip
+			}
+			unpackSkip = true
+		}
```
nwt, line 80: `skip` doesn't appear to do anything other than cause an error if repeated.

compiler/ast/ast.go
mccanne: This has been on my to-do list for a while but it breaks the app so I was waiting for the right moment.

pkg/unpack/tag.go
mccanne: Oops, the code that did the right thing disappeared.  I'll put it back and make a test.  Also, the ",skip" convention will simplify the logic.

pkg/unpack/tag.go
```diff
@@ -0,0 +1,100 @@
...
+// structToRule looks through the tags in a struct to (1) make sure there
+// are no duplicate tags since package json doesn't do this and it results
+// in hard-to-debug behaviors, and (2) it finds an "unpack" tag which indicates
+// which field of the struct to use as the key and what value to match for
+// unmarshling into this concrete type.  If there is not exactly one unpack
+// tag, then an error is returned.
+// unpack="key,skip" unpack="key", unpack="", or unpack="skip"
+func structToUnpackRule(typ reflect.Type) (string, string, bool, error) {
+	if typ.Kind() != reflect.Struct {
+		return "", "", false, errors.New("cannot unpack into non-struct")
```
nwt, line 52:
```suggestion
		return "", "", false, fmt.Errorf("cannot unpack into non-struct %s", type)
```

pkg/unpack/reflector.go
```diff
@@ -0,0 +1,403 @@
...
+}
+
+func (r Reflector) Add(template interface{}) Reflector {
+	typ := reflect.TypeOf(template)
+	unpackKey, unpackVal, skip, err := structToUnpackRule(typ)
+	if err != nil {
+		panic(err.Error())
+	}
+	if unpackKey == "" {
+		panic(fmt.Sprintf("unpack tag not found for Go type '%s'", typ.Name()))
```
nwt, line 30: `reflect.Type.Name` often returns the empty string for non-struct types. `reflect.Type.String` does not.
```suggestion
		panic(fmt.Sprintf("unpack tag not found for Go type '%s'", typ)))
```

pkg/unpack/reflector.go
```diff
@@ -0,0 +1,403 @@
...
+		r.Add(t)
+	}
+	return r
+}
+
+func (r Reflector) Add(template interface{}) Reflector {
+	typ := reflect.TypeOf(template)
+	unpackKey, unpackVal, skip, err := structToUnpackRule(typ)
+	if err != nil {
+		panic(err.Error())
```
nwt, line 27:
```suggestion
		panic(err)
```

pkg/unpack/tag.go
```diff
@@ -0,0 +1,100 @@
...
+	if len(elems) > 1 {
+		opts = elems[1:]
+	}
+	return elems[0], true, opts
+}
+
+func jsonFieldName(f reflect.StructField) (string, bool) {
+	s, ok, _ := parseTag(tagJSON, f)
+	return s, ok
+}
```
nwt, line 41: This looks unused.

pkg/unpack/tag.go
```diff
@@ -0,0 +1,100 @@
...
+		if skip {
+			if unpackSkip {
+				return "", "", false, ErrSkip
+			}
+			unpackSkip = true
+		}
+		if unpackKey != "" {
+			return "", "", false, fmt.Errorf("unpack key appears twice in field '%s' of struct type '%s' may only be 'skip'", field.Name, typ.Name())
+		}
+		if jsonField == "" {
```
nwt, line 90: Might want to return an error if the JSON field name is `-`.

nwt (approved):

pkg/unpack/tag.go
mccanne: used in reflector.go

pkg/unpack/tag.go
mccanne: I made it but didn't update the comment.  Done.

compiler/ast/ast.go
mccanne: Agreed.  I'll put this PR up after I merge sql (hopefully soon!)

mccanne: issue #2222

PR #2190, opened 2021-02-22
adjust type system to realign with zson design

### LINKED ISSUE #1835: refactor resolver into zng.TypeContext and zng.Marshal and move tzngio type parsing elsewhere
BTW, in my last PR on marshaler, I had planned to move it from resolver to ZNG, but to hide the zctx, zng would need to import resolver creating a cycle.  I think the right solution is to move resolver.Context to zng.TypeContext.  resolver/context.go is a bit of a beast and hasn’t ever been revisited really since it’s original incarnation.  I think it can be greatly simplified by moving all the tzng type parsing to zio/tzngio and simplify the type lookup table.  Then we would have zng.Marshal and zng.TypeContext all in zng.  This makes zng bigger but I think it’s the right size for what it does.

mccanne:
This commit provides a long-needed refacting to realign with the
zson/zng focus of the project.  The tzng type parser was moved
to zio/tzngio and the type context logic moved to package zson.
We wrapped the new zson.TypeContext in resolver.Context with the
idea of removing package resovler in a subsequent commit,
With resolver depending on zson, the marshaling needed to move
into zson.  There is also a wrapper for zng.Marshaler that points
back at zson.ZNGMarshaler.

At this point, the types in the type context table are all native
zson and the tzng types are isolated to package tzngio.

This is also a first step toward getting the tzng formatting and
parsing logic out of package zng.

Closes #1835


henridf (approved):

zson/resolver.go
```diff
@@ -0,0 +1,329 @@
+package zson
```
nwt, line 1: Nit: Call this file `typecontext.go`.

zio/zngio/reader.go
```diff
@@ -1,3 +1,14 @@
...
+// directives in binary zng format.  The Reader and Writer types implement the
+// the zbuf.Reader and zbuf.Writer interfaces.  Since these methods
+// read and write only zbuf.Records, but the zng format includes additional
+// functionality, other methods are available to read/write zng comments
+// and include virtual channel numbers in the stream.  Virtual channels
+// provide a way to indicate which output of a flowgraph a result came from
+// when a flowgraph computes multiple output channels.  The zng values in
+// this zng value are "machine format" as prescirbed by the ZNG spec.
+// The vanilla zbuf.Reader and zbuf.Writer implementations ignore application-specific
+// payloads (e.g., channel encodings).
```
nwt, line 11: Nit: Put this in `doc.go`.

zio/tzngio/type-parser.go
```diff
@@ -0,0 +1,339 @@
+package tzngio
```
nwt, line 1: NIt: We don't usually use `-` or `_` in Go source file names, so call this `typeparser.go`.

zson/resolver.go
nwt: Or `context.go` if you rename `TypeContext` to `Context`!

mccanne: as discussed will change zson.TypeContext to zson.Context and name this file context.go

zson/marshal_test.go
```diff
@@ -1,13 +1,6 @@
...
 
-import (
-	"strings"
-	"testing"
-
-	"github.com/brimsec/zq/zson"
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
-)
+/* NOT YET
```
nwt, line 3: Nit: Say `t.Skip("blah blah blah")` in each test rather than commenting them.

nwt (approved):

zson/marshal_test.go
mccanne: Oops, @alfred-landrum  saw that too.  Those tests are fine and a last vestige of the code movements.  Will fix.

PR #2199, opened 2021-02-22
Shaping functions: make '.' subject implicit

henridf:
In the common case of shaping the entire record as in `shape(., <sometype..>)`, allow the `.` argument to be omitted, following the pattern introduced with `is()`.


henridf:
(Maybe I should do this for cut and pick too?)

mccanne (approved):

PR #2200, opened 2021-02-22
add taste operator

mccanne:
This commit adds a "taste" operator (aka proc) that provides
a taste of the shape of the data.  It takes an optional
field-expression argument and computes "taste=any(e) by shape=typeof(e)"
where e is "." if it is omitted.  This is a nice shortcut
for demo, docs, and articles.

There is actually no new proc in the runtime as the parser
does a direct translation to the group-by proc.

### DIFF
--- a/zql/zql.peg
+++ b/zql/zql.peg
@@ -464,7 +464,7 @@ JoinKey
 TasteProc
   = "taste"i e:TasteExpr {
       RETURN(MAP("op": "GroupByProc",
-        "keys":ARRAY(MAP("op": "Assignment",
+        "keys": ARRAY(MAP("op": "Assignment",
                    "lhs": MAP("op": "Identifier", "name": "shape"),
                    "rhs": MAP("op": "FunctionCall", "function": "typeof",
                               "args": ARRAY(e)))),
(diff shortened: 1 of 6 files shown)


zql/zql.peg
```diff
@@ -460,6 +461,26 @@ JoinKey
...
+                   "lhs": MAP("op": "Identifier", "name": "shape"),
+                   "rhs": MAP("op": "FunctionCall", "function": "typeof",
+                              "args": ARRAY(e)))),
+        "reducers": ARRAY(MAP("op": "Assignment",
+                              "lhs": MAP("op": "Identifier", "name": "taste"),
+                              "rhs": MAP("op": "Reducer",
+                                         "operator": "any",
+                                         "expr": e,
+                                         "where": NULL))),
+         "duration": NULL, "limit": 0))
```
nwt, line 477: `"keys"` and `"reducers"` are indented eight spaces.
```suggestion
        "duration": NULL, "limit": 0))
```

zql/zql.peg
```diff
@@ -460,6 +461,26 @@ JoinKey
   = Lval
   / "(" expr:Expr ")" { RETURN(expr) }
 
+TasteProc
+  = "taste"i e:TasteExpr {
+      RETURN(MAP("op": "GroupByProc",
+        "keys":ARRAY(MAP("op": "Assignment",
```
nwt, line 467:
```suggestion
        "keys": ARRAY(MAP("op": "Assignment",
```

nwt (approved):

PR #2203, opened 2021-02-23
merge zson type table

### LINKED ISSUE #2191: merge zson TypeTable into TypeContext
The zson types table should be merged with zson.TypeContext.  The LUT can be removed from the latter.  This will make type lookups a big more efficient by channeling everything through a common types table.

### LINKED ISSUE #2198: type bindings need to be evaluated incrementally in compileConsts
This doesn't work:
```
type port=uint16;
type conn={orig_h:ip,orig_p:port,resp_h:ip,resp_p:port};
put . = shape(., {_path:string,ts:time,uid:bstring,id:conn})
```

mccanne:
The ZSON type table was a temporary structure until the moment we could
merge it with the type context, which is now.  Type lookups now all
go through a hash table that also tracks type value to type bindings
so zng.Value type values can be efficiently turned into zng.Type
interface pointers.

Fixes #2198
Closes #2191


zson/context.go
```diff
@@ -285,34 +244,47 @@ func (c *Context) AddColumns(r *zng.Record, newCols []zng.Column, vals []zng.Val
...
+		// ParseType will ensure the canonical zson is in the toType table,
+		// but the zson argument here may be any conforming zson type string.
+		// Since this string may appear repeatedly (e.g., type values
+		// coming from an external system) we put an extra entry in the
+		// lookup-table to cache it so we don't parse every instance
+		// of a type string when it is not in canonical form.
+		c.toType[zson] = typ
 	}
-	return typ, err
+	return typ, nil
```
nwt, line 283: Nit: I think this reads better with less nesting.
```go
	if typ, ok := c.toType[zson]; ok {
		return typ, nil
	}
	if typ := zng.LookupPrimitive(zson); typ != nil {
		c.toBytes[typ] = zcode.Bytes(zson)
		c.toType[zson] = typ
		return typ, nil
	}
	// ParseType we will re-enter the context and create and/or
	// return an existing type.  Since it's re-entrant we can't
	// (and don't want to) hold the lock.  There can be a race
	// here but it doesn't matter because there is only ever one
	// type that wins the day because of the incremental locking on
	// each component of a nested type.
	c.mu.Unlock()
	typ, err := ParseType(c, zson)
	if err != nil {
		return nil, err
	}
	c.mu.Lock()
	// ParseType will ensure the canonical zson is in the toType table,
	// but the zson argument here may be any conforming zson type string.
	// Since this string may appear repeatedly (e.g., type values
	// coming from an external system) we put an extra entry in the
	// lookup-table to cache it so we don't parse every instance
	// of a type string when it is not in canonical form.
	c.toType[zson] = typ
	return typ, nil
```

nwt (approved):

PR #2209, opened 2021-02-24
change zapi ls output to use output flags

alfred-landrum:
This changes the default output of `zapi ls`, and adds support for our standard output format flags.

The output is generated by marshaling the returned api.Space's from the `SpaceList` call, and then sending the `*zng.Record`'s using the specified format. The default format is tabular:

```bash
$ zapi ls
ID                             NAME DATA_PATH                                                       STORAGE_KIND
sp_1ouEIxXKyGl3bJNLpaZD5S7Vtwm sp1  file:///Users/alfred/tmp/zqdroot/sp_1ouEIxXKyGl3bJNLpaZD5S7Vtwm archivestore
sp_1ouEJDTgHJSGM6C9mz33bGwOHLt sp2  file:///Users/alfred/tmp/zqdroot/sp_1ouEJDTgHJSGM6C9mz33bGwOHLt archivestore
```

### DIFF
--- a/cmd/zapi/cmd/info/ls.go
+++ b/cmd/zapi/cmd/info/ls.go
@@ -2,36 +2,35 @@ package info
 
 import (
 	"flag"
-	"fmt"
-	"os"
-	"strings"
 
 	"github.com/brimsec/zq/api"
+	"github.com/brimsec/zq/cli/outputflags"
 	"github.com/brimsec/zq/cmd/zapi/cmd"
-	"github.com/brimsec/zq/pkg/colw"
+	"github.com/brimsec/zq/zng"
+	"github.com/brimsec/zq/zng/resolver"
+	"github.com/brimsec/zq/zson"
 	"github.com/mccanne/charm"
-	"github.com/mccanne/charm/pkg/termwidth"
 )
 
 var Ls = &charm.Spec{
 	Name:  "ls",
-	Usage: "ls [-l] [glob1 glob2 ...]",
+	Usage: "ls [glob1 glob2 ...]",
 	Short: "list spaces or information about a space",
 	Long: `The ls command lists the names and information about spaces known to the system.
 When run with arguments, only the spaces that match the glob-style parameters are shown
-much like the traditional unix ls command.  When used with "-l", each space specified
-is listed with its detailed information as in the info command.`,
+much like the traditional unix ls command.`,
 	New: NewLs,
 }
 
 type LsCommand struct {
 	*cmd.Command
-	lflag bool
+	outputFlags outputflags.Flags
 }
 
 func NewLs(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
 	c := &LsCommand{Command: parent.(*cmd.Command)}
-	f.BoolVar(&c.lflag, "l", false, "show detail information about each space listed")
+	c.outputFlags.DefaultFormat = "table"
+	c.outputFlags.SetFormatFlags(f)
 	return c, nil
 }
 
@@ -46,7 +45,6 @@ func (c *LsCommand) Run(args []string) error {
 	matches, err := cmd.SpaceGlob(c.Context(), conn, args...)
 	if err != nil {
 		if err == cmd.ErrNoSpacesExist {
-			fmt.Println("no spaces exist")
 			return nil
 		}
 		return err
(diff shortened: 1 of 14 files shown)


mccanne:
cool!

cmd/zapi/cmd/info/ls.go
```diff
@@ -46,33 +45,37 @@ func (c *LsCommand) Run(args []string) error {
 	matches, err := cmd.SpaceGlob(c.Context(), conn, args...)
 	if err != nil {
 		if err == cmd.ErrNoSpacesExist {
-			fmt.Println("no spaces exist")
```
mattnibs, line 49: I think this is a nice to have, maybe not printed to stdout though.

mattnibs (commented):
Code looks fine, just had a comment about output tastes:

As a heavy user of this command I think I would want the name column to be listed first as this is usually what I'm looking for. Also not sure how useful the `DATA_PATH` column is given how much space it takes up. Perhaps these column could be hidden and showed when a `-v` flag is added?

alfred-landrum:
> As a heavy user of this command I think I would want the name column to be listed first as this is usually what I'm looking for. Also not sure how useful the `DATA_PATH` column is given how much space it takes up. Perhaps these column could be hidden and showed when a `-v` flag is added?

I talked with @mattnibs offline a little more, and said that `zapi ls` (with no flags) would use the `text` output format on structs that had only the spaces name, and that `zapi ls -l` would use the full Space struct returned from the client. In both case, you can use any of our output flags:

```
12:37 $ zapi ls
sp1
sp2
✔ ~/work/zq [zapi-ls-zout L|✚ 1⚑ 5]
12:37 $ zapi ls -l
sp_1ouEIxXKyGl3bJNLpaZD5S7Vtwm	sp1	file:///Users/alfred/tmp/zqdroot/sp_1ouEIxXKyGl3bJNLpaZD5S7Vtwm	archivestore
sp_1ouEJDTgHJSGM6C9mz33bGwOHLt	sp2	file:///Users/alfred/tmp/zqdroot/sp_1ouEJDTgHJSGM6C9mz33bGwOHLt	archivestore
✔ ~/work/zq [zapi-ls-zout L|✚ 1⚑ 5]
12:39 $ zapi ls -f table
NAME
sp1
sp2
✔ ~/work/zq [zapi-ls-zout L|✚ 1⚑ 5]
12:39 $ zapi ls -f table -l
ID                             NAME DATA_PATH                                                       STORAGE_KIND
sp_1ouEIxXKyGl3bJNLpaZD5S7Vtwm sp1  file:///Users/alfred/tmp/zqdroot/sp_1ouEIxXKyGl3bJNLpaZD5S7Vtwm archivestore
sp_1ouEJDTgHJSGM6C9mz33bGwOHLt sp2  file:///Users/alfred/tmp/zqdroot/sp_1ouEJDTgHJSGM6C9mz33bGwOHLt archivestore
```

cmd/zapi/cmd/info/ls.go
alfred-landrum: talked with matt, said that the lack of output if no spaces were present matched `ls`, he agreed.

cmd/zapi/cmd/info/ls.go
```diff
@@ -55,24 +55,63 @@ func (c *LsCommand) Run(args []string) error {
...
 	}
-	names := cmd.SpaceNames(matches)
-	width := termwidth.Width()
-	// print listing laid out in columns like ls
-	err = colw.Write(os.Stdout, names, width, 3)
-	if err == colw.ErrDoesNotFit {
-		fmt.Println(strings.Join(names, "\n"))
+	names := make([]string, 0, len(matches))
+	for i := range matches {
+		names = append(names, matches[i].Name)
```
nwt, line 62:
```suggestion
	for _, m := range matches {
		names = append(names, m.Name)
```

mattnibs (approved):

nwt (approved):

cmd/zapi/cmd/info/ls.go
philrz: Sorry to chime in late on what might be a settled topic, but I lean more toward what @mattnibs was implying. If a user has typed what's effectively a "show me something" command, and there's nothing to show, I've generally always been in favor of an explicit "there is nothing to show" sort of response, rather than silence being used to implicitly communicate this. I've always leaned this way because "early exit" bugs can happen where the there _is_ data that should have been shown but the tools just silently failed. So the explicit response helps differentiate between the two.

Also, I agree that stdout would not be the appropriate place for the message in any case, since a user piping might be piping the output to something downstream.

All that said, none of this is a hill I'm prepared to die on, so I'm also prepared to be overruled.

alfred-landrum: I think we should mirror what `ls` does, similar to other unix like tools, where it's acceptable to return exit status of 0 and nothing to stdout on success, so I'm going to merge in as-is.

We already have this behavior today for zq, btw, where you can execute zq and get silent success, because no records were returned or generated.

PR #2210, opened 2021-02-24
add support for three basic joins

mccanne:
This commit extends the current join syntax to support "inner join",
"left join", and "right join".  The default "join" is now an
inner join (to follow SQL convention) so this will break any scripts
or workflow out there that assumes the left-outer behavior of the
previous "join" command.


mccanne:
This is still a little fragile because the join-key expressions only work on their incoming side of data and because RIGHT JOIN is a swapped version of LEFT JOIN you have to right the key expression backward.    See issue #2228.

nwt (approved):

PR #2211, opened 2021-02-24
improve taste

mccanne:
This commit improves the taste operator by outputting just the
sampled values (one per unique schema) instead of a record
comprising the sample and its type.  As Z values are self describing,
the former output was redundant and adds noise that gets in the way.

We also fixed a bug introduced in the dash-z commit where the default 
output-flags for ztests weren't properly expressed as "-f zson -pretty=0".


mccanne:
This is waiting on #2218

nwt (approved):

Issue #2197, opened 2021-02-22
Canonical Z generator


philrz:
(The examples below use the [Zeek `http` NDJSON zq-sample-data](https://github.com/brimsec/zq-sample-data/blob/master/zeek-ndjson/http.ndjson.gz).)

Convenient shorthand in Z can sometimes also be hazardous, particularly for new users. For example, let's say a user gets the hang of chaining together multiple search terms via implicit boolean `and`, matching to this record:

```
$ gzcat http.ndjson.gz | head -1 | zq -f ndjson '_path=http windows' -
{"_path":"http","_write_ts":"2018-03-24T17:15:20.610930Z","host":"10.47.3.200","id.orig_h":"10.164.94.120","id.orig_p":36729,"id.resp_h":"10.47.3.200","id.resp_p":80,"method":"GET","request_body_len":0,"resp_fuids":["FnHkIl1kylqZ3O9xhg"],"resp_mime_types":["text/html"],"response_body_len":56,"status_code":301,"status_msg":"Moved Permanently","tags":[],"trans_depth":1,"ts":"2018-03-24T17:15:20.609736Z","uid":"CpQfkTi8xytq87HW2","uri":"/chassis/config/GeneralChassisConfig.html","user_agent":"Mozilla/4.0 (compatible; MSIE 8.0; Windows NT 5.1; Trident/4.0)","version":"1.1"}
```

Likewise, they learn the ropes of explicit boolean `or`:

```
$ gzcat http.ndjson.gz | head -1 | zq -f ndjson 'chassis OR macintosh' -
{"_path":"http","_write_ts":"2018-03-24T17:15:20.610930Z","host":"10.47.3.200","id.orig_h":"10.164.94.120","id.orig_p":36729,"id.resp_h":"10.47.3.200","id.resp_p":80,"method":"GET","request_body_len":0,"resp_fuids":["FnHkIl1kylqZ3O9xhg"],"resp_mime_types":["text/html"],"response_body_len":56,"status_code":301,"status_msg":"Moved Permanently","tags":[],"trans_depth":1,"ts":"2018-03-24T17:15:20.609736Z","uid":"CpQfkTi8xytq87HW2","uri":"/chassis/config/GeneralChassisConfig.html","user_agent":"Mozilla/4.0 (compatible; MSIE 8.0; Windows NT 5.1; Trident/4.0)","version":"1.1"}
```

But then they try to combine the two concepts and get things like:

```
[...]

PR #2217, opened 2021-02-24
add ast output formatter and "summarize" keyword

mccanne:
This command adds a new package zfmt to output ast's in debuggable
and canonical form.  There are two format variants as we envision
the Z kernel DSL to diverge from the Z syntax over time.  Currently,
this functionality is exposed only through the ast command but
a future PR will add a general output flag option to zq etc once this
approach is stable.

We also added an experimental "summarize" keyword to the Z language
as an optional prefix of an aggregation to go with the canonical
form of output.

We are merging a protoype of this functionality here to experiment
with and gain feedback.  This is a low-risk commit while we fine
tune zfmt since it is only exposed through the ast command for now.

This PR addresses #2197 but does not close it as I would like to merge this 
experimental version first, get feedback, then finalize #2197 later.

### DIFF
--- a/cmd/ast/ast.go
+++ b/cmd/ast/ast.go
@@ -6,11 +6,14 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"log"
 	"os/exec"
 	"strings"
 
 	"github.com/brimsec/zq/compiler"
+	"github.com/brimsec/zq/compiler/semantic"
+	"github.com/brimsec/zq/zfmt"
 	"github.com/brimsec/zq/zql"
 	"github.com/mccanne/charm"
 	"github.com/peterh/liner"
@@ -29,6 +32,10 @@ translated into the analytics requests that is sent to the zqd search endpoint.
 By default, it runs the built-in PEG parser built into this go binary.
 If you specify -js, it will try to run a javascript version of the parser
 by execing node in the currrent directory running the javascript in ./zql/run.js.
+
+The -O flag is handy for turning on and off the compiler, which lets you see
+how the parsed AST is transformed into a runtime object comprised of the
+Z kernel operators.
 `,
 	New: func(parent charm.Command, flags *flag.FlagSet) (charm.Command, error) {
 		return New(flags)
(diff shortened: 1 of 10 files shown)


mccanne:
This is waiting on #2218

cmd/ast/ast.go
```diff
@@ -55,18 +66,38 @@ func New(f *flag.FlagSet) (charm.Command, error) {
 	f.BoolVar(&c.pigeon, "pigeon", true, "run pigeon version of peg parser")
 	f.BoolVar(&c.ast, "ast", false, "run pigeon version of peg parser and show marshaled ast")
 	f.BoolVar(&c.all, "all", false, "run all and show variants")
+	f.BoolVar(&c.optimize, "O", true, "run semantic optimizer on ast version")
+	f.BoolVar(&c.debug, "D", false, "display ast version as lisp-y debugger output")
+	f.BoolVar(&c.canon, "C", false, "display canonical version")
+	f.Var(&c.includes, "I", "source file containing Z query text (may be used multiple times)")
```
nwt, line 72:
```suggestion
	f.Var(&c.includes, "I", "source file containing Z query text (may be repeated)")
```

zfmt/canon.go
```diff
@@ -0,0 +1,386 @@
...
+	c.formatter.close()
+}
+
+func (c *canon) assignments(assignments []ast.Assignment) {
+	first := true
+	for _, a := range assignments {
+		if !first {
+			c.write(",")
+		}
+		first = false
```
nwt, line 37:
```suggestion
	for k, a := range assignments {
		if k > 0 {
			c.write(",")
		}
```

zfmt/canon.go
```diff
@@ -0,0 +1,386 @@
...
+}
+
+func (c *canon) exprs(exprs []ast.Expression) {
+	first := true
+	for _, e := range exprs {
+		if first {
+			first = false
+		} else {
+			c.write(", ")
+		}
```
nwt, line 53:
```suggestion
	for k, e := range exprs {
		if k > 0
			c.write(", ")
		}
```

nwt (approved):

PR #2218, opened 2021-02-24
change ztest default format to include -pretty=0

mccanne:
This commit changes the default output format for ztests to
use -pretty=0 so that tests can be written one value per line.
A test can always override by including -pretty=4 etc in
the ztest output-flags field.

### DIFF
--- a/proc/groupby/ztests/dot.yaml
+++ b/proc/groupby/ztests/dot.yaml
@@ -1,6 +1,6 @@
 zql: 'count() by typeof(.) | sort count'
 
-output-flags: -f zson
+output-flags: -pretty=4
 
 input: |
   #0:record[x:int32,s:string]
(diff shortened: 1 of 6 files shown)


proc/groupby/ztests/dot.yaml
```diff
@@ -1,6 +1,6 @@
 zql: 'count() by typeof(.) | sort count'
 
-output-flags: -f zson
+output-flags: -pretty=4
```
nwt, line 3: Should this be `-f zson -pretty=4`?

mccanne: Just pushed a fix.  -f zson is the proposed default so you don't need it here but I added for clarity and consistency.

nwt (approved):

PR #2220, opened 2021-02-24
Fix groupby sorted input nulls order

### LINKED ISSUE #1912: Non-deterministic time-sorted aggregation
Repro is with `zqd` commit `4ec48f88`.

Consider the attached Zeek TSV log from the full 70 GB wrccdc data set. These the notorious `smb_mapping` events that include a count of 26 that lack a `ts` timestamp.

[smb_mapping.log.gz](https://github.com/brimsec/zq/files/5754689/smb_mapping.log.gz)

If I import it to a Space and repeatedly execute a time-sorted aggregation, I find the appearance of the row with the null timestamp can sometimes appear in a different location in the output order.

```
$ zapi -version
Version: v0.26.0-23-g4ec48f88

$ zapi new smb
smb: space created

$ zapi -s smb postpath smb_mapping.log.gz 
100.0% 1.14MB/1.14MB
posted 1.14MB in 196.802638ms

$ zapi -s smb get -t 'every 1w count()'
0:[1521676800;44121;]
0:[-;26;]
0:[1490227200;2620;]

$ zapi -s smb get -t 'every 1w count()'
0:[1521676800;44121;]
0:[1490227200;2620;]
0:[-;26;]
```

Back when I was documenting grouping, @henridf had reminded me of the default reverse-time-ordered nature when the data gets stored into a Space in `zqd`, therefore a user should be able to rely on that time ordering being maintained during a time-based aggregation and hence not have to include an explicit `sort` on `ts`. Those docs are captured at https://github.com/brimsec/zq/tree/master/zql/docs/grouping#note-undefined-order. The behavior observed above runs counter to what's described there, hence I'm treating this as a bug.

incidentally, this is currently affecting our ability to use the Rug automation for correctness tests at scale. With our legacy system we used to drop all incoming records that lacked a timestamp, hence we eliminated those same events from SQL-based reference results as well. However, since `zqd` now stores them, I went to re-allow them in the reference results, but then found the Z-side results now don't consistently match due to this bug. SQL doesn't have this problem since they have a defined treatment of `NULL` which can also be affected by using `ORDER BY ... NULLS FIRST` or `ORDER BY ... NULLS LAST`. Incidentally, we mimicked these options in our own `sort` processor (https://github.com/brimsec/zq/tree/master/zql/docs/processors#sort) but in this case I'm not invoking an implicit `sort`.

To up-level to the wider problem of matching to reference results, it goes without saying that there's other ways we could do this besides addressing the symptom above, e.g.

1. Including an implicit `sort` in the Z query
2. Changing the Z query to exclude the events with null `ts`
3. Have `zqd` start rejecting records with null timestamps like our legacy system used to do
4. Have `zqd` start adding an explicit timestamp at import time in place of the null one (epoch 0, wall clock time, the time of the previously-read event, ...)

etc. It seems like even if we pursued options like 3 & 4 above, it may not be a "one size fits all" answer and hence might require config options and considering the best default. Therefore I figured it makes sense to start from what we have today.

mattnibs:
If a sorted input is in descending order set nullsMax for the compare function
to false. This follows the behavior of the sort proc on sorted inputs where
null values are always printed last.

This fixes a bug when performing a "count every [duration]" that sporadically
caused records with nil time values to get emitted out-of-order.

Closes #1912

### DIFF
--- a/proc/groupby/groupby.go
+++ b/proc/groupby/groupby.go
@@ -80,10 +80,7 @@ func NewAggregator(zctx *resolver.Context, keyRefs, keyExprs, aggRefs []expr.Eva
 		// As the default sort behavior, nullsMax=true for ascending order and
 		// nullsMax=false for descending order is also expected for streaming
 		// groupby.
-		nullsMax := true
-		if inputSortDir < 0 {
-			nullsMax = false
-		}
+		nullsMax := inputSortDir > 0
 
 		vs := expr.NewValueCompareFn(nullsMax)
 		if inputSortDir < 0 {
--- a/proc/groupby/groupby.go
+++ b/proc/groupby/groupby.go
@@ -77,8 +77,15 @@ func NewAggregator(zctx *resolver.Context, keyRefs, keyExprs, aggRefs []expr.Eva
 
 	nkeys := len(keyExprs)
 	if nkeys > 0 && inputSortDir != 0 {
-		// As the default sort behavior, nullsMax=true is also expected for streaming groupby.
-		vs := expr.NewValueCompareFn(true)
+		// As the default sort behavior, nullsMax=true for ascending order and
+		// nullsMax=false for descending order is also expected for streaming
+		// groupby.
+		nullsMax := true
+		if inputSortDir < 0 {
+			nullsMax = false
+		}
+
+		vs := expr.NewValueCompareFn(nullsMax)
 		if inputSortDir < 0 {
 			valueCompare = func(a, b zng.Value) int { return vs(b, a) }
 		} else {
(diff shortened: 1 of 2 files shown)


mattnibs:
@philrz I have run this branch on the smb_mapping data you used to replicate this issue, it appears to fix the problem.

proc/groupby/groupby.go
```diff
@@ -77,8 +77,15 @@ func NewAggregator(zctx *resolver.Context, keyRefs, keyExprs, aggRefs []expr.Eva
...
 	if nkeys > 0 && inputSortDir != 0 {
-		// As the default sort behavior, nullsMax=true is also expected for streaming groupby.
-		vs := expr.NewValueCompareFn(true)
+		// As the default sort behavior, nullsMax=true for ascending order and
+		// nullsMax=false for descending order is also expected for streaming
+		// groupby.
+		nullsMax := true
+		if inputSortDir < 0 {
+			nullsMax = false
+		}
```
nwt, line 86:
```suggestion
		nullsMax := inputSortDir > 0
```

nwt (approved):

PR #2226, opened 2021-02-25
improved display format for Z syntax errors

mccanne:


zql/parser-support.go
```diff
@@ -114,3 +116,56 @@ func makeUnicodeChar(chars interface{}) string {
 
 	return string(r)
 }
+
+var syntaxRegexp = regexp.MustCompile(`^([0-9]+):([0-9]+) \(([0-9]+)\): no match found`)
+
+func ImproveError(src string, e error) error {
+	fmt.Printf("===\n%s===\n", src)
```
nwt, line 123:
```suggestion
```

zql/parser-support.go
```diff
@@ -114,3 +116,56 @@ func makeUnicodeChar(chars interface{}) string {
...
+		return e
+	}
+	if colNo > 100 {
+		return e
+	}
+	_, err = strconv.Atoi(hits[3])
+	if err != nil {
+		return e
+	}
+	lineNo--
```
nwt, line 143:
```suggestion
	el, ok := e.(errList)
	if !ok || len(el) != 1 {
		return e
	}
	pe, ok := el[0].(*parserError)
	if !ok {
		return e
	}
	lineNo, colNo := pe.pos.line-1, pe.pos.col
```

zql/parser-support.go
```diff
@@ -114,3 +116,56 @@ func makeUnicodeChar(chars interface{}) string {
 
 	return string(r)
 }
+
+var syntaxRegexp = regexp.MustCompile(`^([0-9]+):([0-9]+) \(([0-9]+)\): no match found`)
+
+func ImproveError(src string, e error) error {
```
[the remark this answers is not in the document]
mccanne: Good idea but that method is generated by pigeon so I will use `ParseZ`.

zql/parser-support.go
mccanne: oops

zql/parser-support.go
mccanne: doh!  I went to do that and saw the error values weren't exported without thinking this code lives in the same package :-)

compiler/compiler.go
```diff
@@ -13,16 +13,16 @@ import (
 
 // ParseProc() is an entry point for use from external go code,
 // mostly just a wrapper around Parse() that casts the return value.
-func ParseProc(query string, opts ...zql.Option) (ast.Proc, error) {
-	parsed, err := zql.Parse("", []byte(query), opts...)
+func ParseProc(z string) (ast.Proc, error) {
+	parsed, err := zql.ParseZ(z)
 	if err != nil {
-		return nil, err
+		return nil, zql.ImproveError(z, err)
```
nwt, line 19: Don't want this with zql.ParseZ.
```suggestion
		return nil, err
```

nwt (approved):

compiler/compiler.go
mccanne: I'm getting old, haha

PR #2242, opened 2021-02-26
add zson_parse function

alfred-landrum:
We'll want a way to make metadata available to the shapers that run for an intake. One way we could do that is by creating a ZSON definition of the metadata record, then prepending the `const` definition of that record to the shaper's source. That requires a way to parse ZSON inside of Z, so I put this up to get feedback on the overall approach.

### DIFF
--- a/zql/ztests/zson-parse.yaml
+++ b/zql/ztests/zson-parse.yaml
@@ -1,0 +1,23 @@
+script: |
+  zq -f tzng -I def.zql in.tzng
+
+inputs:
+  - name: def.zql
+    data: |
+      const r=zson_parse('{a:"1",b:2}')
+      put ra=r.a,rb=r.b
+
+  - name: in.tzng
+    data: |
+      #0:record[v:int64]
+      0:[1;]
+      0:[2;]
+      0:[3;]
+
+outputs:
+  - name: stdout
+    data: |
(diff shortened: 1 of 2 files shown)


mccanne (approved):
perfect!

alfred-landrum:
> perfect!

cool :) I'll let it sit up for review for a bit, but marked this as ready for review & removed the 'not for merge' comment.

zql/ztests/zson-parse.yaml
```diff
@@ -0,0 +1,23 @@
...
+      0:[2;]
+      0:[3;]
+
+outputs:
+  - name: stdout
+    data: |
+      #0:record[v:int64,ra:string,rb:int64]
+      0:[1;1;2;]
+      0:[2;1;2;]
+      0:[3;1;2;]
```
nwt, line 23:
```suggestion
zql: |
  const r=zson_parse('{a:"1",b:2}')
  put ra=r.a,rb=r.b

input: |
  {v:1}
  {v:2}
  {v:3}

output: |
  {v:1,ra:"1",rb:2}
  {v:2,ra:"1",rb:2}
  {v:3,ra:"1",rb:2}
```

nwt (approved):

PR #2246, opened 2021-02-26
Dockerfile: move "ARG LDFLAGS" to improve caching

nwt:
Move "ARG LDFLAGS" after "RUN go mod download" to increase the  
likelihood of a build cache hit for the latter, which is slow.

@mattnibs noticed that `go mod download` executes more often than expected during `make docker`. This should help.


alfred-landrum (approved):
nice!

Issue #2207, opened 2021-02-24
slice index out of range panic in zeekio tsv parser


mccanne:
This zeek TSV comes from the zeekio trailing-whitespace.yaml test:
```
#separator \x09
#set_separator	,
#empty_field	(empty)
#unset_field	-
#path	unknown_mime_type_discovery
#open	2020-04-10-18-15-32
#fields	ts	fid	bof
#types	time	string	string
1425634708.454143	FTBt2P3wJBdHGKpZwc
```
Curiously it causes a panic when run wtih `zq -f tzng in.zeek` on master but the unit test on master passes so there is something about the way the ztest module is running this that is avoiding the bug.  I'm disable the test on the dash-z branch and would like to get that merged while this bug can be fixed later...

```
panic: runtime error: index out of range [2] with length 2

goroutine 1 [running]:
github.com/brimsec/zq/zio/zeekio.(*builder).build(0xc0000cf9b0, 0xc000380ea0, 0xc0003595c0, 0x3, 0x4, 0xc00019f170, 0x1b, 0x20, 0xc0003960ae, 0x24, ...)
	/Users/mccanne/repo/zq/zio/zeekio/builder.go:44 +0x8f0
github.com/brimsec/zq/zio/zeekio.(*Parser).ParseValue(0xc0000cf900, 0xc0003960ae, 0x24, 0xff52, 0x0, 0x0, 0x10)
	/Users/mccanne/repo/zq/zio/zeekio/parser.go:329 +0xe5
github.com/brimsec/zq/zio/zeekio.(*Reader).Read(0xc0003796e0, 0x18614c0, 0xc0003796e0, 0x10000)
	/Users/mccanne/repo/zq/zio/zeekio/reader.go:63 +0x205
github.com/brimsec/zq/zio/detector.match(0x1a85540, 0xc0003796e0, 0x191aab2, 0x4, 0x0, 0x0)
	/Users/mccanne/repo/zq/zio/detector/reader.go:104 +0x38
github.com/brimsec/zq/zio/detector.NewReaderWithOpts(0x1a854a0, 0xc000380c60, 0xc0000106e8, 0x7ffeefbffac6, 0x7, 0x191a092, 0x4, 0x1, 0x80000, 0xa00000, ...)
	/Users/mccanne/repo/zq/zio/detector/reader.go:34 +0x595
[...]

PR #2248, opened 2021-02-27
enable zio/zeekio/ztests/trailing-whitespace.yaml

nwt:
I can't reproduce the panic in #2207, so I think it's safe to enable this test and close the issue.


mccanne (approved):
Test passes for me too.  I must have done something temporarily goofy to have triggered the problem.  Weird.

PR #2249, opened 2021-02-27
Makefile: improve bin/minio rule

nwt:
This approach follows the one used for bin/tctl and avoids rebuilding  
bin/minio every time "make test-system" is run.


henridf (approved):
🥳

PR #2251, opened 2021-02-27
use ztest for expr/expr_test.go

mccanne:
This commit changes expr_test.go to use ztest instead of reaching
into the compiler library to compile a standalone expression.
This arrangement is easier to maintain and debug failing tests.

This is needed by a subsequent PR that is changing the compiler
internals.

Also, we refactored ztest a bit so it's easier to run an internal
test from the outside.

### DIFF
--- a/ztest/ztest.go
+++ b/ztest/ztest.go
@@ -327,6 +327,11 @@ func (z *ZTest) check() error {
 	} else {
 		return errors.New("either a zql field or script field must be present")
 	}
+	if z.ErrorRE != "" {
+		var err error
+		z.errRegex, err = regexp.Compile(z.ErrorRE)
+		return err
+	}
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


ztest/ztest.go
```diff
@@ -425,47 +430,57 @@ func (z *ZTest) ShouldSkip(path string) string {
 }
 
 func (z *ZTest) RunScript(testname, shellPath, workingDir, filename string) error {
+	if err := z.check(); err != nil {
+		return fmt.Errorf("%s: bad yaml format: %s", filename, err)
```
nwt, line 434:
```suggestion
		return fmt.Errorf("%s: bad yaml format: %w", filename, err)
```

ztest/ztest.go
```diff
@@ -327,6 +327,11 @@ func (z *ZTest) check() error {
 	} else {
 		return errors.New("either a zql field or script field must be present")
 	}
+	if z.ErrorRE != "" {
+		var err error
+		z.errRegex, err = regexp.Compile(z.ErrorRE)
```
nwt, line 332: This renders the `regexp.Compile` call in `FromYAMLFile` redundant, so maybe remove that?

nwt (approved):

PR #2268, opened 2021-03-02
improve zeekio.Reader performance with zng.Append*

nwt:
Using zng.Append* instead of zng.Encode* eliminates lots of runtime.newobject
calls for byte arrays.

Branch master at 8aade97516c02cecf2fe80755bc697bea182c2e0:
```sh
$ make build && for i in {1..5}; do /usr/bin/time ./dist/zq 'count()' zq-sample-data/zeek-default/* > /dev/null; done
        2.96 real         5.41 user         0.29 sys
        2.80 real         5.39 user         0.28 sys
        2.76 real         5.38 user         0.29 sys
        2.76 real         5.36 user         0.28 sys
        2.76 real         5.38 user         0.28 sys
```

Branch zng.AppendX at b3e742c7721860a372522bd66fd4757cb464282d:
```sh
$ make build && for i in {1..5}; do /usr/bin/time ./dist/zq 'count()' zq-sample-data/zeek-default/* > /dev/null; done
        2.70 real         4.91 user         0.26 sys
        2.47 real         4.92 user         0.26 sys
        2.46 real         4.87 user         0.25 sys
        2.47 real         4.89 user         0.26 sys
        2.49 real         4.92 user         0.26 sys
```

### DIFF
--- a/zng/net.go
+++ b/zng/net.go
@@ -13,21 +13,20 @@ func NewNet(s *net.IPNet) Value {
 	return Value{TypeNet, EncodeNet(s)}
 }
 
-func EncodeNet(subnet *net.IPNet) zcode.Bytes {
-	var b [32]byte
-	ip := subnet.IP.To4()
-	if ip != nil {
-		copy(b[:], ip)
+func AppendNet(zb zcode.Bytes, subnet *net.IPNet) zcode.Bytes {
+	if ip := subnet.IP.To4(); ip != nil {
+		zb = append(zb, ip...)
 		if len(subnet.Mask) == 16 {
-			copy(b[4:], subnet.Mask[12:])
-		} else {
-			copy(b[4:], subnet.Mask)
+			return append(zb, subnet.Mask[12:]...)
 		}
-		return b[:8]
+		return append(zb, subnet.Mask...)
 	}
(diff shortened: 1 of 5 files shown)


zng/net.go
```diff
@@ -13,21 +13,20 @@ func NewNet(s *net.IPNet) Value {
 	return Value{TypeNet, EncodeNet(s)}
 }
 
-func EncodeNet(subnet *net.IPNet) zcode.Bytes {
-	var b [32]byte
-	ip := subnet.IP.To4()
-	if ip != nil {
-		copy(b[:], ip)
+func AppendNet(zb zcode.Bytes, subnet *net.IPNet) zcode.Bytes {
```
alfred-landrum, line 16: Regarding the existing code, not your changes: what's the reason behind the `len(subnet.Mask) == 16` check?

alfred-landrum (approved):

zng/net.go
nwt: `net.IPnet` has no `net.IP.To4` equivalent, so this is checking whether the mask is already a 4-byte IPv4 mask or, alternatively, a 16-byte IPv4-mapped IPv6 mask that must be trimmed to a 4-byte IPv4 mask.

PR #2277, opened 2021-03-03
Unflatten() function

### LINKED ISSUE #2239: Add unflatten Z function
Just like we moved shaping functionality from ad hoc sideshows (types.json) to Z, we should add `unflatten()` to Z so that it can be done as part of a Z shaper script rather than in readers.

---
### Side commentary from @philrz

From a user-facing perspective, "unflatten" in this context refers to what we've traditionally done with turning dot-separated field names back into nested records, due to our Zeek-centric roots. This behavior is still currently the default with explicit `-i ndjson`:

```
$ zq -version
Version: v0.29.0-71-g5cd41763

$ echo '{"a.b": 1}' | zq -f zson -i ndjson -
{
    a: {
        b: 1.
    }
}
```

However, in the recently-added NDJSON ingest that's done via auto-detect, it's not currently the default:

```
$ echo '{"a.b": 1}' | zq -f zson -
{
    "a.b": 1
}
```

henridf:
This PR adds an unflatten() function that turns successive dotted field names into nested records.  For example, unflattening `{"a.a": 1, "a.b": 1}` results in `{a:{a:1,b:1}}`.  

Note that while unflattening is applied recursively from the top-level and applies to arbitrary-depth dotted names, it is not applied to dotted names that start at lower levels (for example {a:{"a.a":1}} is unchanged). This could be added but seems unnecessary given that the sole current use case for this function is zeek json records, where dottedness always starts at the root.

close #2239

### DIFF
--- a/expr/unflattener.go
+++ b/expr/unflattener.go
@@ -1,0 +1,101 @@
+package expr
+
+import (
+	"github.com/brimsec/zq/field"
+	"github.com/brimsec/zq/zng"
+	"github.com/brimsec/zq/zng/builder"
+	"github.com/brimsec/zq/zng/resolver"
+)
+
+type Unflattener struct {
+	zctx        *resolver.Context
+	builders    map[int]*builder.ColumnBuilder
+	recordTypes map[int]*zng.TypeRecord
+	fieldExpr   Evaluator
+}
+
+// NewUnflattener returns a Unflattener that turns successive dotted
+// field names into nested records.  For example, unflattening {"a.a":
+// 1, "a.b": 1} results in {a:{a:1,b:1}}.  Note that while
+// unflattening is applied recursively from the top-level and applies
+// to arbitrary-depth dotted names, it is not applied to dotted names
+// that start at lower levels (for example {a:{"a.a":1}} is
+// unchanged).
+func NewUnflattener(zctx *resolver.Context) *Unflattener {
+	return &Unflattener{
+		zctx:        zctx,
+		builders:    make(map[int]*builder.ColumnBuilder),
+		recordTypes: make(map[int]*zng.TypeRecord),
+	}
+}
+
+func (u *Unflattener) lookupBuilderAndType(in *zng.TypeRecord) (*builder.ColumnBuilder, *zng.TypeRecord, error) {
+	b, ok := u.builders[in.ID()]
+	if ok {
+		b.Reset()
+		return b, u.recordTypes[in.ID()], nil
+	}
+	cols := in.Columns
+	var foundDotted bool
+	var fields []field.Static
+	var types []zng.Type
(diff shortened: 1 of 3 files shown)


expr/unflattener.go
```diff
@@ -0,0 +1,101 @@
...
+	return &Unflattener{
+		zctx:        zctx,
+		builders:    make(map[int]*builder.ColumnBuilder),
+		recordTypes: make(map[int]*zng.TypeRecord),
+	}
+}
+
+func (u *Unflattener) lookupBuilderAndType(in *zng.TypeRecord) (*builder.ColumnBuilder, *zng.TypeRecord, error) {
+	b, ok := u.builders[in.ID()]
+	if ok {
```
nwt, line 34:
```suggestion
	if b, ok := u.builders[in.ID()]; ok {
```

expr/unflattener.go
```diff
@@ -0,0 +1,101 @@
...
+		b.Reset()
+		return b, u.recordTypes[in.ID()], nil
+	}
+	cols := in.Columns
+	var foundDotted bool
+	var fields []field.Static
+	var types []zng.Type
+	for _, c := range cols {
```
nwt, line 42:
```suggestion
	var foundDotted bool
	var fields []field.Static
	var types []zng.Type
	for _, c := range in.Columns {
```

expr/unflattener.go
```diff
@@ -0,0 +1,101 @@
...
+		zctx:        zctx,
+		builders:    make(map[int]*builder.ColumnBuilder),
+		recordTypes: make(map[int]*zng.TypeRecord),
+	}
+}
+
+func (u *Unflattener) lookupBuilderAndType(in *zng.TypeRecord) (*builder.ColumnBuilder, *zng.TypeRecord, error) {
+	b, ok := u.builders[in.ID()]
+	if ok {
+		b.Reset()
```
nwt, line 35: NIt: I think you should do this in Apply rather than here so it's clear there that the builder is being reset before use.

expr/unflattener.go
```diff
@@ -0,0 +1,101 @@
...
+func (u *Unflattener) Apply(in *zng.Record) (*zng.Record, error) {
+	b, typ, err := u.lookupBuilderAndType(in.Type)
+	if err != nil {
+		return nil, err
+	}
+	if b == nil {
+		return in, nil
+	}
+	iter := in.Raw.Iter()
+	for !iter.Done() {
```
nwt, line 78:
```suggestion
	for iter := in.Raw.Iter(); !iter.Done(); {
```

nwt (approved):

expr/unflattener.go
henridf: 👍 agree

PR #2282, opened 2021-03-03
Fix distributed searches

mattnibs:
In order for a search to run as distributed, the Distributed field in
driver.MultiConfig must be set to true.

### DIFF
--- a/ppl/zqd/search/search.go
+++ b/ppl/zqd/search/search.go
@@ -86,6 +86,7 @@ func (s *SearchOp) Run(ctx context.Context, store storage.Storage, output Output
 	switch st := store.(type) {
 	case *archivestore.Storage:
 		return driver.MultiRun(ctx, d, s.query.Proc, zctx, st.MultiSource(), driver.MultiConfig{
+			Distributed: parallelism > 0,
 			Logger:      s.logger,
 			Order:       zbuf.OrderDesc,
 			Parallelism: parallelism,


ppl/zqd/search/search.go
```diff
@@ -86,6 +86,7 @@ func (s *SearchOp) Run(ctx context.Context, store storage.Storage, output Output
 	switch st := store.(type) {
 	case *archivestore.Storage:
 		return driver.MultiRun(ctx, d, s.query.Proc, zctx, st.MultiSource(), driver.MultiConfig{
+			Distributed: distributed,
```
nwt, line 89: How about this instead of adding a parameter?

```suggestion
			Distributed: parallelism > 0,
```

mattnibs: That works. There's a knot here that needs to be untangled.

mattnibs: You could conceivably pass in parallelism > 0 intending to run on a single host with multiple scanners, but we aren't currently doing this.

nwt: Right. I vote for crossing that bridge when we come to it.

By the way, this breakage is all my fault because I left this out of #2178.

mattnibs: I'm aware 😄

nwt (approved):

PR #2289, opened 2021-03-04
Accept vars and aliases in shaper functions

henridf:
Previously the shaper functions required a literal type as second parameter, because they predated const and type literals. Now that those exist, the functions should also accept vars and type expressions. This commit does that.

### DIFF
--- a/expr/shaper.go
+++ b/expr/shaper.go
@@ -246,22 +246,30 @@ func NewShaperType(zctx *resolver.Context, fieldExpr Evaluator, typ *zng.TypeRec
 // to the provided typExpr. (typExpr should evaluate to a type value,
 // e.g. a value of type TypeType).
 func NewShaper(zctx *resolver.Context, fieldExpr, typExpr Evaluator, tf ShaperTransform) (*Shaper, error) {
-	lit, ok := typExpr.(*Literal)
-	if !ok {
-		return nil, fmt.Errorf("shaping functions (crop, fill, cast, order) take a literal as second parameter")
+	switch typExpr.(type) {
+	case *Var, *TypeFunc, *Literal:
+	default:
+		return nil, fmt.Errorf("shaping functions (crop, fill, cast, order) require a type value as second parameter")
 	}
-
-	if lit.zv.Type != zng.TypeType {
-		return nil, fmt.Errorf("shaper needs a type value as second parameter")
+	typVal, err := typExpr.Eval(nil)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


nwt:
Looks like #2290 was merged into this branch. Do you want to drop that commit and reopen the PR?

henridf:
@nwt PRs can't be reopened, but I can create a new one with that commit after this PR goes in. I've removed the commit from this branch.

expr/shaper.go
```diff
@@ -246,22 +246,30 @@ func NewShaperType(zctx *resolver.Context, fieldExpr Evaluator, typ *zng.TypeRec
 // to the provided typExpr. (typExpr should evaluate to a type value,
 // e.g. a value of type TypeType).
 func NewShaper(zctx *resolver.Context, fieldExpr, typExpr Evaluator, tf ShaperTransform) (*Shaper, error) {
-	lit, ok := typExpr.(*Literal)
-	if !ok {
-		return nil, fmt.Errorf("shaping functions (crop, fill, cast, order) take a literal as second parameter")
+	switch typExpr.(type) {
+	case *Var, *TypeFunc, *Literal:
+	default:
+		return nil, fmt.Errorf("shaping functions (crop, fill, cast, order) require a type value as second parameter")
```
nwt, line 252: Nit: This function has some error strings that begin with "shaping functions (crop, fill, cast, order)" and others that begin with "shaper". Either's fine but consistency would be good.

expr/shaper.go
```diff
@@ -246,22 +246,30 @@ func NewShaperType(zctx *resolver.Context, fieldExpr Evaluator, typ *zng.TypeRec
...
+	}
+	if typVal.Type != zng.TypeType {
+		return nil, fmt.Errorf("shaping functions (crop, fill, cast, order) require a type value as second parameter")
+	}
+	s, err := zng.DecodeString(typVal.Bytes)
+	if err != nil {
+		return nil, err
 	}
-	shapeToType, err := zctx.Context.LookupByName(string(lit.zv.Bytes))
+	shapeToType, err := zctx.Context.LookupByName(s)
```
nwt, line 265: Maybe just call this `shapeType`?

nwt (approved):

expr/shaper.go
henridf: done

expr/shaper.go
henridf: done

PR #2294, opened 2021-03-04
More ZQL doc updates after search/expr change

philrz:
The search/expr changes in #2072 introduced new functionality (largely described in a draft in [https://github.com/brimsec/zq/blob/master/zql/expr.md)](https://github.com/brimsec/zq/blob/master/zql/expr.md)) while trying to maintain most of what had been working to date. While some of the mainline docs were updated in that PR, I've done another pass through here and made several additional changes to bring them in sync with how things are currently working.

I'm aware that the shelf life of these Zeek-centric docs is limited, as we've expressed an intent to move a limited amount of Zeek-centric language docs elsewhere (the [Brim wiki](https://github.com/brimsec/zq/blob/master/zql/expr.md) seems a likely destination, as Zeek-centric functionality is likely to be discussed more in the context of the app going forward) and replace these docs with something based on more generic data. However, since the docs have kept mostly correct to date, I figure it'd be helpful to have a point-in-time commit in the repo where these docs are still relevant for the Zeek-centric case, just in case anyone (even if only me) has reason to refer back to them.

### DIFF
--- a/zql/docs/search-syntax/README.md
+++ b/zql/docs/search-syntax/README.md
@@ -262,7 +262,7 @@ When working with named fields, the data type of the field becomes significant i
 
    An attempted field/value match `host=10.47.21.1` would not match the event counted in the middle row of this table, since ZQL recognizes the bare value `10.47.21.1` as an IP address before comparing it to all the fields named `host` that it sees in the input stream. However, `host="10.47.21.1"` would match, since the quotes cause ZQL to treat the value as a string.
 
-2.  The correct operator must be chosen based on whether the field type is primitive or complex.  For example, `id.resp_h=10.150.0.85` will match in our sample data because `id.resp_h` is a primitive type, `addr`. However, to check if the same IP had been a transmitting host in a `files` event, the syntax `10.150.0.85 in tx_hosts` would be used because `tx_hosts` is a complex type, `set[addr]`.
+2.  The correct operator must be chosen based on whether the field type is primitive or complex.  For example, `id.resp_h=10.150.0.85` will match in our sample data because `id.resp_h` is a primitive type, `addr`. However, to check if the same IP had been a transmitting host in a `files` event, the syntax `10.150.0.85 in tx_hosts` would be used because `tx_hosts` is a complex type, `set[addr]`. See the section below on [Containment](#containment) for details in the `in` operator.
 
 See the [Data Types](../data-types/README.md) page for more details on types and the operators for working with them.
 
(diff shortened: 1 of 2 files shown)


zql/docs/search-syntax/README.md
```diff
@@ -20,58 +21,53 @@
...
+zq -z conn.log.gz
 ```
 
 To start a ZQL pipeline with this default search, you can similarly leave out the leading `* |` before invoking your first [processor](#../processors/README.md) or [aggregate function](#../aggregate-functions/README.md).
 
 #### Example #1:
 ```zq-command
-zq -f tzng 'cut server_tree_name' ntlm.log.gz # Shorthand for: zq '* | cut server_tree_name' ntlm.log.gz
+zq -z 'cut server_tree_name' ntlm.log.gz  # Shorthand for: zq -z '* | cut server_tree_name' ntlm.log.gz
 ```
```
marktwallace, line 51: It looks a little weird having the # comment that contains a pipe | character. It makes it easy to miss the #, so splitting the line would be better.

zql/docs/search-syntax/README.md
```diff
@@ -318,7 +315,45 @@ http  2018-03-24T17:17:47.561097Z CgbtuX3gXoYFmEF82l 10.164.94.120 37311     10.
...
-Determining whether the value of a Zeek `addr`-type field is within a subnet also uses the pattern matching operator.
+### Containment
+
+Rather than testing for strict equality or pattern matches, you may want to determine if a value is among the many possible elements of a complex field. This is performed with the `in` operator.
+
+Our Zeek `dns` events include the `answers` field, which is an array of the multiple responses that may have been returned for a query. To determine which responses included hostname `e5803.b.akamaiedge.net`, we'll use `in`.
+
+#### Example:
+```zq-command
+zq -f table '"e5803.b.akamaiedge.net" in answers' dns.log.gz
```
marktwallace, line 326: Just a side comment, so no problem here, but the double-quotes in single quotes pattern looks bad. I don't have a better suggestion ATM.

marktwallace (approved):
This looks good, too. The greatly reduced complexity of your explanations is a good indicator of how much easier it is to talk about ZSON.

I also have to throw in: until I read through this doc, I did not realize what a relief it would be to stop reading TZNG. I think it was the trailing semicolon followed by square bracket that kind of bored me to sleep when reading TZNG syntax. :-)

mccanne (approved):
looks good!

PR #2297, opened 2021-03-05
Support signed-offset ISO8601's in iso()

### LINKED ISSUE #1905: Timestamp offset ±[hh][mm] notation results in "iso: bad argument"
Repro is with `zq` commit `4dff9e90`.

I wanted to try my hand at some of the basic shaping concepts enabled by #1833, so the first thing I tried was importing a Suricata EVE JSON log. Out-of-the-box, the timestamps in the alert records have the ISO format with the timezone expressed as an offset form UTC. However, when I attempt to apply the `iso()` function to turn it into a ZNG `time` value, I get an error.

```
$ echo '{"timestamp":"2020-03-26T11:03:29.986905-0700"}' | zq -t 'put ts=iso(timestamp)' -
iso: bad argument
```

I can see the notation with the colon between hours & minutes _does_ appear to work, though.

```
$ echo '{"timestamp":"2020-03-26T11:03:29.986905-07:00"}' | zq -t 'put ts=iso(timestamp)' -
#0:record[timestamp:string,ts:time]
0:[2020-03-26T11:03:29.986905-07:00;1585245809.986905;]
```

The summary at https://en.wikipedia.org/wiki/ISO_8601#Time_offsets_from_UTC implies that both syntaxes are legal.

henridf:
This variant is seen in Suricata logs.

Fixes #1905.


expr/function/time.go
```diff
@@ -21,8 +22,16 @@ func (i *iso) Call(args []zng.Value) (zng.Value, error) {
...
+	var ts time.Time
+	var err error
+	switch {
+	case bytes.Contains(zv.Bytes, []byte{'Z'}):
+		ts, err = time.Parse(time.RFC3339Nano, string(zv.Bytes))
+	case bytes.Contains(zv.Bytes, []byte{'-'}):
+		// ISO8601 with signed offset
+		ts, err = time.Parse("2006-01-02T15:04:05.999999999-0700", string(zv.Bytes))
+	}
+	if err != nil {
```
nwt, line 34: This isn't quite right. `time.RFC3339Nano` handles ISO 8601 timestamps whose time zone is `Z` or an offset containing a colon. What we need to add is support for an offset without a colon. This would do it.
```
	// Handles ISO 8601 with time zone of Z or an offset not containing a colon.
	format := "2006-01-02T15:04:05.999999999Z0700"
	if l := len(zv.Bytes); l > 2 && zv.Bytes[:l-3] != ':' {
		 // Handles ISO 8601 with time zone of Z or an offset containing a colon.
		layout = time.RFC3339Nano
	ts, err := time.Parse(format, string(zv.Bytes))
	if err != nil {
```

expr/function/ztests/iso.yaml
```diff
@@ -0,0 +1,8 @@
+zql: |
+   put iso=iso(iso), signed=iso(signed)
+
+input: |
+  {"iso": "2018-03-24T17:15:21.255387Z", "signed": "2015-03-05T16:07:13.154566+0100"}
```
nwt, line 5:
```suggestion
  {"z": "2018-03-24T17:15:21.255387Z", "offsetWithoutColon": "2015-03-05T16:07:13.154566+0100", "offsetWithColon": "2015-03-05T16:07:13.154566+01:00"}
```

expr/function/time.go
henridf: thanks! fixed

nwt (approved):

PR #2306, opened 2021-03-06
Temporal helm

marktwallace:
The initial goal of the PR is to add a Temporal subchart to the Brim zservice “umbrella” Helm chart that we use for our zq-test cluster and automated tests. It also includes automated tests of the Temporal deployment and the zqd integration with Temporal.

Temporalio has developed a Helm chart that is relatively new (a few months old) compared to the overall project (which is years old). Their Helm chart is straightforward and easy to customize, but Temporal does not support its use as a subchart. So, in order to incorporate Temporal into our overall deployment, we needed to develop and test a pattern for using Temporal’s Helm chart as a subchart within the zservice chart. 

One of the first things we needed was Helm repository from which we could reference `https://github.com/temporalio/helm-charts`. A Helm repository requires, at minimum, a URL for a tgz file and an index.yaml file with metadata for the chart and it dependencies. We created this within a new github repo, brimsec/helm-chart. This PR has a dependency on the new brimsec/helm-chart repo. 

The procedure we used to create brimsec/helm-chart is documented in k8s/temporal.md. We added temporalio/helm-charts as a submodule of brimsec/helm-chart. We explicitly decided NOT to fork temporalio/helm-charts. We can and will fork temporalio/helm-charts if we need changes outside the capabilities of the current chart. Although temporalio/helm-charts follows some conventions that are different from ours (e.g. database password in environment variable) we will avoid forking it until we have to.

Temporal requires occasional database migrations. The procedures for their migrations are not automated. Because of this there is some manual setup for adding new test instances of Temporal. These are also documented in k8s/temporal.md. Without the detailed guide in k8s/temporal.md, the setup is difficult to get right, because there are multiple dependencies between steps. Temporal's complexity is hidden in the all-in-one docker image that we use for the ZTests in make test-system, but becomes more apparent the the K8s deployment that we use for the zq-test cluster and our future production environment.

Temporal has a concept of namespaces. Currently temporal namespaces require an external tool called `tctl` to set up. We may eventually be able to do this in our zqd code, but for now, follow the namespace setup instructions in k8s/temporal.md.

Minor changes to the zqd personality “temporal” were required to support K8s deployment. I think the temporal personality should have a reduced API, but I am not yet sure what we can remove.

The deployed footprint of the zservice chart is significantly larger after this PR. Seven new deployments and services have been added. (Use `kubectl get pod -A` to see the new pods in ci-master.)

A temporal cluster Ztest similar to the temporal system Ztest has been added to cd.yaml.

The new Makefile rules, test-temporal and helm-install-with-aurora-temporal depend on environment variables that have been added to cd.yaml and are described in k8s/temporal.md.


ppl/zqd/core.go
```diff
@@ -159,6 +159,8 @@ func NewCore(ctx context.Context, conf Config) (*Core, error) {
 	case "recruiter":
 		c.workerPool = recruiter.NewWorkerPool()
 		c.addRecruiterRoutes()
+	case "temporal":
+		c.addAPIServerRoutes()
```
nwt, line 163: Why is this necessary?

marktwallace: It was necessary for the Ztest "compact" to run. I was expecting that the zqd-temporal workers would need different routes. The existing compact test was working personality "all" and depended on APIServer routes.

nwt: Please remove this and change the test so `zapi` talks to the API server instead of the Temporal worker.

marktwallace: OK, so the APIServer also has access to the Temporal API's for added work to the Temporal queue, and the zqd-temporal worker will only dequeue work items added by the APIServer. Is that correct?

nwt: Yes.

.github/workflows/cd.yaml
```diff
@@ -106,20 +106,26 @@ jobs:
           ref: ${{ github.event.pull_request.head.sha }}
       - run: |
           set -x
+          aws eks --region us-east-2 update-kubeconfig --name zq-test
+          kubectl create ns ci-master || true
+          kubectl config set-context --current --namespace=ci-master
           export ZQD_AURORA_USER=cimaster
+          export ZQD_AURORA_PW=$(kubectl get secret aurora --template="{{ index .data \"postgresql-password\" }}" | base64 --decode)
```
nwt, line 113:
```suggestion
          export ZQD_AURORA_PW=$(kubectl get secret aurora --template='{{ index .data "postgresql-password" }}' | base64 --decode)
```

Makefile
```diff
@@ -195,6 +202,28 @@ helm-install-with-aurora:
...
+	--set temporal.server.config.persistence.default.sql.database=$(TEMPORAL_DATABASE) \
+	--set temporal.server.config.persistence.default.sql.user=$(ZQD_AURORA_USER) \
+	--set temporal.server.config.persistence.default.sql.password=$(ZQD_AURORA_PW) \
+	--set temporal.server.config.persistence.default.sql.host=$(ZQD_AURORA_HOST) \
+	--set temporal.server.config.persistence.visibility.sql.database=$(TEMPORAL_VISIBILITY_DATABASE) \
+	--set temporal.server.config.persistence.visibility.sql.user=$(ZQD_AURORA_USER) \
+	--set temporal.server.config.persistence.visibility.sql.password=$(ZQD_AURORA_PW) \
+	--set temporal.server.config.persistence.visibility.sql.host=$(ZQD_AURORA_HOST)
+
+
```
nwt, line 226:
```suggestion

```

.github/workflows/cd.yaml
marktwallace: Single quotes in bash do not interpolate, and are not interchangeable with double quotes. Even if single-quotes work in this line of bash, they would be confusing since the expression wrapped by {{ }} looks like a form of interpolation that is specific to GitHub yaml.

ppl/zqd/core.go
marktwallace: After making this change, the cluster-compact test fails because it is no longer performing compaction. Maybe the root service as it is currently built does not use the new Temporal API?

marktwallace: The resolution to this was to remove the APIServerRoutes from the temporal worker, but keep the case statement so the instance would start. In addition, we added temporal.enabled and other startup flags to the root service.

nwt (approved):

PR #2309, opened 2021-03-08
Remove cropping from shape()

### LINKED ISSUE #2308: Remove crop from shape
Currently, `shape()` is the composition of the four primitive shaping functions cast, fill, crop, and order. Having now implemented some shaper scripts, recalled #1841, and consulted with @philrz, it seems like it would be better for `shape()` not to crop. This makes `shape()` less strict and more in-line with the overall schema flexibility of the system, as opposed to discarding any extra fields from the get-go. The fact that `shape()`  will likely be the default go-to shaping function for most users makes this change especially relevant. Of course, `crop()` remains available as a standalone function for situations that require it.

henridf:
close #2308

### DIFF
--- a/expr/ztests/shape-nested.yaml
+++ b/expr/ztests/shape-nested.yaml
@@ -12,4 +12,6 @@ output: |
   #0:record[x:record[a:record[a:float64,b:float64]]]
   0:[[[1;-;]]]
   0:[[[-;1;]]]
-  0:[[[-;1;]]]
+  #1:record[x:record[a:record[a:float64,b:float64,c:int64]]]
+  1:[[[-;1;1;]]]
+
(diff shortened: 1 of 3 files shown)


expr/ztests/shape-nested.yaml
```diff
@@ -12,4 +12,6 @@ output: |
   #0:record[x:record[a:record[a:float64,b:float64]]]
   0:[[[1;-;]]]
   0:[[[-;1;]]]
-  0:[[[-;1;]]]
+  #1:record[x:record[a:record[a:float64,b:float64,c:int64]]]
+  1:[[[-;1;1;]]]
+
```
nwt, line 17:
```suggestion
```

nwt (approved):

PR #2314, opened 2021-03-09
rename proc.Context.TypeContext to Zctx

### LINKED ISSUE #1876: Rename proc.Context.TypeContext to Zctx
We use the name `zctx` or, occasionally, `sctx` for every field, parameter, or variable of type `*zng/resolver.Context` except for `proc.Context.TypeContext`. We should rename that field to `Zctx` for consistency.

nwt:
Closes #1876.


marktwallace (approved):
This all looks good. I like the new name better.

mccanne (approved):
Now let's get rid of zng/resolver :-)

Issue #2316, opened 2021-03-10
strict-mode csv in auto-detection


mccanne:
Currently, the csv reader is too forgiving and accepts inputs aren't really csv, e.g.,
```
#0:record[_path:string]
0:[conn;1;]
``` 
parses as a csv file with header field `#0:record[_path:string]` and a single value `0:[conn;1;]`.

This causes several of our tests to fail.

A fix would be to have a strict mode that the auto-detector would use, for instance, to require that the CSV field names be identifiers and if the strict-mode auto-detector is too fussy in a corner case, you can always specify `-i csv`. 

This would probably be robust enough to avoid all the test failures.

mccanne:
Closgin as the CSV autodetector is now robust...

PR #2317, opened 2021-03-10
add csv reader

### LINKED ISSUE #999: CSV reader
A user on the Brim public Slack [inquired](https://brimsec.slack.com/archives/C01092MJ9T6/p1594570698005500):

> How can one ingest logs of other formats into Zq-supported format(s)? Say, as the perhaps simplest example, one has a .csv in tabular form (first row column names, followed by value rows, and comma separated etc.), what would you suggest convert these files to Zq formats?

To get them up and running quickly, I noted the availability of tools like [`csvtojson`](https://www.npmjs.com/package/csvtojson`) and showed how it could be used upstream of `zq`. However, for ease of use, a direct CSV input option could be added to `zq`.

mccanne:
This commit adds a reader for CSV formatted input.  It uses the
Go csv package, which conforms to RFC 4180.  CSV is not yet tied
into the auto-detector as it is currently too forgiving of a format
and accepts inputs that are clearly not intended as CSV.  See
issue #2316  for more info.

Closes #999.

### DIFF
--- a/zio/csvio/reader.go
+++ b/zio/csvio/reader.go
@@ -1,0 +1,109 @@
+package csvio
+
+import (
+	"encoding/csv"
+	"errors"
+	"io"
+	"math"
+	"strconv"
+	"strings"
+
+	"github.com/brimsec/zq/zng"
+	"github.com/brimsec/zq/zson"
+)
+
+type Reader struct {
+	zctx      *zson.Context
+	reader    *csv.Reader
+	marshaler *zson.MarshalZNGContext
+	strings   bool
+	valid     bool
+	hdr       []string
+	vals      []interface{}
+}
+
+// XXX This is a placeholder for an option that will allow one to convert
+// all csv fields to strings and defer any type coercion presumably to
+// Z shapers.  Currently, this causes an import cycle because the csvio
+// Writer depends on fuse.  We should refactor this so whatever logic wants
+// to tack on a fuse operator happens outside of zio.  See issue #2315
+//type ReaderOpts struct {
+//	StringsOnly bool
+//}
+
+func NewReader(reader io.Reader, zctx *zson.Context) *Reader {
+	return &Reader{
+		zctx:      zctx,
+		reader:    csv.NewReader(reader),
+		marshaler: zson.NewZNGMarshaler(),
+		//strings:   opts.StringsOnly,
+	}
+}
+
+func (r *Reader) Read() (*zng.Record, error) {
+	for {
+		csvRec, err := r.reader.Read()
+		if err != nil {
+			if err == io.EOF {
+				if !r.valid {
+					err = errors.New("empty csv file")
+				} else {
+					err = nil
+				}
+			}
+			return nil, err
+		}
+		if r.hdr == nil {
+			r.init(csvRec)
+			continue
+		}
+		rec, err := r.translate(csvRec)
+		if err != nil {
+			return nil, err
+		}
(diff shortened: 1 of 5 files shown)


zio/csvio/reader.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+}
+
+func (r *Reader) Read() (*zng.Record, error) {
+	for {
+		csvRec, err := r.reader.Read()
+		if err != nil {
+			if err == io.EOF {
+				if !r.valid {
+					err = errors.New("empty csv file")
```
mattnibs, line 49: I assume this here to get auto-detect working? I would assume that expected behavior for a Reader on an empty file would be a `nil, nil` return on the first read...

zio/csvio/reader.go
```diff
@@ -0,0 +1,109 @@
...
+	switch strings.ToLower(s) {
+	case "+inf", "inf":
+		return math.MaxFloat64
+	case "-inf":
+		return -math.MaxFloat64
+	case "nan":
+		return math.NaN()
+	case "":
+		return nil
+	}
```
nwt, line 101: `strconv.ParseFloat` understands these strings plus `infinity` with optional sign, though it returns `math.Inf(1)` and `math.Inf(-1)` instead of `math.MaxFloat64` and `-math.MaxFloat64`. If that's OK, we could simplify this down to

```go
    if s == "" {
        return nil
    }
```

Otherwise, we should add cases for `"infinity"`, `"+infinity"`, and `"-infinity"` for consistent behavior.

We should also add test coverage for the stuff in this function, but that doesn't have to happen in this PR.

zio/csvio/reader.go
```diff
@@ -0,0 +1,109 @@
...
+// Writer depends on fuse.  We should refactor this so whatever logic wants
+// to tack on a fuse operator happens outside of zio.  See issue #2315
+//type ReaderOpts struct {
+//	StringsOnly bool
+//}
+
+func NewReader(reader io.Reader, zctx *zson.Context) *Reader {
+	return &Reader{
+		zctx:      zctx,
+		reader:    csv.NewReader(reader),
```
nwt, line 37: Avoid a slice allocation on every trip through `csv.Reader.Read`.
```suggestion
    csvReader := csv.NewReader(reader)
    csvReader.ReuseRecord = true
    return &Reader{
        zctx:      zctx,
        reader:    csvReader,
```

zio/csvio/reader.go
mccanne: Yes, that was my first crack at getting the auto-detector to work then I realized it was more involved because CSV is so forgiving to short, random inputs.  I could take this out and it could be part of the "strict mode" that is contemplated in #2319 2316.

zio/csvio/reader.go
mccanne: cool

zio/csvio/reader.go
mccanne: nice

zio/csvio/reader.go
mccanne: I'm inclined to leave it for now and address this in #2316

mattnibs (approved):

zio/csvio/reader.go
nwt: #2320 tracks test coverage for this function.

nwt (approved):

PR #2329, opened 2021-03-12
update to actions/checkout@v2 in CI

nwt:


.github/workflows/ci.yaml
```diff
@@ -13,7 +13,9 @@ jobs:
       - uses: actions/setup-go@v2
         with:
           go-version: '1.16'
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
mattnibs, line 18: What is this all about?

nwt: It fetches tags, which we need for `git describe` (in `Makefile`).

mattnibs (approved):

PR #2334, opened 2021-03-13
fix time and duration casts

### LINKED ISSUE #962: types.json can turn ISO8601 into "time" type, but casting can't
Repro is with `zq` commit `8cdcf4b` and the [streaming JSON zq-sample-data](https://github.com/brimsec/zq-sample-data/tree/master/zeek-ndjson).

If I use the `types.json` config, a field that started life as a JSON string that happens to hold an ISO8601 timestamp can be turned into a ZNG `time` type.

```
$ zq -j ~/work/zq/zeek/types.json -t 'cut _write_ts | put foo=_write_ts | head 1' weird.ndjson.gz 
#0:record[_write_ts:time,foo:time]
0:[1521911720.600843;1521911720.600843;]
```

However, if I try to use the casting syntax "a la carte" on this one field instead of using `types.json`, the cast fails.

```
$ zq -t 'cut _write_ts | put foo=_write_ts:time | head 1' weird.ndjson.gz 
bad cast
#0:record[_write_ts:string]
0:[2018-03-24T17:15:20.600843Z;]
```

mccanne:
This commit fixes the time and duration casting logic to be more
flexible and reasonable. Strings in RFC-3339 can now be cast to
times or durations.  Numerics (float and int) are now presumed to
both be seconds and float64 rounding errors are allowed through
as there are better ways to avoid such rounding errors by using
ISO date/times or native ZNG times from a client.

We also changed the casting logic and iso() function to return
embedded Z errors rather than fatal Go errors to allow computation
to continue in the face of errors.

Given this new behavior and the upcoming additional of ZSON-oriented
Z literals, there is no longer a need for functions sec(), msec(),
etc so they have been removed.

Fixes #962


expr/cast.go
```diff
@@ -103,7 +106,23 @@ func castToIP(zv zng.Value) (zng.Value, error) {
...
-	if zng.IsFloat(zv.Type.ID()) {
+	id := zv.Type.ID()
+	if zng.IsStringy(id) {
+		s := string(zv.Bytes)
+		d, err := time.ParseDuration(s)
+		var ns int64
+		if err == nil {
+			ns = int64(d)
+		} else {
+			f, ferr := strconv.ParseFloat(s, 64)
```
nwt, line 117:
```suggestion
			f, ferr := byteconv.ParseFloat(zv.Bytes, 64)
```

expr/cast.go
```diff
@@ -103,7 +106,23 @@ func castToIP(zv zng.Value) (zng.Value, error) {
 }
 
 func castToDuration(zv zng.Value) (zng.Value, error) {
-	if zng.IsFloat(zv.Type.ID()) {
+	id := zv.Type.ID()
+	if zng.IsStringy(id) {
+		s := string(zv.Bytes)
+		d, err := time.ParseDuration(s)
```
nwt, line 112:
```suggestion
		d, err := time.ParseDuration(byteconv.UnsafeString(zv.Bytes))
```

expr/expr_test.go
```diff
@@ -649,7 +649,7 @@ func TestCasts(t *testing.T) {
 	// Test casts to time
 	ts := zng.Value{zng.TypeTime, zng.EncodeTime(nano.Ts(1589126400_000_000_000))}
 	testSuccessful(t, "1589126400.0 :time", "", ts)
-	ts = zng.Value{zng.TypeTime, zng.EncodeTime(nano.Ts(1589126400))}
+	ts = zng.Value{zng.TypeTime, zng.EncodeTime(nano.Ts(1589126400_000_000_000))}
```
nwt, line 652: This now repeats line 650.
```suggestion
```

expr/function/time.go
```diff
@@ -14,90 +16,49 @@ type iso struct {
...
-	if zv.Bytes == nil {
-		return zng.Value{zng.TypeInt64, nil}, nil
-	}
-	ns, ok := coerce.ToInt(zv)
-	if !ok {
-		ms, ok := coerce.ToFloat(zv)
-		if !ok {
-			return badarg("msec")
+		s := string(zv.Bytes)
+		ts, err := time.Parse(format, s)
```
nwt, line 40:
```suggestion
		ts, err := time.Parse(format, byteconv.UnsafeString(zv.Bytes))
```

expr/function/time.go
```diff
@@ -14,90 +16,49 @@ type iso struct {
...
+		if err != nil {
+			sec, ferr := strconv.ParseFloat(s, 64)
```
nwt, line 42:
```suggestion
			sec, ferr := byteconv.ParseFloat(zv.Bytes, 64)
```

expr/function/time.go
```diff
@@ -14,90 +16,49 @@ type iso struct {
...
-	result.Buffer
+	return zng.Value{zng.TypeTime, i.Time(ts)}, nil
 }
 
-func (s *sec) Call(args []zng.Value) (zng.Value, error) {
-	zv := args[0]
+func CastToTime(zv zng.Value) (nano.Ts, error) {
 	if zv.Bytes == nil {
-		return zng.Value{zng.TypeInt64, nil}, nil
+		// Any nil value is cast to a zero time...
```
nwt, line 28:
```suggestion
		// Any nil value is cast to a zero time.
```

expr/function/time.go
```diff
@@ -14,90 +16,49 @@ type iso struct {
...
+			return 0, fmt.Errorf("cannot convert value of type %s to time", zv.Type)
 		}
-		ns = int64(1000. * us)
-	} else {
-		ns *= 1000
+		ns = int64(sec * 1e9)
 	}
-	return zng.Value{zng.TypeInt64, u.Int(ns)}, nil
+	return nano.Ts(ns), nil
 }
```
nwt, line 62: 1. Should `coerce.ToInt` and `coerce.ToFloat` really swallow errors from `zng.Decode*`?
1. I expected this stretch to go like this.
```go
	if zng.IsInteger(id) {
		if sec, ok := coerce.ToInt(zv); ok {
			return nano.Ts(sec * 1_000_000_000), nil
		}
	}
	if zng.IsFloat(id) {
		if sec, ok := coerce.ToFloat(zv); ok {
			return nano.Ts(sec * 1e9), nil
		}
	}
	return 0, fmt.Errorf("cannot convert value of type %s to time", zv.Type)
}
```

expr/cast.go
mccanne: Given that the logic here already creates the string for ParseDuration, I don't think this helps much, but I'll go ahead and change it.

henridf:
> ZSON-oriented Z literals

🎉

nwt (approved):

PR #2337, opened 2021-03-14
TZNG->ZSON transition in zql/ztests

philrz:


nwt (approved):
When you merge, please edit the commit message to make it a little more specific than just "TZNG->ZSON transition". The PR title would be good for example.

PR #2340, opened 2021-03-14
TZNG->ZSON transition ztests below zio/

philrz:
There's a couple tests in here that still have TZNG because they exit precisely to test I/O of that format. If we should ever reach the point where we're genuinely phasing out TZNG altogether, we could dump 'em then.


zio/detector/ztests/tzng-nonewline.yaml
```diff
@@ -0,0 +1,9 @@
+zql: '*'
+
+input: |-
```
philrz, line 3: This was actually a material change I made to a test while I was in here. The way things were when I started this exercise, `zng.yaml` and `zng-nonewline.yaml` were byte-for-byte exactly the same, so the "nonewline" variant was not testing what was intended. I added the `-` here, following the pattern of the other "nonewline" tests.

nwt (approved):

PR #2355, opened 2021-03-15
extract tzng-related out of packange zng and into tzngio

### LINKED ISSUE #1569: simplify the type formatting methods in package zng
With the simplifications in epic #1566 and dropping of support for zeek tsv as an output format, we should revisit the changes made in issue #1417 and try to further simplify the needed methods in package zng for string-formatting of zng values.  With the unification the value format for csv/table/text/tzng , we should no longer need the zng.OutputFormat* constants.

Also, we still have String and StringOf which seem confusing to me.  I previously suggested changing StringOf to Format but maybe we should bikeshed this a bit more?

mccanne:
This commit does a surgical extraction of the TZNG-related
logic from the package zng and prepares us for the unbundling
of the zeek format.  Unfortunately, we're leaving here a bit of
a mess because comprehensive Z literals have not yet been
implemented; as soon as this goes in, the mess can be removed.
I've tried to leave comments in pointing to the messy bits
that will soon going away so please don't spend much time
reviewing this on the PR.

These changes also nicely highlights the vestigial dependencies
on tzngio so we can now easily grep the source tree to see which
pieces still depend on tzngio and thus still need to be ported
the rest of the way to Z.

Finally, these changes are mostly all code movement, except we
changed the format of durations in NDJSON to conform to RFC 3339.

Closes #1569


ztests/suite/errors/null.yaml
```diff
@@ -10,4 +10,4 @@ inputs:
 outputs:
   - name: stderr
     regexp: |
-      cannot instantiate type null with nonempty value "bleah"
+      .*invalid text body of null value: "bleah"
```
nwt, line 13:
```suggestion
      invalid text body of null value: "bleah"
```

expr/boolean.go
```diff
@@ -390,11 +391,46 @@ func Comparison(op string, literal ast.Literal) (Boolean, error) {
...
 	case int64:
 		return CompareInt64(op, v)
 	default:
 		return nil, fmt.Errorf("unknown type of constant: %s (%T)", literal.Type, v)
 	}
 }
+
+//XXX this shoulw all go away ?
+
+type Port uint32
```
nwt, line 405: This is unused.
```suggestion
```

expr/boolean.go
```diff
@@ -390,11 +391,46 @@ func Comparison(op string, literal ast.Literal) (Boolean, error) {
...
+type Bstring []byte
```
nwt, line 406: No need to export this, but if it's really going away, just leave it.
```suggestion
type bstring []byte
```

cmd/microindex/seek/command.go
```diff
@@ -115,7 +115,7 @@ func (c *Command) Run(args []string) error {
 			}
 			builder = zng.NewBuilder(typ)
 		} else if keyType != k.Type {
-			return fmt.Errorf("key type changed from %s to %s", keyType, k.Type)
+			return fmt.Errorf("key type changed from '%s' to '%s'", keyType.ZSON(), k.Type.ZSON())
```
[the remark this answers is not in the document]
mccanne: I'm still a Go newbie!

zng/split.go
```diff
@@ -0,0 +1,22 @@
...
+	vals := []Value{}
+	for it := b.Iter(); !it.Done(); {
+		zv, _, err := it.Next()
+		if err != nil {
+			return nil, fmt.Errorf("parsing element type '%s' value %q: %w", elemType.ZSON(), zv, err)
+		}
+		vals = append(vals, Value{elemType, zv})
+	}
+	return vals, nil
+}
```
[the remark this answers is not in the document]
mccanne: If you do this, then a zng empty array is returned instead of a nil array and if you change vals as you suggest, then nil array gets returned for empty array.   I got bit by this in the past...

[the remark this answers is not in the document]
mccanne: I'll add a comment.

zng/split.go
```diff
@@ -0,0 +1,25 @@
...
+	// slice of size 0 so that we can differentiate between a non-nil
+	// but empty Z container vs a nil Z container value.
+	if b == nil {
+		return nil, nil
+	}
+	vals := []Value{}
+	for it := b.Iter(); !it.Done(); {
+		zv, _, err := it.Next()
+		if err != nil {
+			return nil, fmt.Errorf("parsing element type '%s' value %q: %w", elemType.ZSON(), zv, err)
```
nwt, line 20: `zv` will always be nil here.

expr/cast.go
```diff
@@ -144,8 +145,12 @@ func castToTime(zv zng.Value) (zng.Value, error) {
 
 func castToStringy(typ zng.Type) func(zng.Value) (zng.Value, error) {
 	return func(zv zng.Value) (zng.Value, error) {
-		if zv.Type.ID() == zng.IdBytes {
-			return zng.Value{typ, zng.EncodeString(string(zv.Bytes))}, nil
+		id := zv.Type.ID()
+		if id == zng.IdBytes || id == zng.IdBstring {
+			if !utf8.Valid(zv.Bytes) {
+				return zng.NewErrorf("non-utf8 bytes cannot be cast to string"), nil
```
nwt, line 151:
```suggestion
				return zng.NewErrorf("non-UTF-8 bytes cannot be cast to string"), nil
```

zng/split.go
mccanne: Yeah that should be `b` instead, right?

nwt: I think so.

expr/boolean.go
```diff
@@ -201,7 +202,7 @@ var compareString = map[string]func(string, string) bool{
 	"<=": func(a, b string) bool { return a <= b },
 }
 
-func CompareBstring(op string, pattern zng.Bstring) (Boolean, error) {
+func CompareBstring(op string, pattern bstring) (Boolean, error) {
```
nwt, line 205: I didn't notice `Bstring` was part of a public interface when I suggested changing it to `bstring`. :facepalm:

I'm happy to let that ride until we rework `parseLiteral`, though.

nwt (approved):

PR #2356, opened 2021-03-16
create uniform naming conventions for AST

### LINKED ISSUE #2165: change Reducer to Aggregator in Z grammar
We should clean up the Reducer ast nodes by change the op to "aggregator" and changing the PEG rule names for Reducer* to Aggregator-oriented names.

### LINKED ISSUE #2222: consistent ast Op names
Shorten and make ast Op names consistent throughout. Update zql.peg and app.  
Every Op name should be the same name as the Go struct type.

Some additional thoughts copied from duplicate issue #[2134:](https://github.com/brimsec/zq/issues/2134)

We should unify the names of AST node and Ops.

The Op names in each of the AST nodes should have a string name that is the same as the ast type.

For example, ast.Foo should have an Op with value "Foo".

In terms of all the Expr nodes we should change Expression to Expr and be consistent throughout.

mccanne:
The various names in the AST data structures evolved over
time and became inconsistent.  This commit cleans it all
up in the favor of short and consistent names.  We also
changed the "reducer" concept to "aggregator" throughout
and replaced ast.TypeExpr with ast.TypeValue as they were
logically equivalent.

Closes #2222
Closes #2165

### DIFF
--- a/compiler/ast/ast.go
+++ b/compiler/ast/ast.go
@@ -22,24 +22,24 @@ type Proc interface {
 
 // Id refers to a syntax element analogous to a programming language identifier.
 type Id struct {
-	Op   string `json:"op" unpack:""`
+	Kind string `json:"kind" unpack:""`
 	Name string `json:"name"`
 }
 
 type Path struct {
-	Op   string   `json:"op" unpack:""`
+	Kind string   `json:"kind" unpack:""`
 	Name []string `json:"name"`
 }
 
 type Ref struct {
-	Op   string `json:"op" unpack:""`
+	Kind string `json:"kind" unpack:""`
 	Name string `json:"name"`
 }
 
 // Root refers to the outer record being operated upon.  Field accesses
 // typically begin with the LHS of a "." expression set to a Root.
 type Root struct {
-	Op string `json:"op" unpack:""`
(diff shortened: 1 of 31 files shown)


mccanne:
Note that the app will break with this change.  I'll fix it on a branch and pass tests before merging here.

compiler/ast/ast.go
```diff
@@ -313,32 +302,32 @@ type (
...
 type SeqExpr struct {
-	Op        string       `json:"op" unpack:""`
-	Name      string       `json:"name"`
-	Selectors []Expression `json:"selectors"`
-	Methods   []Method     `json:"methods"`
+	Op        string   `json:"op" unpack:""`
+	Name      string   `json:"name"`
+	Selectors []Expr   `json:"selectors"`
+	Methods   []Method `json:"methods"`
 }
```
mattnibs, line 315: This should go up with Expr

mattnibs (approved):

nwt (approved):
I like the idea we talked about of swapping the roles of the `Op` and `Kind` fields.

mccanne:
Tests passing in app branch... https://github.com/brimsec/brim/compare/zq-uniform-ast

Issue #2293, opened 2021-03-04
Panic during "zar find" on named field index


philrz:
While the fix in #2273 appeared effective for the example in the original issue, it seems to have introduced a new problem, as this now causes a panic. Using repro steps similar to the `zar` README and testing with `zar` commit `77899a1` that's associated with the change in #2273:

```
$ zar -version
Version: v0.29.0-98-g77899a1f

$ rm -rf logs && mkdir ./logs && ZAR_ROOT=`pwd`/logs && zar import -s 25MB zng/*.gz && zar index create uri && zar find -f table uri=/file
added: file:///Users/phil/logs/zd/20180324/d-1pJ4Vycnq7cg5f5ZZBbr0JJZX6V.zng.zar/idx-1pJ4W3jpHPzoq35O0RxSd9Vg0ou.zng
added: file:///Users/phil/logs/zd/20180324/d-1pJ4VUdChU7I4ziQGZb9pgbARb4.zng.zar/idx-1pJ4W3jpHPzoq35O0RxSd9Vg0ou.zng
added: file:///Users/phil/logs/zd/20180324/d-1pJ4VfjUDKxjLlQdi7VTbW4pNyB.zng.zar/idx-1pJ4W3jpHPzoq35O0RxSd9Vg0ou.zng
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x168908c]

goroutine 50 [running]:
github.com/brimsec/zq/zng.(*Record).ZvalIter(...)
	/Users/phil/scratch/zq/zng/recordval.go:117
github.com/brimsec/zq/zng.(*Record).FieldIter(...)
	/Users/phil/scratch/zq/zng/recordval.go:124
github.com/brimsec/zq/expr.NewKeyCompareFn(0x0, 0xc0005581a0, 0x1, 0x1)
	/Users/phil/scratch/zq/expr/sort.go:100 +0x4c
github.com/brimsec/zq/microindex.NewFinderReader(0x1e20980, 0xc00055a680, 0xc00054a090, 0xc0005700a0, 0x4, 0x0, 0x0, 0x0, 0xc0005700a7, 0x0, ...)
	/Users/phil/scratch/zq/microindex/findreader.go:31 +0xf4
[...]

PR #2361, opened 2021-03-17
Fix flaky test for #2293

### LINKED ISSUE #2360: intermittent failure of ztests/fixedbugs/2241.yaml
This test fails intermittently...
```
ztest.go:475: /home/runner/work/zq/zq/ztests/fixedbugs/2241.yaml: expected and actual stdout differ:
        --- expected
        +++ actual
        @@ -1 +1,2 @@
        +added: file:///tmp/2241719424260/zartest/zd/20180324/d-1ps7DSsgnER2VwIUuWDHvPWRIOU.zng.zar/idx-1ps7DNhgHIEZfWSR83DxdJ4wEJj.zng
```
It looks like there is output status racing against the zar command exiting.  Seems like zar should wait for all such output?

mattnibs:
The ztest for fixedbugs/2293 should have used the "-q" flag since the test does
not care about the output of "zar index create".

Never the less, the flaky test exposed an race conditition with the progress
update functionality, where the program would exit before the update logic
would have a chance to print all updates. Fix this as well.

Also the test was named after the wrong issue, update the name to reflect the
right issue.

Closes #2360


mccanne (approved):
Yay!

ppl/cmd/zar/index/create.go
```diff
@@ -43,29 +44,32 @@ requires specifying the key and output file name. For example:
 
 type CreateCommand struct {
 	*root.Command
+	displayer  *displayer
```
nwt, line 47: Make this `displayer displayer` so you don't have to initialize it?

ppl/cmd/zar/index/create.go
```diff
@@ -43,29 +44,32 @@ requires specifying the key and output file name. For example:
...
 	framesize  int
 	inputFile  string
 	keys       string
 	outputFile string
 	ensure     bool
 	noapply    bool
 	procFlags  procflags.Flags
-	progress   chan string
 	quiet      bool
```
nwt, line 55: Is `quiet` now unused?

ppl/cmd/zar/index/create.go
```diff
@@ -110,14 +114,13 @@ func (c *CreateCommand) Run(args []string) error {
...
-	}
+	c.displayer.run()
+	defer c.displayer.close()
 
-	if err := lake.WriteIndices(ctx, lk, c.progress, defs...); err != nil {
+	if err := lake.WriteIndices(ctx, lk, c.displayer.ch, defs...); err != nil {
 		return err
 	}
+
 	return nil
```
nwt, line 124: Shorten to `return lake.WriteIndices(ctx, lk, c.displayer.ch, defs...)`?

ppl/cmd/zar/index/create.go
mattnibs: yup

ppl/cmd/zar/index/create.go
mattnibs: correct will remove

PR #2363, opened 2021-03-17
zbuf.NewCompareFn: Respect nullsMax on reverse

### LINKED ISSUE #2259: Incorrect results with groupby sorted input and large data set
tl;dr - While the fix in #2220 verified nicely for the smaller data set that made #1912 easy to repro, repeated Rug tests with the full 70 GB wrccdc data set indicate a new problem has been introduced with the fix in #2220.

The issue shows up with several of the Rug `count_every` tests. To have a place to start, I've been focused on the `count_every_1w` test with the full 70 GB wrccdc data set. The Z query is `every 1w count() | head 50`. With the fix to #2220 in place that puts the `null` values last in the sort order, the correct reference output in NDJSON format would be:

```
{
  "count": 3131,
  "ts": "2018-11-08T00:00:00Z"
}
{
  "count": 475593430,
  "ts": "2018-03-22T00:00:00Z"
}
{
  "count": 96242990,
  "ts": "2017-03-23T00:00:00Z"
}
{
  "count": 26,
  "ts": null
}
```

After seeing output diverge from this in recent commits, I created this [cycle.sh](https://github.com/brimsec/zq/files/6062609/cycle.sh.gz) script that helped me confirm that the symptom appears to show up at _query_ _time_.

```
$ gzcat cycle.sh.gz 
#!/bin/bash
SPACE=wrccdc
NUM=1
while true
do
  echo "Run #$NUM:"
  zq -version; zapi new -k archivestore $SPACE
  zapi -s $SPACE postpath /Users/phil/Downloads/sampledata/$SPACE/zeek-logs/[!l]*.log*
  while true
  do
    echo "================================="
    date
    zapi -s $SPACE get -f ndjson 'every 1w count() | head 50' | jq .
  done
done
```

That script ingests the data only once, then in repeated queries, we sometimes see the expected result, but often it's wrong. The following samples are from recent `zq` commit `ab604bb`. Sometimes the `null` records seem to "split" one of the buckets:

```
Sun Feb 28 13:55:18 PST 2021
{
  "count": 3131,
  "ts": "2018-11-08T00:00:00Z"
}
{
  "count": 280917502,
  "ts": "2018-03-22T00:00:00Z"
}
{
  "count": 26,
  "ts": null
}
{
  "count": 194675928,
  "ts": "2018-03-22T00:00:00Z"
}
{
  "count": 96242990,
  "ts": "2017-03-23T00:00:00Z"
}
```

Sometimes there's duplicate buckets with the `null` records on the other side.

```
Sun Feb 28 14:07:42 PST 2021
{
  "count": 3131,
  "ts": "2018-11-08T00:00:00Z"
}
{
  "count": 475593430,
  "ts": "2018-03-22T00:00:00Z"
}
{
  "count": 80397127,
  "ts": "2017-03-23T00:00:00Z"
}
{
  "count": 15845863,
  "ts": "2017-03-23T00:00:00Z"
}
{
  "count": 26,
  "ts": null
}
```

Using the script, I did a bunch of repeated runs that effectively show the problem started to show up at commit `563142a`, which is the commit associated with #2220. Here's the exhaustive set of results at each commit along the path, from oldest to newest:

|**Commit**|**cycle.sh output**|**zqd log during run**|Summary|
|-|-|-|-|
|82630a5|[cycle-82630a5.log.gz](https://github.com/brimsec/zq/files/6062677/cycle-82630a5.log.gz)|[zqd-cycle-82630a5.log.gz](https://github.com/brimsec/zq/files/6062678/zqd-cycle-82630a5.log.gz)|Commit right before #2220 fix. Shows same symptom as #1912: Buckets have correct counts, but `null` bucket skips around.|
|563142a|[cycle-563142a.log.gz](https://github.com/brimsec/zq/files/6062680/cycle-563142a.log.gz)|[zqd-cycle-563142a.log.gz](https://github.com/brimsec/zq/files/6062682/zqd-cycle-563142a.log.gz)|Commit for #2220. Starts to show the symptom shown above with incorrect/duplicate buckets.|
|ab604bb|[cycle-ab604bb.log.gz](https://github.com/brimsec/zq/files/6062687/cycle-ab604bb.log.gz)|[zqd-cycle-ab604bb.log.gz](https://github.com/brimsec/zq/files/6062688/zqd-cycle-ab604bb.log.gz)|A very recent `zq` commit showing the symptom of incorrect/duplicate buckets still occurring.|

One of the theories I pursued while doing the runs was that maybe the bad results only showed up while compaction was running, but it looks like this is not the case. For instance, in `zqd-cycle-ab604bb.log.gz`, the timestamp for the `Compact completed` event is `1614550684`, which is Sunday, February 28, 2021 2:18:04 PM Pacific. Meanwhile, the datestamps of the repeated queries in `cycle-ab604bb.log.gz` show the bad results continue to be logged out beyond 2:30 PM Pacific after compaction has finished.

mattnibs:
This came about because the merge proc on a parallel count by every on
descending sorted data was not handling null values in the same manner as the
child group by procs (i.e. For descending ordered data nil values should be the
last elements in a sorted list). Fix zbuf.NewCompareFn so that it sets nullsMax
to false if the input is in descending order.

Closes #2259


compiler/semantic/parallelize.go
```diff
@@ -227,11 +227,12 @@ func parallelize(p ast.Proc, N int, inputSortField field.Static, inputSortRevers
 
 	seq := ensureSequentialProc(p)
 	orderSensitiveTail := true
+loop:
 	for i := range seq.Procs {
 		switch seq.Procs[i].(type) {
 		case *ast.SortProc, *ast.GroupByProc:
 			orderSensitiveTail = false
-			break
+			break loop
```
mattnibs, line 235: Unrelated to this bug, but thought it should be fixed.

nwt: It should! But it's unrelated to the other change in this PR, so a separate PR would be better.

nwt (approved):

PR #2371, opened 2021-03-18
finalize zson sytnax

### LINKED ISSUE #2357: proper support for RFC-3339 durations
Golang's time.Duration does not support days, weeks, months etc.  We need to either implement our own parser or find a library that support RFC-3339.

mccanne:
This commit changes the definition of a zson duration to be more
comprehensive.  We looked at RFC-3339 but that syntax is unergonomic
and does not cover sub-second granularity.  We decided to follow
the lead of Prometheus duration syntax but extend it with microseconds
and nanoseconds.

To implement this change, we created a nano.Duration type in
package nano.  Further, we changed nano.Span to use Duration instead
of int64.  This means json marshaled nano.Spans (e.g., in the zqd
service API) now use zson duration strings.

Closes #2357


pkg/nano/duration.go
```diff
@@ -0,0 +1,173 @@
...
+var scale = map[string]Duration{
+	"ns": Nanosecond,
+	"us": Microsecond,
+	"ms": Millisecond,
+	"s":  Second,
+	"m":  Minute,
+	"h":  Hour,
+	"d":  Day,
+	"w":  Week,
+	"y":  Year,
```
mattnibs, line 123: Can a year duration be specified given its variable length? I noticed you don't have a test for it...

mccanne: This is a limitation of the syntax and it's worse for months so we don't have a month unit at all.  Unless you know the years over which the duration falls (and even the time zone!), you can't interpret the "y" unit as the precise number of years.  That said, the encoding is precise in that "y" is an abstract unit exactly equal to 365 days... it just doesn't reflect date differences precisely.  For this, we'd need a moments type library accessible via function calls and time difference calculations.  BTW, this is how Prometheus works so we are in good company.

mccanne: That said I can of course add a boomerang for "y".

mccanne: I will clarify the ZSON spec on this point.

mattnibs (approved):

zng/duration.go
```diff
@@ -1,27 +1,30 @@
...
 }
 
-func DecodeDuration(zv zcode.Bytes) (int64, error) {
-	return DecodeInt(zv)
+func DecodeDuration(zv zcode.Bytes) (nano.Duration, error) {
+	i, err := DecodeInt(zv)
+	if err != nil {
+		return 0, err
+	}
+	return nano.Duration(i), nil
```
nwt, line 27:
```suggestion
	return nano.Duration(i), err
```

nwt (approved):

pkg/nano/duration_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"100us",
+	"123ns",
+	"123ms",
+	"1.3s",
+	"0s",
+	"31y259d1h46m40s",
+	"2h43m9.993714061s",
+}
+
+func TestDuration(t *testing.T) {
```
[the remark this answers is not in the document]
mccanne: I'll have a look.  They missed `-292y171d23h47m16.854775808s` :-)

pkg/nano/duration_test.go
```diff
@@ -0,0 +1,69 @@
...
+	checkdur(t, "0ns", "0s")
+	checkdur(t, "1s300ms", "1.3s")
+	checkdur(t, "2716us", "2.716ms")
+	checkdur(t, "1230us", "1.23ms")
+	checkdur(t, "11230us", "11.23ms")
+	checkdur(t, "111230us", "111.23ms")
+	checkdur(t, "1234ns", "1.234us")
+	checkdur(t, "", "0s")
+	checkdur(t, "1w", "7d")
+	checkdur(t, "1.5w", "10d12h")
```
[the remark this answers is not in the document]
mccanne: I'll have a look.  They missed `-292y171d23h47m16.854775808s` :-)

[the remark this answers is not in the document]
mccanne: done

pkg/nano/duration.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+func DurationFromParts(sec, ns int64) Duration {
+	return Duration(sec*1_000_000_000 + ns)
+}
+
+func FloatToDuration(fsec float64) Duration {
+	rsec := math.Round(fsec)
+	ns := fsec - rsec
+	return DurationFromParts(int64(rsec), int64(ns*1e9))
```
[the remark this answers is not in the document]
mccanne: I think this evolved from a pattern where converting a decimal string overflowed when converting to float then converting to Duration, but here, the input is already a float64.  I will change as you suggest.

Issue #2057, opened 2021-02-03
revisit zson scanner and add more comprehensive tests


mccanne:
Since changing the ZSON parser to stream its input, we've run into a couple bugs.  This change was hurriedly put in so we should go back and study the code more completely and write more extensive tests that stress the streaming nature of the zson scanner.

philrz:
Repro is with `zq` commit 79b4821.

As part of the TZNG->ZSON test conversions, I'm working with this test:

[https://github.com/brimsec/zq/blob/master/ppl/zqd/handlers_test.go#L578-L591](https://github.com/brimsec/zq/blob/master/ppl/zqd/handlers_test.go#L578-L591)

Run manually in its original TZNG form, it gives an error message as expected:

```
$ cat in.tzng 
#0:record[_path:string,ts:time,uid:bstring
0:[conn;1;CBrzd94qfowOqJwCHa;]

$ zq -z in.tzng 
in.tzng: format detection error
	tzng: line 1: syntax error parsing type string
...
```

But if I try to construct similar malformed ZSON variations by dropping bits of input, it silently eats them without error.

```
$ cat in.zson 
{_path:"conn",ts:1970-01-01T00:00:01Z,uid:"CBrzd94qfowOqJwCHa" (bstring)} (=0)

$ zq -z in.zson 
{_path:"conn",ts:1970-01-01T00:00:01Z,uid:"CBrzd94qfowOqJwCHa" (bstring)} (=0)

$ cat in1.zson 
{_path:"conn",ts:1970-01-01T00:00:01Z,uid:"CBrzd94qfowOqJwCHa" (bstring)} (=0

$ zq -z in1.zson 
[no output]

$ echo $?
0

$ cat in2.zson 
{_path:"conn",ts:1970-01-01T00:00:01Z,uid:"CBrzd94qfowOqJwCHa" (bstring) (=0)

$ zq -z in2.zson 
[no output]

$ echo $?
0
```

philrz:
Here's another one from this test:

[https://github.com/brimsec/zq/blob/79b4821bf60de2c3671fdfd46e15b83898cbaa35/ppl/zqd/handlers_test.go#L456-L461](https://github.com/brimsec/zq/blob/79b4821bf60de2c3671fdfd46e15b83898cbaa35/ppl/zqd/handlers_test.go#L456-L461)

Manual run in its original TZNG form:

```
$ cat in.tzng 
#0:record[_path:string,ts:time,uid:bstring]
[...]

PR #2377, opened 2021-03-19
TZNG->ZSON transition for 'babble' references

philrz:
In approaching my next round of TZNG->ZSON transitions, I came across my first reference to these "babble" files, so I've decided to resolve these all at once since they're spread out across many directories. Then I'll get back to doing conversions in one directory at a time.

In addition to the many individual ztest files I hit, the chain of dependencies reached into some of the  `.go` tests as well, so with some help from @nwt these got converted in their entirety as well:

*   ppl/lake/index/index\_test.go
*   ppl/lake/lake\_test.go
*   ppl/zqd/handlers\_pcap\_test.go
*   ppl/zqd/handlers\_test.go

As I was doing conversions, I made one material change to fix a test and have a couple other observations, so I've added comments in-line to call attention to those.


microindex/ztests/manual.yaml
```diff
@@ -1,14 +1,14 @@
...
-  microindex create -o index -k s babble.tzng
-  zq -f tzng index.zng > index.tzng
-  zq -o sorted.zng "count() by s | put key=s | cut key | sort key" babble.tzng
-  microindex convert -o manual -k key sorted.zng
-  zq -f tzng manual.zng > manual.tzng
-  diff index.tzng manual.tzng
+  microindex create -o index.zng -k s babble.zson
+  zq -z index.zng > index.zson
+  zq -o sorted.zng "count() by s | put key=s | cut key | sort key" babble.zson
+  microindex convert -o manual.zng -k key sorted.zng
```
philrz, line 5: Here's the material change. The lack of the `.zng` extensions on the `-o` options to `microindex create` was making for zero-length `index.tzng` and `sorted.zng` files. So those zero-length files `diff`'ed successfully and made it look like the test was doing something real and passing when in fact it was effectively a no-op.

I can't help but see this base `zq` behavior as buggy. Given how we want everything to be stream-friendly, I guess I could respect it as expected behavior to output a zero-length stream if a zero-length stream is given as input. OTOH, if an explicit filename is given as the source of that input stream and that file doesn't even exist, it seems like we should give some kind of error/warning or non-zero exit code. Thoughts?

```
$ ls -l doesnotexist
ls: doesnotexist: No such file or directory

$ zq -z doesnotexist > something

$ ls -l something
-rw-r--r--  1 phil  staff  0 Mar 19 09:03 something

$ echo $?
0
```

ppl/lake/ztests/map.yaml
```diff
@@ -1,32 +1,28 @@
...
-      #0:record[v:int64,count:uint64]
-      0:[1;1;]
-      0:[278;6;]
-      #0:record[v:int64,count:uint64]
-      0:[1;1;]
-      0:[14;5;]
+      {v:1,count:1 (uint64)} (=0)
+      {v:278,count:6} (0)
+      {v:1,count:1 (uint64)} (=0)
+      {v:14,count:5} (0)
```
philrz, line 28: I'm guessing this one is not a bug, but point it out just in case it's worthy of further discussion. In these conversions I generally take the approach of converting the TZNG to ZSON and assuming that'll do the trick. In this case it failed on the first time because a straight conversion of the TZNG that was in the YAML gives:

```
$ cat in.tzng 
#0:record[v:int64,count:uint64]
0:[1;1;]
0:[278;6;]
#0:record[v:int64,count:uint64]
0:[1;1;]
0:[14;5;]

$ zq -z in.tzng 
{v:1,count:1 (uint64)} (=0)
{v:278,count:6} (0)
{v:1,count:1} (0)
{v:14,count:5} (0)
```

Notice the TZNG `zar map` output was re-using the `#0` type definition, and the ZSON output of `zar map` shows a similar effect with the `=0`. I guess it's legal to re-use type definitions in this way, though.

ppl/lake/ztests/s3/s3root.yaml
```diff
@@ -23,9 +23,7 @@ outputs:
       [1587518620062237300-1587513611063914690]
       [1587513592062544400-1587508830068523240]
       ===
-      #0:record[count:uint64]
-      0:[1000;]
-      #0:record[key:int64,count:uint64,first:time,last:time]
-      0:[2;1;1587518620.0622373;1587513611.06391469;]
+      {count:1000 (uint64)} (=0)
+      {key:2,count:1 (uint64),first:2020-04-22T01:23:40.0622373Z,last:2020-04-22T00:00:11.06391469Z} (=0)
```
philrz, line 27: Similar observation to the `map.yaml` one above: The `=0` type definition is re-used.

ppl/zqd/handlers_test.go
```diff
@@ -453,6 +447,8 @@ func TestPostZngLogs(t *testing.T) {
 	}, info)
 }
 
+// Skipped trying to convert this one to ZSON for now.
+// See https://github.com/brimsec/zq/issues/2057#issuecomment-803187964
```
philrz, line 451: Just adding a comment to make sure this stands out in the sea of changes.

ppl/zqd/handlers_test.go
```diff
@@ -575,10 +572,12 @@ func TestPostNDJSONLogWarning(t *testing.T) {
 	assert.EqualValues(t, 134, res.BytesRead)
 }
 
+// Other attempts at malformed ZSON closer to the original are not yet flagged
+// as errors. See https://github.com/brimsec/zq/issues/2057#issuecomment-803148819
```
philrz, line 576: Another one I just wanted to make sure folks didn't miss.

ppl/zqd/handlers_test.go
```diff
@@ -412,17 +406,17 @@ func TestRequestID(t *testing.T) {
 	})
 }
 
-func TestPostZngLogs(t *testing.T) {
+func TestPostZsonLogs(t *testing.T) {
```
philrz, line 409: I guess technically this should have been `TestPostTzngLogs()` before, but it was probably a holdover from the days that ZNG was always the text-based format. In any case, the name seemed even more wrong now that it's all ZSON, so I went ahead and changed it.

microindex/ztests/manual.yaml
mccanne: Yes, this is a bug and appears to be a vestige about how the command parsing logic evolved.  What happens here is "doesnotexist" is compiled as a Z query and the rest of the code is perfectly happy with the fact that array of input files is length 0 so nothing happens.  We just need to a check for no files (stdin appears as "-" and doesn't cause this problem obviously).

mccanne: I could swear we used to have a check for this...

mccanne: Fixed in #2382

ppl/lake/ztests/map.yaml
mccanne: This is ok.

ppl/lake/ztests/s3/s3root.yaml
mccanne: Yes, this is ok too.

ppl/zqd/handlers_test.go
mccanne: I think leaving it is fine since this test is really about feeding the log-post logic text that doesn't work with the auto-detector so no need to change the buggy tzng to buggy zson.

mccanne: I added a comment about the funky zson enum error to #2057

ppl/zqd/handlers_test.go
mccanne: Same thing... you can leave this as is.

mccanne (approved):

Issue #2224, opened 2021-02-24
Bugs outputting canonical Z from Brim Query Library entries


philrz:
With the first round of canonical Z support delivered in https://github.com/brimsec/zq/pull/2217, I figured a good first thing to test drive it with would be all the entries in the current Brim default Query Library (Brim commit `38e8e79f`). Indeed this revealed a couple bugs. I'll start with both in this same issue. Repro is with `ast` commit `a22b2a16`.

The one for **HTTP Requests** gives the error `AST bug: dot operator` for the common Zeek dot-containing field references.

```
$ ast -C '_path=http | cut id.orig_h, id.resp_h, id.resp_p, method,host, uri | uniq -c'
filter _path="http"
| cut AST bug: dot operator,AST bug: dot operator,AST bug: dot operator,method,host,uri
| uniq -c
```

The one for **File Activity** drops the reference to `null`.

```
$ ast -C 'filename!=null | cut _path, tx_hosts, rx_hosts, conn_uids, mime_type, filename, md5, sha1'
filter filename!=
| cut _path,tx_hosts,rx_hosts,conn_uids,mime_type,filename,md5,sha1
```

Idea: Would it make sense to extend our automation (ztest?) such that any Z we use in our tests is also subjected to a "boomerang" that turns its AST back into the canonical Z, then turns that canonical Z back into the AST yet again, and confirms the original AST and the final one are exactly the same (they should be, right)? It seems like this would be a quick way to get a whole lot of test coverage.

philrz:
@mccanne mentioned his branch `semantic` (commit 2ade2ec at the moment) where I see this one is already fixed:

```
$ ast -C '_path=http | cut id.orig_h, id.resp_h, id.resp_p, method,host, uri | uniq -c'
filter _path="http"
| cut id.orig_h,id.resp_h,id.resp_p,method,host,uri
| uniq -c
```

This one is still lacking the `null` though:

```
$ ast -C 'filename!=null | cut _path, tx_hosts, rx_hosts, conn_uids, mime_type, filename, md5, sha1'
filter filename!=
| cut _path,tx_hosts,rx_hosts,conn_uids,mime_type,filename,md5,sha1
```

PR #2378, opened 2021-03-19
fix dot and add type values in zfmt

mccanne:
This commit fixes a bug in canonical zfmt to print "."  instead of
nothing for a root record field path.  It also adds support for
printing type values.

This addresses some but not all of the problems in #2224. 
We can add a number of tests for this stuff when we tackle
the rest of #2224.


nwt (approved):
Maybe add test coverage now if it's easy?

mccanne:
> Maybe add test coverage now if it's easy?

I'd like to wait simply because this is still an early first stab meant to provoke conversation and this is all a moving target.  As we finalize canonical form we can write tests that cover the final design.

PR #2380, opened 2021-03-19
add ast.RegExp

mccanne:
This commit removes the regexp concept from the ast.Primitive path
since it is not a Z literal.  To do this, we added a RegExp operator
to the ast and runtime which simply is a boolean regexp pattern
matcher that operates on stringy types.

Now that Z primitives are cleaned up, we have a slate is clean
to add complex Z literals in a subsequent PR.


expr/boolean.go
```diff
@@ -228,33 +228,14 @@ func CheckRegexp(pattern string) (*regexp.Regexp, error) {
...
-			case zng.IdString, zng.IdBstring:
-				return !re.Match(v.Bytes)
-			}
-			return false
-		}, nil
-	default:
-		return nil, fmt.Errorf("unknown pattern comparator: %s", op)
+// NewRegexp returns a Booelan that compares values that must
+// be a stringy the given regexp.
+func NewRegexpBool(re *regexp.Regexp) Boolean {
```
nwt, line 233:
```suggestion
// NewRegexpBoolean returns a Booelan that compares values that must
// be a stringy the given regexp.
func NewRegexpBoolean(re *regexp.Regexp) Boolean {
```

expr/generator.go
```diff
@@ -111,8 +111,11 @@ func (a *AggExpr) Eval(rec *zng.Record) (zng.Value, error) {
 	a.src.Init(rec)
 	for {
 		zv, err := a.src.Next()
+		if err == zng.ErrMissing {
+			continue
+		}
 		if err != nil {
```
nwt, line 117: Our usual pattern is to test for `err != nil` first and then, inside the associated if block, test for specific errors.
```suggestion
		if err != nil {
			if err == zng.ErrMissing {
				continue
			}
```

nwt (approved):

PR #2384, opened 2021-03-20
preserve outer record type names in shaping and put .

### LINKED ISSUE #2379: casting to named type drops typedef name
Running this 
```
type person = {name:string,likes:string,age:int64}
put .=cast(.,type(person))
```
on 
```
{name:"steve",likes:"spicy",age:"52"}
```
gives...
```
{name:"steve",likes:"spicy",age:52}
```
instead of 
```
{name:"steve",likes:"spicy",age:"52"} (=person)
```

mccanne:
This commit fixes a bug where the type name of top-level records
was lost in shaping and in "put .".  The fix is to use zng.Type
instead of zng.TypeRecord where relevant and assert that said Type
is a TypeRecord optionally behind an alias.

Fixes #2379

### DIFF
--- a/expr/shaper.go
+++ b/expr/shaper.go
@@ -9,7 +9,11 @@ import (
 )
 
 // A ShaperTransform represents one of the different transforms that a
-// shaper can apply.
+// shaper can apply.  The transforms are represented as a bit flags that
+// can be bitwide-ored together to create a single shaping operator that
+// represents the composition of all operators.  This composition is efficient
+// as it is carried once per incoming type signature then operator is run
+// for every value of that type.
 type ShaperTransform int
 
 const (
@@ -216,7 +220,7 @@ func (s *step) buildRecord(in zcode.Bytes, b *zcode.Builder) {
 // A shapeSpec is a per-input type ID "spec" that contains the output
 // type and the op to create an output record.
 type shapeSpec struct {
-	typ  *zng.TypeRecord
+	typ  zng.Type
 	step step
 }
 
@@ -224,7 +228,7 @@ type Shaper struct {
 	zctx       *resolver.Context
 	b          zcode.Builder
 	fieldExpr  Evaluator
-	typ        *zng.TypeRecord
+	typ        zng.Type
 	shapeSpecs map[int]shapeSpec // map from type ID to shapeSpec
 	transforms ShaperTransform
 }
@@ -232,7 +236,7 @@ type Shaper struct {
 // NewShaperType returns a shaper that will shape the result of fieldExpr
 // to the provided typExpr. (typExpr should evaluate to a type value,
 // e.g. a value of type TypeType).
-func NewShaperType(zctx *resolver.Context, fieldExpr Evaluator, typ *zng.TypeRecord, tf ShaperTransform) (*Shaper, error) {
+func NewShaper(zctx *resolver.Context, fieldExpr Evaluator, typ zng.Type, tf ShaperTransform) (*Shaper, error) {
 	return &Shaper{
 		zctx:       zctx,
 		fieldExpr:  fieldExpr,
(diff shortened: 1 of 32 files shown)


expr/shaper.go
```diff
@@ -9,7 +9,11 @@ import (
 )
 
 // A ShaperTransform represents one of the different transforms that a
-// shaper can apply.
+// shaper can apply.  The transforms are represented as a bit flags that
+// can be bitwide-ored together to create a single shaping operator that
```
nwt, line 13:
```suggestion
// can be bitwise-ored together to create a single shaping operator that
```

expr/shaper.go
```diff
@@ -9,7 +9,11 @@ import (
...
+// represents the composition of all operators.  This composition is efficient
+// as it is carried once per incoming type signature then operator is run
```
nwt, line 15:
```suggestion
// as it is carried once per incoming type signature and then the resulting operator is run
```

expr/shaper.go
```diff
@@ -267,11 +273,11 @@ func NewShaper(zctx *resolver.Context, fieldExpr, typExpr Evaluator, tf ShaperTr
 		return nil, fmt.Errorf("shaper could not parse type value literal: %s", err)
 	}
 
-	recType, isRecord := zng.AliasedType(shapeType).(*zng.TypeRecord)
+	_, isRecord := zng.AliasedType(shapeType).(*zng.TypeRecord)
 	if !isRecord {
```
nwt, line 277:
```suggestion
	if _, ok := zng.AliasedType(shapeType).(*zng.TypeRecord); !ok {
```

zng/type.go
```diff
@@ -380,3 +380,14 @@ func ReferencedID(typ Type) int {
...
+
+func TypeRecordOf(typ Type) *TypeRecord {
+	for {
+		alias, ok := typ.(*TypeAlias)
+		if !ok {
+			typ, _ := typ.(*TypeRecord)
+			return typ
+		}
+		typ = alias.Type
+	}
```
nwt, line 392:
```suggestion
    typ, _ := zng.AliasOf(typ).(*TypeRecord)
    return typ
```

expr/shaper.go
```diff
@@ -336,7 +343,16 @@ func (s *Shaper) createShapeSpec(inType, spec *zng.TypeRecord) (shapeSpec, error
...
+	var final zng.Type
+	if typ.ID() == s.typ.ID() {
+		// If the underlying records are the same, then use the
+		// spec record as it might be an alias and the intention
+		// would be to cast to the named type.
+		final = s.typ
+	} else {
+		final = typ
+	}
+	return shapeSpec{final, step}, err
```
nwt, line 355:
```suggestion
	final := zng.Type(typ)
	if typ.ID() == s.typ.ID() {
		// If the underlying records are the same, then use the
		// spec type as it might be an alias and the intention
		// would be to cast to the named type.
		final = s.typ
	}
	return shapeSpec{final, step}, err
```

expr/shaper.go
```diff
@@ -336,7 +343,16 @@ func (s *Shaper) createShapeSpec(inType, spec *zng.TypeRecord) (shapeSpec, error
 		}
 	}
 	step, err := createStepRecord(inType, typ)
-	return shapeSpec{typ, step}, err
+	var final zng.Type
+	if typ.ID() == s.typ.ID() {
+		// If the underlying records are the same, then use the
+		// spec record as it might be an alias and the intention
```
nwt, line 349:
```suggestion
		// spec type as it might be an alias and the intention
```

nwt (approved):

zng/type.go
mccanne: This is called AliasedType.  I like AliasOf better so I will change it throughout on this PR.

PR #2386, opened 2021-03-21
allow typedefs to be redefined

### LINKED ISSUE #1568: allow Z type names to be redefined
After further design considerations regarding Z lakes, we realized it would be architecturally favorable to allow Z type names to be redefined.  This is preferred over managing the policing and rejection process when ingesting data into a lake.  This simply means that typedef names may change during a scan and because Z has flexible and forgiving semantics, this will all "just work".  For example, the canonical Z type value will simply have two different values for a type name and the most recent binding in a Z sequence defines the type for any given scan of the sequence. For example, such type definitions will show up in an aggregation by type value and when the resulting aggregate type value keys are scanned the name would simply be redefined in sequence.  Since canonical type values embed the type name mapping, this all just works.

To make this change we should:
* update the tzngio reader, which currently halts with an error with a type tag is reused
* update the zson reader,
* update the zng reader,
* change the zson.Context API to allow alias redefs,
* update the Z spec (ZSON and ZNG)

A technical detail: when type names are redefined, the old type structure and references still exist in the zson.Context with any runtime zng Values referring to them.  For ZST, this is very manageable but means the ZST reader must track which type name bindings are valid in which parts of the columnar data when such typedefs change.  This could be easily stored in the meta-data as another (typically absent or very small) section.  We will put the ZST work in a separate issue.

For tzng, we should allow this in accordance with the XXX in the code (line 132 in tzngio/reader.go)

This will us allow tzng files to be concatenated without error.

While we're at it here, we should change the deprecated "descriptor" references in the tzngio package to refer to record types and type tags instead of descriptors.

With this change, the distrinction between type names as integers and identifiers is less important, though by convention it still is important that locally type names can be chosen from the space of integers as a way to avoid user-defined type names that span a lake.

One more technical point:  in thinking further, the fact that type names can be redefined is really important for high-level types that might represent a common schema where elements of the common schema have union sub-columns and the union content is a function of actual data in the part of the lake that the type refers to.

mccanne:
This commit relaxes the constraint that type names cannot be redefined
to a different type.  This follows the Z design principle of dealing
with whatever the user or client throws at the system as best we can.

The problem of schema versioning is outside the scope of the ZSON
format and can be solved at a higher layer with the appropriate Z
building blocks.

We also removed the end-of-sequence concept from the ZSON spec as
it is not relevant here as the definitions of type contexts and
redefinable types allows an implementation to create a seekable output.

Closes #1568

### DIFF
--- a/zng/docs/zson.md
+++ b/zng/docs/zson.md
@@ -233,11 +233,11 @@ throughout.  That said, external type definitions _may_ vary and an implementati
 must properly handle changing external name by updating the binding between
 the type name and the new type for all subsequent values in the sequence.
 
-> Handling type definition conflicts outside of the scope of ZSON, but an implementation
+> Handling type definition conflicts is outside of the scope of ZSON, but an implementation
 > may police all external definitions so that "type conflicts" are rejected with
-> an error, or it may arrange to name types with a version number, e.g., example.0,
-> example.1, etc then have operators that understand that example.* are all variations
-> of a commonly named type "example."
+> an error, or it may arrange to name types with a version number (e.g., example.0,
+> example.1, etc.), and then have operators that understand that example.* are all variations
+> of a commonly named type "example".
 
 > Note that the semantics of type definitions allows an implementation to "reset" the type
 > context at any point in the sequence causing it to re-emit tyepdefs for each type used
--- a/zng/docs/zson.md
+++ b/zng/docs/zson.md
@@ -1,16 +1,12 @@
 # ZSON - ZNG Structured Object-record Notation
 
-> ### DRAFT 12/2/20
+> ### DRAFT 3/21/21
 > ### Note: This specification is in BETA development.
-> We plan to have a final form established by Spring 2021.
+> We plan to have a final form established soon.
 >
 > ZSON is intended as a literal superset of JSON and NDJSON syntax.
 > If there is anything in this document that conflicts with this goal,
 > please let us know and we will address it.
->
-> Regarding the state of the implementation of the ZSON format in zq:
-> * An input reader is not yet implemented,
-> * End-of-sequence markers are not yet conveyed in the writer.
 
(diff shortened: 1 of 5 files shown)


zng/docs/zson.md
```diff
@@ -236,9 +228,23 @@ Internal type names are represented by integers while external names are
...
 type context, the internal names may be "reused" across type contexts to refer
-to different types but external names must have the same type value across
-subsequences.  It is a "type mismatch" error if an external name has different
-type values across subsequences.
+to different types but external names should generally have the same type value
+throughout.  That said, external type definitions _may_ vary and an implementation
+must properly handle changing external name by updating the binding between
+the type name and the new type for all subsequent values in the sequence.
+
+> Handling type definition conflicts outside of the scope of ZSON, but an implementation
```
nwt, line 236:
```suggestion
> Handling type definition conflicts is outside of the scope of ZSON, but an implementation
```

zng/docs/zson.md
```diff
@@ -236,9 +228,23 @@ Internal type names are represented by integers while external names are
...
+> may police all external definitions so that "type conflicts" are rejected with
+> an error, or it may arrange to name types with a version number, e.g., example.0,
+> example.1, etc then have operators that understand that example.* are all variations
```
nwt, line 239:
```suggestion
> an error, or it may arrange to name types with a version number (e.g., example.0,
> example.1, etc.), and then have operators that understand that example.* are all variations
```

zng/docs/zson.md
```diff
@@ -236,9 +228,23 @@ Internal type names are represented by integers while external names are
...
+> of a commonly named type "example."
```
nwt, line 240:
```suggestion
> of a commonly named type "example".
```

nwt (approved):

PR #2389, opened 2021-03-21
TZNG->ZSON transition ztests below lake/

philrz:


ppl/lake/ztests/import-check.yaml
```diff
@@ -1,14 +1,14 @@
...
-      #0:record[a:string,b:string]
-      0:[a;]
+      {a:"a",b:"b"} (=foo)
+      {a:"a"} (foo)
 
 outputs:
   - name: stderr
     regexp: |
-      .*record missing a field
+      .*record decorator columns \(2\) mismatched with value columns \(1\)
```
philrz, line 14: This was the one test where I made a material change to achieve what seemed to be an equivalent test with ZSON input data. Feel free to propose adjustments or advocate that I leave the TZNG one for now.

mccanne: That error message looks like it could be improved but we can put that in another issue...

mccanne (approved):

ppl/lake/ztests/import-check.yaml
philrz: Thanks @mccanne. I opened #2394 regarding the error message.

PR #2393, opened 2021-03-22
add missing(), has(), and nameof()

mccanne:
This commit adds some sorely needed functions that will be useful for
data discovery, exploration, classification, and shaping.  missing()
returns true iff any of the arguments are missing.  has() returns
true iff all of the arguments are not missing.  nameof() returns
the type name of the single expression if the type is an alias;
otherwise, it returns missing.

### DIFF
--- a/compiler/kernel/expr.go
+++ b/compiler/kernel/expr.go
@@ -386,7 +386,7 @@ func compileShaper(zctx *resolver.Context, scope *Scope, node ast.Call) (*expr.S
 }
 
 func compileCall(zctx *resolver.Context, scope *Scope, call ast.Call) (expr.Evaluator, error) {
-	// For now, we special case stateful functions here.  We shuold generalize this
+	// For now, we special case stateful functions here.  We should generalize this
 	// as we will add many more stateful functions and also resolve this
 	// the changes to create running aggegation functions from reducers.
 	// XXX See issue #1259.
@@ -426,7 +426,7 @@ func compileCall(zctx *resolver.Context, scope *Scope, call ast.Call) (expr.Eval
 	nargs := len(call.Args)
 	fn, root, err := function.New(zctx.Context, call.Name, nargs)
 	if err != nil {
-		return nil, fmt.Errorf("%s: %w", call.Name, err)
+		return nil, fmt.Errorf("%s(): %w", call.Name, err)
 	}
 	args := call.Args
 	if root {
@@ -434,7 +434,7 @@ func compileCall(zctx *resolver.Context, scope *Scope, call ast.Call) (expr.Eval
 	}
 	exprs, err := compileExprs(zctx, scope, args)
 	if err != nil {
-		return nil, fmt.Errorf("%s: bad argument: %w", call.Name, err)
+		return nil, fmt.Errorf("%s(): bad argument: %w", call.Name, err)
 	}
 	return expr.NewCall(zctx, fn, exprs), nil
 }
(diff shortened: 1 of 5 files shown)


expr/eval.go
```diff
@@ -760,6 +760,48 @@ func (e *Exists) Eval(rec *zng.Record) (zng.Value, error) {
...
+}
+
+func NewMissing(exprs []Evaluator) *Missing {
+	return &Missing{exprs}
+}
+
+func (m *Missing) Eval(rec *zng.Record) (zng.Value, error) {
+	for _, e := range m.exprs {
+		_, err := e.Eval(rec)
+		if err != nil {
```
nwt, line 774:
```suggestion
		if _, err := e.Eval(rec); err != nil {
```

expr/eval.go
```diff
@@ -760,6 +760,48 @@ func (e *Exists) Eval(rec *zng.Record) (zng.Value, error) {
...
+}
+
+func NewHas(exprs []Evaluator) *Has {
+	return &Has{exprs}
+}
+
+func (h *Has) Eval(rec *zng.Record) (zng.Value, error) {
+	for _, e := range h.exprs {
+		_, err := e.Eval(rec)
+		if err != nil {
```
nwt, line 795:
```suggestion
		if _, err := e.Eval(rec); err != nil {
```

expr/function/function.go
```diff
@@ -155,6 +157,18 @@ func (t *typeOf) Call(args []zng.Value) (zng.Value, error) {
 	return t.zctx.LookupTypeValue(typ), nil
 }
 
+type nameOf struct{}
+
+func (*nameOf) Call(args []zng.Value) (zng.Value, error) {
+	typ := args[0].Type
+	if alias, ok := typ.(*zng.TypeAlias); ok {
+		// XXX GC
+		//fmt.Println("NAMEOF", typ.ZSON())
```
nwt, line 166:
```suggestion
```

compiler/kernel/expr.go
```diff
@@ -403,9 +403,21 @@ func compileCall(zctx *resolver.Context, scope *Scope, call ast.Call) (expr.Eval
 	case call.Name == "exists":
 		exprs, err := compileExprs(zctx, scope, call.Args)
 		if err != nil {
-			return nil, fmt.Errorf("exists: bad argument: %w", err)
+			return nil, fmt.Errorf("exists(): bad argument: %w", err)
```
nwt, line 406: Looks like error strings here are inconsistent when it comes to adding this `name():` prefix.

Maybe name the return parameters and add then add the prefix in a `defer` ?

```go
defer func() {
    if err != nil {
        err = fmt.Errorf("%s(): %w", call.Name, err)
    }
}()    
```

nwt (approved):

compiler/kernel/expr.go
mccanne: Maybe I'm doing something wrong, but this involves a bit of mess getting the scope right everywhere on err, then defining locally scoped vars for each non-error value.  I think I'll just make it consistent for now and we can rework it later.  Also, I find code with named returned values harder to read compared to explicit return values.

nwt: Does the trouble stem from using unadorned returns? I'm no fan of those either, but you don't have to use them in order to have named result parameters.

mccanne: I will commit as is and you can show me the light...

PR #2395, opened 2021-03-22
TZNG->ZSON transition ztests below ppl/zqd/

philrz:


ppl/zqd/ztests/ls.yaml
```diff
@@ -11,7 +11,7 @@ script: |
   # Verify that output flags are honored.
   zapi -h $ZQD_HOST ls -l -f zson | zq -i zson -f zson "pick name,storage_kind" -
   echo ===
-  zapi -h $ZQD_HOST ls -l -f zng | zq -i zng -f tzng "pick name,storage_kind" -
+  zapi -h $ZQD_HOST ls -l -f zng | zq -i zng -f table "pick name,storage_kind" -
```
nwt, line 14: This fine but I think `-z` would be better.

nwt (approved):

ppl/zqd/ztests/ls.yaml
philrz: @nwt: Given the tear that I was on, `-z` is what I did at first. Then Is saw the comment expressing that a motivation for the test was to "Verify that output flags are honored." Since the prior command already exercised the `-f zson` output flag, I chose to exercise a different one. I'm happy to switch it back to `-z`, but is there an argument for doing so besides "make all the things Z"? :smile:

nwt: I see your point, though I do think "the output flags" here refers to `zapi`'s output flags and changing the `zq` output flag to `-z` would be fine. But so would leaving this as it is.

nwt (approved):

PR #2396, opened 2021-03-22
Populate table of valid durations

philrz:
Now that #2371 is done, it looks like we can phase out this XXX.

### DIFF
--- a/zql/docs/grouping/README.md
+++ b/zql/docs/grouping/README.md
@@ -19,16 +19,21 @@ To create batches of events that are close together in time, specify
 `every <duration>` before invoking your aggregate function(s).
 
 The `<duration>` may be expressed in any of the following units of time. A
-numeric value must precede the unit specification.  The syntax conforms
-to the `dur-time` format of RFC-3339, with the `t` prefix dropped.
+numeric value must precede the unit specification.
 
-| **Unit**  | **Suffix** |
-|-----------|------------|
-| `second`  | `s`        |
-| `minute`  | `m`        |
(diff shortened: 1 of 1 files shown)


zql/docs/grouping/README.md
```diff
@@ -19,16 +19,21 @@ To create batches of events that are close together in time, specify
...
-to the `dur-time` format of RFC-3339, with the `t` prefix dropped.
-
-| **Unit**  | **Suffix** |
-|-----------|------------|
-| `second`  | `s`        |
-| `minute`  | `m`        |
-| `hour`    | `h`        |
-
-XXX finish RFC-3339
+numeric value must precede the unit specification.
```
nwt, line 22: Don't know if you want to get into this since it isn't likely to be useful in practice, but (a) the numeric value can be integer or decimal and (b) a duration can be a single numeric value followed by unit suffix or a sequence of them (e.g., 1h5m10s).

nwt (approved):

zql/docs/grouping/README.md
philrz: @nwt: We're a long way from being comprehensive in the docs, but, yeah, I figure it's worth adding while it's fresh in our minds. I modified the text to include this and also included one more example so we're showing all the variations. Hopefully that doesn't distract too much from the primary use case. Please do give it another look and let me know what you think before I merge.

nwt: I like it!

nwt (approved):
:shipit:

PR #2403, opened 2021-03-24
add complex literals

### LINKED ISSUE #2183: support complex expressions like `put x={a:1+3, b:len(foo)}`
We should add support for complex expressions for records, arrays, sets, maps, etc.  We could start with records and arrays like javascript. 

The syntax for records is 
```
{ <field> : <expr>, <field> : <expr> , ... }
```
The syntax for arrays is 
```
[ <expr>, <expr> , ... ]
```
etc

The runtme would have new expression ast nodes and new expr Evaluators that would evaluate each expression, build the zcode, and look up the zng.Type to form the zng.Value.

This is not particularly difficult at this point given all the progress on the software architecture.

mccanne:
This commit adds support for complex value expressions.
There is still some ambiguity issues with IP6 addresses
and map-literal colons, but there seem to be some broader
problems with IP6 addresses that will be handled after
we switch from :type casts to pythonic/go type() casts.

We also fixed a bug in the zson analyzer where a non-nil
zng.Type holding a nil zng.TypeAlias could appear for
undefined alias names causing a nil derefence instead of
an error.

Closes #2183.


zson/analyzer.go
```diff
@@ -595,10 +595,11 @@ func (a Analyzer) convertType(zctx *Context, typ ast.Type) (zng.Type, error) {
 	case *ast.TypeName:
 		typ, ok := a[t.Name]
 		if !ok {
-			typ = zctx.LookupTypeDef(t.Name)
-			if typ == nil {
+			alias := zctx.LookupTypeDef(t.Name)
+			if alias == nil {
 				return nil, fmt.Errorf("no such type name: %q", t.Name)
 			}
+			typ = alias
```
nwt, line 602: Nit: I think the original code was clearer.

nwt (approved):

zson/analyzer.go
mccanne: It was clearer but it had the interface-nil bug because typ is a zng.Type and LookupTypeDef returns *zng.TypeAlias.  I will add a comment in the code.  (I had put a note in the PR comment.)

expr/values.go
```diff
@@ -0,0 +1,233 @@
...
+		if first == nil {
+			first = typ
+		}
+		if typ != inner && typ != zng.TypeNull {
+			if typ == first || first == zng.TypeNull {
+				a.typ = a.zctx.LookupTypeArray(zv.Type)
+				inner = a.typ.Type
+				container = zng.IsContainerType(inner)
+			} else {
+				return zng.NewErrorf("illegal mixed type array"), nil
```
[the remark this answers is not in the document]
mccanne: I think this is ok for now and yes it was deliberate.  We can fold this into a more comprehensive issue about what to do for mixed type arrays because we could handle this case but also turn them in to array of unions if the mixed elements aren't type compatible.  Also, the current errors are embedded so they would shed light to the user as to the mixed-type nature of their query (instead of just stopping with a type error).  I'm on the fence as to just doing something that works or generating errors and making the user cast, e.g., `put ports=[80:port, id.resp_p, id.resp_p]`.  Also, if you have a missing value should the result be missing or be a partial results (e.g., `foo:null` for a record field that is missing).  See issue #2411

PR #2406, opened 2021-03-24
add ISO time literal

### LINKED ISSUE #2195: Time and duration literals in Z
Z should probably at some point support time and duration literals, like it does for addrs and nets.

### LINKED ISSUE #2335: revamp Z literals
Z literals should be reimplement to better align with ZSON.  This involves adding complex literal values (records, arrays, maps, etc) along with ISO dates and durations.

mccanne:
This commit adds support to the Z grammar for ISO time literals
and exposes durations and time literals into the expression
grammar.  We also added support for searches in the generic
expression compilation path as this can occur when match() is
used in expression context, e.g., put b=match(foo), which
we used for the time/duration literal testing.

Closes #2335 
Closes #2195

### DIFF
--- a/compiler/kernel/expr.go
+++ b/compiler/kernel/expr.go
@@ -68,6 +68,13 @@ func compileExpr(zctx *resolver.Context, scope *Scope, e ast.Expr) (expr.Evaluat
 			return expr.NewVar(ref), nil
 		}
 		return nil, fmt.Errorf("unknown reference: '%s'", e.Name)
+	case *ast.Search:
+		f, err := compileSearch(e)
+		if err != nil {
+			return nil, err
+		}
+		return expr.FilterEvaluator(f), nil
+
 	case *ast.Path:
 		return expr.NewDotExpr(field.Static(e.Name)), nil
 	case *ast.UnaryExpr:
--- a/compiler/kernel/expr.go
+++ b/compiler/kernel/expr.go
@@ -74,7 +74,6 @@ func compileExpr(zctx *resolver.Context, scope *Scope, e ast.Expr) (expr.Evaluat
 			return nil, err
 		}
 		return expr.FilterEvaluator(f), nil
-
 	case *ast.Path:
 		return expr.NewDotExpr(field.Static(e.Name)), nil
 	case *ast.UnaryExpr:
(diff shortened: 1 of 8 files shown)


expr/boolean.go
```diff
@@ -114,6 +114,46 @@ func CompareInt64(op string, pattern int64) (Boolean, error) {
 	}, nil
 }
 
+func CompareTime(op string, pattern int64) (Boolean, error) {
+	CompareInt, ok1 := compareInt[op]
+	CompareFloat, ok2 := compareFloat[op]
+	if !ok1 || !ok2 {
+		return nil, fmt.Errorf("unknown int comparator: %s", op)
+	}
+	// many different zeek data types can be compared with integers
```
nwt, line 123:
```suggestion
	// many different ZNG data types can be compared with integers
```

expr/boolean.go
```diff
@@ -114,6 +114,46 @@ func CompareInt64(op string, pattern int64) (Boolean, error) {
...
+		case zng.IdTime:
+			ts, err := zng.DecodeTime(zv)
+			if err == nil {
+				return CompareInt(int64(ts), pattern)
+			}
+		case zng.IdDuration:
+			v, err := zng.DecodeInt(zv)
+			if err == nil {
+				return CompareInt(int64(v), pattern)
+			}
```
nwt, line 151: Maybe fold these two cases in with the signed integer cases?

compiler/kernel/expr.go
```diff
@@ -68,6 +68,13 @@ func compileExpr(zctx *resolver.Context, scope *Scope, e ast.Expr) (expr.Evaluat
 			return expr.NewVar(ref), nil
 		}
 		return nil, fmt.Errorf("unknown reference: '%s'", e.Name)
+	case *ast.Search:
+		f, err := compileSearch(e)
+		if err != nil {
+			return nil, err
+		}
+		return expr.FilterEvaluator(f), nil
+
```
nwt, line 77: Nit: There are no blank lines between the other cases here.
```suggestion
```

nwt (approved):

expr/boolean.go
mccanne: I did that first and the resulting bug is that int64 is scaled by 1_000_000 for time and dur comparisons as it presumes s not ns as the unit.

PR #2407, opened 2021-03-24
allow record index by string as l-value

### LINKED ISSUE #2326: Dot-containing field name in expression casues "invalid expression on lhs of assignment"
Repro is with `zq` commit 3475ad3.

I was trying to make some enhancements to the Z shaper that merged with #2300. As part of this, I was attempting a data transport on a Zeek field _before_ the `put .=unflatten(.)` call. This means working with the field as if its name contained a literal `.` as opposed to treating it like the nested record it would normally be after the `unflatten()` call.

@mccanne helped me find the correct Z syntax to work with the field despite the special characters in its name, but it currently doesn't work, with the following error:

```
$ echo '{"certificate.not_valid_before":"2011-09-10T19:15:02.000000Z"}' | zq -z 'put .["certificate.not_valid_before"]=iso(.["certificate.not_valid_before"])' -
invalid expression on lhs of assignment
```

mccanne:
This commit fixes a bug where l-values could not be index operations
with fields referenced as string, e.g., put .["x"] = 123.
The fix is to detect this case in the semantic pass and extend
the field-reference path with the string constant.

Fixes #2326

### DIFF
--- a/compiler/semantic/expr.go
+++ b/compiler/semantic/expr.go
@@ -162,7 +162,7 @@ func semBinary(scope *Scope, e *ast.BinaryExpr) (ast.Expr, error) {
 		return nil, err
 	}
 	// If we index a root record with a string constant, then just
-	// extend the path...
+	// extend the path.
 	if op == "[" {
 		if path := isRootIndex(scope, lhs, rhs); path != nil {
 			return path, nil
@@ -191,10 +191,8 @@ func isStringConst(scope *Scope, e ast.Expr) (string, bool) {
 		return p.Text, true
 	}
 	if ref, ok := e.(*ast.Ref); ok {
-		if p := scope.Lookup(ref.Name); p != nil {
-			if c, ok := p.(*ast.Const); ok {
-				return isStringConst(scope, c.Expr)
-			}
+		if c, ok := scope.Lookup(ref.Name).(*ast.Const); ok {
+			return isStringConst(scope, c.Expr)
 		}
 	}
 	return "", false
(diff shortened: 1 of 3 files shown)


compiler/semantic/expr.go
```diff
@@ -161,6 +161,13 @@ func semBinary(scope *Scope, e *ast.BinaryExpr) (ast.Expr, error) {
 	if err != nil {
 		return nil, err
 	}
+	// If we index a root record with a string constant, then just
+	// extend the path...
```
nwt, line 165:
```suggestion
	// extend the path.
```

compiler/semantic/expr.go
```diff
@@ -169,6 +176,30 @@ func semBinary(scope *Scope, e *ast.BinaryExpr) (ast.Expr, error) {
...
+func isStringConst(scope *Scope, e ast.Expr) (string, bool) {
+	if p, ok := e.(*ast.Primitive); ok && p.Type == "string" {
+		return p.Text, true
+	}
+	if ref, ok := e.(*ast.Ref); ok {
+		if p := scope.Lookup(ref.Name); p != nil {
+			if c, ok := p.(*ast.Const); ok {
+				return isStringConst(scope, c.Expr)
+			}
+		}
```
nwt, line 198: Maybe shorten since type assertions fail on `nil`?
```suggestion
		if c, ok := scope.Lookup(ref.Name).(*ast.Const); ok {
			return isStringConst(scope, c.Expr)
		}
```

nwt (approved):

Issue #2412, opened 2021-03-24
"zar map" can't easily access indexes by name anymore


philrz:
The section of the `zar` README on "multi-key custom indexes" has traditionally included this text (as of `zq` commit 693015b):

> So, let's say we wanted a count of all bytes received by 10.47.6.173 as the originator, which is the secondary key. While we could build a different custom index where id.orig\_h is the primary key, we could also just scan the custom2 index using brute force:
> 
> ```
> zar map id.orig_h=10.47.6.173 custom2.zng | zq -f text "sum(resp_bytes)" -
> ```

Using GA `zq` tagged `v0.23.0` that was compatible with that revision of the README, here's the steps that build up to that point, showing how it used to work ok:

```
$ rm -rf logs
$ mkdir ./logs
$ set ZAR_ROOT=`pwd`/logs
$ zq zng/*.gz | zar import -s 25MB -
$ zar index -o custom2.zng -k id.resp_h,id.orig_h -z "_path=conn | resp_bytes=sum(resp_bytes) by id.resp_h,id.orig_h | sort id.resp_h,id.orig_h"
$ zar map id.orig_h=10.47.6.173 custom2.zng | zq -f text "sum(resp_bytes)" -
123488134
```

The reason why it worked ok is because the index was saved in the filename specified via `-o` and hence could easily be accessed as the `file` parameter of `zar map`.

```
$ tree -s logs
logs
├── [         78]  zar.json
└── [         96]  zd
    └── [        576]  20180324
        ├── [   23200224]  d-1qCyQ1rvDAUcsQ3Jhg9kFJQsT7C.zng
        ├── [         96]  d-1qCyQ1rvDAUcsQ3Jhg9kFJQsT7C.zng.zar
        │   └── [      78752]  custom2.zng
        ├── [   22914143]  d-1qCyQ6pmAFUxUqGw7gzQ9828byu.zng
        ├── [         96]  d-1qCyQ6pmAFUxUqGw7gzQ9828byu.zng.zar
        │   └── [      80714]  custom2.zng
        ├── [   10347895]  d-1qCyQPyVGHdMzwf2K6BQPcSaOD2.zng
        ├── [         96]  d-1qCyQPyVGHdMzwf2K6BQPcSaOD2.zng.zar
        │   └── [      68006]  custom2.zng
        ├── [   23382686]  d-1qCyQQRv71VuZHzyTI3CJG6bRcI.zng
[...]

Issue #2398, opened 2021-03-22
Zed: Make the reserved words as limited in scope as possible


philrz:
Repro is with `zar` commit e6b1fda.

While working on bringing the `zar` README current (#2266), I bumped into this example in the "simple graph queries" section that's now a parse error:

```
$ zar map -o forward.zng "id.orig_h != null | put from=id.orig_h,to=id.resp_h | count() by from,to"
error parsing Z at column 29:
id.orig_h != null | put from=id.orig_h,to=id.resp_h | count() by from,to
                        === ^ ===
```

I confirmed with @mccanne that this is due to "from" becoming a reserved word in the Z language as part of the recently-added SQL support. However, he also confirmed that we can do better:

> Yes, this should be quite fixable.  In the meantime, you can also use `\from`
> 
> We should make an issue to make the reserved words as limited in scope as possible.

philrz:
I bumped into this again while assembling chanegs for #2448. The problem was in the context of `type` specifications for the reference shaping config for Zeek data. Here were the specific lines from the config that had stopped working when the changes merged for #2215. You can see the backslashes where I had to add escaping.

```
type intel={_path:string,ts:time,uid:bstring,id:id,seen:{indicator:bstring,indicator_type:zenum,\where:zenum,node:bstring},matched:|[zenum]|,sources:|[bstring]|,fuid:bstring,file_mime_type:bstring,file_desc:bstring,_write_ts:time};
type kerberos={_path:string,ts:time,uid:bstring,id:id,request_type:bstring,client:bstring,service:bstring,success:bool,error_msg:bstring,\from:time,till:time,cipher:bstring,forwardable:bool,renewable:bool,client_cert_subject:bstring,client_cert_fuid:bstring,server_cert_subject:bstring,server_cert_fuid:bstring,_write_ts:time};
type smtp={_path:string,ts:time,uid:bstring,id:id,trans_depth:uint64,helo:bstring,mailfrom:bstring,rcptto:|[bstring]|,date:bstring,\from:bstring,to:|[bstring]|,cc:|[bstring]|,reply_to:bstring,msg_id:bstring,in_reply_to:bstring,subject:bstring,x_originating_ip:ip,first_received:bstring,second_received:bstring,last_reply:bstring,path:[ip],user_agent:bstring,tls:bool,fuids:[bstring],is_webmail:bool,_write_ts:time};
```

philrz:
[...]

PR #2414, opened 2021-03-24
zar README updates

### LINKED ISSUE #2266: Update zar README
We recently considered pointing a community user at `zar` to handle a particular use case. At that moment I recalled a recent PR that changed the meaning of the `-z` flag in `zar find` that's used several times in the README, so I wanted to make sure this didn't break the flow. That motivated me to revisit the README as a user with a recent `zar` commit, which reveals several other things there that are now out of sync:

*   TZNG was featured prominently in the README, but ZSON is now the preferred text format
*   Creating an index with `zar index` now requires an explicit `create` command where it did not before
*   The naming of microindex files has changed, so users need different guidance in the places where we advise them to look at their contents
*   The exercise has also revealed at least one regression in the functionality highlighted there.

I've got a branch named `zar-readme-feb2021` where I'm starting to make these changes as I figure out the correct changes and/or code fixes land. I'll put up a PR once I've got the README in the branch in a state where I can run through it cleanly.

philrz:
The `zar` README right now is hazardous because it's out of sync with how the command is currently actually working in `master` (options have changed, etc.) Some of the command lines in the README are pretty big, so it'd be asking a lot of a new user to figure out the adjustments on their own if they were to try it and see the commands failing. Therefore I'm antsy to get these in sync again.

I'm putting several comments in-line where I want to draw everyone's attention to some of the awkward spots. If folks want to hurry up and make some changes to `zar` functionality to make this less awkward, I'm open to that, assuming we can turn it around pretty quickly.

Closes #2266


ppl/cmd/zar/README.md
```diff
@@ -32,26 +32,24 @@ ln -s zq-sample-data/zng
...
+Let's take those logs and ingest them into a directory.  Often we'd keep our
+archive somewhere like [Amazon S3](https://aws.amazon.com/s3/), but since we're
+just doing a quick test we'll use local temp space.  We'll make it easier to
+run all the commands by setting an environment variable pointing to the root of
+the logs tree.
 ```
-mkdir ./logs
-set ZAR_ROOT=`pwd`/logs
+mkdir /tmp/logs
+ZAR_ROOT=/tmp/logs
```
philrz, line 42: Since the test data is small, I went ahead and advocated building the archive in `/tmp` (and cited S3 as a logical alternative so readers don't get the mistaken impression that we actually advocate putting stuff in temp storage in production). In the past when we had the user just creating it in the current directory, a bunch of the sample outputs further down in the README had to have `/path/to/ZAR_ROOT` in them, whereas using a known dir like `/tmp` lets us avoid that. That addresses two little things:

1.  Whenever I'd have to update the example outputs when we change functionality (it's happened more than you might think :smile: ), I'd have to keep search & replacing the pathnames from the outputs on my system to put back the `/path/to/ZAR_ROOT` part. Not a crisis, just annoying.  
     
2.  If I can get this PR merged, I'd like to see about writing a ztest that executes as much as possible of the stuff in this README so we don't keep regressing. I expect the output checks in that will be much easier to implement/understand if the pathname is "locked in".

ppl/cmd/zar/README.md
```diff
@@ -32,26 +32,24 @@ ln -s zq-sample-data/zng
...
 Now, let's ingest the data using "zar import".  We are working on more
 sophisticated ways to ingest data (e.g., by the standard time partitioning
 techniques of year/month/day/hour etc), but for now zar import just chops
-its input into chunk files of approximately equal size.
-
-Zar import expects its input to be
-in the zng format so we'll use zq to take all the zng logs, gunzip them,
-and feed them to "zar import", which here expects its data on stdin.  We'll chop them
-into chunks of 25MB, which is very small, but in this example the data set is
+its input into chunk files of approximately equal size. We'll chop into chunks
```
philrz, line 48: I expect this wording needs to be updated, since we do at least _some_ time partitioning and sort the data by time, whereas this text looks like an artifact from when we did none of that. I'd appreciate it if someone could make a suggestion with the most accurate summary of how sophisticated (or not) it is currently.

ppl/cmd/zar/README.md
```diff
@@ -32,26 +32,24 @@ ln -s zq-sample-data/zng
...
+of 25MB, which is very small, but in this example the data set is
 fairly small (175MB) and you can always try it out on larger data sets:
 ```
-zq zng/*.gz | zar import -s 25MB -
+zar import -s 25MB zng/*.gz
```
philrz, line 52: I had a [Slack thread](https://brim-labs.slack.com/archives/CP3BMAEEP/p1614885363079800) with @alfred-landrum where I confirmed that this change was appropriate to make. I paused at the time because moving from the prior command line to the newer/simpler one actually changed how the events were distributed across the chunks, and I wanted to make sure that was expected and not a bug. His take at the time:

> It's not a problem, but we should change the readme to use the `zar import <file> ...` since that's the simpler usage. I believe the difference between the two ways is that when `zq` is given multiple files, it tries to merge them into a single stream by looking for a `ts` field, and assumes they are in ascending order, whereas `zar import` does something similar, but assumes the files are using the same order as the archive (which defaults to descending).

ppl/cmd/zar/README.md
```diff
@@ -101,10 +99,10 @@ zar map "count()" | zq -f table -
...
 ```
 COUNT
-188697
-436829
-409224
-427328
+275745
+276326
+290789
+619218
```
philrz, line 105: Here's the change in record distribution across chunks as mentioned in the previous comment.

ppl/cmd/zar/README.md
```diff
@@ -265,54 +278,54 @@ zar ls custom.zng
 ```
 To see what's in it:
 ```
-find $ZAR_ROOT -name custom.zng | head -n 1 | xargs zq -f table 'head 10' -
+find $ZAR_ROOT -name idx-`for file in $ZAR_ROOT/indexdefs/*; do   zq -z $file | grep 'name:"custom.zng"' > /dev/null; if [ $? = 0 ]; then echo $file | sed 's/.*indexdefs\/idxdef-//'; fi done` | head -n 1 | xargs zq -f table 'head 10' -
```
philrz, line 281: Why yes... that `for` loop in the backticks _is_ ugly! :smile:   The problem here is similar in spirit to the one described in #2412, in that the opaque naming of the index files left me scratching my head wondering how to best guide the user toward looking at the contents of one of them going off the index name. Since this is a "show me" moment and not something we expect users would do during production use, I'm ok leaving this as-is, but I'm open to counterarguments.

ppl/cmd/zar/README.md
```diff
@@ -265,54 +278,54 @@ zar ls custom.zng
...
 index:
 ```
-zar find -z -x custom.zng 10.164.94.120 | zq -t -
+zar find -z -x custom.zng 10.164.94.120
 ```
 Now we're talking!  And if you take the results and do a little more math to
 aggregate the aggregations, like this:
 ```
-zar find -z -x custom.zng 10.164.94.120 | zq -f table "count=sum(count) by _path" -
+zar find -x custom.zng 10.164.94.120 | zq -f table "count=sum(count) by _path | sort -r" -
```
philrz, line 302: I intentionally dropped the `-z` here to try and show the best practice of "for best performance, always using binary ZNG when you're piping to downstream `zq`" (but I also didn't wanna get into a big preachy tangent about that either). Part of me wanted to leave the `-z` there just to show me adding to the prior command line and not have the reader wondering if there was significance my dropping it, since it's probably not a lesson that needs to be driven home in this small-scaled emo. A close call in my mind, but probably not a big deal either way.

ppl/cmd/zar/README.md
```diff
@@ -364,7 +384,7 @@ a count of all bytes received by 10.47.6.173 as the originator, which is the
 secondary key.  While we could build a different custom index where id.orig_h
 is the primary key, we could also just scan the custom2 index using brute force:
 ```
-zar map id.orig_h=10.47.6.173 custom2.zng | zq -f text "sum(resp_bytes)" -
+zar map id.orig_h=10.47.6.173 idx-`for file in $ZAR_ROOT/indexdefs/*; do   zq -z $file | grep 'name:"custom2.zng"' > /dev/null; if [ $? = 0 ]; then echo $file | sed 's/.*indexdefs\/idxdef-//'; fi done` | zq -f text "sum(resp_bytes)" -
```
philrz, line 387: _This_ ugly `for` loop in backticks is a workaround for #2412. I don't know if this use case for `zar map` is something we'll want to keep supporting, but if we did address #2412, we could make this one simpler.

ppl/cmd/zar/README.md
```diff
@@ -386,15 +406,15 @@ The classic map-reduce example is word count.  Let's do this example with
 the uri field in http logs.  First, we map each record that has a uri
 to a new record with that uri and a count of 1:
 ```
-zar map -o words.zng "uri != null | cut uri | put count=1"
+zar map -q -o words.zng "uri != null | cut uri | put count=1"
 ```
 again you can look at one of the files...
 ```
-find $ZAR_ROOT -name words.zng | head -n 1 | xargs zq -t -
+find $ZAR_ROOT -name words.zng ! -size 0 | head -n 1 | xargs zq -z -
```
philrz, line 413: The `! -size 0` became necessary because, with the new distribution of records across chunks, some of the chunks completely lack events containing the `uri` field, hence some zero-length `words.zng` files end up in the tree, and we want to avoid those.

ppl/cmd/zar/README.md
```diff
@@ -433,43 +453,43 @@ run the following commands to a create a multi-key search index
...
 where each index includes a count of the occurrences.
 ```
-zar map -o forward.zng "id.orig_h != null | put from=id.orig_h,to=id.resp_h | count() by from,to"
-zar map -o reverse.zng "id.orig_h != null | put from=id.resp_h,to=id.orig_h | count() by from,to"
-zar map -o directed-pairs.zng "count=sum(count) by from,to | sort from,to" forward.zng reverse.zng
-zar index -i directed-pairs.zng -o graph.zng -k from,to -z "*"
+zar map -o forward.zng "id.orig_h != null | put from_addr=id.orig_h,to_addr=id.resp_h | count() by from_addr,to_addr"
+zar map -o reverse.zng "id.orig_h != null | put from_addr=id.resp_h,to_addr=id.orig_h | count() by from_addr,to_addr"
+zar map -o directed-pairs.zng "count=sum(count) by from_addr,to_addr | sort from_addr,to_addr" forward.zng reverse.zng
+zar index create -i directed-pairs.zng -o graph.zng -k from_addr,to_addr -z "*"
```
philrz, line 459: I put in all this `_addr` stuff as a workaround to #2398. @mccanne pointed out that I could have also escaped this as `\from`, but I figured that we could just be more explicit rather than get into a tangent about what is hopefully a temporary workaround, or leaving the reader scratching their head wondering why we put that backslash in there.

ppl/cmd/zar/README.md
nwt, line 42: 1.  To get this into the environment, `export` is required.
2.  Nit: Might as well use the variable instead of repeating the path.

```suggestion
export ZAR_ROOT=/tmp/logs
mkdir $ZAR_ROOT
```

ppl/cmd/zar/README.md
nwt, line 48: Nit: Prevailing style in here is two spaces between sentences.

```suggestion
its input into chunk files of approximately equal size.  We'll chop into chunks
```

ppl/cmd/zar/README.md
```diff
@@ -125,17 +123,40 @@ or...
...
-```
-which gives this somewhat cryptic result in text zng format:
-```
-#port=uint16
-#zenum=string
-#0:record[_path:string,ts:time,uid:bstring,id:record[orig_h:ip,orig_p:port,resp_h:ip,resp_p:port],proto:zenum,service:bstring,duration:duration,orig_bytes:uint64,resp_bytes:uint64,conn_state:bstring,local_orig:bool,local_resp:bool,missed_bytes:uint64,history:bstring,orig_pkts:uint64,orig_ip_bytes:uint64,resp_pkts:uint64,resp_ip_bytes:uint64,tunnel_parents:set[bstring]]
-0:[conn;1521911721.307472;C4NuQHXpLAuXjndmi;[10.10.23.2;11;10.0.0.111;0;]icmp;-;1260.819589;23184;0;OTH;-;-;0;-;828;46368;0;0;-;]
+zar zq -Z "id.orig_h=10.10.23.2"
+```
+which gives this result in the [ZSON](../../../zng/docs/zson.md) format. ZSON
```
nwt, line 128: Nit: Prevailing style in here is two spaces between sentences.
```suggestion
which gives this result in the [ZSON](../../../zng/docs/zson.md) format.  ZSON
```

ppl/cmd/zar/README.md
nwt:
```suggestion
find $ZAR_ROOT -name idx-$(zar index ls -f zng | zq -f text 'desc="zql-custom.zng" | cut id' -).zng | head -n 1 | xargs zq -f table 'head 10' -
```

ppl/cmd/zar/README.md
nwt:
```suggestion
zar map id.orig_h=10.47.6.173 idx-$(zar index ls -f zng | zq -f text 'desc="zql-custom2.zng" | cut id' -).zng | zq -f text "sum(resp_bytes)" -```

File: ppl/cmd/zar/README.md (hunk 1 above)
Comment (philrz) [reply ppl/cmd/zar/README.md:42]: @nwt: Thanks for catching the `export` problem. Everything had been working for me in my runs through the README because I'd had a pre-existing `export ZAR_ROOT` in my `.bashrc` and so it was still having the intended impact even after I was changing the actual value for `ZAR_ROOT`. :roll\_eyes:

File: ppl/cmd/zar/README.md (hunk 2 above)
Comment (mccanne) [reply ppl/cmd/zar/README.md]: Coming soon... arbitrary partition keys and auto-sizing of partitions.

File: ppl/cmd/zar/README.md (hunk 5 above)
Comment (mccanne) [reply ppl/cmd/zar/README.md:281]: This doc change is good for now but this is an area where we will improve the UX with a better model for naming things inside of a lake... space, subspace, meta-layer, etc...

**Review by nwt — APPROVED** [review]

File: ppl/cmd/zar/README.md [hunk 11]
```diff
@@ -32,26 +32,25 @@ ln -s zq-sample-data/zng
...
-in the zng format so we'll use zq to take all the zng logs, gunzip them,
-and feed them to "zar import", which here expects its data on stdin.  We'll chop them
-into chunks of 25MB, which is very small, but in this example the data set is
-fairly small (175MB) and you can always try it out on larger data sets:
+sophisticated ways to ingest data (e.g., arbitrary partition keys and
+auto-sizing of partitions) but for now zar import just chops its input into
+chunk files of approximately equal size, each sorted by timestamp in descending
+order.  We'll chop into chunks of 25MB, which is very small, but in this
+example the data set is fairly small (175MB) and you can always try it out on
+larger data sets:
```
philrz, line 51: @mccanne: I just took a shot at re-wording the `zar import` description to incorporate your suggestion and also reflect current reality as I know it. Could you give it a read-read? Thanks.

philrz: (@mccanne gave a :+1:  of this reword via Slack. Merging...)

PR #2416, opened 2021-03-24
move the Z parser from zq/zql to zq/compiler/parser

mccanne:
This commit moves the "zql" parser to a new package called "parser" under compiler and is the first step in deprecating the "zql" terminology.  This moves only the code and leaves the docs behind. as there is a separate effort to consolidate the various docs to `zq/docs`.  Also, we'll want to leave a copy of the old docs in `github.com/brimsec/zq/zql/docs` for old versions of the app to land on via the built-in docs link.


compiler/parser/Makefile
```diff
@@ -7,10 +7,12 @@ ROLLUP_VERSION = 2.23.1
...
 
-all: zql.go zql.js zql.es.js
+PARSER_PREFIX = parser
 
-.PHONY: zql.go
-zql.go:
+all: $(PARSER_PREFIX).go $(PARSER_PREFIX).js $(PARSER_PREFIX).es.js
+
+.PHONY: $(PARSER_PREFIX).go
+$(PARSER_PREFIX).go:
```
nwt, line 15: Nit: I think this file would read better if you dropped PARSER_PREFIX and instead did s/zql/parser/g, but if you prefer to keep it, call it parser_prefix and put it next to deps and npm since, unlike the other all-caps variables, it isn't really a knob.
```suggestion
all: parser.go parser.js parser.es.js

.PHONY: parser.go
parser.go:
```

Makefile
```diff
@@ -247,9 +247,10 @@ build-python-lib:
 clean-python:
 	@rm -rf python/brim/build
 
-PEG_GEN = zql/zql.go zql/zql.js zql/zql.es.js
-$(PEG_GEN): zql/Makefile zql/parser-support.js zql/zql.peg
-	$(MAKE) -C zql
+PARSER_GEN = compiler/parser/parser
+PEG_GEN = $(PARSER_GEN).go $(PARSER_GEN).js $(PARSER_GEN).es.js
```
nwt, line 251: Maybe shorten?
```suggestion
PEG_GEN := $(addprefix compiler/parser/parser., go js es.js)
```

nwt (approved):

PR #2430, opened 2021-03-26
improve canonical ZSON floating point format

### LINKED ISSUE #2408: cleaner ZSON canonical floating point
Implied floating point numbers should be cleaned up in the canonical ZSON output form.  For example, `1.1e+00` is ugly here:
```
echo '{x:1.1}' | zq -z -
{x:1.1e+00}
```
The design question here is for various scales of floating point numbers, what is the most user-friendly format that includes a "."  For integers, we already print `123.` but everything else has `e+` and that is ugly...

nwt:
Closes #2408.


mccanne (approved):
That change is deep!

Seriously, I looked at g a long time ago and thought "no that doesn't work for ints", then added the int hack, and now g is perfect.  Thanks!

PR #2432, opened 2021-03-26
fix locking bug in zson context

mccanne:
This commit fixes a bug where unlock was called on an unlocked lock.

This was triggered when a typedef name was accessed that did
not yet exist.  We added a test to exercise this path with an
issue() function that looks for a type that is not defuned until
the second record in the input.

### DIFF
--- a/zson/context.go
+++ b/zson/context.go
@@ -254,11 +254,10 @@ func (c *Context) LookupByName(zson string) (zng.Type, error) {
 	// each component of a nested type.
 	c.mu.Unlock()
 	typ, err := ParseType(c, zson)
+	c.mu.Lock()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


zson/context.go
```diff
@@ -255,6 +255,7 @@ func (c *Context) LookupByName(zson string) (zng.Type, error) {
 	c.mu.Unlock()
 	typ, err := ParseType(c, zson)
 	if err != nil {
+		c.mu.Lock()
 		return nil, err
 	}
 	c.mu.Lock()
```
nwt, line 261:
```suggestion
	c.mu.Lock()
	if err != nil {
		return nil, err
	}
```

nwt (approved):

PR #2434, opened 2021-03-26
Add zbuf.Reader post records method

mattnibs:
Add a method for client.Connection that allows users to pass a stream of
records to be ingested into zqd. These changes are needed for the load command
in the upcoming brimcap repo.

Add zngio.IOReader that uses a zngio.Writer to transform a zbuf.Reader into an
io.Reader.


mattnibs:
After pondering this a little further, and getting some useful feedback from @mccanne,  I am now seeing this as unnecessary and dropping this pr.

PR #2435, opened 2021-03-26
zapi post: accept '-' (stdin) as an argument

mattnibs:


api/client/multipartwriter.go
```diff
@@ -42,11 +42,10 @@ func MultipartFileWriter(files ...string) (*MultipartWriter, error) {
...
 			return nil, err
 		}
-		info, err := iosrc.Stat(context.Background(), u)
-		if err != nil {
-			return nil, err
+		info, _ := iosrc.Stat(context.Background(), u)
+		if info != nil {
+			m.BytesTotal += info.Size()
 		}
-		m.BytesTotal += info.Size()
```
nwt, line 49: Maybe continue to return an error if a file doesn't exist?

```go
        if f != iosrc.StdIn {
            info, err := iosrc.Stat(context.Background(), u)
            if err != nil {
                return nil, err
            }
            m.BytesTotal += info.Size()
        }
```

mattnibs: Good point

nwt (approved):

Issue #1762, opened 2020-12-06
improvements to zbuf.Batch and zng.Record memory


mccanne:
zbuf.Batch was designed a long time ago with the original goal of hiding in-place recording filtering behind an interface.  Since the code no longer works that way, we've been stuck with a suboptimal abstraction: arrays for *zng.Record pointers get created suboptimally and there is confusing memory-management across zbuf.Batch.Unref() and zng.Record.Keep().

The proposal here is to use reference-counted slices of []zng.Record which have records that point into reference-counted buffer slabs of zng Raw data (i.e., what a zng scanner decompresses into memory or what a zbuf.Batch implementation efficiently construct for entities like groupby that generats  lots of new records to be consumed downstream).

In this approach, `zbuf.Batch` becomes the interface to the slice of records and hidden behind that would be an implementation that manages and refcounts the slabs.

### New zbuf.Batch 

The current batch interface is:

```
type Batch interface {
	Ref()
	Unref()
	Index(int) *zng.Record
	Length() int
	Records() []*zng.Record
}
```

The proposed interface is:
```
type Batch interface {
    Ref()
    Unref()
    Records() []zng.Record 
}
```

Like old Batch, the semantics are that the record slice and the body of each record in the slice are immutable.  Under the covers, an implementation would typically manage and refcnt both the []zng.Record slices as well as the slabs of data that the records in the record slices point to.

Here indexing the slice of records can be done directly so there is no need for `Batch.Index()` and `Batch.Length()`.

### Example wrapper  implementation

Also, an implementation of batch could be stacked. This would be useful for filtering, e.g., where records aren't change just subsets of the slices are selected.  The outer batch could keep a reference on the inner batch but manage its own pool of record slices.  When the outer batch refcnt hits 0, it calls unref on the underlying batch.  And it's use is something like this in proc filter:

```
func (f *Proc) Pull() (zbuf.Batch, error) {
	for {
		batch, err := f.parent.Pull()
		if proc.EOS(batch, err) {
			return nil, err
		}
		// Create a new batch containing matching records.
		out := zbuf.NewWrapper(batch)
		for _, rec := range batch.Records() {
			if f.Filter(rec) {
				out.Append(rec)
			}
		}
		return out, nil
	}
}
```

compare to the old code:

```
func (f *Proc) Pull() (zbuf.Batch, error) {
	for {
		batch, err := f.parent.Pull()
		if proc.EOS(batch, err) {
			return nil, err
		}
[...]

PR #2437, opened 2021-03-26
clean up zng.Record

### LINKED ISSUE #1801: zng.Record type field should be zng.Type

mccanne:
This commit cleans up zng.Record by recognizing that a zng.Record
is a special case of a zng.Value.  It replaces the type pointers and
Raw field with an embedded zng.Value.  This is the first step toward
improving the batch model as described in issue #1762.

Closes #1801

### DIFF
--- a/zng/builder.go
+++ b/zng/builder.go
@@ -44,9 +44,7 @@ func (b *Builder) Build(zvs ...zcode.Bytes) *Record {
 	// Note that t.rec.nonvolatile is false so anything downstream
 	// will have to copy the record and we can re-use the record value
 	// between subsequent calls.
-	b.rec.Type = b.Type
-	b.rec.Alias = b.Type
-	b.rec.Raw = b.Bytes()
+	b.rec.Value = Value{b.Type, b.Bytes()}
 	return &b.rec
 }
 
--- a/zng/builder.go
+++ b/zng/builder.go
@@ -20,7 +20,7 @@ func NewBuilder(typ *TypeRecord) *Builder {
 	return &Builder{Type: typ}
 }
 
-// Build encodes the top-level zcode.Bytes values as the Raw field
+// Build encodes the top-level zcode.Bytes values as the Bytes field
 // of a record and sets that field and the Type field of the passed-in record.
 // XXX This currently only works for zvals that are properly formatted for
 // the top-level scan of the record, e.g., if a field is record[id:[record:[orig_h:ip]]
--- a/zng/recordval.go
+++ b/zng/recordval.go
@@ -40,50 +40,21 @@ func (r *RecordTypeError) Unwrap() error { return r.Err }
 // operations on zeek data can work with either serialized data or native
 // zng.Records by accessing data via the Record methods.
 type Record struct {
-	Type        *TypeRecord
-	Alias       Type
+	Value
 	nonvolatile bool
-	// Raw is the serialization format for records.  A raw value comprises a
-	// sequence of zvals, one per descriptor column.  The descriptor is stored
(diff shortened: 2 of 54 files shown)


zng/recordval.go
```diff
@@ -40,89 +40,50 @@ func (r *RecordTypeError) Unwrap() error { return r.Err }
...
-		Type:        AliasOf(typ).(*TypeRecord),
-		Alias:       typ,
-		nonvolatile: true,
-		Raw:         raw,
-	}
-}
-
-func NewRecordCheck(typ *TypeRecord, raw zcode.Bytes) (*Record, error) {
+func NewRecordCheck(typ Type, raw zcode.Bytes) (*Record, error) {
 	r := NewRecord(typ, raw)
```
nwt, line 57:
```suggestion
func NewRecordCheck(typ Type, bytets zcode.Bytes) (*Record, error) {
	r := NewRecord(typ, bytes)
```

zng/builder.go
```diff
@@ -44,9 +44,7 @@ func (b *Builder) Build(zvs ...zcode.Bytes) *Record {
 	// Note that t.rec.nonvolatile is false so anything downstream
 	// will have to copy the record and we can re-use the record value
 	// between subsequent calls.
-	b.rec.Type = b.Type
-	b.rec.Alias = b.Type
-	b.rec.Raw = b.Bytes()
+	b.rec.Value = Value{b.Type, b.Bytes()}
```
nwt, line 47: s/Raw/Bytes/ in this function's doc comment, too.

zng/recordval.go
```diff
@@ -131,43 +92,35 @@ func (r *Record) Keep() *Record {
...
 }
 
 // Walk traverses a record in depth-first order, calling a
 // RecordVisitor on the way.
 func (r *Record) Walk(rv Visitor) error {
-	return walkRecord(r.Type, r.Raw, rv)
+	return walkRecord(TypeRecordOf(r.Type), r.Bytes, rv)
 }
 
 // TypeCheck checks that the value coding in Raw is structurally consistent
```
nwt, line 126:
```suggestion
// TypeCheck checks that the value coding in Bytes is structurally consistent
```

nwt (approved):

PR #2440, opened 2021-03-26
rename master branch to main

nwt:


mattnibs:
Should we run brim tests on this to make sure this doesn't break anything there?

nwt:
@mattnibs: brimsec/brim shouldn't notice the change, but if it does, I'll fix it.

mattnibs (approved):

PR #2445, opened 2021-03-27
eliminate and remove proctest

### LINKED ISSUE #2436: remove package proctest
cut_test.go, sort_test.go, and compiler_test.go (in a very small way) are the only tests that use package proctest and sort_test already uses ztest for one of its test.  We should port these ztest and delete proctest.

mccanne:
Closes #2436


mattnibs (approved):

proc/proc.go
```diff
@@ -45,3 +45,13 @@ type Context struct {
...
 	return batch == nil || err != nil
 }
+
+type done struct {
+	zbuf.Puller
+}
+
+func (*done) Done() {}
+
+func NopDone(puller zbuf.Puller) *done {
```
nwt, line 55: Nit: This is more interesting than `type done`, so put it first.

PR #2447, opened 2021-03-27
import charm

mccanne:
This commit imports the charm package from github.com/mccanne.
This way we'll be able to work on it and extend it as discussed.

We also consolidated the uses golang.org/x/crypto/ssh/terminal
with a wrapper package in pkg/terminal (which is also used by
charm to get the interactive terminal width for text formatting).

### DIFF
--- a/pkg/charm/charm.go
+++ b/pkg/charm/charm.go
@@ -1,0 +1,91 @@
+// Package charm is minimilast CLI framework inspired by cobra and urfave/cli.
+package charm
+
+import (
+	"errors"
+	"flag"
+	"fmt"
+	"strings"
+)
+
+var ErrNoRun = errors.New("no run method")
+
+type Constructor func(Command, *flag.FlagSet) (Command, error)
+
+type Command interface {
+	Run([]string) error
+}
+
+type Spec struct {
+	Name          string
+	Usage         string
+	Short         string
+	Long          string
+	New           Constructor
+	Hidden        bool
+	HiddenFlags   string
+	RedactedFlags string
+	Empty         *Spec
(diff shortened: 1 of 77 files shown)


nwt (approved):

pkg/terminal/terminal.go
```diff
@@ -0,0 +1,22 @@
+package terminal
+
+import (
+	"os"
+
+	"golang.org/x/crypto/ssh/terminal"
```
mattnibs, line 6: This package is deprecated, use `golang.org/x/term`

pkg/charm/charm.go
```diff
@@ -0,0 +1,91 @@
...
+
+type Spec struct {
+	Name          string
+	Usage         string
+	Short         string
+	Long          string
+	New           Constructor
+	Hidden        bool
+	HiddenFlags   string
+	RedactedFlags string
```
mattnibs, line 27: Perhaps a comment here for what Hidden vs Redacted flags mean? I was/am confused about the difference.

mattnibs (approved):
Sad to see this leave its home at github.com/mccanne/charm. Would be nice to have this as its own standalone repo at some point, but I get the desire to put it here so we can move faster.

pkg/charm/charm.go
mccanne: Let's move charm in "as is" then improve it.  Frankly I don't remember the difference between hidden and redacted right now :-)

pkg/terminal/terminal.go
mccanne: good idea

PR #2449, opened 2021-03-27
Add ToStdio functionality to emitter

mattnibs:
Have emitter.NewDir and emitter.NewFile return a new Emitter interface, which
is a zbuf.WriteCloser along with a function that returns true if the emitter is
writing to stdio (i.e. stdout or stderr). This can be used to write command
line functionality that changes depending on where zng records are being output
to.


mattnibs:
I added this because I wanted to make brimcap analyze write stats updates to stderr if records are not getting sent to stdout. I think this is useful information.

mattnibs:
Going another direction with this. Closing.

PR #2466, opened 2021-03-31
rename microindex command and move zed/microindex to zed/index

mccanne:
This commit renames the microindex command as "zed index" and moves zed/microindex package to zed/index.  Some of the index file tests changed because of offset changes due to the index file magic field changing from "microindex" to "zed_index".


cmd/zed/index/command.go
```diff
@@ -7,24 +7,27 @@ import (
...
 )
 
-var MicroIndex = &charm.Spec{
-	Name:  "microindex",
-	Usage: "microindex <command> [options] [arguments...]",
-	Short: "create and manipulate microindexes",
+var Cmd = &charm.Spec{
+	Name:  "index",
+	Usage: "index <command> [options] [arguments...]",
+	Short: "create and search zed indexes from zed streams or files",
```
nwt, line 13: This reads roughly because "from zed streams or files" applies to "create" but not "search".
```suggestion
	Short: "create and search zed indexes",
```

cmd/zed/main.go
```diff
@@ -37,6 +37,13 @@ import (
 	_ "github.com/brimdata/zed/cmd/zed/zst/inspect"
 	_ "github.com/brimdata/zed/cmd/zed/zst/read"
 	"github.com/brimdata/zed/pkg/charm"
+
```
nwt, line 40:
```suggestion
```

lake/index/writer.go
```diff
@@ -101,12 +101,12 @@ func (a *onceError) Load() error {
...
-	cutter     *expr.Cutter
-	fgr        zbuf.ReadCloser
-	keyType    zng.Type
-	microindex *microindex.Writer
-	wg         sync.WaitGroup
+	err     onceError
+	cutter  *expr.Cutter
+	fgr     zbuf.ReadCloser
+	keyType zng.Type
+	index   *index.Writer
```
nwt, line 108: Nit: Keep these sorted.
```suggestion
	index   *index.Writer
	keyType zng.Type
```

nwt (approved):

PR #2467, opened 2021-03-31
fix some broken or needlessly absolute links

nwt:


philrz (commented):
Just for background, the ones in the CHANGELOG being absolute is not accidental. After drafting these in the CHANGELOG.md, I always need to cut & paste them into the Release Notes on the GitHub Releases page when we're publishing the artifacts, and relative URLs don't seem to work there. It always was convenient to be able to cut & paste those links as-is rather than having to maintain a set of absolutes in one location and relative in the other. If we have a strong allergy to this I could put effort in the future into maintaining both representations, but it never felt like a a priority.

nwt:
@philrz: Thanks for the background. That seems like a fine reason to have absolute links in `CHANGELOG.md`.

My goal in using relative links is to avoid breakage that occurs when a PR removes a link target but this isn't caught by CI because the target remains available under [https://github.com/brimdata/zed/tree/main](https://github.com/brimdata/zed/tree/main) until the PR is merged into main.

We could accomplish the same thing by using the release branch or tag in place of the main branch in absolute URLs. If that sounds good, I'll make that change here.

philrz (approved):
@nwt and I spoke about the CHANGELOG and my plan going forward is to make sure I use relative links in the `CHANGELOG.md`. This will help users find current info if they find a feature via an old release note but want to use it in a more recent version of the app/language.

Issue #1482, opened 2020-10-13
change name of archive to lake


mccanne:
We agreed to change the name of package archive to lake and rename the `zar` command `zed lake` to better reflect that we're really building a log and analytics lake rather than a log archive.

alfred-landrum:
This first PR renamed the archive package to lake: zq#2053

philrz:
By walking through the steps in the [`zed lake` README](https://github.com/brimdata/zed/blob/main/docs/lake/README.md), I've verified with `zed` commit e113e89 that the operations formerly performed with the command called `zar` are now working via `zed lake`. I did find a couple glitches in the README that should be fixed when the changes in #2490 are merged.

Thanks @mccanne and @alfred-landrum!

PR #2490, opened 2021-04-02
Fix problems in "zed lake" readme

philrz:
While looking to verify #1482, I decided to do a run through the `zed lake` README in its current state and found a couple glitches. I know `zed lake` is due for some significant overhaul, but until then, I figure we might as well keep the README operating.

### DIFF
--- a/docs/lake/README.md
+++ b/docs/lake/README.md
@@ -20,7 +20,7 @@ indexing and searching zed data lakes.
   * [Test data](#test-data)
   * [Ingesting the data](#ingesting-the-data)
   * [Initializing the archive](#initializing-the-archive)
-  * [Counting is our "hello world"](#counting-is-our-hello-world)
+  * [Counting as "hello world"](#counting-as-hello-world)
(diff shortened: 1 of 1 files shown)


docs/lake/README.md
```diff
@@ -125,11 +125,10 @@ to either stdout, or to a new file in the chunk's directory.
 Here's an example using the same "count()" query as before:
 
 ```
-zed lake -f table map "count()"
+zed lake map -f text "count()"
```
philrz, line 128: I opted to fix this one by going to `-f text` because `-f table` gives the ugly output:

```
$ zed lake map -f table "count()"
COUNT
275745
COUNT
276326
COUNT
290789
COUNT
619218
```

Depending on if `zed lake map` lives on in some form in the reworked `zed lake`, perhaps we could write something to emphasize the side effects of the per-file processing you get with `map`. But since this is so early in the text, I figure we might as well keep it simple.

nwt (approved):

PR #2492, opened 2021-04-03
zed query: read inputs sequentially

nwt:
The zed query command merges its inputs by ascending ts field. This  
behavior makes limited sense for Zeek logs and none for general use.  
Remove the merge and read inputs sequentially instead.


mccanne (approved):
Not a big deal but I'm not a fan of the Multi* pattern.  Instead of zbuf.MultiReader how about zbuf.Concatenate, zbuf.ConcatReaders, or something like that.

nwt:
I'm with you on the Multi\* pattern. I'll change it to ConcatReader.

nwt:
@philrz: I'm concerned that the output changes under `docs/language/` render the examples less interesting. Do you think this is something we should address (perhaps by tacking on a `sort ts`)?

philrz:
@nwt: I'd thought the docs casualties from this change would be worse than they actually are. I just looked at them side-by-side and I'd say only one of them became significantly less clear because of the change, but even that's not a catastrophe. You're right to think of adding a `sort`, but part of my my goal with what's there now was to try to introduce concepts without requiring the reader to learn other things at the same time (e.g. a bare search should ideally just be a bare search), so I'm game to let it ride as you have it. Ultimately these examples are not long for this earth, since I'm due to start rewriting them with some generic data similar to @mccanne's fruit examples. In the meantime, the page [https://github.com/brimdata/zed/tree/main/zql/docs](https://github.com/brimdata/zed/tree/main/zql/docs) (which is what the Brim app `v0.24.0` help menu points at) links directly at to how the docs were as of that GA tag `v0.29.0`, so those users can still see the prior rev of the docs whose output will match the behavior of their version of Brim/`zq`.

philrz (approved):

PR #2495, opened 2021-04-03
zed: remove -E flag (display epoch timstamps)

nwt:
It isn't worth its weight.

This affects zed's api, index, lake, query, and zst subcommands.


philrz (approved):

philrz (approved):
I hereby click Approve not as endorsement of the Go code changes (not qualified to speak to that) but as an acknowledgment that @nwt and I spoke about this change and I agree that we can take it out for now, since I'm not aware of users depending on this flag. If someone is affected and speaks up, we can have a discussion later about how to perhaps bring back the presentation options in a better form.

mccanne (approved):

PR #2496, opened 2021-04-03
expr: fix cast to net

### LINKED ISSUE #2113: Implement cast to net
I know we have lots of open issues regarding casting at this point, but I didn't see one for this specific case, so piling on one more. Repro is with `zq` commit `bdbef8b2`.

```
$ echo '{"subnet": "192.168.1.0/24"}' | zq -t 'put subnet=subnet:net' -
rhs of assigment expression: cast to net not implemented
```

nwt:
expr.castToNet incorrectly returns a zng.TypeIP value. Return a  
zng.TypeNet value instead.

Closes #2113.


mccanne (approved):
oopsies

PR #2501, opened 2021-04-05
fix Zed capitalization

mccanne:


docs/lake/README.md
```diff
@@ -233,7 +233,7 @@ output if you repeat the commands.)
 
 ## Indexes
 
-A "zed index" is a zng index file that pertains to just one
+A `zed index` is a zng index file that pertains to just one
```
philrz, line 236:
```suggestion
A Zed index is a zng index file that pertains to just one
```

Nit: Since we're talking about the index construct here rather than the command that's executed at the CLI, normally I'd opt _not_ to do the fixed width thing here.

philrz (approved):

Issue #2473, opened 2021-04-01
Change ZSON methods to string methods


philrz:
After the rework that was done resulting in the closure of #1417, @mccanne indicates there's some remaining tech debt to change the ZSON methods to string methods, now that the string methods have been freed up. This is just renaming, as @mccanne says the code works fine the way it is.

mccanne:
meaining... zng.Type.ZSON() should be replaced with String() and ZSONOf() should be replaced with StringOf()

mccanne:
We actually might want to move all formatting to package ZSON.
```
zson.FormatValue(zng.Value)
zson.FormatType(zng.Type)
```

PR #2511, opened 2021-04-07
add zson.FormatType()

mccanne:
This commits add the FormatType() function that formats up a
zng.Type in its canonical form.  The ZSON() method on zng.Type
didn't quite work right because it would output multiple
embedded typedefs when only the first is needed.  We may want
to get rid of the ZSON() method at some point as is commented
in issue #2473.

### DIFF
--- a/zson/formatter.go
+++ b/zson/formatter.go
@@ -353,42 +353,42 @@ func (f *Formatter) formatTypeBody(typ zng.Type) error {
 }
 
 func (f *Formatter) formatTypeRecord(typ *zng.TypeRecord) {
-	var sep string
 	f.build("{")
-	for _, col := range typ.Columns {
-		f.build(sep)
+	for k, col := range typ.Columns {
+		if k > 0 {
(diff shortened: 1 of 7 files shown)


zson/formatter.go
```diff
@@ -422,3 +422,74 @@ func (t typemap) known(typ zng.Type) bool {
...
+			typedefs[typ] = name
+			b.WriteString("=(")
+			formatType(b, typedefs, t.Type)
+			b.WriteByte(')')
+		}
+	case *zng.TypeRecord:
+		b.WriteByte('{')
+		var sep string
+		for _, col := range t.Columns {
+			b.WriteString(sep)
```
nwt, line 453: Nit: I like this pattern for the separator here and for enums and unions, but what's here is just fine!
```go
		for k, col := range t.Columns {
			if k > 0 {
				b.WriteByte(',')
			}
```

nwt (approved):

PR #2521, opened 2021-04-10
change Id to ID everywhere

nwt:
ID is the Go convention.


mccanne (approved):
Thank you for leaving chunk.Chunk.Id and saving me a rebase.  The lake branch will use ID.

mattnibs (approved):
yay!

PR #2522, opened 2021-04-10
remove fuse from CSV output

### LINKED ISSUE #2402: Don't automatically fuse for CSV output
After some time working with the changes from #1908, consensus is that the "auto-fuse" behavior is undesirable. Per a team discussion on March 23, 2021, the `-csvfuse` option will be removed, a single pass will be made through the data when outputting as CSV, and if the output is halted due to the presence of multiple types, the user would have the option to append a `| fuse` to their Z and re-run.

Right now here's how the error message would look in that case (running `zq` commit 75e8906 with the [zq-sample-data](https://github.com/brimsec/zq-sample-data)):

```
$ zq -f csv -csvfuse=false *
_path,ts,peer,mem,pkts_proc,bytes_recv,pkts_dropped,pkts_link,pkt_lag,events_proc,events_queued,active_tcp_conns,active_udp_conns,active_icmp_conns,tcp_conns,udp_conns,icmp_conns,timers,active_timers,files,active_files,dns_requests,active_dns_requests,reassem_tcp_size,reassem_file_size,reassem_frag_size,reassem_unknown_size
stats,2018-03-24T17:15:20.600725Z,zeek,74,26,29375,,,,404,11,1,0,0,1,0,0,36,32,0,0,0,0,1528,0,0,0
csv output requires uniform records but different types encountered
```

If I could make a plea to whoever implements this issue: Could we add _some_ mention of the `fuse` processor in this output? Perhaps a parenthetical statement like `(consider using 'fuse')`?

nwt:
This reverts fef92f2 (#1908).

As requested in #2402, it also adds `(consider 'fuse')` to the error string, which now looks like this:
```
```

Closes #2402.


philrz (commented):
Yes, I find the error output acceptable. 👍

mccanne (approved):
Nice!

PR #2527, opened 2021-04-11
zio/zngio: remove extra copy in Writer.Write

nwt:


mccanne (approved):
Thanks for addressing the comment I put in zio/zngio/writer.go six months ago :-)

PR #2556, opened 2021-04-15
move scan plan logic out of package commit and into package lake

mccanne:
This commits moves the scan planning logic out of the commit
package and into lake.  It will get bigger over time and
deserve its own dedicated package (perhaps integrate with
generic optimizations) but for now, it's small and can live
inside of lake.

This cleans things up so "squash" will go in more easily.


lake/scanner.go
```diff
@@ -0,0 +1,92 @@
...
+			return ctx.Err()
+		}
+	}
+	return nil
+}
+
+// ScanPartitions partitions all segments in the snapshot overlapping
+// with the given span into non-overlapping partitions and sends the
+// partitions over the given channel sorted by the pool key in the
+// pool's order.
```
nwt, line 57:
```suggestion
// ScanPartitions partitions all segments in snap overlapping
// span into non-overlapping partitions, sorts them by pool key and order,
// and sends them to ch.
```

nwt (approved):

PR #2562, opened 2021-04-16
add squash

### LINKED ISSUE #2543: zed lake squash
There should be a `zed lake squash` command to combine commits in staging.  There should also be a squash option in `zed lake commit` where multiple commit tags from staging are squashed into a single new commit ID when committed to the journal.

mccanne:
This commit adds "zed lake squash" and also cleans up the snapshot
abstraction so delete works and patches to snapshots can be easily
chained together.

Closes #2543

### DIFF
--- a/lake/commit/snapshot.go
+++ b/lake/commit/snapshot.go
@@ -101,7 +101,7 @@ func PlayAction(w Writeable, action actions.Interface) error {
 	return nil
 }
 
-// Play "plays" a recorded transaction into a whiteable snapshot.
+// Play "plays" a recorded transaction into a writeable snapshot.
 func Play(w Writeable, txn *Transaction) error {
 	for _, a := range txn.Actions {
 		if err := PlayAction(w, a); err != nil {
--- a/lake/commit/snapshot.go
+++ b/lake/commit/snapshot.go
@@ -1,48 +1,112 @@
 package commit
 
 import (
+	"errors"
+
+	"github.com/brimdata/zed/lake/commit/actions"
 	"github.com/brimdata/zed/lake/journal"
 	"github.com/brimdata/zed/lake/segment"
 	"github.com/brimdata/zed/pkg/nano"
-	"github.com/brimdata/zed/zbuf"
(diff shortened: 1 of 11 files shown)


lake/commit/snapshot.go
```diff
@@ -1,48 +1,112 @@
...
 	}
 	return segments
 }
+
+type Segments []*segment.Reference
+
+func (s *Segments) Append(segments Segments) {
+	*s = append(*s, segments...)
+}
+
```
nwt, line 92: Is this type going to get more methods? If not, it should go away.

lake/commit/transaction.go
```diff
@@ -17,26 +17,26 @@ type Transaction struct {
...
-	return Transaction{
+func newTransaction(id ksuid.KSUID, capacity int) *Transaction {
+	return &Transaction{
 		ID:      id,
 		Actions: make([]actions.Interface, 0, capacity),
 	}
 }
 
-func NewCommitTxn(id ksuid.KSUID, date nano.Ts, author, message string, segments []segment.Reference) Transaction {
+func NewCommitTxn(id ksuid.KSUID, date nano.Ts, author, message string, segments []segment.Reference) *Transaction {
```
nwt, line 27: Most (all?) of the new code here uses `[]*segment.Reference`, and we should eventually use that instead of `[]segment.Reference` everywhere.

nwt (approved):

lake/commit/snapshot.go
mccanne: I think it will change when we put the spanshots into a sub-pool.  I created the type to abstract things so there will be fewer changes down the road when we make this change.

lake/commit/transaction.go
mccanne: Yes, I realized this and we should keep this mind as we continue to develop these pieces.

mccanne: This was partially a motivation for Segments.

mccanne: We can clean this up in a subsequent PR.

PR #2564, opened 2021-04-16
add "zed lake delete"

### LINKED ISSUE #2544: zed lake delete
Implement the `zed lake delete` command stubbed out in cmd/zed/lake/delete/command.go.
This involves adding delete actions to a commit in staging.
When the staged commit is committed, the data will disappear from HEAD.
We need to update snapshot logic to do deletes.

mccanne:
This commit adds the "zed lake delete" command along with new support
in the commit layer for handling delete actions and querying the
commit journal by commit ID.

Closes #2544

### DIFF
--- a/cmd/zed/lake/delete/command.go
+++ b/cmd/zed/lake/delete/command.go
@@ -78,6 +78,9 @@ func (c *Command) Run(args []string) error {
 		return errors.New("no data or commit tags specified")
 	}
 	ids, err := pool.LookupTags(ctx, tags)
+	if err != nil {
+		return err
+	}
 	commitID, err := pool.Delete(ctx, ids)
 	if err != nil {
 		return err
--- a/cmd/zed/lake/delete/command.go
+++ b/cmd/zed/lake/delete/command.go
@@ -3,9 +3,13 @@ package del
 import (
 	"errors"
 	"flag"
+	"fmt"
+	"os"
 
 	zedlake "github.com/brimdata/zed/cmd/zed/lake"
+	"github.com/brimdata/zed/lake/commit/actions"
 	"github.com/brimdata/zed/pkg/charm"
+	"github.com/brimdata/zed/pkg/signalctx"
 )
 
 var Delete = &charm.Spec{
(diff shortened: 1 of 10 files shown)


cmd/zed/lake/delete/command.go
```diff
@@ -57,5 +64,44 @@ func (c *Command) Run(args []string) error {
...
+		return err
+	}
+	tags, err := zedlake.ParseIDs(args)
+	if err != nil {
+		return err
+	}
+	if len(tags) == 0 {
+		return errors.New("no data or commit tags specified")
+	}
+	ids, err := pool.LookupTags(ctx, tags)
```
nwt, line 80:
```suggestion
	ids, err := pool.LookupTags(ctx, tags)
	if err != nil {
		return err
	}
```

lake/commit/log.go
```diff
@@ -120,3 +121,38 @@ func (l *Log) Snapshot(ctx context.Context, at journal.ID) (*Snapshot, error) {
...
+		if action.CommitID() == commit {
+			ok = true
+			PlayAction(snapshot, action)
+			continue
+		}
+		if del, ok := action.(*actions.Delete); ok && snapshot.Exists(del.ID) {
+			PlayAction(snapshot, action)
+		}
+	}
+	return snapshot, ok, nil
```
nwt, line 157:
```suggestion
			return snapshot, ok, nil
		}
		if action.CommitID() == commit {
			ok = true
			PlayAction(snapshot, action)
			continue
		}
		if del, ok := action.(*actions.Delete); ok && snapshot.Exists(del.ID) {
			PlayAction(snapshot, action)
		}
	}
```

lake/commit/snapshot.go
```diff
@@ -84,6 +84,14 @@ func (s *Snapshot) Select(span nano.Span) Segments {
 	return segments
 }
 
+func (s *Snapshot) SelectAll() Segments {
+	var segments Segments
+	for _, seg := range s.segments {
+		segments = append(segments, seg)
```
nwt, line 90:
```suggestion
		segments.Append(seg)
```

lake/pool.go
```diff
@@ -111,6 +111,16 @@ func (p *Pool) Add(ctx context.Context, zctx *zson.Context, r zbuf.Reader) (ksui
 	return id, nil
 }
 
+func (p *Pool) Delete(ctx context.Context, ids []ksuid.KSUID) (ksuid.KSUID, error) {
+	id := ksuid.New()
+	// IDs aren't vetted here and will fail at commit time if prolematic.
```
nwt, line 116:
```suggestion
	// IDs aren't vetted here and will fail at commit time if problematic.
```

nwt (approved):

lake/commit/snapshot.go
mccanne: That doesn't work. Maybe we should get rid of Segments as you suggested.

PR #2566, opened 2021-04-16
add "-at" option to "zed lake query"

### LINKED ISSUE #2565: time travel
Add an `at` flag to `zed lake query` which takes a journal ID or commit tag (and automatically figures out which is which) and queries the pool at that point in the commit journal.

mccanne:
This commit enables time travel by allowing a "zed lake query"
to run at any point in the commit journal.

Closes #2565

### DIFF
--- a/cmd/zed/lake/query/command.go
+++ b/cmd/zed/lake/query/command.go
@@ -4,6 +4,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"strconv"
 	"text/tabwriter"
 
 	"github.com/brimdata/zed/cli/outputflags"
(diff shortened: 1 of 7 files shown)


cmd/zed/lake/query/command.go
```diff
@@ -74,7 +78,30 @@ func (c *Command) Run(args []string) error {
 	if err != nil {
 		return err
 	}
-	msrc := lake.NewMultiSource(pool)
+	var id journal.ID
+	if c.at != "" {
+		num, err := strconv.Atoi(c.at)
+		if err == nil {
```
nwt, line 84:
```suggestion
		if num, err := strconv.Atoi(c.at); err == nil {
```

nwt (approved):

Issue #2575, opened 2021-04-18
typedef variations and marshaling slice of interfaces


mccanne:
The `zed lake status` command (re-)exposed a bug (we knew about) where arrays of mixed-type values are not handled by zng.Marshal, in particular, `MarshalZNGContext.encodeArray()` uses the last value of the slice elements as the array type, which does not work for mixed-concrete-type interface values. 

The fix is to code mixed-type arrays as an array of unions. Buf if such a slice is typedef'd (as it is with `commit.Transaction.Actions`) then we need to create a typedef for the array-of-union.  However different instances of such a value can have different mixtures of concrete types as elements resulting in different array-of-union types.  This in turn causes any containing typedef to also be different.

We could just let this happen and rely upon Zed typedef renaming, which is perfectly legal.  However a stronger approach would be to add typedef variations to the Zed data model, which allows a single name to refer to multiple variations, e.g., Transaction.1, Transaction.2, etc.  While the marshaling library will not what to do with .1, .2, etc.,  the runtime should deal with it too, perhaps with operators that handle what is normally a rare condition for non-marshaled lake data. 

The zng marshaler context can bookkeep these variations as encountered and take a type-union of the union variations to bound the variations (also, with the union normalization #2145, this should all work out quite well).

mccanne:
NOTE: I worked around this bug in `zed lake status` by scanning the staging area differently and not sending a `commit.Transaction` into the mix.  Instead I added an actions.StagedCommit, which looks like a CommitMessage without the info, and scan this the same way the journal is laid out.

mccanne:
Reminder to see TestBug2575() in marshal_test.go for a test case.

PR #2577, opened 2021-04-18
improve ergonomics of zed lake commands

mccanne:
Now that the dust is settling on the zed lake command set, this
commit unifies the command sets and arranges for all of the
lake-introspection commands (ls, log, status) to produce output
in ZNG format that can then be pretty-printed with "-f lake.
This change is important for a few reasons.  First, it makes
it easy to view introspection results any way you want with zq.
Second, it means that the service API can use exactly the same
lake interface as the CLI tools and there is no need to create
a special API messages for every type of introspection since
it's all just ZNG.  If the API allows for a format (content-encoding?),
then we can be compatible with JSON API's by simply using "-f json".
Third, it is the right abstraction for handling large results from
introspection as the are simply streamed as a result instead
of being packaged into a bulky array that potentially does not
fit into memory.

This commit also adds some fit and finish like terminal colors
(now for the commit log, but we should add to -Z and other places)
a username heuristic from commit messages, and a time-travel "-at"
flag for zed lake ls.

We also commented out much of the worker logic in package driver
as it's awkward dependencies caused an import cycle with the addition
of lakeio.  This prototype anyway does not have the right abstraction
flexibility for an optimizing execution planner and it will be
reworked soon.

Finally, we hit the known-but-undocumented bug with marshaling
mixed-type arrays and filed issue #2575 though we worked around
it by changing the format of the staged-commit scan.


driver/parallel_test.go
```diff
@@ -103,6 +103,7 @@ func trim(s string) string {
 }
 
 func TestParallelOrder(t *testing.T) {
+	t.Skip() // XXX enable after multisource / parallel works fixed
```
nwt, line 106:
```suggestion
	t.Skip("enable after multisource / parallel works"
```

driver/parallel_test.go
```diff
@@ -174,6 +175,8 @@ func (m *scannerCloseMS) SourceFromRequest(context.Context, *api.WorkerChunkRequ
 // TestScannerClose verifies that any open ScannerCloser's will be closed soon
 // after the MultiRun call finishes.
 func TestScannerClose(t *testing.T) {
+	t.Skip() // XXX enable after multisource / parallel works fixed
```
nwt, line 178:
```suggestion
	t.Skip("enable after multisource / parallel works")
```

lake/pool.go
```diff
@@ -239,6 +240,49 @@ func (p *Pool) StoreInStaging(ctx context.Context, txn *commit.Transaction) erro
...
+				}
+			}
+			select {
+			case ch <- &actions.StagedCommit{Commit: id}:
+			case <-ctx.Done():
+				close(ch)
+				return
+			}
+		}
+		close(ch)
```
nwt, line 270:
```suggestion
		defer close(ch)
		for _, id := range ids {
			var txn *commit.Transaction
			txn, err = p.LoadFromStaging(ctx, id)
			if err != nil {
				return
			}
			for _, action := range txn.Actions {
				select {
				case ch <- action:
				case <-ctx.Done():
					return
				}
			}
			select {
			case ch <- &actions.StagedCommit{Commit: id}:
			case <-ctx.Done():
				return
			}
		}
```

zio/lakeio/writer.go
```diff
@@ -0,0 +1,209 @@
...
+	u := zson.NewZNGUnmarshaler()
+	u.Bind(
+		lake.PoolConfig{},
+		segment.Reference{},
+		lake.Partition{},
+		field.Static{},
+		actions.Add{},
+		actions.Delete{},
+		actions.CommitMessage{},
+		actions.StagedCommit{})
```
nwt, line 40: Move this to package scope and share across `Writer`s?

zio/lakeio/writer.go
```diff
@@ -0,0 +1,209 @@
...
+
+func (w *Writer) WriteZSON(rec *zng.Record) error {
+	s, err := w.zson.FormatRecord(rec)
+	if err != nil {
+		return err
+	}
+	if _, err := io.WriteString(w.writer, s); err != nil {
+		return err
+	}
+	_, err = w.writer.Write([]byte("\n"))
```
nwt, line 73:
```suggestion
	_, err = io.WriteString(w.writer, "\n")
```

zson/marshal_test.go
```diff
@@ -182,3 +185,48 @@ func TestBytes(t *testing.T) {
...
+	writer := zngio.NewWriter(&nopCloser{&buffer}, zngio.WriterOpts{})
+	recExpected := zng.NewRecord(zv.Type, zv.Bytes)
+	writer.Write(recExpected)
+	writer.Close()
+
+	r := bytes.NewReader(buffer.Bytes())
+	reader := zngio.NewReader(r, zson.NewContext())
+	recActual, err := reader.Read()
+	exp, _ := zson.FormatValue(recExpected.Value)
+	actual, _ := zson.FormatValue(recActual.Value)
```
nwt, line 230: NIt: Check errors.

zson/marshal_test.go
```diff
@@ -182,3 +185,48 @@ func TestBytes(t *testing.T) {
...
 
 }
+
+type nopCloser struct {
+	io.Writer
+}
+
+func (*nopCloser) Close() error {
+	return nil
+}
```
nwt, line 195: Use `zio.NopCloser`.
```suggestion
```

nwt (approved):

PR #2589, opened 2021-04-19
don't try to ZNG marshal unexported fields in structs

### LINKED ISSUE #2582: zng marshaler shouldn't panic when encountering private struct values
Michael on the brim slack channel [reported this problem](https://brimsec.slack.com/archives/CTT24051C/p1618804259032000). The zng marshaler should instead silently ignore unexported fields as is the design pattern in the Go's `encoding/json` package.

Repro with :

```
package main
import (
    "log"
    "os"
    "time"
    "github.com/brimsec/zq/zio/zngio"
    "github.com/brimsec/zq/zson"
)
type SomeType struct {
    ExportedS string
    ExportedT time.Time
    privateS  string
}
func main() {
    data := []SomeType{
        {
            ExportedS: "Sample",
            ExportedT: time.Now(),
            privateS:  "Not Public",
        },
        {
            ExportedS: "Sample 2",
            ExportedT: time.Now(),
            privateS:  "Not Public 2",
        },
    }
    // Create Marshaller
    marshaller := zson.NewMarshaler()
    // Create File
    f, err := os.Create("sample.zng")
    if err != nil {
        log.Println(err)
    }
    // Create Writer
    zw := zngio.NewWriter(f, zngio.WriterOpts{})
    defer zw.Close()
    // Add Data to file
    for _, event := range data {
        record, err := marshaller.MarshalRecord(event)
        if err != nil {
            log.Println(err)
        }
        if record == nil {
            break
        }
        err = zw.Write(record)
        if err != nil {
            log.Println(err)
        }
    }
}
```

mccanne:
The go reflect package panics when code tries to access
an unexported struct field.  This commit changes the logic
in the ZNG marshal to skip such fields just like Go's
encoding/json package does.

Closes #2582

### DIFF
--- a/zson/marshal.go
+++ b/zson/marshal.go
@@ -401,10 +401,21 @@ func (m *MarshalZNGContext) encodeRecord(sval reflect.Value) (zng.Type, error) {
 	var columns []zng.Column
 	stype := sval.Type()
 	for i := 0; i < stype.NumField(); i++ {
-		if !sval.Field(i).CanInterface() {
-			// If we can't access this field, silently ignore it
-			// like the json marshaler does.  This occurs when
-			// struct fields are not exported.
+		sf := stype.Field(i)
+		isUnexported := sf.PkgPath != ""
+		if sf.Anonymous {
+			t := sf.Type
+			if t.Kind() == reflect.Ptr {
+				t = t.Elem()
+			}
+			if isUnexported && t.Kind() != reflect.Struct {
+				// Ignore embedded fields of unexported non-struct types.
+				continue
+			}
+			// Do not ignore embedded fields of unexported struct types
+			// since they may have exported fields.
+		} else if isUnexported {
+			// Ignore unexported non-embedded fields.
 			continue
 		}
 		field := stype.Field(i)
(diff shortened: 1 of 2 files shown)


zson/marshal.go
```diff
@@ -401,6 +401,12 @@ func (m *MarshalZNGContext) encodeRecord(sval reflect.Value) (zng.Type, error) {
 	var columns []zng.Column
 	stype := sval.Type()
 	for i := 0; i < stype.NumField(); i++ {
+		if !sval.Field(i).CanInterface() {
+			// If we can't access this field, silently ignore it
+			// like the json marshaler does.  This occurs when
+			// struct fields are not exported.
+			continue
+		}
```
nwt, line 409: This is fine but I think we should do this: https://github.com/golang/go/blob/go1.16.3/src/encoding/json/encode.go#L1241-L1257.

nwt (approved):

MichaelRoberts-CyberSecurity:
# Thanks for solving the ZNG Marshalling private fields so rapidly.

PR #2593, opened 2021-04-20
gofmt -s -w .

nwt:


mattnibs (approved):

mattnibs:
Are we adding this to the makefile?

nwt:
@mattnibs: Think we should?

PR #2596, opened 2021-04-20
CLI -h for help and allow subcommands to use parent flags

### LINKED ISSUE #2557: "-h" should print help text consistently in Zed CLI tools
Right now the command `help` is used in all the Zed tools to print help text, e.g. `zed help` gives top-level help that describes all the top-level `zed` commands, and `zed help lake` provides the help text specifically for `zed lake`, and so forth.

As a group we discussed that when seeking the help text for a command it's a bit of a hassle to up-arrow in the shell and surgically add the full word `help` at the right spot in the middle (e.g. to turn `zed lake` into `zed help lake`), and that it would be convenient to be able to add `-h` as an option at any level to see the appropriate help text. Extending our example, `zed -h` would give the same output as `zed help`, and `zed lake -h` would give the same output as `zed lake help`.

A surprise that's waiting for whoever picks up this issue: I can see that `-h` is already used by `zed api` to specify alternate listening IP/port. Maybe we'll find other places where it's being used.

### LINKED ISSUE #2586: Allow CLI subcommands to use parent flags

mccanne:
closes #2557 

closes #2586


pkg/charm/charm.go
```diff
@@ -17,15 +18,19 @@ type Command interface {
...
+	Short string
+	Long  string
+	New   Constructor
+	// Hidden hides this command from help.
+	Hidden bool
+	// Hidden flags (comma-separated) markes these flags as hidden.
+	HiddenFlags string
+	// Redacted flags (comma-separated) marks these flags as redacted,
+	// where a flag is shown (if not hidden) but its default value is hidden,
+	// e.g., as is useful for a password flag.
```
mattnibs, line 32: Thanks for this!

cmd/zed/api/cli.go
```diff
@@ -41,123 +40,84 @@ With "zed api" you can create spaces, list spaces, post data to spaces, and run
...
-	}
-
+func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
+	c := &Command{Command: parent.(*root.Command)}
 	// If not a terminal make nofancy on by default.
 	c.NoFancy = !terminal.IsTerminalFile(os.Stdout)
-
 	defaultHost := "localhost:9867"
-	f.StringVar(&c.Host, "h", defaultHost, "<host[:port]>")
+	f.StringVar(&c.Host, "service", defaultHost, "<host[:port]>")
```
mattnibs, line 51: I don't want to make you change all those tests, but I think `-service` is a lot to type for such a common option. Perhaps `-srv`?

ppl/cmd/gentoken/gentoken.go
```diff
@@ -56,8 +56,7 @@ func (c *Command) Run(args []string) error {
 }
 
 func main() {
-	CLI.Add(charm.Help)
-	_, err := CLI.ExecRoot(os.Args[1:])
+	err := CLI.ExecRoot(os.Args[1:])
 	if err != nil {
```
nwt, line 60:
```suggestion
	if err := CLI.ExecRoot(os.Args[1:]); err != nil {
```

pkg/charm/path.go
```diff
@@ -0,0 +1,37 @@
+package charm
+
+import (
+	"fmt"
+	"strings"
+)
+
+type Path []*instance
```
nwt, line 8: NIt: This isn't part of the `pkg/charm` interface, so don't export it.
```suggestion
type path []*instance
```

pkg/charm/path.go
```diff
@@ -0,0 +1,37 @@
...
+	}
+	return err
+}
+
+func (c Path) last() *instance {
+	return c[len(c)-1]
+}
+
+func (c Path) pathname(args ...string) string {
+	names := make([]string, 0, len(c))
```
nwt, line 23: (But I still like `var name []string` in this situation for lower cognitive load.)
```suggestion
	names := make([]string, 0, len(c)+len(args))
```

pkg/charm/instance.go
```diff
@@ -83,6 +40,100 @@ func (i *instance) options(vflag bool) []string {
...
+		if _, ok := all[f.Name]; !ok {
+			all[f.Name] = f
+			difference[f.Name] = f
+		}
+	})
+	return difference
+}
+
+func parseHelp(spec *Spec, args []string) (Path, error) {
+	var path Path
```
nwt, line 105: NIt: Declare closer to use.

pkg/charm/instance.go
```diff
@@ -83,6 +40,100 @@ func (i *instance) options(vflag bool) []string {
...
+	flags := flag.NewFlagSet(spec.Name, flag.ContinueOnError)
+	var b bool
+	flags.BoolVar(&b, HelpFlag, false, "display help")
+	flags.BoolVar(&b, HiddenFlag, false, "show hidden options")
```
nwt, line 109: Nit: Using `BoolVar` for these results in help output like this:
```
    -h display help (default "false")
    -hidden show hidden options (default "false")
```
Switch to `Func` for this:
```
    -h display help
    -hidden show hidden options
```
```suggestion
	flags.Func(HelpFlag, "display help", func(_ string) error { return nil })
	flags.Func(HiddenFlag, "show hidden options", func(_ string) error { return nil })
```

mattnibs (approved):

nwt (approved):

cmd/zed/api/cli.go
mccanne: I'll do -host.  It's an easy find/sed one liner.

pkg/charm/path.go
mccanne: I prefer using one pattern for when you know the size of slices (for lower cognitive load) rather than switching between the patterns based on performance considerations.  This is the one Go-ism that I think you and I will continue to disagree on :-)

pkg/charm/instance.go
mccanne: That doesn't seem to work.

pkg/charm/path.go
nwt: And that's OK!

Issue #2598, opened 2021-04-20
Improve error message for clash between type name & field reference


philrz:
Repro is with `zq` commit f8d7a87 and the [Zeek NDJSON sample data](https://github.com/brimdata/zed-sample-data/tree/main/zeek-ndjson).

When attempting certain additional Zed operations after applying the [reference shaper for Zeek](https://github.com/brimdata/zed/blob/main/zeek/shaper.zed), we see failures with error messages that aren't too helpful.

```
$ zq -z -I ~/work/zed/zeek/shaper.zed '| head 1 | cut id.resp_h' weird.ndjson.gz 
invalid expression on lhs of assignment

$ zq -z -I ~/work/zed/zeek/shaper.zed '| head 1 | put classnet=network_of(id.resp_h) | count() by classnet | sort classnet' weird.ndjson.gz 
put: a referenced field is missing
```

The root cause turns out to be that there's currently a type definition called `id` within the shaper:

```
type id={orig_h:ip,orig_p:port,resp_h:ip,resp_p:port};
```

This clashes with these later attempts to treat `id` as a reference to a named field in the log data.

We discussed this as a group and there was consensus that the way to address in the short term is to use more descriptive names for the types in the shaper such that they would not conflict with named fields in the data, e.g. `type socket` in this case. However, even if we show best practices in our reference configs and docs, since certain data sources may have hundreds or thousands of fields of different names, we can't rule out that this might still happen in the wild. Therefore the error messages should be improved to indicate what actually happened, i.e. explicitly mention the clash.

mccanne:
This has been fixed as we now require type values to be wrapped in angle brackets...

Issue #2597, opened 2021-04-20
Sensitivity to type names that start with "net"


philrz:
Repro is with `zq` commit 96bc91e, run against the [NDJSON Zeek sample data](https://github.com/brimdata/zed-sample-data/tree/main/zeek-ndjson).

I was doing some variations to try to make the `type` names in the [reference shaper for Zeek](https://github.com/brimdata/zed/blob/main/zeek/shaper.zed) less likely to clash with field names that may appear in the data. Let's start from this simplification of that shaper in its current form, which works ok.

```
$ cat mini-shaper.zed 
type port=uint16;
type id={orig_h:ip,orig_p:port,resp_h:ip,resp_p:port};
put .=unflatten(.)

$ zq -z -I mini-shaper.zed '| head 1' ~/work/zq-sample-data/zeek-ndjson/conn.ndjson.gz 
{_path:"conn",_write_ts:"2018-03-24T17:15:21.400275Z",ts:"2018-03-24T17:15:21.255387Z",uid:"C8Tful1TvM3Zf5x8fl",id:{orig_h:"10.164.94.120",orig_p:39681,resp_h:"10.47.3.155",resp_p:3389},proto:"tcp",duration:0.004266023635864258,orig_bytes:97,resp_bytes:19,conn_state:"RSTR",missed_bytes:0,history:"ShADTdtr",orig_pkts:10,orig_ip_bytes:730,resp_pkts:6,resp_ip_bytes:342}
```

The problem showed up when I tried to use `networkport` in place of `port`.

```
$ cat networkport.zed 
type networkport=uint16;
type socket={orig_h:ip,orig_p:networkport,resp_h:ip,resp_p:networkport};
put .=unflatten(.)

$ zq -z -I networkport.zed '| head 1' ~/work/zq-sample-data/zeek-ndjson/conn.ndjson.gz 
zq: parse error: error parsing Z at line 2, col 72:type networkport=uint16;
type socket={orig_h:ip,orig_p:networkport,resp_h:ip,resp_p:networkport};
                                                                   === ^ ===
put .=unflatten(.)

| head 1
```

After scratching my head at this not-too-helpful parse error for a bit, I fiddled with variations until I deduced that it seems to be the leading `net` that's throwing it off. Example of a workaround:

```
$ cat ntworkport.zed 
type ntworkport=uint16;
type socket={orig_h:ip,orig_p:ntworkport,resp_h:ip,resp_p:ntworkport};
put .=unflatten(.)

[...]

Issue #2594, opened 2021-04-20
Update Zeek JSON typing README


philrz:
The doc at [https://github.com/brimdata/zed/blob/main/zeek/README.md](https://github.com/brimdata/zed/blob/main/zeek/README.md) still references the older `types.json` approach to restoring rich typing to NDJSON data. I plan to morph this doc into something that describes the shaping approach that's now used instead. As part of this, I'll advise that they might want to use some kind of shell alias to apply it all the time if they usually read Zeek NDJSON, though #1059 is expected to eventually improve on this.

philrz:
The functionality covered by the older ["JSON Type Definitions"](https://github.com/brimdata/zed/blob/v0.29.0/zeek/README.md) README is now covered in [https://github.com/brimdata/zed/blob/main/zeek/Shaping-Zeek-NDJSON.md](https://github.com/brimdata/zed/blob/main/zeek/Shaping-Zeek-NDJSON.md) There's also now a top-level [https://github.com/brimdata/zed/blob/main/zeek/README.md](https://github.com/brimdata/zed/blob/main/zeek/README.md) that links to three different Zeek-related docs in that directory, and that shaping doc is one of those three.

PR #2601, opened 2021-04-21
Improvements to reference Zeek shaper

philrz:
As a prerequisite to #2594, here I'm proposing some changes to the reference Zeek shaper to make it less hazardous by renaming the `type id` to `type socket`. This avoids clashes when fields are referenced in the Zeek data that contain the common embedded record also known as `id`. Since we're now aware of the potential for clashes, #2598 is open to track the improving of the error messaging to help make users aware when it happens.

While I was at it, I considered renaming `type port` as well, but my selection of the more descriptive `type networkport` led me to hit new bug #2597. At that point I got to looking closer at the Zeek data and confirmed that there were no actual fields called `port`, which is perhaps unsurprising since that happens to be one of their native types, so perhaps they also have their own possible clashes to avoid. But while I've left `type port` intact, I did notice that there seemed be to be unnecessarily verbose specifications of `:port=(uint16)` throughout the various type definitions when a simple reference to the already-defined `:port` would have sufficed, so I've simplified those here as well.


zeek/shaper.zed
```diff
@@ -1,57 +1,57 @@
 type port=uint16;
 type zenum=string;
-type id={orig_h:ip,orig_p:port,resp_h:ip,resp_p:port};
+type socket={orig_h:ip,orig_p:port,resp_h:ip,resp_p:port};
```
nwt, line 3: The Zeek type for this is [`conn_id`](https://docs.zeek.org/en/lts/scripts/base/init-bare.zeek.html#type-conn_id). I think we should use that instead of our own name.

mccanne: (I don't think we made up the type `socket` but technically this is a `socketpair`.  Another term would be network `flow` though that usually include IP protocol.  That all said, I prefer the `conn_id` approach.)

nwt: True, and I apologize for any implied shade!

"Our own name" was a poor choice of words. "A different name" is what I meant.

philrz: Yes `conn_id` crossed my mind as well, so I've pushed a change to that.

nwt (approved):

PR #2603, opened 2021-04-21
remove python/brim/

nwt:
It is unused.


mattnibs (approved):

nwt:
Discussed this with @mccanne and—as much as I love deleting things—we agreed to keep this around for possible revival.

PR #2613, opened 2021-04-22
First cut of lake index

### LINKED ISSUE #2532: new zed lake: make indexing work again
The indexing logic needs to be updated for the new zed lake layout.

The index writer has been removed from the segment data write-path as it is typically memory intensive and should be done in a separate pass tuned for the indexing workload.  It will be run from orchestration external to the write path and may very well be run by agents outside of the core that simply run indexing work, introspection, and analysis through the service API.  Different indexing events may happen at different times and could even be triggered off reactive algorithms that adapt to the query workload patterns.

This task involves updating the indexing core (not the orchestration) and `zed lake index` commands.  Indexes will be named with ksuid's (but the ksuid does not appear in the object path of the zng index file).  The commit log holds the index rules and index add/del ops for each segment (which refer to the ksuid of the rule).

_(See comment below for modified proposal.)_ 

Proposal is:
* assign a ksuid to each index rule, call it `<xid>`
* put `<xid>`'s rule in the commit log as log-oriented ref
* index add are just like add commits and can be intermixed in a single commit
* add index-add (`<seg>`,`<xid>`) and index-delete actions to commit actions
* store an index of data segment `<segid>` as immutable object `<poolid>/index/<segid>.zng`

mattnibs:
Add ability to create/drop an index rule, add ability to apply an index rule to
a tag(s).

Closes #2532


ztest/shell.go
```diff
@@ -31,5 +31,7 @@ func RunShell(dir *Dir, bindir, script string, stdin io.Reader, useenvs []string
 	cmd.Stdout = &stdout
 	cmd.Stderr = &stderr
 	err := cmd.Run()
+	fmt.Println("stdout", stdout.String())
+	fmt.Println("stderr", stderr.String())
```
nwt, line 35:
```suggestion
```

lake/commit/actions/actions.go
```diff
@@ -71,3 +73,16 @@ func (d *Delete) CommitID() ksuid.KSUID {
 func (d *Delete) String() string {
 	return "DEL " + d.ID.String()
 }
+
+type AddX struct {
```
nwt, line 77: I think AddIndex would be a clearer name.

lake/index/combiner.go
```diff
@@ -0,0 +1,60 @@
...
+
+func (ws Combiner) References() []Reference {
+	references := make([]Reference, len(ws))
+	for i, w := range ws {
+		references[i] = w.Reference
+	}
+	return references
+}
+
+func (ws Combiner) Abort() {
```
nwt, line 56: `c` seems like a better receiver name.

cmd/zed/lake/index/add.go
```diff
@@ -0,0 +1,119 @@
...
+	lake   *zedlake.Command
+	commit bool
+	rids   idArray
+	zedlake.CommitFlags
+}
+
+func NewAdd(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
+	c := &AddCommand{lake: parent.(*Command).Command}
+	f.BoolVar(&c.commit, "commit", false, "commit added index rule if successfully written")
+	f.Var(&c.rids, "rule", "id of rule(s) to index (can be specified multiple times")
```
nwt, line 31: Maybe simplify by dropping `idArray`, changing the type of `AddCommand.rid` to `[]ksuid.KSUID` and using `flag.FlagSet.Func`?
```suggestion
	f.Func("rule", "id of rule to index (can be specified multiple times", func(s string error {
		id, err := zedlake.ParseID(s)
		c.rids = append(c.rids, id)
		return err
	})
```

cmd/zed/lake/index/add.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	if !c.lake.Quiet {
+		fmt.Printf("%s staged\n", commit)
+	}
+
+	return nil
+}
+
+type idArray []ksuid.KSUID
```
nwt, line 98: Nit: If you keep this, since its a slice, not an array, call it `idsArg` or `ids` or `idSlice`.

cmd/zed/lake/index/command.go
```diff
@@ -10,12 +10,13 @@ import (
...
 	Name:  "index",
 	Usage: "index [subcommand]",
-	Short: "perform index related tasks on an archive",
+	Short: "create and drop index rules, index data",
 	New:   New,
 }
 
 func init() {
 	Index.Add(Create)
+	Index.Add(Add)
```
nwt, line 19:
```suggestion
	Index.Add(Add)
	Index.Add(Create)
```

cmd/zed/lake/index/add.go
```diff
@@ -0,0 +1,119 @@
...
+
+	zedlake "github.com/brimdata/zed/cmd/zed/lake"
+	"github.com/brimdata/zed/pkg/charm"
+	"github.com/brimdata/zed/pkg/rlimit"
+	"github.com/segmentio/ksuid"
+)
+
+var Add = &charm.Spec{
+	Name:  "add",
+	Usage: "add [options] [ -rule ruleid ] tag [ ...tag ]",
```
nwt, line 16: Follow the lead in `cmd/zed/lake/commit/command.go` and `cmd/zed/lake/squash/command.go`.
```suggestion
	Usage: "add [options] [ -rule ruleid ] tag [tag ...]",
```

cmd/zed/lake/index/drop.go
```diff
@@ -3,49 +3,66 @@ package index
...
-
-"zar index drop" removes an index definition from the archive then walks through
-the tree removing referenced index files.
-
-If the -noapply option is enabled the command will only removed the index
-definition. The individual index files will still exist but they will no longer
-be queryable.
-`,
-	New: NewDrop,
+	Short: "drop index rule(s) from a lake",
```
nwt, line 16:
```suggestion
	Short: "drop index rules from a lake",
```

zio/lakeio/writer.go
```diff
@@ -195,3 +201,43 @@ func formatDelete(b *bytes.Buffer, indent int, delete *actions.Delete) {
...
+		b.WriteString("field ")
+		b.WriteString(xrule.Value)
+	case index.RuleZed:
+		keys := make([]string, len(xrule.Keys))
+		for i, k := range xrule.Keys {
+			keys[i] = k.String()
+		}
+
+		b.WriteString("field(s) ")
+		b.WriteString(strings.Join(keys, ", "))
```
nwt, line 237:
```suggestion
		b.WriteString("field(s) ")
		for i, k := range xrule.Keys {
			if i > 0 {
				b.WriteString(", ")
			}
		}	b.WriteString(k.String())
```

lake/pool.go
```diff
@@ -68,6 +72,9 @@ func (p *PoolConfig) Create(ctx context.Context, root iosrc.URI) error {
 	if err := iosrc.MkdirAll(StagePath(path), 0700); err != nil {
 		return err
 	}
+	if err := iosrc.MkdirAll(IndexPath(path), 0700); err != nil {
+		return err
+	}
```
nwt, line 77:
```suggestion
	if err := iosrc.MkdirAll(IndexPath(path), 0700); err != nil {
		return err
	}
	if err := iosrc.MkdirAll(StagePath(path), 0700); err != nil {
		return err
	}
```

lake/index/writer.go
```diff
@@ -35,8 +36,8 @@ func NewWriter(ctx context.Context, u iosrc.URI, def *Definition) (*Writer, erro
 }
 
 type Writer struct {
-	Definition *Definition
-	URI        iosrc.URI
+	Reference Reference
```
nwt, line 39: `Reference` is big. Use a pointer.
```suggestion
	Reference *Reference
```

lake/index/writer.go
```diff
@@ -109,36 +110,64 @@ type indexer struct {
 	wg      sync.WaitGroup
 }
 
-func newIndexer(ctx context.Context, u iosrc.URI, def *Definition, r zbuf.Reader) (*indexer, error) {
+func newIndexer(ctx context.Context, path iosrc.URI, ref Reference, r zbuf.Reader) (*indexer, error) {
```
nwt, line 113: `Reference` is big. Use a pointer.
```suggestion
func newIndexer(ctx context.Context, path iosrc.URI, ref *Reference, r zbuf.Reader) (*indexer, error) {
```

lake/index/writer.go
```diff
@@ -14,18 +14,19 @@ import (
 	"github.com/brimdata/zed/pkg/iosrc"
 	"github.com/brimdata/zed/zbuf"
 	"github.com/brimdata/zed/zng"
+	"github.com/brimdata/zed/zqe"
 	"github.com/brimdata/zed/zson"
 )
 
-func NewWriter(ctx context.Context, u iosrc.URI, def *Definition) (*Writer, error) {
+func NewWriter(ctx context.Context, path iosrc.URI, ref Reference) (*Writer, error) {
```
nwt, line 21: `Reference` is big. Use a pointer.
```suggestion
func NewWriter(ctx context.Context, path iosrc.URI, ref *Reference) (*Writer, error) {
```

lake/index/writer.go
```diff
@@ -109,36 +110,64 @@ type indexer struct {
...
+				if err = iosrc.MkdirAll(ref.ObjectDir(path), 0700); err == nil {
+					tried = true
+					continue
+				}
+			}
+		}
+		return w, err
 	}
-	return d, nil
 }
```
nwt, line 171: Simplify.
```go
func newIndexWriter(ctx context.Context, zctx *zson.Context, path string, opts ...index.Option) (*index.Writer, error) {
	w, err := index.NewWriterWithContext(ctx, zctx, path, opts...)
	if zqe.IsNotFound(err) {
		if err := iosrc.MkdirAll(ref.ObjectDir(path), 0700); err != nil {
			return nil, err
		}
		return index.NewWriterWithContext(ctx, zctx, path, opts...)
	}
	return w, err
}
```

lake/commit/actions/actions.go
mccanne: My bad.  I agree.

lake/index/reference.go
```diff
@@ -0,0 +1,94 @@
...
+func (r Reference) String() string {
+	return fmt.Sprintf("%s/%s", r.Rule.ID, r.SegmentID)
+}
+
+func (r Reference) ObjectName() string {
+	return ObjectName(r.SegmentID)
+}
+
+func ObjectName(id ksuid.KSUID) string {
+	return fmt.Sprintf("idx-%s.zng", id)
```
mccanne, line 24: no need for `-idx`

cmd/zed/lake/index/create.go
```diff
@@ -37,136 +43,96 @@ requires specifying the key and output file name. For example:
...
 			return err
 		}
 
-		if c.noapply {
-			return nil
-		}
-
-		if c.ensure {
-			defs, err = lk.ReadDefinitions(ctx)
+		m := zson.NewZNGMarshaler()
```
mccanne, line 102: This should use root.ScanXRules.  There shouldn't be a need to have a marshaling loop like this at the command level.

cmd/zed/lake/index/add.go
```diff
@@ -0,0 +1,119 @@
...
+	ctx, cleanup, err := c.lake.Init()
+	if err != nil {
+		return err
+	}
+	defer cleanup()
+
+	if _, err := rlimit.RaiseOpenFilesLimit(); err != nil {
+		return err
+	}
+
```
mccanne, line 46: We had agreed on a coding convention to not put blank lines like this but add a comment explaining the need for the separation if warranted.

cmd/zed/lake/index/drop.go
```diff
@@ -3,49 +3,66 @@ package index
...
+	if err != nil {
+		return err
+	}
+
+	root, err := c.lake.Open(ctx)
+	if err != nil {
+		return err
+	}
+
+	xrules, err := root.DeleteXRules(ctx, ids)
```
mccanne, line 54: In thinking about names here, I realized we should call a "rule" an "index" and "drop" drops an entire index (as is coded here).  "delete" would still work to delete individual objects from an index.  The "delete" command will should determine when a segment being deleted is actually a data object or a index object.  This is known from the snapshot.  I see there are no updates for delete on this PR.  We can address this part later.

mccanne (changes requested):

cmd/zed/lake/index/add.go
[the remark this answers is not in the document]
mattnibs: ok

lake/index/rule.go
```diff
@@ -7,35 +7,33 @@ import (
...
 	RuleField RuleKind = "field"
-	RuleZQL   RuleKind = "zql"
+	RuleZed   RuleKind = "zed"
 )
 
 // Rule contains the runtime configuration for an indexing rule.
 type Rule struct {
+	Framesize int            `zng:"framesize,omitempty"`
+	ID        ksuid.KSUID    `zng:"id"`
+	Name      string         `zng:"name,omitempty"`
```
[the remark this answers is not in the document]
mattnibs: Why don't I just make the name required in this pr for all rules.

lake/index/reference.go
mattnibs: Yep

cmd/zed/lake/index/add.go
mattnibs: I was not aware of this, will fix.

cmd/zed/lake/index/drop.go
mattnibs: I was going to say we could just use the regular delete command to delete index objects, but am realizing that delete works on both commits and segments so it wouldn't work.

lake/index/rule.go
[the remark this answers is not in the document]
mattnibs: Actually, if its all good I'll do this in the follow up pr when I make Indices (formerly Rules) all zed queries.

cmd/zed/lake/index/create.go
```diff
@@ -37,136 +43,96 @@ requires specifying the key and output file name. For example:
...
+	}
 
-		ctx, cancel := signalctx.New(os.Interrupt)
-		defer cancel()
+	root, err := c.lake.Open(ctx)
+	if err != nil {
+		return err
+	}
+
+	rules, err := c.createXRules(ctx, root, args)
```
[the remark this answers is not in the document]
mattnibs: I deleted the long description and will write a new one when I transform Indices into zed scripts.

cmd/zed/lake/index/add.go
```diff
@@ -0,0 +1,89 @@
...
+	zedlake "github.com/brimdata/zed/cmd/zed/lake"
+	"github.com/brimdata/zed/pkg/charm"
+	"github.com/brimdata/zed/pkg/rlimit"
+	"github.com/segmentio/ksuid"
+)
+
+var Add = &charm.Spec{
+	Name:  "add",
+	Usage: "add [options] [ -index indexid ] tag [tag ...]",
```
nwt, line 16:
```suggestion
	Usage: "add [options] [-index indexid] tag [tag ...]",
```

cmd/zed/lake/index/create.go
```diff
@@ -1,172 +1,100 @@
...
-	f.StringVar(&c.inputFile, "i", "_", "input file relative to each zar directory ('_' means archive log file in the parent of the zar directory)")
-	f.BoolVar(&c.noapply, "noapply", false, "create index rules but do not apply them to the archive")
-	f.StringVar(&c.outputFile, "o", "index.zng", "name of microindex output file (for custom indexes)")
-	//f.BoolVar(&c.displayer.quiet, "q", false, "don't print progress on stdout")
-	f.StringVar(&c.zql, "z", "", "zql for custom indexes")
+	c := &CreateCommand{lake: parent.(*Command).Command}
+	f.IntVar(&c.framesize, "framesize", 32*1024, "minimum frame size used in microindex file")
+	f.StringVar(&c.keys, "k", "key", "one or more comma-separated key fields (for zed script indices only)")
+	f.StringVar(&c.name, "n", "", "name of index (for zed script indices only)")
+	f.StringVar(&c.zed, "zed", "", "zed script for index")
```
nwt, line 40:
```suggestion
	f.StringVar(&c.keys, "k", "key", "one or more comma-separated key fields (for Zed script indices only)")
	f.StringVar(&c.name, "n", "", "name of index (for Zed script indices only)")
	f.StringVar(&c.zed, "zed", "", "Zed script for index")
```

cmd/zed/lake/index/drop.go
```diff
@@ -3,110 +3,57 @@ package index
...
-
-	if len(defs) == 0 {
-		return errors.New("no definitions deleted")
+	root, err := c.lake.Open(ctx)
+	if err != nil {
+		return err
 	}
-
-	if err := lake.RemoveDefinitions(context.TODO(), lk, defs...); err != nil {
+	xrules, err := root.DeleteIndices(ctx, ids)
```
nwt, line 49: Nit: The type of `xrules` is `[]index.Index`, so `indices` seems like a better name.

cmd/zed/lake/index/drop.go
```diff
@@ -3,110 +3,57 @@ package index
...
-	if err := c.Init(); err != nil {
+	ctx, cleanup, err := c.lake.Init()
+	if err != nil {
 		return err
 	}
-
+	defer cleanup()
 	if len(args) == 0 {
-		return errors.New("no index definition specified")
+		return errors.New("must specify one or more xrule tags")
```
nwt, line 36:
```suggestion
		return errors.New("must specify one or more index tags")
```

lake/commit/actions/actions.go
```diff
@@ -71,3 +73,16 @@ func (d *Delete) CommitID() ksuid.KSUID {
...
 	return "DEL " + d.ID.String()
 }
+
+type AddIndex struct {
+	Commit ksuid.KSUID     `zng:"commit"`
+	Index  index.Reference `zng:"index"`
+}
+
+func (a *AddIndex) String() string {
+	return fmt.Sprintf("ADDX %s", a.Index)
```
nwt, line 83:
```suggestion
	return fmt.Sprintf("ADD_INDEX %s", a.Index)
```

zio/lakeio/writer.go
```diff
@@ -195,3 +200,45 @@ func formatDelete(b *bytes.Buffer, indent int, delete *actions.Delete) {
...
+		b.WriteString(idx.Value)
+	case index.IndexZed:
+		b.WriteString("field(s) ")
+		for i, k := range idx.Keys {
+			if i > 0 {
+				b.WriteString(", ")
+			}
+
+			b.WriteString(k.String())
+
```
nwt, line 237:
```suggestion
			b.WriteString(k.String())
```

nwt (approved):

mccanne (approved):

PR #2619, opened 2021-04-23
move zio.ReaderOpts and WriterOpts to zio/anyio

nwt:
This will enable us to move some of the zbuf interfaces into zio without
creating import cycles.


mccanne (approved):

mattnibs (approved):
I'm pretty sure this is going to break advance zed in brimcap FYI

PR #2620, opened 2021-04-23
better error message for unknown sub-command as suggested by nibs

mccanne:


pkg/charm/path.go
```diff
@@ -10,7 +10,11 @@ type path []*instance
 func (p path) run(args []string) error {
 	err := p.last().command.Run(args)
 	if err == ErrNoRun {
-		err = fmt.Errorf("%q: a sub-command is required: %s", p.pathname(args...), p.subCommands())
+		if len(args) == 0 {
+			err = fmt.Errorf("%q: requires a sub-command: %s", p.pathname(), p.subCommands())
+		} else {
+			err = fmt.Errorf("%q: no such sub-command %q: options are: %s", p.pathname(), args[0], p.subCommands())
+		}
```
nwt, line 17: Maybe return instead of setting `err`?
```suggestion
			return fmt.Errorf("%q: requires a sub-command: %s", p.pathname(), p.subCommands())
		}
		return fmt.Errorf("%q: no such sub-command %q: options are: %s", p.pathname(), args[0], p.subCommands())
```

nwt (approved):

PR #2621, opened 2021-04-23
colorize zson

### LINKED ISSUE #2578: zson colorization
Pretty-zson (-Z) should be colorize fields and values (but not other syntax) like `jq` does,
when writing to terminal stdout.

Example use is in lakeio (which currently just colors the commit tag).

mccanne:
This commit add colorization of zson -Z text.  This is a simple start
that we can improve upon over time.  We should probably unify the
colors with the app (though the terminal color codes are limited and
aren't generic RGB values).  Type decorators are toned down in a
light gray to make them less obstrusive.  We also added -color to
outputflags so color can be disabled with -color=false and the
upcoming defaults settings

Closes #2578

### DIFF
--- a/pkg/terminal/color/stack.go
+++ b/pkg/terminal/color/stack.go
@@ -55,7 +55,7 @@ func (s *Stack) ColorEnd(b *strings.Builder) {
 	}
 }
 
-func (s *Stack) ColorStartBytes(b *bytes.Buffer, code Code) {
+func (s *Stack) StartInBytes(b *bytes.Buffer, code Code) {
 	if !Enabled {
 		return
 	}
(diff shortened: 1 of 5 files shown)


pkg/terminal/color/stack.go
```diff
@@ -55,15 +55,33 @@ func (s *Stack) ColorEnd(b *strings.Builder) {
...
 	b.WriteString(code.String())
 }
 
-func (s *Stack) ColorEndBytes(b *bytes.Buffer) {
+func (s *Stack) EndInBytes(b *bytes.Buffer) {
+	if !Enabled {
+		return
+	}
+	op := s.Pop()
+	if op != "" {
```
nwt, line 71:
```suggestion
	if op := s.Pop(); op != "" {
```

pkg/terminal/color/stack.go
```diff
@@ -55,15 +55,33 @@ func (s *Stack) ColorEnd(b *strings.Builder) {
 	}
 }
 
-func (s *Stack) ColorStartBytes(b *bytes.Buffer, code Code) {
+func (s *Stack) StartInBytes(b *bytes.Buffer, code Code) {
```
nwt, line 58: `bytes.Buffer` and `strings.Builder` both implement `io.Writer`, so you don't really need two versions of the `Start` and `End` methods.
```suggestion
func (s *Stack) Start(w io.Writer, code Code) {
```

nwt (approved):

pkg/terminal/color/stack.go
mccanne: sweet!

PR #2650, opened 2021-05-01
change top-level "." to "this"

mccanne:
Gordon pointed out that "." is weird and cryptic in expressions
like "any(.) by typeof(.)" and suggested we use "this" as in C++.
Just because jq has ".", jquery has "$", and underscore has "_"
doesn't mean we need to jump off the cliff too.

This commit adds "this" and leaves "." in the grammar for now
to see if we like it as a short cut for "this".  Now we can
say "any(this) by typeof(this)".  We can deprecate "." later
depending on how it goes.

### DIFF
--- a/compiler/parser/parser.peg
+++ b/compiler/parser/parser.peg
@@ -654,7 +654,7 @@ DerefExpr
   = first:DotID rest:(Deref)* {
       RETURN(makeBinaryExprChain(first, rest))
     }
-  / first:This rest:(Deref)* {
+  / first:RootRecord rest:(Deref)* {
       RETURN(makeBinaryExprChain(first, rest))
     }
   / first:Identifier rest:(Deref)* {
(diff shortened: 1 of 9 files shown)


compiler/parser/parser.peg
```diff
@@ -654,13 +654,18 @@ DerefExpr
...
+      RETURN(makeBinaryExprChain(first, rest))
+    }
   / first:Identifier rest:(Deref)* {
       RETURN(makeBinaryExprChain(first, rest))
     }
   / "." {
       RETURN(MAP("kind":"Root"))
     }
 
+This = "this" { RETURN(MAP("kind":"Root")) }
```
mattnibs, line 667: Should this be called `ThisToken`?

mccanne: If it's called a token, then the action should be to return TEXT but that's not the case. If this is confusing, we could say  `RootRecord = "this"` instead.

mattnibs (approved):

PR #2669, opened 2021-05-03
rework dataflow model and compiler optimizations

mccanne:
This commit is a large change to improve the way dataflow
is structured.  This change allows the optimizer to better
interact with the lake's "scan scheduler".  It also removes
the mess that ended up in package driver over time.  The driver
package is now restored to its roots, where it was meant to be
a simply interface over which to run a search query (e.g., from
the CLI, from the search endpoint, from the ingest shaper, etc).
There is no more mux.

Data source objects are now embedded in the AST and DAG as
"sources" and a From operator provides a mean to connect
sources with trunks of the input path.  Importantly, each trunk
has a convenient place to store pushdown logic, which will be
naturally carried over the network to distributed works via
the forthcoming run-DAG API.

The searchflags package was removed as from/to key ranges and
the data pool our now embedded in the Zed from operator.

To support different storage layers, there is now a proc.DataAdaptor
interfaces and implementations for the file system, the Zed lake,
and internal readers.

The "order" abstraction has been unified into a top-level package
called "order" and zbuf.Order has been deprecated.  This package
supports a simple order concept (asc/desc) plus a scan-direction
conception (asc/desc/unknown).  There is also an order.Layout
absaction to represent sort keys (primary, secondary, ...) and
an order with a text syntax "key1,key2,...:asc|desc".  The
pool create command uses this syntax with a new "-orderby" option.

The dataflow analysis in the compiler has been improved and
generalized.  It should be much easier to extend and add operators.

A stub for HTTP sources is now in place so it should be trivial
to add support from "from https://example.com/etc"

These changes made it trivial to support cross-pool queries.
A new test has been added to exercise a simple cross-pool join.


compiler/compiler.go
```diff
@@ -68,66 +125,17 @@ func (r *Runtime) Entry() dag.Op {
...
 
 // This must be called before the zbuf.Filter interface will work.
 func (r *Runtime) Optimize() error {
-	return r.optimizer.Optimize()
-}
-
-func (r *Runtime) IsParallelizable() bool {
-	return r.optimizer.IsParallelizable()
+	r.optimizer.OptimizeScan()
+	return nil
```
nwt, line 135:
```suggestion
	return r.optimizer.OptimizeScan()
```

compiler/ast/dag/operator.go
```diff
@@ -210,3 +300,45 @@ type TypeProc struct {
...
+		for _, op := range op.Ops {
+			if err := WalkOps(op, visit); err != nil {
+				return err
+			}
+		}
+	default:
+		return visit(op)
+	}
+	return nil
+}
```
nwt, line 344: This is dead code. Add them when you're going to use them?

order/order.go
```diff
@@ -0,0 +1,120 @@
...
+
+func (l Layout) Equal(to Layout) bool {
+	return l.Order == to.Order && l.Keys.Equal(to.Keys)
+}
+
+func (l Layout) String() string {
+	return fmt.Sprintf("%s:%s", field.List(l.Keys), l.Order)
+}
+
+func (w Which) String() string {
```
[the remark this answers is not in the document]
mccanne: I'll break out into which.go and layout.go

order/direction.go
```diff
@@ -0,0 +1,86 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/brimdata/zed/zng"
+	"github.com/brimdata/zed/zson"
+)
+
+type Direction int
```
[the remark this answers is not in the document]
mccanne: I think it's beneficial to have both and for me, Which shows intent that there is no unknown state (e.g., the layout of sorted data in a pool) vs the scan order requested in a query (e.g., asc, desc, or dontcare).  By putting Which in the pool config  data structure (through Layout), there is no notion that the layout could be unknown.

proc/proc.go
```diff
@@ -26,6 +29,19 @@ type Interface interface {
 	Done()
 }
 
+type DataAdaptor interface {
+	Lookup(context.Context, string) (ksuid.KSUID, error)
+	LayoutOf(context.Context, ksuid.KSUID) (order.Layout, error)
+	NewScheduler(context.Context, *zson.Context, *dag.Pool, zbuf.Filter) (Scheduler, error)
+	Open(context.Context, *zson.Context, string, zbuf.Filter) (zbuf.PullerCloser, error)
+	Get(context.Context, *zson.Context, string, zbuf.Filter) (zbuf.PullerCloser, error)
```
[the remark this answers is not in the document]
mccanne: Open is for files and Get is for URLs.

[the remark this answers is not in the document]
mccanne: Lakes (will) support Get but not Open and the file-system adaptor will support both.

compiler/ast/ast.go
```diff
@@ -345,8 +346,62 @@ type (
...
+	Source()
+}
+
+func (*Pool) Source() {}
+func (*File) Source() {}
+func (*HTTP) Source() {}
+
+type Layout struct {
+	Kind  string `json:"kind" unpack:""`
+	Keys  []Expr `json:"keys"`
```
mattnibs, line 395: Wonderful!

driver/driver.go
```diff
@@ -23,77 +32,175 @@ type Driver interface {
 	Stats(api.ScannerStats) error
 }
 
-func Run(ctx context.Context, d Driver, program ast.Proc, zctx *zson.Context, reader zio.Reader, cfg Config) error {
-	ctx, cancel := context.WithCancel(ctx)
-	defer cancel()
+type Config struct {
+	Custom    kernel.Hook
+	Logger    *zap.Logger
+	Span      nano.Span
```
mattnibs, line 38: I take it this will eventually become something like Range with some kind of expression that conveys min and max for a query?

mattnibs: Oh I see: https://github.com/brimdata/zed/pull/2669/files#diff-abd1a262c5c1943d230a7d41ce0c7c1ea7dd417ff7e5c63a470ffcb884dce7b5R163

Doesn't look like this field is being used.

lake/scheduler.go
```diff
@@ -0,0 +1,154 @@
...
+func (s *Scheduler) run() {
+	if err := ScanPartitions(s.ctx, s.snap, s.span, s.pool.Layout.Order, s.ch); err != nil {
+		s.done <- err
+	}
+	close(s.ch)
+	close(s.done)
+}
+
+// PullPartition returns the next span in the schedule.  This is useful for a
+// worker proc that pulls spans from teh scheddule, sends them to a remote
```
mattnibs, line 73:
```suggestion
// worker proc that pulls spans from the scheduler, sends them to a remote
```

lake/root.go
```diff
@@ -189,6 +211,14 @@ func (r *Root) OpenPool(ctx context.Context, id ksuid.KSUID) (*Pool, error) {
...
 }
 
+func (r *Root) OpenPoolByID(ctx context.Context, id ksuid.KSUID) (*Pool, error) {
+	poolRef := r.LookupPool(ctx, id)
+	if poolRef == nil {
+		return nil, fmt.Errorf("%s: pool not found", id)
+	}
+	return poolRef.Open(ctx, r.path)
+}
+
```
mattnibs, line 221: This is a duplicate of the function above it.

lake/root.go
```diff
@@ -189,6 +211,14 @@ func (r *Root) OpenPool(ctx context.Context, id ksuid.KSUID) (*Pool, error) {
 	return poolRef.Open(ctx, r.path)
 }
 
+func (r *Root) OpenPoolByID(ctx context.Context, id ksuid.KSUID) (*Pool, error) {
+	poolRef := r.LookupPool(ctx, id)
+	if poolRef == nil {
+		return nil, fmt.Errorf("%s: pool not found", id)
```
mattnibs, line 217: Should return ErrPoolNotFound.

driver/driver.go
mccanne: Yes, thanks.  Removing Span from driver.Config.  I've been trying to get rid of driver.Config{} and it will soon be down to logger and statstick after we get rid of custom.  Then we can rip it out.

lake/root.go
mccanne: I deleted this function so not relevant anymore.  But using OpenPool() and ErrPoolNotFound means all the callers have to prepend the pool name/ID.

proc/proc.go
```diff
@@ -26,6 +29,19 @@ type Interface interface {
...
+}
+
+type Scheduler interface {
+	Pull() (zbuf.PullerCloser, error)
```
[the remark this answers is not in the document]
mccanne: PullScanTask as we discussed...

lake/root.go
mattnibs: You can wrap it like `fmt.Errorf("%s: %w", ErrPoolNotFound)`, its important to do this so you can introspect the error. The server needs this to know if it needs to return 404.

lake/mock/mock.go
```diff
@@ -0,0 +1,92 @@
...
+		id:     fakeID(name),
+		layout: layout,
+	}, nil
+}
+
+// fakeID creates a ksuid derived from the name string so we can deterministically
+// check test results rather than emulating ksuid.New().
+func fakeID(name string) ksuid.KSUID {
+	var id ksuid.KSUID
+	copy(id[:], []byte(name))
```
mattnibs, line 56:
```suggestion
	copy(id[:], name)
```

cmd/zed/lake/query/command.go
```diff
@@ -37,50 +34,41 @@ type Command struct {
...
-	at          string
 	includes    query.Includes
 	outputFlags outputflags.Flags
 	procFlags   procflags.Flags
-	searchFlags searchflags.Flags
 }
 
 func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
 	c := &Command{lake: parent.(*zedlake.Command)}
-	f.StringVar(&c.at, "at", "", "commit tag or journal ID for time travel")
```
mattnibs, line 49: This feels like a useful option no?

zbuf/merger.go
```diff
@@ -71,24 +72,24 @@ func NewMerger(ctx context.Context, pullers []Puller, cmp expr.CompareFn) *Merge
 	return m
 }
 
-func MergeReadersByTsAsReader(ctx context.Context, readers []zio.Reader, order Order) (zio.Reader, error) {
+func MergeReadersByTsAsReader(ctx context.Context, readers []zio.Reader, o order.Which) (zio.Reader, error) {
```
nwt, line 75: It's odd that `o` is named not for its type but for its type's package. Maybe rename `Which` to `Order` or `Ordering`?

order/direction.go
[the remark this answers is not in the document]
nwt: 👍

lake/root.go
```diff
@@ -181,6 +184,25 @@ func (r *Root) lookupPoolByName(_ context.Context, name string) *PoolConfig {
 	return nil
 }
 
+func (r *Root) Lookup(ctx context.Context, nameOrID string) (ksuid.KSUID, error) {
```
mattnibs, line 187: I might put a comment here that this is part of the DataAdapter interface. Also I might include a:
```
var _ proc.DataAdapter = (*Root)(nil)
```
somewhere at the top...

zng/time.go
```diff
@@ -55,6 +55,7 @@ func (t *TypeOfTime) ZSONOf(zv zcode.Bytes) string {
 	if err != nil {
 		return badZng(err, t, zv)
 	}
+	//XXX shouldn't this be UTC?
```
nwt, line 58: It is, because `nano.Ts.Time` calls `time.Time.UTC`.

lake/root.go
nwt: I was a leading proponent of "Method implements implements package.Interface.Method" doc comments, but they tend to go stale as names change, so I'm now a leading opponent!

But I am a leading proponent of interface assertions, and I fully endorse adding this one.

zbuf/scanner.go
```diff
@@ -41,12 +48,21 @@ type ScannerStats struct {
...
-	s.RecordsMatched += ss.RecordsMatched
+// Add updates its receiver by adding to it the values in ss.
+func (s *ScannerStats) Add(in ScannerStats) {
+	atomic.AddInt64(&s.BytesRead, in.BytesRead)
+	atomic.AddInt64(&s.BytesMatched, in.BytesMatched)
+	atomic.AddInt64(&s.RecordsRead, in.RecordsRead)
+	atomic.AddInt64(&s.RecordsMatched, in.RecordsMatched)
+}
+
+func (s *ScannerStats) Atomic() ScannerStats {
```
nwt, line 59: Nit: I think `Copy` would be a better name.

zng/time.go
nwt: (So you can zap this comment.)

nwt (approved):
I think this looks great! :shipit:

(My unresolved comments are all minor stuff.)

mattnibs (approved):
Let's go!

order/layout.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/brimdata/zed/field"
+)
+
+var Nil = Layout{}
```
nwt, line 11: Nit: I don't think `Nil` is a great name for this because it's a zero value rather than a `nil` value. Maybe call it `Unset`?

order/layout.go
```diff
@@ -0,0 +1,66 @@
...
+	return fmt.Sprintf("%s:%s", field.List(l.Keys), l.Order)
+}
+
+func (w Which) String() string {
+	if w == Desc {
+		return "desc"
+	}
+	return "asc"
+}
+
```
nwt, line 43: Nit: Move me to `which.go`.

cmd/zed/lake/query/command.go
mccanne: It's in the `from` statement now just like the pool name and the search key range, e.g., https://github.com/brimdata/zed/blob/25bfcdc807aca58a4747a9c8709f58d746bfb7cc/lake/ztests/time-travel.yaml#L9

zbuf/merger.go
mccanne: But it's which order of the two, a boolean.  order.Order is weird to me as is order.Ordering.  It's as to change in a subsequent PR as it is for me to change now and I'm not compelled with this proposal.  Maybe there's something better someone will come up and we can make the change later.

mccanne: The important thing is the shape of the order package is now correct, with a thing encompassing either order plus unknown/dontcare and another thing representing a strict order.

nwt: Agreed!

zng/time.go
mccanne: Okay, removing comment and removing the redunandant UTC() call in TypeOfTime.Marshal()

order/layout.go
mccanne: I was following the ksuid pattern which has ksuid.Nil because you can't use `nil` with this type just like Layout. Seemed like a good idea.   I wasn't bother by the idea that nil is the zero value for pointers, interfaces, etc in Go but package.Nil is the zero value of the thing package implements.  You would never have a need to define Nil if you could use nil instead so that seems really clear to me.

order/layout.go
mccanne: oops

mccanne:
Confirmed `npm test` passes in app against this branch...

PR #2672, opened 2021-05-05
change field.Static to field.Path

mccanne:


field/field.go
```diff
@@ -4,29 +4,29 @@ import (
...
 
 // A root is an empty slice (not nil).
-func NewRoot() Static {
-	return Static{}
+func NewRoot() Path {
+	return Path{}
 }
 
-func (f Static) String() string {
+func (f Path) String() string {
```
nwt, line 18: Nit: Change the receiver name in all of the `Path` methods to `p`.

nwt (approved):
Looks like there are still a couple of `field.Static` references remaining.
```
$ git grep Static
docs/language/expressions/expr.md:like field.Static).  More generally, it can operate on an expression and
lake/ztests/ls.yaml:# The keys field should be labeled with (field.)Static.
```

PR #2688, opened 2021-05-08
rework storage engine abstraction

### LINKED ISSUE #2615: recast iosrc into cloud-object model
The iosrc currently implements a subset of file system semantics then maps those onto s3.   This is upside down.  We should have a storage abstraction that implements cloud object store semantics then we should map those semantics onto a file system for local operation.    This should lead to cleaner code and fewer bugs since the cloud peculiarities are hidden behind a layer.  Also, it will then make immutable object caching trivial as a wrapper on the new storage interface.  While we're at it, we should get rid of the single mux and just have entities that want to talk to storage keep an interface pointer to a storage implementation.

mccanne:
The previous "iosrc" package provided a UFS-like interface whereas
the Zed system should be based on a cloud object model with a
UFS-emulation layer.  This commit converts iosrc to this new
interface abstraction and renames it"pkg/storage".

We also moved the immutable object caching implementation from
lake/immcache to pkg/storage/cache and updated the caches to
wrap and implement a storage.Engine.

We also changed storage.URI from struct{*url.URL} to url.URL
and used pointers throughout to be consistent with url.URL
usage patterns.  The stdio URI pattern has been changed to
elide the unnecessary "///" characters, e.g., stdio:stdin
instead of stdio:///stdin, which conforms with the URL RFC.

This change also exposed some structural problems with assumptions
when local files can be accessed.  For example, tests were passing
for s3 when using WriteIfNotExists but that should have failed.
There is now an issue for implementing PutIfNotExists on top of
S3 strong consistency and this is documented in the journal commit
code path.

Closes #2615


lake/root.go
```diff
@@ -115,12 +124,17 @@ func (r *Root) StoreConfig(ctx context.Context) error {
...
-		return json.NewEncoder(w).Encode(r.Config)
-	})
+	//XXX this will soon change to a key-value journal.  for now we write
+	// the new config in its entirety and hope there is no error.
+	b, err := json.Marshal(r.Config)
 	if err != nil {
 		return err
 	}
+	// XXX
+	local := storage.NewLocalEngine()
```
mattnibs, line 134: Shouldn't this use r.engine?

pkg/storage/uri.go
```diff
@@ -0,0 +1,84 @@
...
+
+import (
+	"net/url"
+	"strings"
+
+	"github.com/brimdata/zed/zng"
+	"github.com/brimdata/zed/zson"
+)
+
+type URI url.URL
```
mattnibs, line 11: I kind of think uri should be be its own package in `pkg`. What do you think?

cmd/zed/lake/flags.go
```diff
@@ -31,31 +31,31 @@ func (f *Flags) SetFlags(set *flag.FlagSet) {
...
 }
 
-func (f *Flags) RootPath() (iosrc.URI, error) {
-	return iosrc.ParseURI(f.Root)
+func (f *Flags) RootPath() (*storage.URI, error) {
+	return storage.ParseURI(f.Root)
 }
 
-func (f *Flags) Create(ctx context.Context) (*lake.Root, error) {
+func (f *Flags) Create(ctx context.Context, engine storage.Engine) (*lake.Root, error) {
```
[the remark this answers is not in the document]
mccanne: I would leave this logic to the commands instead of the shared Flags type.  For example, zed lake serve does not directly create a storage engine as service.Core allocates it as a remote engine.

lake/root.go
mccanne: oops

mattnibs (approved):

PR #2689, opened 2021-05-08
fix whitespace bug in ztest regexps

mccanne:
Many times I have written a regexp ztest rule only to be mystified
for why it didn't work.  The mystery lies in a yaml newline that
is embedded in the regexp requiring a silent match of a newline
immediately at the end of any regexp.  This commit adds a
string.TrimSpace to ztest whitespace to fix this.  This means
that you can't rely on whitespace at the start or end of a regexp
and must explicitly put any such whitespace in parens.  None of the
existing tests work this way.

We also removed an extraneous single quote from the regexp error
message.

### DIFF
--- a/ztest/ztest.go
+++ b/ztest/ztest.go
@@ -257,7 +257,7 @@ func (f *File) load(dir string) ([]byte, *regexp.Regexp, error) {
 		return b, nil, err
 	}
 	if f.Re != "" {
-		re, err := regexp.Compile(f.Re)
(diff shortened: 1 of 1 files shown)


mattnibs (approved):

nwt:
I don't think this is the right move.

If you don't want the newline, you can use a flow scalar (i.e., `regexp: no newline`, `regexp: 'no newline'`, or `regexp: "no newline"`) or the block chomping indicator (`regexp: |-`).

mccanne:
Closing based on @nwt feedback.

PR #2691, opened 2021-05-08
silently discard duplicate record fields in zson parser

### LINKED ISSUE #2523: Tolerate duplicate fields in NDJSON records
Repro is with `zq` commit 99477b6.

Consider the following NDJSON:

```
$ cat has_dup.ndjson 
{"foo": 0}
{"foo": 1, "bar": 2, "foo": 3}
{"foo": 2}
```

At the moment, the presence of the duplicate field in the middle record causes `zq` to stop processing the stream entirely.

```
$ zq -f ndjson -i ndjson has_dup.ndjson 
has_dup.ndjson: line 2: field foo is repeated
```

When digging for background on the topic, I found my way to the [JSON RFC 7159](https://tools.ietf.org/html/rfc7159) which covers it pretty clearly:

> The names within an object SHOULD be unique.  
> ...  
> When the names within an object are not unique, the behavior of software that receives such an object is unpredictable.  Many implementations report the last name/value pair only. Other implementations report an error or fail to parse the object, and some implementations report all of the name/value pairs, including duplicates.

By this definition, if we simply skipped over records that contain such duplicates we'd be in-line with the RFC behavior. However, in terms of what's best for the user, we should definitely _not_ be bailing out on the whole stream. For instance, we bumped into this problem in brimdata/brimcap#13 where Suricata v6 was generating output with such duplicate fields. In that case it was causing the entire Zed processing of the log stream to fail and effectively broke the ability to import pcap data via Brimcap. Being tolerant of the input would have been an improvement.

To compare with another tool, `jq` happens to quietly tolerate the presence of the duplicated field, apparently taking the approach of reporting "the last name/value pair only".

```
$ cat has_dup.ndjson | jq -c .
{"foo":0}
{"foo":3,"bar":2}
{"foo":2}
```

I'd advocate that we adapt the same behavior of `jq`, since it's tolerant and would help us work around the problem described in brimdata/brimcap#13. As a team we also discussed printing a warning when this happens.

mccanne:
Closes #2523

### DIFF
--- a/zson/parser-values.go
+++ b/zson/parser-values.go
@@ -319,6 +319,7 @@ func (p *Parser) matchRecord() (*zed.Record, error) {
 func (p *Parser) matchFields() ([]zed.Field, error) {
 	l := p.lexer
 	var fields []zed.Field
+	seen := make(map[string]struct{})
 	for {
 		field, err := p.matchField()
 		if err != nil {
@@ -327,7 +328,10 @@ func (p *Parser) matchFields() ([]zed.Field, error) {
 		if field == nil {
 			break
 		}
-		fields = append(fields, *field)
+		if _, ok := seen[field.Name]; !ok {
+			fields = append(fields, *field)
+		}
+		seen[field.Name] = struct{}{}
 		ok, err := l.match(',')
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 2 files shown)


philrz:
This worked around [Suricata bug 4016](https://redmine.openinfosecfoundation.org/issues/4106).

PR #2692, opened 2021-05-08
Simplify keyword search

mccanne:
This commit simplifies the keyword search experience by eliminating the field shortcuts, replacing "=" for assignment (":=") and comparison ("=="),  and introducing a "matches" keyword for regex match and glob match. The `*=value` and `value in *` patterns are no longer available but will be subsumed by better syntax in a forthcoming PR.

This change will break the app so we need to port and do integration testing before merging this PR.


compiler/parser/parser.peg
```diff
@@ -209,13 +208,17 @@ ShortCut
                                       "rhs": MAP("kind": "ID", "name": "$"),
                                       "lhs": v))))))))
     }
-  / !(SearchGuard EOT) search:PatternSearch { RETURN(search) }
+
+
```
nwt, line 212:
```suggestion

```

nwt (approved):

jameskerr:
I found a bug in this branch when I search "files" in the app.

Screen Shot 2021-05-13 at 3 37 04 PM

Screen Shot 2021-05-13 at 3 37 26 PM

The queries that is being run in the background are:

```
files
```
and 

```
files | every 1d count() by _path
```

nwt:
@jameskerr: brimdata/zed#2721 should fix that bug.

PR #2693, opened 2021-05-09
add support for marshaling zng.Value fields

mccanne:
This commit adds support for marshaling and unmarshaling zng Values
that are embedded in Go structs.  This allows us to encode arbitrary
zng.Values as part of native Go data structures, e.g., to encode
the upper/lower range boundary as zng.Values in the metadata struct
of an arbitrary Zed lake data object.

While here, we also improved the error message when trying to decode
a non-record Zed value into a Go struct.

### DIFF
--- a/zson/marshal.go
+++ b/zson/marshal.go
@@ -275,6 +275,7 @@ func (m *MarshalZNGContext) NamedBindings(bindings []Binding) error {
 }
 
 var nanoTsType = reflect.TypeOf(nano.Ts(0))
+var zngValueType = reflect.TypeOf(zng.Value{})
 
 func (m *MarshalZNGContext) encodeValue(v reflect.Value) (zng.Type, error) {
 	typ, err := m.encodeAny(v)
@@ -292,8 +293,10 @@ func (m *MarshalZNGContext) encodeValue(v reflect.Value) (zng.Type, error) {
 		kind := v.Kind().String()
 		if name != "" && name != kind {
 			// We do not want to further decorate nano.Ts as
-			// it's already been converted to a Zed time.
-			if v.Type() == nanoTsType {
+			// it's already been converted to a Zed time;
+			// likewise for zng.Value, which gets encoded as
+			// itself and its own alias type if it has one.
+			if v.Type() == nanoTsType || v.Type() == zngValueType {
 				return typ, nil
 			}
 			path := v.Type().PkgPath()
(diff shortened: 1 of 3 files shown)


zson/marshal_test.go
```diff
@@ -235,3 +235,50 @@ func TestUnexported(t *testing.T) {
 	_, err := m.Marshal(f)
 	require.NoError(t, err)
 }
+
+type EmbeddedZNG struct {
+	Name string
+	Dawg zng.Value `zng:"yo"`
```
mattnibs, line 241: 😆

zson/marshal_test.go
```diff
@@ -235,3 +235,50 @@ func TestUnexported(t *testing.T) {
 	_, err := m.Marshal(f)
 	require.NoError(t, err)
 }
+
+type EmbeddedZNG struct {
```
mattnibs, line 239: I would think an embedded test would be:

```
type Bar struct {
    V1 string `zng:"v1"`
    V2 string `zng:"v2"`
}

type Foo struct {
    Type string `zng:"type"`
    Bar
}
```

Which the marshal to:

```
{
    "type": "Bar",
    "v1": "v1",
    "v2":"v2",
}
```

zson/marshal.go
```diff
@@ -292,8 +293,10 @@ func (m *MarshalZNGContext) encodeValue(v reflect.Value) (zng.Type, error) {
 		kind := v.Kind().String()
 		if name != "" && name != kind {
 			// We do not want to further decorate nano.Ts as
-			// it's already been converted to a Zed time.
-			if v.Type() == nanoTsType {
+			// it's already been converted to a Zed time;
+			// likewise for zng.Value, which gets encoded as
+			// itself and its own alias type if it has one.
+			if v.Type() == nanoTsType || v.Type() == zngValueType {
```
nwt, line 299: Nit: Call `v.Type` once.
```suggestion
			if t := v.Type(); t == nanoTsType || t == zngValueType {
```

zson/marshal.go
```diff
@@ -597,6 +612,12 @@ func (u *UnmarshalZNGContext) decodeAny(zv zng.Value, v reflect.Value) error {
 			return pv.Interface().(ZNGUnmarshaler).UnmarshalZNG(u, zv)
 		}
 	}
+	if _, ok := v.Interface().(zng.Value); ok {
+		// For embedded zng.Values we simply set the
+		// reflect value to the zng.Value that has been decoded.
+		v.Set(reflect.ValueOf(zv))
+		return nil
+	}
 	if _, ok := v.Interface().(nano.Ts); ok {
```
nwt, line 621: Nit: Move the `zng.Value` if statement after `nano.Ts` one for consistency with `encodeAny`.

zson/marshal_test.go
mccanne: Are you asking me to change the word "embedded" in the test?  I am using it to mean you can embed zng.Values in zng-marshaled Go structs.    We don't hoist structs like your example and neither does encoding/json.  Mapstructure has an option to do this.  It feels low priority.

nwt: @mattnibs point might be that "embedded" in the Go spec denotes something different than what's going on here. For that reason, I'd encourage use of a different term like "zng.Value field".

nwt (approved):
For posterity, it would be good to replace "embedded zng.Value" with `zng.Value field" in the PR title and description as well as the commit message when you merge.

PR #2697, opened 2021-05-10
Allow escaped forward slash to conform with ZSON spec

### LINKED ISSUE #2604: Handling of single-backslash-escaped / in NDJSON value
#1687 described a symptom that was originally specific to the old NDJSON `types.json` support that has since been removed. However, we see the symptom now shows up for NDJSON data in general.

Its origins are in how we've seen Suricata's EVE JSON logs escape `/` with a single `\` like so:

```
$ cat comm.ndjson 
{"community_id":"1:79bZuCMMmkFg6FLG5WIs\/x+8fP0="}
```

As of `zq` commit 8f7e395, we actually end up turning the `\` into `\\`:

```
$ zq -f ndjson comm.ndjson 
{"community_id":"1:79bZuCMMmkFg6FLG5WIs\\/x+8fP0="}
```

Meanwhile, `jq` swallows the single `\`:

```
$ cat comm.ndjson | jq -c .
{"community_id":"1:79bZuCMMmkFg6FLG5WIs/x+8fP0="}
```

As shown in #1687, our older, non-typed NDJSON reader behaved like `jq`, but after seeing our current behavior I did a binary search and confirmed that the `\\`\-style output shown above started appearing with #2123 when NDJSON started being read by the ZSON parser.

I don't really see it from reading the JSON RFC, but in looking around I found comments like [https://github.com/google/jsonnet/issues/679#issuecomment-508419558](https://github.com/google/jsonnet/issues/679#issuecomment-508419558) that seem to imply the `jq` behavior is indeed the expected one, so we should likely conform to that.

FWIW, we've mostly worked around this problem to date because, per the [Suricata docs](https://suricata.readthedocs.io/en/suricata-6.0.0/output/eve/eve-json-output.html#json-flags), this escaping behavior can be turned off, and that's what we've done in the Suricata artifact we've traditionally bundled with Brim and now include as part of Brimcap. However, we can imagine a future when users will want to bring their own custom Suricata, and perhaps there's other systems out there that produce this kind of output.

mccanne:
Closes #2604.

### DIFF
--- a/docs/formats/zson.md
+++ b/docs/formats/zson.md
@@ -424,7 +424,8 @@ value in a record, as an element in an array, etc).
 #### 3.3.1 String Escape Rules
 
 Double-quoted `string` syntax is the same as that of JSON as described
-[RFC 8259](https://tools.ietf.org/html/rfc8259#section-7), specifically:
+[RFC 8259](https://tools.ietf.org/html/rfc8259#section-7), except for
+the special case `\/`, specifically:
 
 * The sequence `\uhhhh` where each `h` is a hexadecimal digit represents
   the Unicode code point corresponding to the given
@@ -432,6 +433,10 @@ Double-quoted `string` syntax is the same as that of JSON as described
 * `\u{h*}` where there are from 1 to 6 hexadecimal digits inside the
   brackets represents the Unicode code point corresponding to the given
   hexadecimal number.
+
+Unlike in the JSON specification, `\/` represents the
+forward slash character (U+2B) though slash may appear without being
+escaped in this fashion.
 
 `\u` followed by anything that does not conform to the above syntax
 is not a valid escape sequence.  The behavior of an implementation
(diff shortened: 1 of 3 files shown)


docs/formats/zson.md
```diff
@@ -424,7 +424,8 @@ value in a record, as an element in an array, etc).
 #### 3.3.1 String Escape Rules
 
 Double-quoted `string` syntax is the same as that of JSON as described
-[RFC 8259](https://tools.ietf.org/html/rfc8259#section-7), specifically:
+[RFC 8259](https://tools.ietf.org/html/rfc8259#section-7), except for
+the special case `\/`, specifically:
```
nwt, line 428: https://tools.ietf.org/html/rfc8259#section-7 specifically says `\/` is a recognized two-character sequence escape representation for `/`, so this isn't an exception.

mccanne: Thanks for checking.  Mystery solved!

PR #2701, opened 2021-05-10
report reasonable error when no zed lake root

### LINKED ISSUE #2671: zed lake serve: Improve error message when no lake root specified
Repro is with Zed commit cbfb2d7.

I happened to forget to set `ZED_LAKE_ROOT` or specify `-R` when attempting to run `zed lake serve`, and this was the observed output:

```
$ zed lake serve
{"level":"info","ts":1620169439.6254737,"msg":"Open files limit raised","limit":65536}
item does not exist: 
```

It took me a few minutes of head-scratching to realize my mistake. It might be helpful if `zed` had provided better guidance along the lines of:

> error: path to Zed Lake store must be specified via `-R` flag or `ZED_LAKE_ROOT` variable

mccanne:
This commit provides a better error when no lake root is specified.
While we were at it, we switch serivce.Config to use a storage.URI
instead of string to specify the root path.

Closes #2671

### DIFF
--- a/cmd/zed/lake/flags.go
+++ b/cmd/zed/lake/flags.go
@@ -49,7 +49,7 @@ func (f *Flags) Open(ctx context.Context, engine storage.Engine) (*lake.Root, er
 		return nil, err
 	}
 	if root.Path == "" {
-		return nil, errors.New("no lake path specied: use -R or ZED_LAKE_ROOT environment")
+		return nil, errors.New("no lake path specied: use -R or set ZED_LAKE_ROOT")
 	}
 	return lake.Open(ctx, engine, root)
 }
--- a/cmd/zed/lake/flags.go
+++ b/cmd/zed/lake/flags.go
@@ -48,6 +48,9 @@ func (f *Flags) Open(ctx context.Context, engine storage.Engine) (*lake.Root, er
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


cmd/zed/lake/flags.go
```diff
@@ -48,6 +48,9 @@ func (f *Flags) Open(ctx context.Context, engine storage.Engine) (*lake.Root, er
 	if err != nil {
 		return nil, err
 	}
+	if root.Path == "" {
+		return nil, errors.New("no lake path specied: use -R or ZED_LAKE_ROOT environment")
```
nwt, line 52: Nit The phrase "use -R or ZED_LAKE_ROOT environment" reads oddly. How about "use -R or ZED_LAKE_ROOT environment variable" or "use -R or set ZED_LAKE_ROOT"?

cmd/zed/lake/serve/command.go
```diff
@@ -75,6 +75,10 @@ func (c *Command) Run(args []string) error {
 		return err
 	}
 	defer cleanup()
+	lakeRoot, err := c.lake.RootPath()
```
nwt, line 78: Nit: No need for `lakeRoot`.
```suggestion
	c.conf.Root, err = c.lake.RootPath()
```

nwt (approved):

PR #2706, opened 2021-05-11
Fix FileStorage size on windows

mattnibs:
Method should use Filepath() not Path, which will have a "/" before the
drive (not a valid windows path).


mattnibs:
Make sure https://github.com/brimdata/brim/pull/1628 is passing before merge.

mccanne (approved):

Issue #1429, opened 2020-10-06
expr errors should be embedded data errors instead of fatal errors


mccanne:
Zng now has an "error" type so that errors encountered in an analytics pipeline can be embedded in data instead of halting the computation (or dropping records) and reporting back an error.  This will be useful for large jobs that might have a few hiccups due to unanticipated data formats.  The user can then see exactly what's going by seeing where and how  errors appear intermixed in the result and adjust their queries (or perhaps the ingest policy) accordingly.

The first step here is to take a pass through package expr and make sure errors get propagated all the way up the the expression evaluation tree.  Such errors should be turned into zng.TypeError values with a go error of nil.  This will require updates to tests that expect the computation to halt or drop records and send the error message to stderr.

mccanne:
This was done in PR #3368.

PR #2708, opened 2021-05-11
change nameof() to return Zed missing instead of Go error

mccanne:
In lieu of the more comprehensive work in issue #1429, this commit
fixes the nameof() operator to return a proper Zed missing value.

### DIFF
--- a/zng/error.go
+++ b/zng/error.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"fmt"
 
-	"github.com/brimdata/zed/pkg/byteconv"
 	"github.com/brimdata/zed/zcode"
 )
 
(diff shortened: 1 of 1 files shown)


zng/error.go
```diff
@@ -64,3 +65,7 @@ func (t *TypeOfError) ZSON() string {
 func (t *TypeOfError) ZSONOf(zv zcode.Bytes) string {
 	return QuotedString(zv, false)
 }
+
+func IsMissing(zv Value) bool {
+	return zv.Type == TypeError && byteconv.UnsafeString(zv.Bytes) == missing
```
nwt, line 70: The compiler is smart enough to avoid allocation here.
```suggestion
	return zv.Type == TypeError && string(zv.Bytes) == missing
```

PR #1612, opened 2020-11-14
Use testing.T.TempDir

nwt:

PR #2716, opened 2021-05-12
use testing.T.TempDir

nwt:
#1612 redux.


mccanne (approved):
Someday we'll stop doing this!

PR #2718, opened 2021-05-13
fix type-value consistency bug

mccanne:
This commit fixes a bug where type values were created both with the
ZSON() method on zng.Type as well as zson.FormatType().  The latter
is correct as the ZSON() method does not create canonical form and
import cycles prevent an easy fix.  The workaround for now is to
move NewTypeType() into package zson and have it call zson.FormatType()
instead of zng.Type.ZSON().


nwt (approved):
Yuck.

PR #2722, opened 2021-05-14
Update page users see when connecting from a browser

philrz:
While updating the Troubleshooting article in the Brim wiki, I noticed that this page still mentions `zqd` and therefore should be updated.

Just to refresh folks on why this page exists, we once had Support issues back in the day because users on Linux didn't realize Brim was a desktop app and therefore they installed via `yum` or `apt` from the shell and then didn't know how to launch it. They poked around and saw the listening `zqd` process and figured they should connect to it from a browser.

The new page:

---

### DIFF
--- a/service/core.go
+++ b/service/core.go
@@ -24,7 +24,7 @@ const indexPage = `
   <title>ZQD daemon</title>
   <body style="padding:10px">
     <h2>zed lake serve</h2>
-    <p>A <a href="https://github.com/brimdata/zed/tree/main/cmd/zed/lake/serve">zed lake serve</a> process is listening on this host/port.</p>
+    <p>A <a href="https://github.com/brimdata/zed/tree/main/cmd/zed/lake/serve">zed lake service</a> is listening on this host/port.</p>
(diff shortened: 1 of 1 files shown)


service/core.go
```diff
@@ -23,10 +23,10 @@ const indexPage = `
 <html>
   <title>ZQD daemon</title>
   <body style="padding:10px">
-    <h2>ZQD</h2>
-    <p>A <a href="https://github.com/brimdata/zed/tree/main/cmd/zed/lake/serve">zqd</a> daemon is listening on this host/port.</p>
+    <h2>zed lake serve</h2>
+    <p>A <a href="https://github.com/brimdata/zed/tree/main/cmd/zed/lake/serve">zed lake serve</a> process is listening on this host/port.</p>
```
mattnibs, line 27: Might be a debate here but IMHO the "what" is zed lake service, `zed lake serve` is the command to init a zed lake service.
```suggestion
    <p>A <a href="https://github.com/brimdata/zed/tree/main/cmd/zed/lake/serve">zed lake service</a> is listening on this host/port.</p>
```

mattnibs (approved):

service/core.go
philrz: @mattnibs: No controversy here. I was already on the fence about "serve" vs. "service" and am happy to flip it, so I just pushed the change and will update the screenshot in the PR. Thanks for the approval!

PR #2729, opened 2021-05-17
add support for arbitrary pool keys

### LINKED ISSUE #2482: Zed lakes should handle partition keys other than "ts"
In the Zeek-centric era, Zed lakes were assumed to always be keyed by a field of type `time` called `ts`. For the future where all generic data will be welcomed, lakes need the ability to handle keys of other data types and named fields.

(@mccanne recently clarified for me that while the Zed commands currently give the impression that this is already possible \[e.g. `zed lake create -p logs -k otherkey -order asc` responds as if it succeeded\], but the lake itself still is wired to expect the key to be `ts` and sort by it. So the alternate-keyed data in this example would get treated as if every record had `ts=0` and so it's effectively unsorted when you try to query the lake for it. So we definitely have work left to to to scale it, provide the same zoom-in/zoom-out experience we've traditionally had with time, etc.)

mccanne:
The backend previously presumed the pool key was of type time. 
This commit generalizes the scanning logic to allow pool keys of any type.

The seek index has been simplified to use a simple, flat
index that is no longer a "micro index".  Instead it is a simple,
single-level list of keys that is used at open to get the
scan range for a row object when the scan is smaller than the
whole object.  This design presumes the index will be cached
when running multiple queries over a sub-range within a row object.

While updating tests, we noticed the segment size and row_size
fields were virtually the same so we deleted the size field.

As part of this commit, the index package was updated to use
zson format for key parsing instead of the deprecated tzng format,
so we removed zio/tzngio/builder.go.

Closes #2482


lake/seekindex/lookup.go
```diff
@@ -0,0 +1,61 @@
...
+			return Range{}, err
+		}
+		if cmp(key, from) > 0 {
+			break
+		}
+		off, err := rec.Access("offset")
+		if err != nil {
+			return Range{}, err
+		}
+		rg.Start, _ = zng.DecodeInt(off.Bytes)
```
nwt, line 34:
```suggestion
		rg.Start, err = zng.DecodeInt(off.Bytes)
		if err != nil {
				return Range{}, err
		}		
```

lake/seekindex/lookup.go
```diff
@@ -0,0 +1,61 @@
...
+		if err != nil {
+			return Range{}, err
+		}
+		if cmp(key, to) > 0 {
+			off, err := rec.Access("offset")
+			if err != nil {
+				return Range{}, err
+			}
+			rg.End, err = zng.DecodeInt(off.Bytes)
+			break
```
nwt, line 50:
```suggestion
			rg.End, err = zng.DecodeInt(off.Bytes)
			if err != nil {
				return Range{}, err
			}
			break
```

lake/seekindex/seekindex_test.go
```diff
@@ -68,32 +71,46 @@ func (e entries) Order() order.Which {
...
-	path := filepath.Join(t.TempDir(), "seekindex.zng")
-	builder, err := NewBuilder(context.Background(), storage.NewLocalEngine(), path, entries.Order())
-	require.NoError(t, err)
+type nopCloser struct {
+	io.Writer
+}
+
+func (*nopCloser) Close() error {
+	return nil
+}
```
nwt, line 105: Use `zio.NopCloser`.

lake/segment/reader.go
```diff
@@ -21,36 +25,44 @@ type Reader struct {
...
+		ReadBytes:  r.RowSize, //XXX
+	}
+	// If a whole segment has nulls for the key values, just return the
+	// whole-segment reader.  Eventually, we will store keyless rows some
+	// other way, perhaps in a sub-pool.
+	if r.First.Bytes == nil || r.Last.Bytes == nil {
+		return sr, nil
+	}
+	segspan := extent.NewGeneric(r.First, r.Last, cmp)
+	if ok := segspan.Crop(scanRange); !ok {
```
nwt, line 47:
```suggestion
	if !segspan.Crop(scanRange) {
```

lake/segment/writer.go
```diff
@@ -1,76 +1,67 @@
...
+	poolKey         field.Path
 }
 
-type WriterOpts struct {
-	Order order.Which
-	Zng   zngio.WriterOpts
-}
-
-func (r *Reference) NewWriter(ctx context.Context, engine storage.Engine, path *storage.URI, opts WriterOpts) (*Writer, error) {
+func (r *Reference) NewWriter(ctx context.Context, engine storage.Engine, path *storage.URI, o order.Which, seekIndexFactor int, poolKey field.Path) (*Writer, error) {
```
nwt, line 34: Nit: `poolKey` feels more important than `o` or `seekIndexFactor`, so maybe put it first.

lake/segment/writer.go
```diff
@@ -1,76 +1,67 @@
...
 	}
 	counter := &writeCounter{bufwriter.New(out), 0}
-	writer := zngio.NewWriter(counter, opts.Zng)
-	seekObjectPath := r.SeekObjectPath(path)
-	seekIndex, err := seekindex.NewBuilder(ctx, engine, seekObjectPath.String(), opts.Order)
+	opts := zngio.WriterOpts{
+		StreamRecordsMax: seekIndexFactor,
+		LZ4BlockSize:     zngio.DefaultLZ4BlockSize,
+	}
+	writer := zngio.NewWriter(counter, opts)
```
nwt, line 44:
```suggestion
	writer := zngio.NewWriter(counter, zngio.WriterOpts{
		StreamRecordsMax: seekIndexFactor,
		LZ4BlockSize:     zngio.DefaultLZ4BlockSize,
	})
```

lake/sorted.go
```diff
@@ -46,19 +50,68 @@ func (s *statScanner) Pull() (zbuf.Batch, error) {
 	return batch, err
 }
 
+type rangeWrapper struct {
```
nwt, line 53: Nit: This file will read better if you put `rangeWrapper` and methods after `newSortedScanner`.

zson/zson.go
```diff
@@ -92,3 +92,11 @@ func TranslateType(zctx *Context, astType zed.Type) (zng.Type, error) {
 func NewTypeType(t zng.Type) zng.Value {
 	return zng.Value{zng.TypeType, zcode.Bytes(FormatType(t))}
 }
+
+func ParseValueFromAST(zctx *Context, ast zed.Value) (zng.Value, error) {
```
nwt, line 96: Nit: Put this after `ParseValue`.

zng/recordval.go
```diff
@@ -235,6 +236,18 @@ func (r *Record) Access(field string) (Value, error) {
...
+func (r *Record) Deref(path field.Path) (Value, error) {
+	var v Value
+	for _, f := range path {
+		var err error
+		v, err = r.Access(f)
+		if err != nil {
+			return Value{}, err
+		}
+	}
+	return v, nil
```
nwt, line 248:
```suggestion
	v := r.Value
	for _, f := range path {
		typ := zng.TypeRecordOf(v.Type)
		if typ == nil {
			return Value{}, errors.New("field access on non-record value")
		}
		var err error
		v, err = zng.NewVolatileRecord(typ, v.Bytes).Access(f)
		if err != nil {
			return Value{}, err
		}
	}
	return v, nil
```

compiler/kernel/proc.go
```diff
@@ -557,3 +570,19 @@ func (b *Builder) LoadConsts(ops []dag.Op) error {
 	}
 	return nil
 }
+
+func evalAtCompileTime(zctx *zson.Context, scope *Scope, in dag.Expr) (zng.Value, error) {
```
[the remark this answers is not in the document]
mccanne: I think I'll leave it because it's not a generic eval.  When we add generalized constant folding, it will be subsumed by that work and we can change it then.

expr/extent/span.go
```diff
@@ -0,0 +1,136 @@
...
+	last  zng.Value
+	cmp   expr.ValueCompareFn
+}
+
+// CompareFunc returns a generic comparator suitable for use in a Range
+// based on the order of values in the range, i.e., when order is desc
+// then the first value is larger than the last value and Before is true
+// for larger values while After is true for smaller values, etc.
+func CompareFunc(o order.Which) expr.ValueCompareFn {
+	// Treat nullsmax as zero by passing false to NewValueCompareFn().
```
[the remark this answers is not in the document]
mccanne: Well, it's min but ok.

expr/extent/span.go
```diff
@@ -0,0 +1,136 @@
...
+	if o == order.Asc {
+		return cmp
+	}
+	return func(a, b zng.Value) int { return cmp(b, a) }
+}
+
+// Create a new Range from generic zng.Values.  If first is greater
+// than last according to cmp, then the values are reversed so that
+// first comes first.
+func NewGeneric(first, last zng.Value, cmp expr.ValueCompareFn) *Generic {
```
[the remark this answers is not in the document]
mccanne: To me a and b is even more confusing.  As I was writing the code I found the convention first/last and lower/upper good pneumonics.  "First" means first wrt to the ordering and "Last" means last wrt to the ordering and this is used consistently throughout the code.  If you try to write all of the methods with a/b, you will quickly get lost in the logic because you will forget when to use < and > on the cmp() results and you will have to reconstruct that Generic.A must be less then Generic.B with respect to Generic.cmp() so better to name them in a way that conveys that invariant in the first place.

expr/extent/span.go
```diff
@@ -0,0 +1,136 @@
...
+	Before(zng.Value) bool
+	After(zng.Value) bool
+	In(zng.Value) bool
+	Overlaps(zng.Value, zng.Value) bool
+	Crop(Span) bool
+	Extend(zng.Value)
+	String() string
+}
+
+type Generic struct {
```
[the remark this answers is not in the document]
mccanne: We can do this later.  Right now, having an outside pointer to extent.Generic is a signal that the work has not yet been done to optimize the innards and the PR that does the first stab at that work can change these pointers, which will serve as a convenient way to update all the (small number of) constructor call sites accordingly when the call logic might change.

expr/extent/span.go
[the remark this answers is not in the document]
mccanne: I used the convention first/last when the order is defined (e.g., in extent.Generic fields) and lower/upper when the order is not defined (e.g., in the ast/dag nodes) so I will use lower/upper here.

lake/segment/writer.go
mccanne: The problem here is the design pattern for calling NewWriter with opts.

zng/recordval.go
mccanne: I don't think we need to do this.  Let's discuss.

mccanne: This speaks to the way we think about zng.Value and the forthcoming changes.

mccanne: My bad, but yuck, this stuff is ancient and needs refactoring.  I think operators should be applied to zng.Value and it would switch inside.  Then, we can transform this pattern into the JIT-CG (just-in-time closure generator) that produces an efficient closure as a value opeator for any top-level schema.

nwt (approved):
:shipit:

PR #2731, opened 2021-05-17
simplify zng mapper

mccanne:


zng/resolver/mapper.go
```diff
@@ -53,6 +53,18 @@ func (m *Mapper) Translate(foreign zng.Type) (zng.Type, error) {
 
 func (m *Mapper) EnterType(td int, typ zng.Type) {
 	m.mu.Lock()
-	m.Slice.Enter(td, typ)
+	if td >= len(m.types) {
+		new := make([]zng.Type, td+1, td*2)
+		copy(new, m.types)
+		m.types = new
+	}
```
nwt, line 60: Nit: No need to copy if `types` has sufficient capacity.
```suggestion
	if td >= cap(m.types) {
		new := make([]zng.Type, td+1, td*2)
		copy(new, m.types)
		m.types = new	
	} else if td >= len(m.types) {
		m.types = t.type[:td+1]
	}
```

zng/resolver/mapper.go
```diff
@@ -53,6 +53,18 @@ func (m *Mapper) Translate(foreign zng.Type) (zng.Type, error) {
...
+	m.types[td] = typ
 	m.mu.Unlock()
 }
+
+func (m *Mapper) Lookup(td int) zng.Type {
+	if td >= 0 && td < len(m.types) {
```
nwt, line 66:
```suggestion
	m.mu.RLock()
	defer m.mu.RUnlock()
	if td < len(m.types) {
```

zng/resolver/mapper.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 type Mapper struct {
-	Slice
+	types     []zng.Type
 	mu        sync.RWMutex
 	outputCtx *zson.Context
```
nwt, line 13: Nit: `mu` protects `types` but not `outputCtx`, so group and order accordingly.
```suggestion
	outputCtx *zson.Context

	mu        sync.RWMutex
	types     []zng.Type
```

nwt (approved):

PR #2737, opened 2021-05-19
Graceful shutdown with SIGTERM

mattnibs:


cmd/zed/root/command.go
```diff
@@ -35,7 +36,7 @@ func (c *Command) Init(all ...cli.Initializer) (context.Context, func(), error)
 	if err := c.cli.Init(all...); err != nil {
 		return nil, nil, err
 	}
-	ctx, cancel := signalctx.New(os.Interrupt)
+	ctx, cancel := signalctx.New(os.Interrupt, syscall.SIGTERM)
```
mccanne, line 39: change os.Interrupt to syscall.SIGINT for consistency with syscall.SIGTERM

mccanne (approved):

cmd/zed/root/command.go
nwt: I'm with @mccanne on this.

nwt (approved):

PR #2753, opened 2021-05-20
zed query: allow from in -I query without positional argument

### LINKED ISSUE #2715: Support "from file" Zed syntax in zq
The verification notes at [https://github.com/brimdata/zed/issues/2483#issuecomment-840048595](https://github.com/brimdata/zed/issues/2483#issuecomment-840048595) show an example `join` that doesn't work quite the way it used to. @mccanne mentioned that another way to accomplish it should be the following Zed:

```
$ cat from-join-on-uid-conn-fused.zed 
from (
  => file conn.log | sort uid
  => file spl.log | sort uid
) | left join on uid=uid orig_spl,resp_spl,orig_spt,resp_spt | fuse
```

That is, the reference to input files is right there in the Zed, such that you could invoke `zq` without mentioning them on the command line, something like:

```
$ zq -z -I file-join-on-uid-conn-fused.zed
```

When attempted today with `zq` commit 60c70f2, the help text is just echo'ed back, as if it's still requiring filenames on the command line. As @mccanne remarked upon seeing this:

> Yeah, that’s a bug in the zq command-line logic.  Once that is fixed, I think it will work.

nwt:
Closes #2715.

### DIFF
--- a/cmd/zed/query/command.go
+++ b/cmd/zed/query/command.go
@@ -131,12 +131,6 @@ func (c *Command) Run(args []string) error {
 		return err
 	}
 	defer cleanup()
-	if len(args) == 0 {
-		return charm.NeedHelp
-	}
-	if err != nil {
-		return err
-	}
 	paths, query, err := ParseSourcesAndInputs(args, c.includes)
 	if err != nil {
 		return fmt.Errorf("zq: %w", err)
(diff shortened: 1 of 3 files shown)


mccanne (approved):
What about 
```
$ zq -z "from in.zson | pass"
zq: no such file: from in.zson | pass
```

mccanne:
> What about
> 
> ```
> $ zq -z "from in.zson | pass"
> zq: no such file: from in.zson | pass
> ```

I meant `file in.zson | pass` and created this issue #2754 so no need to do this here...

Issue #2756, opened 2021-05-21
optimizer should implement real dataflow analysis


mccanne:
The optimizer should implement a dataflow analysis that is general and more comprehensive than the current ad hoc analysis.  This will be needed when we rely upon sorts (i.e. sql order by) at the output to determine when we can omit the sort because it is derived from the scan order, and when we can do an unordered scan because the flowgraph does an ordered input.

See the scoreboard implementation in analyzeCuts() for a start at this...

mccanne:
A better algorithm would compute USE/DEF sets at each node in the flowgraph.  The USE set would indicate an ordering requirement and the DEF set would define ordering knowledge.

Set operators could flow this info through parallel constructs.

mccanne:
These tests should be reenabled as part of this task:
```
compiler/ztests/par-put-ts-rename-count.yaml
compiler/ztests/par-put-ts-rename-sort.yaml
compiler/ztests/par-put-ts-rename.yaml
```

PR #2760, opened 2021-05-21
fix dataflow through cut operators

### LINKED ISSUE #2663: fix dataflow bug when cutting layout key
This doesn't seem to do the right thing...
```
zed compile -C -P 2 "from 'pool-ts' | cut x=ts,ts=1"
```
which gives...
```
from (
  G2eDzBUfU6IEmUSGCa5kHyXMhoO =>
    cut x=ts,ts=1;
  G2eDzBUfU6IEmUSGCa5kHyXMhoO =>
    cut x=ts,ts=1;
)
| merge ts
```
The merge here should be on `x`.  I believe this bug was in the original parallelization code so we will put the hazmat PR up and fix this later.

mccanne:
This commit fixes a problem where a query would be incorrectly
parallelized by merging on the wrong key because sufficient
dataflow info was not properly tracked.  While we fixed this
we broke another optimization that doesn't seem important to
maintain (and anyway was based on flawed logic).  These tests
are marked "skip".  When we implement proper dataflow analysis
as part of issue #2756, we can turn the tests back on.

Closes #2663

### DIFF
--- a/compiler/optimizer/op.go
+++ b/compiler/optimizer/op.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 
 	"github.com/brimdata/zed/compiler/ast/dag"
+	"github.com/brimdata/zed/compiler/ast/zed"
 	"github.com/brimdata/zed/field"
 	"github.com/brimdata/zed/order"
 )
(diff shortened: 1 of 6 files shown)


compiler/optimizer/op.go
```diff
@@ -140,22 +141,83 @@ func analyzeCuts(assignments []dag.Assignment, layout order.Layout) order.Layout
...
+			for _, d := range dependencies {
+				key := fieldKey(d)
+				if _, ok := scoreboard[key]; ok {
+					// There's a dependency on an ordered
+					// field but we're not sophisticated
+					// enough here to know if this preserves
+					// its order...
+					return order.Nil
+				}
+			}
```
nwt, line 187: I'm having trouble understanding why this is necessary. It seems like deleting `lhsKey` and continuing on should always be correct (if conservative), regardless of what's inside `a.RHS`. Do you have a motivating example?

compiler/optimizer/op.go
```diff
@@ -140,22 +141,83 @@ func analyzeCuts(assignments []dag.Assignment, layout order.Layout) order.Layout
...
+			// opportunities, e.g., we could do dependency
+			// analysis of a complex RHS expression.
 			return order.Nil
 		}
-		if lhs.Equal(key) {
-			return layout
+		lhsKey := fieldKey(lhs)
+		if rhs == nil {
+			// If the RHS depends on a well-defined set of fields
+			// (non of which are unambiguous like this.foo[this.bar]),
```
nwt, line 169:
```suggestion
			// (none of which are unambiguous like this.foo[this.bar]),
```

compiler/optimizer/op.go
```diff
@@ -140,22 +141,83 @@ func analyzeCuts(assignments []dag.Assignment, layout order.Layout) order.Layout
...
-		return order.Layout{Keys: field.List{key}, Order: layout.Order}
 	}
-	return order.Nil
+	if len(scoreboard) != 1 {
+		return order.Nil
+	}
+	for _, f := range scoreboard {
+		return order.Layout{Keys: field.List{f}, Order: layout.Order}
+	}
+	panic("unreachable")
```
nwt, line 211: Not sure this is worth the trouble.
```suggestion
	if len(scoreboard) == 0 {
		return order.Nil
	}
	var maxKey string
	for key := range scoreboard {
		if key > maxKey {
			maxKey = key
		}
	}
	return order.Layout{Keys: field.List{scoreboard[maxkey]}, Order: layout.Order}
```

nwt (approved):

compiler/optimizer/op.go
mccanne: It's not conservative to delete something from the scoreboard because to be conservative you need to know there is just one known ordering constraint because the returned layout can only represent one field.  This was essentially the bug in the previous logic.  This will be clarified with the USE-DEF set dataflow analysis when we do this properly.

compiler/optimizer/op.go
mccanne: This seems more complicated to me.

compiler/optimizer/op.go
nwt: > This was essentially the bug in the previous logic.

But the tests pass without this!

compiler/optimizer/op.go
nwt: I meant that I'm not sure my suggestion is worth the trouble because it's definitely more complicated, but its incorrectness is definitely another strike against it!

PR #2761, opened 2021-05-21
expr: fix some comparisons against null

### LINKED ISSUE #2147: len() of a null array is evaluating to > 0 (regression at 9be1403)
Repro is with `zq` commit `9506d27b`.

Consider the following ZSON, where we've got one record containing a populated array and the other with a `null` array.

```
$ cat answers.zson 
{
    answers: [
        "ise.wrccdc.cpp.edu" (bstring),
        "134.71.3.16" (bstring)
    ] (=0)
} (=1)
{
    answers: null
} (1)
```

The `filter` in the following Z is letting through the record with the `null` array, which is unexpected.

```
$ zq -version
Version: v0.29.0-28-g9506d27b

$ zq -i zson -t 'put len_of_answers=len(answers) | filter len(answers) > 0' answers.zson
#0:record[answers:array[bstring],len_of_answers:int64]
0:[[ise.wrccdc.cpp.edu;134.71.3.16;]2;]
0:[-;-;]
```

Interestingly, if I `filter` on the `null` that was placed into `len_of_answers`, I see the expected behavior. So it's as if the stored and computed `null` values are different in some way.

```
$ zq -i zson -t 'put len_of_answers=len(answers) | filter len_of_answers > 0' answers.zson
#0:record[answers:array[bstring],len_of_answers:int64]
0:[[ise.wrccdc.cpp.edu;134.71.3.16;]2;]
```

This change of behavior came with the changes in #2072. If I circle back to `zq` tagged `v0.29.0` that came right before that PR, I see the expected behavior.

```
$ zq -version
Version: v0.29.0

$ zq -i zson -t 'put len_of_answers=len(answers) | filter len(answers) > 0' answers.zson
#0:record[answers:array[bstring],len_of_answers:int64]
0:[[ise.wrccdc.cpp.edu;134.71.3.16;]2;]
```

Incidentally, the reason I stumbled onto this is because [an example Z in the `by` grouping docs](https://github.com/brimsec/zq/tree/master/zql/docs/grouping#example-3) uses a query like this to isolate the responses from DNS queries:

```
zq -f table 'len(answers) > 0 | count() by id.resp_h,num_answers=len(answers) | sort -r num_answers,count' dns.log.gz
```

nwt:
Fixes #2147.

### DIFF
--- a/expr/coerce/coerce.go
+++ b/expr/coerce/coerce.go
@@ -55,6 +55,12 @@ func (c *Pair) Coerce(a, b zng.Value) (int, error) {
 	if aid == bid {
 		return aid, nil
 	}
+	if aid == zng.IDNull {
+		return bid, nil
+	}
+	if bid == zng.IDNull {
+		return aid, nil
+	}
 	if zng.IsNumber(aid) {
 		if !zng.IsNumber(bid) {
 			return 0, ErrIncompatibleTypes
@@ -68,12 +74,6 @@ func (c *Pair) Coerce(a, b zng.Value) (int, error) {
 			id = zng.IDBstring
 		}
 		return id, nil
-	}
-	if aid == zng.IDNull {
-		return bid, nil
-	}
-	if bid == zng.IDNull {
-		return aid, nil
 	}
 	return 0, ErrIncompatibleTypes
 }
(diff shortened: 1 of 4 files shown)


mccanne (approved):
oopsies

PR #2763, opened 2021-05-22
fix read/write race in journal.Queue

### LINKED ISSUE #2699: Error messages during heavy concurrent "zed lake load"
Repro is with Zed commit b70cbfb. This is a repeat of a torture test similar to one that helped isolate #2614, which has since been fixed.

If you've got the [Zeek TSV zed-sample-data](https://github.com/brimdata/zed-sample-data/tree/main/zeek-default) cloned locally, execute these commands while in the `zed-sample-data/zeek-default/` directory. It'll initialize an empty lake+pool and start populating it with repeated concurrent loads of the full set.

```
export ZED_LAKE_ROOT=$HOME/logs
rm -rf $ZED_LAKE_ROOT
zed lake init
zed lake create -p default -orderby ts:desc
while true; do for file in *; do   zed lake load -p default $file & done; wait; done
```

The load operations seem to complete successfully, but if I carefully watch the output I frequently see error messages appear that look like:

```
strconv.ParseUint: parsing "": invalid syntax
```

Or ones like this:

```
file:///Users/phil/logs/1sMN3OLhI34rx4kAZGD1cCkIUsC/L/J: no such journal
```

I've only seen these appear when I background all the load operations with `&` as shown above. I don't know if these are likely to ever show up during less abusive workloads. However, if they occasionally _did_, I'm guessing a user would start to wonder if there was data corruption, a bug, etc., so it would probably help for us to at least know what's going on, since due to its intermittent nature I can't help but suspect some kind of race.

(Side note: @mccanne did mention in https://github.com/brimdata/zed/issues/2614#issuecomment-825174187 that, at least as of the writing of that comment, "transactional consistency [was] not yet implemented", so perhaps that still explains it. However, I know the storage layer was recently re-touched in #2615 and I wasn't sure if it might make these errors magically disappear, but it seems they didn't. So whether this issue should be attached to some future work around transactional consistency or pursued as a one-off, I figured I'd open it so we don't forget.)

mccanne:
When reading HEAD, the journal can read the contents while
a writer has opened and truncated the contents but before the
writer has written the updated HEAD.  This happens only on
file systems, not cloud objects, as cloud writes are atomic.

The fix is to simply retry the read when the read is successful
but the content cannot be parsed.  This fix is preferred over
a more complicated locking model as the HEAD read is just advisory
and not required for correctness of the log abstraction.

Closes #2699

### DIFF
--- a/lake/journal/queue.go
+++ b/lake/journal/queue.go
@@ -142,7 +142,7 @@ func readID(ctx context.Context, engine storage.Engine, path *storage.URI) (ID, 
 		}
 		retry++
 		if retry > MaxReadRetry {
-			return Nil, errors.New("can read but not parse contents of journal HEAD")
+			return Nil, fmt.Errorf("can read but not parse contents of journal HEAD: %s", string(b))
 		}
 		time.Sleep(timeout)
 		timeout *= 2
--- a/lake/journal/queue.go
+++ b/lake/journal/queue.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/brimdata/zed/pkg/byteconv"
 	"github.com/brimdata/zed/pkg/storage"
@@ -23,6 +24,8 @@ var (
 type ID uint64
 
 const Nil ID = 0
+
+const MaxReadRetry = 10
 
 type Queue struct {
 	engine   storage.Engine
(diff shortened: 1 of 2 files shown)


philrz:
I just ran the torture test described at [https://github.com/brimdata/zed/issues/2699#issue-885050213](https://github.com/brimdata/zed/issues/2699#issue-885050213) against this branch (commit 338c420) for several minutes and no longer saw either of the error messages described in #2699, whereas in the past the errors would show up every few seconds. So, functionally, LGTM! :+1:

lake/journal/queue.go
```diff
@@ -127,15 +131,24 @@ func writeID(ctx context.Context, engine storage.Engine, u *storage.URI, id ID)
...
+			return Nil, err
+		}
+		if id, err := byteconv.ParseUint64(b); err == nil {
+			return ID(id), nil
+		}
+		retry++
+		if retry > MaxReadRetry || timeout > 5*time.Second {
+			return Nil, fmt.Errorf("can read but not parse contents of journal HEAD: %s", string(b))
+		}
+		time.Sleep(timeout)
```
nwt, line 148:
```suggestion
		select {
		case <-time.After(timeout):
		case <-ctx.Done():
			return nil, ctx.Err()
		}
```

nwt (approved):

PR #2771, opened 2021-05-26
rework TestMarkdownExamples to use bash

nwt:
TestMarkdownExaamples can only run zq, and only against files in
zed-sample-data/zeek-default/. Rework it so zq-command blocks can
contain arbitrary commands executed by "bash -e -o pipefail" with a
working directory that defaults to the repository root but can be
specified with a word after zq-command in a block's info string.

We should remove zq from the names of lots of things touched here,
but that can happen later.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,7 @@ test-run: build bin/minio
 	@ZTEST_PATH="$(CURDIR)/dist:$(CURDIR)/bin" go test . -run $(TEST)
 
 test-heavy: build $(SAMPLEDATA)
-	@go test -tags=heavy ./tests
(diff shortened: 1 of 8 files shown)


philrz:
Looks good to me so far! Seeing the paths written out in the long form makes me realize we could probably stand to shorten some stuff (e.g. remove  the leading `zed-` in the sample data repo name, and maybe call the directory with the Zeek TSV logs  `tsv` rather than `zeek-default`,  etc.) but I'm happy to deal with that after this merges and we're working on moving things around to make way for another sample data set.

philrz:
@nwt: With your latest changes, you've managed to increase the awesomeness. Thanks! :tada:

mattnibs (approved):

PR #2774, opened 2021-05-28
Reintroduce zapi command

mattnibs:
Reintroduce a new version of the zapi command the uses the exact same
Commands that the zed lake command uses. The idea is to have functional
parity between zapi and zed lake.


cli/lakecli/lakecli.go
```diff
@@ -0,0 +1,126 @@
...
+	// the data will be committed to the pool.
+	Add(ctx context.Context, r zio.Reader, commit *api.CommitRequest) (ksuid.KSUID, error)
+	Delete(ctx context.Context, tags []ksuid.KSUID, commit *api.CommitRequest) (ksuid.KSUID, error)
+	Commit(ctx context.Context, id ksuid.KSUID, commit api.CommitRequest) error
+	ScanLog(ctx context.Context, w zio.Writer, head, tail journal.ID) error
+	ScanStaging(ctx context.Context, w zio.Writer, ids []ksuid.KSUID) error
+	ScanSegments(ctx context.Context, w zio.Writer, at string, partitions bool, span extent.Span) error
+	Squash(ctx context.Context, ids []ksuid.KSUID, commit api.CommitRequest) (ksuid.KSUID, error)
+	Index(ctx context.Context, rules []index.Index, ids []ksuid.KSUID) (ksuid.KSUID, error)
+}
```
nwt, line 57: https://github.com/golang/go/wiki/CodeReviewComments#interfaces says interfaces generally belong in the package using values of that type. Can we follow that with these?

cli/lakecli/lakecli.go
```diff
@@ -0,0 +1,126 @@
...
+		var err error
+		id, err = ksuid.Parse(s)
+		if err != nil {
+			return ksuid.Nil, fmt.Errorf("%s: invalid commit ID", s)
+		}
+	}
+	return id, nil
+}
+
+func ParseIDs(args []string) ([]ksuid.KSUID, error) {
```
nwt, line 95: Nit: Maybe call this `ss` for symmetry with `ParseID`'s `s` parameter.
```suggestion
func ParseIDs(ss []string) ([]ksuid.KSUID, error) {
```

cmd/zed/lake/ztests/log.yaml
```diff
@@ -0,0 +1,33 @@
+script: |
+  export ZED_LAKE_ROOT=test
+  zed lake init -q
+  zed lake create -q -p POOL
+  zed lake load -date 2021-05-28T15:13:42Z -user testuser -q -p POOL 1.zson
+  zed lake load -date 2021-05-28T15:13:43Z -user testuser -q -p POOL 2.zson
```
nwt, line 6: Nit: This is a little easier to read if `-q -p POOL` always comes first.
```suggestion
  zed lake load -q -p POOL -date 2021-05-28T15:13:42Z -user testuser 1.zson
  zed lake load -q -p POOL -date 2021-05-28T15:13:43Z -user testuser 2.zson
```

lake/pool.go
```diff
@@ -327,7 +327,7 @@ func (p *Pool) ScanSegments(ctx context.Context, w zio.Writer, snap *commit.Snap
 		close(ch)
 	}()
 	m := zson.NewZNGMarshaler()
-	m.Decorate(zson.StyleSimple)
+	m.Decorate(zson.StylePackage)
```
nwt, line 330: Is this important?

service/client_test.go
```diff
@@ -29,6 +29,10 @@ func (c *testClient) unmarshal(r *client.ReadCloser, i interface{}) {
 	zw := zsonio.NewWriter(zio.NopCloser(&buf), zsonio.WriterOpts{})
 	require.NoError(c, zio.Copy(zw, zr))
 	require.NoError(c, zw.Close())
+	str := buf.String()
+	if str == "" {
+		return
+	}
 	require.NoError(c, zson.Unmarshal(buf.String(), i))
```
nwt, line 36:
```suggestion
	if s := buf.String(); s != "" {
		require.NoError(c, zson.Unmarshal(s, i))
	}
```

cmd/zed/lake/ztests/log.yaml
```diff
@@ -0,0 +1,33 @@
+script: |
+  export ZED_LAKE_ROOT=test
+  zed lake init -q
+  zed lake create -q -p POOL
```
nwt, line 4: Existing tests almost always use a pool named test, and I think we should stick with that convention.

lake/pool.go
mattnibs: There's a collision between index.Reference and segment.Reference- needs to include package scope to differentiate between the two.

cmd/zed/lake/delete/command.go
```diff
@@ -61,40 +61,38 @@ func (c *Command) Run(args []string) error {
...
+		return err
+	}
+	if !c.lake.Flags.Quiet() {
+		fmt.Printf("%s staged to delete\n", commitID)
+		// XXX
+		// txn, err := pool.LoadFromStaging(ctx, commitID)
+		// if err != nil {
+		// return err
+		// }
+		// fmt.Printf("%s staged to delete %d segments\n", commitID, len(txn.Actions))
```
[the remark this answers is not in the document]
mattnibs: I don't think so for the near future. This isn't supported by the service api right now.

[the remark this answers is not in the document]
mattnibs: Removed.

cli/lakecli/lakecli.go
mattnibs: I take it you are suggesting moving the above interfaces to ./cmd/zed/lake?

I tried doing this and it doesn't work because of return functions like: 
```
CreatePool(context.Context, string, order.Layout, int64) (Pool, error)
```
If I implement it like thus in lakecli:
```
CreatePool(context.Context, string, order.Layout, int64) (*LocalPool, error)
```
Go complains that it doesn't implement the Root interface. Open to suggestions...

cmd/zed/lake/ztests/log.yaml
mattnibs: Sure

mccanne (approved):

Issue #2545, opened 2021-04-13
zed lake vacate & purge


mccanne:
add `zed lake vacate` command stubbed out and described in cmd/zed/lake/vacate/command.go

The [Zed lake design doc](https://github.com/brimdata/zed/blob/main/docs/lake/design.md) also proposes a `zed lake purge` command that's subtly different from `vacate`, and that one's not yet implemented either. So as part of the design we'll surely figure out if both have to exist and implement accordingly.

philrz:
Note to self: When this gets implemented, update the Zed lake design doc to remove the reference to this being an open isssue.

philrz:
Closing as a duplicate of #4113

philrz:
I mistakenly thought `vacuum` did the same thing envisioned by `vacate` so had closed this one in error. Reopening.

philrz:
`super db vacate` is now available as a sub-command and works as documented [here](https://superdb.org/command/db.html#super-db-vacate).

Issue #2547, opened 2021-04-13
lake config should be stored in a journal


mccanne:
The lake config is currently stored in the pools.json static file that has write-concurrency races.
We can use the journal abstraction to make updates ACID and have a history of changes that users can review.

mccanne:
@mattnibs is going to store lake-wide index rules in lake.Root and this will also be converted into the lake config journal when this issue is addressed.

PR #2781, opened 2021-06-02
Add "zed lake init", change pool key spec, and .-to-this conversion

philrz:
Some users on public Slack expressed an interest in beta testing some of the new stuff, at which point they bumped into some out-of-date/missing stuff in the Zed lake design doc. I've addressed the first wave of speedbumps here by:

*   Adding coverage of `zed lake init`
*   Adapting to the new way pool keys are specified
*   Converting `.` references to `this`
*   Updates to `zed query` syntax now that pool specification and key range have moved into the Zed

I've also started linking to open GitHub Issues for the examples we show that aren't yet implemented. We've got stuff in the tooling that already tip the user off about this (e.g. if I execute `zed lake vacate` it responds with `"issue #2545"`) but since we've got users trying to follow along with these examples in beta testing mode, I figure it'd be helpful to save them from trying to craft what they believe is the correct sequence of flags and commits only to then find out that it's not gonna work. If reviewers are on-board with this approach, I'll add notes-to-self in the linked issues with reminders to go back and update this doc as they get implemented.

### DIFF
--- a/docs/lake/design.md
+++ b/docs/lake/design.md
@@ -154,7 +154,7 @@ time range using `over` and `to`.  The default output format is ZNG though this
 can be overridden with `-f` to specify one of the various supported output
 formats.
 
-This example reads every record from the full time range of the `logs` pool
+This example reads every record from the full key range of the `logs` pool
 and sends the results as ZSON to stdout.
 
 ```
--- a/docs/lake/design.md
+++ b/docs/lake/design.md
@@ -382,13 +382,14 @@ commit journal without having to run an explicit `commit` command.
 
 ### Purge and Vacate
 
-Data can be deleted with the DANGER-ZONE command `zed lake purge`.
+Data can be deleted with the DANGER-ZONE command `zed lake purge`
+(implementation tracked in [zed/2545](https://github.com/brimdata/zed/issues/2545)).
 The commits still appear in the log but scans at any time-travel point
 where the commit is present will fail to scan the deleted data.
 
 Alternatively, old data can be removed from the system using a safer
-command (but still in the DANGER-ZONE), `zed lake vacate` (implementation
-tracked in [zed/2545](https://github.com/brimdata/zed/issues/2545)) which moves
+command (but still in the DANGER-ZONE), `zed lake vacate` (also
+[zed/2545](https://github.com/brimdata/zed/issues/2545)) which moves
 the tail of the commit journal forward and removes any data no longer
 accessible through the modified commit journal.
 
--- a/docs/lake/design.md
+++ b/docs/lake/design.md
@@ -437,8 +437,7 @@ and historical content, and accessing its complete detail can provide
 insights about data layout, provenance, history, and so forth.  Thus,
 Zed lake provides a means to query a pool's entire journal in all its
 detail.  To do so, simply query a pool's journal by referring to
-the special sub-pool name `<pool>:journal` (planned implementation tracked in
-[zed/2787](https://github.com/brimdata/zed/issues/2787)):
+the special sub-pool name `<pool>:journal`.
 
 For example, to aggregate a count of each journal entry type of the pool
 called `logs`, you can simply say:
--- a/docs/lake/design.md
+++ b/docs/lake/design.md
@@ -148,27 +148,35 @@ and arbitrary data _shapes_ can coexist side by side.
 
 ### Query
 
-Data is read from a pool with the `query` command.
-A range can be specified with `-from` and/or `-to`.
-The default output format is ZNG though this can be overridden with `-f`
-to specify one of the various supported output formats.
+Data is read from one or more pools with the `query` command.  The pool names
+are specified with `from` at the beginning the Zed query along with an optional
+time range using `over` and `to`.  The default output format is ZNG though this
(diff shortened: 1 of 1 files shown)


docs/lake/design.md
```diff
@@ -140,27 +148,35 @@ and arbitrary data _shapes_ can coexist side by side.
...
+and sends the results as ZSON to stdout.
+
+```
+zed lake query -f zson 'from logs'
 ```
-zed lake query -p logs -f zson -from 2020-1-1T12:00
+
+Or we can narrow the span of the query by specifying the key range.
 ```
+zed lake query -z 'from logs over 2018-03-24T17:36:30.090766Z to 2018-03-24T17:36:30.090758Z'
```
philrz, line 166: @mccanne: I'm effectively explaining this based on my black box testing and what I see in the ztests, so I'd appreciate your confirmation here that I'm not misrepresenting anything. From what I can tell, now that it's in the Zed, the key range:

1.  Must conform to a more strict format (i.e., the relaxed timestamp format like `2020-1-1T12:00` that previously worked when it was a `-from` parameter is now a syntax error if I try to use it inside the Zed)
2.  If I want to specify a key range in the Zed, it seems like I need to specify both the `over` _and_ `to`, whereas the prior example showed only the `-from`.

I'm not implying either of these are _problems_. But since they represent a change in functionality I just want to make sure they were intentional and not surprises.

mattnibs (approved):

docs/lake/design.md
```diff
@@ -121,13 +129,12 @@ Data is then loaded into a lake with the `load` command, .e.g.,
 zed lake load -p logs sample.ndjson
 ```
 where `sample.ndjson` contains logs in NDJSON format.  Any supported format
-(i.e., CSV, JSON, NDJSON, Parquet, ZNG, and ZST) as well multiple files can be used
-here, e.g.,
+(NDJSON, ZNG, ZSON, etc.) as well multiple files can be used here, e.g.,
 ```
-zed lake load -p logs sample1.csv sample2.ndjson sample3.zng
+zed lake load -p logs sample1.ndjson sample2.zng sample3.zson
```
philrz, line 134: The example that was here previously wouldn't work because CSV is _not_ auto-detected currently, and it looks like when specifying `-i` we only accept one format at a time. I therefore replaced the prior "i.e.," (which read as if the list that followed was complete) with the closing ", etc.") (which, to me, reads more like an incomplete list, which I prefer). In addition to encouraging the reader to read on and learn about the `-i` needed for the other supported formats, I figure this hand-waves at the not-explicitly-mentioned supported Zeek TSV format. I respect the purity of not calling out Zeek explicitly, but in the short term a lot of the users are likely to be importing that, so I'm kinda playing both sides here. Thankfully most users who've come to this doc (such as the ones I'm currently engaging with on the #beta channel on Slack) already use `zq` and have come to expect Zeek TSV will "just work", so I'm not too concerned here.

docs/lake/design.md
```diff
@@ -213,7 +229,7 @@ zed lake commit -p logs <tag-1> <tag-2> <tag-3>
 The commit command also takes an optional title and message that is stored
 in the commit journal for reference.  For example,
 ```
-zed lake commit -p logs -u user@example.com -m "new version of prod dataset" <tag>
+zed lake commit -p logs -user user@example.com -message "new version of prod dataset" <tag>
```
philrz, line 232: I found via black box testing that `-u` and `-m` are not accepted. I'm not opposed to adding them, but for now I've updated this to reflect what works.

docs/lake/design.md
```diff
@@ -362,7 +387,8 @@ The commits still appear in the log but scans at any time-travel point
 where the commit is present will fail to scan the deleted data.
 
 Alternatively, old data can be removed from the system using a safer
-command (but still in the DANGER-ZONE), `zed lake vacate`, which moves
+command (but still in the DANGER-ZONE), `zed lake vacate` (implementation
+tracked in [zed/2545](https://github.com/brimdata/zed/issues/2545)) which moves
```
philrz, line 391: While I found the open issue for `zed lake vacate` (which is already shown in `zed lake -h` output) and linked to it, I see `zed lake purge` (mentioned a few lines above) is not yet in the `-h` output nor could I find an open issue. @mccanne: Is this still one we'd like to implement? I'm happy to open an additional issue, but figured I'd check first.

docs/lake/design.md
```diff
@@ -416,7 +442,7 @@ the special sub-pool name `<pool>:journal`.
 For example, to aggregate a count of each journal entry type of the pool
 called `logs`, you can simply say:
 ```
-zed lake query -p logs:journal "count() by typeof(.)"
+zed lake query "from logs:journal | count() by typeof(this)"
```
philrz, line 445: I updated this to phase out the `-p` flag since we know that's coming back, but (per the note just below this example) this journal sub-pool query isn't implemented yet, so I'm basically guessing this will still be the future syntax. Also, I'd like to do my link-to-the-issue-not-yet-implemented thing in that note below, but I'm not sure if #2547 is the correct one or if we have to open a new one. @mccanne could you confirm?

mccanne (approved):

PR #2788, opened 2021-06-04
Don't log error on empty search

mattnibs:


nwt (approved):
Add a test?

mattnibs:
Kind of hard to add a tests here since you would be testing logger output, which might lead to an unstable test.

PR #2791, opened 2021-06-07
Events endpoint

mason-fish:
Adds an `/events` endpoint which will provide clients/subscribers with event updates via the [Server-sent events](https://html.spec.whatwg.org/multipage/server-sent-events.html) mechanism. For now, the implementation provide the following events: `pool-new`, `pool-update`, `pool-delete`, and `pool-commit`. Upon receiving these events the app can then decide how/when it wants to take more action (i.e. fetch new data).

Signed-off-by: Mason Fish <mason@looky.cloud>

### DIFF
--- a/service/core.go
+++ b/service/core.go
@@ -8,7 +8,6 @@ import (
 	"io"
 	"net/http"
 	"net/http/pprof"
-	"strings"
 	"sync"
 	"sync/atomic"
 
@@ -149,7 +148,7 @@ func (c *Core) addAPIServerRoutes() {
 	c.authhandle("/pool/{pool}/staging", handleScanStaging).Methods("GET")
 	c.authhandle("/pool/{pool}/staging/{commit}", handleCommit).Methods("POST")
 	c.authhandle("/pool/{pool}/stats", handlePoolStats).Methods("GET")
-	c.authhandle("/subscribe", handleSubscribe).Methods("GET")
+	c.authhandle("/events", handleEvents).Methods("GET")
 
 	// Deprecated endpoints
 	c.authhandle("/search", handleSearch).Methods("POST")
@@ -200,20 +199,13 @@ func (c *Core) requestLogger(r *http.Request) *zap.Logger {
 	return c.logger.With(zap.String("request_id", api.RequestIDFromContext(r.Context())))
 }
 
-func (c *Core) publishEvent(event string, data string) {
-	eventPayload := fmt.Sprintf("event: %s\n", event)
-	dataLines := strings.Split(data, "\n")
-
-	for _, line := range dataLines {
-		eventPayload += fmt.Sprintf("data: %s\n", line)
-	}
-
+func (c *Core) publishEvent(event string, data []byte) {
+	payload := fmt.Sprintf("event: %s\ndata: %s\n\n", event, string(data))
 	go func() {
 		c.subscriptionsMu.RLock()
-		defer c.subscriptionsMu.RUnlock()
-
 		for sub := range c.subscriptions {
-			sub <- []byte(eventPayload + "\n")
+			sub <- []byte(payload)
 		}
+		c.subscriptionsMu.RUnlock()
 	}()
 }
--- a/service/core.go
+++ b/service/core.go
@@ -41,16 +41,16 @@ type Config struct {
 }
 
 type Core struct {
-	auth      *Auth0Authenticator
-	conf      Config
-	engine    storage.Engine
-	logger    *zap.Logger
-	registry  *prometheus.Registry
-	root      *lake.Root
-	routerAPI *mux.Router
-	routerAux *mux.Router
-	taskCount int64
-	subscriptions map[chan []byte]struct{}
+	auth            *Auth0Authenticator
+	conf            Config
+	engine          storage.Engine
+	logger          *zap.Logger
+	registry        *prometheus.Registry
+	root            *lake.Root
+	routerAPI       *mux.Router
+	routerAux       *mux.Router
+	taskCount       int64
+	subscriptions   map[chan []byte]struct{}
 	subscriptionsMu sync.RWMutex
 }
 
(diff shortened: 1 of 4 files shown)


service/handlers.go
```diff
@@ -177,6 +177,8 @@ func handlePoolPost(c *Core, w *ResponseWriter, r *Request) {
 		w.Error(err)
 		return
 	}
+
+	c.publishEvent("new-pool", pool.ID.String())
```
mattnibs, line 181: Shouldn't there also be an event for put pool (ie update name) and delete pool?

service/handlers.go
```diff
@@ -271,6 +273,19 @@ func handleCommit(c *Core, w *ResponseWriter, r *Request) {
...
 		w.Error(err)
 	}
+
+	b, err := json.Marshal(struct {
+		CommitID string `json:"commit_id"`
+		PoolID   string `json:"pool_id"`
+	}{
+		CommitID: commitID.String(),
+		PoolID:   pool.ID.String(),
+	})
```
mattnibs, line 283:
```suggestion
	b, err := json.Marshal(struct {
		CommitID ksuid.KSUID `json:"commit_id"`
		PoolID   ksuid.KSUID `json:"pool_id"`
	}{
		CommitID: commitID,
		PoolID:   pool.ID,
	})
```

service/core.go
```diff
@@ -144,6 +149,7 @@ func (c *Core) addAPIServerRoutes() {
 	c.authhandle("/pool/{pool}/staging", handleScanStaging).Methods("GET")
 	c.authhandle("/pool/{pool}/staging/{commit}", handleCommit).Methods("POST")
 	c.authhandle("/pool/{pool}/stats", handlePoolStats).Methods("GET")
+	c.authhandle("/subscribe", handleSubscribe).Methods("GET")
```
mattnibs, line 152: I think its generally better (when possible) for rest endpoints to be nouns rather than verbs. Consider calling this `/events` or `/activity`?

service/handlers.go
mattnibs: Also I don't think its good that the payload is an inconsistent content type. I would just stick with json.

service/handlers.go
mattnibs: We keep all the api responses in the api package. I would create a struct there called `EventPoolCommit`.

service/core.go
```diff
@@ -193,3 +199,21 @@ func (c *Core) nextTaskID() int64 {
...
+		eventPayload += fmt.Sprintf("data: %s\n", line)
+	}
+
+	go func() {
+		c.subscriptionsMu.RLock()
+		defer c.subscriptionsMu.RUnlock()
+
+		for sub := range c.subscriptions {
+			sub <- []byte(eventPayload + "\n")
+		}
```
mattnibs, line 217: No need for the defer.
```suggestion
		c.subscriptionsMu.RLock()
		for sub := range c.subscriptions {
			sub <- []byte(eventPayload + "\n")
		}
		c.subscriptionsMu.RUnLock()
```

service/core.go
```diff
@@ -193,3 +199,21 @@ func (c *Core) nextTaskID() int64 {
 func (c *Core) requestLogger(r *http.Request) *zap.Logger {
 	return c.logger.With(zap.String("request_id", api.RequestIDFromContext(r.Context())))
 }
+
+func (c *Core) publishEvent(event string, data string) {
+	eventPayload := fmt.Sprintf("event: %s\n", event)
+	dataLines := strings.Split(data, "\n")
+
```
mattnibs, line 206: The style for the backend is to not have superfluous line breaks in methods.
```suggestion
```

service/core.go
```diff
@@ -193,3 +199,21 @@ func (c *Core) nextTaskID() int64 {
 func (c *Core) requestLogger(r *http.Request) *zap.Logger {
 	return c.logger.With(zap.String("request_id", api.RequestIDFromContext(r.Context())))
 }
+
+func (c *Core) publishEvent(event string, data string) {
+	eventPayload := fmt.Sprintf("event: %s\n", event)
+	dataLines := strings.Split(data, "\n")
```
mattnibs, line 205: You don't have any multi-line data at this point, I would just keep this simple:
```
payload := fmt.Sprintf("event: %s\ndata: %s", event, data)
```

service/core.go
```diff
@@ -193,3 +199,21 @@ func (c *Core) nextTaskID() int64 {
 func (c *Core) requestLogger(r *http.Request) *zap.Logger {
 	return c.logger.With(zap.String("request_id", api.RequestIDFromContext(r.Context())))
 }
+
+func (c *Core) publishEvent(event string, data string) {
```
mattnibs, line 203:
```suggestion
func (c *Core) publishEvent(event, data string) {
```

mattnibs (commented):
Would you mind documenting this endpoint in docs/service-api.md?

service/core.go
```diff
@@ -193,3 +198,14 @@ func (c *Core) nextTaskID() int64 {
 func (c *Core) requestLogger(r *http.Request) *zap.Logger {
 	return c.logger.With(zap.String("request_id", api.RequestIDFromContext(r.Context())))
 }
+
+func (c *Core) publishEvent(event string, data []byte) {
+	payload := fmt.Sprintf("event: %s\ndata: %s\n\n", event, string(data))
```
nwt, line 203: Nits: Might as well do this in the goroutine. And no need to convert `data` to `string` since it'll be formatted by `%s`.

service/core.go
```diff
@@ -193,3 +198,14 @@ func (c *Core) nextTaskID() int64 {
...
+	go func() {
+		c.subscriptionsMu.RLock()
+		for sub := range c.subscriptions {
+			sub <- []byte(payload)
```
nwt, line 207: I could be wrong here, but I think the compiler is going to allocate a new slice on each trip through this loop, so I'd hoist the conversion to `[]byte` out of the loop (and, while you're at it, outside of the critical section).

service/handlers.go
```diff
@@ -177,6 +177,14 @@ func handlePoolPost(c *Core, w *ResponseWriter, r *Request) {
...
 		return
 	}
+
+	b, err := json.Marshal(api.EventPool{
+		PoolID: pool.ID.String(),
+	})
+	if err != nil {
+		w.Error(err)
+	}
+	c.publishEvent("pool-new", b)
```
nwt, line 187: Maybe call `json.Marshal` inside `publishEvent`?
```suggestion
	if err := c.publishEvent("pool-new", api.EventPool{PoolID: pool.ID.String()}); err != nil {
		w.Error(err)
		return
	}
```

nwt (approved):
I think this looks good but leave approval to @mattnibs.

mattnibs (approved):

PR #2792, opened 2021-06-07
zson: add check for match tight

### LINKED ISSUE #2783: "zson syntax error" when reading our own ZSON (regression at 6b17fc49)
Repro is with Zed commit 6b17fc4 using the [Zeek zed-sample-data](https://github.com/brimdata/zed-sample-data/tree/main/zeek-default).

If I read certain Zeek TSV files and output them as ZSON, then read them back as ZSON again, it fails.

```
$ zq -version
Version: v0.29.0-231-g6b17fc49

$ zq -f zson files.log.gz | zq -f zson -i zson 'count()' -
stdio:///stdin: zson syntax error
```

I've narrowed down that this starts happening at a particular threshold.

```
$ zq -f zson 'head 24048' files.log.gz | zq -f zson -i zson 'count()' -
{
    count: 24048 (uint64)
} (=0)

$ zq -f zson 'head 24049' files.log.gz | zq -f zson -i zson 'count()' -
stdio:///stdin: zson syntax error
```

I also narrowed down via binary search that the Zed commit where this started happening is 6b17fc4 that's associated with #2515, which involved a change to the "zson lexer", so that all checks out (cc: @mattnibs).

mattnibs:
If the cursor was empty but underlying stream not empty, match tight
would erroneously return EOF. Add check before match to ensure this does
not happen.

Closes #2783

### DIFF
--- a/zson/lexer.go
+++ b/zson/lexer.go
@@ -106,8 +106,8 @@ func (l *Lexer) match(b byte) (bool, error) {
 }
 
 func (l *Lexer) matchTight(b byte) (bool, error) {
-	if len(l.cursor) == 0 {
-		return false, io.EOF
+	if err := l.check(1); err != nil {
+		return false, err
 	}
 	if b == l.cursor[0] {
 		if err := l.skip(1); err != nil {


nwt (approved):
Is this worth a test? (Honest question. I really don't know.)

philrz:
@nwt: I'm sure you're asking about a micro-level ztest, but FWIW, the nightly test I've put up for review at #2785 would have caught this issue and would hopefully catch the same class of issue in the future if it happens again (that's why I put it up).

Issue #2651, opened 2021-05-03
zapi: Use zed lake commands


mattnibs:
The zapi command should use the same code as zed lake. We will have to add additional api routes to extend functionality of zed lake serve in order to make this happen. The zed lake command will also have to use some interface for lake.Root and lake.Pool so this can be switched out with the api client for zapi.

philrz:
Verified in Zed commit 5569e98 by using a Dev checkout of Brim `5ed0f7f`, which uses this Zed as its backend.

To test, I went through the operations from the [Zed lake design doc](https://github.com/brimdata/zed/blob/main/docs/lake/design.md) using the [Zeek zed-sample-data](https://github.com/brimdata/zed-sample-data/tree/main/zeek-default), but using `zapi` to hit Zed backend over the API rather than the `zed lake` commands shown in the doc that hit the filesystem layer directly.  The `zapi` binary is invoked from inside Brim's `zdeps/` directory.

```
$ ./zapi create -p logs -orderby ts:desc
pool created: logs

$ ./zapi query -z 'from logs | head 1'
{_path:"conn",ts:2018-03-24T17:36:30.158539Z,uid:"ChhAfsfyuz4n2hFMe" (bstring),id:{orig_h:10.239.34.35,orig_p:56602 (port=(uint16)),resp_h:10.47.6.51,resp_p:873 (port)} (=0),proto:"tcp" (=zenum),service:null (bstring),duration:4us,orig_bytes:0 (uint64),resp_bytes:0 (uint64),conn_state:"S0" (bstring),local_orig:null (bool),local_resp:null (bool),missed_bytes:0 (uint64),history:"S" (bstring),orig_pkts:2 (uint64),orig_ip_bytes:88 (uint64),resp_pkts:0 (uint64),resp_ip_bytes:0 (uint64),tunnel_parents:null (1=(|[bstring]|))} (=2)

$ ./zapi query -z 'from logs over 2018-03-24T17:36:30.090766Z to 2018-03-24T17:36:30.090758Z | head 1'
[...]

PR #2794, opened 2021-06-07
query endpoint / zapi query command

mattnibs:
Add new zed lake query endpoint and zapi command that consumes it.

Changes to query endpoint vs deprecated search endpoint:
- Query endpoint accepts zed queries as strings NOT generated ASTs.
- Adjustments to the control message protocol (supported for zng and
  zjson responses):
  - The QueryChannelSet control message indicates the current channel
    for the stream- all subsequent records will be part of the set
    channel until another QueryChannelSet message is received. In the
    old search endpoint, multi channel searches were not supported. They
    are now supported in the new query endpoint.
  - zjson "records" or "objects" are no longer batched. Each record is
    wrapped in an "Object" response.

Part of #2651


api/api.go
```diff
@@ -147,3 +147,30 @@ type IndexPostRequest struct {
 	Patterns []string `json:"patterns"`
 	Zed      string   `json:"zed,omitempty"`
 }
+
+type QueryRequest struct {
+	Query    []string `json:"query"`
+	Includes []string `json:"string"`
```
nwt, line 153:
```suggestion
	Query string `json:"query"`
```

service/handlers.go
```diff
@@ -27,6 +30,39 @@ import (
 	"go.uber.org/zap"
 )
 
+func handleQuery(c *Core, w *ResponseWriter, r *Request) {
+	var req api.QueryRequest
+	if !r.Unmarshal(w, &req) {
+		return
+	}
+	zquery, err := query.ParseSources(req.Query, req.Includes)
```
nwt, line 38: This reads from the file system, so it needs to happen in the client.

mattnibs: Cool

api/queryio/writer.go
```diff
@@ -0,0 +1,89 @@
+package queryio
```
nwt, line 1: Name this file `driver.go`?

api/queryio/client.go
```diff
@@ -0,0 +1,91 @@
...
+	case *api.QueryChannelSet:
+		err = r.flush()
+		r.cid = ctrl.ChannelID
+	case *api.QueryChannelEnd:
+		err = r.driver.ChannelEnd(ctrl.ChannelID)
+	case *api.QueryStats:
+		r.stats.BytesRead = ctrl.ScannerStats.BytesRead
+		r.stats.BytesMatched = ctrl.ScannerStats.BytesMatched
+		r.stats.RecordsRead = ctrl.ScannerStats.RecordsRead
+		r.stats.RecordsMatched = ctrl.ScannerStats.RecordsMatched
```
nwt, line 83:
```suggestion
		r.stats = zbuf.ScannerStats(ctrl.ScannerStats)
```

api/queryio/client.go
```diff
@@ -0,0 +1,91 @@
...
+
+func (r *runner) flush() error {
+	if len(r.recs) > 0 {
+		recs := make([]*zng.Record, len(r.recs))
+		copy(recs, r.recs)
+		r.recs = r.recs[:0]
+		return r.driver.Write(r.cid, zbuf.Array(recs))
+	}
+	return nil
+}
```
nwt, line 69: Maybe simplify by getting rid of `flush` and always handing single-record batches to the driver? We can always get more complicated down the road if the extra allocations become a bottleneck.
```suggestion
func (r *runner) Write(rec *zng.Record) error {
	return r.driver.Write(r.cid, &zbuf.Array{rec})
}
```

nwt (approved):

service/ztests/service.sh
```diff
@@ -22,7 +22,7 @@ mkdir -p $lakeroot
 
 portdir=$(mktemp -d)
 
-zed lake serve -l=localhost:0 -R $lakeroot -portfile=$portdir/lake -log.level=debug $LAKE_EXTRA_FLAGS &> lake.log &
+zed lake serve -l=localhost:0 -R $lakeroot -portfile=$portdir/lake -log.level=warn $LAKE_EXTRA_FLAGS &> lake.log &
```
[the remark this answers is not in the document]
mattnibs: Yeah but should be a different pr.

api/queryio/client.go
```diff
@@ -0,0 +1,74 @@
...
+		}
+		if ctrl != nil {
+			run.handleCtrl(ctrl)
+			continue
+		}
+		if rec != nil {
+			run.Write(rec)
+			continue
+		}
+		return run.stats, run.flush()
```
nwt, line 41:
```suggestion
		if rec == nil {
			return run.stats, nil
		}
		if err := run.Write(rec); err != nil {
			return run.stats, err
		}
```

api/queryio/client.go
```diff
@@ -0,0 +1,74 @@
...
+
+func (r *runner) Write(rec *zng.Record) error {
+	return r.driver.Write(r.cid, &zbuf.Array{rec})
+}
+
+func (r *runner) handleCtrl(ctrl interface{}) error {
+	var err error
+	switch ctrl := ctrl.(type) {
+	case *api.QueryChannelSet:
+		err = r.flush()
```
nwt, line 61:
```suggestion
```

PR #2809, opened 2021-06-17
simplify and fix IDs in zng serializer

mccanne:
There were bugs in some of the zng serialization logic where
zng.Type.ID() was used instead of the zng.TypeID().  This commit
fixes these bugs and simplifies the code.  We also removed the
unused method zngio.Encoder.serializeTypes().

### DIFF
--- a/zio/zngio/ztests/complex-alias.yaml
+++ b/zio/zngio/ztests/complex-alias.yaml
@@ -1,0 +1,13 @@
+# do a round-trip into serialized ZNG to make sure alias encoded properly
+
+script: zq in.zson | zq -z -
+
+inputs:
+  - name: in.zson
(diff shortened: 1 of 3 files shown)


zio/zngio/ztests/complex-alias.yaml
```diff
@@ -0,0 +1,13 @@
+# do a round-trip into serialized ZNG to make sure alias encoded properly
+
+script: zq in.zson | zq -z -
+
+inputs:
+  - name: in.zson
```
nwt, line 6:
```suggestion
script: zq - | zq -z -

inputs:
  - name: stdin
```

nwt (approved):

PR #2810, opened 2021-06-18
pool.ScanStaging: scan all

mattnibs:
If no ids are provided for pool.ScanStaging, scan all objects in
staging.

### DIFF
--- a/lake/pool.go
+++ b/lake/pool.go
@@ -3,6 +3,7 @@ package lake
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"path/filepath"
 	"strings"
@@ -30,6 +31,8 @@ const (
 	LogTag   = "L"
 	StageTag = "S"
 )
+
+var ErrStagingEmpty = errors.New("staging area empty")
 
 type PoolConfig struct {
 	Version   int          `zng:"version"`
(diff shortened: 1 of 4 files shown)


lake/pool.go
```diff
@@ -239,11 +242,22 @@ func (p *Pool) StoreInStaging(ctx context.Context, txn *commit.Transaction) erro
 	return storage.Put(ctx, p.engine, p.StagingObject(txn.ID), bytes.NewReader(b))
 }
 
-func (p *Pool) ScanStaging(ctx context.Context, w zio.Writer, ids []ksuid.KSUID) error {
+// ScanStaging writes the provided commits in staging to the provided
+// zio.Writer.  If not commit tags are passed, all commits in staging are
+// written.
```
nwt, line 247: Nit: I think it's clearer to refer to parameters directly by name rather than indirectly by type or description. (The standard library largely follows this style.)
```suggestion
// ScanStaging writes the staging commits in ids to w.
// If ids is empty, all staging commits are written.
```

lake/pool.go
```diff
@@ -239,11 +242,22 @@ func (p *Pool) StoreInStaging(ctx context.Context, txn *commit.Transaction) erro
...
+func (p *Pool) ScanStaging(ctx context.Context, w zio.Writer, ids []ksuid.KSUID) (err error) {
+	if len(ids) == 0 {
+		ids, err = p.ListStagedCommits(ctx)
```
nwt, line 250: NIt: I think it's clearer to use an explicit declaration here. The named result parameter suggests something more complicated is going on.
```suggestion
func (p *Pool) ScanStaging(ctx context.Context, w zio.Writer, ids []ksuid.KSUID) error {
	if len(ids) == 0 {
		var err error
		ids, err = p.ListStagedCommits(ctx)
```

nwt (approved):
Should this have test coverage?

lake/pool.go
mattnibs: This is much simpler! Thanks!

lake/pool.go
mattnibs: Sure!

mattnibs:
> Should this have test coverage?

@nwt This will get test coverage in #2811 which has a test for zed lake status with commits specified.

PR #2813, opened 2021-06-21
mdtest: change zq- keyword prefix to mdtest-

nwt:
The zq-command and zq-output keywords used in the info strings of fenced
code blocks haven't really fit since the change to allow arbitrary shell
commands.  Now that the package is called mdtest, mdtest-command and
mdtest-output make more sense.

Depends on #2812.


mattnibs (approved):
🙌🏻

philrz (approved):

PR #2814, opened 2021-06-21
mdtest: change head:N to head

nwt:
The line count in "mdtest-output head:N" is unnecessary since it can be
determined from the fenced code block's content, so simplify the syntax
to "mdtest-output head".


philrz:
Neat! I'd gladly accept that simplification. Looks good to me, though I'll let the formal approval come from the code-writers.

mccanne (approved):

PR #2819, opened 2021-06-22
fix bug in zson.String() for null

### LINKED ISSUE #2818: zson.String() problem with null
zson.String() is printing `null (null) (null)` for a null of type null.   We should make type null implied to fix this, which should also fix the extra decorator.

mccanne:
This commit fixes a bug where null(null) was formatted as
"null (null) (null)".  The fix was to make type null "implied"
in package zson.

Closes #2818


nwt (approved):
This definitely makes for better looking ZSON.

philrz:
Verified in Zed commit b28cfe0.

As of Zed commit 6cc190b that was right before the changes in the linked PR, the `(null)` was still explicit.

```
$ zq -version
Version: v0.29.0-392-g6cc190b9

$ echo '{foo:null (null)} (=0)' | zq -z -
{foo:null (null)} (=0)
```

After the changes in the linked PR, it's now an implied type.

```
$ zq -version
Version: v0.29.0-403-gb28cfe0b

$ echo '{foo:null (null)} (=0)' | zq -z -
{foo:null}
```

Thanks @mccanne!

PR #2820, opened 2021-06-22
simplify Zed enums

### LINKED ISSUE #2508: simplify ZSON enum type
The current enum type allows arbitrary ZSON values to be assigned to enum Symbols.
Let's simplify the ZSON enum type to consist solely as a list of symbols.
The symbol syntax shall follow identifiers.
The canonical format for an enum type shall sort the list of symbols alphabetically.
In the zng.TypeEnum struct, change the name of the symbols from Elements to Symbols.

This is a breaking change to ZNG, ZST, and ZSON but nothing uses enum as far as I know.
(The Zed journal will use it.)
The format will not otherwise break if there are no enum types in use.

mccanne:
This commit simplifies Zed enums by removing the values from the
list of symbols.  The benefit of this was always dubious and avoiding
the crossing the type/value boundary (by encoding values in types)
simplifies things quite a bit.

Closes #2508

### DIFF
--- a/docs/formats/zng.md
+++ b/docs/formats/zng.md
@@ -237,19 +237,16 @@ The `<ntypes>` and the type IDs are all encoded as `uvarint`.
 
 #### 2.1.1.5 Enum Typedef
 
-An enum type is encoded as the type code of the enum values as a `uvarint`
-followed by a count of the number of elements in the enum, and in turn,
-followed by the names and values of each element.
+An enum type is encoded as a `uvarint` representing the number of symbols
+in the enumeration followed by the names of each symbol.
 ```
---------------------------------------------------------
-|0xfa|<type-id>|<nelem>|<name1><val-1><name2><val-2>...|
---------------------------------------------------------
+--------------------------------
+|0xfa|<nelem>|<name1><name2>...|
+--------------------------------
 ```
-`<type-id>` and `<nelem>` are encoded as `uvarint`.
+`<nelem>` is encoded as `uvarint`.
 The names have the same UTF-8 format as record field names and are encoded
-as tag-encoded primitive strings.  Each value is encoded as
-a tag-encoded value in accordance with the type indicated
-by `<type-id>`.
+as counted strings following the same convention as record fields.
 
 #### 2.1.1.6 Map Typedef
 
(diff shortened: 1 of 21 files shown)


docs/formats/zng.md
```diff
@@ -237,19 +237,16 @@ The `<ntypes>` and the type IDs are all encoded as `uvarint`.
...
+|0xfa|<nelem>|<name1><name2>...|
+--------------------------------
 ```
-`<type-id>` and `<nelem>` are encoded as `uvarint`.
+`<nelem>` is encoded as `uvarint`.
 The names have the same UTF-8 format as record field names and are encoded
-as tag-encoded primitive strings.  Each value is encoded as
-a tag-encoded value in accordance with the type indicated
-by `<type-id>`.
+as counted strings following the same convention as record fields.
```
nwt, line 249:
```suggestion
as counted strings following the same convention as record field names.
```

docs/formats/zson.md
```diff
@@ -545,21 +544,19 @@ needed to resolve the ambiguity, e.g.,
...
 An enum value has the form
 ```
 <name>
 ```
 where the name is either an identifier or a quoted string and it uniquely
-names one of the enum elements.
+corresponds to one of the enum symbols.
 
-Such a value must appear in a context where the enum type is known, i.e.,
+Such a enum value must appear in a context where the enum type is known, i.e.,
```
nwt, line 557:
```suggestion
Such an enum value must appear in a context where the enum type is known, i.e.,
```

docs/formats/zson.md
```diff
@@ -651,16 +648,12 @@ where there are at least two types in the list.
...
 < <identifier>, <identifier>, ... >
 ```
-and the complex form is:
-```
-< <identifier>:<value>, <identifier>=<value>, ... >
-```
-In the same form, the underlying enum value is equal to its positional
-index in the list of identifiers as an uint64 type.
+The positional index of the each enum symbol is significant, e.g., two enum types
+with the same set of symbols but in different order are distinct.
```
nwt, line 656: Do you feel there's a benefit here over defining enum types with the same set of symbols to be identical? That feels tidier.

nwt (approved):

docs/formats/zson.md
mccanne: On the fence... let's merge and we can change it if we want...

mccanne: issue #2823

PR #2826, opened 2021-06-23
encode type values in binary and move type context into zng

mccanne:
This commit reworks the type value specification and implementation
so that ZNG-serialized type values have their own encoding independent
of the ZSON format.  This breaks the dependencing between zng and zson
and will allow the zng package to be moved to the top-level zed package.

This introduces an backward-incompatible change to the ZNG format as
type values are now encoded differently.  However, this change should
be minimally impactful as type values appear more often dynamically as
the result of a query and not so much in origin data.


docs/formats/zng.md
```diff
@@ -539,13 +549,122 @@ representing machine words are serialized in little-endian format.
...
+| `type`     | 20 | variable | type value byte sequence [as defined below](#4-type-values) |
 | `error`    | 21 | variable | UTF-8 byte sequence of string of error message |
 | `null`     | 22 |    0     | No value, always represents an undefined value |
 
 > TBD: For decimal, ZSON assumes decimal128 and there's not a way to specify
 > different widths.  This should be okay since a 128 will hold the others.
 
+## 4. Type Values
+
+As the ZSON data model support first-class types and because the ZNG designs
```
nwt, line 561:
```suggestion
As the ZSON data model support first-class types and because the ZNG design
```

docs/formats/zng.md
```diff
@@ -539,13 +549,122 @@ representing machine words are serialized in little-endian format.
...
+goals require that values need not be translated between type contexts, type values
```
nwt, line 562:
```suggestion
goals require that values need not be translated when used with other type contexts, type values
```

docs/formats/zng.md
```diff
@@ -539,13 +549,122 @@ representing machine words are serialized in little-endian format.
...
+according to the recursive definition in this section.
+
+The type value of a primitive type (include type `type`) is its primitive ID,
+serialized as a single byte.
+
+The type value of a complex type is serialized recursively according to the
+complex type it represents as described below.
+
+
+#### 4.1 Record Type Value
```
nwt, line 574:
```suggestion
complex type it represents as described below.

#### 4.1 Record Type Value
```

docs/formats/zng.md
```diff
@@ -539,13 +549,122 @@ representing machine words are serialized in little-endian format.
...
+An set type value has the form:
+```
+----------------
+|0x21|<typeval>|
+----------------
+```
+where `<typeval>` is a recursive encoding of a type value.
+
+
+#### 4.4 Union Type Value
```
nwt, line 607:
```suggestion
where `<typeval>` is a recursive encoding of a type value.

#### 4.4 Union Type Value
```

docs/formats/zng.md
```diff
@@ -539,13 +549,122 @@ representing machine words are serialized in little-endian format.
...
+
+A record type value has the form:
+```
+-----------------------------------------------------
+|0x19|<ncolumns>|<name1><typeval><name2><typeval>...|
+-----------------------------------------------------
+```
+where `<ncolumns>` is the number of columns in the record encoded as a `uvarint`,
+`<name1>`, `<name2>`, etc. are the field names encoded as in the
+record typedef and each `<typeval>` is a recursive encoding of a type value.
```
nwt, line 584:
```suggestion
`<name1>` etc. are the field names encoded as in the
record typedef, and each `<typeval>` is a recursive encoding of a type value.
```

docs/formats/zng.md
```diff
@@ -539,13 +549,122 @@ representing machine words are serialized in little-endian format.
...
+A map type value has the form:
+```
+----------------------------
+|0x23|<key-type>|<val-type>|
+----------------------------
+```
+where `<key-type>` and `<val-type>` are recursive encodings of type values.
+
+
+#### 4.7 Alias Type Value
```
nwt, line 639:
```suggestion
where `<key-type>` and `<val-type>` are recursive encodings of type values.

#### 4.7 Alias Type Value
```

zng/typetype_test.go
```diff
@@ -0,0 +1,18 @@
...
+	"github.com/brimdata/zed/zson"
+	"github.com/stretchr/testify/require"
+)
+
+func TestTypeValue(t *testing.T) {
+	const s = "{A:{B:int64},C:int32}"
+	zctx := zson.NewContext()
+	typ, err := zson.ParseType(zson.NewContext(), s)
+	require.NoError(t, err)
+	tv := zctx.LookupTypeValue(typ)
```
nwt, line 16:
```suggestion
	typ, err := zson.ParseType(zson.NewContext(), s)
	require.NoError(t, err)
	tv := zson.NewContext().LookupTypeValue(typ)
```

docs/formats/zng.md
mccanne: I massage this a bit more...

zng/typetype_test.go
mccanne: I had used that test to debug I a problem I had in early code.  It was ok to use the same context.  I will add your logic as a second test.

docs/formats/zng.md
```diff
@@ -539,13 +549,119 @@ representing machine words are serialized in little-endian format.
...
+
+#### 4.5 Enum Type Value
+
+An enum type value has the form:
+```
+--------------------------------
+|0x23|<nelem>|<name1><name2>...|
+--------------------------------
+```
+where `<nelem>` and each symbol name are encoded as in an enum typedef.
```
nwt, line 624:
```suggestion
where `<nelem>` and each symbol name is encoded as in an enum typedef.
```

zng/type.go
```diff
@@ -374,3 +373,71 @@ func TypeID(typ Type) int {
 	}
 	return typ.ID()
 }
+
+func FormatType(typ Type) string {
```
nwt, line 377: Delete this since it's unused and looks equivalent to `FormatTypeValue(EncodeTypeValue(typ))`?

nwt (approved):

zng/type.go
mccanne: Yeah, and formatType has the typedefs bug that I fixed on the other side.  Good call.

mccanne: The type value is stored in the context so I think we should have a method on Context to avoid the extra encoding.  But if you don't have the context, no big deal to encode the type value.

mccanne: I think we can also simplify zson/formatter.go now... we'll do this later

Issue #2627, opened 2021-04-23
In-line inputs for automated checking of docs examples


philrz:
Right now the automation that checks the `zq` examples in our markdown docs is predicated on all the inputs being relative to a checkout of [https://github.com/brimdata/zed-sample-data/tree/main/zeek-default.](https://github.com/brimdata/zed-sample-data/tree/main/zeek-default.) This has served us adequately for our Zeek-centric era, but we're effectively outgrowing this because:

1.  The scope of the Zed platform and the Brim app is expanding beyond security/network data
2.  There may be examples we want to use in the docs that better lend themselves to smaller, in-line example inputs (e.g. a few lines of ZSON) rather than assuming a checkout of a larger sample data repo

To achieve this, @nwt and I discussed an approach that borrows ideas from what we've done with ztest. Extending the use of [fenced code blocks](https://spec.commonmark.org/0.29/#fenced-code-blocks) as we do already in [zq\_example.go,](https://github.com/brimdata/zed/blob/main/tests/zq_example.go,) a straw man proposal might look something like this (precise implementation TBD):

---

````
Given the input file `fruit.zson`:

```zq-input fruit.zson
{name:"apple", color:"red", flavor: "tart"} (=fruit)
{name:"banana", color:"yellow", flavor: "sweet"} (fruit)
{name:"bob", likes: "tart", age:61} (=person)
{name:"strawberry", color:"red", flavor: "sweet"} (fruit)
{name:"dates", color:"brown", flavor: "sweet" } (fruit)
{name:"joe", likes: "sweet", age:14} (person)
{name:"figs", color:"brown", flavor: "plain" } (fruit)
{name:"pomegranate", color:"red", flavor: "tart" } (fruit)
{name:"jim", likes: "plain", age:30} (person)
```

We can isolate the red fruits via the Zed query:

```zq-command requires:fruit.zson
zq -z 'color=red' fruit.zson
```

Which produces the output:

```zq-output
{name:"apple",color:"red",flavor:"tart"} (=fruit)
{name:"strawberry",color:"red",flavor:"sweet"} (fruit)
{name:"pomegranate",color:"red",flavor:"tart"} (fruit)
```
````

---

Which renders as:

---

Given the input file `fruit.zson`:

```
{name:"apple", color:"red", flavor: "tart"} (=fruit)
{name:"banana", color:"yellow", flavor: "sweet"} (fruit)
{name:"bob", likes: "tart", age:61} (=person)
[...]

PR #2830, opened 2021-06-25
Use new mdtest-input in "Zed/Zeek Data Type Compatibility" doc

philrz:
Since the "Zed/Zeek Data Type Compatibility" relies on special ad hoc test data, the new `mdtest-input` support (#2627) is well-suited to ensuring we don't regress there.


nwt (approved):
Nice!

PR #2840, opened 2021-06-30
Cast to bool

### LINKED ISSUE #2838: Implement cast to bool
Repro is with Zed commit 389b120.

While working with some sample data sets as part of #2345, I happened to try shaping the schools data from [http://2016.padjo.org/files/data/starterpack/cde-schools/cdeschools.sqlite,](http://2016.padjo.org/files/data/starterpack/cde-schools/cdeschools.sqlite,) of which one field is effectively a boolean indicator of whether or not it's a "magnet" school. After turning the SQLite into JSON and cutting just the unique values:

```
$ cat magnet.ndjson 
{"Magnet":0}
{"Magnet":1}
{"Magnet":null}
```

This would be great to show off the Zed `bool` data type, but both a straight cast and its use in a shaper currently fail:

```
$ zq -version
Version: v0.29.0-405-g389b1200

$ zq -z 'put Magnet:=bool(Magnet)' magnet.ndjson 
rhs of assigment expression: cast to 'bool' not implemented
```

When we implement this, it'd be great if we could recognize the numeric `0` and `1` values as mapping to `false` and `true`, along with the other obvious mappings like string values `False` / `F` and `True` / `T`, etc.

mattnibs:
Closes #2838

### DIFF
--- a/expr/coerce/coerce.go
+++ b/expr/coerce/coerce.go
@@ -243,7 +243,7 @@ func ToBool(zv zng.Value) (bool, bool) {
 		return v, err == nil
 	}
 	v, ok := ToInt(zv)
-	return v == 1, ok
+	return v != 0, ok
 }
 
 func ToTime(zv zng.Value) (nano.Ts, bool) {
--- a/expr/coerce/coerce.go
+++ b/expr/coerce/coerce.go
@@ -237,6 +237,15 @@ func ToInt(zv zng.Value) (int64, bool) {
 	return 0, false
 }
 
+func ToBool(zv zng.Value) (bool, bool) {
+	if zv.IsStringy() {
+		v, err := strconv.ParseBool(string(zv.Bytes))
+		return v, err == nil
+	}
+	v, ok := ToInt(zv)
+	return v == 1, ok
+}
+
 func ToTime(zv zng.Value) (nano.Ts, bool) {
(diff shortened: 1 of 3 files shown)


expr/coerce/coerce.go
```diff
@@ -237,6 +237,15 @@ func ToInt(zv zng.Value) (int64, bool) {
 	return 0, false
 }
 
+func ToBool(zv zng.Value) (bool, bool) {
+	if zv.IsStringy() {
+		v, err := strconv.ParseBool(string(zv.Bytes))
+		return v, err == nil
+	}
+	v, ok := ToInt(zv)
+	return v == 1, ok
```
mccanne, line 246: Shouldn't this be v != 0

expr/ztests/cast-bool.yaml
```diff
@@ -0,0 +1,13 @@
+zed: put b:=bool(b)
+
+input: |
+  {b:0}
+  {b:"F"}
+  {b:1.0}
```
mccanne, line 6: add `{b:2}`

mccanne (approved):
Pending comments...

expr/coerce/coerce.go
mattnibs: Good point!

Issue #2331, opened 2021-03-13
Add debug output beyond "put: a referenced field is missing"


philrz:
Repro is with `zq` commit 2897dd3.

One of the quirks of how Zeek outputs NDJSON is that it will drop "unset" fields/records entirely rather than output them with explicit `null` values, such as we see in the following pair of `smb_files` events where one contains `times` and one does not.

```
$ cat smb_files.ndjson | jq .
{
  "_path": "smb_files",
  "_write_ts": "2018-03-24T17:23:35.680864Z",
  "ts": "2018-03-24T17:23:35.680290Z",
  "uid": "CJKUcZ2cGcTrU6FpN7",
  "id.orig_h": "10.128.0.233",
  "id.orig_p": 44132,
  "id.resp_h": "10.47.21.82",
  "id.resp_p": 445,
  "action": "SMB::FILE_OPEN",
  "path": "\\\\10.47.21.82\\ADMIN$",
  "name": "Temp\\wdszAPrV.tmp",
  "size": 196,
  "times.modified": "2018-03-24T17:19:18.111192Z",
  "times.accessed": "2018-03-24T17:19:11.777714Z",
  "times.created": "2018-03-24T17:19:11.777714Z",
  "times.changed": "2018-03-24T17:19:18.111192Z"
}
{
  "_path": "smb_files",
  "_write_ts": "2018-03-24T17:25:04.361621Z",
  "ts": "2018-03-24T17:25:04.360742Z",
  "uid": "CjTxiyIZxrlkPWtNi",
  "id.orig_h": "10.164.94.120",
  "id.orig_p": 39947,
  "id.resp_h": "10.47.3.142",
  "id.resp_p": 445,
  "action": "SMB::FILE_OPEN",
  "name": "\\browser",
  "size": 0
}
```

I'm trying to shape this data. One of _our_ quirks at the moment is that the only way to convert ISO timestamps into `time` type is to use the `iso()` function. However, in this case an error message is produced for the one record where `times` was absent.

```
$ zq -z -i ndjson 'put times.modified=iso(times.modified)' smb_files.ndjson 
put: a referenced field is missing
[...]

PR #2843, opened 2021-07-01
put: Fix redundant warnings

mattnibs:
Fix issue where put warnings because of a missing field will get output
for every record (should emit one warning on first occurrence).

### DIFF
--- a/proc/put/put.go
+++ b/proc/put/put.go
@@ -69,10 +69,11 @@ func (p *Proc) maybeWarn(err error) {
 	s := err.Error()
 	_, alreadyWarned := p.warned[s]
 	if !alreadyWarned {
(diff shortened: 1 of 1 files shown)


mattnibs:
I'll add a test case for this!

philrz:
I think I'm fine with this change since messages themselves have never communicated the level of debug info I imagined would prove useful to a user (e.g., a mention of which specific field reference didn't work, some way of seeing the record on which the reference failed, etc.) so having a flood of them hasn't them any more useful. FWIW #2331 happens to be an open issue where I've kept a record of my ongoing disappointment with this error message. :smile:

mccanne (approved):

PR #2861, opened 2021-07-09
lake root: load config from disk

### LINKED ISSUE #2798: "zed lake drop" causes "zapi"-level & "zed lake"-level Pool states to be out-of-sync
Repro is with Zed commit e425777. This issue was originally reported by a community user in a [thread on public Slack](https://brimsec.slack.com/archives/C010DR0HHMF/p1623329084040400). I'm starting this out as a bug, but I'm also prepared to have this be a "yeah don't do that" situation. :smile:  

I start by starting a Zed Lake service with an empty root.

```
$ zed -version
Version: v0.29.0-377-ge425777f

$ zed lake serve
{"level":"info","ts":1623355987.710143,"msg":"Open files limit raised","limit":24576}
{"level":"info","ts":1623355987.7117598,"logger":"core","msg":"Started"}
{"level":"info","ts":1623355987.7120798,"logger":"httpd","msg":"Listening","addr":"[::]:9867"}
...
```

First I'll show the happy path where everything works if I stick to just `zapi` commands to make changes to the Lake.

```
$ zapi create -p mypool
pool created: mypool

$ zapi ls
mypool 1tludINwNd2n259J8jWAE90r44w key ts order desc

$ zed lake ls
mypool 1tludINwNd2n259J8jWAE90r44w key ts order desc

$ zapi drop -p mypool
pool deleted: mypool

$ zed lake ls
[no output... expected!]

$ zapi ls
[no output... expected!]
```

However, if I start over and do the `drop` by using the `zed lake` command to perform the operation at the filesystem layer, checking via `zapi` to ask the Lake service for Pool state makes it seem as if the Pool still exists.

```
$ zapi create -p mypool
pool created: mypool

$ zapi ls
mypool 1tluuz7NfIhMJ5a2kQv5azH2gUD key ts order desc

$ zed lake ls
mypool 1tluuz7NfIhMJ5a2kQv5azH2gUD key ts order desc

$ zed lake drop -p mypool
pool deleted: mypool

$ zed lake ls
[no output... expected!]

$ zapi ls
mypool 1tluuz7NfIhMJ5a2kQv5azH2gUD key ts order desc
```

If I stop and restart the `zed lake serve` at this point, now the `zapi ls` no longer returns `mypool`.

Based on these symptoms, I'm guessing the Zed Lake service may reserve the right to hold some state in memory on the assumption that commands like `zed lake drop` aren't changing stuff at the filesystem layer. I'd encouraged beta users to use `zed lake` commands to give them an early jump on playing with Lakes before we'd finished adding all the equivalent commands to `zapi`, and now that we've added just about all those commands to `zapi`, maybe I just need to walk that back. OTOH if we'd like to be able to say that it's safe to manipulate the Lake directly with `zed lake` commands even while it's fronted by a Lake service, then I guess this would be a bug.

mattnibs:
Do not cache the lake config in the root struct. Instead load the lake config
from disk on each action.

Closes #2798

### DIFF
--- a/lake/root.go
+++ b/lake/root.go
@@ -141,8 +141,8 @@ func (r *Root) ScanPools(ctx context.Context, w zio.Writer) error {
 	if err != nil {
 		return err
 	}
-	for k := range config.Pools {
-		rec, err := m.MarshalRecord(&config.Pools[k])
+	for _, p := range config.Pools {
+		rec, err := m.MarshalRecord(&p)
 		if err != nil {
 			return err
 		}
--- a/lake/root.go
+++ b/lake/root.go
@@ -7,7 +7,6 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"sync"
 	"time"
 
 	"github.com/brimdata/zed/expr/extent"
@@ -35,8 +34,6 @@ type Root struct {
 	engine   storage.Engine
 	path     *storage.URI
 	poolPath *storage.URI
-	configMu sync.Mutex
-	Config
 }
 
 var _ proc.DataAdaptor = (*Root)(nil)
@@ -60,7 +57,7 @@ func newRoot(engine storage.Engine, path *storage.URI) *Root {
 
 func Open(ctx context.Context, engine storage.Engine, path *storage.URI) (*Root, error) {
 	r := newRoot(engine, path)
-	if err := r.LoadConfig(ctx); err != nil {
+	if _, err := r.loadConfig(ctx); err != nil {
 		if zqe.IsNotFound(err) {
 			err = fmt.Errorf("%s: no such lake", path)
 		}
@@ -71,7 +68,7 @@ func Open(ctx context.Context, engine storage.Engine, path *storage.URI) (*Root,
 
 func Create(ctx context.Context, engine storage.Engine, path *storage.URI) (*Root, error) {
 	r := newRoot(engine, path)
-	if r.LoadConfig(ctx) == nil {
+	if _, err := r.loadConfig(ctx); err == nil {
 		return nil, fmt.Errorf("%s: lake already exists", path)
 	}
 	//XXX For now, we write an empty config file to indicate that the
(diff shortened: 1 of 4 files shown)


mattnibs:
This will tide us over until the real fix #2547 comes along.

lake/root.go
```diff
@@ -145,13 +135,14 @@ func (r *Root) storeConfig(ctx context.Context) error {
...
 	m.Decorate(zson.StyleSimple)
-	pools := r.Config.Pools
-	for k := range r.Config.Pools {
-		rec, err := m.MarshalRecord(&pools[k])
+	config, err := r.loadConfig(ctx)
+	if err != nil {
+		return err
+	}
+	for k := range config.Pools {
+		rec, err := m.MarshalRecord(&config.Pools[k])
```
nwt, line 145:
```suggestion
	for _, p := range config.Pools {
		rec, err := m.MarshalRecord(&p)
```

lake/root.go
```diff
@@ -162,40 +153,44 @@ func (r *Root) ScanPools(ctx context.Context, w zio.Writer) error {
...
+}
+
+func (r *Root) LookupPoolByName(ctx context.Context, name string) *PoolConfig {
+	pools, err := r.ListPools(ctx)
+	if err != nil {
+		return nil
+	}
+	return lookupPool(pools, func(p PoolConfig) bool {
+		return p.Name == name
+	})
```
nwt, line 193:
```suggestion
	return lookupPoolByName(pools, name)
```

lake/root.go
```diff
@@ -226,31 +221,36 @@ func (r *Root) OpenPool(ctx context.Context, id ksuid.KSUID) (*Pool, error) {
...
 func (r *Root) RenamePool(ctx context.Context, id ksuid.KSUID, newname string) error {
-	r.configMu.Lock()
-	defer r.configMu.Unlock()
-	if exists := r.lookupPoolByName(ctx, newname); exists != nil {
+	config, err := r.loadConfig(ctx)
+	if err != nil {
+		return err
+	}
+	exists := lookupPool(config.Pools, func(p PoolConfig) bool { return p.Name == newname })
+	if exists != nil {
```
nwt, line 229:
```suggestion
	if lookupPoolByName(config.Pools, newname) != nil {
```

nwt (approved):

PR #2865, opened 2021-07-13
driver: Don't recover from flowgraph panic

### LINKED ISSUE #2841: zed lake serve query: panic in flowgraph
As observed in #2822 a panic in the flowgraph of a query endpoint request does not get reported in the zed lake serve logs. Let's fix that.

mattnibs:
This resolves an issue with lake service where the flowgraph in a query
would panic but there would be no log of the panic. The panic is now
caught by the services panic logger middleware.

Closes #2841

### DIFF
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -17,7 +17,6 @@ import (
 	"github.com/brimdata/zed/zbuf"
 	"github.com/brimdata/zed/zio"
 	"github.com/brimdata/zed/zng"
-	"github.com/brimdata/zed/zqe"
 	"github.com/brimdata/zed/zson"
 	"go.uber.org/zap"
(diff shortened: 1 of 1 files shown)


nwt:
I know I said we should do this, so I apologize for reversing course, but I think we need to keep this.

`driver.Reader.Read` creates a new goroutine for `driver.run` at https://github.com/brimdata/zed/pull/2865/files#diff-bd895b31b8e89a49fd8420f2fbdd9bfc78046bd2eefd9c5f5b848c643e9c8d9bR282-R288.

That means the goroutine handling the HTTP request doesn't see flowgraph panics, and `service.panicCatchMiddleware` can't catch them.

Issue #2345, opened 2021-03-15
Rewrite Z language docs with generic data set


philrz:
The Z language docs are currently all referencing a set of Zeek sample data. To emphasize Brim's refocus away from being security-centric to being applicable to any data, the docs should be rewritten with a generic data set.

Issue #2857, opened 2021-07-07
mdtest: Check for correct output even on a failed command


philrz:
Based on my observations as a user, right now mdtest sees any non-zero exit code returned by an `mdtest-command` as a failure that causes the entire `make test-heavy` to be a failure. This makes sense nearly all the time, but poses a problem if the goal is to intentionally show the user example error output of an intentionally failing command. There's an existing example of this in the [`fuse` docs](https://github.com/brimdata/zed/blob/main/docs/language/operators/README.md#output-8) where right now we're working around this limitation by simply not protecting that example with mdtest. That means the command syntax, successful output, and/or error output might all change spontaneously due to a code change, and we'd be relying on our memory or manually running through the docs examples to notice and fix the doc. In some new docs I'm writing, I found yet another place I'd like to allow failure, so it gave me the urge to solve this problem.

As @nwt has made the most recent round of enhancements to mdtest, we had a chat about if it might be feasible to add some way of tagging an `mdtest-command` to mark it as "failure expected" so that would it would tolerate the non-zero exit code, still check the produced stdout+stderr against the `mdtest-output` block, and if it all lines up, see that mdtest as a success. His take:

> If we change  
>```   
>```mdtest-command a/b/c
>``` 
> to  
>```
>```mdtest-command dir=a/b/c
>```
> then we could have something like  
>``` 
>```mdtest-command dir=a/b/c fails
>```
> where `fails` means the commands in the block should fail.

This is very much along the lines of what I was imagining!

philrz:
This is verified effective in that the `fuse` doc example cited above is now enhanced with this new `fails` option and can be seen at [https://github.com/brimdata/zed/blob/main/docs/language/operators/README.md?plain=1#L228-L237.](https://github.com/brimdata/zed/blob/main/docs/language/operators/README.md?plain=1#L228-L237.) The Zed CI has run successfully since this change was merged, which is evidence that the logic is working as intended.

I also double checked it by removing the `fails` and confirming the test no longer passes in that case, as expected.

```
$ zed -version
Version: v0.31.0-6-gb1ef5443-dirty

$ make test-heavy
--- FAIL: TestMarkdownExamples (0.06s)
    --- FAIL: TestMarkdownExamples/docs/language/operators/README.md (1.50s)
        --- FAIL: TestMarkdownExamples/docs/language/operators/README.md/228 (0.11s)
            mdtest.go:128: 
                docs/language/operators/README.md:228: exit status 1
                output:
[...]

PR #2867, opened 2021-07-13
Draft rewrite of Search Syntax language doc using schools data

philrz:
This is a draft to get early feedback on the proposed rewrite of the Zed language docs using "generic data "(#2345) rather than network security logs. The rendered page can be seen at https://github.com/brimdata/zed/tree/edu-data/docs/language/search-syntax. Here I've rigged up a branch of the zed-sample-data repo with shaped ZSON files derived from two of the tables from the sqlite DB at [http://2016.padjo.org/tutorials/sqlite-data-starterpacks/#more-info-california-school-sat-performance-and-poverty-data,](http://2016.padjo.org/tutorials/sqlite-data-starterpacks/#more-info-california-school-sat-performance-and-poverty-data,) along with a third ZSON file that has current IP addresses for some of the school web sites so we can show `ip` types where necessary.

At this phase I'm just looking to get buy-in that this seen as an improvement over what we had before. It takes a fair amount of effort to come up with decent examples with a new data set, so I want to make sure I'm on the right track before I continue using the same data to rewrite all the examples for the operators and aggregate functions. If I get the green light, in addition to the additional docs rewrites in the Zed repo, I'll also plan to overhaul the zed-sample-data repo to have both the Zeek and schools data "down a level" in the directory hierarchy and the top-level README will place each data set in context, with the schools data described prominently as the one intended for use with the current docs and the Zeek ones kept around for the format checking automation, which seems to still have value. I'll also adjust the format checking automation to make sure it still works correctly with all the new paths.

### DIFF
--- a/docs/language/search-syntax/README.md
+++ b/docs/language/search-syntax/README.md
@@ -24,7 +24,7 @@ The simplest possible Zed search is a match of all records. This search is
 expressed in `zq` with the wildcard `*`. The response will be a dump of all
 records. The default `zq` output is binary [ZNG](../../formats/zng.md), a
 compact format that's ideal for working in pipelines. However, in these docs
-we'll sometimes make use of the `-z` option to output the text-based
+we'll often make use of the `-z` or `-Z` options to output the text-based
 [ZSON](../../formats/zson.md) format, which is readable at the command line.
 
 #### Example:
@@ -115,9 +115,9 @@ Sometimes you may need to search for sequences of multiple words or words that
 contain special characters. To achieve this, wrap your search term in quotes.
 
 Let's say we've noticed that a couple of the school names in our sample data
-include the string `Defunct=`. An attempt to enter this as a bare word search
-causes an error because the language parser interpreted this as the start of
-an attempted [field/value match](#fieldvalue-match) for a field named
+include the string `Defunct=`. An attempt to enter this as a [bare word](#bare-word)
+search causes an error because the language parser interpreted this as the
+start of an attempted [field/value match](#fieldvalue-match) for a field named
 `Defunct`.
 
 #### Example:
@@ -148,8 +148,8 @@ zq -z '"Defunct="' schools.zson
 Wrapping in quotes is particularly handy when you're looking for long, specific
 strings that may have several special characters in them. For example, let's
 say we're looking for information on the Union Hill Elementary district.
-Entered without quotes, we up matching way more records than we intended since
-each space character between words is treated as a [boolean `and`](#and) .
+Entered without quotes, we end up matching way more records than we intended
+since each space character between words is treated as a [boolean `and`](#and).
 
 #### Example:
 ```mdtest-command zed-sample-data/edu/zson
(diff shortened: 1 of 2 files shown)


docs/language/search-syntax/README.md
```diff
@@ -125,97 +114,102 @@ describes ways to perform searches against only fields of a specific
...
+include the string `Defunct=`. An attempt to enter this as a [bare word](#bare-word)
+search causes an error because the language parser interpreted this as the
+start of an attempted [field/value match](#fieldvalue-match) for a field named
+`Defunct`.
 
-1. The leading `O=` would be interpreted as the start of an attempted
-   [field/value match](#fieldvalue-match) for a field named `O`.
+#### Example:
+```mdtest-command zed-sample-data/edu/zson
+zq -z 'Defunct=' *.zson || true
```
philrz, line 125: The `|| true` is only here to work around a current mdtest limitation. I either need to do something like this in the example command line to make it look like "success" or simply not protect it with mdtest. There's already another example in the docs that's unprotected for the same reason, so that motivated me to discuss it with @nwt and then I opened #2857. If that can't be addressed before we want to merge these docs, I could go ahead and leave the example unprotected and leave a note in the issue to circle back when we address #2857.

docs/language/search-syntax/README.md
```diff
@@ -125,97 +114,102 @@ describes ways to perform searches against only fields of a specific
 Sometimes you may need to search for sequences of multiple words or words that
 contain special characters. To achieve this, wrap your search term in quotes.
 
-Let's say we want to isolate the records containing the text
-`O=Internet Widgits` that we saw in the response to the previous example
-search. If typed bare as our Zed query, we'd experience two problems:
+Let's say we've noticed that a couple of the school names in our sample data
+include the string `Defunct=`. An attempt to enter this as a [bare word](#bare-word)
+search causes an error because the language parser interpreted this as the
```
nwt, line 119:
```suggestion
search causes an error because the language parser interprets this as the
```

docs/language/search-syntax/README.md
```diff
@@ -547,153 +466,129 @@ If you enter multiple [value match](#value-match) or
...
 ### Parentheses & Order of Evaluation
 
 Unless wrapped in parentheses, a search is evaluated in _left-to-right order_.
+Terms wrapped in parentheses will be evaluated _first_, overriding the default
+left-to-right evaluation.
 
-For example, the following search leverages the implicit boolean `and` to find
-all `smb_mapping` records in which the `share_type` field is set to a value
-other than `DISK`.
+For example, we've noticed there's some test score records that have `null`
```
nwt, line 549:
```suggestion
For example, we've noticed there're some test score records that have `null`
```

nwt (commented):
Looks fine to me so far.

docs/language/search-syntax/README.md
philrz: @nwt: Good catch. I've now gone with `there are` over `there're`. I know `there're` is valid, but it always looks awkward to me.

PR #2869, opened 2021-07-13
python/zqd: use /query instead of /search

nwt:


mattnibs (approved):

mattnibs:
Was thinking maybe we should rename this zqd->service?

nwt:
@mattnibs: We should definitely rename it, though "service" feels too generic. Maybe "zedlakeservice", "zedlakesvc", or "zls"?

PR #2866, opened 2021-07-13
update language docs with some background/context

mccanne:
This isn't finished but headed in the right direction.  I suggest we merge the work-in-progress and continue to improve.

Issue #2877, opened 2021-07-21
"Zed Data Model" doc


philrz:
Right now, when we want to point users at info about the Zed data model (such as from the top-level [Zed language README](https://github.com/brimdata/zed/blob/main/docs/language/README.md)) we link to the [ZSON spec](https://github.com/brimdata/zed/blob/main/docs/formats/zson.md#2-the-zson-data-model). This is decent coverage for now, but @mccanne has envisioned a separate, standalone doc about the data model that would break out this material. That would make the ZSON spec a little shorter, and would be pointed at from the other format docs (ZNG, ZST, ZJSON).

philrz:
Note: When this gets implemented, the top-level [Zed language README](https://github.com/brimdata/zed/blob/main/docs/language/README.md) should be updated to point to the new doc.

PR #2878, opened 2021-07-21
Tweaks to top-level language README

philrz:
Now that #2866 has been merged, here's the first of what I expect to be several more passes through the docs with iterative improvements.

Like you may have seen in other docs, one of the things I'm fussy about is linking to open issues if we're showing syntax that doesn't actually work that way in the tools yet. While it makes the text a little less pretty, I prefer that to the thought of a user attempting to use the syntax we show and wondering why it doesn't work. We can let the desire for prettiness be our motivation to address the issues. :smile:   I'll be putting comments in the issues as a reminder to go back and remove these links as we address them.

### DIFF
--- a/docs/language/README.md
+++ b/docs/language/README.md
@@ -271,10 +271,10 @@ All other operators are explicitly named.
 
 ### Operators not yet fully documented
 
+* `from`
 * `merge`
 * `split`
 * `switch`
-* `from`
 
 ## Conventions
 
--- a/docs/language/README.md
+++ b/docs/language/README.md
@@ -220,7 +220,7 @@ from ... | switch color (
 ) | ...
 ```
 
-> **Note:** The syntax shown here for `switch` is still in the process of being
+> **Note:** The syntax shown here for `switch` is still being
 > implemented ([zed/2849](https://github.com/brimdata/zed/issues/2849)).
 
 ## Operators
--- a/docs/language/README.md
+++ b/docs/language/README.md
@@ -32,7 +32,7 @@ An ambitious goal of the Zed project is to offer a language
 that provides an easy learning curve and a gentle slope from simple keyword search
 to log-search-style processing and ultimately to sophisticated,
 warehouse-scale queries.  The language also embraces a rich set of type operators
-based on the [Zed data model](../formats/zson.md) for data shaping
+based on the [Zed data model](../formats/zson.md#2-the-zson-data-model) for data shaping
 to provide flexible and easy ETL.
 
 The simplest Zed program is perhaps a single word search, e.g.,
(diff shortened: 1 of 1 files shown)


docs/language/README.md
```diff
@@ -32,7 +32,7 @@ An ambitious goal of the Zed project is to offer a language
 that provides an easy learning curve and a gentle slope from simple keyword search
 to log-search-style processing and ultimately to sophisticated,
 warehouse-scale queries.  The language also embraces a rich set of type operators
-based on the [Zed data model](../formats/zson.md) for data shaping
+based on the [Zed data model](../formats/zson.md#2-the-zson-data-model) for data shaping
```
philrz, line 35: FYI, #2877 has been opened as a reminder to eventually break out this topic into its own doc.

docs/language/README.md
```diff
@@ -75,11 +80,11 @@ To facilitate both a programming-like model as well as an ad hoc search
 experience, Zed has a canonical, long form that can be abbreviated
 using syntax that supports an agile, interactive query workflow.
 For example, the canonical form of an aggregation uses the `summarize`
-keyword, as in
+reserved word, as in
 ```
 summarize count() by color
 ```
-but this can be abbreviated by dropping the keyword whereby the compiler then
+but this can be abbreviated by dropping `summarize` whereby the compiler then
```
philrz, line 87: @mccanne and I talked about these. The prior use of "keyword" is an accurate parser-centric statement, but it has an unfortunate clash with "keyword search" which is also mentioned prominently in this doc. So here's my attempt to avoid that clash.

docs/language/README.md
```diff
@@ -258,12 +269,12 @@ All other operators are explicitly named.
...
+### Operators not yet fully documented
 
-* merge
-* split
-* switch
-* from
+* `merge`
+* `split`
+* `switch`
+* `from`
```
nwt, line 277:
```suggestion
* `from`
* `merge`
* `split`
* `switch`
```

docs/language/README.md
```diff
@@ -211,10 +216,13 @@ using `switch`:
 from ... | switch color (
   "red" => op1 | op2 | ... ;
   "blue" => op1 | op2 | ... ;
-  default => op1 | op2 | ...
+  default => op1 | op2 | ... ;
 ) | ...
 ```
 
+> **Note:** The syntax shown here for `switch` is still in the process of being
```
nwt, line 223:
```suggestion
> **Note:** The syntax shown here for `switch` is still being
```

nwt (approved):

docs/language/README.md
```diff
@@ -211,10 +216,13 @@ using `switch`:
 from ... | switch color (
   "red" => op1 | op2 | ... ;
   "blue" => op1 | op2 | ... ;
-  default => op1 | op2 | ...
+  default => op1 | op2 | ... ;
```
philrz, line 219: FYI, in addition to this one fix that brings this closer to the changes made in #2871, one of my goals in a future pass is to create a set of ztests that effectively fill in the `...` portions of these examples with real operations on real input files. While these will likely all be redundant to existing ztests, since I respect the desire to keep these abbreviated examples in the intro text (as opposed to the bulkier in-line mdtest-protected examples we use elsewhere), I'd like to have some tests close to these docs that are prominently labeled along the lines of **"If you find you need to adjust this test in order to make the CI pass, that means you need to update one of the abbreviated docs examples too!"**

mccanne (approved):
Nice!

PR #2881, opened 2021-07-22
expr: allow shaper to cast to union

nwt:
A shaper can now cast a value to a union type provided the value's
underlying type is the underlying type of one of the union's types.

### DIFF
--- a/expr/shaper.go
+++ b/expr/shaper.go
@@ -231,12 +231,14 @@ func castType(zctx *zson.Context, inType, specType zng.Type) (zng.Type, error) {
 			return zctx.LookupTypeArray(out), nil
 		}
 		return zctx.LookupTypeSet(out), nil
-	default:
-		// Non-castable type pair with at least one
-		// (non-record) container: output column is left
-		// unchanged.
-		return inType, nil
 	}
+	if bestUnionSelector(inType, specType) > -1 {
+		return specType, nil
+	}
+	// Non-castable type pair with at least one
+	// (non-record) container: output column is left
+	// unchanged.
+	return inType, nil
 }
 
 func isCollectionType(t zng.Type) bool {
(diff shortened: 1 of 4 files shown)


expr/shaper.go
```diff
@@ -155,6 +158,22 @@ func (s *step) castPrimitive(in zcode.Bytes, b *zcode.Builder) *zng.Value {
 	return nil
 }
 
+func (s *step) castUnion(in zcode.Bytes, b *zcode.Builder) *zng.Value {
+	if in == nil {
+		b.AppendNull()
+		return nil
+	}
+	b.BeginContainer()
+	b.AppendPrimitive(zng.EncodeInt(int64(s.toSelector)))
```
mattnibs, line 167: Feels like there should be some central place (perhaps in package `zng`) where this secret union handshake takes place.

expr/shaper.go
```diff
@@ -603,10 +626,44 @@ func castType(zctx *zson.Context, inType, specType zng.Type) (zng.Type, error) {
...
+	return inType, nil
+}
+
+// bestUnionSelector tries to return the most specific selector for t in
+// maybeUnion.  It returns -1 if maybeUnion is not a union or contains no type
+// compatible with t.  (Types are compatible if they have the same underlying
+// type.)  If maybeUnion contains t, bestUnionSelector returns its selector.
+// Otherwise, if maybeUnion contains t's underlying type, bestUnionSelector
+// returns its selector.  Finally, bestUnionSelector returns the smallest
+// maybeUnion selector whose type is compatible with t.
```
mattnibs, line 645: Maybe this would be a little easier if `index`/`indexOf` was used in place of `selector`?

expr/shaper.go
```diff
@@ -603,10 +626,44 @@ func castType(zctx *zson.Context, inType, specType zng.Type) (zng.Type, error) {
...
+func bestUnionSelector(maybeUnion, t zng.Type) int {
```
mattnibs, line 646: I think it would be clear if you called these variables:
```suggestion
func bestUnionSelector(inType, specType zng.Type) int {
```

mattnibs: (I switched the order per my comment above).

expr/shaper.go
```diff
@@ -603,10 +626,44 @@ func castType(zctx *zson.Context, inType, specType zng.Type) (zng.Type, error) {
...
+	for i, ut := range u.Types {
+		if ut == t {
+			return i
+		}
+		if ut == aliasOfT && underlying == -1 {
+			underlying = i
+		}
+		if zng.AliasOf(ut) == aliasOfT && compatible == -1 {
+			compatible = i
+		}
```
mattnibs, line 663: Not that we have to do this now, but you'd probably here want it to also have the ability to coerce a primitive if there's no exact matching type (for instance string and bstring), no? Feels like we don't do this consistently enough.

expr/shaper.go
nwt: The choice of "selector" here stems from an exchange with @mccanne about our use in the code of both "index" and "selector" to denote this entity. I have a preference for a single term, and he has a preference for "selector" (reserving "index" for data structures), so I used "selector" here and hope to make the rest of the code consistent in another PR.

expr/shaper.go
nwt: I added `zng.BuildUnion` in 982251cbfd6b99acb8e646469308759c3d609780.

expr/shaper.go
nwt: That sounds intuitive, but the decision is complicated by the fact that it probably isn't the behavior we want for the `fuse` operator.

mattnibs (approved):

PR #2882, opened 2021-07-22
allow shaper to cast null type to any type

### LINKED ISSUE #2779: Equality comparison in join should work between comparable types
Repro is with Zed commit [ff3d2e2](https://github.com/brimdata/zed/commit/ff3d2e27bb029836d3fe963a317a19b45f28d14b).

The tl;dr is that at the moment values of different types (specifically `string` and `bstring` in this case) are being treated as non-equal even when their underlying values appear to be the same.

To start with a base case, with a regular test for equality between `string` and `bstring` values, it does what I expect:

```
$ zq -Z stringy.zson 
{
    foo: "hello",
    bar: "hello" (bstring)
} (=0)

$ zq -z 'foo==bar' stringy.zson 
{foo:"hello",bar:"hello" (bstring)} (=0)
```

Now consider this `join`. This next one works because the field called `bar` is a `bstring` in both files.

```
$ zq -Z right.zson 
{
    bar: "hello" (bstring),
    decoration: "LookAtMe!"
} (=0)

$ cat simple.zed 
from (
  file stringy.zson => sort bar;
  file right.zson => sort bar;
) | left join on bar=bar decoration

$ zq -Z -I simple.zed 
{
    foo: "hello",
    bar: "hello" (bstring),
    decoration: "LookAtMe!"
} (=0)
```

However, if I change the type of `bar` in `right.zson` to a regular `string`, now the `join` fails.

```
$ zq -Z right.zson 
{
    bar: "hello",
    decoration: "LookAtMe!"
}

$ zq -Z -I simple.zed 
{
    foo: "hello",
    bar: "hello" (bstring)
} (=0)
```

As a user, the only way I can see to work around this currently is to cast the `bar` value in one of the inputs so the types line up. Once I got hip to the reason the clause in the `join` was failing, this is not a super difficult workaround to put in place, and in a pinch we could surely document it and hope users lock it into their workflow. However, my concern is that users might spend a lot of time frustrated thinking it was some other mistake on their part (e.g. values not quite lining up in some other way, some other problem in the multi-line Zed, etc.)

When I described this to @mccanne he acknowledged that this is something that could be made to work.

### LINKED ISSUE #2853: null value can't be shaped to an array type
Repro is with Zed commit 2b0b0c9.

Consider the following example where I attempt to shape the values of three records using a single type definition in which the named field is defined to be array-of-strings.

```
$ cat example.ndjson 
{"tags": null}
{"tags": []}
{"tags": ["something"]}

$ zq -version
Version: v0.29.0-411-g2b0b0c97

$ zq -Z example.ndjson 
{
    tags: null
}
{
    tags: [] (0=([null]))
}
{
    tags: [
        "something"
    ]
}

$ zq -Z 'type foo={tags:[string]}; put this:=shape(foo) | put shape:=typeof(this)' example.ndjson 
{
    tags: null,
    shape: ({tags:null})
}
{
    tags: [] (0=([string])),
    shape: (foo=({tags:[string]}))
}
{
    tags: [
        "something"
    ],
    shape: (foo=({tags:[string]}))
}
```

I was not expecting the `{tags:null}` type definition to remain. If there was some intentional reason not to allow that shape to be applied, I'd have expected some kind of warning/error to that effect.

As for my best guess as to whether it was supposed to work, I've found [this portion](https://github.com/brimdata/zed/blob/main/docs/formats/zson.md#24-the-null-type) of the ZSON spec:

> The _null_ type is a primitive type representing only a `null` value. A `null` value can have any type.

That second sentence leads me to believe it should have worked, since an array of strings is "a type" (albeit a complex one), so I'd expect a field defined to hold an array of strings could therefore hold "a `null` value".

### LINKED ISSUE #2879: Can't shape a null value to a nested record type
Repro is with Zed commit dcb5b56. I found this issue while working on a rough prototype attempt to shape some Debezium CDC logs.

Consider the following simplified shaper for a record that contains two nested records.

```
$ cat easyshaper.zed 
type foo={one:{f1:string,f2:string},two:{f1:string,f2:string}};
put this := shape(foo)
```

The shaping works fine if there's values present for all the fields.

```
$ zq -version
Version: v0.29.0-426-gdcb5b56d

$ echo '{"one":{"f1":"hi","f2":"there"},"two":{"f1":"hi","f2":"there"}}' | zq -Z -I easyshaper.zed -
{
    one: {
        f1: "hi",
        f2: "there"
    },
    two: {
        f1: "hi",
        f2: "there"
    }
} (=foo)
```

However, if one of the embedded records takes on a `null` value, an error message appears and the shaping fails (note the lack of `(=foo)`).

```
$ echo '{"one":null,"two":{"f1":"hi","f2":"there"}}' | zq -Z -I easyshaper.zed -
createStep incompatible column types null and {f1:string,f2:string}

{
    one: null,
    two: {
        f1: "hi",
        f2: "there"
    }
}
```

mattnibs:
Closes #2853  
Closes #2879

### DIFF
--- a/expr/shaper.go
+++ b/expr/shaper.go
@@ -106,6 +106,8 @@ func isCollectionType(t zng.Type) bool {
 
 func createStep(in, out zng.Type) (step, error) {
 	switch {
+	case in.ID() == zng.IDNull:
+		return step{op: null}, nil
 	case in.ID() == out.ID():
 		if zng.IsContainerType(in) {
 			return step{op: copyContainer}, nil
@@ -579,6 +581,8 @@ func castRecordType(zctx *zson.Context, input, spec *zng.TypeRecord) (*zng.TypeR
 
 func castType(zctx *zson.Context, inType, specType zng.Type) (zng.Type, error) {
 	switch {
+	case inType.ID() == zng.IDNull:
+		return specType, nil
 	case zng.IsRecordType(inType) && zng.IsRecordType(specType):
 		// Matching field is a record: recurse.
 		inRec := zng.AliasOf(inType).(*zng.TypeRecord)
--- a/expr/ztests/shape-null-container.yaml
+++ b/expr/ztests/shape-null-container.yaml
@@ -1,0 +1,18 @@
+# issues #2853 #2879
+
+zed: |
+  put this := shape({
+    array:[string],
+    record:{c1:int64,c2:bstring,c3:time}
+  })
+
+input: |
+  {array:null,record:null}
+
+output-flags: -pretty=2
+
+output: |
+  {
+    array: null (0=([string])),
+    record: null (1=({c1:int64,c2:bstring,c3:time}))
+  } (=2)


nwt (approved):
The PR title doesn't describe this change very well.  Maybe something like "allow shaper to cast null type to any type"?

Issue #2852, opened 2021-07-06
"fuse" options to consolidate heterogeneous type definitions


philrz:
Repro is with Zed commit 2b0b0c9.

A community user spawned a discussion regarding what happens when `fuse` is applied on inputs that would normally create multiple type definitions for values stored in fields of the same name.

Consider the following NDJSON sample data:

```
$ cat example.ndjson 
{"tags": []}
{"tags": ["something"]}
{"num": 1}
{"num": 2.1}

$ zq -version
Version: v0.29.0-411-g2b0b0c97

$ zq -Z 'put typedef := typeof(this)' example.ndjson 
{
    tags: [] (0=([null])),
    typedef: ({tags:[null]})
}
{
    tags: [
        "something"
    ],
    typedef: ({tags:[string]})
}
{
    num: 1,
    typedef: ({num:int64})
}
{
    num: 2.1,
    typedef: ({num:float64})
}
```

Maybe the user's intent was that `tags` would _always_ be treated as an array of strings and `num` would _always_ be treated as floating point. But since no shaper was applied, the behavior shown all makes sense.

When `fuse` is applied, the heterogeneous type definitions have the effect of automatically creating separate named fields for each variation by appending `_2` to the original field names.

```
$ zq -Z 'fuse' example.ndjson 
{
    tags: [] (0=([null])),
    tags_2: null (1=([string])),
    num: null (int64),
    num_2: null (float64)
}
{
    tags: null (0),
    tags_2: [
        "something"
    ],
    num: null (int64),
    num_2: null (float64)
}
{
    tags: null (0),
    tags_2: null (1),
    num: 1,
    num_2: null (float64)
}
{
    tags: null (0),
    tags_2: null (1),
    num: null (int64),
    num_2: 2.1
}
```

This may be fine for the needs of some users, but it threw this particular community user for a loop, so it's worth considering if we might add one or more options to vary the behavior. Two thoughts came out of the discussion:

1.  (My idea :smile: ) As `fuse` is completing its first pass through the data, perhaps it could track if all the observed type definitions for a named field would fit under just one of them. For instance, the empty array for `tags` could be stored using the type definition for the array of strings, and likewise the numeric value `1` could be stored as a float.
2.  (The user's idea :smile: ) Have `fuse` create a `union` type that covers all the observed values. So for example, `tags` could have been an array of union (string, int) in the fused type.

FWIW, after seeing them laid out side-by-side, the community user offered the following take:

[...]

PR #2885, opened 2021-07-23
replace fuse operator field renaming with union

### LINKED ISSUE #2855: fuse creates extra named fields to hold values of the same type
Repro is with Zed commit 2b0b0c9.

While working with a community user on the issues ultimately captured in #2852 and #2853, I observed the following `fuse` behavior I can't easily explain. These queries are being performed against the [Zeek NDJSON zed-sample-data](https://github.com/brimdata/zed-sample-data/tree/main/zeek-ndjson).

Examined in isolation, the `tags` field of the Zeek `http` NDJSON logs only ever have two possible values.

```
$ zq -version
Version: v0.29.0-411-g2b0b0c97

$ zq -z 'count()' http.ndjson.gz 
{count:144034 (uint64)} (=0)

$ zq -z 'count() by tags' http.ndjson.gz 
{tags:["HTTP::URI_SQLI"],count:17 (uint64)} (=0)
{tags:[] (1=([null])),count:144017 (uint64)} (=2)
```

If we use `fuse` on the unshaped NDJSON to try to unite it under a single shape, we see `tags` values get stored across four different named fields `tags`, `tags_2`, `tags_3`, and `tags_4`. This is the only field that `fuse` ends up giving the "append a number to hold the variations" treatment.

```
$ zq -Z 'fuse | count() by typeof(this)' http.ndjson.gz 
{
    typeof: ({_path:string,_write_ts:string,ts:string,uid:string,id.orig_h:string,id.orig_p:int64,id.resp_h:string,id.resp_p:int64,trans_depth:int64,method:string,host:string,uri:string,version:string,user_agent:string,request_body_len:int64,response_body_len:int64,status_code:int64,status_msg:string,tags:[null],resp_fuids:[string],resp_mime_types:[string],referrer:string,orig_fuids:[string],orig_mime_types:[string],origin:string,proxied:[string],resp_filenames:[string],username:string,info_code:int64,info_msg:string,tags_2:[string],tags_3:[string],tags_4:[string]}),
    count: 144034 (uint64)
} (=0)
```

Examining these in isolation:

```
$ zq -z 'fuse | count() by tags' http.ndjson.gz
{tags:[] (0=([null])),count:144017 (uint64)} (=1)
{tags:null,count:17} (1)

$ zq -z 'fuse | count() by tags_2' http.ndjson.gz
{tags_2:null (0=([string])),count:144021 (uint64)} (=1)
{tags_2:["HTTP::URI_SQLI"],count:13} (1)

$ zq -z 'fuse | count() by tags_3' http.ndjson.gz
{tags_3:null (0=([string])),count:144031 (uint64)} (=1)
{tags_3:["HTTP::URI_SQLI"],count:3} (1)

$ zq -z 'fuse | count() by tags_4' http.ndjson.gz
{tags_4:null (0=([string])),count:144033 (uint64)} (=1)
{tags_4:["HTTP::URI_SQLI"],count:1} (1)
```

After having studied the behavior described in #2852, I was expecting no more than two variations on `tags`, one to hold the "array of nulls" values and one to hold the "array of strings" values. So I can't explain the presence of `tags_3` and `tags_4`.

Bottom line: If it turns out this is working as expected, I'd need some help updating the user-facing docs to explain the behavior.

BTW, the symptom seems to have something to do with the presence of all the other fields in the records, since if we isolate just the `tags` values before we `fuse`, now we see the expected behavior similar to what's reported in #2852.

```
$ zq -z 'cut tags | fuse | count() by typeof(this)' http.ndjson.gz 
{typeof:({tags:[null],tags_2:[string]}),count:144034 (uint64)} (=0)
```

nwt:
When records have a field with the same name but different types, the
fuse operator renames all but one uniquely by appending a suffix.
Remove this behavior and combine the different types in a union instead.

Closes #2855.

Part of #2852.

Depends on #2881.

### DIFF
--- a/expr/shaper.go
+++ b/expr/shaper.go
@@ -206,7 +206,7 @@ func castRecordType(zctx *zson.Context, input, spec *zng.TypeRecord) (*zng.TypeR
 
 func castType(zctx *zson.Context, inType, specType zng.Type) (zng.Type, error) {
 	if _, ok := specType.(*zng.TypeMap); ok {
-		return nil, fmt.Errorf("cannot yet use maps in shaping functions")
+		return nil, fmt.Errorf("cannot yet use maps in shaping functions (issue #2894)")
 	}
 	switch {
 	case inType.ID() == zng.IDNull:
--- a/expr/shaper.go
+++ b/expr/shaper.go
@@ -92,7 +92,10 @@ func (s *ConstShaper) Apply(in *zng.Record) (*zng.Record, error) {
 	if err != nil {
 		return nil, err
 	}
-	return zng.NewRecord(v.Type.(*zng.TypeRecord), v.Bytes), nil
+	if !zng.IsRecordType(v.Type) {
+		return nil, fmt.Errorf("shaper returned non-record value %s", zson.String(v))
+	}
+	return zng.NewRecord(v.Type, v.Bytes), nil
 }
 
 func (c *ConstShaper) Eval(in *zng.Record) (zng.Value, error) {
@@ -100,10 +103,7 @@ func (c *ConstShaper) Eval(in *zng.Record) (zng.Value, error) {
 	if err != nil {
 		return zng.Value{}, err
 	}
-	inType, ok := inVal.Type.(*zng.TypeRecord)
-	if !ok {
-		return inVal, nil
-	}
+	inType := zng.AliasOf(inVal.Type).(*zng.TypeRecord)
 	id := in.Type.ID()
 	s, ok := c.shapers[id]
 	if !ok {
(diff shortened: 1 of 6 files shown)


expr/shaper.go
```diff
@@ -208,6 +205,9 @@ func castRecordType(zctx *zson.Context, input, spec *zng.TypeRecord) (*zng.TypeR
 }
 
 func castType(zctx *zson.Context, inType, specType zng.Type) (zng.Type, error) {
+	if _, ok := specType.(*zng.TypeMap); ok {
+		return nil, fmt.Errorf("cannot yet use maps in shaping functions")
```
mattnibs, line 209: Is there an issue tracking this? Maybe include the issue in the error?

mattnibs (approved):

expr/shaper.go
nwt: Good idea. I created #2894 and added it to this error string in 0926fce46aed162fe1b807b3c618392b16badc5c.

PR #2888, opened 2021-07-23
add "switch expr (literal => ...;)" syntax

### LINKED ISSUE #2849: consistent topology syntax
Currently, `switch`, `split`, and `from` use different conventions for denoting the termination of each parallel path.  We should unify this syntax.

The proposal is to use semi-colon termination consistently.  This would allow us to get rid of the `case` keyword in switch.  We should also have a form of switch where you can switch on an expression and have a list of cases with default instead of just a list boolean predicates, e.g., 
```
switch  _path ( 
  "conn" => ... ;
  "dns" => ... ;
  default => ... ;
) |  ...
```

nwt:
The switch operator now supports two forms:
1. "switch (f => ...;)" where f is a filter
2. "switch e (l => ...;)" where e is an expression and l is a literal

Closes #2849.

### DIFF
--- a/compiler/semantic/proc.go
+++ b/compiler/semantic/proc.go
@@ -249,21 +249,40 @@ func semProc(ctx context.Context, scope *Scope, p ast.Proc, adaptor proc.DataAda
 	case *ast.Sequential:
 		return semSequential(ctx, scope, p, adaptor)
 	case *ast.Switch:
-		var cases []dag.Case
-		for k := range p.Cases {
+		var expr dag.Expr
+		if p.Expr != nil {
 			var err error
-			tr, err := semProc(ctx, scope, p.Cases[k].Proc, adaptor)
+			expr, err = semExpr(scope, p.Expr)
 			if err != nil {
 				return nil, err
 			}
-			e, err := semExpr(scope, p.Cases[k].Expr)
+		}
+		var cases []dag.Case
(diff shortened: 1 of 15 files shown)


compiler/semantic/proc.go
```diff
@@ -249,21 +249,40 @@ func semProc(ctx context.Context, scope *Scope, p ast.Proc, adaptor proc.DataAda
...
+		var cases []dag.Case
+		for _, c := range p.Cases {
+			var e dag.Expr
+			if c.Expr != nil {
+				var err error
+				e, err = semExpr(scope, c.Expr)
+				if err != nil {
+					return nil, err
+				}
+			} else if p.Expr == nil {
```
mattnibs, line 269: Should this be just `} else {` or did you mean to check if the switch expression is nil?

nwt: I do mean to check if the switch expression, `p.Expr`, is nil here. A nil case expression, `c.Expr`, indicates the default case, and its handling depends on whether or not there's a switch expression. I'll update the comment on the following line to make that explicit.

mattnibs (approved):

compiler/semantic/proc.go
nwt: Done in cdc79fd771a5e497516247cc5fa3fc133cb97293.

PR #2889, opened 2021-07-24
expr: replace step.castTypes with fromType and toType

nwt:
The struct type literal is cumbersome, and a type declation doesn't feel
warranted.


mattnibs (approved):
I approve of this move.

Issue #2900, opened 2021-08-03
Attempted sort of aggregation result does not produce deterministic output


philrz:
Repro is with Zed commit 77bdc81.

I'm using this test data:

```
$ cat repro.zson 
{srcip:10.0.0.1,dstip:5.6.7.8,src_bytes:20,dst_bytes:40,status:"ok"}
{srcip:10.0.0.1,dstip:5.6.7.8,src_bytes:20,dst_bytes:40,duration:10}
{srcip:192.168.1.1,dstip:5.6.7.8,src_bytes:10,dst_bytes:30,duration:5,status:"ok"}
{srcip:10.0.0.1,dstip:9.10.11.12,src_bytes:25,dst_bytes:45,status:"ok"}
{srcip:10.0.0.1,dstip:9.10.11.12,src_bytes:20,dst_bytes:40,status:"failed"}
{srcip:10.0.0.1,dstip:13.14.15.16,src_bytes:29,dst_bytes:49}
{srcip:10.0.0.1,dstip:13.14.15.16,src_bytes:23,dst_bytes:43}
{srcip:10.0.0.1,src_bytes:23,dst_bytes:43}
{srcip:10.0.0.1,dstip:17.18.19.20,status:"ok"}
{srcip:10.0.0.1,dstip:17.18.19.20,status:"ok"}
```

With this Zed script:

```
$ cat repro.zed 
not srcip in 192.168.0.0/16
| summarize
    bytes := sum(src_bytes + dst_bytes),
    maxdur := max(duration),
    valid := and(status == "ok")
      by srcip, dstip
| sort
```

I included the `| sort` because I'm trying to create a ztest and I know the output of an aggregation is non-deterministic, but it's been my understanding that tacking on a `sort` like this (even if no explicit sort keys are mentioned) should make for deterministic output. However, after enough repeated runs I can see the order varies.

```
$ zq -z -I repro.zed repro.zson
{srcip:10.0.0.1,dstip:5.6.7.8,bytes:120,maxdur:10,valid:true}
{srcip:10.0.0.1,dstip:9.10.11.12,bytes:130,maxdur:null,valid:false}
{srcip:10.0.0.1,dstip:13.14.15.16,bytes:144,maxdur:null,valid:null (bool)}
{srcip:10.0.0.1,dstip:17.18.19.20,bytes:null,maxdur:null,valid:true}

$ zq -z -I repro.zed repro.zson
[...]

PR #2901, opened 2021-08-03
Top-level language doc README ztests

philrz:
While most of our docs examples are protected via in-line mdtests, the top-level language README is currently written in a more abbreviated way. I'm on-board with this because it emphasizes the most important elements, gets readers started quicker, and provides a jumping-off point for clicking onward to the more comprehensive docs. At the same time, the existence of mdtests and our need to frequently update them indicates that even these intro-level concepts may be subject to change. If we lack automation to protect them, we'd be exposed to the possibility of showing invalid syntax to first-time users, which I'd really like to avoid.

As a way to provide coverage, here I've added a `ztests/` directory with one test that corresponds to each of the unprotected examples in the top-level README. I've put text in the top of each ztest to call attention to which example it's protecting and to explain any small differences, so hopefully the maintenance burden will be minimal.

I'm fully aware that some of these tests are effectively redundant to tests we have elsewhere. But since they only take a few extra seconds to run, I'm in favor of eating this cost rather than trying to get fancy with easy-to-miss comments in other ztests that don't line up precisely 1-for-1 with the examples shown.

I'm also aware that an in-line approach similar to mdtest would be easier to maintain, but mdtest is not up to the task in its current form. I spoke with @nwt and he thinks there's a way using in-line HTML that we could employ to achieve this down the road. But in the interest of us getting some coverage in the short term without adding more distractions to our pre-release to-do list, I've opted to take this tactical step.

As evidence that this was not wholly busy work, the effort did already lead to some significant findings:

1.  The realization that `merge` isn't currently available as an explicit operator.
2.  The behavior described in #2900.

I'll put a couple other PR comments in-line where I'd particularly appreciate attention from reviewers.


docs/language/README.md
```diff
@@ -90,11 +90,12 @@ as in the shorter form
...
-Similarly, the canonical form of a search expression is a `filter` operator
-(and the "AND" operator is explicit in canonical form),
-so the example from above would be written canonically as
+Similarly, the canonical form of a search expression includes a `filter`
+operator and the searched word enclosed in a `match()` function (also, the "AND"
+operator is explicit in canonical form). Therefore the example from above would
+be written canonically as
 ```
-filter widget and price > 1000
+filter match("widget") and price > 1000
```
philrz, line 98: `zc -C` shows the `match()`, so I've opted to include it here and on the other relevant places on the page. I'm not sure if its exclusion was intentional in attempt to keep things looking simpler. However, canonical form is mentioned enough times on this page that I felt like we might as well go all the way with it.

docs/language/README.md
```diff
@@ -201,7 +202,7 @@ A data path can be split with the `split` operator as in
 from PoolOne | split (
   => op1 | op2 | ... ;
   => op1 | op2 | ... ;
-) | merge ts | ...
+) | sort ts | ...
```
philrz, line 205: This change came out of the internal discussion we had after I stumbled onto `merge` not being currently recognized as an explicit operator. After the next GA release goes out, @mccanne, @nwt, and I are due to have a wider chat about implicit/explicit `sort` behaviors, and the topic of implicit/explicit `merge` is related to that. In a preliminary chat, @nwt made the point that one approach we may settle on is that the user may express their desired merge behavior similar to what's shown here via `sort`, at which point our optimizer might be free to _implement_ the sort by potentially sorting the multiple flow graph branches and then merging via the sorted field, but that could be made invisible to the user. Therefore, I've opted to use the explicit `sort` for now so we're not showing an operator that's unavailable to the reader.

docs/language/ztests/language-operators-1.yaml
```diff
@@ -0,0 +1,27 @@
...
+# docs/language/README.md, as "put" is not yet optional
+# (https://github.com/brimdata/zed/issues/2850).
+
+script: |
+  zc -C -I input.zed
+
+inputs:
+  - name: input.zed
+    data: |
+      widget | count() by color | put COLOR := to_upper(color)
```
nwt, line 19:
```suggestion
  zc -C 'widget | count() by color | put COLOR := to_upper(color)'
```

docs/language/ztests/language-directed-acyclic-flow-graphs-3.yaml
```diff
@@ -0,0 +1,47 @@
...
+    data: |
+      {name:"apple","color":"red",price:1.5}
+      {name:"banana","color":"yellow",price:2.1}
+      {name:"blueberry","color":"blue",price:0.1}
+  - name: switch.zed
+    data: |
+      from fruit | switch color (
+        "red" => put price:=price+0.5 | put bigcolor:=to_upper(color) | put tag:="expensive" ;
+        "blue" => put price:=price+0.1 | put colorlen:=len(color) | put tag:="cheaper" ;
+        default => put price:=price-0.1 | put promo:="discounted" | put tag:="promoted" ;
```
nwt, line 39:
```suggestion
        "red" => put price:=price+0.5 | put bigcolor:=to_upper(color) | put tag:="expensive";
        "blue" => put price:=price+0.1 | put colorlen:=len(color) | put tag:="cheaper";
        default => put price:=price-0.1 | put promo:="discounted" | put tag:="promoted";
```

docs/language/ztests/language-directed-acyclic-flow-graphs-2.yaml
```diff
@@ -0,0 +1,56 @@
...
+    data: |
+      {key:"apple",price:1.5}
+      {key:"banana",price:2.1}
+      {key:"avocado",price:3.0}
+      {key:"strawberry",price:0.5}
+  - name: join.zed
+    data: |
+      from (
+        PoolOne => put color:=to_upper(color) | put namelen:=len(key) | sort key ;
+        PoolTwo => put price:=price+0.5 | put tag:="mytag" | sort key ;
```
nwt, line 47:
```suggestion
        PoolOne => put color:=to_upper(color) | put namelen:=len(key) | sort key;
        PoolTwo => put price:=price+0.5 | put tag:="mytag" | sort key;
```

docs/language/ztests/language-directed-acyclic-flow-graphs-1.yaml
```diff
@@ -0,0 +1,49 @@
...
+    data: |
+      {ts:2018-03-24T17:15:21Z,name:"apple",color:"red"}
+      {ts:2018-03-24T17:16:55Z,name:"banana",color:"yellow"}
+      {ts:2018-03-24T17:21:23Z,name:"avocado",color:"green"}
+      {ts:2018-03-24T17:23:29Z,name:"strawberry",color:"red"}
+  - name: split.zed
+    data: |
+      from PoolOne | split (
+        => every 5m allfruit:=collect(name),allcolor:=collect(color) | put numfruit:=len(allfruit) | put numcolor:=len(allcolor) ;
+        => put color:=to_upper(color) | put namelen:=len(name) | put tag:="mytag" ;
```
nwt, line 38: I think we should adopt a Zed style in which a semicolon is not preceded by a space.
```suggestion
        => every 5m allfruit:=collect(name),allcolor:=collect(color) | put numfruit:=len(allfruit) | put numcolor:=len(allcolor);
        => put color:=to_upper(color) | put namelen:=len(name) | put tag:="mytag";
```

nwt (approved):

docs/language/ztests/language-directed-acyclic-flow-graphs-1.yaml
philrz: @nwt: Yes, "semicolon is not preceded by a space" is totally reasonable. I was being dense and blindly filling in the `...` and `op1` / `op2` stuff from the examples in the README and not removing the extra space chars that made sense _there_ to call attention to that stuff being placeholders and not real language elements. Will fix!

docs/language/README.md
philrz: Update: After a conversation with @mccanne, I've pushed a commit that reverses course here. He felt that the `merge` was communicating something important and that this would be lost if replaced with `sort` as I'd been proposing. Based on this he was in favor of just finishing the add of explicit `merge` to the language in the short term. Then @nwt confirmed in a separate chat that it'd be a bit more complicated than just a one-word addition to the grammar (apparently it would need AST changes, test coverage, etc.), plus a proposal in #2905 would have us taking that back out anyway, so the topic should probably live in our backlog behind other current MVP0 priorities. Therefore I've opened #2906 to fill that role and have put back `merge` in this example and linked to that issue from a "Note:" under the example. If we manage to finish implementing explicit `merge` before we release, great. If not, at least now we've got user guidance to go with the currently-non-working example. And for now the `| sort ts` remains in the ztest so at least the other parts of the example's syntax get some test coverage.

nwt (approved):

PR #2910, opened 2021-08-06
reformat ztests with gopkg.in/yaml.v3

nwt:
Lay groundwork for automatic transformations on ztests by reformatting
them with gopkg.in/yaml.v3.

Depends on #2909.


mattnibs (approved):
Maybe add a lint step that runs this?

nwt:
> Maybe add a lint step that runs this?

After go-yaml/yaml#738 is merged.  Without that, a number of multiline literals are converted to double-quoted style for no good reason, making them hard to read.

PR #2911, opened 2021-08-09
remove space before decorator in ZSON -pretty=0/-z output

### LINKED ISSUE #2821: remove decorator space in -z output
The only space that appears in -z output is between a value and its decorator.  We should remove this for consistency (while keeping the space in -Z output).  e.g., `{a:1(int8)}` instead of `{a:1 (int8)}`.

nwt:
The operative change is in zson//formatter.go.  Everything else here
just updates tests.

Depends on #2910.

Closes #2821.


mattnibs (approved):
Jeez!

PR #2915, opened 2021-08-11
fix tzngio type value formatter

mccanne:
This commit fixes a bug in the table output that prints the new,
binary-encoded type value as a string.  For now, the fix here is
to update the tzngio printer though this should eventually go away
and the table writer should probably use package zson.


mattnibs (approved):
ztest?

PR #2924, opened 2021-08-17
service: send 400 for POST /pool/{}/add with no records

### LINKED ISSUE #2883: Attempted load of garbage data returns HTTP 500 error with no auto-detect detail
As described in https://github.com/brimdata/brim/issues/1749#issuecomment-885026987, when a user attempts to load garbage data into Brim, the Zed backend is now returning an HTTP Internal Server Error ("internal error: serialize transaction: empty transaction") with none of the Zed auto-detect detail that it did in the past (a port 9867 pcap that shows this is attached to https://github.com/brimdata/brim/issues/1749). The app is swallowing erroneously swallowing this error at the moment and @mason-fish is planning to address that, but it seems like we have work to do on the backend to return the auto-detect detail once again and (I presume) describe it as something more specific than an HTTP 500.

nwt:
A POST /pool/{pool}/add request whose body is empty or cannot be parsed
elicits a 500 response.  Change that to 400.

Closes #2883.

### DIFF
--- a/lake/commit/transaction.go
+++ b/lake/commit/transaction.go
@@ -13,6 +13,10 @@ import (
 	"github.com/segmentio/ksuid"
 )
 
+// ErrEmptyTransaction is the error returned by Transaction.Serialize when
+// Transaction.Actions is empty.
+var ErrEmptyTransaction = errors.New("empty transaction")
+
 type Transaction struct {
 	ID      ksuid.KSUID         `zng:"id"`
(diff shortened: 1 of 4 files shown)


mattnibs (approved):
Great! At some point we should probably move zqe into the server or api package...

nwt:
> At some point we should probably move zqe into the server or api package...

I like it.

PR #2942, opened 2021-08-20
csvio.Reader: Remove extraneous quotes

### LINKED ISSUE #2862: CSV reader error: extraneous or missing " in quoted-field
Repro is with Zed commit [af73996](https://github.com/brimdata/zed/commit/af73996a690b5b7f3e3a78e7e0689bd007e7ec53).

This symptom was originally captured in a comment of #2332, but since that issue was narrowed in scope such that it only addressed a problem with a different CSV dialect, it's been moved to this separate issue.

A community user was engaged in beta testing and was unable to read some CSV data with Zed. The attached [evtx\_repro.zip](https://github.com/brimdata/zed/files/6671417/evtx_repro.zip) contains everything necessary for the repro. When unpacked, you'll have:

*   `evtx_to_csv.ps1` - A PowerShell script that turns Windows EVTX logs ([**EVTX File Parsing Request #2793**](https://github.com/brimdata/zed/issues/2793)) into CSV
*   `Sysmon_SlothFulMedia.evtx` - A sample EVTX log
*   `Sysmon_SlothFulMedia.csv` - The CSV output from running that EVTX log through the PowerShell script

The repro steps performed in PowerShell on my Windows 10 laptop:

```
PS C:\> zq -version
Version: af73996

PS C:\> .\evtx_to_csv.ps1 .\Sysmon_SlothFulMedia.evtx
Reading in the .evtx.
Finding unique fields.
Processing lines.
Processed 86 events.
Writing output file

PS C:\> zq -z -i csv Sysmon_SlothFulMedia.csv
Sysmon_SlothFulMedia.csv: parse error on line 8, column 380: extraneous or missing " in quoted-field
```

For what it's worth, I was able to read this CSV file without complaint into both Excel and Google Sheets. So while it does seem an argument could be made about how this particular CSV output is technically impure ([RFC4180](https://datatracker.ietf.org/doc/html/rfc4180) covers this topic in some detail), given our intent to have Zed be friendly to as much input data as possible, it seems it would be good for us to be as tolerant as the others. Even if we had to require the user to twiddle some flags to get it right, it seems that would be an improvement over requiring them to preprocess with other tools first, which is what it seems they'd be stuck doing currently.

In terms of the specific place it's choking with this example, that line 8 happens to be the first location of any `"` in the file. In the original input CSV, the value where the `"` appears happens to be:

```
"C:\Windows\SysWOW64\DllHost.exe" /Processid:{776DBC8D-7347-478C-8D71-791E12EF49D8}
```

When Google Sheets reads in this CSV, it appears in Sheets with the quotes stripped away:

```
C:\Windows\SysWOW64\DllHost.exe /Processid:{776DBC8D-7347-478C-8D71-791E12EF49D8}
```

mattnibs:
Add a reader on top of the go csv reader that looks for fields where
quotes do not cover the entirety of a field. If such a case is found
the quotes are stripped from the field.

For instance, the line:
field1,"field2" extra
Would get converted into:
field1,field2 extra

Closes #2862

### DIFF
--- a/zio/csvio/preprocess.go
+++ b/zio/csvio/preprocess.go
@@ -1,15 +1,9 @@
 package csvio
 
 import (
+	"bufio"
 	"bytes"
 	"io"
-
-	"github.com/brimdata/zed/pkg/skim"
-)
-
-const (
-	ReadSize    = 64 * 1024
-	MaxLineSize = 50 * 1024 * 1024
 )
 
 // preprocess is a reader, meant to sit in front of the go csv reader, that
@@ -22,13 +16,13 @@ const (
 // field1,field2 extra
 type preprocess struct {
 	leftover []byte
-	scanner  *skim.Scanner
+	scanner  *bufio.Reader
+	scratch  []byte
 }
 
 func newPreprocess(r io.Reader) *preprocess {
-	buffer := make([]byte, ReadSize)
 	return &preprocess{
-		scanner: skim.NewScanner(r, buffer, MaxLineSize),
+		scanner: bufio.NewReader(r),
 	}
 }
 
@@ -40,12 +34,15 @@ func (p *preprocess) Read(b []byte) (int, error) {
 		}
 	}
 	for {
-		line, err := p.scanner.Scan()
-		if len(line) > 0 {
-			line = checkLine(line)
-			cc := p.copy(b[n:], line)
+		field, err := p.parseField()
+		if len(field) > 0 {
+			cc := p.copy(b[n:], field)
 			n += cc
-			if cc < len(line) {
+			if cc < len(field) {
+				// If copied is less than field size it means there was not
+				// enough space in b to copy the entirety of the field. The
+				// copy function has copied the remaining data into leftover,
+				// just return what we have.
 				return n, err
 			}
 		}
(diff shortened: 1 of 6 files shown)


mccanne (approved):
I don't follow all the logic but it survived my torture tests!   Add these cases to the test...
```
""",""",""" has a couple "" embedded quotes and a , comma",""" """
x,"hello,
"" world , ",y`
```

mccanne:
Oops, I realized this didn't work because the second test line above should result in...
```
x,"hello,
"" world , foo",y`
```

mattnibs:
> Oops, I realized this didn't work because the second test line above should result in...

Yeah this is indeed a trickier case, this pr is greatly simplified by not having to deal with newlines in quotes. I'll take a look at handling this case. At this point we'll not be that far from having our own csv parser.

Also I think the case:
```
x,"hello,
"" world , " foo,y`
```
Should really result in:
```
x,"hello,
 world ,  foo",y`
```

zio/csvio/ztests/extraneous-quotes.yaml
```diff
@@ -0,0 +1,14 @@
+script: zq -i csv -z '*' -
```
nwt, line 1: Make this a Zed test instead of a shell test.

zio/csvio/preprocess.go
```diff
@@ -0,0 +1,127 @@
...
+
+// preProcess is a reader, meant to sit in front of the go csv reader, that
+// looks for fields where quotes do not cover the entirety of a field. If such
+// a case is found the quotes are stripped from the field.
+//
+// For instance, the line:
+// field1,"field2" extra
+// Would get converted into:
+// field1,field2 extra
+type preProcess struct {
```
nwt, line 23: Call this preprocess since preprocess is usually written as one word.

zio/csvio/ztests/extraneous-quotes.yaml
mattnibs: You can't specify input flags for zed tests.

nwt: Not a problem: ztest uses `anyio.NewReader`, which will figure out that the input is CSV if this is working.

mattnibs: This is the error I get:
```
/Users/nibs/Code/zed/zio/csvio/ztests/extraneous-quotes.yaml: format detection error
                	tzng: line 1: bad format
                	zjson: line 1: invalid character 'i' in literal false (expecting 'a')
                	zson: identifier "field1" must be enum and requires decorator
                	zng: zng type ID out of range
                	csv: auto-detection not supported
                	json: auto-detection not supported
                	parquet: auto-detection not supported
                	zst: auto-detection not supported
```

nwt: I guess I'm a dummy. Sorry about that!

ztest/shell.go
```diff
@@ -33,5 +33,6 @@ func RunShell(dir Dir, bindir, script string, stdin io.Reader, useenvs []string)
 	cmd.Stdout = &stdout
 	cmd.Stderr = &stderr
 	err := cmd.Run()
+	fmt.Println("stderr", stderr.String())
```
nwt, line 36:
```suggestion
```

nwt (approved):

mccanne (approved):

PR #2948, opened 2021-08-23
compiler/parser: use ? instead of empty parsing expressions

nwt:


mattnibs (approved):
Interesting. Is there some area in `/compiler` (perhaps semantic?) that was handling NULL expressions? Perhaps this logic can go to?

nwt:
@mattnibs: Not that I know of.

PR #2950, opened 2021-08-24
ztest: Add input flags to zed tests

mattnibs:
This allows for zed tests to specify the input format. Useful when
testing zio Readers such as csv that do not working with auto detect.

### DIFF
--- a/ztest/ztest.go
+++ b/ztest/ztest.go
@@ -129,6 +129,7 @@ import (
 	"testing"
 	"unicode/utf8"
 
+	"github.com/brimdata/zed/cli/inputflags"
 	"github.com/brimdata/zed/cli/outputflags"
 	"github.com/brimdata/zed/compiler"
 	"github.com/brimdata/zed/driver"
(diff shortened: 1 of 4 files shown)


ztest/ztest.go
```diff
@@ -548,18 +552,20 @@ func runzq(path, zed, input string, outputFlags []string) (string, string, error
 	if err != nil {
 		return "", "", err
 	}
+	var outflags outputflags.Flags
+	var inflags inputflags.Flags
+	var flags flag.FlagSet
+	outflags.SetFlags(&flags)
+	inflags.SetFlags(&flags)
+	if err := flags.Parse(append(outputFlags, inputFlags...)); err != nil {
```
nwt, line 560: Using a single `flag.FlagSet` like this allows output flags in the `input-flags` field and input flags in the `output-flags` field, which feels wrong, but maybe it isn't.

mattnibs: Yeah that's a good call, will fix.

nwt (approved):

PR #2967, opened 2021-08-27
Optional Put

### LINKED ISSUE #2850: optional put keyword
The `put` keyword should be optional, e.g., so
```
... | put x := 1 | ...
```
can be abbreviated 
```
... | x := 1 | ...
```

mattnibs:
Allow expressions like `x := 1` to get interpreted as a put proc.

The big change here is a change to how the parser handles a list of expressions
without a proc (e.g. `to_lower(x), to_upper(y)`). They now get parsed as
an AmbiguousExprList proc. It is then on the semantic compiler to look
at the list of expressions and determine if the list of expressions is a
Filter, Summarize or Put proc.

Closes #2850


compiler/ast/ast.go
```diff
@@ -292,6 +292,14 @@ type (
 		Args []Assignment `json:"args"`
 	}
 
+	// An AmbiguousExprList proc is a list of assignments whose parent proc
+	// is unknown: It could be a Filter, Summarize, or Put proc. This
+	// will be determined in the semantic phase.
+	AmbiguousExprList struct {
```
mccanne, line 298: Ambiguity is a property of context not syntactic structure.  Since this is the AST reflecting syntax, it seems better to just call something more reflective than this and that will help us understand what this is.  Since this lives at the operator level, maybe OpExpr?   Also, though the *List naming pattern is used elsewhere, I think *s pattern works better here... OpExprs instead of OpExprList instead of AmbiguousExprList.

compiler/ast/ast.go
```diff
@@ -292,6 +292,14 @@ type (
...
+		Kind string `json:"kind" unpack:""`
+		Args []Expr `json:"args"`
```
mccanne, line 300: Exprs instead of Args to be consistent

compiler/parser/parser.peg
```diff
@@ -466,6 +466,26 @@ SampleProc
...
+  = first:AmbiguousExpr rest:(__ "," __ expr:AmbiguousExpr { RETURN(expr) })* {
+    RETURN(PREPEND(first, rest))
+  }
+
+AmbiguousExpr
+ = Assignment
+// XXX Perhaps SearchFactor should be removed here and we leave it up to the 
+// semantic compiler to convert a raw primitive to a search?
+ / SearchFactor
+ / Expr
```
mccanne, line 487: I thought this would just be Assignment and Function... why do you need Expr and SearchFactor?

compiler/semantic/expr.go
```diff
@@ -494,48 +502,54 @@ func semField(scope *Scope, e ast.Expr) (dag.Expr, error) {
...
-// This way, Zed of the form `... | exists(...) | ...` can be distinguished
-// from `count()` by the name lookup here at compile time.
-func convertFunctionProc(scope *Scope, call *ast.Call) (dag.Op, error) {
-	if _, err := agg.NewPattern(call.Name); err != nil {
-		// Assume it's a valid function and convert.  If not,
-		// the compiler will report an unknown function error.
-		c, err := semCall(scope, call)
-		if err != nil {
-			return nil, err
+func maybeConvertFilterExpr(scope *Scope, e ast.Expr) (*dag.Filter, error) {
```
mccanne, line 505: determineOpExpr()?

compiler/semantic/proc.go
```diff
@@ -575,3 +575,55 @@ func isConst(p ast.Proc) bool {
...
+			if _, ok := assignment.RHS.(*dag.Agg); ok {
+				aggs = append(aggs, assignment)
+			} else {
+				puts = append(puts, assignment)
+			}
+		default:
+			return nil, errors.New("invalid expression in assignment list")
+		}
+	}
+	if len(puts) > 0 && len(aggs) > 0 {
```
mccanne, line 616: 👍

mccanne (changes requested):
At some point in sync, I thought we agreed that the implied puts would require ":=" but now I am forgetting how we relaxed that, e.g,. this is a nice shortcut
```
* | to_lower(s) | ...
```
but it seems extreme compared to requiring ":=" when you don't have put, e..g, 
```
* | put to_lower(s) | ...
```
or 
```
* | to_lower:=to_lower(s) | ...
```
Maybe we can discuss further?

compiler/parser/parser.peg
mattnibs: I think I got ahead of myself and started thinking it would be nice if any expression without a proc should just get parsed as an AmbiguousExprList (or OpExrs) and relying on the semantic compiler to figure out the context (or throw an error if expr list doesn't make any sense), instead of having the parser figure out the context and return a parser error if the context is off. I still think this is the right approach, but not necessary in the scope of this change. I will change it as you suggest.

compiler/ast/ast.go
mattnibs: Yep, much better!

compiler/semantic/expr.go
mattnibs: Ending up dropping this function.

mccanne (approved):
Cool!

compiler/parser/parser.peg
```diff
@@ -466,6 +466,21 @@ SampleProc
                                 "rhs": MAP("kind": "ID", "name": "sample")))))))
     }
 
+// OpExprs is a top level operator that contains a list of function calls or
+// assignment expressions. It is in an ambiguous state and could either be a:
+//   1) A summarize proc
+//   2) a put proc or
+//   3) a filter.
```
nwt, line 473:
```suggestion
// assignment expressions. It is in an ambiguous state and could either be
//   1) a summarize proc,
//   2) a put proc, or
//   3) a filter.
```

compiler/parser/parser.peg
```diff
@@ -466,6 +466,21 @@ SampleProc
...
+//  The semantic pass is responsible for converting to either of the three.
```
nwt, line 474:
```suggestion
//  The semantic pass is responsible for converting it to one of the three.
```

compiler/semantic/expr.go
```diff
@@ -494,48 +501,35 @@ func semField(scope *Scope, e ast.Expr) (dag.Expr, error) {
...
-			Kind: "Filter",
-			Expr: c,
-		}, nil
+		return nil, nil
 	}
 	var e dag.Expr
 	if len(call.Args) > 1 {
+		if call.Name == "min" || call.Name == "max" {
+			// min and max are special cases as they are also functions. If the
+			// args are greater than 1 they're probably a function so do not
```
nwt, line 512:
```suggestion
			// number of args is greater than 1 they're probably a function so do not
```

docs/language/ztests/language-operators-1.yaml
```diff
@@ -6,10 +6,6 @@
 # See https://github.com/brimdata/zed/pull/2901 for details
 # ============================================================================
 
-# The Zed being tested below has been modified from the way it appears in
-# docs/language/README.md, as "put" is not yet optional
-# (https://github.com/brimdata/zed/issues/2850).
-
 script: |
   zc -C 'widget | count() by color | put COLOR := to_upper(color)'
```
nwt, line 10: If you're going to delete the comment about #2850, then you need to delete the `put` here!
```suggestion
  zc -C 'widget | count() by color | COLOR := to_upper(color)'
```

expr/function/function.go
```diff
@@ -123,6 +123,18 @@ func New(zctx *zson.Context, name string, narg int) (Interface, bool, error) {
 	return f, root, nil
 }
 
+// HasBoolResult returns true if the function name returns a boolean value.
```
nwt, line 126:
```suggestion
// HasBoolResult returns true if the function name returns a Boolean value.
```

expr/function/function.go
```diff
@@ -123,6 +123,18 @@ func New(zctx *zson.Context, name string, narg int) (Interface, bool, error) {
...
+// XXX This is a hack so the semantic compiler can determine if a single call
+// expr is a Filter or Put proc. At some point function declarations should have
+// signatures the return type can be introspected.
```
nwt, line 129:
```suggestion
// signatures so the return type can be introspected.
```

PR #2976, opened 2021-08-30
add branches to lake model

### LINKED ISSUE #2537: new zed lake: finish merge command
The new zed lake supports a merge operation that takes a set of tags, reads the commits/objects into a new commit, and outputs the new commit tag.

This task involves implementing the stub in `cmd/zed/lake/merge`.  It is very simple.  Just copy code from `zed lake query` and take stuff out.  It should take 15 minutes to code and a bit more time to write some tests.

### LINKED ISSUE #2626: Git-style branches in Zed lakes
In watching the reactions to the Git-style workflows being built for Zed lakes, a couple of us have thought about extending the metaphor all the way to actually having the concept of "branches". For instance, the Zed lake already provides for the ability to delete by commit to undo mistaken data imports. However, users accustomed to Git workflows might feel even safer working with Zed lakes if they could have the have the sense of "working in a branch". So if they were sitting down to work on shaping or importing a new data source into the lake, by working in such a branch they could do their data adds/deletes off a completely separate journal while feeling there's zero disk that their changes would muck up the "main" production journal that other users/apps are dependent on. If the work in their branch ends up being a failure, they could delete the branch and it'd be as if the data never touched the lake. Or if their work is a success and they want to incorporate the data they'd already shaped and cleaned up in their branch, they could merge it into the main journal.

In thinking this through, @mccanne had the following additional thoughts:

> A branch is just another journal object that points to the main journal and we could pretty easily write the code to  merge a branch journal into the main journal (or into another branch) and the write-conflict logic will be the same as merging commits from staging to main.

Given our need to focus on other priorities, this is feeling like a priority to consider in the MVP1 timeframe.

### LINKED ISSUE #2953: get rid of ScanPools()
lake.Root.ScanPools() should be removed along with the pools API end point but the app currently uses it.  We should coordinate with the app team to use the query endpoint (`from [pools]`) instead of the pools API end point, then we can get rid of this stuff.

### LINKED ISSUE #2975: add commit message to `lake load` endpoint
In the branches PR, we got rid of staging and the API endpoint for `lake load` now generates a dummy commit message because the HTTP body is the posted data.  We could (1) encode the commit message in the URL, (2) include a header containing the marshalled commit message in front of the posted data (yuck), or (3) use mime/multipart?

mccanne:
This commit adds a branch concept to the zed lake.  This required
some rewiring of the snapshot logic to implement merge and a new
approach to locking across two branches for a merge (though the
locking details are not yet implemented).  The TAIL pointer in a
journal now tracks the base pointer in a parent branch.  TAIL is
transactionally guaranteed through locking (TBD).

For now there is only one root branch (main), and branches must
all be descended from this.

"zed lake query" no longer reports "empty log" as it simply
produces no output.

"zed lake ls" no longer lists objects as meta-queries can be used
for this.  Instead, ls lists pools and branches.

The lake commands "add", "commit", and "status" are deprecated.
The lake command "load" does an automatic commit in the backend.

Closes #2626
Closes #2537
Closes #2975


mccanne:
I still need to update lake design.md... will work on that in the morning but wanted to get this up!

lake/branch.go
```diff
@@ -0,0 +1,441 @@
...
+func (b *BranchConfig) Path(poolPath *storage.URI) *storage.URI {
+	return poolPath.AppendPath(b.ID.String())
+}
+
+func (b *BranchConfig) Create(ctx context.Context, engine storage.Engine, poolPath *storage.URI, order order.Which, base journal.ID) error {
+	_, err := commit.Create(ctx, engine, b.Path(poolPath), order, base)
+	return err
+}
+
+func (b *BranchConfig) Open(ctx context.Context, engine storage.Engine, parent *storage.URI, pool *Pool) (*Branch, error) {
```
mattnibs, line 57: I think poolPath makes this clearer, and is consistent with the Create call:
```suggestion
func (b *BranchConfig) Open(ctx context.Context, engine storage.Engine, poolPath *storage.URI, pool *Pool) (*Branch, error) {
```

lake/pool.go
```diff
@@ -67,383 +59,181 @@ func (p *PoolConfig) Path(root *storage.URI) *storage.URI {
...
+	poolPath := p.Path(root)
+	// branchesPath is the path to the kvs journal of BranchConfigs
+	// for the pool while the commit log is stored in <pool-id>/<branch-id>.
+	branchesPath := poolPath.AppendPath(BranchesTag)
+	// create the branches journal store
+	_, err := kvs.Create(ctx, engine, branchesPath, branchStoreTypes)
+	if err != nil {
+		return err
+	}
+	// create the main branche in the branches journal store
```
mattnibs, line 74:
```suggestion
	// create the main branch in the branches journal store
```

lake/branch.go
```diff
@@ -0,0 +1,441 @@
...
+	log    *commit.Log
+	base   commit.View
+}
+
+func newBranchConfig(name string, parent ksuid.KSUID) *BranchConfig {
+	//XXX need to take parent args and check for a cycle.
+	// also check that parent journal ID exists in parent.
+	return &BranchConfig{
+		Name:   name,
+		ID:     ksuid.New(),
```
mattnibs, line 43: I'm kind of wondering why even bother having an ID for branches. I was moving through the zedlake fs and found it harder to follow what was going on with the ids for branches. I get why we did this for pool names- we didn't want to break links when pools were renamed. This feels like less of a problem with branches.

lake/branch.go
```diff
@@ -0,0 +1,441 @@
...
+		Parent: parent,
```
mattnibs, line 44: I don't grasp all the implications of doing this... but: It feels limiting to have a branch point to a parent branch vs doing what git does and having a branch to start at a particular commit. If we followed the way you have it coded up here, would you not be able to delete a branch that has child branches? What would happen to the child branches if their parent was deleted?

I suppose this would require making it so each commit referenced what its parent commit is. It would also mean that you'd have to keep commits centralized in pool... Did I mention I don't follow grasp the implications of this suggestion!?

service/ztests/add-empty.yaml
```diff
@@ -1,7 +1,7 @@
 script: |
   source service.sh
   zapi create -q -p test
-  zapi add -q -p test -
+  zapi load -q -p test -
```
mattnibs, line 4: Rename this file load-empty

service/ztests/add-garbage.yaml
```diff
@@ -1,7 +1,7 @@
 script: |
   source service.sh
   zapi create -q -p test
-  zapi add -q -p test -
```
mattnibs, line 4: Rename this file load-garbage?

service/core.go
```diff
@@ -138,16 +138,15 @@ func (c *Core) addAPIServerRoutes() {
...
-	c.authhandle("/pool/{pool}/add", handleAdd).Methods("POST")
-	c.authhandle("/pool/{pool}/delete", handleDelete).Methods("POST")
-	c.authhandle("/pool/{pool}/squash", handleSquash).Methods("POST")
-	c.authhandle("/pool/{pool}/staging", handleScanStaging).Methods("GET")
-	c.authhandle("/pool/{pool}/staging/{commit}", handleCommit).Methods("POST")
+	c.authhandle("/pool/{pool}/{branch}", handleBranchDelete).Methods("DELETE")
+	c.authhandle("/pool/{pool}", handleBranchPost).Methods("POST")
+	c.authhandle("/pool/{pool}/{branch}", handleBranchLoad).Methods("POST")
+	c.authhandle("/pool/{pool}/{branch}/merge", handleBranchMerge).Methods("POST")
+	c.authhandle("/pool/{pool}/{branch}/delete", handleDelete).Methods("POST")
```
mattnibs, line 149: Usually rest has the object name before id. Maybe put the word `branch` after the poolId and before branchId?

lake/branch.go
mccanne: Good point!

service/ztests/curl-stats.yaml
```diff
@@ -1,10 +1,9 @@
 script: |
   source service.sh
   curl -X POST -d '{"name":"test"}' http://$ZED_LAKE_HOST/pool > pool.json
-  poolID=$(zq -f text 'cut value.id' pool.json)
-  curl -X POST -d @- http://$ZED_LAKE_HOST/pool/$poolID/add > add.json
-  commitID=$(zq -f text 'cut value.commit' add.json)
-  curl -X POST -d '{}' http://$ZED_LAKE_HOST/pool/$poolID/staging/$commitID
+  poolID=$(zq -f text 'cut value.pool.id' pool.json)
+  branchID=$(zq -f text 'cut value.branch.id' pool.json)
+  curl -X POST -d @- http://$ZED_LAKE_HOST/pool/$poolID/$branchID > load.json
```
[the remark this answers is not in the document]
mccanne: But these two lines post the data into the branch to create the stats...?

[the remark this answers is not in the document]
mccanne: `-d @-` is the cryptic option...

api/client/connection.go
```diff
@@ -249,16 +249,30 @@ func (c *Connection) PoolRemove(ctx context.Context, id ksuid.KSUID) error {
...
-		t[i] = tag.String()
+func (c *Connection) BranchPost(ctx context.Context, poolID ksuid.KSUID, payload api.BranchPostRequest) (*Response, error) {
+	req := c.Request(ctx).
+		SetBody(payload)
+	req.Method = http.MethodPost
+	req.URL = path.Join("/pool", poolID.String())
+	resp, err := c.stream(req)
+	var errRes *ErrorResponse
+	if errors.As(err, &errRes) && errRes.StatusCode() == http.StatusConflict {
+		return nil, ErrPoolExists
```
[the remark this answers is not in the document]
mccanne: adding ErrBranchExists

api/client/connection.go
```diff
@@ -249,16 +249,30 @@ func (c *Connection) PoolRemove(ctx context.Context, id ksuid.KSUID) error {
...
-	req.Method = http.MethodGet
-	req.URL = path.Join("/pool", pool.String(), "staging")
-	return c.stream(req)
+		SetBody(api.BranchMergeRequest{at.String()})
+	req.Method = http.MethodPost
+	req.URL = path.Join("/pool", poolID.String(), branchID.String(), "merge")
+	resp, err := c.stream(req)
+	var errRes *ErrorResponse
+	if errors.As(err, &errRes) && errRes.StatusCode() == http.StatusConflict {
+		return nil, ErrPoolExists
```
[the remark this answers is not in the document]
mccanne: Deleting as this doesn't happen for a merge.

api/client/connection.go
```diff
@@ -208,13 +206,15 @@ func (c *Connection) PoolGet(ctx context.Context, id ksuid.KSUID) (*Response, er
...
 	req := c.Request(ctx)
-	req.Method = http.MethodGet
-	req.URL = path.Join("/pool", id.String(), "stats")
+	req.Method = http.MethodPost
+	req.URL = "/ids"
+	req.SetBody(api.IDsRequest{poolName, branchName})
 	r, err := c.stream(req)
 	var errRes *ErrorResponse
 	if errors.As(err, &errRes) && errRes.StatusCode() == http.StatusNotFound {
+		//XXX
```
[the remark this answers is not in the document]
mccanne: deleting

service/ztests/curl-stats.yaml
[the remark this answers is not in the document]
nwt: 🤦

lake/journal/queue.go
```diff
@@ -147,12 +155,16 @@ func (q *Queue) Open(ctx context.Context, head, tail ID) (io.Reader, error) {
 			return nil, err
 		}
 		if head == Nil {
-			return nil, ErrEmpty
+			// Return an empty reader when the journal is empty.
+			// This is preferred over returning ErrEmpty and
+			// havings layers above report the error message instead
+			// of simply processing an empty input without error.
+			return strings.NewReader(""), nil
```
mattnibs, line 162: 👍🏻

api/client/connection.go
```diff
@@ -31,6 +30,8 @@ var (
 	ErrPoolNotFound = errors.New("pool not found")
 	// ErrPoolExists returns when specified the pool already exists.
 	ErrPoolExists = errors.New("pool exists")
+	// ErrBranchExists returns when specified the branch already exists.
```
nwt, line 33:
```suggestion
	// ErrBranchExists is returned when the specified branch already exists.
```

mattnibs (approved):

cmd/zed/lake/ls/command.go
```diff
@@ -64,10 +65,13 @@ func (c *Command) Run(args []string) error {
 		return err
 	}
 	var query string
-	if c.lakeFlags.PoolName == "" {
+	if poolName == "" {
 		query = "from [pools]"
 	} else {
-		query = fmt.Sprintf("from '%s'[objects]", c.lakeFlags.PoolName)
+		if strings.IndexByte(poolName, '\'') >= 0 {
```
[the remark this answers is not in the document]
mccanne: At some point we should define the legal syntax for a pool/branch name.  For now, this is just protecting against a printf-d syntax error regarding the single quote so I will leave it as is.

lake/branch.go
mccanne: Good point but as we discussed... leaving fo now

lake/branch.go
mccanne: Branches v2 coming soon!

service/core.go
mccanne: But the resource is named `/pool/poolID/branchID` then we are applying a verb to that resource, which is not exactly REST-y but I don't see how swapping it makes more REST-y sense, i.e., applying say the verb `delete` to the pool with an branchID under that...?

api/client/connection.go
mccanne: caught the others there too...

mccanne:
Ok, I'll hold off on merging until app integration tests pass...

lake/api/api.go
```diff
@@ -15,28 +17,110 @@ import (
...
+		return nil, fmt.Errorf("internal error: multiple pools found with same name: %s", name)
+	}
+}
+
+func LookupPoolByID(ctx context.Context, api Interface, id ksuid.KSUID) (*lake.PoolConfig, error) {
+	d := newQueryDriver(lake.PoolConfig{})
+	zed := fmt.Sprintf("from [pools] | id == from_hex('%s')", idToHex(id))
+	_, err := api.Query(ctx, d, zed)
+	if err != nil {
+		fmt.Println("XERR", err)
```
nwt, line 65:
```suggestion
```

lake/api/driver.go
```diff
@@ -0,0 +1,41 @@
+package api
+
+import (
+	"github.com/brimdata/zed/api"
+	"github.com/brimdata/zed/zbuf"
+	"github.com/brimdata/zed/zson"
+)
+
+type queryDriver struct {
```
nwt, line 9: Nit: Call this file `querydriver.go` to match this.

lake/api/driver.go
```diff
@@ -0,0 +1,41 @@
...
+		d.results = append(d.results, v)
+	}
+	return nil
+}
+
+func (*queryDriver) Warn(msg string) error {
+	return nil
+}
+
+func (*queryDriver) ChannelEnd(channelID int) error {
```
nwt, line 35:
```suggestion
func (*queryDriver) Warn(string) error {
	return nil
}

func (*queryDriver) ChannelEnd(int) error {
```

cmd/zed/lake/load/command.go
```diff
@@ -21,11 +21,14 @@ import (
...
 var Load = &charm.Spec{
 	Name:  "load",
-	Usage: "load [options] file|S3-object|- ...",
+	Usage: "load [options] -p pool[/branch] file|S3-object|- ...",
 	Short: "add and commit data to a pool",
 	Long: `
 The load command adds data to a pool and commits it in one automatic operation.
 See documentation on "zed lake add" and "zed lake commit" for more details.
+
+A branch must be specified with the -p option.  If a pool name is given
```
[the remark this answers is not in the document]
mccanne: "pool and branch"

lake/api/api.go
mccanne: oopsies

service/core.go
nwt: @mccanne: I think the suggestion is to change `/pool/{pool}/{branch}` to `/pool/{pool}/branch/{branch}`. (I'm in favor.)

service/ztests/curl-pool-get.yaml
```diff
@@ -1,12 +1,15 @@
+skip: GET /pool/* is deprecated
```
nwt, line 1: Revert the changes here and keep this test until we kill `GET /pool` and `GET /pool/{pool}`?

service/handlers_test.go
```diff
@@ -322,6 +314,7 @@ func TestRequestID(t *testing.T) {
 		assert.NotEqual(t, "", res2.Header().Get("X-Request-ID"))
 	})
 	t.Run("PropagatesID", func(t *testing.T) {
+		t.Skip("GET /pool is deprecated")
```
nwt, line 317:
```suggestion
```

service/handlers_test.go
```diff
@@ -313,6 +304,7 @@ func TestNoEndSlashSupport(t *testing.T) {
 func TestRequestID(t *testing.T) {
 	ctx := context.Background()
 	t.Run("GeneratesUniqueID", func(t *testing.T) {
+		t.Skip("GET /pool is deprecated")
```
nwt, line 307:
```suggestion
```

service/handlers_test.go
```diff
@@ -246,9 +237,9 @@ func TestPoolStatsNoData(t *testing.T) {
 
 func TestPoolPostNameOnly(t *testing.T) {
 	_, conn := newCore(t)
-	pool := conn.TestPoolPost(api.PoolPostRequest{Name: "test"})
-	assert.Equal(t, "test", pool.Name)
-	assert.NotEqual(t, "", pool.ID)
+	poolID, _ := conn.TestPoolPost(api.PoolPostRequest{Name: "test"})
+	//assert.Equal(t, "test", pool.Name)
```
nwt, line 241:
```suggestion
```

lake/api/remote.go
```diff
@@ -155,49 +166,15 @@ func unmarshal(res *client.Response, i interface{}) error {
...
-func (r *RemoteSession) Delete(ctx context.Context, poolID ksuid.KSUID, tags []ksuid.KSUID, commit *api.CommitRequest) (ksuid.KSUID, error) {
-	res, err := r.conn.Delete(ctx, poolID, tags)
+func (r *RemoteSession) Delete(ctx context.Context, poolID, branchID ksuid.KSUID, tags []ksuid.KSUID, commit *api.CommitRequest) (ksuid.KSUID, error) {
+	res, err := r.conn.Delete(ctx, poolID, branchID, tags)
 	if err != nil {
 		return ksuid.Nil, err
 	}
 	defer res.Body.Close()
-	var staged api.StagedCommit
+	var staged api.CommitResponse
```
nwt, line 175: Nit: `staged` no longer seems like the right name for this.

lake/commit/snapshot.go
```diff
@@ -19,6 +18,7 @@ var (
 type View interface {
 	Lookup(ksuid.KSUID) (*segment.Reference, error)
 	Select(extent.Span, order.Which) Segments
+	SelectAll() Segments
```
nwt, line 21: Doesn't have to happen now but `SelectAll` should go away since it's equivalent to `Select` with a nil `extent.Span`.

lake/mock/mock.go
```diff
@@ -46,9 +48,11 @@ func NewPool(s string) (Pool, error) {
 		}
 	}
 	return Pool{
-		name:   name,
-		id:     fakeID(name),
-		layout: layout,
+		name:     name,
+		id:       fakeID(name),
+		branch:   branchName,
+		branchID: fakeID(branchName),
```
nwt, line 54: Nit: Initialize fields in the order they're delcared.
```suggestion
		name:     name,
		branch:   branchName,
		id:       fakeID(name),
		branchID: fakeID(branchName),
```

service/core.go
```diff
@@ -138,21 +138,22 @@ func (c *Core) addAPIServerRoutes() {
...
```
nwt, line 149: Nit: Keep these sorted by path.

nwt (approved):

lake/branch.go
mccanne: They will go away in branches v2.

service/core.go
mattnibs: What @nwt said

PR #2985, opened 2021-08-31
Limit scope of reserved words

### LINKED ISSUE #2398: Zed: Make the reserved words as limited in scope as possible

mattnibs:
Allow IDs to have the name of reserved words (e.g. by, from, where, etc).

In order to make this possible change Where clauses to accept a
LogicalOrExpr instead of BooleanSearch as it was previously doing. The
BooleanSearch made it impossible to distinguish between a keyword such
as by and limit and an ID search primitive.

Closes #2398


compiler/parser/parser.peg
```diff
@@ -148,7 +148,7 @@ SearchOrTerm = _ OrToken _ t:SearchAnd { RETURN(ARRAY("or", t)) }
 
 SearchAnd
   = first:SearchFactor
-    rest:((_ AndToken)? _ expr:SearchFactor{ RETURN(ARRAY("and", expr)) })* {
+    rest:(!(_ OrToken _) (_ AndToken)? _ expr:SearchFactor{ RETURN(ARRAY("and", expr)) })* {
```
nwt, line 151:
```suggestion
    rest:((_ AndToken)? _ !OrToken expr:SearchFactor{ RETURN(ARRAY("and", expr)) })* {
```

compiler/parser/parser.peg
```diff
@@ -876,7 +876,7 @@ SQLFrom
 
 SQLAlias
   = _ AS _ id:Lval { RETURN(id) }
-  / _ id:Lval { RETURN(id) }
+  / !(_ SQLTokenSentinels)_ id:Lval { RETURN(id) }
```
nwt, line 879:
```suggestion
  /  _ !SQLTokenSentinels id:Lval { RETURN(id) }
```

compiler/parser/ztests/reserved-word-ids.yaml
```diff
@@ -0,0 +1,8 @@
+zed: 'layout.order == "p1" | put from:="hi"'
+
+input: |
+  {layout:{order:"p1"}}
+
+output: |
+  {layout:{order:"p1"},from:"hi"}
+
```
nwt, line 8:
```suggestion
```

nwt (approved):

compiler/parser/parser.peg
mattnibs: Unless I am missing something you need to ensure the space after the SQLTokenSentinels exists otherwise you can't use ids that start with a token like `fromt`.

compiler/parser/parser.peg
mattnibs: Wouldn't this mean that you could have a search like: `x>1 order>1`?

compiler/parser/parser.peg
nwt: Oh. Tests passed with this change so I figured that meant it was a winner.

mattnibs: I should probably write a test where this fails!

compiler/parser/parser.peg
nwt: Tests passed with this change so I figured that meant it was a winner.

mattnibs: Nevermind, looks like you are right.

compiler/parser/parser.peg
mattnibs: I just tested this out and this one needs to remain as written, but I'll add test for this.

PR #2986, opened 2021-08-31
Post-zed/2967 "put" doc updates

philrz:
Now that #2967 is merged, I re-read the relevant docs as a user and noticed a couple adjustments that seem appropriate.

### DIFF
--- a/docs/language/operators/README.md
+++ b/docs/language/operators/README.md
@@ -685,7 +685,7 @@ time                        uid
 
 |                           |                                                 |
 | ------------------------- | ----------------------------------------------- |
-| **Description**           | Add/update fields based on the results of an expression<br><br>The `put` keyword is optional, though its explicit use may be appropriate to disambiguate from contexts where `:=` is used for [naming fields to hold aggregation results](../aggregate-functions/README.md#field-naming). |
+| **Description**           | Add/update fields based on the results of an expression<br><br>The `put` keyword is optional, though its explicit use may be appropriate to disambiguate from where `:=` is used for [naming fields to hold aggregation results](../aggregate-functions/README.md#field-naming). |
 | **Syntax**                | `[put] <field> := <expression> [, <field> := <expression> ...]`   |
 | **Required arguments**    | `<field>`<br>Field into which the result of the expression will be stored.<br><br>`<expression>`<br>A valid Zed [expression](../expressions/README.md). If evaluation of any expression fails, a warning is emitted and the original record is passed through unchanged. |
 | **Optional arguments**    | None |
--- a/docs/language/operators/README.md
+++ b/docs/language/operators/README.md
@@ -685,7 +685,7 @@ time                        uid
 
 |                           |                                                 |
 | ------------------------- | ----------------------------------------------- |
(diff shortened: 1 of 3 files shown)


docs/language/operators/README.md
```diff
@@ -685,8 +685,8 @@ time                        uid
 
 |                           |                                                 |
 | ------------------------- | ----------------------------------------------- |
-| **Description**           | Add/update fields based on the results of an expression         |
-| **Syntax**                | `put <field> := <expression> [, <field> := <expression> ...]`   |
+| **Description**           | Add/update fields based on the results of an expression<br><br>The `put` keyword is optional, though its explicit use may be appropriate to disambiguate from where `:=` is used for [naming fields to hold aggregation results](../aggregate-functions/README.md#field-naming). |
```
philrz, line 688: @mattnibs @nwt @mccanne: Please keep me honest here! I _think_ I've listed the only other context in the language where `:=` comes up, but I sometimes have difficulty keeping it all mentally cached.

mattnibs: No you do not need to use `:=`. For implied puts the semantic compiler is able to figure out if a call is a normal function (i.e. a call in a put) or an aggregation call (i.e. part of summarize proc).

docs/language/operators/README.md
```diff
@@ -685,8 +685,8 @@ time                        uid
...
+| **Syntax**                | `[put] <field> := <expression> [, <field> := <expression> ...]`   |
```
mattnibs, line 689: You can also just have a call here where a field with the name of the call is create: `to_lower(x)`

docs/language/operators/README.md
```diff
@@ -709,6 +709,25 @@ id.orig_h     id.orig_p id.resp_h       id.resp_p orig_bytes resp_bytes total_by
 ...
 ```
 
+#### Example #2:
+
+As noted above, the `put` keyword is optional. Our prior example could be
+abbreviated as shown below and return the same result.
+
+```mdtest-command zed-sample-data/zeek-default
+zq -q -f table 'total_bytes := orig_bytes + resp_bytes | sort -r total_bytes | cut id, orig_bytes, resp_bytes, total_bytes' conn.log.gz
```
mattnibs, line 718: 👏

docs/language/operators/README.md
mattnibs: So a put is a list of either `<field> := <expression>` or `<call_expression>`

philrz:
Thanks for the feedback @mattnibs. All good points. I've pushed a new commit that attempts to incorporate your feedback. Please give it another look and let me know what you think. I confess I sometimes struggle to summarize these slightly-more-sophisticated syntaxes without going into a full narrative trying to explain how the parser works.

mattnibs:
> I confess I sometimes struggle to summarize these slightly-more-sophisticated syntaxes without going into a full narrative trying to explain how the parser works.

Hmm yeah. I have been thinking about this the last couple weeks while I've been doing more work on the grammar. We've got a lot shorthand tricks to allow you to type less in certain situations, but part of me wonders if this makes the language more difficult to grasp. Does the language have too many secret handshakes? Would be it better to design for slightly more typing but more consistency (e.g. making assignments require a left hand `x:=`)?

mattnibs (approved):

PR #2994, opened 2021-09-02
improve the branches model to be more git-like

mccanne:
This commit improves upon the recent prototype of branching for the
zed lake.  The new model is very much git-like, where commit objects
are pool-wide in scope and stored independently of the branch structure.

Branches now are simply named pointers to commit objects. Just like git.

The commit objects of a pool form a tree rooted at the null commit
and pointing backward from most recent object to oldest.

Scanning can be performed at any commit ID.  There need not be a
branch reference.  Obviously, this provides a form of time travel,
but time-based time travel will be soon implemented by the traveling
backward in the branches and pools tables so that the pool and branch
state can be accessed at any point in time.

All of the marshaling has been changed to use "package style"
types as this will avoid collision in naming.  (In the future,
we can also manging the Zed time names separately from the
Go struct names as you can bind by an arbitrary name.)

Branch merging is done by transactionally and writers can concurrently
commit to either branch while the merge is running.  Rebase will be
implemented in a future PR.  There is no way yet to move a branch
pointer (other deleting it and recreating it).  A rebase will allow
live ingest to a child branch while periodic merges may occur to a
parent then rebasing along with period rebasing of the child to the
tip of the parent.  There is no need to synchronize the merge and
the rebase as they will both be transactionally consistent. (This will
be implemented by rewriting any commit objects in the child past
its merge point in a optimistic fashion until a successful transaction
is written.)

The meta-query syntax is updated from pool/branch[meta] to
pool@brach:meta or pool@commit:meta.

We also added a "rawlog" meta-query to provide access to the
underyling pool and branch config journals (rather than just
getting a snapshot of the tip of the config values).

The lake branch command was added as a preferred way of creating
and deleting branches instead of using the lake create command
(which is no used only for pools).

There is a new "undo" command to revert a commit.  The delete
command is restricted to deleting individual objects and is
no longer overloaded to "delete" a commit.  (There were some
bugs in that logic anyway.)

Now that we have branches, we added branch annotations to the
zed lake log command so you can see here which branches point to
which commit objects.


api/client/connection.go
```diff
@@ -208,15 +211,14 @@ func (c *Connection) PoolStats(ctx context.Context, id ksuid.KSUID) (*Response,
...
 }
 
-func (c *Connection) IDs(ctx context.Context, poolName, branchName string) (*Response, error) {
+func (c *Connection) BranchGet(ctx context.Context, poolID ksuid.KSUID, branchName string) (*Response, error) {
 	req := c.Request(ctx)
-	req.Method = http.MethodPost
-	req.URL = "/ids"
-	req.SetBody(api.IDsRequest{poolName, branchName})
+	req.Method = http.MethodGet
+	req.URL = path.Join("/pool", poolID.String(), "branch", url.QueryEscape(branchName))
```
nwt, line 217: s/QueryEscape/PathEscape/

api/client/connection.go
```diff
@@ -340,24 +357,34 @@ func (c *Connection) IndexPost(ctx context.Context, id ksuid.KSUID, post api.Ind
...
+
+func (c *Connection) Delete(ctx context.Context, poolID ksuid.KSUID, branchName string, ids []ksuid.KSUID, message api.CommitMessage) (*Response, error) {
 	req := c.Request(ctx).
 		SetBody(ids)
+	if err := encodeCommitMessage(req, message); err != nil {
+		return nil, err
+	}
 	req.Method = http.MethodPost
-	req.URL = path.Join("/pool", poolID.String(), branchID.String(), "delete")
+	req.URL = path.Join("/pool", poolID.String(), "branch", branchName, "delete")
```
nwt, line 387: `branchName` should be path-encoded.

To avoid omissions like this, maybe say
```go
	req.URL = urlPath("pool", poolID.String(), "branch", branchName)
```
with
```go
func urlPath(elem ...string) string {
	var s string
	for _, e := range elem {
		s += "/" + url.PathEncode(e)
	}
	return path.Clean(s)
}
```

cli/lakeflags/flags.go
```diff
@@ -32,6 +34,23 @@ func (f *Flags) Branch() (string, string) {
...
 }
 
+func (f *Flags) FromSpec(meta string) (string, error) {
+	poolName, branchName := f.Branch()
+	if strings.IndexByte(poolName, '\'') >= 0 || strings.IndexByte(branchName, '\'') >= 0 {
+		return "", errors.New("pool name may not contain quote characters")
+	}
+	var s string
+	if _, err := parser.ParseID(branchName); err == nil {
+		s = fmt.Sprintf("'%s'@%s", poolName, branchName)
```
nwt, line 44:
```suggestion
		s = fmt.Sprintf("from '%s'@%s", poolName, branchName)
```

cmd/zed/lake/merge/command.go
```diff
@@ -46,31 +45,36 @@ func (c *Command) Run(args []string) error {
...
-	var at ksuid.KSUID
-	if c.at != "" {
-		at, err = parser.ParseID(c.at)
-		if err != nil {
-			return fmt.Errorf("bad -at option: %w", err)
-		}
-	}
-	commit, err := lake.MergeBranch(ctx, poolID, branchID, at)
+	_, err = lake.MergeBranch(ctx, poolID, childBranch, parentBranch, c.commitFlags.CommitMessage())
 	if err != nil {
```
nwt, line 73:
```suggestion
	if _, err := lake.MergeBranch(ctx, poolID, childBranch, parentBranch, c.commitFlags.CommitMessage()); err != nil {
```

api/client/connection.go
```diff
@@ -340,24 +357,34 @@ func (c *Connection) IndexPost(ctx context.Context, id ksuid.KSUID, post api.Ind
...
+		SetBody(r)
+	if err := encodeCommitMessage(req, message); err != nil {
 		return nil, err
 	}
-	req := c.Request(ctx).
-		SetBody(r).
-		SetHeader("Zed-Commit", string(encoded))
 	req.Method = http.MethodPost
-	req.URL = path.Join("/pool", poolID.String(), branchID.String())
+	req.URL = path.Join("/pool", poolID.String(), "branch", url.QueryEscape(branchName))
```
mattnibs, line 367:
```suggestion
	req.URL = path.Join("/pool", poolID.String(), "branch", url.PathEscape(branchName))
```

api/client/connection.go
mattnibs: Or use url.PathEscape

zfmt/dag.go
```diff
@@ -387,8 +387,8 @@ func source(src dag.Source) string {
 		return fmt.Sprintf("%s", p.ID)
 	case *dag.PoolMeta:
 		return fmt.Sprintf("%s[%s]", p.ID, p.Meta)
-	case *dag.BranchMeta:
-		return fmt.Sprintf("%s/%s[%s]", p.ID, p.Branch, p.Meta)
+	case *dag.CommitMeta:
+		return fmt.Sprintf("%s@%s[%s]", p.Pool, p.Commit, p.Meta)
```
nwt, line 391: Do we want to switch these from `[meta]` to `:meta`?

cmd/zed/lake/branch/command.go
```diff
@@ -38,13 +39,13 @@ func init() {
 
 type Command struct {
 	lake      zedlake.Command
-	at        string
+	drop      bool
```
mattnibs, line 42: Maybe call this `delete` instead since the all other text refers to the operation as "delete" not "drop".

service/request.go
```diff
@@ -49,8 +51,31 @@ func (r *Request) PoolID(w *ResponseWriter) (ksuid.KSUID, bool) {
...
+	if commitJSON != "" {
+		if err := json.Unmarshal([]byte(commitJSON), &message); err != nil {
+			w.Error(fmt.Errorf("load endpoint encountered invalid JSON in Zed-Commit header: %w", err))
+			return message, false
+		}
+	}
+	return message, true
+}
+
+func (r *Request) StringArg(w *ResponseWriter, arg string) (string, bool) {
```
nwt, line 70: Nit: Give this a name and signature in keeping with the other methods here.
```suggestion
func (r *Request) StringFromPath(arg string, w *ResponseWriter) (string, bool) {
```

cli/lakeflags/flags.go
```diff
@@ -18,7 +20,7 @@ func (f *Flags) SetFlags(fs *flag.FlagSet) {
 }
 
 func (f *Flags) Names() (string, string) {
-	if i := strings.LastIndexByte(f.PoolName, '/'); i > -1 {
+	if i := strings.LastIndexByte(f.PoolName, '@'); i > -1 {
```
mattnibs, line 23: I'd prefer to instead to have a `-ref` flag  instead of having to know you need to refer to a specific branch in the pool option written as `<pool>@<branch|commit>`. I think this might be a case where it's better to stick with how git works. Actually I'd be okay if you didn't have to specify the pool but if the `-ref` tag where written as `<pool>@<branch|commit>`, kind of like in git how you can write `refs/<remote>/<ref_name>`.

api/api.go
```diff
@@ -112,19 +102,17 @@ type PoolPutRequest struct {
...
 
 type BranchMergeRequest struct {
 	At string `json:"at"`
 }
 
-type CommitRequest struct {
-	Author  string  `zng:"author"`
-	Date    nano.Ts `zng:"date"`
-	Message string  `zng:"message"`
+type CommitMessage struct {
```
nwt, line 113: Nit: I think `Commit`, `CommitInfo`, or something like that would be a better name for this.

lake/segment/reference.go
```diff
@@ -94,7 +94,14 @@ func (r Reference) IsZero() bool {
...
 func (r Reference) String() string {
-	return fmt.Sprintf("%s %d records in %d data bytes", r.ID, r.Count, r.RowSize)
+	return fmt.Sprintf("%s %d record%s in %d data bytes", r.ID, r.Count, plural(int(r.Count)), r.RowSize)
+}
+
+func plural(ordinal int) string {
+	if ordinal == 1 {
+		return ""
+	}
+	return ""
```
nwt, line 104:
```suggestion
	return "s"
```

compiler/semantic/proc.go
```diff
@@ -157,37 +139,60 @@ func semPool(ctx context.Context, scope *Scope, p *ast.Pool, adaptor proc.DataAd
...
+		// We no longer use "at" to refer to a commit tag, but if there
+		// is no commit tag, we could use an "at" time argument to time
+		// travel by going back in the branch log and finding the commit
+		// object with the largest time stamp <= the at time.
+		// This would require commitRef to be branch name not a commit ID.
+		return nil, errors.New("TBD: at clause in from operator needs to use time")
+	}
+	commitRef := p.Spec.Commit
+	var commit ksuid.KSUID
+	if commitRef != "" {
```
nwt, line 154:
```suggestion
	var commit ksuid.KSUID
	if commitRef := p.Spec.Commit; commitRef != "" {
```

lake/api/remote.go
```diff
@@ -44,14 +46,22 @@ func newConnection(host string) *client.Connection {
...
+	return config.ID, nil
+}
+
+func (r *RemoteSession) CommitObject(ctx context.Context, poolID ksuid.KSUID, branchName string) (ksuid.KSUID, error) {
+	res, err := r.conn.BranchGet(ctx, poolID, branchName)
+	if err != nil {
+		return ksuid.Nil, err
+	}
+	var commit api.CommitResponse
+	err = unmarshal(res, &commit)
```
nwt, line 63:
```suggestion
	if err := unmarshal(res, &commit); err != nil {
		return ksuid.Nil, err
	}
```

lake/branch.go
```diff
@@ -90,273 +54,246 @@ func (b *Branch) Load(ctx context.Context, r zio.Reader, date nano.Ts, author, m
...
 		if err != nil {
 			return nil, err
 		}
-		if ok {
-			ids = append(ids, tag)
-			continue
+		object := commits.NewObject(parent.Commit, author, message, retries)
+		for _, id := range ids {
+			if !snap.Exists(id) {
+				return nil, fmt.Errorf("non-existant segment %s: delete operation aborted", id)
```
nwt, line 101:
```suggestion
				return nil, fmt.Errorf("non-existent segment %s: delete operation aborted", id)
```

lake/commits/snapshot.go
```diff
@@ -41,15 +43,15 @@ func NewSnapshot() *Snapshot {
...
+		return fmt.Errorf("%s: add of a duplicate data object: %w", id, ErrWriteConflict)
 	}
 	s.segments[id] = seg
 	return nil
 }
 
 func (s *Snapshot) DeleteSegment(id ksuid.KSUID) error {
 	if _, ok := s.segments[id]; !ok {
-		return ErrNotFound
+		return fmt.Errorf("%s: delete of a non-existant data object: %w", id, ErrWriteConflict)
```
nwt, line 54:
```suggestion
		return fmt.Errorf("%s: delete of a non-existent data object: %w", id, ErrWriteConflict)
```

lake/branch.go
```diff
@@ -90,273 +54,246 @@ func (b *Branch) Load(ctx context.Context, r zio.Reader, date nano.Ts, author, m
...
-	return b.log
-}
-
 func (b *Branch) ApplyIndexRules(ctx context.Context, rules []index.Rule, ids []ksuid.KSUID) (ksuid.KSUID, error) {
 	idxrefs := make([]*index.Reference, 0, len(rules)*len(ids))
 	for _, id := range ids {
 		//XXX make issue for this.
 		// This could be easily parallized with errgroup.
 		refs, err := b.indexSegment(ctx, rules, id)
+		fmt.Println("INDEX", id, len(refs))
```
nwt, line 286:
```suggestion
```

lake/commits/actions.go
```diff
@@ -39,31 +41,22 @@ func (a *Add) String() string {
 	return fmt.Sprintf("ADD %s", a.Segment)
 }
 
-type CommitMessage struct {
-	Commit  ksuid.KSUID `zng:"commit"`
+type Commit struct {
+	ID      ksuid.KSUID `zng:"id"`
+	Parent  ksuid.KSUID `zng:"parent"`
+	Retries uint8       `zng:"retries"`
```
nwt, line 47: What the story on this?

lake/commits/object.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/brimdata/zed/lake/index"
+	"github.com/brimdata/zed/lake/segment"
+	"github.com/brimdata/zed/pkg/nano"
+	"github.com/brimdata/zed/zngbytes"
+	"github.com/brimdata/zed/zson"
+	"github.com/segmentio/ksuid"
+)
+
+// ErrEmptyTransaction is the error returned by Transaction.Serialize when
+// Transaction.Actions is empty.
```
nwt, line 17: Fix or delete this comment.

api/client/connection.go
mccanne: I'm using @nwt's urlEscape() function ok?

cli/lakeflags/flags.go
mccanne: See new discussion at end of `docs/lake/design.md`.  This is complicated enough that I vote we merge this PR and resolve in a subsequent and soonish followup PR...

api/api.go
mccanne: But it's a message.  I was going to change the field name from Message to Body.  So it's a message with an author and a body.  Unless it changes to contain broader and less related information, I wouldn't call it generically "info".

mccanne: And "Commit" by itself implies to me an operation not an annotation.

lake/commits/actions.go
mccanne: I will move this comment from branch.go to actions.go...
```
		// Note that we store the number of retries in the final commit
		// object.  This will allow easily introspection of optimistic
		// locking problems under high commit load by simply issuing
		// a meta-query and looking at the retry count in the persisted
		// commit objects.  If/when this is a problem, we could add
		// pessimistic locking mechanisms alongside the optimistic approach.
```

lake/api/remote.go
```diff
@@ -65,29 +77,23 @@ func (r *RemoteSession) CreatePool(ctx context.Context, name string, layout orde
...
-func (r *RemoteSession) CreateBranch(ctx context.Context, poolID ksuid.KSUID, name string, parent, at ksuid.KSUID) (ksuid.KSUID, error) {
-	res, err := r.conn.BranchPost(ctx, poolID, api.BranchPostRequest{
-		Name:     name,
-		ParentID: parent.String(),
-		At:       at.String(),
+func (r *RemoteSession) CreateBranch(ctx context.Context, poolID ksuid.KSUID, name string, at ksuid.KSUID) error {
+	_, err := r.conn.BranchPost(ctx, poolID, api.BranchPostRequest{
+		Name:   name,
+		Commit: at.String(),
 	})
```
mattnibs, line 87:
```suggestion
	r, err := r.conn.BranchPost(ctx, poolID, api.BranchPostRequest{
		Name:   name,
		Commit: at.String(),
	})
	if err == nil {
		r.Body.Close()
	}
```
I started this pattern in a lot of places, but we need to still close the response if the err is nil otherwise the connection will not close. See https://pkg.go.dev/net/http#Client.Do

We should probably change the connection library to allows return a response Body so we can go without the error check.

api/api.go
nwt: The reason `CommitMessage` feels wrong to me is that it doesn't match the model in `lake/actions.Commit` (or Git or Mercurial), in which author and message are distinct attributes of the commit object (rather than author being an attribute of the message).

lake/commits/actions.go
nwt: Great idea!

nwt: (I hadn't read that comment. :grin:)

lake/journal/store.go
```diff
@@ -73,10 +93,19 @@ func (s *Store) load(ctx context.Context) error {
...
-			delete(table, e.Key)
-		} else {
-			table[e.Key] = e.Value
+		switch e := e.(type) {
+		case *Add:
+			table[e.Entry.Key()] = e.Entry
+		case *Update:
+			key := e.Key()
+			if _, ok := table[key]; !ok {
+				return fmt.Errorf("update to non-existant key in journal store: %T", key)
```
nwt, line 102:
```suggestion
				return fmt.Errorf("update to non-existent key in journal store: %T", key)
```

lake/branch.go
```diff
@@ -90,252 +54,223 @@ func (b *Branch) Load(ctx context.Context, r zio.Reader, date nano.Ts, author, m
...
+	plural := "s"
+	if len(segments) == 1 {
+		plural = ""
+	}
+	b.WriteString(fmt.Sprintf("loaded %d data object%s\n\n", len(segments), plural))
+	for k, segment := range segments {
+		b.WriteString("  ")
+		b.WriteString(segment.String())
+		b.WriteByte('\n')
+		if k >= 10 {
```
mattnibs, line 84:
```suggestion
		if k >= maxMessageObjects {
```

docs/lake/design.md
```diff
@@ -141,23 +141,27 @@ in the Zed language is referred to as "this".
...
 
+> Zed lakes can be used without branches.  When referencing a pool without
+> a branch, the tooling presumes the "main" branch as the default, and everything
+> can be done on main without having to think about branching.
+
 ### Branch
 
-A branch is created from a parent branch (often "main") with the `branch`
-command, e.g.,
+A branch is simply a named pointer to a commit object in the zed lake.
```
philrz, line 150:
```suggestion
A branch is simply a named pointer to a commit object in the Zed lake.
```

The rest of the doc is consistent about capitalizing "Zed" when talking about the project and `zed` when mentioning the command line tool.

cmd/zed/lake/undo/command.go
```diff
@@ -0,0 +1,82 @@
...
+
+	"github.com/brimdata/zed/cli/lakeflags"
+	zedapi "github.com/brimdata/zed/cmd/zed/api"
+	zedlake "github.com/brimdata/zed/cmd/zed/lake"
+	"github.com/brimdata/zed/compiler/parser"
+	"github.com/brimdata/zed/pkg/charm"
+)
+
+var Undo = &charm.Spec{
+	Name:  "undo",
```
mattnibs, line 16: I think `revert` is a better name for this.

docs/lake/design.md
```diff
@@ -141,23 +141,27 @@ in the Zed language is referred to as "this".
...
+zed lake branch -p logs@main staging
 ```
-This creates a new branch called "branch" in pool "logs" whose parent
-is branch "main" forked at the tip of main.  A branch may be forked at
-any point in its parents commit history using the `-at` option.
-
-Note that when specifying a pool, the main branch is assumed unless overridden
-with a branch specifier using the trailing `/branch` syntax described below.
+This creates a new branch called "staging" in pool "logs", which points to
+same commit object as the "main" branch.  Commits to the "staging" branch
```
philrz, line 158:
```suggestion
the same commit object as the "main" branch.  Commits to the "staging" branch
```

lake/root.go
```diff
@@ -26,16 +26,15 @@ import (
 )
 
 var (
-	ErrPoolExists     = errors.New("pool already exists")
-	ErrPoolNotFound   = errors.New("pool not found")
+	//ErrPoolExists = errors.New("pool already exists")
+	//ErrPoolNotFound   = errors.New("pool not found")
 	ErrBranchExists   = errors.New("branch already exists")
 	ErrBranchNotFound = errors.New("branch not found")
 )
```
nwt, line 33: The errors in `lake/pools` make these branch errors redundant, so all of this should go away.

docs/lake/design.md
```diff
@@ -141,23 +141,27 @@ in the Zed language is referred to as "this".
...
+will be added to the commit history without affecting the "main" branch
+and each branch can be queried indepedently at any time.
```
philrz, line 160:
```suggestion
and each branch can be queried independently at any time.
```

lake/commits/patch.go
```diff
@@ -0,0 +1,134 @@
...
+	for _, id := range p.deletes {
+		o.AppendDelete(id)
+	}
+	for _, s := range p.diff.segments {
+		o.appendAdd(s)
+	}
+	return o
+}
+
+//XXX we need to handle more than add/delete
```
mattnibs, line 99: Created an issue for this
```suggestion
//XXX we need to handle more than add/delete #3000 
```

lake/branch.go
```diff
@@ -90,252 +54,223 @@ func (b *Branch) Load(ctx context.Context, r zio.Reader, date nano.Ts, author, m
...
 		}
-		// Note that we only search in this branch not into the parent
-		// since this lookup is used for indexing, deleting, etc and
-		// the parent branch should be used in such cases.
-		snap, ok, err := b.log.SnapshotOfCommit(ctx, 0, tag)
+		return object, nil
+	})
+}
+
+func (b *Branch) Undo(ctx context.Context, commit ksuid.KSUID, author, message string) (ksuid.KSUID, error) {
```
mattnibs, line 109: I mentioned this in undo/command.go but I think this should be called Revert.

docs/lake/design.md
```diff
@@ -174,43 +178,68 @@ zed lake load -p logs -i zst sample5.zst
...
+starting from any commit.  The log can be displayed with the `log` command,
+e.g.,
+```
+zed lake log -p pool@branch
+```
+To understand the log contents, the `load` operation is actually
 decomposed into two steps under the covers:
 an `add` step stores one or more
 new immutable data objects in the lake and a `commit` operation
 materialize the objects into a branch with an ACID transaction.
```
philrz, line 218:
```suggestion
materializes the objects into a branch with an ACID transaction.
```

docs/lake/design.md
```diff
@@ -174,43 +178,68 @@ zed lake load -p logs -i zst sample5.zst
...
+Every commit object and data object is named by and referenced
+using globally unique [KSUIDs](https://github.com/segmentio/ksuid),
+called a `commit ID` or a data `object ID`, respectively.
+
+The log command prints the commit ID of each commit object in that path
+from the current pointer back through history to the first commit object.
+
+A commit object includes
 an optional author and message, along with a required timestamp,
 that is stored in the commit journal for reference.  This values may
```
philrz, line 240:
```suggestion
that is stored in the commit journal for reference.  These values may
```

docs/lake/design.md
```diff
@@ -225,55 +254,40 @@ every journal entry, which has a Zed type signature as follows:
...
+zed lake merge -p logs@updates main
 ```
-A merge operation applies all of the commits in the branch to the tip of
-its parent and transactionally performs this commit while moving the branch
-to its tip.  While the merge operation is performed, data can still be written
-to the branch and queries performed.  Newly written data remains in the
+where the "updates" branch is being merged into the "main" branch.
+
+A merge operation finds a common ancestor in the commit history then
+computes the set of changes needed for target branch to reflect the
```
philrz, line 273:
```suggestion
computes the set of changes needed for the target branch to reflect the
```

docs/lake/design.md
```diff
@@ -387,14 +402,23 @@ If a writer commits data after a reader starts scanning, then the reader
 does not see the new data since it's scanning the snapshot that existed
 before these new writes occur.
```
philrz, line 403:
```suggestion
before these new writes occurred.
```

lake/commits/store.go
```diff
@@ -0,0 +1,279 @@
...
+
+type Store struct {
+	path   *storage.URI
+	engine storage.Engine
+	// We use these poor-man's caches for now since chasing linked lists
+	// over cloud storage will run very slowly... there is an interesting
+	// scaling problem to work on here.
+	cache     map[ksuid.KSUID]*Object
+	snapshots map[ksuid.KSUID]*Snapshot
+	paths     map[ksuid.KSUID][]ksuid.KSUID
```
nwt, line 32: Use github.com/hashicorp/golang-lru for these.

docs/lake/design.md
```diff
@@ -672,8 +642,8 @@ sort of the KSUIDs results in a creation-time ordering (though this ordering
 is not relied on for causal relationships since clock skew can violate
 such an assumption).
 
-Data objects are referred to by zero or more commits, where the commits
-are maintained in a commit journal described below.
+Data objects are referred to by zero or more commit objects, where the =
```
philrz, line 645:
```suggestion
Data objects are referred to by zero or more commit objects, where the
```

docs/lake/design.md
```diff
@@ -672,8 +642,8 @@ sort of the KSUIDs results in a creation-time ordering (though this ordering
...
+commits are maintained referenced by the branches journal described below.
```
philrz, line 646: I'm not sure what suggestion to make to fix this sentence, but "maintained referenced by" reads funny.

docs/lake/design.md
```diff
@@ -725,6 +695,59 @@ a subset of data.
...
+
+A pool's commit history is the definitive record of the evolution of data in
+that pool in a transactionally consistent fashion.
+
+Each commit object entry is identified with its `commit ID`.
+Objects are immutable and uniquely named so there is never a concurrent write
+condition.
+
+The `add` and `commit` operations are transactionally stored
+in a chain of commit object.  Any number of adds (and deletes) may appear
```
philrz, line 708:
```suggestion
in a chain of commit objects.  Any number of adds (and deletes) may appear
```

docs/lake/design.md
```diff
@@ -725,6 +695,59 @@ a subset of data.
...
+in a commit object.  All of the operations that belong to a commit are
+identified with a commit identifier (ID).
+
+As each commit object points to its parent (except for the initial commit
+in main), the collection commit objects in a pool forms a tree.
```
philrz, line 713:
```suggestion
in main), the collection of commit objects in a pool forms a tree.
```

docs/lake/design.md
```diff
@@ -725,6 +695,59 @@ a subset of data.
...
+By default, `zed lake log` outputs an abbreviated form of the log as text to
+stdout, similar to the output of `git log`.
+
+However, the log represents the definitive record of a pool's present
+and historical content, and accessing its complete detail can provide
+insights about data layout, provenance, history, and so forth.  Thus,
+Zed lake provides a means to query a pool's configuration state as well,
+thereby allowing past versions of the complete pool and branch configurations
+as well as all of their underlying data to be subject to time travel.
+To interrogate the underlying trasaction history of the branches and
```
philrz, line 736:
```suggestion
To interrogate the underlying transaction history of the branches and
```

lake/commits/store.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+func (s *Store) Snapshot(ctx context.Context, leaf ksuid.KSUID) (*Snapshot, error) {
+	if snap, ok := s.snapshots[leaf]; ok {
+		return snap, nil
+	}
+	var objects []*Object
+	var base *Snapshot
+	at := leaf
+	for at != ksuid.Nil {
```
nwt, line 92:
```suggestion
	for at := leaf; at != ksuid.Nil; {
```

docs/lake/design.md
```diff
@@ -978,3 +967,69 @@ The initial prototype has been simplified as follows:
 * transaction journal incomplete
 * no recursive journal pool
 * no columnar support
+
+## CLI tool naming conventions
+
+> This is a work in in progress.
```
philrz, line 973:
```suggestion
> This is a work in progress.
```

docs/lake/design.md
```diff
@@ -978,3 +967,69 @@ The initial prototype has been simplified as follows:
...
+Confusion arises because a commit can be referred to by a branch name
+and, depending on context, a branch can refer to a commit or to
+a mutable reference of that branch.  For the technical discussion here, we will
+refer to the branch/commit dichotomy as a `commitish` and the l-value
+nature of a branch as a `ref`.
+
+Both a `commitish` and a `ref` may drop omit `@` suffix
+and refer solely to the pool name, in which case it becomes pool@main.
+
+Complicating matters further, a `ref` may be specified a simple name
```
philrz, line 1002:
```suggestion
Complicating matters further, a `ref` may be specified via a simple name
```

docs/lake/design.md
```diff
@@ -978,3 +967,69 @@ The initial prototype has been simplified as follows:
...
+* delete `ref` `id` `id` ... - delete data objects from branch `ref` and update the branch to point to the new commit object
+* drop `pool` - delete a data pool
+* index `ref` `id` `id` ... - create index objects if they don't exist and record them in branch `ref`
+* load `ref` `file` ... - load the files or stdin into data objects, create a commit object for the new data that points backward to the head of `ref`, and update branch `ref` to point to the new commit
+* log `commitish`- display the commit object history starting at `commitish`
+* ls - list pools in a lake
+* ls `pool` - list branch names of a pool
+* merge `commitish` `sref` - merge a commit history rooted at `commitish` into a branch `sref`
+* query: `from commitish` - data scan
+* query: `from commitish:meta` branch or comit meta-scan (-no defaulting main here because thats pool-level meta)
```
philrz, line 1026:
```suggestion
* query: `from commitish:meta` branch or commit meta-scan (-no defaulting main here because that's pool-level meta)
```

api/api.go
mccanne: Ok, how about we fix it when we do #3001 as it will have a different look and feel then...

cmd/zed/lake/drop/command.go
```diff
@@ -54,26 +52,17 @@ func (c *Command) Run(args []string) error {
 	}
 	defer cleanup()
 	poolName, branchName := c.lakeFlags.Names()
-	if poolName == "" {
+	if poolName == "" || branchName != "" {
 		return errors.New("name of pool must be supplied with -p option")
 	}
```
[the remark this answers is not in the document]
mccanne: I thought the one error was sufficient but will change it as "name of pool" is not a branch.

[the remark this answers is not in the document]
mccanne:
```
branch name not allowed; use the branch command to delete branches
```

lake/api/remote.go
mccanne: Made the change but I had the opposite reaction on the client lib ... the client only needs the response body if it's going to stream back the result.  Otherwise, the library should do all the boiler plate to unmarshal expected structure and return meaningful values instead of a generic response object.

lake/branch.go
```diff
@@ -90,252 +54,223 @@ func (b *Branch) Load(ctx context.Context, r zio.Reader, date nano.Ts, author, m
...
+		// should live in a single tree.
+		//XXX hmm, except if you branch main when it is empty...?
+		// we shoudl detect this and not allow it...?
+		return nil, errors.New("system error: cannot locate common ancestor for branch merge")
+	}
+	// Compute the snapshot of the common ancestor then compute patches
+	// along each branch and make sure the two patches do not have a
+	// delete conflict.  For now, this is the only kind of merge update
+	// conflict we detect.
+	base, err := b.pool.commits.Snapshot(ctx, baseID)
```
mattnibs, line 166: What's the point of getting the base Snapshot? Should you just be able to just collect deletes from the two tips to the baseID and make sure there's no conflict between the two?

lake/commits/patch.go
mccanne: Lucky ticket #3000.  You win a free beer!

lake/commits/store.go
mccanne: I made issue #3002 for this.  Sizing the caches and configuring limits will need some thought.  And the approach may soon change anyway with a design for persisting snapshots.  This is definitely a scaling issue right now, but totally solvable.

lake/branch.go
```diff
@@ -90,252 +54,223 @@ func (b *Branch) Load(ctx context.Context, r zio.Reader, date nano.Ts, author, m
...
+			}
+			return false
+		}
+		if err := b.pool.branches.Update(ctx, config, parentCheck); err != nil {
+			if err == journal.ErrConstraint {
+				// Parent check failed so try again.
+				continue
+			}
+			return ksuid.Nil, err
+		}
```
mattnibs, line 245: Should we delete the object we wrote to the commit store if we fail / need to retry, or will this get cleaned up with garbage collection?

docs/lake/design.md
mccanne: deleting

lake/branches/store.go
```diff
@@ -0,0 +1,106 @@
...
+	list, err := s.All(ctx)
+	if err == nil {
+		for k, config := range list {
+			if config.Commit == commit {
+				return &list[k], nil
+			}
+		}
+	}
+	return nil, fmt.Errorf("%s: %w", commit, ErrNotFound)
+}
```
mattnibs, line 64: This isn't in use, also it seems kind of weird to look up a branch by its commit, no?

lake/commits/store.go
```diff
@@ -0,0 +1,279 @@
...
+		}
+		if at == to {
+			break
+		}
+		at = o.Parent
+	}
+	return path, nil
+}
+
+func (s *Store) GetBytes(ctx context.Context, commit ksuid.KSUID) ([]byte, *Commit, error) {
```
mattnibs, line 162: I would move this next to Get and Get should probably just call GetByes and return the Commit. Also should this populate the cache?

lake/branch.go
```diff
@@ -90,252 +54,223 @@ func (b *Branch) Load(ctx context.Context, r zio.Reader, date nano.Ts, author, m
...
-	txn := commit.NewAddsTxn(id, w.Segments())
-	txn.AppendCommitMessage(id, date, author, message)
-	if _, err := b.log.Commit(ctx, txn); err != nil {
-		return ksuid.Nil, err
-	}
-	return id, nil
+		return ksuid.Nil, commits.ErrEmptyTransaction
+	}
+	if message == "" {
+		message = loadMessage(segments)
```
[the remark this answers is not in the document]
mccanne: I don't agree.  This is the start of data lineage and provenance.  It also makes for bad demos to have blanks.

lake/api/remote.go
mattnibs: I agree.

lake/commits/object.go
```diff
@@ -0,0 +1,142 @@
...
+	o.Append(&Add{Commit: o.Commit, Segment: *s})
+}
+
+func (o *Object) AppendDelete(id ksuid.KSUID) {
+	o.Append(&Delete{Commit: o.Commit, ID: id})
+}
+
+func (o *Object) appendAddIndex(i *index.Reference) {
+	o.Append(&AddIndex{Commit: o.Commit, Index: *i})
+}
```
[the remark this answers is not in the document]
mccanne: indeed

lake/branch.go
```diff
@@ -90,252 +54,223 @@ func (b *Branch) Load(ctx context.Context, r zio.Reader, date nano.Ts, author, m
...
-		// into into the patch.
-		patch := commit.NewPatch(b.base)
-		if err := patch.PlayLog(ctx, b.log, 0); err != nil {
-			return nil, err
-		}
-		return patch, nil
+	parentPath, err := b.pool.commits.Path(ctx, parent.Commit)
+	if err != nil {
+		return nil, err
 	}
```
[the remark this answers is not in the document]
mccanne: This is part of the design-for-scale plan.  For now, I think brute force is easy and correct and will work to a pretty large scale before it falls over.  And anyway, other things will get in the way first.  Here is info on the algorithmic problem:

https://en.wikipedia.org/wiki/Lowest_common_ancestor

Tarjan worked out many non-obvious solutions on this front.

lake/branch.go
mccanne: You need to compare paths to detect write conflicts.  The tip is insufficient.  Right now, it's just detecting simple delete conflicts but this code path will get more sophisticated with fancier update models.

mattnibs (approved):
Nice!

lake/branch.go
mccanne: Very good point!

lake/commits/store.go
mccanne: The cache doesn't save the zng buffer.  These are two different models.  I think it will change anyway a bit when we persist the snapshots so let's just leave this for now.

mccanne:
Will await app changes and integration tests before merging...

api/api.go
nwt: 👍

docs/lake/design.md
```diff
@@ -374,27 +392,37 @@ zed lake query -f lake "from logs/live[branches]"
...
+before these new writes occurred.
 
 Also, arbitrary metadata can be committed to the log as described below,
 e.g., to associate index objects or derived analytics to a specific
 journal commit point potentially across different data pools in
 a transactionally consistent fashion.
 
+#### Time Travel
+
+While time-travel through commit history provides one means to explore
```
nwt, line 416:
```suggestion
While time travel through commit history provides one means to explore
```

nwt (approved):

PR #3004, opened 2021-09-03
allow const identifiers in literal switch cases

### LINKED ISSUE #3003: consts in literal switch cases
The literal version of switch should be able to take constant identifiers in addition to literals.  e.g., while this works because it's expression version of switch cases:
```
const s1=<big type value>
switch (
  typeof(this)==s1 => postal_code:=int64(zip) | drop zip ;
  default => pass ;
)
```
This variation of it does not parse
```
const s1=<big type value>
switch (
  s1 => postal_code:=int64(zip) | drop zip ;
  default => pass ;
)
```
We need to change the parser and the semantic and kernel modules to handle consts in the literal switch cases.

mccanne:
Closes #3003


nwt (approved):
Doh!

PR #3007, opened 2021-09-03
rename segment.Reference and index.Reference as object pattern

### LINKED ISSUE #2927: change segment name
After the config-journal branch is merged, we should change the name of the lake/segment package to conform with the name change to "data object".

The concept of "segment" didn't work very well after all.

The config-journal branch updates the lake design doc to use _data object_.

mccanne:
This commit changes terminology from the word "segment" that never really
stuck to the cloud-object terminology where we now have commit objects,
data objects, index objects, and seek objects.

This is just house keeping and name changes.  This commit contains
no changes to functionality.

Closes #2927


nwt (commented):
`git grep -i segment[^i]` (`[^i]` avoids matching segmentio) still shows lots of segment references.

lake/commits/snapshot.go
```diff
@@ -16,44 +16,47 @@ var (
...
+	Select(extent.Span, order.Which) DataObjects
+	SelectAll() DataObjects
 }
 
 type Writeable interface {
 	View
-	AddSegment(seg *segment.Reference) error
-	DeleteSegment(id ksuid.KSUID) error
+	AddDataObject(*data.Object) error
+	DeleteObject(id ksuid.KSUID) error
```
nwt, line 27: Maybe `DeleteDataObject` for symmetry with `AddDataObject`?

lake/index/writer_test.go
```diff
@@ -17,7 +17,7 @@ import (
 
 func TestWriter(t *testing.T) {
 	r := NewTypeRule("test", zng.TypeInt64)
-	ref := Reference{Rule: r, SegmentID: ksuid.New()}
+	ref := Object{Rule: r, ID: ksuid.New()}
 	w := testWriter(t, ref)
```
nwt, line 21:
```suggestion
	o := Object{Rule: r, ID: ksuid.New()}
	w := testWriter(t, o)
```

lake/index/writer_test.go
```diff
@@ -26,7 +26,7 @@ func TestWriter(t *testing.T) {
 
 func TestWriterWriteAfterClose(t *testing.T) {
 	r := NewTypeRule("test", zng.TypeInt64)
-	ref := Reference{Rule: r, SegmentID: ksuid.New()}
+	ref := Object{Rule: r, ID: ksuid.New()}
 	w := testWriter(t, ref)
```
nwt, line 30:
```suggestion
	o := Object{Rule: r, ID: ksuid.New()}
	w := testWriter(t, o)
```

lake/index/writer_test.go
```diff
@@ -38,7 +38,7 @@ func TestWriterWriteAfterClose(t *testing.T) {
 func TestWriterError(t *testing.T) {
 	const r1 = `{ts:1970-01-01T00:00:01Z,id:"id1"}`
 	const r2 = "{ts:1970-01-01T00:00:02Z,id:2}"
-	ref := Reference{Rule: NewFieldRule("test", "id"), SegmentID: ksuid.New()}
+	ref := Object{Rule: NewFieldRule("test", "id"), ID: ksuid.New()}
 	w := testWriter(t, ref)
```
nwt, line 42:
```suggestion
	o := Object{Rule: NewFieldRule("test", "id"), ID: ksuid.New()}
	w := testWriter(t, o)
```

nwt (approved):
One last "Data segments" at https://github.com/brimdata/zed/blob/2a010b16d0d099acb9600165a495fc04a346633f/cmd/zed/lake/vacate/command.go#L18.

lake/commits/snapshot.go
mccanne: Right now, DeleteObject only takes an id and works for both index and data objects.  I think this should change so that you can store actions in a pool and make the pool key(s) the span boundaries.  If you don't have the span in the delete you can do range queries to get the related Adds/Deletes.  When we make this change, we can rename DeleteObject and we will have DeleteDataObject and DeleteIndexObject and maybe some other DeleteFooObejcts.

Issue #2983, opened 2021-08-30
Spill directories not deleted on interrupt


philrz:
Repro is with Zed commit 520aca6 on macOS using a file from the 70 GB wrccdc data set.

Zed performs spill-to-disk on certain heavyweight operations when they can't be performed in memory alone, e.g., sorting, aggregations, and `fuse`. In the event of a catastrophic failure (e.g., `kill -9` or a sudden hard shutdown of the host) it seems understandable that the temporary storage associated with these operations could be left dangling. However, I've recently discovered that these droppings are left even during more polite forced exits of the Zed process, such as the SIGINT expected when hitting Control-C at the shell.

First, the happy path when it runs to completion:

```
$ zq -version
Version: v0.29.0-482-g520aca64

$ zq 'sort' dns.log.gz > /dev/null

$ echo $?
0

$ ls -l $TMPDIR | grep zed
[no output, since the directory was cleaned up automatically upon exit]
```

Here I hit Control-C about 90 seconds into the run.

```
$ zq 'sort' dns.log.gz > /dev/null
^Ccontext canceled

$ ls -l $TMPDIR | grep zed
drwx------   4 phil  staff    128 Aug 30 13:38 zed-spill-1587859615

$ ls -lR $TMPDIR/zed-spill-1587859615
total 2458584
-rw-r--r--  1 phil  staff  1090275665 Aug 30 13:38 0
-rw-r--r--  1 phil  staff   164128817 Aug 30 13:38 1

$ rm -rf $TMPDIR/zed-spill-1587859615
```

I suspect this becomes even more hazardous when Brim is in use, since if the user issues a new search while a prior one is in progress, this should cancel the query currently being executed by the Zed backend, and hence I expect the same droppings will be left behind. Since this is more of an implicit interrupt, I expect most users would be oblivious to the accumulation of temp space usage that may result. While I can confirm this symptom exists in the current Brim+Zed combo (Brim commit a1dfa38d talking to Zed commit 78ce875), while investigating that I found the more fundamental Brim issue brimdata/brim#1800 such that even an explicit "Kill search" is not currently having the intended effect right now, so we have a couple layers of issues to peel back here.

philrz:
Verified in Zed commit b01f993.

[...]

PR #3009, opened 2021-09-03
wait for proc cleanup after context cancelation

nwt:
Procs (e.g., sort) may need to clean up after context cancelation.  Give
them that opportunity by calling puller.Pull until it indicates
end-of-steam before returning from driver.run.

Partially addresses #2983.


mccanne (approved):
slick!

PR #3013, opened 2021-09-03
Add ability to revert a deleted object action

mattnibs:
Have a Snapshot keep track of deleted objects, then use this to look allow patch.Revert
to look up deleted objects.


mccanne (approved):
cool

Issue #3012, opened 2021-09-03
"zed lake delete" needs a ztest & help text adjustment


philrz:
While reviewing a revision of the Zed lake design doc as a user, I got the mistaken impression that `zed lake delete` could be used on commits, not just objects. @mccanne confirmed this is as designed, though.

> That behavior is now handled by `revert`.  `delete` is still useful (delete individual data objects instead of reverting a commit) and I meant to make a test for it so the issue should just be create a test for delete.

In addition to creating a test, it looks like the help text also needs to be adjusted, since as of Zed commit b2643c2 it still indicates can take "a list of commit tags".

```
$ zed -version
Version: v0.29.0-493-gb2643c22

$ zed lake delete -h
NAME
    zed lake delete - delete commits or data objects from a pool branch

USAGE
    delete -p pool[@commit] id [id ...]

OPTIONS
    -message commit message
    -p name of pool
    -q quiet mode (default "false")
    -user user name for commit message (default "phil@PhilMac.local")
    
    [lake flags]
    -R URI of path to Zed lake store (default "/Users/phil/logs")
    
    [zed flags]
    -cpuprofile write cpu profile to given file name
    -h display help (default "false")
    -hidden show hidden options (default "false")
    -memprofile write memory profile to given file name
    -version print version and exit (default "false")

DESCRIPTION
    "zed lake delete" takes a list of commit tags and/or data object tags
    in the specified pool branch and deletes all of the corresponding data
    objects. Once the delete operation completes, the deleted data is no
    longer seen when read data from the pool.

    No data is actually removed from the lake.  Instead, a delete operation
    is an action in the pool's commit journal.  Any delete can be "undone"
    by adding the commits back to the log using "zed lake add".
```

As for the Zed lake design doc, I've already got a branch in progress with some other fixes, so I'll propose the fix to that one.

mattnibs:
Completed with https://github.com/brimdata/zed/pull/3216

PR #3014, opened 2021-09-04
Fixes to Zed lake design doc after following as a user

philrz:
After the merge of #2994, I did another pass through the Zed lake design doc, reading it as a user would and following along with the example commands. I've proposed a set of corrections that I think are appropriate. I'll add some comments to draw attention to where I think additional input is justified.

### DIFF
--- a/docs/lake/design.md
+++ b/docs/lake/design.md
@@ -360,9 +360,13 @@ optimize layout for performance.
 These structures are introspected using meta-queries that simply
 specify a metadata source using an extended syntax in the `from` operator.
 There are three types of meta-queries:
-* `from :meta` - lake level  
-* `from pool:meta` - pool level  
-* `from pool@branch:meta` - branch level
+* `from :<meta>` - lake level
+* `from pool:<meta>` - pool level
+* `from pool@branch<:meta>` - branch level
+
+`<meta>` is the name of the metadata being queried. The available metadata
+sources varies based on level.
+
 For example, a list of pools with configuration data can be obtained
 in the ZSON format as follows:
 ```
@@ -381,20 +385,20 @@ zed lake query -Z "from logs:branches | branch.name=='main'"
 This meta-query produces a list of the data objects in the `live` branch
 of pool `logs`:
 ```
-zed lake query -Z "from logs@live:branches"
+zed lake query -Z "from logs@live:objects"
 ```
 
 You can also pretty-print in human-readable form most of the metadata Zed records
 using the "lake" format, e.g.,
 ```
-zed lake query -f lake "from logs@live:branches"
+zed lake query -f lake "from logs@live:objects"
 ```
 
 > TODO: we need to document all of the meta-data sources somewhere.
 
 #### Transactional Semantics
 
-The `commit` operation is _transactional_.  This means that a query scanning
+The "commit" operation is _transactional_.  This means that a query scanning
 a pool sees its entire data scan as a fixed "snapshot" with respect to the
 commit history.  In fact, the Zed language allows a commit object (created
 at any point in the past) to be specified with the `@` suffix to a
@@ -424,7 +428,7 @@ and each entry contains a timestamp, branch references can be easily
 navigated by time.  For example, a list of branches of a pool's past
 can be created by scanning the branches log and stopping at the largest
 timestamp less than or equal to the desired timestamp.  Likewise, a branch
-can be located in a similar fashion then its corresponding commit object
+can be located in a similar fashion, then its corresponding commit object
 can be used to construct that data of that branch at that past point in time.
 
 ### Merge Scan and Compaction
@@ -452,8 +456,7 @@ do not overlap.  This is just the basic LSM algorithm at work.
 
 To perform an LSM rollup, the `compact` command (implementation tracked
 via [zed/2977](https://github.com/brimdata/zed/issues/2977))
-is like a "squash" and
-To perform LSM-like compaction function, e.g.,
+is like a "squash" to perform LSM-like compaction function, e.g.,
 ```
 zed lake compact -p logs <tag>
 (merged commit <tag> printed to stdout)
@@ -470,8 +473,8 @@ meta-queries, an agent can introspect the layout of data, perform
 some computational geometry, and decide how and what to compact.
 The nature of this orchestration is highly workload dependent so we plan
 to develop a family of data-management orchestration agents optimized
-for various use cases (e.g., continuously ingested logs vs collections of
-metrics that should be optimized with columnar form vs slowly-changing
+for various use cases (e.g., continuously ingested logs vs. collections of
+metrics that should be optimized with columnar form vs. slowly-changing
 dimensional datasets like threat intel tables).
 
 An orchestration layer outside of the Zed lake is responsible for defining
--- a/docs/lake/design.md
+++ b/docs/lake/design.md
@@ -1051,7 +1051,7 @@ or a `ref`.
 * query: `from :meta`
 * rebase `ref` `commitish` - rebase a branch `ref` onto the commit object history starting at `commitish` (this is peculiar because `ref` could imply the pool name for `commitish` though this will usually be a branch name)
 * rename `pool` `pool` - change name of a data pool
-* revert `ref` `commitish`- undo the commit at `commitish` with a new commit object and updates the branch `ref` to point at the new commit; typically this commit would be in the branches history but it doesn't have to be (but would typically fail a consistency check if it isn't in the history).  We could prevent this condition with a check.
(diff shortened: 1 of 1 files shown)


docs/lake/design.md
```diff
@@ -91,8 +92,8 @@ design patterns of `git`.  In this approach,
 * a _commit_ operation is like a `git commit`,
 * and a pool _snapshot_ is like a `git checkout`.
 
-A core theme of the Zed lake design is _ergonomics_.  Given the git metaphor,
-our goal here is that the Zed lake tooling be as easy and familiar as git is
+A core theme of the Zed lake design is _ergonomics_.  Given the Git metaphor,
+our goal here is that the Zed lake tooling be as easy and familiar as Git is
```
philrz, line 96: I've seen @nwt propose corrections to capitalize "Git" when speaking of the tool, which I support, so I'm continuing in that spirit. I also support `git` being in lowercase & fixed width when we're referencing the command that's invoked at the shell, and it seems the doc is already quite consistent on that one.

docs/lake/design.md
```diff
@@ -221,8 +222,8 @@ zed lake log -p pool@branch
 ```
 To understand the log contents, the `load` operation is actually
 decomposed into two steps under the covers:
-an `add` step stores one or more
-new immutable data objects in the lake and a `commit` operation
+an "add" step stores one or more
+new immutable data objects in the lake and a "commit" operation
```
philrz, line 226: I'm normally a fan of invoking the fixed width font in backticks to add emphasis. However, since we used to have formal `zed lake` commands called `add` and `commit`, but they're gone, I figure _not_ putting them in fixed with helps contrast with the `load` command we just mentioned that _does_ remain.

docs/lake/design.md
```diff
@@ -358,11 +359,15 @@ lake making it easy to measure, tune, and adjust lake parameters to
 optimize layout for performance.
 
 These structures are introspected using meta-queries that simply
-specify a metadata source using a bracket syntax in the `from` operator.
+specify a metadata source using an extended syntax in the `from` operator.
```
philrz, line 362: I remember seeing the "bracket syntax" in some earlier revisions, but it looks like it's gone in favor of the colon/at syntax. My choice of the word "extended" was kind of arbitrary, so please propose a correction if you'd prefer something else!

docs/lake/design.md
```diff
@@ -424,7 +429,7 @@ and each entry contains a timestamp, branch references can be easily
 navigated by time.  For example, a list of branches of a pool's past
 can be created by scanning the branches log and stopping at the largest
 timestamp less than or equal to the desired timestamp.  Likewise, a branch
-can be located in a similar fashion then its corresponding commit object
+can be located in a similar fashion, then its corresponding commit object
```
philrz, line 432: I'm not sure if the comma is strictly necessary, but before it was there I paused on this sentence too long thinking "then" should be "than".

docs/lake/design.md
```diff
@@ -495,11 +499,11 @@ data during any key-range scan.
 ### Delete
 
 Data objects can be deleted with the `delete` command.  This command
-simply removes the data from branch without actually deleting the
+simply removes the data from the branch without actually deleting the
 underlying data objects thereby allowing time travel to work in the face
 of deletes.
 
-For example, this command deletes the three objects and/or commits referenced
+For example, this command deletes the three objects referenced
```
philrz, line 506: We discussed this change on Slack. The help text still speaks as if `delete` would accept commits, so I mentioned that in #3012 along with the request to add a test for it.

docs/lake/design.md
```diff
@@ -509,6 +513,20 @@ zed lake delete -p logs <id> <id> <id>
...
+branch.  Any data loaded in a reverted commit remains in the lake but no longer
+appears in the branch.  The new commit may recursively be reverted by an
+additional revert operation.
+
+For example, this command reverts the commit referenced by commit ID
+`<commit>`.
+```
+zed lake revert -p logs <commit>
+```
+
```
philrz, line 529: At the same time I was educated on Slack about `delete` only applying to objects, I was reminded of the existence of `revert`, at which point I noticed it's absent from the doc. I just copied this text pretty much verbatim from the `zed` help text. Feel free to propose more adjustments to suit the narrative nature of this doc.

docs/lake/design.md
```diff
@@ -588,7 +606,7 @@ in a pool, e.g.,
 ```
 zed lake index apply -p logs IndexGroupEx <tag>
 ```
-The index is created and a transaction put in staging.  Once this transaction
+The index is created and a transaction put (somewhere).  Once this transaction
```
philrz, line 609: This one needs genuine input! I just put some garbage text in there to trigger the PR suggestion workflow.

I know that "staging" as a formal construct is gone, but I don't know what took its place in this context. I know indexing in general is not operational (#2532), but I figure if we're going to make forward-looking statements about how it's expected to work, it might as well be accurate. :smile:

docs/lake/design.md
nwt, line 226: Feels odd to say there are two steps but then call one an operation.
```suggestion
new immutable data objects in the lake and a "commit" step
```

docs/lake/design.md
```diff
@@ -301,9 +302,9 @@ according to configured policies and logic.
 
 Data is read from one or more pools with the `query` command.  The pool/branch names
 are specified with `from` at the beginning of the Zed query along with an optional
-time range using `range` and `to`.  The default output format is ZNG though this
-can be overridden with `-f` to specify one of the various supported output
-formats.
+time range using `range` and `to`.  The default output format is ZSON (if to a
+terminal) or ZNG (if redirected elsewhere) though this can be overridden with
```
nwt, line 306:
```suggestion
time range using `range` and `to`.  The default output format is ZSON for
terminals and ZNG otherwise, though this can be overridden with
```

docs/lake/design.md
```diff
@@ -509,6 +513,20 @@ zed lake delete -p logs <id> <id> <id>
 > whatever reason, it should simply continue on and issue a warning on
 > the query endpoint "warnings channel".
 
+### Revert
+
+The actions in a commit can be reversed with the `revert` command.  This
+command applies the inverse steps in a new commit to the tip of the indicated
+branch.  Any data loaded in a reverted commit remains in the lake but no longer
+appears in the branch.  The new commit may recursively be reverted by an
```
nwt, line 521:
```suggestion
appears in the branch.  The new commit may itself be reverted by an
```

docs/lake/design.md
```diff
@@ -1031,9 +1049,9 @@ or a `ref`.
...
-* rebase `ref` `commitish` - rebase a branch `ref` onto the commit object history starting at `commitish` (this is peculiar because `ref` could implied the pool name for `commitish` though this will usually be a branch name)
+* rebase `ref` `commitish` - rebase a branch `ref` onto the commit object history starting at `commitish` (this is peculiar because `ref` could imply the pool name for `commitish` though this will usually be a branch name)
 * rename `pool` `pool` - change name of a data pool
-* undo `ref` `commitish`- undo reverts the commit at `commitish` with a new commit object and updates the branch `ref` to point at the new commit; typically this commit would be in the branches history but it doesn't have to be (but would typically fail a consistency check if it isn't in the history).  We could prevent this condition with a check.
+* revert `ref` `commitish`- undo the commit at `commitish` with a new commit object and updates the branch `ref` to point at the new commit; typically this commit would be in the branches history but it doesn't have to be (but would typically fail a consistency check if it isn't in the history).  We could prevent this condition with a check.
```
nwt, line 1054:
```suggestion
* revert `ref` `commitish`- undo the commit at `commitish` with a new commit object and update the branch `ref` to point at the new commit; typically this commit would be in the branch's history but it doesn't have to be (but would typically fail a consistency check if it isn't in the history).  We could prevent this condition with a check.
```

docs/lake/design.md
mccanne: Correct, this is described steps in the algorithm and isn't a command.

docs/lake/design.md
mccanne: The index is created and transactionally committed to the branch...

mccanne (approved):
Assuming you'll take Noah's suggestion and I offered text on the item you asked for...

docs/lake/design.md
```diff
@@ -57,8 +58,8 @@ ranges of data over the pool key.
 Data can be efficiently accessed via range scans composed of a
 range of values conforming to the pool key.
 
-A lake has a configured sort order, either ascending or descending
-and data is organized in the lake in accordance with this order.
+A pool key has a configured sort order, either ascending or descending
```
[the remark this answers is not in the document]
philrz: @nwt: Ok, but if that's how it's to be seen, the help text perhaps could use some improvement.

```
$ zed lake create -h
...
OPTIONS
    -orderby comma-separated pool keys with optional :asc or :desc suffix to organize data in pool (cannot be changed) (default "ts:desc")
```

To me that reads as if the "pool key" is the noun attached to `-orderby` and the suffix is further describing said pool key. But I'm happy to merge your suggested wording in the design doc.

PR #3021, opened 2021-09-06
Zeek shaper updates for v4.1.0

philrz:
For some time now, each time a new Zeek version is released, I use our [print-types.zeek](https://github.com/brimdata/zeek/blob/master/brim/print-types.zeek) script to see if there's new log types in the default output and/or new/changed fields in the existing log types. I then update the reference Zed shaper appropriately. Since this is data with which we've got a lot of prior experience, I feel like it's good shaping practice.

When I went to do this for the recently-released Zeek v4.1.0, for the first time since maintaining the reference shaper, I actually saw field names _change_ from what they were prior. This poses a unique challenge because when fields were only being added, the only side effect of applying the newest shaper on older logs was that there'd be some extra fields filled in with `null`. In fact, I've always thought this has an up side, since it "future proofs" the user for when they _do_ upgrade to the newer Zeek release, since their shaped Zeek logs would continue to fit within the type definition they're already been using. But to cover fields that change, it seemed like I'd either have to:

1. Start maintaining separate version-specific reference shapers. Specifically, that would now include one for Zeek v4.0.3 and earlier and another for Zeek v4.1.0 and newer. I'd maintain just those two until yet another set of field name changes comes along.

2. Continue to maintain a single shaper, but introduce flow control logic to detect the changes that are unique to Zeek v4.1.0 and newer and apply them only when necessary.

In the interest of maintaining minimal redundant config and showing an approach that might be applicable in other contexts, in this PR I'm opting for the latter approach.

The diffs on which the changes are based is in the attached [types-diff.txt](https://github.com/brimdata/zed/files/7117335/types-diff.txt).

### DIFF
--- a/zeek/Shaping-Zeek-NDJSON.md
+++ b/zeek/Shaping-Zeek-NDJSON.md
@@ -1,13 +1,14 @@
 # Shaping Zeek NDJSON
 
 - [Summary](#summary)
+- [Zeek Version/Configuration](#zeek-versionconfiguration)
 - [Reference Shaper Contents](#reference-shaper-contents)
   * [Leading Type Definitions](#leading-type-definitions)
-  * [Type Definitions Per Zeek Log `_path`](#type-definitions-per-zeek-log-_path)
+  * [Default Type Definitions Per Zeek Log `_path`](#default-type-definitions-per-zeek-log-_path)
+  * [Version-Specific Type Definitions](#version-specific-type-definitions)
   * [Mapping From `_path` Values to Types](#mapping-from-_path-values-to-types)
   * [Zed Pipeline](#zed-pipeline)
 - [Invoking the Shaper From `zq`](#invoking-the-shaper-from-zq)
-- [Zeek Version/Configuration](#zeek-versionconfiguration)
 - [Importing Shaped Data Into Brim](#importing-shaped-data-into-brim)
 - [Contact us!](#contact-us)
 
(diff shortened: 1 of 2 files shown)


zeek/Shaping-Zeek-NDJSON.md
```diff
@@ -30,6 +31,37 @@ A full description of all that's possible with shapers is beyond the scope of
...
+# Zeek Version/Configuration
+
+The fields and data types in the reference `shaper.zed` reflect the default
+NDJSON-format logs output by Zeek releases up to the version number referenced
+in the comments at the top of that file. They have been revisited periodically
+as new Zeek versions have been released.
+
+Most changes we've observed in Zeek logs between versions have involved only the
+addition of new fields. Because of this, we expect the shaper should be usable
+as-is for Zeek releases older than the one most recently tested, since fields
```
nwt, line 43:
```suggestion
as is for Zeek releases older than the one most recently tested, since fields
```

zeek/shaper.zed
```diff
@@ -1,11 +1,15 @@
 // This reference Zed shaper for Zeek NDJSON logs was most recently tested with
-// Zeek v4.0.1. The fields and data types reflect the default NDJSON
+// Zeek v4.1.0. The fields and data types reflect the default NDJSON
 // logs output by that Zeek version when using the JSON Streaming Logs package.
 
 type port=uint16;
 type zenum=string;
 type conn_id={orig_h:ip,orig_p:port,resp_h:ip,resp_p:port};
 
+// This first block of type definitions cover the fields we've observed in
```
nwt, line 9:
```suggestion
// This first block of type definitions covers the fields we've observed in
```

nwt (approved):

PR #3023, opened 2021-09-06
add back -p for checkout

mccanne:
This commit adds back the pool option -p but only for checkout,
which checks out the main branch of the indicated pool.
This will help with the gentle slope for new users who don't need
to initially undestand or even know about branches.

We also updated various CLI help to reflect that the -p argument
is no longer relevant to them.

### DIFF
--- a/cmd/zed/lake/checkout/command.go
+++ b/cmd/zed/lake/checkout/command.go
@@ -14,7 +14,7 @@ import (
 
 var Checkout = &charm.Spec{
 	Name:  "checkout",
-	Usage: "checkout [-b] branch",
+	Usage: "checkout [-p pool] [-b] branch [base]",
 	Short: "checkout a branch",
 	Long: `
 The lake checkout command sets the working branch as indicated.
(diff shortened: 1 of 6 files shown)


philrz:
Is it possible to mention the pool in the output to `checkout`?

```
$ zed lake checkout -p logs
Switched to branch "main"
```

e.g., \``Switched to branch "main" on pool "logs"`

I feel like _mentioning_ branches in this output is a reasonable way to tug them along on the gentle slop. But once again thinking of that branch-oblivious user, if they only see mention of branches in response to a command that mentions only a pool, that seems to have the potential to confuse.

cmd/zed/lake/checkout/command.go
```diff
@@ -23,6 +23,9 @@ having to specify the working branch.  The branch specifier may also be
 a commit ID, in which case you entered a headless state and commands
 like load that require a branch name for HEAD will report an error.
 
+Checkout may also be run with the -p to indicate a pool name.  In this case,
```
nwt, line 26:
```suggestion
Checkout may also be run with -p to indicate a pool name.  In this case,
```

nwt (approved):

PR #3024, opened 2021-09-06
change behavior of lake checkout state

mccanne:
This commit changes the default behavior of the checkout command
to store the HEAD pointer in $HOME.  The environment variable
ZED_HEAD_PATH can be defined to override this behavior.  In particular,
setting ZED_HEAD_PATH to ".", will revert to the previous behavior
where the HEAD pointer is stored in the current working directory.

### DIFF
--- a/cli/lakeflags/state.go
+++ b/cli/lakeflags/state.go
@@ -3,13 +3,25 @@ package lakeflags
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 )
 
-const headFile = ".zed_head"
+const headFile = ".ZED_HEAD"
+
+func headPath() string {
+	path := os.Getenv("ZED_HEAD_PATH")
+	if path == "" {
+		path = os.Getenv("HOME")
+	}
+	if path == "." {
+		path = ""
+	}
+	return filepath.Join(path, headFile)
+}
 
 func readHead() (string, error) {
-	b, err := os.ReadFile(headFile)
+	b, err := os.ReadFile(headPath())
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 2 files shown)


philrz:
As a user, I like this!

cli/lakeflags/state.go
```diff
@@ -3,13 +3,25 @@ package lakeflags
...
 	"strings"
 )
 
-const headFile = ".zed_head"
+const headFile = ".ZED_HEAD"
+
+func headPath() string {
+	path := os.Getenv("ZED_HEAD_PATH")
+	if path == "" {
+		path = os.Getenv("HOME")
```
nwt, line 15: Use `os.UserHomeDir`.

cli/lakeflags/state.go
```diff
@@ -3,13 +3,25 @@ package lakeflags
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 )
 
-const headFile = ".zed_head"
+const headFile = ".ZED_HEAD"
```
nwt, line 10: I much prefer `.zed_head` to `.ZED_HEAD`. There's no need to shout.

philrz: Unrelated: There will one day be "Zed Head" t-shirts.

philrz: ...and perhaps even more appropriate, hats!

cli/lakeflags/state.go
```diff
@@ -3,13 +3,25 @@ package lakeflags
...
 
+func headPath() string {
+	path := os.Getenv("ZED_HEAD_PATH")
+	if path == "" {
+		path, _ = os.UserHomeDir()
+	}
+	if path == "." {
+		path = ""
+	}
+	return filepath.Join(path, headFile)
```
nwt, line 20: s/path/dir/i (including ZED_HEAD_PATH) to clarify that this is a directory.

nwt (approved):

PR #3031, opened 2021-09-07
change "checkout" to "use" and -HEAD to -use

mccanne:
Closes #3025

PR #3035, opened 2021-09-07
Zed lake README fixes post-zed/3031

philrz:
After the merge of #3031, I did another run through the Zed lake README doc and have a few more adjustments to propose.

### DIFF
--- a/docs/lake/README.md
+++ b/docs/lake/README.md
@@ -172,7 +172,7 @@ The `-use` flag appears in most of the lake commands and indicates
 a commit-history pointer just for that one command invocation.
 
 Supposing the `main` branch of `logs` was already set as the default
-with the "use" command (see below),
+with the `use` command (see below),
 then you could create the new branch called "staging" by simply saying
 ```
 zed lake branch staging
@@ -189,10 +189,20 @@ zed lake branch
 ### Use
 
 The `use` command provides a means to set the current branch context
-to the indicate branch, e.g.,
+to the indicated branch, e.g.,
 ```
 zed lake use staging
 ```
+As previously noted, Zed lakes can be used without thinking about branches.
+The `use` command includes a `-p` option that can specify a pool name, which
+sets the current context to the `main` branch of that pool. A user that's not
+yet ready to fully embrace branches could therefore have locked in this
+context by executing the following command after running [`create`](#create)
+to initiate our new example pool.
+```
+zed lake use -p logs
+```
+
 When you want to specify a branch in another pool, you simply prepend
 the pool name to the branch:
 ```
@@ -226,7 +236,6 @@ zed lake load -i parquet sample4.parquet
 zed lake load -i zst sample5.zst
 ```
 
-By default, the data is committed into the `main` branch of the pool.
 An alternative branch may be specified using the `@` separator,
 i.e., `<pool>@<branch>`.  Supposing there was a branch called `updates`,
 data can be committed into this branch as follows:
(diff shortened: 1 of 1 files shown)


docs/lake/README.md
```diff
@@ -189,10 +189,20 @@ zed lake branch
...
+As previously noted, Zed lakes can be used without thinking about branches.
+The `use` command includes a `-p` option that can specify a pool name, which
+sets the current context to the `main` branch of that pool. A user that's not
+yet ready to fully embrace branches could therefore have locked in this
+context by executing the following command after running [`create`](#create)
+to initiate our new example pool.
+```
+zed lake use -p logs
+```
+
```
philrz, line 205: Hopefully it's not too controversial that I mention this here. Part of why I advocated having a branch-oblivious option is because we'd already revealed Zed lakes to many beta users before branches existed and they seemed satisfied with the functionality thus far. Maybe 100% of them will be Git-aware and on-board with the role of branches as soon as they see the expanded doc. However, for those users that are still climbing the gentle slope, I'd like to have this option revealed somewhere obvious so they know they can effectively still do all their work in a pool without having to think about branches until they're ready.

docs/lake/README.md
```diff
@@ -374,8 +383,8 @@ These range queries are efficiently implemented as the data is laid out
 according to the pool key and seek indexes keyed by the pool key
 are computed for each data object.
 
-Lake queries also can refer to HEAD either implicitly by omitting
-the `from` operator:
+Lake queries also can refer to HEAD (i.e., the branch context set in the most
+recent `use` command) either implicitly by omitting the `from` operator:
```
philrz, line 387: Is this an accurate thing to say? As I was going through the current README I noticed that while `-HEAD` is gone as an option, there's still plenty of "HEAD" references remaining in the doc. The first time the word "HEAD" appears in the doc (in the "Use" section) it's mentioned without any prior context. Once again, users accustomed to the Git concepts will probably understand what this is saying, but then I get concerned that other users might feel alienated. Maybe a separate section about the role of "HEAD" would be justified in the doc. For now, I'm hoping what I'm written here can stand on its own, since it's how I happen to look at it as a user: Once I've done `zed lake use`, I'm free to run all my `query` and `load` commands and know that was my context without having to specify `from` in my queries or reference a pool in my `load`.

docs/lake/README.md
```diff
@@ -226,7 +236,6 @@ zed lake load -i parquet sample4.parquet
 zed lake load -i zst sample5.zst
 ```
 
-By default, the data is committed into the `main` branch of the pool.
```
philrz, line 229: I deleted this line because it doesn't seem accurate anymore. Yes, if someone had done only `zed lake use -p pool` it's true that data would by default be committed into the `main` branch of that pool. But in the absence of _any_ `zed lake use`, the `load` would simply fail.

```
$ zed lake init
using environment variable ZED_LAKE_ROOT
lake created: /Users/phil/logs

$ zed lake create logs
pool created: logs 1xpaHDcvgYOOWBWiJTwPxbZTWLP

$ zed lake load dpd.log.gz 
HEAD not specified: indicate with -use or run the "use" command
```

There's text a little further down in the same section that already says:

> Or, as mentioned above, you can set the default branch for the load command via `use`: ...

Therefore I think we're covered without the line I'm deleting here.

docs/lake/README.md
mccanne: Yes looks good.  HEAD is a concept that shows up in the log listing and is how git refers to this concept, but we can revisit it.  It's really useful when you run zed lake log to see where HEAD is at relative to other branch references.

mccanne (approved):

docs/lake/README.md
```diff
@@ -189,10 +189,20 @@ zed lake branch
...
 
 The `use` command provides a means to set the current branch context
-to the indicate branch, e.g.,
+to the indicated branch, e.g.,
 ```
 zed lake use staging
 ```
+As previously noted, Zed lakes can be used without thinking about branches.
+The `use` command includes a `-p` option that can specify a pool name, which
+sets the current context to the `main` branch of that pool. A user that's not
```
nwt, line 198:
```suggestion
To this end, the `use` command includes a `-p` option that sets the current
branch context to the indicated pool's `main` branch. A user that's not
```

PR #3039, opened 2021-09-08
change "zed lake use" with no args to print HEAD

mccanne:
This commit changes the "use" command to print the current HEAD
if there are no command-line args, which is useful for a quick
reminder of the branch you're on.  We also reworked the logic
a bit for better readability.

### DIFF
--- a/cmd/zed/lake/use/command.go
+++ b/cmd/zed/lake/use/command.go
@@ -83,11 +83,10 @@ func (c *Command) Run(args []string) error {
 	case 0:
 		if c.poolName == "" {
 			head, err := c.lakeFlags.HEAD()
-			if err == nil {
+			if err == nil && !head.IsZero() {
 				fmt.Printf("HEAD at %s\n", head)
-				return nil
 			}
-			return errors.New("a branch name or commit ID must be given")
+			return errors.New("no branch HEAD set: pool name and branch name or commit ID must be given")
 		}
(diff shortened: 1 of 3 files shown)


philrz (commented):
One corner case I spotted: If I've not yet done _any_ `use` (hence no `.zed_head`), this is what I see:

```
$ zed lake use
HEAD at @
```

Could we catch this corner case and provide an error message that advises them to specify `pool@branch` or `-p pool`?

cmd/zed/lake/use/command.go
```diff
@@ -78,34 +78,39 @@ func (c *Command) Run(args []string) error {
...
+	var poolName, branchName, baseName string
 	switch len(args) {
 	case 0:
 		if c.poolName == "" {
+			head, err := c.lakeFlags.HEAD()
+			if err == nil {
+				fmt.Printf("HEAD at %s\n", head)
+				return nil
+			}
 			return errors.New("a branch name or commit ID must be given")
```
nwt, line 90: Nit: Indent the error path.
```suggestion
			if err != nil {
				return errors.New("a branch name or commit ID must be given")
			}
			fmt.Printf("HEAD at %s\n", head)
			return nil
```

nwt (approved):
Add some tests for this in `zed/lake/ztests/`.

PR #3040, opened 2021-09-08
allow any char after backslash when parsing Zed regexps

### LINKED ISSUE #3037: Can't express literal "." in regexp
Repro is with Zed commit d10df29. This issue was originally found by a community user.

Consider this simplified example data and regexp-assisted search.

```
$ cat foo.ndjson 
{"foo": "this.that"}
{"foo": "thisNthat"}

$ zq -z '/this.that/' foo.ndjson 
{foo:"this.that"}
{foo:"thisNthat"}
```

The `.` is playing the common role in regexp of matching any character. So far so good. But now we can't see a way to force it to only match the record that contains the actual dot while still using a regexp. Backslash escape is what normally comes to mind, but that's currently a parser error.

```
$ zq -z '/this\.that/' foo.ndjson 
zq: error parsing Zed at column 7:
/this\.that/
  === ^ ===
```

nwt:
The parser allows only one backslash escape sequence in regular
expressions: "\/".  This is too restrictive.  Instead, allow a backslash
followed by any character.  (This means the parser now allows invalid
escape sequences, but the expr packages will reject them when it tries
to compile them.)

Closes #3037.

### DIFF
--- a/compiler/parser/parser.es.js
+++ b/compiler/parser/parser.es.js
@@ -1140,28 +1140,26 @@ function peg$parse(input, options) {
           },
       peg$c514 = /^[^\/\\]/,
       peg$c515 = peg$classExpectation(["/", "\\"], true, false),
-      peg$c516 = "\\/",
-      peg$c517 = peg$literalExpectation("\\/", false),
-      peg$c518 = /^[\0-\x1F\\]/,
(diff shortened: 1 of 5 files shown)


mccanne (approved):
nice!

PR #3043, opened 2021-09-09
put: Require explicit assignments

### LINKED ISSUE #2997: Revert the <call-expression> shortcut for "put"
When `put` was made optional in #2967, a shorthand that came along with this is the ability to leave off the `fieldname :=` as long as the right-hand-side expression is the name of a function, since this could create a field of the same name as the function. E.g.:

```
$ zq -version
Version: v0.29.0-488-gd30daf4a

$ echo '{"foo": "BAR"}' | zq 'to_lower(foo)' -
{foo:"BAR",to_lower:"bar"}
```

When I wrote the related docs updates in #2986, I found it a little awkward to include this shortcut in the summary of the new syntax. The best idea we had was one from @mattnibs to describe the function example as a `<call-expression>`:

```
[put] (<field> := <expression>) | <call-expression> [, (<field> := <expression>) | <call-expression> ...]
```

The docs and the code are now in sync, but this has given us reason to pause as a team and reflect on whether we like this as users. The current consensus is now that it feels _weird_.

We originally rationalized doing it on the grounds that aggregations behave like this, e.g., a Zed query consisting of just `count()` creates a field called `count` to hold the result. I've claimed this feels a little apples & oranges: When doing an aggregation, you're effectively replacing your original data (which might have dozens of fields) with new summarized data (which might be as narrow as a single field holding a single value, as in the `count()` example). Meanwhile, adding a new field called `to_lower` to what's potentially a wide sea of existing fields seems strange. I expect most users would be inclined to use `:=` to assign a more meaningful field name in that case. And if it's true that users may rarely invoke this shortcut, it calls into question what we're getting for the combined cost of:

1.  The parser being slightly more complicated & expensive to support
2.  The docs being more complicated to write/extend
3.  The docs being more complicated for the user to read/retain

In conclusion, we've reached consensus on backing out this shortcut.

mattnibs:
Make both implicit and explicit put procs require that every expression
have an explicit assignment.

Closes #2997

### DIFF
--- a/compiler/ast/ast.go
+++ b/compiler/ast/ast.go
@@ -292,10 +292,10 @@ type (
 		Args []Assignment `json:"args"`
 	}
 
-	// An OpAssignments proc is a list of assignments whose parent proc
+	// An OpAssignment proc is a list of assignments whose parent proc
 	// is unknown: It could be a Summarize or Put proc. This will be
 	// determined in the semantic phase.
-	OpAssignments struct {
+	OpAssignment struct {
 		Kind        string       `json:"kind" unpack:""`
 		Assignments []Assignment `json:"assignments"`
 	}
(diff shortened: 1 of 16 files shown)


mattnibs:
@philrz I did my best to do the doc changes in this pr as well.

docs/language/operators/README.md
```diff
@@ -711,13 +711,12 @@ id.orig_h     id.orig_p id.resp_h       id.resp_p orig_bytes resp_bytes total_by
...
-`put` keyword or calling a function without specifying a field name in which
-to store the value it returns. Using both of these shortcuts, here we create
-a field to hold the lowercase representation of another field value.
+As noted above the `put` keyword is entirely optional. Here we omit
+the `put` keyword and create a new field to hold the lowercase
+representation of another field value.
 
 ```mdtest-command zed-sample-data/zeek-default
-zq -f table 'cut method | to_lower(method)' http.log.gz
+zq -f table 'cut method | to_lower:=to_lower(method)' http.log.gz
```
philrz, line 719:
```suggestion
zq -f table 'cut method | lower_method:=to_lower(method)' http.log.gz
```

It's a minor thing, but I figure we might as well use a field name different from the function name. If we thought the function names made for good field names we'd probably have had less consensus on proceeding with this PR. :smile: 

Note that it'll require adjusting the field header in the expected output. GitHub won't let me do that via this Suggestion feature since it's too far from the lines being changed. :roll\_eyes:

philrz (commented):
I made one small suggestion on the docs side, but overall LGTM. 👍

compiler/ast/ast.go
```diff
@@ -292,12 +292,12 @@ type (
...
-	// An OpExprs proc is a list of assignments whose parent proc
-	// is unknown: It could be a Filter, Summarize, or Put proc. This
-	// will be determined in the semantic phase.
-	OpExprs struct {
-		Kind  string `json:"kind" unpack:""`
-		Exprs []Expr `json:"exprs"`
+	// An OpAssignments proc is a list of assignments whose parent proc
+	// is unknown: It could be a Summarize or Put proc. This will be
+	// determined in the semantic phase.
+	OpAssignments struct {
```
nwt, line 298: I think we should call this `OpAssignment` so it's singular like all the other AST nodes.

compiler/ztests/filter-non-boolean-call.yaml
```diff
@@ -0,0 +1,13 @@
...
+inputs: 
+  - name: stdin
+    data: '{}'
+
+outputs:
+  - name: stderr
+    data: |
+      bad expression in filter: function "to_lower" does not return a boolean value
+
+
```
nwt, line 13:
```suggestion
```

docs/language/operators/README.md
```diff
@@ -711,18 +711,17 @@ id.orig_h     id.orig_p id.resp_h       id.resp_p orig_bytes resp_bytes total_by
 
 #### Example #2:
 
-As noted above, shorthand syntax may be used, such as leaving out the optional
-`put` keyword or calling a function without specifying a field name in which
-to store the value it returns. Using both of these shortcuts, here we create
-a field to hold the lowercase representation of another field value.
+As noted above the `put` keyword is entirely optional. Here we omit
+the `put` keyword and create a new field to hold the lowercase
```
nwt, line 715:
```suggestion
As noted above the `put` keyword is entirely optional. Here we omit
it and create a new field to hold the lowercase
```

nwt (approved):

PR #3054, opened 2021-09-13
properly format empty string field name

mccanne:
The JSON spec allows empty string to be an object key and
it was ZSON's intention to do the same, but the ZSON formatter
was incorrectly printing out the empty field name without quotes.
This commit fixes this bug and adds a test.

### DIFF
--- a/zng/name.go
+++ b/zng/name.go
@@ -14,6 +14,9 @@ func TypeChar(c rune) bool {
 }
 
 func IsIdentifier(s string) bool {
+	if len(s) == 0 {
+		return false
+	}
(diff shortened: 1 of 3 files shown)


zng/name.go
```diff
@@ -14,6 +14,9 @@ func TypeChar(c rune) bool {
 }
 
 func IsIdentifier(s string) bool {
+	if len(s) == 0 {
```
nwt, line 17:
```suggestion
	if s == "" {
```

nwt (approved):

PR #3064, opened 2021-09-14
handle record aliases in drop

nwt:


mattnibs (approved):
Looks like there's a few more tests to fix.

proc/ztests/cut-empty-record.yaml
```diff
@@ -6,4 +6,5 @@ input: |
   {a:{},b:{}}
 
 output: |
+  {a:{}}
```
mattnibs, line 9: Two bug fixes for the price of one!

mattnibs (approved):

PR #3065, opened 2021-09-15
remove POST /ast and POST /search

nwt:
Both can go now that the Brim app uses POST /query.


mattnibs (approved):
I would recommend run the brim tests against this before merge just to make sure this doesn't break anything.

nwt:
I got a clean run of all the Brim tests.

PR #3079, opened 2021-09-16
index update command

### LINKED ISSUE #3076: index update command
A user friendly version of index apply, index update will ensure that every object in a branch is indexed with the current set of index rules.

mattnibs:
Add index update command which ensures that there is an index for each
member of the current index rule set for every object at a commit point.

Closes #3076

### DIFF
--- a/cmd/zed/lake/index/update.go
+++ b/cmd/zed/lake/index/update.go
@@ -1,0 +1,63 @@
+package index
+
+import (
+	"flag"
+	"fmt"
+
+	"github.com/brimdata/zed/cli/lakeflags"
+	zedlake "github.com/brimdata/zed/cmd/zed/lake"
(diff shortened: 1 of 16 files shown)


cmd/zed/lake/index/update.go
```diff
@@ -0,0 +1,63 @@
...
+	Short: "index all object in a branch using the current set of rules",
+	New:   NewUpdate,
+}
+
+type UpdateCommand struct {
+	*Command
+	zedlake.CommitFlags
+}
+
+func NewUpdate(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
```
nwt, line 25: Nit: No need to export these.

nwt (approved):

PR #3080, opened 2021-09-16
add Zed url_parse function

### LINKED ISSUE #2989: Function to parse URIs
Repro is with Zed commit 59dc184.

A community user recently asked:

> Is there any way to manipulate the data within a field?  For instance, I was looking to break apart the different parameters that are in a URI. `/test/?first=sometext&second=moretext` turning that into `first=sometext` and `second=moretext`

Regexp group support (#2093) has the potential to help here when we implement it. For now, I pointed out the existence of the `split()` function and slice operator that can accomplish much of this. For example:

```
$ zq -version
Version: v0.29.0-487-g59dc1843

$ echo '{"uri": "/test/?first=sometext&second=moretext"}' | zq -Z 'put qstring:=split(uri,"?")[1] | put pairs:=split(qstring,"&")' -
{
    uri: "/test/?first=sometext&second=moretext",
    qstring: "first=sometext&second=moretext",
    pairs: [
        "first=sometext",
        "second=moretext"
    ]
}
```

No doubt this could be taken even further by using the sequence operator to break out the entries in the `pairs` array into separate fields.

However, having shown this, I was also aware that there's probably escaping or other complexity that could make this approach break with some URIs. That made me search & find Go libraries like [https://pkg.go.dev/net/url#example-URL.Query](https://pkg.go.dev/net/url#example-URL.Query) that purport to do this heavy lifting for us. Since URI manipulation is something that may come up frequently for our users, it might be beneficial for Zed to provide a URI-specific parsing function that returns a nested record containing the fully-parsed data.

nwt:
Closes #2989.

### DIFF
--- a/expr/function/function.go
+++ b/expr/function/function.go
@@ -3,6 +3,8 @@ package function
 import (
 	"errors"
 	"fmt"
+	"net/url"
+	"strconv"
 	"strings"
 
 	"github.com/brimdata/zed/anymath"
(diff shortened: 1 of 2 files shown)


mattnibs (approved):
nooice!

PR #3090, opened 2021-09-17
remove stray fmt.Println from ztest.RunShell

nwt:


ztest/shell.go
```diff
@@ -33,6 +33,5 @@ func RunShell(dir Dir, bindir, script string, stdin io.Reader, useenvs []string)
 	cmd.Stdout = &stdout
 	cmd.Stderr = &stderr
 	err := cmd.Run()
-	fmt.Println("stderr", stderr.String())
```
mattnibs, line 36: That would be me. My bad.

mattnibs (approved):

ztest/shell.go
nwt: These things happen!

PR #3095, opened 2021-09-19
change "zqd" to "zed" in python client

mccanne:
This commit gets rid of the "zqd" reference in the python client.
This will break some old examples in online content but we
will have many forthcoming python examples in the near
future.  We also clarified the example in the python
README so newcomers who know python and just want to
quickly try out Zed should have an easier time.


nwt (approved):
`zed.py` mentions zqd twice. Update it?

Issue #1312, opened 2020-09-18
Implement Zed type for float16


mccanne:

philrz:
@mccanne: Should we expand the scope of this to include `float16` also? I see that's also in the updated list of [ZNG Primitive Types](https://github.com/brimsec/zq/blob/master/zng/docs/spec.md#5-primitive-types) but it appears unavailable to me as a user in `zq`, similar to `float32`. For instance, as of zq commit `0c6b642`:

```
$ cat input.tzng 
#0:record[foo:uint64]
0:[12;]

$ zq -t "put f=foo:float64" input.tzng 
#0:record[foo:uint64,f:float64]
0:[12;12;]

$ zq -t "put f=foo:float32" input.tzng 
parse error: 1:11 (10): no match found, expected: "bool", "bstring", "bytes", "duration", "error", "float64", "int16", "int32", "int64", "int8", "ip", "net", "null", "string", "time", "type", "uint16", "uint32", "uint64" or "uint8"

$ zq -t "put f=foo:float16" input.tzng 
parse error: 1:11 (10): no match found, expected: "bool", "bstring", "bytes", "duration", "error", "float64", "int16", "int32", "int64", "int8", "ip", "net", "null", "string", "time", "type", "uint16", "uint32", "uint64" or "uint8"
```

nwt:
#3110 implements float32 but float16 remains unimplemented.

PR #3110, opened 2021-09-24
implement Zed float32 primitive type

nwt:
Half of #1312.

~~Needs more tests.~~


mccanne (approved):
Looks great.  I was thinking we should add test coverage to the marshaler for float32 (and float64 since it's not there either).

PR #3117, opened 2021-09-28
Remove no longer relevant zql directory

mattnibs:


philrz (commented):
I might caution against merging this PR. I put this file here intentionally because the **Help > ZQL Syntax Docs** pull-down menu in the Brim `v0.24.0` release is tied to a URL for this directory. I'd love to be able to say that all users are surely running `v0.25.0` now and hence this is no longer an issue, but that's not the case. First there's the fact that the new installer made it such that Windows `v0.24.0` installs do not get auto-updated nor a notice that a new version is available, so I expect there will be a slow transition of Windows users over time as they find out about the newer release via other channels. Second, while the bulk of new downloads are of `v0.25.0`, we still seem to get a steady trickle of `v0.24.0` downloads that you can see on our internal #downloads channel, e.g, five downloads of the Windows one last night, and there have been sporadic downloads of the Linux ones here and there in recent days. I suspect this may be due to places on the Internet that link directly to the download URL for `v0.24.0`. I did some sleuthing recently by searching for that download URL and one of the hits that stood out was https://wingetgui.com/apps?id=brimdata.brim&v=0.24.0. I contacted someone at that web site a few days ago to advise them the URL should be updated, but, no response yet.

Anyway, I respect the desire to keep the repo tidy and I know it's probably not feasible to whack-a-mole the many places on the Internet that seem to insist on rolling their own approaches to redistributing or pointing at the software, so at some point I know it's fair game to cut ties with the past and maybe let broken links be the way that users learn to upgrade. I just wanted to make sure it's known that some of this backward-looking stuff was laid out this way very intentionally in hopes of allowing for a grace period (see also: the `v0.24.0`-specific articles on the Brim wiki, which surely need to go at some point, but I'd similarly advise to keep them up for a bit if they're not hurting anybody.)

PR #3122, opened 2021-09-29
move contents of package zng/resolver to top level

nwt:
Part of #2824.


mapper.go
```diff
@@ -1,4 +1,4 @@
-package resolver
```
mattnibs, line 1: Does it make more sense to just put this into zson next to Context?

nwt: No, `zson.Context` will go away after `zng.Context` moves to the top level.

https://github.com/brimdata/zed/blob/1af7bb18c0d363c660df71cc28bb89d7524378b8/zson/context.go#L13-L18)

mattnibs (approved):

PR #3126, opened 2021-09-29
remove driver.Copy

nwt:


mccanne:
is there a better way to do this?

https://github.com/brimdata/zinger/blob/1cb3dde45b1d1c4f7bda0f9784ae3c2842280927/fifo/lake.go#L124

nwt:
Nope. Thanks for pointing out that call.

PR #3138, opened 2021-10-01
use seekindex containing nulls

### LINKED ISSUE #2965: nulls in seekindex
@aousterh stumbled across this when using the id.orig_h as the pool key...

When the seekindex has nulls from records without pool keys or records with a pool key that has a null value, it punts.   The if-statement that rejects this case should be removed and the code that computes intersections over keyspans should be updated to handle nullsmax.

https://github.com/brimdata/zed/blob/7359b7de0f2989e809ba82b043dbb0586ef2dcb9/lake/segment/reader.go#L40

Replicate with:
```
zed lake create -p bug -orderby id.orig_h
zed lake load -p bug zq-sample-data/zeek-default/*.gz
```
Then you will see meta.last is null:
```
zed lake query -Z "from bug[objects]"
```
And this query will run slowly because the seek index isn't used:
```
zed lake query -z "from ip2 range 10.128.0.241 to 10.128.0.241 | id.orig_h == 10.128.0.241"
```

nwt:
Closes #2965.

### DIFF
--- a/lake/data/reader.go
+++ b/lake/data/reader.go
@@ -31,20 +31,15 @@ func (o *Object) NewReader(ctx context.Context, engine storage.Engine, path *sto
 	if err != nil {
 		return nil, err
 	}
+	span := extent.NewGeneric(o.First, o.Last, cmp)
+	if !span.Crop(scanRange) {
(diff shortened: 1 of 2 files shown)


mccanne (approved):
cool

PR #3139, opened 2021-10-01
remove dead code from expr/function/strings.go

nwt:


mccanne (approved):
this reminds me we should have a sprintf function

PR #3158, opened 2021-10-07
Use lake indexes to speed up queries

### LINKED ISSUE #3073: lake: Use field indices to speed up searches
We currently have lake index v0, this is lake index v1. Basically wire up lake index functionality to be more of a first class citizen with the final goal of utilizing indices for field lookups in zed queries.

mattnibs:
Similarly to what is done with the filter push down, grab simple
predicates in the immediate filter after the "from" and match these
with any created object indexes along query path.

For now this only supports single Field rules, but will eventually
expand to the other rule types.

Closes #3073

### DIFF
--- a/lake/root.go
+++ b/lake/root.go
@@ -438,10 +438,10 @@ func (r *Root) batchifyIndexRules(ctx context.Context, zctx *zed.Context, f expr
 	return batch, nil
 }
 
-func (r *Root) NewScheduler(ctx context.Context, zctx *zed.Context, src dag.Source, span extent.Span, filter zbuf.Filter, pred []dag.IndexPredicate) (proc.Scheduler, error) {
+func (r *Root) NewScheduler(ctx context.Context, zctx *zed.Context, src dag.Source, span extent.Span, filter zbuf.Filter, predicates []dag.IndexPredicate) (proc.Scheduler, error) {
 	switch src := src.(type) {
 	case *dag.Pool:
-		return r.newPoolScheduler(ctx, zctx, src.ID, src.Commit, span, filter, pred)
+		return r.newPoolScheduler(ctx, zctx, src.ID, src.Commit, span, filter, predicates)
 	case *dag.LakeMeta:
 		return r.newLakeMetaScheduler(ctx, zctx, src.Meta, filter)
 	case *dag.PoolMeta:
--- a/lake/scheduler.go
+++ b/lake/scheduler.go
@@ -2,6 +2,7 @@ package lake
 
 import (
 	"context"
+	"errors"
 	"sync"
 
 	"github.com/brimdata/zed"
@@ -40,7 +41,7 @@ func NewSortedScheduler(ctx context.Context, zctx *zed.Context, pool *Pool, snap
 		span:   span,
 		filter: filter,
 		index:  index,
-		ch:     make(chan Partition, 10),
+		ch:     make(chan Partition),
 	}
 }
 
@@ -72,6 +73,10 @@ func (s *Scheduler) run() {
 	s.group, ctx = errgroup.WithContext(s.ctx)
 	ch := s.ch
 	if s.index != nil {
+		// Make the index out channel buffered so the index filter reads ahead
+		// of the scan pass, but not too far ahead- the search may be aborted
+		// before the scan pass gets to the end and we don't want to waste
+		// resources running index lookups that aren't used.
 		ch = make(chan Partition)
 		s.group.Go(func() error {
 			defer close(s.ch)
@@ -139,7 +144,10 @@ func indexFilterPass(ctx context.Context, snap commits.View, filter *index.Filte
 	for p := range in {
 		objects := make([]*data.Object, 0, len(p.Objects))
 		for _, o := range p.Objects {
-			r, _ := snap.LookupIndexObjectRules(o.ID)
+			r, err := snap.LookupIndexObjectRules(o.ID)
+			if err != nil && !errors.Is(err, commits.ErrNotFound) {
+				return err
+			}
 			if r != nil {
 				hit, err := filter.Apply(ctx, o.ID, r)
 				if err != nil {


mattnibs:
Need to add tests but this is feature complete, and from what I see actually working.

lake/index/match.go
```diff
@@ -0,0 +1,60 @@
...
+	Rule       Rule
+	LookupKeys []*zed.Record
+}
+
+type predicate struct {
+	field field.Path
+	value zed.Value
+}
+
+type matcher func([]Rule) []match
```
nwt, line 20: matcher wants to be close to its buddy, match.

```suggestion
type matcher func([]Rule) []match

type predicate struct {
	field field.Path
	value zed.Value
}
```

lake/index/rule.go
```diff
@@ -176,3 +176,35 @@ func (a *AggRule) RuleKeys() field.List {
 	// XXX can get these by analyzing the compiled script
 	return nil
 }
+
+// NewLookupKey creates a Zed Record that can be used as a lookup key for an
+// index created for the provided Rule. The Values provided must be in order
+// with the Key in the rule that it will be paired with it.
+func NewLookupKey(zctx *zed.Context, r Rule, values []zed.Value) (*zed.Record, error) {
```
nwt, line 183: Is this going to be used in other packages? If not, don't export.

lake/index/rule.go
```diff
@@ -176,3 +176,35 @@ func (a *AggRule) RuleKeys() field.List {
...
+	keys := r.RuleKeys()
+	// XXX Ensure length of values equals the length of Keys in the Rule or
+	// else zed.ColumnBuilder we throw an error on Encode. We should adjust
```
nwt, line 186:
```suggestion
	// else zed.ColumnBuilder will throw an error on Encode. We should adjust
```

lake/index/rule.go
```diff
@@ -176,3 +176,35 @@ func (a *AggRule) RuleKeys() field.List {
...
+	// zed.ColumnBuilder to allow not be so strict.
```
nwt, line 187:
```suggestion
	// zed.ColumnBuilder to be less strict.
```

lake/index/rule.go
```diff
@@ -176,3 +176,35 @@ func (a *AggRule) RuleKeys() field.List {
...
+	n := len(keys) - len(values)
+	if n < 0 {
+		values = values[:len(keys)]
+	}
+	if n < 0 {
+		values = append(values, make([]zed.Value, n)...)
+	}
```
nwt, line 194:
```suggestion
	if n := len(keys) - len(values); n < 0 {
		values = values[:len(keys)]
	} else if n > 0 {
		values = append(values, make([]zed.Value, n)...)
	}
```

lake/ztests/index/query.yaml
```diff
@@ -0,0 +1,36 @@
+script: |
+  export ZED_LAKE_ROOT=test
+  zed lake init -q
+  zed lake create -q test
+  zed lake use -q "test@main"
```
nwt, line 5:
```suggestion
  zed lake use -q test@main
```

lake/index/match.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"github.com/brimdata/zed"
+	"github.com/brimdata/zed/compiler/ast/dag"
+	"github.com/brimdata/zed/field"
+	"github.com/brimdata/zed/zson"
+)
+
+type match struct {
+	Rule       Rule
+	LookupKeys []*zed.Record
```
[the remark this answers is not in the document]
mattnibs: no reason, will fix

lake/index/rule.go
mattnibs: good point

lake/scheduler.go
```diff
@@ -124,6 +135,34 @@ func (s *scannerScheduler) Stats() zbuf.ScannerStats {
 	return s.stats.Copy()
 }
 
+func indexFilterPass(ctx context.Context, snap commits.View, filter *index.Filter, in <-chan Partition, out chan<- Partition) error {
+	for p := range in {
+		objects := make([]*data.Object, 0, len(p.Objects))
+		for _, o := range p.Objects {
+			r, _ := snap.LookupIndexObjectRules(o.ID)
```
nwt, line 142: Handle this error?

lake/root.go
```diff
@@ -438,10 +438,10 @@ func (r *Root) batchifyIndexRules(ctx context.Context, zctx *zed.Context, f expr
 	return batch, nil
 }
 
-func (r *Root) NewScheduler(ctx context.Context, zctx *zed.Context, src dag.Source, span extent.Span, filter zbuf.Filter) (proc.Scheduler, error) {
+func (r *Root) NewScheduler(ctx context.Context, zctx *zed.Context, src dag.Source, span extent.Span, filter zbuf.Filter, pred []dag.IndexPredicate) (proc.Scheduler, error) {
```
nwt, line 441: Call this parameter preds since it's a slice?

lake/scheduler.go
```diff
@@ -52,39 +54,48 @@ func (s *Scheduler) AddStats(stats zbuf.ScannerStats) {
...
 }
 
 func (s *Scheduler) run() {
-	if err := ScanPartitions(s.ctx, s.snap, s.span, s.pool.Layout.Order, s.ch); err != nil {
-		s.done <- err
+	var ctx context.Context
+	s.group, ctx = errgroup.WithContext(s.ctx)
+	ch := s.ch
+	if s.index != nil {
+		ch = make(chan Partition)
```
nwt, line 75: Give this channel the same amount of buffering as its companion?
```suggestion
		ch = make(chan Partition, cap(s.ch))
```

mattnibs: Actually just this channel needs to be buffered since the object partition pass should be fast compared to the index and scan passes. Will change.

PR #3165, opened 2021-10-11
Remove api.ScannerStats

mattnibs:
Package driver should have no notion of package api. Given that
api.ScannerStats and zbuf.ScannerStats are the same, remove
api.ScannerStats entirely and just use zbuf.ScannerStats.

### DIFF
--- a/zbuf/scanner.go
+++ b/zbuf/scanner.go
@@ -38,13 +38,12 @@ type ScannerCloser interface {
 	io.Closer
 }
 
-// ScannerStats holds Scanner statistics. It should be identical to
-// api.ScannerStats.
+// ScannerStats holds Scanner statistics.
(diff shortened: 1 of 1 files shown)


api/api.go
```diff
@@ -29,13 +30,6 @@ func (e Error) Error() string {
 	return e.Message
 }
 
-type ScannerStats struct {
-	BytesRead      int64 `json:"bytes_read"`
-	BytesMatched   int64 `json:"bytes_matched"`
-	RecordsRead    int64 `json:"records_read"`
-	RecordsMatched int64 `json:"records_matched"`
```
nwt, line 36: This change alters the names of these fields in ZJSON responses because zbuf.ScannerStats doesn't have these struct field tags.

PR #3174, opened 2021-10-12
client.Connection: unmarshal zng responses

mattnibs:
Change client.Connection so it:

1. Assumes that all responses are of type zng
2. Marshal the zng responses into the appropriate response type.

The Resty package got in the way of unmarshaling zng responses
so it was removed.


api/client/connection.go
```diff
@@ -70,207 +56,204 @@ func NewConnection() *Connection {
...
+	if rec == nil {
+		return errors.New("expected one record from response but got none")
+	}
+	return zson.UnmarshalZNGRecord(rec, i)
+}
+
+// parseError parses an error from an http.Response with an error status code. For now the response type of errors is assumed to be JSON.
+func parseError(r *http.Response) error {
+	defer r.Body.Close()
+	body, err := io.ReadAll(r.Body)
```
nwt, line 123: No need to defer.
```suggestion
	body, err := io.ReadAll(r.Body)
	r.Body.Close()
```

api/client/request.go
```diff
@@ -0,0 +1,84 @@
...
+
+func (r *Request) HTTPRequest() (*http.Request, error) {
+	if accept := r.Header.Get("Accept"); accept != api.MediaTypeZNG {
+		return nil, fmt.Errorf("%s: invalid 'Accept' header, only %s accepted", accept, api.MediaTypeZNG)
+	}
+	body, err := r.reader()
+	if err != nil {
+		return nil, err
+	}
+	u := r.host + r.Path
```
nwt, line 70:
```suggestion
	u := path.join(r.host, r.Path)
```

api/client/request.go
```diff
@@ -0,0 +1,84 @@
...
+	req.ctx = httptrace.WithClientTrace(ctx, &httptrace.ClientTrace{
+		DNSStart:             func(httptrace.DNSStartInfo) { req.dnsStartTime = time.Now() },
+		GotConn:              func(g httptrace.GotConnInfo) { req.gotConnInfo = g },
+		GetConn:              func(string) { req.getConnTime = time.Now() },
+		GotFirstResponseByte: func() { req.firstByteTime = time.Now() },
+	})
+	return req
+}
+
+func (r *Request) reader() (io.Reader, error) {
```
nwt, line 51: NIt: This is less interesting than HTTPRequest, so I'd put it after that method.

api/client/request.go
```diff
@@ -0,0 +1,84 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return bytes.NewReader(b), nil
+}
+
+func (r *Request) HTTPRequest() (*http.Request, error) {
+	if accept := r.Header.Get("Accept"); accept != api.MediaTypeZNG {
+		return nil, fmt.Errorf("%s: invalid 'Accept' header, only %s accepted", accept, api.MediaTypeZNG)
+	}
```
nwt, line 65: Maybe just stomp on anything supplied by the user?
```suggestion
	r.Header.Set("Accept", api.MediaTypeZNG)
```

nwt (approved):

service/ztests/service.sh
```diff
@@ -26,7 +26,7 @@ zed lake serve -l=localhost:0 -R $lakeroot -portfile=$portdir/lake -log.level=wa
 lakepid=$!
 awaitfile $portdir/lake
 
-trap "rm -rf $portdir; kill $lakepid;" EXIT
+trap "rm -rf $portdir; kill $lakepid; cat lake.log;" EXIT
```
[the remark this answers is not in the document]
mattnibs: Whoops

api/client/request.go
mattnibs: This wasn't working for me.

PR #3185, opened 2021-10-13
fix bug where consts blocked "from" operator wiring logic

mccanne:
This commit fixes a bug where type and const declarations interfered
with the logic to determine how to wire up inputs to the flowgraph.
The fix is to require all const/type declarations to appear at the
outer, top block and pull them out at the very start of compilation.

This will be reworked when we add modules, subqueries, UDFs, etc.

### DIFF
--- a/compiler/semantic/proc.go
+++ b/compiler/semantic/proc.go
@@ -552,63 +552,43 @@ func semProc(ctx context.Context, scope *Scope, p ast.Proc, adaptor proc.DataAda
 			Type: typ,
 			As:   as,
 		}, nil
+	case *ast.Const:
+		return nil, errors.New("const declaration must appear in outer block")
+	case *ast.TypeProc:
+		return nil, errors.New("type declaration must appear in outer block")
 	}
 	return nil, fmt.Errorf("semantic transform: unknown AST type: %v", p)
 }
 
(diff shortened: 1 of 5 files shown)


compiler/semantic/proc.go
```diff
@@ -552,63 +552,43 @@ func semProc(ctx context.Context, scope *Scope, p ast.Proc, adaptor proc.DataAda
 			Type: typ,
 			As:   as,
 		}, nil
+	case *ast.Const:
+		return nil, errors.New("const declaration must appear in outer block")
+	case *ast.TypeProc:
+		return nil, errors.New("type declaration must appear in outer block")
```
nwt, line 558: Maybe "at top level" instead of "in outer block"?

expr/ztests/type-map.yaml
```diff
@@ -5,7 +5,7 @@ zed: |
     "conn": conn,
     "dns": dns
   }|
-  split (
+  * | split (
```
nwt, line 8: 🤮

nwt (approved):

expr/ztests/type-map.yaml
mccanne: This problem is already there... the bug fix uncovered it.
:-)

PR #3188, opened 2021-10-15
index: Use KeyValue for lookup

mattnibs:
Instead of constructing a record for Lookup of an index value, have
finder.Lookup* accept KeyValue.

### DIFF
--- a/index/finder.go
+++ b/index/finder.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/brimdata/zed"
 	"github.com/brimdata/zed/expr"
+	"github.com/brimdata/zed/field"
 	"github.com/brimdata/zed/order"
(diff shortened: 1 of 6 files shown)


index/finder.go
```diff
@@ -147,14 +153,33 @@ func (f *Finder) search(compare expr.KeyCompareFn) (zio.Reader, error) {
 	return f.newSectionReader(0, off)
 }
 
-func (f *Finder) Lookup(keys *zed.Value) (*zed.Value, error) {
+func compareFn(kvs []KeyValue) expr.KeyCompareFn {
+	accessors := make([]expr.Evaluator, len(kvs))
+	values := make([]zed.Value, len(kvs))
+	fn := expr.NewValueCompareFn(false)
```
nwt, line 159: NIt: Move this definition closer to its fn's use.

PR #3189, opened 2021-10-15
remove zed.Record type alias

nwt:
Generated automatically with the following script.

```sh
sed -i '' '/^type Record = Value/,+1d' recordval.go
gofmt -w -r 'Record -> Value' *.go
gofmt -w -r 'zed.Record -> zed.Value' .
```


mccanne (approved):
awesome

PR #3193, opened 2021-10-19
Add missing Kind to sql agg assignment

mattnibs:
Closes brimdata/brim#1884

### DIFF
--- a/service/ztests/query-sql-agg.yaml
+++ b/service/ztests/query-sql-agg.yaml
@@ -2,10 +2,10 @@
 
 script: |
   source service.sh
-  zed lake create -q test
-  zed lake use -q test@main
-  zed lake load -q data.zson
-  zed lake query -z 'SELECT count(), id GROUP BY id | sort id'
+  zapi create -q test
+  zapi use -q test@main
+  zapi load -q data.zson
+  zapi query -z 'SELECT count(), id GROUP BY id | sort id'
 
 inputs:
   - name: service.sh


lake/ztests/query-sql-agg.yaml
```diff
@@ -0,0 +1,22 @@
+# brimdata/brim#1884
+
+script: |
+  export ZED_LAKE_ROOT=test
+  zed lake init -q
+  zed lake create -q test
+  zed lake use -q test@main
+  zed lake load -q data.zson
+  zed lake query -z 'from test | SELECT count(), id GROUP BY id | sort id'
```
nwt, line 9: No need for `from` after `zed lake use`.
```suggestion
  zed lake query -z 'SELECT count(), id GROUP BY id | sort id'
```

service/ztests/query-sql-agg.yaml
```diff
@@ -0,0 +1,22 @@
+# brimdata/brim#1884
+
+script: |
+  source service.sh
+  zed lake create -q test
+  zed lake use -q test@main
+  zed lake load -q data.zson
+  zed lake query -z 'from test | SELECT count(), id GROUP BY id | sort id'
```
nwt, line 8:
```suggestion
  zapi create -q test
  zapi use -q test@main
  zapi load -q data.zson
  zapi query -z 'SELECT count(), id GROUP BY id | sort id'
```

mattnibs: I'd prefer to keep using `zed lake` since if I make a change I can easily copy this to lake and vice versa.

nwt: But this needs to use `zapi` or `zed api` in order to talk to `zed lake serve`. `zed lake` just goes straight to the file system.

mccanne: Good catch!  I missed that one.

nwt (approved):

Issue #3183, opened 2021-10-13
more fully featured index predicate expressions


mattnibs:
Right now the optimize pass that computes what part of a filter expression can be used in an index lookup is kind of dumb; it merely creates a list of field equal expression. The index then takes this list and does a look up for each field/value combination and if it has at least one match among the list, the data object is scanned. There's some features in a filter expression that if taken into account could speed up the index:
- Binary `and` operations
- `>`, `>=`, `<`, `<=` operators
- `in` operators

## And operations
If you are searching for a chain of `or` key/value expressions this works fine (e.g. `foo == 1 or bar == 2`), but what if your query looks like `foo == 1 bar == 2`? While currently an object would be scanned if there is a match for `foo == 1` or `bar ==2` but ideally you would only scan the object if both predicates exist in the index.

## gt, gte, lt, lte operators

Since an index is essentially a list of ordered values we could check an index to see if there are any values that meet these conditions

## in operators

There's two flavors of lookups here. The first is seeing if an index has values within a range (e.g. an ip subnet). Here the logic is similar to the comparator operators mentioned above. The second is seeing if an index has a match amongst a set of values. This is will be the same as doing a chan of `or` index lookups on all the members of the set.

## != operators
~~Additionally the optimize pass completely ignore unary expressions and `!=` comparators which is another area where the index could be used to filter out objects. XXX I am recalling some trickiness here in previous incarnations of index having to do with the fact that the index is looking at batches of records and not individual records and the naive approach created some false negatives with `!=` expressions. I do not recall the exact situations, I need to dig back in the old source code and figure this out.~~
EDIT You cannot actually use the `!=` operation in the index because while an index may contain a value, there may be a record that does not contain that value. For instance a naive implementation for the query `x != 1`, would skip a data object if there was a hit in the index for `x == 1`, but there could be a whole of records in the data object where `x != 1` is true.

## unary operations
For the same reason that you can't use != operators, I am skeptical that use of the unary expression in an index filter. I need to fully think this case through.

nwt:
Closing because #4972 removed the index package.

PR #3196, opened 2021-10-19
index.Finder.Nearest: support lt and gt

mattnibs:
Change the name of closet to nearest, make it a public method and
support > and < operations. Get rid of ClosetGTE/LTE and have users just
call nearest with the correct operator.

Part of #3183

### DIFF
--- a/index/finder.go
+++ b/index/finder.go
@@ -76,8 +76,7 @@ func lookupAsc(reader zio.Reader, fn expr.KeyCompareFn, op Operator) (*zed.Value
 			}
 			return prev, err
 		}
-		cmp := fn(rec)
-		if cmp >= 0 {
+		if cmp := fn(rec); cmp >= 0 {
 			if cmp == 0 && op.hasEqual() {
 				return rec, nil
 			}
--- a/index/finder.go
+++ b/index/finder.go
@@ -215,7 +215,7 @@ func compareFn(kvs []KeyValue) expr.KeyCompareFn {
 }
 
 // Nearest finds the zed.Value in the index that is nearest to kvs according to
-// the provided operator.
+// operator.
 func (f *Finder) Nearest(operator string, kvs ...KeyValue) (*zed.Value, error) {
 	op := Operator(operator)
 	if !op.valid() {
--- a/index/finder.go
+++ b/index/finder.go
@@ -273,7 +273,6 @@ func (o Operator) valid() bool {
 	switch o {
 	case EQL, GT, GTE, LT, LTE:
 		return true
-	default:
-		return false
 	}
+	return false
 }
--- a/index/finder.go
+++ b/index/finder.go
@@ -264,9 +264,8 @@ func (o Operator) hasEqual() bool {
 	switch o {
 	case EQL, GTE, LTE:
 		return true
-	default:
-		return false
 	}
+	return false
 }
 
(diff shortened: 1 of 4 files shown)


index/finder.go
```diff
@@ -45,72 +55,69 @@ func NewFinder(ctx context.Context, zctx *zed.Context, engine storage.Engine, ur
...
-			if op == eql {
-				rec = nil
-			}
-			if op == gte {
+			if op == GTE || op == GT {
 				return prev, nil
 			}
-			return rec, nil
+			if op == EQL {
+				return nil, nil
```
nwt, line 116: Should this case remain consistent with the corresponding case in lookupAsc? (They were previously identical but now diverge.)

index/finder.go
```diff
@@ -231,7 +214,13 @@ func compareFn(kvs []KeyValue) expr.KeyCompareFn {
 	}
 }
 
-func (f *Finder) closest(kvs []KeyValue, op operator) (*zed.Value, error) {
+// Nearest finds the zed.Value in the index that is nearest to kvs according to
+// the provided operator.
```
nwt, line 218:
```suggestion
// operator.
```

index/finder.go
```diff
@@ -270,3 +259,21 @@ func (f *Finder) ParseKeys(inputs ...string) ([]KeyValue, error) {
...
 	return kvs, nil
 }
+
+func (o Operator) hasEqual() bool {
+	switch o {
+	case EQL, GTE, LTE:
+		return true
+	default:
+		return false
+	}
```
nwt, line 269:
```suggestion
	}
	return false
```

index/finder.go
```diff
@@ -270,3 +259,21 @@ func (f *Finder) ParseKeys(inputs ...string) ([]KeyValue, error) {
...
+}
+
+func (o Operator) valid() bool {
+	switch o {
+	case EQL, GT, GTE, LT, LTE:
+		return true
+	default:
+		return false
+	}
```
nwt, line 278:
```suggestion
	}
	return false
```

index/index_test.go
```diff
@@ -85,59 +85,51 @@ func TestCompare(t *testing.T) {
...
-			}
-
+			rec, err := finder.Nearest(op, kvs...)
 			require.NoError(t, err)
-			if expected == -1 {
-				assert.Nil(t, rec)
-			} else {
-				require.NotNil(t, rec)
-				v, err := rec.AccessInt("ts")
+			var v int64 = -1
```
nwt, line 103:
```suggestion
			v := int64(-1)
```

nwt (approved):

index/finder.go
mattnibs: They are the same thing but I'll change the one above to this.

index/finder.go
```diff
@@ -45,72 +55,68 @@ func NewFinder(ctx context.Context, zctx *zed.Context, engine storage.Engine, ur
...
-			if cmp == 0 {
+			if cmp == 0 && op.hasEqual() {
 				return rec, nil
 			}
-			if op == eql {
+			if op == LTE || op == LT {
+				return prev, nil
+			}
+			if op == EQL {
 				rec = nil
```
mattnibs, line 87: return nil, nil

PR #3197, opened 2021-10-19
lake index.Filter: Use expression tree for lookups

mattnibs:
First this pr modifies the contents of the index Filter pushdown so
rather than just storing predicates it creates a simplified verision
of binary expression that the index can use.

Secondly the evaluation part of index.Filter is modified so it make use
the expression tree. For instance if the left hand side of an or
operation returns a hit, the expression will return true without having
to wait for the right hand expression to complete. The index.Filter also
uses the index.Finder.Nearest method so compare expressions with the
operators >, >=, <, <= can also be used.

Closes 3183

### DIFF
--- a/service/ztests/index/query-relative.yaml
+++ b/service/ztests/index/query-relative.yaml
@@ -1,18 +1,18 @@
 script: |
   source service.sh
-  zed lake create -q test
-  zed lake use -q test@main
-  zed lake index create -q s field s
+  zapi create -q test
+  zapi use -q test@main
+  zapi index create -q s field s
   # Load these separately so we have 3 different objects.
-  zed lake load -q 1.zson
-  zed lake load -q 2.zson
-  zed lake load -q 3.zson
-  zed lake index update -q
-  zed lake query -z -s 's>3 | count()'
-  zed lake query -z -s 's<5.09 | count()'
-  zed lake query -z -s 's>=9.5 | count()'
-  zed lake query -z -s 's<=2 | count()'
-  zed lake query -z -s 's<1 | count()'
+  zapi load -q 1.zson
+  zapi load -q 2.zson
+  zapi load -q 3.zson
+  zapi index update -q
+  zapi query -z -s 's>3 | count()'
+  zapi query -z -s 's<5.09 | count()'
+  zapi query -z -s 's>=9.5 | count()'
+  zapi query -z -s 's<=2 | count()'
+  zapi query -z -s 's<1 | count()'
 
 inputs:
   - name: service.sh


compiler/ast/dag/operator.go
```diff
@@ -150,14 +150,10 @@ type (
 	}
 
 	Pushdown struct {
-		Scan  Op               `json:"scan"`
-		Index []IndexPredicate `json:"index"`
+		Scan  Op      `json:"scan"`
+		Index *Filter `json:"index"`
```
nwt, line 154: This feels to me like it wants to be called IndexFilter/index_filter (but I'm OK with the current name if you don't agree).

service/ztests/index/query.yaml
```diff
@@ -0,0 +1,38 @@
...
+  zed lake load -q 1.zson
+  zed lake load -q 2.zson
+  zed lake load -q 3.zson
+  zed lake query -s -o /dev/null 's==127.0.0.1' 2> stats.zson
+  zed lake index update -q
+  zed lake query -s -o /dev/null 's==127.0.0.1' 2>> stats.zson
+  zed lake query -s -o /dev/null 's==1' 2>> stats.zson
+  zed lake query -s -o /dev/null 's=="hello"' 2>> stats.zson
+  zed lake query -s -o /dev/null 's=="hello" or s==1 or s==127.0.0.1' 2>> stats.zson
+  zq -z 'cut records_read, records_matched' stats.zson
```
nwt, line 16: Any reason not to let `zed lake query` write results to standard output and stats to standard error (and then check both) like you do in `query-relative.yaml`?

service/ztests/index/query.yaml
```diff
@@ -0,0 +1,38 @@
...
+  # Load these separately so we have 3 different objects.
+  zed lake load -q 1.zson
+  zed lake load -q 2.zson
+  zed lake load -q 3.zson
+  zed lake query -s -o /dev/null 's==127.0.0.1' 2> stats.zson
+  zed lake index update -q
+  zed lake query -s -o /dev/null 's==127.0.0.1' 2>> stats.zson
+  zed lake query -s -o /dev/null 's==1' 2>> stats.zson
+  zed lake query -s -o /dev/null 's=="hello"' 2>> stats.zson
+  zed lake query -s -o /dev/null 's=="hello" or s==1 or s==127.0.0.1' 2>> stats.zson
```
nwt, line 15: s/zed lake/zapi/

service/ztests/index/query.yaml
mattnibs: Yeah probably should do that. I didn't want to check bytes because that could change, but then again zson is pretty stable.

compiler/ast/dag/operator.go
mattnibs: Would you also change Scan to ScanFilter?

service/ztests/index/query.yaml
mattnibs: Oh snap, these should be zapi as well.

PR #3202, opened 2021-10-21
Use zson.Unmarshal request bodies

mattnibs:
As we already do for endpoint responses, use zson.Unmarshal to unmarshal
the body of requests.


service/request.go
```diff
@@ -137,8 +138,34 @@ func (r *Request) BoolFromQuery(param string, w *ResponseWriter) (bool, bool) {
...
 }
 
-func (r *Request) Unmarshal(w *ResponseWriter, body interface{}) bool {
-	if err := json.NewDecoder(r.Body).Decode(body); err != nil {
+func (r *Request) Unmarshal(w *ResponseWriter, body interface{}, templates ...interface{}) bool {
+	typ := r.Header.Get("Content-Type")
+	if typ == "" || typ == "application/x-www-form-urlencoded" {
+		// If Content-Type is set or is a form (probably set from curl), assume
+		// json.
+		typ = api.MediaTypeJSON
```
nwt, line 146: I think we should use `curl --data-binary` instead of ignoring `application/x-www-form-urlencoded` like this. 
```suggestion
	if typ == "" {
		typ = api.MediaTypeJSON
```

api/client/request.go
```diff
@@ -66,11 +69,21 @@ func (r *Request) reader() (io.Reader, error) {
 	if b, ok := r.Body.(io.Reader); ok {
 		return b, nil
 	}
-	b, err := json.Marshal(r.Body)
+	var buf bytes.Buffer
```
nwt, line 72: Nit: Declare this closer to its use.

api/client/request.go
```diff
@@ -66,11 +69,21 @@ func (r *Request) reader() (io.Reader, error) {
...
+	m := zson.NewZNGMarshaler()
+	m.Decorate(zson.StylePackage)
+	v, err := m.Marshal(r.Body)
```
nwt, line 75: Nit: s/v/zv/ to make it a little clearer that this is a zed.Value.

service/request.go
```diff
@@ -137,8 +138,34 @@ func (r *Request) BoolFromQuery(param string, w *ResponseWriter) (bool, bool) {
 	return b, true
 }
 
-func (r *Request) Unmarshal(w *ResponseWriter, body interface{}) bool {
-	if err := json.NewDecoder(r.Body).Decode(body); err != nil {
+func (r *Request) Unmarshal(w *ResponseWriter, body interface{}, templates ...interface{}) bool {
+	typ := r.Header.Get("Content-Type")
+	if typ == "" || typ == "application/x-www-form-urlencoded" {
+		// If Content-Type is set or is a form (probably set from curl), assume
+		// json.
```
nwt, line 145:
```suggestion
		// If Content-Type is unset or is a form (probably set from curl), assume
		// JSON.
```

PR #3229, opened 2021-10-29
Query endpoint: default to ZSON response

mattnibs:
For the query endpoint, if the "Accept" endpoint is not specified
default to ZSON.

Part of brimdata/brim#1905


mattnibs:
Closing and going with the approach in #3242 instead.

PR #3230, opened 2021-10-30
allow leading split operator

nwt:
Allow a leading split operator, including one whose branches all have a
leading from operator, as in "split (=> from (...); => from (...);)".


mccanne (approved):
nice!

PR #3234, opened 2021-11-02
handle pool and branch errors in service.errorResponse

nwt:
Pool and branch error handling is duplcated accross some handlers and
missing from others.  Move it into service.errorResponse to remove
duplication and to apply it uniformly when handlers call
service.Request.Error.

### DIFF
--- a/service/handlers.go
+++ b/service/handlers.go
@@ -10,11 +10,9 @@ import (
 	"github.com/brimdata/zed/compiler"
 	"github.com/brimdata/zed/driver"
 	"github.com/brimdata/zed/lake"
-	"github.com/brimdata/zed/lake/branches"
 	"github.com/brimdata/zed/lake/commits"
 	"github.com/brimdata/zed/lake/index"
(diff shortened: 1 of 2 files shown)


mattnibs (commented):
I'm not sure how I feel about this. On one hand the I like less code, but also I feel like this makes the error handling kind of opaque. Maybe there's some way to set the expected error codes in the response writer at the beginning of each request?

nwt:
That wouldn't really do anything to ensure that we always handle these errors, though.

We could write wrappers for methods like `lake.Root.OpenPool` in which we transmute known errors into a `zqe.Error`, but that's going to be more verbose and less robust (since we might forget to use a wrapper). Or we could use `zqe` in `lake/pools` and `lake/branches`, but ugh.

mattnibs:
>  Or we could use zqe in lake/pools and lake/branches, but ugh.

I was thinking of dropping `zqe` soon.

mattnibs (approved):
I suppose these errors are common enough that this is fine.

PR #3237, opened 2021-11-03
Enable embedding custom metadata in the commit journal

silveryfu:
This PR allows embedding a metadata field (`meta`) in the commit object at `zed lake load`. The `meta` field can include any ZSON/ZNG data such that external applications can embed any custom metadata in the commit journal and implement arbitrary data provenance and governance capabilities later.

For example,

Load data with `-meta` option:
`zed lake load -meta '{class: "sensitive"}' in.zson`

Filter commits by metadata:
`zed lake query "from pool@main:log | meta.class=='sensitive'"`

One can embed any ZSON/ZNG data, e.g.,
`zed lake load -meta '[{a:1}, {b:"hi there"}]' in.zson`

This PR does not do/check: allowing data to be arbitrarily large (that it can't fit in an HTTP header over the load endpoint with remote driver).

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"context"
+	
 	"github.com/brimdata/zed/lake/index"
 	"github.com/brimdata/zed/lakeparse"
 	"github.com/brimdata/zed/order"
--- a/api/api.go
+++ b/api/api.go
@@ -54,9 +54,9 @@ type BranchMergeRequest struct {
 }
 
 type CommitMessage struct {
-	Author string    `zed:"author"`
-	Body   string    `zed:"body"`
-	Meta   string 	 `zed:"meta"`
+	Author string `zed:"author"`
+	Body   string `zed:"body"`
+	Meta   string `zed:"meta"`
 }
 
 type CommitResponse struct {
--- a/api/api.go
+++ b/api/api.go
@@ -2,7 +2,7 @@ package api
 
 import (
 	"context"
-	
+
 	"github.com/brimdata/zed/lake/index"
 	"github.com/brimdata/zed/lakeparse"
 	"github.com/brimdata/zed/order"
--- a/api/api.go
+++ b/api/api.go
@@ -2,7 +2,6 @@ package api
 
 import (
 	"context"
-
 	"github.com/brimdata/zed/lake/index"
 	"github.com/brimdata/zed/lakeparse"
 	"github.com/brimdata/zed/order"
@@ -55,8 +54,9 @@ type BranchMergeRequest struct {
 }
 
 type CommitMessage struct {
-	Author string `zed:"author"`
-	Body   string `zed:"body"`
+	Author string    `zed:"author"`
+	Body   string    `zed:"body"`
+	Meta   string 	 `zed:"meta"`
 }
 
 type CommitResponse struct {
--- a/lake/branch.go
+++ b/lake/branch.go
@@ -28,7 +28,7 @@ const (
 
 var (
 	ErrCommitFailed      = fmt.Errorf("exceeded max update attempts (%d) to branch tip: commit failed", maxCommitRetries)
-	ErrInvalidCommitMeta = errors.New("cannot parse zson string")
+	ErrInvalidCommitMeta = errors.New("cannot parse ZSON string")
 )
 
 type Branch struct {
--- a/lake/branch.go
+++ b/lake/branch.go
@@ -27,7 +27,7 @@ const (
 )
 
 var (
-	ErrCommitFailed = fmt.Errorf("exceeded max update attempts (%d) to branch tip: commit failed", maxCommitRetries)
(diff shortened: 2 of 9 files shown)


api/api.go
```diff
@@ -2,7 +2,6 @@ package api
 
 import (
 	"context"
-
```
nwt, line 5: We separate standard library imports from 3rd-party imports with a blank line, so keep this. (If configured to manage imports, `gopls` should include this.)

lake/commits/actions.go
```diff
@@ -56,15 +57,20 @@ type Commit struct {
...
 
 func (c *Commit) CommitID() ksuid.KSUID {
 	return c.ID
 }
 
-func (c *Commit) String() string {
-	//XXX need to format Message field for single line
-	return fmt.Sprintf("COMMIT %s -> %s %s %s %s", c.ID, c.Parent, c.Date, c.Author, c.Message)
+func (c *Commit) AppMeta() zed.Value {
+	return c.Meta
```
nwt, line 73: These method changes don't appear to be necessary for this PR, so remove them to minimize the diff.

api/api.go
silveryfu: thanks Noah! done!

lake/commits/actions.go
silveryfu: Ah, done! I'll double-check diff in future. thanks!

lake/commits/actions.go
```diff
@@ -67,6 +69,10 @@ func (c *Commit) String() string {
 	return fmt.Sprintf("COMMIT %s -> %s %s %s %s", c.ID, c.Parent, c.Date, c.Author, c.Message)
 }
 
+func (c *Commit) AppMeta() zed.Value {
+	return c.Meta
+}
+
```
nwt, line 75: This is unused, so let's not add it.

lake/commits/object.go
```diff
@@ -21,7 +22,7 @@ type Object struct {
 	Actions []Action    `zed:"actions"`
 }
 
-func NewObject(parent ksuid.KSUID, author, message string, retries int) *Object {
+func NewObject(parent ksuid.KSUID, author, message string, retries int, meta zed.Value) *Object {
```
nwt, line 25: Nit: Put this next to `author` and `message`.
```suggestion
func NewObject(parent ksuid.KSUID, author, message string, meta zed.Value, retries int) *Object {
```

nwt (commented):
Forgot to say that this looks great overall! A few more tweaks and we'll merge it.

lake/commits/actions.go
silveryfu: nice catch! removed!

lake/commits/object.go
silveryfu: makes sense and done! thank you! :)

mccanne:
Nice work @silveryfu!

lake/branch.go
```diff
@@ -89,6 +94,17 @@ func loadMessage(objects []data.Object) string {
 	return b.String()
 }
 
+func loadMeta(meta string) (zed.Value, error) {
+	if meta == "" {
+		return zed.Value{zed.TypeNull, nil}, nil
+	}
+	zv, err := zson.ParseValue(zed.NewContext(), meta)
+	if err != nil {
+		return zed.Missing, fmt.Errorf("cannot parse zson string %s: %v", zv, err)
```
mattnibs, line 103: You should probably create a public error called `ErrInvalidCommitMeta` that looks like:
```
var ErrInvalidCommitMeta = errors.New("cannot parse zson string")
```
Then the `handleBranchLoad` handler on line 316 of service/handlers.go should check for this error and return ErrInvalid so the error code will be 400 in the response.

```suggestion
		return zed.Missing, fmt.Errorf("%w %s: %v", ErrInvalidCommitMeta, zv, err)
```

silveryfu: Thanks Matt! This makes a lot sense! Done!

mattnibs (approved):
lgtm

lake/branch.go
```diff
@@ -25,7 +26,10 @@ const (
 	maxMessageObjects = 10
 )
 
-var ErrCommitFailed = fmt.Errorf("exceeded max update attempts (%d) to branch tip: commit failed", maxCommitRetries)
+var (
+	ErrCommitFailed      = fmt.Errorf("exceeded max update attempts (%d) to branch tip: commit failed", maxCommitRetries)
+	ErrInvalidCommitMeta = errors.New("cannot parse zson string")
```
nwt, line 31:
```suggestion
	ErrInvalidCommitMeta = errors.New("cannot parse ZSON string")
```

nwt (commented):
One last request before we merge this.

lake/branch.go
silveryfu: Thanks Noah! Done!

PR #3239, opened 2021-11-03
capitalize first letter of Zed where appropriate

nwt:
Capitalize the first letter of Zed unless it's used as command name or
an identifier.


mattnibs (approved):
Good to know!

PR #3246, opened 2021-11-05
Zedify events handler

mattnibs:
Zedify objects fed into the events handler. This fixes an issue where ksuids where getting formatted differently for the /events endpoint vs the other endpoints in the app.

### DIFF
--- a/service/core.go
+++ b/service/core.go
@@ -234,7 +234,7 @@ func (c *Core) publishEvent(w *ResponseWriter, name string, data interface{}) {
 		return
 	}
 	go func() {
-		ev := event{name: name, value: zv}
+		ev := event{name: name, value: &zv}
 		c.subscriptionsMu.RLock()
 		for sub := range c.subscriptions {
 			sub <- ev
--- a/service/core.go
+++ b/service/core.go
@@ -230,7 +230,7 @@ func (c *Core) requestLogger(r *http.Request) *zap.Logger {
 func (c *Core) publishEvent(w *ResponseWriter, name string, data interface{}) {
 	zv, err := zson.MarshalZNG(data)
 	if err != nil {
-		w.Logger.Error("error marshaling published event", zap.Error(err))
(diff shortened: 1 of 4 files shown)


service/core.go
```diff
@@ -226,17 +227,17 @@ func (c *Core) requestLogger(r *http.Request) *zap.Logger {
 	return c.logger.With(zap.String("request_id", api.RequestIDFromContext(r.Context())))
 }
 
-func (c *Core) publishEvent(event string, data interface{}) {
+func (c *Core) publishEvent(w *ResponseWriter, name string, data interface{}) {
+	zv, err := zson.MarshalZNG(data)
+	if err != nil {
+		w.Logger.Error("error marshaling published event", zap.Error(err))
```
nwt, line 233: Capitalize the first word of a log string.
```suggestion
		w.Logger.Error("Error marshaling published event", zap.Error(err))
```

service/handlers.go
```diff
@@ -126,9 +126,7 @@ func handlePoolPost(c *Core, w *ResponseWriter, r *Request) {
 		w.Error(err)
 		return
 	}
-	c.publishEvent("pool-new", api.EventPool{
-		PoolID: pool.ID.String(),
-	})
+	c.publishEvent(w, "pool-new", api.EventPool{PoolID: pool.ID})
 	w.Respond(http.StatusOK, meta)
```
nwt, line 130: I think we should publish events after sending the response.

service/eventstream.go
```diff
@@ -0,0 +1,37 @@
...
+	"io"
+
+	"github.com/brimdata/zed"
+	"github.com/brimdata/zed/zio"
+	"github.com/brimdata/zed/zio/anyio"
+)
+
+type event struct {
+	name  string
+	value zed.Value
```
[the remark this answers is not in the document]
mattnibs: Ok I wasn't sure which was the way of the future 🚀

nwt (approved):

PR #3257, opened 2021-11-08
change zed.Value.Copy to return a pointer

nwt:
This will let us remove zed.Value.Keep.


mccanne (approved):
We should discuss the long-term plan to eliminate copies altogether with the new Batch implementation... the contract is you only have to copy when you want your ownership to last beyond the Batch's lifetime.

nwt:
That's a nice statement of the contract. And let's discuss the long-term plan tomorrow.

PR #3263, opened 2021-11-09
remove TZNG and most of zio/tzngio

nwt:
TZNG has served its purpose.  Now it's just getting in the way of CSV
detection in zio/anyio.

Other packages depend on what remains of zio/tzngio.  We'll unwind them
and remove the remnants later.


mattnibs (approved):
🎉

Issue #2517, opened 2021-04-08
Auto-detect all input formats for files


nwt:
`zed` doesn't try to detect CSV, Parquet, JSON, or ZST inputs, but for file inputs, it should.

It doesn't because `zio/detector.NewReaderWithOpts` wraps its `io.Reader` parameter with `Track` and `Recorder`, which don't implement `io.ReadSeeker` and so aren't compatible with `zio/parquetio.NewReader` or `zio/zstio.NewReader`. But if the `io.Reader` parameter does implement `io.ReadSeeker`, `NewReaderWithOpts` can try Parquet, JSON, and ZST first, using `Seek` to rewind for the next format after each try.

philrz:
Note to self: For now I've added a comment in the "Custom Brimcap Configuration" article linking to this as an open issue, so that way the reader understands the current limitation is temporary. If/when we address this issue, I should update the wiki article to remove the workaround and the comment.

philrz:
Verified in Zed commit e569877.

Here's an example of reading each of the supported formats (with the exception of `line`) using auto-detect.

```
$ zq -version
Version: v1.3.0-42-ge5698777

$ for format in arrows zng vng json zeek zjson csv parquet zson; do   echo '{"hello": "world", "pi": 3.14}' | zq -f $format -o sample.$format - ; done

$ for file in *; do echo -n "$file: ";   zq 'count()' $file; done
sample.arrows: {count:1(uint64)}
sample.csv: {count:1(uint64)}
sample.json: {count:1(uint64)}
sample.parquet: {count:1(uint64)}
sample.vng: {count:1(uint64)}
sample.zeek: {count:1(uint64)}
sample.zjson: {count:1(uint64)}
sample.zng: {count:1(uint64)}
sample.zson: {count:1(uint64)}
```

`line` format still requires an explicit `-i`, since there's no way to automatically determine intent to treat input this way.

```
$ zq -i line sample.zson
"{"
"    hello: \"world\","
"    pi: 3.14"
"}"

$ zq -i line sample.csv
"hello,pi"
"world,3.14"
```

And, as mentioned in #4270, the non-seekable formats like Parquet and VNG aren't readable if compressed, whether via auto-detect or expliit.

```
$ gzip *

$ for file in *; do echo -n "$file: ";   zq 'count()' $file; done
sample.arrows.gz: {count:1(uint64)}
sample.csv.gz: {count:1(uint64)}
sample.json.gz: {count:1(uint64)}
[...]

PR #3277, opened 2021-11-12
attempt to detect CSV input automatically

nwt:
In zio/anyio.NewReaderWithOpts, return a zio/csvio.Reader if no other
format matches, the first line of input contains a comma, and a call to
match succeeds.

For #2517.


mccanne (approved):
cool

PR #3283, opened 2021-11-14
add zed lake serve -rootcontentfile flag

nwt:
The -rootcontentflag flag to "zed lake serve" specifies a file whose
contents are served for GET / requests (in place of the default content
in service.indexPage).


mattnibs (commented):
Idk, I feel like this makes it more difficult to set up a cloud environment in that I now have to have a index.html on hand to service the right index. Couldn't we just have a separate cloud index file in repo that gets served when a flag is enabled?

nwt:
I think that file belongs in another repository. If it's here, then deploying a copy change also deploys a new `zed` binary, and I don't think we want that coupling.

mattnibs (approved):

Issue #3231, opened 2021-11-01
traverse operator


mccanne:
Add a "traverse" operator to pivot between streams of records at operator level and streams of values based on descending into and interating over complex values.

For example,  for input `{a:[0,1,2,3,},{a:[2,3,4,5]}`, this Zed
```
traverse a ( 
  sum(this) by bin:=this/2 
)
```
will give
```
{sum:1,bin:=0}
{sum:5,bin:=1}
{sum:5,bin:=1}
{sum:9,bin:=2}
```
This will involve unify `proc.Interface` and `expr.Generator`.

This works like "every" where the group-by within the traverse clause is reset for every input.

PR #3286, opened 2021-11-17
First cut of traverse Proc

mattnibs:
This commit adds a first cut of the traverse Proc.

This current version of traverse is limited in the following ways:

1. It only works with a subset of the available procs. This is because
   traverse requires the procs in its sub-sequence to adhere to a new
   definition of what proc.EOS means. Current procs completely shutdown
   when EOS is encountered; future work will change the various procs so
   that EOS signals them to reset all state but not to shutdown.
   Traverse should work fine on all stateless operators.

2. Traverse currently doesn't support iteration over records and maps.
   For records and sets the root record to each element of the array. There's
   some design work needed to decide how the key values in a record/map
   elements should be referenced.

PartOf: #3231


compiler/semantic/proc.go
```diff
@@ -552,6 +552,20 @@ func semProc(ctx context.Context, scope *Scope, p ast.Proc, adaptor proc.DataAda
 			Type: typ,
 			As:   as,
 		}, nil
+	case *ast.Traverse:
+		e, err := semExpr(scope, p.Expr)
+		if err != nil {
+			return nil, err
+		}
+		seq, err := semSequential(ctx, scope, p.Seq, adaptor, head)
```
mattnibs, line 560: At some point this should check that the resulting proc does not start with `From`.

mccanne (approved):

Issue #3027, opened 2021-09-07
Performance degradation as commits accumulate


philrz:
@mccanne identified a performance exposure as the commit log gets long, so I ran the following script on my Mac laptop to quantify the degradation. The test data is the first Zeek `conn` record from the [zed-sample-data](https://github.com/brimdata/zed-sample-data) in ZNG format.

```
#!/bin/sh
rm -rf $ZED_LAKE_ROOT
zed lake init
zed lake create logs
zed lake checkout logs@main
NUM=1
while true; do
  echo "Performing commit #${NUM}:"
  time zed lake load one_conn.zng
  echo
  echo "Performing minimal query:"
  time zed lake query 'head 1' > /dev/null
  echo
  NUM=$(expr $NUM + 1)
done
```

As shown in the chart below, the simple `head 1` query goes from 23 milliseconds when the pool contains just the one commit to 4+ seconds when there's 21k commits.

![](https://user-images.githubusercontent.com/5934157/132365667-9e7d8c86-d31e-47d1-b73b-0bf52462cb68.png)

mccanne:
Just to be clear, this is all "by design" and shall be taken care of by not-yet-implemented cats and snanpshots of the commit history.  Since everything is immutable (and since ZNG can be simply concatenated), it is trivial to concatenate a commit history at any object and to create snapshots of the current state of the object store at any commit.

mccanne:
I'm also curious about our commit-per-second speed.  There are some easy things here we can do here to speed this up too...

mccanne:
(The commit history overhead slows down the commits because you have to build the commit state for each commit made...)

mccanne:
Also query times are slow because of the massive number of single-record ZNG objects in the store.  Small commits will be address with LSM-style rollups via the forthcoming compat command (which is trivial to implement... just read a bunch of small objects in a mergescan and write them back).

philrz:
The addition of `zed compact` via #3940 justifies revisiting the tests performed here to see if compacting the objects from the large number of commits might improve query performance.,

As the initial tests were performed back in September of 2021, it first seemed appropriate to re-baseline by repeating the original test with current Zed. The Zed commands have changed a bit in the time since, so here's an update of the repro script that's in sync with current Zed commit 28029f06.

```
#!/bin/sh
export ZED_LAKE="$HOME/.zed-3027-lake"
rm -rf $ZED_LAKE
zed init
zed create logs
zed use logs@main
NUM=1
while true; do
[...]

PR #3296, opened 2021-11-29
serialize snapshots in lake/commit.Store.Snapshot

nwt:
Reduce the number of storage objects read by lake/commit.Store.Snapshot
in the common case by storing and retrieving serialized snapshots.  This
implementation stores every snapshot it ~~creates, including intermediates~~ returns.

For #3027.


lake/commits/store.go
```diff
@@ -86,6 +87,12 @@ func (s *Store) Snapshot(ctx context.Context, leaf ksuid.KSUID) (*Snapshot, erro
...
 	}
+	if snap, err := s.getSnapshot(ctx, leaf); err != nil && !errors.Is(err, fs.ErrNotExist) {
+		return nil, err
+	} else if err == nil {
+		s.snapshots[leaf] = snap
+		return snap, nil
+	}
 	var objects []*Object
 	var base *Snapshot
 	for at := leaf; at != ksuid.Nil; {
```
mattnibs, line 98: Should we be also doing `getSnapshots` when we're reversing through history?

nwt: I'd been thinking we shouldn't because that would add extra latency, but it won't if we fetch the next object and next snapshot concurrently, so let me see how it looks.

mattnibs (approved):
Nicely done 👍🏻

mason-fish:
I've verified that this does increase the performance time when querying branches that have previously been queried. I used the following script to get a feel for the change in performance:
```bash
#!/bin/sh
rm -rf $ZED_LAKE_ROOT
zed lake init
zed lake create logs
zed lake use logs@main

TIMEFORMAT=%R
NUM=1
while true; do
  if (( $NUM % 1000 == 0))
  then
  echo "Performing minimal query after ${NUM} commits:"
  time zed lake query 'head 1' > /dev/null
  echo
  echo "Re-running same query against same commit"
  time zed lake query 'head 1' > /dev/null
  echo
  fi
  zed lake load first_conn.zng > /dev/null
  NUM=$(expr $NUM + 1)
done
``` 

which produced this data
| Commits | First Query Time (s)  | Second Query Time (s) |
| ------------- | ------------- | ------------|
| 1000  | 0.297  | 0.144 | 
| 2000  | 0.432  | 0.273 | 
| 3000  | 0.590  | 0.397 | 
| 4000  | 0.783  | 0.524 | 
| 5000  | 1.721  | 0.651 | 
| 6000  | 1.872  | 0.775 | 
| 7000  | 1.127  | 0.934 | 
| 8000  | 1.346  | 1.149 | 
| 9000  | 1.609  | 1.532 | 
| 10000  | 3.602  | 2.065 | 

The results are interesting, the non-initial queries can still be slow (at 10k commits even the second+ queries can take 2 seconds...) but they are definitely faster than the first query to a commit.

Thanks @nwt!

Issue #3018, opened 2021-09-04
seek range in index objects


mccanne:
Search index objects should have an option to keep the seek range in the index.  This would speed up access a bit for needle searches.  When the seek range is large it can simply be omitted with a null value.

PR #3305, opened 2021-12-01
Allow aggregation functions in expression context

mattnibs:
Allow the use of aggregation functions in the expression context. These
so called "streaming aggregations" will output the current result of the
stateful aggregation each time a new value is added to it.

Part of #3018

### DIFF
--- a/compiler/kernel/expr.go
+++ b/compiler/kernel/expr.go
@@ -100,6 +100,12 @@ func compileExpr(zctx *zed.Context, scope *Scope, e dag.Expr) (expr.Evaluator, e
 		return compileSetExpr(zctx, scope, e)
 	case *dag.MapExpr:
 		return compileMapExpr(zctx, scope, e)
+	case *dag.Agg:
+		agg, err := compileAgg(zctx, scope, e)
+		if err != nil {
+			return nil, err
+		}
+		return expr.NewAggregatorExpr(agg), nil
 	default:
 		return nil, fmt.Errorf("invalid expression type %T", e)
 	}
--- a/compiler/parser/parser.peg
+++ b/compiler/parser/parser.peg
@@ -171,7 +171,8 @@ TBD
                 MAP("kind": "Call", "name": "map",
                     "args": ARRAY(MAP("kind": "BinaryExpr", "op": "=",
                                       "lhs": MAP("kind": "ID", "name": "$"),
-                                      "rhs": v))))))))
+                                      "rhs": v)),
+                    "where": NULL))))))
     }
   / match:PatternMatch !ExprGuard { RETURN(match) }
   / v:SearchValue _ InToken _ "*" {
@@ -183,7 +184,8 @@ TBD
                 MAP("kind": "Call", "name": "map",
                     "args": ARRAY(MAP("kind": "BinaryExpr", "op": "in",
                                       "rhs": MAP("kind": "ID", "name": "$"),
-                                      "lhs": v))))))))
+                                      "lhs": v)),
+                    "where": NULL))))))
     }
 
 SearchExpr
@@ -445,7 +447,8 @@ SampleProc
           "keys": ARRAY(MAP("kind": "Assignment",
                    "lhs": MAP("kind": "ID", "name": "shape"),
                    "rhs": MAP("kind": "Call", "name": "typeof",
-                              "args": ARRAY(e)))),
+                              "args": ARRAY(e),
+                              "where": NULL))),
           "aggs": ARRAY(MAP("kind": "Assignment",
                               "lhs": MAP("kind": "ID", "name": "sample"),
                               "rhs": MAP("kind": "Agg",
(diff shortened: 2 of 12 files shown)


mattnibs:
Hold off on review this, I have to add one more feature.

compiler/kernel/expr.go
```diff
@@ -100,11 +100,32 @@ func compileExpr(zctx *zed.Context, scope *Scope, e dag.Expr) (expr.Evaluator, e
...
 	case *dag.MapExpr:
 		return compileMapExpr(zctx, scope, e)
+	case *dag.Agg:
+		return compileAggExpr(zctx, scope, e)
 	default:
 		return nil, fmt.Errorf("invalid expression type %T", e)
 	}
 }
 
+func compileAggExpr(zctx *zed.Context, scope *Scope, e *dag.Agg) (expr.Evaluator, error) {
```
nwt, line 110: Nits: Call this compileAgg since it compiles a dag.Agg, not a dag.AggExpr. And rename e to a for a similar reason.

compiler/parser/parser.peg
```diff
@@ -443,7 +443,8 @@ SampleProc
           "keys": ARRAY(MAP("kind": "Assignment",
                    "lhs": MAP("kind": "ID", "name": "shape"),
                    "rhs": MAP("kind": "Call", "name": "typeof",
-                              "args": ARRAY(e)))),
+                              "args": ARRAY(e),
+                              "where": NULL))),
```
nwt, line 447: Should we add `"where": NULL` to all of the `"kind": "Call"` maps? (Honest question because I forget what our philosophy is here regarding NULL map values.)

mattnibs: Good point let me double check, though I'm pretty sure I wouldn't have added this if this wasn't causing something to fail.

mattnibs: This was needed because this was causing our ztests to fail that was comparing the raw results of pigeon to the raw results of ast.Proc. I fixed this by having ast.Proc omit the where clause in ast.Call if where is empty. I think this is a better route.

compiler/parser/parser.peg
```diff
@@ -725,7 +725,10 @@ Cast
     }
 
 Function
-  = !FuncGuard fn:IdentifierName __ "(" __ args:OptionalExprs __ ")" {
+  = !FuncGuard fn:IdentifierName __ "(" __ args:OptionalExprs __ ")" where:WhereClause {
+      RETURN(MAP("kind": "Call", "name": fn, "args": args, "where": where))
+    }
+  / !FuncGuard fn:IdentifierName __ "(" __ args:OptionalExprs __ ")" {
```
nwt, line 731: I think we should be consistent with the way we've handled this elsewhere, which would be to add
```
  = !FuncGuard fn:IdentifierName __ "(" __ args:OptionalExprs __ ")" where:WhereClause? {
      RETURN(MAP("kind": "Call", "name": fn, "args": args, "where": where))
    }
```
here and then add `"where": NULL` to the other `"kind": "Call"` maps.

compiler/ztests/where-on-func.yaml
```diff
@@ -0,0 +1,7 @@
+script: |
+  ! zq 'cut hex := to_hex(this) where this % 2 == 0' -
```
nwt, line 2: You could make this a Zed-style test.

compiler/parser/parser.peg
mattnibs: Ok!

compiler/ztests/where-on-func.yaml
mattnibs: zed-style tests report errors?

nwt: Yes, with `errorRE: ...`.

compiler/parser/parser.peg
mattnibs: Back to the way it was!

expr/agg.go
```diff
@@ -59,3 +59,27 @@ func (a *Aggregator) filter(rec *zed.Value) bool {
 	}
 	return !a.where(rec)
 }
+
+// AggregatorExpr is an expression variant of an Aggregator also referred to as
+// a streaming aggregation. They are stateful like Aggregations in a 'reduce'
+// context, but they return their current state after every input is consumed.
+func NewAggregatorExpr(agg *Aggregator) Evaluator {
```
nwt, line 66: Reword this to describe NewAggregatorExpr or else s/AggregatorExpr/aggregatorExpr/ and then move the comment to that type.

compiler/parser/parser.peg
```diff
@@ -445,7 +445,8 @@ SampleProc
           "keys": ARRAY(MAP("kind": "Assignment",
                    "lhs": MAP("kind": "ID", "name": "shape"),
                    "rhs": MAP("kind": "Call", "name": "typeof",
-                              "args": ARRAY(e)))),
+                              "args": ARRAY(e),
+                              "where": NULL))),
```
nwt, line 449: Maybe add `"where": NULL` to the `"kind": "Call"` maps up in the `TBD` rule to avoid surprises should someone resurrect that rule?

nwt (approved):

compiler/parser/parser.peg
mattnibs: Why does that `TBD` rule still exist?

PR #3306, opened 2021-12-01
Add record number to seek index

mattnibs:
Add the unique record number to each entry in a lake's data object seek index.

Part of #3018

### DIFF
--- a/lake/seekindex/writer.go
+++ b/lake/seekindex/writer.go
@@ -22,7 +22,7 @@ func NewWriter(w zio.Writer) *Writer {
 	}
 }
 
-func (w *Writer) Write(key zed.Value, offset int64) error {
+func (w *Writer) Write(key zed.Value, count uint64, offset int64) error {
 	b := w.builder
 	b.Reset()
 	if zed.IsContainerType(key.Type) {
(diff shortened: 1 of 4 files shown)


mattnibs:
There's no test for this feature but testing will be added when this feature is actually used in #3018

lake/seekindex/writer.go
```diff
@@ -22,22 +24,31 @@ func NewWriter(w zio.Writer) *Writer {
...
 	}
 	return w.writer.Write(zed.NewValue(w.recType, b.Bytes()))
 }
+
+func (w *Writer) Close() error {
+	if closer, ok := w.writer.(io.Closer); ok {
+		return closer.Close()
+	}
+	return nil
+}
```
nwt, line 54: This doesn't look necessary. It also has foot-shooting potential because it makes it easy to call NewWriter with an io.WriterCloser wrapped by something returning an io.Writer and hiding io.WriterCloser.Close.

lake/seekindex/writer.go
```diff
@@ -22,22 +24,31 @@ func NewWriter(w zio.Writer) *Writer {
...
 		b.AppendContainer(key.Bytes)
 	} else {
 		b.AppendPrimitive(key.Bytes)
 	}
+	b.AppendPrimitive(zed.EncodeUint(number))
 	b.AppendPrimitive(zed.EncodeInt(offset))
 	if w.typ != key.Type {
 		var schema = []zed.Column{
 			{"key", key.Type},
+			{"record", zed.TypeUint64},
```
nwt, line 40: Since we're no longer restricted to indexing records, maybe call this count or index (in the sense of an array index)?

lake/seekindex/writer.go
mattnibs: Fair.

lake/seekindex/writer.go
```diff
@@ -22,18 +22,20 @@ func NewWriter(w zio.Writer) *Writer {
 	}
 }
 
-func (w *Writer) Write(key zed.Value, offset int64) error {
+func (w *Writer) Write(key zed.Value, number uint64, offset int64) error {
```
nwt, line 25: Be consistent with the Zed record column name?
```suggestion
func (w *Writer) Write(key zed.Value, count uint64, offset int64) error {
```

nwt (approved):

PR #3307, opened 2021-12-02
Use seek index to narrow index lookups

### LINKED ISSUE #3018: seek range in index objects

mattnibs:
Store the value count range for each key when creating indexes and use these use
 this range to narrow the scan of an index optimized query.

 This pr also plumbs the SeekIndexStride option into the branch.Load api
 which was necessary in order to get server tests working for this
 feature.

 Closes #3018

### DIFF
--- a/lake/index/expr.go
+++ b/lake/index/expr.go
@@ -7,6 +7,8 @@ import (
 
 	"github.com/brimdata/zed/compiler/ast/dag"
 	astzed "github.com/brimdata/zed/compiler/ast/zed"
+	"github.com/brimdata/zed/expr/extent"
+	"github.com/brimdata/zed/field"
 	"github.com/brimdata/zed/index"
 	"github.com/brimdata/zed/zson"
 	"github.com/segmentio/ksuid"
@@ -15,8 +17,8 @@ import (
 type expr func(context.Context, *Filter, ksuid.KSUID, []Rule) <-chan result
 
 type result struct {
-	err error
-	hit bool
+	err  error
+	span extent.Span
 }
 
 func compileExpr(node dag.Expr) (expr, error) {
@@ -71,9 +73,11 @@ func logicalExpr(lhs, rhs expr, op string) expr {
 func orExpr(c <-chan result) result {
 	var res result
 	for r := range c {
-		res = r
-		if res.hit || res.err != nil {
-			break
+		if r.span != nil {
+			res = appendResult(res, r)
+		}
+		if r.err != nil {
+			return r
 		}
 	}
 	return res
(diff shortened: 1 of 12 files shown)


mattnibs:
This pr is dependent on prs #3305 #3306. Since I can only set one branch as my base, the commit of #3306 is added to this pr but will be removed once #3306 is merged to main.

mattnibs:
I'm going to move the strideIndex plumbing into a separate pr.

lake/index/expr.go
```diff
@@ -112,15 +126,19 @@ func compareExpr(kv index.KeyValue, op string) (expr, error) {
...
 	for _, rule := range rules {
 		// XXX support indexes with multiple keys #3162
 		// and other rule types.
 		if fr, ok := rule.(*FieldRule); ok && in.Key.Equal(fr.Fields[0]) {
-			return rule
+			out := index.KeyValue{
+				Key:   append(field.New("key"), in.Key...),
+				Value: in.Value,
+			}
+			return out, rule
```
nwt, line 138: Unless you find this clearer with `out`, of course.
```suggestion
			return index.KeyValue{
				Key:   append(field.New("key"), in.Key...),
				Value: in.Value,
			}, rule
```

lake/index/filter.go
```diff
@@ -2,15 +2,31 @@ package index
...
 )
 
+var (
+	minExpr = zedexpr.NewDotExpr(field.Dotted("seek.min"))
+	maxExpr = zedexpr.NewDotExpr(field.Dotted("seek.max"))
+	MaxSpan = extent.NewGenericFromOrder(
+		zed.NewUint64(0),
+		zed.NewUint64(math.MaxUint64),
+		order.Asc,
+	)
```
nwt, line 27: Nit: We don't usually wrap after each argument.
```suggestion
	MaxSpan = extent.NewGenericFromOrder(zed.NewUint64(0), zed.NewUint64(math.MaxUint64), order.Asc)
```

lake/index/filter.go
```diff
@@ -31,27 +47,33 @@ func NewFilter(engine storage.Engine, path *storage.URI, dag *dag.Filter) (*Filt
...
 	u := ObjectPath(f.path, rid, oid)
 	finder, err := index.NewFinder(ctx, zed.NewContext(), f.engine, u)
 	if err != nil {
-		return false, err
+		return nil, err
 	}
-	rec, err := finder.Nearest(op, kv)
-	if err != nil {
-		return false, err
+	zv, err := finder.Nearest(op, kv)
```
nwt, line 65: Maybe change `zv` to `val` here and below? (We can always do that later, too.)

lake/seekindex/range.go
```diff
@@ -17,6 +17,20 @@ func (r Range) Size() int64 {
 	return r.End - r.Start
 }
 
+func (r Range) Crop(r2 Range) Range {
```
nwt, line 20: Maybe put Crop next to Trim since they're similar?

lake/ztests/index/query-with-seek.yaml
```diff
@@ -0,0 +1,30 @@
...
+  zed create -q -seekstride 1KiB -orderby dname:asc test
+  zed use -q test@main
+  zed load -q  testscores.zson
+  zed query -s -z 'AvgScrMath == 306 | cut dname'
+  echo === | tee >(cat >&2)
+  zed index create -q AvgScrMath field AvgScrMath
+  zed index update -q AvgScrMath
+  zed query -s -z 'AvgScrMath == 306 | cut dname'
+
+
```
nwt, line 13:
```suggestion

```

service/ztests/index/query-with-seek.yaml
```diff
@@ -0,0 +1,31 @@
...
+  zed create -q -seekstride 1KiB -orderby dname:asc test
+  zed use -q test@main
+  zed load -q  testscores.zson
+  zed query -s -z 'AvgScrMath == 306 | cut dname'
+  echo === | tee >(cat >&2)
+  zed index create -q AvgScrMath field AvgScrMath
+  zed index update -q AvgScrMath
+  zed query -s -z 'AvgScrMath == 306 | cut dname'
+
+
```
nwt, line 12:
```suggestion

```

nwt (approved):

PR #3314, opened 2021-12-08
Migrate all zed lake subcommands to zed

mattnibs:
Make all the subcommands of zed lake, subcommands of zed and get rid of
`zed lake`. The pr also unifies zed api with zed lake- Zed lake services
can be operated on by setting the -lake argument to the URI of a Zed lake
service.

Additional changes:
- zed index has been renamed to indexfile and moved underneath the zed
  dev command
- zst and compile have also been moved under zed dev.


cmd/zed/main.go
```diff
@@ -17,37 +21,57 @@ import (
...
+	wrapnew := spec.New
+	spec.New = func(parent charm.Command, fs *flag.FlagSet) (charm.Command, error) {
+		lk, err := lake.New(parent.(*root.Command), fs)
+		if err != nil {
+			return nil, err
+		}
+		return wrapnew(lk, fs)
+	}
+	root.Zed.Add(spec)
+}
```
nwt, line 77: This and `lake.New` don't feel quite right.

How about adding something like `type LakeFlag { Lake string }` to package `cli` (I know we already have `cli/lakeflags`, but we can rename that later) to handle `-lake` and giving it `Connection` and `Open` methods (to take the place of the corresponding methods in `cmd/zed/lake/command.go`)?

mattnibs: What command would use `type LakeFlag`, root.Command?

mattnibs: Should `LakeFlag` be present in the `dev` sub-commands?

nwt: No, `LakeFlag` would go into any command that accesses a lake (i.e., the same commands passed to `lakeCmd` here).

mattnibs: Oh I see. I kind of like that!

mattnibs: Would need to also add ConfigDir to LakeFlag (along with the credential access methods) but this simplifies things nicely!

lake/api/api.go
```diff
@@ -36,6 +37,10 @@ type Interface interface {
 	UpdateIndex(ctx context.Context, names []string, pool ksuid.KSUID, branchName string) (ksuid.KSUID, error)
 }
 
+func IsRemoteLake(u *storage.URI) bool {
```
nwt, line 40: I don't think remote is the right term here since a lake on S3 is also remote in a sense. Maybe IsAPILake, IsHTTPLake, or IsRESTLake?

service/core.go
```diff
@@ -36,7 +36,7 @@ const indexPage = `
     <h2>zed lake serve</h2>
     <p>A <a href="https://github.com/brimdata/zed/tree/main/cmd/zed/lake/serve">zed lake service</a> is listening on this host/port.</p>
     <p>If you're a <a href="https://www.brimdata.io/">Brim</a> user, connect to this host/port from the <a href="https://github.com/brimdata/brim">Brim application</a> in the graphical desktop interface in your operating system (not a web browser).</p>
-    <p>If your goal is to perform command line operations against this Zed lake, use the <a href="https://github.com/brimdata/zed/tree/main/cmd/zapi">zapi</a> client.</p>
+    <p>If your goal is to perform command line operations against this Zed lake, use the <a href="https://github.com/brimdata/zed/tree/main/cmd/zed">Zed</a> client.</p>
```
nwt, line 39:
```suggestion
    <p>If your goal is to perform command line operations against this Zed lake, use the <a href="https://github.com/brimdata/zed/tree/main/cmd/zed"><code>zed</code></a> command.</p>
```

README.md
```diff
@@ -53,18 +53,18 @@ Once installed, you can run the query engine from the command-line using `zq`:
 ```
 echo '{"s":"hello, world"}' | zq -Z -
 ```
-Or you can run a Zed lake server, load it with data using `zapi`, and hit the API.
+Or you can run a Zed lake service, load it with data using `zed load`, and hit the API.
 In one shell, run the server:
 ```
 mkdir scratch
-zed lake serve -R scratch
+zed serve -R scratch
```
nwt, line 60:
```suggestion
zed serve -lake scratch
```

cli/lake.go
```diff
@@ -0,0 +1,145 @@
...
+			return nil, err
+		}
+		return api.NewRemoteWithConnection(conn), nil
+	}
+	return api.OpenLocalLake(ctx, uri)
+}
+
+func (l *LakeFlags) URI() (*storage.URI, error) {
+	if l.Lake == "" {
+		return nil, errors.New("lake URI must be set (either with the -lake flag or ZED_LAKE environment variable)")
```
nwt, line 70:
```suggestion
		return nil, errors.New("lake location must be set (either with the -lake flag or ZED_LAKE environment variable)")
```

cli/lake.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+	u, err := storage.ParseURI(l.Lake)
+	if err != nil {
+		err = fmt.Errorf("error parsing lake path: %w", err)
```
nwt, line 74:
```suggestion
		err = fmt.Errorf("error parsing lake location: %w", err)
```

cmd/zed/ztests/no-lake-root.yaml
```diff
@@ -1,7 +1,7 @@
 script: |
-  ! zed lake ls
+  ! zed ls
```
nwt, line 2:
```suggestion
  ! zed ls -lake ''
```
Because the default lake should be http://localhost:9867.

lake/api/api.go
mattnibs: Right. Perhaps sine we call it service everywhere `IsLakeService`?

nwt: That's perfect!

cmd/zed/ztests/no-lake-root.yaml
mattnibs: Didn't know this was the case. Ok.

cli/commitflags.go
```diff
@@ -1,4 +1,4 @@
-package lake
+package cli
```
nwt, line 1: Maybe change ZED_LAKE_USER to ZED_USER below? And there's a "zed lake" in the CommitFlags comment.

service/core.go
```diff
@@ -36,7 +36,7 @@ const indexPage = `
     <h2>zed lake serve</h2>
```
nwt, line 36: Update this and the title.

nwt (commented):
cmd/zed/init/command.go, cmd/zed/load/command.go, cmd/zed/ls/command.go, 
and cmd/zed/query/command.go all still contain "zed lake <command>".

nwt (approved):
:shipit:

PR #3321, opened 2021-12-11
allow cut of non-record value to root

nwt:
Remove the restriction on cutting a non-record value to root left over
from when a top-level value was required to be a record.


nwt:
Talked about this with @mccanne and it isn't the way.

PR #3323, opened 2021-12-13
change "traverse" to "over", simplify, and implement double-EOS protocol

mccanne:
How about we merge this into the "traverse" branch, then continue to edit that PR...?


docs/language/operators.md
```diff
@@ -21,7 +21,7 @@ The following available operators are documented in detail below:
 * [`rename`](#rename)
 * [`sort`](#sort)
 * [`tail`](#tail)
-* [`traverse`](#traverse)
+* [`over`](#over)
```
nwt, line 24: NIt: Keep these sorted.

docs/language/operators.md
```diff
@@ -990,35 +990,39 @@ zq -z 'County=="Los Angeles" | tail 5' schools.zson
 
 ---
 
-## `traverse`
+## `over`
```
nwt, line 993: Nit: Keep these sorted.

docs/language/operators.md
```diff
@@ -990,35 +990,39 @@ zq -z 'County=="Los Angeles" | tail 5' schools.zson
...
 ```
 
-`traverse` emits each element in collection (i.e., values of type `array`,
-`set`, `record`\*, `map`\*) `foo` as a separate value to child
-nodes in the flowgraph. If the optional sequence `seq`\*\* argument is provided the
-elements of `foo` are read into `seq` as a separate stream sending an `EOS` when
-the last element is traversed. The each element can be accessed inside `seq` by
-referencing the `this` root record.
+For each input value,
+`over` yields a sequence of output values terminated by `EOS`
```
nwt, line 1000: EOS is an implementation detail that does not appear anywhere else in the documentation. I don't think we should confuse the reader with it.

docs/language/operators.md
```diff
@@ -990,35 +990,39 @@ zq -z 'County=="Los Angeles" | tail 5' schools.zson
...
+where the values are determined by the one or more expressions
+evaluated left to right.
+When an expression results in a complex type, each of the composite value's top-level
+elements is yielded:
+* for an array or set, each element is yielded;
+* for a record, each field value is yielded as a single-field record; and
+* for a map, each entry is yielded as a record containing two fields
+`key` and `value`.
```
nwt, line 1008:
```suggestion
When an expression results in a complex type, `over` yields each of the composite value's top-level elements:
* for an array or set, it yields each element;
* for a record, it yields each field value as a single-field record; and
* for a map, it yields each entry as a record containing two fields, `key` and `value`.
```

proc/traverse/ztests/filter.yaml
```diff
@@ -1,4 +1,4 @@
-zed: "traverse a => (this % 2 == 0)"
+zed: "over a | this % 2 == 0"
```
nwt, line 1:
```suggestion
zed: over a | this % 2 == 0
```

proc/traverse/ztests/bare.yaml
```diff
@@ -1,8 +1,9 @@
-zed: "traverse a | sum(this)"
+zed: "over a | sum(this)"
```
nwt, line 1:
```suggestion
zed: over a | sum(this)
```

proc/traverse/over.go
```diff
@@ -0,0 +1,97 @@
+package traverse
```
nwt, line 1: Do you want to change the package name and directory name from travers to over?

proc/traverse/over.go
```diff
@@ -0,0 +1,100 @@
...
+	if o.eof {
+		o.eof = false
+		return nil, nil
+	}
+	o.eof = true
+	out, err := o.over(&o.vals[0])
+	o.vals = o.vals[1:]
+	if len(o.vals) == 0 {
+		o.batch.Unref()
+	}
```
nwt, line 45: This is unsafe since `out.Bytes` might refer to `o.vals[0].Bytes`.
```suggestion
```

proc/groupby/groupby.go
```diff
@@ -172,22 +172,29 @@ func (p *Proc) Done() {
...
 					return
 				}
-				p.sendResult(b, err)
+				if b == nil {
+					break
+				}
+				p.sendResult(b, nil)
 			}
+			eof = true
+			continue
```
nwt, line 195: Might be clearer to label the outer loop as PullLoop and do this.
```suggestion
				if b == nil {
					eof = true
					continue PullLoop
				}
				p.sendResult(b, nil)
			}
```

proc/traverse/over.go
```diff
@@ -0,0 +1,100 @@
...
+func appendOver(out *zbuf.Array, zv *zed.Value) error {
+	if zed.IsPrimitiveType(zv.Type) {
+		out.Append(zv)
+		return nil
+	}
+	typ := zed.InnerType(zv.Type)
+	if typ == nil {
+		// XXX Support records and maps.
+		// we shouldn't error on this...
+		//zerr := zed.NewErrorf("value must be of type array or set, got: %s", zv.Type)
```
mattnibs, line 78:
```suggestion
```

docs/language/operators.md
mccanne: I agreed and copied what nibs had.  We need to document batching.  Let's leave this here for now since we are redoing a lot in the next step.

mattnibs (approved):

proc/traverse/over.go
mccanne: No, because over/into both belong in traverse.  This will get rejiggered in the next step.

proc/traverse/over.go
mccanne: There's a copy for this reason.  Better to send it to GC?

We should discuss this protocol.  I think a much better buffering model that isn't the full blown "mbuf" proposal from long ago shouldn't be too hard.

proc/groupby/groupby.go
mccanne: That doesn't work because of nested loops.  It would be cleaner to put that loop in another method.

proc/traverse/over.go
nwt: I missed the copy, so this is indeed fine.

nwt (approved):

proc/groupby/groupby.go
nwt: It works if you label the outer loop.

proc/traverse/over.go
mccanne: I created issue #3324

PR #3326, opened 2021-12-14
add yield operator and drop "this" as an lval

mccanne:
This commit adds the "yield" operator which takes one or more
expressions and produces the values as "this" on the output.
Put and cut are now exclusively record operators and no longer
allow "this" as an lvalue (since yield is more natural for this).


docs/language/operators.md
```diff
@@ -1094,3 +1095,46 @@ zq -z 'cut District,County | uniq' schools.zson
...
+| **Optional arguments**    | None |
+
+#### Example #1:
+
+This example produce two simpler records for every school record listing
+the average math score with the school name and the county name.
+
+```mdtest-command dir=testdata/edu
+zq -Z 'AvgScrMath!=null | yield {school:sname,avg:AvgScrMath}, {county:cname,zvg:AvgScrMath}' testscores.zson
+```
```
mattnibs, line 1117: I know this example is just following the convention established in this file, but I personally find it easier to read examples when I can see the input data, for example:

```
echo '{x:1}' | zq -Z '[the query]' -
```

docs/language/operators.md
```diff
@@ -1094,3 +1095,46 @@ zq -z 'cut District,County | uniq' schools.zson
...
+---
+
+## `yield`
+
+|                           |                                                 |
+| ------------------------- | ----------------------------------------------- |
+| **Description**           | For each input value, produces one or more values downstream.
+| **Syntax**                | `yield <expression> [, (<expression>)...]` |
+| **Required arguments**    | One or more of `<expression>`, which may be any valid Zed [expression](expressions.md).
+| **Optional arguments**    | None |
```
mattnibs, line 1108: I know this is following the convention established in this file, but I'd prefer following the way (over is documented)[https://github.com/brimdata/zed/blob/main/docs/language/operators.md#over]. The tables look nice when the markdown is rendered but the markdown itself can become unwieldy to edit. See the table in the [join section](https://github.com/brimdata/zed/blob/main/docs/language/operators.md#join).

mattnibs (commented):
Great feature! I think you need to update zfmt to handle yield, but otherwise this looks good to go.

docs/language/operators.md
mattnibs: I am planning at some point to reformat these to get rid of the table.

docs/language/operators.md
mccanne: Agreed.  Let's do this in another docs overhaul.  The schools data should be in a tutorial and the reference docs should just have simple, generic examples like `{x:1}` suggested here.

docs/language/operators.md
mccanne: I was following convention and didn't realize there is precedent for not following it.  Let's do the change in the subsequent PR as suggested above.

mattnibs (approved):

expr/ztests/cut-to-root-error-unset-value.yaml
```diff
@@ -3,6 +3,4 @@ zed: 'cut this:=a'
 input: |
   {a:null({b:int32})}
 
-warnings: |
-  cut: cannot cut an unset value to .
-  cut: no record found with columns a
+errorRE: "cut: 'this' not allowed \\(use record literal\\)"
```
nwt, line 6: I think we should just keep one of cut-to-root-error-non-record.yaml and cut-to-root-error-unset-value.yaml, and we should rename it to cut-to-root-error.yaml.

nwt (approved):

PR #3334, opened 2021-12-16
fix join when first left key value is non-null but zero-length

nwt:
The join operator ignores the first value encountered for the left join
key if its zed.Value.Bytes is non-null but zero-length.  Fix this by
preserving the nil-ness of that field when setting join.Proc.joinKey.

### DIFF
--- a/value.go
+++ b/value.go
@@ -154,6 +154,8 @@ func (v Value) IsUnsetOrNil() bool {
 	return v.Bytes == nil
 }
 
+// Copy returns a copy of v that does not share v.Bytes.  The copy's Bytes field
+// is nil if and only if v.Bytes is nil.
 func (v Value) Copy() *Value {
 	var b zcode.Bytes
 	if v.Bytes != nil {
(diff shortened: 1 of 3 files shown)


value.go
```diff
@@ -163,6 +165,20 @@ func (v Value) Copy() *Value {
...
+	if from.Bytes == nil {
+		v.Bytes = nil
+	} else if v.Bytes == nil {
+		v.Bytes = make(zcode.Bytes, len(from.Bytes))
+		copy(v.Bytes, from.Bytes)
+	} else {
+		v.Bytes = append(v.Bytes[:0], from.Bytes...)
+	}
+}
+
```
nwt, line 181: I'm confident we want something like this, but I'm not sure I got the interface quite right.

mccanne: That looks good for now but I think we want to package all this up inside of batch for most use cases so we can efficiently clone  values that share byte slices where they all live inside the same batch.

mccanne (approved):

PR #3335, opened 2021-12-16
load -seekStride: Don't set global var

mattnibs:
In the zed load command setting the global SeekIndexStride variable to the
value of the -seekStride flag made it so this value was ignored for
service load requests. Instead add seekStride option to the lake Load
command and propagate this value to the object writers. Add seekStride
query parameter for load requests.

### DIFF
--- a/api/client/connection.go
+++ b/api/client/connection.go
@@ -271,8 +271,9 @@ func (c *Connection) Query(ctx context.Context, head *lakeparse.Commitish, src s
 	return res, err
 }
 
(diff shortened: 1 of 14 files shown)


nwt:
This feels like a lot of wiring for a knob that won't get much use. Should we maybe make it part of a lake's or pool's configuration?

mattnibs:
> This feels like a lot of wiring for a knob that won't get much use. Should we maybe make it part of a lake's or pool's configuration?

Let's discuss

mattnibs:
Closing and bringing this up as hidden pool config option.

PR #3337, opened 2021-12-16
Add SeekStride to pool configuration

mattnibs:
Instead of having SeekStride as a global variable, move this to an
option in pools.Config. Drop -seekstride as a flag in zed command load
and add -seekstride as a hidden flag in zed create.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -36,9 +36,10 @@ type VersionResponse struct {
 }
 
 type PoolPostRequest struct {
-	Name   string       `json:"name"`
-	Layout order.Layout `json:"layout"`
-	Thresh int64        `json:"thresh"`
+	Name       string       `json:"name"`
+	Layout     order.Layout `json:"layout"`
+	Thresh     int64        `json:"thresh"`
+	SeekStride int          `json:"seek_stride"`
 }
 
 type PoolPutRequest struct {
--- a/api/api.go
+++ b/api/api.go
@@ -38,8 +38,8 @@ type VersionResponse struct {
 type PoolPostRequest struct {
 	Name       string       `json:"name"`
 	Layout     order.Layout `json:"layout"`
-	Thresh     int64        `json:"thresh"`
 	SeekStride int          `json:"seek_stride"`
+	Thresh     int64        `json:"thresh"`
 }
 
 type PoolPutRequest struct {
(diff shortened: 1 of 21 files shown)


lake/data/object.go
```diff
@@ -14,7 +14,10 @@ import (
 	"github.com/segmentio/ksuid"
 )
 
-const DefaultThreshold = 500 * 1024 * 1024
+const (
+	DefaultThreshold  = 500 * 1024 * 1024
+	DefaultSeekStride = 64 * 1024
```
nwt, line 19:
```suggestion
	DefaultSeekStride = 64 * 1024
	DefaultThreshold  = 500 * 1024 * 1024
```

api/api.go
```diff
@@ -36,9 +36,10 @@ type VersionResponse struct {
 }
 
 type PoolPostRequest struct {
-	Name   string       `json:"name"`
-	Layout order.Layout `json:"layout"`
-	Thresh int64        `json:"thresh"`
+	Name       string       `json:"name"`
+	Layout     order.Layout `json:"layout"`
+	Thresh     int64        `json:"thresh"`
+	SeekStride int          `json:"seek_stride"`
```
nwt, line 42:
```suggestion
	SeekStride int          `json:"seek_stride"`
	Thresh     int64        `json:"thresh"`
```

lake/api/remote.go
```diff
@@ -47,11 +47,12 @@ func (r *RemoteSession) CommitObject(ctx context.Context, poolID ksuid.KSUID, br
 	return res.Commit, err
 }
 
-func (r *RemoteSession) CreatePool(ctx context.Context, name string, layout order.Layout, thresh int64) (ksuid.KSUID, error) {
+func (r *RemoteSession) CreatePool(ctx context.Context, name string, layout order.Layout, thresh int64, seekStride int) (ksuid.KSUID, error) {
```
nwt, line 50: Maybe keep the order of these consistent with that of the api.PoolPostRequest fields?
```suggestion
func (r *RemoteSession) CreatePool(ctx context.Context, name string, layout order.Layout, seekStride int, thresh int64) (ksuid.KSUID, error) {
```

nwt (approved):

lake/data/object.go
mattnibs: Doh!

PR #3342, opened 2021-12-17
add jq-like "." shortcut for yield

mccanne:
This commit allows a simple jq-like shortcut where "." is the same
as "pass" and ".x" is the same as yield x.  You can use any deref
operators after the ".".

We also removed expression support for leading "." to mean "this".

### DIFF
--- a/compiler/parser/parser.peg
+++ b/compiler/parser/parser.peg
@@ -579,8 +579,19 @@ OverProc
     }
 
 YieldProc
-  = "yield"i _ exprs:Exprs {
+  = "yield"i _  exprs:Exprs {
 	  RETURN(MAP("kind":"Yield", "exprs":exprs))
+    }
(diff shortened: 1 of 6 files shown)


compiler/parser/parser.peg
```diff
@@ -579,9 +579,20 @@ OverProc
     }
 
 YieldProc
-  = "yield"i _ exprs:Exprs {
+  = "yield"i _  exprs:Exprs {
```
mattnibs, line 582:
```suggestion
  = "yield"i _ exprs:Exprs {
```

mattnibs (commented):
If you're going to allow `.` in yield shouldn't one be able to use this wherever `this` is used?  What makes `yield` different?

mccanne:
I don't think we should do this.

PR #3346, opened 2021-12-18
replace deprecated concept of "unset" with null

mccanne:
This commit replaces the deprecated concept of unset with Zed null.
Zed does not have unset record fields as a record with an unset
field is simply a different record type without that field.

We also removed the null check from zed.AppendInt and moved it
into zcode where it belongs.


value.go
```diff
@@ -143,14 +143,8 @@ func (v Value) IsNil() bool {
...
 
-// IsUnset returns true iff v is an unset value.  Unset values are represented
-// with a zero-valued Value.  A zero-valued value that is not unset is represented
-// by a non-nil slice for Bytes of zero length.
-func (v Value) IsUnset() bool {
-	return v.Bytes == nil && v.Type != nil
-}
-
-func (v Value) IsUnsetOrNil() bool {
+// IsNull returns true iff v is any type of null
```
nwt, line 146:
```suggestion
// IsNull returns true if and only if v is a null value of any type.
```

nwt (approved):
There are some names and comments in which we should change "unset" to "null", but we can scrub those later.

nwt (approved):
:shipit:

PR #3349, opened 2021-12-18
remove zio/tzngio

nwt:
Move live code to zio/zeekio and remove dead code.


nwt:
I'd rather replace this stuff than move it.

PR #3352, opened 2021-12-19
remove misguided select expressions

mccanne:
Now that we have properly figured out how to traverse values
into a sub-flowgraph, this commit removes the select expression
and expr.Generator, which was never quite right anyway.


compiler/kernel/bufferfilter.go
```diff
@@ -91,38 +86,7 @@ func isFieldEqualOrIn(e *dag.BinaryExpr) (*astzed.Primitive, string) {
...
-		}
-		if lhs, ok := pred.LHS.(*astzed.Primitive); ok && lhs.Type != "net" {
-			return lhs, pred.Op, true
-		}
-	}
-	return nil, "", false
-}
-
-func isDollar(e dag.Expr) bool {
+func xisDollar(e dag.Expr) bool {
```
nwt, line 89: Delete this too since it's no longer used?

nwt (approved):

PR #3368, opened 2021-12-21
add expr.Context and change Evaluator interface to take context

mccanne:
This commit changes the Eval() interface to expr.Evaluator to
add an "expression context" representing the state of a locally
executing flowgraph.  This includes references to named values
via a forthcoming let/with capability of Zed flowgraphs along
with a means to allocate new Zed values inside of the context
so they may be easily freed to a pool when no longer needed
downstream.  Some scaffolding for let and ast.Scope has been put in
for a subsequent PR to finish these bits.

zbuf.Batch now has a method to carry its expression context.
This is important as flowgraph concurrency requires that batches and
context be carried together.  Since a batch is a bunch of values
and the context allows the creation of new values in the same
batch, this design provides a foundation for more efficient batch
allocations in the future.  (For example, batches could be wrapped
so new values in a new batch could refer to byte slices in the
underlying batch and the top batch would ref the underlying batch.)

Also as part of this change, we reworked the convention in
expr.Evaluator.Eval() to return all Zed runtime erros as Zed
values and only unrecoverable system errors (e.g., i/o errors
or corrupt data errors) as Go errors.  This can cause uncaught
panics to occur from proc goroutines but a soon-to-be-merged PR
will add recover() to each proc goroutine.

Missing values are now generally not ignored in group-bys, cut,
put etc.  If you want to ignore them in these context,
you can wrap the underlying expressions in the new quiet() function,
where quiet turns error("missing") into error("quiet") and quiet
errors *are ignored*.  That says, yield ignores top-levelmissing
values.

We also updated many places in the code to use *zed.Value pointers
instead of zed.Values.  The pick proc has been removed.

proc.Function has been replaced with expr.Applier and various cleanup
made to appliers.


cmd/zed/dev/indexfile/create/command.go
```diff
@@ -113,6 +113,7 @@ func (c *Command) buildTable(zctx *zed.Context, reader zio.Reader) (*index.MemTa
 		return nil, err
 	}
 	table := index.NewMemTable(zctx, c.keys)
+	ctx := expr.NewContext()
```
nwt, line 116: In the same way we use `zctx` as the name for a `zed.Context` to avoid confusion with `context.Context`, I think we should stick with `ectx` as the name for these things.

compiler/semantic/proc.go
```diff
@@ -561,6 +552,22 @@ func semProc(ctx context.Context, scope *Scope, p ast.Proc, adaptor proc.DataAda
 			Kind:  "Over",
 			Exprs: exprs,
 		}, nil
+	case *ast.Scope:
+		locals, err := semAssignments(scope, p.Locals)
+		if err != nil {
+			return nil, err
+		}
+		//XXX need to add locals to scope data structure and
+		// extent it to handle local frame as well as consts, etc
```
nwt, line 561:
```suggestion
		// extend it to handle local frame as well as consts, etc.
```

docs/language/functions.md
```diff
@@ -42,6 +41,7 @@
   - [`nameof`](#nameof)
   - [`typeof`](#typeof)
   - [`typeunder`](#typeunder)
+  - [`quiet`](#quiet)
```
nwt, line 44: Put this between nameof and typeof so the list remains alphabetical (and likewise with the target of this link).

docs/language/functions.md
```diff
@@ -763,6 +741,27 @@ echo  '{flavor:"chocolate"}(=flavor)' \
...
 
+### `quiet`
+
+```
+quiet(a <any>) -> type
+```
+
+`quiet` returns `a` unless `a` is `error("missing")` in which case
+it returns `error("quiet")`.  Quiet errors are ignored by operators
+`yield`, `put`, `cut`, and `summarize`.
```
nwt, line 752:
```suggestion
`cut`, `put`, `summarize`, and `yield`.
```

docs/language/operators.md
```diff
@@ -646,7 +641,7 @@ value. For instance the sum of elements in an array can be computed with
 records and maps. This will be added shortly.
 
 \*\* `over` is currently in beta and works on a subset of the available
-operators including `filter`, `cut` and `pick`, and `summarize`.  Other operators
+operators including `filter`, `cut`, and `summarize`.  Other operators
```
nwt, line 644:
```suggestion
operators including `cut`, `filter`, and `summarize`.  Other operators
```

cmd/zed/dev/indexfile/create/command.go
mccanne: Ugh that's a lot of typing

docs/language/grouping.md
```diff
@@ -103,20 +103,26 @@ Name_Length count
...
 
 #### Output:
-```mdtest-output
+```mdtest-output head
+cname           dnmae   avg                count
+Los Angeles     missing 450.83037974683543 469
+San Diego       missing 496.74789915966386 168
+San Bernardino  missing 465.11764705882354 117
+Riverside       missing 463.8170731707317  110
+Orange          missing 510.91011235955057 107
```
nwt, line 125:
```suggestion
Orange          missing 510.91011235955057 107
...
```

expr/agg/collect.go
```diff
@@ -1,30 +1,31 @@
...
 
-func (c *Collect) update(v zed.Value) {
-	stash := make(zcode.Bytes, len(v.Bytes))
-	copy(stash, v.Bytes)
-	c.values = append(c.values, zed.Value{v.Type, stash})
-	c.size += len(v.Bytes)
+func (c *Collect) update(val *zed.Value) {
+	stash := make(zcode.Bytes, len(val.Bytes))
+	copy(stash, val.Bytes)
+	c.values = append(c.values, zed.Value{val.Type, stash})
```
nwt, line 27:
```suggestion
	c.values = append(c.values, *val.Copy())
```

error.go
```diff
@@ -12,16 +12,18 @@ import (
...
+var quiet = []byte("quiet")
 
 // Missing is value that represents the error condition that a field
 // referenced was not present.  The Missing value can be propagated through
 // functions and expressions and each operator must clearly defined its
 // semantics with respect to the Missing value.  For example, "true AND MISSING"
 // is MISSING.
-var Missing = NewError(ErrMissing)
+var Missing = &Value{TypeError, missing}
+var Quiet = &Value{TypeError, quiet}
```
nwt, line 26: NIt: There's no benefit to having `var missing = ...` and `var quiet = ...` up above.
```suggestion
var Missing = &Value{TypeError, []byte("missing")}
var Quiet = &Value{TypeError, []byte("quiet")}
```

expr/agg/fuse.go
```diff
@@ -11,34 +11,35 @@ type fuse struct {
...
+	}
 }
 
-func (f *fuse) Consume(v zed.Value) error {
+func (f *fuse) Consume(val *zed.Value) {
 	// only works for record types, e.g., fuse(foo.x) where foo.x is a record
-	typ, ok := v.Type.(*zed.TypeRecord)
-	if !ok {
-		return nil
+	if typ, ok := val.Type.(*zed.TypeRecord); ok {
```
nwt, line 24:
```suggestion
	if typ := zed.TypeRecordOf(val.Type); typ != nil {
```

expr/applier.go
```diff
@@ -0,0 +1,11 @@
+package expr
+
+import (
+	"fmt"
+)
+
+type Applier interface {
```
nwt, line 7: I think this should stay in proc/apply.go since that's the only place it's used.

expr/context.go
```diff
@@ -0,0 +1,52 @@
+package expr
+
+import (
+	"github.com/brimdata/zed"
+	"github.com/brimdata/zed/zcode"
+)
+
+// Context is an interface to a scope and value allocator for expressions.
+// This allows external packages to implement efficint temporary buffering
```
nwt, line 9:
```suggestion
// This allows external packages to implement efficient temporary buffering
```

expr/context.go
```diff
@@ -0,0 +1,52 @@
...
+// of Zed values both for let-style temporary variables accessible via
+// Context.Scope() and for allocating results.
```
nwt, line 11:
```suggestion
// the scope and for allocating results.
```

expr/eval.go
```diff
@@ -11,34 +10,29 @@ import (
...
 }
 
-func (n *Not) Eval(rec *zed.Value) (zed.Value, error) {
-	zv, err := evalBool(n.expr, rec)
-	if err != nil {
-		return zv, err
-	}
+func (n *Not) Eval(ectx Context, val *zed.Value) *zed.Value {
+	zv := evalBool(ectx, n.expr, val)
 	if zed.IsTrue(zv.Bytes) {
```
nwt, line 32:
```suggestion
func (n *Not) Eval(ectx Context, this *zed.Value) *zed.Value {
	val := evalBool(ectx, n.expr, this)
	if zed.IsTrue(val.Bytes) {
```

expr/eval.go
```diff
@@ -219,27 +196,32 @@ func NewCompareEquality(lhs, rhs Evaluator, operator string) (*Equal, error) {
 	return e, nil
 }
 
-func (e *Equal) Eval(rec *zed.Value) (zed.Value, error) {
-	_, err := e.numeric.eval(rec)
+func (e *Equal) Eval(ectx Context, this *zed.Value) *zed.Value {
+	_, err := e.numeric.eval(ectx, this)
 	if err != nil {
```
nwt, line 201:
```suggestion
	if _, err := e.numeric.eval(ectx, this); err != nil {
```

expr/eval.go
```diff
@@ -279,32 +255,62 @@ func newNumeric(lhs, rhs Evaluator) numeric {
 	}
 }
 
-func enumify(v zed.Value) (zed.Value, error) {
+func enumify(v *zed.Value) *zed.Value {
```
nwt, line 258:
```suggestion
func enumify(val *zed.Value) *zed.Value {
```

expr/extent/span.go
```diff
@@ -66,48 +66,50 @@ func NewGenericFromOrder(first, last zed.Value, o order.Which) *Generic {
...
-	return g.cmp(g.first, g.last) <= 0
+	return g.cmp(&g.first, &g.last) <= 0
 }
 
-func (g *Generic) Extend(zv zed.Value) {
-	if g.cmp(zv, g.first) < 0 {
-		g.first = zv
-	} else if g.cmp(zv, g.last) > 0 {
-		g.last = zv
+//XXX don't we have to copy the bytes from the zed.Value here?
```
nwt, line 106: Are you looking for zed.Value.Copy?

expr/filter.go
```diff
@@ -11,25 +11,40 @@ import (
 	"github.com/brimdata/zed/zcode"
 )
 
-type Filter func(*zed.Value) bool
+type Filter func(Context, *zed.Value) bool
 
 func LogicalAnd(left, right Filter) Filter {
-	return func(p *zed.Value) bool { return left(p) && right(p) }
+	return func(ectx Context, val *zed.Value) bool {
```
nwt, line 17: Do we want to call the zed.Value parameter `this` in Filter implementation as we do in Eval implementation?

expr/function/bytes.go
```diff
@@ -11,64 +11,61 @@ import (
 // https://github.com/brimdata/zed/blob/main/docs/language/functions.md#from_base64
 type FromBase64 struct{}
 
-func (*FromBase64) Call(args []zed.Value) (zed.Value, error) {
+func (f *FromBase64) Call(ctx zed.Allocator, args []zed.Value) *zed.Value {
```
nwt, line 14: I don't love the ctx name here but let's hash that out after we merge this.

expr/function/function.go
```diff
@@ -231,7 +222,60 @@ func (i *Is) Call(args []zed.Value) (zed.Value, error) {
...
+func newString(ctx zed.Allocator, native string) *zed.Value {
+	return ctx.NewValue(zed.TypeString, zed.EncodeString(native))
+}
+
+func newError(ctx zed.Allocator, err error) *zed.Value {
+	return ctx.CopyValue(zed.NewError(err))
+}
+
+func newErrorf(ctx zed.Allocator, format string, args ...interface{}) *zed.Value {
+	return ctx.CopyValue(zed.NewErrorf(format, args...))
```
nwt, line 280: Copying these is unnecessary since they their bytes.

zbuf/array.go
```diff
@@ -38,13 +54,27 @@ func (a *Array) Write(r *zed.Value) error {
...
+}
+
+func (*Array) NewValue(typ zed.Type, bytes zcode.Bytes) *zed.Value {
+	// XXX can make this more efficient later
+	return zed.NewValue(typ, bytes)
+}
+
+func (*Array) CopyValue(val zed.Value) *zed.Value {
+	// XXX can make this more efficient later
+	return zed.NewValue(val.Type, val.Bytes)
```
nwt, line 79: If the CopyValue contract is that it returns a value that does not share val.Bytes, then we need an actual copy here.
```suggestion
	return val.Copy()
```

zio/zngio/batch.go
```diff
@@ -33,6 +35,15 @@ func (b *batch) add(r *zed.Value) { b.vals = append(b.vals, *r) }
 
 func (b *batch) Values() []zed.Value { return b.vals }
 
+//XXX
+func (b *batch) NewValue(typ zed.Type, bytes zcode.Bytes) *zed.Value {
+	return zed.NewValue(typ, bytes)
+}
+
+func (b *batch) CopyValue(val zed.Value) *zed.Value {
+	return zed.NewValue(val.Type, val.Bytes)
```
nwt, line 44: If the CopyValue contract is that it returns a value that does not share val.Bytes, then we need an actual copy here.
```suggestion
	return val.Copy()
```

expr/context.go
```diff
@@ -0,0 +1,52 @@
...
+func NewContext() *allocator {
+	return &allocator{}
+}
+
+func (*allocator) NewValue(typ zed.Type, bytes zcode.Bytes) *zed.Value {
+	return zed.NewValue(typ, bytes)
+}
+
+func (*allocator) CopyValue(val zed.Value) *zed.Value {
+	return zed.NewValue(val.Type, val.Bytes)
```
nwt, line 33: If the CopyValue contract is that it returns a value that does not share val.Bytes, then we need an actual copy here.
```suggestion
	return val.Copy()
```

expr/agg/collect.go
mccanne: This sends the pointer to GC

error.go
mccanne: It's for IsMissing and IsQuiet with one place to define the message.

expr/applier.go
mccanne: as we discussed

expr/eval.go
mccanne: also fixed bug where missing was hidden

expr/agg/collect.go
nwt: The compiler's smart enough to put it on the stack.

expr/applier.go
mccanne: When we get rid of warnings, we won't need the interface anymore anyway!  It will just be expr.Eval.

expr/function/function.go
mccanne: But it returns a zed.Value not *zed.Value

zbuf/array.go
mccanne: This was consistent with current behavior.  I think we should fix it when we change the caller's to presume fate share between batch refcnt and the result from Eval().  Right now, they copy the value so no sense copying multiple times.  The other change needed is for the batch NewValue() alllocator to allocate zcode.Bytes inside of the batch memory pools.

zio/zngio/batch.go
mccanne: same as above

expr/context.go
mccanne: same as above

docs/language/operators.md
```diff
@@ -17,7 +17,6 @@ The following available operators are documented in detail below:
 * [`head`](#head)
 * [`join`](#join)
 * [`over`](#over)
-* [`pick`](#pick)
```
mattnibs, line 20: How about a doc for `let`?

expr/function/parse.go
```diff
@@ -62,7 +57,7 @@ func (p *ParseURI) Call(args []zed.Value) (zed.Value, error) {
 	if ss := u.Port(); ss != "" {
 		u64, err := strconv.ParseUint(ss, 10, 16)
 		if err != nil {
-			return zed.Value{}, fmt.Errorf("parse_uri: %q: invalid port: %s", s, errors.Unwrap(err))
+			return newErrorf(ctx, "parse_uri: %q: invalid port: %s", s, ss)
```
nwt, line 60: You should update this to follow your new `%s (%q)` convention.

expr/agg/collect.go
```diff
@@ -92,29 +93,30 @@ func (c *Collect) buildUnion(zctx *zed.Context, selectors map[zed.Type]int) (zed
 	}
 	unionType := zctx.LookupTypeUnion(types)
 	arrayType := zctx.LookupTypeArray(unionType)
-	return zed.Value{arrayType, b.Bytes()}, nil
+	return zed.NewValue(arrayType, b.Bytes())
 }
 
-func (c *Collect) ConsumeAsPartial(zv zed.Value) error {
+func (c *Collect) ConsumeAsPartial(val *zed.Value) {
```
mattnibs, line 99: Did we decide to go with val of zv?

expr/expr_test.go
```diff
@@ -241,33 +215,33 @@ func TestCompareNumbers(t *testing.T) {
 		record = fmt.Sprintf(
 			`{x:%s (%s),s:"hello",bs:"world" (bstring),i:10.1.1.1,n:10.1.0.0/16} (=0)`, one, typ)
 
-		testWarning(t, "x == s", record, expr.ErrIncompatibleTypes, "comparing integer and string")
```
nwt, line 244: `testWarning` is now unused, so delete it.

expr/agg/collect.go
nwt: @mattnibs: @mccanne and I think we should, both because it's less cryptic and for agreement with `vals := batch.Values()`.

docs/language/operators.md
mattnibs: Or actually it doesn't look like its actually implemented...

mattnibs: It's implemented up to semantic, but not it appears in to be in the kernel (perhaps I am missing something?). If not, maybe get rid of this change- it'll make it easier to review it when it goes in.

nwt (commented):
Maybe rename expr/ztests/count-by.yaml to rename.yaml?

docs/language/operators.md
mccanne: I'm trying to move fast rather than make the PR perfect.  The motivation for this PR was Let and scopes, so I put a few things in as scaffolding.  This is mentioned in the PR description and git comment.

expr/ztests/unit-logical.yaml
```diff
@@ -0,0 +1,32 @@
+script: |
```
nwt, line 1: Nit: Just call this logical.yaml

index/keys.go
```diff
@@ -0,0 +1,54 @@
...
+		cutter: cutter,
+		valid:  make(map[zed.Type]struct{}),
+	}, nil
+}
+
+func (k *Keyer) Keys() []field.Path {
+	return k.keys
+}
+
+func (k *Keyer) KeyOf(ectx expr.Context, val *zed.Value) (*zed.Value, error) {
```
nwt, line 36: Nit: Just call this Key.

lake/root.go
```diff
@@ -175,34 +175,35 @@ func (r *Root) readLakeMagic(ctx context.Context) error {
...
+func (r *Root) batchifyPools(ctx context.Context, zctx *zed.Context, f expr.Filter) ([]zed.Value, error) {
 	m := zson.NewZNGMarshalerWithContext(zctx)
 	m.Decorate(zson.StylePackage)
 	pools, err := r.ListPools(ctx)
 	if err != nil {
 		return nil, err
 	}
-	var batch zbuf.Array
+	ectx := expr.NewContext()
+	var batch []zed.Value
```
nwt, line 186: Nit: vals would be a better name for this variable in all these batchify* functions, but we can do that in another PR.

value.go
```diff
@@ -14,11 +14,38 @@ var (
...
+	NullDuration = &Value{Type: TypeDuration}
+	NullTime     = &Value{Type: TypeTime}
+	NullFloat32  = &Value{Type: TypeFloat32}
+	NullFloat64  = &Value{Type: TypeFloat64}
+	NullBool     = &Value{Type: TypeBool}
+	NullBytes    = &Value{Type: TypeBytes}
+	NullString   = &Value{Type: TypeString}
+	NullIP       = &Value{Type: TypeIP}
+	NullNet      = &Value{Type: TypeNet}
+	NullTypeType = &Value{Type: TypeNet}
```
nwt, line 35:
```suggestion
	NullTypeType = &Value{Type: TypeType}
```

value.go
```diff
@@ -14,11 +14,38 @@ var (
...
+	Null         = &Value{Type: TypeNull}
+)
+
 type Value struct {
 	Type  Type
 	Bytes zcode.Bytes
 }
 
+type Allocator interface {
```
nwt, line 44: Nit: Put this before the Value declaration so the Value declaration stays with its constructor.

value.go
```diff
@@ -178,13 +201,15 @@ func (v Value) IsStringy() bool {
...
 
-var missingAsBytes = []byte(missing)
-
 func (v Value) IsMissing() bool {
-	return v.Type == TypeError && bytes.Equal(v.Bytes, missingAsBytes)
+	return v.Type == TypeError && bytes.Equal(v.Bytes, missing)
+}
+
+func (v Value) IsQuiet() bool {
+	return v.Type == TypeError && bytes.Equal(v.Bytes, quiet)
```
nwt, line 212: Maybe use Missing and Quiet here? And should v.Type be AliasOf(v.Type)?
```suggestion
	return v.Type == Missing.Type && bytes.Equal(v.Bytes, Missing.Bytes)
}

func (v Value) IsQuiet() bool {
	return v.Type == Quiet.Type && bytes.Equal(v.Bytes, Quiet.Bytes)
```

zio/zngio/scanner.go
```diff
@@ -232,7 +234,11 @@ func (w *worker) wantRecord(rec *zed.Value, progress *zbuf.Progress) bool {
 	// negatives because it expects a buffer of ZNG value messages, and
 	// rec.Bytes is just a ZNG value.  (A ZNG value message is a header
 	// indicating a type ID followed by a value of that type.)
-	if w.filter == nil || w.filter(rec) {
+	// We pass nil to filter for the scope since a scanner cannot appear
+	// inside of a nested scope obviously and buffer filters cannot include
+	// scoped references.
```
nwt, line 239: Is this forward looking or should it be removed?

nwt (approved):
My eyes have gone numb. :shipit:

value.go
mccanne: This is nice because it will still work when we changed to structured errors.  Missing and Quiet are internally created things and I'm not sure we should support customized variations of them with aliases, but maybe there's a use case for that?  We can also ad AliasOf later...

mccanne: Actually this is a little complicated because Missing and Quiet will need a type context when we have structured errors...

zio/zngio/scanner.go
mccanne: No that's a mistake... will remove.

PR #3372, opened 2021-12-22
Enable streaming stats on query endpoint

### LINKED ISSUE #3371: "Gateway Time-out" on query against cloud lake
I'm testing with Zed CLI commit 4b30ab4 against a test Zed lake in the cloud `https://cedar.lake.brimdata.io`.

```
$ echo $ZED_LAKE
https://cedar.lake.brimdata.io

$ zed -version
Version: v0.33.0-36-g4b30ab42

$ zed use
HEAD at fdns@main
```

Querying my populated `fdns` pool, this works:

```
$ zed query 'count()'
{count:100000000(uint64)}
```

However, this fails:

```
$ zed query 'count() by value | sort'
status code 504: <html>
<head><title>504 Gateway Time-out</title></head>
<body>
<center><h1>504 Gateway Time-out</h1></center>
</body>
</html>
```

mattnibs:
Stats were not getting sent on a periodic basis for the query endpoint.
This was causing long running queries on a service to time-out when
served behind an AWS load balancer: The load balancer would see no data
getting transmitted and time-out the connection.

Add a time Ticker in the query handler and also flush the response data
when a control message is sent.

Closes #3371

### DIFF
--- a/service/handlers.go
+++ b/service/handlers.go
@@ -3,6 +3,7 @@ package service
 import (
 	"errors"
 	"net/http"
+	"time"
 
 	"github.com/brimdata/zed"
 	"github.com/brimdata/zed/api"
@@ -19,6 +20,8 @@ import (
 	"github.com/brimdata/zed/zio/anyio"
 	"github.com/brimdata/zed/zqe"
 )
+
+const queryStatsInterval = time.Second
 
 func handleQuery(c *Core, w *ResponseWriter, r *Request) {
 	var req api.QueryRequest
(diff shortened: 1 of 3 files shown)


service/handlers.go
```diff
@@ -39,12 +42,18 @@ func handleQuery(c *Core, w *ResponseWriter, r *Request) {
...
 	if err != nil {
 		w.Error(err)
 	}
+	var statsTicker <-chan time.Time
+	if !noctrl {
+		t := time.NewTicker(queryStatsInterval)
+		statsTicker = t.C
+		defer t.Stop()
+	}
 	defer d.Close()
```
nwt, line 55: Nit: Move defers closer to their corresponding assignments to keep idioms intact.
```suggestion
	d, err := queryio.NewDriver(w.ResponseWriter, format, !noctrl)
	if err != nil {
		w.Error(err)
	}
	defer d.Close()
	var statsTicker <-chan time.Time
	if !noctrl {
		t := time.NewTicker(queryStatsInterval)
		defer t.Stop()
		statsTicker = t.C
	}
```

service/handlers.go
```diff
@@ -20,6 +21,8 @@ import (
 	"github.com/brimdata/zed/zqe"
 )
 
+const queryStatsInterval = time.Second
```
nwt, line 24: Nit: Put this inside handleQuery since it's unlikely it'll ever be used anywhere else.

nwt (approved):

service/handlers.go
mattnibs: Where would you put this at the top of function?

nwt: I'd just put it right before `t := time.NewTicker(queryStatsInterval)`, but top of the function works tool.

mccanne:
I am working on a branch to rework driver.Driver and runtime.Query.  driver.Interface will go away and the stats ticker won't be handled inside runtime.Query anymore.  Instead the client of runtime.Query will read batches or zed.Values and will implement its own ticker.  So, I guess I'm saying I think this will all change soon.  (I want to put this up right after expr-context.)

mattnibs:
@mccanne happy to close this for your pr if you plan on having it up soon...

PR #3374, opened 2021-12-22
Cors support

mason-fish:
Adds cors support to the api. The `OPTIONS` method must be added to any endpoints which may trigger a [pre-flight](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request) request (as opposed to a [simple](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#simple_requests) request). For now, as we play with integrations on observablehq, I've only found it to be necessary for the /query route.

I also baked in the observable notebook's domain into the allowed origins. We may want to refactor this to be a configuration defined by cli flags instead though.


service/core.go
```diff
@@ -164,7 +167,7 @@ func (c *Core) addAPIServerRoutes() {
 	c.authhandle("/pool/{pool}/branch/{branch}/merge/{child}", handleBranchMerge).Methods("POST")
 	c.authhandle("/pool/{pool}/branch/{branch}/revert/{commit}", handleRevertPost).Methods("POST")
 	c.authhandle("/pool/{pool}/stats", handlePoolStats).Methods("GET")
-	c.authhandle("/query", handleQuery).Methods("POST")
+	c.authhandle("/query", handleQuery).Methods("POST", "OPTIONS")
```
mattnibs, line 170: I believe the cors middleware will handle this, so the OPTIONS method isn't required no?

mason-fish: It is required because the router won't run the middleware unless there is a matching route (including method)

service/middleware.go
```diff
@@ -29,6 +32,35 @@ func requestIDMiddleware() mux.MiddlewareFunc {
 	}
 }
 
+func getAllowedOrigins() []string {
+	return []string{"observableusercontent.com", "localhost"}
+}
```
mattnibs, line 37:
```suggestion
var allowedOrigins = []string{"observableusercontent.com", "localhost"}
```

service/core.go
nwt:
```suggestion
	c.authhandle("/query", handleQuery).Methods("OPTIONS", "POST")
```

service/middleware.go
```diff
@@ -29,6 +32,35 @@ func requestIDMiddleware() mux.MiddlewareFunc {
...
+	url, err := url.Parse(unescapedOrigin)
+	if err != nil {
+		return valid
+	}
+	for _, allowedDomain := range getAllowedOrigins() {
+		if strings.HasSuffix(url.Host, allowedDomain) {
+			valid = true
+		}
+	}
+	return valid
```
nwt, line 54: I think this would be clearer without valid.
```suggestion
	unescapedOrigin, err := url.QueryUnescape(origin)
	if err != nil {
		return false
	}
	url, err := url.Parse(unescapedOrigin)
	if err != nil {
		return false
	}
	for _, allowedDomain := range getAllowedOrigins() {
		if strings.HasSuffix(url.Host, allowedDomain) {
			return true
		}
	}
	return false
```

service/core.go
```diff
@@ -102,6 +102,8 @@ func NewCore(ctx context.Context, conf Config) (*Core, error) {
 	}
 
 	routerAux := mux.NewRouter()
+	routerAux.Use(corsMiddleware())
+
```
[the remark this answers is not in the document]
mason-fish: For the /status and /version endpoints, I think yes we do if we want clients to be able to use those as health-checks the way we do in the app

service/middleware.go
```diff
@@ -29,6 +32,35 @@ func requestIDMiddleware() mux.MiddlewareFunc {
...
+}
+
+func corsMiddleware() mux.MiddlewareFunc {
+	return cors.New(cors.Options{
+		AllowOriginFunc:  originValidator,
+		AllowCredentials: true,
+	}).Handler
+}
```
mattnibs, line 62: I think you should be able to get away with this:
```suggestion
func corsMiddleware() mux.MiddlewareFunc {
	return cors.New(cors.Options{
		AllowCredentials: true,
		AllowedMethods: []string{
			http.MethodHead,
			http.MethodGet,
			http.MethodPost,
			http.MethodPut,
			http.MethodPatch,
			http.MethodDelete,
		},
		AllowedOrigins:  allowedOrigins,
	}).Handler
}
```

mattnibs (commented):
Can we get a system test for this using curl? See /service/ztests/curl-*

service/middleware.go
```diff
@@ -29,6 +32,35 @@ func requestIDMiddleware() mux.MiddlewareFunc {
...
+	unescapedOrigin, err := url.QueryUnescape(origin)
+	if err != nil {
+		return valid
+	}
+	url, err := url.Parse(unescapedOrigin)
+	if err != nil {
+		return valid
+	}
+	for _, allowedDomain := range getAllowedOrigins() {
+		if strings.HasSuffix(url.Host, allowedDomain) {
```
nwt, line 50: This will validate adversarialobservableusercontent.com but shouldn't.

service/middleware.go
nwt: Much better!

mason-fish: I think we still need the validator func because each observable notebook's origin has a dynamically generated sub-domain. I'll fix the suffix issue you caught though @nwt 👌

mattnibs: @mason-fish then use wildcards for the observablehq domain name.

mattnibs: e.g. `var allowedOrigins = []string{"*.observableusercontent.com", "localhost"}

mason-fish: Oh neato, hadn't noticed the rc/cors supported that 👍

mattnibs: Actually now that I'm thinking about it, if auth isn't enabled we should just allow cors on all domains. Some
```suggestion
func corsMiddleware(authEnabled bool) mux.MiddlewareFunc {
	opts := cors.Options{
		AllowCredentials: authEnabled,
		AllowedMethods: []string{
			http.MethodHead,
			http.MethodGet,
			http.MethodPost,
			http.MethodPut,
			http.MethodPatch,
			http.MethodDelete,
		}
	}
	if opts.AllowCredentials {
		opts.AllowedOrigins = allowedOrigins
	}
	return cors.New(opts).Handler
}
```

service/core.go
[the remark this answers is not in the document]
mattnibs: @mason-fish would these handlers then need the options method as well?

service/core.go
mattnibs: Frustrating! But makes sense.

service/core.go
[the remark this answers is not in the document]
mason-fish: Nope, the `OPTIONS` method is only invoked in a pre-flight cors request, and those would both qualify as simple cors requests. (see description for links to docs on those)

service/middleware.go
mattnibs: I thought about it, disregard my last suggestion. When we eventually add a cli option people can configure this to their hearts content.

service/ztests/curl-cors.yaml
```diff
@@ -0,0 +1,51 @@
...
+      === OPTIONS: allowed ===
+      Access-Control-Allow-Credentials: true
+      Access-Control-Allow-Headers: Content-Type
+      Access-Control-Allow-Methods: POST
+      Access-Control-Allow-Origin: http://test.observableusercontent.com
+      === OPTIONS: not allowed ===
+      === POST: allowed ===
+      Access-Control-Allow-Credentials: true
+      Access-Control-Allow-Origin: http://test.observableusercontent.com
+      === POST: not allowed ===
```
mattnibs, line 51: Great tests!

mattnibs (approved):

service/ztests/curl-cors.yaml
```diff
@@ -0,0 +1,51 @@
+script: |
+  source service.sh
+  echo "=== OPTIONS: allowed ==="
+  curl -sD - \
+    -X OPTIONS \
+    -H "Access-Control-Request-Method: POST" \
+    -H "Access-Control-Request-Headers: content-type" \
+    -H "Origin: http://test.observableusercontent.com" \
+    -H "Accept: application/json" \
+    -H "Content-Type: application/json" \
```
nwt, line 10: Tests pass without these.
```suggestion
```

service/ztests/curl-cors.yaml
```diff
@@ -0,0 +1,51 @@
+script: |
+  source service.sh
+  echo "=== OPTIONS: allowed ==="
```
nwt, line 3: We don't quotes this stuff in other ztests.
```suggestion
  echo === OPTIONS: allowed ===
```

service/ztests/curl-cors.yaml
```diff
@@ -0,0 +1,51 @@
...
+  curl -sD - \
+    -X OPTIONS \
+    -H "Access-Control-Request-Method: POST" \
+    -H "Access-Control-Request-Headers: content-type" \
+    -H "Origin: http://test.observableusercontent.com" \
+    -H "Accept: application/json" \
+    -H "Content-Type: application/json" \
+    $ZED_LAKE/query | grep -Fi "Access-Control-Allow"  | tr -d '\015'
```
nwt, line 11: This is sufficient.
```suggestion
    $ZED_LAKE/query | grep Access-Control-Allow  | tr -d '\015'
```

service/ztests/curl-cors.yaml
```diff
@@ -0,0 +1,51 @@
...
+  echo "=== OPTIONS: not allowed ==="
+  curl -sD - \
+    -X OPTIONS \
+    -H "Access-Control-Request-Method: POST" \
+    -H "Access-Control-Request-Headers: content-type" \
+    -H "Origin: http://adversarialobservableusercontent.com" \
+    -H "Accept: application/json" \
+    -H "Content-Type: application/json" \
+    $ZED_LAKE/query | grep -Fi "Access-Control-Allow"  || true
```
nwt, line 20: Use `!` instead of `|| true` in ztests.
```suggestion
  ! curl -sD - \
    -X OPTIONS \
    -H "Access-Control-Request-Method: POST" \
    -H "Access-Control-Request-Headers: content-type" \
    -H "Origin: http://adversarialobservableusercontent.com" \
    $ZED_LAKE/query | grep Access-Control-Allow
```

service/ztests/curl-cors.yaml
nwt: Agreed!

nwt (approved):
:shipit:

Please make sure the commit message is tidy when you merge!

PR #3375, opened 2021-12-22
Peeker: fill at least until min

mattnibs:
Peeker was previously holding up reads until it had been able to
completely fill up its entire buffer. This caused a problem when reading
from slow drip streams of data- you wouldn't see any data until the
buffer was full which may take a while when the buffer is large. Change
this behavior so Peeker.fill reads until it has at least enough data to
meet the min argument.

### DIFF
--- a/pkg/peeker/reader.go
+++ b/pkg/peeker/reader.go
@@ -47,22 +47,14 @@ func (r *Reader) fill(min int) error {
 	r.buffer = r.buffer[:cap(r.buffer)]
 	copy(r.buffer, r.cursor)
 	clen := len(r.cursor)
-	space := len(r.buffer) - clen
-	for space > 0 {
-		cc, err := r.Reader.Read(r.buffer[clen:])
-		if cc > 0 {
-			clen += cc
-			space -= cc
-		}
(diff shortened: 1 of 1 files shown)


pkg/peeker/reader.go
```diff
@@ -47,22 +47,14 @@ func (r *Reader) fill(min int) error {
...
-		if cc > 0 {
-			clen += cc
-			space -= cc
-		}
-		if err != nil {
-			if err == io.EOF {
-				r.eof = true
-				break
-			}
+	cc, err := io.ReadAtLeast(r.Reader, r.buffer[clen:], min)
```
nwt, line 50: I like how this is shorter.

Honestly curious: Is there a mnemonic (or something) else behind `cc`? (`io.ReadAtLeast` calls  it `n`.)

nwt (approved):

pkg/peeker/reader.go
mattnibs: @mccanne started it

nwt: Then call it `n`!

PR #3376, opened 2021-12-22
zed load: add fancy stats updates

### LINKED ISSUE #3354: zed load: progress bar
Add progress bar for zed load provided the current shell is a tty.

mattnibs:
Add stats updates to zed load if -q is not enabled and the current shell
is a tty.

Closes #3354

### DIFF
--- a/cli/inputflags/flags.go
+++ b/cli/inputflags/flags.go
@@ -6,6 +6,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"sync/atomic"
 
 	"github.com/brimdata/zed"
 	"github.com/brimdata/zed/cli/auto"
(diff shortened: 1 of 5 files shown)


cli/inputflags/flags.go
```diff
@@ -66,3 +69,90 @@ func (f *Flags) Open(ctx context.Context, zctx *zed.Context, engine storage.Engi
...
+			}
+			fmt.Fprintln(os.Stderr, err)
+			continue
+		}
+		readers = append(readers, file)
+	}
+	return readers, nil
+}
+
+func openFile(ctx context.Context, zctx *zed.Context, engine storage.Engine, path string, opts anyio.ReaderOpts) (*StatsReader, error) {
```
mattnibs, line 93: It was kind of gross having to replicate the logic of anyio.Open here but I need to wrap the Readers to count the bytes read- didn't see another way. Open to suggestions.

pkg/display/display.go
```diff
@@ -61,7 +65,11 @@ func (d *Display) Bypass() io.Writer {
 }
 
 func (d *Display) Close() {
-	close(d.close)
+	select {
+	case <-d.close:
+	default:
+		close(d.close)
+	}
```
nwt, line 72: I think this has a race condition. If you enter the default case but another goroutine closes `d.close` before you do, you'll panic.

mattnibs: I'll trade you a close for a ctx, cancel?

cli/inputflags/flags.go
nwt: Create a storage.Engine wrapper with a Get method that, in turn, wraps a storage.Reader with something like your StatsReader before returning it and then provides a way to get stats for the wrapped readers!

mattnibs: Great idea!

cmd/zed/load/command.go
```diff
@@ -92,3 +117,72 @@ func (c *Command) Run(args []string) error {
...
+	e.readers = append(e.readers, counter)
+	return counter, nil
+}
+
+func (e *engineWrap) status() (units.Bytes, units.Bytes, int) {
+	var read units.Bytes
+	for _, r := range e.readers {
+		read += units.Bytes(r.bytesRead())
+	}
+	return read, e.bytesTotal, int(atomic.LoadInt32(&e.completed))
```
nwt, line 168:
```suggestion
	var read int64
	for _, r := range e.readers {
		read += r.bytesRead()
	}
	return units.Bytes(read), e.bytesTotal, int(atomic.LoadInt32(&e.completed))
```

nwt (approved):

pkg/display/display.go
```diff
@@ -18,17 +19,24 @@ type Display struct {
 	interval time.Duration
 	updater  Displayer
 	buffer   *bytes.Buffer
-	close    chan struct{}
+	ctx      context.Context
+	cancel   context.CancelFunc
```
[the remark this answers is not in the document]
mattnibs: Agreed, good idea.

Issue #3343, opened 2021-12-17
Template literals


mattnibs:
Add the ability to include template literals, denoted by `${var}`, embedded in StringLiteral values.

For example:
```bash
echo '{var:"world"}' | zq -z 'yield "hello ${var}"' -
```

would return:
```
"hello world"
```

PR #3377, opened 2021-12-23
Add Template Expressions

mattnibs:
Add the ability to include template expressions inside of
StringLiterals.

### DIFF
--- a/compiler/parser/support.go
+++ b/compiler/parser/support.go
@@ -33,6 +33,20 @@ func makeArgMap(args interface{}) (interface{}, error) {
 	return m, nil
 }
 
+func makeTemplateExprChain(in interface{}) interface{} {
+	rest := in.([]interface{})
+	ret := rest[0]
+	for _, part := range rest[1:] {
+		ret = map[string]interface{}{
+			"kind": "BinaryExpr",
+			"op":   "+",
+			"lhs":  ret,
+			"rhs":  part,
+		}
+	}
+	return ret
+}
+
 func joinChars(in interface{}) string {
 	str := bytes.Buffer{}
 	for _, i := range in.([]interface{}) {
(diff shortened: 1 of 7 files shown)


mattnibs:
Closes #3343

compiler/parser/ztests/string-templates.yaml
```diff
@@ -0,0 +1,24 @@
...
+   "hello 776f726c64"
+   "hello ${this}"
+   "hello ${to_hex(this)}"
+   "==="
+   "incompatible types"(error)
+   "hello 7f000001"
+   "hello ${this}"
+   "hello ${to_hex(this)}"
+   "==="
+
```
nwt, line 24:
```suggestion
```

compiler/parser/ztests/template-literals.yaml
```diff
@@ -0,0 +1,26 @@
...
+   "hello 776f726c64"
+   "hello ${this}"
+   "hello ${to_hex(this)}"
+   "yo dawg world"
+   "==="
+   "incompatible types"(error)
+   "hello 7f000001"
+   "hello ${this}"
+   "hello ${to_hex(this)}"
+   "yo incompatible types"(bstring)
```
mattnibs, line 25: This feels like a bug, but not from this pr.

mccanne: > This feels like a bug, but not from this pr.

This should be fixed in #3396

compiler/parser/support.go
```diff
@@ -33,6 +33,20 @@ func makeArgMap(args interface{}) (interface{}, error) {
 	return m, nil
 }
 
+func makeTemplateExprChain(in interface{}) interface{} {
```
nwt, line 36: Maybe someday we can get rid of this and just use makeBinaryExprChain since that's so similar.

nwt (approved):

nwt (approved):

PR #3380, opened 2021-12-23
Refactor client side auth

mattnibs:
- Move the functionality of cmd/zed/auth/devauth into api/client/auth0.
- Add auth0.Store responsible for storing and retrieving auth0
  credentials.


api/client/auth0/store.go
```diff
@@ -0,0 +1,107 @@
...
+
+import (
+	"errors"
+	"io/fs"
+
+	zedfs "github.com/brimdata/zed/pkg/fs"
+	"github.com/brimdata/zed/pkg/storage"
+)
+
+func normalizeLakeURI(lake string) string {
```
nwt, line 11: Nit: This isn't that interesting so put it down below, after it's used.

api/client/auth0/store.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+type ServiceInfo struct {
+	URL    string `json:"url"`
+	Tokens Tokens `json:"tokens"`
+}
+
+type Credentials struct {
+	Version  int           `json:"version"`
+	Services []ServiceInfo `json:"services"`
```
nwt, line 71: Any reason not to simplify by changing this to a URL-keyed map?

cli/lake.go
```diff
@@ -43,13 +42,13 @@ func (l *LakeFlags) Connection() (*client.Connection, error) {
 	if !api.IsLakeService(uri) {
 		return nil, errors.New("cannot open connection on local lake")
 	}
-	creds, err := l.LoadCredentials()
+	conn := client.NewConnectionTo(uri.String())
+	tokens, err := l.AuthStore().LakeTokens(uri.String())
 	if err != nil {
 		return nil, err
 	}
```
nwt, line 49:
```suggestion
	tokens, err := l.AuthStore().LakeTokens(uri.String())
	if err != nil {
		return nil, err
	}
	conn := client.NewConnectionTo(uri.String())
```

cli/lake.go
```diff
@@ -69,6 +68,14 @@ func (l *LakeFlags) Open(ctx context.Context) (api.Interface, error) {
 	return api.OpenLocalLake(ctx, uri)
 }
 
+func (l *LakeFlags) AuthStore() *auth0.Store {
+	return auth0.NewStore(l.authStorePath())
+}
+
+func (l *LakeFlags) authStorePath() string {
+	return filepath.Join(l.ConfigDir, credsFileName)
```
nwt, line 76:
```suggestion
	return auth0.NewStore(filepath.Join(l.ConfigDir, credsFileName)
```

cmd/zed/auth/login.go
```diff
@@ -51,42 +58,55 @@ func (c *LoginCommand) Run(args []string) error {
...
-				browser.OpenURL(res.VerificationURL)
-			}
-			return nil
-		},
-	})
+	auth0client, err := auth0.NewClient(*method.Auth0)
+	if err != nil {
+		return err
+	}
+	dcr, err := auth0client.GetDeviceCode(ctx, deviceCodeScope)
```
nwt, line 72: Nit: Ditch deviceCodeScope.
```suggestion
	dcr, err := auth0client.GetDeviceCode(ctx, "email offline_access openid profile")
```

cmd/zed/auth/login.go
```diff
@@ -51,42 +58,55 @@ func (c *LoginCommand) Run(args []string) error {
...
 	if err != nil {
 		return err
 	}
-	creds, err := c.LoadCredentials()
+	fmt.Println("Complete authentication at:", dcr.VerificationURIComplete)
```
nwt, line 76:
```suggestion
	fmt.Println("Complete authentication at", dcr.VerificationURIComplete)
```

api/client/auth0/auth0.go
```diff
@@ -0,0 +1,131 @@
...
+	RefreshToken string `json:"refresh_token,-"`
+}
+
+type tokenResponse struct {
+	AccessToken  string `json:"access_token"`
+	ExpiresIn    int    `json:"expires_in"`
+	IDToken      string `json:"id_token"`
+	RefreshToken string `json:"refresh_token"`
+	TokenType    string `json:"token_type"`
+}
```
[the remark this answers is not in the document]
mattnibs: They will be used in the follow up refresh tokens pr so I'll keep these where they are.

api/client/auth0/store.go
mattnibs: Avoiding a migration here, but yes we should do this. I'll do this in a follow up pr that will make the auth0 endpoint to services endpoints.

nwt: > I'll do this in a follow up pr that will make the auth0 endpoint to services endpoints.
I don't follow. What did you mean here?

mattnibs: We need to really map the auth0 api endpoint to each token instead of the service uri. As is (with my refresh token pr) if you have an auth0 authenticated lake service and you authenticate a different lake service using the same auth0 api endpoint, the previous access token is invalidated. Actually I might as well address this in my refresh token pr.

api/client/auth0/auth0.go
```diff
@@ -0,0 +1,131 @@
...
+	var res DeviceCodeResponse
+	err := c.post(ctx, "/oauth/device/code", deviceCodeRequest{
+		Audience: c.config.Audience,
+		ClientID: c.config.ClientID,
+		Scope:    scope,
+	}, &res)
+	return res, err
+}
+
+func (c *Client) GetAuthTokens(ctx context.Context, dcr DeviceCodeResponse) (Tokens, error) {
```
nwt, line 84: Maybe just call this GetTokens?

api/client/auth0/store.go
```diff
@@ -0,0 +1,107 @@
...
+func (s Store) SetLakeTokens(lake string, tokens Tokens) error {
+	creds, err := s.load()
+	if err != nil {
+		return err
+	}
+	creds.AddTokens(lake, tokens)
+	return s.save(creds)
+}
+
+func (s Store) RemoveLakeTokens(lake string) error {
```
nwt, line 36: Maybe remove "Lake" from these names?

cmd/zed/auth/login.go
```diff
@@ -51,42 +56,55 @@ func (c *LoginCommand) Run(args []string) error {
...
+		return err
+	}
+	if err := c.AuthStore().SetLakeTokens(c.Lake, tokens); err != nil {
 		return fmt.Errorf("failed to save credentials file: %w", err)
 	}
 	fmt.Printf("Login successful, stored credentials for %s\n", c.Lake)
 	return nil
 }
+
+func (c *LoginCommand) pollForTokens(ctx context.Context, client *auth0.Client, dcr auth0.DeviceCodeResponse) (auth0.Tokens, error) {
```
nwt, line 90: This doesn't make much sense as a `LoginCommand` method since the receiver is unused. But I think it does make sense as an `auth0.Client` method.

cmd/zed/auth/login.go
```diff
@@ -40,6 +42,9 @@ func (c *LoginCommand) Run(args []string) error {
 	if len(args) > 0 {
 		return errors.New("login command takes no arguments")
 	}
+	if err := os.MkdirAll(c.ConfigDir, 0700); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
mattnibs: I kind of wanted to make auth0.Store not have to be aware of the directory structure and just assume it already exists, but I think what you are saying makes it all simpler, will do.

cmd/zed/auth/login.go
```diff
@@ -51,42 +56,55 @@ func (c *LoginCommand) Run(args []string) error {
 	switch method.Kind {
 	case api.AuthMethodAuth0:
 	case api.AuthMethodNone:
-		return fmt.Errorf("Zed lake service at %s does not support authentication", c.Lake)
+		return fmt.Errorf("Zed service at %s does not support authentication", c.Lake)
 	default:
-		return fmt.Errorf("Zed lake service at %s requires unknown authentication method %s", c.Lake, method.Kind)
+		return fmt.Errorf("Zed service at %s requires unknown authentication method %s", c.Lake, method.Kind)
```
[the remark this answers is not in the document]
mattnibs: Good point.

api/client/auth0/auth0.go
mattnibs: Yep.

cmd/zed/auth/login.go
mattnibs: Yeah, good idea.

cmd/zed/auth/login.go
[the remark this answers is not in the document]
nwt: In case it hasn't already occurred to you, it doesn't have to be aware of the directory structure if it does `os.MkDirall(filepath.Dir(pathToFile))`.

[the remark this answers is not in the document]
mattnibs: Yeah I got that.

api/client/auth0/store.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+	creds.AddTokens(uri, tokens)
+	return s.save(creds)
+}
+
+func (s Store) RemoveTokens(lake string) error {
+	creds, err := s.load()
+	if err != nil {
+		return err
+	}
```
nwt, line 42:
```suggestion
func (s Store) RemoveTokens(uri string) error {
	creds, err := s.load()
	if err != nil {
		return err
	}
	creds.RemoveTokens(uri)
```

api/client/auth0/store.go
```diff
@@ -0,0 +1,112 @@
...
+	URL    string `json:"url"`
+	Tokens Tokens `json:"tokens"`
+}
+
+type Credentials struct {
+	Version  int           `json:"version"`
+	Services []ServiceInfo `json:"services"`
+}
+
+func (c *Credentials) ServiceTokens(uri string) *Tokens {
```
nwt, line 74: Nit: I think we should just call this `Tokens`.

nwt (approved):

api/client/auth0/store.go
mattnibs: Oh shoot!

PR #3381, opened 2021-12-23
zed: auto refresh expired access tokens

mattnibs:
If a request fails due to using an expired access token, have the
client.Connection automatically retrieve a new access token from Auth0
and retry the request.

### DIFF
--- a/api/client/auth0/auth0.go
+++ b/api/client/auth0/auth0.go
@@ -122,6 +122,29 @@ func (c *Client) getDeviceCodeTokens(ctx context.Context, dcr DeviceCodeResponse
 	}, nil
 }
 
+func (c *Client) RefreshToken(ctx context.Context, refreshToken string) (Tokens, error) {
+	var res tokenResponse
+	err := c.post(ctx, "/oauth/token", tokenRequest{
+		ClientID:     c.config.ClientID,
+		GrantType:    "refresh_token",
+		RefreshToken: refreshToken,
+	}, &res)
+	if err != nil {
+		return Tokens{}, err
+	}
+	if res.RefreshToken != "" {
+		// res.RefreshToken is only set when refresh token rotation is
+		// enabled for the Auth0 application specified by c.config.ClientID.
+		refreshToken = res.RefreshToken
+	}
+	return Tokens{
+		Access:     res.AccessToken,
+		Expiration: time.Now().Add(time.Duration(res.ExpiresIn) * time.Second),
+		ID:         res.IDToken,
+		Refresh:    refreshToken,
+	}, nil
+}
+
 func (c *Client) post(ctx context.Context, path string, body, out interface{}) error {
 	b, err := json.Marshal(body)
 	if err != nil {
--- a/api/client/connection.go
+++ b/api/client/connection.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/brimdata/zed"
 	"github.com/brimdata/zed/api"
+	"github.com/brimdata/zed/api/client/auth0"
 	"github.com/brimdata/zed/compiler/parser"
 	"github.com/brimdata/zed/lake"
 	"github.com/brimdata/zed/lake/branches"
@@ -42,6 +43,7 @@ var (
 )
 
 type Connection struct {
+	auth          *auth0.Store
 	client        *http.Client
 	defaultHeader http.Header
 	hostURL       string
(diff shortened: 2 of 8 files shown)


api/client/auth0/auth0.go
```diff
@@ -0,0 +1,149 @@
...
+		RefreshToken: refreshToken,
+	}, &res)
+	if err != nil {
+		return Tokens{}, err
+	}
+	return Tokens{
+		Access:     res.AccessToken,
+		Expiration: time.Now().Add(time.Duration(res.ExpiresIn) * time.Second),
+		ID:         res.IDToken,
+		Refresh:    res.RefreshToken,
```
nwt, line 116:
```suggestion
	if res.RefreshToken != "" {
		// res.RefreshToken is only set when refresh token rotation is
		// enabled for the Auth0 application specified by c.config.ClientID.
		refreshToken = res.RefreshToken
	}
	return Tokens{
		Access:     res.AccessToken,
		Expiration: time.Now().Add(time.Duration(res.ExpiresIn) * time.Second),
		ID:         res.IDToken,
		Refresh:    refreshToken,
```

api/client/connection.go
```diff
@@ -83,21 +97,30 @@ type Response struct {
...
+		if err != nil {
+			return nil, err
+		}
+		res, err := c.client.Do(httpreq)
+		if err != nil {
+			return nil, err
+		}
+		if res.StatusCode < 200 || res.StatusCode > 299 {
+			err = parseError(res)
+			if i == 0 && res.StatusCode == 401 && err.(*ErrorResponse).Err.Error() == "invalid token" {
```
nwt, line 111: Retrying the request like this unfortunately is not going to work if `req.Body` is an `io.Reader`. One way to handle that is with a lightweight preflight request like `client.Connection.AuthIdentity`, either for all requests or for just `io.Reader` requests. Another way to handle it is to rely on the token's expiration time to trigger the refresh. (I like the expiration time approach because it avoids the extra request.)

mattnibs: Excellent point.

mattnibs: If the body is the request I'm going to have it use a track/recorder. package netutil used to have something similar for similar situations. I believe the standard library does this for redirects in requests.

nwt: That works too!

mattnibs: Alright I added request body replay functionality. This will actually save us on redirects as well since we were passing non-recognized io.Readers to http request.

nwt, line 111: Better to use `errors.As` than a type assertion. And we as well use `api.Error`.
```suggestion
			var apierr *api.Error
			if i == 0 && res.StatusCode == 401 && errors.As(err, &apierr) && apierr.Message == "invalid token" {
```

api/client/connection.go
```diff
@@ -83,21 +97,30 @@ type Response struct {
...
+				var access string
+				if access, err = c.refreshAuthToken(req.ctx); err == nil {
+					req.Header.Set("Authorization", "Bearer "+access)
+					continue
+				}
```
nwt, line 116: If `c.refreshAuthToken` fails, this will return a `Response` that doesn't match `err`.
```suggestion
				access, err = c.refreshAuthToken(req.ctx)
				if err != nil {
					return nil, err
				}
				req.Header.Set("Authorization", "Bearer "+access)
				continue
```

api/client/connection.go
mattnibs: Can't use api.Error since for some reason this endpoint is returning an api.Error, I'll change it to ErrorResponse.

api/client/request.go
```diff
@@ -61,10 +62,71 @@ func (r *Request) HTTPRequest() (*http.Request, error) {
...
+}
+
+func (r *replayReader) Read(b []byte) (int, error) {
+	if r.buf.Len() > 0 {
+		return r.buf.Read(b)
+	}
+	return r.Reader.Read(b)
+}
+
+func (r *Request) GetBody() (io.ReadCloser, error) {
```
nwt, line 107: Nits: I don't think this is really part of the public interface, so I wouldn't export it. I also think it's more interesting than `recordReader` and `replayReader` and deserves to come first.

api/client/request.go
```diff
@@ -61,10 +62,71 @@ func (r *Request) HTTPRequest() (*http.Request, error) {
...
+	// Set GetBody so posted data is preserved on redirects or requests with
+	// expired access tokens.
+	req.GetBody = r.GetBody
 	req.Header = r.Header
 	return req, nil
 }
 
+type recordReader struct {
+	io.Reader
+	buf      *bytes.Buffer
```
nwt, line 74:
```suggestion
	buf      bytes.Buffer
```

nwt (approved):

PR #3390, opened 2021-12-24
eliminate driver and proc warnings channel

mccanne:
This commit eliminates the driver paakage and drops the warning
channel from proc.Context.  Warnings are now delivered via
top-level error results in the dataflow output.

Queries are now run from a simpler API in package runtime, which
returns a query object that is a puller and is easily turned into
a zio.Reader or the new interface zio.ProgressReader.  Progress updates
are now more cleanly handled through this interface.

We also fixed the compiler to properly return meters for file system
scanners not just lakes and simplified the interface for retrieving
a meter for the flowgraph.

The logic to teardown a flowgraph is now all in proc.  In particular,
proc.Mux and the new type proc.Latcher (for single-output graphs)
both tear down the flowgraph at EOS and convert the double-EOS batch
protocol to a single-EOS semantic.  We also added proc.Catcher to
implement recover() to provide an easy means to catch panics that
can be raised in the expr package when encountering fatal errors
(like system I/O problems or corrupt ZNG data).

Since the driver interface is gone, we needed a way to signal channel
completions.  This is accomplished by sending a proc.EndOfChannel
batch that is otherwise an empty batch but is detected by the
query endpoint in the server so it can send the corresponding events
to a client.

To carry control messages through the zio.Reader interface, we create
zbuf.Control, which implements the error interface, so control messages
including progress updates (zio.Progress) can be "read" for a zio.Reader.


api/queryio/writer.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/brimdata/zed/zio"
+	"github.com/brimdata/zed/zio/anyio"
+	"github.com/brimdata/zed/zio/jsonio"
+)
+
+type ControlWriter interface {
+	zio.WriteCloser
+	WriteControl(interface{}) error
+}
+
```
nwt, line 19: No point in exporting this.
```suggestion
```

api/queryio/writer.go
```diff
@@ -0,0 +1,93 @@
...
+func (w *Writer) Error(err error) {
+	w.WriteControl(api.QueryError{err.Error()}, true)
+}
+
+func (w *Writer) WriteControl(value interface{}, ctrl bool) error {
+	var err error
+	if ctrl {
+		if ctrl, ok := w.writer.(ControlWriter); ok {
```
nwt, line 81:
```suggestion
		type controlWriter interface {
			WriteControl(interface{}) error
		}
		if ctrl, ok := w.writer.(controlWriter); ok {
```

proc/latch.go
```diff
@@ -0,0 +1,103 @@
...
+	if l.eos {
+		l.pctx.Cancel()
+		return nil, nil
+	}
+	for {
+		batch, err := l.parent.Pull()
+		if err == EndOfBatch {
+			continue
+		}
+		if err != nil {
```
nwt, line 87:
```suggestion
		if err != nil {
			if err == EndOfBatch {
				continue
			}
```

lake/api/api.go
```diff
@@ -19,7 +18,7 @@ import (
 )
 
 type Interface interface {
-	Query(ctx context.Context, d driver.Driver, head *lakeparse.Commitish, src string, srcfiles ...string) (zbuf.Progress, error)
+	Query(ctx context.Context, head *lakeparse.Commitish, ctrl bool, src string, srcfiles ...string) (zbuf.ProgressReader, error)
```
nwt, line 21: Adding `ctrl` here feels wrong.
```suggestion
	Query(ctx context.Context, head *lakeparse.Commitish, ctrl bool, src string, srcfiles ...string) (zbuf.ProgressReader, error)
```

api/queryio/writer.go
mccanne: Ok, but we can't just delete it.

api/queryio/writer.go
mccanne: I'd rather change it so it's not exported and keep the compile-time checks that the writers implement the interface.

lake/api/api.go
mccanne: Ok, see dc3bb75

api/queryio/client.go
```diff
@@ -1,69 +1,59 @@
...
-	return run.stats, ctx.Err()
+type Query struct {
+	reader *zngio.Reader
 }
 
-type runner struct {
-	driver driver.Driver
-	cid    int
-	stats  zbuf.Progress
+//  NewQuery returns a Query that reads ZNG-encoded query response
```
nwt, line 19:
```suggestion
// NewQuery returns a Query that reads a ZNG-encoded query response
```

lake/api/querydriver.go
```diff
@@ -1,41 +1,29 @@
 package api
```
nwt, line 1: `querydriver.go` no longer makes sense as a name for this file. Move the contents to `lake/api/api.go` since that's the only file in which they're used and both files are small?

cli/zq/command.go
```diff
@@ -144,19 +144,14 @@ func (c *Command) Run(args []string) error {
 	if err != nil {
 		return err
 	}
-	d := driver.NewCLI(writer)
-	if !c.quiet {
```
nwt, line 148: `Command.quiet` is unused now. Delete it!

zbuf/reader.go
```diff
@@ -0,0 +1,80 @@
...
+type meterReader struct {
+	zio.Reader
+	progress Progress
+}
+
+var _ ProgressReader = (*meterReader)(nil)
+
+func MeterReader(r zio.Reader) *meterReader {
+	return &meterReader{Reader: r}
+}
```
nwt, line 55:
```suggestion
```suggestion
func MeterReader(r zio.Reader) ProgressReader {
	return &meterReader{Reader: r}
}

type meterReader struct {
	zio.Reader
	progress Progress
}
```
```

File: zbuf/reader.go [hunk 9]
```diff
@@ -0,0 +1,80 @@
...
+
+func (n *noControl) Read() (*zed.Value, error) {
+	for {
+		val, err := n.Reader.Read()
+		if err != nil {
+			if _, ok := err.(*Control); ok {
+				continue
+			}
+		}
```
nwt, line 36: Type assertion will fail if `err` is nil.
```suggestion
		if _, ok := err.(*Control); ok {
			continue
		}
```

zbuf/reader.go
```diff
@@ -0,0 +1,80 @@
...
+func (m *meterReader) Read() (*zed.Value, error) {
+	for {
+		val, err := m.Reader.Read()
+		if err != nil {
+			if ctrl, ok := err.(*Control); ok {
+				if progress, ok := ctrl.Message.(Progress); ok {
+					m.progress = progress
+				}
+			}
+		}
```
nwt, line 70:
```suggestion
		if ctrl, ok := err.(*Control); ok {
			if progress, ok := ctrl.Message.(Progress); ok {
				m.progress = progress
			}
		}
```

cli/zq/command.go
mccanne: The WarningsReader is coming back very soon and will wire back into quiet.

lake/api/remote.go
```diff
@@ -104,13 +104,43 @@ func (r *RemoteSession) Revert(ctx context.Context, poolID ksuid.KSUID, branchNa
...
+type queryReader struct {
+	reader zio.Reader
+	res    *client.Response
+}
+
+var _ zio.Reader = (*queryReader)(nil)
+
+func (q *queryReader) Read() (*zed.Value, error) {
+	val, err := q.reader.Read()
+	if (val == nil || err != nil) && q.res != nil {
```
nwt, line 116: I don't think the `q.res` test is necessary since it should never be nil.

service/ztests/curl-query-split.yaml
```diff
@@ -0,0 +1,21 @@
+# Test the output mux to make sure channel end messages appear in ZJSON
+
+script: |
+  source service.sh
+  zed create -q test
+  zed load -q -use test@main -
+  curl --stderr ignore -H "Accept: application/x-zjson" -d '{"query":"from test | split (=>pass;=>pasd;)"}' $ZED_LAKE/query\?ctrl=true |
```
nwt, line 7:
```suggestion
  curl --stderr ignore -H "Accept: application/x-zjson" -d '{"query":"from test | split (=>pass;=>pass;)"}' $ZED_LAKE/query\?ctrl=true |
```

proc/catcher.go
```diff
@@ -0,0 +1,33 @@
...
+func (c *Catcher) Pull() (b zbuf.Batch, err error) {
+	defer func() {
+		r := recover()
+		if r == nil {
+			return
+		}
+		err = zqe.RecoverError(r)
+	}()
+	b, err = c.parent.Pull()
+	return
```
nwt, line 28:
```suggestion
	defer func() {
		if r := recover(); r != nil {
			err = zqe.RecoverError(r)
		}
	}()
	return c.parent.Pull()
```

proc/catcher.go
```diff
@@ -0,0 +1,33 @@
...
+	parent Interface
+}
+
+func NewCatcher(parent Interface) *Catcher {
+	return &Catcher{parent}
+}
+
+// SafePull runs a Pull and catches panics and turns them out errors.
+// This should be called out the output puller of a flowgraph and by
+// the top-level puller of all new goroutine created inside of a flowgraph.
```
nwt, line 18: Move comment to Catcher and reword.
```suggestion
// Catcher wraps an Interface with a Pull method that recovers panics and turns them into errors.
// It should be wrapped around the output puller of a flowgraph and
// the top-level puller of any goroutine created inside of a flowgraph.
```

proc/latch.go
```diff
@@ -0,0 +1,103 @@
+package proc
+
+import (
+	"errors"
+
+	"github.com/brimdata/zed/zbuf"
+)
+
+// EndOfBatch signals the end of batch via the error value returned by a Pull()
+// on a Latch. EndOfBatch should be detected and acted upon by a caller of
```
nwt, line 10:
```suggestion
// EndOfBatch is returned by Latch.Pull to signal the end of a batch.
// It should be handled by a caller of
```

proc/latch.go
```diff
@@ -0,0 +1,103 @@
...
+func (l *Latch) Pull() (zbuf.Batch, error) {
+	if l.done {
+		return nil, l.err
+	}
+	b, err := l.parent.Pull()
+	if err == EndOfBatch {
+		// This breaks the protocol and shouldn't happen so we panic.
+		panic("proc.Latch received EndOfBatch")
+	}
+	if err != nil {
```
nwt, line 39:
```suggestion
	if err != nil {
		if err == EndOfBatch {
			// This breaks the protocol and shouldn't happen so we panic.
			panic("proc.Latch received EndOfBatch")
		}
```

runtime/query.go
```diff
@@ -0,0 +1,102 @@
...
+// that gracefully tears down the flowgraph.  Its AsReader() and AsProgressReader()
+// methods provide a convenient means to run a flowgraph as zio.Reader.
+type Query struct {
+	zbuf.Puller
+	pctx      *proc.Context
+	flowgraph *compiler.Runtime
+}
+
+var _ zbuf.Puller = (*Query)(nil)
+var _ io.Closer = (*Query)(nil)
```
nwt, line 28:
```suggestion
var _ zbuf.PullerCloser = (*Query)(nil)
```

runtime/query.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/brimdata/zed/compiler/ast"
+	"github.com/brimdata/zed/lakeparse"
+	"github.com/brimdata/zed/order"
+	"github.com/brimdata/zed/proc"
+	"github.com/brimdata/zed/zbuf"
+	"github.com/brimdata/zed/zio"
+	"go.uber.org/zap"
+)
+
+// Query runs a flowgraph as zubf.Puller and implements a Close() method
```
nwt, line 18:
```suggestion
// Query runs a flowgraph as a zubf.Puller and implements a Close() method
```

nwt (approved):

Issue #3224, opened 2021-10-28
improve error message for "incompatible type" error


mattnibs:
If I do`echo '{x:"blah"}' | zq -z 'x := x/2' -`

I get the output: 

```
incompatible types
{x:"blah"}
```

This is undesirable. The output we really want looks something like this:

```
{x:"cannot coerce a value of type string into an int"(error)}
```

This change could happen inside of the functions expr/coerce or it could happen wherever coerce is called.

mccanne:
I'm going to do this in two PRs: the first one will fix package expr to properly propagate coercion errors and the second one will improve the error message detail as suggested here.

PR #3395, opened 2021-12-25
propagate arithmetic expression errors properly

mccanne:
Errors were not propagated properly where error values could
be used incorrectly as actual values in expressions like string
concatenation.  This fixes arithmetic expressions to detect and
propagate errors correctly.

We will improve the error message for "incompatible types" in a 
subsequent PR.  See #3224.

This should fix the "bstring" problem in #3377.

### DIFF
--- a/expr/eval.go
+++ b/expr/eval.go
@@ -152,8 +152,12 @@ func (i *In) inContainer(typ zed.Type, elem, container *zed.Value) *zed.Value {
 		if err != nil {
 			panic(err)
 		}
-		_, err = i.vals.Coerce(elem, &zed.Value{typ, zv})
-		if err == nil && i.vals.Equal() {
+		_, errVal := i.vals.Coerce(elem, &zed.Value{typ, zv})
+		if errVal != nil {
+			if errVal != coerce.IncompatibleTypes {
+				return errVal
+			}
+		} else if i.vals.Equal() {
 			return zed.True
 		}
 	}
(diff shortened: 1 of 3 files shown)


expr/eval.go
```diff
@@ -152,8 +152,12 @@ func (i *In) inContainer(typ zed.Type, elem, container *zed.Value) *zed.Value {
 		if err != nil {
 			panic(err)
 		}
-		_, err = i.vals.Coerce(elem, &zed.Value{typ, zv})
-		if err == nil && i.vals.Equal() {
+		_, errVal := i.vals.Coerce(elem, &zed.Value{typ, zv})
+		if errVal != nil {
```
nwt, line 156:
```suggestion
		if _, errVal := i.vals.Coerce(elem, &zed.Value{typ, zv}); errVal != nil {
```

expr/eval.go
```diff
@@ -168,16 +172,24 @@ func (i *In) inMap(typ *zed.TypeMap, elem, container *zed.Value) *zed.Value {
 		if err != nil {
 			panic(err)
 		}
-		_, err = i.vals.Coerce(elem, &zed.Value{keyType, zv})
-		if err == nil && i.vals.Equal() {
+		_, errVal := i.vals.Coerce(elem, &zed.Value{keyType, zv})
+		if errVal != nil {
```
nwt, line 176:
```suggestion
		if _, errVal := i.vals.Coerce(elem, &zed.Value{keyType, zv}); errVal != nil {
```

expr/eval.go
```diff
@@ -168,16 +172,24 @@ func (i *In) inMap(typ *zed.TypeMap, elem, container *zed.Value) *zed.Value {
...
 			return zed.True
 		}
 		zv, _, err = iter.Next()
 		if err != nil {
 			panic(err)
 		}
-		_, err = i.vals.Coerce(elem, &zed.Value{valType, zv})
-		if err == nil && i.vals.Equal() {
+		_, errVal = i.vals.Coerce(elem, &zed.Value{valType, zv})
+		if errVal != nil {
```
nwt, line 188:
```suggestion
		if _, errVal = i.vals.Coerce(elem, &zed.Value{valType, zv}); errVal != nil {
```

nwt (approved):
Maybe add an overflow test?

mccanne:
> Maybe add an overflow test?

This PR doesn't change the semantics of overflow and we don't produce them.  Maybe we should have another look at the overflow semantics and/or the logic in coerce?

PR #3402, opened 2021-12-27
fix failing query in scripts/perf-compare.sh

nwt:
As of #3368, the command "zq -f zeek 'count() by id.orig_h' ..." in
scripts/perf-compare.sh fails because its output now includes Zed error
values, on which the Zeek writer chokes.  Fix by wrapping id.orig_h with
quiet() in that command.


nwt:
https://github.com/brimdata/zed/runs/4643336257 show `.github/workflows/perf-compare.yml` succeeding on this branch.

mccanne (approved):

PR #3404, opened 2021-12-27
update filters to handle zed.Values and propagate errors

### LINKED ISSUE #3393: update expr.Filter to properly handle embedded Zed errors and non-record values
expr.Filter should be change from type bool to *zed.Value so errors can be appropriately propagated and the logic throughout needs to be updated to not presume a top-levelrecord, e.g., SearchRecordOther crashes with this input:
```
echo 1 | zq "not this" -
```

mccanne:
The old filter logic had not yet been updated to deal with the
operator change from records to zed.Values.  Also, they were
unable to propagate errors as the filter type had a native
bool result.

This commit updates the filtering system to use expr.Evaluator
for filters.  Also, we clenaed up the search code, gave it better
names, and fixed some bugs, adding tests where needed.  Field name
search now properly decends into complex types.

There is now only one set of logical operators rather than
separate sets for filters and expressions.  We also handle
bool(null) in logical expressions without a panic.

Fixes #3393


expr/filter.go
```diff
@@ -11,83 +11,53 @@ import (
...
-			} else if !isRecord && eval(&zed.Value{c.Type, val}) {
-				return true
+	var match bool
+	recType := zed.TypeRecordOf(typ)
+	if recType != nil {
+		var nameIter stringsearch.FieldNameIter
+		nameIter.Init(recType)
+		for !nameIter.Done() {
+			body := nameIter.Next()
+			if s.pred(&zed.Value{Type: zed.TypeString, Bytes: body}) {
```
nwt, line 52:
```suggestion
			if s.pred(zed.NewValue(zed.TypeString, nameIter.Next())) {
```

expr/filter.go
```diff
@@ -113,85 +83,137 @@ func stringSearch(a, b string) bool {
...
-		// record columns for each unique record type.
-		match, ok := fieldNameCheck[val.Type]
-		if !ok {
-			nameIter.Init(zed.TypeRecordOf(val.Type))
-			for !nameIter.Done() {
-				if stringSearch(byteconv.UnsafeString(nameIter.Next()), term) {
-					match = true
-					break
-				}
+		} else if s.compare(&zed.Value{Type: typ, Bytes: body}) {
```
nwt, line 111:
```suggestion
		} else if s.compare(zed.NewValue(typ, body)) {
```

zio/zngio/scanner.go
```diff
@@ -226,16 +226,21 @@ func (w *worker) scanBatch(buf *buffer, mapper *zed.Mapper, streamZctx *zed.Cont
 	return batch, nil
 }
 
-func (w *worker) wantRecord(rec *zed.Value, progress *zbuf.Progress) bool {
-	progress.BytesRead += int64(len(rec.Bytes))
+func check(ectx expr.Context, this *zed.Value, filter expr.Evaluator) bool {
+	val := filter.Eval(ectx, this)
+	return val.Type == zed.TypeBool && zed.IsTrue(val.Bytes)
+}
```
nwt, line 232: Nit: wantValue is more interesting so put this after it.

expr/ztests/search-ip-scalar.yaml
```diff
@@ -0,0 +1,11 @@
+zed: 10.0.0.0/8
```
nwt, line 1: Nit: s/scalar/primitive/ in the file name.

expr/ztests/search-scalars.yaml
```diff
@@ -0,0 +1,10 @@
+zed: foo
```
nwt, line 1: Nit: s/scalars/primitives/ in the file name.

lake/pool.go
```diff
@@ -144,7 +144,15 @@ func (p *Pool) OpenBranchByName(ctx context.Context, name string) (*Branch, erro
...
 }
 
-func (p *Pool) batchifyBranches(ctx context.Context, recs []zed.Value, m *zson.MarshalZNGContext, f expr.Filter) ([]zed.Value, error) {
+func filter(ectx expr.Context, this *zed.Value, e expr.Evaluator) bool {
+	if e == nil {
+		return true
+	}
+	val, ok := expr.EvalBool(ectx, this, e)
+	return ok && val.Bytes != nil && zed.IsTrue(val.Bytes)
+}
```
nwt, line 153: Nit: The batchify functions are more interesting so put this after them.

nwt (approved):

PR #3407, opened 2021-12-28
change references of "root" to "this"

mccanne:


compiler/semantic/expr.go
```diff
@@ -348,8 +348,8 @@ func semAssignment(scope *Scope, a ast.Assignment) (dag.Assignment, error) {
 		lhs = &dag.Path{"Path", []string{call.Name}}
 	} else if agg, ok := a.RHS.(*ast.Agg); ok {
 		lhs = &dag.Path{"Path", []string{agg.Name}}
-	} else if _, ok := a.RHS.(*ast.Root); ok {
-		lhs = &dag.Path{"Path", []string{"."}}
+	} else if _, ok := a.RHS.(*ast.This); ok {
+		return dag.Assignment{}, errors.New("cannot assign to \"this\"")
```
nwt, line 352:
```suggestion
		return dag.Assignment{}, errors.New(`cannot assign to "this"`)
```

proc/groupby/ztests/count-by-this-err.yaml
```diff
@@ -0,0 +1,3 @@
+zed: 'count() by this'
+
+errorRE: "cannot assign to \"this\""
```
nwt, line 3:
```suggestion
errorRE: cannot assign to "this"
```

proc/groupby/ztests/count-by-this-err.yaml
```diff
@@ -0,0 +1,3 @@
+zed: 'count() by this'
```
nwt, line 1:
```suggestion
zed: count() by this
```

proc/groupby/ztests/count-by-this.yaml
```diff
@@ -0,0 +1,14 @@
+zed: 'count() by val:=this | sort this'
```
nwt, line 1:
```suggestion
zed: count() by val:=this | sort this
```

nwt (approved):
I think these are worth changing as well.
```
$ git grep -i root -- {compiler,expr,proc}/**/*.go
compiler/kernel/expr.go:        fn, root, err := function.New(zctx, call.Name, nargs)
compiler/kernel/expr.go:        if root {
compiler/semantic/expr.go:      // If we index a root record with a string constant, then just
compiler/semantic/expr.go:              if path := isRootIndex(scope, lhs, rhs); path != nil {
compiler/semantic/expr.go:func isRootIndex(scope *Scope, lhs, rhs dag.Expr) *dag.Path {
compiler/semantic/expr.go:                      if path := isRootIndex(scope, lhs, rhs); path != nil {
expr/function/function.go:      var root bool
expr/function/function.go:              root = true
expr/function/function.go:      return f, root, nil
proc/put/put.go:        root                 // values are being copied to the root record (put .=)
proc/put/put.go:        case root:
proc/put/put.go:                // top-level op should be root or record
```

PR #3415, opened 2021-12-29
fix scoping model and update const implementation

mccanne:
This commit makes the long-wanted change to the compiler to handle consts
properly.  ConstProc and TypeProc are now gone and const statements can
appear at the front of any sequential flowgraph segment and are carried
in ast.Sequential.

The dag no longer depends on the AST of zed values and types.  Instead,
they are encoded as ZSON strings.  The dag also knows nothing about consts
as the semantic pass now performs constant propagation and all consts
are turned into dag literals.

These changes were necessary at this time because without them, scoped
consts were getting hairy in compiler/kernal and the AsFilter() method
on zbuf.Filter needed to retain the table of consts.  By having the
semantic pass resolve all consts, this is all simpler and will work
well over the network for distributed queries.

To support dynamic type name resolution, we added the typename() function,
which takes a string type name and returns the zed type if it exists.
When named types are referred to and can't be resolved at compile time,
the semantic pass automatically inserts a call to typename().

This scoping work tees up the forthcoming PR on scoped variables
so subgraphs can refer to values from the outer "this" when traversing
new values for "this".


compiler/compiler.go
```diff
@@ -93,19 +92,15 @@ func New(pctx *proc.Context, inAST ast.Proc, adaptor proc.DataAdaptor, head *lak
...
 	if err != nil {
 		return nil, err
 	}
 	builder := kernel.NewBuilder(pctx, adaptor)
-	if err := builder.LoadConsts(consts); err != nil {
-		return nil, err
-	}
 	return &Runtime{
 		pctx:      pctx,
 		builder:   builder,
```
nwt, line 102: Nit: There's no longer any need to define `builder` up above.
```suggestion
		builder:   kernel.NewBuilder(pctx, adaptor),
```

compiler/kernel/bufferfilter.go
```diff
@@ -38,23 +37,27 @@ func CompileBufferFilter(e dag.Expr) (*expr.BufferFilter, error) {
...
+		val, err := zson.ParseValue(zctx, e.Value)
+		if err != nil {
+			return nil, err
+		}
+		if val.Type == zed.TypeNet {
 			return nil, nil
 		}
-		if e.Value.Type == "string" {
-			pattern := norm.NFC.Bytes(zed.UnescapeBstring([]byte(e.Value.Text)))
+		if val.Type == zed.TypeString {
```
nwt, line 59: Unlikely as there are to appear here, we should probably armor this against named types.
```go
		switch zed.AliasOf(val.Type) {
		case zed.TypeNet:
			return nil, nil
		case zed.TypeString:
```

compiler/kernel/bufferfilter.go
```diff
@@ -63,7 +66,11 @@ func CompileBufferFilter(e dag.Expr) (*expr.BufferFilter, error) {
 			return expr.NewOrBufferFilter(left, right), nil
 		}
 		left := expr.NewBufferFilterForStringCase(e.Text)
-		right, err := newBufferFilterForLiteral(e.Value)
+		literal, err := zson.ParseValue(zctx, e.Value)
+		if err != nil {
+			return nil, err
+		}
+		right, err := newBufferFilterForLiteral(&literal)
```
nwt, line 73: I don't think there's any need to parse `e.Value` a second time.
```suggestion
		right, err := newBufferFilterForLiteral(&val)
```

compiler/kernel/bufferfilter.go
```diff
@@ -73,41 +80,47 @@ func CompileBufferFilter(e dag.Expr) (*expr.BufferFilter, error) {
...
+			return &val, "="
 		}
 	} else if dag.IsTopLevelField(e.RHS) && e.Op == "in" {
-		if literal, ok := e.LHS.(*astzed.Primitive); ok && literal.Type != "net" {
-			return literal, "in"
+		if literal, ok := e.LHS.(*dag.Literal); ok {
+			val, err := zson.ParseValue(zctx, literal.Value)
+			if err != nil {
+				return nil, ""
+			}
```
nwt, line 100: We should probably change the signature of `isFieldEqualOrIn` to return these `zson.ParseValue` errors rather than swallowing them.

compiler/kernel/bufferfilter.go
```diff
@@ -73,41 +80,47 @@ func CompileBufferFilter(e dag.Expr) (*expr.BufferFilter, error) {
...
-	case "string":
+	case *zed.TypeOfString:
 		// Match the behavior of zed.ParseLiteral.
-		l.Type = "bstring"
-	}
-	v, err := zson.ParsePrimitive(l.Type, l.Text)
-	if err != nil {
-		return nil, err
+		val = zed.NewValue(zed.TypeBstring, val.Bytes)
 	}
```
nwt, line 121: 1. The first case is missing `zed.TypeOfUint8` and shouldn't include `zed.TypeOfBool` since it doesn't require multiple patterns. We should just use `zed.IsNumber`.
2. There's no need to turn Zed strings into bstrings since we're no longer trying to get `zson.ParsePrimitive` to behave like `zed.ParseLiteral`.
3. Might as well armor this against named types.
```go
	if id := val.Type.ID(); zed.IsNumber(id) || id == zed.IDNull {
		// All numbers are all comparable, so they can require up to three
		// patterns: float, varint, and uvarint.
		return nil, nil
	}
```

compiler/kernel/filter.go
```diff
@@ -23,20 +23,34 @@ func (f *Filter) AsEvaluator() (expr.Evaluator, error) {
...
 func (f *Filter) AsBufferFilter() (*expr.BufferFilter, error) {
 	if f == nil {
 		return nil, nil
 	}
-	return CompileBufferFilter(f.pushdown)
+	return CompileBufferFilter(f.builder.pctx.Zctx, f.pushdown)
 }
 
-func compileCompareField(zctx *zed.Context, scope *Scope, e *dag.BinaryExpr) (expr.Evaluator, error) {
+func isLiteral(zctx *zed.Context, e dag.Expr) (*zed.Value, error) {
```
nwt, line 36: 1. Nit: `compileCompareField` is more interesting so put this after it.
2. `isLiteral` doesn't feel like the right name for something that doesn't return a Boolean. How about `asLiteral`, `parseLiteral`, `maybeParseLiteral`, or `tryParseLiteral`?

compiler/kernel/filter.go
```diff
@@ -69,37 +83,40 @@ func compileCompareField(zctx *zed.Context, scope *Scope, e *dag.BinaryExpr) (ex
...
-			return match(zv) || contains(zv)
+		pred := func(val *zed.Value) bool {
+			return match(val) || contains(val)
 		}
-
 		return expr.SearchByPredicate(pred), nil
 	}
-	if node.Value.Type == "string" {
-		term := norm.NFC.Bytes(zed.UnescapeBstring([]byte(node.Value.Text)))
+	if val.Type == zed.TypeString {
```
nwt, line 109: Handle named types?
```go
	switch zed.AliasOf(val.Type) {
	case zed.TypeNet:
		...
	case zed.TypeString:
```

compiler/kernel/proc.go
```diff
@@ -40,18 +40,14 @@ var ErrJoinParents = errors.New("join requires two upstream parallel query paths
 
 type Builder struct {
 	pctx       *proc.Context
-	scope      *Scope
 	adaptor    proc.DataAdaptor
 	schedulers map[dag.Source]proc.Scheduler
 }
 
 func NewBuilder(pctx *proc.Context, adaptor proc.DataAdaptor) *Builder {
 	// Create scope and enter the global scope.
```
nwt, line 48:
```suggestion
```

compiler/kernel/proc.go
```diff
@@ -590,47 +586,22 @@ func (b *Builder) PushdownOf(trunk *dag.Trunk) (*Filter, error) {
...
-			scope.Bind(name, zv)
-		default:
-			return fmt.Errorf("kernel.LoadConsts: not a const: '%T'", p)
-		}
-	}
-	return nil
-}
-
-func evalAtCompileTime(zctx *zed.Context, scope *Scope, in dag.Expr) (*zed.Value, error) {
+func EvalAtCompileTime(zctx *zed.Context, in dag.Expr) (val *zed.Value, err error) {
```
nwt, line 589: Maybe call this `EvalNow`?

compiler/kernel/proc.go
```diff
@@ -590,47 +586,22 @@ func (b *Builder) PushdownOf(trunk *dag.Trunk) (*Filter, error) {
...
+	e, err := compileExpr(zctx, in)
 	if err != nil {
 		return nil, err
 	}
+	// Catch panic as the runtime will panic if there is a
+	// reference to a var not in scope, a field access null this, etc
+	defer func() {
+		if recover() != nil {
+			val = nil
+			err = errors.New("panic")
```
nwt, line 602:
```suggestion
		if e := recover(); e != nil {
			err = fmt.Errorf("%+v", e)
```

compiler/kernel/proc.go
```diff
@@ -590,47 +586,22 @@ func (b *Builder) PushdownOf(trunk *dag.Trunk) (*Filter, error) {
...
 	if in == nil {
 		return zed.Null, nil
 	}
-	e, err := compileExpr(zctx, scope, in)
+	e, err := compileExpr(zctx, in)
 	if err != nil {
 		return nil, err
 	}
+	// Catch panic as the runtime will panic if there is a
+	// reference to a var not in scope, a field access null this, etc
```
nwt, line 598:
```suggestion
	// reference to a var not in scope, a field access null this, etc.
```

compiler/semantic/ast.go
```diff
@@ -10,18 +10,12 @@ import (
...
 	if err != nil {
-		return nil, nil, err
+		return nil, err
 	}
-	entry, err := semSequential(ctx, scope, seq, adaptor, head)
-	if err != nil {
-		return nil, nil, err
-	}
-	return entry, consts, nil
+	return entry, nil
```
nwt, line 18:
```go
	return semSequential(ctx, NewScope(), seq, adaptor, head)
```

compiler/semantic/ast.go
```diff
@@ -10,18 +10,12 @@ import (
 )
 
 // Analyze analysis the AST and prepares it for runtime compilation.
```
nwt, line 12:
```suggestion
// Analyze analyzes the AST and prepares it for runtime compilation.
```

compiler/semantic/expr.go
```diff
@@ -117,13 +117,26 @@ func semExpr(scope *Scope, e ast.Expr) (dag.Expr, error) {
...
 	case *astzed.TypeValue:
-		tv, err := semType(scope, e.Value)
+		typ, err := semType(scope, e.Value)
 		if err != nil {
+			// If this is a type name, then we check to see if it's in the
+			// context because it has been defined locally.  If not, then
+			// the type needs to come from the data, in which case we replace
+			// the literal reference with a type_by_name() call.
+			// Note that we just check the top value here but there can be
+			// nested dynamic type references inside a complex type but this
```
nwt, line 127:
```suggestion
			// nested dynamic type references inside a complex type; this
```

compiler/semantic/expr.go
```diff
@@ -117,13 +117,26 @@ func semExpr(scope *Scope, e ast.Expr) (dag.Expr, error) {
 			Type: typ,
 		}, nil
 	case *astzed.TypeValue:
-		tv, err := semType(scope, e.Value)
+		typ, err := semType(scope, e.Value)
 		if err != nil {
+			// If this is a type name, then we check to see if it's in the
+			// context because it has been defined locally.  If not, then
+			// the type needs to come from the data, in which case we replace
+			// the literal reference with a type_by_name() call.
```
nwt, line 125:
```suggestion
			// the literal reference with a typename() call.
```

compiler/semantic/expr.go
```diff
@@ -252,14 +287,10 @@ func isIndexOfThis(scope *Scope, lhs, rhs dag.Expr) *dag.Path {
...
-	if p, ok := e.(*astzed.Primitive); ok && p.Type == "string" {
-		return p.Text, true
-	}
-	if ref, ok := e.(*dag.Ref); ok {
-		if c, ok := scope.Lookup(ref.Name).(*dag.Const); ok {
-			return isStringConst(scope, c.Expr)
-		}
+func isStringConst(scope *Scope, e dag.Expr) (field string, ok bool) {
+	val, err := kernel.EvalAtCompileTime(scope.zctx, e)
+	if err == nil && val != nil && val.Type == zed.TypeString {
```
nwt, line 292:
```suggestion
	if err == nil && val != nil && zed.AliasOf(val.Type) == zed.TypeString {
```

compiler/semantic/proc.go
```diff
@@ -208,20 +208,24 @@ func semSequential(ctx context.Context, scope *Scope, seq *ast.Sequential, adapt
...
 			return nil, err
 		}
 		ops = append(ops, converted)
 	}
 	return &dag.Sequential{
-		Kind: "Sequential",
-		Ops:  ops,
+		Kind:   "Sequential",
+		Ops:    ops,
+		Consts: consts,
```
nwt, line 228: Nit: Follow field definition order.
```suggestion
		Consts: consts,
		Ops:    ops,
```

compiler/semantic/scope.go
```diff
@@ -1,18 +1,24 @@
...
+	stack []*Binder
 }
 
 func NewScope() *Scope {
-	return &Scope{}
+	return &Scope{zctx: zed.NewContext()}
 }
 
-func (s *Scope) tos() Binder {
+func (s *Scope) tos() *Binder {
```
nwt, line 21: `tos` is a little cryptic, so maybe call this `top`?

compiler/semantic/scope.go
```diff
@@ -24,31 +30,72 @@ func (s *Scope) Exit() {
...
+	b := s.tos()
+	if _, ok := b.symbols[name]; ok {
+		return fmt.Errorf("symbol %q redefined", name)
+	}
+	val, err := kernel.EvalAtCompileTime(s.zctx, def)
+	if err != nil {
+		return err
+	}
+	if val.IsError() {
+		return fmt.Errorf("cannot resolve const %q at compile time", name)
```
nwt, line 57: Should we include `zson.MustFormatValue(*val)` in this error?

compiler/semantic/scope.go
```diff
@@ -24,31 +30,72 @@ func (s *Scope) Exit() {
...
-			return e.op
+			return e.ref
 		}
 	}
 	return nil
 }
 
+func (s *Scope) nvars() int {
+	var n int
+	for k := len(s.stack) - 1; k >= 0; k-- {
```
nwt, line 79: No need to iterate in reverse here.
```suggestion
	for k :=  range s.stack {
```

compiler/semantic/sql.go
```diff
@@ -220,13 +219,9 @@ func convertSQLTableRef(scope *Scope, e ast.Expr) (dag.Op, error) {
...
 	// the type name will be defined by the data stream.
-	if id, ok := dag.TopLevelField(converted); ok && scope.Lookup(id) == nil {
-		converted = &astzed.TypeValue{
-			Kind: "TypeValue",
-			Value: &astzed.TypeName{
-				Kind: "TypeName",
-				Name: id,
-			},
+	if id, ok := dag.TopLevelField(converted); ok {
+		if ref := scope.Lookup(id); ref == nil {
```
nwt, line 223:
```suggestion
		if scope.Lookup(id) == nil {
```

docs/language/functions.md
```diff
@@ -720,8 +721,30 @@ echo  '{x:"missing"(error),y:"hello"}'  | zq -z 'cut x:=quiet(x), y:=quiet(y)' -
...
+```
+
+**Output:**
+```mdtest-output
+(port=(int16))
+```
+
 ### `typeof`
 
+
```
nwt, line 747:
```suggestion
```

expr/boolean.go
```diff
@@ -411,62 +409,58 @@ func Contains(compare Boolean) Boolean {
...
-		primitive = astzed.Primitive{Kind: "Primitive", Type: "bstring", Text: primitive.Text}
+	if val.Type == zed.TypeString {
+		val = zed.NewValue(zed.TypeBstring, val.Bytes)
 	}
-	zv, err := zson.ParsePrimitive(primitive.Type, primitive.Text)
-	if err != nil {
-		return nil, err
-	}
-	switch zv.Type.(type) {
+	switch val.Type.(type) {
```
nwt, line 420: Should we use `zed.AliasOf(val.Type)` instead of `val.Type` here?

expr/boolean.go
```diff
@@ -411,62 +409,58 @@ func Contains(compare Boolean) Boolean {
...
+func Comparison(op string, val *zed.Value) (Boolean, error) {
 	// String literals inside Zed are parsed as zng bstrings
 	// (since bstrings can represent a wider range of values,
 	// specifically arrays of bytes that do not correspond to
 	// UTF-8 encoded strings).
-	if primitive.Type == "string" {
-		primitive = astzed.Primitive{Kind: "Primitive", Type: "bstring", Text: primitive.Text}
+	if val.Type == zed.TypeString {
+		val = zed.NewValue(zed.TypeBstring, val.Bytes)
 	}
```
nwt, line 419: `zed.TypeOfString` and `zed.TypeOfBstring` are handled by the same case in the switch statement below, so you can delete all of this.

expr/function/function.go
```diff
@@ -190,6 +192,38 @@ func (n *NameOf) Call(ctx zed.Allocator, args []zed.Value) *zed.Value {
 	return zed.Missing
 }
 
+// https://github.com/brimdata/zed/blob/main/docs/language/functions.md#typename
+type typeName struct {
+	zctx *zed.Context
+}
+
+func (t *typeName) Call(ectx zed.Allocator, args []zed.Value) *zed.Value {
+	if args[0].Type != zed.TypeString {
```
nwt, line 201:
```suggestion
	if zed.AliasOf(args[0].Type) != zed.TypeString {
```

expr/function/function.go
```diff
@@ -190,6 +192,38 @@ func (n *NameOf) Call(ctx zed.Allocator, args []zed.Value) *zed.Value {
...
+			return zed.Missing
+		}
+		return t.zctx.LookupTypeValue(typ)
+	}
+	// XXX second argument is a type value that wants to be named or
+	// if not a type value then we can take the type of the value.
+	//if len(args) == 2 {
+	//	return zctx.
+	//}
+	if args[1].Type != zed.TypeType {
```
nwt, line 217:
```suggestion
	if zed.AliasOf(args[1].Type) != zed.TypeType {
```

expr/function/function.go
```diff
@@ -190,6 +192,38 @@ func (n *NameOf) Call(ctx zed.Allocator, args []zed.Value) *zed.Value {
...
 }
 
+// https://github.com/brimdata/zed/blob/main/docs/language/functions.md#typename
+type typeName struct {
+	zctx *zed.Context
+}
+
+func (t *typeName) Call(ectx zed.Allocator, args []zed.Value) *zed.Value {
+	if args[0].Type != zed.TypeString {
+		return newErrorf(ectx, "typename: first argument not type string")
```
nwt, line 202:
```suggestion
		return newErrorf(ectx, "typename: first argument not a string")
```

expr/var.go
```diff
@@ -3,19 +3,16 @@ package expr
 import "github.com/brimdata/zed"
 
 type Var struct {
-	ref *zed.Value
+	slot int
 }
```
nwt, line 7: Or maybe `type Var int`?

compiler/kernel/filter.go
mccanne: On 2), I use this naming pattern because the `val!=nil` condition is a boolean indicator.  You get back nil if and only if it *is* not the thing being looed for so I think the word "is" is appropriate.

compiler/kernel/proc.go
mccanne: "Now" is always implied for any function call.  This name is meant to clarify that this is only called without a running flowgraph in expression context without this.  Maybe `Compile AndEvalExprWithoutRuntime` but I kind of prefer it how it is.

compiler/kernel/proc.go
mccanne: This error value is always ignored and the error conditions is used to detect an non-constant expression (i.e., not dependent on flowgraph/this).    Thus, it expects to typically panic on this == nil as all the expr code assumes this != nil.  The right way to make this better would be to carefully make sure there is a detectable difference between unrecoverable errors encountered here vs expected panics due to null this and for unrecoverable errors report them as suggested here.  This seems low priority.

compiler/semantic/scope.go
mccanne: I've used this forever... I think it's pretty common

https://www.acronymfinder.com/Top-of-the-Stack-(TOS).html
https://users.ece.cmu.edu/~koopman/stack_computers/sec3_2.html#:~:text=One%20of%20the%20data%20element,as%20used%20by%20the%20programmer.&text=This%20scheme%20allows%20using%20a,the%20top%20two%20stack%20elements.

compiler/semantic/scope.go
mccanne: It's an error so better to just take the error string.

compiler/kernel/bufferfilter.go
```diff
@@ -73,41 +80,41 @@ func CompileBufferFilter(e dag.Expr) (*expr.BufferFilter, error) {
...
 }
 
-func newBufferFilterForLiteral(l astzed.Primitive) (*expr.BufferFilter, error) {
-	switch l.Type {
-	case "bool", "byte", "int16", "uint16", "int32", "uint32", "int64", "uint64",
-		"float32", "float64", "time", "duration":
-		// These are all comparable, so they can require up to three
+func newBufferFilterForLiteral(val *zed.Value) (*expr.BufferFilter, error) {
+	if id := val.Type.ID(); zed.IsNumber(id) || id == zed.IDNull {
+		// All numbers are all comparable, so they can require up to three
```
nwt, line 112: This typo was in my suggestion. Sorry about that.
```suggestion
		// All numbers are comparable, so they can require up to three
```

compiler/kernel/proc.go
```diff
@@ -590,47 +585,22 @@ func (b *Builder) PushdownOf(trunk *dag.Trunk) (*Filter, error) {
...
-	e, err := compileExpr(zctx, scope, in)
+	e, err := compileExpr(zctx, in)
 	if err != nil {
 		return nil, err
 	}
+	// Catch panic as the runtime will panic if there is a
+	// reference to a var not in scope, a field access null this, etc.
+	defer func() {
+		if recover() != nil {
+			val = nil
```
nwt, line 600: Fine to do this but not necessary (`val` is initialized to nil and will remain nil if control reaches this point).
```suggestion
```

nwt (approved):
:shipit:!!!

Issue #3367, opened 2021-12-21
finish double-EOS protocol


mccanne:
Sort and procs that merge upstreams branches into a downstream branch (i.e., combine, merge, etc) all need to be updated to correctly implement the double EOS protocol.  Need good test coverage here.

mccanne:
Sort is merged #3417

mccanne:
Done!

PR #3417, opened 2021-12-30
adapt sort operator with double EOS protocol

mccanne:
This commit adds support to the sort operator for the double EOS
protocol and adds a test to sort the output of over.

We also restructured the sort logic so it's now "batch oriented"
instead of "run oriented" which will make adapting it for
scoped batches easier in a forthcoming PR.

This is part of issue #3367


proc/sort/sort.go
```diff
@@ -55,112 +56,140 @@ func (p *Proc) Done() {
...
+	var spiller *spill.MergeSort
+	var eof bool
 	var nbytes int
 	var out []zed.Value
 	for {
 		batch, err := p.parent.Pull()
 		if err != nil {
-			return nil, false, err
+			p.sendResult(nil, err)
+			return
```
nwt, line 69: If `spiller` is non-nil, this will return without calling `spiller.Cleanup`. But rather than insert a call here, I think you should simplify the protocol: Remove all the existing calls (including the one in `sendSpills`). Add one call immediately before `spiller = nil`. And add a deferred call immediately after `var spiller *spill.MergeSort`:
```go
	 defer func() {
	 	 if spiller != nil {
	 	 	 spiller.Cleanup()
		}
	}()
```

proc/sort/sort.go
```diff
@@ -55,112 +56,140 @@ func (p *Proc) Done() {
...
+	unseen := p.unseenFieldTracker.unseen()
+	vals := make([]zed.Value, 0, len(unseen))
+	for _, f := range unseen {
 		name, _ := expr.DotExprToString(f)
+		if name == "this" {
+			continue
+		}
 		e := fmt.Sprintf("warning: sort field %q not present in input", name)
-		a.Append(zed.NewValue(zed.TypeError, []byte(e)))
+		vals = append(vals, *zed.NewValue(zed.TypeError, []byte(e)))
```
nwt, line 187:
```go
		if name, _ := expr.DotExprToString(f); name != "this" {
			vals = append(vals, *zed.NewErrorf("warning: sort field %q not present in input", name))
		}
```

proc/sort/sort.go
mccanne: good idea!

mccanne: The cleanup in sendSpills is needed on success because otherwise the tmp files will never be deleted.  Maybe Cleanup() should be called Close()?

nwt: That's why I suggested adding one call immediately before `spiller = nil`. Keeping the call in `sendSpills` as you have makes the protocol less clear—readers have to examine both methods to see that `spiller.Cleanup` is called appropriately.

mccanne: Ok, fixed

nwt (approved):

Issue #3420, opened 2021-12-31
catchers everywhere


mccanne:
Add defer/recover calls everywhere needed in the operator data path wherever a new goroutine is created.  proc.Catcher is nice but we night want to lift the recovery point higher up in various cases.

We should audit the code and make sure we have coverage of this item.

PR #3421, opened 2022-01-01
update threaded operators to handle batching protocol

mccanne:
This commit has a number of changes to accommodate the batching protocol
in various threaded operators including switch, exprswitch, combine,
and merge.  For merge and combine, there's no need to detect double-EOS
but rather convey double-EOS through the fan-in of nodes where we need
to block paths that have sent EOS until all of the paths have sent EOS,
then resume all paths at the same time.  Doing this in combination with
a deadlock-free done protocol was a bit tricky, but we have clarified
the Done contract to say if an operator initiates a Done it must immediately
call Pull until receiving EOS then send exactly one EOS downstream and
resume as if initialized.

We added an explicit merge operator in the Zed language.  Previously, merge
was used only by the compiler on lakes but having a merge operator in
Zed is by itself useful but also makes for easier testing.  This will be
undocumented for now.

These changes exposed a bug in the "from" operator where a race between
EOF of the scanner and a Done coming upstream could result in a file
being erroneously closed twice.

We also added various catchers near the top of goroutines.  We might want
to make this even a bit tighter by adding the defer/recover calls right
at the point of goroutine creation.  See issue #3420.

### DIFF
--- a/proc/exprswitch/exprswitch.go
+++ b/proc/exprswitch/exprswitch.go
@@ -1,8 +1,6 @@
 package exprswitch
 
 import (
-	"sync"
-
 	"github.com/brimdata/zed"
 	"github.com/brimdata/zed/expr"
 	"github.com/brimdata/zed/proc"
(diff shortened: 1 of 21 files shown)


proc/router.go
```diff
@@ -0,0 +1,203 @@
+package proc
+
+import (
+	"sync"
+
+	"github.com/brimdata/zed/zbuf"
+)
+
+type Selector interface {
+	Forward(*Router, zbuf.Batch) error
```
nwt, line 10:
```suggestion
type Forwarder interface {
	Forward(*Router, zbuf.Batch) error
```

proc/split/split.go
```diff
@@ -14,113 +12,25 @@ import (
...
-	// Send SplitProc a request, then read the result. On EOS we send a nil
-	// request to let SplitProc know we're done, which will cause it to exit
-	// when it sees that all of us SplitChannels are gone.
-	p.request <- p.ch
-	result := <-p.ch
+var _ proc.Selector = (*Splitter)(nil)
 
-	if result.Batch == nil || result.Err != nil {
-		p.Done()
+func New(pctx *proc.Context, parent proc.Interface, n int) (*proc.Router, []proc.Interface) {
```
nwt, line 19: The `*proc.Router` return parameter never gets used, so maybe drop it.

proc/split/split.go
```diff
@@ -14,113 +12,25 @@ import (
...
-
-func send(flight []chan<- proc.Result, result proc.Result) {
-	for _, ch := range flight {
-		if result.Batch != nil {
-			result.Batch.Ref()
-		}
-		ch <- result
-	}
-}
+type Splitter []proc.Interface
```
nwt, line 15: This thing's only used in this package.
```suggestion
type splitter []proc.Interface
```

proc/router.go
nwt: Or, if you prefer Selector.
```suggestion
type Selector interface {
	Select(*Router, zbuf.Batch) error
```

proc/exprswitch/exprswitch.go
```diff
@@ -1,123 +1,84 @@
...
-
-func (p *Proc) Done() {
-	go func() {
-		for {
-			if _, ok := <-p.ch; !ok {
-				return
-			}
-		}
-	}()
+	return nil
```
nwt, line 83:
```suggestion
	batch.Unref()
	return nil
```

proc/head/head.go
```diff
@@ -36,13 +47,25 @@ func (p *Proc) Pull() (zbuf.Batch, error) {
...
+	for {
+		batch, err := p.parent.Pull()
+		if err != nil {
+			p.err = err
+			return
+		}
+		if batch == nil {
+			p.done = true
+			return
+		}
```
nwt, line 69:
```suggestion
		}
		batch.Unref()
```

proc/merge/merge.go
```diff
@@ -1,39 +1,274 @@
...
+	}
+}
+
+func (p *Proc) Pull() (zbuf.Batch, error) {
+	p.once.Do(p.run)
+	if p.err != nil {
+		return nil, p.err
+	}
+	if p.Len() == 0 {
+		// No more batches in head of line.  So, lets resume
```
nwt, line 57:
```suggestion
		// No more batches in head of line.  So, let's resume
```

proc/merge/merge.go
```diff
@@ -1,39 +1,274 @@
...
+	defer close(p.resultCh)
+	for {
+		batch, err := p.Pull()
+		select {
+		case p.resultCh <- proc.Result{batch, err}:
+			if err != nil {
+				return
+			}
+			if batch == nil {
+				if ok := p.resume(); !ok {
```
nwt, line 220:
```suggestion
				if !p.resume() {
```

proc/merge/merge.go
```diff
@@ -1,39 +1,274 @@
...
+					return
+				}
+			}
+		case <-p.doneCh:
+			// Drop the pending batch and initiate a done...
+			if ok := p.done(); !ok {
```
nwt, line 226:
```suggestion
			// Drop the pending batch and initiate a done.
			if !p.done() {
```

proc/merge/merge.go
```diff
@@ -1,39 +1,274 @@
...
+	u := p.hol[0]
+	val := &u.vals[0]
+	u.vals = u.vals[1:]
+	if len(u.vals) == 0 {
+		ok, err := p.replenish(u)
+		if err != nil {
+			return nil, err
+		}
+		if !ok {
+			heap.Pop(p)
```
nwt, line 126: My recollection is that these heap operations are spendy and avoiding unnecessary calls to `heap.Fix` in `zbuf.Merger.Read` made a measurable difference.
```suggestion
			heap.Pop(p)
			return val, nil
```

proc/router.go
```diff
@@ -0,0 +1,203 @@
...
+
+func (r *Router) run() {
+	defer func() {
+		// At double-EOS completion, we close all channels so the
+		// downstream routes will automatically get their double-EOS completions.
+		for _, ch := range r.routes {
+			close(ch)
+		}
+	}()
+	var EOS bool
```
nwt, line 57:
```suggestion
	var eos bool
```

proc/switcher/switch.go
```diff
@@ -1,169 +1,78 @@
...
-	for i, ch := range s.requests {
-		if results[i].Batch != nil {
-			results[i].Batch.Ref()
-			ch <- results[i]
-			s.requests[i] = nil
-			results[i].Batch = nil
+	// Send each case that has vals from this batch.
+	// We have vals that point into the current batch so we
+	// ref the batch for each outgoing new batch, then unref
+	// the batch once after the loop.
```
nwt, line 65: This comment reads like this function calls Unref but it's actually the router that calls it.
```suggestion
	// We have vals that point into the current batch so we
	// ref the batch for each outgoing new batch.
```

proc/exprswitch/exprswitch.go
```diff
@@ -1,123 +1,84 @@
...
-	} else {
-		for k, ch := range s.cases {
-			if ch == doneCh {
-				delete(s.cases, k)
-				break
-			}
+	// Send each case that has vals from this batch.
+	// We have vals that point into the current batch so we
+	// ref the batch for each outgoing new batch, then unref
+	// the batch once after the loop.
```
nwt, line 65: This comment reads like this function calls Unref but it's actually the router that calls it.
```suggestion
	// We have vals that point into the current batch so we
	// ref the batch for each outgoing new batch.
```

proc/router.go
```diff
@@ -0,0 +1,203 @@
...
+}
+
+func (r *route) Pull() (zbuf.Batch, error) {
+	r.router.once.Do(func() {
+		go r.router.run()
+	})
+	if result, ok := <-r.resultCh; ok {
+		return result.Batch, result.Err
+	}
+	// Router err protected by result channel
```
nwt, line 197: Nit: We usually put comments like this in the type declaration adjacent to the applicable field.

nwt (approved):

PR #3422, opened 2022-01-01
language improvements (drop semicolon, case keyword, match to search)

mccanne:
This commit makes a number of improvements to the language for
clarity and consistency. Semicolons are no longer used anywhere.
Instead, new instroduced keywords to provide needed disambiguation:
"pool" (inside of a multi-way from) and "case" (for switch cases).

The new keywords provide familiarity and while they increase typing
a bit, they are used in contexts where you would be "editing code"
compared to "typing searches".

We also changed the "match()" expression syntax to "search()" as this
is clearer in intent.  This change was made in combination with the
fact that swithc case expressions are now always expressions whereas
before one style was boolean search and the other style was expression.
This just created confusion for me.  So, now you assume expression
syntax and where you want boolean search, you just wrap the text
with search().


compiler/parser/parser.peg
```diff
@@ -63,67 +63,72 @@ Consts
...
+SplitLeg
+  = __ "=>" __ s:Sequential { RETURN(s) }
+
+SwitchLegs
+  = __ first:SwitchLeg rest:(_ leg:SwitchLeg { RETURN(leg) })* {
+    RETURN(PREPEND(first, rest))
+  }
+
+SwitchLeg
+  = expr:Case __ "=>" __ proc:Sequential {
```
nwt, line 108: You could use `SwitchLeg+` instead of `SwitchLegs` in the switch rules if you added a `__` prefix.
```suggestion
SwitchLeg
  = __ expr:Case __ "=>" __ proc:Sequential {
```

compiler/parser/parser.peg
```diff
@@ -63,67 +63,72 @@ Consts
...
+  = "case" _ expr:Expr { RETURN(expr) }
+  / "default" { RETURN(NULL) }
+
+FromLeg
+  = __ source:FromHead __  seq:Sequential {
+      RETURN(MAP("kind": "Trunk", "source": source, "seq": seq))
+    }
+  / __ source:FromSource {
+      RETURN(MAP("kind": "Trunk", "source": source, "seq": NULL))
     }
```
nwt, line 122: Maybe shorten?
```suggestion
  = __ source:FromHead seq:(__  s:Sequential { RETURN(s) })? {
      RETURN(MAP("kind": "Trunk", "source": source, "seq": seq))
    }
```

nwt (approved):

compiler/parser/parser.peg
```diff
@@ -63,67 +63,72 @@ Consts
...
   / f:Function &EndOfOp { RETURN(f) }
   / a:OpAssignment &EndOfOp { RETURN(a) }
   / a:Aggregation &EndOfOp  { RETURN(a) }
   / expr:SearchBoolean !AggGuard {
       RETURN(MAP("kind": "Filter", "expr": expr))
     }
 
-SwitchLiteralClause
-  = __ expr:(DefaultToken / Literal / Identifier) __ "=>" __ proc:Sequential __ ";" {
+EndOfOp = __ (Pipe / SearchKeywordGuard / "=>" / [);] / EOF)
```
mattnibs, line 96: Should we remove the `;` from here or keep this for backwards compatability?

mattnibs (approved):
This is great!

compiler/parser/parser.peg
mccanne: This changes the empty case to require "=>".  We could make this change but I think it would look strange.

nwt: Indeed. I missed that the first rule uses `FromHead` but the second uses `FromSource`. Sorry for the noise.

jameskerr:
Can you give an example of how you'd rewrite this query now?

```
"* | split ( => every 1h count(); => count() by id.resp_h;)"
```

PR #3424, opened 2022-01-03
Peeker: return EOF is cursor is empty after fill.

mattnibs:


nwt (approved):
Is this worth a unit test?

mattnibs:
> Is this worth a unit test?

Yeah probably...

PR #3425, opened 2022-01-04
Make build/install: build only zq and zed

mattnibs:
Change make build, install and create-release-assets to only build
zq and zed. Add builddev and installdev to install the other
development-centric commands (e.g., mockbrim, zc, zst, etc.).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ ARCH = "amd64"
 VERSION = $(shell git describe --tags --dirty --always)
 LDFLAGS = -s -X github.com/brimdata/zed/cli.Version=$(VERSION)
 MINIO_VERSION := 0.0.0-20201211152140-453ab257caf5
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -56,33 +57,42 @@ test-generate: generate
...
+output-check: builddev $(SAMPLEDATA)
 	scripts/output-check.sh
 
 # If the build recipe changes, please also change npm/build.
 build: $(PEG_DEP)
+	@mkdir -p dist
+	@go build -ldflags='$(LDFLAGS)' -o dist $(BUILD_COMMANDS)
+
+.PHONY: builddev
+builddev: $(PEG_DEP)
```
nwt, line 87: Do we really need both `build` and `builddev`? If not, I'd much prefer to leave `build` as is.

mattnibs: Sure

nwt (approved):

PR #3426, opened 2022-01-04
fix bug with joining record unions

mccanne:
This commit fixes a bug in the join logic to handle unions by
simply extracting the union-tagged value from the input before
processing.

### DIFF
--- a/proc/join/ztests/join-union.yaml
+++ b/proc/join/ztests/join-union.yaml
@@ -1,21 +1,18 @@
 script: |
-  zq -z 'split ( => sort a => sort b ) | inner join on a=b hit:=c | sort a' a.zson b.zson
-
+  zq -z 'inner join on a=b' a.zson b.zson
 inputs:
   - name: a.zson
     data: |
       {a:1}(({a:int64},{a:string}))
       {a:2}(({a:int64},{a:string}))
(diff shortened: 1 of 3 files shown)


proc/join/ztests/join-union.yaml
```diff
@@ -0,0 +1,21 @@
...
+    data: |
+      {b:1,c:"hit"}
+      {b:"bar",c:"hit"}
+      {b:3,c:"hit"}
+
+outputs:
+  - name: stdout
+    data: |
+      {a:1,hit:"hit"}
+      {a:"bar",hit:"hit"}
```
nwt, line 21: Simplify? (This triggers the same panic.)
```suggestion
script: |
  zq -z 'inner join on a=b' a.zson b.zson

inputs:
  - name: a.zson
    data: |
      {a:1}(({a:int64},{a:string}))
      {a:2}(({a:int64},{a:string}))
      {a:"bar"}(({a:int64},{a:string}))

  - name: b.zson
    data: |
      {b:1}
      {b:3}
      {b:"bar"}

outputs:
  - name: stdout
    data: |
      {a:1}
      {a:"bar"}
```

nwt (approved):

PR #3430, opened 2022-01-06
zbuf.PullerReader: Unref batch

mattnibs:
Not calling Batch.Unref when the PullerReader is done reading a batch
was causing batches to not get recycled. Fix this.

### DIFF
--- a/zbuf/batch.go
+++ b/zbuf/batch.go
@@ -122,6 +122,7 @@ type pullerReader struct {
 	p     Puller
 	batch Batch
 	idx   int
+	val   zed.Value
 }
 
 func (r *pullerReader) Read() (*zed.Value, error) {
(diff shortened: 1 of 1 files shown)


zbuf/batch.go
```diff
@@ -143,6 +143,7 @@ func (r *pullerReader) Read() (*zed.Value, error) {
 	rec := &vals[r.idx]
 	r.idx++
 	if r.idx == len(vals) {
+		r.batch.Unref()
```
nwt, line 146: Once `r.batch.Unref` is called, the storage referenced by both `rec` and `rec.Bytes` is subject to immediate reuse, so we need to copy both. We could do that with `rec = rec.Copy()` here, but it's more efficient to add a `val zed.Value` field to `pullerReader` and do this:
```suggestion
		r.val.CopyFrom(rec)
		rec = &r.val
		r.batch.Unref()
```

nwt (approved):
Good catch!

PR #3432, opened 2022-01-06
lake.Writer: Use CopyFrom to copy values

mattnibs:
lake.Writer was using value.Copy to copy a value bytes into the values
slice. This was resulting in an extra burden for allocations for every
added value, since a new byte slice had to be allocated. Instead use
value.CopyFrom which will attempt to use already allocated bytes slices.

### DIFF
--- a/lake/writer.go
+++ b/lake/writer.go
@@ -77,11 +77,7 @@ func (w *Writer) Write(rec *zed.Value) error {
 		}
 		return w.ctx.Err()
 	}
-	// XXX This call leads to a ton of one-off allocations that burden the GC
-	// and slow down import. We should instead copy the raw record bytes into a
-	// recycled buffer and keep around an array of ts + byte-slice structs for
-	// sorting.
-	w.vals = append(w.vals, *rec.Copy())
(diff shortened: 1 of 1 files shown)


lake/writer.go
```diff
@@ -91,6 +87,19 @@ func (w *Writer) Write(rec *zed.Value) error {
...
+func (w *Writer) append(val *zed.Value) {
+	n := len(w.vals)
+	if n >= cap(w.vals) {
+		newvals := make([]zed.Value, n, (n+1)*2)
+		for i, val := range w.vals {
+			newvals[i] = val
+		}
+		w.vals = newvals
+	}
+	w.vals = w.vals[:n+1]
```
nwt, line 99: I think we should just use `append` in these situations. It calls `runtime.growslice`, which does 2x for small slices and 1.25x for large slices.
```suggestion
	if n < cap(w.vals) {
		w.vals = w.vals[:n+1]
	} else {
		w.vals = append(w.vals, zed.Value)
	}
```

nwt (approved):

PR #3441, opened 2022-01-11
add scopes to dataflow model

mccanne:
This commit adds the concept of a scope to the dataflow model.  A scope
allows a subgraph in an inner scope to perform computation on values that
are derived from the sequence of values in the outer scope.  In this way,
the generated values run to completion in the scope and the scope yields
it results to the outer scope.  A scope can only exist in the context of
a traversal operator like "over" ("into" is coming soon).

A scope also allows the binding of variables (via a "let" operator)
to outer-scope values that appear in the inner scope as so named.
This is accomplished by modifying each batch that enters the scope
with the variables added by the scope.  Scopes can be nested and
each level of nesting pushes any new variables on top of any existing
scoped variables and popping them off the batch on exit from that scope.

To accommodate this new model, the double-EOS protocol is now restricted
to run only within a scope and batches that come out of a scope adhere
to a sinlge-EOS protocol.  We call each single-EOS-terminated sequence
of batches, terminated by a double-EOS, a platoon.

In making these changes, we realized we needed to simplify the semantics
of fanout operators (split/switch) and fanin operators (combined/merge)
while also eliminating the Done method on proc.Interface by having a
done Boolean passed as a parameter to Pull.  proc.Interface is now
entirely subsumed by zbuf.Puller.  EOS and Done events are now always
resolved by a go-channel select on an inbound done and an outbound write
thus resolving races between sending EOS downstream and propagating done
indications upstream.


compiler/parser/parser.peg
```diff
@@ -546,13 +547,30 @@ MergeProc
...
 OverProc
-  = "over"i _ exprs:Exprs {
-	  RETURN(MAP("kind":"Over", "exprs":exprs))
+  = ScopedOver
+  / "over"i _ exprs:Exprs {
+	  RETURN(MAP("kind":"Over", "exprs":exprs, "scope":NULL))
+    }
+
+ScopedOver
+  = "over"i _ exprs:Exprs __ scope:Scope {
```
nwt, line 556: Nit: Crop the Scope nonterminal since it doesn't appear to be used anywhere else.
```suggestion
  = "over"i _ exprs:Exprs __ "=>" __ "(" __ scope:Sequential __ ")" {
```

compiler/semantic/proc.go
```diff
@@ -558,41 +558,29 @@ func semProc(ctx context.Context, scope *Scope, p ast.Proc, adaptor proc.DataAda
...
+	case *ast.Let:
+		if p.Over == nil {
+			return nil, errors.New("let operator missing traversal in AST")
+		}
+		if p.Over.Scope == nil {
+			return nil, errors.New("let operator missing scope in AST")
+		}
+		scope.Enter()
+		defer scope.Exit()
+		locals, err := semLets(scope, p.Locals)
```
nwt, line 571: Should this be called `semLet` to agree with `ast.Let`?

expr/context.go
```diff
@@ -10,7 +10,7 @@ import (
 // of Zed values both for let-style temporary variables accessible via
 // the scope and for allocating results.
 type Context interface {
-	Scope() []zed.Value
+	Vars() []zed.Value
```
nwt, line 13: Maybe update the doc comment to make clear that the `Vars` method provides access to the scope?

zbuf/batch.go
```diff
@@ -67,10 +67,18 @@ func readBatch(zr zio.Reader, n int) (Batch, error) {
 	return NewArray(recs), nil
 }
 
-// A Puller produces Batches of records, signaling end-of-stream by returning
-// a nil Batch and nil error.
+// A Puller produces Batches of records, signaling end-of-stream (EOS) by returning
+// a nil Batch and nil error.  The done argument to Pull indicates that the stream
+// should be terminated before its natural EOS.  An implementation must return
+// EOS in response to a Pull call when the done parameter is true.  After seneind EOS,
```
nwt, line 73:
```suggestion
// EOS in response to a Pull call when the done parameter is true.  After seeing EOS,
```

zbuf/batch.go
```diff
@@ -67,10 +67,18 @@ func readBatch(zr zio.Reader, n int) (Batch, error) {
...
+// (either via done or its natural end), an implementation of an operator that
+// processes pulled data) should respond to additional calls to Pull as if restarting
```
nwt, line 75:
```suggestion
// processes pulled data should respond to additional calls to Pull as if restarting
```

zbuf/batch.go
```diff
@@ -67,10 +67,18 @@ func readBatch(zr zio.Reader, n int) (Batch, error) {
...
+// in its initial state.  For original sources of data (e.g., the from operator),
+// once EOS is reached, additional calls to Pull after the first EOS should always
+// return EOS.   Calls to Pull() are presumed to be single threaded
+// and multiple calls to Pull() cannot be issued concurrently.
```
nwt, line 79:
```suggestion
// return EOS.  Pull is not safe to call concurrently.
```

proc/yield/yield.go
```diff
@@ -39,13 +37,9 @@ func (p *Proc) Pull() (zbuf.Batch, error) {
 				recs = append(recs, *out.Copy())
 			}
 		}
-		batch.Unref()
+		defer batch.Unref()
 		if len(recs) > 0 {
-			return zbuf.NewArray(recs), nil
+			return zbuf.NewBatch(batch, recs), nil
 		}
```
nwt, line 43: Best to avoid deferring inside a loop like this since it will require allocations if the trip count is high.
```go
		if len(recs) > 0 {
			defer batch.Unref()
			return zbuf.NewBatch(batch, recs), nil
		}
		defer batch.Unref()
```

proc/traverse/ztests/nested-over-tail-head.yaml
```diff
@@ -0,0 +1,15 @@
+zed: |
+  over this => (
+    over this => (
+      tail 2 | head 1
+   )
+  )
```
nwt, line 6:
```suggestion
    )
  )
```

proc/traverse/ztests/nested-done-head.yaml
```diff
@@ -0,0 +1,22 @@
...
+  over this => (
+    over this => (
+      split (
+        => head 1 | a:=collect(this)
+        => head 2 | a:=collect(this)
+        => summarize a:=collect(this)
+      )
+      | sort this
+      | head 1
+    ) | sort this
```
nwt, line 11:
```suggestion
    )
    | sort this
```

proc/traverse/ztests/nested-done-tail.yaml
```diff
@@ -0,0 +1,26 @@
...
+  over this => (
+    over this => (
+      split (
+        => tail 1 | a:=collect(this)
+        => tail 2 | a:=collect(this)
+        => summarize a:=collect(this)
+      )
+      | sort this
+      | tail 2
+    ) | sort this
```
nwt, line 11:
```suggestion
    )
    | sort this
```

proc/traverse/ztests/nested.yaml
```diff
@@ -0,0 +1,23 @@
+zed: |
+  let TOP=top over outer => (
+      yield {msg:"OUTER", TOP:TOP, inner:inner, b:b}
+    | let BOTTOM=b over inner => (
+        yield {top:TOP, bottom:BOTTOM, elem:this}
+      )
+  )
```
nwt, line 7:
```suggestion
  let TOP=top over outer => (
    yield {msg:"OUTER", TOP:TOP, inner:inner, b:b}
    | let BOTTOM=b over inner => (
      yield {top:TOP, bottom:BOTTOM, elem:this}
    )
  )
```

proc/ztests/over-split-sort-outer.yaml
```diff
@@ -0,0 +1,26 @@
+zed: |
+  over a => (
+    split (
+      => head 1
+      => yield this+10
+    )
+  ) | sort this
```
nwt, line 7:
```suggestion
  over a => (
    split (
      => head 1
      => yield this+10
    )
  )
  | sort this
```

proc/ztests/over-split-sort.yaml
```diff
@@ -1,12 +1,14 @@
 zed: |
-  over a | split (
-    => head 1
-    => yield this+10
-  ) | sort this
+  over a => (
+    split (
+      => head 1
+      => yield this+10
+    ) | sort this
```
nwt, line 6:
```suggestion
    )
    | sort this
```

proc/groupby/groupby.go
```diff
@@ -154,95 +155,143 @@ func New(pctx *proc.Context, parent proc.Interface, keys []expr.Assignment, aggN
...
+			err = pullErr
+		}
+		if err != nil {
+			select {
+			case p.resultCh <- proc.Result{Err: err}:
+				return true, false
+			case <-p.pctx.Done():
+				return false, false
+			}
+		}
```
nwt, line 270: This doesn't feel right. If `err != nil` on the way in, shouldn't we send it on `p.resultCh` regardless of the value of `pullErr`?

proc/switcher/switch.go
```diff
@@ -70,8 +69,10 @@ func (s *Selector) Forward(router *proc.Router, batch zbuf.Batch) error {
 			batch.Ref()
 			out := zbuf.NewArray(c.vals)
 			c.vals = nil
-			router.Send(c.route, out, nil)
+			if ok := router.Send(c.route, out, nil); !ok {
```
nwt, line 72:
```suggestion
			if !router.Send(c.route, out, nil) {
```

proc/split/split.go
```diff
@@ -27,10 +27,12 @@ func New(pctx *proc.Context, parent proc.Interface, n int) []proc.Interface {
 }
 
 // Forward copies every batch to every output thus implementing split.
-func (s splitter) Forward(r *proc.Router, b zbuf.Batch) error {
+func (s splitter) Forward(r *proc.Router, b zbuf.Batch) bool {
 	for _, exit := range s {
 		b.Ref()
-		r.Send(exit, b, nil)
+		if ok := r.Send(exit, b, nil); !ok {
```
nwt, line 33:
```suggestion
		if !r.Send(exit, b, nil) {
```

proc/sort/sort.go
```diff
@@ -41,18 +41,25 @@ func New(pctx *proc.Context, parent proc.Interface, fields []expr.Evaluator, ord
...
+			}
+			if r.Batch == nil || r.Err != nil {
+				return nil, r.Err
+			}
+		}
+	}
 	if r, ok := <-p.resultCh; ok {
 		return r.Batch, r.Err
 	}
 	return nil, p.pctx.Err()
```
nwt, line 60: Maybe simplify?
```suggestion
	for {
		r, ok := <-p.resultCh
		if !ok {
			return nil, p.pctx.Err()
		}
		if !done || r.Batch == nil || r.Err != nil {
			return r.Batch, r.Err
		}
		r.Batch.Unref()
	}
```

proc/sort/sort.go
```diff
@@ -61,36 +68,42 @@ func (p *Proc) run() {
...
-			return
-		}
-		if batch == nil {
-			if eof {
+			if ok := p.sendResult(nil, err); !ok {
 				return
 			}
-			eof = true
+			continue
+		}
```
nwt, line 80: If `err != nil`, don't we want to return regardless of what happened in `p.sendResult`?

proc/combine/combine.go
```diff
@@ -12,139 +12,213 @@ import (
...
+		case <-p.doneCh:
+			if batch == nil {
+				// Combiner tells us we're done but we just
+				// received an EOS from upstream, so we don't want
+				// to call Pull(true) as they would break the contract.
+				// Since the combiner thinks we're done and our parent
+				// thinks we're done, there's nothing to do.
+				// Just continue the loop and reach for the next
+				// platoon.
+				if ok := p.wait(); !ok {
```
nwt, line 193: Nit: This pattern appears often in this change. I think we should simplify it everywhere.
```suggestion
				if !p.wait() {
```

proc/traverse/scope.go
```diff
@@ -0,0 +1,280 @@
...
+		if err != nil {
+			if ok := s.sendEOS(err); !ok {
+				return
+			}
+		}
+		if batch == nil {
+			if ok := s.sendEOS(nil); !ok {
+				return
+
+			}
```
nwt, line 76:
```suggestion
		if batch == nil || err != nil {
			if ok := s.sendEOS(err); !ok {
				return
			}
```

proc/traverse/over.go
```diff
@@ -1,85 +1,91 @@
...
 		o.batch = batch
-		o.vals = batch.Values()
+		o.outer = batch.Values()
 	}
-	if o.eof {
-		o.eof = false
+	if done {
+		o.outer = nil
 		return nil, nil
 	}
```
nwt, line 45: When called with `done` equal to true, shouldn't `Pull` always call `o.parent.Pull(true)`?

My understanding is that it should, so I was expecting this done/outer prelude to look like this:
```go
	if done {
		o.outer = nil
		return o.parent.Pull(true)
	}
	if len(o.outer) == 0 {
		batch, err := o.parent.Pull(false)
		if batch == nil || err != nil {
			return nil, err
		}
		o.batch = batch
		o.outer = batch.Values()
	}
```

compiler/semantic/proc.go
mccanne: I don't understand...

expr/context.go
mccanne: Yes! Definitely a good idea to leave a meaningful comment here as this somewhat complicated interface evolves.

proc/traverse/ztests/over-split-head.yaml
```diff
@@ -0,0 +1,7 @@
+zed: over this => (split (=> head 1))
+
+input: |
+  1
+
+output: |
+  1
```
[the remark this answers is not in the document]
mccanne: I don't understand.  Having the two test for different conditions seems good.

proc/yield/yield.go
nwt: I looked a bit at the implementation and [`defer` in a loop requires an allocation regardless of trip count](https://github.com/golang/go/blob/2580d0e08d5e9f979b943758d3c49877fb2324cb/src/runtime/panic.go#L14-L19).

proc/traverse/ztests/over-split-head.yaml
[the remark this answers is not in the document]
nwt: It's certainly good to test the different conditions, but given that both tests files have the same Zed program (viz., `over this => (split (=> head 1))`, aggregate code coverage is the same if you combine them, so we might as well.

proc/sort/sort.go
mccanne: let's discuss.  contract is you keep going and the propagating error cancels context

proc/combine/combine.go
mccanne: The reason I like my pattern is that the wait return is that things aren't ok vs the wait is over.  There are two things going on.  But I will change it to your pattern since its shorter and you can see two things at once!

compiler/parser/parser.peg
```diff
@@ -546,13 +547,30 @@ MergeProc
...
+    }
+
+Scope = "=>" __ "(" __ seq:Sequential __ ")" { RETURN(seq) }
+
+LetAssignments
+  = first:LetAssignment rest:(__ "," __ a:LetAssignment { RETURN(a) })* {
+      RETURN(PREPEND(first, rest))
+    }
+
+LetAssignment = id:IdentifierName __ "=" __ expr:Expr {
```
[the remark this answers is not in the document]
mccanne: `:=` means put / mutation

proc/groupby/groupby.go
```diff
@@ -303,14 +352,14 @@ func (a *Aggregator) Consume(ectx expr.Context, this *zed.Value) {
...
+		row.reducers.consumeAsPartial(this, a.aggRefs, batch)
 	} else {
-		row.reducers.apply(ectx, a.aggs, this)
+		row.reducers.apply(batch, a.aggs, this)
 	}
 }
 
-func (a *Aggregator) spillTable(eof bool, ectx expr.Context) error {
-	batch, err := a.readTable(true, true)
+func (a *Aggregator) spillTable(eof bool, ref zbuf.Batch) error {
```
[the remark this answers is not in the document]
mccanne: This change propagates... let's fix later.

proc/yield/yield.go
nwt: It is the same. 🤦

Here's what I meant to suggest:
```go
		if len(recs) > 0 {
			defer batch.Unref()
			return zbuf.NewBatch(batch, recs), nil
		}
		batch.Unref()
```

compiler/semantic/proc.go
nwt: I fully admit that this is marginal territory, but `semLets` feels a little off to me because it's plural while `ast.Let` is singular. But after another look at the code, `semLet` doesn't feel right since it doesn't have an `ast.Let` parameter.

Given the similarity to `semConsts`, how about calling it `semVars`? (They're identical except that the latter calls `scope.DefineVar` instead of `scope.DefineConst`.)

proc/sort/sort.go
mccanne: as we discussed...

proc/groupby/groupby.go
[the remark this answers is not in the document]
mccanne: As we discussed, as we improved the batch/pool/buffering model these code paths will be updated and opportunities will arise to tighten it up...

proc/merge/merge.go
```diff
@@ -173,43 +170,42 @@ func (m *Proc) Pop() interface{} {
 }
 
 type puller struct {
```
nwt, line 172: As in `proc/combine`, I think we can simplify all of this if we change things around so this type is just a `zbuf.Puller` that wraps its upstream/parent in a goroutine, but let's worry about that later.

nwt (approved):

proc/combine/combine.go
```diff
@@ -12,139 +12,213 @@ import (
...
+		goto drain
+	default:
+	}
+	// Now that everyone is blocked either because they sent us an EOS,
+	// we sent them a done, or and EOS/done collided at the same time,
+	// we can unblock everything.
+	return p.unwait()
+}
+
+type puller struct {
```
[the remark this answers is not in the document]
mccanne: issue #3453

proc/merge/merge.go
mccanne: issue #3453

Issue #2213, opened 2021-02-24
bucketing time and durations, duration syntax


mccanne:
We should have more flexible ways to bucket time.  The `trunc` function should be renamed to `bin` or `bucket` or something else.

With zeek logs, I should be able to say 
```
_path=conn | count() by bucket(duration, 10m) 
```

The expected output of this would be a table with contents like:
* A count of connections with 0 <= duration < 10m
* A count of connections with 10m <= duration < 20m
* etc.

Currently, the semantic analyzer changes group-by "every" clauses a group-by key of ts=trunc(every-dur).

As pointed out by @nwt , "every" is an unusual pattern.  Instead, we could align the Z syntax with common practice by changing every from a group-by clause to a shortcut for 
```
ts = bucket(ts, span)
```
where `span` is the every duration.  This way you could say 
```
count() by every(1m)
```
We should also adopt RFC 3339 syntax for duration strings as is implemented in the golang time package. 

So... instead of
```
_path=conn | every 10min count()
```
you would say 
```
_path=conn | count() by every(10m)
```

mccanne:
The duration syntax is in issue #2214

PR #3450, opened 2022-01-14
Rename function trunc to bucket

mattnibs:
This change also makes it so if first argument is a duration, the
duration type is kept and is truncated by the second duration argument.

Part of #2213

### DIFF
--- a/expr/function/time.go
+++ b/expr/function/time.go
@@ -36,7 +36,7 @@ func (b *Bucket) Call(ctx zed.Allocator, args []zed.Value) *zed.Value {
 		}
 		bin = nano.Duration(d) * nano.Second
 	}
-	if tsArg.Type == zed.TypeDuration {
+	if zed.TypeUnder(tsArg.Type) == zed.TypeDuration {
 		dur, err := zed.DecodeDuration(tsArg.Bytes)
 		if err != nil {
 			panic(err)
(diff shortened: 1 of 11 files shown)


expr/function/time.go
```diff
@@ -38,9 +32,20 @@ func (t *Trunc) Call(ctx zed.Allocator, args []zed.Value) *zed.Value {
 	} else {
 		d, ok := coerce.ToInt(binArg)
 		if !ok {
-			return newErrorf(ctx, "trunc: second arg must be duration or number")
+			return newErrorf(ctx, "bucket: second arg must be duration or number")
 		}
 		bin = nano.Duration(d) * nano.Second
 	}
-	return newTime(ctx, nano.Ts(ts.Trunc(bin)))
+	if tsArg.Type == zed.TypeDuration {
```
nwt, line 39:
```suggestion
	if zed.TypeUnder(tsArg.Type) == zed.TypeDuration {
```

nwt (approved):
That branch name!

mattnibs:
> That branch name!

Sometimes you need to have a little fun 😄

expr/function/time.go
mattnibs: Ugh, I should have known this...

PR #3451, opened 2022-01-14
Change group by every syntax

### LINKED ISSUE #2213: bucketing time and durations, duration syntax

mattnibs:
Remove the current "every [duration]" syntax for the summarize proc and
instead add function "every" that can be used as a key in summarize.
This change also keeps intact the optimizations done on every queries by
looking for an every call in a summarize proc's keys.

Closes #2213


compiler/kernel/expr.go
```diff
@@ -350,13 +350,14 @@ func compileCall(zctx *zed.Context, call dag.Call) (expr.Evaluator, error) {
 		return compileShaper(zctx, call)
 	}
 	nargs := len(call.Args)
-	fn, this, err := function.New(zctx, call.Name, nargs)
+	fn, wantPath, err := function.New(zctx, call.Name, nargs)
```
nwt, line 353: Nit: `wantPath` reads like a Boolean. I'd call this `path` (and call the `dag.Path` below `dagPath`).

expr/function/function.go
```diff
@@ -20,18 +21,21 @@ type Interface interface {
 	Call(zed.Allocator, []zed.Value) *zed.Value
 }
 
-func New(zctx *zed.Context, name string, narg int) (Interface, bool, error) {
+func New(zctx *zed.Context, name string, narg int) (Interface, field.Path, error) {
 	argmin := 1
 	argmax := 1
-	var wantThis bool
+	var wantPath field.Path
```
nwt, line 27: Nit: `wantPath` reads like a Boolean. I'd call this `path`.

compiler/optimizer/optimizer.go
```diff
@@ -95,10 +95,10 @@ func (o *Optimizer) propagateScanOrder(op dag.Op, parent order.Layout) (order.La
 	case *dag.Summarize:
 		//XXX handle only primary key for now
 		key := parent.Primary()
-		if len(op.Keys) > 0 {
-			groupByKey := fieldOf(op.Keys[0].LHS)
+		for _, a := range op.Keys {
+			groupByKey := fieldOf(a.LHS)
 			if groupByKey.Equal(key) {
```
nwt, line 100:
```suggestion
			if groupByKey := fieldOf(a.LHS); groupByKey.Equal(key) {
```

compiler/optimizer/optimizer.go
```diff
@@ -95,10 +95,10 @@ func (o *Optimizer) propagateScanOrder(op dag.Op, parent order.Layout) (order.La
 	case *dag.Summarize:
 		//XXX handle only primary key for now
 		key := parent.Primary()
-		if len(op.Keys) > 0 {
-			groupByKey := fieldOf(op.Keys[0].LHS)
+		for _, a := range op.Keys {
```
nwt, line 98:
```suggestion
		for _, k := range op.Keys {
```

expr/function/time.go
```diff
@@ -45,7 +47,14 @@ func (b *Bucket) Call(ctx zed.Allocator, args []zed.Value) *zed.Value {
...
+		return newErrorf(ctx, "%s: time arg required", b)
 	}
 	return newTime(ctx, ts.Trunc(bin))
 }
+
+func (t *Bucket) String() string {
+	if t.name == "" {
+		return "bucket"
+	}
+	return t.name
```
nwt, line 59:
```suggestion
func (b *Bucket) String() string {
	if b.name == "" {
		return "bucket"
	}
	return b.name
```

nwt (approved):

PR #3454, opened 2022-01-17
update ZSON string escaping, delete bstring, simplify zeekio

### LINKED ISSUE #3041: Recognize \u escape sequences in ZSON strings
The ZSON parser does not recognize the `\u` escape sequences for strings [described in the ZSON specification](https://github.com/brimdata/zed/blob/main/docs/formats/zson.md#331-string-escape-rules).

mccanne:
As part of the final updates to the Zed formats, this commit
gets rid of the bstring type and updates the ZSON escaping
rules.  We now convert Zeek TSV strings, with their \x escapes,
into normal UTF-8 Zed strings.  Where an \x escape results in
illegal code points, we escape the the backslash before the x
just as the Zeek does for JSON logs.

Also, with this change it became clear that tzngio was no longer
needed and was consolidated into zeekio.  tableio and textio now
use zeekio.FormatValue().

Closes #3041


zio/zeekio/builder.go
```diff
@@ -166,11 +167,31 @@ func (b *builder) appendPrimitive(typ zed.Type, val []byte) error {
 		b.buf = zed.AppendBool(b.buf[:0], v)
 	case zed.IDString:
 		// Zeek's enum type is aliased to string.
+		val = unescapeZeekString(val)
+		if !utf8.Valid(val) {
+			// Zeek has an unusual escaping model for non-valid UTF
+			// strings in their JSON integration: invalid bytes are
+			// formatted as the sequence '\' 'x' h h to indicate
```
mattnibs, line 174:
```suggestion
			// formatted as the sequence '\' 'x' to indicate
```

zio/zeekio/builder.go
```diff
@@ -166,11 +167,31 @@ func (b *builder) appendPrimitive(typ zed.Type, val []byte) error {
...
+			// the presence of unexpected, invalid binary data where
+			// a string was expeceted, e.g., in a field of data coming
+			// of the network.  This is a reasonable scheme; however,
```
mattnibs, line 177:
```suggestion
			// off the network.  This is a reasonable scheme; however,
```

zio/zeekio/escape.go
```diff
@@ -0,0 +1,104 @@
...
+
+import (
+	"bytes"
+	"unicode"
+	"unicode/utf8"
+
+	"github.com/brimdata/zed"
+)
+
+// ShouldEscape determines if the given code point at the given position
```
mattnibs, line 11:
```suggestion
// shouldEscape determines if the given code point at the given position
```

zio/zeekio/escape.go
```diff
@@ -0,0 +1,104 @@
...
+// in a value should be escaped for the given output format.  This function
+// does not account for unprintable characters, its main purpose is to
+// centralize the logic about which characters are syntatically significant
+// in each output format and hence must be escaped.  The inContainer parameter
+// specifies whether this values is inside a set or vector (which is needed
+// to correctly implement  zeek log escaping rules).
```
mattnibs, line 17:
```suggestion
// to correctly implement zeek log escaping rules).
```

zson/lexer.go
```diff
@@ -300,6 +318,123 @@ func (l *Lexer) scanString() (string, error) {
...
+					k += 6
+					if dec := utf16.DecodeRune(r, r2); dec != unicode.ReplacementChar {
+						// A valid pair; consume.
+						if _, err := b.WriteRune(dec); err != nil {
+							return "", err
+						}
+					}
+					//XXX
+					// Invalid surrogate; fall back to replacement rune.
+					//rr = unicode.ReplacementChar
```
mattnibs, line 399: Should this just return an error?
```suggestion
					return "", errors.New("invalid utf16 surrogate pair")
```

mattnibs (approved):

proc/join/ztests/comparable-types.yaml
```diff
@@ -1,20 +0,0 @@
...
-    data: |
-      {a:"hello",va:"value a"}
-  - name: b.zson
-    data: |
-      {b:"hello"(bstring),vb:"value b"}
-
-outputs:
-  - name: stdout
-    data: |
-      {a:"hello",va:"value a",vb:"value b"}
```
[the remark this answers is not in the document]
mccanne: This is covered by the other tests.  The test I took out was making sure join on string=bstring worked too but there is no longer a bstring.

zio/zeekio/builder.go
mccanne: There is probably a better way to say this but the point is subtle and I think dropping the salient piece of the encoding only makes it more confusing.

zson/lexer.go
mccanne: Yes, I made an issue.

zio/zeekio/escape_test.go
```diff
@@ -0,0 +1,32 @@
+package zeekio
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+//XXX move back to package zng
+func TestUnescapeBstring(t *testing.T) {
```
nwt, line 10:
```suggestion
func TestUnescapeZeekString(t *testing.T) {
```

zio/zeekio/ztests/zeek-utf8-json.yaml
```diff
@@ -0,0 +1,80 @@
+zed: "*"
+
+# These test inputs are from src/util.cc in the Zeek source code as of Jan 2022.
+# See https://github.com/zeek/zeek/blob/3d9d6e953be5ae8b72336031b6e6659222f52cea/src/util.cc#L2521
+
+# XXX PR question: do we want the funny chars to be u-escaped?  also, our code
```
nwt, line 6: If you're asking whether we want to u-escape characters whose Go rune value exceeds `utf.RuneSelf`, then I think the answer is no.

zio/zeekio/ztests/zeek-utf8-json.yaml
```diff
@@ -0,0 +1,80 @@
...
+# differs from Zeek on "Private Use Area (E000-F8FF) are always invalid"... "\xee\x8b\xa0
```
nwt, line 7: I think this difference is fine.

zio/zeekio/ztests/zeek-utf8-json.yaml
```diff
@@ -0,0 +1,80 @@
+zed: "*"
+
+# These test inputs are from src/util.cc in the Zeek source code as of Jan 2022.
+# See https://github.com/zeek/zeek/blob/3d9d6e953be5ae8b72336031b6e6659222f52cea/src/util.cc#L2521
```
nwt, line 4: These are great tests, and I'm happy to have them here if that's your preference, but I do think they'd be more readable if moved to `escape_test.go`.

nwt (approved):

zio/zeekio/ztests/zeek-utf8-json.yaml
mccanne: As discussed, we can move them later or add another column to the Zeek TSV values as a comment explaining each test (taking the comments from the respective tests in the Zeek source code)

PR #3458, opened 2022-01-17
add structured errors, change ZSON enum syntax

### LINKED ISSUE #3320: structured errors as complex types
We should implement errors as complex types so users can produce richer errors and errors can be stacked/wrapped.

The proposal is that an error will be a complex type with type signature 
```
error<type> 
```
where `type` is any Zed type.

Error values are created simply by casting a value to an error, e.g., 
```
error({msg: "my error message", value:this})
```
This error value would have type 
```
error<{msg:string,value:typeOfThis>}
```
were `typeOfThis` is the type signature of `this`, say for example,
```
error<{msg:string,value:{foo:string,bar:int64}}>
```

Note that this is compatible with the current syntax of casting strings to the primitiive error type, e.g., 
```
error("my error message")
```

Another motivation is in shaping, e.g., 
```
switch Country (
 "US"  => do something for US ;
 "FR" => do something  for FR;
  null => yield error({err:"no country code", value:this})
  default => yield error({err:"unknown country: ${Country}", value:this})
) | ...
```
Note `yield` is not yet implemented nor are template literals (i.e., `${this.Country}`) but they are coming soon.  `yield x` is shorthand for `put this:=x`

Note that for a multi-stage ETL pipeline, error values would simply propagate and get wrapped in each other making debugging far easier than navigate and correlated errors messages from different logs in different stages.

e.g., suppose this downstream logic 
```
const shape1 = type({ ...})
const shape2 = type({ ...})
switch typeof(this) (
  shape1 => ... ;
  shape2 => ... ;
  default => yield error({msg:"unknown shape", value:this});
) | ...
```
received the value 
```
error({err:"no country code", value:{foo:"hello",bar:1})
```
Then it would yield the following wrapped error...
```
error({
  err:"unknown shape",
  value:error({
    err:"no country code",
    value:{foo:"hello",bar:1}
  })
})
```
So, if this error value ultimately landed in an output pool, then a data engineer could easily pinpoint where the error began and how it propagated by searching for errors using Zed queries.

In practice, you would want to include the pool key (e.g., timestamp) or maybe just a ts field in your error set to "now()", but this is up to the developer designing the error patterns.

mccanne:
This commit adds structured errors to Zed as part of the Dec 2021
changes to the Zed formats.  Since a parser ambiguity arose that was
solved by the new enum syntax, we rolled the enum changes into this PR
as well.

Also bstring and stringy errors are now gone, so we got rid of the IsStringy
pattern and replaced it everywhere with a compare to type string.

The diffs are somewhat verbose because the change requires that errors
are now allocated from a type context so many more entities now need a zctx.

Closes #3320


index/keys.go
```diff
@@ -44,7 +44,7 @@ func (k *Keyer) KeyOf(ectx expr.Context, val *zed.Value) (*zed.Value, error) {
 	}
 	if _, ok := k.valid[key.Type]; !ok {
 		for _, col := range recType.Columns {
-			if col.Type == zed.TypeError {
+			if _, ok := col.Type.(*zed.TypeError); ok {
```
nwt, line 47:
```suggestion
			if _, ok := zed.TypeUnder(col.Type).(*zed.TypeError); ok {
```

nwt (approved):

PR #3469, opened 2022-01-20
fix bug that confused ZNG error and JSON error in query end point

mccanne:
This commit fixes a bug that is tripped up by the ZNG 2022 format
changes where the query handler could return a ZNG-encoded error
instead of a JSON error but the client expected JSON.  Also, it was
hard to track down because of the vague error message, so the
error messages are improved here too.

### DIFF
--- a/api/queryio/client.go
+++ b/api/queryio/client.go
@@ -32,11 +32,11 @@ func (q *Query) Read() (*zed.Value, error) {
 		}
 		value, err := zson.ParseValue(zed.NewContext(), string(ctrl.Bytes))
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("client query: unable to parse Zed control message: %w (%s)", err, string(ctrl.Bytes))
 		}
 		var v interface{}
 		if err := unmarshaler.Unmarshal(value, &v); err != nil {
-			return nil, err
+			return nil, fmt.Errorf("client query: unable to unmarshal Zed control message: %w (%s)", err, string(ctrl.Bytes))
 		}
 		return nil, controlToError(v)
 	}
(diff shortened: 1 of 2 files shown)


api/queryio/client.go
```diff
@@ -32,11 +32,11 @@ func (q *Query) Read() (*zed.Value, error) {
 		}
 		value, err := zson.ParseValue(zed.NewContext(), string(ctrl.Bytes))
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("client query: unable to parse Zed control message: %w (%s)", err, string(ctrl.Bytes))
 		}
 		var v interface{}
 		if err := unmarshaler.Unmarshal(value, &v); err != nil {
-			return nil, err
+			return nil, fmt.Errorf("client query: unable to unmarshal Zed control message: %w (%s)", err, string(ctrl.Bytes))
```
nwt, line 39: Nit: I'd leave off the "client query: " prefix.

service/handlers.go
```diff
@@ -38,18 +47,21 @@ func handleQuery(c *Core, w *ResponseWriter, r *Request) {
...
-	flowgraph, err := runtime.NewQueryOnLake(r.Context(), zed.NewContext(), query, c.root, &req.Head, r.Logger)
+	flusher, _ := w.ResponseWriter.(http.Flusher)
+	writer, err := queryio.NewWriter(zio.NopCloser(w), format, flusher)
 	if err != nil {
 		w.Error(err)
 		return
 	}
+	// Once we defer writer.Close() are going to write ZNG to the HTTP
+	// response body and for errors after this point, we must call
+	// writer.WriterError() isntead of w.Error().
```
nwt, line 63:
```suggestion
	// writer.WriterError() instead of w.Error().
```

nwt (approved):

PR #3477, opened 2022-01-21
new version of ZNG

mccanne:
This commit updates the ZNG protocol specification and implementation
to conform to the December 2021 upadtes to Zed along with a better way
to frame payloads.  This new approach will make it easy to add
parallelism to the output compression path.

As part of the refactor, we create a synchronous ZNG scanner for
test and debug alongside the parallel scanner.  There is also a
new input flag (-threads) to specify the number of threads used
by the zng scanner (default of 0 uses go maxprocs and setting it
to 1 forces the synchronous scanner).

On no output, the ZNG writer emits now outputs nothing instead of a
single-char EOS.

We also added a new dev command zngdump that prints metadata
about serialized ZNG frames.  This will come in handy down the
road as we explore performance and how different types of ZNG
impact perf in different ways.

This also fixes the null value bug (echo "null(int64)" | zq - | zq -).

Co-authored-by: Noah Treuhaft <noah.treuhaft@gmail.com>


cli/inputflags/flags.go
```diff
@@ -28,6 +29,7 @@ func (f *Flags) Options() anyio.ReaderOpts {
 func (f *Flags) SetFlags(fs *flag.FlagSet, validate bool) {
 	fs.StringVar(&f.Format, "i", "auto", "format of input data [auto,zng,zst,json,ndjson,zeek,zjson,csv,parquet]")
 	fs.BoolVar(&f.ZNG.Validate, "validate", validate, "validate the input format when reading ZNG streams")
+	fs.IntVar(&f.ZNG.Threads, "threads", 0, "number of threads used for scanning ZNG input")
```
nwt, line 32: Nits: I don't think this knob is worth its weight. But if we keep it around, I think the field should be `Goroutines` and the flag should be `zng.goroutines`.

cmd/zed/dev/zngdump/command.go
```diff
@@ -0,0 +1,180 @@
...
+		f, err = os.Open(fname)
+		if err != nil {
+			return err
+		}
+	}
+	if c.from != 0 || c.to != 0 {
+		chunk := io.NewSectionReader(f, int64(c.from), int64(c.to)-int64(c.from))
+		_, err := io.Copy(os.Stdout, chunk)
+		return err
+	}
```
nwt, line 61: This is weird and doesn't feel like it belongs here. I expected -from and -to to determine the range fed to the reader below, so maybe do that instead?

cmd/zed/dev/zngdump/command.go
```diff
@@ -0,0 +1,180 @@
...
+		if err != nil {
+			return noEOF(err)
+		}
+		fmt.Printf("% 10d ", pos)
+		if code == 0xff {
+			fmt.Println("ff EOS")
+			continue
+		}
+		if (code & 0x80) != 0 {
+			return errors.New("zngio: encountered wrong version bit in framing")
```
nwt, line 77:
```suggestion
			return errors.New("encountered wrong version bit in ZNG framing")
```

cmd/zed/dev/zngdump/command.go
```diff
@@ -0,0 +1,180 @@
...
+			return fmt.Errorf("unknown ZNG message block type: %d", typ)
+		}
+		if (code & 0x40) != 0 {
+			format, size, zlen, err := r.readComp(code)
+			if err != nil {
+				return noEOF(err)
+			}
+			fmt.Printf("comp fmt %d zlen %d size %d (net %d)\n", format, zlen, size, size-zlen)
+		} else {
+			size, err := r.readUncomp(code)
```
nwt, line 96: Nit: I think calling these methods readCompressed and readUncompressed would make what's going on a little clearer.

cmd/zed/dev/zngdump/command.go
```diff
@@ -0,0 +1,180 @@
...
+
+func (r *reader) decodeLength(code byte) (int, error) {
+	v, err := r.uvarint()
+	if err != nil {
+		return 0, err
+	}
+	return (v << 4) | (int(code) & 0xf), nil
+}
+
+func (r *reader) uvarint() (int, error) {
```
nwt, line 170: NIts: These methods read from r but the names don't communicate that. I'd call them readLength and readUvarint.

docs/data-model/zng.md
```diff
@@ -61,69 +56,54 @@ or otherwise rewritten to be merged in this fashion.
...
-| `0xfe` | application-defined message    |
-| `0xff` | end-of-stream                  |
-
-The application-defined messages are available to higher-layer protocols and
-potential future variations of ZNG.  A ZNG implementation that
-merely skips over all of the application-defined messages is guaranteed by
-this specification to decode all of the data as described herein even if such
-messages provide additional semantics on top of the base ZNG format.
+A ZNG stream comprises a sequence of interleaved frames of data
+where a frame contains one of three types of data:
```
nwt, line 60:
```suggestion
A ZNG stream comprises a sequence of frames where
each frame contains one of three types of data:
```

docs/data-model/zng.md
```diff
@@ -61,69 +56,54 @@ or otherwise rewritten to be merged in this fashion.
...
 
-The type ID is encoded as a `uvarint`, an encoding used throughout the ZNG format.
+Bits 5 and 4 indicate the frame type:
+values (`00`), typedefs (`01`), control (`10`),
+or end of stream (`11`).
+
+The length of the frame body exclusive of the code and frame length
+is encoded in part as a variable length unsigned integer referred
+to here as a `uvarint`.  The frame length is equal to the uvarint `len`
+times 8 plus the low 4-bit integer value of the `code`.
```
nwt, line 106: I don't feel strongly about this, but I do favor simplifying the specification and implementation by reserving the four extra code bits rather than adding them to the frame length. I think it would be a reasonable move since the decrease in entropy seems negligible, particularly for large frames. But again, this isn't something I feel strongly about!

docs/data-model/zng.md
```diff
@@ -61,69 +56,54 @@ or otherwise rewritten to be merged in this fashion.
...
-> uses application-defined message `0xfe` to embed search and server stats in
-> the return stream of ZNG data, e.g., as a long-running search progresses on
-> the server.
+Each frame begins with its version bit and a code indicating
+whether it is compressed or not:
+
+|    code    |  len    | body            |
+|------------|---------|-----------------|
+| `00xxxxxx` | uvarint | uncompressed data |
+| `01xxxxxx` | uvarint | compressed data |
```
nwt, line 74: Maybe add an RFC-style figure describing the code byte, along these lines?
```
    7 6 5 4 5 3 1 0 
   +-+-+-+-+-+-+-+-+
   |V|C|  T|      L|
   +-+-+-+-+-+-+-+-+

   V: 1 bit

     Version number.  Must be zero.

   C: 1 bit

     Indicates compressed frame data.

   T: 2 bits

     Type of frame data.

       00: Types
       01: Values
       10: Control
       11: End of stream

   L: 4 bits

     High-order bits of frame length.
```

docs/data-model/zng.md
mccanne: What would we reserve them for?  The version bit controls any future meaning of those bits.

expr/coerce/promote.go
```diff
@@ -0,0 +1,41 @@
...
+	zed.IDInt64,      // IDTime       = 13
+	zed.IDFloat16,    // IDFloat16    = 14
+	zed.IDFloat32,    // IDFloat32    = 15
+	zed.IDFloat64,    // IDFloat64    = 16
+	zed.IDFloat128,   // IDFloat64    = 17
+	zed.IDFloat256,   // IDFloat64    = 18
+	zed.IDDecimal32,  // IDDecimal32  = 19
+	zed.IDDecimal64,  // IDDecimal64  = 20
+	zed.IDDecimal128, // IDDecimal128 = 21
+	zed.IDDecimal256, // IDDecimal256 = 22
```
nwt, line 30: You can use the constants as keys.
```suggestion
	zed.IDUint8:      zed.IDInt8,
	zed.IDUint16:     zed.IDInt16,
	zed.IDUint32:     zed.IDInt32,
	zed.IDUint64:     zed.IDInt64,
	zed.IDUint128:    zed.IDInt128,
	zed.IDUint256:    zed.IDInt256,
	zed.IDInt8:       zed.IDInt8,
	zed.IDInt16:      zed.IDInt16,
	zed.IDInt32:      zed.IDInt32,
	zed.IDInt64:      zed.IDInt64,
	zed.IDInt128:     zed.IDInt128,
	zed.IDInt256:     zed.IDInt256,
	zed.IDDuration:   zed.IDInt64,
	zed.IDTime:       zed.IDInt64,
	zed.IDFloat16:    zed.IDFloat16,
	zed.IDFloat32:    zed.IDFloat32,
	zed.IDFloat64:    zed.IDFloat64,
	zed.IDFloat128:   zed.IDFloat128,
	zed.IDFloat256:   zed.IDFloat256,
	zed.IDDecimal32:  zed.IDDecimal32,
	zed.IDDecimal64:  zed.IDDecimal64,
	zed.IDDecimal128: zed.IDDecimal128,
	zed.IDDecimal256: zed.IDDecimal256,
```

cli/inputflags/flags.go
mccanne: I would hope that most people who use zq don't care that it's implemented in Go and more people know what a thread is compared to a goroutine.

cmd/zed/dev/zngdump/command.go
mccanne: This is an early start to more useful tooling and I'd rather make it better later.  I needed to use this and found out dd also does the trick so I will take it out.

mccanne: Actually, let's just leave it and fix it later.

cmd/zed/dev/zngdump/command.go
mccanne: Later.  This should also just use the code in zngio now that it is stable.

cmd/zed/dev/zngdump/command.go
mccanne: This is all going to change when we clean it up later.

index/writer.go
```diff
@@ -333,7 +335,7 @@ func (w *indexWriter) write(rec *zed.Value) error {
 		if err != nil {
 			return err
 		}
-		w.frameKey = key
+		w.frameKey = key.Copy()
```
mattnibs, line 338: Maybe use CopyFrom here?
```suggestion
		w.frameKey.CopyFrom(key)
```

nwt (approved):

docs/data-model/zng.md
mccanne: I found it easier to write and follow while I was coding to unravel the various cases but I will write it up this way.

cli/inputflags/flags.go
nwt: "Threads" made me think of runtime.LockOSThread.

But I suppose no one will use this so the name doesn't really matter.

docs/data-model/zng.md
nwt: For making the specification and implementation simpler, but that doesn't strike me as an issue worth pursuing. What's here is fine.

expr/coerce/promote.go
nwt: We can do this later.

docs/data-model/zng.md
nwt: We can do this later.

nwt (approved):
Let's merge this!

index/writer.go
mccanne: This doesn't work in general because the memory for the Value struct can get clobbered too.  Also, I think we'll soon end up making CopyFrom() go away.

expr/coerce/promote.go
mccanne: As agreed we can fix this one later...

PR #3482, opened 2022-01-21
add expr.Sorter

nwt:
expr.Sorter holds state used to improve the performance of its
SortStable method.


mccanne (commented):
I was thinking the sort item would be 
```
int64
*zed.Value
```
the resolution of nulls and zero and stuff should all be captured in the int value when you build the slice of []{int64,*zed.Value}
Then after the sort you can stream these *zed.Values from this sorted slice in order of sort.  Maybe I'm missing something.

nwt:
> I was thinking the sort item would be
> 
> ```
> int64
> *zed.Value
> ```

Caching the first Evaluator's Eval result in expr0val yields about a 40% speedup for non-integer types, so that seems worth keeping.

> the resolution of nulls and zero and stuff should all be captured in the int value when you build the slice of []{int64,*zed.Value}

Were you thinking we'd map null to math.MaxInt64/MinInt64 (according to nullsMax) and then, if we're comparing two of those, fall back to comparing zed.Values instead?

> Then after the sort you can stream these *zed.Values from this sorted slice in order of sort. Maybe I'm missing something.

I did SortStable because it would be easy to swap into lake.Writer, proc/sort.Proc, and proc/spill.MergeSort. But we should definitely do what you suggest, either in this PR or another. Any preference?

nwt:
@mccanne: Sorting signed integers does indeed go faster with the sort item pared back to just `type item { index int32; i64 int64 }`, but it goes faster still with separate slices (i.e., `type Sorter struct { indices []uint32; i64s []int64 }`), so I went that direction. And it puts the marginal cost of caching the first Evaluator's Eval result close to zero, so I kept that.

mccanne (approved):
Very cool.

I think there is a bug when you have null(int64) and 9223372036854775807 but we can live with that for now.

When we generalize to other types, we'll want some sort of interface type, which should also save additional memory.

nwt:
> I think there is a bug when you have null(int64) and 9223372036854775807 but we can live with that for now.

As discussed elsewhere, I think the behavior here is a consequence of the `sort` operator's `-nulls first`/`-nulls last` semantics, a result of which is that `-r` does not affect the position of nulls in the output.

> When we generalize to other types, we'll want some sort of interface type, which should also save additional memory.

👍

PR #3483, opened 2022-01-22
move marshal logic from zed package to jsonio

mccanne:
This commit moves the marshal methods from the zed package,
which shouldn't have any serialization knowledge, into the
jsonio package.

zed.Type.String() and zed.Type.Format() are next in line.


mccanne:
This also fixes a bunch of bugs that were previously there.

zio/jsonio/marshal.go
```diff
@@ -0,0 +1,129 @@
...
+	case *zed.TypeSet:
+		return marshalSet(typ, bytes)
+	case *zed.TypeMap:
+		return marshalMap(typ, bytes)
+	case *zed.TypeUnion:
+		return marshalUnion(typ, bytes)
+	case *zed.TypeEnum:
+		return marshalEnum(typ, bytes)
+	case *zed.TypeError:
+		return &zedErr{marshalAny(typ.Type, bytes)}
```
nwt, line 64: Nit: Clearer I think to drop zedErr and do this.
```suggestion
		return map[string]interface{}{"error": marshalAny(typ.Type, bytes)}
```

service/ztests/curl-add-gzip.yaml
```diff
@@ -14,5 +14,5 @@ inputs:
 
 outputs:
   - name: stdout
-    regexp: |
-      {"kind":"api.CommitResponse","value":{"commit":"0x\w{40}","warnings":\[]}}
+    data: |
+      {"type":"Error","kind":"item does not exist","error":"missing: pool not found"}
```
nwt, line 18: Non-nit: This shouldn't change. It probably did because the contents of pool.json changed and `zq -f text "cut value.pool.id" pool.json)` isn't printing the pool ID.

zio/jsonio/marshal.go
mccanne: It seems you should use map[string]interface{} instead of struct when the keys are dynamic/unknown but ok.

service/ztests/curl-add-gzip.yaml
mccanne: Oops, I thought I fixed that one.

mccanne: I just realized I had fixed this in another branch.   I have too many branches stacked up... the weekend effect.

nwt (approved):

PR #3486, opened 2022-01-23
unify trailers and add dev dig commands

### LINKED ISSUE #3470: improve zngdump
zngdump wasa created to help debug and test the 2022 ZNG format.

It will come in handy down the road when we start to experiment with performance issues, e.g., to understand what kind of data, compression, 
type information, etc impacts performance of the runtime.

We should improve zng to output more stats about ZNG (looking into the various frames, e.g., number of values in a frame) and output the data as Zed instead of text.

mccanne:
This commit unifies the ZNG trailer structure used by ZST and
where the section layout is generically shared.  We also added
tooling as a new command "zed dev dig" to generically explore
sectioned files with trailers.

"zed dev zngdump" is now called "zed dev dig frames" and now
outputs the frame metadata as Zed instead of printed text.

The "zst inspect" was removed since its functionality is now
subsumed by "zed dev dig".

Closes #3470


zio/zngio/trailer.go
```diff
@@ -1,34 +1,130 @@
...
 		}
 		if off == 0 || b[off-1] == EOS {
 			if ok := validStream(b, off); ok {
-				return b[off:], nil
+				return off
 			}
 		}
 	}
 }
+func decodeTrailer(b []byte) *zed.Value {
```
nwt, line 124: Nit: This doesn't decode the trailer, so maybe call it readTrailer?

zst/trailer.go
```diff
@@ -1,159 +1,37 @@
...
-}
-
-func recordToTrailer(rec *zed.Value) (*Trailer, error) {
-	var trailer Trailer
-	var err error
-	trailer.Magic, err = rec.AccessString(MagicField)
-	if err != nil || trailer.Magic != MagicVal {
-		return nil, ErrNotZst
+	if trailer.Type != FileType {
+		return nil, nil, fmt.Errorf("not a zst file: trailer type is %q", trailer.Type)
```
nwt, line 27: While you're here, maybe change zst to ZST in all these error strings?

nwt (approved):

zst/trailer.go
mccanne: This is being changed on zst PR

PR #3488, opened 2022-01-24
simplify ZST reassembly maps

mccanne:
Now that ZST handles any typ of top-level value, this commit simplifies
the reassembly section that defines the reassembly maps.
This is a baby step toward a larger PR that will simplify the
column maps and clean up how the column maps are built.

We updated the ZST format doc to reflect the changes and the
new dig tooling and adding nulls to the tests as the rebase onto
zng-2022 fixed the bug with nulls in zst.  There is still a bug
with null container types but this will get fixed with an
ancipated refactor of the column maps.

We also fixed a bug with the handling of sets where set type instead
of the element type was passed to the column constructor.

### DIFF
--- a/docs/data-model/zst.md
+++ b/docs/data-model/zst.md
@@ -192,11 +192,11 @@ values in the ZST file have the same super ID,
 the super column will compress trivially.
 
 The reassembly map appears as the next value in the reassembly section
-is of type `<segmap>`.
+and is of type `<segmap>`.
 
 #### The Reassembly Records
 
-Following the root reassembly map are N reassembly maps one for each unique super type.
+Following the root reassembly map are N reassembly maps, one for each unique super type.
 
 Each reassembly record is a record of type `<any_column>`, as defined below,
 where each reassembly record appears in the same sequence as the original N schemas.
@@ -411,7 +411,8 @@ zed dev dig section -Z 1 hello.zst
 ```
 which provides the Zed output (comments added with explanations):
 ```
-// First, the all types are declared with null values (just one here).
+// First, all of the types of the encoded value sequence are declared
+// with null values (just one here).
 
 null ({a:string,b:string})
 
--- a/docs/data-model/zst.md
+++ b/docs/data-model/zst.md
@@ -191,16 +191,12 @@ will compress very significantly, e.g., in the special case that all the
 values in the ZST file have the same super ID,
 the super column will compress trivially.
 
-The type of the reassembly record for the super column stream has the
-type signature:
-```
-{root:<segmap>}
-```
+The reassembly map appears as the next value in the reassembly section
+is of type `<segmap>`.
 
 #### The Reassembly Records
 
-The remaining N records in the reassembly stream define the reassembly
-maps for each super type.
+Following the root reassembly map are N reassembly maps one for each unique super type.
 
 Each reassembly record is a record of type `<any_column>`, as defined below,
 where each reassembly record appears in the same sequence as the original N schemas.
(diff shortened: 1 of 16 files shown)


docs/data-model/zst.md
```diff
@@ -191,16 +191,12 @@ will compress very significantly, e.g., in the special case that all the
 values in the ZST file have the same super ID,
 the super column will compress trivially.
 
-The type of the reassembly record for the super column stream has the
-type signature:
-```
-{root:<segmap>}
-```
+The reassembly map appears as the next value in the reassembly section
+is of type `<segmap>`.
```
nwt, line 195:
```suggestion
The reassembly map appears as the next value in the reassembly section
and is of type `<segmap>`.
```

docs/data-model/zst.md
```diff
@@ -191,16 +191,12 @@ will compress very significantly, e.g., in the special case that all the
...
 
 #### The Reassembly Records
 
-The remaining N records in the reassembly stream define the reassembly
-maps for each super type.
+Following the root reassembly map are N reassembly maps one for each unique super type.
```
nwt, line 199:
```suggestion
Following the root reassembly map are N reassembly maps, one for each unique super type.
```

docs/data-model/zst.md
```diff
@@ -408,32 +404,29 @@ gracie
 0
 ===
 ```
-
 To see the detailed ZST structure described as ZSON, you can use the `zst`
 command like this:
 ```
-zst inspect -Z -trailer hello.zst
+zed dev dig section -Z 1 hello.zst
```
nwt, line 410: Maybe use mdtest? Or maybe not since then you'd lose the ability to include inline comments.

docs/data-model/zst.md
```diff
@@ -408,32 +404,29 @@ gracie
...
 ```
 which provides the Zed output (comments added with explanations):
 ```
-// First, the schemas are defined (just one here).
-{
-    a: null (string),               
-    b: null (string)
-}
+// First, the all types are declared with null values (just one here).
```
nwt, line 414: The all types?

zst/column/column.go
```diff
@@ -92,14 +92,16 @@ func Unmarshal(typ zed.Type, in zed.Value, r io.ReaderAt) (Any, error) {
...
 	case *zed.TypeUnion:
 		u := &Union{}
 		err := u.UnmarshalZNG(typ, in, r)
 		return u, err
-	case *zed.TypeMap, *zed.TypeEnum:
-		return nil, errors.New("unsupported")
+	case *zed.TypeMap:
+		return nil, errors.New("types 'map' is currently unsupported by the ZST implementation")
+	case *zed.TypeEnum:
+		return nil, errors.New("types 'enum' is currently unsupported by the ZST implementation")
```
nwt, line 104:
```suggestion
		return nil, errors.New("type 'map' is currently unsupported by the ZST implementation")
	case *zed.TypeEnum:
		return nil, errors.New("type 'enum' is currently unsupported by the ZST implementation")
```

zst/column/record.go
```diff
@@ -70,7 +70,7 @@ type Record []*Field
 func (r *Record) UnmarshalZNG(utyp zed.Type, in zed.Value, reader io.ReaderAt) error {
 	typ, ok := zed.TypeUnder(utyp).(*zed.TypeRecord)
 	if !ok {
-		return errors.New("cannot unmarshal non-record into record")
+		return errors.New("corrupt zst object: record_column is not a record")
```
nwt, line 73: s/zst/ZST/ in all these error strings you're touching?

zst/object.go
```diff
@@ -120,50 +120,46 @@ func (o *Object) IsEmpty() bool {
...
+		if val == nil {
+			return nil, errors.New("zst: corrupt reassembly section: number of reassembly maps not equal to number of types")
+		}
+		assembly.maps = append(assembly.maps, val.Copy())
 	}
-	rec, _ = reader.Read()
-	if rec != nil {
-		return nil, errors.New("extra records in reassembly section")
+	val, _ = reader.Read()
+	if val != nil {
```
nwt, line 157:
```suggestion
	if val, _ := reader.Read(); val != nil {
```

nwt (approved):

docs/data-model/zst.md
mccanne: We can improve later

PR #3489, opened 2022-01-24
fix textio for non-record values

mccanne:
This commit fixes the textio record to handle non-record value by
printing them using the zeek TSV value format.

### DIFF
--- a/zio/textio/writer.go
+++ b/zio/textio/writer.go
@@ -27,7 +27,15 @@ func (w *Writer) Close() error {
 	return w.writer.Close()
 }
 
-func (w *Writer) Write(rec *zed.Value) error {
+func (w *Writer) Write(val *zed.Value) error {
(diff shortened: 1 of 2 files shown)


zio/textio/writer.go
```diff
@@ -27,7 +27,15 @@ func (w *Writer) Close() error {
 	return w.writer.Close()
 }
 
-func (w *Writer) Write(rec *zed.Value) error {
+func (w *Writer) Write(val *zed.Value) error {
+	if _, ok := zed.TypeUnder(val.Type).(*zed.TypeRecord); ok {
+		return w.writeRecord(val)
+	}
+	_, err := fmt.Fprintf(w.writer, "%s\n", zeekio.FormatValue(*val))
```
nwt, line 34:
```suggestion
	_, err := fmt.Fprintln(w.writer, zeekio.FormatValue(*val))
```

nwt (approved):

PR #3497, opened 2022-01-24
fix bug with enum type value in zson formatter

mccanne:


zson/ztests/enum-typeval.yaml
```diff
@@ -0,0 +1,7 @@
+zed: yield (typeof(this))
```
nwt, line 1:
```suggestion
zed: yield typeof(this)
```

nwt (approved):

PR #3498, opened 2022-01-24
improve syntax of ZSON decorator typedefs

mccanne:
This commit improves the syntax to drop the right-hand side enclosing
angle brackets as they are no longer needed to resolve any ambiguity.
James Kerr suggested this excellent improvement.


nwt (approved):
This is a win.

PR #3501, opened 2022-01-24
semantic compiler: no scopes for lvals

mattnibs:
Do not have the semantic compiler replace lvals with scope variables.

Closes #3941

### DIFF
--- a/proc/traverse/ztests/scope-groupby-key.yaml
+++ b/proc/traverse/ztests/scope-groupby-key.yaml
@@ -11,4 +11,3 @@ output: |
   {s:"a",sum:3}
   {s:"b",sum:9}
   {s:"b",sum:13}
-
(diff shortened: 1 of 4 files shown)


proc/traverse/ztests/scope-groupby-key.yaml
```diff
@@ -0,0 +1,14 @@
...
+  {s:"a",x:[3]}
+  {s:"b",x:[4,5]}
+  {s:"b",x:[6,7]}
+
+output: |
+  {s:"a",sum:3}
+  {s:"a",sum:3}
+  {s:"b",sum:9}
+  {s:"b",sum:13}
+
```
nwt, line 14:
```suggestion
```

nwt (approved):

mccanne (commented):
Would you consider adding semLval() instead of passing the boolean around?  Then semField should be called something else...

mattnibs:
> Would you consider adding semLval() instead of passing the boolean around? Then semField should be called something else...

Yeah I could do that. I thought about it but didn't want to duplicate a bunch of code for one if statement. I'll think about a better approach.

mccanne:
Why don't you merge and we can refactor later next time we come across this...

PR #3505, opened 2022-01-26
updates to ZJSON for Zed 2022

mccanne:
This commits updates the ZJSON implementation to reflect the
December 2021 changes to the Zed formats.  We changed the
specification a bit based on implementation gaps.

We also changed the service protocol to not wrap ZJSON
in the template {kind:"Object",value:<zjson value>}.
ZJSON values all now have signature {type:<>,value:<>}
so we can easily encode server control messages using
the same type field since control messages will have
a string type field and ZJSON data values all have a
JSON object type field.

ZJSON now supports the full Zed model and handles null
values better throughout (e.g., null type values) as well as
any top-level value, not just records.


zio/zjsonio/types.go
```diff
@@ -185,6 +278,6 @@ func (d decoder) decodeTypeMap(zctx *zed.Context, m *astzed.TypeMap) (*zed.TypeM
 	return zctx.LookupTypeMap(keyType, valType), nil
 }
 
-func (d decoder) decodeTypeEnum(zctx *zed.Context, enum *astzed.TypeEnum) (*zed.TypeEnum, error) {
+func (d decoder) decodeTypeEnum(zctx *zed.Context, enum *Enum) (*zed.TypeEnum, error) {
 	return nil, errors.New("TBD: issue #2508")
```
nwt, line 282: Is this the right time to fix this and enable `zio/zjsonio/ztests/enum.yaml`?

zio/zjsonio/types.go
```diff
@@ -5,147 +5,240 @@ import (
...
-	astzed "github.com/brimdata/zed/compiler/ast/zed"
 )
 
-type encoder map[zed.Type]string
+type Type interface {
+	typeNode()
+}
+
+type (
+	Primitive struct {
```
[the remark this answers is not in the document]
mccanne: ok

zio/zjsonio/types.go
mccanne: yes will do

nwt (approved):

PR #3506, opened 2022-01-26
refactor "this" representation in AST

mccanne:
This commit changes the representation of "this" to a normal
identifier in the AST and eliminates the ast.This.  We also changed
dag.Path to dag.This as this feels like a more sensible name.

This PR sets us up to experiment with binding "this" to different
scopes and binding "as" names to this in a subscope.


mattnibs (approved):
lgtm

nwt (approved):

PR #3509, opened 2022-01-26
add scoping of "this" to Zed language

mccanne:
This commit introduces an "as" clause to a scoped over, which
allows "this" to be renamed so the inner scope can access this
via the name and also access the outer this via the name "this".
The scoping technique here works recursively into a hierarchy
of nested scopes.

We also fixed some problems with the semantic analysis of field
l-values and eliminated redundant code in the analysis of
the dot operator.  This was needed to correctly drop in scoping
for "this".

### DIFF
--- a/compiler/parser/parser.peg
+++ b/compiler/parser/parser.peg
@@ -560,9 +560,8 @@ As
   = _ "as" _ id:IdentifierName { RETURN(id) }
   / "" { RETURN("") }
 
-
 LetProc
-  = "let"i _ locals:LetAssignments __ over:ScopedOver  {
+  = "let"i _ locals:LetAssignments __ over:ScopedOver {
       RETURN(MAP("kind":"Let", "locals":locals, "over":over))
     }
 
--- a/compiler/parser/parser.peg
+++ b/compiler/parser/parser.peg
@@ -544,18 +544,25 @@ MergeProc
     }
 
 OverProc
-  = ScopedOver
+  = over:ScopedOver {
+      RETURN(MAP("kind":"Let", "locals":NULL, "over":over))
+    }
   / "over"i _ exprs:Exprs {
-      RETURN(MAP("kind":"Over", "exprs":exprs, "scope":NULL))
+      RETURN(MAP("kind":"Over", "exprs":exprs, "scope":NULL, "as":""))
     }
 
 ScopedOver
-  = "over"i _ exprs:Exprs __ scope:Scope {
-      RETURN(MAP("kind":"Over", "exprs":exprs, "scope":scope))
+  = "over"i _ exprs:Exprs as:As __ scope:Scope {
+      RETURN(MAP("kind":"Over", "exprs":exprs, "scope":scope, "as":as))
     }
 
+As
+  = _ "as" _ id:IdentifierName { RETURN(id) }
+  / "" { RETURN("") }
+
+
 LetProc
-  = "let"i _ locals:LetAssignments __ over:ScopedOver {
+  = "let"i _ locals:LetAssignments __ over:ScopedOver  {
       RETURN(MAP("kind":"Let", "locals":locals, "over":over))
     }
 
--- a/compiler/semantic/expr.go
+++ b/compiler/semantic/expr.go
@@ -48,7 +48,15 @@ func semExpr(scope *Scope, e ast.Expr) (dag.Expr, error) {
 			Value: zson.MustFormatValue(val),
 		}, nil
 	case *ast.ID:
-		return semField(scope, e, false)
+		// We use static scoping here to see if an identifier is
+		// a "var" reference to the name or a field access
+		// and transform the AST node appropriately.  The resulting DAG
+		// doesn't have Identifiers as they are resolved here
+		// one way or the other.
+		if ref := scope.Lookup(e.Name); ref != nil {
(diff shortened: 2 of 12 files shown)


compiler/parser/parser.peg
```diff
@@ -544,18 +544,25 @@ MergeProc
...
     }
 
+As
+  = _ "as" _ id:IdentifierName { RETURN(id) }
+  / "" { RETURN("") }
+
+
 LetProc
-  = "let"i _ locals:LetAssignments __ over:ScopedOver {
+  = "let"i _ locals:LetAssignments __ over:ScopedOver  {
```
nwt, line 565:
```suggestion
  = "let"i _ locals:LetAssignments __ over:ScopedOver {
```

compiler/parser/parser.peg
```diff
@@ -544,18 +544,25 @@ MergeProc
...
-      RETURN(MAP("kind":"Over", "exprs":exprs, "scope":scope))
+  = "over"i _ exprs:Exprs as:As __ scope:Scope {
+      RETURN(MAP("kind":"Over", "exprs":exprs, "scope":scope, "as":as))
     }
 
+As
+  = _ "as" _ id:IdentifierName { RETURN(id) }
+  / "" { RETURN("") }
+
+
```
nwt, line 563:
```suggestion
```

compiler/semantic/expr.go
```diff
@@ -374,14 +398,32 @@ func semAssignment(scope *Scope, a ast.Assignment) (dag.Assignment, error) {
...
 	} else {
-		lhs, err = semField(scope, a.RHS, true)
+		lhs, err = semExpr(scope, a.RHS)
 		if err != nil {
 			return dag.Assignment{}, errors.New("assignment name could not be inferred from rhs expression")
 		}
 	}
+	if summarize {
+		// Summarize always outputs its results as new records of "this"
+		// this so if we have an "as" that overrides "this", we just
```
nwt, line 411:
```suggestion
		// Summarize always outputs its results as new records of "this"
		// so if we have an "as" that overrides "this", we just
```

compiler/semantic/expr.go
```diff
@@ -374,14 +398,32 @@ func semAssignment(scope *Scope, a ast.Assignment) (dag.Assignment, error) {
...
+		// convert it back to a local this.
+		if dot, ok := lhs.(*dag.Dot); ok {
+			if v, ok := dot.LHS.(*dag.Var); ok && v.Name == "this" {
+				lhs = &dag.This{Kind: "This", Path: []string{dot.RHS}}
+			}
+		}
+	}
+	// Make sure we have a valid l-value for lhs.
```
nwt, line 419: Nit: Existing comments say "lval" instead of "l-value".
```suggestion
	// Make sure we have a valid lval for lhs.
```

compiler/semantic/expr.go
```diff
@@ -374,14 +398,32 @@ func semAssignment(scope *Scope, a ast.Assignment) (dag.Assignment, error) {
...
+	this, ok := lhs.(*dag.This)
+	if !ok {
+		return dag.Assignment{}, errors.New("illegal left-hand side of assignment'")
+	}
+	if len(this.Path) == 0 {
+		return dag.Assignment{}, errors.New("cannot assign to \"this\"")
```
nwt, line 425: Nit: Other error strings use single quotes in this circumstance.
```suggestion
		return dag.Assignment{}, errors.New("cannot assign to 'this'")
```

compiler/semantic/proc.go
```diff
@@ -532,6 +519,27 @@ func semProc(ctx context.Context, scope *Scope, p ast.Proc, adaptor proc.DataAda
...
+			// "as" name resolves to "this" in the scope above, which is
+			// where it was bound.
+			if err := scope.DefineAs(as); err != nil {
+				return nil, err
+			}
+			if err := scope.DefineVar("this"); err != nil {
+				return nil, err
+			}
+			// We append "this" to locals so it will be eval'd
+			// but we don't but it in the scope.
```
nwt, line 537:
```suggestion
			// but we don't put it in the scope.
```

compiler/semantic/scope.go
```diff
@@ -45,6 +45,18 @@ func (s *Scope) DefineVar(name string) error {
...
 
+func (s *Scope) DefineAs(name string) error {
+	b := s.tos()
+	if _, ok := b.symbols[name]; ok {
+		return fmt.Errorf("symbol %q redefined", name)
+	}
+	ref := &dag.This{Kind: "This"}
+	// We add the symbol to the table but don't bump nvars because
+	// it's not a var and doesn't take a slot in the batch vars.
+	b.Define(name, ref)
```
nwt, line 56:
```suggestion
	// We add the symbol to the table but don't bump nvars because
	// it's not a var and doesn't take a slot in the batch vars.
	b.Define(name, &dag.This{Kind: "This"})
```

nwt (approved):

compiler/semantic/scope.go
mccanne: My brain doesn't think that way (you are creating a ref that will be returned and putting it in the symbol table not creating two things that happen to be the same and returning one of them).  I will lift the comment.

PR #3511, opened 2022-01-26
over: work with maps and records

mattnibs:
Make over work with maps and records. For each element in an record/map
over will emit `{key:<key>,value:<value>}` as the context.

Also:
- Use un-union'd values for over.

### DIFF
--- a/proc/traverse/over.go
+++ b/proc/traverse/over.go
@@ -7,6 +7,7 @@ import (
 	"github.com/brimdata/zed/expr"
 	"github.com/brimdata/zed/proc"
 	"github.com/brimdata/zed/zbuf"
+	"github.com/brimdata/zed/zcode"
 )
 
 type Over struct {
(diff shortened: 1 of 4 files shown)


proc/traverse/over.go
```diff
@@ -65,26 +68,52 @@ func (o *Over) over(batch zbuf.Batch, this *zed.Value) zbuf.Batch {
...
+			// this copy through the batch.
+			val := expr.ValueUnder(zed.NewValue(typ, it.Next()))
+			vals = append(vals, *val.Copy())
+		}
 		return vals
+	case *zed.TypeMap:
+		rtyp := zctx.MustLookupTypeRecord([]zed.Column{
+			zed.NewColumn("key", typ.KeyType),
+			zed.NewColumn("value", typ.ValType),
+		})
```
mattnibs, line 93: Hmm, I'm thinking value should also be de-unioned here as we do with arrays?

mattnibs: Actually also the key as well.

mccanne:
Did you want to merge this and fix the union problem later?  Might be easier after we update other parts of the system to handle unions more uniformly...

mattnibs:
> Did you want to merge this and fix the union problem later? Might be easier after we update other parts of the system to handle unions more uniformly...

Sure!

mccanne (approved):

PR #3515, opened 2022-01-27
remove cut() function

mccanne:
The cut() function was added to the Zed lanuage before it had
record expressions (and before it was called Zed). This commit
removes it and updates the tests to use record expressions.


nwt (approved):
You truly love to see it.

PR #3516, opened 2022-01-27
move put proc to expr.Putter as an expr.Applier

mccanne:
This refactor is needed to add a with-clause to record expressions
coming in a subsequent PR.


expr/putter.go
```diff
@@ -82,31 +76,31 @@ func (p *Proc) eval(ectx expr.Context, this *zed.Value) []zed.Value {
...
-	fromInput  op = iota // copy field from input record
-	fromClause           // copy field from put assignment
-	record               // recurse into record below us
+	putFromInput  putOp = iota // copy field from input record
+	putFromClause              // copy field from put assignment
+	putRecord                  // recurse into record below us
 )
 
-func (s step) build(in zcode.Bytes, b *zcode.Builder, vals []zed.Value) zcode.Bytes {
+func (s putStep) build(in zcode.Bytes, b *zcode.Builder, vals []zed.Value) zcode.Bytes {
```
nwt, line 101: These things occupy six words, so use a pointer receiver.
```suggestion
func (p *putStep) build(in zcode.Bytes, b *zcode.Builder, vals []zed.Value) zcode.Bytes {
```

mccanne:
Turns out the record-with clause can't use the putter because of the putter's assumptions about the input zcode relative to the out zcodes.  Anyway, it wasn't too hard to code up record-with a different way and it still seems like this PR is a win.

expr/putter.go
```diff
@@ -305,36 +298,23 @@ func (p *Proc) lookupRule(inType *zed.TypeRecord, vals []zed.Value) putRule {
...
-		// Copy is necessary because put can return its argument.
-		recs = append(recs, *rec.Copy())
-	}
-	batch.Unref()
-	return zbuf.NewArray(recs), nil
+func (*Putter) String() string { return "put" }
+
+func (*Putter) Warning() string {
+	return ""
 }
```
nwt, line 320:
```suggestion
func (*Putter) Warning() string { return "" }
```

nwt (approved):
I think you should add a "put-" prefix to any of the moved ztest files that don't already have it.

mccanne:
> I think you should add a "put-" prefix to any of the moved ztest files that don't already have it.

will do

PR #3517, opened 2022-01-27
add "with clause" to record literals

mccanne:
This commit extends record expressions to be able to create new
records from a template record and additional fields, e.g., {r with x:1}
would give {a:1,b:2,x:1} if r equaled {a:1,b:2}.  The template can be 
any expression.  This will be particularly useful in an inner scope
to join the outer this with inner-traversed values.

### DIFF
--- a/compiler/parser/parser.peg
+++ b/compiler/parser/parser.peg
@@ -778,8 +778,11 @@ Primary
   / "(" __ expr:Expr __ ")" { RETURN(expr) }
 
 Record
-  = "{" __ fields:Fields __ "}" {
-      RETURN(MAP("kind":"RecordExpr", "fields":fields ))
+  = "{" __ expr:Expr _ "with" _ fields:Fields __ "}" {
+      RETURN(MAP("kind":"RecordExpr", "fields":fields, "with":expr))
+    }
+  / "{" __ fields:Fields __ "}" {
+      RETURN(MAP("kind":"RecordExpr", "fields":fields, "with":NULL))
     }
 
 Fields
--- a/expr/values.go
+++ b/expr/values.go
@@ -103,12 +103,12 @@ func (r *RecordExprWith) Eval(ectx Context, this *zed.Value) *zed.Value {
 	b := r.builder
 	b.Reset()
 	var dirty bool
-	it := zcode.Iter(with.Bytes)
+	it := with.Iter()
 	for k, col := range typ.Columns {
 		if pos, ok := r.position[col.Name]; ok {
 			val := r.exprs[pos].Eval(ectx, this)
 			b.Append(val.Bytes)
-			col = zed.Column{col.Name, val.Type}
+			col.Type = val.Type
 			it.Next()
 		} else {
 			b.Append(it.Next())
@@ -116,7 +116,7 @@ func (r *RecordExprWith) Eval(ectx Context, this *zed.Value) *zed.Value {
 		if k >= len(cols) {
 			dirty = true
 			cols = append(cols, col)
-		} else if cols[k].Name != col.Name || cols[k].Type != col.Type {
+		} else if col != cols[k] {
 			dirty = true
 			cols[k] = col
 		}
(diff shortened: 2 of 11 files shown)


compiler/parser/parser.peg
```diff
@@ -778,8 +778,11 @@ Primary
   / "(" __ expr:Expr __ ")" { RETURN(expr) }
 
 Record
-  = "{" __ fields:Fields __ "}" {
-      RETURN(MAP("kind":"RecordExpr", "fields":fields ))
+  = "{" __ expr:Expr _ "with" _ fields:Fields __ "}" {
+      RETURN(MAP("kind":"RecordExpr", "fields":fields, "with":expr))
+    }
+  / "{" __ fields:Fields __ "}" {
+      RETURN(MAP("kind":"RecordExpr", "fields":fields, "with":NULL))
```
nwt, line 785:
```suggestion
  = "{" __ with_:(e:Expr _ "with" _ { RETURN(e) })? fields:Fields __ "}" {
      RETURN(MAP("kind":"RecordExpr", "fields":fields, "with":with_))
```

expr/values.go
```diff
@@ -54,6 +55,100 @@ func (r *RecordExpr) Eval(ectx Context, this *zed.Value) *zed.Value {
...
+	cols := r.columns
+	b := r.builder
+	b.Reset()
+	var dirty bool
+	it := zcode.Iter(with.Bytes)
+	for k, col := range typ.Columns {
+		if pos, ok := r.position[col.Name]; ok {
+			val := r.exprs[pos].Eval(ectx, this)
+			b.Append(val.Bytes)
+			col = zed.Column{col.Name, val.Type}
```
nwt, line 111:
```suggestion
			col.Type = val.Type
```

expr/values.go
```diff
@@ -54,6 +55,100 @@ func (r *RecordExpr) Eval(ectx Context, this *zed.Value) *zed.Value {
...
+			it.Next()
+		} else {
+			b.Append(it.Next())
+		}
+		if k >= len(cols) {
+			dirty = true
+			cols = append(cols, col)
+		} else if cols[k].Name != col.Name || cols[k].Type != col.Type {
```
nwt, line 119:
```suggestion
		} else if cols[k] != col {
```

expr/values.go
```diff
@@ -54,6 +55,100 @@ func (r *RecordExpr) Eval(ectx Context, this *zed.Value) *zed.Value {
...
+	// cols is a cache of the columns of the output record.
+	// As long as the type doesn't change, the columns will stay the
+	// same and the dirty boolean will stay false.  If something is
+	// diferrent, dirty becomes true, we look up the new type, and
+	// it stays in the cache for the next input.
+	cols := r.columns
+	b := r.builder
+	b.Reset()
+	var dirty bool
+	it := zcode.Iter(with.Bytes)
```
nwt, line 106:
```suggestion
	it := with.Iter()
```

expr/values.go
```diff
@@ -54,6 +55,100 @@ func (r *RecordExpr) Eval(ectx Context, this *zed.Value) *zed.Value {
...
+		if typ.HasField(r.names[k]) {
+			continue
+		}
+		val := e.Eval(ectx, this)
+		b.Append(val.Bytes)
+		col := zed.Column{r.names[k], val.Type}
+		if colno >= len(cols) {
+			dirty = true
+			cols = append(cols, col)
+		} else if cols[colno].Name != col.Name || cols[colno].Type != col.Type {
```
nwt, line 135:
```suggestion
		} else if cols[colno] != col {
```

nwt (approved):

compiler/parser/parser.peg
mccanne: I find this pattern a little harder to follow and maintain especially when debugging backtracking problems so I'd like to leave it in the explanded form.

PR #3519, opened 2022-01-27
jsonio reader: preserve JSON attributes

### LINKED ISSUE #2122: json-strict mode for zson

mattnibs:
Preserve json attributes when translating JSON to ZSON:
- Parse numbers as float64s.
- Maintain column ordering of the original JSON.
- Do not convert primitive values to records.

This changed also required some changes to zson.Unmarshal allowing
json parsed numbers to be appropriately cast to the correct go type
destination.

Closes #2122

### DIFF
--- a/zio/jsonio/ztests/reader.yaml
+++ b/zio/jsonio/ztests/reader.yaml
@@ -28,19 +28,19 @@ outputs:
   - name: stdout
     data: |
       {
+          ts: 1.521911721926018e+09,
           a: "hello, world",
           b: {
-              x: 4611686018427388000,
-              y: "127.0.0.1",
-              z: "&, <, and > should not be escaped"
-          },
-          ts: 1.521911721926018e+09
+              z: "&, <, and > should not be escaped",
+              x: 4611686018427387904.,
+              y: "127.0.0.1"
+          }
       }
       ===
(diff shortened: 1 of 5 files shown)


zio/jsonio/ztests/reader.yaml
```diff
@@ -28,19 +28,19 @@ outputs:
...
           b: {
-              x: 4611686018427388000,
-              y: "127.0.0.1",
-              z: "&, <, and > should not be escaped"
-          },
-          ts: 1.521911721926018e+09
+              z: "&, <, and > should not be escaped",
+              x: 4611686018427387904.,
+              y: "127.0.0.1"
+          }
```
nwt, line 37: These fields should remain sorted. (That's is the "canonical order" mentioned in #2122.)

mattnibs:
Closing in favor of another (faster) approach from @nwt

PR #3524, opened 2022-01-27
Add zed.CompareTypes

mattnibs:
Add zed.CompareTypes function which compares two types according to the
type ordering spec (https://github.com/brimdata/zed/blob/main/docs/formats/zed.md)
and use this function to implement type value sorts in package expr. A
follow up pr will use the functionality to create cannonical union
values with types properly and predictably sorted.


nwt (approved):

expr/sort.go
```diff
@@ -367,6 +367,14 @@ func LookupCompare(typ zed.Type) comparefn {
 			return va.Compare(vb)
 		}
 
+	case zed.IDType:
+		zctx := zed.NewContext()
+		return func(a, b zcode.Bytes) int {
+			va, _ := zctx.DecodeTypeValue(a)
+			vb, _ := zctx.DecodeTypeValue(b)
+			return zed.CompareTypes(va, vb)
+		}
```
nwt, line 376: I think we ultimately want `func CompareTypeValue(a, b zcode.Bytes) int` because `DecodeTypeValue` isn't cheap, but this is fine for now!

nwt (approved):

expr/sort.go
mattnibs: Agreed!

mattnibs: I'll put a comment here.

PR #3525, opened 2022-01-27
Canonical Unions

### LINKED ISSUE #2145: Normalize order of union types
Union types are currently order sensitive, meaning that `union[float64,bool]` and `union[bool,float64]` are different types. We should make those types be the same.

mattnibs:
Normalize union types by sorting types in a union according to the
specified type ordering.

Closes #2145

### DIFF
--- a/context.go
+++ b/context.go
@@ -4,6 +4,7 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
+	"sort"
 	"sync"
 
 	"github.com/brimdata/zed/zcode"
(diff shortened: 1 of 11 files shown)


context.go
```diff
@@ -169,6 +169,7 @@ func (c *Context) LookupTypeArray(inner Type) *TypeArray {
 }
 
 func (c *Context) LookupTypeUnion(types []Type) *TypeUnion {
+	SortTypes(types)
```
nwt, line 172: Nit: I don't think SortTypes is worth adding.
```suggestion
	sort.SliceStable(types, func(i, j int) bool {
		return CompareTypes(types[i], types[j]) < 0
	})
```

nwt (approved):

PR #3527, opened 2022-01-28
eliminate "." as a syntactical shortcut for "this"

mccanne:


nwt (approved):

nwt:
(But I think we'll eventually bring `.` back!)

PR #3531, opened 2022-01-28
collect: ensure correct selector for type

mattnibs:
With canonical unions the collect aggregation was not deriving the
correct selector index for a union of values since the type order may be
rearranged when the zed.Context creates the union type. Fix this and add
a lookup table to zed.TypeUnion so it's easier to find a type's
selector.

### DIFF
--- a/context.go
+++ b/context.go
@@ -184,6 +184,24 @@ func (c *Context) LookupTypeUnion(types []Type) *TypeUnion {
 	return typ
 }
 
+func (c *Context) LookupTypeOfValues(vals []Value) Type {
+	if len(vals) == 0 {
+		return TypeNull
+	}
+	var types []Type
+	m := make(map[Type]struct{})
+	for _, val := range vals {
+		if _, ok := m[val.Type]; !ok {
+			m[val.Type] = struct{}{}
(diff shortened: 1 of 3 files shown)


context.go
```diff
@@ -184,6 +184,24 @@ func (c *Context) LookupTypeUnion(types []Type) *TypeUnion {
 	return typ
 }
 
+func (c *Context) LookupTypeOfValues(vals []Value) Type {
```
nwt, line 187: I don't think I buy that this belongs here. I'd leave it in `expr/agg/collect.go` unless you plan to use it elsewhere, in which case putting it here seems fine for now.

nwt (approved):

context.go
mattnibs: Yeah I'd prefer to keep this since I'm working on `into` and this will be needed there as well. Or perhaps it could go elsewhere? In `analyze.go` we duplicate this same logic but can't use it since values or not zed.Values. I'm also imaging this same problem will show up whenever you're constructing maps/arrays/sets.

nwt: Let's keep it here for now then. I'm sure its proper home will eventually be apparent.

PR #3532, opened 2022-01-28
simplify methods to zed.Value and allow method chaining

### LINKED ISSUE #1347: improve UX for package zed with method chaining
Make a package for easily creating on-the-fly zng records, e.g.,

```
func TestZ(t *testing.T) {
	var zc z.Container
	rec := zc.NewRecord(
		z.Int64("cnt", 12),
		zc.Array("a", z.Int64v(1), z.Int64v(2), z.Int64v(3)),
		z.String("s", "hello"),
		zc.Record("r",
			z.String("a", "foo"),
			z.String("b", "bar")))

	assert.Equal(t, rec.Type.String(), "record[cnt:int64,a:array[int64],s:string,r:record[a:string,b:string]]")
	assert.Equal(t, rec.String(), "record[12,array[1,2,3],hello,record[foo,bar]]")
}
```
There's a branch called package-z in the zq repo. This is on icebox till we find a compelling reason to finish (e.g., zap logger integration)

### LINKED ISSUE #2166: create better Golang API for external consumption
Our external development-oriented users want to embed Z in their apps but our API for using and manipulating Z data has been designed for the Z runtime internals and can be cumbersome for external use.  We should design a first cut at an API that is performant and easy to use. 

For example, a simple thing that extract an IP address from `id.orig_h` shouldn't be so hard:
```
id, err := rec.Access("id")
if err != nil {
  // no such field
}
idType, ok := id.Type.(*zng.TypeRecord)
if !ok {
  // id is not a record 
}
id := NewRecord(idType, id.Bytes)
ip, err := id.AccessIP("orig_h")
// etc...
```

### LINKED ISSUE #2637: Store/return keyless data
When a user creates a new pool, they're required to specify the key(s) by which data will be sorted in that pool. However, the user may then try to import some data to the pool that lacks the key. As an alternative to simply dropping the data, we've discussed an approach where a sub-pool would be automatically created to hold the keyless data. The API could then provide ways to query/return that keyless data, e.g. perhaps the keyless records could always be returned at the beginning or end of results relative to the results sorted by the actual keys, and this could be configurable. Perhaps the Brim app could show it in a separate window so the user could easily look at it side-by-side next to the keyed data.

A classic use case for this from our Zeek-centric days was the notorious `loaded_scripts` log, which contains points without the `ts` timestamps that are present on nearly all other Zeek logs. This was a log that showed which Zeek scripts were loaded into the Zeek system that produced the logs. Being able to store & retrieve this could indeed be useful, such as if a user was expecting to see logs associated with a certain Zeek package and wanted to know from the loaded scripts logs whether the package was installed but saw no network traffic that triggered log outputs vs. the package not being loaded at all.

Yet another use case we've bumped into is bugs or limitations in systems that simply fail to output the keys on some events. By storing the data off to the side in a way that makes it easy to query/retrieve, this would make the user aware and potentially tee off a workflow where they investigate/fix the bug in the system that's originating the logs or perhaps write a Zed preprocessor that works around the problem by assigning a key in these cases. Once the root cause has been addressed, the user might choose to enhance the data in the sub-pool to add a key for those records while importing it to the main pool, then delete it from the sub-pool.

@mccanne and I agreed this would be a likely MVP1 target.

mccanne:
This commit simplifies the use of zed.Values by allowing method
chaining.  In this approach, a missing value is modeled as nil
since the receiver does not have access to a type context to create
a missing error.  Instead a caller with a type context can check
for nil and create a missing error value if desired.

In debugging tests, we noticed that the old issue to allow values with
missing pool keys is moot so we simply removed the relevant check for
missing added test coverage and are closing it.

We are also closing #1347 as half of this is addressed here and the other
half (constructing values using method chaining) is also moot now that we
have comprehensive marshaling.  This may come back later but is not
high priority right now.  This also address the intent of #2166.

In a subsequent PR, we will merge recordval.go into value.go and delete
the former.  For now, the diffs are easier to review on this commit.

Closes #2637
Closes #1347
Closes #2166


recordval.go
```diff
@@ -31,39 +29,30 @@ func (r *Value) Walk(rv Visitor) error {
...
-}
-
-func (r *Value) ValueByField(field string) (Value, error) {
-	col, ok := r.ColumnOfField(field)
-	if !ok {
-		return Value{}, ErrMissing
+func (v *Value) DerefByColumn(col int) *Value {
+	if v != nil {
+		if bytes := v.nth(col); bytes != nil {
+			v = &Value{v.Columns()[col].Type, bytes}
```
mattnibs, line 50: This would panic if Columns is nil (because the value is a record) or not the right size, I think we want to turns nil here right?

index/finder.go
```diff
@@ -155,10 +155,11 @@ func (f *Finder) search(compare keyCompareFn) (zio.Reader, error) {
 			// smaller than the smallest key present.
 			return nil, ErrNotFound
 		}
-		off, err = rec.AccessInt(f.meta.ChildOffsetField)
-		if err != nil {
-			return nil, fmt.Errorf("b-tree child field: %w", err)
+		child := rec.Deref(f.meta.ChildOffsetField)
+		if child == nil {
+			return nil, fmt.Errorf("b-tree child field is missing")
```
nwt, line 160:
```suggestion
			return nil, fmt.Errorf("B-tree child field is missing")
```

recordval.go
mccanne: Yes, will fix.

recordval.go
```diff
@@ -74,98 +63,63 @@ func (r *Value) TypeOfColumn(col int) Type {
 	return TypeRecordOf(r.Type).Columns[col].Type
 }
 
-func (r *Value) Access(field string) (Value, error) {
-	col, ok := r.ColumnOfField(field)
+func (v *Value) Deref(field string) *Value {
+	if v == nil {
+		return nil
+	}
+	col, ok := v.ColumnOfField(field)
```
mattnibs, line 70: This will panic if value is not a record (we want a nil instead right)?

mccanne:
I think we should merge this as a first step then clean it up, document it, add a bunch of tests , and add all the methods one would expect.

recordval.go
```diff
@@ -74,98 +63,63 @@ func (r *Value) TypeOfColumn(col int) Type {
...
-		v := DecodeUint(v.Bytes)
-		if v > math.MaxInt64 {
-			return 0, errors.New("conversion from uint64 to signed int results in overflow")
+func (v *Value) AsInt() int64 {
+	if v != nil {
+		switch TypeUnder(v.Type).(type) {
+		case *TypeOfUint8, *TypeOfUint16, *TypeOfUint32:
+			return int64(DecodeUint(v.Bytes))
+		case *TypeOfUint64:
+			return int64(DecodeUint(v.Bytes))
```
nwt, line 104:
```suggestion
		case *TypeOfUint8, *TypeOfUint16, *TypeOfUint32, *TypeOfUint64:
			return int64(DecodeUint(v.Bytes))
```

recordval.go
```diff
@@ -2,12 +2,10 @@ package zed
```
mattnibs, line 2: Maybe  move all of this to value.go and remove this file?

recordval.go
```diff
@@ -74,98 +63,63 @@ func (r *Value) TypeOfColumn(col int) Type {
...
+	return 0
 }
 
-func (r *Value) AccessTimeByColumn(colno int) (nano.Ts, error) {
-	zv, err := r.Slice(colno)
-	if err != nil {
-		return 0, err
+func (v *Value) MissingAsNull() *Value {
+	if v == nil {
+		v = Null
```
nwt, line 121:
```suggestion
	if v.IsMIssing() {
		return Null
```

zio/tableio/writer.go
```diff
@@ -54,13 +54,16 @@ func (w *Writer) Write(r *zed.Value) error {
 	var out []string
 	for k, col := range r.Columns() {
 		var v string
-		value := r.ValueByColumn(k)
+		value := r.DerefByColumn(k)
+		if value == nil {
+			value = zed.Null
+		}
```
nwt, line 60:
```suggestion
		value := r.DerefByColumn(k).MIssingAsNull()
```

zio/zeekio/parser_test.go
```diff
@@ -79,9 +79,8 @@ func TestLegacyZeekValid(t *testing.T) {
 	record, err := sendLegacyValues(parser, values)
 	require.NoError(t, err)
 
-	ts, err := record.AccessTime("ts")
-	assert.ErrorIs(t, err, zed.ErrMissing)
-	assert.Equal(t, nano.MinTs, ts)
+	assert.Equal(t, record.Deref("ts").MissingAsNull(), zed.Null)
+	//assert.Equal(t, nano.MinTs, ts)
```
nwt, line 83:
```suggestion
	assert.Equal(t, 0, ts)
```

recordval.go
mccanne: From PR comment:
```
In a subsequent PR, we will merge recordval.go into value.go and delete
the former. For now, the diffs are easier to review on this commit.
```

mattnibs: My bad!

recordval.go
mccanne: Fixed in ColumnOfField, which fixes it here.

mattnibs: If Value is an array or set you'll get a panic here.

recordval.go
```diff
@@ -31,141 +29,96 @@ func (r *Value) Walk(rv Visitor) error {
...
-	return TypeRecordOf(r.Type).ColumnOfField(field)
-}
-
-func (r *Value) TypeOfColumn(col int) Type {
-	return TypeRecordOf(r.Type).Columns[col].Type
-}
-
-func (r *Value) Access(field string) (Value, error) {
-	col, ok := r.ColumnOfField(field)
+func (v *Value) Deref(field string) *Value {
```
nwt, line 65: We should call this `Field` but later is fine for that.

recordval.go
```diff
@@ -31,141 +29,96 @@ func (r *Value) Walk(rv Visitor) error {
...
 	return TypeRecordOf(r.Type).Columns
 }
 
-// Value returns the indicated column as a Value.  If the column doesn't
-// exist or another error occurs, the nil Value is returned.
-func (r *Value) ValueByColumn(col int) Value {
-	zv, err := r.Slice(col)
-	if err != nil {
-		return Value{}
+func (v *Value) DerefByColumn(col int) *Value {
```
nwt, line 47: We should call this `Index` and make it work for other types but later is fine for all that.

recordval.go
mattnibs: Or should this private and only called by record-y functions?

nwt (approved):

recordval.go
mccanne: Table/text writer call DerefByColumn but only with records.  My point is for this method, it needs to be a record as implied by the name.  We will document all this and improve the names further in a subsequent PR.

recordval.go
mccanne: I pushed a fix for this based on your earlier comment.

mattnibs (approved):

PR #3535, opened 2022-01-29
fig bug unmarshaling zed.Value

mccanne:
This commit fixes a bug where the unmarshaler would utilize the bytes
buffer of the zed.Value being decoded when returning an unmarshaled
zed.Value.  This can cause corruption to that zed.Value whenever the
original byte slice is changed as ownership is unclear in this scenario.
The fix is top copy the needed bytes from the incoming Zed value when
unmarshaling zed.Values.

This code path is not used much but manifested in the unmarshaling of
the commit journal when creating the scan scnapshots.  It went
unnoticed because of the way the scan would get built: with all
of the buggy spans the same, every data object in the pool would be
merge-scanned for any query.  While  correct, this can obviously
cause really poor performance.

### DIFF
--- a/zson/marshal_zng_test.go
+++ b/zson/marshal_zng_test.go
@@ -537,8 +537,7 @@ func TestEmbeddedInterface(t *testing.T) {
 }
 
 func TestMultipleZedValues(t *testing.T) {
-	bytes := make([]byte, 3)
-	copy(bytes, []byte("foo"))
+	bytes := []byte("foo")
 	u := zson.NewZNGUnmarshaler()
 	var foo zed.Value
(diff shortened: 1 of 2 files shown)


zson/marshal_zng_test.go
```diff
@@ -535,3 +535,19 @@ func TestEmbeddedInterface(t *testing.T) {
 	assert.Equal(t, true, ok)
 	assert.Equal(t, "It's a thing one", thingB.Who())
 }
+
+func TestMultipleZedValues(t *testing.T) {
+	bytes := make([]byte, 3)
+	copy(bytes, []byte("foo"))
```
nwt, line 541:
```suggestion
	bytes := []byte("foo")
```

nwt (approved):

PR #3538, opened 2022-01-30
fix bug in zcode.DecodeTagLength that caused bufferfilter panic

mccanne:


zcode/bytes.go
```diff
@@ -88,7 +88,10 @@ func DecodeTagLength(b Bytes) int {
 	if n <= 0 {
 		panic(fmt.Sprintf("bad uvarint: %d", n))
 	}
-	return int(u64) + n
+	if u64 == 0 {
```
nwt, line 91:
```suggestion
	if tagIsNull(u64) {
```

nwt (approved):

PR #3540, opened 2022-01-30
fix buffer pool for uncompressed buffers

mccanne:
This quick commit speeds up zq searches by 25% by sending the
decompressed buffers to a pool along with the compressed buffers.
It's a bit hacky and this could all use some more work but
I think it's okay to check in for now.

### DIFF
--- a/zio/zngio/buffer.go
+++ b/zio/zngio/buffer.go
@@ -10,11 +10,12 @@ type buffer struct {
 	off  int
 }
 
-var bufferPool sync.Pool
+var bigBuffers sync.Pool
+var smallBuffers sync.Pool
 
 func newBuffer(length int) *buffer {
 	if length <= DefaultLZ4BlockSize {
-		b, ok := bufferPool.Get().(*buffer)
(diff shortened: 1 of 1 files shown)


nwt (approved):
A simpler way to address this is to do what we used to do and have zngio.Writer.Write call Writer.flush before (rather than after) len(Writer.values) exceeds Writer.thresh.

mccanne:
> A simpler way to address this is to do what we used to do and have zngio.Writer.Write call Writer.flush before (rather than after) len(Writer.values) exceeds Writer.thresh.

flush() means multiple things that you don't know ahead of time and even if you just focused on the compressed buffers there are uncompressed buffers that need to be managed also.  To me, having the buffer pool adapt to the size of the buffers in use rather than making sure the encode path always stays below the max buffer size is "simpler" but perhaps this is chocolate vs vanilla.  And anyway, you can't have an invariant that all buffers will always be smaller than the max because you can have an arbitrarily large value so you have to deal with the exception anyway.

PR #3543, opened 2022-01-30
fix peeker bug and rework interactive behavior

mccanne:
This commit fixes a bug introduced in #3375 where ReadAtLeast()
could be called with a bigger request than the buffer provided.

### DIFF
--- a/pkg/peeker/reader.go
+++ b/pkg/peeker/reader.go
@@ -7,11 +7,10 @@ import (
 
 type Reader struct {
 	io.Reader
-	limit       int
-	buffer      []byte
-	cursor      []byte
-	eof         bool
-	interactive bool
+	limit  int
+	buffer []byte
+	cursor []byte
+	eof    bool
 }
 
 var (
@@ -19,14 +18,13 @@ var (
 	ErrTruncated      = errors.New("truncated input")
 )
 
-func NewReader(reader io.Reader, size, max int, interactive bool) *Reader {
+func NewReader(reader io.Reader, size, max int) *Reader {
 	b := make([]byte, size)
 	return &Reader{
-		Reader:      reader,
-		limit:       max,
-		buffer:      b,
-		cursor:      b[:0],
-		interactive: interactive,
+		Reader: reader,
+		limit:  max,
+		buffer: b,
+		cursor: b[:0],
 	}
 }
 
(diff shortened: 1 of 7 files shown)


pkg/peeker/reader.go
```diff
@@ -37,17 +39,23 @@ func (r *Reader) Limit() int {
...
 	copy(r.buffer, r.cursor)
 	clen := len(r.cursor)
-	n, err := io.ReadAtLeast(r.Reader, r.buffer[clen:], min)
+	var request int
+	if r.interactive {
+		request = need - clen
+	} else {
+		request = len(r.buffer) - clen
+	}
+	n, err := io.ReadAtLeast(r.Reader, r.buffer[clen:], request)
```
nwt, line 58: I don't think this path offers any benefit. `io.ReadAtLeast`'s first call to `r.Reader.Read` will fill `r.buffer[clen:]` if enough bytes are available (as they usually are when reading from a file or object store). But if it doesn't, why block here until it's filled instead of allowing the caller to make forward progress once `min` bytes are available?

mccanne: The effect here isn't as pronounced for ZNG now that we have frames but the behavior for the old scanner (and for possible future users of Peeker... this is in pkg) will do a system call for every small read when reading from a reader like the file system that otherwise doesn't run asynchronously.  In terms of letting the reader make forward progress, pipelining will work well reading a large buffer at a time as the reader issues the next big read and the scanners are busy working on the previous buffer.  If reading is the bottleneck, doing lots of small reads to let the scanners make forward progress doesn't solve the problem.

mccanne: Maybe I'm missing something.  Is it that the read is filling the buffer when data is available beyond the request size but waiting only until the min is satisfied?

nwt: > Is it that the read is filling the buffer when data is available beyond the request size but waiting only until the min is satisfied?

That's correct. `io.Reader.Read` semantics are conventionally "give me as much as you can without blocking." The only thing `io.ReadAtLeast` adds to that is a loop that calls `Read` until it has at least `min` bytes.

You can see this if you put something like this right after the call to `io.ReadAtLeast` and then read a large file with `zq`. The first few and last few reads will be short but all the rest will be full-length.
```go
	if n < len(r.buffer[clen:]) {
		fmt.Println("short", n, len(r.buffer[clen:]))
	} else {
		fmt.Println("full", n)
	}
```

nwt (approved):

PR #3545, opened 2022-01-31
make default LZ buffer size much bigger

mccanne:
This commit changes the default LZ buffer size in the ZNG writer
from 16KB to 500KB leading to a speedup of 5X (in combination
with the checked-in buffering improvements) for Boyer-Moore string
search.  The compressed ZNG files are materially smaller.

I tried 1MB too (after fixing the bug in #3543) but it didn't seem
to make a difference over 500KB.


zio/zngio/writer.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 // DefaultLZ4BlockSize is a reasonable default for WriterOpts.LZ4BlockSize.
-const DefaultLZ4BlockSize = 16 * 1024
+const DefaultLZ4BlockSize = 500 * 1024
```
nwt, line 12: No objection here—just curious—but why not 512?

nwt (approved):

zio/zngio/writer.go
mccanne: No reason

PR #3547, opened 2022-01-31
zed.Context.LookupTypeUnion: dedupe unions

mattnibs:
LookupTypeUnion shoud produce canonical union values, which means
it should not produce union values with deplicate types. Fix this.

### DIFF
--- a/context.go
+++ b/context.go
@@ -173,13 +173,21 @@ func (c *Context) LookupTypeUnion(types []Type) *TypeUnion {
 	sort.SliceStable(types, func(i, j int) bool {
 		return CompareTypes(types[i], types[j]) < 0
 	})
-	tv := EncodeTypeValue(&TypeUnion{Types: types})
+	out := make([]Type, 0, len(types))
+	var prev Type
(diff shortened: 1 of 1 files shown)


mccanne (commented):
I think this check should go in zngio.Validate and zson.

mattnibs:
> I think this check should go in zngio.Validate and zson.
@mccanne would this result in an error?

mccanne:
> > I think this check should go in zngio.Validate and zson.
> > @mccanne would this result in an error?

ZNG yes an error (and we should clarify the spec).  ZSON would be flexible and fix it.

mccanne:
> > > I think this check should go in zngio.Validate and zson.
> > > @mccanne would this result in an error?
> 
> ZNG yes an error (and we should clarify the spec). ZSON would be flexible and fix it.

ZNG spec already says "The type IDs of a union must be unique.".  We should clariify that they must follow the Zed type total order.

mattnibs:
I'm going to close this and we'll revisit this later.

PR #3548, opened 2022-01-31
Add flatten function

mattnibs:
Add flatten function which flattens record values, returning a map where
where keys are the path to each value, and value is a union of all
non-record values.

### DIFF
--- a/docs/language/functions.md
+++ b/docs/language/functions.md
@@ -23,6 +23,7 @@
   - [`parse_zson`](#parse_zson)
 - [Records](#records)
   - [`fields`](#fields)
+  - [`flatten`](#flatten)
   - [`unflatten`](#unflatten)
 - [Strings](#strings)
   - [`join`](#join)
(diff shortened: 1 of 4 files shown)


docs/language/functions.md
```diff
@@ -426,6 +427,28 @@ echo '{a:1,b:2,c:{d:3,e:4}}' | zq -z 'over fields(this) | yield join(this,".")'
...
 
+### `into`
+
+```
+into(r record) -> |{[string]:[<any>]}|
+```
+
+`into` returns a map of the flattened key/values of record r where key is a
+string array of the path to each flattened non-record value. If there are
+multiple types for the values in r the type of each value will be the union of
```
nwt, line 438:
```suggestion
multiple types for the values in r, then the type of each value will be the union of
```

expr/function/into.go
```diff
@@ -0,0 +1,104 @@
...
+			valType = n.zctx.LookupTypeUnion(types)
+		}
+		n.mapper.EnterType(typ.ID(), valType)
+	}
+	if n.keyType == nil {
+		n.keyType = n.zctx.LookupTypeArray(zed.TypeString)
+	}
+	n.Reset()
+	n.encode(typ.Columns, valType, field.Path{}, val.Bytes)
+	return ctx.NewValue(n.zctx.LookupTypeMap(n.keyType, valType), n.Bytes()).Copy()
```
nwt, line 48: No need for `Copy` here since the `zcode.Builder.Bytes` caller own the returned slice.
```suggestion
	return ctx.NewValue(n.zctx.LookupTypeMap(n.keyType, valType), n.Bytes())
```

expr/function/into.go
```diff
@@ -0,0 +1,104 @@
...
+	sort.SliceStable(types, func(i, j int) bool {
+		return zed.CompareTypes(types[i], types[j]) < 0
+	})
+	out := make([]zed.Type, 0, len(types))
+	var prev zed.Type
+	for _, typ := range types {
+		if typ != prev {
+			out = append(out, typ)
+		}
+		prev = typ
```
nwt, line 61:
```suggestion
			out = append(out, typ)
			prev = typ
		}
```

expr/function/into.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+	return out
+}
+
+func (n *Into) encode(cols []zed.Column, inner zed.Type, base field.Path, b zcode.Bytes) {
+	iter := b.Iter()
+	for _, col := range cols {
+		val := iter.Next()
```
nwt, line 69: At this point, nearly all of our iterator variables are called `it`.
```suggestion
	it := b.Iter()
	for _, col := range cols {
		val := it.Next()
```

nwt (approved):
Nice!

PR #3550, opened 2022-01-31
Add a function that get's the path of the executable.

jameskerr:
This will be very useful for our brim tests.

### DIFF
--- a/npm/index.d.ts
+++ b/npm/index.d.ts
@@ -1,0 +1,1 @@
+export declare function getPath(name: "zed" | "zq"): string;
--- a/npm/index.js
+++ b/npm/index.js
@@ -1,0 +1,11 @@
+const path = require("path");
+const os = require("os");
+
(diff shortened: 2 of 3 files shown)


nwt:
@jameskerr: I'd like minimize the amount of JavaScript here. Can this go in brimdata/brim instead?

mattnibs:
@jameskerr I echo @nwt comment. Can this logic go in brim?

jameskerr:
Yeah sure it can go in brim. But having this here would be very convenient. Might I ask why you'd like to not have too much javascript in here?

jameskerr:
Why don't we have a talk at the meeting tomorrow about the zed executables and how they get consumed. There have been several solutions without holistic thoughts behind them. I'm sure there's a good way forward.

jameskerr:
I am closing this in favor of a more thoughtful solution we will think of later.

PR #3553, opened 2022-02-02
StringLiteral: fix empty string

mattnibs:
Fix bug on empty strings introduced with template literals.


compiler/parser/parser.peg
```diff
@@ -1036,7 +1036,7 @@ ComplexType
 
 TemplateLiteral
   = v:TemplateLiteralParts {
-      if ISNULL(v) {
+      if ARRAY_ISLEN(v, 0) {
```
nwt, line 1039: My vote's for `if (ARRAY_LEN(v) == 0) {` but this is OK.

nwt (approved):

compiler/parser/parser.peg
mattnibs: I'll try this.

PR #3554, opened 2022-02-02
change data_object row_size to size

mccanne:


mattnibs (approved):
Going to cause backward incompatibility so we best get this in before the next release.

nwt (approved):

mccanne:
Yes, this breaks current lakes but now is the time to change it!  It's been bugging me in demos.  I had that there thinking ZST size would be column_size but that metadata will go somewhere else...

PR #3557, opened 2022-02-03
Wrap zson type errors with zson.FormatType

mattnibs:
Make zson type errors more legible by wrapping them with zson.String.

### DIFF
--- a/zson/analyzer.go
+++ b/zson/analyzer.go
@@ -168,7 +168,7 @@ func (a Analyzer) typeCheck(cast, parent zed.Type) error {
 		// as any union incompability will be caught in convertAnyValue().
 		return nil
 	}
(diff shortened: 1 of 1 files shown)


zson/analyzer.go
```diff
@@ -503,7 +503,7 @@ func (a Analyzer) convertEnum(zctx *zed.Context, val *astzed.Enum, cast zed.Type
 			}, nil
 		}
 	}
-	return nil, fmt.Errorf("symbol %q not a member of type %q", val.Name, FormatType(enum))
+	return nil, fmt.Errorf("symbol %q not a member of type %q", val.Name, String(enum))
```
nwt, line 506: String's OK but I vote for using FormatType everywhere instead since it's type-safe.

nwt (approved):

PR #3574, opened 2022-02-07
improve zed.Context.LookupType* performance

nwt:
zio/jsonio.Reader tends to call at least one zed.Context.LookupType*
method per value read.  Speed up those methods with two improvements.

1. Some zed.Context.LookupType* methods allocate type values that are
   discarded if the type already exists in the the context.  Obtain type
   value storage from a sync.Pool to reduce allocations.

2. zed.Context.LookupTypeRecord uses a map to check for duplicate
   fields.  Replace with a faster sorted list obtained from a sync.Pool.

### DIFF
--- a/context.go
+++ b/context.go
@@ -91,6 +91,15 @@ func (c *Context) Lookup(id int) *TypeRecord {
 	return nil
 }
 
+var tvPool = sync.Pool{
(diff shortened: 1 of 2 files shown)


mccanne (approved):
cool

PR #3581, opened 2022-02-07
speed up zio/jsonio.Reader with pkg/jsonlexer

nwt:
zio/jsonio.Reader performance suffers from the large number of
allocations performed by encoding/json.Decoder.Token.  Replace it with
pkg/jsonlexer, a JSON lexer that minimizes the number of allocations.

There are a number of JSON lexers in Go out there, and I'd hoped to use one, but all were slower than I'd hoped.

Reading NDJSON with ~~`-f json`~~ `-i json` is about three times faster for me with this branch compared to main.

### DIFF
--- a/pkg/jsonlexer/lexer.go
+++ b/pkg/jsonlexer/lexer.go
@@ -1,0 +1,171 @@
+package jsonlexer
+
+import (
+	"bufio"
+	"errors"
+	"fmt"
+	"io"
+)
+
+type Token byte
+
+const (
+	TokenErr Token = iota
+
+	TokenBeginArray
+	TokenBeginObject
+	TokenEndArray
+	TokenEndObject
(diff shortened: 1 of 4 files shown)


mccanne (approved):
Very cool.

unquoteBytes looks like zson.Lexer.scanString() but done differently (the specs are the same).  Should we combine them?

mccanne:
> Reading NDJSON with `-f json` is about three times faster for me with this branch compared to main.

You mean `-i json`, right?

nwt:
> unquoteBytes looks like zson.Lexer.scanString() but done differently (the specs are the same). Should we combine them?

Yep. I didn't want to deal with that hear but it's definitely something we should do.

> You mean `-i json`, right?

I do!

PR #3584, opened 2022-02-08
uniq -c: wrap value in new record

### LINKED ISSUE #3582: uniq -c panics on non-record inputs
Repro panic with:
```
echo 1 | zq -z 'uniq -c' -
```

Let's simplify things here and output records of the format `{value:any,count:uint64}` instead of adding a column to the input record.

The uniq operator is documented this way in the upcoming docs-update PR.

mattnibs:
For uniq -c wrap the output in {value:any,count:uint64}. This ensures
that uniq -c also works for non-record values.

Closes #3582


mccanne (approved):
cool

PR #3587, opened 2022-02-08
put: make proc work with quiet

### LINKED ISSUE #3579: quiet should work with put
This should emit empty record:
```
echo '{}' | zq 'x:=quiet(y)' -
```
but emits
```
{x:error("quiet")}
```

mattnibs:
Like the cut proc, put should work with quiet and ignore expressions
that return missing.

Closes #3579

### DIFF
--- a/expr/putter.go
+++ b/expr/putter.go
@@ -19,7 +19,9 @@ type Putter struct {
 	zctx    *zed.Context
 	builder zcode.Builder
 	clauses []Assignment
-	// vals is a fixed array to avoid re-allocating for every record
+	// valClauses is a slice to avoid re-allocating for every value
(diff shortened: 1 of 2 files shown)


expr/putter.go
```diff
@@ -19,6 +19,8 @@ type Putter struct {
 	zctx    *zed.Context
 	builder zcode.Builder
 	clauses []Assignment
+	// valClauses is a fixed array to avoid re-allocating for every record
+	valClauses []Assignment
 	// vals is a fixed array to avoid re-allocating for every record
```
nwt, line 24: Nit: These are slices, not arrays.

nwt (approved):

PR #3589, opened 2022-02-08
Make record literal an implicit yield

### LINKED ISSUE #3576: record literal as default yield
A record literal by itself may appear as an operator in which implied operator is yield.  When combined with record spreads, this makes generating jq-like transformation very easy, e.g.,
```
echo '{a:{x:1},b:2,c:3}' | zq '{...a,y:b+c}' -
```
would give
```
{x:1,y:5}
```

mattnibs:
Closes #3576

### DIFF
--- a/compiler/parser/parser.peg
+++ b/compiler/parser/parser.peg
@@ -92,7 +92,7 @@ Operation
   / a:OpAssignment &EndOfOp { RETURN(a) }
   / a:Aggregation &EndOfOp  { RETURN(a) }
   / r:Record &EndOfOp {
-	    RETURN(MAP("kind":"Yield", "exprs":ARRAY(r)))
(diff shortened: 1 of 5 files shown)


compiler/parser/parser.peg
```diff
@@ -91,6 +91,9 @@ Operation
   / f:Function &EndOfOp { RETURN(f) }
   / a:OpAssignment &EndOfOp { RETURN(a) }
   / a:Aggregation &EndOfOp  { RETURN(a) }
+  / r:Record &EndOfOp {
+	    RETURN(MAP("kind":"Yield", "exprs":ARRAY(r)))
```
nwt, line 95:
```suggestion
      RETURN(MAP("kind": "Yield", "exprs": ARRAY(r)))
```

nwt (approved):

PR #3591, opened 2022-02-08
fix Parquet writer data corruption

nwt:
zio/parquetio.Writer.Write retains pointers into zed.Value.Bytes for Zed
bytes and string values.  This can lead to corruption if that storage is
reused before the current row group is flushed to the Parquet file.  Fix
by copying instead of retaining pointers for bytes and string values in
parquetio.newData.


mccanne (approved):
Looks good but  `case *zed.TypeOfType:` also needs attention, but instead of copying the byte slice,  it should return the zson string of the type value.

nwt:
> Looks good but case *zed.TypeOfType: also needs attention, but instead of copying the byte slice, it should return the zson string of the type value.

I'll do that in a separate PR.

nwt:
The separate PR is #3592.

PR #3593, opened 2022-02-08
consolidate source files in top-level package zed

mccanne:
This commit consolidates various sources files in the top-level
zed package into fewer files, in part, to reduce the clutter in the
github repository listing.  There are a few more moves to make but
this is a start.


zed_test.go
```diff
@@ -236,3 +243,193 @@ func findInputs(t *testing.T, dirs map[string]struct{}, script string, isValidIn
 	}
 	return out, nil
 }
+
+func TestRecordAccessNamed(t *testing.T) {
```
nwt, line 247: I don't think we should move any of these tests in here since it's conventional to put `x.go` tests in `x_test.go`.

(The one move I'd make is to rename `set_test.go` to `complex_test.go`.)

mccanne: This conflicts with the goal of narrowing down the zed package files.  Is there something else we could do to not have the *_test.go files at the top level?

mccanne:
zed.Builder has got to go.

builder_test.go
```diff
@@ -4,12 +4,13 @@ import (
...
 	"testing"
 
+	"inet.af/netaddr"
+
 	"github.com/brimdata/zed"
 	"github.com/brimdata/zed/zcode"
 	"github.com/brimdata/zed/zio/zsonio"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"inet.af/netaddr"
```
nwt, line 12: Nit: Don't move inet.af/netaddr.

nwt (approved):

nwt:
> zed.Builder has got to go.

Truth.

PR #3599, opened 2022-02-09
Select variable path for implied quiet assignment

### LINKED ISSUE #3596: quiet() LHS name shouldn't be quiet
The code that derives the LHS id (when omitted) from the RHS expression should handle quiet() by looking at the argument of quiet instead of using "quiet" as the name. 

So this query...
```
echo '{a:1}{b:1}' | zq 'cut quiet(b)' -
```
wouldn't give
```
{quiet:1}
```
but instead would give
```
{b:1}
```

mattnibs:
Closes #3596

### DIFF
--- a/compiler/semantic/expr.go
+++ b/compiler/semantic/expr.go
@@ -415,13 +415,20 @@ func semAssignment(scope *Scope, a ast.Assignment, summarize bool) (dag.Assignme
 			return dag.Assignment{}, fmt.Errorf("lhs of assigment expression: %w", err)
 		}
 	} else if call, ok := a.RHS.(*ast.Call); ok {
-		// If LHS is nil and the call is every() make the LHS field ts since
-		// field ts assumed with every.
-		name := call.Name
-		if name == "every" {
-			name = "ts"
+		path := []string{call.Name}
+		switch call.Name {
+		case "every":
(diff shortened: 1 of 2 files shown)


compiler/semantic/expr.go
```diff
@@ -415,13 +415,26 @@ func semAssignment(scope *Scope, a ast.Assignment, summarize bool) (dag.Assignme
...
+		case "every":
+			// If LHS is nil and the call is every() make the LHS field ts since
+			// field ts assumed with every.
+			path = []string{"ts"}
+		case "quiet":
+			if len(call.Args) > 0 {
+				if p, ok := rhs.(*dag.Call).Args[0].(*dag.This); ok {
+					path = p.Path
+				} else {
+					return dag.Assignment{}, fmt.Errorf("implied quiet variable must be a path")
```
mccanne, line 429: Shouldn't this be the same error as the default case?

mattnibs: I'm not quite sure I follow. If do something like `cut quiet(a.b[0])` this is valid selector but we can't imply a variable name here.

mccanne (approved):
I just mean it seems a little non-orthogonal to say something specific about quiet() but non-specific about all the other cases.   Maybe the behavior here should be that you get "quiet" like other function calls if the quiet body is not inferrable?

PR #3612, opened 2022-02-12
fix cast bugs

mccanne:
This commit fixes bug where casts of type time, duration, and net would
fail when the input type was already the desired type.

### DIFF
--- a/runtime/expr/cast.go
+++ b/runtime/expr/cast.go
@@ -144,6 +144,9 @@ type casterNet struct {
 }
 
 func (c *casterNet) Eval(ectx Context, val *zed.Value) *zed.Value {
+	if val.Type.ID() == zed.IDNet {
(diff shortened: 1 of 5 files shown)


expr/cast.go
```diff
@@ -191,6 +197,9 @@ type casterTime struct {
 
 func (c *casterTime) Eval(ectx Context, val *zed.Value) *zed.Value {
 	id := val.Type.ID()
+	if id == zed.IDTime {
+		return ectx.CopyValue(*val)
+	}
 	var ts nano.Ts
 	switch {
```
nwt, line 204:
```suggestion
	var ts nano.Ts
	switch {
	case id == zed.IDTime:
		return ectx.CopyValue(*val)
```

nwt (approved):

Issue #3613, opened 2022-02-12
math functions and float32


mccanne:
Some math functions like sqrt() always return float64 while other math functions like (ceil) return a float32 for a float32.  We should revisit these type decisions and possibly change things to make them more consistent.

See test in `expr/function/ztests/float32.yaml`

PR #3614, opened 2022-02-12
fix float32 bugs

mccanne:
This commit fixes bugs in the math functions where float32's were
being decoded as float64's causing panics.  The new test reveals
that some functions turn float32's into float64's.  We might want
to make this more consistent as described in issue #3613.

### DIFF
--- a/runtime/expr/function/function.go
+++ b/runtime/expr/function/function.go
@@ -168,6 +168,10 @@ func newUint(ctx zed.Allocator, typ zed.Type, native uint64) *zed.Value {
 	return ctx.NewValue(typ, zed.EncodeUint(native))
 }
 
+func newFloat32(ctx zed.Allocator, native float32) *zed.Value {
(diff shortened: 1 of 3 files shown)


expr/function/math.go
```diff
@@ -101,7 +111,12 @@ func (r *reducer) Call(ctx zed.Allocator, args []zed.Value) *zed.Value {
 	if zed.IsFloat(id) {
 		//XXX this is wrong like math aggregators...
 		// need to be more robust and adjust type as new types encountered
-		result := zed.DecodeFloat64(zv.Bytes)
+		var result float64
+		if id == zed.IDFloat64 {
+			result = zed.DecodeFloat64(zv.Bytes)
+		} else {
+			result = float64(zed.DecodeFloat32(zv.Bytes))
+		}
```
nwt, line 119:
```suggestion
		result := zed.DecodeFloat(zv.Bytes)
```

nwt (approved):

PR #3619, opened 2022-02-13
release: use goreleaser

mattnibs:
Use goreleaser to create release assets, publish them to github as a
release and update the brew taps in github.com/brimdata/homebrew-tap to
the latest version.

This change also adds arm64 builds to releases for linux, macos, and
windows.


.github/workflows/release.yml
```diff
@@ -24,12 +24,10 @@ jobs:
...
+    - uses: goreleaser/goreleaser-action@v2
       with:
-        repo_token: ${{ secrets.GITHUB_TOKEN }}
-        tag: ${{ github.ref }}
-        file: dist/release/*
-        file_glob: true
-        overwrite: true
+        distribution: goreleaser
+        version: latest
+        args: release --rm-dist -f scripts/goreleaser.yaml
```
nwt, line 31: Any reason not to put the config file in the standard location (`.goreleaser.yaml` in the root directory)?

.github/workflows/release.yml
```diff
@@ -24,12 +24,10 @@ jobs:
     - run: make test-unit
     - run: make test-system
     - run: make test-heavy
-    - run: make create-release-assets
```
nwt, line 27: Might as well remove the `create-release-assets` target from `Makefile`.

scripts/goreleaser.yaml
```diff
@@ -0,0 +1,80 @@
...
+      owner: brimdata
+      name: homebrew-tap
+    commit_author:
+      name: brim-bot
+      email: bot@brimdata.io
+    homepage: "https://github.com/brimdata/zed"
+    description: |
+      A command-line tool for creating, configuring, ingesting into, querying,
+      and orchestrating Zed data lakes.
+    folder: Formula
```
nwt, line 57: Might as well put fromulae in the root of the repository since they'll be the only things there.
```suggestion
```

README.md
```diff
@@ -69,13 +69,13 @@ If you have Go installed, you can easily install `zed` and `zq` by running
 ```
 go install github.com/brimdata/zed/cmd/{zed,zq}@latest
 ```
-On a Mac, you can use `brew` to install both `zq` and `zed`:
+On a Mac, you can use `brew` to install `zq`:
```
nwt, line 72:
```suggestion
On macOS and Linux, you can use `brew` to install `zq`:
```

.github/workflows/release.yml
mattnibs: I just didn't want to add another file to the root when we're trying to cut down the number of files here, is this silly?

scripts/goreleaser.yaml
mattnibs: Good point.

.github/workflows/release.yml
mattnibs: I'll just put this in the standard location.

.goreleaser.yaml
```diff
@@ -0,0 +1,69 @@
...
+      - windows
+      - darwin
+archives:
+  - name_template: zed-{{ .Version }}.{{ .Os }}-{{ .Arch }}
+    format_overrides:
+      - goos: windows
+        format: zip
+    files:
+      - LICENSE*
+      - README*
```
nwt, line 35: I think we want these two files.
```suggestion
      - LICENSE.txt
      - acknowledgments.txt
```

nwt (approved):

Issue #3604, opened 2022-02-10
docs update v2


mccanne:
Finish updating docs for Feb 2022 release...

* finish zq language description and drop expr-TODO.md
* convert lake docs to zed docs 
* move lake format to a design doc 
* update API docs and give examples in zed/python/curl with many python lib calls TBD

mccanne:
This was completed with the big docs project.

PR #3620, opened 2022-02-14
change filter to search/where and add implied yield

mccanne:
This commit changes the filter keyword to "where" and also allows
"search" to explicitly cause a parse of a keyword boolean search.
To do this, we changed ast.Filter to ast.OpExpr, which represents
an unresolved expression that requires semantic analysis to
determine if it is a boolean search, an implied yield, or an
simply function-call style aggregation.

Though "where/search" are now operator keywords, the filter operator
is still the implementation of these concepts so dag.Filter
remains and there is no dag.Where.

More thorough documentation for all of this will be soon merged
as part of issue #3604.  The stuff in where.md is draft-y and
will move around.


compiler/parser/parser.peg
```diff
@@ -88,14 +88,16 @@ Operation
...
+  / "search" _ expr:SearchBoolean {
+      RETURN(MAP("kind": "Where", "expr": expr))
     }
-  / expr:SearchBoolean !AggGuard {
-      RETURN(MAP("kind": "Filter", "expr": expr))
+  / expr:SearchBoolean {
+      RETURN(MAP("kind": "OpExpr", "expr": expr))
+    }
+  / expr:Expr {
+      RETURN(MAP("kind": "OpExpr", "expr": expr))
```
nwt, line 100: Maybe combine?
```suggestion
  / expr:(Expr / SearchBoolean) {
      RETURN(MAP("kind": "OpExpr", "expr": expr))
```

compiler/semantic/proc.go
```diff
@@ -645,3 +646,78 @@ func semOpAssignment(scope *Scope, p *ast.OpAssignment) (dag.Op, error) {
...
+	case *dag.Search, *dag.RegexpMatch, *dag.RegexpSearch:
+		return true
+	default:
+		return false
+	}
+}
+
+func semCallOp(scope *Scope, call *ast.Call) (dag.Op, error) {
+	agg, err := maybeConvertAgg(scope, call)
+	if err == nil && agg != nil {
```
nwt, line 700:
```suggestion
	if agg, err := maybeConvertAgg(scope, call); agg != nil && err == nil {
```

docs/zq/operators/where.md
```diff
@@ -1,31 +1,29 @@
...
+[where] <expr>
 ```
 ### Description
 
-The `filter` operator copies a filtered version of its input to its output by
-applying a [search expression](../language.md#search-expressions) to each
-input value and dropping each value not matched by the expression.
+The `where` operator filters its input by applying a Boolean expression `<expr>`
+to each input value and dropping each value for which the expression is `false`
+or an error.
```
nwt, line 13:
```suggestion
to each input value and dropping each value for which the expression evaluates
to `false` or to an error.
```

zfmt/ast.go
```diff
@@ -442,6 +445,22 @@ func (c *canon) pool(p *ast.Pool) {
...
+		}
+		return &ast.Summarize{
+			Kind: "Summarize",
+			Aggs: []ast.Assignment{{
+				Kind: "Assignment",
+				RHS:  call,
+			}},
+		}
+	}
+	return nil
```
nwt, line 461: NIt: Indent less.
```suggestion
	call, ok := e.(*ast.Call)
	if !ok {
		return nil
	}
	if _, err := agg.NewPattern(call.Name); err != nil {
		return nil
	}
	return &ast.Summarize{
		Kind: "Summarize",
		Aggs: []ast.Assignment{{
			Kind: "Assignment",
			RHS:  call,
		}},
	}
```

nwt (approved):

compiler/parser/parser.peg
mccanne: I think I will leave it since the PEG order matters and separate lines stress that.  You introduced a bug where identifiers will be matched before keywords so I believe `foo` would be `yield foo` instead of `search("foo")`

jameskerr:
Now that this is changed, the program.test.ts test in brim is failing and I'm looking for help in a new solution for how to fix it. Here is an example of a failing test.

Issue #3615, opened 2022-02-12
fix flatten/map problem


mccanne:
With flatten using a map value, there is no way to retain the field order of the original record.  I think we should instead use `[{key:[string],value:any}]` for flatten and change `over` on a record to generate a sequence of `{key:[string],value:any}` without the need to create a common union type for the array.

mattnibs:
Yeah this is a good idea.

PR #3624, opened 2022-02-14
flatten: array of key/value records

mattnibs:
Change flatten to return an array of key/value records where the
key is a string array of the value path and value is the value (any
type).

Part of #3615

### DIFF
--- a/runtime/expr/function/flatten.go
+++ b/runtime/expr/function/flatten.go
@@ -1,7 +1,6 @@
 package function
 
 import (
-	"fmt"
 	"sort"
 
 	"github.com/brimdata/zed"
@@ -20,7 +19,7 @@ type Flatten struct {
 
 func NewFlatten(zctx *zed.Context) *Flatten {
 	return &Flatten{
-		entryTypes: make(map[int]zed.Type),
+		entryTypes: make(map[zed.Type]zed.Type),
 		keyType:    zctx.LookupTypeArray(zed.TypeString),
 		mapper:     zed.NewMapper(zctx),
 		zctx:       zctx,
@@ -55,7 +54,7 @@ func (n *Flatten) collectTypes(cols []zed.Column) []zed.Type {
 		if typ := zed.TypeRecordOf(col.Type); typ != nil {
 			types = append(types, n.collectTypes(typ.Columns)...)
 		} else {
-			typ, ok := n.entryTypes[col.Type.ID()]
+			typ, ok := n.entryTypes[col.Type]
 			if !ok {
 				n.entryTypes[col.Type] = n.zctx.MustLookupTypeRecord([]zed.Column{
 					zed.NewColumn("key", n.keyType),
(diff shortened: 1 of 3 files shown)


runtime/expr/function/flatten.go
```diff
@@ -11,15 +12,18 @@ import (
 // https://github.com/brimdata/zed/blob/main/docs/language/functions.md#flatten
 type Flatten struct {
 	zcode.Builder
-	keyType zed.Type
-	mapper  *zed.Mapper
-	zctx    *zed.Context
+	keyType    zed.Type
+	mapper     *zed.Mapper
+	entryTypes map[int]zed.Type
```
nwt, line 17: Nit: This obviously works either way, but I think the intention is clearer if it's a map from zed.Type to zed.Type.
```suggestion
	entryTypes map[zed.Type]zed.Type
```

runtime/expr/function/flatten.go
```diff
@@ -40,12 +44,35 @@ func (n *Flatten) Call(ctx zed.Allocator, args []zed.Value) *zed.Value {
...
+			if !ok {
+				var err error
+				typ, err = n.zctx.LookupTypeRecord([]zed.Column{
+					zed.NewColumn("key", n.keyType),
+					zed.NewColumn("value", col.Type),
+				})
+				if err != nil {
+					panic(fmt.Errorf("flatten: error creating record type %w", err))
+				}
+				n.entryTypes[col.Type.ID()] = typ
```
nwt, line 70:
```suggestion
				n.entryTypes[col.Type] = n.zctx.MustLookupTypeRecord([]zed.Column{
					zed.NewColumn("key", n.keyType),
					zed.NewColumn("value", col.Type),
				})
```

runtime/expr/function/flatten.go
```diff
@@ -40,12 +44,35 @@ func (n *Flatten) Call(ctx zed.Allocator, args []zed.Value) *zed.Value {
...
+	return ctx.NewValue(n.zctx.LookupTypeArray(valType), n.Bytes())
+}
+
+func (n *Flatten) collectTypes(cols []zed.Column) []zed.Type {
+	var types []zed.Type
+	for _, col := range cols {
+		if typ := zed.TypeRecordOf(col.Type); typ != nil {
+			for _, typ := range n.collectTypes(typ.Columns) {
+				types = append(types, typ)
+			}
```
nwt, line 58:
```suggestion
			types = append(types, n.collectTypes(typ.Columns)...)
```

nwt (approved):

runtime/expr/function/flatten.go
mattnibs: Good point.

mattnibs:
I need to update the flatten docs.

runtime/expr/function/flatten.go
mattnibs: This doesn't set `typ` but I'll fix this.

PR #3632, opened 2022-02-15
TypeUnion.SplitZNG: panic on invalid selector

mattnibs:


complex.go
```diff
@@ -374,15 +374,15 @@ func (t *TypeUnion) Selector(typ Type) int {
 }
 
 // SplitZNG takes a zng encoding of a value of the receiver's union type and
-// returns the concrete type of the value, its selector, and the value encoding.
-func (t *TypeUnion) SplitZNG(zv zcode.Bytes) (Type, int64, zcode.Bytes, error) {
+// returns the concrete type of the value, its selector, and the value encoding. SplitZNG panics if the selector of zv is not a member of the union.
```
nwt, line 377:
```suggestion
// returns the concrete type of the value, its selector, and the value encoding.
// SplitZNG panics if the selector is invalid.
```

nwt (approved):
s/union.SplitZNG/TypeUnion.SplitZNG/ in the PR title and merge commit message since there's no union.SplitZNG.

PR #3637, opened 2022-02-15
union agg: Allow mixed types

### LINKED ISSUE #3610: union aggregator mixed types
A union() on heterogeneous types should create an output that is type union of the types encountered.  Currently, union() discards types that don't conform to the first type seen.

mattnibs:
Allow mixed types in the union aggregation function. If multiple types
exist the resultant set is of type union.

Closes #3610


runtime/expr/agg/union.go
```diff
@@ -6,36 +6,33 @@ import (
...
-	if _, ok := u.val[string(b)]; !ok {
-		u.val[string(b)] = struct{}{}
+func (u *Union) update(typ zed.Type, b zcode.Bytes) {
+	m, ok := u.types[typ]
+	if !ok {
+		m = make(map[string]struct{})
+		u.types[typ] = m
+	}
+	if _, ok := m[string(b)]; !ok {
+		m[string(b)] = struct{}{}
```
nwt, line 35: Since `string([]byte{})` and `string([]byte(nil))` both evaluate to `""`, this won't distinguish between zero-length and nil values of `b`. One way to deal with that would be to change `Union.types` to `map[zed.Type]map[string]bool` and then store true if the `string` is `""` but represents a nil value.

runtime/expr/agg/union.go
```diff
@@ -47,34 +44,59 @@ func (u *Union) update(b zcode.Bytes) {
...
+	inner := types[0]
+	if len(types) > 1 {
+		inner = zctx.LookupTypeUnion(types)
+	}
 	var b zcode.Builder
-	for v := range u.val {
-		b.Append([]byte(v))
+	for typ, m := range u.types {
+		for v := range m {
+			if union, ok := inner.(*zed.TypeUnion); ok {
```
nwt, line 75:
```suggestion
			if union, ok := zed.TypeUnder(inner).(*zed.TypeUnion); ok {
```

runtime/expr/agg/union.go
```diff
@@ -47,34 +44,59 @@ func (u *Union) update(b zcode.Bytes) {
...
+	}
+	styp, ok := val.Type.(*zed.TypeSet)
+	if !ok {
+		panic("union: partial not a set type")
 	}
 	for it := val.Iter(); !it.Done(); {
-		u.update(it.Next())
+		typ := styp.Type
+		b := it.Next()
+		if union, ok := typ.(*zed.TypeUnion); ok {
```
nwt, line 96:
```suggestion
		if union, ok := zed.TypeUnder(typ).(*zed.TypeUnion); ok {
```

type.go
```diff
@@ -374,6 +375,24 @@ func TypeID(typ Type) int {
 	return typ.ID()
 }
 
+// CanonicalUnionOfTypes returns the set of unique Types in types in sorted
+// order. types will be sorted in place, but a new slice of unique types will
+// be returned.
+func CanonicalUnionOfTypes(types []Type) []Type {
```
nwt, line 381: Maybe call this `UniqueTypes`?

type.go
```diff
@@ -374,6 +375,24 @@ func TypeID(typ Type) int {
...
+	sort.SliceStable(types, func(i, j int) bool {
+		return CompareTypes(types[i], types[j]) < 0
+	})
+	out := make([]Type, 0, len(types))
```
nwt, line 385: Any reason not to reuse `types`'s storage here (and adjust the doc comment accordingly)?
```suggestion
	out := types[:0]
```

type.go
mattnibs: Doesn't quite capture that its also sorted, but I like the simpler name.

runtime/expr/agg/union.go
mattnibs: Actually this is moot since null values are ignored by the union aggregator.

nwt: Excellent point! Sorry for missing that.

nwt (approved):

PR #3644, opened 2022-02-18
remove apt-get commands from .github/workflows/release.yml

nwt:
The release workflow installs some OS packages that haven't been needed
for a long time.  Remove the apt-get commands that install them.


mattnibs (approved):
🎉

PR #3647, opened 2022-02-18
Aggregations: check arg

mattnibs:
When creating an aggregation function if the aggregation requires an
argument but receives none, return an error.

### DIFF
--- a/compiler/semantic/expr.go
+++ b/compiler/semantic/expr.go
@@ -488,7 +488,7 @@ func semField(scope *Scope, f ast.Expr) (*dag.This, error) {
 }
 
 func maybeConvertAgg(scope *Scope, call *ast.Call) (dag.Expr, error) {
-	if p, _ := agg.NewPattern(call.Name, true); p == nil {
+	if _, err := agg.NewPattern(call.Name, true); err != nil {
 		return nil, nil
 	}
 	var e dag.Expr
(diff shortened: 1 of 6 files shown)


compiler/semantic/expr.go
```diff
@@ -488,7 +488,7 @@ func semField(scope *Scope, f ast.Expr) (*dag.This, error) {
 }
 
 func maybeConvertAgg(scope *Scope, call *ast.Call) (dag.Expr, error) {
-	if _, err := agg.NewPattern(call.Name); err != nil {
+	if p, _ := agg.NewPattern(call.Name, true); p == nil {
```
nwt, line 491: Nit: I think you should leave these all of these tests in terms of `err`.

runtime/expr/agg/ztests/args.yaml
```diff
@@ -0,0 +1,33 @@
...
+  ! zq -z 'any()' in.zson
+  ! zq -z 'collect()' in.zson
+  zq -z 'count()' in.zson
+  ! zq -z 'dcount()' in.zson
+  ! zq -z 'fuse()' in.zson
+  ! zq -z 'min()' in.zson
+  ! zq -z 'max()' in.zson
+  ! zq -z 'or()' in.zson
+  ! zq -z 'sum()' in.zson
+  ! zq -z 'union()' in.zson
```
nwt, line 12: Maybe shorten?
```suggestion
  zq -z 'count()' in.zson
  for agg in and any collect dcount fuse min max or sum union; do
    ! zq -z "$agg()" in.zson
  done
```

nwt (approved):

PR #3649, opened 2022-02-20
fix empty batches in runtime/op/traverse.Over

nwt:
The flowgraph for a scoped over operator (i.e., "over ... => (...)")
shuts down prematurely upon ingesting a null container in part because
the inner scope produces an empty batch.  Fix by modifying
runtime/op/traverse.Over so it does not produce empty batches.

There's still problem here because
```
echo '[1] [2]' | zq 'over this => (where this > 1)' -
```
produces no output, but I'm working on that.


mattnibs (approved):
test?

nwt:
> test?

Indeed. Added in cbe2157c17f66d9ca761fea9d235f05ffd0abb6b.

PR #3652, opened 2022-02-22
Make unflatten inverse of flatten

mattnibs:
Make unflatten inverse of flatten meaning the unflatten now accepts an
array of key value records and returns them into a single record. Rename
the former unflatten to zeek.

### DIFF
--- a/compiler/kernel/expr.go
+++ b/compiler/kernel/expr.go
@@ -380,7 +380,7 @@ func compileCall(zctx *zed.Context, call dag.Call) (expr.Evaluator, error) {
 			return nil, fmt.Errorf("has(): bad argument: %w", err)
 		}
 		return expr.NewHas(exprs), nil
-	case call.Name == "unflatten":
+	case call.Name == "nest_dotted":
 		return expr.NewUnflattener(zctx), nil
 	case isShaperFunc(call.Name):
 		return compileShaper(zctx, call)
--- a/docs/zq/functions/unflatten.md
+++ b/docs/zq/functions/unflatten.md
@@ -1,23 +1,33 @@
 ### Function
 
-&emsp; **unflatten** &mdash; transform a record with dotted names to a nested record
+&emsp; **unflatten** &mdash; transform an array of key/value records into a
+record.
 
 ### Synopsis
 
 ```
-unflatten(val: record) -> record
+unflatten(val: [{key:[string],value:any}) -> record
 ```
 ### Description
-The _unflatten_ function returns a copy of `val` with all dotted field names converted
-into nested records. If no argument is supplied to `unflatten`, `unflatten`
-operates on `this`.
+The _unflatten_ function converts the key/value records in array `val` into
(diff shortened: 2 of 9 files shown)


compiler/kernel/expr.go
```diff
@@ -317,7 +317,7 @@ func compileCall(zctx *zed.Context, call dag.Call) (expr.Evaluator, error) {
 			return nil, fmt.Errorf("has(): bad argument: %w", err)
 		}
 		return expr.NewHas(exprs), nil
-	case call.Name == "unflatten":
+	case call.Name == "zeek":
```
nwt, line 320: We're looking for a different name for this, right?

mattnibs: Oh yeah I'll change that

docs/zq/functions/unflatten.md
```diff
@@ -1,23 +1,24 @@
...
 ```
 ### Description
-The _unflatten_ function returns a copy of `val` with all dotted field names converted
-into nested records. If no argument is supplied to `unflatten`, `unflatten`
-operates on `this`.
+The _unflatten_ functions converts the key/values records in array `val` into
+a single record. _unflatten_ is the inverse of _flatten_ i.e., `unflatten(flatten(r))`
+will produce an equivalent record to `r`.
 
 ### Examples
```
nwt, line 17: Maybe lead with an example of `unflatten` on its own, like `echo '[{key:"a",value:1},{key:"b",value:2}] | zq 'yield unflatten()' -`?

runtime/expr/function/unflatten.go
```diff
@@ -0,0 +1,153 @@
...
+		return &val
+	}
+	if err := u.closeChildren(u.stack); err != nil {
+		return u.zctx.NewErrorf("unflatten: %v", err)
+	}
+	typ, err := u.typeOfRecord(u.stack[0])
+	if err != nil {
+		return u.zctx.NewErrorf("unflatten: %v", err)
+	}
+	return ctx.NewValue(typ, u.builder.Bytes()).Copy()
```
nwt, line 54:
```suggestion
	return ctx.NewValue(typ, u.builder.Bytes())
```

runtime/expr/function/unflatten.go
```diff
@@ -0,0 +1,153 @@
...
+
+type record struct {
+	names []string
+	types []zed.Type
+}
+
+func (r *record) reset() {
+	r.names = r.names[0:]
+	r.types = r.types[0:]
+}
```
nwt, line 65: You could use `[]zed.Column` instead of this new type.

runtime/expr/function/unflatten.go
mattnibs: See: https://github.com/brimdata/zed/pull/3660#discussion_r814153988

nwt (approved):

runtime/expr/function/unflatten.go
mattnibs: Truth.

PR #3657, opened 2022-02-23
add some missing zio/zngio.Reader.Close calls

nwt:
A zio/zngio.Reader can leak goroutines if not closed.


mattnibs (approved):
Good to have these fixed 👍🏻

PR #3660, opened 2022-02-23
zed: handle unions for map/array/sets

mattnibs:


runtime/expr/values.go
```diff
@@ -275,48 +243,105 @@ type Entry struct {
...
+		// If there's only two types and one type is null, return single
+		// non-null type.
+		if types[0] == zed.TypeNull {
+			return types[1]
+		} else if types[1] == zed.TypeNull {
+			return types[0]
+		}
 	}
-	return ectx.CopyValue(zed.Value{m.typ, zed.NormalizeMap(bytes)})
+	return zctx.LookupTypeUnion(types)
```
nwt, line 346: I think we should discard the null type whenever we have at least one other type.
```go
	 unique := types[:0]
	 for _, t := range zed.UniqueTypes(types) {
		if t != zed.TypeNull {
			unique = append(unique, t)
		}
	}
	if len(unique) == 0 {
		return zed.TypeNull
	}
	if len(unique) == 1 {
		return types[0]
	}
	return zctx.LookupTypeUnion(types)
```

runtime/expr/values.go
```diff
@@ -178,95 +178,63 @@ func (r *recordSpreadExpr) invalidate(object map[string]column) {
 }
 
 type ArrayExpr struct {
-	zctx    *zed.Context
-	typ     *zed.TypeArray
-	builder *zcode.Builder
+	builder zcode.Builder
+	elems   collectionBuilder
 	exprs   []Evaluator
+	zctx    *zed.Context
```
nwt, line 184: No need to adopt this if it doesn't warm your heart, but I find it helpful to group fields into stuff from the outside world that was passed to the constructor followed by a separate group of the purely internal stuff that's either created by the constructor or just left as the zero value.
```suggestion
	exprs []Evaluator
	zctx  *zed.Context

	builder zcode.Builder
	elems   collectionBuilder
```

nwt (approved):

runtime/expr/values.go
```diff
@@ -178,95 +178,63 @@ func (r *recordSpreadExpr) invalidate(object map[string]column) {
...
-		bytes = []byte{}
+	if len(a.elems.types) == 0 {
+		return ectx.NewValue(a.zctx.LookupTypeArray(zed.TypeNull), []byte{})
+	}
+	it := a.elems.iter(a.zctx)
+	for !it.done() {
+		it.appendNext(&a.builder)
 	}
-	return ectx.NewValue(a.typ, bytes)
+	return ectx.NewValue(a.zctx.LookupTypeArray(it.typ), a.builder.Bytes()).Copy()
```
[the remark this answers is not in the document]
mattnibs: Why does this not need a copy? Won't the bytes get overwritten one the next value evaluated?

runtime/expr/values.go
mattnibs: What then would the selector of the null value be then?

runtime/expr/values.go
```diff
@@ -275,48 +243,105 @@ type Entry struct {
...
+	c.types = append(c.types, val.Type)
+	c.bytes = append(c.bytes, val.Bytes)
+}
+
+func (c *collectionBuilder) iter(zctx *zed.Context) collectionIter {
+	// uniqueTypes must be copied since zed.UniqueTypes operates on the type
+	// array in place and thus we'll lose order.
+	for _, typ := range c.types {
+		c.uniqueTypes = append(c.uniqueTypes, typ)
+	}
```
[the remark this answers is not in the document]
mattnibs: doh

runtime/expr/values.go
mattnibs: Discussed this offline, will fix.

runtime/expr/values.go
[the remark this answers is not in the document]
nwt: No, because at present, `zcode.Builder.Reset` sets `Builder.bytes` to nil.

PR #3661, opened 2022-02-23
zed: Allow floats to have no fractional numbers

mattnibs:


nwt (approved):
This deserves a test!

PR #3662, opened 2022-02-24
dot expressions: work on type values

### LINKED ISSUE #3562: has on types
It seems like this should be true but it's not
```
echo '<{foo:int64}>' | zq -z 'yield has(foo)' -
```

mattnibs:
If a value passed into a dot expression is a type value of a record,
select the appropiated field in the record.

Closes #3562

### DIFF
--- a/runtime/expr/dot.go
+++ b/runtime/expr/dot.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/brimdata/zed"
 	"github.com/brimdata/zed/pkg/field"
+	"github.com/brimdata/zed/zcode"
 )
 
 type This struct{}
@@ -56,6 +57,9 @@ func ValueUnder(val *zed.Value) *zed.Value {
 
 func (d *DotExpr) Eval(ectx Context, this *zed.Value) *zed.Value {
 	rec := d.record.Eval(ectx, this)
+	if _, ok := rec.Type.(*zed.TypeOfType); ok {
+		return d.evalTypeOfType(ectx, rec.Bytes)
+	}
 	val := ValueUnder(rec)
 	recType, ok := val.Type.(*zed.TypeRecord)
 	if !ok {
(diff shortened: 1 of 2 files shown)


runtime/expr/dot.go
```diff
@@ -56,6 +57,9 @@ func ValueUnder(val *zed.Value) *zed.Value {
 
 func (d *DotExpr) Eval(ectx Context, this *zed.Value) *zed.Value {
 	rec := d.record.Eval(ectx, this)
+	if _, ok := rec.Type.(*zed.TypeOfType); ok {
+		return d.evalTypeOfType(ectx, rec.Bytes)
+	}
 	val := ValueUnder(rec)
```
nwt, line 63:
```suggestion
	val := ValueUnder(rec)
	if _, ok := val.Type.(*zed.TypeOfType); ok {
		return d.evalTypeOfType(ectx, val.Bytes)
	}

```

runtime/expr/dot.go
```diff
@@ -75,6 +79,16 @@ func (d *DotExpr) Eval(ectx Context, this *zed.Value) *zed.Value {
 	return ectx.NewValue(recType.Columns[idx].Type, field)
 }
 
+func (d *DotExpr) evalTypeOfType(ectx Context, vb zcode.Bytes) *zed.Value {
+	typ, _ := d.zctx.DecodeTypeValue(vb)
```
nwt, line 83: Nit: Call this `b`, `bytes`, or `zb`.
```suggestion
func (d *DotExpr) evalTypeOfType(ectx Context, b zcode.Bytes) *zed.Value {
	typ, _ := d.zctx.DecodeTypeValue(b)
```

mattnibs: ok

nwt (approved):

PR #3666, opened 2022-02-25
unions: Don't use selector for nulls

mattnibs:
Change union.SplitZNG so if the passed value is null (i.e. has no
selector) the returned type is the type of the union value.

This change has implications for unions in zst- the readers and writers
of which expected every value to have a selector. Add a presence
reader/writer to the union writer to handle null values in the union.

### DIFF
--- a/complex.go
+++ b/complex.go
@@ -376,14 +376,17 @@ func (t *TypeUnion) Selector(typ Type) int {
 // SplitZNG takes a zng encoding of a value of the receiver's union type and
 // returns the concrete type of the value, its selector, and the value encoding.
 // SplitZNG panics if the selector is invalid.
-func (t *TypeUnion) SplitZNG(zv zcode.Bytes) (Type, int64, zcode.Bytes) {
+func (t *TypeUnion) SplitZNG(zv zcode.Bytes) (Type, zcode.Bytes) {
 	it := zv.Iter()
+	if it.Done() {
+		return t, nil
+	}
 	selector := DecodeInt(it.Next())
 	inner, err := t.Type(int(selector))
 	if err != nil {
 		panic(err)
 	}
-	return inner, selector, it.Next()
+	return inner, it.Next()
 }
 
 func (t *TypeUnion) Kind() Kind {
--- a/complex.go
+++ b/complex.go
@@ -377,10 +377,10 @@ func (t *TypeUnion) Selector(typ Type) int {
 // returns the concrete type of the value, its selector, and the value encoding.
 // SplitZNG panics if the selector is invalid.
 func (t *TypeUnion) SplitZNG(zv zcode.Bytes) (Type, zcode.Bytes) {
-	it := zv.Iter()
-	if it.Done() {
+	if zv == nil {
 		return t, nil
 	}
+	it := zv.Iter()
 	selector := DecodeInt(it.Next())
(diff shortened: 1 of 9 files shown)


complex.go
```diff
@@ -376,14 +376,17 @@ func (t *TypeUnion) Selector(typ Type) int {
 // SplitZNG takes a zng encoding of a value of the receiver's union type and
 // returns the concrete type of the value, its selector, and the value encoding.
 // SplitZNG panics if the selector is invalid.
-func (t *TypeUnion) SplitZNG(zv zcode.Bytes) (Type, int64, zcode.Bytes) {
+func (t *TypeUnion) SplitZNG(zv zcode.Bytes) (Type, zcode.Bytes) {
 	it := zv.Iter()
+	if it.Done() {
+		return t, nil
+	}
```
nwt, line 383: We should test for nil. A non-nil, zero-length `zcode.Bytes` here is a bug and we should panic on that.
```suggestion
	if zv == nil {
		return t, nil
	}
	it := zv.Iter()
```

zst/column/union.go
```diff
@@ -24,11 +25,18 @@ func NewUnionWriter(typ *zed.TypeUnion, spiller *Spiller) *UnionWriter {
 		typ:      typ,
 		values:   values,
 		selector: NewIntWriter(spiller),
+		presence: NewPresenceWriter(spiller),
 	}
 }
 
 func (u *UnionWriter) Write(body zcode.Bytes) error {
-	_, selector, zv := u.typ.SplitZNG(body)
+	if len(body) == 0 {
```
nwt, line 33:
```suggestion
	if body == nil {
```

zio/zeekio/format.go
```diff
@@ -198,8 +198,11 @@ func unescape(r rune) []byte {
 }
 
 func formatUnion(t *zed.TypeUnion, zv zcode.Bytes) string {
-	typ, selector, iv := t.SplitZNG(zv)
-	s := strconv.FormatInt(selector, 10) + ":"
+	if len(zv) == 0 {
```
nwt, line 201:
```suggestion
	if zv == nil {
```

zst/column/union.go
```diff
@@ -24,11 +25,18 @@ func NewUnionWriter(typ *zed.TypeUnion, spiller *Spiller) *UnionWriter {
...
+		u.presence.TouchNull()
+		return nil
+	}
+	u.presence.TouchValue()
+	typ, zv := u.typ.SplitZNG(body)
+	selector := u.typ.Selector(typ)
```
nwt, line 39: I think we should keep the `SplitZNG` selector return parameter so we don't have to do this lookup, but I don't feel strongly about that.

zst/column/union.go
```diff
@@ -103,13 +117,36 @@ func NewUnionReader(utyp zed.Type, in zed.Value, r io.ReaderAt) (*UnionReader, e
 	if err != nil {
 		return nil, err
 	}
+	presence := rec.Deref("presence").MissingAsNull()
+	if presence.IsNull() {
+		return nil, errors.New("ZST union mission presence")
```
nwt, line 122:
```suggestion
		return nil, errors.New("ZST union missing presence")
```

nwt (approved):

PR #3667, opened 2022-02-27
change "in" evaluator to use walk and simplify compiler

mccanne:
This commit changes the "in" logic to walk the input value
 it is now fully complete.  We also removed the network
type as an "in" operand since it was awkward to combine
the semantics of a cidr match with a test for value containment
in a nested value.  To this end, we added a cidr_match() function,
which also walks its value but it doesn't match the network
value straight up.

We also simplified the compiler and removed old vestiges
of filter compilation by removing compileFilter().  The logic
to determine when a constant "in" can be used is simpler
and more general now as we call EvalAtCompileTime instead of
checking for dag.Literal.

We also simplified the logic for regexp searches where a closure
was doing a redundant check with the walk as the walk checks the
top-level value not just values below.


compiler/kernel/expr.go
```diff
@@ -117,6 +118,16 @@ func compileBinary(zctx *zed.Context, e *dag.BinaryExpr) (expr.Evaluator, error)
 	if slice, ok := e.RHS.(*dag.BinaryExpr); ok && slice.Op == ":" {
 		return compileSlice(zctx, e.LHS, slice)
 	}
+	if e.Op == "in" {
+		// Do a faster comparison if the LHS is a compile-time constant expression.
+		in, err := compileConstIn(zctx, e)
+		if in != nil && err == nil {
```
nwt, line 124:
```suggestion
		if in, err := compileConstIn(zctx, e); in != nil && err == nil {
```

runtime/expr/filter.go
```diff
@@ -113,6 +121,31 @@ func (s *search) Eval(_ Context, this *zed.Value) *zed.Value {
...
+
+func (s *searchCIDR) Eval(_ Context, val *zed.Value) *zed.Value {
+	if errMatch == val.Walk(func(typ zed.Type, body zcode.Bytes) error {
+		switch typ.ID() {
+		case zed.IDNet:
+			if bytes.Compare(body, s.bytes) == 0 {
+				return errMatch
+			}
+		case zed.IDIP:
+			addr := zed.DecodeIP(body).IPAddr().IP
```
nwt, line 137: `netaddr.IP.IPAddr` allocates memory. Replacing the `*net.IPNet` with a `netaddr.IPPrefix` would avoid that.

compiler/kernel/expr.go
```diff
@@ -139,8 +152,66 @@ func compileBinary(zctx *zed.Context, e *dag.BinaryExpr) (expr.Evaluator, error)
...
+func compileConstIn(zctx *zed.Context, e *dag.BinaryExpr) (expr.Evaluator, error) {
+	literal, err := EvalAtCompileTime(zctx, e.LHS)
+	if err != nil || literal.IsError() {
+		// If the RHS here is a literal value, it would be good
+		// to optimize this too.  However, this will all be handled
+		// by the JIT compiler that will create optimized closures
+		// on a per-type basis.
+		return nil, nil
+	}
+	eql, err := expr.Comparison("=", literal)
```
nwt, line 168: In light of #3669.
```suggestion
	eql, err := expr.Comparison("==", literal)
```

compiler/kernel/expr.go
```diff
@@ -139,8 +152,66 @@ func compileBinary(zctx *zed.Context, e *dag.BinaryExpr) (expr.Evaluator, error)
...
+	operand, err := compileExpr(zctx, e.RHS)
+	if err != nil {
+		return nil, err
 	}
+	return expr.NewFilter(operand, expr.Contains(eql)), nil
+}
+
+func compileConstCompare(zctx *zed.Context, e *dag.BinaryExpr) (expr.Evaluator, error) {
+	switch e.Op {
+	case "=", "!=", "<", "<=", ">", ">=":
```
nwt, line 181: In light of #3669.
```suggestion
	case "==", "!=", "<", "<=", ">", ">=":
```

nwt (approved):

PR #3670, opened 2022-02-28
differentiate "in" and search with "grep" function

mccanne:
This commit finishes the work to separate containment checking
with "in" from nested string searches with a new function "grep".
It also cleans up the parse tree to fully disambiguate between
a "where operator" that uses a boolean expression and a
"search operator" that uses search syntax.  We do so by adding
AST nodes that are present only in searches.  We updated zfmt
with a check to determine whether an OpExpr is a "where" or a
"search" using these new AST nodes and formatting the canonical
form accordingly.

Since grep subsumes the "matches" operator for most cases, we
removed "matches" but we wanted to still allow regexp compare
in expression context so we revided the old ~ operator to provide
for a regexp comparison where the RHS must be a regexp.

This will all be more thoroughly documented in the forthcoming
zed-language PR.


compiler/parser/parser.peg
```diff
@@ -689,23 +700,34 @@ FuncGuard = NotFuncs __ "("
...
 Function
-  = !FuncGuard fn:IdentifierName __ "(" __ args:OptionalExprs __ ")" where:WhereClause? {
+  = Grep
+  / !FuncGuard fn:IdentifierName __ "(" __ args:OptionalExprs __ ")" where:WhereClause? {
       RETURN(MAP("kind": "Call", "name": fn, "args": args, "where": where))
     }
 
+Grep
+  = "grep" __ "(" __ pattern:Pattern __ ")" {
+      RETURN(MAP("kind":"Grep", "pattern":pattern,"expr":MAP("kind":"ID", "name":"this")))
```
nwt, line 718:
```suggestion
      RETURN(MAP("kind": "Grep", "pattern": pattern, "expr": MAP("kind": "ID", "name": "this")))
```

compiler/parser/parser.peg
```diff
@@ -689,23 +700,34 @@ FuncGuard = NotFuncs __ "("
...
+    }
+  / "grep" __ "(" __ pattern:Pattern __  "," __ expr:Expr __ ")" {
+      RETURN(MAP("kind":"Grep", "pattern":pattern,"expr":expr))
```
nwt, line 721:
```suggestion
      RETURN(MAP("kind": "Grep", "pattern": pattern, "expr": expr))
```

compiler/parser/parser.peg
```diff
@@ -689,23 +700,34 @@ FuncGuard = NotFuncs __ "("
...
+    }
+
+Pattern
+  = Regexp
+  / Glob
+  / s:QuotedString {
+      RETURN(MAP("kind":"String","text":s))
```
nwt, line 728:
```suggestion
      RETURN(MAP("kind": "String", "text": s))
```

docs/zq/operators/search.md
```diff
@@ -0,0 +1,101 @@
...
+### Description
+
+The `search` operator filters its input by applying a search expression `<sexpr>`
+to each input value and dropping each value for which the expression evaluates
+to `false` or to an error.
+
+The "search" keyword may be omitted in which case `<sexpr>` follows
+the [search expression](../language.md#search-expressions) syntax.
+
+When Zed queries are run interactively, it is highly convenient to be able to omit
```
nwt, line 18:
```suggestion
When Zed queries are run interactively, it is convenient to be able to omit
```

nwt (approved):

PR #3677, opened 2022-03-01
Simplify ZSON representation of mixed-type arrays

### LINKED ISSUE #3507: mixed-type arrays in zson
We should simplify ZSON representation of mixed-type arrays.

A mixed-type array implies a array of union so the only time you need to decorate such an array should be when union type includes elements that are not in the array.

This will simplify the ZSON of a typical JSON response in an API where you have an array of objects of varying types.

mattnibs:
Do not decorate the elements of a mixed-type array. Decorate a mixed
type array if when the union type includes elements that are not in
the array.

Closes #3507

### DIFF
--- a/zson/formatter.go
+++ b/zson/formatter.go
@@ -274,27 +274,50 @@ func (f *Formatter) formatRecord(indent int, typ *zed.TypeRecord, bytes zcode.By
 
 func (f *Formatter) formatVector(indent int, open, close string, inner zed.Type, zv zed.Value, known, parentImplied bool) (bool, error) {
 	f.build(open)
-	len, err := zv.ContainerLength()
+	n, err := zv.ContainerLength()
 	if err != nil {
 		return true, err
 	}
-	if len == 0 {
+	if n == 0 {
 		f.build(close)
 		return true, nil
 	}
 	indent += f.tab
 	sep := f.newline
 	it := zv.Iter()
+	union, _ := zed.TypeUnder(inner).(*zed.TypeUnion)
+	seen := make(map[zed.Type]struct{})
 	for !it.Done() {
 		f.build(sep)
 		f.indent(indent, "")
-		if err := f.formatValue(indent, inner, it.Next(), known, parentImplied, true); err != nil {
+		b := it.Next()
+		typ := inner
+		if union != nil {
+			if b == nil {
+				// The type returned from union.SplitZNG for a null value will
+				// be the union type. While this is the correct type, for
+				// display purposes we do not want to see the decorator so just
+				// set the type to null.
+				typ = zed.TypeNull
+			} else {
+				typ, b = union.SplitZNG(b)
+				if _, ok := seen[typ]; !ok {
+					seen[typ] = struct{}{}
+				}
+			}
+		}
+		if err := f.formatValue(indent, typ, b, known, parentImplied, true); err != nil {
 			return true, err
 		}
 		sep = "," + f.newline
 	}
 	f.build(f.newline)
 	f.indent(indent-f.tab, close)
+	if _, isnamed := inner.(*zed.TypeNamed); union != nil && (isnamed || len(seen) != len(union.Types)) {
+		// If we haven't seen all the types in the union, print the decorator
+		// so the fullness of the union is persevered.
+		f.decorate(zv.Type, false, true)
+	}
 	return false, nil
 }
 
--- a/zson/ztests/mixed-array.yaml
+++ b/zson/ztests/mixed-array.yaml
@@ -1,24 +1,17 @@
-script: |
-  zq -f zson -i zson -pretty=0 in.zson
+zed: "*"
 
-inputs:
-  - name: in.zson
-    data: |
-      [1((int64,string)),"b"((int64,string)),2((int64,string))]
-      [8((int64,string)),3((int64,string))]
-      ["hello"((int64,string)),"goodbye"((int64,string))]
-      [null,1,"hello"]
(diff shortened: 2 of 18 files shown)


zson/formatter.go
```diff
@@ -274,27 +274,54 @@ func (f *Formatter) formatRecord(indent int, typ *zed.TypeRecord, bytes zcode.By
...
 	}
-	if len == 0 {
+	if n == 0 {
 		f.build(close)
 		return true, nil
 	}
 	indent += f.tab
 	sep := f.newline
 	it := zv.Iter()
+	union, _ := zed.TypeUnder(inner).(*zed.TypeUnion)
```
mccanne, line 288: This will lose the named type.  I think you want to just not do the TypeUnder...

zson/formatter.go
```diff
@@ -274,27 +274,54 @@ func (f *Formatter) formatRecord(indent int, typ *zed.TypeRecord, bytes zcode.By
...
-		if err := f.formatValue(indent, inner, it.Next(), known, parentImplied, true); err != nil {
+		b := it.Next()
+		typ := inner
+		if union != nil {
+			if b == nil {
+				// The type returned from union.SplitZNG for a null value will
+				// be the union type. While this is the correct type, for
+				// display purposes we do not want to see the decorator so just
+				// set the type to null.
+				typ = zed.TypeNull
```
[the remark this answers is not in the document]
mattnibs: This is already done. The zcode bytes in the above will not be nil (it'll have a selector) so this check will be skipped.

zson/formatter.go
mattnibs: You are right that we're losing the name on named unions. Having said that, I don't think we want the individual elements to get decorated as they would if we removed this on a named union here. Maybe we should decorate the entire array if its inner type is a named union?

```
echo '[1(named=(int64,string)),"foo"(named=(int64,string))]' | go run ./cmd/zq -
```
=>
```
[1,"foo"]([=named])
```
?

mattnibs: (I added a change so the above example now works.

zson/ztests/mixed-array.yaml
```diff
@@ -1,24 +1,17 @@
...
-  zq -f zson -i zson -pretty=0 in.zson
+zed: "*"
 
-inputs:
-  - name: in.zson
-    data: |
-      [1((int64,string)),"b"((int64,string)),2((int64,string))]
-      [8((int64,string)),3((int64,string))]
-      ["hello"((int64,string)),"goodbye"((int64,string))]
-      [null,1,"hello"]
```
nwt, line 10: Why not keep this?

zson/ztests/mixed-array.yaml
```diff
@@ -1,24 +1,17 @@
...
-      {version:[1((int64,string)),"b"((int64,string)),2((int64,string))]}
-      {version:[8((int64,string)),3((int64,string))]}
-      {version:["hello"((int64,string)),"goodbye"((int64,string))]}
+output: |
+  [1,"b",2]
+  [8,3]([(int64,string)])
+  ["hello","goodbye"]([(int64,string)])
+  {version:[1,"b",2]}
+  {version:[8,3]([(int64,string)])}
+  {version:["hello","goodbye"]([(int64,string)])}
```
nwt, line 17: We should be able to do this, right?
```yaml
input: &input |
  [1,"b",2]
  [8,3]([(int64,string)])
  ["hello","goodbye"]([(int64,string)])
  {version:[1,"b",2]}
  {version:[8,3]([(int64,string)])}
  {version:["hello","goodbye"]([(int64,string)])}

output: *input
```

mattnibs: Yeah but I liked how the inputs have explicit decorators on the elements so I can see its resolved correctly.

zson/ztests/mixed-array.yaml
mattnibs: You mean why did I make this a zed test not a script test? I like doing zed tests when I can since they're faster run... is there some direction we're headed w.r.t to when something is a zed test vs. script test?

zson/formatter.go
```diff
@@ -274,27 +274,54 @@ func (f *Formatter) formatRecord(indent int, typ *zed.TypeRecord, bytes zcode.By
...
 	f.build(f.newline)
 	f.indent(indent-f.tab, close)
+	if _, isnamed := inner.(*zed.TypeNamed); union != nil && (isnamed || len(seen) != len(union.Types)) {
+		// If we haven't seen all the types in the union, print the decorator
+		// so the fullness of the union is persevered.
+		f.startColor(color.Gray(200))
+		f.build("(")
+		f.formatType(zv.Type)
+		f.build(")")
+		f.endColor()
```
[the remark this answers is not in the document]
mattnibs: I tried doing that but without that last `true`, maybe that will give us what we want!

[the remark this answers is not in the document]
mattnibs: By job it does work, much better!

zson/ztests/mixed-array.yaml
nwt: No, I'm referring to the `[null,1,"hello"]` case, which has disappeared.

nwt (approved):

zson/ztests/mixed-array.yaml
mattnibs: Oh I removed it because there's a test in mixed-null-array that does exactly the same thing. Honestly those two files should probably just be merged into array.yaml....

PR #3684, opened 2022-03-03
Simply ZSON formatting of mixed-type maps

### LINKED ISSUE #3400: simplify ZSON array of union formatting
For mixed-type arrays, we should simplify ZSON formatting so the implied mixed types imply the type of the array.

e.g., this mixed-type input value implies array of union
```
  {a:[{c:"foo"},{b:1}]}
```
and is formatted as 
```
{a:[{c:"foo"}(({c:string},{b:int64})),{b:1}(({c:string},{b:int64}))]}
```
but she be formatted the same as the input.

mattnibs:
Do not decorate the elements of a mixed-type map. Decorate a mixed
typed map if when the union type includes elements that are not in
the array or if elements are a named union.

Closes #3400

### DIFF
--- a/zson/formatter.go
+++ b/zson/formatter.go
@@ -285,27 +285,11 @@ func (f *Formatter) formatVector(indent int, open, close string, inner zed.Type,
 	indent += f.tab
 	sep := f.newline
 	it := zv.Iter()
-	union, _ := zed.TypeUnder(inner).(*zed.TypeUnion)
-	seen := make(map[zed.Type]struct{})
+	elems := newElemBuilder(inner)
 	for !it.Done() {
 		f.build(sep)
 		f.indent(indent, "")
-		b := it.Next()
-		typ := inner
-		if union != nil {
-			if b == nil {
-				// The type returned from union.SplitZNG for a null value will
-				// be the union type. While this is the correct type, for
-				// display purposes we do not want to see the decorator so just
-				// set the type to null.
-				typ = zed.TypeNull
-			} else {
-				typ, b = union.SplitZNG(b)
-				if _, ok := seen[typ]; !ok {
-					seen[typ] = struct{}{}
-				}
-			}
-		}
+		typ, b := elems.add(it.Next())
 		if err := f.formatValue(indent, typ, b, known, parentImplied, true); err != nil {
 			return true, err
 		}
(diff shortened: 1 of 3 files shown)


zson/formatter.go
```diff
@@ -341,19 +360,22 @@ func (f *Formatter) formatMap(indent int, typ *zed.TypeMap, bytes zcode.Bytes, k
 	f.build("|{")
 	indent += f.tab
 	sep := f.newline
+	keyElems := newElemBuilder(typ.KeyType)
+	valElems := newElemBuilder(typ.ValType)
 	for it := bytes.Iter(); !it.Done(); {
-		keyBytes := it.Next()
+		kbytes := it.Next()
 		if it.Done() {
 			return empty, errors.New("truncated map value")
```
mattnibs, line 368: This check was already there but I'm think this should be a panic, no?

zson/formatter.go
```diff
@@ -285,42 +285,61 @@ func (f *Formatter) formatVector(indent int, open, close string, inner zed.Type,
...
+			return zed.TypeNull, b
+		} else {
+			typ, b := e.union.SplitZNG(b)
+			if _, ok := e.seen[typ]; !ok {
+				e.seen[typ] = struct{}{}
+			}
+			return typ, b
+		}
+	}
+	return e.typ, b
```
nwt, line 335: Indent less!
```suggestion
	if e.union == nil {
		return e.typ, b
	}
	if b == nil {
		// The type returned from union.SplitZNG for a null value will
		// be the union type. While this is the correct type, for
		// display purposes we do not want to see the decorator so just
		// set the type to null.
		return zed.TypeNull, b
	}
	typ, b := e.union.SplitZNG(b)
	if _, ok := e.seen[typ]; !ok {
		e.seen[typ] = struct{}{}
	}
	return typ, b
```

zson/formatter.go
```diff
@@ -341,19 +360,22 @@ func (f *Formatter) formatMap(indent int, typ *zed.TypeMap, bytes zcode.Bytes, k
 	f.build("|{")
 	indent += f.tab
 	sep := f.newline
+	keyElems := newElemBuilder(typ.KeyType)
+	valElems := newElemBuilder(typ.ValType)
 	for it := bytes.Iter(); !it.Done(); {
-		keyBytes := it.Next()
+		kbytes := it.Next()
```
nwt, line 366: Nit: I'd stick with `keyBytes`. It's a better name.

zson/formatter.go
nwt: I'd just leave it.

zson/formatter.go
```diff
@@ -341,19 +360,22 @@ func (f *Formatter) formatMap(indent int, typ *zed.TypeMap, bytes zcode.Bytes, k
...
 		if it.Done() {
 			return empty, errors.New("truncated map value")
 		}
 		empty = false
-		valBytes := it.Next()
 		f.build(sep)
 		f.indent(indent, "")
-		if err := f.formatValue(indent, typ.KeyType, keyBytes, known, parentImplied, true); err != nil {
+		var ktyp zed.Type
+		ktyp, kbytes = keyElems.add(kbytes)
```
nwt, line 374:
```suggestion
		keyType, keyBytes := keyElems.add(keyBytes)
```

zson/formatter.go
```diff
@@ -362,13 +384,17 @@ func (f *Formatter) formatMap(indent int, typ *zed.TypeMap, bytes zcode.Bytes, k
 		if f.tab > 0 {
 			f.build(" ")
 		}
-		if err := f.formatValue(indent, typ.ValType, valBytes, known, parentImplied, true); err != nil {
+		vtyp, vb := valElems.add(it.Next())
+		if err := f.formatValue(indent, vtyp, vb, known, parentImplied, true); err != nil {
```
nwt, line 388:
```suggestion
		valType, valBytes := valElems.add(it.Next())
		if err := f.formatValue(indent, valType, valBytes, known, parentImplied, true); err != nil {
```

zson/formatter.go
```diff
@@ -285,42 +285,61 @@ func (f *Formatter) formatVector(indent int, open, close string, inner zed.Type,
...
-	if _, isnamed := inner.(*zed.TypeNamed); union != nil && (isnamed || len(seen) != len(union.Types)) {
+	if elems.needsDecoration() {
 		// If we haven't seen all the types in the union, print the decorator
 		// so the fullness of the union is persevered.
 		f.decorate(zv.Type, false, true)
 	}
 	return false, nil
 }
 
+type elemBuilder struct {
```
nwt, line 308: I don't know what to call this but it isn't a builder.

nwt (approved):

zson/formatter.go
mattnibs: Truth.

zson/formatter.go
mattnibs: This is true.

zson/formatter.go
mattnibs: Fair.

PR #3693, opened 2022-03-06
add calculator-style usage for zq

mccanne:
This commit changes the zq command parser to accept a single
argument that consists of a yield operator (or implied yield)
followed by optional pipelines operators so that the common
pattern of running 'echo null | zq "..." -' can be replaced simply
with 'zq "..."'.  e.g., zq '1+1' prints out 2.

### DIFF
--- a/cli/zq/command.go
+++ b/cli/zq/command.go
@@ -133,7 +133,7 @@ func (c *Command) Run(args []string) error {
 		// Prevent ParseSourcesAndInputs from treating args[0] as a path.
 		args = append(args, "-")
 	}
-	paths, flowgraph, err := c.queryFlags.ParseSourcesAndInputs(args)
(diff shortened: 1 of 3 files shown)


cli/zq/command.go
```diff
@@ -146,9 +146,14 @@ func (c *Command) Run(args []string) error {
 	}
 	zctx := zed.NewContext()
 	local := storage.NewLocalEngine()
-	readers, err := c.inputFlags.Open(ctx, zctx, local, paths, c.stopErr)
-	if err != nil {
-		return err
+	var readers []zio.Reader
+	if null {
+		readers = []zio.Reader{&singleNull{}}
```
nwt, line 151:
```suggestion
		readers = []zio.Reader{zbuf.NewArray([]zed.Value{*zed.Null})}
```

nwt (approved):

PR #3698, opened 2022-03-07
zfmt: fix format type values

### LINKED ISSUE #3674: type values, type() and zfmt
```
zc -C "is(<foo>) bar"
```
prints 
```
search is(type(foo)) and bar 
```
instead of `<foo>`.

(This is super easy to fix.)

mattnibs:
Closes #3674


zfmt/ztests/typevalues.yaml
```diff
@@ -0,0 +1,12 @@
+script: |
```
nwt, line 1: Nit: Call this `type-value.yaml`.

nwt (approved):

PR #3699, opened 2022-03-07
merge: make argument an expression not field

### LINKED ISSUE #3678: merge operator should take expression not field
The ast and dag nodes for merge require a field path, but the code generated by compiler/kernel creates a compare func that could work with any expression.

e.g., this should work
```
echo '1 2' | zq "fork (=>pass =>pass) | merge this" -
merge: key must be a field
```

mattnibs:
Closes #3678


mccanne (approved):
Nice!

PR #3703, opened 2022-03-08
tidy runtime/expr/shaper.go

nwt:
* Replace copyPrimitive and copyContainer with copyOp.
* Use != consistently in ShaperTransform flag tests.
* Remove isCollectionType and step.append.
* Rename some functions, parameters, and variables.
* Reorder some constants, fields, functions, and parameters.
* Adjust some comments.


mccanne (approved):
Excellent Swiss-French to English translation.

PR #3704, opened 2022-03-08
interpret lone zq arg as query if it begins with file, from, get

nwt:
zq fails with a "no such file" error when run with a lone argument that
is a valid query beginning with file, from, or get.  Make that work.


mccanne (approved):
Me likey.

PR #3707, opened 2022-03-09
support for unary minus

mccanne:


runtime/expr/ztests/unary-minus.yaml
```diff
@@ -0,0 +1,47 @@
+zed: yield -this
+
+input: |
```
nwt, line 3: Add cases for `null`, `null(uint8)`, etc.?

nwt (approved):

runtime/expr/ztests/unary-minus.yaml
mccanne: oopsies

PR #3708, opened 2022-03-09
add docs for zed language

mccanne:


docs/zq/language.md
```diff
@@ -1,341 +1,1970 @@
...
+> is handy for simply emitting new, arbitrary record literals based on
+> input values and mixing in these shaping functions in an embedded record literal.
+> The [fuse aggregate function](aggregates/fuse.md) is useful for fusing
+> values into a common schema within a group-by schema though a type is returned
+> rather than values.
+
+In the examples below, we will use the following named type `connection`
+that is stored in a file [connection.zed](connection.zed)
+and is included in the example Zed queries with `-I` option of `zq`:
+```
```
nwt, line 1693: For example files like this, we should use the content here in the Markdown rather than duplicate it in a separate file.

To do that, change `mdtest-command dir=docs/zq` to just `mdtest-command` everywhere and then add `mdtest-input <filename>` to the fenced code blocks for the example files, like this:
~~~suggestion
```mdtest-input connection.zed
~~~

Comment (mccanne) [reply docs/zq/language.md:1693]: cool!

File: docs/zq/functions/cast.md [hunk 2]
```diff
@@ -0,0 +1,60 @@
...
+```
+<t>(<val>)
+```
+e.g., the result of `cast(1, <string>)` is the same as `string(1)` which is `"1"`.
+
+For complex types, the cast function visits each leaf value in `val` and
+casts that value to the corresponding type in `t`.
+When a complex value has multiple levels of nesting,
+casting is applied recursively down the tree.  For example, cast is recursively
+applied to each element in array of records and recursively applied to each record.
```
nwt, line 25: `cast(val, t)` doesn't walk `val` like this, though I think we do want to add a variant that does.

docs/zq/language.md
mccanne: done

docs/zq/functions/cast.md
mccanne: I will clarify.  It's just referring to the idea that for complex types it casts each component of the complex type.  This came from Henri's description in the issue.

docs/zq/functions/order.md
```diff
@@ -0,0 +1,71 @@
...
+
+_Order a record_
+```mdtest-command
+echo '{b:"foo", a:1}' | zq -z 'order(this, <{a:int64,b:string}>)' -
+```
+produces
+```mdtest-output
+{a:1,b:"foo"}
+```
+TBD: fix this bug or remove example...
```
nwt, line 40: Fixed in #3712.

docs/zq/language.md
```diff
@@ -1,341 +1,1969 @@
...
+landing at the final output stage.
+
+Errors will unfortunately and inevitably occur even in production,
+but having a first-class data type to manage them all while allowing them to
+peacefully co-exist with valid production data is a novel and
+useful approach that Zed enables.
+
+#### 5.3.1 Missing and Quiet
+
+Zed's heterogenous data model allows for queries
```
silveryfu, line 718:
```suggestion
Zed's heterogeneous data model allows for queries
```

docs/zq/language.md
```diff
@@ -1,341 +1,1969 @@
...
+results from accessing a field that is not present.  Thus, `x==NULL` and
+`x==MISSING` could disambiguate the two cases above.
+
+Zed, instead, recognizes that the SQL value is `MISSING` is a paradox:
+I'm here but I'm not.  
+
+In reality, a MISSING value is not a value.  It's an error condition
+that resulted from trying to reference something that didn't exist.
+
+So why should pretend that this is a bona fide value?  SQL adopted this
```
silveryfu, line 752:
```suggestion
So why should we pretend that this is a bona fide value?  SQL adopted this
```

docs/zq/language.md
```diff
@@ -1,341 +1,1969 @@
...
+is presumed to be a key and the corresponding value for that key is
+the result of the operation.  If no such key exists in the map, then
+`error("missing")` results.
+
+If the `<value>` expression is a string, then the `<index>` operand
+must be coercible to an integer and the result is an integer representing
+the unicode code point at that offset in the string.
+
+If the `<value>` expression is type bytes, then the `<index>` operand
+must be coercible to an integer and the result is a unsigned 8-bit integer
```
silveryfu, line 908:
```suggestion
must be coercible to an integer and the result is an unsigned 8-bit integer
```

docs/zq/language.md
```diff
@@ -1,341 +1,1969 @@
...
+to integer and represent a range of index values to form a subset of elements
+from the `<value>` term provided.  The range begins at the `<from>` position
+and ends one before the `<to>` position.  A negative
+value of `<from>` or `<to>` represents a position relative to the
+end of the value being sliced.
+
+If the `<value>` expression is an array, then the result is an array of
+elements comprising the indicated range.
+
+If the `<value>` expression is an set, then the result is a set of
```
silveryfu, line 927:
```suggestion
If the `<value>` expression is a set, then the result is a set of
```

docs/zq/language.md
```diff
@@ -1,341 +1,1969 @@
...
+If the `<value>` expression is type bytes, then the result is a bytes sequence
+consisting of bytes comprising the given range.
+
+### 6.8 Conditional
+
+A conditional expression has the form
+```
+<boolean> ? <expr> : <expr>`
+```
+The `<boolean>` expression is evaluated and must be result it type `bool`.
```
silveryfu, line 942:
```suggestion
The `<boolean>` expression is evaluated and the result must be of type `bool`.
```

docs/zq/language.md
```diff
@@ -1,341 +1,1969 @@
...
+value of `<from>` or `<to>` represents a position relative to the
+end of the value being sliced.
+
+If the `<value>` expression is an array, then the result is an array of
+elements comprising the indicated range.
+
+If the `<value>` expression is an set, then the result is a set of
+elements comprising the indicated range ordered by total order of Zed values.
+
+If the `<value>` expression is an string, then the result is a substring
```
silveryfu, line 930:
```suggestion
If the `<value>` expression is a string, then the result is a substring
```

docs/zq/language.md
```diff
@@ -1,341 +1,1969 @@
...
+```
+
+### 6.10 Aggregate Function Calls
+
+[Aggregate functions](reference.md#aggregates) may be called within an expression.
+Unlike the aggregation context provided by a summarizing group-by, such calls
+in expression context yield an output value for each input value.
+
+Note that because aggregate functions carry state which is typically dependent
+dependent on the order of input values, their use can prevent the runtime
```
silveryfu, line 989:
```suggestion
Note that because aggregate functions carry state which is typically
dependent on the order of input values, their use can prevent the runtime
```

docs/zq/language.md
```diff
@@ -1,341 +1,1969 @@
...
+["hello","world"]
+```
+
+#### 6.11.4 Set Expressions
+
+Set literals have the form
+```
+|[ <expr>, <expr>, ... ]|
+```
+When the expressions results in values of non-uniform type, then the implied
```
silveryfu, line 1104:
```suggestion
When the expressions result in values of non-uniform type, then the implied
```

docs/zq/language.md
```diff
@@ -1,341 +1,1969 @@
...
+A union value can be created with a cast, e.g.,
+```mdtest-command
+echo '1 "foo"' | zq -z 'yield cast(this,<(int64,string)>)' -
+```
+produces
+```mdtest-output
+1((int64,string))
+"foo"((int64,string))
+```
+The value underlying a union-tagged value is accesse with the
```
silveryfu, line 1152:
```suggestion
The value underlying a union-tagged value is accessed with the
```

docs/zq/language.md
```diff
@@ -1,341 +1,1969 @@
...
+
+### 6.13 Type Definitions
+
+Named types may be declared and bound to an identifier with a `type` statement,
+which has the form
+```
+type <name> = <type>
+```
+This statement evaluates the type `<type>` before any input is processed
+to determine the a type value assigned to the identifier `<id>`.
```
silveryfu, line 1193:
```suggestion
to determine the type value assigned to the identifier `<id>`.
```

docs/zq/language.md
```diff
@@ -1,341 +1,1969 @@
...
-To build effective queries, it is also important to become familiar with the
-Zed _[Data Types](#data-types)_.
+##### 7.2.1.6 Predicate Search Term
 
-Each of the sections hyperlinked above describes these elements of the language
-in more detail. To make effective use of the materials, it is recommended to
-first review the XXX.  Just put conventions here.  Coming in subsequent PR.
+Any Boolean-valued [function](reference.md) like `is()`, `has()`,
+`grep()` etc. and any [comparison expression](#comparisons)
+may be used a search term and mixed into a search expression.
```
silveryfu, line 1553:
```suggestion
may be used as a search term and mixed into a search expression.
```

docs/zq/language.md
```diff
@@ -1,341 +1,1969 @@
...
+```
+where `<elem>` has either an assignment form
+```
+<var>=<expr>
+```
+or a field reference form
+```
+<field>
+```
+For each input value to the outer scope, the assignment form create a binding
```
silveryfu, line 1641:
```suggestion
For each input value to the outer scope, the assignment form creates a binding
```

docs/zq/language.md
```diff
@@ -1,341 +1,1969 @@
...
+
+The first technique is to simply combine types with a type union.
+For example, an `int64` and a `string` can be merged into a common
+type of union `(int64,string)`, e.g., the value sequence `1 "foo"`
+can be fused into the single-type sequence:
+```
+1((int64,string))
+"foo"((int64,string))
+```
+The second technique is to merge fields of records, analagous to a spread
```
silveryfu, line 1878:
```suggestion
The second technique is to merge fields of records, analogous to a spread
```

docs/zq/language.md
```diff
@@ -1,341 +1,1969 @@
...
+
+To perform fusion, Zed currently includes two key mechanisms
+(though this is an active area of development):
+* the [fuse operator](operators/fuse.md#operator), and
+* the [fuse aggregate function](aggregates/fuse.md).
+
+### 10.1 Fuse Operator
+
+The _fuse operator_ reads all of its input, computes a fused type using
+the techniuqes above, and outputs the result, e.g.,
```
silveryfu, line 1902:
```suggestion
the techniques above, and outputs the result, e.g.,
```

docs/zq/functions/cast.md
```diff
@@ -0,0 +1,60 @@
...
+
+### Description
+
+The _cast_ function performs type casts but handles both primitive types and
+complex types.  If the input type `t` is a primitive type, then the result
+is equivalent to
+```
+<t>(<val>)
+```
+e.g., the result of `cast(1, <string>)` is the same as `string(1)` which is `"1"`.
```
mattnibs, line 19: It's also equivalent to <string>(1) (you can also use complex types for this syntax).

docs/zq/language.md
mccanne: must have a result of type `bool`.

docs/zq/functions/cast.md
mccanne: This is explained in language.md and I don't feel like it needs to be included here too as most the of the time you wouldn't type the angle brackets for primitive types.

mccanne: I updated language.md with a bit more clarity on this...

docs/zq/language.md
```diff
@@ -1,341 +1,2091 @@
...
+you can specify an operator including keyword search terms, Boolean predicates,
+etc. using the same syntax at any point in the pipeline as
+[described below](#7-search-expressions)
 
-## Record Literal
+For example,
+the predicate `message_length > 100` can simply be tacked onto the keyword search
+from above, e.g.,
+```
+sample.com urgent message_length > 100
```
silveryfu, line 133:
```suggestion
example.com urgent message_length > 100
```
or,
```suggestion
urgent sample.com message_length > 100
```
or,
```suggestion
'sample.com' urgent message_length > 100
```
(`sample` is reserved for the zed operator - renaming to `example.com` might make most sense?)

docs/zq/language.md
```diff
@@ -1,341 +1,2091 @@
...
+which computes an aggregation table of different message types (e.g.,
+from a hypothetical field called `type`) into a new, aggregated field
+called `kinds` and grouped by by the network of all the source IP address
+in the input
+(e.g.,from a hypothetical field called `srcip`) as a derived field called `net`.
 
-TBD: this is linked from below
+The short-hand query from above might be typed into a search box while the
+latter query might be composed in a query editor or in Zed source files
+maintained in Github.  Both forms are valid Zed queries.
```
silveryfu, line 152:
```suggestion
maintained using git ("gitops for data").  Both forms are valid Zed queries.
```

mccanne: I'm not sure Noah would go for this :-)

docs/zq/operators/switch.md
```diff
@@ -0,0 +1,70 @@
...
+Only one default case is allowed and it may appear anywhere in the list of cases;
+where it appears does not influence the result.
+
+The output of a switch consists of multiple legs that must be merged.
+If the downstream operator expects a single input, then the output legs are
+merged with an automatically inserted [combine operator](combine.md).
+
+### Examples
+
+_Split input into even and odds_
```
nwt, line 49:
```suggestion
_Split input into evens and odds_
```

docs/zq/operators/switch.md
```diff
@@ -0,0 +1,70 @@
...
+echo '1 2 3 4' | zq -z 'switch ( case this%2==0 => {even:this} case this%2==1 => {odd:this}) | sort odd,even' -
+```
+=>
+```mdtest-output
+{odd:1}
+{odd:3}
+{even:2}
+{even:4}
+```
+_Switch on this with a constant case_
```
nwt, line 60:
```suggestion
_Switch on `this` with a constant case_
```

docs/zq/language.md
nwt: I think either "in GitHub" or "in Git" is fine, though my heart lies with the more generic "in a version control system".

docs/zq/language.md
```diff
@@ -1,341 +1,2091 @@
...
+
+
+### 2.6 Implied Operators
+
+When Zed is run in an application like [Brim](https://github.com/brimdata/brim),
+queries are often composed interactively in a "search bar" experience.
+The language design here attempts to support both this "lean forward" pattern of usage
+along with a "coding style" of query writing where the queries might be large
+and complex, e.g., to perform transformations in a data pipeline, where
+the Zed queries are stored under source-code control perhaps in Github or
```
nwt, line 358:
```suggestion
the Zed queries are stored under source-code control perhaps in GitHub or
```

docs/zq/language.md
```diff
@@ -1,341 +1,2091 @@
...
+input record field is modified, as compared to `=` which is used in constant
+and variable assignments.  For example, the operation
+```
+put y:=2*x+1
+```
+can be expressed simply as
+```
+y:=2*x+1
+```
+When composing long-form queries that are shared via Brim or managed in Github,
```
nwt, line 413:
```suggestion
When composing long-form queries that are shared via Brim or managed in GitHub,
```

docs/zq/language.md
```diff
@@ -1,341 +1,2091 @@
...
 ```
-from (
-  pool PoolOne => op1 | op2 | ...
-  pool PoolTwo => op1 | op2 | ...
-) | join on key=key | ...
+const <id> = <expr>
 ```
-Similarly, data can be routed to different paths with replication
-using `switch`:
+where `<id>` is an identifier and `<expr>` or a constant [expression](#6-expressions)
```
nwt, line 450:
```suggestion
where `<id>` is an identifier and `<expr>` is a constant [expression](#6-expressions)
```

docs/zq/language.md
```diff
@@ -1,341 +1,2091 @@
...
+Const statements may appear intermixed with type statements.
+
+## 4. Type Statements
+
+Named types may be created with the syntax
+```
+type <id> = <type>
+```
+where `<id>` is an identifier and `<type>` or a [Zed type](#first-class-types).
+This create a new type with the given name in the Zed test system, e.g.,
```
nwt, line 480:
```suggestion
This create a new type with the given name in the Zed type system, e.g.,
```

docs/zq/language.md
```diff
@@ -1,341 +1,2091 @@
...
+## 3. Const Statements
 
-Or multiple pools can be accessed and, for example, joined:
+Constants may be defined and assigned to a symbolic name with the syntax
 ```
-from (
-  pool PoolOne => op1 | op2 | ...
-  pool PoolTwo => op1 | op2 | ...
-) | join on key=key | ...
+const <id> = <expr>
```
nwt, line 448: To avoid confusion with Zed type values, I think we should adopt a convention other than angle brackets for syntax examples like this. Caps or italics, maybe?

docs/zq/language.md
```diff
@@ -1,341 +1,2091 @@
...
+* a union of string and integer  - `(string,int64)`
+```
+Complex types may be composed, as in `[({s:string},{x:int64})]` which is
+an array of type union of two types of records.
+
+The [typeof function](functions/typeof.md) returns a value's type as
+a value, e.g., `typeof(1)` is `<int64>` and `typeof(<int64>)` is `<type>`.
+
+First-class types are quite powerful because types can
+serve as group-by keys or used in "data shaping" logic.
```
nwt, line 550:
```suggestion
serve as group-by keys or be used in "data shaping" logic.
```

docs/zq/language.md
```diff
@@ -1,341 +1,2091 @@
...
+where `<name>` is typically an identifier (though a type name can have any
+characters in it so `<name>` can also be a quoted string), and `<type>` is
+any Zed type.  For example,
+```
+type port=uint16
+```
+defines a type named "port" for type `uint16`.
+
+Type statements (along with const statements) must appear at the beginning
+of a query or the beginning of a [parenthesized scope](#scopes).
```
nwt, line 600: This section seems to repeat material in section 4.

docs/zq/language.md
```diff
@@ -1,341 +1,2091 @@
...
+results in
+```mdtest-output
+{typeof:<bar=int64>,count:1(uint64)}
+{typeof:<foo=int64>,count:2(uint64)}
+{typeof:<foo=string>,count:1(uint64)}
+```
+Here, the two versions of type "foo" are retained in the group-by results.
+
+In general, it is bad practice to define multiple versions a single named type
+though the Zed system and Zed data model accommodates such dynamic bindings.
```
nwt, line 663:
```suggestion
though the Zed system and Zed data model accommodate such dynamic bindings.
```

docs/zq/language.md
```diff
@@ -1,341 +1,2091 @@
...
+In general, expressions and functions that result in errors simply return
+a value of type error as a result.  This encourages a powerful flow-style
+of error handling where errors simply propagate from one operation to the
+next and land in the output alongside non-error values to provide a very helpful
+context and rich information for tracking down the source of errors.  There is
+no need to check for error conditions everywhere or look through auxiliary
+logs to find out what happened.
+
+For example,
+the following values can be transformed to errors as follows:
```
nwt, line 713:
```suggestion
input values can be transformed to errors as follows:
```

docs/zq/language.md
```diff
@@ -1,341 +1,2091 @@
...
+<error(string)>
+```
+First-class errors are particularly useful for creating structured errors.
+When a Zed query encounters a problematic condition,
+instead of silently dropping the problematic error
+and logging an error obscurely into some hard-to-find system log as so many
+ETL pipelines do, the Zed logic can
+preferably wrap the offending value as an error and propagate it to its output.
+
+For example, suppose a bad value shows up...
```
nwt, line 749:
```suggestion
For example, suppose a bad value shows up:
```

docs/zq/language.md
```diff
@@ -1,341 +1,2091 @@
...
+```
+{kind:"bad", stuff:{foo:1,bar:2}}
+```
+a Zed shaper could catch the bad value (e.g., as a default case in a
```
nwt, line 753:
```suggestion
A Zed shaper could catch the bad value (e.g., as a default case in a
```

docs/zq/language.md
```diff
@@ -1,341 +1,2091 @@
...
+switch topology) and propagate it as an error using the Zed expression:
+```
+yield error({message:"unrecognized input",input:this})
+```
+then such errors could be detected and searched for downstream with the
```
nwt, line 758:
```suggestion
Then such errors could be detected and searched for downstream with the
```

docs/zq/language.md
```diff
@@ -1,341 +1,2091 @@
...
+to land such errors as they can simply land next to the output values themselves.
+
+And when transformations cascade one into the next as different stages of
+an ETL pipeline, errors can be wrapped one by one forming a "stack trace"
+or lineage of where the error started and what stages it traversed before
+landing at the final output stage.
+
+Errors will unfortunately and inevitably occur even in production,
+but having a first-class data type to manage them all while allowing them to
+peacefully co-exist with valid production data is a novel and
```
nwt, line 778:
```suggestion
peacefully coexist with valid production data is a novel and
```

docs/zq/language.md
```diff
@@ -1,341 +1,2091 @@
...
+echo "{x:1} {y:2}" | zq -z 'yield x' -
+```
+produces
+```mdtest-output
+1
+error("missing")
+```
+Sometimes you want missing errors to show up and sometimes you don't.
+The [quiet function](functions/quiet.md) transforms missing errors into
+"quiet errors".  A quiet error is value `error("quiet")` and is ignored
```
nwt, line 833:
```suggestion
"quiet errors".  A quiet error is the value `error("quiet")` and is ignored
```

docs/zq/language.md
```diff
@@ -1,341 +1,2091 @@
...
+produces
+```mdtest-output
+{a:[1,2]}
+{d:{e:1}}
+```
+
+### 6.4 Logic
+
+The keywords `and`, `or`, and `not` perform logic on operands of type `bool`.
+The binary operands of `and` and `or` operate on Boolean values and result in
```
nwt, line 922:
```suggestion
The binary operators `and` and `or` operate on Boolean values and result in
```

docs/zq/language.md
```diff
@@ -1,341 +1,2091 @@
...
+an error value if either operand is not a boolean.  Likewise, `not` operates
```
nwt, line 923:
```suggestion
an error value if either operand is not a Boolean.  Likewise, `not` operates
```

docs/zq/language.md
```diff
@@ -1,341 +1,2091 @@
...
+### 6.5 Field Dereference
+
+Records fields are dereferenced with the dot operator `.` as is customary
+in other languages and have the form
+```
+<value> . <id>
+```
+where `<id>` is an identifier representing the field name referenced.
+If a field name is not representable as an identifier, then [indexing](#indexing)
+may be used with a quoted to string to represent any valid field name.
```
nwt, line 937:
```suggestion
may be used with a quoted string to represent any valid field name.
```

docs/zq/language.md
mccanne: I agree but this is used everywhere.  I created #3270

docs/zq/language.md
```diff
@@ -1,341 +1,2091 @@
...
+```mdtest-command
+echo '{diameter:1}{diameter:5}' | zq -z 'const PI = 3.14159 circumference:=2*PI*diameter' -
+```
+produces
+```mdtest-output
+{diameter:1,circumference:6.28318}
+{diameter:5,circumference:31.4159}
+```
+
+### 6.13 Type Definitions
```
nwt, line 1266: These sections largely duplicate sections 3 and 4, but maybe that's OK.

docs/zq/language.md
```diff
@@ -1,341 +1,2091 @@
...
+true
+true
+false
+```
+
+#### 7.1.2 Globs
+
+Globs provide a convenient short-hand for regular expressions and follow
+the familiar pattern of "file globbing" supported by Unix shells.
+Zed globs are a simple, special case utilize only the `*` wildcard.
```
nwt, line 1425:
```suggestion
Zed globs are a simple, special case utilizing only the `*` wildcard.
```

docs/zq/language.md
```diff
@@ -1,341 +1,2091 @@
...
+is a glob match for any matching string value in the input, but
+```
+a*b==c
+```
+is a Boolean comparison between the product `a*b` and `c`.
+
+### 7.2 Search Logic
+
+The search patterns described above can be combined with other elements
+to a search expression comprised of "search terms" that may be combined
```
nwt, line 1480: /to a/to form a/, maybe?

docs/zq/language.md
```diff
@@ -1,341 +1,2091 @@
...
+
+> Note that when processing ZNG data, the Zed runtime performs a multi-threaded
+> Boyer-Moore scan over decompressed data buffers before parsing any data.
+> This allows large buffers of data to be efficiently discarded and skipped when
+> searching for rarely occurring values.  For a [Zed lake](../zed), search indexes
+> may also be configured to further accelerate searches.
+> In a forthcoming release, Zed will also offer an approach for locating
+> delimited words within string fields, which will allow accelerated
+> search using a full-text search index.  Currently, search indexes may be built
+> for exact value match as text segmentation is in the works.a
```
nwt, line 1491:
```suggestion
> for exact value match as text segmentation is in the works.
```

docs/zq/language.md
mccanne: Indeed, will rework.

docs/zq/language.md
```diff
@@ -1,341 +1,2091 @@
...
+> Also, the [yield operator](operators/yield.md#operator)
+> is handy for simply emitting new, arbitrary record literals based on
+> input values and mixing in these shaping functions in an embedded record literal.
+> The [fuse aggregate function](aggregates/fuse.md) is useful for fusing
+> values into a common schema within a group-by schema though a type is returned
+> rather than values.
+
+In the examples below, we will use the following named type `connection`
+that is stored in a file `connection.zed`
+and is included in the example Zed queries with `-I` option of `zq`:
```
nwt, line 1813:
```suggestion
and is included in the example Zed queries with the `-I` option of `zq`:
```

PR #3713, opened 2022-03-11
extend cast to shape to a named type when given a string

mccanne:


mccanne:
I'll rebase zed-language to this addition and update docs in that branch.

nwt (approved):

PR #3714, opened 2022-03-11
parser: support IEEE Inf and NaN

### LINKED ISSUE #3702: add NaN and Inf to Zed language
IEEE NaN and Inf should be added to the Zed PEG parser and made part of the language.

This is an easy change because they should need to be accepted as float literals and the existing Go code should handle them without change since strconv etc knows what do to do with these.

mattnibs:
Closes #3702


mccanne:
Looks good to me.  I'll let @nwt approve in case he thinks of a corner case.

nwt (approved):
Looks good

PR #3724, opened 2022-03-15
has_error function

### LINKED ISSUE #3719: has_error()
We should have a has_error() function that returns true for any value that has an error somewhere inside of it.  Since errors must appear in the type signature, this could cache results in a type map.

mattnibs:
Add has_error function which will recurse a zed value and return true if
the value or any of its leafs an error. has_error memoizes the types it
has seen so it won't have to recurse types more than once (unless the it
contains an error type in a union in which case it'll always have to
recurse).

Closes #3719

### DIFF
--- a/runtime/expr/function/types.go
+++ b/runtime/expr/function/types.go
@@ -131,6 +131,10 @@ func (h *HasError) hasError(t zed.Type, b zcode.Bytes) (bool, bool) {
 	if _, ok := typ.(*zed.TypeError); ok {
 		return true, false
 	}
+	// If a value is null we can skip since an null error is not an error.
+	if b == nil {
+		return false, false
+	}
 	if hasErr, ok := h.cached[t.ID()]; ok {
 		return hasErr, true
 	}
--- a/runtime/expr/function/types.go
+++ b/runtime/expr/function/types.go
@@ -2,6 +2,7 @@ package function
 
 import (
 	"github.com/brimdata/zed"
+	"github.com/brimdata/zed/zcode"
 	"github.com/brimdata/zed/zson"
 )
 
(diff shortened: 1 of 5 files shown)


mattnibs:
Also of note, has_error doesn't support type values, should it?

runtime/expr/function/types.go
```diff
@@ -107,6 +108,78 @@ func (i *Is) Call(_ zed.Allocator, args []zed.Value) *zed.Value {
...
+	var hasErr bool
+	canCache := true
+	switch typ := typ.(type) {
+	case *zed.TypeRecord:
+		it := b.Iter()
+		for _, col := range typ.Columns {
+			e, c := h.hasError(col.Type, it.Next())
+			hasErr = hasErr || e
+			canCache = !canCache || c
+		}
```
nwt, line 146: This will panic on a null record value.

runtime/expr/function/types.go
```diff
@@ -107,6 +108,78 @@ func (i *Is) Call(_ zed.Allocator, args []zed.Value) *zed.Value {
...
+
+func (h *HasError) Call(_ zed.Allocator, args []zed.Value) *zed.Value {
+	v := args[0]
+	if yes, _ := h.hasError(v.Type, v.Bytes); yes {
+		return zed.True
+	}
+	return zed.False
+}
+
+func (h *HasError) hasError(t zed.Type, b zcode.Bytes) (bool, bool) {
```
nwt, line 129: I think you could simplify this if you used `Value.Walk`.

runtime/expr/function/types.go
nwt: This still panics on `null({a:int64})`.

nwt: This still panics on `null({a:int64})`.

runtime/expr/function/types.go
mattnibs: I had considered this but there was some case that I thought had made it unworkable. Let me look again.

mattnibs: Using walk is tougher because I need to keep track of whether a part of value tree is cacheable or not.

nwt (approved):

PR #3729, opened 2022-03-17
fix premature EOS in scoped over operator

nwt:
The flowgraph for a scoped over operator (i.e., "over ... => (...)")
shuts down prematurely if the scoped flowgraph produces no output for a
given input, as with a where operator.  Fix by adding a channel in
runtime/op/traverse.Scope called parentEOSCh, sending to it when a
Scope's parent returns an EOS, trying to receive from it when
Exit.pullPlatoon pulls an empty platoon, and interpreting the empty
platoon as EOS only if the receive succeeds.

This seems to work but doesn't feel particularly robust.  I think we
might need a different design here if we want something that's more
obviously correct.

### DIFF
--- a/runtime/op/traverse/scope.go
+++ b/runtime/op/traverse/scope.go
@@ -11,25 +11,29 @@ import (
 )
 
 type Scope struct {
-	ctx        context.Context
-	parent     zbuf.Puller
-	enter      *Enter
-	subgraph   zbuf.Puller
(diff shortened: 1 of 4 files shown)


mccanne (approved):
Yeah, this stuff is hard and I iterated on the design a few times before finally getting something to work but I agree, there must be better way.

Issue #3730, opened 2022-03-18
Failure of "go install" command line in top-level README


philrz:
I attempted to run the `go install` command line shown in the top-level README as of Zed commit f24ab19adfa951aed96733fd0fbe97fa4ec58b79 and it failed.

```
$ go version
go version go1.17.8 darwin/amd64

$ go install github.com/brimdata/zed/cmd/{zed,zq}@latest
go install: github.com/brimdata/zed/cmd/zed@latest (in github.com/brimdata/zed@v0.33.0):
	The go.mod file for the module providing named packages contains one or
	more replace directives. It must not contain directives that would cause
	it to be interpreted differently than if it were the main module.
```

philrz:
As of Zed GA tagged `v1.0.0`, this no longer fails. However, a different issue is now revealed, which is tracked in #3777.

PR #3731, opened 2022-03-18
Add a few hyperlinks in top-level Zed README

philrz:
While reviewing the current top-level Zed README as if I were a new user, I thought of a few hyperlinks it might be helpful to add. Most of these tools are surely familiar to the user base, but I figure it never hurts to be inclusive to novices.

As noted to the Dev team on Slack, I also noticed in passing that the `go install` and `brew` command lines aren't currently working, but I suspect this is an expected side effect of there not being a tagged release yet based on the work described in #3619. Before I pieced together this theory about the root cause, I opened #3730 about `go install` but held off on opening one about `brew` for now. If I'm right about any of this and someone is maintaining a checklist of things to check when the next release is tagged, these would be good things to test for in final steps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,8 @@ And the same search/analytics system that works at cloud scale should run easily
 a lightweight command-line tool on your laptop.
 
 And rather than having to set up complex ETL pipelines with brittle
-transformation logic, managing your data lake should be as easy as `git`.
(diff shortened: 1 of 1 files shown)


mccanne (approved):
Thanks!

PR #3732, opened 2022-03-19
Fix example and a few typos

philrz:
I noticed that the `cut shape` in the example didn't match the spirit of the preceding SQL query, so I've adjusted it here. I found a couple other typos to fix while I was in there.

### DIFF
--- a/docs/formats/README.md
+++ b/docs/formats/README.md
@@ -211,8 +211,8 @@ for schema interpretation and management.
 ### 2.2 Type Combinatorics
 
 A common objection to using a type system to represent schemas is that
-xdiverse applications generating arbitrarily structured data can produce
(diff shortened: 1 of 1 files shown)


docs/formats/README.md
```diff
@@ -266,7 +265,7 @@ and more natural given the super-structure of Zed.  For example, the above
 two SQL-like queries could be written as:
 ```
   count() by shape:=typeof(this)
-  any(this) by shape:=typeof(this) | cut shape
+  any(this) by shape:=typeof(this) | cut any
```
mccanne, line 268: Oops... let's get rid of `shape:=` since we are cutting any...

philrz: Good idea. I pushed a commit to make that adjustment.

PR #3735, opened 2022-03-22
add zed command README

mccanne:
Ths doc adds the zed command README and factors out the design
discussion as design notes in an unlinked markdown doc design.md.

This needs quite a bit more work to add detail, document all the
options, and make the presentation more uniform but we will get to
this in our next big chunk of work on the lake.

PR #3744, opened 2022-03-23
make zed use command and -use flag more pool-centric

### LINKED ISSUE #3243: make -use flag more pool-centric
Zed lake commands interpret the `-use` flag's value as a branch or commit if it does not contain the `@` character. We should change that interpretation from branch or commit to pool to shorten the common case of specifying a pool's `main` branch from `-use $pool@main` to `-use $pool`.

nwt:
The zed use command and -use flag interpret their argument as a branch
or commit if it does not contain the @ character.  Change them to
interpret it as a pool with implied branch main, and remove the now
redundant zed use -p flag and the confusing -b flag.

Closes #3243.

@mccanne: If this looks good, I'll suggest necessary updates on #3735.


lakeparse/head.go
```diff
@@ -12,13 +12,16 @@ type Commitish struct {
 }
 
 func ParseCommitish(commitish string) (*Commitish, error) {
+	if commitish == "" {
+		return nil, errors.New("empty pool and branch")
```
mattnibs, line 16: Should this "branch" be "ref" since this could be a branch or commit sha? I do think "branch" is clearer though...

mccanne (approved):
Hurrah!

lakeparse/head.go
nwt: Excellent question! I think we do want to change our terminology here, though "ref" doesn't feel quite right because in Git, a commit ID isn't a ref. What I'd like to do is commit this as is and improve it later.

PR #3749, opened 2022-03-24
A few more fixes in zed command README

philrz:
With #3735 now merged, I gave the `zed` command README a final run through and found a couple broken links and other final fixes.

### DIFF
--- a/docs/zed/README.md
+++ b/docs/zed/README.md
@@ -4,7 +4,7 @@
 
 > `zed` is a command-line tool to manage and query Zed data lakes.
 > You can import data from a variety of formats and `zed` will automatically
-> commit the data in the Zed data model's [super-structured](../formats/zed.md)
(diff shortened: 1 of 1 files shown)


docs/zed/README.md
```diff
@@ -4,7 +4,7 @@
 
 > `zed` is a command-line tool to manage and query Zed data lakes.
 > You can import data from a variety of formats and `zed` will automatically
-> commit the data in the Zed data model's [super-structured](../formats/zed.md)
+> commit the data in the Zed data model's [super-structured](../formats/README.md)
```
philrz, line 7: I suggested linking to this instead because the `zed.md` actually doesn't speak of "super-structured" at all, so it seems it would have had the potential to confuse.

nwt (approved):

PR #3752, opened 2022-03-24
add -j flag to zed and zq as a shortcut for -f json

nwt:
Our plan was to make `-j` a shortcut for `-i json -f json`, but paring it back to just `-f json` made both the implementation (given the separation between `cli/inputflags` and `cli/outputflags`) and documentation easier. But if anyone feels strongly about it, I can add `-i json` either here or in another PR.


mccanne (approved):
No worries about the input side if and until people ask.

PR #3754, opened 2022-03-25
Cast template expressions as strings

mattnibs:


mattnibs:
I'm not sure if this the correct fix for the `'${1}'` problem. Perhaps the ast should be of template StringExpr with a list of expressions and the semantic compiler adds the cast. Either way this fixes the bug for now.

compiler/parser/parser.peg
```diff
@@ -1081,7 +1081,16 @@ TemplateSingleQuotedChar
...
-  = "${" __ e:Expr __ "}" { RETURN(e) }
+  = "${" __ e:Expr __ "}" { 
+      RETURN(MAP(
+        "kind": "Cast",
+        "expr": e,
+        "type": MAP(
+          "kind": "TypeValue",
+          "value": MAP("kind": "TypePrimitive", "name": "string"),
+        ),
+      ))
```
nwt, line 1092: We don't put closing parens on separate lines anywhere else in here.
```suggestion
          "value": MAP("kind": "TypePrimitive", "name": "string"))))
```

nwt (approved):

PR #3761, opened 2022-03-25
A few more README fixes

philrz:
I missed the boat on the #3757 review, so here's a few more suggestions. I'll add some in-line comments in places where I'd appreciate extra attention.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,11 +79,11 @@ is to download a pre-built release binary.
 You can find these binaries on the GitHub
 [releases](https://github.com/brimdata/zed/releases) page.
 
-On macOS and Linux, you can also use [Homebrew](https://brew.sh/) to install `zq`:
+On macOS and Linux, you can also use [Homebrew](https://brew.sh/) to install `zq`, run:
 ```
 brew install brimdata/tap/zq
 ```
-To install `zed` for working with lakes, use
+To install `zed` for working with lakes, run
 ```
 brew install brimdata/tap/zed
 ```
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ proper superset of JSON, `zq` also works natively with JSON.
 
 While `zq` and the Zed data formats are production quality, the Zed project's
 [Zed data lake](docs/zed/README.md) is a bit [earlier in development](docs/zed/README.md#status).
-The Zed lake will look somewhat like a lakehouse but will utilize the
+The Zed lake will look somewhat like a [lakehouse](https://databricks.com/blog/2020/01/30/what-is-a-data-lakehouse.html) but will utilize the
 Zed type system to organize its underlying data instead of often hard-to-manage
 relational tables and schemas.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -83,7 +83,7 @@ On macOS and Linux, you can also use [Homebrew](https://brew.sh/) to install `zq
 ```
 brew install brimdata/tap/zq
 ```
-To install `zed`, use
+To install `zed` for working with lakes, use
```
philrz, line 86: Maybe these extra words are overkill. But I suggested adding them because the above text in this README does a nice job of setting expectations that `zq` is a standalone command-line tool because it's compared to `jq`. However, it's not as clearly spelled out that `zed` the command line tool is the one for lakes. But since the word "Zed" appears so many times already in this doc, I figure a little extra disambiguation here can't hurt. It's admittedly all made even more clear in the lines that follow.

README.md
```diff
@@ -164,7 +164,7 @@ See the [contributing guide](CONTRIBUTING.md) on how you can help improve Zed!
 
 ## Join the Community
 
-Join our [Public Slack](https://www.brimdata.io/join-slack/) workspace for announcements, Q&A, and to trade tips!
+Join our [public Slack](https://www.brimdata.io/join-slack/) workspace for announcements, Q&A, and to trade tips!
```
philrz, line 167: I think I was the one that incorrectly capitalized "Public" back in the day, so I'm just fixing my own error here. :wink:

docs/README.md
```diff
@@ -72,38 +72,38 @@ built from self-describing data structures.  Everything in a Zed lake
...
-Once copied, a new service can be instantiated by pointing a `zed service`
+Once copied, a new service can be instantiated by pointing a `zed serve`
 at the copy of the lake.
 
 Functionality like indexing, data compaction, and retention are all
 API-driven.
 
 Bite-sized components are unified by the Zed data, usually in the ZNG format:
-* All lake meta-data is available via metaqueries.
+* All lake meta-data is available via meta-queries.
```
philrz, line 95: This appears hyphenated at a couple other docs already in the repo.

docs/README.md
```diff
@@ -72,38 +72,38 @@ built from self-describing data structures.  Everything in a Zed lake
...
 * All like operations available through the service API are also available
 directly via the `zed` command.
-* Everything available to the client is
+* Everything available to the client is XXX
```
philrz, line 98: I noticed this sentence ends unfinished. I'm not sure what was intended, so maybe someone can step in and finish it. Otherwise the bullet should probably be removed for now.

README.md
mccanne: I also think "use" should be "run" here

docs/README.md
mccanne: oops, I think this was a point about everything the Brim app has access to is available in the API, but I think since this in the zed repo, let's just delete this bullet.

mccanne:
Thanks, Phil!

philrz:
@mccanne: I pushed commits for the use->run change and removing the XXX bullet, so it's ready for approve+merge if someone on your end is up for it.

mccanne (approved):

PR #3768, opened 2022-03-26
fix some issues in .goreleaser.yaml

nwt:
* Use .Tag instead of .Version in templates
* Use relative path for CHANGELOG.md
* Set ldflags for zq
* Move zq Homebrew formula to top level to match zed formula
* Install one binary per Homebrew formula

(This is the `.goreleaser.yaml` that I used for https://github.com/brimdata/zed/releases/tag/v1.0.0.)

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -5,7 +5,7 @@ builds:
     env:
       - CGO_ENABLED=0
     ldflags:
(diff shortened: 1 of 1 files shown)


mattnibs (approved):

nwt:
> Should we also use the BRIMBOT's personal access token so we fix the issue with it being able to push the brew changes?

Let's do that separately since it happens in `.github/workflows/release.yml`.

PR #3753, opened 2022-03-25
add zq primer to docs/tutorials

mccanne:

PR #3769, opened 2022-03-26
zq primer fixes

philrz:
I did another proofreading pass on the `zq` primer after the merge #3753 and found some more fixes to propose. I'll include some inline comments among the diffs where I'd like to draw some extra attention.

In addition to what's captured in the diffs, I did bump into one higher-level snag with the flow. As I was pretending to be a new user, I obediently executed the command shown:

```
zq -f json https://api.github.com/repos/brimdata/zed/pulls\?state\=all\&sort\=desc\&per_page=30 > prs.json
```

...and started working off my own JSON of the old PRs rather than use the one checked into the repo. However, the results returned by the GitHub API didn't have the two empty object `{}` entries referenced in the doc, so when the example commands get to counting and avoiding the empties, my results started to diverge from what's shown in the doc. It wasn't too difficult for me to switch to the one in the repo and start to see the expected results. However, I could imagine a new user being more slowed down at this spot and thinking they maybe did something wrong.

I'm not sure if the empty object `{}` entries were something put there intentionally in the version in the repo or if if they're legit and there's some adjustment to the GitHub API call that could ensure that the user will always get the same result as what's in the repo. If they're an intentional fudge just for the repo version, then maybe the doc could just be enhanced to emphasize that the user needs to stick to the one in the repo.

### DIFF
--- a/docs/tutorials/zq-primer.md
+++ b/docs/tutorials/zq-primer.md
@@ -17,7 +17,7 @@ pulled from the GitHub API.
 
 If you want to follow along on the command line,
 just make sure the `zq` command [is installed](../../README.md#quick-start)
-as well as [jq](https://stedolan.github.io/jq/).
+as well as [`jq`](https://stedolan.github.io/jq/).
 
 ## But JSON
 
@@ -195,7 +195,7 @@ produces
 
 ## Comprehensive Types
 
-ZSON also has a [comprehensive type system](../formats/zed.md#).
+ZSON also has a [comprehensive type system](../formats/zed.md).
 
 For example, here is ZSON "record" with a taste of different types
 of values as record fields:
@@ -983,7 +983,7 @@ in the graph and each set of reviewers is another node.
 
 So as a first step, let's figure out how to create each edge, where an edge
 is a relation between the requesting user and the set of reviewers.  We can
-create this in Zed with a ["lateral query"](../zq/language.md#8-lateral-queries).
+create this in Zed with a ["lateral subquery"](../zq/language.md#8-lateral-subqueries).
 Instead of computing a set-union over all the reviewers across all PRs,
 we instead want to compute the set-union over the reviewers in each PR.
 We can do this as follows:
--- a/docs/tutorials/zq-primer.md
+++ b/docs/tutorials/zq-primer.md
@@ -22,7 +22,7 @@ as well as [`jq`](https://stedolan.github.io/jq/).
 ## But JSON
 
 While `zq` is based on a new type of [data model called Zed](../formats/zed.md),
-it just so happens to be a superset of JSON.
+Zed just so happens to be a superset of JSON.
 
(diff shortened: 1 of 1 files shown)


docs/tutorials/zq-primer.md
```diff
@@ -17,20 +17,20 @@ pulled from the GitHub API.
...
 If you want to follow along on the command line,
 just make sure the `zq` command [is installed](../../README.md#quick-start)
-as well as [jq](https://stedolan.github.io/jq/).
+as well as [`jq`](https://stedolan.github.io/jq/).
 
 ## But JSON
 
 While `zq` is based on a new type of [data model called Zed](../formats/zed.md),
-it just so happens to be a superset of JSON.
+Zed just so happens to be a superset of JSON.
```
philrz, line 25: I know the back-to-back comma-separated "Zed" is may seem a little odd, but as originally written, I felt like the "it" could have been read as a reference to `zq` _or_ to Zed, hence better to be explicit since this is intro text for many readers.

docs/tutorials/zq-primer.md
```diff
@@ -892,20 +892,20 @@ argument to run all the transformations in one fell swoop:
...
+the structure and clarity of the relational model while retaining the flexibility
 of JSON's document model.  No need to create tables then issue SQL insert commands
 to put your clean data into all the right places.
 
 Let's start with something simple.  How about we output a "PR Report" listing
-the title of each PR along with it's PR number and creation date:
+the title of each PR along with its PR number and creation date:
 ```mdtest-command dir=docs/tutorials
-zq -f table '{DATE:created_at,"NUMBER":"PR #${number}",TITLE:title}' prs.zng
+zq -f table '{DATE:created_at,NUMBER:"PR #${number}",TITLE:title}' prs.zng
```
philrz, line 908: At first I assumed the quotes around `"NUMBER"` were significant for some reason, but the surrounding text didn't say anything and removing them didn't change the results, so...

docs/tutorials/zq-primer.md
```diff
@@ -249,13 +249,13 @@ written as `[int64]`.
...
 
-`v10` is a "map" type, which in other languages are often called a "table"
+`v10` is a "map" type, which in other languages is often called a "table"
 or a "dictionary".  In Zed, a value of any type can be used for the key or the
 value though all of the keys and all of the values must have the same type.
 
 Finally, `v11` is a Zed "record", which is similar to a JSON "object", but the
 keys are called "fields", the order of the fields is significant and
-is always preserved, and the field values all have a uniform type.
+is always preserved.
```
philrz, line 258: The removed text "...and the field values all have a uniform type" didn't make sense to me, so I checked with @mccanne on Slack and confirmed that he doesn't remember what he meant by that, so I've proposed removing it. Maybe someone can recall why it was there and suggest something to put in its place that's clearer.

nwt (approved):

PR #3771, opened 2022-03-27
Additional fixes to zed-primer

philrz:
I did one more proofreading pass on the Zed primer and found a double hash in a hyperlink was messing up the ability to jump right to the intended spot. While I was in there I found one additional item to clarify.


nwt (approved):

mattnibs:
Looks like we have a flaky test failure unrelated to this pr.

PR #3783, opened 2022-03-31
api: Handle base base62 tags in requests

mattnibs:
Closes #3775

PR #3788, opened 2022-04-02
API doc improvements

philrz:
The examples shown in this PR rely on the fixes in #3783.

I did another review pass after #3751 merged and ended up doing some actual black box testing and finding genuine corrections and improvements. By the time I was done I had working examples for each, so I figured I'd just go all the way and include them.


docs/zed/api.md
```diff
@@ -44,13 +40,55 @@ POST /pool
 | Name | Type | In | Description |
 | ---- | ---- | -- | ----------- |
 | name | string | body | **Required.** Name of the pool. Must be unique to lake. |
-| layout.order | string | body | Order of value storage in pool. Possible values: desc, asc. Default: asc. |
-| layout.keys | [string] | body | Primary key(s) of pool. Default: ts. |
+| layout.order | string | body | Order of storage by primary key(s) in pool. Possible values: desc, asc. Default: asc. |
+| layout.keys | [[string]] | body | Primary key(s) of pool. The element of each inner string array should reflect the hierarchical ordering of named fields within indexed records. Default: [[ts]]. |
```
philrz, line 44: As pointed out by @jameskerr when I asked about this on Slack, I suppose this could be simplified to accept array elements that are mere strings if the user is seeking to treat a top-level field as a key.

docs/zed/api.md
```diff
@@ -200,42 +462,26 @@ data: {"pool_id": "1sMDXpVwqxm36Rc2vfrmgizc3jz", "commit_id": "1tisISpHoWI7MAZdF
...
-POST /index
-```
-
-#### Delete Index Rule
 
-Deletes the specified index rule. Any created object indexes will persist.
-
-```
-DELETE /index
-```
```
philrz, line 222: I reviewed this one with @mattnibs before proposing to remove it in this PR. What I observed via black box testing was that the `zed index create` seemed to be cooking up its KSUIDs and timestamps client-side and posting them along with the new rule details. I'm not a pro at interpreting the code, but I think these snippets confirm that:

[https://github.com/brimdata/zed/blob/7e9f5e5a51ee061e7362244a5245bbdd9766a3a3/cmd/zed/index/create.go#L109](https://github.com/brimdata/zed/blob/7e9f5e5a51ee061e7362244a5245bbdd9766a3a3/cmd/zed/index/create.go#L109)  
[https://github.com/brimdata/zed/blob/7e9f5e5a51ee061e7362244a5245bbdd9766a3a3/lake/index/rule.go#L50-L55](https://github.com/brimdata/zed/blob/7e9f5e5a51ee061e7362244a5245bbdd9766a3a3/lake/index/rule.go#L50-L55)  
[https://github.com/brimdata/zed/blob/7e9f5e5a51ee061e7362244a5245bbdd9766a3a3/lake/index/rule.go#L23-L27](https://github.com/brimdata/zed/blob/7e9f5e5a51ee061e7362244a5245bbdd9766a3a3/lake/index/rule.go#L23-L27)  
  
If I'm right about that, this seems like it would be awkward to explain to an API user and ask them to adapt to, so we agreed that for now we could just leave these endpoints undocumented until @mattnibs has some time to look closer, and maybe the API will change regardless.

docs/zed/api.md
```diff
@@ -200,42 +462,26 @@ data: {"pool_id": "1sMDXpVwqxm36Rc2vfrmgizc3jz", "commit_id": "1tisISpHoWI7MAZdF
...
-receive responses in the desired format, include the mime type of the format in
-the requests ACCEPT HTTP header.
+For response content types, the service can produce a variety of formats. To
+receive responses in the desired format, include the MIME type of the format in
+the request's Accept HTTP header.
 
-If the ACCEPT header is not specified, the service will return json as default.
+If the Accept header is not specified, the service will return JSON as the
+default response format for the endpoints described above, with the exception
+of the [query](#query) endpoint that returns ZSON by default.
```
philrz, line 477: I also pre-cleared this one with @mattnibs since I wasn't sure if the ZSON default returned by `/query` was a bug or a feature, and the consensus was to document it as expected.

philrz, line 477: I also pre-cleared this one with @mattnibs since I wasn't sure if the ZSON default returned by `/query` was a bug or a feature, and the consensus was to document it as expected.

docs/zed/api.md
```diff
@@ -43,13 +40,55 @@ POST /pool
...
+| layout.order | string | body | Order of storage by primary key(s) in pool. Possible values: desc, asc. Default: asc. |
+| layout.keys | [[string]] | body | Primary key(s) of pool. The element of each inner string array should reflect the hierarchical ordering of named fields within indexed records. Default: [[ts]]. |
 | thresh | int | body | The size in bytes of each seek index. |
 
+**Example Request**
+
+```
+curl -X POST \
+     -H 'Content-Type: application/json' \
+     -d '{"name": "inventory", "layout": {"keys": [["product","serial_number"],["warehouse"]]}}' \
```
mattnibs, line 52: I like where your head is at with this but I do not think we want to highlight primary/secondary keys since I'm not sure this functionality is supported throughout the code base.

docs/zed/api.md
mattnibs: Hadn't had a chance to look at this in depth, but if we are cooking up ksuids client side this should be changed.

docs/zed/api.md
mattnibs: Thought about this a little more. It's weird to have an exception like this. My proposal is this: ZSON by default on every endpoint. If the content-type is `application/x-www-form-urlencoded` (the default for curl) it should return JSON. @philrz what do you think about this?

docs/zed/api.md
philrz: @mattnibs: Ok, that sounds seems like a valid concern. FWIW, the help text in `zed create -h` speaks of “comma-separated pool keys” (plural) so it’s not a well-hidden secret overall. But if you think showing it in an actual example is taking it too far, I’d be happy to trim the example to show only the primary.

That said, part of my motivation for showing it how I did was to make sense of the array-of-arrays that `layout.keys` happens to be. In addition to it having been described incorrectly in the prior doc revision as only a single-dimensional array, even after I saw via Wireshark sniffing that it was a double array, it didn’t make sense to me until @jameskerr explained how it was used to handle keys in embedded objects. But if you’d rather have it show an array-of-arrays and not have it really be obvious why it’s like that, I can live with that.

mattnibs (approved):

PR #3791, opened 2022-04-03
zed: delete by predicate

mattnibs:
Add -predicate argument to zed delete command. The value of -predicate
is a simple zed predicate the must compare against the pool key of the
selected pool. If the flag is specified a commit will be created that
deletes all data that evaluates true for the expression.


mattnibs:
This is local only, if we are ok with this approach I'll put up a pr that wire this up for remote lakes.

compiler/compiler.go
```diff
@@ -224,3 +226,43 @@ func (m *meter) Progress() zbuf.Progress {
 	}
 	return out
 }
+
+func ParseRangeExpr(zctx *zed.Context, src string, layout order.Layout) (*zed.Value, string, error) {
```
mattnibs, line 230: I'm not really a fan of putting this in compiler, considering moving this into branch...

mattnibs: This also could be gotten rid of and I could just use strings.Split instead 🤔

lake/branch.go
```diff
@@ -123,6 +127,154 @@ func (b *Branch) Delete(ctx context.Context, ids []ksuid.KSUID, author, message
...
+	if err := zio.CopyWithContext(ctx, w, q.AsReader()); err != nil {
+		return nil, err
+	}
+	if err := w.Close(); err != nil {
+		return nil, err
+	}
+	return w.Objects(), nil
+}
+
+// getDbpObjects gets the object IDs of objects affected by a delete by predicate
```
mattnibs, line 226:
```suggestion
// getDbpObjects gets the object IDs of objects effected by a delete by predicate
```

cmd/zed/delete/command.go
```diff
@@ -1,15 +1,18 @@
...
 	"github.com/brimdata/zed/cli"
 	"github.com/brimdata/zed/cli/lakeflags"
 	"github.com/brimdata/zed/cmd/zed/root"
+	"github.com/brimdata/zed/lake/api"
 	"github.com/brimdata/zed/lakeparse"
 	"github.com/brimdata/zed/pkg/charm"
+	"github.com/segmentio/ksuid"
 )
 
 var Cmd = &charm.Spec{
```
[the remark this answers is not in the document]
mccanne: and zed/README.md...?

lake/branch.go
```diff
@@ -123,6 +127,154 @@ func (b *Branch) Delete(ctx context.Context, ids []ksuid.KSUID, author, message
...
+  case lower %s THRESH => deletes:=collect(id)
+  case upper %s THRESH => copies:=collect(id)
+)`, op, op)
+	} else {
+		src += fmt.Sprintf(`
+| switch (
+  case upper %s THRESH => deletes:=collect(id)
+  case lower %s THRESH => copies:=collect(id)
+)`, op, op)
+	}
```
nwt, line 246:
```suggestion
| switch (
  case %s %s THRESH => deletes:=collect(id)
  case %s %s THRESH => copies:=collect(id)
)`
	deletesField, copiesField := "upper", "lower"
	if op == ">=" || op == ">" {
		deletesField, copiesField = copiesField, deletesField        
        }
	src := fmt.Sprintf(dbp, zson.MustFormatValue(*val), b.pool.ID, commit, deletesField, op, copiesField, op)
```

lake/branch.go
```diff
@@ -123,6 +127,154 @@ func (b *Branch) Delete(ctx context.Context, ids []ksuid.KSUID, author, message
...
+	flowgraph, err := compiler.ParseProc(src)
+	if err != nil {
+		return nil, nil, err
+	}
+	if _, err := rlimit.RaiseOpenFilesLimit(); err != nil {
+		return nil, nil, err
+	}
+	q, err := runtime.NewQueryOnLake(ctx, zed.NewContext(), flowgraph, lake, nil, nil)
```
nwt, line 254: Defer `q.Close` if this succeeds.

lake/branch.go
```diff
@@ -123,6 +127,154 @@ func (b *Branch) Delete(ctx context.Context, ids []ksuid.KSUID, author, message
...
+		defer reader.Close()
+		readers[i] = zngio.NewReader(reader, zctx)
+	}
+	// Keeps values that don't fit the filter by adding "not".
+	flowgraph, err := compiler.ParseProc("not " + filter)
+	if err != nil {
+		return nil, err
+	}
+	r := zio.NewCombiner(ctx, readers)
+	q, err := runtime.NewQueryOnReader(ctx, zctx, flowgraph, r, nil)
```
nwt, line 209: Defer `q.Close` if this succeeds.

lake/branch.go
```diff
@@ -123,6 +127,154 @@ func (b *Branch) Delete(ctx context.Context, ids []ksuid.KSUID, author, message
...
+	w, err := NewWriter(ctx, zctx, b.pool)
+	if err != nil {
+		return nil, err
+	}
+	if err := zio.CopyWithContext(ctx, w, q.AsReader()); err != nil {
+		return nil, err
+	}
+	if err := w.Close(); err != nil {
+		return nil, err
+	}
```
nwt, line 222: 1. Call `w.Close` whether `zio.CopyWithContext` succeeds or fails.
2. Clean up if either `zio.CopyWithContext` or `w.Close` fails.
```suggestion
	err = zio.CopyWithContext(ctx, w, q.AsReader())
	if err2 := w.Close(); err == nil {
		err = err2
	}
	if err != nil {
		for _, o := range w.Objects() {
			o.Remove(ctx, b.pool.engine, b.pool.DataPath)
		}
		return err
	}
```

lake/branch.go
```diff
@@ -123,6 +127,154 @@ func (b *Branch) Delete(ctx context.Context, ids []ksuid.KSUID, author, message
...
+		}
+		copies, deletes, err := b.getDbpObjects(ctx, lake, val, op, parent.Commit)
+		if err != nil {
+			return nil, err
+		}
+		count := len(copies) + len(deletes)
+		if count == 0 {
+			return nil, errors.New("nothing to delete")
+		}
+		// XXX Copied should be deleted if the commit fails.
```
nwt, line 153: Do this in `dpbCopies`.
```suggestion
```

mattnibs: There's a separate case where if the entire commit operation fails these should also be cleaned up, and this cannot happen in dpbCopies.

cmd/zed/delete/command.go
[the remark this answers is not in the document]
mattnibs: Added, please take a look.

cmd/zed/delete/command.go
```diff
@@ -23,6 +27,11 @@ delete operation to HEAD.
 Once the delete operation completes, the deleted data is no longer seen
 when read data from the pool.
 
+If the -where flag is specified, delete will remove all values that return true
+the for the provided boolean expressions . Currently the filter expressions is
+limited in that it must be against the pool key and it must be a single relative
+comparison (e.g., -where 'ts <= now() - 3h').
```
nwt, line 33:
```suggestion
If the -where flag is specified, delete will remove all values for which the
provided filter expression is true. The filter expression must be single comparison
against the pool key using <, <=, > or >= (e.g., -where 'ts <= now() - 3h').
```

docs/zed/README.md
```diff
@@ -494,6 +494,11 @@ simply removes the data from the branch without actually deleting the
 underlying data objects thereby allowing time travel to work in the face
 of deletes.
 
+If the -where flag is specified, delete will remove all values that return true
+the for the provided boolean expressions . Currently the filter expressions is
+limited in that it must be against the pool key and it must be a single relative
+comparison (e.g., `-where 'ts <= now() - 3h'`).
```
nwt, line 500:
```suggestion
If the -where flag is specified, delete will remove all values for which the
provided filter expression is true. The filter expression must be single comparison
against the pool key using <, <=, > or >= (e.g., -where 'ts <= now() - 3h').
```

lake/api/remote.go
```diff
@@ -129,6 +129,10 @@ func (r *RemoteSession) Delete(ctx context.Context, poolID ksuid.KSUID, branchNa
 	return res.Commit, err
 }
 
+func (r *RemoteSession) DeleteByPredicate(ctx context.Context, poolID ksuid.KSUID, branchName, src string, commit api.CommitMessage) (ksuid.KSUID, error) {
+	return ksuid.Nil, errors.New("unsupported")
```
nwt, line 133: We should implement this, but I guess that can wait.

nwt (approved):

lake/api/remote.go
mattnibs: Separate pr.

PR #3795, opened 2022-04-04
rework expression compilation to use kernel.Builder

mccanne:
This commit adapts the compiler to use methods on kernel.Builder
for compiling expressions.  This will be used in a subsequent PR
to allow dataflow operators to appear inside of expressions.
We also moved CompileAssignments to package expr and renamed it
NewAssignments since it was just creating runtime objects and not
actually doing any compilation.

### DIFF
--- a/runtime/expr/eval.go
+++ b/runtime/expr/eval.go
@@ -880,7 +880,7 @@ type Assignment struct {
 
 func NewAssignments(zctx *zed.Context, dsts field.List, srcs field.List) (field.List, []Evaluator) {
 	if len(srcs) != len(dsts) {
-		panic("CompileAssignments: argument mismatch")
+		panic("NewAssignments: argument mismatch")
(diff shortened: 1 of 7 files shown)


mattnibs (approved):
lgtm

runtime/expr/eval.go
```diff
@@ -877,3 +877,16 @@ type Assignment struct {
 	LHS field.Path
 	RHS Evaluator
 }
+
+func NewAssignments(zctx *zed.Context, dsts field.List, srcs field.List) (field.List, []Evaluator) {
+	if len(srcs) != len(dsts) {
+		panic("CompileAssignments: argument mismatch")
```
nwt, line 883:
```suggestion
		panic("NewAssignments: argument mismatch")
```

nwt (approved):

PR #3797, opened 2022-04-04
add support for laterally scoped "over" expressions

mccanne:
This commit extends the "over" lateral query concept to appear
within expression context, lifting the complicated constraint that
all results from a single over operators must exit a dataflow scope
and be pasted together.

This change introduces a notable slowdown in parsing for "over"
expressions embedded in complex nested literals.  We diagnosed
the problem as customary backtracking behavior intrinsic to the
PEG parsing model itself and have identified several opportunities
to restructure PEG grammar rules to avoid excessing backtracking.
These improvements will be submitted in a subsequent PR.

### DIFF
--- a/compiler/semantic/expr.go
+++ b/compiler/semantic/expr.go
@@ -224,6 +224,30 @@ func semExpr(scope *Scope, e ast.Expr) (dag.Expr, error) {
 			Kind:    "MapExpr",
 			Entries: entries,
 		}, nil
+	case *ast.OverExpr:
+		exprs, err := semExprs(scope, e.Exprs)
+		if err != nil {
+			return nil, err
+		}
+		if e.Scope == nil {
+			return nil, errors.New("over expression missing scope in AST")
+		}
+		scope.Enter()
+		defer scope.Exit()
+		locals, err := semVars(scope, e.Locals)
+		if err != nil {
+			return nil, err
+		}
+		seq, err := semSequential(nil, scope, e.Scope, nil, nil)
+		if err != nil {
+			return nil, err
+		}
+		return &dag.OverExpr{
+			Kind:  "OverExpr",
+			Defs:  locals,
+			Exprs: exprs,
+			Scope: seq,
+		}, nil
 	}
 	return nil, fmt.Errorf("invalid expression type %T", e)
 }
--- a/compiler/semantic/proc.go
+++ b/compiler/semantic/proc.go
@@ -236,7 +236,7 @@ func semSequential(ctx context.Context, scope *Scope, seq *ast.Sequential, adapt
 func semProc(ctx context.Context, scope *Scope, p ast.Proc, adaptor op.DataAdaptor, head *lakeparse.Commitish) (dag.Op, error) {
 	switch p := p.(type) {
 	case *ast.From:
-		if ctx == nil {
+		if adaptor == nil {
 			return nil, errors.New("from operator not allowed inside of expression")
 		}
 		return semFrom(ctx, scope, p, adaptor, head)
--- a/compiler/semantic/proc.go
+++ b/compiler/semantic/proc.go
@@ -236,6 +236,9 @@ func semSequential(ctx context.Context, scope *Scope, seq *ast.Sequential, adapt
 func semProc(ctx context.Context, scope *Scope, p ast.Proc, adaptor op.DataAdaptor, head *lakeparse.Commitish) (dag.Op, error) {
 	switch p := p.(type) {
 	case *ast.From:
+		if ctx == nil {
+			return nil, errors.New("from operator not allowed inside of expression")
+		}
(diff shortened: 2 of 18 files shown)


compiler/semantic/expr.go
```diff
@@ -224,6 +224,30 @@ func semExpr(scope *Scope, e ast.Expr) (dag.Expr, error) {
...
+		if e.Scope == nil {
+			return nil, errors.New("over expression missing scope in AST")
+		}
+		scope.Enter()
+		defer scope.Exit()
+		locals, err := semVars(scope, e.Locals)
+		if err != nil {
+			return nil, err
+		}
+		seq, err := semSequential(nil, scope, e.Scope, nil, nil)
```
nwt, line 241: The package context documentation discourages passing a nil context.  (This requires changing `if ctx == nil {` to `if adapter == nil {` in `semProc`.)
```suggestion
		seq, err := semSequential(ctx, scope, e.Scope, nil, nil)
```

compiler/semantic/proc.go
```diff
@@ -236,6 +236,9 @@ func semSequential(ctx context.Context, scope *Scope, seq *ast.Sequential, adapt
 func semProc(ctx context.Context, scope *Scope, p ast.Proc, adaptor op.DataAdaptor, head *lakeparse.Commitish) (dag.Op, error) {
 	switch p := p.(type) {
 	case *ast.From:
+		if ctx == nil {
```
nwt, line 239: As mentioned elsewhere, the package context documentation discourages passing a nil context.  Conditioning this on having an adaptor also feels more natural.
```suggestion
		if adaptor == nil {
```

runtime/op/traverse/eval.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/brimdata/zed"
+	"github.com/brimdata/zed/runtime/expr"
+	"github.com/brimdata/zed/zbuf"
+	"github.com/brimdata/zed/zcode"
+)
+
+// Expr provides provides glue to run a traversal subquery in expression
+// context.  It implements zbuf.Puller so it can serve as the data source
+// to the subquery as well as expr.Evalulator so it can be called from an
+// expression.  Each time it's Eval method is called, it propagates the value
```
nwt, line 15:
```suggestion
// expression.  Each time its Eval method is called, it propagates the value
```

runtime/op/traverse/eval.go
```diff
@@ -0,0 +1,131 @@
...
+// to the batch channel to be pulled into the scope.  If there is
+// just one result, then the value is returned.  If there are multiple results
+// then they are returned in an array create union elements if the type varies.
```
nwt, line 18:
```suggestion
// then they are returned in an array (with union elements if the type varies).
```

runtime/op/traverse/eval.go
```diff
@@ -0,0 +1,131 @@
...
+	for typ := range types {
+		utypes = append(utypes, typ)
+	}
+	union := e.zctx.LookupTypeUnion(utypes)
+	var b zcode.Builder
+	for _, val := range vals {
+		b.BeginContainer()
+		b.Append(zed.EncodeInt(int64(union.Selector(val.Type))))
+		b.Append(val.Bytes)
+		b.EndContainer()
```
nwt, line 114: Use `zed.BuildUnion` for null handling consistent with what we generally have elsewhere. (To be clear on the difference, for `val.Bytes==nil`, `zed.BuildUnion` will append a null union value, while the code here will append a non-null union value containing a (typed) null value.)
```suggestion
		zed.BuildUnion(union.Selector(val.Type), val.Bytes)
```

compiler/semantic/expr.go
mccanne: Context is available here as it is not needed along this code path so I get the convention but I think it would be best to pass nil.  If the nil gets ends up getting used below somehow thats a bug that I would want to panic on.  But I will change the check from ctx to adaptor != nil.

runtime/op/traverse/eval.go
mccanne: Good catch!

docs/zq/language.md
```diff
@@ -1760,6 +1760,48 @@ produces
 {collect:[1,4,7]}
 {collect:[1,2,3]}
 ```
+
+### 8.2 Lateral Expressions
+
+Lateral subqueries can also appear in expression context using the
+parenthesized form:
+```
+( over <expr> [, <expr>...] [with <var>=<expr> [, ... <var>[=<expr>]] | <lateral> )
```
nwt, line 1769: Having different syntax for the operator and expression forms is surprising. I think we should unify them.

nwt (approved):

docs/zq/language.md
mccanne: Agreed this needs some more thought.  Let's merge and revisit soon.

PR #3799, opened 2022-04-06
Service support for delete by predicate

mattnibs:
Add support for delete by predicate in the service delete endpoint.

PR #3800, opened 2022-04-06
Small improvements to delete docs

philrz:
After the merge of #3791 and #3799 I tried out the new `zed delete -where` and the experience was :+1:. I did notice a couple small doc improvements to suggest, though, so here they are.

I'll also point out that the top-level command usage summary is technically inaccurate now:

```
zed delete [options] <id> [<id>...]
```

...since `zed delete -where` doesn't require an `<id>`. I'm not suggesting a specific way to address this, but options might seem to include:

1. Shorten to `zed delete [options] [<id>...]`, since it's revealed below that one of the options makes the `<id>` portion optional.
2. Add a second usage line that looks something like `zed delete [options] -where <filter>`
3. Not change the top-line usage guidance at all, such as if this delete-by-predicate approach is considered new enough that it's not intended to draw a lot of attention yet

If reviewers chime in with a preferred approach I'd be happy to push another change on this PR.


mattnibs (approved):

mattnibs:
Thansk @philrz !

philrz:
@mattnibs: Any opinion on the three options I laid out above on the usage text?

mattnibs:
I think I like 1. I'd say 2. but I'm not sure charm in its current incarnation supports multiple usage lines.

philrz:
@mattnibs: FWIW, I hadn’t thought of pushing multiple usage lines into charm. I just meant showing both of them in that Zed README doc, like I’ve seen done on some other pages, e.g. [https://github.com/brimdata/zed/tree/main/docs/zq#1-usage](https://github.com/brimdata/zed/tree/main/docs/zq#1-usage). If that swings you back to preferring option 2 in the doc, I can push a commit of that, otherwise option 1. Will wait to hear from you.

mattnibs:
@philrz you are right! In that case go with 2.

philrz:
@mattnibs: Cool. I just pushed a commit for option 2. I don't have the permissions to merge, so feel free to hit that from your end if you're satisfied.

PR #3807, opened 2022-04-07
zson.Unmarshal: Improve handling of null values

mattnibs:
This pr fixes a couple issues when unmarshaling null zed values:
- zson.Unmarshal was panicing when trying to unmarshal a "null" value
   into an array.
- Typed null values (e.g. "null(int64)") were not getting typed check if
   the go value was a pointer.
- Ensure nested pointers are getting properly traversed.

### DIFF
--- a/zson/marshal.go
+++ b/zson/marshal.go
@@ -614,7 +614,7 @@ func UnmarshalZNGRecord(rec *zed.Value, v interface{}) error {
 }
 
 func incompatTypeError(zt zed.Type, v reflect.Value) error {
-	return fmt.Errorf("incompatible type translation: zng type %v go type %v go kind %v", zt, v.Type(), v.Kind())
+	return fmt.Errorf("incompatible type translation: zng type %v go type %v go kind %v", FormatType(zt), v.Type(), v.Kind())
 }
 
 func (u *UnmarshalZNGContext) Unmarshal(zv zed.Value, v interface{}) error {
@@ -680,6 +680,13 @@ func (u *UnmarshalZNGContext) decodeAny(zv zed.Value, v reflect.Value) error {
 		v.Set(reflect.ValueOf(*zv.Copy()))
 		return nil
 	}
+	if zed.TypeUnder(zv.Type) == zed.TypeNull && v.CanSet() {
+		// Since a zed TypeNull can resolve to any value go ahead an set the
+		// zero value if the value can be set. Otherwise the reflect type is
+		// probably a pointer and we'll take care of this below.
+		v.Set(reflect.Zero(v.Type()))
+		return nil
+	}
 	switch v.Kind() {
 	case reflect.Array:
 		return u.decodeArray(zv, v)
@@ -724,14 +731,16 @@ func (u *UnmarshalZNGContext) decodeAny(zv zed.Value, v reflect.Value) error {
 		}
 		return nil
 	case reflect.Ptr:
-		if zv.Bytes == nil {
-			v.Set(reflect.Zero(v.Type()))
-			return nil
+		for v.Kind() == reflect.Ptr {
+			if v.CanSet() && zv.Bytes == nil {
+				return u.decodeNull(zv, v)
+			}
+			if v.IsNil() {
+				v.Set(reflect.New(v.Type().Elem()))
+			}
+			v = v.Elem()
 		}
-		if v.IsNil() {
-			v.Set(reflect.New(v.Type().Elem()))
-		}
-		return u.decodeAny(zv, v.Elem())
+		return u.decodeAny(zv, v)
 	case reflect.String:
 		// XXX We bundle string, type, error all into string.
 		// See issue #1853.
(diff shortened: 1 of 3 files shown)


zson/marshal.go
```diff
@@ -680,6 +680,13 @@ func (u *UnmarshalZNGContext) decodeAny(zv zed.Value, v reflect.Value) error {
 		v.Set(reflect.ValueOf(*zv.Copy()))
 		return nil
 	}
+	if typ := zed.TypeUnder(zv.Type); typ == zed.TypeNull && v.Kind() != reflect.Ptr {
+		// Since zed.TypeNull can be any type, set the interface to zero value.
+		// We need to make sure the reflect value isn't a pointer since pointers
+		// structs on null values should remain a nil pointer.
```
nwt, line 686: I don't follow this. Can you elaborate on "pointers structs on null values should remain a nil pointer"?

zson/marshal_zng_test.go
```diff
@@ -202,6 +202,31 @@ func TestUnmarshalRecord(t *testing.T) {
 	require.Equal(t, *v1.T1f1, *v3.T4f1)
 }
 
+func TestUnmarshalNull(t *testing.T) {
+	v, err := zson.ParseValue(zed.NewContext(), "null")
+	require.NoError(t, err)
```
nwt, line 207: Use `zed.Null`!

zson/marshal_zng_test.go
```diff
@@ -202,6 +202,31 @@ func TestUnmarshalRecord(t *testing.T) {
...
+	var array []int
+	array = []int{1}
+	require.NoError(t, zson.UnmarshalZNG(*v, &array))
+	assert.Len(t, array, 0)
```
nwt, line 211:
```suggestion
	slice := []int{1}
	require.NoError(t, zson.UnmarshalZNG(*zed.Null, &slice))
	assert.Nil(t, slice)
```

zson/marshal_zng_test.go
```diff
@@ -202,6 +202,31 @@ func TestUnmarshalRecord(t *testing.T) {
...
+	type testobj struct {
+		Val int
+	}
+	var obj *testobj
+	require.NoError(t, zson.UnmarshalZNG(*v, &obj))
+	assert.Nil(t, obj)
+	// Test if a non-array null returns an error.
+	v, err = zson.ParseValue(zed.NewContext(), "null(int64)")
+	require.NoError(t, err)
+	assert.EqualError(t, zson.UnmarshalZNG(*v, &array), "not an array")
```
nwt, line 227:
```suggestion
	assert.EqualError(t, zson.UnmarshalZNG(*zed.NullInt64, &slice), "not an array")
```

zson/marshal.go
mattnibs: Yeah I'm going to do another pass of comments on this logic because it is confusing.

zson/marshal_zng_test.go
```diff
@@ -202,6 +202,28 @@ func TestUnmarshalRecord(t *testing.T) {
...
+	require.NoError(t, zson.UnmarshalZNG(*v, &integer))
+	assert.Equal(t, integer, 0)
+	m := map[string]string{"key": "value"}
+	require.NoError(t, zson.UnmarshalZNG(*v, &m))
+	assert.Nil(t, m)
+	type testobj struct {
+		Val int
+	}
+	var obj *testobj
+	require.NoError(t, zson.UnmarshalZNG(*v, &obj))
```
nwt, line 221:
```suggestion
	slice := []int{1}
	require.NoError(t, zson.UnmarshalZNG(*zed.Null, &slice))
	assert.Nil(t, slice)
	integer := -1
	require.NoError(t, zson.UnmarshalZNG(*zed.Null, &integer))
	assert.Equal(t, integer, 0)
	m := map[string]string{"key": "value"}
	require.NoError(t, zson.UnmarshalZNG(*zed.Null, &m))
	assert.Nil(t, m)
	type testobj struct {
		Val int
	}
	var obj *testobj
	require.NoError(t, zson.UnmarshalZNG(*zed.Null, &obj))
```

zson/marshal.go
mattnibs: Updated, I hope this is clearer.

zson/marshal_zng_test.go
mattnibs: I reformatted this test.

mccanne:
I added this to the end of TestUnmarshalNull() and it failed...
```
v, err = zson.ParseValue(zed.NewContext(), "null([int64])")
require.NoError(t, err)
slice = []int{1}
require.NoError(t, zson.UnmarshalZNG(*v, &slice))
assert.Nil(t, slice)
```

mattnibs:
@mccanne yeah it wasn't doing inner type checking for maps, arrays, sets. I suppose we should do this here while we are at it.

mattnibs:
@mccanne thanks I resolved the issue you found plus fixed the same issue for maps and bytes. Node: the code is not currently checking the inner types of null values- this doesn't seem trivial to do so I just put a comment in the code and perhaps revisit or I can fix now?

mccanne (approved):

PR #3810, opened 2022-04-08
simplfiy lake/api for external use

mccanne:
This commit simplifies package lake/api by not exporting
the implementations and making it easier to open a lake
with either a string URL or parsed storage.URI.  We added
the OpenLake method which figures out if a local or remote
instance should be used based on the URI.

### DIFF
--- a/lake/api/api.go
+++ b/lake/api/api.go
@@ -4,15 +4,16 @@ import (
 	"context"
 	"encoding/hex"
 	"fmt"
+	"strings"
 
 	"github.com/brimdata/zed"
 	"github.com/brimdata/zed/api"
+	"github.com/brimdata/zed/api/client"
 	"github.com/brimdata/zed/lake"
 	"github.com/brimdata/zed/lake/index"
 	"github.com/brimdata/zed/lake/pools"
 	"github.com/brimdata/zed/lakeparse"
 	"github.com/brimdata/zed/order"
-	"github.com/brimdata/zed/pkg/storage"
 	"github.com/brimdata/zed/zbuf"
 	"github.com/brimdata/zed/zio"
 	"github.com/brimdata/zed/zson"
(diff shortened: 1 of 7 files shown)


lake/api/api.go
```diff
@@ -40,6 +40,17 @@ type Interface interface {
...
+	uri, err := storage.ParseURI(u)
+	if err != nil {
+		return nil, err
+	}
+	if IsLakeService(uri) {
+		return OpenRemoteLakeWithURI(ctx, uri)
+	}
+	return OpenLocalLakeWithURI(ctx, uri)
+}
+
```
mattnibs, line 53: This isn't being used, nor can it with the cli since it needs to set the auth store on the connection.

lake/api/remote.go
```diff
@@ -12,42 +12,47 @@ import (
...
 }
 
-func NewRemoteWithConnection(conn *client.Connection) *RemoteSession {
-	return &RemoteSession{conn}
+func OpenRemoteLakeWithURI(ctx context.Context, uri *storage.URI) (Interface, error) {
+	return OpenRemoteLake(ctx, uri.String())
 }
 
-func (r *RemoteSession) PoolID(ctx context.Context, poolName string) (ksuid.KSUID, error) {
+func NewRemoteWithConnection(conn *client.Connection) *remote {
```
nwt, line 38:
```suggestion
func NewRemoteWithConnection(conn *client.Connection) Interface {
```

lake/api/api.go
nwt: @mattnibs: I think it's for use in example code.

lake/api/remote.go
```diff
@@ -12,42 +12,47 @@ import (
...
+	return &remote{conn}
+}
```
nwt, line 40: 1. The context parameters and error return parameters of `OpenRemoteLake` and `OpenRemoteLakeWithURI` are unused.
2. The name of `NewRemoteWithConnection` and its `*remote` return type are inconsistent with the other two constructors.
3. Three constructors feels pretty busy.

Based on all that, I think we should just have this:
```go
func NewRemoteLake(conn *client.Connection) Interface {
	return &remote{conn}
}
```

PR #3811, opened 2022-04-08
fix some ordering and numbering in docs/formats/zson.md

nwt:
@philrz pointed out these inconsistencies.


philrz:
Thanks @nwt! That does indeed seem to address the problems I spotted. 👍

mccanne (approved):

PR #3818, opened 2022-04-14
docs: Clarify 'Accept' header for query endpoint

mattnibs:
Add documentation clarifying the use of the Accept header for the Zed
service query endpoint. Add csv as an Media-Type to docs.


mattnibs:
Closes #3818

mccanne (approved):

PR #3819, opened 2022-04-14
zson: Pretty print type values

mattnibs:
Add indenting and coloring when formatting type values. Also
standardizes all formatting of type values into single implementation.

### DIFF
--- a/zson/formatter.go
+++ b/zson/formatter.go
@@ -196,14 +196,164 @@ func (f *Formatter) formatValue(indent int, typ zed.Type, bytes zcode.Bytes, par
 		err = f.formatValue(indent, t.Type, bytes, known, parentImplied, false)
 		f.build(")")
 	case *zed.TypeOfType:
-		f.startColorPrimitive(zed.TypeType)
-		f.buildf("<%s>", FormatTypeValue(bytes))
+		f.startColor(color.Gray(200))
+		f.build("<")
+		f.formatTypeValue(indent, bytes)
+		f.build(">")
 		f.endColor()
 	}
 	if err == nil && decorate {
 		f.decorate(typ, parentKnown, null)
 	}
 	return err
+}
+
+func (f *Formatter) formatTypeValue(indent int, tv zcode.Bytes) zcode.Bytes {
+	n, tv := zed.DecodeLength(tv)
+	if tv == nil {
+		f.truncTypeValueErr()
+		return nil
+	}
+	switch n {
+	default:
+		if n < 0 || n > zed.TypeValueMax {
+			f.buildf("<ERR bad type ID %d in type value>", n)
+			return nil
+		}
+		typ := zed.LookupPrimitiveByID(int(n))
+		f.startColorPrimitive(typ)
+		f.build(zed.PrimitiveName(typ))
+		f.endColor()
+	case zed.TypeValueNameDef:
+		name, tv := zed.DecodeName(tv)
+		if tv == nil {
+			f.truncTypeValueErr()
+			return nil
+		}
+		f.build(name)
+		f.build("=")
+		return f.formatTypeValue(indent, tv)
+	case zed.TypeValueNameRef:
+		name, tv := zed.DecodeName(tv)
+		if tv == nil {
+			f.truncTypeValueErr()
+			return nil
+		}
+		f.build(name)
+		return tv
+	case zed.TypeValueRecord:
+		f.build("{")
+		var n int
+		n, tv = zed.DecodeLength(tv)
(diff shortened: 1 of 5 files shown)


mccanne:
A couple thoughts
* The color scheme is a little jarring to me... how about we live it as gray then we can experiment with colored approaches and we use it more
* Array, set, and map types will typically be simple, e.g.,  I don't think `[int64]` should be vertically formatted.  Maybe we could have a heuristic that vertically formats these only when there is a compound type inside, e.g., array of object or even array of array could be 
```
[
    [int64]
]
```

mattnibs:
@mccanne updated, I took away the colors and gave primitives a slightly darker gray:
image

mattnibs:
Also I just based the heuristic for indenting arrays, sets, maps based off of if the inner type is a primitive. I could enable this also for enums, errors, named types but this didn't seem quite as straightforward, not to mention unraveling to types for maps where you need to find the type for the key and value is a harder problem.

zson/formatter.go
```diff
@@ -206,6 +208,174 @@ func (f *Formatter) formatValue(indent int, typ zed.Type, bytes zcode.Bytes, par
...
+			return nil
+		}
+		for k := 0; k < n; k++ {
+			if k > 0 {
+				f.build(",")
+			}
+			var symbol string
+			symbol, tv = zed.DecodeName(tv)
+			if tv == nil {
+				return nil
```
nwt, line 339:
```suggestion
				f.truncTypeValueErr()
				return nil
```

zson/formatter.go
```diff
@@ -206,6 +208,174 @@ func (f *Formatter) formatValue(indent int, typ zed.Type, bytes zcode.Bytes, par
...
+	return tv
+}
+
+func (f *Formatter) formatVectorTypeValue(token string, indent int, tv zcode.Bytes) zcode.Bytes {
+	f.build(token)
+	end := "]"
+	if token == "|[" {
+		end += "|"
+	}
+	if n, _ := zed.DecodeLength(tv); n <= zed.IDTypeComplex {
```
nwt, line 361: Check for an error and call `Formatter.truncTypeValueErr`.

zson/formatter.go
```diff
@@ -206,6 +208,174 @@ func (f *Formatter) formatValue(indent int, typ zed.Type, bytes zcode.Bytes, par
...
+		tv = f.formatVectorTypeValue("[", indent, tv)
+	case zed.TypeValueSet:
+		tv = f.formatVectorTypeValue("|[", indent, tv)
+	case zed.TypeValueMap:
+		f.build("|{")
+		newline := f.newline
+		indent += f.tab
+		if n, itv := zed.DecodeLength(tv); n <= zed.IDNull {
+			n, _ = zed.DecodeLength(itv)
+			if n <= zed.IDNull {
```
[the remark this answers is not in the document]
mattnibs: The errors from decode length will get caught below when this is actually rendered.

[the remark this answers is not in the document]
mattnibs: Also curious why `< zed.IDTypeComplex` is superior to `<= zed.IDNull`?

zson/formatter.go
mattnibs: No need, as this will get caught on the call to `formatTypeValue`.

mattnibs: But this should be `n < zed.IDTypeComplex`

zson/formatter.go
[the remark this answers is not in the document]
nwt: Unlike `IDNull`, it's robust to addition of primitive types. If we ever add more, `IDTypeComplex` will change accordingly.

zson/ztests/pretty-type-value.yaml
```diff
@@ -0,0 +1,52 @@
+zed: "yield this, '==='"
```
nwt, line 1:
```suggestion
zed: yield this, "==="
```

zson/ztests/pretty-type-value.yaml
```diff
@@ -0,0 +1,52 @@
...
+input: |
+  <({foo:int64},{bar:string},uint64)>
+  <named=|[({foo:int64},uint64)]|>
+  <|{{x:string}:{ip:ip,port:uint16}}|>
+  <|{string:uint64}|>
+  <[int64]>
+  <error([string])>
+  <[[string]]>
+
+output-flags: "-pretty=4"
```
nwt, line 12:
```suggestion
output-flags: -pretty=4
```

nwt (approved):

PR #3820, opened 2022-04-19
Fixes in ZNG format spec

philrz:
From a proofreading pass.

### DIFF
--- a/docs/formats/zng.md
+++ b/docs/formats/zng.md
@@ -135,7 +135,7 @@ Values for the `format` byte are defined in the
 
 > This arrangement of message blocks separating types and values allows
 > for efficient scanning and parallelization.  In general, values depend
-> on type definitions but as long as all of the types are known before
+> on type definitions but as long as all of the types are known when
 > values are used, decoding can be done in parallel.  Likewise, since
 > each block is independently compressed, the blocks can be decompressed
 > in parallel.  Moreover, efficient filtering can be carried out over
--- a/docs/formats/zng.md
+++ b/docs/formats/zng.md
@@ -57,7 +57,7 @@ allowing an implementation to easily skip from frame to frame.
 
 Each frame begins with a single-byte "frame code":
 ```
-    7 6 5 4 5 3 1 0
+    7 6 5 4 3 2 1 0
    +-+-+-+-+-+-+-+-+
    |V|C|  T|      L|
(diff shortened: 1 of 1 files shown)


docs/formats/zng.md
```diff
@@ -135,8 +135,8 @@ Values for the `format` byte are defined in the
 
 > This arrangement of message blocks separating types and values allows
 > for efficient scanning and parallelization.  In general, values depend
-> on type definitions but as long as all of the types are known by the
-> time values are used, decoding can be done in parallel.  Likewise, since
+> on type definitions but as long as all of the types are known before
+> values are used, decoding can be done in parallel.  Likewise, since
```
philrz, line 139: I proposed this change because upon first read I had difficulty parsing the sentence, as my eyes got stuck on "time values" and I found myself wondering "Wait... what do the values of `time` types have to do with this?" Once I figured out how I was misinterpreting the sentence, I figured an alternate wording might be justified.

docs/formats/zng.md
```diff
@@ -420,7 +420,7 @@ messages provide additional semantics on top of the base ZNG format.
 Any such application-defined messages not known by
 a ZNG data receiver shall be ignored.
 
-The body of control message is JSON, ZSON, ZNG, binary, or UTF-8 text.
+The body of a control message is JSON, ZSON, ZNG, binary, or UTF-8 text.
```
philrz, line 423: Side note: I noticed that the section titles consistently speak of "frames" but then "message" is often used interchangeably within the text. I'm not sure if it might be appropriate to standardize on"frame" everywhere. I didn't try to take that step, but figured I'd point it out in case there's an interest in standardizing.

philrz:
Also, I didn't come up with specific changes to propose to this sentence, so I can't comment on it in the PR diffs. But I wanted to point out that I had real difficulty reading this sentence as is and couldn't tell if it's got mistakes in it or if I'm just too dumb to understand it. It would read clean to me if the middle text "...key and value encoded as..." was dropped, but I don't know if that changes the correctness of the statement.

> A map value is encoded as a container as a sequence of alternating tag-encoded key and value encoded as keys and values of the underlying key and value types.

philrz:
Oh yeah. And I picked the wrong branch name. Oops. :smile:

nwt (approved):

PR #3823, opened 2022-04-20
rework some parser rules to reduce backtracking

nwt:
Unnecessary backtracking can really slow things down.  For example, this
change makes "zc '[1,1.5,{s:[1,{x:1}]}]'" run about 200 times faster.

### DIFF
--- a/compiler/parser/parser.peg
+++ b/compiler/parser/parser.peg
@@ -49,11 +49,8 @@
 start = __ ast:Sequential __ EOF { RETURN(ast) }
 
 Sequential
-  = consts:Consts __ first:Operation rest:SequentialTail+ {
+  = consts:Consts __ first:Operation rest:SequentialTail* {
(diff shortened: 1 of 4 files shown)


compiler/parser/parser.peg
```diff
@@ -758,12 +747,9 @@ Primary
   / "(" __ expr:Expr __ ")" { RETURN(expr) }
 
 OverExpr
-  = "over" _ exprs:Exprs _ "with" _ locals:LetAssignments __ "|" __ scope:Sequential {
+  = "over" _ exprs:Exprs locals:(_ "with" _ l:LetAssignments { RETURN(l) })? __ "|" __ scope:Sequential {
```
mattnibs, line 750: Idk about having a RETURN in the middle of the definition here, maybe break the `with` into its own statement?

mattnibs (approved):

PR #3826, opened 2022-04-21
Proper coercion for aggregation math fns

mattnibs:
Instead of using the first type to set the type for aggregation min,
max, and sum, use coercion rules to select the right type from the range
of values.


runtime/expr/agg/math.go
```diff
@@ -39,32 +48,46 @@ func (m *mathReducer) Consume(val *zed.Value) {
...
-	}
-	if val.IsNull() {
-		return
+	var id int
+	if m.math != nil {
+		var err error
+		// XXX We're not using the value coercion parts of coerce.Pair here.
+		// Would be better if coerce had a function that just compared types
+		// and returned the type to coerce to.
+		id, err = m.pair.Coerce(&m.nilVal, val)
```
nwt, line 57: No need for `nilVal` here. The `Coerce` arguments don't escape to the heap so the `zed.NewValue` return value goes on the stack. (You can see this in the output of `go build -gcflags=-m ./runtime/expr/agg` and `go tool objdump -S -s mathReducer.*consumeVal dist/zq`).
```suggestion
		id, err = m.pair.Coerce(val, zed.NewValue(m.math.typ(), nil))
```

runtime/expr/agg/ztests/math-coercion.yaml
```diff
@@ -0,0 +1,13 @@
+zed: 'sum(this), max(this), min(this), avg(this)'
+
+input: '2 2.5'
+
+output-flags: '-pretty 4'
```
nwt, line 5:
```suggestion
zed: sum(this), max(this), min(this), avg(this)

input: |
  2
  2.5

output-flags: -pretty=4
```

nwt (approved):
Seems like there should be a simpler way to do this, but I'm happy to worry about that later.

runtime/expr/agg/ztests/math-coercion.yaml
mattnibs: I get getting rid of the quotes on zed and output-flags but what does breaking the input into multiple lines give us?

mattnibs:
> Seems like there should be a simpler way to do this, but I'm happy to worry about that later.

🤷🏻‍♂️

runtime/expr/agg/ztests/math-coercion.yaml
nwt: Makes it clearer that there are two inputs, but no offense taken if you prefer it as is.

PR #3827, opened 2022-04-21
Have expr/function import expr

mattnibs:
Instead of having package expr import expr/function have function import
package expr. Move functions that formerly had to be in expr into package
function.


runtime/expr/function/nestdotted.go
```diff
@@ -1,37 +1,35 @@
...
+package function
 
 import (
 	"github.com/brimdata/zed"
 	"github.com/brimdata/zed/pkg/field"
 )
 
-type Unflattener struct {
+// https://github.com/brimdata/zed/blob/main/docs/language/functions.md#nest_dotted.md
+type NestDottedFields struct {
```
nwt, line 9: Nit: Call this NestDotted to match the Zed function name and the file name.

runtime/expr/function/has.go
```diff
@@ -0,0 +1,31 @@
...
+// https://github.com/brimdata/zed/blob/main/docs/language/functions.md#has
+type Has struct{}
+
+func (h *Has) Call(ctx zed.Allocator, args []zed.Value) *zed.Value {
+	for _, val := range args {
+		if val.IsError() {
+			if val.IsMissing() || val.IsQuiet() {
+				return zed.False
+			}
+			return &val
```
nwt, line 14:
```suggestion
			return ctx.CopyValue(val)
```

runtime/expr/function/has.go
```diff
@@ -0,0 +1,31 @@
...
+
+// https://github.com/brimdata/zed/blob/main/docs/language/functions.md#missing
+type Missing struct {
+	has Has
+}
+
+func (m *Missing) Call(ctx zed.Allocator, args []zed.Value) *zed.Value {
+	val := m.has.Call(ctx, args)
+	if val.Type == zed.TypeBool {
+		val = zed.Not(val.Bytes)
```
nwt, line 28:
```suggestion
		return zed.Not(val.Bytes)
```

nwt (approved):

PR #3832, opened 2022-04-22
Add compare function

mattnibs:
Add compare function and use it in delete-by-predicate. This fixes an
issue with delete-by-predicate where values are not deleted if they
can't be coerced with the comparison value.

### DIFF
--- a/docs/language/functions/README.md
+++ b/docs/language/functions/README.md
@@ -11,6 +11,7 @@ take Zed values as arguments and produce a value as a result.
 * [cast](cast.md) - coerce a value to a different type
 * [ceil](ceil.md) - ceiling of a number
 * [cidr_match](cidr_match.md) - test if IP is in a network
+* [compare](compare.md) - returns an int comparing two values
 * [crop](crop.md) - remove fields from a value that are missing in a specified type
 * [error](error.md) - wrap a value as an error
 * [every](every.md) - bucket `ts` using a duration
--- a/docs/language/functions/README.md
+++ b/docs/language/functions/README.md
@@ -11,7 +11,7 @@ take Zed values as arguments and produce a value as a result.
 * [cast](cast.md) - coerce a value to a different type
 * [ceil](ceil.md) - ceiling of a number
 * [cidr_match](cidr_match.md) - test if IP is in a network
-* [compare](compare.md) - returns an int comparing two values
+* [compare](compare.md) - return an int comparing two values
 * [crop](crop.md) - remove fields from a value that are missing in a specified type
 * [error](error.md) - wrap a value as an error
 * [every](every.md) - bucket `ts` using a duration
--- a/docs/language/functions/compare.md
+++ b/docs/language/functions/compare.md
@@ -12,7 +12,7 @@ compare(a: any, b: any) -> int64
 
 The _compare_ function returns an integer comparing two values. The result will
 be 0 if a is equal to b, +1 if a is greater than b, and -1 if a is less than b.
-_compare_ differs from <, >, <=, >=, == comparators in that it will
+_compare_ differs from `<`, `>`, `<=`, `>=`, `==`, and `!=` in that it will
 work for any type (e.g., `compare(1, "1")`). `null` values are the max value.
 
 ### Examples
--- a/docs/language/functions/compare.md
+++ b/docs/language/functions/compare.md
@@ -1,0 +1,26 @@
+### Function
+
+&emsp; **compare** &mdash; returns an integer comparing two values
+
+### Synopsis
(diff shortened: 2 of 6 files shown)


runtime/expr/function/compare.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/brimdata/zed/runtime/expr"
+)
+
+func NewCompare() *Compare {
+	return &Compare{
+		fn: expr.NewValueCompareFn(true),
+	}
+}
+
+type Compare struct {
```
nwt, line 14: 1. Include a doc comment.
2. Move this before the constructor.
```suggestion
// https://github.com/brimdata/zed/blob/main/docs/language/functions.md#compare
type Compare struct {
```

runtime/expr/function/compare.go
```diff
@@ -0,0 +1,28 @@
...
+	fn expr.CompareFn
```
nwt, line 15: We usually call `cmp` or `compare`.
```suggestion
	cmp expr.CompareFn
```

runtime/expr/function/compare.go
```diff
@@ -0,0 +1,28 @@
...
+func (e *Compare) Call(ctx zed.Allocator, args []zed.Value) *zed.Value {
+	lhs, rhs := &args[0], &args[1]
+	if lhs.IsError() || rhs.IsError() {
+		if lhs.IsError() {
+			return lhs
+		}
+		return rhs
+	}
+	cmp := e.fn(lhs, rhs)
+	return ctx.NewValue(zed.TypeInt64, zed.EncodeInt(int64(cmp)))
```
nwt, line 27: 1. I think we want to be able to compare error values.
2. Let `newInt64` allocate the `zcode.Bytes`. (It just calls `zed.EncodeInt` for now, but it'll be more efficient eventually.)
```suggestion
	return newInt64(ctx, int64(e.cmp(&args[0], &args[1]))
```

docs/language/functions/compare.md
```diff
@@ -0,0 +1,26 @@
...
+
+### Synopsis
+
+```
+compare(a: any, b: any) -> int64
+```
+
+### Description
+
+The _compare_ function returns an int comparing two values. The result with
```
nwt, line 13:
```suggestion
The _compare_ function returns an integer comparing two values. The result will
```

docs/language/functions/compare.md
```diff
@@ -0,0 +1,26 @@
+### Function
+
+&emsp; **compare** &mdash; returns an int comparing two values
```
nwt, line 3:
```suggestion
&emsp; **compare** &mdash; returns an integer comparing two values
```

docs/language/functions/README.md
```diff
@@ -11,6 +11,7 @@ take Zed values as arguments and produce a value as a result.
 * [cast](cast.md) - coerce a value to a different type
 * [ceil](ceil.md) - ceiling of a number
 * [cidr_match](cidr_match.md) - test if IP is in a network
+* [compare](compare.md) - returns an int comparing two values
```
nwt, line 14: Other entries begin with "return" instead of "returns".
```suggestion
* [compare](compare.md) - return an int comparing two values
```

docs/language/functions/compare.md
```diff
@@ -0,0 +1,26 @@
+### Function
+
+&emsp; **compare** &mdash; returns an integer comparing two values
```
nwt, line 3:
```suggestion
&emsp; **compare** &mdash; return an integer comparing two values
```

docs/language/functions/compare.md
```diff
@@ -0,0 +1,26 @@
...
+
+```
+compare(a: any, b: any) -> int64
+```
+
+### Description
+
+The _compare_ function returns an integer comparing two values. The result will
+be 0 if a is equal to b, +1 if a is greater than b, and -1 if a is less than b.
+_compare_ differs from <, >, <=, >=, == comparators in that it will
```
nwt, line 15:
```suggestion
_compare_ differs from `<`, `>`, `<=`, `>=`, `==`, and `!=` in that it will
```

docs/language/functions/compare.md
```diff
@@ -0,0 +1,26 @@
...
+work for any type (e.g., `compare(1, "1")`). `null` values are the max value.
```
nwt, line 16: "Works for any type" is a little vague, but we can fix that later.
```suggestion
work for any type (e.g., `compare(1, "1")`). `null` values compare greater than non-`null` values.
```

nwt (approved):

PR #3834, opened 2022-04-25
Improvements to join docs

philrz:
Having just looked closer at these while working on brimdata/brim#2331, here's some suggested improvements I found along the way.

### DIFF
--- a/docs/language/operators/join.md
+++ b/docs/language/operators/join.md
@@ -14,7 +14,7 @@ The `join` operator combines records from two inputs based on whether
 the `<left-key>` expression (evaluated in the context of the left input)
 is equal to the `<right-key>` expression (evaluated in the context of
 the right input) omitting values where there is no match (or including them
-int the case of anti join).
+in the case of anti join).
 
 The available join types are:
 * _inner_ - output only values that match
@@ -31,4 +31,5 @@ For anti join, the `<right-expr>` is undefined and thus cannot be specified.
 
 ### Examples
 
-TBD
+The [Join Overview](../../../tutorials/join.md) tutorial includes several
+examples.
--- a/docs/tutorials/join.md
+++ b/docs/tutorials/join.md
@@ -7,12 +7,12 @@ sidebar_label: Join
 
 ---
 
-This is a brief primer on Zed's experimental [join operator](../language/operators/join.md).
+This is a brief primer on Zed's experimental [`join` operator](../language/operators/join.md).
 
-Currently, join is limited in the following ways:
+Currently, `join` is limited in the following ways:
 * the joined inputs both come from the parent so the query must be split before join,
-* only merge join is implemented requiring inputs to be explicitly sorted, and
-* only equ-join is supported.
+* only merge join is implemented, requiring inputs to be explicitly sorted, and
+* only equi-join is supported.
 
 A more comprehensive join design with easier-to-use syntax is forthcoming.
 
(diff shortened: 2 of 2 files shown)


docs/tutorials/join.md
```diff
@@ -7,12 +7,12 @@ sidebar_label: Join
...
-This is a brief primer on Zed's experimental [join operator](../language/operators/join.md).
+This is a brief primer on Zed's experimental [`join` operator](../language/operators/join.md).
 
-Currently, join is limited in the following ways:
+Currently, `join` is limited in the following ways:
 * the joined inputs both come from the parent so the query must be split before join,
-* only merge join is implemented requiring inputs to be explicitly sorted, and
-* only equ-join is supported.
+* only merge join is implemented, requiring inputs to be explicitly sorted, and
+* only equi-join is supported.
```
philrz, line 15: In addition to fixing this spelling, I'd normally be inclined to suggest a hyperlink here, perhaps to something like [https://en.wikipedia.org/wiki/Join_(SQL)#Equi-join.](https://en.wikipedia.org/wiki/Join_(SQL)#Equi-join.) However, I've held back from that because in recent chats with @nwt I've deduced that I may be overzealous with suggesting hyperlinks in a way that runs somewhat counter to the preferences of the core Brim Dev team. Instead I'll just state my case and see if it persuades anyone.

For something like join, it seems great if we can assume the audience is made up of SQL jockeys, since they'll quickly look at these docs through the lens of "What's the bare minimum of what I need to know in Zed to replicate what I already know how to do in SQL?" However, I suspect there's a subset of users (maybe the search-centric ones, `jq` crossovers, and/or hold-overs from the days of Brim's network security focus) who have minimal/zero SQL expertise, and I still think it's worth helping them along the "gentle slope". In this regard, I often prefer hyperlinking potentially unfamiliar terms to the sources where we'd want them to learn the concepts rather than rely on the "Well, anyone can just Google it" and risk that they end up Googling to a suboptimal resource. Another term that's touched in in these join docs for which I also didn't suggest a hyperlink (anti-join) the exposure seems a little more significant because there's the SQL context as well as the relational algebra concept that both show up in search results, and the two have a lot in common but aren't 1-for-1 swaps for each other, so it seems all the more helpful to push the user to a solid resource.

In conclusion, I always get nervous of giving the impression of "You must be this SQL/math/whatever to ride this ride" and I feel like liberal use of hyperlinks helps avoid this. If I've persuaded anyone, I think picking out good hyperlinks for equi-join and anti-join would be a solid improvement, but since join is a bit of a power feature & described as experimental regardless, it's surely also not a crisis if left as is.

docs/language/operators/join.md
```diff
@@ -31,4 +31,5 @@ For anti join, the `<right-expr>` is undefined and thus cannot be specified.
 
 ### Examples
 
-TBD
+The [Join Overview](../../tutorials/join.md) tutorial includes several
+examples.
```
nwt, line 35:
```suggestion
The [join tutorial](../../tutorials/join.md) includes several examples.
```

docs/tutorials/join.md
nwt: How about just defining the term here?
```suggestion
* only equi-join (i.e., a join predicate containing `=`) is supported.

nwt (approved):

docs/tutorials/join.md
philrz: @nwt: Yes, works for me!

PR #3837, opened 2022-04-25
Remove horizontal rules from docs

jameskerr:
Many pages in the docs site have a horizontal rule under the heading. This breaks the automatic description tag used by the social networks. I don't think they are necessary as a visual style and removing them makes our twitter descriptions work automatically.

This is how it used to look:

image

We want those "---" under the title to actually be a preview of the content.

If we want to control the exact description of a page, add a 'description' key in the front-matter.

image

### DIFF
--- a/docs/commands/README.md
+++ b/docs/commands/README.md
@@ -5,8 +5,6 @@ sidebar_label: Commands
 
 # Command Tooling
 
----
-
 The Zed system is managed and queried with the [`zed` command](zed.md),
 which is organized into numerous subcommands like the familiar command patterns
 of `docker` or `kubectrl`.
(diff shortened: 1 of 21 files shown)


docs/commands/README.md
```diff
@@ -20,6 +18,6 @@ Zed data model, and interoperates with a number of other formats beyond JSON.
 If you don't need a Zed lake, you can install just the
 slimmer `zq` command which omits lake support and dev tools.
 
-`zq` is always installed alongside `zed`.  You might find yourself mixing and
+`zq` is always installed alongside `zed`. You might find yourself mixing and
```
nwt, line 21: We use two spaces after a period in the other Markdown files here.
```suggestion
`zq` is always installed alongside `zed`.  You might find yourself mixing and
```

docs/commands/zq.md
```diff
@@ -1,6 +1,7 @@
 ---
 sidebar_position: 0
 sidebar_label: zq
+description: A command-line tool that uses the Zed Language for pipeline-style search and analytics.
```
nwt, line 4: Did you mean to add a description to just this file?

jameskerr: Yes, since the first line of the content starts with "tldr". Doesn't seem too classy to include that in a twitter card link.

PR #3842, opened 2022-04-26
Update join.md

henridf:
fix typo

PR #3846, opened 2022-04-26
Revert "fix typo in docs/language/operators/join.md (#3842)"

philrz:
This reverts #3842 .

The mdtest-protected examples in the [join tutorial](https://zed.brimdata.io/docs/tutorials/join/) all confirm that `:=` is indeed the correct syntax for field assignment with `join`, so there wasn't a typo here.

cc: @henridf

### DIFF
--- a/docs/language/operators/join.md
+++ b/docs/language/operators/join.md
@@ -6,7 +6,7 @@
 
 ```
 ( => left path => right path )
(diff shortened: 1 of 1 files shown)


nwt:
🤦

nwt (approved):

PR #3852, opened 2022-04-27
zed: Move common flags to root

### LINKED ISSUE #3843: zed -lake and -use flags should be allowed before subcommand
`zed` exits with an error if the `-lake` flag precedes the subcommand. The same is true for `-use`.
```
$ zed -lake mylake ls
flag provided but not defined: -lake
at flag: "-lake mylake ls": flag provided but not defined: -lake
```

These flags, and perhaps all flags, should be allowed before the subcommand.

mattnibs:
Attach flags -q, -lake, -configdir, and -use to the zed root command.
This enables use to specify these flags before the subcommand:
zed -lake $HOME/.lake ls

Closes #3843


cli/cli.go
```diff
@@ -41,15 +45,30 @@ func (f *Flags) Init(all ...Initializer) error {
...
+
+type interruptedContext struct{ context.Context }
+
+func (s *interruptedContext) Err() error {
+	err := s.Context.Err()
+	if errors.Is(err, context.Canceled) {
+		return errors.New("interrupted")
+	}
+	return err
 }
```
nwt, line 69: I'm surprised to see this here. Does it need to move? If not, I think it should stay where it is.

mattnibs: Here's the story: I moved lake flags into cmd/zed/root command, zq uses this command and so now zq had the lake flags. zq only uses the root command because it has the convenient context/signal init stuff. It seem sensible to put this in cli.Flags since all our commands use this and also need ctx. I could move this to another struct, maybe something like cli.Base, and maybe that is the right thing to do, but I wanted to avoid adding another thing.

cli/cli.go
```diff
@@ -29,7 +33,7 @@ type Initializer interface {
 	Init() error
 }
 
-func (f *Flags) Init(all ...Initializer) error {
+func (f *Flags) Init(all ...Initializer) (context.Context, func(), error) {
```
nwt, line 36:
```suggestion
func (f *Flags) Init(all ...Initializer) (context.Context, context.CancelFunc, error) {
```

cli/cli.go
```diff
@@ -41,15 +45,30 @@ func (f *Flags) Init(all ...Initializer) error {
...
+		cancel()
+		f.cleanup()
+	}
+	return &interruptedContext{ctx}, cleanup, nil
+}
+
+type interruptedContext struct{ context.Context }
+
+func (s *interruptedContext) Err() error {
+	err := s.Context.Err()
```
nwt, line 64:
```suggestion
func (i *interruptedContext) Err() error {
	err := i.Context.Err()
```

cli/cli.go
```diff
@@ -41,15 +45,30 @@ func (f *Flags) Init(all ...Initializer) error {
...
 	if err != nil {
-		return err
+		return nil, nil, err
 	}
 	if f.cpuprofile != "" {
 		f.runCPUProfile(f.cpuprofile)
 	}
-	return nil
+	ctx, cancel := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
+	var cleanup = func() {
```
nwt, line 54:
```suggestion
	cleanup := func() {
```

cli/lakeflags/flags.go
```diff
@@ -33,3 +63,48 @@ func (f *Flags) HEAD() (*lakeparse.Commitish, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		return api.NewRemoteLake(conn), nil
+	}
+	return api.OpenLocalLake(ctx, uri.String())
+}
+
+func (l *Flags) AuthStore() *auth0.Store {
+	return auth0.NewStore(filepath.Join(l.ConfigDir, credsFileName))
```
nwt, line 98: Nit: Get rid of credsFileName.
```suggestion
	return auth0.NewStore(filepath.Join(l.ConfigDir, "credentials.json"))
```

cmd/zed/vacate/command.go
```diff
@@ -32,12 +31,10 @@ they cannot be recovered.
 
 type Command struct {
 	*root.Command
-	cli.LakeFlags
 }
 
 func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
 	c := &Command{Command: parent.(*root.Command)}
-	c.LakeFlags.SetFlags(f)
 	return c, nil
```
nwt, line 38:
```go
	return &Command{Command: parent.(*root.Command)}, nil
```

cli/zq/command.go
```diff
@@ -94,22 +93,19 @@ https://github.com/brimdata/zed for more information.
 }
 
 type Command struct {
-	*root.Command
 	canon       bool
 	quiet       bool
 	stopErr     bool
+	cli         cli.Flags
```
nwt, line 99:
```suggestion
	cliFlags         cli.Flags
```

cmd/zed/auth/command.go
```diff
@@ -27,14 +25,10 @@ func init() {
...
 	*root.Command
-	cli.LakeFlags
-	lakeFlags lakeflags.Flags
 }
 
 func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
 	c := &Command{Command: parent.(*root.Command)}
-	c.LakeFlags.SetFlags(f)
-	c.lakeFlags.SetFlags(f)
 	return c, nil
```
nwt, line 32:
```go
	return &Command{Command: parent.(*root.Command)}, nil
```

cmd/zed/index/command.go
```diff
@@ -49,14 +47,10 @@ func init() {
...
 	*root.Command
-	cli.LakeFlags
-	lakeFlags lakeflags.Flags
 }
 
 func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
 	c := &Command{Command: parent.(*root.Command)}
-	c.LakeFlags.SetFlags(f)
-	c.lakeFlags.SetFlags(f)
 	return c, nil
```
nwt, line 54:
```go
	return &Command{Command: parent.(*root.Command)}, nil
```

cmd/zed/init/command.go
```diff
@@ -24,14 +22,10 @@ var Cmd = &charm.Spec{
...
 	*root.Command
-	cli.LakeFlags
-	lakeFlags lakeflags.Flags
 }
 
 func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
 	c := &Command{Command: parent.(*root.Command)}
-	c.LakeFlags.SetFlags(f)
-	c.lakeFlags.SetFlags(f)
 	return c, nil
```
nwt, line 29:
```go
	return &Command{Command: parent.(*root.Command)}, nil
```

cmd/zed/root/command.go
```diff
@@ -23,42 +21,27 @@ querying, and orchestrating Zed data lakes.`,
 
 type Command struct {
 	charm.Command
+	lakeflags.Flags
```
nwt, line 24: We don't usually embed flag structs, so I don't think we should embed this one.
```suggestion
	LakeFlags lakeflags.Flags
```

cmd/zed/use/command.go
```diff
@@ -60,16 +59,12 @@ in the lake and is not copied to this local directory.
...
-	lakeFlags lakeflags.Flags
+	branch   bool
+	poolName string
 }
 
 func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
 	c := &Command{Command: parent.(*root.Command)}
-	c.LakeFlags.SetFlags(f)
-	c.lakeFlags.SetFlags(f)
 	return c, nil
```
nwt, line 68:
```go
	return &Command{Command: parent.(*root.Command)}, nil
```

cmd/zed/use/command.go
```diff
@@ -60,16 +59,12 @@ in the lake and is not copied to this local directory.
 
 type Command struct {
 	*root.Command
-	cli.LakeFlags
-	branch    bool
-	poolName  string
-	lakeFlags lakeflags.Flags
+	branch   bool
+	poolName string
```
nwt, line 63: These are unused.
```suggestion
```

PR #3854, opened 2022-04-27
Fixes to ZST format doc

philrz:
From a proofreading pass.

I'm including comments in a couple places where I want to make extra sure that my clarifications to the wording didn't affect the correctness of the message.

### DIFF
--- a/docs/formats/zst.md
+++ b/docs/formats/zst.md
@@ -257,7 +257,7 @@ when there are null values (as described below).
 
 If there are no null values, then the `presence` field contains an empty `<segmap>`.
 If all of the values are null, then the `column` field is null (and the `presence`
-both contain empty `<segmap>s`).  For empty `<segmap>s`, there is no
+contains an empty `<segmap>`).  For an empty `<segmap>`, there is no
 corresponding data stored in the data section.  Since a `<segmap>` is a Zed
 array, an empty `<segmap>` is simply the empty array value `[]`.
 
--- a/docs/formats/zst.md
+++ b/docs/formats/zst.md
@@ -134,8 +134,8 @@ from column streams, i.e., to map columns back to composite values.
 >
 > Also, the reassembly section is in general vastly smaller than the data section
 > so the goal here isn't to express information in cute and obscure compact forms
-> but rather to represent data in easy-to-digest, programmer-friendly forms that
-> leverage ZNG.
+> but rather to represent data in an easy-to-digest, programmer-friendly form that
+> leverages ZNG.
 
 The reassembly section is a ZNG stream.  Unlike Parquet,
 which uses an externally described schema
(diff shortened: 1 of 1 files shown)


docs/formats/zst.md
```diff
@@ -213,7 +213,7 @@ In other words, the reassembly record of the super column
 combined with the N reassembly records collectively define the original sequence
 of Zed data values in the original order.
 Taken in pieces, the reassembly records allow efficient access to sub-ranges of the
-rows, to subset of columns of the rows, to sub-ranges of columns of the rows, and so forth.
+rows, to subsets of columns of the rows, to sub-ranges of columns of the rows, and so forth.
```
philrz, line 216: Maybe this could also be "...to a subset of columns..." but based on everything else in this paragraph, I assumed plural was the intent.

docs/formats/zst.md
```diff
@@ -270,11 +270,21 @@ An `<array_column>` has the form:
...
+#### Map Column
+
+A `<map_column>` has the form:
+```
+{key:<any_column>,value:<any_column>}
+```
+where
+* `key` encodes the column of map keys, and
+* `value` encodes the column of map values.
+
```
philrz, line 287: I moved this section above the one for "union" because @nwt conditioned me to make Zed-related complex type docs to follow the order in [https://github.com/brimdata/zed/blob/main/type.go](https://github.com/brimdata/zed/blob/main/type.go).

docs/formats/zst.md
```diff
@@ -365,7 +365,7 @@ column values from.
 The top-level reassembly fetches column values as a `<record_column>`.
 
 For any `<record_column>`, a value from each field is read from each field's column,
-accounting for the presence column indicating nil,
+accounting for the presence column indicating null,
```
philrz, line 368: The word `nil` never appears anywhere else in this doc and `null` is referenced several times when speaking about the role of the presence column, so I assume that's what's intended here.

docs/formats/zst.md
```diff
@@ -251,13 +251,13 @@ where
...
-run-length encoding the locations of column values in their respective rows,
+* the `presence` columns are `int32` ZNG column streams comprised of a
+run-length encoding of the locations of column values in their respective rows,
 when there are null values (as described below).
 
 If there are no null values, then the `presence` field contains an empty `<segmap>`.
-If all of the values are null, then then `column` field is null (and the `presence`
-both contain empty `<segmap>'s`).  For empty `<segmap>'s`, there is no
+If all of the values are null, then the `column` field is null (and the `presence`
+both contain empty `<segmap>s`).  For empty `<segmap>s`, there is no
```
nwt, line 260:
```suggestion
contains an empty `<segmap>`).  For an empty `<segmap>`, there is no
```

docs/formats/zst.md
```diff
@@ -270,11 +270,21 @@ An `<array_column>` has the form:
...
 The `<array_column>` structure is used for both Zed arrays and sets.
 
+#### Map Column
+
+A `<map_column>` has the form:
+```
+{key:<any_column>,value:<any_column>}
+```
+where
+* `key` encodes the column of map keys, and
```
nwt, line 285: Nit: The original text contains a serial comma, but since there are only two list items, I'd lose it.
```suggestion
* `key` encodes the column of map keys and
```

nwt (approved):

PR #3858, opened 2022-04-28
add assert operator

### LINKED ISSUE #3853: Feature idea: assert keyword
I came across Zed and really like it. It'd be a great tool for validating data, but for this it'd be nice if it has an "assert" operator defined as

```
switch(
  case not CONDITION => yield error( "assertion error: CONDITION not met for ..." ) 
  default => yield this
)
```

In such a case, it would be possible to write data assertion pipes like

```
assert Age > 0 |
assert Age < 120 |
assert Height > 0 |
...
```

In case any assertion is not met it throws an error like

```
error("Assertion error: Age > 0 not met for ...")
```

nwt:
"assert EXPR" yields the input value if the Boolean expression EXPR
evaluates to true or a structured error if it does not.  The error
structure follows that used in zed.Context.WrapError.

Closes #3853.

### DIFF
--- a/compiler/parser/parser.es.js
+++ b/compiler/parser/parser.es.js
@@ -324,10 +324,48 @@ function peg$parse(input, options) {
             }
             return result
           },
-      peg$c114 = "sort",
-      peg$c115 = peg$literalExpectation("sort", false),
-      peg$c116 = function(args, l) { return l },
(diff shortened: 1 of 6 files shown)


mccanne (approved):
Nice work!

PR #3861, opened 2022-04-29
change zed.Value parameters to pointers in zson

nwt:
At five words, a zed.Value isn't tiny, so avoid unnecessary copies.

Depends on #3860.

### DIFF
--- a/api/client/connection.go
+++ b/api/client/connection.go
@@ -136,7 +136,7 @@ func (c *Connection) doAndUnmarshal(req *Request, v interface{}, templates ...in
 	}
 	m := zson.NewZNGUnmarshaler()
 	m.Bind(templates...)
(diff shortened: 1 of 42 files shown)


mattnibs (commented):
Considering the zson package is in the slow path, I question the perf savings here. Is the objective to have zed.Value pointers in method calls by default and deviating where it makes sense?

nwt:
> Considering the zson package is in the slow path, I question the perf savings here. Is the objective to have zed.Value pointers in method calls by default and deviating where it makes sense?

You're mostly right about the questionable performance savings here. I suppose my real goal here is a consistent call-by-reference pattern for zed.Value parameters.

mattnibs (approved):

PR #3866, opened 2022-05-01
Fix incorrect collect vs union usage

iloveitaly:


nwt:
Thank you, @iloveitaly!

nwt (approved):

PR #3867, opened 2022-05-01
Add static array containment example

iloveitaly:


nwt (approved):

docs/language/overview.md
```diff
@@ -847,6 +847,15 @@ produces
 {a:[1,2]}
 {d:{e:1}}
 ```
+You can also use this operator with a static array:
+```mdtest-command
+echo '{accounts:[{id:1},{id:2},{id:3}]}' | zq -z 'over accounts | where id in [1,2]' -
```
nwt, line 852: @iloveitaly: Maybe simplify this a little?
```suggestion
echo '{id:1} {id:2} {id:3}' | zq -z 'where id in [1,2]' -
```

iloveitaly: @nwt I actually like my example better: it demonstrates how to extract data from a JSON blob using `zq`. I wish there were more examples like this in the docs.

Happy to use yours if you'd like, just want to give the feedback that more JSON-based examples would be helpful for newcomers.

PR #3870, opened 2022-05-02
service: Improve accept header handling

### LINKED ISSUE #3859: API query fails with list of Accept values
Repro is with Zed commit 8ae1a73e.

I'm trying to query the Zed API through a client that adds its own HTTP Accept header. I'm starting from the following successful query that's the same example one shown at https://zed.brimdata.io/docs/lake/api/#query but with the `Accept` header changed to `application/json` which is what the client ultimately needs.

Request;
```
curl -X POST \
     -H 'Content-Type: application/json' \
     -H 'Accept: application/json' \
     http://localhost:9867/query -d '{"query":"from inventory@main | count() by warehouse"}'
```

Response:
```
[{"warehouse":"miami","count":1},{"warehouse":"chicago","count":2}]
```

However, if I change the `Accept` header to the same list that's generated by my client, now it fails:

Request:
```
curl -X POST \
     -H 'Content-Type: application/json' \
     -H 'Accept: application/json, text/plain, */*' \
     http://localhost:9867/query -d '{"query":"from inventory@main | count() by warehouse"}'
```

Response:
```
{"type":"Error","kind":"invalid operation","error":"mime: unexpected content after media subtype"}
```

I was hoping it would have just recognized `application/json` as the first in the list and used that.

mattnibs:
Import handling of accept headers so that HTTP Accept headers with multiple
types are handled without error. This pr also improves and centralizes
how the derived format type from the selected Accept mime type is used
in constructing the response.

Closes #3859

### DIFF
--- a/api/mime.go
+++ b/api/mime.go
@@ -69,6 +69,6 @@ func FormatToMediaType(format string) string {
 	case "zson":
 		return MediaTypeZSON
 	default:
-		return ""
+		panic(fmt.Sprintf("unknown format type: %s", format))
 	}
 }
--- a/api/mime.go
+++ b/api/mime.go
@@ -53,3 +53,22 @@ func MediaTypeToFormat(s string, dflt string) (string, error) {
 	}
 	return "", &ErrUnsupportedMimeType{typ}
 }
+
+func FormatToMediaType(format string) string {
+	switch format {
+	case "csv":
+		return MediaTypeCSV
+	case "json":
+		return MediaTypeJSON
+	case "ndjson":
+		return MediaTypeNDJSON
+	case "zjson":
+		return MediaTypeZJSON
+	case "zng":
+		return MediaTypeZNG
+	case "zson":
+		return MediaTypeZSON
+	default:
+		return ""
+	}
+}
--- a/service/request.go
+++ b/service/request.go
@@ -49,7 +49,7 @@ func newRequest(w http.ResponseWriter, r *http.Request, logger *zap.Logger, defa
 	if len(ss) == 0 {
 		ss = []string{""}
 	}
-	for _, mime := range strings.Split(r.Header.Get("Accept"), ",") {
+	for _, mime := range ss {
 		format, err := api.MediaTypeToFormat(mime, defaultFormat)
 		if err != nil {
 			continue
--- a/service/request.go
+++ b/service/request.go
@@ -32,7 +32,7 @@ type Request struct {
 	Logger *zap.Logger
 }
 
-func newRequest(w http.ResponseWriter, r *http.Request, logger *zap.Logger, defaultFormat string) (*ResponseWriter, *Request, bool) {
+func newRequest(w http.ResponseWriter, r *http.Request, logger *zap.Logger) (*ResponseWriter, *Request, bool) {
 	logger = logger.With(zap.String("request_id", api.RequestIDFromContext(r.Context())))
 	req := &Request{
 		Request: r,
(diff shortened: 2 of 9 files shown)


philrz:
I just cherry-picked the commit from this branch into my Grafana test environment and indeed it got past the problem that was blocking me, so, 👍  from me!

service/ztests/accept-header.yaml
```diff
@@ -0,0 +1,22 @@
+script: |
+  source service.sh
+  zed create -q test
+  echo '{ts:0}' | zed load -use test -q -
+  for accept in "text/plain, application/json" "application/xml, text/css"; do
+    echo === $accept ===
```
nwt, line 6: If you use `//`, the output will be valid ZSON, so `runParquetBoomerangs` and `runZSONBoomerangs` in `zed_test.go` will be able to understand it.
```suggestion
    echo // $accept
```

service/ztests/accept-header.yaml
```diff
@@ -0,0 +1,22 @@
...
+    curl -X POST \
+      -H "Accept: $accept" \
+      -d '{"query":"from test@main"}' \
+      $ZED_LAKE/query
```
nwt, line 10: `curl -d` implies `-X POST`.
```suggestion
    curl -H "Accept: $accept" -d '{"query":"from test@main"}' $ZED_LAKE/query
```

service/request.go
```diff
@@ -44,12 +45,20 @@ func newRequest(w http.ResponseWriter, r *http.Request, logger *zap.Logger) (*Re
...
-	return res, req
+	for _, mime := range strings.Split(r.Header.Get("Accept"), ",") {
+		format, err := api.MediaTypeToFormat(mime, defaultFormat)
+		if err != nil {
+			continue
+		}
+		res.Format = format
+		return res, req, true
+	}
+	res.Error(srverr.ErrInvalid("could not find supported mime type in ACCEPT header"))
```
nwt, line 60:
```suggestion
	res.Error(srverr.ErrInvalid("could not find supported MIME type in Accept header"))
```

api/mime.go
```diff
@@ -53,3 +53,22 @@ func MediaTypeToFormat(s string, dflt string) (string, error) {
...
+	case "ndjson":
+		return MediaTypeNDJSON
+	case "zjson":
+		return MediaTypeZJSON
+	case "zng":
+		return MediaTypeZNG
+	case "zson":
+		return MediaTypeZSON
+	default:
+		return ""
```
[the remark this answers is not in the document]
mattnibs: Agreed, let's go with panic since this should never happen.

service/request.go
```diff
@@ -44,12 +45,20 @@ func newRequest(w http.ResponseWriter, r *http.Request, logger *zap.Logger) (*Re
 		Logger:         logger,
 		marshaler:      m,
 	}
-	accept := r.Header.Get("Accept")
-	if accept == "" || accept == "*/*" {
-		accept = "application/json"
+	ss := strings.Split(r.Header.Get("Accept"), ",")
+	if len(ss) == 0 {
+		ss = []string{""}
 	}
```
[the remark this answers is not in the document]
mattnibs: doh

service/ztests/accept-header.yaml
mattnibs: The more you know

service/request.go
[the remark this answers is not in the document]
mattnibs: This is there to handle the empty Accept header case and to set an appropriate default. Will fix.

nwt (approved):

PR #3879, opened 2022-05-10
Renable filter tests

mattnibs:
Enabling the filter tests also surfaced a bug in BufferFilter where
lengths of record values where getting improperly decoded.

### DIFF
--- a/zcode/bytes.go
+++ b/zcode/bytes.go
@@ -88,7 +88,7 @@ func DecodeTagLength(b Bytes) int {
 	if tagIsNull(u64) {
 		return n
 	}
-	return int(u64) + n - 1
+	return int(u64) + n
 }
 
 func toTag(length int) uint64 {
--- a/zcode/bytes.go
+++ b/zcode/bytes.go
@@ -88,7 +88,7 @@ func DecodeTagLength(b Bytes) int {
 	if tagIsNull(u64) {
 		return n
 	}
-	return int(u64) + n
+	return int(u64) + n - 1
 }
 
 func toTag(length int) uint64 {
--- a/runtime/expr/filter_test.go
+++ b/runtime/expr/filter_test.go
@@ -77,18 +77,14 @@ func runCasesHelper(t *testing.T, record string, cases []testcase, expectBufferF
 			bf, err := filterMaker.AsBufferFilter()
 			assert.NoError(t, err, "filter: %q", c.filter)
 			if bf != nil {
-				expected := c.expected
-				if expectBufferFilterFalsePositives {
-					expected = true
-				}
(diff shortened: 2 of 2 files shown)


zcode/bytes.go
```diff
@@ -88,7 +88,7 @@ func DecodeTagLength(b Bytes) int {
 	if tagIsNull(u64) {
 		return n
 	}
-	return int(u64) + n - 1
+	return int(u64) + n
```
nwt, line 91: I don't think this change is correct. The return value here is meant to be the length of the tag (which is `n`) plus the length of the value (which is `u64 - 1` at this point).

There's clearly a bug somewhere though since the tests panic without this. I'll take a look.

nwt: The bug's in `runtime/expr/filter_test.go`, which didn't get the necessary update in #3301. Here's what I think we want.
```diff
diff --git a/runtime/expr/filter_test.go b/runtime/expr/filter_test.go
index 0a4f062b..60ca26b0 100644
--- a/runtime/expr/filter_test.go
+++ b/runtime/expr/filter_test.go
@@ -81,13 +81,11 @@ func runCasesHelper(t *testing.T, record string, cases []testcase, expectBufferF
                                if expectBufferFilterFalsePositives {
                                        expected = true
                                }
-                               // For fieldNameFinder.find coverage, we need to
-                               // hand BufferFilter.Eval a buffer containing a
-                               // ZNG value message for rec, assembled here.
-                               require.Less(t, rec.Type.ID(), 0x40)
-                               buf := []byte{byte(rec.Type.ID())}
-                               buf = zcode.AppendUvarint(buf, uint64(len(rec.Bytes)))
-                               buf = append(buf, rec.Bytes...)
+                               // For FieldNameFinder.Find coverage, we need to
+                               // hand BufferFilter.Eval a ZNG values frame
+                               // containing rec, assembled here.
+                               buf := zcode.AppendUvarint(nil, uint64(rec.Type.ID()))
+                               buf = zcode.Append(buf, rec.Bytes)
                                assert.Equal(t, expected, bf.Eval(zctx, buf),
                                        "filter: %q\nbuffer:\n%s", c.filter, hex.Dump(buf))
                        }
diff --git a/zcode/bytes.go b/zcode/bytes.go
index 8d71c23b..4dc871e7 100644
--- a/zcode/bytes.go
+++ b/zcode/bytes.go
@@ -88,7 +88,7 @@ func DecodeTagLength(b Bytes) int {
        if tagIsNull(u64) {
                return n
        }
-       return int(u64) + n
+       return int(u64) + n - 1
 }
```

mattnibs: Perfect, thanks!

nwt (approved):

PR #3880, opened 2022-05-10
Simplify pushdown

mattnibs:
Simplify pushdown by making the filter dag reachable in kernel.Filter
and moving full compilation of index into index package. Remove the index
filter dag argument from DataAdaptor.NewScheduler


compiler/ztests/pushdown-index.yaml
```diff
@@ -1,47 +0,0 @@
-script: |
-  zc -C -O "from 'pool-ts'| x=='hello' or x==1.0"
-  echo ===
-  zc -C -O "from 'pool-ts'| x > 1 y <= 1.0"
-  echo ===
-  zc -C -O "from 'pool-ts'| x=='hello' or x!=1.0"
-  echo ===
-  zc -C -O "from 'pool-ts'| x=='hello' or !(y==2 or y==3)"
```
nwt, line 8: I think this is our only coverage of `zc -C -O` output that includes pushdowns, so maybe keep it (with the necessary changes to the output) and rename to `pushdown.yaml`?

lake/index/expr.go
```diff
@@ -21,39 +20,40 @@ type result struct {
 	span extent.Span
 }
 
-func compileExpr(node dag.Expr) (expr, error) {
+// compileExpr returns a watered-down version of a filter that can be digested
+// by index. All parts of the expression tree are removed that are not:
+// - A '=', '>', '>=', '<', '<=', 'and' or 'or' BinaryExpr
```
nwt, line 25:
```suggestion
// - A '==', '>', '>=', '<', '<=', 'and' or 'or' BinaryExpr
```

nwt (approved):
Nice!

Issue #3881, opened 2022-05-10
zq command fails with "zngio: uncompressed length exceeds MaxSize" for large files


muthu-rk:
Hello,

I am trying to query a 100MB JSON file converted to ZNG format. However the query fails with **"zngio: uncompressed length exceeds MaxSize"** error. I tried setting -readsize and -readmax to higher values (1GiB). The error still persist.

However querying the JSON file directly instead of ZNG format, gave expected results. But it took more time to complete. I am trying to optimize it with ZNG format.

Also, the same query worked well for smaller files (about 10MB JSON)

Errors:
`
$ zq -z -i zng 'over arr | where id==10001 | cut level0.level1' 100mb.zng 
100mb.zng: zngio: uncompressed length exceeds MaxSize

$ zq -z -i zng -readmax 1GiB -readsize 1GiB 'over arr | where id==10001 | cut level0.level1' 100mb.zng 
100mb.zng: zngio: uncompressed length exceeds MaxSize
`
How files were converted from JSON to ZNG format:

`
$ zq -f zng -i json 100mb.json > 100mb.zng

$ ll -h
total 1.7G
drwxrwxr-x 2 user user  4.0K May 11 02:09 ./
drwxrwxr-x 9 user user  4.0K May  6 10:52 ../
-rw-rw-r-- 1 user user  102M May 11 01:25 100mb.json
-rw-rw-r-- 1 user user  728K May 11 01:26 100mb.zng
-rw-rw-r-- 1 user user   11M May 11 01:24 10mb.json
-rw-rw-r-- 1 user user   60K May 11 01:26 10mb.zng

`
Can you suggest how to solve this problem?

mccanne:
Hmm, look like a bug.  We'll get to the bottom of it.

muthu-rk:
Thanks @mccanne for quick response.

We have use case to perform CRUD on 1GiB size JSON file. It equivalent ZNG file comes to about 8.7 MB size.

I hope there wont be any file size restriction with the fix, given that querying a json file with input of 1GiB instead of equivalent ZNG input is already works as expected.

nwt:
@muthu-rk: Is it possible for you to share that 100mb.zng file by [attaching it here](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/attaching-files)?

muthu-rk:
[100mb.zng.zip](https://github.com/brimdata/zed/files/8669438/100mb.zng.zip)
[...]

PR #3884, opened 2022-05-11
zio/zngio: increase MaxSize from 10 MB to 1 GB

nwt:
The MaxSize constant limits the size of ZNG frames (and, consequently,
values) read by zngio, providing a defense against bad inputs that might
otherwise trigger an arbitrarily-large memory allocation.  Experience
has shown that the current value of 10 MB is too small, so increase it
to 1 GB.

For #3881.

### DIFF
--- a/zio/zngio/reader.go
+++ b/zio/zngio/reader.go
@@ -14,7 +14,7 @@ import (
 
 const (
 	ReadSize  = 512 * 1024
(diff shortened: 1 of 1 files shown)


mccanne (approved):
We should also add a change so a user can override this max, correct?

nwt:
> We should also add a change so a user can override this max, correct?

Yes, working on that PR now!

PR #3896, opened 2022-05-13
compiler/parser: allow over expression without extra parentheses

nwt:
An over expression often must be wrapped in extra parentheses.  For
example, the parser accepts "collect((over a | b))" but not
"collect(over a | b)".  Fix that.


mccanne (approved):
cool

PR #3898, opened 2022-05-16
Add nullsMax argument to compare function

mattnibs:
Delete by predicate (along with future internal zed lake programs
that with use this function) should behave in a manner consistent with
how data is stored in the lake (i.e. the filter ts < 0 should delete records
where ts == null for descending ordered lakes). As such add an optional
nullsMax argument to the zed compare function so behavior can be changed
to match pool sort order.


docs/language/functions/compare.md
```diff
@@ -5,15 +5,18 @@
...
 
 ### Description
 
 The _compare_ function returns an integer comparing two values. The result will
 be 0 if a is equal to b, +1 if a is greater than b, and -1 if a is less than b.
 _compare_ differs from `<`, `>`, `<=`, `>=`, `==`, and `!=` in that it will
-work for any type (e.g., `compare(1, "1")`). `null` values compare greater than non-`null` values.
+work for any type (e.g., `compare(1, "1")`).
+
+`nullsMax` is an optional value (true by default) that sets whether `null`
```
nwt, line 18:
```suggestion
`nullsMax` is an optional value (true by default) that determines whether `null`
```

nwt (approved):

PR #3902, opened 2022-05-17
zfmt: add parentheses to binary expressions

mattnibs:
Add parentheses on binary expressions when requested by the parent
caller. This change fixes an issue where zfmt would lose needed order
of operations in certain circumstances.


mattnibs:
hmm... not working as expected. Closing to rework this.

PR #3905, opened 2022-05-19
docs: Language overview fixes

philrz:
Fixes of the usual variety: Links, formatting, typos, wording. I'll add inline comments where I want to draw particular attention.

### DIFF
--- a/docs/language/overview.md
+++ b/docs/language/overview.md
@@ -681,8 +681,9 @@ For example, suppose a bad value shows up:
 ```
 {kind:"bad", stuff:{foo:1,bar:2}}
 ```
-A Zed [shaper](#9-shaping) could catch the bad value (e.g., as a default case in a
-[`switch`](operators/switch.md) topology) and propagate it as an error using the Zed expression:
+A Zed [shaper](#9-shaping) could catch the bad value (e.g., as a default
+case in a [`switch`](operators/switch.md) topology) and propagate it as
+an error using the Zed expression:
 ```
 yield error({message:"unrecognized input",input:this})
 ```
--- a/docs/language/overview.md
+++ b/docs/language/overview.md
@@ -1078,7 +1078,7 @@ where a `<spec>` has one of three forms:
 The first form is a customary colon-separated field and value similar to JavaScript,
 where `<field>` may be an identifier or quoted string.
 The second form is an implied field reference `<ref>`, which is shorthand
-for `<ref>:<ref>`.  The third form is the `...` spread operator which expects
+for `<ref>:<ref>` (XXX).  The third form is the `...` spread operator which expects
 a record value as the result of `<expr>` and inserts all of the fields from
 the resulting record.
 If a spread expression results in a non-record type (e.g., errors), then that
--- a/docs/language/overview.md
+++ b/docs/language/overview.md
@@ -138,7 +138,7 @@ Each operator is identified by name and performs a specific operation
 on a stream of records.
 
 Some operators, like
-[`summarize`](operators/summarize.md) or [`sort`](operators/sort.md)
+[`summarize`](operators/summarize.md) or [`sort`](operators/sort.md),
 read all of their input before producing output, though
 `summarize` can produce incremental results when the group-by key is
 aligned with the order of the input.
--- a/docs/language/overview.md
+++ b/docs/language/overview.md
@@ -480,7 +480,7 @@ a few examples:
 * a union of string and integer  - `(string,int64)`
 
 Complex types may be composed, as in `[({s:string},{x:int64})]` which is
-an array of type union of two types of records.
+an array of type `union` of two types of records.
 
 The [`typeof` function](functions/typeof.md) returns a value's type as
 a value, e.g., `typeof(1)` is `<int64>` and `typeof(<int64>)` is `<type>`.
@@ -596,7 +596,7 @@ the scope of the Zed data model and language.  That said, Zed provides flexible
 building blocks so systems can define their own schema versioning and schema
 management policies on top of these Zed primitives.
 
-Zed's [super-structured data model](..//formats/README.md#2-zed-a-super-structured-pattern) is a superset of relational tables and
+Zed's [super-structured data model](../formats/README.md#2-zed-a-super-structured-pattern) is a superset of relational tables and
 the Zed language's type system can easily make this connection.
 As an example, consider this type definition for "employee":
 ```
@@ -1107,7 +1107,7 @@ Array literals have the form
 [ <expr>, <expr>, ... ]
 ```
 When the expressions result in values of non-uniform type, then the implied
-type of the array is an array of type union of the types that appear.
+type of the array is an array of type `union` of the types that appear.
 
 For example,
 ```mdtest-command
@@ -1126,7 +1126,7 @@ Set literals have the form
 |[ <expr>, <expr>, ... ]|
 ```
 When the expressions result in values of non-uniform type, then the implied
-type of the set is a set of type union of the types that appear.
+type of the set is a set of type `union` of the types that appear.
 
 Set values are always organized in their "natural order" independent of the order
 they appear in the set literal.
@@ -1178,7 +1178,7 @@ produces
 "foo"((int64,string))
 ```
 The value underlying a union-tagged value is accessed with the
-[under function](functions/under.md):
+[`under` function](functions/under.md):
 ```mdtest-command
 echo '1((int64,string))' | zq -z 'yield under(this)' -
 ```
@@ -1201,13 +1201,13 @@ produces
 
 ### 6.12 Casts
 
-Type conversion is performed with casts and the built-in function `cast()`.
+Type conversion is performed with casts and the built-in [`cast` function](functions/cast.md).
 
 Casts for primitive types have a function-style syntax of the form
 ```
 <type> ( <expr> )
 ```
-where `<type>` is a Zed type and `<expr>` is any Zed expression.
+where `<type>` is a [Zed type](#51-first-class-types) and `<expr>` is any Zed expression.
 In the case of primitive types, the type-value angle brackets
 may be omitted, e.g., `<string>(1)` is equivalent to `string(1)`.
 If the result of `<expr>` cannot be converted
@@ -1238,7 +1238,7 @@ produces
 1970-10-07T00:00:00Z
 ```
 
-Casts of complex or named types may be performed using type values
+Casts of complex or [named types](#52-named-types) may be performed using type values
 either in functional form or with `cast`:
 ```
 <type-value> ( <expr> )
@@ -1257,7 +1257,6 @@ produces
 Casts may be used with complex types as well.  As long as the target type can
 accommodate the value, the case will be recursively applied to the components
 of a nested value.  For example,
-For example
 ```mdtest-command
 echo '["10.0.0.1","10.0.0.2"]' | zq -z 'cast(this,<[ip]>)' -
 ```
@@ -1287,7 +1286,7 @@ long form built from the expression syntax above in combination with the
 ### 7.1 Search Patterns
 
 Several styles of string search can be performed with a search expression
-(as well as the [grep function](functions/grep.md)) using "patterns",
+(as well as the [`grep` function](functions/grep.md)) using "patterns",
 where a pattern is a regular expression, glob, or simple string.
 
 #### 7.1.1 Regular Expressions
@@ -1328,7 +1327,7 @@ false
 
 Globs provide a convenient short-hand for regular expressions and follow
 the familiar pattern of "file globbing" supported by Unix shells.
-Zed globs are a simple, special case utilize only the `*` wildcard.
+Zed globs are a simple, special case that utilize only the `*` wildcard.
 
 Valid glob characters include `a` through `z`, `A` through `Z`,
 any valid string escape sequence
@@ -1369,8 +1368,20 @@ produces
 {a:1}
 ```
 
+Globs may also appear in the `grep` function:
+```mdtest-command
+echo '"foo" {s:"bar"} {s:"baz"} {foo:1}' | zq -z 'yield grep(ba*, s)' -
+```
+produces
+```mdtest-output
+false
+true
+true
+false
+```
+
 Note that a glob may look like multiplication but context disambiguates
-these condition, e.g.,
+these conditions, e.g.,
 ```
 a*b
 ```
@@ -1383,7 +1394,7 @@ is a Boolean comparison between the product `a*b` and `c`.
 ### 7.2 Search Logic
 
 The search patterns described above can be combined with other elements
-to a search expression comprised of "search terms" that may be combined
+of a search expression comprised of "search terms" that may be combined
 using Boolean logic.
 
 > Note that when processing ZNG data, the Zed runtime performs a multi-threaded
@@ -1394,7 +1405,7 @@ using Boolean logic.
 > In a forthcoming release, Zed will also offer an approach for locating
 > delimited words within string fields, which will allow accelerated
 > search using a full-text search index.  Currently, search indexes may be built
-> for exact value match as text segmentation is in the works.a
+> for exact value match as text segmentation is in the works.
 
 #### 7.2.1 Search Terms
 
@@ -1409,7 +1420,7 @@ A "search term" is one of the following;
 
 A regular expression `/re/` is equivalent to
 ```
-grep(/reg/, this)
+grep(/re/, this)
 ```
 but shorter and easier to type in a search expression.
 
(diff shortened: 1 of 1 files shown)


docs/language/overview.md
```diff
@@ -105,14 +105,14 @@ All available operators are listed on the [reference page](operators/README.md).
 ### 2.1 Dataflow Sources
 
 In addition to the data sources specified as files on the `zq` command line,
-a source may also be specified with the [from operator](operators/from.md#operator).
+a source may also be specified with the [`from` operator](operators/from.md).
```
philrz, line 108: This was a convention I pre-sold with @nwt. In many places, this doc was already saying "\<some> operator", "\<some> function", etc., which I think does a nice job of helping the user start to understand the differences between operators, simple functions, and aggregate functions, and also disambiguating in cases where there's one of both flavors (e.g., `fuse`). However, we also agreed that it would be good practice to start putting the operator name in fixed width font, so I've started doing that here. The other thing I've adopted as a convention is dropping the extra `#operator` that was at the end, which is how the function links were already written.

docs/language/overview.md
```diff
@@ -482,16 +480,15 @@ a few examples:
...
 a value, e.g., `typeof(1)` is `<int64>` and `typeof(<int64>)` is `<type>`.
 
 First-class types are quite powerful because types can
-serve as group-by keys or be used in "data shaping" logic.
-A common query for data introspection is to perform some search query
-slicing and dicing some exploratory data then counting the shapes of
-each type of data as follows:
+serve as group-by keys or be used in ["data shaping"](#9-shaping) logic.
+A common workflow for data introspection is to first perform a search of
+exploratory data and then count the shapes of each type of data as follows:
```
philrz, line 491: The original statement felt a bit too wordy, so I've tried to simplify. Please confirm I have not wrecked the essence of its intended meaning, though.

docs/language/overview.md
```diff
@@ -965,13 +959,17 @@ produces
 -2
 ```
 
+Note that if the expression has side effects,
+as with [aggregate function calls](#610-aggregate-function-calls), only the selected expression
+will be evaluated.
```
philrz, line 964: I moved this "note" after the example because, as originally written, I'd initially expected the example to show something related to what was just described in the note.

docs/language/overview.md
```diff
@@ -1181,68 +1199,15 @@ produces
...
-of a query or the beginning of a [parenthesized scope](#scopes).
-
-```mdtest-command
-echo '{s:1}{s:10.0.0.1}' | zq -z 'type foo = {s:string} cast(this, foo)' -
-```
-produces
-```mdtest-output
-{s:"1"}(=foo)
-{s:"10.0.0.1"}(=foo)
-```
```
philrz, line 1235: Upon seeing this text seemed redundant with sections 3 & 4, I pre-cleared with @nwt that:

> `const` and `type` are statements, not expressions, so 6.12 and 6.13 should go away.

docs/language/overview.md
```diff
@@ -1572,13 +1548,13 @@ the "in" operator, e.g.,
 
 ##### 7.2.1.6 Predicate Search Term
 
-Any Boolean-valued [function](functions/README.md) like `is()`, `has()`,
-`grep()` etc. and any [comparison expression](#comparisons)
+Any Boolean-valued [function](functions/README.md) like `is`, `has`,
+`grep` etc. and any [comparison expression](#62-comparisons)
```
philrz, line 1552: In other language docs it is indeed common to use the `()` as shorthand for describing a function. However, since in so many places in the doc the word "function" was already being used, I changed these to adopt that convention.

docs/language/overview.md
```diff
@@ -1572,13 +1548,13 @@ the "in" operator, e.g.,
...
 may be used as a search term and mixed into a search expression.
 
 For example,
 ```
-is(<foo>) has(bar) baz x==y+z
+is(<foo>) has(bar) baz x==y+z timestamp > 2018-03-24T17:17:55Z
```
philrz, line 1557: I added this comparison with a time literal because @henridf recently happened to ask me how to "filter out events in a given date range" and said he'd not seen how from the docs. I happened to know that the docs already talk a fair amount about literals and comparison functions, but I respect it's not a given that all audiences will put that together, especially if they're newer to the language. I figure it can't hurt to have examples cover common use cases, and this seemed like a harmless place to do that.

docs/language/overview.md
```diff
@@ -1639,7 +1615,7 @@ Here the lateral scope, described below, creates a subquery
 ```
 yield {name,elem:this}
 ```
-for each sub-sequence of values derived from each outer input value.
+for each subsequence of values derived from each outer input value.
```
philrz, line 1618: When I started the proofreading, the count on "subsequence" was 3 and "sub-sequence" was 2. My dictionary searches also seemed to imply "subsequence" was the more common form.

docs/language/overview.md
```diff
@@ -1763,37 +1739,36 @@ To unify disparate data sources, data is often cleaned up to fit into
...
-> Also, the [yield operator](operators/yield.md#operator)
+> which is supported by the [`rename` operator](operators/rename.md).
+> Also, the [`yield` operator](operators/yield.md)
 > is handy for simply emitting new, arbitrary record literals based on
 > input values and mixing in these shaping functions in an embedded record literal.
-> The [fuse aggregate function](aggregates/fuse.md) is useful for fusing
-> values into a common schema within a group-by schema though a type is returned
-> rather than values.
+> The [`fuse` aggregate function](aggregates/fuse.md) is also useful for fusing
+> values into a common schema, though a type is returned rather than values.
```
philrz, line 1767: Please review this one closely. The original sentence seemed confusing and I thought it could be simplified, but I want to make sure I didn't wreck its essence.

docs/language/overview.md
```diff
@@ -2025,35 +2000,35 @@ requires a type union for field `x` and produces:
...
 ```mdtest-output
 {fuse:<{x:(int64,string),y:string}>}
 ```
-Since the _fuse_ here is an aggregate function, it can also be used with
-group-by keys.  Supposing we wanted to fuse different type records into
+Since the `fuse` here is an aggregate function, it can also be used with
+group-by keys.  Supposing we wanted to fuse different categories of records into
 different types, we can use a group-by.  In this simple example, we will
-fuse records based on their number of fields using the `len()` function:
+fuse records based on their number of fields using the [`len` function:](functions/len.md)
```
philrz, line 2021: Please review this one closely. "...different type records into different types..." seemed tautological, so hopefully my change is ok.

docs/language/overview.md
```diff
@@ -1418,18 +1394,18 @@ is a Boolean comparison between the product `a*b` and `c`.
 ### 7.2 Search Logic
 
 The search patterns described above can be combined with other elements
-to a search expression comprised of "search terms" that may be combined
+of a search expression comprised of "search terms" that may be combined
 using Boolean logic.
```
nwt, line 1398: I find this sentence a bit confusing. How about this?

```
The search patterns described above can be combined with other "search terms"
using Boolean logic to form search expressions.
```

@mccanne: You might want to weigh in here.

docs/language/overview.md
nwt, line 1552:
```suggestion
`grep`, etc. and any [comparison expression](#62-comparisons)
```

docs/language/overview.md
```diff
@@ -1699,10 +1675,10 @@ the value `this` refers to the inner sequence generated from the `over` expressi
 This query runs to completion for each inner sequence and emits
 each subquery result as each inner sequence traversal completes.
 
-This structure is powerful because _any_ Zed query can be appear in the body of
+This structure is powerful because _any_ Zed query can appear in the body of
 the lateral scope.  In contrast to the `yield` example, a sort could be
-applied to each sub-sequence in the subquery, where sort
-reads all of values of the subsequence, sorts them, emits them, then
+applied to each subsequence in the subquery, where sort
+reads all of the values of the subsequence, sorts them, emits them, then
```
nwt, line 1681: This is a long sentence, so removing a couple words that don't affect its meaning seem like a win.
```suggestion
reads all values of the subsequence, sorts them, emits them, then
```

docs/language/overview.md
nwt: @mccanne: You should probably weigh in here.

docs/language/overview.md
nwt: I think that's better but still not super clear. Maybe this?

```
Supposing we want to divide records into categories and fuse the records in each
category, we can use a group-by.
```

docs/language/overview.md
philrz: @nwt: I agree it's important. Therefore, I've added a second example. Please give it another look.

docs/language/overview.md
philrz: @nwt: Sure, I like that wording. I've pushed a commit to use that wording and also adjusted the post-fix wrapping.

docs/language/overview.md
```diff
@@ -1060,11 +1078,11 @@ where a `<spec>` has one of three forms:
 The first form is a customary colon-separated field and value similar to JavaScript,
 where `<field>` may be an identifier or quoted string.
 The second form is an implied field reference `<ref>`, which is shorthand
-for `<ref>:<ref>`.  The third form is the `...` spread operator which expects
+for `<ref>:<ref>` (XXX).  The third form is the `...` spread operator which expects
```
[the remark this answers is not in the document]
philrz: @nwt: Yes, `<field>` instead of `<ref>` definitely improves it for me, so I've pushed that change. I also realize now that when highlighting this text, I'd already forgotten that there was separate section already in the doc that explicitly describes what an "implied field reference" is, so I added a link to that as well.

I think it still leaves open the question of how to briefly describe what it's shorthand for. As it now stands, I've got it as "...which is shorthand for `<field>:<ref>`." In my mind, instead of `<ref>` it's be great if there's a way it could effectively say `<the value associated with <field>>` but I know that's wordy and awkward, so I'm happy with `<ref>` if you are. :smile:

docs/language/overview.md
```diff
@@ -342,10 +340,10 @@ the implied record field named `foo`.
 
 Another common query pattern involves adding or mutating fields of records
 where the input is presumed to be a sequence of records.
-The [put operator](operators/put.md#operator) provides this mechanism and the `put`
+The [`put` operator](operators/put.md) provides this mechanism and the `put`
 keyword is implied by the mutator syntax `:=`, which is used in Zed when an
-input record field is modified, as compared to `=` which is used in constant
-and variable assignments.  For example, the operation
+input record field is modified, as compared to `=` which is used in [constant](#3-const-statements)
+and [variable](#8-lateral-subqueries) assignments.  For example, the operation
```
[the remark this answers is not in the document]
philrz: @nwt: I agree the repetition can be removed, though with your proposed wording that would be the only time in the doc that "mutator" is mentioned, whereas the prior section that covers the topic speaks of it as "field assignment". I've pushed a change to re-use that terminology and added a link to the section which seems like a reasonable way to provide coverage should someone start their journey by landing right in this section. Please give it a look.

docs/language/overview.md
```diff
@@ -152,23 +151,23 @@ on values as they are produced.  For example, a long running query that
 produces incremental output will stream results as they are produced, i.e.,
 running `zq` to standard output will display results incrementally.
 
-The `search` and `where` operators "find" values in their input and drop
+The [`search`](operators/search.md) and [`where`](operators/where.md) operators "find" values in their input and drop
```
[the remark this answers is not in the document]
philrz: @nwt: Sure, happy to accept all these wrapping suggestions. I'm sometimes unsure where people land on this topic. I've known some reviewers to get annoyed when formatting-only changes accompany meaningful changes in content and hence prefer some long lines rather than additional lines being flagged in the diff. I also wasn't sure where your preferences were on follow-on wraping changes, e.g., by accepting just this suggestion, now the next line becomes way shorter than 80 columns, even though it has a line break mid-sentence, so I wasn't sure if I'd then want to reformat the rest of the paragraph, but I think I see your preferences now. I'm fine with whatever, so I'll try to remember this for future passes in docs.

docs/language/overview.md
philrz: @nwt: Good catch. I am also a fan of the Oxford comma!

docs/language/overview.md
philrz: @nwt: That wording is fine with me, but I'll wait to see if @mccanne has input before pushing a change.

docs/language/overview.md
mccanne: LGTM

docs/language/overview.md
```diff
@@ -288,8 +287,7 @@ or
 summarize salary:=sum(income) by address:=lower(address)
 ```
 This style of "assignment" to a record value is distinguished from the `=`
-token which binds a locally scoped name to a value that can be referenced
-in later expressions.
+token which is used in [constant](#3-const-statements) and [variable](#8-lateral-subqueries) assignments.
```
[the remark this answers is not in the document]
mccanne: I wouldn't make this change because the language I had refers generically to all these cases and as Noah points out, we were going to take it out for `type` (but I wasn't sure we should take it out for `const`).  Currently, you can say this for example...
```
over foo with x => (
  const y = 10
  yield this+x+y
)
```

docs/language/overview.md
mccanne: +1 for @nwt suggestion

docs/language/overview.md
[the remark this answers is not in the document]
philrz: @mccanne: Ok, I interpret your comment as declining @nwt's proposal, but I'd therefore appreciate your take on my original proposed change. I agree that your original generic wording seemed accurate, but my motive for linking to the "constant" & "variable" sections is because I sometimes get nervous that certain words might only be clear to people with language chops ("token", "locally scoped"). Given it's a language overview doc, I suppose that's fair game, but I find linking to examples can sometimes help less-experienced audiences connect the dots. But if you feel it becomes inaccurate once changed from you original text, I'd be fine with restoring that.

docs/language/overview.md
```diff
@@ -2025,35 +2012,36 @@ requires a type union for field `x` and produces:
...
 ```mdtest-output
 {fuse:<{x:(int64,string),y:string}>}
 ```
-Since the _fuse_ here is an aggregate function, it can also be used with
-group-by keys.  Supposing we wanted to fuse different type records into
-different types, we can use a group-by.  In this simple example, we will
-fuse records based on their number of fields using the `len()` function:
+Since the `fuse` here is an aggregate function, it can also be used with
+group-by keys.  Supposing we want to divide records into categories and fuse
+the records in each category, we can use a group-by. In this simple example, we
```
nwt, line 2032:
```suggestion
the records in each category, we can use a group-by.  In this simple example, we
```

docs/language/overview.md
```diff
@@ -1054,17 +1085,17 @@ Record literals have the form
...
 ...<expr>
 ```
 The first form is a customary colon-separated field and value similar to JavaScript,
 where `<field>` may be an identifier or quoted string.
-The second form is an implied field reference `<ref>`, which is shorthand
-for `<ref>:<ref>`.  The third form is the `...` spread operator which expects
-a record value as the result of `<expr>` and inserts all of the fields from
-the resulting record.
+The second form is an [implied field reference](#24-implied-field-references)
+`<ref>`, which is shorthand for `<field>:<ref>`.  The third form is the `...`
```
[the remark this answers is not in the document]
philrz: @nwt: Ok, I'm putting it back to the way it was originally then. That is, not only am I accepting your most recent suggestion but also changing it back to `<ref>` in the table right above. My reason for making any change was because I couldn't understand what it was trying to express as originally written, but once I realized there was a section to link to with more info on implied field references, that alone helps.

docs/language/overview.md
```diff
@@ -1474,7 +1450,7 @@ Searches for any string that begins with `foo` has the string
 
 Keywords and string literals are equivalent search terms so it is often
 easier to quote a string search term instead of using escapes in a keyword.
-Keywords are useful in interactive modes of use where searches can be issued
+Keywords are useful in interactive contexts where searches can be issued
```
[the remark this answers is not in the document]
philrz: @nwt: Maybe I'm officially overthinking it at this point, but "settings" makes me think of configurable settings. I just pushed a change so that it now says "interactive workflows". I even found a spot earlier in the doc where text-not-written-by-me already said "interactive query workflows", so it even seems there's precedent.

docs/language/overview.md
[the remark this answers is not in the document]
philrz: @nwt: Ok, I've pushed a commit putting back the original text then. I still think it could use some improvement, but it doesn't have to happen now. Since you guys made it sound like `=` may be phased out in other contexts soon anyway, happy to wait for the dust to settle.

nwt (approved):

PR #3908, opened 2022-05-19
Add support of index operations on sets

mattnibs:


runtime/expr/eval.go
```diff
@@ -706,7 +706,7 @@ func (i *Index) Eval(ectx Context, this *zed.Value) *zed.Value {
 	}
 }
 
-func indexArray(zctx *zed.Context, ectx Context, typ *zed.TypeArray, array zcode.Bytes, index *zed.Value) *zed.Value {
+func indexVector(zctx *zed.Context, ectx Context, inner zed.Type, b zcode.Bytes, index *zed.Value) *zed.Value {
```
nwt, line 709: Nit: Stick with the convention of giving the `zcode.Bytes` parameter a meaningful name.
```suggestion
func indexVector(zctx *zed.Context, ectx Context, inner zed.Type, vector zcode.Bytes, index *zed.Value) *zed.Value {
```

nwt (approved):

philrz:
This happens to be top-of-mind for me because I just opened #3905, but the language overview doc would also need an addition to mention this case, probably in the section [https://github.com/brimdata/zed/blob/main/docs/language/overview.md#66-indexing](https://github.com/brimdata/zed/blob/main/docs/language/overview.md#66-indexing).

PR #3912, opened 2022-05-23
use net/netip instead of inet.af/netaddr

nwt:
Note that this really commits us to Go 1.18 since that version introduces net/netip.


mattnibs (approved):
Great!

Issue #3895, opened 2022-05-13
Put query is failing on large json zng file


muthu-rk:
@nwt: This is related to #3881. Large JSON files shown in this issue are present in #3881.

I am using zq from main branch after #3881 is present. Unfortunately, the -version is showing unknown.

Update.zed:
over arr | switch id (
  case 2 =>  put level0.level1.value:="2-changed"
  default => pass 
)
| merge id
| arr:=collect(this)

$ ~/go/bin/zq -I update.zed 500mb.zng | zq -Z 'over arr | id==2' -
stdio:stdin: format detection error
	zeek: line 1: bad types/fields definition in zeek header
	zjson: line 1: invalid character 'O' looking for beginning of value
	zson: zson syntax error
	zng: zngio: uncompressed length exceeds MaxSize
	zng21: zng type ID out of range
	csv: record on line 1023: wrong number of fields
	json: invalid character 'O' looking for beginning of value
	parquet: auto-detection not supported
	zst: auto-detection not supported

The same issue is observed for 100MB file too. However the above query works for 10MB zng file.
[10mb.zng.zip](https://github.com/brimdata/zed/files/8689673/10mb.zng.zip)

Interestingly, cut operation like below works on 1GB zng file too.
$ ~/go/bin/zq -z -i zng 'over arr | where id==101 | cut level0.level1' 1gb.zng

Are there any commandline parameters I can tweak to make this query work on large files?

mccanne:
The auto-detector can't handle the large files and it's the second `zq` in the pipeline above causing the problem.  You can avoid the auto-detector with `-i`.  If you run this it should work:
```
~/go/bin/zq -I update.zed 500mb.zng | zq -i zng -Z 'over arr | id==2' -
```

mccanne:
We should probably improve the ZNG auto-detector...

muthu-rk:
@mccanne: With -i option, I dont see the error anymore. But I dont get any output printed on stdin from second zq.

Also, I missed to call, the latest zq binary, the second time. Fixed the query and ran this.

`~/go/bin/zq -I update.zed 500mb.zng | ~/go/bin/zq -i zng -Z 'over arr | id==2' -
`

Can you try this at your end to see if it works for you? I see no output for all these files.

[100mb](https://github.com/brimdata/zed/files/8669438/100mb.zng.zip)
[500mb](https://github.com/brimdata/zed/files/8669461/500mb.zng.zip)
[...]

Issue #1813, opened 2020-12-14
per-row size limits on aggregators


mccanne:
There should be a maxrowsize option on aggregations that also influences the max-read-size buffer in zngio.  Maximum row size is more intuitive and is the real bottleneck here.  When we implement this in aggregators we should add a comand-line option to zq to go with the maxreadsize flag.

mccanne:
We should have a general limit on aggregation outputs, perhaps both in terms of "length", i.e., number of contained values as as in terms of bytes occupied by the row, e.g.,
```
union(query) by id.resp_h with limit=10
```
which would limit each row to 10 set elements or 
```
union(query) by id.resp_h with limit=5MiB
```
which would (roughly) limit each row to 5MiB of row memory footprint.

mccanne:
This supersedes issue #1494.

mccanne:
Moving to backlog to confirm this is implemented.

philrz:
This is a note-to-self because I seem prone to forgetting the history as regards wide rows/values & often wonder if we can do some kind of spill-to-disk rather than requiring limits & knobs.

In a recent chat @mccanne explained that our design currently assumes that a value can fit in memory and hence there has to be some kind of circuit breaker if a user's program tries to make a single value that's larger than would fit in memory, such as is the case with these aggregations. He also explained that this is also how other data warehouses work today. To make our circuit breaker friendly, we've considered wrapping in some kind of error so the user has the benefit of partial results in addition to being made aware that they've exceeded limits.

PR #3921, opened 2022-05-26
add -aggmem flag to zed and zq

nwt:
The -aggmem fleg sets runtime/expr/agg.MaxValueSize, which limits the
size of values produced by aggregate functions.

For #1813 and #3895.


cli/runtimeflags/flags.go
```diff
@@ -29,11 +30,14 @@ func defaultMemMaxBytes() uint64 {
...
 type Flags struct {
 	// these memory limits should be based on a shared resource model
+	aggMemMax  auto.Bytes
 	sortMemMax auto.Bytes
 	fuseMemMax auto.Bytes
 }
 
 func (f *Flags) SetFlags(fs *flag.FlagSet) {
+	f.aggMemMax = auto.NewBytes(uint64(agg.MaxValueSize))
+	fs.Var(&f.aggMemMax, "aggmem", "maximum memory used by aggregate function values in MiB, MB, etc")
```
mattnibs, line 40: The description made me think this was the max memory for an entire aggregate function (which is unbounded) rather than the max memory for an individual value.

nwt: @mattnibs: How about "maximum memory used per aggregate function value" or "maximum memory used per aggregate function result"?

mattnibs: Sorry I totally dropped the ball on this! I like "maximum memory used per aggregate function value".

nwt: No worries!

Made this change in c6c2533dc6b21ba3a65ef12f351afa47ce5db66c.

mattnibs (approved):

PR #3927, opened 2022-06-02
Add array/set spread operators to language docs

mattnibs:


docs/language/overview.md
```diff
@@ -1086,8 +1086,19 @@ produces
...
+where a `<spec>` has one of two forms:
+```
+<expr>
+...<expr>
+```
+
+The first form is simply an element in the array, the result of `<expr>`. The
+second form is the `...` spread operator which expects an array or set value as
+the result of `<expr>` and inserts all of the values from the result. If a spread
+expression results in neither a set or array an error is returned.
```
nwt, line 1100:
```suggestion
The first form is simply an element in the array, the result of `<expr>`.  The
second form is the `...` spread operator which expects an array or set value as
the result of `<expr>` and inserts all of the values from the result.  If a spread
expression results in neither an array nor set, then the value is elided.
```

docs/language/overview.md
```diff
@@ -1101,12 +1112,32 @@ produces
...
+where a `<spec>` has one of two forms:
+```
+<expr>
+...<expr>
+```
+
+The first form is simply an element in the set, the result of `<expr>`. The
+second form is the `...` spread operator which expects an array or set value as
+the result of `<expr>` and inserts all of the values from the result. If a spread
+expression results in neither a set or array an error is returned.
```
nwt, line 1139:
```suggestion
The first form is simply an element in the set, the result of `<expr>`.  The
second form is the `...` spread operator which expects an array or set value as
the result of `<expr>` and inserts all of the values from the result.  If a spread
expression results in neither an array nor set, then the value is elided.
```

nwt (approved):
:shipit:

PR #3928, opened 2022-06-03
zed index create: add long description

mattnibs:


cmd/zed/index/create.go
```diff
@@ -20,9 +20,22 @@ import (
 
 var create = &charm.Spec{
 	Name:  "create",
-	Usage: "create [options] rule-name pattern",
+	Usage: "create [options] rule-name rule-type value",
 	Short: "create an index rule for a lake",
-	New:   newCreate,
+	Long: `The index create command creates and index rule that can be applied to any
```
nwt, line 25: Other `charm.Spec.Long` values begin with a newline. And s/and/an/.
```suggestion
	Long: `
The index create command creates an index rule that can be applied to any
```

cmd/zed/index/create.go
```diff
@@ -20,9 +20,23 @@ import (
...
 	Short: "create an index rule for a lake",
-	New:   newCreate,
+	Long: `
+The index create command creates an index rule that can be applied to any
+pool in the Zed lake. The command takes three arguments: the name of the rule,
+the type of the rule, and the value for the rule.
+
+The name of the index rule must be unique.
+
+The rule's type can be either: field, type, or agg (currently only field rules
```
nwt, line 32:
```suggestion
The rule's type can be either field, type, or agg (currently only field rules
```

cmd/zed/index/create.go
```diff
@@ -20,9 +20,23 @@ import (
 
 var create = &charm.Spec{
 	Name:  "create",
-	Usage: "create [options] rule-name pattern",
+	Usage: "create [options] rule-name rule-type value",
 	Short: "create an index rule for a lake",
-	New:   newCreate,
+	Long: `
+The index create command creates an index rule that can be applied to any
+pool in the Zed lake. The command takes three arguments: the name of the rule,
```
nwt, line 27: Most other `Long` values use two spaces after a sentence.
```suggestion
pool in the Zed lake.  The command takes three arguments: the name of the rule,
```

nwt (approved):

PR #3930, opened 2022-06-03
Optimized pool key predicates

### LINKED ISSUE #3745: Optimize filters on pool keys
If a user is filtering on a pool key field (i.e. `ts >= now()-3h`, the optimizer should recognize this and limit the range of scanned data.

mattnibs:
This pr introduces new functionality that uses pool key predicates in the
pushdown to optimize both object and object seek selection. Both objects
and their seek offsets are selected by using augmented version of the
filter used in the original query. Optimized pool key lookups are limited in
the following ways:

- Will only work on the basic comparison operators "==", ">=", ">",
  "<=", "<"
- The right hand side must be a literal.
- Will not work on expressions nested under unary operators.

This pr also removes the from pool range syntax that was the former
approach to limiting the scope of a query.

Fixes #3745


mattnibs:
Still pending before I take this out of draft mode:
- [ ] Write a more comprehensive test suite for these changes
- [ ] Update documentation- remove references to range syntax and add documentation about pool key predicates are optimized
- [ ] Fix issue with parentheses in zfmt

compiler/parser/parser.peg
```diff
@@ -480,8 +480,8 @@ Pool
   = "pool" _ body:PoolBody { RETURN(body) }
 
 PoolBody
-  = spec:PoolSpec at:PoolAt? over:PoolRange? order:OrderArg? {
-      RETURN(MAP("kind": "Pool", "spec": spec, "at": at, "range": over, "scan_order": order))
+  = spec:PoolSpec at:PoolAt? order:OrderArg? {
```
nwt, line 483: To avoid breakage for users, @mccanne and I think we should keep this in the language but remove it from the documentation and examples. That probably means keeping it in the AST, removing it from the DAG, and creating a `dag.Filter` for it during semantic analysis. Happy to talk that over if you'd like.

compiler/kernel/op.go
```diff
@@ -651,6 +643,11 @@ func (b *Builder) evalAtCompileTime(in dag.Expr) (val *zed.Value, err error) {
 	return e.Eval(expr.NewContext(), b.zctx().Missing()), nil
 }
 
+func CompileExpr(in dag.Expr) (expr.Evaluator, error) {
```
nwt, line 646: Don't export this (unless you have plans to use this from another package).

lake/data/reader.go
```diff
@@ -36,46 +18,18 @@ type Reader struct {
...
-			return nil, err
-		}
-	}
-	if !o.ScanRange.IsZero() {
-		rg = rg.Crop(o.ScanRange)
-	}
-	rg = rg.TrimEnd(sr.TotalBytes)
 	sr.ReadBytes = rg.Size()
 	sr.Reader, err = rg.Reader(reader)
 	return sr, err
```
nwt, line 35: Call `reader.Clsoe` if `rg.Reader(reader)` fails.
```go
	r, err := rg.Reader(reader)
	if err != nil {
		reader.Close()
		return nil, err
	}
	return &Reader{
		Reader:     r,
		Closer:     reader,
		TotalBytes: o.Size,
		ReadBytes:  rg.Size(),
	}, nil
```

lake/index/expr.go
```diff
@@ -25,6 +25,9 @@ type result struct {
 // - Equals comparisons chained with 'and' or 'or' statements.
 // - Leaf BinaryExprs with the LHS of *dag.Path and RHS of *dag.Literal.
 func compileExpr(node dag.Expr) expr {
+	if node == nil {
+		return nil
+	}
```
nwt, line 30: You don't need this since the type assertion on the next line will fail if `node` is nil.
```suggestion
```

lake/partition.go
```diff
@@ -138,12 +129,12 @@ func sortObjects(o order.Which, objects []*data.Object) {
 	}
 }
 
-func partitionReader(ctx context.Context, zctx *zed.Context, snap commits.View, span extent.Span, order order.Which) (zio.Reader, error) {
+func partitionReader(ctx context.Context, zctx *zed.Context, pool *Pool, snap commits.View) (zio.Reader, error) {
```
nwt, line 132: Nit: Pass an `order.Layout` instead of a `*Pool`.

compiler/ast/dag/keyfilter.go
```diff
@@ -0,0 +1,137 @@
...
+			return nil
+		}
+		rhs, ok := e.RHS.(*Literal)
+		if !ok {
+			return nil
+		}
+		// Copy this.
+		var lhs This
+		lhs.Path = append(lhs.Path, this.Path...)
+		return v(e.Op, &lhs, rhs)
```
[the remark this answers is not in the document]
nwt: More generally, initialize the Kind field whenever you create a dag node.

compiler/parser/parser.peg
mattnibs: Ok.

lake/index/expr.go
mattnibs: Did not know that, thanks.

docs/commands/zed.md
```diff
@@ -247,7 +247,10 @@ When pool data is not structured as records/objects (e.g., scalar or arrays or o
 non-record types), then the pool key would typically be configured
 as the [special value `this`](../language/README.md#23-the-special-value-this).
 
-Data can be efficiently scanned via ranges of values conforming to the pool key.
+Data can be efficiently scanned if a query has a filter operating on the pool
+key. For example on a pool with pool key `ts`, the query `ts == 100`
```
nwt, line 251: Prevailing style here is two spaces between sentences.
```suggestion
key.  For example on a pool with pool key `ts`, the query `ts == 100`
```

docs/commands/zed.md
```diff
@@ -247,7 +247,10 @@ When pool data is not structured as records/objects (e.g., scalar or arrays or o
...
+will be optimized to scan only the objects where the value `100` could be
```
nwt, line 252:
```suggestion
will be optimized to scan only the data objects where the value `100` could be
```

compiler/ast/dag/keyfilter.go
```diff
@@ -0,0 +1,119 @@
...
+	return &KeyFilter{e}
+}
+
+// SpanFilter creates an Expr that returns true if the KeyFilter has a value
+// within a span of values. The span compared against must be a record with the
+// fields "lower" and "upper", where "lower" is the inclusive lower bounder and
+// "upper" is the exclusive upper bound.
+func (k *KeyFilter) SpanFilter(o order.Which, prefix ...string) Expr {
+	lower := append([]string{}, append(prefix, "lower")...)
+	upper := append([]string{}, append(prefix, "upper")...)
```
nwt, line 39: Copy `prefix` before appending to avoid unexpectedly modifying its underlying array.
```suggestion
	lower := append(append([]string{}, prefix...), "lower")
	upper := append(append([]string{}, prefix...), "upper")
```

compiler/ast/dag/keyfilter.go
```diff
@@ -0,0 +1,119 @@
...
+			this.Path = upper
+		}
+		return relativeToCompare(cmp, this, val, o)
+	})
+}
+
+// CroppedByFilter produces an expression which returns true if a span known
+// to overlap the key filter is cropped by filter (i.e., not all values in the
+// span evaluate true against the KeyFilter.
+func (k *KeyFilter) CroppedByFilter(o order.Which, prefix ...string) Expr {
```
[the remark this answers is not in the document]
mattnibs: This doesn't quite work since lower and upper get switched for "<", etc. when checking if cropped.

lake/ztests/seek-index-null.yaml
```diff
@@ -5,7 +5,7 @@ script: |
     echo // $o | tee /dev/stderr
     zed create -q -seekstride 2KB -orderby ts:$o $o
     zed load -q -use $o babble.zson null.zson
-    zed query -z -s "from $o range 2020-04-21T23:59:26.063Z to 2020-04-21T23:59:38.069Z"
+    zed query -z -s -use $o "ts >= 2020-04-21T23:59:26.063Z and ts <= 2020-04-21T23:59:38.069Z"
```
nwt, line 8: Any reason for changing from `from` to `-use`? (Just curious.)

mattnibs: It's more a matter of taste. I don't like writing from in single pool queries and like the consistency of using the `-use` arg as with the other commands. Also with this change `from` isn't necessary.

lake/scheduler.go
```diff
@@ -2,53 +2,48 @@ package lake
...
 }
 
 var _ op.Scheduler = (*Scheduler)(nil)
 
-func NewSortedScheduler(ctx context.Context, zctx *zed.Context, pool *Pool, snap commits.View, span extent.Span, filter zbuf.Filter) (*Scheduler, error) {
-	var idx *index.Filter
-	if kf := filter.(*kernel.Filter); kf != nil {
-		idx = index.NewFilter(pool.engine, pool.IndexPath, kf.Pushdown)
+func NewSortedScheduler(ctx context.Context, zctx *zed.Context, pool *Pool, snap commits.View, filter zbuf.Filter) (*Scheduler, error) {
+	ranger, err := newSeekFinder(pool, snap, filter)
```
nwt, line 35: Nit: s/ranger/finder/

runtime/expr/spanfilter.go
```diff
@@ -0,0 +1,48 @@
...
+	o.cols[1].Type = upper.Type
+	if o.typ == nil || o.cols[0].Type != lower.Type && o.cols[1].Type != upper.Type {
+		o.typ = o.zctx.MustLookupTypeRecord(o.cols)
+	}
+	o.builder.Reset()
+	o.builder.Append(lower.Bytes)
+	o.builder.Append(upper.Bytes)
+	val := o.eval.Eval(o.ectx, o.ectx.NewValue(o.typ, o.builder.Bytes()))
+	if val.Type != zed.TypeBool {
+		panic(fmt.Errorf("result of SpanFilter not a boolean: %s", zson.String(val.Type)))
```
nwt, line 45:
```suggestion
		panic(fmt.Errorf("result of SpanFilter not a boolean: %s", zson.MustFormatValue(val.Type)))
```

runtime/expr/spanfilter.go
```diff
@@ -0,0 +1,48 @@
...
+func (o *SpanFilter) Eval(lower, upper *zed.Value) bool {
+	o.cols[0].Type = lower.Type
+	o.cols[1].Type = upper.Type
+	if o.typ == nil || o.cols[0].Type != lower.Type && o.cols[1].Type != upper.Type {
+		o.typ = o.zctx.MustLookupTypeRecord(o.cols)
+	}
+	o.builder.Reset()
+	o.builder.Append(lower.Bytes)
+	o.builder.Append(upper.Bytes)
+	val := o.eval.Eval(o.ectx, o.ectx.NewValue(o.typ, o.builder.Bytes()))
```
nwt, line 43: Replace the `typ` field with `val zed.Value` to avoid allocating a `*zed.Value` here.
```suggestion
	if o.val.Type == nil || o.cols[0].Type != lower.Type && o.cols[1].Type != upper.Type {
		o.val.Type = o.zctx.MustLookupTypeRecord(o.cols)
	}
	o.builder.Reset()
	o.builder.Append(lower.Bytes)
	o.builder.Append(upper.Bytes)
	o.val.Bytes = o.builder.Bytes()
	val := o.eval.Eval(o.ectx, &o.val)
```

philrz:
@mattnibs: Does the filtering need to happen in a single pipeline element to be recognized as optimize-able? I ask because my recent close scrutiny of the Brim app's correlation queries led me to notice that the app currently does the time filtering in two pieces (e.g., `| ts >= 2015-03-05T14:25:12.943Z | ts <= 2015-04-13T09:34:44.733Z`) so if that's going to miss the optimizer we'll need to make sure that gets changed.

nwt (approved):
I don't really have my head all the way around this, but I think we should go ahead and merge it.

runtime/expr/spanfilter.go
```diff
@@ -0,0 +1,49 @@
...
+		},
+		ectx: NewContext(),
+		zctx: zed.NewContext(),
+	}
+}
+
+func (o *SpanFilter) Eval(lower, upper *zed.Value) bool {
+	o.cols[0].Type = lower.Type
+	o.cols[1].Type = upper.Type
+	if o.val.Type == nil || o.cols[0].Type != lower.Type && o.cols[1].Type != upper.Type {
```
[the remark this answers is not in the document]
mattnibs: This sounds right, good catch.

compiler/semantic/op.go
```diff
@@ -179,8 +178,6 @@ func semPool(ctx context.Context, scope *Scope, p *ast.Pool, ds *data.Source, he
 				Meta:      p.Spec.Meta,
 				Pool:      poolID,
 				Commit:    commitID,
-				ScanLower: lower,
-				ScanUpper: upper,
```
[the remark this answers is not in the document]
mattnibs: Because this isn't be used any more. See https://github.com/brimdata/zed/pull/3930/files#diff-072c6f760a72cdd01fcb450006cda7ebdbf0ac0ff18baffff5db2c742b017ce2R88 range expressions get added to the pushdown filter.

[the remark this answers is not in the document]
nwt: In that case, can/should we remove remove these field from compiler/ast/dag.CommitMeta?

compiler/ast/dag/keyfilter.go
```diff
@@ -0,0 +1,108 @@
...
+// filter values that would not be not also filtered by the original filter).
+// Currently KeyFilter only recognizes simple key predicates against a literal
+// value and using the comparators ==, >=, >, <, and <=; otherwise the predicate is
+// ignored.
+func NewKeyFilter(key field.Path, node Expr) *KeyFilter {
+	e := visitLeaves(node, func(cmp string, lhs *This, rhs *Literal) Expr {
+		if !key.Equal(lhs.Path) {
+			return nil
+		}
+		return NewBinaryExpr(cmp, lhs, rhs)
```
mattnibs, line 25: Ugh there's an issue here on a filter: `poolkey == "foo" or notpoolkey == "bar"`. This will get optimized but should not as it could result in false negatives (unless `notpoolkey` is indexed but that's a situation we do not handle).

compiler/semantic/op.go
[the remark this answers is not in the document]
mattnibs: Indeed we should.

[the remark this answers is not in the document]
mattnibs: Actually nevermind this removal is a mistake and should be added back.

PR #3933, opened 2022-06-04
Index Writer: fix bug with LZ4 size

mattnibs:
Fix bug in the IndexWriter where the zngio.Writer LZ4 options was not
getting set, resulting in the index getting created with a bunch of
single value value frames.


mattnibs:
Closing as testing will need another nob and will rebuild on https://github.com/brimdata/zed/pull/3934

PR #3935, opened 2022-06-04
indexfile: use default zng lz4 block size

mattnibs:
Fix bug in the IndexWriter where the zngio.Writer LZ4 block size was not
getting set, resulting in the index getting created with a bunch of
single value value frames.


cmd/zed/dev/indexfile/create/command.go
```diff
@@ -59,6 +60,8 @@ func newCommand(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
 	})
 	f.StringVar(&c.outputFile, "o", "index.zng", "name of index output file")
 	f.StringVar(&c.keys, "k", "", "comma-separated list of field names for keys")
+	f.IntVar(&c.opts.ZNGWriterOpts.LZ4BlockSize, "znglz4blocksize", zngio.DefaultLZ4BlockSize,
+		"LZ4 block size in bytes for ZNG compression (nonpositive to disable)")
```
nwt, line 64: The behavior of `-znglz4blocksize=0` is different here than it is everywhere else. I think you should make this consistent, but if you keep it as is, fix the usage string.
```suggestion
		"LZ4 block size in bytes for ZNG compression (negative to disable)")
```

nwt (approved):
The current PR title ("indexfile: use default zng lz4 block size") suggests that this change only affects "zed dev indexfile". Please change it to something that better reflects what's really happening, like "index: enable ZNG compression by default" (and don't forget to use it as the commit title when you merge).

cmd/zed/dev/indexfile/create/command.go
mattnibs: > The behavior of -znglz4blocksize=0 is different here than it is everywhere else. I think you should make this consistent

I don't quite follow, I literally copied this from outputflags.

mattnibs: A couple of other things:

Setting LZ4BlockSize to 0 disables compressions, (so the negative to disable comment is not right):
```
func NewWriter(w io.WriteCloser, opts WriterOpts) *Writer {
	var comp *compressor
	if opts.LZ4BlockSize > 0 {
		comp = &compressor{}
	}
	return &Writer{
		writer:     w,
		compressor: comp,
		opts:       opts,
		types:      NewEncoder(),
		thresh:     opts.LZ4BlockSize, //XXX
	}
}
```
Additionally any value <= 0 for WriterOpts.LZ4BlockSize makes the Writer output a new value block for every value:
```
func (w *Writer) Write(val *zed.Value) error {
...
	if len(w.values) >= w.thresh || len(w.types.bytes) >= w.thresh {
		return w.flush()
	}
```
Which makes the current output for LZ4BlockSize < 0 pretty unusable (unless have a bunch of single value zed blocks is not a big deal).

nwt: > I don't quite follow, I literally copied this from outputflags.

I know, but in `index.NewWriterWithContext` you're adding
```go
	if opts.ZNGWriterOpts.LZ4BlockSize == 0 {
		opts.ZNGWriterOpts.LZ4BlockSize = zngio.DefaultLZ4BlockSize
	}
```
and that means `zed dev indexfile create -znglz4blocksize=0` gives you compression with the default block size, rigfht?

mattnibs: Oh shoot I forgot I put that in there. You are right this is weird. Since I already merged this, I'll put up a separate pr to fix this.

PR #3939, opened 2022-06-13
Object writer: derive spans from values

mattnibs:
Change the object writer so the first and last keys are derrived from
the values written to the object.


lake/writer.go
```diff
@@ -190,17 +184,3 @@ func (s *ImportStats) Copy() ImportStats {
...
-	layout.Keys = field.List{poolKey(layout)}
-	return zbuf.NewComparator(zctx, layout)
-}
-
-func poolKey(layout order.Layout) field.Path {
-	if len(layout.Keys) != 0 {
-		// XXX We don't yet handle multiple pool keys.
-		return layout.Keys[0]
-	}
-	return field.New("ts")
```
nwt, line 205: We should make this change (dropping "ts" as the default), but in a separate PR.

lake/data/writer.go
```diff
@@ -90,11 +85,10 @@ func (w *Writer) writeIndex(key zed.Value) error {
 	w.seekIndexTrigger += len(key.Bytes)
 	if w.first {
 		w.first = false
-		w.firstKey = key
-		w.lastKey = key
+		w.object.First.CopyFrom(&key)
 		return w.seekIndex.Write(key, 0, 0)
 	}
-	if w.seekIndexTrigger < w.seekIndexStride || bytes.Equal(key.Bytes, w.lastKey.Bytes) {
+	if w.seekIndexTrigger < w.seekIndexStride || bytes.Equal(key.Bytes, w.object.Last.Bytes) {
```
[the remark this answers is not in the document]
mattnibs: Oh no, I should add back the part where I set lastKey when I set firstKey.

lake/writer.go
mattnibs: Ok.

nwt (approved):

PR #3941, opened 2022-06-17
refactor compiler a little bit

mccanne:
This commit refactors the compiler and lake so the compiler can import
the lake and do more sophisticated lake things as we integrate warehouse
style optimizations in the Zed lake.  This reverses the previous pattern
where op.DataAdaptor was used by the compiler to talk to the lake.

For now, op.DataAdaptor still exists but we will get rid of this in a
subsequent PR and localize the lakey-ness disposition into the compiler.

### DIFF
--- a/compiler/job.go
+++ b/compiler/job.go
@@ -141,12 +141,6 @@ func (j *Job) Parallelize(n int) error {
 	return j.optimizer.Parallelize(n)
 }
 
-// Parse concatenates the source files in filenames followed by src and parses
-// the resulting program.
-func (*anyCompiler) Parse(src string, filenames ...string) (ast.Op, error) {
-	return Parse(src, filenames...)
-}
-
 func Parse(src string, filenames ...string) (ast.Op, error) {
 	parsed, err := parser.ParseZed(filenames, src)
 	if err != nil {
@@ -192,6 +186,12 @@ func (j *Job) Puller() zbuf.Puller {
 }
 
 type anyCompiler struct{}
+
+// Parse concatenates the source files in filenames followed by src and parses
+// the resulting program.
+func (*anyCompiler) Parse(src string, filenames ...string) (ast.Op, error) {
+	return Parse(src, filenames...)
+}
 
 func (a *anyCompiler) ParseRangeExpr(zctx *zed.Context, src string, layout order.Layout) (*zed.Value, string, error) {
 	o, err := a.Parse(src)
--- a/compiler/job.go
+++ b/compiler/job.go
@@ -143,7 +143,7 @@ func (j *Job) Parallelize(n int) error {
 
 // Parse concatenates the source files in filenames followed by src and parses
 // the resulting program.
-func (*any) Parse(src string, filenames ...string) (ast.Op, error) {
+func (*anyCompiler) Parse(src string, filenames ...string) (ast.Op, error) {
 	return Parse(src, filenames...)
 }
 
@@ -157,7 +157,7 @@ func Parse(src string, filenames ...string) (ast.Op, error) {
 
 // MustParse is like Parse but panics if an error is encountered.
 func MustParse(query string) ast.Op {
-	o, err := (*any)(nil).Parse(query)
+	o, err := (*anyCompiler)(nil).Parse(query)
 	if err != nil {
 		panic(err)
 	}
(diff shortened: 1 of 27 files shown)


ztest/ztest.go
```diff
@@ -504,7 +505,7 @@ func runzq(path, zedProgram, input string, outputFlags []string, inputFlags []st
 		// tests.
 		return outbuf.String(), errbuf.String(), err
 	}
-	proc, err := compiler.ParseOp(zedProgram)
+	proc, err := compiler.Parse(zedProgram)
```
nwt, line 508: Just do this so we don't need to thread the compiler down through all these calls (or even make it part of the ztest interface)?
```suggestion
	proc, err := zruntime.NewCompiler().Parse(zedProgram)
```

service/core.go
```diff
@@ -51,6 +53,7 @@ type Config struct {
 type Core struct {
 	auth            *Auth0Authenticator
 	conf            Config
+	compiler        runtime.Compiler
```
nwt, line 56:
```suggestion
	compiler        runtime.Compiler
	conf            Config
```

service/core.go
```diff
@@ -133,6 +136,7 @@ func NewCore(ctx context.Context, conf Config) (*Core, error) {
 	c := &Core{
 		auth:          authenticator,
 		conf:          conf,
+		compiler:      compiler.NewLakeCompiler(root),
```
nwt, line 139:
```suggestion
		compiler:      compiler.NewLakeCompiler(root),
		conf:          conf,
```

compiler/job.go
```diff
@@ -128,93 +127,74 @@ func isSequentialWithLeadingFrom(o ast.Op) bool {
...
-	return r.optimizer.Parallelize(n)
+func (j *Job) Parallelize(n int) error {
+	return j.optimizer.Parallelize(n)
 }
 
-// ParseOp concatenates the source files in filenames followed by src and parses
+// Parse concatenates the source files in filenames followed by src and parses
 // the resulting program.
-func ParseOp(src string, filenames ...string) (ast.Op, error) {
+func (*anyCompiler) Parse(src string, filenames ...string) (ast.Op, error) {
```
nwt, line 146: Nit: Move this down below the `anyCompiler` type declaration.

compiler/lake.go
```diff
@@ -2,34 +2,45 @@ package compiler
 
 import (
 	"errors"
-	"runtime"
+	goruntime "runtime"
```
nwt, line 5: Nit: When one of our packages clashes with the standard library like this, I think it'll be a little less surprising for other readers if we rename our package rather than the the standard library package.

lake/index/rule.go
```diff
@@ -65,10 +64,6 @@ func NewTypeRule(name string, typ zed.Type) *TypeRule {
 }
 
 func NewAggRule(name, prog string) (*AggRule, error) {
-	// make sure it compiles
-	if _, err := compiler.ParseOp(prog); err != nil {
-		return nil, err
-	}
```
nwt, line 71: Are you sure you want to get rid of this?

If you do, consider removing the error return parameter.

runtime/op/sort/sort_test.go
```diff
@@ -130,7 +132,7 @@ func cat(ss ...string) string {
 	return out
 }
 
-func runTest(t *testing.T, cmd, input, output string) {
+func runTest(t *testing.T, c runtime.Compiler, cmd, input, output string) {
```
nwt, line 135: This parameter is no longer used, so you can revert the changes to this file.

service/handlers.go
```diff
@@ -38,12 +38,12 @@ func handleQuery(c *Core, w *ResponseWriter, r *Request) {
 	// The client must look at the return code and interpret the result
 	// accordingly and when it sees a ZNG error after underway,
 	// the error should be relay that to the caller/user.
-	query, err := compiler.ParseOp(req.Query)
+	query, err := compiler.Parse(req.Query)
```
nwt, line 41:
```suggestion
	query, err := c.compiler.Parse(req.Query)
```

nwt (approved):

lake/index/rule.go
mccanne: Ok I put it back.  Needed to pass in the a runtime.Compiler because lake packages can't import compiler.

service/handlers.go
mccanne: Hmm, I don't know why this passed our tests because c.compiler has the lake pointer but compiler.Parse() doesn't have a lake.

mccanne: Oh, figured it out.  Parse() just builds the AST and the lake work is done when building the query from the AST.

PR #3945, opened 2022-06-20
remove mock lake and zc REPL

mccanne:
This commit removes the mock lake that was used for testing in favor
of an approach that uses actual lakes for testing the DAG optimizer.
This will make work on the optimizer easier and more reliable as we
add more functionality like the vector engine and summary objects.

As part of this, we updated all the ztests that covered DAG optimizations
to create an actual lake with the pool key under test.

While we were here, we also removed the REPL from zed dev compile.


cmd/zed/dev/compile/command.go
```diff
@@ -142,7 +134,14 @@ func (c *Command) Run(args []string) error {
 		}
 	}
 	src += strings.Join(args, " ")
-	return c.parse(src)
+	var lk *lake.Root
+	if c.semantic || c.optimize || c.parallel != 0 {
+		lakeAPI, err := c.LakeFlags.Open(context.Background())
```
nwt, line 139: `c.Init` up above returns a `context.Context`. Use that here.

compiler/semantic/op.go
```diff
@@ -72,6 +72,9 @@ func semSource(ctx context.Context, scope *Scope, source ast.Source, adaptor op.
 			Layout: layout,
 		}, nil
 	case *ast.Pool:
+		if adaptor == nil {
+			return nil, fmt.Errorf("semantic analyzer: from pool cannot be used without a lake")
```
nwt, line 76:
```suggestion
			return nil, errors.New("semantic analyzer: from pool cannot be used without a lake")
```

compiler/ztests/par-count.yaml
```diff
@@ -1,13 +1,18 @@
-script: zc -C -P 2 "from 'pool-ts' | count() by y"
+script: |
+  export ZED_LAKE=test
+  zed init -q
+  zed create -q -orderby ts test
+  zed use -q test
```
nwt, line 5: I don't think we need this.
```suggestion
```

nwt (approved):

PR #3950, opened 2022-06-23
add vector form of data objects to lake/data

mccanne:
This commit creates the vector concept for cloud data objects and
adds a simple test to make sure the creation and deletion of such
objects works.

We also updated some of the docs and code to replace the "row/column"
terminology with "sequence/vector" terminology to better describe
Zed's model of heterogeneous sequences of deeply typed values compared
to the relational model of tables, schemas, columns, and rows.
There is more work to do here in the report so we created issue ##3949
to track it.

### DIFF
--- a/docs/lake/format.md
+++ b/docs/lake/format.md
@@ -112,7 +112,7 @@ allowing a scan to do a byte-range retrieval of the ZNG object when
 processing only a subset of data.
 
 > Note the ZST format allows individual vector segments to be read in isolation
-> and the in-memory ZST representation is support for random access so there is
+> and the in-memory ZST representation supports random access so there is
 > no need to have a seek index for the vector object.
 
 #### Commit History
--- a/docs/lake/format.md
+++ b/docs/lake/format.md
@@ -73,6 +73,7 @@ such an assumption).
 
 A data object is created by a single writer using a globally unique name
 with an embedded KSUID.  
+
 New objects are written in their entirety.  No updates, appends, or modifications
 may be made once an object exists.  Given these semantics, any such object may be
 trivially cached as neither its name nor content ever change.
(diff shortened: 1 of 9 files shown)


docs/lake/format.md
```diff
@@ -110,9 +111,9 @@ The seek index maps pool key values to seek offsets in the ZNG file thereby
 allowing a scan to do a byte-range retrieval of the ZNG object when
 processing only a subset of data.
 
-> Note the ZST format will have seekable checkpoints based on the sort key that
-> are encoded into its metadata section so there is no need to have a separate
-> seek index for the columnar object.
+> Note the ZST format allows individual vector segments to be read in isolation
+> and the in-memory ZST representation is support for random access so there is
```
nwt, line 115:
```suggestion
> and the in-memory ZST representation supports for random access so there is
```

lake/data/vector.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		get.Close()
+		return err
+	}
+	writer, err := zstio.NewWriter(bufwriter.New(put), zstio.WriterOpts{
+		ColumnThresh: zstio.DefaultColumnThresh,
+		SkewThresh:   zstio.DefaultSkewThresh,
+	})
+	if err != nil {
+		return err
```
nwt, line 33:
```suggestion
		get.Close()
		put.Close()
		DeleteVector(ctx, engine, path, id)
		return err
```

lake/data/vector.go
```diff
@@ -0,0 +1,56 @@
...
+	if closeErr := writer.Close(); err == nil {
+		err = closeErr
+	}
+	if closeErr := reader.Close(); err == nil {
+		err = closeErr
+	}
+	if closeErr := get.Close(); err == nil {
+		err = closeErr
+	}
+	return err
```
nwt, line 48:
```suggestion
	if err != nil {
		DeleteVector(ctx, engine, path, id)
	}
	return err
```

nwt (approved):

PR #3957, opened 2022-06-29
add -splitsize flag to zed and zq

### LINKED ISSUE #3955: Split output on size threshold
As requested by a community user in a [Slack thread](https://brimdata.slack.com/archives/CTSMAK6G7/p1654696247596769):

> Does anyone have any recommendations for chunking large ZNG files outside of pushing to a Zed lake? We have some edge compute cases where we'd like to take very large ZNG files (sometimes in the 10s if not 100 of GBs each) and do some ZQ CLI processing on them and other processes we have with Go and Python libraries. The issue is the ZNG files often are too large to load into memory for some of what we do. My thought was if there was a way to chunk out the file into smaller files, we could work in memory of these remote systems and do some pre-processing before the data is returned to our infrastructure.
If there is nothing in Zed query itself, do you think there would be data corruption issues if we used something like `split`? Something like `split -b 1000k --additional-suffix=.zng largefile.zng smallfileprefix`

nwt:
The -splitsize flag to zed and zq changes the behavior of -split,
splitting output into files of at least the given size rather than by
data type.  Files may exceed the target size substantially due to
buffering in the underlying writer as determined by -f et al.

Closes #3955.

### DIFF
--- a/zio/emitter/sizesplitter.go
+++ b/zio/emitter/sizesplitter.go
@@ -1,0 +1,106 @@
+package emitter
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"strconv"
+
+	"github.com/brimdata/zed"
+	"github.com/brimdata/zed/pkg/bufwriter"
+	"github.com/brimdata/zed/pkg/storage"
+	"github.com/brimdata/zed/zio"
+	"github.com/brimdata/zed/zio/anyio"
+)
+
+type sizeSplitter struct {
+	ctx    context.Context
+	engine storage.Engine
+	dir    *storage.URI
+	prefix string
+	opts   anyio.WriterOpts
+	size   int64
+
+	cwc countingWriteCloser
+	ext string
+	seq int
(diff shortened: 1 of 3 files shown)


mattnibs (approved):

zio/emitter/sizesplitter.go
```diff
@@ -0,0 +1,106 @@
...
+	zwc zio.WriteCloser
+}
+
+// NewSizeSplitter returns a zio.WriteCloser that writes to sequentially
+// numbered files created by engine in dir with optional prefix and with opts,
+// creating a new file after the current one reaches size bytes.  Files may
+// exceed size substantially due to buffering in the underlying writer as
+// determined by opts.Format.
+func NewSizeSplitter(ctx context.Context, engine storage.Engine, dir *storage.URI, prefix string,
+	opts anyio.WriterOpts, size int64) (zio.WriteCloser, error) {
```
mattnibs, line 36: +1 to making long signatures like this multiline.

PR #3958, opened 2022-06-30
zson marshal: support ignore field tag

mattnibs:


zson/marshal_test.go
```diff
@@ -294,3 +294,17 @@ func TestJSONFieldTag(t *testing.T) {
...
+	type ignoreTag struct {
+		Value  string       `zed:"value"`
+		Ignore func() error `zed:"-"`
+	}
+	b, err := zson.Marshal(ignoreTag{Value: "test"})
+	require.NoError(t, err)
+	assert.Equal(t, expected, b)
+	var j ignoreTag
+	require.NoError(t, zson.Unmarshal(b, &j))
+	assert.Equal(t, ignoreTag{Value: "test"}, j)
```
nwt, line 309: Nits: The names here are a little off because there's no "ignore" field tag in play here, just the "zed" field tag. And I don't think `expected` improves readability.
```suggestion
func TestIgnoreField(t *testing.T) {
	type s struct {
		Value  string       `zed:"value"`
		Ignore func() error `zed:"-"`
	}
	b, err := zson.Marshal(s{Value: "test"})
	require.NoError(t, err)
	assert.Equal(t, `{value:"test"}`, b)
	var v s
	require.NoError(t, zson.Unmarshal(b, &v))
	assert.Equal(t, s{Value: "test"}, v)
```

nwt (approved):

zson/marshal_test.go
mattnibs: Pretty!

PR #3966, opened 2022-07-05
zio/zngio: improve Writer constructor ergonomics

nwt:
Creating a zio/zngio.Writer with reasonable default options is
cumbersome.  Make it easier by renaming the current constructor,
NewWriter, to NewWriterWithOpts and adding a new version of NewWriter
that calls NewWriterWithOpts with reasonable default options.

(I plan to introduce separate options for compression and frame threshold in a subsequent PR.)

### DIFF
--- a/runtime/op/spill/file.go
+++ b/runtime/op/spill/file.go
@@ -27,7 +27,7 @@ type File struct {
 // records via the zio.Reader interface.
 func NewFile(f *os.File) *File {
 	return &File{
-		Writer: zngio.NewWriter(bufwriter.New(zio.NopCloser(f)), zngio.WriterOpts{}),
+		Writer: zngio.NewWriterWithOpts(bufwriter.New(zio.NopCloser(f)), zngio.WriterOpts{}),
 		file:   f,
 	}
 }
--- a/zngbytes/serializer.go
+++ b/zngbytes/serializer.go
@@ -20,7 +20,7 @@ func NewSerializer() *Serializer {
 	s := &Serializer{
 		marshaler: m,
 	}
-	s.writer = zngio.NewWriter(zio.NopCloser(&s.buffer), zngio.WriterOpts{})
+	s.writer = zngio.NewWriterWithOpts(zio.NopCloser(&s.buffer), zngio.WriterOpts{})
 	return s
 }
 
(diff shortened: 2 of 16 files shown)


zngbytes/serializer.go
```diff
@@ -20,7 +20,7 @@ func NewSerializer() *Serializer {
 	s := &Serializer{
 		marshaler: m,
 	}
-	s.writer = zngio.NewWriter(zio.NopCloser(&s.buffer), zngio.WriterOpts{})
+	s.writer = zngio.NewWriterWithOpts(zio.NopCloser(&s.buffer), zngio.WriterOpts{})
```
mccanne, line 23: Seems like this should use compression

zst/writer.go
```diff
@@ -163,7 +163,7 @@ func (w *Writer) finalize() error {
 	}
 	// At this point all the column data has been written out
 	// to the underlying spiller, so we start writing zng at this point.
-	zw := zngio.NewWriter(w.writer, zngio.WriterOpts{})
+	zw := zngio.NewWriterWithOpts(w.writer, zngio.WriterOpts{})
```
mccanne, line 166: Compression here too

runtime/op/spill/file.go
```diff
@@ -27,7 +27,7 @@ type File struct {
 // records via the zio.Reader interface.
 func NewFile(f *os.File) *File {
 	return &File{
-		Writer: zngio.NewWriter(bufwriter.New(zio.NopCloser(f)), zngio.WriterOpts{}),
+		Writer: zngio.NewWriterWithOpts(bufwriter.New(zio.NopCloser(f)), zngio.WriterOpts{}),
```
mccanne, line 30: We should test but I think compression would increase perf overall.  Let's make an issue.

mccanne (approved):
Yay.  See comments.

zngbytes/serializer.go
nwt: Agreed. Coming in a separate PR.

runtime/op/spill/file.go
nwt: This'll be my next PR after #3967.

PR #3967, opened 2022-07-05
rework ZNG writer options

nwt:
zio/zngio.WriterOpts contains a single field, LZ4BlockSize, that both
enables compression and specifies the frame threshold, making it
impossible to create a writer with no compression and a positive frame
threshold.  Replace it with separate Compress and FrameThresh fields,
and accordingly, replace the -znglz4blocksize flag to zed, zq, et al.
with new -zngcompress and -zngframethresh flags.

Depends on #3966.

### DIFF
--- a/api/client/request.go
+++ b/api/client/request.go
@@ -102,7 +102,7 @@ func (r *Request) reader() (io.Reader, error) {
 		return nil, err
 	}
 	var buf bytes.Buffer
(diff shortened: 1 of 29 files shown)


mccanne (approved):
cool

mattnibs (approved):

Issue #3924, opened 2022-05-31
Events endpoint no longer returning "Content-Type: text/event-stream" (regression at 8b54766)


philrz:
This issue caused a regression that can be reproduced in the the Brim app's current tip of `main` (commit `9c88527` at the moment.)

To briefly summarize the symptom, if Brim is running and I then perform a `zed create testpool` operation at the shell, the pool name is _not_ automatically reflected in the app. This last worked in at commit `b1092e4` which was right before its Zed dependency was advanced to take on the changes in #3870 (cc: @mattnibs).

I've attached pcaps [works-b1092e46.pcapng.gz](https://github.com/brimdata/zed/files/8801248/works-b1092e46.pcapng.gz) and [fails-e35edbad.pcapng.gz](https://github.com/brimdata/zed/files/8801250/fails-e35edbad.pcapng.gz) that capture the network traffic between Brim and the Zed backend for the last working Brim commit and the first failing commit, respectively. But below are simplified Zed-only repros that capture the essence of what I see as the problem.

At Zed commit 27ceb6ba that was right before the changes in #3870, a running `zed serve` responds on the `/events` endpoint with `Content-Type: text/event-stream`.

```
$ curl http://localhost:9867/version
{"version":"v1.0.0-84-g27ceb6ba"}

$ curl -v -H 'Accept: application/json' http://localhost:9867/events
*   Trying 127.0.0.1:9867...
* Connected to localhost (127.0.0.1) port 9867 (#0)
> GET /events HTTP/1.1
> Host: localhost:9867
> User-Agent: curl/7.79.1
> Accept: application/json
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Content-Type: text/event-stream
< Vary: Origin
< X-Request-Id: 29uOq2EkWKzL0Earkk8FQ1plL7E
< Date: Tue, 31 May 2022 01:14:53 GMT
< Transfer-Encoding: chunked
< 
event: pool-new
data: {"pool_id":"0x0f231e0deb32417a497856c94d4f6fc1829816f2"}
```

However, once we advance to the commit 8b547669 that;'s associated with #3870, now it responds with `Content-Type: application/json`.

```
$ curl http://localhost:9867/version
{"version":"v1.0.0-85-g8b547669"}

$ curl -v -H 'Accept: application/json' http://localhost:9867/events
*   Trying 127.0.0.1:9867...
* Connected to localhost (127.0.0.1) port 9867 (#0)
[...]

PR #3972, opened 2022-07-07
Clarify Accept header significance on /events endpoint

philrz:
While verifying #3924, I double checked that the API docs are still accurate. They effectively are, but since the server-side events concept might not be familiar to some audiences, I figured a little extra clarity was justified.

### DIFF
--- a/docs/lake/api.md
+++ b/docs/lake/api.md
@@ -432,8 +432,8 @@ curl -X POST \
 
 Subscribe to an events feed, which returns an event stream in the format of
 [server-sent events](https://html.spec.whatwg.org/multipage/server-sent-events.html).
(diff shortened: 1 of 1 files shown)


docs/lake/api.md
```diff
@@ -432,6 +432,8 @@ curl -X POST \
 
 Subscribe to an events feed, which returns an event stream in the format of
 [server-sent events](https://html.spec.whatwg.org/multipage/server-sent-events.html).
+The MIME type specified in the request's Accept HTTP header only affects the format
+of the objects returned in the `data` fields of the event stream.
```
nwt, line 436:
```suggestion
The MIME type specified in the request's Accept HTTP header determines the format
of `data` field values in the event stream.
```

nwt (approved):

Issue #2780, opened 2021-06-02
Slow aggregation (also slower when spilled to disk)


philrz:
A community user [reported on public Slack](https://brimsec.slack.com/archives/C010DR0HHMF/p1622562315017200) that they were forming some Zed queries to show potential malware beacons. I've already provided some test results in that thread using the last GA `zq` release `v0.29.0` since that what the user had. The performance deltas reproduce with current `main` of Zed as well, so I'm collecting up some numbers and observations here for further consideration.

Repro is with Zed commit fd273d1. The original test pcap is at http://mawi.nezu.wide.ad.jp/mawi/samplepoint-F/2021/202101011400.pcap.gz however I found it took me a long time to download, so I've put a copy of it up on a public storage bucket at https://storage.googleapis.com/brimsec-public-test-data/zed-2780/202101011400.pcap.gz. It turns out there's more than 12-million flows in the pcap, so dragging it into my Brim `v0.24.0` to turn it into an `all.zng` took a very long time, so I've copied that ZNG to the same storage bucket such that it can be downloaded from https://storage.googleapis.com/brimsec-public-test-data/zed-2780/202101011400.zng.gz.

Using that ZNG now with current `zq` fd273d11, here's a baseline on my Mac laptop of something like what the user was first attempting (I dropped a `sort` they had at the end, so we can focus on _just_ the aggregation performance):

```
$ zq -version
Version: v0.29.0-366-gfd273d11

$ time zq '_path=="conn" duration < 1.0 | every 1m count(id.orig_h) by id.orig_h, id.resp_h, id.resp_p, proto, service' 202101011400.zng.gz > counts.zng

real	3m47.191s
user	4m25.353s
sys	0m6.546s
```

We find that the aggregation produced almost 11-million rows.

```
$ zq -f table 'count()' counts.zng
COUNT
10708394
```

When first hearing the user's report of slow performance, I'd thought about spill-to-disk perhaps causing the performance drag, hence thought of suggesting the knob on the aggregate function `with -limit <N>`, which makes it such that the spill-to-disk only kicks in when the number of unique aggregation keys in memory exceeds N. It defaults to 1-million (`1000000`). Not having yet seen the data and knowing the significance of the almost-11-million-rows, I happened to suggest the setting of 10-million first, which we see doesn't happen to improve the performance:

```
[...]

PR #3974, opened 2022-07-08
use default ZNG writer options where possible

nwt:
Enable compression and use the default frame threshold for most ZNG
writers, with two notable exceptions:

* runtime/op/spill.NewFile, where enabling compression decreases write
  throughput by a factor of 10 (but using the default frame threshold
  improves read throughput by a factor of 70)

* index.WriterOpts values, because the index package uses
  zio/zngio.Writer.Position in a way that requires a single value per
  ZNG frame

(The runtime/op/spill.NewFile change should help to address the spilling slowness described in #2780.)

### DIFF
--- a/index/index_test.go
+++ b/index/index_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/brimdata/zed/pkg/storage"
 	"github.com/brimdata/zed/runtime"
 	"github.com/brimdata/zed/zio"
+	"github.com/brimdata/zed/zio/zngio"
 	"github.com/brimdata/zed/zio/zsonio"
 	"github.com/brimdata/zed/zson"
 	"github.com/stretchr/testify/assert"
@@ -30,7 +31,9 @@ func TestSearch(t *testing.T) {
 {key:"key5",value:"value5"}
 {key:"key6",value:"value6"}
 `
-	finder := buildAndOpen(t, storage.NewLocalEngine(), reader(data), field.DottedList("key"), index.WriterOpts{})
+	finder := buildAndOpen(t, storage.NewLocalEngine(), reader(data), field.DottedList("key"), index.WriterOpts{
+		ZNGWriterOpts: zngio.WriterOpts{Compress: true},
+	})
 	kv, err := finder.ParseKeys(`"key2"`)
 	require.NoError(t, err)
 	rec, err := finder.Lookup(kv...)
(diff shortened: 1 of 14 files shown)


service/request.go
```diff
@@ -203,7 +204,13 @@ func (w *ResponseWriter) ZioWriter() zio.WriteCloser {
 	if w.zw == nil {
 		w.Header().Set("Content-Type", api.FormatToMediaType(w.Format))
 		var err error
-		w.zw, err = anyio.NewWriter(zio.NopCloser(w), anyio.WriterOpts{Format: w.Format})
+		w.zw, err = anyio.NewWriter(zio.NopCloser(w), anyio.WriterOpts{
+			Format: w.Format,
+			ZNG: zngio.WriterOpts{
```
mattnibs, line 209: Am wondering if this should be a pointer and if nil just does zngio.NewWriter?

index/index_test.go
```diff
@@ -47,7 +50,9 @@ func TestMicroIndex(t *testing.T) {
 	ctx := context.Background()
 	zctx := zed.NewContext()
 	engine := storage.NewLocalEngine()
-	writer, err := index.NewWriter(ctx, zctx, engine, path, field.DottedList("key"), index.WriterOpts{})
+	writer, err := index.NewWriter(ctx, zctx, engine, path, field.DottedList("key"), index.WriterOpts{
+		ZNGWriterOpts: zngio.WriterOpts{Compress: true},
```
mattnibs, line 54: Like anyio.WriterOpts am wondering if ZNGWriterOpts should be a pointer and if nil just does the normal thing?

service/request.go
nwt: Excellent idea. I created #3978 for that.

lake/index/writer.go
```diff
@@ -117,7 +117,9 @@ func newIndexer(ctx context.Context, c runtime.Compiler, engine storage.Engine,
 		return nil, err
 	}
 	keys := rule.RuleKeys()
-	writer, err := index.NewWriter(ctx, zctx, engine, object.Path(path).String(), keys, index.WriterOpts{})
+	writer, err := index.NewWriter(ctx, zctx, engine, object.Path(path).String(), keys, index.WriterOpts{
+		ZNGWriterOpts: zngio.WriterOpts{Compress: true},
```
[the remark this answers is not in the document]
nwt: Yes, eventually. But right now, index.indexWriter.write expects zngio.Writer.Position to return a meaningful value without a preceding call to zngio.Writer.EndStream, and it only does so if zngio.WriterOpts.FrameThresh is zero.

I'm working on a separate fix for that.

index/index_test.go
nwt: My preference would be to remove ZNGWriterOpts since it feels like a knob no one is going to turn.

mattnibs: I'm guessing the knob is used/necessary for testing.

nwt: I certainly agree that it's necessary for the current tests. And though I do think those tests can be made to work without it, that can wait.

What do you say to a separate PR to change index.WriterOpts.ZNGWriterOpts to a pointer?

nwt: I created #3983 for this.

lake/index/writer.go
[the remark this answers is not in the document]
nwt: #3982 tracks the incorrect use of zngio.Writer.Position in index.indexWriter.write.

PR #3984, opened 2022-07-09
add error checks to journal replay and simplify snapshots

mccanne:
This commit adds error checks to the logic that plays a log
to build a snapshot.  This change exposed some bugs that
led to the realization that the snapshot abstraction was
tracking deleted stuff when such tracking belongs in Patches
not Snaphots.  We removed the deleted objects from Snapshot
and updated all the surrounding logic to use Patches instead.

While we were at it, we changed the branch merge logic to use
a new method on Patch to compute a diff between patches and also
detect deletion conflicts.  This allowed us to remove the
OverlappingDeletes check that is better done in the Diff computation.

### DIFF
--- a/lake/branch.go
+++ b/lake/branch.go
@@ -373,14 +373,17 @@ func (b *Branch) buildMergeObject(ctx context.Context, parent *branches.Config, 
 	if err != nil {
 		return nil, err
 	}
-	if overlap := childPatch.OverlappingDeletes(parentPatch); overlap != nil {
-		//XXX add IDs of (some of the) overlaps
-		return nil, errors.New("write conflict on merge")
-	}
 	if message == "" {
 		message = fmt.Sprintf("merged %s into %s", b.Name, parent.Name)
 	}
-	return childPatch.NewCommitObject(parent.Commit, retries, author, message, zed.Value{zed.TypeNull, nil}), nil
+	// Now compute the diff between the parent patch and the child path so that
+	// the diff patch will reflect into the parent the changes in the child.
+	// Diff() will also check for delete conflicts.
+	diff, err := commits.Diff(parentPatch, childPatch)
+	if err != nil {
+		return nil, fmt.Errorf("error trying to merge %s into %s: %w", b.Name, parent.Name, err)
+	}
+	return diff.NewCommitObject(parent.Commit, retries, author, message, zed.Value{zed.TypeNull, nil}), nil
 }
 
 func commonAncestor(a, b []ksuid.KSUID) ksuid.KSUID {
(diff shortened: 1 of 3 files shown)


lake/commits/patch.go
```diff
@@ -174,16 +201,70 @@ func (p *Patch) Revert(tip *Snapshot, commit, parent ksuid.KSUID, retries int, a
...
+	for _, idx := range child.deletedIndexes {
+		if IndexExists(parent, idx.ruleID, idx.id) {
+			if err := p.DeleteIndexObject(idx.ruleID, idx.id); err != nil {
+				return nil, err
+			}
+			dirty = true
+		}
+	}
+	if !dirty {
+		return nil, errors.New("nothing to merge")
```
mattnibs, line 267: "nothing to merge" seems like the wrong error message given the function is called Diff.

mattnibs (approved):

lake/branch.go
```diff
@@ -373,14 +373,17 @@ func (b *Branch) buildMergeObject(ctx context.Context, parent *branches.Config,
...
 	if message == "" {
 		message = fmt.Sprintf("merged %s into %s", b.Name, parent.Name)
 	}
-	return childPatch.NewCommitObject(parent.Commit, retries, author, message, zed.Value{zed.TypeNull, nil}), nil
+	// Now compute the diff between the parent patch and the child path so that
+	// the diff patch will reflect into the parent the changes in the child.
+	// Diff() will also check for delete conflicts.
+	diff, err := commits.Diff(parentPatch, childPatch)
+	if err != nil {
+		return nil, fmt.Errorf("error trying to merge %s into %s: %w", b.Name, parent.Name, err)
```
nwt, line 384:
```suggestion
		return nil, fmt.Errorf("merging %s into %s: %w", b.Name, parent.Name, err)
```

lake/commits/patch.go
```diff
@@ -142,29 +155,43 @@ func (p *Patch) NewCommitObject(parent ksuid.KSUID, retries int, author, message
...
 	for _, id := range p.deletedObjects {
-		dataObject, _ := tip.LookupDeleted(id)
-		if dataObject != nil {
+		// Reach back to get the object before it was deleted in this patch.
+		dataObject, err := p.base.Lookup(id)
+		if err != nil {
+			return nil, err
+		}
+		if dataObject == nil {
+			return nil, fmt.Errorf("corrupt snapshot: id %s is in patch's deleted objects but not in patch's base snapshot", id)
```
nwt, line 172:
```suggestion
			return nil, fmt.Errorf("corrupt snapshot: ID %s is in patch's deleted objects but not in patch's base snapshot", id)
```

lake/commits/patch.go
nwt: Maybe address this by changing the function name to Merge?

mccanne: I'll do "difference is empty" then this error is decorated by the merge logic with additional context so that works nicely.

PR #3991, opened 2022-07-13
support marshaling of net.IP to Zed type ip

mccanne:
This commit adds back support for marshaling IP addresses
from the Go net package.  When we moved to the netip package,
we lost this support.

### DIFF
--- a/zson/marshal.go
+++ b/zson/marshal.go
@@ -329,6 +329,12 @@ func (m *MarshalZNGContext) encodeAny(v reflect.Value) (zed.Type, error) {
 		m.Builder.Append(zv.Bytes)
 		return typ, nil
 	}
+	if ip, ok := v.Interface().(net.IP); ok {
+		if a, err := netip.ParseAddr(ip.String()); err == nil {
+			m.Builder.Append(zed.EncodeIP(a))
+			return zed.TypeIP, nil
+		}
+	}
 	switch v.Kind() {
 	case reflect.Array:
 		if v.Type().Elem().Kind() == reflect.Uint8 {
--- a/zson/marshal_test.go
+++ b/zson/marshal_test.go
@@ -2,6 +2,7 @@ package zson_test
 
 import (
 	"bytes"
+	"net"
 	"strings"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


nwt (approved):

zson/marshal.go
```diff
@@ -329,6 +329,12 @@ func (m *MarshalZNGContext) encodeAny(v reflect.Value) (zed.Type, error) {
 		m.Builder.Append(zv.Bytes)
 		return typ, nil
 	}
+	if ip, ok := v.Interface().(net.IP); ok {
+		if a, err := netip.ParseAddr(ip.String()); err == nil {
```
nwt, line 333:
```suggestion
		if a, ok := netip.AddrFromSlice(ip); ok {
```

zson/marshal_test.go
```diff
@@ -307,3 +308,9 @@ func TestIgnoreField(t *testing.T) {
 	require.NoError(t, zson.Unmarshal(b, &v))
 	assert.Equal(t, s{Value: "test"}, v)
 }
+
+func TestMarshalNetIP(t *testing.T) {
+	b, err := zson.Marshal(net.ParseIP("10.0.0.1"))
+	require.NoError(t, err)
+	assert.Equal(t, `10.0.0.1`, b)
```
nwt, line 315:
```suggestion
	// A successful net.ParseIP call always returns an IP address in 16-byte
	// form, effectively converting IPv4 dotted-quad inputs to IPv4-mapped
	// IPv6 outputs.
	s, err := zson.Marshal(net.ParseIP("10.0.0.1"))
	require.NoError(t, err)
	assert.Equal(t, "::ffff:10.0.0.1", s)

	s, err = zson.Marshal(net.ParseIP("10.0.0.1").To4())
	require.NoError(t, err)
	assert.Equal(t, "10.0.0.1" , s)
```

mccanne:
I'm going to commit this as is and create an issue for @nwt's comments.  The problem with these changes is the marshaler will always create an IPv6 Zed ip value whereas the path through the string will create a Zed IPv4 address when it can.

PR #3995, opened 2022-07-13
fix bugs in IP marshaling

mccanne:
This commit fixes bugs in the IP marshaling code where Go types
net.IP and netip.Addr were being decorated when they should just
convert to native Zed ip type.  Also, we added suport for unmarshaling
net.IP in addition to netip.Addr.  Test coverage was extended to
cover these use cases.

### DIFF
--- a/zson/marshal_test.go
+++ b/zson/marshal_test.go
@@ -3,6 +3,7 @@ package zson_test
 import (
 	"bytes"
 	"net"
+	"net/netip"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


zson/marshal_test.go
```diff
@@ -322,3 +339,16 @@ func TestMarshalGoTime(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, `2006-01-02T15:04:05.123Z`, b)
 }
+
+func TestMarshalDecoratedIPs(t *testing.T) {
```
nwt, line 343: Nit: Put this between TestMarshalNetipAddr and TestMarshalGoTime.

nwt (approved):

PR #3999, opened 2022-07-14
lake/api: handle errors in remote.Load

nwt:
The goroutine in remote.Load ignores errors.  Handle them by passing
them to io.PipeWriter.CloseWithError instead.


lake/api/remote.go
```diff
@@ -95,9 +95,12 @@ func (r *remote) RenamePool(ctx context.Context, pool ksuid.KSUID, name string)
 func (r *remote) Load(ctx context.Context, _ *zed.Context, poolID ksuid.KSUID, branchName string, reader zio.Reader, commit api.CommitMessage) (ksuid.KSUID, error) {
 	pr, pw := io.Pipe()
 	go func() {
-		w := zngio.NewWriter(pw)
-		zio.CopyWithContext(ctx, w, reader)
-		w.Close()
+		w := zngio.NewWriter(zio.NopCloser(pw))
```
mattnibs, line 98: What's the reason for wrapping this in a NopCloser?

mattnibs (approved):

PR #4001, opened 2022-07-15
service: Add back query ctrl flag

### LINKED ISSUE #3990: Make /query response formats consistent & add option to include control messages
This issue is being filed as of Zed commit e0683b10.

As discussed in https://github.com/brimdata/brim/issues/2426, we've recently become aware of some inconsistency in the format of responses from the Zed lake API's `/query` endpoint. Specifically, when requesting MIME types `application/x-zng` or `application/x-zjson` via the `Accept` header, in addition to the actual query response data being returned in that format, control messages are also included. For other formats, just the response data of the requested format is returned.

An example with Zed commit e0683b10 (please excuse the binary noise due to ZNG):

```
$ zed -version
Version: v1.1.0-60-ge0683b10

$ zed query -use foo '*'
{hello:"world"}

$ curl --output - -X POST  -H 'Accept: application/json' -H 'Content-Type: application/json' http://localhost:9867/query -d '{"query":"from foo@main"}'
[{"hello":"world"}]

$ curl --output - -X POST  -H 'Accept: application/x-zng' -H 'Content-Type: application/json' http://localhost:9867/query -d '{"query":"from foo@main"}'
"{channel_id:0}(=QueryChannelSet)
	helloworld"{channel_id:0}(=QueryChannelEnd)
`	?? {start_time:2022-07-12T22:57:06.135226Z,update(	?6558Z,Progress:?matched:6,record1!@1}(=J?ress)}(=QueryStats)
```

This inconsistency didn't really cause a problem in the past because the Brim/Zui app and `zed` CLI tool were the only clients hitting these endpoints, and there was a `?noctrl=true` option these clients could use to request the response data without the control messages. However, this option happened to disappear due to some other changes, and also the API now has user-facing documentation at https://zed.brimdata.io/docs/next/lake/api/#query.

The team reached consensus that the best way to make this consistent & clean will be:

1. No control messages by default in the `/query` responses
2. Add an option to `/query` so a client can request the control messages when desired
3. Have the Brim/Zui app and `zed` CLI command invoke the option when appropriate

mattnibs:
By default have the query endpoint return no control messages for
supporting formats (zng and zjson). If the query params ctrl is set to
true enable control messages. For the go client enable control messages
on queries by default.

Closes #3990

### DIFF
--- a/api/queryio/writer.go
+++ b/api/queryio/writer.go
@@ -10,8 +10,6 @@ import (
 	"github.com/brimdata/zed/zio"
 	"github.com/brimdata/zed/zio/anyio"
 	"github.com/brimdata/zed/zio/jsonio"
-	"github.com/brimdata/zed/zio/zjsonio"
-	"github.com/brimdata/zed/zio/zngio"
 )
 
 type controlWriter interface {
@@ -23,29 +21,23 @@ type Writer struct {
 	cid     int
 	start   nano.Ts
 	writer  zio.WriteCloser
+	ctrl    bool
 	flusher http.Flusher
 }
 
 func NewWriter(w io.WriteCloser, format string, flusher http.Flusher, ctrl bool) (*Writer, error) {
 	d := &Writer{
 		cid:     -1,
+		ctrl:    ctrl,
 		start:   nano.Now(),
 		flusher: flusher,
 	}
 	var err error
 	switch format {
 	case "zng":
-		if ctrl {
-			d.writer = NewZNGWriter(w)
-		} else {
-			d.writer = zngio.NewWriter(w)
-		}
+		d.writer = NewZNGWriter(w)
 	case "zjson":
-		if ctrl {
-			d.writer = NewZJSONWriter(w)
-		} else {
-			d.writer = zjsonio.NewWriter(w)
-		}
+		d.writer = NewZJSONWriter(w)
 	case "json":
 		// A JSON response is always an array.
 		d.writer = jsonio.NewArrayWriter(w)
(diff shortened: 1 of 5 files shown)


api/queryio/writer.go
```diff
@@ -33,9 +35,17 @@ func NewWriter(w io.WriteCloser, format string, flusher http.Flusher) (*Writer,
...
+			d.writer = NewZNGWriter(w)
+		} else {
+			d.writer = zngio.NewWriter(w)
+		}
 	case "zjson":
-		d.writer = NewZJSONWriter(w)
+		if ctrl {
+			d.writer = NewZJSONWriter(w)
+		} else {
+			d.writer = zjsonio.NewWriter(w)
```
nwt, line 47: I think what's going on here would be substantially clearer if we added a `ctrl bool` field to `Writer` and then checked it in `WriteControl`.

service/handlers.go
```diff
@@ -29,6 +30,14 @@ func handleQuery(c *Core, w *ResponseWriter, r *Request) {
...
 		return
 	}
+	var ctrl bool
+	if s := r.URL.Query().Get("ctrl"); s != "" {
+		var err error
+		if ctrl, err = strconv.ParseBool(s); err != nil {
+			w.Error(srverr.ErrInvalid("non-boolean value for \"ctrl\" query parameter: %q", s))
+			return
+		}
+	}
```
nwt, line 40:
```suggestion
	ctrl, ok := r.BoolFromQuery("ctrl")
	if !ok {
		return
	}
```

api/queryio/writer.go
mattnibs: I can dig it 👍🏻

nwt (approved):

PR #4009, opened 2022-07-19
Include ctrl URL param in query API doc

philrz:
Now that #4001 has brought back a `ctrl` URL param for the `/query` endpoint, I figure we might as well call it in the doc.

This looks like the first time we're documenting a URL param, so feel free to critique the style as I'd intend to follow the pattern for future URL params.

### DIFF
--- a/docs/lake/api.md
+++ b/docs/lake/api.md
@@ -423,6 +423,7 @@ POST /query
 | query | string | body | Zed query to execute. All data is returned if not specified. ||
 | head.pool | string | body | Pool to query against Not required if pool is specified in query. |
 | head.branch | string | body | Branch to query against. Defaults to "main". |
+| ctrl | string | url | Set to "T" to include control messages in ZNG or ZJSON responses. Defaults to "F". |
 
 **Example Request**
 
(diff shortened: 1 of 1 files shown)


docs/lake/api.md
```diff
@@ -423,6 +423,7 @@ POST /query
 | query | string | body | Zed query to execute. All data is returned if not specified. ||
 | head.pool | string | body | Pool to query against Not required if pool is specified in query. |
 | head.branch | string | body | Branch to query against. Defaults to "main". |
+| ctrl | string | url | Set to "T" to include control messages in ZNG or ZJSON responses. Defaults to "F". |
```
nwt, line 426: RFC 3986 calls this the query component of a URI.
```suggestion
| ctrl | string | query | Set to "T" to include control messages in ZNG or ZJSON responses. Defaults to "F". |
```

nwt (approved):

docs/lake/api.md
philrz: Thanks @nwt. I must admit I'm slightly conflicted about that because "query" could potentially be mistaken as a reference a Zed query. But the example provides additional clarity so I'm happy to take the suggestion.

PR #4028, opened 2022-07-22
Fix anchor links between docs

philrz:
The new coverage from our markdown link checker automation described in #4004 finds broken links for missing anchor destinations within the same markdown doc, but it still doesn't catch broken links to anchors within _other_ docs. When clicked, those bad links still just successfully end up at the top-level page, which has made these difficult to catch/fix. But I've been working on getting the ruzickap/action-my-broken-link-checker Action (the same one already used on the brimdata.io web site) to check for broken links on the deployed docs site https://zed.brimdata.io/, and thankfully that _does_ pick up on such breakages. So here I'm trying to fix those back at the origin. I guess once these changes are approved/merged I'll want to backport this to the older versioned doc revs. Fun times!

### DIFF
--- a/docs/commands/zq.md
+++ b/docs/commands/zq.md
@@ -88,7 +88,7 @@ emits
 ```mdtest-output
 2
 ```
-Note here that the query `1+1` [implies](../language/README.md#26-implied-operators)
+Note here that the query `1+1` [implies](../language/overview.md#26-implied-operators)
 `yield 1+1`.
 
 ## 2. Input Formats
@@ -351,7 +351,7 @@ If you are ever stumped about how the `zq` compiler is parsing your query,
 you can always run `zq -C` to compile and display your query in canonical form
 without running it.
 This can be especially handy when you are learning the language and
-[its shortcuts](../language/README.md#26-implied-operators).
+[its shortcuts](../language/overview.md#26-implied-operators).
 
 For example, this query
 ```mdtest-command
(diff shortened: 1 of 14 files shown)


docs/commands/zq.md
```diff
@@ -421,7 +421,7 @@ echo <values> | zq <query> -
 which is used throughout the [language documentation](../language/README.md)
 and [operator reference](../language/operators/README.md).
 
-The language documentation and [tutorials directory](../tutorials)
+The language documentation and [tutorials directory](../tutorials/README.md)
```
philrz, line 424: These link destinations that just point to a directory work ok when browsing the docs within GitHub itself, but on the deployed site they produce a "Page Not Found" error when clicked.

docs/commands/zq.md
```diff
@@ -582,7 +582,7 @@ Next, a JSON file can be converted from ZNG using:
 zq -f json conn.zng > conn.json
 ```
 Note here that we lose information in this conversion because the rich data types
-of Zed (that were [translated from the Zeek format](../../zeek/Data-Type-Compatibility.md)) are lost.
+of Zed (that were [translated from the Zeek format](https://github.com/brimdata/zed/blob/main/zeek/Data-Type-Compatibility.md)) are lost.
```
philrz, line 585: These kinds of relative links to things outside the `/docs` directory work fine when browsing within GitHub itself, but they understandably break on the deployed docs site. I've therefore fixed them by using an absolute URL.

nwt (approved):

Issue #4012, opened 2022-07-19
finish union support in zson/zng marshaler


mccanne:
The recent reddit thread exposed a few problems in the marshaling implementation for unions.  In particular, unmarshaling a ZSON array of union types can only work in Go for interface values so the unmarshaler needs to create an array of empty interface values of the proper type in which to unmarshal (or the logic might want to be turned around where the concrete values are created from below then fit into an array of interface values).

PR #4007 was a first attempt at this.  While we got the marshal direction much improved for unions, we encountered some complexity in the unmarshal path (as described above).  It's probably not too difficult to get this right but I didn't find the solution and a day's work and need to get back to higher priority items.

So, we're closing #4007 and we'll leave the `fix-array-marshal` branch in the tree to get picked up later.  That branch has a failing test that illustrates the problem.

Moving this to Back Burner for now.

mccanne:
We're merging branch `fix-array-marshal` in an upcoming PR as it is needed by the upcoming ZST changes and partially addresses this issue.

mccanne:
Added a TestSimpleUnionUnmarshal in branch `fix-array-marshal` which is skipped and should be addressed in this issue.

mccanne:
To complete this issue we should add a recursive test (where implied union is an array of unions of unions) and a "combinatoric" test, where there are multiple interface values with combos otherwise creating combinatoric problems but I think Zed solve this. becomes the variations become unions.  That said, we can get combinatoric  behavior when all the type mixtures don't appear in one array so we should have a way to declare the complete union as part of the Go package, like  Bind but upfront declarations of what bindings go in what unions.

philrz:
@mccanne confirmed there's some overlap between this issue and #4012.

PR #4029, opened 2022-07-24
add (partial) support for marshaling mixed-type arrays and zed.Type

mccanne:
This commit partially fixes a known bug where mixed-type arrays
are erroneously marshaled into zng.  The fix is to detect when the
array has multiple types.  This is tricky because the type of
each element is only determined upon a recursive descent which
also builds the value.  So, we opportunistically go this route
with a temp builder and track all the types encounted.  If there
is more than one type, we throw  out the improperly built array
and recode it as an array of unions.

This solution currently works when the mixed-type array elements
are Zed named types, allowing the decoder to allocate the proper
concrete type for each underlying mixed type.  We're leaving
issue #4012 open to handle the case where the concrete types
can be recursively determined without the Zed named types.

We also added some tests that are skipped to be addressed by #4012
and added a test to cover the case where an interface value
points to a zero-valued concrete object.

Finally, we added support for marshaling type values to/from
interface values of zed.Type.  This is needed by the upcoming
changes to the ZST metadata implementation.

### DIFF
--- a/zson/marshal.go
+++ b/zson/marshal.go
@@ -558,20 +558,15 @@ func (m *MarshalZNGContext) encodeArray(arrayVal reflect.Value) (zed.Type, error
 		// so we can recode the array with tagged union elements.
 		// We throw out the array computed above and start over with
 		// an empty builder.
-		m.Builder = *zcode.NewBuilder()
 		m.Builder.Truncate()
 		for i, typ := range types {
 			m.Builder.BeginContainer()
 			tag := unionType.TagOf(typ)
 			m.Builder.Append(zed.EncodeInt(int64(tag)))
 			item := arrayVal.Index(i)
-			itemType, err := m.encodeValue(item)
+			_, err := m.encodeValue(item)
 			if err != nil {
 				return nil, err
-			}
-			tagType, _ := unionType.Type(tag)
-			if typ != tagType {
-				return nil, fmt.Errorf("internal error: type mismatch in MarshalZNGContext.encodeArray: %s does not match %s; %p %p", String(typ), String(itemType), typ, itemType)
 			}
 			m.Builder.EndContainer()
 		}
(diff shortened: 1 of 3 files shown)


zson/marshal.go
```diff
@@ -514,27 +519,66 @@ func (m *MarshalZNGContext) encodeArrayBytes(arrayVal reflect.Value) (zed.Type,
...
+		for typ := range typeMap {
+			unionTypes = append(unionTypes, typ)
+		}
+		unionType := m.Context.LookupTypeUnion(unionTypes)
+		// Now we know all the types and have the union type
+		// so we can recode the array with tagged union elements.
+		// We throw out the array computed above and start over with
+		// an empty builder.
+		m.Builder = *zcode.NewBuilder()
+		m.Builder.Truncate()
```
nwt, line 562: No need for a new builder; Truncate's sufficient.
```suggestion
		m.Builder.Truncate()
```

zson/marshal.go
```diff
@@ -514,27 +519,66 @@ func (m *MarshalZNGContext) encodeArrayBytes(arrayVal reflect.Value) (zed.Type,
...
+			m.Builder.BeginContainer()
+			tag := unionType.TagOf(typ)
+			m.Builder.Append(zed.EncodeInt(int64(tag)))
+			item := arrayVal.Index(i)
+			itemType, err := m.encodeValue(item)
+			if err != nil {
+				return nil, err
+			}
+			tagType, _ := unionType.Type(tag)
+			if typ != tagType {
```
nwt, line 573: Nit: Doesn't seem like there's any reason not to use itemType here.
```suggestion
			if typ != itemType {
```

zson/marshal.go
```diff
@@ -906,9 +970,13 @@ func (u *UnmarshalZNGContext) decodeMap(zv *zed.Value, mapVal reflect.Value) err
 }
 
 func (u *UnmarshalZNGContext) decodeRecord(zv *zed.Value, sval reflect.Value) error {
+	if union, ok := zv.Type.(*zed.TypeUnion); ok {
```
nwt, line 973:
```suggestion
	if union, ok := zed.TypeUnder(zv.Type).(*zed.TypeUnion); ok {
```

zson/marshal.go
```diff
@@ -1095,33 +1167,72 @@ func (u *UnmarshalZNGContext) lookupType(typ zed.Type) (reflect.Type, error) {
...
+		}
+		elemType, err := u.lookupGoType(typ.Type, bytes)
 		if err != nil {
 			return nil, err
 		}
 		return reflect.SliceOf(elemType), nil
-	case *zed.TypeUnion, *zed.TypeEnum:
+	case *zed.TypeUnion:
+		t, bytes := typ.Untag(bytes)
+		return u.lookupGoType(t, bytes)
```
nwt, line 1216:
```suggestion
		return u.lookupGoType(typ.Untag(bytes))
```

nwt (approved):

PR #4032, opened 2022-07-26
fix bug in EncodeTypeValue to honor DFS order of named bindings

mccanne:
This commit fixes a bug where a child typedef would override the
parent typedef violating the DFS ordering semantics of name bindings
that refer to the same name.  This fix is to simply update the
typedefs map after the child is traversed.

### DIFF
--- a/zed_test.go
+++ b/zed_test.go
@@ -11,7 +11,9 @@ import (
 	"github.com/brimdata/zed"
 	"github.com/brimdata/zed/zio/anyio"
 	"github.com/brimdata/zed/zio/parquetio"
+	"github.com/brimdata/zed/zson"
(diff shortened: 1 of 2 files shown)


zed_test.go
```diff
@@ -238,3 +240,18 @@ func findInputs(t *testing.T, dirs map[string]struct{}, script string, isValidIn
 	}
 	return out, nil
 }
+
+func TestTranslateNameConflictUnion(t *testing.T) {
+	// This test confirms that a union with complicated type renamig is properly
```
nwt, line 245:
```suggestion
	// This test confirms that a union with complicated type renaming is properly
```

nwt (approved):

PR #4033, opened 2022-07-26
fix zson formatter bug with redefined named types

mccanne:
The zson formatter did not properly track type name bindings.
The data structure was backwards: you need to keep track
of the binding from a name to a type so that when bindings
change in DFS order they can be tracked by their name.

This came up in unions where the named type could be redefined
in a child of a previous type and thus needed to be rebound
instead of referring erronously to the child type.

### DIFF
--- a/zson/formatter.go
+++ b/zson/formatter.go
@@ -15,8 +15,8 @@ import (
 )
 
 type Formatter struct {
-	typedefs  typemap
-	permanent typemap
+	typedefs  map[string]*zed.TypeNamed
+	permanent map[string]*zed.TypeNamed
 	persist   *regexp.Regexp
 	tab       int
 	newline   string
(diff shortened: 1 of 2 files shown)


zson/formatter.go
```diff
@@ -117,26 +117,46 @@ func (f *Formatter) Format(zv *zed.Value) (string, error) {
...
+		if _, ok = f.permanent[named.Name]; ok {
+			return true
+		}
+	}
+	return false
+}
+
+func (f *Formatter) exists(typ zed.Type) bool {
+	return false
 }
```
nwt, line 137: This is unused.
```suggestion
```

zson/formatter.go
```diff
@@ -117,26 +117,46 @@ func (f *Formatter) Format(zv *zed.Value) (string, error) {
...
-		ok = f.permanent.exists(typ)
+	named, ok := typ.(*zed.TypeNamed)
+	if !ok {
+		return false
+	}
+	if _, ok := f.typedefs[named.Name]; ok {
+		return true
 	}
-	return ok
+	if f.persist != nil {
```
nwt, line 127: Nit: Test f.permanent since that's the field that this guards.
```suggestion
	if f.permanent != nil {
```

nwt (approved):

PR #4035, opened 2022-07-26
update ZST format to use marshaled metadata

mccanne:
This commit updates the ZST format to use marshaling instead of
hand-coded zed values in the metadata section.  To do so, we created
a new set of Go metadata structures that are simply marshaled into
the metadata section of the ZST file.  We bumped the version number
to reflect the change in format.

In subsequent PRs, we will change the name of package zst/column
to zst/vector and update the ZST specification.


zst/cutter.go
```diff
@@ -42,50 +41,34 @@ type CutAssembler struct {
...
-			// aren't records but all this will be subsumed when we
-			// work on predicate pushdown, so no sense spending time
-			// on this right now.
-			return nil, fmt.Errorf("ZST cut requires all top-level records to be records: encountered type %s", zson.FormatType(typ))
-		}
-		reader, err := column.NewRecordReader(recType, a.maps[k], object.seeker)
-		if err != nil {
-			return nil, err
+func NewCutAssembler(zctx *zed.Context, path field.Path, object *Object) (*CutAssembler, error) {
+	root := column.NewIntReader(object.root, object.seeker)
```
nwt, line 53: Nit: This looks like it's used only in the CutAssembler composite literal, so move it down there.

zst/cutter.go
```diff
@@ -138,20 +94,12 @@ func (a *CutAssembler) Read() (*zed.Value, error) {
 			// Skip records that don't have the field we're cutting.
 			continue
 		}
-		for n := cut.depth; n > 0; n-- {
-			a.builder.BeginContainer()
-		}
-		err = cut.reader.Read(&a.builder)
-		if err != nil {
+		if err = cut.reader.Read(&a.builder); err != nil {
```
nwt, line 97:
```suggestion
		if err := cut.reader.Read(&a.builder); err != nil {
```

zst/object.go
```diff
@@ -10,20 +10,12 @@
...
-// and introspection), an Assembly is created from the Object then zng records
-// are read from the assembly, which implements zio.Reader.  The Assembly
-// keeps track of where each column is, which is why you need a separate
-// Assembly per scan.
-//
-// You can have multiple Assembly's referring to one Object as once an
-// object is created, it's state never changes.  That said, each assembly
-// will issue reads to the underlying storage object and the read pattern
-// may create performance issues.
+// and introspection), a Metadata instance is created from the Object then zng records
```
nwt, line 13:
```suggestion
// and introspection), a Metadata instance is created from the Object and then Zed records
```

zst/object.go
```diff
@@ -112,47 +108,38 @@ func (o *Object) IsEmpty() bool {
...
+	if err != nil {
+		return err
+	}
+	u := zson.NewZNGUnmarshaler()
+	u.SetContext(o.zctx)
+	u.Bind(column.Template...)
+	if err := u.Unmarshal(val, &o.root); err != nil {
+		return err
+	}
+	// The rest of the values column.Metadata one for each
```
nwt, line 126:
```suggestion
	// The rest of the values are column.Metadata, one for each
```

zst/stream.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/brimdata/zed/zcode"
+	"github.com/brimdata/zed/zio"
+	"github.com/brimdata/zed/zst/column"
+)
+
+var ErrBadTypeNumber = errors.New("bad type number in ZST root reassembly map")
+
+// Stream reads a columnar ZST object to generate a stream of zed.Values.
+// It also has methods to read metadata for test and debugging.
+type Stream struct {
```
nwt, line 18: Nit: I'd just fold this into Reader since I don't see the benefit to keeping it separate.

zst/stream.go
```diff
@@ -0,0 +1,57 @@
...
+	s.builder.Reset()
+	typeNo, err := s.root.Read()
+	if err == io.EOF {
+		return nil, nil
+	}
+	if typeNo < 0 || int(typeNo) >= len(s.readers) {
+		return nil, ErrBadTypeNumber
+	}
+	tr := s.readers[typeNo]
+	if err = tr.reader.Read(&s.builder); err != nil {
```
nwt, line 53:
```suggestion
	if err := tr.reader.Read(&s.builder); err != nil {
```

nwt (approved):
Much nicer this way!

zst/column/int.go
```diff
@@ -11,7 +11,7 @@ type IntWriter struct {
 }
 
 func NewIntWriter(spiller *Spiller) *IntWriter {
-	return &IntWriter{*NewPrimitiveWriter(spiller)}
+	return &IntWriter{*NewPrimitiveWriter(zed.TypeInt32, spiller)}
 }
 
 func (p *IntWriter) Write(v int32) error {
```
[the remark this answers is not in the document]
mccanne: I will make every consistent and explicit as Int64

zst/object.go
mccanne: Deleting this paragraph as it's no longer accurate.

zst/stream.go
mccanne: Ok, cleaned this all up.

PR #4038, opened 2022-07-27
simplify zst to use io.ReaderAt instead of storage.Seeker

mccanne:
This commit simplifies the zst package to use io.ReaderAt
throughout and get rid of storage.Seeker.  We also noticed
that the check in zstio for a seekable input did not work
and instead failed on the first call to ReadAt (which gives
an error for stdin).  We fixed this and added a test to
make sure stdin fails properly.

### DIFF
--- a/zio/zstio/reader.go
+++ b/zio/zstio/reader.go
@@ -11,12 +11,15 @@ import (
 
 func NewReader(zctx *zed.Context, r io.Reader) (*zst.Reader, error) {
 	reader, ok := r.(storage.Reader)
-	if !ok {
-		return nil, errors.New("zst must be used with a seekable input")
+	if ok {
+		if !storage.IsSeekable(reader) {
+			return nil, errors.New("zst must be used with a seekable input")
+		}
+		return zst.NewReaderFromStorageReader(zctx, reader)
 	}
-	seeker, err := storage.NewSeeker(reader)
-	if err != nil {
-		return nil, errors.New("zst must be used with a seekable input")
-	}
-	return zst.NewReaderFromSeeker(zctx, seeker)
+	// This can't be the zed system (which always using package storage)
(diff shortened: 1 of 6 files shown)


zst/object.go
```diff
@@ -59,36 +59,33 @@ func NewObject(zctx *zed.Context, s *storage.Seeker, size int64) (*Object, error
...
+	r, err := engine.Get(ctx, uri)
 	if err != nil {
 		return nil, err
 	}
-	o, err := NewObject(zctx, seeker, size)
-	if err == nil {
-		o.closer = r
-	}
-	return o, err
+	return NewObjectFromStorageReader(zctx, r)
```
nwt, line 88: We should close the storage.Reader on error.
```suggestion
	object, err := NewObjectFromStorageReader(zctx, r)
	if err != nil {
		r.Close()
		return nil, err
	}
	return object, nil
```

zio/zstio/reader.go
```diff
@@ -11,12 +11,15 @@ import (
...
-	seeker, err := storage.NewSeeker(reader)
-	if err != nil {
-		return nil, errors.New("zst must be used with a seekable input")
-	}
-	return zst.NewReaderFromSeeker(zctx, seeker)
+	// This can't be the zed system (which always using package storage)
+	// so it must be a third party using he zst library.  We could assert
+	// for io.ReaderAt and io.Seeker and use the seeker to get the size
+	// and make a zst reader this way, but for now, we just say this is not supported.
+	return nil, errors.New("zst does not yet support non-storage package implementations")
```
nwt, line 24: This part's hard to love, but I say we merge it as is and see about cleaning it up later.

nwt (approved):

zio/zstio/reader.go
mccanne: :-)

PR #4045, opened 2022-08-10
manage monitor command

mattnibs:
Add zed manage monitor command- a long running process that monitors the
pools in a zed lake and executes opportunistic compactions as data is
loaded into the various pools.

### DIFF
--- a/api/client/events.go
+++ b/api/client/events.go
@@ -1,7 +1,6 @@
 package client
 
 import (
-	"errors"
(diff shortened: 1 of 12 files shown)


mattnibs:
TODOs:
- [ ] Test and observe this actually helps the live ingest case.
- [ ] Recover when an error occurs whilst compacting/scanning a pool.

nwt (approved):

PR #4070, opened 2022-09-01
Fix ZSON doc Type Decorator examples

### LINKED ISSUE #4056: Incorrect union type decorator syntax shown in ZSON spec
Repro is with Zed commit `2d52400`.

Section [2.2 Type Decorators](https://zed.brimdata.io/docs/next/formats/zson/#22-type-decorators) in the ZSON spec shows example union decorators that are currently parse errors.

```
$ zq -version
Version: v1.2.0-29-g2d52400f

$ echo '123 (float64) (int64,float64)' | zq -i zson -z -
stdio:stdin: parse error: mismatched parentheses while parsing type decorator

$ echo '123.0 (int64,float64)' | zq -i zson -z -
stdio:stdin: parse error: mismatched parentheses while parsing type decorator
```

@nwt pointed out that the correct way would use double parens, since the outer set defines a type decorator while the inner describes the union type.

```
$ echo '123.0 ((int64,float64))' | zq -i zson -z -
123.((int64,float64))

$ echo '123 (float64) ((int64,float64))' | zq -i zson -z -
stdio:stdin: type mismatch: "int64" cannot be used as "float64"
```

I guess that second one now becomes a separate issue, since even with the non-union example:

```
$ echo '123 (float64)' | zq -i zson -z -
stdio:stdin: type mismatch: "int64" cannot be used as "float64"
```

If that one's not a bug, perhaps we could change the example to something like this...?

```
$ echo '123.0 (float32) ((int64,float32,float64))' | zq -i zson -z -
123.(float32)((int64,float32,float64))
```

In any case, it seems the spec needs an update.

philrz:
These now work and come back the same when passed through `zq`.

```
$ zq -version
Version: v1.2.0-38-gd07901b2

$ echo '123.0 (float32) ((int64,float32,float64))' | zq -z -
123.(float32)((int64,float32,float64))

$ echo '123.0 ((int64,float64))' | zq -z -
123.((int64,float64))
```

Also, I paused briefly at this part:

---
image

---

...and wondered if it needed the second set of parens inside the `[ ]`, but I don't think so. Since the rightmost `<type>` is a union type as the text says, it could come with its own set of extra parens, or lack them if it was a reference to a named type. Just figured I'd point it out.

Fixes #4056

### DIFF
--- a/docs/formats/zson.md
+++ b/docs/formats/zson.md
@@ -52,11 +52,11 @@ For union values, multiple decorators might be
 required to distinguish the union-member type from the possible set of
 union types when there is ambiguity, as in
 ```
-123.0 (float32) ((int64,float32,float64))
+123. (float32) ((int64,float32,float64))
 ```
 In contrast, this union value is unambiguous:
 ```
-123.0 ((int64,float64))
+123. ((int64,float64))
 ```
 
 The syntax of a union value decorator is
--- a/docs/formats/zson.md
+++ b/docs/formats/zson.md
@@ -52,11 +52,11 @@ For union values, multiple decorators might be
 required to distinguish the union-member type from the possible set of
 union types when there is ambiguity, as in
 ```
-123 (float64) (int64,float64)
+123.0 (float32) ((int64,float32,float64))
 ```
 In contrast, this union value is unambiguous:
 ```
-123.0 (int64,float64)
+123.0 ((int64,float64))
 ```
 
 The syntax of a union value decorator is


docs/formats/zson.md
```diff
@@ -52,11 +52,11 @@ For union values, multiple decorators might be
 required to distinguish the union-member type from the possible set of
 union types when there is ambiguity, as in
 ```
-123 (float64) (int64,float64)
+123.0 (float32) ((int64,float32,float64))
```
mattnibs, line 55: ZSON doesn't print the 0 here:
```suggestion
123. (float32) ((int64,float32,float64))
```

philrz: @mattnibs: Indeed, the current ZSON implementation in the Zed tooling does not print the trailing `0`. At times I've heard @mccanne make the distinction between what the spec says and what our implementation does and imply they don't always have to match 100%. For instance, there's an existing example already in the text just a little further down that shows it with the `123.0` so I assumed that was intentional/harmless and left it intact, hence made the example here match that one.  If @mccanne or @nwt have an opinion perhaps they can chime in. If not, I'm happy to change both to just `123.`. so they both match the current implementation.

mccanne (approved):

docs/formats/zson.md
philrz: I checked in with @nwt to see if he had an opinion. He would have been fine leaving the trailing zeros but is also fine dropping them. Therefore I've dropped in both examples so they both now say `123.`.

PR #4071, opened 2022-09-01
Adding array containment example

iloveitaly:
It's not obvious that this is possible from the existing documentation


philrz (approved):
Thank you, @iloveitaly!

PR #4079, opened 2022-09-09
fix "fork (=> ... => ...) | head" hang

### LINKED ISSUE #4013: Adding "head" after a forked/merged query causes a hang
Repro is with Zed commit 3953933.

I was attempting to use a forked query in the Brim app, but it seemed to hang. It didn't hang when I executed the same query against the same lake using `zed query`.

Looking in Wireshark, I could see the app tacks on `| { i: count(), v: this} | i > 0 | head 500` at the end of queries to help with pagination, but in this case the `head` is causing a hang.

Below is a simple repro at the CLI using `zq`. Here's an initial program that works fine:

```
$ zq -version
Version: v1.2.0-7-g39539336

$ seq 10 | zq -z '
fork (
  => this % 2 == 0 | yield {evencount: count() }
  => this % 2 == 1 | total_odds := count()
)' -
{evencount:1(uint64)}
{evencount:2(uint64)}
{evencount:3(uint64)}
{evencount:4(uint64)}
{evencount:5(uint64)}
{total_odds:5(uint64)}
```

If I up the sequence count, it, still runs fine.

```
$ seq 10000 | zq -z '
fork (
  => this % 2 == 0 | yield {evencount: count() }
  => this % 2 == 1 | total_odds := count()
)' -
...
{evencount:4999(uint64)}
{evencount:5000(uint64)}
{total_odds:5000(uint64)}
```

However, if I add a `head`, it prints partial results and then hangs indefinitely, as if perhaps deadlocked.

```
$ seq 10000 | zq -z '
fork (
  => this % 2 == 0 | yield {evencount: count() }
  => this % 2 == 1 | total_odds := count()
) | head 500' -

{evencount:448(uint64)}
{evencount:449(uint64)}
{evencount:450(uint64)}
```

nwt:
"fork (=> ... => ...) | head" with one leg that pulls until EOS and one that does not will hang on an input containing multiple batches.  Fix in runtime/op/combine.Proc.propagateDone and runtime/op.Router.sendEOS.

Closes #4013.

### DIFF
--- a/runtime/op/combine/combine.go
+++ b/runtime/op/combine/combine.go
@@ -108,25 +108,34 @@ func (p *Proc) block(parent *puller) {
 }
 
 func (p *Proc) propagateDone() error {
+	var wg sync.WaitGroup
 	for _, parent := range p.parents {
 		if parent.blocked {
 			continue
 		}
-	again:
-		select {
-		case <-p.queue:
-			// If a parent is waiting on the queue, we need to
-			// read the queue to avoid deadlock.  Since we
-			// are going to throw away the batch anyway, we can
-			// simply ignore which parent it is as we will hit all
-			// of them eventually as we loop over each unblocked parent.
-			goto again
-		case parent.doneCh <- struct{}{}:
-			p.block(parent)
-		case <-p.ctx.Done():
-			return p.ctx.Err()
-		}
+		parent := parent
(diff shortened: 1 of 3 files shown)


mccanne (approved):
Ugh

PR #4089, opened 2022-09-15
Clean up summarize doc

### LINKED ISSUE #3868: Docs: summarize vs sum
https://zed.brimdata.io/docs/language/operators/summarize/

These docs reference `sum` but don't give an example of using `summarize`. To someone new to the project it is unclear if these are the same or different.

philrz:
In #3868, a community user helpfully let us know that the `summarize` doc may be confusing to new users because it made ample reference to the `sum` aggregate function while not making use of the explicit `summarize` keyword in the examples. I've tried to make this all clearer with the changes in this PR, and cleaned up a couple other things with the doc while I was in there.

Closes #3868

### DIFF
--- a/docs/language/operators/summarize.md
+++ b/docs/language/operators/summarize.md
@@ -5,26 +5,29 @@
 ### Synopsis
 
 ```
-summarize [<field>:=]<agg> [where <expr>][, [<field>:=]<agg> [where <expr>]] [by [<field>][:=<expr>] ...]
+summarize [<field>:=]<agg> [where <expr>][, [<field>:=]<agg> [where <expr>] ...] [by [<field>][:=<expr>] ...]
 ```
 ### Description
 
-The `summarize` operator consumes all of its input, applies an aggregate function
+The `summarize` operator consumes all of its input, applies an [aggregate function](../aggregates/README.md)
 to each input value optionally organized with the group-by keys specified after
-the `by` keyword and at the end of input, and produces one or more aggregations
+the `by` keyword, and at the end of input produces one or more aggregations
 for each unique set of group-by key values.
 
-Each aggregate function may be optionally followed by a `where` clause, which
-indicates a Boolean expression that indicates, for each input value,
-whether to deliver it to that aggregate.
+The keyword `summarize` can be treated as an [implied operator](../overview.md#26-implied-operators)
+and hence may be omitted before an aggregate function call.
+
+Each aggregate function may be optionally followed by a [`where`](where.md) clause, which
+applies a Boolean expression that indicates, for each input value,
+whether to deliver it to that aggregate function.
 
 The output field names for each aggregate and each key are optional.  If omitted,
-a field name is inferred from each right-hand side, e.g, the output field for the `sum`
-aggregate function is simply `sum`.
+a field name is inferred from each right-hand side, e.g, the output field for the [`count`](../aggregates/count.md)
+aggregate function is simply `count`.
 
-A key may be either an expression or a field.  If the key field is omitted it,
+A key may be either an expression or a field.  If the key field is omitted,
 it is inferred from the expression, e.g., the field name for `by lower(s)`
-is `lower.`
+is `lower`.
 
 If the cardinality of group-by keys causes the memory footprint to exceed
 a limit, then each aggregate's partial results are spilled to temporary storage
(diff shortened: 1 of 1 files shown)


docs/language/operators/summarize.md
```diff
@@ -5,26 +5,29 @@
 ### Synopsis
 
 ```
-summarize [<field>:=]<agg> [where <expr>][, [<field>:=]<agg> [where <expr>]] [by [<field>][:=<expr>] ...]
+summarize [<field>:=]<agg> [where <expr>][, [<field>:=]<agg> [where <expr>] ...] [by [<field>][:=<expr>] ...]
```
philrz, line 8: I figure the `...` should be added because the user is free to keep tacking on as many aggregate function calls as they want.

Also, I briefly thought about changing `summarize` to `[summarize]` to emphasize its optional nature due to the "implied operator" concept. However, I wasn't sure if we could say it's _always_ optional or if there's times it's necessary to remove ambiguity that I can't think of, so I've left that one as is. But please speak up if you do feel it should be `[summarize]`.

docs/language/operators/summarize.md
```diff
@@ -5,26 +5,29 @@
...
+the `by` keyword, and at the end of input produces one or more aggregations
 for each unique set of group-by key values.
 
-Each aggregate function may be optionally followed by a `where` clause, which
-indicates a Boolean expression that indicates, for each input value,
-whether to deliver it to that aggregate.
+The keyword `summarize` can be treated as an [implied operator](../overview.md#26-implied-operators)
+and hence may be omitted before an aggregate function call.
+
+Each aggregate function may be optionally followed by a [`where`](where.md) clause, which
```
philrz, line 20: Are folks ok with me linking to the `where` operator doc from here? I don't know if it's technically 100% the same thing since it's being used to affect the behavior of another operator rather than as a standalone operator. However, assuming they share the same code, I figure linking is the best way to answer the user's possible question of "What are the details on how this `where` clause is used?"

docs/language/operators/summarize.md
```diff
@@ -5,26 +5,29 @@
...
+applies a Boolean expression that indicates, for each input value,
+whether to deliver it to that aggregate function.
 
 The output field names for each aggregate and each key are optional.  If omitted,
-a field name is inferred from each right-hand side, e.g, the output field for the `sum`
-aggregate function is simply `sum`.
+a field name is inferred from each right-hand side, e.g, the output field for the [`count`](../aggregates/count.md)
+aggregate function is simply `count`.
```
philrz, line 26: I chose to reference `count` since it's one more opportunity to avoid any new user confusion about `summarize` vs. `sum`.

docs/language/operators/summarize.md
```diff
@@ -34,7 +37,16 @@ to a cluster of workers in an adaptive shuffle, though this is not yet implement
 
 ### Examples
 
-Sum the input sequence:
+Average the input sequence:
+```mdtest-command
+echo '1 2 3 4' | zq -z 'summarize avg(this)' -
```
philrz, line 42: Once again, I chose a different aggregate function so we'd not be _leading_ with `summarize sum` and perhaps confuse new users.

docs/language/operators/summarize.md
```diff
@@ -34,7 +37,16 @@ to a cluster of workers in an adaptive shuffle, though this is not yet implement
...
+```
+=>
+```mdtest-output
+{avg:2.5}
+```
+
+Sum the input sequence, leaving out the `summarize` keyword:
```
philrz, line 49: I figured that since we used the explicit `summarize` in one example and call attention here to intentionally leaving it out, now we're free to leave it out of all the other examples.

docs/language/operators/summarize.md
```diff
@@ -5,26 +5,29 @@
...
 to each input value optionally organized with the group-by keys specified after
-the `by` keyword and at the end of input, and produces one or more aggregations
+the `by` keyword, and at the end of input produces one or more aggregations
 for each unique set of group-by key values.
 
-Each aggregate function may be optionally followed by a `where` clause, which
-indicates a Boolean expression that indicates, for each input value,
-whether to deliver it to that aggregate.
+The keyword `summarize` can be treated as an [implied operator](../overview.md#26-implied-operators)
+and hence may be omitted before an aggregate function call.
```
nwt, line 18: Nit: Copy the corresponding language that appears in put.md.
```suggestion
The `summarize` keyword is optional since it is an
[implied operator](../overview.md#26-implied-operators).
```

docs/language/operators/summarize.md
nwt: I do think this is a bit confusing because this is a clause, not an operator. How about something like this?

```
Each aggregate function may be optionally followed by a `where` clause, which
applies a Boolean expression that indicates, for each input value,
whether to deliver it to that aggregate. (`where` clauses are analogous
to the [`where` operator](where.md).)
```

docs/language/operators/summarize.md
philrz: @nwt: Ah, thanks for the pointer to precedent! Pushed a commit going with `[summarize]`.

PR #4100, opened 2022-09-21
Delete Where v2

### LINKED ISSUE #3992: "zed delete -where" on more than just pool key
A user has expressed a requirement to do a `zed delete` with a richer `-where` query, which would be an enhancement over the current limitation of [just a single comparison against the pool key](https://zed.brimdata.io/docs/commands/zed/#24-delete).

@mccanne had the following to say on the possible implementation (I'm paraphrasing 😄):

> Super rich `-where` queries would run much slower since we'd need to read the whole pool. But if the query consists of a comparison against the pool key "`and` [something-other-than-the-pool-key]" it could run faster.

@mattnibs has also noted that the "could run faster" is dependent on merging what's in #3930.

mattnibs:
Closes #3992


service/handlers.go
```diff
@@ -477,7 +479,15 @@ func handleDelete(c *Core, w *ResponseWriter, r *Request) {
...
+		var program ast.Op
+		if program, err = c.compiler.Parse(payload.Where); err != nil {
+			w.Error(srverr.ErrInvalid(err))
+			return
+		}
+		commit, err = branch.DeleteWhere(r.Context(), c.compiler, program, message.Author, message.Body, message.Meta)
+	}
+	if errors.Is(err, &compiler.InvalidDeleteWhereQuery{}) {
+		err = srverr.ErrInvalid(err)
 	}
```
nwt, line 491:
```suggestion
		commit, err = branch.DeleteWhere(r.Context(), c.compiler, program, message.Author, message.Body, message.Meta)
		if errors.Is(err, &compiler.InvalidDeleteWhereQuery{}) {
			err = srverr.ErrInvalid(err)
		}
	}
```

compiler/lake.go
```diff
@@ -49,3 +54,70 @@ func (l *lakeCompiler) NewLakeQuery(pctx *op.Context, program ast.Op, parallelis
...
+			},
+		}},
+	})
+	entry, err := semantic.Analyze(pctx.Context, seq, src, head)
+	if err != nil {
+		return nil, err
+	}
+	if _, ok := entry.Ops[1].(*dag.Filter); !ok {
+		return nil, &InvalidDeleteWhereQuery{}
+	}
```
nwt, line 117: Could we get rid of compiler/kernel.DeleteFilter by doing something like this?
```suggestion
	f, ok := entry.Ops[1].(*dag.Filter)
	if !ok {
		return nil, &InvalidDeleteWhereQuery{}
	}
	// Invert the filter so we keep values that don't match the original.
	f.Expr = &dag.UnaryExpr{
		Kind:    "UnaryExpr",
		Op:      "!",
		Operand: f.Expr,
	}
```

lake/branch.go
```diff
@@ -124,155 +125,51 @@ func (b *Branch) Delete(ctx context.Context, ids []ksuid.KSUID, author, message
...
+		// anyways. Unfortunately there's quite a few layers of plumbing needed
+		// to get this working in compiler.
+		commitish := &lakeparse.Commitish{
+			Pool:   b.pool.Name,
+			Branch: parent.Commit.String(),
+		}
+		query, err := c.NewLakeDeleteQuery(pctx, program, commitish)
 		if err != nil {
 			return nil, err
 		}
```
nwt, line 147:
```suggestion
		query, err := c.NewLakeDeleteQuery(pctx, program, commitish)
		if err != nil {
			return nil, err
		}
		defer query.Pull(true)
```

runtime/exec/delete.go
```diff
@@ -0,0 +1,138 @@
...
+					err = err2
+				}
+				s.err = err
+				var count uint64
+				for _, b := range s.batches {
+					count += uint64(len(b.Values()))
+				}
+				if count == s.object.Count {
+					// Only return batches from records where objects have been
+					// deleted.
```
nwt, line 122:
```suggestion
					// Only return batches from objects where values have been
					// deleted.
```

nwt (approved):

compiler/lake.go
mattnibs: We talked about this but this can't be done since adding the not expression here would disable the object filter pass (which doesn't work with not expressions) and would there for require every object to be scanned on a delete that could otherwise be optimized.

PR #4101, opened 2022-09-21
return named types in zson.MarshalZNGContext.lookupType

### LINKED ISSUE #4083: another named type zson bug
Repro is at Zed commit 520b145.

I noticed this problem when looking at a zst section.  The named type is not being properly decorated.  If you output the section as ZNG then yield typeof(this), it is correct, but when outputing as ZSON and parsing the zson, you get the wrong typeof(this).

Make a t.zst file from this t.zson
```
{a:[1,2],b:[]([int64])}
{a:null([int64]),b:[1,2,3]}
```
using
```
zq -o t.zst -f zst t.zson
```
Now compare the outputs from these two commands:
```
zed dev dig section 1 t.zst | zq -Z 'typeof(this)' -
zed dev dig section -Z 1 t.zst | zq -Z 'typeof(this)' -
```
They should be the same obviously but they are not.

The problem is that the ZSON output from the section command does not decorate the different named types with the same name `Array`.  There is probably a simple name comparison somewhere in the zson formatter when the underlying type needs to be compared.

nwt:
A zson.MarshalZNGContext configured to decorate types doesn't always do so because the lookupType method doesn't return named types.  Fix that.

Closes #4083.

### DIFF
--- a/zson/marshal.go
+++ b/zson/marshal.go
@@ -264,37 +264,15 @@ func (m *MarshalZNGContext) encodeValue(v reflect.Value) (zed.Type, error) {
 	if err != nil {
 		return nil, err
 	}
-	if m.decorator != nil || m.bindings != nil {
-		// Don't create named types for interface types as this is just
-		// one value for that interface and it's the underlying concrete
-		// types that implement the interface that we want to name.
-		if !v.IsValid() || v.Kind() == reflect.Interface {
-			return typ, nil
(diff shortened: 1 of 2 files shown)


mccanne (approved):
Nice!  Cleaner too.   I experimented with this branch with the zst metadata section and it's all hanging together nicely now.

Issue #3200, opened 2021-10-20
fuzzy string matching


mccanne:
We should have functions for fuzzy string matching.

Postgres has a nice set of options to consider:

https://www.postgresql.org/docs/9.1/fuzzystrmatch.html

It looks like there are so open-source Go implementations of some of these algorithms...

philrz:
We attempted a [poll](https://brimdata.slack.com/archives/CTT24051C/p1663192286626839) of the community to see if there were opinions about a preferred fuzzy matching algorithm to start with, and with a whopping single vote for Levenshtein, we've decided to try that one first. 😉

philrz:
Verified in Zed commit 28f95a3.

```
$ echo '{"name": "Phillip"}' | zq -Z 'yield {name, distance_1: levenshtein(name, "Philip"), distance_2: levenshtein(name, "Phil")}' -
{
    name: "Phillip",
    distance_1: 1,
    distance_2: 3
}
```

Thanks @nwt!

PR #4104, opened 2022-09-22
add levenshtein() function

nwt:
For #3200.


mccanne (approved):
Love it how easily new functionality drops in!

PR #4109, opened 2022-09-27
Ignore trailing slash in lake uri

### LINKED ISSUE #4099: Ignore trailing slash on lake URL
Repro is with Zed commit b242ea7.

I'm not sure how exactly it got in there, but the cloud lake URL in my password manager somehow ended up with a trailing slash. Therefore a simple cut and paste led to this experience:

```
$ export ZED_LAKE="https://shasta.lake.brimdata.io/"

$ zed -version
Version: v1.2.0-60-gb242ea74

$ zed ls
status code 405: 
```

I scratched my head for a bit before I realized the trailing slash was the problem.

```
$ export ZED_LAKE="https://shasta.lake.brimdata.io"

$ zed ls
wrccdc 2BXU7qiuq6QX762ROF3DP9YnZI1 key ts order desc
github-metrics 27lUwGj6FyeYGO3ksWpfTNCUEJ3 key ts order desc
```

This problem can be reproduced with the local lake as well.

```
$ export ZED_LAKE="http://localhost:9867/"

$ zed ls
status code 405: 
```

Incidentally, while that HTTP 405 error was visible on the CLI, the `zed serve` log remained silent the whole time.

So, two possible Zed issues here:

1. We probably want the HTTP 405 error to show up in the log
2. Trailing slashes seem to have a way of showing up out in the wild, so we might want to have the client drop any trailing slash before trying to use the URL (or inform the user that it's invalid and make them correct it)

mattnibs:
Closes #4099

### DIFF
--- a/cli/lakeflags/flags.go
+++ b/cli/lakeflags/flags.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"strings"
 
 	"github.com/brimdata/zed/api/client"
 	"github.com/brimdata/zed/api/client/auth0"
@@ -101,7 +100,7 @@ func (l *Flags) URI() (*storage.URI, error) {
 	if l.Lake == "" {
 		return nil, errors.New("lake location must be set (either with the -lake flag or ZED_LAKE environment variable)")
 	}
-	u, err := storage.ParseURI(strings.TrimRight(l.Lake, "/"))
+	u, err := storage.ParseURI(l.Lake)
 	if err != nil {
 		err = fmt.Errorf("error parsing lake location: %w", err)
 	}
--- a/cli/lakeflags/flags.go
+++ b/cli/lakeflags/flags.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/brimdata/zed/api/client"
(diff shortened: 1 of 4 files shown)


cli/lakeflags/flags.go
```diff
@@ -100,7 +101,7 @@ func (l *Flags) URI() (*storage.URI, error) {
 	if l.Lake == "" {
 		return nil, errors.New("lake location must be set (either with the -lake flag or ZED_LAKE environment variable)")
 	}
-	u, err := storage.ParseURI(l.Lake)
+	u, err := storage.ParseURI(strings.TrimRight(l.Lake, "/"))
```
nwt, line 104: Instead of doing this here, I think we should do it in api/client.NewConnectionTo so anyone using that package benefits.

If you agree, let's also make the equivalent change in python/zed/zed.py.
```diff
     def __init__(self,
                  base_url=os.environ.get('ZED_LAKE', 'http://localhost:9867'),
                  config_dir=os.path.expanduser('~/.zed')):
-        self.base_url = base_url
+        self.base_url = base_url.rstrip('/')
         self.session = requests.Session()
         self.session.headers.update({'Accept': 'application/x-zjson'})
         token = self.__get_auth_token(config_dir)
```

mattnibs: Good idea 👍🏻

nwt (approved):

PR #4120, opened 2022-10-04
connection: Add SubscribeEvents

mattnibs:


api/client/events.go
```diff
@@ -0,0 +1,48 @@
...
+		if errors.Is(err, io.EOF) {
+			return "", nil, nil
+		}
+		return "", nil, err
+	}
+	var v interface{}
+	if err := l.unmarshaler.Unmarshal(data, &v); err != nil {
+		return "", nil, err
+	}
+	return kind, v, err
```
nwt, line 43: Nit: Returning kind doesn't provide any additional information. Maybe drop it?

api/client/events.go
```diff
@@ -0,0 +1,48 @@
...
+	"errors"
+	"fmt"
+	"io"
+
+	"github.com/brimdata/zed/api"
+	"github.com/brimdata/zed/zson"
+)
+
+type EventsClient struct {
+	resp        *Response
```
nwt, line 13: Nit: Don't really need the response here, just the body.
```suggestion
	rc        io.ReadCloser
```

nwt (approved):

api/client/events.go
mattnibs: You need kind because it distinguishes between pool-new and pool-deleted. I don't like that you need it but this is where we are.

PR #4135, opened 2022-10-07
wait for groupby and sort cleanup on exit

nwt:
The groupby and sort operators create goroutines that may need to remove temporary files, but no mechanism ensures they've finished that cleanup before the main goroutine exits.  Address that by adding a sync.WaitGroup to runtime/op/Context, registering the groupby and sort goroutines with it, and waiting for it in Context.Cancel (which appears to be called in all the right places).

For #1431.


mccanne (approved):
Nice!

PR #4137, opened 2022-10-10
implement complex nulls consistently in ZST

### LINKED ISSUE #3466: fix zst nulls
ZST doesn't properly support nulls anywhere as nulls are encoded as 
presence vectors for record fields.  This should be fixed probably by just encoding the null in the column and getting rid of presence vectors altogether and using null run-lengths in all of the value columns.

mccanne:
We never quite correctly implemented nulls for complex types in ZST. They were implemented in an ad hoc way using the "presence" concept inside of fields and inside of unions.

We noted that primitives encode null explicitly with their ZNG representation but complex types don't have a way to do that in ZST since they are representing the arrangement of primitives in a hierarchy so the nulls must be explicitly encoded for each complex type.

In this new approach, we simplify the concept with a generic Nulls metadata type that wraps each complex type.  The NullsWriter encodes nulls/values as run lengths and if there are not any nulls at EOF, it simply omits itself from the metadata output. The NullsReader wraps an underlying Reader and returns the sequence of nulls and underlying values by decoding the run lengths.

Closes #3466


zst/vector/nulls.go
```diff
@@ -0,0 +1,111 @@
...
+func (n *NullsReader) Read(b *zcode.Builder) error {
+	run := n.run
+	for run == 0 {
+		n.null = !n.null
+		v, err := n.runs.Read()
+		if err != nil {
+			return err
+		}
+		run = int(v)
+	}
```
mattnibs, line 104: I am probably missing something here but does this work if the first value is null? This seems to assume the first value is not null...

mccanne: Yes, I should document the encoding better.  I will add a comment and at some point we need to document it in the format spec.  I will create an issue to update the spec.

mattnibs (approved):

zst/vector/nulls.go
mccanne: BTW, there's probably a slightly better encoding for the runs but we can fix that later.  I will make an issue for that too.

PR #4145, opened 2022-10-11
Add regexp function

### LINKED ISSUE #2093: Regex group support
For some regex use cases, a user may need to go beyond just knowing whether there was a match and actually extract the strings that matched. An example implementation of this is [Python regex grouping](https://docs.python.org/3/howto/regex.html#grouping).

As a specific example use case, this could be a possible alternative to the implementation envisioned for #2092. Instead of using functions to split/slice desired fields, and re-join back into a string, a regex approach in Python could look like:

```
$ python
>>> import re
>>> query='www.brimsecurity.com'
>>> m = re.search(r'(\w+\.\w+)$', query)
>>> print(m.group(1))
brimsecurity.com
```

mattnibs:
Add a simple regexp function that returns the left most match of a string on a regular expression and the results of its sub expressions.

Closes #2093

### DIFF
--- a/docs/language/functions/regexp.md
+++ b/docs/language/functions/regexp.md
@@ -9,8 +9,9 @@ regexp(re: string, s: string) -> any
 ```
 ### Description
 The _regexp_ function returns an array of strings holding the text
-of the left most match of the regular expression string re and the
-results of its sub expression if there are any. A null value indicates
+of the left most match of the regular expression string `re` and the
+matches of each parenthesized subexpression (also known as capturing groups)
+if there are any. A null value indicates
 no match.
 
 ### Examples
--- a/docs/language/functions/regexp.md
+++ b/docs/language/functions/regexp.md
@@ -1,0 +1,34 @@
+### Function
+
+&emsp; **regexp** &mdash; perform a regular expression search on a string
+
+### Synopsis
+
+```
+regexp(re: string, s: string) -> any
+```
+### Description
+The _regexp_ function returns an array of strings holding the text
+of the left most match of the regular expression string re and the
+results of its sub expression if there are any. A null value indicates
+no match.
+
+### Examples
+
+Regexp returns an array of the match and its subexpressions:
+```mdtest-command
+echo '"seafood fool"' | zq -z 'yield regexp("foo(.?)", this)' -
+```
+=>
+```mdtest-output
+["food", "d"]
+```
+
+A null is returned if there is no match:
+```mdtest-command
+echo '"foo"' | zq -z 'yield regexp("bar", this)' -
+```
+=>
+```mdtest-output
+null([string])
+```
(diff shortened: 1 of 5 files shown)


docs/language/functions/regexp.md
```diff
@@ -0,0 +1,34 @@
...
+&emsp; **regexp** &mdash; perform a regular expression search on a string
+
+### Synopsis
+
+```
+regexp(re: string, s: string) -> any
+```
+### Description
+The _regexp_ function returns an array of strings holding the text
+of the left most match of the regular expression string re and the
```
philrz, line 12:
```suggestion
of the left most match of the regular expression string `re` and the
```

docs/language/functions/regexp.md
```diff
@@ -0,0 +1,34 @@
...
+results of its sub expression if there are any. A null value indicates
```
philrz, line 13:
```suggestion
matches of each parenthesized subexpression (also known as capturing groups)
if there are any. A null value indicates
```

docs/language/functions/regexp.md
```diff
@@ -0,0 +1,34 @@
...
+no match.
+
+### Examples
+
+Regexp returns an array of the match and its subexpressions:
+```mdtest-command
+echo '"seafood fool"' | zq -z 'yield regexp("foo(.?)", this) -'
```
philrz, line 20:
```suggestion
echo '"seafood fool friend"' | zq -z 'yield regexp("foo(.?) (\\w+) fr.*", this)' -
```

I think it'd be helpful for the example to show multiple capture groups to help users recognize that the approach is very similar to what they may have seen in Python, TCL, etc.

docs/language/functions/regexp.md
```diff
@@ -0,0 +1,34 @@
...
+```
+=>
+```mdtest-output
+["food", "d"]
```
philrz, line 24:
```suggestion
["food fool friend","d","fool"]
```

runtime/expr/function/function.go
```diff
@@ -150,6 +150,9 @@ func New(zctx *zed.Context, name string, narg int) (expr.Function, field.Path, e
 		f = &ParseURI{zctx: zctx, marshaler: zson.NewZNGMarshalerWithContext(zctx)}
 	case "parse_zson":
 		f = &ParseZSON{zctx: zctx}
+	case "regexp":
+		argmin, argmax = 2, 2
+		f = &Regexp{zctx: zctx}
 	case "quiet":
 		f = &Quiet{zctx: zctx}
```
nwt, line 157:
```suggestion
	case "quiet":
		f = &Quiet{zctx: zctx}
	case "regexp":
		argmin, argmax = 2, 2
		f = &Regexp{zctx: zctx}
```

nwt (approved):
Nicely done!

PR #4150, opened 2022-10-13
Move logger related flags to cli package

mattnibs:


mattnibs:
We need logging for the forthcoming manage monitor command, so I thought now is a time to centralize log related flags for easier use.

service/core.go
```diff
@@ -65,9 +64,9 @@ type Core struct {
 	subscriptionsMu sync.RWMutex
 }
 
-func NewCore(ctx context.Context, conf Config) (*Core, error) {
-	if conf.Logger == nil {
-		conf.Logger = zap.NewNop()
+func NewCore(ctx context.Context, conf Config, logger *zap.Logger) (*Core, error) {
```
nwt, line 67: I'd leave logger as a field in Config but don't have strong feelings about that.

nwt (approved):

service/core.go
mattnibs: ok

PR #4151, opened 2022-10-13
Parse non-search regular expressions as strings

### LINKED ISSUE #4148: Make regexps compile to a string outside the search context
As of the filing of this issue, Zed is currently at commit e9624c3.

With the arrival of the `regexp()` function via #2093, it has been observed that in the current implementation, the backslashes that often appear in regexps need to be escaped to be passed into the function as part of a string representation of the regexp. E.g.:

```
$ zq -version
Version: v1.2.0-85-ge9624c3b

$ echo '{"query": "www.brimdata.io"}' | zq -z 'yield regexp("(\\w+\\.\\w+)$", query)' -
["brimdata.io","brimdata.io"]
```

It has been proposed that we improve on this by having regular expressions compile to a string outside the search context. For instance in when writing a regex for the new regexp function it would be able to do:

```
regexp(/(\w+\.\w+)/", query)
```

mattnibs:
Parse regular expressions outside of the search context as string literals.

Close #4148

### DIFF
--- a/compiler/parser/parser.es.js
+++ b/compiler/parser/parser.es.js
@@ -873,59 +873,62 @@ function peg$parse(input, options) {
       peg$c381 = "null",
       peg$c382 = peg$literalExpectation("null", false),
       peg$c383 = function() { return {"kind": "Primitive", "type": "null", "text": ""} },
-      peg$c384 = "0x",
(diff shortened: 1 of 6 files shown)


philrz (approved):
I'll let the codes give the official approval, but I just tested it out in the branch and it's a functional 👍 from me.

PR #4157, opened 2022-10-19
index apply: ability to apply multiple rules

mattnibs:
Update the index apply command line and service endpoint so that multiple rules can be applied to an indexing operation.

The pr also adjusts the index rule lookup logic so that either the rule name or rule ID can be used in reference to an index rule.


nwt (commented):
Add tests for multiple -r flags and for lookup-by-ref so we know those work and stay working down the road?

nwt (approved):
Nice. :shipit:

PR #4158, opened 2022-10-19
regexp func: Allow regexp literal as first argument

### LINKED ISSUE #4148: Make regexps compile to a string outside the search context

mattnibs:
Allow the use of a regexp literal as the first argument for the regexp function.

Close #4148

### DIFF
--- a/compiler/parser/parser.peg
+++ b/compiler/parser/parser.peg
@@ -721,7 +721,11 @@ Cast
 
 Function
   = Grep
-  / RegexpFunc
+  // Special case to handle `regexp(/x/, y)`.
+  / "regexp" __ "(" __ arg0Text:RegexpPattern __ "," __ arg1:Expr __ ")" where:WhereClause? {
+      VAR(arg0) = MAP("kind": "Primitive", "type": "string", "text": arg0Text)
+      RETURN(MAP("kind": "Call", "name": "regexp", "args": ARRAY(arg0, arg1), "where": where))
+    }
   / !FuncGuard fn:IdentifierName __ "(" __ args:FunctionArgs __ ")" where:WhereClause? {
       RETURN(MAP("kind": "Call", "name": fn, "args": args, "where": where))
     }
--- a/compiler/parser/parser.peg
+++ b/compiler/parser/parser.peg
@@ -750,22 +750,6 @@ Pattern
       RETURN(MAP("kind": "String", "text": s))
     }
 
-RegexpFunc
-  = "regexp" __ "(" __ args:RegexpFuncArgs __ ")" where:WhereClause? {
-      RETURN(MAP("kind": "Call", "name": "regexp", "args": args, "where": where))
-    }
-
-RegexpFuncArgs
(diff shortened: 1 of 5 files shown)


compiler/parser/parser.peg
```diff
@@ -721,6 +721,7 @@ Cast
 
 Function
   = Grep
+  / RegexpFunc
```
nwt, line 724:
```suggestion
  // Special case to handle `regexp(/x/, y)`.
  / "regexp" __ "(" __ arg0Text:RegexpPattern __ "," __ arg1:Expr __ ")" where:WhereClause? {
      VAR(arg0) = MAP("kind": "Primitive", "type": "string", "text": arg0Text)
      RETURN(MAP("kind": "Call", "name": "regexp", "args": ARRAY(arg0, arg1), "where": where))
    }
```

docs/language/functions/regexp.md
```diff
@@ -18,7 +18,7 @@ no match.
 
 Regexp returns an array of the match and its subexpressions:
 ```mdtest-command
-echo '"seafood fool friend"' | zq -z 'yield regexp("foo(.?) (\\w+) fr.*", this)' -
+echo '"seafood fool friend"' | zq -z 'yield regexp(/foo(.?) (\w+) fr.*/, this)' -
```
nwt, line 21: We should update the synopsis and description, too.

compiler/parser/parser.peg
mattnibs: Oh yeah good call 👍🏻

docs/language/functions/regexp.md
```diff
@@ -5,20 +5,20 @@
...
+regexp(re: string|regexp, s: string) -> any
 ```
 ### Description
 The _regexp_ function returns an array of strings holding the text
-of the left most match of the regular expression string `re` and the
-matches of each parenthesized subexpression (also known as capturing groups)
-if there are any. A null value indicates
-no match.
+of the left most match of the regular expression `re`, which can be either
+a string value or [regular expression](https://zed.brimdata.io/docs/next/language/overview#711-regular-expressions),
```
nwt, line 13: The link in glob.md looks like this.
```suggestion
a string value or a [regular expression](../overview.md#711-regular-expressions),
```

nwt (approved):

docs/language/functions/regexp.md
mattnibs: Oh, shoot that's right.

PR #4173, opened 2022-11-02
Add map aggregation function

mattnibs:


docs/language/aggregates/map.md
```diff
@@ -0,0 +1,36 @@
+### Aggregate Function
+
+&emsp; **map** &mdash; aggregates map values into a single map
```
nwt, line 3: For consistency with docs/language/aggregates/collect.md.
```suggestion
&emsp; **map** &mdash; aggregate map values into a single map
```

docs/language/aggregates/map.md
```diff
@@ -0,0 +1,36 @@
...
+
+### Synopsis
+```
+map(|{ any: any, ...}|) -> |{ any: any, ...}|
```
nwt, line 7:
```suggestion
map(|{any:any}|) -> |{any:any}|
```

docs/language/aggregates/map.md
```diff
@@ -0,0 +1,36 @@
...
+```
+### Description
+
+The _map_ aggregate function combines map typed inputs into a single map output.
```
nwt, line 11:
```suggestion
The _map_ aggregate function combines map inputs into a single map output.
```

docs/language/aggregates/map.md
```diff
@@ -0,0 +1,36 @@
...
+If _map_ receives multiple values for the same key the last value receieved is
```
nwt, line 12:
```suggestion
If _map_ receives multiple values for the same key, the last value received is
```

docs/language/aggregates/map.md
```diff
@@ -0,0 +1,36 @@
...
+retained. If the input key/values vary in type, the return type will be a map
+of union of the types encountered.
```
nwt, line 14:
```suggestion
retained. If the input keys or values vary in type, the return type will be a map
of union of those types.
```

runtime/expr/agg/map.go
```diff
@@ -0,0 +1,98 @@
...
+		return
+	}
+	mtyp, ok := zed.TypeUnder(val.Type).(*zed.TypeMap)
+	if !ok {
+		return
+	}
+	it := val.Bytes.Iter()
+	for !it.Done() {
+		ktyp, kb := untagUnionValue(mtyp.KeyType, it.Next())
+		vtyp, vb := untagUnionValue(mtyp.ValType, it.Next())
```
nwt, line 37: Use expr.ValueUnder instead of untagUnionValue.

philrz:
`docs/language/aggregates/README.md` needs a link to the new `map.md`.

philrz (commented):
I tested out the branch and was able to rewrite my recent Hack Day project to take advantage of it and it worked nicely, so, it's a functional 👍 from me.

runtime/expr/agg/map.go
mattnibs: This will strip named types, is that something we want to do here?

mattnibs: You want to removed the named type if its a named type of a union (I should fix that in `untagUnionValue`).

nwt: > This will strip named types, is that something we want to do here?

Good point!

I think that means we want something like this:
```go
// valueUnder is like expr.ValueUnder but it preserves non-union named types.
func valueUnder(typ zed.Type, b zcode.Bytes) *zed.Value {
	val := zed.NewValue(typ, b)
	if _, ok := zed.TypeUnder(typ).(*zed.TypeUnion); !ok {
		return val
	}
	return expr.ValueUnder(val)
}
```

runtime/expr/agg/map.go
```diff
@@ -0,0 +1,98 @@
...
+
+func (m *Map) Consume(val *zed.Value) {
+	if val.IsNull() {
+		return
+	}
+	mtyp, ok := zed.TypeUnder(val.Type).(*zed.TypeMap)
+	if !ok {
+		return
+	}
+	it := val.Bytes.Iter()
```
nwt, line 34:
```suggestion
	// Copy val.Bytes since we're going to keep slices of it.
	it := zcode.Iter(slices.Clone(val.Bytes))
```

runtime/expr/agg/map.go
```diff
@@ -0,0 +1,98 @@
...
+	for !it.Done() {
+		ktyp, kb := untagUnionValue(mtyp.KeyType, it.Next())
+		vtyp, vb := untagUnionValue(mtyp.ValType, it.Next())
+		m.scratch = zed.AppendTypeValue(m.scratch[:0], ktyp)
+		m.scratch = append(m.scratch, kb...)
```
nwt, line 39: Using kb here means this can't distinguish between the null key and the zero-length key. Here's one way to do that:

```go
	keyTagAndBody := it.NextTagAndBody()
	ktyp, kb := untagUnionValue(mtyp.KeyType, keyTagAndBody.Body())
	m.scratch = zed.AppendTypeValue(m.scratch[:0], ktyp)
	m.scratch = append(m.scratch, keyTagAndBody...)
```

```

File: runtime/expr/agg/map.go (hunk 6 above)
Comment (mattnibs) [reply runtime/expr/agg/map.go:37]: Ugh, I can't use expr.ValueUnder since it creates an import cycle.

Comment (nwt) [reply runtime/expr/agg/map.go:37]: Maybe make it a method on zed.Value?

Comment (mattnibs) [reply runtime/expr/agg/map.go:37]: Interesting

File: runtime/expr/agg/map.go [hunk 9]
```diff
@@ -0,0 +1,100 @@
...
+
+func unionOf(zctx *zed.Context, types []zed.Type) zed.Type {
+	types = zed.UniqueTypes(types)
+	if len(types) == 1 {
+		return types[0]
+	}
+	return zctx.LookupTypeUnion(types)
+}
+
+// valueUnder is like expr.ValueUnder but it preserves non-union named types.
```
nwt, line 93:
```suggestion
// valueUnder is like zed.(*Value).Under but it preserves non-union named types.
```

runtime/expr/agg/map.go
```diff
@@ -0,0 +1,100 @@
...
+func valueUnder(typ zed.Type, b zcode.Bytes) *zed.Value {
+	val := zed.NewValue(typ, b)
+	if _, ok := zed.TypeUnder(typ).(*zed.TypeUnion); ok {
```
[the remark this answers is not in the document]
mattnibs: whoops, thanks!

runtime/expr/agg/map.go
```diff
@@ -0,0 +1,103 @@
...
+	return zed.NewValue(typ, b)
+}
+
+func (m *Map) ResultAsPartial(zctx *zed.Context) *zed.Value {
+	return m.Result(zctx)
+}
+
+func appendMapVal(b *zcode.Builder, typ zed.Type, val *zed.Value, uniq int) {
+	if u, ok := typ.(*zed.TypeUnion); ok && uniq > 1 {
+		zed.BuildUnion(b, u.TagOf(val.Type), val.Bytes)
```
nwt, line 82: Nit: If uniq>1 but typ isn't a union, something's gone very sideways and I think we should let the type assertion panic so we find out.
```suggestion
	if uniq > 1 {
		u := typ.(*zed.TypeUnion)
		zed.BuildUnion(b, .TagOf(val.Type), val.Bytes)
```

mattnibs: Good call 👍🏻

PR #4174, opened 2022-11-02
runtime/op/groupby: clean up on Proc.agg.Consume error

### LINKED ISSUE #4171: Spill directory not removed after Ctrl-C'ing zq aggregation
Repro is with Zed commit d8f34f5 using test data `s3://brim-sampledata/fdns/fdns-a.zng.gz`.

To watch for the creation of a spill directory, in one window I run:

```
# while true; date; do ls -l ${TMPDIR}zed-spill-*; sleep 1; done
```

Then I run an aggregation that will spill-to-disk.

```
# zq 'count() by value' fdns-a.zng.gz
```

It doesn't reproduce every time, but it seems that about half the time if I hit Ctrl-C to stop `zq` some time after the spill directory starts being populated, the `zq` process exits and leaves the spill directory behind.

See the video below for an example repro using these steps.

https://user-images.githubusercontent.com/5934157/199375242-a6ca3862-62d1-4258-ac3a-1b5970336e3e.mp4

nwt:
runtime/op/groupby.Proc.run will exit without calling Proc.agg.spiller.Cleanup if Proc.agg.Consume returns an error and Proc.sendResult succeeds in sending that error to Proc.resultCh.  Defer a call to spiller.Cleanup at the top of Proc.run so cleanup always happens on exit.

This change makes the call to Proc.reset after <-p.ctx.Done() in Proc.sendResult unnecessary, so remove it.

Fixes #4171.

### DIFF
--- a/runtime/op/groupby/groupby.go
+++ b/runtime/op/groupby/groupby.go
@@ -173,8 +173,13 @@ func (p *Proc) Pull(done bool) (zbuf.Batch, error) {
 }
 
 func (p *Proc) run() {
-	// Tell p.ctx's cancel function that we've finished our cleanup.
-	defer p.pctx.WaitGroup.Done()
+	defer func() {
+		if p.agg.spiller != nil {
+			p.agg.spiller.Cleanup()
+		}
+		// Tell p.ctx's cancel function that we've finished our cleanup.
+		p.pctx.WaitGroup.Done()
+	}()
 	sendResults := func(p *Proc) bool {
 		for {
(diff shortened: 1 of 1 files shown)


philrz (commented):
FYI, I tested out the #4171 repro steps on this branch and after about 10 attempts indeed I no longer see spill directories left behind after Ctrl-C'ing, so this is a functional 👍 from me.

mattnibs (approved):

PR #4193, opened 2022-11-08
Add a link from README to map aggregate function

### LINKED ISSUE #4141: Create a map by gathering key/value pairs
A community zync user inquired:

> Does Zed provide a way to `collect()` values into a map instead of an array?

It currently does not, but the Dev team recognizes this would be cool & useful. Two approaches come to mind:

1. An operator that could turn a table of aggregation results into a map, e.g., using this Zeek example:

   ```
   count() by _path | map _path,count | ...
   ```

2. An aggregation function, e.g.:

   ```
   summarize map(foo,bar) by other_val
   ```

   Where the `foo` & `bar` become the key/value pairs, respectively, of the created map.

We expect both approaches are probably something we'd want to support.

philrz:
A loose end related to #4173.

Closes #4141.

### DIFF
--- a/docs/language/aggregates/README.md
+++ b/docs/language/aggregates/README.md
@@ -12,6 +12,7 @@ or expression context and produce an aggregate value for a sequence of inputs va
 - [count](count.md) - count input values
 - [dcount](dcount.md) - count distinct input values
 - [fuse](fuse.md) - compute a fused type of input values
+- [map](map.md) - aggregate map values into a single map
 - [max](max.md) - maximum value of input values
 - [min](min.md) - minimum value of input values
 - [or](or.md) - logical OR of input values


docs/language/aggregates/README.md
```diff
@@ -12,6 +12,7 @@ or expression context and produce an aggregate value for a sequence of inputs va
 - [count](count.md) - count input values
 - [dcount](dcount.md) - count distinct input values
 - [fuse](fuse.md) - compute a fused type of input values
+- [map](map.md) - aggregate map values into a single map
 - [max](max.md) - maximum value of input values
```
nwt, line 16: Keep these alphabetized.
```suggestion
- [max](max.md) - maximum value of input values
- [map](map.md) - aggregate map values into a single map
```

PR #4195, opened 2022-11-09
UnmarshalZNG: Explicitly check for type null null

mattnibs:
The simple equality check for type null null values in UnmarshalZNG was resulting erroneous unmarshal errors. Fix this.

### DIFF
--- a/zson/marshal_zng_test.go
+++ b/zson/marshal_zng_test.go
@@ -254,11 +254,18 @@ func TestUnmarshalNull(t *testing.T) {
 		var obj struct {
 			Test *testobj `zed:"test"`
 		}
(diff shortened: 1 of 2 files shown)


zson/marshal_zng_test.go
```diff
@@ -251,14 +251,21 @@ func TestUnmarshalNull(t *testing.T) {
 		type testobj struct {
 			Val int
 		}
-		var obj struct {
+		var ptr struct {
```
nwt, line 254: Nit: I think obj is a better name for this than ptr since this isn't a pointer.

nwt (approved):

PR #4196, opened 2022-11-09
lake manage: Add indexing support

mattnibs:
Add indexing support to lake manage commands.

This support also renames the compact command to update. The update is functionally similar to the monitor command only it updates the lake once and exits whereas monitor continuously performs updates.


cmd/zed/manage/update/command.go
```diff
@@ -0,0 +1,55 @@
...
+	New:   New,
+}
+
+func init() {
+	manage.Cmd.Add(Cmd)
+}
+
+type Command struct {
+	*manage.Command
+	commitFlags commitflags.Flags
```
nwt, line 26: Is this used?

cmd/zed/manage/update/command.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/brimdata/zed/cli/logflags"
+	"github.com/brimdata/zed/cmd/zed/manage"
+	"github.com/brimdata/zed/cmd/zed/manage/lakemanage"
+	"github.com/brimdata/zed/pkg/charm"
+)
+
+var Cmd = &charm.Spec{
+	Name:  "update",
+	Usage: "update",
+	Short: "performs compaction and index on all pools and exit",
```
nwt, line 16:
```suggestion
	Short: "compact and index pools and then exit",
```

lake/api/api.go
```diff
@@ -60,6 +60,32 @@ func ScanIndexRules(ctx context.Context, api Interface) (zio.ReadCloser, error)
...
+		val, err := r.Read()
+		if val == nil || err != nil {
+			return rules, err
+		}
+		var r index.Rule
+		if err := u.Unmarshal(val, &r); err != nil {
+			return nil, err
+		}
+		rules = append(rules, r)
+	}
```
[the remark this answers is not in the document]
mattnibs: Nice!

cmd/zed/manage/update/command.go
mattnibs: It is not!

cmd/zed/manage/lakemanage/branch.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+		found++
+		compacted += len(run.Objects)
+		b.logger.Debug("compacted", zap.Stringer("commit", commit), zap.Int("objects_compacted", len(run.Objects)))
+	}
+	if compacted == 0 {
+		b.logger.Debug("compaction completed", zap.Int("runs_found", found), zap.Int("objects_compacted", compacted))
+	} else {
+		b.logger.Info("compaction completed", zap.Int("runs_found", found), zap.Int("objects_compacted", compacted))
+	}
```
nwt, line 73: But to be honest, I think we should log everything at the info level.
```suggestion
	level := zap.InfoLevel
	if compacted == 0 {
		level = zap.DebugLevel
	}
	b.logger.Log(level, "compaction completed", zap.Int("runs_found", found), zap.Int("objects_compacted", compacted))
```

cmd/zed/manage/lakemanage/compact.go
```diff
@@ -72,3 +72,49 @@ func Scan(ctx context.Context, it ObjectIterator, pool *pools.Config,
...
+		return nil, err
+	}
+	return &o, nil
+}
+
+func (r *PoolDataIterator) Close() error {
+	return r.reader.Close()
+}
+
+type DataIterator interface {
```
nwt, line 118: DataObjectIterator, maybe? Or maybe not!

cmd/zed/manage/lakemanage/config.go
```diff
@@ -1,7 +1,56 @@
...
+	if !c.Enabled() {
+		return nil
+	}
+	rules, err := api.GetIndexRules(ctx, lk)
+	if err != nil {
+		return err
+	}
+	ss := lakeparse.FormatIDs(c.RuleNames)
+loop:
+	for _, s := range ss {
```
nwt, line 41:
```suggestion
loop:
	for _, id := range lakeparse.FormatIDs(c.RuleNames) {
```

cmd/zed/manage/lakemanage/lakemanage.go
```diff
@@ -0,0 +1,211 @@
...
+	for {
+		switch err := runMonitor(ctx, conf, conn, logger); {
+		case errors.Is(err, syscall.ECONNREFUSED):
+			logger.Info("cannot connect to lake, retrying in 5 seconds")
+		case err != nil:
+			return err
+		}
+		timer.Reset(time.Second * 5)
+		select {
+		case <-timer.C:
```
nwt, line 65: I don't think a time.Timer is needed here since efficiency isn't a concern.
```suggestion
		case <-time.After(5 * time.Second):
```

cmd/zed/manage/lakemanage/branch.go
mattnibs: I can dig it 👍🏻

cmd/zed/manage/lakemanage/lakemanage.go
mattnibs: Good point

cmd/zed/manage/lakemanage/branch.go
```diff
@@ -0,0 +1,104 @@
...
+		}
+		found++
+		compacted += len(run.Objects)
+		b.logger.Debug("compacted", zap.Stringer("commit", commit), zap.Int("objects_compacted", len(run.Objects)))
+	}
+	level := zap.InfoLevel
+	if compacted == 0 {
+		level = zap.DebugLevel
+	}
+	b.logger.Log(level, "compaction completed", zap.Int("runs_found", found), zap.Int("objects_compacted", compacted))
```
nwt, line 73: Doh! This method appeared in v1.22.0 but we're at v1.16.0.

cmd/zed/manage/lakemanage/lakemanage.go
```diff
@@ -0,0 +1,211 @@
...
+	if logger == nil {
+		logger = zap.NewNop()
+	}
+	timer := time.NewTimer(0)
+	<-timer.C
+	defer func() {
+		if !timer.Stop() {
+			<-timer.C
+		}
+	}()
```
nwt, line 55:
```suggestion
```

cmd/zed/manage/lakemanage/lakemanage.go
```diff
@@ -0,0 +1,211 @@
...
+	for {
+		switch err := runMonitor(ctx, conf, conn, logger); {
+		case errors.Is(err, syscall.ECONNREFUSED):
+			logger.Info("cannot connect to lake, retrying in 5 seconds")
+		case err != nil:
+			return err
+		}
+		timer.Reset(time.Second * 5)
```
nwt, line 63:
```suggestion
```

cmd/zed/manage/lakemanage/lakemanage.go
```diff
@@ -0,0 +1,211 @@
...
+		return
+	}
+	go func() {
+		timer := time.NewTimer(0)
+		<-timer.C
+		defer func() {
+			if !timer.Stop() {
+				<-timer.C
+			}
+		}()
```
nwt, line 198: Don't think we need this.
```suggestion
```

cmd/zed/manage/lakemanage/branch.go
mattnibs: I'll upgrade

cmd/zed/manage/lakemanage/lakemanage.go
mattnibs: Whoops.

cmd/zed/manage/lakemanage/lakemanage.go
mattnibs: Good point, I don't think so either.

cmd/zed/manage/lakemanage/lakemanage.go
```diff
@@ -0,0 +1,203 @@
...
+		if err != nil {
+			if errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) {
+				// Ignore EOF error from lost connection.
+				return nil
+			}
+			return err
+		}
+		switch kind {
+		case "pool-new":
+			ev := detail.(*api.EventPool)
```
nwt, line 114: Nits: Why switch on kind instead of the type? And it's surprising to shadow ev with something of a different type.
```suggestion
		switch detail := detail.(type) {
		case *api.EventPool:
```

cmd/zed/manage/lakemanage/branch.go
```diff
@@ -0,0 +1,104 @@
...
+	return b
+}
+
+func (b *branch) compact(ctx context.Context) (*time.Time, error) {
+	b.logger.Debug("compaction started")
+	head := lakeparse.Commitish{Pool: b.pool.Name, Branch: b.name}
+	it, err := NewPoolDataObjectIterator(ctx, b.lake, &head, b.pool.Layout)
+	if err != nil {
+		return nil, err
+	}
```
nwt, line 51:
```suggestion
	it, err := NewPoolDataObjectIterator(ctx, b.lake, &head, b.pool.Layout)
	if err != nil {
		return nil, err
	}
	defer it.Close()
```

cmd/zed/manage/lakemanage/index.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/brimdata/zed/zson"
+	"github.com/segmentio/ksuid"
+)
+
+func IndexScan(ctx context.Context, lk api.Interface, pool, branch string, thresh time.Duration,
+	rules []index.Rule, ch chan<- ObjectIndexes) (*time.Time, error) {
+	it, err := newIndexIterator(ctx, lk, pool, branch)
+	if err != nil {
+		return nil, err
+	}
```
nwt, line 21:
```suggestion
	it, err := newIndexIterator(ctx, lk, pool, branch)
	if err != nil {
		return nil, err
	}
	defer it.reader.Close()
```

cmd/zed/manage/lakemanage/lakemanage.go
mattnibs: "pool-new" and "pool-delete" both use `*api.EventPool` but mean different things.

nwt: > I'm not quite sure I understand this. The design as is on this endpoint isn't ideal- i'd like to at some point make these zson, but kept this to avoid a migration for now.

ev is a *client.EventsClient outside this switch statement but a *api.EventWhatever inside it. Maybe do this?

```suggestion
		switch kind {
		case "pool-new":
			detail := detail.(*api.EventPool)
```

nwt (approved):

cmd/zed/manage/lakemanage/lakemanage.go
mattnibs: I can dig it

PR #4200, opened 2022-11-10
Clean up sidebar indexing

philrz:
I sat down to make an addition to the docs, but then noticed that there were some inconsistencies with the sidebar indexing that could stand to be cleaned up first. I've eyeballed the rendering after these changes to confirm that the content/order in the sidebar is exactly the same as it was. I'll add some inline comments to point to the reason behind certain less-obvious changes.

### DIFF
--- a/docs/commands/README.md
+++ b/docs/commands/README.md
@@ -1,8 +1,3 @@
----
-sidebar_position: 3
-sidebar_label: Commands
----
-
 # Command Tooling
 
 The Zed system is managed and queried with the [`zed` command](zed.md),
--- a/docs/commands/zq.md
+++ b/docs/commands/zq.md
@@ -1,5 +1,5 @@
 ---
-sidebar_position: 0
+sidebar_position: 1
 sidebar_label: zq
 description: A command-line tool that uses the Zed Language for pipeline-style search and analytics.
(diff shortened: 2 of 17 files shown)


docs/commands/README.md
```diff
@@ -1,8 +1,3 @@
----
-sidebar_position: 3
-sidebar_label: Commands
----
-
```
philrz, line 5: For these README files that act as intro text for a folder ("category") of docs (such as **Commands**), it's the `_category_.yaml` that has the active config that determines the position/label in the sidebar, so this stuff I've removed here was effectively a no-op.

docs/commands/zq.md
```diff
@@ -1,5 +1,5 @@
 ---
-sidebar_position: 0
+sidebar_position: 1
 sidebar_label: zq
 description: A command-line tool that uses the Zed Language for pipeline-style search and analytics.
```
philrz, line 4: It's unclear to me from the Docusaurus docs how this `description` field is actually used. I don't think it's populating anything reader-facing for us at the moment so I was tempted to remove it, but I decided it's not hurting anything so I left it alone.

nwt (approved):

PR #4202, opened 2022-11-10
Add a docs page about AWS S3

philrz:
I recently fielded a question from a community user about S3 support in Zed. This led me to realize that we don't have any kind of formal docs at the moment that speak to the current S3 support, so I've added that here. Since I couldn't find a great home in one of the current docs categories, I created a new section called "Integrations" of which this is the first doc.

A follow-on I plan after this PR is to link to it from some other docs that mention S3, including the `from` doc which I also recognized needs a significant update.

### DIFF
--- a/docs/integrations/amazon-s3.md
+++ b/docs/integrations/amazon-s3.md
@@ -35,6 +35,11 @@ are typically input in the first three settings entered during `aws configure`.
    Default region name [*********]: 
    ```
 
+## Endpoint
+
+To use S3-compatible storage not provided by AWS, set the `AWS_S3_ENDPOINT`
+environment variable to the hostname or URI of the provider.
+
 ## Wildcard Support
 
 [Like the AWS CLI tools themselves](https://aws.amazon.com/premiumsupport/knowledge-center/s3-event-notification-filter-wildcard/#:~:text=Because%20the%20wildcard%20asterisk%20character,suffix%20object%20key%20name%20filter.),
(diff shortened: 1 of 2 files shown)


docs/integrations/amazon-s3.md
```diff
@@ -0,0 +1,44 @@
...
+can be set in environment variables `AWS_ACCESS_KEY_ID`,
+`AWS_SECRET_ACCESS_KEY`, and `AWS_REGION`. The values for these are the same as
+are typically input in the first three settings entered during `aws configure`.
+
+   ```
+   $ aws configure
+   AWS Access Key ID [********************]: 
+   AWS Secret Access Key [********************]: 
+   Default region name [*********]: 
+   ```
```
nwt, line 36:
```suggestion
## Region

You must specify an AWS region via one of the following:
* The `AWS_REGION` environment variable
* The `~/.aws/config` file
* The file specified by the `AWS_CONFIG_FILE` environment variable

You can create `~/.aws/config` by installing the
[AWS CLI](https://aws.amazon.com/cli/) and running `aws configure`.

## Credentials

You must specify AWS credentials via one of the following:
* The `AWS_ACCESS_KEY_ID` and`AWS_SECRET_ACCESS_KEY` environment variables
* The `~/.aws/credentials` file
* The file specified by the `AWS_SHARED_CREDENTIALS_FILE` environment variable

You can create `~/.aws/credentials` by installing the
[AWS CLI](https://aws.amazon.com/cli/) and running `aws configure`.
```

philrz:
Thanks @nwt. I've incorporated your suggestions, which led me to think of one more change about having to set `AWS_REGION` even using non-AWS S3 that doesn't require it, such as we do in https://github.com/brimdata/zed/blob/main/testdata/minio.sh. As a user, once I was using Minio it felt odd that the region setting was required, but I assume this is a side effect of the SDK coming from Amazon where it's always required. I figure pointing this out in our docs might avoid some users raising it as if it were a bug, so I've pushed one more commit.

nwt (approved):

PR #4206, opened 2022-11-14
Fix meta queries on empty pools

### LINKED ISSUE #3029: Improve error message for empty "zed lake log"
Repro is with Zed commit 8824189.

```
$ zed -version
Version: v0.29.0-503-g88241892

$ zed lake init
using environment variable ZED_LAKE_ROOT
lake created: /Users/phil/logs

$ zed lake create -orderby ts:desc logs
pool created: logs 1xoqIDTLBuvs5gHbhRrvWPIBj2X

$ zed lake checkout -p logs
Switched to branch "main" on pool "logs"

$ zed lake log
unknown pool metadata type: "log"
```

mattnibs:
Fix bug with running meta commit queries on empty pools. Previously the error message would state that the meta type for the commit did not exist when it did. Change this so the query remains a meta commit query and a query on an empty snapshot is run.

This pr also fixes a panic that occurred when running a partitions meta query on an empty pool.

Closes #3029

### DIFF
--- a/compiler/semantic/op.go
+++ b/compiler/semantic/op.go
@@ -218,7 +218,7 @@ func semPoolWithName(ctx context.Context, scope *Scope, p *ast.Pool, poolName st
 		}
 	}
 	if p.Spec.Meta != "" {
-		if commitID != ksuid.Nil {
+		if commit != "" {
 			return &dag.CommitMeta{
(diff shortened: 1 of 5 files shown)


nwt (approved):
🏆

Issue #4212, opened 2022-11-16
further refactor meta-data in dag


mccanne:
After merging the first refactor of metadata handling for scanner logic,
this issue defines some further simplifictions:
* get rid of the channel-based scans and simplify to use the snapshot directly,
* rework the indexer to operate directly on object streams as a standalone op,
* change partitioner to be meta op slicer and take sorted objects in and emit partitions,
* rework SequenceScanner to take either objects or partitions so it can be hooked up either way,
* simplify zbuf.Filter by removing the methods having to do with object filtering and make those parameters to the meta ops,
* add dag nodes to mirror the meta ops and maybe get rid of dag.From
* fix objectRange by embedding info in metadata returned by Pull so we don't have to pass the snapshot into Deleter and SequenceScanner
* remove runtime.DeleteQuery and make the deletions a parameter inside query
* use slice terminology instead of "partition"

PR #4213, opened 2022-11-16
Convert scan planner to runtime ops

### LINKED ISSUE #4163: refactor lake scheduler and indexer to be part of DAG
It will be easier to integrate the vcache and support clustering if we changed the logic implemented by the lake scheduler which calls PullScanTask to instead be implemented as dag operators.  In this approach, the scheduler would be a zbuf.Puller and return the data objects and zed values, which will be expected by downstream objects like a zng-scanner, vcache-projection, search index pruner, etc.

This will also simplify tests of all the various optimizations as we can use zc to simply output the dag representing pushdowns etc.

mccanne:
This commit refactors all the Go native logics that schedules data scans as elements of the run-time DAG.  This will make it easier to extend the scanning logic with new optimizations like use of the vector cache and more sophisticated search logic. It will also make it easier to distribute the DAG across a cluster as the objects to be scanned can now be serialized over the network as Zed without the need for a specialiead protocol.

These changes represent a first step at the refactoring where we have retained as much as possible the pre-existing logic.  Once this has been merged, we will refactor further in a subsequent PR to further simplify the logic, e.g., with some simplifying changes to how deletion and search-index lookups work.  Also, we will introduce package dag components for the new runtime ops to further simplify debugging and tests.  This will be important as we add further complexity and optimiztion to the scanning and pushdown logic.

These planned changes are documented in #4212.

Closes #4163


zbuf/scanner.go
```diff
@@ -93,6 +103,20 @@ func NewScanner(ctx context.Context, r zio.Reader, filterExpr Filter) (Scanner,
...
+	s, err := newScanner(ctx, r, filterExpr)
+	if err != nil {
+		return nil, err
+	}
+	if stringer, ok := r.(fmt.Stringer); ok {
+		s = NamedScanner(s, stringer.String())
+	}
+	return s, nil
+}
+
```
nwt, line 119: Nit: Put this before newScanner.

zbuf/scanner.go
```diff
@@ -163,3 +187,42 @@ func (n *namedScanner) Pull(done bool) (Batch, error) {
...
+var _ Scanner = (*MultiScanner)(nil)
+
+func NewMultiScanner(scanners ...Scanner) *MultiScanner {
+	return &MultiScanner{
+		scanners: scanners,
+	}
+}
+
+func (m *MultiScanner) Pull(done bool) (Batch, error) {
```
nwt, line 205: Nits: No point exporting this type. And call the function MultiScanner for symmetry with zio.ConcatReader and zio.MultiWriter.
```suggestion
func MultiScanner(scanners ...Scanner) Scanner {
	return &multiScanner{
		scanners: slices.Clone(scanners),
	}
}

type multiScanner struct {
	scanners []Scanner
	progress Progress
	current  Scanner
}

func (m *multiScanner) Pull(done bool) (Batch, error) {
```

zbuf/scanner.go
```diff
@@ -163,3 +187,42 @@ func (n *namedScanner) Pull(done bool) (Batch, error) {
...
+		batch, err := m.current.Pull(done)
+		if err != nil {
+			return nil, err
+		}
+		if batch != nil {
+			return batch, nil
+		}
+		m.progress.Add(m.current.Progress())
+		m.current = nil
+	}
```
nwt, line 223: Nit: Don't really need m.current.
```suggestion
	for len(m.scanners) > 0 {
		if batch, err := m.scanners[0].Pull(done); batch != nil || err != nil {
			return batch, err
		}
		m.progress.Add(m.scanners[0].Progress())
		m.scanners = m.scanners[1:]
	}
	return nil, nil
```

compiler/kernel/op.go
```diff
@@ -42,14 +43,24 @@ var ErrJoinParents = errors.New("join requires two upstream parallel query paths
 type Builder struct {
 	pctx     *op.Context
 	source   *data.Source
-	planners map[dag.Source]from.Planner
+	listers  map[dag.Source]*meta.Lister
+	pools    map[dag.Source]*lake.Pool
+	progress *zbuf.Progress
```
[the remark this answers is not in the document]
mccanne: After the compilation is done, GC could reclaim the compiler and leave behind the *zbuf.Progress, though we maybe need to make sure the compiler pointer goes away during the query.  It seemed distasteful to me to have runtime data structures point into the (ephemeral) Builder struct.

compiler/kernel/op.go
```diff
@@ -42,14 +43,24 @@ var ErrJoinParents = errors.New("join requires two upstream parallel query paths
...
+		pctx:    pctx,
+		source:  source,
+		listers: make(map[dag.Source]*meta.Lister),
+		pools:   make(map[dag.Source]*lake.Pool),
+		progress: &zbuf.Progress{
+			BytesRead:      0,
+			BytesMatched:   0,
+			RecordsRead:    0,
+			RecordsMatched: 0,
+		},
```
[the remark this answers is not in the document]
mccanne: see above

compiler/kernel/op.go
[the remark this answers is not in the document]
mccanne: Also, this will need to change when we distribute the scanners as the stats will need to flow back over the network to be consolidated and returned to the client.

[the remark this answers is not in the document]
nwt: 👍

runtime/op/meta/slicer.go
```diff
@@ -46,31 +46,46 @@ func PartitionObjects(objects []*data.Object, o order.Which) []meta.Partition {
...
 
-func (s *stack) pop() meta.Partition {
+/* unused right now
+func (s *stack) pop() Range {
 	n := len(*s)
 	p := (*s)[n-1]
 	*s = (*s)[:n-1]
 	return p
 }
+*/
```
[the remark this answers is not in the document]
mccanne: I think this will come back when we improve the algorithm.

compiler/kernel/op.go
nwt, line 63: Nit: No need to initialize fields to the zero value.
```suggestion
		progress: &zbuf.Progress{},
```

zbuf/scanner.go
mccanne: You need to do repeated pulls from current until EOS before you move on to the next scanner.

nwt: This'll do that! Clearer this way?
```go
	for len(m.scanners) > 0 {
		batch, err := m.scanners[0].Pull(done)
		if batch != nil || err != nil {
			return batch, err
		}
		m.progress.Add(m.scanners[0].Progress())
		m.scanners = m.scanners[1:]
	}
	return nil, nil
```

nwt (approved):

PR #4218, opened 2022-11-17
simplify scans of snapshot metadata

mccanne:
This commit simplifies how snapshot metadata is scanned by embracing the fact that snapshots are stored entirely in memory.  To this end, we removed the channels that streamed metadata as the original intention was that the snapshot would not necessarily fit entirely in memory. However, we can go a long way toward huge Zed lakes with the in-memory constraint and when we want to scale further, we can work a design that is consistent with where the Zed lake architecture lands in the future.

We also removed the unnecessary calls the zbuf.MultiScanner when there was only one scanner and updated some functions that did not need to be exported.

This is part of #4212

### DIFF
--- a/runtime/op/meta/lister.go
+++ b/runtime/op/meta/lister.go
@@ -8,7 +8,6 @@ import (
 	"github.com/brimdata/zed/lake"
 	"github.com/brimdata/zed/lake/commits"
 	"github.com/brimdata/zed/lake/data"
-	"github.com/brimdata/zed/lake/index"
 	"github.com/brimdata/zed/order"
 	"github.com/brimdata/zed/runtime/expr"
 	"github.com/brimdata/zed/runtime/expr/extent"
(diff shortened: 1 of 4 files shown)


runtime/op/meta/lister.go
```diff
@@ -69,58 +66,30 @@ func (l *Lister) Snapshot() commits.View {
...
-		val, err := l.marshaler.Marshal(part)
+	if l.err != nil {
+		return nil, l.err
+	}
+	if l.parts == nil {
+		var err error
+		l.parts, err = sortedPartitions(l.snap, l.pool.Layout, l.filter)
 		if err != nil {
+			l.err = err
 			return nil, err
```
nwt, line 79:
```suggestion
		l.parts, l.err = sortedPartitions(l.snap, l.pool.Layout, l.filter)
		if l.err != nil {
			return nil, l.err
```

nwt (approved):
So nice!

PR #4231, opened 2022-11-30
manage monitor: Fix premature thread exit

mattnibs:
This pr fixes an issue with manage monitor where newly ingested data would get ignored because the commit arrived after the object/index scans started. Get the current branch HEAD before the task then when finished run again if there is no remaining work but the task HEAD is out of date.

### DIFF
--- a/cmd/zed/manage/lakemanage/lakemanage.go
+++ b/cmd/zed/manage/lakemanage/lakemanage.go
@@ -31,8 +31,12 @@ func Update(ctx context.Context, lk lakeapi.Interface, conf Config, logger *zap.
 	for _, branch := range branches {
 		branch := branch
 		group.Go(func() error {
+			head, err := branch.head(ctx)
+			if err != nil {
+				return err
+			}
 			for _, task := range branch.tasks {
-				if _, err := task.run(ctx); err != nil {
+				if _, err := task.run(ctx, head); err != nil {
 					task.logger().Error("task error", zap.Error(err))
 					return err
 				}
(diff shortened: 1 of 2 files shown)


mattnibs:
This pr won't totally fix the race here, as there could be a commit in between the final head check and the thread exiting. Its an edge case IMHO this is good enough for now.

cmd/zed/manage/lakemanage/lakemanage.go
```diff
@@ -192,20 +197,44 @@ func (t *thread) run() {
...
+				// to see if there are any new commits since the task was run.
+				current, ok := t.head()
+				if !ok {
+					return
+				}
+				if current != head {
+					continue
+				}
 				t.task.logger().Info("thread exiting")
 				return
```
nwt, line 220: Nit: Save head across loop iterations to avoid batch to back t.branch.head calls.
```go
	 	var head ksuid.KSUID
		for t.ctc.Err() == nil {
			current, err := t.branch.head(t.ctx)
			if err != nil {
				t.task.logger().Error("error fetching branch head", zap.Error(err))
				return
			}
			if current == head {
				t.task.logger().Info("thread exiting")
				return
			}
			head = current
			next, err := t.task.run(t.ctx, head)
			if err != nil {
				t.task.logger().Error("thread exited with error", zap.Error(err))
				return
			}
			if next == nil {
				// This means there's no further work, but before exiting check
				// to see if there are any new commits since the task was run.
				continue

nwt (approved):

cmd/zed/manage/lakemanage/lakemanage.go
mattnibs: Hell yeah!

PR #4235, opened 2022-12-02
service: support loading Parquet via API

### LINKED ISSUE #4228: Add backend support for "parquet" input
Repro is with GA Zed tagged `v1.3.0`.

Parquet can currently be imported via `-i parquet` with the Zed CLI tooling. However, the Zed CLI tooling currently handles all input by first turning it into ZNG within the client and posting it to the backend as ZNG. Meanwhile, the Brim app and [lake API](https://zed.brimdata.io/docs/lake/api#branches) both post data in native formats and rely on the backend to read in that format. Therefore to have feature parity between our different clients & API it would be ideal to add Parquet load support in the backend as well.

How things look at the moment in GA Zed tagged `v1.3.0` and the [dns.parquet](https://github.com/brimdata/zed/blob/main/zio/parquetio/ztests/dns.parquet) test data from the Zed repo:

```
$ zed -version
Version: v1.3.0

{"version":"v1.3.0"}

$ zed create foo
pool created: foo 2IExZje6MXr9psNAiOBWpjkUaNc

$ zed use foo
Switched to branch "main" on pool "foo"

$ curl -X POST -H 'Accept: application/json' -H 'Content-Type: application/x-parquet' -d @dns.parquet http://localhost:9867/pool/foo/branch/main
{"type":"Error","kind":"invalid operation","error":"reader cannot seek"}
```

This shows the same error message as #4199 and might very well be a duplicate of that.

nwt:
When loading data via POST /pool/{pool}/branch/{branch}, if the request content type is application/x-parquet, copy the request body to a temporary file for the benefit of the Parquet reader, which needs a reader that implements io.ReaderAt and io.Seeker.

In the future, we should probably add a way to disable this or limit file size to prevent an adversary from filling the file system. ~~I'll create an issue for this.~~ See #4236.

Closes #4228.

### DIFF
--- a/service/handlers.go
+++ b/service/handlers.go
@@ -2,7 +2,9 @@ package service
 
 import (
 	"errors"
+	"io"
 	"net/http"
+	"os"
 	"time"
 
 	"github.com/brimdata/zed"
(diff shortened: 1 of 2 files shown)


service/handlers.go
```diff
@@ -366,13 +368,36 @@ func handleBranchLoad(c *Core, w *ResponseWriter, r *Request) {
...
+		//
+		// TODO: Add a way to disable this or limit file size.
+		f, err := os.CreateTemp("", "zed-serve-load-")
+		if err != nil {
+			w.Error(err)
+			return
+		}
+		defer f.Close()
+		defer os.Remove(f.Name())
+		if _, err := io.Copy(f, reader); err != nil {
```
mccanne, line 384: I would wrap this in a LimitedReader and pick a big number as a const and we can worry about configuring the limit later.  Maybe 10GB?

mccanne (approved):

service/handlers.go
nwt: I created #4236 for this. I think what we want is a way to disable the limit for Zui where this doesn't create a new threat (since anyone who can talk to the service can already fill the file system by loading data).

PR #4247, opened 2022-12-05
network_of: Allow ip value as mask argument

### LINKED ISSUE #4248: network_of() should accept a written-out netmask
Repro is with Zed commit 46f8877.

A community user recently inquired in a [public Slack thread](https://brimdata.slack.com/archives/C010DR0HHMF/p1670255425294199):

> am i abusing zui when trying this: `anet:=net(Subnet+"/"+"255.255.255.0")`, apparently netmask has to be /bitsite (eg. /24) to work. I'm trying to do some ETL during load. Do I have to transform this first to read like this: `anet:=net(Subnet+"/24")`? Reason I'm asking, I've multiple csv containing IP and netmask in separate columns.

We'd like to be able to point this user at the [`network_of()`](https://zed.brimdata.io/docs/language/functions/network_of) function, but there's currently a mismatch with the user's needs. As the docs show, right now its `mask` parameter accepts either an integer parameter, specifying the count of consecutive leading 1-bits (from left to right) in the network mask:

```
$ zq -version
Version: v1.3.0-11-g46f88777

$ echo '10.1.2.129' | zq -z 'yield network_of(this, 25)' -
10.1.2.128/25
```

...or it takes a `net` parameter, which in Zed has to include both the address _and_ netmask portion.

```
$ echo '10.1.2.129' | zq -z 'yield network_of(this, 255.255.255.128/25)' -
10.1.2.128/25
```

However, the user's incoming incoming CSV data has only the written-out netmask like the `255.255.255.0` shown above. This is a valid Zed `ip` type value, so it seems that `network_of()` could be enhanced to accept this and treat it the same as it currently does the integer value `24`.

mattnibs:
Closes #4248

### DIFF
--- a/docs/language/functions/network_of.md
+++ b/docs/language/functions/network_of.md
@@ -17,7 +17,7 @@ network is inferred from `val`, which in this case, must be an IPv4 address.
 
 ### Examples
 
-Compute the network address of an IP using an `ip`-type mask argument:
+Compute the network address of an IP using an `ip` mask argument:
 ```mdtest-command
 echo '10.1.2.129' | zq -z 'yield network_of(this, 255.255.255.128)' -
 ```
--- a/docs/language/functions/network_of.md
+++ b/docs/language/functions/network_of.md
@@ -5,7 +5,7 @@
 ### Synopsis
 
 ```
-network_of(val: ip [, mask: net|int|uint]) -> net
+network_of(val: ip [, mask: net|ip|int|uint]) -> net
 ```
 ### Description
 
--- a/docs/language/functions/network_of.md
+++ b/docs/language/functions/network_of.md
@@ -5,19 +5,28 @@
 ### Synopsis
 
 ```
-network_of(val: ip [, mask: net|ip|int|uint]) -> net
+network_of(val: ip [, mask: ip|net|int|uint]) -> net
 ```
 ### Description
 
 The _network_of_ function returns the network of the IP address given
 by `val` as determined by the optional `mask`.  If `mask` is an integer rather
 than a net, it is presumed to be a network prefix of the indicated length.
-If `mask` is ommited, then a class A (8 bit), B (16 bit), or C (24 bit)
+If `mask` is omitted, then a class A (8 bit), B (16 bit), or C (24 bit)
 network is inferred from `val`, which in this case, must be an IPv4 address.
 
 ### Examples
 
-Compute the network address of an IP given a network mask argument:
+Compute the network address of an IP using an `ip`-type mask argument:
+```mdtest-command
+echo '10.1.2.129' | zq -z 'yield network_of(this, 255.255.255.128)' -
+```
+=>
+```mdtest-output
+10.1.2.128/25
+```
+
+Compute the network address of an IP using a `net`-type mask argument:
 ```mdtest-command
 echo '10.1.2.129' | zq -z 'yield network_of(this, 255.255.255.128/25)' -
 ```
@@ -25,6 +34,7 @@ echo '10.1.2.129' | zq -z 'yield network_of(this, 255.255.255.128/25)' -
 ```mdtest-output
 10.1.2.128/25
 ```
+
 Compute the network address of an IP given an integer prefix argument:
 ```mdtest-command
 echo '10.1.2.129' | zq -z 'yield network_of(this, 25)' -
@@ -50,4 +60,13 @@ echo 1 | zq -z 'yield network_of(this)' -
 =>
 ```mdtest-output
 error("network_of: not an IP")
+```
+
+Network masks must be contiguous:
+```mdtest-command
+echo '10.1.2.129' | zq -z 'yield network_of(this, 255.255.128.255)' -
+```
+=>
+```mdtest-output
+error("network_of: mask \"255.255.128.255\" is non-contiguous")
 ```
--- a/docs/language/functions/network_of.md
+++ b/docs/language/functions/network_of.md
@@ -26,7 +26,7 @@ echo '10.1.2.129' | zq -z 'yield network_of(this, 255.255.255.128)' -
 10.1.2.128/25
 ```
 
-Compute the network address of an IP using a `net`-type mask argument:
+Compute the network address of an IP using a `net` mask argument:
 ```mdtest-command
 echo '10.1.2.129' | zq -z 'yield network_of(this, 255.255.255.128/25)' -
 ```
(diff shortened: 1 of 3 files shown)


runtime/expr/function/ip.go
```diff
@@ -30,15 +30,22 @@ func (n *NetworkOf) Call(ctx zed.Allocator, args []zed.Value) *zed.Value {
...
 			}
 			mask = cidrMask.Mask
-		} else if zed.IsInteger(id) {
+		case id == zed.IDIP:
+			ip := zed.DecodeIP(body)
+			if ip.Is6() {
+				return newErrorf(n.zctx, ctx, "network_of: mask cannot be an IPv6 address")
+			}
+			b := ip.As4()
+			mask = net.IPv4Mask(b[0], b[1], b[2], b[3])
```
mccanne, line 47: Can't this just be...?
```
mask = net.IPMask(zed.DecodeIP(body).AsSlice())
```
Also, I noticed that allows a non-contiguous mask into a zed network value which then formats it in a non-ZSON compatible form.  So maybe we should detect non-contiguous masks?

runtime/expr/ztests/network-of.yaml
```diff
@@ -1,7 +1,10 @@
...
 
 input: |
   {a:10.1.2.129,mask:255.255.255.128/25}
 
 output: |
-  {a:10.1.2.129,mask:255.255.255.128/25,n1:10.1.2.128/25,n2:10.1.2.0/24}
+  10.1.2.128/25
+  10.1.2.0/24
+  10.1.2.128/25
+  error("network_of: mask \"255.128.255.255\" is not in canonical form")
```
philrz, line 10:
```suggestion
  error("network_of: mask \"255.128.255.255\" is non-contiguous")
```

runtime/expr/function/ip.go
```diff
@@ -30,15 +30,21 @@ func (n *NetworkOf) Call(ctx zed.Allocator, args []zed.Value) *zed.Value {
...
 			if !bytes.Equal(cidrMask.IP, cidrMask.Mask) {
 				return newErrorf(n.zctx, ctx, "network_of: network arg not a cidr mask")
 			}
 			mask = cidrMask.Mask
-		} else if zed.IsInteger(id) {
+		case id == zed.IDIP:
+			ip := zed.DecodeIP(body)
+			mask = net.IPMask(ip.AsSlice())
+			if ones, bits := mask.Size(); ones == 0 && bits == 0 {
+				return newErrorf(n.zctx, ctx, "network_of: mask %q is not in canonical form", ip.String())
```
philrz, line 45:
```suggestion
				return newErrorf(n.zctx, ctx, "network_of: mask %q is non-contiguous", ip.String())
```

philrz:
@mattnibs: I pushed a commit with updates to the user-facing docs that includes new examples. It'll run clean if you accept my PR suggestion about the phrasing "non-contiguous" instead of "not in canonical form".

Also, I admit that the `net`-type argument here has always eluded me. Is there a use case where the address and mask portion of the `net` can be different and provide a meaningful result? The only way I've ever thought to use it is with the mask and prefix length being a redundant expression the same, e.g., the way the examples in my update show mask arg of `25` and `255.255.255.128` both giving the same result as `255.255.255.128/25`. If they're always redundant, I'd wonder about removing the `net`-type argument altogether since I think it has the potential to confuse, though as long as the examples show it being redundant I guess it's not hurting anything. OTOH if there _is_ a meaningful use case where they're _not_ redundant, I'd ask that someone chime in with an example that shows it.

(cc: @mccanne and @nwt in case they've got a take on any of the above.)

docs/language/functions/network_of.md
```diff
@@ -5,26 +5,36 @@
...
 by `val` as determined by the optional `mask`.  If `mask` is an integer rather
 than a net, it is presumed to be a network prefix of the indicated length.
-If `mask` is ommited, then a class A (8 bit), B (16 bit), or C (24 bit)
+If `mask` is omitted, then a class A (8 bit), B (16 bit), or C (24 bit)
 network is inferred from `val`, which in this case, must be an IPv4 address.
 
 ### Examples
 
-Compute the network address of an IP given a network mask argument:
+Compute the network address of an IP using an `ip`-type mask argument:
```
nwt, line 20:
```suggestion
Compute the network address of an IP using an `ip` mask argument:
```

docs/language/functions/network_of.md
```diff
@@ -5,26 +5,36 @@
...
+```mdtest-command
+echo '10.1.2.129' | zq -z 'yield network_of(this, 255.255.255.128)' -
+```
+=>
+```mdtest-output
+10.1.2.128/25
+```
+
+Compute the network address of an IP using a `net`-type mask argument:
```
nwt, line 29:
```suggestion
Compute the network address of an IP using a `net` mask argument:
```

runtime/expr/function/ip.go
nwt, line 45: Nit: No real need to quote this, and I think the embedded quotes it produces make the error string less readable.
```suggestion
				return newErrorf(n.zctx, ctx, "network_of: mask %s is not in canonical form", ip.String())
```

docs/language/functions/network_of.md
```diff
@@ -51,3 +61,12 @@ echo 1 | zq -z 'yield network_of(this)' -
...
 error("network_of: not an IP")
 ```
+
+Network masks must be contiguous:
+```mdtest-command
+echo '10.1.2.129' | zq -z 'yield network_of(this, 255.255.128.255)' -
+```
+=>
+```mdtest-output
+error("network_of: mask \"255.255.128.255\" is non-contiguous")
```
philrz, line 71:
```suggestion
error("network_of: mask 255.255.128.255 is non-contiguous")
```

Derp. Forgot to match this up with @nwt's nit.

nwt (approved):

mattnibs:
> @mattnibs: I pushed a commit with updates to the user-facing docs that includes new examples. It'll run clean if you accept my PR suggestion about the phrasing "non-contiguous" instead of "not in canonical form".
> 
> Also, I admit that the `net`-type argument here has always eluded me. Is there a use case where the address and mask portion of the `net` can be different and provide a meaningful result? The only way I've ever thought to use it is with the mask and prefix length being a redundant expression the same, e.g., the way the examples in my update show mask arg of `25` and `255.255.255.128` both giving the same result as `255.255.255.128/25`. If they're always redundant, I'd wonder about removing the `net`-type argument altogether since I think it has the potential to confuse, though as long as the examples show it being redundant I guess it's not hurting anything. OTOH if there _is_ a meaningful use case where they're _not_ redundant, I'd ask that someone chime in with an example that shows it.
> 
> (cc: @mccanne and @nwt in case they've got a take on any of the above.)

@philrz agreed, the net argument doesn't make a lot of sense. I think we should drop it.

PR #4253, opened 2022-12-08
network_of: drop net mask argument

mattnibs:


docs/language/functions/network_of.md
```diff
@@ -26,9 +26,9 @@ echo '10.1.2.129' | zq -z 'yield network_of(this, 255.255.255.128)' -
 10.1.2.128/25
 ```
 
-Compute the network address of an IP using a `net` mask argument:
+Compute the network address of an IP using an `ip` mask argument:
 ```mdtest-command
-echo '10.1.2.129' | zq -z 'yield network_of(this, 255.255.255.128/25)' -
+echo '10.1.2.129' | zq -z 'yield network_of(this, 255.255.255.128)' -
```
philrz, line 31: Instead of adjusting this example you probably should just remove it entirely, as I'd already added an example that shows it working with an `ip` mask, so at the moment in this branch you've got redundant examples.

nwt: What @philrz said.

mattnibs: Whoops!

nwt (approved):

PR #4263, opened 2022-12-10
replace net.IPNet with netip.Prefix

nwt:
Creating a net.IPNet allocates memory; creating a netip.Prefix does not.


mattnibs (approved):
Awesome!

PR #4264, opened 2022-12-11
User defined functions

### LINKED ISSUE #4026: user-defined functions (UDF)
The time has come to implement a simple form of user-defined function.

We can debate the syntax but I propose this pythonic form:
```
def f(x,y,...): ( <any expression>)
```
We need parens to define the scope of the expression (since we don't use whitespace
scoping like python)

In the initial implementation no recursion or control structures will be allowed.
This can come later.  For now, we will support just the existing expression syntax.

mattnibs:
Add ability to for users to define simple functions that have a body for evaluating a singular expression against variables defined in scope.

Closes #4026


runtime/expr/ztests/udf.yaml
```diff
@@ -0,0 +1,14 @@
...
+  yield fib(this)
+
+input: |
+  3
+  10
+
+output: |
+  2
+  55
+
```
nwt, line 14:
```suggestion
```

compiler/ast/ast.go
```diff
@@ -194,6 +198,22 @@ func (*OverExpr) ExprAST() {}
...
+type ConstDecl struct {
+	Kind string `json:"kind" unpack:""`
+	Name string `json:"name"`
+	Expr Expr   `json:"expr"`
+}
+
+type FuncDecl struct {
+	Kind   string `json:"kind" unpack:""`
+	Name   string `json:"name"`
+	Params []*ID  `json:"params"`
```
nwt, line 210: Any reason not to make this a []string?

compiler/ast/dag/expr.go
```diff
@@ -38,6 +38,12 @@ type (
 		LHS  Expr   `json:"lhs"`
 		RHS  string `json:"rhs"`
 	}
+	Func struct {
+		Kind    string   `json:"func" unpack:""`
+		Name    string   `json:"name"`
+		Formals []string `json:"formals"`
```
nwt, line 44: Calls this Params so it matches ast.Func.Params.

nwt:
@mattnibs: This is beautiful work!

compiler/kernel/op.go
```diff
@@ -308,6 +310,11 @@ func splitAssignments(assignments []expr.Assignment) (field.List, []expr.Evaluat
 }
 
 func (b *Builder) compileSequential(seq *dag.Sequential, parents []zbuf.Puller) ([]zbuf.Puller, error) {
+	for _, fn := range seq.Funcs {
+		if err := b.compileFunc(fn); err != nil {
+			return nil, err
+		}
+	}
```
nwt, line 317: I think we should consider compiling all of the functions together, as in b.compileFuncs(seq.Funcs). That would let us add all of the names to b.funcs before compiling any of the bodies. And that would in turn allow both declaring a caller before a callee and mutual recursion (i.e., a calls b and b calls a).

runtime/expr/udf.go
```diff
@@ -0,0 +1,41 @@
+package expr
+
+import (
+	"github.com/brimdata/zed"
+	"golang.org/x/exp/slices"
+)
+
+const maxStackSize = 10_000
```
nwt, line 8: Nit: I think maxStackDepth would be a better name for this.

compiler/parser/parser.peg
```diff
@@ -49,28 +49,39 @@
...
     }
   / "type" _ id:(IdentifierName / QuotedString) __ "=" __ typ:Type {
       RETURN(MAP(
+        "kind":"ConstDecl",
         "name":id,
         "expr":MAP("kind":"TypeValue","value":MAP("kind":"TypeDef","name":id,"type":typ))))
     }
 
+FuncDecl
+  = "def" _ id:IdentifierName __ "(" __ params:IdentifierList __ ")" __ ":" __ "(" __ expr:Expr __ ")" {
```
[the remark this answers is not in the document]
mattnibs: I wonder if `func` is more clear than `fn`. But I would defer to @mccanne for this one since the issue he created specifies `def`.

compiler/ast/ast.go
mattnibs: None, let's do that.

compiler/kernel/op.go
mattnibs: Good idea.

compiler/parser/parser.peg
```diff
@@ -49,28 +49,39 @@
 start = __ ast:Sequential __ EOF { RETURN(ast) }
 
 Sequential
-  = consts:Consts __ first:Operation rest:SequentialTail* {
-      RETURN(MAP("kind": "Sequential", "ops": PREPEND(first, rest), "consts":consts))
+  = decls:Decls __ first:Operation rest:SequentialTail* {
+      RETURN(MAP("kind": "Sequential", "ops": PREPEND(first, rest), "decls":decls))
```
nwt, line 53:
```suggestion
      RETURN(MAP("kind": "Sequential", "ops": PREPEND(first, rest), "decls": decls))
```

runtime/expr/udf.go
```diff
@@ -0,0 +1,41 @@
...
+	stack int
+	vars  []zed.Value
+}
+
+func (f *frame) Vars() []zed.Value {
+	return f.vars
+}
+
+func (f *frame) exit() {
+	f.stack -= 1
```
nwt, line 40:
```suggestion
	f.stack--
```

nwt (commented):
We should add test for mutual recursion.

runtime/expr/ztests/udf-mutual-recursion.yaml
```diff
@@ -0,0 +1,21 @@
...
+input: |
+  10
+  11
+
+output: |
+  true
+  false
+  false
+  true
+
```
nwt, line 21:
```suggestion
```

runtime/expr/udf.go
```diff
@@ -0,0 +1,41 @@
...
+	Body Evaluator
+}
+
+func (u *UDF) Call(ectx zed.Allocator, args []zed.Value) *zed.Value {
+	stack := 1
+	if f, ok := ectx.(*frame); ok {
+		stack += f.stack
+	}
+	if stack > maxStackDepth {
+		panic("stack overflow")
```
nwt, line 20: Rather than panic here, do you think we should return a Zed error that includes the name of the current function?

nwt (commented):
A few things we should add:
1. Compilation error tests (function declared twice, calling something that isn't a function, wrong number of arguments in call)
2. Support (and tests) for this in zfmt
3. Documentation in docs/language/overview.md

I took a stab the documentation and am happy to share that by pushing a commit or otherwise.

philrz:
The docs changes from @nwt read well to me. However, I did a scratch deploy of the docs site based on this branch and it confirmed that a bunch of other docs that link to the moved sections will now have broken links. Since @mattnibs was ok with having the docs changes pushed directly to the branch, I'll similarly push a commit of link fixes.

nwt (approved):
:shipit:!

philrz:
Verified in Zed commit 5f4c4e8.

Here's an example that uses two UDF functions. One uses a regular expression to extract the area code from a phone number string and the other computes a factorial recursively.

```
$ cat functions.zed 
func factorial(n): (
  n == 1 ? 1 : n*factorial(n-1)
 )

func area_code(str): (
  regexp(/^\((\d{3})\) *(\d{3}\-\d{4})$/, str)[1]
)

$ zq -version
Version: v1.3.0-31-g5f4c4e81

$ echo '{"num": 5, "phone": "(415) 555-1212"}' | zq -Z -I functions.zed 'yield {num, fac:factorial(num), phone, area_code:area_code(phone)}' -
{
    num: 5,
    fac: 120,
    phone: "(415) 555-1212",
    area_code: "415"
}
```

Documentation is at https://zed.brimdata.io/docs/next/language/overview#4-func-statements.

Thanks @mattnibs!

PR #4270, opened 2022-12-14
zio/anyio: auto-detect Parquet and VNG

### LINKED ISSUE #2517: Auto-detect all input formats for files

nwt:
Both formats require a seekable reader so this only works with files and URIs that are not gzip-compressed.  It does not work with standard input, fifos, or gzip-compressed data.

Closes #2517.


zio/anyio/reader.go
```diff
@@ -32,6 +34,36 @@ func NewReaderWithOpts(zctx *zed.Context, r io.Reader, opts ReaderOpts) (zio.Rea
...
+				return nil, err
+			}
+		} else {
+			parquetErr = err
+			vngErr = err
+		}
+		parquetErr = fmt.Errorf("parquet: %w", parquetErr)
+		vngErr = fmt.Errorf("vng: %w", vngErr)
+	} else {
+		parquetErr = errors.New("parquet: reader cannot seek")
```
mccanne, line 63: "reader cannot seek" is describing the implementation not the supported types of inputs begging the question "what's a reader?".  How about `seekable input required for auto-detection (consider "-i parquet")`

nwt: @mccanne: Good point, though "consider '-i parquet'" isn't quite the right advice since that won't make the input seekable, and that's the requirement here. I'll go with "seekable input required for auto-detection" unless you've got other ideas.

nwt: Done in f82fb695cb1b26ba86ccfab3abea21ece1376fbd, though the exact language I used is "auto-detection requires seekable input" to align with `lineErr := errors.New("line: auto-detection not supported")`.

mattnibs (approved):

PR #4274, opened 2022-12-19
Add journal snapshots

mattnibs:
Create snapshots of the journal table so the entire journal log need not be read when a branch tip has been updated. This greatly improves performance when creating a bunch of commits during small batch loads.

### DIFF
--- a/lake/journal/store.go
+++ b/lake/journal/store.go
@@ -4,12 +4,14 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io/fs"
 	"os"
 	"sync"
 	"time"
 
 	"github.com/brimdata/zed"
 	"github.com/brimdata/zed/pkg/storage"
+	"github.com/brimdata/zed/zio/zngio"
 	"github.com/brimdata/zed/zngbytes"
 	"github.com/brimdata/zed/zson"
 )
@@ -74,34 +76,41 @@ func (s *Store) load(ctx context.Context) error {
 		return err
 	}
 	s.mu.RLock()
-	at := s.at
+	current := s.at
 	s.mu.RUnlock()
-	if head == at {
+	if head == current {
 		return nil
 	}
-	r, err := s.journal.OpenAsZNG(ctx, zed.NewContext(), head, 0)
+	at, table, err := s.getSnapshot(ctx)
+	if err != nil && !errors.Is(err, fs.ErrNotExist) {
+		return err
+	}
+	r, err := s.journal.OpenAsZNG(ctx, zed.NewContext(), head, at)
 	if err != nil {
 		return err
 	}
 	defer r.Close()
-	table := make(map[string]Entry)
 	for {
-		rec, err := r.Read()
+		v, err := r.Read()
 		if err != nil {
 			return err
 		}
-		if rec == nil {
+		if v == nil {
 			now := time.Now()
 			s.mu.Lock()
 			s.table = table
 			s.at = head
 			s.loadTime = now
 			s.mu.Unlock()
+			// Reduce the amount of times we write snapshots to disk by only writing when there are 10 or more new entries since
+			// the last snapshot.
+			if head-at > 10 {
+				return s.putSnapshot(ctx, head, table)
+			}
 			return nil
-
 		}
 		var e Entry
-		if err := s.unmarshaler.Unmarshal(rec, &e); err != nil {
+		if err := s.unmarshaler.Unmarshal(v, &e); err != nil {
 			return err
 		}
 		switch e := e.(type) {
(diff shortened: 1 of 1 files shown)


nwt (commented):
lake/journal.Store only provides access to the current journal.Entry table, so keeping a snapshot of anything but the current table doesn't seem beneficial.

With that in mind, how about just overwriting any existing snapshot? (Since the corresponding journal head value wouldn't be part of the snapshot's path, we'd need to put it inside the snapshot, but that's easy enough.)

mattnibs:
> With that in mind, how about just overwriting any existing snapshot?

I can dig that, lemme give it a try.

mattnibs:
Actually @nwt I'm concerned about concurrent writes to the same file and this resulting in corrupt snapshots. For writes to the file system we could do the old write to tempfile then rewrite trick, but I'm not sure how concurrent writes work on s3...

nwt:
@mattnibs: You're right to be concerned!

I propose that we continue here with the assumption that pkg/storage.Engine.Put is atomic while working separately on #4277 to make that true.

lake/journal/store.go
```diff
@@ -74,34 +76,41 @@ func (s *Store) load(ctx context.Context) error {
...
+		if v == nil {
 			now := time.Now()
 			s.mu.Lock()
 			s.table = table
 			s.at = head
 			s.loadTime = now
 			s.mu.Unlock()
+			// Reduce the amount of times we write snapshots to disk by only writing when there are 10 or more new entries since
+			// the last snapshot.
+			if head-at > 10 {
```
nwt, line 107: Comment says "10 or more" but code says `head-at > 10`, and that means more than 10, right?

lake/journal/store.go
```diff
@@ -115,12 +124,70 @@ func (s *Store) load(ctx context.Context) error {
...
+		var e Entry
+		if err := s.unmarshaler.Unmarshal(v, &e); err != nil {
+			return at, nil, err
+		}
+		table[e.Key()] = e
+	}
+}
+
+func (s *Store) putSnapshot(ctx context.Context, at ID, table map[string]Entry) error {
+	// XXX This needs to be an atomic write for file systems.
```
nwt, line 162: #4277 covers this, right?

lake/journal/store.go
```diff
@@ -115,12 +124,70 @@ func (s *Store) load(ctx context.Context) error {
...
+	w, err := s.journal.engine.Put(ctx, s.snapshotURI())
+	if err != nil {
+		return err
+	}
+	defer w.Close()
+	zw := zngio.NewWriter(w)
```
nwt, line 168: We should call zw.Close on error and let that call w.Close.
```suggestion
	zw := zngio.NewWriter(w)
	defer zw.Close()
```

lake/journal/store.go
```diff
@@ -115,12 +124,70 @@ func (s *Store) load(ctx context.Context) error {
...
+	v := zed.NewValue(zed.TypeUint64, zed.EncodeUint(uint64(at)))
+	if err := zw.Write(v); err != nil {
```
nwt, line 170:
```suggestion
	if err := zw.Write(zed.NewUint64(at)); err != nil {
```

lake/journal/store.go
mattnibs: Yes, will update comment.

lake/journal/store.go
```diff
@@ -74,34 +76,41 @@ func (s *Store) load(ctx context.Context) error {
...
+	}
+	r, err := s.journal.OpenAsZNG(ctx, zed.NewContext(), head, at)
 	if err != nil {
 		return err
 	}
 	defer r.Close()
-	table := make(map[string]Entry)
 	for {
-		rec, err := r.Read()
+		v, err := r.Read()
```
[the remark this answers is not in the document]
mattnibs: Where's my head at?

lake/journal/store.go
mattnibs: Yeah, i'll reference the issue in the comment.

lake/journal/store.go
mattnibs: Tangential to this pr but, it's confusing that the zngio.Writer will close the writer passed to it but the reader will not.

lake/journal/store.go
```diff
@@ -115,12 +124,69 @@ func (s *Store) load(ctx context.Context) error {
...
+func (s *Store) getSnapshot(ctx context.Context) (ID, map[string]Entry, error) {
+	table := make(map[string]Entry)
+	r, err := s.journal.engine.Get(ctx, s.snapshotURI())
+	if err != nil {
+		return Nil, table, err
+	}
+	defer r.Close()
+	zr := zngio.NewReader(zed.NewContext(), r)
+	defer zr.Close()
+	v, err := zr.Read()
```
nwt, line 140: s/v/val/

lake/journal/store.go
```diff
@@ -115,12 +124,69 @@ func (s *Store) load(ctx context.Context) error {
...
+	if v == nil || err != nil {
+		return Nil, table, err
+	}
+	if v.Type.ID() != zed.IDUint64 {
+		return Nil, table, errors.New("corrupted journal snapshot")
+	}
+	at := ID(zed.DecodeUint(v.Bytes))
+	for {
+		v, err := zr.Read()
```
nwt, line 149: s/v/val/

lake/journal/store.go
```diff
@@ -115,12 +124,69 @@ func (s *Store) load(ctx context.Context) error {
...
+	}
+	zw := zngio.NewWriter(w)
+	defer zw.Close()
+	if err := zw.Write(zed.NewUint64(uint64(at))); err != nil {
+		return err
+	}
+	marshaler := zson.NewZNGMarshaler()
+	marshaler.Decorate(s.style)
+	for _, entry := range table {
+		v, err := marshaler.Marshal(entry)
```
nwt, line 175: s/v/val/

nwt (approved):

nwt (commented):
I could be mistaken but I don't think we have any tests that cover this.

mattnibs:
> I could be mistaken but I don't think we have any tests that cover this.

If we have a test that creates more than 10 pools or adds 10 commits to a pool we have coverage for this. Let me check tho.

philrz:
I've verified that this definitely makes a big difference. At Zed commit 3143419 that was right before the changes in this PR, posting 10,000 synthetic records as fast as possible took 38+ minutes.

```
{"version":"v1.3.0-39-g31434194"}

$ time ./record-poster.py --seed 1 --keyless 100 --exitafter 10000
...
real	38m30.412s
user	0m12.039s
sys	0m1.786s
```

CPU and memory use during the run:

image

The same at Zed commit 3d0a8c5 that includes the changes in this PR completed in just 50 seconds.

```
{"version":"v1.3.0-44-g3d0a8c5b"}

$ time ./record-poster.py --seed 1 --keyless 100 --exitafter 10000
...
real	0m50.091s
user	0m15.345s
sys	0m2.251s
```

image

FWIW, in addition to the memory usage being similar both runs, the storage consumed by the lake was the same in both (156 MB). Not that I was expecting differences, but I happened to check so am just reporting the data.

PR #4283, opened 2022-12-23
zed: rename ColumnBuilder to RecordBuilder

nwt:
This thing builds records so RecordBuilder seems like a better name.


mattnibs (approved):
Indeed, much better.

PR #4284, opened 2022-12-23
Mention that left/right joins are sometimes known as "outer"

philrz:
A community user recently inquired:

> Is there a way to perform outer joins in Zed?

I pointed them at the [join tutorial](https://zed.brimdata.io/docs/next/tutorials/join) doc and explained how what we call "left join" and "right join" are the same as what's also sometimes called "left outer join" and "right outer join" in SQL contexts. The user is all set now that they know this, but the exchange made he realize that we could stand to call it out explicitly in the tutorial just like they do at SQL references like https://www.w3schools.com/sql/sql_join_left.asp.

The syntax I'm using here render in Docusaurus like this:

image

### DIFF
--- a/docs/tutorials/join.md
+++ b/docs/tutorials/join.md
@@ -74,6 +74,10 @@ produces
 
 ## Left Join
 
+:::tip note
+In some databases a left join is called a _left outer join_.
+:::
+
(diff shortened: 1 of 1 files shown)


mattnibs (approved):
Fancy!

PR #4299, opened 2023-01-05
add VNG segment compression

nwt:
In vng/vector:

* Add a Segment.CompressionFormat field and accompanying CompressionFormatNone and CompressionFormatLZ4 constants

* Add a Segment.MemLength field indicating the segment's in-memory (i.e., uncompressed) length (in contrast with Segment.Length, its in-file length)

* Add (*Segment).Read, which can read both compressed and uncompressed segments

* Update (*Spiller).Write to write a compressed segment if compression is effective and an uncompressed segment if not


mattnibs (approved):
lgtm

mccanne:
Cool!

PR #4306, opened 2023-01-09
zed: rename Column type to Field

nwt:
The changes outside package zed are as limited as possible, leaving a similar renaming of identifiers in those packages for later.


value.go
```diff
@@ -215,14 +215,14 @@ func (r *Value) nth(column int) zcode.Bytes {
 	return zv
 }
 
-func (r *Value) Columns() []Column {
-	return TypeRecordOf(r.Type).Columns
+func (r *Value) Fields() []Field {
+	return TypeRecordOf(r.Type).Fields
 }
 
 func (v *Value) DerefByColumn(col int) *Value {
```
mattnibs, line 222: Should this be renamed `DerefByField`?

recordbuilder.go
```diff
@@ -189,10 +189,10 @@ func (r *RecordBuilder) Encode() (zcode.Bytes, error) {
 // TypedColumns takes an array of Types for the individual fields
 // and constructs an array of Columns that reflects the fullly
 // typed structure.  This is suitable for e.g. allocating a descriptor.
```
mattnibs, line 191:
```suggestion
// Fields takes an array of Types for the individual fields
// and constructs an array of Fields that reflects the fullly
// typed structure.  This is suitable for e.g. allocating a descriptor.
```

mattnibs (approved):

value.go
nwt: Indeed.

recordbuilder.go
nwt: Good catch!

value.go
nwt: On second thought, I think this should become DerefByIndex or FieldByIndex, so I'll leave it as is and we'll sort out the name in another PR.

PR #4308, opened 2023-01-10
Mention comments in language overview doc

### LINKED ISSUE #4302: Mention comments in Language Overview doc
A community zync user recently remarked:

> I recently discovered we can use comments `//` in Zed query code. I guess I should have read the ZSON docs more throughly and figure this out sooner...

ZSON and the Zed language have some overlap but are in many ways wholly separate topics. I can see that comments are not currently mentioned at all in the [Language Overview](https://zed.brimdata.io/docs/language/overview) doc where they surely should be. In a quick scan my first instinct is to put it at the end of the Introduction right after it's shown how the language can extend from simple one-liner searches to longer source files that one might maintain in GitHub.

philrz:
Fixes #4302

### DIFF
--- a/docs/language/overview.md
+++ b/docs/language/overview.md
@@ -95,6 +95,17 @@ The short-hand query from above might be typed into a search box while the
 latter query might be composed in a query editor or in Zed source files
 maintained in GitHub.  Both forms are valid Zed queries.
 
+To further ease the maintenance and readability of source files, comments
+beginning with `//` may appear in Zed.
+
+```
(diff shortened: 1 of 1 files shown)


mccanne (approved):
We should add multiline comments (`/* ... */`) at some point.  I think there was a problem a long time ago with the PEG rules.  Just need to go back and fix it.

philrz:
Thanks @mccanne. I've opened #4310 as a reminder to add them.

PR #4334, opened 2023-01-25
add -cors.origin flag to "zed serve"

### LINKED ISSUE #4297: "zed serve" flag to specify additional AllowedOrigins
As described in the [README](https://github.com/philrz/zed-datasource/blob/main/README.md) for the prototype Zed data source Grafana plugin, such a plugin cannot currently query an out-of-the-box Zed service because the CORS configuration does not permit requests from the `http://localhost:3000` origin.

As an interim hack, I was able to get it working in branch by adding the additional entry for `http://localhost:3000` here:

https://github.com/brimdata/zed/blob/52c5061b6fbbe9abd1dd754d240104cca647c5aa/service/middleware.go#L33

In a discussion with the team, we agreed that we should introduce a flag to allow this, e.g., `zed serve -origin http://localhost:3000`. It seems like a flag that the user should be able to invoke multiple times on the same command line so multiple additional origins can be specified, similar to what we do with multiple includes on `zed query -I`.

nwt:
The -cors.origin flag specifies a CORS allowed origin.  The flag may be repeated.

This change removes the two baked-in allowed origins, \*.observableusercontent.com and localhost, and replaces them with a default allowed origin of \*.  As a consequence, "zed serve" with no -cors.origin flag behaves like "zed serve -cors.origin '*'".

Closes #4297.


philrz:
I've tested out this branch at commit 990520e and confirmed that it does the trick for allowing connections from the Grafana data source proxy at origin `http://localhost:3000`. Therefore it's a functional 👍 from me!

cmd/zed/serve/command.go
```diff
@@ -53,6 +53,8 @@ func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
 	c.conf.Version = cli.Version
 	c.logflags.SetFlags(f)
 	f.IntVar(&c.brimfd, "brimfd", -1, "pipe read fd passed by brim to signal brim closure")
+	c.conf.CORSAllowedOrigins = []string{"*.observableusercontent.com", "localhost"}
+	f.Var((*cli.CommaStringsFlag)(&c.conf.CORSAllowedOrigins), "cors.origins", "comma-separated list of CORS allowed origins")
```
mattnibs, line 57: I don't really like doing comma strings, can we rather just have the ability for the user to specify the flag multiple times? I know we did this somewhere...

nwt: Yes, happy to make this change.

philrz: Ah yes, this one:

```
    -I source file containing Zed query text (may be used multiple times)
```

nwt: @mattnibs: Done in 0ecdef640cd450f951233341f7d65168e52e9443.

service/middleware.go
```diff
@@ -30,9 +30,7 @@ func requestIDMiddleware() mux.MiddlewareFunc {
 	}
 }
 
-var allowedOrigins = []string{"*.observableusercontent.com", "localhost"}
```
nwt, line 33: Note that I've removed these baked-in values!

philrz:
I've re-tested this in the branch at commit 1452539 and it's still a functional 👍 for me. The removal of the baked-in values actually works great for my use case since it means that my target user won't even need to invoke the new option, though it's nice to have it there in case someone wants to lock down the service more tightly.

service/middleware.go
mattnibs: Cool, we need to make sure brim uses this new command argument to add these origins.

mattnibs (approved):

service/middleware.go
nwt: I think it'll work fine without them.

We will however need to add https://*.observableusercontent.com on cloud deployments to keep the Observable integration working.

mattnibs: Oh ok, good point.

PR #4336, opened 2023-01-26
use "zng." prefix for ZNG-related flags

nwt:
Give all ZNG-related flags a common "zng." prefix for clarity and to group them in help text (where flags are alphabetized).


nwt:
The output-check test passes locally with brimdata/zed-sample-data#40.

philrz (approved):

mattnibs (approved):

Issue #4351, opened 2023-01-31
commit-order scan guarantees


mccanne:
It might be desirable to provide ordering guarantees in commit order
when the pool key value is unifrom across a large amount of data.
We previously supported this at small scale but our design breaks at 
large scale and we should contemplate how we might implement this feature.

PR #4352, opened 2023-01-31
add meta-data slicer

mccanne:
This commit changes the algorithm for forming partitions of non-overlapping metadata objects for in-order scans.  This gets us closer to fine-grained metadata operators so we can add vcache integration to the compiler.

A major implication of these changes is that commit-order scanning when pool keys are of uniform value is no longer maintained.  This isn't as bad as it sounds because this feature never worked anyway at larger scale (when the size of the uniform-key data exceeds the batch parallelism of the merge op the order guarantee can fail and we did not have test coverage for this).  See issue #4351.

We also simplified things to use data.Objects without extent.Spans and changed the logic to use the underlying From/To fields of the data.Object. There is more work to do on this front.

We rearranged the metadata operators to take an explicit zctx so there can be type separation between the metadata processing and downstream operators, but no type separation for metaqueries where the downstream user query operates on the metadata.

### DIFF
--- a/compiler/kernel/op.go
+++ b/compiler/kernel/op.go
@@ -593,11 +593,12 @@ func (b *Builder) compileTrunk(trunk *dag.Trunk, parent zbuf.Puller) ([]zbuf.Pul
 			// the metadata here is intercepted by the scanner and these zed values never enter
 			// the flowgraph.  For the metaqueries below, we pass in the flowgraph's type context
 			// because this data does, in fact, flow into the downstream flowgraph.
-			l, err := meta.NewSortedLister(b.pctx.Context, b.pctx.Zctx, lk, pool, src.Commit, filter)
+			zctx := zed.NewContext()
+			l, err := meta.NewSortedLister(b.pctx.Context, zctx, lk, pool, src.Commit, filter)
 			if err != nil {
 				return nil, err
 			}
-			slicer = meta.NewSlicer(l, b.pctx.Zctx, pool.Layout.Order)
+			slicer = meta.NewSlicer(l, zctx, pool.Layout.Order)
 			b.pools[src] = pool
 			b.slicers[src] = slicer
 		}
--- a/compiler/kernel/op.go
+++ b/compiler/kernel/op.go
@@ -43,7 +43,7 @@ var ErrJoinParents = errors.New("join requires two upstream parallel query paths
 type Builder struct {
 	pctx     *op.Context
 	source   *data.Source
-	listers  map[dag.Source]*meta.Lister
+	slicers  map[dag.Source]*meta.Slicer
 	pools    map[dag.Source]*lake.Pool
 	progress *zbuf.Progress
 	deletes  *sync.Map
@@ -54,7 +54,7 @@ func NewBuilder(pctx *op.Context, source *data.Source) *Builder {
 	return &Builder{
 		pctx:    pctx,
 		source:  source,
-		listers: make(map[dag.Source]*meta.Lister),
+		slicers: make(map[dag.Source]*meta.Slicer),
 		pools:   make(map[dag.Source]*lake.Pool),
 		progress: &zbuf.Progress{
 			BytesRead:      0,
@@ -220,6 +220,9 @@ func (b *Builder) compileLeaf(o dag.Op, parent zbuf.Puller) (zbuf.Puller, error)
 			return nil, err
 		}
 		as, err := compileLval(v.As)
+		if err != nil {
+			return nil, err
+		}
 		if len(as) != 1 {
 			return nil, errors.New("explode field must be a top-level field")
 		}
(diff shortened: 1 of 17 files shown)


mccanne:
I'd like to get this merged then simplify things further in a subsequent PR to remove extent.Spans completely from the metadata path.  We can do this by generating the span-filter logic depending on the order and just using the underlying First and Last fields of data.Object.

compiler/kernel/op.go
```diff
@@ -578,29 +581,34 @@ func (b *Builder) compileTrunk(trunk *dag.Trunk, parent zbuf.Puller) ([]zbuf.Pul
...
 			if err != nil {
 				return nil, err
 			}
-			l, err := meta.NewSortedLister(b.pctx.Context, lk, pool, src.Commit, filter)
+			// We pass a new type context in here because we don't want the metadata types to interfere
+			// with the downstream flowgraph.  And there's no need to map between contexts because
+			// the metadata here is intercepted by the scanner and these zed values never enter
+			// the flowgraph.  For the metaqueries below, we pass in the flowgraph's type context
+			// because this data does, in fact, flow into the downstream flowgraph.
+			l, err := meta.NewSortedLister(b.pctx.Context, b.pctx.Zctx, lk, pool, src.Commit, filter)
```
nwt, line 596: Given the comment, should `b.pctx.Zctx` be `zed.NewContext()`?

mccanne: doh!  yes

mccanne: slicer needs to pick it up too

lake/data/seekindex.go
```diff
@@ -4,13 +4,14 @@ import (
...
 	"github.com/brimdata/zed/lake/seekindex"
+	"github.com/brimdata/zed/order"
 	"github.com/brimdata/zed/pkg/storage"
 	"github.com/brimdata/zed/runtime/expr"
 	"github.com/brimdata/zed/runtime/expr/extent"
 )
 
 func LookupSeekRange(ctx context.Context, engine storage.Engine, path *storage.URI,
-	o *Object, cmp expr.CompareFn, filter *expr.SpanFilter, countSpan extent.Span) (seekindex.Range, error) {
+	o *Object, cmp expr.CompareFn, filter *expr.SpanFilter, countSpan extent.Span, dir order.Which) (seekindex.Range, error) {
```
nwt, line 14: Nit: We use `o` or `order` for `order.Which` symbols everywhere else.

runtime/op/meta/lister.go
```diff
@@ -10,16 +12,16 @@ import (
...
 	"github.com/brimdata/zed/zson"
 	"github.com/segmentio/ksuid"
 	"golang.org/x/sync/errgroup"
 )
 
-// XXX Lister enumerates all the partitions in a scan.  After we get this working,
-// we will modularize further by listing just the objects and having
-// the slicer to organize objects into slices (formerly known as partitions).
+// Lister enumerates all the data.Objects in a scan.  A Slicer downstream may
+// optionally organizes objects into non-overlapping partitions for merge on read.
```
nwt, line 22:
```suggestion
// optionally organize objects into non-overlapping partitions for merge on read.
```

lake/data/seekindex.go
```diff
@@ -23,7 +24,13 @@ func LookupSeekRange(ctx context.Context, engine storage.Engine, path *storage.U
 		if s == nil || err != nil {
 			return rg, err
 		}
-		if filter != nil && filter.Eval(s.Keys.First(), s.Keys.Last()) {
+		first := s.Keys.First()
+		last := s.Keys.Last()
+		swapper := expr.NewValueCompareFn(order.Asc, dir == order.Asc)
```
nwt, line 29: Maybe hoist this out of the loop?

runtime/op/meta/lister.go
```diff
@@ -68,51 +72,88 @@ func (l *Lister) Snapshot() commits.View {
...
-	if len(l.parts) == 0 {
+	if l.err != nil || len(l.objects) == 0 {
+		return nil, l.err
+	}
+	// End after the last object.  XXX we could change this so a scan can appear
+	// inside of a subgraph and be restarted after each time done is called
+	// (like how head/tail work).
+	if len(l.objects) == 0 {
 		return nil, l.err
 	}
```
nwt, line 89:
```suggestion
	// End after the last object.  XXX we could change this so a scan can appear
	// inside of a subgraph and be restarted after each time done is called
	// (like how head/tail work).
	if l.err != nil || len(l.objects) == 0 {
		return nil, l.err
	}
```

runtime/op/meta/sequence.go
```diff
@@ -167,42 +167,52 @@ func (s *statScanner) Pull(done bool) (zbuf.Batch, error) {
...
 		if err != nil {
-			return seekindex.Range{}, err
+			return nil, err
+		}
+		r, err := data.LookupSeekRange(ctx, pool.Storage(), pool.DataPath, o, cmp, spanFilter, indexSpan, pool.Layout.Order)
+		if err != nil {
+			return nil, err
 		}
-		return data.LookupSeekRange(ctx, pool.Storage(), pool.DataPath, o, cmp, spanFilter, indexSpan)
+		return &r, nil
```
nwt, line 214: We can do this later, but data.LookupSeekRange should return a pointer so this can just be
```go
		return data.LookupSeekRange(ctx, pool.Storage(), pool.DataPath, o, cmp, spanFilter, indexSpan, pool.Layout.Order)
```

runtime/op/meta/slicer.go
```diff
@@ -1,165 +1,136 @@
...
-func (s *stack) tos() *Range {
-	return &(*s)[len(*s)-1]
+func NewSlicer(parent zbuf.Puller, zctx *zed.Context, o order.Which) *Slicer {
+	m := zson.NewZNGMarshalerWithContext(zctx)
+	m.Decorate(zson.StylePackage)
+	return &Slicer{
+		parent:      parent,
+		marshaler:   m,
+		unmarshaler: zson.NewZNGUnmarshaler(),
+		objects:     []*data.Object{},
```
nwt, line 37: Nit: Tests pass without this.
```suggestion
```

runtime/op/meta/slicer.go
```diff
@@ -1,165 +1,136 @@
...
+		if err != nil {
+			return nil, err
+		}
+		if batch == nil {
+			return s.nextPartition()
+		}
+		vals := batch.Values()
+		if len(vals) != 1 {
+			// We currently support only one object per batch.
+			return nil, errors.New("system error: Searcher encountered multi-valued batch")
```
nwt, line 66:
```suggestion
			return nil, errors.New("system error: Slicer encountered multi-valued batch")
```

runtime/op/meta/slicer.go
```diff
@@ -1,165 +1,136 @@
...
+		if first == nil {
+			first = &o.First
+			last = &o.Last
+		} else {
+			if s.cmp(&o.First, first) < 0 {
+				first = &o.First
+			}
+			if s.cmp(&o.Last, last) > 0 {
+				last = &o.Last
+			}
```
nwt, line 96:
```suggestion
	first := &s.objects[0].First
	last := &s.objects[0].Last
	for _, o := range s.objects[1:] {
		if s.cmp(&o.First, first) < 0 {
			first = &o.First
		}
		if s.cmp(&o.Last, last) > 0 {
			last = &o.Last
		}
```

nwt (approved):
:shipit:

PR #4373, opened 2023-02-09
Add user defined operators

### LINKED ISSUE #4152: User-defined operators
Whereas user-defined functions (#4026) only allow for the execution of expressions, we've also recognized the need for more sophisticated user-defined operators (or "user ops" for short, also formerly called "parameterized subgraphs"). These would allow for callable Zed pipelines whose results are passed back into the data flow from which they were called.

A specific community zync user recently cited the need for this functionality in their use case. Implementation is TBD, but here was the request in their own words.

> I had a long conversation with [our team] about this topic of a 'stored procedure'. The next real time saver for us would be to have an new operator, let's call `execute` which would go like this: `execute <query string>,p1,p2,...pN`. The `<query string>` could look like this: `execute "from $p1 | cut $p2 | sort $p3"`, which would substitute the `p1,p2,p3` value, execute the query and pass the result to the data flow.

mattnibs:
Closes #4152

### DIFF
--- a/compiler/ast/ast.go
+++ b/compiler/ast/ast.go
@@ -211,8 +211,16 @@ type FuncDecl struct {
 	Expr   Expr     `json:"expr"`
 }
 
+type OpDecl struct {
+	Kind   string      `json:"kind" unpack:""`
+	Name   string      `json:"name"`
+	Params []string    `json:"params"`
+	Scope  *Sequential `json:"scope"`
+}
+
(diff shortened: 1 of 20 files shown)


mattnibs:
TODO:
- [ ] Add more robust error test cases (e.g., op cycles)
- [ ] Add test for call user defined ops within user defined ops
- [ ] Add support for calling user-defined ops with no arguments.

PR #4375, opened 2023-02-13
remove scan range and scan order from DAG

mccanne:
This commit removes the scan range options and scan order options from the DAG (and AST).  Now that we have decent inference of pool key predicates this options are no longer needed.  Furthermore, the scan order was being ignored.  If we want the functionality later, this will be better implemented by the user specifying a sort and the optimizer detecting that a reverse scan can be executed instead of an actual sort.


mccanne:
This won't break zui, right?

nwt (approved):

PR #4377, opened 2023-02-14
improve UX for metadata queries

mccanne:
This commit improves the UX for metadata queries by using the metadata type name to distinguish between commit-level queries and pool-level queries rather than requiring the presence of an @branch commit tag for commit-level queries.  This way you can query the objects on the main branch of a pool without specifying @main.  We also fixed a bug in the optimization logic where the layout of metadata queries was presumed to be the same layout of the pool.  The fix is to presume the layout of metadata queries is unknown.

### DIFF
--- a/compiler/ast/dag/operator.go
+++ b/compiler/ast/dag/operator.go
@@ -200,6 +200,25 @@ type (
 	}
 )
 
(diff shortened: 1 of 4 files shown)


nwt (approved):
Nice.

PR #4380, opened 2023-02-14
compiler: optimize by merging adjacent filters

nwt:
Optimize queries by merging adjacent filter operators so that, e.g., "where a | where b" becomes "where a and b".

Note that this implementation does not merge filters in "over" expressions like "sum(over a | where b | where c)".  Doing so requires a lot more code and doesn't seem worth the effort right now.

### DIFF
--- a/compiler/optimizer/optimizer.go
+++ b/compiler/optimizer/optimizer.go
@@ -37,39 +37,46 @@ func (o *Optimizer) Entry() *dag.Sequential {
 //
 // Note: MergeFilters does not descend into dag.OverExpr.Scope, so it cannot
 // merge filters in "over" expressions like "sum(over a | where b | where c)".
-func (o *Optimizer) MergeFilters() { mergeFilters(o.entry) }
-
-func mergeFilters(op dag.Op) {
-	switch op := op.(type) {
-	case *dag.From:
-		for _, t := range op.Trunks {
-			mergeFilters(t.Seq)
-		}
-	case *dag.Over:
-		mergeFilters(op.Scope)
-	case *dag.Parallel:
-		for _, o := range op.Ops {
-			mergeFilters(o)
-		}
-	case *dag.Sequential:
-		if op == nil {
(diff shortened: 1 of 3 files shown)


compiler/ztests/merge-filters.yaml
```diff
@@ -0,0 +1,39 @@
...
+        file a =>
+          where b and c
+        file d =>
+          where e and f
+      )
+      ===
+      from (
+        (internal reader) =>
+          over a
+      )
```
nwt, line 29: This isn't correct because zfmt doesn't print dag.Over.Scope. I'll fix that in another PR.

compiler/optimizer/optimizer.go
```diff
@@ -31,6 +32,46 @@ func (o *Optimizer) Entry() *dag.Sequential {
...
+		for _, t := range op.Trunks {
+			mergeFilters(t.Seq)
+		}
+	case *dag.Over:
+		mergeFilters(op.Scope)
+	case *dag.Parallel:
+		for _, o := range op.Ops {
+			mergeFilters(o)
+		}
+	case *dag.Sequential:
```
mattnibs, line 54: I am wondering if we want to add something like dag.Walk and use this here. We have now a few areas of optimizer that does this and as we add additional operators I'm concerned we'll miss the different places in optimizer that crawls over dags looking for leafs. For instance my user defined operators pr will need to do these kind of updates...

mattnibs (approved):

mccanne:
Nice!

compiler/optimizer/optimizer.go
nwt: This was a really good suggestion!

I didn't try to add a fully-fledged dag.Walk because I wasn't confident I'd get the interface right, but I think reimplementing what was here in terms of a bare-bones walkOp function made it a lot clearer, and we can expand walkOp into dag.Walk down the road.

PR #4386, opened 2023-02-17
unify data.Object and seekindex frames and improve seekindex

mccanne:
This commit unifies the data.Object and seekindex metadata so we use the consistent pattern of From/To to represent a range of Zed values.  This will simplify the pruning logic in a subsequent PR where the pruning logic for data.Objects and seekindex frames can be the same.

We now use the From/To pattern everywhere and no longer have First/Last or Lower/Upper.

This commit also improves the seekindex data structure.  Instead of having seekindex.Reader build the data structure that is desired from the old seekindex format, we simply put the desired data structure directly in the seekindex zng gile and use a standard zio.Reader to unmarshal the index entries without additional processing.

We also got rid of data.Object.Meta and repositioned these fields as top-level data.Object fields simplifying the layout here.

Because this changes the lake format, we bumped the lake version from 1 to 2.


lake/data/seekindex.go
```diff
@@ -17,28 +21,37 @@ func LookupSeekRange(ctx context.Context, engine storage.Engine, path *storage.U
...
+		if swapper(from, to) > 0 {
+			from, to = to, from
 		}
-		if countSpan != nil && !countSpan.Overlaps(s.Counts.First(), s.Counts.Last()) {
+		if filter != nil && filter.Eval(from, to) {
 			continue
 		}
-		if rg.Start == -1 {
-			rg.Start = s.Range.Start
+		if countSpan != nil {
```
mattnibs, line 45: If we're dropping indexes we no longer need this count stuff, but this can be handled later...

lake/data/writer.go
```diff
@@ -63,38 +62,49 @@ func (o *Object) NewWriter(ctx context.Context, engine storage.Engine, path *sto
...
+	return w.flushSeekIndex()
+}
+
+func (w *Writer) flushSeekIndex() error {
+	if w.from != nil {
+		w.seekIndexTrigger = 0
+		from := w.from
+		to := w.object.To.Copy()
+		w.from = nil
+		return w.seekIndex.Write(from, to, w.count, uint64(w.writer.Position()))
```
mattnibs, line 105: On line 77 you remove the count increment and yet this is passed to the `seekindex.Writer`. As I said above I think we can remove the counts since we're nixing the indexer.

lake/data/seekindex.go
mccanne: Yes, this in my next PR.  I just wanted all the tests to continue to pass in this commit.

lake/data/writer.go
mccanne: We'll want the counts for selection vectors from the vcache.

mattnibs (approved):

PR #4393, opened 2023-02-19
replace keyfilter with pruner and unhook search index logic

mccanne:
This commit replaces the keyfilter pattern with a pruner pattern that is more reminiscent of OLAP and data lake architectures. The logic for detecting whether we can prune based on pool keys has moved from the DAG package to the compiler optimizer and now sits adjacent to the logic that decides whether filter predicates can be pushed into the scanning layer.

We also added a "tap" option to the meta-queries to ease test and debug.  When the tap keyword is appended to a pool:objects query, the downstream flowgraph is compiled as if it were a normal query but the query runs as a metaquery, generating the objects after they have been pruned in accordance with any pushdowns present in the downstream flowgraph.  The flowgraph is not applied to the stream of objects of course.

We also improved the seekindex implementation where we scan only the seekindex frames that survive the pruner rather than the smallest single span that covers all such frames as was previously implemented.


compiler/kernel/op.go
```diff
@@ -588,28 +588,42 @@ func (b *Builder) compileTrunk(trunk *dag.Trunk, parent zbuf.Puller) ([]zbuf.Pul
 			if err != nil {
 				return nil, err
 			}
+			var pruner expr.Evaluator
+			if trunk.KeyPruner != nil {
+				pruner, err = compileExpr(trunk.KeyPruner)
+				if err != nil {
+					return nil, err
+				}
+			}
```
nwt, line 597: Nit: This also appears immediately after the enclosing if statement, so you can delete both copies in favor of one placed before the if statement.

compiler/kernel/op.go
```diff
@@ -618,7 +632,16 @@ func (b *Builder) compileTrunk(trunk *dag.Trunk, parent zbuf.Puller) ([]zbuf.Pul
...
-		scanner, err := meta.NewCommitMetaScanner(b.pctx.Context, b.pctx.Zctx, b.source.Lake(), src.Pool, src.Commit, src.Meta, pushdown)
+		var pruner expr.Evaluator
+		if src.Tap {
+			if trunk.KeyPruner != nil {
+				pruner, err = compileExpr(trunk.KeyPruner)
+				if err != nil {
+					return nil, err
+				}
+			}
+		}
```
nwt, line 643:
```suggestion
		if src.Tap && trunk.KeyPruner != nil {
			pruner, err = compileExpr(trunk.KeyPruner)
			if err != nil {
				return nil, err
			}
		}
```

compiler/optimizer/optimizer.go
```diff
@@ -110,6 +110,16 @@ func (o *Optimizer) OptimizeScan() error {
 	}
 	for k := range from.Trunks {
 		pushDown(&from.Trunks[k])
+		// Check to see if we can add a range pruner when the pool-key is used
+		// in a normal filtering operation.
+		trunk := &from.Trunks[k]
```
nwt, line 115:
```suggestion
		trunk := &from.Trunks[k]
		pushDown(trunk)
		// Check to see if we can add a range pruner when the pool key is used
		// in a normal filtering operation.
```

compiler/optimizer/optimizer.go
```diff
@@ -110,6 +110,16 @@ func (o *Optimizer) OptimizeScan() error {
...
+		if layout, ok := o.layouts[trunk.Source]; ok {
+			if pushdown, ok := trunk.Pushdown.(*dag.Filter); ok {
+				if p := newRangePruner(pushdown.Expr, layout.Primary(), layout.Order); p != nil {
+					trunk.KeyPruner = p
+				}
```
nwt, line 120: I might be missing something, but it trunk.KeyPruner looks to me as though it's always nil here, so you could simplify a little.
```suggestion
				trunk.KeyPruner = newRangePruner(pushdown.Expr, layout.Primary(), layout.Order)
```

compiler/optimizer/optimizer.go
```diff
@@ -224,7 +234,15 @@ func (o *Optimizer) getLayout(s dag.Source, parent order.Layout) (order.Layout,
 		return s.Layout, nil
 	case *dag.Pool:
 		return o.source.Layout(o.ctx, s), nil
-	case *dag.LakeMeta, *dag.PoolMeta, *dag.CommitMeta:
+	case *dag.CommitMeta:
+		if s.Tap && s.Meta == "objects" {
+			// For a tap into the object stream, we compile the downstream
+			// DAG as if it were a normal query (so the optimizer can prune
+			// objects etc) but we execute it in the end as a meta-query.
```
nwt, line 241:
```suggestion
			// objects etc.) but we execute it in the end as a meta-query.
```

compiler/optimizer/optimizer.go
```diff
@@ -322,3 +340,144 @@ func pushDown(trunk *dag.Trunk) {
 	}
 	trunk.Pushdown = filter
 }
+
+// newRangePruner returns a new predicate based on the input predicate pred
+// that when applied to an input value (i.e., "this") with fields from/to, returns
+// true if comparisons in pred against literal values can for certain rule out
+// that pred would be true for any value in the from/to range.  From/to are presumed
+// to be ordered according to the order o.  This is used to prune metadata objects
+// from a scan when we know the pool-key range of the object could not satisfy
```
nwt, line 349: Nit: We don't usually hyphenate this.
```suggestion
// from a scan when we know the pool key range of the object could not satisfy
```

compiler/optimizer/optimizer.go
```diff
@@ -322,3 +340,144 @@ func pushDown(trunk *dag.Trunk) {
...
+	upper := &dag.This{Kind: "This", Path: field.New("to")}
+	if o == order.Desc {
+		lower, upper = upper, lower
+	}
+	e, _ := buildRangePruner(pred, fld, lower, upper, o)
+	return e
+}
+
+// buildRangePruner creates a DAG comparison expression that can evalaute whether
+// a Zed value adhering to the From/To pattern can be excluded from a scan because
```
nwt, line 362: Nit: This is styled as "from/to" in the newRangePruner doc comment.
```suggestion
// a Zed value adhering to the from/to pattern can be excluded from a scan because
```

compiler/optimizer/optimizer.go
```diff
@@ -322,3 +340,144 @@ func pushDown(trunk *dag.Trunk) {
...
+// the expression "pred" would evalate to false for all values of "fld" in the
```
nwt, line 363:
```suggestion
// the expression pred would evaluate to false for all values of fld in the
```

runtime/op/meta/ztests/pruner.yaml
```diff
@@ -0,0 +1,44 @@
...
+      {from:8,to:12,count:5(uint64),size:30}
+      {from:20,to:25,count:6(uint64),size:34}
+      ===
+      {from:8,to:12,count:5(uint64),size:30}
+      {from:10,to:15,count:6(uint64),size:34}
+      {from:14,to:16,count:3(uint64),size:22}
+      {from:20,to:25,count:6(uint64),size:34}
+      ===
+      {from:20,to:25,count:6(uint64),size:34}
+
```
nwt, line 44:
```suggestion
```

runtime/op/meta/ztests/pruner.yaml
```diff
@@ -0,0 +1,44 @@
...
+  echo ===
+  zed query "from tmp:objects tap | k <= 10" | zq -z "drop id" -
+  echo ===
+  zed query "from tmp:objects tap | k >= 15 and k < 20" | zq -z "drop id" -
+  echo ===
+  zed query  "from tmp:objects tap | k <= 9 or k > 24" | zq -z "drop id" -
+  echo ===
+  zed query  'from tmp:objects tap | a[k] == "foo" or k >= 20' | zq -z "drop id" -
+  echo ===
+  zed query  'from tmp:objects tap | a[k] == "foo" and k >= 20' | zq -z "drop id" -
```
nwt, line 22: Nit: No need for zq here.
```suggestion
  zed query "from tmp:objects tap | k > 18 | drop id"
  echo ===
  zed query "from tmp:objects tap | k <= 10 | drop id"
  echo ===
  zed query "from tmp:objects tap | k >= 15 and k < 20 | drop id"
  echo ===
  zed query "from tmp:objects tap | k <= 9 or k > 24 | drop id"
  echo ===
  zed query 'from tmp:objects tap | a[k] == "foo" or k >= 20 | drop id'
  echo ===
  zed query 'from tmp:objects tap | a[k] == "foo" and k >= 20 | drop id'
```

compiler/ast/dag/operator.go
```diff
@@ -157,10 +157,11 @@ type (
...
 	Trunk struct {
-		Kind     string      `json:"kind" unpack:""`
-		Source   Source      `json:"source"`
-		Seq      *Sequential `json:"seq"`
-		Pushdown Op          `json:"pushdown"`
+		Kind      string      `json:"kind" unpack:""`
+		Source    Source      `json:"source"`
+		Seq       *Sequential `json:"seq"`
+		Pushdown  Op          `json:"pushdown"`
+		KeyPruner Expr        `json:"key_pruner"`
```
nwt, line 164: Unless you think we'll introduce another type of pruner, I'm for shortening this name to simply Pruner, but I don't feel strongly about that!

lake/data/seekindex.go
```diff
@@ -9,48 +9,43 @@ import (
...
+	}
 	r, err := engine.Get(ctx, obj.SeekIndexURI(path))
 	if err != nil {
 		return nil, err
 	}
 	defer r.Close()
+	var ranges []seekindex.Range
 	var rg *seekindex.Range
 	unmarshaler := zson.NewZNGUnmarshaler()
 	reader := zngio.NewReader(zed.NewContext(), r)
```
nwt, line 30: We should close this ZNG reader.
```suggestion
	reader := zngio.NewReader(zed.NewContext(), r)
	defer reader.Close()
```

runtime/op/meta/pruner.go
```diff
@@ -0,0 +1,27 @@
...
+		pred: e,
+		ectx: expr.NewContext(),
+	}
+}
+
+func (p *Pruner) prune(val *zed.Value) bool {
+	if p == nil {
+		return false
+	}
+	result := p.pred.Eval(expr.NewContext(), val)
```
nwt, line 25:
```suggestion
	result := p.pred.Eval(p.ectx, val)
```

runtime/op/meta/pruner.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/brimdata/zed/order"
+	"github.com/brimdata/zed/runtime/expr"
+)
+
+type Pruner struct {
+	pred expr.Evaluator
+	ectx expr.Context
+}
+
+func NewPruner(e expr.Evaluator, o order.Which) *Pruner {
```
nwt, line 14: Nits: No need to export this. And o is unused.
```suggestion
func newPruner(e expr.Evaluator) *Pruner {
```

runtime/op/meta/pruner.go
```diff
@@ -0,0 +1,27 @@
+package meta
+
+import (
+	"github.com/brimdata/zed"
+	"github.com/brimdata/zed/order"
+	"github.com/brimdata/zed/runtime/expr"
+)
+
+type Pruner struct {
```
nwt, line 9: Nit: No need to export this.
```suggestion
type pruner struct {
```

runtime/op/meta/lister.go
```diff
@@ -73,62 +76,31 @@ func (l *Lister) Pull(done bool) (zbuf.Batch, error) {
...
-		}
+		l.objects = initObjectScan(l.snap, l.pool.Layout)
 	}
-	// End after the last object.  XXX we could change this so a scan can appear
-	// inside of a subgraph and be restarted after each time done is called
-	// (like how head/tail work).
-	if l.err != nil || len(l.objects) == 0 {
+	if l.err != nil {
 		return nil, l.err
 	}
```
nwt, line 83: Nit: Move this ahead of the l.objects if statement.

compiler/optimizer/optimizer.go
```diff
@@ -322,3 +340,144 @@ func pushDown(trunk *dag.Trunk) {
...
+}
+
+func rangePrunerPred(op string, literal *dag.Literal, lower, upper *dag.This, o order.Which) *dag.BinaryExpr {
+	switch op {
+	case "<":
+		// key < CONST
+		return compare("<=", literal, lower, o)
+	case "<=":
+		// key <= CONST
+		return compare("<", literal, lower, o)
```
nwt, line 426: Why does < become <= and vice versa in the call to compare?

compiler/optimizer/optimizer.go
```diff
@@ -322,3 +340,144 @@ func pushDown(trunk *dag.Trunk) {
...
+		lower, upper = upper, lower
+	}
+	e, _ := buildRangePruner(pred, fld, lower, upper, o)
+	return e
+}
+
+// buildRangePruner creates a DAG comparison expression that can evalaute whether
+// a Zed value adhering to the From/To pattern can be excluded from a scan because
+// the expression "pred" would evalate to false for all values of "fld" in the
+// From/To value range.  If a pruning decision can be reliably determined then
```
nwt, line 364:
```suggestion
// from/to value range.  If a pruning decision can be reliably determined then
```

compiler/optimizer/optimizer.go
```diff
@@ -322,3 +340,144 @@ func pushDown(trunk *dag.Trunk) {
...
+// the boolean return value is true; otherwise, false.
+func buildRangePruner(pred dag.Expr, fld field.Path, lower, upper *dag.This, o order.Which) (*dag.BinaryExpr, bool) {
```
nwt, line 366: I don't see any way for this this function to return (nil, true), so I think you can simplify it by dropping the bool return parameter.

mattnibs (approved):
lgtm

compiler/kernel/op.go
mccanne: That doesn't work because the lister and the scanner need their own copies as Evaluators can't be used concurrently.

compiler/optimizer/optimizer.go
mccanne: This avoids the dreaded non-nil interface that points to a concrete nil pointer.  I could change rangePruner to return the interface instead of the struct pointer.  Which pattern do you prefer?

compiler/optimizer/optimizer.go
mccanne: I thought compound nouns used as an adjective should be hyphenated but I will change it.

runtime/op/meta/ztests/pruner.yaml
mccanne: While this happens to work, the idea of the "tap" option is to analyze the flowgraph provided then emit the objects in accordance with the downstream flowgraph so tacking on additional processing that would interfere with the analysis being tested seems like a bad idea.  In this case it's ok because the drop op doesn't change the analysis but it might down the road so I'm inclined to leave the zq commands here to emphasize this point.

compiler/optimizer/optimizer.go
nwt: Ah. Stick with what you've got.

runtime/op/meta/ztests/pruner.yaml
nwt: 👍

compiler/kernel/op.go
nwt: 👍

nwt (approved):
:shipit:!

PR #4399, opened 2023-02-22
lake: log cache object access error and continue

### LINKED ISSUE #4328: fall back to original data on error reading commit or journal cache object
An error in lake/commits.Store.getSnapshot or lake/journal.Store.getSnapshot is unrecoverable even though those methods read objects containing only cached information. (We've seen a lake rendered inaccessible by a corrupt journal snap.zng.) We should change those methods to log the error and then fall back to reading the original data.

nwt:
Errors from lake/commits.Store.getSnapshot and putSnapshot and lake/journal.Store.getSnapshot and putSnapshot are unrecoverable even though those methods operate on objects containing only cached information.  Change commits.Store.Snapshot and journal.Store.load to log these errors and continue instead of failing unrecoverably. (Changes elsewhere are just logger plumbing.)

Closes #4328.

### DIFF
--- a/cli/lakeflags/flags.go
+++ b/cli/lakeflags/flags.go
@@ -14,6 +14,7 @@ import (
 	"github.com/brimdata/zed/lake/api"
 	"github.com/brimdata/zed/lakeparse"
 	"github.com/brimdata/zed/pkg/storage"
+	"go.uber.org/zap"
 )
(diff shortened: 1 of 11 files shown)


mattnibs (approved):
This is lovely!

PR #4411, opened 2023-02-28
Change "pass" references to "yield this"

philrz:
I recently went to add an example to the `join` tutorial that required the functionality provided by `pass`, which led me to confront that `pass` was not currently documented. That led to the observation that `yield this` is functionally equivalent and not much more verbose, so we could make the language simpler by dropping `pass`. As a prerequisite to putting up the `join` docs PR, here I'm just switching over the references in the docs.

### DIFF
--- a/docs/language/operators/combine.md
+++ b/docs/language/operators/combine.md
@@ -21,7 +21,7 @@ and its semantics of undefined merge order.
 
 _Copy input to two paths and combine with the implied operator_
 ```mdtest-command
-echo '1 2' | zq -z 'fork (=>pass =>pass) | sort this' -
+echo '1 2' | zq -z 'fork (=>yield this =>yield this) | sort this' -
 ```
 =>
 ```mdtest-output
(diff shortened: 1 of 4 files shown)


philrz:
While the changes to the examples here are syntactically correct, @nwt spotted that the `join` example from #4403 that originally needed `pass` won't work with `yield this`. If I change it:

```
$ cat three-way-join.zed 
from (
  file fruit.ndjson => sort flavor
  file people.ndjson => sort likes
) | inner join on flavor=likes eater:=name | sort name
| from (
  yield this
  file prices.ndjson => sort name
) | inner join on name=name price:=price

$ zq -I three-way-join.zed 
zq: error parsing Zed in three-way-join.zed at line 6, column 9:
  yield this
    === ^ ===
```

Therefore @nwt and I agreed that we can move forward with keeping `pass` for now and perhaps look at making the parser changes down the road to standardize on `yield this`. Closing this PR.

PR #4415, opened 2023-03-01
convert from/to pattern to min/max for object metadata and seekindex

mccanne:
This commit changes the from/to pattern to an absolute min/max approach for the range of values in a data.Object and seekindex.Entry.  Overall this simplifies the code a bit with a small increae in complexity when sorting object snapshots.  This will also make things consistent with generalized pruning that we will implement in the vcache when we have vector stats (including min/max).


lake/data/writer.go
```diff
@@ -26,7 +26,7 @@ type Writer struct {
 	seekIndexStride  int
 	seekIndexTrigger int
 	first            bool
-	from             *zed.Value
+	seek_min         *zed.Value
```
nwt, line 29: Nit: Ditch the underscore.
```suggestion
	seekMin         *zed.Value
```

zbuf/merger.go
```diff
@@ -17,6 +17,17 @@ func NewComparator(zctx *zed.Context, layout order.Layout) *expr.Comparator {
 	return expr.NewComparator(nullsMax, !nullsMax, exprs...).WithMissingAsNull()
 }
 
+func NewComparatorNullsMax(zctx *zed.Context, layout order.Layout) *expr.Comparator {
```
nwt, line 20: Am I missing something or is the behavior of this identical to that of NewComparator?

mccanne: Oh yeah meant to fix this

mccanne: Actually, I am futzing with this.  The merge op unit test doesn't seem to work with the NullsMax variation and there's no way to use NewComparator in a way where nullsmax is always true and reverse is true only for order.Desc.  I propose we merge this is and clean it up later.

nwt: Ah, I totally missed that the first argument to expr.NewComparator here is always true.

nwt (approved):

runtime/op/meta/slicer.go
```diff
@@ -108,18 +110,29 @@ func (s *Slicer) stash(o *data.Object) (zbuf.Batch, error) {
...
 			var err error
 			batch, err = s.nextPartition()
 			if err != nil {
 				return nil, err
 			}
-			s.to = nil
+			s.min = nil
+			s.max = nil
+		} else {
+
```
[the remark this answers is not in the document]
mccanne: oops

PR #4417, opened 2023-03-01
Add User Ops

### LINKED ISSUE #4152: User-defined operators

mattnibs:
Add support for user defined operators or "user ops". User ops can call other ops within the same scope but recursive op calls are not allowed.

Closes #4152


compiler/parser/parser.peg
```diff
@@ -82,6 +82,15 @@ FuncDecl
         "expr":expr))
     }
 
+OpDecl
+  = "op" _ name:IdentifierName __ "(" __ params:IdentifierNames __ ")" scope:Scope {
```
nwt, line 86: I'm for making this consistent with FuncDecl by using : instead of => between the parameters and body.
```suggestion
  = "op" _ name:IdentifierName __ "(" __ params:IdentifierNames __ ")" __ ":" __ "(" __ scope:Sequential __ ")" {
```

compiler/semantic/scope.go
```diff
@@ -45,14 +59,13 @@ func (s *Scope) DefineVar(name string) error {
 	return nil
 }
 
-func (s *Scope) DefineAs(name string) error {
+func (s *Scope) DefineAs(name string, e dag.Expr) error {
 	b := s.tos()
 	if _, ok := b.symbols[name]; ok {
+
```
nwt, line 65:
```suggestion
```

runtime/op/user/user.go
```diff
@@ -0,0 +1,41 @@
...
+	batch, err := p.parent.Pull(false)
+	if batch == nil || err != nil {
+		return nil, err
+	}
+	vals := make([]zed.Value, len(batch.Values()))
+	for i, v := range batch.Values() {
+		vals[i] = *p.expr.Eval(batch, &v)
+		if vals[i].IsError() {
+			fmt.Println("isanerror")
+		}
```
nwt, line 38:
```suggestion
```

zfmt/ztests/decls.yaml
```diff
@@ -36,4 +40,6 @@ outputs:
       from (
         (internal reader)
       )
-      | pass
+      | stamp({outer:this,assignee:"bob"}) => (
+        yield {...outer,assignee:assignee,ts:now()}
+      )
```
nwt, line 45: Unless I'm mistaken, `zc -C` output has always been a valid Zed program. Is there a compelling reason to break with that here?

mattnibs: You are mistaken as was I. See how we display pushdowns in filters or specify if the output in a summarize proc is partials, I'm sure there are other examples. I did of course want to make it a valid zed program but there's some difficulties in doing this, most notably in order to scoping right I'd have to figure out how to compile the op declarations in place vs compiling where they are called. While I think this is eventually where we want to go this would to require an additional pass over the dag and this isn't a trivial change. So I made the decision to ship this and worry about making zfmt -s on op decls valid zed later because I don't think this should get in the way to getting this out.

compiler/parser/parser.peg
mattnibs: Not opposed to this.

runtime/op/user/user.go
mattnibs: Whoops!

compiler/semantic/scope.go
```diff
@@ -65,6 +77,15 @@ func (s *Scope) DefineFunc(f *dag.Func) error {
...
 }
 
+func (s *Scope) DefineOp(o *ast.OpDecl) error {
+	b := s.tos()
+	if _, ok := b.symbols[o.Name]; ok {
+		return fmt.Errorf("symbol %q redefined", o.Name)
+	}
+	b.Define(o.Name, o)
+	return nil
+}
```
nwt, line 87: This is so similar to DefineAs that it doesn't seem worth its weight. Maybe just change the type of the second DefineAs parameter to any and use that?

compiler/semantic/scope.go
```diff
@@ -89,16 +110,76 @@ func (s *Scope) DefineConst(name string, def dag.Expr) error {
...
+
+func (s *Scope) opCycle(name string) error {
+	for _, n := range s.opstack {
+		if name == n {
+			stack := append(slices.Clone(s.opstack), name)
+			return errOpCycle(stack)
+
 		}
 	}
```
nwt, line 173:
```suggestion
	if slices.Index(s.opstack, name) > -1 {
		stack := append(slices.Clone(s.opstack), name)
		return errOpCycle(stack)
	}
```

compiler/semantic/op.go
```diff
@@ -824,3 +831,68 @@ func semCallOp(scope *Scope, call *ast.Call) (dag.Op, error) {
...
+func maybeConvertUserOp(ctx context.Context, scope *Scope, call *ast.Call, ds *data.Source, head *lakeparse.Commitish) (dag.Op, error) {
+	op, err := scope.LookupOp(call.Name)
+	if op == nil || err != nil {
+		if _, ok := err.(errOpCycle); !ok {
+			err = nil
+		}
+		return nil, err
+	}
+	if call.Where != nil {
+		return nil, fmt.Errorf("%s(): user defined ops cannot have a where clause", call.Name)
```
nwt, line 846: Nit: s/op/operator/ in all error strings for consistency with the documentation, which uses operator.
```suggestion
		return nil, fmt.Errorf("%s(): user defined operators cannot have a where clause", call.Name)
```

compiler/semantic/scope.go
mattnibs: Yeah that's a fair point.

philrz (commented):
In addition to the suggestions I've made, two other things:

1. I know this one was a bit complex to implement, and the docs as they stand seem to reflect some of the underlying complexity of the implementation. It's not clear to me if the user needs to have user-defined ops explained to them with terms like "context" and "scope". I don't feel proficient enough with language concepts to know quite what to suggest instead, though. Maybe it's reasonable that a user advanced enough to be wanting to create user-defined ops is going to be comfortable with the concepts. So, maybe no action required there, but just wanted to make the observation.

2. Adding a new numbered section early in the doc does break a bunch of links as you've already fixed within the PR. However, there's cross-links from other docs that are going to break as well. We went through a similar exercise when user-defined functions went in, so you can see at https://github.com/brimdata/zed/pull/4264/commits/db57e46535d217629f86652d66a4d19fd979961a which other files I had to hit. If you want me to take care of that again and push a commit directly to the PR just let me know and I can put that on my to-do list.

mattnibs:
> In addition to the suggestions I've made, two other things:
> 
> 1. I know this one was a bit complex to implement, and the docs as they stand seem to reflect some of the underlying complexity of the implementation. It's not clear to me if the user needs to have user-defined ops explained to them with terms like "context" and "scope". I don't feel proficient enough with language concepts to know quite what to suggest instead, though. Maybe it's reasonable that a user advanced enough to be wanting to create user-defined ops is going to be comfortable with the concepts. So, maybe no action required there, but just wanted to make the observation.

Yeah this is fair, a lot of this is because I was in the weeds while writing this. I think the term "scope" will be necessary and I wanted to refer to the "lateral scope" section of the docs but this doesn't quite capture the evolving nature of scopes. When I put up my scope pr I plan on moving scopes to its own section towards the top of the language overview. A scope is basically:
```
(
  decl
  sequence
)
```


> 2. Adding a new numbered section early in the doc does break a bunch of links as you've already fixed within the PR. However, there's cross-links from other docs that are going to break as well. We went through a similar exercise when user-defined functions went in, so you can see at [db57e46](https://github.com/brimdata/zed/commit/db57e46535d217629f86652d66a4d19fd979961a) which other files I had to hit. If you want me to take care of that again and push a commit directly to the PR just let me know and I can put that on my to-do list.

I think we should remove the numbered sections in the docs. They are nice to have but they are pain to manage especially with the language evolving as it is.

docs/language/overview.md
```diff
@@ -457,13 +457,103 @@ scope but can override identifiers defined in ancestor scopes.
...
+
+A user-defined operator can then be called with using the familiar call syntax
+```
+<id> ( [<expr> [, <expr> ...]] )
+```
+where `<id>` is the identifier of the user-defined operator and `<expr>` is a list
+of [expressions](#8-expressions) matching the number of `<param>`s defined in
+the operator's signature.
+
+### 5.1 Scope Context
```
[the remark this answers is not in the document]
mattnibs: Ugh, I hate the titles of these sections. I believe these concepts are important for the user to understand but am unsure how to explain them in simple, user facing language. "Inner" and "Outer" scope isn't really right since really its the same scope. The word "Context" is confusing but saying "the `this` value" everywhere also seems weird. I will think about this but am also open to suggestion.

docs/language/overview.md
```diff
@@ -457,13 +457,103 @@ scope but can override identifiers defined in ancestor scopes.
...
+echo '{}' | zq -z -I myop.zed -
+```
+would produce
+```mdtest-output
+{a:"foo",b:"bar"}
+```
+
+### 5.2 Outer Scope Context
+
+If the an `<expr>` in the call to an operator is a variable then the variable's
```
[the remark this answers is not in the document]
mattnibs: I agree "variable" isn't right but saying "the parameters path for the op's `this` value" also seems weird. I'm open to suggestions here.

philrz:
Thanks @mattnibs. Responses on two topics:

1. Yes, on further reflection and after hearing your clarifications, I agree that "scope" surely needs to stay, and I like your idea about having a separation section just talking about scope as an independent concept. Other language docs have it. If we can find a way to avoid having to say "context" I still think that would  be an improvement.

2. I'm on the fence about the numbered sections. On one side there's things like RFCs that have always had numbered sections. I don't recall if I had a hand in introducing numbered sections to our docs back in the day, but if I had to guess, maybe it started because there were data format specs and such that had the feel of RFC-like docs that would hopefully get locked in place and similarly referenced almost like legal texts. Then on the other hand there's plenty of docs out there laid out somewhat like ours (Kubernetes, Grafana, Docusaurus, ...) that don't have numbered sections. I figure we can talk about it as a team and if nobody's super attached to them, I'd be happy to make the changes.

As for the deeper decisions on wording, I'm game to let @nwt have a look with fresh eyes, then see where that lands us.

compiler/ast/ast.go
```diff
@@ -525,3 +537,19 @@ type Agg struct {
...
 }
+
+// ----------------------------------------------------------------------------
+// Params
+
+type (
+	SpreadParam struct {
+		Kind string `json:"kind" unpack:""`
+	}
+	VarParam struct {
```
nwt, line 548: Maybe call this NamedParam?

compiler/ast/dag/op.go
```diff
@@ -101,10 +101,11 @@ type (
...
 	}
 	Scope struct {
-		Kind   string  `json:"kind" unpack:""`
-		Consts []Def   `json:"consts"`
-		Funcs  []*Func `json:"funcs"`
-		Body   Seq     `json:"seq"`
+		Kind   string    `json:"kind" unpack:""`
+		Consts []Def     `json:"consts"`
+		Funcs  []*Func   `json:"funcs"`
+		UserOp []*UserOp `json:"user_ops"`
```
nwt, line 107: Nit: The other slice field names are plural.
```suggestion
		UserOps []*UserOp `json:"user_ops"`
```

compiler/ast/dag/op.go
```diff
@@ -139,6 +140,19 @@ type (
...
+		Params []Param `json:"params"`
+		Body   Seq     `json:"body"`
+	}
+	UserOpCall struct {
+		Kind   string  `json:"kind" unpack:""`
+		Name   string  `json:"name"`
+		Params []Param `json:"params"`
+		Exprs  []Expr  `json:"exprs"`
+		Body   Seq     `json:"body"`
+	}
```
nwt, line 155: Nit: Move these after Uniq.

compiler/kernel/op.go
```diff
@@ -329,6 +330,39 @@ func (b *Builder) compileLeaf(o dag.Op, parent zbuf.Puller) (zbuf.Puller, error)
...
+		}
+		var exit zbuf.Puller
+		if len(exits) == 1 {
+			exit = exits[0]
+		} else {
+			// This can happen when output of the body
+			// is a fork or switch.
+			exit = combine.New(b.octx, exits)
+		}
+		return exit, nil
```
nwt, line 365: Nit: Simplify.
```suggestion
		if len(exits) > 1 {
			// This can happen when output of the body
			// is a fork or switch.
			return combine.New(b.octx, exits), nil
		}
		return exits[0], nil
```

compiler/parser/parser.peg
```diff
@@ -85,6 +85,24 @@ FuncDecl
         "expr": expr))
     }
 
+OpDecl
+  = "op" _ name:IdentifierName __ "(" __ params:Params __ ")" __ ":" __ "(" __ body:Seq __ ")" {
+      RETURN(MAP(
+        "kind":"OpDecl",
+        "name":name,
+        "params":params,
+        "body":body))
```
nwt, line 94: Nit: We have spaces everywhere else.
```suggestion
        "kind": "OpDecl",
        "name": name,
        "params": params,
        "body": body))
```

compiler/ast/ast.go
mattnibs: Good idea

zfmt/dag.go
```diff
@@ -468,6 +468,16 @@ func (c *canonDAG) op(p dag.Op) {
...
 	case *dag.Over:
 		c.over(p)
+	case *dag.UserOpCall:
+		c.next()
+		c.write("%s(", p.Name)
+		for i := range p.Exprs {
+			if i > 0 {
+				c.write(", ")
+			}
+			c.expr(p.Exprs[i], "")
```
nwt, line 478:
```suggestion
		for i, e := range p.Exprs {
			if i > 0 {
				c.write(", ")
			}
			c.expr(e, "")
```

zfmt/dag.go
```diff
@@ -589,6 +602,36 @@ func (c *canonDAG) scope(s *dag.Scope) {
...
 }
 
+func (c *canonDAG) semUserOp(o *dag.UserOp) {
+	c.head = true
+	c.write("op %s(", o.Name)
+	for i := range o.Params {
+		if i > 0 {
+			c.write(", ")
+		}
+		c.semParam(o.Params[i])
```
nwt, line 612:
```suggestion
	for i, p := range o.Params {
		if i > 0 {
			c.write(", ")
		}
		c.semParam(p)
```

zfmt/dag.go
```diff
@@ -589,6 +602,36 @@ func (c *canonDAG) scope(s *dag.Scope) {
 	}
 }
 
+func (c *canonDAG) semUserOp(o *dag.UserOp) {
```
nwt, line 605:
```suggestion
func (c *canonDAG) userOp(o *dag.UserOp) {
```

zfmt/dag.go
```diff
@@ -589,6 +602,36 @@ func (c *canonDAG) scope(s *dag.Scope) {
...
+	c.seq(o.Body)
+	c.close()
+	c.ret()
+	c.flush()
+	c.write(")")
+	c.ret()
+	c.flush()
+}
+
+func (c *canonDAG) semParam(p dag.Param) {
```
nwt, line 624:
```suggestion
func (c *canonDAG) param(p dag.Param) {
```

zfmt/ast.go
```diff
@@ -339,6 +339,24 @@ func (c *canon) decl(d ast.Decl) {
 		c.ret()
 		c.flush()
 		c.write(")")
+	case *ast.OpDecl:
+		c.write("op %s(", d.Name)
+		for k := range d.Params {
+			if k > 0 {
+				c.write(", ")
+			}
+			c.param(d.Params[k])
```
nwt, line 348:
```suggestion
		for k, p := range d.Params {
			if k > 0 {
				c.write(", ")
			}
			c.param(p)
```

docs/language/overview.md
```diff
@@ -457,13 +457,102 @@ scope but can override identifiers defined in ancestor scopes.
...
+<id> ( [<expr> [, <expr> ...]] )
+```
+where `<id>` is the identifier of the user-defined operator and `<expr>` is a list
+of [expressions](#8-expressions) matching the number of `<param>`s defined in
+the operator's signature.
+
+### 5.1 Sequence `this` Value
+
+The `this` value of a user-defined operator's sequence is a record value
+comprised of the paramters provided in the operator's signature.
```
philrz, line 484:
```suggestion
comprised of the parameters provided in the operator's signature.
```

zfmt/ztests/decls.yaml
```diff
@@ -10,7 +10,9 @@ inputs:
       const foo = "bar"
       func add(a,b): (a+b)
       const bar = "baz"
-      pass
+      op stamp(outer, assignee): ( yield {...outer, assignee, ts: now()} )
+      op addbar(...): ( pass )
```
nwt, line 14: Nit: This doesn't add bar.
```suggestion
      op nop(...): ( pass )
```

compiler/semantic/scope.go
```diff
@@ -84,42 +68,60 @@ func (s *Scope) DefineConst(zctx *zed.Context, name string, def dag.Expr) error
...
-	for k := len(s.stack) - 1; k >= 0; k-- {
-		if e, ok := s.stack[k].symbols[name]; ok {
-			e.refcnt++
-			return e.ref
+func (s *Scope) LookupExpr(name string) (dag.Expr, error) {
+	for scope := s; scope != nil; scope = scope.parent {
+		if entry, ok := scope.symbols[name]; ok {
+			e, ok := entry.ref.(dag.Expr)
+			if !ok {
+				return nil, fmt.Errorf("symbol %q is not an expression", name)
```
nwt, line 80:
```suggestion
				return nil, fmt.Errorf("symbol %q is not bound to an expression", name)
```

compiler/semantic/scope.go
```diff
@@ -84,42 +68,60 @@ func (s *Scope) DefineConst(zctx *zed.Context, name string, def dag.Expr) error
...
+		}
+	}
+	return nil, nil
+}
+
+func (s *Scope) LookupOp(name string) (*dag.UserOp, error) {
+	if entry := s.lookupEntry(name); entry != nil {
+		d, ok := entry.ref.(*dag.UserOp)
+		if !ok {
+			return nil, fmt.Errorf("symbol %q is not an operator", name)
```
nwt, line 93:
```suggestion
			return nil, fmt.Errorf("symbol %q is not bound to an operator", name)
```

compiler/semantic/scope.go
```diff
@@ -84,42 +68,60 @@ func (s *Scope) DefineConst(zctx *zed.Context, name string, def dag.Expr) error
...
 }
 
-func (s *Scope) Lookup(name string) dag.Expr {
-	for k := len(s.stack) - 1; k >= 0; k-- {
-		if e, ok := s.stack[k].symbols[name]; ok {
-			e.refcnt++
-			return e.ref
+func (s *Scope) LookupExpr(name string) (dag.Expr, error) {
+	for scope := s; scope != nil; scope = scope.parent {
+		if entry, ok := scope.symbols[name]; ok {
```
nwt, line 77: Can you use s.lookupEntry here?

compiler/semantic/analyzer.go
```diff
@@ -14,28 +14,114 @@ import (
 // to DAG form, resolving syntax ambiguities, and performing constant propagation.
 // After semantic analysis, the DAG is ready for either optimization or compilation.
 func Analyze(ctx context.Context, seq ast.Seq, source *data.Source, head *lakeparse.Commitish) (dag.Seq, error) {
-	c := newAnalyzer(ctx, source, head)
-	return c.check(seq)
+	a := newAnalyzer(ctx, source, head)
+	s, err := a.check(seq)
```
nwt, line 18: Maybe just call a.semSeq directly?
```suggestion
	s, err := a.semSeq(seq)
```

compiler/semantic/op.go
```diff
@@ -934,3 +1007,50 @@ func (a *analyzer) semCallOp(call *ast.Call, seq dag.Seq) (dag.Seq, error) {
...
+	// defer this until the check has completed and we have determined there is
+	// not a cycle in op calls.
+	a.resolve = append(a.resolve, func() error {
+		if op.Body == nil {
+			// This should never happen so panic if it does.
+			panic("op.Body is nil")
+		}
+		c.Body = op.Body
+		return nil
+	})
```
nwt, line 1054: How about replacing UserOpCall's Body field with a `UserOp *dag.UserOp` field so this resolve step can go away?

compiler/semantic/analyzer.go
mattnibs: Yeah...

compiler/semantic/scope.go
mattnibs: doh

compiler/semantic/op.go
mattnibs: I actually thought about this but didn't consider this w.r.t not needing the resolve. Another advantage is we can drop the Params argument in UserOpCall. Let's try it out.

compiler/semantic/scope.go
```diff
@@ -84,42 +68,58 @@ func (s *Scope) DefineConst(zctx *zed.Context, name string, def dag.Expr) error
...
-	return &Binder{symbols: make(map[string]*entry)}
-}
-
-func (b *Binder) Define(name string, ref dag.Expr) {
-	b.symbols[name] = &entry{ref: ref}
+func (s *Scope) sortedEntries() []*entry {
+	var entries []*entry
+	for _, entry := range s.symbols {
+		entries = append(entries, entry)
+	}
```
nwt, line 120:
```suggestion
	entries := maps.Values(s.symbols)
```

compiler/semantic/scope.go
```diff
@@ -7,68 +7,52 @@ import (
...
+		parent.children = append(parent.children, s)
+	}
+	return s
 }
 
-func (s *Scope) Exit() {
-	s.stack = s.stack[:len(s.stack)-1]
+type entry struct {
+	ref    any
+	refcnt int
```
nwt, line 30: Is this field used?

compiler/semantic/scope.go
mattnibs: ooh

compiler/semantic/scope.go
mattnibs: but I kind of think since it already exists it should be removed in a separate pr.

compiler/semantic/op.go
```diff
@@ -805,6 +814,65 @@ func (a *analyzer) semFuncBody(params []string, body ast.Expr) (dag.Expr, error)
...
+				return nil, err
+			}
+		}
+		if err := a.scope.DefineAs(op.Name, op); err != nil {
+			return nil, err
+		}
+		a.opDeclMap[op] = &opDecl{op: op}
+		ops[i] = op
+	}
+	// Now compiler op bodies.
```
nwt, line 838:
```suggestion
	// Now compile op bodies.
```

compiler/semantic/op.go
```diff
@@ -805,6 +814,65 @@ func (a *analyzer) semFuncBody(params []string, body ast.Expr) (dag.Expr, error)
...
+	a.enterScope()
+	defer func() {
+		a.opDecl = old
+		a.exitScope()
+	}()
+	var err error
+	if op.Body, err = a.semSeq(seq); err != nil {
+		return err
+	}
+	return nil
```
nwt, line 859: Nit: No need for defer or if here.
```suggestion
	var err error
	op.Body, err = a.semSeq(seq)
	a.opDecl = old
	a.exitScope()
	return err
```

compiler/semantic/op.go
```diff
@@ -934,3 +1007,38 @@ func (a *analyzer) semCallOp(call *ast.Call, seq dag.Seq) (dag.Seq, error) {
 	}
 	return append(seq, dag.NewFilter(c)), nil
 }
+
+// maybeConvertUserOp returns nil, nil if the call is determined to not be a
+// UserOp, otherwise it returns the compile op or the encountered error.
```
nwt, line 1012:
```suggestion
// UserOp, otherwise it returns the compiled op or the encountered error.
```

docs/commands/zq.md
```diff
@@ -366,12 +366,12 @@ where has(foo)
 ```
 while this query
 ```mdtest-command
-zq -C 'lower(foo)'
+zq -C 'a := x+1'
```
nwt, line 369: Nit: We usually omit these spaces.
```suggestion
zq -C 'a:=x+1'
```

docs/commands/zq.md
```diff
@@ -366,12 +366,12 @@ where has(foo)
...
 ```
-is an implied [yield operator](../language/operators/yield.md), which produces the lower case
-version of the presumed string in field `foo`, i.e.,
+is an implied [put operator](../language/operators/put.md), which creates a new field `a`
+with the value `x+1`, i.e.,
 ```mdtest-output
-yield lower(foo)
+put a := x+1
```
nwt, line 374: zfmt doesn't include these spaces.
```suggestion
put a:=x+1
```

compiler/semantic/analyzer.go
```diff
@@ -14,28 +14,104 @@ import (
...
+	path = append(path, op)
+	visited[op], onStack[op] = true, true
+	for _, neighbor := range a.opDeclMap[op].deps {
+		if !visited[neighbor] {
+			if p := a.isCyclic(neighbor, visited, onStack, path); p != nil {
+				return p
+			}
+		} else if onStack[neighbor] {
+			path = append(path, neighbor)
+			return path
```
nwt, line 99: You don't need `onStack` since `path` holds the current stack. And given that, `stack` would be a better name for it than `path`.
```suggestion
		} else if slices.Contains(path, neighbor) {
			return append(path, neighbor)
```

runtime/op/user/user.go
```diff
@@ -0,0 +1,34 @@
...
+	batch, err := p.parent.Pull(false)
+	if batch == nil || err != nil {
+		return nil, err
+	}
+	vals := make([]zed.Value, len(batch.Values()))
+	for i, v := range batch.Values() {
+		vals[i] = *p.expr.Eval(batch, &v)
+	}
+	return zbuf.NewBatch(batch, vals), nil
+}
```
nwt, line 34: If we merge #4627, you can use op.NewApplier instead of adding all this.

compiler/semantic/analyzer.go
mattnibs: Great point.

nwt (approved):
:shipit:!

PR #4420, opened 2023-03-03
single row implied aggregates default to values instead of records

### LINKED ISSUE #4418: Returning value rather than record from aggregation functions
Repro is with Zed commit 08ee509.

Consider the simple program:

```
$ zed -version
Version: v1.5.0-53-g08ee509e

$ echo '1 2 3' | zq -z 'collect(this)' -
{collect:[1,2,3]}
```

When writing Zed, many of us have found ourselves wanting just the value and not a record with a named field that holds the value. Therefore we've often had to write the verbose:

```
$ echo '1 2 3' | zq -z 'collect(this) | yield collect' -
[1,2,3]
```

It has been proposed that the behavior be changed so that a simple call to an aggregation function in this case would have returned just the value. If the user still wanted the prior behavior, they could use an explicit assignment, e.g., `collect:=collect(this)`.

mccanne:
This commit changes the semantics of aggregate functions of an implied summarize op to simply emit the values of the aggregation instead of putting them in a record.  This changes the UX for aggregations to be more Zed-like and less SQL-like when trying to form simple aggregations that result in a single result without one or more group-by keys.

Many of us found ourselves typing the pattern 'agg() | yield agg' and realized this approach would simplify things.  This way, 'echo "1 2 3"' | sum(this)' results in 6 instead of {sum:6}.

The change here is rather small but the ramifications on tests are non-trivial.

Fixes #4418


philrz:
I did just confirm that this change breaks a couple of the end-to-end Zui tests, but the fix looks trivial and is something I'll be able to take care of.

docs/language/aggregates/and.md
```diff
@@ -18,7 +18,7 @@ echo 'true false true' | zq -z 'and(this)' -
 ```
 =>
 ```mdtest-output
-{and:false}
+false
```
philrz, line 21: This PR can effectively be seen as one big bug fix. I see now that all this time the usage guidance for all our aggregate functions have looked like:

```
and(bool) -> bool
```

i.e., they've been showing the primitive return type this PR introduces and not the record type they've always returned. So now it finally matches. 😂

nwt:
@mccanne: I think we should do this only if we have a single aggregation. I say that because I think
```console
$ echo 1 2 3 | zq 'count(),sum(this)' -
{count:3(uint64),sum:6}
```
is both more intuitive and more useful than
```console
$ echo 1 2 3 | zq 'count(),sum(this)' -
3(uint64)
6
```

philrz:
Having read through the updated examples, I think I agree with @nwt's proposal of only doing it with a single aggregation.

Whatever we ultimately decide, somewhere in the docs (probably in `docs/language/operators/summarize.md`) I think we should have some brief-but-explicit text explaining when they can expect what's returned to be a value vs. a record.

mccanne:
Agreed!  Will fix.

docs/language/aggregates/and.md
mccanne: > Whatever we ultimately decide, somewhere in the docs (probably in `docs/language/operators/summarize.md`) I think we should have some brief-but-explicit text explaining when they can expect what's returned to be a value vs. a record.

Ok, just pushed an update to summarize.md

docs/language/operators/summarize.md
```diff
@@ -30,6 +30,11 @@ A key may be either an expression or a field.  If the key field is omitted,
 it is inferred from the expression, e.g., the field name for `by lower(s)`
 is `lower`.
 
+When the result of `summarize` is a single value (e.g., a single aggregate
+function without group-by keys) and there is no field name specified, then
+output is the Zed value of that result rather than a single-field record
```
nwt, line 35:
```suggestion
the output is that single value rather than a single-field record
```

docs/language/operators/summarize.md
```diff
@@ -44,7 +49,27 @@ echo '1 2 3 4' | zq -z 'summarize avg(this)' -
 ```
 =>
 ```mdtest-output
-{avg:2.5}
+2.5
+```
+
+To format the output of a single-valued aggregation into a record, simply specify
+an explicit field for the output as in:
```
nwt, line 56:
```suggestion
an explicit field for the output:
```

docs/language/operators/summarize.md
```diff
@@ -44,7 +49,27 @@ echo '1 2 3 4' | zq -z 'summarize avg(this)' -
...
+```mdtest-command
+echo '1 2 3 4' | zq -z 'summarize mean:=avg(this)' -
+```
+=>
+```mdtest-output
+{mean:2.5}
+```
+
+When multiple aggregate functions are specified, even without explicit field names,
+a record result is generated from the implied fields names of the functions:
```
nwt, line 66:
```suggestion
a record result is generated with field names implied by the functions:
```

nwt (approved):
Looks good!

Issue #2908, opened 2021-08-05
In-line doc example tests that are not shown to the reader


philrz:
The opening text of #2901 describes how it was a stopgap effort to give us some test coverage on docs examples where we want to show abbreviated Zed syntax or spare the reader from having to see all the inputs/outputs just so a test can be executed by mdtest in its present form. In a discussion with @nwt, we agreed that a better solution would be to have the test-not-shown-in-the-doc actually still live as part of the doc itself, but not be shown as part of the rendered page. This would help with maintenance since having the test code very close to the relevant docs text maximizes the likelihood that a developer that's fixing the broken test will also remember to update the corresponding docs text. Unfortunately it seems like markdown itself doesn't provide a great way to achieve this, but @nwt sensed he could cobble something together that leverages in-line HTML, since that way we could use comment blocks.

When we address this issue, I'll plan to modify the tests that were added as part of #2901 to take advantage of it, and assuming #2857 has not been addressed by then, I could also modify the two unprotected examples linked from that issue to use this approach (and perhaps decide this is "good enough" and close #2857).

philrz:
A specific variation of this came up recently in. In brief, we had an unprotected example in the docs where named ZSON named type syntax had changed from `port=(uint16)` to `port=uint16`. In https://github.com/brimdata/zed/pull/5000#issuecomment-1904518089 made this suggestion on how we might address:

> maybe for these kinds of tests we want to add a mdtest boomerang flag that 1) verifies it can serialize the value and 2) deserialize it to zson and 3) ensures the input/output values match?

PR #4422, opened 2023-03-03
Remove language doc ztests

philrz:
**tl;dr** - These tests are no longer needed and can be removed.

**Details** - These were originally added in #2901 after several incidents where non-mdtest-protected examples in the docs kept breaking and we had to catch/fix them manually. However, in the time since they were added, the examples they were protecting disappeared during a docs overhaul that brought language intro out of the top-level `docs/language/README.md` and into `docs/language/overview.md` with most of the content/examples being redone at that time. I also see that in the ~1.5 years since the ztests were added, per the commit log, nobody has ever had to make changes to them, so it seems we've gotten better about not breaking the fundamentals. If we want ideal coverage, perhaps at some point we can take on #2908.


nwt:
@philrz: Before merging this, let's verify that it doesn't reduce coverage.

nwt (approved):

nwt:
@philrz: Let's merge this.

PR #4437, opened 2023-03-15
Mention collect() with "zq -j"

### LINKED ISSUE #3875: zq JSON ouptut doesn't pass 'parse'
when using zq to create multi-line/record json files (-f json or -j) the resulting output won't parse as JSON without error..

this may have been an unintended side effect of collapsing ndjson and json output types into one??

the simplest way to replicate and see parse error is to open resulting json file in firefox where you'll get:
"SyntaxError: JSON.parse: unexpected non-whitespace character after JSON data at line 2 column 1 of the JSON data"

example output data which doesn't parse:
{"Date":"1994-12-31T00:00:00Z","AWE":630.1,"AAE":32765.2,"S":38000}
{"Date":"1995-06-30T00:00:00Z","AWE":648.2,"AAE":33706.4,"S":40000}
{"Date":"1995-12-31T00:00:00Z","AWE":661.2,"AAE":34382.4,"S":40000}

browser expecting, at minimum (array boundaries and separators):
[{"Date":"1994-12-31T00:00:00Z","AWE":630.1,"AAE":32765.2,"S":38000},
{"Date":"1995-06-30T00:00:00Z","AWE":648.2,"AAE":33706.4,"S":40000},
{"Date":"1995-12-31T00:00:00Z","AWE":661.2,"AAE":34382.4,"S":40000}]

philrz:
In #3875 a community user was tripped up by the subtle differences between the sequence of JSON values the Zed tooling outputs with `-j` vs. the single JSON values expected by much downstream tooling. Here I've attempted to clarify by emphasizing "values" (plural) and referencing `collect()`.

Closes #3875

### DIFF
--- a/docs/tutorials/zq.md
+++ b/docs/tutorials/zq.md
@@ -65,7 +65,7 @@ echo '"hello" "world"' | zq -j 'collect(this)' -
 
 which produces
 
-```
+```mdtest-output
 ["hello","world"]
 ```
 
--- a/docs/tutorials/zq.md
+++ b/docs/tutorials/zq.md
@@ -42,32 +42,14 @@ than `jq`.
 
 To this end, if you want full JSON compatibility without having to delve into the
 details of Zed, just use the `-j` option with `zq` and this will tell `zq` to
-expect JSON input and produce JSON output. For example:
+expect JSON values as input and produce JSON values as output, much like `jq`.
 
-```mdtest-command
-echo '"hello" "world"' | zq -j -
-```
-
-produces
-
-```mdtest-output
-"hello"
-"world"
-```
-
-If your downstream JSON tooling expects only a single JSON value, we can
-[`collect()`](../language/aggregates/collect.md) the sequence of values into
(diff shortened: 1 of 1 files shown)


mattnibs (approved):

philrz:
Thanks for the approval @mattnibs. As it turns out, @nwt gave me some verbal feedback on this one today and said that he felt that the examples may have interrupted the flow of this doc, and I can see how he's right about that. To maintain some incremental improvement while still trying to avoid a repeat of the issue that spawned this PR, I've pushed a commit to drop the examples and instead put the `collect()` reference in a "tip" since these are intended to point out such things without breaking the flow. In the preceding text I also spoke of input/output "values" (plural) to describe the default behavior and its similarity to `jq`, since despite the number of tools that demand JSON arrays as input, the precedent of `jq` shows we're not the only tool that works the way we do.

Here's the rendered section in Docusaurus:

image

nwt (approved):

PR #4447, opened 2023-03-21
solution to allowing keys of various types

### LINKED ISSUE #4167: Map with a key of union type returns error("missing") for all elements
Repro is with Zed commit 7396ea9. I originally bumped into this issue while reproducing #3360, but I have reason to believe it's got a separate root cause so am opening this one separately.

Start with this simple map and successfully yield a value from it.

```
$ zq -version
Version: v1.2.0-97-g7396ea9c

$ cat data.zson 
{
    m: |{
        1: "a"
    }|
}

$ zq -z 'yield m[1]' data.zson 
"a"
```

If I add an element with a key that has a different data type, I assume the key type now becomes a union and this would work. But instead I find that now I can't access any elements.

```
$ cat repro2.zson 
{
    m: |{
        1: "a",
        "foo": "bar"
    }|
}

$ zq -z 'yield m[1]' repro2.zson 
error("missing")

$ zq -z 'yield m["foo"]' repro2.zson 
error("missing")
```

Incidentally, I originally found this issue because I was programmatically creating a map and accidentally created an element with both key and value of `error("missing")`. Given what we just saw, unsurprisingly this shows the same problem.

```
$ cat repro1.zson 
{
    m: |{
        1: "a",
        error("missing"): error("missing")
    }|
}

$ zq -z 'yield m[1]' foo.zson 
error("missing")
```

dianetc:
Fixes #4167 by utilizing the ``BuildUnion`` function for proper encoding.


runtime/expr/eval.go
```diff
@@ -766,8 +766,20 @@ func indexMap(zctx *zed.Context, ectx Context, typ *zed.TypeMap, mapBytes zcode.
 		return zctx.Missing()
 	}
 	if key.Type != typ.KeyType {
-		// XXX issue #3360
-		return zctx.Missing()
+		if union, ok := zed.TypeUnder(typ.KeyType).(*zed.TypeUnion); ok {
+			for _, t := range union.Types {
+				if key.Type == t {
```
nwt, line 771: union.TagOf does roughly the same thing we're doing here, so we can simplify a little by calling it and checking for a nonnegative result.
```go
			if tag := union.TagOf(key.Type); tag >= 0 {
```

runtime/expr/eval.go
```diff
@@ -766,8 +766,20 @@ func indexMap(zctx *zed.Context, ectx Context, typ *zed.TypeMap, mapBytes zcode.
...
+					zed.BuildUnion(&b, union.TagOf(key.Type), key.Bytes)
+					if valBytes, ok := lookupKey(mapBytes, b.Bytes().Body()); ok {
+						return deunion(ectx, typ.ValType, valBytes)
+					}
+				}
+			}
+			return zctx.Missing()
+		} else {
+			return zctx.Missing()
+		}
```
nwt, line 782: Nit: I think we can simplify with just a single return here.
```suggestion
		}
		return zctx.Missing()
```

runtime/expr/ztests/union-key.yaml
nwt, line 1: Nit: Maybe call this `index-map-union-key.yaml` so it appears near some of the other indexing tests (which have names that begin with "index-") in file listings?

nwt (approved):
Looks good!

Issue #4444, opened 2023-03-21
String interpolation clash causing named capture group fail in regexp_replace()


philrz:
While verifying the changes for #4296, I spotted what I think may be a bug. Repro is with Zed commit 5b40e33.

I saw in the reviews of #4435 that we're using functionality described at https://pkg.go.dev/regexp#Regexp.Expand. Therefore I tried to use the example shown there in my verification. The following baseline shows it working much as it's shown in the Go docs.

```
$ zq -version
Version: v1.6.0-9-g5b40e33b

$ cat lines.txt 
# comment line
option1: value1
option2: value2

# another comment line
option3: value3

$ zq -i line lines.txt 
"# comment line"
"option1: value1"
"option2: value2"
""
"# another comment line"
"option3: value3"

$ zq -i line lines.txt | zq -z 'regexp_replace(this,/(?m)(?P<key>\w+):\s+(?P<value>\w+)$/,"$key=$value")' -
"# comment line"
"option1=value1"
"option2=value2"
null(string)
"# another comment line"
"option3=value3"
```

However, the docs also indicate the syntax `${key}` should work, and I find it doesn't.

```
$ zq -i line lines.txt | zq -z 'regexp_replace(this,/(?m)(?P<key>\w+):\s+(?P<value>\w+)$/,"${key}=${value}")' -
error("regexp_replace: string values required for all args")
error("regexp_replace: string values required for all args")
error("regexp_replace: string values required for all args")
error("regexp_replace: string values required for all args")
error("regexp_replace: string values required for all args")
error("regexp_replace: string values required for all args")
```

I think it's a clash with Zed's [string interpolation](https://zed.brimdata.io/docs/language/overview#7111-string-interpolation). 🙁

I sense it's needed in cases where the variable reference could be ambiguous. For instance, if I wanted the key/value output to be bookended by `FOO` and `BAR` and tried to write it without the `{}`, I don't get the `BAR`.

```
$ zq -i line lines.txt | zq -z 'regexp_replace(this,/(?m)(?P<key>\w+):\s+(?P<value>\w+)$/,"FOO$key=$valueBAR")' -
"# comment line"
"FOOoption1="
"FOOoption2="
null(string)
"# another comment line"
"FOOoption3="
```

As we're fixing this, it seems it would be good to add a test of named capture groups and an example in the docs that shows them.

philrz:
Verified in Zed commit e40a039.

[...]

PR #4435, opened 2023-03-14
Add regexp_replace function

mattnibs:
Closes #4296

PR #4449, opened 2023-03-21
regexp_replace doc improvements & test adjustment

### LINKED ISSUE #4296: String replacement for matches found via regexp
A community zync user recently inquired:

>  I think either `grep()` or `regexp()` should support the substitute function. Something like `grep /foo/Foo/s`.

In further chat, we up-leveled the requirement to the need to find the replacement targets via regexp. So in this regard we might think about enhancing `replace()` to accept a pattern as opposed to the simple string replacement it does today.

philrz:
While verifying #4296, I tried some of the enhancements that were added late in the #4435 reviews and found a few problems. Strictly speaking even these docs are not 100% safe to follow until #4444 is fixed, but for now it shows an example that's safe to use without the curly braces.

I'll put some comments in-line to explain some of the changes.

Fixes #4296

### DIFF
--- a/docs/language/functions/regexp_replace.md
+++ b/docs/language/functions/regexp_replace.md
@@ -16,8 +16,8 @@ the string `new`.
 Variables in `new` are replaced with corresponding matches drawn from `s`.
 A variable is a substring of the form `$name` or `${name}`, where `name` is a non-empty
 sequence of letters, digits, and underscores. A purely numeric name like `$1` refers
-to the submatch with the corresponding index; other names refer to parenthesized
-capturing groups named with the `(?P<name>...)` syntax. A reference to an out of range or
+to the submatch with the corresponding index; other names refer to capturing
+parentheses named with the `(?P<name>...)` syntax. A reference to an out of range or
 unmatched index or a name that is not present in the regular expression is replaced
 with an empty string.
 
--- a/docs/language/functions/regexp_replace.md
+++ b/docs/language/functions/regexp_replace.md
@@ -10,28 +10,50 @@ regexp_replace(s: string, re: string|regexp, new: string) -> string
 ### Description
 
 The _regexp_replace_ function substitutes all characters matching the
-[regular expression](../overview.md#regular-expressions) `re` in string `s` with
+[regular expression](../overview.md#811-regular-expressions) `re` in string `s` with
 the string `new`.
 
 Variables in `new` are replaced with corresponding matches drawn from `s`.
-A variable is a substring of the form $name or ${name}, where name is a non-empty
-sequence of letters, digits, and underscores. A purely numeric name like $1 refers
-to the submatch with the corresponding index; other names refer to capturing
-parentheses named with the `(?P<name>...)` syntax. A reference to an out of range or
+A variable is a substring of the form `$name` or `${name}`, where `name` is a non-empty
+sequence of letters, digits, and underscores. A purely numeric name like `$1` refers
+to the submatch with the corresponding index; other names refer to parenthesized
+capturing groups named with the `(?P<name>...)` syntax. A reference to an out of range or
 unmatched index or a name that is not present in the regular expression is replaced
-with an empty slice.
+with an empty string.
 
-In the $name form, name is taken to be as long as possible: $1x is equivalent to
-${1x}, not ${1}x, and, $10 is equivalent to ${10}, not ${1}0.
+In the `$name` form, `name` is taken to be as long as possible: `$1x` is equivalent to
+`${1x}`, not `${1}x`, and, `$10` is equivalent to `${10}`, not `${1}0`.
 
-To insert a literal $ in the output, use $$ in the template.
+To insert a literal `$` in the output, use `$$` in the template.
 
-#### Example:
+#### Examples:
+
+Replace regular expression matches with a letter:
 
 ```mdtest-command
-echo '"-ab-axxb-"' | zq -z 'yield regexp_replace(this, /a(x*)b/, "T")' -
+echo '"-ab-axxb-"' | zq -z 'yield regexp_replace(this, /ax*b/, "T")' -
 ```
 =>
 ```mdtest-output
 "-T-T-"
+```
+
+Replace regular expression matches using numeric references to submatches:
+
+```mdtest-command
+echo '"option: value"' | zq -z 'yield regexp_replace(this,/(\w+):\s+(\w+)$/,"$1=$2")' -
+```
+=>
+```mdtest-output
+"option=value"
+```
+
+Replace regular expression matches using named references:
+
+```mdtest-command
+echo '"option: value"' | zq -z 'yield regexp_replace(this,/(?P<key>\w+):\s+(?P<value>\w+)$/,"$key=$value")' -
+```
+=>
+```mdtest-output
+"option=value"
 ```
--- a/runtime/expr/function/ztests/regexp-replace.yaml
+++ b/runtime/expr/function/ztests/regexp-replace.yaml
@@ -1,7 +1,7 @@
 zed: "yield regexp_replace(in, re, new)"
 
 input: |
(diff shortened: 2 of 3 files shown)


docs/language/functions/regexp_replace.md
```diff
@@ -10,28 +10,50 @@ regexp_replace(s: string, re: string|regexp, new: string) -> string
...
 
 Variables in `new` are replaced with corresponding matches drawn from `s`.
-A variable is a substring of the form $name or ${name}, where name is a non-empty
-sequence of letters, digits, and underscores. A purely numeric name like $1 refers
-to the submatch with the corresponding index; other names refer to capturing
-parentheses named with the `(?P<name>...)` syntax. A reference to an out of range or
+A variable is a substring of the form `$name` or `${name}`, where `name` is a non-empty
+sequence of letters, digits, and underscores. A purely numeric name like `$1` refers
+to the submatch with the corresponding index; other names refer to parenthesized
+capturing groups named with the `(?P<name>...)` syntax. A reference to an out of range or
```
philrz, line 20: In the top of this doc there's a link to the [regular expressions section](https://zed.brimdata.io/docs/language/overview#811-regular-expressions) of the Zed language overview doc, which in turn points to the [RE2](https://github.com/google/re2) repo, so strictly speaking I guess it's fair game to quite the RE2 docs verbatim like we'd done here. However, it still seems likely users may try to work off this doc in isolation, so I wanted to improve the wording a bit to help along those that don't have as much regexp experience. While "capturing parentheses" is indeed a phrase used in the [Expand docs](https://pkg.go.dev/regexp#Regexp.Expand) being quoted here, the phrase "parenthesized subexpressions (also known as capturing groups)" appears higher up in that same doc, so I've shortened that and proposed it instead since I happen to find it clearer.

docs/language/functions/regexp_replace.md
```diff
@@ -10,28 +10,50 @@ regexp_replace(s: string, re: string|regexp, new: string) -> string
...
 unmatched index or a name that is not present in the regular expression is replaced
-with an empty slice.
+with an empty string.
```
philrz, line 22: While [slices](https://zed.brimdata.io/docs/language/overview#77-slices) are present in Zed, I'm pretty sure we want to say "string" here.

docs/language/functions/regexp_replace.md
```diff
@@ -10,28 +10,50 @@ regexp_replace(s: string, re: string|regexp, new: string) -> string
...
+To insert a literal `$` in the output, use `$$` in the template.
 
-#### Example:
+#### Examples:
+
+Replace regular expression matches with a letter:
 
 ```mdtest-command
-echo '"-ab-axxb-"' | zq -z 'yield regexp_replace(this, /a(x*)b/, "T")' -
+echo '"-ab-axxb-"' | zq -z 'yield regexp_replace(this, /ax*b/, "T")' -
```
philrz, line 34: As far as I can tell, the parens aren't needed here since we're not referencing the contents of the capturing group. I made a similar adjustment in the ztest.

runtime/expr/function/ztests/regexp-replace.yaml
```diff
@@ -1,8 +1,9 @@
 zed: "yield regexp_replace(in, re, new)"
 
 input: |
-  {in:"-ab-axxb-",re:"a(x*)b",new:"T"}
+  {in:"-ab-axxb-",re:"ax*b",new:"T"}
   {in:"-ab-axxb-",re:"a(x*)b",new:"$1"}
+  {in:"-ab-axxb-",re:"a(?P<ZeroOrMoreX>x*)b",new:"foo$ZeroOrMoreX bar"}
```
philrz, line 6: The space char included in ` bar` here is intentional. When #4444 is addressed I'd propose adding another test that uses curly braces so the space char can be dropped.

docs/language/functions/regexp_replace.md
```diff
@@ -10,28 +10,50 @@ regexp_replace(s: string, re: string|regexp, new: string) -> string
 ### Description
 
 The _regexp_replace_ function substitutes all characters matching the
-[regular expression](../overview.md#regular-expressions) `re` in string `s` with
+[regular expression](../overview.md#811-regular-expressions) `re` in string `s` with
```
nwt, line 13: Do you know why .github/workflows/markdown-lint.yml didn't catch as part of #4435?

runtime/expr/function/ztests/regexp-replace.yaml
```diff
@@ -1,8 +1,9 @@
 zed: "yield regexp_replace(in, re, new)"
 
 input: |
-  {in:"-ab-axxb-",re:"a(x*)b",new:"T"}
+  {in:"-ab-axxb-",re:"ax*b",new:"T"}
```
nwt, line 4: Keep this as is so it remains the same as the expression on the next line.
```suggestion
  {in:"-ab-axxb-",re:"a(x*)b",new:"T"}
```

runtime/expr/function/ztests/regexp-replace.yaml
nwt, line 6: Use less text so this is more readable.
```suggestion
  {in:"-ab-axxb-",re:"a(?P<X>x*)b",new:"$X"}
```

docs/language/functions/regexp_replace.md
nwt: I like the original "capturing parentheses" better.

docs/language/functions/regexp_replace.md
philrz: @nwt: Well, I'm not surprised by it... does that count? 😄  I first confronted this limitation in #4028. While the [markdown link checker](https://github.com/gaurav-nelson/github-action-markdown-link-check) we use can detect broken anchor links within the same doc, it's never been smart enough to detect them when pointing to other docs. I think this is partially GitHub's fault because if you follow the link you get an HTTP 200 and it puts you at the top of the doc and sends back an HTTP 200. Here's that with the broken link of the moment:

```
$ curl -s -o /dev/null -w "%{http_code}" "https://github.com/brimdata/zed/blob/main/docs/language/overview.md#regular-expressions"
200
```

Meanwhile the [link checker we use on deployed sites](https://github.com/ruzickap/action-my-broken-link-checker) has always been able to catch these. I can't say why all the way down to the code level since the behavior & response code seem much the same on the rendered docs site.

```
$ curl -s -o /dev/null -w "%{http_code}" "https://zed.brimdata.io/docs/next/language/overview#regular-expressions"
200
```

However, that link checker is somehow still able to see this as a breakage, as shown in the output of a [recent zed-docs-site Actions run](https://github.com/brimdata/zed-docs-site/actions/runs/4483517347/jobs/7882899690):

```
...
2023-03-21T20:31:56.3361518Z https://zed.brimdata.io/docs/next/language/functions/regexp_replace
...
2023-03-21T20:31:56.3381315Z    id #regular-expressions not found       https://zed.brimdata.io/docs/next/language/overview#regular-expressions
...
```

If I had to guess, I think the [muffet](https://github.com/raviqqe/muffet) tool that's used by the checker that's catching the breakage is more sophisticated, whereas the markdown link checker depends on https://github.com/tcort/markdown-link-check which looks simpler and has a bunch of confusing open/closed issues grumping about problems with anchors, and as far as I can tell they've decided not to cover this case. I can't remember if at one time I explored phasing out the markdown-level one altogether but I seem to recall that treating `https://github.com/brimdata/zed/tree/main/docs` like a rendered site caused its own set of problems (HTTP 429, maybe?). So long ago I resolved that it was ok to have a "belt & suspenders" approach where the markdown one caught _most_ of the breakages early and then the second link checker was there just in case any snuck by.

docs/language/functions/regexp_replace.md
philrz: @nwt: Ok, I've put back "capturing parentheses".

PR #4451, opened 2023-03-22
Allow head operator to take in expression

### LINKED ISSUE #4168: Have "head" accept an expr rather than only a static integer
A community zync user inquired:

> Do you have any idea why `head 10` works while `head 10+1` (with an expression) doesn't?

A look at the [docs](https://zed.brimdata.io/docs/next/language/operators/head) indicates this is an expected limitation of the current implementation, since it shows the parameter being an integer `[ n ]` rather than `[ expr ]`. I'm not certain if it would be feasible/easy to accept an expression in this context, but this user indicated they have a likely use for this in a Zed BI proxy server they're building. If we end up doing it for `head`, surely we'd want `tail` to have the same treatment. Looking through the docs, I don't see any other operators that take a simple integer parameter at the moment.

dianetc:
Fixes #4168
The logic allows the `head` operation to take in an `<expr>` so that calls such as `head 1+1`  can be interpreted as `head 2`. Note, this does not allow for variable expressions, i.e.,  `head x`  fails.


compiler/parser/parser.peg
```diff
@@ -367,8 +367,9 @@ DropOp
     }
 
 HeadOp
-  = "head" _ count:UInt { RETURN(MAP("kind": "Head", "count": count)) }
-  / "head" { RETURN(MAP("kind": "Head", "count": 1)) }
+  = "head" _ !EndOfOp count:Expr { RETURN(MAP("kind": "Head", "count": count)) }
+  / "head"  { RETURN(MAP("kind": "Head", "count": MAP("kind": "Primitive", "type": "int64", "text": "1"))) }
+
```
nwt, line 372:
```suggestion
```

compiler/semantic/op.go
```diff
@@ -418,13 +418,14 @@ func semOp(ctx context.Context, scope *Scope, o ast.Op, ds *data.Source, head *l
 			NullsFirst: o.NullsFirst,
 		}, nil
 	case *ast.Head:
-		limit := o.Count
-		if limit == 0 {
-			limit = 1
+		expr, _ := semExpr(scope, o.Count)
+		val, _ := kernel.EvalAtCompileTime(scope.zctx, expr)
```
nwt, line 422: We should handle these errors.
```suggestion
		expr, err := semExpr(scope, o.Count)
		if err != nil {
			return nil, fmt.Errorf("head: %w", err)
		}		
		val, err := kernel.EvalAtCompileTime(scope.zctx, expr)
		if err != nil {
			return nil, fmt.Errorf("head: %w", err)
		}		
```

compiler/semantic/op.go
```diff
@@ -418,13 +418,14 @@ func semOp(ctx context.Context, scope *Scope, o ast.Op, ds *data.Source, head *l
...
+		if val.AsInt() < 0 {
+			return nil, errors.New("expression must be of positive value")
```
nwt, line 424:
```suggestion
		if val.AsInt() < 1 {
			return nil, errors.New("head: expression must be of positive value")
```

compiler/semantic/op.go
```diff
@@ -418,13 +418,20 @@ func semOp(ctx context.Context, scope *Scope, o ast.Op, ds *data.Source, head *l
...
+		expr, err := semExpr(scope, o.Count)
+		if err != nil {
+			return nil, fmt.Errorf("head: %w", err)
+		}
+		val, err := kernel.EvalAtCompileTime(scope.zctx, expr)
+		if err != nil {
+			return nil, fmt.Errorf("head: %w", err)
+		}
+		if val.AsInt() < 1 {
+			return nil, errors.New("head: expression must be of positive value")
```
nwt, line 430: After testing this branch some, I think we ought to tune the error message a bit, both to make clear that the value must be a positive integer and to show what the actual value is.
```suggestion
			return nil, fmt.Errorf("head: expression value is not a positive integer: %s", zson.MustFormatValue(val))
```

runtime/op/traverse/ztests/head.yaml
```diff
@@ -1,4 +1,4 @@
-zed: over a | head 2
+zed: over a | head 2+1
```
nwt, line 1: Rather than change this test, whose primary intent is to cover the interaction between `over` and `head`, let's add `compiler/ztests/head.yaml` with a few `zc -C -s` tests (similar to `compiler/ztests/from-error.yaml`).

Test cases I think we should include:
1. `head 1`
2. `const x=1 head x + 1`
3. `head 1.`
4. `head "1"`
5. `head x`

(`zc -C -s` should succeed for the first two and fail for the last three.)

nwt (commented):
@dianetc: This looks great! I added a couple last comments (about an error message and tests). We should also update `docs/language/operators/head.md` to change `n` in the synopsis and description to an `<expr>` with a note about the requirement that it evaluate to a positive integer value at compile time.

PR #4456, opened 2023-03-23
zed manage: Fix disabled config

mattnibs:
The logic to create tasks for each pool was not taking into account pool specific configurations. Fix this.


nwt (approved):

nwt:
Is this worth a test?

mattnibs:
> Is this worth a test?

Not right now.

PR #4465, opened 2023-03-28
Add docs for -log.level option of "zed serve"

### LINKED ISSUE #4448: Document -log.level on "zed serve"
At the time of the filing of this issue, Zed is at commit 052f2c0.

A community zync user pointed out that we currently lack documentation of the `-log.level` option on `zed serve`, and indeed this should be added to https://zed.brimdata.io/docs/commands/zed. I think we'd want to add:

1. What are the options (I think it's the ones listed [here](https://github.com/uber-go/zap/blob/845ca51d5b8d9fed9fe14f35ab13b6b160d5762d/level.go#L131-L133))
2. What's the default if `-log.level` is not specified (`info`, per `zed serve -h`)
3. Some basic guidance, e.g.:
   * `debug` when debugging issues
   * If the default is too verbose for your production use, what do we recommend instead (`warn`?)

philrz:
A community zync user reported having looked in the current Zed docs hoping to find guidance on the `zed serve` log levels, but came up empty. While we don't currently try to exhaustively cover each available option at the docs site and instead rely on built-in help for much of that, I figure if a user went looking for it on the docs site and took the time to ask for guidance when nothing came back, that means it's likely to come up again. This PR adds some minimal guidance to cover that for the future.

I've also adjusted some of the top-level "section 2" intro text to be more explicit about the fact that we're only describing highlights of the available options.

Fixes #4448

### DIFF
--- a/docs/commands/zed.md
+++ b/docs/commands/zed.md
@@ -389,7 +389,7 @@ or [`kubectl`](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-co
 commands.
 
 The following sections describe each of the available commands and highlight
-some key options. Built-in help shows the commands and their options:
+some key options.  Built-in help shows the commands and their options:
 
 * `zed -h` with no args displays a list of `zed` commands.
 * `zed command -h`, where `command` is a sub-command, displays help
@@ -862,7 +862,7 @@ specified by the `-l` option, executes the requests, and returns results.
 
 The `-log.level` option controls log verbosity.  Available levels, ordered
 from most to least verbose, are `debug`, `info` (the default), `warn`,
-`error`, `dpanic`, `panic`, and `fatal`. Suggestions for use:
+`error`, `dpanic`, `panic`, and `fatal`.  Suggestions for use:
 
 * As its name implies, `debug` level is likely to only be required when actively debugging issues.
 
--- a/docs/commands/zed.md
+++ b/docs/commands/zed.md
@@ -389,8 +389,7 @@ or [`kubectl`](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-co
 commands.
 
 The following sections describe each of the available commands and highlight
-some key options. Built-in help shows all commands and options that can be
-invoked as follows:
+some key options. Built-in help shows the commands and their options:
 
 * `zed -h` with no args displays a list of `zed` commands.
 * `zed command -h`, where `command` is a sub-command, displays help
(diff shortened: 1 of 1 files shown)


mattnibs (approved):

docs/commands/zed.md
```diff
@@ -384,13 +384,15 @@ case but this design allows different workloads like these to be custom tuned.
...
+[`docker`](https://docs.docker.com/engine/reference/commandline/cli/)
+or [`kubectl`](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands)
 commands.
 
-The following sections describe each of the available commands, but built-in
-help is also available:
-* `zed -h` with no args displays a list of zed commands.
+The following sections describe each of the available commands and highlight
+some key options. Built-in help shows all commands and options that can be
+invoked as follows:
```
nwt, line 393: 1. The new sentence reads oddly to me.
2. Prevailing style in here is two spaces between sentences.
```suggestion
some key options. Built-in help shows the commands and their options:
```

docs/commands/zed.md
```diff
@@ -854,11 +856,20 @@ pool `<existing>`, which may be referenced by its ID or its previous name.
...
 specified by the `-l` option, executes the requests, and returns results.
 
+The verbosity of the server's log output can be varied with the `-log.level`
+option. Available levels ordered from most to least verbose are: `debug`, `info` (the
+default), `warn`, `error`, `dpanic`, `panic`, and `fatal`. Suggestions for use:
+
+* As its name implies, `debug` level is likely to only be required when actively debugging issues.
+
+* If the volume of logging output at the default `info` level seems too
+excessive for production use, `warn` level is recommended.
```
nwt, line 871: These suggestions feel quite obvious. I wouldn't waste the reader's time with them.

nwt (approved):

docs/commands/zed.md
philrz: @nwt: I've shortened the suggestions to drop the `debug` mention (yeah, you're right that it's very obvious) but left the suggestion of `warn`. Quoting the user that spawned the issue:

> Is production, Zed Serve log to many info, I know that exist zed serve `-log.level=debug` to debug but, did exist a log level that show only warning and error? I searched on https://zed.brimdata.io/docs  I did not found details about log.level.
> After try and error, I found -log.level=warn look like what I need.

While it's possible they'd get what they need from the ordered list alone, I think it's worth making a specific recommendation to save them from hacking around with it more than necessary. Speaking just for myself, I've always had a terrible time remembering log levels and have to look them up every time, so I'd have wanted something like this if I was a user, too.

PR #4466, opened 2023-03-28
seek index: fix incorrect entry min/max values

mattnibs:
Fix a bug occurring when seek indexes are generated that produced
incorrect values for the min and max values for seek entries

### DIFF
--- a/lake/data/writer.go
+++ b/lake/data/writer.go
@@ -64,17 +64,12 @@ func (o *Object) NewWriter(ctx context.Context, engine storage.Engine, path *sto
 
 func (w *Writer) Write(val *zed.Value) error {
 	w.count++
-	key := val.DerefPath(w.poolKey).MissingAsNull()
-	if w.seekIndex != nil {
-		if err := w.writeIndex(key); err != nil {
-			return err
-		}
-	}
 	if err := w.writer.Write(val); err != nil {
 		return err
 	}
+	key := val.DerefPath(w.poolKey).MissingAsNull()
 	w.object.Max.CopyFrom(key)
-	return nil
+	return w.writeIndex(key)
 }
 
 func (w *Writer) writeIndex(key *zed.Value) error {
@@ -86,7 +81,6 @@ func (w *Writer) writeIndex(key *zed.Value) error {
 	if w.seekMin == nil {
 		w.seekMin = key.Copy()
 	}
-	w.object.Max.CopyFrom(key)
 	if w.seekIndexTrigger < w.seekIndexStride {
 		return nil
 	}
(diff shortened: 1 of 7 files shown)


nwt (approved):
1. PR title says "inclusive inclusive".
2. I don't think the PR description is right. The correctness bug I've observed is that seekindex.Entry.Min can be larger than the key for the first value in that entry's range. It's also true that seekindex.Entry.Max can be larger than the key for the last value in that entry's range, and that's a bug, but it affect performance, not correctness.

mattnibs:
> * I don't think the PR description is right. The correctness bug I've observed is that seekindex.Entry.Min can be larger than the key for the first value in that entry's range. It's also true that seekindex.Entry.Max can be larger than the key for the last value in that entry's range, and that's a bug, but it affect performance, not correctness.

Try and run the tests on this branch on main and you'll notice that they fail. The correct value is omitted. Pretty sure this is a correctness bug. I have not observed any problem with the seekindex.Entry.Min, nor looking at the code could I see how that could happen. Could you point me to a repo of that?

mattnibs:
> * PR title says "inclusive inclusive".
> * I don't think the PR description is right. The correctness bug I've observed is that seekindex.Entry.Min can be larger than the key for the first value in that entry's range. It's also true that seekindex.Entry.Max can be larger than the key for the last value in that entry's range, and that's a bug, but it affect performance, not correctness.

Try and run the tests on this branch on main and you'll notice that they fail. The correct values are omitted. Pretty sure this is a correctness bug. The problem stems from the fact that writeIndex is called before the writing of the value which causes the current value to get marked as the max of the present stream and the next value to get marked as the min value of the new stream. Thus if I do an an equals compare on the max value of the entry it reads the first stream and skips the second stream (since the min of the seconds stream is greater than the max of the first stream), which is a problem since the value is in the second stream and not the first.

nwt:
> The pr title is wrong I'll fix.

It still says "inclusive inclusive".

> Try and run the tests on this branch on main and you'll notice that they fail. The correct values are omitted. Pretty sure this is a correctness bug.

There's definitely a correctness bug and this PR definitely fixes it. I didn't mean to dispute that and could have been clearer there.

> The problem stems from the fact that writeIndex is called before the writing of the value which causes the current value to get marked as the max of the present stream and the next value to get marked as the min value of the new stream.

Agreed.

> Thus if I do an an equals compare on the max value of the entry it reads the first stream and skips the second stream (since the min of the seconds stream is greater than the max of the first stream), which is a problem since the value is in the second stream and not the first.

This is where we disagree. The problem doesn't arise when the min of the second stream is greater than the max of the first stream. That's often the case. The problem instead arises when a stream's min (i.e., its seekindex.Entry.Min) is greater than the first stream value's key, and that can currently happen whenever the first value's key is less than the second value's key. That (i.e., a stream min that's larger than the first stream value's key) is what's causing us to miss values.

PR #4471, opened 2023-03-30
Load operator that adds and commits data to a pool

### LINKED ISSUE #2765: Pool "load" operator in Zed
tl;dr - Inspired by the kinds of ideas proposed in #2754, @mccanne thought of the idea of a `load` operator that could be used for populating scratch Pools based on data from other Pools.

Consider the following workflow using current Zed commit 7adbb36 and the [Zeek zed-sample-data](https://github.com/brimdata/zed-sample-data/tree/main/zeek-default). Let's say I start by loading it all to one pool:

```
$ rm -rf $ZED_LAKE_ROOT

$ zed lake create -p samples
pool created: samples

$ zed lake load -p samples *
1szVwPnjln5K2tiUUXSckKGxqQ3 committed 1 segments

$ zed lake query -z 'from samples | count() by _path'
{_path:"http",count:144034 (uint64)} (=0)
{_path:"ntp",count:904} (0)
{_path:"ssh",count:22} (0)
...
```

If I wanted to extract a subset of that data to another pool, this can be done currently using a pipeline that invokes Zed multiple times:

```
$ zed lake create -p weird
pool created: weird

$ zed lake query 'from samples | _path=="weird"' | zed lake load -p weird -
1szWGf4n4AUAfFElCHtVxjwa05m committed 1 segments

$ zed lake query -z 'from weird | count() by _path'
{_path:"weird",count:24048 (uint64)} (=0)
```

However, @mccanne pointed out that a `load` operator could potentially be added in Zed to accomplish this in one shot, such as:

```
$ zed lake query 'from samples | _path=="weird" | load weird'
```

This could be particularly useful in cloud-centric use cases, since it would open up the potential to create scratch pools consisting of subsets of larger data sets, without having to require data to bounce all the way down to the client just to be sent back up to the cloud again.

dianetc:
Fixes #2765

This PR creates a `load` operator that  populates the specified `<pool>` with the values it
receives as input. The operator is used to save query results from your Zed
pipeline to a pool in the same Zed lake


compiler/optimizer/op.go
```diff
@@ -36,7 +36,7 @@ func (o *Optimizer) analyzeSortKey(op dag.Op, in order.SortKey) (order.SortKey,
 	case *dag.Lister:
 		// This shouldn't happen.
 		return order.Nil, errors.New("internal error: dag.Lister encountered in anaylzeSortKey")
-	case *dag.Filter, *dag.Head, *dag.Pass, *dag.Uniq, *dag.Tail, *dag.Fuse:
+	case *dag.Filter, *dag.Head, *dag.Pass, *dag.Uniq, *dag.Tail, *dag.Fuse, *dag.Load:
```
nwt, line 39: This clobbers the input order, so we should just let the default case catch it.
```suggestion
	case *dag.Filter, *dag.Head, *dag.Pass, *dag.Uniq, *dag.Tail, *dag.Fuse:
```

compiler/ast/ast.go
```diff
@@ -377,6 +377,14 @@ type (
 		Kind   string  `json:"kind" unpack:""`
 		Trunks []Trunk `json:"trunks"`
 	}
+	Load struct {
+		Kind    string `json:"kind" unpack:""`
+		Pool    string `json:"pool"`
+		Author  string `json:"author"`
+		Message string `json:"message"`
+		Meta    string `json:"meta"`
+		Branch  string `json:"branch"`
```
nwt, line 386: Put Branch after Pool since they're coupled.
```suggestion
		Pool    string `json:"pool"`
		Branch  string `json:"branch"`
		Author  string `json:"author"`
		Message string `json:"message"`
		Meta    string `json:"meta"`
```

compiler/ast/dag/op.go
```diff
@@ -58,6 +58,14 @@ type (
 		RightKey Expr         `json:"right_key"`
 		Args     []Assignment `json:"args"`
 	}
+	Load struct {
+		Kind    string `json:"kind" unpack:""`
+		Pool    string `json:"pool"`
+		Author  string `json:"author"`
+		Message string `json:"message"`
+		Meta    string `json:"meta"`
+		Branch  string `json:"branch"`
```
nwt, line 67: Put Branch after Pool since they're coupled.
```suggestion
		Kind    string `json:"kind" unpack:""`
		Pool    string `json:"pool"`
		Branch  string `json:"branch"`
		Author  string `json:"author"`
		Message string `json:"message"`
		Meta    string `json:"meta"`
```

compiler/parser/parser.peg
```diff
@@ -478,6 +479,23 @@ SampleExpr
   = _ lval:Lval { RETURN(lval)}
   / "" { RETURN(MAP("kind":"ID", "name":"this")) }
 
+LoadOp
+    = "load" _ pool:PoolNameString author:AuthorArg? message:MessageArg? meta:MetaArg? branch:BranchSpecification? {
```
nwt, line 483: Put branch immediately after pool so the syntax is "pool@branch", which matches what we do with PoolCommit.
```suggestion
    = "load" _ pool:PoolNameString branch:BranchSpecification? author:AuthorArg? message:MessageArg? meta:MetaArg?  {
```

compiler/parser/parser.peg
```diff
@@ -478,6 +479,23 @@ SampleExpr
...
+  = _ "author" _ val:QuotedString { RETURN(val) }
+
+MessageArg
+  = _ "message" _ val:QuotedString { RETURN(val) }
+
+MetaArg
+  = _ "meta" _ val:QuotedString { RETURN(val) }
+
+BranchSpecification
+  = _ "@" _ branch:QuotedString { RETURN(branch) }
```
nwt, line 497: 1. This is pretty similar to PoolCommit, so call it PoolBranch.
2. In keeping with PoolCommit, don't require (or allow) spaces around "@".
3. Allow PoolIdentifier in addition to QuotedString so we can say `pool@main` instead of having to say `pool@"main"`.
```suggestion
PoolBranch
  = "@" branch:(PoolIdentifier / QuotedString) { RETURN(branch) }
```

compiler/parser/parser.peg
```diff
@@ -478,6 +479,23 @@ SampleExpr
   = _ lval:Lval { RETURN(lval)}
   / "" { RETURN(MAP("kind":"ID", "name":"this")) }
 
+LoadOp
+    = "load" _ pool:PoolNameString branch:PoolBranch? author:AuthorArg? message:MessageArg? meta:MetaArg? {
+        RETURN(MAP("kind": "Load", "pool": pool, "branch": branch, "author": author, "message": message, "meta": meta))
```
nwt, line 484:
```suggestion
  = "load" _ pool:PoolNameString branch:PoolBranch? author:AuthorArg? message:MessageArg? meta:MetaArg? {
      RETURN(MAP("kind": "Load", "pool": pool, "branch": branch, "author": author, "message": message, "meta": meta))
```

zfmt/dag.go
```diff
@@ -364,6 +364,9 @@ func (c *canonDAG) op(p dag.Op) {
 			c.space()
 			c.exprs(p.Args)
 		}
+	case *dag.Load:
+		c.next()
+		c.write("load %s", p.Pool)
```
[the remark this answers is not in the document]
dianetc: This was so that we have an output of 
```
from (
  pool 2NyhaM5nfRSwx8Wimz1rpT5KZWj
  pool 2NyhaM5nfRSwx8Wimz1rpT5KZWj
)
| merge ts:desc
| load dst <---- here
```
so by "handle other fields" do you mean something like:
```
c.next()
c.write("load %s, %s, %s, %s, %s", p.Pool, p.Branch, p.Author, p.Message, p.Meta)
```
so  you'd get `load pool, branch, author, message, meta` printed out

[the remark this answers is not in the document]
nwt: Yes, but match the Zed syntax, so `load <pool>@<branch> author <author> message <message> meta <meta>` if all options are specified, `load <pool>` if just the pool is specified, `load <pool>@<branch>` if pool and branch are specified, etc.

zfmt/ast.go
```diff
@@ -518,6 +518,22 @@ func (c *canon) proc(p ast.Op) {
 			c.space()
 			c.exprs(p.Args)
 		}
+	case *ast.Load:
+		c.next()
+		c.write("load %s", p.Pool)
+		if len(p.Branch) != 0 {
```
nwt, line 524: The conventional way to check for a non-empty string is `s != ""` rather than `len(s) != 0`.
```suggestion
		if len(p.Branch) != 0 {
```

zfmt/ast.go
```diff
@@ -518,6 +518,22 @@ func (c *canon) proc(p ast.Op) {
...
+			c.write("@%s", p.Branch)
```
nwt, line 525:
```suggestion
			c.write(p.Branch)
```

zfmt/ast.go
```diff
@@ -518,6 +518,21 @@ func (c *canon) proc(p ast.Op) {
 			c.space()
 			c.exprs(p.Args)
 		}
+	case *ast.Load:
+		c.next()
+		c.write("load %s", zson.QuotedString([]byte(p.Pool)))
+		if p.Branch != "" {
+			c.write(p.Branch)
```
nwt, line 525: I wasn't thinking straight when I suggested changing this to `c.write(p.Branch)`! We definitely want the "@" in there.
```suggestion
			c.write("@%s", p.Branch)
```

zfmt/ast.go
```diff
@@ -518,6 +518,21 @@ func (c *canon) proc(p ast.Op) {
...
+		}
+		if p.Author != "" {
+			c.write(" author %s ", p.Author)
```
nwt, line 528: We want just a leading space on these so they'll look good regardless of which are set.
```suggestion
			c.write(" author %s", p.Author)
```

zfmt/ast.go
```diff
@@ -518,6 +518,21 @@ func (c *canon) proc(p ast.Op) {
...
+		}
+		if p.Message != "" {
+			c.write("message %s ", p.Message)
```
nwt, line 531:
```suggestion
			c.write(" message %s", p.Message)
```

zfmt/ast.go
```diff
@@ -518,6 +518,21 @@ func (c *canon) proc(p ast.Op) {
...
+		}
+		if p.Meta != "" {
+			c.write("meta %s", p.Meta)
```
nwt, line 534:
```suggestion
			c.write(" meta %s", p.Meta)
```

zfmt/dag.go
```diff
@@ -364,6 +364,21 @@ func (c *canonDAG) op(p dag.Op) {
...
 			c.exprs(p.Args)
 		}
+	case *dag.Load:
+		c.next()
+		c.write("load %s ", p.Pool)
+		if p.Branch != "" {
+			c.write("@%s", p.Branch)
+		}
+		if p.Author != "" {
+			c.write(" author %s ", p.Author)
```
nwt, line 374:
```suggestion
			c.write(" author %s", p.Author)
```

zfmt/dag.go
```diff
@@ -364,6 +364,21 @@ func (c *canonDAG) op(p dag.Op) {
...
+		}
+		if p.Message != "" {
+			c.write("message %s ", p.Message)
```
nwt, line 377:
```suggestion
			c.write(" message %s", p.Message)
```

zfmt/dag.go
```diff
@@ -364,6 +364,21 @@ func (c *canonDAG) op(p dag.Op) {
...
+		}
+		if p.Meta != "" {
+			c.write("meta %s", p.Meta)
```
nwt, line 380:
```suggestion
			c.write(" meta %s", p.Meta)
```

compiler/optimizer/parallelize.go
```diff
@@ -253,7 +253,7 @@ func (o *Optimizer) concurrentPath(ops []dag.Op, sortKey order.SortKey) (int, or
 				return 0, order.Nil, false, false, nil
 			}
 			return k, newKey, false, true, nil
-		case *dag.Parallel, *dag.Head, *dag.Tail, *dag.Uniq, *dag.Fuse, *dag.Sequential, *dag.Join:
+		case *dag.Parallel, *dag.Head, *dag.Tail, *dag.Uniq, *dag.Fuse, *dag.Sequential, *dag.Join, *dag.Load:
```
nwt, line 256: dag.Load wants ordered input (hence `true, true`) but doesn't produce ordered output (hence `order.Nil`).
```suggestion
		case *dag.Load:
			return k, order.Nil, true, true, nil
		case *dag.Parallel, *dag.Head, *dag.Tail, *dag.Uniq, *dag.Fuse, *dag.Sequential, *dag.Join:
```

philrz (commented):
@dianetc: Could you add a line for this operator in `docs/language/operators/README.md`?

philrz:
I think `meta` might need a little more attention. The way it's currently described as just another string seems slightly hazardous since, extending the latter example in the docs, this doesn't work:

```
$ zed query -z 'from samples | County=="Orange" | load Orange@test author "Steve" meta "this is meta"'
cannot parse ZSON string "this is meta": ZSON syntax error
```

I'd never had reason to use the `-meta` parameter of `zed load` so I had to do a close read of [these docs](https://zed.brimdata.io/docs/commands/zed#28-load) to figure out I'd have to do something like these:

```
$ zed query -z 'from samples | County=="Orange" | load Orange@test author "Steve" meta "\"this is meta\""'
0x10d9f256a07d5bdad5d54cc80fc8771b88537954

$ zed query -z 'from samples | County=="Orange" | load Orange@test author "Steve" meta  "{MyMeta: \"This meta is in a record\"}"'
0x10d9f290a630a46fece73f7a7a7724a6d9669de5
```

Such that the metadata ends up in there as structured Zed data:

```
$ zed log -f zson
{
    Name: "test",
    Commit: 0x10d9f290a630a46fece73f7a7a7724a6d9669de5 (=ksuid.KSUID)
} (=lake.BranchTip)
{
    Name: "main",
    Commit: 0x0000000000000000000000000000000000000000 (=ksuid.KSUID)
} (=lake.BranchTip)
{
    id: 0x10d9f290a630a46fece73f7a7a7724a6d9669de5 (=KSUID),
    parent: 0x10d9f256a07d5bdad5d54cc80fc8771b88537954 (KSUID),
    retries: 0 (uint8),
    author: "Steve",
    date: 2023-04-28T21:54:24.808077Z,
    message: "loaded 1 data object\n\n  2P4aZd40YUfGhxayQkSyCQS3JtX 886 records in 47172 data bytes\n",
    meta: {
        MyMeta: "This meta is in a record"
    }
} (=Commit)
{
    id: 0x10d9f256a07d5bdad5d54cc80fc8771b88537954 (=KSUID),
    parent: 0x0000000000000000000000000000000000000000 (KSUID),
    retries: 0 (uint8),
    author: "Steve",
    date: 2023-04-28T21:53:26.069294Z,
    message: "loaded 1 data object\n\n  2P4aSPowRWirm3NziOsLJf59fql 886 records in 47172 data bytes\n",
    meta: "this is meta"
} (=Commit)
```

Since there's other places in the language where strings and complex literals can be written out without escaping I'd assume we'd want to do that rather than require this parameter to be a string, but perhaps there's a reason why that's easier said than done. If that's too difficult, then my next idea would be to show examples like these that show the necessary escaping so users don't get hung up on it like I did.

philrz:
Regarding that last comment from me, @nwt and I spoke offline and agreed we can deal with that in a separate PR after this one merges.

dianetc:
@nwt @philrz For the `meta` issue, I'm not sure how long it will take to get resolved (i.e., how pressing it is). Should I add a note about this in the docs for how the handle `meta` ,  in the meantime?

philrz:
@dianetc: I'm working on a branch with a draft of the `load` docs I'll submit for your consideration soon. Right now I've got a simple working example that uses `meta` and that will probably suffice for now.

philrz:
@dianetc: I've got a branch https://github.com/brimdata/zed/tree/load-op-docs that has proposed revisions to the docs. I've gone over them with @nwt and he's ok with them, so feel free to merge them to your branch if you're ok with them too. Here's a brief summary of why I proposed changing what I did:

1. I removed references to these being "scratch" pools since the changes in your PR seem to make the `load` operator functionally equivalent to what we have at the command line with `zed load`, i.e., it's just a general mechanism for loading data to an existing pool. At some point perhaps we'll add a way to create the destination pool if it doesn't already exist, and that's an option I'd describe as "scratch", but that can come later.

2. I'm using "mdtest" (https://github.com/brimdata/zed/blob/main/mdtest/mdtest.go) to protect the examples so that way the Zed commands the users see in the docs will run in CI and we'll be alerted if the Zed commands start failing or the output changes, since we'd want to update the docs in that case. I started out trying to use the same "schools" data as you'd started from in your own examples, but of a limitation  I hit in mdtest, so instead I just cooked up that ultra simple coin flip data like we've done at other docs pages.

3. The redirect of the output of the `load` commands to `/dev/null` is not ideal, but one of the limitations of mdtest is that things like unique hash outputs can cause the test to fail. We normally get around that by invoking the Zed commands with `-q` so they print no output, but this exercise led me to notice for the first time that `zed query -q` still displays query output. @nwt agreed that it's probably worth opening a bug about it but it's not urgent, so for now I recommend doing the `/dev/null` redirect and I can circle back and remove that when we `zed query -q` is capable of quietness.

Let me know if you've got any questions.

nwt (approved):

Issue #2839, opened 2021-06-30
change join syntax to use subquery


nwt:
A join using `from` currently looks like this.
```
from (
    left => pass;
    right => pass;
) | join on left_field == right_field
```

This might be more readable if the upstream operation provides the left input and a subquery provides the right input, like this.
```
from left | join (from right) on left_field == right_field
```

philrz:
The description text for #2836 includes a three-way `join` example I was hoping to show, but I could never come up with a workable syntax in the current implementation. When we implement this issue, I can validate it by expressing the example in the revised syntax and add it to the docs.

mccanne:
Steve and Noah agree that we will keep join on `=` to mean strict equi-join and in the future when we have generalized join expressions, join on `==` will mean coercible equi-join.

philrz:
The main [`join` operator docs page](https://zed.brimdata.io/docs/next/language/operators/join) now leads with the new syntax and the [join tutorial](https://zed.brimdata.io/docs/next/tutorials/join) shows several working examples that use the new syntax. The prior syntax is still available and shown in the docs as well.

Thanks @nwt!

PR #4473, opened 2023-03-31
Allow fork operator with from and non-from legs

nwt:
The fork operator requires that either all of its legs begin with a from operator or that none do.  Relax this restriction to allow programs like "fork (=> from (...) => pass)".

This is actually a side effect of the changes to
compiler/kernel.Builder.compileParallel that are necessary to implement subquery join syntax, which requires a dag.Parallel node containing one entry operation and one non-entry operation.

For #2839.


compiler/ztests/fork-from-and-pass.yaml
```diff
@@ -0,0 +1,13 @@
+zed: |
+  zq 'fork (=> from (file a.zson) => pass)' a.zson
```
mattnibs, line 2: Aren't we getting rid of pass maybe change to `yield this`?

mattnibs (approved):

compiler/ztests/fork-from-and-pass.yaml
nwt: I'd prefer to make the switch if and when we do drop pass.

PR #4474, opened 2023-03-31
Fix optimized filters on object and partition queries

### LINKED ISSUE #4472: Filters not working on partition, object meta queries (regression at 772322c)
I create a space and load data, thusly:

```bash
zed create test
zed use test
seq 20 | zq 'yield {ts:this}' - | zed load -
```

When I run a query on the objects metadata that should return no data

```bash
zed query 'from test@main:objects | max > 21'
```

I nevertheless get results:
```
{id:0x10b4c1b8fb68b535bdff358f7ad4803af26f86ce(=ksuid.KSUID),min:1,max:20,count:20(uint64),size:91}(=data.Object)
``` 

Note that if I put an operator between the from operator and the filter operator

```bash
zed query 'from del@main:objects | id := ksuid(id) | max > 20'
```

I get the expected results. This suggests that there's something wrong with optimized filters.

mattnibs:
Closes #4472

### DIFF
--- a/lake/ztests/meta-optimized-filter.yaml
+++ b/lake/ztests/meta-optimized-filter.yaml
@@ -1,0 +1,17 @@
+script: |
+  export ZED_LAKE=test
+  zed init -q
+  zed create -q test
+  zed use -q test
+  seq 20 | zq 'yield {ts: this}' - | zed load -q -
+  seq 21 40 | zq 'yield {ts: this}' - | zed load -q -
+  zed query -z 'from test:objects | max > 20 | cut min, max'
+  echo ===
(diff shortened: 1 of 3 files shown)


lake/ztests/meta-optimized-filter.yaml
```diff
@@ -0,0 +1,18 @@
...
+  echo ===
+  zed query -z 'from test:partitions | max > 20 | cut min, max'
+
+outputs:
+  - name: stdout
+    data: |
+      {min:21,max:40}
+      ===
+      {min:21,max:40}
+
```
nwt, line 18:
```suggestion
```

nwt (approved):

PR #4475, opened 2023-04-01
Adjust compiler/parser/Makefile for GNU cpp

nwt:
GNU cpp (on Ubuntu) needs -std=gnu90 to prevent errors on the \uHHHH sequences in compiler/parser/parser.peg, and Clang cpp (on macOS) with -std=gnu90 needs -Wno-unicode to prevent warnings on them.

@dianetc: This should fix the errors from "make -C compiler/parser" you've been seeing.


mattnibs (approved):
Lovely!

PR #4476, opened 2023-04-01
Remove runtime/expr.Sorter

### LINKED ISSUE #4121: zed load: +36% mem use, -15% load time (commit 7dfef14)
Through a set of iterative runs as described at https://github.com/brimdata/zed-perf-testing/tree/main/memory-load, I've observed that at Zed commit commit 7dfef14 (associated with #3482), the `zed load` of the `fdns-a.zeek.gz` data set saw an increase of 36% in reported `go_memstats_sys_bytes` in exchange for a 15% reduction in the time for the load to complete.

Here's a macro-level plot of the trend across many commits:

image

In a team discussion, @nwt was unsurprised that the perf improvement came with some memory cost. This issue has been opened to log the observation and consider if the trade is acceptable or if there's ways to potentially maintain the better perf while using less memory.

nwt:
In retrospect, this thing was a bad idea.  Removing it modestly reduces running time and peak memory footprint for both "zed load" and the sort operator on multi-GB inputs.

Closes #4121

### DIFF
--- a/lake/writer.go
+++ b/lake/writer.go
@@ -34,7 +34,6 @@ type Writer struct {
 	// data efficiently in one big backing store.
 	buffer      chan []zed.Value
 	comparator  *expr.Comparator
-	sorter      expr.Sorter
 	memBuffered int64
 	stats       ImportStats
 }
@@ -128,7 +127,7 @@ func (w *Writer) writeObject(object *data.Object, recs []zed.Value) error {
 	if !w.inputSorted {
 		done := make(chan struct{})
 		go func() {
-			w.sorter.SortStable(recs, w.comparator)
+			w.comparator.SortStable(recs)
 			close(done)
 		}()
 		select {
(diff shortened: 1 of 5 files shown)


philrz:
@nwt: Hmm! That would imply this issue will address the changes reported in #4121. Those findings indicated that upon adding `expr.Sorter` run time _improved_, but came with the increase in memory usage as you say. Here you indicate that _removing_ it may improve run time. Well, I'll see what the Autoperf trends have to say when this merges. 👍

philrz:
Looking good! I used this as an excuse to add a new load/count test pair to Autoperf using the FDNS data since that's what I'd used in #4121 and I know that the differences in that test data vs. wrccdc can have observable differences in the results (FDNS has smaller records and an `ip` pool key). On these charts the left point is the baseline at Zed commit 36dd506 and the right is at commit 832c134 on this branch. The reduction in sys memory usage is observable with both the wrccdc and FDNS data set and the run times seem about the same.

image

image

mattnibs (approved):
Nice

PR #4484, opened 2023-04-04
Add runtime/expr.Comparator.SortStableReader

nwt:
SortStableReader returns a zio.Reader over values in sorted order. Because it allocates less memory than SortStable, using it in place of SortStable modestly reduces running time and peak memory footprint for both "zed load" and the sort operator on multi-GB inputs.

### DIFF
--- a/lake/writer.go
+++ b/lake/writer.go
@@ -124,10 +124,13 @@ func (w *Writer) Close() error {
 }
 
 func (w *Writer) writeObject(object *data.Object, recs []zed.Value) error {
-	if !w.inputSorted {
+	var zr zio.Reader
+	if w.inputSorted {
+		zr = zbuf.NewArray(recs)
+	} else {
 		done := make(chan struct{})
 		go func() {
-			w.comparator.SortStable(recs)
+			zr = w.comparator.SortStableReader(recs)
 			close(done)
 		}()
 		select {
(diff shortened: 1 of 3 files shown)


lake/writer.go
```diff
@@ -124,10 +124,13 @@ func (w *Writer) Close() error {
...
 func (w *Writer) writeObject(object *data.Object, recs []zed.Value) error {
-	if !w.inputSorted {
+	var zr zio.Reader
+	if w.inputSorted {
+		zr = zbuf.NewArray(recs)
+	} else {
 		done := make(chan struct{})
 		go func() {
-			w.comparator.SortStable(recs)
+			zr = w.comparator.SortStableReader(recs)
```
mattnibs, line 133: Just a though, and not to be addressed in this pr, but perhaps we should add a context to the comparator Sort* methods so we don't have to do things like this?

nwt: You might be on to something here.

philrz:
Now that this has merged, I can see in the Autoperf trend for loading the full wrccdc data set that this change does indeed provide a modest but statistically significant improvement in the numbers. The rightmost point in these plots are at commit 729f9c1 which is associated with the changes in this PR. Compared to the runs at the prior commit 30e7c5a, the load time improved from 12.6 minutes to 12.3 minutes (so, a 2.3% speed-up) and the average sys memory usage dropped from 5.56 GB to 4.91 GB (so, an 11.6% improvement).

image

PR #4511, opened 2023-04-20
Shorten canonical DAG kernel.Reader representation

nwt:


mccanne (approved):
I have this as "reader" not "<reader>" in the tip of my branch...

PR #4512, opened 2023-04-20
Describe both request & response MIME types for lake API

philrz:
Up until now, the [Media Types](https://zed.brimdata.io/docs/lake/api#media-types) section at the bottom of the lake API doc only spoke of response formats. However, once #4345 and #4235 merged, we had symmetry with the API such that most formats were supported for both requests and responses. That made me want to make these changes so the docs would be more symmetrical to match that and also divulge which formats could be used in only one direction.

### DIFF
--- a/docs/lake/api.md
+++ b/docs/lake/api.md
@@ -30,6 +30,8 @@ POST /pool
 | layout.order | string | body | Order of storage by primary key(s) in pool. Possible values: desc, asc. Default: asc. |
 | layout.keys | [[string]] | body | Primary key(s) of pool. The element of each inner string array should reflect the hierarchical ordering of named fields within indexed records. Default: [[ts]]. |
 | thresh | int | body | The size in bytes of each seek index. |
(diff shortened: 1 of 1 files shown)


docs/lake/api.md
```diff
@@ -85,12 +87,12 @@ PUT /pool/{pool}
 | ---- | ---- | -- | ----------- |
 | pool | string | path | **Required.** ID or name of the requested pool. |
 | name | string | body | **Required.** The desired new name of the pool. Must be unique to lake. |
+| Content-Type | string | header | [MIME type](#mime-types) of the request payload. |
 
 **Example Request**
 
 ```
 curl -X PUT \
-     -H 'Accept: application/json' \
```
philrz, line 93: This was one I spotted along the way: Since the successful HTTP 204 for this endpoint has no response payload and errors would come back only as JSON, it seems like this was unnecessary.

mattnibs (approved):

PR #4526, opened 2023-04-21
move metadata ops into the DAG and refactor optimizer

mccanne:
This commit makes a big shift in the optimizer to support the forthcoming vector cache and engine.  With these changes, the listing and manipulation of scanned objects is now part of the DAG instead of being constructed at the last phase of runtime generation.  This means we can implement DAG transormations that replace or combine in vector operations with the traditional sequence scanner.  This also will set us up for distributed computation as the lister can transmit objects over the network to different pieces of the distributed DAG.

We also adapted the "delete where" logic to use the refactored approach.

These changes improve performance when a scan is not order sensitive as we now separate the slicing of objects into partitions from the listing of raw objects.  SeqScanner and Deleter both handle reading sorted and non-sorted cases by creating a merge-on-read scanner when order is required for overlapping objects or simply scanning one object at a time when order is not needed.


compiler/ast/dag/op.go
```diff
@@ -137,26 +141,25 @@ type (
 // Input structure
 
 type (
-	From struct {
-		Kind   string  `json:"kind" unpack:""`
-		Trunks []Trunk `json:"trunks"`
+	Lister struct {
+		Kind      string      `json:"kind" unpack:""`
+		Pool      ksuid.KSUID `json:"id"`
```
nwt, line 146:
```suggestion
		Pool      ksuid.KSUID `json:"pool"`
```

zfmt/dag.go
```diff
@@ -429,32 +429,35 @@ func (c *canonDAG) op(p dag.Op) {
...
-			c.write("%s", source(trunk.Source))
-			if trunk.Seq != nil && len(trunk.Seq.Ops) != 0 {
-				c.open()
-				c.head = true
-				c.write(" =>")
-				c.op(trunk.Seq)
-				c.close()
-			}
+		c.open("lister")
+		c.write(fmt.Sprintf(" pool %s commit %s", p.Pool, p.Commit))
```
nwt, line 435:
```suggestion
		c.write(" pool %s commit %s", p.Pool, p.Commit)
```

zfmt/dag.go
```diff
@@ -429,32 +429,35 @@ func (c *canonDAG) op(p dag.Op) {
...
+			c.expr(p.KeyPruner, "")
+			c.write(")")
 		}
-		c.ret()
 		c.close()
-		c.write(")")
+	case *dag.SeqScan:
+		c.next()
+		c.open("seqscan")
+		c.write(fmt.Sprintf(" pool %s", p.Pool))
```
nwt, line 445:
```suggestion
		c.write(" pool %s", p.Pool)
```

zfmt/dag.go
nwt, line 435: c.write calls fmt.Sprintf for you.
```suggestion
		c.write(" pool %s commit %s", p.Pool, p.Commit)
```

compiler/kernel/op.go
```diff
@@ -325,7 +416,21 @@ func (b *Builder) compileSequential(seq *dag.Sequential, parents []zbuf.Puller)
...
 
-func (b *Builder) compileParallel(par *dag.Parallel, parents []zbuf.Puller) ([]zbuf.Puller, error) {
+func (b *Builder) compileParallel(parallel *dag.Parallel, parents []zbuf.Puller) ([]zbuf.Puller, error) {
+	if parallel.Any {
+		if len(parents) != 1 {
+			return nil, errors.New("internal compiler error: any-path parallel dag operators requires a single parent")
+		}
+		var ops []zbuf.Puller
+		for _, o := range parallel.Ops {
+			op, err := b.compile(o, []zbuf.Puller{parents[0]})
```
nwt, line 426: Nit: A slice expression would be simpler.
```suggestion
			op, err := b.compile(o, parents[:1])
```

compiler/kernel/op.go
```diff
@@ -325,7 +416,21 @@ func (b *Builder) compileSequential(seq *dag.Sequential, parents []zbuf.Puller)
 	return parents, nil
 }
 
-func (b *Builder) compileParallel(par *dag.Parallel, parents []zbuf.Puller) ([]zbuf.Puller, error) {
+func (b *Builder) compileParallel(parallel *dag.Parallel, parents []zbuf.Puller) ([]zbuf.Puller, error) {
```
nwt, line 419: Nit: Stick with par since that's what we call a *dag.Parallel everywhere else.
```suggestion
func (b *Builder) compileParallel(par *dag.Parallel, parents []zbuf.Puller) ([]zbuf.Puller, error) {
```

zfmt/dag.go
mccanne: I fixed all the others that were in there.

compiler/ast/dag/op.go
```diff
@@ -178,19 +182,19 @@ type (
...
 	}
 	CommitMetaScan struct {
-		Kind   string      `json:"kind" unpack:""`
-		Pool   ksuid.KSUID `json:"pool"`
-		Commit ksuid.KSUID `json:"branch"`
-		Meta   string      `json:"meta"`
-		Tap    bool        `json:"tap"`
+		Kind      string      `json:"kind" unpack:""`
+		Pool      ksuid.KSUID `json:"pool"`
+		Commit    ksuid.KSUID `json:"branch"`
```
nwt, line 194:
```suggestion
		Commit    ksuid.KSUID `json:"commit"`
```

runtime/op/meta/sequence.go
```diff
@@ -48,42 +48,48 @@ func (s *SequenceScanner) Pull(done bool) (zbuf.Batch, error) {
...
-			s.current, err = newSortedPartitionScanner(s, part)
+			vals := batch.Values()
+			if len(vals) != 1 {
+				if len(vals) != 1 {
+					// We currently support only one partition per batch.
+					err := errors.New("system error: SequenceScanner encountered multi-valued batch")
+					s.close(err)
+					return nil, err
+				}
+			}
```
nwt, line 77:
```suggestion
			if len(vals) != 1 {
				// We currently support only one partition per batch.
				err := errors.New("system error: SequenceScanner encountered multi-valued batch")
				s.close(err)
				return nil, err
			}
```

runtime/op/meta/sequence.go
```diff
@@ -92,28 +98,38 @@ func (s *SequenceScanner) close(err error) {
...
+	pullers := make([]zbuf.Puller, 0, len(objects))
 	pullersDone := func() {
 		for _, puller := range pullers {
 			puller.Pull(true)
 		}
 	}
-	for _, object := range part.Objects {
-		s, err := newObjectScanner(p.octx.Context, p.octx.Zctx, p.pool, object, p.filter, p.pruner, p.progress)
+	for _, object := range objects {
+		s, err := newStatScanner(ctx, zctx, pool, object, pruner, filter, progress)
```
nwt, line 132: And zap newStatScanner.
```suggestion
		s, err := newObjectScanner(ctx, zctx, pool, object, filter, pruner, progress)
```

compiler/optimizer/op.go
```diff
@@ -2,121 +2,120 @@ package optimizer
...
 	"github.com/brimdata/zed/order"
 	"github.com/brimdata/zed/pkg/field"
 )
 
-// analyzeOp returns how an input order maps to an output order based
+// analyzeSortKey returns how an input order maps to an output order based
 // on the semantics of the operator.  Note that an order can go from unknown
-// to known (e.g., sort) or from known to unknown (e.g., parallel paths).
+// to known (e.g., sort) or from known to unknown (e.g., conflicting parallel paths).
 // Also, when op is a Summarize operator, it's input direction (where the
```
nwt, line 16: While you're here.
```suggestion
// Also, when op is a Summarize operator, its input direction (where the
```

compiler/optimizer/op.go
```diff
@@ -2,121 +2,120 @@ package optimizer
...
 // order key is presumed to be the primary group-by key) is set based
-// on the sortKey argument.  This is clumsy and needs to change.
+// on the in sort key.  This is clumsy and needs to change.
 // See issue #2658.
-func (o *Optimizer) analyzeOp(op dag.Op, sortKey order.SortKey) (order.SortKey, error) {
+func (o *Optimizer) analyzeSortKey(op dag.Op, in order.SortKey) (order.SortKey, error) {
+	if scan, ok := op.(*dag.PoolScan); ok {
+		// Ignore the null in value and just return the sort order of the pool.
```
nwt, line 22:
```suggestion
		// Ignore in and just return the sort order of the pool.
```

compiler/optimizer/op.go
```diff
@@ -220,15 +220,15 @@ func fieldOf(e dag.Expr) field.Path {
 	return nil
 }
 
-func copyOps(ops []dag.Op) []dag.Op {
+func CopyOps(ops []dag.Op) []dag.Op {
```
nwt, line 223: Nit: This isn't used in any other package.
```suggestion
func copyOps(ops []dag.Op) []dag.Op {
```

compiler/optimizer/op.go
```diff
@@ -220,15 +220,15 @@ func fieldOf(e dag.Expr) field.Path {
...
 	var copies []dag.Op
 	for _, o := range ops {
-		copies = append(copies, copyOp(o))
+		copies = append(copies, CopyOp(o))
 	}
 	return copies
 }
 
-func copyOp(o dag.Op) dag.Op {
+func CopyOp(o dag.Op) dag.Op {
```
nwt, line 231: Nit: This isn't used in any other package.
```suggestion
func copyOp(o dag.Op) dag.Op {
```

runtime/op/meta/deleter.go
```diff
@@ -47,42 +46,82 @@ func (d *Deleter) Pull(done bool) (zbuf.Batch, error) {
...
 				return nil, err
 			}
-			d.current, err = newDeleterScanner(d, part)
+			vals := batch.Values()
+			if len(vals) != 1 {
+				if len(vals) != 1 {
+					// We currently support only one partition per batch.
+					return nil, errors.New("system error: meta.Deleter encountered multi-valued batch")
+				}
+			}
```
nwt, line 97:
```suggestion
			if len(vals) != 1 {
				// We currently support only one partition per batch.
				return nil, errors.New("system error: meta.Deleter encountered multi-valued batch")
			}
```

compiler/semantic/op.go
```diff
@@ -20,22 +20,29 @@ import (
...
-func semFrom(ctx context.Context, scope *Scope, from *ast.From, source *data.Source, head *lakeparse.Commitish) (*dag.From, error) {
-	var trunks []dag.Trunk
-	for _, in := range from.Trunks {
-		converted, err := semTrunk(ctx, scope, in, source, head)
-		if err != nil {
-			return nil, err
+func semFrom(ctx context.Context, scope *Scope, from *ast.From, source *data.Source, head *lakeparse.Commitish) (dag.Op, error) {
+	switch len(from.Trunks) {
+	case 0:
+		return nil, errors.New("from command has no paths")
```
nwt, line 26:
```suggestion
		return nil, errors.New("from operator has no paths")
```

compiler/semantic/op.go
```diff
@@ -97,10 +111,11 @@ func semSource(ctx context.Context, scope *Scope, source ast.Source, ds *data.So
 		}
 		return semPool(ctx, scope, p, ds, head)
 	case *ast.Pass:
-		return []dag.Source{&dag.Pass{Kind: "Pass"}}, nil
+		//XXX just connect parent
+		return []dag.Op{&dag.Pass{Kind: "Pass"}}, nil
```
nwt, line 115:
```suggestion
		return []dag.Op{dag.PassOp, nil
```

compiler/semantic/op.go
```diff
@@ -239,12 +254,28 @@ func semPoolWithName(ctx context.Context, scope *Scope, p *ast.Pool, poolName st
...
+			},
+		}, nil
+	}
 	return &dag.PoolScan{
 		Kind:   "PoolScan",
 		ID:     poolID,
 		Commit: commitID,
-		Delete: p.Delete,
 	}, nil
+
```
nwt, line 278:
```suggestion
```

compiler/ztests/par-count.yaml
```diff
@@ -2,32 +2,40 @@ script: |
   export ZED_LAKE=test
   zed init -q
   zed create -q -orderby ts test
-  zc -C -P 2 "from test | count() by y" | sed -e 's/pool .* =>/pool POOL =>/'
+  zc -C -P 2 "from test | count() by y" | sed -e 's/lister pool.*/lister .../' -e 's/seqscan pool .*/seqscan .../'
```
nwt, line 5: Nit: Simplify all these sed commands.
```suggestion
  zc -C -P 2 "from test | count() by y" | sed -e 's/pool .*$/.../'
```

compiler/kernel/op.go
```diff
@@ -244,6 +247,94 @@ func (b *Builder) compileLeaf(o dag.Op, parent zbuf.Puller) (zbuf.Puller, error)
 			return nil, err
 		}
 		return b.compileOver(parent, v.Over, names, exprs)
+	case *dag.PoolScan:
+		if parent != nil {
+			return nil, errors.New("a pool scan cannot have a parent")
```
nwt, line 252: Nit: Make this consistent with the dag.Lister error.
```suggestion
			return nil, errors.New("pool scan cannot have a parent operator")
```

compiler/kernel/op.go
```diff
@@ -514,177 +612,36 @@ func (b *Builder) compile(o dag.Op, parents []zbuf.Puller) ([]zbuf.Puller, error
...
-}
-
-func (b *Builder) compileTrunk(trunk *dag.Trunk, parent zbuf.Puller) ([]zbuf.Puller, error) {
-	pushdown, err := b.PushdownOf(trunk)
+func (b *Builder) compilePoolScan(scan *dag.PoolScan) (zbuf.Puller, error) {
+	// Here we convert PoolScan to lister->slicer->seqscan for the slow path as
+	// optimizer should do this conversion, but this allows us to run
+	// unoptimized scans too.
+	lk := b.source.Lake()
+	pool, err := lk.OpenPool(b.octx.Context, scan.ID)
```
nwt, line 620:
```suggestion
	pool, err := b.lookupPool(scan.ID)
```

compiler/kernel/op.go
```diff
@@ -326,6 +417,20 @@ func (b *Builder) compileSequential(seq *dag.Sequential, parents []zbuf.Puller)
 }
 
 func (b *Builder) compileParallel(par *dag.Parallel, parents []zbuf.Puller) ([]zbuf.Puller, error) {
+	if par.Any {
+		if len(parents) != 1 {
+			return nil, errors.New("internal compiler error: any-path parallel dag operators requires a single parent")
```
nwt, line 422:
```suggestion
			return nil, errors.New("internal error: any-path parallel operator requires a single parent")
```

compiler/kernel/op.go
```diff
@@ -716,3 +673,27 @@ func EvalAtCompileTime(zctx *zed.Context, in dag.Expr) (val *zed.Value, err erro
...
+		if len(op.Ops) == 0 {
+			return false
+		}
+		for _, o := range op.Ops {
+			if s, ok := o.(*dag.Sequential); !ok || !isEntry(s) {
+				return false
+			}
+		}
+		return true
+
```
nwt, line 696:
```suggestion
```

compiler/optimizer/op.go
```diff
@@ -2,121 +2,120 @@ package optimizer
...
+		return order.Nil
+	}
+	return order.NewSortKey(o, field.List{key})
+}
+
+// isKeyOfSummarize returns true iff its first groupby key is the
+// same as the given sort order or an order-preserving function thereof.
+func isKeyOfSummarize(summarize *dag.Summarize, in order.SortKey) bool {
+	key := in.Keys[0]
+	for _, outputKeyExpr := range summarize.Keys {
```
nwt, line 108: My understanding is that we're only interested in whether or not the first summarize key (`summarize.Keys[0]`) matches `in`, so why examine anything past that first key?

compiler/optimizer/op.go
```diff
@@ -150,6 +149,7 @@ func analyzeCuts(assignments []dag.Assignment, sortKey order.SortKey) order.Sort
 	// sort key we return.  In a future version of the optimizer, we will
 	// generalize this scoreboard concept across the flowgraph for a
 	// comprehensive approach to dataflow analysis.  See issue #2756.
+	//XXX 2756
```
nwt, line 152: Nit: The prior line already refers to 2756.
```suggestion
```

compiler/optimizer/op.go
mccanne: The comment is wrong.  I'll fix it.  We want to know if any group-by key is corresponds to sort order so we can release group-by results as soon as its known that there won't be anymore for a given value.

runtime/op/meta/deleter.go
```diff
@@ -47,42 +46,80 @@ func (d *Deleter) Pull(done bool) (zbuf.Batch, error) {
...
+			// not a partition.
+			batch, err := d.parent.Pull(false)
+			if batch == nil || err != nil {
 				return nil, err
 			}
-			d.current, err = newDeleterScanner(d, part)
+			vals := batch.Values()
+			if len(vals) != 1 {
+				// We currently support only one partition per batch.
+				return nil, errors.New("system error: meta.Deleter encountered multi-valued batch")
```
nwt, line 94:
```suggestion
				return nil, errors.New("internal error: meta.Deleter encountered multi-valued batch")
```

compiler/kernel/op.go
```diff
@@ -514,177 +612,35 @@ func (b *Builder) compile(o dag.Op, parents []zbuf.Puller) ([]zbuf.Puller, error
...
 	}
-	f, ok := trunk.Pushdown.(*dag.Filter)
-	if !ok {
-		return nil, errors.New("non-filter pushdown operator not yet supported")
+	return &Filter{e, b}
+}
+
+func (b *Builder) lookupPool(id ksuid.KSUID) (*lake.Pool, error) {
+	if b.source == nil || b.source.Lake() == nil {
+		return nil, errors.New("system error: lake operation cannot be used in non-lake context")
```
nwt, line 640:
```suggestion
		return nil, errors.New("internal error: lake operation cannot be used in non-lake context")
```

compiler/ztests/pushdown.yaml
```diff
@@ -2,53 +2,41 @@ script: |
   export ZED_LAKE=test
   zed init -q
   zed create -q -orderby ts pool-ts
-  zc -C -O "from 'pool-ts'| x=='hello' or x==1.0" | sed -e 's/pool .*/pool POOL/'
+  zc -C -O "from 'pool-ts' | x=='hello' or x==1.0" | sed -e 's/lister .*/lister/' -e 's/seqscan .*filter/seqscan filter/'
```
nwt, line 5: Nit: Simplify the sed commands in here.
```suggestion
  zc -C -O "from 'pool-ts' | x=='hello' or x==1.0" | sed -e 's/pool .*/.../'
```

compiler/semantic/op.go
```diff
@@ -20,22 +20,29 @@ import (
...
-func semFrom(ctx context.Context, scope *Scope, from *ast.From, source *data.Source, head *lakeparse.Commitish) (*dag.From, error) {
-	var trunks []dag.Trunk
-	for _, in := range from.Trunks {
-		converted, err := semTrunk(ctx, scope, in, source, head)
-		if err != nil {
-			return nil, err
+func semFrom(ctx context.Context, scope *Scope, from *ast.From, source *data.Source, head *lakeparse.Commitish) (dag.Op, error) {
+	switch len(from.Trunks) {
+	case 0:
+		return nil, errors.New("from operator has no paths")
```
nwt, line 26: This is an internal error, right?
```suggestion
		return nil, errors.New("internal error: from operator has no paths")
```

compiler/optimizer/optimizer.go
```diff
@@ -2,29 +2,36 @@ package optimizer
...
 )
 
 type Optimizer struct {
-	ctx      context.Context
-	entry    *dag.Sequential
-	source   *data.Source
-	sortKeys map[dag.Source]order.SortKey
+	ctx     context.Context
+	entry   *dag.Sequential
+	sources map[dag.Op]struct{}
```
nwt, line 21: Can this be a map[*dag.Sequential]struct{}?

compiler/optimizer/optimizer.go
```diff
@@ -55,103 +62,219 @@ func (o *Optimizer) mergeFilters() {
...
-		for _, t := range op.Trunks {
-			if t.Seq != nil {
-				walkOp(t.Seq, post)
+func removePassOps(op dag.Op) {
+	walkOp(op, true, func(op dag.Op) {
+		if seq, ok := op.(*dag.Sequential); ok {
+			// Start at the next to last element and work toward the first.
+			for i := 0; i < len(seq.Ops); i++ {
+				if _, ok := seq.Ops[i].(*dag.Pass); ok {
+					seq.Ops = slices.Delete(seq.Ops, i, i+1)
```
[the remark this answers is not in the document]
mccanne: Doesn't the loop still increment?

compiler/ztests/pushdown.yaml
mccanne: That clobbers the seqscan filter predicate under test.

compiler/optimizer/optimizer.go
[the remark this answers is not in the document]
nwt: It sure does. 🤦

Maybe this instead?
```suggestion
					seq.Ops = slices.Delete(seq.Ops, i, i+1)
					i--
```

compiler/optimizer/optimizer.go
```diff
@@ -209,137 +321,123 @@ func (o *Optimizer) propagateScanOrder(op dag.Op, parent order.SortKey) (order.S
...
+	return lister, slicer, ops
 }
 
-func extend(trunk *dag.Trunk, op dag.Op) {
-	if trunk.Seq == nil {
-		trunk.Seq = &dag.Sequential{Kind: "Sequential"}
+// filterPushdown attempts to move any filter from the front of op chain
+// and returns the filter's expression (and the modified chain) so that
+// the runtime can push the filter predicate into the scanner.
+func filterPushdown(in []dag.Op) (dag.Expr, []dag.Op) {
```
nwt, line 432: Nit: Give this a verby name. Maybe extractLeadingFilter?

compiler/optimizer/optimizer.go
```diff
@@ -209,137 +321,123 @@ func (o *Optimizer) propagateScanOrder(op dag.Op, parent order.SortKey) (order.S
...
 	default:
-		return order.Nil, fmt.Errorf("unknown dag.Source type %T", s)
+		return order.Nil, fmt.Errorf("internal error: unknown source type %T", op)
 	}
 }
 
-// Parallelize takes a sequential operation and tries to
-// parallelize it by splitting as much as possible of the sequence
-// into n parallel branches. The boolean return argument indicates
-// whether the flowgraph could be parallelized.
```
nwt, line 260: Nit: I think it's worth the effort to keep a doc comment on Parallelize.

compiler/optimizer/parallelize.go
```diff
@@ -2,189 +2,211 @@ package optimizer
...
-// possible insert an operator at seq.Op[1]
-func (o *Optimizer) parallelizeTrunk(seq *dag.Sequential, trunk *dag.Trunk, replicas int) error {
-	from, ok := seq.Ops[0].(*dag.From)
-	if !ok {
-		return errors.New("internal error: parallelizeTrunk: entry is not a From")
+func (o *Optimizer) parallelizeScan(ops []dag.Op, replicas int) ([]dag.Op, error) {
+	// For now we parallelize only pool scans and no metadata scans.
+	// We can do the latter when we want to scale the performance of metadata.
+	if replicas < 2 {
+		panic(fmt.Sprintf("parallelizeScan: bad replicas: %d", replicas))
```
nwt, line 24:
```suggestion
		return nil, fmt.Errorf("internal error: parallelizeScan: bad replicas: %d", replicas)
```

compiler/optimizer/parallelize.go
```diff
@@ -2,189 +2,211 @@ package optimizer
...
 	}
-	if len(from.Trunks) > 1 {
-		// No support for multi-trunk from's yet, which only arise for
-		// joins and other peculaliar mixins of different sources.
-		// We need to handle join parallelization differently, though
-		// the logic is not very far from what's here.
-		return nil
+	if len(ops) < 1 {
+		panic("parallelizeScan: short path")
```
nwt, line 27:
```suggestion
		return nil, errors.New("internal error: parallelizeScan: short path")
```

compiler/optimizer/parallelize.go
```diff
@@ -2,189 +2,211 @@ package optimizer
...
-	// as checked above.
-	sortKey, err := o.sortKeyOfSource(trunk.Source, order.Nil)
-	if err != nil {
-		return err
+	if len(ops) == 1 && scan.Filter == nil {
+		return nil, nil
 	}
-	inlineSequentials(seq)
-	// Check that the path consisting of the original from
+	// concurrentPath will check that the path consisting of the original from
```
nwt, line 44: Say "source" here since we no longer have dag.From?
```suggestion
	// concurrentPath will check that the path consisting of the original source
```

compiler/optimizer/optimizer.go
mccanne: I'll call it `matchFilter`.  As we do more pattern matching in the optimizer, I think the `match*` pattern makes sense.

compiler/optimizer/parallelize.go
```diff
@@ -197,66 +219,6 @@ func inlineSequentials(op dag.Op) {
...
-			Source:    src,
-			Seq:       newSeq,
-			Pushdown:  pd,
-			KeyPruner: trunk.KeyPruner,
-		}
-		from.Trunks = append(from.Trunks, replica)
-	}
-}
-
 func canOptimizeEvery(s *dag.Summarize) bool {
```
nwt, line 222: Delete this since it's now unused.

compiler/optimizer/parallelize.go
```diff
@@ -2,189 +2,211 @@ package optimizer
...
+	case *dag.Cut, *dag.Drop, *dag.Put, *dag.Rename, *dag.Filter:
+		if merge != nil {
+			// See if drop, put, or rename would disrupt the merge-on key
+			mergeKey, err := o.propagateSortKey(merge, order.Nil)
+			if err != nil || mergeKey.IsNil() {
+				// Bail if there's a merge with a non-key expression.
+				return
+			}
+			key, err := o.propagateSortKey(op, mergeKey)
+			if !key.Equal(mergeKey) {
```
nwt, line 190:
```suggestion
			if err != nil || !key.Equal(mergeKey) {
```

compiler/optimizer/parallelize.go
```diff
@@ -2,189 +2,211 @@ package optimizer
...
-		if sortKey.IsNil() {
-			// Unknown order: we can't parallelize because
-			// we can't maintain this unknown order at the merge point,
-			// but we shouldn't care if the client doesn't care
-			// so this goes back to whether the scan order was
-			// specified in the source. See Issue #2661.
-			return nil
+	case *dag.Cut, *dag.Drop, *dag.Put, *dag.Rename, *dag.Filter:
+		if merge != nil {
+			// See if drop, put, or rename would disrupt the merge-on key
```
nwt, line 183:
```suggestion
			// See if this op would disrupt the merge-on key
```

compiler/optimizer/optimizer.go
```diff
@@ -55,103 +62,219 @@ func (o *Optimizer) mergeFilters() {
...
-func walkOp(op dag.Op, post func(dag.Op)) {
-	switch op := op.(type) {
-	case *dag.From:
-		for _, t := range op.Trunks {
-			if t.Seq != nil {
-				walkOp(t.Seq, post)
+func removePassOps(op dag.Op) {
+	walkOp(op, true, func(op dag.Op) {
+		if seq, ok := op.(*dag.Sequential); ok {
+			// Start at the next to last element and work toward the first.
```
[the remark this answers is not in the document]
mccanne: It seems correct to me though the comment is wrong.  I edited the comment and added a test.  The test exposed the fact that an intervening `pass` confounded `mergeFilters` so I added an extra `removePassOps`.  In later PR, we can use a loop-until-no-changes approach in the optimizer (standard technique in PL optimization).

compiler/optimizer/optimizer.go
```diff
@@ -209,137 +321,125 @@ func (o *Optimizer) propagateScanOrder(op dag.Op, parent order.SortKey) (order.S
...
-	if trunk.Seq == nil {
-		return
-	}
-	for _, op := range trunk.Seq.Ops {
-		if cut, ok := op.(*dag.Cut); ok {
-			cut.Quiet = true
-		}
+func (o *Optimizer) lookupPool(id ksuid.KSUID) (*lake.Pool, error) {
+	if o.lake == nil {
+		return nil, errors.New("system error: lake operation cannot be used in non-lake context")
```
nwt, line 412: This is an internal error, right?
```suggestion
		return nil, errors.New("internal error: lake operation cannot be used in non-lake context")
```

compiler/optimizer/parallelize.go
```diff
@@ -2,189 +2,211 @@ package optimizer
...
+		return nil, nil
+	}
+	head := ops[:n+1]
+	tail := ops[n+1:]
+	par := &dag.Parallel{Kind: "Parallel", Any: true}
+	for k := 0; k < replicas; k++ {
+		par.Ops = append(par.Ops, &dag.Sequential{Kind: "Sequential", Ops: copyOps(head)})
+	}
+	var merge dag.Op
+	if !needMerge {
```
nwt, line 61: Nit: I think this would be a little clearer if you reversed the sense here (and swapped the "if" and "else" blocks).
```suggestion
	if needMerge {
```

compiler/optimizer/op.go
nwt: But runtime/op/groupby/groupby.go only looks at the first summarize key (via Aggregator.keyCompare) to make that decision. Given that, is there any reason to look beyond the first key here?

compiler/ztests/pushdown.yaml
nwt: How about this?
```suggestion
  zc -C -O "from 'pool-ts' | x=='hello' or x==1.0" | sed -e 's/pool [^ ]*/.../'
```

compiler/optimizer/parallelize.go
```diff
@@ -2,189 +2,211 @@ package optimizer
...
+		return nil, nil
 	}
-	// This logic requires that there is only one trunk in the From,
-	// as checked above.
-	sortKey, err := o.sortKeyOfSource(trunk.Source, order.Nil)
-	if err != nil {
-		return err
+	if len(ops) == 1 && scan.Filter == nil {
+		return nil, nil
 	}
```
nwt, line 43: Nit: Move this ahead of the call to o.sortKeyofSource.

compiler/optimizer/op.go
```diff
@@ -2,121 +2,121 @@ package optimizer
...
 	// See issue #2657.
-	key := sortKey.Primary()
+	key := in.Primary()
 	if key == nil {
 		return order.Nil, nil
 	}
 	switch op := op.(type) {
+	case *dag.Lister:
+		// This shouldn't happen.
+		return order.Nil, errors.New("system error: dag.Lister encountered in anaylzeSortKey")
```
nwt, line 38:
```suggestion
		return order.Nil, errors.New("internal error: dag.Lister encountered in anaylzeSortKey")
```

compiler/optimizer/parallelize.go
```diff
@@ -2,189 +2,211 @@ package optimizer
...
 		}
-		replicateTrunk(from, trunk, replicas)
-		return nil
+	})
+}
+
+// parPullUp examines a sequence of Ops to see if there's an opportunity to
+// lift operations downstream from a parallel Op into the parallel paths to
+// enhance concurrency.  If so, we modify the sequential ops in place.
+func (o *Optimizer) parPullUp(ops []dag.Op) {
```
nwt, line 91: Maybe call this liftParPullups.
```suggestion
// liftParPullups examines a sequence of Ops to see if there's an opportunity to
// lift operations downstream from a parallel Op into the parallel paths to
// enhance concurrency.  If so, we modify the sequential ops in place.
func (o *Optimizer) liftParPullups(ops []dag.Op) {
```

mccanne: I landed on `optimizeParallels` and `liftIntoParPaths`

compiler/optimizer/parallelize.go
```diff
@@ -2,189 +2,211 @@ package optimizer
...
 		// for partials to flow between them.
-		egress := copyOp(ingress).(*dag.Summarize)
-		egress.PartialsOut = true
-		ingress.PartialsIn = true
+		if op.PartialsIn || op.PartialsOut {
+			// Need an unmodified summarize to split into its parials pieces.
+			return
+		}
+		for k := 0; k < len(par.Ops); k++ {
+			path := par.Ops[k].(*dag.Sequential)
```
nwt, line 123:
```suggestion
		for _, o := range par.Ops {
			path := o.(*dag.Sequential)
```

compiler/optimizer/parallelize.go
```diff
@@ -2,189 +2,211 @@ package optimizer
...
-		replicateTrunk(from, trunk, replicas)
-		return nil
-	case *dag.Sort:
-		if len(ingress.Args) > 1 {
-			// Unknown or multiple sort fields: we sort after
-			// the merge, which can be unordered.
-			replicateTrunk(from, trunk, replicas)
-			return nil
+		for k := 0; k < len(par.Ops); k++ {
+			path := par.Ops[k].(*dag.Sequential)
```
nwt, line 153:
```suggestion
		for _, o := range par.Ops {
			path := o.(*dag.Sequential)
```

compiler/optimizer/parallelize.go
```diff
@@ -2,189 +2,211 @@ package optimizer
...
-		sortKey := order.NewSortKey(ingress.Order, field.List{mergeKey})
-		return replicateAndMerge(seq, sortKey, from, trunk, replicas)
 	case *dag.Head, *dag.Tail:
-		if sortKey.IsNil() {
-			// Unknown order: we can't parallelize because we can't maintain this unknown order at the merge point.
-			return nil
+		// Copy the head or tail into the parallel path and leave the original in
+		// place which will apply another head or tail after the merge.
+		for k := 0; k < len(par.Ops); k++ {
+			path := par.Ops[k].(*dag.Sequential)
```
nwt, line 178:
```suggestion
		for _, o := range par.Ops {
			path := o.(*dag.Sequential)
```

compiler/optimizer/parallelize.go
```diff
@@ -2,189 +2,211 @@ package optimizer
...
-		// merge. Otherwise, no parallelization.
-		if sortKey.IsNil() {
-			// Unknown order: we can't parallelize because
-			// we can't maintain this unknown order at the merge point,
-			// but we shouldn't care if the client doesn't care
-			// so this goes back to whether the scan order was
-			// specified in the source. See Issue #2661.
-			return nil
+		for k := 0; k < len(par.Ops); k++ {
+			path := par.Ops[k].(*dag.Sequential)
```
nwt, line 197:
```suggestion
		for _, o := range par.Ops {
			path := o.(*dag.Sequential)
```

compiler/optimizer/parallelize.go
```diff
@@ -2,189 +2,211 @@ package optimizer
...
-				return nil
+		if merge == nil {
+			merge = &dag.Merge{
+				Kind:  "Merge",
+				Expr:  op.Args[0],
+				Order: op.Order,
 			}
+			if egress == 2 {
+				ops[1] = merge
+				ops[2] = &dag.Pass{Kind: "Pass"}
```
nwt, line 165:
```suggestion
				ops[2] = dag.PassOp
```

compiler/optimizer/parallelize.go
```diff
@@ -2,189 +2,211 @@ package optimizer
...
+			} else {
+				ops[egress] = merge
+			}
+		} else {
+			// There already was an appropriate merge.
+			// Smash the sort into a nop.
+			ops[egress] = &dag.Pass{Kind: "Pass"}
```
nwt, line 172:
```suggestion
			ops[egress] = dag.PassOp
```

compiler/optimizer/parallelize.go
```diff
@@ -2,189 +2,211 @@ package optimizer
...
-			// so this goes back to whether the scan order was
-			// specified in the source. See Issue #2661.
-			return nil
+		for k := 0; k < len(par.Ops); k++ {
+			path := par.Ops[k].(*dag.Sequential)
+			path.Ops = append(path.Ops, copyOp(op))
 		}
-		return replicateAndMerge(seq, sortKey, from, trunk, replicas)
+		// this will get removed later
+		ops[egress] = &dag.Pass{Kind: "Pass"}
```
nwt, line 201:
```suggestion
		ops[egress] = dag.PassOp
```

nwt (approved):

PR #4532, opened 2023-04-23
separate scope from sequential op

mccanne:
This commit creates a new operator for scopes and changes Sequential to exclude local decls.  This gets us ready to remove the Sequential op altogether and replace it was a new type `Seq` defined as `[]Op`, which is not itself an `Op`.  This is coming in a subsequent PR on branch `seq`.

We also updated the grammar so that declarations require a parenthesized scope block, except for the main scope, which requires no parentheses.

The clunkiness in handling the Scope operator as the entry point to the DAG will go away
soon when the `seq` branch is merged.


compiler/parser/parser_test.go
```diff
@@ -99,6 +99,7 @@ func TestZtestZed(t *testing.T) {
 	require.NoError(t, err)
 	for _, z := range zed {
 		testZed(t, z)
+		return
```
nwt, line 102: This means we'll only call testZed once.
```suggestion
```

zfmt/dag.go
```diff
@@ -526,11 +510,48 @@ func (c *canonDAG) over(o *dag.Over) {
...
-		c.op(o.Scope)
+		c.op(o.Body)
+		c.close()
+		c.ret()
+		c.flush()
+		c.write(")")
+	}
+}
+
+func (c *canonDAG) scope(p *dag.Scope, parens bool) {
```
nwt, line 525:
```suggestion
func (c *canonDAG) scope(s *dag.Scope, parens bool) {
```

compiler/parser/parser_test.go
mccanne: Oops, had that in there while debugging

compiler/job.go
```diff
@@ -146,14 +147,26 @@ func (j *Job) Parallelize(n int) error {
...
+	return wrapScope(op), err
+}
+
+func wrapScope(op ast.Op) ast.Op {
+	if seq, ok := op.(*ast.Sequential); ok {
+		op = &ast.Scope{Kind: "Scope", Body: seq}
+	}
+	return op
+}
+
```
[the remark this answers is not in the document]
mccanne: This is just for this PR.  Going away in the `seq` branch.  I hacked things up so `ast.Scope` can be the entry but this will change to `ast.Seq` in the next bit of work.  This is what I meant in the PR comment about this being clunky.  It is temporary.

compiler/parser/parser.peg
```diff
@@ -46,21 +46,24 @@
...
+start
+  = __ scope:Scope __ EOF { RETURN(scope) }
+  / __ seq:Sequential __ EOF { RETURN(seq) }
 
-Sequential
-  = decls:Decls __ first:Operation rest:SequentialTail* {
-      RETURN(MAP("kind": "Sequential", "ops": PREPEND(first, rest), "decls": decls))
+Scope
+  = decls:Decl+ seq:Sequential {
+      RETURN(MAP("kind": "Scope", "body": seq, "decls": decls))
```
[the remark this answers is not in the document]
mccanne: I'm not sure what you want me to do.  Change scope.Body to scope.Sequential?  Or change body:Sequential to seq:Sequential and then go with "body":seq?  To me it's most clear that the seq is in the body of a scope so I prefer it as I've written it and in the next PR body will be an ast.Seq which is a []dag.Op rather than a struct.

compiler/job.go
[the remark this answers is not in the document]
nwt: I guess my point is that the diffs in this PR and the the seq branch PR will be smaller if you deal with this in parser.peg, that's just a nit.

compiler/parser/parser.peg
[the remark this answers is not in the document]
nwt: No, I'm just suggesting changing the name `seq` to `body` here.

compiler/parser/parser.peg
```diff
@@ -580,12 +583,15 @@ MergeOp
...
+      RETURN(MAP("kind": "Over", "exprs": exprs, "locals": locals, "body": body))
     }
 
-Scope
-  = __ "=>" __ "(" __ seq:Sequential __ ")" { RETURN(seq) }
+// This is a little weird that you can have double parens, e.g., => ((foo | count()))
+// But we need the parens to define the subquery boundaries. Not sure how to fix it.
+Lateral
+  = __ "=>" __ "(" __ scope:Scope __ ")" { RETURN(scope) }
+  / __ "=>" __ "(" __ seq:Sequential __ ")" { RETURN(seq) }
```
nwt, line 594: Maybe shorten?
```suggestion
  = __ "=>" __ "(" __ s:(Scope / Sequential) __ ")" { RETURN(s) }
```

compiler/parser/parser.peg
```diff
@@ -46,21 +46,24 @@
 #define VAR(x) let x
 #endif
 
-start = __ ast:Sequential __ EOF { RETURN(ast) }
+start
+  = __ scope:Scope __ EOF { RETURN(scope) }
+  / __ seq:Sequential __ EOF { RETURN(seq) }
```
nwt, line 51: Maybe shorten?
```suggestion
  = __ s:(Scope / Sequential) __ EOF { RETURN(s) }
```

compiler/parser/parser.peg
```diff
@@ -580,12 +583,15 @@ MergeOp
...
-  = "over" _ exprs:Exprs locals:Locals? scope:Scope? {
-      RETURN(MAP("kind": "Over", "exprs": exprs, "locals": locals, "scope": scope))
+  = "over" _ exprs:Exprs locals:Locals? body:Lateral? {
+      RETURN(MAP("kind": "Over", "exprs": exprs, "locals": locals, "body": body))
     }
 
-Scope
-  = __ "=>" __ "(" __ seq:Sequential __ ")" { RETURN(seq) }
+// This is a little weird that you can have double parens, e.g., => ((foo | count()))
+// But we need the parens to define the subquery boundaries. Not sure how to fix it.
```
nwt, line 591: Are you sure this is true?

```console
$ go run ./cmd/zc -C 'over a => ((b | c))'
error parsing Zed at column 15:
over a => ((b | c))
          === ^ ===
exit status 1
```

zfmt/ast.go
```diff
@@ -636,11 +637,31 @@ func (c *canon) over(o *ast.Over) {
...
+		c.ret()
+	}
+	for _, d := range s.Decls {
+		c.decl(d)
+		c.ret()
+	}
+	//XXX functions?
+	c.flush()
+	c.proc(s.Body)
+	if parens {
```
nwt, line 664: Something's amiss here.

```console
$ go run ./cmd/zc -C 'a | (const b = 1 c)'
search a(
  const b = 1

  | search c
)
```

compiler/parser/parser.peg
mccanne: I made your change above so they return different things now.

zfmt/ast.go
mccanne: Yes, I saw that.  Rather than spend time on this now, how about we make an issue to fix the tab/ret logic in zfmt to use a stack and simplify things.  It's hard to use and when I first put it together it was a quick hack.

mccanne: I can futz with this on a separate branch.

nwt: Sounds good to me.

compiler/parser/parser.peg
mccanne: It's true for `over a => ((count()))`

How about we make an issue to  work on this some more?  All the tests are passing.

nwt: Sounds good.

compiler/parser/parser.peg
```diff
@@ -46,21 +46,26 @@
 #define VAR(x) let x
 #endif
 
-start = __ ast:Sequential __ EOF { RETURN(ast) }
+start
+  = __ scope:Scope __ EOF { RETURN(scope) }
+  / __ body:Sequential __ EOF {
+      RETURN(MAP("kind": "Scope", "body": body, "decls": nil))
```
nwt, line 52: NULL makes both JavaScript and Go happy.
```suggestion
      RETURN(MAP("kind": "Scope", "body": body, "decls": NULL))
```

nwt (approved):
Fix the nil/NULL thing in parser.peg and then :shipit:.

compiler/parser/parser.peg
mccanne: Hmm, you're right that the comment was wrong but I still don't understand why the double-parent count parses.

zfmt/ast.go
mccanne: #4534

PR #4543, opened 2023-04-25
update unpacker comments since mapstructure no longer used

mccanne:


nwt (approved):
There's also one on pkg/unpack/unpack_test.go.

PR #4547, opened 2023-04-25
Use pqarrow in zio/parquetio

### LINKED ISSUE #764: Support "Avro Old List Structure" in Parquet reader
I'm kinda parroting back concepts here that I don't fully understand, so please bear with me.

When outputting Parquet sample data with the Nifi **ParquetRecordSetWriter**, data that had started life as JSON arrays was written in a format that `zq` (as of commit `c1360a8`) choked on. These two are one such example:

[dns.parquet.gz](https://github.com/brimsec/zq/files/4618269/dns.parquet.gz)
[http.parquet.gz](https://github.com/brimsec/zq/files/4618272/http.parquet.gz)

The error message is:

```
$ zq -t -i parquet dns.parquet 
dns.parquet: LIST element (Answers) should have 1 child
```

Looking the `zq` code for where this error came from led me to https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#lists, which in turn inspired me to fiddle with the knobs in the **ParquetRecordSetWriter**. Through trial and error, I stumbled onto this setting:

That setting normally defaults to True, and the attachments above that `zq` choked on were generated when that was set to True. When I change it to False as is shown here, now the outputs are the attachments shown below, which `zq` reads without complaint.

[dns.parquet.gz](https://github.com/brimsec/zq/files/4618275/dns.parquet.gz)
[http.parquet.gz](https://github.com/brimsec/zq/files/4618276/http.parquet.gz)

At some point we should investigate this further and confirm if we need to enhance the variations we support such that we might be able to read the default output from Nifi without problems.

### LINKED ISSUE #4278: Replace zio/parquetio with zio/arrowio and pqarrow
We can replace the Parquet reader and writer in zio/parquetio with a combination of zio/arrowio and github.com/apache/arrow/go/v11/parquet/pqarrow, and we probably should since
* it'll let us remove about 750 lines from zio/parquetio,
* the Arrow Parquet implementation will probably receive more attention the one we're currently using, and
* it fixes #764.

### LINKED ISSUE #4527: Zed can't read Parquet output from DuckDB
Repro is with Zed commit 0a375f8. The test data originally came from a community user that reported https://github.com/brimdata/zui/issues/2751 and https://github.com/brimdata/zui/issues/2754.

Start with the attached test data [imdb.csv](https://github.com/brimdata/zed/files/11299489/imdb.csv).

```
$ zq -version
Version: v1.7.0-28-g0a375f88

$ zq -z 'count()' imdb.csv 
347(uint64)
```

A user imports it into DuckDB as a table and then exports that table as Parquet.

```
$ duckdb
v0.7.1 b00b93f0b1

D create table imdb as select * from 'imdb.csv';
D copy imdb to 'output.parquet' (FORMAT PARQUET);
```

Zed cannot read the Parquet that was produced.

```
$ zq -i parquet output.parquet 
output.parquet: *parquet.PageHeader field 15 read error: don't know what type: 13
```

However, tools like https://parquetreader.com seem able to read it without complaint.

I know we've suspected that following through on #4278 may allow us to cover more Parquet dialects, so I'll mark this issue as dependent on that one just in case that alone fixes it.

nwt:
Reading and writing are much faster with it than with github.com/fraugster/parquet-go.  Its only apparent drawback is that it offers no easy way to support Zed's duration and float16 types, and writing a value containing either produces a cryptic error.

    $ echo '{a:1.(float16)}' | zq -f parquet -
    parquetio: unsupported type: not implemented yet

Closes #764, closes #4278, and closes #4527.


philrz:
@nwt: Regarding the lack of `float16` and `duration` support, in your travels did you happen to find any open issues for the library pointing at this limitation? If not, would it be appropriate to open our own? It definitely seems there's net positives from moving to the new lib, but since we're still losing a little it might be nice if there were issues I could watch via Notifications so we know if/when things change. Given that it's two libraries that both purport to offer "Parquet" support I'd hope they'd one day converge, though maybe I shouldn't hold my breath. 😄

nwt:
@philrz: Our current support for these types is entirely on our side. There's no support for them in github.com/fraugster/parquet-go (I don't count the INTERVAL converted type because of its month component) and no issue for either in github.com/apache/arrow, and I doubt there ever will be given the pace of progress on apache/parquet-format#43, apache/parquet-format#165, and apache/parquet-format#184.

philrz:
This comment is basically a note-to-self to summarize an offline conversation @nwt and I had, as I may need to refer back to this in the future. What I now understand is that even though our prior Parquet support was able to output what started as Zed values of the `float16` or `duration` types, what ended up in the Parquet files were actually `float32` and `interval`. In essence we've never had true "round-trip" support with Parquet (i.e., if you wrote out values of those types with our old Parquet writer and then read that Parquet back in with Zed, you'd not get back values with Zed `float16` or `duration` types) so if it could be said anything was lost here in terms of functionality it was arguably only ever a placebo. As for the "cryptic error" that appears now, it's understood that ideally it would point to the specific type/value that caused it to choke, but the work involved to create the ideal error message appears non-trivial so we're keen to defer that effort until users actually bump into it.

mattnibs (approved):

PR #4563, opened 2023-04-28
use json.Unmarshal pattern in unpack

mccanne:
This commit further simplies the unpacker by adopting the method signature of json.Unmarshal.  Since reworking the packer design, it became apparent that the unpackVal method could be used with an arbitrary destination from the callee thereby providing the same semantics of json.Unmarshal, which simplifies the logic.

This also created an opportunity to simplify the zjson decoder to avoid the ugly workarounds with the previous unpack API.

### DIFF
--- a/zio/zjsonio/reader.go
+++ b/zio/zjsonio/reader.go
@@ -97,7 +97,7 @@ func (r *Reader) decodeValue(b *zcode.Builder, typ zed.Type, body interface{}) e
 		}
 		var t zType
 		if err := unpacker.UnmarshalObject(body, &t); err != nil {
-			return fmt.Errorf("zjsonio reader:type value is not a valid ZJSON type: %w", err)
+			return fmt.Errorf("type value is not a valid ZJSON type: %w", err)
 		}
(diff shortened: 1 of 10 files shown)


zio/zjsonio/reader.go
```diff
@@ -95,13 +95,9 @@ func (r *Reader) decodeValue(b *zcode.Builder, typ zed.Type, body interface{}) e
...
-		typeObj, err := unpacker.UnmarshalObject(body)
-		if err != nil {
-			return err
-		}
-		t, ok := typeObj.(zType)
-		if !ok {
-			return errors.New("type value is not a valid ZJSON type")
+		var t zType
+		if err := unpacker.UnmarshalObject(body, &t); err != nil {
+			return fmt.Errorf("zjsonio reader:type value is not a valid ZJSON type: %w", err)
```
nwt, line 100: Nit: Other error strings in this package don't have this style of prefix.
```suggestion
			return fmt.Errorf("type value is not a valid ZJSON type: %w", err)
```

nwt (approved):

PR #4568, opened 2023-05-02
Fix perf script and update results

philrz:
The `perf-compare.sh` script has been doing its job in CI of confirming `zq` still handles most permutations of input & output formats. But upon dusting it off today I found it's been a little broken since the changes in #3555, so I've fixed those in this PR. Specifically:

1. The move of the `zed=${ZED_QUERIES[$n]}` line caused the Zed query to no longer be output correctly in the markdown report.

2. The prior (admittedly ugly) `if`/`else` logic was making sure the Zed shaper was only applied with the one permutation of NDJSON input + Zeek TSV output (since it was only need to work around [this problem](https://github.com/brimdata/zed/pull/2123#issuecomment-859164320)), but the redo into `case` logic has had the shaper being invoked on _all_ permutations with NDJSON output or Zeek TSV output. In the interest of maintaining the cleanliness of the `case` logic, here I've changed it to apply the shaper on all the runs with NDJSON input and added a note up top explaining that the shaper has a constant cost.

Since it had been almost 2 years since the last time I updated the markdown results summary, I ran it at a current commit and did it on an AWS `t2.2xlarge` instead of the Google Cloud ones we'd used in the past, since we use AWS pretty much exclusively now.

### DIFF
--- a/performance/README.md
+++ b/performance/README.md
@@ -2,13 +2,10 @@
 
 The tables below provide a summary of simple operations and how `zq`
 performs at them relative to `zeek-cut` and `jq`. All operations were performed
-on a Google Cloud `n1-standard-8` VM (8 vCPUs, 30 GB memory) with the logs
-stored on a local SSD. `make perf-compare` was used to generate the results.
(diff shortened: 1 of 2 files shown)


nwt (approved):
🤦

PR #4576, opened 2023-05-04
Language Overview broken into separate files & without numeric headings

philrz:
A community zync user recently got stumped looking through the docs for help with [boolean logic](https://zed.brimdata.io/docs/language/overview#83-boolean-logic) and using `in` to test for [containment](https://zed.brimdata.io/docs/language/overview#73-containment). Once we helped point them at the right docs, we asked for input on what changes might have helped them find it on their own, since we've also recognized that some adjustments might be justified but hadn't yet come up with concrete ideas. In their own words:

> I think it would be perfect to have a submenu on the left for the "Overview" section the same way it is done for "Operators", "Functions", and "Aggregate Functions". In this submenu, we could find all the topics documented in "Overview". That way, they would be easier to find without the need to duplicate anything.

I spent some time considering this proposal and have come to agree with it. In its current state, the [Language Overview](https://zed.brimdata.io/docs/language/overview) doc is about double the size of the next biggest doc (the [`zq` Tutorial](https://zed.brimdata.io/docs/tutorials/zq)) when measured by either line count or word count(*). In my own work writing/debugging Zed and supporting users, Language Overview is easily the one I reference most often. Its current length and format does lead to some shortcomings, such as:

1. The lack of section headers in the left-hand nav, as cited by the user. So if a user is scanning the current list, the expandable nature of the sections for Operators/Functions/etc gives heavier weight to their contents, whereas it seems users interested in topics like Expressions/Data Types/etc. are less likely to click on Overview and scan through its large contents to find those.

image

2. Due to its current length, the right-hand nav when a reader clicks on **Overview** is much longer than will fit on a screen of typical size. We already know some users fail to even notice the right-hand nav, so this makes it worse.

image

3. Even at its current length, the number of "levels" in the right-hand nav is such that helpful things are not shown,  E.g., right now for **Expressions**, the user sees:

image

...but if **Expressions** was its own doc such that all the "levels" were moved "up by one", the right hand nav would now show:

image

This PR attempts to follow the user's suggestion and break out Language Overview into separate docs. Now there's entires in the left-hand nav for many of what have traditionally been section headers, including some particularly important ones like Expressions. The right-hand nav now shows an extra "level" for each of those smaller docs compared to what was there before.

image

As I was pulling the thread on that change, the leading numbers in each of the section headers started to look awkward, which led me to recall a discussion at a prior sync meeting where a team member had expressed a dislike of the numbers. Indeed, I can speak from personal experience that more link breakages have come about from this particular doc than any other, since we've often added new sections where it makes the most sense, but that means renumbering subsequent sections, which means fixing more links in other docs, etc. Contrast this with some other docs like [Zed Formats](https://zed.brimdata.io/docs/formats) specs, which are more "static" in nature and read more like RFCs, so the numbers seem more natural and less hazardous there.

In addition to this PR, I've put up a draft of the rendered docs in a staging site at https://spiffy-gnome-8f2834.netlify.app/docs/next/language/overview so folks can test drive it in its finished form.

---

(*) Strictly speaking, the [Schools Tutorial](https://zed.brimdata.io/docs/tutorials/schools) is our second longest doc. But I rarely find myself needing to reference it and I struggle to remember a user having mentioned reading it. It's crossed my mind at times to just dump it, so for the sake of this discussion I'm disregarding it.


jameskerr (commented):
This is great! I very much like the new format.

philrz:
After a chat with @nwt, I've adjusted the bucketing a bit from how it was in the first pass.

1. The previously separate sections for **Const Statements**, **Func Statements**, and **Type Statements** have been consolidated into a single **Const, Func, and Type Statements**. Each was fairly small and simple and they seem to go together logically. Having the words called out in section header in the left nav should hopefully catch the reader's attention if that's the stuff they're looking for.

2. The section for **Data Types** has been moved "up" so it's now right below **Dataflow Model**. It seems to flow better this way.

3. The previously separate sections for **Shaping** and **Type Fusion** have been consolidated into a single **Shaping and Type Fusion**. They seem similar enough in their respective missions.

You can see it in its rendered form at the staging site https://spiffy-gnome-8f2834.netlify.app/docs/next/language/overview.

philrz:
I had one more round of offline feedback from @nwt, which has led to the most recent (and maybe final?) revision. The big change is that the individual files previously broken out from what was the large **Overview** have all been "moved up a level". Now the left nav in the rendered version at https://spiffy-gnome-8f2834.netlify.app/docs/next/language looks like:

image

Keeping with the spirit of what started this PR, the thinking behind this most recent change is that some of those new sections like **Expressions** are probably just as important as the companion/reference groups like **Operators** and **Functions**, therefore bringing them up a level makes them even easier to spot vs. when they were in an expandable grouping like I had for **Overview** in a previous revision.

I've once again fixed all the broken links within the Zed repo and have a clean link checker run against the staging site at https://github.com/philrz/broken-link-checker/actions/runs/5094968820/jobs/9159411083. If this PR merges, there will absolutely be a wave of broken links in other repos and blog posts when we tag the next docs version, so I've made a note-to-self to plan for that with careful timing.

Once I made this most recent change, I felt compelled to fit in a semi-related change. I've always felt the **Conventions** doc felt a bit awkward, and this most recent change made it seem even more awkward to me. I feel like I understood its previous placement in that it was placed in the left nav right before the **Operators**, **Functions**, and **Aggregate Functions** reference docs about which it's supposed to be giving clarity, so a user reading all the docs top-down would come to the **Conventions** before they dive into those reference docs. However, I have a sneaking suspicion that users are more likely to access those reference-style materials on an individual basis (e.g., follow a hyperlink in another doc to learn more about an operator, or look up an operator in the Search tool after seeing it referenced in a program, hence jump right to that one doc about that one operator) and hence are unlikely to have ever seen **Conventions**. Granted, I feel like the prototypes in those reference docs are fairly intuitive, so I'm not even sure **Conventions** is must-see content, but I also don't see a reason to do away with it entirely. But if a user is scanning the left hand nav, I _did_ think that after this most recent change it kinda broke up the flow even more than before. Therefore I've done two things:

1. Below the **Synopsis** on each individual **Function** and **Aggregate Function** page, I've added a small "See also: [Type conventions](https://spiffy-gnome-8f2834.netlify.app/docs/next/language/conventions)" link. Since that's right below where the user is seeing references to specific types, classes of types, or "any", if they feel any confusion at all I figure that's where we want to offer help. I don't love the repetition, but it's a small enough that I feel it's worthwhile.

2. In the left nav, I moved it below the groupings of the reference sections for **Operators**, **Functions**, and **Aggregate Functions**. I don't feel super strongly about this particular change, but after the other changes I found myself liking everything about the latest look of the left nav except for how **Conventions** seemed to interrupt the flow, so if others like how I'm linking to it from the individual reference docs, it feels ok to move it out of the way a bit.

One more thing about that **Conventions** doc. One of the things it says:

> ...the function and operator prototypes in the Zed documentation include several type classes as follows:

Strictly speaking, only the function & aggregate functions specify types currently in their prototypes. It's for this reason that I did _not_ link to the **Conventions** doc from the synopses of the pages for the individual operators. This made me wonder if I should also remove the reference to operators from within the **Conventions** doc, but I'd already kinda let some scope-creep happening, so I stopped short of that.

docs/language/README.md
```diff
@@ -2,6 +2,11 @@
 
 The language documents:
 * provide an [overview](overview.md) of the Zed language,
-* layout some [conventions](conventions.md) for the documenation, and
+* describe Zed's [dataflow model](dataflow-model.md),
+* show the syntax of [statements](statements.md) that define constants, functions, and named types,
+* explain how Zed [data types](data-types.md) are used,
+* describe [expression](expressions.md) syntax, including for [search](search-expressions.md),
+* explain how [lateral queries](lateral-subqueries.md) are used,
+* describe the the concepts of [shaping and type fusion](shaping.md), and
```
philrz, line 10: It felt awkward to write such a long extension of this narrative reveal of hyperlinks, so if folks wanna go another direction here, I'm open to that.

docs/language/README.md
```diff
@@ -2,6 +2,11 @@
 
 The language documents:
 * provide an [overview](overview.md) of the Zed language,
-* layout some [conventions](conventions.md) for the documenation, and
+* describe Zed's [dataflow model](dataflow-model.md),
+* show the syntax of [statements](statements.md) that define constants, functions, and named types,
+* explain how Zed [data types](data-types.md) are used,
```
nwt, line 7:
```suggestion
* explain Zed's [data types](data-types.md),
* show the syntax of [statements](statements.md) that define constants, functions, and named types,

```

docs/language/README.md
```diff
@@ -2,6 +2,11 @@
...
+* describe [expression](expressions.md) syntax, including for [search](search-expressions.md),
```
nwt, line 8:
```suggestion
* describe the syntax of [expressions](expressions.md) and [search expressions](search-expressions.md),
```

docs/language/README.md
```diff
@@ -2,6 +2,11 @@
...
+* explain how [lateral queries](lateral-subqueries.md) are used,
```
nwt, line 9:
```suggestion
* explain [lateral subqueries](lateral-subqueries.md),
```

docs/language/README.md
nwt, line 10:
```suggestion
* describe [shaping and type fusion](shaping.md), and
```

nwt (approved):
:shipit:!

PR #4577, opened 2023-05-05
zed vacuum

### LINKED ISSUE #4113: Vacuum support
The following text was present in a retired "lake design" document (see #3803). It has been established that this was really a pending to-do, so this issue tracks its ultimate implementation and corresponding updates to docs.

```
## Vacuum Support

While data objects currently can be deleted from a lake, the underlying data
is retained to support time travel.

The system must also support purging of old data so that retention policies
can be implemented.

This could be supported with the DANGER-ZONE command `zed vacuum`
(implementation tracked in [zed/2545](https://github.com/brimdata/zed/issues/2545)).
The commits still appear in the log but scans at any time-travel point
where the commit is present will fail to scan the deleted data.
In this case, perhaps we should emit a structured Zed error describing
the meta-data of the object that was unavailable.

Alternatively, old data can be removed from the system using a safer
command (but still in the DANGER-ZONE), `zed vacate` (also
[zed/2545](https://github.com/brimdata/zed/issues/2545)) which moves
the tail of the commit journal forward and removes any data no longer
accessible through the modified commit journal.
```

mattnibs:
Add vacuum command to zed the removes objects that are no longer visible from a specific commit point.

Closes #4113

### DIFF
--- a/api/client/connection.go
+++ b/api/client/connection.go
@@ -421,8 +421,8 @@ func (c *Connection) delete(ctx context.Context, poolID ksuid.KSUID, branchName 
 	return commit, err
 }
 
-func (c *Connection) Vacuum(ctx context.Context, pool, ref string, dryrun bool) (api.VacuumResponse, error) {
-	path := urlPath("pool", pool, "revision", ref, "vacuum")
+func (c *Connection) Vacuum(ctx context.Context, pool, revision string, dryrun bool) (api.VacuumResponse, error) {
+	path := urlPath("pool", pool, "revision", revision, "vacuum")
 	if dryrun {
 		path += "?dryrun=true"
 	}
--- a/api/client/connection.go
+++ b/api/client/connection.go
@@ -421,6 +421,17 @@ func (c *Connection) delete(ctx context.Context, poolID ksuid.KSUID, branchName 
 	return commit, err
 }
 
+func (c *Connection) Vacuum(ctx context.Context, pool, ref string, dryrun bool) (api.VacuumResponse, error) {
+	path := urlPath("pool", pool, "revision", ref, "vacuum")
+	if dryrun {
+		path += "?dryrun=true"
+	}
+	req := c.NewRequest(ctx, http.MethodPost, path, nil)
+	var res api.VacuumResponse
+	err := c.doAndUnmarshal(req, &res)
+	return res, err
+}
+
 func (c *Connection) SubscribeEvents(ctx context.Context) (*EventsClient, error) {
 	req := c.NewRequest(ctx, http.MethodGet, "/events", nil)
 	req.Header.Set("Accept", api.MediaTypeZSON)
--- a/cmd/zed/vacuum/command.go
+++ b/cmd/zed/vacuum/command.go
@@ -53,26 +53,19 @@ func (c *Command) Run(args []string) error {
 	if err != nil {
 		return err
 	}
-	if c.dryrun {
-		oids, err := lk.Vacuum(ctx, at.Pool, at.Branch, true)
-		if err != nil {
+	verb := "would vacuum"
+	if !c.dryrun {
+		verb = "vacuumed"
+		if err := c.confirm(at.String()); err != nil {
 			return err
 		}
-		// It's a bit weird to specify -q and -dryrun, but do as they say.
-		if !c.LakeFlags.Quiet {
-			fmt.Printf("would vacuum %d object%s\n", len(oids), plural.Slice(oids, "s"))
-		}
-		return nil
 	}
-	if err := c.confirm(at.String()); err != nil {
-		return err
-	}
-	oids, err := lk.Vacuum(ctx, at.Pool, at.Branch, false)
+	oids, err := lk.Vacuum(ctx, at.Pool, at.Branch, c.dryrun)
 	if err != nil {
 		return err
 	}
 	if !c.LakeFlags.Quiet {
-		fmt.Printf("vacuumed %d object%s\n", len(oids), plural.Slice(oids, "s"))
+		fmt.Printf("%s %d object%s\n", verb, len(oids), plural.Slice(oids, "s"))
 	}
 	return nil
 }
--- a/cmd/zed/vacuum/command.go
+++ b/cmd/zed/vacuum/command.go
@@ -47,7 +47,7 @@ func (c *Command) Run(args []string) error {
 	defer cleanup()
 	at, err := c.LakeFlags.HEAD()
 	if err != nil {
-		return errors.New("default pool and branch unset")
+		return err
 	}
 	lk, err := c.LakeFlags.Open(ctx)
 	if err != nil {
--- a/cmd/zed/vacuum/command.go
+++ b/cmd/zed/vacuum/command.go
@@ -1,0 +1,94 @@
+package vacuum
+
+import (
+	"errors"
+	"flag"
+	"fmt"
+	"strings"
+
+	"github.com/brimdata/zed/cmd/zed/root"
+	"github.com/brimdata/zed/pkg/charm"
+	"github.com/brimdata/zed/pkg/plural"
+)
+
+var Cmd = &charm.Spec{
+	Name:  "vacuum",
+	Usage: "vacuum [options]",
+	Short: "clear space by removing unreferenced objects",
+	Long: `
+"zed vacuum" clears up space in a pool by removing objects that are not visible 
+from a pool's branch or commit.
+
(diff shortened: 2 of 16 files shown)


lake/api/api.go
```diff
@@ -44,6 +44,8 @@ type Interface interface {
 	UpdateIndex(ctx context.Context, names []string, pool ksuid.KSUID, branchName string) (ksuid.KSUID, error)
 	AddVectors(ctx context.Context, pool ksuid.KSUID, branch string, objects []ksuid.KSUID, message api.CommitMessage) (ksuid.KSUID, error)
 	DeleteVectors(ctx context.Context, pool ksuid.KSUID, branch string, objects []ksuid.KSUID, message api.CommitMessage) (ksuid.KSUID, error)
+	Vacuumable(ctx context.Context, pool, ref string) ([]ksuid.KSUID, error)
+	Vacuum(ctx context.Context, pool string, objects []ksuid.KSUID) ([]ksuid.KSUID, error)
```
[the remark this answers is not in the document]
mattnibs: 1. Yeah I was thinking about this over the weekend and thing we could do with just one endpoint with a dry run / delete flag. Will fix.
2. Yeah I think that makes sense will fix.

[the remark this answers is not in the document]
mattnibs: In terms of the delete flag, I think the flag should be called "dryrun" since if the endpoint's name is vacuum it suggests that the default is that the object is deleted.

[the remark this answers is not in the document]
nwt: @mattnibs: For the HTTP API, do you think GET /pool/{pool}/vacuum/{revision} for a dry run and POST with the same URL for deletion?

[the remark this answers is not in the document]
mattnibs: Having thought about this far more than  warranted, I'm leaning towards keeping this all one endpoint. This is more based on intuition than anything but it seems that if Vacuum exists as only a single call in lake/api.Interface than it should also only have one endpoint.

api/client/connection.go
```diff
@@ -421,6 +421,17 @@ func (c *Connection) delete(ctx context.Context, poolID ksuid.KSUID, branchName
 	return commit, err
 }
 
+func (c *Connection) Vacuum(ctx context.Context, pool, ref string, dryrun bool) (api.VacuumResponse, error) {
+	path := urlPath("pool", pool, "revision", ref, "vacuum")
```
nwt, line 425:
```suggestion
func (c *Connection) Vacuum(ctx context.Context, pool, revision string, dryrun bool) (api.VacuumResponse, error) {
	path := urlPath("pool", pool, "revision", revision, "vacuum")
```

lake/pool.go
```diff
@@ -139,6 +144,20 @@ func (p *Pool) OpenBranchByName(ctx context.Context, name string) (*Branch, erro
 	return p.openBranch(ctx, branchRef)
 }
 
+// ResolveRevision returns the commit id for revision. revision can be either a
+// commit ID in string form or a branch name.
+func (p *Pool) ResolveRevision(ctx context.Context, ref string) (ksuid.KSUID, error) {
+	id, err := lakeparse.ParseID(ref)
+	if err != nil {
+		branch, err := p.LookupBranchByName(ctx, ref)
```
nwt, line 152:
```suggestion
func (p *Pool) ResolveRevision(ctx context.Context, revision string) (ksuid.KSUID, error) {
	id, err := lakeparse.ParseID(revision)
	if err != nil {
		branch, err := p.LookupBranchByName(ctx, revision)
```

cmd/zed/vacuum/command.go
```diff
@@ -0,0 +1,94 @@
...
+type Command struct {
+	*root.Command
+	force  bool
+	dryrun bool
+}
+
+func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
+	c := &Command{Command: parent.(*root.Command)}
+	f.BoolVar(&c.force, "f", false, "do not prompt for confirmation")
+	f.BoolVar(&c.dryrun, "dryrun", false, "vacuum without deleting objects")
```
nwt, line 38:
```suggestion
	f.BoolVar(&c.dryrun, "dryrun", false, "print objects that would be removed but do not remove them")
	f.BoolVar(&c.force, "f", false, "do not prompt for confirmation")
```

cmd/zed/vacuum/command.go
```diff
@@ -0,0 +1,94 @@
...
+
+func (c *Command) Run(args []string) error {
+	ctx, cleanup, err := c.Init()
+	if err != nil {
+		return err
+	}
+	defer cleanup()
+	at, err := c.LakeFlags.HEAD()
+	if err != nil {
+		return errors.New("default pool and branch unset")
```
nwt, line 50: Every command except for "zed use" just returns the error from LakeFlags.HEAD.
```suggestion
		return err
```

cmd/zed/vacuum/command.go
```diff
@@ -0,0 +1,94 @@
...
+		return err
+	}
+	oids, err := lk.Vacuum(ctx, at.Pool, at.Branch, false)
+	if err != nil {
+		return err
+	}
+	if !c.LakeFlags.Quiet {
+		fmt.Printf("vacuumed %d object%s\n", len(oids), plural.Slice(oids, "s"))
+	}
+	return nil
```
nwt, line 77:
```suggestion
	verb := "would vacuum"
	if !c.dryrun {
		verb = "vacuumed"
		if err := c.confirm(at.String()); err != nil {
			return err
		}
	}
	oids, err := lk.Vacuum(ctx, at.Pool, at.Branch, c.dryrun)
	if err != nil {
		return err
	}
	if !c.LakeFlags.Quiet {
		fmt.Printf("%s %d object%s\n", verb, len(oids), plural.Slice(oids, "s"))
	}
	return nil
```

api/client/connection.go
```diff
@@ -421,6 +421,17 @@ func (c *Connection) delete(ctx context.Context, poolID ksuid.KSUID, branchName
 	return commit, err
 }
 
+func (c *Connection) Vacuum(ctx context.Context, pool, revision string, dryrun bool) (api.VacuumResponse, error) {
+	path := urlPath("pool", pool, "revision", revision, "vacuum")
+	if dryrun {
+		path += "?dryrun=true"
```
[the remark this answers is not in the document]
mattnibs: I'm going to keep this for now but you're preference is duly noted.

lake/api/local.go
```diff
@@ -41,7 +40,6 @@ func OpenLocalLake(ctx context.Context, logger *zap.Logger, lakePath string) (In
 	return &local{
 		root:     root,
 		compiler: compiler.NewLakeCompiler(root),
-		engine:   engine,
 	}, nil
```
nwt, line 43:
```suggestion
	return FromRoot(root), nil
```

lake/api/local.go
```diff
@@ -56,11 +54,14 @@ func CreateLocalLake(ctx context.Context, logger *zap.Logger, lakePath string) (
 		return nil, err
 	}
 	return &local{
-		root:   root,
-		engine: engine,
+		root: root,
 	}, nil
```
nwt, line 58:
```suggestion
	return FromRoot(root), nil
```

runtime/op/meta/sequence.go
```diff
@@ -147,6 +148,12 @@ func newObjectScanner(ctx context.Context, zctx *zed.Context, pool *lake.Pool, o
 	}
 	rc, err := object.NewReader(ctx, pool.Storage(), pool.DataPath, ranges)
 	if err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
+			// If the object does not exist return a zbuf.Puller that returns a
+			// single value- a zed error.
+			vals := zbuf.NewArray([]zed.Value{*zctx.NewError(err)})
+			return zbuf.NewPuller(vals), nil
+		}
```
nwt, line 156: This is a neat idea and we should eventually add this sort of error tolerance as option, but I don't think it's the right default behavior.

mattnibs: This was part of the original issue description https://github.com/brimdata/zed/issues/4113:

> In this case, perhaps we should emit a structured Zed error describing the meta-data of the object that was unavailable.

Are you saying that the query should just return an error?

nwt: Yes, it should return an error.

mattnibs: ok

nwt (approved):

PR #4578, opened 2023-05-05
Use PEP 440 version in python/zed/setup.py

nwt:
service/ztests/python.yaml and service/ztests/python-auth.yaml fail with Python 3.10 or better because setuptools now requires a package version that is valid according to PEP 440.  Fix that.


mattnibs (approved):
Not sure what this does but it looks beautiful 👍🏻

PR #4581, opened 2023-05-08
Allow script-style ztests to run on Windows

nwt:
Script-style ztests require Bash on every OS and run with PATH set to the equivalent of "$ZTEST_PATH:$PATH".  This is a change from the previous value of "/bin:/usr/bin:$ZTEST_PATH".


mattnibs (approved):
This is great!

PR #4589, opened 2023-05-11
Improve JSON output for Zed maps

### LINKED ISSUE #4567: Improve JSON output of Zed map
Repro is with Zed commit 4ffdf3e.

As is described in more detail in #4565, right now outputting a Zed `map` type as JSON results in this extended `key` / `value` representation.

```
$ zq -version
Version: v1.7.0-55-g4ffdf3ef

$ echo '|{"hello":"world"}|' | zq -j -
[{"key":"hello","value":"world"}]
```

After polling the team, the consensus seems to be that this should instead be output as a simpler JSON object:

```
{"hello":"world"}
```

nwt:
In JSON output, a Zed map becomes an array of {"key":KEY,"value":VALUE} objects.  Change that to a single object object whose keys are ZSON representations of the map keys, as in {"KEY":VALUE}.

* String keys are used directly.
* Non-string primitive keys are converted to undecorated ZSON.
* Union keys are untagged and then converted to ZSON (with decoration for uniqueness).
* Enum keys are converted to their corresponding symbol.
* All other keys are converted to ZSON.

Closes #4567.


nwt:
I think this is probably what we want, but I'm definitely open to tweaking any of the key representations.

The one thing here I'm not sure about is floating point keys with integer values. Right now this converts them to, e.g., "1." because that's the ZSON representation, but we might want to make that "1.0".

philrz:
Great ztests! It's just one person's opinion, but I'd be happy with what you've got it doing right now.

I'm not inclined to sweat the float keys too hard. FWIW, I've sometimes paused at the trailing dot in the ZSON output. Since we're converting to JSON, this seems maybe relevant:

```
$ echo '5. 5.0' | zq -j -
5
5
```

Which is to say it seems like we could drop with the dot altogether. But since in other logic you're converting to the ZSON equivalent then I could see why keeping the trailing dot would make it consistent because:

```
$ echo '5. 5.0' | zq -z -
5.
5.
```

 If we decided the trailing `.0` was desirable in the map-as-JSON output then it'd make me wonder why it's not just as desirable in the general ZSON output case.

PR #4598, opened 2023-05-19
Fix vacuum -dryrun

### LINKED ISSUE #4597: "zed vacuum -dryrun" deletes objects
Repro is with Zed commit 89bb056.

I bumped into this while verifying the `zed vacuum` command added in #4577. Here I follow the same verification steps shown in https://github.com/brimdata/zed/issues/4113#issuecomment-1554909295, but rather than actually executing the `vacuum`, I tried the `-dryrun` option with the expectation that it would tell me about objects that would be deleted from storage without actually deleting them.

```
$ zed -version
Version: v1.7.0-68-g89bb0569

$ zed create zeek

$ zed use zeek

$ for file in *; do   zed -use zeek load $file; done

$ du -sh ~/lake
 60M	/Users/phil/lake

$ for logtype in $(zed query -f text 'by _path | sort -r' | tail -n +2); do   zed delete -where "_path==\"$logtype\""; done
2Q1NSmxWgFcz5TjggjXi9v9MB2I delete committed
2Q1NSjpPW2A8YGul7s7ELhiNvkd delete committed
...

$ du -sh ~/lake
 61M	/Users/phil/lake

$ zed vacuum -dryrun
would vacuum 49 objects

$ du -sh ~/lake
1.1M	/Users/phil/lake

$ zed query 'from zeek@2Q1NPCKMW3Ivcj2hPpgWWTwmFqr | count() by _path | sort -r'
file:///Users/phil/lake/2Q1NOJEYPIJUZO5wfVHcZE39Vlo/data/2Q1NP6SaBnvr5w7zAu8rN97ZBw3.zng: file does not exist
```

In addition to the storage in the lake directory being reduced when I did not expect it, I also noticed that the object count in the `would vacuum` message is roughly double the count of 25 objects shown in https://github.com/brimdata/zed/issues/4113#issuecomment-1554909295 when I was letting the `vacuum` actually execute.

mattnibs:
Closes #4597

### DIFF
--- a/lake/pool.go
+++ b/lake/pool.go
@@ -239,6 +239,7 @@ func (p *Pool) Vacuum(ctx context.Context, commit ksuid.KSUID, dryrun bool) ([]k
 				}
 				return err
 			})
+			continue
 		}
 		group.Go(func() error {
 			err := p.engine.Delete(ctx, data.SequenceURI(p.DataPath, o.ID))
--- a/lake/ztests/vacuum.yaml
+++ b/lake/ztests/vacuum.yaml
@@ -6,6 +6,7 @@ script: |
   echo {x:1} | zed load -q -
   r=$(echo {x:2} | zed load - | head -1 | awk '{print $1}')
   r=$(zed revert $r | awk '{print $5}')
+  zed vacuum -dryrun
   zed vacuum -f
   zed revert -q $r
   ! zed query -z '*'
@@ -13,6 +14,7 @@ script: |
 outputs:
   - name: stdout
     data: |
+      would vacuum 1 object
       vacuumed 1 object
   - name: stderr
     regexp:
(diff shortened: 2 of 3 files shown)


lake/ztests/vacuum-dryrun.yaml
```diff
@@ -0,0 +1,19 @@
+script: |
+  export ZED_LAKE=test
+  zed init -q
+  zed create -q -orderby x:asc test
+  zed use -q test
+  echo {x:1} | zed load -q -
+  r=$(echo {x:2} | zed load - | head -1 | awk '{print $1}')
+  r=$(zed revert $r | awk '{print $5}')
+  zed vacuum -f -dryrun
```
nwt, line 9: Instead of adding separate tests, can't we get the same coverage by just doing `zed vacuum -f -dryrun` before `zed vacuum -f` in the existing tests?

mattnibs: Yeah ok

nwt (approved):

PR #4600, opened 2023-05-19
Add vacuum docs

philrz:


docs/lake/api.md
```diff
@@ -120,13 +120,46 @@ DELETE /pool/{pool}
...
+     http://localhost:9867/pool/inventory
 ```
 
 On success, HTTP 204 is returned with no response payload.
 
 ---
 
+#### Vacuum pool
+
+Clear storage space by permanently deleting underlying data objects that have
```
nwt, line 132:
```suggestion
Free storage space by permanently removing underlying data objects that have
```

docs/lake/api.md
```diff
@@ -120,13 +120,46 @@ DELETE /pool/{pool}
...
+POST /pool/{pool}/revision/{revision}/vacuum
+```
+
+**Params**
+
+| Name | Type | In | Description |
+| ---- | ---- | -- | ----------- |
+| pool | string | path | **Required.** ID or name of the requested pool. |
+| revision | string | path | **Required.** The starting point for locating objects that can be vacuumed. Can be the name of a branch (whose tip would be used) or a commit ID. |
+| dryrun | string | query | Set to "T" to return the list of objects that could be vacuumed, but don't actually delete them. Defaults to "F". |
```
nwt, line 145:
```suggestion
| dryrun | string | query | Set to "T" to return the list of objects that could be vacuumed, but don't actually remove them. Defaults to "F". |
```

nwt (approved):
I suggest we use the term "remove" rather than "delete" to denote removal of data objects. I think that might prevent some confusion about what "delete" means given the existence of `zed delete`.

docs/commands/zed.md
```diff
@@ -894,3 +893,15 @@ the pool name to the desired branch:
...
+```
+zed vacuum [options]
+```
+
+The `vacuum` command permanently deletes underlying data objects that have
+previously been subject to a [`delete`](#24-delete) operation.  As this is a
+DANGER ZONE command, you must confirm that you want to delete
+the pool to proceed.  The `-f` option can be used to force the deletion
+without confirmation.  The `-dryrun` option may also be used to see a summary
+of how many objects would be deleted by a `vacuum` but without deleting them.
```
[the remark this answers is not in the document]
philrz: @nwt: Good idea. And thanks for catching "pool". Stupid copy/paste mistake.

PR #4603, opened 2023-05-24
semantic compiler: Add analyzer struct

mattnibs:
Add central analyzer struct for semantic compilation.

### DIFF
--- a/compiler/semantic/expr.go
+++ b/compiler/semantic/expr.go
@@ -15,7 +15,7 @@ import (
 	"github.com/brimdata/zed/zson"
 )
 
-func semExpr(scope *Scope, e ast.Expr) (dag.Expr, error) {
+func (a *analyzer) semExpr(e ast.Expr) (dag.Expr, error) {
 	switch e := e.(type) {
 	case nil:
 		return nil, errors.New("semantic analysis: illegal null value encountered in AST")
@@ -32,7 +32,7 @@ func semExpr(scope *Scope, e ast.Expr) (dag.Expr, error) {
 			Expr:    pathOf("this"),
 		}, nil
 	case *ast.Grep:
-		return semGrep(scope, e)
+		return a.semGrep(e)
 	case *astzed.Primitive:
 		val, err := zson.ParsePrimitive(e.Type, e.Text)
 		if err != nil {
@@ -43,7 +43,7 @@ func semExpr(scope *Scope, e ast.Expr) (dag.Expr, error) {
 			Value: zson.MustFormatValue(val),
 		}, nil
 	case *ast.ID:
-		return semID(scope, e), nil
+		return a.semID(e), nil
 	case *ast.Term:
 		var val string
 		switch t := e.Value.(type) {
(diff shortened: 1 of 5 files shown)


compiler/semantic/ast.go
```diff
@@ -13,5 +13,6 @@ import (
 // to DAG form, resolving syntax ambiguities, and performing constant propagation.
 // After semantic analysis, the DAG is ready for either optimization or compilation.
 func Analyze(ctx context.Context, seq ast.Seq, source *data.Source, head *lakeparse.Commitish) (dag.Seq, error) {
-	return semSeq(ctx, NewScope(), seq, source, head)
+	c := newChecker(ctx, source, head)
+	return c.check(seq)
```
mattnibs, line 17: In the forthcoming user ops pr this will have additional steps.

compiler/semantic/op.go
```diff
@@ -144,17 +144,17 @@ func semSortKey(p *ast.SortKey) (order.SortKey, error) {
...
+func (c *checker) semPool(scope *Scope, p *ast.Pool) ([]dag.Op, error) {
 	var poolNames []string
 	var err error
 	switch specPool := p.Spec.Pool.(type) {
 	case nil:
 		// This is a lake meta-query.
 		poolNames = []string{""}
 	case *ast.Glob:
-		poolNames, err = matchPools(ctx, ds, reglob.Reglob(specPool.Pattern), specPool.Pattern, "glob")
+		poolNames, err = matchPools(c.ctx, c.source, reglob.Reglob(specPool.Pattern), specPool.Pattern, "glob")
```
nwt, line 155: Make this a method so you don't have to pass c.ctx and c.source?

compiler/semantic/op.go
```diff
@@ -165,7 +165,7 @@ func semPool(ctx context.Context, scope *Scope, p *ast.Pool, ds *data.Source, he
 	}
 	var sources []dag.Op
 	for _, name := range poolNames {
-		source, err := semPoolWithName(ctx, scope, p, name, ds, head)
+		source, err := semPoolWithName(c.ctx, scope, p, name, c.source, c.head)
```
nwt, line 168: Make this a checker method so you don't have to pass c.ctx, c.source, and c.head?

compiler/semantic/expr.go
```diff
@@ -15,7 +15,7 @@ import (
 	"github.com/brimdata/zed/zson"
 )
 
-func semExpr(scope *Scope, e ast.Expr) (dag.Expr, error) {
+func (c *checker) semExpr(scope *Scope, e ast.Expr) (dag.Expr, error) {
```
nwt, line 18: Any reason not to drop the scope parameter everywhere in favor of checker.scope?

mattnibs: Yeah I was just thinking the same thing, I'll do this.

compiler/semantic/ast.go
nwt, line 17: Existing comments and error strings almost always refer to this step as semantic analysis, so should we rename "checker" to "analyzer" for consistency with that?

mattnibs: ok

compiler/semantic/op.go
```diff
@@ -931,9 +928,9 @@ func semCallOp(scope *Scope, call *ast.Call, seq dag.Seq) (dag.Seq, error) {
 	if !function.HasBoolResult(call.Name) {
 		return nil, nil
 	}
-	c, err := semCall(scope, call)
+	out, err := a.semCall(call)
```
nwt, line 931: Nit: Leave this name unchanged.
```suggestion
	c, err := a.semCall(call)
```

nwt (approved):

PR #4607, opened 2023-05-24
Remove unnecessary type conversions

nwt:
https://github.com/mdempsky/unconvert found these.


mattnibs (approved):
Cool, can we add this to our vet check?

nwt:
> Cool, can we add this to our vet check?

We can!

PR #4624, opened 2023-06-05
Enhanced user ops example

philrz:
When reviewing #4417 I thought this example could maybe be improved. I tried proposing this change via the GitHub Suggestion workflow but the excessive use of triple backticks with mdtest made the Preview look all weird such that I wasn't convinced it was gonna work correctly. 😛 

In summary, with this change:

1. Multiple data types are shown in the record rather than just strings, including a nested record just to make it more obvious that all the usual stuff is fair game
2. `this` is shown working simultaneously in the different scopes
3. The example is now protected with mdtest


mattnibs (approved):
I like it!

PR #4629, opened 2023-06-06
Add JSON omitempty option to ztest.ZTest.ErrorRE

nwt:
All other exported fields in ZTest have it.


nwt:
@mccanne: They're occasionally useful for automated stuff like #2910 and #4630.

PR #4633, opened 2023-06-07
fix zson marshaling of unknown interface values

mccanne:
This commit changes the logic that determines Zed types from Go native types to always use zed type null for interface types since the underlying concrete type is not known.  Most of the marshaling logic uses values to infer types rather than type lookups but when there is a nil value of a complex type with embedded interfaces, this heuristic is needed.  This still doesn't cover all the possible cases here but it is an improvement over what was here before.

### DIFF
--- a/zson/marshal.go
+++ b/zson/marshal.go
@@ -410,16 +410,9 @@ func (m *MarshalZNGContext) encodeMap(v reflect.Value) (zed.Type, error) {
 }
 
 func (m *MarshalZNGContext) encodeNil(t reflect.Type) (zed.Type, error) {
-	var typ zed.Type
-	if t.Kind() == reflect.Interface {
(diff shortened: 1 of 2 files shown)


zson/marshal.go
```diff
@@ -578,6 +571,9 @@ func (m *MarshalZNGContext) lookupType(t reflect.Type) (zed.Type, error) {
 		typ = zed.TypeFloat32
 	case reflect.Float64:
 		typ = zed.TypeFloat64
+	case reflect.Interface:
+		// Encode interfaces when we don't knwo the underlying concrete type as null type.
```
nwt, line 575:
```suggestion
		// Encode interfaces when we don't know the underlying concrete type as null type.
```

nwt (approved):

PR #4635, opened 2023-06-07
UserOps: Add const param

mattnibs:
Add a const param to user ops that requires a literal argument.


docs/language/statements.md
```diff
@@ -135,6 +135,30 @@ produces
 {foo:"foo",bar:"bar",ts:2021-01-01T00:00:00Z}
 ```
 
+### Const Parameters
+
+User-defined operators may use the `const` keyword to indicate that a parameter
+is expecting a const value. Const parameters are different from standard named
```
philrz, line 141:
```suggestion
is expecting a constant value. Const parameters are different from standard named
```

Not a biggie, but saying "const" twice felt tautological.

runtime/op/scope.go
```diff
@@ -0,0 +1,60 @@
...
+	if len(vars) == 0 {
+		return parent
+	}
+	return &enterScope{parent, vars}
+}
+
+func (s *enterScope) Pull(done bool) (zbuf.Batch, error) {
+	batch, err := s.parent.Pull(done)
+	if batch != nil {
+		vars := append(batch.Vars(), s.vars...)
```
nwt, line 23: Batches can be accessed concurrently, so appending to batch.Vars() is racy. Clone it first.
```suggestion
		vars := append(slices.Clone(batch.Vars()), s.vars...)
```

mattnibs: Good catch.

runtime/op/scope.go
```diff
@@ -42,7 +43,7 @@ func (s *exitScope) Pull(done bool) (zbuf.Batch, error) {
 	batch, err := s.parent.Pull(done)
 	if batch != nil {
 		vars := batch.Vars()
-		vars = vars[:len(vars)-s.nvars]
+		vars = slices.Clone(vars[:len(vars)-s.nvars])
```
nwt, line 46: No need to clone here since we're not writing to the array backing vars.

nwt (approved):
:shipit:

PR #4638, opened 2023-06-08
add dictionary encoding to vng

mccanne:
This commit adds dictionary encoding to the vng format where primitive vectors are optionally encoded with dictionaries when their cardinality is small.  Currently, the cardinality is limited to 256 values; we will allow larger dictionaries in the future.

When the dictionary has length 1, the vector is simplified into a "constant" vector and the dictionary and stats are omitted.

### DIFF
--- a/vng/vector/primitive.go
+++ b/vng/vector/primitive.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"sort"
 
 	"github.com/brimdata/zed"
 	"github.com/brimdata/zed/order"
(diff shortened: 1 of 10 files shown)


vng/vector/primitive.go
```diff
@@ -2,37 +2,83 @@ package vector
...
+	}
+	if body != nil {
+		val := zed.NewValue(p.typ, body)
+		if p.min == nil || p.cmp(val, p.min) < 0 {
+			p.min = val
+		}
+		if p.max == nil || p.cmp(val, p.max) > 0 {
+			p.max = val
+		}
+	}
```
nwt, line 65: Given the previous if statement, the condition here is always true.
```suggestion
	val := zed.NewValue(p.typ, body)
	if p.min == nil || p.cmp(val, p.min) < 0 {
		p.min = val
	}
	if p.max == nil || p.cmp(val, p.max) > 0 {
		p.max = val
	}
```

vng/vector/primitive.go
```diff
@@ -2,37 +2,83 @@ package vector
...
+	if p.dict != nil {
+		p.dict[string(body)] += 1
```
nwt, line 67:
```suggestion
		p.dict[string(body)]++
```

vng/vector/primitive.go
```diff
@@ -41,15 +87,88 @@ func (p *PrimitiveWriter) Flush(eof bool) error {
...
+		Max:    p.max,
+	}
+}
+
+func (p *PrimitiveWriter) makeDict() []DictEntry {
+	dict := make([]DictEntry, 0, len(p.dict)+p.hasNull)
+	for key, cnt := range p.dict {
+		dict = append(dict, DictEntry{
+			zed.NewValue(p.typ, zcode.Bytes(key)),
+			uint32(cnt),
```
nwt, line 161: This is already a uint32.
```suggestion
			cnt,
```

vng/vector/primitive.go
```diff
@@ -41,15 +87,88 @@ func (p *PrimitiveWriter) Flush(eof bool) error {
...
+		})
 	}
+	if p.nulls != 0 {
+		dict = append(dict, DictEntry{
+			zed.NewValue(p.typ, nil),
+			p.nulls,
+		})
+	}
+	sortDict(dict, expr.NewValueCompareFn(order.Asc, false))
```
nwt, line 170: Nit: Move sortDict's definition after this method since this is the only call to it.

nwt (approved):

PR #4644, opened 2023-06-08
Allow user ops with no parameters

### LINKED ISSUE #4643: User Op without parameters won't parse
Repro is with Zed commit 1a9b356.

I'm guessing that a valid use case for User Ops may be situations that don't require parameters but are instead just frequently-invoked chunks of pipeline that would benefit from a terse syntax to invoke them. I therefore attempted the following, but found it did not parse.

```
$ zq -version
Version: v1.8.1-24-g1a9b356b

$ cat bar.zed 
op CallMe(): (
  yield "hello"
)
CallMe()

$ echo "null" | zq -z -I bar.zed -
zq: error parsing Zed in bar.zed at line 1, column 12:
op CallMe(): (
       === ^ ===
```

However, if I add a parameter and pass it a value that's never used, that makes it work.

```
$ cat foo.zed 
op CallMe(const param): (
  yield "hello"
)
CallMe("goodbye")

$ echo "null" | zq -z -I foo.zed -
"hello"
```

mattnibs:
Closes #4643

### DIFF
--- a/compiler/parser/parser.peg
+++ b/compiler/parser/parser.peg
@@ -86,7 +86,7 @@ FuncDecl
     }
 
 OpDecl
-  = "op" _ name:IdentifierName __ "(" __ params:Params __ ")" __ ":" __ "(" __ body:Seq __ ")" {
+  = "op" _ name:IdentifierName __ "(" __ params:Params? __ ")" __ ":" __ "(" __ body:Seq __ ")" {
       RETURN(MAP(
         "kind": "OpDecl",
         "name": name,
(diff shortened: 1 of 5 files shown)


compiler/parser/parser.peg
```diff
@@ -98,6 +98,7 @@ Params
   = first:Param rest:(__ "," __ p:Param { RETURN(p) })* {
       RETURN(PREPEND(first, rest))
     }
+  / __ { RETURN(ARRAY()) }
```
nwt, line 101: s/Param/Params?/ would be clearer.

nwt (approved):

PR #4646, opened 2023-06-08
Allow user defined functions with no params

mattnibs:


compiler/parser/parser.peg
```diff
@@ -1211,6 +1211,7 @@ IdentifierNames
   = first:IdentifierName rest:(__ "," __ id:IdentifierName { RETURN(id) })* {
     RETURN(PREPEND(first, rest))
   }
+  / __ { RETURN(ARRAY()) }
```
nwt, line 1214: s/IdentifierNames/IdentifierNames?/ in the FuncDecl rule would be clearer.

mattnibs: Yeah

nwt (approved):

PR #4647, opened 2023-06-09
Remove ZSON formatter error result parameters

nwt:
ZSON formatting function and methods, like zson.FormatValue, are cumbersome to use because they return an error.  (For example, MustFormatValue is used far more than FormatValue.)  But the errors those functions and methods return only concern invalid input, which isn't very useful since they validate only a subset of what zed.Value.Validate does.  Remove their error result parameters, and remove MustFormatValue, which is now unneeded.


mattnibs (approved):
kewl

PR #4655, opened 2023-06-15
Simplify lake manage

mattnibs:
Make changes to the lake manage logic to simplify the logic for compaction:

- Manage now compacts any overlapping data (no methods for reducing write amplification are currently used).
- Instead of timing compaction runs dynamically in accordance to changes to a pool, compact now just executes on a configurable interval.
- Removed index updates for lake manage.
- Change lake manage to a single command with no subcommands.


philrz:
It won't let me put this as an inline review comment since it's not code that was touched in this PR, but should the `[subcommand]` part of the usage be removed or replaced with `[options]` since the subcommands are gone for now?

cmd/zed/manage/ztests/once.yaml
```diff
@@ -0,0 +1,15 @@
+script: |
+  export ZED_LAKE=test1
+  zed init -q
+  zed create -S 10KB -q test1
```
philrz, line 4:
```suggestion
  zed create -q test1
```

In the past I remember we had to use these small object sizes in order to see compaction in action because the heuristic was (something like) that there had to be an accumulation of a certain amount of uncompacted data relative to the object size. Based on the text description attached to your PR and my black box testing of this branch at commit 2cd633a it appears that even with the default object size compaction will be triggered, and so if that's all working as intended, then it seems we might as well test with default object size.

mattnibs: I'm going to add one more test that makes sure a non-overlapping object that is at least half the size of pool threshold is left alone.

mattnibs: Oh wait nevermind, I'll create another pool for the test. Will change to what you suggest.

cmd/zed/manage/command.go
```diff
@@ -1,30 +1,69 @@
...
+	logFlags logflags.Flags
+	once     bool
+	config   lakemanage.Config
 }
 
 func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
-	return &Command{Command: parent.(*root.Command)}, nil
+	c := &Command{Command: parent.(*root.Command)}
+	c.logFlags.SetFlags(f)
+	f.Func("config", "path of manage yaml config file", func(s string) error {
```
nwt, line 32:
```suggestion
	f.Func("config", "path of manage YAML config file", func(s string) error {
```

cmd/zed/manage/lakemanage/branch.go
```diff
@@ -74,26 +50,25 @@ type compactTask struct {
 	log *zap.Logger
 }
 
-func (b *compactTask) run(ctx context.Context, at ksuid.KSUID) (*time.Time, error) {
+func (b *compactTask) run(ctx context.Context) error {
```
nwt, line 53:
```suggestion
func (c *compactTask) run(ctx context.Context) error {
```

nwt (approved):

cmd/zed/manage/lakemanage/branch.go
```diff
@@ -74,26 +50,25 @@ type compactTask struct {
 	log *zap.Logger
```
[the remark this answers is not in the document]
mattnibs: The task logger should have an extra field that has the name of the task (e.g. "compact" or "vector") but since we only got one I'll do this for the time being.

PR #4656, opened 2023-06-15
zed create: Add -use flag

mattnibs:
Add -use flag to zed create that if activated sets the created pool as the current pool for future operations.

### DIFF
--- a/cmd/zed/create/command.go
+++ b/cmd/zed/create/command.go
@@ -5,6 +5,7 @@ import (
 	"flag"
 	"fmt"
 
+	"github.com/brimdata/zed/cli/poolflags"
 	"github.com/brimdata/zed/cmd/zed/root"
 	"github.com/brimdata/zed/lake/data"
 	"github.com/brimdata/zed/order"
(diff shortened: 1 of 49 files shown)


cmd/zed/create/command.go
```diff
@@ -48,9 +50,10 @@ func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
 		seekStride: units.Bytes(data.DefaultSeekStride),
 	}
 	c.thresh = data.DefaultThreshold
+	f.Var(&c.seekStride, "seekstride", "size of seek-index unit for ZNG data, as '32KB', '1MB', etc.")
 	f.Var(&c.thresh, "S", "target size of pool data objects, as '10MB' or '4GiB', etc.")
```
nwt, line 54: Nit: Keep the c.thresh assignment adjacent to the corresponding f.Var call.
```suggestion
	f.Var(&c.seekStride, "seekstride", "size of seek-index unit for ZNG data, as '32KB', '1MB', etc.")
	c.thresh = data.DefaultThreshold
	f.Var(&c.thresh, "S", "target size of pool data objects, as '10MB' or '4GiB', etc.")
```

nwt (approved):

Issue #4658, opened 2023-06-15
Parameterized nested field reference inside User Op


philrz:
Repro is with Zed commit 2314053.

Let's start from this very simple example of outputting the value from a field in an embedded record.

```
$ zq -version
Version: v1.8.1-28-g23140538

$ echo '{foo: {bar: "baz"}}' | zq -Z 'yield foo.bar' -
"baz"
```

If we want to do the same when `this` is passed into a User Op, it's easy enough if we always know the precise field path and can express it in dotted notation or with [indexing](https://zed.brimdata.io/docs/next/language/expressions#indexing).

```
$ cat yield-in-op-1.zed 
op CallMe(...): (
  yield {dotted: foo.bar, indexed: this["foo"]["bar"]}
)
CallMe(this)

$ echo '{foo: {bar: "baz"}}' | zq -Z -I yield-in-op-1.zed -
{
    dotted: "baz",
    indexed: "baz"
}
```

However, we expect users will want to write User Ops when the field name/path is not known in advance and it's passed in from the main scope as a parameter. This currently works fine with a top-level field reference.

```
$ cat yield-in-op-2.zed 
op CallMe(..., const f): (
  yield this[f]
)
CallMe(this, "foo")

$ echo '{foo: {bar: "baz"}}' | zq -Z -I yield-in-op-2.zed -
{
    bar: "baz"
}
```

However, there's not currently a way to reference a nested field path this way. For instance, if I was a user hacking away, I might attempt:

```
$ cat yield-in-op-3.zed 
op CallMe(..., const f): (
  yield this[f]
)
CallMe(this, ["foo"]["bar"])

$ echo '{foo: {bar: "baz"}}' | zq -Z -I yield-in-op-3.zed -
zq: error parsing Zed in yield-in-op-3.zed at line 4, column 21:
CallMe(this, ["foo"]["bar"])
                === ^ ===
```

We've discussed as a team implementing an approach to this using "first class paths".

philrz:
Verified in Zed commit e57e90e.

With the changes in the linked PR #4663, I can now pass in a dotted reference to a nested field or use [field referencing with indexing](https://zed.brimdata.io/docs/next/language/expressions#field-dereference) to retrieve a field's value from inside the user op. Updating the repro steps to align with the revised syntax, we now see:

```
$ zq -version
Version: v1.8.1-37-ge57e90ee

$ cat yield-in-op-revised.zed
op CallMe(f): (
  yield f
)

[...]

Issue #4650, opened 2023-06-13
Primitive values dropped when "this" passed into User Op with spread


philrz:
Repro is with Zed commit 2314053.

The main scope is able to process both primitive and complex values as `this`, e.g.:

```
$ cat kind-in-main.zed
yield {kind: kind(this),ThisWas: this}

$ echo '{foo: "bar"} "hello"' | zq -z -I kind-in-main.zed -
{kind:"record",ThisWas:{foo:"bar"}}
{kind:"primitive",ThisWas:"hello"}
```

However, if I wrap that `yield` inside a User Op and try to pass `this` from main scope into the Op's `this` using the `...` spread operator, the primitive value gets dropped.

```
$ cat kind-in-op.zed 
op CallMe(...): (
  yield {kind: kind(this),ThisWas: this}
)
CallMe(this)

$ echo '{foo: "bar"} "hello"' | zq -z -I kind-in-op.zed -
{kind:"record",ThisWas:{foo:"bar"}}
{kind:"record",ThisWas:{}}
```

Once a user is aware of this limitation, they could work around it by passing the main scope's `this` into a named parameter, e.g.:

```
$ cat kind-in-op-nested.zed
op CallMe(nested): (
  yield {kind: kind(nested),ThisWas: nested}
)
CallMe(this)

$ echo '{foo: "bar"} "hello"' | zq -z -I kind-in-op-nested.zed -
{kind:"record",ThisWas:{foo:"bar"}}
{kind:"primitive",ThisWas:"hello"}
```

Admittedly, the [docs](https://zed.brimdata.io/docs/next/language/statements#sequence-this-value) do call out how the use of the spread operator means that only records will be expected:

> ...user-defined operators may use the spread operator `...` to indicate that the operator expects a record value whose key/values will be expanded as entries in the operator's `this` record value.

However, if we start from the design principle stated in #4649, we may want to avoid having this limitation. When discussing as a team, there were two reactions:

1. If we kept the current behavior, it could be improved by throwing an `error` when such values are dropped rather than its current silence
2. We may instead change to an approach that allows `this` to passed in from the main scope in a way that primitive values don't get dropped

philrz:
Verified in Zed commit e57e90e.

Of the two likely approaches described above, the changes in linked PR #4650 took the latter approach of allowing `this` to be carried through into the User Op such that primitive values are no longer dropped. Updating the repro steps to align with the revised syntax, we now see:

```
$ zq -version
[...]

PR #4663, opened 2023-06-21
Rework user ops

mattnibs:
Rework user ops in the following ways:
- User ops take either a const or path as an argument.
- The this value of the parent is carried through when a user op is invoked.


philrz:
FWIW, I tested out this branch at commit 6fb851c and confirmed it seems to address what's described in #4650. That is, since we're no longer using spread to pass `this` into the user op, now primitive values make it through ok.

```
$ zed -version
Version: v1.8.1-32-g6fb851c9

$ cat kind-in-op-4663.zed
op CallMe(): (
  yield {kind: kind(this),ThisWas: this}
)
CallMe()

$ echo '{foo: "bar"} "hello"' | zq -z -I kind-in-op-4663.zed -
{kind:"record",ThisWas:{foo:"bar"}}
{kind:"primitive",ThisWas:"hello"}
```

philrz:
It looks like #4658 is addressed as well. I can now pass in a dotted reference to a nested field or use [field referencing with indexing](https://zed.brimdata.io/docs/next/language/expressions#field-dereference) and then reference the field ok from inside the user op.

```
$ zq -version
Version: v1.8.1-32-g6fb851c9

$ cat yield-in-op-4663.zed
op CallMe(f): (
  yield f
)

$ echo '{foo: {bar: "baz"}}' | zq -Z -I yield-in-op-4663.zed 'CallMe(foo.bar)' -
"baz"

$ echo '{foo: {bar: "baz"}}' | zq -Z -I yield-in-op-4663.zed 'CallMe(this["foo"]["bar"])' -
"baz"
```

nwt (commented):
The stuff in runtime/op/scope.go is now unused, so you can remove that.

compiler/semantic/op.go
```diff
@@ -1066,42 +1059,64 @@ func (a *analyzer) maybeConvertUserOp(call *ast.Call, seq dag.Seq) (dag.Op, erro
...
+		for i, p := range params {
+			if err := a.scope.DefineAs(p, exprs[i]); err != nil {
+				return nil, err
+			}
+		}
+		if body, err = a.semSeq(decl.ast.Body); err != nil {
+			return nil, err
+		}
 	}
+	// Else we need to
```
nwt, line 1115:
```suggestion
```

runtime/op/ztests/user-no-params.yaml
```diff
@@ -1,10 +0,0 @@
-zed: |
-  op noparams(): (
-    yield "hello"
-  )
-  noparams()
-
-input: "null"
-
-output: |
-  "hello"
```
nwt, line 10: Why delete this?

compiler/semantic/op.go
```diff
@@ -1066,42 +1059,64 @@ func (a *analyzer) maybeConvertUserOp(call *ast.Call, seq dag.Seq) (dag.Op, erro
...
 	// Add references so we can construct the graph of user op calls in order
 	// to later detect cycles.
 	if a.opDecl != nil {
 		a.opDecl.deps = append(a.opDecl.deps, op)
+	} else {
+		// appendOpPath will return an error if the opDecl is already in the
+		// path, which means we have a cycle.
+		if err := a.appendOpPath(op); err != nil {
+			return nil, err
+		}
```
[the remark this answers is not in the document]
mattnibs: great!

runtime/op/ztests/user-no-params.yaml
mattnibs: I suppose it is still useful.

nwt (approved):

PR #4671, opened 2023-06-22
Prefer zed.NewFloat et al. over zed.NewValue

nwt:
Unlike NewValue, the NewFloat functions all return a zed.Value that uses the native representation.

### DIFF
--- a/runtime/expr/cast.go
+++ b/runtime/expr/cast.go
@@ -106,7 +106,7 @@ func (c *casterFloat16) Eval(ectx Context, val *zed.Value) *zed.Value {
 	if !ok {
 		return c.zctx.WrapError("cannot cast to float16", val)
 	}
-	return ectx.NewValue(zed.TypeFloat16, zed.EncodeFloat16(float32(f)))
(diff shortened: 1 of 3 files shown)


runtime/expr/cast.go
```diff
@@ -106,7 +106,7 @@ func (c *casterFloat16) Eval(ectx Context, val *zed.Value) *zed.Value {
 	if !ok {
 		return c.zctx.WrapError("cannot cast to float16", val)
 	}
-	return ectx.NewValue(zed.TypeFloat16, zed.EncodeFloat16(float32(f)))
+	return ectx.CopyValue(zed.NewFloat16(float32(f)))
```
mattnibs, line 109: Why are we copying the value here if zed.NewFloat16 creates a new Value?

nwt: So it'll be allocated by ectx rather than on the heap. This doesn't work the way we want right now because Context is an interface, and that cause its argument to escape to the heap. But I'm planning to fix that by converting Context from an interface to a struct.

mattnibs (approved):

PR #4686, opened 2023-06-27
Add an "Error Handling" section to Shaping docs

### LINKED ISSUE #2776: Notify users of shaper "misses"
brimsec/zq#2369 and its fix in brimsec/zq#2370 describe an example where the shaper currently imports additional fields that are not covered by the shaping config, assigning them inferred types. In that case it resulted in an IP address being stored as a string rather than an `ip` type, which has the follow-on effect of type-specific operations not working (e.g. CIDR matches in this case).

This behavior was last visited in brimsec/zq#2309. Before that change, such additional fields would have been "cropped". After that change, now they're let in with inferred types. If choosing between these two options, indeed the current "infer" approach seems slightly better, but from a user perspective, they _both_ seem sub-optimal: Whether the unexpected fields are lost entirely or stored incorrectly, I imagine _the user will want to know what happened_ so they can adjust their queries, ask us for help, etc. Before brimsec/zq#2298 was merged, the legacy `types.json` approach at least gave the user a heads-up in the form of a warning (`"Unexpected additional field(s) 'tunnel.src_ip' 'tunnel.src_port' 'tunnel.dest_ip' 'tunnel.dest_port' 'tunnel.proto' 'tunnel.depth'. Please email this error message to support@brimsecurity.com."`) which is how we got a heads-up about brimsec/zq#2369 in the first place. But now we've lost that, hence this issue.

It seems the first order of business will be to just have the warnings generated by `zqd` and presented to the user in Brim. I expect the _ultimate_ solution will be when we have a fairly rich shaper UX that will allow the user to react to the warnings themselves, which means them being able to see/modify/apply their own shaper configs. But as long as they're stuck with static shapers that we provide, some kind of notification like we had before feels like it would be an improvement.

philrz:
I recently spent some quality time with Zed's shaping features and found I was pretty happy with how things have improved. In particular the improvements in error handling (such as highlighted in the [recent blog post](https://www.brimdata.io/blog/debugging-data-first-class-errors/) from @jameskerr) definitely help plug some remaining UX gaps. To that end, I went back through some old issues to see if it's appropriate to close some, and #2776 definitely looked like fair game. However, I figured it's worthwhile to augment the docs with some examples that show how to catch shaping errors, hence this PR.

Fixes #2776

### DIFF
--- a/docs/language/shaping.md
+++ b/docs/language/shaping.md
@@ -30,7 +30,7 @@ that currently perform shaping are:
 * [`shape`](functions/shape.md) - apply `cast`, `fill`, and `order`
 
 They all have the same signature, taking two parameters: the value to be
-transformed and a type value for the target type.
+transformed and a [type value](data-types.md) for the target type.
 
 > Another type of transformation that's needed for shaping is renaming fields,
 > which is supported by the [`rename` operator](operators/rename.md).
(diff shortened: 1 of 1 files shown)


docs/language/shaping.md
```diff
@@ -208,6 +208,156 @@ drops the `uid` field after shaping:
...
+  "kind": "dns",
+  "server": {
+    "addr": "10.0.0.100",
+    "port": 53
+  },
+  "client": {
+    "addr": "39 Elm Street",
+    "port": 41772
+  },
+  vlan: "available"
```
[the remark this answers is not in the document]
philrz: Thanks @nwt. ZSON fallback is too forgiving! 🤪

docs/language/shaping.md
```diff
@@ -208,6 +208,156 @@ drops the `uid` field after shaping:
...
+        on: "available"
+    })
+}
+```
+
+Since these error values are nested inside an otherwise healthy record, adding
+[`has_error(this)`](functions/has_error.md) downstream in our Zed pipeline
+could help find or exclude such records.  If the failure to shape _any_ single
+field is considered severe enough to render the entire input record unhealthy,
+[conditional logic](expressions.md#conditional)
```
nwt, line 265:
```suggestion
[a conditional expression](expressions.md#conditional)
```

docs/language/shaping.md
```diff
@@ -208,6 +208,156 @@ drops the `uid` field after shaping:
...
+        } (=socket),
+        vlan: error({
+            message: "cannot cast to uint16",
+            on: "available"
+        })
+    }
+}) (error({msg:string,original:{kind:string,server:{addr:string,port:int64},client:{addr:string,port:int64},vlan:string},shaped:{kind:string,client:{addr:error({message:string,on:string}),port:port=uint16},server:socket={addr:ip,port:port},vlan:error({message:string,on:string})}}))
+```
+
+If you require awareness about changes made by the core shaping functions that
```
nwt, line 316: I don't think "core" clarifies anything here.
```suggestion
If you require awareness about changes made by the shaping functions that
```

nwt (approved):

PR #4694, opened 2023-07-03
Move Zeek docs to Integrations area

philrz:
I've spotted some material changes I'd like to make to our docs about how Zed works with Zeek. These have been in living in a [separate spot](https://github.com/brimdata/zed/tree/main/zeek) away from the other Zed docs, which I believe was motivated by an instinct to create some separation from tooling's prior Zeek-centric focus. However, we now have an Integrations area in the Zed docs where it feels like these could safely live such that they'd benefit from the better rendering in Docusaurus and they'd also be easier for readers to find but still be clearly separate from the core Zed mission. This PR just takes care of moving them to the new location and updating the formatting/links. Once this merges I'll put up a follow-on PR with actual material changes to the content.

If you want to see how it looks rendered, I've put a copy of on a personal staging site at https://spiffy-gnome-8f2834.netlify.app/docs/next/integrations/zeek.


nwt:
`zio/zeekio/` is all about Zeek TSV I/O, so it doesn't feel like the right place for `shaper.zed` and its test. How about embedding them in `shaping-zeek-ndjson.md` as a fenced code block (for the shaper) and mdtest?

philrz:
@nwt: Ok, I've followed your recommendation and moved the `shaper.zed` into the markdown and created an mdtest to replicate what was formerly in the ztest. As we discussed offline, that iteration of the `shaper.zed` is particularly huge but I have plans to shorten it significantly in the changes I have coming up along with a wider overhaul of that doc about shaping Zeek, so try not to get too caught up in its present contents.

nwt (approved):
:shipit:

PR #4697, opened 2023-07-04
Extract Zeek shaper from markdown

philrz:
After #4694 moved the reference shaper for Zeek into the markdown docs, I realized this performance comparison script was depending on the shaper file at its prior location. This updates the script to extract the shaper from the markdown into a temporary file so it can be used as it was before.

### DIFF
--- a/scripts/perf-compare.sh
+++ b/scripts/perf-compare.sh
@@ -6,6 +6,9 @@ set -eo pipefail
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 cd "$SCRIPT_DIR" || exit 1
 
+SHAPER=$(mktemp)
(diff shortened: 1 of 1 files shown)


scripts/perf-compare.sh
```diff
@@ -6,6 +6,9 @@ set -eo pipefail
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 cd "$SCRIPT_DIR" || exit 1
 
+SHAPER=$(mktemp)
+sed -e '1,/```mdtest-input shaper.zed/d' ../docs/integrations/zeek/shaping-zeek-ndjson.md | sed -e '/```/,$d' > "$SHAPER"
```
nwt, line 10: Nits: All-caps is for environment variables and shouting. And just invoke sed once.
```suggestion
shaper=$(mktemp)
sed -e '1,/```mdtest-input shaper.zed/d' -e '/```/,$d' ../docs/integrations/zeek/shaping-zeek-ndjson.md > "$shaper"
```

nwt (approved):

PR #4726, opened 2023-07-24
Add vector service endpoints

mattnibs:


api/client/connection.go
```diff
@@ -432,6 +432,25 @@ func (c *Connection) Vacuum(ctx context.Context, pool, revision string, dryrun b
 	return res, err
 }
 
+func (c *Connection) VectorPost(ctx context.Context, pool, revision string, objectIDs []ksuid.KSUID, message api.CommitMessage) (api.CommitResponse, error) {
+	return c.vector(ctx, pool, revision, objectIDs, message, http.MethodPost)
+}
+
+func (c *Connection) VectorDelete(ctx context.Context, pool, revision string, objectIDs []ksuid.KSUID, message api.CommitMessage) (api.CommitResponse, error) {
```
nwt, line 439: Call these AddVectors and DeleteVectors so they match the lake/api.Interface? That's what we've done with the all the other methods.

api/client/connection.go
```diff
@@ -432,6 +432,25 @@ func (c *Connection) Vacuum(ctx context.Context, pool, revision string, dryrun b
...
+	return c.vector(ctx, pool, revision, objectIDs, message, http.MethodDelete)
+}
+
+func (c *Connection) vector(ctx context.Context, pool, revision string, objectIDs []ksuid.KSUID, message api.CommitMessage, method string) (api.CommitResponse, error) {
```
nwt, line 443: Nit: Call this doVector so the name is verby.

service/ztests/vector.yaml
```diff
@@ -0,0 +1,34 @@
...
+  zed vector add -q $id
+  zed query -Z 'from POOL@main:vectors | drop id'
+  echo ===
+  zed vector delete -q $id
+  zed query -Z 'from POOL@main:vectors | drop id'
+  echo ===
+
+inputs:
+  - name: service.sh
+    source: service.sh
```
nwt, line 15: Nit: Don't need this.
```suggestion
```

service/ztests/vector.yaml
```diff
@@ -0,0 +1,33 @@
+script: |
```
nwt, line 1: We should add the local lake equivalent of this test in lake/ztests/vector.yaml so we have coverage for the lake/api/local.go change.

nwt (commented):
We should add this to docs/lake/api.md.

service/ztests/vector.yaml
mattnibs: This test is an exact copy of lake/ztests/vector.yaml

nwt: My apologies. I totally missed that.

nwt (approved):
I'm OK merging this without the docs/lake/api.md update if you'd rather do that later.

PR #4728, opened 2023-07-24
zed manage: Change default behavior

mattnibs:
Change default behavior of zed manage so the default behavior runs compact once and exits. The -monitor flag can be used to continuously monitor and update pools.

### DIFF
--- a/cmd/zed/manage/command.go
+++ b/cmd/zed/manage/command.go
@@ -15,15 +15,15 @@ import (
 var Cmd = &charm.Spec{
 	Name:  "manage",
 	Usage: "manage",
-	Short: "continuously run compaction and other maintenance tasks on a lake",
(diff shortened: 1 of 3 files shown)


cmd/zed/manage/command.go
```diff
@@ -15,14 +15,14 @@ import (
...
+	Short: "run compaction and other maintenance tasks on a lake",
 	New:   New,
 }
 
 type Command struct {
 	*root.Command
 	logFlags logflags.Flags
-	once     bool
+	monitor  bool
 	config   lakemanage.Config
```
nwt, line 26: Nit: Swap these to match the order in which New initializes them.
```suggestion
	config   lakemanage.Config
	monitor  bool
```

nwt (approved):

PR #4736, opened 2023-08-01
Use sync/atomic.Pointer in zed.Context

nwt:
Context.Missing, Quiet, and StringTypeError are relatively expensive because they synchronize via Context.mu, and mutex overhead is relatively high.  Change those methods to synchronize using atomic.Pointer, which has lower overhead than a mutex.

### DIFF
--- a/context.go
+++ b/context.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"sort"
 	"sync"
+	"sync/atomic"
 	"unicode/utf8"
 
 	"github.com/brimdata/zed/zcode"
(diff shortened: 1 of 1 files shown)


mattnibs (approved):

philrz:
@nwt: I've spotted what look like some statistically significant deltas in the Autoperf trend that correlate with this PR. I know the changes here were intended to approve performance so this is probably not a surprise, but figured I'd report the specifics to confirm that they line up with expectations.

The first is marked by the right-side red/vertical dotted line shown in this screenshot where the time to run these particular queries came down from ~6 minutes to ~4.5 minutes. It also looks like memory usage improved, e.g., "sys" memory down from ~95 MB to ~85 MB. The specific queries:

```
host_p==36 or port_num==36 or id.orig_p==36 or id.resp_p==36 or data_channel.resp_p==36 or p==36 or bound_p==36 | count() by _path | sort _path
host_p==80 or port_num==80 or id.orig_p==80 or id.resp_p==80 or data_channel.resp_p==80 or p==80 or bound_p==80 | count() by _path | sort _path
```

image

The other visible delta shows up on the memory side but not run time. These queries look like:

```
notice==true | count()
timedout==false | count()
```

image

PR #4756, opened 2023-08-28
Compact: write vectors

mattnibs:
Allow writing of vectors when compacting objects

### DIFF
--- a/runtime/exec/compact.go
+++ b/runtime/exec/compact.go
@@ -13,7 +13,7 @@ import (
 	"github.com/segmentio/ksuid"
 )
 
-func Compact(ctx context.Context, lk *lake.Root, pool *lake.Pool, branchName string, objectIDs []ksuid.KSUID, author, message, info string) (ksuid.KSUID, error) {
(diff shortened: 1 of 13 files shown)


runtime/exec/compact.go
```diff
@@ -39,6 +39,9 @@ func Compact(ctx context.Context, lk *lake.Root, pool *lake.Pool, branchName str
 	slicer := meta.NewSlicer(lister, zctx)
 	puller := meta.NewSequenceScanner(octx, slicer, pool, nil, nil, nil)
 	w := lake.NewSortedWriter(ctx, zctx, pool)
+	if writeVectors {
+		w.EnableVectorWrite()
+	}
```
nwt, line 44: Nit: Just pass this flag to NewSortedWriter?
```suggestion
	w := lake.NewSortedWriter(ctx, zctx, pool, writeVectors)
```

nwt (approved):

runtime/exec/compact.go
mattnibs: ok

PR #4758, opened 2023-09-01
zed: Default app data location

mattnibs:
If no lake location is specified, zed will use a per-os default
location. Also if the default location is used, zed will first check if
there's a service on 9867 and opt to connect to that instead of
operating locally.

### DIFF
--- a/cli/lakeflags/datadir.go
+++ b/cli/lakeflags/datadir.go
@@ -1,0 +1,36 @@
+package lakeflags
+
+import (
+	"os"
+	"path/filepath"
+	"runtime"
+)
+
+var defaultDataDir string
+
+func init() {
+	defaultDataDir = getDefaultDataDir()
+}
+
+// getDefaultDataDir returns the default data directory for the current user.
+// Derived from https://github.com/btcsuite/btcd/blob/master/btcutil/appdata.go
+func getDefaultDataDir() string {
+	// Resolve the XDG data home directory if set.
+	if xdgDataHome := os.Getenv("XDG_DATA_HOME"); xdgDataHome != "" {
+		return filepath.Join(xdgDataHome, "zed")
+	}
+	if runtime.GOOS == "windows" {
+		if appData := os.Getenv("LOCALAPPDATA"); appData != "" {
+			return filepath.Join(appData, "zed")
+		}
+	}
+	if homeDir, _ := os.UserHomeDir(); homeDir != "" {
+		// Follow the XDG spec which states:
+		// If $XDG_DATA_HOME is either not set or empty, a default equal to
+		// $HOME/.local/share should be used.
+		return filepath.Join(homeDir, ".local", "share", "zed")
+	}
+	// Return an empty string which will cause an error if a default data
+	// directory cannot be found.
+	return ""
+}
--- a/cli/lakeflags/datadir.go
+++ b/cli/lakeflags/datadir.go
@@ -6,12 +6,6 @@ import (
 	"runtime"
 )
 
-var defaultDataDir string
-
-func init() {
-	defaultDataDir = getDefaultDataDir()
-}
-
 // getDefaultDataDir returns the default data directory for the current user.
 // Derived from https://github.com/btcsuite/btcd/blob/master/btcutil/appdata.go
 func getDefaultDataDir() string {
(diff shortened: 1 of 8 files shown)


cli/lakeflags/datadir.go
```diff
@@ -0,0 +1,51 @@
...
+			return filepath.Join(appData, "zed")
+		}
+	default:
+		if homeDir != "" {
+			// For linux and windows follow the XDG spec which states:
+			// If $XDG_DATA_HOME is either not set or empty, a default equal to
+			// $HOME/.local/share should be used.
+			return filepath.Join(homeDir, ".local", "share", "zed")
+		}
+	}
```
nwt, line 47: Maybe simplify this?
```suggestion
	if runtime.GOOS == "windows" {
		if appData := os.Getenv("LOCALAPPDATA"); appData != "" {
			return filepath.Join(appData, "zed")
		}
	}
	if homeDir, _ := os.UserHomeDir(); homeDir != "" {
		// Follow the XDG spec which states:
		// If $XDG_DATA_HOME is either not set or empty, a default equal to
		// $HOME/.local/share should be used.
		return filepath.Join(homeDir, ".local", "share", "zed")
	}
```

philrz:
@mattnibs: I'd remembered you indicating this branch was maybe finished, so I just did a retest of the branch at e6afe28 and it's still not making sense to me. For instance If I don't have `ZED_LAKE` or `XDG_DATA_HOME` set, I see:

```
$ zed -version
Version: v1.10.0-2-ge6afe283

$ zed serve
serve command available for local lakes only

$ zed init
lake created: /Users/phil/.local/share/zed

$ zed serve
serve command available for local lakes only

$ zed ls
Post "http://localhost:9867/query?ctrl=T": dial tcp [::1]:9867: connect: connection refused

$ ls -l $HOME/.local/share/zed
total 8
drwxr-xr-x  4 phil  staff  128 Oct  2 09:40 index_rules
-rw-r--r--  1 phil  staff   53 Oct  2 09:40 lake.zng
drwxr-xr-x  4 phil  staff  128 Oct  2 09:40 pools

$ zed -lake $HOME/.local/share/zed ls
Post "http://localhost:9867/query?ctrl=T": dial tcp [::1]:9867: connect: connection refused
```

And even if I set `XDG_DATA_HOME` to a specific path, the `zed init` is still the only command that seems to do what I'd expected.

mattnibs:
@philrz I marked it finished but it is not, I'm still finding bugs in it.

mattnibs:
@philrz et al this pr is now ready for review.

philrz:
Thanks @mattnibs. Indeed, the latest changes do seem to address much of what I showed in that prior comment. However, testing with commit b0e2d08 that's the current tip of this branch, the attached video shows something that's still not making sense to me. With none of the environment variables set, a `zed ls` gives me hints about the lack of a lake at the default location, which makes sense. However, once I launch Zui and it starts a lake service on `http://localhost:9867`, repeating `zed ls` still returns the same error. The spirit in the comments I see in the PR make me think the CLI tools would check for & find the running lake service at that point.

https://github.com/brimdata/zed/assets/5934157/143ca57f-78e1-4da4-9437-49c22358bcff

cli/lakeflags/flags.go
```diff
@@ -95,3 +91,22 @@ func (l *Flags) URI() (*storage.URI, error) {
...
+// the defaultDataDir, it first checks if a zed service is running on
+// localhost:9867 and if so uses http://localhost:9867 as the lake location.
+func (l *Flags) ClientURI() (*storage.URI, error) {
+	u, err := l.URI()
+	if err != nil {
+		return nil, err
+	}
+	if u.String() == storage.MustParseURI(defaultDataDir).String() && localServer() {
+		u = storage.MustParseURI("http://localhost:9867")
+	}
```
nwt, line 105: @mattnibs: This logic doesn't distinguish between not specifying a lake and specifying a lake at the same location as defaultDataDir. To make that more concrete, if `zed serve` is listening at localhost:9867 and I run `zed -lake $HOME/.local/share/zed ls`, it'll connect to `zed serve` even though I asked it to use $HOME/.local/share/zed. I don't think it should.

We could fix that by not assigning a default value to l.lake and then calling localServer and getDefaultLake only when l.lake isn't set.

mattnibs: Good call, let's do that. Will fix.

mattnibs: Actually that's not going to work since it's going to override it if someone does `zed ls -lake ''`.

mattnibs: @nwt added back lackSpecified, ready for another review.

nwt (approved):

PR #4761, opened 2023-09-03
docs: add inverse containment logic example

iloveitaly:


docs/language/operators/where.md
```diff
@@ -60,6 +60,18 @@ echo '1 2 3 4' | zq -z 'where this in [1,4]' -
...
+ echo '
+{status: "active", num: 1}
+{status: "inactive", num: 2}
+{status: "ready", num: 3}
+' | zq 'where ! (status in ["active", "ready"])' -
+```
+=>
+```mdtest-output
+{status:"inactive",num:2}
+```
```
philrz, line 74: Thanks @iloveitaly! A few suggested changes:

1. Fixed a typo: login -> logic
2. Dropped the leading space before `echo`. The test tools actually interpreted the leading space as if it was part of the name of the binary to run, which caused it to fail in CI.
3. While I like the verbose example I sent you on Slack and you echoed back here [ 😄  ], I know the rest of the Zed team tends to prefer the more terse examples, so I morphed it to be an inverse of the test that came right before it. I figure that does a decent job of isolating the what-it-takes-to-flip-the-logic essence.

```suggestion
_A filter with inverse containment logic_
```mdtest-command
echo '1 2 3 4' | zq -z 'where ! (this in [1,4])' -
``` 
=>
```mdtest-output
2
3
```
```

iloveitaly:
looks great, thanks!

iloveitaly:
fixed!

mattnibs (approved):

Issue #4215, opened 2022-11-16
Failed response over API looks like success


philrz:
Repro is with Zed commit 0066c8a.

I was tinkering in a Brim branch https://github.com/brimdata/brim/tree/parquet-export to add Parquet as an **Export** format. It works fine for the happy path, but the backend currently won't create Parquet output when the query response has one or more of the following:

1. Multiple shapes
2. A `union` type

With the Zed CLI tooling, this is well-reported to the user. The following examples use test data [sample.zng](https://github.com/brimdata/brim/blob/main/src/test/shared/data/sample.zng?raw=true) from the Brim repo.

```
$ zq -f parquet sample.zng > /dev/null
Parquet output requires uniform records but multiple types encountered (consider 'fuse')

$ echo $?
1

$ zq -f parquet 'fuse' sample.zng > /dev/null
union type unsupported

$ echo $?
1

$ curl http://localhost:9867/version
{"version":"v1.2.0-132-g0066c8aa"}

$ zed create foo
pool created: foo 2HeFzcYxjPT6hCQ7oZ54rUu7MPa

$ zed -use foo load sample.zng 
(2/1) 4910B/4910B 4910B/s 100.00%
2HeG09yuVGPz3gQg5d0c4yUC8ru committed

$ zed query -f parquet 'from foo' > /dev/null
Parquet output requires uniform records but multiple types encountered (consider 'fuse')

$ echo $?
1

$ zed query -f parquet 'from foo | fuse' > /dev/null
union type unsupported

$ echo $?
1
```

However, If I attempt those queries over the API (as the Brim app would need to do), the errors seem to be masked by the fact that the response comes back as HTTP 200.

```
$ curl -v --fail-with-body -X POST --output /dev/null -H 'Accept: application/x-parquet' -H 'Content-Type: application/json' http://localhost:9867/query -d '{"query":"from foo"}'

Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying 127.0.0.1:9867...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0* Connected to localhost (127.0.0.1) port 9867 (#0)
> POST /query HTTP/1.1
> Host: localhost:9867
> User-Agent: curl/7.84.0
> Accept: application/x-parquet
> Content-Type: application/json
> Content-Length: 20
> 
[...]

PR #4763, opened 2023-09-04
service: Log query runtime errors

mattnibs:
Part of #4215

### DIFF
--- a/service/handlers.go
+++ b/service/handlers.go
@@ -88,6 +88,7 @@ func handleQuery(c *Core, w *ResponseWriter, r *Request) {
 		select {
 		case <-timer.C:
 			if err := writer.WriteProgress(meter.Progress()); err != nil {
+				w.Logger.Warn("Error writing progress", zap.Error(err))
 				writer.WriteError(err)
 				return
 			}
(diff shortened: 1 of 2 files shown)


service/handlers.go
```diff
@@ -88,19 +88,22 @@ func handleQuery(c *Core, w *ResponseWriter, r *Request) {
 		select {
 		case <-timer.C:
 			if err := writer.WriteProgress(meter.Progress()); err != nil {
+				w.Logger.Warn("Error Writing Progress", zap.Error(err))
```
nwt, line 91: Capitalize just the first word.
```suggestion
				w.Logger.Warn("Error writing progress", zap.Error(err))
```

service/ztests/query-runtime-error-log.yaml
```diff
@@ -0,0 +1,20 @@
+script: |
+  LAKE_EXTRA_FLAGS='-log.path=service.log' source service.sh
+  zed create -use -q test
+  echo '{ts:1,union:"hi"((int64,string))}' | zed load -q -
+  curl -o /dev/null -H 'Accept: application/x-parquet' -d '{"query":"from test"}' $ZED_LAKE/query 
+  zq -Z 'cut level, msg, error' service.log
+
+
```
nwt, line 8:
```suggestion

```

service/ztests/query-runtime-error-log.yaml
```diff
@@ -0,0 +1,20 @@
+script: |
+  LAKE_EXTRA_FLAGS='-log.path=service.log' source service.sh
+  zed create -use -q test
+  echo '{ts:1,union:"hi"((int64,string))}' | zed load -q -
```
nwt, line 4: Nit: You can use a simpler input here (though it'll give you a different error).
```suggestion
  echo {} | zed load -q -
```

mattnibs: Fine by me as long as we get the error.

PR #4766, opened 2023-09-07
Query Status endpoint is GET, not POST

philrz:
I bumped into this while following the user-facing API docs during verification.


mattnibs (approved):
Whoop, good catch.

PR #4768, opened 2023-09-08
Allow "not" token in expressions

### LINKED ISSUE #4000: inconsistency between docs and code regarding "not" in expressions
This issue was pointed out by a community member in a [Slack thread](https://brimdata.slack.com/archives/CTSMAK6G7/p1657900246373639).

The docs say `not true` is a valid expression:  
https://github.com/brimdata/zed/blob/3b24a114cabc32c5a306defd7f582ba967d43786/docs/language/overview.md#L861-L868

But the parser says it isn't (though `! true`, which isn't documented, is):  
https://github.com/brimdata/zed/blob/3b24a114cabc32c5a306defd7f582ba967d43786/compiler/parser/parser.peg#L683-L687

What we should do:

*   [ ] Update the versioned docs in https://github.com/brimdata/zed-docs-site/tree/main/versioned_docs/ so they're accurate
*   [ ] Decide what we want in the language
*   [ ] Update docs and code to reflect that

mattnibs:
Closes #4000

### DIFF
--- a/compiler/parser/parser.peg
+++ b/compiler/parser/parser.peg
@@ -718,7 +718,7 @@ MultiplicativeExpr
 MultiplicativeOperator = ("*" / "/" / "%") { RETURN(TEXT) }
 
 NotExpr
-  = "!" __ e:NotExpr {
+  = (NotToken _ / "!" __) e:NotExpr {
         RETURN(MAP("kind": "UnaryExpr", "op": "!", "operand": e))
     }
   / NegationExpr
--- a/docs/language/expressions.md
+++ b/docs/language/expressions.md
@@ -90,7 +90,7 @@ produces
 
 ## Logic
 
-The keywords `and`, `or`, and `not` or `!` perform logic on operands of type `bool`.
+The keywords `and`, `or`, `not`, and `!` perform logic on operands of type `bool`.
 The binary operators `and` and `or` operate on Boolean values and result in
 an error value if either operand is not a Boolean.  Likewise, `not` (and its
 equivalent `!`) operates on its unary operand and results in an error if its
--- a/docs/language/expressions.md
+++ b/docs/language/expressions.md
@@ -90,12 +90,12 @@ produces
 
 ## Logic
 
-The keywords `and`, `or`, and `not` perform logic on operands of type `bool`.
+The keywords `and`, `or`, and `not` or `!` perform logic on operands of type `bool`.
 The binary operators `and` and `or` operate on Boolean values and result in
-an error value if either operand is not a Boolean.  Likewise, `not` operates
-on its unary operand and results in an error if its operand is not type `bool`.
-Unlike many other languages, non-Boolean values are not automatically converted to
-Boolean type using "truthiness" heuristics.
+an error value if either operand is not a Boolean.  Likewise, `not` (and its
+equivalent `!`) operates on its unary operand and results in an error if its
+operand is not type `bool`. Unlike many other languages, non-Boolean values are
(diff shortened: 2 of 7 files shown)


compiler/parser/parser.peg
```diff
@@ -721,6 +721,9 @@ NotExpr
   = "!" __ e:NotExpr {
         RETURN(MAP("kind": "UnaryExpr", "op": "!", "operand": e))
     }
+  / NotToken _ e:NotExpr {
+        RETURN(MAP("kind": "UnaryExpr", "op": "!", "operand": e))
+    }
```
nwt, line 726: Nit: Do this the same way as the SearchFactor rule.
```suggestion
  = (NotToken _ / "!" __) e:NotExpr {
        RETURN(MAP("kind": "UnaryExpr", "op": "!", "operand": e))
    }
```

runtime/expr/ztests/unary-not.yaml
```diff
@@ -0,0 +1,16 @@
...
+input: |
+  {status: "active"}
+  {status: "inactive"}
+
+output: |
+  [false,false]
+  [true,true]
+
+
+
```
nwt, line 16: Nits: Could simplify this.  And delete the trailing empty lines.
```yaml
zed: |
  yield [!this, not this]

input: |
  false
  true

output: |
  [true,true]
  [false,false]
```

nwt (commented):
Please update `docs/language/expressions.md` and `docs/language/search-expressions.md` to reflect the fact that both "!" and "not" are accepted.

philrz:
@mattnibs: I reviewed the docs changes and they read fine. 👍

I suppose it's a separate topic, but I'd personally be in favor of deprecating one approach (probably deprecate `!` so we can start working toward `not` being the only way) which would make the docs even simpler. The fact we support `and` and `or` but not `&&` and `||`, but we _do_ support both `not` and `!` feels a little strange. But other things feel more important so I'm fine merging what we've already got here and banking the improvement.

docs/language/expressions.md
```diff
@@ -90,12 +90,12 @@ produces
 
 ## Logic
 
-The keywords `and`, `or`, and `not` perform logic on operands of type `bool`.
+The keywords `and`, `or`, and `not` or `!` perform logic on operands of type `bool`.
```
nwt, line 93:
```suggestion
The keywords `and`, `or`, `not`, and `!` perform logic on operands of type `bool`.
```

docs/language/search-expressions.md
```diff
@@ -293,9 +293,10 @@ is not.
 ### Boolean Logic
 
 Search terms may be combined into boolean expressions using logical operators
-`and`, `or` and `not`.  `and` may be elided; i.e., concatenation of search terms
-is a logical `and`.  `not` has highest precedence and `and` has precedence over
-`or`.  Parentheses may be used to override natural precedence.
+`and`, `or` and `not` or `!`.  `and` may be elided; i.e., concatenation of
```
nwt, line 296:
```suggestion
`and`, `or`, `not`, and `!`.  `and` may be elided; i.e., concatenation of
```

nwt (approved):

mattnibs:
> I suppose it's a separate topic, but I'd personally be in favor of deprecating one approach (probably deprecate ! so we can start working toward not being the only way) which would make the docs even simpler. The fact we support and and or but not && and ||, but we do support both not and ! feels a little strange. But other things feel more important so I'm fine merging what we've already got here and banking the improvement.

That's a good point, one would be better.

PR #4770, opened 2023-09-08
join: Add automatic sort

mattnibs:
Add functionality to join operator to automatically sort any unsorted input.

### DIFF
--- a/runtime/op/join/join.go
+++ b/runtime/op/join/join.go
@@ -10,6 +10,7 @@ import (
 	"github.com/brimdata/zed/pkg/field"
 	"github.com/brimdata/zed/runtime/expr"
 	"github.com/brimdata/zed/runtime/op"
+	"github.com/brimdata/zed/runtime/op/sort"
 	"github.com/brimdata/zed/zbuf"
 	"github.com/brimdata/zed/zio"
 )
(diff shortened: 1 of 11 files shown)


runtime/op/join/join.go
```diff
@@ -33,11 +34,25 @@ type Op struct {
...
+		if err != nil {
+			return nil, err
+		}
+	}
+	if !rightSorted || leftOrder != rightOrder {
+		right, err = sort.New(octx, right, []expr.Evaluator{rightKey}, rightOrder, true)
+		if err != nil {
+			return nil, err
+		}
+	}
```
nwt, line 55: Something's wrong here.

Suppose you create two files.
```console
$ cat <<EOF >a.zson
{a:1}
{a:2}
EOF
$ cat <<EOF >b.zson
{b:2}
{b:1}
EOF
```

Joining those files on `a=b` should always produce two output values.  But in these cases it doesn't.
```console
$ zq 'file a.zson | join (file b.zson | sort -r b) on a=b'
{a:2}
$ zq 'file a.zson | sort a | join (file b.zson | sort -r b) on a=b'
{a:2}
$ zq 'file a.zson | sort -r a | join (file b.zson) on a=b'
{a:1}
$ zq 'file a.zson | sort -r a | join (file b.zson | sort b) on a=b'
{a:1}
```

mattnibs: This is v strange, thanks for digging into this.

mattnibs: Actually a much better test as well will use this.

nwt (commented):
You should update the Description section of docs/language/operators/join.md so that it doesn't say inputs must be sorted.

nwt (commented):
And docs/tutorial/join.md also contains references to sorting that should be updated.

nwt (approved):
:shipit: when CI is green.

philrz (approved):
Docs changes look good too! 👍

PR #4773, opened 2023-09-19
docs: cut with default value

iloveitaly:


philrz (approved):
There's a couple changes I might like to make here, but I know from experience that GitHub's PR review "suggestions" workflow does not play nicely with these kinds of diffs (the multiple levels of triple backticks seem to confuse it) so I'll just go ahead and merge this and put up additional changes in my own PR. Thanks @iloveitaly!

iloveitaly:
Perfect, sounds like a plan!

PR #4784, opened 2023-09-27
Remove unnecessary sorts before joins in ztests

nwt:
Remove sort operators before joins in ztests that are unnecessary now that join automatically sorts its inputs.

compiler/ztests/join-desc.yaml is left unmodified since it is explicitly testing inputs with descending order.

zfmt/ztests/decls.yaml is modified for consistency, even though removing the sorts affects the output (which is formatted Zed, not query results).

### DIFF
--- a/runtime/op/join/ztests/kinds.yaml
+++ b/runtime/op/join/ztests/kinds.yaml
@@ -1,12 +1,12 @@
 script: |
   echo === ANTI ===
-  zq -z 'anti join (file B.zson) on a=b' A.zson
+  zq -z 'anti join (file B.zson) on a=b | sort a' A.zson
   echo === LEFT ===
-  zq -z 'left join (file B.zson) on a=b hit:=sb' A.zson
+  zq -z 'left join (file B.zson) on a=b hit:=sb | sort a' A.zson
   echo === INNER ===
-  zq -z 'inner join (file B.zson) on a=b hit:=sb' A.zson
+  zq -z 'inner join (file B.zson) on a=b hit:=sb | sort a' A.zson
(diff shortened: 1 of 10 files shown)


runtime/op/join/ztests/kinds.yaml
```diff
@@ -1,12 +1,12 @@
...
-  zq -z 'anti join (file B.zson) on a=b | sort a' A.zson
+  zq -z 'anti join (file B.zson) on a=b' A.zson
   echo === LEFT ===
-  zq -z 'left join (file B.zson) on a=b hit:=sb | sort a' A.zson
+  zq -z 'left join (file B.zson) on a=b hit:=sb' A.zson
   echo === INNER ===
-  zq -z 'inner join (file B.zson) on a=b hit:=sb | sort a' A.zson
+  zq -z 'inner join (file B.zson) on a=b hit:=sb' A.zson
   echo === RIGHT ===
-  zq -z 'right join (file C.zson) on b=c hit:=sb | sort c' B.zson
```
philrz, line 9: While reviewing, I happened to notice that the original actually looks broken here, since the `| sort` are being tacked on _after_ the `join`. It looks like it only worked consistently because the inputs happened to be pre-sorted by the respective keys.

philrz (approved):

mattnibs (approved):

runtime/op/join/ztests/kinds.yaml
nwt: Well spotted! I'll back out the change to this file since these sorts aren't before the join.

PR #4786, opened 2023-09-28
Add hint to init command if no local lake exists

mattnibs:


cmd/zed/ztests/init-hint.yaml
```diff
@@ -0,0 +1,11 @@
+script: |
+  export ZED_LAKE=path/to/zed
+  ! zed ls
```
nwt, line 3:
```suggestion
  ! zed -lake=path/to/zed ls
```

nwt (approved):

cmd/zed/ztests/init-hint.yaml
mattnibs: Fair.

PR #4787, opened 2023-09-28
zed use: print lake with no args

mattnibs:
Change zed use to also print the current lake location if no args are given.

### DIFF
--- a/cmd/zed/use/command.go
+++ b/cmd/zed/use/command.go
@@ -26,7 +26,8 @@ The use command is like "git checkuout" but there is no local copy of
 the lake data.  Rather, the local HEAD state influences commands as
 they access the lake.
 
-With no argument, use prints the working pool and branch.
+With no argument, use prints the working pool and branch as well as the
+location of the current lake.
 
 With an argument of the form "pool", use sets the working pool as indicated
 and the working branch to "main".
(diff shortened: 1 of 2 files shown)


cmd/zed/use/command.go
```diff
@@ -26,7 +26,8 @@ The use command is like "git checkuout" but there is no local copy of
 the lake data.  Rather, the local HEAD state influences commands as
 they access the lake.
 
-With no argument, use prints the working pool and branch.
+With no argument, use prints the working pool and branch as well of the
```
nwt, line 29:
```suggestion
With no argument, use prints the working pool and branch as well as the
```

lake/ztests/use.yaml
```diff
@@ -20,14 +20,16 @@ outputs:
...
-
   - name: stderr
     data: |
       default pool and branch unset
       default pool unset
       "branch": branch not found
+  - name: no-args-use.txt
+    regexp: |
+      HEAD at POOL@main
+      Lake: file.*/test
```
nwt, line 35: I think it looks funny to use both "at" and ":".

mattnibs: I don't think either is right but I'll go your way for now.

nwt (approved):

PR #4789, opened 2023-09-29
Fix panic compiling record expression with duplicate field

### LINKED ISSUE #3365: compileRecordExpr should check for dup columns
compileRecordExpr should check for dup columns at compile time so the runtime won't panic on a bad TypeRecord lookup.

nwt:
Closes #3365.

### DIFF
--- a/compiler/kernel/expr.go
+++ b/compiler/kernel/expr.go
@@ -13,6 +13,7 @@ import (
 	"github.com/brimdata/zed/runtime/op/traverse"
 	"github.com/brimdata/zed/zbuf"
 	"github.com/brimdata/zed/zson"
+	"golang.org/x/exp/slices"
 	"golang.org/x/text/unicode/norm"
 )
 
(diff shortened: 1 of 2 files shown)


compiler/kernel/expr.go
```diff
@@ -397,6 +398,9 @@ func (b *Builder) compileRecordExpr(record *dag.RecordExpr) (expr.Evaluator, err
 	for _, elem := range record.Elems {
 		switch elem := elem.(type) {
 		case *dag.Field:
+			if slices.ContainsFunc(elems, func(r expr.RecordElem) bool { return r.Name == elem.Name }) {
+				return nil, fmt.Errorf("record expression: %w", &zed.DuplicateFieldError{Name: elem.Name})
+			}
```
mattnibs, line 403: I think this check should go in semantic check rather than kernel

mattnibs (approved):

PR #4795, opened 2023-10-04
Support dynamic paths for put and cut op

### LINKED ISSUE #4555: Zed shorthand to create a record field name from string
At the time of the filing of this issue, Zed is at commit 1874aeb.

The attached [scan.json.gz](https://github.com/brimdata/zed/files/11348260/scan.json.gz) is a trimmed [masscan](https://github.com/robertdavidgraham/masscan) report of the kind referenced in [this tweet](https://twitter.com/acut3hack/status/1648346514467463169). The `jq` example shown there formats it thusly:

```
$ jq --version
jq-1.6

$ jq 'reduce .[] as $e ({}; . + { ($e.ip): (.[$e.ip] + $e.ports) })' scan.json 
{
  "192.168.5.53": [
    {
      "port": 515,
      "proto": "tcp",
      "status": "open",
      "reason": "syn-ack",
      "ttl": 64
    },
    {
      "port": 21,
      "proto": "tcp",
      "status": "open",
      "reason": "syn-ack",
      "ttl": 64
    }
  ],
  "192.168.5.49": [
    {
      "port": 49152,
      "proto": "tcp",
      "status": "open",
      "reason": "syn-ack",
      "ttl": 64
    },
    {
      "port": 3401,
      "proto": "tcp",
      "status": "open",
      "reason": "syn-ack",
      "ttl": 64
    }
  ]
}
```

The cleanest Zed I thought of that outputs something _close_ to that is:

```
$ zq -version
Version: v1.7.0-48-g1874aeb4

$ zq -Z 'over this | ports:=collect(ports[0]) by ip' scan.json 
{
    ip: "192.168.5.53",
    ports: [
        {
            port: 515,
            proto: "tcp",
            status: "open",
            reason: "syn-ack",
            ttl: 64
        },
        {
            port: 21,
            proto: "tcp",
            status: "open",
            reason: "syn-ack",
            ttl: 64
        }
    ]
}
{
    ip: "192.168.5.49",
    ports: [
        {
            port: 49152,
            proto: "tcp",
            status: "open",
            reason: "syn-ack",
            ttl: 64
        },
        {
            port: 3401,
            proto: "tcp",
            status: "open",
            reason: "syn-ack",
            ttl: 64
        }
    ]
}
```

However, notice in this case that we've got named fields `ip` and `ports` whereas in the `jq` example the IP address string was used as an object key with the value being the array of port data. If a user wanted to get the exact same output `jq` was producing, the best approach the team came up with is:

```
$ zq -Z 'over this | ports:=collect(ports[0]) by ip | unflatten([{key:ip,value:ports}])' scan.json 
{
    "192.168.5.53": [
        {
            port: 515,
            proto: "tcp",
            status: "open",
            reason: "syn-ack",
            ttl: 64
        },
        {
            port: 21,
            proto: "tcp",
            status: "open",
            reason: "syn-ack",
            ttl: 64
        }
    ]
}
{
    "192.168.5.49": [
        {
            port: 49152,
            proto: "tcp",
            status: "open",
            reason: "syn-ack",
            ttl: 64
        },
        {
            port: 3401,
            proto: "tcp",
            status: "open",
            reason: "syn-ack",
            ttl: 64
        }
    ]
}
```

(There's actually a subtle difference between this output and what came from `jq`, but that's further explored in #4565.)

This idiom of `collect()` + `unflatten()` has come up a few times before (example: #4332) and I'm starting to see why: `unflatten([{key:...,value:...}])` is the only way I can see to create new field/value pairs in records where the field name is based on a string that's been programmatically generated within the Zed pipeline. It therefore seems like we could benefit from some shorthand that would achieve the same. I'm at a loss for what precisely to propose, but I sense it would be something like the [indexing](https://zed.brimdata.io/docs/language/expressions#indexing) syntax but usable on in "left hand" contexts. For instance, when hacking at this, things I tried in vain hoping they might work included:

```
$ zq -Z 'over this | ports:=collect(ports[0]) by ip | yield {under(ip):ports}' scan.json 
zq: error parsing Zed at column 58:
over this | ports:=collect(ports[0]) by ip | yield {under(ip):ports}
                                                     === ^ ===

$ zq -Z 'over this | ports:=collect(ports[0]) by ip | put this[ip]:=ports' scan.json 
illegal left-hand side of assignment

$ zq -Z 'over this | ports:=collect(ports[0]) by ip | rename this[ip]:=ports' scan.json 
'rename' requires explicit field references
```

mattnibs:
Fixes #4555

### DIFF
--- a/compiler/kernel/expr.go
+++ b/compiler/kernel/expr.go
@@ -263,15 +263,41 @@ func (b *Builder) compileDotExpr(dot *dag.Dot) (expr.Evaluator, error) {
 	return expr.NewDotExpr(b.zctx(), record, dot.RHS), nil
 }
 
-func compileLval(e dag.Expr) (field.Path, error) {
-	if this, ok := e.(*dag.This); ok {
-		return field.Path(this.Path), nil
+func (b *Builder) compileLval(e dag.Expr) (*expr.Lval, error) {
+	switch e := e.(type) {
+	case *dag.BinaryExpr:
+		if e.Op != "[" {
+			return nil, fmt.Errorf("internal error: invalid lval %#v", e)
+		}
+		lhs, err := b.compileLval(e.LHS)
+		if err != nil {
+			return nil, err
+		}
+		rhs, err := b.compileExpr(e.RHS)
+		if err != nil {
+			return nil, err
+		}
+		lhs.Elems = append(lhs.Elems, expr.NewExprLvalElem(b.zctx(), rhs))
+		return lhs, nil
+	case *dag.Dot:
+		lhs, err := b.compileLval(e.LHS)
+		if err != nil {
+			return nil, err
+		}
+		lhs.Elems = append(lhs.Elems, &expr.StaticLvalElem{Name: e.RHS})
+		return lhs, nil
+	case *dag.This:
+		var elems []expr.LvalElem
+		for _, elem := range e.Path {
+			elems = append(elems, &expr.StaticLvalElem{Name: elem})
+		}
+		return expr.NewLval(elems), nil
 	}
-	return nil, errors.New("invalid expression on lhs of assignment")
+	return nil, fmt.Errorf("internal error: invalid lval %#v", e)
 }
 
 func (b *Builder) compileAssignment(node *dag.Assignment) (expr.Assignment, error) {
-	lhs, err := compileLval(node.LHS)
+	lhs, err := b.compileLval(node.LHS)
 	if err != nil {
 		return expr.Assignment{}, err
 	}
--- a/compiler/kernel/groupby.go
+++ b/compiler/kernel/groupby.go
@@ -44,12 +44,12 @@ func (b *Builder) compileAggAssignment(assignment dag.Assignment) (field.Path, *
 	if !ok {
 		return nil, nil, errors.New("aggregator is not an aggregation expression")
 	}
-	lhs, err := compileLval(assignment.LHS)
-	if err != nil {
-		return nil, nil, fmt.Errorf("lhs of aggregation: %w", err)
+	this, ok := assignment.LHS.(*dag.This)
+	if !ok {
+		return nil, nil, fmt.Errorf("internal error: aggregator assignment LHS is not a static path: %#v", assignment.LHS)
 	}
 	m, err := b.compileAgg(aggAST)
-	return lhs, m, err
+	return this.Path, m, err
 }
 
 func (b *Builder) compileAgg(agg *dag.Agg) (*expr.Aggregator, error) {
--- a/compiler/kernel/op.go
+++ b/compiler/kernel/op.go
@@ -109,10 +109,7 @@ func (b *Builder) compileLeaf(o dag.Op, parent zbuf.Puller) (zbuf.Puller, error)
 			return nil, err
 		}
 		lhs, rhs := splitAssignments(assignments)
-		cutter, err := expr.NewCutter(b.octx.Zctx, lhs, rhs)
-		if err != nil {
-			return nil, err
-		}
+		cutter := expr.NewCutter(b.octx.Zctx, lhs, rhs)
 		if v.Quiet {
 			cutter.Quiet()
 		}
@@ -174,36 +171,13 @@ func (b *Builder) compileLeaf(o dag.Op, parent zbuf.Puller) (zbuf.Puller, error)
 		if err != nil {
 			return nil, err
 		}
-		putter, err := expr.NewPutter(b.octx.Zctx, clauses)
-		if err != nil {
-			return nil, err
-		}
+		putter := expr.NewPutter(b.octx.Zctx, clauses)
 		return op.NewApplier(b.octx, parent, putter), nil
 	case *dag.Rename:
 		var srcs, dsts field.List
-		for _, fa := range v.Args {
-			dst, err := compileLval(fa.LHS)
-			if err != nil {
-				return nil, err
-			}
-			// We call CompileLval on the RHS because renames are
-			// restricted to dotted field name expressions.
-			src, err := compileLval(fa.RHS)
-			if err != nil {
-				return nil, err
-			}
-			if len(dst) != len(src) {
-				return nil, fmt.Errorf("cannot rename %s to %s", src, dst)
-			}
-			// Check that the prefixes match and, if not, report first place
-			// that they don't.
-			for i := 0; i <= len(src)-2; i++ {
-				if src[i] != dst[i] {
-					return nil, fmt.Errorf("cannot rename %s to %s (differ in %s vs %s)", src, dst, src[i], dst[i])
-				}
-			}
-			dsts = append(dsts, dst)
-			srcs = append(srcs, src)
+		for _, a := range v.Args {
+			dsts = append(dsts, a.LHS.(*dag.This).Path)
+			srcs = append(srcs, a.RHS.(*dag.This).Path)
 		}
 		renamer := expr.NewRenamer(b.octx.Zctx, srcs, dsts)
 		return op.NewApplier(b.octx, parent, renamer), nil
@@ -388,9 +362,9 @@ func (b *Builder) compileAssignments(assignments []dag.Assignment) ([]expr.Assig
 	return keys, nil
 }
 
-func splitAssignments(assignments []expr.Assignment) (field.List, []expr.Evaluator) {
+func splitAssignments(assignments []expr.Assignment) ([]*expr.Lval, []expr.Evaluator) {
 	n := len(assignments)
-	lhs := make(field.List, 0, n)
+	lhs := make([]*expr.Lval, 0, n)
 	rhs := make([]expr.Evaluator, 0, n)
 	for _, a := range assignments {
 		lhs = append(lhs, a.LHS)
--- a/compiler/semantic/expr.go
+++ b/compiler/semantic/expr.go
@@ -507,10 +507,10 @@ func (a *analyzer) semExprs(in []ast.Expr) ([]dag.Expr, error) {
 	return exprs, nil
 }
 
-func (a *analyzer) semAssignments(assignments []ast.Assignment, summarize bool) ([]dag.Assignment, error) {
+func (a *analyzer) semAssignments(assignments []ast.Assignment) ([]dag.Assignment, error) {
 	out := make([]dag.Assignment, 0, len(assignments))
 	for _, e := range assignments {
-		a, err := a.semAssignment(e, summarize)
+		a, err := a.semAssignment(e)
 		if err != nil {
 			return nil, err
 		}
@@ -519,64 +519,69 @@ func (a *analyzer) semAssignments(assignments []ast.Assignment, summarize bool) 
 	return out, nil
 }
 
-func (a *analyzer) semAssignment(assign ast.Assignment, summarize bool) (dag.Assignment, error) {
+func (a *analyzer) semAssignment(assign ast.Assignment) (dag.Assignment, error) {
 	rhs, err := a.semExpr(assign.RHS)
 	if err != nil {
-		return dag.Assignment{}, fmt.Errorf("rhs of assignment expression: %w", err)
-	}
-	if _, ok := rhs.(*dag.Agg); ok {
-		summarize = true
+		return dag.Assignment{}, fmt.Errorf("right-hand side of assignment: %w", err)
 	}
 	var lhs dag.Expr
-	if assign.LHS != nil {
-		lhs, err = a.semExpr(assign.LHS)
+	if assign.LHS == nil {
+		path, err := deriveLHSPath(rhs)
 		if err != nil {
-			return dag.Assignment{}, fmt.Errorf("lhs of assigment expression: %w", err)
+			return dag.Assignment{}, err
 		}
-	} else if call, ok := assign.RHS.(*ast.Call); ok {
-		path := []string{call.Name}
-		switch call.Name {
+		lhs = &dag.This{Kind: "This", Path: path}
+	} else if lhs, err = a.semExpr(assign.LHS); err != nil {
+		return dag.Assignment{}, fmt.Errorf("left-hand side of assignment: %w", err)
+	}
+	if !isLval(lhs) {
+		return dag.Assignment{}, errors.New("illegal left-hand side of assignment")
+	}
+	if this, ok := lhs.(*dag.This); ok && len(this.Path) == 0 {
+		return dag.Assignment{}, errors.New("cannot assign to 'this'")
+	}
+	return dag.Assignment{Kind: "Assignment", LHS: lhs, RHS: rhs}, nil
+}
+
+func isLval(e dag.Expr) bool {
+	switch e := e.(type) {
+	case *dag.BinaryExpr:
+		return e.Op == "[" && isLval(e.LHS)
+	case *dag.Dot:
+		return isLval(e.LHS)
+	case *dag.This:
+		return true
+	}
+	return false
+}
+
+func deriveLHSPath(rhs dag.Expr) ([]string, error) {
+	var path []string
+	switch rhs := rhs.(type) {
+	case *dag.Call:
+		path = []string{rhs.Name}
+		switch rhs.Name {
 		case "every":
 			// If LHS is nil and the call is every() make the LHS field ts since
 			// field ts assumed with every.
 			path = []string{"ts"}
 		case "quiet":
-			if len(call.Args) > 0 {
-				if p, ok := rhs.(*dag.Call).Args[0].(*dag.This); ok {
-					path = p.Path
+			if len(rhs.Args) > 0 {
+				if this, ok := rhs.Args[0].(*dag.This); ok {
+					path = this.Path
 				}
 			}
 		}
-		lhs = &dag.This{Kind: "This", Path: path}
-	} else if agg, ok := assign.RHS.(*ast.Agg); ok {
-		lhs = &dag.This{Kind: "This", Path: []string{agg.Name}}
-	} else if v, ok := rhs.(*dag.Var); ok {
-		lhs = &dag.This{Kind: "This", Path: []string{v.Name}}
-	} else {
-		lhs, err = a.semExpr(assign.RHS)
-		if err != nil {
-			return dag.Assignment{}, errors.New("assignment name could not be inferred from rhs expression")
-		}
+	case *dag.Agg:
+		path = []string{rhs.Name}
+	case *dag.Var:
+		path = []string{rhs.Name}
+	case *dag.This:
+		path = rhs.Path
+	default:
+		return nil, errors.New("cannot infer field from expression")
 	}
-	if summarize {
-		// Summarize always outputs its results as new records of "this"
-		// so if we have an "as" that overrides "this", we just
-		// convert it back to a local this.
-		if dot, ok := lhs.(*dag.Dot); ok {
-			if v, ok := dot.LHS.(*dag.Var); ok && v.Name == "this" {
-				lhs = &dag.This{Kind: "This", Path: []string{dot.RHS}}
-			}
-		}
-	}
-	// Make sure we have a valid lval for lhs.
-	this, ok := lhs.(*dag.This)
-	if !ok {
-		return dag.Assignment{}, errors.New("illegal left-hand side of assignment")
-	}
-	if len(this.Path) == 0 {
-		return dag.Assignment{}, errors.New("cannot assign to 'this'")
-	}
-	return dag.Assignment{Kind: "Assignment", LHS: lhs, RHS: rhs}, nil
+	return path, nil
 }
 
 func (a *analyzer) semFields(exprs []ast.Expr) ([]dag.Expr, error) {
--- a/compiler/semantic/op.go
+++ b/compiler/semantic/op.go
@@ -15,6 +15,7 @@ import (
 	"github.com/brimdata/zed/pkg/field"
 	"github.com/brimdata/zed/pkg/plural"
 	"github.com/brimdata/zed/pkg/reglob"
+	"github.com/brimdata/zed/runtime/expr"
 	"github.com/brimdata/zed/runtime/expr/function"
 	"github.com/brimdata/zed/zson"
 	"github.com/segmentio/ksuid"
@@ -410,18 +411,24 @@ func (a *analyzer) semOp(o ast.Op, seq dag.Seq) (dag.Seq, error) {
 	case *ast.From:
 		return a.semFrom(o, seq)
 	case *ast.Summarize:
-		keys, err := a.semAssignments(o.Keys, true)
+		keys, err := a.semAssignments(o.Keys)
 		if err != nil {
 			return nil, err
+		}
+		if assignmentHasDynamicLHS(keys) {
+			return nil, errors.New("summarize: key output field must be static")
 		}
 		if len(keys) == 0 && len(o.Aggs) == 1 {
 			if seq := a.singletonAgg(o.Aggs[0], seq); seq != nil {
 				return seq, nil
 			}
 		}
-		aggs, err := a.semAssignments(o.Aggs, true)
+		aggs, err := a.semAssignments(o.Aggs)
 		if err != nil {
 			return nil, err
+		}
+		if assignmentHasDynamicLHS(aggs) {
+			return nil, errors.New("summarize: aggregate output field must be static")
 		}
 		// Note: InputSortDir is copied in here but it's not meaningful
 		// coming from a parser AST, only from a worker using the kernel DSL,
@@ -497,9 +504,19 @@ func (a *analyzer) semOp(o ast.Op, seq dag.Seq) (dag.Seq, error) {
 	case *ast.Shape:
 		return append(seq, &dag.Shape{Kind: "Shape"}), nil
 	case *ast.Cut:
-		assignments, err := a.semAssignments(o.Args, false)
+		assignments, err := a.semAssignments(o.Args)
 		if err != nil {
 			return nil, err
+		}
+		// Collect static paths so we can check on what is available.
+		var fields field.List
+		for _, a := range assignments {
+			if this, ok := a.LHS.(*dag.This); ok {
+				fields = append(fields, this.Path)
+			}
+		}
+		if _, err = zed.NewRecordBuilder(a.zctx, fields); err != nil {
+			return nil, fmt.Errorf("cut: %w", err)
 		}
 		return append(seq, &dag.Cut{
 			Kind: "Cut",
@@ -596,9 +613,19 @@ func (a *analyzer) semOp(o ast.Op, seq dag.Seq) (dag.Seq, error) {
 			Limit: o.Limit,
 		}), nil
 	case *ast.Put:
-		assignments, err := a.semAssignments(o.Args, false)
+		assignments, err := a.semAssignments(o.Args)
 		if err != nil {
 			return nil, err
+		}
+		// We can do collision checking on static paths, so check what we can.
+		var fields field.List
+		for _, a := range assignments {
+			if this, ok := a.LHS.(*dag.This); ok {
+				fields = append(fields, this.Path)
+			}
+		}
+		if err := expr.CheckPutFields(fields); err != nil {
+			return nil, fmt.Errorf("put: %w", err)
 		}
 		return append(seq, &dag.Put{
 			Kind: "Put",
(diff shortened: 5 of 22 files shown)


philrz:
As a sign that we're on the right track, here's verification that this branch at commit 271bb18 seems to be addressing the use case laid out in #4555. As shown there, creating a new record field with a name derived from a string value required me to use a record literal + `unflatten()` idiom, which I've never found super intuitive.

```
$ zq -Z 'over this | ports:=collect(ports[0]) by ip | unflatten([{key:ip,value:ports}])' scan.json | tee 1
{
    "192.168.5.53": [
        {
            port: 515,
            proto: "tcp",
            status: "open",
            reason: "syn-ack",
            ttl: 64
        },
        {
            port: 21,
            proto: "tcp",
            status: "open",
            reason: "syn-ack",
            ttl: 64
        }
    ]
}
{
    "192.168.5.49": [
        {
            port: 49152,
            proto: "tcp",
            status: "open",
            reason: "syn-ack",
            ttl: 64
        },
        {
            port: 3401,
            proto: "tcp",
            status: "open",
            reason: "syn-ack",
            ttl: 64
        }
    ]
}
```

Running this branch at commit 271bb18, I can now obtain the same output using the syntax below, which previously triggered the `illegal left-hand side of assignment` error.

```
$ zq -version
Version: v1.10.0-5-g271bb189

$ zq -Z 'over this | ports:=collect(ports[0]) by ip | put this[ip]:=ports | drop ip,ports' scan.json | tee 2
{
    "192.168.5.53": [
        {
            port: 515,
            proto: "tcp",
            status: "open",
            reason: "syn-ack",
            ttl: 64
        },
        {
            port: 21,
            proto: "tcp",
            status: "open",
            reason: "syn-ack",
            ttl: 64
        }
    ]
}
{
    "192.168.5.49": [
        {
            port: 49152,
            proto: "tcp",
            status: "open",
            reason: "syn-ack",
            ttl: 64
        },
        {
            port: 3401,
            proto: "tcp",
            status: "open",
            reason: "syn-ack",
            ttl: 64
        }
    ]
}

$ diff 1 2
[no output]
```

mattnibs:
Thanks @philrz. 

Btw, I didn't realize this until working on the pr but

```
over this | ports:=collect(ports[0]) by ip | put this[ip]:=ports | drop ip,ports
```

can be shortened to
```
over this | ports:=collect(ports[0]) by ip | cut this[ip]:=ports
```

compiler/ast/dag/unpack.go
```diff
@@ -36,6 +36,8 @@ var unpacker = unpack.New(
 	Over{},
 	OverExpr{},
 	Pass{},
+	Path{},
+	StaticPathElem{},
```
nwt, line 40: Nit: Move this down so these stay alphabetized.

compiler/ast/dag/op.go
```diff
@@ -95,8 +95,9 @@ type (
 		Args []Assignment `json:"args"`
 	}
 	Rename struct {
-		Kind string       `json:"kind" unpack:""`
-		Args []Assignment `json:"args"`
+		Kind string  `json:"kind" unpack:""`
+		Dsts []*This `json:"dsts"`
+		Srcs []*This `json:"srcs"`
```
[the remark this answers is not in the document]
mattnibs: Yeah I'll rework this.

compiler/kernel/expr.go
```diff
@@ -263,15 +263,42 @@ func (b *Builder) compileDotExpr(dot *dag.Dot) (expr.Evaluator, error) {
...
+	switch e := e.(type) {
+	case *dag.This:
+		var elems []expr.LvalElem
+		for _, elem := range e.Path {
+			elems = append(elems, &expr.StaticLvalElem{Name: elem})
+		}
+		return expr.NewLval(elems), nil
+	case *dag.BinaryExpr:
+		if e.Op != "[" {
+			return nil, fmt.Errorf("internal error: invalid path expression %T", e)
```
nwt, line 276:
```suggestion
			return nil, fmt.Errorf("internal error: invalid lval %#v", e)
```

compiler/kernel/expr.go
```diff
@@ -263,15 +263,42 @@ func (b *Builder) compileDotExpr(dot *dag.Dot) (expr.Evaluator, error) {
...
+		return lhs, nil
+	case *dag.Dot:
+		lhs, err := b.compileLval(e.LHS)
+		if err != nil {
+			return nil, err
+		}
+		lhs.Elems = append(lhs.Elems, &expr.StaticLvalElem{Name: e.RHS})
+		return lhs, nil
+	default:
+		return nil, fmt.Errorf("internal error: invalid path expression %T", e)
```
nwt, line 296:
```suggestion
		return nil, fmt.Errorf("internal error: invalid lval %#v", e)
```

compiler/semantic/expr.go
```diff
@@ -519,64 +519,82 @@ func (a *analyzer) semAssignments(assignments []ast.Assignment, summarize bool)
 	return out, nil
 }
 
-func (a *analyzer) semAssignment(assign ast.Assignment, summarize bool) (dag.Assignment, error) {
-	rhs, err := a.semExpr(assign.RHS)
+func (a *analyzer) semAssignment(e ast.Assignment) (dag.Assignment, error) {
```
nwt, line 522: Nit: `e` suggests an expression so stick with `assign` here.
```suggestion
func (a *analyzer) semAssignment(assign ast.Assignment) (dag.Assignment, error) {
```

compiler/semantic/expr.go
```diff
@@ -519,64 +519,82 @@ func (a *analyzer) semAssignments(assignments []ast.Assignment, summarize bool)
...
+	rhs, err := a.semExpr(e.RHS)
 	if err != nil {
-		return dag.Assignment{}, fmt.Errorf("rhs of assignment expression: %w", err)
-	}
-	if _, ok := rhs.(*dag.Agg); ok {
-		summarize = true
+		return dag.Assignment{}, err
```
nwt, line 525: Maybe stick with
```go
 		return dag.Assignment{}, fmt.Errorf("rhs of assignment expression: %w", err)
```
(and likewise for an LHS error) since this seems like it provides useful context?

compiler/kernel/expr.go
```diff
@@ -263,15 +263,42 @@ func (b *Builder) compileDotExpr(dot *dag.Dot) (expr.Evaluator, error) {
 	return expr.NewDotExpr(b.zctx(), record, dot.RHS), nil
 }
 
-func compileLval(e dag.Expr) (field.Path, error) {
-	if this, ok := e.(*dag.This); ok {
-		return field.Path(this.Path), nil
+func (b *Builder) compileLval(e dag.Expr) (*expr.Lval, error) {
+	switch e := e.(type) {
+	case *dag.This:
```
nwt, line 268: Maybe alphabetize these cases?

compiler/semantic/expr.go
```diff
@@ -519,64 +519,82 @@ func (a *analyzer) semAssignments(assignments []ast.Assignment, summarize bool)
...
+		if len(this.Path) == 0 {
+			return dag.Assignment{}, errors.New("cannot assign to 'this'")
+		}
+	}
+	return dag.Assignment{Kind: "Assignment", LHS: lhs, RHS: rhs}, nil
+}
+
+func (a *analyzer) isLval(e dag.Expr) bool {
+	switch e := e.(type) {
+	case *dag.This:
```
nwt, line 552: Maybe alphabetize these cases?

compiler/semantic/expr.go
mattnibs: Is lhs/rhs a user friendly abbreviation or should it be left-hand side / right-hand side?

philrz: I'd advocate for the more verbose "left-hand side / right-hand side".

compiler/kernel/op.go
```diff
@@ -174,36 +174,13 @@ func (b *Builder) compileLeaf(o dag.Op, parent zbuf.Puller) (zbuf.Puller, error)
...
-			for i := 0; i <= len(src)-2; i++ {
-				if src[i] != dst[i] {
-					return nil, fmt.Errorf("cannot rename %s to %s (differ in %s vs %s)", src, dst, src[i], dst[i])
-				}
-			}
-			dsts = append(dsts, dst)
-			srcs = append(srcs, src)
+		for k := range v.Args {
+			dsts = append(dsts, v.Args[k].LHS.(*dag.This).Path)
+			srcs = append(srcs, v.Args[k].RHS.(*dag.This).Path)
```
nwt, line 183:
```suggestion
		for _, a := range v.Args {
			dsts = append(dsts, a.LHS.(*dag.This).Path)
			srcs = append(srcs, z.RHS.(*dag.This).Path)
```

compiler/semantic/expr.go
```diff
@@ -519,64 +519,82 @@ func (a *analyzer) semAssignments(assignments []ast.Assignment, summarize bool)
...
-		lhs, err = a.semExpr(assign.LHS)
+	if assign.LHS == nil {
+		path, err := derriveLHSPath(rhs)
 		if err != nil {
-			return dag.Assignment{}, fmt.Errorf("lhs of assigment expression: %w", err)
+			return dag.Assignment{}, err
+		}
+		lhs = &dag.This{Kind: "This", Path: path}
+	} else {
+		if lhs, err = a.semExpr(assign.LHS); err != nil {
```
nwt, line 535:
```suggestion
	} else if lhs, err = a.semExpr(assign.LHS); err != nil {
```

compiler/semantic/expr.go
```diff
@@ -519,64 +519,82 @@ func (a *analyzer) semAssignments(assignments []ast.Assignment, summarize bool)
...
+			return dag.Assignment{}, fmt.Errorf("left-hand side of assignment expression: %w", err)
+		}
+	}
+	if !a.isLval(lhs) {
+		return dag.Assignment{}, errors.New("illegal left-hand side of assignment expression")
+	}
+	if this, ok := lhs.(*dag.This); ok {
+		if len(this.Path) == 0 {
+			return dag.Assignment{}, errors.New("cannot assign to 'this'")
+		}
```
nwt, line 545:
```suggestion
	if this, ok := lhs.(*dag.This); ok && len(this.Path) == 0 {
		return dag.Assignment{}, errors.New("cannot assign to 'this'")
```

compiler/semantic/expr.go
```diff
@@ -519,64 +519,82 @@ func (a *analyzer) semAssignments(assignments []ast.Assignment, summarize bool)
...
+	}
+	return dag.Assignment{Kind: "Assignment", LHS: lhs, RHS: rhs}, nil
+}
+
+func (a *analyzer) isLval(e dag.Expr) bool {
```
nwt, line 550: Doesn't look like this needs to be a method.

compiler/semantic/expr.go
```diff
@@ -519,64 +519,82 @@ func (a *analyzer) semAssignments(assignments []ast.Assignment, summarize bool)
...
+		return e.Op == "[" && a.isLval(e.LHS)
+	case *dag.Dot:
+		return a.isLval(e.LHS)
+	case *dag.This:
+		return true
+	}
+	return false
+}
+
+func assignHasDynamicPath(assignments []dag.Assignment) bool {
```
nwt, line 562: Nits: I think this would be a clearer name. And maybe move to op.go since that's the only file in which it's used?
```suggestion
func assignmentHasDynamicLHS(assignments []dag.Assignment) bool {
```

compiler/kernel/groupby.go
```diff
@@ -44,12 +43,12 @@ func (b *Builder) compileAggAssignment(assignment dag.Assignment) (field.Path, *
 	if !ok {
 		return nil, nil, errors.New("aggregator is not an aggregation expression")
 	}
-	lhs, err := compileLval(assignment.LHS)
-	if err != nil {
-		return nil, nil, fmt.Errorf("lhs of aggregation: %w", err)
+	this, ok := assignment.LHS.(*dag.This)
+	if !ok {
+		return nil, nil, errors.New("internal error: aggregator assignment must be a static path")
```
nwt, line 48:
```suggestion
		return nil, nil, errors.New("internal error: aggregator assignment LHS must be a static path")
```

compiler/semantic/expr.go
```diff
@@ -519,64 +519,69 @@ func (a *analyzer) semAssignments(assignments []ast.Assignment, summarize bool)
...
-	if len(this.Path) == 0 {
-		return dag.Assignment{}, errors.New("cannot assign to 'this'")
+	case *dag.Agg:
+		path = []string{rhs.Name}
+	case *dag.Var:
+		path = []string{rhs.Name}
+	case *dag.This:
+		path = rhs.Path
+	default:
+		return nil, errors.New("assignment name could not be inferred from rhs expression")
```
nwt, line 582: Given the context in which this error occurs (i.e., there's no `:=` to provide an RHS), I don't think it makes sense for the error string to use "rhs".
```suggestion
		return nil, errors.New("cannot infer field name from expression")
```

compiler/semantic/op.go
```diff
@@ -410,19 +410,25 @@ func (a *analyzer) semOp(o ast.Op, seq dag.Seq) (dag.Seq, error) {
...
 			if seq := a.singletonAgg(o.Aggs[0], seq); seq != nil {
 				return seq, nil
 			}
 		}
-		aggs, err := a.semAssignments(o.Aggs, true)
+		aggs, err := a.semAssignments(o.Aggs)
 		if err != nil {
 			return nil, err
 		}
+		if assignmentHasDynamicLHS(keys) {
```
nwt, line 429: These error cases should have test coverage.
```suggestion
		if assignmentHasDynamicLHS(aggs) {
```

compiler/semantic/op.go
```diff
@@ -596,10 +612,33 @@ func (a *analyzer) semOp(o ast.Op, seq dag.Seq) (dag.Seq, error) {
...
+			}
+		}
+		for i, f := range fields {
+			if f.IsEmpty() {
+				return nil, fmt.Errorf("put: LHS cannot be 'this' (use 'yield' operator)")
+			}
+			for j, c := range fields {
+				if i == j {
+					continue
+				}
```
nwt, line 633:
```suggestion
			for _, c := range fields[i+1:] {
```

runtime/expr/ztests/cut-dup-fields.yaml
```diff
@@ -3,6 +3,7 @@ script: |
   ! zq -z "cut rec.sub1,rec.sub1" in.zson
   ! zq -z "cut rec.sub,rec.sub.sub" in.zson
   ! zq -z "cut rec.sub.sub,rec.sub" in.zson
+  # XXX Add runtime evaluatable cut (i.e., dynamic path).
```
nwt, line 6: Do this now?

runtime/expr/eval.go
```diff
@@ -807,19 +807,23 @@ func (c *Call) Eval(ectx Context, this *zed.Value) *zed.Value {
...
 }
 
-func NewAssignments(zctx *zed.Context, dsts field.List, srcs field.List) (field.List, []Evaluator) {
+func NewAssignments(zctx *zed.Context, dsts field.List, srcs field.List) ([]*Lval, []Evaluator) {
 	if len(srcs) != len(dsts) {
 		panic("NewAssignments: argument mismatch")
 	}
 	var resolvers []Evaluator
-	var fields field.List
+	var paths []*Lval
```
nwt, line 819: Nit: `lvals` seems like a better name for this.

compiler/semantic/op.go
```diff
@@ -596,10 +612,33 @@ func (a *analyzer) semOp(o ast.Op, seq dag.Seq) (dag.Seq, error) {
...
+				if f.Equal(c) {
+					return nil, fmt.Errorf("put: multiple assignments to %s", f)
+				}
+				if c.HasStrictPrefix(f) {
+					return nil, fmt.Errorf("put: conflicting nested assignments to %s and %s", f, c)
+				}
+			}
+		}
```
nwt, line 641: This is a nearly verbatim copy of runtime/expr/checkFields. Export that as CheckPutFields or similar and call it here.

runtime/expr/putter.go
```diff
@@ -273,19 +262,48 @@ func hasField(name string, fields []zed.Field) bool {
...
+
+func checkFields(fields field.List) error {
+	for i, f := range fields {
+		if f.IsEmpty() {
+			return fmt.Errorf("put: LHS cannot be 'this' (use 'yield' operator)")
+		}
+		for j, c := range fields {
+			if i == j {
+				continue
+			}
```
nwt, line 297:
```suggestion
		for _, c := range fields[i+1:] {
```

compiler/semantic/op.go
mattnibs: Good idea

compiler/semantic/expr.go
mattnibs: Cool

compiler/semantic/op.go
mattnibs: Indeed they should not in the pr.

runtime/expr/ztests/cut-dup-fields.yaml
mattnibs: We already have a test for this compiler/ztests/dynamic-field-cut

compiler/semantic/op.go
mattnibs: Hmm I guess I'll add them here since this wasn't actually possible before this pr

compiler/semantic/op.go
```diff
@@ -410,19 +411,25 @@ func (a *analyzer) semOp(o ast.Op, seq dag.Seq) (dag.Seq, error) {
 	case *ast.From:
 		return a.semFrom(o, seq)
 	case *ast.Summarize:
-		keys, err := a.semAssignments(o.Keys, true)
+		keys, err := a.semAssignments(o.Keys)
 		if err != nil {
 			return nil, err
 		}
+		if assignmentHasDynamicLHS(keys) {
+			return nil, errors.New("summarize: key output fields must be static")
```
nwt, line 419: Make this consistent with the other assignmentHasDynamicLHS error string below.
```suggestion
			return nil, errors.New("summarize: key output field must be static")
```

compiler/ztests/dynamic-field-cut.yaml
```diff
@@ -0,0 +1,41 @@
+script: |
```
nwt, line 1: Nit: There are some cut and put operator tests in compiler/op/ztests so move this and dynamic-field-put.yaml there and name them {cut,put}-dynamic-field.yaml in keeping with the existing tests.

compiler/ztests/summarize-lhs-error.yaml
```diff
@@ -0,0 +1,9 @@
+script: |
+  ! zc -s 'count() by this[a] := key'
+  ! zc -s 'this[a] := count() by key'
+
```
nwt, line 4: Add coverage for the OpAssignment case.
```suggestion
  ! zc -s 'this[a] := count()'

```

runtime/expr/lval.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	"github.com/brimdata/zed"
+	"github.com/brimdata/zed/pkg/field"
+)
+
+type LvalElem interface {
+	Eval(ectx Context, this *zed.Value) (string, *zed.Value)
+}
+
+type Lval struct {
```
nwt, line 12: Nit: Lval is more interesting so put it first.

compiler/ztests/dynamic-field-cut.yaml
mattnibs: Fair

nwt: Indeed, I do.

runtime/expr/cutter.go
```diff
@@ -67,30 +55,42 @@ func (c *Cutter) FoundCut() bool {
 // receiver's configuration.  If the resulting record would be empty, Apply
 // returns zed.Missing.
 func (c *Cutter) Eval(ectx Context, in *zed.Value) *zed.Value {
+	rb, paths, verr := c.lookupBuilder(ectx, in)
+	if verr != nil {
+		if !verr.IsMissing() && !verr.IsQuiet() {
+			verr = ectx.CopyValue(*c.zctx.WrapError(fmt.Sprintf("cut: %s", zed.DecodeError(verr.Bytes())), in))
```
nwt, line 61: `zed.DecodeError` is broken. See #4810.

mattnibs: What do you propose to do instead here?

mattnibs: Just decode it as a string?

mattnibs: We're just dropping the wrap for now and will revisit this...

runtime/expr/lval.go
```diff
@@ -0,0 +1,80 @@
...
+
+func (l *ExprLvalElem) Eval(ectx Context, this *zed.Value) (string, *zed.Value) {
+	val := l.eval.Eval(ectx, this)
+	if val.IsError() {
+		return "", val
+	}
+	if !val.IsString() {
+		if val = l.caster.Eval(ectx, val); val.IsError() {
+			return "", val
+		}
```
nwt, line 77: This doesn't feel quite right. How about returning an error, like this:
```suggestion
		return ectx.CopyValue(*l.zctx.WrapError("field reference is not a string", val))
```

runtime/expr/putter.go
```diff
@@ -39,43 +39,32 @@ type putRule struct {
...
-		if !val.IsQuiet() {
-			p.vals = append(p.vals, val)
-			p.valClauses = append(p.valClauses, cl)
+		if val.IsQuiet() {
+			continue
+		}
+		p.vals = append(p.vals, val)
+		path, verr := cl.LHS.Eval(ectx, this)
+		if verr != nil {
+			return nil, nil, verr
```
nwt, line 63:
```suggestion
		path, errVal := cl.LHS.Eval(ectx, this)
		if errVal != nil {
			return nil, nil, errVal
```

runtime/expr/putter.go
```diff
@@ -19,12 +19,12 @@ type Putter struct {
...
-	valClauses []Assignment
+	rules   map[int]map[string]putRule
+	warned  map[string]struct{}
 	// vals is a slice to avoid re-allocating for every value
-	vals   []zed.Value
-	rules  map[int]putRule
-	warned map[string]struct{}
+	vals []zed.Value
+	// paths is a slice to avoid re-allocating for every path
+	paths []field.Path
```
nwt, line 27:
```suggestion
	paths field.List
```

runtime/expr/putter.go
```diff
@@ -39,43 +39,32 @@ type putRule struct {
...
-		rules:   make(map[int]putRule),
+		rules:   make(map[int]map[string]putRule),
 		warned:  make(map[string]struct{}),
-	}, nil
+	}
 }
 
-func (p *Putter) eval(ectx Context, this *zed.Value) ([]zed.Value, []Assignment) {
-	p.valClauses = p.valClauses[:0]
+func (p *Putter) eval(ectx Context, this *zed.Value) ([]zed.Value, []field.Path, *zed.Value) {
```
nwt, line 52:
```suggestion
func (p *Putter) eval(ectx Context, this *zed.Value) ([]zed.Value, field.List, *zed.Value) {
```

runtime/expr/putter.go
```diff
@@ -175,20 +164,20 @@ func (ig *getter) nth(n int) (zcode.Bytes, error) {
 	return nil, fmt.Errorf("getter.nth: array index %d out of bounds", n)
 }
 
-func findOverwriteClause(path field.Path, clauses []Assignment) (int, field.Path, bool) {
-	for i, cand := range clauses {
-		if path.Equal(cand.LHS) || cand.LHS.HasStrictPrefix(path) {
-			return i, cand.LHS, true
+func findOverwriteClause(path field.Path, paths []field.Path) (int, field.Path, bool) {
```
nwt, line 167:
```suggestion
func findOverwriteClause(path field.Path, paths field.List) (int, field.Path, bool) {
```

runtime/expr/putter.go
```diff
@@ -175,20 +164,20 @@ func (ig *getter) nth(n int) (zcode.Bytes, error) {
...
+		if path.Equal(lpath) || lpath.HasStrictPrefix(path) {
+			return i, lpath, true
 		}
 	}
 	return -1, nil, false
 }
 
-func (p *Putter) deriveSteps(inType *zed.TypeRecord, vals []zed.Value, clauses []Assignment) (putStep, zed.Type) {
-	return p.deriveRecordSteps(field.Path{}, inType.Fields, vals, clauses)
+func (p *Putter) deriveSteps(inType *zed.TypeRecord, vals []zed.Value, paths []field.Path) (putStep, zed.Type) {
```
nwt, line 176:
```suggestion
func (p *Putter) deriveSteps(inType *zed.TypeRecord, vals []zed.Value, paths field.List) (putStep, zed.Type) {
```

runtime/expr/putter.go
```diff
@@ -175,20 +164,20 @@ func (ig *getter) nth(n int) (zcode.Bytes, error) {
...
+	return p.deriveRecordSteps(field.Path{}, inType.Fields, vals, paths)
 }
 
-func (p *Putter) deriveRecordSteps(parentPath field.Path, inFields []zed.Field, vals []zed.Value, clauses []Assignment) (putStep, *zed.TypeRecord) {
+func (p *Putter) deriveRecordSteps(parentPath field.Path, inFields []zed.Field, vals []zed.Value, paths []field.Path) (putStep, *zed.TypeRecord) {
```
nwt, line 180:
```suggestion
func (p *Putter) deriveRecordSteps(parentPath field.Path, inFields []zed.Field, vals []zed.Value, paths field.List) (putStep, *zed.TypeRecord) {
```

runtime/expr/cutter.go
nwt: OK, here's my proposal.

1. lookupBuilder should return a Go error instead of a zed.Value.
2. lookupBuilder should convert a Zed error to a Go error like this.
   ```go
   	for _, lval := range c.lvals {
   		path, errVal := lval.Eval(ectx, in)
   		if errVal != nil {
   			val := zed.NewValue(errVal.Type.(*zed.TypeError).Type, errVal.Bytes())
   			if val.IsString() {
   				return errors.New(val.AsString())
   			}
   			return errors.New(zson.FormatValue(val))
   		}
   		...
   	}
   ```
3. Cutter.Eval should handle an error from lookupBuilder like this.
   ```go
   	b, paths, err := c.lookupBuilder(ectx, in)
   	if err != nil {
   		return ectx.CopyValue(*c.zctx.WrapError(fmt.Sprintf("cut: %s", err, in))
   	}
   ```
4. We should create an issue to revisit this.

mattnibs: Not great but an improvement to what I have.

runtime/expr/cutter.go
```diff
@@ -67,30 +55,42 @@ func (c *Cutter) FoundCut() bool {
...
-	b := c.builder
-	b.Reset()
 	droppers := c.dropperCache[:0]
 	for k, e := range c.fieldExprs {
 		val := e.Eval(ectx, in)
 		if val.IsQuiet() {
 			// ignore this field
-			if c.droppers[k] == nil {
-				c.droppers[k] = NewDropper(c.zctx, c.fieldRefs[k:k+1])
+			// This no worky because the path may be dynamic.
```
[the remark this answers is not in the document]
mattnibs: I don't know I put this here I think this can go.

index/keys.go
```diff
@@ -13,10 +13,7 @@ type Keyer struct {
...
 func NewKeyer(zctx *zed.Context, keys []field.Path) (*Keyer, error) {
 	fields, resolvers := expr.NewAssignments(zctx, keys, keys)
-	cutter, err := expr.NewCutter(zctx, fields, resolvers)
-	if err != nil {
-		return nil, err
-	}
+	cutter := expr.NewCutter(zctx, fields, resolvers)
 	return &Keyer{
 		keys:   keys,
 		cutter: cutter,
```
nwt, line 19:
```suggestion
	return &Keyer{
		keys:   keys,
		cutter: expr.NewCutter(zctx, fields, resolvers),
```

runtime/op/join/join.go
```diff
@@ -35,19 +34,17 @@ type Op struct {
...
 
 func New(octx *op.Context, anti, inner bool, left, right zbuf.Puller, leftKey, rightKey expr.Evaluator,
-	leftDir, rightDir order.Direction, lhs field.List,
+	leftDir, rightDir order.Direction, lhs []*expr.Lval,
 	rhs []expr.Evaluator) (*Op, error) {
-	cutter, err := expr.NewCutter(octx.Zctx, lhs, rhs)
-	if err != nil {
-		return nil, err
-	}
+	cutter := expr.NewCutter(octx.Zctx, lhs, rhs)
```
nwt, line 39: Nit: Drop this and replace
```go
		cutter:      cutter,
```
with
```go
		cutter:      expr.NewCutter(octx.Zctx, lhs, rhs)
```
in the composite literal down below.

nwt (approved):
:shipit:!!!

PR #4802, opened 2023-10-11
Dynamic fields in rename

mattnibs:
Add the ability to rename fields dynamic for the rename op.

### DIFF
--- a/compiler/kernel/op.go
+++ b/compiler/kernel/op.go
@@ -174,10 +174,18 @@ func (b *Builder) compileLeaf(o dag.Op, parent zbuf.Puller) (zbuf.Puller, error)
 		putter := expr.NewPutter(b.octx.Zctx, clauses)
 		return op.NewApplier(b.octx, parent, putter), nil
 	case *dag.Rename:
-		var srcs, dsts field.List
+		var srcs, dsts []*expr.Lval
 		for _, a := range v.Args {
-			dsts = append(dsts, a.LHS.(*dag.This).Path)
-			srcs = append(srcs, a.RHS.(*dag.This).Path)
+			src, err := b.compileLval(a.RHS)
+			if err != nil {
+				return nil, err
+			}
+			dst, err := b.compileLval(a.LHS)
+			if err != nil {
+				return nil, err
+			}
+			srcs = append(srcs, src)
+			dsts = append(dsts, dst)
 		}
 		renamer := expr.NewRenamer(b.octx.Zctx, srcs, dsts)
 		return op.NewApplier(b.octx, parent, renamer), nil
--- a/compiler/semantic/op.go
+++ b/compiler/semantic/op.go
@@ -640,25 +640,23 @@ func (a *analyzer) semOp(o ast.Op, seq dag.Seq) (dag.Seq, error) {
 	case *ast.Rename:
 		var assignments []dag.Assignment
 		for _, fa := range o.Args {
-			dst, err := a.semField(fa.LHS)
+			assign, err := a.semAssignment(fa)
 			if err != nil {
-				return nil, errors.New("rename: requires explicit field references")
+				return nil, fmt.Errorf("rename: %w", err)
 			}
-			src, err := a.semField(fa.RHS)
-			if err != nil {
-				return nil, errors.New("rename: requires explicit field references")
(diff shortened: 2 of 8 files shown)


compiler/kernel/op.go
```diff
@@ -174,10 +174,18 @@ func (b *Builder) compileLeaf(o dag.Op, parent zbuf.Puller) (zbuf.Puller, error)
...
-		for _, a := range v.Args {
-			dsts = append(dsts, a.LHS.(*dag.This).Path)
-			srcs = append(srcs, a.RHS.(*dag.This).Path)
+		var srcs, dsts []*expr.Lval
+		for k := range v.Args {
+			src, err := b.compileLval(v.Args[k].RHS)
+			if err != nil {
+				return nil, err
+			}
+			dst, err := b.compileLval(v.Args[k].LHS)
```
nwt, line 183:
```suggestion
		for _, a := range v.Args {
			src, err := b.compileLval(a.RHS)
			if err != nil {
				return nil, err
			}
			dst, err := b.compileLval(a.LHS)
```

runtime/expr/renamer.go
```diff
@@ -17,13 +17,87 @@ type Renamer struct {
...
+	if !ok {
+		var err error
+		typ, err = r.computeType(zed.TypeRecordOf(this.Type), srcs, dsts)
+		if err != nil {
+			return ectx.CopyValue(*r.zctx.WrapError(fmt.Sprintf("rename: %s", err), this))
+		}
+		m[string(r.fieldsStr)] = typ
+	}
+	out := this.Copy()
+	return ectx.NewValue(typ, out.Bytes())
```
nwt, line 57: This copy isn't necessary.
```suggestion
	return ectx.NewValue(typ, this.Bytes())
```

pkg/field/field.go
```diff
@@ -15,6 +15,20 @@ func (p Path) String() string {
 	return strings.Join(p, ".")
 }
 
+// AppendTo appends the string representation of the path to byte slice b.
+func (p Path) AppendTo(b []byte) []byte {
```
nwt, line 19: Simplify Path.String.
```go
func (p Path) String() string { return string(p.AppendTo(nil)) }
```

pkg/field/field.go
```diff
@@ -65,6 +79,17 @@ func (l List) String() string {
 	return strings.Join(paths, ",")
 }
 
+// AppendTo appends the string representation of the list to byte slice b.
+func (l List) AppendTo(b []byte) []byte {
```
nwt, line 83: Simplify List.String.
```go
func (l List) String() string { return string(l.AppendTo(nil)) }

File: compiler/semantic/op.go [hunk 5]
```diff
@@ -640,22 +640,27 @@ func (a *analyzer) semOp(o ast.Op, seq dag.Seq) (dag.Seq, error) {
...
+			// If both paths are static validate them. Otherwise this will be
+			// done at runtime.
+			srcThis, srcOk := src.(*dag.This)
+			dstThis, dstOk := dst.(*dag.This)
+			if dstOk && srcOk {
+				if err := expr.CheckRenameField(srcThis.Path, dstThis.Path); err != nil {
+					return nil, fmt.Errorf("rename: %w", err)
 				}
 			}
 			assignments = append(assignments, dag.Assignment{Kind: "Assignment", LHS: dst, RHS: src})
```
nwt, line 666: Simplify with semAssignments?
```go
			as, err := a.semAssignment(fa)
			if err != nil {
				return nil, fmt.Errorf("rename: %w", err)
			}
			if !isLval(as.RHS) {
				return nil, fmt.Errorf("rename: illegal right-hand side of assignment")
			}
			// If both paths are static validate them. Otherwise this will be
			// done at runtime.
			lhs, lhsOk := as.LHS.(*dag.This)
			rhs, rhsOk := as.RHS.(*dag.This)
			if lhsOk && rhsOk {
				if err := expr.CheckRenameField(lhs.Path, rhs.Path); err != nil {
					return nil, fmt.Errorf("rename: %w", err)
				}
			}
			assignments = append(assignments, as)
```

mattnibs: I like it.

nwt (approved):
:shipit: when CI is green.

PR #4805, opened 2023-10-12
Semantically check function calls

### LINKED ISSUE #4593: No implied "yield" with cast expression
Repro is with Zed commit a2626a7.

This works:

```
$ zq -version
Version: v1.7.0-66-ga2626a7c

$ echo '{ts: "2023-05-17T19:19:32Z"}' | zq -z 'yield time(ts)' -
2023-05-17T19:19:32Z
```

But this fails:

```
$ echo '{ts: "2023-05-17T19:19:32Z"}' | zq -z 'time(ts)' -
time(): no such function
```

Per the [Implied Operators docs](https://zed.brimdata.io/docs/language/overview#26-implied-operators), as a user I expect `time(ts)` to have been interpreted as an expression and hence it should have worked without the explicit `yield`.

Issue #4556 seems like it might share some root cause with this one, so whoever takes up one might want to consider the other at the same time.

mattnibs:
Previously most checking of function calls was done in kernel, but move this to the semantic pass. Also change the parser so it doesn't convert a primitive function call to cast and instead lets the semantic pass do this.

Closes #4593


runtime/expr/function/function.go
```diff
@@ -165,13 +165,20 @@ func New(zctx *zed.Context, name string, narg int) (expr.Function, field.Path, e
 	case "unflatten":
 		f = NewUnflatten(zctx)
 	}
+	if err := CheckArgCount(narg, argmin, argmax); err != nil {
+		return nil, nil, err
+	}
+	return f, path, nil
+}
+
+func CheckArgCount(narg int, argmin int, argmax int) error {
```
nwt, line 174: 👍

compiler/semantic/expr.go
```diff
@@ -471,26 +472,48 @@ func (a *analyzer) semCall(call *ast.Call) (dag.Expr, error) {
...
+		if len(f.Params) != nargs {
+			return nil, fmt.Errorf("%s(): expects %d argument(s)", name, len(f.Params))
+		}
+	case ptyp != nil:
+		// Primitive function call, change this to a cast.
+		if err := function.CheckArgCount(nargs, 1, 1); err != nil {
+			return nil, fmt.Errorf("%s(): %w", name, err)
+		}
+		name = "cast"
+		exprs = append(exprs, &dag.Literal{Kind: "Literal", Value: fmt.Sprintf("<%s>", zson.FormatType(ptyp))})
```
nwt, line 501:
```suggestion
	case zed.LookupPrimitive(name) != nil:
		// Primitive function call, change this to a cast.
		if err := function.CheckArgCount(nargs, 1, 1); err != nil {
			return nil, fmt.Errorf("%s(): %w", name, err)
		}
		exprs = append(exprs, &dag.Literal{Kind: "Literal", Value: "<"+name+">"})
		name = "cast"
```

compiler/semantic/expr.go
```diff
@@ -471,26 +472,48 @@ func (a *analyzer) semCall(call *ast.Call) (dag.Expr, error) {
...
+	case shaper != 0:
```
nwt, line 502:
```suggestion
	case expr.NewShaperTransform(name) != 0:
```

compiler/parser/ztests/primitive-cast.yaml
```diff
@@ -0,0 +1,14 @@
+script: |
+  zc -C -s 'int64(this)'
+  zc -C -s 'time(this)'
+  zc -C -s 'string(this)'
```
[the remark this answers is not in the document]
mattnibs: So just don't test this since there's no change to input/output? I guess I'm ok with that

nwt (approved):

PR #4806, opened 2023-10-12
Add map function

### LINKED ISSUE #4610: Call a function on every element of a complex value
Repro is with Zed commit efdcd27.

Imagine you have an complex value such as the array of numbers `[1, 4, 9, 16]` and you want to execute some function on each element to produce a new array, such as multiplying each element by two. To accomplish this in Zed today requires a multi-part idiom:

```
$ zq -version
Version: v1.8.0-5-gefdcd272

$ echo '[1, 4, 9, 16]' | zq -z 'over this | yield this*2 | collect(this)' -
[2,8,18,32]
```

Compare this to something like [JavaScript's `map()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) (not to be confused with Zed's [`map` type](https://zed.brimdata.io/docs/formats/zed#24-map)) which can perform the equivalent in one shot. We've recognized Zed should offer a similar shorthand.

While in JavaScript we see this approach applied to array types, it seems that in Zed it could be used for not only arrays but also sets, maybe maps, and perhaps into records (recursively?) though I'm not sure. Implementation TBD, I suppose.

mattnibs:
The map function that applies a function to every element in an array or set value.

Also: rename the map aggregator to collect_map to avoid collision with this new function.

Closes #4610


compiler/ast/dag/expr.go
```diff
@@ -62,6 +62,11 @@ type (
 		Kind  string `json:"kind" unpack:""`
 		Value string `json:"value"`
 	}
+	MapCall struct {
+		Kind  string `json:"kind" unpack:""`
+		Expr  Expr   `json:"expr"`
+		Inner Expr   `json:"inner"`
+	}
```
nwt, line 69: Is there a downside to using `Call` rather than adding `MapCall`?

mattnibs: Its kind of a different thing than a call. A call is made of args that you get the value of all by passing `this` into it, a map call you evaluate this, then iterate over each value an call the inner eval on each value. So I kind of think this should remain MapCall

runtime/expr/agg/map.go
```diff
@@ -6,13 +6,13 @@ import (
 	"golang.org/x/exp/slices"
 )
 
-type Map struct {
+type CollectMap struct {
```
nwt, line 9: Please rename this file to collectmap.go

runtime/expr/agg/map.go
```diff
@@ -22,7 +22,7 @@ type mapEntry struct {
 	val *zed.Value
 }
 
-func (m *Map) Consume(val *zed.Value) {
+func (m *CollectMap) Consume(val *zed.Value) {
```
nwt, line 25: Nit: Change receiver name to `c`.

docs/language/functions/README.md
```diff
@@ -40,6 +40,7 @@ Zed's [primitive types](../../formats/zed.md#1-primitive-types), e.g.,
 * [levenshtein](levenshtein.md) Levenshtein distance
 * [log](log.md) - natural logarithm
 * [lower](lower.md) - convert a string to lower case
+* [map](map.md) - apply a function to each element of an array
```
nwt, line 43:
```suggestion
* [map](map.md) - apply a function to each element of an array or set
```

docs/language/functions/map.md
```diff
@@ -0,0 +1,42 @@
+### Function
+
+&emsp; **map** &mdash; calls a function on each element of an array and returns the results
```
nwt, line 3:
```suggestion
&emsp; **map** &mdash; apply a function to each element of an array or set
```

runtime/expr/map.go
```diff
@@ -0,0 +1,71 @@
...
+	if err != nil {
+		return ectx.CopyValue(*a.zctx.WrapError(err.Error(), in))
+	}
+	if len(elems) == 0 {
+		return v
+	}
+	a.vals = a.vals[:0]
+	a.types = a.types[:0]
+	for _, elem := range elems {
+		out := a.inner.Eval(ectx, &elem)
```
nwt, line 39: Nit: s/out/val/ because that's what we usually call a zed.Value.

docs/language/functions/map.md
```diff
@@ -0,0 +1,42 @@
...
+
+### Synopsis
+
+```
+map(v: array|set, f: function) -> array|set
```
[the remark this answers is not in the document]
philrz: @nwt: Can you say more about why this is your preference? It feels strange to me for the data type to be changed as a side effect of an operation like this. If we end up going with your suggestion, then I'd be in favor of the docs mentioning your `|[...map(v,f)]|` workaround in the same doc right below this function signature. Those of us who work on Zed knowing something is possible is not the same thing as it being something average users will think to put together. I think of the spread operator as a more advanced part of the language, so leaving this as an "exercise for the reader" doesn't sit with me right.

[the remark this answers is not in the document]
mattnibs: Yeah I agree with @philrz. The map function is really targeting each element in the array/set and it feels weird to have it change the type of the container.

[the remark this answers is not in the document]
nwt: I suggest this because having `map(some_large_set, is_interesting)` return `|[true, false]|` isn't useful.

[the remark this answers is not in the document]
philrz: @nwt: Maybe I'm missing some context, but I'm not sure how that example is the one that would make the case for this. I admittedly don't find myself reaching for the `set` data type very often, but to reach for our old friend Zeek that's always happened to have `set` data, I can see places to use this, e.g. with the [zed-sample-data](https://github.com/brimdata/zed-sample-data):

```
$ zq -version
Version: v1.10.0-15-g7be4063c

$ zq -Z 'lower_analyzers:=map(analyzers, lower) | cut analyzers,lower_analyzers | head 1' files.log.gz
{
    analyzers: |[
        "MD5",
        "SHA1"
    ]|,
    lower_analyzers: [
        "md5",
        "sha1"
    ]
}
```

Indeed, that shows it's turning the set into an array even in the current state of the branch, and I'm still not sure why that's helpful here to me as a user.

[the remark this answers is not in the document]
mattnibs: In the case of `map(some_large_set, is_interesting)` it seems like the thing to do would be to convert `some_large_set` into an array if you don't want the output values to get squashed. The normal case IMHO would be something like what @philrz is proposing.

docs/language/functions/map.md
```diff
@@ -0,0 +1,42 @@
...
+```
+
+### Description
+
+The _map_ function applies function f to every element in array/set v and
+returns the augmented array/set. Function f must be a function that takes
```
nwt, line 14: Nits: "or" is classier than "/". And "augmented" suggests that values returned by f are added to v.
```suggestion
The _map_ function applies function f to every element in array or set v and
returns an array or set of the results. Function f must be a function that takes
```

docs/language/functions/map.md
```diff
@@ -0,0 +1,42 @@
...
+Using a user defined function to convert an epoch float to a time:
+
+```mdtest-input udf.zed
+func floatToTime(x): ( cast(x*1000000000, <time>) )
+yield map(this, floatToTime)
+```
+
+```mdtest-command
+echo '[1697151533.41415,1697151540.716529]' | zq -z -I udf.zed -
+```
```
nwt, line 38: Nit: I think this would be clearer without udf.zed.
```mdtest-command
echo '[1697151533.41415,1697151540.716529]' | zq -z '
  func floatToTime(x): ( cast(x*1000000000, <time>) )
  yield map(this, floatToTime)
' -
```

nwt (approved):

PR #4807, opened 2023-10-16
Parser error for functions starting with a token

mattnibs:
Fix parsing error when a function starts with the true, false or null tokens.

### DIFF
--- a/compiler/parser/parser.peg
+++ b/compiler/parser/parser.peg
@@ -1051,11 +1051,11 @@ IntegerLiteral
     }
 
 BooleanLiteral
-  = "true"  { RETURN(MAP("kind": "Primitive", "type": "bool", "text": "true")) }
-  / "false" { RETURN(MAP("kind": "Primitive", "type": "bool", "text": "false")) }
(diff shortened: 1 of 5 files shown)


compiler/parser/parser.peg
```diff
@@ -1190,9 +1190,12 @@ FieldName
...
 AndToken = ("and" / "AND") !IdentifierRest { RETURN("and") }
 OrToken = ("or" / "OR") !IdentifierRest { RETURN("or") }
-InToken = "in" !IdentifierRest { RETURN("in") }
 NotToken = ("not" / "NOT") !IdentifierRest { RETURN("not") }
-ByToken = "by" !IdentifierRest { RETURN("by") }
+ByToken = "by" !IdentifierRest
+InToken = "in" !IdentifierRest
+NullToken = "null" !IdentifierRest
+TrueToken = "true" !IdentifierRest
+FalseToken = "false" !IdentifierRest
```
nwt, line 1198: Maybe alphabetize these?

nwt (approved):

compiler/parser/parser.peg
mattnibs: Sure but I think it `true` and `false` should be grouped together.

PR #4808, opened 2023-10-17
zq: Call user-defined op with source

### LINKED ISSUE #4701: "redundant inputs" failure when calling -I included User Op
Repro is with Zed commit 7ba4ecc.

The following all work ok.

```
$ zq -version
Version: v1.8.1-59-g7ba4ecc6

$ zq 'file sample.zng | count()'
31(uint64)

$ cat file_main.zed 
file sample.zng | count()

$ zq -I file_main.zed 
31(uint64)

$ cat file_in_op.zed
op CallMe(): (
  file sample.zng | count()
)
CallMe()

$ zq -I file_in_op.zed
31(uint64)
```

However, if I have _just_ the User Op in my included Zed and attempt to make the call to it at the shell level, now I get an error.

```
$ cat file_in_op_no_call.zed
op CallMe(): (
  file sample.zng | count()
)

$ zq -I file_in_op_no_call.zed 'CallMe()'
redundant inputs specified: use either command-line files or a Zed source operator
```

I'd not be surprised if there's ways to trigger this that don't involve User Ops, but since I bumped into it accidentally while attempting some basic User Ops use cases, I'm opening it in that context.

mattnibs:
The change allows zq users to start a query with a call to a user-defined op that contains a source (file) with no additional input sources. Previously attempting to do this would result in error "redundant inputs".

Fixes #4701

### DIFF
--- a/cli/queryflags/flags.go
+++ b/cli/queryflags/flags.go
@@ -1,6 +1,7 @@
 package queryflags
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"net/url"
@@ -9,8 +10,8 @@ import (
 	"github.com/brimdata/zed/cli"
 	"github.com/brimdata/zed/compiler"
 	"github.com/brimdata/zed/compiler/ast"
+	"github.com/brimdata/zed/compiler/semantic"
 	"github.com/brimdata/zed/zbuf"
-	"github.com/brimdata/zed/zfmt"
 	"github.com/brimdata/zed/zson"
 	"golang.org/x/exp/slices"
 )
@@ -36,11 +37,13 @@ func (f *Flags) ParseSourcesAndInputs(paths []string) ([]string, ast.Seq, bool, 
 			// and appears to start with a from or yield operator.
 			// Otherwise, consider it a file.
 			if query, err := compiler.Parse(src, f.Includes...); err == nil {
-				if isFrom(query) {
-					return nil, query, false, nil
-				}
-				if isYield(query) {
-					return nil, query, true, nil
+				if s, err := semantic.Analyze(context.Background(), query, nil, nil); err == nil {
+					if semantic.HasSource(s) {
+						return nil, query, false, nil
+					}
+					if semantic.StartsWithYield(s) {
+						return nil, query, true, nil
+					}
 				}
 			}
 			return nil, nil, false, fmt.Errorf("no such file: %s", src)
@@ -51,32 +54,6 @@ func (f *Flags) ParseSourcesAndInputs(paths []string) ([]string, ast.Seq, bool, 
 		return nil, nil, false, err
 	}
 	return paths, query, false, nil
-}
-
-func isFrom(seq ast.Seq) bool {
-	if len(seq) > 0 {
-		switch op := seq[0].(type) {
-		case *ast.From:
-			return true
-		case *ast.Scope:
-			return isFrom(op.Body)
-		}
-	}
-	return false
-}
-
-func isYield(seq ast.Seq) bool {
-	if len(seq) > 0 {
-		switch op := seq[0].(type) {
-		case *ast.Yield:
-			return true
-		case *ast.Scope:
-			return isYield(op.Body)
-		case *ast.OpExpr:
-			return !zfmt.IsSearch(op.Expr) && !zfmt.IsBool(op.Expr)
-		}
-	}
-	return false
 }
 
 func isURLWithKnownScheme(path string, schemes ...string) bool {
--- a/compiler/job.go
+++ b/compiler/job.go
@@ -40,7 +40,7 @@ func NewJob(octx *op.Context, in ast.Seq, src *data.Source, head *lakeparse.Comm
 	if len(seq) == 0 {
 		return nil, errors.New("internal error: AST seq cannot be empty")
 	}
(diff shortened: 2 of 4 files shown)


philrz:
Testing this branch at commit 777dc58, it seems to address #4701. The final command shown below now returns the expected output rather than the `redundant inputs` error shown in #4701.

```
$ zq -version
Version: v1.10.0-11-g777dc583

$ cat file_in_op_no_call.zed
op CallMe(): (
  file sample.zng | count()
)

$ zq -I file_in_op_no_call.zed 'CallMe()'
31(uint64)
```

That makes it a functional 👍 for me. I'll mark it in the opening text as "Fixes #4701" so the issue will close when this merges.

cli/queryflags/flags.go
```diff
@@ -36,11 +37,13 @@ func (f *Flags) ParseSourcesAndInputs(paths []string) ([]string, ast.Seq, bool,
...
-				if isYield(query) {
-					return nil, query, true, nil
+				if a, err := semantic.NewAnalyzer(context.Background(), query, nil, nil); err == nil {
+					if a.HasSource() {
+						return nil, query, false, nil
+					}
+					if a.StartsWithYield() {
+						return nil, query, true, nil
+					}
 				}
```
nwt, line 47: This interface feels off because the methods interrogate the DAG, not the semantic analyzer.

How about something like this instead?
```suggestion
				if queryDag, err := semantic.Analyze(context.Background(), query, nil, nil); err == nil {
					if semantic.HasSource(queryDag) {
						return nil, query, false, nil
					}
					if semantic.StartsWithYield(queryDag) {
						return nil, query, true, nil
					}
				}
```

mattnibs: I can dig that.

mattnibs: ok

compiler/lake.go
```diff
@@ -99,17 +99,17 @@ func newDeleteJob(octx *op.Context, in ast.Seq, src *data.Source, head *lakepars
...
-	if _, ok := entry[1].(*dag.Filter); !ok {
+	if _, ok := s[1].(*dag.Filter); !ok {
 		return nil, &InvalidDeleteWhereQuery{}
 	}
 	return &Job{
 		octx:      octx,
 		builder:   kernel.NewBuilder(octx, src),
 		optimizer: optimizer.New(octx.Context, src),
-		entry:     entry,
+		entry:     s,
```
nwt, line 113: Nit: I think `entry` is a better name here.

compiler/job.go
```diff
@@ -40,17 +40,17 @@ func NewJob(octx *op.Context, in ast.Seq, src *data.Source, head *lakeparse.Comm
...
 	}
-	reader, _ := entry[0].(*kernel.Reader)
+	reader, _ := s[0].(*kernel.Reader)
 	return &Job{
 		octx:      octx,
 		builder:   kernel.NewBuilder(octx, src),
 		optimizer: optimizer.New(octx.Context, src),
 		reader:    reader,
-		entry:     entry,
+		entry:     s,
```
nwt, line 53: Nit: I think `entry` is a better name here.

compiler/semantic/analyzer.go
```diff
@@ -49,26 +58,26 @@ func newAnalyzer(ctx context.Context, source *data.Source, head *lakeparse.Commi
...
-	a.scope = NewScope(a.scope)
-}
-
-func (a *analyzer) exitScope() {
-	a.scope = a.scope.parent
-}
-
-func (a *analyzer) buildFrom(op dag.Op) (dag.Op, error) {
-	switch op := op.(type) {
+func HasSource(s dag.Seq) bool {
```
nwt, line 61: Nit: We call a dag.Seq parameter `seq` everywhere else.
```suggestion
func HasSource(seq dag.Seq) bool {
```

compiler/semantic/analyzer.go
```diff
@@ -49,26 +58,26 @@ func newAnalyzer(ctx context.Context, source *data.Source, head *lakeparse.Commi
...
-		return a.buildFrom(op.Paths[0][0])
+		return HasSource(op.Paths[0])
 	case *dag.Scope:
-		return a.buildFrom(op.Body[0])
+		return HasSource(op.Body)
+	}
+	return false
+}
+
+func (a *analyzer) addDefaultSource(s *dag.Seq) error {
```
nwt, line 73: Nit: We call a dag.Seq parameter `seq` everywhere else.
```suggestion
func (a *analyzer) addDefaultSource(s *dag.Seq) error {
```

compiler/semantic/analyzer.go
```diff
@@ -81,9 +90,28 @@ func (a *analyzer) buildFrom(op dag.Op) (dag.Op, error) {
...
 	ops, err := a.semPool(pool)
 	if err != nil {
-		return nil, err
+		return err
+	}
+	s.Prepend(ops[0])
+	return nil
+}
+
+func StartsWithYield(s dag.Seq) bool {
```
nwt, line 99: Nit: We call a dag.Seq parameter `seq` everywhere else.
```suggestion
func StartsWithYield(s dag.Seq) bool {
```

nwt (approved):

PR #4811, opened 2023-10-20
Add docs related to zed/4758 changes

philrz:
This is my attempt to update the docs to reflect what changed in #4758. This took longer than I thought because the flexibility of having `-lake`, multiple possible environment variables, and the new fallback default paths play with the different `zed` personalities in many ways. Thankfully the hints from the `zed` CLI seem to point the user in the right direction if they attempt a combo that's not intended to work, so I stopped short of trying to enumerate every combo and instead continue to lean on what was already written in the "personalities" section to fill in some of the detail (e.g., when URLs can be used vs. file system paths).

I'm expecting 99% of users will stay on the happy path, i.e., when they use `zed` in its client personality they won't have specified `-lake` or any environment variables but if their Zui is running it'll have started a lake service and it'll "just work". We do still need to make changes on the Zui side to make it "just work" in the other direction (i.e., so Zui will start defaulting to using storage at `%LOCALAPPDATA%\zed` on Windows or `$HOME/.local/share/zed` on Linux/macOS), but that'll be my next conquest once folks sign off on what's in this PR.

All that said, improvements welcomed!

### DIFF
--- a/docs/commands/zed.md
+++ b/docs/commands/zed.md
@@ -74,7 +74,7 @@ components come together.
 While the CLI-first approach provides these benefits,
 all of the functionality is also exposed through [an API](../lake/api.md) to
 a Zed service.  Many use cases involve an application like
-[Zui](https://github.com/brimdata/zui) or a
+[Zui](https://zui.brimdata.io/) or a
 programming environment like Python/Pandas interacting
 with the service API in place of direct use with the `zed` command.
 
@@ -130,18 +130,17 @@ model onto a file system so that Zed lakes can also be deployed on standard file
 
 The `zed` command provides a single command-line interface to Zed lakes, but
 different personalities are taken on by `zed` depending on the particular
-sub-command executed and the disposition of its `-lake` option
-(which defaults to the value of `ZED_LAKE` environment variable or,
-if `ZED_LAKE` is not set, to the client personality `https://localhost:9867`).
+sub-command executed and where the [lake is located](#locating-the-lake).
 
 To this end, `zed` can take on one of three personalities:
+
 * _Direct Access_ - When the lake is a storage path (`file` or `s3` URI),
 then the `zed` commands (except for `serve`) all operate directly on the
 lake located at that path.
 * _Client Personality_ - When the lake is an HTTP or HTTPS URL, then the
 lake is presumed to be a Zed lake service endpoint and the client
 commands are directed to the service managing the lake.
-* _Server Personality_ - When the `zed serve` command is executed, then
+* _Server Personality_ - When the [`zed serve`](#serve) command is executed, then
 the personality is always the server personality and the lake must be
 a storage path.  This command initiates a continuous server process
 that serves client requests for the lake at the configured storage path.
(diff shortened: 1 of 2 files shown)


docs/commands/zed.md
```diff
@@ -167,6 +166,25 @@ all adhere to the consistency semantics of the Zed lake.
...
+`zed` checks each of the following in order to attempt to locate an existing
+lake.
+
+1. The contents of the `-lake` option (if specified)
+2. The contents of the `ZED_LAKE` environment variable (if defined)
+3. A `zed` subdirectory below a path in the
+   [`XDG_DATA_HOME`](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)
+   environment variable (if defined)
+4. A Zed lake service running locally at `http://localhost:9867` (if a socket
+   is listening at that port)
```
nwt, line 183: http://localhost:9867 has precedence over XDG_DATA_HOME.

philrz: @nwt: Thanks for spotting that! It's a subtlety I didn't pick up on in my black box testing. I've swapped the order in the list.

docs/commands/zed.md
```diff
@@ -167,6 +166,25 @@ all adhere to the consistency semantics of the Zed lake.
...
+At times you may want the Zed CLI tools to access the same lake storage
+used by other tools such as [Zui](https://zui.brimdata.io/). To help
+enable this by default while allowing for separate lake storage when desired,
+`zed` checks each of the following in order to attempt to locate an existing
+lake.
+
+1. The contents of the `-lake` option (if specified)
+2. The contents of the `ZED_LAKE` environment variable (if defined)
+3. A Zed lake service running locally at `http://localhost:9867` (if a socket
+   is listening at that port)
```
nwt, line 180: Maybe tighten this?
```suggestion
3. A Zed lake service running at `http://localhost:9867`
```

docs/commands/zed.md
```diff
@@ -167,6 +166,25 @@ all adhere to the consistency semantics of the Zed lake.
...
+4. A `zed` subdirectory below a path in the
+   [`XDG_DATA_HOME`](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)
+   environment variable (if defined)
```
nwt, line 183: Maybe tighten this?
```suggestion
4. `$XDG_DATA_HOME/zed` (if the `XDG_DATA_HOME` environment variable is defined)
```

docs/commands/zed.md
```diff
@@ -167,6 +166,25 @@ all adhere to the consistency semantics of the Zed lake.
...
+5. A default file system location based on detected OS platform:
+   - `%LOCALAPPDATA%\zed` on Windows
+   - `$HOME/.local/share/zed` on Linux/macOS
```
nwt, line 186: Nit: "and" is classier.
```suggestion
   - `$HOME/.local/share/zed` on Linux and macOS
```

nwt (approved):

docs/commands/zed.md
philrz: @nwt: I actually kept this one un-tight intentionally. During black box testing I noticed that if I run a simple program that just listens on `9867` and accepts the incoming connection, and the lake location logic makes it to this step, `zed` effectively hangs indefinitely. I guess it's a hair-split to think that the text as I wrote it communicates this subtlety. But I was trying to express that all it takes is a listening socket for it to commit to this step.

```
#!/usr/local/bin/python3
import socket
import sys

if (len(sys.argv) != 2 or not sys.argv[1].isdigit()):
  print('Usage: listen <port>')
  exit()

p = int(sys.argv[1])
l = []
s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
s.bind(('', p))
s.listen(1)
while 1:
  (c, a) = s.accept()
  l.append(c)
  print('%d: connection from %s' % (len(l), a))
```

docs/commands/zed.md
philrz: @nwt: This one was also intentionally un-tight because even on Windows the `XDG_DATA_HOME` environment variable may be obeyed. Obviously users have their choice of shell on Windows, and some use `$` to expand. But since the out-of-the-box Command Prompt uses dual `%` to expand environment variables, I was keeping it higher level. Maybe another futile hair-split, though.

docs/commands/zed.md
nwt: @philrz: That's a bug!

philrz: @nwt: Cool! Glad we discussed. I've just opened https://github.com/brimdata/zed/issues/4824 to track this.

PR #4823, opened 2023-10-25
DONT MERGE. Vam.

jamii:
(I'm just opening this to get the review ui).


vector/array.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+type Array struct {
+	mem
+	Typ     *zed.TypeArray //XXX type array or set
+	Lengths []int32
+	Values  Any
+}
+
+var _ Any = (*Array)(nil)
```
jamii, line 14: This is an assertion that it fits the `Any` interface?

vector/record.go
```diff
@@ -0,0 +1,55 @@
+package vector
+
+import (
+	"github.com/brimdata/zed"
+	"github.com/brimdata/zed/zcode"
+)
+
+// XXX need to create memory model
+type mem struct{}
```
jamii, line 9: What is this for?

vector/array.go
nwt: Exactly.

vector/string.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"github.com/brimdata/zed"
+	"github.com/brimdata/zed/zcode"
+)
+
+type String struct {
+	mem
+	Typ    zed.Type
+	Values []string
+	Nulls  Nullmask
```
jamii, line 12: Why are strings in particular nullable?

vector/nulls.go
```diff
@@ -0,0 +1,23 @@
+package vector
+
+type Nullmask []byte //XXX change to uint64
```
jamii, line 3: How about

```
type Nullable struct {
    mask []byte
    vector Any
}
```

Then we don't have to embed the null mask in every other type.

vector/record.go
nwt: Looks like it's just a way to add no-op Ref and Unref methods to the types that embed this.

runtime/vcache/primitive.go
```diff
@@ -1,68 +1,144 @@
...
+	case *zed.TypeOfString:
+		var vals []string
+		var nullslots []uint32
+		it := zcode.Bytes(bytes).Iter()
+		for !it.Done() {
+			val := it.Next()
+			if val == nil {
+				nullslots = append(nullslots, uint32(len(vals)))
+			} else {
+				vals = append(vals, zed.DecodeString(val))
```
jamii, line 92: `zed.DecodeString(val)`

If I'm reading this right this is not a copy - all the resulting strings share the same underlying byte buffer?

jamii:
Notes:

* Adds a command that tests the aggs.
* Alters the project command to use vectors.
* Adds a VecScan op which projects some paths.
  * Exec in compiler/kernel/op
  * Work happens in VecScanner
* Adds some hacky aggregate ops.
  * (Ignoring these for now)
* Adds vcache to lake.
* Adds Index - a selection vector.
  * Adds commented out materialization code.
* vam/projection splits into vectors by type (should be replaced by producing Union vector where appropriate)
* Adds VecScanner. 
  * parent feeds lake uris
  * (Don't follow the cancellation logic yet)
  * Loads a vector for each (type, path) combo
     * Probably path should include union branches instead.
* Changes vcache loading code to call loadVector. 
  * array/set (only handles ints)
  * map
  * nulls (not finished)
  * object (on demand?)
  * primitives (only a few types)
    * stored as zng plus precomputed offsets
    * commented out code for dictionary compression?
  * records (comment about missing locking?)
  * union
* Adds a mutex to vcache cache fetch.
* Deletes old vcache projection code.
* Comments out vcache reader.
* Adds vector.
  * Any is the interface.
  * Builder takes a zcode.Builder and adds 1 value to it.
     * (zcode.Builder just helps track lengths of nested trees in zng)
  * mem ???
  * Const probably needs a length field.
  * String vector

runtime/vcache/primitive.go
nwt: That's what we want but there's an implicit copy here as the compiler will insert a call to runtime.slicebytetostring when it sees this. Our usual way around that is this: https://github.com/brimdata/zed/blob/40639dce6c333da1c55844b026052dc3718cc2d2/pkg/byteconv/byteconv.go#L11-L16

vector/record.go
jamii: Are the Ref/Unref methods some existing idiom in the codebase?

nwt: Yes, they appear in zbuf.Batch.

https://github.com/brimdata/zed/blob/40639dce6c333da1c55844b026052dc3718cc2d2/zbuf/batch.go#L12-L34

Issue #3151, opened 2021-10-06
"Live" data import with Logstash


philrz:
I recently ran an experiment to see what would happen if I used an external agent (specifically, [Logstash](https://www.elastic.co/logstash/)) to push data to a Zed lake rather than using `zapi load`. I'll note up front that Logstash is surely not current state-of-the-art in terms of import agents, but it's one with which I'm already familiar, so I figured I'd start there in the interest of time.

Overall, it showed that the necessary pieces are all in place, but there were some speedbumps along the way that point to issues that may be worth addressing. To that end, I've started this issue off as an Epic just in case folks want to start linking additional issues to it.

The steps below were performed with Brim commit 5da5e7d (which uses Zed commit `fe1cb75`) and Logstash `7.15.0`. I also happened to use Zeek `4.1.1` running on my local laptop as the "live" data source, just because it was familiar and close at hand. Other than pointing it at the wireless interface on my Mac laptop to sniff "live" traffic, the only other change from Zeek defaults was the addition of the j[son-streaming-logs](https://github.com/corelight/json-streaming-logs) package. I also did not do any tuning to Logstash, using just this simple config file `logstash.conf`:

```
input {
  file {
    path => "/usr/local/zeek-4.1.1/logs/current/json_streaming_*"
  }
}

filter {
  json {
    source => "message"
  }
  mutate {
    remove_field => [ "message", "@timestamp", "@version", "path", "host" ]
  }
}

output {
  stdout {
    codec => "json"
  }
  http {
    url => "http://127.0.0.1:9867/pool/1ys2uIgnrlu5iM9hGc8ObhhxBOp/branch/main"
    http_method => "post"
    pool_max => 1
  }
}
```

In brief, the `input` stanza watches the logs generated by the json-streaming-logs package in a `tail`\-like manner. The `filter` stanza ensures that the lines in the log file are interpreted as JSON and turned into key/value pairs, then a bunch of Logstash's automatically-added metadata fields are dropped so they won't appear in the output. Finally, the `output` stanza sends a copy of all the posted data to the console for debug purposes, but most importantly posts data to the Zed API endpoint that accepts data to be loaded to a pool. The `pool_max` setting also turns out to be important because by default Logstash attempts to open 50 connections in parallel to post data in the name of performance, but Zed doesn't care for this, so we've lowered it to just one connection (thanks to @nwt for isolating the importance of this setting!)

[...]

PR #4827, opened 2023-10-26
Add grok function

### LINKED ISSUE #4140: Parsing text lines into records (Grok)
A community user asked in a [public Slack thread](https://brimdata.slack.com/archives/CTSMAK6G7/p1665158278523859):

> What would the best process be to parse a field of large text (syslog events in this case) into a record so that data within the longer string can be easily searched and integrated with zed query syntax?

For now the user managed to get by using the `split()` function and then using record literal syntax to assign field names to the value of each element of the resulting array. However, the first thing I thought of upon hearing the inquiry was Grok, as tools like Logstash have long used this to do the heavy lifting for such parsing. Some of the benefits of Grok is that there's off-the-shelf patterns that can be invoked to parse standard log formats as well as debug tools like https://grokdebug.herokuapp.com/ that help users create their own custom parsers.

I can see that there's a Go library https://github.com/vjeantet/grok, so maybe it would not be too difficult to add a Zed function that uses it.

If we ever wanted to get fancy, it seems we could extend the Grok syntax to also assign a Zed data type to the created fields. There's some precedent here in the Grok that's used in Logstash as described in [this StackOverflow post](https://stackoverflow.com/questions/29742482/logstash-output-to-elasticsearch-with-valid-types#comment47618824_29742787). In their case they only support destination data types for `int` and `float`, but Zed could differentiate by supporting its full set of rich types. I assume the Go library is only generating strings, but perhaps we could extend it.

mattnibs:
Closes #4140


philrz:
## tl;dr

I've done some preliminary testing on this branch at commit 2ad8892. The tl;dr is that it seems like what's already working here meets a defensible definition of "doing Grok". I'd not stand in the way of shipping it as is and seeing what users do with it. That said, I did hit a few speedbumps in my testing such that if we ship what we've got, I'd feel like we'd need to highlight several caveats in the docs and be prepared for users to struggle and frequently come to Slack for help. In this regard it may be a pay-now-or-later zero-sum game where it might be worth addressing some gaps in the code now if it's not too difficult so it'll mean less user/support pain in the future. I'll dive into details in separate sections below.

## Grok specification

When reading through dozens of online references, it's clear how under-specified Grok is. It's not like there's an RFC or other standard that lays out what specifically Grok is/isn't. That said, almost all pages point back at the [Grok filter plugin in Logstash](https://www.elastic.co/guide/en/logstash/current/plugins-filters-grok.html#plugins-filters-grok) as if it's the reference implementation. Further digging finds a [`grok` CLI tool](https://github.com/jordansissel/grok) written by Jordan Sissel who is basically "the Logstash guy". The first commit on that CLI tool was in 2007 while the first commit in [the Logstash repo](https://github.com/elastic/logstash) was in 2009. But the CLI tool is barely referenced in Google search results and hasn't seen a commit since 2017, so I'm going with the consensus that Logstash is the reference implementation.

## Testing

After reproducing the example in the docs included in this PR, the next thing I wanted to test was using a "custom pattern" via the `extra` parameter. Due to an (un-)lucky coincidence, the way I went about this was to follow an example I found in [this article](https://www.elastic.co/blog/slow-and-steady-how-to-build-custom-grok-patterns-incrementally). I figured it was a fine choice since it's written by Elastic (i.e., the people who bring us Logstash) and has a "for dummies" feel (so, our own users that don't know Grok might stumble onto it and see it as a good way to get started). Unfortunately, the example shown doesn't work as is, even before I bring the `extra` parameter into the picture.

```
$ zq -version
Version: v1.10.0-13-g2ad88925

$ echo '"55.3.244.1 GET /index.html 15824 0.043 other stuff"' | zq -Z 'yield grok("%{IP:host.ip} %{WORD:http.request.method} %{URIPATHPARAM:url.original} %{NUMBER:http.request.bytes:int} %{NUMBER:event.duration:double} %{GREEDYDATA:my_greedy_match}", this)' -
error("grok(): no match")
```

After further digging, there's two things in this example that don't seem to be supported with the current implementation in this branch, resulting in a failure to parse.

1. The trailing "data type conversion" specifications (e.g., `:int`).
2. Field names containing dots (e.g, `http.request.method`).

If I drop the data types and change the dots to something else like underscores, now it parses.

```
$ echo '"55.3.244.1 GET /index.html 15824 0.043 other stuff"' | zq -Z 'yield grok("%{IP:host_ip} %{WORD:http_request_method} %{URIPATHPARAM:url_original} %{NUMBER:http_request_bytes} %{NUMBER:event_duration} %{GREEDYDATA:my_greedy_match}", this)' -
{
    host_ip: "55.3.244.1",
    http_request_method: "GET",
    url_original: "/index.html",
    http_request_bytes: "15824",
    event_duration: "0.043",
    my_greedy_match: "other stuff"
}
```

Before I figured this out, I coincidentally happened to _also_ try testing out the original pattern from that same article in https://grokdebugger.com because the more-frequently-linked-to https://grokdebug.herokuapp.com seems to be dead. Well, despite grokdebugger.com having what feels like a nice UX, it turns out it _also_ fails completely when asked to create fields whose names contain dots, and this is disclosed in the [repo's README](https://github.com/cjslack/grok-debugger) along with other limitations.

I'd have ideally hoped to be able to tell users in our docs "feel free to educate yourself and steal from all the Grok materials already available online" and/or "we suggest using a Grok debugger to develop your patterns", but this one-two punch of unexpected findings gave me reason to pause and go down the rabbit hole about these gotchas.

## Proposals

### Field names containing dots

It seems there's a rich and miserable history in ELK stack as relates to field names that contain dots. [This doc from around 2015](https://www.elastic.co/guide/en/elasticsearch/reference/2.0/breaking_20_mapping_changes.html#_field_names_may_not_contain_dots) describes how field names containing dots were ok in Elasticsearch 1.x, then weren't ok in 2.0, but then [this doc](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/dots-in-names.html) describes how it became possible again in 5.0. Also regarding Elasticsearch, it's worth pointing out how [this article](https://www.elastic.co/blog/slow-and-steady-how-to-build-custom-grok-patterns-incrementally) I referenced previously actually showed Grok being used in a modern (year 2020) context of an ingest pipeline (i.e., Logstash isn't mentioned once) and interestingly it shows the dotted field names turning directly into nested fields with no special handling.

While that's all interesting, I'm once again more interested in Logstash as the reference example. This is not only because it's the one most often referenced but also because users that need Grok (or other parsing) functionality we can't yet provide natively in Zed can still use Logstash (see #3151). Here what we see is that by default Logstash leaves the dotted field names alone. (This and other Logstash tests are being performed with the current latest available, which is v8.10.4).

Using config `logstash-fields-with-dots.conf`:

```
input {
  file {
    path => "/Users/phil/work/logstash/input.txt"
  }
}

filter {
  grok {
    match => { "message" =>  "%{IP:host.ip} %{WORD:http.request.method} %{URIPATHPARAM:url.original} %{NUMBER:http.request.bytes:int} %{NUMBER:event.duration:double} %{GREEDYDATA:my_greedy_match}" }
  }
}

output {
  stdout {
    codec => "json"
  }
}
```

And input file `input.txt`:

```
55.3.244.1 GET /index.html 15824 0.043 other stuff
```

We get:

```
$ bin/logstash -f logstash-fields-with-dots.conf
{
  "message": "55.3.244.1 GET /index.html 15824 0.043 other stuff",
  "event.duration": "0.043",
  "http.request.method": "GET",
  "url.original": "/index.html",
  "@version": "1",
  "log": {
    "file": {
      "path": "/Users/phil/work/logstash/input.txt"
    }
  },
  "@timestamp": "2023-10-31T21:03:31.952526Z",
  "event": {
    "original": "55.3.244.1 GET /index.html 15824 0.043 other stuff"
  },
  "host": {
    "name": "wpm.local"
  },
  "http.request.bytes": 15824,
  "my_greedy_match": "other stuff",
  "host.ip": "55.3.244.1"
}
```

However, [this 2015 article](https://www.elastic.co/blog/introducing-the-de_dot-filter) (right around the time Elasticsearch was reckoning with its own treatment of dotted fields) Logstash added a [`de_dot` filter](https://www.elastic.co/guide/en/logstash/current/plugins-filters-de_dot.html) that can turn these into nested fields before producing output. So adding this filter to create an alternate config `logstash-de-dot.conf`:

```
input {
  file {
    path => "/Users/phil/work/logstash/input.txt"
  }
}

filter {
  grok {
    match => { "message" =>  "%{IP:host.ip} %{WORD:http.request.method} %{URIPATHPARAM:url.original} %{NUMBER:http.request.bytes:int} %{NUMBER:event.duration:double} %{GREEDYDATA:my_greedy_match}" }
  }
  de_dot {
    nested => true
  }
}

output {
  stdout {
    codec => "json"
  }
}
```

Now we see the nesting.

```
$ bin/logstash -f logstash-de-dot.conf
{
  "http": {
    "request": {
      "method": "GET",
      "bytes": 15824
    }
  },
  "message": "55.3.244.1 GET /index.html 15824 0.043 other stuff",
  "event": {
    "original": "55.3.244.1 GET /index.html 15824 0.043 other stuff",
    "duration": "0.043"
  },
  "log": {
    "file": {
      "path": "/Users/phil/work/logstash/input.txt"
    }
  },
  "@version": "1",
  "@timestamp": "2023-10-31T21:08:04.107244Z",
  "my_greedy_match": "other stuff",
  "host": {
    "name": "wpm.local",
    "ip": "55.3.244.1"
  },
  "url": {
    "original": "/index.html"
  }
}
```

Strictly speaking, `de_dot` is just one of many filters in Logstash and not "part of Grok", though.

**Conclusion:** In theory we could maintain the current behavior of the code in this branch and require users to create dot-free field names and use downstream Zed operators like `put` and `cut` to move the values to nested fields. However, in addition to this being potentially lots of extra typing for users, the fact there's dot-having examples out there in Google search results may set many users up for some total-failure-to-parse pain. Meanwhile, Zed has its own [`nest_dotted` function](https://zed.brimdata.io/docs/language/functions/nest_dotted) that's doing the equivalent of what's shown with `de_dot` above. Therefore, if it's not too difficult, I'd advocate doing what Logstash's Grok does here and just let the dot-having field names through so the users can make their own decision about whether to leave them alone or apply `nest_dotted` to them.

### Square brackets for nesting

The limitations disclosed in the [Grok debugger README](https://github.com/cjslack/grok-debugger) also made me hip to an alternate syntax for creating nested fields that may be used within the Grok pattern. This syntax is only [shown in passing in the Logstash Grok filter doc](https://www.elastic.co/guide/en/logstash/current/plugins-filters-grok.html#plugins-filters-grok-overwrite) with no attention being called to it. Further digging reveals this is because once again we're looking at a [general Logstash concept for how nested fields can be referenced](https://www.elastic.co/guide/en/logstash/current/event-dependent-configuration.html#logstash-config-field-references), i.e., not something _specific_ to Grok but that we may run into just because there's Logstash-centric Grok examples out there that use it.

When used with Logstash, it provides a handy "a la carte" way someone might create some nested fields while leaving others as dots, e.g., with config `logstash-brackets.conf`:

```
input {
  file {
    path => "/Users/phil/work/logstash/input.txt"
  }
}

filter {
  grok {
    match => { "message" =>  "%{IP:host.ip} %{WORD:[http][request][method]} %{URIPATHPARAM:url.original} %{NUMBER:http.request.bytes:int} %{NUMBER:event.duration:double} %{GREEDYDATA:my_greedy_match}" }
  }
}

output {
  stdout {
    codec => "json"
  }
}
```

And processing the same input as before, we see:

```
$ bin/logstash -f logstash-brackets.conf
{
  "message": "55.3.244.1 GET /index.html 15824 0.043 other stuff",
  "url.original": "/index.html",
  "http.request.bytes": 15824,
  "@version": "1",
  "@timestamp": "2023-10-31T21:25:49.545440Z",
  "log": {
    "file": {
      "path": "/Users/phil/work/logstash/input.txt"
    }
  },
  "event": {
    "original": "55.3.244.1 GET /index.html 15824 0.043 other stuff"
  },
  "host.ip": "55.3.244.1",
  "host": {
    "name": "wpm.local"
  },
  "event.duration": "0.043",
  "http": {
    "request": {
      "method": "GET"
    }
  },
  "my_greedy_match": "other stuff"
}
```

Whereas with what's on this branch at the moment, we're in another fail-to-parse situation. If I start from the working example above that used underscores and use just the bracketed notation for that one field, we see:

```
$ echo '"55.3.244.1 GET /index.html 15824 0.043 other stuff"' | zq -Z 'yield grok("%{IP:host_ip} %{WORD:[http][request][method]} %{URIPATHPARAM:url_original} %{NUMBER:http_request_bytes} %{NUMBER:event_duration} %{GREEDYDATA:my_greedy_match}", this)' -
error("grok(): no match")
```

**Conclusion:** Similar to the argument above regarding field names containing dots, it would be great if this syntax could create the nested fields like Logstash does just so more found examples will work out-of-the-box and users would see fewer "fail to parse" situations. That said, compared to field names containing dots, in my reading I've seen fewer examples that use the square bracket notation, so if it was deemed too much effort I'd be prepared to live without it, disclose it in the docs, and see if users notice/care.

### Data type conversion

The data type conversion syntax (e.g., `:int`) was something I'd anticipated in my write-up in #4140. Indeed, whereas the [Logstash Grok filter doc](https://www.elastic.co/guide/en/logstash/current/plugins-filters-grok.html#_grok_basics) says they only support `int` and `float` right now, I'd thought a future Zed differentiator would be to support all the Zed primitive types in this syntax to cast to our many rich types. That said, since shaping is such a first class concept in Zed (and hence our users should hopefully know what to do if they see only string data coming back from our Grok function), I'd be less inclined to beg for this in a first implementation. And supporting just `:int` and `:float` to have feature parity with Logstash feels like it would be a pretty lame mini step.

**Conclusion:** While I'm fine living without any actual casting being triggered by this syntax, the current failure-to-parse behavior seen on the code in this branch feels hazardous, once again due to the fact that there's plenty of examples out there that use it. If not too difficult, I'd be in favor of effectively recognizing the trailing `:whatever` as valid syntax but then ignoring it. That would make it so more found examples at least produce output, and then we could divulge in the docs that it's currently unsupported and point to an open issue to gather interest for doing this with the Zed types. (Incidentally, the "allow but ignore" approach is what I see in the [Grok debugger](https://grokdebugger.com) I've linked to a couple times above, though it also seems to do some clever things where it turns numeric values into numbers rather than leaving them as strings like Logstash does).

### Regular expression libraries

I'm a bit out of my league with the specifics here, but something @mattnibs and @nwt mentioned that I've also seen referenced frequently in various docs is how [Logstash's implementation uses the Oniguruma regexp library](https://www.elastic.co/guide/en/logstash/current/plugins-filters-grok.html#_custom_patterns) whereas ours currently uses the same Go one that Zed already uses in other contexts like the [`regexp` function](https://zed.brimdata.io/docs/language/functions/regexp). After trying to read some stuff online about the differences, much of it seems to boil down to (paraphrasing) "Oniguruma potentially performs much worse, but can do things the other libraries can't". What I _don't_ see are write-ups that quantify the degree to which this affects real world log parsing use cases, e.g., "the following log formats simply cannot be parsed without Oniguruma!" or "look how much uglier I have to make this Go regexp compared to this beautiful Oniguruma one!") Instead I just see some version of "Waah! Waah! 👶 🍼 The patterns I've copy/pasted directly from Logstash don't work in this other Grok tool!" which is valid but I find less helpful.

Perhaps the difference I see most often cited is the [first syntax for custom patterns](https://www.elastic.co/guide/en/logstash/current/plugins-filters-grok.html#_custom_patterns) shown in the Logstash Grok filter doc. Using the example they show there, if we have config `logstash-oniguruma-named-capture.conf`:

```
input {
  file {
    path => "/Users/phil/work/logstash/custom.txt"
  }
}

filter {
  grok {
    match => { "message" => "(?<queue_id>[0-9A-F]{10,11})" }
  }
}

output {
  stdout {
    codec => "json"
  }
}
```

And input `custom.txt`:

```
BEF25A72965
```

We get:

```
$ bin/logstash -f logstash-oniguruma-named-capture.conf
{
  "message": "BEF25A72965",
  "event": {
    "original": "BEF25A72965"
  },
  "@version": "1",
  "host": {
    "name": "wpm.local"
  },
  "log": {
    "file": {
      "path": "/Users/phil/work/logstash/custom.txt"
    }
  },
  "queue_id": "BEF25A72965",
  "@timestamp": "2023-10-31T21:50:24.419974Z"
}
```

So, ok, that's a handy shorthand. However, even that same Logstash doc goes on to describe how the equivalent is possible by defining a named custom Grok pattern and referencing it, and indeed, the same is possible with the current implementation on this branch.

```
$ echo '"BEF25A72965"' | zq -Z 'yield grok("POSTFIX_QUEUEID [0-9A-F]{10,11}", "%{POSTFIX_QUEUEID:queue_id}", this)' -
{
    queue_id: "BEF25A72965"
}
```

Incidentally, this same issue is cited in https://github.com/logrusorgru/grokky/issues/4 (which is from the repo for the code the implementation this branch is based on) and https://github.com/vjeantet/grok/issues/36 (which is from the repo I found when I wrote up #4140) so we're definitely not alone in this being a limitation. In fact, the implementation currently on this branch lines up with what's described in second of those quoted issues:

```
$ echo '"BEF25A72965"' | zq -Z 'yield grok("(?<queue_id>[0-9A-F]{10,11})", this)' -
error("grok(): error parsing regexp: invalid or unsupported Perl syntax: `(?<`")
```

**Conclusion:** @mattnibs and @nwt have already set my expectations that using an alternate regexp library like Oniguruma would be a non-trivial amount of work. Furthermore, having one regexp library for Grok and another for other parts of the language seems like it could come with its own set of hazards, and moving _all_ Zed regexp functionality to depend on a library that's believed to perform worse is also probably a bad choice. Since Logstash's own doc shows two approaches to achieve the same result, and we support one of those approaches, I'd be fine in sticking with what we've got here and disclosing in the docs that the other is unsupported and see if it turns out to be a problem. The error message currently being surfaced is perhaps not ideal, but it's specific enough that users can react to it (e.g., if we reference it in our Grok doc, a search for "perl" on the docs site would surely bring the user to the right place).

## References

I read through all the following as research for the write-up above. They're all pages that come up as top links for Google searches like "grok log parsing", "grok specification", and "grok create nested fields".

* https://logz.io/blog/grok-pattern-examples-for-log-parsing/
* https://newrelic.com/blog/how-to-relic/how-to-use-grok-log-parsing
* https://www.elastic.co/guide/en/logstash/current/plugins-filters-grok.html
* https://coralogix.com/blog/logstash-grok-tutorial-with-examples/
* https://graylog.org/post/getting-started-with-grok-patterns/
* https://docs.mezmo.com/2.8/telemetry-pipelines/using-grok-to-parse
* https://manpages.org/grok
* https://www.tutorialspoint.com/logstash/logstash_parsing_the_logs.htm
* https://www.youtube.com/watch?v=0vV9iVltJck
* https://stackoverflow.com/questions/71843978/java-log-parsing-with-logstash-grok
* https://logz.io/blog/logstash-grok/
* https://www.elastic.co/guide/en/elasticsearch/reference/current/grok.html
* https://streamsets.com/blog/what-are-grok-patterns/
* https://stackoverflow.com/questions/23523790/list-of-syntax-for-logstashs-grok
* https://stackoverflow.com/questions/28748674/have-a-grok-filter-create-nested-fields-as-a-result
* https://github.com/vectordotdev/vector/issues/4465
* https://www.reddit.com/r/elasticsearch/comments/n6l7j6/logstash_ecs_fields/
* https://github.com/elastic/elasticsearch/issues/15951
* https://www.elastic.co/blog/slow-and-steady-how-to-build-custom-grok-patterns-incrementally

runtime/expr/function/grok.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		return nil, err
+	}
+	p, err := h.getPattern(g.zctx, patternArg.AsString())
+	if err != nil {
+		return nil, err
+	}
+	ss := p.ParseValues(inputArg.AsString())
+	if ss == nil {
+		return nil, errors.New("no match")
```
philrz, line 58: Instead of just `error("grok(): no match")`, might we consider wrapping the original input string in the `error` value? This way if a user has set up a `zq` Grok pipeline output to `zed load` where they're expecting all the input data to be Grok-able, and some un-parseable data shows up in the input stream, they'll still have the option to find the original data inside the error and parse/load it after they've updated their Grok pattern to catch it going forward. This feels like it's in line with some of our "why Zed errors are great" stuff that's covered in places like [this blog post](https://www.brimdata.io/blog/debugging-data-first-class-errors/). It's also kind of similar to Logstash's [`_grokparsefailure` tag](https://www.elastic.co/guide/en/logstash/8.10/plugins-filters-grok.html#plugins-filters-grok-tag_on_failure) that they add on parse failures. Their overall approach is a little different than ours since by default they always include the original input string in the output as a field called `message` (see the many Logstash outputs in my recent [big comment](https://github.com/brimdata/zed/pull/4827#issuecomment-1788123901)) but it would seem to enable a similar "repair after the fact" workflow.

docs/language/functions/grok.md
```diff
@@ -0,0 +1,44 @@
+### Function
+
+&emsp; **grok** &mdash; parse a string using a grok pattern
+
+### Synopsis
+
+```
+grok(pattern: string, s: string) -> any
+grok(extra: string, pattern: string, s: string) -> any
```
nwt, line 9: It's surprising to see `extra` before rather than after the parameters common to both forms. Any reason not to move it after `s`?

philrz:
@mattnibs: While I was proud to show my work in my [exhaustive write-up above](https://github.com/brimdata/zed/pull/4827#issuecomment-1788123901) [ 😄 ] I realize the conclusions are in danger of getting lost in the text. Just to summarize the changes I advocate making to what's currently in this branch:

1. **Accept fields with dots in their names**. We don't need to automatically turn these into nested fields, though.
2. **Accept the square bracket syntax for creating nested fields**. I'm willing to lose this if it's super hard to do, though.
3. **Accept the `:int`-style syntax for type conversion, but ignore it**. Fields can all be string-typed and the user can use Zed casting. We could maybe do casting to Zed types with this syntax one day.
4. **When a string fails to parse via Grok, wrap it in the error value**. That way users can study/repair later.
5. **Move the `extra` parameter to the end**. This point from @nwt makes sense to me, as it's what I see with optional parameters in other languages.

docs/language/functions/grok.md
mattnibs: ok

mattnibs:
Thanks for the detail writeup @philrz 

> Accept fields with dots in their names. We don't need to automatically turn these into nested fields, though.
> Accept the square bracket syntax for creating nested fields. I'm willing to lose this if it's super hard to do, though.

I'll make sure all characters possible are allowed as the name, however in order to pair down the size of an already large pr I'm going to punt on handling to another pr. 

> Accept the :int-style syntax for type conversion, but ignore it.

For now I'm just going to have any type hints as part of the resultant field name.

> When a string fails to parse via Grok, wrap it in the error value.
> Move the extra parameter to the end. 

Will fix these.

nwt:
> For now I'm just going to have any type hints as part of the resultant field name.

Ignoring or honoring type hints strikes me as MVP functionality so I think we ought to do that here.

philrz:
@mattnibs: `docs/language/functions/README.md` needs an entry that links to the new `grok.md`.

docs/language/functions/README.md
```diff
@@ -28,6 +28,7 @@ Zed's [primitive types](../../formats/zed.md#1-primitive-types), e.g.,
 * [flatten](flatten.md) - transform a record into a flattened map
 * [floor](floor.md) - floor of a number
 * [grep](grep.md) - search strings inside of values
+* [grok)](grok.md) - parse a string into a structured record
```
philrz, line 31:
```suggestion
* [grok](grok.md) - parse a string into a structured record
```

docs/language/functions/grok.md
```diff
@@ -0,0 +1,44 @@
...
+
+When provided with three arguments the third argument, definitions, is a string
+of named patterns seperated by new lines in the format `PATTERN_NAME PATTERN`.
+The named patterns can then be referenced in the grok pattern argument.
+
+#### Included Patterns
+
+The _grok_ function by default includes a set of builtin named patterns
+that can be referenced in any pattern. The included named patterns can be seen
+[here](https://raw.githubusercontent.com/brimdata/zed/main/pkg/grok/grok-patterns).
```
nwt, line 28:
```suggestion
[here](https://raw.githubusercontent.com/brimdata/zed/main/pkg/grok/base.go).
```

docs/language/functions/grok.md
```diff
@@ -0,0 +1,44 @@
...
+
+The _grok_ function parses a string using a grok pattern and returns
+a record containing the parsed fields. The syntax for a grok pattern
+is `{%pattern:field_name}` where _pattern_ is a the name of the pattern
+to match text with and _field_name_ is resultant field name of the capture
+value.
+
+When provided with three arguments the third argument, definitions, is a string
+of named patterns seperated by new lines in the format `PATTERN_NAME PATTERN`.
+The named patterns can then be referenced in the grok pattern argument.
```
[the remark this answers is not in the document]
nwt: I'm in favor of @philrz's s/pattern/p/ suggestion.

Also, I think it'd be clearer to mention the pattern parameter by name in the first sentence, like this:

```
The _grok_ function parses a string `s` using a grok pattern `p` and returns
a record containing the parsed fields.
```

docs/language/functions/grok.md
```diff
@@ -0,0 +1,44 @@
...
+The _grok_ function by default includes a set of builtin named patterns
+that can be referenced in any pattern. The included named patterns can be seen
+[here](https://raw.githubusercontent.com/brimdata/zed/main/pkg/grok/base.go).
+
+### Examples
+
+Parsing a simple log line using the builtin named patterns:
+```mdtest-command
+echo '"2020-09-16T04:20:42.45+01:00 DEBUG This is a sample debug log message"' \
+  | zq -Z 'yield grok("%{TIMESTAMP_ISO8601:timestamp} %{LOGLEVEL:level} %{GREEDYDATA:message}", this)' -
```
nwt, line 35:
```suggestion
echo '"2020-09-16T04:20:42.45+01:00 DEBUG This is a sample debug log message"' |
  zq -Z 'yield grok("%{TIMESTAMP_ISO8601:timestamp} %{LOGLEVEL:level} %{GREEDYDATA:message}", this)' -
```

nwt (approved):

PR #4829, opened 2023-10-26
Basic demand inference.

jamii:


compiler/optimizer/demand.go
```diff
@@ -0,0 +1,177 @@
...
+				}
+			}
+		}
+	}
+
+	return DemandUnion([2]Demand{a, b})
+}
+
+func demandForSeq(seq dag.Seq) map[*dag.Op]Demand {
+	demands := make(map[*dag.Op]Demand)
```
jamii, line 66: There are no unique ids for ops and I didn't want to add a demand field to every op, so I'm using pointers as map keys.

compiler/optimizer/demand.go
```diff
@@ -0,0 +1,177 @@
...
+	"github.com/brimdata/zed/compiler/ast/dag"
+)
+
+type Demand interface {
+	isDemand()
+}
+
+// TODO Think about normal form for Demand.
+
+// A nil Demand means 'demand nothing'.
```
jamii, line 14: Maybe it would be safer to have an explicit `DemandNone` and have the zero value for `Demand` just be an error?

compiler/optimizer/demand.go
```diff
@@ -0,0 +1,177 @@
...
+	//    return seq
+	//})
+
+	return demands
+}
+
+func demandForSeqInto(demands map[*dag.Op]Demand, demandOnSeq Demand, seq dag.Seq) {
+	var demand = demandOnSeq
+	for i := len(seq) - 1; i >= 0; i-- {
+		op_ptr := &seq[i]
```
mccanne, line 86: Don't you want the dag.Op interface value (seq[i]) rather than the address of the slot in the slice?

compiler/optimizer/demand.go
mccanne: This seems reasonable but I think you want map[dag.Op] instead of map[*dag.Op]

compiler/optimizer/demand.go
mccanne: The Go idiom is return `(nil, err)` is the error indicator so `(nil, nil)` usually has reasonable non-error semantics.  So I think this is ok.

compiler/optimizer/demand.go
jamii: Hmmm. If I write `Seq([op, op])` then they both have the same value as `dag.Op` but different values as `*dag.Op`. But even `*dag.Op` is not safe if two different seqs share an underlying array.

jamii: I'd love to just do `map[[]int]` but alas...

jamii: Maybe we can number them in post-order or something.

jamii: I don't think there is a satisfying solution without invasive changes, so I'll just use `dag.Op` as the key and settle for checking that there are no duplicate ops.

compiler/optimizer/demand.go
```diff
@@ -0,0 +1,198 @@
+package optimizer
+
+import (
+	//"fmt"
```
nwt, line 4:
```suggestion
```

compiler/optimizer/demand.go
```diff
@@ -0,0 +1,198 @@
...
+func demandForKey(demand Demand, key string) Demand {
+	switch demand := demand.(type) {
+	case DemandAll:
+		return demand
+	case DemandKeys:
+		if value, ok := demand[key]; ok {
+			return value
+		} else {
+			return demandNone()
+		}
```
nwt, line 194:
```suggestion
		}
		return demandNone()
```

compiler/optimizer/demand.go
```diff
@@ -0,0 +1,198 @@
...
+	for _, demand := range demands {
+		if !demandIsValid(demand) {
+			panic("Invalid demand")
+		}
+	}
+	return demands
+}
+
+func demandForSeqInto(demands map[dag.Op]Demand, demandOnSeq Demand, seq dag.Seq) {
+	var demand = demandOnSeq
```
nwt, line 105: Nit: We generally reserve `var` for declaring zero-valued variables.
```suggestion
	demand := demandOnSeq
```

compiler/optimizer/demand.go
```diff
@@ -0,0 +1,198 @@
...
+		return demandUnion(
+			demandForExpr(DemandAll{}, expr.LHS),
+			demandForExpr(DemandAll{}, expr.RHS),
+		)
+	case *dag.Dot:
+		return demandKey(expr.RHS, demandForExpr(demandOnExpr, expr.LHS))
+	case *dag.Literal:
+		return demandNone()
+	case *dag.MapExpr:
+		var demand Demand = demandNone()
```
nwt, line 154:
```suggestion
		demand := demandNone()
```

compiler/optimizer/demand.go
```diff
@@ -0,0 +1,198 @@
...
+		for _, entry := range expr.Entries {
+			demand = demandUnion(demand, demandForExpr(DemandAll{}, entry.Key))
+			demand = demandUnion(demand, demandForExpr(DemandAll{}, entry.Value))
+		}
+		return demand
+	case *dag.RecordExpr:
+		var demand Demand = demandNone()
```
nwt, line 161:
```suggestion
		demand := demandNone()
```

compiler/optimizer/demand.go
```diff
@@ -0,0 +1,198 @@
...
+				if d := demandForKey(demandOnExpr, elem.Name); !demandIsEmpty(d) {
+					demand = demandUnion(demand, demandForExpr(d, elem.Value))
+				}
+			case *dag.Spread:
+				demand = demandUnion(demand, demandForExpr(demandOnExpr, elem.Expr))
+			}
+		}
+		return demand
+	case *dag.This:
+		var demand Demand = demandOnExpr
```
nwt, line 174:
```suggestion
		demand := demandNone()
```

compiler/optimizer/demand.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+		for k, v := range b {
+			if v2, ok := a[k]; ok {
+				demand[k] = demandUnion(v, v2)
+			} else {
+				demand[k] = v
+			}
+		}
+		return demand
+	}
```
nwt, line 82: Nit: Indent this less.
```suggestion
	a, b := a.(DemandKeys), b.(DemandKeys)
	demand := DemandKeys(make(map[string]Demand, len(a)+len(b)))
	for k, v := range a {
		demand[k] = v
	}
	for k, v := range b {
		if v2, ok := a[k]; ok {
			demand[k] = demandUnion(v, v2)
		} else {
			demand[k] = v
		}
	}
	return demand
```

compiler/optimizer/demand.go
```diff
@@ -0,0 +1,198 @@
...
+	demands := make(map[dag.Op]Demand)
+	demandForSeqInto(demands, DemandAll{}, seq)
+
+	//walk(seq, true, func(seq dag.Seq) dag.Seq {
+	//    for i := range seq {
+	//        fmt.Println(seq[i], " ", demands[&seq[i]])
+	//    }
+	//    return seq
+	//})
+
```
nwt, line 95: Nit: Maybe delete this?
```suggestion
```

compiler/optimizer/demand.go
jamii: This does not compile - `no new variables`.

nwt: Ah, I forgot that `a := a.(DemandKeys)` requires a new scope. 🤦

compiler/optimizer/optimizer.go
```diff
@@ -132,6 +132,7 @@ func (o *Optimizer) Optimize(seq dag.Seq) (dag.Seq, error) {
 	seq = removePassOps(seq)
 	o.optimizeParallels(seq)
 	seq = mergeFilters(seq)
+	seq = insertDemandTests(seq)
```
jamii, line 135: @nwt It would be nice to hide this behind some cli flag if it's not too fiddly. Otherwise I'll just take it out before merging.

nwt: Since this is just temporary, how about an environment variable, which requires a little less plumbing?
```suggestion
	if os.Getenv("ZED_DEMAND_TESTS") != "" {
		seq = insertDemandTests(seq)
	}
```

nwt (approved):

PR #4832, opened 2023-10-27
dynamic cut: fix bug in keys

### LINKED ISSUE #4830: cut not producing output equivalent to put+drop
Repro is with Zed commit e03665e.

Start with the successful verification example shown in https://github.com/brimdata/zed/issues/4555#issuecomment-1780316810 that uses attached test data [scan.json.gz](https://github.com/brimdata/zed/files/13183527/scan.json.gz).

```
$ zq -version
Version: v1.10.0-20-ge03665e6

$ zq -Z 'over this | ports:=collect(ports[0]) by ip | put this[ip]:=ports | drop ip,ports' scan.json.gz 
{
    "192.168.5.53": [
        {
            port: 515,
            proto: "tcp",
            status: "open",
            reason: "syn-ack",
            ttl: 64
        },
        {
            port: 21,
            proto: "tcp",
            status: "open",
            reason: "syn-ack",
            ttl: 64
        }
    ]
}
{
    "192.168.5.49": [
        {
            port: 49152,
            proto: "tcp",
            status: "open",
            reason: "syn-ack",
            ttl: 64
        },
        {
            port: 3401,
            proto: "tcp",
            status: "open",
            reason: "syn-ack",
            ttl: 64
        }
    ]
}
```

In https://github.com/brimdata/zed/pull/4795#issuecomment-1749695080, @mattnibs pointed out that the `put`/`drop` combo could be improved by replacing it with `cut` as shown below. However, this produces incorrect output: Notice that instead of the separate keys `"192.168.5.49"` and `"192.168.5.53"`, the same key `"192.168.5.53"` appears twice.

```
$ zq -Z 'over this | ports:=collect(ports[0]) by ip | cut this[ip]:=ports' scan.json.gz 
{
    "192.168.5.53": [
        {
            port: 515,
            proto: "tcp",
            status: "open",
            reason: "syn-ack",
            ttl: 64
        },
        {
            port: 21,
            proto: "tcp",
            status: "open",
            reason: "syn-ack",
            ttl: 64
        }
    ]
}
{
    "192.168.5.53": [
        {
            port: 49152,
            proto: "tcp",
            status: "open",
            reason: "syn-ack",
            ttl: 64
        },
        {
            port: 3401,
            proto: "tcp",
            status: "open",
            reason: "syn-ack",
            ttl: 64
        }
    ]
}
```

FWIW, I've noticed that if I run this in a loop sometimes the key `"192.168.5.49"` will show up in both spots instead.

mattnibs:
Fix bug in dynamic cut where two output records with different keys but equal value types would overwrite each other.

Closes #4830

### DIFF
--- a/runtime/expr/cutter.go
+++ b/runtime/expr/cutter.go
@@ -13,6 +13,7 @@ type Cutter struct {
 	fieldRefs  field.List
 	fieldExprs []Evaluator
 	lvals      []*Lval
+	outTypes   *zed.TypeVectorTable
 	typeCache  []zed.Type
 
 	builders     map[string]*recordBuilderCachedTypes
@@ -34,6 +35,7 @@ func NewCutter(zctx *zed.Context, fieldRefs []*Lval, fieldExprs []Evaluator) *Cu
 		fieldRefs:    make(field.List, n),
 		fieldExprs:   fieldExprs,
 		lvals:        fieldRefs,
+		outTypes:     zed.NewTypeVectorTable(),
 		typeCache:    make([]zed.Type, n),
 		droppers:     make(map[string]*Dropper),
 		dropperCache: make([]*Dropper, n),
@@ -79,7 +81,7 @@ func (c *Cutter) Eval(ectx Context, in *zed.Value) *zed.Value {
 	if err != nil {
 		panic(err)
 	}
-	rec := ectx.NewValue(rb.Type(types), bytes)
+	rec := ectx.NewValue(rb.Type(c.outTypes.Lookup(types), types), bytes)
 	for _, d := range droppers {
 		rec = d.Eval(ectx, rec)
 	}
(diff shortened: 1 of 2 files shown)


runtime/expr/cutter.go
```diff
@@ -9,15 +9,13 @@ import (
 )
 
 type Cutter struct {
-	zctx        *zed.Context
-	fieldRefs   field.List
-	fieldExprs  []Evaluator
-	lvals       []*Lval
-	outTypes    *zed.TypeVectorTable
```
nwt, line 16: I think this can stay here and doesn't need to be in recordBuilderCachedTypes.

nwt: (This being `outTypes`.)

mattnibs: yep

nwt (approved):
:shipit:!

philrz (approved):
Branch tests out well for me too! 👍

PR #4834, opened 2023-10-30
Vector MVP

jamii:
* Vector representation
* Reader, reusing vng.vector.Reader
  * Likely to be slow due to interface overhead, but removing the overhead would require inlining and duplicating a lot of existing code.
* Materializer, implementing zio.Reader
  * Likely to be slow due to interface overhead, but difficult to vectorize because zng is variable-length

I haven't been able to produce a valid vng file for thorough testing, but `ZED_USE_VECTOR=1 go test ./...` passes.


jamii:
* This currently reads the entire file as a single vector which is almost certainly too much. Other formats read a row group at a time, but there is no equivalent in vng. We could maybe just fix a maximum number of values to stream per vector batch, but we'll still pay for lz4 decompression to touch even the first row.
* I reused much of the existing vng.vector.Reader code. In theory we should be able to make the vector code fast enough that it's the preferred pathway and the direct vng->Value readers can go away.
* Everything is hidden behind `ZED_USE_VECTOR` for now.

jamii:
Perhaps the existing tests are not sufficient - https://github.com/brimdata/zed/pull/4834/commits/aa16b6c3b67e63e2703ac425cf132d7921fb9e61. I'm going to mark this as a draft and figure out some more thorough tests.

jamii:
Performance is terrible:

```
> hyperfine --warmup=1 'duckdb ./wrccdc_mono_10m.db -c "select service, count(*) from wrccdc_mono_10m where _path = \'conn\' and service is not null group by service order by service limit 10"'
Benchmark 1: duckdb ./wrccdc_mono_10m.db -c "select service, count(*) from wrccdc_mono_10m where _path = 'conn' and service is not null group by service order by service limit 10"
  Time (mean ± σ):     106.4 ms ±  11.7 ms    [User: 667.6 ms, System: 124.9 ms]
  Range (min … max):    79.8 ms … 121.2 ms    24 runs

> hyperfine --warmup=1  'zq -f zson \'_path=="conn" service!=null | count() by service | sort service | head 10\' wrccdc_mono_10m.vng'
Benchmark 1: zq -f zson '_path=="conn" service!=null | count() by service | sort service | head 10' wrccdc_mono_10m.vng
  Time (mean ± σ):      6.331 s ±  0.012 s    [User: 6.271 s, System: 0.106 s]
  Range (min … max):    6.314 s …  6.351 s    10 runs

> time ZED_USE_VECTOR=1 zq -f zson '_path=="conn" service!=null | count() by service | sort service | head 10' wrccdc_mono_10m.vng > /dev/null
________________________________________________________
Executed in   55.50 secs    fish           external
   usr time  111.28 secs    1.58 millis  111.27 secs
   sys time    1.63 secs    0.07 millis    1.63 secs
```

vng/vector/io.go
```diff
@@ -76,7 +76,7 @@ func NewWriter(typ zed.Type, spiller *Spiller) Writer {
 		if !zed.IsPrimitiveType(typ) {
 			panic(fmt.Sprintf("unsupported type in VNG file: %T", typ))
 		}
-		return NewPrimitiveWriter(typ, spiller, true)
```
jamii, line 79: Previously nulls in primitive values would be written as zng nulls in the primitive column rather than producing a nulls column. This was hidden by the vng reader using a zcode.Builder to reconstruct them.

jamii:
This is as well-tested as I can manage working around known bugs. 

* I fixed a bug in the vng writer that skipped null columns for primitives.
* @nwt is working on the bug with NaNs in dicts.
* The fuzzer triggers some bug that appears to be in segment handling.

With workarounds for the above, this branch survives fuzzing and can run simple benchmarks on a filtered and truncated version of the wrccdc dataset.

mccanne:
> Performance is terrible:

As expected

jamii:
I wasn't expecting it to be terrible.

vng/vector/io.go
mccanne: This was by design at the time as the Nulls column was used only for complex types.

vector/types.go
```diff
@@ -0,0 +1,123 @@
...
+import (
+	"net/netip"
+
+	"github.com/brimdata/zed"
+	"github.com/brimdata/zed/pkg/nano"
+
+	"github.com/RoaringBitmap/roaring"
+)
+
+type any interface {
```
nwt, line 12: `any` is a built-in alias for `interface{}`. Is there another name that might make sense here? Maybe `materializable`?

vector/read.go
```diff
@@ -0,0 +1,432 @@
...
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"net/netip"
+
+	"github.com/brimdata/zed"
+	"github.com/brimdata/zed/pkg/nano"
+	"github.com/brimdata/zed/vng"
+	vngVector "github.com/brimdata/zed/vng/vector"
```
nwt, line 12: We use lowercase package names.
```suggestion
	vngvector "github.com/brimdata/zed/vng/vector"
```

vector/vector.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"github.com/brimdata/zed"
+)
+
+// len(values) == len(Types)
+type Vector struct {
+	Context *zed.Context
+	Types   []zed.Type
+	values  []any
```
nwt, line 11: A comment immediately before an exported symbol is a [doc comment](https://tip.golang.org/doc/comment) and will appear in generated documentation. Sinc this one's about internals, hang it on `values` to keep it out of those docs.
```suggestion
type Vector struct {
	Context *zed.Context
	Types   []zed.Type
	// len(values) == len(types)
	values  []any
```

vng/reader.go
```diff
@@ -11,48 +11,48 @@ import (
...
+	Readers []TypedReader
 	builder zcode.Builder
 	val     zed.Value
 }
 
-type typedReader struct {
-	typ    zed.Type
-	reader vector.Reader
+type TypedReader struct {
+	Typ    zed.Type
```
nwt, line 21: We say `typ` because `type` is a Go keyword, but it's ugly.
```suggestion
	Type    zed.Type
```

vector/read.go
```diff
@@ -0,0 +1,432 @@
...
+	"io"
+	"net/netip"
+
+	"github.com/brimdata/zed"
+	"github.com/brimdata/zed/pkg/nano"
+	"github.com/brimdata/zed/vng"
+	vngvector "github.com/brimdata/zed/vng/vector"
+	"github.com/brimdata/zed/zcode"
+
+	"github.com/RoaringBitmap/roaring"
```
nwt, line 15: We just do two groups: standard library and everything else.
```suggestion
	"github.com/RoaringBitmap/roaring"
	"github.com/brimdata/zed"
	"github.com/brimdata/zed/pkg/nano"
	"github.com/brimdata/zed/vng"
	vngvector "github.com/brimdata/zed/vng/vector"
	"github.com/brimdata/zed/zcode"
```

vector/types.go
```diff
@@ -0,0 +1,123 @@
+package vector
+
+import (
+	"net/netip"
+
+	"github.com/brimdata/zed"
+	"github.com/brimdata/zed/pkg/nano"
+
+	"github.com/RoaringBitmap/roaring"
```
nwt, line 9:
```suggestion
	"github.com/RoaringBitmap/roaring"
	"github.com/brimdata/zed"
	"github.com/brimdata/zed/pkg/nano"
```

nwt (approved):

PR #4836, opened 2023-10-31
Query additional filter option

mattnibs:
Add a filter options to the query command that applies filter expressions immediately after the query source.

### DIFF
--- a/api/client/connection.go
+++ b/api/client/connection.go
@@ -297,7 +297,7 @@ func (c *Connection) Revert(ctx context.Context, poolID ksuid.KSUID, branchName 
 //
 // As for Connection.Do, if the returned error is nil, the user is expected to
 // call Response.Body.Close.
-func (c *Connection) Query(ctx context.Context, head *lakeparse.Commitish, src string, filenames ...string) (*Response, error) {
+func (c *Connection) Query(ctx context.Context, head *lakeparse.Commitish, src string, addFilters []string, filenames ...string) (*Response, error) {
 	src, srcInfo, err := parser.ConcatSource(filenames, src)
 	if err != nil {
 		return nil, err
--- a/cli/queryflags/flags.go
+++ b/cli/queryflags/flags.go
@@ -17,14 +17,19 @@ import (
 )
 
 type Flags struct {
-	Verbose  bool
-	Stats    bool
-	Includes Includes
+	Verbose    bool
+	Stats      bool
+	AddFilters []string
+	Includes   Includes
 }
 
 func (f *Flags) SetFlags(fs *flag.FlagSet) {
 	fs.BoolVar(&f.Stats, "s", false, "display search stats on stderr")
 	fs.Var(&f.Includes, "I", "source file containing Zed query text (may be used multiple times)")
+	fs.Func("filter", "add a filter expression to the query (may be used multiple times)", func(s string) error {
+		f.AddFilters = append(f.AddFilters, s)
+		return nil
+	})
 }
 
 func (f *Flags) ParseSourcesAndInputs(paths []string) ([]string, ast.Seq, bool, error) {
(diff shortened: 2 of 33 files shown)


api/client/connection.go
```diff
@@ -297,7 +297,7 @@ func (c *Connection) Revert(ctx context.Context, poolID ksuid.KSUID, branchName
 //
 // As for Connection.Do, if the returned error is nil, the user is expected to
 // call Response.Body.Close.
-func (c *Connection) Query(ctx context.Context, head *lakeparse.Commitish, src string, filenames ...string) (*Response, error) {
+func (c *Connection) Query(ctx context.Context, head *lakeparse.Commitish, src string, addFilters []string, filenames ...string) (*Response, error) {
```
nwt, line 300: Why add this?

cli/queryflags/flags.go
```diff
@@ -17,14 +17,19 @@ import (
...
+	Verbose    bool
+	Stats      bool
+	AddFilters []string
+	Includes   Includes
 }
 
 func (f *Flags) SetFlags(fs *flag.FlagSet) {
 	fs.BoolVar(&f.Stats, "s", false, "display search stats on stderr")
 	fs.Var(&f.Includes, "I", "source file containing Zed query text (may be used multiple times)")
+	fs.Func("filter", "add a filter expression to the query (may be used multiple times)", func(s string) error {
```
nwt, line 29: Why add this?

nwt (commented):
This change seems like it's a lot bigger than it needs to be. We can just add a field to api.QueryRequest and add handling for that field in service.handleQuery.

mattnibs:
> We can just add a field to api.QueryRequest and add handling for that field in service.handleQuery.

How are we correctly manipulating the ast, inserting the where clauses into the proper location if all the functionality is confined to service.handleQuery? It seems to me that if we are going to do this feature, then the semantic analyzer is that part of the code that should be responsible for rewriting the dag.

nwt:
But it'd be so much simpler like this!
```diff
diff --git a/api/api.go b/api/api.go
index 1e3bf75b3..b393964bb 100644
--- a/api/api.go
+++ b/api/api.go
@@ -113,8 +113,9 @@ type EventBranch struct {
 }
 
 type QueryRequest struct {
-	Query string              `json:"query"`
-	Head  lakeparse.Commitish `json:"head"`
+	Query       string              `json:"query"`
+	Head        lakeparse.Commitish `json:"head"`
+	HeadFilters []string            `json:"head_filters"`
 }
 
 type QueryChannelSet struct {
diff --git a/service/handlers.go b/service/handlers.go
index 6918b0de5..fe733601d 100644
--- a/service/handlers.go
+++ b/service/handlers.go
@@ -42,6 +42,14 @@ func handleQuery(c *Core, w *ResponseWriter, r *Request) {
 	if !ok {
 		return
 	}
+	var headFilters ast.Seq
+	for _, f := range req.HeadFilters {
+		f, err := c.compiler.Parse(f)
+		if err != nil {
+			w.Error(srverr.ErrInvalid(err))
+		}
+		headFilters = append(headFilters, f...)
+	}
 	// A note on error handling here.  If we get an error setting up
 	// before the query starts to run, we call w.Error() and return
 	// an HTTP status error and a JSON formatted error.  If the query
@@ -56,7 +64,10 @@ func handleQuery(c *Core, w *ResponseWriter, r *Request) {
 		w.Error(srverr.ErrInvalid(err))
 		return
 	}
-	flowgraph, err := runtime.CompileLakeQuery(r.Context(), zed.NewContext(), c.compiler, query, &req.Head, r.Logger)
+	flowgraph, err := runtime.CompileLakeQuery(r.Context(), zed.NewContext(), c.compiler, append(headFilters, query...), &req.Head, r.Logger)
+	if err != nil && err.Error() == "internal error: pool scan cannot have a parent operator" {
+		flowgraph, err = runtime.CompileLakeQuery(r.Context(), zed.NewContext(), c.compiler, query, nil, r.Logger)
+	}
 	if err != nil {
 		w.Error(srverr.ErrInvalid(err))
 		return
diff --git a/service/ztests/curl-query-filter.yaml b/service/ztests/curl-query-filter.yaml
new file mode 100644
index 000000000..42a3d1f91
--- /dev/null
+++ b/service/ztests/curl-query-filter.yaml
@@ -0,0 +1,24 @@
+script: |
+  source service.sh
+  zed create -q test
+  zed load -q -use test -
+  curl -d '{"query":"sort","head":{"pool":"test"},"head_filters":["this>1","this<3"]}' $ZED_LAKE/query
+  echo //
+  curl -d '{"query":"from test | sort","head":{"pool":"test"},"head_filters":["this>1","this<3"]}' $ZED_LAKE/query
+
+inputs:
+  - name: service.sh
+  - name: stdin
+    data: |
+      1
+      2
+      3
+
+outputs:
+  - name: stdout
+    data: |
+      2
+      //
+      1
+      2
+      3
```

PR #4847, opened 2023-11-03
Vector perf improvements

jamii:
In this PR:

* @nwt pointed out that roaring is very slow. Replaced that with `runs []int64`. When we need random access we can build a simple index over `runs`.
* Store bytes/strings in a single allocation with an array of offsets.
* Fill out all the int/uint types (saves memory).

Other things I tried:
* Replacing EncodeFoo with AppendFoo. This ends up requiring Begin/EndContainer because we don't know the lengths of the things we're appending. Minimal improvement and gross code.
* Replacing the materializer functions with an explicit switch. Slower.
* Guessing the size of vectors so the slices can be pre-allocated instead of growing. This shaves off ~15% total but the method I used to guess the sizes is not very robust. Probably better deferred to when we figure out vector chunking.

Before:

```
> time ZED_USE_VECTOR=1 zq -f zson '_path=="conn" service!=null | count() by service | sort service | head 10' ~/zed-perf-testing/autoperf/vs-duckdb/wrccdc_mono_10m.vng > /d
ev/null

________________________________________________________
Executed in   31.82 secs    fish           external
   usr time   62.94 secs    1.17 millis   62.94 secs
   sys time    1.13 secs    0.02 millis    1.13 secs
```

After:

```
> time ZED_USE_VECTOR=1 zq -f zson '_path=="conn" service!=null | count() by service | sort service | head 10' ~/zed-perf-testing/autoperf/vs-duckdb/wrccdc_mono_10m.vng

________________________________________________________
Executed in    5.89 secs    fish           external
   usr time    6.96 secs    1.11 millis    6.96 secs
   sys time    0.44 secs    0.02 millis    0.44 secs
```

Compared to existing vng reader:

```
> time zq -f zson '_path=="conn" service!=null | count() by service | sort service | head 10' ~/zed-perf-testing/autoperf/vs-duckdb/wrccdc_mono_10m.vng > /dev/null

________________________________________________________
Executed in    3.26 secs    fish           external
   usr time    3.20 secs  568.00 micros    3.20 secs
   sys time    0.08 secs   88.00 micros    0.08 secs
```

### DIFF
--- a/vector/read.go
+++ b/vector/read.go
@@ -6,7 +6,6 @@ import (
 	"io"
 	"net/netip"
 
-	"github.com/RoaringBitmap/roaring"
 	"github.com/brimdata/zed"
 	"github.com/brimdata/zed/pkg/nano"
 	"github.com/brimdata/zed/vng"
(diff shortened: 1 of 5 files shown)


vector/read.go
```diff
@@ -187,7 +175,9 @@ func readPrimitive(context *zed.Context, typ zed.Type, readBytes func() ([]byte,
 		return vector, nil
 
 	case zed.TypeBytes:
-		values := make([][]byte, 0)
+		data := bytes.NewBuffer(nil)
+		offsets := make([]int, 1)
+		offsets[0] = 0
```
nwt, line 180:
```suggestion
		offsets := []int{0}
```

vector/read.go
```diff
@@ -331,7 +378,9 @@ func readPrimitive(context *zed.Context, typ zed.Type, readBytes func() ([]byte,
 		return vector, nil
 
 	case zed.TypeString:
-		values := make([]string, 0)
+		data := bytes.NewBuffer(nil)
+		offsets := make([]int, 1)
+		offsets[0] = 0
```
nwt, line 383:
```suggestion
		offsets := []int{0}
```

nwt (approved):
:shipit:

PR #4850, opened 2023-11-03
Use demand in vector.read

jamii:
Hooking up demand to vector.Read.
* Use demand in vector.Read to avoid reading unneeded vectors. Also changes the type of the output value, removing any unused fields so that we can avoid having to materialize a bunch of nulls.
* Add vngio.ReaderOpts, where opts.Demand defaults to demand.All{}.
* Split the demand representation into a separate package to avoid import loops.
* Pull the fuzzer code from https://github.com/brimdata/zed/pull/4833 into a fuzz package. (I've manually copied across the requested changes from that pr).
* Add a fuzzer that runs random queries through both zng and vng paths and compares the results. (Found #4854). This uses a gross hack to get around the fact that readers are created before queries are analyzed.
* Change the constants vector to hold []byte instead of zed.Value, since all we ever do is copy the bytes out.

This doesn't yet actually do anything for `zq` or `zed query`. I can work on the latter, and @nwt is thinking about how to refactor to make `zq` possible.


jamii:
@brimdata/zed This is ready for review, but before merging I'll need to rebase it on https://github.com/brimdata/zed/pull/4833 to make the vng tests use the new fuzz package.

jamii:
@nwt I finished rebasing this on the other two PRs.

zio/vngio/reader.go
```diff
@@ -6,12 +6,29 @@ import (
...
 	"github.com/brimdata/zed"
+	"github.com/brimdata/zed/compiler/optimizer/demand"
 	"github.com/brimdata/zed/vector"
 	"github.com/brimdata/zed/vng"
 	"github.com/brimdata/zed/zio"
 )
 
+type ReaderOpts struct {
+	Demand demand.Demand
+}
```
nwt, line 17: In my dream world, all of the zio.Reader constructors will accept a demand.Demand, so rather than put it in here, I'm in favor of making it a NewReader parameter, like so:
```
func NewReader(zctx *zed.Context, r io.Reader, d demand.Demand) (zio.Reader, error)
```

compiler/optimizer/demand/demand.go
```diff
@@ -0,0 +1,100 @@
...
+			}
+		}
+		return true
+	default:
+		panic("Unreachable")
+	}
+}
+
+func None() Demand {
+	return keys(make(map[string]Demand, 0))
```
nwt, line 30:
```suggestion
	return keys{}
```

compiler/optimizer/demand/demand.go
```diff
@@ -0,0 +1,100 @@
...
+	return ok
+}
+
+func Key(key string, value Demand) Demand {
+	if IsNone(value) {
+		return value
+	}
+	demand := keys(make(map[string]Demand, 1))
+	demand[key] = value
+	return demand
```
nwt, line 59:
```suggestion
	return keys{key: value}
```

compiler/optimizer/demand/demand.go
```diff
@@ -0,0 +1,100 @@
...
+		return a
+	}
+	if _, ok := b.(all); ok {
+		return b
+	}
+
+	{
+		a, b := a.(keys), b.(keys)
+
+		demand := keys(make(map[string]Demand, len(a)+len(b)))
```
nwt, line 73:
```suggestion
		demand := make(keys, len(a)+len(b))
```

zio/vngio/reader.go
```diff
@@ -29,17 +46,34 @@ func NewReader(zctx *zed.Context, r io.Reader) (zio.Reader, error) {
...
+			materializer: nil,
+		}
+		return reader, nil
+	} else {
+		return vng.NewReader(o)
+	}
+}
+
+var _ zio.Reader = (*Reader)(nil)
+
```
nwt, line 68: NewReader returns a zio.Reader, making this redundant.
```suggestion
```

vector/materializer.go
```diff
@@ -219,10 +219,9 @@ func (v *arrays) newMaterializer() materializer {
 	}
 }
 
-func (v *constants) newMaterializer() materializer {
-	bytes := v.value.Bytes()
+func (vector *constants) newMaterializer() materializer {
```
nwt, line 222: For consistency.
```suggestion
func (v *constants) newMaterializer() materializer {
```

vector/read.go
```diff
@@ -39,15 +41,22 @@ func Read(reader *vng.Reader) (*Vector, error) {
 	return vector, nil
 }
 
-func read(context *zed.Context, reader vngvector.Reader) (vector, error) {
+func read(context *zed.Context, reader vngvector.Reader, demandOut demand.Demand) (vector, error) {
+	if demand.IsNone(demandOut) {
+		vector := &constants{
+			bytes: nil,
+		}
+		return vector, nil
```
nwt, line 49:
```suggestion
		return &constants{}, nil
```

vector/read.go
```diff
@@ -490,7 +501,7 @@ func readPrimitive(context *zed.Context, typ zed.Type, readBytes func() ([]byte,
 
 	case zed.TypeNull:
 		vector := &constants{
-			value: *zed.NewValue(zed.TypeNull, nil),
+			bytes: nil,
 		}
 		return vector, nil
```
nwt, line 506:
```suggestion
		return &constants{}, nil
```

vector/vector_test.go
```diff
@@ -0,0 +1,43 @@
...
+		var zngBuf bytes.Buffer
+		fuzz.WriteZng(t, values, &zngBuf)
+		resultZng := fuzz.RunQueryZng(t, &zngBuf, querySource)
+
+		var vngBuf bytes.Buffer
+		fuzz.WriteVng(t, values, &vngBuf, vngio.WriterOpts{
+			SkewThresh:   vngio.DefaultSkewThresh,
+			ColumnThresh: vngio.DefaultColumnThresh,
+		})
+		resultVng := fuzz.RunQueryVng(t, &vngBuf, querySource)
```
nwt, line 39:
```suggestion
		resultVNG := fuzz.RunQueryVng(t, &vngBuf, querySource)
```

fuzz/fuzz.go
```diff
@@ -0,0 +1,410 @@
...
+	"github.com/brimdata/zed/zcode"
+	"github.com/brimdata/zed/zio"
+	"github.com/brimdata/zed/zio/vngio"
+	"github.com/brimdata/zed/zio/zngio"
+	"github.com/brimdata/zed/zson"
+	"github.com/stretchr/testify/require"
+	"github.com/x448/float16"
+)
+
+func ReadZng(bs []byte) ([]zed.Value, error) {
```
nwt, line 34: It's conventional to use [consistent case for initialisms](https://github.com/golang/go/wiki/CodeReviewComments#initialisms).
```suggestion
func ReadZNG(bs []byte) ([]zed.Value, error) {
```

zio/vngio/reader.go
jamii: Should it be in anyio.ReaderOpts then? Or as an extra argument to anyio.NewReader too? Eventually it's going to get passed down from the dag in kernel.compileLeaf.

nwt: I like it as an extra argument to anyio.NewReader.

zio/vngio/reader.go
```diff
@@ -29,17 +46,32 @@ func NewReader(zctx *zed.Context, r io.Reader) (zio.Reader, error) {
...
-		vector, err := vector.Read(reader)
+		reader := &Reader{
+			reader:       vngReader,
+			Opts:         opts,
+			materializer: nil,
+		}
+		return reader, nil
+	} else {
+		return vng.NewReader(o)
+	}
```
nwt, line 64:
```suggestion
		return &Reader{
			reader:       vngReader,
			Opts:         opts,
			materializer: nil,
		}, nil
	}
	return vng.NewReader(o)
```

fuzz/fuzz.go
```diff
@@ -0,0 +1,410 @@
...
+	}
+	var a zbuf.Array
+	err = zio.Copy(&a, reader)
+	if err != nil {
+		return nil, err
+	}
+	return a.Values(), nil
+}
+
+func WriteZNG(t *testing.T, valuesIn []zed.Value, buf *bytes.Buffer) {
```
nwt, line 62: Nits:
1. Could use io.Writer here to make this a little more flexible.
2. We usually put the destination parameter first (as with zio.Copy).
3. We usually use val as the name for zed.Value slice.
```suggestion
func WriteZNG(t *testing.T, w io.Writer, vals []zed.Value) {
```

nwt (approved):
I'd like to have the `NewReader(zctx *zed.Context, r io.Reader, d demand.Demand)` change but then :shipit:.

zio/vngio/reader.go
jamii: https://github.com/brimdata/zed/pull/4850/commits/2c177b19eccd0c81a6994cd0b4678cb7c0f15ac1

PR #4860, opened 2023-11-04
map: Fix bug with result container type

### LINKED ISSUE #4855: Running map() on a set may return an array
Repro is with Zed commit 3a40788.

Regarding the new [`map()` function](https://zed.brimdata.io/docs/next/language/functions/map) added in #4806, if I modify the complex value in the docs example to be a set type instead of an array, `map()` still returns a set.

```
$ zq -version
Version: v1.10.0-31-g3a40788b

$ echo '|["foo","bar","baz"]|' | zq -z 'yield map(this, upper)' -
|["BAR","BAZ","FOO"]|
```

However, if I put that set value inside of a record and have `map()` act on that value, now what comes back is an array.

```
$ echo '{m:|["foo","bar","baz"]|}' | zq -z 'yield map(m, upper)' -
["BAR","BAZ","FOO"]
```

mattnibs:
Fix bug in map where the result container type was getting set of the incorrect value.

Closes #4855

### DIFF
--- a/runtime/expr/map.go
+++ b/runtime/expr/map.go
@@ -22,16 +22,16 @@ func NewMapCall(zctx *zed.Context, e, inner Evaluator) Evaluator {
 }
 
 func (a *mapCall) Eval(ectx Context, in *zed.Value) *zed.Value {
-	v := a.eval.Eval(ectx, in)
-	if v.IsError() {
-		return v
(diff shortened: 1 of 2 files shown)


nwt (approved):

philrz (approved):
Branch tests out well for me to! 👍

PR #4872, opened 2023-11-08
Allow using vectors in lake

jamii:
When ZED_USE_VECTOR is set, assume that all vng versions of zng files are present, and transform PoolScan into VecLister+VecScan.

This also finally connects the demand inference to the vector load, but it turns out that the cost of loading data is tiny compared to the overhead of decoding/encoding, so even on queries where demand inference can remove almost all of the data loading we're still slower than just running the same query against zng.

### DIFF
--- a/compiler/optimizer/demand.go
+++ b/compiler/optimizer/demand.go
@@ -81,7 +81,7 @@ func inferDemandExprIn(demandOut demand.Demand, expr dag.Expr) demand.Demand {
 	switch expr := expr.(type) {
 	case *dag.Agg:
 		// Since we don't know how the expr.Name will transform the inputs, we have to assume demand.All.
-		demandIn = demand.Union(
+		return demand.Union(
 			inferDemandExprIn(demand.All(), expr.Expr),
 			inferDemandExprIn(demand.All(), expr.Where),
 		)
--- a/compiler/optimizer/demand.go
+++ b/compiler/optimizer/demand.go
@@ -9,8 +9,7 @@ func insertDemand(seq dag.Seq) dag.Seq {
 	demands := InferDemandSeqOut(seq)
 	return walk(seq, true, func(seq dag.Seq) dag.Seq {
 		for _, op := range seq {
-			switch vecOp := op.(type) {
-			case *dag.VecSeqScan:
+			if vecOp, ok := op.(*dag.VecSeqScan); ok {
 				vecOp.Demand = demands[op]
 			}
 		}
--- a/compiler/optimizer/demand.go
+++ b/compiler/optimizer/demand.go
@@ -5,6 +5,19 @@ import (
 	"github.com/brimdata/zed/compiler/optimizer/demand"
 )
 
+func insertDemand(seq dag.Seq) dag.Seq {
+	demands := InferDemandSeqOut(seq)
+	return walk(seq, true, func(seq dag.Seq) dag.Seq {
+		for _, op := range seq {
+			switch vecOp := op.(type) {
+			case *dag.VecSeqScan:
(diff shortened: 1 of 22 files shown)


lake/commits/snapshot.go
```diff
@@ -3,6 +3,7 @@ package commits
 import (
 	"errors"
 	"fmt"
+	"golang.org/x/exp/maps"
```
nwt, line 6: We're on Go 1.21 now.
```suggestion
	"maps"
```

jamii: Huh, the maps packages doesn't have the Keys function :S 

I'll just rewrite that section.

vng/vector/primitive.go
```diff
@@ -213,9 +212,10 @@ func (p *PrimitiveReader) ReadBytes() (zcode.Bytes, error) {
 func (p *PrimitiveReader) next() error {
 	segment := p.segmap[0]
 	p.segmap = p.segmap[1:]
-	if segment.Length > 2*MaxSegmentThresh {
-		return errors.New("corrupt VNG: segment too big")
-	}
+	//TODO Segments are currently disabled.
+	//if segment.Length > 2*MaxSegmentThresh {
+	//    return errors.New("corrupt VNG: segment too big")
+	//}
```
nwt, line 218: I don't see anything failing without the changes in this file. Unless I'm missing something, let's leave it as is.

compiler/optimizer/demand.go
```diff
@@ -29,73 +42,87 @@ func inferDemandSeqOutWith(demands map[dag.Op]demand.Demand, demandSeqOut demand
...
+		return demand.None()
+	}
+	if expr == nil {
+		return demand.None()
 	}
+	var demandIn demand.Demand
 	switch expr := expr.(type) {
+	case *dag.Agg:
+		// Since we don't know how the expr.Name will transform the inputs, we have to assume demand.All.
+		demandIn = demand.Union(
```
nwt, line 85:
```suggestion
		return demand.Union(
```

vng/writer.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 const (
-	MaxSegmentThresh = vector.MaxSegmentThresh
+	MaxSegmentThresh = math.MaxInt
```
nwt, line 15: Should we make this change to vector.MaxSegmentThresh intead?

vector/vector_test.go
```diff
@@ -41,3 +49,88 @@ func FuzzQuery(f *testing.F) {
 		fuzz.CompareValues(t, resultZNG, resultVNG)
 	})
 }
+
+const N = 10000000
+
+func BenchmarkReadZng(b *testing.B) {
```
nwt, line 55: This doesn't really have anything to do with using vectors in the lake so if you want to add it let's do so in a separate PR.

vng/vector/primitive.go
jamii: On main:

```
> zq -f vng -o ints.vng 'yield this, this' ints.zng                                                      

> zq 'count()' ints.vng                                                                                  
ints.vng: corrupt VNG: segment too big
```

You changed the write path to make one big segment, but the read path still panics if the segment is over the max length.

vng/writer.go
jamii: Good point.

nwt (commented):
I was thinking we'd do something more along the lines of https://github.com/brimdata/zed/tree/noah-lake-vector, which is a little simpler and also more complete in that it falls back to ZNG if VNG isn't available.

nwt (approved):
Let's merge this and worry about simplification in another PR.

PR #4881, opened 2023-11-14
Low-hanging fruit in vector.Read

jamii:
* Read directly from segmap rather than using `ReadBytes`. 
* Include a count in each segment (this is a breaking change to vng). Use this to pre-size slices.
* Reuse a single buffer across all internal read calls.
* Avoid unnecessary copies for string/bytes vectors.
* Correctly handle dict vectors, now that they're enabled.
* Add a benchmark tracking vng read performance and comparing it to zng and raw varints.

On wrccdc_mono_10m.vng this cuts vector.Read time from 270ms -> 140ms for id.resp_p and 470ms -> 180ms for uid.

Materializer performance is still terrible. I had some success improving it by only exposing a Puller instead of a zio.Reader, but the changes ended up being quite invasive elsewhere so I backed it out. It's probably worth trying again later once we only have one set of code for reading vng.

### DIFF
--- a/vector/read.go
+++ b/vector/read.go
@@ -15,26 +15,25 @@ import (
 	"github.com/brimdata/zed/zcode"
 )
 
-func Read(reader *vng.Reader, demandOut demand.Demand) (*Vector, error) {
-	context := zed.NewContext()
-	tags, err := ReadInt64s(reader.Root)
+func Read(object *vng.Object, demandOut demand.Demand) (*Vector, error) {
+	tags, err := ReadInt64s(object.ReaderAt, object.Root)
 	if err != nil {
 		return nil, err
 	}
-	types := make([]zed.Type, len(reader.Readers))
-	values := make([]vector, len(reader.Readers))
-	for i, typedReader := range reader.Readers {
-		typeCopy, _ := context.DecodeTypeValue(zed.EncodeTypeValue(typedReader.Type))
-		typ := typeAfterDemand(context, typedReader.Reader, demandOut, typeCopy)
+	types := make([]zed.Type, len(object.Maps))
+	values := make([]vector, len(object.Maps))
+	for i, metadata := range object.Maps {
+		typeCopy := metadata.Type(object.Zctx)
+		typ := typeAfterDemand(object.Zctx, metadata, demandOut, typeCopy)
 		types[i] = typ
-		value, err := read(context, typedReader.Reader, demandOut)
+		value, err := read(object.Zctx, object.ReaderAt, metadata, demandOut)
 		if err != nil {
 			return nil, err
 		}
 		values[i] = value
 	}
 	vector := &Vector{
-		Context: context,
+		Context: object.Zctx,
 		Types:   types,
 		values:  values,
 		tags:    tags,
@@ -42,19 +41,19 @@ func Read(reader *vng.Reader, demandOut demand.Demand) (*Vector, error) {
 	return vector, nil
 }
 
-func read(context *zed.Context, reader vngvector.Reader, demandOut demand.Demand) (vector, error) {
+func read(zctx *zed.Context, readerAt io.ReaderAt, meta vngvector.Metadata, demandOut demand.Demand) (vector, error) {
 	if demand.IsNone(demandOut) {
 		return &constants{}, nil
 	}
 
-	switch reader := reader.(type) {
+	switch meta := meta.(type) {
 
-	case *vngvector.ArrayReader:
-		lengths, err := ReadInt64s(reader.Lengths)
+	case *vngvector.Array:
+		lengths, err := ReadInt64s(readerAt, meta.Lengths)
 		if err != nil {
 			return nil, err
 		}
-		elems, err := read(context, reader.Elems, demand.All())
+		elems, err := read(zctx, readerAt, meta.Values, demand.All())
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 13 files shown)


vector/read.go
```diff
@@ -14,570 +16,597 @@ import (
...
-		return vector, nil
+		}, nil
 
 	default:
-		return nil, fmt.Errorf("unknown VNG vector Reader type: %T", reader)
+		return nil, fmt.Errorf("unknown VNG meta type: %T", meta)
+	}
+}
+
+func ReadInt64s(readerAt io.ReaderAt, buf *[]byte, segmap []vngvector.Segment) ([]int64, error) {
```
nwt, line 177: No need to export this (unless you plan to in the future).

vector/read.go
```diff
@@ -14,570 +16,597 @@ import (
...
-		typeCopy, _ := context.DecodeTypeValue(zed.EncodeTypeValue(typedReader.Type))
-		typ := typeAfterDemand(context, typedReader.Reader, demandOut, typeCopy)
+	types := make([]zed.Type, len(object.Maps))
+	values := make([]vector, len(object.Maps))
+	for i, metadata := range object.Maps {
+		typeCopy := metadata.Type(object.Zctx)
+		typ := typeAfterDemand(object.Zctx, metadata, demandOut, typeCopy)
 		types[i] = typ
-		value, err := read(context, typedReader.Reader, demandOut)
+		value, err := read(object.Zctx, object.ReaderAt, &buf, metadata, demandOut)
```
nwt, line 31: Nit: It looks as though the internal read functions (this one, ReadInt64s, readPrimitive, and readSegment at least) mostly share the same zed.Context, io.ReaderAt, and []byte parameters, so it might make things a little cleaner to stash them in a struct like
```go
type reader struct {
	zctx *zed.Context
	ra io.ReaderAt
	buf []byte
}
```
so you can hang the read functions on it as methods and then say
```go
		value, err := (&reader{object.Zctx, object.ReaderAt, nil}).read(metadata, demandOut)
```
here.

vector/read.go
```diff
@@ -14,570 +16,597 @@ import (
...
+			if err != nil {
+				return nil, err
+			}
+			return &dict{
+				dict: meta.Dict,
+				tags: tags,
+			}, nil
+		} else {
+			return readPrimitive(zctx, readerAt, buf, meta.Segmap, meta.Type(zctx))
+		}
```
nwt, line 122: Nits: We usually minimize indentation when doing so doesn't otherwise make things less readable. And if an `if` branch block returns, we usually omit `else`.
```suggestion
		if len(meta.Dict) == 0 {
			return readPrimitive(zctx, readerAt, buf, meta.Segmap, meta.Type(zctx))
		}
		tags, err := readSegmap(readerAt, meta.Segmap)
		if err != nil {
			return nil, err
		}
		return &dict{
			dict: meta.Dict,
			tags: tags,
		}, nil
```

vector/read.go
```diff
@@ -14,570 +16,597 @@ import (
...
 			if err != nil {
-				if err == io.EOF {
-					break
-				} else {
-					return nil, err
-				}
+				return nil, err
+			}
+			it := zcode.Iter(*buf)
+			for !it.Done() {
```
nwt, line 202: Nit:
```suggestion		
			for it := zcode.Iter(*buf); !it.Done(); {
```

vector/read.go
```diff
@@ -14,570 +16,597 @@ import (
...
+
 	switch typ {
 	case zed.TypeBool:
-		var values []bool
-		for {
-			bs, err := readBytes()
+		values := make([]bool, 0, count)
+		for _, segment := range segmap {
+			err := readSegment(readerAt, buf, segment)
 			if err != nil {
```
nwt, line 198: Nit:
```suggestion
			if err := readSegment(readerAt, buf, segment); err != nil {
```

vector/read.go
```diff
@@ -14,570 +16,597 @@ import (
...
-				} else {
-					return nil, err
-				}
+				return nil, err
+			}
+			it := zcode.Iter(*buf)
+			for !it.Done() {
+				bs := it.Next()
+				value := zed.DecodeBool(bs)
+				values = append(values, value)
```
nwt, line 205: Nit: Maybe shorten?
```suggestion
				values = append(values, zed.DecodeBool(it.Next()))
```

vector/read.go
```diff
@@ -14,570 +16,597 @@ import (
...
+
+func readSegmap(readerAt io.ReaderAt, segmap []vngvector.Segment) ([]byte, error) {
+	var memLength int
+	for _, segment := range segmap {
+		memLength += int(segment.MemLength)
+	}
+	data := make([]byte, memLength)
+	offset := 0
+	for _, segment := range segmap {
+		if err := segment.Read(readerAt, data[offset:offset+int(segment.MemLength)]); err != nil {
```
nwt, line 538: Nit: No need to specify the high bound since Segment.Read takes care of it.
```suggestion
		if err := segment.Read(readerAt, data[offset:]); err != nil {
```

vector/read.go
```diff
@@ -14,570 +16,597 @@ import (
...
-		ints = append(ints, int)
+		offset += int(segment.MemLength)
 	}
-	return ints, nil
+	return data, nil
+}
+
+func stripContainers(data []byte, countHint int) ([]byte, []int, error) {
+	offsetFrom := 0
+	offsetTo := 0
```
nwt, line 548: Nit: Use var for zero values.
```suggestion
	var offsetFrom, offsetTo int
```

vector/vector_test.go
```diff
@@ -41,3 +46,86 @@ func FuzzQuery(f *testing.F) {
...
 	})
 }
+
+const N = 10000000
+
+func BenchmarkReadZng(b *testing.B) {
+	rand := rand.New(rand.NewSource(42))
+	valuesIn := make([]zed.Value, N)
+	for i := range valuesIn {
+		valuesIn[i] = *zed.NewValue(zed.TypeInt64, zed.EncodeInt(int64(rand.Intn(N))))
```
nwt, line 56:
```suggestion
		valuesIn[i] = *zed.NewInt64(rand.Int63n(N))
```

nwt (approved):

nwt (approved):

PR #4887, opened 2023-11-16
grep: accept variables as pattern arguments

### LINKED ISSUE #4022: grep should accept a variable as a pattern
A community user reported the following via private gist. Repro is with Zed commit [d10050f](https://github.com/brimdata/zed/commit/d10050fee0f75534f95b91d555431f553179c4f8).

Start with this simple `grep` example.

```
$ zq -version
Version: v1.2.0-8-gd10050fe

$ echo '{a:"hello"} {a:"world"}' | zq -z 'yield {value:a,matches:grep("hello",a)}' -
{value:"hello",matches:true}
{value:"world",matches:false}
```

We can do something similar to search array elements using `over`.

```
$ echo '{a:["hello","world"]}' | zq -z 'over a => (yield {value:this,matches:grep("hello",this)})' -
{value:"hello",matches:true}
{value:"world",matches:false}
```

However, if we wanted to do something like have the target search pattern be a variable in a lateral query, this is not yet possible.

```
$ echo '{a:["hello","world"],target:"hello"}' | zq -z 'over a with searchfor=target => (yield {value:this,matches:grep(searchfor,this)})' -
grep(): syntax error
```

### LINKED ISSUE #4692: grep syntax error when string is passed into user op
Repro is with Zed commit ed3b086.

These both work as expected:

```
$ zq -version
Version: v1.8.1-55-ged3b0865

$ echo '"hello" "world"' | zq -Z 'where grep("ell")' -
"hello"

$ echo '"hello" "world"' | zq -Z 'op CallMe(): ( where grep("ell") ) CallMe()' -
"hello"
```

However, if I try to turn the string into a parameter and pass it in, now I get an error.

```
$ echo '"hello" "world"' | zq -Z 'op CallMe(MyStr): ( where grep(MyStr) ) CallMe("ell")' -
compiling filter: grep(): syntax error
```

This is actually a simplified repro of my more ambitious attempt to turn what's in https://github.com/brimdata/zed/issues/4050#issuecomment-1615227518 into a user op.

```
$ cat change_fields.zed 
op change_fields(pattern, change_to): (
  over flatten(this) => (
    value:=grep(pattern)
      ? change_to
      : value
    | collect(this)
    | unflatten(this)
  )
)

$ zq -Z -I change_fields.zed 'change_fields(f*, "changed")' records.json 
zq: error parsing Zed at column 17:
change_fields(f*, "changed")
            === ^ ===
```

By a strict reading of the [docs](https://zed.brimdata.io/docs/next/language/statements#arguments) I can see that this was probably never intended to work since the `f*` is neither a literal nor a field reference. It'd be great if it _could_ work though. 😉  However, even I change that to a string literal it fails.

```
$ zq -Z -I change_fields.zed 'change_fields("f", "changed")' records.json 
rhs of assigment expression: grep(): syntax error
```

mattnibs:
This pr adjusts grep functionality so that it can accept variables for the pattern argument. An additional improvement is that any compile time evaluatable string will get converted into an optimized search.

Closes #4692, closes #4022

### DIFF
--- a/compiler/parser/parser.peg
+++ b/compiler/parser/parser.peg
@@ -770,19 +770,12 @@ FunctionArgs
   / OptionalExprs
 
 Grep
-  = "grep" __ "(" __ pattern:Pattern __ opt:("," __ (OverExpr / Expr) __)? ")" {
+  = "grep" __ "(" __ pattern:(Regexp / Glob / Expr) __ opt:("," __ (OverExpr / Expr) __)? ")" {
       VAR(m) = MAP("kind": "Grep", "pattern": pattern, "expr": MAP("kind": "ID", "name": "this"))
       if ISNOTNULL(opt) {
         m["expr"] = ASSERT_ARRAY(opt)[2]
       }
       RETURN(m)
-    }
-
-Pattern
-  = Regexp
-  / Glob
-  / s:QuotedString {
-      RETURN(MAP("kind": "String", "text": s))
     }
 
 OptionalExprs
--- a/docs/language/functions/grep.md
+++ b/docs/language/functions/grep.md
@@ -12,9 +12,9 @@ grep(<pattern> [, e: any]) -> bool
 
 The _grep_ function searches all of the strings in its input value `e`
 (or `this` if `e` is not given)
- using the `<pattern>` argument, which must be a
+ using the `<pattern>` argument, which can be a
 [regular expression](../search-expressions.md#regular-expressions),
-[glob pattern](../search-expressions.md#globs), or string literal.
+[glob pattern](../search-expressions.md#globs), or string.
 If the pattern matches for any string, then the result is `true`.  Otherwise, it is `false`.
 
 > Note that string matches are case insensitive while regular expression
--- a/runtime/expr/function/grep.go
+++ b/runtime/expr/function/grep.go
@@ -1,0 +1,41 @@
+package function
+
+import (
+	"github.com/brimdata/zed"
+	"github.com/brimdata/zed/runtime/expr"
+	"golang.org/x/text/unicode/norm"
+)
+
+type Grep struct {
+	grep    expr.Evaluator
+	pattern string
+	zctx    *zed.Context
+}
+
+func (g *Grep) Call(ectx zed.Allocator, vals []zed.Value) *zed.Value {
+	patternVal, inputVal := &vals[0], &vals[1]
+	if zed.TypeUnder(patternVal.Type) != zed.TypeString {
+		return g.error(ectx, "pattern argument must be a string", patternVal)
+	}
+	if p := patternVal.AsString(); g.grep == nil || g.pattern != p {
+		g.pattern = p
+		term := norm.NFC.Bytes(patternVal.Bytes())
+		g.grep = expr.NewSearchString(string(term), nil)
+	}
+	return g.grep.Eval(wrapAllocator{ectx}, inputVal)
+}
+
+// XXX This is gross, any reason function shouldn't just accept an expr.Context?
+type wrapAllocator struct {
+	zed.Allocator
+}
+
+func (wrapAllocator) Vars() []zed.Value { return nil }
+
+func (g *Grep) error(ectx zed.Allocator, msg string, val *zed.Value) *zed.Value {
+	msg = "grep(): " + msg
+	if val == nil {
+		return ectx.CopyValue(*g.zctx.NewErrorf(msg))
+	}
+	return ectx.CopyValue(*g.zctx.WrapError(msg, val))
+}
--- a/runtime/expr/function/ztests/grep.yaml
+++ b/runtime/expr/function/ztests/grep.yaml
@@ -1,0 +1,17 @@
+# This test tests grep as a function call which only happens when the pattern
+# arg is not a glob, regular expression, or resolvable to a string at compile
+# time.
+
+script: |
+  echo '{pattern:"hello",input:[{a:{hello:"world"}},{hi:"world"}]}' | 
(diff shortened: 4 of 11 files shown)


philrz:
I tested this out and it definitely seems to address the examples from the two linked issues so it's a big improvement and a functional 👍 for me.

Once these changes make it possible to pass a string literal into a User Op and then use it in `grep()`, I suspect users might start hoping they could achieve the same with a `/regexp/` or glob. I just discussed this with @mattnibs offline and he helped me understand why this is easier said than done, so I'm content to wait to see if it ever becomes a concern.

compiler/parser/parser.peg
```diff
@@ -770,7 +770,7 @@ FunctionArgs
   / OptionalExprs
 
 Grep
-  = "grep" __ "(" __ pattern:Pattern __ opt:("," __ (OverExpr / Expr) __)? ")" {
+  = "grep" __ "(" __ pattern:(Regexp / Glob / Expr) __ opt:("," __ (OverExpr / Expr) __)? ")" {
```
nwt, line 773: `Pattern` is now unused so remove it.

docs/language/functions/grep.md
```diff
@@ -12,9 +12,9 @@ grep(<pattern> [, e: any]) -> bool
 
 The _grep_ function searches all of the strings in its input value `e`
 (or `this` if `e` is not given)
- using the `<pattern>` argument, which must be a
+ using the `<pattern>` argument, which can be a
 [regular expression](../search-expressions.md#regular-expressions),
-[glob pattern](../search-expressions.md#globs), or string literal.
+[glob pattern](../search-expressions.md#globs), or a string.
```
nwt, line 17:
```suggestion
[glob pattern](../search-expressions.md#globs), or string.
```

runtime/expr/function/grep.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+type Grep struct {
+	grep   expr.Evaluator
+	grestr string
+	zctx   *zed.Context
+}
+
+func (g *Grep) Call(ectx zed.Allocator, vals []zed.Value) *zed.Value {
+	patternArg, inputArg := vals[0], vals[1]
```
nwt, line 18: Nits: We usually give these things a "Val" suffix. And no need to copy them.
```suggestion
	patternVal, inputVal := &vals[0], &vals[1]
```

runtime/expr/function/grep.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+// XXX This is gross, any reason function shouldn't just accept an expr.Context?
+type wrapAllocator struct {
+	zed.Allocator
+}
+
+func (wrapAllocator) Vars() []zed.Value { return nil }
+
+func (g *Grep) error(ectx zed.Allocator, err string, val *zed.Value) *zed.Value {
```
nwt, line 37: Nit: `err` is for `error`s.
```suggestion
func (g *Grep) error(ectx zed.Allocator, msg string, val *zed.Value) *zed.Value {
```

runtime/expr/function/grep.go
```diff
@@ -0,0 +1,43 @@
...
+	err = fmt.Sprintf("grep(): %s", err)
```
nwt, line 38: Nit: Just use `+`.
```suggestion
	msg = "grep(): " + msg
```

runtime/expr/function/grep.go
```diff
@@ -0,0 +1,43 @@
...
+	"fmt"
+
+	"github.com/brimdata/zed"
+	"github.com/brimdata/zed/runtime/expr"
+	"golang.org/x/text/unicode/norm"
+)
+
+type Grep struct {
+	grep   expr.Evaluator
+	grestr string
```
nwt, line 13: Nit: I think `pattern` would be a better name.
```suggestion
	pattern string
```

runtime/expr/function/ztests/grep.yaml
```diff
@@ -0,0 +1,16 @@
+# This test tests grep as a function call which only happens when the pattern
+# arg is not a glob, regular expression, or compile time resolvable to a string.
+
+script: |
+  echo '{pattern:"hello",input:[{a:{hello:"world"}},{hi:"world"}]}' \
+    | zq -z 'over input with p=pattern => ( grep(p) )' -
+  echo "// ==="
+  echo '{a:{foo:"bar"}} {b:{baz:"foo"}}' \
+    | zq -z 'func g(s, e): ( grep(s, e) ) where g("baz", this)' -
```
nwt, line 9:
```suggestion
  echo '{pattern:"hello",input:[{a:{hello:"world"}},{hi:"world"}]}' |
    zq -z 'over input with p=pattern => ( grep(p) )' -
  echo //
  echo '{a:{foo:"bar"}} {b:{baz:"foo"}}' |
    zq -z 'func g(s, e): ( grep(s, e) ) where g("baz", this)' -
```

runtime/expr/function/ztests/grep.yaml
```diff
@@ -0,0 +1,16 @@
+# This test tests grep as a function call which only happens when the pattern
+# arg is not a glob, regular expression, or compile time resolvable to a string.
```
nwt, line 2:
```suggestion
# arg is not a glob, regular expression, or resolvable to a string at compile time.
```

nwt (approved):

runtime/expr/function/ztests/grep.yaml
mattnibs: We typically use `// ===` to mark different tests, why change this now?

nwt: `//` feels sufficient but I'm fine with `// ===`.

PR #4889, opened 2023-11-17
CSV Writer: Allow different types in output

### LINKED ISSUE #4781: Allow CSV output when columns are consistent even if types aren't
Repro is with Zed commit 80dbcda.

Consider this community user's CSV output example that motivated the changes in #4773 

```
$ zq -version
Version: v1.9.0-24-g80dbcda5

$ echo '{a:1,b:null}{a:1,b:2}' | zq -f csv -
a,b
1,
```

Or here's another for a non-`null` case:

```
$ echo '{"a":1} {"a":"hi"}' | zq -f csv -
a
1
```

As the error messages indicate, adding `fuse` does work around the problem in both cases by taking a first pass through the data to coerce all the input values to construct a single merged record type. However, since CSV effectively lacks real data typing, what's truly important in the record type is that the _fields_ are consistent, since once the field names in the header row are output there's no way to deal with additional fields later encountered. But using the examples above, a `null` value is going to be output the same in CSV (i.e., "nothing" between the comma delimiters) regardless of whether it had a particular type in Zed. Similarly, if we run `fuse` on the second example, indeed the `union` type is established to indicate the field could hold an integer or string:

```
$ echo '{"a":1} {"a":"hi"}' | zq -z 'fuse | count() by typeof(this)' -
{typeof:<{a:(int64,string)}>,count:2(uint64)}
```

but when those values are being output on the second pass, they're still printed in the same column of the CSV output as a number or string.

```
$ echo '{"a":1} {"a":"hi"}' | zq -f csv 'fuse' -
a
1
hi
```

i.e., the brief existence of the `union` type was just to satisfy a current constraint of the CSV writer, but didn't really enhance the output in any way.

In a discussion with @nwt he agreed that we could probably relax this constraint of the CSV writer. This would reduce the number of times that new users encounter the `(consider 'fuse')` message and have to take a detour to learn about why it's needed and how to use it.

mattnibs:
Adjust the csvio.Writer so that it can handle records with the same field names but different types.

Closes #4781

### DIFF
--- a/zio/csvio/writer.go
+++ b/zio/csvio/writer.go
@@ -66,7 +66,7 @@ func (w *Writer) Write(rec *zed.Value) error {
 			return err
 		}
 	} else if _, ok := w.types[rec.Type.ID()]; !ok {
-		if !fieldsEqual(w.first.Fields, rec.Fields()) {
+		if !fieldNamesEqual(w.first.Fields, rec.Fields()) {
 			return ErrNotDataFrame
 		}
 		w.types[rec.Type.ID()] = struct{}{}
@@ -103,7 +103,7 @@ func formatValue(typ zed.Type, bytes zcode.Bytes) string {
 	return zson.FormatValue(zed.NewValue(typ, bytes))
 }
 
-func fieldsEqual(a, b []zed.Field) bool {
+func fieldNamesEqual(a, b []zed.Field) bool {
 	return slices.EqualFunc(a, b, func(a, b zed.Field) bool {
 		return a.Name == b.Name
 	})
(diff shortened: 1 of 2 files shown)


philrz:
I just tested out the branch at commit dc95dd8 and it seems to work great for the examples from #4781. That makes it a functional 👍 for me!

zio/csvio/writer.go
```diff
@@ -96,3 +102,9 @@ func formatValue(typ zed.Type, bytes zcode.Bytes) string {
 	}
 	return zson.FormatValue(zed.NewValue(typ, bytes))
 }
+
+func fieldsEqual(a, b []zed.Field) bool {
```
nwt, line 106:
```suggestion
func fieldNamesEqual(a, b []zed.Field) bool {
```

nwt (approved):

PR #4903, opened 2023-11-28
Disallow creation of multi-key pools

### LINKED ISSUE #4596: Prevent creation of multi-key pools (or inform users it's unsupported)
At the time this issue is being created, Zed it at commit a2626a7.

At the moment, `zed` does allow the creation of pools with multiple keys;

```
$ zed -version
Version: v1.7.0-66-ga2626a7c

$ zed create -orderby foo,bar MyPool
pool created: MyPool 2Pykr133SNb0NUI0YKcGZTtzbqy

$ zed ls
MyPool 2Pykr133SNb0NUI0YKcGZTtzbqy key foo,bar order asc
```

However, the code for handling multiple keys is known to be incomplete and it's likely users will hit bugs if they try to work with multi-key pools right now. While we don't expect the work involved to be huge to start supporting multi-key pools (#2657), since we're not getting to it right this minute, for now we should probably prevent users from being able to create such pools, or at least inform them that they're running something unsupported and that bugs are likely.

mattnibs:
Closes #4596

### DIFF
--- a/lake/root.go
+++ b/lake/root.go
@@ -333,6 +333,9 @@ func (r *Root) CreatePool(ctx context.Context, name string, sortKey order.SortKe
 	if thresh == 0 {
 		thresh = data.DefaultThreshold
 	}
+	if len(sortKey.Keys) > 1 {
+		return nil, errors.New("multiple pool keys not supported")
+	}
 	config := pools.NewConfig(name, sortKey, thresh, seekStride)
 	if err := CreatePool(ctx, r.engine, r.logger, r.path, config); err != nil {
 		return nil, err
--- a/cmd/zed/create/command.go
+++ b/cmd/zed/create/command.go
@@ -15,13 +15,13 @@ import (
 
 var Cmd = &charm.Spec{
 	Name:  "create",
-	Usage: "create [-orderby key[,key...][:asc|:desc]] name",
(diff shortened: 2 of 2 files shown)


lake/root.go
```diff
@@ -333,6 +333,9 @@ func (r *Root) CreatePool(ctx context.Context, name string, sortKey order.SortKe
 	if thresh == 0 {
 		thresh = data.DefaultThreshold
 	}
+	if len(sortKey.Keys) > 1 {
+		return nil, errors.New("multiple pool keys not supported")
+	}
```
nwt, line 338: This is all we need for #4596. I think we should leave everything else as is.

nwt: My reasoning here is that we'll have to revert all the other changes when we add support for multiple pool keys.

mattnibs: I can go along with this (in fact my first revision of this pr was similar) with the caveat that we should keep the help document changes and also if we do it this way the server will have to be adjusted so this is 400 error.

mattnibs: I also think once we decide to add support for multiple pool keys we will have to augment order.SortKeys so that each key can have a different sort order, so we'll have to make changes on everything this pr touches anyway.

nwt (approved):

PR #4904, opened 2023-11-28
api: Ensure pool key default is set

mattnibs:
This commit ensures that the default pool key is set to ts:desc which was previously being set as null when not specified in an API request.

### DIFF
--- a/lake/pools/config.go
+++ b/lake/pools/config.go
@@ -4,6 +4,7 @@ import (
 	"github.com/brimdata/zed/lake/data"
 	"github.com/brimdata/zed/lake/journal"
 	"github.com/brimdata/zed/order"
+	"github.com/brimdata/zed/pkg/field"
 	"github.com/brimdata/zed/pkg/nano"
 	"github.com/brimdata/zed/pkg/storage"
 	"github.com/segmentio/ksuid"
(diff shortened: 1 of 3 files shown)


lake/pools/config.go
```diff
@@ -27,6 +28,9 @@ func NewConfig(name string, sortKey order.SortKey, thresh int64, seekStride int)
 	if seekStride == 0 {
 		seekStride = data.DefaultSeekStride
 	}
+	if sortKey.IsNil() {
+		sortKey = order.NewSortKey(order.Desc, field.DottedList("ts"))
+	}
```
nwt, line 33: Nit: Move this ahead of `if thresh == 0 {...}` so these remain in parameter order.

service/ztests/curl-default-poolkey.yaml
```diff
@@ -0,0 +1,30 @@
+script: |
+  source service.sh
+  curl -X POST -d '{"name":"test"}' $ZED_LAKE/pool > pool.json
+  curl -X POST \
+    -H "Accept: application/json" \
+    -d '{"query":"from :pools"}' \
+    $ZED_LAKE/query | zq -Z "yield this[0] | id:=0,ts:=0" -
```
nwt, line 7: Instead of adding this test, just remove `defaultSortKey` from service/handlers_test.go so the tests in there get the actual default.

nwt (approved):

PR #4912, opened 2023-11-29
csv: Disable TrimLeadingSpace for space delimiters

### LINKED ISSUE #4911: TSV reader fails on multiple null values
## tl;dr

With this example we fail to read our own TSV output back as TSV input.

```
$ echo '{foo: "bar", b: null, c: null}' | zq -f tsv - | zq -i tsv -Z -
stdio:stdin: record on line 2: wrong number of fields
```

## Details

Repro is with Zed commit fa48157.

The CSV reader doesn't seem to have the problem.

```
$ zq -version
Version: v1.11.1-13-gfa48157f

$ echo '{foo: "bar", b: null, c: null}' | zq -f csv - | zq -i csv -Z -
{
    foo: "bar",
    b: null,
    c: null
}
```

And it works fine as TSV if I stick to just a single `null` value.

```
$ echo '{foo: "bar", b: null}' | zq -f tsv - | zq -i tsv -Z -
{
    foo: "bar",
    b: null
}
```

mattnibs:
Closes #4911

### DIFF
--- a/zio/anyio/ztests/tsv.yaml
+++ b/zio/anyio/ztests/tsv.yaml
@@ -1,13 +1,16 @@
+# This test purposely has two null values to test issue found in
+# brimdata/zed#4911
 script: |
   zq - | zq -f tsv -
+
 
 inputs:
   - name: stdin
     data: &stdin |
       #	Year	Winery
       1	2019	Brutocao
-      1	2021	Cabana
-      1	2020	Castoro
+      2		
+      3	2020	Castoro
 
 outputs:
   - name: stdout
--- a/zio/anyio/ztests/tsv.yaml
+++ b/zio/anyio/ztests/tsv.yaml
@@ -1,8 +1,7 @@
-# This test purposely has two null values to test issue found in
-# brimdata/zed#4911
+# This test purposely has two null values to test the issue found in
+# brimdata/zed#4911.
 script: |
   zq - | zq -f tsv -
-
 
 inputs:
   - name: stdin
--- a/zio/csvio/reader.go
+++ b/zio/csvio/reader.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"slices"
 	"strconv"
+	"unicode"
 
 	"github.com/brimdata/zed"
 	"github.com/brimdata/zed/zson"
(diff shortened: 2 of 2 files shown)


philrz:
@mattnibs: Thanks for that fix. It definitely addresses the example from #4911, so it's a functional 👍 for me.

Unfortunately, while testing that I bumped into another problem with a similar flavor that was waiting in line right behind. It has a similar "CSV works but TSV doesn't" flavor so I figured I'd flag it here rather than open yet another issue in the event it can be knocked out easily. I'm testing at commit dcbcbf8 on this branch.

This works with CSV but not TSV:

```
$ zq -version
Version: v1.11.1-14-gdcbcbf82

$ echo '{foo: "hello,world", bar: ["baz"]}' | zq -f csv - | zq -i csv -
{foo:"hello,world",bar:"[\"baz\"]"}

$ echo '{foo: "hello,world", bar: ["baz"]}' | zq -f tsv - | zq -i tsv -
stdio:stdin: parse error on line 2, column 7: bare " in non-quoted-field
```

Granted, I recognize that even in the CSV case it gets read back as a string rather than an array. But I accept that there's going to be complexity with some types that simple won't be reversible in these primitive formats. I'm just reacting to the parse error since it means the user doesn't get to see any data.

---
BTW, if you're curious about the workflow that's coming up with these, what I stumbled onto during verification was trying `zq -f tsv 'fuse' sample.zng | zq -i tsv -` with the [sample.zng](https://github.com/brimdata/zui/blob/main/apps/zui/src/test/shared/data/sample.zng) test data from the Zui repo. It seems like an abusive thing to do but it does happen to boomerang without complaint in CSV format so I figured I'd watch the same with TSV. When I've seen it fail I've trimmed down the test data to the simpler repros like the one in this comment and the one from #4911 as I figured that was kinder than just dumping the massive repro on you with the hundreds of fields. But if you're concerned there'll be a third round of whack-a-mole feel free to try running that one. 😄

zio/anyio/ztests/tsv.yaml
```diff
@@ -1,13 +1,16 @@
+# This test purposely has two null values to test issue found in
+# brimdata/zed#4911
```
nwt, line 2:
```suggestion
# This test purposely has two null values to test the issue found in
# brimdata/zed#4911.
```

nwt (approved):

philrz:
The bonus bug described in https://github.com/brimdata/zed/pull/4912#issuecomment-1832895486 above didn't get taken care of here so I've opened a new bug #4918 to track having that looked at.

PR #4915, opened 2023-12-01
Fix bug with nest_dotted argument

### LINKED ISSUE #4914: nest_dotted not working on a named record value
## tl;dr

The following output seems incorrect.

```
$ echo '{"inner":{"foo.bar":1}}' | zq -Z 'yield nest_dotted(inner)' -
{
    inner: {
        "foo.bar": 1
    }
}
```

I'd expected it to be:

```
{
    foo: {
        bar: 1
    }
}
```

## Details

Repro is with Zed commit 40739bd. This issue was spotted during an exchange with a community user in https://github.com/brimdata/zed/issues/4899#issuecomment-1833821769.

`nest_dotted()` seems to work as expected if applied on a top-level `this` value, e.g.

```
$ zq -version
Version: v1.11.1-16-g40739bdc

$ echo '{"foo.bar":1}' | zq -Z 'nest_dotted(this)' -
{
    foo: {
        bar: 1
    }
}
```

I see that the [nest_dotted.yaml](https://github.com/brimdata/zed/blob/main/runtime/expr/ztests/nest_dotted.yaml) ztest only has `this` examples, so perhaps we're just looking at untested territory.

mattnibs:
The pr fixes an issue with nest_dotted function not using the argument passed into it but instead always using the `this` value.

Closes #4914

### DIFF
--- a/runtime/expr/function/nestdotted.go
+++ b/runtime/expr/function/nestdotted.go
@@ -56,16 +56,19 @@ func (n *NestDotted) lookupBuilderAndType(in *zed.TypeRecord) (*zed.RecordBuilde
 }
 
 func (n *NestDotted) Call(ctx zed.Allocator, args []zed.Value) *zed.Value {
-	this := &args[0]
-	b, typ, err := n.lookupBuilderAndType(zed.TypeRecordOf(this.Type))
+	val := &args[0]
+	if len(args) > 1 {
+		val = &args[1]
+	}
+	b, typ, err := n.lookupBuilderAndType(zed.TypeRecordOf(val.Type))
 	if err != nil {
-		return wrapError(n.zctx, ctx, "nest_dotted: "+err.Error(), this)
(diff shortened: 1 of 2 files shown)


runtime/expr/function/nestdotted.go
```diff
@@ -56,16 +56,19 @@ func (n *NestDotted) lookupBuilderAndType(in *zed.TypeRecord) (*zed.RecordBuilde
 }
 
 func (n *NestDotted) Call(ctx zed.Allocator, args []zed.Value) *zed.Value {
-	this := &args[0]
-	b, typ, err := n.lookupBuilderAndType(zed.TypeRecordOf(this.Type))
+	val := &args[0]
+	if len(args) > 1 {
+		val = &args[1]
+	}
```
nwt, line 62: Maybe shorten?
```suggestion
	val := &args[len(args)-1]
```

nwt (approved):

PR #4919, opened 2023-12-01
csvio: Fix quote bug with non-comma delimiters

### LINKED ISSUE #4918: TSV reader fails to read back record with string+array
## tl;dr

With these examples we fail to read our own TSV output back as TSV input.

```
$ echo '{foo: "hello,world", bar: ["baz"]}' | zq -f tsv - | zq -i tsv -
stdio:stdin: parse error on line 2, column 7: bare " in non-quoted-field

$ echo '{foo: "helloworld", bar: ["baz"]}' | zq -f tsv - | zq -i tsv -
stdio:stdin: record on line 2: wrong number of fields
```

## Details

Repro is with Zed commit 4ae15b0.

The CSV reader doesn't seem to have the problem.

```
$ zq -version
Version: v1.11.1-17-g4ae15b03

$ echo '{foo: "hello,world", bar: ["baz"]}' | zq -f csv - | zq -i csv -
{foo:"hello,world",bar:"[\"baz\"]"}

$ echo '{foo: "helloworld", bar: ["baz"]}' | zq -f csv - | zq -i csv -
{foo:"helloworld",bar:"[\"baz\"]"}
```

It's understood that even in the CSV case `bar` gets read back as a string rather than an array. But I accept that there's going to be complexity with some types that simple won't be reversible in these primitive formats. I'm just reacting to the parse error since it means the user doesn't get to see any data.

If you're curious about the workflow that's coming up with bugs like this and #4911, what I stumbled onto during verification was trying `zq -f tsv 'fuse' sample.zng | zq -i tsv -` with the [sample.zng](https://github.com/brimdata/zui/blob/main/apps/zui/src/test/shared/data/sample.zng) test data from the Zui repo. It seems like an abusive thing to do but it does happen to boomerang without complaint in CSV format so I figured I'd watch the same with TSV. When I've seen it fail I've trimmed down the test data to the simpler repros like this and the one from #4911 as I figured that was kinder than just dumping the massive repro on you with the hundreds of fields. But if you're concerned there'll be another round of whack-a-mole feel free to try running a test with that bigger file. And when there's a PR up with a fix for this new issue I'll be sure to run it myself on the branch before we merge.

mattnibs:
Closes #4918

### DIFF
--- a/zio/anyio/ztests/tsv.yaml
+++ b/zio/anyio/ztests/tsv.yaml
@@ -10,6 +10,7 @@ inputs:
       1	2019	Brutocao
       2		
       3	2020	Castoro
+      4	2021	"[""baz""]"
 
 outputs:
   - name: stdout
--- a/zio/csvio/preprocess.go
+++ b/zio/csvio/preprocess.go
@@ -15,14 +15,19 @@ import (
 // Would get converted into:
 // field1,field2 extra
 type preprocess struct {
-	leftover []byte
-	scanner  *bufio.Reader
-	scratch  []byte
+	delimiter rune
+	leftover  []byte
+	scanner   *bufio.Reader
+	scratch   []byte
 }
 
-func newPreprocess(r io.Reader) *preprocess {
+func newPreprocess(r io.Reader, delim rune) *preprocess {
+	if delim == 0 {
+		delim = ','
+	}
 	return &preprocess{
-		scanner: bufio.NewReader(r),
+		delimiter: delim,
+		scanner:   bufio.NewReader(r),
 	}
 }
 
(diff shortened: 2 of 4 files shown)


mattnibs:
Note: If someone specifies a unicode delimiter this isn't going to work but I suggest we cross this bridge when we get there- this pr doesn't make that problem worse than it already was.

philrz:
I just tested this branch at commit 55eb303 and can see that it addresses both of the repro examples shown in the leading text of #4918. Also, the `zq -f tsv 'fuse' sample.zng | zq -i tsv -` I'd been using to find the last couple bugs now runs successfully. So I'm a functional 👍 on this one.

And thanks @mattnibs for pointing out the unicode delimiter exposure. I'm guessing that's not going to come up since tabs and semicolons are the only two alternates I've heard of people wanting to use, but indeed, I guess we'll see! 🤪

nwt (approved):

PR #4925, opened 2023-12-05
Add a vector engine

nwt:
* Add a vector engine in compiler/vam (for vector abstract machine) with a few initial operations

* Add dag.Vectorize, which tells the compiler to execute operations in its body using the vector engine

* Add compiler/optimize.Optimizer.Vectorize, which identifies a very limited set of vectorizable operations and wraps them in dag.Vectorize

* Add complier/kernel.Builder.compileVectorize, which compiles dag.Vectorize

Co-authored-by: @mccanne


nwt:
Tests in runtime/vcache/ztests/ are broken because "zed dev vcache copy" no longer works. I'd like to fix it rather than disable the tests.

mccanne (approved):
Merge when tests pass.

nwt:
> Tests in runtime/vcache/ztests/ are broken because "zed dev vcache copy" no longer works. I'd like to fix it rather than disable the tests.

We're going to fix "zed dev vcache copy" (and "zed dev vcache project", which is also broken) later, so I've disabled the corresponding tests.

PR #4926, opened 2023-12-05
switch: allow multiple parents

### LINKED ISSUE #4727: Can't chain back-to-back "switch" operators
Repro is with Zed commit e002355.

Input data in `input.csv`:

```
foo,bar
Ex,Gd
other,Po
```

User Op defined in `include.zed`:

```
op rating_to_num_op(rating): (
  switch rating (
    case "Ex" => rating := 5
    case "Gd" => rating := 4
    case "TA" => rating := 3
    case "Fa" => rating := 2
    case "Po" => rating := 1
    default => rating := 0
  )
)
```

I can call the User Op successfully on a single field.

```
$ cat one_field.zed
rating_to_num_op(foo)

$ zq -version
Version: v1.8.1-71-ge0023555

$ zq -I include.zed -I one_field.zed input.csv 
{foo:0,bar:"Po"}
{foo:5,bar:"Gd"}
```

However, if I try to chain multiple calls so I can run the User Op on multiple fields, I get an error.

```
$ cat two_fields.zed 
rating_to_num_op(foo)
| rating_to_num_op(bar)

$ zq -I include.zed -I two_fields.zed input.csv 
expression switch has multiple parents
```

mattnibs:
Fix the switch operator to not return an error when it receives multiple parents and instead merge the parents into a single parent.

Closes #4727

### DIFF
--- a/runtime/op/switcher/ztests/switch-chained.yaml
+++ b/runtime/op/switcher/ztests/switch-chained.yaml
@@ -1,23 +1,19 @@
 zed: |
-  switch foo (
-    case "Ex" => foo := 5
-    case "Gd" => foo := 4
-    case "TA" => foo := 3
-    case "Fa" => foo := 2
-    case "Po" => foo := 1
-    default => foo := 0
+  switch this (
+    case 0 => pass
+    case 1 => yield 2
   )
-  | switch bar (
-    case "Ex" => bar := 5
-    case "Gd" => bar := 4
-    case "TA" => bar := 3
-    case "Fa" => bar := 2
-    case "Po" => bar := 1
-    default => bar := 0
+  | switch this (
+    case 0 => pass
+    case 2 => yield 3
+  )
+  | switch (
+    case this==0 => pass
+    case this==3 => yield 4
   )
 
 input: |
-  {foo:"Gd",bar:"Fa"}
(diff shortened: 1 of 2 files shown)


runtime/op/switcher/ztests/switch-chained.yaml
```diff
@@ -0,0 +1,23 @@
...
+    case "Fa" => bar := 2
+    case "Po" => bar := 1
+    default => bar := 0
+  )
+
+input: |
+  {foo:"Gd",bar:"Fa"}
+
+output: |
+  {foo:4,bar:2}
```
nwt, line 23: Nit: Simplify program and input.

Not nit: Test both switch forms since you modified both compileExprSwitch and compileSwitch.

```suggestion
zed: |
  switch this (
    case 0 => pass
    case 1 => yield 2
  )
  | switch this (
    case 0 => pass
    case 2 => yield 3
  )
  | switch (
    case this==0 => pass
    case this==3 => yield 4
  )

input: |
  1

output: |
  4
```

mattnibs: Clever!

nwt (approved):
:shipit:

PR #4927, opened 2023-12-07
Fix typo in pip3 git url

rafaelsgirao:
Command to pip install zed is missing the `v` at the beginning of tags, making it fetch, e.g, `1.10.0` instead of `v1.10.0` - which fails. Adding a v before the local zed version fixes this.

### DIFF
--- a/docs/libraries/python.md
+++ b/docs/libraries/python.md
@@ -26,7 +26,7 @@ pip3 install "git+https://github.com/brimdata/zed#subdirectory=python/zed"
 
 Install the version compatible with a local `zed` like this:
 ```sh
-pip3 install "git+https://github.com/brimdata/zed@$(zed -version | cut -d ' ' -f 2)#subdirectory=python/zed"
+pip3 install "git+https://github.com/brimdata/zed@v$(zed -version | cut -d ' ' -f 2)#subdirectory=python/zed"
 ```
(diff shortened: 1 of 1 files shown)


rafaelsgirao:
Thanks for your reply!

I see now this is an issue with the `zed` package in the distribution I'm using (NixOS).

Your `zed version` returns `v1.10.0` while mine returns:

```console
❯ zed -version
Version: 1.10.0
```
The missing `v` in my distribution's  packaging is what's causing this issue.

I'll be closing this since this isn't at all a problem on your end.

(Just for reference, here's the offending line: 
 https://github.com/NixOS/nixpkgs/blob/d65bceaee0fb1e64363f7871bc43dc1c6ecad99f/pkgs/development/tools/zed/default.nix#L26C1-L26C56 )

Cheers!

philrz:
Thanks for the detail @rafaelsgirao. Indeed, we had no role in creating that NixOS package config so we had no idea this exposure exists. I really don't have much experience with NixOS but looking at several other things they repackage it looked like for each one they try to force version strings to look a certain way (i.e., without the leading `v`) so users may keep bumping into this. I've made a note-to-self to look closer at this at some point, and if you have NixOS experience you can share we'd be happy to hear it. 😄

PR #4930, opened 2023-12-07
semantic: return error on redefined constant

### LINKED ISSUE #4684: Earliest "type" statement is winning
Repro is with Zed commit 0c2f62f.

The [Named Types docs](https://zed.brimdata.io/docs/next/language/data-types#named-types) include the statement:

> Each instance of a named type definition overrides any earlier definition. In this way, types are local in scope.

It's not clear if that statement is speaking only of named types defined in the data stream or if it also applies to `type` definitions in the Zed program. Since it's vague, as a reader I'm assuming it's expected to be true for both, but upon testing, this appears not to be the case. For instance, if that quoted statement had been correct, I've have expected `int32` to be output here instead of `int16`.

```
$ zq -version
Version: v1.8.1-49-g0c2f62fa

$ echo {} | zq -z 'type myport=int16 type myport=int32 put b:=type(myport)' -
{b:<myport=int16>}
```

Since `type` definitions can only be placed in one spot currently (i.e., before the Zed pipeline), as a user I'd also have been ok with this generating an error, since I'm not certain if there's even a use case where it would make sense to define the same named type twice like this in a single Zed code block.

mattnibs:
Closes #4684

### DIFF
--- a/compiler/semantic/scope.go
+++ b/compiler/semantic/scope.go
@@ -66,8 +66,7 @@ func (s *Scope) DefineConst(zctx *zed.Context, name string, def dag.Expr) error 
 		Kind:  "Literal",
 		Value: zson.FormatValue(val),
 	}
-	s.DefineAs(name, literal)
-	return nil
+	return s.DefineAs(name, literal)
 }
 
 func (s *Scope) LookupExpr(name string) (dag.Expr, error) {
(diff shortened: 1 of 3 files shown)


nwt (approved):

mccanne:
Maybe add this test?
```
$ echo null | zq 'const a=1 ( const a=2 yield a )' -
2
```

Issue #4923, opened 2023-12-04
Changing search in "case" match results in join failure


philrz:
Repro is with Zed commit 6ff0586.

In a [recent Slack thread](https://brimdata.slack.com/archives/C03MW6XT7HC/p1701288001536269), a community user was getting help debugging their `join` use case. We were building off this working example of a "[self-join](https://zed.brimdata.io/docs/tutorials/join#self-joins)". 

```
$ cat all.json 
{"attr": {"addrrss": "1.2.3.4"}, "sessionid": 1, "thing": "evilthing", "s": "start"}
{"attr": {"addrrss": "1.2.3.4"}, "sessionid": 1, "thing": "evilthing", "s": "end", "message": "A match!"}

$ cat self-join.zed 
switch (
  case thing=="evilthing" and s=="start" => Fieldx:=ip(attr.addrrss), sessionstartid:=sessionid
  case thing=="evilthing" and s=="end" => Fieldx:=ip(attr.addrrss), sessionendid:=sessionid
) | left join on sessionstartid=sessionendid did_it_match:=message

$ zq -version
Version: v1.11.1-21-g6ff05865

$ zq -I self-join.zed all.json 
{attr:{addrrss:"1.2.3.4"},sessionid:1,thing:"evilthing",s:"start",Fieldx:1.2.3.4,sessionstartid:1,did_it_match:"A match!"}
```

The user had some difficulty applying this to their specific data, then circled back to report where they struggled. In their own words:

> Omg I found the problem - I was doing a regular search as part of the switch conditions - so something like case "searchterm" and attr=="val"

We can repro their experience by dropping the leading `thing==` portions in each `case`, which results in the `join` no longer producing output.

```
$ cat case-oops.zed 
switch (
  case "evilthing" and s=="start" => Fieldx:=ip(attr.addrrss), sessionstartid:=sessionid
  case "evilthing" and s=="end" => Fieldx:=ip(attr.addrrss), sessionendid:=sessionid
) | left join on sessionstartid=sessionendid did_it_match:=message

$ zq -I case-oops.zed all.json 
[no output]
```

What I find interesting is that if I run the `case` matches from each variant a la carte they produce the same values.

```
$ zq 'thing=="evilthing" and s=="start"' all.json 
{attr:{addrrss:"1.2.3.4"},sessionid:1,thing:"evilthing",s:"start"}

$ zq 'thing=="evilthing" and s=="end"' all.json 
[...]

PR #4931, opened 2023-12-11
switch: Check non-expression case is boolean expr

mattnibs:
Add check to ensure expressions for cases in non-expression switch statements are boolean expressions.

Part of #4923

### DIFF
--- a/compiler/semantic/op.go
+++ b/compiler/semantic/op.go
@@ -492,7 +492,7 @@ func (a *analyzer) semOp(o ast.Op, seq dag.Seq) (dag.Seq, error) {
 			// If this is a non-expression switch verify that the case
 			// is a boolean expression.
 			if expr == nil {
-				if !a.isBool(e) {
+				if !a.isBool(e, false) {
 					return nil, fmt.Errorf("switch: case expression must be a boolean expression")
 				}
 			}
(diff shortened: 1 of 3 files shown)


compiler/semantic/op.go
```diff
@@ -489,6 +489,13 @@ func (a *analyzer) semOp(o ast.Op, seq dag.Seq) (dag.Seq, error) {
 					Value: "true",
 				}
 			}
+			// If this is a non-expression switch verify that the case
+			// is a boolean expression.
+			if expr == nil {
+				if !a.isBool(e) {
+					return nil, fmt.Errorf("switch: case expression must be a boolean expression")
```
nwt, line 496: This is too strict. For example, it won't allow `switch (case a => yield b)`.

compiler/semantic/op.go
```diff
@@ -1021,25 +1028,29 @@ func (a *analyzer) semOpExpr(e ast.Expr, seq dag.Seq) (dag.Seq, error) {
...
 		switch e.Op {
 		case "and", "or", "in", "==", "!=", "<", "<=", ">", ">=":
 			return true
+		case "[":
+			return !strict
 		default:
 			return false
 		}
+	case *dag.This:
+		return !strict
```
mattnibs, line 1047: @nwt what about changing `isBool` to do something like this? Seems to work but I'm not sure if this covers all cases.

nwt: I think this is the right direction. I too am not sure if this covers all the cases (but we need to be before we merge something like this).

PR #4956, opened 2023-12-24
change vam.Puller.PullVec() to Pull()

mccanne:
This commit changes PullVec to Pull and adds an explicit materialization step between vector world and sequence world.  We also fixed a bug in the CountByString and Sum hacks introduced by the vector.Nulls wrapper.

### DIFF
--- a/runtime/vam/materialize.go
+++ b/runtime/vam/materialize.go
@@ -1,94 +1,43 @@
 package vam
 
 import (
-	"errors"
+	"bytes"
 
-	"github.com/brimdata/zed/vector"
+	"github.com/brimdata/zed"
 	"github.com/brimdata/zed/zbuf"
+	"github.com/brimdata/zed/zcode"
 )
 
+type Materializer struct {
+	parent Puller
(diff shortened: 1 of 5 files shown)


runtime/vam/materialize.go
```diff
@@ -1,94 +1,43 @@
...
+	for {
+		if !b(builder) {
+			return zbuf.NewArray(vals), nil
 		}
-		return false
+		//XXX Body should only be called for container types, but this
+		// will change soon anyway when we change out vector Builder to a
+		// slot-based indexing approach that isn't based on closures.
+		zv := zed.NewValue(typ, bytes.Clone(builder.Bytes().Body()))
+		vals = append(vals, *zv)
```
nwt, line 40: Nit: We've switched from zv to val everywhere.
```suggestion
		val := zed.NewValue(typ, bytes.Clone(builder.Bytes().Body()))
		vals = append(vals, val)
```

nwt (approved):

Issue #4815, opened 2023-10-24
vector engine stubs


mccanne:
This work involves stubbing out the initial and simple Vector interface type for the vector engine and locating it in a package under the runtime.  A rudimentary implementation will allow for vectors to be loaded through the vector cache and operated upon from a `zed dev` command.  This code will otherwise not be available to the general zed runtime as that will come later.

PR #4959, opened 2023-12-29
Remove vector.Vector et al. in favor of vector.Any

nwt:


mccanne (approved):

mccanne:
Doesn't this close an issue?

nwt:
@mccanne: Are you thinking of #4815? If so, #4925 probably comes closer to closing that one.

PR #4961, opened 2023-12-30
Docs for "zed manage" (i.e., compaction)

### LINKED ISSUE #4115: Document compaction
The following text was present in a retired "lake design" document (see #3803). It has been established that this was really a pending to-do, so this issue tracks its ultimate implementation and corresponding updates to docs.

```
### Compaction

To perform an LSM rollup, the `compact` command (implementation tracked
via [zed/2977](https://github.com/brimdata/zed/issues/2977))
is like a "squash" to perform LSM-like compaction function, e.g.,

zed compact <id> [<id> ...]
(merged commit <id> printed to stdout)

After compaction, all of the objects comprising the new commit are sorted
and non-overlapping.
Here, the objects from the given commit IDs are read and compacted into
a new commit.  Again, until the data is actually committed,
no readers will see any change.

Unlike other systems based on LSM, the rollups here are envisioned to be
run by orchestration agents operating on the Zed lake API.  Using
meta-queries, an agent can introspect the layout of data, perform
some computational geometry, and decide how and what to compact.
The nature of this orchestration is highly workload dependent so we plan
to develop a family of data-management orchestration agents optimized
for various use cases (e.g., continuously ingested logs vs. collections of
metrics that should be optimized with columnar form vs. slowly-changing
dimensional datasets like threat intel tables).

An orchestration layer outside of the Zed lake is responsible for defining
policy over
how data is ingested and committed and rolled up.  Depending on the
use case and workflow, we envision that some amount of overlapping data objects
would persist at small scale and always be "mixed in" with other overlapping
data during any key-range scan.

> Note: since this style of data organization follows the LSM pattern,
> how data is rolled up (or not) can control the degree of LSM write
> amplification that occurs for a given workload.  There is an explicit
> tradeoff here between overhead of merging overlapping objects on read
> and LSM write amplification to organize the data to avoid such overlaps.
>
> Note: we are showing here manual, CLI-driven steps to accomplish these tasks
> but a live data pipeline would automate all of this with orchestration that
> performs these functions via a service API, i.e., the same service API
> used by the CLI operators.
```

Note that the referenced issue #2977 has been completed and so this is at least partially implemented. However, the docs at https://zed.brimdata.io/docs/next/commands/zed#28-load still speak of compaction as a future enhancement, and indeed, much of the functionality described in the text above is expected to be covered by the "lake manager" tracked in #3923. Therefore, it seems _this_ specific issue can be used to track when we think we've completed enough to reveal compaction as a feature we want users to start using, and make sure we've updated the docs with enough guidance that they'll be successful.

philrz:
Now that compaction is available via `zed manage` and we have at least one environment where we know it's being used successfully, we're due to reveal it in the docs rather than speaking of it as a planned feature.

Closes #4115

### DIFF
--- a/docs/commands/zed.md
+++ b/docs/commands/zed.md
@@ -589,7 +589,7 @@ alternate check frequency in [duration format](../formats/zson.md#23-primitive-v
 If `-monitor` is not specified, a single maintenance pass is performed on the
 lake.
 
-The log output from `manage` provides a per-pool summary of the maintenance
+The output from `manage` provides a per-pool summary of the maintenance
 performed, including a count of `objects_compacted`.
 
 ### Merge
--- a/docs/commands/zed.md
+++ b/docs/commands/zed.md
@@ -465,13 +465,9 @@ When data is loaded, it is broken up into objects of a target size determined
 by the pool's `threshold` parameter (which defaults 500MiB but can be configured
 when the pool is created).  Each object is sorted by the pool key but
 a sequence of objects is not guaranteed to be globally sorted.  When lots
-of small or unsorted commits occur, data can be fragmented impacting performance.
-
-> Note that data is easily compacted by reading from a fragmented pool and writing
-> it back to a target pool so that it is globally sorted and compacted into
-> contiguous large objects.  We will soon introduce a compaction feature that
-> does this automatically inside of a pool and can either be run manually or
-> configured to run automatically by the server.
+of small or unsorted commits occur, data can be fragmented.  The performance
+impact of fragmentation can be eliminated by regularly [compacting](#manage)
+pools.
 
 For example, this command
 ```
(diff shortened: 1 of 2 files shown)


docs/commands/zed.md
```diff
@@ -574,6 +570,28 @@ API for automation.
 
 > Note that the branchlog meta-query source is not yet implemented.
 
+### Manage
+```
+zed manage [options]
+```
+Maintenance tasks are performed on a lake with the `manage` command.
```
nwt, line 577: Nit: Most of the other command sections start like this. I think it reads better too.
```suggestion
The `manage` command performs maintenance tasks on a lake.
```

docs/commands/zed.md
```diff
@@ -574,6 +570,28 @@ API for automation.
...
+
+Currently the only supported task is _compaction_, which reduces fragmentation
+by reading data objects in a pool and writing them back globally sorted and
+with their contents stored in contiguous large objects.
```
nwt, line 581: I don't think "globally sorted" and "contiguous" are accurate descriptions of what's going on here. How about "non-overlapping"?

```suggestion
by reading data objects in a pool and writing their contents back to large,
non-overlapping objects.
```

docs/commands/zed.md
```diff
@@ -574,6 +570,28 @@ API for automation.
...
+If the `-monitor` option is specified and the lake is [located](#locating-the-lake)
+via network connection, `zed manage` will run continuously and perform updates
+as needed.  By default a check is performed once per minute to determine if
+updates are necessary.  The `-interval` option may be used to specify an
+alternate check frequency in [duration format](../formats/zson.md#23-primitive-values).
+
+If `-monitor` is not specified, a single maintenance pass is performed on the
+lake.
+
+The log output from `manage` provides a per-pool summary of the maintenance
```
nwt, line 592: Nit: I don't think "log" adds anything here.
```suggestion
The output from `manage` provides a per-pool summary of the maintenance
```

docs/commands/zed.md
philrz: @nwt: Thanks for this suggestion. Happy to take it. In case it wasn't noticed, the wording I was using was taken directly from text that I was removing elsewhere in this PR, i.e., what I'm quoting below was an earlier forward-looking statement that's still present right now in our docs but will be removed if this PR merges. Now that compaction has been implemented I certainly recognize our way of talking about it may have evolved, but that was not obvious to me.

> Note that data is easily compacted by reading from a fragmented pool and writing it back to a target pool so that it is globally sorted and compacted into contiguous large objects.

nwt (approved):

docs/commands/zed.md
nwt: @philrz: I hadn't realized that.

PR #4963, opened 2024-01-02
Add keepalive endpoint

### LINKED ISSUE #4863: Add "long poll" lake API endpoint
Needed for https://github.com/brimdata/zui/issues/2884.

mattnibs:
This is used by Zui to prevent orhpaned Zed processes by exiting when a client stops polling the keepalive endpoint.

Closes #4863


service/core.go
```diff
@@ -49,6 +49,12 @@ type Config struct {
 	RootContent           io.ReadSeeker
 	Version               string
 	Logger                *zap.Logger
+	// EnableStayAlive when true enables the stayalive endpoint. If a client
+	// begins polling the stayalive endpoint then exits, the Zed process will
+	// also exit. This is used by the Zui application to ensure that the Zed
+	// process exits in the event that the Zui process is terminated with
+	// SIGKILL.
+	EnableStayAlive bool
```
nwt, line 57: For parity with -brimfd, I think we want this to be something like `StayAliveTimeout time.Duration` that, if greater than zero, both enables the /stayalive route and starts a timer that will kill the server on expiration if it hasn't received a /stayalive request.

mattnibs: Done.

mccanne:
Cool!  I think `keepalive` should return the reason the server exits but not sure how to wire that up, e.g., `"zed serve exited on SIGINT"`.  Not a big deal and could do this later if warranted.

mattnibs:
Closing, going in a different direction: https://github.com/brimdata/zui/pull/2956

PR #4970, opened 2024-01-08
Remove core.Shutdown()

mattnibs:
It does nothing except emit a log message that doesn't tell us anything we don't already know.


nwt (commented):
Any reason not to remove Core.Shutdown in service/core.go?

nwt (approved):

PR #4983, opened 2024-01-11
Change most remaining instances of *zed.Value to zed.Value

nwt:
Remaining instances generally appear in contexts in which nil signals that no zed.Value is available, as with zio.Reader.Read's first return parameter.

This change removes all of zed.Allocator's methods and renders runtime/expr.ResetContext unnecessary.  Both types will eventually go away.

### DIFF
--- a/runtime/op/uniq/uniq.go
+++ b/runtime/op/uniq/uniq.go
@@ -26,7 +26,7 @@ func New(octx *op.Context, parent zbuf.Puller, cflag bool) *Op {
 	}
 }
 
-func (o *Op) wrap(t *zed.Value) *zed.Value {
+func (o *Op) wrap(t *zed.Value) zed.Value {
 	if o.cflag {
 		o.builder.Reset()
 		o.builder.Append(t.Bytes())
@@ -37,20 +37,20 @@ func (o *Op) wrap(t *zed.Value) *zed.Value {
 		})
 		return zed.NewValue(typ, o.builder.Bytes()).Copy()
 	}
-	return t
+	return *t
 }
 
 func (o *Op) appendUniq(out []zed.Value, t *zed.Value) []zed.Value {
 	if o.count == 0 {
-		o.last = t.Copy()
+		o.last = t.Copy().Ptr()
 		o.count = 1
 		return out
 	} else if bytes.Equal(t.Bytes(), o.last.Bytes()) {
 		o.count++
 		return out
 	}
-	out = append(out, *o.wrap(o.last))
-	o.last = t.Copy()
+	out = append(out, o.wrap(o.last))
+	o.last = t.Copy().Ptr()
 	o.count = 1
 	return out
 }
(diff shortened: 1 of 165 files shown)


value.go
```diff
@@ -351,22 +350,21 @@ func (r *Value) nth(n int) zcode.Bytes {
 	return zv
 }
 
-func (r *Value) Fields() []Field {
+func (r Value) Fields() []Field {
 	return TypeRecordOf(r.Type()).Fields
 }
 
 func (v *Value) DerefByColumn(col int) *Value {
```
mattnibs, line 357: Shouldn't this return a non-pointer Value?

value.go
```diff
@@ -351,22 +350,21 @@ func (r *Value) nth(n int) zcode.Bytes {
...
 	if v != nil {
 		if bytes := v.nth(col); bytes != nil {
-			v = NewValue(v.Fields()[col].Type, bytes)
-		} else {
-			v = nil
+			val := NewValue(v.Fields()[col].Type, bytes)
+			return &val
```
mattnibs, line 361:
```suggestion
			return NewValue(v.Fields()[col].Type, bytes).Ptr()
```

value.go
mattnibs, line 357: Maybe this should just return a non-pointer Value and bool if it exists?

value.go
```diff
@@ -429,36 +427,30 @@ func (v *Value) AsTime() nano.Ts {
 
 func (v *Value) MissingAsNull() *Value {
```
mattnibs, line 428: Why keep this as a non-pointer? And also perhaps `v.IsMissing()` should get rid of the check for `v` being nil since this isn't a thing any longer?

runtime/op/uniq/uniq.go
```diff
@@ -26,7 +26,7 @@ func New(octx *op.Context, parent zbuf.Puller, cflag bool) *Op {
 	}
 }
 
-func (o *Op) wrap(t *zed.Value) *zed.Value {
+func (o *Op) wrap(t *zed.Value) zed.Value {
```
mattnibs, line 29: Why not nix the rest of the pointers in this file including `Op.last`, the `t` arg here and in appendUniq?

value.go
nwt: Quite possibly, though then we'll lose the ability to do method chaining, as in val.DerefByColumn(1).DerefByColumn(2). But given the size of this change so far, I think it's something we should address later. Sound OK?

value.go
nwt: Good catch. No reason for this to return a pointer.

To return a value I'll have to remove the second line here, but that seems fine.

https://github.com/brimdata/zed/blob/60bbaa1706c3596dfbd0fa3f0701280d7c165669/zio/zeekio/parser_test.go#L82-L83

runtime/op/uniq/uniq.go
nwt: I tried but there's a comparison between `Op.last` and `nil` in `Op.Pull` that appears to be significant.

value.go
nwt: Done in 1af46ba90da7bae8bbd61223eaffbcfc5d0eec37.

mccanne (approved):
Looks great!  So much nicer.  Let's rock and roll!!!

PR #4984, opened 2024-01-15
simplify and improve VNG format

mccanne:
This commit simplifies the VNG format with a new VNG metadata object called Variant, which represents a sequence of typed values (without putting the types into a union), in line with the Zed data model. This replaces the ad hoc data structures that were encoded in the VNG trailer.  We also changed vectors so they are all represented in a single segment in the data section, which simplifies vector stats and reading.  To support streaming, we replaced the trailer with a header and added framing so that VNG objects can simply be concatenated in a stream or file.  We added the command "zed dev vng" to displace the fixed VNG framing header (with magic) and the metadata as Zed data.

Nulls are now handled completely by the Nulls metadata structure and vector metadata now includes length information, which will be used by the vcache in a subsequent commit to faciliate easier flattening of leaf vectors so that all leaf values within a type all have the same length (except perhaps for vectors below unions, TBD).

The VNG write path now stores each object entirely in memory.  Once all writes have completed, the vectors are encoded in parallel (e.g., applying compression). Then a DFS traversal computes the data layout and produces a single Zed value representing the vector metadata with data locations.  The header, metadatasection, and data vectors are  then written out sequentially from a single thread such that the buffers end up being written in the same locations that were computed in the metadata pass.


cmd/zed/dev/vng/command.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/brimdata/zed/vng"
+	"github.com/brimdata/zed/zio"
+	"github.com/brimdata/zed/zio/zngio"
+	"github.com/brimdata/zed/zson"
+)
+
+var Cmd = &charm.Spec{
+	Name:  "vng",
+	Usage: "vng uri",
+	Short: "dump vng meta data",
```
nwt, line 25:
```suggestion
	Short: "dump vng metadata",
```

zio/vngio/writer.go
```diff
@@ -3,28 +3,10 @@ package vngio
...
-	DefaultColumnThresh = 5 * 1024 * 1024
-	DefaultSkewThresh   = 25 * 1024 * 1024
-)
-
-type WriterOpts struct {
-	ColumnThresh units.Bytes
-	SkewThresh   units.Bytes
-}
-
 // NewWriter returns a writer to w with reasonable default options.
```
nwt, line 9: Since the options have gone away.
```suggestion
// NewWriter returns a writer to w.
```

cli/outputflags/flags.go
```diff
@@ -42,12 +41,6 @@ func (f *Flags) Options() anyio.WriterOpts {
 func (f *Flags) setFlags(fs *flag.FlagSet) {
 	// zio stuff
 	fs.BoolVar(&f.color, "color", true, "enable/disable color formatting for -Z and lake text output")
-	f.VNG = &vngio.WriterOpts{
-		ColumnThresh: vngio.DefaultColumnThresh,
-		SkewThresh:   vngio.DefaultSkewThresh,
-	}
-	fs.Var(&f.VNG.ColumnThresh, "vng.colthresh", "minimum VNG frame size")
-	fs.Var(&f.VNG.SkewThresh, "vng.skewthresh", "minimum VNG skew size")
```
nwt, line 50: Since we're removing this, let's remove vng/ztests/skewthresh.yaml too.

cmd/zed/dev/vng/command.go
```diff
@@ -0,0 +1,144 @@
...
+	Long: `
+vng decodes an input uri and emits the metadata sections in the format desired.`,
```
nwt, line 27: Maybe shorten this?
```suggestion
The vng command prints a file's metadata sections.`,
```

cmd/zed/dev/vng/command.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+func (c *Command) Run(args []string) error {
+	ctx, cleanup, err := c.Init(&c.outputFlags)
+	if err != nil {
+		return err
+	}
+	defer cleanup()
+	if len(args) != 1 {
+		return errors.New("a single file required")
```
nwt, line 53:
```suggestion
		return errors.New("a single file is required")
```

cmd/zed/dev/vng/command.go
```diff
@@ -0,0 +1,144 @@
...
+			return r.marshaler.Marshal(hdr)
+		}
+		val, err := r.meta.Read()
+		if err != nil {
+			return nil, err
+		}
+		if val != nil {
+			return val, nil
+		}
+		r.meta = nil
```
nwt, line 117: Should close this before discarding it.
```suggestion
		r.meta.Close()
		r.meta = nil
```

vng/object.go
```diff
@@ -1,120 +1,97 @@
...
 }
 
-func NewObjectFromURI(ctx context.Context, zctx *zed.Context, engine storage.Engine, uri *storage.URI) (*Object, error) {
-	r, err := engine.Get(ctx, uri)
+func (o *Object) NewReader() (zio.Reader, error) {
+	return vector.NewZedReader(o.zctx, o.meta, o.readerAt)
+}
+
+func readMetadata(zctx *zed.Context, r io.Reader) (vector.Metadata, error) {
+	zr := zngio.NewReader(zctx, r)
```
nwt, line 73:
```suggestion
	zr := zngio.NewReader(zctx, r)
	defer zr.Close()
```

cmd/zed/dev/vng/command.go
```diff
@@ -0,0 +1,144 @@
...
+			r.meta = zngio.NewReader(r.zctx, io.LimitReader(r.reader, int64(hdr.MetaSize)))
+			r.dataSize = int(hdr.DataSize)
+			return r.marshaler.Marshal(hdr)
+		}
+		val, err := r.meta.Read()
+		if err != nil {
+			return nil, err
+		}
+		if val != nil {
+			return val, nil
```
nwt, line 115:
```suggestion
		if val != nil || err != nil {
			return val, err
```

cmd/zed/dev/vng/command.go
mccanne: Input doesn't need to be a file and it doesn't "print" since it serializes in the output format desired.  Maybe there's a better way to say this.

mccanne: I don't think of -z as printing.

mccanne: Since this is ubiquitous in all of our commands, maybe we should make an issue to come up with generic language for this and update all the help text.

nwt: Let's do that.

vng/vector/array.go
```diff
@@ -5,45 +5,57 @@ import (
...
 }
 
-func (a *ArrayWriter) Metadata() Metadata {
-	return &Array{
-		Lengths: a.lengths.Segmap(),
-		Values:  a.values.Metadata(),
+func (a *ArrayWriter) Metadata(off uint64) (uint64, Metadata) {
+	var lens, vals Metadata
+	off, lens = a.lengths.Metadata(off)
+	off, vals = a.values.Metadata(off)
```
nwt, line 54:
```suggestion
	off, lens := a.lengths.Metadata(off)
	off, vals := a.values.Metadata(off)
```

vng/vector/io.go
```diff
@@ -34,50 +34,52 @@ package vector
...
-	// between vectors.
-	Write(zcode.Bytes) error
-	// Push all in-memory vector data to the storage layer.
-	Flush(bool) error
+	// Write encodes the given value into memory.
+	Write(zcode.Bytes)
+	// Encoded all in-memory vector data into its storage-ready serialized format.
+	// Vectors may be encoded concurrently and errgroup.Group is used to sync
+	// and return errors.
+	Encode(*errgroup.Group)
```
nwt, line 49: This is nice.

vng/vector/io.go
```diff
@@ -34,50 +34,52 @@ package vector
...
 	// Metadata returns the data structure conforming to the VNG specification
 	// describing the layout of vectors.  This is called after all data is
-	// written and flushed by the Writer with the result marshaled to build
-	// the metadata section of the VNG file.
-	Metadata() Metadata
+	// written and encoded by the Writer with the result marshaled to build
+	// the header section of the VNG object.  An offset is passed down into
+	//  the traversal representing where in the data section the vector data
```
nwt, line 54:
```suggestion
	// the traversal representing where in the data section the vector data
```

vng/vector/variant.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	for _, value := range v.values {
+		if err := value.Emit(w); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+type VariantBuilder struct {
```
nwt, line 91: Maybe don't export this or VectorBuilder?

vng/writer.go
```diff
@@ -63,96 +38,43 @@ func (w *Writer) Close() error {
...
+	if err != nil {
+		//XXX can encode return an error?
+		return fmt.Errorf("system error: could not encode VNG metadata: %w", err)
 	}
 	// At this point all the vector data has been written out
 	// to the underlying spiller, so we start writing zng at this point.
-	zw := zngio.NewWriter(w.writer)
-	dataSize := w.spiller.Position()
+	metaBytes := &bytes.Buffer{}
+	zw := zngio.NewWriter(zio.NopCloser(metaBytes))
```
nwt, line 54:
```suggestion
	var metaBytes bytes.Buffer
	zw := zngio.NewWriter(zio.NopCloser(&metaBytes))
```

nwt: Also, metaBuf might be a better name.

vng/vector/array.go
mccanne: Didn't realize you could do that!

cmd/zed/dev/vng/command.go
mccanne: > Let's do that.

#4986

cmd/zed/dev/vng/command.go
nwt: Don't forget!

vng/writer.go
```diff
@@ -63,96 +38,43 @@ func (w *Writer) Close() error {
...
-	return w.root.Flush(eof)
+	return w.variant.Write(val)
 }
 
 func (w *Writer) finalize() error {
-	if err := w.flush(true); err != nil {
-		return err
+	meta, dataSize, err := w.variant.Encode()
+	if err != nil {
+		//XXX can encode return an error?
```
nwt, line 48: Seems like the answer here, at least for now, is yes.

nwt (approved):

PR #4987, opened 2024-01-15
consolidate package vng/vector into package vng

mccanne:
This commit moves package vng/vector into package vng and renames the Reader/Writer pattern to Builder/Encoder to better reflect the transitioned design.

There are no code changes here.  It is strictly renaming.


runtime/vcache/array.go
```diff
@@ -7,10 +7,10 @@ import (
 	"github.com/brimdata/zed/pkg/field"
 	"github.com/brimdata/zed/vector"
 	"github.com/brimdata/zed/vng"
-	meta "github.com/brimdata/zed/vng/vector" //XXX rename package
+	//XXX rename package
```
nwt, line 10:
```suggestion
```

vng/object.go
```diff
@@ -1,14 +1,36 @@
-// Package vng implements the reading and writing of VNG storage objects
-// to and from any Zed format.  The VNG storage format is described
-// at https://github.com/brimdata/zed/blob/main/docs/formats/vng.md.
+// Package vng implements the reading and writing of VNG serialization objects.
+// The VNG format is described at https://github.com/brimdata/zed/blob/main/docs/formats/vng.md.
+//
+// A VNG object is created by allocating an Encoder for any top-level Zed type
+// via NewEncoder, which recursively descends into the Zed type, allocating an Encoder
+// for each node in the type tree.  The top-level zng body is written via a call
```
nwt, line 6:
```suggestion
// for each node in the type tree.  The top-level ZNG body is written via a call
```

vng/object.go
```diff
@@ -1,14 +1,36 @@
...
+// to Write.  Each vector buffers its data in memory until the the object is encoded.
```
nwt, line 7:
```suggestion
// to Write.  Each vector buffers its data in memory until the object is encoded.
```

vng/object.go
```diff
@@ -1,14 +1,36 @@
...
+//
+// After all of the Zed data is written, a metadata section is written consisting
+// of a single Zed value describing the lahyout of all the vector data obtained by
```
nwt, line 10:
```suggestion
// of a single Zed value describing the layout of all the vector data obtained by
```

vng/object.go
```diff
@@ -1,14 +1,36 @@
...
+// calling the Metadata method on the vng.Writer interface.
```
nwt, line 11:
```suggestion
// calling the Metadata method on the Writer interface.
```

vng/object.go
```diff
@@ -1,14 +1,36 @@
...
+//
+// Nulls for complex types are encoded by a special Nulls object.  Each complex
+// type is wrapped by a NullsEncoder, which runlength encodes any alternating
```
nwt, line 14:
```suggestion
// type is wrapped by a NullsEncoder, which run-length encodes alternating
```

vng/record.go
```diff
@@ -11,39 +11,39 @@ import (
 
 var ErrVectorMismatch = errors.New("zng record value doesn't match vector writer")
```
nwt, line 13: While you're here, delete this since it's unused?
```suggestion
```

nwt (approved):

vng/object.go
mccanne: Encoder

PR #4988, opened 2024-01-16
serve: Add version, root and auth to startup log

### LINKED ISSUE #4882: Print version & lake storage info in "zed serve" log
Repro is with Zed commit c17a623.

Consider what's shown currently when `zed serve` starts.

```
$ zed -lake lake serve
{"level":"info","ts":1700000387.53475,"logger":"core","msg":"Started"}
{"level":"info","ts":1700000387.53509,"logger":"httpd","msg":"Listening","addr":"[::]:9867"}
```

When debugging a problem reported recently by a community user, I came to recognize how much more helpful it would be if the "Started" message included some other detail such as the Zed version. As a support person, this saves me and the user the round-trip asking them to run `zed -version` and share the output. Also, in situations where they might have been running multiple Zed versions over time (e.g., the `zlake.log` with Zui) this provides a historical record of which Zed version _was_ running long ago.

As long as I'm asking for log improvements, I could guess another detail that might be helpful in the future that I've not yet found myself needing. For instance, especially given the recent changes in #4758 ([docs](https://zed.brimdata.io/docs/commands/zed#locating-the-lake)), I'm guessing it would also be helpful to know what storage location `zed serve` is using and how it was determined (e.g., `-lake` flag, an environment variable \[and if so, which one\], or fallback to default) since it would once again save a round-trip to ask the user.

mattnibs:
Closes #4882

### DIFF
--- a/service/core.go
+++ b/service/core.go
@@ -157,9 +157,9 @@ func NewCore(ctx context.Context, conf Config) (*Core, error) {
 
 	c.addAPIServerRoutes()
 	c.logger.Info("Started",
-		zap.String("version", conf.Version),
-		zap.String("root", path.String()),
 		zap.Bool("auth_enabled", conf.Auth.Enabled),
+		zap.Stringer("root", path),
+		zap.String("version", conf.Version),
 	)
 	return c, nil
 }
--- a/service/core.go
+++ b/service/core.go
@@ -156,7 +156,11 @@ func NewCore(ctx context.Context, conf Config) (*Core, error) {
 	}
 
 	c.addAPIServerRoutes()
-	c.logger.Info("Started")
+	c.logger.Info("Started",
+		zap.String("version", conf.Version),
+		zap.String("root", path.String()),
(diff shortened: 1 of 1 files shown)


service/core.go
```diff
@@ -156,7 +156,11 @@ func NewCore(ctx context.Context, conf Config) (*Core, error) {
 	}
 
 	c.addAPIServerRoutes()
-	c.logger.Info("Started")
+	c.logger.Info("Started",
+		zap.String("version", conf.Version),
+		zap.String("root", path.String()),
+		zap.Bool("auth_enabled", conf.Auth.Enabled),
```
nwt, line 162: Nits: Alphabetize these. And use zap.Stringer.
```suggestion
		zap.Bool("auth_enabled", conf.Auth.Enabled),
		zap.Stringer("root", path),
		zap.String("version", conf.Version),
```

nwt (approved):

Issue #4719, opened 2023-07-19
Smaller float types are broken


philrz:
## tl;dr

The `float16` and `float32` Zed types are broken.

## Details

Repro is with Zed commit 91ee553.

Start from this simple math.

```
$ zq -version
Version: v1.8.1-67-g91ee553a

$ zq 'yield 5 - 4'
1
```

If I perform this with all combinations of numeric Zed types that are currently implemented, I get bad results for the combinations that mix in the smaller float types like `float16` and `float32`.

```
$ cat types.txt 
uint8
uint16
uint32
uint64
int8
int16
int32
int64
float16
float32
float64

$ for type1 in $(cat types.txt); do for type2 in $(cat types.txt); do   echo -n "$type1 $type2: " ; echo "{num1: 5($type1), num2: 4($type2)}" | zq -z "yield num1 - num2" -; done; done | egrep -v " 1$"\|" 1\.$"\|" 1\.\("\|" 1\("
float16 float32: -Inf(float16)
float16 float64: -Inf(float16)
float32 float16: -17403.(float32)
float32 float64: -4616189618054758400.(float32)
float64 float16: -17403.
float64 float32: -1082130427.
```

philrz:
While fielding an inquiry from a community zync user I became aware that these smaller float types also have a subtle bug when used as join keys. Repro is with Zed commit 38763f8.

Start from this working example. In absence of explicit casting, the `num` value from the CSV input is stored as a `float64` and the `num` value from the ZSON input is stored as `int64`. The comparison of the `join` keys succeeds.

```
$ zq -version
Version: v1.14.0-16-g38763f82

$ cat f64.csv 
num,word
1,hello

$ cat i64.zson 
{num:1,word:"goodbye"}

$ cat join-1.zed 
file f64.csv
| inner join (
  file i64.zson
) on num=num otherword:=word

$ zq -I join-1.zed 
{num:1.,word:"hello",otherword:"goodbye"}
```

However, if we make the `num` value from the ZSON input into a `float32`, the `join` fails.

```
$ cat f32.zson 
{num:1 (float32),word:"goodbye"}

$ cat join-2.zed 
file f64.csv
| inner join (
  file f32.zson
) on num=num otherword:=word

$ zq -I join-2.zed 
[no output]
```

[...]

PR #4993, opened 2024-01-17
Change runtime/expr.testSuccessful expected param to ZSON

nwt:
testSuccessful takes its expected output as a zed.Value, which it must then format as ZSON.  Simplify things by changing that parameter to a ZSON string.

### DIFF
--- a/runtime/expr/expr_test.go
+++ b/runtime/expr/expr_test.go
@@ -2,18 +2,14 @@ package expr_test
 
 import (
 	"fmt"
-	"net/netip"
 	"strings"
 	"testing"
 
 	"github.com/brimdata/zed"
-	"github.com/brimdata/zed/pkg/nano"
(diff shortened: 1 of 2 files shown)


mattnibs (approved):

nwt:
Tests are failing because `zq 'time(float16(1589126400000000000))'` prints `2262-04-11T23:47:16.854775807Z` on arm64 but `1677-09-21T00:12:43.145224192Z` on amd64. Ugh.

philrz:
Related to #4719?

nwt:
It's a badly-designed test. `float16(1589126400000000000)` is `+Inf`, which is then converted to an integer, and https://go.dev/ref/spec#Conversions says, "In all non-constant conversions involving floating-point or complex values, if the result type cannot represent the value the conversion succeeds but the result value is implementation-dependent."

PR #4996, opened 2024-01-18
Fix parsing for type literals

mattnibs:
The currently parsing rules for type literals allow a type literal to be specified inside of a type literal (e.g., `<(int64, <string>)>)`. This pr fixes this undesired behavior.


mattnibs:
This pr actually changes a few behaviors that reviewers should be aware of:

1. The `type` constant will now not allow the type literal syntax (e.g., `type mytype=<int64>`). I think this makes sense but just wanted to point this out.
2. The explode op (which isn't even documented) will also not use allow the type literal syntax for the argument of `by`.

nwt (approved):

PR #5000, opened 2024-01-19
docs: Fix ZSON named type example

mattnibs:


philrz (approved):
Thanks for catching this @mattnibs.

I just took a brief tour of the history of changes in this area since I find it helps illustrate the value of mdtest-protecting the docs whenever we can. The syntax you're correcting here was first changed in the tooling to `(port=<uint16>)` with the changes in #3458. Then the `(port=uint16)` syntax you're now correctly showing arrived with the changes in #3498.

For readability I know not every example in the docs lends itself to the full mdtest treatment, but as long as they have to exist, maybe there's something else we can do to help make sure they get updated. For instance, while the relevant part of the `zson.md` didn't get touched in those two PRs, the tests certainly did. Would it help at all if we added comments to one or more of those test files effectively calling attention "Hey, if you find yourself updating this, these other parts of these docs probably need the same treatment"? Or does everyone only do bulk `sed`-style test updates such that comments would not be noticed? I'm curious what others might think (cc: @nwt).

nwt (approved):

mattnibs:
@philrz maybe for these kinds of tests we want to add a mdtest boomerang flag that 1) verifies it can serialize the value and 2) deserialize it to zson and 3) ensures the input/output values match?

philrz:
@mattnibs: I like your idea! We had an existing issue for this class of mdtest improvement so I just added your idea as https://github.com/brimdata/zed/issues/2908#issuecomment-1904744301. If someone wanted to knock that out as a palate cleanser I would not fight it since these little glitches in the docs have a way of sneaking up on us now and then and it bums me out to think of users that have taken the time to read the docs getting confused. But it's not a crisis if we leave it for later. The worst that happens is someone pings us on Slack.

PR #5001, opened 2024-01-21
improved vcache and beginnings of vector runtime

mccanne:
This commit substantially improves the vector cache by adding a mutable shadow layer that caches all the big pieces of vector data and builds projections of runtime vectors by traversing the shadow and building new mutable vectors with the query zctx whose data buffers point into the cache.

All the vector types now have proper nulls bitmasks.

We also added the mapping logic to adjust slot numbera a code paths go into and out of sparse unions and variants.

The vcache projection command has been re-enabled and some additional tests added to cover some corner cases.

Also, we added dictionary counterparts to all of the vector primitive types.  This exposes dictionary encoding to the runtime and will allow them to be exploited for better performance.

The hacky aggregate demos (sum and count-by-string) have been updated with all these changes and are working.


runtime/vam/agg.go
```diff
@@ -110,6 +153,7 @@ func NewSum(zctx *zed.Context, parent Puller, name string) *Sum {
 		parent: parent,
 		zctx:   zctx,
 		name:   name,
+		field:  NewDotExpr(zctx, &This{}, name),
```
nwt, line 156: Nit: Initialize in declaration order for reader's benefit.
```suggestion
		field:  NewDotExpr(zctx, &This{}, name),
		name:   name,
```

runtime/vam/agg.go
```diff
@@ -133,18 +177,34 @@ func (c *Sum) Pull(done bool) (vector.Any, error) {
...
+	}
+	vec = c.field.Eval(vec)
+	switch vec := vec.(type) {
+	case *vector.Int:
+		for _, x := range vec.Values {
+			c.sum += x
+		}
+	case *vector.DictInt:
+		for k := range vec.Values {
+			c.sum += vec.Values[k] * int64(vec.Counts[k])
```
nwt, line 199:
```suggestion
		for k, x := range vec.Values {
			c.sum += x * int64(vec.Counts[k])
```

runtime/vam/agg.go
```diff
@@ -133,18 +177,34 @@ func (c *Sum) Pull(done bool) (vector.Any, error) {
...
+		for _, x := range vec.Values {
+			c.sum += int64(x)
 		}
-		if vec, ok := vec.(*vector.Uint); ok {
-			for _, x := range vec.Values {
-				c.sum += int64(x)
-			}
+	case *vector.DictUint:
+		for k := range vec.Values {
+			c.sum += int64(vec.Values[k]) * int64(vec.Counts[k])
```
nwt, line 207:
```suggestion
		for k, x := range vec.Values {
			c.sum += int64(x) * int64(vec.Counts[k])
```

runtime/vam/agg.go
```diff
@@ -45,29 +47,49 @@ func (c *CountByString) Pull(done bool) (vector.Any, error) {
...
+	if vec, ok := vec.(*vector.Variant); ok {
+		for _, vec := range vec.Values {
+			c.update(vec)
 		}
-		//xxx
-		fmt.Printf("vector.CountByString: bad vec %s %T\n", zson.String(vec.Type()), vec)
+		return
+	}
+	vec = c.field.Eval(vec)
+	switch vec := vec.(type) {
```
nwt, line 62:
```suggestion
	switch vec := c.field.Eval(vec).(type) {
```

runtime/vam/dot.go
```diff
@@ -0,0 +1,56 @@
...
+	case *vector.Record:
+		i, ok := vec.Typ.IndexOfField(d.field)
+		if !ok {
+			return vector.NewMissing(d.zctx, vec.Len())
+		}
+		return vec.Fields[i]
+	case *vector.Map:
+		panic("vam.DotExpr Map TBD")
+	case *vector.TypeValue:
+		panic("vam.DotExpr TypeValue TBD")
```
nwt, line 46: Nit: Put this first so the cases follow the type system total order.

vector/tagmap.go
```diff
@@ -0,0 +1,48 @@
...
+	return NewTagMapFromLens(tags, lens)
+}
+
+func NewTagMapFromLens(tags []uint32, lens []uint32) *TagMap {
+	forward := make([]uint32, len(tags))
+	space := make([]uint32, len(tags))
+	reverse := make([][]uint32, len(lens))
+	var off uint32
+	for tag := range lens {
+		n := lens[tag]
```
nwt, line 27:
```suggestion
	for tag, n := range lens {
```

runtime/vcache/path.go
```diff
@@ -0,0 +1,61 @@
+package vcache
+
+import (
+	"fmt"
+
+	"github.com/brimdata/zed/pkg/field"
+)
+
+// A path is an array of string or Forks
+type Path []interface{} //XXX clean this up later
```
mattnibs, line 10:
```suggestion
type Path []any //XXX clean this up later
```

vector/union.go
```diff
@@ -6,40 +6,45 @@ import (
...
 var _ Any = (*Union)(nil)
 
-func NewUnion(typ *zed.TypeUnion) *Union {
-	return &Union{Typ: typ, Values: make([]Any, len(typ.Types))}
+func NewUnion(typ *zed.TypeUnion, tags []uint32, vals []Any, nulls *Bool) *Union {
+	return &Union{
+		Typ:    typ,
+		Tags:   tags,
+		TagMap: *NewTagMap(tags, vals),
+		Values: vals,
```
nwt, line 23: Don't forget nulls!

In constructors, I think a composite literal with a limited number of fields is often clearer without fields names. And if you omit them, you'll get a compilation error if you forget a field like this.
```suggestion
	return &Union{typ, tags, *NewTagMap(tags, vals), vals, nulls}
```

runtime/vam/materialize.go
```diff
@@ -25,19 +26,22 @@ func (m *Materializer) Pull(done bool) (zbuf.Batch, error) {
...
+		typ = vec.Type()
+	}
 	builder := zcode.NewBuilder()
 	var vals []zed.Value
-	for {
-		if !b(builder) {
-			return zbuf.NewArray(vals), nil
+	n := vec.Len()
+	for slot := uint32(0); slot < n; slot++ {
+		vec.Serialize(builder, slot)
```
mattnibs, line 38: This more just a curiosity but going forward with @nwt's ongoing work on native Values do we imagine that some point soon we won't want to call Serialize here and instead build using native values?

runtime/vcache/cache.go
```diff
@@ -19,27 +19,50 @@ type Cache struct {
...
+	c.Lock(id)
+	object, err := NewObject(ctx, c.engine, uri)
 	if err != nil {
+		c.Unlock(id)
 		return nil, err
 	}
+	c.mu.Lock()
 	c.objects[id] = object
+	c.mu.Unlock()
+	c.Unlock(id)
```
mattnibs, line 66: I think there's a race here and the second request to create the object will overwrite the first. I think you need to keep `c.mu` locked for the entire call here.

nwt, line 66: If you and I call `c.Lock(id)` concurrently and I block waiting for you, then you'll set `c.object[id]` to the object you've created and, when you call `c.Unlock(id)`, I'll overwrite that with the object I've created.

I'm not sure if that's OK, but if it isn't, we could do this.
```suggestion
	object, err := NewObject(ctx, c.engine, uri)
	if err != nil {
		return nil, err
	}
	c.mu.Lock()
	if o, ok := c.objects[id]; ok {
		object = o
	} else {
		c.objects[id] = object
	}
	c.mu.Unlock()
```

runtime/vam/agg.go
```diff
@@ -77,29 +99,50 @@ func (c *countByString) countFixed(vec *vector.Const) {
...
+	var nulls *vector.Bool
+	if c.nulls > 0 {
+		length++
+		counts = slices.Grow(counts, length)[0:length]
+		offs = slices.Grow(offs, length+1)[0 : length+1]
+		counts[k] = c.nulls
+		k++
+		offs[k] = uint32(len(bytes))
+		nulls = vector.NewBoolEmpty(uint32(k), nil)
+		nulls.Set(uint32(k - 1))
```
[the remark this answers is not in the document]
mccanne: This code is all going away and I futzed with it to bit to get it to work.  It's just for the hacky demo and there's no test coverage so I'd rather not manually not have to check that the change works for just right now.  It will disappear soon (vam-summarize branch).

runtime/vam/dot.go
mccanne: sequence world has this order

runtime/vam/materialize.go
mccanne: Yes, will need a better name for the method then.

runtime/vcache/cache.go
mccanne: Good catch but you can't hold mu because it protects the locks map. You just have to recheck the object condition after taking the fine-grained lock.

mccanne: There's probably a cleaner way to do this.

mccanne: It doesn't seem right to created an object that's not used

runtime/vcache/nulls.go
```diff
@@ -2,42 +2,104 @@ package vcache
...
+		var null bool
+		var off int
+		b := n.local
+		for {
+			run, err := runlens.Next()
+			if err != nil {
+				if err == io.EOF {
+					err = nil
+				}
+				n.meta = nil
```
nwt, line 48: Setting this to nil in response to a transient error seems like it would irrecoverably poison the shadow.
```suggestion
				if err == io.EOF {
					n.meta = nil
					return nil
				}
```

runtime/vcache/nulls.go
```diff
@@ -2,42 +2,104 @@ package vcache
...
+	})
+}
+
+func (n *nulls) flatten(parent *vector.Bool) *vector.Bool {
+	if n == nil {
+		return parent
+	}
+	n.mu.Lock()
+	defer n.mu.Unlock()
+	if n.done {
```
nwt, line 69: Drop the done field and just check for a non-nil flat?
```suggestion
	if n.flat != nil {
```

cmd/zed/dev/vcache/project/command.go
```diff
@@ -4,25 +4,28 @@ import (
...
 )
 
 var Project = &charm.Spec{
 	Name:  "project",
-	Usage: "project [flags] field[,field...] path",
+	Usage: "project [flags] path [field ...]",
 	Short: "read a VNG file and run a projection as a test",
 	Long: `
 The project command reads VNG vectors from
 a VNG storage objects (local files or s3 objects) and outputs
```
nwt, line 26:
```suggestion
VNG storage objects (local files or s3 objects) and outputs
```

runtime/vam/agg.go
```diff
@@ -133,18 +176,34 @@ func (c *Sum) Pull(done bool) (vector.Any, error) {
...
+	case *vector.Uint:
+		for _, x := range vec.Values {
+			c.sum += int64(x)
 		}
-		if vec, ok := vec.(*vector.Uint); ok {
-			for _, x := range vec.Values {
-				c.sum += int64(x)
-			}
+	case *vector.DictUint:
+		for k, val := range vec.Values {
```
nwt, line 205: Nit: Use the same variable names in all of these loops but avoid `val` since that's a `zed.Value` almost everywhere else. `x` is good, though I like `v`, which we often use for this sort of generic Go value.

runtime/vcache/cache.go
```diff
@@ -19,27 +19,55 @@ type Cache struct {
...
+		c.locks[id] = mu
+	}
+	mu.Lock()
+}
+
+func (c *Cache) Unlock(id ksuid.KSUID) {
+	c.mu.Lock()
+	c.locks[id].Unlock()
+	c.mu.Unlock()
+}
```
nwt, line 48: Nit: Don't export these (unless you plan to use them outside this package).

runtime/vcache/loader.go
```diff
@@ -0,0 +1,589 @@
...
+	"github.com/brimdata/zed/vector"
+	"github.com/brimdata/zed/vng"
+	"github.com/brimdata/zed/zcode"
+	"github.com/brimdata/zed/zson"
+	"golang.org/x/sync/errgroup"
+)
+
+// loader handles loading vector data on demand for only the fields needed
+// as specified in the projection.  Each load is executed with a multiphase
+// process: first, we build a mirror of the vng metadata where each node has a
```
nwt, line 19:
```suggestion
// process: first, we build a mirror of the VNG metadata where each node has a
```

runtime/vcache/loader.go
```diff
@@ -0,0 +1,589 @@
...
+// lock and places to store the bulk data so that it may be reused across
+// projections.  This is called the shadow object.  Then, we fill in the shadow
+// with data vectors dynamically and create runtime vectors as follows:
+//
+//	(1) create a mirror data structure (shadow)
+//	(2) concurrently load all the nulls and tags/lens/etc that will be needed (fetchNulls)
```
nwt, line 25:
```suggestion
//	(2) concurrently load all the nulls and tags, lens, etc. that will be needed (fetchNulls)
```

runtime/vcache/loader.go
```diff
@@ -0,0 +1,589 @@
...
+//	(3) compute top-down flattening of nulls (concurrently) (flatten)
```
nwt, line 26:
```suggestion
//	(3) compute top-down flattening of nulls concurrently (flatten)
```

runtime/vcache/nulls.go
```diff
@@ -2,42 +2,104 @@ package vcache
...
+	out := vector.NewBoolEmpty(n, nil)
+	var childSlot uint32
+	for slot := uint32(0); slot < n; slot++ {
+		if !parent.Value(slot) {
+			if child.Value(childSlot) {
+				out.Set(slot)
+			}
+			childSlot++
+		} else {
+			out.Set(slot)
```
nwt, line 101: Nit: I think this is clearer without the `!`.
```suggestion
		if parent.Value(slot) {
			out.Set(slot)
		} else {
			if child.Value(childSlot) {
				out.Set(slot)
			}
			childSlot++
		}
```

runtime/vcache/loader.go
```diff
@@ -0,0 +1,589 @@
...
+	r    io.ReaderAt
+}
+
+// Load all vector data into the in-memory shadow that is needed and not yet loaded
+// and return a new vector.Any using the data vectors in cache.  This may be called
+// concurrently on the same shadow and fine-grained locking insures that any given
+// data vector is loaded just once and such loads may be executed concurrently (even
+// when only one thread is calling load).  If paths is nil, then the entire value
+// is loaded.
+func (l *loader) load(paths Path, s shadow) (vector.Any, error) {
```
nwt, line 45: Nit: Shadow feels like it should come first because it's the thing from which you're loading. The paths just says what parts of the shadow to load.  (Same goes for all the other functions and methods that take a shadow and paths.)
```suggestion
func (l *loader) load(s shadow, paths Path) (vector.Any, error) {
```

runtime/vcache/project.go
```diff
@@ -0,0 +1,122 @@
...
+		}
+		return s.vec
+	case *const_:
+		if len(paths) > 0 {
+			return vector.NewMissing(zctx, s.length())
+		}
+		return s.vec
+	case *named:
+		v := project(zctx, paths, s.vals)
+		typ, _ := zctx.LookupTypeNamed(s.name, v.Type())
```
nwt, line 43:
```suggestion
		typ, err := zctx.LookupTypeNamed(s.name, v.Type())
		if err != nil {
			panic(err)
		}
```

runtime/vcache/path.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+	}
+	// No common prefix so add the addition to the fork.
+	return append(fork, convertFieldPath(addition))
+}
+
+func convertFieldPath(path field.Path) Path {
+	var out []interface{}
+	for _, s := range path {
+		out = append(out, interface{}(s))
```
mattnibs, line 58:
```suggestion
	var out []any
	for _, s := range path {
		out = append(out, s)
```

runtime/vcache/shadow.go
```diff
@@ -0,0 +1,203 @@
...
+type count struct {
+	vals  uint32
+	nulls uint32
+}
+
+func (c count) length() uint32 {
+	return c.nulls + c.vals
+}
+
+// newShadow converts the vng metadata structure to a complete vector.Any
```
nwt, line 120:
```suggestion
// newShadow converts the VNG metadata structure to a complete vector.Any
```

runtime/vcache/shadow.go
```diff
@@ -0,0 +1,203 @@
...
+			count: count{m.Len(), nullsCnt},
+			loc:   m.Tags,
+			vals:  vals,
+			nulls: nulls{meta: n},
+		}
+	case *vng.Primitive:
+		return &primitive{
+			count: count{m.Len(), nullsCnt},
+			nulls: nulls{meta: n},
+			vng:   m,
```
nwt, line 192:
```suggestion
			vng:   m,
			nulls: nulls{meta: n},
```

runtime/vcache/shadow.go
```diff
@@ -0,0 +1,203 @@
...
+		}
+	case *vng.Const:
+		return &const_{
+			count: count{m.Len(), nullsCnt},
+			nulls: nulls{meta: n},
+			val:   m.Value,
```
nwt, line 198:
```suggestion
			val:   m.Value,
			nulls: nulls{meta: n},
```

runtime/vcache/shadow.go
```diff
@@ -0,0 +1,203 @@
...
+			vals:  newShadow(m.Values, nil, 0),
+			nulls: nulls{meta: n},
+		}
+	case *vng.Map:
+		return &map_{
+			nulls: nulls{meta: n},
+			count: count{m.Len(), nullsCnt},
+			loc:   m.Lengths,
+			keys:  newShadow(m.Keys, nil, 0),
+			vals:  newShadow(m.Values, nil, 0),
```
nwt, line 175:
```suggestion
			count: count{m.Len(), nullsCnt},
			loc:   m.Lengths,
			keys:  newShadow(m.Keys, nil, 0),
			vals:  newShadow(m.Values, nil, 0),
			nulls: nulls{meta: n},
```

runtime/vcache/project.go
```diff
@@ -0,0 +1,122 @@
...
+	default:
+		panic(fmt.Sprintf("vector cache: shadow type %T not supported", s))
+	}
+}
+
+func projectVariant(zctx *zed.Context, paths Path, s *variant) vector.Any {
+	vecs := make([]vector.Any, 0, len(s.vals))
+	for _, m := range s.vals {
+		vec := project(zctx, paths, m)
+		vecs = append(vecs, vec)
```
nwt, line 54:
```suggestion
		vecs = append(vecs, project(zctx, paths, m))
```

runtime/vcache/project.go
```diff
@@ -0,0 +1,122 @@
...
+func projectRecord(zctx *zed.Context, paths Path, s *record) vector.Any {
+	if len(paths) == 0 {
+		// Build the whole record.  We're either loading all on demand (nil paths)
+		// or loading this record because it's referenced at the end of a projected path.
+		vals := make([]vector.Any, 0, len(s.fields))
+		types := make([]zed.Field, 0, len(s.fields))
+		for _, f := range s.fields {
+			v := project(zctx, nil, f.val)
+			vals = append(vals, v)
+			types = append(types, zed.Field{Name: f.name, Type: v.Type()})
```
nwt, line 68:
```suggestion
			vec := project(zctx, nil, f.val)
			vals = append(vals, vec)
			types = append(types, zed.Field{Name: f.name, Type: vec.Type()})
```

runtime/vcache/project.go
```diff
@@ -0,0 +1,122 @@
...
+	return vector.NewRecord(zctx.MustLookupTypeRecord(fields), vecs, length, nulls)
+}
+
+func projectUnion(zctx *zed.Context, paths Path, s *union) vector.Any {
+	vals := make([]vector.Any, 0, len(s.vals))
+	types := make([]zed.Type, 0, len(s.vals))
+	for _, val := range s.vals {
+		v := project(zctx, paths, val)
+		vals = append(vals, v)
+		types = append(types, v.Type())
```
nwt, line 119:
```suggestion
		vec := project(zctx, paths, val)
		vals = append(vals, vec)
		types = append(types, vec.Type())
```

runtime/vcache/nulls.go
```diff
@@ -2,42 +2,104 @@ package vcache
...
+		n.mu.Unlock()
+		return
+	}
+	n.mu.Unlock()
+	g.Go(func() error {
+		n.mu.Lock()
+		defer n.mu.Unlock()
+		if n.meta == nil {
+			return nil
 		}
```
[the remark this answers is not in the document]
mccanne: The idea is to be able to walk the shadow metadata quickly when the data is all in cache and not do anything heavier weight.  I know go channels are supposed to be lightweight but it seems like a bad idea to create a new thread for every node in a value hierarchy to load a vector that is already entirely in memory.  The idea is to create the new channel only when there is going to be I/O.

runtime/vcache/nulls.go
mccanne: Ok, I'll make this change but we should handle the error at the higher layer maybe by kicking the object out of the cache?  I'm not sure how the multi-phase algorithm would be easily restartable but maybe.

runtime/vam/agg.go
mccanne: I don't like x.  v is ok.  I think val is ok because when you are in vector world a value vector is called val.  If we want to change this convention there are a lot of places in VNG where we would need to change it.  This code in particular is going away soon.

runtime/vcache/loader.go
```diff
@@ -0,0 +1,589 @@
...
+	}
+}
+
+func indexOfField(name string, fields []field_) int {
+	for k, f := range fields {
+		if name == f.name {
+			return k
+		}
+	}
+	return -1
```
nwt, line 588:
```suggestion
	return slices.IndexFunc(fields, func (f field_) bool {
		return f.name == name
	})
```

runtime/vcache/loader.go
```diff
@@ -0,0 +1,589 @@
...
+		}
+	case *array:
+		s.nulls.flatten(parent)
+		flattenNulls(paths, s.vals, nil)
+	case *set:
+		s.nulls.flatten(parent)
+		flattenNulls(paths, s.vals, nil)
+	case *map_:
+		s.nulls.flatten(parent)
+		flattenNulls(nil, s.vals, nil)
```
nwt, line 564:
```suggestion
		flattenNulls(nil, s.keys, nil)
```

runtime/vcache/loader.go
mccanne: I'm gonna leave this one

runtime/vcache/project.go
mccanne: I'm going with val here for reason described above

nwt (approved):
I haven't looked as closely at runtime/vcache/loader.go as I'd like but I've been at this for a few hours now and am pretty worn.

runtime/vcache/project.go
mccanne: ask discussed

runtime/vcache/loader.go
mccanne: Oopsies.  Ugh not enough test coverage

PR #5005, opened 2024-01-23
align vector and sequence runtimes for easier coexistence

mccanne:
This commit rearranges the runtime codebase so that the sequence and vector runtimes coexist in parallel layout. We added the path runtime/sam for the sequential op and expr packages etc. and created op and expr packages for runtime/vam.

We moved op.Context to runtime.Context since it is shared between vam and sam.  To do so and avoid import loops, we moved runtime.Query and DeleteQuery into runtime/exac where they more naturally belonged and made interfaces to these implementations.

We also removed runtime/vam/index.go since it won't be needed.


runtime/compiler.go
```diff
@@ -0,0 +1,64 @@
...
+		io.Closer
+		zbuf.Meter
+	}{AsReader(q), q, q}
+}
+
+func CompileQuery(ctx context.Context, zctx *zed.Context, c Compiler, program ast.Seq, readers []zio.Reader) (Query, error) {
+	octx := NewContext(ctx, zctx)
+	q, err := c.NewQuery(octx, program, readers)
+	if err != nil {
+		octx.Cancel()
```
nwt, line 50:
```suggestion
	rctx := NewContext(ctx, zctx)
	q, err := c.NewQuery(rctx, program, readers)
	if err != nil {
		rctx.Cancel()
```

runtime/sam/op/merge/merge_test.go
```diff
@@ -96,7 +96,7 @@ func TestParallelOrder(t *testing.T) {
 	for i, c := range cases {
 		t.Run(strconv.Itoa(i), func(t *testing.T) {
 			zctx := zed.NewContext()
-			octx := &op.Context{Context: context.Background(), Zctx: zctx}
+			octx := &runtime.Context{Context: context.Background(), Zctx: zctx}
```
nwt, line 99: This is one of two remaining occurrences of octx. You could rename it to rctx but I favor removing it and replacing `octx.Context` with `context.Background()` down below.
```suggestion
```

cmd/zed/dev/vcache/agg/command.go
```diff
@@ -11,6 +11,7 @@ import (
 	"github.com/brimdata/zed/pkg/charm"
 	"github.com/brimdata/zed/pkg/storage"
 	"github.com/brimdata/zed/runtime/vam"
+	vamop "github.com/brimdata/zed/runtime/vam/op"
```
nwt, line 14: Nit: No need to rename this import since `op` doesn't conflict with any of the others.
```suggestion
	"github.com/brimdata/zed/runtime/vam/op"
```

runtime/vam/op/puller.go
```diff
@@ -0,0 +1,3 @@
+package op
+
+//XXX put in vector?
```
nwt, line 3: Nit: Don't create this file until we have something useful to put in it.

nwt (approved):

PR #5013, opened 2024-01-30
add vector compiler skeleton and basic logic

mccanne:
This commit adds basic comparison logic to the vector runtime and some initial compiler framework to translate vectorized DAGs more generally into runtime ops and exprs.  This is partially complete but we want to merge as is so the team can begin work on vector-enabled field search in parallel.

We also renamed the "zed dev vcache" command to "zed dev vector".

In a subsequent PR, we will finish out the mechanisms for handling heterogeneously typed data with mixed-in errors building up a library of helper functions for Zed variants to do so.  At that point, we will begin adding comprehensive tests for the vector runtime.


runtime/vam/expr/compare.go
```diff
@@ -0,0 +1,960 @@
...
+}
+
+func NewCompare(zctx *zed.Context, lhs, rhs Evaluator, operator string) *Compare {
+	return &Compare{zctx, operator, lhs, rhs}
+}
+
+func (c *Compare) Eval(val vector.Any) vector.Any {
+	l := vector.Under(c.lhs.Eval(val))
+	r := vector.Under(c.rhs.Eval(val))
+	lhs, rhs, _ := coerceVals(c.zctx, l, r)
```
mattnibs, line 25: Forgive the naive comment, but this is going to panic if val is a Variant (since calling Type(), as coerceVals does, on vector.Variant panics). Would the idea in a future update have this recursively call Compare if Variant is encountered here?

mattnibs: Nvm, I suppose, I see the comment `//XXX unions and variants and single-value-with-error variant`.

mattnibs (approved):

cmd/zed/dev/vector/run/command.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/brimdata/zed/runtime/vcache"
+	"github.com/brimdata/zed/zbuf"
+	"github.com/brimdata/zed/zio"
+	"github.com/segmentio/ksuid"
+)
+
+var run = &charm.Spec{
+	Name:  "run",
+	Usage: "run [flags] query path",
+	Short: "run a Zed query on a VNG file",
```
nwt, line 24: Would it make sense to rename this later to "zed dev vector query" for symmetry with "zed query"?

nwt (approved):

cmd/zed/dev/vector/run/command.go
mccanne: Yes!  I'll change it now...

PR #5017, opened 2024-02-01
Add ability to run manage from the serve command

mattnibs:
This commit adds the ability to run zed manage in the same process as the zed serve command. If the -manage flag is set with a duration greater than 0 manage will run.

This commit also move the manage package to lake/api/manage package and simplifies the logic for the monitor function.


cmd/zed/manage/command.go
```diff
@@ -8,8 +8,8 @@ import (
 
 	"github.com/brimdata/zed/cli/lakeflags"
 	"github.com/brimdata/zed/cli/logflags"
-	"github.com/brimdata/zed/cmd/zed/manage/lakemanage"
 	"github.com/brimdata/zed/cmd/zed/root"
+	lakemanage "github.com/brimdata/zed/lake/api/manage"
```
nwt, line 12: I think we should put this in cmd/zed/internal/manage (or maybe cmd/zed/internal/lakemanage).

While I understand the thinking behind putting it here (i.e., that it interacts with lakes only through the lake/api package), everything else in or under the lake package is part of the lake implementation or API. This, on the other hand, is an API consumer. Moreover, it's intended for use only under cmd/zed.

lake/api/manage/config.go
```diff
@@ -19,32 +19,27 @@ func (c *Config) poolConfig(p *pools.Config) PoolConfig {
...
+		Pool:    p.Name,
+		Branch:  "main",
+		Vectors: c.Vectors,
 	}
 }
 
-type PoolConfig struct {
-	Pool     string         `yaml:"pool"`
-	Branch   string         `yaml:"branch"`
-	Interval *time.Duration `yaml:"interval"`
```
nwt, line 41: Why remove this field?

mattnibs: It just seemed like having a per-pool interval added a lot of complexity to the code that might not ultimately get used so I'm simplified things and just put interval on a global level. Do you think this is a mistake?

mattnibs: I imagine the real version of manage (not this toy version we currently have) will not have intervals but rather will compact opportunistically while data comes in. Following this logic it made sense to me to just keep this as simple as possible.

nwt: Makes sense to me.

cmd/zed/manage/command.go
mattnibs: Not opposed to doing this but I think if you're going to import the package into `service` as we do in this pr, its a little weird to have inside of `cmd/zed`.

nwt: But this PR doesn't touch package service.

mattnibs: oh wait you are right. Ok we can try this.

cmd/zed/serve/command.go
```diff
@@ -59,6 +64,7 @@ func New(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
 	})
 	f.StringVar(&c.conf.DefaultResponseFormat, "defaultfmt", service.DefaultZedFormat, "default response format")
 	f.StringVar(&c.listenAddr, "l", ":9867", "[addr]:port to listen on")
+	f.DurationVar(&c.manage, "manage", 0, "manage lake (e.g., compact) every interval (default 0=never)")
```
nwt, line 67: The usage message shows the default twice.
```console
$ go run ./cmd/zed serve -h |& grep manage
    -manage manage lake (e.g., compact) every interval (default 0=never) (default "0s")
```

```suggestion
	f.DurationVar(&c.manage, "manage", 0, "when positive, run lake maintenance tasks at this interval")
```

nwt (commented):
I'd like to have @philrz give this a spin before we merge it.

mattnibs:
> I'd like to have @philrz give this a spin before we merge it.

Agreed.

philrz:
I just tested this branch out at commit 9ce0f4f and it seems to work fine. One thing I found a little odd though is that I went looking for a ztest in this PR since that usually gives me a starting point for seeing how the new functionality is invoked and it looks like the only ztest change is to something that only exercises the pre-existing approach that used the `zed manage` approach. Does the continuous nature of `zed serve -manage` make it too difficult to hit with ztest?

nwt (approved):

philrz:
After an offline discussion with @nwt he agreed to take up the testing angle separately/later via what's now tracked in #5024.

PR #5022, opened 2024-02-06
seekindex.Writer: accept io.WriteCloser

mattnibs:
Follow the pattern we use for our other writers have an seekindex.Writer accept a io.WriteCloser as its argument, constructing the relied-on zngio.Writer under the covers.

### DIFF
--- a/lake/seekindex/writer.go
+++ b/lake/seekindex/writer.go
@@ -17,12 +17,12 @@ type Entry struct {
 
 type Writer struct {
 	marshal *zson.MarshalZNGContext
-	writer  zio.Writer
+	writer  zio.WriteCloser
 	offset  uint64
 	valoff  uint64
 }
 
-func NewWriter(w zio.Writer) *Writer {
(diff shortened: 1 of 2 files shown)


lake/seekindex/writer.go
```diff
@@ -17,15 +20,15 @@ type Entry struct {
...
 	offset  uint64
 	valoff  uint64
 }
 
-func NewWriter(w zio.Writer) *Writer {
+func NewWriter(w io.WriteCloser) *Writer {
 	return &Writer{
 		marshal: zson.NewZNGMarshaler(),
-		writer:  w,
+		writer:  zngio.NewWriter(w),
```
nwt, line 31: Tying this package to ZNG doesn't feel quite right. How about changing the io.WriteCloser parameter to zio.WriteCloser and letting the caller create the ZNG writer?

mattnibs: Ok

nwt (approved):

PR #5023, opened 2024-02-07
Buffer zbuf.Batches in runtime/sam/op/tail.Op

nwt:
The tail operator buffers zed.Values.  Change it to buffer zbuf.Batches, which fits better with planned changes to memory management.


mccanne (approved):
cool

PR #5026, opened 2024-02-12
Fix panics due to bad batch reference counts

nwt:
* runtime/sam/op/groupby.Op.run accesses its batch after unreffing it.

* zbuf.NewBatch does not ref its batch parameter, allowing premature reuse when its vals parameter contains values that live in that batch.  Fix by adding proper reference counting for zbuf.batch.

### DIFF
--- a/zbuf/batch.go
+++ b/zbuf/batch.go
@@ -199,40 +199,39 @@ func (r *pullerReader) Read() (*zed.Value, error) {
 	return val, nil
 }
 
-// XXX at some point the stacked scopes should not make copies of values
-// but merely refer back to the value in the wrapped batch, and we should
-// ref the wrapped batch then downstream entities will unref it, but how
-// do we carry the var frame through... protocol needs to be that any new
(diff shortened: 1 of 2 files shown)


zbuf/batch.go
```diff
@@ -199,40 +199,39 @@ func (r *pullerReader) Read() (*zed.Value, error) {
...
-	Batch
-	vars []zed.Value
+	refs   int32
+	parent Batch
+	vals   []zed.Value
+	vars   []zed.Value
 }
 
 func NewBatch(b Batch, vals []zed.Value) Batch {
+	b.Ref()
```
mattnibs, line 210: Instead of keeping the parent batch around, why not require that the invoker do a fully copy of the values that took from the parent? Seems like this would require keeping around memory that isn't be used in a lot of cases. What am I missing?

nwt: Glad you asked.

At present we almost always do that copy, so holding onto the parent batch is pointless, wasting memory with no reduction in copying, so I'll close this.

PR #5044, opened 2024-02-21
vector search

mattnibs:
Add vector search command that for a lake with VNG files will perform
a vector enhanced search and use the resultant bit vector to trim the
scanned ranges of the accompanying sequence files.

### DIFF
--- a/cmd/zed/dev/vector/search/command.go
+++ b/cmd/zed/dev/vector/search/command.go
@@ -73,7 +73,7 @@ func (c *Command) Run(args []string) error {
 	if err != nil {
 		return err
 	}
-	if err := zio.Copy(writer, zbuf.PullerReader(puller)); err != nil {
+	if err := zio.CopyPuller(writer, puller); err != nil {
 		writer.Close()
 		return err
 	}
--- a/cmd/zed/dev/vector/search/command.go
+++ b/cmd/zed/dev/vector/search/command.go
@@ -15,7 +15,6 @@ import (
 	"github.com/brimdata/zed/pkg/storage"
 	"github.com/brimdata/zed/runtime"
 	"github.com/brimdata/zed/zbuf"
-	"github.com/brimdata/zed/zio"
 )
 
 var search = &charm.Spec{
@@ -73,7 +72,7 @@ func (c *Command) Run(args []string) error {
 	if err != nil {
 		return err
 	}
-	if err := zio.CopyPuller(writer, puller); err != nil {
+	if err := zbuf.CopyPuller(writer, puller); err != nil {
 		writer.Close()
 		return err
 	}
--- a/cmd/zed/dev/vector/search/command.go
+++ b/cmd/zed/dev/vector/search/command.go
@@ -1,0 +1,81 @@
+package search
+
+import (
+	"errors"
+	"flag"
+
+	"github.com/brimdata/zed"
+	"github.com/brimdata/zed/cli/outputflags"
+	"github.com/brimdata/zed/cli/poolflags"
+	devvector "github.com/brimdata/zed/cmd/zed/dev/vector"
+	"github.com/brimdata/zed/cmd/zed/root"
+	"github.com/brimdata/zed/compiler"
+	"github.com/brimdata/zed/compiler/data"
+	"github.com/brimdata/zed/pkg/charm"
+	"github.com/brimdata/zed/pkg/storage"
+	"github.com/brimdata/zed/runtime"
+	"github.com/brimdata/zed/zbuf"
+	"github.com/brimdata/zed/zio"
+)
+
+var search = &charm.Spec{
+	Name:  "search",
+	Usage: "search [flags] filter_expr",
+	Short: "run a VNG optimized search on a lake",
+	New:   newCommand,
+}
+
+func init() {
+	devvector.Cmd.Add(search)
+}
+
+type Command struct {
+	*root.Command
+	outputFlags outputflags.Flags
+	poolFlags   poolflags.Flags
+}
+
+func newCommand(parent charm.Command, f *flag.FlagSet) (charm.Command, error) {
+	c := &Command{Command: parent.(*root.Command)}
+	c.outputFlags.SetFlags(f)
+	c.poolFlags.SetFlags(f)
+	return c, nil
+}
+
+func (c *Command) Run(args []string) error {
+	ctx, cleanup, err := c.Init(&c.outputFlags)
+	if err != nil {
+		return err
+	}
+	defer cleanup()
+	if len(args) != 1 {
+		return errors.New("usage: filter expression")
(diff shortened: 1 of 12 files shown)


lake/data/seekindex.go
```diff
@@ -23,30 +24,55 @@ func LookupSeekRange(ctx context.Context, engine storage.Engine, path *storage.U
...
+func RangeFromBitVector(ctx context.Context, engine storage.Engine, path *storage.URI,
+	o *Object, b *vector.Bool) ([]seekindex.Range, error) {
+	entries, err := seekEntries(ctx, engine, path, o)
+	if err != nil {
+		return nil, err
+	}
+	return entries.Filter(b).Range(), nil
+}
+
+func seekEntries(ctx context.Context, engine storage.Engine, path *storage.URI, o *Object) (seekindex.Entries, error) {
```
mccanne, line 58: call this readSeekIndex?

lake/seekindex/entry.go
```diff
@@ -0,0 +1,53 @@
...
+		Offset: int64(e.Offset),
+		Length: int64(e.Length),
+		ValOff: int64(e.ValOff),
+		ValCnt: int64(e.ValCnt),
+	}
+}
+
+type Entries []Entry
+
+func (entries Entries) Filter(b *vector.Bool) Entries {
```
mccanne, line 28: call this Lookup?  if you changed the name of Entries, then the naming would make more sense, e.g., do a lookup on an Index, e.g., `func (i Index) Lookup(b *vector.Bool) []Range`

mccanne (approved):

cmd/zed/dev/vector/search/command.go
```diff
@@ -0,0 +1,81 @@
...
+	rctx := runtime.NewContext(ctx, zed.NewContext())
+	puller, err := compiler.VectorFilterCompile(rctx, text, data.NewSource(nil, root), head)
+	if err != nil {
+		return err
+	}
+	writer, err := c.outputFlags.Open(ctx, storage.NewLocalEngine())
+	if err != nil {
+		return err
+	}
+	if err := zio.Copy(writer, zbuf.PullerReader(puller)); err != nil {
```
nwt, line 76:
```suggestion
	if err := zio.CopyPuller(writer, puller); err != nil {
```

cmd/zed/dev/vector/ztests/search.yaml
```diff
@@ -0,0 +1,17 @@
+script: |
+  export ZED_LAKE=test
+  zed init -q
+  zed create -q -use test
+  seq 50000 | zq '{ts:int64(this),y:3.14,z:"foo"}' - | zed load -q -
```
nwt, line 5:
```suggestion
  seq -f '{ts:%g,y:3.14,z:"foo"}' 50000 | zed load -q -
```

compiler/kernel/op.go
```diff
@@ -86,6 +87,28 @@ func (b *Builder) BuildWithPuller(seq dag.Seq, parent vector.Puller) ([]vector.P
...
+	if err != nil {
+		return nil, err
+	}
+	cache := b.source.Lake().VectorCache()
+	search, err := vop.NewSearcher(b.rctx, cache, l, pool, e, demand.Fields(d))
+	if err != nil {
+		return nil, err
+	}
+	f := &Filter{pushdown: filter, builder: b}
+	return meta.NewSearchScanner(b.rctx, search, pool, f, b.progress), nil
```
nwt, line 109:
```suggestion
	return meta.NewSearchScanner(b.rctx, search, pool, b.PushdownOf(filter), b.progress), nil
```

runtime/sam/op/meta/sequence.go
```diff
@@ -92,6 +94,62 @@ func (s *SequenceScanner) close(err error) {
...
+			if err != nil {
+				return nil, err
+			}
+		}
+		batch, err := s.scanner.Pull(false)
+		if batch == nil {
+			s.scanner = nil
+			return nil, err
+		}
+		return batch, nil
```
nwt, line 149: This won't always return a non-nil error.
```suggestion
		if batch == nil {
			s.scanner = nil
		}
		return batch, err
```

runtime/sam/op/meta/sequence.go
```diff
@@ -92,6 +94,62 @@ func (s *SequenceScanner) close(err error) {
...
+func (s *SearchScanner) Pull(done bool) (zbuf.Batch, error) {
+	if done {
+		var err error
+		if s.scanner != nil {
+			_, err = s.scanner.Pull(true)
+			s.scanner = nil
+		}
+		return nil, err
+	}
+	for {
```
nwt, line 129: This loop's trip count is always one so why have it?

runtime/vam/op/searcher.go
```diff
@@ -0,0 +1,113 @@
...
+			return
+		}
+		s.sendResult(meta, b, nil)
+	}
+}
+
+func (s *Searcher) sendResult(o *data.Object, b *vector.Bool, err error) {
+	select {
+	case s.resultCh <- searchResult{o, b, err}:
+		return
```
nwt, line 93:
```suggestion
```

runtime/vam/op/searcher.go
```diff
@@ -0,0 +1,113 @@
...
+			return
+		}
+		vec, err := object.Fetch(s.rctx.Zctx, s.projection)
+		if err != nil {
+			s.sendResult(nil, nil, err)
+			return
+		}
+		b, ok := s.filter.Eval(vec).(*vector.Bool)
+		if !ok {
+			s.sendResult(nil, nil, fmt.Errorf("system error: vam.Searcher encountered a non-boolean filter result"))
```
nwt, line 83:
```suggestion
			s.sendResult(nil, nil, errors.New("system error: vam.Searcher encountered a non-boolean filter result"))
```

runtime/vam/op/searcher.go
```diff
@@ -0,0 +1,113 @@
...
+		resultCh:   make(chan searchResult),
+		doneCh:     make(chan struct{}),
+	}, nil
+}
+
+func (s *Searcher) Pull(done bool) (*data.Object, *vector.Bool, error) {
+	s.once.Do(func() {
+		// Block p.ctx's cancel function until p.run finishes its
+		// cleanup.
+		s.rctx.WaitGroup.Add(1)
```
nwt, line 46: run does no cleanup so this is useless.
```suggestion
```

runtime/sam/op/meta/sequence.go
mattnibs: There's an error here because if the scanner gets eof on the next pull is needs to loop back and get the next scanner. Will fix.

nwt (approved):

Issue #5052, opened 2024-02-28
zq "-query" flag mentioned in docs but doesn't exist


philrz:
## tl;dr

The `zq` docs  mention a `-query` flag that doesn't appear to have ever existed. It seems we should either implement it or remove reference to it from the docs.

## Details

At the time this issue is being filed, Zed is at commit 87deeeb.

The text in question is from the [`zq` usage docs](https://zed.brimdata.io/docs/commands/zq#usage) after describing the heuristic approach for determining whether the first argument is a query or an input.

> This heuristic is convenient but can result in a rare surprise when a simple Zed query (like a keyword search) happens to correspond with a file of the same name in the local directory. To avoid this, you can provide the query with the `-query` flag, which specifies the Zed program to run and forces all arguments to be interpreted as inputs.

The text in question was first introduced as part of #3722, but compiling `zq` at the commit associated with that PR doesn't show the flag as existing then and it doesn't appear to have been introduced since. But it does seem like its existence could be justified since the "surprise" still seems to exist as described. For instance, starting with this directory and single input file:

```
$ ls -l
total 8
-rw-r--r--  1 phil  staff  29 Feb 27 16:27 input

$ cat input 
{"foo": "bar"}
{"baz": "in"}

$ zq -version
Version: v1.14.0-6-g87deeebe

$ zq in input 
{baz:"in"}

$ echo '{"hello": "world"}' > in

$ zq in input 
{hello:"world"}
{foo:"bar"}
{baz:"in"}
```

Per the docs, to disambiguate that last example it seems like `zq -query in input` would be a reasonable way to return just the `{baz:"in"}` result despite the existence of the file named `in`.

The wider context in which I happened to be revisiting these docs is brainstorming with @mattnibs on improvements related to #5046. I was reviewing the docs for possible improvements we could make and came across this `-query` option as something that we might be able to recommend in `zq` error outputs, only to find it doesn't exist.

PR #5053, opened 2024-02-28
zq: Better error messages for single argument commands

### LINKED ISSUE #5046: Zed syntax error causes in confusing error message about file input
## tl;dr

A community user was confused by this error message:

```
$ zq -version
Version: v1.14.0-2-g4dc62369

$ zq -Z 'fork(=> file "./csv_1.csv" | yield email | collect(this), => file "./csv_2.csv" | yield this["Recommended Email"])'
zq: no such file: fork(=> file "./csv_1.csv" | yield email | collect(this), => file "./csv_2.csv" | yield this["Recommended Email"])
```

## Details

Repro is with Zed commit 4dc6236. The issue was initially reported in a [community Slack thread](https://brimdata.slack.com/archives/CTT24051C/p1708106640649629).

Many `zq` command lines are unambiguous when the first argument is the Zed program text and the remaining arguments are file inputs. The example above presents a challenge because there's just a single argument, so `zq` needs to use some kind of heuristic to determine how to treat that single argument. It could be a filename (e.g., a simple command line like `zq MyFile.zng` may often be used to view a file's contents as ZSON) or, as was clearly the user's intent here, it could be a Zed program that specifies the inputs from inside using the `file` operator. In this case there was a syntax error in the Zed program (the `,` before the `=>`) so the heuristic fell back to treating the single input as a filename, and the current error message about this understandably confused the user.

We discussed this one as a group and there was some thought that the heuristic could maybe stand to be revisited and/or a new flag could be introduced similar to `jq -n`. However, as a short term improvement there was some consensus that we might get by just improving the error message here to be explicit about what the heuristic attempted to make it more obvious to the user in a situation like this that it's actually a Zed syntax error. Most of the error messages from the Zed CLI tooling are fairly terse, but this situation is tricky enough that it may be one where some verbosity is justified.

mattnibs:
Fixes #5046

### DIFF
--- a/cmd/zq/ztests/single-arg-error.yaml
+++ b/cmd/zq/ztests/single-arg-error.yaml
@@ -1,11 +1,9 @@
 script: |
-  ! zq -z doesnotexist
+  ! zq 'file sample.zson | count('
 
 outputs:
-  - name: stdout
-    data: ""
   - name: stderr
     data: |
       zq: could not invoke zq with a single argument because:
        - the argument did not parse as a valid Zed query
-       - a file could not be found with the name "doesnotexist"
+       - a file could not be found with the name "file sample.zson | c..."
--- a/cmd/zq/ztests/single-arg-error.yaml
+++ b/cmd/zq/ztests/single-arg-error.yaml
@@ -5,5 +5,7 @@ outputs:
   - name: stderr
     data: |
       zq: could not invoke zq with a single argument because:
-       - the argument did not parse as a valid Zed query
        - a file could not be found with the name "file sample.zson | c..."
+       - the argument could not be compiled as a valid Zed query due to parse error (column 25):
+         file sample.zson | count(
+                              === ^ ===
--- a/cli/queryflags/flags.go
+++ b/cli/queryflags/flags.go
@@ -14,6 +14,7 @@ import (
 	"github.com/brimdata/zed/compiler"
 	"github.com/brimdata/zed/compiler/ast"
 	"github.com/brimdata/zed/compiler/data"
+	"github.com/brimdata/zed/compiler/parser"
 	"github.com/brimdata/zed/compiler/semantic"
 	"github.com/brimdata/zed/pkg/storage"
 	"github.com/brimdata/zed/zbuf"
@@ -40,7 +41,8 @@ func (f *Flags) ParseSourcesAndInputs(paths []string) ([]string, ast.Seq, bool, 
 			// Consider a lone argument to be a query if it compiles
 			// and appears to start with a from or yield operator.
 			// Otherwise, consider it a file.
-			if query, err := compiler.Parse(src, f.Includes...); err == nil {
+			query, err := compiler.Parse(src, f.Includes...)
+			if err == nil {
 				if s, err := semantic.Analyze(context.Background(), query, data.NewSource(storage.NewLocalEngine(), nil), nil); err == nil {
 					if semantic.HasSource(s) {
 						return nil, query, false, nil
@@ -50,7 +52,7 @@ func (f *Flags) ParseSourcesAndInputs(paths []string) ([]string, ast.Seq, bool, 
 					}
 				}
 			}
-			return nil, nil, false, singleArgError(src)
+			return nil, nil, false, singleArgError(src, err)
 		}
 	}
 	query, err := compiler.Parse(src, f.Includes...)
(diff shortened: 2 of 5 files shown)


philrz:
I just tested this branch out at commit f361f112 using the repro example from #5046.

```
$ zq -version
Version: v1.14.0-7-gf361f112

$ zq -Z 'fork(=> file "./csv_1.csv" | yield email | collect(this), => file "./csv_2.csv" | yield this["Recommended Email"])'
 - the argument did not parse as a valid Zed query
 - a file could not be found with the name "fork(=> file \"./csv_..."
```

This seems like a great improvement to me so I'm functionally 👍 on this.

As @mattnibs and I have discussed offline, the user would surely appreciate being able to see the parse error in a situation like this. Two thoughts come to mind:

1. We could present it as part of this two-part message. If we do, I might suggest swapping the order since I expect the parse error will take up multiple lines.
2. We could implement the `-query` flag as mention in the docs and #5052 and make some reference to it in this error message (e.g., "consider `-query`").

philrz:
I just re-tested the branch now at commit 2699090 once again using that example from #5046. It looks great with the parse error!

```
$ zq -version
Version: v1.14.0-8-g2699090d

$ zq -Z 'fork(=> file "./csv_1.csv" | yield email | collect(this), => file "./csv_2.csv" | yield this["Recommended Email"])'
 - a file could not be found with the name "fork(=> file \"./csv_..."
 - the argument could not be compiled as a valid Zed query due to parse error (column 58):
   fork(=> file "./csv_1.csv" | yield email | collect(this), => file "./csv_2.csv" | yield this["Recommended Email"])
                                                         === ^ ===
```

I then thought to see how it would look if I had some multi-line Zed.

```
$ zq -Z 'fork (
>   => file "./csv_1.csv" | yield email | collect(this),
>   => file "./csv_2.csv" | yield this["Recommended Email"]
> )'
 - a file could not be found with the name "fork (\n  => file \"./..."
 - the argument could not be compiled as a valid Zed query due to parse error (line 2, column 2):
     => file "./csv_2.csv" | yield this["Recommended Email"]
   = ^ ===
```

That's great since it mentions the line & column but shows just the line that had the error. Very nice! 👍 

(I guess it's a partial bummer that the parser doesn't point directly at the `,` as being the problem, but that doesn't appear to be this PR's fault.)

cmd/zq/ztests/single-arg-error.yaml
```diff
@@ -0,0 +1,11 @@
...
+  ! zq 'file sample.zson | count('
+
+outputs:
+  - name: stderr
+    data: |
+      zq: could not invoke zq with a single argument because:
+       - a file could not be found with the name "file sample.zson | c..."
+       - the argument could not be compiled as a valid Zed query due to parse error (column 25):
+         file sample.zson | count(
+                              === ^ ===
```
nwt, line 11: This is pretty wordy. How about something more concise?
```
zq: not a file or valid query: "from ( file a ) | co..."
  error parsing Zed at column 25:
  from ( file a ) | count(
                      === ^ ===
```

As a bonus, that would make for a much simpler `singleArgError`.
```go
func singleArgError(src string, err error) error {
	if len(src) > 20 {
		src = src[:20] + "..."
	}
	msg := fmt.Sprintf("not a file or valid query: %q", src)
	if errors.As(err, new(*parser.Error)) {
		msg += "\n" + err.Error()
	}
	msg = strings.ReplaceAll(msg, "\n", "\n   ")
	return errors.New(msg)
}
```

nwt: (To be clear, I'm OK with the existing message if you guys prefer it.)

philrz: Thanks @nwt. Indeed, I was in the middle of writing a reply when that last comment of yours came in, so I'll paste that below for completeness, but basically, yeah, I happen to prefer what's already there. 😄

I know I can tend toward wordiness in my own work and so I try to keep that in check, but I think this is one place where verbosity is maybe justified. For instance, in the #5046 case the user is almost certainly thinking entirely in terms of their argument being a Zed query so I'm concerned something concise like "not a file" might still be confusing. "Not a file? What? The file refs are in my program. I don't get it." This particular error is triggered rarely enough and has confused users enough in is previous form in enough that I feel pretty strongly that the benefits of super clear communication outweigh any assumed benefits from being concise.

Come to think of it, the user that last hit #5046 is responsive on Slack today, so I'll see if they're game to weigh in to the discussion, though if we already have consensus and want to get on with it and merge I'm fine with that too.

nwt: @mattnibs: Any preference here?

nwt (approved):

cmd/zq/ztests/single-arg-error.yaml
iloveitaly: fwiw, as someone who doesn't use zed every day, but reaches for it probably once a week when manipulating data I'd prefer the more verbose message. It's hard to understand exactly what might be wrong, especially given the unfamiliarity of the zq syntax compared to other tools.

Issue #4537, opened 2023-04-24
Inconsistent errors when reading NDJSON with bad last line


philrz:
Repro is with Zed commit d599839.

The attached NDJSON test data files [lines-9.ndjson.gz](https://github.com/brimdata/zed/files/11314643/lines-9.ndjson.gz) and [lines-10.ndjson.gz](https://github.com/brimdata/zed/files/11314644/lines-10.ndjson.gz) both consist of several lines of valid NDJSON and a closing incomplete line:

```
{"syntaxerror
```

They otherwise only differ in that `lines-10` contains one more valid NDJSON record than `lines-9` before that bad last line.

Reading both with `zq`, the reported errors differ.

```
$ zq -version
Version: v1.7.0-35-gd5998393

$ zq -z lines-9.ndjson
lines-9.ndjson: parse error: string literal: unescaped line break

$ zq -z lines-10.ndjson
lines-10.ndjson: EOF
```

The difference becomes a little more significant when loading to a pool, since no error at all is reported for `line-10`.

```
$ zed -use foo load lines-9.ndjson
(1/1) 3724B/3724B 3724B/s 100.00%
Post "http://localhost:9867/pool/2OtB1htK17ZmmjRhXGdwCbXpoPp/branch/main": parse error: string literal: unescaped line break

$ zed -use foo load lines-10.ndjson
(1/1) 4134B/4134B 4134B/s 100.00%
2OtBn63BBXfmUMaoGYmM1nQSCrp committed

$ echo $?
0
```

philrz:
The fixes in #5055 have significantly improved the errors shown here. Repeating the original repro steps with Zed commit 38763f8, we now see:

```
$ zq -version
Version: v1.14.0-16-g38763f82

$ zq -z lines-9.ndjson
lines-9.ndjson: parse error: string literal: unescaped line break

$ zq -z lines-10.ndjson
lines-10.ndjson: unexpected end of JSON input
```

@mattnibs explains in https://github.com/brimdata/zed/pull/5055#issuecomment-1971566624 why we saw the improvement here for `lines-10.ndjson` but not `lines-9.ndjson`.

> this is a separate issue since in the example of `lines-9.ndjson` zq is choosing the zsonio reader which is where the error is coming from. [...] if I run it with the json reader specified [...] I get the expected error message.

Indeed this is the case.

```
$ zq -i json lines-9.ndjson
lines-9.ndjson: unexpected end of JSON input
```

The improvements are similar for `zed load`.

[...]

PR #5055, opened 2024-02-29
jsonio: Fix parse errors on EOF

mattnibs:
Fix an issue with jsonio.Reader where parse errors occurring at the end of an input (EOF) are often ignored or return an ambiguous EOF error.

### DIFF
--- a/pkg/jsonlexer/lexer.go
+++ b/pkg/jsonlexer/lexer.go
@@ -93,6 +93,9 @@ func (l *Lexer) readLiteral(s string, t Token) Token {
 	for i := range s {
 		c, err := l.br.ReadByte()
 		if err != nil {
+			if err == io.EOF {
+				err = io.ErrUnexpectedEOF
+			}
 			l.err = err
 			return TokenErr
 		}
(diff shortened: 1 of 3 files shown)


philrz:
@mattnibs: The description of the fix reminded me of a familiar bug, so I dug up #4537. Indeed, it seems to improve the error message in one of the cases described there. Whereas it looks like this on tip of main:

```
$ zq -version
Version: v1.14.0-6-g87deeebe

$ zq -z lines-10.ndjson
lines-10.ndjson: EOF
```

On your branch at commit 1d3d2c3 it's now improved to:

```
$ zq -version
Version: v1.14.0-7-g1d3d2c37

$ zq -z lines-10.ndjson
lines-10.ndjson: unexpected end of JSON input
```

FWIW, the other case shown there isn't affected.

```
$ zq -version
Version: v1.14.0-6-g87deeebe

$ zq -z lines-9.ndjson
lines-9.ndjson: parse error: string literal: unescaped line break

$ zq -version
Version: v1.14.0-7-g1d3d2c37

$ zq -z lines-9.ndjson
lines-9.ndjson: parse error: string literal: unescaped line break
```

Since you've just spent time under the hood looking at this code, maybe you'd like to make an assessment of whether `unescaped line break` is the best we're ever going to do for that case or if there's an opportunity to improve on that as well, such that we could declare victory on that whole issue.

nwt (approved):

PR #5056, opened 2024-02-29
docs: uppercase ex with substring

iloveitaly:
from: https://brimdata.slack.com/archives/CTT24051C/p1709235005584779


philrz (approved):
Looks great! I made one small adjustment to add a link, but if you're ok accepting that I'll go ahead and merge right after.

PR #5069, opened 2024-03-13
Reset stateful expressions on EOS

### LINKED ISSUE #4943: count() in lateral query not resetting to zero
## tl;dr

@mccanne's statement upon seeing the repro below:

> That looks like a bug!  We're not resetting the streaming agg state at end-of-stream.

## Details

Repro is with Zed commit 59b72d8. I bumped into this while working on a solution for the question posed in #4937.

Compare these two behaviors.

```
$ zq -version
Version: v1.11.1-27-g59b72d88

$ echo '["a","b","c","d","e","f","g"] ["h","i","j","k"]' | zq -z 'over this => (count())' -
7(uint64)
4(uint64)

$ echo '["a","b","c","d","e","f","g"] ["h","i","j","k"]' | zq -z 'over this => ({id:count()})' -
{id:1(uint64)}
{id:2(uint64)}
{id:3(uint64)}
{id:4(uint64)}
{id:5(uint64)}
{id:6(uint64)}
{id:7(uint64)}
{id:8(uint64)}
{id:9(uint64)}
{id:10(uint64)}
{id:11(uint64)}
```

The [lateral subquery docs](https://zed.brimdata.io/docs/language/lateral-subqueries) speak of how the subquery "is evaluated once per outer value" (i.e., per array in this case) so the top behavior made sense to me while the bottom one didn't. I'd expected the `count()` result to reset to zero when processing began on the second array value.

mattnibs:
This commit fixes a bug where stateful expressions were not getting reset when an op encounters EOS. This would most notably result in unexpected values when using stateful expressions inside of lateral queries.

Closes #4943

### DIFF
--- a/compiler/kernel/expr.go
+++ b/compiler/kernel/expr.go
@@ -96,7 +96,9 @@ func (b *Builder) compileExpr(e dag.Expr) (expr.Evaluator, error) {
 		if err != nil {
 			return nil, err
 		}
-		return expr.NewAggregatorExpr(b.zctx(), agg), nil
+		aggexpr := expr.NewAggregatorExpr(b.zctx(), agg)
+		b.resetters = append(b.resetters, aggexpr)
+		return aggexpr, nil
 	case *dag.OverExpr:
 		return b.compileOverExpr(e)
 	default:
@@ -497,7 +499,7 @@ func (b *Builder) compileOverExpr(over *dag.OverExpr) (expr.Evaluator, error) {
 		return nil, err
 	}
 	parent := traverse.NewExpr(b.rctx.Context, b.zctx())
-	enter := traverse.NewOver(b.rctx, parent, exprs)
+	enter := traverse.NewOver(b.rctx, parent, exprs, expr.Resetters{})
 	scope := enter.AddScope(b.rctx.Context, names, lets)
 	exits, err := b.compileSeq(over.Body, []zbuf.Puller{scope})
 	if err != nil {
--- a/compiler/kernel/op.go
+++ b/compiler/kernel/op.go
@@ -50,14 +50,15 @@ import (
 var ErrJoinParents = errors.New("join requires two upstream parallel query paths")
 
 type Builder struct {
-	rctx     *runtime.Context
-	mctx     *zed.Context
-	source   *data.Source
-	readers  []zio.Reader
-	progress *zbuf.Progress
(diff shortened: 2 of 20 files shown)


compiler/kernel/expr.go
```diff
@@ -46,7 +59,7 @@ import (
 // TBD: string values and net.IP address do not need to be copied because they
 // are allocated by go libraries and temporary buffers are not used.  This will
 // change down the road when we implement no-allocation string and IP conversion.
-func (b *Builder) compileExpr(e dag.Expr) (expr.Evaluator, error) {
+func (b *Builder) compileExpr(ectx *exprContext, e dag.Expr) (expr.Evaluator, error) {
```
nwt, line 62: I think we talked about putting this inside Builder as another possible approach, and I think I prefer that now that we can see how much noise we're creating by plumbing it through as a method parameter.

runtime/sam/op/router.go
```diff
@@ -93,6 +94,11 @@ func (r *Router) blocked() bool {
 // after receiving the EOS, it's done will be captured as soon as we unblock
 // all channels.
 func (r *Router) sendEOS(err error) bool {
+	defer func() {
+		if r, ok := r.selector.(expr.Resetter); ok {
+			r.Reset()
+		}
+	}()
```
nwt, line 101: How about embedding expr.Resetter in op.Selector? That feels more straightforward to me.

compiler/kernel/op.go
```diff
@@ -50,14 +50,15 @@ import (
...
-	funcs    map[string]expr.Function
+	rctx      *runtime.Context
+	mctx      *zed.Context
+	source    *data.Source
+	readers   []zio.Reader
+	progress  *zbuf.Progress
+	arena     *zed.Arena // For zed.Values created during compilation.
+	deletes   *sync.Map
+	funcs     map[string]expr.Function
+	resetters expr.Resetters
```
nwt, line 61: I think we might need an expr.Resetters stack here to handle stateful expressions nested at different levels correctly but this looks good for now.

nwt (approved):
I think we might need a stack

compiler/kernel/op.go
mattnibs: I think the one area for concern here is with over expressions. Everywhere else we just collect resetters for the expressions before recursing down into the flowgraph.

PR #5081, opened 2024-03-18
nameof: Work with type values

### LINKED ISSUE #5067: nameof() should support type values
## tl;dr

The following should return `"port"`.

```
$ echo '80(port=int16)' | zq -z 'yield nameof(<port>)' -
error("missing")
```

## Details

Repro is with Zed commit 38763f8.

A community user recently inquired in a [Slack thread](https://brimdata.slack.com/archives/CTT24051C/p1710202640349919):

> I am using typeof(this) to get the types of records in my JSON file. Would it be be possible to return just a simplified version of the type (i.e. "http", "files") rather than the entire type itself?
>
> image

Based on the screenshot, it appears the user is following the Zed docs for [shaping Zeek NDJSON](https://zed.brimdata.io/docs/integrations/zeek/shaping-zeek-ndjson). Starting from here, there's a couple ways to get the Zeek record type from the record _values_, e.g., `yield _path` (since with Zeek data the `_path` field happens to map 1-to-1 with the record types), or [`nameof(this)`](https://zed.brimdata.io/docs/language/functions/nameof), such as in this simplified example based on the Zeek shaping docs:

```
$ cat shaper.zed 
type port=uint16
type conn_id={orig_h:ip,orig_p:port,resp_h:ip,resp_p:port}
type zenum=string
type dns={_path:string,ts:time,uid:string,id:conn_id,proto:zenum,trans_id:uint64,rtt:duration,query:string,qclass:uint64,qclass_name:string,qtype:uint64,qtype_name:string,rcode:uint64,rcode_name:string,AA:bool,TC:bool,RD:bool,RA:bool,Z:uint64,answers:[string],TTLs:[duration],rejected:bool,_write_ts:time}
shape(this, <dns>)

$ zq -version
Version: v1.14.0-16-g38763f82

$ echo {} | zq -I shaper.zed '| yield nameof(this)' -
"dns"
```

However, if we start from the user's position of wanting to derive this info from the `type` value that returrned from the [`typeof` function](https://zed.brimdata.io/docs/language/functions/typeof), that's not currently possible via `nameof`.

```
$ echo {} | zq -I shaper.zed '| yield typeof(this)' -
<dns={_path:string,ts:time,uid:string,id:conn_id={orig_h:ip,orig_p:port=uint16,resp_h:ip,resp_p:port},proto:zenum=string,trans_id:uint64,rtt:duration,query:string,qclass:uint64,qclass_name:string,qtype:uint64,qtype_name:string,rcode:uint64,rcode_name:string,AA:bool,TC:bool,RD:bool,RA:bool,Z:uint64,answers:[string],TTLs:[duration],rejected:bool,_write_ts:time}>

$ echo {} | zq -I shaper.zed '| yield nameof(typeof(this))' -
error("missing")
```

We discussed this one as a group and there was consensus that it would be good to add support for this. In the meantime, in addition to the workarounds shown above that derive the type name from the original value, I also offered the user this hacky workaround that extracts the name from the string representation of the record type definition.

```
$ echo {} | zq -I shaper.zed '| yield split(string(typeof(this)), "=")[0][1:]' -
"dns"
```

mattnibs:
Closes #5067

### DIFF
--- a/runtime/sam/expr/function/types.go
+++ b/runtime/sam/expr/function/types.go
@@ -31,14 +31,17 @@ type NameOf struct {
 
 func (n *NameOf) Call(_ zed.Allocator, args []zed.Value) zed.Value {
 	typ := args[0].Type()
+	if named, ok := typ.(*zed.TypeNamed); ok {
+		return zed.NewString(named.Name)
+	}
 	if typ.ID() == zed.IDType {
 		var err error
 		if typ, err = n.zctx.LookupByValue(args[0].Bytes()); err != nil {
 			panic(err)
 		}
-	}
-	if named, ok := typ.(*zed.TypeNamed); ok {
-		return zed.NewString(named.Name)
+		if named, ok := typ.(*zed.TypeNamed); ok {
+			return zed.NewString(named.Name)
+		}
 	}
 	return n.zctx.Missing()
 }
--- a/runtime/sam/expr/function/types.go
+++ b/runtime/sam/expr/function/types.go
@@ -31,6 +31,12 @@ type NameOf struct {
 
 func (n *NameOf) Call(_ zed.Allocator, args []zed.Value) zed.Value {
 	typ := args[0].Type()
+	if typ.ID() == zed.IDType {
+		var err error
+		if typ, err = n.zctx.LookupByValue(args[0].Bytes()); err != nil {
+			panic(err)
+		}
+	}
 	if named, ok := typ.(*zed.TypeNamed); ok {
 		return zed.NewString(named.Name)
 	}
--- a/runtime/sam/expr/function/ztests/nameof.yaml
+++ b/runtime/sam/expr/function/ztests/nameof.yaml
@@ -3,11 +3,13 @@ zed: count() by nameof(this) | sort this
 input: |
   {x:1}
   {x:1}(=foo)
+  <foo={x:int64}>
   {y:1}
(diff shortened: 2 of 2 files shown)


runtime/sam/expr/function/types.go
```diff
@@ -31,6 +31,12 @@ type NameOf struct {
...
 	typ := args[0].Type()
+	if typ.ID() == zed.IDType {
+		var err error
+		if typ, err = n.zctx.LookupByValue(args[0].Bytes()); err != nil {
+			panic(err)
+		}
+	}
 	if named, ok := typ.(*zed.TypeNamed); ok {
 		return zed.NewString(named.Name)
 	}
```
nwt, line 42: This changes the output of
```
echo '<int64>(=a) <b=int64> <c=int64>(=d)' | zq 'yield nameof(this)' -
```
from
```
"a"
error("missing")
"d"
```
to
```
error("missing")
"b"
"c"
```
but I think we ultimately want
```
"a"
"b"
"d"
```

More explicitly, I propose that the `nameof()` algorithm be this:
1. If the argument is a named type, return its name.
2. Otherwise, if the argument is a type value that encodes a named type, return its name.
3. Otherwise, return `error("missing")`.

mattnibs: Good point didn't consider the case of a named type with the value of type, dawg.

nwt (approved):

PR #5088, opened 2024-03-26
vector search: Fix fields selector

mattnibs:
Fix an issue with vector search where all fields where getting selected for the vector search portion of every search. Use optimizer.FieldsOf instead of the demand package.

### DIFF
--- a/compiler/kernel/op.go
+++ b/compiler/kernel/op.go
@@ -11,6 +11,7 @@ import (
 	"github.com/brimdata/zed"
 	"github.com/brimdata/zed/compiler/ast/dag"
 	"github.com/brimdata/zed/compiler/data"
+	"github.com/brimdata/zed/compiler/optimizer"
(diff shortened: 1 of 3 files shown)


compiler/kernel/op.go
```diff
@@ -101,7 +101,7 @@ func (b *Builder) BuildVamToSeqFilter(filter dag.Expr, d demand.Demand, poolID,
 		return nil, err
 	}
 	cache := b.source.Lake().VectorCache()
-	search, err := vop.NewSearcher(b.rctx, cache, l, pool, e, demand.Fields(d))
+	search, err := vop.NewSearcher(b.rctx, cache, l, pool, e, fields)
```
nwt, line 104: Nit: Move the FieldsOf call into this method. 
```suggestion
	project, _ := optimzer.FieldsOf(filter)
	search, err := vop.NewSearcher(b.rctx, cache, l, pool, e, project)
```

nwt (approved):

PR #5086, opened 2024-03-21
Rework package sam/expr/coerce

mattnibs:
Rework the coerce package so that it uses native values instead of their byte representation.

Fixes #4719

PR #5091, opened 2024-03-26
Add a ztest for math between all numeric types

philrz:
As discussed in #5086, this PR proposes a ztest that performs simple math between all numeric types. https://github.com/brimdata/zed/pull/5086#issuecomment-2021216766 shows an example of how drastically the output would differ if a bug like #4719 crept back in.


nwt (approved):
This duplicates some of TestArithmetic in runtime/sam/epxr/expr_test.go and I think it'd make sense to consolidate the two, but we can do that later.

PR #5096, opened 2024-04-01
JSON pretty print

### LINKED ISSUE #5064: Formatted ("pretty print") JSON output
The `-z` gives non-formatted output. For "pretty output", there is `-Z`

```
zq -z '{"foo":1, "bar":2}'
{foo:1,bar:2}

zq -Z '{"foo":1, "bar":2}'
{
    foo: 1,
    bar: 2
}
```

However, there is no JSON equivalent for `-Z`
```
zq -j '{"foo":1, "bar":2}'
{"foo":1,"bar":2}

zq -J '{"foo":1, "bar":2}'
flag provided but not defined: -J
```

For pretty output, i have to opt to secondary tool
```
zq -j '{"foo":1, "bar":2}' | jq
{
  "foo": 1,
  "bar": 2
}
```

Is there a way to get JSON formatted output without reaching to a another tool ?

mattnibs:
Add support for JSON pretty printing (with color if supported by terminal) when activated by either using the -J flag or setting format and the -pretty flag.

Closes #5064


philrz:
I just tried it out on the branch at commit edd77e6 and it looks good to me!

docs/tutorials/zq.md
```diff
@@ -1170,7 +1170,7 @@ which produces
 Of course, if you'd like the query output in JSON, you can just say `-f json` and
```
nwt, line 1170:
```suggestion
Of course, if you'd like the query output in JSON, you can just say `-j` and
```

go.mod
```diff
@@ -32,8 +32,8 @@ require (
 	go.uber.org/zap v1.23.0
 	golang.org/x/exp v0.0.0-20231006140011-7918f672742d
 	golang.org/x/sync v0.4.0
-	golang.org/x/sys v0.13.0
-	golang.org/x/term v0.13.0
+	golang.org/x/sys v0.14.0
+	golang.org/x/term v0.14.0
```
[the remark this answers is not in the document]
mattnibs: No, I will back it out.

zio/jsonio/writer.go
```diff
@@ -1,26 +1,27 @@
...
-	"encoding/json"
 	"io"
 
 	"github.com/brimdata/zed"
 )
 
 type Writer struct {
 	io.Closer
-	encoder *json.Encoder
+	encoder *encoder
```
nwt, line 11: Is there a benefit to having the separate `encoder` type rather than just adding its fields and methods to `Writer`?

zio/jsonio/encode.go
```diff
@@ -0,0 +1,245 @@
...
+func (e *encoder) writeColor(b []byte, code []byte) {
+	if e.tab > 0 && color.Enabled {
+		e.buf.Write(code)
+		defer e.buf.WriteString(color.Reset.String())
+	}
+	e.buf.Write(b)
+}
+
+func (e *encoder) newline() {
+	e.buf.WriteString(e.newlineChar)
```
nwt, line 240: Nit: No need for newlineChar if you do this.
```suggestion
	if e.tab > 0 {
		e.buf.WriteByte(`\n`)
	}
```

zio/jsonio/encode.go
```diff
@@ -0,0 +1,245 @@
...
+	fieldColor  = []byte("\x1b[34;1m")
+	nullColor   = []byte("\x1b[2m")
+	numberColor = []byte("\x1b[36m")
+	puncColor   = []byte{} // no color
+	stringColor = []byte("\x1b[32m")
+)
+
+type encoder struct {
+	w           io.Writer
+	buf         bytes.Buffer
```
nwt, line 29: A bufio.Writer would be simpler.

zio/jsonio/writer.go
mattnibs: Yeah we can do this.

zio/jsonio/encode.go
mattnibs: I believe you would then have to check the error on every write in that case because it might flush to io.Writer and you might get an error there.

mattnibs: Looking at the source code it looks like it does cache the first error encountered so perhaps we can get away with using bufio.Writer.

zio/jsonio/writer.go
```diff
@@ -1,26 +1,250 @@
...
 }
 
 func (w *Writer) Write(val zed.Value) error {
-	return w.encoder.Encode(marshalAny(val.Type(), val.Bytes()))
+	w.marshal(0, val)
+	w.writer.WriteByte('\n')
+	return w.writer.Flush()
+}
+
+func (w *Writer) marshal(tab int, val zed.Value) {
```
nwt, line 61: Nit: This no longer feels like the right name so I'd do s/marshal/write/g in this file.

zio/jsonio/writer.go
```diff
@@ -1,26 +1,250 @@
...
+	w.primEnc = json.NewEncoder(&w.primBuf)
+	w.primEnc.SetEscapeHTML(false)
+	return w
 }
 
 func (w *Writer) Write(val zed.Value) error {
-	return w.encoder.Encode(marshalAny(val.Type(), val.Bytes()))
+	w.marshal(0, val)
+	w.writer.WriteByte('\n')
+	return w.writer.Flush()
```
nwt, line 58: Nit: Instead of calling Flush for every zed.Value, call it once from Writer.Close.

zio/jsonio/writer.go
```diff
@@ -1,26 +1,250 @@
...
 
 type Writer struct {
 	io.Closer
-	encoder *json.Encoder
+	writer *bufio.Writer
+	tab    int
+
+	// Use json.Encoder to get marshal primitive Values. Have to use
+	// json.Encoder instead of json.Marshal because this is the only way to get
+	// pkg json to turn off HTML escaping.
```
nwt, line 35:
```suggestion
	// Use json.Encoder for primitive Values. Have to use
	// json.Encoder instead of json.Marshal because it's
	// the only way to turn off HTML escaping.
```

zio/jsonio/writer.go
```diff
@@ -1,26 +1,250 @@
...
+		tab:    opts.Pretty,
 	}
+	w.primEnc = json.NewEncoder(&w.primBuf)
+	w.primEnc.SetEscapeHTML(false)
+	return w
 }
 
 func (w *Writer) Write(val zed.Value) error {
-	return w.encoder.Encode(marshalAny(val.Type(), val.Bytes()))
+	w.marshal(0, val)
```
nwt, line 56: Maybe add a comment here about why this doesn't return an error?

zio/jsonio/writer.go
```diff
@@ -1,26 +1,250 @@
...
+		w.marshalArray(tab, typ.Type, val.Bytes())
+	case *zed.TypeMap:
+		w.marshalMap(tab, typ, val.Bytes())
+	case *zed.TypeEnum:
+		w.marshalEnum(typ, val.Bytes())
+	case *zed.TypeError:
+		w.marshalError(tab, typ, val.Bytes())
+	default:
+		m := fmt.Sprintf("<unsupported type: %s>", zson.FormatType(typ))
+		w.writeColor(w.marshalJSON(m), stringColor)
```
nwt, line 86: Nit: I think we should panic here.

zio/jsonio/writer.go
```diff
@@ -1,26 +1,250 @@
...
+	case id == zed.IDString:
+		v = val.AsString()
+	case id == zed.IDIP:
+		v = zed.DecodeIP(val.Bytes()).String()
+	case id == zed.IDNet:
+		v = zed.DecodeNet(val.Bytes()).String()
+	case id == zed.IDType:
+		v = zson.FormatValue(val)
+	default:
+		v = fmt.Sprintf("<unsupported id=%d>", id)
```
nwt, line 217: Nit: I think we should panic here.

zio/jsonio/writer.go
mattnibs: I don't think this is a good idea since if you have a broken pipe or something like that you won't know it till EOS.

mattnibs: We could do `return w.writer.WriteByte` and do the flush() in close but I do worry about holding back values on slow emitting streams and if you flush() on every value you let the underlying writer figure out what it wants to do (e.g., nagles).

nwt: > Instead of calling Flush for every zed.Value, call it once from Writer.Close.

This is what we do in csvio.Writer, by the way.

mattnibs: For zson we write every value. IDK, for csv I care less about the slow emitting stream problem than I do for json or zson.

nwt (approved):

PR #5099, opened 2024-04-04
Remove javascript parser

mattnibs:
This commit removes the javascript parsers, a feature this is no longer necessary now that the Zui client uses the /compile endpoint to parse Zed queries.


nwt (commented):
We don't need `/node_modules/` in `.gitignore` any more so let's remove that.

nwt (approved):
:shipit:!

PR #5100, opened 2024-04-05
ast: IndexExpr and SliceExpr

mattnibs:
This commit makes index and slice expressions their own ast Nodes instead of representing them as BinaryExpr.


compiler/ast/ast.go
```diff
@@ -83,6 +83,19 @@ type Cast struct {
...
+	Kind  string `json:"kind" unpack:""`
+	Expr  Expr   `json:"expr"`
+	Index Expr   `json:"index"`
+}
+
+type SliceExpr struct {
+	Kind string `json:"kind" unpack:""`
+	Expr Expr   `json:"expr"`
+	Low  Expr   `json:"low"`
+	High Expr   `json:"high"`
```
nwt, line 96: runtime/sam/expr.Slice and compiler/kernel.Builder.compileSlice (and maybe other functions) call these "from" and "to". I don't have any preference between those names (though I do have my usual, mild preference for keeping the diff small) but let's stay consistent.

compiler/kernel/expr.go
```diff
@@ -73,6 +73,10 @@ func (b *Builder) compileExpr(e dag.Expr) (expr.Evaluator, error) {
 		return b.compileConditional(*e)
 	case *dag.Call:
 		return b.compileCall(*e)
+	case *dag.IndexExpr:
+		return b.compileIndexExpr(e)
+	case *dag.SliceExpr:
+		return b.compileSlice(e)
```
nwt, line 79: Nit: Call this compileSliceExpression so it matches the type as with the other Builder.compile* methods.

compiler/semantic/expr.go
```diff
@@ -116,6 +116,42 @@ func (a *analyzer) semExpr(e ast.Expr) (dag.Expr, error) {
 			Name: "cast",
 			Args: []dag.Expr{expr, typ},
 		}, nil
+	case *ast.IndexExpr:
+		expr, err := a.semExpr(e.Expr)
+		if err != nil {
+			return nil, err
+		}
+		// XXX This should be type checked as either an int, string or var.
```
nwt, line 124: I think we have to allow any type here since a map key can have any type.
```suggestion
```

compiler/semantic/expr.go
```diff
@@ -116,6 +116,42 @@ func (a *analyzer) semExpr(e ast.Expr) (dag.Expr, error) {
...
+		index, err := a.semExpr(e.Index)
+		if err != nil {
+			return nil, err
+		}
+		// If container is a path and the index expression is a string, then
```
nwt, line 129:
```suggestion
		// If expr is a path and index is a string, then
```

compiler/semantic/expr.go
```diff
@@ -116,6 +116,42 @@ func (a *analyzer) semExpr(e ast.Expr) (dag.Expr, error) {
...
+		// just extend the path.
+		if path := a.isIndexOfThis(expr, index); path != nil {
+			return path, nil
+		}
+		return &dag.IndexExpr{Kind: "IndexExpr", Expr: expr, Index: index}, nil
```
nwt, line 134: Nit: Almost every other compound literal in here is on multiple lines.
```suggestion
		return &dag.IndexExpr{
			Kind:  "IndexExpr",
			Expr:  expr,
			Index: index
		}, nil
```

compiler/semantic/expr.go
```diff
@@ -116,6 +116,42 @@ func (a *analyzer) semExpr(e ast.Expr) (dag.Expr, error) {
...
+	case *ast.SliceExpr:
+		expr, err := a.semExpr(e.Expr)
+		if err != nil {
+			return nil, err
+		}
+		// XXX Low/High should be type checked as either int or var.
```
nwt, line 140: Since slice indices can be arbitrary expressions, I think we can only check their types if they're literals.
```suggestion
		// Literal indices should be type checked as int.
```

compiler/semantic/expr.go
mattnibs: Good point!

compiler/ast/ast.go
mattnibs: The go std library uses high low but lets just keep it from/to.

nwt (approved):
:shipit:!

PR #5102, opened 2024-04-05
ast: A TypeDecl Node

mattnibs:
This commit adds a separate AST decl type for type declarations as opposed to using const deleclartions for this case. This is part of a series of changes preparing the AST code base to support position information for AST nodes.


compiler/ast/ast.go
```diff
@@ -210,6 +210,12 @@ type ConstDecl struct {
 	Expr Expr   `json:"expr"`
 }
 
+type TypeDecl struct {
```
nwt, line 213: Nit: Put this after OpDecl.

compiler/semantic/op.go
```diff
@@ -893,6 +899,21 @@ func (a *analyzer) semConstDecl(c *ast.ConstDecl) (dag.Def, error) {
...
+		return dag.Def{}, err
+	}
+	e := &dag.Literal{
+		Kind:  "Literal",
+		Value: fmt.Sprintf("<%s=%s>", zson.QuotedName(d.Name), typ),
+	}
+	if err := a.scope.DefineConst(a.zctx, d.Name, e); err != nil {
+		return dag.Def{}, err
+	}
+	return dag.Def{Name: d.Name, Expr: e}, nil
```
nwt, line 914: Nit: These are on multiple lines everywhere else in here.
```suggestion
	return dag.Def{
		Name: d.Name,
		Expr: e
	}, nil
```

zfmt/ast.go
```diff
@@ -316,6 +316,9 @@ func (c *canon) decl(d ast.Decl) {
 	case *ast.ConstDecl:
 		c.write("const %s = ", d.Name)
 		c.expr(d.Expr, "")
+	case *ast.TypeDecl:
```
nwt, line 319: Nit: Put this after OpDecl.

nwt (approved):

compiler/semantic/op.go
mattnibs: For me I follow my made up rule that if a struct declaration fits below 80 characters, I do it single line.  I think that seems reasonable, no?

zfmt/ast.go
mattnibs: We need to at some point go through and alphabetize ast nodes elsewhere.

compiler/semantic/op.go
nwt: Totally reasonable. I just prefer to see changes follow the prevailing style. Just a nit though so do as you like!

nwt (approved):
So nice I'm approving it twice.

PR #5103, opened 2024-04-05
Buffer zngio.scanner.workerCh for more concurrency

nwt:
Tracing with package runtime/trace shows the scanner.start and worker.run goroutines block frequently on the workerCh channel.  Add buffering to workerCH to reduce blocking and increase concurrency.

When scanning a 4 GB ZNG file containing Zeek logs, this yields a 1.1X speedup on my 10-core machine.


mattnibs (approved):
💥

PR #5104, opened 2024-04-06
Move assert out of peg

mattnibs:
Move the internals of the assert operator out of the parser PEG rules and into the code for the semantic checker.


compiler/semantic/op.go
```diff
@@ -798,6 +799,48 @@ func (a *analyzer) semOp(o ast.Op, seq dag.Seq) (dag.Seq, error) {
...
+							Elems: []dag.RecordElem{
+								&dag.Field{
+									Kind:  "Field",
+									Name:  "message",
+									Value: &dag.Literal{Kind: "Literal", Value: `"assertion failed"`},
+								},
+								&dag.Field{
+									Kind:  "Field",
+									Name:  "expr",
+									Value: &dag.Literal{Kind: "Literal", Value: strconv.Quote(o.Text)},
```
nwt, line 831: I think we want zson.QuotedString here since dag.Literal.Value is ZSON.

nwt (approved):
:shipit:!

PR #5108, opened 2024-04-16
Add positional information to AST Nodes

mattnibs:


mattnibs:
I should say this pr comes with a couple of loose ends that will need to be addressed in later work.
1. Template literals are a hot mess and are going to have totally wrong positions. They need to have their own ast node instead of the hacked up solution we currently use.
2. String positions are off- they currently refer to only the contents of the string and not the quotes. This will get resolved once we fix Template literals.
3. Positional information for operator arguments / flags is not great and this lead me to have to do hacky stuff to get positions working for various operators that rely on them (e.g., load, file, get, pool). The big problem here is that each Op implements flags or optional args in its own way are there is no standardized ast.Node for flags that is used across all ops. We should do this.

compiler/ast/ast.go
```diff
@@ -131,38 +154,59 @@ func (*QuotedString) PatternAST() {}
 func (*Regexp) PatternAST()       {}
 func (*String) PatternAST()       {}
 
+func (x *QuotedString) Pos() int { return x.Lquote }
+func (x *String) Pos() int       { return x.TextPos }
+
+func (x *QuotedString) End() int { return x.Lquote + 2 + len(x.Text) }
+func (x *String) End() int       { return x.TextPos + len(x.Text) }
```
nwt, line 161: In the interest of keeping things consistent with our style elsewhere, are you up for
(a) moving the Pos and End methods so they come right after the receiver type declarations and 
(b) changing the receiver names from `x` to the first character of the type name?

compiler/parser/api.go
```diff
@@ -17,7 +19,7 @@ func ParseZed(filenames []string, src string) (interface{}, error) {
 	if err != nil {
 		return nil, ImproveError(err, src, srcInfo)
 	}
-	return p, nil
+	return ast.Seq(sliceOf[ast.Op](p)), nil
```
nwt, line 22: Why not change the return type to ast.Seq instead?
```suggestion
	return sliceOf[ast.Op](p), nil
```

test.zed
```diff
@@ -0,0 +1,2 @@
+
+
```
nwt, line 2: Delete me.

compiler/parser/parser.peg
```diff
@@ -1,130 +1,164 @@
...
-#define PREPEND(value, arr) append([]interface{}{value}, (arr.([]interface{}))...)
-#define PRINT(...) fmt.Println(__VA_ARGS__)
-#define RETURN(x) return x, nil
-#define TEXT string(c.text)
-#define VAR(x) var x
 
 start
-  = __ scope:Scope __ EOF { RETURN(ARRAY(scope)) }
-  / __ seq:Seq __ EOF { RETURN(seq) }
+  = __ scope:Scope __ EOF { return []any{scope.(*ast.Scope)}, nil }
```
nwt, line 12:
```suggestion
  = __ scope:Scope __ EOF { return []any{scope}, nil }
```

compiler/job.go
```diff
@@ -93,7 +93,7 @@ func Parse(src string, filenames ...string) (ast.Seq, error) {
 	if err != nil {
 		return nil, err
 	}
-	return ast.UnmarshalObject(parsed)
+	return parsed.(ast.Seq), nil
```
nwt, line 96: This leave ast.UnmarshalObject so you can remove that.

compiler/ast/ast.go
mattnibs: If you think this will help readability, sure.

compiler/parser/parser.peg
```diff
@@ -1,130 +1,164 @@
...
+        Body: sliceOf[ast.Op](body),
+        TokenPos: c.pos.offset,
+        Rparen: lastPos(c, ")"),
+      }, nil
     }
 
 OpDeclBody
-  = scope:Scope { RETURN(ARRAY(scope)) }
-  / seq:Seq { RETURN(seq) }
+  = scope:Scope { return []any{scope.(*ast.Scope)}, nil }
```
nwt, line 69:
```suggestion
  = scope:Scope { return []any{scope}, nil }
```

compiler/parser/api.go
mattnibs: Definitely.

compiler/ast/ast.go
mattnibs: What do we want to do with ops considering that they are declared in a big `var ()` statement?

compiler/ast/ast.go
```diff
@@ -214,42 +261,115 @@ func (*RecordExpr) ExprAST() {}
...
+func (x *SQLExpr) End() int    { return -1 }
 
 type ConstDecl struct {
-	Kind string `json:"kind" unpack:""`
-	Name string `json:"name"`
-	Expr Expr   `json:"expr"`
+	Kind     string `json:"kind" unpack:""`
+	Name     string `json:"name"`
+	Expr     Expr   `json:"expr"`
+	TokenPos int    `json:"token_pos"`
```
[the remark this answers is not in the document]
mattnibs: I'm for this where it makes sense. It this for the position of an operator like `top` that has a keyword cool. If this is for unary expression that doesn't have a keyword but a token, I think token_pos makes sense.

compiler/ast/ast.go
nwt: I'm for breaking them out into a separate statement per type but let's do that later. If you agree, then I'd say put the methods for those types where ever you like.

mattnibs: Alright, then we'll just leave the op stuff where it is and reformat this in a future pr once this gets in.

compiler/ast/ast.go
```diff
@@ -16,53 +16,79 @@ import (
 // license that can be found in the LICENSE file.
 
 // Op is the interface implemented by all AST operator nodes.
+
+type Node interface {
+	Pos() int // Position of first character belonging to the node.
+	End() int // Position of first character immediately after the node.
+}
+
 type Op interface {
```
nwt, line 25: Move the Op doc comment so it remains adjacent to the Op declaration.

compiler/ast/ast.go
```diff
@@ -71,58 +97,103 @@ type Conditional struct {
...
 }
 
+func (s *SliceExpr) Pos() int { return s.Expr.Pos() }
+func (s *SliceExpr) End() int { return s.Rbrack + 1 }
+
 type Grep struct {
 	Kind    string `json:"kind" unpack:""`
 	Pattern Expr   `json:"pattern"`
 	Expr    Expr   `json:"expr"`
+	Left    int    `json:"left"`
```
nwt, line 152: If I understand correctly, this is the position of the "grep" keyword, so maybe call this KeywordPos and put it between Kind and Pattern?

compiler/ast/ast.go
```diff
@@ -179,21 +274,30 @@ type VectorValue struct {
...
+func (m *MapExpr) End() int { return m.Rpipe + 1 }
+
 type EntryExpr struct {
 	Key   Expr `json:"key"`
 	Value Expr `json:"value"`
 }
 
 type OverExpr struct {
 	Kind   string `json:"kind" unpack:""`
+	Left   int    `json:"left"`
```
nwt, line 292: If I understand correctly, this is the position of the "over" keyword, so maybe call this KeywordPos?

compiler/ast/ast.go
```diff
@@ -214,42 +318,56 @@ func (*RecordExpr) ExprAST() {}
...
 
+func (x *ConstDecl) Pos() int { return x.KeywordPos }
+func (x *FuncDecl) Pos() int  { return x.KeywordPos }
+func (x *OpDecl) Pos() int    { return x.KeywordPos }
+func (x *TypeDecl) Pos() int  { return x.KeywordPos }
+
+func (x *ConstDecl) End() int { return x.Expr.End() }
+func (x *FuncDecl) End() int  { return x.Rparen }
+func (x *OpDecl) End() int    { return x.Rparen }
+func (x *TypeDecl) End() int  { return x.Type.End() }
```
nwt, line 369: Nits: Put these after their type declarations. And replace x with the first letter of the receiver name.

compiler/ast/zed/type.go
```diff
@@ -1,58 +1,75 @@
...
-		Kind string `json:"kind" unpack:""`
+		Kind       string `json:"kind" unpack:""`
+		KeywordPos int    `json:"pos"`
 	}
 	TypeError struct {
-		Kind string `json:"kind" unpack:""`
-		Type Type   `json:"type"`
+		Kind       string `json:"kind" unpack:""`
+		Type       Type   `json:"type"`
+		KeywordPos int    `json:"keyword_pos"`
```
nwt, line 60: Nit: Maybe put this before Type?

compiler/parser/support.go
```diff
@@ -4,17 +4,41 @@ import (
...
-		ret = map[string]interface{}{
-			"kind": "BinaryExpr",
-			"op":   part[0],
-			"lhs":  ret,
-			"rhs":  part[1],
+func sliceOf[E any](s any) []E {
+	if s == nil {
+		return nil
+	}
+	arr := s.([]any)
```
nwt, line 16: Nit: This is a slice, not an array, so don't call it arr.

compiler/parser/support.go
```diff
@@ -33,20 +57,54 @@ func makeArgMap(args interface{}) (interface{}, error) {
...
+	}
+	return c.pos.offset + i
+}
+
+func firstPos(c *current, s string) int {
+	i := bytes.Index(c.text, []byte(s))
+	if i == -1 {
+		panic(fmt.Sprintf("system error: character %s not found in %s", s, string(c.text)))
+	}
+	return c.pos.offset
```
nwt, line 101: Shoudl this be `c.pos.offset + i`?

zfmt/ast.go
```diff
@@ -443,6 +442,21 @@ func (c *canon) op(p ast.Op) {
 		c.ret()
 		c.flush()
 		c.write(")")
+	case *ast.Pool:
+		c.next()
+		c.open("")
+		c.pool(p, true)
```
nwt, line 448: `true` is pretty opaque here. How about something more explicit?
```suggestion
		c.write("from ")
		c.pool(p)
```

compiler/semantic/expr.go
```diff
@@ -342,9 +342,12 @@ func dynamicTypeName(name string) *dag.Call {
 }
 
 func (a *analyzer) semGrep(grep *ast.Grep) (dag.Expr, error) {
-	e, err := a.semExpr(grep.Expr)
-	if err != nil {
-		return nil, err
+	var e dag.Expr = &dag.This{}
```
nwt, line 345:
```suggestion
	e := dag.Expr(&dag.This{Kind: "This"}
```

compiler/semantic/op.go
```diff
@@ -773,9 +808,12 @@ func (a *analyzer) semOp(o ast.Op, seq dag.Seq) (dag.Seq, error) {
 			Body:  body,
 		}), nil
 	case *ast.Sample:
-		e, err := a.semExpr(o.Expr)
-		if err != nil {
-			return nil, err
+		var e dag.Expr = &dag.This{}
```
nwt, line 811:
```suggestion
		e := dag.Expr(&dag.This{Kind: "This"}
```

compiler/semantic/op.go
```diff
@@ -404,6 +410,12 @@ func (a *analyzer) semOp(o ast.Op, seq dag.Seq) (dag.Seq, error) {
 	switch o := o.(type) {
 	case *ast.From:
 		return a.semFrom(o, seq)
+	case *ast.Pool:
+		return a.semOpSource(o, seq)
+	case *ast.File:
+		return a.semOpSource(o, seq)
+	case *ast.HTTP:
+		return a.semOpSource(o, seq)
```
nwt, line 418:
```suggestion
	case *ast.File, *ast.HTTP, *ast.Pool:
		return a.semOpSource(o, seq)
```

compiler/semantic/op.go
```diff
@@ -540,13 +552,15 @@ func (a *analyzer) semOp(o ast.Op, seq dag.Seq) (dag.Seq, error) {
...
 		}), nil
 	case *ast.Head:
-		expr, err := a.semExpr(o.Count)
-		if err != nil {
-			return nil, fmt.Errorf("head: %w", err)
-		}
-		val, err := kernel.EvalAtCompileTime(a.zctx, expr)
-		if err != nil {
-			return nil, fmt.Errorf("head: %w", err)
+		val := zed.NewInt(zed.TypeInt64, 1)
```
nwt, line 555:
```suggestion
		val := zed.NewInt64(1)
```

compiler/semantic/op.go
```diff
@@ -556,13 +570,15 @@ func (a *analyzer) semOp(o ast.Op, seq dag.Seq) (dag.Seq, error) {
...
 		}), nil
 	case *ast.Tail:
-		expr, err := a.semExpr(o.Count)
-		if err != nil {
-			return nil, fmt.Errorf("tail: %w", err)
-		}
-		val, err := kernel.EvalAtCompileTime(a.zctx, expr)
-		if err != nil {
-			return nil, fmt.Errorf("tail: %w", err)
+		val := zed.NewInt(zed.TypeInt64, 1)
```
nwt, line 573:
```suggestion
		val := zed.NewInt64(1)
```

compiler/semantic/op.go
```diff
@@ -83,6 +83,21 @@ func (a *analyzer) semTrunk(trunk ast.Trunk, out dag.Seq) (dag.Seq, error) {
...
+	sources, err := a.semSource(source)
+	if err != nil {
+		return nil, err
+	}
+	if len(sources) == 1 {
+		return append(out, append(dag.Seq{sources[0]}, out...)...), nil
+	}
+	paths := make([]dag.Seq, len(sources))
+	for k, s := range sources {
+		paths[k] = dag.Seq{s}
```
nwt, line 96: Minor nit: I think loops like this are a little clearer with `var` and `append`.
```suggestion
	var paths []dag.Seq
	for _, s := range sources {
		paths = append(paths, dag.Seq{s})
```

compiler/semantic/op.go
```diff
@@ -83,6 +83,21 @@ func (a *analyzer) semTrunk(trunk ast.Trunk, out dag.Seq) (dag.Seq, error) {
 
 //XXX make sure you can't read files from a lake instance
 
+func (a *analyzer) semOpSource(source ast.Source, out dag.Seq) (dag.Seq, error) {
+	sources, err := a.semSource(source)
+	if err != nil {
+		return nil, err
+	}
+	if len(sources) == 1 {
+		return append(out, append(dag.Seq{sources[0]}, out...)...), nil
```
nwt, line 92:
```suggestion
		return append(sources, out...), nil
```

compiler/ast/zed/value.go
```diff
@@ -29,15 +34,26 @@ type Any interface {
...
 }
 
+// Only Primitive and TypeValue are used in zed parser so only these are given
+// ast.Node features.
+
+var (
+	_ Node = (*Primitive)(nil)
+	_ Node = (*TypeValue)(nil)
+)
+
```
nwt, line 44: No need for this since we'll get errors in compiler/semantic/expr.go and zfmt/ast.go if these types don't implement this interface.
```suggestion
```

compiler/ast/ast.go
nwt: But also, why not use compiler/ast/zed.Node instead of declaring this?

compiler/ast/zed/value.go
```diff
@@ -1,5 +1,10 @@
 package zed
 
+type Node interface {
```
nwt, line 3: Nit: I think types.go would be a better place for this since it's embedded in `Type`.

compiler/ast/ast.go
mattnibs: It just seems weird that Node is part of astzed package, where as ast is the main package.

compiler/ast/ast.go
mattnibs: For sure, not sure why I did this

compiler/ast/ast.go
mattnibs: Like at some point we'll have a node Walk function if we want to do things like formatting and it seems weird to me that this would use `astzed.Node`.

compiler/ast/zed/value.go
```diff
@@ -29,15 +34,26 @@ type Any interface {
...
+		Text    string `json:"text"`
+		TextPos int    `json:"text_pos"`
 	}
 	Record struct {
-		Kind   string  `json:"Kind" unpack:""`
-		Fields []Field `json:"fields"`
+		Kind    string  `json:"kind" unpack:""`
+		Opening int     `json:"opening"`
+		Fields  []Field `json:"fields"`
+		Closing int     `json:"closing"`
```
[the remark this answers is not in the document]
mattnibs: Actually this isn't needed since zed uses ast.RecordExpr for records, so I just removed this.

compiler/parser/support.go
mattnibs: Whoops!

mattnibs: Oh this isn't even being used removed...

compiler/ast/zed/value.go
mattnibs: Actually don't need this at all in astzed so I got rid of it.

compiler/parser/parser.peg
```diff
@@ -156,110 +190,147 @@ SearchKeywordGuard
...
   = Literal
   / !RegexpPattern v:KeyWord {
-      RETURN(MAP("kind": "Primitive", "type": "string", "text": v))
+      return newPrimitive(c, "string", v.(string)), nil
+    }
+
+QuotedStringPattern 
+  = s:QuotedString {
+      return &ast.QuotedString{Kind: "QuotedString", Text: s.(string), Lquote: c.pos.offset}, nil
     }
```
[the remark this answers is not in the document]
mattnibs: QuotedString is of Pattern interface type in ast.go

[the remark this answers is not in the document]
mattnibs: Though I go back in forth: *Pattern and *Node are both kind of weird.

compiler/ast/ast.go
```diff
@@ -257,6 +377,24 @@ func (*TypeDecl) DeclAST()  {}
...
+
+func (s Seq) End() int {
+	if len(s) == 0 {
+		return -1
+	}
+	return s[len(s)-1].End()
+}
+
+func (seq *Seq) Prepend(front Op) {
+	*seq = append([]Op{front}, *seq...)
```
nwt, line 395:
```suggestion
func (s *Seq) Prepend(front Op) {
	*s = append([]Op{front}, *s...)
```

compiler/ast/ast.go
```diff
@@ -273,87 +411,110 @@ type (
...
+		Kind       string       `json:"kind" unpack:""`
+		KeywordPos int          `json:"keyword_pos"`
+		Args       []Assignment `json:"args"`
 	}
 	Drop struct {
-		Kind string `json:"kind" unpack:""`
-		Args []Expr `json:"args"`
+		Kind       string `json:"kind" unpack:""`
+		Args       []Expr `json:"args"`
+		KeywordPos int    `json:"keyword_pos"`
```
nwt, line 442:
```suggestion
		KeywordPos int    `json:"keyword_pos"`
		Args       []Expr `json:"args"`
```

compiler/ast/ast.go
```diff
@@ -513,9 +731,146 @@ func (*Sample) OpAST()       {}
...
+func (x *Summarize) End() int {
+	if len(x.Keys) > 0 {
+		return x.Keys[len(x.Keys)-1].End()
+	}
+	return x.Aggs[len(x.Aggs)-1].End()
+}
+
+// End for Top isn't right because positions do not work for -flush flag.
+// Positions for flags in operators will be addressed in a future PR.
+func (x *Top) End() int {
```
nwt, line 822: Nit: These comments aren't doc comments so put them inside the methods.

compiler/parser/parser.peg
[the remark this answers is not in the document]
nwt: My beef is that
```
QuotedStringPattern -> QuotedString
```
reverses the convention in the existing rules
```
Glob -> GlobPattern
Regexp -> RegexpPattern
```

QuotedStringNode certainly isn't ideal but I think it's less confusing than QuotedStringPattern.

But if you think QuotedStringPattern is the best alternative, I can live with it and you're welcome to mark this conversation resolved.

compiler/semantic/expr.go
```diff
@@ -342,9 +342,12 @@ func dynamicTypeName(name string) *dag.Call {
 }
 
 func (a *analyzer) semGrep(grep *ast.Grep) (dag.Expr, error) {
-	e, err := a.semExpr(grep.Expr)
-	if err != nil {
-		return nil, err
+	e := dag.Expr(&dag.This{})
```
nwt, line 345:
```suggestion
	e := dag.Expr(&dag.This{Kind: "This"})
```

compiler/semantic/op.go
```diff
@@ -761,9 +792,12 @@ func (a *analyzer) semOp(o ast.Op, seq dag.Seq) (dag.Seq, error) {
 			Body:  body,
 		}), nil
 	case *ast.Sample:
-		e, err := a.semExpr(o.Expr)
-		if err != nil {
-			return nil, err
+		e := dag.Expr(&dag.This{})
```
nwt, line 795:
```suggestion
		e := dag.Expr(&dag.This{Kind: "This"})
```

nwt (approved):

compiler/parser/parser.peg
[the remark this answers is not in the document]
mattnibs: Oh yeah I see, good point. Node it is.

PR #5109, opened 2024-04-18
Drop SQL language syntax from Zed language

mattnibs:


nwt (approved):
Rebased onto main as discussed.

PR #5114, opened 2024-04-23
Enabled JSON and ZSON color by default

mattnibs:
This commit changes the behavior of the ZSON and JSON outputs where if the writer is writing to a terminal the output will be colorized- independent of whether the output is indented or not.

### DIFF
--- a/zson/formatter.go
+++ b/zson/formatter.go
@@ -14,18 +14,19 @@ import (
 )
 
 type Formatter struct {
-	typedefs  map[string]*zed.TypeNamed
-	permanent map[string]*zed.TypeNamed
-	persist   *regexp.Regexp
-	tab       int
-	newline   string
-	builder   strings.Builder
-	stack     []strings.Builder
-	implied   map[zed.Type]bool
-	colors    color.Stack
+	typedefs      map[string]*zed.TypeNamed
+	permanent     map[string]*zed.TypeNamed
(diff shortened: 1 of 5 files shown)


zson/formatter.go
```diff
@@ -14,18 +14,19 @@ import (
...
+	tab           int
+	newline       string
+	builder       strings.Builder
+	stack         []strings.Builder
+	implied       map[zed.Type]bool
+	colors        color.Stack
+	colorDisabled bool
+}
+
+func NewFormatter(pretty int, colorDisabled bool, persist *regexp.Regexp) *Formatter {
```
nwt, line 29: I think this would be clearer with `color bool` instead.

mattnibs: Word.

mattnibs: The problem with this is that color = true wouldn't necessarily mean color is enabled since it depends when the writer is outputting to a terminal.

nwt: Ah, right. To make that work we'd need to get rid of pkt/terminal/color.Enabled. Let's do that later.

nwt (approved):

zson/formatter.go
mattnibs: Agreed I think we should get rid of that but yeah, later.

PR #5119, opened 2024-05-02
zq: Correct error message when no source is found

### LINKED ISSUE #5110: Implicit stdin support / Unintuitive behaviour when omitting '-'
When attempting to onboard from `jq`, I've had somewhat unintuitive results exploring basic commands:

`echo '{a:1,b:2,c:3}' | zq 'yield this' -` => `{a:1,b:2,c:3}`

So far so good, though the docs list the default query when omitting one as being `*`, so let's try that:

`echo '{a:1,b:2,c:3}' | zq '*' -` => `{a:1,b:2,c:3}`

Cool, equivalent behaviour to what I get from having no query:

`echo '{a:1,b:2,c:3}' | zq -` => `{a:1,b:2,c:3}`

Except... This option of omitting the query entirely leads into some interesting territory. According to the `--help` docs:
> If the first argument is a path to a valid file rather than a Zed query, then the Zed query is assumed to be "*", i.e., match and output all of the input.
> If the first argument is both a valid Zed query and an existing file, then the file overrides.

So what happens if we make a request with a valid query and allow standard input to be assumed as the input?

`echo '{a:1,b:2,c:3}' | zq 'yield this'` => `null`

Here's our first unexpected behaviour: not passing a query seems to ignore `stdin`. I thought maybe this was due to `-` counting as a valid zed query, but the following experiment shows otherwise:

`echo '{a:1,b:2,c:3}' | zq '-' -` => `{a:1,b:2,c:3}`

So let's try passing the real "default" query given by the CLI docs:

```
> echo '{a:1,b:2,c:3}' | zq '*'
 - a file could not be found with the name "*"
 - the argument did not parse as a valid Zed query
```

So the default zed query of '*' is not a valid zed query?

... Except when you pass stdin explicitly:

`echo '{a:1,b:2,c:3}' | zq '*' -` => `{a:1,b:2,c:3}`

As a new user, the issue seems to be that allowing implicit queries is a significant footgun and should be replaced with the more POSIX-friendly implicit `stdin` input convention. There may be some nuance here that makes shell scripting a lower-priority target than some other workflow that the `zq` tutorial hasn't introduced, but this is not apparent from day one, and the surprising nature of the CLI behaviour makes me hesitant to adopt this in lieu of JQ.

mattnibs:
This commit fixes an issue where the wrong error message was returned for a single argument zq invocation where the argument is a valid Zed query but the query contains no source.

Closes #5110

### DIFF
--- a/cli/queryflags/flags.go
+++ b/cli/queryflags/flags.go
@@ -42,17 +42,23 @@ func (f *Flags) ParseSourcesAndInputs(paths []string) ([]string, ast.Seq, *parse
 			// and appears to start with a from or yield operator.
 			// Otherwise, consider it a file.
 			query, sset, err := compiler.Parse(src, f.Includes...)
-			if err == nil {
-				if s, err := semantic.Analyze(context.Background(), query, data.NewSource(storage.NewLocalEngine(), nil), nil); err == nil {
-					if semantic.HasSource(s) {
-						return nil, query, sset, false, nil
-					}
-					if semantic.StartsWithYield(s) {
-						return nil, query, sset, true, nil
-					}
-				}
+			if err != nil {
+				return nil, nil, nil, false, singleArgError(src, err)
 			}
-			return nil, nil, nil, false, singleArgError(src, err)
+			s, err := semantic.Analyze(context.Background(), query, data.NewSource(storage.NewLocalEngine(), nil), nil)
+			if err != nil {
+				if list, ok := err.(parser.ErrorList); ok {
+					list.SetSourceSet(sset)
+				}
+				return nil, nil, nil, false, err
+			}
+			if semantic.HasSource(s) {
+				return nil, query, sset, false, nil
+			}
+			if semantic.StartsWithYield(s) {
+				return nil, query, sset, true, nil
+			}
+			return nil, nil, nil, false, singleArgError(src, nil)
 		}
 	}
 	query, sset, err := parser.ParseZed(f.Includes, src)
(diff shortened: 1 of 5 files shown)


cli/queryflags/flags.go
```diff
@@ -97,6 +97,8 @@ func singleArgError(src string, err error) error {
 		for _, l := range strings.Split(perr.ParseErrorContext(), "\n") {
 			fmt.Fprintf(&b, "\n   %s", l)
 		}
+	} else if err == nil {
+		b.WriteString("\n - the argument is a valid Zed query but does not begin with a source or yield operator (e.g., \"file input.zson\")")
```
philrz, line 101:
```suggestion
		b.WriteString("\n - the argument is a valid Zed query but does not begin with a source (e.g., \"file input.zson\") or yield operator")
```

cmd/zq/ztests/from-pool-error.yaml
```diff
@@ -4,6 +4,6 @@ script: |
 outputs:
   - name: stderr
     data: |
-      zq: could not invoke zq with a single argument because:
-       - a file could not be found with the name "from ( pool a )"
-       - the argument did not parse as a valid Zed query
+      zq: from pool cannot be used without a lake (line 1, column 8):
```
philrz, line 7:
```suggestion
      zq: "from pool" cannot be used without a lake (line 1, column 8):
```

philrz (commented):
@mattnibs: Thanks for adjusting the error message. I'm good with these improvements, but since there's actual code changes involved here it'd probably be better for the official approval to come from @nwt.

nwt (approved):

PR #5120, opened 2024-05-03
Localized Errors

mattnibs:
This commit introduces localized errors- functionality for tying generated errors to a location in a Zed source set. A version of this somewhat exists currently but it is limited in that it is specific only to Parse errors and currently only supports returning a single error. In this pr localized errors are only used for parser errors but a follow up commit will enable localized errors for semantic errors as well.


mattnibs:
Reformatting we'll be right back...

PR #5121, opened 2024-05-08
Add positional information to semantic errors

mattnibs:
This commit changes errors produced from semantic analysis so they include a reference to the AST node which produced the error. The allows for highlighting the specific portion of the source that caused a semantic error.

This commit also changes how semantic errors are collected in that the semantic analyzer no longer exits as soon as the first error is found and instead goes on the analyze the rest of the AST, reporting all semantic errors in the AST.

### DIFF
--- a/compiler/ast/ast.go
+++ b/compiler/ast/ast.go
@@ -110,7 +110,7 @@ func (c *Call) End() int {
 	if c.Where != nil {
 		return c.Where.End()
 	}
-	return c.Rparen
+	return c.Rparen + 1
 }
 
 type Cast struct {
@@ -180,7 +180,7 @@ type Regexp struct {
 }
 
 func (r *Regexp) Pos() int { return r.PatternPos }
-func (r *Regexp) End() int { return r.PatternPos + len(r.Pattern) }
+func (r *Regexp) End() int { return r.PatternPos + len(r.Pattern) + 2 }
 
 type String struct {
 	Kind    string `json:"kind" unpack:""`
@@ -432,9 +432,9 @@ type (
 		NullsFirst bool        `json:"nullsfirst"`
 	}
 	Cut struct {
-		Kind       string       `json:"kind" unpack:""`
-		KeywordPos int          `json:"keyword_pos"`
-		Args       []Assignment `json:"args"`
+		Kind       string      `json:"kind" unpack:""`
+		KeywordPos int         `json:"keyword_pos"`
+		Args       Assignments `json:"args"`
 	}
 	Drop struct {
 		Kind       string `json:"kind" unpack:""`
@@ -471,10 +471,10 @@ type (
 		Kind string `json:"kind" unpack:""`
 		// StartPos is not called KeywordPos for Summarize since the "summarize"
 		// keyword is optional.
-		StartPos int          `json:"start_pos"`
-		Limit    int          `json:"limit"`
-		Keys     []Assignment `json:"keys"`
-		Aggs     []Assignment `json:"aggs"`
+		StartPos int         `json:"start_pos"`
+		Limit    int         `json:"limit"`
+		Keys     Assignments `json:"keys"`
+		Aggs     Assignments `json:"aggs"`
 	}
 	Top struct {
 		Kind       string `json:"kind" unpack:""`
@@ -484,9 +484,9 @@ type (
 		Flush      bool   `json:"flush"`
 	}
 	Put struct {
-		Kind       string       `json:"kind" unpack:""`
-		KeywordPos int          `json:"keyword_pos"`
-		Args       []Assignment `json:"args"`
+		Kind       string      `json:"kind" unpack:""`
+		KeywordPos int         `json:"keyword_pos"`
+		Args       Assignments `json:"args"`
 	}
 	Merge struct {
 		Kind       string `json:"kind" unpack:""`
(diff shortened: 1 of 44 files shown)


compiler/semantic/op.go
```diff
@@ -915,70 +813,55 @@ func (a *analyzer) singletonAgg(agg ast.Assignment, seq dag.Seq) dag.Seq {
...
+	}
 }
 
-func (a *analyzer) semTypeDecl(d *ast.TypeDecl) (dag.Def, error) {
+func (a *analyzer) semTypeDecl(d *ast.TypeDecl) dag.Def {
 	typ, err := a.semType(d.Type)
 	if err != nil {
-		return dag.Def{}, err
+		a.error(d.Type, err)
+		typ = "error(\"string\")"
```
nwt, line 852: Nit: `null` feels more appropriate to me.
```suggestion
		typ = "null"
```

compiler/semantic/op.go
```diff
@@ -987,118 +870,117 @@ func (a *analyzer) semFuncDecls(decls []*ast.FuncDecl) ([]*dag.Func, error) {
...
 }
 
-func assignmentHasDynamicLHS(assignments []dag.Assignment) bool {
-	for _, a := range assignments {
-		if _, ok := a.LHS.(*dag.This); !ok {
+func (a *analyzer) checkStaticAssignment(asts []ast.Assignment, assignments []dag.Assignment) bool {
+	for k, assign := range assignments {
+		if _, ok := assign.LHS.(*dag.BadExpr); ok {
+			continue
+		}
```
nwt, line 961: I don't think this is necessary.
```suggestion
```

compiler/semantic/op.go
```diff
@@ -987,118 +870,117 @@ func (a *analyzer) semFuncDecls(decls []*ast.FuncDecl) ([]*dag.Func, error) {
...
+		if _, ok := assign.LHS.(*dag.This); !ok {
+			a.error(asts[k].LHS, errors.New("assignment key must be static"))
```
nwt, line 963: Nit: Make this consistent with the similar error in semOpAssignment.

```suggestion
			a.error(asts[k].LHS, errors.New("output field must be static"))
```

compiler/ast/ast.go
```diff
@@ -674,14 +674,19 @@ type Assignment struct {
...
-func (a *Assignment) Pos() int {
+func (a Assignment) Pos() int {
 	if a.LHS != nil {
 		return a.LHS.Pos()
 	}
 	return a.RHS.Pos()
 }
 
-func (a *Assignment) End() int { return a.RHS.Pos() }
+func (a Assignment) End() int { return a.RHS.End() }
```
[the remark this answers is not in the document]
mattnibs: okeee

compiler/semantic/expr.go
```diff
@@ -51,109 +52,78 @@ func (a *analyzer) semExpr(e ast.Expr) (dag.Expr, error) {
...
 			if err != nil {
-				return nil, err
+				a.error(e, err)
+				return badExpr()
 			}
 			val = "<" + tv + ">"
 		default:
-			return nil, fmt.Errorf("unexpected term value: %s", e.Kind)
+			a.error(e, fmt.Errorf("unexpected term value: %s", e.Kind))
+			return badExpr()
```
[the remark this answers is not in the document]
mattnibs: Agreed

compiler/semantic/expr.go
```diff
@@ -167,70 +137,61 @@ func (a *analyzer) semExpr(e ast.Expr) (dag.Expr, error) {
...
 	case *ast.RecordExpr:
 		fields := map[string]struct{}{}
 		var out []dag.RecordElem
 		for _, elem := range e.Elems {
 			switch elem := elem.(type) {
 			case *ast.Field:
 				if _, ok := fields[elem.Name]; ok {
-					return nil, fmt.Errorf("record expression: %w", &zed.DuplicateFieldError{Name: elem.Name})
+					a.error(elem, fmt.Errorf("record expression: %w", &zed.DuplicateFieldError{Name: elem.Name}))
+					// XXX Add BadRecordElem?
```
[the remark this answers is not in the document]
mattnibs: No idea, dropping that comment like its hot

compiler/semantic/op.go
mattnibs: It is necessary because then if you have an error with the LHS you will also get error "assignment key must be static" from below.

nwt: 🤦

compiler/semantic/expr.go
```diff
@@ -167,70 +136,60 @@ func (a *analyzer) semExpr(e ast.Expr) (dag.Expr, error) {
...
 				out = append(out, &dag.Field{
 					Kind:  "Field",
 					Name:  elem.Name,
 					Value: e,
 				})
 			case *ast.ID:
 				if _, ok := fields[elem.Name]; ok {
-					return nil, fmt.Errorf("record expression: %w", &zed.DuplicateFieldError{Name: elem.Name})
+					a.error(elem, fmt.Errorf("record expression: %w", &zed.DuplicateFieldError{Name: elem.Name}))
+					// XXX Add BadRecordElem?
```
nwt, line 181: Nit:
```suggestion
```

compiler/semantic/op.go
```diff
@@ -987,118 +869,117 @@ func (a *analyzer) semFuncDecls(decls []*ast.FuncDecl) ([]*dag.Func, error) {
...
-func assignmentHasDynamicLHS(assignments []dag.Assignment) bool {
-	for _, a := range assignments {
-		if _, ok := a.LHS.(*dag.This); !ok {
+func (a *analyzer) checkStaticAssignment(asts []ast.Assignment, assignments []dag.Assignment) bool {
+	for k, assign := range assignments {
+		if _, ok := assign.LHS.(*dag.BadExpr); ok {
+			continue
+		}
+		if _, ok := assign.LHS.(*dag.This); !ok {
+			a.error(asts[k].LHS, errors.New("output key must be static"))
```
nwt, line 962: Nit: I'd go with "field" for consistency with the error in semOpAssignment.
```suggestion
			a.error(asts[k].LHS, errors.New("output field must be static"))
```

nwt (approved):
:shipit:

PR #5122, opened 2024-05-13
ast.Errors

mattnibs:
This commit introduces ast.Errors- an error type that should be used whenever a generated error refers to a place in the query source. This commit switches the parser package to use ast.Error but a follow up commit will use ast.Errors for the semantic package as well.

### DIFF
--- a/lake/api/api.go
+++ b/lake/api/api.go
@@ -22,8 +22,8 @@ import (
 
 type Interface interface {
 	Root() *lake.Root
-	Query(ctx context.Context, head *lakeparse.Commitish, src string, srcfiles ...string) (zio.ReadCloser, error)
-	QueryWithControl(ctx context.Context, head *lakeparse.Commitish, src string, srcfiles ...string) (zbuf.ProgressReadCloser, error)
+	Query(ctx context.Context, head *lakeparse.Commitish, src string) (zio.ReadCloser, error)
+	QueryWithControl(ctx context.Context, head *lakeparse.Commitish, src string) (zbuf.ProgressReadCloser, error)
 	PoolID(ctx context.Context, poolName string) (ksuid.KSUID, error)
 	CommitObject(ctx context.Context, poolID ksuid.KSUID, branchName string) (ksuid.KSUID, error)
 	CreatePool(context.Context, string, order.SortKey, int, int64) (ksuid.KSUID, error)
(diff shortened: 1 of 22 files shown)


lake/ztests/query-parse-error.yaml
```diff
@@ -34,34 +34,34 @@ outputs:
   - name: stderr
     data: |
       =1=
-      error parsing Zed at column 11:
+      error parsing Zed (line 1, column 11):
```
nwt, line 37: This looks like a purely cosmetic change. Can we save it in a separate PR?

lake/api/api.go
```diff
@@ -22,8 +22,8 @@ import (
 
 type Interface interface {
 	Root() *lake.Root
-	Query(ctx context.Context, head *lakeparse.Commitish, src string, srcfiles ...string) (zio.ReadCloser, error)
-	QueryWithControl(ctx context.Context, head *lakeparse.Commitish, src string, srcfiles ...string) (zbuf.ProgressReadCloser, error)
+	Query(ctx context.Context, head *lakeparse.Commitish, src string) (zio.ReadCloser, error)
+	QueryWithControl(ctx context.Context, head *lakeparse.Commitish, src string) (zbuf.ProgressReadCloser, error)
```
nwt, line 26: Do these APIs need to change?

mattnibs: I think it should be this way since if you let `lake/api/api` do the ConcatSources for you won't be able to map error locations to your source set.

lake/ztests/query-parse-error.yaml
mattnibs: Not really since with semantic errors you have different error messages that may not syntactically work in the "append this to a sentence" way that errors are currently formatted.

lake/api/api.go
nwt: I was thinking that the mapping would continue to happen inside these methods in the same way it effectively does now. Shifting that burden to the caller seems like it makes for a more complicated interface since the caller now has to deal with parser.ConcatSource and clierrors.Format.

Does shifting that burden have some benefit I'm not seeing?

mattnibs: I eventually want to rewrite the error text to use asni escape sequences for pretty console output- where should the package that formats the error messages go?

mattnibs:
Closing in favor of https://github.com/brimdata/zed/pull/5128

PR #5124, opened 2024-05-14
Reference the "zed serve -manage" option in zed command docs

philrz:
With the merge of https://github.com/brimdata/zui/pull/3006 and wanting to generally be able to point users at the benefits of compaction when they have lots of small commits, I realized the docs could stand to mention the availability of the `zed serve -manage` option since I expect many users would prefer this over having to run `zed manage` as a separate process.

### DIFF
--- a/cmd/zed/manage/command.go
+++ b/cmd/zed/manage/command.go
@@ -19,7 +19,30 @@ var Cmd = &charm.Spec{
 	Name:  "manage",
 	Usage: "manage",
 	Short: "run compaction and other maintenance tasks on a lake",
-	New:   New,
+	Long: `
+The manage command performs maintenance tasks on a lake.
+
+Currently the only supported task is compaction, which reduces
+fragmentation by reading data objects in a pool and writing their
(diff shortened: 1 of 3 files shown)


mattnibs (commented):
Great! You wanted to add this to the manage command line instructions as well.

philrz:
Thanks @mattnibs. Good idea. I just pushed another commit to add the `zed manage` command help. I ended up basically taking the text from the docs site verbatim. Then I noticed that the `zed serve` command help was somewhat out of date, so I did the same there.

That all got me to thinking how the true docs nerd thing to do might be to "single source" the content in some way so that both the docs site and the command help could be more easily kept in sync. But I can foresee just enough complexity with doing that properly that I'm prepared to put it off for another time. 😅

mattnibs:
@philrz indeed it would be nice to have a single source for docs and cli help. I know that cobra allows you to output your cli help to markdown docs. ..

mattnibs (approved):

PR #5126, opened 2024-05-17
Add query describe endpoint

mattnibs:
This commit adds the query describe endpoint- a service endpoint the returns information about a posted query. This endpoint is meant for internal use from within Zui and is not meant for public consumption.

### DIFF
--- a/compiler/describe.go
+++ b/compiler/describe.go
@@ -1,0 +1,26 @@
+package compiler
+
+import (
+	"context"
+
+	"github.com/brimdata/zed/compiler/data"
+	"github.com/brimdata/zed/compiler/describe"
+	"github.com/brimdata/zed/compiler/parser"
+	"github.com/brimdata/zed/compiler/semantic"
+	"github.com/brimdata/zed/lakeparse"
+)
+
+func Describe(ctx context.Context, query string, src *data.Source, head *lakeparse.Commitish) (*describe.Info, error) {
+	seq, sset, err := Parse(query)
+	if err != nil {
+		return nil, err
+	}
+	entry, err := semantic.AnalyzeAddSource(ctx, seq, src, head)
+	if err != nil {
+		if list, ok := err.(parser.ErrorList); ok {
+			list.SetSourceSet(sset)
+		}
+		return nil, err
+	}
+	return describe.Analyze(ctx, src, entry)
+}
--- a/compiler/describe/analyze.go
+++ b/compiler/describe/analyze.go
@@ -1,0 +1,156 @@
+package describe
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/brimdata/zed/compiler/ast/dag"
+	"github.com/brimdata/zed/compiler/data"
+	"github.com/brimdata/zed/compiler/optimizer"
+	"github.com/brimdata/zed/lake"
+	"github.com/brimdata/zed/order"
+	"github.com/brimdata/zed/pkg/field"
+	"github.com/segmentio/ksuid"
+)
+
+type Info struct {
+	Sources  []Source  `json:"sources"`
+	Channels []Channel `json:"channels"`
+}
+
+type Source interface {
+	Source()
+}
+
+type (
+	LakeMeta struct {
+		Kind string `json:"kind"`
+		Meta string `json:"meta"`
+	}
+	Pool struct {
+		Kind string      `json:"kind"`
+		Name string      `json:"name"`
+		ID   ksuid.KSUID `json:"id"`
+	}
+	Path struct {
+		Kind string `json:"kind"`
+		URI  string `json:"uri"`
+	}
+)
+
+func (*LakeMeta) Source() {}
+func (*Pool) Source()     {}
+func (*Path) Source()     {}
+
+type Channel struct {
+	AggregationKeys field.List     `json:"aggregation_keys"`
+	Sort            *order.SortKey `json:"sort"`
+}
+
+func Analyze(ctx context.Context, source *data.Source, seq dag.Seq) (*Info, error) {
+	var info Info
+	var err error
+	if info.Sources, err = describeSources(ctx, source.Lake(), seq[0]); err != nil {
+		return nil, err
+	}
+	sortKeys, err := optimizer.New(ctx, source).SortKeys(seq)
+	if err != nil {
+		return nil, err
+	}
+	aggKeys := describeAggs(seq, []field.List{nil})
+	for i := range sortKeys {
+		// Convert SortKey to a pointer so a nil sort is encoded as null for
+		// JSON/ZSON.
+		var s *order.SortKey
+		if !sortKeys[i].IsNil() {
+			s = &sortKeys[i]
+		}
+		info.Channels = append(info.Channels, Channel{
+			Sort:            s,
+			AggregationKeys: aggKeys[i],
+		})
+	}
+	return &info, nil
+}
+
+func describeSources(ctx context.Context, lk *lake.Root, o dag.Op) ([]Source, error) {
+	switch o := o.(type) {
+	case *dag.Fork:
+		var s []Source
+		for _, p := range o.Paths {
+			out, err := describeSources(ctx, lk, p[0])
+			if err != nil {
+				return nil, err
+			}
+			s = append(s, out...)
+		}
+		return s, nil
+	case *dag.DefaultScan:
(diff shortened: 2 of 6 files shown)


compiler/describe/analyze.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/brimdata/zed/compiler/optimizer"
+	"github.com/brimdata/zed/lake"
+	"github.com/brimdata/zed/order"
+	"github.com/brimdata/zed/pkg/field"
+	"github.com/segmentio/ksuid"
+)
+
+type Info struct {
+	Sources      []Source        `json:"sources"`
+	Aggregations []*Aggregation  `json:"aggregations"`
```
nwt, line 19: Nit: Simplify this (unless you expect to add more fields to the Aggregation type down the road).
```suggestion
	AggregationKeys field.List   `json:"aggregation_keys"
```

compiler/optimizer/optimizer.go
```diff
@@ -185,6 +185,10 @@ func (o *Optimizer) OptimizeDeleter(seq dag.Seq, replicas int) (dag.Seq, error)
 	return dag.Seq{lister, scatter, merge}, nil
 }
 
+func (o *Optimizer) SortKeys(seq dag.Seq) ([]order.SortKey, error) {
+	return o.propagateSortKey(copyOps(seq), []order.SortKey{order.Nil})
+}
+
```
nwt, line 191: Nit: Put this right before propagateSortKey since it's just a wrapper around that?

service/ztests/query-describe.yaml
```diff
@@ -0,0 +1,76 @@
...
+              },
+              {
+                  "kind": "Pool",
+                  "name": "test2",
+                  "id": "XXX"
+              }
+          ],
+          "aggregations": [
+              null
+          ],
```
nwt, line 44: Since there's no aggregation in this query, I'd expect `null` or `[]` here. Is there a reason `[null]` makes sense?

mattnibs: `null` let's you know that for this particular leg of the flowgraph there is no aggregation.

compiler/describe.go
```diff
@@ -0,0 +1,29 @@
...
+
+func Describe(query string, src *data.Source, head *lakeparse.Commitish) (*describe.Info, error) {
+	seq, err := Parse(query)
+	if err != nil {
+		return nil, err
+	}
+	if len(seq) == 0 {
+		return nil, errors.New("internal error: AST seq cannot be empty")
+	}
+	rctx := runtime.NewContext(context.Background(), zed.NewContext())
```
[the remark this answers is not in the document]
mattnibs: 👍🏻 I'm not sure why I did this, probably copy pasta from NewJob

service/ztests/query-describe.yaml
nwt: Then the `aggregations` and `sort_keys` arrays each have one entry per channel? That isn't obvious from the structure here. How about making it clear with something like this?

```json
      {
          "sources": [...],
          "channels": [
              {
                    "aggregation_keys": [...],
                    "sort_keys": [...]
              }
          ]
      }
```

mattnibs: One thing this approach doesn't support is letting you know if there is an aggregation in the channel. We could add a `has_aggregation` field or keep the aggregation structure, any preference?

nwt: I think it does support that. Each entry in the `channels` array is a record containing the aggregation keys and sort keys for the channel. If `aggregation_keys` is `null`, the channel has no aggregation. Is that sufficient.

mattnibs: So `[]` means has aggregation but no keys whereas `null` means no aggregation? I suppose this works since an empty array in javascript evaluations as truthy.

compiler/describe/analyze.go
```diff
@@ -0,0 +1,160 @@
...
+		}
+		return aggs
+	case *dag.Scatter:
+		var aggs []field.List
+		for _, p := range op.Paths {
+			aggs = append(aggs, describeAggs(p, []field.List{nil})...)
+		}
+		return aggs
+	case *dag.Summarize:
+		// The field list for aggregation with no keys in an empty slice and
```
nwt, line 147:
```suggestion
		// The field list for aggregation with no keys is an empty slice and
```

compiler/describe/analyze.go
```diff
@@ -0,0 +1,160 @@
...
+	case *dag.CommitMetaScan:
+		return sourceOfPool(ctx, lk, o.Pool)
+	case *dag.LakeMetaScan:
+		return []Source{&LakeMeta{Kind: "LakeMeta", Meta: o.Meta}}, nil
+	default:
+		return nil, fmt.Errorf("unsupported source type %T", o)
+	}
+}
+
+func sourceOfPool(ctx context.Context, lk *lake.Root, id ksuid.KSUID) ([]Source, error) {
```
nwt, line 110: Nit: Given that this returns a slice, `sourcesOfPool` seems like a better name.

compiler/describe/analyze.go
```diff
@@ -0,0 +1,160 @@
...
+	if lk == nil {
+		panic(errors.New("internal error: lake operation cannot be used in non-lake context"))
```
nwt, line 112: Don't panic when you can return an error.
```suggestion
		return nil, errors.New("internal error: lake operation cannot be used in non-lake context")
```

service/ztests/query-describe.yaml
```diff
@@ -0,0 +1,94 @@
...
+      from (
+        pool test1
+        pool test2
+      ) | put foo := "bar"
+  - name: agg.zed
+    data: |
+      count() by key1:=v1, key2
+  - name: agg-no-keys.zed
+    data: |
+      sum(this)
```
nwt, line 26: How about a test with two channels? And one with a channel that has both an aggregation and a sort?

compiler/describe/analyze.go
```diff
@@ -0,0 +1,160 @@
...
+		Kind string `json:"kind" unpack:""`
+		Meta string `json:"meta"`
+	}
+	Pool struct {
+		Kind string      `json:"kind" unpack:""`
+		Name string      `json:"name"`
+		ID   ksuid.KSUID `json:"id"`
+	}
+	Path struct {
+		Kind string `json:"kind" unpack:""`
```
nwt, line 37: Nit: No need for `unpack` on these.

compiler/describe/analyze.go
mattnibs: We do a lot of returning errors in cases like this where this is clearly error that should have been caught in the semantic pass. My instinct is to stop returning these `internal errors: ...` and just panicking because if we get here it means something is seriously wrong. What do you think?

compiler/describe/analyze.go
mattnibs: I mean it only ever returns a single pool since a pool can only have one source. I originally had this return a single source then had to return it as a list of sources, but doing this meant I had to do this three times which felt tedious.

service/ztests/query-describe.yaml
mattnibs: Yeah probably should add these.

compiler/describe/analyze.go
nwt: I'm with you.

With that in mind then, how about removing this check and just letting the call to lk.OpenPool panic if lk is nil?

mattnibs: Let's go!

compiler/describe.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+func Describe(ctx context.Context, query string, src *data.Source, head *lakeparse.Commitish) (*describe.Info, error) {
+	seq, sset, err := Parse(query)
+	if err != nil {
+		return nil, err
+	}
+	if len(seq) == 0 {
+		return nil, errors.New("internal error: AST seq cannot be empty")
+	}
```
nwt, line 21: Nit: In line with the policy described at https://github.com/brimdata/zed/pull/5126#discussion_r1612923925, drop this?
```suggestion
```

service/ztests/query-describe.yaml
nwt: @mattnibs: Add these and I'll approve.

mattnibs: working on it....

nwt (approved):

PR #5127, opened 2024-05-23
JSON pretty: empty containers

mattnibs:
For the JSON writer in pretty mode ensure that empty objects and array are printed on the same line.

### DIFF
--- a/zio/jsonio/writer.go
+++ b/zio/jsonio/writer.go
@@ -94,6 +94,10 @@ func (w *Writer) writeAny(tab int, val zed.Value) {
 func (w *Writer) writeRecord(tab int, typ *zed.TypeRecord, bytes zcode.Bytes) {
 	tab += w.tab
 	w.punc('{')
+	if len(bytes) == 0 {
+		w.punc('}')
+		return
+	}
 	it := bytes.Iter()
 	for i, f := range typ.Fields {
 		if i != 0 {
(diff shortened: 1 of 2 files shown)


zio/jsonio/writer.go
```diff
@@ -95,47 +95,50 @@ func (w *Writer) writeRecord(tab int, typ *zed.TypeRecord, bytes zcode.Bytes) {
...
+			w.newline()
+			w.indent(tab - w.tab)
+			break
+		}
+		w.punc(',')
 	}
-	w.newline()
-	w.indent(tab - w.tab)
 	w.punc('}')
 }
```
nwt, line 108: This pattern's OK but I think what's going on here would be a lot clearer if it were conditioned on `len(bytes)`, as in

```go
	w.punc('{')
	if len(bytes) == 0 {
		w.punc('}')
		return
	}
	it := bytes.Iter()
		if i != 0 {
			w.punc(',')
		}
		w.writeEntry(tab, f.Name, w.arena.New(f.Type, it.Next()))
	}
	w.newline()
	w.indent(tab - w.tab)
	w.punc('}')
```

or

```go
	w.punc('{')
	if len(bytes) > 0 {
		it := bytes.Iter()
			if i != 0 {
				w.punc(',')
			}
			w.writeEntry(tab, f.Name, w.arena.New(f.Type, it.Next()))
		}
		w.newline()
		w.indent(tab - w.tab)
	}
	w.punc('}')
```

nwt (approved):

PR #5132, opened 2024-06-05
ZSON: Quoted name type values

mattnibs:
This commit fixes an issue in the ZSON formatter where the name of a named type value is not getting placed in quotes when the name is not a valid identifier.

### DIFF
--- a/zson/ztests/type-value.yaml
+++ b/zson/ztests/type-value.yaml
@@ -3,6 +3,6 @@ zed: "*"
 input: &input |
   <int64>
   {typ:<int64>}
-  <{a:"@foo"={bar:string},b:"@foo"}>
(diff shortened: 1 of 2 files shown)


zson/ztests/type-value.yaml
```diff
@@ -3,5 +3,6 @@ zed: "*"
 input: &input |
   <int64>
   {typ:<int64>}
+  <{a:"@foo"={bar:string},b:"@foo"}>
```
nwt, line 6: Nit: Different field names and a simpler underlying type would make the intent behind this case a little clearer.
```suggestion
  <{def:"@foo"=int64,ref:"@foo">
```

nwt (approved):

PR #5133, opened 2024-06-05
Parser: Allowed quoted named types

mattnibs:
The commit adds support in the Zed parser to allow parsing of quoted named type values.

### DIFF
--- a/compiler/parser/parser.go
+++ b/compiler/parser/parser.go
@@ -8271,35 +8271,44 @@ var g = &grammar{
 								&labeledExpr{
 									pos:   position{line: 1228, col: 5, offset: 30072},
 									label: "name",
-									expr: &ruleRefExpr{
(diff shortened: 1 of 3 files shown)


compiler/ztests/quoted-named-type.yaml
```diff
@@ -0,0 +1,8 @@
+zed: |
+  yield <"@foo"={x:int64}>
+
+input: |
+  null
+
+output: |
+  <"@foo"={x:int64}>
```
nwt, line 8: Nit: This is primarily a parser test so maybe put it in compiler/parser/ztests?

nwt (approved):

compiler/ztests/quoted-named-type.yaml
mattnibs: Sure

PR #5136, opened 2024-06-06
parser: Use *ast.ID for identifers

mattnibs:
The commit changes the AST representation for declaration identifiers as well as parameters in user-defined ops/funcs and over expressions. Previously these were represented as strings but are now represented as *ast.ID which has the advantage of having location information - meaning errors having to do with these nodes can be set to them rather than the parent.

### DIFF
--- a/compiler/ast/ast.go
+++ b/compiler/ast/ast.go
@@ -96,15 +96,14 @@ func (c *Conditional) End() int { return c.Else.End() }
 // a function call has the standard semantics where it takes one or more arguments
 // and returns a result.
 type Call struct {
-	Kind    string `json:"kind" unpack:""`
-	Name    string `json:"name"`
-	NamePos int    `json:"name_pos"`
-	Args    []Expr `json:"args"`
-	Rparen  int    `json:"rparen"`
-	Where   Expr   `json:"where"`
+	Kind   string `json:"kind" unpack:""`
+	Name   *ID    `json:"name"`
+	Args   []Expr `json:"args"`
+	Rparen int    `json:"rparen"`
+	Where  Expr   `json:"where"`
 }
 
-func (c *Call) Pos() int { return c.NamePos }
+func (c *Call) Pos() int { return c.Name.Pos() }
 
 func (c *Call) End() int {
 	if c.Where != nil {
@@ -185,7 +184,7 @@ func (r *Regexp) End() int { return r.PatternPos + len(r.Pattern) + 2 }
 type String struct {
 	Kind    string `json:"kind" unpack:""`
 	Text    string `json:"text"`
-	TextPos int    `json:"start_pos"`
+	TextPos int    `json:"text_pos"`
 }
 
 func (s *String) Pos() int { return s.TextPos }
@@ -364,7 +363,7 @@ func (*FString) ExprAST()    {}
 type ConstDecl struct {
 	Kind       string `json:"kind" unpack:""`
 	KeywordPos int    `json:"keyword_pos"`
-	Name       string `json:"name"`
+	Name       *ID    `json:"name"`
 	Expr       Expr   `json:"expr"`
 }
 
@@ -372,24 +371,24 @@ func (c *ConstDecl) Pos() int { return c.KeywordPos }
 func (c *ConstDecl) End() int { return c.Expr.End() }
 
 type FuncDecl struct {
-	Kind       string   `json:"kind" unpack:""`
-	KeywordPos int      `json:"keyword_pos"`
-	Name       string   `json:"name"`
-	Params     []string `json:"params"`
-	Expr       Expr     `json:"expr"`
-	Rparen     int      `json:"rparen"`
+	Kind       string `json:"kind" unpack:""`
+	KeywordPos int    `json:"keyword_pos"`
+	Name       *ID    `json:"name"`
+	Params     []*ID  `json:"params"`
+	Expr       Expr   `json:"expr"`
+	Rparen     int    `json:"rparen"`
 }
 
 func (f *FuncDecl) Pos() int { return f.KeywordPos }
 func (f *FuncDecl) End() int { return f.Rparen }
 
 type OpDecl struct {
-	Kind       string   `json:"kind" unpack:""`
-	KeywordPos int      `json:"keyword_pos"`
-	Name       string   `json:"name"`
-	Params     []string `json:"params"`
-	Body       Seq      `json:"body"`
-	Rparen     int      `json:"rparen"`
+	Kind       string `json:"kind" unpack:""`
+	KeywordPos int    `json:"keyword_pos"`
+	Name       *ID    `json:"name"`
+	Params     []*ID  `json:"params"`
+	Body       Seq    `json:"body"`
+	Rparen     int    `json:"rparen"`
 }
 
 func (o *OpDecl) Pos() int { return o.KeywordPos }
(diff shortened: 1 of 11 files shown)


compiler/ast/ast.go
```diff
@@ -398,7 +397,7 @@ func (o *OpDecl) End() int { return o.Rparen }
 type TypeDecl struct {
 	Kind       string      `json:"kind" unpack:""`
 	KeywordPos int         `json:"keyword_pos"`
-	Name       string      `json:"name"`
+	Name       Pattern     `json:"name"`
```
nwt, line 400: `Pattern` feels very wrong since this can't be a `Glob` or a `Regexp`. Are you using it because `String` and `QuotedString` both implement it?

mattnibs: I agree it is really weird. Yes I'm using it because String and QuotedString implement it. Open to suggestions.

mattnibs: I mean ideally this would just be an identifier but we are in need of a stop gap until that decision can be made.

nwt: (The `TypeDecl` rule in parser.peg could then go something like this.

```
TypeDecl
  = "type" _ id:(Identifier / QuotedIdentifier) __ "=" __ typ:Type {
      ...
    }

QuotedIdentifier
  = name:QuotedString {
      return &ast.ID{
          Kind: "ID",
          Name: name.(string),
          NamePos: c.pos.offset,
      }, nil
    }
```
)

compiler/parser/parser.peg
```diff
@@ -32,33 +32,33 @@ Decl
   = v:(ConstDecl / FuncDecl / OpDecl / TypeDecl) _ { return v, nil }
 
 ConstDecl
-  = "const" _ id:IdentifierName __ "=" __ expr:Expr {
+  = "const" _ id:Identifier __ "=" __ expr:Expr {
```
nwt, line 35: Nit: While we're here, our usual convention is to use the name of destination field as the label.

```suggestion
  = "const" _ name:Identifier __ "=" __ expr:Expr {
```

compiler/parser/parser.peg
```diff
@@ -32,33 +32,33 @@ Decl
...
-          Name: id.(string),
+          Name: id.(*ast.ID),
           Expr: expr.(ast.Expr),
           KeywordPos: c.pos.offset,
       }, nil
     }
 
 FuncDecl
-  = "func" _ id:IdentifierName __ "(" __ params:IdentifierNames? __ ")" __ ":" __ "(" __ expr:Expr __ ")" {
+  = "func" _ id:Identifier __ "(" __ params:Identifiers? __ ")" __ ":" __ "(" __ expr:Expr __ ")" {
```
nwt, line 45: Nit:

```suggestion
  = "func" _ name:Identifier __ "(" __ params:Identifiers? __ ")" __ ":" __ "(" __ expr:Expr __ ")" {
```

compiler/parser/parser.peg
```diff
@@ -32,33 +32,33 @@ Decl
...
+        Params: sliceOf[*ast.ID](params),
         Expr: expr.(ast.Expr),
         KeywordPos: c.pos.offset,
         Rparen: lastPos(c, ")"),
       }, nil
     }
 
 OpDecl
-  = "op" _ name:IdentifierName __ "(" __ params:IdentifierNames? __ ")" __ ":" __ "(" __ body:OpDeclBody __ ")" {
+  = "op" _ id:Identifier __ "(" __ params:Identifiers? __ ")" __ ":" __ "(" __ body:OpDeclBody __ ")" {
```
nwt, line 57: Nit:

```suggestion
  = "op" _ name:Identifier __ "(" __ params:Identifiers? __ ")" __ ":" __ "(" __ body:OpDeclBody __ ")" {
```

compiler/parser/parser.peg
```diff
@@ -70,15 +70,25 @@ OpDeclBody
   / seq:Seq { return seq, nil }
 
 TypeDecl
-  = "type" _ id:(IdentifierName / QuotedString) __ "=" __ typ:Type {
+  = "type" _ id:TypeDeclID __ "=" __ typ:Type {
```
nwt, line 73: Nit:

```suggestion
  = "type" _ name:TypeDeclID __ "=" __ typ:Type {
```

compiler/parser/parser.peg
```diff
@@ -791,11 +801,10 @@ Locals
     }
 
 LocalsAssignment
-  = name:IdentifierName opt:(__ "=" __ Expr)? {
+  = id:Identifier opt:(__ "=" __ Expr)? {
```
nwt, line 804: Nit:

```suggestion
  = name:Identifier opt:(__ "=" __ Expr)? {
```

compiler/semantic/scope.go
```diff
@@ -29,28 +30,28 @@ type entry struct {
 	order int
 }
 
-func (s *Scope) DefineVar(name string) error {
+func (s *Scope) DefineVar(id *ast.ID) error {
```
nwt, line 33: Nit: Use a consistent name for the `*ast.ID` parameter to this, `DefineAs`, and `DefineConst`. I like `name` since it makes the parameter's role clear but `id` is fine too.

zfmt/ast.go
```diff
@@ -64,7 +64,7 @@ func (c *canon) defs(defs []ast.Def, separator string) {
 }
 
 func (c *canon) def(d ast.Def) {
-	c.write(d.Name)
+	c.expr(d.Name, "")
```
nwt, line 67: Nit: `write` is clearer. (Same for all the other places below where `write` is changed to `expr`.)

```suggestion
	c.write(d.Name.Name, "")
```

mattnibs: You are probably right but I do find `d.Name.Name` to be really jarring to look at.

nwt: Agreed. We could address that (later) by renaming `ast.ID.Name` to `Text`.

compiler/ast/ast.go
mattnibs: The problem with this approach is you now have inaccurate node positioning for `QuotedIdentifier` given that it won't take into account the space of the quotes when finding the location of end.

mattnibs: I suppose the least lazy, most correct way would be to create a new interface that supports both ID and QuotedString. It's not great but neither is this feature IMHO.

nwt: > The problem with this approach is you now have inaccurate node positioning for QuotedIdentifier given that it won't take into account the space of the quotes when finding the location of end.

Could we say `NamePos: c.pos.offset+1` (instead of `c.pos.offset`) and get errors that look like this (where the tildes don't extend beneath the quotation marks)?

```
type "something bad" = int64
      ~~~~~~~~~~~~~
```

Seems fine to me if so.

mattnibs: I still hate it, but ok.

mattnibs: As long as we both can agree that this is a hack and not a rock solid solution, I'm ok with this.

nwt: We can agree.

And you're right that a new interface supporting both ID and QuotedString is the most correct way to solve this. I just don't think it's worth its weight.

nwt (approved):
:shipit:

PR #5145, opened 2024-06-19
lake api: Add nop logger if logger is nil

mattnibs:


nwt (approved):
Fine but I think lake.newRoot is a better place to do this.

PR #5146, opened 2024-06-20
Rework vam/expr.Compare.Eval to use generated code

nwt:
By generating code, we can handle comparisons for more Zed types (bytes and strings are added) with fewer lines of authored code.

Run "make generate" to run gencomparefuncs.go, which writes comparefuncs.go.

Mostly extracted from @mccanne's vam-stitch branch.


runtime/vam/expr/gencomparefuncs.go
```diff
@@ -0,0 +1,316 @@
...
+	}
+}
+
+const flat_flat = `
+	for k := uint32(0); k < n; k++ {
+		if l.Values[k] %s r.Values[k] {
+			out.Set(k)
+		}
+	}
+`
```
mattnibs, line 99: Maybe get rid of the extra line here so the generated file doesn't have the extra line?

runtime/vam/expr/gencomparefuncs.go
```diff
@@ -0,0 +1,316 @@
...
+
+const flat_idx = `
```
mattnibs, line 101: Any reason to not name these s/idx/dict ?

runtime/vam/expr/gencomparefuncs.go
```diff
@@ -0,0 +1,316 @@
...
+		case vector.FormConst:
+			return fmt.Sprintf(idx_const_s, pkg, op)
+		}
+	case vector.FormConst:
+		switch rform {
+		case vector.FormFlat:
+			return fmt.Sprintf(const_flat_s, pkg, op)
+		case vector.FormDict:
+			return fmt.Sprintf(const_idx_s, pkg, op)
+		}
```
mattnibs, line 260: For my own edification- there no such thing as having a const on both the lhs and rhs?

mattnibs (approved):

runtime/vam/expr/gencomparefuncs.go
nwt: Right, because in that case it's more efficient to return a vector.Const, and that's what vam/expr.compareConsts does.

runtime/vam/expr/gencomparefuncs.go
nwt: Done in 8135e285738f3be5e950ecd2ef8162904d4659a6.

PR #5149, opened 2024-06-26
Add vector arithmetic

nwt:
Implement addition, subtraction, multiplication, division, and modulo for numeric (unsigned integer, signed interger, and floating point) vectors as well as addition for string vectors.

We haven't yet worked out a way to write concise tests for vector operations so test coverage is minimal for now.


runtime/vam/expr/genarithfuncs.go
```diff
@@ -0,0 +1,133 @@
...
+	var ents strings.Builder
+	for _, op := range []string{"+", "-", "*", "/", "%"} {
+		for _, typ := range []string{"Int", "Uint", "Float", "String"} {
+			if typ == "Float" && op == "%" ||
+				typ == "String" && op != "+" {
+				continue
+			}
+			for lform := vector.Form(0); lform < 3; lform++ {
+				for rform := vector.Form(0); rform < 3; rform++ {
+					name := "arith" + opToAlpha[op] + typ + lform.String() + rform.String()
```
mattnibs, line 43: Your naming convention differs from your compare pr in that the generated functions are seperated by underscores (e.g., `cmp_EQ_Int_Flat_Flat`). I think you should pick one for consistency.

vector/int.go
```diff
@@ -18,6 +18,14 @@ func NewInt(typ zed.Type, values []int64, nulls *Bool) *Int {
 	return &Int{Typ: typ, Values: values, Nulls: nulls}
 }
 
+func NewIntEmpty(typ zed.Type, length uint32, nulls *Bool) *Int {
+	return NewInt(typ, make([]int64, 0, length), nulls)
+}
+
+func (i *Int) Append(v int64) {
+	i.Values = append(i.Values, v)
```
mattnibs, line 26: Do nulls need to be append to here?

runtime/vam/expr/genarithfuncs.go
nwt: I'm planning to remove the underscores from gencomparefuncs.go in a separate PR.

vector/int.go
nwt: No need. The model for vector operations whose output is a fixed-size primitive type (like `Int`, `Uint`, `Float`, and `Bool`) is that we compute a value for every slot in the output vector regardless of nulls in the input vectors and separately compute an output nulls vector (the `Nulls *Bool` field present in most of the vector types) from the input nulls vectors.

This PR doesn't include that nulls handling and neither does #5146 or any of the existing code under runtime/vam/expr. But it's coming.

mattnibs (approved):

PR #5161, opened 2024-07-01
parser: Name collisions with uniq, head, top ops

### LINKED ISSUE #5160: possible parser error if user-defined op or func starts with an internal operator name
v1.16.0

```
❯ zq -z 'func headfunc(): ( 1 ) headfunc()'  
 - a file could not be found with the name "func headfunc(): ( 1..."
 - the argument could not be compiled as a valid Zed query:
   error parsing Zed at line 1, column 28:
   func headfunc(): ( 1 ) headfunc()
                          === ^ ===

❯ zq -z 'func _headfunc(): ( 1 ) _headfunc()' 
1
```

```
❯ echo '{}' | zq -z 'op headop(): ( pass ) headop()' - 
zq: error parsing Zed at line 1, column 27:
op headop(): ( pass ) headop()
                      === ^ ===

❯ echo '{}' | zq -z 'op _headop(): ( pass ) _headop()' -
{}
```

It seems the parser isn't grabbing the entire name of the user-defined op as one thing.

mattnibs:
This commit fix an issue with the parser where yield-less calling a functions named uniq, head, top would result in a parser error.

Closes #5160

### DIFF
--- a/compiler/parser/parser.go
+++ b/compiler/parser/parser.go
@@ -3060,11 +3060,41 @@ var g = &grammar{
 					&actionExpr{
 						pos: position{line: 445, col: 5, offset: 11043},
 						run: (*parser).callonHeadOp10,
-						expr: &litMatcher{
-							pos:        position{line: 445, col: 5, offset: 11043},
-							val:        "head",
-							ignoreCase: false,
-							want:       "\"head\"",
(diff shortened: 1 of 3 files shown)


nwt (approved):
Does `pass` need the same treatment?

PR #5171, opened 2024-07-05
merge: Batches not getting Unreferenced

mattnibs:
This commit fixes an issue where the merge operator was not properly unreferencing batches.

### DIFF
--- a/runtime/sam/op/merge/merge.go
+++ b/runtime/sam/op/merge/merge.go
@@ -30,7 +30,8 @@ type Op struct {
 	// The head-of-line (hol) queue is maintained as a min-heap on cmp of
 	// hol.vals[0] (see Less) so that the next Read always returns
 	// hol[0].vals[0].
-	hol []*puller
+	hol   []*puller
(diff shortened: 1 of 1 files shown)


philrz:
Since this PR held the promise of reducing the memory usage during compaction, I re-ran the tests that had previously flagged the problem and posted a [write-up](https://github.com/brimdata/zed-perf-testing/blob/main/object-size/README.md#merge-unref). The tl;dr is that it does indeed reduce memory during compaction by about 3.5x at the smaller object sizes, though memory consumption is still higher during the compaction phase than the load phase at these smaller object sizes.

Also just to test for general stability, I did a run through all the Autoperf tests using this branch at commit `6f67560`. I can testify that it survived all the tests and I reviewed the summary dashboard and rather than assemble a bunch of screenshots that effectively show flat lines I can also say that it didn't cause any statistically significant deltas in the macro trends compared to other recent test runs at tip of `main`.

In conclusion, it's a functional 👍 for me.

Issue #5168, opened 2024-07-04
casting complex type is order dependent when docs say different


chrismo:
The [cast docs](https://zed.brimdata.io/docs/v1.16.0/language/functions/cast) mention:
> If val is a record (or if any of its nested value is a record):
> ...
> - fields are matched by name and are order independent and the input order is retained.

```
❯ zq -version
Version: v1.16.0

❯ zq 'type foo={a:string,b:int64} yield {} | fill(this,<foo>) | cast(this,<foo>)'
{a:null(string),b:null(int64)}(=foo)

❯ zq 'type foo={a:string,b:int64} yield {b:1} | fill(this,<foo>) | cast(this,<foo>)'
{b:1,a:null(string)}

❯ zq 'type foo={a:string,b:int64} yield {a:"x"} | fill(this,<foo>) | cast(this,<foo>)'
{a:"x",b:null(int64)}(=foo)
```

The middle example shows that the casting fails (returns the original input) because the order of fields isn't correct, but the docs say that "fields are matched by name and are order independent".

As a result of looking for similar issues, I discovered the `shape` function, which is what I can use going forward and not be impacted by this.

philrz:
@chrismo: Actually, I think the middle example is behaving in a way that's aligned with the docs, though your experience shows there's room for improvement.

Let's start from the base of your middle example.

```
$ zq 'type foo={a:string,b:int64} yield {b:1}'
{b:1}
```

So that `{b:1}` becomes the `this` at the input to `fill()`. Adding that on:

```
$ zq 'type foo={a:string,b:int64} yield {b:1} | fill(this,<foo>)'
{b:1,a:null(string)}
```

That's is in line with where the [`fill` docs](https://zed.brimdata.io/docs/language/functions/fill) say:

> The _fill_ function adds to the input record `val` any fields that are present in the output type `t` but not in the input.

What's implied here (but should likely be called out explicitly) is that the order of such added fields is undefined. There's a separate [`order` function](https://zed.brimdata.io/docs/language/functions/order) that would take care of this.

```
$ zq 'type foo={a:string,b:int64} yield {b:1} | fill(this,<foo>) | order(this,<foo>)'
{a:null(string),b:1}
```

Putting `order` aside for now and getting back to your original middle example, now that `fill`'s behavior is rationalized, you can probably see that:

```
$ zq 'type foo={a:string,b:int64} yield {b:1} | fill(this,<foo>) | cast(this,<foo>)'
{b:1,a:null(string)}
[...]

PR #5172, opened 2024-07-07
Improve docs for shape and a la carte shaping functions

philrz:
When it comes to working with records, we've had some internal consensus that users will probably prefer `shape` most of the time whereas the "a la carte" functions `cast`, `fill`, and `order` are probably going to see more limited use. Addressing the questions posed by the community user that opened #5168 made me realize this is not really called out in the docs, so I'm attempting to do that here, along with a couple other bits of polish I saw in the pages along the way.

### DIFF
--- a/docs/language/functions/fill.md
+++ b/docs/language/functions/fill.md
@@ -18,7 +18,7 @@ you want to be sure that all fields in a schema are present in a record.
 
 The order of newly added fields relative to fields already present is
 undefined.  If maintaining relative order is important, consider applying the
-[`order`](./order.md) or [`shape`](./shape) function.
+[`order`](./order.md) or [`shape`](./shape.md) function.
 
 If `val` is not a record, it is returned unmodified.
 
--- a/docs/language/functions/fill.md
+++ b/docs/language/functions/fill.md
@@ -11,14 +11,12 @@ fill(val: any, t: type) -> any
 ### Description
 
 The _fill_ function adds to the input record `val` any fields that are
-present in the output type `t` but not in the input.
+present in the output type `t` but not in the input.  Such fields are added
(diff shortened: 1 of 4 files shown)


docs/language/functions/fill.md
```diff
@@ -16,8 +16,18 @@ present in the output type `t` but not in the input.
 Filled fields are added with a `null` value.  Filling is useful when
 you want to be sure that all fields in a schema are present in a record.
 
+The order of newly added fields relative to fields already present is
+undefined.  If maintaining relative order is important, consider applying the
+[`order`](./order.md) or [`shape`](./shape) function.
```
philrz, line 21: I stuck with "undefined" since I don't know the code. If we think there's something more detailed we could say that's not too complicated I'd be happy to set more explicit expectations (e.g., if/when it can be expected to append the new fields).

docs/language/functions/fill.md
```diff
@@ -16,8 +16,18 @@ present in the output type `t` but not in the input.
 Filled fields are added with a `null` value.  Filling is useful when
 you want to be sure that all fields in a schema are present in a record.
 
+The order of newly added fields relative to fields already present is
+undefined.  If maintaining relative order is important, consider applying the
+[`order`](./order.md) or [`shape`](./shape.md) function.
```
nwt, line 21: Fields are added in the order they appear in the input record.

nwt (approved):

docs/language/functions/fill.md
philrz: After seeing @nwt's comments here we had a quick offline chat and agreed on some text to clarify where `fill` adds the fields rather than the "undefined" I had before. I've pushed a commit with that change.

PR #5185, opened 2024-07-17
csv reader: fix EOF on quotes

### LINKED ISSUE #4732: CSV reader error when file doesn't end in newline: bare " in non-quoted-field
Repro is with Zed tagged `v1.9.0`.

A user in a [public Slack thread](https://brimdata.slack.com/archives/CTSMAK6G7/p1690568261979059) observed the following when attempting to read the attached test data [chatgpt-vcs-2.csv](https://github.com/brimdata/zed/files/12197294/chatgpt-vcs-2.csv).

```
$ zq -version
Version: v1.9.0

$ zq -i csv 'count()' chatgpt-vcs-2.csv 
chatgpt-vcs-2.csv: parse error on line 630, column 27: bare " in non-quoted-field
```

Closer inspection revealed the problem was triggered because the file lacks a newline on the last line. Therefore this works around the problem:

```
$ echo >> chatgpt-vcs-2.csv 

$ zq -i csv 'count()' chatgpt-vcs-2.csv 
629(uint64)
```

This can surely be added to the list of CSV dialect headaches for which there may not be one 100% correct way to handle, but it seems like a simple enough corner cases that hopefully we can handle it better. Could we just detect this case and pretend the newline was there even if it isn't? If not, could we at least have a better error message so has a better chance of making the fix themselves?

mattnibs:
This commit fixes an with the csv reader where and EOF occurring on a quoted field would cause a parsing error.

Closes #4732

### DIFF
--- a/zio/csvio/preprocess.go
+++ b/zio/csvio/preprocess.go
@@ -73,15 +73,21 @@ func (p *preprocess) parseField() ([]byte, error) {
 		}
 		if c == '"' {
 			hasstr = true
-			str, err := p.parseString()
-			p.scratch = append(p.scratch, str...)
-			if err != nil {
+			var s []byte
+			s, err = p.parseString()
+			p.scratch = append(p.scratch, s...)
+			if err == nil {
+				continue
+			}
+			if err != io.EOF {
 				return p.scratch, err
 			}
-			continue
 		}
-		if rune(c) == p.delimiter || c == '\n' {
-			ending := []byte{c}
+		if rune(c) == p.delimiter || c == '\n' || err == io.EOF {
+			var ending []byte
+			if err != io.EOF {
+				ending = []byte{c}
+			}
 			if hasstr {
 				// If field had quotes wrap entire field in quotes.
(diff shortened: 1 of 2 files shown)


philrz:
The branch seems to fix the two user examples from #4732 so it's a functional 👍 for me!

```
$ zq -version
Version: v1.16.0-16-gf7d8e511

$ zq -i csv 'count()' chatgpt-vcs-2.csv
629(uint64)

$ echo -e -n "foo,bar\n\"foo\",\"bar\"" | zq -i csv -z 'yield this' -
{foo:"foo",bar:"bar"}
```

Compared to baseline on `main`:

```
$ zq -version
Version: v1.17.0

$ zq -i csv 'count()' chatgpt-vcs-2.csv
chatgpt-vcs-2.csv: parse error on line 630, column 27: bare " in non-quoted-field

$ echo -e -n "foo,bar\n\"foo\",\"bar\"" | zq -i csv -z 'yield this' -
stdio:stdin: parse error on line 2, column 10: bare " in non-quoted-field
```

nwt (approved):

Issue #4271, opened 2022-12-15
"Live" data import with Fluentd


philrz:
In an exercise similar to #3151 with Logstash, I've run a test with external agent [Fluentd](https://www.fluentd.org/) to push "live" data continuously to a Zed lake.

tl;dr - As Fluentd is a more modern tool than Logstash, in the end it did indeed seem to play more nicely with Zed in a minimal out-of-the-box config.

The test was performed with a Zed lake running behind Zui insiders [0.30.1-142](https://github.com/brimdata/zui-insiders/releases/tag/v0.30.1-142) (which is Zed commit `101d358`), Fluentd `1.15.3`, and Zeek `5.1.1` (with the [json-streaming-logs](https://github.com/corelight/json-streaming-logs) package) sniffing the coincidental live local traffic on my wireless interface during a typical work-from-home day.

The `fluentd.conf` I ultimately settled on:

```
<source>
  @type tail
  path /usr/local/zeek-5.1.1/logs/current/json_streaming_*
  follow_inodes true
  pos_file /Users/phil/work/fluentd/fluentd.pos
  tag zeek
  <parse>
    @type json
  </parse>
</source>

<match zeek>
  @type http
  endpoint http://127.0.0.1:9988/pool/zeek/branch/main
  content_type application/json

  <format>
    @type json
  </format>
</match>
```

Then to start Fluentd:

```
$ ulimit -n 65535
$ fluentd -c fluentd.conf 
```

The `ulimit` guidance came from the ["Before Installation" docs](https://docs.fluentd.org/installation/before-install) and did end up being significant, as the default setting on my Macbook was `256` and in an early run Fluentd did indeed complain of running out of file descriptors.

The only minor speedbump I hit along the way was needing to specify `content_type application/json`. At first after reading the relevant [docs](https://docs.fluentd.org/output/http#content_type) I thought the defaults would work since the doc explains that `json_array` is `false` by default and hence it posts in NDJSON (which `zed serve` accepts) and sets Content Type to `application/x-ndjson`. However, this caused an HTTP 400 error because the Zed backend handles both regular JSON and NDJSON through the same reader and this is wired up only through Content Type `application/json`, so I had to set that explicit `content_type`.

[...]

PR #5190, opened 2024-07-21
Add an integrations doc for Fluentd

philrz:
## What's Changing

This PR adds a new article to the "Integrations" section of the Zed docs site that shows example configuration using [Fluentd](https://www.fluentd.org/) for "live" log ingest of Zeek logs to a Zed lake.

## Why

Features like compaction have made it such that the Zed lake should be suited for such use cases, at least at modest scale. A community user found open issue #4271 that had a preliminary configuration and they expressed an interest in using it in a Zeek-related project. That motivated me to more thoroughly test the configuration using current software versions and commit it to a proper article. Once on the docs site, I could potentially promote this content within the Zeek community and perhaps find interest from more users seeking alternatives to their current log storage solutions.

## Discussion

While I'd be happy if a reviewer actually ran through the steps in the article, if that's not palatable, since I expect I'll likely be providing front-line support for any users attempting to follow it, I'd be content with reviews just for style and correctness of how I describe the relevant Zed components.

### DIFF
--- a/docs/integrations/fluentd.md
+++ b/docs/integrations/fluentd.md
@@ -1,0 +1,413 @@
+---
+sidebar_position: 3
+sidebar_label: Fluentd
+---
+
+# Fluentd
+
+The [Fluentd](https://www.fluentd.org/) open source data collector can be used
+to push log data to a [Zed lake](../commands/zed.md) in a continuous manner.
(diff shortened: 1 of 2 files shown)


mattnibs (approved):
This is great!

PR #5191, opened 2024-07-22
output operator

mattnibs:
Add output operator which allows users to set the name of the resultant channel. Channels are no longer identified by an integer but instead with a string- if a channel doesn't not have an output it will be assigned to channel "main".

### DIFF
--- a/compiler/job.go
+++ b/compiler/job.go
@@ -23,7 +23,7 @@ type Job struct {
 	rctx      *runtime.Context
 	builder   *kernel.Builder
 	optimizer *optimizer.Optimizer
-	outputs   []zbuf.Puller
+	outputs   map[string]zbuf.Puller
 	puller    zbuf.Puller
 	entry     dag.Seq
 }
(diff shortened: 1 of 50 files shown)


compiler/job.go
```diff
@@ -113,6 +113,9 @@ func (j *Job) Build(readers ...zio.Reader) error {
 	if err != nil {
 		return err
 	}
+	if outputs == nil || len(outputs) == 0 {
+		panic("system error: no outputs")
+	}
```
nwt, line 118: Nits: I don't think we need this check but if you want to add it, I think
1. a better place for it would be kernel/Builder.Build,
2. `if len(outputs) == 0` is sufficient, and
3. it should return an error instead of panicking.

mattnibs: It's gone.

nwt (approved):
Before you merge, have we verified that this doesn't break Zui?

mattnibs:
@nwt no I will test this before mergin'

philrz (approved):

PR #5193, opened 2024-07-24
Add vector.View

nwt:
A vector.View provides a view of another vector's elements (usually a subset of them).  The type isn't used here but will be later when we add support for comparisons and arithmetic involving vector.Union and vector.Variant (aka stitch).

Mostly extracted from @mccanne's vam-stitch branch.


mccanne (approved):
Looks great!

PR #5194, opened 2024-07-24
Handle unions and variants in vector arithmetic and comparison

nwt:
Add vam/expr.stitch and call it from vam/expr.Arith.eval and Compare.eval so arithmetic and comparison operators can handle vector.Union and vector.Variant operands.

The stitch function takes two same-length vector arguments along with a two-argument callback function.  If either vector is a union or variant, stitch applies the callback to the vectors and returns a result vector. At a high level, it does so via the following steps.

1. Identify a union or variant argument.

2. Split that argument into a set of single-type (i.e., neither union nor variant) subvectors.

3. Split the other argument into a corresponding set of vector.Views.

4. Apply the callback in turn to each single-type subvector and its corresponding vector.View and collect the result vectors.

5. Combine the result vectors into a vector.Variant and return it.

The stitch name and algorithm come from @mccanne's vam-stitch branch.

Depends on #5193.


mccanne (approved):
cool... we'll generalize rip-and-stitch in a subsequent PR

PR #5196, opened 2024-07-30
Debug operator

### LINKED ISSUE #5181: Something like jq's "debug"
## tl;dr

In a recent [community Slack thread](https://brimdata.slack.com/archives/CTSMAK6G7/p1719613079600409), a user asked of `zq`:

> is there a way to do `jq`’s [debug](https://jqlang.github.io/jq/manual/#debug)?

Zed doesn't currently have a 1-to-1 equivalent to this, but a lot of the building blocks are there. To benefit users that come to Zed from tools like `jq`, it does seem like a worthwhile exercise to consider what it might take to provide something similar and rationalize where we think it's appropriate to take a different approach.

## Details

At the time this issue is being filed, Zed is at commit 6e61673.

The [`jq` docs](https://jqlang.github.io/jq/manual/#debug) don't actually show an example of their most simple `debug` usage with no parameters, but we can trim their more advanced example to:

```
$ jq -n '1 as $x | 2 | debug | (.+1)'
["DEBUG:",2]
3
```

Here's something similar in `zq`.

```
$ cat debug.zed 
op debug(): (
  fork (
    => yield error(["DEBUG:", this]) | yield {k:0, v:this}
    => pass | yield {k:1, v:this}
  )
  | sort k
  | yield v
)

$ zq -version
Version: v1.16.0-15-g6e61673c

$ zq -I debug.zed 'const x=1 yield 2 | debug() | yield this+1'
error(["DEBUG:",2])(error([(int64,string)]))
3
```

`jq` takes the approach of sending the error outputs to stderr, which means there's ways at the shell to redirect or filter them out. `zq` doesn't currently take this approach, which is why I chose to wrap the debug messages in Zed `error` values since there's filtering available within the language itself, e.g., making the following the equivalent of tacking `2> /dev/null` onto the end of the `jq` command line above.

```
$ zq -I debug.zed 'const x=1 yield 2 | debug() | yield this+1 | !is_error(this)'
3
```

Of course, this begs the question of whether this close coupling between the tool and the shell is _ideal_. A proficient `jq` user that later comes to `zq` might indeed have the muscle memory that would make this seem appealing. However, even if this were enabled specifically with `zq` to benefit those shell-centric `jq`-like users, what would happen in other places where the Zed language can be used? For instance, over the Zed lake API there's no concept of "stderr", so where would the debug output go in that case? In an API-dependent app like Zui, we could perhaps imagine some likely UI, e.g., perhaps a separate window or panel could separate the debug output from query output, or perhaps it could be interleaved with the query output in a way that makes stand out visually and easy to filter. So this would all require design insight.

Moving on to the more advanced example as it appears in the `jq` docs:

```
$ jq -n '1 as $x | 2 | debug("Entering function foo with $x == \($x)", .) | (.+1)'
["DEBUG:","Entering function foo with $x == 1"]
["DEBUG:",2]
3
```

I have what I believe would be a working enhancement that once again approximates what `jq` is doing, taking the approach of passing the multiple "message" parameters as elements of an array. However, I hit what appears to be a bug #5180.

```
$ cat debug-msgs.zed 
op debug(msgs): (
  fork (
    => over msgs => (
      yield error(["DEBUG:", this])
      | yield {k:count(), v:this}
    )
    => pass | yield {k:Inf, v:this}
  )
  | sort k
  | yield v
)

$ zq -I debug-msgs.zed 'const x=1 yield 2 | debug([f"Entering function foo with $x == {x}", this]) | yield this+1'
error(["DEBUG:","Entering function foo with $x == 1"])
error(["DEBUG:",error("missing")])(error([(string,error(string))]))
3
```

I'll plan to pick up this effort once that's addressed.

mattnibs:
This commit introduces the debug operator- an operator intending to help debugging complex queries. When running the debug operator using zq or zed all encounterd values at the point of the operator are written to stderr.

Closes #5181

### DIFF
--- a/api/queryio/client.go
+++ b/api/queryio/client.go
@@ -20,16 +20,15 @@ type scanner struct {
 	progress zbuf.Progress
 }
 
-func NewScanner(ctx context.Context, rc io.ReadCloser) zbuf.Scanner {
+func NewScanner(ctx context.Context, rc io.ReadCloser) (zbuf.Scanner, error) {
 	s, err := zngio.NewReader(zed.NewContext(), rc).NewScanner(ctx, nil)
 	if err != nil {
-		// XXX This shouldn't happen since we don't have a filter.
-		panic(err)
+		return nil, err
 	}
 	return &scanner{
 		scanner: s,
 		closer:  rc,
-	}
+	}, nil
 }
 
 func (s *scanner) Progress() zbuf.Progress {
@@ -49,7 +48,7 @@ again:
 	if !ok {
 		return nil, err
 	}
-	v, err := s.marshalControl(zctrl)
+	v, err := marshalControl(zctrl)
 	if err != nil {
 		return nil, err
 	}
@@ -70,7 +69,7 @@ again:
 	}
 }
 
-func (s *scanner) marshalControl(zctrl *zbuf.Control) (any, error) {
+func marshalControl(zctrl *zbuf.Control) (any, error) {
 	ctrl, ok := zctrl.Message.(*zngio.Control)
 	if !ok {
 		return nil, fmt.Errorf("unknown control type: %T", zctrl.Message)
@@ -82,11 +81,11 @@ func (s *scanner) marshalControl(zctrl *zbuf.Control) (any, error) {
 	defer arena.Unref()
 	value, err := zson.ParseValue(zed.NewContext(), arena, string(ctrl.Bytes))
 	if err != nil {
-		return nil, fmt.Errorf("unable to parse Zed control message: %w (%s)", err, string(ctrl.Bytes))
+		return nil, fmt.Errorf("unable to parse Zed control message: %w (%s)", err, ctrl.Bytes)
 	}
 	var v interface{}
 	if err := unmarshaler.Unmarshal(value, &v); err != nil {
-		return nil, fmt.Errorf("unable to unmarshal Zed control message: %w (%s)", err, string(ctrl.Bytes))
+		return nil, fmt.Errorf("unable to unmarshal Zed control message: %w (%s)", err, ctrl.Bytes)
 	}
 	return v, nil
 }
--- a/api/queryio/client.go
+++ b/api/queryio/client.go
@@ -1,6 +1,7 @@
 package queryio
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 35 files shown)


cmd/zed/internal/lakemanage/scan.go
```diff
@@ -66,20 +66,20 @@ from %q@%q:objects
...
 	query := fmt.Sprintf(iteratorQuery, head.Pool, head.Branch, head.Pool, head.Branch)
-	r, err := lake.Query(ctx, nil, query)
+	q, err := lake.Query(ctx, nil, query)
 	if err != nil {
 		return nil, err
 	}
 	arena := zed.NewArena()
 	return &objectIterator{
-		reader:      r,
+		reader:      zbuf.ScannerReader(q),
```
nwt, line 82: Instead of adding zbuf.ScannerReader, how about

```suggestion
		puller:      q,
		reader:      zbuf.PullerReader(q),
```

plus

```go
func (r *objectIterator) close() {
	return r.puller.Pull(true)
}
```

plus

```go
	defer q.Pull(true)
	reader := zbuf.PullerReader(q)
```

in docs/library/go.md?

zbuf/mux.go
```diff
@@ -0,0 +1,58 @@
+package zbuf
+
+import (
+	"github.com/brimdata/zed"
+	"github.com/brimdata/zed/zio"
+)
+
+func Wrap(label string, batch Batch) Batch {
```
nwt, line 8: There's already a function in this package called WrapBatch so this really wants another name. Maybe Label?

mattnibs: Yeah that's confusing Label/Unlabel sounds good.

cmd/zed/internal/lakemanage/scan.go
mattnibs: sure

docs/language/operators/debug.md
```diff
@@ -0,0 +1,27 @@
...
+```
+### Description
+
+The `debug` operator write the value of `expr` to the debug channel. If no
+`expr` is provided `this` is written to the debug channel. If the query is
+ran on command line via `zq` or `zed` all output written to the debug
+channel is displayed on stderr.
+
+The `debug` operator is useful to view intermediate values as part of
+workflow when debugging complex zed query.
```
philrz, line 18:
```suggestion
The `debug` operator writes the value of `expr` to the debug channel. If no
`expr` is provided, `this` is written to the debug channel. If the query is
run on the command line via `zq` or `zed` all output written to the debug
channel is displayed on stderr.

The `debug` operator is useful to view intermediate values when debugging a
complex Zed query.
```

mattnibs: Much better, thanks

docs/language/operators/debug.md
```diff
@@ -0,0 +1,27 @@
...
+
+### Examples
+
+The following query:
+```
+echo '"foo"' | zq -z 'debug f"debug: {this}" | yield f"{this}_bar"' -
+```
+will display `"debug: foo"` on stderr whereas `"foo_bar"` will display
+on stdout.
```
philrz, line 27:
````suggestion
The following query uses expressions containing [f-strings](../expressions.md#formatted-string-literals)
to display `"debug: foo"` on stderr whereas `"foo_bar"` will display
on stdout.

```
echo '"foo"' | zq -z 'debug f"debug: {this}" | yield f"{this}_bar"' -
```
````

**Review by philrz — APPROVED** [review]

File: zfmt/dag.go [hunk 5]
```diff
@@ -302,6 +302,22 @@ func (c *canonDAG) op(p dag.Op) {
...
+			c.write("=>")
+			c.open()
+			c.head = true
+			c.seq(seq)
+			c.close()
+		}
+		c.close()
+		c.ret()
+		c.flush()
+		c.write(")")
```
nwt, line 320: Test this.

zfmt/ast.go
```diff
@@ -630,6 +630,13 @@ func (c *canon) op(p ast.Op) {
 	case *ast.Output:
 		c.next()
 		c.write("output %s", p.Name.Name)
+	case *ast.Debug:
+		c.next()
+		c.write("debug")
+		if p.Expr != nil {
+			c.write(" ")
+			c.expr(p.Expr, "")
+		}
```
nwt, line 639: Test this.

api/queryio/client.go
```diff
@@ -12,58 +13,80 @@ import (
...
+		return nil, fmt.Errorf("unknown control type: %T", zctrl.Message)
+	}
+	if ctrl.Format != zngio.ControlFormatZSON {
+		return nil, fmt.Errorf("unsupported app encoding: %v", ctrl.Format)
+	}
+	arena := zed.NewArena()
+	defer arena.Unref()
+	value, err := zson.ParseValue(zed.NewContext(), arena, string(ctrl.Bytes))
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse Zed control message: %w (%s)", err, string(ctrl.Bytes))
```
nwt, line 85: Nit: `%s` formats `[]byte` as a string.
```suggestion
		return nil, fmt.Errorf("unable to parse Zed control message: %w (%s)", err, ctrl.Bytes)
```

api/queryio/client.go
```diff
@@ -12,58 +13,80 @@ import (
...
+	}
+	var v interface{}
+	if err := unmarshaler.Unmarshal(value, &v); err != nil {
+		return nil, fmt.Errorf("unable to unmarshal Zed control message: %w (%s)", err, string(ctrl.Bytes))
```
nwt, line 89: Nit: `%s` formats `[]byte` as a string.
```suggestion
		return nil, fmt.Errorf("unable to unmarshal Zed control message: %w (%s)", err, ctrl.Bytes)
```

docs/libraries/go.md
```diff
@@ -134,11 +135,12 @@ func main() {
 	if err != nil {
 		log.Fatalln(err)
 	}
-	reader, err := lake.Query(ctx, nil, "from Demo")
+	q, err := lake.Query(ctx, nil, "from Demo")
 	if err != nil {
 		log.Fatalln(err)
 	}
-	defer reader.Close()
+    defer q.Pull(true)
```
nwt, line 142: Tab, not spaces.
```suggestion
	defer q.Pull(true)
```

nwt (approved):
@philrz and I talked about extending the Zed syntax so you can specify a channel other than "debug". We both think that's a good idea since Zui will eventually be able to display channels separately but we can add it later.

mattnibs:
> @philrz and I talked about extending the Zed syntax so you can specify a channel other than "debug". We both think that's a good idea since Zui will eventually be able to display channels separately but we can add it later.

Is this different than using the output operator?

nwt:
> > @philrz and I talked about extending the Zed syntax so you can specify a channel other than "debug". We both think that's a good idea since Zui will eventually be able to display channels separately but we can add it later.
> 
> Is this different than using the output operator?

Yes, in two regards. First, it includes an expression (e.g., `debug some_channel some.path`). And second, it doesn't have `output`'s "read to EOS" behavior.

PR #5197, opened 2024-07-30
Add strftime function

mattnibs:


runtime/sam/expr/function/time.go
```diff
@@ -46,3 +49,49 @@ func (b *Bucket) Call(ectx expr.Context, args []zed.Value) zed.Value {
...
+	strftimeSet = strftime.NewSpecificationSet()
+	strftimeSet.Delete('n') // newline
+	strftimeSet.Delete('R') // equivalent to %H:%M
+	strftimeSet.Delete('r') // equivalent to %I:%M:%S %p
+	strftimeSet.Delete('t') // tab
+	strftimeSet.Delete('T') // equivalent to %H:%M:%S
+	strftimeSet.Delete('v') // equivalent to %e-%b-%Y
+	strftimeSet.Delete('V') // the week number of the year (Monday as the first day of the week) as a decimal number (01-53)
+	strftimeSet.Delete('z') // timezone
+	strftimeSet.Delete('Z') // timezone
```
nwt, line 65: I think we should keep these since they are all in the POSIX strftime specification.

runtime/sam/expr/function/time.go
```diff
@@ -46,3 +49,49 @@ func (b *Bucket) Call(ectx expr.Context, args []zed.Value) zed.Value {
...
+		return s.zctx.WrapError(ectx.Arena(), "strftime: string value required for format arg", formatArg)
+	}
+	if formatArg.IsNull() {
+		return zed.NullString
+	}
+	format := formatArg.AsString()
+	v, ok := coerce.ToInt(args[1])
+	if !ok {
+		return s.zctx.WrapError(ectx.Arena(), "strftime: time value required for time arg", args[1])
+	}
```
nwt, line 87: Using coerce.ToInt here results in behavior that's inconsistent with both this error message and the documentation in docs/language/functions/strftime.md.

runtime/sam/expr/function/ztests/strftime.yaml
```diff
@@ -0,0 +1,6 @@
+zed: 'strftime("%Y-%a", this)'
+
+input: '2024-07-30T06:15:01.062681Z'
+
+output: |
+  "2024-Tue"
```
nwt, line 6: Test error handling too.

docs/language/functions/strftime.md
```diff
@@ -0,0 +1,59 @@
...
+strftime(format: string, t: time) -> string
+```
+
+### Description
+The _strftime_ function returns a string represenation of time `t`
+as specified by the provided string `format`. `format` is a string
+containing format directives that dictate how the time string is
+formatted.
+
+The directives can be used:
```
nwt, line 16:
```suggestion
These directives are supported:
```

runtime/sam/expr/function/time.go
mattnibs: ok

mattnibs: I was weary about allowing timezone since we don't really support timezones with zed time, but I suppose we can allow timezones and it will always be `GMT`.

runtime/sam/expr/function/time.go
```diff
@@ -46,3 +49,34 @@ func (b *Bucket) Call(ectx expr.Context, args []zed.Value) zed.Value {
...
+	if formatArg.IsNull() {
+		return zed.NullString
+	}
+	format := formatArg.AsString()
+	if zed.TypeUnder(timeArg.Type()) != zed.TypeTime {
+		return s.zctx.WrapError(ectx.Arena(), "strftime: time value required for time arg", args[1])
+	}
+	t := timeArg.AsTime()
+	if s.format != format {
```
nwt, line 73: Nit: No need for Strftime.format.
```suggestion
	if s.formatter == nil || s.formatter.Pattern() != format {
```

runtime/sam/expr/function/ztests/strftime.yaml
```diff
@@ -0,0 +1,11 @@
+zed: 'strftime(f, v)'
+
+input: |
+  {f:"%Y-%a",v:2024-07-30T06:15:01.062681Z}
+  {f:1,v:2024-07-30T06:15:01.062681Z}
```
nwt, line 5: Nit: Test a malformed format string.
```suggestion
  {f:1,v:2024-07-30T06:15:01.062681Z}
  {f:"%",v:2024-07-30T06:15:01.062681Z}
```

nwt (approved):

PR #5174, opened 2024-07-09
Support lift optimization for dag.Fork

mattnibs:
Also fixes bug where the lift optimization would not get applied on eligible sequences with a length of 2.

Closes #4559, closes #4503, closes #5070

PR #5199, opened 2024-08-02
Fix flakiness in runtime/ztests/issue-4013.yaml

nwt:
This test went flaky with #5174 because the optimizer began lifting the head operator into the legs of the fork operator, causing the pass leg to stop pulling until EOS.


nwt:
Thanks @mattnibs for figuring out that this flakiness began with #5174!

mattnibs (approved):

PR #5204, opened 2024-08-05
Fix strftime docs and add another example

philrz:
While trying out the new `strftime` function added in #5197 I found a couple small errors to fix, and while I was at it I figured it could use another example.


docs/language/functions/README.md
```diff
@@ -59,6 +59,7 @@ Zed's [primitive types](../../formats/zed.md#1-primitive-types), e.g.,
 * [rune_len](rune_len.md) - length of a string in Unicode code points
 * [shape](shape.md) - apply cast, fill, and order
 * [split](split.md) - slice a string into an array of strings
+* [strftime](strftime.md) - format time values
 * [sqrt](sqrt.md) - square root of a number
```
nwt, line 63:
```suggestion
* [sqrt](sqrt.md) - square root of a number
* [strftime](strftime.md) - format time values
```

nwt (approved):

PR #5214, opened 2024-08-10
Simplify vam/expr.stitchLHS

nwt:


nwt:
stitchLHS is going away.

PR #5216, opened 2024-08-12
Add summarize docs for "by" without an aggregate function

philrz:
A recent [community Slack thread](https://brimdata.slack.com/archives/CTSMAK6G7/p1722612689782589?thread_ts=1722611911.071209&cid=CTSMAK6G7) made me realize we were lacking usage docs and an example for `by` in the absence of an aggregate function.

While I was making this change, I saw that the existing `by` coverage didn't seem to show that multiple keys could be specified, so I've tried to address that as well.

### DIFF
--- a/docs/language/operators/summarize.md
+++ b/docs/language/operators/summarize.md
@@ -5,14 +5,20 @@
 ### Synopsis
 
 ```
-[summarize] [<field>:=]<agg> [where <expr>][, [<field>:=]<agg> [where <expr>] ...] [by [<field>][:=<expr>] ...]
+[summarize] [<field>:=]<agg> [where <expr>][, [<field>:=]<agg> [where <expr>] ...] [by [<field>][:=<expr>][, [<field>][:=<expr>]] ...]
+[summarize] by [<field>][:=<expr>][, [<field>][:=<expr>] ...]
 ```
 ### Description
 
-The `summarize` operator consumes all of its input, applies an [aggregate function](../aggregates/README.md)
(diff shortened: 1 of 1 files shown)


docs/language/operators/summarize.md
```diff
@@ -5,14 +5,23 @@
 ### Synopsis
 
 ```
-[summarize] [<field>:=]<agg> [where <expr>][, [<field>:=]<agg> [where <expr>] ...] [by [<field>][:=<expr>] ...]
+[summarize] [<field>:=]<agg>
+[summarize] [<field>:=]<agg> [where <expr>][, [<field>:=]<agg> [where <expr>] ...]
+[summarize] [<field>:=]<agg> [by [<field>][:=<expr>][, [<field>][:=<expr>]] ...]
+[summarize] [<field>:=]<agg> [where <expr>][, [<field>:=]<agg> [where <expr>] ...] [by [<field>][:=<expr>][, [<field>][:=<expr>]] ...]
+[summarize] by [<field>][:=<expr>][, [<field>][:=<expr>] ...]
```
philrz, line 12: After seeing the first iteration of this PR, @nwt correctly pointed out that the expanded/corrected original synopsis line was way too long and complex. His first thought on how to improve it was to break it out into more lines, so that's what I've done on this next iteration and indeed I'm liking it more. I like leading with the shorter one that represents a very common use, then builds up to the full/beastly one, so if a user is learning incrementally hopefully they'll get the idea.

nwt (approved):

PR #5222, opened 2024-08-15
vector: Rudimentary function support

mattnibs:
This commit adds rudimentary support for functions. There are a lot of shortcomings compared to sequence notably: only supports function lower(), does not variants/unions and wrapped errors are not handled, and it is not currently tested. Future work will brings things in feature parity.

### DIFF
--- a/runtime/vam/expr/function/string.go
+++ b/runtime/vam/expr/function/string.go
@@ -1,0 +1,30 @@
+package function
+
+import (
+	"strings"
+
+	"github.com/brimdata/zed"
+	"github.com/brimdata/zed/vector"
+)
+
+// https://github.com/brimdata/zed/blob/main/docs/language/functions.md#lower
+type ToLower struct {
+	zctx *zed.Context
+}
+
+func (t *ToLower) Call(args []vector.Any) vector.Any {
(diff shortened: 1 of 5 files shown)


runtime/vam/expr/function/string.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+func (t *ToLower) Call(args []vector.Any) vector.Any {
+	v := vector.Under(args[0])
+	if v.Type() != zed.TypeString {
+		// XXX This should be a wrapped error as seen in sequential world.
+		return vector.NewStringError(t.zctx, "lower: string arg required", v.Len())
+	}
+	out := vector.NewStringEmpty(v.Len(), vector.NewBoolEmpty(v.Len(), nil))
+	for i := uint32(0); i < v.Len(); i++ {
```
mccanne, line 22: I think we want a pattern where there is a string function applied to string/const/dict/view and you get back a vector.Any.  So instead of looping over slots you apply a function to a vector and get a vector back.  So you would pass the go func strings.ToLower to the apply function and get back the vector.  We would have to make an applier for every possible type signature but that seems reasonable since it's linear in problem size.

PR #5223, opened 2024-08-15
ztest: Drop unused warnings field

mattnibs:


nwt (approved):
Please also remove the following comment from `runzq` since it concerns warnings.

```go
// If there was an error, errbuf could potentially hold both warnings
// and error messages, but that's not currently an issue with existing
// tests.
```

PR #5224, opened 2024-08-16
Replace vam/expr.stitch with vector.Apply

nwt:
The stitch function performs the rip-eval-stitch operation on a pair of vectors.  Replace it with vector.Apply, which generalizes stitch to operate on a slice of vectors and also allows the caller to control whether or not it rips vector.Unions.  (stitch always rips unions).


mattnibs (approved):
Really cool

mccanne (approved):

PR #5226, opened 2024-08-16
Show named channels in the lake API doc

philrz:
## What's Changing

The lake API doc will now show named channels in the output of the Query example.

## Why

#5191 changed channels to use string-based names rather than numbers, so the API examples should stay in sync with this.

### DIFF
--- a/docs/lake/api.md
+++ b/docs/lake/api.md
@@ -433,11 +433,11 @@ curl -X POST \
 **Example Response**
 
 ```
-{"type":"QueryChannelSet","value":{"channel_id":0}}
+{"type":"QueryChannelSet","value":{"channel":"main"}}
 {"type":{"kind":"record","id":30,"fields":[{"name":"warehouse","type":{"kind":"primitive","name":"string"}},{"name":"count","type":{"kind":"primitive","name":"uint64"}}]},"value":["miami","1"]}
 {"type":{"kind":"ref","id":30},"value":["chicago","2"]}
-{"type":"QueryChannelEnd","value":{"channel_id":0}}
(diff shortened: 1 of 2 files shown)


docs/lake/api.md
```diff
@@ -433,11 +433,11 @@ curl -X POST \
...
 
 ```
-{"type":"QueryChannelSet","value":{"channel_id":0}}
+{"type":"QueryChannelSet","value":{"channel":"main"}}
 {"type":{"kind":"record","id":30,"fields":[{"name":"warehouse","type":{"kind":"primitive","name":"string"}},{"name":"count","type":{"kind":"primitive","name":"uint64"}}]},"value":["miami","1"]}
 {"type":{"kind":"ref","id":30},"value":["chicago","2"]}
-{"type":"QueryChannelEnd","value":{"channel_id":0}}
-{"type":"QueryStats","value":{"start_time":{"sec":1658193276,"ns":964207000},"update_time":{"sec":1658193276,"ns":964592000},"bytes_read":55,"bytes_matched":55,"records_read":3,"records_matched":3}}
+{"type":"QueryChannelEnd","value":{"channel":"main"}}
+{"type":"QueryStats","value":{"start_time":{"sec":1723830077,"ns":168637000},"update_time":{"sec":1723830077,"ns":169204000},"bytes_read":55,"bytes_matched":55,"records_read":3,"records_matched":3}}
```
nwt, line 440: Please leave this line as is.

nwt (approved):

docs/lake/api.md
philrz: @nwt: Hmm. I guess I can. I thought about that but I just chose to take the output wholesale of a re-run of the same API command rather than think it wise to surgically update only the parts I'd _think_ should change and risk not noticing if I'd made the output somehow internally inconsistent in other ways. I agree that in this case that doesn't seem to be the case, so I'll yield to your preference, but I FWIW I feel differently about it.

Issue #5221, opened 2024-08-14
Presenting times in non-UTC timezones


philrz:
## tl;dr

Issues like https://github.com/brimdata/zui/issues/1057 and https://github.com/brimdata/brimcap/issues/352 reflect users' desire to sometimes work in non-UTC timezones, such as their local time. The [`strftime` function](https://zed.brimdata.io/docs/next/language/functions/strftime) added in #5197 provides the `%z` and `%Z` formatting directives that could allow output of such timezones, but right now they only reflect the UTC timezone.

To improve on this, we've discussed approaches for a user to provde a hint for an alternate timezone such that the rendered time value and output of `%z` / `%Z` directives would reflect the delta. In a preliminary discussion on this topic, @mattnibs made the following proposal for `strftime` and the string-based rendering of time values it outputs:

>I could imagine `strftime` taking a third `duration` argument that specifies the offset of timezone and then you could use the `%z` directive to display the timezone. We could have a timezone function that would return the offset.
>
> Eg `strftime(“%z”, now(), tz(“PST”))`

### Comparison With Other Tools

Users are likely to compare our timezone support to that of other tools and/or copy string-based time values between Zed tools and other tools, so it may be helpful to know how others approach this problem. Since they both support the same formatting options as our `strftime`, I happened to start with `jq` and GNU Date.

Since all these tools offer their own default/alternate formats, as a neutral starting point, I'll start from a UTC seconds-since-epoch value `1723662771` that roughly matches the time when this issue was written. I'm ignoring the fractional seconds for now because Zed has orthogonal catch-up to do there (#5220).

#### `jq`

`jq` has its `strftime` function that supports formatting options like Zed's function of the same name. Per their [docs](https://jqlang.github.io/jq/manual/#dates), it _only_ is intended to format a time in UTC (well, they call it GMT).

```
$ echo '1723662771' | jq 'strftime("%Y-%m-%dT%H:%M:%S")' -
"2024-08-14T19:12:51"
```

For output in non-UTC timezones, `jq` offers a _separate_ `strflocaltime` function. By default it outputs time in the local timezone for the system on which `jq` is running, though it overrides that behavior if the `TZ` environment variable is set to something else. My Macbook happens to be in Pacific Daylight Time at the moment, so:

```
$ echo '1723662771' | jq 'strflocaltime("%Y-%m-%dT%H:%M:%S")' -
"2024-08-14T12:12:51"

[...]

PR #5232, opened 2024-08-19
Add Grok pattern for ISO 8601 datestamp

philrz:
## What's Changing

Adding a [`grok`](https://zed.brimdata.io/docs/language/functions/grok) pattern for a date of the format `2024-08-15`.

## Why

A user went looking for such a pattern in a recent [community Slack thread](https://brimdata.slack.com/archives/CTSMAK6G7/p1723833517874349) and indeed one didn't exist in our set of included patterns.

## Details

While I showed the user how to define/use one immediately via Grok's optional third `definitions` argument, this pattern seems common enough that it's worth adding to our included base set.

I've done a fair amount of timestamp-related research lately while putting together the write-up in #5221, so I put a little extra throught into the name of this pattern, e.g., should it be `DATESTAMP_ISO8601` or something else like `DATESTAMP_RFC3339`. However, I went with the former because:

1. A date of this format is the top entry at https://en.wikipedia.org/wiki/ISO_8601, and,
2. RFC3339 is considered a "profile" of ISO 8601, making ISO 8601 the "wider umbrella" thing to reference


pkg/grok/base.go
```diff
@@ -87,6 +87,7 @@ func NewBase() Host {
 	h.Must("ISO8601_TIMEZONE", `Z|[+-]%{HOUR}(?::?%{MINUTE})`)
 	h.Must("ISO8601_SECOND", `%{SECOND}|60`)
 	h.Must("TIMESTAMP_ISO8601", `%{YEAR}-%{MONTHNUM}-%{MONTHDAY}[T ]%{HOUR}:?%{MINUTE}(?::?%{SECOND})?%{ISO8601_TIMEZONE}?`)
+	h.Must("DATESTAMP_ISO8601", `%{YEAR}-%{MONTHNUM}-%{MONTHDAY}`)
```
mattnibs, line 90: Might as well use this in TIMESTAMP and also get the benefit of a little test coverage:
```suggestion
	h.Must("DATESTAMP_ISO8601", `%{YEAR}-%{MONTHNUM}-%{MONTHDAY}`)
	h.Must("TIMESTAMP_ISO8601", `%{DATESTAMP_ISO8601}[T ]%{HOUR}:?%{MINUTE}(?::?%{SECOND})?%{ISO8601_TIMEZONE}?`)
```

mattnibs (approved):
I was skeptical about adding this since it's not included in grok-patterns for logstash, but then again we do have extra base rules extended beyond logstash and being that this is a subset of TIMESTAMP_ISO8601 so I don't think it hurts to add this as well.

PR #5235, opened 2024-08-20
vam: Use vector.Apply for lower()

### LINKED ISSUE #5211: Most functions don't support unions

```
0x666f6f
1
```

mattnibs:
This commit changes vam lower() so that it uses vector.Apply and can now handle Variant and Union types. Also adjust sam lower() so that it de-unions parameters before evaluation.

Partially fixes #5211

### DIFF
--- a/vector/bool.go
+++ b/vector/bool.go
@@ -28,7 +28,9 @@ func (b *Bool) Type() zed.Type {
 }
 
 func (b *Bool) Value(slot uint32) bool {
-	return (b.Bits[slot>>6] & (1 << (slot & 0x3f))) != 0
+	// Because Bool is used to store nulls for many vectors and it is often
+	// nil check to see if receiver is nil and return false.
+	return b != nil && (b.Bits[slot>>6]&(1<<(slot&0x3f))) != 0
 }
 
 func (b *Bool) Set(slot uint32) {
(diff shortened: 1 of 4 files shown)


vector/bool.go
```diff
@@ -28,7 +28,7 @@ func (b *Bool) Type() zed.Type {
 }
 
 func (b *Bool) Value(slot uint32) bool {
-	return (b.Bits[slot>>6] & (1 << (slot & 0x3f))) != 0
+	return b != nil && (b.Bits[slot>>6]&(1<<(slot&0x3f))) != 0
```
mattnibs, line 31: I did this because we were getting panics in vector.StringValue when the Null boolean vector was nil. I could change the code there to check for nulls but it seemed to make more sense to just put the check here.

mccanne: I think all the vector types should behave the same way with regard to nil receiver and my inclination here is to check at the caller and assume b not nil. Since we're here though, I noticed that Value should check for the nulls vector as in other vector types.

mccanne: Hmm after looking deeper I think you're change is ok but include a comment about the nil receiver for Bool.Value() specifically referring to the Nulls use case.

mccanne (approved):
Add comment to vector.Bool.Value()

PR #5240, opened 2024-08-21
vam: replace() function

mattnibs:


runtime/vam/expr/function/function.go
```diff
@@ -1,25 +1,60 @@
...
+		f = &Replace{zctx}
 	default:
 		return nil, nil, function.ErrNoSuchFunction
 	}
 	if err := function.CheckArgCount(narg, argmin, argmax); err != nil {
 		return nil, nil, err
 	}
+	if !dontapply {
+		f = applier(f.Call)
+	}
```
nwt, line 33: For which Zed functions do you anticipate that `dontapply` will be true?

nwt (approved):

PR #5242, opened 2024-08-21
vam: Add remaining string functions

### LINKED ISSUE #5211: Most functions don't support unions

mattnibs:
Partially fixes #5211

### DIFF
--- a/vector/array.go
+++ b/vector/array.go
@@ -38,3 +38,32 @@ func (a *Array) Serialize(b *zcode.Builder, slot uint32) {
 	}
 	b.EndContainer()
 }
+
+func ContainerOffset(val Any, slot uint32) (uint32, uint32, bool) {
+	switch val := val.(type) {
+	case *Array:
+		return val.Offsets[slot], val.Offsets[slot+1], val.Nulls.Value(slot)
+	case *Set:
+		return val.Offsets[slot], val.Offsets[slot+1], val.Nulls.Value(slot)
+	case *Map:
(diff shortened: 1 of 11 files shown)


vector/array.go
```diff
@@ -38,3 +38,30 @@ func (a *Array) Serialize(b *zcode.Builder, slot uint32) {
...
+	switch val := val.(type) {
+	case *Array:
+		return val.Offsets[slot], val.Offsets[slot+1], val.Nulls.Value(slot)
+	case *View:
+		slot = val.Index[slot]
+		a := val.Any.(*Array)
+		return a.Offsets[slot], a.Offsets[slot+1], a.Nulls.Value(slot)
+	}
+	panic(val)
+}
```
mattnibs, line 53: @mccanne am I right here in that Dict and Consts are not used for Arrays? Also the function name is kind of weird if you have any better ideas.

vector/array.go
```diff
@@ -38,3 +38,33 @@ func (a *Array) Serialize(b *zcode.Builder, slot uint32) {
 	}
 	b.EndContainer()
 }
+
+func OffsetsOf(val Any, slot uint32) (uint32, uint32, bool) {
+	// Dict / Const do not appear for Array, Sets, Maps?
```
mattnibs, line 43: @mccanne am I right here in that Dict and Consts are not used for Arrays? Also the function name is kind of weird if you have any better ideas.

nwt (approved):

PR #5243, opened 2024-08-22
Improve Grok function docs

philrz:
A recent [community Slack thread](https://brimdata.slack.com/archives/CTSMAK6G7/p1723833517874349) inspired me to finish a long standing to-do to improve these Grok function docs.

If you'd like to review by seeing it rendered, I've got it on a personal staging site at:

https://66c760c8bd7394547de20d6c--spiffy-gnome-8f2834.netlify.app/docs/next/language/functions/grok

### DIFF
--- a/docs/language/functions/grok.md
+++ b/docs/language/functions/grok.md
@@ -1,38 +1,159 @@
 ### Function
 
-&emsp; **grok** &mdash; parse a string using a grok pattern
+&emsp; **grok** &mdash; parse a string using a Grok pattern
 
 ### Synopsis
 
 ```
-grok(p: string, s: string) -> any
-grok(p: string, s: string, definitions: string) -> any
+grok(p: string, s: string) -> record
+grok(p: string, s: string, definitions: string) -> record
 ```
 
 ### Description
 
-The _grok_ function parses a string `s` using grok pattern `p` and returns
+The _grok_ function parses a string `s` using Grok pattern `p` and returns
 a record containing the parsed fields. The syntax for pattern `p`
 is `%{pattern:field_name}` where _pattern_ is the name of the pattern
 to match in `s` and _field_name_ is the resultant field name of the capture
 value.
 
 When provided with three arguments, `definitions` is a string
-of named patterns in the format `PATTERN_NAME PATTERN` each separated by newlines.
-The named patterns can then be referenced in argument `p`.
+of named patterns in the format `PATTERN_NAME PATTERN` each separated by
+newlines (`\n`). The named patterns can then be referenced in argument `p`.
 
-#### Included Patterns
+### Included Patterns
 
-The _grok_ function by default includes a set of builtin named patterns
+The `grok` function by default includes a set of built-in named patterns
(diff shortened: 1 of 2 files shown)


docs/language/functions/grok.md
```diff
@@ -1,38 +1,159 @@
...
-&emsp; **grok** &mdash; parse a string using a grok pattern
+&emsp; **grok** &mdash; parse a string using a Grok pattern
 
 ### Synopsis
 
 ```
-grok(p: string, s: string) -> any
-grok(p: string, s: string, definitions: string) -> any
+grok(p: string, s: string) -> record
+grok(p: string, s: string, definitions: string) -> record
```
philrz, line 9: I'm not sure if the return of `any` that was there before was trying to communicate something subtle that I'm missing. However, since the text below has always said it "returns a record" I went ahead and made this change. But if there's more to the story we can change it back and figure out what else needs to change in the text.

docs/language/functions/grok.md
```diff
@@ -1,38 +1,159 @@
...
 
-#### Included Patterns
+### Included Patterns
 
-The _grok_ function by default includes a set of builtin named patterns
+The `grok` function by default includes a set of built-in named patterns
 that can be referenced in any pattern. The included named patterns can be seen
 [here](https://raw.githubusercontent.com/brimdata/zed/main/pkg/grok/base.go).
 
+### Comparison to Other Implementations
```
philrz, line 30: I know this section is long, but it's my best attempt to deal with the fact that web searches for "grok parsing" and such will reveal tons of examples that, taken unmodified, will not work in our implementation. Since we collectively don't have the cycles or stomach to build full compatibility with Logstash, I figure the next best thing is just disclose as much as we can and hope users are patient.

docs/language/functions/grok.md
mattnibs: `any` is probably more correct since `record` isn't an explicit type, but then again we does use `record` in docs for `nest_dotted` and it does probably make things clearer to the new user so I think this fine.

nwt: We should add "record" to docs/language/conventions.md.

mattnibs:
I think my problem with this change is it basically reads like here are the features we don't have that logstash has when docs usually read like here are the features. It also assumes that user looking at this doc is familiar with logstash's implementation of grok which probably is true but also might now be. I do think we could add example showing how to capture nested fields using `nest_dotted` and an example of how to do type conversion after the function is applied. I do think it would be ok to place a section called "differences from logstash" at the bottom of the docs after the examples.

nwt (approved):
I spoke with @mattnibs about this today. We agreed that working through [his suggestions](https://github.com/brimdata/zed/pull/5243#issuecomment-2340844180) comes second to work he and I are focused on now, so we should merge this as is (and possibly revisit the suggestions later).

philrz:
Thanks @nwt and @mattnibs! If we're lucky, some users may actually work off these docs by the time the topic next comes around and we'll have the benefit of their input as well. I'll go ahead and merge this for now.

PR #5248, opened 2024-08-23
Break many docs examples across multiple lines

philrz:
## What's Changing

I've reformatted many of the Zed docs examples across multiple lines to minimize the number of times readers need to "scroll right".

If you'd like to see the rendered site as part of review, I've put a copy of what's on this branch at a personal staging site at:

https://66c7d4d2d4764d84e6cdf484--spiffy-gnome-8f2834.netlify.app/docs/next

## Why

Several of us have noticed this before, but a user recently spoke up about it in a [community Slack thread](https://brimdata.slack.com/archives/CTSMAK6G7/p1724275452056539) and that inspired me to finally take action.

## Details

The majority of them were fixable by just changing from:

```
echo "input data" | zq -z '...zed code...`
```

to

```
echo "input data" |
  zq -z '...zed code...`
```

i.e., I adopted the convention of putting the pipe for shell continuation at end of line when it would help.

For situations where I needed continuation of Zed pipelines, I put the pipes at beginning of next line, e.g.:

```
from pool
| count()
```

Overall, I stopped short of being totally dogmatic. In some cases I was able to break long inputs across multiple lines (e.g., when there were many `echo'`ed individual records) but some things would have required using some of the more ugly shell continuation tricks (e.g., breaking up single long strings) so there remain a handful if situations where "scroll right" is still needed to see everything. I figure this is not the end of the world.

I found a couple bonus things I fixed up along the way, including:

1. Examples that weren't previously protected by `mdtest` at all but could be
2. Switched from `"` to `'` for surrounding the Zed code in some examples, since most already use `'` and I figured we should be consistent (plus `'` is a good habit for users since it avoids some shell surprises)

### DIFF
--- a/docs/language/aggregates/union.md
+++ b/docs/language/aggregates/union.md
@@ -38,18 +38,19 @@ echo '1 2 3 3' | zq -z 'yield union(this)' -
 ```
 
 Mixed types create a union type for the set elements:
-```mdtest-command-issue-3610
+```mdtest-command
 echo '1 2 3 "foo"' | zq -z 'set:=union(this) | yield this,typeof(set)' -
 ```
 =>
-```mdtest-output-issue-3610
+```mdtest-output
 {set:|[1,2,3,"foo"]|}
 <|[(int64,string)]|>
 ```
 
 Create sets of values bucketed by key:
 ```mdtest-command
-echo '{a:1,k:1} {a:2,k:1} {a:3,k:2} {a:4,k:2}' | zq -z 'union(a) by k | sort' -
+echo '{a:1,k:1} {a:2,k:1} {a:3,k:2} {a:4,k:2}' |
+  zq -z 'union(a) by k | sort' -
 ```
 =>
 ```mdtest-output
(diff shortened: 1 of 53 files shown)


docs/language/aggregates/union.md
```diff
@@ -38,18 +38,19 @@ echo '1 2 3 3' | zq -z 'yield union(this)' -
 ```
 
 Mixed types create a union type for the set elements:
-```mdtest-command-issue-3610
```
philrz, line 41: This fix was unique. Through black box testing I confirmed that the trailing `-issue-3610` was having the effect of _not_ running the example via `mdtest`. Furthermore, I looked at #3610 and saw it had been fixed, and indeed, I found that if I dropped that trailing `-issue-3610` and let mdtest do its thing, it runs clean (and doesn't run clean if I intentionally mucked the output). So it's now going to be mdtest-protected going forward. Please speak up if I that was the wrong decision, though!

nwt (approved):

PR #5250, opened 2024-08-23
vam: Add fields() func

mattnibs:


runtime/vam/expr/function/fields.go
```diff
@@ -0,0 +1,94 @@
...
+func (f *Fields) Call(args ...vector.Any) vector.Any {
+	val := vector.Under(args[0])
+	switch typ := val.Type().(type) {
+	case *zed.TypeRecord:
+		paths := buildPath(typ, nil)
+		s := vector.NewStringEmpty(val.Len(), nil)
+		inOffs, outOffs := []uint32{0}, []uint32{0}
+		nulls := vector.NewBoolEmpty(val.Len(), nil)
+		for i := uint32(0); i < val.Len(); i++ {
+			if vector.IsNull(val, i) {
```
nwt, line 33: runtime/sam/expr/function/fields.go doesn't treat a null record value any differently than a non-null record value. Is there a reason for doing that here?

mattnibs: Good point, I'll change this.

runtime/vam/expr/function/fields.go
```diff
@@ -0,0 +1,88 @@
...
+			if rtyp == nil {
+				errcnt++
+				tags[i] = 1
+				continue
+			}
+			inOffs, outOffs = appendPaths(buildPath(rtyp, nil), s, inOffs, outOffs)
+		}
+		inner := vector.NewArray(f.innerTyp, inOffs, s, nil)
+		out := vector.NewArray(f.outerTyp, outOffs, inner, nil)
+		return vector.NewVariant(tags, []vector.Any{out, vector.NewStringError(f.zctx, "missing", errcnt)})
```
mccanne, line 53: The common case is to not have a variant result here.  I think we should have a check for `errcnt == 0` check and just return `out` in this case...?

nwt (approved):

PR #5256, opened 2024-08-29
vam: Add array, set, record index expressions

mattnibs:


runtime/vam/expr/eval.go
```diff
@@ -32,3 +33,91 @@ func (c *Call) Eval(this vector.Any) vector.Any {
...
+	case *vector.Map:
+		panic("vector index operations on maps not supported")
+	default:
+		return vector.NewMissing(i.zctx, this.Len())
+	}
+}
+
+func indexVector(zctx *zed.Context, offsets []uint32, vals, index vector.Any, nulls *vector.Bool) vector.Any {
+	if !zed.IsInteger(index.Type().ID()) {
+		return vector.NewWrappedError(zctx, "index is not an integer", index)
```
nwt, line 72: Nit:
```suggestion
		return vector.NewWrappedError(zctx, "array or set index is not an integer", index)
```

runtime/vam/expr/eval.go
```diff
@@ -32,3 +33,91 @@ func (c *Call) Eval(this vector.Any) vector.Any {
...
+}
+
+func indexRecord(zctx *zed.Context, record *vector.Record, index vector.Any) vector.Any {
+	if index.Type().ID() != zed.IDString {
+		return vector.NewWrappedError(zctx, "record index is not a string", index)
+	}
+	var errcnt uint32
+	tags := make([]uint32, record.Len())
+	n := len(record.Typ.Fields)
+	viewtags := make([][]uint32, n)
```
nwt, line 105: Nit: These are indexes, not tags, so name accordingly.

nwt (commented):
Will finish reviewing and approve after lunch.

vector/any.go
```diff
@@ -21,3 +23,23 @@ type Puller interface {
 }
 
 type Builder func(*zcode.Builder) bool
+
+func Combine(desiredLen uint32, vec Any, index []uint32, add Any) Any {
```
nwt, line 27: Nit: I don't think desiredLen is necessary since I think it's always equal to vec.Len()+uint32(len(index)).

nwt (approved):
Nit: Remove runtime/sam/expr/ztests/negative-index.yaml since it's redundant with runtime/ztests/expr/index.yaml.

Nits:Move./runtime/sam/expr/ztests/index-deunion-vector.yaml

vector/any.go
mattnibs: It is necessary the way it's being used in index expressions but perhaps I am approaching this wrong. Happy to talk about this.

vector/any.go
```diff
@@ -21,3 +23,24 @@ type Puller interface {
 }
 
 type Builder func(*zcode.Builder) bool
+
+func Combine(vec Any, index []uint32, add Any) Any {
+	var tags []uint32
+	var vecs []Any
+	size := int(vec.Len())
```
nwt, line 30: I think you have to include index here.
```suggestion
	size := int(vec.Len()) + len(index)
```

mattnibs: There's more to this here and I'm not exactly sure what to do about this.

nwt: In that case, I think you should restore the desiredLen parameter to Combine since that version worked, and then we can maybe try removing it again later on.

mattnibs: On second thought your suggestion was right but I needed to make adjustments to how combine worked.

nwt (approved):
:shipit:

Issue #4565, opened 2023-04-30
Handling JSON-compatible pseudo-maps


philrz:
Repro is with Zed commit 4ffdf3e.

Consider again the `jq` example with test data [scan.json.gz](https://github.com/brimdata/zed/files/11348260/scan.json.gz) shown at the top of #4555 and the Zed that tries to come close to it.

```
$ jq --version
jq-1.6

$ jq 'reduce .[] as $e ({}; . + { ($e.ip): (.[$e.ip] + $e.ports) })' scan.json
{
  "192.168.5.53": [
    {
      "port": 515,
      "proto": "tcp",
      "status": "open",
      "reason": "syn-ack",
      "ttl": 64
    },
    {
      "port": 21,
      "proto": "tcp",
      "status": "open",
      "reason": "syn-ack",
      "ttl": 64
    }
  ],
  "192.168.5.49": [
    {
      "port": 49152,
      "proto": "tcp",
      "status": "open",
      "reason": "syn-ack",
      "ttl": 64
    },
    {
      "port": 3401,
      "proto": "tcp",
      "status": "open",
      "reason": "syn-ack",
      "ttl": 64
    }
  ]
}

$ zq -version
Version: v1.7.0-55-g4ffdf3ef

$ zq -Z 'over this | ports:=collect(ports[0]) by ip | unflatten([{key:ip,value:ports}])' scan.json
{
    "192.168.5.53": [
        {
            port: 515,
            proto: "tcp",
            status: "open",
            reason: "syn-ack",
            ttl: 64
        },
        {
            port: 21,
            proto: "tcp",
            status: "open",
            reason: "syn-ack",
            ttl: 64
        }
    ]
}
{
    "192.168.5.49": [
        {
            port: 49152,
            proto: "tcp",
            status: "open",
            reason: "syn-ack",
            ttl: 64
        },
        {
            port: 3401,
            proto: "tcp",
            status: "open",
            reason: "syn-ack",
            ttl: 64
        }
    ]
}
```

A subtle difference is that the `zq` output is two separate records while the `jq` output is a single JSON object. There's nothing about the JSON object output by `jq` that says it must ultimately be used like a map, but docs like [Go's `json.Marshal()`](https://pkg.go.dev/encoding/json#Marshal) (and the equivalent in other languages) set expectations about bouncing between this JSON representation and a true `map` within a language.

[...]

PR #5264, opened 2024-09-06
Update docs for lateral subqueries and over operator

philrz:
## What's Changing

The docs for Lateral Subqueries and the `over` operator are revised to clarify some subtleties revealed during a recent internal discussion. I made other changes along the way such as adding hyperlinks, reducing redundancy, and addressing conflicting text.

If you would like to review the content in rendered form, I've published it on a personal staging site at:
* https://whimsical-flan-b2925b.netlify.app/docs/next/language/lateral-subqueries
* https://whimsical-flan-b2925b.netlify.app/docs/next/language/operators/over

## Why

The internal discussion was about Lateral Expressions and when they return primitive values or arrays. The discussion created enough debate that once we reached consensus it was clear this needed to be divulged to users.

## Details

As I was adding the new text in the Lateral Expressions section, I reviewed the other text at the overall Lateral Subqueries page as well as the related `over` docs and saw many things that could stand to be polished up. For starters, there was no hyperlinking from the `over` page back to the Lateral Subqueries page, which seemed like a big exposure since subtleties like the one covered in this new text represent things a new user of `over` might need to know.

I also noticed that there was some text in the `over` page that effectively described Lateral Subqueries, but some of that text used language that conflicted with what's at the Lateral Subqueries page. Rather than end up with redundant text that would need to be kept in sync in future revisions, I took the step of trimming the `over` page to narrow its scope (moving some relevant text that wasn't previously covered at the Lateral Subqueries page) and relying on repeated hyperlinking to the Lateral Subqueries page to ensure the user knows where they can find the additional info.

It would be great if I could get an approval on https://github.com/brimdata/zed-docs-site/pull/83 before this merges so we can be spared yet another round of "broken link checker failures" due to a known issue.


docs/language/lateral-subqueries.md
```diff
@@ -112,13 +126,13 @@ parenthesized form:
...
-> dataflow operator.
 
-This form must always include a lateral scope as indicated by `<lateral>`,
-which can be any dataflow operator sequence excluding [`from` operators](operators/from.md).
-As with the `over` operator, values from the outer scope can be brought into
-the lateral scope using the `with` clause.
+:::tip
+The parentheses disambiguate a lateral expression from a lateral
+dataflow operator.
+:::
```
philrz, line 133: I'd appreciate some input on this text. All I've done with it thus far is reformat it. However, as a user/reader, I struggle to understand what it's trying to clarify. I'd be in favor of expanding the text a bit to make it clearer or drop it until some user's confusion reminds us why we put it there. 😄

docs/language/operators/over.md
```diff
@@ -12,45 +12,21 @@ The `over` operator traverses complex values to create a new sequence
...
 Each expression `<expr>` is evaluated in left-to-right order and derived sequences are
 generated from each such result depending on its types:
-* an array value generates each of its element,
+* an array value generates each of its elements,
 * a map value generates a sequence of records of the form `{key:<key>,value:<value>}` for each
 entry in the map, and
 * all other values generate a single value equal to itself.
 
-Records can be converted to maps with the [_flatten_ function](../functions/flatten.md)
+Records can be converted to maps with the [`flatten` function](../functions/flatten.md)
```
philrz, line 25: > Records can be converted to maps...

Strictly speaking, it doesn't seem like `flatten` converts them to maps. Rather, it generates a sequence of values similar to what `over` does with maps, but even then, there's still the difference that `flatten` creates an array for the path of the record keys whereas `over` on a map outputs the keys as is. Does this bother anyone? It doesn't _quite_ bother me enough to insist on a major expansion of this text. However, user confusion about maps has come up a few times (e.g., the subtle difference between true Zed map types and what I've been calling "JSON-compatible pseudo-maps" (#4565)) which is why I perked up when I saw a possible oversimplified reference to "maps" here.

Anyone have reactions?

docs/language/lateral-subqueries.md
philrz: Update: I've spent more time staring at it and I think I at least understand what it's trying to say, but I'm still struggling for the text to propose instead. I _think_ the point being made is that the open parenthesis appearing _before_ the keyword `over` is what makes it clear that we're in expression context and not dataflow context.

The reason the text was bothering me is because, assuming a new user reading these docs for the first time, we've just showered them with a bunch of related terms on this page: Lateral Subquery, Lateral Scope, Lateral Expression... and it feels like in this one sentence we cooked up "lateral dataflow operator" as yet another, but unlike the others, it's not a specific term that's defined in some place we can link to.

I think I'm still leaning toward just taking it out. I'm feeling like the overall Lateral Subquery topic is one of those growth points as a user learning Zed where there's no shortcuts and they just need to closely read the docs, hack with the examples, apply it to their own use cases, and when they come out the other side it'll hopefully have stuck. And I feel like this particular point about the parens comes through already in the text without this additional note. But I'm open to other thoughts!

docs/language/lateral-subqueries.md
```diff
@@ -7,11 +7,19 @@ sidebar_label: Lateral Subqueries
...
 Lateral subqueries provide a powerful means to apply a Zed query
 to each subsequence of values generated from an outer sequence of values.
-The inner query may be _any Zed query_ and may refer to values from
+The inner query may be _any_ dataflow operator sequence (excluding
+[`from` operators](operators/from.md)) and may refer to values from
 the outer sequence.
 
+:::tip Note
+This pattern rhymes with the SQL pattern of a "lateral
+join", which runs a SQL subquery for each row of the outer query's table.
```
nwt, line 16: Nit: SQL seems redundant here. And "results" is clearer than "table".
```suggestion
join", which runs a subquery for each row of the outer query's results.
```

nwt (approved):

docs/language/lateral-subqueries.md
philrz: @nwt: Works for me! That was actually existing text that I was just moving and reformatting. In general I am not too assertive about overhauling other people's statements about SQL, so I appreciate the suggestion.

docs/language/lateral-subqueries.md
nwt: I think this should stay and is fine as is. Rather than adding more text, how about just making "lateral dataflow operator" into a link to the `over` operator page?

PR #5265, opened 2024-09-08
Update cast function docs

philrz:
## What's Changing

A few small improvements are proposed to the `cast` function docs and a related section on Named Types.

## Why

A user recently asked a question about the `cast` function on community Slack and from their question it was apparent that they were unfamiliar with the concept of Named Types. Reviewing the docs myself, I could see there were hyperlinks missing between the relevant pages, so by adding them now I'm hoping future users will fare better.

## Details

While reviewing I found some other hyperlinks that seemed worth adding. I also made a wording improvement.

### DIFF
--- a/docs/language/data-types.md
+++ b/docs/language/data-types.md
@@ -89,6 +89,7 @@ persist into the data model and thus into the serialized input and output.
 
 Named types may be defined in three ways:
 * with a [`type` statement](statements.md#type-statements),
+* with the [`cast` function](functions/cast.md),
 * with a definition inside of another type, or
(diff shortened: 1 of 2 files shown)


docs/language/data-types.md
```diff
@@ -89,6 +89,7 @@ persist into the data model and thus into the serialized input and output.
 
 Named types may be defined in three ways:
```
mattnibs, line 90:
```suggestion
Named types may be defined in four ways:
```

mattnibs (approved):

PR #5269, opened 2024-09-10
Add vector Format

mattnibs:
Add vector.Format function which displays a given vector in human-readable form.

### DIFF
--- a/vector/format.go
+++ b/vector/format.go
@@ -1,0 +1,87 @@
+package vector
+
+import (
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/brimdata/zed/zson"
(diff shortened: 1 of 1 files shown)


mattnibs:
credit to @nwt

mccanne:
I think this should serialize to Zed.  You can put MarshalZNG methods on each vector and use zq to format to human-readable ZSON.  Format() can then just call zson.String().  I'm assuming you want this for printf debugging?  If we do it this way, then we could also have a `zed dev` command that outputs a VNG file's vectors as Zed.  Good for demos and learning.  You can load that output into a lake for a serious debugging session too!

PR #5272, opened 2024-09-11
Change references to "dataflow" in docs to "pipeline"

philrz:
## What's Changing

References to "dataflow" in the user-facing docs are changed to "pipeline" and references to a "leg" or "path" of dataflow are changed to "branches" of pipelines. This is a prerequisite to revisiting the changes proposed in #5264.

If you'd like to review the docs in rendered form, I've published a build from this branch at a personal staging site https://spiffy-gnome-8f2834.netlify.app/docs/next.

## Why

@mattnibs recently reacted to a line of pre-existing docs text that got moved around in the proposed #5264 changes. The line mentioned a "dataflow operator sequence" which indeed seems like a mouthful. In a group discussion about this, there was consensus that the word "pipeline" could be used here instead of "dataflow" and in so doing make the extra word "sequence" unnecessary.

We then discussed the possibility of rippling this kind of change throughout the rest of the docs. While in the past we had what we felt were defensible reasons for citing Unix pipelines one time early in the docs but then preferring the word "dataflow", @mccanne pointed out how several similar technologies in recent years have all settled on "pipeline". That led to consensus that it was worth making a pass through the docs and seeing how it reads with this wider change.

## Details

When summarizing the changes from this PR's first commit with the team, I noted how text that used to read "multiple legs of the dataflow path" would now read "multiple legs of the pipeline", which led @mccanne to propose also switching from "legs" to "branches". When making that change I also spotted several places where we spoke of just "paths" in the abstract and so to be consistent I also changed these to "branches" or "pipeline branches" depending on which reads clearer in a given context.

Having made the full set of changes, I'm pretty happy with how it reads. "Leg" by itself was not particularly ambiguous before, but "path" was arguably overloaded since we sometimes use it to refer to a storage location and sometimes to the hierarchical set of names referring to a field in a nested record, and these multiple uses sometimes showed up within the same page (e.g., the `from` operator doc). Therefore starting to say "branch" seems like an improvement. It's true we also use "branch" in multiple ways (e.g., branches of a pool in the Zed lake) but those seem to come up in different contexts.

I poked around to check what links outside the Zed repo might break from this change. Indeed, there's one hyperlink in a blog article on the [brimdata.io](https://brimdata.io) site that I'd update, which is no biggie. However, something else I noticed at the same time is that the [Introducing zq](https://www.brimdata.io/blog/introducing-zq/) blog article currently says "dataflow" 20+ times but says "pipeline" once just to say how Zed's is "kind of like a Unix pipeline but not really". @mccanne already signed off on leaving that one alone, and I'm inclined to agree since the article speaks of dataflow in a way that seems academic enough (e.g., describing `jq`’s computational model as "stateless dataflow") and it reads fine as a standalone doc.


docs/language/expressions.md
```diff
@@ -8,7 +8,7 @@ sidebar_label: Expressions
 Zed expressions follow the typical patterns in programming languages.
 Expressions are typically used within data flow operators
```
nwt, line 9:
```suggestion
Expressions are typically used within pipeline operators
```

docs/language/lateral-subqueries.md
```diff
@@ -116,12 +116,12 @@ parenthesized form:
 > dataflow operator.
```
nwt, line 116:
```suggestion
> pipeline operator.
```

philrz: @nwt: It was actually intentional that I didn't touch this one because I'd already posed a question about it at https://github.com/brimdata/zed/pull/5264/files#r1747648931 and so expected to change it there based on what reviewers said. In essence, I feel like this text needs more help than just the dataflow->pipeline change, and perhaps the text should be removed entirely. If it stays, I'd certainly make sure it says "pipeline" regardless. Could you offer your take on if the overall statement is fine as is, and if not, how it could be clarified or if it should be removed?

nwt: This comment has my take: https://github.com/brimdata/zed/pull/5264#discussion_r1761003157.

docs/language/operators/merge.md
```diff
@@ -9,14 +9,14 @@
...
+The `merge` operator merges inputs from multiple upstream branches of
+the pipeline into a single output.  The order of values in the combined
 output is determined by the `<expr>` arguments, which act as sort expressions
-where the values from the upstream paths are forwarded based on these expressions.
+where the values from the upstream pipeline branches are forwarded based on these expressions.
 
 ### Examples
 
-_Copy input to two paths and combine_
+_Copy input to two pipeline branches and merge_
```
nwt, line 19: Nit: Make the same change you made to the similar example in docs/language/operators/fork.md.

nwt (approved):
Merge after you fix https://github.com/brimdata/zed/pull/5272#pullrequestreview-2307451176 and CI is green.

PR #5203, opened 2024-08-05
sort: Allow ordering on a per key basis

mattnibs:
Change the sort operator so that different ordering can be applied to each key in the sort.

Closes #4945

Issue #4591, opened 2023-05-12
Improve top operator


nwt:
The top operator is limited to field references. It should should accept arbitrary expressions like the sort operator does.

It should also allow selection of either ascending or descending order for each expression.

philrz:
As of Zed commit 4290afd, it looks like `top` does take arbitrary expressions like `sort` does, e.g., modifying an example from the `sort` docs:

```
$ echo '{word:"hello"}{word:"Hi"}{word:"WORLD"}' |   zq -z 'top 2 lower(word)' -
{word:"WORLD"}
{word:"Hi"}
```

However, it remains true that `top` lacks the per-key ordering of `sort` like was added in #5203.

philrz:
Verified in super commit 7a5d7b95.

The same kind of per-key ordering used with `sort` is now available with `top`. For instance, we can start with the "Sort by secondary key in reverse order when the primary keys are identical" example from the [`sort` docs](https://superdb.org/docs/language/operators/sort/) and use `top 1` instead of `sort` to get just the first value.

```
$ super -version
Version: v1.18.0-403-g7a5d7b95c

$ echo '{s:"bar",k:2} {s:"bar",k:3} {s:"foo",k:2}' | super -z -c 'top 1 k,s desc' -
{s:"foo",k:2}
```

Thanks @nwt!

PR #5276, opened 2024-09-13
Document top operator

### LINKED ISSUE #4590: Document top operator
There's no documentation for the top operator, but there should be.

philrz:
## What's Changing

Docs are added for the `top` operator.

## Why

It's been undocumented for a long time, primarily due to resistance from me, but I am now officially reversing my position. 😂

## Details

https://github.com/brimdata/zed/issues/4590#issuecomment-1548256191 summarizes my prior resistance, the tl;dr being that tools like Splunk and Humio (now LogScale) each have their own `top` that had different functionality, and log-centric use cases were a big enough focus of the Zed tools that I was concerned this would create confusion. However, the more recent focus has been on applying Zed in general data use cases, and in that world I see less room for confusion, e.g., some SQL implementations have [`SELECT TOP`](https://www.w3schools.com/sql/sql_top.asp), but this is only in limited dialects (e.g., others use `LIMIT`) and so those are more synonymous with Zed's `head` operator. In conclusion, it seems the word is kinda used all over the place and I'm now recognizing the futility of my holding out this long hoping it would become clearer.

In this PR I've revived an old branch from @mattnibs where he started drafting the `top` docs in #5066. In addition to general polish, I also added an additional example that happens to show how to achieve the Splunk-like `top` equivalent by combining it with `count()`.

The astute reader may notice that the comparison with `sort` I've left in the text is perhaps now dated because #5203 gave us per-key ordering in `sort` and, as called out in #4591, `top` lacks that. Personally, I'm fine with that distinction remaining until users speak up or we have spare cycles.

Closes #4590.


docs/language/operators/top.md
```diff
@@ -0,0 +1,40 @@
...
+
+### Synopsis
+
+```
+top <uint> <expr> [, <expr> ...]
+```
+### Description
+
+The `top` operator returns the top N values from a sequence sorted in descending
+order by one or more expressions. `top` is functionally similar to
```
nwt, line 13: `<uint>` isn't right since the first argument isn't an unsigned integer but rather a compile-time constant expressions that evaluates to a positive integer. For example, this works.

```
const limit = 5
top limit this
```

Maybe say something like this (inspired by the synopsis in docs/language/operators/switch.md)?

```suggestion
top <const-expr> <expr> [, <expr> ...]
```
### Description

The `top` operator returns the top N values from a sequence sorted in descending
order by one or more expressions. N is given by `<const-expr`, a compile-time
constant expression evaluating to a positive integer.

`top` is functionally similar to
```

nwt (approved):

PR #5277, opened 2024-09-13
Mention that conditionals can be chained

philrz:
## What's Changing

An example is added to the user-facing docs for conditional expressions pointing out that they can be chained to provide an "else if" equivalent.

## Why

While astute readers would hopefully make this connection on their own, given the ternary syntax is already seen as obtuse by some, I figure the more we make explicit regarding its use, the better.

### DIFF
--- a/docs/language/expressions.md
+++ b/docs/language/expressions.md
@@ -195,7 +195,7 @@ produces
 -2
 ```
 
-Conditional expressions can be chained, providing the logical equivalent of
+Conditional expressions can be chained, providing behavior equivalent of
 "else if" as appears in other languages.
 
 For example,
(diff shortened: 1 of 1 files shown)


docs/language/expressions.md
```diff
@@ -195,6 +195,21 @@ produces
 -2
 ```
 
+Conditional expressions can be chained, providing the logical equivalent of
```
nwt, line 198: Nit: Avoid any possibly confusion with the "Logic" section.
```suggestion
Conditional expressions can be chained, providing behavior equivalent of
```

nwt (approved):

Issue #5230, opened 2024-08-17
Post-merge operator prevents debug on flowgraph branches


philrz:
Repro is with Zed commit e7623c5.

This works as expected:

```
$ zq -version
Version: v1.17.0-27-ge7623c57

$ echo "1 3" | zq 'fork (
  => count() | debug f"debug: count {this}" | yield f"post debug count {this}"
  => sum(this) | debug f"sum is {this}" | yield f"post debug sum {this}"
)' -
"post debug count 2(uint64)"
"post debug sum 4"
"sum is 4"
"debug: count 2(uint64)"
```

But once I add an operator after the merge point, it no longer parses.

```
$ echo "1 3" | zq 'fork (
  => count() | debug f"debug: count {this}" | yield f"post debug count {this}"
  => sum(this) | debug f"sum is {this}" | yield f"post debug sum {this}"
) | yield this + " (merged)"' -
output operator must be at flowgraph leaf at line 2, column 16:
  => count() | debug f"debug: count {this}" | yield f"post debug count {this}"
               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
output operator must be at flowgraph leaf at line 3, column 18:
  => sum(this) | debug f"sum is {this}" | yield f"post debug sum {this}"
                 ~~~~~~~~~~~~~~~~~~~~~~
```

This poses a challenge for allowing `debug` in Zui since the app appends additional operators `head` after the user's program to help with pagination.

philrz:
I just checked in on this issue and confirmed it's still with us as of tagged release `v0.2.0`. Updating the repro query for current SuperSQL syntax:

```
$ super -version
Version: v0.2.0

$ echo "1 3" | super -c '
fork
  (count() | debug f"debug: count {this}" | values f"post debug count {this}")
  (sum(this) | debug f"sum is {this}" | values f"post debug sum {this}"
) | values f"{this} (merged)"' -

output operator must be at flowgraph leaf at line 2, column 12:
(count() | debug f"debug: count {this}" | values f"post debug count {this}")
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
output operator must be at flowgraph leaf at line 3, column 14:
(sum(this) | debug f"sum is {this}" | values f"post debug sum {this}"
             ~~~~~~~~~~~~~~~~~~~~~~
```

philrz:
Verified in super commit 02986c2.

The query forked query now executes and produces the expected debug outputs from the flowgraph branches.

```
[...]

PR #5288, opened 2024-09-17
Hide "debug" operator docs for now

philrz:
## What's Changing

The user-facing `debug` operator docs are being hidden for now.

## Why

We're planning to put out a new set of GA releases soon. The `debug` operator currently feels a little hazardous to expose because of open issue #5230 and lack of support for its output in Zui. Current consensus is to just hide the docs for now on the assumption we'll wrap up those loose ends and reveal it in another release in the near future.


mattnibs (approved):

nwt:
@philrz: I changed the title since this removes rather than hides the debug docs.

nwt (approved):

PR #5289, opened 2024-09-17
Implement vector pass operator

nwt:


mattnibs (approved):

nwt:
@mattnibs: I just realized there's no need for runtime/vam/op.Pass so I removed it. I'll hold off on merging until you take another look.

mattnibs (approved):

PR #5300, opened 2024-09-23
vam: Add preliminary quiet support

mattnibs:
This commit add the quiet function and support for error("quiet") in the yield operator. Quiet errors are not supported elsewhere including cut and put.


type.go
```diff
@@ -469,7 +469,7 @@ func CompareTypes(a, b Type) int {
 		}
 		return 0
 	case ErrorKind:
-		ea, eb := a.(*TypeError), b.(*TypeError)
+		ea, eb := TypeUnder(a).(*TypeError), TypeUnder(b).(*TypeError)
```
mattnibs, line 472: This had to be fixed because it was causing a parquet boomerang value when creating a union type of a named error.

vector/error.go
```diff
@@ -42,9 +42,8 @@ func NewStringError(zctx *zed.Context, msg string, len uint32) *Error {
 }
 
 func NewMissing(zctx *zed.Context, len uint32) *Error {
-	missing := zctx.Missing()
-	vals := NewConst(missing, len, nil)
-	return &Error{Typ: missing.Type().(*zed.TypeError), Vals: vals}
+	vals := NewConst(zed.NewString("missing"), len, nil)
+	return &Error{Typ: zctx.LookupTypeError(zed.TypeString), Vals: vals}
```
nwt, line 46:
```suggestion
	return NewStringError(zctx, string(zed.Missing), len)
```

type.go
nwt: The `MapKind`, `UnionKind`, and `EnumKind` cases all need the same treatment. But rather than add `TypeUnder` calls in each, how about adding `a, b = TypeUnder(a), TypeUnder(b)` before the `switch` statement?

runtime/ztests/expr/function/quiet.yaml
```diff
@@ -0,0 +1,19 @@
+zed: quiet(this)
```
nwt, line 1: Nit: Merge this file with runtime/ztests/op/yield-quiet.yaml since they contain equivalent Zed programs.

runtime/vam/op/yield.go
```diff
@@ -63,3 +58,59 @@ func interleave(vals []vector.Any) vector.Any {
...
+			index = append(index, k)
+		}
+	}
+	if len(index) == 0 {
+		return nil
+	}
+	if len(index) == int(n) {
+		return vec
+	}
+	return vector.NewView(index, vec)
```
nwt, line 85:
```suggestion
	masked, _ := applyMask(vec, mask)
	return masked
```

runtime/vam/expr/function/errors.go
```diff
@@ -0,0 +1,34 @@
...
+	if _, ok := arg.Typ.Type.(*zed.TypeOfString); !ok {
+		return args[0]
+	}
+	n := arg.Len()
+	vec := vector.NewStringEmpty(n, vector.NewBoolEmpty(n, nil))
+	for i := uint32(0); i < n; i++ {
+		s, isnull := vector.StringValue(arg.Vals, i)
+		if isnull {
+			vec.Nulls.Set(i)
+		}
```
nwt, line 27: Nit: Just use arg.Nulls.
```suggestion
	vec := vector.NewStringEmpty(n, arg.Nulls)
	for i := uint32(0); i < n; i++ {
		s, _ := vector.StringValue(arg.Vals, i)
```

nwt (approved):

PR #5306, opened 2024-09-25
vam: Add network_of() func

mattnibs:


runtime/vam/expr/function/fields.go
```diff
@@ -47,11 +47,9 @@ func (f *Fields) Call(args ...vector.Any) vector.Any {
...
 		}
 		inner := vector.NewArray(f.innerTyp, inOffs, s, nil)
-		out := vector.NewArray(f.outerTyp, outOffs, inner, nil)
-		if len(errs) > 0 {
-			return vector.Combine(out, errs, vector.NewStringError(f.zctx, "missing", uint32(len(errs))))
-		}
-		return out
+		b := vector.NewCombiner(vector.NewArray(f.outerTyp, outOffs, inner, nil))
+		b.Add(errs, vector.NewStringError(f.zctx, "missing", uint32(len(errs))))
+		return b.Result()
```
nwt, line 52: vector.Combine feels like a better API here than vector.NewCombiner since the latter allocates a vector.Error even when len(errs)==0. Why not stick with it?

mattnibs: Doesn't work for multiple error values

nwt: Then how about using an API that works for multiple error values when you need that and an API that doesn't do unnecessary allocations when you don't?

vector/dict.go
```diff
@@ -17,6 +19,14 @@ func NewDict(vals Any, index []byte, counts []uint32, nulls *Bool) *Dict {
 	return &Dict{vals, index, counts, nulls}
 }
 
+func NewDictBuildCounts(vals Any, index []byte, nulls *Bool) *Dict {
+	counts := make([]uint32, vals.Len())
+	for _, slot := range index {
+		counts[slot]++
+	}
+	return &Dict{vals, index, counts, nulls}
+}
```
nwt, line 28: Instead of a new constructor, how about changing NewDict to build counts if that parameter is nil?

```go
func NewDict(vals Any, index []byte, counts []uint32, nulls *Bool) *Dict {
	if counts == nil {
		counts = make([]uint32, vals.Len())
		for _, slot := range index {
			counts[slot]++
		}
	}
	return &Dict{vals, index, counts, nulls}
}
```

runtime/vam/expr/function/network_of.go
```diff
@@ -0,0 +1,218 @@
...
+		return vector.Combine(out, errs, errNotIP4(n.zctx, vector.NewView(errs, vec)))
+	}
+	return out
+}
+
+func (n *NetworkOf) singleIPLoop(vec *vector.IP, sel *vector.Selection) ([]netip.Prefix, []uint32) {
+	size := vec.Len()
+	var nets []netip.Prefix
+	var errs []uint32
+	for i := uint32(0); i < size; i++ {
```
nwt, line 69:
```suggestion
	var nets []netip.Prefix
	var errs []uint32
	for i := range vec.Len() {
```

(This evaluates `vec.Len()` once.)

runtime/vam/expr/function/network_of.go
```diff
@@ -0,0 +1,218 @@
...
+	default:
+		return 24
+	}
+}
+
+func (n *NetworkOf) ipMask(ipvec, maskvec vector.Any) vector.Any {
+	var nets []netip.Prefix
+	var errsLen, errsCont []uint32
+	size := ipvec.Len()
+	for i := uint32(0); i < size; i++ {
```
nwt, line 97:
```suggestion
	for i := ipvec.Len() {
```

runtime/vam/expr/function/network_of.go
```diff
@@ -0,0 +1,218 @@
...
+			if invalidNetMask(ip, int(bits)) {
+				return errCIDRRange(n.zctx, ipvec, maskvec)
+			}
+			net := netip.PrefixFrom(ip, int(bits)).Masked()
+			return vector.NewConst(zed.NewNet(net), ipvec.Len(), nil)
+		}
+		nets, errs = n.intMaskFast(ipvec, int(bits))
+	} else {
+		size := ipvec.Len()
+		for i := uint32(0); i < size; i++ {
```
nwt, line 135:
```suggestion
		for i := ipvec.Len() {
```

runtime/vam/expr/function/network_of.go
```diff
@@ -0,0 +1,218 @@
...
+	default:
+		panic(vec)
+	}
+}
+
+func (n *NetworkOf) intMaskFastLoop(vec *vector.IP, sel *vector.Selection, bits int) ([]netip.Prefix, []uint32) {
+	var errs []uint32
+	var nets []netip.Prefix
+	size := vec.Len()
+	for i := uint32(0); i < size; i++ {
```
nwt, line 183:
```suggestion
	for i := vec.Len() {
```

vector/combine.go
```diff
@@ -0,0 +1,68 @@
...
+		baseVecs = []Any{c.base}
+		baseTags = make([]uint32, c.base.Len())
+	}
+	size := c.base.Len()
+	for _, vec := range c.vecs {
+		size += vec.Len()
+	}
+	tags := make([]uint32, int(size))
+	n := uint32(len(baseVecs))
+	for i := uint32(0); i < size; i++ {
```
nwt, line 52:
```suggestion
	for i := range size {
```

vector/dict_test.go
```diff
@@ -0,0 +1,26 @@
...
+	s := vector.NewStringEmpty(0, nil)
+	s.Append("foo")
+	s.Append("bar")
+	s.Append("baz")
+	index := []byte{0, 1, 2, 0, 1, 2, 0, 1, 2}
+	d := vector.NewDictBuildCounts(s, index, nil)
+	var dropped []uint32
+	d, dropped = d.DeleteEntries(0, 2)
+	var vals []string
+	for i := uint32(0); i < d.Len(); i++ {
```
nwt, line 20:
```suggestion
	for i := range d.Len() {
```

vector/dict.go
nwt: On second thought, since it looks like you want this solely for TestDictDeleteEntries, I don't think we should add it. Just use a composite literal in the test or build the slice there.

runtime/vam/expr/function/network_of.go
```diff
@@ -0,0 +1,218 @@
...
+
+func errCIDRRange(zctx *zed.Context, ipvec, maskvec vector.Any) vector.Any {
+	vec := addressAndMask(zctx, ipvec, maskvec)
+	return vector.NewWrappedError(zctx, "network_of: CIDR bit count out of range", vec)
+}
+
+func intMaskValid(ip netip.Addr, bits int) bool {
+	return bits > 128 || bits > 32 && ip.Is4()
+}
+
```
nwt, line 207: This is unused.
```suggestion
```

vector/dict.go
mattnibs: I also need to know the indexes that we dropped so I can construct the error view.

mattnibs: > On second thought, since it looks like you want this solely for TestDictDeleteEntries, I don't think we should add it. Just use a composite literal in the test or build the slice there.

I don't follow this. I want this because I use it network_of for intMastFast

mattnibs: Fair

mattnibs: I do think we're going to want to have this in `NewDict` as you suggested but we'll wait for that day.

mattnibs: Actually I'm being totally silly about this. We don't want DeleteEntries here of ips we want to create a new nets dictionary. I'm reworking this and this will be removed.

mattnibs: ok fixed.

runtime/vam/expr/function/network_of.go
```diff
@@ -0,0 +1,213 @@
...
+		netVals, derrs := n.singleIPLoop(vec.Any.(*vector.IP), nil)
+		index, counts, nulls := vec.Index, vec.Counts, vec.Nulls
+		if len(derrs) > 0 {
+			index, counts, nulls, errs = vec.RebuildDropTags(derrs...)
+		}
+		nets = vector.NewDict(netVals, index, counts, nulls)
+	}
+	if len(errs) > 0 {
+		outer := vector.Combine(nets, errs, errNotIP4(n.zctx, vector.NewView(errs, vec)))
+		return outer
```
nwt, line 61: Nit:
```suggestion
		return vector.Combine(nets, errs, errNotIP4(n.zctx, vector.NewView(errs, vec)))
```

runtime/vam/expr/function/network_of.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+	b := vector.NewCombiner(vector.NewNet(nets, nil))
+	m := addressAndMask(n.zctx, ipvec, maskvec)
+	b.WrappedError(n.zctx, errsLen, "network_of: address and mask have different lengths", m)
+	b.WrappedError(n.zctx, errsCont, "network_of: mask is non-contiguous", maskvec)
+	return b.Result()
+}
+
+func (n *NetworkOf) intMask(ipvec, maskvec vector.Any) vector.Any {
+	id := maskvec.Type().ID()
```
nwt, line 118: Nit: Move this into the else branch.

vector/dict.go
```diff
@@ -14,6 +14,9 @@ type Dict struct {
 var _ Any = (*Dict)(nil)
 
 func NewDict(vals Any, index []byte, counts []uint32, nulls *Bool) *Dict {
+	if counts == nil {
+
+	}
```
nwt, line 19:
```suggestion
```

runtime/vam/expr/function/network_of.go
```diff
@@ -0,0 +1,213 @@
...
+		return outer
+	}
+	return nets
+}
+
+func (n *NetworkOf) singleIPLoop(vec *vector.IP, sel *vector.Selection) (*vector.Net, []uint32) {
+	var nets []netip.Prefix
+	var errs []uint32
+	for i := range vec.Len() {
+		idx := sel.Get(i)
```
nwt, line 70: Nit: vector.Selection saves three short lines here. I don't think it's worth its weight.
```suggestion
func (n *NetworkOf) singleIPLoop(vec *vector.IP, index []uint32) (*vector.Net, []uint32) {
	var nets []netip.Prefix
	var errs []uint32
	for i := range vec.Len() {
		idx := i
		if index != nil {
			idx = index[i]
		}
```

runtime/vam/expr/function/network_of.go
```diff
@@ -0,0 +1,213 @@
+package function
```
nwt, line 1: Nit: We don't use underscores in Go file names (except for test files and build constraints).

runtime/vam/expr/function/network_of.go
```diff
@@ -0,0 +1,213 @@
...
+	var out vector.Any
+	if c, ok := maskvec.(*vector.Const); ok {
+		bits, _ := c.AsInt()
+		if _, ok := ipvec.(*vector.Const); ok {
+			ip, _ := vector.IPValue(ipvec, 0)
+			if invalidNetMask(ip, int(bits)) {
+				return errCIDRRange(n.zctx, ipvec, maskvec)
+			}
+			net := netip.PrefixFrom(ip, int(bits)).Masked()
+			return vector.NewConst(zed.NewNet(net), ipvec.Len(), nil)
```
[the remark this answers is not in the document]
mattnibs: Nice!

nwt (approved):
:shipit:!!!

PR #5307, opened 2024-09-25
vam: Initial support for summarize

mattnibs:
This commit adds initial support of summarize in the vector runtime.
Currently summarize is limited to use of the count() aggregator only.


compiler/kernel/vop.go
```diff
@@ -150,12 +152,14 @@ func (b *Builder) compileVamLeaf(o dag.Op, parent vector.Puller) (vector.Puller,
 		}
 		return vamop.NewSort(b.rctx, parent, sortExprs, o.NullsFirst, o.Reverse, b.resetters), nil
 	case *dag.Summarize:
-		if name, ok := optimizer.IsCountByString(o); ok {
-			return vamop.NewCountByString(b.zctx(), parent, name), nil
-		} else if name, ok := optimizer.IsSum(o); ok {
-			return vamop.NewSum(b.zctx(), parent, name), nil
+		// XXX This only works if the key is a string which is unknowable at
```
mccanne, line 155: Please uncomment this so I can run in demo.  It's only a problem if you put vectors in a pool which none of our users are doing yet.

runtime/vam/op/summarize/summarize.go
```diff
@@ -0,0 +1,110 @@
...
+			return nil, err
+		}
+		if vec == nil {
+			for _, t := range s.tables {
+				s.results = append(s.results, t)
+			}
+			s.tables = nil
+			return s.next(), nil
+		}
+		// XXX Eval of keys and vals can be done in parallel.
```
mccanne, line 64: I think this was my comment but not clear we want to parallelize here if we want to keep the morsel-based scheduling philosophy.  Would be easy to measure both paths.  Maybe update the comment here?

runtime/vam/op/summarize/summarize.go
```diff
@@ -0,0 +1,110 @@
...
+		var keys, vals []vector.Any
+		for _, e := range s.keyExprs {
+			keys = append(keys, e.Eval(vec))
+		}
+		for _, e := range s.aggs {
+			vals = append(vals, e.Eval(vec))
+		}
+		vector.Apply(false, func(args ...vector.Any) vector.Any {
+			s.consume(args[:len(keys)], args[len(keys):])
+			// XXX Perhaps there should be a "consume" version of Apply where
```
mccanne, line 74: Or change Apply to handle nil return value?

mccanne (approved):
Let's work on the fast paths and full suite of agg funcs!

compiler/kernel/vop.go
mattnibs: The problem with this is it causes count by any single argument type other than string to fail. I disabled this because it was causing these tests to fail. I already have a separate branch that adds this back but in runtime. How about we merge this and post up this branch for review shortly after?

runtime/vam/expr/agg.go
```diff
@@ -0,0 +1,53 @@
+package expr
+
+import (
+	zed "github.com/brimdata/super"
+	"github.com/brimdata/super/runtime/vam/expr/agg"
+	"github.com/brimdata/super/vector"
+)
+
+type Aggregator struct {
```
nwt, line 9: Nit: Name this file aggregator.go to match this name.

PR #5312, opened 2024-10-01
vector.View: rebuild nulls

mattnibs:
The commit changes the behavior when wrapping Const, Errors and Unions. Consts and Errors are now rebuilt rather than getting wrapped in a View. For all these values if there is a non-nil nulls value the Bool vector is rebuilt.

### DIFF
--- a/vector/view.go
+++ b/vector/view.go
@@ -46,16 +46,16 @@ func nullsView(nulls *Bool, index []uint32) *Bool {
 	if nulls == nil {
 		return nil
 	}
-	out := NewBoolEmpty(uint32(len(index)), nil)
+	var out *Bool
 	for k, slot := range index {
 		if nulls.Value(slot) {
+			if out == nil {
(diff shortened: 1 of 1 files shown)


vector/view.go
```diff
@@ -38,6 +42,22 @@ func NewView(index []uint32, val Any) Any {
...
+	out := NewBoolEmpty(uint32(len(index)), nil)
+	for k, slot := range index {
+		if nulls.Value(slot) {
+			out.Set(uint32(k))
+		}
+	}
+	if out.Len() > 0 {
+		return out
+	}
+	return nil
```
nwt, line 58: Nit: Maybe allocate a vector.Bool only when necessary?
```suggestion
	var out *vector.Bool
	for k, slot := range index {
		if nulls.Value(slot) {
			if out == nil {
				out = NewBoolEmpty(uint32(len(index)), nil)
			}
			out.Set(uint32(k))
		}
	}
	return out
```

nwt (approved):

PR #5320, opened 2024-10-07
vng: Use uint64 for Data Lengths

mattnibs:


vng/writer.go
```diff
@@ -63,7 +63,7 @@ func (w *Writer) finalize() error {
 	zw.EndStream()
 	metaSize := zw.Position()
 	// Header
-	if _, err := w.writer.Write(Header{Version, uint32(metaSize), uint32(dataSize)}.Serialize()); err != nil {
+	if _, err := w.writer.Write(Header{Version, uint64(metaSize), uint64(dataSize)}.Serialize()); err != nil {
```
nwt, line 66: Nit: dataSize is a uint64.
```suggestion
	if _, err := w.writer.Write(Header{Version, uint64(metaSize), dataSize}.Serialize()); err != nil {
```

nwt (approved):

PR #5327, opened 2024-10-08
Implement vector over operator (minus "with" clause)

nwt:
This doesn't include the "with" clause because it requires Zed variables and we don't have those in runtime/vam yet.

Depends on #5325.


runtime/vam/op/over.go
```diff
@@ -0,0 +1,163 @@
...
+	for {
+		if len(o.vecs) == 0 || o.idx >= o.vecs[0].Len() {
+			vec, err := o.parent.Pull(done)
+			if vec == nil || err != nil {
+				return nil, err
+			}
+			o.vecs = o.vecs[:0]
+			for _, e := range o.exprs {
+				vec2 := e.Eval(vec)
+				vec2 = vector.Apply(true, func(vecs ...vector.Any) vector.Any { return vecs[0] }, vec2)
```
mattnibs, line 40: What is the point of the apply here?

nwt: `e.Eval(vec)` can return a vector.Dynamic containing vector.Unions, and over needs to "deunion" them. Is there any other way than this?

For completeness, here's the failure that appears without this.

```console
$ go test . -run /runtime/ztests/op/
--- FAIL: TestZed (0.23s)
    --- FAIL: TestZed/runtime/ztests/op (0.01s)
        --- FAIL: TestZed/runtime/ztests/op/over-union (0.01s)
            ztest.go:409: runtime/ztests/op/over-union.yaml: === vector ===
                expected and actual output differ:
                --- expected
                +++ actual
                @@ -1,6 +1,5 @@
                 1
                 "foo"
                -3
                -4
                -"bar"
                +[3,4]((ip,[int64]))
                +"bar"((int64,string))

FAIL
FAIL	github.com/brimdata/zed	0.829s
FAIL
```

mattnibs: Makes sense

mattnibs (approved):

PR #5329, opened 2024-10-09
Move runtime/ztests/op/sort/*.yaml to runtime/ztests/op

nwt:
All other tests under runtime/ztests/op live directly in that directory and have names that begin with the operator under test.


mattnibs:
IMHO when an op has a lot of tests as sort does I like having it in a directory so I don't have a huge list of tests to look through

PR #5332, opened 2024-10-10
Implement vector drop operator

nwt:
As part of this, convert the script ztest in
runtime/sam/expr/ztests/drop-nested.yaml to three separate Zed ztests with "vector: true" in runtime/ztests/op/drop-nested-{1,2,3}.yaml.


runtime/vam/expr/dropper.go
```diff
@@ -0,0 +1,127 @@
...
+			return nil, true
+		}
+		newRecType := d.zctx.MustLookupTypeRecord(newFields)
+		return vector.NewRecord(newRecType, newVecs, vec.Len(), vec.Nulls), true
+	case *vector.Dict:
+		if newVec, ok := d.drop(vec.Any, fm); ok {
+			return vector.NewDict(newVec, vec.Index, vec.Counts, vec.Nulls), true
+		}
+	case *vector.View:
+		if newVec, ok := d.drop(vec.Any, fm); ok {
```
mattnibs, line 104: Though this is slick, if you we do have a large filter here you're going to be doing a lot of extra work. Consider passing the index in here?

mattnibs (approved):
Nice

PR #5336, opened 2024-10-10
Rewrite top-level docs page with "super" naming

philrz:
A rendered draft is staged at https://spiffy-gnome-8f2834.netlify.app/docs/next

## Goals in renaming as affects docs

We know we want to:

1. Allow our existing users to be able to continue using the tools as they have been, even as the project/command names change (so, even as SuperSQL becomes the hot new item, we still want materials about SuperPipe, the data formats, etc., to still be readable and accurate even in the new naming scheme)
2. Make way to begin introducing new concepts like SuperSQL

## What I'm attempting here

This is a small step in the overall renaming effort to try out the new "super"-centric names.

We know the renaming is not a simple search-and-replace of "zed" with "super", so I wanted to confirm that the current proposed from/to list feels ok when applied to the existing materials. I've started with a rewrite of the first page users see when they land at https://zed.brimdata.io/docs/next

I've done some minimal reworking of sentence structure and moving paragraphs around where it felt necessary, but I've kept the overall content intact.

If this passes muster my hope is that we can begin re-using the new terms in similar ways as we attack more docs pages, READMEs in the repos, web site content, etc.

## What I'm **not** attempting (yet)

In this pass I'm intentionally **not** trying to:

1. Overhaul the top-level positioning
2. Introduce SuperSQL

...though I fully expect these things will happen in subsequent passes.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -3,101 +3,102 @@ sidebar_position: 1
 sidebar_label: Introduction
 ---
 
-# The Zed Project
+# SuperDB
 
-Zed offers a new approach to data that makes it easier to manipulate and manage
-your data.
-
-With Zed's new [super-structured data model](formats/README.md#2-zed-a-super-structured-pattern),
+SuperDB offers a new approach that makes it easier to manipulate and manage
+your data.  With its [super-structured data model](formats/README.md#2-zed-a-super-structured-pattern),
 messy JSON data can easily be given the fully-typed precision of relational tables
 without giving up JSON's uncanny ability to represent eclectic data.
 
 ## Getting Started
 
-Trying out Zed is easy: just [install](install.md) the command-line tool
-[`zq`](commands/zq.md) and run through the [zq tutorial](tutorials/zq.md).
+Trying out SuperDB is easy: just [install](install.md) the command-line tool
+[`super`](commands/zq.md) and run through the [tutorial](tutorials/zq.md).
 
-`zq` is a lot like [`jq`](https://stedolan.github.io/jq/)
-but is built from the ground up as a search and analytics engine based
-on the [Zed data model](formats/zed.md).  Since Zed data is a
-proper superset of JSON, `zq` also works natively with JSON.
+Compared to putting JSON data in a relational column, the
+[super-structured data model](formats/zed.md) makes it really easy to
+mash up JSON with your relational tables.  The `super` command is a little
+like [DuckDB](https://duckdb.org/) and a little like
+[`jq`](https://stedolan.github.io/jq/) but super-structured data ties the
+two patterns together with strong typing of dynamic values.
 
-While `zq` and the Zed data formats are production quality, the Zed project's
-[Zed data lake](commands/zed.md) is a bit [earlier in development](commands/zed.md#status).
-
-For a non-technical user, Zed is as easy to use as web search
-while for a technical user, Zed exposes its technical underpinnings
+For a non-technical user, SuperDB is as easy to use as web search
+while for a technical user, SuperDB exposes its technical underpinnings
 in a gradual slope, providing as much detail as desired,
 packaged up in the easy-to-understand
-[ZSON data format](formats/zson.md) and
-[Zed language](language/README.md).
+[Super JSON data format](formats/zson.md) and
+[SuperPipe language](language/README.md).
+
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -3,101 +3,102 @@ sidebar_position: 1
...
+[`super`](commands/zq.md) and run through the [tutorial](tutorials/zq.md).
 
-While `zq` and the Zed data formats are production quality, the Zed project's
-[Zed data lake](commands/zed.md) is a bit [earlier in development](commands/zed.md#status).
+Compared to putting JSON data in a relational column, the
+[super-structured data model](formats/zed.md) makes it really easy to
+mash up JSON with your relational tables.  The `super` command is a little
+like [DuckDB](https://duckdb.org/) and a little like
+[`jq`](https://stedolan.github.io/jq/) but super-structured data ties the
+two patterns together with strong typing of dynamic values.
```
philrz, line 23: This was some new content proposed by @mccanne when he responded on Slack after seeing an early draft. Indeed, even as I was doing the first pass I knew we'd need to move beyond just the `jq` comparison, but I didn't have the confidence to start launching into SQL-centric content on my own, so happy to include this!

docs/README.md
```diff
@@ -3,101 +3,102 @@ sidebar_position: 1
...
+while for a technical user, SuperDB exposes its technical underpinnings
 in a gradual slope, providing as much detail as desired,
 packaged up in the easy-to-understand
-[ZSON data format](formats/zson.md) and
-[Zed language](language/README.md).
+[Super JSON data format](formats/zson.md) and
+[SuperPipe language](language/README.md).
+
+While `super` and its accompanying data formats are production quality, the project's
+[SuperDB data lake](commands/zed.md) is a bit [earlier in development](commands/zed.md#status).
```
philrz, line 33: I intentionally moved this paragraph to the bottom of this section. One of my challenges in this rewrite was the contrast between SuperDB-the-project (which covers everything production quality that's available right now) and "SuperDB data lake" / `super db`-the-command (which we've acknowledged will be de-emphasized in the short term... but they still invoke the words "super db"). So I want to avoid sending mixed message.

While I would not advocate amputating lake coverage from the docs (especially since we have users relying on it in production!) I'm sensing that moving lake materials to less prominent places may help reduce confusion among new readers.

docs/README.md
```diff
@@ -3,101 +3,102 @@ sidebar_position: 1
...
-[sequential (ZNG)](formats/zng.md), [columnar (VNG)](formats/vng.md),
-and [human-readable (ZSON)](formats/zson.md) formats that all adhere to the
-same abstract Zed data model.
-* A [Zed lake](commands/zed.md) is a collection of Zed data stored
-across one or more [data pools](commands/zed.md#data-pools) with ACID commit semantics and
-accessed via a [Git](https://git-scm.com/)-like API.
-* The [Zed language](language/README.md) is the system's pipeline language for performing
+* The [super data model](formats/zed.md) is the abstract definition of the data types and semantics
+that underlie the super-structured data formats.
+* The [super data formats](formats/README.md) are a family of
```
philrz, line 41: In a first iteration I had these as "super-structured data model" and "super-structured data formats", but I recognized it was a lot of words and could get unwieldy considering how often we mention the data model in the docs. @mccanne was in favor of dropping the "-structured" and keeping it lowercase "s". We agreed that saying "SuperDB" in this spot would feel odd, and "Super" with a capital "S" is not something we're attempting anywhere else, so "super" basically wins by default at that point.

docs/README.md
```diff
@@ -3,101 +3,102 @@ sidebar_position: 1
...
-and online help.  The same language query that works for `zq` operating
-on local files or streams also works for `zed query` operating on a lake.
+The [`super db` documentation](commands/zed.md)
+is the best way to learn about the SuperDB data lake.
+All of its examples use `super db` commands run on the command line.
+Run `super db -h` or `-h` with any subcommand for a list of command options
+and online help.  The same language query that works for `super` operating
+on local files or streams also works for `super db query` operating on a lake.
 
 ## Design Philosophy
```
philrz, line 73: This whole paragraph basically helps justify the lake and therefore should perhaps move or be dropped entirely, but I intentionally did not attempt that in this pass.

mccanne (approved):
Looks great!  Let's merge this and and keep iterating.

PR #5346, opened 2024-10-17
rename zed/zq to super

mccanne:
This commit changes the command structure from "zed" to "super". The zq command is now "super query" and the zed command is now "super db".  "super query" no longer attempts to differentiate a single argument file from a single argument query and instead has a flag "-c" for specifying the query text.  "super db query" does not have a "-c" flag since there are no command-line file args.

The ZED_LAKE environement is now SUPER_LAKE.

There is still work to transition from zed terminology both in the source code as well as documentation but here we are getting all of the automated and docs tests passing.

We will rename the file formats in a subsequent PR.

Some of us will miss typing zq in place of "super query" but shell aliases are your friend, e.g., sup='super query'


philrz:
Two suggestions for your consideration:

1. For the env var, use `SUPER_DB_LAKE` instead of `SUPER_LAKE`. The rationale:
   *  `super db` is the command for all lake operations.
   * We've started saying "SuperDB data lake" in docs text, but `SUPER_DB_DATA_LAKE` is pretty darn long. 😛
   *  I'm instinctively flagging whenever we say just "super" since we've reached consensus that what we were doing before with overloading "zed" all by itself had its drawbacks, so I wanna make sure we're really passionate and selective about the times we choose to do it.

2. For the SuperPipe query-in-a-file, perhaps an extension other than `.pipe`? At one time I suggested `.spq` (pronounced "spike") and some seemed to like it. The rationale:
   * I'm instinctively drawn to something that's a three-letter alternative to the common `.sql`.
   * It contains the letters `s` and `q`, but is visually distinctive from `.sql` because the `l` is tall while the `q` dips down.
   * When I saw `.pipe` in the PR my first thought was Unix named pipes and `mkfifo`. (Maybe few others would make this connection and it wouldn't come up often enough to create wide confusion, though.)

mccanne:
There is still a ton of reworking to change the zed names but this changes the command names and updates all the tests...

docs/tutorials/join.md
```diff
@@ -134,34 +134,34 @@ produces
...
-```mdtest-input inner-join-pools.zed
+```mdtest-input inner-join-pools.spq
 from fruit
 | inner join (
   from people
 ) on flavor=likes eater:=name
 ```
 
-Populating the pools, then executing the Zed script:
+sPopulating the pools, then executing the query:
```
philrz, line 155:
```suggestion
Populating the pools, then executing the query:
```

philrz:
I noticed that as part of these changes  most (but not quite all) of the `.ndjson` file extensions were changed to `.json`. I respect these file extensions are just a hint and the two formats are very close so this is effectively splitting hairs. But I know of at least one example (#3875) where a community user went out of their way to point out where we were being loose in the difference between the two because there are downstream tools that have separate handling for the two formats, so this is part of why I usually go out of my way in docs/etc. to be explicit when we can about what's JSON vs. NDJSON. If we're in a hurry to get this merged I can follow up on the topic later though.

philrz (approved):

PR #5354, opened 2024-10-20
remove ast.Trunk

mccanne:
This commit simplifies the AST by removing the ast.Trunk object and replacing it with ast.Seq.  This is in preparation for unifying the SuperPipe and SQL from semantics.


mattnibs (approved):
Bravo

PR #5362, opened 2024-10-23
sam: Primitive cast fixes

mattnibs:
This commit makes several changes to primitive casting in sequential runtime in preparation for primitive casting in the vector runtime:
- Fix numeric to string casting (don't include type annotation in string value)
- Check for negative values when casting from float to uint
- Move coerce.To* functions to cast package. Have these functions require cast type and do bounds checking.
- Introduct cast.Overflow functions which determine if overflow checks are necessary when cast from numeric to numeric type. This is necessary because in vector runtime it is helpful to skip overflow check if it is not needed.


runtime/sam/expr/agg/avg.go
```diff
@@ -5,7 +5,7 @@ import (
 	"fmt"
 
 	"github.com/brimdata/super"
-	"github.com/brimdata/super/runtime/sam/expr/coerce"
+	"github.com/brimdata/super/runtime/sam/expr/cast"
```
nwt, line 8: Nit: I think you should leave this stuff in runtime/sam/expr/coerce. "Coerce" is a more evocative name than "cast". And it's one less package.

runtime/sam/expr/expr_test.go
```diff
@@ -501,6 +501,9 @@ func TestCasts(t *testing.T) {
 	testSuccessful(t, "uint8(-1)", "", `error({message:"cannot cast to uint8",on:-1})`)
 	testSuccessful(t, "uint8(300)", "", `error({message:"cannot cast to uint8",on:300})`)
 	testSuccessful(t, `uint8("foo")`, "", `error({message:"cannot cast to uint8",on:"foo"})`)
+	testSuccessful(t, `uint8("-1")`, "", `error({message:"cannot cast to uint8",on:"-1"})`)
+	testSuccessful(t, "uint8(this)", "258.", `error({message:"cannot cast to uint8",on:258.})`)
```
nwt, line 505: Nit:
```suggestion
	testSuccessful(t, "uint8(258.)", "", `error({message:"cannot cast to uint8",on:258.})`)
```

runtime/sam/expr/cast/overflow.go
```diff
@@ -0,0 +1,141 @@
+package cast
```
nwt, line 1: Nit: No need to export this stuff.

nwt (approved):

runtime/sam/expr/cast/overflow.go
mattnibs: For a follow up pr it needs to be exported.

PR #5366, opened 2024-10-24
refactor "super dev compile" into "super compile" and "super db compile"

mccanne:
This commit reparents the "super dev compile" command under the top-level "super" command and creates a parallel command "super db compile". This makes it much clearer as to when you are compiling a query for a lake vs a file system.

We also changed the compile commands to take a single query argument instead of concatenating multiple arguments.  Also, the command options are a little different: you now say "-dag" if you want the DAG output and get the AST output otherwise.

There were some inconsitencies in how commands were parented so we added charm.NoRun() and fixed up the command hierarchy linkages. The invariant should now be that every internal command properly appears in the command hierarchy and NoRun() is invoked by the Run() method of an interior command to give proper help.

The command-line help has also been updated.


cmd/super/compile/command.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/brimdata/super/pkg/charm"
+)
+
+var spec = &charm.Spec{
+	Name:  "compile",
+	Usage: "compile [ options ] spq|sql",
+	Short: "compile a local query for inspection and debugging",
+	Long: `
+This command parses a query and emits the resulting abstract syntax
+tree (AST) or runtime directed-acycle graph (DAG) in the output format desired.
```
nwt, line 16:
```suggestion
tree (AST) or runtime directed acyclic graph (DAG) in the output format desired.
```

cmd/super/compile/shared.go
```diff
@@ -0,0 +1,135 @@
...
+
+type valReader struct {
+	*super.Value
+}
+
+func (r *valReader) Read() (*super.Value, error) {
+	val := r.Value
+	r.Value = nil
+	return val, nil
+}
```
nwt, line 135:
```suggestion
	err = zio.CopyWithContext(ctx, writer, zbuf.NewArray([]super.Value{val}))
	if closeErr := writer.Close(); err == nil {
		err = closeErr
	}
	return err
}
```

compiler/ztests/from-error.yaml
```diff
@@ -1,24 +1,18 @@
 script: |
-  ! super dev compile -lake='' -C -s 'from p'
```
nwt, line 2: Should we keep this test case until you finish reworking the from operator?

cmd/super/compile/shared.go
mccanne: I was trying to remember that pattern.  Too much surfing LOL

nwt (approved):

PR #5367, opened 2024-10-25
update README for SuperDB

mccanne:

PR #5368, opened 2024-10-25
Super-fy the Search Expressions doc

philrz:
## What's Changing

This is a first pass at adapting the Search Expressions doc for the "super" era.

## Details

It's expected all docs pages will have to be somewhat revised to ease out the "Zed era" terminology and make way for the "SuperDB era". In an attempt to chip away at them incrementally, I'm starting from the pages hyperlinked from the revised README (#5367) and will fan out from there. I expect these may need more passes when SuperSQL is further along since right now the examples are all pure SuperPipe since that's what's working in `main`.

Like other early docs rewrite attempts, rather than replacing references to "Zed" with 1-for-1 references to something "super", I feel like I'm having some success with reworking sentences to drop the reference entirely.

Since I had to review everything, I made a few additional fixes and added some polish along the way.

### DIFF
--- a/docs/language/search-expressions.md
+++ b/docs/language/search-expressions.md
@@ -207,11 +207,11 @@ where grep("foo", this)
 ```
 
 Note that the "search" keyword may be omitted.
-For example, the simplest SuperPipe program is perhaps a single keyword search, e.g.,
+For example, the simplest SuperPipe query is perhaps a single keyword search, e.g.,
 ```
 foo
 ```
-As above, this program searches the implied input for values that
+As above, this query searches the implied input for values that
 contain the string "foo".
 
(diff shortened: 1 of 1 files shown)


docs/language/search-expressions.md
```diff
@@ -205,7 +207,7 @@ where grep("foo", this)
 ```
 
 Note that the "search" keyword may be omitted.
-For example, the simplest Zed program is perhaps a single keyword search, e.g.,
+For example, the simplest SuperPipe program is perhaps a single keyword search, e.g.,
```
mccanne, line 210: program -> query

mccanne (approved):

PR #5369, opened 2024-10-25
Super-fy the Formats README doc

philrz:
## What's Changing

This is a first pass at adapting the top-level Formats doc for the "super" era.

## Details

See #5368.


mccanne (approved):
This looks great.  It feels a little bit dated so let's merge this and iterate a bit more.

PR #5371, opened 2024-10-26
simplify locators in AST

mccanne:
This commit simplifies the construction of location information attached to each AST node.  The previous design was based on the Go compiler, which is necessarily more complicated because of its recursive descent and separate lexical analysis.  In our PEG grammar, each matching rule knows its precise location of the text in the input.  Thus, we can simplify things so every node simply attaches this info to a common ast.Loc implementation of the ast.Node interface.  This will pattern will be used in subequent PRs to implement the forthcoming SQL extensions.


mattnibs (approved):

compiler/parser/parser.peg
```diff
@@ -1624,7 +1603,7 @@ UnicodeEscape
     }
 
 RegexpPattern
-  = "/" body:RegexpBody "/" !KeyWordStart { return body, nil }
+  = "/" body:RegexpBody "/" !KeyWordStart { return body, nil } // XXX s/b Node
```
nwt, line 1606: What's "s/b" here?

mccanne: Oops, will delete!

PR #5373, opened 2024-10-26
unify name-y things in grammar

mccanne:
This commit unifies the various types of named things in the PEG grammar to a single pattern called ast.Name and gets rid of ast.String and ast.QuotedString.  Now that the locators are all computed from the underlying rule, we can easily have a single ast.Name that represents different variations of names.  These changes will be used by a subequent PR that will unify the get/file/pool operators into a single "from", which is in turn needed by the upcoming SuperSQL additions.


mccanne:
Note that the `decs.yaml` ztest breaks the canonical superpipe query text by wrapping the user-op arguments `left_file` and `right_file` in quotes.  The way file names and const identifiers are mixed here isn't robust but will be fixed soon in the `from-any` branch.  This isn't all that important (just the zfmt output) and it will be promptly fixed in a subsequent PR so I would like to merge "as is".

compiler/parser/parser.peg
```diff
@@ -1603,7 +1594,7 @@ UnicodeEscape
     }
 
 RegexpPattern
-  = "/" body:RegexpBody "/" !KeyWordStart { return body, nil }
+  = "/" body:RegexpBody "/" !KeyWordStart { return body, nil } // XXX s/b Node
```
nwt, line 1597:
```suggestion
  = "/" body:RegexpBody "/" !KeyWordStart { return body, nil }
```

nwt (approved):

compiler/parser/parser.peg
mccanne: Rebase auto-merge doh!

PR #5374, opened 2024-10-26
change super compile so pretty output is default 

mccanne:


cli/outputflags/flags.go
```diff
@@ -22,6 +22,7 @@ import (
 type Flags struct {
 	anyio.WriterOpts
 	DefaultFormat string
+	DefaultPretty bool
```
nwt, line 25: Nit: Call this DefaultZSONPretty since both JSON and ZSON can be pretty?

nwt (approved):

cli/outputflags/flags.go
mccanne: Hmm, good point.  I think the logic is off.  I'll look a little closer and write some tests.

mccanne: Here's the problem...
```
; super compile -f json 'from x'
cannot use -z or -Z with -f
```

mccanne:
I don't see any easy way to do what I wanted without messing up the generic output flags logic.  We probably need to rethink the logic to do what I want.  I'll live with typing -Z for now...

PR #5378, opened 2024-10-27
improvements to from operator

mccanne:
This commit improves the from operator and brings it into closer alignment with SQL.  We also deprecated pool, get, and file though they can still be used but appear as "from" when running super compile.

Targets of the from operator are now self describing and names are interpreted as files when running locally (super -c) and pools when running on a lake (super db).  URLs continue to work in both cases.

Const expressions continue to work but now need to be enclosed in brackets, e.g., "from [poolConst]".  This makes easy to distinguish a file reference e.g., "from foo.json", from a record deref. e.g., "from [this.pool]"  Currently these expressions must evaluate at compile time, but this sets us up to configure "from" to have a parent and make this all dynamic. Note you can also now say 'from ["a.json", "b.json"]', which will do a merge of the scans (not sequential eval).

The semantic pass now does existence checks of files and pools so "super compile" tests needed to be updated to "touch" files etc.

We will update docs to reflect these changes after we add the parent concept to "from" and have a bit of experience with the new syntax and semantics.


compiler/parser/parser.peg
```diff
@@ -294,7 +292,7 @@ AggAssignment
 Agg
   = !FuncGuard op:AggName __ "(" __ expr:(OverExpr / Expr)? __ ")" !(__ ".") where:WhereClause? {
       agg := &ast.Agg{
-          Kind:"Agg", 
+          Kind:"Agg",
```
nwt, line 295:
```suggestion
          Kind: "Agg",
```

zfmt/ast.go
```diff
@@ -640,6 +613,20 @@ func (c *canon) op(p ast.Op) {
...
+	switch e := e.(type) {
+	case *ast.ExprEntity:
+		c.expr(e.Expr, "")
+	case *ast.Glob, *ast.Regexp:
+		c.pattern(e)
+	case *ast.Name:
+		c.write(zson.QuotedName(e.Text))
+	default:
+		c.open("unknown from expression: %T", e)
+		c.close()
```
nwt, line 626: I think this should panic.

zfmt/ast.go
```diff
@@ -680,33 +667,28 @@ func (c *canon) scope(s *ast.Scope, parens bool) {
...
-		return p.Pattern
+		c.write(p.Pattern)
 	case *ast.Regexp:
-		return "/" + p.Pattern + "/"
-	case *ast.Name:
-		return zson.QuotedName(p.Text)
+		c.write("/" + p.Pattern + "/")
 	default:
-		return fmt.Sprintf("(unknown pattern type %T)", p)
+		c.write(fmt.Sprintf("(unknown pattern type %T)", p))
```
nwt, line 691: I think this should panic.

zfmt/ast.go
```diff
@@ -782,42 +764,37 @@ func IsSearch(e ast.Expr) bool {
...
+	switch args := args.(type) {
+	case *ast.PoolArgs:
+		c.poolArgs(args)
+	case *ast.HTTPArgs:
+		c.httpArgs(args)
+	case *ast.FormatArg:
+		c.formatArg(args)
 	default:
-		c.write("unknown source type: %T", src)
+		c.write("unknown argument type in from operaetor: %T", args)
```
nwt, line 798: I think this should panic.

compiler/parser/parser.peg
```diff
@@ -607,46 +605,87 @@ DebugOp
...
+  = "from"
+  / DeprecatedFroms
+
+DeprecatedFroms = "get" / "file" / "pool"
+
+FromBody
+  = entity:FromEntity args:FromArgs {
+      from := &ast.From{
+        Kind: "From",
+        Entity: entity.(ast.FromEntity),
```
nwt, line 624:
```suggestion
        Entity: entity.(ast.FromEntity),
        Loc: loc(c),
```

compiler/parser/parser.peg
```diff
@@ -607,46 +605,87 @@ DebugOp
...
-      h := &ast.HTTP{
-        Kind: "HTTP",
-        URL: url.(ast.Pattern),
+FromArgs
+  = commit:PoolCommit meta:PoolMeta? tap:TapArg {
+      return &ast.PoolArgs{
+        Kind: "PoolArgs",
+        Commit: nullableName(commit),
+        Meta: nullableName(meta),
+        Tap: tap.(bool),
```
nwt, line 665:
```suggestion
        Tap: tap.(bool),
        Loc: loc(c),
```

compiler/parser/parser.peg
```diff
@@ -607,46 +605,87 @@ DebugOp
...
+      }, nil
+    }
+  / meta:PoolMeta tap:TapArg {
+      return &ast.PoolArgs{
+        Kind: "PoolArgs",
+        Meta: nullableName(meta),
+        Tap: tap.(bool),
```
nwt, line 672:
```suggestion
        Tap: tap.(bool),
        Loc: loc(c),
```

compiler/dag/op.go
```diff
@@ -206,20 +206,18 @@ type (
 		SortKeys order.SortKeys `json:"sort_keys"`
 	}
 	FileScan struct {
-		Kind     string         `json:"kind" unpack:""`
-		Path     string         `json:"path"`
-		Format   string         `json:"format"`
-		SortKeys order.SortKeys `json:"sort_keys"`
```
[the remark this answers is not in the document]
mccanne: I agree that the system should be able to use sort keys from file objects but this doesn't belong in the query language.  Instead this should be part of the interface between the compiler and the file objects as it is with the pool adaptor.

compiler/semantic/op.go
```diff
@@ -33,156 +36,213 @@ func (a *analyzer) semSeq(seq ast.Seq) dag.Seq {
...
 	}
-	return append(out, &dag.Fork{Kind: "Fork", Paths: paths})
 }
 
-//XXX make sure you can't read files from a lake instance
+func (a *analyzer) semFile(nameLoc ast.Node, name string, args ast.FromArgs) dag.Op {
+	if _, err := os.Stat(name); err != nil {
+		a.error(nameLoc, fmt.Errorf("%q: not such file", name))
+		return badOp()
+	}
```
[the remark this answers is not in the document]
mccanne: Good question.  It keeps it symmetric with pool lookup (which we need to create dag nodes) and we will eventually need to lookup sort keys (as mentioned above) and types (for static typing checking) but we don't need that till optimization.  But maybe it's nice to run `compile -dag` without needing the file...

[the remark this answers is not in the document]
nwt: > But maybe it's nice to run compile -dag without needing the file...

That's what I was thinking.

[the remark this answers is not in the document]
mccanne: Removed!

nwt (approved):
:shipit:

PR #5379, opened 2024-10-27
add support for SQL-style case expressions

mccanne:
This commit adds support to the grammar for SQL-style case expressions. The case-when form is mapped to ast.Conditional and the case-expr-when form currently reports that it is unsupported in the semantic pass (we'll need to add runtime support for this other form).

This will be documented when we add docs for SuperSQL.

### DIFF
--- a/compiler/parser/parser.peg
+++ b/compiler/parser/parser.peg
@@ -1021,9 +1021,9 @@ CaseExpr
         whens := sliceOf[ast.When](cases)
         cond := &ast.Conditional{
             Kind: "Conditional",
-            Loc: loc(c),
             Cond: whens[0].Cond,
             Then: whens[0].Then,
+            Loc: loc(c),
         }
         last := cond
         for _, when := range whens[1:] {
@@ -1031,7 +1031,7 @@ CaseExpr
                Kind: "Conditional",
                Cond: when.Cond,
                Then: when.Then,
-               Loc: loc(c),
+               Loc: when.Loc,
             }
             last.Else = next
             last = next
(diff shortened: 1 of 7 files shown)


compiler/parser/parser.peg
```diff
@@ -998,6 +999,53 @@ Function
...
     }
 
+CaseExpr
+  = "case"i cases:When+ else_:(_ "else" _ Expr)? _ "end"i (_ "case"i)? {
+        whens := sliceOf[ast.When](cases)
+        cond := &ast.Conditional{
+            Kind: "Conditional",
+            Loc: loc(c),
+            Cond: whens[0].Cond,
+            Then: whens[0].Then,
```
nwt, line 1009:
```suggestion
            Cond: whens[0].Cond,
            Then: whens[0].Then,
            Loc: loc(c),
```

compiler/parser/parser.peg
```diff
@@ -998,6 +999,53 @@ Function
...
+        }
+        last := cond
+        for _, when := range whens[1:] {
+            next := &ast.Conditional{
+               Kind: "Conditional",
+               Cond: when.Cond,
+               Then: when.Then,
+               Loc: loc(c),
```
nwt, line 1017:
```suggestion
               Loc: when.Loc,
```

compiler/parser/parser.peg
```diff
@@ -998,6 +999,53 @@ Function
...
+        }
+        return cond, nil
+    }
+
+When
+  = _ "when"i __ cond:Expr _ "then"i _ then:Expr {
+      return ast.When{
+        Kind: "When",
+        Cond: cond.(ast.Expr),
+        Then: then.(ast.Expr),
```
nwt, line 1045:
```suggestion
```suggestion
        Then: then.(ast.Expr),
        Loc: loc(c),
```
```

PR #5380, opened 2024-10-28
clarify PEG grammar around lean-forward idioms of SuperPipe

mccanne:
This commit reorganizes the PEG grammar a bit to put all the pedantic SuperPipe operators under one rule and then to enumerate all of the syntatic shortcuts under the leaner LeanOp rule (which used to be Operation).  With this layout, it's really clear where the lean grammar diverges from the pedantic grammar.  We can subsequently use this pattern to make a lean SuperSQL alongside the pedantic SuperSQL.

We also deleted the Cast rule from the LeanOp location as it was redundant with the Expr rule for implied yield.  This undoes the change in #3697 but retains the test, which still passes.

### DIFF
--- a/compiler/parser/parser.peg
+++ b/compiler/parser/parser.peg
@@ -291,7 +291,7 @@ Operator
   / DebugOp
 
  ForkOp =
-   "fork" __ "(" paths:Leg+ __ ")" {
+   "fork" __ "(" paths:Path+ __ ")" {
       var seqs []ast.Seq
       for _, p := range paths.([]any) {
         seqs = append(seqs, sliceOf[ast.Op](p))
@@ -303,10 +303,10 @@ Operator
       }, nil
     }
 
-Leg = __ "=>" __ seq:Seq { return seq, nil }
+Path = __ "=>" __ seq:Seq { return seq, nil }
 
 SwitchOp
-  = "switch" _ expr:Expr _ "(" cases:SwitchLeg+ __ ")" {
+  = "switch" _ expr:Expr _ "(" cases:SwitchPath+ __ ")" {
       return &ast.Switch{
           Kind: "Switch",
           Expr: expr.(ast.Expr),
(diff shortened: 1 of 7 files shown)


compiler/parser/parser.peg
```diff
@@ -349,6 +290,78 @@ Operator
...
+        seqs = append(seqs, sliceOf[ast.Op](p))
+      }
+      return &ast.Parallel{
+          Kind: "Parallel",
+          Paths: seqs,
+          Loc: loc(c),
+      }, nil
+    }
+
+Leg = __ "=>" __ seq:Seq { return seq, nil }
```
nwt, line 306: Nit: Since we've moved away from the "leg" terminology.
```suggestion
Path = __ "=>" __ seq:Seq { return seq, nil }
```

compiler/parser/parser.peg
```diff
@@ -349,6 +290,78 @@ Operator
...
+    }
+  / "switch" __ "(" cases:SwitchLeg+ __ ")" {
+      return &ast.Switch{
+          Kind: "Switch",
+          Cases: sliceOf[ast.Case](cases),
+          Loc: loc(c),
+      }, nil
+    }
+
+SwitchLeg
```
nwt, line 325: Nit:
```suggestion
SwitchPath
```

compiler/parser/parser.peg
```diff
@@ -349,6 +290,78 @@ Operator
...
+FromForkOp
+  = "from" __ "(" trunks:FromLeg+ __ ")" {
+      return &ast.Parallel{
+          Kind: "Parallel",
+          Paths: sliceOf[ast.Seq](trunks),
+          Loc: loc(c),
+      }, nil
+    }
+
+FromLeg
```
nwt, line 347: Nit:
```suggestion
FromPath
```

PR #5383, opened 2024-10-28
vam: Add to cast primitive

mattnibs:


runtime/sam/expr/cast.go
```diff
@@ -268,7 +265,7 @@ func (c *casterType) Eval(ectx Context, val super.Value) super.Value {
 	}
 	typval, err := zson.ParseValue(c.zctx, val.AsString())
 	if err != nil {
-		return c.zctx.WrapError(err.Error(), val)
+		return c.zctx.WrapError("cannot cast to type", val)
 	}
 	if typval.Type().ID() != super.IDType {
 		return c.zctx.WrapError("cannot cast to type", val)
```
nwt, line 271:
```go
	if err != nil || typeval.Type().ID() != super.IDType {
		return c.zctx.WrapError("cannot cast to type", val)
```

ztest/ztest.go
```diff
@@ -273,6 +273,7 @@ type ZTest struct {
 	OutputFlags string `yaml:"output-flags,omitempty"`
 	ErrorRE     string `yaml:"errorRE,omitempty"`
 	Vector      bool   `yaml:"vector"`
+	VectorOnly  bool   `yaml:"vector_only"`
```
nwt, line 276: You're not adding any tests that use this.
```suggestion
```

mattnibs: whoops

vector/view.go
```diff
@@ -52,7 +52,7 @@ func NewView(index []uint32, val Any) Any {
 	return &View{val, index}
 }
 
-func nullsView(nulls *Bool, index []uint32) *Bool {
+func NullsView(index []uint32, nulls *Bool) *Bool {
```
nwt, line 55: Nit: I've been meaning to change the NewView signature to `NewView(vec Any, index []uint32) Any` because that feels more natural so you might want to leave this as is.

runtime/ztests/expr/cast/bool.yaml
```diff
@@ -0,0 +1,39 @@
...
+  null(bool)
+  true
+  true
+  false
+  false
+  error({message:"cannot cast to bool",on:"blah"})
+  null(bool)
+  true
+
+
```
nwt, line 39:
```suggestion
```

nwt (approved):
:shipit:

vector/view.go
nwt: #5395 makes the NewView change.

mattnibs: I just wanted to make this consistent with NewView so if you're going to change NewView I'll do as you say.

PR #5386, opened 2024-10-29
Super-fy the Super JSON format doc

philrz:
In the same spirit as #5368.

### DIFF
--- a/docs/formats/jsup.md
+++ b/docs/formats/jsup.md
@@ -149,7 +149,7 @@ conforming to any floating point representation that cannot be
 interpreted as an integer, e.g., `1.` or `1.0` instead of
 `1` or `1e3` instead of `1000`.  Unlike JSON, a floating point number can
 also be one of:
-`Inf`, `+Inf`, `-Inf`, or `NaN`.
+`+Inf`, `-Inf`, or `NaN`.
 
 A floating point value may be expressed with an integer string provided
 a type decorator is applied, e.g., `123 (float64)`.
--- a/docs/formats/jsup.md
+++ b/docs/formats/jsup.md
@@ -23,7 +23,7 @@ Super JSON values.
 A Super JSON text is a sequence of UTF-8 characters organized either as a bounded input
 or an unbounded stream.
 
-The input text is organized as a sequence of one or more Zed values optionally
+The input text is organized as a sequence of one or more values optionally
 separated by and interspersed with whitespace.
 Single-line (`//`) and multi-line (`/* ... */`) comments are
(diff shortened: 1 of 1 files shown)


docs/formats/jsup.md
```diff
@@ -149,7 +149,7 @@ conforming to any floating point representation that cannot be
 interpreted as an integer, e.g., `1.` or `1.0` instead of
 `1` or `1e3` instead of `1000`.  Unlike JSON, a floating point number can
 also be one of:
-`Inf`, `+Inf`, `-Inf`, or `Nan`.
+`Inf`, `+Inf`, `-Inf`, or `NaN`.
```
philrz, line 152: Turns out case is important here.

```
$ super -version
Version: v1.18.0-69-g5ab9615c

$ echo '1 NaN 2' | super -z -c 'yield typeof(this)' -
<int64>
<float64>
<int64>
```

The following causes a syntax error, which makes sense to me:

```
$ echo '1 Nan 2' | super -z -c 'yield typeof(this)' -
stdio:stdin: Super JSON syntax error
```

Not sure how to explain this though. A bug?

```
$ echo 'Nan' | super -z -c 'yield typeof(this)' -
[no output]
```

Also related, it looks like `Inf` requires the explicit `+` or `-` at the moment.

```
$ echo '1 +Inf -Inf 2' | super -z -c 'yield typeof(this)' -
<int64>
<float64>
<float64>
<int64>

$ echo '1 Inf 2' | super -z -c 'yield typeof(this)' -
stdio:stdin: Super JSON syntax error

$ echo 'Inf' | super -z -c 'yield typeof(this)' -
[no output]
```

Whereas `NaN` seemed like an easy docs fix since that's the way it's shown in Excel/etc., it's not clear to me if we'd rather add support for plain `Inf` to the grammar or drop it from the spec.

mccanne: This is a bug in the csup parser for any input that is a single identifier then EOF.  It shouldn't be too hard to fix.

mccanne: Regarding Inf, in IEEE floating point, there's a `+Inf` and a `-Inf`.  I think we should have one way to represent `+Inf` in Super JSON adn we could change `+Inf` to `Inf`.  I think we went with `+Inf` because the golang `fmt` library prints it that way.

philrz: Thanks @mccanne. I've opened #5391 to track the JSUP parser bug and #5392 to track possibly changing `+Inf` to just `Inf`.

PR #5388, opened 2024-10-30
add SuperSQL to SuperDB

mccanne:
This commit is a first cut at adding SQL support to SuperDB.

This initial work represents the world's first foundation for a SQL that blends traditional static analysis of relational queries (when types are available for the entities queried) with dynamic execution of polymorphic operators when types are unknown (like SQL++).

It's also the world's first implementation of an (eventually) SQL-compliant query language embedded within a pipeline-sytle language that allows aggressive abbrevitions for SQL idioms to provide lean-forward log-search style experience.  The pipeline language includes many new operators enabled by super-structured data that are not available in SQL.

Needless to say, this is also the world's first SQL for super-structured data.

Pipelines can be added to the end of any SQL expression and joins are limping along.  SQL unions do not work yet.

While this initial commit provides a very incomplete implementation of SQL, the foundations are plumbed into the pipeline syntax in a clean way and should serve as a solid foundation going forward.

It's also all setup for a "binder-in-the-data" approach to SQL scoping.  More on this coming soon.

### DIFF
--- a/compiler/ast/ast.go
+++ b/compiler/ast/ast.go
@@ -615,8 +615,8 @@ type SortExpr struct {
 	Kind  string `json:"kind" unpack:""`
 	Expr  Expr   `json:"expr"`
 	Order *ID    `json:"order"`
+	Nulls *ID    `json:"nulls"`
 	Loc   `json:"loc"`
-	Nulls *ID `json:"nulls"`
 }
 
 type Case struct {
--- a/compiler/ast/ast.go
+++ b/compiler/ast/ast.go
@@ -161,7 +161,7 @@ type Name struct {
 
 type FromEntity interface {
 	Node
-	fromEntityAST()
+	fromEntity()
 }
 
 type ExprEntity struct {
@@ -170,11 +170,28 @@ type ExprEntity struct {
 	Loc  `json:"loc"`
 }
 
-func (*Glob) fromEntityAST()       {}
-func (*Regexp) fromEntityAST()     {}
-func (*ExprEntity) fromEntityAST() {}
-func (*LakeMeta) fromEntityAST()   {}
-func (*Name) fromEntityAST()       {}
+func (*Glob) fromEntity()       {}
+func (*Regexp) fromEntity()     {}
+func (*ExprEntity) fromEntity() {}
+func (*LakeMeta) fromEntity()   {}
+func (*Name) fromEntity()       {}
+func (*CrossJoin) fromEntity()  {}
+func (*SQLJoin) fromEntity()    {}
+func (*SQLPipe) fromEntity()    {}
+
+type FromElem struct {
+	Kind       string      `json:"kind" unpack:""`
+	Entity     FromEntity  `json:"entity"`
+	Args       FromArgs    `json:"args"`
+	Ordinality *Ordinality `json:"ordinality"`
+	Alias      *Name       `json:"alias"`
+	Loc        `json:"loc"`
+}
+
+type Ordinality struct {
+	Kind string `json:"kind" unpack:""`
+	Loc  `json:"loc"`
+}
 
 type RecordExpr struct {
 	Kind  string       `json:"kind" unpack:""`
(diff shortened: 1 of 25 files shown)


compiler/ast/ast.go
```diff
@@ -599,6 +616,7 @@ type SortExpr struct {
 	Expr  Expr   `json:"expr"`
 	Order *ID    `json:"order"`
 	Loc   `json:"loc"`
+	Nulls *ID `json:"nulls"`
```
nwt, line 619: Nit:
```suggestion
	Nulls *ID `json:"nulls"`
	Loc   `json:"loc"`
```

zfmt/ast.go
```diff
@@ -613,6 +610,25 @@ func (c *canon) op(p ast.Op) {
...
+}
+
+func (c *canon) fromElem(elem *ast.FromElem) {
+	c.fromEntity(elem.Entity)
+	if elem.Args != nil {
+		c.fromArgs(elem.Args)
+	}
+	if elem.Alias != nil {
+		c.write(" ")
+		c.write(zson.QuotedName(elem.Alias.Text))
```
nwt, line 628: Nit:
```suggestion
		c.write(" %s", zson.QuotedName(elem.Alias.Text))
```

zfmt/ast.go
```diff
@@ -621,6 +637,10 @@ func (c *canon) fromEntity(e ast.FromEntity) {
 		c.pattern(e)
 	case *ast.Name:
 		c.write(zson.QuotedName(e.Text))
+	case *ast.CrossJoin:
+		c.write("cross join XXX")
+	case *ast.SQLJoin:
+		c.write("sql join XXX")
```
nwt, line 643: Are you deliberately leaving these unfinished?

compiler/semantic/op.go
```diff
@@ -36,15 +36,46 @@ func (a *analyzer) semSeq(seq ast.Seq) dag.Seq {
 }
 
 func (a *analyzer) semFrom(from *ast.From, out dag.Seq) dag.Seq {
-	sources := a.semFromEntity(from.Entity, from.Args)
-	if len(sources) == 1 {
-		return append(sources, out...)
+	if len(from.Elems) > 1 {
+		a.error(from, errors.New("cross join implied from multiple elements in from clause is not yet supported"))
```
nwt, line 40: Nit:
```suggestion
		a.error(from, errors.New("cross join implied by multiple elements in from clause is not yet supported"))
```

lake/ztests/query-parse-error.yaml
```diff
@@ -34,34 +34,34 @@ outputs:
   - name: stderr
     data: |
       =1=
-      error parsing SuperPipe at line 1, column 11:
+      error parsing SuperPipe at line 1, column 12:
       from test \ count()
-            === ^ ===
+             === ^ ===
```
nwt, line 39: Pointing at the space after the offending token is confusing and we should fix that.

compiler/semantic/sql.go
```diff
@@ -0,0 +1,458 @@
+package semantic
+
+import (
+	"errors"
+	"fmt"
+
+	zed "github.com/brimdata/super"
```
mattnibs, line 7: Use super instead of zed

nwt: 👍

lake/ztests/query-parse-error.yaml
nwt: @mccanne and I spoke about this and it's something we can look into later.

zfmt/ast.go
mccanne: yes

mccanne: I could take them out and get the panic.  -C for sql is a whole little project.  e.g., select isn't handled yet at all.

nwt: I'd go with panic but this is fine.

compiler/semantic/sql.go
```diff
@@ -0,0 +1,457 @@
...
+		if limit < 1 {
+			a.error(op.Count, errors.New("expression value must be a positive integer"))
+		}
+		head := &dag.Head{
+			Kind:  "Head",
+			Count: int(limit),
+		}
+		return append(a.semSQLOp(op.Op, seq), head)
+	default:
+		panic(fmt.Sprintf("semSQLOp: unknown op: %T", op))
```
nwt, line 157: Nit: More detail via %#v in all these switch-default panics could be nice.
```suggestion
		panic(fmt.Sprintf("semSQLOp: unknown op: %#v", op))
```

nwt (approved):
:shipit:

PR #5398, opened 2024-10-30
vam: Add min, max, sum and avg agg functions

mattnibs:


nwt (approved):
You can move runtime/sam/op/groupby/ztests/not-present.yaml over to runtime/ztests/op/summarize and add `vector: true`.

PR #5401, opened 2024-10-31
refactor api between compiler and parser

mccanne:
This commit clears up the interface between the parser and compiler and separates the collating of errors to source files into a separate package that is shared by the parser and semantic pass.  This gets rid of the awkward error handling that required binding the source-file set back to the error list with a complex type assertion when handling what should just be a normal Go error.  The only time this binding trick now must happen is in the API when the client library gets a response form the server. (We could further clean things up here by pushing this model through the service API but we don't do this in this commit.)


mattnibs (approved):
cool

compiler/semantic/analyzer.go
```diff
@@ -9,30 +9,30 @@ import (
...
+func Analyze(ctx context.Context, ast *parser.AST, source *data.Source, head *lakeparse.Commitish) (dag.Seq, error) {
+	files := ast.Files()
+	a := newAnalyzer(ctx, files, source, head)
+	s := a.semSeq(ast.Copy())
 	s = a.checkOutputs(true, s)
-	if a.errors != nil {
-		return nil, a.errors
-	}
-	return s, nil
+	return s, files.Error() //XXX
```
nwt, line 24: What's the XXX for?

compiler/semantic/analyzer.go
```diff
@@ -9,30 +9,30 @@ import (
...
 // Analyze performs a semantic analysis of the AST, translating it from AST
 // to DAG form, resolving syntax ambiguities, and performing constant propagation.
 // After semantic analysis, the DAG is ready for either optimization or compilation.
-func Analyze(ctx context.Context, seq ast.Seq, source *data.Source, head *lakeparse.Commitish) (dag.Seq, error) {
-	a := newAnalyzer(ctx, source, head)
-	s := a.semSeq(seq)
+func Analyze(ctx context.Context, ast *parser.AST, source *data.Source, head *lakeparse.Commitish) (dag.Seq, error) {
+	files := ast.Files()
+	a := newAnalyzer(ctx, files, source, head)
+	s := a.semSeq(ast.Copy())
```
nwt, line 22: We should get rid of this copy since semSeq doesn't modify its parameter.

compiler/srcfiles/file.go
```diff
@@ -0,0 +1,75 @@
...
+	return string(b)
+}
+
+func searchLine(lines []int, offset int) int {
+	return sort.Search(len(lines), func(i int) bool { return lines[i] > offset }) - 1
+
+}
+
+type Position struct {
+	Pos    int `json:"pos"`    // Offset relative to SourceSet.
```
nwt, line 69: Fix this comment.

nwt (approved):
I think this look good. My comments just nit-level stuff.

compiler/semantic/analyzer.go
```diff
@@ -9,30 +9,30 @@ import (
...
 
 // AnalyzeAddSource is the same as Analyze but it adds a default source if the
 // DAG does not have one.
-func AnalyzeAddSource(ctx context.Context, seq ast.Seq, source *data.Source, head *lakeparse.Commitish) (dag.Seq, error) {
-	a := newAnalyzer(ctx, source, head)
-	s := a.semSeq(seq)
+func AnalyzeAddSource(ctx context.Context, ast *parser.AST, source *data.Source, head *lakeparse.Commitish) (dag.Seq, error) {
+	files := ast.Files()
+	a := newAnalyzer(ctx, files, source, head)
+	s := a.semSeq(ast.Copy())
```
nwt, line 32: I think this copy can go away, too.

compiler/semantic/analyzer.go
mccanne: forgot to remove

PR #5404, opened 2024-10-31
make aggregate functions and expression functions case insensitive

mccanne:
This commit changes the semantic pass (and PEG parser for a couple cases) to accept mixed-case names for predefined functions and aggregate functions.  User-defined functions remain case sensitive. The invariant is that the transformations are all handled in the semantic pass and the DAG presumes everything relevant here has been converted to lower case.

### DIFF
--- a/compiler/parser/parser.go
+++ b/compiler/parser/parser.go
@@ -4384,7 +4384,7 @@ var g = &grammar{
 					},
 				},
 			},
-			leader:        false,
+			leader:        true,
 			leftRecursive: true,
 		},
 		{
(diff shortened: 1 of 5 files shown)


nwt (approved):

philrz:
Verified in super commit 2965a1d.

Before:

```
$ super -version
Version: v1.18.0-86-g9ef560ec

$ super -c 'SELECT COUNT() FROM "customers.bsup"'
super: no such function at line 1, column 8:
SELECT COUNT() FROM "customers.bsup"
       ~~~~~~~
cannot infer field from expression at line 1, column 8:
SELECT COUNT() FROM "customers.bsup"
       ~~~~~~~
illegal left-hand side of assignment at line 1, column 8:
SELECT COUNT() FROM "customers.bsup"
       ~~~~~~~
AS clause not a field at line 1, column 1:
SELECT COUNT() FROM "customers.bsup"
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Now:

```
$ super -version
Version: v1.18.0-87-g2965a1d0

$ super -c 'SELECT COUNT() FROM "customers.bsup"'
{count:1(uint64)}
{count:2(uint64)}
...
```

Thanks @mccanne!

PR #5413, opened 2024-11-01
add support for "*" in sql select

mccanne:
This commit adds super for `"*"` in select clauses as an empty ast.Assignment.  There isn't yet support for `id.*` style selections. We also reworked the selection naming convention to use the last name in a dotted path, which is what SQL does.

The SQL semantic logic was also reorged and renamed a bit to better position it for the comming work on scoping and binding.


compiler/semantic/sql.go
```diff
@@ -468,9 +534,16 @@ func (a *analyzer) deriveAs(as ast.Assignment) (field.Path, error) {
...
-		return f.Path
+func (a *analyzer) semGroupByKey(in ast.Expr) (*dag.This, bool) {
+	e := a.semExpr(in)
+	this, ok := e.(*dag.This)
+	if !ok {
+		a.error(in, errors.New("GROUP BY expressions are not yet supported; try expression in the selection with an AS"))
+		return nil, false
+	}
+	if len(this.Path) == 0 {
+		a.error(in, errors.New("cannot use 'this' as group-by expression"))
```
nwt, line 545: Nit: Matching styling in other error strings.
```suggestion
		a.error(in, errors.New("cannot use 'this' as GROUP BY expression"))
```

nwt (approved):

philrz:
Verified in super commit 9de86ad.

Before:

```
$ super -version
Version: v1.18.0-95-gac8b7e0f

$ super -c 'SELECT * FROM "customers.bsup" LIMIT 1'
super: parse error at line 1, column 15:
SELECT * FROM "customers.bsup" LIMIT 1
          === ^ ===
```

Now:

```
$ super -version
Version: v1.18.0-99-g9de86ad2

$ super -c 'SELECT * FROM "customers.bsup" LIMIT 1'
{CustomerID:1(int32),CustomerName:"Alfreds Futterkiste",ContactName:"Maria Anders",Address:"Obere Str. 57",City:"Berlin",PostalCode:"12209",Country:"Germany"}
```

Thanks @mccanne!

PR #5415, opened 2024-11-01
rework how command-line file args are wired into DAG

mccanne:
This commit moves the logic that determines if there should be a null source from queryflags to the compiler.  This decision really belongs in the semantic pass (as opposed to just after it) but that change will wait until we evolve the interface to file storage to handle broader reflection.

### DIFF
--- a/cli/queryflags/flags.go
+++ b/cli/queryflags/flags.go
@@ -1,18 +1,10 @@
 package queryflags
 
 import (
-	"context"
-	"errors"
 	"flag"
 	"fmt"
-	"net/url"
 	"os"
-	"slices"
 
-	"github.com/brimdata/super/compiler/data"
-	"github.com/brimdata/super/compiler/parser"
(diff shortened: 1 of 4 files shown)


cli/queryflags/flags.go
```diff
@@ -28,44 +20,6 @@ func (f *Flags) SetFlags(fs *flag.FlagSet) {
...
-		s, err := semantic.Analyze(context.Background(), ast, data.NewSource(storage.NewLocalEngine(), nil), nil)
-		if err != nil {
-			return nil, nil, false, err
-		}
-		//XXX we should simplify this logic, e.g., by inserting a null source
-		// if no source is given (this is how sql "select count(*)" works with no from)
-		if semantic.HasSource(s) {
-			return nil, ast, false, nil
-		}
-		if semantic.StartsWithYield(s) {
```
nwt, line 49: Nit: This is the only use of this function so you can remove it.

PR #5422, opened 2024-11-02
allow select operator to receive input from parent

mccanne:
This commit allows the SQL select operator to receive its input from its parent in a pipeline when no from clause is present.  This means that bare select queries work on constant values now since the default parent will be a single null value.

We also brought the column naming semantics into closer alignment with SQL where selected columns also have a string name and not a dotted field path.

The SQL semantic logic has been simplified and expand further as we make progress toward our binder-in-the-data approach to SQL scoping.

### DIFF
--- a/compiler/semantic/sql.go
+++ b/compiler/semantic/sql.go
@@ -48,7 +48,7 @@ func (a *analyzer) semSelect(sel *ast.Select, seq dag.Seq) dag.Seq {
 		}
 	} else if sel.Selection.Args != nil {
 		if sel.Having != nil {
-			a.error(sel, errors.New("HAVING clause used without GROUP BY"))
+			a.error(sel.Having, errors.New("HAVING clause used without GROUP BY"))
 			return append(seq, badOp())
 		}
 		seq = a.convertProjection(sel.Selection.Loc, proj, seq)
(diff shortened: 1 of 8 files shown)


compiler/semantic/sql.go
```diff
@@ -47,33 +39,19 @@ func (a *analyzer) semSelect(sel *ast.Select, seq dag.Seq) dag.Seq {
...
-		// For SELECT *, cutter is nil.
-		selector, err := projectSelect(proj)
-		if err != nil {
-			a.error(sel, err)
-			seq = append(seq, badOp())
-		} else {
-			seq = append(seq, selector)
+	} else if sel.Selection.Args != nil {
+		if sel.Having != nil {
+			a.error(sel, errors.New("HAVING clause used without GROUP BY"))
```
mattnibs, line 51: Shouldn't the error be on the having clause?
```suggestion
			a.error(sel.Having, errors.New("HAVING clause used without GROUP BY"))
```

mattnibs (approved):

compiler/semantic/sql.go
mccanne: good catch!

PR #5426, opened 2024-11-03
update help text for top-level super command

mccanne:


cmd/super/root/command.go
```diff
@@ -24,35 +24,39 @@ import (
...
 providing search, analytics, and extensive transormations using
-the Zed query language.
+the SuperSQL query language.
 A query typically applies Boolean logic or keyword search to filter
 the input and then transforms or analyzes the filtered stream.
 Output is written to one or more files or to standard output.
 
-A Zed query is comprised of one or more operators interconnected
+A query is comprised of one or more operators interconnected
 into a pipeline using the Unix pipe character "|".
```
philrz, line 39:
```suggestion
into a pipeline using the Unix pipe character "|>".
```

Want to aspirationally make this change now on the assumption #5427 will go through?

BTW, I briefly paused on what to call `|>` since it's not a single character, but I see the Google paper still calls it " the pipe character '|>'" so it seems like we can just go with that (though I also think the Unix comparison is still totally worth keeping).

cmd/super/root/command.go
```diff
@@ -24,35 +24,39 @@ import (
...
 See https://github.com/brimdata/super/tree/main/docs/language
```
philrz, line 40:
```suggestion
See https://zed.brimdata.io/docs/language
```

cmd/super/root/command.go
```diff
@@ -24,35 +24,39 @@ import (
...
-for details.
+for details.  The "select" and "from" operators provide backward
+compatible with SQL. In fact, you can use SQL exclusively and
```
philrz, line 42:
```suggestion
compatibility with SQL. In fact, you can use SQL exclusively and 
```

cmd/super/root/command.go
```diff
@@ -24,35 +24,39 @@ import (
...
 into a pipeline using the Unix pipe character "|".
 See https://github.com/brimdata/super/tree/main/docs/language
-for details.
+for details.  The "select" and "from" operators provide backward
+compatible with SQL. In fact, you can use SQL exclusively and 
+avoid pipeline operators altogether if you prefer.
 
-Supported input formats include CSV, JSON, NDJSON, Parquet,
+Supported file formats include Arrow, CSV, JSON, Parquet,
 Super JSON, Super Binary, Super Columnar,  and Zeek TSV.
```
philrz, line 46:
```suggestion
Super JSON, Super Binary, Super Columnar, and Zeek TSV. 
```

cmd/super/root/command.go
```diff
@@ -24,35 +24,39 @@ import (
...
-be file system paths; "-" for standard input; or HTTP, HTTPS, or S3 URLs.
+Input files may be file system paths;
+"-" for standard input; or HTTP, HTTPS, or S3 URLs.
 For most types of data, the input format is automatically detected.
 If multiple files are specified, each file format is determined independently
 so you can mix and match input types.  If multiple files are concatenated
 into a stream and presented as standard input, the files must all be of the
 same type as the beginning of stream will determine the format.
 
+If no input file is specified, the default of a single null value will be
```
philrz, line 56:
```suggestion
If no input file is specified, the default of a single null input value will be
```

cmd/super/root/command.go
```diff
@@ -24,35 +24,39 @@ import (
...
+fed to the query.  This is analogous to SQL's default input of a single
```
philrz, line 57:
```suggestion
fed to the query.  This is analogous to SQL's default input of a single
```

cmd/super/root/command.go
```diff
@@ -62,33 +66,25 @@ flag to indicate a destination directory for separate output files for each
...
-of the input.  If the first argument is both a valid Zed query
-and an existing file, then the file overrides.
-
-The Zed query text may include source files using -I, which is particularly
+The query text may include source files using -I, which is particularly
 convenient when a large, complex query spans multiple lines.  In this case,
 these source files are concatenated together along with the command-line query text
-in the order appearing on the command line.
+in the order appearing on the command line.  Any error messages are properly
+collated to the included file in which it occurred.
```
philrz, line 79:
```suggestion
collated to the included file in which they occurred.
```

cmd/super/root/command.go
```diff
@@ -24,35 +24,39 @@ import (
...
 
-"zq" must be run with at least one input.  Input files can
-be file system paths; "-" for standard input; or HTTP, HTTPS, or S3 URLs.
+Input files may be file system paths;
+"-" for standard input; or HTTP, HTTPS, or S3 URLs.
 For most types of data, the input format is automatically detected.
 If multiple files are specified, each file format is determined independently
 so you can mix and match input types.  If multiple files are concatenated
 into a stream and presented as standard input, the files must all be of the
 same type as the beginning of stream will determine the format.
```
philrz, line 54:
```suggestion
same type as the beginning of the stream will determine the format.
```

philrz (approved):

cmd/super/root/command.go
mccanne: using the pipe symbol "|>"

PR #5428, opened 2024-11-03
add tuple expressions

mccanne:
This commit adds a tuple expression to the query language.  Since the data model and runtime do not support native tuples, these expressions are mapped onto records with columns c0, c1, ...  Datafusion does this and other SQLs map tuples to structs with unnamed fields.

This will be needed by a subsequent PR implementing the VALUES clause.


compiler/semantic/expr.go
```diff
@@ -244,6 +244,20 @@ func (a *analyzer) semExpr(e ast.Expr) dag.Expr {
 			Kind:    "MapExpr",
 			Entries: entries,
 		}
+	case *ast.TupleExpr:
+		elems := make([]dag.RecordElem, 0, len(e.Elems))
+		for colno, elem := range e.Elems {
+			e := a.semExpr(elem)
+			elems = append(elems, &dag.Field{
+				Kind:  "Field",
+				Name:  fmt.Sprintf("c%d", colno),
```
mattnibs, line 253: Isn't there a concern here that this could collide with pre-existing column names?

nwt (approved):

compiler/semantic/expr.go
mccanne: The tuple type doesn't have column names so this is assigning `c0,c1,...` as the column names of the value that comprises the tuple.  No possibility of field name conflict.

PR #5430, opened 2024-11-03
add bitwise operators

mccanne:
This commit adds bitwise operators to the PEG grammar, but doesn't otherwise support them.  If you use them, you'll get an error.

We've been living without bitwise OR in the grammar for a long time and this time and this change illustates the conflict between "|" vs "|>" as the pipeline separator.

### DIFF
--- a/compiler/parser/parser.peg
+++ b/compiler/parser/parser.peg
@@ -983,7 +983,8 @@ BitShiftExpr
 
 BitShiftOperator = ("<<" / ">>") { return string(c.text), nil }
 
-AdditiveExpr = first:MultiplicativeExpr
+AdditiveExpr
+  = first:MultiplicativeExpr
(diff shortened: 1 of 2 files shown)


compiler/parser/parser.peg
```diff
@@ -963,8 +963,27 @@ ComparisonExpr
...
+
+BitShiftExpr
+  = first:AdditiveExpr
+    rest:(__ op:BitShiftOperator __ expr:AdditiveExpr { return []any{op, expr}, nil })* {
+        return makeBinaryExprChain(first, rest, c), nil
+    }    
+
+BitShiftOperator = ("<<" / ">>") { return string(c.text), nil }
+
+AdditiveExpr = first:MultiplicativeExpr
```
nwt, line 986: Nit:
```suggestion
AdditiveExpr
  = first:MultiplicativeExpr
```

nwt (approved):

PR #5437, opened 2024-11-06
allow from operators to have parents for robot scans

mccanne:
This commit allows from operators to pull data from its parent and for each incoming value compute a target to open and read. This is repeated successively until the parent fails.  When failures occur, structured errors are propgated downstream.

There is a disabled test mimicing the example in the top-level README.  It will be enabled in a subsequent upon adding support for "join using".

### DIFF
--- a/compiler/semantic/op.go
+++ b/compiler/semantic/op.go
@@ -35,20 +35,16 @@ func (a *analyzer) semSeq(seq ast.Seq) dag.Seq {
 	return converted
 }
 
-func (a *analyzer) semFrom(from *ast.From, out dag.Seq) dag.Seq {
+func (a *analyzer) semFrom(from *ast.From, seq dag.Seq) dag.Seq {
 	if len(from.Elems) > 1 {
 		a.error(from, errors.New("cross join implied by multiple elements in from clause is not yet supported"))
 		return dag.Seq{badOp()}
 	}
-	if len(out) > 0 {
-		a.error(from, errors.New("from operator with parent input is not yet supported"))
-		return dag.Seq{badOp()}
-	}
-	return a.semFromElem(from.Elems[0])
+	return a.semFromElem(from.Elems[0], seq)
 }
 
-func (a *analyzer) semFromElem(elem *ast.FromElem) dag.Seq {
-	seq := a.semFromEntity(elem.Entity, elem.Args)
+func (a *analyzer) semFromElem(elem *ast.FromElem, seq dag.Seq) dag.Seq {
+	seq = a.semFromEntity(elem.Entity, elem.Args, seq)
 	if elem.Ordinality != nil {
 		a.error(elem.Ordinality, errors.New("WITH ORDINALITY clause is not yet supported"))
 		return dag.Seq{badOp()}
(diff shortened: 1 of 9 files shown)


compiler/semantic/sql.go
```diff
@@ -18,11 +18,20 @@ import (
...
-		if len(seq) > 0 {
-			a.error(sel, errors.New("SELECT cannot have both an embedded FROM claue and input from parents"))
-			return append(seq, badOp())
+		off := len(seq)
+		hasParent := off > 0
+		seq = a.semFrom(sel.From, seq)
+		// If we have parents with both a from and select, report an error
+		// but only if it's not a RobotScan where the parent needs to feed
+		// the from.
+		if off < len(seq) {
```
nwt, line 27: `off < len(seq)` will never be true here because `off := len(seq)`.

runtime/sam/op/robot/robot.go
```diff
@@ -0,0 +1,193 @@
...
+		var headers http.Header
+		f, err := o.env.OpenHTTP(o.rctx.Context, o.rctx.Zctx, u.String(), o.format, method, headers, body, nil)
+		if err != nil {
+			return nil, err
+		}
+		return f, err
+	}
+	if o.env.IsLake() {
+		return nil, fmt.Errorf("%s: cannot open in a data lake environment", path)
+	}
```
nwt, line 187: This check could happen in semantic.analyzer.semFromExpr.

runtime/sam/op/robot/robot.go
```diff
@@ -0,0 +1,193 @@
...
+			o.batch.Unref()
+		}
+		o.batch = nil
+		src := o.src
+		o.src = nil
+		if closer, ok := src.(io.Closer); ok {
+			if err := closer.Close(); err != nil {
+				return nil, err
+			}
+		}
```
nwt, line 55: Call Pull(true) when discarding a zbuf.Puller. (Neither runtime.Environment.Open nor OpenHTTP returns a value implementing io.Closer.)

```suggestion
		if o.src != nil {
			if _, err := o.src.Pull(true); err != nil {
				return nil, err
			}
		}
```

runtime/sam/op/robot/robot.go
```diff
@@ -0,0 +1,193 @@
...
+	if puller != nil {
+		return puller, nil
+	}
+	if len(o.targets) > 0 {
+		src, err := o.openNext()
+		o.src = src
+		return src, err
+	}
+	src, err := o.nextPuller()
+	o.src = src
```
nwt, line 89: As mentioned, call Pull(true) when discarding a zbuf.Puller. In this territory, I think that means anytime a non-nil value in o.src is overwritten.

compiler/semantic/sql.go
mccanne: semFrom is called between off getting set and getting checked.  The idea here was to protect the seq[off] check if semFrom didn't lengthen the chain.  I will rework it so it's more clear to return of the chain isn't lengthened.

runtime/sam/op/robot/robot.go
mccanne: It's gonna go away when we add support for lake's in robotscan.

compiler/semantic/sql.go
nwt: Completely missed that. 🤦

runtime/sam/op/robot/robot.go
```diff
@@ -0,0 +1,193 @@
...
+		return f, err
+	}
+	if o.env.IsLake() {
+		return nil, fmt.Errorf("%s: cannot open in a data lake environment", path)
+	}
+	f, err := o.env.Open(o.rctx.Context, o.rctx.Zctx, path, o.format, o.filter, o.demand)
+	if err != nil {
+		return nil, err
+	}
+	return f, err
```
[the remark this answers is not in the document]
mccanne: Oh yeah,  I thought I had the interface nil pointer problem but it turned out it was something else.

compiler/semantic/op.go
```diff
@@ -78,42 +74,79 @@ func wrapAlias(alias string, seq dag.Seq) dag.Seq {
 	})
 }
 
-func (a *analyzer) semFromEntity(entity ast.FromEntity, args ast.FromArgs) dag.Seq {
+func (a *analyzer) semFromEntity(entity ast.FromEntity, args ast.FromArgs, seq dag.Seq) dag.Seq {
 	switch entity := entity.(type) {
 	case *ast.Glob:
+		if a.hasFromParent(entity, seq) {
+			return seq
+		}
```
nwt, line 82: Nit: Returning seq when hasFromParent failes doesn't really make sense. Clearer to return dag.Seq{badOp()}.

mccanne: Well that pattern that is used throughout it to march along with the bad seq's.. I thought this was useful for debugging but I will change this.

compiler/semantic/op.go
```diff
@@ -78,42 +74,79 @@ func wrapAlias(alias string, seq dag.Seq) dag.Seq {
...
+		return a.semFromConstVal(val, entity, args)
+	}
+	// This is an expression so set up a robot scanner that pulls values from
+	// parent to decide what to scan.
+	e := a.semExpr(entity.Expr)
+	return append(seq, &dag.RobotScan{
+		Kind:   "RobotScan",
+		Expr:   e,
```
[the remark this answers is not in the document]
mccanne: Doh!

compiler/semantic/op.go
nwt: I guess what I'm getting at is that seq is here for use in analysis but isn't meant to be returned. Even the successful returns from this method don't incorporate seq.

nwt (approved):

PR #5439, opened 2024-11-06
Adapt perf-compare script for zed#5436 changes

philrz:
The perf-compare script has always had an explicit `*` query in it so it'll show up in the results table, and that needs to change to `? *` now.

### DIFF
--- a/scripts/perf-compare.sh
+++ b/scripts/perf-compare.sh
@@ -46,7 +46,7 @@ declare -a DESCRIPTIONS=(
 )
 
 declare -a SPQS=(
-    '*'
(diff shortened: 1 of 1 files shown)


scripts/perf-compare.sh
```diff
@@ -46,7 +46,7 @@ declare -a DESCRIPTIONS=(
 )
 
 declare -a SPQS=(
-    '*'
+    '? *'
```
nwt, line 49:
```suggestion
    'pass'
```

nwt (approved):

PR #5446, opened 2024-11-07
Fix bug affecting vector min(), max(), sum(), and avg() functions

nwt:
They return incorrect results because the minOf, maxOf, and sumOf functions in runtime/vam/expr/agg update their state with vector slot numbers.  Fix them so they use the values in those slots instead.


mattnibs (approved):
Whoops!

PR #5449, opened 2024-11-07
Add projection pushdown to the Parquet reader

nwt:
This includes the addition of compiler/dag.FileScan.Fields to communicate projection pushdown configuration to
compiler/kernel.Builder.

Depends on #5448.

### DIFF
--- a/zio/parquetio/reader.go
+++ b/zio/parquetio/reader.go
@@ -10,10 +10,11 @@ import (
 	"github.com/apache/arrow-go/v18/parquet/file"
 	"github.com/apache/arrow-go/v18/parquet/pqarrow"
 	"github.com/brimdata/super"
+	"github.com/brimdata/super/pkg/field"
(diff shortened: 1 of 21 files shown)


zio/parquetio/reader.go
```diff
@@ -31,7 +32,8 @@ func NewReader(zctx *super.Context, r io.Reader) (*arrowio.Reader, error) {
 		pr.Close()
 		return nil, err
 	}
-	rr, err := fr.GetRecordReader(context.TODO(), nil, nil)
+	cols := columnIndexes(fr.Manifest, fields)
```
mattnibs, line 35: Shouldn't this check if fields is `nil` and make sure we get all columns?

nwt: No, if `fields==nil` then `columnIndexes` will return a nil slice of column indexes and that will tell `GetRecordReader` to read all columns.

mattnibs (approved):

PR #5450, opened 2024-11-07
Add vector Parquet reader

nwt:
It supports projection pushdown and concurrent access (via NewConcurrentPuller).

Note that the new runtime/exec.Environment.VectorOpen method does not yet arrange to close the storage.Reader that it creates.

Depends on #5449.

### DIFF
--- a/zio/parquetio/vectorreader.go
+++ b/zio/parquetio/vectorreader.go
@@ -1,0 +1,336 @@
+package parquetio
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"sync/atomic"
+	"time"
+	"unsafe"
+
+	"github.com/apache/arrow-go/v18/arrow"
+	"github.com/apache/arrow-go/v18/arrow/array"
+	"github.com/apache/arrow-go/v18/arrow/memory"
+	"github.com/apache/arrow-go/v18/parquet"
+	"github.com/apache/arrow-go/v18/parquet/file"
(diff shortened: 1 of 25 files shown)


zio/parquetio/vectorreader.go
```diff
@@ -0,0 +1,345 @@
...
+	length := uint32(a.Len())
+	// Order here follows that of the arrow.Type constants.
+	switch dt.ID() {
+	case arrow.NULL:
+		bits := make([]uint64, (length+7)/8)
+		for i := range bits {
+			bits[i] = ^uint64(0)
+		}
+		nulls := vector.NewBool(bits, length, nil)
+		return vector.NewConst(super.Null, length, nulls), nil
```
mattnibs, line 118: Why would a null const value have a nulls vector? Is there such thing as a null null?

mattnibs (approved):

zio/parquetio/vectorreader.go
nwt: To the vector runtime, a slot in a vector is null if and only if that vector has a nulls vector and its bit corresponding to that slot is set. A vector.Const containing a null zed.Value but having no nulls vector is generally going to look to the runtime like it contains zero values rather than nulls.

We could of course change the runtime to check for a null slot by first checking the nulls vector and then falling back to checking for a vector.Const containing a null zed.Value, but my take is that's more complicated than the current approach without offering any benefit.

This is fairly subtle so my explanation here might not be very convincing, in which case I'm happy to elaborate.

PR #5451, opened 2024-11-07
sam: SQL semantics for null arithmetic

mattnibs:


runtime/sam/expr/eval.go
```diff
@@ -407,6 +407,9 @@ func (a *Add) Eval(ectx Context, this super.Value) super.Value {
 	if errVal != nil {
 		return *errVal
 	}
+	if lhsVal.IsNull() || rhsVal.IsNull() {
+		return super.NewValue(typ, nil)
+	}
```
nwt, line 412: Nit: Instead of adding this to each operator implementation, you could do it in evalAndPromote.

nwt (approved):

runtime/sam/expr/eval.go
mattnibs: I don't think this'll work because it'll still coerce the values to 0 if we change lhs and rhs to null in coerceAndPromote.

nwt: I was thinking of this.

```go
func (n *numeric) evalAndPromote(ectx Context, this super.Value) (super.Value, super.Value, super.Type, *super.Value) {
	lhsVal, rhsVal, errVal := n.eval(ectx, this)
	if errVal != nil {
		return super.Null, super.Null, nil, errVal
	}
	id, err := coerce.Promote(lhsVal, rhsVal)
	if err != nil {
		return super.Null, super.Null, nil, n.zctx.NewError(err).Ptr()
	}
	typ, err := n.zctx.LookupType(id)
	if err != nil {
		return super.Null, super.Null, nil, n.zctx.NewError(err).Ptr()
	}
	if lhsVal.IsNull() || rhsVal.IsNull() {
		return super.Null, super.Null, nil, super.NewValue(typ, nil).Ptr()
	}
	return lhsVal, rhsVal, typ, nil
}

```

nwt (approved):
Approving even harder than last time.

PR #5455, opened 2024-11-08
vam: SQL semantics for nulls in arithmetic

mattnibs:


runtime/vam/expr/genarithfuncs.go
```diff
@@ -77,13 +77,13 @@ func genFunc(name, op, typ string, lhs, rhs vector.Form) string {
 		} else {
 			s += fmt.Sprintf("val := super.New%s64(lconst %s rconst)\n", typ, op)
 		}
-		s += "return vector.NewConst(val, lhs.Len(), nil)\n"
+		s += "return vector.NewConst(val, lhs.Len(), vector.Or(l.Nulls, r.Nulls))\n"
```
nwt, line 80: Nit: Instead of repeating this, up above change

```go
```

to

```go
	s += "nulls := vector.Or(l.Nulls, r.Nulls)\n"
```

and then just do this here (and in the two spots with `vector.Or` below):

```go
		s += "return vector.NewConst(val, lhs.Len(), nulls)\n"
```

nwt (approved):

PR #5456, opened 2024-11-08
vam: divide by zero error

mattnibs:


runtime/vam/expr/arith.go
```diff
@@ -48,5 +51,90 @@ func (a *Arith) eval(vecs ...vector.Any) vector.Any {
 	if !ok {
 		return vector.NewStringError(a.zctx, coerce.ErrIncompatibleTypes.Error(), lhs.Len())
 	}
-	return f(lhs, rhs)
+	out, ok := f(lhs, rhs)
+	if !ok {
+		return a.evalDivideByZero(kind, lhs, rhs)
+	}
+	return out
```
nwt, line 58: Keep genarithfuncs.go a little simpler by letting the hardware do the check?
```suggestion
	if a.opCode == vector.ArithDiv || a.opCode == vector.ArithMod {
		defer func() {
			if v := recover(); v != nil {
				if err, ok := v.(runtime.Error); ok && err.Error() == "runtime error: integer divide by zero" {
					out = a.evalDivideByZero(kind, lhs, rhs)
				} else {
					panic(v)
				}
			}
		}()
	}
	return f(lhs, rhs)	
```

runtime/vam/expr/arith.go
```diff
@@ -48,5 +51,90 @@ func (a *Arith) eval(vecs ...vector.Any) vector.Any {
...
+				floats = append(floats, 0)
+				continue
+			}
+			if r == 0 {
+				errs = append(errs, i)
+				continue
+			}
+			floats = append(floats, l/r)
+		}
+		out = vector.NewFloat(super.TypeFloat64, floats, nulls)
```
nwt, line 132: Floating point division by zero isn't usually an error. (The IEEE 754 spec says it results in NaN, +Inf, or -Inf depending on the operands.)
```suggestion
```

runtime/vam/expr/arith.go
mattnibs: Yeah I was thinking about this sounds good

runtime/vam/expr/arith.go
```diff
@@ -48,5 +52,74 @@ func (a *Arith) eval(vecs ...vector.Any) vector.Any {
...
+				if err, ok := v.(runtime.Error); ok && err.Error() == "runtime error: integer divide by zero" {
+					out = a.evalDivideByZero(kind, lhs, rhs)
+				} else {
+					panic(v)
+				}
+			}
+		}()
+	}
+	out = f(lhs, rhs)
+	return
```
nwt, line 67: Nit: Naked returns are the worst.
```suggestion
	return f(lhs, rhs)
```

runtime/vam/expr/genarithfuncs.go
```diff
@@ -49,7 +49,7 @@ func main() {
 		}
 	}
 
-	fmt.Fprintln(&buf, "var arithFuncs = map[int]func(vector.Any, vector.Any) vector.Any{")
+	fmt.Fprintln(&buf, "var arithFuncs = map[int]func(vector.Any, vector.Any) vector.Any {")
```
nwt, line 52: Nit: Don't change this file.

nwt (approved):

PR #5458, opened 2024-11-09
vam: Add bucket() function

mattnibs:


runtime/sam/expr/function/time.go
```diff
@@ -21,30 +21,36 @@ type Bucket struct {
...
 func (b *Bucket) Call(_ super.Allocator, args []super.Value) super.Value {
+	args = underAll(args)
 	tsArg := args[0]
 	binArg := args[1]
 	if tsArg.IsNull() || binArg.IsNull() {
+		if tsArg.Type().ID() == super.IDDuration {
+			return super.NullDuration
+		}
 		return super.NullTime
 	}
```
nwt, line 32: Nit: Check argument types before nulls. Doing it this way means `yield bucket(string(null), ip(null))` evaluates to `null(time)`, which is weird.

runtime/sam/expr/function/time.go
```diff
@@ -21,30 +21,36 @@ type Bucket struct {
...
-		bin = nano.Duration(binArg.Int())
-	} else {
-		d, ok := coerce.ToInt(binArg, super.TypeDuration)
-		if !ok {
-			return b.zctx.WrapError(b.name+": second argument is not a duration or number", binArg)
-		}
-		bin = nano.Duration(d) * nano.Second
+	i, ok := coerce.ToInt(binArg, super.TypeInt64)
+	if !ok {
+		return b.zctx.WrapError(b.name+": second argument is not a duration or number", binArg)
```
nwt, line 35: Nit: This error string isn't accurate since coerce.ToInt can succeed when the statement isn't true and fail when it is true.

runtime/sam/expr/function/time.go
```diff
@@ -21,30 +21,36 @@ type Bucket struct {
...
+		if bin != 0 {
+			dur = dur.Trunc(bin)
+		}
+		return super.NewDuration(dur)
 	}
-	v, ok := coerce.ToInt(tsArg, super.TypeInt64)
+	v, ok := coerce.ToInt(tsArg, super.TypeTime)
 	if !ok {
-		return b.zctx.WrapError(b.name+": first argument is not a time", tsArg)
+		return b.zctx.WrapError(b.name+": first argument is not a time or duration", tsArg)
```
nwt, line 47: Nit: This error string isn't accurate either.

mattnibs: Alright I'm just going to make it require a duration for the second argument and time or duration for the first.

nwt (approved):

PR #5462, opened 2024-11-11
vam: Array Expressions

mattnibs:
Add vector array expressions. When building an array in vector land you are often creating a non-union inner type that includes values from multiple vectors. This presents a problem because you cannot merge the vectors using a dynamic because dynamics are supposed to bubble up on be on top. As such I've added vector.Builder that is used to rebuild the multiple vectors to a single vector when this is encountered.


philrz:
I happened to be revisiting the mgbench queries and since I get `vector expression type *dag.ArrayExpr: not supported` on the tip of `main` when trying to run bench1/q1 with vectors, I remembered this branch just went up. I now get:

```
$ super -version
Version: v1.18.0-139-ge69d3fd6

$ super dev vector query "where
     log_time >= 2017-01-11T00:00:00Z
     and machine_name in ['anansi','aragog','urd']" bench1.csup

invalid binary operator in
```

Not sure if that's expected to work here or would come later, but figured I'd report.

mattnibs:
@philrz no support for `in` operator in this pr

runtime/vam/expr/arrayexpr.go
```diff
@@ -0,0 +1,112 @@
...
+		innerVals = vecs[0]
+	} else {
+		var all []super.Type
+		for _, vec := range vecs {
+			all = append(all, vec.Type())
+		}
+		types := super.UniqueTypes(all)
+		if len(types) == 1 {
+			typ = types[0]
+			innerVals = vector.NewDynamic(tags, vecs)
```
mattnibs, line 84: Actually I think this is a problem since we'll have a dynamic not on top... I'm not sure what to do about this...

mattnibs:
@nwt can we handle records in a separate pr?

mattnibs:
Given there's some controversy over this I'm just going to get rid of spread on primitives for now

mattnibs:
@nwt ready for another review I added back the behavior where non-spreadable elements are ignored.

compiler/kernel/vexpr.go
```diff
@@ -218,3 +218,32 @@ func (b *Builder) compileVamRecordExpr(e *dag.RecordExpr) (vamexpr.Evaluator, er
...
+
+func (b *Builder) compileVamArrayExpr(e *dag.ArrayExpr) (vamexpr.Evaluator, error) {
+	elems, err := b.compileVamVectorElems(e.Elems)
+	if err != nil {
+		return nil, err
+	}
+	return vamexpr.NewArrayExpr(b.zctx(), elems), nil
+}
+
+func (b *Builder) compileVamVectorElems(elems []dag.VectorElem) ([]vamexpr.ListElem, error) {
```
nwt, line 230: Nit: Maybe call this compileVamListElems to avoid overloading vector?

runtime/vam/expr/arrayexpr.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+	offsets := []uint32{0}
+	var tags []uint32
+	for i := range n {
+		var size uint32
+		for tag, spreadOff := range spreadOffs {
+			if len(spreadOff) == 0 {
+				tags = append(tags, uint32(tag))
+				size++
+				continue
```
nwt, line 68: Nit: Don't need continue since the other path is in an else block.
```suggestion
```

nwt (approved):

PR #5474, opened 2024-11-12
Reduce allocations in runtime/vam/op/summarize

nwt:
* Add a sync.Pool for the index slices allocated in superTable.update.

* Change map values to pointers to limit assignments to string-keyed maps in superTable.update and various countByString methods.  (In these assignments, the key escapes to the heap even if already present in the map.)

### DIFF
--- a/runtime/vam/op/summarize/agg.go
+++ b/runtime/vam/op/summarize/agg.go
@@ -2,6 +2,7 @@ package summarize
 
 import (
 	"fmt"
+	"sync"
 
 	"github.com/brimdata/super"
 	"github.com/brimdata/super/runtime/vam/expr"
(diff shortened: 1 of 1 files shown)


mccanne:
This slows down `count() by repo.name` on the gha dataset by 10%.  Is the intention to trade off speed for memory?

nwt:
Rats. The intention was to improve performance, which it did for `count() by event`. I'll look into what's happening with `count() by repo.name`.

nwt:
Looks like `count() by repo.name` goes slower because it has high cardinality with low counts, so there aren't enough updates to existing map entries over which to amortize the extra cost of allocating the objects to which those entries point.

PR #5477, opened 2024-11-13
bring up JOIN USING and make README query work

mccanne:
This commit gets the using condition for joins working and now allows the query that is now in the top-level README to work.  We unified some of the join logic between SQL and SPQ and relaxed some of the checking of group-by keys with selection, which had been too aggressive. We also added a test that emulates the README query with smaller data in local files rather than APIs.

### DIFF
--- a/compiler/semantic/sql.go
+++ b/compiler/semantic/sql.go
@@ -178,7 +178,7 @@ func (a *analyzer) semSQLJoin(join *ast.SQLJoin, seq dag.Seq) dag.Seq {
 		a.error(join.Right, errors.New("SQL joins currently require a table alias on the right lef of the join"))
 		seq = append(seq, badOp())
 	}
-	leftKey, rightKey, err := a.semJoinCond(join.Cond)
+	leftKey, rightKey, err := a.semSQLJoinCond(join.Cond)
 	if err != nil {
 		a.error(join.Cond, errors.New("SQL joins currently limited to equijoin on fields"))
 		return append(seq, badOp())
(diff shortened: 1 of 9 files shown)


compiler/semantic/sql.go
```diff
@@ -318,21 +337,16 @@ func (a *analyzer) semGroupBy(exprs []ast.Expr, proj projection, seq dag.Seq) (d
...
-		if !path.In(paths) {
+	for k, col := range scalars {
+		path := col.scalar
+		if this, ok := col.scalar.(*dag.This); ok {
+			if field.Path(this.Path).In(paths) {
+				continue
+			}
+		}
+		if !(field.Path{col.name}).In(paths) {
 			a.error(exprs[k], fmt.Errorf("'%s': selected expression is missing from GROUP BY clause (and is not an aggregation)", path))
```
nwt, line 350: Nit:
```suggestion
	for k, col := range scalars {
		if this, ok := col.scalar.(*dag.This); ok {
			if paths.Has(this.Path) {
				continue
			}
		}
		if !paths.Has(col.name) {
			a.error(exprs[k], fmt.Errorf("%q: selected expression is missing from GROUP BY clause (and is not an aggregation)", col.scalar))
```

zfmt/ast.go
```diff
@@ -565,11 +565,14 @@ func (c *canon) op(p ast.Op) {
...
-			c.write("=")
-			c.expr(p.RightKey, "")
+		switch cond := p.Cond.(type) {
+		case *ast.JoinOnExpr:
+			c.write("on ")
+			c.expr(cond.Expr, "")
+		case *ast.JoinUsingExpr:
+			c.write("using (")
+			c.exprs(cond.Fields)
+			c.write(")")
```
nwt, line 575: Nit:
```suggestion
			c.write(")")
		default:
			panic(cond)
```

nwt (approved):

PR #5479, opened 2024-11-14
add SQL syntactic sugar to SPQ pipe operators

mccanne:
This commit adds some syntactic sugar so that UNNEST can be used in place of OVER, ORDER BY in place of SORT, and LIMIT in place of HEAD.  We also unified the PEG grammar for the sort operaetor with the existing rules for SQL ORDER BY, and we added syntactic sugar for IS NULL and IS NOT NULL.

### DIFF
--- a/compiler/parser/parser.peg
+++ b/compiler/parser/parser.peg
@@ -384,11 +384,13 @@ AssertOp
     }
 
 SortOp
-  = SORT &EOKW args:SortArgs exprs:(__ e:SortExprs { return e, nil })? {
+  = (SORT / ORDER _ BY) &EOKW args:SortArgs exprs:(__ e:OrderByList { return e, nil })? {
       o := &ast.Sort{
         Kind: "Sort",
-        Args: sliceOf[ast.SortExpr](exprs),
         Loc: loc(c),
+      }
+      if exprs != nil {
+        o.Args = exprs.([]ast.SortExpr)
       }
       argm := args.(map[string]any)
       if _, ok := argm["r"]; ok {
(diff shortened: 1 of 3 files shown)


compiler/parser/parser.peg
```diff
@@ -962,6 +954,10 @@ ComparisonExpr
         Loc: loc(c),
       }, nil
     }
+ 
+ IsNull
+  = IS _ NULL { return "==", nil }
+  / IS _ NOT _ NULL { return "!=", nil }
```
mattnibs, line 961: Should this be its own expression? Once we get to SQL semantics for null values `IS NULL` will behave differently than `foo == null`...

compiler/ztests/sql/assignees.yaml
```diff
@@ -0,0 +1,26 @@
+
+script: |
+  super -z -I query.sql
+
+inputs:
+  - name: query.sql
+    data: |
+      FROM in.jsup
+      | UNNEST [...assignees,assignee]
+      | WHERE this IS NOT NULL
```
mattnibs, line 10: This doesn't feel right since currently `FROM in.jsup | UNNEST [...assignees,assignee]` evaluates to:
```
{login:"joe"}
{login:"fred"}
error("missing")
{login:"sue"}
{login:"joe"}
{login:"sue"}
error("missing")
```
or is `error("missing")` equivalent to null for `this IS NOT NULL`?

nwt (approved):

compiler/ztests/sql/assignees.yaml
mccanne: This came from a SQL query where there is no MISSING or error("missing") and the behavior works with our current `this!=null` semantics and is also consistent with SQL++ semantics, so I think we're ok.

Issue #5482, opened 2024-11-15
yaml support


1nhann:
use `-i yaml` to recieve yaml data as input

use `-f yaml` to output data as yaml

not support auto detect yet

1nhann:
https://github.com/brimdata/super/pull/5483

PR #5483, opened 2024-11-15
add yaml support

1nhann:
support yaml

### DIFF
--- a/cli/inputflags/flags.go
+++ b/cli/inputflags/flags.go
@@ -27,7 +27,7 @@ func (f *Flags) Options() anyio.ReaderOpts {
 }
 
 func (f *Flags) SetFlags(fs *flag.FlagSet, validate bool) {
-	fs.StringVar(&f.Format, "i", "auto", "format of input data [auto,arrows,bsup,csup,csv,json,jsup,line,parquet,tsv,zeek,zjson]")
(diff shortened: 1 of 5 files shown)


philrz:
@1nhann: Thanks for the contribution! The core Dev team looked at the code and are going to decline for now because it's not quite production ready. A few gaps identified:

* What's here modifies the JSON I/O support, but in the project each reader/writer typically stands alone.
* It lacks the auto-detect read support as you pointed out in #5482, and we typically try to auto-detect input formats whenever possible for consistency.
* It needs some tests.

Ideally we'd have more contributor documentation we could point you to that provides guidance on how to approach these. Right now the small core Dev team is quite busy with the effort morphing the project from Zed to SuperDB (you can see lots of activity of this in the commit history if you're curious to watch the progress) so are short on time to write those materials right now or walk you through the equivalent via PR comments. We expect we'll have more time to work with the community in 2025 after the wider SuperDB launch.

I've expanded the text at the issue you opened at #5482 so other users interested in YAML support that find it via issue searches are aware of the interim workarounds like `yq` and can also find this PR and your branch if they want to make use of it before an official enhancement can be added.

I'll close this PR for now. Thanks for helping to improve Zed/SuperDB!

PR #5484, opened 2024-11-15
vam: Summarize use vector.Builder to materialize

mattnibs:
The commit changes slow path aggregations to use vector.Builder when materializing the aggregation table. The previous approaching of creating a dynamic of consts was causing the system to run out of memory when querying larger datasets.


philrz:
Note to self: This should address a problem I was seeing with high memory usage when running mgbench bench2/q5, so I should re-test that once this merges.

mccanne (approved):

PR #5486, opened 2024-11-17
add SQL LENGTH() function as alias for len()

mccanne:


mattnibs (approved):

runtime/ztests/expr/length.yaml
```diff
@@ -0,0 +1,9 @@
+zed: SELECT LENGTH('W3Schools.com');
```
mattnibs, line 1: Put this in `runtime/ztests/expr/function/length` where the test for `len` is

philrz:
Verified in super commit 5c1851b.

Before:

```
$ super -version
Version: v1.18.0-156-gb2c46ef7

$ super -c "SELECT LENGTH('W3Schools.com');"
SELECT LENGTH('W3Schools.com');
       ~~~~~~~~~~~~~~~~~~~~~~~
```

Now:

```
$ super -version
Version: v1.18.0-157-g5c1851be

$ super -c "SELECT LENGTH('W3Schools.com');"
{length:13}
```

Thanks @mccanne!

PR #5481, opened 2024-11-14
update docs on super command for SQL/OLAP audience

mccanne:

PR #5487, opened 2024-11-17
Minor edits to super command doc

philrz:
These are some minor changes that come to mind while I was reviewing #5481.

This only touches the first half of the doc and not the second half with the Performance section. I've been writing some scripts to make it easy to run those queries and assemble the results so we can revisit them frequently and also point users at a way to run them easily on an AWS instance as neutral turf. I'll put up a separate PR next with those changes.

### DIFF
--- a/docs/commands/super.md
+++ b/docs/commands/super.md
@@ -33,7 +33,7 @@ check out the [`super db`](super-db.md) set of commands.
 By invoking the `-c` option, a query expressed in the [SuperSQL language](../language/README.md)
 may be specified and applied to the input stream.
 
-Super's data model is based on super-structured data, meaning that all data
+The [super data model](../formats/zed.md) is based on [super-structured data](../formats/README.md#2-a-super-structured-pattern), meaning that all data
 is both strongly _and_ dynamically typed and need not conform to a homogeneous
 schema.  The type structure is self-describing so it's easy to daisy-chain
 queries and inspect data at any point in a complex query or data pipeline.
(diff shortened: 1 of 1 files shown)


docs/commands/super.md
```diff
@@ -52,27 +52,31 @@ do not haphazardly change when input data changes in subtle ways.
...
+super -c "FROM example.json | SELECT typeof(this)"
 ```
 is equivalent to
 ```
-super -c "select typeof(this)" example.json
+super -c "SELECT typeof(this)" example.json
+```
+and both are equivalent to the classic SQL
+```
+super -c "SELECT typeof(this) FROM example.json"
```
philrz, line 66: I'm still getting the feel for when and how much to reassure readers that standard SQL is still in there. I know we've talked about covering this topic in more detail when we get to overhauling the Language docs. Regardless, my instincts are to make some nods to classic SQL early in individual docs pages since we can't always be sure where users will first land and we can't be confident they'll always click on links when we hope they will. While we hope that users will quickly catch on to SuperSQL's pipe extensions so they can do things with super-structured data that classic SQL can't, I also want to make it as clear as possible to the new users that learning new elements are not a _prerequisite_ to getting started, hence my addition of this new text.

Also related to this is when to use ALL CAPS. Once again, it feels worthwhile to use all caps in contexts where we're showing pure SQL that could drop directly into some other classic relational system. Likewise if we're showing an example that learns mostly/exclusively on SuperSQL shortcuts it seems reasonable to opt for lowercase, even if the example shares a few keywords common in classic SQL. Then there seems to be a gray area when it's a mix. On the whole though, my instincts are to start from all caps when possible to maximize approachability to the SQL-centric audience.

docs/commands/super.md
```diff
@@ -139,14 +143,14 @@ The input format is typically [detected automatically](#auto-detection) and the
 "Auto" is "yes" in the table above support _auto-detection_.
 Formats without auto-detection require the `-i` option.
 
-### Hard-wired Input Format
+#### Hard-wired Input Format
```
philrz, line 146: In #5481, a new "Data Formats" section was created and the "Input Formats" and "Output Formats" sub-sections were moved below that, which all makes sense. However, the sub-sections under "Input Formats" and "Output Formats" were left at their prior level, so the hierarchical order was disrupted. This and other `####`-style changes in this PR are to correct that.

mccanne (approved):

PR #5490, opened 2024-11-19
sam: Use SQL semantics for null comparison

mattnibs:
The commit changes the sequential runtime to embrace SQL semantics when comparing values to nulls. This commit also changes the IS NULL expression to not rely on generic equality compare which would produce incorrect results with SQL null semantics.

### DIFF
--- a/compiler/parser/parser.peg
+++ b/compiler/parser/parser.peg
@@ -948,12 +948,11 @@ BetweenExpr
   / ComparisonExpr
 
 ComparisonExpr
-  = lhs:AdditiveExpr _ op:IsNull {
-      return &ast.BinaryExpr{
-        Kind: "BinaryExpr",
-        Op: op.(string),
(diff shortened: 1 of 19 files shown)


compiler/parser/parser.peg
```diff
@@ -948,12 +948,11 @@ BetweenExpr
   / ComparisonExpr
 
 ComparisonExpr
-  = lhs:AdditiveExpr _ op:IsNull {
-      return &ast.BinaryExpr{
-        Kind: "BinaryExpr",
-        Op: op.(string),
-        LHS: lhs.(ast.Expr),
-        RHS: newPrimitive(c, "null", ""),
+  = expr:AdditiveExpr _ IS not:(_ NOT)? _ NULL {
```
mccanne, line 951: You should delete the IsNull peg grammar rule

mccanne (approved):

PR #5494, opened 2024-11-21
vam: Add support for conditionals (case expr)

mattnibs:


runtime/vam/expr/conditional.go
```diff
@@ -0,0 +1,125 @@
...
+				}
+			}
+		}
+	case *vector.Error:
+		if ridx != nil {
+			errs.AddMany(ridx)
+		} else {
+			errs.AddRange(0, uint64(vec.Len()))
+		}
+	}
```
nwt, line 124: Is the absence of a default case here intentional?

runtime/vam/op/filter.go
```diff
@@ -31,18 +31,13 @@ func (f *Filter) Pull(done bool) (vector.Any, error) {
...
+	b, _ := expr.BoolMask(mask)
+	if b.IsEmpty() {
 		return nil, false
 	}
-	if len(index) == int(n) {
+	if b.GetCardinality() == uint64(mask.Len()) {
 		return vec, true
 	}
-	return vector.NewView(vec, index), true
+	return vector.NewView(vec, b.ToArray()), true
```
nwt, line 42: What does this change have to do with implementing conditionals?

mattnibs: Just figured not to have two different boolMask functions plus the new one is more optimized

runtime/vam/expr/conditional.go
mattnibs: 👍🏻 should panic

mattnibs: On second thought I don't think it should. Have no default just sets the return value to false for non-bool vectors just like vector.BoolValue does

nwt (approved):

PR #5495, opened 2024-11-21
vam: type value comparisons

mattnibs:


runtime/vam/expr/gencomparefuncs.go
```diff
@@ -66,11 +66,15 @@ func main() {
...
+	if !(op == "==" || op == "!=") && typ == "TypeValue" {
+		s += "return vector.NewConst(super.False, lhs.Len(), nil)\n}\n"
+		return s
+	}
 	s += genVarInit("l", typ, lhs)
 	s += genVarInit("r", typ, rhs)
 	lexpr := genExpr("l", lhs)
 	rexpr := genExpr("r", rhs)
-	if typ == "Bytes" {
+	if typ == "Bytes" || typ == "TypeValue" {
```
nwt, line 77: This should match what happens on the other side of the fence. https://github.com/brimdata/super/blob/3d706f23fc3c7cc340aee12769cdb4a72659e787/runtime/sam/expr/sort.go#L205-L211.

And since performance on type comparisons doesn't matter much, I don't think it's worth generating code and suggest we just handle it with a loop over in compare.go.

mattnibs: ok

mattnibs: You are referencing sort for equality comparisons on type values we just compare bytes:
https://github.com/brimdata/super/blob/98f94ff9299489ed4f4cae7cef2be11434f7c362/runtime/sam/expr/coerce/coerce.go#L18-L20

nwt (approved):

PR #5496, opened 2024-11-22
Rename: NullsView to BoolView

mattnibs:
Rename NullsView to BoolView since this can be used for creating views of Bool vectors as well. In addition have all an false value views of a boolean vector return a non-nil vector. When dealing with boolean values there's a difference between a false values and a nil vector.

Also have NewView of a boolean vector create a new boolean vector.

### DIFF
--- a/runtime/vam/expr/genarithfuncs.go
+++ b/runtime/vam/expr/genarithfuncs.go
@@ -71,20 +71,19 @@ func genFunc(name, op, typ string, lhs, rhs vector.Form) string {
 	s := fmt.Sprintf("func %s(lhs, rhs vector.Any) vector.Any {\n", name)
 	s += genVarInit("l", typ, lhs)
 	s += genVarInit("r", typ, rhs)
-	s += "nulls := vector.Or(l.Nulls, r.Nulls)\n"
 	if lhs == vector.FormConst && rhs == vector.FormConst {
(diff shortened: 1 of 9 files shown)


runtime/vam/expr/genarithfuncs.go
```diff
@@ -71,7 +71,6 @@ func genFunc(name, op, typ string, lhs, rhs vector.Form) string {
 	s := fmt.Sprintf("func %s(lhs, rhs vector.Any) vector.Any {\n", name)
 	s += genVarInit("l", typ, lhs)
 	s += genVarInit("r", typ, rhs)
-	s += "nulls := vector.Or(l.Nulls, r.Nulls)\n"
```
mattnibs, line 74: This was incorrect when getting nulls of a view.

vector/view.go
```diff
@@ -52,19 +54,19 @@ func NewView(val Any, index []uint32) Any {
 	return &View{val, index}
 }
 
-func NullsView(nulls *Bool, index []uint32) *Bool {
-	if nulls == nil {
+func BoolView(vec *Bool, index []uint32) *Bool {
```
nwt, line 57: Nit: Callt this NewBoolView so it starts with "New" like the other vector constructors.

nwt (approved):

PR #5503, opened 2024-11-25
vam: Fix summarize where

### LINKED ISSUE #5468: Vector query with "where" clause returns incorrect count
Repro is with super commit 411514f.

This is a simplification of the [bench2/q4 query](https://github.com/brimdata/mgbench-private/blob/main/bench2/q4/q4.spq).

Test data is the contents of [repro.jsup.gz](https://github.com/user-attachments/files/17709299/repro.jsup.gz):

```
{log_time:2012-01-01T00:00:44Z,client_ip:249.92.17.134,request:"/courses/cs106/2004/Assignments/rudimentary-interp.html",status_code:304(uint16),object_size:0(uint64)}(=bench2)
{log_time:2012-10-01T00:24:30Z,client_ip:249.92.17.134,request:"/people/sr/",status_code:200(uint16),object_size:2242(uint64)}(=bench2)
{log_time:2012-05-12T10:23:22Z,client_ip:251.58.48.137,request:"/robots.txt",status_code:404(uint16),object_size:506(uint64)}(=bench2)
```

This query against the original Super JSON returns the expected result.

```
$ super -version
Version: v1.18.0-142-g411514fd

$ super -c '
summarize
    num_requests := count()
    where log_time >= 2012-10-01T00:00:00Z
    by client_ip
' repro.jsup.gz 

{client_ip:249.92.17.134,num_requests:1(uint64)}
{client_ip:251.58.48.137,num_requests:0(uint64)}
```

However if I turn the Super JSON into Super Columnar and repeat the same query, now the counts are too high.

```
$ super -f csup -o repro.csup repro.jsup.gz 

$ super dev vector query '
summarize
    num_requests := count()
    where log_time >= 2012-10-01T00:00:00Z
    by client_ip
' repro.csup

{client_ip:251.58.48.137,num_requests:1(uint64)}
{client_ip:249.92.17.134,num_requests:2(uint64)}
```

If I drop the `where` clause, the results match.

mattnibs:
Fix issue with incorrect counts when using where clauses on a aggregation function in vector runtime. If a value fails the where clause make the value as null so it is skipped by the aggregation function.

Closes #5468

### DIFF
--- a/runtime/vam/expr/aggregator.go
+++ b/runtime/vam/expr/aggregator.go
@@ -41,15 +41,19 @@ func (a *Aggregator) Eval(this vector.Any) vector.Any {
 
 func (a *Aggregator) apply(args ...vector.Any) vector.Any {
 	vec, where := args[0], args[1]
-	var tags []uint32
-	// If type is not bool then we want to filter everything.
-	if where.Type().ID() == super.IDBool {
-		for slot := uint32(0); slot < where.Len(); slot++ {
-			// XXX Feels like we should have a optimzed version of this.
-			if vector.BoolValue(where, slot) {
-				tags = append(tags, slot)
-			}
-		}
+	bools, _ := BoolMask(where)
+	if bools.IsEmpty() {
+		// everything is filtered.
+		return vector.NewConst(super.NewValue(vec.Type(), nil), vec.Len(), nil)
 	}
-	return vector.NewView(vec, tags)
+	bools.Flip(0, uint64(vec.Len()))
+	if !bools.IsEmpty() {
+		nulls := vector.NewBoolEmpty(vec.Len(), nil)
+		bools.WriteDenseTo(nulls.Bits)
+		if origNulls := vector.NullsOf(vec); origNulls != nil {
+			nulls = vector.Or(nulls, origNulls)
+		}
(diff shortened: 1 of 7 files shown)


runtime/vam/expr/aggregator.go
```diff
@@ -41,15 +41,19 @@ func (a *Aggregator) Eval(this vector.Any) vector.Any {
...
+		return vector.NewConst(super.NewValue(vec.Type(), nil), vec.Len(), nil)
+	}
+	bools.Flip(0, uint64(vec.Len()))
+	if !bools.IsEmpty() {
+		nulls := vector.NewBoolEmpty(vec.Len(), nil)
+		bools.WriteDenseTo(nulls.Bits)
+		if origNulls := vector.NullsOf(vec); origNulls != nil {
+			nulls = vector.Or(nulls, origNulls)
 		}
+		vector.SetNulls(vec, nulls)
```
nwt, line 56: Modifying a vector that might be shared (as I think `vec` is since it arrived here via `args`) is unsafe since other branches of the flowgraph (and thus other goroutines) may also have a reference to it.

mattnibs: Hmm good point no sure what to do here

PR #5507, opened 2024-11-26
CSUP: Allow multiple objects

mattnibs:
This pr changes the protocol for CSUP files so that a single file can contain multiple vng Objects. When writing a CSUP file a new object is created every 120,000 values.


philrz:
Note to self: Per a heads up from @mattnibs, I'll need to recreate the CSUP files used for mgbench once this merges so we can take advantage of parallel reads.

mccanne (approved):
cool!

PR #5506, opened 2024-11-26
Scripts and AWS results for perf section of super command doc

philrz:
## What's Changing

A set of scripts are introduced to automate the running of the same [perf](https://zed.brimdata.io/docs/next/commands/super#performance) queries shown in #5481. I've updated the super command doc to reflect results from these scripts run on an AWS `m6idn.2xlarge` instance.

For ease of review, I've published a copy of the docs from this branch in a personal staging site [here](https://whimsical-flan-b2925b.netlify.app/docs/next/commands/super#performance).

## Why

1. Automated scripts will make it easier to re-run/publish results over time as devices-under-test evolve.
2. AWS represents "neutral turf" on which anyone can reproduce the results.

## Details

@mccanne established a set of five queries in #5481 that illustrate perf differences between SuperDB, DuckDB, ClickHouse, and DataFusion. The results shown as of that PR were from manual runs on @mccanne's Macbook. The scripts introduced here can be run on an AWS reference instance as well as Macbooks. 

Because the scripts only exist on this PR's branch but we want to hyperlink to them from the docs, there are expected link checker failures at the moment that will go away once this PR merges. I've placed inline comments in the PR to the expected failures.

Regarding the style in which the shell scripts are written, I opt for a style that adheres to most [shellcheck](https://github.com/koalaman/shellcheck) conventions. This may be somewhat more verbose than other styles, but it's consistent and it provides a handy lint-like way to catch mistakes as they're being modified.

PR #5509, opened 2024-11-27
Hyperlink fixes

philrz:
## tl;dr

Make some link checker failures go away.

## Details

I made a mistake in one of my aspirational "it'll start working once it merges" hyperlinks in #5506, so I tried fixing that here. That revealed that the markdown link checker we've been using also seems to have some kind of bug with hyperlinks containing [percent encoding](https://en.wikipedia.org/wiki/Percent-encoding). I just added a [comment](https://github.com/gaurav-nelson/github-action-markdown-link-check/issues/102#issuecomment-2504463659) to an existing issue with a simple repro example, but for now I've added an exclusion in our link checker config. This also led me to notice that the guy who maintained that markdown link checker has a new tool [linkspector](https://github.com/UmbrellaDocs/linkspector) that seems to be more actively maintained, so I'll separately look into whether we can move over to that one and maybe not have to do exclusions.

While I'm at it, I'm also adjusting the hyperlink for an Apache Hive reference. Their design docs happen to be on a public Confluence site and it must not be built for scale because it seems like 50% of the time our link checker experiences some kind of timeout trying to crawl it. Rather than drop the hyperlink entirely or add an exclusion to the link checker I'm opting to just point to something one link further back in their "real" web site.

### DIFF
--- a/.github/workflows/markdown-link-check-config.json
+++ b/.github/workflows/markdown-link-check-config.json
@@ -5,6 +5,9 @@
     },
     {
       "pattern": "^https://suricata.io"
+    },
+    {
+      "pattern": "^https://github.com/brimdata/super/blob/main/scripts/super-cmd-perf/queries/search"
     }
   ],
   "retryOn429": true,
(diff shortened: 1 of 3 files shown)


philrz:
It looks like moving to a different link checker may be a better fix here. Closing this one without merging.

PR #5513, opened 2024-12-03
Include ClickHouse beta JSON type results in super command doc

philrz:
## What's Changing

Query performance results are added to the [`super` command doc](https://zed.brimdata.io/docs/next/commands/super) for the ClickHouse [beta JSON type](https://clickhouse.com/blog/a-new-powerful-json-data-type-for-clickhouse).

## Why

In initial attempts on our Macbooks to load the test data into a table column of this type, we consistently experienced an "too many open files" error. However, when I got the automated scripts running on a larger AWS instance in #5506, I found the data would load successfully into ClickHouse, hence the motivation to circle back and support running this permutation in the scripts and including the results in the doc,


scripts/super-cmd-perf/benchmark.sh
```diff
@@ -69,6 +59,19 @@ if command -v dmidecode && [ "$(sudo dmidecode --string system-uuid | cut -c1-3)
...
 
+  # Install ClickHouse
+  if ! command -v clickhouse-client > /dev/null 2>&1; then
+    sudo apt-get install -y apt-transport-https ca-certificates curl gnupg
+    curl -fsSL 'https://packages.clickhouse.com/rpm/lts/repodata/repomd.xml.key' | sudo gpg --dearmor -o /usr/share/keyrings/clickhouse-keyring.gpg
+    echo "deb [signed-by=/usr/share/keyrings/clickhouse-keyring.gpg] https://packages.clickhouse.com/deb stable main" | sudo tee \
+        /etc/apt/sources.list.d/clickhouse.list
+    sudo apt-get update
+    sudo DEBIAN_FRONTEND=noninteractive apt-get install -y clickhouse-server clickhouse-client
+    sudo cp clickhouse-storage.xml /etc/clickhouse-server/config.d
```
philrz, line 70: I moved the ClickHouse installation after the cloning of the super repo so I could grab this XML config file to make sure the instance SSD was used for DB storage.

docs/commands/super.md
```diff
@@ -759,7 +766,7 @@ For the _search_ test, we'll search for the string pattern
 ```
 in the field `payload.pull_request.body`
 and we'll just count the number of matches found.
-The number of matches is small (3) so the query performance is dominated
+The number of matches is small (2) so the query performance is dominated
```
philrz, line 769: I happened to notice and fix this small error that had been there previously. It's actually the "Search+" test that returns 3 matches.

mccanne (approved):

PR #5519, opened 2024-12-05
Refactor record field name matching in runtime/sam/expr

nwt:
Introduce FieldNameMatcher and use it for all record field name matching.  It will also see use in runtime/vam/expr when we add field name matching there.

The new code fixes a bug in the existing implementation: It now matches fields names inside empty values and null values.


nwt:
@mccanne: I've changed that text to "The new code fixes a bug in the existing implementation: It now matches fields names inside empty values and null values."

PR #5521, opened 2024-12-09
Implement vector non-predicate search terms

nwt:
Specifically, this includes regular expression, glob, keyword, and literal search terms (i.e., dag.Search and dag.RegexpSearch).

Depends on #5520.


nwt:
Note that vector.search.eval is based on (and very similar to) vector.in.eval.

mccanne (approved):

PR #5523, opened 2024-12-10
Implement vector grep() function

nwt:


mattnibs (approved):
We are probably going to want to implement a fast path for string consts on the grep pattern but we can do this later.

PR #5526, opened 2024-12-11
dev vector query: support concatenated csup files

mattnibs:
Update the super dev vector query command so that it support concatenated csup files.


nwt:
The title is "dev vector query: fix CSUP reader" but nothing appears to be broken (all tests pass on main) so what are you fixing?

nwt (approved):

PR #5539, opened 2024-12-18
Implement vector error() function

nwt:
This includes a ztest for the assert operator because the error() function is the final missing piece required to make that operator work on vectors.


mattnibs (approved):
Lovely!

Issue #5550, opened 2024-12-25
Vector CSUP search query consumes 75+ GB memory


philrz:
## tl;dr

I found that running the [Search Test from the `super` command doc](https://zed.brimdata.io/docs/next/commands/super#search-test) with CSUP in vector runtime consumed 75+ GB of RAM, which was enough to hang the EC2 instance with 32 GB of RAM that I'd used to successfully in the past to run the equivalent query with other RDBMS and as well as sequential `super` with BSUP input.

## Details

Repro is with super commit 4084e01. The test data is available at `s3://brim-sampledata/super-cmd-perf/gha.csup` which was generated similar to BSUP as shown [here](https://zed.brimdata.io/docs/next/commands/super#appendix-1-preparing-the-test-data), i.e.,

```
$ super -f csup -o gha.csup gharchive_gz/*.json.gz
```

The query is run like this:

```
$ super -version
Version: v1.18.0-206-g4084e011

$ SUPER_VAM=1 super -c "SELECT count()
FROM 'gha.csup'
WHERE grep('in case you have any feedback 😊', payload.pull_request.body)"
```

In the past I've run these queries successfully on an AWS EC2 `m6idn.2xlarge` instance which has 32 GB of RAM (but no swap), and that's always been enough resource to run the equivalent query successfully as shown in the doc on DuckDB, ClickHouse, DataFusion, and sequential `super` with BSUP input. Previously we'd not been able to run the query at all with `super` in vector runtime, but with the merge of #5523 it was time to give it a go. On the first try with the EC2 instance it consumed all the memory and hung the system.

To give it a closer look, I re-ran it on my Macbook which only has 16 GB of RAM but it does have swap. Keeping an eye on the process in Activity Monitor as it ran for a couple hours, I watched as it climbed to 75+ GB of RAM consumed before it finally did finish. Hoping to get a bit more detail for this issue, I attempted to get a memory profile on a re-run:

```
$ SUPER_VAM=1 super -memprofile=mem.pprof -c "SELECT count()
FROM 'gha.csup'
WHERE grep('in case you have any feedback 😊', payload.pull_request.body)"
```

However, at the very end after it showed the correct result a `Killed: 9` appeared and my `mem.pprof` showed zero length, so I'm not sure what to make of that.

philrz:
The results in #5552 also show that the same test was able to run successfully with vector runtime within the 32 GB memory footprint if Parquet format was used instead of CSUP.

philrz:
[...]

PR #5552, opened 2024-12-28
Run super command perf tests in vector runtime that couldn't before

philrz:
## What's Changing

Query performance results are added to the [`super` command doc](https://zed.brimdata.io/docs/next/commands/super) for tests that couldn't be run previously, thanks to the merge of recent PRs that enabled new functionality in vector runtime.

## Why

This allows us to get rid of some cells in the results summary where we previously had disclaimers saying we couldn't run these yet.

## Details

When making these updates I hoped to show vector CSUP performance with these queries as well, but I bumped into new issue #5550. Therefore I've left the relevant parts of the scripts commented out for now, but will revisit once #5550 is addressed.

### DIFF
--- a/docs/commands/super.md
+++ b/docs/commands/super.md
@@ -681,12 +681,12 @@ with the following software versions:
 
 |**Software**|**Version**|
 |-|-|
-|`super`|Commit `cc6949f`|
+|`super`|Commit `3900a40`|
 |`duckdb`|`v1.1.3` 19864453f7|
 |`datafusion-cli`|datafusion-cli `43.0.0`|
-|`clickhouse`|ClickHouse local version `24.11.1.2557` (official build)|
+|`clickhouse`|ClickHouse local version `24.12.1.1614` (official build)|
 
-The complete run logs are [archived here](https://super-cmd-perf.s3.us-east-2.amazonaws.com/2024-12-03_00-43-29.tgz).
+The complete run logs are [archived here](https://super-cmd-perf.s3.us-east-2.amazonaws.com/2024-12-27_21-58-22.tgz).
 
 ### The Test Data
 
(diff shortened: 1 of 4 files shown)


docs/commands/super.md
```diff
@@ -733,19 +733,18 @@ file:
 ```
 super gharchive_gz/*.json.gz > gha.bsup
 ```
-This code path in `super` is not multi-threaded so not particularly performant but,
-on our test machine, this runs more than 2x faster than the `duckdb` method of
-creating a schema-fused table and just a bit faster than `clickhouse` could
-load the data to its beta JSON type.
+This code path in `super` is not multi-threaded so not particularly performant,
+but on our test machine it runs a bit faster than both the `duckdb` method of
+creating a schema-fused table or loading the data to the `clickhouse` beta JSON type.
```
philrz, line 738: Even though I used the same DuckDB release version and AWS instance type as the last time I ran these tests, for some reason DuckDB did happen to run this "table creation" step substantially faster than last time (328 seconds now vs. 513 seconds last time). I can't explain it offhand and I'm not inspired to take a detour and study it deeply, but in the interest of science I've corrected the summary here to reflect that. I'll keep an eye on it in future runs and maybe the operation just turns out to have high variance.

docs/commands/super.md
```diff
@@ -966,33 +965,29 @@ FROM 'gha.bsup'
...
+_Note: we were not able to successfully run the [union query](#union) with
 ClickHouse's beta JSON type_
 
 Since DuckDB with its native format could successfully run all queries with
 decent performance, we used it as the baseline for all of the speed-up factors.
 
 To summarize,
 `super` with Super Binary is substantially faster than multiple relational systems for
-the search use cases and performs on par with the others for traditional OLAP queries,
+the search use cases, and with Parquet performs on par with the others for traditional OLAP queries,
```
philrz, line 990: Especially now that we have across-the-boards Parquet results, I figured I might as well clarify that those are the ones where `super` perf comes closest with OLAP queries. Looking forward to seeing CSUP showing the best performance. 🤞

mattnibs (approved):

mccanne (approved):

PR #5555, opened 2025-01-06
vector: Fix union nulls

mattnibs:
This commit fixes nulls in vector Unions by added a const null vector of union type into the Union dynamic.

### DIFF
--- a/vector/union.go
+++ b/vector/union.go
@@ -26,13 +26,13 @@ func (u *Union) Type() super.Type {
 func (u *Union) Serialize(b *zcode.Builder, slot uint32) {
 	if u.Nulls.Value(slot) {
 		b.Append(nil)
-	} else {
-		b.BeginContainer()
-		tag := u.Typ.TagOf(u.TypeOf(slot))
-		b.Append(super.EncodeInt(int64(tag)))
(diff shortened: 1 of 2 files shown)


vector/union.go
```diff
@@ -24,11 +24,15 @@ func (u *Union) Type() super.Type {
...
-	b.EndContainer()
+	if u.Nulls.Value(slot) {
+		b.Append(nil)
+	} else {
+		b.BeginContainer()
+		tag := u.Typ.TagOf(u.TypeOf(slot))
+		b.Append(super.EncodeInt(int64(tag)))
+		u.Dynamic.Serialize(b, slot)
+		b.EndContainer()
+	}
```
nwt, line 35: Nit: Indent less.
```suggestion
	if u.Nulls.Value(slot) {
		b.Append(nil)
		return
	}
	b.BeginContainer()
	tag := u.Typ.TagOf(u.TypeOf(slot))
	b.Append(super.EncodeInt(int64(tag)))
	u.Dynamic.Serialize(b, slot)
	b.EndContainer()
```

nwt (approved):
Nice!

PR #5557, opened 2025-01-07
vng: Drop builder

mattnibs:
This pr removes the builder interface from package vng. Instead when loading CSUP files into the sequential runtime the file is loaded as a vector then serialized into row format.


vng/int.go
```diff
@@ -18,18 +19,17 @@ func (p *Int64Encoder) Write(v int64) {
...
+	buf := make([]byte, loc.MemLength)
+	if err := loc.Read(r, buf); err != nil {
+		if err == io.EOF {
+			return nil, nil
+		}
+		return nil, err
+	}
+	var vals []uint32
+	for it := zcode.Iter(buf); !it.Done(); {
+		vals = append(vals, uint32(super.DecodeInt(it.Next())))
```
mattnibs, line 32: These values are currently encoded as signed ints when they should really be encoded as unints, but we'll fix this later.

vng/int.go
```diff
@@ -18,18 +19,17 @@ func (p *Int64Encoder) Write(v int64) {
...
-
-func NewInt64Decoder(loc Segment, r io.ReaderAt) *Int64Decoder {
-	return &Int64Decoder{*NewPrimitiveBuilder(&Primitive{Typ: super.TypeInt64, Location: loc}, r)}
-}
-
-func (p *Int64Decoder) Next() (int64, error) {
-	zv, err := p.ReadBytes()
-	if err != nil {
-		return 0, err
+func ReadUint32Vector(loc Segment, r io.ReaderAt) ([]uint32, error) {
```
nwt, line 22: Nit: I'm aware of the precedent leading to the appearance of "Vector" in this name but think it's confusing given that what "vector" means in the rest of the system, so maybe call this something like ReadUint32s or ReadZcodeUint32s?

nwt (approved):
Nice! Definitely a "less is more" type of situation here.

vng/int.go
mattnibs: Calling it ReadUint32s since using intcomp for uint32 vectors in dynamics, nulls, unions, etc. is the first place I'm using the intcomp library so the zcode stuff will go soon.

PR #5569, opened 2025-01-10
csup: Store dict values in data section

mattnibs:
This pr change the structure of how dicts in CSUP files are store so that the values in a dictionary are no longer stored in the dictionary's metadata but rather in it's data section.

### DIFF
--- a/vng/dict.go
+++ b/vng/dict.go
@@ -1,0 +1,143 @@
+package vng
+
+import (
+	"io"
+	"math"
+
+	"github.com/brimdata/super"
+	"github.com/brimdata/super/zcode"
+	"golang.org/x/sync/errgroup"
+)
+
+type DictEncoder struct {
+	typ    super.Type
+	tags   map[string]byte
+	index  []byte
+	counts *Uint32Encoder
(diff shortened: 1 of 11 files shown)


vng/dict.go
```diff
@@ -0,0 +1,143 @@
...
+		bytes = []byte(b)
+	}
+	return &Const{
+		Value: super.NewValue(d.typ, bytes),
+		Count: uint32(len(d.index)),
+	}
+}
+
+func (d *DictEncoder) isValid() bool {
+	return d.tags != nil && len(d.tags) >= 1 && len(d.index) > math.MaxInt8
```
nwt, line 69: `len(d.index) > math.MaxInt8` doesn't feel right. In particular, there's no benefit to dictionary encoding if `len(d.index) == len(d.tags)`, so I think what we want is something like `len(d.index) > len(d.tags)` or maybe `len(d.index) > len(d.tags) + nonzeroThreshold`.

mattnibs: Good point

nwt (approved):

PR #5571, opened 2025-01-14
Document "line" output format

philrz:
## tl;dr

This adds user-facing docs for the `-f line` output format added in #5570.

## Details

It seemed appropriate to disclose the backslash escape sequences that affect this output. Since the changes in #5570 indicate we hand the job off to Go's `fmt.Fprintln()`, I'd normally have been fine just hyperlinking to some kind of table in the Go docs that disclose this info, but I couldn't seem to find such a table. I asked ChatGPT and it also complained that the official docs were sparse, but it did produce a table of its own and I confirmed via black box testing which ones seem to work, so that's what's in my table here. If those of you who know the code can point to a better reference or can spot entries missing from the table, please speak up.

While making these changes I noticed that I let the tables for input/output formats get out of alphabetical order in #5387, so I address that here as well.

### DIFF
--- a/docs/commands/super.md
+++ b/docs/commands/super.md
@@ -133,11 +133,11 @@ in the order appearing on the command line forming the input stream.
 | `csv`     |  yes | [Comma-Separated Values (RFC 4180)](https://www.rfc-editor.org/rfc/rfc4180.html) |
 | `json`    |  yes | [JSON (RFC 8259)](https://www.rfc-editor.org/rfc/rfc8259.html) |
 | `jsup`    |  yes | [Super JSON](../formats/jsup.md) |
-| `zjson`   |  yes | [Super JSON over JSON](../formats/zjson.md) |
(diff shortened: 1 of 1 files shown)


docs/commands/super.md
```diff
@@ -399,12 +400,39 @@ be used with any output format.
 
 #### Simplified Text Outputs
 
-The `text` and `table` formats simplify data to fit within the
-limitations of text-based output. Because they do not capture all the
-information required to reconstruct the original data, they are not supported
-input formats. They may be a good fit for use with other text-based shell
+The `line`, `text`, and `table` formats simplify data to fit within the
+limitations of text-based output. They may be a good fit for use with other text-based shell
```
philrz, line 404: I decided to keep the details for all three of these formats still in one section but gave `line` top billing because it comes with the fewest caveats.

nwt (approved):

PR #5582, opened 2025-01-17
Handle dag.DefaultScan in kernel.Builder.compileVamLeaf

nwt:
This allows the super command to read from files specified on the command line (e.g., "super f.zson") when using the vector runtime.

Depends on #5581.

### DIFF
--- a/compiler/kernel/vop.go
+++ b/compiler/kernel/vop.go
@@ -176,6 +176,12 @@ func (b *Builder) compileVamLeaf(o dag.Op, parent vector.Puller) (vector.Puller,
 			return nil, err
 		}
 		return vamop.NewYield(b.zctx(), parent, []vamexpr.Evaluator{e}), nil
+	case *dag.DefaultScan:
+		zbufPuller, err := b.compileLeaf(o, nil)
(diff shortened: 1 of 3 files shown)


compiler/kernel/vop.go
```diff
@@ -176,6 +176,12 @@ func (b *Builder) compileVamLeaf(o dag.Op, parent vector.Puller) (vector.Puller,
 			return nil, err
 		}
 		return vamop.NewYield(b.zctx(), parent, []vamexpr.Evaluator{e}), nil
+	case *dag.DefaultScan:
```
mattnibs, line 179: Won't this change make it so CSUP and Parquet are materialized then dematerialized before hitting the vector runtime?

nwt: Yes, but it's a step forward from this.

```console
$ SUPER_VAM=1 super f.csup
vector runtime does not support internal readers
```

Fixing things so it we skip that extra work requires some refactoring of the way files on the command line are handled, and I look forward to doing that.

mattnibs (approved):

PR #5586, opened 2025-01-18
vam: Add collect aggregation

mattnibs:


mattnibs:
This implementation does not keep the in-order behavior that the sequential `collect` version does (i.e., the resultant array is in order from which values were received). I do not think it would be easy to have this behavior without completely rewriting how we do aggregations writ large.

runtime/sam/expr/agg/collect.go
```diff
@@ -2,61 +2,64 @@ package agg
...
 	"github.com/brimdata/super"
 	"github.com/brimdata/super/zcode"
 	"github.com/brimdata/super/zson"
 )
 
 type Collect struct {
-	values []super.Value
-	size   int
+	types []super.Type
+	bytes []zcode.Bytes
```
nwt, line 14: Why replace `values` with `types` and `bytes`?

philrz:
In a group discussion @mattnibs clarified that the "does not keep the in-order behavior" cited in the comment above only becomes true when there's values of multiple types being collected. So for the simple case of something like:

```
$ seq 1 100000 | super -c 'collect(this)' -
```

...the contents of the array are still expected to be in order.

runtime/sam/expr/agg/collect.go
mattnibs: I'll just change it back

mattnibs: Actually never mind. I change it because otherwise you need to do `super.NewValue` when adding values for sequential ConsumeAsPartial and both vector Consume methods, so it felt like better just keep two arrays. We already have to create an array of all types when the innerType is created.

runtime/ztests/op/summarize/collect-union.yaml
```diff
@@ -0,0 +1,11 @@
+zed: summarize collect(this) | sort this
+
+vector: true
+
+input: |
+  {a:1}
+  {a:2}
+  {b:1.5}
```
nwt, line 8: Add a `null` so we know we're doing the right thing with it?

runtime/ztests/op/summarize/collect-union.yaml
```diff
@@ -0,0 +1,11 @@
+zed: summarize collect(this) | sort this
```
nwt, line 1: `sort` does nothing here since the output is a single value.
```suggestion
zed: summarize collect(this)
```

runtime/sam/expr/agg/collect.go
nwt: OK. Fine to merge it.

But if the concern around `super.NewValue` is performance, it's free since it's inlined. And even if it weren't, complicating code in pursuit of performance without measurement is usually the wrong move.

nwt (approved):

runtime/ztests/op/summarize/collect-union.yaml
mattnibs: Hopefully ignoring it

runtime/sam/expr/agg/collect.go
mattnibs: After thinking about it I agree I think my changes does complicate this a bit. I'll change it back.

PR #5588, opened 2025-01-20
Allow sequence index expression on array or set to return null

nwt:
In the sequence runtime, indexing a null value inside a non-null array
or set returns error("missing").  Return the null value instead both for
consistency with the vector runtime and because it's less surprising.

### DIFF
--- a/runtime/sam/expr/function/ip.go
+++ b/runtime/sam/expr/function/ip.go
@@ -82,8 +82,6 @@ type CIDRMatch struct {
 var errMatch = errors.New("match")
 
 func (c *CIDRMatch) Call(_ super.Allocator, args []super.Value) super.Value {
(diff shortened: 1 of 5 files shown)


runtime/sam/expr/function/ip.go
```diff
@@ -82,6 +82,8 @@ type CIDRMatch struct {
 var errMatch = errors.New("match")
 
 func (c *CIDRMatch) Call(_ super.Allocator, args []super.Value) super.Value {
+	//fmt.Println(zson.FormatValue(args[0]))
+	//fmt.Println(zson.FormatValue(args[1]))
```
mattnibs, line 86: delete these lines
```suggestion
```

mattnibs (approved):

PR #5592, opened 2025-01-21
Implement vector cidr_match() function

nwt:


runtime/vam/expr/function/cidrmatch.go
```diff
@@ -0,0 +1,46 @@
...
+			continue
+		}
+		ip, null := vector.IPValue(ipVec, i)
+		if null {
+			continue
+		}
+		if net.Contains(ip) {
+			out.Set(i)
+		}
+	}
```
mattnibs, line 44: In order for this to be consistent with other boolean logic I think we need to have this return `null(bool)` when nulls are encountered.

nwt: @mattnibs: I think I fixed the null handling in here so please have another look.

mattnibs (approved):
Lovely

PR #5611, opened 2025-01-30
vam: Add grok()

mattnibs:


runtime/ztests/expr/function/grok.yaml
```diff
@@ -0,0 +1,60 @@
+zed: yield grok(pattern, input, defs)
+
+vector: true
+
+input: |
+  {
+    pattern: "%{TIMESTAMP_ISO8601:event_time} %{LOGLEVEL:log_level} %{GREEDYDATA:log_message}",
+    input: "2020-09-16T04:20:42.45+01:00 DEBUG This is a sample debug log message",
+    defs: null(string)
+  }
```
nwt, line 10: While you're here, maybe add test cases for null(string) pattern and input along with non-string pattern, input, and defs?

runtime/vam/expr/function/grok.go
```diff
@@ -0,0 +1,134 @@
...
+func newGrok(zctx *super.Context) *Grok {
+	return &Grok{
+		zctx:  zctx,
+		hosts: make(map[string]*host),
+	}
+}
+
+func (g *Grok) Call(args ...vector.Any) vector.Any {
+	patternArg, inputArg := args[0], args[1]
+	var defArg vector.Any = vector.NewConst(super.NullString, args[0].Len(), nil)
```
nwt, line 29: Nit:
```suggestion
	defArg := vector.Any(vector.NewConst(super.NullString, args[0].Len(), nil))
```

nwt (approved):

runtime/ztests/expr/function/grok.yaml
mattnibs: What do you think they should do with null values? Is the current implementation incorrect?

mattnibs: I'll just have it return a null empty record when null inputs are encountered. Seems consistent.

PR #5613, opened 2025-01-31
vam: Support NullScan

mattnibs:


compiler/kernel/vop.go
```diff
@@ -202,6 +202,8 @@ func (b *Builder) compileVamLeaf(o dag.Op, parent vector.Puller) (vector.Puller,
 		return vamop.NewFilter(b.zctx(), parent, e), nil
 	case *dag.Head:
 		return vamop.NewHead(parent, o.Count), nil
+	case *dag.NullScan:
+		return vector.NewPuller(vector.NewConst(super.Null, 1, nil)), nil
```
nwt, line 206: Nit: No need for vector.NewPuller.
```suggestion
		return vam.NewDematerializer(zbuf.NewPuller(zbuf.NewArray([]super.Value{super.Null}))), nil

```

nwt (approved):

compiler/kernel/vop.go
mattnibs: I think this is kind of weird but will do it your way until there's a legit use case for a performant vector.NewPuller (perhaps this day will not come!)

PR #5616, opened 2025-02-03
vam: regexp()

mattnibs:


runtime/sam/expr/function/regexp.go
```diff
@@ -27,7 +27,7 @@ func (r *Regexp) Call(_ super.Allocator, args []super.Value) super.Value {
 		r.re, r.err = regexp.Compile(r.restr)
 	}
 	if r.err != nil {
-		return r.zctx.NewErrorf("regexp: %s", r.err)
+		return r.zctx.WrapError("regexp: invalid regular expression", args[0])
```
nwt, line 30: Nit: I think we should include regexp/syntax.Error.Code in the error string. (Similar change required in runtime/vam/expr/function/regexp.go of course.)
```suggestion
		msg := "regexp: invalid regular expression"
		if syntaxErr, ok := r.err.(*syntax.Error); ok {
			msg += ": " + syntaxErr.Code.String()
		}
		return r.zctx.WrapError(msg, args[0])
```

nwt (approved):

PR #5618, opened 2025-02-03
vam: flatten()

mattnibs:


runtime/vam/expr/function/flatten.go
```diff
@@ -0,0 +1,33 @@
...
+func (f *flatten) Call(args ...vector.Any) vector.Any {
+	vec := vector.Under(args[0])
+	rtyp := super.TypeRecordOf(vec.Type())
+	if rtyp == nil {
+		return args[0]
+	}
+	builder := vector.NewDynamicBuilder()
+	var b zcode.Builder
+	for i := range vec.Len() {
+		b.Reset()
```
nwt, line 27: It's safe to use this here.
```suggestion
		b.Truncate()
```

nwt (approved):

PR #5619, opened 2025-02-04
Remove typeunder()

mattnibs:


nwt (approved):
docs/language/functions/under.md still mentions typeunder() so remove that before you merge.

PR #5623, opened 2025-02-05
vam: pow()

mattnibs:


nwt (commented):
Add `vector: true` to float32.yaml!

nwt (approved):

PR #5632, opened 2025-02-06
vam: parse_uri()

mattnibs:
Also fix output flags for vector ztests

### DIFF
--- a/runtime/sam/expr/function/parse.go
+++ b/runtime/sam/expr/function/parse.go
@@ -1,6 +1,7 @@
 package function
 
 import (
+	"fmt"
 	"net/url"
 	"strconv"
 	"strings"
@@ -16,17 +17,12 @@ type ParseURI struct {
 	marshaler *zson.MarshalZNGContext
 }
 
+func NewParseURI(zctx *super.Context) *ParseURI {
+	return &ParseURI{zctx, zson.NewZNGMarshalerWithContext(zctx)}
+}
+
 func (p *ParseURI) Call(_ super.Allocator, args []super.Value) super.Value {
-	in := args[0]
-	if !in.IsString() || in.IsNull() {
-		return p.zctx.WrapError("parse_uri: non-empty string arg required", in)
-	}
-	s := super.DecodeString(in.Bytes())
-	u, err := url.Parse(s)
-	if err != nil {
-		return p.zctx.WrapError("parse_uri: "+err.Error(), in)
-	}
-	var v struct {
+	type uri struct {
 		Scheme   *string    `zed:"scheme"`
 		Opaque   *string    `zed:"opaque"`
 		User     *string    `zed:"user"`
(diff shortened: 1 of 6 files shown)


runtime/sam/expr/function/parse.go
```diff
@@ -1,79 +1,119 @@
...
 	"github.com/brimdata/super/zson"
 )
 
 // https://github.com/brimdata/super/blob/main/docs/language/functions.md#parse_uri
 type ParseURI struct {
-	zctx      *super.Context
-	marshaler *zson.MarshalZNGContext
+	zctx    *super.Context
+	typ     super.Type
+	builder zcode.Builder
```
nwt, line 19: Nit: Rewriting this in terms of zcode.Builder doesn't seem necessary for adding parse_uri() to vam and makes makes the code more complicated.

runtime/sam/expr/function/parse.go
```diff
@@ -1,79 +1,119 @@
...
+		for k, list := range q {
+			mb.Append([]byte(k))
+			mb.BeginContainer()
+			for _, v := range list {
+				mb.Append([]byte(v))
+			}
+			mb.EndContainer()
+		}
+		mb.EndContainer()
+		b.Append(super.NormalizeMap(mb.Bytes().Body()))
```
nwt, line 107: Instead of creating another zoce.Builder like this, use zcode.Builder.TransformContainer. 
```suggestion
		b.BeginContainer()
		for k, list := range q {
			b.Append([]byte(k))
			b.BeginContainer()
			for _, v := range list {
				b.Append([]byte(v))
			}
			b.EndContainer()
		}
		b.TransformContainer(super.NormalizeMap)
		b.EndContainer()
```

runtime/vam/expr/function/parse.go
```diff
@@ -0,0 +1,48 @@
...
+				continue
+			}
+			vb.Write(p.builder.Bytes())
+		}
+	}
+	out := vb.Build()
+	if len(errs) > 0 {
+		return vector.Combine(out, errs, vector.NewVecWrappedError(p.zctx, errMsgs, vector.NewView(args[0], errs)))
+	}
+	return out
```
nwt, line 47: Nit: If samfunc.EncodeURI always returned a super.Value, you could simplify this to
```go
	nullType := samfunc.URIType()
	b := NewDynamicBuilder()
	for i := range vec.Len()
		var val super.Value
		s, null := vector.StringValue(vec, i)
		if null {
			val = super.NewValue(nullType, nil)
		} else {
			val = samfunc.EncodeURI(s)
		}
		b.Write(val)
	}
	return b.Build()
```

nwt (approved):
I'm in favor of a smaller diff in runtime/sam/expr/function/parse.go and a simpler ParseURI.Call in runtime/vam/expr/function/parse.go but this works so merge if you like it.

runtime/sam/expr/function/parse.go
mattnibs: I had one this to support null return values but am now realizing I can still do this using the old method. I'll change it back.

PR #5635, opened 2025-02-06
vam: unflatten()

mattnibs:


runtime/vam/expr/function/function.go
```diff
@@ -115,6 +115,8 @@ func New(zctx *super.Context, name string, narg int) (expr.Function, field.Path,
 		f = &TypeOf{zctx}
 	case "upper":
 		f = &ToUpper{zctx}
+	case "unflatten":
+		f = newUnflatten(zctx)
 	case "under":
 		f = &Under{zctx}
```
nwt, line 121: Nit:
```suggestion
	case "under":
		f = &Under{zctx}
	case "unflatten":
		f = newUnflatten(zctx)
	case "upper":
		f = &ToUpper{zctx}
```

runtime/vam/expr/function/flatten.go
```diff
@@ -31,3 +31,25 @@ func (f *flatten) Call(args ...vector.Any) vector.Any {
...
+	return &unflatten{samfunc.NewUnflatten(zctx)}
+}
+
+func (u *unflatten) Call(args ...vector.Any) vector.Any {
+	vec := vector.Under(args[0])
+	typ := vec.Type()
+	builder := vector.NewDynamicBuilder()
+	var b zcode.Builder
+	for i := range vec.Len() {
+		b.Reset()
```
nwt, line 49: Nit: Should be safe to use Truncate here since vector.DynamicBuilder.Write doesn't retain a reference to the bytes returned by val.Bytes().
```suggestion
		b.Truncate()
```

nwt (approved):

PR #5639, opened 2025-02-07
vam: Propagate errors in arith exprs

mattnibs:


runtime/vam/expr/arith.go
```diff
@@ -29,6 +29,12 @@ func (a *Arith) Eval(val vector.Any) vector.Any {
 func (a *Arith) eval(vecs ...vector.Any) (out vector.Any) {
 	lhs := enumToIndex(vector.Under(vecs[0]))
 	rhs := enumToIndex(vector.Under(vecs[1]))
+	if vector.IsError(lhs) {
+		return lhs
+	}
+	if vector.IsError(rhs) {
```
nwt, line 35: Nit: We already have a way to do this so I don't think we should add vector.IsError.
```suggestion
	if lhs.Type().Kind() == super.ErroKind {
		return lhs
	}
	if rhs.Type().Kind() == super.ErroKind {
```

nwt (approved):

PR #5640, opened 2025-02-07
vam: has(), missing()

mattnibs:


runtime/vam/expr/function/has.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func bitmapToBool(b *roaring.Bitmap, len uint32) *vector.Bool {
+	if b.IsEmpty() {
+		return nil
+	}
+	return vector.NewBool(b.ToDense(), len, nil)
+}
+
+func missingOrQuietV2(verr *vector.Error) *roaring.Bitmap {
```
nwt, line 60: Nit: Why V2? How about isMissingOrQuiet?

runtime/vam/expr/function/has.go
```diff
@@ -0,0 +1,100 @@
...
+	b := roaring.New()
+	inner := verr.Vals
+	if inner.Type() != super.TypeString {
+		// Non-error types should
```
nwt, line 64: Should what?

nwt (approved):

runtime/vam/expr/function/has.go
mattnibs: ha my bad at some point this was v2

runtime/vam/expr/function/has.go
mattnibs: I have no idea what I was going for here I'll drop the comment

PR #5642, opened 2025-02-07
Another iteration on the Introduction doc

philrz:
As I confront the need to get back to de-Zed-ifying the docs, I'm inclined to take it from the top again. The last time I revised this "Introduction" doc we were in the brief phase when we were saying "SuperPipe", so I yank that out here along with a few other adjustments that feel appropriate now that things have settled a bit more.

I recognize the whole "Design Philosophy" section is still somewhat suspect since we're not heavily emphasizing the lake at the moment, but I'm inclined to leave it where it is until we have a better sense of its fate.

### DIFF
--- a/docs/_index.md
+++ b/docs/_index.md
@@ -13,7 +13,7 @@ without giving up JSON's uncanny ability to represent eclectic data.
 ## Getting Started
 
 Trying out SuperDB is easy: just [install](install.md) the command-line tool
-[`super`](commands/super.md) and run through the [tutorial](tutorials/zq.md).
+[`super`](commands/super.md) and run through its [usage documentation](commands/super.md).
 
 Compared to putting JSON data in a relational column, the
 [super-structured data model](formats/zed.md) makes it really easy to
@@ -27,7 +27,7 @@ while for a technical user, SuperDB exposes its technical underpinnings
 in a gradual slope, providing as much detail as desired,
 packaged up in the easy-to-understand
 [Super JSON data format](formats/jsup.md) and
-[SuperPipe language](language/_index.md).
+[SuperSQL language](language/_index.md).
 
 While `super` and its accompanying data formats are production quality, the project's
 [SuperDB data lake](commands/super-db.md) is a bit earlier in development.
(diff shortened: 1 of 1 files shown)


docs/_index.md
```diff
@@ -13,7 +13,7 @@ without giving up JSON's uncanny ability to represent eclectic data.
 ## Getting Started
 
 Trying out SuperDB is easy: just [install](install.md) the command-line tool
-[`super`](commands/super.md) and run through the [tutorial](tutorials/zq.md).
+[`super`](commands/super.md) and run through its [usage documentation](commands/super.md).
```
philrz, line 16: We talked in a team meeting recently about how the doc formerly known as the "zq tutorial" now reads more like a "`super` guide for `jq` users". It seems like it'll be good to keep that on in some form, but it's no longer the best starting point for new users. Meanwhile, @mccanne already overhauled the `super` command doc with lots of SQL-centric content, perf comparisons, and other goodies, so it's clearly a superior choice to link to.

docs/_index.md
```diff
@@ -38,30 +38,28 @@ While `super` and its accompanying data formats are production quality, the proj
 a number of different elements of the system:
 * The [super data model](formats/zed.md) is the abstract definition of the data types and semantics
 that underlie the super-structured data formats.
-* The [super data formats](formats/_index.md) are a family of
+* The [super-structured data formats](formats/_index.md) are a family of
```
philrz, line 41: With the exception of "super data model", I'm continuing to make all attempts to avoid using "super" as a standalone adjective more than necessary. We actually haven't been saying "super data model" anywhere else yet, so it seems appropriate to match it here with how it was described in the previous bullet.

docs/_index.md
```diff
@@ -38,30 +38,28 @@ While `super` and its accompanying data formats are production quality, the proj
...
-* A  [SuperPipe query](language/overview.md) is a script that performs
-search and/or analytics.
-* A [SuperPipe shaper](language/shaping.md) is a script that performs
-data transformation to _shape_
-the input data into the desired set of organizing super-structured data types called "shapes",
-which are traditionally called _schemas_ in relational systems but are
-much more flexible in SuperDB.
+* A [SuperSQL pipe query (SPQ)](language/overview.md) is a query that
+employs SuperSQL's unique pipeline extensions and shortcuts to perform data
+operations that are difficult or impossible in standard SQL.
```
philrz, line 50: I'd appreciate attention here that folks are on-board with this description. We want it to be clear to users that we'll be able to handle standard SQL queries, but we've got the pipeline extensions and shortcuts, so I've heard rough consensus around "SPQ" as the shorthand when talking about SuperSQL queries that include more than the standard stuff, and how it could also appear as a `.spq` extension for queries saved in files so a user knows they can't expect it contents to run seamlessly on some other SQL-based RDBMS without modification. When we get to filling out the intro language page more we'll surely go into detail, but I figure it's appropriate in the intro to plant the seed that we're offering more than standard SQL.

docs/_index.md
```diff
@@ -38,30 +38,28 @@ While `super` and its accompanying data formats are production quality, the proj
...
-The [SuperPipe language documentation](language/_index.md)
-is the best way to learn about `super` in depth.
-All of its examples use `super` commands run on the command line.
-Run `super -h` for a list of command options and online help.
+The [SuperSQL language documentation](language/_index.md)
+is the best way to learn about `super` in depth. Most
+[examples](commands/super.md#examples) that appear throughout the docs can be
+executed right in your browser and can easily be copied to the command line
+for execution with `super`. Run `super -h` for a list of command options and
+brief help.
```
philrz, line 62: This revision is largely a nod to the new Playground presentation that also includes the CLI tab. See also #5627.

PR #5659, opened 2025-02-17
Implement vector join operator

nwt:
This is a simple hash join with lots of room for improvement.  Because
it must read its entire right input before reading any values from its
left input, it can deadlock on self joins.

For deterministic output, it requires the addition of an output sort in
tests that don't already have one.

(runtime/ztests/parallel.yaml is removed as it duplicates other tests.)


philrz:
A couple of the mgbench benchmarks have not been able to run yet in vector runtime due to lack of `join` support, so I decided to test out this branch. Taking [bench3/q2](https://github.com/brimdata/mgbench-private/blob/main/bench3/q2/q2.spq) as an example, I suspect there may be a deadlock. Using the test data at `s3://s3://brim-sampledata/mgbench/bench3.parquet`, when I run the following on my Macbook:

```
$ super -version
Version: v1.18.0-286-gfa0918bf

$ SUPER_VAM=1 super -c 'from "/Users/phil/bench3.parquet"
| where
    event_type=="power"
    and log_time >= 2019-11-01T00:00:00Z
| yield {
    hr: uint8(strftime("%H", log_time)),
    device_id,
    device_name,
    device_category: grep(coffee*, device_name)
      ? "coffee"
      : grep(printer*, device_name)
        ? "printer"
        : grep(projector*, device_name)
          ? "projector"
          : grep(vending*, device_name)
            ? "vending"
            : "other",
    device_floor,
    event_value
  }
| fork (
  => summarize
       power_avg := avg(event_value)
       by hr,
          device_id,
          device_name,
          device_category,
          device_floor

  => fork (
       => yield {device_id, device_category}
          | summarize union(this)
          | over this

       => pass
     )
     | join on device_category=device_category power_hourly := this

     | where device_id != power_hourly.device_id
     | summarize
         category_power_avg := avg(power_hourly.event_value)
         by device_id,
            power_hourly.hr,
            power_hourly.device_category
)
| switch (
  case has(power_avg) => put left_key := {device_id, hr}
  case has(category_power_avg) => put right_key := {device_id, hr:power_hourly.hr}
)
| join on left_key=right_key category_power_avg
| where power_avg > category_power_avg * 2
| drop left_key
| sort
    device_id,
    hr'
```

...the `super` process in Activity Monitor stays consistently at 0% CPU consumed no matter how long I wait. But if I drop the `SUPER_VAM=1` and repeat, it chugs away at ~150% CPU consumed for a couple minutes and does return a result.

nwt:
@philrz: This is expected. I neglected to mention in the PR description that this implementation deadlocks on self joins with `... | fork (=> ... => ...) | join ...` if neither fork branch contains a source operator (like `from`). We'll remove this limitation soon.

mattnibs (approved):

PR #5661, opened 2025-02-18
Add fix to intcomp

mattnibs:


nwt (approved):
`go mod tidy` and then :shipit:.

PR #5663, opened 2025-02-18
compare(): return error when error is encountered

mattnibs:


nwt (approved):
Nit: Add test coverage for this in runtime/sam/expr/function/ztests/compare.yaml.

PR #5665, opened 2025-02-19
Set expectations in README and first docs pages

philrz:
As @mccanne has plans to start introducing some new audiences to SuperDB, we want to be careful in setting expectations, since those users may arrive at the project README and/or docs pages ready to tinker and we want their first impressions to be in line with the project's status. To that end, the changes in this branch adjust the tone of the README and some of the first docs pages from something like "come 'n' get it!" to something more like "try it... expect some speedbumps... we welcome your feedback".

You can see the rendered version of the key part of the README [here](https://github.com/brimdata/super/blob/setting-expectations/README.md#try-it), and the rendered change in one of the docs pages looks like the bottom box shown in this screenshot:

image

Once we've got consensus on the language used here or some adjusted variation, I plan to do two immediate follow-on changes:

1. I'd put up a PR to propose some similar adjustments in the https://superdb.org front page, as that also has a bit of "come 'n' get it!" tone at the moment.

2. I'd hide the **Releases** area circled in red in the following screenshot, since (in line with the text) we're not yet offering GA releases of SuperDB, and we don't want to confuse new users if they should stumble across the releases from the older "Zed" project from which SuperDB evolved in this same repo. We have many legacy users that still use Zed and there's links to the releases from blog articles and other web sites, and hiding that **Releases** area does _not_ disable the actual https://github.com/brimdata/super/releases page or release artifacts if the user happens to access them by URL. Indeed, at the moment we still mention the availability of the older Zed releases at https://www.brimdata.io and this approach allows the links to the various artifacts from the [Download](https://www.brimdata.io/download/) page to remain active. Perhaps at some point we may choose to separate from all history of the Zed project to eliminate any possible confusion among the new SuperDB user base. However, as we intend to help those users transition to SuperDB if possible, I believe we have consensus on the "hide" approach for now.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,11 @@
 # SuperDB [![Tests][tests-img]][tests] [![GoPkg][gopkg-img]][gopkg]
 
+> 🔴 **NOTICE OF PROJECT READINESS** 🔴
+>
+> SuperDB is still under development so there's not yet a GA release.
+> You're welcome to [try it out](#try-it) in its early form (i.e.,
+> tip-of-`main`) and we'd love to hear your feedback. Read on for more info!
+
 SuperDB is a new analytics database that supports relational tables and JSON
 on an equal footing.  It shines when it comes to data wrangling where
 you need to explore or process large eclectic data sets.  It's also pretty
--- a/README.md
+++ b/README.md
@@ -67,34 +67,44 @@ representation of Super JSON (a bit like Parquet).
 Even though SuperDB is based on these super-structured data formats, it can read and write
 most common data formats.
 
-## Try It
-
-Trying out SuperDB is super easy: just [install](https://superdb.org/docs/install)
-the command-line tool [`super`](https://superdb.org/docs/commands/super).
-
-Detailed documentation for the entire SuperDB system and its piped SQL syntax
-is available on the [SuperDB docs site](https://superdb.org/docs).
-
-The SuperDB query engine can run locally without a storage engine by accessing
-files, HTTP endpoints, or S3 paths using the `super` command. While
-[earlier in its development](https://superdb.org/docs/commands/super-db/#status),
-SuperDB can also run on a
-[super-structured data lake](https://superdb.org/docs/commands/super-db/#the-lake-model)
-using the `super db` sub-commands.
-
 ## Project Status
 
 Our long-term goal for SuperSQL is to be Postgres-compatible and interoperate
 with existing SQL tooling. In the meantime, SuperSQL is a bit of a moving 
-target and we would love community engagement to evolve and fine tune its
+target and we would love [community engagement](#join-the-community) to evolve and fine tune its
 syntax and semantics.
 
 Our areas of active development include:
 * the SuperSQL query language,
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -67,35 +67,45 @@ representation of Super JSON (a bit like Parquet).
...
-is available on the [SuperDB docs site](https://superdb.org/docs).
-
-The SuperDB query engine can run locally without a storage engine by accessing
-files, HTTP endpoints, or S3 paths using the `super` command. While
-[earlier in its development](https://superdb.org/docs/commands/super-db/#status),
-SuperDB can also run on a
-[super-structured data lake](https://superdb.org/docs/commands/super-db/#the-lake-model)
-using the `super db` sub-commands.
-
 ## Project Status
```
philrz, line 70: I moved the **Project Status** section _above_ the **Try It** section since I figured it was good to draw their attention first to the project's early status.

docs/_index.md
```diff
@@ -29,7 +44,7 @@ packaged up in the easy-to-understand
 [Super JSON data format](formats/jsup.md) and
 [SuperSQL language](language/_index.md).
 
-While `super` and its accompanying data formats are production quality, the project's
+While `super` and its accompanying data formats are production quality for some use cases, the project's
```
philrz, line 47: I figured saying a flat "production quality" clashed with our message of "expect speedbumps". Those of us who work on it know that years of real world usage of `zq` has made `super` in its current form quite solid for many production use cases, e.g., things for which one might otherwise reach for `jq`. But that history will not be obvious to a new SQL-centric user, so I figured we could just tone down the language a tad.

docs/install.md
```diff
@@ -20,19 +37,12 @@ brew install brimdata/tap/super
 
 Once installed, run a [quick test](#quick-tests).
 
-## Binary Download
-
-We offer pre-built binaries for macOS, Windows and Linux for both amd64/x86 and arm
-architectures in the super [GitHub Release page](https://github.com/brimdata/super/releases).
-
-Once the `super` binary is unpacked from a downloaded package, run a [quick test](#quick-tests).
-
```
philrz, line 29: Since the plan is to hide the Releases link from the repo's front page to reflect that there's not yet GA SuperDB releases, it seems we should drop it here as well. Legacy Zed users will hopefully find their way to the older releases via https://www.brimdata.io and https://www.brimdata.io/download .

docs/install.md
```diff
@@ -20,19 +37,12 @@ brew install brimdata/tap/super
...
 ## Building From Source
 
 If you have Go installed, you can easily build `super` from source:
 
 ```bash
-go install github.com/brimdata/super/cmd/super@latest
+go install github.com/brimdata/super/cmd/super@main
```
philrz, line 45: `@latest` would still get them the last GA Zed v1.18.0 release, which is not what we want. Since the README/docs changes in this PR set expectations that they're installing code that's still under development, tip-of-`main` is exactly what we want them to do for now.

mccanne (approved):

PR #5667, opened 2025-02-19
csup: Add metadata filtering

mattnibs:


philrz:
FYI, testing with this branch at commit 00939e9 I've confirmed that all 16 of the mgbench queries that can currently be run in vector runtime still return correct results. 👍

vng/metadata.go
```diff
@@ -265,6 +266,68 @@ func (d *Dynamic) Len() uint32 {
 	return d.Length
 }
 
+func ZoneMapValues(zctx *super.Context, m Metadata) []super.Value {
```
nwt, line 269: There's no zone map here so I think MetadataValues would be a better name.

compiler/dag/op.go
```diff
@@ -210,11 +210,12 @@ type (
...
-		Path   string       `json:"path"`
-		Format string       `json:"format"`
-		Fields []field.Path `json:"fields"`
-		Filter Expr         `json:"filter"`
+		Kind          string       `json:"kind"  unpack:""`
+		Path          string       `json:"path"`
+		Format        string       `json:"format"`
+		Fields        []field.Path `json:"fields"`
+		Filter        Expr         `json:"filter"`
+		ZoneMapPruner Expr         `json:"zone_map_pruner"`
```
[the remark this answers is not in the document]
mattnibs: I think calling it `Filter` is a misleading term since it doesn't mean the same thing as how it's used elsewhere. It's a highly augmented version of a queries filter.

compiler/optimizer/pruner.go
```diff
@@ -0,0 +1,231 @@
...
+		return ">"
+	case ">":
+		return "<="
+	case ">=":
+		return "<"
+	}
+	panic("unknown op")
+}
+
+func newZoneMapPruner(pred dag.Expr) dag.Expr {
```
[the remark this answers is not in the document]
mattnibs: Min/Max is too limiting because at least at some point you are going to want to put bloom filters here. Zone maps are what these are typically referred to as but we can go with newMetadataPruner for now.

vng/metadata.go
mattnibs: I mean they are [zone maps](https://docs.oracle.com/en/database/oracle/oracle-database/21/dwhsg/using-zone-maps.html) but I'm ok with `MetadataValues`

zfmt/dag.go
```diff
@@ -548,6 +548,17 @@ func (c *canonDAG) op(p dag.Op) {
 		if len(p.Fields) > 0 {
 			c.fields(p.Fields)
 		}
+		if p.MetadataPruner != nil {
+			c.ret()
+			c.open()
+			c.open(" zonemap (")
```
nwt, line 554:
```suggestion
			c.open(" pruner (")
```

zio/vngio/vectorreader.go
```diff
@@ -65,6 +67,24 @@ func (v *VectorReader) Pull(done bool) (vector.Any, error) {
...
+		return objects
+	}
+	return slices.DeleteFunc(objects, func(o *vng.Object) bool {
+		return pruneObject(zctx, pruner, o.Metadata())
+	})
+}
+
+func pruneObject(zctx *super.Context, pruner expr.Evaluator, m vng.Metadata) bool {
+	for _, val := range vng.MetadataValues(zctx, m) {
+		if out := pruner.Eval(nil, val); out.AsBool() {
```
nwt, line 81:
```suggestion
		if pruner.Eval(nil, val).AsBool() {
```

vng/metadata.go
```diff
@@ -265,6 +266,68 @@ func (d *Dynamic) Len() uint32 {
...
+			values = append(values, super.NewValue(typ, b.Bytes().Body()))
+		}
+	} else {
+		typ := zoneMapValue(zctx, &b, m)
+		values = append(values, super.NewValue(typ, b.Bytes().Body()))
+	}
+	return values
+}
+
+func zoneMapValue(zctx *super.Context, b *zcode.Builder, m Metadata) super.Type {
```
nwt, line 285:
```suggestion
func metadataValue(zctx *super.Context, b *zcode.Builder, m Metadata) super.Type {
```

vng/metadata.go
```diff
@@ -265,6 +266,68 @@ func (d *Dynamic) Len() uint32 {
...
+		return zoneMapLeaf(zctx, b, super.NewInt(m.Typ, m.Min), super.NewInt(m.Typ, m.Max))
+	case *Uint:
+		return zoneMapLeaf(zctx, b, super.NewUint(m.Typ, m.Min), super.NewUint(m.Typ, m.Max))
+	case *Const:
+		return zoneMapLeaf(zctx, b, m.Value, m.Value)
+	}
+	panic(m)
+}
+
+func zoneMapLeaf(zctx *super.Context, b *zcode.Builder, min, max super.Value) super.Type {
```
nwt, line 320:
```suggestion
func metadataLeaf(zctx *super.Context, b *zcode.Builder, min, max super.Value) super.Type {
```

compiler/ztests/zonemap.yaml
```diff
@@ -0,0 +1,12 @@
+script: |
```
nwt, line 1: Nit: Rename to pruner.yaml

nwt (approved):

PR #5669, opened 2025-02-20
add proper SQL scoping and table alias resolution

mccanne:
This commit implements proper SQL scoping among expressions in the various clauses of SQL select expressions etc.  It works by mapping the SQL relation concept to various runtime record patterns, e.g., for select wrapping data in a record {in:...,out:...} so that order by can reach into the select scope and order elements either by the input relation or the selected columns.  Solving this in general proved a bit tricky but solution is pretty straightforward in the end.  We do not yet support lateral joins or correlated subqueries.

With these changes, where clauses now operate on both the input and column aliases, any column expression can refer to any aliases defined to its left, group-by expression matching works, joins are spread into a joined record, order-by can reach into the input data or selected data, having clauses can include aggregate functions, and column aggregates can be composed of multiple aggregate functions in an expression.

There is a new resolve function in the semantic analyzer that implements rules for mapping dotted paths to expression elements based on precedence, e.g., for x.y a known column x is chosen before a table expression with alias x.  This is complicated when data types are not known in advance. A later PR will add type propagation and methods for obtaining type information (instead of schemas) from sources.  This will allow more precise SQL compatibility.


compiler/semantic/sql.go
```diff
@@ -16,61 +16,292 @@ import (
...
-			}
+		seq = a.genYield(proj, where, sch, seq)
+		schOut = sch
+		if sel.Distinct {
+			seq = a.genDistinct(pathOf("out"), seq)
 		}
 	}
-	if sel.Value {
-		return a.semSelectValue(sel, seq)
+	return seq, schOut
```
nwt, line 69: Nit: Indent less.
```go
	if len(funcs) != 0 || len(keyExprs) != 0 {
		seq = a.genSummarize(sel.Loc, proj, where, keyExprs, funcs, having, seq)
		return seq, sch.out
	}
	if having != nil {
		a.error(sel.Having, errors.New("HAVING clause requires aggregations and/or a GROUP BY clause"))
	}
	seq = a.genYield(proj, where, sch, seq)
	if sel.Distinct {
		seq = a.genDistinct(pathOf("out"), seq)
	}
	return seq, sch
```

compiler/semantic/sql.go
```diff
@@ -16,61 +16,292 @@ import (
...
 	}
-	if sel.GroupBy != nil {
-		if proj.hasStar() {
-			a.error(sel, errors.New("aggregate mixed with *-selector not yet supported"))
-			return append(seq, badOp())
+	return seq
+}
+
+func (a *analyzer) genColumns(proj projection, sch *schemaSelect, seq dag.Seq) dag.Seq {
+	notFirst := false
```
nwt, line 91: Nit: Use `var` for the zero value.
```suggestion
	var notFirst bool
```

compiler/semantic/sql.go
```diff
@@ -16,61 +16,292 @@ import (
...
+
+func (a *analyzer) genSummarize(loc ast.Loc, proj projection, where dag.Expr, keyExprs []exprloc, funcs aggfuncs, having dag.Expr, seq dag.Seq) dag.Seq {
+	if proj.hasStar() {
+		// XXX take this out and figure out to incorporate this especially if we know the input schema
+		a.error(loc, errors.New("aggregate mixed with *-selector not yet supported"))
+		return append(seq, badOp())
+	}
+	if len(proj) != len(proj.aggCols()) {
+		// Yield expressions for potentially left-to-right-dependent
+		// columns expressions of the group-by expression components.
```
nwt, line 165:
```suggestion
		// column expressions of the group-by expression components.
```

compiler/semantic/sql.go
```diff
@@ -16,61 +16,292 @@ import (
...
+}
+
+func (a *analyzer) semSelectFrom(loc ast.Loc, from *ast.From, seq dag.Seq) (dag.Seq, schema) {
+	if from == nil {
+		return seq, &schemaDynamic{}
+	}
+	off := len(seq)
+	hasParent := off > 0
+	var sch schema
+	seq, sch = a.semFrom(from, seq)
```
nwt, line 287: Nit:
```suggestion
	seq, sch := a.semFrom(from, seq)
```

compiler/semantic/sql.go
```diff
@@ -16,61 +16,292 @@ import (
...
+// an OrderBy, it can reach into the "in" part of the select scope (for non-aggregates)
+// and also sort by the out elements.  It's up to the caller to unwrap the in/out
+// record when returning to pipeline context.
+func (a *analyzer) semSelect(sel *ast.Select, seq dag.Seq) (dag.Seq, schema) {
+	if len(sel.Selection.Args) == 0 {
+		a.error(sel, errors.New("SELECT clause has no selection"))
+		return seq, badSchema()
+	}
+	var fromSchema schema
+	seq, fromSchema = a.semSelectFrom(sel.Loc, sel.From, seq)
```
nwt, line 31: Nit:
```suggestion
	seq, fromSchema := a.semSelectFrom(sel.Loc, sel.From, seq)
```

compiler/semantic/sql.go
```diff
@@ -16,61 +16,292 @@ import (
...
+	}
+	sch := &schemaSelect{in: fromSchema}
+	var funcs aggfuncs
+	proj := a.semProjection(sch, sel.Selection.Args, &funcs)
+	var where dag.Expr
+	if sel.Where != nil {
+		where = a.semExprSchema(sch, sel.Where)
+	}
+	keyExprs := a.semGroupBy(sch, sel.GroupBy)
+	having, err := a.semHaving(sch, sel.Having, &funcs)
```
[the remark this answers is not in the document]
mccanne: I want to leave this one as is for now.

nwt (approved):
:shipit:

chrismo:
browsing all of the test .yaml files - very cool stuff, need to find some time to play with all this more. :)

PR #5678, opened 2025-02-25
De-Zed the Data Model doc

philrz:
## tl;dr

Recent adjustments to the [superdb.org](https://superdb.org/) web site led me to link to the Data Model page, which made me confront that it's ready for the de-Zed-ification I've done here. I did a bit of other clean-up along the way.

## Details

I intentionally avoided ever saying "SuperDB", since even though we only be the full implementation of this stuff, I assume we want things like the Data Model to read as independent from any single implementation.

### DIFF
--- a/docs/formats/data-model.md
+++ b/docs/formats/data-model.md
@@ -102,7 +102,7 @@ is distinct from type `{b:string,a:string}`.
 
 A field name is any UTF-8 string.
 
-A field value is any value of any type.
+A field value is a value of any type.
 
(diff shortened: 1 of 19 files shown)


docs/formats/data-model.md
```diff
@@ -98,9 +102,9 @@ is distinct from type `{b:string,a:string}`.
 
 A field name is any UTF-8 string.
 
-A field value is any Zed value.
+A field value is any value of any type.
```
mccanne, line 105: is any value -> is a value

mccanne (approved):

PR #5694, opened 2025-03-06
Add support for SQL Timestamp syntax

mattnibs:


compiler/parser/parser.peg
```diff
@@ -1293,6 +1294,15 @@ Tuple
       }, nil
     }
 
+SQLTimestamp
+  = TIMESTAMP _ s:StringLiteral {
+        return &ast.SQLTimestamp {
+          Kind: "SQLTimestamp",
+          Value: s.(*ast.Primitive),
```
nwt, line 1301: Some mgbench queries also use SQL DATE syntax. How about handling that, too?

```suggestion
SQLDateTime
  = typ:(DATE / TIMESTAMP) _ value:StringLiteral {
        return &ast.SQLTimestamp {
          Kind: "SQLTimestamp",
          Type: typ,
          Value: value.(*ast.Primitive),
```

nwt: If we do this, I don't think there's any need to constrain DATE values to exclude a time (since there's no problem for now with accepting the larger language).

mattnibs: If we change the name to ast.SQLTimeValue I think we can go this route. Postgres does indeed allow you to parse timestamp values as date however if you have a timestamp string w/ non-zero hh:mm:ss I believe you need to have logic to zero those out if we are going to be consistent w/ date casting in https://github.com/brimdata/super/pull/5695

nwt: I'm in favor of doing this, with rounding for DATE in semantic.analyzer.semExpr.

compiler/ztests/sql-timeval.yaml
```diff
@@ -0,0 +1,10 @@
+zed: "yield TIMESTAMP '2012-10-02 12:01:01', DATE '2012-10-02 12:01:01'"
```
nwt, line 1: Nit: No need to quote this.
```suggestion
zed: yield TIMESTAMP '2012-10-02 12:01:01', DATE '2012-10-02 12:01:01'
```

nwt (approved):

PR #5695, opened 2025-03-06
Support SQL type cast expressions

mattnibs:
Also includes a hack to support casts to dates but adding bucket(this,1d) to a time cast.

### DIFF
--- a/compiler/ast/sql.go
+++ b/compiler/ast/sql.go
@@ -122,3 +122,12 @@ type AsExpr struct {
 }
 
 func (*AsExpr) ExprAST() {}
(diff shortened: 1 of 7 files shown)


compiler/ast/ast.go
```diff
@@ -155,6 +155,13 @@ type SliceExpr struct {
 	Loc  `json:"loc"`
 }
 
+type SQLCast struct {
```
nwt, line 158: I'd put this in compiler/ast/sql.go.

nwt (approved):

PR #5697, opened 2025-03-06
Support regexp prefix for metadata pruner

mattnibs:


compiler/optimizer/pruner.go
```diff
@@ -145,10 +149,40 @@ func reverseComparator(op string) string {
...
+			compare(">=", end, &dag.This{Kind: "This", Path: append(this.Path, "max")}))
+	default:
 		return nil
 	}
+}
+
+func incrementLastRune(s string) string {
+	b := []byte(s)
+	r, size := utf8.DecodeLastRune(b)
+	return string(utf8.AppendRune(b[:len(b)-size], r+1))
```
mattnibs, line 182: I need to verify this doesn't result in an invalid utf8 value which I think it probably can.

mattnibs: Or maybe this doesn't matter for comparison purposes?

mattnibs:
Closing this to revaluate how to do regexp prefix ranges.

PR #5699, opened 2025-03-07
Fix multiple close calls on concurrent csup readers

### LINKED ISSUE #5644: Attempted sort of CSUP query result in vector runtime: "file already closed" error
## tl;dr

With test data `data.jsup`:

```
{
    URL: "https://www.google.com",
    EventTime: 2013-07-14T20:08:25Z (=arrow_timestamp_us)
}
```

This fails in vector runtime:

```
$ super -f csup -o data.csup data.jsup && SUPER_VAM=1 super -c "SELECT * FROM 'data.csup' WHERE URL LIKE '%google%' ORDER BY EventTime"
close /Users/phil/work/newbug/data.csup: file already closed
```

## Details

Repro is with super commit 203b0de. The query and data are a simplification of the [ClickBench q23 query](https://github.com/ClickHouse/ClickBench/blob/acdd156c11043d45e9e54f791747bf51e9564d06/duckdb/queries.sql#L24).

It runs fine with CSUP in sequential runtime.

```
$ super -version
Version: v1.18.0-272-g203b0de2

$ super -f csup -o data.csup data.jsup && super -c "SELECT * FROM 'data.csup' WHERE URL LIKE '%google%' ORDER BY EventTime"
{URL:"https://www.google.com",EventTime:2013-07-14T20:08:25Z(=arrow_timestamp_us)}
```

It also runs fine in vector runtime if I use Parquet instead of CSUP.

```
$ super -f parquet -o data.parquet data.jsup && SUPER_VAM=1 super -c "SELECT * FROM 'data.parquet' WHERE URL LIKE '%google%' ORDER BY EventTime"
{URL:"https://www.google.com",EventTime:2013-07-14T20:08:25Z}
```

And it runs fine with CSUP in vector runtime if I drop the `ORDER BY`.

```
$ super -f csup -o data.csup data.jsup && SUPER_VAM=1 super -c "SELECT * FROM 'data.csup' WHERE URL LIKE '%google%'"
{URL:"https://www.google.com",EventTime:2013-07-14T20:08:25Z(=arrow_timestamp_us)}
```

mattnibs:
Closes #5644

### DIFF
--- a/zio/vngio/vectorreader.go
+++ b/zio/vngio/vectorreader.go
@@ -24,6 +24,7 @@ type VectorReader struct {
 	objects       []*vng.Object
 	projection    vcache.Path
 	readerAt      io.ReaderAt
+	hasClosed     bool
 }
 
 func NewVectorReader(ctx context.Context, zctx *super.Context, r io.Reader, fields []field.Path, pruner expr.Evaluator) (*VectorReader, error) {
@@ -48,7 +49,15 @@ func NewVectorReader(ctx context.Context, zctx *super.Context, r io.Reader, fiel
 
 func (v *VectorReader) NewConcurrentPuller() vector.Puller {
 	v.activeReaders.Add(1)
-	return v
+	return &VectorReader{
+		ctx:           v.ctx,
+		zctx:          v.zctx,
+		activeReaders: v.activeReaders,
+		nextObject:    v.nextObject,
+		objects:       v.objects,
+		projection:    v.projection,
+		readerAt:      v.readerAt,
+	}
 }
 
 func (v *VectorReader) Pull(done bool) (vector.Any, error) {
@@ -86,7 +95,8 @@ func pruneObject(zctx *super.Context, pruner expr.Evaluator, m vng.Metadata) boo
 }
 
 func (v *VectorReader) close() error {
-	if v.activeReaders.Add(-1) <= 0 {
+	if !v.hasClosed && v.activeReaders.Add(-1) <= 0 {
+		v.hasClosed = true
 		if closer, ok := v.readerAt.(io.Closer); ok {
 			return closer.Close() // coffee is for closers
 		}


zio/vngio/vectorreader.go
```diff
@@ -86,10 +97,13 @@ func pruneObject(zctx *super.Context, pruner expr.Evaluator, m vng.Metadata) boo
 }
 
 func (v *VectorReader) close() error {
-	if v.activeReaders.Add(-1) <= 0 {
-		if closer, ok := v.readerAt.(io.Closer); ok {
-			return closer.Close() // coffee is for closers
+	var err error
+	v.closeOnce.Do(func() {
+		if v.activeReaders.Add(-1) <= 0 {
```
nwt, line 102: This isn't correct because v.activeReaders.Add can only be called once now.

mattnibs: I create a new sync.Once for every ConcurrentPuller and so it will be called once per thread. Am I missing something?

nwt: Right you are. Missed that.

zio/vngio/vectorreader.go
```diff
@@ -48,7 +50,16 @@ func NewVectorReader(ctx context.Context, zctx *super.Context, r io.Reader, fiel
...
-	return v
+	return &VectorReader{
+		ctx:           v.ctx,
+		zctx:          v.zctx,
+		activeReaders: v.activeReaders,
+		nextObject:    v.nextObject,
+		objects:       v.objects,
+		projection:    v.projection,
+		readerAt:      v.readerAt,
+		closeOnce:     sync.Once{},
```
nwt, line 61: Nit: No need to initialize this.
```suggestion
```

nwt (approved):

zio/vngio/vectorreader.go
nwt: But then it can just be a `bool` since a VectorReader isn't safe to share across goroutines.

mattnibs: Right you are I change

PR #5703, opened 2025-03-11
Add filter pushdown to vector Parquet reader

nwt:


zio/parquetio/rowgroupfilter.go
```diff
@@ -0,0 +1,196 @@
...
+		return 1, true
+	case []byte:
+		if id := aVal.Type().ID(); id != super.IDBytes && id != super.IDString {
+			return 0, false
+		}
+		a := aVal.Bytes()
+		n := min(len(a), len(b))
+		return bytes.Compare(a[:n], b[:n]), true
+	case float64:
+		a, ok := coerce.ToFloat(aVal, super.TypeFloat64)
```
mattnibs, line 147: This and the other coerce functions here are going to coerce string values to floats which I don't think we want to do here.

nwt: I know but don't think that matters because the `where` operator containing this filter will drop those values.

mattnibs (approved):

PR #5706, opened 2025-03-14
Optimize by merging adjacent yield operators

nwt:
Optimize queries by merging adjacent yield operators when the first yields one record expression so that, e.g., "yield {a:b} | yield {c:a}" becomes "yield {c:b}".

### DIFF
--- a/compiler/optimizer/optimizer.go
+++ b/compiler/optimizer/optimizer.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"reflect"
+	"slices"
 
 	"github.com/brimdata/super/compiler/dag"
 	"github.com/brimdata/super/compiler/optimizer/demand"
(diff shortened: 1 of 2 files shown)


mattnibs:
Not sure if this is something you want to address in this pr. But if I do:
```SELECT machine_name,
       log_time
FROM '__SOURCE__'
WHERE (machine_name LIKE 'cslab%' OR
       machine_name LIKE 'mslab%')
  AND load_one IS NULL
  AND log_time >= TIMESTAMP '2017-01-10 00:00:00'
ORDER BY machine_name,
         log_time;```

i get:
```
file __SOURCE__
| yield {in:this,out:{...{machine_name:machine_name},log_time:log_time}}
| where (regexp_search(/(?s)^cslab.*?$/, out.machine_name) or regexp_search(/(?s)^mslab.*?$/, out.machine_name)) and in.load_oneIS NULL and out.log_time>=2017-01-10T00:00:00Z
| sort out.machine_name asc, out.log_time asc
| yield out
| output main
```

Seems like on the opening yield there should be the spread operator on `machine_name`, no?

nwt:
@mattnibs: Planning to address that in a separate PR.

mattnibs (approved):

PR #5716, opened 2025-03-18
Support pruning on regular expression prefixes

mattnibs:


compiler/optimizer/pruner.go
```diff
@@ -145,10 +149,61 @@ func reverseComparator(op string) string {
...
+		prefix := regexpPrefix(e.Pattern)
+		maxPrefix := regexpMaxPrefix(prefix)
+		if prefix == "" || maxPrefix == "" {
+			return nil
+		}
+		min := &dag.Literal{Kind: "Literal", Value: zson.QuotedString(prefix)}
+		max := &dag.Literal{Kind: "Literal", Value: zson.QuotedString(maxPrefix)}
+		return dag.NewBinaryExpr("and",
+			compare("<", min, &dag.This{Kind: "This", Path: append(this.Path, "min")}),
+			compare(">=", max, &dag.This{Kind: "This", Path: append(this.Path, "max")}))
```
nwt, line 169: If len(this.Path) < cap(this.Path), then both calls to append here will write to the same underlying array with the second call overwriting "min" with "max".
```suggestion
			compare("<", min, &dag.This{Kind: "This", Path: append(slices.Clone(this.Path), "min")}),
			compare(">=", max, &dag.This{Kind: "This", Path: append(slices.Clone(this.Path), "max")}))
```

nwt (approved):

PR #5717, opened 2025-03-19
Morph "super tutorial" doc into "For jq Users"

philrz:
## What's Changing

As a de-Zed-ing exercise, in this PR I propose morphing the [Super Tutorial](https://superdb.org/docs/tutorials/zq/) doc into a doc titled "For jq Users". Most of the original content is maintained, but I attempt to adjust its approach slightly to nod toward the SuperDB project's current lead-with-SuperSQL approach.

To review by reading the doc in its final form (which I recommend) rather than as diffs, I've put a copy of the rendered doc in a personal staging site: https://capable-khapse-fa9d06.netlify.app/docs/tutorials/super-for-jq/

## Details

Back when this doc was originally the "`zq` Tutorial" our target audience largely consisted of users of `jq` and similar command-line tools. In the current era of SuperSQL that's no longer the case, and yet everything about the data model, pipeline queries, and language shortcuts that first made the doc compelling are still present in the SuperDB project. We'd probably not _lead_ with this doc today--instead we'd be more likely to prominently point to the [`super` command doc](https://superdb.org/docs/commands/super/) that provides an introduction more tailored to a SQL-centric audience. But considering the original comparison to `jq` accumulated [460+ upvotes on HackerNews](https://news.ycombinator.com/item?id=31166956) and a [corresponding quick bump in ~500 GitHub stars](https://www.star-history.com/#brimdata/super&Date) back in April of 2022, it feels like the text is worth keeping in some form, either for a secondary audience of users that find SuperDB after having had a `jq`-centric background _or_ a SQL-centric user that quickly catches on that SuperSQL's approach provides the option to avoid static schemas, tables, and SQL itself if that's their preference.

### DIFF
--- a/docs/tutorials/super-for-jq.md
+++ b/docs/tutorials/super-for-jq.md
@@ -3,7 +3,7 @@ weight: 1
 title: For jq Users
 ---
 
-SuperSQL’s pipes and shortcuts provide a flexible and powerful way for SQL
+[SuperSQL](../language/_index.md)’s pipes and shortcuts provide a flexible and powerful way for SQL
 users to query their JSON data while leveraging their existing skills.
 However, users who've traditionally wrangled their JSON with other tools such
 as [`jq`](https://stedolan.github.io/jq/) will find `super` equally powerful


docs/tutorials/super-for-jq.md
```diff
@@ -1,46 +1,59 @@
...
+it can serve you well as a handy, go-to tool.  In this way, `super` is kind of
 like `jq`.  As you probably know, `jq`
 is a popular command-line tool for taking a sequence of JSON values as input,
 doing interesting things on that input, and emitting results, of course, as JSON.
 
 `jq` is awesome and powerful, but its syntax and computational model can
-sometimes be daunting and difficult.  We tried to make `zq` really easy and intuitive,
-and it is usually faster, sometimes [much faster](../commands/super.md#performance),
+sometimes be daunting and difficult.  We tried to make `super` really easy and intuitive,
+and it is usually faster, sometimes [much faster](https://github.com/brimdata/super/tree/main/performance),
```
philrz, line 52: The ["performance" section of the `super` command doc](https://superdb.org/docs/commands/super/#performance) this was previously linking to has already (appropriately!) been morphed into a perf comparison vs. SQL-centric solutions, not `jq`. Therefore I'm instead pointing to the place in the repo that still has some `jq` perf comparisons. That said, I also recognize that same table of perf results includes `zeek-cut` stuff that's probably due to be dropped. If this PR passes muster, as a follow-on task I'll overhaul that perf table and the underlying script that generates the results.

docs/tutorials/super-for-jq.md
```diff
@@ -449,7 +470,7 @@ But more powerfully, types can be used anywhere a value can be used and
 in particular, they can be group-by keys, e.g.,
 ```mdtest-command
 echo '{x:1,y:2}{s:"foo"}{x:3,y:4}' |
-  super -f table -c "count() by \`shape\`:=typeof(this) | sort count" -
+  super -f table -c "count() by this['shape']:=typeof(this) | sort count" -
```
philrz, line 473: I felt like showing backticks while having to escape them to prevent interference by the shell was a little esoteric, so here I opted to instead show the field reference with `this['...']` which still feels a little esoteric for a tutorial doc but a little less so IMO. I'm open to debate though. 😄

docs/tutorials/super-for-jq.md
```diff
@@ -579,20 +600,34 @@ To appreciate all this, let's have a look next at some real-world data...
...
+As we get into the exercise below, we'll reach a step where we encounter some
+unexpected empty objects in the original data. It seems the GitHub API
+must have been having a bad day when we first ran this exercise, as these
+empty records no longer appear if the download is repeated today using the same
+URL shown above. But taming glitchy data is a big part of data discovery, so to
+relive the magic of our original experience, you can download
+[this archived copy](https://superdb.org/docs/tutorials/prs.json) of the
+`prs.json` we originally saw.
+
+{{% /tip %}}
```
philrz, line 627: It's always bothered me a little that the steps in the doc were not strictly repeatable if the reader followed our guidance about downloading the GitHub data anew. Rather than risk confusion, I figured we might as well just disclose.

mccanne (approved):
"SuperSQL’s pipeline extensions and shortcuts" -> SuperDB's Pipe SQL and SQL shortcuts

"SQL lovers" -> SQL users

PR #5718, opened 2025-03-19
Merge adjacent yield operators containing bare this

nwt:
The optimization that merges adjacent yield operators does not alter "yield {a} | yield {b:this}" because the second yield operator contains a bare reference to "this".  Remove that restriction so it will merge the operators to become "yield {b:{a}}".


nwt:
I don't think we need this.

PR #5726, opened 2025-03-20
vam: Support shape expressions using samexpr

mattnibs:


compiler/kernel/vexpr.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/brimdata/super/compiler/dag"
 	"github.com/brimdata/super/pkg/field"
 	"github.com/brimdata/super/runtime/sam/expr"
+	samexpr "github.com/brimdata/super/runtime/sam/expr"
```
nwt, line 11: This package is already imported as expr by the previous line.

mattnibs: Right

nwt (approved):

PR #5727, opened 2025-03-21
Enable vector tests on working sam exprs tests

mattnibs:


mattnibs:
Also removed a few redundant tests.

runtime/ztests/expr/function/split.yaml
```diff
@@ -5,6 +5,7 @@ vector: true
 input: |
   {s:"a,b,c",sep:","((int64,string))}
   {s:"abc",sep:""}
+  {s:"a",sep:","}
```
nwt, line 8: runtime/sam/expr/ztests/split.yaml also includes an empty-string case so I think we should add that here.
```suggestion
  {s:"",sep:","}
```

runtime/sam/expr/ztests/time-duration-cast.yaml
```diff
@@ -1,11 +0,0 @@
-zed: cut t:=time(t),d:=duration(d)
-
-input: |
-  {t:"1",d:"1"}
-  {t:1,d:1}
```
nwt, line 5: runtime/ztests/expr/cast/duration.yaml doesn't include this case (casting an integer) so I think we should add it over there.

nwt (approved):

PR #5744, opened 2025-03-27
vam: Fix nested cuts / puts

### LINKED ISSUE #5741: Cut on multiple nested fields in vector runtime
There's a bug in vector runtime when doing a cut on multiple nested fields sharing the same root path:
```
$ echo {rec:{a:1,b:2}} | SUPER_VAM=1 super -c 'cut rec.a,rec.b' -
```

This should return the original record but it instead returns `{rec:{b:2}}`

mattnibs:
Closes #5741

### DIFF
--- a/compiler/kernel/vop.go
+++ b/compiler/kernel/vop.go
@@ -3,6 +3,7 @@ package kernel
 import (
 	"errors"
 	"fmt"
+	"slices"
 
 	"github.com/brimdata/super"
 	"github.com/brimdata/super/compiler/dag"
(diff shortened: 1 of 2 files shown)


compiler/kernel/vop.go
```diff
@@ -302,27 +303,31 @@ func (b *Builder) compileVamLeaf(o dag.Op, parent vector.Puller) (vector.Puller,
...
+		addPathToRecord(rec, lhs.Path, a.RHS)
 	}
-	return b.compileVamRecordExpr(&dag.RecordExpr{Kind: "RecordExpr", Elems: elems})
+	return b.compileVamRecordExpr(rec)
 }
 
-func newDagRecordExprForPath(path []string, expr dag.Expr) *dag.RecordExpr {
-	if len(path) > 1 {
-		expr = newDagRecordExprForPath(path[1:], expr)
+func addPathToRecord(rec *dag.RecordExpr, path []string, expr dag.Expr) {
```
nwt, line 317:
```suggestion
func addPathToRecordExpr(rec *dag.RecordExpr, path []string, expr dag.Expr) {
```

nwt (approved):
:shipit:

PR #5745, opened 2025-03-28
Remove mentions of "SuperPipe" from docs, plus other fixes

philrz:
This PR removes from the docs all mentions of the name "SuperPipe" that was briefly used to refer to the query language's use of pipes.

As I gave a thorough read through each affected doc, I made other changes along the way, e.g., fixing typos, adding/changing hyperlinks, de-Zed-ing, and a few material changes I'll point out with in-line comments.

### DIFF
--- a/docs/language/data-types.md
+++ b/docs/language/data-types.md
@@ -3,7 +3,7 @@ weight: 3
 title: Data Types
 ---
 
-The SuperPipe language includes most data types of a typical programming language
+The SuperSQL language includes most data types of a typical programming language
 as defined in the [super data model](../formats/data-model.md).
 
 The syntax of individual literal values generally follows
@@ -21,19 +21,19 @@ constant values like Super JSON and can be composed from [literal expressions](e
 
 ## First-class Types
 
-As in the super data model, the SuperPipe language has first-class types:
+As in the super data model, the SuperSQL language has first-class types:
 any type may be used as a value.
 
 The primitive types are listed in the
 [data model specification](../formats/data-model.md#1-primitive-types)
-and have the same syntax in SuperPipe.  Complex types also follow
+and have the same syntax in SuperSQL.  Complex types also follow
 the Super JSON syntax.  Note that the type of a type value is simply `type`.
 
 As in Super JSON, _when types are used as values_, e.g., in an expression,
 they must be referenced within angle brackets.  That is, the integer type
 `int64` is expressed as a type value using the syntax `<int64>`.
 
-Complex types in the SuperPipe language follow the Super JSON syntax as well.  Here are
+Complex types in the SuperSQL language follow the Super JSON syntax as well.  Here are
 a few examples:
 * a simple record type - `{x:int64,y:int64}`
 * an array of integers - `[int64]`
@@ -108,7 +108,7 @@ type socket = {addr:ip,port:port=uint16}
 defines a named type `socket` that is a record with field `addr` of type `ip`
 and field `port` of type "port", where type "port" is a named type for type `uint16` .
 
-Named types may also be defined by the input data itself, as super data is
+Named types may also be defined by the input data itself, as super-structured data is
 comprehensively self describing.
 When named types are defined in the input data, there is no need to declare their
 type in a query.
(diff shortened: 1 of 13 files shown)


docs/language/data-types.md
```diff
@@ -108,7 +108,7 @@ type socket = {addr:ip,port:port=uint16}
 defines a named type `socket` that is a record with field `addr` of type `ip`
 and field `port` of type "port", where type "port" is a named type for type `uint16` .
 
-Named types may also be defined by the input data itself, as super data is
+Named types may also be defined by the input data itself, as super-structured data is
```
philrz, line 111: I'm making all attempts to avoid the use of "super data" as a standalone adjective+noun pairing. On a case by case basis, I've been trying to replace with "super-structured data" or some reference to the "super data model".

nwt: I thought we were going with polymorphic instead of super-structured.

docs/language/expressions.md
```diff
@@ -81,6 +78,22 @@ For example,
...
+# input
+{s:"foo"}
+{s:"foo",t:"bar"}
+{a:{s:"foo"}}
+[1,{s:"foo"},2]
+# expected output
+{s:"foo"}
+{a:{s:"foo"}}
+[1,{s:"foo"},2]
+```
```
philrz, line 95: I broke out this separate example because the "Search Expressions" doc (that's also touched in this PR) makes a reference to using `in this` with complex values and so I wanted to link to this section of the "Expressions doc", but that's when I noticed we didn't show an example of it here.

docs/language/operators/over.md
```diff
@@ -22,8 +22,8 @@ generated from each such result depending on its types:
 entry in the map, and
 * all other values generate a single value equal to itself.
 
-Records can be converted to maps with the [`flatten` function](../functions/flatten.md)
-resulting in a map that can be traversed,
+A record can be converted with the [`flatten` function](../functions/flatten.md) to
+an array that can be traversed,
```
philrz, line 26: The word "map" seemed inappropriate here, so I reworded. I think I see the spirit of what was being said before because the breaking out of the separately-named `key` and `value` portions allow for treating these in map-like ways, but since we already have an official `map` type (not to mention the `map` function), I figured it'd be good to avoid lumping on yet another use of the word "map".

docs/language/pipeline-model.md
```diff
@@ -19,11 +19,11 @@ refers to a collection of data called a "data pool" and is referenced using
 the pool's name much as SQL references database tables by their name.
 
 For more detail, see the reference page of the [`from` operator](operators/from.md),
-but as an example, you might use the `get` form of `from` to fetch data from an
+but as an example, you might use its `get` form to fetch data from an
 HTTP endpoint and process it with `super`, in this case, to extract the description
 and license of a GitHub repository:
 ```
-super -f text -c 'get https://api.github.com/repos/brimdata/super
+super -f line -c 'get https://api.github.com/repos/brimdata/super
```
philrz, line 26: Mode `line` gives the same output as `text` in this example, so I opted to flip it because we sometimes seem close to dumping `text` as an output format.

docs/language/pipeline-model.md
```diff
@@ -241,14 +240,14 @@ is unambiguously an aggregation and can be shortened to
 ```
 count() by id
 ```
-Likewise, a very common lean-forward use pattern is "searching" so by default,
-expressions are interpreted as keyword searches, e.g.,
+Likewise, a very common lean-forward use pattern is "searching", so with the
+use of leading `?` shorthand, expressions are interpreted as keyword searches, e.g.,
```
philrz, line 244: The text here definitely needed to be changed (#5436). I'm less certain if we should stop mentioning `search` entirely in this section on Implied Operators. What I've done instead is still mention it but alongside explaining the `?` shorthand. From a language parsing perspective I respect that this makes it different from all the genuine "implied operators" since the `?` is still explicit. But I feel we could also justify mentioning it here in the spirit of "ways you can express your query with less typing", plus it contrasts nicely with `where`. I'm open to debate though!

docs/language/data-types.md
mccanne: I think polymorphic is good when explaining the theoretical foundations when talking to a more academic audience but maybe we should avoid it as the headline.  Super-structured is good when comparing/contrasting with semi-structured and structured data.  "Super data model" is good for the data model and "super data" I'm actually ok with as a generic terms to mean any data that conforms to the super data model, e.g., there is super data inside of a CSUP file,  the super.Value Go type represents super data, super data is a superset of relational and JSON data.  So I guess here I would just leave it as "super data" but super-structured is ok too.

mccanne (approved):

docs/language/data-types.md
nwt: 👍

nwt (approved):
This looks fine to me.

PR #5748, opened 2025-03-30
remove super.Mapper from flattener

mccanne:
This commit changes the use of super.Mapper in the flattener to a simple hash table.  Mapper has unnecessary overhead for this use case (for example, computing type values to translate between contexts when this code used same context as the output) and a simple map implementation suffices here.  Also, we plan to change the super.Mapper API in a subsequent PR and its use here would otherwise complicate that change.

### DIFF
--- a/runtime/sam/expr/flattener.go
+++ b/runtime/sam/expr/flattener.go
@@ -8,8 +8,8 @@ import (
 )
 
 type Flattener struct {
-	zctx   *super.Context
-	mapper *super.Mapper
+	zctx *super.Context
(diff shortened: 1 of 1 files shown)


nwt (approved):
The only knock on this I see is that the map access here will be observably slower than the slice access it replaces, but that doesn't really matter given our focus on the vector runtime.

mccanne:
> The only knock on this I see is that the map access here will be observably slower than the slice access it replaces, but that doesn't really matter given our focus on the vector runtime.

It's only used by the writers that need to flatten stuff...

PR #5749, opened 2025-03-31
lighter weight type context management

mccanne:
This commit makes a big performance improvement for type context management by removing the need to serialize and store type values for every type in the context.  This was especially problematic for large (wide and deep) types where the system would reserialize every node in the type hierarchy from bottom to top so there was a quadratic effect going on.  We never noticed this for small types but for the large, nested values of CSUP metadata the problem became painfully obvious.

Part of the fix is to remove super.Mapper, which was based on mapping types by serialized type value, and replace it with a direct mapper in the BSUP reader where local type IDs are mapped directly to the shared super.Context.  As part of this change, we cleaned up the mapper cache by creating the super.TypeFetcher interface and arraging for super.Context, zngio.Decoder, and super.TypeCache to all implement this interface.  This way the bufferfilter can simply take super.TypeFetcher interface value and not worry that the type mappings are handled in the BSUP decoder.

Types are now interned by using type fingerprints based on local type IDs instead of large, serialized type values.  There is a dedicated map in the super.Context for each complex type category but the maps are allocated only when needed.

These changes mean that all values in the query runtime must be typed from the shared query super.Context.  This was generally the case before, but we found a problem in network_of where error values were using zson.MarshalZNG, which created a temporary super.Context and caused mixed-context values to be created and thus foiled the type interning in the shared context.  This has been fixed by passing the context to the marshaler.

Finally, since we were touching the code, we added some circuit-breaker checks on the number of union types and number of enum symbols in the BSUP reader.


context.go
```diff
@@ -101,19 +117,29 @@ func (d *DuplicateFieldError) Error() string {
 // this type context.  If you want to use fields from a different type context,
 // use TranslateTypeRecord.
 func (c *Context) LookupTypeRecord(fields []Field) (*TypeRecord, error) {
-	tv := tvPool.Get().(*[]byte)
-	*tv = AppendTypeValue((*tv)[:0], &TypeRecord{Fields: fields})
+	key := keyPool.Get().(*[]byte)
+	bytes := (*key)[:0]
+	for _, field := range fields {
+		bytes = binary.LittleEndian.AppendUint32(bytes, uint32(len(field.Name)))
+		bytes = append(bytes, []byte(field.Name)...)
```
nwt, line 124: Nit: No need to convert string to []byte here.
```suggestion
		bytes = append(bytes, field.Name...)
```

context.go
```diff
@@ -155,76 +181,102 @@ func (c *Context) MustLookupTypeRecord(fields []Field) *TypeRecord {
...
 }
 
 func (c *Context) LookupTypeEnum(symbols []string) *TypeEnum {
-	tv := tvPool.Get().(*[]byte)
-	*tv = AppendTypeValue((*tv)[:0], &TypeEnum{Symbols: symbols})
+	key := keyPool.Get().(*[]byte)
+	bytes := (*key)[:0]
+	for _, symbol := range symbols {
+		bytes = binary.LittleEndian.AppendUint32(bytes, uint32(len(symbol)))
+		bytes = append(bytes, []byte(symbol)...)
```
nwt, line 265: Nit:
```suggestion
		bytes = append(bytes, symbol...)
```

context.go
```diff
@@ -246,32 +301,42 @@ func (c *Context) LookupTypeNamed(name string, inner Type) (*TypeNamed, error) {
 	if LookupPrimitive(name) != nil {
 		return nil, fmt.Errorf("bad type name %q: primitive type name", name)
 	}
-	tv := tvPool.Get().(*[]byte)
-	*tv = AppendTypeValue((*tv)[:0], &TypeNamed{Name: name, Type: inner})
+	key := keyPool.Get().(*[]byte)
+	bytes := (*key)[:0]
+	bytes = binary.LittleEndian.AppendUint32(bytes, uint32(len(name)))
+	bytes = append(bytes, []byte(name)...)
```
nwt, line 307: Nit:
```suggestion
	bytes = append(bytes, name...)
```

context.go
```diff
@@ -305,23 +374,20 @@ func (c *Context) TranslateType(ext Type) (Type, error) {
...
 
 func (c *Context) LookupTypeValue(typ Type) Value {
 	c.mu.Lock()
-	bytes, ok := c.toValue[typ]
-	c.mu.Unlock()
-	if ok {
-		return NewValue(TypeType, bytes)
+	if c.toValue != nil {
+		bytes, ok := c.toValue[typ]
+		if ok {
```
nwt, line 385: Nit:
```suggestion
		if bytes, ok := c.toValue[typ]; ok {
```

zio/zngio/parser.go
```diff
@@ -34,7 +34,7 @@ func (p *parser) read() (frame, error) {
 			// and the old mapper and context will continue on just fine as
 			// everything gets properly mappped to the shared context
 			// under concurrent locking within super.Context.
-			p.types.reset()
+			p.types = NewDecoder(p.types.sctx)
```
nwt, line 37: The comment preceding this line could use an update. It refers to local context, shared context, and mappers, but those change largely eliminates those concepts.

nwt (approved):
My comments about updating comments are just nits so :shipit:.

mccanne:
> My comments about updating comments are just nits so :shipit:.

Gee, I somehow missed the important ones!

PR #5750, opened 2025-03-31
vam: Fix bug with puts on nested records

mattnibs:


compiler/kernel/vop.go
```diff
@@ -302,16 +308,16 @@ func (b *Builder) compileVamLeaf(o dag.Op, parent vector.Puller) (vector.Puller,
 	}
 }
 
-func (b *Builder) compileVamAssignmentsToRecordExpression(initial []dag.RecordElem, assignments []dag.Assignment) (vamexpr.Evaluator, error) {
-	rec := &dag.RecordExpr{Kind: "RecordExpr", Elems: initial}
+func vamAssignmentsToRecordExpression(assignments []dag.Assignment) (*dag.RecordExpr, error) {
```
nwt, line 311: Nit: I think namNewRecordExprFromAssignments would be a better name.

nwt (approved):
Nice work!

PR #5758, opened 2025-04-02
change vng to csup in the code

mccanne:


nwt (approved):
Maybe style it CSup or Csup instead of CSUP so it doesn't present as an initialism?

PR #5760, opened 2025-04-02
Manage the mockgen command as a Go tool dependency

nwt:
This moves the mockgen command version specification from the Makefile to go.mod, forcing the same version for the mockgen code and the command.  It also simplifies the Makefile a little.

(Tool dependencies are new in Go 1.24.)


mattnibs (approved):
Cool

PR #5763, opened 2025-04-03
Manage the goimports and pigeon commands as Go tool dependencies

nwt:
This moves the goimports and pigeon command version specifications from compiler/parser/Makefile to go.mod, moves the invocation of those commands to Makefile, and removes compiler/parser/Makefile.

To force rebuilding of the PEG parser, run "make -B peg" instead of "make -C compiler/parser".  (The latter no longer works because compiler/parser/Makefile no longer exists.)


mattnibs (approved):
Lovely!

PR #5773, opened 2025-04-06
move vcache.Path to field.Projection

mccanne:
This commit moves the projection data structure used by vcache into the generic field package so it can be shared in more places without creating import cycles.

This is needed by an upcoming PR that will add projection pushdown to the CSUP metadata filter, which is needed by a subsequent PR that will make CSUP more efficient by allowing lazy loading of types.

### DIFF
--- a/pkg/field/projection.go
+++ b/pkg/field/projection.go
@@ -1,17 +1,15 @@
-package vcache
+package field
 
 import (
 	"fmt"
-
-	"github.com/brimdata/super/pkg/field"
 )
 
-// A path is an array of string or Forks
-type Path []any //XXX clean this up later
-type Fork []Path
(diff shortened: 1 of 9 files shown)


pkg/field/projection.go
```diff
@@ -1,25 +1,23 @@
...
 import (
 	"fmt"
-
-	"github.com/brimdata/super/pkg/field"
 )
 
-// A path is an array of string or Forks
-type Path []any //XXX clean this up later
-type Fork []Path
+// A Projection is an array of string or Forks
```
nwt, line 7: Nit:
```suggestion
// A Projection is an slice of string or Forks
```

nwt (approved):

PR #5775, opened 2025-04-07
unify pushdown interface

mccanne:
This commit unifies the pushdown logic for filtering, BSUP filtering, CSUP metadata filtering, and projection into a single interface called zbuf.Pushdown replacing zbuf.Filter.  We also added projections to the data filters and arranged for the CSUP reader to project only the needed metadata values according to this projection.

These changes are needed to an upcoming PR that will improve the efficiency of CSUP by doing lazy type loading.


compiler/kernel/op.go
```diff
@@ -713,11 +717,24 @@ func (b *Builder) compilePoolScan(scan *dag.PoolScan) (zbuf.Puller, error) {
 	return meta.NewSequenceScanner(b.rctx, slicer, pool, nil, nil, b.progress), nil
 }
 
-func (b *Builder) PushdownOf(e dag.Expr) *Filter {
-	if e == nil {
+func (b *Builder) NewPushdown(e dag.Expr, projection []field.Path) *pushdown {
```
nwt, line 720: But: Return a zbuf.Pushdown here since it's a little opaque for a public method to return an unexported type.

nwt: 👍

zfmt/dag.go
```diff
@@ -552,24 +552,37 @@ func (c *canonDAG) op(p dag.Op) {
...
+				c.ret()
+				c.open()
+				c.open(" pruner (")
+				c.ret()
+				c.write("expr: ")
+				c.expr(mf.Expr, "")
+				c.ret()
+				if len(mf.Projection) > 0 {
+					c.write("proj:")
+					c.fields(mf.Projection)
```
nwt, line 579: Nit: This winds up printing "proj: fields a,b,c,..." so maybe drop the "proj:"?

mattnibs: or drop the word "fields"

compiler/optimizer/demand.go
```diff
@@ -109,8 +109,16 @@ func demandForOp(op dag.Op, downstream demand.Demand) demand.Demand {
 	case *dag.CommitMetaScan, *dag.DefaultScan, *dag.Deleter, *dag.DeleteScan, *dag.LakeMetaScan:
 		return demand.None()
 	case *dag.FileScan:
-		d := demand.Union(downstream, demandForExpr(op.Filter))
-		op.Fields = demand.Fields(d)
+		d := downstream
+		if mf := op.Pushdown.MetaFilter; mf != nil {
+			filterDemand := demandForExpr(mf.Expr)
+			mf.Projection = demand.Fields(filterDemand)
+			d = demand.Union(d, filterDemand)
```
nwt, line 116: I know this doesn't hurt but I don't think we need to merge the fields needed by the metadata filter into those needed by the overall query.

nwt (approved):

compiler/optimizer/demand.go
mccanne: Yes this is important clarificaiton.

PR #5776, opened 2025-04-07
streaming of CSUP objects

mccanne:
This commit changes the CSUP reader to stream the objects found within any given file instead of reading them in all at once. The vector pullers provide concurrency so no additional attempts to read-ahead are implemented here.

### DIFF
--- a/zio/csupio/stream.go
+++ b/zio/csupio/stream.go
@@ -1,0 +1,31 @@
+package csupio
+
+import (
+	"io"
+	"math"
+	"sync"
+
+	"github.com/brimdata/super/csup"
+)
+
+type stream struct {
+	mu  sync.Mutex
+	r   io.ReaderAt
+	off int64
+}
+
+func (s *stream) next() (*csup.Object, error) {
+	// NewObject puts the right end to the passed in SectionReader and since
+	// the end is unkown just pass MaxInt64.
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	o, err := csup.NewObject(io.NewSectionReader(s.r, s.off, math.MaxInt64))
+	if err != nil {
+		if err == io.EOF {
+			err = nil
+		}
+		return nil, err
+	}
+	s.off += int64(o.Size())
+	return o, nil
+}
--- a/zio/csupio/stream.go
+++ b/zio/csupio/stream.go
@@ -15,10 +15,12 @@ type stream struct {
 }
 
 func (s *stream) next() (*csup.Object, error) {
-	// NewObject puts the right end to the passed in SectionReader and since
-	// the end is unkown just pass MaxInt64.
+
 	s.mu.Lock()
 	defer s.mu.Unlock()
+	// We read the next object by creating a section reader that starts
+	// at the end of the previous object without an upper bound.  The csup
+	// package won't read off the end of the object so this is not a problem.
 	o, err := csup.NewObject(io.NewSectionReader(s.r, s.off, math.MaxInt64))
 	if err != nil {
 		if err == io.EOF {
--- a/zio/csupio/vectorreader.go
+++ b/zio/csupio/vectorreader.go
@@ -105,6 +105,7 @@ func (v *VectorReader) Pull(done bool) (vector.Any, error) {
 
 func pruneObject(zctx *super.Context, metaFilters *sync.Pool, m csup.Metadata) bool {
 	mf := metaFilters.Get().(*metafilter)
+	defer metaFilters.Put(mf)
 	vals := csup.ProjectMetadata(zctx, m, mf.projection)
 	for _, val := range vals {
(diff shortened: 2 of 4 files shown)


zio/csupio/stream.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"io"
+	"math"
+	"sync"
+
+	"github.com/brimdata/super/csup"
+)
+
+type stream struct {
```
nwt, line 11: Nit: Maybe call this objectStream to make its function a little more explicity?

zio/csupio/vectorreader.go
```diff
@@ -33,24 +31,11 @@ func NewVectorReader(ctx context.Context, zctx *super.Context, r io.Reader, push
...
-			return nil, err
-		}
-	}
 	return &VectorReader{
 		ctx:           ctx,
 		zctx:          zctx,
 		activeReaders: &atomic.Int64{},
-		nextObject:    &atomic.Int64{},
-		objects:       filterObjects(zctx, metaFilter, objects, metaProjection),
+		stream:        &stream{r: ra},
```
mattnibs, line 38: You're not setting the pushdown field here and you should right?

zio/csupio/stream.go
```diff
@@ -0,0 +1,31 @@
...
+	mu  sync.Mutex
+	r   io.ReaderAt
+	off int64
+}
+
+func (s *stream) next() (*csup.Object, error) {
+	// NewObject puts the right end to the passed in SectionReader and since
+	// the end is unkown just pass MaxInt64.
```
nwt, line 19: 1. The meaning of "puts the right end to the passing in SectionReader" isn't clear.
2. s/unkown/unknown
3. Maybe move this below the defer so it's closer to the NewObject call?

zio/csupio/vectorreader.go
```diff
@@ -77,27 +62,29 @@ func (v *VectorReader) Pull(done bool) (vector.Any, error) {
...
 }
 
-func filterObjects(zctx *super.Context, pruner expr.Evaluator, objects []*csup.Object, projection field.Projection) []*csup.Object {
-	if pruner == nil {
-		return objects
+func pruneObject(zctx *super.Context, pushdown zbuf.Pushdown, m csup.Metadata) bool {
+	metaFilter, metaProjection, err := pushdown.MetaFilter()
+	if err != nil {
+		panic(err)
 	}
```
nwt, line 84: Bit of a shame to recompile this filter every time it's run. Maybe call pushdown.MetaFilter once in NewConcurrentPuller and put the results in VectorReader?

zio/csupio/vectorreader.go
nwt: True!

mccanne: yikes

zio/csupio/stream.go
mccanne: will update

zio/csupio/vectorreader.go
mccanne: @mattnibs pointed out to me that we need separate runtime closures to run concurrently.   Should we recycle compiled expressions in a sync.Pool...?

nwt: I don't think that's necessary. If you call pushdown.MetaFilter in NewConcurrentPuller and store the results in VectorReader, you'll wind up with one expr.Evaluator per goroutine, which should be sufficient.

mccanne: Oops I saw this after pushing the pool change.  How about we keep the pool so that we can change the way concurrency is done and not have to change it?  I'm find reverting to your suggestion here too.  Easy change.

nwt: Either's fine with me.

nwt (approved):

zio/csupio/vectorreader.go
```diff
@@ -77,27 +88,26 @@ func (v *VectorReader) Pull(done bool) (vector.Any, error) {
...
-	return slices.DeleteFunc(objects, func(o *csup.Object) bool {
-		return pruneObject(zctx, pruner, o.Metadata(), projection)
-	})
 }
 
-func pruneObject(zctx *super.Context, pruner expr.Evaluator, m csup.Metadata, projection field.Projection) bool {
-	vals := csup.ProjectMetadata(zctx, m, projection)
+func pruneObject(zctx *super.Context, metaFilters *sync.Pool, m csup.Metadata) bool {
+	mf := metaFilters.Get().(*metafilter)
```
mattnibs, line 107: Put back the metafilter once you are done?

mccanne: oopsies

mattnibs (approved):

PR #5778, opened 2025-04-08
lazy type decoding for CSUP

mccanne:
This commit addresses a scaling challenge for CSUP with respect to large numbers of types by arranging for the per-type metadata to be decoded on an as-needed basis.  To do so, we changed the metadata representation from a single super value to an array of flattened metadata records indexed by ID.  This allows us to incrementally unmarshal and build the shadow vectors for only the types needed by a query.  Additionally, the metadata object filter now performs projection too so that only the metadata values from the types needed are deserialized and acted upon by the metadata filter.

These changes affect the CSUP file format so we bumped its version number from 8 to 9.

The code that unmarshals the shadow vectors is not currently reentrant. If/when we want to allow concurrent updates (e.g., vcache used by parallel lake requests) we will need to invoke the proper locking protocol.

Some rough perf measurements indicate a 5X speedup for the Bluesky Million data set on a simple query projecting a single column.

### DIFF
--- a/csup/context.go
+++ b/csup/context.go
@@ -1,0 +1,70 @@
+package csup
+
+import (
+	"fmt"
+	"io"
+	"sync"
+
+	"github.com/brimdata/super"
+	"github.com/brimdata/super/sup"
+	"github.com/brimdata/super/zio/bsupio"
+)
+
+type Context struct {
+	mu     sync.Mutex
+	local  *super.Context // holds the types for the Metadata values
+	metas  []Metadata     // id to Metadata
+	values []super.Value  // id to unmarshaled Metadata
+	uctx   *sup.UnmarshalBSUPContext
+}
+
+type ID uint32
+
+func NewContext() *Context {
+	return &Context{local: super.NewContext()}
+}
+
+func (c *Context) enter(meta Metadata) ID {
+	id := ID(len(c.metas))
+	c.metas = append(c.metas, meta)
+	return id
+}
+
+func (c *Context) Lookup(id ID) Metadata {
+	if id >= ID(len(c.metas)) {
+		panic(fmt.Sprintf("csup.Context ID (%d) out of range (len %d)", id, len(c.values)))
+	}
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if c.metas[id] == nil {
+		if err := c.unmarshal(id); err != nil {
+			panic(err) //XXX
+		}
+	}
+	return c.metas[id]
+}
+
+func (c *Context) unmarshal(id ID) error {
+	if c.uctx == nil {
+		c.uctx = sup.NewBSUPUnmarshaler()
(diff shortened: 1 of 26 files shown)


runtime/vcache/object.go
```diff
@@ -46,7 +46,7 @@ func NewObject(ctx context.Context, engine storage.Engine, uri *storage.URI) (*O
 func NewObjectFromCSUP(object *csup.Object) *Object {
 	return &Object{
 		object: object,
-		root:   newShadow(object.Metadata(), nil, 0),
+		root:   nil,
```
nwt, line 49: Nit: No need to specify the zero value.
```suggestion
```

runtime/vcache/shadow.go
```diff
@@ -16,131 +16,142 @@ import (
 // mutable shadow pieces that are dynamically loaded and maintained here.
 // The invariant is that runtime vectors are immutable while vcache.shadow
 // vectors are mutated under locks here as needed.
+//
+// Shadows are created incrementally so that a sequence of projections will do the
+// minimal work unmarshaling the csup metadata as needed.  When processing a sequence
+// of csup files with a single projection, the incremental capability is not important
+// but when caching csup objects (e.g., in local from S3), multiple threads operating
```
nwt, line 23: Nit:
```suggestion
// minimal work unmarshaling the CSUP metadata as needed.  When processing a sequence
// of CSUP files with a single projection, the incremental capability is not important
// but when caching CSUP objects (e.g., in local from S3), multiple threads operating
```

runtime/vcache/shadow.go
```diff
@@ -152,108 +163,213 @@ func (c count) length() uint32 {
 	return c.nulls + c.vals
 }
 
-// newShadow converts the CSUP metadata structure to a complete vector.Any
-// without loading any leaf columns.  Nulls are read from storage and unwrapped
+// unmarshal decodes the CSUP metadata structure to a (partial) shadow according
+// to the procided projection.  No vector data is actually loaded.
```
nwt, line 167:
```suggestion
// to the provided projection.  No vector data is actually loaded.
```

csup/context.go
```diff
@@ -0,0 +1,70 @@
...
+	"sync"
+
+	"github.com/brimdata/super"
+	"github.com/brimdata/super/sup"
+	"github.com/brimdata/super/zio/bsupio"
+)
+
+type Context struct {
+	mu     sync.Mutex
+	local  *super.Context // holds the types for the Metadata values
```
nwt, line 15: Maybe call this sctx for consistency?

nwt (approved):
Nice!

csup/context.go
mccanne: This raises an interesting point. I find it helpful to know it's different than the shared query context (i.e., created locally vs passed in) so that you need to do translations.

PR #5784, opened 2025-04-09
vam: Deunion array elements in array expressions

### LINKED ISSUE #5780: vam: Spreading union array values in array expressions
If you try to spread a union type array in an array expression in the vector runtime we do not de-union the individual values as is done in the sequential runtime:

```
$ echo '{a:[1],b:[2.,"bar"]}' | SUPER_VAM=1 super -c 'yield [a[1],...b]' -
[1,2.((float64,string)),"bar"((float64,string))]
```
When the result should be:
```
[1,2.,"bar"]
```

mattnibs:
Closes #5780

### DIFF
--- a/runtime/vam/expr/arrayexpr.go
+++ b/runtime/vam/expr/arrayexpr.go
@@ -41,8 +41,10 @@ func (a *ArrayExpr) eval(in ...vector.Any) vector.Any {
 		return vector.NewConst(super.Null, 0, nil)
 	}
 	var spreadOffs [][]uint32
+	unionTags := make([][]uint32, len(a.elems))
 	var viewIndexes [][]uint32
 	var vecs []vector.Any
+	var vecTags []uint32
 	for i, elem := range a.elems {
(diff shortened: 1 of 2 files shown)


runtime/vam/expr/arrayexpr.go
```diff
@@ -53,25 +55,41 @@ func (a *ArrayExpr) eval(in ...vector.Any) vector.Any {
...
+		for k, spreadOff := range spreadOffs {
+			tag := vecTags[k]
+			utags := unionTags[k]
 			if len(spreadOff) == 0 {
-				tags = append(tags, uint32(tag))
+				if utags != nil {
+					tags = append(tags, tag+utags[i])
+				} else {
+					tags = append(tags, tag)
+				}
```
nwt, line 80: Nit: Simplify?
```suggestion
				if utags != nil {
					tag += utags[i])
				}
				tags = append(tags, tag)
```

nwt (approved):

PR #5785, opened 2025-04-09
vam: set expressions

mattnibs:


runtime/vam/expr/setexpr.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+	return vector.Apply(false, s.eval, vecs...)
+}
+
+func (a *setExpr) eval(in ...vector.Any) vector.Any {
+	offsets, inner := buildList(a.sctx, a.elems, in)
+	// Dedupe list elems
+	vb := vector.NewBuilder(a.sctx.LookupTypeSet(inner.Type()))
+	var b zcode.Builder
+	for i := 0; i < len(offsets)-1; i++ {
```
nwt, line 35: Nit:
```suggestion
	for i := range offsets {
```

nwt (approved):

runtime/vam/expr/setexpr.go
mattnibs: You can't do this because you don't want traverse the last element in offsets

PR #5794, opened 2025-04-10
Add vector runtime support for is()

mattnibs:
Also make it so the type argument must be a type value not a string that can be parsed as a type value.

### DIFF
--- a/compiler/parser/ztests/match-is.yaml
+++ b/compiler/parser/ztests/match-is.yaml
@@ -1,5 +1,7 @@
 spq: |
-  ? a is(s2,'string')
+  ? a is(s2,<string>)
+
+vector: true
 
 input: |
   {s1:"a",s2:"b"}
--- a/runtime/sam/expr/function/types.go
+++ b/runtime/sam/expr/function/types.go
@@ -2,7 +2,6 @@ package function
 
 import (
 	"github.com/brimdata/super"
-	"github.com/brimdata/super/sup"
 	"github.com/brimdata/super/zcode"
 )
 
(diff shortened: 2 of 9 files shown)


runtime/sam/expr/function/types.go
```diff
@@ -87,11 +86,10 @@ func (i *Is) Call(_ super.Allocator, args []super.Value) super.Value {
 	}
 	var typ super.Type
 	var err error
-	if zvTypeVal.IsString() {
-		typ, err = sup.ParseType(i.sctx, string(zvTypeVal.Bytes()))
-	} else {
-		typ, err = i.sctx.LookupByValue(zvTypeVal.Bytes())
+	if zvTypeVal.Type().ID() != super.IDType {
+		return i.sctx.WrapError("is(): type value argument expected", zvTypeVal)
```
nwt, line 90: We don't usually include parentheses in these error strings.
```suggestion
		return i.sctx.WrapError("is: type value argument expected", zvTypeVal)
```

runtime/vam/expr/function/types.go
```diff
@@ -5,6 +5,37 @@ import (
...
+
+func (i *Is) Call(args ...vector.Any) vector.Any {
+	vec := args[0]
+	typeVal := args[1]
+	if len(args) == 3 {
+		vec = args[1]
+		typeVal = args[2]
+	}
+	if typeVal.Type().ID() != super.IDType {
+		return vector.NewWrappedError(i.sctx, "is(): type value argument expected", typeVal)
```
nwt, line 21:
```suggestion
		return vector.NewWrappedError(i.sctx, "is: type value argument expected", typeVal)
```

compiler/parser/ztests/match-is.yaml
```diff
@@ -1,5 +1,5 @@
 spq: |
-  ? a is(s2,'string')
+  ? a is(s2,<string>)
```
nwt, line 3: Add `vector: true`?

nwt (approved):

PR #5795, opened 2025-04-10
Add support for SQL substring function

mattnibs:


compiler/ast/sql.go
```diff
@@ -130,4 +130,13 @@ type SQLCast struct {
 	Loc  `json:"loc"`
 }
 
-func (*SQLCast) ExprAST() {}
+type SQLSubstring struct {
+	Kind string `json:"kind" unpack:""`
+	Expr Expr   `json:"expr"`
+	From Expr   `json:"from"`
+	For  Expr   `json:"for"`
```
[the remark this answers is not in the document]
mattnibs: I don't know `from` / `for` is what they are referred to in the syntax for the SQL expression if this was in the pipe language I'd agree with you.

compiler/parser/parser.peg
```diff
@@ -1118,6 +1118,20 @@ Function
         Loc: loc(c),
       }, nil
     }
+  / !FuncGuard SUBSTRING __ "(" __ e:Expr fromExpr:(_ FROM _ f:Expr { return f, nil })? forExpr:(_ FOR _ f:Expr { return f, nil })? ")" {
```
nwt, line 1121: Nit: These things are generally named to match the corresponding fields.
```suggestion
  / !FuncGuard SUBSTRING __ "(" __ expr:Expr from:(_ FROM _ f:Expr { return f, nil })? for_:(_ FOR _ f:Expr { return f, nil })? ")" {
```

mattnibs: `for` is a reserved word in go that's I add expr in front of everything for consistency.

nwt: Then say For or for_?

mattnibs: ok

compiler/semantic/expr.go
```diff
@@ -175,6 +175,40 @@ func (a *analyzer) semExpr(e ast.Expr) dag.Expr {
...
+		}
+		return &dag.Conditional{
+			Kind: "Conditional",
+			Cond: is,
+			Then: slice,
+			Else: &dag.Call{Kind: "Call", Name: "error", Args: []dag.Expr{
+				&dag.RecordExpr{
+					Kind: "RecordExpr",
+					Elems: []dag.RecordElem{
+						&dag.Field{Kind: "Field", Name: "message", Value: &dag.Literal{Kind: "Literal", Value: `"SUBSTRING: string value expected"`}},
```
nwt, line 206: Nit: These error strings usually read "value required".
```suggestion
						&dag.Field{Kind: "Field", Name: "message", Value: &dag.Literal{Kind: "Literal", Value: `"SUBSTRING: string value required"`}},
```

runtime/ztests/expr/sql-substring.yaml
```diff
@@ -0,0 +1,15 @@
+spq: yield SUBSTRING(this FROM 4 FOR 2), SUBSTRING(this FROM 4), SUBSTRING(this FOR 3)
```
nwt, line 1: Nit: Put this in compiler/ztests/sql/ since most (all?) of the SQL tests are in there.

nwt (approved):

runtime/ztests/expr/sql-substring.yaml
mattnibs: ok

compiler/ztests/sql/sql-substring.yaml
```diff
@@ -0,0 +1,15 @@
+spq: yield SUBSTRING(this FROM 4 FOR 2), SUBSTRING(this FROM 4), SUBSTRING(this FOR 3)
```
nwt, line 1: Nit: Just call this substring.yaml.

PR #5796, opened 2025-04-10
refactor null bit vectors and vector.Bool

mccanne:
This commit separates the implementation of bit vectors from the vector.Bool runtime vector.  We were using vector.Bool as a bit vector even when it wasn't a runtime vector causing confusion in the code (my bad for doing this originally).  We have moved the bit vector logic from vector.Bool into a new package in vector/bitvec and replaced all the vector.Any.Null bitmaps with a bitvec.Bits.


vector/view.go
```diff
@@ -20,36 +21,36 @@ func NewView(vec Any, index []uint32) *View {
...
+			nulls = bitvec.NewFalse(uint32(len(index)))
 			for k, idx := range index {
-				if val.Nulls.Value(idx) {
+				if val.Nulls.IsSet(idx) {
 					nulls.Set(uint32(k))
 				}
 				v := val.Index[idx]
 				index2[k] = v
-				counts[v]++
+				counts[v]++ //XXX this looks wrong
```
nwt, line 39:
```suggestion
				counts[v]++
```

vector/view.go
```diff
@@ -20,36 +21,36 @@ func NewView(vec Any, index []uint32) *View {
...
 			}
 		} else {
 			for k, idx := range index {
 				v := val.Index[idx]
 				index2[k] = v
-				counts[v]++
+				counts[v]++ //XXX this looks wrong
```
nwt, line 45:
```suggestion
				counts[v]++
```

runtime/vam/expr/function/math.go
```diff
@@ -151,14 +152,14 @@ func (l *Log) Call(args ...vector.Any) vector.Any {
...
 	for i := range vec.Len() {
 		v, isnull := vector.FloatValue(vec, i)
 		if isnull {
-			if nulls == nil {
-				nulls = vector.NewBoolEmpty(vec.Len(), nil)
+			if nulls.IsZero() {
+				nulls = bitvec.NewFalse(vec.Len())
 			}
-			nulls.Set(uint32(len(floats)))
+			nulls.Set(uint32(len(floats))) //XXX
```
nwt, line 162:
```suggestion
			nulls.Set(uint32(len(floats)))
```

runtime/vam/expr/function/types.go
```diff
@@ -115,7 +116,6 @@ func (t *TypeName) Call(args ...vector.Any) vector.Any {
 			out.Append(t.sctx.LookupTypeValue(typ).Bytes())
 		}
 	}
-	out.Nulls.SetLen(out.Len())
```
nwt, line 118: We still want this in case out.Len()<vec.Len().

vector/bitvec/bits.go
```diff
@@ -0,0 +1,168 @@
...
+	return n
+}
+
+// helpful to have around for debugging
+func (b Bits) String() string {
+	if b.IsZero() {
+		return "empty"
+	}
+	var s strings.Builder
+	for k := uint32(0); k < b.Len(); k++ {
```
nwt, line 109: Nit:
```suggestion
	for k := range b.Len() {
```

nwt (approved):

PR #5799, opened 2025-04-10
Move stray SQL tests into compiler/ztests/sql

nwt:
That's where all the other SQL tests live.


mattnibs:
Shouldn't the time and date stamp test be in the sql folder?

nwt:
It absolutely should. Thanks for catching that!

mattnibs (approved):

nwt:
Moved it in 83204fc7db9bb6ebeee6df34f9f3679cf82cbd9b.

PR #5800, opened 2025-04-10
add vector.BytesTable

mccanne:
This commits adds a bytes table for sharing across bytes, strings, and type-value vectors.  It abstracts away the complex offsets/bytes indexing logic and provides an interface that can act as the insertion point for more efficient mechanisms like German strings down the road.

This will also be used by the forthcoming lazy-types PR.

### DIFF
--- a/runtime/vam/expr/cast/string.go
+++ b/runtime/vam/expr/cast/string.go
@@ -115,7 +115,7 @@ func castToString(vec vector.Any, index []uint32) (vector.Any, []uint32, bool) {
 			offs = append(offs, uint32(len(bytes)))
 		}
 	}
-	return vector.NewString(offs, bytes, nulls), nil, true
+	return vector.NewString(vector.NewBytesTable(offs, bytes), nulls), nil, true
 }
 
 func timeToString(vec *vector.Int, index []uint32, n uint32) ([]uint32, []byte) {
--- a/vector/bytes.go
+++ b/vector/bytes.go
@@ -42,7 +42,7 @@ func (b *Bytes) Value(slot uint32) []byte {
 	if b.Nulls.IsSet(slot) {
 		return nil
 	}
-	return b.Table().Bytes(slot)
+	return b.table.Bytes(slot)
 }
 
 func (b *Bytes) Table() BytesTable {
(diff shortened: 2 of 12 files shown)


mattnibs (approved):

vector/string.go
```diff
@@ -7,36 +7,39 @@ import (
...
-	return NewString(make([]uint32, 1, length+1), nil, nulls)
+func NewStringEmpty(cap uint32, nulls bitvec.Bits) *String {
+	return NewString(NewBytesTableEmpty(cap), nulls)
 }
 
 func (s *String) Append(v string) {
-	s.Bytes = append(s.Bytes, v...)
-	s.Offsets = append(s.Offsets, uint32(len(s.Bytes)))
+	s.table.bytes = append(s.table.bytes, v...)
+	s.table.offsets = append(s.table.offsets, uint32(len(s.table.bytes)))
```
nwt, line 26: Nit: Use s.table.Append. The compiler's smart enough not to allocate memory for []byte(v).
```suggestion
	s.table.Append([]byte(v))
```

runtime/vam/expr/cast/string.go
```diff
@@ -115,7 +115,7 @@ func castToString(vec vector.Any, index []uint32) (vector.Any, []uint32, bool) {
 			offs = append(offs, uint32(len(bytes)))
 		}
 	}
-	return vector.NewString(offs, bytes, nulls), nil, true
+	return vector.NewString(vector.NewBytesTable(offs, bytes), nulls), nil, true
```
nwt, line 118: Nit: This happens often enough that I think it's worth continuing to have NewBytes accept slices and adding a new constructor like NewBytesFromTable (or else having NewBytes accept a table and adding NewBytesFromSlices, but I think the slice version will get called a lot more than the table version). Same goes for NewString.

vector/bytes.go
```diff
@@ -43,7 +42,11 @@ func (b *Bytes) Value(slot uint32) []byte {
 	if b.Nulls.IsSet(slot) {
 		return nil
 	}
-	return b.Bytes[b.Offs[slot]:b.Offs[slot+1]]
+	return b.Table().Bytes(slot)
```
nwt, line 45: Nit: 
```suggestion
	return b.table.Bytes(slot)
```

nwt (approved):

PR #5801, opened 2025-04-10
Fix issues with math coercion in vector runtime

mattnibs:


runtime/vam/expr/coerce.go
```diff
@@ -36,235 +34,7 @@ func coerceVals(sctx *super.Context, a, b vector.Any) (vector.Any, vector.Any, v
...
-		out := make([]int64, n)
-		for k := 0; k < n; k++ {
-			out[k] = int64(vals[k])
-		}
-		return vector.NewInt(super.TypeInt64, out, val.Nulls)
-	default:
-		panic(fmt.Sprintf("intToFloat invalid type: %T", val))
-	}
+	id, _ := coerce.Promote(super.NewValue(a.Type(), nil), super.NewValue(b.Type(), nil))
+	typ, _ := super.LookupPrimitiveByID(id)
```
nwt, line 38: I know neither of these is expected to fail, but if that changes, ignoring these errors will probably make someone's life harder. 
```suggestion
	id, err := coerce.Promote(super.NewValue(a.Type(), nil), super.NewValue(b.Type(), nil))
	if err != nil {
		panic(err)
	}
	typ, err := super.LookupPrimitiveByID(id)
		if err != nil {
		panic(err)
	}
```

mattnibs: yeah good point

runtime/vam/expr/index.go
```diff
@@ -47,7 +48,8 @@ func indexArrayOrSet(sctx *super.Context, vec, indexVec vector.Any) vector.Any {
 		vec, index = view.Any, view.Index
 	}
 	offsets, vals, nulls := arrayOrSetContents(vec)
-	indexVec = promoteToSigned(indexVec)
+	// XXX With uints there's a possiblity of overflow errors getting returned.
```
nwt, line 51: Let's deal with this and add a test for it. I think we just need something like this.

```go
	indexVec = cast.To(sctx, indexVec, super.TypeInt64)
	dyn, _ := indexVec.(*vector.Dynamic)
	var errs []uint32
	var viewIndexes []uint32
	for i := range indexVec.Len() {
		idx := i
		if index != nil {
			idx = index[i]
		}
		if dyn != nil && dyn.TypeOf(uint32(i)).ID() != super.IDInt64 {
			errs = append(errs, i)
			continue
		}
```

mattnibs: this doesn't work

mattnibs: Alright I pushed a fix for this.

runtime/sam/expr/eval.go
```diff
@@ -668,8 +668,15 @@ func (i *Index) Eval(ectx Context, this super.Value) super.Value {
...
+		return index
+	}
 	id := index.Type().ID()
-	if !super.IsInteger(id) {
+	if super.IsUnsigned(id) {
+		index = LookupPrimitiveCaster(sctx, super.TypeInt64).Eval(ectx, index)
+		goto again
+	} else if !super.IsInteger(id) {
 		return sctx.WrapError("index is not an integer", index)
 	}
```
nwt, line 681: Nit: Goto is gross. Maybe do something like this?

```go
	id := index.Type().ID()
	if super.IsUnsigned(id) {
		index = LookupPrimitiveCaster(sctx, super.TypeInt64).Eval(ectx, index)
		id = index.Type().ID()
	}
	if !super.IsInteger(id) {
		if index.IsError() {
			return index
		}
		return sctx.WrapError("index is not an integer", index)
	}
```

nwt (approved):

runtime/sam/expr/eval.go
mattnibs: Nice

runtime/sam/expr/eval.go
```diff
@@ -669,7 +669,13 @@ func (i *Index) Eval(ectx Context, this super.Value) super.Value {
 
 func indexArrayOrSet(sctx *super.Context, ectx Context, inner super.Type, vector zcode.Bytes, index super.Value) super.Value {
 	id := index.Type().ID()
+	if super.IsUnsigned(id) {
+		index = LookupPrimitiveCaster(sctx, super.TypeInt64).Eval(ectx, index)
```
nwt, line 673: I think you need to set id again: `id = index.Type().ID()`.

mattnibs: right

PR #5808, opened 2025-04-11
Kind bit overflow in vector.FuncCode

mattnibs:
Because vector.Kind has more than 8 values but vector.FuncCode only alots 3 bits for Kind, values over 8 overflow causing the incorrect arithfunc to get selected. Fix this by using 4 bits instead.

Also fix the error message for arithmetic on incompatible types so it matches the output of sequential runtime.


vector/kind.go
```diff
@@ -207,7 +207,23 @@ func ArithOpFromString(op string) int {
 	panic(op)
 }
 
+func OpCodeToString(op int) string {
```
nwt, line 210: Nit: For symmetry with ArithOpFromString.

```suggestion
func ArithOpToString(op int) string {
```

nwt (approved):

PR #5822, opened 2025-04-15
vam: Support net comparisons

mattnibs:


runtime/sam/expr/coerce/coerce.go
```diff
@@ -17,7 +19,13 @@ func Equal(a, b super.Value) bool {
 	}
 	switch aid, bid := a.Type().ID(), b.Type().ID(); {
 	case !super.IsNumber(aid) || !super.IsNumber(bid):
-		return aid == bid && bytes.Equal(a.Bytes(), b.Bytes())
+		if aid != aid {
```
nwt, line 22:
```suggestion
		if aid != bid {
```

mattnibs: whoops!

runtime/sam/expr/coerce/coerce.go
```diff
@@ -17,7 +19,13 @@ func Equal(a, b super.Value) bool {
 	}
 	switch aid, bid := a.Type().ID(), b.Type().ID(); {
 	case !super.IsNumber(aid) || !super.IsNumber(bid):
-		return aid == bid && bytes.Equal(a.Bytes(), b.Bytes())
+		if aid != bid {
+			return false
+		}
+		if aid == super.IDNet {
+			return NetCompare(super.DecodeNet(a.Bytes()), super.DecodeNet(b.Bytes())) == 0
+		}
```
nwt, line 27: I don't think we need this. Tests pass without it.

mattnibs: I don't like have the compare be different than what we do on the vector side. Maybe it all works out fine, maybe it doesn't our testing isn't the most robust in this area. If you want I can change vector side to convert nets to bytes and compare that?

nwt: Then my vote's for using `==` on netip.Prefixes in both runtimes.

nwt (approved):

PR #5823, opened 2025-04-15
don't deunion vector.Union in constructor

mccanne:
This commit moves the deunioning logic for a Union vector will nulls to vector.Deunion intead of living inside the union vector constructor. The union vector shouldn't have to know about Dynamic's null union values in Serialize.  If we are concerned about performance, we can cache the result of Deunion under a lock inside the union vector but I wouldn't worry about it for now.

The upcoming lazy-vectors PR will make use of this refactoring.


nwt (commented):
I think this is fine but leave approval to @mattnibs.

mattnibs (approved):

PR #5824, opened 2025-04-15
Add position function

mattnibs:


docs/language/functions/position.md
```diff
@@ -0,0 +1,26 @@
...
+```
+
+### Description
+
+The _position_ function returns the 1-based index where string `sub` first
+occurs in string `s`. If `sub` is not a sub-string of `s` then 0 is returned.
+
+### Examples
+
+```mdtest-spq {data-layout="stacked"}
```
nwt, line 19: @philrz: Do we need `{data-layout="stacked"}` here?

philrz:
The `super` on this branch is allowing us to run the original SQL [bench2/q3](https://github.com/andrewcrotty/mgbench/blob/master/bench2/q3/q3.sql) and get the correct result, so that's a functional 👍 for me! 🎉 

```
$ super -version
Version: v1.18.0-414-g602cfef2b

$ SUPER_VAM=1 super -f csv -c "
SELECT top_level,
       AVG(LENGTH(request) - LENGTH(REPLACE(request, '/', ''))) AS depth_avg
FROM (
  SELECT SUBSTRING(request FROM 1 FOR len) AS top_level,
         request
  FROM (
    SELECT POSITION('/' IN SUBSTRING(request FROM 2)) AS len,
           request
    FROM 'bench2.csup'
    WHERE status_code >= 200
      AND status_code < 300
      AND log_time >= TIMESTAMP '2012-12-01 00:00:00'
  ) AS r
  WHERE len > 0
) AS s
WHERE top_level IN ('/about','/courses','/degrees','/events',
                    '/grad','/industry','/news','/people',
                    '/publications','/research','/teaching','/ugrad')
GROUP BY top_level
ORDER BY top_level;"

top_level,depth_avg
/about,4.180225643663013
/courses,5.138040610251345
/degrees,3.385165907612232
/events,3.105441792057925
/grad,3.6942030710629687
/industry,4.094312105007292
/news,2.963451710977424
/people,4.485873513820637
/publications,2.245890410958904
/research,5.040612211618354
/teaching,2
/ugrad,2.4527363184079602
```

docs/language/functions/position.md
mattnibs: I have no idea what is it!

nwt: @philrz confirms that it isn't needed.
```suggestion
```mdtest-spq
```

nwt (approved):

docs/language/functions/position.md
philrz: `{data-layout="stacked"}` was added to hint the formatting of playground-style examples that use wider queries/data or produce wider outputs that would otherwise require the user to "scroll right" to see everything. We had complaints from the community in the past when reading examples required lots of "scroll right".

For this specific one, it's a close call, but yes, the `{data-layout="stacked"}` can be dropped. FWIW, in the default non-stacked formatting, a horizontal scrollbar _does_ appear because the error text is just a _little_ too wide, but scrolling right doesn't actually reveal any additional text.

image

And in the `{data-layout="stacked"}` formatting, it looks like this:

image

philrz: Derp. And of course in that prior comment my screenshots were only showing the horizontal scrollbar because I'd not updated my local superdb-wasm to a commit that had support for your new `position` function. @nwt's suggestion is definitely correct.

mattnibs:
@mccanne brought up a good point that this approach breaks things if you want to do a call to a udf call named `position` with an `in` expression. Going to reformat to accommodate this case.

nwt (approved):
:shipit:

PR #5835, opened 2025-04-16
csup: Fix MemLength and Length for ints and uint

mattnibs:
MemLength and Length for csup int and uint vectors were set incorrectly. Since they bypass general compression the length for both of these should be the size of the compressed data.


mattnibs:
I'm not sure we want to merge this as is but it definitely is something bares discussion.

PR #5838, opened 2025-04-17
Remove nullsMax param from runtime/sam/expr.NewComparator

nwt:
Remove nullsMax param from runtime/sam/expr.NewComparator

In its place:

* Add a new type, order.Nulls, representing the position (first or last)
  of nulls in a sort ordering, as in SQL.

* Add an order.Nulls field to runtime/sam/expr.SortExpr and a
  corresponding paramater to NewSortExpr.

* Honor the new field in runtime/sam/expr.Comparator.Compare and
  Compartor.sortStableIndices.

* Add an order.Nulls argument to NewSortExpr calls that maintains
  current behavior.

* Set the new field in runtime/sam/op/sort.NewComparator according to
  the nullsFirst parameter.

### DIFF
--- a/order/nulls.go
+++ b/order/nulls.go
@@ -1,0 +1,9 @@
+package order
+
+// Nulls represents the position of nulls in an ordering of values.
+type Nulls bool
+
+const (
+	NullsLast  Nulls = false
+	NullsFirst Nulls = true
+)
--- a/cmd/super/db/ztests/version.yaml
+++ b/cmd/super/db/ztests/version.yaml
@@ -1,3 +1,5 @@
+skip: Enable after creating first Git release tag.
+
 script: |
   super -version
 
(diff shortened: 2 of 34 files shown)


order/nulls.go
```diff
@@ -0,0 +1,9 @@
+package order
+
+// Nulls represents the position of nulls in an ordering of values.
+type Nulls bool
+
+const (
+	NullsLast  Nulls = false
+	NullsFirst Nulls = true
```
mattnibs, line 8: Shouldn't this be NullsMin/NullsMax to match the method in order/which?

nwt: I've found that this nulls stuff is cleaner and easier if as much of the code as possible reflects the "nulls first"/"nulls last" model in SPQ and SQL.

nwt: I should say that you don't really see that here but you do in #5847.

mattnibs (approved):

PR #5845, opened 2025-04-19
remove csup.Metadata.Type method

mccanne:
This commit removes the Type method from the csup.Metadata interface. It was unused and overly constrains the design of CSUP.  In a future PR, we plan to relax the correspondence between Metadata entities and their polymorphic types and the Type method would otherwise interfere with this change.


nwt (approved):
The all-red diff is the best diff.

PR #5847, opened 2025-04-23
Remove sort and top -nulls flag in favor of "nulls X" in sort exprs

nwt:
* Remove the -nulls flag from sort and top operators in the parser, the {ast,dag}.{Sort,Top}.NullsFirst fields, and the nullsFirst function parameters in runtime/sam/op/{sort,top} and runtime/vam/op.

* Add a dag.SortExpr.Null field with type order.Null

* Set dag.SortExpr.Null in compiler/semantic.semSortExpr according to the existing field ast.SortExpr.Nulls, which is already wired up in the parser to reflect "nulls first" and "nulls last" in sort expressions.

* Pass dag.SortExpr.Null to expr.NewSortExpr in compiler/kernel.


philrz:
@nwt: Docs changes look good to me. 👍

PR #5848, opened 2025-04-23
Remove CSUP reset interface and just create new encoders

mattnibs:


csup/dict.go
```diff
@@ -87,7 +84,7 @@ func (d *DictEncoder) encodeValues(group *errgroup.Group) {
 	for key, tag := range d.tags {
 		byteSlices[tag] = zcode.Bytes(key)
 	}
-	d.values.(resetter).reset()
+	d.values = newPrimitiveEncoder(d.typ)
```
nwt, line 87: Do you need this given that d.values is already initialized with the encoder passed to NewDictEncoder?

mattnibs: Yes because current values is not deduplicated so therefore needs to be rewritten.

nwt: Ah, I get it now.

nwt (approved):

PR #5852, opened 2025-04-25
Add compiler support for unordered dag.FileScan

nwt:
This provides a way for the optimizer to inform readers that they do not
need to preserve the order of values in a file and instead may return
values in arbitrary order.

It adds:

* A dag.Pushdown.Unordered field of type bool

* Logic in compiler/optimizer.setPushdownUnordered to set
  dag.Pushdown.Unordered in a dag.FileScan to true input order to the
  downstream DAG does not matter.

* An Unordered method in the zbuf.Pushdown interface to
  communicate the value of dag.Pushdown.Unordered to readers.


mccanne (approved):
Nice work!  Add support for join as we discussed and merge it!

PR #5853, opened 2025-04-25
csup: Encode offsets and bytes separate for strings

mattnibs:
This commit changes the CSUP encoding for strings, type values, and bytes
    values where the offsets of string/byte vectors are stored as an array of
    uint32 values and bytes are stored as compressed bytes.

### DIFF
--- a/csup/bytes.go
+++ b/csup/bytes.go
@@ -1,0 +1,79 @@
+package csup
+
+import (
+	"bytes"
+	"io"
+
+	"github.com/brimdata/super"
+	"github.com/brimdata/super/zcode"
+	"golang.org/x/sync/errgroup"
+)
+
+type BytesEncoder struct {
+	typ      super.Type
+	min, max []byte
+	bytes    zcode.Bytes
+	offsets  Uint32Encoder
+
+	// These values are used for the Encode pass.
+	bytesFmt uint8
+	bytesOut []byte
+	bytesLen uint64
+}
+
+func NewBytesEncoder(typ super.Type) *BytesEncoder {
+	return &BytesEncoder{typ: typ, offsets: Uint32Encoder{vals: []uint32{0}}}
+}
+
+func (b *BytesEncoder) Write(vb zcode.Bytes) {
+	if len(b.bytes) == 0 || bytes.Compare(vb, b.min) < 0 {
+		b.min = append(b.min[:0], vb...)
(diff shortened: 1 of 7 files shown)


csup/metadata.go
```diff
@@ -150,6 +150,38 @@ func (f *Float) Len(*Context) uint32 {
...
+}
+
+func (s *String) Type(*Context, *super.Context) super.Type {
+	return super.TypeString
+}
+
+func (s *String) Len(*Context) uint32 {
+	return s.BytesTable.Count
+}
+
```
nwt, line 184: You could add just a single type with a super.Type field here. That'd let you eliminate duplicate code in csup/bytes.go too.

```suggestion
type Bytes struct {
	Typ     super.Type `super:"Type"`
	Bytes   Segment
	Offsets Segment
	Min     []byte
	Max     []byte
	Count   uint32
}

func (b *Bytes) Type(*Context, *super.Context) super.Type {
	return b.Typ
}

func (b *Bytes) Len(*Context) uint32 {
	return b.Count
}

Comment (mattnibs) [reply csup/metadata.go:184]: Great idea

File: csup/bytes.go [hunk 2]
```diff
@@ -0,0 +1,81 @@
...
+func (b *BytesEncoder) Metadata(cctx *Context, off uint64) (uint64, ID) {
+	bytesLoc := Segment{
+		Offset:            off,
+		Length:            uint64(len(b.bytesOut)),
+		MemLength:         b.bytesLen,
+		CompressionFormat: b.bytesFmt,
+	}
+	off += bytesLoc.Length
+	var offsLoc Segment
+	off, offsLoc = b.offsets.Segment(off)
```
nwt, line 63: Nit:
```suggestion
	off, offsLoc := b.offsets.Segment(off)
```

csup/metadata.go
```diff
@@ -150,6 +150,22 @@ func (f *Float) Len(*Context) uint32 {
 	return f.Count
 }
 
+type Bytes struct {
+	Typ      super.Type `super:"Type"`
+	Bytes    Segment
+	Offsets  Segment
+	Min, Max []byte
```
nwt, line 157: Nit: Do these like the other types in here.
```suggestion
	Min []byte
	Max []byte
```

nwt (approved):

PR #5855, opened 2025-04-29
csup: Create metafilter for each concurrent reader

mattnibs:
Using the pool was slowing things down.

### DIFF
--- a/zio/csupio/vectorreader.go
+++ b/zio/csupio/vectorreader.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"io"
-	"sync"
 	"sync/atomic"
 
 	"github.com/brimdata/super"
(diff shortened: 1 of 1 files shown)


zio/csupio/vectorreader.go
```diff
@@ -22,59 +21,60 @@ type VectorReader struct {
 
 	activeReaders *atomic.Int64
 	stream        *stream
-	metaFilters   *sync.Pool
+	pushdown      zbuf.Pushdown
+	metaFilter    *metafilter
 	projection    field.Projection
```
nwt, line 26: Nit: Drop projection since it's now available as pushdown.Projection().

```suggestion
	pushdown      zbuf.Pushdown
	metaFilter    *metafilter
```

nwt (approved):

zio/csupio/vectorreader.go
mattnibs: From compiler/kernel/pushdown.go, it looks like the projection and metaProjection are two different things or am I missing something?

nwt: I think you might be. VectorReader.project is initialized to pushdown. VectorReader.projection is initialized to pushdown.Projection(). So the projection field isn't necessary and tests pass with the following diff:

mattnibs: gotcha

PR #5859, opened 2025-05-01
csup.Segment: Don't read empty Segments

mattnibs:
In addition to the benefit of skipping unnecessary work, this commit fixes an issue exposed in a future pr where reads to Segments of 0 length returns unexpected EOFs.

### DIFF
--- a/csup/segment.go
+++ b/csup/segment.go
@@ -30,6 +30,9 @@ var zbufPool = sync.Pool{
 // first s.MemLength bytes of b. If the length of b is less than s.MemLength,
 // Read returns [io.ErrShortBuffer].
 func (s *Segment) Read(r io.ReaderAt, b []byte) error {
+	if s.Length == 0 {
+		return nil
+	}
 	if len(b) < int(s.MemLength) {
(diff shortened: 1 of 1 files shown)


csup/segment.go
```diff
@@ -33,6 +33,9 @@ func (s *Segment) Read(r io.ReaderAt, b []byte) error {
 	if len(b) < int(s.MemLength) {
 		return fmt.Errorf("csup: segment read: %w", io.ErrShortBuffer)
 	}
+	if s.Length == 0 {
+		return nil
+	}
 	b = b[:s.MemLength]
```
nwt, line 39: Nit: Don't insert this between the check and use of b and s.MemLength.
```suggestion
	if s.Length == 0 {
		return nil
	}
	if len(b) < int(s.MemLength) {
		return fmt.Errorf("csup: segment read: %w", io.ErrShortBuffer)
	}
	b = b[:s.MemLength]
```

nwt (approved):
In the commit message, "this commit fixes exposed in future work" doesn't make sense.

PR #5864, opened 2025-05-01
csup.Dict: Move dictionary creation to Encode pass

mattnibs:
The commit changes the process of creating dictionaries for CSUP files so that:
1. Dictionaries creation is done in a multi-threaded manner in during the Encode pass (i.e., no slowing down single threaded calls to Write).
2. The dictionaries are created using the native values inside primitive vectors.


csup/dict.go
```diff
@@ -2,113 +2,75 @@ package csup
...
-func NewDictEncoder(typ super.Type, values Encoder) *DictEncoder {
-	var tags map[string]byte
-	if id := typ.ID(); id != super.IDUint8 && id != super.IDInt8 && id != super.IDBool {
-		// Don't bother using a dictionary (which takes 8-bit tags) to encode
-		// other 8-bit values.
-		tags = make(map[string]byte)
+func NewDictEncoder(typ super.Type, values Encoder) Encoder {
+	if _, ok := values.(primitive); !ok {
+		panic("Dict values encoder must be a primitive")
+	}
```
nwt, line 23: Feels off. How about this?

```suggestion
func NewDictEncoder(typ super.Type, values PrimitiveEncoder) Encoder {
```

PrimitiveEncoder would look like this:

```go
type PrimitiveEncoder interface {
	Encoder
	Dict() (Encoder, []byte, []uint32)
	ConstValue() super.Value
}
```

newPrimitiveEncoder would return that. And DictEncoder would embed it instead of Encoder.

mattnibs: I can dig it

mattnibs: Thanks this is better

csup/dict.go
```diff
@@ -2,113 +2,71 @@ package csup
...
-	"github.com/brimdata/super/zcode"
 	"golang.org/x/sync/errgroup"
 )
 
 type DictEncoder struct {
-	typ    super.Type
-	tags   map[string]byte
-	index  []byte
+	typ super.Type
+	PrimitiveEncoder
```
nwt, line 12: Nit: We usually put embedded fields first.
```suggestion
	PrimitiveEncoder
	typ super.Type
```

csup/primitive.go
```diff
@@ -12,7 +13,7 @@ import (
 
 const MaxDictSize = 256
 
-type PrimitiveEncoder struct {
+type ZcodeEncoder struct {
```
nwt, line 16: If you're going to rename this type, then you should also rename the file and the method receivers. That's a lot of noise unrelated to the goal of this PR so how about doing it in a separate PR?

mattnibs: We've taken the name PrimitiveEncoder in this pr so this had to be renamed which I think is fine

nwt: Call it primitiveEncoder then? (It doesn't actually need to be exported.)

nwt (approved):

csup/primitive.go
mattnibs: I'm all for not exporting this stuff but if you are going to do it there then you need to do it elsewhere. Also I don't like the term `PrimitiveEncoder` because there are already other primitive encoders.

PR #5878, opened 2025-05-07
Zero null Booleans and numerics in vector Parquet reader

### LINKED ISSUE #5530: avg calculation with Parquet file in vector runtime seems more "off"
## tl;dr

This is a follow-on from #5516. Repeating the `avg` calculation using the test data from that issue, consider the results in the table below. Treating the result `1058.5234017218875` from DuckDB as a baseline, with the different super formats and runtime options we see:

|**Format**|**Runtime**|**Result**|**Delta**|
|-|-|-|-|
|BSUP|sequential|`1058.523401720017`|`0.0000000018705`|
|CSUP|sequential|`1058.523401720017`|`0.0000000018705`|
|Parquet|sequential|`1058.523401720017`|`0.0000000018705`|
|CSUP|vector|`1058.5234017218877`|`0.0000000000002`|
|Parquet|vector|`1058.5297770606794`|`0.0063753387919`|

## Details

Repro is with super commit 883ffd2.

The the runs that generated the results in the table above are in https://github.com/brimdata/super/issues/5516#issuecomment-2539695424 and the test data is linked from that issue.

Users are accustomed to seeing small differences in precision with floating point math, so I expect this might all be explained by something about parallel operations. Indeed, we've seen some non-deterministic floating point calculations with other database systems, for instance. However, seeing the delta as soon as the 4th digit with vector Parquet is surprising compared to out at 9+ digits like the others, so I figured I'd surface this in case it's worthy of closer scrutiny.

nwt:
For Boolean and numeric types, the vector runtime requires that null vector slots contain the zero value.  This isn't always true for Arrow vectors coming from pqarrow, so explicitly enforce it when building Super vectors from them.

Fixes #5530.

### DIFF
--- a/zio/parquetio/vectorreader.go
+++ b/zio/parquetio/vectorreader.go
@@ -152,6 +152,10 @@ type vectorBuilder struct {
 func (v *vectorBuilder) build(a arrow.Array) (vector.Any, error) {
 	dt := a.DataType()
 	length := uint32(a.Len())
+	// For Boolean and numeric types, the runtime requires that null vector
+	// slots contain the zero value.  This isn't always true for Arrow
+	// vectors coming from pqarrow, so the code below must enforce it.
+	nulls := makeNulls(a)
 	// Order here follows that of the arrow.Type constants.
 	switch dt.ID() {
 	case arrow.NULL:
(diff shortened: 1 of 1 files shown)


zio/parquetio/vectorreader.go
```diff
@@ -375,10 +394,25 @@ func makeNulls(a arrow.Array) bitvec.Bits {
 	return bitvec.New(bits, uint32(n))
 }
 
-func convertSlice[Out, In uint8 | uint16 | uint32 | uint64 | int8 | int16 | int32 | int64 | float32 | float64](in []In) []Out {
+type numeric interface {
+	uint8 | uint16 | uint32 | uint64 | int8 | int16 | int32 | int64 | float32 | float64
```
mattnibs, line 398: nit: 
```suggestion
	constraints.Integer | constraints.Float
```

mattnibs (approved):

PR #5881, opened 2025-05-08
Allow string primitives for SQL AS labels

mattnibs:


compiler/parser/parser.peg
```diff
@@ -2169,7 +2169,7 @@ SelectElem
     }
 
 OptAsClause 
-  = _ AS _ id:Identifier { return id, nil }
+  = _ AS _ label:(Identifier / StringLiteral) { return label, nil }
   / _ !SQLGuard id:Identifier { return id, nil }
```
nwt, line 2173: PostgreSQL allows a double-quoted string here too.

```suggestion
  / _ !SQLGuard label:(Identifier / StringLiteral) { return label, nil }
```

compiler/semantic/sql.go
```diff
@@ -580,16 +580,28 @@ func dedupeColname(m map[string]struct{}, name string) string {
...
+			valid = label.Type == "string"
+			name = label.Text
+		case *ast.ID:
+			valid = true
+			name = label.Name
+		}
+		if !valid {
+			a.error(as.Label, errors.New("unexpected label type"))
+		} else if name == "" {
+			a.error(as.Label, errors.New("label name cannot be an empty string"))
```
nwt, line 603: Nit:

```suggestion
			a.error(as.Label, errors.New("label cannot be an empty string"))
```

nwt (approved):

PR #5891, opened 2025-05-12
Add min/max docs examples that use string values

philrz:
To further show the capabilities and limitations of the enhancement added in #5886.

### DIFF
--- a/docs/language/aggregates/max.md
+++ b/docs/language/aggregates/max.md
@@ -4,7 +4,7 @@
 
 ### Synopsis
 ```
-max(number) -> number
+max(number|string) -> number|string
 ```
 
 ### Description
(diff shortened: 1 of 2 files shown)


docs/language/aggregates/max.md
```diff
@@ -39,7 +42,31 @@ yield max(this)
...
+max(this)
+# input
+"bar"
+"foo"
+"baz"
+# expected output
+"foo"
+```
+
+Determining a maximum among mixed numeric and string inputs is not yet possible:
```
mattnibs, line 57: I would change this to say something like feeding `max` a mix of string and numeric values results in an error.

mattnibs (approved):

PR #5893, opened 2025-05-13
csup: Parallelize reading of CSUP metadata

mattnibs:
This commit improves the speed of reading CSUP file metadata by changing the single threaded csup.stream to only read the header of CSUP frames and instead have individual VectorReader threads read and serialize the frame's metadata.

### DIFF
--- a/zio/csupio/stream.go
+++ b/zio/csupio/stream.go
@@ -1,33 +1,60 @@
 package csupio
 
 import (
+	"context"
 	"io"
 	"math"
+	"runtime"
 	"sync"
 
 	"github.com/brimdata/super/csup"
 )
 
 type stream struct {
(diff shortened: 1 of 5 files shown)


zio/csupio/stream.go
```diff
@@ -1,33 +1,60 @@
...
-	r   io.ReaderAt
-	off int64
+	r    io.ReaderAt
+	ch   chan chunk
+	once sync.Once
+	ctx  context.Context
 }
 
-func (s *stream) next() (*csup.Object, error) {
+type chunk struct {
```
nwt, line 20: Nit: We'd usually call a type that packages values for a channel like this "result" or "streamResult".

nwt (approved):

PR #5894, opened 2025-05-13
length(): Count UTF-8 characters instead of byte length

### LINKED ISSUE #5873: LENGTH function on strings should count chars like SQL, not bytes
## tl;dr

Given input `data.json`:

```
{"foo": "hello😀"}
```

`super` currently returns a `LENGTH` of `9` (a byte count) for string `foo`. But to be compliant with standard SQL behavior, it should return `6` (a character count).

## Details

Repro is with super commit 124b7b9.

```
$ super -version
Version: v1.18.0-452-g124b7b9ec

$ super -c "SELECT LENGTH(foo) FROM 'data.json';"
{length:9}
```

Our approach with SuperDB is to be SQL-compatible to a similar degree as Postgres, and indeed, here Postgres returns a `LENGTH` of `6`. If you want the byte length, they also offer `OCTET_LENGTH`.

```
$ psql postgres
psql (17.4 (Homebrew))

postgres=# CREATE TABLE data ( foo TEXT NOT NULL );
CREATE TABLE
postgres=# INSERT INTO data (foo) values ('hello😀');
INSERT 0 1
postgres=# SELECT LENGTH(foo) FROM data;
 length 
--------
      6
(1 row)

postgres=# SELECT OCTET_LENGTH(foo) FROM data;
 octet_length 
--------------
            9
(1 row)
```

As they also tend to follow Postgres, DuckDB also returns a `LENGTH` of `6` and to get the byte length they offer `STRLEN`.

```
$ duckdb --version
v1.2.2 7c039464e4

$ duckdb -c "SELECT LENGTH(foo) FROM 'data.json';"
┌─────────────┐
│ length(foo) │
│    int64    │
├─────────────┤
│      6      │
└─────────────┘

$ duckdb -c "SELECT STRLEN(foo) FROM 'data.json';"
┌─────────────┐
│ strlen(foo) │
│    int64    │
├─────────────┤
│      9      │
└─────────────┘
```

Interestingly, ClickHouse goes against the SQL standard here and returns `LENGTH` of `9` here, and they offer a `CHAR_LENGTH` function that returns the `6`.

```
$ clickhouse --version
duckdb ClickHouse local version 25.4.2.31 (official build).

$ clickhouse --query "SELECT LENGTH(foo) FROM 'data.json'"
9

$ clickhouse --query "SELECT CHAR_LENGTH(foo) FROM 'data.json'"
6
```

In these situations, we intend to lean the SQL standard way, hence plan to change `LENGTH` to count characters like what Postgres and DuckDB were doing.

It's unclear whether we'd introduce another convenience function to still count bytes. I assume after we make a change the following will still return the same `9` the following does currently:

```
$ super -c "SELECT LENGTH(bytes(foo)) FROM 'data.json';"
{length:9}
```

...in which case we'd already have it covered via building blocks that we could call out in the docs. TBD.

mattnibs:
Closes #5873

### DIFF
--- a/runtime/ztests/expr/function/length.yaml
+++ b/runtime/ztests/expr/function/length.yaml
@@ -1,9 +1,11 @@
-spq: SELECT LENGTH('W3Schools.com');
+spq: SELECT LENGTH(s);
 
 vector: true
 
 input: |
-  null
+  {s:"W3Schools.com"}
+  {s:"emoji👀"}
 
 output: |
-  {"LENGTH(\"W3Schools.com\")":13}
+  {"LENGTH(s)":13}
+  {"LENGTH(s)":6}
--- a/runtime/sam/expr/function/len.go
+++ b/runtime/sam/expr/function/len.go
@@ -1,6 +1,8 @@
 package function
 
 import (
+	"unicode/utf8"
+
 	"github.com/brimdata/super"
 )
 
@@ -22,7 +24,9 @@ func (l *LenFn) Call(_ super.Allocator, args []super.Value) super.Value {
 		if err != nil {
 			panic(err)
 		}
-	case *super.TypeOfBytes, *super.TypeOfString, *super.TypeOfIP, *super.TypeOfNet:
+	case *super.TypeOfString:
+		length = utf8.RuneCount(val.Bytes())
+	case *super.TypeOfBytes, *super.TypeOfIP, *super.TypeOfNet:
 		length = len(val.Bytes())
 	case *super.TypeError:
 		return l.sctx.WrapError("len()", val)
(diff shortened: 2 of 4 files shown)


runtime/ztests/expr/function/length.yaml
```diff
@@ -1,9 +1,11 @@
-spq: SELECT LENGTH('W3Schools.com');
+spq: SELECT LENGTH(s);
```
nwt, line 1: Nit: No need for records here. Just make the input a sequence of strings.

```suggestion
spq: SELECT LENGTH(this);
```

nwt (approved):

runtime/ztests/expr/function/length.yaml
mattnibs: You get a semantic error if you pass `this` into this expression

mattnibs:
```
super compile -dag 'SELECT LENGTH(this);'
cannot reference 'this' in relational context; consider the 'yield' operator at line 1, column 15:
SELECT LENGTH(this);
              ~~~~
```

PR #5900, opened 2025-05-15
Add skip operator

mattnibs:


compiler/parser/parser.peg
```diff
@@ -517,6 +518,15 @@ TailOp
       }, nil
     }
 
+OffsetOp
+  = OFFSET _ !EndOfOp count:Expr {
```
nwt, line 522: No need for !EndOfOp here since an expression always follows OFFSET (unlike HeadOp and TailOp, where the expression is optional).

```suggestion
  = OFFSET _ count:Expr {
```

compiler/semantic/op.go
```diff
@@ -678,46 +678,27 @@ func (a *analyzer) semOp(o ast.Op, seq dag.Seq) dag.Seq {
 			Reverse: o.Reverse && len(sortExprs) == 0,
 		})
 	case *ast.Head:
-		val := super.NewInt64(1)
+		val := 1
```
nwt, line 681: Nit: Change this to something other than val since it's is no longer a super.Value.

```suggestion
		count := 1
```

compiler/semantic/op.go
```diff
@@ -678,46 +678,27 @@ func (a *analyzer) semOp(o ast.Op, seq dag.Seq) dag.Seq {
...
+			val = a.evalPositiveInteger(o.Count)
 		}
 		return append(seq, &dag.Head{
 			Kind:  "Head",
-			Count: int(val.AsInt()),
+			Count: val,
 		})
 	case *ast.Tail:
-		val := super.NewInt64(1)
+		val := 1
```
nwt, line 690: Nit: Change this to something other than val since it's no longer a super.Value.

```suggestion
		count := 1
```

nwt (commented):
I know "offset" was inspired by SQL, but it's not very intuitive as an operator name and I think "skip" would be better. For example, if you don't know the language, I think the meaning of "skip 5" is going a lot more obvious than the meaning of "offset 5".

What's your take?

compiler/parser/parser.peg
mattnibs: Thanks for the clarification!

mattnibs:
> I know "offset" was inspired by SQL, but it's not very intuitive as an operator name and I think "skip" would be better. For example, if you don't know the language, I think the meaning of "skip 5" is going a lot more obvious than the meaning of "offset 5".

> What's your take?

@nwt I actually originally had this called `skip` but changed it to more closely align with SQL. Now that you mention it I do think `skip` is a much more intuitive word so yeah let's go with that.

philrz:
FWIW, we're not the only ones. I did some poking around and while it looks like the major SQL implementations don't call it "skip", both [Informix](https://www.ibm.com/docs/en/informix-servers/14.10.0?topic=clause-restricting-return-values-skip-limit-first-options#ids_sqs_0982__title__2) and [Firebird](https://www.firebirdsql.org/refdocs/langrefupd20-select.html#langrefupd20-first-skip) both happen to use it. 🤷‍♂️

zfmt/dag.go
```diff
@@ -445,6 +445,9 @@ func (c *canonDAG) op(p dag.Op) {
 	case *dag.Tail:
 		c.next()
 		c.write("tail %d", p.Count)
+	case *dag.Skip:
+		c.next()
+		c.write("offset %d", p.Count)
```
nwt, line 450:
```suggestion
		c.write("skip %d", p.Count)
```

compiler/parser/parser.peg
```diff
@@ -517,6 +518,15 @@ TailOp
       }, nil
     }
 
+SkipOp
+  = OFFSET _ count:Expr {
```
nwt, line 522:
```suggestion
  = SKIP _ count:Expr {
```

compiler/dag/unpack.go
```diff
@@ -45,6 +45,7 @@ var unpacker = unpack.New(
 	Merge{},
 	Mirror{},
 	NullScan{},
+	Skip{},
```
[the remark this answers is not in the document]
mattnibs: Lazy of me apologies

compiler/optimizer/demand.go
```diff
@@ -117,6 +117,8 @@ func demandForSimpleOp(op dag.Op, downstream demand.Demand) demand.Demand {
 		return demandForAssignments(op.Args, downstream)
 	case *dag.Shape:
 		return downstream
+	case *dag.Skip:
+		return downstream
```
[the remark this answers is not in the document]
mattnibs: I'm not a fan of this since these aren't logically grouped.

PR #5903, opened 2025-05-15
join: Use smallest parent for hash table

### LINKED ISSUE #5671: Address deadlock in vector runtime self join
## tl;dr

#5659 was merged with the known limitation that a "self `join`" would deadlock if performed in the vector runtime, so we'll need to address this.

## Details

Repro is with super commit 0f8e7db.

Starting from the [self `join` example from the docs](https://superdb.org/docs/tutorials/join/#self-joins):

```
$ super -version
Version: v1.18.0-293-g0f8e7dbe

$ cat fruit.json people.json | super -z -I inner-join-streamed.spq -
{name:"figs",color:"brown",flavor:"plain",eater:"jessie"}
{name:"banana",color:"yellow",flavor:"sweet",eater:"quinn"}
{name:"strawberry",color:"red",flavor:"sweet",eater:"quinn"}
{name:"dates",color:"brown",flavor:"sweet",note:"in season",eater:"quinn"}
{name:"apple",color:"red",flavor:"tart",eater:"morgan"}
{name:"apple",color:"red",flavor:"tart",eater:"chris"}
```

If we set the environment variable to run the same query in vector runtime, it hangs indefinitely.

```
$ cat fruit.json people.json | SUPER_VAM=1 super -z -I inner-join-streamed.spq -
[hangs with no output]
```

mattnibs:
This commit changes the vector runtime's join operator to concurrently read from both the left and right parents till the first end of stream, then uses the smallest parent with which to build the hashtable from.

Closes #5671


philrz:
I can see this branch does indeed address the deadlock shown in the simple query in the linked issue #5671. Unfortunately it looks like the deadlock still happens when running the mgbench [bench3/q2 SPQ](https://github.com/brimdata/benchmarks/blob/main/benchmarks/mgbench/shared/bench3/q2.spq) in vector runtime.

To repro, download test data from `s3://brim-sampledata/mgbench/bench3.csup` and then run the query as:

```
$ super -version
Version: 2f7001dca

$ SUPER_VAM=1 super -c "
SELECT EXTRACT(HOUR FROM log_time) AS hr,
       device_id,
       device_name,
       CASE WHEN device_name LIKE 'coffee%' THEN 'coffee'
            WHEN device_name LIKE 'printer%' THEN 'printer'
            WHEN device_name LIKE 'projector%' THEN 'projector'
            WHEN device_name LIKE 'vending%' THEN 'vending'
            ELSE 'other'
       END AS device_category,
       device_floor,
       event_value
FROM 'bench3.csup'
WHERE event_type = 'power'
  AND log_time >= TIMESTAMP '2019-11-01 00:00:00'

// Fork the 'power_hourly' data so we can calculate 'power_avg' separately from
// 'category_power_avg'. We can't do them at the same time in a single
// aggregate because 'power_avg' includes a device's own measurements in
// the average while 'category_power_avg' excludes a device's measurements
// in that average. I can't really understand in human terms if/how how this
// better answers the top-line question so I assume it's just done to exercise
// conditional subqueries in the benchmark.

| fork (

  // Create the equivalent of the 'power_avg' from the SQL.
  => aggregate
       power_avg := avg(event_value)
       by hr,
          device_id,
          device_name,
          device_category,
          device_floor

  // Create the equivalent of the 'category_power_avg' from the SQL. Since this
  // is like a conditional subquery, we know it's going to be O(N^2). We create
  // copies of the 'power_hourly' data in which a 'device_id' is filtered out
  // of each, then calculate the unique per-device category averages by running
  // avg() on each of those copies.

  => fork (
       => yield {device_id, device_category}
          | aggregate union(this)
          | over this

       => pass
     )
     | join on device_category=device_category power_hourly := this

     // The filtering is done as a separate advance step rather than a 'where'
     // clause in the aggregate because there are some instances where a device
     // is the only one of a particular category. When that happens and we use
     // 'where', a bucket is still created for that device_id but since no
     // power_hourly.event_value values make it past the 'where' clause, we end up
     // with a null value for category_power_avg for that device_id and that
     // pollutes the results. By doing the filtering as a separate step in advance
     // we filter out all such devices such that no bucket for such a device is
     // created in the aggregation.

     | where device_id != power_hourly.device_id
     | aggregate
         category_power_avg := avg(power_hourly.event_value)
         by device_id,
            power_hourly.hr,
            power_hourly.device_category
)

// The 'power_avg' and 'category_power_avg' are now effectively separate data
// sequences blended together, so we isolate and re-join them to complete the
// final threshold calculation and comparison.

| switch (
  case has(power_avg) => put left_key := {device_id, hr}
  case has(category_power_avg) => put right_key := {device_id, hr:power_hourly.hr}
)
| join on left_key=right_key category_power_avg
| where power_avg > category_power_avg * 2
| drop left_key
| sort
    device_id,
    hr;"
```

What I observe in the Activity Monitor on my Intel-based Macbook is that the `super` process consumes around 100% CPU for close to a minute, but then its CPU usage drops to 0% and stays there indefinitely, which indicates to me that it's still deadlocking. This is just a guess, but I wonder if it's happening because the query contains multiple self-joins.

Issue #5902, opened 2025-05-15
Structured presentation of f-string errors


philrz:
There's an enhancement that's been proposed via inline text in the [Expressions doc](https://superdb.org/docs/language/expressions/#formatted-string-literals):

> If any expression results in an error, then the value of the f-string is the first error encountered in left-to-right order.
>
> TBD: we could improve an error result here by creating a structured error containing the string template text along with a list of values/errors of the expressions.

## Details

Repro is with super commit 9fcbd22.

Consider these attempted uses of f-strings, each of which generates an error:

```
$ super -version
Version: 9fcbd222b

$ echo '{numerator:22.0, denominator:0}' | super -c 'values f"pi is approximately {numerator / denominator}"' -
error("divide by zero")

$ echo '{numerator:22.0, denominator:0}' | super -c 'values f"attempt to concat missing strings: {foo + bar}" ' -
error("missing")
```

Per the docs text quoted above, if we attempt both expressions at once, we only ever see the first error.

```
$ echo '{numerator:22.0, denominator:0}' | super -c 'values f"pi is approximately {numerator / denominator} attempt to concat missing strings: {foo + bar}"' -
error("divide by zero")

$ echo '{numerator:22.0, denominator:0}' | super -c 'values f"attempt to concat missing strings: {foo + bar} attempt to concat missing strings: {foo + bar}" ' -
error("missing")
```

It might indeed be a nice way to show off the capabilities of the error type by following through on the proposed enhancement of presenting all the errors in a structured way.

PR #5905, opened 2025-05-16
De-Zed the Expressions doc

philrz:
I wanted to make a material change to the Expressions doc, but upon confronting it I could see it still had Zed remnants. Therefore as a prerequisite I'm giving it a first pass de-Zed-ing here along with a couple other corrections.

### DIFF
--- a/docs/language/expressions.md
+++ b/docs/language/expressions.md
@@ -3,12 +3,14 @@ weight: 5
 title: Expressions
 ---
 
-Zed expressions follow the typical patterns in programming languages.
-Expressions are typically used within pipeline operators
+Expressions follow the typical patterns in programming languages.
+Expressions are often used within [operators](operators/_index.md)
 to perform computations on input values and are typically evaluated once per each
 input value [`this`](pipeline-model.md#the-special-value-this).
 
-For example, `yield`, `where`, `cut`, `put`, `sort` and so forth all take
+For example, [`yield`](operators/yield.md), [`where`](operators/where.md),
+[`cut`](operators/cut.md), [`put`](operators/put.md),
+[`sort`](operators/sort.md) and so forth all take
 various expressions as part of their operation.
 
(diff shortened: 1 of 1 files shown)


docs/language/expressions.md
```diff
@@ -342,10 +344,6 @@ yield f"pi is approximately {numerator / denominator}"
 If any expression results in an error, then the value of the f-string is the
 first error encountered in left-to-right order.
 
-> TBD: we could improve an error result here by creating a structured error
-> containing the string template text along with a list of values/errors of
-> the expressions.
-
```
philrz, line 348: I thought this text was making a good point, but we generally track enhancements to-do items in issues so I opened #5902 and am proposing deleting the text here. If there's consensus that this is important enough to still mention in the doc I'd be open to morphing this text into something that includes a link to the open issue, but my prior attempts to include issue links in docs text haven't always been well-received in the past so I assume folks would prefer to drop it.

docs/language/expressions.md
```diff
@@ -548,7 +546,7 @@ yield under(this)
 Union values are powerful because they provide a mechanism to precisely
 describe the type of any nested, semi-structured value composed of elements
 of different types.  For example, the type of the value `[1,"foo"]` in JavaScript
-is simply a generic JavaScript "object".  But in Zed, the type of this
+is simply a generic JavaScript "object".  But in SuperSQL, the type of this
```
philrz, line 549: I paused for a bit here unsure if I should say Super (the format, since the value `[1,"foo"]` we just showed can be interpreted as a value in the Super format) or SuperSQL (the language, since `[1,"foo"]` can also be interpreted as an array literal in the language). Since this doc is technically part of the Language docs I went with the latter, though I'd be open to mentioning both if that's what folks prefer.

mccanne (approved):

PR #5907, opened 2025-05-18
Clarify that comparisons are via byte order

philrz:
## tl;dr

This adds clarity to the docs that comparison between values for things like `min`, `max`, `sort`, `compare`, and comparison expressions are done via byte order.

## Details

After the merge of #5886, a community user [asked on Slack](https://super-db.slack.com/archives/CTSMAK6G7/p1747236636614269):

>will y’all have docs (or maybe you do already) on specifics when it comes to unicode / collation issues like i’m semi-used to in postgres?

The Dev team confirmed that it's done via byte order similar to the C/POSIX collation in Postgres, so I've added text to that effect.

While a user may assume that all of these order-centric operations rely on the same underlying mechanism, I can tell there's some differing code between them, so if there's subtle differences in their behaviors that I'm not picking up on, please do chime in.

I waffled a bit in terms of how much redundancy to have here, e.g., if maybe the topic should be covered in one place and then linked from everywhere else, but I landed on the side of "one-stop shopping" in the different pages with some repetition. That said, I chose not to paste the text yet again on the `top` page since that operator is already framed as a special kind of `sort`, so I figured we can lean on the existing linkage there.

### DIFF
--- a/docs/language/aggregates/max.md
+++ b/docs/language/aggregates/max.md
@@ -11,6 +11,11 @@ max(number|string) -> number|string
 
 The _max_ aggregate function computes the maximum value of its input.
 
+When determining the _max_ of string inputs, values are compared via byte
+order. This is equivalent to
+[C/POSIX collation](https://www.postgresql.org/docs/current/collation.html#COLLATION-MANAGING-STANDARD)
+as found in other SQL databases such as Postgres.
+
 ### Examples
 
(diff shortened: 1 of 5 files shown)


mccanne (approved):
Let's merge this then in a subsequent change we can move the order semantics to a common place that can be referred to by these various functions.

PR #5913, opened 2025-05-21
proper double-quoted SQL identifiers

mccanne:
This commit cleans up the handling of SQL identifiers in double quotes. In SQL expressions, double-quoted strings are always now interpreted as identifiers.  In pipe expressions, double-quoted strings are string literals.  In either context, single-quoted strings are always string literals and backtick-quoted strings are always identifiers.

To do this, we added a new AST expression type to represent double-quoted strings so that this version of a string literal can be differentiated in the semantic pass and handled according to the rules above.

We also updated the grammar to allow backtick-quoted and double-quoted strings to be on the RHS of a dot operator.

### DIFF
--- a/compiler/ast/ast.go
+++ b/compiler/ast/ast.go
@@ -54,7 +54,7 @@ type ID struct {
 // originally double quoted to perform this analysis.
 type DoubleQuote struct {
 	Kind string `json:"kind" unpack:""`
-	Text string `json:"name"`
+	Text string `json:"text"`
 	Loc  `json:"loc"`
 }
 
--- a/compiler/ast/ast.go
+++ b/compiler/ast/ast.go
@@ -48,6 +48,16 @@ type ID struct {
 	Loc  `json:"loc"`
 }
 
+// DoubleQuote is specialized from the other primitive types because
+// these values can be interpreted either as a string value or an identifier based
+// on SQL vs pipe context.  The semantic pass needs to know the string was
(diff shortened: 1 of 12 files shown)


compiler/ast/ast.go
```diff
@@ -48,6 +48,16 @@ type ID struct {
 	Loc  `json:"loc"`
 }
 
+// DoubleQuote is specialized from the other primitive types because
+// these values can be interpreted either as a string value or an identifier based
+// on SQL vs pipe context.  The semantic pass needs to know the string was
+// originally double quoted to perform this analysis.
+type DoubleQuote struct {
+	Kind string `json:"kind" unpack:""`
+	Text string `json:"name"`
```
nwt, line 57:
```suggestion
	Text string `json:"text"`
```

zfmt/ast.go
```diff
@@ -103,6 +103,8 @@ func (c *canon) expr(e ast.Expr, parent string) {
 		c.literal(*e)
 	case *ast.ID:
 		c.write(e.Name)
+	case *ast.DoubleQuote:
+		c.expr(&ast.Primitive{Kind: "Primitive", Type: "string", Text: e.Text}, parent)
```
nwt, line 107: Nit: Simplify.
```suggestion
		c.write("%q", e.Text)
```

nwt (approved):

PR #5915, opened 2025-05-21
add SQL VALUES clause and column aliases

mccanne:
This commit adds support for a VALUES clause as well as an optional columns specification on table aliases.  The values listed in the values clause must all be uniformly typed.  When polymorphic data is desired, yield can be used inside of a SQL pipe instead of this values clause.  VALUES is not ANSI SQL but provides compatibility with PostgreSQL and duckdb.

### DIFF
--- a/compiler/parser/parser.peg
+++ b/compiler/parser/parser.peg
@@ -712,7 +712,7 @@ FromElem
         elem.Ordinality = o.(*ast.Ordinality)
       }      
       if alias != nil {
-        elem.Alias = alias.(*ast.Name)
+        elem.Alias = alias.(*ast.TableAlias)
       }
       return elem, nil
     }
@@ -1951,6 +1951,7 @@ SelectExpr
     body:(
         SetOperation
       / Select
+      / Values
       / "(" __ s:SQLPipe __")"      { return s, nil }
     )
     orderby:OptOrderByClause
@@ -2010,6 +2011,15 @@ Select
       return sel, nil
     }
 
+Values
+  = VALUES _ exprs:Exprs {
+      return &ast.Values{
+          Kind: "Values",
+          Exprs: sliceOf[ast.Expr](exprs),
+          Loc: loc(c),
+      }, nil
+    }
+
 OptDistinct
   = _ ALL        { return false, nil }
   / _ DISTINCT   { return true, nil }
@@ -2045,7 +2055,7 @@ CteList = first:Cte rest:( __ "," __ cte:Cte { return cte, nil} )* {
 Cte
   = name:SQLIdentifier _ AS m:OptMaterialized __ "(" __ s:SQLPipe __ ")" {
         return ast.CTE{
-            Name: name.(string),
+            Name: name.(*ast.ID).Name,
             Materialized: m.(*bool),
             Op: s.(ast.Op),
             Loc: loc(c),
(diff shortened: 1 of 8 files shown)


compiler/parser/parser.peg
```diff
@@ -2158,9 +2168,19 @@ OptAlias
...
+        alias.Columns = cols.([]*ast.ID)
+      }
+      return alias, nil
+    }
+
+OptColumns 
+  = "(" __ first:SQLIdentifier rest:( __ "," __ s:SQLIdentifier{ return s, nil })* __ ")" {
+        return sliceOf[*ast.ID](prepend(first,rest)), nil 
     }
+  / "" { return nil, nil }
```
nwt, line 2183:
```suggestion
  = (AS _)? !(SQLGuard / DeprecatedFroms) name:IdentifierName cols:Columns? { 
      alias := &ast.TableAlias{Kind: "TableAlias", Name: name.(string), Loc: loc(c)}
      if cols != nil {
        alias.Columns = cols.([]*ast.ID)
      }
      return alias, nil
    }

Columns 
  = "(" __ first:SQLIdentifier rest:( __ "," __ s:SQLIdentifier{ return s, nil })* __ ")" {
        return sliceOf[*ast.ID](prepend(first,rest)), nil 
    }
```

compiler/semantic/sql.go
```diff
@@ -322,35 +325,141 @@ func (a *analyzer) semSelectValue(sel *ast.Select, sch schema, seq dag.Seq) (dag
...
+		if err != nil {
+			a.error(astExpr, errors.New("expressions in values clause must be constant"))
+			return seq, badSchema()
+		}
+		recType, ok := super.TypeUnder(val.Type()).(*super.TypeRecord)
+		if !ok {
+			// Turn non record values into single column tuple
+			f := &dag.Field{Kind: "Field", Name: "c0", Value: e}
+			e = &dag.RecordExpr{Kind: "RecordExpr", Elems: []dag.RecordElem{f}}
+			val, _ := kernel.EvalAtCompileTime(sctx, e)
```
nwt, line 344: I know this should never fail but that could always change down the road and we want to know if it does.
```suggestion
			val, err := kernel.EvalAtCompileTime(sctx, e)
			if err != nil {
				panic(err)
			}
```

compiler/semantic/sql.go
```diff
@@ -322,35 +325,141 @@ func (a *analyzer) semSelectValue(sel *ast.Select, sch schema, seq dag.Seq) (dag
...
 	}
 	if len(seq) > 0 {
 		panic("semSQLOp: SQL pipes can't have parents")
 	}
-	return a.semSeq(op.Ops), &dynamicSchema{name: alias}
+	var name string
+	if alias != nil {
+		name = alias.Name
+		if len(alias.Columns) != 0 {
+			a.error(alias, errors.New("cannot apply aliased columns to dynamically typed data"))
```
nwt, line 392: Nit: Use the same string as below.
```suggestion
			a.error(alias, errors.New("cannot apply column aliases to dynamically typed data"))
```

compiler/semantic/sql.go
```diff
@@ -322,35 +325,141 @@ func (a *analyzer) semSelectValue(sel *ast.Select, sch schema, seq dag.Seq) (dag
...
+	case *staticSchema:
+		return mapColumns(sch.columns, alias, seq)
+	default:
+		return seq, sch, errors.New("cannot apply column aliases to dynamically typed data")
+	}
+}
+
+func mapColumns(in []string, alias *ast.TableAlias, seq dag.Seq) (dag.Seq, schema, error) {
+	if len(alias.Columns) > len(in) {
+		return nil, nil, fmt.Errorf("cannot apply %d aliased columns in alias %q to table with %d columns", len(alias.Columns), alias.Name, len(in))
```
nwt, line 431: Nit: Maybe clearer like this?
```suggestion
		return nil, nil, fmt.Errorf("cannot apply %d column aliases to table %q with %d columns", len(alias.Columns), alias.Name, len(in))
```

zfmt/ast.go
```diff
@@ -683,8 +683,21 @@ func (c *canon) fromElem(elem *ast.FromElem) {
 	if elem.Args != nil {
 		c.fromArgs(elem.Args)
 	}
-	if elem.Alias != nil {
-		c.write(" %s", sup.QuotedName(elem.Alias.Text))
+	c.tableAlias(elem.Alias)
```
nwt, line 686: Nit: The usual pattern in here is to do the nil check in the caller rather the callee.
```suggestion
	if elem.Alias != nil {
		c.tableAlias(elem.Alias)
	}
```

nwt (approved):

PR #5917, opened 2025-05-21
zfmt: Add missing style attribute to join operator

mattnibs:


compiler/parser/parser.peg
```diff
@@ -599,7 +599,7 @@ JoinStyle
   / INNER _ { return "inner", nil }
   / LEFT  _ { return "left", nil }
   / RIGHT _ { return "right", nil }
-  / ""        { return "inner", nil }
+  / ""        { return "", nil }
```
nwt, line 602: Nit: Fix this indentation while you're here.

```suggestion
  / ""      { return "", nil }
```

nwt (approved):

Issue #5910, opened 2025-05-20
Bucket mis-labels zero output


chrismo:
```shell
❯ super --version                 
Version: v1.18.0-350-g733dc02a1

❯ super -c "bucket(59s, 1m)"                                                     
0s

❯ super -c "bucket(59m, 1h)"
0s
```

I think `bucket` should retain the label of the second argument if possible. When it outputs 0 seconds for 59m in a 1h bucket, that's misleading IMO. I think it should say `0h` for zero hours, implying it's anything less than 1 hour, not less than 1 second.

philrz:
@chrismo and I had a detailed offline discussion on this one, but I'll summarize here as I morph this issue into something we can circle back to in the future.

The effect shown above is not unique to the `bucket` function, but instead follows from the nature of the `duration` data type and how the tools present values in a single, unique form.

As regards the data type, we refer to the [Data Model doc](https://superdb.org/docs/formats/data-model/#1-primitive-types) which describes a `duration` as:

>signed 64-bit integer as nanoseconds

i.e., input values like `0s`, `0m`, and `0h` are all equivalent ways of expressing what's ultimately stored as "zero nanoseconds". Then when such a value is read back, it's shown in a unique form, i.e., the tools have one way of expressing "zero nanoseconds", which in this case is `0s` for all three.

```
$ super -version
Version: 9fcbd222b

$ super -S -c "
values {
  zero_seconds: 0s,
  zero_minutes: 0m,
  zero_hours: 0h
}"
{
    zero_seconds: 0s,
    zero_minutes: 0s,
    zero_hours: 0s
}
```

The tools don't make any attempt to "remember" the units in which a `duration` was originally expressed in a query (e.g., the `m` in the call to `bucket`) and express results with those same units, instead preferring to have one way to express whatever values make up a query result.

For when a user wants to output a `duration` value in a non-default way, the way we expect we'd go about it is similar to what currently exists with the [`strftime` function](https://superdb.org/docs/language/functions/strftime/) for `time` values, i.e., a function that could use formatting rules to turn a `duration` into a `string`, which could potentially be cast back into a `duration` value again if desired (e.g., it could morph the `duration` value of `0s` from `bucket` above into the `string` value `0m`).

I'm going to change the title of this issue to reflect the intent to create such a function in the future.

chrismo:
Sounds good, thx again Phil! :)

PR #5919, opened 2025-05-22
Document canonical value presentation

philrz:
## What's Changing

This is my attempt to document how the tools present values in a canonical form in the human-readable output formats such as SUP.

## Why

When fielding community inquiry #5910 and writing up a response, I recognized that this topic might benefit from a little more coverage.

## Details

I expect that some aspects of canonical value presentation shown here would be unsurprising to most users, e.g., dropping trailing zeros in floating point values. OTOH, things like the `duration` type may be slightly less familiar to users, leading to confusion such as expressed in #5910. While docs enhancements like I'm doing here may not guarantee future users would never be confused by such things again, I figure it doesn't hurt to cover it _somewhere_, plus it gives me something in the docs to point at if similar questions from users come up in the future.

### DIFF
--- a/docs/commands/super.md
+++ b/docs/commands/super.md
@@ -315,6 +315,41 @@ produces
 When pretty printing, colorization is enabled by default when writing to a terminal,
 and can be disabled with `-color false`.
 
+#### Canonical Value Presentation
+
+Values are stored internally based on their types as defined in the
+[data model](../formats/data-model.md), and values that are output in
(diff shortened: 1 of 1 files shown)


philrz:
We've discussed this one and there's consensus that it probably shouldn't be merged in its current form, but @mccanne has expressed an interest in addressing the concern somehow in the ongoing revisions to the SUP format and related docs, so we're holding this one open as a reminder of that.

PR #5926, opened 2025-05-23
update join docs to conform with planned join syntax

mccanne:


docs/language/operators/join.md
```diff
@@ -6,48 +6,91 @@
 
 ```
 <left-input>
-| [anti|inner|left|right] join (
+| [anti|inner|left|right] join [as { <left-name>,<right-name> }] (
```
mattnibs, line 9: should there be the ability to do an anti right join?

mccanne: Yeah I don't think I have this quite right

PR #5931, opened 2025-05-30
Fix bug selecting values from nested SQL group by

mattnibs:


nwt (approved):
This looks fine to me but @mccanne should really approve since he understands the subtleties here.

One nit: Seems like PR the PR title could be more accurate since it mentions "nested SQL group by" but the diff doesn't have anything to do with aggregation.

mattnibs:
Closing in favor of #5937

mccanne:
#5937 fixes this.  Let's merge that this rebase this and merge the test defined here?

mattnibs:
@mccanne we can do but I think it would be easier just to close this pr and move the test here to your pr.

PR #5936, opened 2025-06-02
Support using integers in record index expressions

mattnibs:


runtime/sam/expr/eval.go
```diff
@@ -685,29 +685,38 @@ func indexArrayOrSet(sctx *super.Context, ectx Context, inner super.Type, vector
...
+		}
+		bytes, idx := getNthFromContainer(record, idx)
+		if idx < 0 {
+			return sctx.Missing()
+		}
+		return super.NewValue(typ.Fields[idx].Type, bytes)
+	}
 	if id != super.IDString {
-		return sctx.WrapError("record index is not a string", index)
+		return sctx.WrapError("invalid value for record index", index)
```
nwt, line 719: Nit: Maybe shorten?

```suggestion
		return sctx.WrapError("bad record index", index)
```

runtime/sam/expr/eval.go
```diff
@@ -685,29 +685,38 @@ func indexArrayOrSet(sctx *super.Context, ectx Context, inner super.Type, vector
...
+	if idx < 0 {
 		return sctx.Missing()
 	}
 	return deunion(inner, bytes)
 }
 
 func indexRecord(sctx *super.Context, typ *super.TypeRecord, record zcode.Bytes, index super.Value) super.Value {
 	id := index.Type().ID()
+	if super.IsInteger(id) {
+		idx := numericIndex(index)
```
nwt:
```suggestion
		idx := int(index.AsInt())
```

runtime/sam/expr/eval.go
philrz: Just one person's opinion, but I'm a fan of the more detailed message.

nwt: @philrz: But is it actually more detailed? If so, what additional information does it contain?

mattnibs: The word "bad" here to me sounds like it could mean there was something corrupt about the value passed in whereas saying "invalid value for" is more explicit in saying the value is not acceptable as a record index... Don't feel super strong about this though

runtime/vam/expr/index.go
```diff
@@ -86,8 +86,20 @@ func indexArrayOrSet(sctx *super.Context, vec, indexVec vector.Any) vector.Any {
 }
 
 func indexRecord(sctx *super.Context, vec, indexVec vector.Any) vector.Any {
-	if indexVec.Type().ID() != super.IDString {
-		return vector.NewWrappedError(sctx, "record index is not a string", indexVec)
+	if _, ok := indexVec.(*vector.Error); ok {
```
nwt: This doesn't handle vector.Named or vector.View. Maybe check for`indexVec.Kind() == super.ErrorKind`?

Also, I think moving this into the default case of the switch down below would make the overall behavior clearer.

mattnibs: Sure

nwt (approved):

PR #5938, opened 2025-06-02
Add SQL group by column ordinal

mattnibs:


compiler/semantic/sql.go
```diff
@@ -649,6 +657,14 @@ func (a *analyzer) semGroupBy(sch *selectSchema, in []ast.Expr) []exprloc {
 	return out
 }
 
+func (a *analyzer) isOrdinal(e dag.Expr) bool {
+	if literal, ok := e.(*dag.Literal); ok {
+		v := sup.MustParseValue(a.sctx, literal.Value)
+		return super.IsSigned(v.Type().ID())
```
nwt: This doesn't have to be signed, does it?

```suggestion
		return super.IsInterger(v.Type().ID())
```

compiler/ztests/sql/group-by-ordinal.yaml
```diff
@@ -0,0 +1,15 @@
...
+    data: |
+      {a:1,b:"foo"}
+      {a:1,b:"foo"}
+      {a:1,b:"bar"}
+
+outputs:
+  - name: stdout
+    data: |
+      {c:1(uint64)}
+      {c:2(uint64)}
```
nwt: I think you can make this a SPQ-style test.

```suggestion
spq: |
  select count() as c group by 2 order by c

vector: true

input: |
  {a:1,b:"foo"}
  {a:1,b:"foo"}
  {a:1,b:"bar"}

output: |
  {c:1(uint64)}
  {c:2(uint64)}
```

compiler/semantic/sql.go
```diff
@@ -638,6 +639,13 @@ func (a *analyzer) semGroupBy(sch *selectSchema, in []ast.Expr) []exprloc {
 	var funcs aggfuncs
 	for k, expr := range in {
 		e := a.semExprSchema(sch, expr)
+		if a.isOrdinal(e) {
+			e = &dag.IndexExpr{
+				Kind:  "IndexExpr",
+				Expr:  &dag.This{Kind: "This", Path: []string{"in"}},
```
nwt, line 645: Did you convince yourself that "in" is the right thing to use here or is that still an open question?

mattnibs: I think it has to be `in` b/c keys in a group by can only take values from `in`

nwt (approved):

PR #5947, opened 2025-06-09
analyzer.semJoinCond: Don't check result of semField

mattnibs:
This was resulting in two errors getting reported for a bad using expression.

### DIFF
--- a/compiler/semantic/sql.go
+++ b/compiler/semantic/sql.go
@@ -619,10 +619,7 @@ func (a *analyzer) semJoinCond(cond ast.JoinExpr) (dag.Expr, dag.Expr, error) {
 		if len(cond.Fields) > 1 {
 			return nil, nil, errors.New("join using currently limited to a single field")
 		}
-		key, ok := a.semField(cond.Fields[0]).(*dag.This)
-		if !ok {
-			return nil, nil, errors.New("join using key must be a field reference")
-		}
+		key := a.semField(cond.Fields[0])
 		return key, key, nil
 	default:
 		panic(fmt.Sprintf("semJoinCond: unknown type: %T", cond))


compiler/ztests/sql/join-ambiguous-using.yaml
```diff
@@ -0,0 +1,6 @@
+spq: SELECT city FROM weather JOIN cities USING(city)
+
+error: |
+  "city": ambiguous column reference at line 1, column 44:
+  SELECT city FROM weather JOIN cities USING(city)
+                                             ~~~~
```
nwt: This doesn't make sense. How is `using (city)` ambiguous?

```suggestion
  "city": ambiguous column reference at line 1, column 44:
  SELECT city FROM weather JOIN cities USING(city)
                                             ~~~~
```

mattnibs: Yeah you might be right about that. If you look at main right now this query will report two errors. So this does fix a problem but perhaps I need to rename the test:
```
super compile -C -dag 'SELECT city, temp_lo, temp_hi, prcp, date, location
    FROM weather JOIN cities USING (city);'
"city": ambiguous column reference at line 2, column 37:
    FROM weather JOIN cities USING (city);
                                    ~~~~
join using key must be a field reference at line 2, column 29:
    FROM weather JOIN cities USING (city);
                            ~~~~~~~~~~~~~
````

mattnibs: I'll just remove the test case

mattnibs: Definitely a bug with the `USING` case

nwt (approved):

mattnibs:
Closing, am going to fix this and the bad using clause stuff in an upcoming pr.

PR #5950, opened 2025-06-10
Fix join using

mattnibs:
Fix join using functionality by adapting joinSchema for the using clause to check that a given column exists for both the left and right schemas.

### DIFF
--- a/compiler/semantic/sql.go
+++ b/compiler/semantic/sql.go
@@ -619,11 +619,22 @@ func (a *analyzer) semJoinCond(cond ast.JoinExpr) (dag.Expr, dag.Expr, error) {
 		if len(cond.Fields) > 1 {
 			return nil, nil, errors.New("join using currently limited to a single field")
 		}
-		key, ok := a.semField(cond.Fields[0]).(*dag.This)
-		if !ok {
-			return nil, nil, errors.New("join using key must be a field reference")
+		if a.scope.schema != nil {
+			sch := a.scope.schema.(*joinSchema)
+			a.scope.schema = &joinUsingSchema{sch}
+			defer func() { a.scope.schema = sch }()
 		}
-		return key, key, nil
+		e := a.semField(cond.Fields[0])
+		key, ok := e.(*dag.This)
+		if !ok {
+			return e, e, nil
+		}
+		leftKey, rightKey := key, key
+		if a.scope.schema != nil {
+			leftKey = &dag.This{Kind: "This", Path: append([]string{"left"}, key.Path...)}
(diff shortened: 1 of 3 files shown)


compiler/semantic/sql.go
```diff
@@ -619,16 +619,45 @@ func (a *analyzer) semJoinCond(cond ast.JoinExpr) (dag.Expr, dag.Expr, error) {
...
 	}
 }
 
+type joinUsingSchema struct {
+	*joinSchema
+}
+
+func (j *joinUsingSchema) resolveColumn(col string) (field.Path, error) {
+	if _, err := j.left.resolveColumn(col); err != nil {
+		return nil, fmt.Errorf("column %q in USING clause does not exist in right table", col)
```
```suggestion
		return nil, fmt.Errorf("column %q in USING clause does not exist in left table", col)
```

compiler/semantic/sql.go
```diff
@@ -619,16 +619,45 @@ func (a *analyzer) semJoinCond(cond ast.JoinExpr) (dag.Expr, dag.Expr, error) {
...
+	}
+	if _, err := j.right.resolveColumn(col); err != nil {
+		return nil, fmt.Errorf("column %q in USING clause does not exist in left table", col)
```
```suggestion
		return nil, fmt.Errorf("column %q in USING clause does not exist in right table", col)
```

compiler/semantic/sql.go
```diff
@@ -619,16 +619,45 @@ func (a *analyzer) semJoinCond(cond ast.JoinExpr) (dag.Expr, dag.Expr, error) {
...
+	}
+	return field.Path{col}, nil
+}
+
+func (j *joinUsingSchema) resolveTable(table string) (schema, field.Path, error) {
```
nwt: Nit:

```suggestion
func (*joinUsingSchema) resolveTable(string) (schema, field.Path, error) {
```

compiler/semantic/sql.go
```diff
@@ -619,16 +619,45 @@ func (a *analyzer) semJoinCond(cond ast.JoinExpr) (dag.Expr, dag.Expr, error) {
...
+			rightKey = &dag.This{Kind: "This", Path: append([]string{"right"}, key.Path...)}
 		}
-		return key, key, nil
+		return leftKey, rightKey, nil
 	default:
 		panic(fmt.Sprintf("semJoinCond: unknown type: %T", cond))
 	}
 }
 
+type joinUsingSchema struct {
```
nwt: Nit: I'd put this in schema.go.

compiler/semantic/sql.go
mattnibs: doh

PR #5959, opened 2025-06-17
Add DefaultScan before semantic analysis

mattnibs:


compiler/semantic/analyzer.go
```diff
@@ -15,18 +15,20 @@ import (
 // Analyze performs a semantic analysis of the AST, translating it from AST
 // to DAG form, resolving syntax ambiguities, and performing constant propagation.
 // After semantic analysis, the DAG is ready for either optimization or compilation.
-func Analyze(ctx context.Context, ast *parser.AST, env *exec.Environment, extInput bool) (dag.Seq, error) {
-	files := ast.Files()
+func Analyze(ctx context.Context, p *parser.AST, env *exec.Environment, extInput bool) (dag.Seq, error) {
```
nwt, line 18: Nit: ast is a better name than p.

mattnibs: agreed but conflicts with the `ast` package

nwt: But I don't see the ast package used here.

nwt: Doh! I see it now.

PR #5962, opened 2025-06-17
Implement new version of join operator

mattnibs:


cmd/super/internal/lakemanage/scan.go
```diff
@@ -62,7 +62,9 @@ func scan(ctx context.Context, it *objectIterator, pool *pools.Config, runCh cha
 
 const iteratorQuery = `
 from %q@%q:objects
-| left join (from %q@%q:vectors) on id=id vector:=true
+| left join (from %q@%q:vectors) on id
+| left.vector := has(right)
+| yield left
```
nwt:
```suggestion
| yield {...left, vector: has(right)}
```

compiler/optimizer/demand.go
```diff
@@ -24,15 +24,10 @@ func demandForOp(op dag.Op, downstreams []demand.Demand) []demand.Demand {
...
 		return out
 	case *dag.Join:
-		left := downstreams[0]
-		for _, a := range op.Args {
-			if _, ok := a.LHS.(*dag.This); ok {
-				// Assignment clobbers a static field.
-				left = demand.Delete(left, demandForExpr(a.LHS))
-			}
-		}
+		in := downstreams[0]
```
nwt: Please change in to downstream so it's consistent with the rest of this file.

compiler/parser/parser.peg
```diff
@@ -578,19 +578,19 @@ ShapeOp
     }
 
 JoinOp
-  = style:JoinStyle JOIN rightInput:JoinRightInput _ e:JoinExpr optArgs:(_ FlexAssignments)? {
+  = style:JoinStyle JOIN alias:OptJoinAlias rightInput:JoinRightInput _ e:JoinExpr {
```
nwt: I think the alias and the condition should be adjacent because

```
join (
  from ...
) as {l,r} where l.a=r.b
```

reads better and is clearer than

```
join as {l,r} (
  from ...
) on l.a=r.b
```

mattnibs: I think agree but this from the doc written by @mccanne so let's hear what he has to say

compiler/ztests/pushdown.yaml
```diff
@@ -4,11 +4,11 @@ script: |
   echo === distinct
   super compile -C -O 'from file | distinct a | yield b'
   echo === fork-join
-  super compile -C -O 'fork (=> from file1 => from file2) | join on a b | yield c'
+  super compile -C -O 'fork (=> from file1 => from file2) | join on right.b=left.a | yield left.c'
```
nwt: I'd like to change this one so it references fields of both join inputs in the join condition as well as downstream in the yield.
```suggestion
  super compile -C -O 'fork (=> from file1 => from file2) | join on left.a=right.b | yield left.c,right.d'
```

mattnibs: The join condition does reference both inputs and is in purposely in reverse order to demonstrate the compiler can figure this out. I'm good with the yield though.

compiler/ztests/pushdown.yaml
```diff
@@ -4,11 +4,11 @@ script: |
   echo === distinct
   super compile -C -O 'from file | distinct a | yield b'
   echo === fork-join
-  super compile -C -O 'fork (=> from file1 => from file2) | join on a b | yield c'
+  super compile -C -O 'fork (=> from file1 => from file2) | join on right.b=left.a | yield left.c,right.d'
   echo === join
-  super compile -C -O "from file1 | join (from file2) on a b | yield c"
+  super compile -C -O "from file1 | join (from file2) on a | yield {c:left.c,b:right.b}"
```
nwt:
```suggestion
  super compile -C -O "from file1 | join (from file2) on left.a=right.b | yield left.c,right.d"
```

compiler/ztests/pushdown.yaml
```diff
@@ -4,11 +4,11 @@ script: |
   echo === distinct
   super compile -C -O 'from file | distinct a | yield b'
   echo === fork-join
-  super compile -C -O 'fork (=> from file1 => from file2) | join on a b | yield c'
+  super compile -C -O 'fork (=> from file1 => from file2) | join on right.b=left.a | yield left.c,right.d'
```
nwt: I'd like to keep the fields in alphabetical order for readability.

```suggestion
  super compile -C -O 'fork (=> from file1 => from file2) | join on right.a=left.b | yield left.c,right.d'
```

compiler/ztests/pushdown.yaml
```diff
@@ -4,11 +4,11 @@ script: |
...
   echo === join
-  super compile -C -O "from file1 | join (from file2) on a b | yield c"
+  super compile -C -O "from file1 | join (from file2) on a | yield {c:left.c,b:right.b}"
   echo === switch-join
-  super compile -C -O 'from file | switch a (case b => put x:=c case d => put x:=e ) | join on f g | yield h'
+  super compile -C -O 'from file | switch a (case b => put x:=c case d => put x:=e ) | join on f | yield {h:left.h,g:right.g}'
```
nwt:
```suggestion
  super compile -C -O 'from file | switch a (case b => put x:=c case d => put x:=e ) | join on left.f=right.g | yield left.h,right.i'
```

compiler/ast/ast.go
```diff
@@ -579,13 +579,17 @@ type (
...
-		Kind       string      `json:"kind" unpack:""`
-		Style      string      `json:"style"`
-		RightInput Seq         `json:"right_input"`
-		Cond       JoinExpr    `json:"cond"`
-		Args       Assignments `json:"args"`
+		Kind       string     `json:"kind" unpack:""`
+		Style      string     `json:"style"`
+		Alias      *JoinAlias `json:"alias"`
+		RightInput Seq        `json:"right_input"`
+		Cond       JoinExpr   `json:"cond"`
```
nwt: Nit: Keep the order here (and elsewhere) in sync with PEG rule?

```suggestion
		Style      string     `json:"style"`
		RightInput Seq        `json:"right_input"`
		Alias      *JoinAlias `json:"alias"`
		Cond       JoinExpr   `json:"cond"`
```

compiler/semantic/op.go
```diff
@@ -812,22 +820,27 @@ func (a *analyzer) semOp(o ast.Op, seq dag.Seq) dag.Seq {
...
+			Kind:       "Join",
+			Style:      style,
+			LeftAlias:  leftAlias,
+			LeftDir:    order.Unknown,
+			LeftKey:    leftKey,
+			RightAlias: rightAlias,
+			RightDir:   order.Unknown,
+			RightKey:   rightKey,
+		}
+		if o.RightInput != nil {
```
nwt: Nit: Indent less.

```suggestion
		if o.RightInput == nil {
			return append(seq, join)
		}
```

compiler/semantic/sql.go
```diff
@@ -629,17 +620,90 @@ func (a *analyzer) semJoinCond(cond ast.JoinExpr) (dag.Expr, dag.Expr, error) {
...
-			rightKey = &dag.This{Kind: "This", Path: append([]string{"right"}, key.Path...)}
-		}
-		return leftKey, rightKey, nil
+		return key, key, nil
 	default:
 		panic(fmt.Sprintf("semJoinCond: unknown type: %T", cond))
 	}
 }
 
+func (a *analyzer) semEquiJoin(leftAlias, rightAlias string, cond *ast.JoinOnExpr, lhs, rhs ast.Expr) (dag.Expr, dag.Expr) {
```
nwt: Nits: The usual pattern in this package is to name a method after the thing it analyzes and to pass the thing as the first argument.

```suggestion
func (a *analyzer) semJoinOnExpr(cond *ast.JoinOnExpr, leftAlias, rightAlias string, lhs, rhs ast.Expr) (dag.Expr, dag.Expr) {
```

compiler/semantic/sql.go
```diff
@@ -629,17 +620,90 @@ func (a *analyzer) semJoinCond(cond ast.JoinExpr) (dag.Expr, dag.Expr, error) {
...
+	for _, field := range fields {
+		if len(field.Path) == 0 {
+			a.error(in, errors.New("cannot join on this"))
+			return badExpr(), ""
+		}
+		aliases = append(aliases, field.Path[0])
+	}
+	slices.Sort(aliases)
+	aliases = slices.Compact(aliases)
+	if len(aliases) > 1 {
```
nwt: Nit: No need to sort these since you only care whether there's more than one distinct value.
```suggestion
	if len(slices.Compact(aliases)) > 1 {
```

compiler/semantic/sql.go
```diff
@@ -629,17 +620,90 @@ func (a *analyzer) semJoinCond(cond ast.JoinExpr) (dag.Expr, dag.Expr, error) {
...
+		a.error(in, errors.New("no field references in join expression"))
+		return badExpr(), ""
+	}
+	var aliases []string
+	for _, field := range fields {
+		if len(field.Path) == 0 {
+			a.error(in, errors.New("cannot join on this"))
+			return badExpr(), ""
+		}
+		aliases = append(aliases, field.Path[0])
```
nwt, line 680: Nit: Strip the alias here and you can drop the other for loop.

```suggestion
		aliases = append(aliases, field.Path[0])
		// Strip alias.
		field.Path = field.Path[1:]
```

nwt: Now that I think about it, stripping alias names in the DAG won't work when an alias appears on both sides of a comparison. I know this code isn't meant to handle that case but we might want to change the stripping approach in anticipation.

runtime/ztests/op/join-expr.yaml
```diff
@@ -1,11 +1,9 @@
 script: |
-  super -s -c 'left join (file B.sup) on s b | sort' A.sup
+  super -s -c 'left join (file B.sup) as {l,r} on s | yield {...l,...r} | sort' A.sup
   echo ===
-  super -s -c 'left join (file B.sup) on s=(lower(s)) b | sort' A.sup
+  super -s -c 'left join (file B.sup) as {l,r} on lower(r.s)=l.s | sort' A.sup
```
nwt, line 4: Nit: Output from this one's pretty gross so maybe add `| yield {...l,...r}`?

docs/tutorials/join.md
```diff
@@ -152,15 +157,15 @@ The query `anti-join.spq`:
 file fruit.json
 | anti join (
   file people.json
-) on flavor=likes
+) as {fruit,people} on fruit.flavor=people.likes
```
nwt, line 160: Nit: Add `yield fruit` since that's the prevailing pattern in here?

compiler/ztests/sql/readme-query.yaml
```diff
@@ -14,7 +14,8 @@ inputs:
         | SELECT VALUE {user:login,created_at:time(created_at)}
         => PASS
         )
-      | JOIN USING (user) repos
+      | JOIN USING (user)
+      | yield {...left,repos:right.repos}
```
nwt, line 18: Add this to the corresponding query in README.md.

compiler/ztests/join.yaml
```diff
@@ -0,0 +1,15 @@
...
+  - name: stdout
+    data: |
+      null
+      | inner join as {j,k} on jkey=kkey
+      | output main
+  - name: stderr
+    data: |
+      left and right join aliases cannot be the same at line 1, column 6:
+      join as {j,j} on key
+           ~~~~~~~~
```
nwt: Nit: I don't think the span should include "as" on these errors.

compiler/ztests/join.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+outputs:
+  - name: stdout
+    data: |
+      null
+      | inner join as {j,k} on jkey=kkey
+      | output main
+  - name: stderr
+    data: |
+      left and right join aliases cannot be the same at line 1, column 6:
```
nwt: We should have tests for the other possible errors as well.

compiler/semantic/sql.go
mattnibs: Yep that's a good point I think we need to change the approach here.

runtime/ztests/op/join-expr.yaml
mattnibs: I agree it's gross but it is the output of the operator and I thought it was good to have to one test somewhere that shows the unfiltered output.

compiler/ztests/join.yaml
mattnibs: It's easiest to do it the way I did but I can see if I can rewrite the peg rules to allow for this.

runtime/sam/op/join/join.go
```diff
@@ -23,18 +24,20 @@ type Op struct {
...
+	leftAlias   string
+	rightAlias  string
 	getLeftKey  expr.Evaluator
 	getRightKey expr.Evaluator
 	resetter    expr.Resetter
 	compare     expr.CompareFn
-	cutter      *expr.Cutter
 	joinKey     *super.Value
 	joinSet     []super.Value
 	splicer     *RecordSplicer
```
nwt: This is unused now.

nwt: And you can remove RecordSplicer too.

compiler/ztests/join.yaml
nwt: Just a nit so your call.

compiler/semantic/sql.go
mattnibs: We talked about this and decided to punt on this for the time being.

README.md
```diff
@@ -55,7 +55,8 @@ FROM 'https://data.gharchive.org/2015-01-01-15.json.gz'
    | SELECT VALUE {user:login,created_at:time(created_at)}
   => PASS
   )
-| JOIN USING (user) repos
+| JOIN USING (user)
+| yield {...left,repos:right.repos}
```
nwt: Nit: Be consistent with the rest of the query.

```suggestion
| YIELD {...left,repos:right.repos}
```

nwt (commented):
:shipit:

nwt (approved):
:shipit: for real

PR #5964, opened 2025-06-18
ztest: Combine runseq and runvec into runInternal

nwt:
runseq and runvec are nearly identical so combine them into a single function.


mattnibs (approved):
Bravo!

PR #5967, opened 2025-06-19
Fix record order for right joins

mattnibs:


compiler/kernel/vop.go
```diff
@@ -37,23 +37,12 @@ func (b *Builder) compileVam(o dag.Op, parents []vector.Puller) ([]vector.Puller
...
-		var anti, inner bool
-		switch o.Style {
-		case "anti":
-			anti = true
-		case "inner":
-			inner = true
-		case "left":
-		case "right":
-			leftAlias, rightAlias = rightAlias, leftAlias
+		if o.Style == "right" {
```
mccanne: I think this swapping logic should go inside the join.New() function.  Then the API to join.New() is clean.  Otherwise, the API is that you need to do the swap then the logic inside the runtime op does the swap in wrap.  Cleaner to just let the runtime op handle the swapping semantics.

mccanne (approved):

compiler/kernel/vop.go
mattnibs: Agreed, I'll fix that here

PR #5970, opened 2025-06-21
remove fat arrows

mccanne:
This commit removes fat arrows (=>) from SuperSQL and replaces them with parenthesised query fragments.

As part of this, we also removed the from (=> ... => ...) syntax where one can simply instead use fork with embedded froms. The deprecated forms of from (get/pool/file) have also been removed.

This change introduced a potential ambiguity for over so we added the keyword "into" to replace "=>" for over operators with lateral scopes.  The "into" concept will be reused when we add other forms of lateral scopes including partition-by, range windows, sliding windows, and so forth.

We did not carefully update the docs subdirectory as all of these changes will be accurately and comprehensively documented in the upcoming book branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,11 +50,10 @@ FROM 'https://data.gharchive.org/2015-01-01-15.json.gz'
   GROUP BY user
   ORDER BY len(repos) DESC
   LIMIT 5
-| FORK (
-  => FROM eval(f'https://api.github.com/users/{user}')
-   | SELECT VALUE {user:login,created_at:time(created_at)}
-  => PASS
-  )
+| FORK
+  ( FROM eval(f'https://api.github.com/users/{user}')
+    SELECT VALUE {user:login,created_at:time(created_at)} )
+  ( PASS )
 | JOIN USING (user)
 | YIELD {...left,repos:right.repos}
 ```
--- a/compiler/parser/parser.peg
+++ b/compiler/parser/parser.peg
@@ -54,7 +54,7 @@ FuncDecl
     }
 
 OpDecl
-  = OP _ name:Identifier __ "(" __ params:Identifiers? __ ")" __ ":" __ "(" __ body:OpDeclBody __ ")" {
+  = OP _ name:Identifier __ "(" __ params:Identifiers? __ ")" __ ":" __ body:ScopeBody {
       return &ast.OpDecl{
         Kind: "OpDecl",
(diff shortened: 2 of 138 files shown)


README.md
```diff
@@ -50,11 +50,10 @@ FROM 'https://data.gharchive.org/2015-01-01-15.json.gz'
...
   ORDER BY len(repos) DESC
   LIMIT 5
-| FORK (
-  => FROM eval(f'https://api.github.com/users/{user}')
-   | SELECT VALUE {user:login,created_at:time(created_at)}
-  => PASS
-  )
+| FORK
+  ( FROM eval(f'https://api.github.com/users/{user}')
+    SELECT VALUE {user:login,created_at:time(created_at)} )
```
mattnibs, line 55: Isn't this missing the `|` character in from of `SELECT`?
```suggestion
    | SELECT VALUE {user:login,created_at:time(created_at)} )
```

compiler/parser/parser.peg
```diff
@@ -865,29 +824,19 @@ MergeOp
     }
 
 OverOp
-  = (OVER / UNNEST) _ exprs:Exprs locals:Locals? body:Lateral? {
+   = (OVER / UNNEST) _ exprs:Exprs locals:Locals? (_ "into"i _)? __ body:ScopeBody? {
```
mattnibs: So the `into` keyword is not necessary to have a body for over? If that's the case can we just get rid of `into`?

compiler/ztests/sql/readme-query.yaml
```diff
@@ -9,11 +9,12 @@ inputs:
...
         ORDER BY len(repos) DESC
         LIMIT 2
-      | FORK (
-        => FROM eval(f'{user}.sup')
-        | SELECT VALUE {user:login,created_at:time(created_at)}
-        => PASS
+      | FORK
+        (
+          FROM eval(f'{user}.sup')
+          | SELECT VALUE {user:login,created_at:time(created_at)}
```
mattnibs: Get rid of the pipe here since its removed in the README
```suggestion
          SELECT VALUE {user:login,created_at:time(created_at)}
```

compiler/parser/parser.peg
mattnibs: My problem is that with this you can do `over this into` which I don't think makes sense.

mattnibs (approved):

PR #5986, opened 2025-06-27
Rename yield op to values

mattnibs:


compiler/semantic/sql.go
```diff
@@ -346,24 +337,17 @@ func (a *analyzer) semValues(values *ast.Values, seq dag.Seq) (dag.Seq, schema)
...
+		// always record values.
+		recType := super.TypeUnder(val.Type()).(*super.TypeRecord)
 		if schema == nil {
 			schema = recType
 		} else if schema != recType {
 			a.error(astExpr, errors.New("values clause must contain uniformly typed values"))
 		}
 		exprs = append(exprs, e)
 	}
+	// XXX I don't think this necessary.
```
mccanne: The idea is that if you are using SQL values semantics, it should be a table and have a schema that can be type checked.  If you want dynamic data, then you should use the values tuples, you should use SuperSQL literals in which case this code path shouldn't be encountered.

mccanne (approved):

Issue #5958, opened 2025-06-17
SQL: Support for "::" casting notation


philrz:
## tl;dr

```
$ super -c "SELECT 3::string"
parse error at line 1, column 11:
SELECT 3::string
      === ^ ===
```

## Details

Repro is with super commit cabf51f. This is a query from [a sqllogictest](https://github.com/duckdb/duckdb/blob/beb0fb08b2b3abee8605e8008503aae02e243ec5/test/sql/projection/coalesce_error.test#L10).

SuperDB currently has a couple supported syntaxes for type casting, e.g.,

```
$ super -version
Version: cabf51f1a

$ super -c "SELECT string(3);"
{"string(3)":"3"}

$ super -c "SELECT CAST(3 AS string);"
{"CAST(3 AS string)":"3"}
```

However, the following syntax (allegedly first popularized by Postgres) is not currently supported.

```
$ super -c "SELECT 3::string"
parse error at line 1, column 11:
SELECT 3::string
      === ^ ===
```

An example of another SQL system that supports it:

```
$ duckdb --version
v1.3.0 71c5c07cdd

$ duckdb -c "SELECT 3::string"
┌────────────────────┐
│ CAST(3 AS VARCHAR) │
│      varchar       │
├────────────────────┤
│ 3                  │
└────────────────────┘
```

philrz:
Verified in super commit b8311a3.

The `::` casting notation is now available in SQL queries, e.g.,

```
$ super -version
Version: b8311a333

$ super -c "SELECT 3::string"
{"3::<string>":"3"}
```

The available set of types to which a user can cast does not yet include the full set typically available in SQL (e.g., `VARCHAR` doesn't yet work here) but that coverage is tracked in separate issue #5953.

Thanks @mccanne!

PR #5994, opened 2025-07-01
add double-colon style casts

mccanne:


compiler/parser/ztests/colon-cast.yaml
```diff
@@ -0,0 +1,45 @@
+script: |
+  super -s -c "values 1::uint8, 2::int32, 3::float64, '2025-07-01T21:53:23.26504Z'::time"
+  super -s -c "values [1,2,3]::[string]"
+  super -s -c "values {x:1,s:'foo'}::{x:string,s:string}"
+  super -s -c "values 3.3.3.0/24::net, ::1/128::net, 2::/16::net"
+  super -s -c "values 3.3.3.3::ip, ::::ip, 2::::ip, (::)::ip, (2::)::ip"
```
mattnibs, line 6: Can we add a test for a union and perhaps a named type?

compiler/parser/parser.peg
```diff
@@ -992,7 +992,10 @@ UnaryPlusOrMinus
 PlusOrMinusOp = ("+" / "-") { return string(c.text), nil }
 
 DerefExpr
-  = expr:DerefExpr "[" __ from:AdditiveExpr __ ":" __ to:AdditiveExpr? __ "]" {
+  = expr:DerefExpr __ "::" __ typ:(TypeAsValue / TypeLiteral) {
+      return &ast.Cast{Kind: "Cast", Expr: expr.(ast.Expr), Type: typ.(ast.Expr), Loc: loc(c)}, nil
```
mattnibs: We've been going with the pattern that different syntaxes should be described using a different syntax node and here we are borrowing the node used for function call-like casts (e.g., `string(1)`).

philrz:
I see this covers us for SQL context as well, so I've linked this PR up to the issue #5958 I opened recently. E.g., testing with this branch with the repro example from that issue:

```
$ super -version
Version: 22f5e5fa4

$ super -c "SELECT 3::string"
{"<string>(3)":"3"}
```

compiler/parser/parser.peg
mccanne: Good point.  I decided to use the ast.BinaryExpr pattern and now the RHS of the `::` can be any expression including function calls that return type values.  Hmm, I need to add a test for that.  Will do.

PR #6000, opened 2025-07-05
change union syntax to use pipe(|) instead of comma separator

mccanne:
This commit changes the SuperSQL grammar and SUP format to use the pipe symbol for unions instead of parenthesized comma-separated list of types.  This also required a new variation of a embedded named type inside of a union where you can parenthesize the type, e.g., to disambigate foo=int64|string.  If foo=int64 should be inside of the union then the syntax is foo=(int64)|string.  Also, unions can be parenthesized when they are inside another union, e.g., int64|(uint8|int8)|string.

Comprehensive docs updates are coming in the book branch.


philrz:
What can be done about older format SUP/ZSON? Using this branch I tried reading a file that had a union of the prior format and it looks hazardous at the moment.

```
$ super -version
Version: f2fd31dbf

$ cat previous-union.sup 
{a:1((int64,string))}
{a:"hi"((int64,string))}

$ super -S previous-union.sup 
{
    "{a:1((int64": "{a:hi((int64",
    "string))}": "string))}"
}
```

mccanne:
> What can be done about older format SUP/ZSON? Using this branch I tried reading a file that had a union of the prior format and it looks hazardous at the moment.

This is autodetecting to CSV.  It wouldn't be hard to add backward compat for old style union syntax.  How about we get the breaking changes into SUP (we're also changing decorators to use :: cast syntax) and merge book with all the updated mdtests, then decide how we want to support backward compat?

nwt:
> This also required a new variation of a embedded named type inside of a union where you can parenthesize the type, e.g., to disambigate foo=int64|string. If foo=int64 should be inside of the union then the syntax is foo=(int64)|string.

`foo=(int64)|string` reads oddly to me. How about resolving the ambiguity by giving `|` higher precedence than `=` so that `foo=int64|string` is the named type `foo` with underlying type `int64|string` while `(foo=int64)|string` is the union of `foo=int64` and `string`?

PR #6029, opened 2025-07-15
update SUPER_DB_LAKE env to SUPER_DB in python lib

mccanne:


nwt (approved):
Nit: Remove lake at line 185, which looks like this.

```py
        description='Query default SuperDB lake service and print results.',
```

PR #6030, opened 2025-07-15
new docs infrastructure

mccanne:
This commit is the first step toward a new docs design based on mdbook from the Rust eco-system.  It also implements a better playground for query examples allowing inputs, editing of the query text, and copy-to-clipboard of the equivalent shell command.

The SUMMARY.md file defines the outline for the docs that are under development.  Subsequent PRs will post updates to this outline to fill out the book.

### DIFF
--- a/book/super-example/super-example.js
+++ b/book/super-example/super-example.js
@@ -82,7 +82,7 @@ for (const [i, pre] of preNodes.entries()) {
 
   const commandCode = node.querySelector('.super-command code')
   SuperPlayground.setup(node, (query, input) => {
-    commandCode.textContent = `echo '${input}'\\\n| super -s -c '${query}' -`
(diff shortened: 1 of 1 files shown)


book/super-example/super-example.js
```diff
@@ -0,0 +1,90 @@
...
+  div.innerHTML = html;
+  const node = div.children[0]
+  pre.replaceWith(node);
+
+  const tablist = node.querySelector('[role="tablist"]');
+  AriaTabs.setup(tablist);
+
+  const commandCode = node.querySelector('.super-command code')
+  SuperPlayground.setup(node, (query, input) => {
+    commandCode.textContent = `echo '${input}'\\\n| super -s -c '${query}' -`
```
nwt: I forgot a space here and think it improves readability of the command a little.

```suggestion
    commandCode.textContent = `echo '${input}' \\\n| super -s -c '${query}' -`
```

PR #6035, opened 2025-07-16
cleanup the enum type

mccanne:
This commit cleans up the enum design and brings it into alignment with other SQLs.  In this new approach, an enum is represented as a string cast to an enum type.  We updated the SUP parser and formatter, added enum types to the SuperSQL grammaer (that weren't previously there), and adapted the cast operator in the runtime to handle casts of strings to enum values.

Doc updates for SUP and SuperSQL will appear in the forthcoming book PRs.


mattnibs (approved):

runtime/sam/expr/cast.go
```diff
@@ -269,3 +273,21 @@ func (c *casterType) Eval(ectx Context, val super.Value) super.Value {
...
+
+type casterEnum struct {
+	sctx *super.Context
+	enum *super.TypeEnum
+}
+
+func (c *casterEnum) Eval(ectx Context, val super.Value) super.Value {
+	id := val.Type().ID()
+	if id != super.IDString {
+		return c.sctx.WrapError("cannot cast to type", val)
```
nwt:
```suggestion
	if val.Type().ID() != super.IDString {
		return c.sctx.WrapError("cannot cast to enum", val)
```

sup/formatter.go
```diff
@@ -178,8 +178,14 @@ func (f *Formatter) formatValue(indent int, typ super.Type, bytes zcode.Bytes, p
 	case *super.TypeMap:
 		null = f.formatMap(indent, t, bytes, known, parentImplied)
 	case *super.TypeEnum:
-		f.build("%")
-		f.build(t.Symbols[super.DecodeUint(bytes)])
+		f.build("\"")
+		code := super.DecodeUint(bytes)
```
nwt: Nit: We call this thing "selector" almost everywhere else.

sup/formatter.go
```diff
@@ -178,8 +178,14 @@ func (f *Formatter) formatValue(indent int, typ super.Type, bytes zcode.Bytes, p
...
+		if code < uint64(len(t.Symbols)) {
+			f.build(t.Symbols[code])
+		} else {
+			f.build("<sup-error: bad enum index>")
```
nwt: Nit: We generally panic on malformed values.

nwt (approved):

PR #6036, opened 2025-07-16
update format specs from old docs to new book docs

mccanne:

PR #6037, opened 2025-07-16
add new intro for SuperDB in the new mdbook docs

mccanne:


mccanne:
The bad markdown link will be resolved when #6036 is merged

book/src/intro.md
```diff
@@ -1 +1,376 @@
...
+
+SuperDB is implemented in a single, standalone executable called
+[`super`](command/super.md).
+There are no external dependencies to futz with.
+Just [install the binary](getting-started/install.md) and you're off and running.
+
+SuperDB separates compute and storage and is decomposed into
+a runtime system that
+* runs directly on any data inputs like files, streams, or APIs, or 
+* manipulates and queries data in a persistant
```
philrz:
```suggestion
* manipulates and queries data in a persistent 
```

book/src/intro.md
```diff
@@ -1 +1,376 @@
...
+argues that any attempt to supplant SQL or the RM is doomed to fail
+because any good ideas that arise from such efforts will simply be incorporated 
+into SQL and the RM.
+
+Yet, the incorporation of the JSON data model into the 
+relational model has left much to be desired.  One must typically choose 
+between creating columns of "JSON type" that layers in a parallel set of 
+operators and behaviors that diverge from core SQL semantics, or
+rely upon schema inference to convert JSON into relational tables,
+which unfornately does not always work.
```
philrz:
```suggestion
which unfortunately does not always work.
```

book/src/intro.md
```diff
@@ -1 +1,376 @@
...
+
+To understand the difficulty of schema inference,
+super this simple line of JSON data is in a file called `example.json`:
+```json
+{"a":[1,"foo"]}
+```
+
+> The literal `[1,"foo"]` is a contrived example but adequatedly
```
philrz:
```suggestion
> The literal `[1,"foo"]` is a contrived example but adequately 
```

book/src/intro.md
```diff
@@ -1 +1,376 @@
...
+> an API returning an array of JSON objects with varying shape.
+
+Surprisingly, this simple JSON input causes unpredictable schema inference
+across different SQL systems.
+Clickhouse converts the JSON number `1` to a string:
+```sh
+$ clickhouse -q "SELECT * FROM 'example.json'"
+['1','foo']
+```
+DuckDB does only partial schema infererence and leaves the contents
```
philrz:
```suggestion
DuckDB does only partial schema inference and leaves the contents 
```

book/src/intro.md
```diff
@@ -1 +1,376 @@
...
+DataFusion CLI v46.0.1
+Error: Arrow error: Cast error: Cannot cast string 'foo' to value of Int64 type
+```
+
+The more recent innovation of an open
+["variant type"](https://github.com/apache/spark/blob/master/common/variant/README.md)
+is more general than JSON but suffers from similar problems.
+In both these cases, the JSON type and the variant
+type are not individual types but rather entire type systems that differ 
+from the base relational type sysetem and so are shoehorned into the relational model
```
philrz:
```suggestion
from the base relational type system and so are shoehorned into the relational model
```

book/src/intro.md
```diff
@@ -1 +1,376 @@
...
+as a parallel type system masquerading as specialized type to make it all work.
```
philrz:
```suggestion
as a parallel type system masquerading as a specialized type to make it all work.
```

book/src/intro.md
```diff
@@ -1 +1,376 @@
...
+In this super-structured representation, the `type` field is a first-class type value
+representing an array type of elements having a sum type of `int64` and `string`.
+
+## SuperSQL
+
+Since super-structured data is a superset of the relational model, it turns out that
+a query language for super-structured data can be devised that is a superset of SQL.
+The SuperDB query language is a
+[Pipe SQL](https://research.google/pubs/sql-has-problems-we-can-fix-them-pipe-syntax-in-sql/)
+adapted for super-structured data scalled
```
philrz:
```suggestion
adapted for super-structured data called
```

book/src/intro.md
```diff
@@ -1 +1,376 @@
...
+{R:1.75}
+{kind:"rect",width:4.,height:1.}
+{kind:"rect",width:2.,height:3.5}
+```
+
+So what's going on here?  The data model here is acting both 
+as a strongly typed representation of JSON-like sequences as well as a 
+means to represent relational tables.  And SuperSQL is behaving like SQL
+when applied to table-like data, but at the same time is a
+pipe-syntax language for arbitrarilty typed data.
```
philrz:
```suggestion
pipe-syntax language for arbitrarily typed data.
```

book/src/intro.md
```diff
@@ -1 +1,376 @@
...
+But rather than replace SQL entirely, a very different approach is taken in 
+[SQL Has Problems. We Can Fix Them: Pipe Syntax In SQL](https://research.google/pubs/sql-has-problems-we-can-fix-them-pipe-syntax-in-sql/).
+Here the authors argue that except for compositional syntax, SQL is
+perfectly reasonable and we should live with its anachronisms (see Section 2.4).
+Thus, their
+[Pipe SQL specification](https://github.com/google/zetasql/blob/master/docs/pipe-syntax.md)
+carries forward SQL eccentricities into their modern adaptation
+of pipes for SQL.
+
+SuperSQL follows the Pipe SQL pattern but seizes the opportunity to modernrize
```
philrz:
```suggestion
SuperSQL follows the Pipe SQL pattern but seizes the opportunity to modernize
```

book/src/intro.md
```diff
@@ -1 +1,376 @@
...
+
+* array indexing is one-based in SQL clauses but zero-based in pipe operators,
+* column names in SQL clauses are case insensitive while record field references 
+are case sensitive in pipe operators,
+* complex scoping rules for table aliases and column references are required in
+relational SQL while binding from names to data in pipe operators is managed
+in a uniform and simple way as derefenced paths on `this`,
+* the syntactic structure of SQL clauses means all data must conform to a table
+whereas pipe operators can emit any data type desired in a varying fashion.
+* sum types are integral to piped data allowing mix-typed data proceessing
```
philrz:
```suggestion
* sum types are integral to piped data allowing mix-typed data processing
```

book/src/intro.md
```diff
@@ -1 +1,376 @@
...
+and results that need not fit in a uniform table,
+
+With this approach, SuperSQL can be adopted and used for existing use cases
+based on legacy SQL while incrementally expanding and embracing the 
+the pipe model tied to super-structured data.
```
philrz:
```suggestion
based on legacy SQL while incrementally expanding and embracing
the pipe model tied to super-structured data.
```

book/src/intro.md
```diff
@@ -1 +1,376 @@
...
+The super-structured data model is inherently more complex
+than the relational model and thus poses performance challenges.
+
+But since the relational model is a special case of the super-structured model,
+the myriad of optimization techniques developed for SQL and the relational 
+model can apply when super-structured data is homogeneously typed.  Moreover, it turns 
+out that you can implement a vectorized runtime for super-structured data by 
+organizing the data into vectorized units based on types instead of relational columns.
+
+While SuperDB is not yet setting speed records for relational queries, it's performance
```
philrz:
```suggestion
While SuperDB is not yet setting speed records for relational queries, its performance
```

book/src/intro.md
```diff
@@ -1 +1,376 @@
...
+## Summary
+
+To summarize, SuperDB provides a unified approach to analytics for eclectic JSON data
+and uniformly typed relational tables.  It does so in a SQL compatible fashion when 
+data looks like relational tables while offering 
+a modern pipe syntax suited for arbitrarily typed collections of data.
+
+In other words,
+SuperDB does not attempt to replace the relational model but rather leverages it
+in a more general approaach based on the super-structured data model where:
```
philrz:
```suggestion
in a more general approach based on the super-structured data model where:
```

book/src/intro.md
```diff
@@ -1 +1,376 @@
...
+
+* SuperSQL is a superset of SQL,
+* the super-structured data model is a superset of the relational model,
+* JSON easily and automatically represented with strong typing,
+* a polymorphic algebra generalizes the relational algebra, and
+* it all has an efficient vectorized implementation.
+
+## Next Steps
+
+If SuperDB or super-structured data has piqued you're interest at all,
```
philrz:
```suggestion
If SuperDB or super-structured data has piqued your interest at all,
```

book/src/intro.md
```diff
@@ -1 +1,376 @@
...
+feel free to dive in deeper:
+* explore the [SuperSQL](super-sql/intro.md) query language,
+* learn about the
+[super-structured data model](formats/model.md) and
+[formats](formats/intro.md) underlyng SuperDB, or
```
philrz:
```suggestion
[formats](formats/intro.md) underlying SuperDB, or
```

book/src/intro.md
```diff
@@ -1 +1,376 @@
...
+in a gradual slope, providing as much detail as desired,
+packaged up in the easy-to-understand
+[Super (SUP) data format](formats/sup.md) and
+[SuperSQL query language](super-sql/intro.md).
+
+## Super-structured Data
+
+Super-structured data is strongly typed and self describing.
+While compatible with relational schemas, SuperDB does not require such schemas
+as they can be modeled as [super-structured records](formats/model.md#21-record)
```
philrz:
```suggestion
as they can be modeled as [super-structured records](formats/model.md#21-record).
```

book/src/intro.md
```diff
@@ -1 +1,376 @@
...
+
+More specifically, a _relational table_ is simply a collection of tuples
+defined by a statically typed record,
```
philrz: Is it possible we mean "schema" here, not "record"?

book/src/intro.md
```diff
@@ -1 +1,376 @@
...
+
+Yet, the incorporation of the JSON data model into the 
+relational model has left much to be desired.  One must typically choose 
+between creating columns of "JSON type" that layers in a parallel set of 
+operators and behaviors that diverge from core SQL semantics, or
+rely upon schema inference to convert JSON into relational tables,
+which unfornately does not always work.
+
+To understand the difficulty of schema inference,
+super this simple line of JSON data is in a file called `example.json`:
```
philrz:
```suggestion
consider this simple line of JSON data is in a file called `example.json`:
```

book/src/intro.md
```diff
@@ -1 +1,376 @@
...
+```sh
+$ duckdb -c "SELECT * FROM 'example.json'"
+┌──────────────┐
+│      a       │
+│    json[]    │
+├──────────────┤
+│ [1, '"foo"'] │
+└──────────────┘
+```
+And Datafusion fails with an error:
```
philrz:
```suggestion
And DataFusion fails with an error:
```

book/src/intro.md
```diff
@@ -1 +1,376 @@
...
+Code: 386. DB::Exception: There is no supertype for types UInt8, String because some of them are String/FixedString/Enum and some of them are not. (NO_COMMON_TYPE)
+$ duckdb -c "SELECT [1,'foo']"
+Conversion Error:
+Could not convert string 'foo' to INT32
+
+LINE 1: SELECT [1,'foo']
+                  ^
+$ datafusion-cli -c "SELECT [1,'foo']" 
+DataFusion CLI v46.0.1
+Error: Arrow error: Cast error: Cannot cast string 'foo' to value of Int64 type
```
philrz:
```suggestion
$ clickhouse -q "SELECT [1,'foo']"
Code: 386. DB::Exception: There is no supertype for types UInt8, String because some of them are String/FixedString/Enum and some of them are not. (NO_COMMON_TYPE)

$ duckdb -c "SELECT [1,'foo']"
Conversion Error:
Could not convert string 'foo' to INT32

LINE 1: SELECT [1,'foo']
                  ^

$ datafusion-cli -c "SELECT [1,'foo']" 
DataFusion CLI v46.0.1
Error: Arrow error: Cast error: Cannot cast string 'foo' to value of Int64 type
```

It's personal preference, but when there's multiple command lines each showing their own output, I prefer the extra line breaks so it's more obvious where one ends and the next begins.

book/src/intro.md
```diff
@@ -1 +1,376 @@
...
+> <center><img src="codd-footnote.png" alt="Codd's Footnote" title="Codd's Footnote" width="500"></center>
+>
+> But sum types were notably absent.
+
+Armed with both sum and product types, super-structured data provides a 
+comprehensive algebraic type system that can represent any
+[JSON value as a concrete type](https://www.researchgate.net/publication/221325979_Union_Types_for_Semistructured_Data).
+And since relations are simply product types 
+as originally envisioned by Codd, any relational table can be represented
+also as a super-structure product type.  Thus, JSON and relational tables
```
philrz:
```suggestion
also as a super-structured product type.  Thus, JSON and relational tables
```

book/src/intro.md
```diff
@@ -1 +1,376 @@
...
+Instead of operating upon statically typed relational tables as SQL does,
+SuperSQL operates upon super-structured data.
+When such data happens to look like a table, then SuperSQL can work just 
+like SQL:
+```sh
+$ super -c '''
+SELECT a+b AS x, a-b AS y FROM (
+  VALUES (1,2),	(3,0)
+) AS T(a,b)
+'''
```
philrz: We probably shouldn't use the "triple quotes" that's currently in this and other examples here. It's well-known from Python of course (maybe other places I'm not familiar with?) but all the chatter I find online about it (such as [this Reddit thread](https://www.reddit.com/r/bash/comments/m8wyey/what_do_triple_double_quotes_do_in_bash/)) say that in Bourne shells the first two quotes in each triple basically evaluates to an empty string, so in this regard it's kind of a no-op but with the potential to puzzle a reader because they might mistaken it for some special syntax that's unique to SuperDB.

In terms of alternatives, it seems we could just do the usual of being careful to use one kind of quotes for the outer and one for the inner. Another approach would be to use a [heredoc](https://linuxize.com/post/bash-heredoc/) as I think @nwt suggested at one point, though I've just been hacking with it and I can't seem to come up with an approach that looks quite as clean as quoted input.

For instance, there's:

```
$ super -c "$(cat << EOF
SELECT a+b AS x, a-b AS y FROM (
  VALUES (1,2),(3,0)
) AS T(a,b)
)"
{x:3,y:-1}
{x:3,y:3}
```

Which does indeed allow for mixing the multiple quote types within the query, e.g.,

```
$ super -c "$(cat << EOF
values 'a',"b"
)"
"a"
"b"
```

But it's a little ugly. 🫤

book/src/intro.md
```diff
@@ -1 +1,376 @@
...
+    {kind:'rect',width:2.0,height:1.0},
+    {kind:'circle',radius:2},
+    {kind:'rect',width:1.0,height:3.5}
+)
+"""
+{R:1.75,W:1.5}
+```
+
+Things get more interesting when you want to do different types of processing 
+for differently type entities, e.g., let compute an average radius of circles,
```
philrz:
```suggestion
for differently typed entities, e.g., let compute an average radius of circles,
```

philrz:
```suggestion
for differently type entities, e.g., let's compute an average radius of circles,
```

book/src/intro.md
```diff
@@ -1 +1,376 @@
...
+
+So what's going on here?  The data model here is acting both 
+as a strongly typed representation of JSON-like sequences as well as a 
+means to represent relational tables.  And SuperSQL is behaving like SQL
+when applied to table-like data, but at the same time is a
+pipe-syntax language for arbitrarilty typed data.
+The super-structured data model ties it all together.
+
+To make this all work, the runtime must handle arbitrarily typed data.  Hence, every 
+operator in the SuperSQL has defined behavior for every possible input type.
```
philrz:
```suggestion
operator in SuperSQL has defined behavior for every possible input type.
```

book/src/intro.md
```diff
@@ -1 +1,376 @@
...
+Here the authors argue that except for compositional syntax, SQL is
+perfectly reasonable and we should live with its anachronisms (see Section 2.4).
+Thus, their
+[Pipe SQL specification](https://github.com/google/zetasql/blob/master/docs/pipe-syntax.md)
+carries forward SQL eccentricities into their modern adaptation
+of pipes for SQL.
+
+SuperSQL follows the Pipe SQL pattern but seizes the opportunity to modernrize
+the ergonomics of the language.  The vision here is that comprehensive backward
+compatibility can reside in the SQL operators while an modernized syntax and
```
philrz:
```suggestion
compatibility can reside in the SQL operators while a modernized syntax and
```

book/src/intro.md
```diff
@@ -1 +1,376 @@
...
+and improved ergonomics can reside in the pipe operators, e.g.,
+
+* array indexing is one-based in SQL clauses but zero-based in pipe operators,
+* column names in SQL clauses are case insensitive while record field references 
+are case sensitive in pipe operators,
+* complex scoping rules for table aliases and column references are required in
+relational SQL while binding from names to data in pipe operators is managed
+in a uniform and simple way as derefenced paths on `this`,
+* the syntactic structure of SQL clauses means all data must conform to a table
+whereas pipe operators can emit any data type desired in a varying fashion.
```
philrz:
```suggestion
whereas pipe operators can emit any data type desired in a varying fashion, and
```

book/src/intro.md
```diff
@@ -1 +1,376 @@
...
+* sum types are integral to piped data allowing mix-typed data proceessing
+and results that need not fit in a uniform table,
```
philrz:
```suggestion
and results that need not fit in a uniform table.
```

book/src/intro.md
```diff
@@ -1 +1,376 @@
...
+data looks like relational tables while offering 
+a modern pipe syntax suited for arbitrarily typed collections of data.
+
+In other words,
+SuperDB does not attempt to replace the relational model but rather leverages it
+in a more general approaach based on the super-structured data model where:
+
+* SuperSQL is a superset of SQL,
+* the super-structured data model is a superset of the relational model,
+* JSON easily and automatically represented with strong typing,
```
philrz:
```suggestion
* JSON is easily and automatically represented with strong typing,
```

book/src/intro.md
mccanne: That changes the meaning.  The point is to relate schema to records.  I will try to reword.

book/src/intro.md
mccanne: I like the cat EOF trick in general but we don't need it here so let's go with your first suggestion.

PR #6042, opened 2025-07-17
Rework UDFs

mattnibs:
This commit changes how udfs work so that they no longer rely on variables but instead rewrite the context value as a record with a field for every function parameter paired with its call expression.


mattnibs:
Going to close and fix this so we don't need to hack expr.Context to keep track of stack size and also make this work for vector runtime.

PR #6044, opened 2025-07-17
Rework UDFs

mattnibs:
This commit changes how udfs work so that they no longer rely on variables but instead rewrite the context value as a record with a field for every function parameter paired with its call expression. It also removes UDFs dependency on expr.Context to keep track of current stack size of recursive function calls.

### DIFF
--- a/compiler/kernel/expr.go
+++ b/compiler/kernel/expr.go
@@ -318,9 +318,9 @@ func (b *Builder) compileCall(call dag.Call) (expr.Evaluator, error) {
 	// First check if call is to a user defined function, otherwise check for
 	// builtin function.
 	var fn expr.Function
-	if e, ok := b.udfs[call.Name]; ok {
+	if u, ok := b.udfs[call.Name]; ok {
 		var err error
-		if fn, err = b.compileUDFCall(call.Name, e); err != nil {
+		if fn, err = b.compileUDFCall(call.Name, u); err != nil {
 			return nil, err
 		}
 	} else {
@@ -342,16 +342,20 @@ func (b *Builder) compileCall(call dag.Call) (expr.Evaluator, error) {
 	return expr.NewCall(fn, exprs), nil
 }
 
-func (b *Builder) compileUDFCall(name string, body dag.Expr) (expr.Function, error) {
+func (b *Builder) compileUDFCall(name string, f *dag.Func) (expr.Function, error) {
 	if fn, ok := b.compiledUDFs[name]; ok {
 		return fn, nil
 	}
-	fn := &expr.UDF{Name: name, Sctx: b.sctx()}
+	if b.udfStackDepth == nil {
+		b.udfStackDepth = new(int)
+		defer func() { b.udfStackDepth = nil }()
+	}
+	fn := expr.NewUDF(b.sctx(), name, f.Params, b.udfStackDepth)
 	// We store compiled UDF calls here so as to avoid stack overflows on
 	// recursive calls.
 	b.compiledUDFs[name] = fn
 	var err error
-	if fn.Body, err = b.compileExpr(body); err != nil {
+	if fn.Body, err = b.compileExpr(f.Expr); err != nil {
 		return nil, err
 	}
 	delete(b.compiledUDFs, name)
--- a/compiler/kernel/op.go
+++ b/compiler/kernel/op.go
@@ -52,16 +52,17 @@ import (
 var ErrJoinParents = errors.New("join requires two upstream parallel query paths")
 
 type Builder struct {
-	rctx         *runtime.Context
-	mctx         *super.Context
-	env          *exec.Environment
-	readers      []zio.Reader
-	progress     *zbuf.Progress
-	channels     map[string][]zbuf.Puller
-	deletes      *sync.Map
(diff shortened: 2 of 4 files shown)


compiler/kernel/op.go
```diff
@@ -52,16 +53,17 @@ import (
...
-	compiledUDFs map[string]*expr.UDF
-	resetters    expr.Resetters
+	rctx                *runtime.Context
+	mctx                *super.Context
+	env                 *exec.Environment
+	readers             []zio.Reader
+	progress            *zbuf.Progress
+	channels            map[string][]zbuf.Puller
+	deletes             *sync.Map
+	udfs                map[string]*udf
```
nwt: No need for type udf.  Just use dag.Func.

```suggestion
	udfs                map[string]*dag.Func
```

compiler/kernel/expr.go
```diff
@@ -342,16 +343,20 @@ func (b *Builder) compileCall(call dag.Call) (expr.Evaluator, error) {
...
+func (b *Builder) compileUDFCall(name string, f *dag.Func) (expr.Function, error) {
 	if fn, ok := b.compiledUDFs[name]; ok {
 		return fn, nil
 	}
-	fn := &expr.UDF{Name: name, Sctx: b.sctx()}
+	if b.udfStackSizeCounter == nil {
+		b.udfStackSizeCounter = new(atomic.Int64)
+		defer func() { b.udfStackSizeCounter = nil }()
+	}
+	fn := expr.NewUDF(b.sctx(), name, f.Params, b.udfStackSizeCounter)
```
nwt: Sharing this counter across goroutines is surprising and doesn't feel like the right move.

mattnibs: Open to suggestions. We are trying to get rid of using expr.Context which will give you use a solution that will also work in vector runtime. Since we are in expression land this shouldn't normally go across go-routines with the exception in some weird cases of unnest expressions and probably some less weird cases when we add subquery expressions (am I'm missing something here?). I don't think our current approach works in those circumstances either because once you go out of expression land into operator land and back the frame expr.Context is carried across these boundaries.

nwt: > Open to suggestions. We are trying to get rid of using expr.Context which will give you use a solution that will also work in vector runtime. Since we are in expression land this shouldn't normally go across go-routines with the exception in some weird cases of unnest expressions and probably some less weird cases when we add subquery expressions (am I'm missing something here?).

I now have a better understanding of what's going on here and think it's fine because an expr.Evaluator should actually never be shared across goroutines. And given that, there's no need for a *atomic.Int64. A *int suffices.

compiler/kernel/op.go
```diff
@@ -52,16 +53,17 @@ import (
...
+	rctx                *runtime.Context
+	mctx                *super.Context
+	env                 *exec.Environment
+	readers             []zio.Reader
+	progress            *zbuf.Progress
+	channels            map[string][]zbuf.Puller
+	deletes             *sync.Map
+	udfs                map[string]*dag.Func
+	compiledUDFs        map[string]*expr.UDF
+	udfStackSizeCounter *atomic.Int64
```
nwt: Nit: I think udfStackDepth would be a better name.

runtime/sam/expr/udf.go
```diff
@@ -1,43 +1,44 @@
...
 
 type UDF struct {
-	Body Evaluator
-	Name string
-	Sctx *super.Context
+	Body      Evaluator
+	sctx      *super.Context
+	name      string
+	fields    []super.Field
+	stackSize *atomic.Int64
```
nwt: Nit: I think stackDepth would be a better name.

runtime/sam/expr/udf.go
```diff
@@ -1,43 +1,43 @@
...
 }
 
-func (u *UDF) Call(ectx super.Allocator, args []super.Value) super.Value {
-	stack := 1
-	if f, ok := ectx.(*frame); ok {
-		stack += f.stack
-	}
-	if stack > maxStackDepth {
-		return u.Sctx.NewErrorf("stack overflow in function %q", u.Name)
+func NewUDF(sctx *super.Context, name string, params []string, stackSize *int) *UDF {
```
nwt:
```suggestion
func NewUDF(sctx *super.Context, name string, params []string, stackDepth *int) *UDF {
```

runtime/sam/expr/udf.go
```diff
@@ -1,43 +1,43 @@
...
+		return u.sctx.NewErrorf("stack overflow in function %q", u.name)
+	}
+	defer func() { *u.stackDepth-- }()
+	if len(args) == 0 {
+		return u.Body.Eval(NewContext(), super.Null)
+	}
+	u.builder.Reset()
+	for i := range args {
+		u.fields[i].Type = args[i].Type()
+		u.builder.Append(args[i].Bytes())
```
nwt: Nit: Maybe make this a little more concise?

```suggestion
	for i, a := range args {
		u.fields[i].Type = a.Type()
		u.builder.Append(a.Bytes())
```

nwt (approved):
Nice!

PR #6045, opened 2025-07-17
Remove dag.Var and expr.Context

mattnibs:
This commit removes the no longer needed dag.Var as well as expr.Context required in calls to expr.Evaluator which are no longer needed if dag.Var does not exist.


compiler/semantic/scope.go
```diff
@@ -34,19 +34,6 @@ type cte struct {
 	schema schema
 }
 
-func (s *Scope) DefineVar(name *ast.ID) error {
-	ref := &dag.Var{
-		Kind: "Var",
-		Name: name.Name,
-		Slot: s.nvars(),
```
nwt, line 41: Remove Scope.nvar and Scope.nvars too.

zbuf/batch.go
```diff
@@ -29,8 +29,6 @@ type Batch interface {
 	Ref()
 	Unref()
 	Values() []super.Value
-	// Vars accesses the variables reachable in the current scope.
-	Vars() []super.Value
```
nwt, line 33: Remove zbuf.pullerBatch.Vars, zbuf.EndOfChannel.Vars, and zio/bsupio.batch.Vars too.

nwt (approved):
You truly love to see it.

PR #6052, opened 2025-07-18
replace sequential join with vector join

nwt:
Remove the sequential join implementation and use the vector join implementation when compiling a join for the sequential runtime.  This frees us from maintaining two join implementations.

Changes to tests and to docs/tutorials/join.md account for the vector join operator's undefined output order.


mattnibs (approved):
I think we can also get rid of the test `runtime/ztests/op/join-auto-sort.yaml` since its no longer relevant...

nwt:
> I think we can also get rid of the test `runtime/ztests/op/join-auto-sort.yaml` since its no longer relevant...

Good call. Done in c3d314f5913e65c05b744aabb11d7eb0e67defb3.

PR #6054, opened 2025-07-18
remove semVars since it is now unused

mccanne:


mattnibs:
We got a pr removing this and much more: https://github.com/brimdata/super/pull/6045

mccanne:
Redundant with #6045

PR #6062, opened 2025-07-24
filter nulls and errors on input to shapes

mccanne:


compiler/semantic/op.go
```diff
@@ -905,9 +905,19 @@ func (a *analyzer) semOp(o ast.Op, seq dag.Seq) dag.Seq {
...
 			e = a.semExpr(o.Expr)
-		}
+			copy = a.semExpr(o.Expr)
+		}
+		seq = append(seq, &dag.Filter{
+			Kind: "Filter",
+			Expr: &dag.UnaryExpr{
+				Kind:    "UnaryExpr",
+				Op:      "!",
+				Operand: &dag.IsNullExpr{Kind: "IsNullExpr", Expr: copy},
```
nwt: Nit: You don't really need to copy `e` here.

```suggestion
				Operand: &dag.IsNullExpr{Kind: "IsNullExpr", Expr: e},
```

runtime/ztests/op/shapes-null.yaml
```diff
@@ -0,0 +1,16 @@
...
+  {s1:"a",s2:0}
+  {s1:"a",s2:null::string}
+  {s1:"a",s2:null::{x:int64}}
+  {s1:"a"}
+  {s1:"a",s2:error(1)}
+  {s1:"a",s2:error({x:1})}
+
+output: |
+  0
+  "b"
```
nwt: Nit: Since this test is just shapes-expr.yaml with some additional input values, add the additional values to that test instead of adding a new test.

nwt (approved):

PR #6063, opened 2025-07-24
use file extension to imply format and avoid auto-detection

mccanne:
This commit changes heuristic for detecting input file (or URL) formats by first checking for known file extensions when the input path is known.  Only when this heuristic fails is autodetection employed.  This was already being done for parquet and csup files (not URLs) and this commit generalizes the approach to all support file types and includes the check for URLs.

When files are gzipped, this test fails and the autodetection logic detects the gzip and the enclosing format.  We might want to extend this approach here to gzipped files in the future, e.g., where ".json.gz" implies gzipped JSON without having to do autodetection.

### DIFF
--- a/zio/zio.go
+++ b/zio/zio.go
@@ -3,6 +3,7 @@ package zio
 import (
 	"context"
 	"io"
+	"path/filepath"
 	"slices"
 
 	"github.com/brimdata/super"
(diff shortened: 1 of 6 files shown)


zio/zio.go
```diff
@@ -35,6 +36,29 @@ func Extension(format string) string {
...
+	case ".json", ".ndjson", ".jsonl":
+		return "json"
+	case ".parquet":
+		return "parquet"
+	case ".sup":
+		return "sup"
+	case ".txt":
+		return "text"
+	case ".jsup":
+		return "jsup"
```
nwt: Nit:
```suggestion
	case ".json", ".jsonl", ".ndjson":
		return "json"
	case ".jsup":
		return "jsup"
	case ".parquet":
		return "parquet"
	case ".sup":
		return "sup"
	case ".text", ".txt":
		return "text"
```

nwt (approved):

PR #6070, opened 2025-07-28
Query expressions

mattnibs:
The commit adds limited support for query expressions. Typical query expressions not covered here:
- IN, EXIST, etc. expressions for comparing against multiple values in query expressions.
- Correlated subqueries.

### DIFF
--- a/compiler/optimizer/demand.go
+++ b/compiler/optimizer/demand.go
@@ -190,6 +190,8 @@ func demandForExpr(expr dag.Expr) demand.Demand {
 			d = demand.Union(d, demandForExpr(e.Value))
 		}
 		return d
+	case *dag.QueryExpr:
+		return demand.Union(DemandForSeq(expr.Body, demand.All())...)
 	case *dag.RecordExpr:
 		d := demand.None()
 		for _, e := range expr.Elems {
--- a/compiler/semantic/expr.go
+++ b/compiler/semantic/expr.go
@@ -175,6 +175,8 @@ func (a *analyzer) semExpr(e ast.Expr) dag.Expr {
 			Kind:  "Literal",
 			Value: sup.FormatValue(val),
 		}
+	case *ast.QueryExpr:
+		return a.semQueryExpr(e)
 	case *ast.RecordExpr:
 		fields := map[string]struct{}{}
 		var out []dag.RecordElem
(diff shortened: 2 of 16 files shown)


compiler/optimizer/demand.go
```diff
@@ -214,6 +214,12 @@ func demandForExpr(expr dag.Expr) demand.Demand {
 	case *dag.SliceExpr:
 		return demand.Union(demandForExpr(expr.Expr),
 			demand.Union(demandForExpr(expr.From), demandForExpr(expr.To)))
+	case *dag.QueryExpr:
+		d := demand.None()
+		for _, seq := range DemandForSeq(expr.Body, demand.None()) {
+			d = demand.Union(d, seq)
+		}
+		return d
```
nwt: You need demand.All instead of demand.None since you want every field produced by expr.Body.

And demand.Union is variadic.

And put this after dag.MapExpr.

```suggestion
	case *dag.QueryExpr:
		return demand.Union(DemandForSeq(expr.Body, demand.All())...)
```

compiler/semantic/expr.go
```diff
@@ -408,6 +408,8 @@ func (a *analyzer) semExpr(e ast.Expr) dag.Expr {
 			Expr: expr,
 			Body: a.semSeq(e.Body),
 		}
+	case *ast.QueryExpr:
```
nwt: Nit: Put this after ast.Primitive.

compiler/semantic/expr.go
```diff
@@ -992,6 +994,28 @@ func (a *analyzer) semFString(f *ast.FString) dag.Expr {
 	return out
 }
 
+func (a *analyzer) semQueryExpr(q *ast.QueryExpr) dag.Expr {
+	e := &dag.QueryExpr{
+		Kind: "QueryExpr",
+		Body: a.semSeq(q.Body),
+	}
+	// XXX We should probably add a check to return error on nonsensical
+	// subqueries like (head 1)?
```
nwt: Is `head 1` really nonsensical since it will produce output (from the dag.NullScan added below)?

nwt (approved):

compiler/semantic/expr.go
```diff
@@ -992,6 +994,28 @@ func (a *analyzer) semFString(f *ast.FString) dag.Expr {
...
+	if isSQLOp(q.Body[0]) {
+		// SQL expects a record with a single column result so fetch the first
+		// value.
+		return &dag.IndexExpr{
+			Kind:  "IndexExpr",
+			Expr:  e,
+			Index: &dag.Literal{Kind: "Literal", Value: "1"},
+		}
+	}
+	return e
```
nwt: Nit: Indent less.

```suggestion
	if !isSQLOp(q.Body[0]) {
    		return e
    	}
	// SQL expects a record with a single column result so fetch the first
	// value.
	return &dag.IndexExpr{
		Kind:  "IndexExpr",
		Expr:  e,
		Index: &dag.Literal{Kind: "Literal", Value: "1"},
	}
```

compiler/semantic/expr.go
mattnibs: I believe any subquery without a source (including values) is non-sensical

mattnibs: We'll just let this be for now.

runtime/sam/op/traverse/queryexpr.go
```diff
@@ -0,0 +1,39 @@
...
+func NewQueryExpr(rctx *runtime.Context, puller zbuf.Puller) *QueryExpr {
+	return &QueryExpr{rctx: rctx, puller: puller}
+}
+
+func (q *QueryExpr) Eval(this super.Value) super.Value {
+	if q.cached == nil {
+		q.cached = q.exec().Ptr()
+	}
+	return *q.cached
+
```
nwt:
```suggestion
```

runtime/vam/expr/queryexpr.go
```diff
@@ -0,0 +1,94 @@
...
+		if this.Len() > 1 {
+			// This is an array so just create a view that repeats this.Len().
+			return vector.Pick(vec, make([]uint32, this.Len()))
+		}
+		return vec
+	}
+
+}
+
+func (q *QueryExpr) exec(this vector.Any) vector.Any {
```
nwt: Nit: We don't need really need `this` here.

```suggestion
func (q *QueryExpr) exec(length uint32) vector.Any {
```

PR #6072, opened 2025-07-29
unify and simplify field assignments

mccanne:
This commit makes field assignments and the optionality of "name:=..." a consistent element of the SuperSQL syntax. This means the put operator now works this way. We updated the semantic pass to always derive a name for all missing LHS using zfmt.  Otherwise the semantic logic that analyzed assignments in various operators was already handling the general case of a missing/null LHS so there were no changes needed here.


compiler/semantic/ztests/put-expr.yaml
```diff
@@ -0,0 +1,9 @@
+spq: put 1+1
```
nwt: Nit: Most (all?) of the other put operation tests are in runtime/ztests/op/.

compiler/semantic/expr.go
```diff
@@ -820,28 +821,29 @@ func isLval(e dag.Expr) bool {
 	return false
 }
 
-func deriveLHSPath(rhs dag.Expr) ([]string, error) {
-	switch rhs := rhs.(type) {
+func deriveNameFromExpr(e dag.Expr, a ast.Expr) ([]string, error) {
```
nwt: Drop the error return parameter since it's always nil now.

nwt (approved):

PR #6079, opened 2025-07-30
zfmt: add support for SQL AST nodes

### LINKED ISSUE #6073: Error messages appearing in field names
Repro is with super commit cc6b1e6.

Thanks to the initial subquery support added in #6070, the simple subquery shown in #6013 produces the correct result now, but the field name happens to contain what looks like an error message.

```
$ super -version
Version: cc6b1e6c2

$ super -c "SELECT 1+(SELECT 1);"
{"1+(\n  unknown operator: *ast.Select\n)":2}
```

As a user, I initially reacted to this as if I was seeing a bug with the subquery support, but @mattnibs educated me that this is an unrelated symptom due to lack of support for `ast.Select` in `zfmt.Expr`.

It's not clear to me if this is a one-off or if this is a pattern that's likely to come up repeatedly in the future such that we'd need to whack-a-mole such symptoms each time they surface.

nwt:
Depends on #6077 and #6078.

Fixes #6073

### DIFF
--- a/zfmt/ztests/sql.yaml
+++ b/zfmt/ztests/sql.yaml
@@ -1,6 +1,4 @@
 script: |
-  echo === expressions
-  super compile -C "select cast(a as b), substring(c from d for e), date 'f', timestamp 'g'"
   echo === select
   super compile -C 'select a from b where c group by d having e order by f limit g offset h'
   echo === ornate select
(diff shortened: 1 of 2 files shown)


philrz:
Looks like a good improvement on what's shown in #6073. Curious why the extra spaces and `\n` are in there.

```
$ super -version
Version: 8fe64aa83

$ super -c "SELECT 1+(SELECT 1);"
{"1+(  select 1\n)":2}
```

zfmt/ztests/sql.yaml
```diff
@@ -0,0 +1,60 @@
...
+      )
+      select c
+      === ornate with
+      with recursive a as (
+        select b
+        from c
+      ), d as (
+        select e
+      )
+      select f
```
mattnibs, line 60: I think it would be nice to have a test with a query expression.

nwt: Indeed.

nwt: @mattnibs: Added in 04eb18f9fd4801634f25118274cdcfe799b18350. How's it look?

PR #6083, opened 2025-07-31
Enable unnest expressions in vector runtime

mattnibs:


mattnibs:
I spent a fair amount of time trying to get this to work natively in vector runtime but all approaches led me to relying on heavy use of vector.DynamicBuilder which led me to this approach. Perhaps we can reevaluate the approach later down the line.

PR #6095, opened 2025-08-04
zfmt: add support for ast.SQLJoin

nwt:


zfmt/ztests/sql.yaml
```diff
@@ -19,6 +23,12 @@ script: |
 outputs:
   - name: stdout
     data: |
+      === join with on condition
+      select a
+      from b inner join c on d
```
mattnibs: Shouldn't the inner join statement be on a new line?

nwt: It certainly could be.

Should we do the same for cross joins?  They'd look like this:

```
select a
from b
cross join c
```

mattnibs: Yeah I think so

mattnibs (approved):

PR #6097, opened 2025-08-05
compiler/semantic: create static schema for Parquet file scan

nwt:


philrz:
Here's a summary of the user-facing benefits of this change.

Given the following Parquet data as input:

```
$ super -s foo.parquet 
{foo:"bar"}
```

First, here's the baseline behavior as of super commit 97818e4 right before the changes in this PR were merged. An attempt to access a field not present in the file would result in an `error("missing")`.

```
$ super -version
Version: 97818e43e

$ super -c "SELECT foo,hello FROM 'foo.parquet';"
{foo:"bar",hello:error("missing")}
```

Then at super commit 0f83358 which is right after the changes merged from this PR, now an error message is returned immediately informing the user that the `SELECT` will fail to find a value at the referenced field.

```
$ super -version
Version: 0f83358ad

$ super -c "SELECT foo,hello FROM 'foo.parquet';"
column "hello": does not exist at line 1, column 12:
SELECT foo,hello FROM 'foo.parquet';
           ~~~~~
```

This is in line with what's typical of the relational model as seen in other SQL implementations, e.g.,

```
$ duckdb --version
v1.3.2 (Ossivalis) 0b83e5d2f6

$ duckdb -c "SELECT foo,hello FROM 'foo.parquet';"
Binder Error:
Referenced column "hello" not found in FROM clause!
Candidate bindings: "foo"

LINE 1: SELECT foo,hello FROM 'foo.parquet';
                   ^
```

PR #6099, opened 2025-08-05
zfmt: split SQL cross join across multiple lines

nwt:
This matches the output for ast.SQLJoin.

The tests removed from zfmt/ztests/join.yaml provide no additional coverage beyond that of the test moved to zfmt/ztests/sql.yaml.

### DIFF
--- a/zfmt/ztests/sql.yaml
+++ b/zfmt/ztests/sql.yaml
@@ -1,4 +1,6 @@
 script: |
+  echo === cross join
+  super compile -C 'from a cross join b, c'
   echo === join with on condition
   super compile -C 'from a join b on c'
   echo === join with using condition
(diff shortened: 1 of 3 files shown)


zfmt/ztests/sql.yaml
```diff
@@ -23,6 +25,10 @@ script: |
 outputs:
   - name: stdout
     data: |
+      === cross join
+      from a
+      cross join b
+      cross join c
```
mattnibs, line 31: In the ast we parse `cross join b,c` as two separate nodes?

nwt: Exactly. SQL has the two syntaxes for cross join (`from a cross join b` and `from a, b` are equivalent) so `from a cross join b, c` is a composition of two cross joins and becomes two AST nodes.

mattnibs: interesting, good to know

mattnibs (approved):

PR #6100, opened 2025-08-05
First pass at correlated subqueries

mattnibs:
This commit allow use of correlated subqueries for pipe syntax. Correlated subqueries using the outer scope for SQL queries remains unsupported.


compiler/parser/ztests/unnest-expr.yaml
```diff
@@ -1,14 +1,14 @@
 script: |
   echo === aggregation
-  super compile -C 'collect(unnest {a:b,c} | ?d)'
+  super compile -C 'collect((unnest {a:b,c} | ?d))'
```
nwt, line 3: The need for double parentheses here is a bringdown. Did you already burn time looking for a way to avoid it?

compiler/rungen/vexpr.go
```diff
@@ -300,17 +297,19 @@ func (b *Builder) compileVamRecordExpr(e *dag.RecordExpr) (vamexpr.Evaluator, er
 }
 
 func (b *Builder) compileVamQueryExpr(query *dag.QueryExpr) (vamexpr.Evaluator, error) {
-	exits, err := b.compileVamSeq(query.Body, nil)
+	eval, err := b.compileQueryExpr(query)
```
nwt: Nit: We usually call an Evaluator `e` or `somethingExpr`.

runtime/ztests/expr/queryexpr-multiple-err.yaml
```diff
@@ -0,0 +1,6 @@
+spq: values (values 1,2,3)
+
+vector: true
+
+output: |
+  error({message:"encountered 3 values in expression subquery (consider collect())",on:1})
```
nwt: This message doesn't make sense. It says there are 3 values but the `on` field only contains a single value.

compiler/parser/ztests/unnest-expr.yaml
mattnibs: This was what @mccanne did with the parsing logic. I suppose we could add back the `Expr / UnnestExpr` to function calls and just change it to `Expr / QueryExpr`, but I'd like to hear from @mccanne

compiler/rungen/vexpr.go
mattnibs: ok

runtime/ztests/expr/queryexpr-multiple-err.yaml
mattnibs: Fair

mccanne: Could change `on` to `at`.  It's even more confusing if you wrap the multi-value in a single-value array and say it encountered multiple values when on is a single value.

mattnibs: I'm just going to get rid of the wrapped values and return an error string.

mattnibs: Or we can do the `at` as you suggest @mccanne either way.

runtime/sam/op/traverse/queryexpr.go
```diff
@@ -1,38 +1,115 @@
...
+	"context"
+
 	"github.com/brimdata/super"
 	"github.com/brimdata/super/runtime"
+	"github.com/brimdata/super/runtime/sam/expr"
 	"github.com/brimdata/super/zbuf"
 )
 
-type QueryExpr struct {
+type Cache struct {
```
nwt: Nit: I think CachedQueryExpr would be a better name.

runtime/sam/op/traverse/queryexpr.go
```diff
@@ -1,38 +1,115 @@
...
 }
+
+// QueryExpr is a simple subquery mechanism where it has both an Eval
+// method to implement expressions and a Pull method to act as the parent
+// of a subgraph that is embedded in an expression.  Whenever eval
+// is called, it constructs a single valued batch using the passed-in
+// this, posts that batch to the embedded query, then pulls from the
+// query until eos.  When the subquery is not correlated (i.e., because
+// of a new from clause or perhaps a constant values clase) then it wraps
+// the from in a cache that always returns the same batch.  Subqueries always
```
nwt: Nit: This looks out of date.

```suggestion
// query until eos.  Subqueries always
```

runtime/sam/op/traverse/queryexpr.go
```diff
@@ -1,38 +1,115 @@
...
+		return batch, nil
+	case <-q.ctx.Done():
+		return nil, q.ctx.Err()
+	}
+}
+
+func (q *QueryExpr) Eval(this super.Value) super.Value {
+	b := zbuf.NewArray([]super.Value{this})
+	select {
+	case q.batchCh <- b:
```
nwt: Nit:

```suggestion
	select {
	case q.batchCh <- zbuf.NewArray([]super.Value{this}):
``````

File: runtime/sam/op/traverse/queryexpr.go [hunk 7]
```diff
@@ -1,38 +1,115 @@
...
+		if b == nil {
+			if count > 1 {
+				return q.sctx.NewErrorf("encountered %d values in expression subquery (consider collect())", count)
+			}
+			return val
+		}
+		if count == 0 {
+			val = b.Values()[0].Copy()
+		}
+		count += len(b.Values())
```
nwt, line 105:
```suggestion
		count += len(b.Values())
		b.Unref()
```

runtime/sam/op/traverse/queryexpr.go
```diff
@@ -1,38 +1,115 @@
...
+	val := super.Null
+	var count int
+	for {
+		b, err := q.body.Pull(false)
+		if err != nil {
+			panic(err)
+		}
+		if b == nil {
+			if count > 1 {
+				return q.sctx.NewErrorf("encountered %d values in expression subquery (consider collect())", count)
```
nwt: Nits: "encountered %d values" doesn't really suggest that at most one value is expected. And it would be nice to use consistent terminology ("query expression" or "expression subquery" or whatever) across the code and this error.

```suggestion
				return q.sctx.NewErrorf("query expression produced multiple values (consider collect())")
```

runtime/sam/op/traverse/queryexpr.go
```diff
@@ -1,38 +1,115 @@
...
+// return single values expecting multi-valued results to be wrapped in a collect.
+// If more than one value is returned, then a structured error results.
+type QueryExpr struct {
+	ctx     context.Context
+	sctx    *super.Context
+	batchCh chan zbuf.Batch
+	eos     bool
+
+	body     zbuf.Puller
+	resetter expr.Resetter //XXX why this?
```
nwt: Nit: No point including this if we're not going to use it.

```suggestion
```

runtime/ztests/expr/from-eval-subquery.yaml
```diff
@@ -0,0 +1,22 @@
+script: super -s -I query.spq
```
nwt: Nit: Call this querexpr-from-eval.yaml so it's grouped with the other queryepxr-* tests.

nwt (approved):

PR #6101, opened 2025-08-06
Add support for exists() subqueries

mattnibs:


compiler/semantic/expr.go
```diff
@@ -996,17 +1020,23 @@ func (a *analyzer) semQueryExpr(q *ast.QueryExpr) dag.Expr {
...
+		// value.
+		// XXX this should be a structured error... or just allow it
+		// SQL expects a record with a single column result so fetch the first
+		// value.  Or we should be descoping.
+		e.Body.Append(&dag.Values{
+			Kind: "Values",
+			Exprs: []dag.Expr{
+				&dag.IndexExpr{
+					Kind:  "IndexExpr",
+					Expr:  &dag.This{},
```
```suggestion
					Expr:  &dag.This{Kind: "This"},
```

compiler/semantic/expr.go
```diff
@@ -436,6 +438,28 @@ func (a *analyzer) semDoubleQuote(d *ast.DoubleQuote) dag.Expr {
 	})
 }
 
+func (a *analyzer) semExists(e *ast.Exists) dag.Expr {
+	q := a.semSeqExpr(e.Body)
+	// Append collect(this) to ensure array of results is returned.
+	q.Body = append(q.Body,
+		&dag.Aggregate{
```
nwt: We only need to see one value.

```suggestion
	q.Body = append(q.Body,
		&dag.Head{Kind: "Head", Count: 1},
		&dag.Aggregate{
```

mattnibs: Nice

PR #6102, opened 2025-08-06
Full support for "in" subquery expressions

mattnibs:


runtime/ztests/expr/in-subquery.yaml
```diff
@@ -0,0 +1,17 @@
...
+vector: true
+
+outputs:
+  - name: stdout
+    data: |
+      true
+      // ===
+      true
+      // ===
+      false
```
nwt: Nit: Make this a spq-style test.

```suggestion
spq: |
  values
    "foo" in (values "foo", "bar", "baz"),
    "===",
    "foo" not in (values "fool"),
    "===",
    "foo" in (values "fool")

vector: true

output: |
  true
  "==="
  true
  "==="
  false
```

nwt (approved):

PR #6105, opened 2025-08-08
book: always rebuild src/super.wasm when making "all"

nwt:
It's fast since it's just "go build".

### DIFF
--- a/book/README.md
+++ b/book/README.md
@@ -10,7 +10,7 @@ brew install mdbook
 ```
 
 The easiest way to work on docs is to run an mdbook service in this directory
-and point your browser at its embedded web server, e.g., 
+and point your browser at its embedded web server, e.g.,
 ```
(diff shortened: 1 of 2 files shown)


book/README.md
```diff
@@ -20,17 +20,16 @@ Then connect to localhost:3000.
...
 and build the book manually like this:
 ```
 mdbook build
 ```
-This was the service doesn't do things like recreating a file that you have 
+This way the service doesn't do things like recreating a file that you have
 removed when you are trying to rearrange things.
 
-After editing JavaScript or Go files in `super-example` run `make`.
+After editing any JavaScript or Go files run `make` in this directory.
```
mccanne, line 34: Add something like: "This will update the wasm binary so that playground examples use the version of SuperDB built against the Go source files in the local repository."

PR #6108, opened 2025-08-13
Add subquery schema

mattnibs:
The commit adds a subquery schema to the semantic analyzer that currently just returns an unsupported error when a subquery is correlated.

### DIFF
--- a/compiler/semantic/schema.go
+++ b/compiler/semantic/schema.go
@@ -43,10 +43,16 @@ type joinUsingSchema struct {
 	*joinSchema
 }
 
+type subquerySchema struct {
+	outer schema
+	inner schema
+}
+
 func (s *staticSchema) Name() string  { return s.name }
 func (d *dynamicSchema) Name() string { return d.name }
 func (*selectSchema) Name() string    { return "" }
 func (*joinSchema) Name() string      { return "" }
+func (*subquerySchema) Name() string  { return "" }
 
 func badSchema() schema {
 	return &dynamicSchema{}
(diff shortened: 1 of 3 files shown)


compiler/semantic/schema.go
```diff
@@ -217,6 +250,10 @@ func (j *joinUsingSchema) resolveOrdinal(col int) (dag.Expr, error) {
 	return nil, errors.New("ordinal column selection in join not supported")
 }
 
+func (s *subquerySchema) resolveOrdinal(col int) (dag.Expr, error) {
+	return nil, errors.New("ordinal column selection in subqueries not supported")
```
nwt: Nit: Be consistent with joinSchema.resolveOrdinal.

```suggestion
	return nil, errors.New("ordinal column selection in subquery not supported")
```

compiler/semantic/schema.go
```diff
@@ -293,6 +330,10 @@ func joinSpread(left, right dag.Expr) *dag.RecordExpr {
 	}
 }
 
+func (s *subquerySchema) deref(name string) (dag.Expr, schema) {
+	panic(name) // This shouldn't happen.
```
nwt: Nit: The panic() makes this obvious.

```suggestion
	panic(name)
```

compiler/ztests/sql/subqueries-correlated.yaml
```diff
@@ -0,0 +1,62 @@
...
+inputs:
+  - name: a.sup
+    data: |
+      {colA:1,colB:3}
+      {colA:2,colB:4}
+  - name: b.sup
+    data: |
+      {colA:3,colB:1}
+      {colA:4,colB:2}
+
```
nwt: These are unused.

```suggestion
```

compiler/semantic/schema.go
mattnibs: Fair

PR #6112, opened 2025-08-15
add r-strings

### LINKED ISSUE #4917: Passing regexp or glob into a User Op
## tl;dr

These don't currently work.

```
$ echo '"foo" "bar" "baz"' | zq -Z 'op foo(pattern): ( grep(pattern) ) foo(*a*)' -
zq: error parsing Zed at column 40:
op foo(pattern): ( grep(pattern) ) foo(*a*)
                                   === ^ ===

$ echo '"foo" "bar" "baz"' | zq -Z 'op foo(pattern): ( grep(pattern) ) foo(/a/)' -
zq: error parsing Zed at column 40:
op foo(pattern): ( grep(pattern) ) foo(/a/)
                                   === ^ ===
```

## Details

Repro is with Zed commit 40739bd.

This was discussed in the context of #4692 where it became possible to pass a string into a User Op and then use that as an argument to `grep()`. So this works now where it didn't before:

```
$ zq -version
Version: v1.11.1-16-g40739bdc

$ echo '"foo" "bar" "baz"' | zq -Z 'op foo(pattern): ( grep(pattern) ) foo("bar")' -
"bar"
```

However, the glob and regexp variations shown in the opening tl;dr text above still don't work yet. I recently got @mccanne on-record saying we should probably make them work at some point, hence this issue. 😉

mccanne:
Add raw strings where a string prefixed with an r doesn't interpret backslashes as escapes.  This is useful for strings that are regular expressions.  Docs updates forthcoming from the book branch.

Partially fixes #4917

### DIFF
--- a/compiler/parser/ztests/rstring.yaml
+++ b/compiler/parser/ztests/rstring.yaml
@@ -3,9 +3,6 @@ spq: |
 
 vector: true
 
-input: |
-  null
-
 output: |
   ""
   ""


compiler/parser/ztests/rstring.yaml
```diff
@@ -0,0 +1,15 @@
+spq: |
+  values r'', r"", r'foo', r'\f\o\o\n', r"bar", r"\b\a\r\n"
+
+vector: true
+
+input: |
+  null
+
```
nwt: Nit: Don't need this.

```suggestion
```

PR #6115, opened 2025-08-15
simplify regexp/grep syntax

mccanne:
This commit cleans up the regexp/grep syntax by isolating the regexp pattern and glob syntax to search expressions and using conventional strings for the functions regexp(), regexp_replace(), and grep().  This allowed us to remove a bunch of special case syntax for these functions from the PEG grammar.

This change means that globbing is no longer available for grep, which now uses only regular expressions.  If we want to allow globbing, we can enable that with a function that converts glob strings to regexps or perhaps introduce a g"glob" style string.

We also removed the implied "this" argument from grep and now require two arguments.

Detailed docs updates are forthcoming in the book branch.

### DIFF
--- a/docs/language/functions/grep.md
+++ b/docs/language/functions/grep.md
@@ -5,7 +5,7 @@
 ### Synopsis
 
 ```
-grep(<pattern> [, e: any]) -> bool
+grep(<pattern>, e: any) -> bool
 ```
 
 ### Description
@@ -36,7 +36,7 @@ if a complex type,
 _Reach into nested records_
 ```mdtest-spq
 # spq
-grep("baz")
+grep("baz", this)
 # input
 {foo:10}
 {bar:{s:"baz"}}
@@ -47,7 +47,7 @@ grep("baz")
 _It only matches string fields_
 ```mdtest-spq
 # spq
-grep("10")
+grep("10", this)
 # input
 {foo:10}
 {bar:{s:"baz"}}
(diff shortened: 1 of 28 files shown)


docs/language/functions/grep.md
```diff
@@ -5,7 +5,7 @@
 ### Synopsis
 
 ```
-grep(<pattern> [, e: any]) -> bool
+grep(<pattern>, e: any) -> bool
```
nwt, line 8: The first argument to grep() is now a string containing a regular expression, right?  If so, the description also needs an update.

```suggestion
grep(re: string, e: any) -> bool
```

docs/language/functions/regexp.md
```diff
@@ -5,7 +5,7 @@
 ### Synopsis
 
 ```
-regexp(re: string|regexp, s: string) -> any
+regexp(re: string, s: string) -> any
```
nwt, line 8: Description also needs an update.

docs/language/functions/grep.md
mccanne: Updated on book branch

docs/language/functions/regexp.md
mccanne: Updated in book branch

chrismo:
In the tests, I'm seeing `grep` without an r prefix on the strings, and `regexp_replace` with an r prefix on strings ... are those just two different options available anywhere, or a requirement of those functions? (Also, TIL: `regexp_search`?)

---

Ah, https://github.com/brimdata/super/pull/6112 - I assume then these are available anywhere. Cool. :)

PR #6118, opened 2025-08-15
Support SQL case-matching style expressions

### LINKED ISSUE #5979: SQL: CASE matching-style expressions
## tl;dr

```
$ super -c "SELECT CASE a WHEN 42 THEN 100 WHEN 43 THEN 200 ELSE 300 END FROM exprtest.sup;"
case matching-style expressions not yet supported at line 1, column 8:
SELECT CASE a WHEN 42 THEN 100 WHEN 43 THEN 200 ELSE 300 END FROM exprtest.sup;
       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

## Details

Repro is with super commit e75dfda. This was found via a query from [a sqllogictest](https://github.com/duckdb/duckdb/blob/049b11bfeedb3d6139ba10c2da3882aaf15222b0/test/sql/projection/test_complex_expressions.test#L39-L46).

This works in Postgres.

```
$ psql postgres
psql (17.5 (Homebrew))
Type "help" for help.

postgres=# SELECT * FROM exprtest;
 a  |  b  
----+-----
 42 |  10
 43 | 100
    |   1
 45 |  -1
(4 rows)

postgres=# SELECT CASE a WHEN 42 THEN 100 WHEN 43 THEN 200 ELSE 300 END FROM exprtest;
 case 
------
  100
  200
  300
  300
(4 rows)
```

But not yet in SuperDB.

```
$ cat exprtest.sup 
{a:42::int32,b:10::int32}
{a:43::int32,b:100::int32}
{a:null::int32,b:1::int32}
{a:45::int32,b:-1::int32}

$ super -version
Version: e75dfdaf1

$ super -c "SELECT CASE a WHEN 42 THEN 100 WHEN 43 THEN 200 ELSE 300 END FROM exprtest.sup;"
case matching-style expressions not yet supported at line 1, column 8:
SELECT CASE a WHEN 42 THEN 100 WHEN 43 THEN 200 ELSE 300 END FROM exprtest.sup;
       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

mattnibs:
Closes #5979


philrz:
This is a functional 👍 for me! Using a `super` built from this branch at commit cd28659, 1052 of the queries from the sqlite "select" suites that previously could not run at all due to the lack of this functionality now run and produce the expected result. 🎉

runtime/ztests/expr/case-match.yaml
```diff
@@ -0,0 +1,18 @@
+spq: 'values case x when 1 then "foo" when 2 then "bar" else {y:12} end'
```
nwt: Nit: No need to quote this.

```suggestion
spq: values case x when 1 then "foo" when 2 then "bar" else {y:12} end
```

compiler/semantic/expr.go
```diff
@@ -645,6 +644,29 @@ func (a *analyzer) semDotted(e *ast.BinaryExpr) ([]string, dag.Expr) {
...
+				RHS:  a.semExpr(when.Cond),
+			},
+			Then: a.semExpr(when.Then),
+			Else: cond,
+		}
+		if i == len(c.Whens)-1 {
+			cond.Else = a.semExprNullable(c.Else)
+		}
+	}
+	return cond
```
nwt: Nit: Simplify.

```suggestion
	out := a.semExprNullable(c.Else)
	for i := len(c.Whens) - 1; i >= 0; i-- {
		when := c.Whens[i]
		out = &dag.Conditional{
			Kind: "Conditional",
			Cond: &dag.BinaryExpr{
				Kind: "BinaryExpr",
				Op:   "==",
				LHS:  dag.CopyExpr(e),
				RHS:  a.semExpr(when.Cond),
			},
			Then: a.semExpr(when.Then),
			Else: out,
		}
	}
	return out
```

nwt (approved):

compiler/semantic/expr.go
mattnibs: Nice!

PR #6122, opened 2025-08-17
simplify text entity syntax

mccanne:
This commit simplifies the syntax rules for text entites by creating a new grammar rule for a "simple URL".  This makes usage and documentation clearer.

While here, we changed the ":meta" meta-modifier syntax to a "name" (the same rules as field names), where the it can be anything if quoted and otherwise must have identifier syntax.

We also eliminated the unused PEG rule PoolAt and simplified the rule for KSUID as it's now used in just one place (KSUID overlaps with TextChars so it's not needed in the Text rule).

Documentation changes are forthcoming in the book branch.

### DIFF
--- a/compiler/parser/parser.peg
+++ b/compiler/parser/parser.peg
@@ -701,7 +701,7 @@ Text
 SimpleURL =
   "http" "s"? "://"
   ([a-zA-Z0-9_-]+) ( "." [a-zA-Z0-9_-]+)*
-  ("/" TextChars)? { return string(c.text), nil }
+  ("/" TextChars?)? { return string(c.text), nil }
 
 TextChars
   = (IdentifierRest / "." / "/")+ { return string(c.text), nil }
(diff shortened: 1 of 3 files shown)


compiler/parser/parser.peg
```diff
@@ -697,13 +697,14 @@ FromEntity
...
       return &ast.Text{Kind:"Text", Text:s.(string), Loc:loc(c) }, nil
     }
 
-UnquotedURL = ("http://" / "https://" / "s3://") URLChar+ { return string(c.text), nil }
-
-URLChar = [0-9a-zA-Z!@$%&_=,./?:[\]~+-] //XXX fix this?
+SimpleURL =
+  "http" "s"? "://"
+  ([a-zA-Z0-9_-]+) ( "." [a-zA-Z0-9_-]+)*
+  ("/" TextChars)? { return string(c.text), nil }
```
nwt: This doesn't accept a path of "/".

```console
$ go run ./cmd/super compile -C 'from http://x/'
parse error at line 1, column 15:
from http://x/
          === ^ ===
exit status 1
```

```suggestion
  ("/" TextChars?)? { return string(c.text), nil }
```

PR #6127, opened 2025-08-18
Support out of order CTEs

### LINKED ISSUE #6082: SQL: Referencing a CTE before it's defined
## tl;dr

With the following query, right now SuperDB treats it as a failed attempt to read from a file. Other implementations either run the query ok or have a better error message, and we might consider one of those approaches as an improvement.

```
$ super -c "with cte3 as (select ref2.j as i from cte1 as ref2), cte1 as (Select i as j from a), cte2 as (select ref.j+1 as k from cte1 as ref) select * from cte2 union all select * FROM cte3;"
cte1: file does not exist
```

## Detail

Repro is with super commit fd1e48f.

The test data is simply:

```
$ super -version
Version: fd1e48f4f

$ super -s a
{i:42::int32}
```

Comparing with other implementations, Postgres reports an error that at first makes it sound like a failed table reference, but then it goes on to explain that it recognized it as an attempted CTE reference.

```
$ psql postgres
psql (17.5 (Homebrew))
Type "help" for help.

postgres=# create table a(i integer);
CREATE TABLE

postgres=# insert into a values (42);
INSERT 0 1

postgres=# with cte3 as (select ref2.j as i from cte1 as ref2), cte1 as (Select i as j from a), cte2 as (select ref.j+1 as k from cte1 as ref) select * from cte2 union all select * FROM cte3;

ERROR:  relation "cte1" does not exist
LINE 1: with cte3 as (select ref2.j as i from cte1 as ref2), cte1 as...
                                              ^
DETAIL:  There is a WITH item named "cte1", but it cannot be referenced from this part of the query.
HINT:  Use WITH RECURSIVE, or re-order the WITH items to remove forward references.
```

DuckDB handles the query without complaint.

```
$ duckdb test_cte.db 
DuckDB v1.3.2 (Ossivalis) 0b83e5d2f6

D SELECT * FROM a;
┌───────┐
│   i   │
│ int32 │
├───────┤
│  42   │
└───────┘

D with cte3 as (select ref2.j as i from cte1 as ref2), cte1 as (Select i as j from a), cte2 as (select ref.j+1 as k from cte1 as ref) select * from cte2 union all select * FROM cte3;
┌───────┐
│   k   │
│ int32 │
├───────┤
│    43 │
│    42 │
└───────┘
```

ClickHouse handles it ok too.

```
$ clickhouse --query "with cte3 as (select ref2.j as i from cte1 as ref2), cte1 as (Select i as j from a), cte2 as (select ref.j+1 as k from cte1 as ref) select * from cte2 union all select * FROM cte3;"
42
43
```

mattnibs:
Closes #6082


compiler/semantic/op.go
```diff
@@ -129,6 +125,21 @@ func (a *analyzer) semFromEntity(entity ast.FromEntity, alias *ast.TableAlias, a
 	}
 }
 
+func (a *analyzer) fromCTE(entity ast.FromEntity, c *cte, alias *ast.TableAlias) (dag.Seq, schema) {
```
nwt, line 128: Nit: You don't need an ast.FromEntity here, just an ast.Node.

```suggestion
func (a *analyzer) fromCTE(n ast.Node, c *cte, alias *ast.TableAlias) (dag.Seq, schema) {
```

compiler/semantic/analyzer.go
```diff
@@ -40,13 +40,14 @@ func Analyze(ctx context.Context, p *parser.AST, env *exec.Environment, extInput
...
-	files   *srcfiles.List
-	opStack []*ast.OpDecl
-	outputs map[*dag.Output]ast.Node
-	env     *exec.Environment
-	scope   *Scope
-	sctx    *super.Context
+	ctx      context.Context
+	files    *srcfiles.List
+	opStack  []*ast.OpDecl
+	cteStack []*ast.SQLCTE
```
nwt: Nit: I think what's going on here would be substantially clearer if this were a cte slice.

```suggestion
	cteStack []*cte
```

compiler/semantic/op.go
```diff
@@ -129,6 +125,21 @@ func (a *analyzer) semFromEntity(entity ast.FromEntity, alias *ast.TableAlias, a
...
+		return dag.Seq{badOp()}, badSchema()
+	}
+	a.cteStack = append(a.cteStack, c.ast)
+	body, inSchema := a.semSQLPipe(c.ast.Body, nil, &ast.TableAlias{Name: c.ast.Name.Name})
+	a.cteStack = a.cteStack[:len(a.cteStack)-1]
+	seq, outSchema, err := derefSchemaWithAlias(inSchema, alias, body)
+	if err != nil {
+		a.error(alias, err)
+	}
+	return seq, outSchema
```
nwt: Nit: inSchema and outSchema aren't very descriptive names so just say schema.

```suggestion
	body, schema := a.semSQLPipe(c.ast.Body, nil, &ast.TableAlias{Name: c.ast.Name.Name})
	a.cteStack = a.cteStack[:len(a.cteStack)-1]
	seq, schema, err := derefSchemaWithAlias(schema, alias, body)
	if err != nil {
		a.error(alias, err)
	}
	return seq, schema
```

compiler/ztests/sql/cte-out-of-order.yaml
```diff
@@ -0,0 +1,15 @@
+# Tests to ensure CTEs can be called out of order and still work.
+spq: |
+  with cte3 
+     as (select ref2.j as i
+         from cte1 as ref2),
+     cte1
+     as (select i as j from (values (42)) as x(i) ),
+     cte2
+     as (select ref.j+1 as k from cte1 as ref)
```
nwt: Nit: `super compile` formats CTEs like so, like the PostgreSQL documentation. Maybe do likewise?

```sql
with a as (
  select ...
), b as (
  select ...
), c as (
  select ...
)
select ...
```

compiler/ztests/sql/cte-recursive.yaml
```diff
@@ -0,0 +1,16 @@
...
+    SELECT x+1
+    FROM   t
+    WHERE  x < 200000
+  )
+  SELECT count(t1) FROM t AS t1;
+
+input: null
+
+error: |
+  recursive WITH relations not currently supported at line 5, column 10:
```
nwt, line 14: The real error here is that the CTE refers to itself but doesn't include the RECURSIVE keyword.

mattnibs: Yeah but we also don't support the RECURSIVE keyword so I wanted to avoid telling them to use the RECURSIVE keyword then have it not work.

compiler/ztests/sql/cte-out-of-order.yaml
mattnibs: I have yet to see a format that is pretty so I'll go with this

compiler/semantic/op.go
mattnibs: I'm fine with this but I don't think `n` is a good name for the variable, it should at least imply that this is supposed to be the FromEntity node.

nwt: Just a nit so do as you please.

nwt (approved):

PR #6130, opened 2025-08-19
rename QueryExpr to Subquery and delete traverse.Expr

mccanne:
This is a simple rename and removes unused code and doesn't otherwise change anything.

### DIFF
--- a/compiler/ast/ast.go
+++ b/compiler/ast/ast.go
@@ -303,7 +303,7 @@ type TupleExpr struct {
 	Loc   `json:"loc"`
 }
 
-type QueryExpr struct {
+type Subquery struct {
 	Kind string `json:"kind" unpack:""`
 	Body Seq    `json:"body"`
 	Loc  `json:"loc"`
@@ -372,7 +372,7 @@ func (*SQLTimeValue) exprNode() {}
 func (*FString) exprNode()      {}
 func (*Primitive) exprNode()    {}
 func (*TypeValue) exprNode()    {}
-func (*QueryExpr) exprNode()    {}
+func (*Subquery) exprNode()     {}
 
 type ConstDecl struct {
 	Kind string `json:"kind" unpack:""`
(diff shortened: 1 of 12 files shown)


compiler/ast/ast.go
```diff
@@ -303,7 +303,7 @@ type TupleExpr struct {
 	Loc   `json:"loc"`
 }
 
-type QueryExpr struct {
+type Subquery struct {
```
nwt, line 306: Can we call this SubqueryExpr to make clear that it's an expression?

mccanne: At it's core its an expression that wraps an operator so really both at the same time.  In the runtime, it implements both Evaluator and Op.    I realize in the AST it just implements exprNode but I think the bridge concept is clear enough that subquery can stand alone.

mccanne: I want to make another move to separate unnest, subquery, and lateral scope in the runtime but wanted to make this name change first.

compiler/dag/expr.go
```diff
@@ -83,7 +83,7 @@ type (
 		Kind    string  `json:"kind" unpack:""`
 		Entries []Entry `json:"entries"`
 	}
-	QueryExpr struct {
+	Subquery struct {
```
nwt: Nit: Move below SortExpr to keep these alphabetized.

compiler/optimizer/demand.go
```diff
@@ -190,7 +190,7 @@ func demandForExpr(expr dag.Expr) demand.Demand {
 			d = demand.Union(d, demandForExpr(e.Value))
 		}
 		return d
-	case *dag.QueryExpr:
+	case *dag.Subquery:
```
nwt: Nit: Move below SliceExpr to keep these alphabetized.

compiler/ast/ast.go
nwt: I can live with Subquery.

But while I get that in the runtime a subquery implements both Evaluator and Puller, in the AST and DAG they're strictly expressions. Yes, they wrap operators, but they aren't operators themselves!

PR #6139, opened 2025-08-20
change zio to sio

mccanne:


docs/libraries/go.md
```diff
@@ -46,7 +46,7 @@ import (
 	"os"
 
 	"github.com/brimdata/super"
-	"github.com/brimdata/super/zio/supio"
+	"github.com/brimdata/super/sio/supio"
```
nwt, line 49: You should also update the package names that appear in the text up above.

mccanne: I was going to do that later after pkg.go.dev is eventually consistent

mccanne: Also these docs are obsolete and will be replaced by the copy now in the book branch

nwt (approved):

Issue #6143, opened 2025-08-21
Failed pipe syntax "from" with URL, method, and headers


philrz:
```
$ super -c 'from http://ifconfig.co ( method GET headers {Accept:"application/json"} )'
panic: interface conversion: interface {} is semantic.argExpr, not *ast.ArgText

goroutine 1 [running]:
github.com/brimdata/super/compiler/semantic.(*analyzer).exprArg(0xc00016f838, 0xc0001267e0?, {0x117cf2a9, 0x7})
	/Users/phil/work/super/compiler/semantic/op.go:1344 +0x126
github.com/brimdata/super/compiler/semantic.(*analyzer).semFromURL(0xc00016f838, {0x11f03458, 0xc00058ea80}, {0xc000418c18, 0x12}, {0xc0001267e0, 0x2, 0x2})
	/Users/phil/work/super/compiler/semantic/op.go:297 +0x31a
github.com/brimdata/super/compiler/semantic.(*analyzer).semFromName(0xc00016f838, {0x11f03458, 0xc00058ea80}, {0xc000418c18, 0x12}, {0xc0001267e0, 0x2, 0x2})
	/Users/phil/work/super/compiler/semantic/op.go:236 +0xc5
github.com/brimdata/super/compiler/semantic.(*analyzer).semFromEntity(0xc00016f838, {0x11f0ba00?, 0xc00058ea80}, 0x0, {0xc0001267e0, 0x2, 0x2}, {0x0, 0x0, 0x0})
	/Users/phil/work/super/compiler/semantic/op.go:102 +0x5b4
github.com/brimdata/super/compiler/semantic.(*analyzer).semFromElem(0xc00016f838, 0xc0005b0ea0, {0x0?, 0x2?, 0x2?})
	/Users/phil/work/super/compiler/semantic/op.go:57 +0x5b
github.com/brimdata/super/compiler/semantic.(*analyzer).semFrom(0x0?, 0xc000795b78?, {0x0?, 0x1?, 0x0?})
	/Users/phil/work/super/compiler/semantic/op.go:47 +0x3b
github.com/brimdata/super/compiler/semantic.(*analyzer).semOp(0xc00016f838, {0x11f0bb20?, 0xc00033a740}, {0x0, 0x0, 0x0})
	/Users/phil/work/super/compiler/semantic/op.go:559 +0x1b66
[...]

Issue #6050, opened 2025-07-18
Enhance flatten function to cover map and type types


philrz:
The new `unnest` functionality introduced in #6021 did not carry over some functionality for the `map` type that was present in the previous `over` operator. It's been proposed this gap be filled via enhancements to the `flatten` function.

## Details

As this issue is being opened, super is at commit f016b90.

At super commit fa4174f right before the merge of the `unnest` enhancements in #6021, the `over` operator could unwrap values of the `map` type.

```
$ super -version
Version: fa4174f32

$ echo '|{"APPL":145.03,"GOOG":87.07}|' | super -c 'over this' -
{key:"APPL",value:145.03}
{key:"GOOG",value:87.07}
```

This functionality did not make it through the transition, so now that `over` in its prior form is gone, here's what a user sees if `unnest` is applied to a map value at current tip-of-`main`:

```
$ super -version
Version: f016b909b

$ echo '|{"APPL":145.03,"GOOG":87.07}|' | super -c 'unnest this' -
error({message:"unnest: encountered non-array value",on:|{"APPL":145.03,"GOOG":87.07}|})

$ echo '|{"APPL":145.03,"GOOG":87.07}|' | super -c 'unnest flatten(this)' -
error({message:"unnest: encountered non-array value",on:|{"APPL":145.03,"GOOG":87.07}|})

$ echo '|{"APPL":145.03,"GOOG":87.07}|' | super -c 'flatten(this)' -
|{"APPL":145.03,"GOOG":87.07}|
```

This was anticipated by @mccanne in the [draft of updated `unnest` docs](https://superdb.org/book/super-sql/operators/unnest.html):

> TODO: `flatten` should take maps and do what `over m` currently does when `m` is a map.

In that same note, @mccanne made a point about similar treatment of `type` values. I'll assume whoever takes up this issue would address both `map` and `type` types at the same time, but this can always be broken our to a separate issue if necessary.

>  (flatten should take type values too and do the right thing producing elemental type values from the a complex type value

philrz:
Note to self: When this gets addressed I'll be able to update the `widen.spq` script in Autoperf. It's still using `over` and an older `super` binary that includes it because its functionality is dependent on being able to unpack maps.

Issue #6016, opened 2025-07-10
SQL: CTE with column aliasing


philrz:
The following query doesn't currently work in SuperDB but does work in Postgres.

```
$ super -version
Version: 7f23f6560

$ super -c "with cte1(xxx) as (Select i as j from a) select xxx from cte1;"
parse error at line 1, column 10:
with cte1(xxx) as (Select i as j from a) select xxx from cte1;
     === ^ ===
```

## Details

Repro is with super commit 7f23f65. This was found via a query from [a sqllogictest](https://github.com/duckdb/duckdb/blob/223ff0a7dba7900039d5910a009247ef097fff3c/test/sql/cte/test_cte.test#L24-L27).

The Postgres docs only seem to cover this concept in the context of [Recursive CTEs](https://www.postgresql.org/docs/current/queries-with.html#QUERIES-WITH-RECURSIVE). However, as the sqllogictest shows, used with a regular CTE as in the query above, the `(xxx)` has the effect of renaming the column returned by the query defined in the CTE.

```
$ psql postgres
psql (17.5 (Homebrew))
Type "help" for help.

postgres=# create table a(i integer);
CREATE TABLE
postgres=# insert into a values (42);
INSERT 0 1
postgres=# with cte1(xxx) as (Select i as j from a) select xxx from cte1;
 xxx 
-----
  42
(1 row)
```

philrz:
Verified in super commit 72ac4de.

Given input data:

```
$ super a
{i:42}
```

The repro query now returns the same result as we saw with Postgres.

```
$ super -version
Version: 72ac4deb7

$ super -c "with cte1(xxx) as (Select i as j from a) select xxx from cte1;"
{xxx:42}
```

Thanks @mccanne!

PR #6151, opened 2025-08-22
Improvements to new Pipe Operators docs

philrz:


philrz:
In `intro.md`, the docs currently show:

```
put lower(s), a.b.c, x+1
```

Saying it's equivalent to:

```
put lower:=lower(s), c:=a.b.c, `x+1`:=x+1
```

However, when I run it, the middle one in the shorthand looks to me like a no-op, e.g.,

```
$ super -version
Version: c81a6353d

$ echo '{s:"HI",a:{b:{c:1}},x:2}' | super -c 'put lower(s), a.b.c, x+1' -
{s:"HI",a:{b:{c:1}},x:2,lower:"hi","x+1":3}

$ echo '{s:"HI",a:{b:{c:1}},x:2}' | super -c 'put lower:=lower(s), c:=a.b.c, `x+1`:=x+1' -
{s:"HI",a:{b:{c:1}},x:2,lower:"hi",c:1,"x+1":3}
```

The doc sets expectations:

> for a dotted path expression, the name is the last element of the path;

Assuming my test data matches what this example had in mind, does the code need a fix, or should we drop it from the docs?

philrz:
`from.md` currently shows arguments like `header` but I could not get them to work in a simple example, so I opened #6143. Could someone confirm if my findings there are a valid bug or user error? If it's a valid bug then I'd advocate we commit to fixing it before GA release or remove mention of these arguments from the docs.

book/src/super-sql/operators/unnest.md
```diff
@@ -52,8 +52,8 @@ sequence has the form:
 where the first field is copied to each derived value and the second field is
 the unnested elements of the array `elem0`, `elem1`, etc.
 
-To explode the fields of records or the key-value paris of maps, use the  
-[`flatten`](../functions/records/flatten.md) function, which produces an array that
+To explode the fields of records or the key-value pairs of maps, use the
+[`flatten` function](../functions/records/flatten.md), which produces an array that
```
philrz: This is not currently accurate as written since per open issue #6050 `flatten` does not currently support maps. I'd advocate we either commit to fixing that issue before GA release or add a caveat here that support for maps is planned for the future.

philrz:
Right now the `unnest` doc shows off error messages like this one that are currently surfaced by `super`:

```
error({message:"unnest: encountered record without two columns",on:<value>})
```

Since it's being shown in the context of a pipe operator, I was a little surprised to see the use of the word "columns" (which I associate with the relational context) rather than "fields" (which I associate with the general concept of records). Could someone confirm if this is intentional?

book/src/super-sql/operators/values.md
```diff
@@ -49,7 +49,7 @@ WITH points(x,y) AS (
   VALUES (2,1),(4,2),(6,3)
 )
 SELECT vals
-FROM points CROSS JOIN LATERAL (VALUES (x+y), (x-y)) t(vals)"
+FROM points CROSS JOIN LATERAL (VALUES (x+y), (x-y)) t(vals)
```
philrz, line 52: Beyond just dropping that `"` to make it paste-able as valid SQL, this example doesn't currently work in SuperDB because of open issue #6016.

```
$ super -version
Version: c81a6353d

$ super -c "
WITH points(x,y) AS (
  VALUES (2,1),(4,2),(6,3)
)
SELECT vals
FROM points CROSS JOIN LATERAL (VALUES (x+y), (x-y)) t(vals)"
parse error at line 2, column 12:
WITH points(x,y) AS (
       === ^ ===
```

But it is indeed valid SQL, e.g., in Postgres:

```
$ psql postgres -c "
WITH points(x,y) AS (
  VALUES (2,1),(4,2),(6,3)
)
SELECT vals
FROM points CROSS JOIN LATERAL (VALUES (x+y), (x-y)) t(vals)"
 vals 
------
    3
    1
    6
    2
    9
    3
(6 rows)
```

The text as written is not explicit about whether it's intended to be SQL code that runs in SuperDB or is just a setup before showing the simpler pipe example as follows. However, I'd advocate we commit to either:

1. Fixing #6016 before a GA release,
2. Change the example to something that runs ok with what we expect to have working in the GA release, or,
3. Add a caveat saying the SQL one doesn't work yet in SuperDB but will soon.

book/src/super-sql/patterns.md
```diff
@@ -10,14 +10,14 @@ path can be expressed as a [text entity](#text-entity) and need not be quoted:
...
-Likewise, in the search operator, the syntax for a [regular expression](#regular-expression)
+Likewise, in the [`search` operator](operators/search.md), the syntax for a [regular expression](#regular-expression)
 search can be specified as
 ```
 search /\w+(foo|bar)/
 ```
 instead of the `regexp` function call required in expression context
 ```
-where regexp(r'\w+(foo|bar)')
+where len(regexp(r'\w+(foo|bar)', this)) > 0
```
philrz, line 20: The side-by-side `search` and `where` examples were originally introduced as if they'd produce similar results, but the `where` one didn't work because `regexp` returns either a `null` or an array of matches rather than the boolean that `where` expects, so I've rewritten this `where` example.

book/src/super-sql/operators/aggregate.md
```diff
@@ -47,7 +47,7 @@ Each aggregate function `<agg-func>` may be optionally followed by a `where` cla
 which applies a Boolean expression `<expr>` that indicates, for each input value,
 whether to include it in the values operated upon by the aggregate function.
 `where` clauses are analogous
-to the [`where`](where.md) operator but apply their filter to the input
+to the [`where` operator](where.md) but apply their filter to the input
```
philrz: Edits like this one are to maintain a prior convention that when we found ourselves hyperlinking to specific operators/functions, etc., we'd try to include the word "operator"/"function"/etc. and also put it inside the hyperlinked text. The thinking has been that this would help disambiguate in the small number of cases where we've got the same keyword for two different things, e.g., "`fuse` operator" and "`fuse` aggregate function", but the convention is maintained regardless.

book/src/super-sql/operators/from.md
```diff
@@ -18,7 +18,7 @@ a query and transmits that data to its output.
 
 It has two forms:
 * a `from` pipe operator with dataflow scoping as described here, or
-* a [`FROM`](../sql/from.md) SQL clause with relational scoping.
+* a [SQL `FROM` clause](../sql/from.md) with relational scoping.
```
philrz: Similar to my previous comment about a convention for hyperlinking to pipeline element docs, it seems like a reasonable convention would be to hyperlink to `SQL <something> clause`, since lots of SQL resources I see out there (e.g., [w3schools](https://www.w3schools.com/sql/sql_where.asp)) seem to lead with the "SQL".

book/src/super-sql/operators/join.md
```diff
@@ -64,8 +64,8 @@ within nested scopes, the pipeline join operator uses dataflow scoping to join d
 Here, all data is combined into joined records that can be operated upon 
 like any other record without complex scoping logic.
 
-If relational scoping is desired, the [SQL join](../sql/join.md)
-operator can be used instead.
+If relational scoping is desired, a [SQL `JOIN` clause](../sql/join.md)
```
philrz: It seems like capitalizing the SQL keywords would help emphasize when we're talking about SQL.

book/src/super-sql/operators/tail.md
```diff
@@ -36,7 +36,8 @@ tail 2
 _Grab last two values of arbitrary sequence, using a different representation of two_
 ```mdtest-spq
 # spq
-tail 1+1
+const ONE = 1
+tail ONE+1
```
philrz, line 40: @mccanne enhanced the example in the `head` doc to use this `const` approach so I figured we might as well do the same in `tail`.

book/src/super-sql/patterns.md
```diff
@@ -30,13 +30,19 @@ and is documented in the
...
+represent a regular expression as a string, prefix with `r`, e.g.,
+```
+regexp(r'\w+(foo|bar)', this)
 ```
 
 But when used outside of expressions where an explicit indication of
 a regular expression is required (e.g., in a
 [`search`](operators/search.md) or 
-[`from`](operators/from.md) operator), the RE2 is instead
+[`from`](operators/from.md#database-operation) operator), the RE2 is instead
```
philrz, line 45: From what I can see, the only place regexps are currently used with `from` is to match database pools, so I figured a subtle clarification would be to hyperlink to the relevant section.

mccanne:
> However, when I run it, the middle one in the shorthand looks to me like a no-op, e.g.,

This is because the code isn't quite up to date with the docs I wrote (I missed this). I wanted to unify the semantics of a field assignment across all the ops that use it so we need to update `put` to follow the semantics or decide the old behavior is what we want and document the deviation.

As an example `put x.y` should behave like `put y:=x.y` here...
```
; echo '{x:{y:1}}' | super -c "put y:=x.y" -
{x:{y:1},y:1}
```

mccanne:
> `from.md` currently shows arguments like `header` but I could not get them to work in a simple example, so I opened #6143. Could someone confirm if my findings there are a valid bug or user error? If it's a valid bug then I'd advocate we commit to fixing it before GA release or remove mention of these arguments from the docs.

Oops, the docs introduced the bug as the keyword is `headers` instead of `header`.  I think we should just fix the new docs (it's always been `headers`).

mccanne:
> Right now the `unnest` doc shows off error messages like this one that are currently surfaced by `super`:
> 
> ```
> error({message:"unnest: encountered record without two columns",on:<value>})
> ```
> 
> Since it's being shown in the context of a pipe operator, I was a little surprised to see the use of the word "columns" (which I associate with the relational context) rather than "fields" (which I associate with the general concept of records). Could someone confirm if this is intentional?

See #6154

book/src/super-sql/operators/values.md
mccanne: Let's discuss.

book/src/super-sql/operators/aggregate.md
mccanne: I moved to the opposite convention and prefer it but we can discuss.

book/src/super-sql/operators/unnest.md
mccanne: Yeah we decided to do the generalized flatten logic later.  Let's add the caveat.

mccanne: Yeah we decided to do the generalized flatten logic later.  Let's add the caveat.

book/src/super-sql/operators/search.md
```diff
@@ -15,7 +15,7 @@ along the lines of web search, email search, or log search.
 A search operation filters its input by applying the search expression `<sexpr>`
 to each input value and emitting all values that match.
 
-The `search` keyword can be abbreviated as `?`. 
+The `search` keyword can be [abbreviated](../shortcuts.md) as `?`.
```
mccanne: I don't think of `?` as a short-cut the way other structures are syntactic shortcuts where you can omit the operator keyword.  Here search is being abbreviated as ? and I didnt plan on documenting this in the shortcuts section.

philrz: Got it. I took out the hyperlink.

philrz:
@mccanne: Thanks for the tip on it being `headers` rather than `header` in the `from` with URL. Now instead of a parse error I get a stack trace, so I've updated #6143 to reflect that. 😛  But I've updated the doc to say `headers` so we're on the right track.

PR #6161, opened 2025-08-27
eliminate implied function arguments

mccanne:
This commit eliminates function arguments that are implied, i.e., is() now requires two arguments and an explicit "this", the every() function is removed in favor of bucket(ts, ...) as SuperDB no longer treats "ts" as a special value, and nest_dotted() no longer allows for an implied "this".

We also removed the logic in rungen and in the runtime function packages that converted implied arguments to actuals since it is no longer needed.


nwt (approved):
:shipit:

PR #6167, opened 2025-08-28
add support for arbitrary join conditions

nwt:
The pipe join operator and SQL join clause support only equi-join conditions with a single comparison.  Extend them to support arbitrary join conditions by reworking the existing cross join implementation into a full-feature nested loop join.

For efficiency, single-comparison equi-join continues to use the existing hash join implementation.  The optimizer now determines when this is possible and communicates this to the runtime generator by altering the DAG, replacing dag.Join with dag.HashJoin.


compiler/ztests/pushdown.yaml
```diff
@@ -3,12 +3,18 @@ script: |
...
+  super compile -C -O 'fork ( from file1 ) ( from file2 ) | join on right.a>left.b | values left.c,right.d'
+  echo === hash join
   super compile -C -O "from file1 | join (from file2) on left.a=right.b | values left.c,right.d"
-  echo === switch-join
+  echo === nested loop join
+  super compile -C -O "from file1 | join (from file2) on left.a>right.b | values left.c,right.d"
+  echo === switch into hash join
   super compile -C -O 'from file | switch a case b ( put x:=c ) case d ( put x:=e ) | join on left.f=right.g | values left.h,right.i'
+  echo === switch into nested loop join
+  super compile -C -O 'from file | switch a case b ( put x:=c ) case d ( put x:=e ) | join on left.f>right.g | values left.h,right.i'
```
mattnibs, line 17: I might add one trickier test to make sure a reverse order expression equality expression that is somewhat complicated gets correctly identified and flipped (appears to be working in your pr):
```sql
select *
from (values (1,2),(2,3)) as t1(a,b)
left join (values (2),(3)) as t2(c) on c = a+b
```

nwt: @mattnibs: We've got a couple tests like that at https://github.com/brimdata/super/blob/e4f3f5b1d51e2609abc54aa4fd7f53f7415a79ab/runtime/ztests/op/join-expr.yaml#L4-L6. Does that feel sufficient?

mattnibs: Oh cool didn't see that

PR #6169, opened 2025-08-29
generalize CAST-AS syntax

mccanne:
This commit updates the CAST-AS syntax to allow the complete type system for the AS types.  We updated a test from INT64 to int64 since we should add case-insensitivity of type names to the grammar instead of handling them only in CAST AS.

Updated docs coming from the book branch in a subsequent PR.

### DIFF
--- a/compiler/sfmt/ztests/sql.yaml
+++ b/compiler/sfmt/ztests/sql.yaml
@@ -20,7 +20,7 @@ script: |
   echo === query expressions
   super compile -C 'select a+(select b from c), d+(select e) as f'
   echo === other expressions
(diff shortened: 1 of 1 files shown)


compiler/sfmt/ztests/sql.yaml
```diff
@@ -20,7 +20,7 @@ script: |
   echo === query expressions
   super compile -C 'select a+(select b from c), d+(select e) as f'
   echo === other expressions
-  super compile -C "select cast(a as b), substring(c from d for e), date 'f', timestamp 'g'"
+  super compile -C "select cast(a as int64), substring(c from d for e), date 'f', timestamp 'g'"
```
nwt: Nit: Is there any reason to change this? It passes with "b".

PR #6176, opened 2025-08-30
make user ops more like thunks

mccanne:
This commit removes code to allow user ops arguments to be full-fledged thunks.  We had already implemented thunk logic here and it just needed to be exposed by removing the code that didn't allow it.


compiler/semantic/op.go
```diff
@@ -1218,35 +1218,19 @@ func (a *analyzer) maybeConvertUserOp(call *ast.Call) dag.Seq {
...
-					a.error(arg, errors.New(string(val.Bytes())))
-				}
-			}
-			e = &dag.Literal{
-				Kind:  "Literal",
-				Value: sup.FormatValue(val),
-			}
-		}
-		exprs[i] = e
+		exprs[i] = a.semExpr(arg)
```
nwt: Nit: Ditch this loop in favor of semExprs.

```go
	exprs := a.semExprs(args)
```

compiler/ztests/thunk.yaml
```diff
@@ -0,0 +1,7 @@
+spq: |
```
nwt: Nit: Call this user-op-thunk.yaml so it'll appear next to the tests in #6177.

PR #6177, opened 2025-08-31
automatically promote a user-op in an expression to a subquery

mccanne:
This commit changes the handling of user-ops referenced in expressions from an error to a subquery.  When the user-op begins with a from, then the subquery is spliced into the expression as an uncorrelated query. Otherwise, it is dependenent on its input and the user-op is expanded into the pattern (values this | userOp(params) | collect(this)).

### DIFF
--- a/compiler/semantic/expr.go
+++ b/compiler/semantic/expr.go
@@ -629,13 +629,12 @@ func (a *analyzer) semCall(call *ast.Call) dag.Expr {
 		return badExpr()
 	}
 	if userOp := a.maybeConvertUserOp(call); userOp != nil {
-		// When a user op is encountered inside an expression, we turn it into
+		// When a user op is encountered inside an expression, we turn it into a
 		// subquery operating on a single-shot "this" value unless it's uncorrelated
 		// (i.e., starts with a from), in which case we put the uncorrelated body
 		// in the subquery without the values/collect logic.
-		var correlated bool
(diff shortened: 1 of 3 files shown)


compiler/semantic/expr.go
```diff
@@ -628,6 +628,41 @@ func (a *analyzer) semCall(call *ast.Call) dag.Expr {
 		a.error(call, errors.New("'where' clause on non-aggregation function"))
 		return badExpr()
 	}
+	if userOp := a.maybeConvertUserOp(call); userOp != nil {
+		// When a user op is encountered inside an expression, we turn it into
+		// subquery operating on a single-shot "this" value unless it's uncorrelated
```
```suggestion
		// When a user op is encountered inside an expression, we turn it into a
		// subquery operating on a single-shot "this" value unless it's uncorrelated
```

compiler/semantic/expr.go
```diff
@@ -628,6 +628,41 @@ func (a *analyzer) semCall(call *ast.Call) dag.Expr {
...
+		// (i.e., starts with a from), in which case we put the uncorrelated body
+		// in the subquery without the values/collect logic.
+		var correlated bool
+		if isCorrelated(userOp) {
+			correlated = true
```

```suggestion
		correlated := isCorrelated(userOp)
		if correlated {
```

mccanne:
We're gonna do user-op syntax differently so this doesn't make sense any more.

PR #6179, opened 2025-09-02
Per-page table of contents in book

philrz:
## tl;dr

A feature I miss from the old docs site was the per-page table of contents that allowed a user to easily browse and jump to the section headers on a single page. I did some digging and found a way to enable this in mdbook. However, I seem to recall  some mixed feelings about this feature in the old docs site, so I'm submitting here for consideration.

## Details

To configure, I followed the instructions at https://github.com/slowsage/mdbook-pagetoc by creating a `theme/` directory into which I copied the `pagetoc.css` and `pagetoc.js` files from that repo. I also had to `cargo install mdbook-pagetoc` to install a binary tool needed to help with the rendering, which for some reason wasn't mentioned in the README. I guess Rust people just know these things intuitively. 🤷‍♂️

A pro/con of the implementation (depending on how you feel about it) is that it's pretty conservative about when the TOC shows up, i.e., the window needs to be pretty wide for it to appear. The attached video shows the effect when rendering the book from this branch. If my memory serves me correctly, those that were less keen on the TOC thought it was visual clutter and/or had the potential to be confusing when in view at the same time as the left hand nav. If that's true, then maybe its conservative nature is a good thing since it's likely to only appear for users that actually want it, like me. 😄

https://github.com/user-attachments/assets/7ce34175-2067-4166-93e8-81d66e3d6f5c

If this passes muster and gets merged, I'll add the `cargo install mdbook-pagetoc` install step to the website deployment Actions workflow.

### DIFF
--- a/book/.gitignore
+++ b/book/.gitignore
@@ -1,8 +1,8 @@
 book
+theme
 
 node_modules
 package-lock.json
 src/super.wasm
 super-example.bundle.js
 super-example/wasm_exec.js
-theme


book/book.toml
```diff
@@ -7,9 +7,10 @@ title = "SuperDB"
 [output.html.fold]
 enable = true
 
+[preprocessor.pagetoc]
```
nwt, line 10:
```suggestion
[preprocessor.pagetoc]

```

book/theme/pagetoc.css
```diff
@@ -0,0 +1,57 @@
+@media only screen and (max-width:1439px) {
```
nwt: Since we're not customizing this and pagetoc.js, let mdbook-pagetoc generate them instead of adding them to version control.

Then you just need to add "theme" to book/.gitignore:

```
book
theme

node_modules
package-lock.json
src/super.wasm
super-example.bundle.js
super-example/wasm_exec.js
```

philrz:
@nwt: Thanks for the idea. That tested out well so I just pushed a commit that does what you suggested. I'd never have put that together based on the docs I was following.

book/.gitignore
```diff
@@ -5,3 +5,4 @@ package-lock.json
 src/super.wasm
 super-example.bundle.js
 super-example/wasm_exec.js
+theme
```
nwt: Put this after "book" with a blank line after it as in https://github.com/brimdata/super/pull/6179#discussion_r2319316706.

PR #6181, opened 2025-09-03
new syntax for user operator declarations and invocations

mccanne:
This commit improves the syntax for user operators to better align with built-in operators and distinguish their syntax from user functions.

The new syntax for op declarations is:

     op name arg, arg, ... : ( ... )

and invoking an operator is now:

    call name arg, arg, ...

with a short-cut where you can drop the "call" as a shortcut:

    name arg, arg, ...

(as long as the shortcut is not a single operator with no arguments inside of a subquery)

Updated documentation is coming in a subsequent PR from the book branch.

### DIFF
--- a/compiler/parser/parser.peg
+++ b/compiler/parser/parser.peg
@@ -54,7 +54,7 @@ FuncDecl
     }
 
 OpDecl
-  = OP _ name:Identifier __ "(" __ params:Identifiers? __ ")" __ ":" __ body:ScopeBody {
+  = OP _ name:Identifier params:(_ ids:Identifiers { return ids, nil })? __ ":" __ body:ScopeBody {
       return &ast.OpDecl{
         Kind: "OpDecl",
         Name: name.(*ast.ID),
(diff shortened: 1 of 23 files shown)


compiler/parser/parser.peg
```diff
@@ -79,21 +79,30 @@ TypeDecl
...
 PipeOp
   = Operator
   / "(" __ scope:Scope __ ")" { return scope, nil }
   / a:OpAssignment &EndOfOp { return a, nil }
   / !(Function EndOfOp) a:Aggregation &EndOfOp { return a, nil }
+  / !CallIDGuard id:Identifier _ !CallExprGuard args:Exprs {
+      return &ast.CallOp{Kind:"CallOp", Name: id.(*ast.ID), Args:sliceOf[ast.Expr](args), Loc:loc(c)}, nil
+    }
+  / !CallIDGuard id:Identifier &EndOfOp {
+      return &ast.CallOp{Kind:"CallOp", Name: id.(*ast.ID), Loc:loc(c)}, nil
```
nwt: Nit: We usually name these for the field they initialize rather than their rule.

```suggestion
  / !CallIDGuard name:Identifier _ !CallExprGuard args:Exprs {
      return &ast.CallOp{Kind: "CallOp", Name: name.(*ast.ID), Args: sliceOf[ast.Expr](args), Loc: loc(c)}, nil
    }
  / !CallIDGuard name:Identifier &EndOfOp {
      return &ast.CallOp{Kind: "CallOp", Name: name.(*ast.ID), Loc: loc(c)}, nil
```

compiler/parser/parser.peg
```diff
@@ -409,6 +419,15 @@ TopOp
       return op, nil
     }
 
+CallOp
+  = CALL _ id:Identifier args:((_ args:Exprs) { return args, nil })? {
+      op := &ast.CallOp{Kind:"CallOp", Name: id.(*ast.ID), Loc:loc(c)}
+      if args != nil {
+        op.Args = sliceOf[ast.Expr](args)
+      }
+      return op, nil
```
nwt: Nits: Name the Identifier label for the field it initializes. And passing nil to sliceOf is safe. 

```suggestion
  = CALL _ name:Identifier args:((_ args:Exprs) { return args, nil })? {
      return &ast.CallOp{
        Kind: "CallOp",
        Name: name.(*ast.ID),
        Args: sliceOf[ast.Expr](args),
        Loc: loc(c)
      }, nil
    }
```

compiler/parser/parser.peg
```diff
@@ -79,21 +79,30 @@ TypeDecl
     }
 
 // An PipeOp is any pipeline operator, or a parenthesized pipe scope, or
-// one of three shortcuts: a shortcut assignment for "put", a shortcut "aggregate",
-// or an expression that is either an implied "values" (non-boolean) or an implied
-// "where" (boolean).
+// one of four shortcuts: a shortcut assignment for "put", a shortcut "aggregate",
+// a shortcut "call" operator,  or an expression that is either an implied "values"
```
nwt: Nit: For consistency with the other items in this list.

```suggestion
// a shortcut "call",  or an expression that is either an implied "values"
```

nwt:
I think this new syntax looks really good!

PR #6183, opened 2025-09-03
Correctly propogate schema on select * in from subquery

### LINKED ISSUE #6125: SQL: Not finding named column in result from "select *" subquery
## tl;dr

The SQL query in this command line doesn't currently work in `super`, but does in Postgres.

```
super -c "select id from (select * from (values (1)) as x(id))"
```

## Details

Repro is with super commit f5d8fc6.

```
$ super -version
Version: f5d8fc63d

$ super -c "select id from (select * from (values (1)) as x(id))"
column "id": does not exist at line 1, column 8:
select id from (select * from (values (1)) as x(id))
       ~~
```

@mattnibs stumbled onto this problem and made this observation when contemplating the root cause:

> The `*` is just ignored when resultant schema is computed from a `select`

Indeed, it works if we use named column `id` at both layers.

```
$ super -c "select id from (select id from (values (1)) as x(id))"
{id:1}
```

And the original query works ok in Postgres.

```
$ psql postgres
psql (17.5 (Homebrew))
Type "help" for help.

postgres=# select id from (select id from (values (1)) as x(id));
 id 
----
  1
(1 row)
```

mattnibs:
Closes #6125

### DIFF
--- a/compiler/ztests/sql/scope-from-subquery.yaml
+++ b/compiler/ztests/sql/scope-from-subquery.yaml
@@ -1,0 +1,17 @@
+script: |
+  super -s -c 'select id from ( select * from (values (1)) x(id) )'
+  echo // ===
+  # Ensure id can be selected when subquery has select * on a dynamic schema.
+  super -s -c 'select id from ( select * from in.sup )'
+
+inputs:
+  - name: in.sup
+    data: |
+      {id:1}
+
+outputs:
+  - name: stdout
+    data: |
+      {id:1}
+      // ===
+      {id:1}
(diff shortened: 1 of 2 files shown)


compiler/ztests/sql/scope-from-subquery.yaml
```diff
@@ -0,0 +1,17 @@
+script: |
```
nwt: Nit: Call this something like subqueries-select-star.yaml so it's grouped with the other subqueries-*.yaml tests.

nwt (approved):

PR #6187, opened 2025-09-04
fix markdownlint errors

nwt:
Fixing these errors will allow us to run markdownlint against the entire repository in CI.

### DIFF
--- a/book/src/intro.md
+++ b/book/src/intro.md
@@ -175,7 +175,8 @@ a product type &mdash; it unfortunately did not anticipate sum types.
 > _Codd's original paper on the relational model has a footnote that
 > essentially describes as a product type:_
 >
-> ![Codd's Footnote](codd-footnote.png "Codd's Footnote")
+> <!-- markdownlint-disable-next-line MD033 -->
+> <center><img src="codd-footnote.png" alt="Codd's Footnote" title="Codd's Footnote" width="500"></center>
 >
 > _But sum types were notably absent._
 


book/src/intro.md
```diff
@@ -168,37 +168,37 @@ What's missing here is an easy and native way to represent mixed-type entities.
...
-While the original conception of the relational data model anticipated 
+While the original conception of the relational data model anticipated
 "product types" &mdash; in fact, describing a relation's schema in terms of
 a product type &mdash; it unfortunately did not anticipate sum types.
 
 > _Codd's original paper on the relational model has a footnote that
 > essentially describes as a product type:_
 >
-> <center><img src="codd-footnote.png" alt="Codd's Footnote" title="Codd's Footnote" width="500"></center>
+> ![Codd's Footnote](codd-footnote.png "Codd's Footnote")
```
philrz: This happens to result in a noticeable change in the formatting. Rendered at tip of `main` the relevant section looks like this:

image

Whereas on this branch it looks like this:

image

I don't necessarily find the new rendering distasteful, but since @mccanne wrote the original text he might have used HTML with that specific `width` setting to get the formatting just how he wanted it, so checking if he wants to chime in here.

nwt: @philrz: @mccanne would like to keep the 500-pixel width so 513ed2d211f4b568259715068b574d833ccb655f restores that.

PR #6189, opened 2025-09-04
don't panic if -C specified without -c

mccanne:


compiler/sfmt/ast.go
```diff
@@ -12,6 +12,9 @@ import (
 )
 
 func AST(p ast.Seq) string {
+	if len(p) == 0 {
+		return ""
+	}
```
nwt, line 17: Nit: Tests pass without this.

```suggestion
```

PR #6194, opened 2025-09-05
remove performance/README.md et al.

nwt:
scripts/perf-compare.sh, which generates it, depends on docs/integrations/zeek/shaping-zeek-json.md, and that was removed in #6190.  Getting it working again doesn't seem worth the effort.


nwt:
See #6195.

Issue #2071, opened 2021-02-04
z backtick string literal


mccanne:
Z should have backtick string literals like ZSON

philrz:
The recently-added SuperSQL support for raw strings (#6112) seems to cover the use cases we previously imagined would be covered by backtick string literals (e.g., newlines per #6208) so there's consensus that we can close this issue

Issue #3568, opened 2022-02-06
zson backtick strings broken


mccanne:
This `echo '`foo`' | zq -`  should produce `"foo"` but instead it produces no output.

philrz:
It looks like this issue is still with us as of super commit 18c86ef, but the [current SUP docs](https://superdb.org/book/formats/sup#23-primitive-values) make it sound like it should work.

Revisiting the original repro with current syntax, the symptom is as previously described:

```
$ super -version
Version: 18c86efe5

$ echo '`foo`' | super -i sup -
[no output]
```

Back in January 2024, I can see @mattnibs bumped into what looks like a variation on broken backtick support.

```
$ super -i sup - << 'EOF'
{
  test: `              
        Hello Darkness
        My Old Friend`
}
EOF

stdio:stdin: parse error: mismatched string backticks
```

We recently made it so most SUP literals could appear in `values`, so I tried this one too, which didn't parse.

```
$ super -c '
values {
  test: `
        Hello Darkness
        My Old Friend`
}'

parse error at line 3, column 10:
  test: `
     === ^ ===
```

It should be noted that backtick support _does_ work as relates to identifiers in SuperSQL, e.g., such as shown in the [backtick-id.yaml](https://github.com/brimdata/super/blob/main/compiler/ztests/sql/backtick-id.yaml) ztest.

```
$ super -c 'select `x` as `y` from (values (1),(2)) T(x)'
{y:1}
{y:2}
```

philrz:
Per #6210, we've decided to drop this from the SUP format spec for now, and the non-working code that tried to cover it has been removed from the tooling as well. Closing this issue.

PR #6197, opened 2025-09-05
Improvements to the grok function doc

philrz:
In addition to minor formatting/conventions stuff, here I'm pushing the use of raw strings which I think is an improvement over the double backslashes that were shown before.

### DIFF
--- a/book/src/super-sql/functions/parsing/grok.md
+++ b/book/src/super-sql/functions/parsing/grok.md
@@ -47,10 +47,10 @@ been published by Elastic and others that provide helpful guidance on becoming
 proficient in Grok. To help you adapt what you learn from these resources to
 the use of the `grok` function, review the tips below.
 
-> As these represent areas of possible future SuperSQL enhancement, links to open
+> _As these represent areas of possible future SuperSQL enhancement, links to open
 > issues are provided. If you find a functional gap significantly impacts your
 > ability to use the `grok` function, please add a comment to the relevant
-> issue describing your use case.
+> issue describing your use case._
 
 1. Logstash's Grok offers an optional data type conversion syntax,
    e.g.,
@@ -106,7 +106,7 @@ the use of the `grok` function, review the tips below.
    avoid compatibility issues, we recommend building configurations starting
    from the RE2-based [included patterns](#included-patterns).
 
-> If you absolutely require features of Logstash's Grok that are not currently
+> _If you absolutely require features of Logstash's Grok that are not currently
 > present in SuperSQL, you can create a Logstash-based preprocessing
 > pipeline that uses its
 > [Grok filter plugin](https://www.elastic.co/guide/en/logstash/current/plugins-filters-grok.html)
@@ -114,7 +114,7 @@ the use of the `grok` function, review the tips below.
 > [super/3151](https://github.com/brimdata/super/issues/3151) provides some tips for
 > getting started. If you pursue this approach, please add a comment to the
 > issue describing your use case or come talk to us on
-> [community Slack](https://www.brimdata.io/join-slack/).
+> [community Slack](https://www.brimdata.io/join-slack/)._
 
 #### Debugging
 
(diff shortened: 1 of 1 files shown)


book/src/super-sql/functions/parsing/grok.md
```diff
@@ -190,16 +189,16 @@ values grok("%{MY_NUMTZ:tz}",
...
 ```mdtest-spq {data-layout="stacked"}
 # spq
-values grok("\\(%{PH_PREFIX:prefix}\\)-%{PH_LINE_NUM:line_number}",
-           this,
-          "PH_PREFIX \\d{3}\n" +
-          "PH_LINE_NUM \\d{4}")
+values grok(r"\(%{PH_PREFIX:prefix}\)-%{PH_LINE_NUM:line_number}",
+            this,
+            r"PH_PREFIX \d{3}" + "\n" +
+            r"PH_LINE_NUM \d{4}")
```
philrz: Showing a combo of raw strings concat'ed with a regular `"\n"` string is a minor bummer, but I still feel like it's an improvement over double backslashes. What I probably really wanted is the oft-discussed backtick string literals, but #2071.

mccanne: We use backticks now for identifiers like some SQLs.  Now that we have raw strings, I think we should close #2071 

We could do what duckdb does and allow newlines inside of strings...
```
; duckdb -c "select '  foo
  bar '"
┌─────────────────┐
│ '  foo\n  bar ' │
│     varchar     │
├─────────────────┤
│   foo\n  bar    │
└─────────────────┘
```

philrz: @mccanne: Yes, I spent a fair amount of time contemplating raw strings and the open issues about backtick strings while I was updating this.

I noticed that we already do something like what you showed for DuckDB raw strings, but not for regular strings.

```
$ super -version
Version: 54aa3ff47

$ super -c "values 'foo
bar'"
parse error at line 1, column 12:
values 'foo
       === ^ ===

$ super -c "values r'foo
bar'"
"foo\nbar"
```

I'd be open to seeing it supported with regular strings, but this behavior seems more in line with what I expect of raw strings, so I'm not personally gunning for a change there. Open to discussion though.

Back on the Grok example, this means that I could have used raw strings, but because the indents still make their way into the string, to make it work I have to write it this way that's kind of ugly.

```
$ echo '"(555)-1212"' | super -c '
values grok(r"\(%{PH_PREFIX:prefix}\)-%{PH_LINE_NUM:line_number}",
            this,
            r"PH_PREFIX \d{3}
PH_LINE_NUM \d{4}")' -

{prefix:"555",line_number:"1212"}
```

I can improve slightly on that approach by putting the multi-line definition in a `const` so I can maintain the tidy formatting of my function call.

```
$ echo '"(555)-1212"' | super -c '
const defs = r"PH_PREFIX \d{3}
PH_LINE_NUM \d{4}"

values grok(r"\(%{PH_PREFIX:prefix}\)-%{PH_LINE_NUM:line_number}",
            this,
            defs)' -

{prefix:"555",line_number:"1212"}
```

My travels also had me revisit and update #3568, as the backtick support currently described for SUP sounds like it would have allowed me to put the multi-line string in the original location and still keep my formatting. An excerpt from the [current SUP docs](https://superdb.org/book/formats/sup#23-primitive-values):

> A string may be backtick-quoted with the backtick character &#96;. None of the text between backticks is escaped, but by default, any newlines followed by whitespace are converted to a single newline and the first newline of the string is deleted.

i.e., if the backtick literals inside SuperSQL worked like what's described for SUP, then maybe this would have been an option:

```
$ echo '"(555)-1212"' | super -c '
values grok(r"\(%{PH_PREFIX:prefix}\)-%{PH_LINE_NUM:line_number}",
            this,
            `PH_PREFIX \d{3}
             PH_LINE_NUM \d{4}`)' -
```

Anyway, just going off what's currently working, if there's any preferences between my original update that's already been approved vs. the `const` variation, I'm all ears. It's probably not a big deal regardless, but I like it when our examples are showing what we think of as "best practices".

philrz: Update: I walked through these variations on a Zoom with @mccanne and @nwt. A raw string with a leading newline happens to work in this particular example and there seemed to be some rough consensus that could be seen as a best practice using what's currently working in the code, so I've updated the example to that.

PR #6210, opened 2025-09-08
Drop backtick strings from SUP

philrz:
Per #3568, the support for backtick strings in SUP (and going back to ZSON days) hasn't been working and maybe never worked. I've established consensus that it be removed from the SUP format doc and that we could revisit the topic in the future, such as in response to user demand.

That said, I poked around and can see that it looks like there _is_ code that alleges to handle this:

https://github.com/brimdata/super/blob/215dceede0f7090218575c32f8c4d68eaeaf118e/sup/parser-values.go#L246-L279

I'm always nervous about touching the Go code, so I think some collab with one of y'all developers is justified here. Seems our options include:

1. Drop the Go code as well
2. Leave the Go code but add a comment mentioning that it's known to not work
3. Fix the code and keep it in the SUP spec after all

### DIFF
--- a/sup/lexer.go
+++ b/sup/lexer.go
@@ -445,8 +445,6 @@ func unhexRune(b []byte) (rune, error) {
 	return r0<<12 | r1<<8 | r2<<4 | r3, nil
 }
 
-var newline = []byte{'\n'}
-
 func (l *Lexer) scanTypeName() (string, error) {
 	ok, err := l.match('"')
 	if err != nil {


mccanne:
> Per #3568, the support for backtick strings in SUP (and going back to ZSON days) hasn't been working and maybe never worked. I've established consensus that it be removed from the SUP format doc and that we could revisit the topic in the future, such as in response to user demand.
> 
> That said, I poked around and can see that it looks like there _is_ code that alleges to handle this:
> 
> https://github.com/brimdata/super/blob/215dceede0f7090218575c32f8c4d68eaeaf118e/sup/parser-values.go#L246-L279
> 
> I'm always nervous about touching the Go code, so I think some collab with one of y'all developers is justified here. Seems our options include:
> 
> 1. Drop the Go code as well
> 2. Leave the Go code but add a comment mentioning that it's known to not work
> 3. Fix the code and keep it in the SUP spec after all

Yeah take out the go code

And `var arrow`
The two lines where matchBacktick are called can be deleted

PR #6217, opened 2025-09-09
add lambdas and function references

mccanne:
This commit adds lambdas and function references to SuperSQL. Lambdas are not first class as they cannot be referenced in generic functions but they are callable and they are passable to user operators and the map function.

We updated the docs for the map() function to explain lambdas and function references but more comprehensive updates are forthcoming from the book branch.

The way the function code is structured now does not easily allow for passing lambdas to functions so this is not currently supported. This can change if we update the model to look more like the operator calling conventions where we could resolve and expand lambas at compile time within function bodies, but this will be deferred for now.


compiler/ztests/lambda-op.yaml
```diff
@@ -0,0 +1,7 @@
+spq: |
+  op foo f : (values f(1))
+  call foo lambda x:x+1
+
+output: |
+  2
+
```
nwt:
```suggestion
```

book/src/super-sql/functions/generics/map.md
```diff
@@ -38,7 +50,19 @@ Using a user-defined function to convert epoch floats to time values:
...
-values map(this, floatToTime)
+values map(this, &floatToTime)
+# input
+[1697151533.41415,1697151540.716529]
+# expected output
+[2023-10-12T22:58:53.414149888Z,2023-10-12T22:59:00.716528896Z]
+```
+
+---
+_Same as above but with an anonymous lambda expression_
```
nwt: Nit: The text above doesn't use "anonymous" so no reason to introduce it here.

```suggestion
_Same as above but with an lambda expression_
```

compiler/ast/ast.go
```diff
@@ -123,11 +123,11 @@ type When struct {
...
 // and returns a result.
 type Call struct {
-	Kind  string `json:"kind" unpack:""`
-	Name  *ID    `json:"name"`
-	Args  []Expr `json:"args"`
-	Where Expr   `json:"where"`
-	Loc   `json:"loc"`
+	Kind    string `json:"kind" unpack:""`
+	Fn      FnRef  `json:"fn"`
+	Actuals []Expr `json:"actuals"`
```
nwt: Nit: I like "func" over "fn" and "args" and "params" over "actuals" and "formals", but I can live with either. But do you like the new terminology enough to do the follow-up PRs to update all the uses of the old terminology? 'Cause there are a lot.

compiler/ast/ast.go
```diff
@@ -339,6 +339,13 @@ type SQLTimeValue struct {
 	Loc   `json:"loc"`
 }
 
+type MapCall struct {
+	Kind string `json:"kind" unpack:""`
+	Expr Expr   `json:"expr"`
+	Fn   FnRef  `json:"func"`
```
nwt:
```suggestion
	Fn   FnRef  `json:"fn"`
```

compiler/semantic/expr.go
```diff
@@ -84,9 +84,10 @@ func (a *analyzer) semExpr(e ast.Expr) dag.Expr {
 	case *ast.Cast:
 		expr := a.semExpr(e.Expr)
 		typ := a.semExpr(e.Type)
+		name := "cast"
 		return &dag.Call{
 			Kind: "Call",
-			Name: "cast",
+			Fn:   &dag.FnName{Kind: "Name", Name: name},
```
```suggestion
			Fn:   &dag.FnName{Kind: "FnName", Name: name},
```

compiler/sfmt/ast.go
```diff
@@ -112,8 +112,9 @@ func (c *canon) expr(e ast.Expr, parent string) {
 		c.write(" : ")
 		c.expr(e.Else, "")
 	case *ast.Call:
-		c.write("%s(", e.Name.Name)
-		c.exprs(e.Args)
+		c.fnRef(e.Fn)
+		c.write("(")
+		c.exprs(e.Actuals)
 		c.write(")")
```
nwt: What about ast.MapCall?

compiler/sfmt/dag.go
```diff
@@ -98,7 +98,7 @@ func (c *canonDAG) expr(e dag.Expr, parent string) {
 		c.write(" : ")
 		c.expr(e.Else, "")
 	case *dag.Call:
-		c.write("%s(", e.Name)
+		c.write("%s(", e.Name())
 		c.exprs(e.Args)
 		c.write(")")
```
nwt: This output is wrong.

```console
$ super compile -C -dag 'op o a: (values a(this))  o lambda x: x+1'
null
| (
  values (this)
  | output main
)
```

compiler/sfmt/ast.go
```diff
@@ -939,7 +993,11 @@ func IsBool(e ast.Expr) bool {
 	case *ast.Conditional:
 		return IsBool(e.Then) && IsBool(e.Else)
 	case *ast.Call:
-		return function.HasBoolResult(e.Name.Name)
+		named, ok := e.Fn.(*ast.FnName)
+		if !ok {
+			return false
+		}
+		return function.HasBoolResult(named.ID.Name)
```
nwt: Nit: Why not name instead of named? And simplify.

```suggestion
		name, ok := e.Fn.(*ast.FnName)
		return ok && function.HasBoolResult(name.ID.Name)
```

compiler/sfmt/ast.go
```diff
@@ -911,7 +961,11 @@ func isAggFunc(e ast.Expr) *ast.Aggregate {
 	if !ok {
 		return nil
 	}
-	if _, err := agg.NewPattern(call.Name.Name, false, true); err != nil {
+	named, ok := call.Fn.(*ast.FnName)
+	if !ok {
+		return nil
+	}
+	if _, err := agg.NewPattern(named.ID.Name, false, true); err != nil {
```
nwt: Nit:

```suggestion
	name, ok := call.Fn.(*ast.FnName)
	if !ok {
		return nil
	}
	if _, err := agg.NewPattern(name.ID.Name, false, true); err != nil {
```

compiler/sfmt/ast.go
```diff
@@ -752,6 +772,36 @@ func (c *canon) op(p ast.Op) {
...
+			continue
+		}
+		switch a := a.(type) {
+		case *ast.FnLambda:
+			c.lambda(a)
+		case *ast.FnName:
+			c.write("&")
+			c.expr(a.ID, "")
+		default:
+			c.write(fmt.Sprintf("unknown type for op call actuals: %T", a))
```
nwt:
```suggestion
			panic(fmt.Sprintf("unknown type for op call actuals: %T", a))
```

compiler/rungen/expr.go
```diff
@@ -308,41 +308,64 @@ func (b *Builder) compileAssignment(node *dag.Assignment) (expr.Assignment, erro
...
 func (b *Builder) compileCall(call dag.Call) (expr.Evaluator, error) {
-	if tf := expr.NewShaperTransform(call.Name); tf != 0 {
-		return b.compileShaper(call.Args, tf)
+	switch fn := call.Fn.(type) {
+	case *dag.FnName:
+		return b.compileCallByName(fn.Name, call.Args)
+	case *dag.Lambda:
+		return b.compileLambda(fn, call.Args)
+	default:
+		panic("compileCall")
```
nwt:
```suggestion
		panic(call)
```

compiler/semantic/expr.go
```diff
@@ -628,33 +618,67 @@ func (a *analyzer) semCall(call *ast.Call) dag.Expr {
...
-	name, nargs := call.Name.Name, len(call.Args)
-	nameLower := strings.ToLower(name)
+	actuals := a.semExprs(call.Actuals)
+	switch fn := call.Fn.(type) {
+	case *ast.FnName:
+		return a.semCallByName(call, fn.ID.Name, actuals)
+	case *ast.FnLambda:
+		return a.semCallLambda(fn, actuals)
+	default:
+		panic("semCall")
```
```suggestion
		panic(fn)
```

compiler/semantic/op.go
```diff
@@ -1209,12 +1229,22 @@ func (a *analyzer) semCallOp(call *ast.CallOp, seq dag.Seq) dag.Seq {
...
-	exprs := a.semExprs(args)
+	exprs := make([]dag.Expr, len(actuals))
+	fns := make([]dag.FnRef, len(actuals))
+	for i, e := range actuals {
+		if expr, ok := e.(ast.Expr); ok {
+			exprs[i] = a.semExpr(expr)
+		} else if fn, ok := e.(ast.FnRef); ok {
+			fns[i] = a.semFnRef(fn)
+		} else {
+			panic("semCallOp")
```
nwt:
```suggestion
			panic(e)
```

compiler/ast/ast.go
mccanne: As I was coding this, I was a bit confused with args/params for calling conventions so I changed the convention to formals/actuals (for just things that have formals) in the AST in line with compiler literature.   There are no formals in the DAG so I left the DAG as args and there isn't an args concept to distinguish between formal and actual.  Arguments for all the other operators makes sense as args sense there are no formals.  Func is Fn here as it's new and in line with the Fn* pattern of other names.  Func isn't really used elsewhere in dag/ast except for Funcs in the scope.  Could change that one to Fns.  I could use FnRef here if you prefer.      Now that I coded up the lambdas, I could change the pattern back to params/args if you prefer but I like formals/actuals to differentiate it from args  usage that doesn't have corresponding formals.

compiler/sfmt/dag.go
mccanne: Hmm need more test coverage.  Will work on this...

compiler/ast/ast.go
nwt: I don't much care which names we use so long as we use them consistently. Using both "actuals" and "args" just isn't consistent. Neither is using both "fn" and "func".

If we adopt new names, someone is going to do a bunch of typing to make things consistent again. Neither of us wants to do that work, and that's really my argument for sticking with the existing names.

compiler/sfmt/dag.go
mccanne: Fixed it and added test

compiler/ast/ast.go
```diff
@@ -756,3 +763,35 @@ type Agg struct {
...
+type FnLambda struct {
+	Kind    string `json:"kind" unpack:""`
+	Formals []*ID  `json:"formals"`
+	Expr    Expr   `json:"expr"`
+	Loc     `json:"loc"`
+}
+
+type FnName struct {
+	Kind string `json:"kind" unpack:""`
+	ID   *ID    `json:"id"`
```
nwt: Nit: Is there any upside to using ID here instead of string?

```suggestion
	Text string `json:"text"`
```

compiler/sfmt/ast.go
```diff
@@ -939,7 +1020,8 @@ func IsBool(e ast.Expr) bool {
 	case *ast.Conditional:
 		return IsBool(e.Then) && IsBool(e.Else)
 	case *ast.Call:
-		return function.HasBoolResult(e.Name.Name)
+		named, ok := e.Fn.(*ast.FnName)
+		return ok && function.HasBoolResult(named.ID.Name)
```
nwt: Nit:

```suggestion
		name, ok := e.Fn.(*ast.FnName)
		return ok && function.HasBoolResult(name.ID.Name)
```

compiler/rungen/op.go
```diff
@@ -478,8 +478,8 @@ func (b *Builder) compileScope(scope *dag.Scope, parents []sbuf.Puller) ([]sbuf.
 	parentUDFs := b.udfs
 	b.udfs = maps.Clone(parentUDFs)
 	defer func() { b.udfs = parentUDFs }()
-	for _, f := range scope.Funcs {
-		b.udfs[f.Name] = f
+	for k, f := range scope.Funcs {
+		b.udfs[f.Name] = &scope.Funcs[k].Lambda
```
nwt: Nit:

```suggestion
	for _, f := range scope.Funcs {
		b.udfs[f.Name] = &f.Lambda
```

compiler/rungen/vop.go
```diff
@@ -205,8 +205,8 @@ func (b *Builder) compileVamScope(scope *dag.Scope, parents []vector.Puller) ([]
 	parentUDFs := b.udfs
 	b.udfs = maps.Clone(parentUDFs)
 	defer func() { b.udfs = parentUDFs }()
-	for _, f := range scope.Funcs {
-		b.udfs[f.Name] = f
+	for k, f := range scope.Funcs {
+		b.udfs[f.Name] = &scope.Funcs[k].Lambda
```
nwt:
```suggestion
	for _, f := range scope.Funcs {
		b.udfs[f.Name] = &f.Lambda
```

compiler/semantic/expr.go
```diff
@@ -84,9 +84,10 @@ func (a *analyzer) semExpr(e ast.Expr) dag.Expr {
 	case *ast.Cast:
 		expr := a.semExpr(e.Expr)
 		typ := a.semExpr(e.Type)
+		name := "cast"
 		return &dag.Call{
 			Kind: "Call",
-			Name: "cast",
+			Fn:   &dag.FnName{Kind: "FnName", Name: name},
```
```suggestion
			Fn:   &dag.FnName{Kind: "FnName", Name: "cast"},
```

compiler/semantic/expr.go
```diff
@@ -628,33 +618,67 @@ func (a *analyzer) semCall(call *ast.Call) dag.Expr {
...
+			Kind:    "Lambda",
+			Formals: formals,
+			Expr:    a.semExpr(lambda.Expr),
+		},
+		Args: actuals,
+	}
+}
+
+func (a *analyzer) semCallByName(call *ast.Call, name string, actuals []dag.Expr) dag.Expr {
+
```
```suggestion
```

compiler/parser/parser.peg
```diff
@@ -42,23 +42,45 @@ ConstDecl
...
+          Formals: sliceOf[*ast.ID](formals),
+          Expr: expr.(ast.Expr),
+          Loc: loc(c),
+        },
+        Loc: loc(c),
+      }, nil
+    }
+
+LambdaExpr
+  = "lambda" formals:(_ ids:Identifiers { return ids, nil })? __ ":" __ expr:Expr {
```
nwt:
```suggestion
  = "lambda"i formals:(_ ids:Identifiers { return ids, nil })? __ ":" __ expr:Expr {
```

compiler/semantic/expr.go
```diff
@@ -810,15 +835,19 @@ func deriveNameFromExpr(e dag.Expr, a ast.Expr) []string {
 	case *dag.Agg:
 		return []string{e.Name}
 	case *dag.Call:
-		switch strings.ToLower(e.Name) {
+		name := e.Name()
+		if name == "" {
+			name = "lambda"
+		}
```
nwt: Nit: Maybe move this into dag.Call.Name?

compiler/sfmt/dag.go
```diff
@@ -214,6 +215,24 @@ func (c *canonDAG) binary(e *dag.BinaryExpr, parent string) {
...
+				c.write(", ")
+			}
+			c.write(fn.Formals[i])
+		}
+		c.write(":")
+		c.expr(fn.Expr, "")
+		c.write(")")
+	case *dag.FnName:
+		c.write(fn.Name)
+	}
```
nwt, line 235:
```suggestion
	default:
		panic(fn)
	}
```

compiler/semantic/op.go
```diff
@@ -1033,17 +1032,23 @@ func (a *analyzer) semTypeDecl(d *ast.TypeDecl) {
 	}
 }
 
-func (a *analyzer) semFuncDecls(decls []*ast.FuncDecl) []*dag.Func {
-	funcs := make([]*dag.Func, 0, len(decls))
+func formalsAsStrings(formals []*ast.ID) []string {
```
nwt: Use idsToStrings.

nwt (approved):

compiler/rungen/vop.go
mccanne: Isn't this a bug because go copies the value into f or did they fix this?

PR #6233, opened 2025-09-14
remove scopes from dag and simplify functions, fix scoping bugs

mccanne:
This commit removes dag.Scope and simplifies function handling in the DAG by having a global table functions indexed by a tag. There is no need for scoping in the DAG since lexical scoping is all resolved by the end of the semantic pass.

This undoes the complexity of the recent dag.FuncRef addition, which took things in the wrong direction (my bad).

All lambda functions are now compiled alongside named user functions in the DAG function table.  This sets us up to handle passing lambdas to functions (in addition to ops), but this will come in a subsequent PR.

This commit also fixes some scoping problems where the lookup-by-name approach for function resolution didn't work quite right.  See new tests in compiler/semantic/ztests.


cmd/super/compile/shared.go
```diff
@@ -90,8 +90,7 @@ func (s *Shared) Run(ctx context.Context, args []string, dbFlags *dbflags.Flags,
 		return s.writeValue(ctx, description)
 	}
 	if s.optimize {
-		dag, err = compiler.Optimize(rctx, dag, env, s.parallel)
-		if err != nil {
+		if err = compiler.Optimize(rctx, dag, env, s.parallel); err != nil {
```
nwt: Nit:

```suggestion
		if err := compiler.Optimize(rctx, dag, env, s.parallel); err != nil {
```

compiler/dag/op.go
```diff
@@ -129,11 +134,7 @@ type (
 		Kind  string `json:"kind" unpack:""`
 		Paths []Seq  `json:"paths"`
 	}
-	Scope struct {
-		Kind  string     `json:"kind" unpack:""`
-		Funcs []*FuncDef `json:"funcs"`
-		Body  Seq        `json:"seq"`
-	}
+
```
nwt: Nit:

```suggestion
```

compiler/semantic/ztests/op-func-thunk-esc.yaml
```diff
@@ -0,0 +1,10 @@
+spq: |
+  fn outer(x):x+1
+  op foo f,g,val: (values f(val),g(val))
+  (
+    fn inner(x):x
+    foo &inner,&outer,10
+  )
+output: |
```
nwt: Nit:

```suggestion
  )

output: |
```

compiler/semantic/analyzer.go
```diff
@@ -35,8 +38,16 @@ func Analyze(ctx context.Context, p *parser.AST, env *exec.Environment, extInput
 			seq.Prepend(&dag.NullScan{Kind: "NullScan"})
 		}
 	}
+	var funcs []*dag.FuncDef
+	for _, f := range a.funcs {
+		funcs = append(funcs, f)
+	}
```
nwt: Nit:

```suggestion
	funcs := slices.Collect(maps.Values(a.funcs))
```

compiler/semantic/analyzer.go
```diff
@@ -89,6 +100,18 @@ func (a *analyzer) exitScope() {
 	a.scope = a.scope.parent
 }
 
+func (a *analyzer) newFunc(name string, params []string, e dag.Expr) string {
+	tag := fmt.Sprintf("%d", len(a.funcs))
```
nwt: Nit:

```suggestion
	tag := strconv.Itoa(len(a.funcs))
```

compiler/semantic/analyzer.go
```diff
@@ -35,8 +38,16 @@ func Analyze(ctx context.Context, p *parser.AST, env *exec.Environment, extInput
...
 		}
 	}
+	var funcs []*dag.FuncDef
+	for _, f := range a.funcs {
+		funcs = append(funcs, f)
+	}
+	// Sort by tag so we get deterministic sfmt output.
+	slices.SortFunc(funcs, func(a, b *dag.FuncDef) int {
+		return strings.Compare(a.Tag, b.Tag)
+	})
```
nwt: Nit: If this is for the sake of sfmt, is there any reason not to do it there?

compiler/semantic/scope.go
```diff
@@ -72,24 +72,18 @@ func (s *Scope) LookupExpr(name string) (dag.Expr, error) {
...
+func (s *Scope) LookupFunc(name string) (*dag.FuncDef, string, error) {
+	entry := s.lookupEntry(name)
+	if entry == nil {
+		return nil, "", nil
+	}
+	if f, ok := entry.ref.(*dag.FuncDef); ok {
+		return f, "", nil
+	}
+	if b, ok := entry.ref.(*builtin); ok {
+		return nil, string(*b), nil
```

```suggestion
	switch ref := entry.ref.(type) {
	case *dag.FuncDef:
		return ref, "", nil
	case *builtin:
		return nil, string(*ref), nil
```

compiler/sfmt/dag.go
```diff
@@ -627,25 +604,18 @@ func (c *canonDAG) unnest(u *dag.Unnest) {
...
-		for i := range f.Lambda.Params {
+func (c *canonDAG) main(main *dag.Main) {
+	for _, f := range main.Funcs {
+		c.write("fn %s/%s(", f.Tag, f.Name)
+		for i := range f.Params {
 			if i != 0 {
 				c.write(", ")
 			}
-			c.write(f.Lambda.Params[i])
+			c.write(f.Params[i])
```
nwt: Nit:

```suggestion
		for i, p := range f.Params {
			if i != 0 {
				c.write(", ")
			}
			c.write(p)
```

compiler/package.go
```diff
@@ -144,20 +146,19 @@ func VectorFilterCompile(rctx *runtime.Context, query string, env *exec.Environm
...
 	}
-	scan, ok := seq[0].(*dag.DefaultScan)
+	scan, ok := main.Body[0].(*dag.DefaultScan)
 	if !ok {
 		return nil, errors.New("CompileWithSortKey: expected a reader")
 	}
 	scan.SortKeys = order.SortKeys{sortKey}
-	seq, err = Optimize(rctx, seq, env, 0)
-	if err != nil {
+	if err = Optimize(rctx, main, env, 0); err != nil {
```
nwt: Nit:

```suggestion
	if err := Optimize(rctx, main, env, 0); err != nil {
```

nwt (approved):

compiler/semantic/analyzer.go
mccanne: sfmt should print what's there.  I'll update the comment.

Issue #6223, opened 2025-09-11
SQL: Query with math in WHERE stopped working at super c018cba


philrz:
## tl;dr

The following query returns no output, but worked ok at the commit before the merge of #6219.

```
$ super -f parquet -o t1 -c "values {a:1,b:2,c:3}" &&
  super -c "SELECT a,b,c FROM t1 (FORMAT parquet) WHERE b+1 BETWEEN a+1 and c+1"

[no output]
```

## Details

Repro is with super commit c018cba. This is a simplified version of the query from [a sqllogictest](https://github.com/duckdb/duckdb/blob/f6b2306dfd7368bfc61bee6ca6f42d0c4b269b65/test/sqlite/select1.test_slow#L5918-L5927) that originally surfaced the issue.

The query worked ok at the prior commit ec253be that was right before the merge of the changes in #6219.

```
$ super -version
Version: ec253be2b

$ super -f parquet -o t1 -c "values {a:1,b:2,c:3}" &&
  super -c "SELECT a,b,c FROM t1 (FORMAT parquet) WHERE b+1 BETWEEN a+1 and c+1"

{a:1,b:2,c:3}
```

It also works ok at the problematic commit if I drop the math from the `WHERE` clause.

```
$ super -version
Version: c018cba16

$ super -f parquet -o t1 -c "values {a:1,b:2,c:3}" &&
  super -c "SELECT a,b,c FROM t1 (FORMAT parquet) WHERE b BETWEEN a and c"

{a:1,b:2,c:3}
```

philrz:
Verified in super commit 0e9052b.

The query now returns the expected result when math is used in the `WHERE` clause.

```
$ super -version
Version: 0e9052b6d

$ super -f parquet -o t1 -c "values {a:1,b:2,c:3}" &&   super -c "SELECT a,b,c FROM t1 (FORMAT parquet) WHERE b+1 BETWEEN a+1 and c+1"

{a:1,b:2,c:3}
```

Thanks @nwt!

PR #6236, opened 2025-09-16
Makefile targets to run sqllogic ztests

philrz:
## What's Changing

This is a first attempt at running some of the successful sqllogictest queries as ztests to catch SuperDB regressions. I expect I'll need some help improving the Makefile. To run:

```
make test-sqllogic
```

## Why

While working on sqllogictests to find new problems I've bumped into some regressions such as #6223, so there's been recent interest within the Dev team in "locking in" the automated running of successful tests so such breakages can be caught early.

## Details

As this first step, I'm just trying to make it possible to run the tests and will leave it up to group consensus where this fits into the Dev workflow, e.g., if/when it runs in CI.

What I've got in this PR is basically following the existing "sampledata" pattern in the Makefile. I've noticed that once the `test-sqllogic/` directory exists, the logic that finds all the ztest YAML in the source tree seems to run these tests even in targets like `test-ci` that don't explicitly target these tests, so this is one area where I'll surely need help to improve.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,13 @@ $(SAMPLEDATA):
 
 sampledata: $(SAMPLEDATA)
 
+SQLLOGIC_ZTESTS:=sqllogic-ztests/README.md
+
+$(SQLLOGIC_ZTESTS):
+	git clone --depth=1 https://github.com/brimdata/sqllogic-ztests $(@D)
+
+sqllogic-ztests: $(SQLLOGIC_ZTESTS)
+
 bin/minio: Makefile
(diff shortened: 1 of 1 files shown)


philrz:
If anyone is curious how these would fare in CI, I just did a [run](https://github.com/brimdata/super/actions/runs/17750915173/job/50445525549) on a branch and the step that runs these tests took 4.5 minutes. However, that should go much faster once we've finished the optimizations for filtered cross joins.

nwt (approved):
I couldn't come up with anything better.

PR #6243, opened 2025-09-17
Simplify syntax for array-wrapped subquery

mattnibs:
This commit allows users to specify an array-wrapped subquery using only brackets wrapping the subquery and not requiring the parenthesis usually needed for subqueries.


mccanne (approved):
cool

PR #6244, opened 2025-09-18
allow functions/lambdas to be passed as args to functions

mccanne:
This commit adds support for function arguments to functions. Since functions are not first-class, the approach here is to do a compile-time analysis of passed-in functions and create variations of each function that takes different passed-in functions.  The variation still calls the reference function by name by does so in a static fashion, which means each variation of static binding must be unraveled to a separate function definition in the DAG. This is done in two passes in the semantic phase, where the first pass decorages the DAG with function references and the second pass creates the variations based on the actual references passed in.

We also changed the syntax so that user functions are referenced as values simply with their name while built-in functions use the &id syntax, e.g., myFunction vs. &upper.

We also fixed a bug where const decls could override the name of a parameter when the parameter should override the const.  There's a new test to cover this.

These changes also meant that ast.MapCall could be removed since ast.Call now covers this use case.  We left dag.MapCall in place for now but if/when we add more built-ins that take function args, we can generalize dag.MapCall and extend the runtime package to support different variations of such functions.


compiler/semantic/expr.go
```diff
@@ -760,36 +818,23 @@ func (a *analyzer) maybeSubqueryCall(call *ast.Call, name string) *dag.Subquery
...
+		a.error(loc, errors.New("map requires two arguments"))
+		return badExpr()
 	}
-	return &dag.MapCall{
+	f, ok := args[1].(*dag.FuncRef)
+	if !ok {
+		a.error(loc, errors.New("second argument to map must be a function"))
+		return badExpr()
+	}
+	e := &dag.MapCall{
```
mattnibs, line 831: Should we just get rid of MapCall have this be a subquery with a collect on it? Or is this something you wanted to do in a subsequent pr?

compiler/semantic/resolver.go
```diff
@@ -0,0 +1,580 @@
...
+type resolver struct {
+	analyzer *analyzer
+	in       map[string]*dag.FuncDef
+	fixed    map[string]*dag.FuncDef
+	variants []*dag.FuncDef
+	params   []map[string]string
+	ntag     int
+}
+
+func newResolver(a *analyzer, in map[string]*dag.FuncDef) *resolver {
```
mattnibs: Maybe just pass in analyzer since `in` is derived from analyzer.funcsByTag?

compiler/semantic/resolver.go
```diff
@@ -0,0 +1,580 @@
...
+func (r *resolver) resolve(seq dag.Seq) (dag.Seq, []*dag.FuncDef) {
+	out := r.resolveSeq(seq)
+	funcs := r.variants
+	for _, f := range r.fixed {
+		funcs = append(funcs, f)
+	}
+	return out, funcs
+}
+
+// resolve traverses a DAG and substitues all instances of CallParam by rewriting
```
mattnibs:
```suggestion
// resolveSeq traverses a DAG and substitutes all instances of CallParam by rewriting
```

compiler/semantic/resolver.go
```diff
@@ -0,0 +1,580 @@
...
+func (r *resolver) nextTag() string {
+	tag := strconv.Itoa(r.ntag)
+	r.ntag++
+	return tag
+}
+
+func (r *resolver) bindParam(param, tag string) {
+	r.params[len(r.params)-1][param] = tag
+}
+
```
mattnibs: This function is unused:
```suggestion
```

compiler/semantic/expr.go
mattnibs: As is the MapCall isn't implemented in the vector runtime so this would allow it to run there.

mccanne: Good idea but can we do this on a subsequent PR?

mattnibs (approved):

PR #6251, opened 2025-09-20
support for restartable dag.FileScans so repeated from-subqueries work

mccanne:
This commit adds support for restarting file scans so that repeated from-subqueries involving file paths work. This is a temporary solution to make progress on docs and testing.  It does not work for databases and fails silently for non-seekable streams.

A more comprhensive solution will come in a future PR.

### DIFF
--- a/runtime/vam/op/hashjoin.go
+++ b/runtime/vam/op/hashjoin.go
@@ -324,7 +324,7 @@ func (b *bufPuller) pull(done bool) (vector.Any, error) {
 	if vec == nil || err != nil {
 		b.done = true
 	}
-	return vec, nil
+	return vec, err
 }
 
 func hashKey(val super.Value) string {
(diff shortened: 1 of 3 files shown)


runtime/vam/op/hashjoin.go
```diff
@@ -312,7 +313,18 @@ func (b *bufPuller) Pull(done bool) (vector.Any, error) {
...
+
+func (b *bufPuller) pull(done bool) (vector.Any, error) {
+	if b.done {
+		return nil, nil
+	}
+	vec, err := b.puller.Pull(done)
+	if vec == nil || err != nil {
+		b.done = true
+	}
+	return vec, nil
```
nwt:
```suggestion
	return vec, err
```

nwt (approved):

PR #6253, opened 2025-09-20
add support for expression-only fields in records

mccanne:
This commit adds a convenient shortcut for record expressions to allow a standalone expression as a field producing a name for the field in the saw way that SQL SELECT expressions create names for columns when the AS clause is absent.


compiler/semantic/expr.go
```diff
@@ -213,13 +214,27 @@ func (a *analyzer) semExpr(e ast.Expr) dag.Expr {
...
+			if _, ok := fields[name]; ok {
+				a.error(elem, fmt.Errorf("record expression: %w", &super.DuplicateFieldError{Name: name}))
+				continue
+			}
+			fields[name] = struct{}{}
+			out = append(out, &dag.Field{
+				Kind:  "Field",
+				Name:  name,
+				Value: e,
+			})
```
nwt: Nit: I think what's going on would be a little clearer if you promote this block to be the default case in the preceding type switch instead of adding continue at the end of its existing cases.

Issue #5977, opened 2025-06-23
SQL: Duplicate column names in query results


philrz:
## tl;dr

SuperSQL does not yet support duplicate column names in query results as is supported in standard SQL.

## Details

Repro is with super commit 8f4cd56. This was found via a query from [a sqllogictest](https://github.com/duckdb/duckdb/blob/0cbb56d14cc1464d945a06cb8bac051485fa2ef7/test/sql/aggregate/group/test_group_by.test#L176-L183).

The referenced sqllogictest is actually trying to test something more subtle, but seeing it fail with `super` reminded me that the more fundamental ability to cope with duplicate column names in query results is still on our SuperSQL to-do list.

With input data `integers2.sup`:

```
{i:1::int32}
{i:2::int32}
{i:3::int32}
{i:null::int32}
```

The following query produces column names `i` and `i_1`.

```
$ super -version
Version: 8f4cd564b

$ super -c "SELECT i, i % 2 AS i FROM 'integers2.sup';"
{i_1:1::int32,i:1}
{i_1:2::int32,i:0}
{i_1:3::int32,i:1}
{i_1:null::int32,i:null::int64}
```

Whereas in Postgres, we get two columns both with name `i`.

```
$ psql postgres
psql (17.5 (Homebrew))
Type "help" for help.

# SELECT * FROM integers2;
 i 
---
 1
 2
 3
  
(4 rows)

postgres=# SELECT i, i % 2 AS i FROM integers2;
 i | i 
---+---
 1 | 1
 2 | 0
 3 | 1
   |  
(4 rows)
```

philrz:
I think this is another example that would require support for duplicate column names, but since it manifests as absent columns rather than the name with `_1` appended I'm logging it separately just in case it requires special treatment.

## Details

Repro is with super commit e75dfda. The repro query shown above was taken from [a sqllogictest](https://github.com/duckdb/duckdb/blob/223ff0a7dba7900039d5910a009247ef097fff3c/test/sql/join/cross_product/test_cross_product.test#L14-L20).

Here's the query with test data running successfully in Postgres. Note it shows two columns named `a` and two named `b`.

```
$ psql postgres
psql (17.5 (Homebrew))
Type "help" for help.

postgres=# CREATE TABLE test (a INTEGER, b INTEGER);
CREATE TABLE

postgres=# INSERT INTO test VALUES (11, 1), (12, 2);
INSERT 0 2

postgres=# SELECT * FROM test t1, test t2;
[...]

PR #6263, opened 2025-09-22
unify SQL and pipe name inference and simplify single by-only agg output

### LINKED ISSUE #5315: Cast shorthand in record expression creates field name "cast"
As a user, the field name `cast` seems unexpected here.

```
$ echo '{x:1, y:2}' | zq 'cut x,float64(y)' -
{x:1,cast:2.}
```

## Details

Repro is with Zed commit b8fe312.

In the language, field names are commonly derived from function names when the calls are made in record expressions like this. For instance:

```
$ zq -version
Version: v1.18.0-11-gb8fe3121

$ echo '{x:1, y:-2}' | zq 'cut x,abs(y)' -
{x:1,abs:2}
```

And of course, if a user wants to explicitly assign a field name to avoid this effect, they can include explicit assignment.

```
$ echo '{x:1, y:-2}' | zq 'cut x,y:=abs(y)' -
{x:1,y:2}
```

That said, going back to our first example, the field name `cast` would only make sense if the user knows the implementation detail of `float64()` being a wrapper around the `cast()` function. Instead one of the following would seem more intuitive:

1. Using the field name `float64` to match the behavior we just saw with `abs()`.
2. Maintaining the original field name (`y` in this case).

In a group discussion, there was consensus on maintaining the original field name, which happens to match my instincts as well.

In that same discussion, @mattnibs pointed out a related surprise that when dealing with nested fields, this casting operation creates a top-level field, whereas he's felt it should keep the value in its nested path location.

```
$ echo '{x:1, y:2, a: {b: 3}}' | zq 'cut x,float64(a.b)' -
{x:1,cast:3.}
```

If we don't address that when taking up this issue, I can open a separate issue at that time.

### LINKED ISSUE #6157: Dotted path expr in field assignment should create value named for last element in path
## tl;dr

The docs updates in #6128 imply that this:

```
$ super -version
Version: 6bbf94507

$ echo '{x:{y:1}}' | super -c "put x.y" -
{x:{y:1}}
```

should have the same output as:

```
$ echo '{x:{y:1}}' | super -c "put y:=x.y" -
{x:{y:1},y:1}
```

## Details

Repro is with super commit 6bbf945.

The full example in which this appeared in #6128 can be seen [here](https://github.com/brimdata/super/blob/6bbf94507c8a60266b2f55db555bad9089d7bd75/book/src/super-sql/operators/intro.md?plain=1#L69-L87). @mccanne and @philrz had some further discussion of the topic in a follow-on PR/comment https://github.com/brimdata/super/pull/6151#issuecomment-3217132816.

mccanne:
This commit unifies the way l-values are inferred for expressions that omit the LHS of a field assignment or the AS clause in a select SELECT expression.  ANSI SQL does not have specific guidance on how to infer arbitrary expressions so we blended in some of the existing logic from the pipe syntax.

We also changed how "this" is handled.  Instead of raising an error, the l-value becomes "that".  We also adapted the heuristic for a single column agg with no groupings to work also for a single grouping with no aggs (i.e., emitting the values instead of a single-column record). These two changes  means that "by this" now works and behaves like distinct.

Overall, this improves the consistency and orthogonality of the language.

Fixes #6157
Fixes #5315

### DIFF
--- a/compiler/semantic/ztests/by-this.yaml
+++ b/compiler/semantic/ztests/by-this.yaml
@@ -1,4 +1,4 @@
-spq: by this | sort
+spq: by this | sort this
 
 input: |
   1


compiler/semantic/ztests/field-expr.yaml
```diff
@@ -5,4 +5,4 @@ input: |
   {x:{y:123},s:"Hello, world", z:0.5}
 
 output: |
-  {y:123,"upper(s)":"HELLO, WORLD","sqrt(2)/z":2.8284271247461903}
+  {y:123,upper:"HELLO, WORLD","sqrt(2)/z":2.8284271247461903}
```
philrz, line 8: I'm sure there'll always be ways to trip over this, but the loss of the `(s)` feels like maybe a step backwards since now it's much easier to hit this:

```
$ echo '{x:{y:123},s:"Hello, world", r:"bye", z:0.5}' | super -c 'values {x.y,upper(s),upper(r),sqrt(2)/z}' -
record expression: duplicate field: "upper" at line 1, column 22:
values {x.y,upper(s),upper(r),sqrt(2)/z}
                     ~~~~~~~~
```

But I guess a temporary problem since we'll eventually get to #5977?

runtime/ztests/op/aggregate/count-by-this-plain.yaml
philrz, line 1: I assume the name of this test should change (or be deleted entirely?) since its original raison d'etre seemed to be confirming that query produced an error and as your change shows it no longer does that.

compiler/semantic/ztests/field-expr.yaml
mccanne: Yes temporary but I didn't realize the extent of the impact here.

Issue #5984, opened 2025-06-26
SQL: NULL values absent from JOIN output


philrz:
When Postgres runs this query, it has `NULL` values for `k` and `l` in the bottom output record.

```
$ super -f parquet -o integers.parquet -c "
values
  {i:1::int32,j:2::int32},
  {i:2::int32,j:3::int32},
  {i:3::int32,j:4::int32}
" &&
super -f parquet -o integers2.parquet -c "
values
  {k:1::int32,l:10::int32},
  {k:2::int32,l:20::int32}
" &&
super -c "SELECT * FROM integers.parquet LEFT OUTER JOIN integers2.parquet ON integers.i=integers2.k ORDER BY i;"

{i:1::int32,j:2::int32,k:1::int32,l:10::int32}
{i:2::int32,j:3::int32,k:2::int32,l:20::int32}
{i:3::int32,j:4::int32}
```

## Details

Repro is with super commit a90585f. This was found via a query from [a sqllogictest](https://github.com/duckdb/duckdb/blob/d524c57a58c4984406ca7ee7251bb8603614eb7a/test/sql/join/left_outer/test_left_outer.test#L23-L28).

Here's the query running in Postgres with a client setting to make it easier to see `NULL` values.

```
$ psql --version
psql (PostgreSQL) 17.6 (Homebrew)

$ psql postgres -P null=NULL -c "
DROP TABLE IF EXISTS integers;
DROP TABLE IF EXISTS integers2;
CREATE TABLE integers(i INTEGER, j INTEGER);
INSERT INTO integers VALUES (1, 2), (2, 3), (3, 4);
CREATE TABLE integers2(k INTEGER, l INTEGER);
INSERT INTO integers2 VALUES (1, 10), (2, 20);
SELECT * FROM integers LEFT OUTER JOIN integers2 ON integers.i=integers2.k ORDER BY i;"

DROP TABLE
DROP TABLE
CREATE TABLE
INSERT 0 3
CREATE TABLE
INSERT 0 2

 i | j |  k   |  l   
---+---+------+------
 1 | 2 |    1 |   10
 2 | 3 |    2 |   20
 3 | 4 | NULL | NULL
(3 rows)
```

Whereas in SuperDB, the bottom output record lacked values for the fields `k` and `l`.

Now, as a user, I know that if I want the `NULL` values, I can just tack on a piped `fuse`.

```
$ super -version
Version: a90585f6d

$ super -c "
[...]

PR #6264, opened 2025-09-23
SQL select star static schema column names

mattnibs:
This commit changes the generated values operator dag in a SQL query that uses select * so that if the underlying schema is static all the selected columns will be enumerated instead of using record spread expressions. This change allows us to optimize queries that could otherwise not be optimized since the source of a field is obfuscated when spreads are used.

### DIFF
--- a/compiler/semantic/sql.go
+++ b/compiler/semantic/sql.go
@@ -108,12 +108,7 @@ func (a *analyzer) genColumns(proj projection, sch *selectSchema, seq dag.Seq) d
 			notFirst = true
 		}
 		if col.isStar() {
-			for _, path := range unravel(sch, nil) {
-				elems = append(elems, &dag.Spread{
-					Kind: "Spread",
-					Expr: dag.NewThis(path),
-				})
-			}
+			elems = unravel(elems, sch, nil)
 		} else {
 			elems = append(elems, &dag.Field{
 				Kind:  "Field",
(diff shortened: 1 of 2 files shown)


philrz:
I'm not suggesting any action right now related to this, but just want to mention that seeing this PR made me curious about its impact on #5984. The tl;dr is that it seems to get us closer, because before the query results had spots filled in with `NULL` in Postgres that were fully absent in SuperDB (potentially entire columns worth), whereas on this PR's branch those spots are now filled in with `error("missing")` values. Compare these with what's shown in the original repros in #5984:

```
$ super -version
Version: e96d1be11

$ super -f parquet -o integers.parquet -c "
values
  {i:1::int32,j:2::int32},
  {i:2::int32,j:3::int32},
  {i:3::int32,j:4::int32}
" && super -f parquet -o integers2.parquet -c "
values
  {k:1::int32,l:10::int32},
  {k:2::int32,l:20::int32}
" && super -c super -c "SELECT * FROM integers.parquet LEFT OUTER JOIN integers2.parquet ON integers.i=integers2.k ORDER BY i;"

{i:1::int32,j:2::int32,k:1::int32,l:10::int32}
{i:2::int32,j:3::int32,k:2::int32,l:20::int32}
{i:3::int32,j:4::int32,k:error("missing"),l:error("missing")}
```

```
$ super -c "
SELECT * FROM integers.parquet LEFT OUTER JOIN integers2.parquet ON 1=2 ORDER BY i;"

{i:1::int32,j:2::int32,k:error("missing"),l:error("missing")}
{i:2::int32,j:3::int32,k:error("missing"),l:error("missing")}
{i:3::int32,j:4::int32,k:error("missing"),l:error("missing")}
```

PR #6267, opened 2025-09-26
add semantic tree

mccanne:
This commit adds a new data structure that represents the query tree in between the AST and the DAG.  This will set us up to bring type checking to dynamic data using the semantic tree.

Each entity in the semantic tree has an exported ast.Node interface pointer so everything implements ast.Node and locators still work (unlike in the DAG).

We also fixed how consts are evaluated and how EvalAtCompileTime is invoked.  In this new approach, we re-enter the semantic analysis for any constant function needing eval and have a "maybe" version that ignores errors and a "must" version that captures errors.  Functions and subqueries are now properly handled when evaluating consts.  We also walk the sem tree to figure out which expressions are const instead of relying on the unreliable error("missing") hack that we previously used.

Expression matching for SQL group-by expressions is now done on the sem tree (instead of DAG) and utilizes some deep-equal logic that traverses the tree.

This is some formatting logic for sem tree in compiler/semantic/fmt.go, which is useful for printf debugging and will be tied into "super compile" in a subsequent PR.

There are still some problems with declaration dependencies (e..g, fn decls that depend on const decls in same block and vice versa) and we added some skipped tests for this.  We will fix this in an upcoming branch decl-dep.

Type checking will come in a subsequent PR.


compiler/dag/op.go
```diff
@@ -22,6 +22,11 @@ type Main struct {
 	Body  Seq        `json:"body"`
 }
 
+type MainExpr struct {
+	Funcs []*FuncDef `json:"funcs"`
+	Expr  Expr       `json:"expr"`
+}
+
```
nwt: Nit: This is more expression than operator so maybe put it in expr.go?

compiler/semantic/sem/expr.go
```diff
@@ -0,0 +1,252 @@
...
+		},
+	}
+	return &CallExpr{
+		Node: n,
+		Tag:  "error",
+		Args: []Expr{rec},
+	}
+}
+
+// XXX
```
nwt: What's this about?

compiler/semantic/sem/op.go
```diff
@@ -0,0 +1,291 @@
+// Package sem represents an semantic-internal AST representation that
```
nwt:
```suggestion
// Package sem represents a semantic-internal AST representation that
```

compiler/semantic/sem/op.go
```diff
@@ -0,0 +1,291 @@
...
+// is used for analysis between the AST and DAG representations.
+// This is never serialized so there are no Kind fields or unpacker hooks.
+// The sem AST has no scope and all functions, ops, etc are resolved
```
nwt:
```suggestion
// The sem AST has no scope and all functions, ops, etc. are resolved
```

compiler/semantic/sem/op.go
```diff
@@ -0,0 +1,291 @@
...
+// to flattened tables.  Likewise, all SQL  structure is transalted to
```
nwt:
```suggestion
// to flattened tables.  Likewise, all SQL structure is translated to
```

compiler/semantic/resolver.go
```diff
@@ -4,31 +4,40 @@ import (
...
-func (r *resolver) resolve(seq dag.Seq) (dag.Seq, []*dag.FuncDef) {
-	out := r.resolveSeq(seq)
+func (r *resolver) resolve(seq sem.Seq) (sem.Seq, []*sem.FuncDef) {
+	out := r.seq(seq)
+	funcs := r.variants
+	for _, f := range r.fixed {
+		funcs = append(funcs, f)
+	}
+	return out, funcs
+}
```
nwt, line 38: Nit: This can reuse r.variant's underlying array across calls. Clearly that isn't a problem right now but you might want to future-proof with `funcs := slices.Clone(r.variants)`.

nwt (approved):

compiler/semantic/sem/expr.go
mccanne: Oops, removed.  I'll address the thought in another PR.

PR #6268, opened 2025-09-27
improve declaration bindings with lazy evaluation

mccanne:
This commit improves how declaration bindings work in a given scope by binding all the symbols as encountered then resolving declaration bodies on demand.  This means that dependencies are handled in an as-used basis and the order of declaration does not matter (for a given scope block).

This is needed by the forthcoming type checker.

### DIFF
--- a/compiler/ztests/udf-err.yaml
+++ b/compiler/ztests/udf-err.yaml
@@ -38,4 +38,3 @@ outputs:
       "q" is not a function in call-query-err.spq at line 2, column 8:
       values q(this)
              ~~~~~~~
- 


compiler/ztests/udf-err.yaml
```diff
@@ -30,3 +35,7 @@ outputs:
       "f": expected 2 params but called with 1 in wrong-args.spq at line 2, column 8:
       values f(this)
              ~~~~~~~
+      "q" is not a function in call-query-err.spq at line 2, column 8:
+      values q(this)
+             ~~~~~~~
+
```
nwt: Nit:

```suggestion
```

PR #6272, opened 2025-09-30
Add support for join filter pullups

### LINKED ISSUE #6074: SQL: Large cartesian product causes very long query runtime
## tl;dr

Even when they contain filters that might reduce the result set, queries containing cross joins that produce a large cartesian product currently have running times that seem to hang indefinitely.

## Details

Repro is with super commit 230007c. This was found via a query from [a sqllogictest](https://github.com/duckdb/duckdb/blob/041b68304fee39b0eb770f745d1b282b7f9fe004/test/sqlite/select4.test_slow#L29191-L29200).

Many of the sqllogictests in that particular suite operate perform a cross join that, based on the `SELECT` and `FROM` alone, would generate a massive cartesian product, but they also include filtering that cause them to ultimately produce a small result.

To see how Postgres handles it, we'll load the data using the attached [ddl.txt](https://github.com/user-attachments/files/21497338/ddl.txt) and time the running of the query.

```
$ psql --version
psql (PostgreSQL) 17.5 (Homebrew)

$ psql postgres -q -c "CREATE DATABASE foo;"

$ psql foo -q -f ddl.txt

$ time psql foo -c "
SELECT b2, d6, b9*398, c1, e3*353+b9
  FROM t1, t9, t6, t3, t2
 WHERE d6 in (885,924,457,578,786,664)
   AND 488=d2
   AND a3=b9
   AND c9=688
   AND a1=d9;"

 b2 | d6  | ?column? | c1  | ?column? 
----+-----+----------+-----+----------
 93 | 924 |     1990 | 585 |   286994
 93 | 457 |     1990 | 585 |   286994
 93 | 786 |     1990 | 585 |   286994
 93 | 578 |     1990 | 585 |   286994
 93 | 664 |     1990 | 585 |   286994
 93 | 885 |     1990 | 585 |   286994
(6 rows)

real	0m0.024s
user	0m0.010s
sys	0m0.006s
```

The attached [foo.tgz](https://github.com/user-attachments/files/21497404/foo.tgz) contains Parquet files of the dumped contents of the same tables referenced in the SQL query. If I kick off `super` running the same query against that data, my Intel Macbook shows the process consuming 100%+ CPU continuously and runs for tens of minutes without ever producing a result.

```
$ super -version
Version: 230007cb7

$ super -c "
SELECT b2, d6, b9*398, c1, e3*353+b9
  FROM t1, t9, t6, t3, t2
 WHERE d6 in (885,924,457,578,786,664)
   AND 488=d2
   AND a3=b9
   AND c9=688
   AND a1=d9;"

[runs for tens of minutes without producing a result]
```

As @nwt confirmed, with the initial cross join support added in #6064, no optimizations yet exist to reduce the runtime in the same way that Postgres seems to be doing. Therefore the high runtime we're observing follows directly from the large cartesian product, i.e.:

```
$ for file in t*; do   super -c "count()" $file; done | super -f text -
128
113
129
92
98

$ expr 128 \* 113 \* 129 \* 92 \* 98
16822557696
```

Cranking through 16-million combinations would indeed take a while.

mattnibs:
This commit adds functionality to the optimizer to pull up simple join predicates. Currently this will only pullup simple predicate expressions where keys are compared against constant values.

Partially fixes #6074

### DIFF
--- a/compiler/optimizer/optimizer.go
+++ b/compiler/optimizer/optimizer.go
@@ -144,6 +144,7 @@ func (o *Optimizer) Optimize(main *dag.Main) error {
 	seq = mergeFilters(seq)
 	seq = mergeValuesOps(seq)
 	inlineRecordExprSpreads(seq)
+	seq = joinFilterPullup(seq)
 	seq = removePassOps(seq)
 	seq = replaceSortAndHeadOrTailWithTop(seq)
 	o.optimizeParallels(seq)
@@ -504,6 +505,103 @@ func inlineRecordExprSpreads(v any) {
 		}
 		return r
 	})
+}
+
+func joinFilterPullup(seq dag.Seq) dag.Seq {
+	seq = mergeFilters(seq)
+	for i := 0; i <= len(seq)-3; i++ {
+		fork, isfork := seq[i].(*dag.Fork)
+		leftAlias, rightAlias, isjoin := isJoin(seq[i+1])
+		filter, isfilter := seq[i+2].(*dag.Filter)
+		if !isfork || !isjoin || !isfilter {
+			continue
+		}
+		if len(fork.Paths) != 2 {
+			panic(seq[i])
+		}
+		var remaining []dag.Expr
+		for _, e := range splitPredicate(filter.Expr) {
+			if pullup, ok := pullupExpr(leftAlias, e); ok {
+				fork.Paths[0] = append(fork.Paths[0], dag.NewFilter(pullup))
+				continue
+			}
+			if pullup, ok := pullupExpr(rightAlias, e); ok {
+				fork.Paths[1] = append(fork.Paths[1], dag.NewFilter(pullup))
+				continue
+			}
+			remaining = append(remaining, e)
+		}
+		if len(remaining) == 0 {
+			// Filter has been fully pulled up and can be removed.
+			seq.Delete(i+2, i+3)
+		} else {
+			out := remaining[0]
+			for _, e := range remaining[1:] {
+				out = dag.NewBinaryExpr("and", e, out)
+			}
+			seq[i+2] = dag.NewFilter(out)
+		}
+		fork.Paths[0] = joinFilterPullup(fork.Paths[0])
+		fork.Paths[1] = joinFilterPullup(fork.Paths[1])
+	}
+	return seq
+}
+
+func isJoin(op dag.Op) (string, string, bool) {
+	switch op := op.(type) {
+	case *dag.HashJoin:
+		return op.LeftAlias, op.RightAlias, true
+	case *dag.Join:
+		return op.LeftAlias, op.RightAlias, true
+	default:
+		return "", "", false
+	}
+}
+
+func splitPredicate(e dag.Expr) []dag.Expr {
+	if b, ok := e.(*dag.BinaryExpr); ok && b.Op == "and" {
+		return append(splitPredicate(b.LHS), splitPredicate(b.RHS)...)
+	}
+	return []dag.Expr{e}
+}
+
+func pullupExpr(alias string, expr dag.Expr) (dag.Expr, bool) {
+	e, ok := expr.(*dag.BinaryExpr)
+	if !ok {
+		return nil, false
+	}
+	if e.Op == "and" {
+		lhs, lok := pullupExpr(alias, e.LHS)
+		rhs, rok := pullupExpr(alias, e.RHS)
+		if !lok || !rok {
+			return nil, false
+		}
+		return dag.NewBinaryExpr("and", lhs, rhs), true
+	}
+	if e.Op == "or" {
+		lhs, lok := pullupExpr(alias, e.LHS)
+		rhs, rok := pullupExpr(alias, e.RHS)
+		if !lok || !rok {
+			return nil, false
+		}
+		return dag.NewBinaryExpr("or", lhs, rhs), true
+
+	}
+	var literal *dag.Literal
+	var this *dag.This
+	for _, e := range []dag.Expr{e.RHS, e.LHS} {
+		if l, ok := e.(*dag.Literal); ok && literal == nil {
+			literal = l
+			continue
+		}
+		if t, ok := e.(*dag.This); ok && this == nil && len(t.Path) > 1 && t.Path[0] == alias {
+			this = t
+			continue
+		}
+		return nil, false
+	}
+	path := slices.Clone(this.Path[1:])
+	return dag.NewBinaryExpr(e.Op, dag.NewThis(path), literal), true
 }
 
 func liftFilterOps(seq dag.Seq) dag.Seq {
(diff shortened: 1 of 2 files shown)


compiler/optimizer/optimizer.go
```diff
@@ -506,6 +507,108 @@ func inlineRecordExprSpreads(v any) {
 	})
 }
 
+func joinFilterPullup(seq dag.Seq) dag.Seq {
+	seq = mergeFilters(seq)
+	if len(seq) < 3 {
+		return seq
+	}
```
nwt: Nit: This duplicates the loop condition.

```suggestion
```

compiler/optimizer/optimizer.go
```diff
@@ -506,6 +507,108 @@ func inlineRecordExprSpreads(v any) {
...
+		if !isfork || !isjoin || !isfilter {
+			continue
+		}
+		if len(fork.Paths) != 2 {
+			panic(seq[i])
+		}
+		var remaining []dag.Expr
+		for _, e := range breakupFilter(filter.Expr) {
+			pullup, ok := pullupExpr(leftAlias, e)
+			if ok {
```
nwt: Nit:

```suggestion
			if pullup, ok := pullupExpr(leftAlias, e); ok {
```

compiler/optimizer/optimizer.go
```diff
@@ -506,6 +507,108 @@ func inlineRecordExprSpreads(v any) {
...
+				fork.Paths[0] = append(fork.Paths[0], dag.NewFilter(pullup))
+				continue
+			}
+			pullup, ok = pullupExpr(rightAlias, e)
+			if ok {
```
nwt: Nit:

```suggestion
			if pullup, ok := pullupExpr(rightAlias, e); ok {
```

compiler/optimizer/optimizer.go
```diff
@@ -506,6 +507,108 @@ func inlineRecordExprSpreads(v any) {
...
+			seq[i+2] = dag.NewFilter(out)
+		}
+		fork.Paths[0] = joinFilterPullup(fork.Paths[0])
+		fork.Paths[1] = joinFilterPullup(fork.Paths[1])
+	}
+	return seq
+}
+
+func isJoin(op dag.Op) (string, string, bool) {
+	switch join := op.(type) {
```
nwt: Nit: We usually shadow this variable in a type switch.

```suggestion
	switch op := op.(type) {
```

compiler/optimizer/optimizer.go
```diff
@@ -506,6 +507,108 @@ func inlineRecordExprSpreads(v any) {
...
+	case *dag.HashJoin:
+		return join.LeftAlias, join.RightAlias, true
+	case *dag.Join:
+		return join.LeftAlias, join.RightAlias, true
+	default:
+		return "", "", false
+	}
+}
+
+func breakupFilter(e dag.Expr) []dag.Expr {
```
nwt: Nit: I think a name like splitConjunction would make the behavior of this clearer.

compiler/optimizer/optimizer.go
```diff
@@ -506,6 +507,108 @@ func inlineRecordExprSpreads(v any) {
...
+				continue
+			}
+			if t, ok := e.(*dag.This); ok && this == nil && len(t.Path) > 1 && t.Path[0] == alias {
+				this = t
+				continue
+			}
+			return nil, false
+		}
+		path := slices.Clone(this.Path[1:])
+		return dag.NewBinaryExpr(e.Op, &dag.This{Kind: "This", Path: path}, dag.CopyExpr(literal)), true
```
nwt: Nits: Use dag.NewThis. And I don't think there's any need to copy the dag.Literal.

```suggestion
		return dag.NewBinaryExpr(e.Op, dag.NewThis(path), literal), true
```

compiler/optimizer/optimizer.go
```diff
@@ -506,6 +507,108 @@ func inlineRecordExprSpreads(v any) {
...
+	if b, ok := e.(*dag.BinaryExpr); ok && b.Op == "and" {
+		return append(breakupFilter(b.LHS), breakupFilter(b.RHS)...)
+	}
+	return []dag.Expr{e}
+}
+
+func pullupExpr(alias string, e dag.Expr) (dag.Expr, bool) {
+	switch e := e.(type) {
+	case *dag.BinaryExpr:
+		if e.Op == "and" {
```
nwt: Nit: Indent less.

```suggestion
func pullupExpr(alias string, expr dag.Expr) (dag.Expr, bool) {
	e, ok := e.(*dag.BinaryExpr)
	if !ok {
		return nil, false
	}
	if e.Op == "and" {
```

philrz:
It's a functional 👍 for me! While the sqllogictests contain lots of heavy cross joins it didn't seem like I could easily spot any that had only predicate expressions with constant comparisons, so I modified one. There's probably no need to make this into a ztest in the super repo since we'll have thousands of similar tests that are even more complex once the full feature work is complete, but FWIW with the attached test data [data.tgz](https://github.com/user-attachments/files/22644907/data.tgz), this query runs fast and produces the same result as the equivalent in Postgres:

```
$ super -version
Version: 0471667fa

$ time super -s -c "
SELECT x7,x3,x62,x57,x47,x25,x51,x55,x44,x49,x53,x32
  FROM t47 (FORMAT parquet), t55 (FORMAT parquet), t62 (FORMAT parquet), t25 (FORMAT parquet), t32 (FORMAT parquet), t7 (FORMAT parquet), t3 (FORMAT parquet), t57 (FORMAT parquet), t53 (FORMAT parquet), t51 (FORMAT parquet), t44 (FORMAT parquet), t49 (FORMAT parquet)
WHERE b32=1
  AND b51=1
  AND b7=1
  AND a49=1
  AND b53=1
  AND b25=7
  AND b57=1
  AND a44=1
  AND a62=1
  AND a49=1
  AND a7=1
  AND a25=1
  AND a3=1
  AND a47=1
  AND b55=1;"

{x7:"table t7 row 1",x3:"table t3 row 1",x62:"table t62 row 1",x57:"table t57 row 9",x47:"table t47 row 1",x25:"table t25 row 1",x51:"table t51 row 10",x55:"table t55 row 1",x44:"table t44 row 1",x49:"table t49 row 1",x53:"table t53 row 5",x32:"table t32 row 6"}

real	0m0.094s
user	0m0.107s
sys	0m0.068s
```

...whereas at current tip of `main` it effectively runs forever due to the massive cartesian product.

mattnibs:
thanks @philrz good to hear

compiler/optimizer/optimizer.go
mattnibs: This tru

compiler/optimizer/optimizer.go
mattnibs: I like split but Conjunction to me doesn't necessarily imply filter or predicate. Maybe splitPredicate?

nwt: My beef with splitPredicate is that it doesn't really say anything about how the function parameter will be split. splitConjunction at least suggests that if the parameter is a logical conjunction then it will be split into its operands.

PR #6276, opened 2025-10-02
Distrbute prerelease super via Homebrew Cask

philrz:
## What's Changing

This builds on #6271 by creating artifacts as each PR merges and updating a Homebrew Cask to make those artifacts available on a public S3 bucket.

## Why

1. As described in #6271, GoReleaser is pushing us toward using Casks rather than Formulas, so this is is a step toward where we'll want to be at GA
2. I'd been updating the prerelease Formula manually on a "best effort" basis, but the approach here is automatic each time a PR merges to `main`, which should get us better user testing in this prerelease phase

## Details

Consensus has been that we don't want any artifacts on the [super Releases](https://github.com/brimdata/super/releases) page on GitHub until we're ready to tag a GA release. The approach here gets around that putting the artifacts on a public S3 bucket instead and having the Homebrew Cask point at those URLs.

Part of the goal here is to get prerelease users off the Formula and onto the Cask before GA. Based on my research there's no way to silently _force_ that transition, so when this merges, my next planned step will be to replace the old [`super.rb` Homebrew Formula](https://github.com/brimdata/homebrew-tap/blob/main/super.rb) with one that installs a shim that will advise users of the minimal steps to switch over to the Cask. At that point they'll start getting updated `super` binaries as part of every general `brew update` / `brew upgrade`. I'd plan to remove that `super.db` Formula entirely right before we tag the GA release.

I've tested this extensively in a [personal super fork](https://github.com/philrz/super) and [personal Homebrew tap](https://github.com/brimdata/homebrew-tap) so if you want to see proof it works run `brew install philrz/tap/super` on a scratch macOS system with Homebrew installed.

### DIFF
--- a/.github/workflows/homebrew-update.yaml
+++ b/.github/workflows/homebrew-update.yaml
@@ -1,0 +1,60 @@
+name: "Update Homewbrew Cask"
+
+on:
+  pull_request:
+    branches:
+      - main
+    types: closed
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: 'Commit SHA to build'
+        required: true
+        default: 'main'
+        type: string
+
+jobs:
+  build-and-update:
+    if: github.event.pull_request.merged == true || github.event_name == 'workflow_dispatch'
+    runs-on: ubuntu-22.04
+    steps:
+      - id: checkout-ref
+        run: |
+          if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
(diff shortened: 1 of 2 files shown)


philrz:
In a group discussion there was consensus that this should be a nightly job and should run in the Homebrew Tap repo. I'll open a new PR over there that takes that approach. Closing this one.

PR #6278, opened 2025-10-03
add new cast implementation for complex types

### LINKED ISSUE #2894: shaping functions cannot cast map types
As of #2885, the shaping functions cannot cast map types because `castType` in `expr/shaper.go` does not handle them.

Here's a simple repro as of Zed commit f967c99.

```
$ zq -version
Version: v1.8.1-50-gf967c99d

$ echo '|{"foo": "1", "bar": "2"}|' | zq -z 'type MyMap = |{string:int64}| yield cast(this, MyMap)' -
error("cannot yet use maps in shaping functions (issue #2894)")
```

When this addressed the expected output would be `|{"bar":2,"foo":1}|`.

### LINKED ISSUE #4315: cast() should return an error/warning on failed casts to complex types
#4314 was a recent community issue where the user's intent was to cast to the primitive `ip` type but they accidentally wrote their call as `cast(host,<[ip]>))` such that it was actually attempting to cast to an array-of-`ip`. Their call to `cast()` was effectively behaving like a no-op and just returning the same `string`-type value.

A close reading of the [docs](https://zed.brimdata.io/docs/language/functions/cast) shows this is actually expected behavior at the moment.

> If a cast fails, an error is returned when casting to primitive types and the input value is returned when casting to complex types.

However, when reviewing with the team, no one felt that this behavior was _preferred_ and is therefore likely just a detail of the original `cast()` implementation which pre-dated much of the improved error handling/reporting now present in Zed. Therefore we'll look at starting to return some kind of error/warning, e.g., "attempt to cast [value] to [complex type] failed".

### LINKED ISSUE #5168: casting complex type is order dependent when docs say different

nwt:
Casting of complex types is implemented in runtime/sam/expr/shaper.go.
Replace it with a new implementation in
runtime/sam/expr/function/cast.go with these differences:

* Always generates a structured error when casting fails 

* Preserves target type field order when casting records

* Able to cast maps

* More maintainable

Fixes #2894
Fixes #4315
Closes #5168

### DIFF
--- a/runtime/sam/expr/function/cast.go
+++ b/runtime/sam/expr/function/cast.go
@@ -33,7 +33,7 @@ func (c *cast) Call(args []super.Value) super.Value {
 		}
 		return c.cast(from, typ)
 	}
-	return c.sctx.WrapError("cannot cast to "+sup.FormatValue(to), from)
+	return c.sctx.WrapError("cast target must be a type or type name", to)
 }
 
 func (c *cast) cast(from super.Value, to super.Type) super.Value {
--- a/runtime/sam/expr/function/cast.go
+++ b/runtime/sam/expr/function/cast.go
@@ -1,0 +1,225 @@
+package function
+
+import (
+	"maps"
+	"slices"
+
+	"github.com/brimdata/super"
+	"github.com/brimdata/super/runtime/sam/expr"
+	"github.com/brimdata/super/scode"
+	"github.com/brimdata/super/sup"
+)
+
+type cast struct {
+	sctx *super.Context
+}
+
+func (c *cast) Call(args []super.Value) super.Value {
+	from, to := args[0], args[1]
+	if from.IsError() {
+		return from
+	}
+	switch toUnder := to.Under(); toUnder.Type().ID() {
+	case super.IDString:
+		typ, err := c.sctx.LookupTypeNamed(toUnder.AsString(), super.TypeUnder(from.Type()))
+		if err != nil {
+			return c.sctx.WrapError("cannot cast to named type: "+err.Error(), from)
+		}
+		return super.NewValue(typ, from.Bytes())
+	case super.IDType:
+		typ, err := c.sctx.LookupByValue(toUnder.Bytes())
+		if err != nil {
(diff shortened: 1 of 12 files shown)


mattnibs (approved):

runtime/sam/expr/function/cast.go
```diff
@@ -0,0 +1,225 @@
...
+		}
+		return super.NewValue(typ, from.Bytes())
+	case super.IDType:
+		typ, err := c.sctx.LookupByValue(toUnder.Bytes())
+		if err != nil {
+			panic(err)
+		}
+		return c.cast(from, typ)
+	}
+	return c.sctx.WrapError("cannot cast to "+sup.FormatValue(to), from)
```
mccanne: Maybe an error here about the second argument when the second arg isn't a type or string, e.g., instead of 
```
; super -c "cast(1,2)"
error({message:"cannot cast to 2",on:1})
```
maybe 
```
error({message:"cast target must be a type or name of type",on:2})
```

mccanne (approved):

PR #6279, opened 2025-10-05
unify naming in package dag

mccanne:
This commit cleans up the naming conventions in package dag, remove some unused code, and reorders some code to conform with alphabetical order of type names.  All dag.Ops now have a suffix of "Op" or "Scan" and all dag.Exprs have a suffix of "Expr".

This is strictly a naming/style change and no logic is changed or introduced here.


nwt (approved):
This is the way.

Since I merged #6278 there's a logical conflict between this branch and main (it has a new reference to dag.Call that needs to become dag.CallExpr), so be sure to resolve that before you merge.

PR #6284, opened 2025-10-06
add type checking

mccanne:
This commit adds type checking to the semantic analyzer thus bringing static typing to dynamic data.  As far as we know, this has not been done before in a general fashion in any SQL-like query language for dynamically typed data (e.g., SQL++, Asterix, search languages, etc).

This works by computing fused types of each operator's output and propagating these types in a dataflow analysis.  When types are unknown, the analysis flexibly models them as having any possible type.  The CSUP and BSUP formats for dynamically typed data will be updated in future PRs to include fused-type information so robust type checking can be carried out for any super-structured data.

Type checking for built-in functions and aggregate functions is not yet done as we need support from the functions packages to provide type signatures.  This will be done in a subsequent PR.

Many existing tests were updated since they had problematic type behavior. A number of new tests were added to test the type checker but coverage is light.


compiler/semantic/ztests/checker-func.yaml
```diff
@@ -0,0 +1,8 @@
+spq: |
+  fn z(a):a[0]
+  values z(1)
+
+error: |
+  indexed entity is not indexable at line 1, column 9:
+  fn z(a):a[0]
+          ~
```
nwt, line 8: Nit: This error will be confusing in a program with multiple calls to z. Is there any way to get include the call site as context?

compiler/semantic/ztests/checker-parquet.yaml
```diff
@@ -0,0 +1,10 @@
+script: |
+  super -f parquet -o t.parquet -c "values {x:1,y:2}"
+  ! super -c "from t.parquet | values z"
+
+outputs:
+  - name: stderr
+    data: |
+      "z" no such field at line 1, column 25:
```
nwt: Nit: Error might read better like this.

```suggestion
      no such field "z" at line 1, column 25:
```

philrz:
The sqllogictests spotted something that broke in this branch. Here's the baseline working as expected on current tip of `main`:

```
$ super -version
Version: 525853d26

$ super -f parquet -o data.parquet -c "values {col1:1,col2:1},{col1:2,col2:3},{col1:4,col2:4}" &&
  super -c "SELECT * FROM data.parquet WHERE col1 IN (col2);"

{col1:1,col2:1}
{col1:4,col2:4}
```

Here's Postgres handling it the same.

```
$ psql postgres
psql (17.6 (Homebrew))
Type "help" for help.

postgres=# CREATE TABLE DATA (col1 INTEGER, col2 INTEGER);
CREATE TABLE
postgres=# INSERT INTO DATA (col1, col2) VALUES (1,1),(2,3),(4,4);
INSERT 0 3

postgres=# SELECT * FROM data WHERE col1 IN (col2);
 col1 | col2 
------+------
    1 |    1
    4 |    4
(2 rows)
```

And here it is failing on the branch:

```
$ super -f parquet -o data.parquet -c "values {col1:1,col2:1},{col1:2,col2:3},{col1:4,col2:4}" &&
  super -c "SELECT * FROM data.parquet WHERE col1 IN (col2);"

bad type for right-hand side of in operator: int64 at line 1, column 43:
SELECT * FROM data.parquet WHERE col1 IN (col2);
                                          ~~~~
```

mccanne:
@philrz this behavior was previously present but exposed by the type-checking system.  Currently, the RHS of `... IN ( expr )` doesn't compile into a tuple/array but the runtime `IN` operator returns true for `scalar IN scalar`.  I think this is questionable and would should revisit these semantics.  So, we need to fix the parser to recognize `... IN ( expr )` as a tuple and update the runtime to produce dynamic errors when testing if something is `IN` a scalar value.

I would recommend merging this PR then fixing the existing problems in a subsequent PR.

compiler/semantic/ztests/checker-plus-ip.yaml
```diff
@@ -0,0 +1,6 @@
+spq: values 10.1.1.1 + 1
+
+error: |
+  type mismatch at line 1, column 8:
+  values 10.1.1.1 + 1
+         ~~~~~~~~
```
nwt: Nit: Maybe underline the entire offending expression instead of just one operand?

compiler/semantic/ztests/no-such-builtin.yaml
```diff
@@ -1,12 +1,8 @@
 spq: |
   fn foo(f,x):f(x)
-  values foo(nosuchfunction, 1)
-  | values foo(&nosuchfunction, 2)
+  values foo(&nosuchfunction, 2)
 
 error: |
-  function called via parameter "f" is bound to a non-function at line 1, column 13:
-  fn foo(f,x):f(x)
-              ~~~~
```
nwt, line 9: This error is covered by this test only. If you can't maintain it here, maybe add a new test that covers it?

runtime/ztests/op/join-error.yaml
```diff
@@ -11,4 +11,6 @@ inputs:
 outputs:
   - name: stderr
     data: |
-      join requires two upstream parallel query paths
+      join requires two query inputs at line 1, column 1:
```
nwt: Nit: "query" feels unnecessary here.

```suggestion
      join requires two inputs at line 1, column 1:
```

compiler/semantic/checker.go
```diff
@@ -0,0 +1,1081 @@
...
+func (c *checker) epop() errlist {
+	n := len(c.estack) - 1
+	errs := c.estack[n]
+	c.estack = c.estack[:n]
+	return errs
+}
+
+func (c *checker) ekeep(errs errlist) {
+	n := len(c.estack) - 1
+	c.estack[n] = append(c.estack[n], errs...)
```
nwt: Nit: The "e" prefix on these isn't very suggestive. How about "err" instead?

compiler/semantic/checker.go
```diff
@@ -0,0 +1,1081 @@
...
+			return super.TypeBool
+		default:
+			panic(e.Op)
+		}
+	default:
+		panic(e)
+	}
+}
+
+func (c *checker) isContainer(containerType super.Type) super.Type {
```
nwt: Nit: It's more idiomatic to add a bool as a second return value to signal that a predicate has evaluated to false instead of signaling false by returning nil.

compiler/semantic/checker.go
```diff
@@ -0,0 +1,1081 @@
...
+	return false
+}
+
+func hasUnknown(typ super.Type) bool {
+	if u, ok := super.TypeUnder(typ).(*super.TypeUnion); ok {
+		for _, t := range u.Types {
+			if hasUnknown(t) {
+				return true
+			}
+		}
```
nwt: Nits:

```suggestion
		if slices.ContainsFunc(u.Types, hasUnknown) {
			return true
		}
```

compiler/semantic/checker.go
```diff
@@ -0,0 +1,1081 @@
...
+	case *super.TypeError:
+		return isUnknown(typ)
+	case *super.TypeOfString, *super.TypeOfNull:
+		return true
+	case *super.TypeUnion:
+		for _, t := range typ.Types {
+			if hasString(t) {
+				return true
+			}
+		}
```

```suggestion
		return slices.ContainsFunc(typ.Types, hasString(t))
```

compiler/semantic/checker.go
```diff
@@ -0,0 +1,1081 @@
...
+	return types
+}
+
+func (c *checker) join(types []super.Type, op *sem.JoinOp) super.Type {
+	if len(types) != 2 {
+		c.error(op, errors.New("join requires two query inputs"))
+	}
+	typ := c.t.sctx.MustLookupTypeRecord([]super.Field{
+		{Name: op.LeftAlias, Type: types[0]},
+		{Name: op.RightAlias, Type: types[1]},
```

```suggestion
		super.NewField(op.LeftAlias, types[0]),
		super.NewField(op.RightAlias, types[1]),
```

compiler/semantic/checker.go
```diff
@@ -0,0 +1,1081 @@
...
+	for _, path := range paths {
+		fuser.fuse(c.pathToRec(path.typ, path.elems))
+	}
+	return fuser.Type(c)
+}
+
+func (c *checker) pathToRec(typ super.Type, elems []string) super.Type {
+	for ; len(elems) > 0; elems = elems[:len(elems)-1] {
+		last := elems[len(elems)-1]
+		typ = c.t.sctx.MustLookupTypeRecord([]super.Field{{Name: last, Type: typ}})
```
nwt:
```suggestion
	```suggestion
	for _, elem := range slices.Backward(elems) {
typ}})
		typ = c.t.sctx.MustLookupTypeRecord([]super.Field{{Name: elem, Type: typ}})
```

compiler/semantic/checker.go
```diff
@@ -0,0 +1,1081 @@
...
+	id := super.TypeUnder(typ).ID()
+	if super.IsNumber(id) || id == super.IDNull {
+		return true
+	}
+	if u, ok := super.TypeUnder(typ).(*super.TypeUnion); ok {
+		for _, t := range u.Types {
+			if hasNumber(t) {
+				return true
+			}
+		}
```

```suggestion
		if slices.ContainsFunc(u.Types, hasNumber) {
			return true
		}
```

compiler/semantic/ztests/checker-func.yaml
mccanne: I agree but we don't have a way to report errors tying together different locations in the code.  We need to add this.  How about we fix this in a subsequent PR?

nwt: Just a nit so later is fine.

mccanne:
Ok, regarding the SQL problem with the IN operator, we decided on zoom today that IN semantics will include scalar equality in addition to containment equality.  This means we don't need to change the runtime.  Instead, I just pushed changes to update the type checker.  Docs for the IN operator will come soon to reflect this.

philrz:
FYI, once this merged to `main` (commit c7dcd26) I re-ran the 1+ million sqllogictests that had previously been successful and 100% of them passed. So, `IN` fix is confirmed a success and no other surprises were hiding behind that one. 👍

PR #6286, opened 2025-10-07
remove crop(), fill(), fit(), order(), and shape()

nwt:
cast() is generally sufficient though we may add a variant that doesn't add missing fields with a null value when casting records.

runtime/sam/expr.ConstShaper has to stick around because it's used in runtime/sam/op/fuse.  We can remove it in favor of runtime/sam/expr/function.cast after we enhance that latter to handle union-typed record fields in the target type.


nwt:
@mattnibs: I don't think there are any right now. I get no output from this:

```sh
git grep '(crop|fill|fit|order|shape)\(' -- book/src
```

PR #6292, opened 2025-10-10
add table references and "this" to SQL expressions

### LINKED ISSUE #6241: What happens when a user tries to use "this" in relational scoping
## tl;dr

Users that are familiar with SuperSQL pipe context may instinctively try to use `this` in relational context, but it's not feasible to provide this for the general case. Some combination of helpful error messages, documentation, and additional language elements may be necessary to help users get what they want through a different approach.

## Details

This came up recently in a user question in a [community Slack thread](https://super-db.slack.com/archives/CTSMAK6G7/p1757471218849919).

We'll start from this query that seemed to work ok back at super commit d9c1886.

```
$ super -version
Version: d9c1886cc

$ super -s -c "
  type foo = {id:int64}

  values [{id:2},{id:3},{id:'a'}]
  | unnest this
  | shape(this, <foo>)
  | select *
    where is(<foo>)"

{id:2}
{id:3}
```

Then at the next commit 5075037 after the merge of #6161, the query no longer worked as written.

```
$ super -version
Version: 5075037c8

$ super -s -c "
  type foo = {id:int64}

  values [{id:2},{id:3},{id:'a'}]
  | unnest this
  | shape(this, <foo>)
  | select *
    where is(<foo>)"

too few arguments at line 8, column 11:
    where is(<foo>)
          ~~~~~~~~~
```

The user noticed and understood the changes in the PR, so they attempted to get their query working again by adding an explicit `this` in the call to the `is` function, but:

```
$ super -s -c "
  type foo = {id:int64}

  values [{id:2},{id:3},{id:'a'}]
  | unnest this
  | shape(this, <foo>)
  | select *
    where is(this, <foo>)"

cannot reference 'this' in relational context; consider the 'values' operator at line 8, column 14:
    where is(this, <foo>)
             ~~~~
```

This guidance on its own was not particularly helpful, as it was not enough for the user to see how to adjust the query to get the result they had before. Indeed, one way to get there is to avoid SQL altogether and rewrite that portion with just pipes, which doesn't require explicit use of `values` at that part of the query.

```
$ super -s -c "
  type foo = {id:int64}

  values [{id:2},{id:3},{id:'a'}]
  | unnest this
  | shape(this, <foo>)
  | where is(this, <foo>)"

{id:2}::=foo
{id:3}::=foo
```

@mccanne joined the thread and gave the following explanation, including a take on getting the query to work while still using `select`:

> This is a challenge we're facing in making the mix of relational and super-structured understandable.   Once you go from pipe operators to SQL operators, you go from dataflow scoping to relational scoping.  In the relational model, there's not really a way to refer to "this" because there is no notion of `this`... instead you refer to columns and how column names bind to different parts of the dataflow is based on the ad hoc scoping rules of SQL (e.g., the `WHERE` clause can refer to aggregate output, the input relation, or the `SELECT` columns based on statically defined scoping).  We've thought about adding a `this()` function that could be called from SQL which would give you the input to the SQL operator but we haven't done this yet.  That all said, you can make your query above work by wrapping `this` in a single column record then operate on the "SQL input column" as in:
>```
>$ super -s -c "
>  type foo = {id:int64}
>
>  values [{id:2},{id:3},{id:'a'}]
>  | unnest this
>  | shape(this, <foo>)
>  | {that:this}
>  | select that
>    where is(that, <foo>)
>"
>
>{that:{id:2}::=foo}
>{that:{id:3}::=foo}
>```

I also had a [thread](https://brim-labs.slack.com/archives/CP3BMAEEP/p1757966340757879) with @mccanne after my own experience bumping into this with a previously-working attempt to use the `grep` function on `this` in a SQL `WHERE` clause.

mccanne:
This commit adds support for referring to tables in a SQL expression, resulting in a record representing the table row (as in duckdb and somewhat similarly in postgres).  We also added support for referencing "this" in a SQL expression, which refers to the input relation in SELECT and WHERE expressions, the output relation in HAVING clauses, and the input relation for arguments (and where clauses) of agg functions in HAVING clauses.

This new logic causes an error for table references of dynamic schemas. This is to avoid a situation where "select T from T" refers to the table in a dynamic schema, then when a schema shows up for that same data, the query compiles differently to a field reference of T inside T (following postgres and duckdb scope precedence).  When it is desied to refer to the table of a dynamic source, the special value "this" can be used instead. In general, query semantics should be identical when types/schemas are known and unknown; if this isn't the case anywhere here, then it's a design bug.

We also added an escape valve for referring to a SQL column named "this", which is simply denoted with double quotes.

Finally, these changes exposed a problem in the as-name inference algorithm, where the internal DAG paths would show up, so we updated the inference code to strictly use the AST instead of a mix of the AST and sem tree.

Fixes #6241

### DIFF
--- a/compiler/semantic/expr.go
+++ b/compiler/semantic/expr.go
@@ -1001,14 +1001,10 @@ func (t *translator) semAggFunc(n ast.Node, name string, arg ast.Expr, where ast
 			t.scope.schema = save
 		}()
 	}
-	var e sem.Expr
-	if arg != nil {
-		e = t.semExpr(arg)
-	}
 	return &sem.AggFunc{
 		Node:     n,
 		Name:     name,
-		Expr:     e,
+		Expr:     t.semExprNullable(arg),
 		Where:    t.semExprNullable(where),
 		Distinct: distinct,
 	}


compiler/semantic/expr.go
```diff
@@ -947,14 +983,34 @@ func (t *translator) maybeConvertAgg(call *ast.CallExpr) sem.Expr {
...
+		e = t.semExpr(arg)
 	}
 	return &sem.AggFunc{
-		Node:  call,
-		Name:  nameLower,
-		Expr:  e,
-		Where: t.semExprNullable(call.Where),
+		Node:     n,
+		Name:     name,
+		Expr:     e,
```

```suggestion
		Expr:     t.semExprNullable(arg),
```

philrz:
Here's a failure that happens on the branch that runs ok on tip of `main`. Unpack the attached [data.tgz](https://github.com/user-attachments/files/22855622/data.tgz), then run:

```
$ super -version
Version: 321566076

$ super -c "SELECT tab1.col1 FROM tab0 (FORMAT parquet), tab2 (FORMAT parquet) AS cor0 CROSS JOIN tab1, tab2 AS cor1;"
"tab1": ambiguous column reference at line 1, column 8:
SELECT tab1.col1 FROM tab0 (FORMAT parquet), tab2 (FORMAT parquet) AS cor0 CROSS JOIN tab1, tab2 AS cor1;
       ~~~~~~~~~
```

philrz:
Here's one that runs on the branch but produces only error outputs. It uses the same test data as in https://github.com/brimdata/super/pull/6292#issuecomment-3391616800.

```
$ super -version
Version: 321566076

$ super -c "SELECT tab2.col1 AS col2 FROM tab1 (FORMAT parquet), tab1 (FORMAT parquet) AS cor0 CROSS JOIN tab2;"
{col2:error("missing")}
{col2:error("missing")}
{col2:error("missing")}
{col2:error("missing")}
...
```

Whereas on tip of `main`:

```
$ super -c "SELECT tab2.col1 AS col2 FROM tab1 (FORMAT parquet), tab1 (FORMAT parquet) AS cor0 CROSS JOIN tab2;"
{col2:31}
{col2:31}
{col2:31}
{col2:17}
...
```

philrz:
@mccanne and @nwt spotted that the failures in my two comments above could both be explained by the absence of the `(FORMAT parquet)` on some of the table references that resulted in getting dynamic schemas rather than static. The fact they'd been working correctly on `main` looks like it was a happy accident. Therefore I'm updating my sqllogictest scripts to handle those additional `(FORMAT parquet)` corner cases and can then retest in full to sniff out if there's any failures hiding behind those with other root causes. In the meantime, no reason to hold off on getting this PR merged. 👍

philrz:
Following up on the prior comment, after adding the additional `(FORMAT parquet)` where needed (https://github.com/brimdata/sqllogic-ztests/commit/d836ae9cbf7f693d0299d261ca59cabd053d17ed) a run of the 1+ million sqllogictest queries that had previously run successfully at super commit 0dd0e87 before this PR merged also ran successfully at commit cad1342 after this PR merged. 👍

PR #6295, opened 2025-10-13
integrate resolver's lambda unraveling into semantic translator pass

mccanne:
This commit integrates the lambda unraveling technique used in the resolver into the first-pass of the semantic translator.  This is possible by indexing the lambda permutation using decl IDs instead of call tags based on the intuition that the only thing that can give rise to variation in unraveled combinations is the declared function, i.e., at the outermost call site, each variation of lamba-parameterized function depends only on the function passed in, which are uniquely defined by their ID not their unraveled instances.

This means the evaluator can simply traverse an expression to determine if it's constant and then dagen it.  There is no need anymore to re-enter the resolver.

This is also the key step needed to integrate type checking into the translator, which will then allow us to utilize dataflow-analyzed fused types from pipe queries in schema-column resolution for SQL queries that include pipe subqueries.  This will be forthcoming in a subsequent PR.

### DIFF
--- a/compiler/semantic/expr.go
+++ b/compiler/semantic/expr.go
@@ -98,7 +98,7 @@ func (t *translator) semExpr(e ast.Expr) sem.Expr {
 		// an arg to another function.  It could be a built-in (as in &upper),
 		// or a user function (as in fn foo():... &foo)...
 		id := e.Name
-		if boundID, _ := t.scope.lookupFuncDeclOrParam(e.Name); boundID != "" {
+		if boundID, _ := t.scope.lookupFuncDeclOrParam(id); boundID != "" {
 			id = boundID
 		}
 		return &sem.FuncRef{
@@ -785,7 +785,7 @@ func (t *translator) semMapCall(call *ast.CallExpr, args []sem.Expr) sem.Expr {
 		return badExpr()
 	}
 	mapArgs := []sem.Expr{sem.NewThis(call.Args[1], nil)}
-	e := t.resolveCall(call.Args[1], ref.ID, mapArgs)
+	e := t.resolver.resolveCall(call.Args[1], ref.ID, mapArgs)
 	if callExpr, ok := e.(*sem.CallExpr); ok {
 		return &sem.MapCallExpr{
 			Node:   call,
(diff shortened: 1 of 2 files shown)


compiler/semantic/expr.go
```diff
@@ -95,15 +95,15 @@ func (t *translator) semExpr(e ast.Expr) sem.Expr {
...
 		// We get here for &refs that are in a call expression. e.g.,
-		// an arg to another function.  These are only built-ins as
-		// user functions should be referenced directly as an ID.
-		tag := e.Name
-		if boundTag, _ := t.scope.lookupFunc(e.Name); boundTag != "" {
-			tag = boundTag
+		// an arg to another function.  It could be a built-in (as in &upper),
+		// or a user function (as in fn foo():... &foo)...
+		id := e.Name
+		if boundID, _ := t.scope.lookupFuncDeclOrParam(e.Name); boundID != "" {
```
nwt: Nit: Using ID here would make the data flow a little clearer.

```suggestion
		id := e.Name
		if boundID, _ := t.scope.lookupFuncDeclOrParam(id); boundID != "" {
```

compiler/semantic/resolver.go
```diff
@@ -2,549 +2,230 @@ package semantic
...
+		fixed: make(map[string]string),
+	}
+}
+
+// There is a funcDef for every unique lambda-unraveled instance and
+// exactly one for each such combination of decl IDs.  The instances
+// are unqiue up to the lamba params, which are in turn identified by
+// their decl ID (or built-in name).  Since lambda params originate only
+// at a built-in // or function declaration and they can't be modified
+// *(only passed by reference), the variants are determined by decl ID
```
nwt: Nit:

```suggestion
// at a built-in or function declaration and they can't be modified
// (only passed by reference), the variants are determined by decl ID
```

compiler/semantic/resolver.go
```diff
@@ -2,549 +2,230 @@ package semantic
...
+type funcDef struct {
+	tag     string
+	name    string   // original name in decl or "lambda" (for errors)
+	params  []string // params of args without any lambda args
+	lambdas []lambda
+	body    sem.Expr
+}
+
+type lambda struct {
+	param string // parameter name this lambda arg appeared asa
```
nwt: Nit:

```suggestion
	param string // parameter name this lambda arg appeared as
```

compiler/semantic/resolver.go
```diff
@@ -2,549 +2,230 @@ package semantic
...
-	return newTag
+func (r *resolver) resolveFixed(d *funcDecl, tag string) *funcDef {
+	save := r.t.scope
+	r.t.scope = NewScope(d.scope)
+	defer func() {
+		r.t.scope = save
+	}()
+	r.t.enterScope()
+	params := idsToStrings(d.lambda.Params)
+	for _, param := range idsToStrings(d.lambda.Params) {
```
nwt: Nit:

```suggestion
	params := idsToStrings(d.lambda.Params)
	for _, param := range params {
```

compiler/semantic/expr.go
```diff
@@ -779,19 +779,35 @@ func (t *translator) semMapCall(call *ast.CallExpr, args []sem.Expr) sem.Expr {
...
+		// Check argument count here for builtin functions.
+		if _, err := function.New(super.NewContext(), id, len(args)); err != nil {
+			t.error(n, err)
+			return badExpr()
+		}
+		return sem.NewCall(n, id, args)
+	}
+	return t.resolver.mustResolveCall(n, id, args)
+}
+
```
nwt: Nit: Putting this immediately before mustResolveCall would improve readability a bit.

PR #6301, opened 2025-10-13
Add sfmt.DAGSeq to format sequences

mattnibs:


mattnibs:
I added this because I often use sfmt.DAG to see what a sequence looks like a found it cumbersome to have to constantly wrap a sequence in a dag.Main all the time.

PR #6303, opened 2025-10-14
support column-id resolution in SQL pipe subqueries

mccanne:
This commit adds a pipe schema concept to the semantic pass so that column references can be properly resolved for pipe subqueries embedded as SQL table expressions.  It calls the type checker to compute the super.Type of the pipe subquery and uses that type to infer the available columns for SQL expressions.

This logic will need to be extended when we add support for correlated subqueries, likely requiring type tracking to be more tightly integrated with the translator pass (e.g., tracking the type all the way from the input to the query instead of locally analyzing the pipe subquery).

### DIFF
--- a/compiler/semantic/checker.go
+++ b/compiler/semantic/checker.go
@@ -445,7 +445,7 @@ func (c *checker) recordElems(typ super.Type, elems []sem.RecordElem) super.Type
 			elemType := c.expr(typ, elem.Expr)
 			if hasUnknown(elemType) {
 				// If we're spreading an unknown type into this record, we don't
-				// the result at all.  Return unknown for the whole thing.
+				// know the result at all.  Return unknown for the whole thing.
 				return c.unknown
 			}
 			fuser.fuse(c.expr(typ, elem.Expr))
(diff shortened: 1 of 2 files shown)


compiler/semantic/checker.go
```diff
@@ -442,6 +442,12 @@ func (c *checker) recordElems(typ super.Type, elems []sem.RecordElem) super.Type
 	for _, elem := range elems {
 		switch elem := elem.(type) {
 		case *sem.SpreadElem:
+			elemType := c.expr(typ, elem.Expr)
+			if hasUnknown(elemType) {
+				// If we're spreading an unknown type into this record, we don't
+				// the result at all.  Return unknown for the whole thing.
```

```suggestion
				// If we're spreading an unknown type into this record, we don't
				// know the result at all.  Return unknown for the whole thing.
```

compiler/semantic/checker.go
```diff
@@ -883,6 +889,20 @@ func hasUnknown(typ super.Type) bool {
...
+		return typ
+	case *super.TypeUnion:
+		for _, typ := range typ.Types {
+			if typ := recordOf(typ); typ != nil {
+				return typ
+			}
+		}
+	}
+	return nil
+}
```
nwt: Nit: Put this in schema.go after newPipeSchema since that's where it's called.

PR #6327, opened 2025-10-27
0-based index and slice expressions in pipes

mattnibs:
This commit changes the behavior of slice and index expressions so that slice and index expressions in pipe queries are 0-based while slice and index expression in SQL remain 1-based.

### DIFF
--- a/book/src/super-sql/functions/records/unflatten.md
+++ b/book/src/super-sql/functions/records/unflatten.md
@@ -37,7 +37,7 @@ _Flatten to unflatten_
 ```mdtest-spq
 # spq
 unnest flatten(this) into (
-  key[1] != "rm"
+  key[0] != "rm"
   | values collect(this)
 )
 | values unflatten(this)
--- a/book/src/super-sql/functions/strings/upper.md
+++ b/book/src/super-sql/functions/strings/upper.md
@@ -31,7 +31,7 @@ values upper(this)
 ```mdtest-spq
 # spq
 fn capitalize(str): (
-  upper(str[1:2]) + str[2:]
+  upper(str[0:1]) + str[1:]
 )
 values capitalize(this)
 # input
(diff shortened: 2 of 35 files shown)


chrismo:
my knee-jerk one word review of this idea: blerg 😆 

aren't there situations where the context of pipe v. sql isn't obvious? like maybe a lateral subquery or somesuch? 

I think in general this is going to be pretty confusing to me - but esp. if there's a place where I didn't even know I'd switched into a pipe context (or vice-versa), which ... maybe there's not ... but that sounds potentially maddening. 

But, y'all have given it more thought - what's the sales pitch for this change?

chrismo:
> aren't there situations where the context of pipe v. sql isn't obvious? like maybe a lateral subquery or somesuch?

What about this? Inspired by a test change in this PR:

```
#!/usr/bin/env bash

set -euo pipefail

export ASDF_SUPERDB_VERSION=ref:c26f3a3a1f4bcd5d7ed1d72691ba04de4d8b32d5

super --version

echo '"foobar"' |
  super -c "
    select SUBSTRING(this FROM 4 FOR 2), SUBSTRING(this FROM 4), SUBSTRING(this FOR 3)
    | flatten(this) | unnest this | values value
  " -
```
```
"ba"
"bar"
"foo"
```
_.  o O (That's a bit ugly to have to do all that to extract just the values ... oh, that's because I'm using select which will generate a column alias ... I can just use values)_
```
echo '"foobar"' |
  super -c "
    values SUBSTRING(this FROM 4 FOR 2), SUBSTRING(this FROM 4), SUBSTRING(this FOR 3)
  " -
```
```
"ar"
"ar"
"foob"
```

wait what?

---

Probably not the most convincing, since I actually said "column alias" and maybe have a chance of realizing I'm switching contexts ... but `SUBSTRING(this FROM 4 FOR 2)` feels _very_ SQL to me, fwiw.

chrismo:
as an aside - is this maybe a bug in and of itself? `values SUBSTRING(this FOR 3)` => "foob"

mattnibs:
Redoing this pr going to reopen with different functionality.

PR #6334, opened 2025-10-30
rework handling of command line input files

nwt:
Input files appearing on the command line are fed to the compiler as a single sio.Reader, preventing the semantic analyzer from seeing their super.Types.  Feed them in as paths instead.  (The semantic analyzer does not take advantage of this yet.)

As a side effect, dag.FileScan is now restartable in the vector runtime so it can now execute repeated "from <file>" subqueries.

### DIFF
--- a/runtime/exec/environment.go
+++ b/runtime/exec/environment.go
@@ -154,7 +154,7 @@ func (c *closePuller) Pull(done bool) (sbuf.Batch, error) {
 	return batch, err
 }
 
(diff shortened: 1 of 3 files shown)


nwt:
@mattnibs: As we discussed, I renamed the parameter "n" to "concurrentReaders" in 6762ddf3da977efde7f6f02f35abc30eead58276.

mattnibs (approved):

PR #6335, opened 2025-10-31
runtime/vam/expr/agg: fix panic in mathReducer.ResultAsPartial

nwt:
ResultAsPartial passes a nil *super.Context to Result, which can dereference it.


mattnibs:
Is there a test needed here?

nwt:
I'm not sure how to write a test that reliably exercises this method, but I think we're OK without one.

PR #6344, opened 2025-11-06
Count operator

### LINKED ISSUE #5182: Accessing the index of current array element within an "over" scope
## tl;dr

In a recent [community Slack thread](https://brimdata.slack.com/archives/CTSMAK6G7/p1719801076425009), a user was manipulating an array in an `over` scope in a way that required knowing the index of each value. I responded with an example that achieved this using `count()`, to which @mattnibs commented:

>I kind of think having to using streaming `count()` for this case here is kind of a hack and that you should be able to access the index of the current element from within the `over` scope. The hard part is coming up with a elegant way of doing this...

## Details

Repro is with Zed commit 6e61673.

The specific question posed by the user:

>is there a way to change this: `["a","b"]` to `[{"a":1},{"b":2}]`based on the index location of each item? (for any length array...)

My response:

```
$ zq -version
Version: v1.16.0-15-g6e61673c

$ echo '["a","b","c"]' | zq -j '
over this => (yield [{key:this,value:count()}] | unflatten(this))
| collect(this)
' -
[{"a":1},{"b":2},{"c":3}]
```

The user noted how Ruby provides this via `each_with_index` [in Ruby](https://rubyapi.org/3.3/o/s?q=each_with_index).

### LINKED ISSUE #5308: Processing values from adjacent pipeline inputs
## tl;dr

A community user recently inquired about how to compute the delta between the values in back-to-back input records. The only solution we could come up with that uses existing building blocks used `collect()` to turn the whole of the input to an an array and process it with `over`, but this has perf/memory limitations. An approach that works in a streaming mode in pipeline context would be preferred.

## Details

Repro is with Zed commit d103420.

The user question as posed in a [community Slack thread](https://brimdata.slack.com/archives/CTSMAK6G7/p1727389431796989):

> Is there a way to get the previous entry in an array?
> I want to analyze the gap between the timestamps in this input [data.zson.gz](https://github.com/user-attachments/files/17167215/data.zson.gz):
>```
>{level:"info",message:"retrieving batch",ts:"2024-09-26T22:05:45.705018615Z",offset:167000,page_size:1000,table:"NextTransactionLineLink",total_results:892574}
>{level:"info",message:"retrieving batch",ts:"2024-09-26T22:06:55.352124847Z",offset:168000,page_size:1000,table:"NextTransactionLineLink",total_results:892574}
>{level:"info",message:"retrieving batch",ts:"2024-09-26T22:08:04.096384542Z",offset:169000,page_size:1000,table:"NextTransactionLineLink",total_results:892574}
>{level:"info",message:"retrieving batch",ts:"2024-09-26T22:09:13.185300393Z",offset:170000,page_size:1000,table:"NextTransactionLineLink",total_results:892574}
>{level:"info",message:"retrieving batch",ts:"2024-09-26T22:10:21.898506504Z",offset:171000,page_size:1000,table:"NextTransactionLineLink",total_results:892574}
>{level:"info",message:"retrieving batch",ts:"2024-09-26T22:11:31.302608426Z",offset:172000,page_size:1000,table:"NextTransactionLineLink",total_results:892574}
>{level:"info",message:"retrieving batch",ts:"2024-09-26T22:12:40.698136642Z",offset:173000,page_size:1000,table:"NextTransactionLineLink",total_results:892574}
>{level:"info",message:"retrieving batch",ts:"2024-09-26T22:13:49.398172997Z",offset:174000,page_size:1000,table:"NextTransactionLineLink",total_results:892574}
>{level:"info",message:"retrieving batch",ts:"2024-09-26T22:14:58.007201613Z",offset:175000,page_size:1000,table:"NextTransactionLineLink",total_results:892574}
>```

The best we could come up with is a program like this that reads the whole of the input to an array first.

```
$ zq -version
Version: v1.18.0-4-gd1034203

$ zq -z '
put id:=count()
| collect(this)
| over this with obj=this => (
  put last_ts := (id<2) ? time(ts) : time(obj[id-2].ts), ts:=time(ts)
  | duration:=ts-last_ts
  | drop last_ts,id
)              
' data.zson.gz

{level:"info",message:"retrieving batch",ts:2024-09-26T22:05:45.705018615Z,offset:167000,page_size:1000,table:"NextTransactionLineLink",total_results:892574,duration:0s}
{level:"info",message:"retrieving batch",ts:2024-09-26T22:06:55.352124847Z,offset:168000,page_size:1000,table:"NextTransactionLineLink",total_results:892574,duration:1m9.647106232s}
{level:"info",message:"retrieving batch",ts:2024-09-26T22:08:04.096384542Z,offset:169000,page_size:1000,table:"NextTransactionLineLink",total_results:892574,duration:1m8.744259695s}
{level:"info",message:"retrieving batch",ts:2024-09-26T22:09:13.185300393Z,offset:170000,page_size:1000,table:"NextTransactionLineLink",total_results:892574,duration:1m9.088915851s}
{level:"info",message:"retrieving batch",ts:2024-09-26T22:10:21.898506504Z,offset:171000,page_size:1000,table:"NextTransactionLineLink",total_results:892574,duration:1m8.713206111s}
{level:"info",message:"retrieving batch",ts:2024-09-26T22:11:31.302608426Z,offset:172000,page_size:1000,table:"NextTransactionLineLink",total_results:892574,duration:1m9.404101922s}
{level:"info",message:"retrieving batch",ts:2024-09-26T22:12:40.698136642Z,offset:173000,page_size:1000,table:"NextTransactionLineLink",total_results:892574,duration:1m9.395528216s}
{level:"info",message:"retrieving batch",ts:2024-09-26T22:13:49.398172997Z,offset:174000,page_size:1000,table:"NextTransactionLineLink",total_results:892574,duration:1m8.700036355s}
{level:"info",message:"retrieving batch",ts:2024-09-26T22:14:58.007201613Z,offset:175000,page_size:1000,table:"NextTransactionLineLink",total_results:892574,duration:1m8.609028616s}
```

The user is quite happy with this solution since it works fine for the amount of data they've got. However, knowing Zed's per-value size limitations, this approach can only go so far. The fact the entire array needs to be assembled first also had a performance hit, plus the disadvantage that it can't stream outputs incrementally.

While the design of a precise solution is TBD, in preliminary chats, the topic of "window functions" and `LAG` / `LEAD` as they appear in SQL come to mind.

### LINKED ISSUE #5348: Passing batches of input values to a lateral subquery
## tl;dr

A proposed new `batch` operator:

```
batch 2 => (
  collect(this) |  // do something with 2-element array
)
```

## Details

Repro is with `super` commit 0af34e0.

A community user recently asked about this in the same [Slack thread](https://brimdata.slack.com/archives/CTSMAK6G7/p1729213973068049?thread_ts=1728833757.636849&cid=CTSMAK6G7) that spawned [this comment](https://github.com/brimdata/super/issues/5308#issuecomment-2412075711). In their original question they were attempting to operate on just two back-to-back input values, but then widened their inquiry to the more general case of needing to handle an indefinite number of back-to-back input vales of a similar nature.

Let's use the example of 6 incrementing numbers.

```
$ seq 6
1
2
3
4
5
6
```

Using what's in the language currently, much like the response to the user's original question, it seems this would require some initial aggregation of the input pairs into a complex value and then picking apart the complex values, potentially using `over` and a lateral subquery. For instance if we wanted to add each pairing of numbers together:

```
$ super -version
Version: v1.18.0-35-g0af34e0f

$ seq 6 | zq -z '
yield {i:(count()+1)/2,v:this}
| values:=collect(v) by i
| sort i
| over values => (
  sum(this)
)' -

3
7
11
```

However, @mccanne pointed out that for a use case like this a purpose-built operator like `batch` proposed above would provide a more direct and intuitive way to get the values into the lateral subquery.

mattnibs:
This commit adds the count operator an operator that outputs the current count of values into a record expression.

Partially fixes #5348
Partially fixes #5308
Partially fixes #5182

### DIFF
--- a/compiler/ztests/par-count.yaml
+++ b/compiler/ztests/par-count.yaml
@@ -1,14 +1,12 @@
 # Test ensures that flowgraphs utilizing the count operator
 # are not parallelized.
 script: |
-  export SUPER_VAM=1
-  super -f csup -o a.csup -c 'values {x:1}'
-  super compile -C -P 2 'from a.csup | count {row_number,...this} | aggregate sum(x)'
+  SUPER_VAM=1 super compile -C -P 2 'from /dev/null | count {row_number,...this} | aggregate sum(x)'
 
 outputs:
   - name: stdout
     data: |
-      file a.csup format csup unordered
+      file /dev/null unordered
       | count {row_number,...this}
       | aggregate
           sum:=sum(x)


compiler/ast/op.go
```diff
@@ -60,6 +60,11 @@ type (
 		Args []Expr `json:"args"`
 		Loc  `json:"loc"`
 	}
+	CountOp struct {
```
nwt, line 63: Add this to compiler/ast/unpack.go.

compiler/dag/op.go
```diff
@@ -63,6 +63,11 @@ type (
 	CombineOp struct {
 		Kind string `json:"kind" unpack:""`
 	}
+	CountOp struct {
```
nwt, line 66: Add this to compiler/dag/unpack.go.

compiler/parser/parser.peg
```diff
@@ -467,6 +468,19 @@ CallOp
...
+      return &ast.CountOp{
+        Kind: "CountOp",
+        Expr: rec.(ast.Expr),
+      }, nil
+    }
+  / COUNT &EndOfOp {
+      return &ast.CountOp{
+        Kind: "CountOp",
+      }, nil
+    }
```
nwt: Nit: I don't think this form is worth its weight.

mattnibs: ~~I needed to do this otherwise it would try and parse and fail something like `count := count+1`~~

Sorry I now see what you are saying, I suppose I don't have strong feelings

mattnibs: I think the `as` keyword is almost always optional should we get rid of that syntax here then?

nwt: Yes, I vote for dropping `as`.

mccanne: What would this do?
```
values 1,2,3 | count
```

mattnibs: I don't have a dog in this fight but if we go with what @nwt says this would have an error that there is no user defined op called `count` or something like that.

mccanne: I don't think I'm following

mattnibs:
```
$ super -c 'values 1,2,3 | count'
no such user operator: "count" at line 1, column 16:
values 1,2,3 | count
               ~~~~~
```

compiler/rungen/op.go
```diff
@@ -267,6 +268,15 @@ func (b *Builder) compileLeaf(o dag.Op, parent sbuf.Puller) (sbuf.Puller, error)
...
 	case *dag.AggregateOp:
 		return b.compileAggregate(parent, v)
+	case *dag.CountOp:
+		var e expr.Evaluator
+		if v.Expr != nil {
+			var err error
+			if e, err = b.compileExpr(v.Expr); err != nil {
+				return nil, err
+			}
+		}
```
nwt, line 278:
```suggestion
		e, err := b.compileExpr(v.Expr)
		if err != nil {
			return nil, err
		}
```

compiler/rungen/vop.go
```diff
@@ -206,6 +206,15 @@ func (b *Builder) compileVamLeaf(o dag.Op, parent vector.Puller) (vector.Puller,
...
 	case *dag.AggregateOp:
 		return b.compileVamAggregate(o, parent)
+	case *dag.CountOp:
+		var e vamexpr.Evaluator
+		if o.Expr != nil {
+			var err error
+			if e, err = b.compileVamExpr(o.Expr); err != nil {
+				return nil, err
+			}
+		}
```
nwt, line 216:
```suggestion
		e, err := b.compileVamExpr(o.Expr)
		if err != nil {
			return nil, err
		}
```

compiler/rungen/op.go
mattnibs: `v.Expr` can be nil if `count` only has an identifier expression `count {count}`

compiler/rungen/vop.go
mattnibs: v.Expr can be nil if `count` only has an identifier expression `count {count}`

compiler/sfmt/dag.go
```diff
@@ -220,6 +205,25 @@ func (c *canonDAG) binary(e *dag.BinaryExpr, parent string) {
...
+		switch e := elem.(type) {
+		case *dag.Field:
+			c.write(sup.QuotedName(e.Name))
+			c.write(":")
+			c.expr(e.Value, "")
+		case *dag.Spread:
+			c.write("...")
+			c.expr(e.Expr, "")
+		default:
+			c.write("sfmt: unknown record elem type: %T", e)
```
nwt:
```suggestion
			panic(e)
```

runtime/vam/op/count.go
```diff
@@ -0,0 +1,48 @@
...
+
+func NewCount(sctx *super.Context, parent vector.Puller, alias string, in expr.Evaluator) *Count {
+	o := &Count{parent: parent, alias: alias}
+	var elems []expr.RecordElem
+	elems = append(elems, expr.RecordElem{Name: alias, Expr: evalfunc(o.evalCount)})
+	if in != nil {
+		elems = append(elems, expr.RecordElem{Expr: in})
+	}
+	o.expr = expr.NewRecordExpr(sctx, elems)
+	return o
```
nwt, line 25: Nit: Clearer to create the return value here rather than up above.

```suggestion
	return &Count{parent, alias, expr.NewRecordExpr(sctx, elems), 0}
```

mattnibs: I have to have it this way since the alias argument needs to reference `o.evalCount` in its evaluator.

compiler/ztests/par-count.yaml
```diff
@@ -1,45 +1,16 @@
...
 script: |
-  export SUPER_DB=test
-  super db init -q
-  super db create -q -orderby ts test
-  super db compile -C -P 2 "from test | count() by y" | sed -e 's/pool .*/.../'
-  echo ===
-  super db compile -C -P 2 "from test | count()" | sed -e 's/pool .*/.../'
+  export SUPER_VAM=1
+  super -f csup -o a.csup -c 'values {x:1}'
+  super compile -C -P 2 'from a.csup | count {row_number,...this} | aggregate sum(x)'
```
nwt: Nit: No need to create a.csup.

```suggestion
  SUPER_VAM=1 super compile -C -P 2 'from /dev/null | count {row_number,...this} | aggregate sum(x)'
```

PR #6348, opened 2025-11-10
Add back 0-based indexes

mattnibs:
This commit adds back 0-based index/slice expressions to the language. It also institutes pragmas declarations that allow users to specify 1-based or 0-based indexes for a given scope.

### DIFF
--- a/compiler/ast/decl.go
+++ b/compiler/ast/decl.go
@@ -27,6 +27,13 @@ type OpDecl struct {
 	Loc    `json:"loc"`
 }
 
+type PragmaDecl struct {
+	Kind string `json:"kind" unpack:""`
+	Name *ID    `json:"name"`
+	Expr Expr   `json:"expr"`
+	Loc  `json:"loc"`
+}
+
 type QueryDecl struct {
 	Kind string `json:"kind" unpack:""`
 	Name *ID    `json:"name"`
@@ -41,8 +48,9 @@ type TypeDecl struct {
 	Loc  `json:"loc"`
 }
 
-func (*ConstDecl) declNode() {}
-func (*FuncDecl) declNode()  {}
-func (*OpDecl) declNode()    {}
-func (*QueryDecl) declNode() {}
-func (*TypeDecl) declNode()  {}
+func (*ConstDecl) declNode()  {}
+func (*FuncDecl) declNode()   {}
+func (*OpDecl) declNode()     {}
+func (*PragmaDecl) declNode() {}
+func (*QueryDecl) declNode()  {}
+func (*TypeDecl) declNode()   {}
--- a/compiler/parser/parser.peg
+++ b/compiler/parser/parser.peg
@@ -33,7 +33,7 @@ Seq
 SeqTail = __ Pipe __ o:PipeOp { return o, nil }
 
 Decl
-  = v:(ConstDecl / FuncDecl / OpDecl / QueryDecl / TypeDecl) _ { return v, nil }
+  = v:(ConstDecl / FuncDecl / OpDecl / PragmaDecl / QueryDecl / TypeDecl ) _ { return v, nil }
 
 ConstDecl
   = CONST _ name:Identifier __ "=" __ expr:Expr {
@@ -102,6 +102,16 @@ OpDecl
 ScopeBody
   = "(" __ scope:Scope __ ")"   { return []any{scope}, nil }
   / "(" __ seq:Seq __ ")"       { return seq, nil }
+
+PragmaDecl
+  = PRAGMA _ name:Identifier __ "=" __ e:Expr {
+      return &ast.PragmaDecl{
+        Kind: "PragmaDecl",
+        Name: name.(*ast.ID),
+        Expr: e.(ast.Expr),
+        Loc: loc(c),
+      }, nil
+    }
 
 TypeDecl
   = TYPE _ name:Identifier __ "=" __ typ:Type {
@@ -2452,6 +2462,7 @@ OUTER      = "OUTER"i           !IdentifierRest
 OUTPUT     = "OUTPUT"i          !IdentifierRest
 PASS       = "PASS"i            !IdentifierRest
 PUT        = "PUT"i             !IdentifierRest
+PRAGMA     = "PRAGMA"i          !IdentifierRest
 RECURSIVE  = "RECURSIVE"i       !IdentifierRest
 RENAME     = "RENAME"i          !IdentifierRest
 RIGHT      = "RIGHT"i           !IdentifierRest
--- a/compiler/semantic/expr.go
+++ b/compiler/semantic/expr.go
@@ -148,6 +148,7 @@ func (t *translator) expr(e ast.Expr) sem.Expr {
 			Node:  e,
 			Expr:  expr,
 			Index: index,
+			Base1: t.scope.indexBase() == 1,
 		}
 	case *ast.IsNullExpr:
 		expr := t.expr(e.Expr)
@@ -246,10 +247,11 @@ func (t *translator) expr(e ast.Expr) sem.Expr {
 		from := t.exprNullable(e.From)
 		to := t.exprNullable(e.To)
 		return &sem.SliceExpr{
-			Node: e,
-			Expr: expr,
-			From: from,
-			To:   to,
+			Node:  e,
+			Expr:  expr,
+			From:  from,
+			To:    to,
+			Base1: t.scope.indexBase() == 1,
 		}
 	case *ast.SQLTimeExpr:
 		if e.Value.Type != "string" {
@@ -308,9 +310,10 @@ func (t *translator) expr(e ast.Expr) sem.Expr {
 		// XXX type checker should remove this check when it finds it redundant
 		is := sem.NewCall(e, "is", []sem.Expr{expr, &sem.LiteralExpr{Node: e.Expr, Value: "<string>"}})
 		slice := &sem.SliceExpr{
-			Node: e,
-			Expr: expr,
-			From: t.exprNullable(e.From),
+			Node:  e,
+			Expr:  expr,
+			From:  t.exprNullable(e.From),
+			Base1: true,
 		}
 		if e.For != nil {
 			to := t.expr(e.For)
@@ -1152,7 +1155,7 @@ func scalarSubqueryCheck(n ast.Node) *sem.ValuesOp {
 	indexExpr := &sem.IndexExpr{
 		Node:  n,
 		Expr:  sem.NewThis(n, nil),
-		Index: &sem.LiteralExpr{Node: n, Value: "1"},
+		Index: &sem.LiteralExpr{Node: n, Value: "0"},
 	}
 	innerCond := &sem.CondExpr{
 		Node: n,
--- a/compiler/semantic/op.go
+++ b/compiler/semantic/op.go
@@ -1075,6 +1075,8 @@ func (t *translator) decls(decls []ast.Decl) {
 			t.funcDecl(d)
 		case *ast.OpDecl:
 			t.opDecl(d)
+		case *ast.PragmaDecl:
+			t.pragmaDecl(d)
 		case *ast.QueryDecl:
 			t.queryDecl(d)
 		case *ast.TypeDecl:
@@ -1196,6 +1198,33 @@ func (t *translator) opDecl(d *ast.OpDecl) {
 	}
 	if err := t.scope.BindSymbol(d.Name.Name, &opDecl{ast: d, scope: t.scope}); err != nil {
 		t.error(d, err)
+	}
+}
+
+func (t *translator) pragmaDecl(d *ast.PragmaDecl) {
+	switch d.Name.Name {
+	case "index_base":
+		if _, ok := t.scope.pragmas["index_base"]; ok {
+			t.error(d.Name, errors.New("index_base defined multiple times in scope"))
+			return
+		}
+		e := t.expr(d.Expr)
+		val, ok := t.mustEval(e)
+		if !ok {
+			return
+		}
+		if !super.IsInteger(val.Type().ID()) || val.IsNull() {
+			t.error(d.Expr, fmt.Errorf("index_base value must be an integer: %s", sup.FormatValue(val)))
+			return
+		}
+		v := int(val.AsInt())
+		if v < 0 || v > 1 {
+			t.error(d.Name, errors.New("index_base must be 0 or 1"))
+			return
+		}
+		t.scope.pragmas["index_base"] = v
+	default:
+		t.error(d.Name, fmt.Errorf("unknown pragma %q", d.Name.Name))
 	}
 }
 
@@ -1474,11 +1503,12 @@ func (t *translator) mustEvalPositiveInteger(ae ast.Expr) int {
 	}
 	if !super.IsInteger(val.Type().ID()) || val.IsNull() {
 		t.error(ae, fmt.Errorf("expression value must be an integer value: %s", sup.FormatValue(val)))
-		return -1
+		return 0
 	}
 	v := int(val.AsInt())
 	if v < 0 {
 		t.error(ae, errors.New("expression value must be a positive integer"))
+		return 0
 	}
 	return v
 }
--- a/compiler/semantic/scope.go
+++ b/compiler/semantic/scope.go
@@ -11,13 +11,19 @@ import (
 
 type Scope struct {
 	parent  *Scope
+	pragmas map[string]any
 	symbols map[string]*entry
 	ctes    map[string]*ast.SQLCTE
 	schema  schema
 }
 
 func NewScope(parent *Scope) *Scope {
-	return &Scope{parent: parent, symbols: make(map[string]*entry), ctes: make(map[string]*ast.SQLCTE)}
+	return &Scope{
+		parent:  parent,
+		pragmas: make(map[string]any),
+		symbols: make(map[string]*entry),
+		ctes:    make(map[string]*ast.SQLCTE),
+	}
 }
 
 type entry struct {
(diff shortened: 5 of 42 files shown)


compiler/ast/decl.go
```diff
@@ -41,8 +41,16 @@ type TypeDecl struct {
 	Loc  `json:"loc"`
 }
 
-func (*ConstDecl) declNode() {}
-func (*FuncDecl) declNode()  {}
-func (*OpDecl) declNode()    {}
-func (*QueryDecl) declNode() {}
-func (*TypeDecl) declNode()  {}
+type PragmaDecl struct {
```
nwt: Put this between OpDecl and QueryDecl.

compiler/ast/decl.go
```diff
@@ -41,8 +41,16 @@ type TypeDecl struct {
...
+	Expr Expr   `json:"expr"`
+	Loc  `json:"loc"`
+}
+
+func (*ConstDecl) declNode()  {}
+func (*FuncDecl) declNode()   {}
+func (*OpDecl) declNode()     {}
+func (*QueryDecl) declNode()  {}
+func (*TypeDecl) declNode()   {}
+func (*PragmaDecl) declNode() {}
```
nwt:
```suggestion
func (*OpDecl) declNode()     {}
func (*PragmaDecl) declNode() {}
func (*QueryDecl) declNode()  {}
func (*TypeDecl) declNode()   {}
```

compiler/parser/parser.peg
```diff
@@ -33,7 +33,7 @@ Seq
 SeqTail = __ Pipe __ o:PipeOp { return o, nil }
 
 Decl
-  = v:(ConstDecl / FuncDecl / OpDecl / QueryDecl / TypeDecl) _ { return v, nil }
+  = v:(ConstDecl / FuncDecl / OpDecl / QueryDecl / TypeDecl / PragmaDecl) _ { return v, nil }
```
nwt:
```suggestion
  = v:(ConstDecl / FuncDecl / OpDecl / PragmaDecl / QueryDecl / TypeDecl) _ { return v, nil }
```

compiler/parser/parser.peg
```diff
@@ -113,6 +113,16 @@ TypeDecl
       }, nil
     }
 
+PragmaDecl
```
nwt, line 106: Put this between ScopeBody and TypeDecl.

compiler/semantic/op.go
```diff
@@ -1055,6 +1055,8 @@ func (t *translator) decls(decls []ast.Decl) {
 			t.queryDecl(d)
 		case *ast.TypeDecl:
 			t.typeDecl(d)
+		case *ast.PragmaDecl:
```
nwt: Put this between OpDecl and QueryDecl.

compiler/semantic/op.go
```diff
@@ -1175,6 +1177,33 @@ func (t *translator) opDecl(d *ast.OpDecl) {
 	}
 }
 
+func (t *translator) pragmaDecl(d *ast.PragmaDecl) {
+	switch d.Name.Name {
+	case "index_base":
+		if _, ok := t.scope.pragmas["index_base"]; ok {
+			t.error(d.Name, errors.New("\"index_base\" defined multiple times in scope"))
```
nwt:
```suggestion
			t.error(d.Name, errors.New(`"index_base" defined multiple times in scope`))
```

compiler/semantic/expr.go
```diff
@@ -1126,12 +1129,12 @@ func (t *translator) subqueryExpr(astExpr ast.Expr, array bool, body ast.Seq) *s
...
 		// SQL expects a record with a single column result so unravel this
 		// condition with this complex cleanup...
-		e.Body.Append(scalarSubqueryCheck(astExpr))
+		e.Body.Append(t.scalarSubqueryCheck(astExpr))
 	}
 	return e
 }
 
-func scalarSubqueryCheck(n ast.Node) *sem.ValuesOp {
+func (t *translator) scalarSubqueryCheck(n ast.Node) *sem.ValuesOp {
```
nwt: This doesn't appear to be necessary.

compiler/semantic/op.go
```diff
@@ -1175,6 +1177,33 @@ func (t *translator) opDecl(d *ast.OpDecl) {
...
+		if !ok {
+			return
+		}
+		if !super.IsInteger(val.Type().ID()) || val.IsNull() {
+			t.error(d.Expr, fmt.Errorf("index_base value must be an integer: %s", sup.FormatValue(val)))
+			return
+		}
+		v := int(val.AsInt())
+		if v < 0 || v > 1 {
+			t.error(d.Name, errors.New("invalid value for index_base (must be either 0 or 1)"))
```
nwt:
```suggestion
			t.error(d.Name, errors.New("index_base must be 0 or 1"))
```

compiler/semantic/op.go
```diff
@@ -1175,6 +1177,33 @@ func (t *translator) opDecl(d *ast.OpDecl) {
...
+			return
+		}
+		t.scope.pragmas["index_base"] = v
+	default:
+		t.error(d.Name, fmt.Errorf("unknown directive %q", d.Name.Name))
```
nwt:
```suggestion
		t.error(d.Name, fmt.Errorf("unknown pragma %q", d.Name.Name))
```

compiler/semantic/scope.go
```diff
@@ -11,13 +11,19 @@ import (
...
 	schema  schema
 }
 
 func NewScope(parent *Scope) *Scope {
-	return &Scope{parent: parent, symbols: make(map[string]*entry), ctes: make(map[string]*ast.SQLCTE)}
+	return &Scope{
+		parent:  parent,
+		symbols: make(map[string]*entry),
+		ctes:    make(map[string]*ast.SQLCTE),
+		pragmas: make(map[string]any),
```
nwt: Nit: Follow the declaration order.

```suggestion
		pragmas: make(map[string]any),
		symbols: make(map[string]*entry),
		ctes:    make(map[string]*ast.SQLCTE),
```

compiler/semantic/scope.go
```diff
@@ -153,6 +159,19 @@ func (s *Scope) resolve(n ast.Node, path field.Path) (sem.Expr, error) {
...
+	scope := s
+	for {
+		if v, ok := s.pragmas["index_base"]; ok {
+			return v.(int)
+		}
+		if scope.parent == nil {
+			return 0
+		}
+		scope = scope.parent
+	}
```
nwt: This doesn't seem right since it's equivalent this:

```go
	if v, ok := s.pragmas["index_base"]; ok {
		return v.(int)
	}
	return 0
```

compiler/semantic/op.go
```diff
@@ -1175,6 +1177,33 @@ func (t *translator) opDecl(d *ast.OpDecl) {
 	}
 }
 
+func (t *translator) pragmaDecl(d *ast.PragmaDecl) {
+	switch d.Name.Name {
+	case "index_base":
+		if _, ok := t.scope.pragmas["index_base"]; ok {
+			t.error(d.Name, errors.New(`"index_base" defined multiple times in scope`))
```
nwt: This error string has quotes around index_base but the others don't.

compiler/semantic/scope.go
mattnibs: I should be use `scope.pragmas` instead of `s.pragmas`

mattnibs: How the heck are the test passing tho

nwt: OK. This should have test coverage.

runtime/vam/expr/slice.go
```diff
@@ -53,17 +55,17 @@ func (s *sliceExpr) eval(vecs ...vector.Any) vector.Any {
...
+		return s.evalStringOrBytes(container, from, to, s.base1)
 	case vector.KindError:
 		return container
 	default:
 		return vector.NewWrappedError(s.sctx, "sliced value is not array, set, bytes, or string", container)
 	}
 }
 
-func (s *sliceExpr) evalArrayOrSlice(vec, fromVec, toVec vector.Any) vector.Any {
+func (s *sliceExpr) evalArrayOrSlice(vec, fromVec, toVec vector.Any, sql bool) vector.Any {
```
nwt: s/sql/base1/ in this file.

runtime/vam/expr/index.go
```diff
@@ -29,23 +30,23 @@ func (i *Index) eval(args ...vector.Any) vector.Any {
...
+		return indexRecord(i.sctx, container, index, i.base1)
 	case vector.KindMap:
 		panic("vector index operations on maps not supported")
 	default:
 		return vector.NewMissing(i.sctx, this.Len())
 	}
 }
 
-func indexArrayOrSet(sctx *super.Context, vec, indexVec vector.Any) vector.Any {
+func indexArrayOrSet(sctx *super.Context, vec, indexVec vector.Any, sql bool) vector.Any {
```
nwt: s/sql/base1/ in this file.

compiler/semantic/scope.go
mattnibs: agreed

runtime/sam/expr/eval.go
```diff
@@ -649,73 +649,87 @@ type Index struct {
...
+		return indexRecordByName(i.sctx, typ, container.Bytes(), index)
 	case *super.TypeMap:
 		return indexMap(i.sctx, typ, container.Bytes(), index)
 	default:
 		return i.sctx.Missing()
 	}
 }
 
-func indexArrayOrSet(sctx *super.Context, inner super.Type, vector scode.Bytes, index super.Value) super.Value {
+func getNumericIndex(sctx *super.Context, sql bool, index super.Value) (int, bool, *super.Value) {
```
nwt:
```suggestion
func getNumericIndex(sctx *super.Context, index super.Value, base1 bool) (int, bool, *super.Value) {
```

compiler/semantic/scope.go
```diff
@@ -153,6 +159,19 @@ func (s *Scope) resolve(n ast.Node, path field.Path) (sem.Expr, error) {
...
+func (s *Scope) indexBase() int {
+	scope := s
+	for {
+		if v, ok := scope.pragmas["index_base"]; ok {
+			return v.(int)
+		}
+		if scope.parent == nil {
+			return 0
+		}
+		scope = scope.parent
```
nwt: Nit: Just use s.

```suggestion
	for {
		if v, ok := s.pragmas["index_base"]; ok {
			return v.(int)
		}
		if s.parent == nil {
			return 0
		}
		s = s.parent
```

compiler/ztests/pragma-index-base-index.yaml
```diff
@@ -0,0 +1,26 @@
...
+      const s = "base1"
+      values {s,v:this[1]}
+    )
+    (
+      const s = "base0"
+      pragma index_base = 0
+      values {s,v:this[1]}
+    )
+  | sort s
+
```
nwt, line 14:
```suggestion

vector: true

```

compiler/ztests/pragma-index-base-slice.yaml
```diff
@@ -0,0 +1,26 @@
...
+      const s = "base1" 
+      values {s, from:a[from:], to:a[:to], both:a[from:to]}
+    )
+    (
+      const s = "base0"
+      pragma index_base = 0
+      values {s, from:a[from:], to:a[:to], both:a[from:to]}
+    )
+  | sort s
+
```
nwt, line 14:
```suggestion

vector: true

```

runtime/sam/expr/slice.go
```diff
@@ -85,7 +87,7 @@ func (s *Slice) Eval(this super.Value) super.Value {
 	return super.NewValue(elem.Type(), bytes)
 }
 
-func sliceIndex(this super.Value, slot Evaluator, length int) (int, error) {
+func sliceIndex(this super.Value, slot Evaluator, length int, sql bool) (int, error) {
```
nwt:
```suggestion
func sliceIndex(this super.Value, slot Evaluator, length int, base1 bool) (int, error) {
```

runtime/vam/expr/index.go
```diff
@@ -85,12 +86,12 @@ func indexArrayOrSet(sctx *super.Context, vec, indexVec vector.Any) vector.Any {
 	return out
 }
 
-func indexRecord(sctx *super.Context, vec, indexVec vector.Any) vector.Any {
+func indexRecord(sctx *super.Context, vec, indexVec vector.Any, sql bool) vector.Any {
```
nwt:
```suggestion
func indexRecord(sctx *super.Context, vec, indexVec vector.Any, base1 bool) vector.Any {
```

runtime/vam/expr/slice.go
```diff
@@ -125,11 +127,11 @@ func (s *sliceExpr) evalArrayOrSlice(vec, fromVec, toVec vector.Any) vector.Any
 	return out
 }
 
-func (s *sliceExpr) evalStringOrBytes(vec, fromVec, toVec vector.Any) vector.Any {
+func (s *sliceExpr) evalStringOrBytes(vec, fromVec, toVec vector.Any, sql bool) vector.Any {
```
nwt:
```suggestion
func (s *sliceExpr) evalStringOrBytes(vec, fromVec, toVec vector.Any, base1 bool) vector.Any {
```

runtime/vam/expr/slice.go
```diff
@@ -170,30 +172,30 @@ func (s *sliceExpr) evalStringOrBytes(vec, fromVec, toVec vector.Any) vector.Any
 	return out
 }
 
-func (s *sliceExpr) evalStringOrBytesFast(vec vector.Any, from, to int) (vector.Any, bool) {
+func (s *sliceExpr) evalStringOrBytesFast(vec vector.Any, from, to int, sql bool) (vector.Any, bool) {
```
nwt:
```suggestion
func (s *sliceExpr) evalStringOrBytesFast(vec vector.Any, from, to int, base1 bool) (vector.Any, bool) {
```

runtime/vam/expr/slice.go
```diff
@@ -279,8 +281,8 @@ func sliceIsConstIndex(vec vector.Any) (int, bool) {
 	return 0, false
 }
 
-func sliceIndex(idx, size int) int {
-	if idx > 0 {
+func sliceIndex(idx, size int, sql bool) int {
```
nwt:
```suggestion
func sliceIndex(idx, size int, base1 bool) int {
```

chrismo:
👍🏻 I like this much more than the prior PR (#6327)

chrismo:
I am still curious what this behavior will look like when pragma is set to 1?

`values SUBSTRING(this FOR 3) => "foob"`

philrz:
@chrismo: I suspect you may have already done some of your own testing between the merge of this PR and the related #6351. But the answer to the question in your [comment](https://github.com/brimdata/super/pull/6348#issuecomment-3514191258) is that for this specific example the result is the same regardless of the pragma setting:

```
$ super -version
Version: d84d65640

$ echo '"foobar"' | super -c "values SUBSTRING(this FOR 3)" -
"foo"

$ echo '"foobar"' | super -c "pragma index_base = 1 values SUBSTRING(this FOR 3)" -
"foo"
```

That happens to match what traditional SQL systems do as well.

```
$ duckdb -c "SELECT SUBSTRING(s FOR 3) FROM (SELECT 'foobar' AS s);"
┌────────────────────────────────────────────┐
│ main."substring"(s, 1, CAST(3 AS INTEGER)) │
│                  varchar                   │
├────────────────────────────────────────────┤
│ foo                                        │
└────────────────────────────────────────────┘

$ psql postgres -c "SELECT SUBSTRING(s FOR 3) FROM (SELECT 'foobar' AS s);"
 substring 
-----------
 foo
(1 row)
```

I think this makes sense since, as it's used here, the concept of "index base" doesn't really come into play, since the query is just asking for the first three characters of the string. In this regard, I think you were certainly correct to call out that unexpected 4-character `"foob"` result as a likely bug in that closed-not-merged #6327. I'm not sure if there was a reason at the time for the behavior that we may have thought was defensible, but it's all in the past now.

Where the pragma still has some impact is if we include the "start position" (to use the term from the [SQL spec](https://www.contrib.andrew.cmu.edu/~shadow/sql/sql1992.txt)) since it behaves like an index base. Per #6351, we made the decision for now to let that be affected by the pragma setting. That means the default behavior does diverge from SQL, but strict backward compatibility with SQL is why we added the pragma.

```
$ duckdb -c "SELECT SUBSTRING(s FROM 2 FOR 3) FROM (SELECT 'foobar' AS s);"
┌───────────────────────────┐
│ main."substring"(s, 2, 3) │
│          varchar          │
├───────────────────────────┤
│ oob                       │
└───────────────────────────┘

$ psql postgres -c "SELECT SUBSTRING(s FROM 2 FOR 3) FROM (SELECT 'foobar' AS s);"
 substring 
-----------
 oob
(1 row)

$ super -c "SELECT SUBSTRING(s FROM 2 FOR 3) FROM (SELECT 'foobar' AS s);"
{"SUBSTRING(s FROM 2 FOR 3)":"oba"}

$ super -c "pragma index_base = 1 SELECT SUBSTRING(s FROM 2 FOR 3) FROM (SELECT 'foobar' AS s);"
{"SUBSTRING(s FROM 2 FOR 3)":"oob"}
```

Since you showed an interest in this topic and your feedback has been helpful, I'll flag you down on community Slack to give you a bit more detail about the approaches we considered here and see if you have any other reactions.

chrismo:
> unexpected 4-character `"foob"` result as a likely bug in that closed-not-merged #6327

Yeah, that clearly seemed to be a bug. Glad to see that it didn't propagate into this PR :)

chrismo:
> I think this makes sense since, as it's used here, the concept of "index base" doesn't really come into play, since the query is just asking for the first three characters of the string.

Yeah, I dunno if I agree. 🤔 I guess this is the sort of scenario that led y'all to attempt "1-based in SQL, 0-based elsewhere" ... It seems inconsistent as just `FOR 3` still implies an index_base?

PR #6350, opened 2025-11-10
UNION ALL: merge schemas if possible

### LINKED ISSUE #6342: SQL: Schema info not present after UNION ALL
## tl;dr

```
$ super -c "
select * from (
  select 1 as x
  union all
  select 2 as x
)
join (
  select 1 as y
  union all
  select 3 as y
) on x=y"

"x": ambiguous column reference at line 11, column 6:
) on x=y
     ~
"y": ambiguous column reference at line 11, column 8:
) on x=y
       ~
```

## Details

Repro is with super commit 6f83f8d.

By comparison, if the joined data comes from Parquet file inputs rather than as results from the `UNION ALL`, the join is successful due to the schema info being present.

```
$ super -version
Version: 6f83f8d4f

$ super -f parquet -o x.parquet -c "
    select 1 as x
    union all
    select 2 as x" &&
  super -f parquet -o y.parquet -c "
    select 1 as y
    union all
    select 3 as y" &&
  super -c "select * from x.parquet join (select * from y.parquet) on x=y"

{x:1,y:1}
```

And the original query works as expected in Postgres.

```
$ psql --version

$ psql postgres -c "
select * from (
  select 1 as x
  union all
  select 2 as x
)
join (
  select 1 as y
  union all
  select 3 as y
) on x=y"

 x | y 
---+---
 1 | 1
(1 row)
```

mattnibs:
This commit adds functionality in the semantic pass of UNION ALL that merges unioned schemas if possible.

Closes #6342

### DIFF
--- a/compiler/semantic/sql.go
+++ b/compiler/semantic/sql.go
@@ -513,9 +513,9 @@ func (t *translator) sqlQueryBody(query ast.SQLQueryBody, seq sem.Seq) (sem.Seq,
 		return seq, sch
 	case *ast.SQLUnion:
 		left, leftSch := t.sqlQueryBody(query.Left, seq)
-		left, _ = endScope(query.Left.(ast.Node), leftSch, left)
+		left, leftSch = endScope(query.Left.(ast.Node), leftSch, left)
 		right, rightSch := t.sqlQueryBody(query.Right, seq)
-		right, _ = endScope(query.Right.(ast.Node), rightSch, right)
+		right, rightSch = endScope(query.Right.(ast.Node), rightSch, right)
 		out := sem.Seq{
 			&sem.ForkOp{Node: query, Paths: []sem.Seq{left, right}},
 			// This used to be dag.Combine but we don't have combine in the sem tree,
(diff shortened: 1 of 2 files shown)


nwt:
This isn't specific to UNION ALL. The output schema for UNION and UNION ALL is the same.

compiler/semantic/sql.go
```diff
@@ -527,12 +527,21 @@ func (t *translator) sqlQueryBody(query ast.SQLQueryBody, seq sem.Seq) (sem.Seq,
...
 	}
 }
 
+func unionSchema(a, b schema) schema {
+	asch, aok := a.(*staticSchema)
+	bsch, bok := b.(*staticSchema)
+	if !aok || !bok || !slices.Equal(asch.columns, bsch.columns) {
+		return &dynamicSchema{}
+	}
+	return &staticSchema{slices.Clone(asch.columns)}
```
nwt, line 542: I think this is correct as far as it goes but not sufficient.  Per other systems, the input tables just need the same number of columns, and then the output table has the same columns (in the same order) as the left input.

In terms of a hypothetical schema.outColumns method that returns the output table columns for a schema:

```go
func unionSchema(a, b schema) schema {
	if cols := a.outColumns(); len(cols) == len(b.outColumns()) {
		return &staticSchema{cols}
	}
	return &dynamicSchema{}
```

mattnibs:
Going to reopen this in another form with different functionality.

PR #6355, opened 2025-11-12
Drop support for streaming aggregations

### LINKED ISSUE #5969: Nested aggregate functions
## tl;dr

SuperSQL is currently accepting nested aggregate functions, but it looks like they're not supposed to be permitted in SQL. Therefore it seems we should either stop supporting them or document why SuperSQL's approach is somehow valid/appropriate in contrast to others.

## Details

Repro is with super commit 3c0e543. This was found via a query from [a sqllogictest](https://github.com/duckdb/duckdb/blob/8be6ed418aff9e8b2c6a5faa549e93e142adf297/test/sql/aggregate/group/test_group_by.test#L91-L94).

Applying the data and query from the test in Postgres:

```
$ psql postgres
psql (17.5 (Homebrew))
Type "help" for help.

postgres=# select * from test;
 a  | b  
----+----
 11 | 22
 13 | 22
 12 | 21
(3 rows)

postgres=# SELECT b % 2 AS f, COUNT(SUM(a)) FROM test GROUP BY f;
ERROR:  aggregate function calls cannot be nested
LINE 1: SELECT b % 2 AS f, COUNT(SUM(a)) FROM test GROUP BY f;
```

Whereas in SuperDB:

```
$ cat test.sup
{a:11(int32),b:22(int32)}
{a:13(int32),b:22(int32)}
{a:12(int32),b:21(int32)}

$ super -version
Version: 3c0e54399

$ super -c "SELECT b % 2 AS f, COUNT(SUM(a)) FROM 'test.sup' GROUP BY f;"
{f:0,"COUNT(SUM(a))":2(uint64)}
{f:1,"COUNT(SUM(a))":1(uint64)}
```

I wasn't previously familiar with the concept of nested aggregate functions, but the tl;dr of what I see in web searches is that they're supposedly not permitted in SQL, and users that think they want them should instead use multi-stage aggregations with subqueries or CTEs.

The fact SuperDB is producing a result at all is curious. If the query did work in basic SQL, I'm not sure what the expected result would have been. But taking my own stab at it, if I drop the `COUNT` and make it a valid SQL query we get:

```
postgres=# SELECT b % 2 AS f, SUM(a) FROM test GROUP BY f;
 f | sum 
---+-----
 0 |  24
 1 |  12
(2 rows)
```

So since the `SUM` is producing single values as we'd expect, I'd assume the `COUNT` of each would have been expected to be `1`, making SuperDB's result of `2` incorrect. If it's actually doing something smart and I'm not seeing it, then we probably should document it and defend it. 😄

mattnibs:
Closes #5969

### DIFF
--- a/book/src/super-sql/aggregates/and.md
+++ b/book/src/super-sql/aggregates/and.md
@@ -25,24 +25,10 @@ true
 false
 ```
 
-Continuous AND of simple sequence:
-```mdtest-spq
-# spq
-values and(this)
-# input
-true
-false
-true
-# expected output
-true
-false
-false
-```
-
 Unrecognized types are ignored and not coerced for truthiness:
 ```mdtest-spq
 # spq
-values and(this)
+and(this)
 # input
 true
 "foo"
@@ -50,10 +36,6 @@ true
 false
 true
 # expected output
-true
-true
-true
-false
 false
 ```
 
--- a/book/src/super-sql/aggregates/any.md
+++ b/book/src/super-sql/aggregates/any.md
@@ -27,19 +27,6 @@ any(this)
 1
 ```
 
-Continuous any over a simple sequence:
-```mdtest-spq
-# spq
-values any(this)
-# input
-1 2 3 4
-# expected output
-1
-1
-1
-1
-```
-
 Any is not sensitive to mixed types as it just picks one:
 ```mdtest-spq
 # spq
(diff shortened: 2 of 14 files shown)


nwt (commented):
Looks good but I'll leave approval to @mccanne in case he has input on the docs.

PR #6356, opened 2025-11-12
Add schema support for SQL set operations

### LINKED ISSUE #5755: SQL-compatible UNION behavior (e.g., column-based)
## tl;dr

We've reached consensus that the initial GA release of SuperDB will include the `UNION ALL` support added in #5735, which is in some ways not fully compatible with typical SQL behavior, such as how SuperDB's is effectively "union by name" while SQL's is "union by column".

## Details

At the time this issue is being filed, super is at commit a6eae50. The example shown below is likely not the only area where SuperSQL's current `UNION` behavior differs somewhat from that of traditional SQL.

This is illustrated using an abbreviated example of something from the [w3schools SQL tutorial](https://www.w3schools.com/sql/sql_union.asp).

Given input data files:

```
$ cat customers.csv 
City,Country
Berlin,Germany
México D.F.,Mexico
México D.F.,Mexico

$ cat suppliers.csv 
City,Country
Londona,UK
New Orleans,USA
```

Here's `UNION ALL` behaving as expected.

```
$ super -version
Version: v1.18.0-355-ga6eae509

$ super -c "
SELECT City,Country FROM customers.csv
UNION ALL
SELECT City,Country FROM suppliers.csv
ORDER BY City;"

{City:"Berlin",Country:"Germany"}
{City:"México D.F.",Country:"Mexico"}
{City:"México D.F.",Country:"Mexico"}
{City:"Londona",Country:"UK"}
{City:"New Orleans",Country:"USA"}
```

However, if we swap the order of the fields in the second `SELECT`, we see the relative order of the fields in the output remains the same as in the respective input tables, and the city values remain with the `City` field names and country values remain with `Country` field names.

```
$ super -c "
SELECT City,Country FROM customers.csv
UNION ALL
SELECT Country,City FROM suppliers.csv
ORDER BY City;"

{City:"Berlin",Country:"Germany"}
{City:"México D.F.",Country:"Mexico"}
{City:"México D.F.",Country:"Mexico"}
{Country:"UK",City:"Londona"}
{Country:"USA",City:"New Orleans"}
```

By comparison, in a tool like DuckDB that follows traditional SQL behavior, the names in the output table are based on the first input table while the values land in relative column order, such that the country values from the second `SELECT` are shown under the column header for `City`.

```
$ duckdb -c "
SELECT City,Country FROM customers.csv
UNION ALL
SELECT Country,City FROM suppliers.csv
ORDER BY City;"
┌─────────────┬─────────────┐
│    City     │   Country   │
│   varchar   │   varchar   │
├─────────────┼─────────────┤
│ Berlin      │ Germany     │
│ México D.F. │ Mexico      │
│ México D.F. │ Mexico      │
│ UK          │ Londona     │
│ USA         │ New Orleans │
└─────────────┴─────────────┘
```

Putting aside any debate about which behavior might be more intuitive, since our goal is for SuperSQL to be as Postgres-compatible as possible, we'll want to follow this behavior by default in SuperSQL.

While the following doesn't provide the behavior that matches traditional SQL, it should be noted that tacking on the [`fuse` operator](https://superdb.org/docs/language/operators/fuse) does adjust the relative order of SuperDB's output to appear columnar, though the city and country values remain under their original field names.

```
$ super -c "
SELECT City,Country FROM customers.csv
UNION ALL
SELECT Country,City FROM suppliers.csv
ORDER BY City
| fuse"

{City:"Berlin",Country:"Germany"}
{City:"México D.F.",Country:"Mexico"}
{City:"México D.F.",Country:"Mexico"}
{City:"Londona",Country:"UK"}
{City:"New Orleans",Country:"USA"}
```

### LINKED ISSUE #6342: SQL: Schema info not present after UNION ALL

### LINKED ISSUE #6369: SQL: Parse error in query containing cross join in subquery (fails at 0ae3bcf)
## tl;dr

This SQL query now causes a parse error, but previously parsed ok and produced the expected result.

```
$ super -c "SELECT + + 0 AS col1 FROM ( tab2 cor0 CROSS JOIN tab1 AS cor1 );"
parse error at line 1, column 63:
SELECT + + 0 AS col1 FROM ( tab2 cor0 CROSS JOIN tab1 AS cor1 );
                                                          === ^ ===
```

## Details

Repro is with super commit 0ae3bcf, which is associated with the merge of the changes in #6352. This issue was found via [a sqllogictest](https://github.com/dolthub/sqllogictest/blob/a1f4aa3c1d691552a4a03d9603c45bcf93d0bb9d/test/random/aggregates/slt_good_86.test#L64118-L64121).

The query was working at at commit d84d656, which was right before the merge of the changes in #6352. The test data needed for this repro is available if you have a local checkout of the [sqllogic-ztests](https://github.com/brimdata/sqllogic-ztests) repo and `cd` to the appropriate test directory.

```
$ cd ztests/sqlite/random/aggregates/slt_good_86/ztests-1

$ super -version
Version: d84d65640

$ super -c "SELECT + + 0 AS col1 FROM ( tab2 cor0 CROSS JOIN tab1 AS cor1 );"
{col1:0}
{col1:0}
{col1:0}
{col1:0}
{col1:0}
{col1:0}
{col1:0}
{col1:0}
{col1:0}
```

mattnibs:
This commit changes the behavior of SQL set operations in that it requires both the left and right side to have static schemas and for both sides to have an equal number of columns otherwise errors are returned. The right side record fields are renamed to that of the left side if the column names are not the same. A valid schema is returned so that two set operations can be joined.

This commit also fixes an union with SQLUnion where the parsed AST Loc was not set and adds the Node interface to ast.SQLQueryBody so errors can be reported on these elements.

Fixes #6342
Fixes #5755
Fixes #6369

### DIFF
--- a/compiler/semantic/sql.go
+++ b/compiler/semantic/sql.go
@@ -6,6 +6,7 @@ import (
 	"maps"
 	"reflect"
 	"slices"
+	"strconv"
 	"strings"
 
 	"github.com/brimdata/super"
@@ -478,9 +479,38 @@ func (t *translator) sqlQueryBody(query ast.SQLQueryBody, seq sem.Seq) (sem.Seq,
 		return seq, sch
 	case *ast.SQLUnion:
 		left, leftSch := t.sqlQueryBody(query.Left, seq)
-		left, _ = endScope(query.Left.(ast.Node), leftSch, left)
+		left, leftSch = endScope(query.Left.(ast.Node), leftSch, left)
+		leftCols, lok := leftSch.outColumns()
+		if !lok {
+			t.error(query.Left, errors.New("set operations cannot be applied to dynamic sources"))
+		}
 		right, rightSch := t.sqlQueryBody(query.Right, seq)
-		right, _ = endScope(query.Right.(ast.Node), rightSch, right)
+		right, rightSch = endScope(query.Right.(ast.Node), rightSch, right)
+		rightCols, rok := rightSch.outColumns()
+		if !rok {
+			t.error(query.Right, errors.New("set operations cannot be applied to dynamic sources"))
+		}
+		if !lok || !rok {
+			return sem.Seq{badOp()}, badSchema()
+		}
+		if len(leftCols) != len(rightCols) {
+			t.error(query, errors.New("set operations can only be applied to sources with the same number of columns"))
+			return sem.Seq{badOp()}, badSchema()
+		}
+		if !slices.Equal(leftCols, rightCols) {
+			// Rename fields on the right to match the left.
+			var elems []sem.RecordElem
+			for i, col := range leftCols {
+				elems = append(elems, &sem.FieldElem{
+					Name: col,
+					Value: &sem.IndexExpr{
+						Expr:  sem.NewThis(nil, nil),
+						Index: &sem.LiteralExpr{Value: strconv.Itoa(i)},
+					},
+				})
+			}
+			right = append(right, sem.NewValues(nil, &sem.RecordExpr{Elems: elems}))
+		}
 		out := sem.Seq{
 			&sem.ForkOp{Node: query, Paths: []sem.Seq{left, right}},
 			// This used to be dag.Combine but we don't have combine in the sem tree,
@@ -492,7 +522,7 @@ func (t *translator) sqlQueryBody(query ast.SQLQueryBody, seq sem.Seq) (sem.Seq,
 		if query.Distinct {
 			out = t.genDistinct(sem.NewThis(query, nil), out)
 		}
-		return out, &dynamicSchema{}
+		return out, leftSch
 	default:
 		panic(query)
 	}
--- a/compiler/ast/sql.go
+++ b/compiler/ast/sql.go
@@ -1,6 +1,7 @@
 package ast
 
 type SQLQueryBody interface {
+	Node
 	sqlQueryBodyNode()
 }
 
--- a/compiler/parser/parser.go
+++ b/compiler/parser/parser.go
@@ -4591,7 +4591,7 @@ var g = &grammar{
 					},
 				},
 			},
-			leader:        true,
+			leader:        false,
 			leftRecursive: true,
 		},
 		{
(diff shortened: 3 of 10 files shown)


compiler/semantic/sql.go
```diff
@@ -478,9 +478,40 @@ func (t *translator) sqlQueryBody(query ast.SQLQueryBody, seq sem.Seq) (sem.Seq,
...
+		if !lok {
+			t.error(query.Left, errors.New("set operations cannot be applied to dynamic sources"))
+		}
 		right, rightSch := t.sqlQueryBody(query.Right, seq)
-		right, _ = endScope(query.Right.(ast.Node), rightSch, right)
+		right, rightSch = endScope(query.Right.(ast.Node), rightSch, right)
+		rightStatic, rok := rightSch.(*staticSchema)
+		if !rok {
+			t.error(query.Right, errors.New("set operations cannot be applied to dynamic sources"))
+		}
```
nwt: Shouldn't this work for any schema that knows its output columns? Any schema implementations that wraps another schema might know that (depending on what it wraps).

mattnibs: My thinking here was that after the `endScope` call the `schema` should be either static or dynamic- I had tested this with selects but not with joins schemas. Turns out this is correct but there's an issue with select/joins on a query like this:
```
select *
from (values {x:1})
inner join (values {y:1}) on x = y
union all
select * from (values {y:2,z:2})
```
that has the left side returning dynamic and so we errantly get the error message:
```
set operations cannot be applied to dynamic sources
```

I'm going to investigate why this is happening.

mattnibs: @nwt I fixed this issue here https://github.com/brimdata/super/pull/6360 will rebase this branch on top of this. Used your idea of adding `getColumns` to the schema interface.

compiler/semantic/sql.go
```diff
@@ -478,9 +478,40 @@ func (t *translator) sqlQueryBody(query ast.SQLQueryBody, seq sem.Seq) (sem.Seq,
...
+		if !lok || !rok {
+			return sem.Seq{badOp()}, badSchema()
+		}
+		if len(leftCols) != len(rightCols) {
+			t.error(query, errors.New("set operations can only be applied to sources with the same number of columns"))
+			return sem.Seq{badOp()}, badSchema()
+		}
+		if !slices.Equal(leftCols, rightCols) {
+			// If columns are not equal rewrite the columns on the right side
+			// so they match that of the left.
```
nwt: Nit:

```suggestion
			// Rename fields on the right to match the left.
```

compiler/semantic/sql.go
```diff
@@ -478,9 +478,40 @@ func (t *translator) sqlQueryBody(query ast.SQLQueryBody, seq sem.Seq) (sem.Seq,
...
+			var elems []sem.RecordElem
+			for i, col := range leftCols {
+				idx := super.NewInt64(int64(i))
+				elems = append(elems, &sem.FieldElem{
+					Name: col,
+					Value: &sem.IndexExpr{
+						Expr:  sem.NewThis(nil, nil),
+						Index: &sem.LiteralExpr{Value: sup.FormatValue(idx)},
```
nwt: Nit:

```suggestion
						Index: &sem.LiteralExpr{Value: strconv.Itoa(i)},
```

compiler/semantic/sql.go
```diff
@@ -478,9 +478,40 @@ func (t *translator) sqlQueryBody(query ast.SQLQueryBody, seq sem.Seq) (sem.Seq,
...
+		if !lok {
+			t.error(query.Left, errors.New("set operations cannot be applied to dynamic sources"))
+		}
 		right, rightSch := t.sqlQueryBody(query.Right, seq)
-		right, _ = endScope(query.Right.(ast.Node), rightSch, right)
+		right, rightSch = endScope(query.Right.(ast.Node), rightSch, right)
+		rightCols, rok := rightSch.outColumns()
+		if !rok {
+			t.error(query.Right, errors.New("set operations cannot be applied to dynamic sources"))
+		}
```
nwt, line 492: Add test coverage for this.

nwt (approved):

compiler/semantic/sql.go
mattnibs: I wasn't quite sure how to test for a dynamic source that won't have to be deleted soon. Like I could query against a jsup file but aren't we planning to read types of small jsup files?

philrz: Not sure if this alone is motivation enough to do it now, but in a recent group discussion I think @mccanne proposed adding a `-dynamic` flag to `super` that explicitly bypassed type checking.

mattnibs: This sounds like its just skipping the type check which is separate from this part but I suppose I'd like to hear more from @mccanne about how he envisions this feature to work.

nwt: > I wasn't quite sure how to test for a dynamic source that won't have to be deleted soon. Like I could query against a jsup file but aren't we planning to read types of small jsup files?

Let's cross that bridge when we come to it.

PR #6360, opened 2025-11-13
Propogate SQL static schemas on join queries

mattnibs:
This commit fixes an issue with the semantic pass for SQL queries where join queries with static schemas would end up getting reported as dynamic schemas. This would cause certain queries to lose out on potential optimizations as well as inhibit future functionality that relies on knowing if child schemas are static or not.

### DIFF
--- a/compiler/semantic/schema.go
+++ b/compiler/semantic/schema.go
@@ -483,9 +483,8 @@ func (j *joinSchema) outColumns() ([]string, bool) {
 	if !lok || !rok {
 		return nil, false
 	}
-	// Append all columns from the left except for those that overlap with
-	// the right as these will be overwritten. We/If we allow duplicate columns
-	// this should be changed to just concat both columns.
+	// Include all columns from the left except for those appearing on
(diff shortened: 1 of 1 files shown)


compiler/semantic/schema.go
```diff
@@ -460,6 +461,48 @@ func (s *subquerySchema) tableOnly(n ast.Node, name string, path []string) (sem.
...
+
+func (j *joinSchema) outColumns() ([]string, bool) {
+	left, lok := j.left.outColumns()
+	right, rok := j.right.outColumns()
+	if !lok || !rok {
+		return nil, false
+	}
+	// Append all columns from the left except for those that overlap with
+	// the right as these will be overwritten. We/If we allow duplicate columns
+	// this should be changed to just concat both columns.
```
nwt: "Append all columns from the left" is not an accurate description of what you're doing here.

Maybe this?

```suggestion
	// Include all columns from the left except for those appearing on 
	// the right since that's what {...left,...right} does.
```

PR #6366, opened 2025-11-17
remove output-check Makefile target and CI job

nwt:
They haven't caught a problem in a long time.


philrz:
FWIW, #1065 was when this was originally added.

To summarize, it originated in time when we were frequently changing the data formats and wanted a way to know if we'd made a change by accident that would affect users that had started saving data in our formats. Since it read/wrote values of most types in all supported formats, over time it also found a couple panics specific to input/output of certain values/formats. However, as noted above by @nwt it hadn't found any problems in a while, plus there was overhead related to making its test keep passing after it started "failing" due to benign changes. Meanwhile we have a lot more ztests that provide test coverage, and with few exceptions our data formats have been holding still. It was a fine time to retire it.

PR #6370, opened 2025-11-17
change GET /query/describe to send aggregate input keys

nwt:
The GET /query/describe route sends a response containing the output keys for aggregation operations but Zui is interested in the input keys, so send those instead.

This wasn't a problem for Zui for most queries until #6263, which changed the output key for "aggregate by a.b" from a.b (the path of the input key) to b (the last path component of the input key).

Fixes brimdata/zui#3210.


philrz (approved):
Normally I hold back on clicking Approve on such PRs, but I the fix seems simple and I just confirmed it's effective. I'm not offended if you want to wait for a second approval from a proper Go developer, though.

PR #6373, opened 2025-11-19
Fix a few broken links

philrz:
Now that the book docs are deployed at https://superdb.org/ and the old SuperDB site (which included the old docs) is gone, the link checker picked up on some absolute URLs that were hiding in hyperlinks that are now broken. Those are fixed here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
-SuperDB's first GA release is expected in the Summer of 2025. Ongoing changes
+SuperDB's first GA release is expected in the 2026. Ongoing changes
 will be documented here once that release is made available. To try out
 SuperDB while it's still under development we recommend starting from the
-[`super` command doc](https://superdb.org/docs/commands/super/).
+[`super` command doc](https://superdb.org/command/super.html).
 
 The SuperDB codebase originates from the [Zed project](https://zed.brimdata.io/)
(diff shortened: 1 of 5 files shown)


book/src/command/db-auth.md
```diff
@@ -18,8 +18,3 @@ Additional options of the [db sub-command](db.md#options)
 
 > **TODO: rename this command. it's really about connecting to a database.
 > authenticating is something you do to connect.**
-
-Access to a lake can be secured with [Auth0 authentication](https://auth0.com/).
-A [guide](../dev/integrations/auth.md) is available with example configurations.
-Please reach out to us on our [community Slack](https://www.brimdata.io/join-slack/)
-if you have feedback on your experience or need additional help.
```
philrz, line 25: @mccanne: I saw that the article this was linking to had a `!!` prepended to it in the `SUMMARY.md`, so it's not currently alive to link to:

https://github.com/brimdata/super/blob/2bfb13b989ae948309b678ea95776d8ca719482b/book/src/SUMMARY.md?plain=1#L216

I took that to mean you were intentionally disabling it and hence figured I'd drop this paragraph rather than leave orphan text describing an absent article. Let me know if I misunderstood your intent though.

mccanne (approved):

PR #6380, opened 2025-11-20
Add colored backgrounds for mdbook admonitions

philrz:
## tl;dr

This enables colored backgrounds for mdbook [admonitions](https://rust-lang.github.io/mdBook/format/markdown.html#admonitions).

## Details

This is a handy feature we can take advantage of now that we're running mdbook v0.5 which improves on the simple blockquotes we've been doing:

https://github.com/brimdata/super/blob/a3c31a5470d8d2a89a9427397dd15ed803eb96cf/book/src/intro.md?plain=1#L386-L387

Now can become:

image

I happen to be partial to the colored backgrounds like we had with Docusaurus admonitions, and the CSS I've added here gets us that. I mostly reach for "Tip" and "Note", in my writing, but here's the full set:

image

### DIFF
--- a/book/admonitions/admonitions.css
+++ b/book/admonitions/admonitions.css
@@ -1,0 +1,25 @@
+/* Add light background colors to mdbook admonitions */
+.blockquote-tag {
+  padding: 1em;
+  border-radius: 4px;
+}
+
+.blockquote-tag-note {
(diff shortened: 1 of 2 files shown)


mattnibs (approved):
Not a css experiment but this lgtm

PR #6387, opened 2025-11-24
aggfunc.subst: Use original ast.Node for sem.This.Node

mattnibs:
This commit fixes a bug where the Node for the newly created sem.This object was being set as a sem.Expr (which implements ast.Node) where it should be the original ast.Node. This change will be tested in an upcoming pr.


mattnibs:
Closing this not needed.

PR #6388, opened 2025-11-25
Allow group by keys to be used in arbitrary expressions

### LINKED ISSUE #5973: Using GROUP BY column in math
## tl;dr

The following query works in Postgres but not SuperDB.

```
SELECT i, i + 10 FROM integers GROUP BY i ORDER BY i;
```

## Details

Repro is with super commit 3c0e543. This was found via a query from [a sqllogictest](https://github.com/duckdb/duckdb/blob/0cbb56d14cc1464d945a06cb8bac051485fa2ef7/test/sql/aggregate/group/test_group_by.test#L113-L118).

Repro with test data:

```
$ super -version
Version: 3c0e54399

$ cat integers.sup 
{i:3(int32),j:4(int32)}
{i:3(int32),j:4(int32)}
{i:2(int32),j:4(int32)}

$ super -c "
SELECT i, i + 10 FROM 'integers.sup' GROUP BY i ORDER BY i;"

no corresponding grouping element for non-aggregate "i+10" at line 2, column 11:
SELECT i, i + 10 FROM 'integers.sup' GROUP BY i ORDER BY i;
```

Here it is working in Postgres:

```
$ psql postgres
psql (17.5 (Homebrew))
Type "help" for help.

$ psql postgres
psql (17.5 (Homebrew))
Type "help" for help.

postgres=# SELECT * FROM integers;
 i | j 
---+---
 3 | 4
 3 | 4
 2 | 4
(3 rows)

postgres=# SELECT i, i + 10 FROM integers GROUP BY i ORDER BY i;
 i | ?column? 
---+----------
 2 |       12
 3 |       13
(2 rows)
```

mattnibs:
Previously group by keys could only be used in select clauses if the select expression exactly matched the expression in a group by. This change allows them to be used in any abitrary expression including expressions that involve multiple group-by keys.

Closes #5973

### DIFF
--- a/compiler/semantic/projection.go
+++ b/compiler/semantic/projection.go
@@ -157,3 +157,122 @@ func (a *aggfuncs) substArrayElems(elems []sem.ArrayElem) []sem.ArrayElem {
 	}
 	return out
 }
+
+func keySubst(e sem.Expr, exprs []exprloc) (sem.Expr, bool) {
+	if i := exprMatch(e, exprs); i >= 0 {
+		return sem.NewThis(e, []string{"in", fmt.Sprintf("k%d", i)}), true
+	}
+	ok := true
+	switch e := e.(type) {
+	case nil:
+	case *sem.AggFunc:
+		// This shouldn't happen.
+		panic(e)
+	case *sem.ArrayExpr:
+		e.Elems, ok = keySubstArrayElems(e.Elems, exprs)
+	case *sem.BinaryExpr:
+		if e.LHS, ok = keySubst(e.LHS, exprs); ok {
+			e.RHS, ok = keySubst(e.RHS, exprs)
+		}
+	case *sem.CallExpr:
+		var args []sem.Expr
+		for _, arg := range e.Args {
+			if arg, ok = keySubst(arg, exprs); !ok {
+				return nil, false
+			}
+			args = append(args, arg)
+		}
+		e.Args = args
+	case *sem.CondExpr:
+		var ok1, ok2, ok3 bool
+		e.Cond, ok1 = keySubst(e.Cond, exprs)
+		e.Then, ok2 = keySubst(e.Then, exprs)
+		e.Else, ok3 = keySubst(e.Else, exprs)
+		ok = ok1 && ok2 && ok3
+	case *sem.DotExpr:
+		e.LHS, ok = keySubst(e.LHS, exprs)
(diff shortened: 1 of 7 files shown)


compiler/semantic/sem/expr.go
```diff
@@ -246,3 +248,201 @@ func NewStructuredError(n ast.Node, message string, on Expr) Expr {
...
+				Expr: CopyExpr(elem.Expr),
+			})
+		default:
+			panic(elem)
+		}
+	}
+	return out
+}
+
+func copyNode(n ast.Node) ast.Node {
```
nwt: @mccanne is right: You don't need #6387 if you remove this function and just use the incoming ast.Node instead of copying it.

compiler/semantic/projection.go
```diff
@@ -157,3 +157,122 @@ func (a *aggfuncs) substArrayElems(elems []sem.ArrayElem) []sem.ArrayElem {
 	}
 	return out
 }
+
+func keySubst(e sem.Expr, exprs []exprloc) (sem.Expr, bool) {
+	if i := exprMatch(e, exprs); i >= 0 {
+		return sem.NewThis(ast.NewLoc(e.Pos(), e.End()), []string{"in", fmt.Sprintf("k%d", i)}), true
```
nwt: Nit:

```suggestion
		return sem.NewThis(e, []string{"in", fmt.Sprintf("k%d", i)}), true
```

compiler/semantic/sql.go
```diff
@@ -229,31 +229,32 @@ func (t *translator) genAggregateOutput(loc ast.Node, proj projection, keyExprs
...
-			// grouping expression.
+		key, ok := keySubst(sem.CopyExpr(col.expr), keyExprs)
+		if !ok {
 			alias := sem.NewThis(col.loc, []string{"out", col.name})
-			which = exprMatch(alias, keyExprs)
+			which := exprMatch(alias, keyExprs)
+			if ok = which >= 0; ok {
+				key = sem.NewThis(col.loc, []string{"in", fmt.Sprintf("k%d", which)})
+				ok = true
+			}
```
```suggestion
			if which := exprMatch(alias, keyExprs); which >= 0 {
				key = sem.NewThis(col.loc, []string{"in", fmt.Sprintf("k%d", which)})
				ok = true
			}
```

nwt (approved):

PR #6396, opened 2025-11-26
Edits to "intro" docs at top-level and super-sql

philrz:
See inline comments for edits that require additional explanation.

### DIFF
--- a/book/src/super-sql/intro.md
+++ b/book/src/super-sql/intro.md
@@ -309,7 +309,7 @@ produces
 {x:2,y:5}
 {x:3,y:5}
 ```
-In contrast, relational scoping using [identifier scoping](https://en.wikipedia.org/wiki/Scope_(computer_programming)) in a `SELECT` clause
+In contrast, [relational scoping](#relational-scoping) in a `SELECT` clause
 with the table source identified in `FROM` and `JOIN` clauses, e.g., this query
 produces the same result:
 ```
@@ -394,7 +394,7 @@ type checking, instead creating runtime errors when type errors are encountered.
 
 In other words, dynamic data is statically type checked when possible.
 This works by computing fused types of each operator's output and propagating
-these types in a [dataflow](https://en.wikipedia.org/wiki/Dataflow) analysis.  When types are unknown, the
+these types in a [dataflow analysis](https://en.wikipedia.org/wiki/Data-flow_analysis).  When types are unknown, the
 analysis flexibly models them as having any possible type.
 
 For example, this query produces the expected output


book/README.md
```diff
@@ -47,35 +47,42 @@ The following should be rendered in `fixed width` font:
 * References to operator/function usage (e.g., "the `<expr>` argument to `eval`)"
 * Input/output/parameter values (e.g., "the value of N defaults to `1`)
 
-### Pipeline elements
+> **Exception**: Don't use fixed width when hyperlinking, since rendered
+  hyperlinks already have color/underline emphasis.
```
philrz, line 51: This is a change from a prior convention as first touched on [here](https://github.com/brimdata/super/pull/6363#discussion_r2528547786) then discussed offline with @mccanne. Hyperlinks within this branch have been adjusted to fit this new convention.

book/src/intro.md
```diff
@@ -182,8 +182,7 @@ DataFusion CLI v46.0.1
 Error: Arrow error: Cast error: Cannot cast string 'foo' to value of Int64 type
 ```
 
-The more recent innovation of an open
-[variant type](https://github.com/apache/spark/blob/master/common/variant/README.md)
```
philrz, line 186: I dropped the hyperlink here since I added the hyperlink to the first reference to "variant" within this section of text.

book/src/intro.md
```diff
@@ -284,28 +283,33 @@ SELECT avg(radius) as R, avg(width) as W FROM (
 {R:1.75,W:1.5}
 ```
 
-Things get more interesting when you want to do different types of processing
+Things get more interesting when you want to do different kinds of processing
```
philrz, line 286: I figured this synonym was a slight improvement since the word "type" is already used in the technical sense close to here.

book/src/intro.md
```diff
@@ -284,28 +283,33 @@ SELECT avg(radius) as R, avg(width) as W FROM (
...
 | switch kind
     case 'circle' (
         R:=avg(radius)
     )
     case 'rect' (
         width:=width*2
     )
+    default (
+        error({message:'unknown shape',value:this})
+    )
```
philrz, line 307: I added this along with the "triangle" record above in attempt to make good on the "...and employ first-class errors to flag unknown types" statement that was already present in the text. Or if the original/shorter example is preferred, we could drop that part of the text.

book/src/super-sql/intro.md
```diff
@@ -202,10 +201,10 @@ Here, a pipe operator takes any sequence of input values
 and produces any computed sequence of output values and _all
 data references are limited to these inputs and outputs_.
 Since there is just one sequence of values, it may be
-referenced as special value with a special name, which for
+referenced as a special value with a special name, which for
 SuperSQL is the value `this`.
 
-This scoping model can be summarized as follows:
+The pipe scoping model can be summarized as follows:
```
philrz, line 207: Made this adjustment just to avoid using the word "this" again, since it appears so many times close to here when talking about special value `this`.

book/src/super-sql/intro.md
```diff
@@ -310,7 +309,7 @@ produces
 {x:2,y:5}
 {x:3,y:5}
 ```
-In contrast, relational scoping using identifier scoping in a `SELECT` clause
+In contrast, relational scoping using [identifier scoping](https://en.wikipedia.org/wiki/Scope_(computer_programming)) in a `SELECT` clause
```
philrz: While this is all happening within a doc that talks a lot about "scoping", since we just defined specific examples like "pipe scoping", "relational scoping", and are in the middle of describing "join scoping", a term like "identifier scoping" read to me like a specific thing that justified a definition or a link. That said, it seems like most resources I found online talk about it in the context of one particular programming language, so I went with the Wikipedia article.

book/src/super-sql/intro.md
```diff
@@ -339,7 +338,7 @@ collection of unnested values separately as a unit for each outer value.
 The `into ( <scope> )` body is an optional component of `unnest`, and if absent,
 the unnested collection boundaries are ignored and all of the unnested data is output.
 
-With the `unnest` operator, we can now consider how a correlated subquery from
+With the `unnest` operator, we can now consider how a [correlated subquery](https://en.wikipedia.org/wiki/Correlated_subquery) from
```
philrz, line 341: There's problems with the correlated subquery example below that I'll discuss offline.

book/src/super-sql/intro.md
```diff
@@ -440,21 +439,27 @@ appear in the file.  Likewise, data stored in a database organized by
...
-Each operator defines whether or not it is order is preserved,
-created, or discarded.
+The top of the documentation page for each [operator](operators/intro.md)
+is marked with an icon to describe its data order behavior:
+
+|Marker|Data Order behavior    |
+|:----:|:----------------------|
+| ✅   | Order-preserving      |
+| 🔀   | Order-creating        |
+| 🎲   | Undefined output order|
```
philrz, line 453: We've discussed at times that it'd be good to explicitly disclose the order behaviors of the various operators. Since the text I'm replacing here seemed to be nodding toward that concept, I've tried to bite that off here. Since it's a first take, I'm open to input on the approach, e.g., whether folks like the "icon" approach, choice of icons, now they're presented in the individual operator pages, etc.

This first approach I'm proposing is trying to be brief (i.e., just a hyperlinked icon at the top of each operator page, as opposed to a whole line of text) but still catch the reader's attention (they seem to stand out such that a user reading a page for the first time may hover over them, see the link, then learn about what it means, which is what we want).

In addition to reviewing for style, please double check that I've not misrepresented the order behavior of any given operator!

book/src/super-sql/operators/from.md
```diff
@@ -1,6 +1,6 @@
 ### Operator
 
-&emsp; **from** &mdash; source data from databases, files, or URLs
+[✅](../intro.md#data-order)[🎲](../intro.md#data-order)&emsp; **from** &mdash; source data from databases, files, or URLs
```
philrz, line 3: This was one where I was iffy on how to best decorate. As the `from` doc itself describes, the "undefined order" concern only comes up when there's multiple sources. So if I was being conservative, I'd show just the 🎲, and if I'm being generous and speaking of the usage that's probably more commonly invoked, I'd go with just the ✅. Instead I'm currently opting to show both, and if a reader is puzzled as to why there's two icons, hopefully that would inspire them to read the text and appreciate the subtlety.

book/src/super-sql/intro.md
mccanne: identifier scoping doesn't read right.  How about just "relational scoping" with the link to the super-sql/intro.md#relational....

book/src/super-sql/intro.md
```diff
@@ -395,17 +394,17 @@ type checking, instead creating runtime errors when type errors are encountered.
 
 In other words, dynamic data is statically type checked when possible.
 This works by computing fused types of each operator's output and propagating
-these types in a dataflow analysis.  When types are unknown, the
+these types in a [dataflow](https://en.wikipedia.org/wiki/Dataflow) analysis.  When types are unknown, the
```
mccanne: wrong link... try https://en.wikipedia.org/wiki/Data-flow_analysis

mccanne (approved):
see comments but looks great

PR #6405, opened 2025-11-30
cleanly separate pipe from and SQL FROM

mccanne:
This commit cleanly separates the syntax of the pipe-operator "from" and the SQL FROM clause.  This means that table aliases are no longer allowed with pipe from (even though they were ignored).  Also, a relational FROM is no longer possible without a SELECT clause (required a "SELECT *" to do a straight relational join).  We can revisit this if there's interest as the grammar should be able to handle it.

This better aligns the PEG grammar and semantic pass with the actual SQL syntax and will make it easier to document the FROM clause and the "from" pipe operator.

Documentation is forthcoming from the book-sql branch.

### DIFF
--- a/compiler/ast/sql.go
+++ b/compiler/ast/sql.go
@@ -83,12 +83,12 @@ type SQLTableInput interface {
 func (*FromItem) sqlTableInputNode() {}
 func (*SQLPipe) sqlTableInputNode()  {}
 
-// SQL table expression structure all of which implement FromEntity
-
 type SQLTableExpr interface {
 	Node
 	sqlTableExprNode()
 }
+
+// SQL table expression structure all of which implement SQLTableExpr
 
 type (
 	SQLCrossJoin struct {
--- a/compiler/parser/parser.peg
+++ b/compiler/parser/parser.peg
@@ -752,14 +752,11 @@ SQLTableExpr
 
 FromItem
   = source:FromSource args:CommitishOpArgs? {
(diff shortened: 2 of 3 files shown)


compiler/ast/sql.go
```diff
@@ -74,25 +74,46 @@ type (
...
+// SQLTableInput is a table expression that can be aliased with an AS clause.
+type SQLTableInput interface {
+	Node
+	sqlTableInputNode()
+}
+
+func (*FromItem) sqlTableInputNode() {}
+func (*SQLPipe) sqlTableInputNode()  {}
+
 // SQL table expression structure all of which implement FromEntity
```
nwt: Looks like this is now out of date.

compiler/parser/parser.peg
```diff
@@ -705,44 +705,70 @@ DebugOp
...
+FromItem
+  = source:FromSource args:CommitishOpArgs? {
+      s := &ast.FromItem{
+        Source: source.(ast.FromSource),
+        Loc: loc(c),
+      }
+      if args != nil {
+        s.Args = sliceOf[ast.OpArg](args)
+      }
+      return s, nil
```
nwt: Nit: sliceOf returns nil for a nil argument.

```suggestion
      return &ast.FromItem{
        Source: source.(ast.FromSource),
        Args: sliceOf[ast.OpArg](args),
        Loc: loc(c),
      }, nil
```

PR #6411, opened 2025-12-02
coalesce: ignore all errors

### LINKED ISSUE #4329: coalesce() should ignore all errors
coalesce() returns the first value that is not null, error("missing"), or error("quiet"). Based on user feedback, it would be more useful to ignore all errors and return the first value that is not null or an error.

mattnibs:
This commit changes the behavior of the coalesce function so that all errors are ignored not just error("missing") and error("quiet").

Fixes #4329


chrismo:
🥳 🎉

PR #6412, opened 2025-12-02
Handle no else clause on case expressions

### LINKED ISSUE #6333: "case" without "else"
## tl;dr

As a user, I recently tried using `case` expressions that didn't include an `else` portion and bumped into some problems. Maybe these should be forked off to different issues, but since they're related I'll start by showing them together.

## Details

Repro is with super commit bec8934.

### 1. `error("missing")` when `else` is absent

As a starting point, I'll use this example that _does_ have an `else` portion.

```
$ super -version
Version: bec8934e0

$ echo '"one" "two" "three"' | super -c "
values upper(this)
| values case
    when this=='ONE' then 1
    when this=='TWO' then 2
    else -1
  end" -

1
2
-1
```

Dropping the `else` portion, we get `error("missing")` for the input value that would have matched our `else`.

```
$ echo '"one" "two" "three"' | super -c "
values upper(this)
| values case
    when this=='ONE' then 1
    when this=='TWO' then 2
  end" -

1
2
error("missing")
```

As someone that knows the tooling, I sort of understand why `error("missing")` might show up here, but would it make sense to have an error that more explicitly describes its origin, e.g, `error({message:"unmatched case expression",on:"three"})`? Perhaps this is just an example of #4608.

### 2. `null expression not allowed`

Now I'll switch the example with `else` to have the call to `upper` as part of the `case`.

```
$ echo '"one" "two" "three"' | super -c "
values case upper(this)
  when 'ONE' then 1
  when 'TWO' then 2
  else -1
end" -

1
2
-1
```

When we drop the `else`:

```
$ echo '"one" "two" "three"' | super -c "
values case upper(this)
  when 'ONE' then 1
  when 'TWO' then 2
end" -   

null expression not allowed
```

This one is not very helpful in its current form since in addition to being vague it doesn't point to what part of the query caused the failure.

### 3. In SQL, unmatched `CASE` produces `NULL` values

In Postgres:

```
$ psql postgres
psql (17.6 (Homebrew))
Type "help" for help.

postgres=# \pset null 'NULL'
Null display is "NULL".
postgres=# CREATE TABLE data (a VARCHAR);
CREATE TABLE
postgres=# INSERT INTO data (a) VALUES ('one'), ('two'), ('three');
INSERT 0 3

postgres=# SELECT                       
  CASE
    WHEN a='one' THEN 1    
    WHEN a='two' THEN 2    
  END
FROM data;
 case 
------
    1
    2
 NULL
(3 rows)
```

Whereas in `super`, we once again see `error("missing")`.

```
$ super -f parquet -o data.parquet -c "values {a:'one'},{a:'two'},{a:'three'}" && super -c "
SELECT                       
  CASE
    WHEN a='one' THEN 1    
    WHEN a='two' THEN 2
  END AS case  
FROM data.parquet;"

{case:1}
{case:2}
{case:error("missing")}
```

This seems similar in spirit to #5984, i.e., while we feel strongly about `error("missing")` being an improvement over using `NULL` for so many things the way SQL does, users bringing over their legacy SQL queries likely expect to see `NULL` values in results and have those match against `NULL`-centric filtering in those queries.

mattnibs:
The commit changes case expressions so that in a pipe query if an expression does not match any clause and there is no else clause a structured error is returned. In a SQL query a null value is returned.

Closes #6333

### DIFF
--- a/compiler/semantic/expr.go
+++ b/compiler/semantic/expr.go
@@ -613,7 +613,14 @@ func (t *translator) semDotted(e *ast.BinaryExpr, lval bool) ([]string, sem.Expr
 
 func (t *translator) semCaseExpr(c *ast.CaseExpr) sem.Expr {
 	e := t.expr(c.Expr)
-	out := t.exprNullable(c.Else)
+	var out sem.Expr
+	if c.Else != nil {
+		out = t.expr(c.Else)
+	} else if t.scope.schema != nil {
+		out = &sem.LiteralExpr{Node: c, Value: "null"}
+	} else {
+		out = sem.NewStructuredError(c, "case: no clause matched and no else provided", e)
+	}
 	for i := len(c.Whens) - 1; i >= 0; i-- {
 		when := c.Whens[i]
 		out = &sem.CondExpr{
--- a/runtime/ztests/expr/case-noelse.yaml
+++ b/runtime/ztests/expr/case-noelse.yaml
@@ -1,0 +1,8 @@
+spq: |
+  values 1,2 | values case this when 1 then 'foo' end
+
+vector: true
+
+output: |
+  "foo"
+  error({message:"case: no clause matched and no else provided",on:2})
--- a/compiler/ztests/sql/case-noelse.yaml
+++ b/compiler/ztests/sql/case-noelse.yaml
@@ -1,0 +1,8 @@
+spq: |
+  select case x when 1 then 'foo' end as x from (values (1),(2)) T(x)
+
+vector: true
+
+output: |
+  {x:"foo"}
+  {x:null}


runtime/ztests/expr/case-noelse.yaml
```diff
@@ -0,0 +1,13 @@
+script: |
+  super -s -c "values 1,2 | values case this when 1 then 'foo' end"
+  echo // ===
+  super -s -c "select case x when 1 then 'foo' end as x from (values (1),(2)) T(x)"
+
```
nwt, line 3:
```suggestion

vector: true

```

compiler/semantic/expr.go
```diff
@@ -613,7 +613,16 @@ func (t *translator) semDotted(e *ast.BinaryExpr, lval bool) ([]string, sem.Expr
...
+	var out sem.Expr
+	if c.Else != nil {
+		out = t.expr(c.Else)
+	} else {
+		if t.scope.schema != nil {
+			out = &sem.LiteralExpr{Node: c, Value: "null"}
+		} else {
+			out = sem.NewStructuredError(c, "case: no clause matched and no else provided", sem.CopyExpr(e))
+		}
+	}
```
nwt, line 623: Nits: Indent less. And there's no need to copy e.

```suggestion
	} else if t.scope.schema != nil {
		out = &sem.LiteralExpr{Node: c, Value: "null"}
	} else {
		out = sem.NewStructuredError(c, "case: no clause matched and no else provided", e)
	}
```

runtime/ztests/expr/case-noelse.yaml
```diff
@@ -0,0 +1,13 @@
+script: |
+  super -s -c "values 1,2 | values case this when 1 then 'foo' end"
+  echo // ===
+  super -s -c "select case x when 1 then 'foo' end as x from (values (1),(2)) T(x)"
```
nwt: I think tests for SQL behavior should go with rest of the SQL tests in compiler/ztests/sql.

runtime/ztests/expr/case-noelse.yaml
mattnibs: doh

nwt (approved):

chrismo:
@philrz with an older version this test case from the commit still returned a "error"-y message, yeah? so this isn't much of a breaking change, or is it?

```
❯ super --version                                               
Version: v0.0.0-20251016221528-bdb38bbc4fef

❯ super -c "values 1,2 | values case this when 1 then 'foo' end"
null expression not allowed
```

EDIT: Ah, I see more details -> https://github.com/brimdata/super/issues/6333

philrz:
@chrismo: Yeah, as you may have gathered after your find of the details, I'd frame this as an improvement rather than a breaking change. As I see it, that old `null expression not allowed` error had two faults:

1. It was vague
2. It stopped the query from running entirely

The new structured error only shows up when triggered by an input value, and since it's a first class `error` value it can be filtered/quieted and all the usual treatments.

```
$ super -version
Version: 7b4ab522e

$ echo '"one" "two" "three"' | super -c "
values case upper(this)
  when 'ONE' then 1
  when 'TWO' then 2
end" -
1
2
error({message:"case: no clause matched and no else provided",on:"THREE"})
```

As you may have also noticed, the ink isn't quite try on all the changes related to #6333, but I expect #6415 will merge soon.

PR #6415, opened 2025-12-02
case: Add no else error to non-match case expressions

### LINKED ISSUE #6333: "case" without "else"

mattnibs:
This commit follows up #6412 by adding the same error/null values added in that pr but to non-matching case expressions. Part of the solution is to use the same ast node (ast.CaseExpr) for both match and non-matching case expressions.

Closes #6333

### DIFF
--- a/compiler/semantic/expr.go
+++ b/compiler/semantic/expr.go
@@ -612,20 +612,26 @@ func (t *translator) semDotted(e *ast.BinaryExpr, lval bool) ([]string, sem.Expr
 }
 
 func (t *translator) semCaseExpr(c *ast.CaseExpr) sem.Expr {
-	e := t.expr(c.Expr)
+	e := t.exprNullable(c.Expr)
 	var out sem.Expr
 	if c.Else != nil {
 		out = t.expr(c.Else)
 	} else if t.scope.schema != nil {
 		out = &sem.LiteralExpr{Node: c, Value: "null"}
-	} else {
+	} else if e != nil {
 		out = sem.NewStructuredError(c, "case: no clause matched and no else provided", e)
+	} else {
+		out = sem.NewStringError(c, "case: no clause matched and no else provided")
 	}
 	for i := len(c.Whens) - 1; i >= 0; i-- {
 		when := c.Whens[i]
+		cond := t.expr(when.Cond)
+		if e != nil {
+			cond = sem.NewBinaryExpr(c, "==", e, cond)
+		}
 		out = &sem.CondExpr{
 			Node: c,
-			Cond: sem.NewBinaryExpr(c, "==", e, t.expr(when.Cond)),
+			Cond: cond,
 			Then: t.expr(when.Then),
 			Else: out,
 		}
--- a/compiler/semantic/sem/expr.go
+++ b/compiler/semantic/sem/expr.go
@@ -249,6 +249,16 @@ func NewStructuredError(n ast.Node, message string, on Expr) Expr {
 	}
 }
 
+func NewStringError(n ast.Node, message string) Expr {
+	return &CallExpr{
+		Node: n,
+		Tag:  "error",
+		Args: []Expr{
+			&LiteralExpr{Node: n, Value: sup.FormatValue(super.NewString(message))},
+		},
+	}
+}
+
 func CopyExpr(e Expr) Expr {
 	switch e := e.(type) {
 	case nil:
--- a/compiler/parser/parser.go
+++ b/compiler/parser/parser.go
@@ -4699,7 +4699,7 @@ var g = &grammar{
 					},
 				},
 			},
-			leader:        false,
+			leader:        true,
 			leftRecursive: true,
 		},
 		{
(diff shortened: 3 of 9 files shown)


compiler/semantic/sem/expr.go
```diff
@@ -249,6 +249,16 @@ func NewStructuredError(n ast.Node, message string, on Expr) Expr {
 	}
 }
 
+func NewStringError(n ast.Node, message string) Expr {
+	return &CallExpr{
+		Node: n,
+		Tag:  "error",
+		Args: []Expr{
+			&LiteralExpr{Node: n, Value: sup.QuotedString(message)},
```
nwt: Why is the LiteralExpr here different from the one in NewStructuredError?

compiler/semantic/expr.go
```diff
@@ -612,20 +612,26 @@ func (t *translator) semDotted(e *ast.BinaryExpr, lval bool) ([]string, sem.Expr
...
+	} else if e != nil {
 		out = sem.NewStructuredError(c, "case: no clause matched and no else provided", e)
+	} else {
+		out = sem.NewStringError(c, "case: no clause matched and no else provided")
 	}
 	for i := len(c.Whens) - 1; i >= 0; i-- {
 		when := c.Whens[i]
+		cond := t.expr(when.Cond)
+		if e != nil {
+			cond = sem.NewBinaryExpr(c, "==", e, t.expr(when.Cond))
```
```suggestion
			cond = sem.NewBinaryExpr(c, "==", e, cond)
```

compiler/semantic/sem/expr.go
mattnibs: What would be the value for `on` be in this circumstance since you don't have a single matching expression?

mattnibs: I suppose we could make `on` the value of `this`, you'll still have to do something different.

nwt: What I'm getting at is that the message parameter should be handled the same by both functions but isn't.

NewStrucutredError:

```go
&LiteralExpr{Node: n, Value: sup.FormatValue(super.NewString(message))}
```

NewStringError:
```go
&LiteralExpr{Node: n, Value: sup.QuotedString(message)}
```

mattnibs: right

mattnibs: apologies I totally read your original comment wrong.

PR #6417, opened 2025-12-03
Remove const optimizations in rungen.compileBinary

### LINKED ISSUE #6416: SQL: Incorrect result from CASE+COALESCE+AVG combo at commit 79545c1
## tl;dr

The following SQL query produces the result `1` in Postgres but `null` in SuperDB.

```
$ super -version
Version: 79545c1a3

$ super -c "SELECT CASE -71 WHEN COALESCE (-AVG(71), 2) THEN 1 ELSE NULL END AS col0;"
{col0:null}
```

## Details

Repro is with super commit 79545c1, which is associated with the merge of changes in #6411. This is a simplification of a query from [a sqllogictest](https://github.com/dolthub/sqllogictest/blob/a1f4aa3c1d691552a4a03d9603c45bcf93d0bb9d/test/random/expr/slt_good_116.test#L66100-L66105).

Here's the expected result as produced by Postgres.

```
$ psql postgres
Type "help" for help.

postgres=# SELECT CASE -71 WHEN COALESCE (-AVG(71), 2) THEN 1 ELSE NULL END AS col0;
 col0 
------
    1
(1 row)
```

It looks like the use of `COALESCE` is key here, as `super` returned the expected result at the commit right before the merge of #6411.

```
$ super -c "SELECT CASE -71 WHEN COALESCE (-AVG(71), 2) THEN 1 ELSE NULL END AS col0;"
{col0:1}
```

mattnibs:
This commit removes the compileConstIn and compileConstCompare optimizations from the rungen of the sequential runtime. These functions rely on evalAtCompileTime to determine if an expression is a const which is insufficient for this purpose: evalAtCompileTime can report a value is a const when it is not.

Closes #6416

### DIFF
--- a/compiler/rungen/expr.go
+++ b/compiler/rungen/expr.go
@@ -105,15 +105,6 @@ func (b *Builder) compileExprWithEmpty(e dag.Expr) (expr.Evaluator, error) {
 }
 
 func (b *Builder) compileBinary(e *dag.BinaryExpr) (expr.Evaluator, error) {
-	if e.Op == "in" {
-		// Do a faster comparison if the LHS is a compile-time constant expression.
-		if in, err := b.compileConstIn(e); in != nil && err == nil {
-			return in, err
-		}
-	}
-	if e, err := b.compileConstCompare(e); e != nil && err == nil {
-		return e, nil
-	}
 	lhs, err := b.compileExpr(e.LHS)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


philrz:
I've confirmed that `super` based on this branch at commit a5b42c7 now gives the expected response for both the simplified query and original sqllogictest query mentioned in #6416. 👍

PR #6422, opened 2025-12-05
Fix nulls in in expressions in sequential runtime

### LINKED ISSUE #6421: SQL: "NULL IN" behavior changed at commit d7e8921
## tl;dr

This query returns a result, but per SQL conventions should return nothing.

```
$ super -version
Version: d7e89211b

$ super -c "SELECT 1 WHERE NULL IN (NULL);"
{"1":1}
```

## Details

Repro is with super commit d7e8921, which is associated with the merge of the changes in #6417. The query above is a simplification of a query from [a sqllogictest](https://github.com/dolthub/sqllogictest/blob/a1f4aa3c1d691552a4a03d9603c45bcf93d0bb9d/test/random/aggregates/slt_good_78.test#L65139-L65142).

Here's the expected behavior when the query is run in Postgres.

```
$ psql postgres
Type "help" for help.

postgres=# \pset null 'NULL'
Null display is "NULL".
postgres=# SELECT 1 WHERE NULL IN (NULL);
 ?column? 
----------
(0 rows)
```

At commit 8ebde76 right before the merge of #6417, `super` was matching the SQL behavior.

```
$ super -version
Version: 8ebde7627

$ super -c "SELECT 1 WHERE NULL IN (NULL);"
[no output]
```

mattnibs:
Closes #6421

### DIFF
--- a/runtime/sam/expr/eval.go
+++ b/runtime/sam/expr/eval.go
@@ -138,11 +138,18 @@ func (i *In) Eval(this super.Value) super.Value {
 	if elem.IsError() {
 		return elem
 	}
+	if elem.IsNull() {
+		return super.NullBool
+	}
 	container := i.container.Eval(this)
 	if container.IsError() {
 		return container
 	}
+	errNull := errors.New("null")
 	err := container.Walk(func(typ super.Type, body scode.Bytes) error {
+		if body == nil {
+			return errNull
+		}
 		if coerce.Equal(elem, super.NewValue(typ, body)) {
 			return errMatch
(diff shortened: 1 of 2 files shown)


philrz:
I've confirmed by running this branch against the full set of "random aggregates" sqllogictest queries that all the affected queries with the #6421 symptom run successfully again.

mattnibs:
This is revealing an bug with `in` and nulls on the vector side, going to close this and create a new pr once I get that resolved.

PR #6424, opened 2025-12-05
vector.Or: Handle nulls

mattnibs:
This commit adjusts the behavior of vector.Or so that null values are properly set.


nwt (commented):
Test?

mattnibs:
@nwt I have an upcoming pr that adds tests that will cover this

PR #6425, opened 2025-12-05
Fix bugs with nulls and in expressions

### LINKED ISSUE #6003: SQL: Any comparison with NULL should yield UNKNOWN
While the SQL `NOT IN` support added in #5992 seems to work correctly for the general case, according to SQL conventions, there should have been no rows returned here.

```
$ super -c "SELECT * FROM intest WHERE a NOT IN (NULL, NULL)"
{a:42::int32,b:42::int32,c:42::int32}
{a:43::int32,b:42::int32,c:42::int32}
{a:44::int32,b:41::int32,c:44::int32}
```

Similar here:

```
$ super -c "SELECT * FROM intest WHERE NULL NOT IN ('a', 'b');"
{a:42::int32,b:42::int32,c:42::int32}
{a:43::int32,b:42::int32,c:42::int32}
{a:44::int32,b:41::int32,c:44::int32}
```

## Details

Repro is with with super commit e75dfda. This was found via queries from [this sqllogictest](https://github.com/duckdb/duckdb/blob/c677fa112c5597c16cd83ed24739129696d38d53/test/sql/projection/test_complex_expressions.test#L135-L137) and [this other sqllogictest](https://github.com/duckdb/duckdb/blob/c677fa112c5597c16cd83ed24739129696d38d53/test/sql/projection/test_complex_expressions.test#L159-L161).

Here's the queries returning no rows in Postgres, per the SQL convention.

```
$ psql postgres
psql (17.5 (Homebrew))
Type "help" for help.

postgres=# CREATE TABLE intest (a INTEGER, b INTEGER, c INTEGER);
INSERT INTO intest VALUES (42, 42, 42), (43, 42, 42), (44, 41, 44);
CREATE TABLE
INSERT 0 3

postgres=# SELECT * FROM intest WHERE a NOT IN (NULL, NULL)
postgres-# ;
 a | b | c 
---+---+---
(0 rows)

postgres=# SELECT * FROM intest WHERE NULL NOT IN ('a', 'b');
 a | b | c 
---+---+---
(0 rows)
```

Whereas in SuperDB, the entire table contents are returned

```
$ super -version
Version: e75dfdaf1

$ super -s intest 
{a:42::int32,b:42::int32,c:42::int32}
{a:43::int32,b:42::int32,c:42::int32}
{a:44::int32,b:41::int32,c:44::int32}

$ super -c "SELECT * FROM intest WHERE a NOT IN (NULL, NULL)"
{a:42::int32,b:42::int32,c:42::int32}
{a:43::int32,b:42::int32,c:42::int32}
{a:44::int32,b:41::int32,c:44::int32}

$ super -c "SELECT * FROM intest WHERE NULL NOT IN ('a', 'b');"
{a:42::int32,b:42::int32,c:42::int32}
{a:43::int32,b:42::int32,c:42::int32}
{a:44::int32,b:41::int32,c:44::int32}
```

While admittedly kind of weird examples, web searches show folks explaining the behavior as being related to `NULL` values in SQL representing "unknown" (i.e., of SQL's three logic states `TRUE`, `FALSE`, and `UNKNOWN`), so in a test for whether something is in a set of values of which some are unknown, the answer is also said to be unknown. Or to put it another way, `NOT IN` will return no rows if the list contains `NULL`. Similar logic (but flipped around) applies for the `NULL NOT IN` case.

### LINKED ISSUE #6421: SQL: "NULL IN" behavior changed at commit d7e8921

mattnibs:
This commit fixes an issue with in expressions in both vector and sequential runtimes where null::bool was not getting returned when encountering null values in various contexts.

Closes #6421
Closes #6003

### DIFF
--- a/runtime/vam/expr/logic.go
+++ b/runtime/vam/expr/logic.go
@@ -219,7 +219,8 @@ func (i *In) eval(vecs ...vector.Any) vector.Any {
 					nullTags = append(nullTags, i)
 				}
 			}
-			out := vector.NewConst(super.NullBool, uint32(len(nullTags)), bitvec.Zero)
+			length := uint32(len(nullTags))
+			out := vector.NewConst(super.NullBool, length, bitvec.NewTrue(length))
 			rhs = vector.ReversePick(rhs, nullTags)
 			return vector.Combine(rhs, nullTags, out)
 		}
--- a/runtime/ztests/expr/in.yaml
+++ b/runtime/ztests/expr/in.yaml
@@ -29,4 +29,4 @@ output: |
   [true,false,null::bool]
   [true,false,null::bool]
   [true,false,null::bool]
-  [error(0),error(0),null::bool]
+  [error(0),error(0),null]::[bool|error(int64)]
--- a/runtime/vam/expr/arrayexpr.go
+++ b/runtime/vam/expr/arrayexpr.go
@@ -109,7 +109,8 @@ func buildList(sctx *super.Context, elems []ListElem, in []vector.Any) ([]uint32
 	if len(types) == 1 {
 		return offsets, mergeSameTypeVecs(types[0], tags, vecs)
 	}
-	return offsets, vector.NewUnion(sctx.LookupTypeUnion(types), tags, vecs, bitvec.Zero)
+	nulls := unionNulls(tags, vecs)
+	return offsets, vector.NewUnion(sctx.LookupTypeUnion(types), tags, vecs, nulls)
 }
 
 func unwrapSpread(vec vector.Any) (vector.Any, []uint32, []uint32) {
@@ -123,6 +124,23 @@ func unwrapSpread(vec vector.Any) (vector.Any, []uint32, []uint32) {
 		return vals, offsets, vec.Index
 	}
 	return nil, nil, nil
+}
+
+func unionNulls(tags []uint32, vecs []vector.Any) bitvec.Bits {
+	for i, vec := range vecs {
+		if !vector.NullsOf(vec).IsZero() {
+			nulls := bitvec.NewFalse(uint32(len(tags)))
+			offs := make([]uint32, len(vecs))
+			for j, t := range tags {
+				if t >= uint32(i) && vector.NullsOf(vecs[t]).IsSet(offs[t]) {
+					nulls.Set(uint32(j))
+				}
+				offs[t]++
+			}
+			return nulls
+		}
+	}
+	return bitvec.Zero
 }
 
 func mergeSameTypeVecs(typ super.Type, tags []uint32, vecs []vector.Any) vector.Any {


runtime/ztests/expr/in.yaml
nwt: Nit: Add null here instead of adding another test.

```suggestion
  values [1 in this, 9 in this, null in this]
```

runtime/sam/expr/eval.go
```diff
@@ -138,11 +138,18 @@ func (i *In) Eval(this super.Value) super.Value {
...
+		return super.NullBool
+	}
 	container := i.container.Eval(this)
 	if container.IsError() {
 		return container
 	}
+	var hasnull bool
 	err := container.Walk(func(typ super.Type, body scode.Bytes) error {
+		if body == nil {
+			hasnull = true
```
nwt, line 151: There's no reason to continue the walk once you've seen a null.

```suggestion
			return errNull
```

mattnibs: There is `1 in [null,1]` should return `true`.

nwt: Good point!

nwt:
@mattnibs: One way to get both runtimes to behave the same for `echo 'error(0)' | super -c 'values [1 in this, null in this]'` is this patch.

```diff
diff --git a/runtime/vam/expr/arrayexpr.go b/runtime/vam/expr/arrayexpr.go
index 4f4773b84..d1d890d6c 100644
--- a/runtime/vam/expr/arrayexpr.go
+++ b/runtime/vam/expr/arrayexpr.go
@@ -109,7 +109,11 @@ func buildList(sctx *super.Context, elems []ListElem, in []vector.Any) ([]uint32
 	if len(types) == 1 {
 		return offsets, mergeSameTypeVecs(types[0], tags, vecs)
 	}
-	return offsets, vector.NewUnion(sctx.LookupTypeUnion(types), tags, vecs, bitvec.Zero)
+	nulls := unionNulls(tags, vecs)
+	// When nulls is not empty, this will return a union with nulls
+	// sprinkled across its values vectors rather than confined to a single
+	// union-typed vector.Const.
+	return offsets, vector.NewUnion(sctx.LookupTypeUnion(types), tags, vecs, nulls)
 }
 
 func unwrapSpread(vec vector.Any) (vector.Any, []uint32, []uint32) {
@@ -139,3 +143,20 @@ func mergeSameTypeVecs(typ super.Type, tags []uint32, vecs []vector.Any) vector.
 	}
 	return vb.Build(bitvec.Zero)
 }
+
+func unionNulls(tags []uint32, vecs []vector.Any) bitvec.Bits {
+	for i, vec := range vecs {
+		if !vector.NullsOf(vec).IsZero() {
+			nulls := bitvec.NewFalse(uint32(len(tags)))
+			offs := make([]uint32, len(vecs))
+			for j, t := range tags {
+				if t >= uint32(i) && vector.NullsOf(vecs[t]).IsSet(offs[t]) {
+					nulls.Set(uint32(j))
+				}
+				offs[t]++
+			}
+			return nulls
+		}
+	}
+	return bitvec.Zero
+}
```

This doesn't feel ideal because it doesn't produce the usual union representation (i.e., with a union-typed const vector for nulls), but I think it's fine for now. (And don't forget that it'll go away when we ditch typed nulls).

runtime/vam/expr/logic.go
```diff
@@ -210,6 +211,18 @@ func (i *In) eval(vecs ...vector.Any) vector.Any {
...
 	if rhs.Type().Kind() == super.ErrorKind {
+		if nulls := vector.NullsOf(lhs); !nulls.IsZero() {
+			// Nulls in LHS should still be null.
+			var nullTags []uint32
+			for i := range rhs.Len() {
+				if nulls.IsSetDirect(i) {
+					nullTags = append(nullTags, i)
+				}
+			}
+			out := vector.NewConst(super.NullBool, uint32(len(nullTags)), bitvec.Zero)
```
nwt: Nulls should all be true here.

```suggestion
			length := uint32(len(nullTags))
			out := vector.NewConst(super.NullBool, length, bitvec.NewTrue(length))
```

PR #6363, opened 2025-11-14
update SuperSQL in book docs

mccanne:
This commit adds substantial content to the SuperSQL section of the book docs.  The SQL section is still a placeholder.

PR #6427, opened 2025-12-06
Edits to super-sql/queries doc

philrz:
Like #6396, this is a bite-sized edit to one of the docs created/modified in #6363.

In this pass I've started trying to make use of the recent mdtest Playground enhancements from @nwt to drop labels and make them narrower. If the approach here passes scrutiny I'll apply it in similar upcoming edits.

In addition to the inline comments, a couple questions for @mccanne about parts of this doc I spotted but did not propose edits:

1. Is detail about the availability of `|>` planned as part of the SQL docs you're currently drafting? It's mentioned here in passing and if we expect to have text about it I'll make a note-to-self to circle back and link to it. No doubt users that have seen Google's pipe SQL will be familiar with it, but I'm more thinking of users coming to SuperSQL for their first pipe SQL experience being curious when they see it mentioned in passing.

2. Are we standardizing on "SQL Clauses" over "SQL Operators", or are these two different concepts we expect to describe separately in the upcoming SQL docs? I ask because this doc currently links the text "SQL operator" to a doc titled "SQL Clauses" in the `SUMMARY.md`, so I wasn't sure if this was something to make consistent now or if it's about to all become clearer in what you're drafting.


book/src/super-sql/queries.md
```diff
@@ -32,15 +32,14 @@ along with an operator sequence in the parentheses having the structure:
...
+* as the body of a [declared operator](declarations/operators.md).
 
 The parenthesized block forms a
-[lexical scope](https://en.wikipedia.org/wiki/Scope_(computer_science)#Lexical_scope)
-and the bindings created by declarations
-within the scope are reachable only within that scope inclusive
-of other scopes defined within the scope.
+[lexical scope](https://en.wikipedia.org/wiki/Scope_(computer_science)#Lexical_scope).
+Bindings created by declarations within the scope are reachable within that
+scope and any nested scopes, but not outside it.
```
philrz: I had a little difficulty parsing this sentence in its original form, so I've taken a shot at a rewrite. Please confirm it preserves essence of the original, though.

book/src/super-sql/queries.md
```diff
@@ -168,12 +178,11 @@ A text entity represents a string where quotes can be omitted for
...
-[`from`](operators/from.md) operator and as named arguments
-to `from` and the
-[`load`](operators/load.md) operator.
+[from](operators/from.md) operator and as named arguments
+to the `from` and [load](operators/load.md) operators.
 
 Specifically, a text entity is one of:
-* a string literal (double quoted, single quoted, or raw string),
+* a [string literal](types/string.md) (double quoted, single quoted, or raw string),
 * a path consisting of a sequence of characters consisting of letters, digits, `_`,  `$`,  `.`, and `/`, or
```
philrz: Is "path" here a reference to "file paths" as mentioned at the top of this section, the "path" through nested records to a field reference, or both? If meant to be "file paths" then I'd change it to be explicit. If one of the the others, then it leads me to wonder if we have a formal definition of that somewhere or a planned home for it, because I couldn't seem to find one when I went looking for it (but would link to it if so).

book/src/super-sql/queries.md
mccanne: This isn't right and needs to be further rephrased

book/src/super-sql/queries.md
```diff
@@ -51,22 +50,34 @@ of references to data input is defined by
 [pipe scoping](intro.md#pipe-scoping) and
 [relational scoping](intro.md#relational-scoping).
 
-For example, [this example](declarations/constants.md#examples) of a constant declaration
-```
+For example, this query containing a constant declaration emits the value `3.14`
+```mdtest-spq {data-layout='no-labels'} {style='text-align:left;width:20%'}
```
mccanne: This formatting isn't ready for prime time.  When there's no input, there shouldn't be an input pain.  There should just be a query and the result.

book/src/super-sql/queries.md
```diff
@@ -187,7 +196,7 @@ subsequent newline.
 
 Multi-line comments are C style and begin with `/*` and end with `*/`.
 
-```mdtest-spq
+```mdtest-spq {data-layout='no-labels'} {style='text-align:left;width:50%'}
```
mccanne: I was imagining the playground pane being centered when width is less than 100%.  This doesn't look good to me left justified.

philrz: This was just a first try based on personal preference as I happened to find the jump to the center jarring in the doc flow, but I'm not married to that. In the spirit of iteration I just pushed a commit to make the examples in this branch all centered so you can see how you like that instead. I recall this morning you also wondered aloud if some kind of "indent" solution might be preferred over "left" or "centered", and I know @nwt is working on some other style variations so your reaction here may help inform that.

FYI in case you want to toggle it back and forth on your own, going from "left" to "centered" meant changing:

```
{style='text-align:left;width:50%'}
```

to:

```
{style='margin:auto;width:50%'}
```

book/src/super-sql/queries.md
philrz: As covered on Zoom this morning, a `make` in the `book/` directory was just needed to see the latest changes including the dropped input pane, so I'm marking this comment Resolved.

book/src/super-sql/queries.md
nwt: For what it's worth, "and any nested scopes" certainly feels clearer to me than "inclusive of other scopes defined within the scope".

book/src/super-sql/queries.md
mccanne: We could just call that "an unquoted string"

book/src/super-sql/queries.md
mccanne: I'm not saying original text is good, but the current text is unclear because "nested scope" is meant to be child scope but it is a generic concept and would arguably include parent scopes.  "inclusive of other scopes defined within the scope" is referring to child scopes referred to by the first scope.  I was trying to avoid having to define parent/child scopes.

nwt: Makes sense. Does "within that scope and any scopes nested within it" resolve the ambiguity well enough?

mccanne: sounds good

philrz: Ok, that sounds like consensus. I just pushed a commit that adjusts this to include @nwt's suggestion. I also dropped the "but not outside it" part. I hadn't heard any explicit 👍 or 👎 about that part of my first proposal, but in an offline exchange @nwt confirmed my suspicion that it read as superfluous, so figured I might as well drop it since it wasn't in the original.

book/src/super-sql/queries.md
philrz: @mccanne: Ok, just pushed a commit so now that bullet reads as:

> * an unquoted string consisting of a sequence of characters consisting of letters, digits, `_`,  `$`,  `.`, and `/`, or

philrz:
@mccanne: I think I've incorporated the suggested changes to the content, and I've also updated the conventions in the README to reflect your latest preferences regarding labels/width on the inline playground examples. While we've been using the examples here as the reference for the iterations on vertical space and Copy button, since I expect those to come in a separate branch from @nwt maybe you're game to Approve the enhancements here thus far and we'll keep tracking the vertical/Copy as a separate effort.

PR #6432, opened 2025-12-09
Reformat count operator

### LINKED ISSUE #6413: count operator without argument
## tl;dr

It has been proposed that when the `count` operator is used without an argument it should produce output like this:

```
$ echo '{foo:"bar",a:"b"} {foo:"baz",a:"c"}' | super -c 'count' -
{that:{foo:"bar",a:"b"},count:1::uint64}
{that:{foo:"baz",a:"c"},count:2::uint64}
```

## Details

At the time this issue is being opened, super is at commit 30f64bb.

The `count` operator added in #6344 currently requires an argument in record syntax that specifies the name of a record field to hold the numeric count and optional additional elements that determine what becomes of the input value. However, to spare the user from having to remember the record syntax, it's been proposed that `count` without argument would behave as shown above where the input value would be embedded under a field called `that` (per conventions for [derived field names](https://superdb.org/super-sql/types/record.html#derived-field-names)) and the numeric count would be placed in a field called `count`.

If the user called `count` in that simplified manner and actually wanted to spread an input record value next to the numeric count, it's expected they'd tack on something like:

```
$ echo '{foo:"bar",a:"b"} {foo:"baz",a:"c"}' | super -c 'count | values {...that,count}' -
{foo:"bar",a:"b",count:1::uint64}
{foo:"baz",a:"c",count:2::uint64}
```

If the user wants to assign a name to the numeric count field and do `values`-like output assembly all in one shot, an optional argument to `count` in record syntax would still be available, e.g.,

```
$ echo '{foo:"bar",a:"b"} {foo:"baz",a:"c"}' | super -c 'count {foo,a,cnt}' -
{foo:"bar",a:"b",cnt:1::uint64}
{foo:"baz",a:"c",cnt:2::uint64}
```

mattnibs:
This commit reformats the count operator so the record expression is no longer required and the count alias in the record expression is now the last argument.

Closes #6413

### DIFF
--- a/compiler/semantic/checker.go
+++ b/compiler/semantic/checker.go
@@ -88,12 +88,14 @@ func (c *checker) op(typ super.Type, op sem.Op) super.Type {
 	case *sem.BadOp:
 		return c.unknown
 	case *sem.CountOp:
-		elems := []sem.RecordElem{
-			&sem.FieldElem{Name: op.Alias, Value: &sem.LiteralExpr{Value: "0(uint64)"}},
-		}
+		var elems []sem.RecordElem
 		if op.Expr != nil {
 			elems = append(elems, op.Expr.(*sem.RecordExpr).Elems...)
 		}
+		elems = append(elems, &sem.FieldElem{
+			Name:  op.Alias,
+			Value: &sem.LiteralExpr{Value: "0(uint64)"},
+		})
 		return c.recordElems(typ, elems)
 	case *sem.CutOp:
 		return c.pathsToType(c.assignments(typ, op.Args))
--- a/compiler/semantic/op.go
+++ b/compiler/semantic/op.go
@@ -652,23 +652,38 @@ func (t *translator) semOp(o ast.Op, seq sem.Seq) sem.Seq {
 			Cases: cases,
 		})
 	case *ast.CountOp:
-		if len(o.Expr.Elems) == 0 {
-			t.error(o.Expr, errors.New("count record expression must not be empty"))
-			return append(seq, badOp())
-		}
-		first, ok := o.Expr.Elems[0].(*ast.ExprElem)
-		if !ok {
-			t.error(o.Expr, errors.New("first element in record expression for count must be an identifier"))
-			return append(seq, badOp())
-		}
-		alias := first.Expr.(*ast.IDExpr).Name
+		var alias string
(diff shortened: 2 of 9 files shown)


philrz:
I just tested out this branch at commit d053f6e and it matches the details captured in #6413 from the last time we discussed the specifics as a team, so it's a functional 👍 for me.

Also, assuming this merges, I'm game to write up the user-facing doc for it.

compiler/semantic/checker.go
```diff
@@ -88,12 +88,17 @@ func (c *checker) op(typ super.Type, op sem.Op) super.Type {
...
 		if op.Expr != nil {
 			elems = append(elems, op.Expr.(*sem.RecordExpr).Elems...)
 		}
+		elems = append(
+			elems,
+			&sem.FieldElem{
+				Name:  op.Alias,
+				Value: &sem.LiteralExpr{Value: "0(uint64)"},
+			},
+		)
```
nwt: Nit: We usually format this this sort of thing with fewer newlines.

```suggestion
		elems = append(elems, &sem.FieldElem{
			Name:  op.Alias,
			Value: &sem.LiteralExpr{Value: "0(uint64)"},
		})
```

compiler/semantic/op.go
```diff
@@ -652,23 +652,34 @@ func (t *translator) semOp(o ast.Op, seq sem.Seq) sem.Seq {
...
+			if n == 0 {
+				t.error(o.Expr, errors.New("count record expression must not be empty"))
+				return append(seq, badOp())
+			}
+			last, ok := o.Expr.Elems[n-1].(*ast.ExprElem)
+			if !ok {
+				t.error(o.Expr, errors.New("last element in record expression for count must be an identifier"))
+				return append(seq, badOp())
+			}
+			alias = last.Expr.(*ast.IDExpr).Name
```
nwt: This panics on `count {1}`.

```suggestion
			if last, ok := o.Expr.Elems[n-1].(*ast.ExprElem); ok {
				if idExpr, ok := last.Expr.(*ast.IDExpr); ok {
					alias = idExpr.Name
				}
			}
			if alias == "" {
				t.error(o.Expr, errors.New("last element in record expression for count must be an identifier"))
				return append(seq, badOp())
			}
```

PR #6436, opened 2025-12-11
semantic/schema: Alphabetize schemas

mattnibs:


mattnibs:
Things are getting a bit unwieldy in this file hopefully this brings a little organization to it.

PR #6439, opened 2025-12-11
Document "count" operator

### LINKED ISSUE #6413: count operator without argument

philrz:
A user-facing doc to go alongside #6432.

Closes #6413

### DIFF
--- a/book/src/super-sql/operators/count.md
+++ b/book/src/super-sql/operators/count.md
@@ -19,9 +19,9 @@ When `<record-expr>` is present, it must be a
 rightmost element is the name of a field to hold the numeric count and any
 preceding elements are evaluated on the input value.
 
-If the optional `<record-expr>` is absent, the output record contains the
-input value in an element of [derived field name](../types/record.md#derived-field-names)
-`that` and a field of name `count` that contains the numeric count.
+If the optional `<record-expr>` is absent, the output record is created with a
+[derived field name](../types/record.md#derived-field-names) resulting in the
+equivalent of `count {that:this,count}`.
 
 ### Examples
 


book/src/super-sql/aggregates/count.md
```diff
@@ -1,6 +1,9 @@
 ### Aggregate Function
 
-&emsp; **count** &mdash; count input values
+&emsp; **count** &mdash; count all input values
+
+>[!TIP]
+> For running counts as values arrive, see the [count](../operators/count.md) operator.
```
nwt: Nit:

```suggestion
> For a running count as values arrive, see the [count](../operators/count.md) operator.
```

book/src/super-sql/operators/count.md
```diff
@@ -0,0 +1,87 @@
+### Operator
+
+[✅](../intro.md#data-order)&emsp; **count** &mdash; emit a running count of input values
+
+>[!TIP]
+> For final counts across all input values, see the [count](../aggregates/count.md) aggregate function.
```
nwt: Nit:

```suggestion
> For a final count of all input values, see the [count](../aggregates/count.md) aggregate function.
```

book/src/super-sql/operators/count.md
```diff
@@ -0,0 +1,87 @@
+### Operator
+
+[✅](../intro.md#data-order)&emsp; **count** &mdash; emit a running count of input values
```
nwt: Nit:

```suggestion
[✅](../intro.md#data-order)&emsp; **count** &mdash; emit records containing a running count of input values
```

book/src/super-sql/operators/count.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Synopsis
+
+```
+count [ <record-expr> ]
+```
+### Description
+
+The `count` operator produces a running count of input values as a field in
+an output record, including some or all of parts of input values in the output.
```
nwt:
```suggestion
The `count` operator produces records that include a running count of its input values.

File: book/src/super-sql/operators/count.md [hunk 5]
```diff
@@ -0,0 +1,87 @@
...
+
+If the optional `<record-expr>` is absent, the output record contains the
+input value in an element of [derived field name](../types/record.md#derived-field-names)
+`that` and a field of name `count` that contains the numeric count.
```
nwt: Nit: I think a "show, don't tell" approach makes this clearer and more concise. I'd also move it after the next paragraph.

```suggestion
Without `<record-expr>`, `count` is equivalent to `count {that:this,count}`.
```

book/src/super-sql/operators/count.md
```diff
@@ -0,0 +1,87 @@
...
+
+When `<record-expr>` is present, it must be a
+[record expression](../types/record.md#record-expressions) in which the
+rightmost element is the name of a field to hold the numeric count. Any
+preceding elements determine what parts of the input value to include in the
+output record, similar to how a record expression may be used with the
+[values](values.md) operator.
```
nwt: Nit: The second sentence here isn't really accurate since the record expression can include fields that don't reference the input value, like `{a:1,count}`. I think we'd be better off using language similar to that in the first paragraph of the description in values.md ("evaluat[ed] ... on each input value").

```suggestion
When `<record-expr>` is present, it must be a
[record expression](../types/record.md#record-expressions) in which the
rightmost element is the name of a field to hold the numeric count and any
preceding elements are evaluated on the input value.
```

book/src/super-sql/operators/count.md
philrz: @nwt: I had to change "counts" to "count" to make the grammar work here, but, accepted.

book/src/super-sql/operators/count.md
philrz: @nwt: I understand your point and it's a close call, but in the spirit of the rule of nits, I've just moved the paragraph as you suggested while keeping the text. Here's my thinking. The concept of "special value `this`" appears all over our docs, plus readers may be familiar with the concept from multiple other programming languages. Meanwhile, how `this` can sometimes become `that` only comes up in a couple places. Given how new our stuff is, in the event this doc happens to be the first place they encounter the concept, I lean toward having the opportunity to say a little more and link off to the "derived field names" docs coverage lest they be puzzled at the thought this is something unique to `count` or go off on a difficult scavenger hunt trying to learn more about it. (BTW, the hunt is made slightly worse by the fact that the search feature that comes with mdbook shows no hits for "that", which I assume is because it's such a common English word that it would just give thousands of useless hits. But I'm not looking to slay that dragon today either.)

nwt: Maybe there's some middle ground where you mention derived field names but also show the equivalent in lieu of that textual description of the record (which isn't really complete since it doesn't specify the field order)?

philrz: @nwt: Ok, I've pushed a change that I think meets your definition of "middle path", so I'm declaring victory and merging. Thanks for the reviews!

PR #6441, opened 2025-12-12
ztest: allow multiple ztests per YAML file

nwt:
Specify multiple ztests in a single file by separating them with a line containing three hyphens ("---").

    spq: values 1

    output: |
      1

    ---

    spq: values 2

    output: |
      2


mattnibs (approved):
🎉

PR #6447, opened 2025-12-15
Allow env var override for running ztests in vector runtime

philrz:
## What's Changed

This adds support for env var `ZTEST_VECTOR` which can be set to `true` to have the same effect as having added `vector: true` to each ztest YAML.

## Why

I want to sometimes run the [sqllogic-ztests](https://github.com/brimdata/sqllogic-ztests) in vector runtime, but since there's still lots more failures there than in sequential, I'd like to keep the "golden" set of ztest YAML set for just the default of running sequential, but have an easy way to override that. The only way I saw to do this before was to keep a separate copy of all the ztest YAML files edited to include `vector: true`, which is cumbersome.

## Details

I'm not much of a Go developer, so I had Claude AI create this diff. It's functionally giving me what I need (e.g., [this Actions run](https://github.com/brimdata/sqllogic-ztests/actions/runs/20245068983/job/58122863390)) but I respect AI's code may have its shortcomings, so if someone's game to rewrite it better or I have to just live off my branch, I'm ok with that.


philrz:
We've closed much of the gap between vector and sequential runtime on the sqllogictests so I'm optimistic that I won't need this functionality even in the sqllogic-ztests repo. Once they're at parity I'll be fine just marking all the ztests as `vector: true` in the classic way. I'm closing this PR and will live off the branch until we reach that point soon, at which point I'll retire the branch too.

PR #6448, opened 2025-12-16
implement nullif() function

### LINKED ISSUE #6249: SQL: NULLIF function
SuperSQL does not yet have support for the standard SQL function [`NULLIF`](https://www.postgresql.org/docs/current/functions-conditional.html#FUNCTIONS-NULLIF).

## Details

Repro is with super commit c6ebc21. This issue was found via [a sqllogictest](https://github.com/dolthub/sqllogictest/blob/a1f4aa3c1d691552a4a03d9603c45bcf93d0bb9d/test/random/groupby/slt_good_0.test#L53154-L53159).

Here's the function being used in Postgres.

```
$ psql postgres -P "null=NULL"
psql (17.6 (Homebrew))
Type "help" for help.

postgres=# SELECT NULLIF(1,1) AS null_if_equal;
 null_if_equal 
---------------
          NULL
(1 row)

postgres=# SELECT NULLIF(1,2) AS null_if_equal;
 null_if_equal 
---------------
             1
(1 row)
```

Whereas currently in `super`:

```
$ super -version
Version: c6ebc2139

$ super -c "SELECT NULLIF(1,1) AS null_if_equal;"
SELECT NULLIF(1,1) AS null_if_equal;
       ~~~~~~~~~~~
```

I can see this function mentioned as far back as the [1992 SQL spec](https://www.contrib.andrew.cmu.edu/~shadow/sql/sql1992.txt).

nwt:
Closes #6249.

### DIFF
--- a/runtime/sam/expr/function/function.go
+++ b/runtime/sam/expr/function/function.go
@@ -115,6 +115,9 @@ func New(sctx *super.Context, name string, narg int) (expr.Function, error) {
 		argmax = 0
 		argmin = 0
 		f = &Now{}
+	case "nullif":
+		argmin, argmax = 2, 2
+		f = newNullIf()
 	case "parse_sup":
 		f = newParseSUP(sctx)
 	case "parse_uri":
--- a/runtime/sam/expr/function/nullif.go
+++ b/runtime/sam/expr/function/nullif.go
@@ -1,0 +1,29 @@
+package function
+
+import (
+	"github.com/brimdata/super"
+	"github.com/brimdata/super/order"
+	"github.com/brimdata/super/runtime/sam/expr"
+)
+
+type NullIf struct {
+	compareFn expr.CompareFn
+}
+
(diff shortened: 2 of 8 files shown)


philrz:
I tested out this branch with the sqllogic-ztests and I think it's solid. There were perviously 105,358 queries that contain `NULLIF` that were all doomed to failure. With this branch at commit e4ec65a, that number comes down to 15,275 failures in sequential runtime and 21,039 in vector runtime. Because we have a lot of other open issues causing sqllogictest failures, it's unsurprising that fixing an issue may expose another issue that was hiding a layer beneath. Of course, given the "fuzzy" nature of most queries, it's difficult to know with 100% certainty that none of these failures are due to some implementation problem with `NULLIF`. But I've scanned the failures and randomly audited a handful of them more closely and they all have the look & feel of some of our other known issues, so I'm in favor of banking this progress so we can keep chipping away at other problems.

PR #6455, opened 2025-12-18
Edits to super-sql/declarations docs

philrz:
This is another set of bite-sized edits to docs created/modified in #6363.

---

In addition to the inline comments, I spotted some text that I think needs attention but I'm not confident in a proposed edit so I can't mention it inline. Currently in the [Operators declarations doc](https://superdb.org/super-sql/declarations/operators.html) one of the examples is introduced with the text:

> _Error checking works as expected for non-l-values used as l-values_

The term "l-values" seems esoteric enough that it seems like it needs to be defined somewhere and linked to, but I'm not sure where and how. The only other place I see it currently used is in the [join operator doc](https://superdb.org/super-sql/operators/join.html). If that's the correct definition for this context as well (I'm actually not sure...) then maybe we could move that text somewhere else (to where? not sure...) and link to it from both places.

### DIFF
--- a/book/src/super-sql/declarations/operators.md
+++ b/book/src/super-sql/declarations/operators.md
@@ -51,7 +51,7 @@ These expression arguments can be viewed as a
 [closure](https://en.wikipedia.org/wiki/Closure_(computer_programming))
 though there is no persistent state stored in the closure.
 The [jq](https://github.com/jqlang/jq/wiki/jq-Language-Description#the-jq-language) language
-describes its expression semantics as closures as well, though unlike `jq`,
+describes its expression semantics as closures as well, though unlike jq,
 the operator expressions here are not generators and do not implement backtracking.
 
 ### Examples
--- a/book/src/super-sql/declarations/types.md
+++ b/book/src/super-sql/declarations/types.md
@@ -17,7 +17,7 @@ forward references to other named types.  In particular, named types cannot be r
 > A future version of SuperSQL may include recursive types.  This is a research topic
 > for the SuperDB project.
 
-[Input data may create named types](../../formats/model.md#3-named-type) that conflict with type declarations.  In this case,
+Input data may create [named types](../../formats/model.md#3-named-type) that conflict with type declarations.  In this case,
 a reference to a declared type in the query text uses the type definition of the nearest
 containing scope that binds the type name independent of types in the input.
 


book/src/super-sql/declarations/functions.md
```diff
@@ -25,21 +25,21 @@ fn add(a,b): a+b
...
+values {x:1} | values add(x,2)
 ```
 is passed the record the `{a:x,b:1}`, which after resolving `x` to `1`,
-is `{a:1,b:1}` and thus evaluates the expression
+is `{a:1,b:2}` and thus evaluates the expression
 ```
 this.a + this.b
 ```
-which results in `2`.
+which results in `3`.
```
philrz, line 35: About this example:

1. I changed the literal `1` to `2` because it becomes easier for the reader to follow the journey of `x` and the literal number if they're different.
2. While I generally try to avoid unprotected examples, I respect that this one reads better when introduced piecemeal like this. I've made a note-to-self to bring it up again when we get to talking about ways to get some second-order protection on these.

book/src/super-sql/declarations/functions.md
```diff
@@ -25,21 +25,21 @@ fn add(a,b): a+b
...
 
-Any function-as-value arguments passed to a function do not appear in the `this`
+Any [function references](../expressions/functions.md#function-references) passed to a function do not appear in the `this`
```
philrz, line 37: I assume this is an appropriate substitution (wanted something we could link to rather than invoking a new term), but please double check.

book/src/super-sql/declarations/functions.md
```diff
@@ -67,24 +67,35 @@ fn <id> ( [<param> [, <param> ...]] ) : [
...
   unnest {outer:val,item:a}
   | collect(outer+item)
 )
 values apply([1,2,3], 1)
+# input
+
+# expected output
+[2,3,4]
 ```
-See the example below.
```
philrz, line 87: I spent a fair amount of time fiddling with this one. Unlike the other example further up in this Functions doc (that I thought reads best piecemeal and hence has to currently stay unprotected), in this one the declaration+query are big enough that it seemed odd to reference the expected error but not show it and instead point the user off to the Examples section. Also, as the original declaration was written it happened to contain a syntax error, which speaks to why I think protecting is the better approach whenever it can be justified.

That said, I respect that these inline examples still look a little rough until we have the benefit of the expected "vertical spacing" mdtest enhancements from @nwt.

If this approach of showing the full examples here is acceptable, then a follow on question is whether to keep the same queries as they still appear below in the Examples section, as they're now effectively redundant. Personally, I think it's fine to have both since it's important enough of a concept that I like how it's introduced here with the extra narrative, but if a user happens to just be browsing the Examples it's good that it's highlighted there as well.

Finally, the reference to "alternative form of unnest" here and "compound unnest form" in the text in Examples below seemed vague to me at first. Because I'm very close to these tools I caught on pretty quickly that it was a reference to calling `unnest` with a two-field record, but I'm concerned new users may not "get it" right away. To try to improve on that, I made an edit in this PR to the `unnest` doc so it now uses the word "compound" where that usage is introduced, so at least the user can find that if they follow the link to the `unnest` doc.

book/src/super-sql/declarations/intro.md
```diff
@@ -12,7 +12,7 @@ Declarations may be created for
 * [operators](operators.md), or
 * [pragmas](pragmas.md).
 
-All of the names defined in a given scope are available to other declarations defined
+With the exception of types that reference other types, all of the names defined in a given scope are available to other declarations defined
```
philrz, line 15: While I don't love having to add this big exception text up front, I saw the text in the "types" doc that contradicted the way this statement was originally written.

book/src/super-sql/declarations/operators.md
```diff
@@ -43,29 +43,29 @@ In contrast to function calls, where the arguments are evaluated at the call sit
...
 ```mdtest-spq
 # spq
 op echo: (
   values this
 )
-echo
+call echo
 # input
 {x:1}
 # expected output
```
philrz, line 71: Since we introduced `call` above I figured at least one example might as well use it. 🤷‍♂️

book/src/super-sql/declarations/queries.md
```diff
@@ -25,12 +25,11 @@ As with any expression subquery, multiple values result in an error, so when
 this is expected, the query reference may be enclosed in brackets to form
 an array subquery.
 
-To create a query that can be used as an operator and thus can operate on its input,
-declare an [operator](operators.md).
+To create a query that can operate on its input, declare an [operator](operators.md).
```
philrz, line 28: It seemed like this could be simplified, but please double check.

book/src/super-sql/operators/unnest.md
```diff
@@ -47,7 +47,7 @@ If `<expr>` is a record, it must have two fields of the form:
 {<first>: <any>, <second>:<array>}
 ```
 where `<first>` and `<second>` are arbitrary field names, `<any>` is any
-SuperSQL value, and `<array>` is an array value.  In this case, the derived
+SuperSQL value, and `<array>` is an array value.  In this compound form of `unnest`, the derived
```
philrz, line 50: This is the edit I made to go with the use of the word "compound" in the Functions declarations doc.

book/src/super-sql/declarations/functions.md
mccanne: LGTM

book/src/super-sql/declarations/operators.md
```diff
@@ -43,29 +43,29 @@ In contrast to function calls, where the arguments are evaluated at the call sit
...
+by these expressions (e.g., [constants](constants.md), [functions](functions.md), [named queries](queries.md), etc.) are bound to
 those entities using the lexical scope where they are defined rather than the
 scope of their expanded use in the operator's definition.
 
 These expression arguments can be viewed as a
 [closure](https://en.wikipedia.org/wiki/Closure_(computer_programming))
 though there is no persistent state stored in the closure.
 The [jq](https://github.com/jqlang/jq/wiki/jq-Language-Description#the-jq-language) language
-describes its expression semantics as closures as well, though unlike jq,
+describes its expression semantics as closures as well, though unlike `jq`,
```
mccanne: I feel like this reference is to jq the system, not `jq` the command.  Kind of like `super` vs SuperDB.  Wouldn't say `SuperDB`.  Or `clickhouse` the command vs ClickHouse the system.

book/src/super-sql/declarations/types.md
```diff
@@ -4,25 +4,25 @@ Named types are declared with the syntax
...
-declared in the same scope or in an ancestor scope, i.e., types cannot contained
+declared in the same scope or in an ancestor scope, i.e., types cannot contain
 forward references to other named types.  In particular, named types cannot be recursive.
 
 >[!NOTE]
 > A future version of SuperSQL may include recursive types.  This is a research topic
 > for the SuperDB project.
 
-Input data may create named types that conflict with type declarations.  In this case,
+[Input data may create named types](../../formats/model.md#3-named-type) that conflict with type declarations.  In this case,
```
mccanne: Change link to [named types] instead of the whole clause?

book/src/super-sql/declarations/operators.md
philrz: @mccanne: Fair point. I've dropped the backticks around "jq".

book/src/super-sql/declarations/types.md
philrz: @mccanne: Ok, I've shortened the link to be just around "named types". FWIW, I was doing something intentional in my original attempt. The section in which this link appears happens to open by mentioning "named types" so I was trying to call attention to there being another area of the wider architecture where the input data may create named types and hence make the link stand out (i.e., trying to make it obvious that I'm not just linking back to this same page). But I respect all that's probably too subtle and the longer linked text is ugly, hence going with your suggestion.

PR #6456, opened 2025-12-18
improve parquet error message requiring seekable input

mccanne:


sio/parquetio/reader.go
```diff
@@ -15,10 +15,12 @@ import (
 	"github.com/brimdata/super/sio/arrowio"
 )
 
+var errNotSeekable = errors.New("parquet format requires seekable input")
```
nwt:
```suggestion
var errNotSeekable = errors.New("Parquet format requires seekable input")
```

nwt (approved):

PR #6465, opened 2025-12-23
change "where" clause of agg functions to "filter"

### LINKED ISSUE #6261: SQL: Can't do math between aggregate function calls with modifiers (e.g., DISTINCT)
This query doesn't work in `super` but does in Postgres.

```
$ super -f parquet -o data.parquet -c "values {a:1},{a:1},{a:2}" &&
  super -c "SELECT SUM(DISTINCT a) + SUM(DISTINCT a) FROM data.parquet;"

parse error at line 1, column 24:
SELECT SUM(DISTINCT a) + SUM(DISTINCT a) FROM data.parquet;
                   === ^ ===
```

## Details

Repro is with super commit a87150f. This issue was found via [a sqllogictest](https://github.com/dolthub/sqllogictest/blob/a1f4aa3c1d691552a4a03d9603c45bcf93d0bb9d/test/random/expr/slt_good_0.test#L681-L685).

Here it is working as expected in Postgres.

```
$ psql postgres
psql (17.6 (Homebrew))
Type "help" for help.

postgres=# CREATE TABLE data (a INTEGER);
CREATE TABLE

postgres=# INSERT INTO data values (1), (1), (2);
INSERT 0 3

postgres=# SELECT SUM(DISTINCT a) + SUM(DISTINCT a) FROM data;
 ?column? 
----------
        6
(1 row)
```

Note that it executes math ok as long as I leave out the `DISTINCT` modifier.

```
$ super -version
Version: a87150f14

$ super -c "SELECT SUM(a) + SUM(a) FROM data.parquet;"
{"SUM(a)+SUM(a)":8}
```

I'm pretty sure `ALL` will need the same fix once we've added support for it as an aggregate modifier (#6258).

### LINKED ISSUE #6330: SQL: Extra parens around aggregation causes error value
## tl;dr

It's my understanding that the extra parens here should have no effect, but when run with `super` this SQL query produces an `error` value.

```
$ super -c "SELECT (MAX (DISTINCT 1));"
{"(  aggregate\n      MAX(distinct 1)\n)":error({message:"?-operator: bool predicate required",on:error({message:"len: bad type",on:1})})}
```

## Details

Repro is with super commit bec8934. This is a simplification of an issue surfaced by [a sqllogictest](https://github.com/dolthub/sqllogictest/blob/a1f4aa3c1d691552a4a03d9603c45bcf93d0bb9d/test/random/aggregates/slt_good_2.test#L127-L131).

It works without the extra parens.

```
$ super -version
Version: bec8934e0

$ super -c "SELECT MAX (DISTINCT 1);"
{MAX:1}
```

And both variations work the same in Postgres.

```
$ psql --version

$ psql postgres -c "SELECT MAX (DISTINCT 1);"
 max 
-----
   1
(1 row)

$ psql postgres -c "SELECT (MAX (DISTINCT 1));"
 max 
-----
   1
(1 row)
```

### LINKED ISSUE #6430: SQL: Parse error from CAST+MAX+ALL combination
## tl;dr

Despite the `ALL` support added in #6258, the following produces a parse error.

```
$ super -c "SELECT CAST(MAX(ALL 6) AS INTEGER);"
parse error at line 1, column 21:
SELECT CAST(MAX(ALL 6) AS INTEGER);
                === ^ ===
```

## Details

Repro is with super commit 30f64bb. This is a simplification of a query from [a sqllogictest](https://github.com/dolthub/sqllogictest/blob/a1f4aa3c1d691552a4a03d9603c45bcf93d0bb9d/test/random/expr/slt_good_61.test#L63662-L63668) (aka `random/expr/slt_good_61/q10764`).

Here's the query returning the expected result in Postgres.

```
$ psql postgres -c "SELECT CAST(MAX(ALL 6) AS INTEGER);"
 max 
-----
   6
(1 row)
```

It does parse in `super` if the `ALL` is dropped.

```
$ super -version
Version: 30f64bbac

$ super -c "SELECT CAST(MAX(6) AS INTEGER);"
{"CAST(MAX(6) AS int32)":6::int32}
```

mccanne:
This commit changes the "where <expr>" clause on agg functions to use the SQL syntax "filter ( <expr> )".

While we were here, we changed ast.Agg to ast.AggFunc, removed the Where field from ast.CallExpr, and cleaned up the grammar for how agg functions are integrated into expressions and shortcuts.

We also fixed a big where "count(*) where ..." did not previously parse, but "count(*) filter ( ... )" now works.

Fixes #6261
Fixes #6330
Fixes #6430

### DIFF
--- a/compiler/ast/op.go
+++ b/compiler/ast/op.go
@@ -376,12 +376,3 @@ func (*ValuesOp) opNode()     {}
 func (*WhereOp) opNode()      {}
 
 func (*DefaultScan) opNode() {}
-
-type AggFunc struct {
-	Kind     string `json:"kind" unpack:""`
-	Name     string `json:"name"`
-	Distinct bool   `json:"distinct"`
-	Expr     Expr   `json:"expr"`
-	Filter   Expr   `json:"filter"`
-	Loc      `json:"loc"`
-}
--- a/compiler/parser/parser.peg
+++ b/compiler/parser/parser.peg
@@ -267,8 +267,8 @@ AggAssignment
 
 AggFunc
   = name:AggName __ "(" __ expr:Expr? __ ")" filter:FilterClause? {
-      agg := &ast.AggFunc{
-          Kind: "AggFunc",
+      agg := &ast.AggFuncExpr{
+          Kind: "AggFuncExpr",
           Name: name.(string),
           Loc: loc(c),
       }
@@ -282,8 +282,8 @@ AggFunc
     }
   / name:AggName __ "(" __ q:(ALL / DISTINCT)  _ expr:Expr __ ")" filter:FilterClause? {
       qualifier := string(q.([]any)[0].([]byte))
-      agg := &ast.AggFunc{
-          Kind: "AggFunc",
+      agg := &ast.AggFuncExpr{
+          Kind: "AggFuncExpr",
           Name: name.(string),
           Distinct: strings.EqualFold(qualifier, "distinct"),
           Expr: expr.(ast.Expr),
@@ -295,8 +295,8 @@ AggFunc
       return agg, nil    
     }
   / cs:CountStar filter:FilterClause? {
-      agg := &ast.AggFunc{
-          Kind: "AggFunc",
+      agg := &ast.AggFuncExpr{
+          Kind: "AggFuncExpr",
           Name: "count",
           Loc: loc(c),
       }
@@ -312,8 +312,6 @@ AggName
   / OR
 
 FilterClause = _ FILTER __ "(" __ expr:LogicalOrExpr __ ")" { return expr, nil }
-
-WhereClause = _ WHERE _ expr:LogicalOrExpr { return expr, nil }
 
 AggAssignments
   = first:AggAssignment rest:(__ "," __ AggAssignment)* {
@@ -2099,6 +2097,8 @@ FromSelect
       }
       return sel, nil
     }
+
+WhereClause = _ WHERE _ expr:LogicalOrExpr { return expr, nil }
 
 SQLValues
   = VALUES __ tuples:SQLTuples {
--- a/compiler/parser/support.go
+++ b/compiler/parser/support.go
@@ -58,13 +58,12 @@ func makeArgMap(args any) (any, error) {
 }
 
 func newCall(c *current, name, args any) ast.Expr {
-	call := &ast.CallExpr{
+	return &ast.CallExpr{
 		Kind: "CallExpr",
 		Func: name.(ast.Expr),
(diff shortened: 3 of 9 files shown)


compiler/ast/op.go
```diff
@@ -377,17 +377,11 @@ func (*WhereOp) opNode()      {}
...
 func (*DefaultScan) opNode() {}
 
-// An Agg is an AST node that represents a aggregate function.  The Name
-// field indicates the aggregation method while the Expr field indicates
-// an expression applied to the incoming records that is operated upon by them
-// aggregate function.  If Expr isn't present, then the aggregator doesn't act
-// upon a function of the record, e.g., count() counts up records without
-// looking into them.
-type Agg struct {
+type AggFunc struct {
```
nwt: Nit: While you're here, maybe move this to expr.go?

compiler/parser/parser.peg
```diff
@@ -258,61 +258,61 @@ LimitArg
...
     }
 
 AggName
   = IdentifierName
   / AND
   / OR
 
+FilterClause = _ FILTER __ "(" __ expr:LogicalOrExpr __ ")" { return expr, nil }
+
 WhereClause = _ WHERE _ expr:LogicalOrExpr { return expr, nil }
```
nwt: Nit: Since this is now used only in FromSelect and Select, maybe move it down where OptWhereClause used to live?

compiler/parser/support.go
```diff
@@ -57,16 +57,13 @@ func makeArgMap(args any) (any, error) {
 	return m, nil
 }
 
-func newCall(c *current, name, args, where any) ast.Expr {
+func newCall(c *current, name, args any) ast.Expr {
 	call := &ast.CallExpr{
```
nwt: Nit:

```suggestion
	return &ast.CallExpr{
```

compiler/semantic/op.go
```diff
@@ -1404,6 +1406,23 @@ func (t *translator) callExpr(call *ast.CallExpr, seq sem.Seq) sem.Seq {
...
+		Aggs: []sem.Assignment{
+			{
+				Node: aggFunc,
+				LHS:  sem.NewThis(agg, []string{name}),
+				RHS:  aggFunc,
+			},
+		},
+	}
+	values := sem.NewValues(agg, sem.NewThis(agg, []string{name}))
+	return append(append(seq, aggregate), values)
```
nwt: Nit:

```suggestion
	return append(seq, aggregate, values)
```

compiler/ast/op.go
mccanne: Also gonna call it `AggFuncExpr`

PR #6471, opened 2025-12-29
vector.Not: Set true/null to false/null

### LINKED ISSUE #6431: SQL: Incorrect result from OR+BETWEEN combo with NULL value in vector runtime
The following query returns a result in vector runtime but should return nothing.

```
$ super -f parquet -o data.parquet -c "values {d:null,e:15}" &&
SUPER_VAM=1 super -c "
SELECT d,e
FROM data.parquet
WHERE d>e OR d NOT BETWEEN 10 AND 20;"

{d:null,e:15}
```
## Details

Repro is with super commit 30f64bb. The query above is a simplification of one from [this sqllogictest](https://github.com/dolthub/sqllogictest/blob/a1f4aa3c1d691552a4a03d9603c45bcf93d0bb9d/test/select2.test#L4519-L4528) (aka `select2/q394`).

This query has characteristics in common with the one from #6277 in that it seems to be unique to `NULL` values being subject to comparisons, but for some reason was not fixed by the changes in #6426.

Interestingly, if I do an a la carte `WHERE` of what's on each end of the `OR`, they each return the expected "nothing".

```
$ super -f parquet -o data.parquet -c "values {d:null,e:15}" &&
SUPER_VAM=1 super -c "
SELECT d,e
FROM data.parquet
WHERE d>e;"

[no output]
```

```
$ super -f parquet -o data.parquet -c "values {d:null,e:15}" &&
SUPER_VAM=1 super -c "
SELECT d,e
FROM data.parquet
WHERE d NOT BETWEEN 10 AND 20;"

[no output]
```

And here it is in sequential runtime and DuckDB with the full `OR`'d together `WHERE` clause giving the expected "nothing" response.

```
$ super -f parquet -o data.parquet -c "values {d:null,e:15}" &&
super -c "
SELECT d,e
FROM data.parquet
WHERE d>e OR d NOT BETWEEN 10 AND 20;"

[no output]
```

```
$ duckdb
DuckDB v1.4.2 (Andium) 68d7555f68
D SELECT d,e
  FROM data.parquet
  WHERE d>e OR d NOT BETWEEN 10 AND 20;
┌───────┬───────┐
│   d   │   e   │
│ int32 │ int64 │
├───────┴───────┤
│    0 rows     │
└───────────────┘
```

mattnibs:
Closes #6431

### DIFF
--- a/runtime/ztests/expr/logical-not.yaml
+++ b/runtime/ztests/expr/logical-not.yaml
@@ -15,3 +15,21 @@ output: |
   null::bool
   error("missing")
   error("foo")
+
+---
+
+# Ensure true/null values are converted to false/null.
+spq: |
+  values false or not this
+
+vector: true
+
+input: |
+  true
+  false
+  null::bool
+
+output: |
+  false
+  true
+  null::bool
--- a/vector/bool.go
+++ b/vector/bool.go
@@ -64,6 +64,16 @@ func Or(a, b *Bool) *Bool {
 	return NewBool(bits, nulls)
 }
 
+func Not(vec *Bool) *Bool {
+	bits := bitvec.Not(vec.Bits)
+	if vec.Nulls.IsZero() {
+		return NewBool(bits, bitvec.Zero)
+	}
+	// Flip true/null values to false/null.
+	bits = bitvec.And(bits, bitvec.Not(vec.Nulls))
+	return NewBool(bits, vec.Nulls)
+}
+
 // BoolValue returns the value of slot in vec if the value is a Boolean.  It
 // returns false otherwise.
 func BoolValue(vec Any, slot uint32) (bool, bool) {
--- a/runtime/vam/expr/logic.go
+++ b/runtime/vam/expr/logic.go
@@ -27,8 +27,11 @@ func (n *Not) Eval(val vector.Any) vector.Any {
 func (n *Not) eval(vecs ...vector.Any) vector.Any {
 	switch vec := vecs[0].(type) {
 	case *vector.Bool:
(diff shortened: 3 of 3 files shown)


runtime/ztests/expr/logical-not.yaml
```diff
@@ -15,3 +15,20 @@ output: |
   null::bool
   error("missing")
   error("foo")
+
+---
+# Ensure true/null values are converted to false/null.
```
nwt, line 21: Nit: I think we should establish a convention of surrounding the document separator (`---`) with single blank lines.

```suggestion

---

# Ensure true/null values are converted to false/null.
```

vector/bool.go
```diff
@@ -64,6 +64,16 @@ func Or(a, b *Bool) *Bool {
 	return NewBool(bits, nulls)
 }
 
+func Not(vec *Bool) *Bool {
+	bits := bitvec.Not(vec.Bits)
+	if vec.Nulls.IsZero() {
+		return NewBool(bits, bitvec.Zero)
+	}
+	// Flip true/null values to false/null.
+	bits = bitvec.And(bits, bitvec.Not(bitvec.And(bits, vec.Nulls)))
```
nwt: Nit: No need for two logical conjunctions.
```
A & ^(A & B) = A & (^A | ^B)        By De Morgan's law.
             = (A & ^A) | (A & ^B)  By the distributive property.
             = A & ^B               Because A & ^A is always false.
```

```suggestion
	bits = bitvec.And(bits, bitvec.Not(vec.Nulls))
```

nwt (approved):

vector/bool.go
mattnibs: Thanks!

PR #6475, opened 2025-12-29
vector: report const as null if base value is null

mattnibs:
When getting the value of a primitive const at a given slot, report the value as null if the base value is null.

### DIFF
--- a/vector/bool.go
+++ b/vector/bool.go
@@ -81,7 +81,7 @@ func BoolValue(vec Any, slot uint32) (bool, bool) {
 	case *Bool:
 		return vec.Bits.IsSet(slot), vec.Nulls.IsSet(slot)
 	case *Const:
-		return vec.Value().Ptr().AsBool(), vec.Nulls.IsSet(slot)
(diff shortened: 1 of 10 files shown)


vector/bool.go
```diff
@@ -81,7 +81,7 @@ func BoolValue(vec Any, slot uint32) (bool, bool) {
 	case *Bool:
 		return vec.Bits.IsSet(slot), vec.Nulls.IsSet(slot)
 	case *Const:
-		return vec.Value().Ptr().AsBool(), vec.Nulls.IsSet(slot)
+		return vec.Value().Ptr().AsBool(), vec.val.IsNull() || vec.Nulls.IsSet(slot)
```
nwt, line 84: I really think this is the wrong approach and that the nulls vector should be the source of truth, with the implication that a vector.Const with a null super.Value should always have every bit set in its nulls vector.

mattnibs: I don't feel strongly about this but I'd like to make sure this works someday in the near future. I can't assess whether what you are proposing is a quick fix.

nwt: Check out #6477.

Issue #6402, opened 2025-11-28
Dropping "text" and "table" output formats


philrz:
Because of their multiple problems, we've reached consensus that the the `text` and `table` output formats should be dropped. However, as users have been known to reach for these features, it would be ideal to cover some functional gaps in alternate formats to allow for creating equivalent/better output (#5961, #6381).

## Details

At the time this issue is being opened, super is at commit a3f7d15.

Some specific shortcomings of `text` and `table` outputs are described in #4023 and #4595. Because of these problems, we initially gave much more coverage in docs/examples to other output formats we believed were superior. However some users apparently were seeking the simpler functionality implied by their names, as several discovered them via help text, experienced their glitches, then came to community Slack asking for help/fixes. This led to my attempt to document their uses/limitations via docs so I'd have somewhere to point users each time this topic came up, and these are still visible in the old Zed docs [here](https://zed.brimdata.io/docs/commands/zq#simplified-text-outputs). But since they're still difficult to use even when aware of those limitations and their flaws were seen as difficult to fix, they've been dropped again from the most recent docs and so we've decided to drop them from the tools as well.

Thankfully, we do have alternate formats that cover _most_ of what these formats could do. Specifically:

1. `line` mode is useful for printing values without formatting, e.g., strings without quotes, similar to `text`. One limitation of `line` mode is that records are printed in SUP format whereas `line` would print just values, but combining `line` mode with [f-strings](https://superdb.org/super-sql/expressions/f-strings.html) can achieve this. We've also considered adding `sprintf`-like functionality (#3147) to provide more options here.

   ```
   $ super -version
   Version: a3f7d1558
   
   $ echo '{foo: "bar"}' | super -f text -
   bar
   
   $ echo '{foo: "bar"}' | super -f line -
   {foo:"bar"}
   
   $ echo '{foo: "bar"}' | super -f line -c 'f"{foo}"' -
   bar
   ```

2. `csv` or `tsv` modes offer a columnar output similar in nature to `table`, with the advantage that they steer users toward `fuse` to avoid the problem described in #4023.

[...]

PR #6483, opened 2026-01-01
fix: add separator between tables in table output

### LINKED ISSUE #4023: Lack of headers makes table output confusing
A community user reported the following via private gist. Repro is with Zed commit 116f5c0.

The user reports:

> The output from the command below was initially confusing. Maybe put a --- between when printing multiple tables.
> ```
> $ echo '{s:"foo", val:1}{s:"bar"}' | zq -f table 'cut s, val' -
> s   val
> foo 1
> s   val
> bar -

majiayu000:
Fixes #4023

## Changes
- Add `---` separator between tables when record type changes
- Add test for multiple types in table output

### DIFF
--- a/sio/tableio/writer.go
+++ b/sio/tableio/writer.go
@@ -43,6 +43,7 @@ func (w *Writer) Write(r super.Value) error {
 	if r.Type() != w.typ {
 		if w.typ != nil {
 			w.flush()
+			w.writer.Write([]byte("---\n"))
 			w.nline = 0
 		}
 		// First time, or new descriptor, print header
--- a/sio/tableio/ztests/multiple-types.yaml
+++ b/sio/tableio/ztests/multiple-types.yaml
@@ -1,0 +1,14 @@
+spq: pass
+
+input: |
+  {s:"foo",val:1}
+  {s:"bar"}
+
+output-flags: -f table
+
+output: |
+  s   val
+  foo 1
+  ---
+  s
+  bar


philrz:
@majiayu000: Thanks for your contribution and interest in the project!

While it's true that `table` output is still present in `super` at the moment, per #6402, due to problems with the feature beyond the one you're addressing here, our current plan is to drop support for it as an output format and instead encourage the use of other formats, such as `tsv` combined with the use of the [fuse operator](https://superdb.org/super-sql/operators/fuse.html) if needed to deal with multiple record types. Could you help us understand if there's something about your use case for `super` that would make the `table` output in its current format still preferable to `tsv`+`fuse`?

majiayu000:
Thanks for the context. If table output is planned for removal, I’m happy to close this PR. If it helps users, I can instead add docs or a hint that suggests  when multiple record types are present. Let me know what you prefer.

philrz:
@majiayu000: FWIW, there's actually a hint present in the tools already. For instance, because of the multiple record types, if I try to print the input data from your PR in CSV or TSV format we see partial output, then the hint once the second record type is encountered:

```
$ echo '{s:"foo",val:1} {s:"bar"}' | super -f tsv -
s	val
foo	1
```

If a new user is seeing this, we're hopeful this will inspire them to look at the [fuse operator docs](https://superdb.org/super-sql/operators/fuse.html) at which point it should be clear they can do:

```
$ echo '{s:"foo",val:1} {s:"bar"}' | super -f tsv -c "fuse" -
s	val
foo	1
bar	
```

I do suspect that at some point we may add a proper table/box style output format (e.g., something like [duckbox](https://duckdb.org/docs/stable/guides/snippets/importing_duckbox_tables)) since most other SQL tools have that and so I suspect users may start asking about it. But the format we've been calling `table` has had too many problems to be fixable in its prior form, so we figure taking it out of the tools entirely will help us establish demand and get user input for a design of a better replacement.

Let us know if you have any questions, and thanks again for your interest in the project!

PR #6486, opened 2026-01-05
Add concat function, drop "+" for string concatenation

### LINKED ISSUE #6347: SQL: Add CONCAT function, drop "+" for string concatenation
Right now `super` supports the SQL standard `||` for string concatenation but doesn't support the common SQL `CONCAT` function. `super` also has support for doing concatenation via `+`. Consensus is to add support for `CONCAT` and drop it for `+`. There's also a bug with `||` currently.

## Details

Repro is with super commit d388d44.

Here's how this all looks in Postgres.

```
$ psql postgres
psql (17.6 (Homebrew))
Type "help" for help.

postgres=# CREATE TABLE data (a VARCHAR, b VARCHAR);
CREATE TABLE
postgres=# INSERT INTO data (a, b) VALUES ('foo', 'bar');
INSERT 0 1

postgres=# SELECT a || b FROM data;
 ?column? 
----------
 foobar
(1 row)

postgres=# SELECT CONCAT(a, b) FROM data;
 concat 
--------
 foobar
(1 row)

postgres=# SELECT a+b FROM data;
ERROR:  operator does not exist: character varying + character varying
LINE 1: SELECT a+b FROM data;
                ^
HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.
```

And in `super`:

```
$ super -version
Version: d388d4472

$ super -f parquet -o data -c "values {a:'foo',b:'bar'}"

$ super -c "SELECT a || b FROM data;"
{"a||b":"foobar"}

$ super -c "SELECT CONCAT(a, b) FROM data;"
SELECT CONCAT(a, b) FROM data;
       ~~~~~~~~~~~~

$ super -c "SELECT a+b FROM data;"
{"a+b":"foobar"}
```

@mccanne also recently spotted that `||` is incorrectly doing mathematical addition when given numeric inputs.

```
$ super -version
Version: d388d4472

$ super -c "values 1 || 2"
3
```

mattnibs:
Closes #6347


nwt (approved):
Nicely done.

PR #6487, opened 2026-01-06
new approach to relational scoping

### LINKED ISSUE #5972: SQL: ORDER BY aggregation that does not appear in SELECT clause
The following query works in Postgres but not SuperDB.

```
SELECT b, SUM(a) FROM test GROUP BY b ORDER BY COUNT(a);
```

## Details

Repro is with super commit e75dfda. This was found via a query from [a sqllogictest](https://github.com/duckdb/duckdb/blob/0cbb56d14cc1464d945a06cb8bac051485fa2ef7/test/sql/aggregate/group/test_group_by.test#L58-L63).

Repro with test data:

```
$ super -version
Version: e75dfdaf1

$ cat test.sup
{a:11::int32,b:22::int32}
{a:13::int32,b:22::int32}
{a:12::int32,b:21::int32}

$ super -c "
SELECT b, SUM(a) FROM 'test.sup' GROUP BY b ORDER BY COUNT(a);"

column "a": does not exist at line 2, column 60:
SELECT b, SUM(a) FROM 'test.sup' GROUP BY b ORDER BY COUNT(a);
                                                           ~
```

Here it is working in Postgres:

```
$ psql postgres
psql (17.5 (Homebrew))
Type "help" for help.

postgres=# SELECT * FROM test;
 a  | b  
----+----
 11 | 22
 13 | 22
 12 | 21
(3 rows)

postgres=# SELECT b, SUM(a) FROM test GROUP BY b ORDER BY COUNT(a);
 b  | sum 
----+-----
 21 |  12
 22 |  24
(2 rows)
```

### LINKED ISSUE #5977: SQL: Duplicate column names in query results

### LINKED ISSUE #6088: SQL: SELECT DISTINCT ineffective when combined with GROUP BY
This query returns two values in SuperDB but (in line with the expectations of `SELECT DISTINCT`) only one value in Postgres.

```
$ super -f parquet -o test -c "values
  {a:11::int32,b:22::int32},
  {a:13::int32,b:22::int32},
  {a:11::int32,b:21::int32},
  {a:11::int32,b:22::int32}" &&
super -c "SELECT DISTINCT MAX(b) FROM test GROUP BY a;"

{MAX:22}
{MAX:22}
```

## Details

Repro is with super commit 1201d20. This was found via a query from [a sqllogictest](https://github.com/duckdb/duckdb/blob/459203413e86af5c4897c0c0cd58afb03a5840ac/test/sql/aggregate/distinct/test_distinct.test#L47-L49).

We'll start by prepping data and running queries in both systems _without_ using `DISTINCT`.

Postgres:

```
$ psql postgres
psql (17.5 (Homebrew))
Type "help" for help.

postgres=# CREATE TABLE test (a INTEGER, b INTEGER);
CREATE TABLE
postgres=# INSERT INTO test VALUES (11, 22), (13, 22), (11, 21), (11, 22);
INSERT 0 4
postgres=# SELECT MAX(b) FROM test GROUP BY a;
 max 
-----
  22
  22
(2 rows)
```

SuperDB:

```
$ super -version
Version: 1201d205e

$ super -c "SELECT MAX(b) FROM test GROUP BY a;"
{MAX:22}
{MAX:22}
```

Once we add the `DISTINCT`, we see that in Postgres the output is trimmed to a single value, whereas in SuperDB it's still two values.

```
postgres=# SELECT DISTINCT MAX(b) FROM test GROUP BY a;
 max 
-----
  22
(1 row)

$ super -c "SELECT DISTINCT MAX(b) FROM test GROUP BY a;"
{MAX:22}
{MAX:22}
```

### LINKED ISSUE #6227: SQL: "column not found" when Parquet schema is leveraged
This query works if I drop the use of `(FORMAT parquet)`.

```
$ super -f parquet -o test -c "
values
  {a:11,b:22},
  {a:13,b:22},
  {a:12,b:21}
" && super -c "SELECT test.b, SUM(a) FROM test (FORMAT parquet) GROUP BY test.b ORDER BY (SELECT SUM(a) FROM test (FORMAT parquet) t WHERE test.b=t.b) DESC;"

column "test" not found at line 1, column 125:
SELECT test.b, SUM(a) FROM test (FORMAT parquet) GROUP BY test.b ORDER BY (SELECT SUM(a) FROM test (FORMAT parquet) t WHERE test.b=t.b) DESC;
                                                                                                                            ~~~~~~
```

## Details

Repro is with super commit d0077f2. This was found via a query from [a sqllogictest](https://github.com/duckdb/duckdb/blob/15fba9f47c2c547d8a9fa2cdb23f91e5eb8ea533/test/sql/aggregate/having/test_having.test#L74-L79).

As noted above, if works if I drop the use of `(FORMAT parquet)`.

```
$ super -version
Version: d0077f2bf

$ super -f parquet -o test -c "
values
  {a:11,b:22},
  {a:13,b:22},
  {a:12,b:21}
" && super -c "SELECT test.b, SUM(a) FROM test GROUP BY test.b ORDER BY (SELECT SUM(a) FROM test t WHERE test.b=t.b) DESC;"

{b:22,"SUM(a)":24}
{b:21,"SUM(a)":12}
```

It also works with the `(FORMAT parquet)` if I go back to the commit before #6097 was merged, so the problem seems unique to that Parquet schema support.

```
$ super -version
Version: 97818e43e

$ super -f parquet -o test -c "
values
  {a:11,b:22},
  {a:13,b:22},
  {a:12,b:21}
" && super -c "SELECT test.b, SUM(a) FROM test (FORMAT parquet) GROUP BY test.b ORDER BY (SELECT SUM(a) FROM test (FORMAT parquet) t WHERE test.b=t.b) DESC;"

{b:22,"SUM(a)":24}
{b:21,"SUM(a)":12}
```

### LINKED ISSUE #6250: SQL: aggregate mixed with *-selector not yet supported
```
$ super -f parquet -o data.parquet -c "values {a:1,b:2},{a:3, b:4},{a:1, b:2}" &&
  super -c "SELECT * FROM data.parquet GROUP BY a, b"

aggregate mixed with *-selector not yet supported at line 1, column 1:
SELECT * FROM data.parquet GROUP BY a, b
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

## Details

Repro is with super commit c6ebc21. This issue was found via [a sqllogictest](https://github.com/dolthub/sqllogictest/blob/a1f4aa3c1d691552a4a03d9603c45bcf93d0bb9d/test/random/groupby/slt_good_0.test#L119-L122).

I learned this is a somewhat rare combination, but when working as intended, one way it may be used is to get distinct rows (i.e., the equivalent of a `SELECT DISTINCT`). Here it is in Postgres doing just that with the input data.

```
$ psql postgres
psql (17.6 (Homebrew))
Type "help" for help.

postgres=# CREATE TABLE data (a INTEGER, b INTEGER);
CREATE TABLE

postgres=# INSERT INTO data VALUES (1, 2), (3, 4), (1, 2);
INSERT 0 3

postgres=# SELECT * FROM data GROUP BY a, b;
 a | b 
---+---
 3 | 4
 1 | 2
(2 rows)
```

For that specifically, `SELECT DISTINCT` is indeed already available in `super`.

```
$ super -version
Version: c6ebc2139

$ super -c "SELECT DISTINCT * FROM data.parquet"
{a:1,b:2}
{a:3,b:4}
```

### LINKED ISSUE #6403: SQL: Unable to extract record field from subquery result using dot notation
The following query fails with `super`, but succeeds in other SQL implementations.

```
$ super -version
Version: a3f7d1558

$ super -c "SELECT t.a FROM (VALUES({a: 1})) AS t(t)"
table "t": column "a": does not exist at line 1, column 8:
SELECT t.a FROM (VALUES({a: 1})) AS t(t)
       ~~~
```

## Details

Repro is with super commit a3f7d15.

Here's the expected result for the same query as seen with DuckDB.

```
$ duckdb
DuckDB v1.4.2 (Andium) 68d7555f68
D SELECT t.a FROM (VALUES({a: 1})) AS t(t);
┌───────┐
│   a   │
│ int32 │
├───────┤
│   1   │
└───────┘
```

### LINKED ISSUE #6404: SQL: Improve error message for duplicate aliases
The error message from the following attempted query does not accurately describe what's wrong with it.

```
$ super -f parquet -o a.parquet -c "values {a:1},{a:2}" &&
  super -f parquet -o b.parquet -c "values {b:1},{b:2}" &&
  super -c "select t from a.parquet t cross join b.parquet t"

"t": not found (column "t": does not exist, column "t": does not exist) at line 1, column 8:
select t from a.parquet t cross join b.parquet t
       ~
```

## Details

Repro is with super commit a3f7d15.

The query above is a variation on this other query that works fine since the table aliases `t` and `u` do not conflict.

```
$ super -version
Version: a3f7d1558

$ super -c "select t from a.parquet t cross join b.parquet u"
{t:{a:1}}
{t:{a:1}}
{t:{a:2}}
{t:{a:2}}
```

And here's DuckDB giving a more helpful error message on the original query that contains the duplicate aliases for `t`.

```
$ duckdb
DuckDB v1.4.2 (Andium) 68d7555f68
D select t from a.parquet t cross join b.parquet t;
Binder Error:
Ambiguous reference to table "t" (duplicate alias "t", explicitly alias one of the tables using "AS my_alias")
```

mccanne:
This commit makes a big change to how relational scoping is implemented better aligning the design with SQL semantics.  In this approach, SQL operators always produce tables even if the input is not tabular.

Lateral column alias resolution is now performed statically and semantically defined as substitution of input-scoped expressions corresponding to the column alias.  To ensure that scope bindings do not depend on input data, we chose to prioritize lateral column binding over input table binding thus deviating from Postgres semantics in favor of Google SQL semantics.

The select-* pattern no longer works with dynamic inputs and instead matches all columns in the input schema.  Also, table.* pattern matching now works.  We also fixed join with the USING clause to limit the USING expressions to non-dotted identifiers and to deduplicate the USING columns and place them in the leftmost column position.

Table and column scope bindings are now case insensitive so that "select A" works on column "a" and "select t.a" works on table "T", etc.

Redundant agg functions are now detected and computed only once.  Also, expressions to ORDER BY can correctly reach into the SELECT scope and refer to aggregate functions, grouping expressions, input table columns, and so forth.

We've updated the pipe join docs here to reflect the change to USING but more complete doc updates for relational will be added in a subsequent PR.

Fixes #5972
Fixes #6088
Fixes #6227
Fixes #6250
Fixes #6403
Fixes #6404
Fixes #5977


philrz:
I started a run of the sqllogictests that had been passing on tip of `main` to see how they fare on this branch. So far I'm seeing a few hundred failures similar to this repro based on [random/groupby/slt_good_11/q9885.yaml](https://github.com/brimdata/sqllogic-ztests/blob/main/ztests/sqlite/random/groupby/slt_good_11/ztests-1/q9885.yaml):

```
$ cat tab1.csv
col0,col1,col2
22,6,8
28,57,45
82,44,71

$ super -version
Version: dbe0a949c

$ super -c "SELECT + 99 AS col2 FROM tab1.csv AS cor0 GROUP BY col1, col2;"
position 99 is not in select list at line 1, column 58:
SELECT + 99 AS col2 FROM tab1.csv AS cor0 GROUP BY col1, col2;
                                                         ~~~~
```

The query does contain the kind of ambiguity that I know matters here. But in other situations the ambiguous queries run while producing a different results, so these stood out because they refuse to run at all.

On tip of `main`, this had been returning the same result we get form Postgres and ZetaSQL.

```
$ super -version
Version: b6e73785c

$ super -c "SELECT + 99 AS col2 FROM tab1.csv AS cor0 GROUP BY col1, col2;"
{col2:99}
{col2:99}
{col2:99}

$ psql postgres -c "SELECT + 99 AS col2 FROM tab1 AS cor0 GROUP BY col1, col2;"
 col2 
------
   99
   99
   99
(3 rows)

$ execute_query_macos --table_spec StringTab1=csv:tab1.csv "
WITH tab1 AS (
  SELECT
    CAST (col0 AS INT64) as col0,
    CAST (col1 AS INT64) as col1,
    CAST (col2 AS INT64) as col2
  FROM StringTab1
)
SELECT + 99 AS col2 FROM tab1 AS cor0 GROUP BY col1, col2;"
+------+
| col2 |
+------+
| 99   |
| 99   |
| 99   |
+------+
```

---

Update: Nice! Fix looks good.

```
$ super -version
Version: c93d852db

$ super -c "SELECT + 99 AS col2 FROM tab1.csv AS cor0 GROUP BY col1, col2;"
{col2:99}
{col2:99}
{col2:99}
```

Will start the run over at the newer commit and let you know if anything surfaces.

compiler/semantic/projection.go
```diff
@@ -3,108 +3,49 @@ package semantic
...
+}
+
+func exprMatch(target sem.Expr, exprs []exprloc) int {
+	for which, e := range exprs {
+		if eqExpr(target, e.expr) {
+			return which
+		}
+	}
+	return -1
 }
```
nwt, line 47: Nits: I think exprIndex would be a better name. And the conventional signature for index function has the target parameter after the slice parameter. And simplify with slices.IndexFunc.

```suggestion
func exprIndex(exprs []exprloc, target sem.Expr) int {
	return slices.IndexFunc(exprs, func(e exprloc) bool {
		return eqExpr(target, e.expr)
	})
}
```

compiler/semantic/schema.go
```diff
@@ -14,44 +13,85 @@ import (
...
+
+		// grouped is set when processing an expression that should resolve
+		// "this" to the aggregate projection (i.e., "g" column)
+		grouped bool
+
+		// aggs is the set of agg functions computed in a grouped table
+		// inclusive of agg functions used subsequently in an order by
+		aggs []*sem.AggFunc
+
+		// groupings hold the grouping expression which are needed to do
```
```suggestion
		// groupings hold the grouping expressions which are needed to do
```

compiler/semantic/schema.go
```diff
@@ -14,44 +13,85 @@ import (
...
+		// expression matching for having, order by, and the projection.
+		// These expressions are always relative to the "in" scope
+		// grouped is true when resolution has access only to the outputs
```
```suggestion
		// These expressions are always relative to the "in" scope.
		// The grouped field is true when resolution has access only to the outputs.
```

compiler/semantic/projection.go
mattnibs: I like exprMatch better, exprIndex sounds like something to do with index expressions.

compiler/ztests/sql/subqueries-correlated.yaml
```diff
@@ -47,6 +47,6 @@ outputs:
       correlated subqueries not currently supported at line 5, column 38:
                                 where colB=colA)
                                            ~~~~
-      column "colC" not found at line 5, column 38:
+      "colC": not a column or table at line 5, column 38:
```
nwt: Nit: I think this would read better as `"colC" is not a column or table`.

compiler/semantic/sql.go
```diff
@@ -640,113 +660,61 @@ type exprloc struct {
...
 	return out
 }
 
-func isOrdinal(sctx *super.Context, e sem.Expr) (int, bool) {
-	if literal, ok := e.(*sem.LiteralExpr); ok {
-		v := sup.MustParseValue(sctx, literal.Value)
-		return int(v.AsInt()), super.IsInteger(v.Type().ID())
+func isOrdinal(e ast.Expr) (int, bool) {
+	if e, ok := e.(*ast.Primitive); ok && e.Type == "int64" {
+		colno, _ := strconv.Atoi(e.Text)
```
```suggestion
		colno, err := strconv.Atoi(e.Text)
		if err != nil {
			panic(err)
		}
```

compiler/semantic/projection.go
mccanne: `slices.IndexFunc` is improvement but the name change is "what" vs "how" and I agree with @mattnibs

compiler/semantic/schema.go
```diff
@@ -89,258 +136,277 @@ func recordOf(typ super.Type) *super.TypeRecord {
...
 }
 
-func (j *joinUsingSchema) resolveColumn(col string) (field.Path, bool, error) {
-	if _, _, err := j.left.resolveColumn(col); err != nil {
-		return nil, false, fmt.Errorf("column %q in USING clause does not exist in left table", col)
-	}
-	if _, _, err := j.right.resolveColumn(col); err != nil {
-		return nil, false, fmt.Errorf("column %q in USING clause does not exist in right table", col)
+func (j *joinUsingSchema) resolveUnqualified(col string) (field.Path, bool, error) {
+	if slices.Contains(j.columns, col) {
```
nwt, line 277: Should this case-insensitive?

nwt:
Do you want to improve these errors?

```console
$ go run ./cmd/super -c 'select * from (values (1,2)) T1(a,b) join (values (1,3)) T2(A,B) using (a, b) group by a, b'
column "left" must appear in GROUP BY clause at line 1, column 8:
select * from (values (1,2)) T1(a,b) join (values (1,3)) T2(A,B) using (a, b) group by a, b
       ~
column "left" must appear in GROUP BY clause at line 1, column 8:
select * from (values (1,2)) T1(a,b) join (values (1,3)) T2(A,B) using (a, b) group by a, b
       ~
column "right" must appear in GROUP BY clause at line 1, column 8:
select * from (values (1,2)) T1(a,b) join (values (1,3)) T2(A,B) using (a, b) group by a, b
       ~
column "right" must appear in GROUP BY clause at line 1, column 8:
select * from (values (1,2)) T1(a,b) join (values (1,3)) T2(A,B) using (a, b) group by a, b
       ~
exit status 1
```

compiler/semantic/schema.go
mccanne: I'll back out case folding as we discussed

nwt (approved):
I think this looks good overall!

PR #6488, opened 2026-01-06
vam: Fix null+null arithmetic

### LINKED ISSUE #6484: SQL: Vector runtime errors on arithmetic with CASE expression returning NULL
These all produce unexpected error values in vector runtime.

```
$ SUPER_VAM=1 super -c "SELECT CASE 1 WHEN 0 THEN 1 END + NULL;"
{"case 1 when 0 then 1 end+":error("type null incompatible with '+' operator")}

$ SUPER_VAM=1 super -c "SELECT CASE 1 WHEN 0 THEN 1 END * NULL;"
{"case 1 when 0 then 1 end*":error("type null incompatible with '*' operator")}

$ SUPER_VAM=1 super -c "SELECT CASE 1 WHEN 0 THEN 1 END / NULL;"
{"case 1 when 0 then 1 end/":error("type null incompatible with '/' operator")}
```

## Details

Repro is with super commit 5ea0cb5. The queries shown above are simplifications from sqllogictests that produce the same errors, such as [random/expr/slt_good_117/q1055](https://github.com/brimdata/sqllogic-ztests/blob/main/ztests/sqlite/random/expr/slt_good_117/ztests-1/q1055.yaml) (addition), [random/expr/slt_good_102/q2684](https://github.com/brimdata/sqllogic-ztests/blob/main/ztests/sqlite/random/expr/slt_good_102/ztests-1/q2684.yaml) (multiplication), and [random/expr/slt_good_105/q9770](https://github.com/brimdata/sqllogic-ztests/blob/main/ztests/sqlite/random/expr/slt_good_105/ztests-1/q9770.yaml) (division).

Here's the expected results being returned in sequential runtime and Postgres.

```
$ super -version
Version: 5ea0cb5d6

$ super -c "SELECT CASE 1 WHEN 0 THEN 1 END + NULL;"
{"case 1 when 0 then 1 end+":null}

$ super -c "SELECT CASE 1 WHEN 0 THEN 1 END * NULL;"
{"case 1 when 0 then 1 end*":null}

$ super -c "SELECT CASE 1 WHEN 0 THEN 1 END / NULL;"
{"case 1 when 0 then 1 end/":null}
```

```
$ psql postgres
psql (17.7 (Homebrew))
Type "help" for help.

postgres=# \pset null 'NULL'
Null display is "NULL".
postgres=# SELECT CASE 1 WHEN 0 THEN 1 END + NULL;
 ?column? 
----------
     NULL
(1 row)

postgres=# SELECT CASE 1 WHEN 0 THEN 1 END * NULL;
 ?column? 
----------
     NULL
(1 row)

postgres=# SELECT CASE 1 WHEN 0 THEN 1 END / NULL;
 ?column? 
----------
     NULL
(1 row)
```

mattnibs:
Fixes #6484

### DIFF
--- a/runtime/vam/expr/arith.go
+++ b/runtime/vam/expr/arith.go
@@ -31,10 +31,10 @@ func (a *Arith) Eval(val vector.Any) vector.Any {
 func (a *Arith) eval(vecs ...vector.Any) (out vector.Any) {
 	lhs := enumToIndex(vector.Under(vecs[0]))
 	rhs := enumToIndex(vector.Under(vecs[1]))
-	if vector.KindOf(lhs) == vector.KindError {
+	if k := vector.KindOf(lhs); k == vector.KindNull || k == vector.KindError {
 		return lhs
 	}
-	if vector.KindOf(rhs) == vector.KindError {
+	if k := vector.KindOf(rhs); k == vector.KindNull || k == vector.KindError {
 		return rhs
 	}
 	lhs, rhs, errVal := coerceVals(a.sctx, lhs, rhs)
--- a/runtime/ztests/expr/arith-nulls.yaml
+++ b/runtime/ztests/expr/arith-nulls.yaml
@@ -6,6 +6,7 @@ input: |
   {a:1,b:null::int64}
   {a:null::int64,b:1}
   {a:null::uint64,b:2}
+  {a:null,b:null}
 
 output: |
   null::int64
@@ -23,3 +24,8 @@ output: |
   null::int64
   null::int64
   null::int64
+  null
+  null
+  null
+  null
+  null
--- a/vector/kind.go
+++ b/vector/kind.go
@@ -19,11 +19,12 @@ const (
 	KindIP      = 6
 	KindNet     = 7
 	KindType    = 8
-	KindError   = 9
-	KindArray   = 10
-	KindSet     = 11
-	KindMap     = 12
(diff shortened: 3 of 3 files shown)


runtime/vam/expr/arith.go
```diff
@@ -48,6 +48,9 @@ func (a *Arith) eval(vecs ...vector.Any) (out vector.Any) {
 	if kind == vector.KindFloat && a.opCode == vector.ArithMod {
 		return vector.NewStringError(a.sctx, "type float64 incompatible with '%' operator", lhs.Len())
 	}
+	if kind == vector.KindNull {
+		return vector.NewConst(super.Null, lhs.Len(), bitvec.NewTrue(lhs.Len()))
+	}
```
nwt: Nit: Checking for KindNull earlier, when we check for KindError. That way we don't have to call vector.NewConst twice (once in coerceVals and then again here).

```go
	if k := vector.KindOf(lhs); k == vector.KindNull || k == vector.KindError {
		return lhs
	}
	if k := vector.KindOf(rhs); k == vector.KindNull || k == vector.KindError {
		return rhs
	}
```

nwt (approved):

runtime/vam/expr/arith.go
mattnibs: Nice!

PR #6489, opened 2026-01-06
vam: Fix toBool handling of Const with nulls

### LINKED ISSUE #6482: SQL: Vector runtime evaluation of "(NOT NULL = NULL) OR FALSE"
Vector runtime gives an unexpected output value here:

```
$ SUPER_VAM=1 super -c "
SELECT * 
FROM (VALUES (1)) AS t(col0) 
WHERE (NOT NULL = NULL) OR FALSE;"

{col0:1}
```

## Details

Repro is with super commit 5ea0cb5. The query above is a simplification of the one from sqllogictest [random/groupby/slt_good_10/q1452](https://github.com/brimdata/sqllogic-ztests/blob/main/ztests/sqlite/random/groupby/slt_good_10/ztests-1/q1452.yaml).

Sequential runtime gives the expected result of "no output":

```
$ super -version
Version: 5ea0cb5d6

$ super -c "
SELECT * 
FROM (VALUES (1)) AS t(col0) 
WHERE (NOT NULL = NULL) OR FALSE;"

[no output]
```

Postgres does the same.

```
$ psql postgres -c "
SELECT * 
FROM (VALUES (1)) AS t(col0) 
WHERE (NOT NULL = NULL) OR FALSE;"
 col0 
------
(0 rows)
```

mattnibs:
When a vector.Const had a true value with nulls set, toBool incorrectly created true bits for null positions. Properly flip true/null values to false/null.

Fixes #6482

### DIFF
--- a/runtime/vam/expr/logic.go
+++ b/runtime/vam/expr/logic.go
@@ -167,9 +167,13 @@ func toBool(vec vector.Any) *vector.Bool {
 	case *vector.Const:
 		val := vec.Value()
 		if val.Bool() {
-			out := vector.NewTrue(vec.Len())
-			out.Nulls = vec.Nulls
-			return out
+			var bits bitvec.Bits
+			if vec.Nulls.IsZero() {
+				bits = bitvec.NewTrue(vec.Len())
+			} else {
+				bits = bitvec.Not(vec.Nulls)
+			}
+			return vector.NewBool(bits, vec.Nulls)
 		} else if val.IsNull() {
 			return vector.NewBoolEmpty(vec.Len(), bitvec.NewTrue(vec.Len()))
 		}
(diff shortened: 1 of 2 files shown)


runtime/vam/expr/logic.go
```diff
@@ -167,9 +167,12 @@ func toBool(vec vector.Any) *vector.Bool {
...
 		val := vec.Value()
 		if val.Bool() {
-			out := vector.NewTrue(vec.Len())
-			out.Nulls = vec.Nulls
-			return out
+			bits := bitvec.NewTrue(vec.Len())
+			if !vec.Nulls.IsZero() {
+				// Flip true/null values to false/null.
+				bits = bitvec.And(bits, bitvec.Not(vec.Nulls))
+			}
```
nwt: Nit: We could allocate less here.

```suggestion
			var bits bitvec.Bits
			if vec.Nulls.IsZero() {
				bits = bitvec.NewTrue(vec.Len())
			} else {		
				bits = bitvec.Not(vec.Nulls)
			}
```

nwt (approved):

runtime/vam/expr/logic.go
mattnibs: Nice!

PR #6491, opened 2026-01-06
join(): Fix inconsistency between sam and vam variants

mattnibs:
This commit changes the sam version of join() to return a null string when a null string array is encountered. Also moves the tests to the standard locations and makes the testing more complete.

### DIFF
--- a/runtime/ztests/expr/function/join.yaml
+++ b/runtime/ztests/expr/function/join.yaml
@@ -1,15 +1,45 @@
-spq: join(s, sep)
+spq: |
+  values join(this, "->")
 
 vector: true
 
 input: |
-  {s:["a","b","c"],sep:", "}
-  {s:"join",sep:","}
-  {s:["a"],sep:["b"]}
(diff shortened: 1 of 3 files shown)


runtime/ztests/expr/function/join.yaml
```diff
@@ -1,15 +1,43 @@
...
+output: |
+  "foo->bar->baz"
+  "foo"
+  null::string
+  ""
+  ""
+  error({message:"join: array of string arg required",on:"not an array"})
+
+---
+# Single arg
```
nwt, line 25: Nit:

```suggestion
---

# Single arg
```

runtime/ztests/expr/function/join.yaml
```diff
@@ -1,15 +1,43 @@
...
+vector: true
+
+input: |
+  ["f",".bar.com"]
+
+output: |
+  "f.bar.com"
+
+---
+# Bad separator arg
```
nwt, line 39: Nit:

```suggestion
---

# Bad separator arg
```

nwt (approved):

PR #6492, opened 2026-01-06
vam: Add ksuid() function

mattnibs:
Along with adding the ksuid function to the vector runtime, this commit:
- Uses a generic error message when failing to parse bytes or string input.
- Returns a null value when a null input is encountered.
- Adds a NeedsInput interface to vector functions that require the input value (needed for now() and ksuid()).
- Moves tests to runtime/ztests
- Centralizes testing in to a single test.


runtime/ztests/expr/function/ksuid.yaml
```diff
@@ -0,0 +1,29 @@
...
+  0x0e8afd4a5a0a2372e218818aa996ad96f5739dd2
+  "24eT188EIisOiwVWwmej1h9t4Ms"
+  null::string
+  null::bytes
+  error({message:"ksuid: invalid ksuid value",on:"invalid ksuid"})
+  error({message:"ksuid: invalid ksuid value",on:0x696e76616c6964206b73756964})
+  error({message:"ksuid: argument must a bytes or string type",on:1})
+
+---
+spq: len(ksuid())
```
nwt, line 25: Nit:

```suggestion
---

spq: len(ksuid())
```

PR #6495, opened 2026-01-07
switch to pragma-controlled Postgres style relational scoping

### LINKED ISSUE #5974: SQL: Using an alias identical to a column name
Repro is with super commit e75dfda. This was found via a query from [a sqllogictest](https://github.com/duckdb/duckdb/blob/0cbb56d14cc1464d945a06cb8bac051485fa2ef7/test/sql/aggregate/group/test_group_by.test#L144-L149).

Repro with test data:

```
$ cat integers.sup
{i:3::int32,j:4::int32}
{i:3::int32,j:4::int32}
{i:2::int32,j:4::int32}

$ super -version
Version: e75dfdaf1

$ super -c "SELECT 1 AS i, SUM(i) FROM 'integers.sup' GROUP BY i ORDER BY 2;"
{i:1,"SUM(i)":8}
```

Whereas in Postgres:

```
$ psql postgres
psql (17.5 (Homebrew))
Type "help" for help.

postgres=# SELECT * FROM integers;
 i | j 
---+---
 3 | 4
 3 | 4
 2 | 4
(3 rows)

postgres=# SELECT 1 AS i, SUM(i) FROM integers GROUP BY i ORDER BY 2;
 i | sum 
---+-----
 1 |   2
 1 |   6
(2 rows)
```

### LINKED ISSUE #6410: SQL: Remaining "no corresponding grouping element for non-aggregate" errors
## tl;dr

The fix in #6388 addressed most of the ~176k incidents of the "no corresponding grouping element for non-aggregate" errors from sqllogictests originally tracked in #5973. However, now that a full re-run of all the sqllogictests has been completed with the benefit of that fix there's still 979 queries that produce that error. It's unclear at first glance if these are just corner cases that can be addressed with the same approach used in #6388 or if they present a unique challenge.

## Details

Repro is with super commit a814e1b.

The attached bundle [groupby_slt_good_8_repro.tgz](https://github.com/user-attachments/files/23868556/groupby_slt_good_8_repro.tgz) contains 75 such queries from the 8th "random groupby" sqllogictest set along with test data. As with all the "random" queries, these are "fuzzy" in nature and therefore look unrealistic in some ways. Just run the following and it will show you each query producing the error:

```
$ super -version
Version: a814e1b66

$ sh repro.sh 
no corresponding grouping element for non-aggregate "+tab0.col2" at line 1, column 8:
SELECT + tab0.col2, tab0.col2 FROM tab0 GROUP BY col2;
       ~~~~~~~~~~~
no corresponding grouping element for non-aggregate "col2_1" at line 1, column 12:
SELECT ALL col2, col2 AS col2 FROM tab2 GROUP BY col2;
           ~~~~
...
```

There's plenty of repetition in the style of these queries so I won't enumerate them all. But to my eyes, these first two actually seem representative of what may be two buckets of queries:

## Simple math putting `+` / `-` before column names

That first one shown above is an example of this, and I'm optimistic that it may indeed just be a variation of the kind of queries shown in the original #5973 repro. For example, DuckDB happens to give the `+`-led column a unique name in the output.

```
$ duckdb -c "SELECT + tab0.col2, tab0.col2 FROM tab0.parquet GROUP BY col2;"
┌──────────────┬───────┐
│ +(tab0.col2) │ col2  │
│    int64     │ int64 │
├──────────────┼───────┤
│           24 │    24 │
│           38 │    38 │
│           79 │    79 │
└──────────────┴───────┘
```

## What Would be Duplicate Columns

The second one shown above seems like it falls into this bucket, evidenced by the trailing `_1` in the error message. Indeed, DuckDB uses the same column name twice in the output, and I know this already presents a challenge for us (#5977). I suspect that if we let the `_1` column appear in the output that would be ok (at least for now) but it would be an improvement if we didn't reject the query.

```
$ duckdb -c "SELECT ALL col2, col2 AS col2 FROM tab2.parquet GROUP BY col2;"
┌───────┬───────┐
│ col2  │ col2  │
│ int64 │ int64 │
├───────┼───────┤
│    58 │    58 │
│    79 │    79 │
│    87 │    87 │
└───────┴───────┘
```

mccanne:
This commit changes the relational scoping rules to follow Postgres except for GROUP BY expressions which check column aliases before the input table (like Google SQL).  This is favored as the default for SuperSQL because it doesn't change binding when a dynamic input transitions to a typed input (and thus a field that was present but unknown in the input changes the GROUP BY binding of a column alias).

This behavior can be overridden now with "pragma pg=true" which says to follow Postgres semantics and favor the input table over any column alias in GROUP BY expressions.

As a result, column aliases are no longer visible in WHERE clauses or SELECT expressions in line with Postgres semantics (and ANSI and Google SQL).

We added some light documentation to the pragma section of the book in this PR, but more details and an explanation of the rationale for this design will be forthcoming in a future PR from the book-sql branch.

Fixes #5974
Fixes #6410

### DIFF
--- a/book/src/super-sql/declarations/pragmas.md
+++ b/book/src/super-sql/declarations/pragmas.md
@@ -22,7 +22,7 @@ Currently, there are two supported pragmas.
     * `1` for one-based indexing
 * `pg` - controls the precedence of scoping for GROUP BY clauses in SQL operators
     * `false` to follow Google SQL semantics of resolving identifiers first from column aliases then from the input table
-    * `true` to follow Postgres semantics of resolving identifiers first from the input table then from the column aliases
+    * `true` to follow PostgreSQL semantics of resolving identifiers first from the input table then from the column aliases
 
 ### Example
 
--- a/compiler/semantic/op.go
+++ b/compiler/semantic/op.go
@@ -1552,7 +1552,11 @@ func (t *translator) mustEvalString(e sem.Expr) (field string, ok bool) {
 
 func (t *translator) mustEvalBool(in ast.Expr) (bool, bool) {
 	val, ok := t.mustEval(t.expr(in))
-	if ok && !val.IsError() && super.TypeUnder(val.Type()) == super.TypeBool {
+	if ok {
+		if super.TypeUnder(val.Type()) != super.TypeBool {
+			t.error(in, errors.New("expected type bool"))
+			return false, false
+		}
 		return val.AsBool(), true
 	}
 	return false, false
@@ -1560,7 +1564,7 @@ func (t *translator) mustEvalBool(in ast.Expr) (bool, bool) {
 
 func (t *translator) maybeEvalString(e sem.Expr) (field string, ok bool) {
 	val, ok := t.maybeEval(e)
-	if ok && !val.IsError() && super.TypeUnder(val.Type()) == super.TypeString {
+	if ok && super.TypeUnder(val.Type()) == super.TypeString {
 		return string(val.Bytes()), true
 	}
 	return "", false
@@ -1573,12 +1577,12 @@ func (t *translator) mustEvalPositiveInteger(ae ast.Expr) int {
 		return 0
 	}
 	if !super.IsInteger(val.Type().ID()) || val.IsNull() {
-		t.error(ae, fmt.Errorf("expression must be an integer: %s", sup.FormatValue(val)))
+		t.error(ae, errors.New("expected integer"))
 		return 0
 	}
 	v := int(val.AsInt())
 	if v < 0 {
-		t.error(ae, errors.New("expression must be a positive integer"))
+		t.error(ae, errors.New("expected positive integer"))
 		return 0
 	}
 	return v
--- a/compiler/semantic/schema.go
+++ b/compiler/semantic/schema.go
@@ -76,9 +76,7 @@ type (
 
 		// Columns holds the projection.  We bookkeep both the ast expression
 		// and sem tree expr so we can do lateral column alias macro substition.
-		// Also, a SELECT column can refer to only the lateral columns to its left
-		// so latStop limits the extent of the column lookup to avoid
-		// an infinite expansion cycle.
+		// The lateral field is true when lateral column resolution is allowed.
 		columns []column
 		lateral bool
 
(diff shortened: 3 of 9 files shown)


book/src/super-sql/declarations/pragmas.md
```diff
@@ -14,12 +14,15 @@ Pragmas must appear in the declaration section of a [scope](../queries.md#scope)
...
-Currently, there is just one supported pragma.
+Currently, there are two supported pragmas.
 
 * `index_base` - controls whether [index expressions](../expressions/index.md) and
     [slice expressions](../expressions/slices.md) are 0-based or 1-based.
     * `0` for zero-based indexing
     * `1` for one-based indexing
+* `pg` - controls the precedence of scoping for GROUP BY clauses in SQL operators
+    * `false` to follow Google SQL semantics of resolving identifiers first from column aliases then from the input table
+    * `true` to follow Postgres semantics of resolving identifiers first from the input table then from the column aliases
```
nwt:
```suggestion
    * `true` to follow PostgreSQL semantics of resolving identifiers first from the input table then from the column aliases
```

compiler/semantic/schema.go
```diff
@@ -80,7 +80,7 @@ type (
 		// so latStop limits the extent of the column lookup to avoid
```
```suggestion
		// so lateral limits the extent of the column lookup to avoid
```

compiler/semantic/op.go
```diff
@@ -1555,6 +1550,14 @@ func (t *translator) mustEvalString(e sem.Expr) (field string, ok bool) {
 	return "", false
 }
 
+func (t *translator) mustEvalBool(in ast.Expr) (bool, bool) {
+	val, ok := t.mustEval(t.expr(in))
+	if ok && !val.IsError() && super.TypeUnder(val.Type()) == super.TypeBool {
```
mattnibs: Don't need the error check
```suggestion
	if ok && super.TypeUnder(val.Type()) == super.TypeBool {
```

compiler/semantic/schema.go
mccanne: I update this comment as it was stale.

compiler/semantic/op.go
mccanne: Good point.  I blindly copied that pattern and it should really report a type error.  Also for some of the other cases.  I will address.

PR #6496, opened 2026-01-07
allow short pragma to indicate bool true

mccanne:


mccanne:
Rebased on bug fix so tests should pass now

PR #6498, opened 2026-01-07
fix relational scoping bug introduced in #6495

mccanne:
This commit fixes a bug where ORDER BY scope was improperly handled because the logic favored the input table over the column aliases but the materialized output table has priority over the input table once the SELECT body scope is left. This would typically be handled by first looking into the output table but we can't do this because we do expression matching on the sem-tree expressions and we can't mix and match input expressions with the output expressions.  The fix is to look in the columns first (producing input table expressions) and matching those expressions to the grouped output table columns with the expression matcher.

### DIFF
--- a/compiler/semantic/scope.go
+++ b/compiler/semantic/scope.go
@@ -151,7 +151,7 @@ func (s *Scope) resolve(t *translator, n ast.Node, path field.Path) (sem.Expr, e
 	// SELECT body (e.g., in ORDER BY), so we need to first check
 	// the select column names before the inputs.  This is handled by
 	// other SQLs by looking in the output table, but we can't do that
-	// since we must can't mix output columns and input columns
+	// since we can't mix output columns and input columns
 	// in the grouping expression matcher.  So we rely upon this column
 	// lookup to give us the input bindings and then match everything to its
 	// grouped output.
@@ -163,7 +163,7 @@ func (s *Scope) resolve(t *translator, n ast.Node, path field.Path) (sem.Expr, e
 		}
 	}
 	// We give priority to lateral column aliases which diverges from
-	// Postgres semantics (which checks the input table first) but follows
+	// PostgreSQL semantics (which checks the input table first) but follows
 	// Google SQL semantics.  We favor this approach so we can avoid
 	// precedence problems for dynamic inputs or super-structured inputs
 	// that evolve, e.g., with a super-structured input, whether a column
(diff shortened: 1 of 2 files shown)


mccanne:
This should fix the intermittent test failure in #6496

compiler/semantic/scope.go
```diff
@@ -147,6 +147,21 @@ func (s *Scope) resolve(t *translator, n ast.Node, path field.Path) (sem.Expr, e
 			inputFirst = b
 		}
 	}
+	// If the output projection is in place, then we are outside the
+	// SELECT body (e.g., in ORDER BY), so we need to first check
+	// the select column names before the inputs.  This is handled by
+	// other SQLs by looking in the output table, but we can't do that
+	// since we must can't mix output columns and input columns
```
nwt:
```suggestion
	// since we can't mix output columns and input columns
```

compiler/ztests/sql/order-by-pg-scope.yaml
```diff
@@ -0,0 +1,16 @@
+script: |
+  super compile -C -O -I query.spq | grep sort
+
+inputs:
+  - name: query.spq
+    data: |
+      pragma pg=false
```
nwt: Is this what you want? I was expecting pg=true based on the file name.

compiler/semantic/scope.go
```diff
@@ -155,7 +170,8 @@ func (s *Scope) resolve(t *translator, n ast.Node, path field.Path) (sem.Expr, e
 	// is present in the input shouldn't control the scoping decision compared
 	// to a column aliases otherwise as data evolves (and such a column shows up)
 	// the binding of identifiers changes and query results can change.  By
-	// resolving lateral aliases first, we avoid this problem.
+	// resolving lateral aliases first, we avoid this problem.  This can
+	// be overridden with "pragma pg" to favor Postgres precedence.
```

```suggestion
	// be overridden with "pragma pg" to favor PostgreSQL precedence.
```

nwt (approved):

compiler/ztests/sql/order-by-pg-scope.yaml
mccanne: Oops, you're right.  The bug was when PG scope was off and it was the output table causing the problem.  I'll rename the file.

mccanne: And remove pg=false

PR #6500, opened 2026-01-07
vam: Add support for fuse() aggregation

mattnibs:


runtime/ztests/op/aggregate/fuse.yaml
```diff
@@ -1,4 +1,19 @@
...
+
+input: |
+  {a:"hello",b:"world"}
+  {a:"goodnight",b:123::int32}
+
+output: |
+  <{a:string,b:int32|string}>
+
+---
+# Test partials.
```
nwt, line 14: Nit:

```suggestion
---

# Test partials.
```

PR #6511, opened 2026-01-08
vam: Add support for map literals

mattnibs:


runtime/vam/expr/mapexpr.go
```diff
@@ -0,0 +1,78 @@
...
+		entries: entries,
+	}
+}
+
+func (m *MapExpr) Eval(this vector.Any) vector.Any {
+	if len(m.entries) == 0 {
+		mtyp := m.sctx.LookupTypeMap(super.TypeNull, super.TypeNull)
+		offsets := make([]uint32, this.Len()+1)
+		key := vector.NewConst(super.Null, 0, bitvec.Zero)
+		val := vector.NewConst(super.Null, 0, bitvec.Zero)
```
nwt: Nit: Maybe use the same vector for the keys and values? That would be safe to do.

PR #6522, opened 2026-01-13
represent complex literals as expressions

mccanne:
This commit changes complex literals in the sem tree and DAG to always be represented by expressions.  To enforce this, we changed sem.LiteralExpr and dag.LiteralExpr to PrimitiveExpr and created sem.NewLiteral() to handle all creations of complex literal values.  NewLiteral() can also be used on primitive types as it creates a PrimitiveExpr for primitive types and an expression for complex literals.

This refactoring exposed the fact that complex literals could be handled by the bsup bufferfilter and the pruning logic in the optimizer.  With these changes, bufferfilter and pruning is now limited to primitive types.

We also fixed super.NewTypeValue which was using the wrong type ID.

### DIFF
--- a/compiler/semantic/checker.go
+++ b/compiler/semantic/checker.go
@@ -370,13 +370,14 @@ func (c *checker) expr(typ super.Type, e sem.Expr) super.Type {
 		}
 		return c.t.sctx.LookupTypeMap(c.fuse(keyTypes), c.fuse(valTypes))
 	case *sem.PrimitiveExpr:
-		if val, err := sup.ParseValue(c.t.sctx, e.Value); err == nil {
-			if !super.IsPrimitiveType(val.Type()) {
-				panic(e)
-			}
-			return val.Type()
+		val, err := sup.ParseValue(c.t.sctx, e.Value)
+		if err != nil {
+			return c.unknown
 		}
-		return c.unknown
+		if !super.IsPrimitiveType(val.Type()) {
+			panic(e)
+		}
+		return val.Type()
 	case *sem.RecordExpr:
 		return c.recordElems(typ, e.Elems)
 	case *sem.RegexpMatchExpr:
--- a/compiler/semantic/sem/expr.go
+++ b/compiler/semantic/sem/expr.go
@@ -489,11 +489,11 @@ func recordToExpr(loc ast.Node, typ *super.TypeRecord, bytes scode.Bytes) Expr {
 	}
 	var elems []RecordElem
 	it := bytes.Iter()
-	for k, f := range typ.Fields {
+	for _, f := range typ.Fields {
 		if it.Done() {
 			panic(it)
 		}
-		elem := &FieldElem{Node: loc, Name: typ.Fields[k].Name, Value: valueToExpr(loc, f.Type, it.Next())}
+		elem := &FieldElem{Node: loc, Name: f.Name, Value: valueToExpr(loc, f.Type, it.Next())}
 		elems = append(elems, elem)
 	}
 	return &RecordExpr{
@@ -508,8 +508,7 @@ func arrayToExpr(loc ast.Node, typ *super.TypeArray, bytes scode.Bytes) Expr {
 	}
 	var elems []ArrayElem
 	inner := super.InnerType(typ)
-	it := bytes.Iter()
-	for !it.Done() {
+	for it := bytes.Iter(); !it.Done(); {
 		elems = append(elems, &ExprElem{Node: loc, Expr: valueToExpr(loc, inner, it.Next())})
 	}
 	return &ArrayExpr{
@@ -524,8 +523,8 @@ func setToExpr(loc ast.Node, typ *super.TypeSet, bytes scode.Bytes) Expr {
 	}
 	var elems []ArrayElem
 	inner := super.InnerType(typ)
-	it := bytes.Iter()
-	for !it.Done() {
+
+	for it := bytes.Iter(); !it.Done(); {
 		elems = append(elems, &ExprElem{Node: loc, Expr: valueToExpr(loc, inner, it.Next())})
 	}
 	return &SetExpr{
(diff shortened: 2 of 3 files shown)


compiler/rungen/bufferfilter.go
```diff
@@ -81,16 +81,18 @@ func CompileBufferFilter(sctx *super.Context, e dag.Expr) (*expr.BufferFilter, e
 }
 
 func isFieldEqualOrIn(sctx *super.Context, e *dag.BinaryExpr) (*super.Value, error) {
+	// XXX This could work for complex literal but we need to walk the
+	// expression and compute a super.Value or call eval.
```
mattnibs, line 85: I don't think this comment applies at this point.

mccanne: It does apply.  We're giving up bufferfiltering on complex values with these changes but we could get them back here by converting a dag.Expr that is a constant-valued complex expression into a super.Value that could be used in bufferfilter.

compiler/semantic/checker.go
```diff
@@ -374,6 +369,14 @@ func (c *checker) expr(typ super.Type, e sem.Expr) super.Type {
...
 		}
 		return c.t.sctx.LookupTypeMap(c.fuse(keyTypes), c.fuse(valTypes))
+	case *sem.PrimitiveExpr:
+		if val, err := sup.ParseValue(c.t.sctx, e.Value); err == nil {
+			if !super.IsPrimitiveType(val.Type()) {
+				panic(e)
+			}
+			return val.Type()
+		}
+		return c.unknown
```
nwt: Nit: Indent error paths instead of the common path to improve readability.

```suggestion
		val, err := sup.ParseValue(c.t.sctx, e.Value)
		if err != nil {
			return c.unknown
		}
		if !super.IsPrimitiveType(val.Type()) {
			panic(e)
		}
		return val.Type()
```

value.go
```diff
@@ -89,7 +89,7 @@ func NewBytes(b []byte) Value           { return NewValue(TypeBytes, b) }
 func NewString(s string) Value          { return Value{TypeString, nonNilUnsafeStringData(s), uint64(len(s))} }
 func NewIP(a netip.Addr) Value          { return NewValue(TypeIP, EncodeIP(a)) }
 func NewNet(p netip.Prefix) Value       { return NewValue(TypeNet, EncodeNet(p)) }
-func NewTypeValue(t Type) Value         { return NewValue(TypeNet, EncodeTypeValue(t)) }
+func NewTypeValue(t Type) Value         { return NewValue(TypeType, EncodeTypeValue(t)) }
```
nwt, line 92: Yikes. This bug's been here since I added NewTypeValue in January 2023.

compiler/sfmt/dag.go
```diff
@@ -111,7 +111,7 @@ func (c *canonDAG) expr(e dag.Expr, parent string) {
 	case *dag.IsNullExpr:
 		c.expr(e.Expr, "")
 		c.write(" IS NULL")
-	case *dag.LiteralExpr:
+	case *dag.PrimitiveExpr:
 		c.write("%s", e.Value)
```
nwt: Nit: Move this after dag.MapExpr.

compiler/semantic/sem/expr.go
```diff
@@ -452,3 +454,120 @@ func copyArrayElems(elems []ArrayElem) []ArrayElem {
...
+	if bytes == nil {
+		return NewCast(loc, valueToExpr(loc, super.TypeNull, nil), typ)
+	}
+	var elems []RecordElem
+	it := bytes.Iter()
+	for k, f := range typ.Fields {
+		if it.Done() {
+			panic(it)
+		}
+		elem := &FieldElem{Node: loc, Name: typ.Fields[k].Name, Value: valueToExpr(loc, f.Type, it.Next())}
```
nwt: Nit:

```suggestion
	for _, f := range typ.Fields {
		if it.Done() {
			panic(it)
		}
		elem := &FieldElem{Node: loc, Name: f.Name, Value: valueToExpr(loc, f.Type, it.Next())}
```

compiler/semantic/sem/expr.go
```diff
@@ -452,3 +454,120 @@ func copyArrayElems(elems []ArrayElem) []ArrayElem {
...
+}
+
+func arrayToExpr(loc ast.Node, typ *super.TypeArray, bytes scode.Bytes) Expr {
+	if bytes == nil {
+		return NewCast(loc, valueToExpr(loc, super.TypeNull, nil), typ)
+	}
+	var elems []ArrayElem
+	inner := super.InnerType(typ)
+	it := bytes.Iter()
+	for !it.Done() {
```
nwt: Nit:

```suggestion
	for it := bytes.Iter(); !it.Done(); {
```

compiler/semantic/sem/expr.go
```diff
@@ -452,3 +454,120 @@ func copyArrayElems(elems []ArrayElem) []ArrayElem {
...
+}
+
+func setToExpr(loc ast.Node, typ *super.TypeSet, bytes scode.Bytes) Expr {
+	if bytes == nil {
+		return NewCast(loc, valueToExpr(loc, super.TypeNull, nil), typ)
+	}
+	var elems []ArrayElem
+	inner := super.InnerType(typ)
+	it := bytes.Iter()
+	for !it.Done() {
```
nwt: Nit:

```suggestion
	for it := bytes.Iter(); !it.Done(); {
```

compiler/semantic/sem/expr.go
```diff
@@ -452,3 +454,120 @@ func copyArrayElems(elems []ArrayElem) []ArrayElem {
...
+
+func mapToExpr(loc ast.Node, typ *super.TypeMap, bytes scode.Bytes) Expr {
+	if bytes == nil {
+		return NewCast(loc, valueToExpr(loc, super.TypeNull, nil), typ)
+	}
+	keyType := super.TypeUnder(typ.KeyType)
+	valType := super.TypeUnder(typ.ValType)
+	it := bytes.Iter()
+	var entries []Entry
+	for !it.Done() {
```
nwt: Nit:

```suggestion
	var entries []Entry
	for it := bytes.Iter(); !it.Done(); {
```

value.go
mccanne: LOL

PR #6523, opened 2026-01-14
Add map() function support to vector runtime

mattnibs:


runtime/sam/expr/map.go
```diff
@@ -28,7 +28,7 @@ func (a *mapCall) Eval(in super.Value) super.Value {
 	}
 	elems, err := val.Elements()
 	if err != nil {
-		return a.sctx.WrapError(err.Error(), in)
+		return a.sctx.WrapError("expected array or set value", in)
```
nwt: Include the function name?

```suggestion
		return a.sctx.WrapError("map: expected array or set value", in)
```

runtime/ztests/expr/map-func.yaml
```diff
@@ -0,0 +1,32 @@
...
+  |["foo","bar",null::string]|
+  null::|[string]|
+  "invalid"
+
+output: |
+  ["FOO","BAR",null::string]
+  null::[string]
+  |[null::string,"BAR","FOO"]|
+  null::|[string]|
+  error({message:"expected array or set value",on:"invalid"})
```
nwt: Nit: This would be better with the function name.

```suggestion
  error({message:"map: expected array or set value",on:"invalid"})
```

nwt (approved):

PR #6526, opened 2026-01-14
has_error: null error values return false

mattnibs:
This commit fixes has_error so that null error values return false for having an error.

### DIFF
--- a/runtime/sam/expr/function/types.go
+++ b/runtime/sam/expr/function/types.go
@@ -100,10 +100,6 @@ func (h *HasError) Call(args []super.Value) super.Value {
 }
 
 func (h *HasError) hasError(t super.Type, b scode.Bytes) (bool, bool) {
-	typ := super.TypeUnder(t)
-	if _, ok := typ.(*super.TypeError); ok {
-		return true, false
-	}
 	// If a value is null we can skip since an null error is not an error.
 	if b == nil {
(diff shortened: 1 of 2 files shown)


runtime/sam/expr/function/types.go
```diff
@@ -148,6 +144,8 @@ func (h *HasError) hasError(t super.Type, b scode.Bytes) (bool, bool) {
 			hasErr, cc = h.hasError(typ, b)
 			canCache = !canCache || cc
 		}
+	case *super.TypeError:
+		hasErr = true
```
nwt, line 148: hasErr isn't cacheable here so the second output value in this example is incorrect. (It should be false since field a is null.)

```console
$ go run ./cmd/super -c 'values {a:error(1)}, {a:null::error(int64)} | values [this,has_error(this)]'
true
true
```

The fix:

```suggestion
		hasErr, canCache = true, false
```

Add a test too.

mattnibs: Yeah I got rid of the cacheable stuff in this pr: https://github.com/brimdata/super/pull/6530

But we can do this change here if you want.

nwt: Then this is fine.

For next time, a note on the PR or comment in the code pointing out known a bug like this will save us both some time.

PR #6528, opened 2026-01-14
is_error: return false for null error

mattnibs:


runtime/sam/expr/function/types.go
```diff
@@ -64,7 +64,7 @@ func (e *Error) Call(args []super.Value) super.Value {
 type IsErr struct{}
 
 func (*IsErr) Call(args []super.Value) super.Value {
-	return super.NewBool(args[0].IsError())
+	return super.NewBool(!args[0].IsNull() && args[0].IsError())
```
nwt: Nit: I think this reads better with the IsError first.

```suggestion
	return super.NewBool(args[0].IsError() && !args[0].IsNull())
```

PR #6529, opened 2026-01-14
vam: Add support for is_error() function

mattnibs:


runtime/vam/expr/function/types.go
```diff
@@ -36,6 +36,19 @@ func (i *Is) Call(args ...vector.Any) vector.Any {
...
+
+func (*IsErr) Call(args ...vector.Any) vector.Any {
+	errvec, ok := vector.Under(args[0]).(*vector.Error)
+	if !ok {
+		return vector.NewConst(super.False, args[0].Len(), bitvec.Zero)
+	}
+	if errvec.Nulls.IsZero() {
+		return vector.NewConst(super.True, errvec.Len(), bitvec.Zero)
+	}
+	return vector.NewBool(bitvec.Not(errvec.Nulls), bitvec.Zero)
```
nwt: I don't think this will work for vector.Dict or vector.View.

```suggestion
	vec := args[0]
	if vector.KindOf(vec) == vector.KindError {
		return vector.NewConst(super.False, vec.Len(), bitvec.Zero)
	}
	nulls := vector.NullsOf(vec)
	if nulls.IsZero() {
		return vector.NewConst(super.True, vec.Len(), bitvec.Zero)
	}
	return vector.NewBool(bitvec.Not(nulls), bitvec.Zero)
```

mattnibs: Views are pushed down when applied to errors. From vector/view.go.Pick:
```
	case *Error:
		return NewError(val.Typ, Pick(val.Vals, index), val.Nulls.Pick(index))
```
In CSUP we only do dicts on primitive types. Is there anywhere else a dict can wrap an error?

nwt: It's true at present that we only create a vector.Dict for primitive types and that vector.Pick on a vector.Error always returns a vector.Error, but there's no guarantee those facts won't change (and there are reasons why they might). Baking them into this code feels like a misstep.

mattnibs: I'll do it your way here, but we should discuss this. I just did a quick check and there are a lot of places in the vector runtime that have the assumption errors aren't wrapped in views or dicts. I could see an argument for wapping errors in views (does make the code more complex) but the upside to wrapping them in dicts I'm having a hard time seeing.

nwt: Happy to discuss it.

PR #6530, opened 2026-01-15
Remove cache for HasError function

### LINKED ISSUE #6199: Array element order foils has_error
This returns no output.

```
$ echo '[1,error("hi")]' | super -c "has_error(this)" -
[no output]
```

## Details

Repro is with super commit 54aa3ff.

Working my way up to the repro, these variations all work as expected.

```
$ super -version
Version: 54aa3ff47

$ echo 'error("hi")' | super -c "has_error(this)" -
error("hi")

$ echo '[error("hi")]' | super -c "has_error(this)" -
[error("hi")]

$ echo '[error("hi"),1]' | super -c "has_error(this)" -
[error("hi"),1]
```

It's only when I put the non-error value first in the array that I was able to break it.

mattnibs:
This commit removes the caching logic from the sequential runtime implementation of the HasError function. There was a bug in the current implementation where unions values that had an error type but were not errors themselves would report the entire union type as a cacheable non-error then subsequent error values of that union would get false reported as not having an error. This caching behavior could be fixed with additional logic but since we are not so interested in performance in the sequential runtime remove the caching and simplify the logic.

Fixes #6199

### DIFF
--- a/runtime/sam/expr/function/types.go
+++ b/runtime/sam/expr/function/types.go
@@ -1,6 +1,8 @@
 package function
 
 import (
+	"slices"
+
 	"github.com/brimdata/super"
 	"github.com/brimdata/super/scode"
 )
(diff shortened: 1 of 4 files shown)


runtime/sam/expr/function/types.go
```diff
@@ -83,76 +85,46 @@ func (i *Is) Call(args []super.Value) super.Value {
...
-			canCache = !canCache || isErr
-		}
-		if typ, b := typ.Untag(b); b != nil {
-			// Check mb is not nil to avoid infinite recursion.
-			var cc bool
-			hasErr, cc = h.hasError(typ, b)
-			canCache = !canCache || cc
-		}
+		utyp, b := typ.Untag(b)
+		return b != nil && h.hasError(utyp, b)
```
nwt: Nit: This check for nil duplicates the one done at the top of the method.

```suggestion
		return h.hasError(typ.Untag(b))
```

runtime/sam/expr/ztests/has_error.yaml
```diff
@@ -5,10 +5,13 @@ input: |
   error(0)
   {f1:127.0.0.1,f2:error("error")}
   [1,2]::[int64|error(string)]
-  |{"key":error(0)::(int64|error(int64))}|
```
nwt, line 8: Don't remove this. It's the only test case with a non-null union containing an error.

mattnibs: I removed this because this test also contains a non-null union containing an error `|{"k2":1,"k1":error(0)}|` and is a more comprehensive.

nwt: Forgive me if I'm missing something but `|{"k2":1,"k1":error(0)}|` doesn't contain a union, does it?

nwt: Maybe change

```
{"k2":1,"k1":error(0)}|
```

to

```
{"k2":1,"k1":error(0)::(int64|error(int64))}|
```

?

mattnibs: I was coming from the view point of the vector runtime where the vector that represents the values in the map will get represented as a union of `int64|error(int64)`. Not as familiar with what is going on in the sequential runtime perhaps I am missing something?

```
$ super -c 'values typeof(|{"k2":1,"k1":error(0)}|)'
<|{string:int64|error(int64)}|>
```

nwt: Ah, I was slow on the uptake here. I didn't realize that the map value type has to be a union, but of course it does since the map has two values with different types.

PR #6531, opened 2026-01-15
integrate relational scoping and pipe scoping

mccanne:
This commit integrates relational and pipe scoping by moving the type analaysis from the semantic checker into the translator. With this change, the fused type that emerges from a pipe query can be fed into SQL and converted into a relational table and SQL output relations can be converted back into super types for type analysis in a pipe query segment.

### DIFF
--- a/compiler/semantic/op.go
+++ b/compiler/semantic/op.go
@@ -653,7 +653,7 @@ func (t *translator) semOp(o ast.Op, seq sem.Seq, inType super.Type) (sem.Seq, s
 					&sem.FieldElem{Name: "that", Value: sem.NewThis(nil, nil)},
 				},
 			}
-			fields = []super.Field{{Name: "that", Type: inType}}
+			fields = []super.Field{super.NewField("that", inType)}
 		} else {
 			n := len(o.Expr.Elems)
 			if n == 0 {
@@ -697,7 +697,7 @@ func (t *translator) semOp(o ast.Op, seq sem.Seq, inType super.Type) (sem.Seq, s
 				fields = slices.Clone(recType.Fields)
 			}
 		}
-		fields = append(fields, super.Field{Name: alias, Type: super.TypeInt64})
+		fields = append(fields, super.NewField(alias, super.TypeInt64))
 		return append(seq, &sem.CountOp{
 			Node:  o,
 			Alias: alias,
@@ -875,8 +875,7 @@ func (t *translator) semOp(o ast.Op, seq sem.Seq, inType super.Type) (sem.Seq, s
 		var paths []pathType
 		for _, fa := range o.Args {
 			assign, path := t.assignment(&fa, inType)
-			_, ok := isLval(assign.RHS)
-			if !ok {
+			if _, ok := isLval(assign.RHS); !ok {
 				t.error(fa.RHS, fmt.Errorf("illegal right-hand side of assignment"))
 			}
 			// If both paths are static validate them. Otherwise this will be
@@ -1142,8 +1141,6 @@ func (t *translator) switchOp(op *ast.SwitchOp, seq sem.Seq, inType super.Type) 
 		if c.Expr != nil {
 			e, _ = t.expr(c.Expr, inType)
 		} else if op.Expr == nil {
-			// c.Expr == nil indicates the default case,
-			// whose handling depends on p.Expr.
 			e = sem.NewLiteral(op, super.True)
 		}
 		path, typ := t.seq(c.Path, inType)
--- a/compiler/semantic/resolver.go
+++ b/compiler/semantic/resolver.go
@@ -159,7 +159,7 @@ func (r *resolver) lookupFixed(id string, inType super.Type) (string, super.Type
 		if f, ok := r.funcs[tag]; ok {
 			return tag, r.t.checker.expr(inType, f.body)
 		}
-		// This function is recursive so we can't know it's type here.
+		// This function is recursive so we can't know its type here.
 		// At some point, we will add function prototypes so type checking
 		// can work more completely on recursive functions.
 		return tag, r.t.checker.unknown
--- a/compiler/semantic/schema.go
+++ b/compiler/semantic/schema.go
@@ -313,7 +313,7 @@ func (s *selectScope) resolveUnqualified(col string) (field.Path, bool, error) {
 		// This isn't a problem when the table has aggregate output (!isGrouped())
 		// (because there is no way to reference the input table values outside
 		// of agg func arguments), and it can't work because expression matching
-		// would be foiled by paths with out.col..
+		// would be foiled by paths with out.col.
 		path, dyn, err := s.out.resolveUnqualified(col)
 		if err != nil {
 			return nil, false, err
--- a/compiler/semantic/scope.go
+++ b/compiler/semantic/scope.go
@@ -266,8 +266,7 @@ func resolveLateralColumn(t *translator, scope *selectScope, col string, inType 
 		// lateral columns available only inside select bodies
 		return nil, t.checker.unknown
 	}
-	for k := range len(scope.columns) {
-		c := scope.columns[k]
+	for _, c := range scope.columns {
 		if c.lateral && c.name == col {
 			defer func() {
 				scope.lateral = true
--- a/compiler/semantic/sql.go
+++ b/compiler/semantic/sql.go
@@ -60,7 +60,7 @@ func (t *translator) sqlSelect(sel *ast.SQLSelect, demand []ast.Expr, seq sem.Se
 	scope.lateral = true
 	scope.groupings = t.groupBy(scope, sel.GroupBy, inType)
 	scope.aggOk = true
-	// Make sure any aggs needed by ORDER BY and HAVING computed and
+	// Make sure any aggs needed by ORDER BY and HAVING are computed and
 	// placed into the scope.aggs table.
 	// This is a little tricky because we need to find the agg functions
 	// and type check their args against the input table, but we can't
@@ -289,12 +289,12 @@ func groupTmp(k int) string {
 }
 
 // selectFrom analyzes the FROM clause which is a cross-joined set of table expressions.
-// inType is immediate pipe parent in a top-level select, nil otherwise
+// inType is from the immediate pipe parent in a top-level select, nil otherwise
 func (t *translator) selectFrom(loc ast.Loc, exprs []ast.SQLTableExpr, seq sem.Seq, inType super.Type) (sem.Seq, relScope) {
 	if len(exprs) == 0 {
 		// No FROM clause is modeled by a single null value, which we represent
-		//  with dynamic, e.g., so "select this" results in {that:null},
-		// or its the parent pipe input.
+		// with dynamic, e.g., so "select this" results in {that:null},
+		// or the parent pipe input.
 		return seq, newTableFromType(inType, t.checker.unknown)
 	}
 	off := len(seq)
@@ -321,7 +321,7 @@ func (t *translator) selectFrom(loc ast.Loc, exprs []ast.SQLTableExpr, seq sem.S
 }
 
 // sqlValues analyzes the values expression which currently cannot depend
-// on the input data (until we add support for lateral joins then a values
+// on the input data (until we add support for lateral joins; then a values
 // clause in the FROM can refer to the left relation).
 func (t *translator) sqlValues(values *ast.SQLValues, seq sem.Seq) (sem.Seq, *staticTable) {
 	exprs := make([]sem.Expr, 0, len(values.Exprs))
@@ -337,7 +337,7 @@ func (t *translator) sqlValues(values *ast.SQLValues, seq sem.Seq) (sem.Seq, *st
 
 // With type checking integrated into the SQL analysis, the logic here
 // can always infer a staticTable.  Eventually, this will need to support
-// lateral join subqueries which have runtime values (but staticTable columns still known).
+// lateral join subqueries which have runtime values (but staticTable columns are still known).
 func (t *translator) inferSchema(loc ast.Node, types []super.Type) *staticTable {
 	recType, ok := super.TypeUnder(t.checker.fuse(types)).(*super.TypeRecord)
 	if !ok {


compiler/semantic/checker.go
```diff
@@ -429,6 +419,35 @@ func (c *checker) expr(typ super.Type, e sem.Expr) super.Type {
...
+		return super.TypeBool
+	case "==", "!=", "=", "<>":
+		return c.equality(lhs, rhs)
+	case "<", "<=", ">", ">=":
+		return c.comparison(lhs, rhs)
+	case "+", "-", "*", "/", "%":
+		if op == "+" {
+			return c.plus(loc, lhs, rhs)
+		}
+		return c.arithmetic(lloc, rloc, lhs, rhs)
```

```suggestion
	case "+":
		return c.plus(loc, lhs, rhs)
	case "-", "*", "/", "%":
		return c.arithmetic(lloc, rloc, lhs, rhs)
```

compiler/semantic/checker.go
```diff
@@ -905,6 +948,24 @@ func hasUnknown(typ super.Type) bool {
...
+	if u, ok := super.TypeUnder(typ).(*super.TypeUnion); ok {
+		for _, t := range u.Types {
+			if elemType, ok := c.hasArray(t); ok {
+				return elemType, true
+			}
+		}
+	}
+	if a, ok := super.TypeUnder(typ).(*super.TypeArray); ok {
+		return a.Type, true
+	}
```
nwt: Nit: I think a type switch is a little clearer.

```suggestion
	switch typ := super.TypeUnder(typ).(type) {
	case *super.TypeUnion:
		for _, t := range typ.Types {
			if elemType, ok := c.hasArray(t); ok {
				return elemType, true
			}
		}
	case *super.TypeArray:
		return typ.Type, true
	}
```

compiler/ztests/join.yaml
```diff
@@ -24,17 +24,23 @@ outputs:
...
       // ===
-      ambiguous field reference "foo" at line 1, column 9:
-      join on foo.key=left.key
-              ~~~~~~~~~~~~~~~~
+      ambiguous field reference "foo" at line 1, column 30:
+      fork (pass) (pass) | join on foo.key=left.key
+                                   ~~~~~~~~~~~~~~~~
+      no such field "foo" at line 1, column 30:
+      fork (pass) (pass) | join on foo.key=left.key
+                                   ~~~
```
nwt: Nit: To me the "ambiguous field reference" error implies that the field exist in both join inputs, so the "no such field" error is surprising. Maybe just show one error?

compiler/semantic/ztests/checker-drop.yaml
```diff
@@ -1,6 +1,9 @@
 spq: values {x:{y:1,w:1},z:1} | drop x.y | drop x.y
 
 error: |
-  no such field to drop: "y" at line 1, column 44:
+  no such field "y" at line 1, column 46:
+  values {x:{y:1,w:1},z:1} | drop x.y | drop x.y
+                                               ~
+  expression not present in drop input at line 1, column 44:
   values {x:{y:1,w:1},z:1} | drop x.y | drop x.y
```
nwt: Nit: The second error feels superfluous. If it always follows a "no such field" error, which I figure it does, maybe just remove the check that produces it?

compiler/semantic/ztests/const-err-quoted-path.yaml
```diff
@@ -3,6 +3,6 @@ spq: |
   values x
 
 error: |
-  cannot reference '"a b"."c d"' in constant expression at line 1, column 11:
+  "a b" no such field at line 1, column 11:
```
nwt: Nit: Other tests contain errors in which the field name follows "no such field". Make this consistent with that.

compiler/semantic/expr.go
```diff
@@ -19,31 +19,31 @@ import (
 	"github.com/shellyln/go-sql-like-expr/likeexpr"
 )
 
-func (t *translator) expr(e ast.Expr) sem.Expr {
+func (t *translator) expr(e ast.Expr, inType super.Type) (sem.Expr, super.Type) {
```
nwt, line 23: Nit: Maybe call the super.Type parameter eType instead of inType in keeping with the x/xType naming pattern that appears throughout this method?

```suggestion
func (t *translator) expr(e ast.Expr, eType super.Type) (sem.Expr, super.Type) {
```

compiler/semantic/expr.go
```diff
@@ -19,31 +19,31 @@ import (
...
-		return t.binaryExpr(e)
+		return t.binaryExpr(e, inType)
 	case *ast.BetweenExpr:
-		val := t.expr(e.Expr)
-		lower := t.expr(e.Lower)
-		upper := t.expr(e.Upper)
+		val, valType := t.expr(e.Expr, inType)
+		lower, lowerType := t.expr(e.Lower, inType)
+		upper, upperType := t.expr(e.Upper, inType)
 		// Copy val so an optimizer change to one instance doesn't affect the other.
```
nwt, line 48: This sounds like a good idea but I don't see val being copied, so add that?

compiler/semantic/expr.go
```diff
@@ -116,89 +124,98 @@ func (t *translator) expr(e ast.Expr) sem.Expr {
...
 		funcDecl := t.resolver.newFuncDecl("lambda", e, t.scope)
 		return &sem.FuncRef{
 			Node: e,
 			ID:   funcDecl.id,
-		}
+		}, t.checker.unknown
 	case *ast.MapExpr:
 		var entries []sem.Entry
+		var keyTypes []super.Type
+		var valTypes []super.Type
```
nwt, line 167: Nit:

```suggestion
		var keyTypes, valTypes []super.Type
```

compiler/semantic/checker.go
mattnibs: I also think you can get rid of the special case for `+` and just do arithmetic for `+`. This was here because `+` used to be able to concat strings together but now it can't.

nwt: > I also think you can get rid of the special case for `+`

Nice!

compiler/semantic/expr.go
mccanne: It's the input type to eval not the type of `e`.

compiler/semantic/expr.go
```diff
@@ -585,82 +639,92 @@ func (t *translator) isIndexOfThis(lhs, rhs sem.Expr) *sem.ThisExpr {
 	return nil
 }
 
-func (t *translator) exprNullable(e ast.Expr) sem.Expr {
+func (t *translator) exprNullable(e ast.Expr, inType super.Type) (sem.Expr, super.Type) {
 	if e == nil {
-		return nil
+		return nil, super.TypeNull
```
nwt, line 645: Is there a specific reason to use super.TypeNull rather than t.checker.unknown?

compiler/semantic/expr.go
```diff
@@ -585,82 +639,92 @@ func (t *translator) isIndexOfThis(lhs, rhs sem.Expr) *sem.ThisExpr {
...
 	} else if e != nil {
 		out = sem.NewStructuredError(c, "case: no clause matched and no else provided", e)
+		elseType = t.checker.unknown
 	} else {
 		out = sem.NewStringError(c, "case: no clause matched and no else provided")
+		elseType = t.checker.unknown
 	}
+	types := []super.Type{elseType}
 	for i := len(c.Whens) - 1; i >= 0; i-- {
 		when := c.Whens[i]
```

```suggestion
	 for _, when := range slices.Backwards(c.Whens) {
```

compiler/semantic/expr.go
```diff
@@ -676,62 +740,71 @@ func (t *translator) semCallByName(call *ast.CallExpr, name string, args []sem.E
...
+			return t.resolver.mustResolveCall(ref, ref.ID, args, argTypes)
 		case *funcDecl:
-			return t.resolver.mustResolveCall(call, ref.id, args)
+			return t.resolver.mustResolveCall(call, ref.id, args, argTypes)
 		case *constDecl, *queryDecl:
 			t.error(call, fmt.Errorf("%q is not a function", name))
-			return badExpr
+			return badExpr, t.checker.unknown
+		case thunk:
+			// We're calling a function that is an user-operator parameter.
```
```suggestion
			// We're calling a function that is a user-operator parameter.
```

compiler/semantic/expr.go
```diff
@@ -676,62 +740,71 @@ func (t *translator) semCallByName(call *ast.CallExpr, name string, args []sem.E
...
+			// It must be bound to a function.
+			f, _ := t.resolveThunk(ref, inType)
+			if ref, ok := f.(*sem.FuncRef); ok {
+				return t.resolver.mustResolveCall(call, ref.ID, args, argTypes)
+			}
+			t.error(call, fmt.Errorf("function called via parameter %q is not a function", name))
+			return badExpr, t.checker.unknown
```
nwt: Nit: Indent the error path.

```suggestion
			f, _ := t.resolveThunk(ref, inType)
			ref, ok := f.(*sem.FuncRef)
			if !ok {
				t.error(call, fmt.Errorf("function called via parameter %q is not a function", name))
				return badExpr, t.checker.unknown
			}
			return t.resolver.mustResolveCall(call, ref.ID, args, argTypes)
```

compiler/semantic/expr.go
```diff
@@ -676,62 +740,71 @@ func (t *translator) semCallByName(call *ast.CallExpr, name string, args []sem.E
...
 		pattern, ok := t.maybeEvalString(args[0])
 		if !ok {
-			return sem.NewCall(call, "grep", args)
+			return sem.NewCall(call, "grep", args), super.TypeBool
 		}
 		re, err := expr.CompileRegexp(pattern)
 		if err != nil {
 			t.error(call.Args[0], err)
-			return badExpr
+			return badExpr, t.checker.unknown
```
nwt, line 795: This pattern appears repeatedly.

```go
t.error(...)
return badExpr, t.checker.unknown
```

Have you considered adding

```go
func (t *translator) error(n ast.Node, err error) (sem.Expr, super.Type) {
	t.reporter.error(n, err)
 	return badExpr, t.checker.unknown
}
```

so you can replace that pattern with

```go
return t.error(...)
```

?

compiler/semantic/expr.go
mccanne: In theory, null is more strict so more thorough at type checking, e.g., you can't deref a null but you can deref an unknown though this is sort of a corner case because if the expression isn't there it's not processed but we do pass the type in some cases.

compiler/semantic/expr.go
mccanne: that doesn't compile

compiler/semantic/expr.go
mccanne: t.error() is not specialized to expr and is used lots of places that isn't expr.  Generating a specialized return value inside a general error function seems to me like mixing things that don't go together.

compiler/semantic/checker.go
mccanne: This is a good idea but let's do it on a subsequent PR and add the types to the error (like I did for plus) in the arithmetic case.

mccanne: I made a note to finish this branch after we merge this PR

https://github.com/brimdata/super/compare/new-checker...merge-plus-into-arithmetic

compiler/semantic/resolver.go
```diff
@@ -159,16 +154,22 @@ func (r *resolver) resolveFixed(d *funcDecl, tag string) *funcDef {
...
 }
 
-func (r *resolver) lookupFixed(id string) string {
+func (r *resolver) lookupFixed(id string, inType super.Type) (string, super.Type) {
 	if tag, ok := r.fixed[id]; ok {
-		return tag
+		if f, ok := r.funcs[tag]; ok {
+			return tag, r.t.checker.expr(inType, f.body)
+		}
+		// This function is recursive so we can't know it's type here.
```
nwt:
```suggestion
		// This function is recursive so we can't know its type here.
```

compiler/semantic/schema.go
```diff
@@ -296,10 +303,17 @@ func (s *selectScope) resolveUnqualified(col string) (field.Path, bool, error) {
...
+		// higher than the input table regardless of "pramga pg".
+		// Otherwise, this query resolves the order-by to the input
+		//   pragma pg
+		//   select a as b from (values (0,3),(1,2),(2,0)) T(a,b)
+		//   order by b
+		// but it should use the output table b for order-by.
+		// This isn't a problem when the table has aggregate output (!isGrouped())
+		// (because there is no way to reference the input table values outside
+		// of agg func arguments), and it can't work because expression matching
+		// would be foiled by paths with out.col..
```
```suggestion
		// would be foiled by paths with out.col.
```

compiler/semantic/scope.go
```diff
@@ -109,32 +126,19 @@ func (s *Scope) lookupFuncDeclOrParam(name string) (string, error) {
...
-
-func (s *Scope) resolveThis(t *translator, n ast.Node) sem.Expr {
+func (s *Scope) resolveThis(t *translator, n ast.Node, inType super.Type) (sem.Expr, super.Type) {
 	if e := s.sql.this(n, nil); e != nil {
-		return e
+		return e, t.checker.expr(inType, e)
 	}
 	t.error(n, errors.New("cannot reference 'this' in this context"))
-	return badExpr
+	return badExpr, badType
```
nwt: Nit: Indent the error path.

compiler/semantic/scope.go
```diff
@@ -252,23 +261,22 @@ func extend(n ast.Node, e sem.Expr, rest []string) sem.Expr {
...
+func resolveLateralColumn(t *translator, scope *selectScope, col string, inType super.Type) (sem.Expr, super.Type) {
+	if !scope.lateral {
 		// lateral columns available only inside select bodies
-		return nil
+		return nil, t.checker.unknown
 	}
-	for k := range len(sel.columns) {
-		c := sel.columns[k]
+	for k := range len(scope.columns) {
+		c := scope.columns[k]
```

```suggestion
	for _, c := range scope.columns {
```

compiler/semantic/op.go
```diff
@@ -571,84 +601,64 @@ func (t *translator) semOp(o ast.Op, seq sem.Seq) sem.Seq {
...
 		var expr sem.Expr
+		var fields []super.Field
 		if o.Expr == nil {
 			alias = "count"
 			expr = &sem.RecordExpr{
 				Elems: []sem.RecordElem{
 					&sem.FieldElem{Name: "that", Value: sem.NewThis(nil, nil)},
 				},
 			}
+			fields = []super.Field{{Name: "that", Type: inType}}
```
nwt: Nit:

```suggestion
			fields = []super.Field{super.NewField("that", inType}}
```

compiler/semantic/op.go
```diff
@@ -658,21 +668,41 @@ func (t *translator) semOp(o ast.Op, seq sem.Seq) sem.Seq {
...
+					return append(seq, &sem.CountOp{
+						Node:  o,
+						Alias: alias,
+						Expr:  expr,
+					}), t.checker.unknown
+				}
+				fields = slices.Clone(recType.Fields)
 			}
 		}
+		fields = append(fields, super.Field{Name: alias, Type: super.TypeInt64})
```
nwt: Nit:

```suggestion
		fields = append(fields, super.NewField(alias, super.TypeInt64))
```

compiler/semantic/op.go
```diff
@@ -816,14 +866,17 @@ func (t *translator) semOp(o ast.Op, seq sem.Seq) sem.Seq {
...
+		return append(seq, op), typ
 	case *ast.RenameOp:
 		var assignments []sem.Assignment
+		var paths []pathType
 		for _, fa := range o.Args {
-			assign := t.assignment(&fa)
-			if !isLval(assign.RHS) {
+			assign, path := t.assignment(&fa, inType)
+			_, ok := isLval(assign.RHS)
+			if !ok {
```
nwt: Nit:
```suggestion
			if _, ok := isLval(assign.RHS); !ok {
```

compiler/semantic/op.go
```diff
@@ -999,47 +1027,173 @@ func (t *translator) semOp(o ast.Op, seq sem.Seq) sem.Seq {
...
+	return sem.Seq{fork, join}, typ
+}
+
+func (t *translator) pipeJoinCond(cond ast.JoinCond, leftAlias, rightAlias string, inType super.Type) sem.Expr {
+	switch cond := cond.(type) {
+	case *ast.JoinOnCond:
+		e, typ := t.expr(cond.Expr, inType)
+		t.checker.boolean(cond.Expr, typ)
+		// hack: e is wrapped in []sem.Expr to work around CanSet() model in WalkT
+		dag.WalkT(reflect.ValueOf([]sem.Expr{e}), func(e *sem.ThisExpr) *sem.ThisExpr {
```
nwt, line 1112: Tests pass with `reflect.ValueOf(e)`. Was there something else that motivated this?

compiler/semantic/op.go
```diff
@@ -999,47 +1027,173 @@ func (t *translator) semOp(o ast.Op, seq sem.Seq) sem.Seq {
...
+		expr, _ = t.expr(op.Expr, inType)
+	}
+	var cases []sem.Case
+	for _, c := range op.Cases {
+		var e sem.Expr
+		if c.Expr != nil {
+			e, _ = t.expr(c.Expr, inType)
+		} else if op.Expr == nil {
+			// c.Expr == nil indicates the default case,
+			// whose handling depends on p.Expr.
```
nwt: `p` isn't in scope here so what's `p.Expr`?

runtime/ztests/op/user-cycle.yaml
```diff
@@ -7,6 +7,6 @@ spq: |
 vector: true
 
 error: |
-  operator cycle found: callA -> callB -> callC -> callA at line 3, column 12:
+  operator cycle found: callA -> callB -> callC -> callA -> etc... at line 3, column 12:
```
nwt, line 10: Nit: I don't think we should add "etc..." here. "callA -> callB -> callC -> callA" illustrates the cycle quite clearly.

compiler/semantic/sql.go
```diff
@@ -48,45 +45,78 @@ func (t *translator) sqlSelect(sel *ast.SQLSelect, demand []ast.Expr, seq sem.Se
...
+		var typ super.Type
+		where, typ = t.expr(sel.Where, inType)
+		t.checker.boolean(sel.Where, typ)
 	}
 	scope.lateral = true
-	scope.groupings = t.groupBy(scope, sel.GroupBy)
+	scope.groupings = t.groupBy(scope, sel.GroupBy, inType)
 	scope.aggOk = true
-	// Make sure any aggs needed by ORDER BY and computed.
+	// Make sure any aggs needed by ORDER BY and HAVING computed and
```
```suggestion
	// Make sure any aggs needed by ORDER BY and HAVING are computed and
```

compiler/semantic/sql.go
```diff
@@ -260,11 +288,14 @@ func groupTmp(k int) string {
 	return fmt.Sprintf("k%d", k)
 }
 
-func (t *translator) selectFrom(loc ast.Loc, exprs []ast.SQLTableExpr, seq sem.Seq) (sem.Seq, relScope) {
+// selectFrom analyzes the FROM clause which is a cross-joined set of table expressions.
+// inType is immediate pipe parent in a top-level select, nil otherwise
```
```suggestion
// inType is from the immediate pipe parent in a top-level select, nil otherwise.
```

compiler/semantic/sql.go
```diff
@@ -260,11 +288,14 @@ func groupTmp(k int) string {
...
+func (t *translator) selectFrom(loc ast.Loc, exprs []ast.SQLTableExpr, seq sem.Seq, inType super.Type) (sem.Seq, relScope) {
 	if len(exprs) == 0 {
 		// No FROM clause is modeled by a single null value, which we represent
-		//  with dynamic, e.g., so "select this" results in {that:null}.
-		return seq, &dynamicTable{}
+		//  with dynamic, e.g., so "select this" results in {that:null},
+		// or its the parent pipe input.
```
```suggestion
		// with dynamic, e.g., so "select this" results in {that:null},
		// or the parent pipe input.
```

compiler/semantic/sql.go
```diff
@@ -282,47 +313,37 @@ func (t *translator) selectFrom(loc ast.Loc, exprs []ast.SQLTableExpr, seq sem.S
...
 	// Handle comma-separated table expressions in FROM clause.
 	for _, e := range exprs[1:] {
 		seq, scope = t.sqlAppendCrossJoin(e, seq, scope, e)
 	}
 	return seq, scope
 }
 
-func (t *translator) sqlValues(values *ast.SQLValues, seq sem.Seq) (sem.Seq, tableScope) {
+// sqlValues analyzes the values expression which currently cannot depend
+// on the input data (until we add support for lateral joins then a values
```
```suggestion
// on the input data (until we add support for lateral joins; then a values
```

compiler/semantic/sql.go
```diff
@@ -282,47 +313,37 @@ func (t *translator) selectFrom(loc ast.Loc, exprs []ast.SQLTableExpr, seq sem.S
...
-		}
-		fuser.Mixin(val.Type())
-	}
-	recType, ok := super.TypeUnder(fuser.Type()).(*super.TypeRecord)
+	return seq, t.inferSchema(values, types)
+}
+
+// With type checking integrated into the SQL analysis, the logic here
+// can always infer a staticTable.  Eventually, this will need to support
+// lateral join subqueries which have runtime values (but staticTable columns still known).
```
```suggestion
// lateral join subqueries which have runtime values (but staticTable columns are still known).
```

compiler/ztests/head.yaml
```diff
@@ -24,6 +24,9 @@ outputs:
       expected integer at line 1, column 6:
       head "1"
            ~~~
-      cannot reference 'x' in constant expression at line 1, column 6:
+      expected integer at line 1, column 6:
+      head x
+           ~
+      no such field "x" at line 1, column 6:
       head x
            ~
```
nwt, line 32: Nit: This error feels misleading since by definition a constant expression can't include a field.

compiler/semantic/op.go
mccanne: I don't understand either.  Removing comment.

runtime/ztests/op/user-cycle.yaml
mccanne: This isn't true in general.  We can make an issue to put in extra logic to make a better error message but I thought this was ok.

compiler/ztests/head.yaml
mccanne: I'll make a note to improve this error message in a subsequent PR.

runtime/ztests/op/user-cycle.yaml
mccanne: I'll make a note to improve this error message in a subsequent PR.

PR #6534, opened 2026-01-15
vam: Add support for has_error() function

mattnibs:


runtime/vam/expr/function/types.go
```diff
@@ -1,11 +1,81 @@
...
+	case *vector.Map:
+		keys := listHasError(index, vec.Offsets, h.hasError(vec.Keys))
+		vals := listHasError(index, vec.Offsets, h.hasError(vec.Values))
+		return expr.EvalOr(nil, keys, vals)
+	default:
+		return vector.Apply(true, IsErr{}.Call, in)
+	}
+}
+
+func listHasError(index, offsets []uint32, inner vector.Any) vector.Any {
```
nwt: Nit: Since this is similar to search.evalForList, use a similar signature.

```suggestion
func listHasError(inner vector.Any, offsets, index []uint32) vector.Any {
```

runtime/vam/expr/function/types.go
```diff
@@ -1,11 +1,81 @@
...
+		start, end := offsets[idx], offsets[idx+1]
+		n := end - start
+		if n == 0 {
+			continue
+		}
+		index2 = slices.Grow(index2[:0], int(n))[:n]
+		for k := range n {
+			index2[k] = k + start
+		}
+		view := vector.Pick(inner, index2)
```
nwt, line 76: Reusing the index2 slice across calls to vector.Pick is unsafe. That function can return a vector containing a reference to its parameter's backing array, so writing to slice2 can modify vectors returned by previous Pick calls.

mattnibs: FlattenBool is written in a way that it doesn't retain a reference to `index2` and therefore its use is discarded before the next iteration. There is a risk here that future versions of `FlattenBool` might somehow retain a reference but I don't think that's likely since given the point of flatten to flatten into a single boolean vector. The fix would be that we just allocate index2 on every iteration but I think we can agree that isn't ideal. Thoughts?

nwt: Good point. Totally safe. How about adding a comment like "Reusing index2 across calls here is safe because view does not escape this loop body."?

mattnibs: Good call

PR #6535, opened 2026-01-16
Enable more tests for vector runtime

mattnibs:
Move and enable more tests that now work in the vector runtime.


mattnibs:
@nwt they aren't enabled for any runtime! I just don't like have these tests in two places and am going to remove the folder `runtime/sam/expr/ztests` soon

nwt:
@mattnibs: runtime/ztests/expr/in-star.yaml would, if fixed, duplicate runtime/sam/expr/ztests/in-star.yaml so how about removing it instead of moving it.

PR #6538, opened 2026-01-16
sam: Filter any error value

### LINKED ISSUE #6521: SQL: error("missing") in sequential runtime on "HAVING NULL =" query
The following sqllogictest query gives an unexpected `error("missing")` result in sequential runtime.

```
$ super -f parquet -o tab0 -c "values {col0:83,col1:0,col2:38},{col0:26,col1:0,col2:79},{col0:43,col1:81,col2:24}" &&
  super -c "SELECT DISTINCT tab0.col1 FROM tab0 GROUP BY col1 HAVING NULL = + col1 / tab0.col1;"

{col1:error("missing")}
```

In vector runtime, it matches Postgres in returning no output (though as is sometimes the case, maybe it's doing so for the wrong reasons).

```
$ SUPER_VAM=1 super -c "SELECT DISTINCT tab0.col1 FROM tab0 GROUP BY col1 HAVING NULL = + col1 / tab0.col1;"

[no output]
```

## Details

Repro is with super commit 05e88a9. This query is from sqllogictest [random/groupby/slt_good_7/q2952](https://github.com/brimdata/sqllogic-ztests/blob/main/ztests/sqlite/random/groupby/slt_good_7/ztests-1/q2952.fail)

Here's Postgres returning nothing on the query:

```
$ psql postgres
psql (17.7 (Homebrew))
Type "help" for help.

postgres=# CREATE TABLE tab0(col0 INTEGER, col1 INTEGER, col2 INTEGER);
CREATE TABLE
postgres=# INSERT INTO tab0 VALUES(83,0,38);
INSERT 0 1
postgres=# INSERT INTO tab0 VALUES(26,0,79);
INSERT 0 1
postgres=# INSERT INTO tab0 VALUES(43,81,24);
INSERT 0 1
postgres=# SELECT DISTINCT tab0.col1 FROM tab0 GROUP BY col1 HAVING NULL = + col1 / tab0.col1;
 col1 
------
(0 rows)
```

This is just me hacking, but picking apart what I see in `super compile`:

```
$ super -version
Version: 05e88a9fb

$ super compile -O -C "SELECT DISTINCT tab0.col1 FROM tab0 GROUP BY col1 HAVING NULL = + col1 / tab0.col1;"
file tab0 unordered fields col1
| aggregate
     by k0:=col1
| where null==k0/k0
| values {g:this,out:{col1:k0}}
| distinct out
| values out
| output main
```

The "divide by zero" stands out:

```
$ super -c "
from tab0
| aggregate
     by k0:=col1
| where null==k0/k0"

error("divide by zero")
```

```
$ SUPER_VAM=1 super -c "
from tab0
| aggregate
     by k0:=col1
| where null==k0/k0"

[no output]
```

I asked Claude AI to speculate on this one a bit, and it theorized that Postgres may be using query optimization to recognize that a `HAVING NULL =` expression can never be true, so it short-circuits or skips evaluating the division entirely. Maybe our vector runtime is doing something similar, but sequential is not?

mattnibs:
This commit changes the behavior of filters in sequential runtime where any error values encountered in a where clause will be filtered. Previously filters would only filter error("missing") or error("quiet"). Filters in vector runtime already behaved in this manner.

Fixes #6521


runtime/sam/expr/filter.go
```diff
@@ -214,11 +214,8 @@ func NewFilterApplier(sctx *super.Context, e Evaluator) Evaluator {
...
-	if val.Type().ID() == super.IDBool {
-		if val.Bool() {
-			return this
-		}
-		return f.sctx.Missing()
+	if val.Type().ID() == super.IDBool && val.Bool() {
+		return this
 	}
-	return val
+	return f.sctx.Missing()
```
nwt: Nits: There's no longer any reason to use EvalBool. And indent the "error" path.

```go
	val := f.expr.Eval(this)
	if val.Type().ID() != super.IDBool || !val.Bool() {
		return f.sctx.Missing()
	}
	return this
```

mattnibs: Nice!

PR #6543, opened 2026-01-19
Edits to super-sql/expressions docs

philrz:


book/src/super-sql/expressions/intro.md
```diff
@@ -1,31 +1,31 @@
...
-[`sort`](../operators/sort.md) and so forth all utilize various expressions
+For example, [values](../operators/values.md), [where](../operators/where.md),
+[cut](../operators/cut.md), [put](../operators/put.md),
+[sort](../operators/sort.md) and so forth all utilize various expressions
 as part of their semantics.
 
-Likewise, the projected columns of a
-[`SELECT`](../sql/select.md) from the very same expression syntax
+Likewise, the projected columns of a SQL
+[SELECT](../sql/select.md) come from the very same expression syntax
```
philrz: There definitely was a word missing here, but please confirm "come" is the right addition.

book/src/super-sql/expressions/intro.md
```diff
@@ -34,29 +34,29 @@ Expressions are composed from operands and operators over operands.
...
-  * [arithmetic](./arithmetic.md) to add, subtract, multiply, divide, etc,
+  * [arithmetic](./arithmetic.md) to add, subtract, multiply, divide, etc.,
   * [cast](cast.md) to convert values from one type to another,
-  * [comparisons](comparisons.md) to compare two values resulting in a Boolean,
+  * [comparisons](comparisons.md) to compare two values resulting in a [Boolean](../types/bool.md),
   * [concatenation](concat.md) of strings,
-  * [conditionals](conditional.md) including C-style `?-:` operator and SQL `CASE` expressions,
-  * [containment](containment.md) to test for the existing value inside an array or set,
-  * [dot](dot.md) to access a field of a record (or a SQL column of a table),
+  * [conditionals](conditional.md) including C-style `?:` operator and SQL `CASE` expressions,
```
philrz, line 47: There was some inconsistency between showing `?-:` here vs. `?:` that appears in the **Precedence** list at the bottom of this same page. Since `?:` seems like the common shorthand in wider web search hits that talk about C-style ternary, I've proposed going with `?:` here.

book/src/super-sql/expressions/containment.md
```diff
@@ -7,27 +7,22 @@ a value in another value and has the form
...
 The inverse of `in` has the syntax
 ```
 <item> not in <target>
 ```
-and is true when `<item>` is not contained in the `<target>`.
-
->[!NOTE]
-> The `in` operator currently does not support SQL NULL semantics in that
-> `1 not in [2,NULL]` is false instead of NULL.  This will be fixed
-> in a future version.
```
philrz, line 30: I confirmed that we can drop this note because this was addressed in #6425.

```
$ super -version
Version: ed5f009bf

$ super -c "values 1 not in [2,NULL]"
null::bool
```

book/src/super-sql/expressions/exists.md
```diff
@@ -1,6 +1,6 @@
 ## Exists
 
-The `exists` operator is Boolean-valued function that tests whether
+The [Boolean](../types/bool.md) predicate `exists` tests whether
```
philrz: I paused at seeing "operator" and "function" both used to describe `exists`, since we carefully define the two as different things elsewhere in the language. I bounced this off Claude AI and it was sympathetic and saw it as a kind of hybrid thing. After some back & forth, it seemed like we could get by describing it as a "predicate" as I've attempted here, but let me know if that's inappropriate for some reason.

book/src/super-sql/expressions/inputs.md
```diff
@@ -29,11 +29,11 @@ tables or columns are referenced.
...
 then the input data can still be referred to as the value `this` and placed
 into an output relation using [SELECT](../sql/select.md).
-When referring to non-relational with `*`, there are no input columns and
+When referring to non-relational inputs with `*`, there are no columns and
 thus the select value is empty, i.e., the value `{}`.
 
 When non-record data is referenced in a SQL operator and the input
-schema is dynamic and unknown, runtime errors like `error("missing")`
+schema is dynamic and unknown, runtime [errors](../types/error.md) like `error("missing")`
 will generally arise and be present in the output data.
```
philrz, line 37: It seems like some more detail would help here. I started digging in when I sensed a missing word after "non-relational" and so proposed moving the "input" word forward, but please confirm.

Then when trying to trigger the effects described, I don't see the empty `{}` or `error("missing")` implied by the text. But maybe my example is bad and this is a sign we just need to add examples.

```
$ cat data.json 
[1,2,3]

$ super -version
Version: ed5f009bf

$ super -c "SELECT * FROM data.json"
the all-columns (*) pattern cannot be used for dynamic inputs at line 1, column 8:
SELECT * FROM data.json
       ~
```

book/src/super-sql/expressions/cast.md
```diff
@@ -144,7 +144,7 @@ then the named type is reattached to the result.
 
 ### Errors
 
-Casts attempted between a value and a type that are not defined
+Casts attempted between a value and a type that are not permitted
```
philrz: "defined" read strangely to me here. I assume it could be interpreted as "defined within the code" or "defined within a spec", but reading it as a user "permitted" seemed more appropriate. (AI is also suggesting "supported" might be ok.)

mccanne: How about "Casting a value to an incompatible type results in a structured..."

book/src/super-sql/expressions/conditional.md
```diff
@@ -50,13 +50,13 @@ is generated at run time as the result of the conditional expression.
...
 
-The ternary form follows the C language and has syntax
+The [ternary](https://en.wikipedia.org/wiki/Ternary_conditional_operator) form follows the C language and has syntax
 ```
 <predicate> ? <true-expr> : <false-expr>
 ```
 where `<predicate>` is a Boolean-valued expression
 and `<true-expr>` and `<false-expr>` are any [expressions](intro.md).
-When `<predicate>` is true, then `<true-expr>` is evaluated and becomes
+When `<predicate>` is `true`, then `<true-expr>` is evaluated and becomes
```
mccanne: Some of the instances of "true" I would leave plain.  If you are talking about the specific Boolean value `true` then it can be back-quoted.  I would think here if you are saying `<predicate>` is `true`, it means the expression is literally the text `true` whereas saying `<predicate>` is true means the condition resulted in a true outcome, e.g., `values 1 > 0` produces the literal value `true` but the comparison "is 1 greater than 0?" is true in the abstract sense of the word.

book/src/super-sql/expressions/index.md
```diff
@@ -17,24 +17,24 @@ Indexable entities include
 [strings](../types/string.md), and
 [bytes](../types/bytes.md).
 
-If `<entity>` is a record, then the `<index>` operand
-must be coercible to a string and the result is the record's field
+If `<entity>` is a `record`, then the `<index>` operand
```
mccanne: I don't think record, array, etc should be in backticks since they aren't language keywords.

book/src/super-sql/expressions/inputs.md
```diff
@@ -29,11 +29,11 @@ tables or columns are referenced.
 In a SQL operator, if the input is not a record (i.e., not relational),
 then the input data can still be referred to as the value `this` and placed
 into an output relation using [SELECT](../sql/select.md).
-When referring to non-relational with `*`, there are no input columns and
+When referring to non-relational inputs with `*`, there are no columns and
```
mccanne, line 32: `*` is no longer allowed with dynamic inputs

book/src/super-sql/expressions/intro.md
mccanne: How about "utilize"

philrz: @mccanne: "utilize" works for me. Committing.

book/src/super-sql/expressions/inputs.md
philrz: @mccanne: Ok, so trying to turn that into action, does that mean both of these sentences currently in these "Inputs" docs should just be deleted?

> When referring to non-relational inputs with `*`, there are no columns and thus the select value is empty, i.e., the value `{}`.

> When non-record data is referenced in a SQL operator and the input schema is dynamic and unknown, runtime errors like `error("missing")` will generally arise and be present in the output data.

book/src/super-sql/expressions/conditional.md
philrz: @mccanne: Ok, that makes sense to me, so after looking over all my changes with that in mind I just de-backticked the instances of "true" I'd previously changed in `comparisons.md`, `conditional.md`, and `containment.md` so they're effectively back to the way you had them originally. I actually ended up de-backticking one of your original ones as I was doing that since the prior doc had:

> "...are evaluated in order until a true result is encountered..."

and
> "If there is no `true` result..."

both in the same paragraph, and it seems that from the proposed convention they'd both want to be without backtick.

I also pushed a change to the [Conventions](https://github.com/brimdata/super/blob/review-super-sql-expr-docs/book/README.md#fixed-width-font) as a reminder to stick to this going forward.

Please do look things over and let me know if you think I've tilted it too far in the other direction, though.

book/src/super-sql/expressions/exists.md
philrz: Per offline discussion, this has been simplified to just:

> The `exists` operator tests whether a subquery has a non-empty result...

book/src/super-sql/expressions/inputs.md
philrz: @mccanne has indicated he'll clarify this in his upcoming SQL docs update, so I'm leaving this text as it is for now and will revisit in the context of his PR when it's up.

book/src/super-sql/expressions/index.md
philrz: @mccanne: Ok, I put things back to the way you had them before.

PR #6550, opened 2026-01-20
add support for CTEs with column aliases

### LINKED ISSUE #6016: SQL: CTE with column aliasing

mccanne:
Closes #6016

### DIFF
--- a/compiler/ast/sql.go
+++ b/compiler/ast/sql.go
@@ -49,7 +49,7 @@ func (*SQLValues) sqlQueryBodyNode() {}
 
 type (
 	SQLCTE struct {
-		Name         *ID          `json:"name"`
+		Name         *TableAlias  `json:"name"`
 		Materialized bool         `json:"materialized"`
 		Body         SQLQueryBody `json:"body"`
 		Loc          `json:"loc"`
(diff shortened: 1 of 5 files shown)


mccanne:
In a subsequent PR, let's fix TableAlias to use  SQLIdentifier instead of IdentifierName (it was already broken)
,

mattnibs (approved):

PR #6555, opened 2026-01-21
wire pipe output type to sql input in type checker

mccanne:
This commit fixes a bug where the type output from a pipe sequence was not properly forwarded to the
input a SQL operator for type checking.

Additionally a SQL operator with neither a FROM clause nor a pipe parent was using type unknown instead of empty table thus not properly type checking missing column references.


compiler/ztests/sql/select-x.yaml
```diff
@@ -0,0 +1,15 @@
...
+         ~
+
+---
+
+spq: values {a:1} | select x
+
+error: |
+  "x" is not a column or table at line 1, column 23:
+  values {a:1} | select x
+                        ~
```
nwt: Nit: Add these tests to compiler/ztests/sql/select-nonexistent-column.yaml instead of a new file.

PR #6557, opened 2026-01-22
Simplify logic for unnest <expr> into <query>

mattnibs:
This commit changes the behavior for unnest into for the runtime. Instead of creating scopes in the runtime, the semantic pass rewrites unnest into as unnest [unnest <expr> | <query>].


compiler/optimizer/optimizer.go
```diff
@@ -73,27 +73,23 @@ func removePassOps(seq dag.Seq) dag.Seq {
 }
 
 func Walk(seq dag.Seq, post func(dag.Seq) dag.Seq) dag.Seq {
-	return walk(seq, true, post)
+	return walk(seq, post)
 }
 
-func walk(seq dag.Seq, over bool, post func(dag.Seq) dag.Seq) dag.Seq {
+func walk(seq dag.Seq, post func(dag.Seq) dag.Seq) dag.Seq {
```
nwt: Nit: There's no longer any reason to have both Walk and walk so combine them.

compiler/rungen/op.go
nwt: Nit: Since you're removing compileVamUnnest, remove this too.

compiler/ztests/merge-filters.yaml
```diff
@@ -2,8 +2,9 @@ script: |
   super compile -C -O 'from /dev/null | where a | where b'
   echo ===
   super compile -C -O -dynamic 'fork ( from /dev/null | where b | where c ) ( from /dev/zero | where e | where f ) | where g'
-  echo ===
-  super compile -C -O 'from /dev/null | unnest [a] into ( where b | where c )'
+  # This test is disable until we can enable merging filters in subquery expressions.
+  # echo ===
+  # super compile -C -O 'from /dev/null | unnest [a] into ( where b | where c )'
```
nwt: I think you can keep this working with this.

```diff
diff --git a/compiler/optimizer/optimizer.go b/compiler/optimizer/optimizer.go
index 97d228749..7893336b3 100644
--- a/compiler/optimizer/optimizer.go
+++ b/compiler/optimizer/optimizer.go
@@ -40,7 +40,7 @@ func New(ctx context.Context, env *exec.Environment) *Optimizer {
 // Note: mergeFilters does not descend into dag.OverExpr.Scope, so it cannot
 // merge filters in "over" expressions like "sum(over a | where b | where c)".
 func mergeFilters(seq dag.Seq) dag.Seq {
-       return walk(seq, func(seq dag.Seq) dag.Seq {
+       walkT(reflect.ValueOf(&seq), func(seq dag.Seq) dag.Seq {
                // Start at the next to last element and work toward the first.
                for i := len(seq) - 2; i >= 0; i-- {
                        if f1, ok := seq[i].(*dag.FilterOp); ok {
@@ -54,6 +54,7 @@ func mergeFilters(seq dag.Seq) dag.Seq {
                }
                return seq
        })
+       return seq
 }

 func removePassOps(seq dag.Seq) dag.Seq {
```

runtime/vam/op/unnest.go
```diff
@@ -96,7 +97,15 @@ func (u *Unnest) flatten(vec vector.Any, slot uint32) vector.Any {
...
-		return vector.NewWrappedError(u.sctx, "unnest: encountered non-array value", vec)
+		if c, ok := vec.(*vector.Const); ok && c.Value().IsNull() {
+			return nil
+		}
+		errvec := vector.NewWrappedError(u.sctx, "unnest: encountered non-array value", vec)
+		if nulls := vector.NullsOf(vec); !nulls.IsZero() {
+			index := roaring.FromDense(nulls.GetBits(), false).ToArray()
+			return vector.Pick(vec, index)
+		}
+		return errvec
```
nwt: This isn't right. It does this.

```console
$ echo 1 2 | go run ./cmd/super -c 'unnest this' -
error({message:"unnest: encountered non-array value",on:1})
error({message:"unnest: encountered non-array value",on:2})

$ echo 1 2 | SUPER_VAM=1 go run ./cmd/super -c 'unnest this' -
error({message:"unnest: encountered non-array value",on:1})
error({message:"unnest: encountered non-array value",on:2})
error({message:"unnest: encountered non-array value",on:1})
error({message:"unnest: encountered non-array value",on:2})
```

And this.

```console
$ echo 1 null::int64 | go run ./cmd/super -c 'unnest this' -
error({message:"unnest: encountered non-array value",on:1})

$ echo 1 null::int64 | SUPER_VAM=1 go run ./cmd/super -c 'unnest this' -
null::int64
null::int64
```

I think this is what we want.

```suggestion
		if vector.NullsOf(vec).IsSet(slot) {
			return nil
		}
		slotVec := vector.Pick(vec, []uint32{slot})
		return vector.NewWrappedError(u.sctx, "unnest: encountered non-array value", slotVec)
```

compiler/ztests/merge-filters.yaml
mattnibs: I tried this and it causes stuff to fail. A simple example is failing to merge filters this query:
```
$ super compile -C -O 'from /dev/null | where a | where b'
file /dev/null filter (a)
| where b
| output main
```

runtime/vam/op/unnest.go
mattnibs: doh!

compiler/ztests/merge-filters.yaml
nwt: I think that might be because you wrote `reflect.ValueOf(seq)` instead of `reflect.ValueOf(&seq)`.

mattnibs: Great this works

PR #6558, opened 2026-01-22
Add scalar aggregate for keyless aggregations

### LINKED ISSUE #4708: Explicitly confirming the absence of results
Repro is with Zed commit 70e7129.

We started a debate years ago about this difference between SQL and Zed behaviors. Start with this [simple-comma.csv](https://github.com/brimdata/zed/files/11992248/simple-comma.csv) test data.

```
id,key,value
1,max,hi
2,zap,my
3,patty,name
4,zap,is
5,max,phil
6,gris,these
7,carrot,are
8,thomas,a
9,max,bunch
10,max,of
11,zap,lines
```

In DuckDB, we'll search & count to determine the presence and absence of a particular field value.

```
$ duckdb
v0.8.1 6536a77232
D create table pets as select * from 'simple-comma.csv';
D select count(*) from pets where key='max';
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│            4 │
└──────────────┘
D select count(*) from pets where key='foo';
┌──────────────┐
│ count_star() │
│    int64     │
├──────────────┤
│            0 │
└──────────────┘
```

Compare this with Zed behavior.

```
$ zq -version
Version: v1.8.1-61-g70e71293

$ zq 'grep("max") | count()' simple-comma.csv 
4(uint64)

$ zq 'grep("foo") | count()' simple-comma.csv 
[no output]

$ echo $?
0
```

The last time this was debated, there were differing opinions about if Zed should adapt to SQL's behavior of having `count()` always return a scalar (so, `0` in this case) vs. the way it's currently silent and leaving that for the user to interpret.

Having left the topic alone for some time while the project has evolved, here's my updated thinking on the topic. I sense we may not be able to definitively settle on one true behavior. Instead, like we've done in other areas, perhaps we can employ Zed's first-class errors to make it possible to test for this condition.

One thought I had was that if pipeline elements are able to detect the condition that their upstream dataflow has never produced results and has effectively shut down (is that called EOF in the code?), the pipeline element (`count()` in this case) could raise `error(missing)`, so the user would have the option to invoke `quiet()` if they wanted to restore the prior behavior of seeing pure silence when there's no results.

Another approach might be to let pipeline elements behave as they currently do but have an operator with the specific role of detecting this condition & raising such an error. Since we already have a `pass` operator, perhaps this could be a new option on `pass`. For illustrative purposes I'll call it `-mustflow`. So if the user wanted the SQL-like behavior it might look like:

```
$ zq 'grep("foo") | count() | pass -mustflow | yield is_error(this) ? 0 : this' simple-comma.csv
0
```

### LINKED ISSUE #6087: SQL: Adding a column alias for a NULL result causes empty result
In the words of community user @chrismo in a recent [Slack thread](https://super-db.slack.com/archives/CTSMAK6G7/p1754004229114909):

>Adding a column alias for a null result returns nothing at all. Using the default column alias is still fine.
>```
>❯ super --version
>Version: v0.0.0-20250725204616-e75dfdaf1809
>
>❯ super -c "[{id:1,s:'a'},{id:2,s:'b'}] | unnest this | select max(id) where s like '%c%'"
>{"max(id)":null}
>
>❯ super -c "[{id:1,s:'a'},{id:2,s:'b'}] | unnest this | select max(id) as max_id where s like '%c%'"
>[no output]
>```

## Details

Repro is with super commit e75dfda.

Since the user's query relies on some functionality that's unique to SuperSQL, the following Postgres comparison starts with a statically loaded table that contains the same data that came out of the `unnest`.

```
$ psql postgres
psql (17.5 (Homebrew))
Type "help" for help.

postgres=# CREATE TABLE foo(id INTEGER, s VARCHAR);
CREATE TABLE
postgres=# INSERT INTO foo(id, s) VALUES (1, 'a'), (2, 'b');
INSERT 0 2
postgres=# SELECT * FROM foo;
 id | s 
----+---
  1 | a
  2 | b
(2 rows)

postgres=# \pset null 'NULL'
Null display is "NULL".

postgres=# select max(id) from foo where s like '%b%';
 max 
-----
   2
(1 row)

postgres=# select max(id) from foo where s like '%c%';
 max  
------
 NULL
(1 row)

postgres=# select max(id) as max_id from foo where s like '%c%';
 max_id 
--------
   NULL
(1 row)
```

### LINKED ISSUE #6506: SQL: Aggregate without GROUP BY on empty input
As I understand it, the SQL standard indicates there should be at least one row returned on aggregations without `GROUP BY`, even if the input is empty. Instead SuperDB is currently still doing what it's traditionally done in pipe context, which is to return nothing.

## Details

Repro is with super commit a021857. This effect surfaces in sqllogictests such as [random/aggregates/slt_good_0/q343](https://github.com/brimdata/sqllogic-ztests/blob/main/ztests/sqlite/random/aggregates/slt_good_0/ztests-1/q343.fail) and [random/aggregates/slt_good_0/q464](https://github.com/brimdata/sqllogic-ztests/blob/main/ztests/sqlite/random/aggregates/slt_good_0/ztests-1/q464.fail).

Here's two simplified repro examples that both return nothing with `super`:

```
$ super -version
Version: a021857b4

$ super -c "SELECT COUNT(*) WHERE FALSE;"
[no output]

$ super -c "SELECT SUM(1) WHERE FALSE;"
[no output]
```

Whereas in Postgres, these return `0` and `NULL`, respectively.

```
$ psql postgres
psql (17.7 (Homebrew))
Type "help" for help.

postgres=# \pset null 'NULL'
Null display is "NULL".
postgres=# SELECT COUNT(*) WHERE FALSE;
 count 
-------
     0
(1 row)

postgres=# SELECT SUM(1) WHERE FALSE;
 sum  
------
 NULL
(1 row)
```

mattnibs:
This commit adds a separate runtime operator for scalar aggregates. If no input is encountered scalar aggregates return default values for each aggregate value.

Fixes #6506
Closes #4708
Fixes #6087

### DIFF
--- a/runtime/sam/op/aggregate/scalar.go
+++ b/runtime/sam/op/aggregate/scalar.go
@@ -1,0 +1,90 @@
+package aggregate
+
+import (
+	"github.com/brimdata/super"
+	"github.com/brimdata/super/pkg/field"
+	"github.com/brimdata/super/runtime"
+	"github.com/brimdata/super/runtime/sam/expr"
+	"github.com/brimdata/super/sbuf"
+)
+
+type scalarAggregate struct {
+	sctx        *super.Context
+	parent      sbuf.Puller
+	builder     *super.RecordBuilder
+	aggRefs     []expr.Evaluator
+	aggs        []*expr.Aggregator
+	partialsIn  bool
+	partialsOut bool
+	row         valRow
+}
+
+func NewScalar(rctx *runtime.Context, parent sbuf.Puller, aggNames []field.Path, aggs []*expr.Aggregator, partialsIn, partialsOut bool) (sbuf.Puller, error) {
+	builder, err := super.NewRecordBuilder(rctx.Sctx, aggNames)
+	if err != nil {
+		return nil, err
+	}
+	aggRefs := make([]expr.Evaluator, 0, len(aggNames))
+	for _, fieldName := range aggNames {
+		aggRefs = append(aggRefs, expr.NewDottedExpr(rctx.Sctx, fieldName))
+	}
+	return &scalarAggregate{
+		sctx:        rctx.Sctx,
+		parent:      parent,
+		builder:     builder,
+		aggRefs:     aggRefs,
+		aggs:        aggs,
+		partialsIn:  partialsIn,
+		partialsOut: partialsOut,
+		row:         newValRow(aggs),
+	}, nil
+}
+
+func (s *scalarAggregate) Pull(done bool) (sbuf.Batch, error) {
+	if done {
+		s.row = nil
+		return nil, nil
+	}
+	if s.row == nil {
+		s.row = newValRow(s.aggs)
+		return nil, nil
+	}
+	for {
+		batch, err := s.parent.Pull(false)
+		if err != nil {
+			return nil, err
+		}
+		if batch == nil {
+			return s.result(), nil
+		}
+		for _, val := range batch.Values() {
+			if s.partialsIn {
+				s.row.consumeAsPartial(val, s.aggRefs)
+			} else {
+				s.row.apply(s.sctx, s.aggs, val)
+			}
+		}
+	}
+}
+
+func (s *scalarAggregate) result() sbuf.Batch {
+	var typs []super.Type
+	s.builder.Reset()
+	for _, agg := range s.row {
+		var val super.Value
+		if s.partialsOut {
+			val = agg.ResultAsPartial(s.sctx)
+		} else {
+			val = agg.Result(s.sctx)
+		}
+		typs = append(typs, val.Type())
+		s.builder.Append(val.Bytes())
+	}
+	s.row = nil
+	typ := s.builder.Type(typs)
+	b, err := s.builder.Encode()
+	if err != nil {
+		panic(err)
+	}
+	return sbuf.NewBatch([]super.Value{super.NewValue(typ, b)})
+}
--- a/runtime/vam/op/aggregate/scalar.go
+++ b/runtime/vam/op/aggregate/scalar.go
@@ -1,0 +1,98 @@
+package aggregate
+
+import (
+	"github.com/brimdata/super"
+	"github.com/brimdata/super/pkg/field"
+	"github.com/brimdata/super/runtime/vam/expr"
+	"github.com/brimdata/super/runtime/vam/expr/agg"
+	"github.com/brimdata/super/vector"
+	"github.com/brimdata/super/vector/bitvec"
+)
+
+type scalarAggregate struct {
+	parent      vector.Puller
+	sctx        *super.Context
+	aggExprs    []expr.Evaluator
+	aggs        []*expr.Aggregator
+	builder     *vector.RecordBuilder
+	partialsIn  bool
+	partialsOut bool
+
+	funcs []agg.Func
+}
+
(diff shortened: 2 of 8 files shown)


runtime/ztests/op/aggregate/scalar-empty.yaml
```diff
@@ -0,0 +1,24 @@
...
+  | count(*), sum(this), and(this), or(this), min(this), avg(this), collect(this)
+
+vector: true
+
+output: |
+  {count:0,sum:null,and:null::bool,or:null::bool,min:null,avg:null::float64,collect:null}
+
+---
+
+# Test that scalar aggregations with no inputs work with inside of unnest.
```
nwt:
```suggestion
# Test that scalar aggregations with no inputs work in an unnset subquery.
```

philrz:
This tests out well! I ran this branch against the sqllogic-ztests last night at commit f2f6d0b and it allows 31,251 additional queries to execute successfully and return the expected result.

runtime/vam/op/aggregate/aggregate.go
```diff
@@ -26,11 +26,14 @@ type Aggregate struct {
...
 
-func New(parent vector.Puller, sctx *super.Context, aggNames []field.Path, aggExprs []expr.Evaluator, aggs []*expr.Aggregator, keyNames []field.Path, keyExprs []expr.Evaluator, partialsIn, partialsOut bool) (*Aggregate, error) {
+func New(parent vector.Puller, sctx *super.Context, aggNames []field.Path, aggExprs []expr.Evaluator, aggs []*expr.Aggregator, keyNames []field.Path, keyExprs []expr.Evaluator, partialsIn, partialsOut bool) (vector.Puller, error) {
 	builder, err := vector.NewRecordBuilder(sctx, append(keyNames, aggNames...))
 	if err != nil {
 		return nil, err
 	}
+	if len(keyNames) == 0 {
+		return newScalarAggregate(parent, sctx, builder, aggs, aggExprs, partialsIn, partialsOut), nil
+	}
```
nwt: Nit: I think it would make more sense to export newScalarAggregate and then make this decision in rungen.Builder.compileAggregate.

runtime/sam/op/aggregate/scalar.go
```diff
@@ -0,0 +1,81 @@
+package aggregate
+
+import (
+	"github.com/brimdata/super"
+	"github.com/brimdata/super/runtime"
+	"github.com/brimdata/super/runtime/sam/expr"
+	"github.com/brimdata/super/sbuf"
+)
+
+type scalar struct {
```
nwt: Nit: Call this scalarAggregate for consistency with runtime/vam/op/aggregate.scalarAggregate.

runtime/vam/op/aggregate/scalar.go
```diff
@@ -0,0 +1,93 @@
...
+			for _, e := range s.aggExprs {
+				vals = append(vals, e.Eval(vec))
+			}
+		} else {
+			for _, e := range s.aggs {
+				vals = append(vals, e.Eval(vec))
+			}
+		}
+		vector.Apply(false, func(args ...vector.Any) vector.Any {
+			for i, arg := range args {
```
nwt: Nit: We usually calls these vecs.

```suggestion
		vector.Apply(false, func(vecs ...vector.Any) vector.Any {
			for i, vec := range vecs {
```

nwt (approved):

PR #6566, opened 2026-01-27
remove pragma "pg" and disallow column alias in dynamic group-by

mccanne:


philrz:
This is 👍 for me also! I tested by using the 75 queries mentioned in https://github.com/brimdata/super/issues/6410#issuecomment-3717155253 that previously needed the `pragma pg` to be able to run and produce the same result as Postgres. Now they all run ok and match Postgres _without_ the pragma.

mccanne:
After consulting with @philrz , we decided there are too many problems with this in the dynamic domain and `pg` is easier to explain so I'm closing this for now. We can revisit if we can come up with something in between.

PR #6581, opened 2026-01-29
Dynamic runtime selection logic

mattnibs:
This commit changes the logic for selection of either vam or sam runtime. If a given query contains a vectorized file then the vector runtime is selected. This commit also introduces the -vam -sam runtime flags that if enabled will force the selection of either specified runtime.


mccanne (approved):
I wouldn't have had runtimeflags reach into exec but no big deal

PR #6588, opened 2026-01-30
update README for zuma

mccanne:


README.md
philrz: Any thoughts on linking from here to https://superdb.org/super-sql/sql/intro.html#sql-compatibility? No offense taken if you think it's too early, but if you think SQL lovers will want quantitative assurance that their legacy queries are likely to work, maybe worth it?

philrz:
```suggestion
This extended SQL is called [SuperSQL](https://superdb.org/super-sql/intro.html).
```

I know you've got the general link to the docs site below in the "Try It" section, but since we're linking to "shortcuts" up here it feels appropriate to link early to general SuperSQL docs too.

philrz (approved):
Noticed a couple small things, but looks good!

PR #6589, opened 2026-01-30
update install docs with download links from GitHub releases

mccanne:


book/src/getting-started/install.md
```diff
@@ -1,11 +1,16 @@
...
-You can [install a pre-built binary](#homebrew)
-or [build from source code](#building-from-source).
+SuperDB is downloadable software available as a single binary embodied
+in the [super](../command/super.md) command.  This software includes support
+for [SuperSQL](../super-sql/intro.md),
+SuperDB's new query language for super-structured data.
+
+Several options for installation are available:
+* download and install pre-built binaries via links on the
+  [GitHub Releases page](https://github.com/brimdata/super/releases/tag/v0.1.0),
```
philrz:
```suggestion
  [GitHub Releases page](https://github.com/brimdata/super/releases),
```

Might as well go with the shorter link so we don't have to update this each time we tag a release. The latest will always be at the top of the Releases page.

book/src/getting-started/install.md
```diff
@@ -1,11 +1,16 @@
...
+* automatically install a pre-built binary for a Mac or Linux environment
+  with [HomeBrew](#homebrew), or
```
philrz:
```suggestion
  with [Homebrew](#homebrew), or
```

Looks like they don't use the capital B. 🤷‍♂️

philrz (approved):
Spotted a couple small things, but looks good!

PR #6602, opened 2026-02-05
vam: Casting an error should propagate error

mattnibs:


nwt (commented):
Might as well be thorough and add an error value to every file in runtime/ztests/expr/cast.

nwt (approved):

PR #6603, opened 2026-02-05
runtime: Return error when comparing container values

mattnibs:


mattnibs:
Actually we should allow this.

PR #6611, opened 2026-02-06
vam.regexp: Only shorten nulls if nulls is non-zero

mattnibs:
This was causing panics on inputs that have no nulls. Testing on this will occur once we enable vector runtime on mdtest (forthcoming).


mattnibs:
Indeed

PR #6621, opened 2026-02-09
vam: Fix under() skipping const vectors

mattnibs:


runtime/ztests/expr/function/under.yaml
```diff
@@ -29,3 +29,13 @@ output: |
...
+
+---
+
+# Test under for const values in vector runtime.
+spq: under(<foo={x:int64}>)
+
+vector: true
+
+output: |
+  <{x:int64}>
```
nwt: This passes on main so it doesn't exercise the new code. There's really no reliable way to write a ztest for a specific vector form so I think leaving it untested is OK.

PR #6623, opened 2026-02-10
vam.rename: update type of nested records

mattnibs:
This commit fixes an issue in the vector runtime version of the rename operator where nested records were not getting assigned updated types.

### DIFF
--- a/runtime/vam/expr/renamer.go
+++ b/runtime/vam/expr/renamer.go
@@ -1,6 +1,8 @@
 package expr
 
 import (
+	"slices"
+
 	"github.com/brimdata/super"
 	"github.com/brimdata/super/runtime/sam/expr"
 	"github.com/brimdata/super/vector"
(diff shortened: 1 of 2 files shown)


runtime/vam/expr/renamer.go
```diff
@@ -31,5 +31,14 @@ func (r *Renamer) eval(vecs ...vector.Any) vector.Any {
...
 		return vector.NewWrappedError(r.sctx, err.Error(), vec)
 	}
-	return vector.NewRecord(val.Type().(*super.TypeRecord), recVec.Fields, recVec.Len(), recVec.Nulls)
+	return renameRec(recVec, val.Type().(*super.TypeRecord))
+}
+
+func renameRec(vec *vector.Record, typ *super.TypeRecord) *vector.Record {
+	for i, f := range typ.Fields {
+		if rtyp, ok := f.Type.(*super.TypeRecord); ok {
+			vec.Fields[i] = renameRec(vec.Fields[i].(*vector.Record), rtyp)
```
nwt: vec may be shared so modifying vec.Fields is unsafe.

mattnibs: good call 👍🏻

runtime/vam/expr/renamer.go
```diff
@@ -31,5 +33,15 @@ func (r *Renamer) eval(vecs ...vector.Any) vector.Any {
 	if err != nil {
 		return vector.NewWrappedError(r.sctx, err.Error(), vec)
 	}
-	return vector.NewRecord(val.Type().(*super.TypeRecord), recVec.Fields, recVec.Len(), recVec.Nulls)
+	return renameRec(recVec, val.Type().(*super.TypeRecord))
+}
+
+func renameRec(vec *vector.Record, typ *super.TypeRecord) *vector.Record {
```
nwt: Nit: I think something like changeRecordType would be a better name since renameRec reads like it renames the record but that's not what it does.

runtime/ztests/op/rename-nested.yaml
```diff
@@ -11,3 +11,16 @@ output: |
   {net:{src:10.164.94.120,orig_p:39681::(port=uint16),dst:10.47.3.155,resp_p:3389::port}}
   {net:{src:10.164.94.121,orig_p:39681::(port=uint16),dst:10.47.3.155,resp_p:3390::port}}
   {net:{orig_p:39681::(port=uint16),resp_p:3389::port}}
+
+---
+
+# Test nested renamed records in vector runtime are deeply renamed.
+spq: rename x.x:=x.y | values {...x}
```
nwt: Nit:

```suggestion
spq: rename x.x:=x.y | values x
```

PR #6626, opened 2026-02-10
vam math aggs: Persist mixed types errors for partials

mattnibs:


runtime/vam/expr/agg/math.go
```diff
@@ -106,6 +109,15 @@ func (m *mathReducer) consumeNumeric(vec vector.Any) {
 }
 
 func (m *mathReducer) ConsumeAsPartial(vec vector.Any) {
+	if vec.Len() != 1 {
+		panic("invalid length for partial")
+	}
+	if errvec, ok := vec.(*vector.Error); ok && errvec.Vals.Kind() == vector.KindString {
+		if vector.StringValue(errvec.Vals, 0) == mathMixTypesErr.Error() {
+			m.mixedTypesErr = true
+		}
```
nwt: Nit: Unless I'm mistaken, the only way an error can get here is via mathReducer.Result so just checking for an error should be sufficient.

```suggestion
	if vec.Kind() == vector.KindError {
		m.mixedTypesErr = true
```

mattnibs: Right you are

PR #6635, opened 2026-02-17
update fuser terminology in code

mccanne:
This commit updates the old "schema" naming pattern to unify it with the concept of type fusion.

We also moved the caching of seen types into the type fuse agg function.

### DIFF
--- a/runtime/vam/expr/agg/fuse.go
+++ b/runtime/vam/expr/agg/fuse.go
@@ -9,20 +9,29 @@ import (
 )
 
 type fuse struct {
-	types    map[super.Type]struct{}
+	// XXX This convulated data structure of seen/types is here to preserve the
+	// order of the types encountered since some tests rely upon this
+	// and fusion gives different field order for records based on the
+	// input order.  This can happen anyway due to spilling so it's not a
+	// complete solution.  We should decide what to do here.  Maybe the
+	// non-deterministic output is ok.
+	seen     map[super.Type]struct{}
+	types    []super.Type
 	partials []super.Value
 }
 
 func newFuse() *fuse {
 	return &fuse{
-		types: make(map[super.Type]struct{}),
+		seen: make(map[super.Type]struct{}),
 	}
 }
 
 func (f *fuse) Consume(vec vector.Any) {
-	if _, ok := f.types[vec.Type()]; !ok {
-		f.types[vec.Type()] = struct{}{}
+	typ := vec.Type()
+	if _, ok := f.seen[typ]; !ok {
+		f.seen[typ] = struct{}{}
 	}
+	f.types = append(f.types, typ)
 }
 
 func (f *fuse) Result(sctx *super.Context) super.Value {
@@ -37,7 +46,7 @@ func (f *fuse) Result(sctx *super.Context) super.Value {
 		}
 		fuser.Fuse(typ)
 	}
-	for typ := range f.types {
+	for _, typ := range f.types {
 		fuser.Fuse(typ)
 	}
 	return sctx.LookupTypeValue(fuser.Type())


runtime/sam/expr/agg/fuser.go
```diff
@@ -6,40 +6,46 @@ import (
...
-	return &Schema{sctx: sctx, missingFieldsNullable: true}
+func NewFuserWithMissingFieldsAsNullable(sctx *super.Context) *Fuser {
+	f := NewFuser(sctx)
+	f.missingFieldsNullable = true
+	return f
 }
 
-// Mixin mixes t into the fused type.
-func (s *Schema) Mixin(t super.Type) {
+func (s *Fuser) Fuse(t super.Type) {
```
nwt:
```suggestion
func (f *Fuser) Fuse(t super.Type) {
```

runtime/sam/expr/agg/fuser.go
```diff
@@ -6,40 +6,46 @@ import (
...
 	} else {
-		s.typ = s.merge(s.typ, t)
+		s.typ = s.fuse(s.typ, t)
 	}
 }
 
-// Type returns the fused type.
-func (s *Schema) Type() super.Type {
+// Type returns the computed supertype.
+func (s *Fuser) Type() super.Type {
```
nwt:
```suggestion
func (f *Fuser) Type() super.Type {
```

runtime/sam/expr/agg/fuser.go
```diff
@@ -6,40 +6,46 @@ import (
...
 	return s.typ
 }
 
-func (s *Schema) merge(a, b super.Type) super.Type {
+func (s *Fuser) fuse(a, b super.Type) super.Type {
```
nwt:
```suggestion
func (f *Fuser) fuse(a, b super.Type) super.Type {
```

runtime/sam/expr/agg/fuser.go
```diff
@@ -6,40 +6,46 @@ import (
...
-	typ super.Type
+	typ   super.Type
+	types map[super.Type]struct{}
 
 	missingFieldsNullable bool
 }
 
-func NewSchema(sctx *super.Context) *Schema {
-	return &Schema{sctx: sctx}
+// XXX this is used by type checker but I think we can use the other one
```
nwt, line 19: Might be true but the nullable stuff is gross and should go away!

runtime/sam/expr/agg/fuser.go
```diff
@@ -135,15 +141,19 @@ func indexOfField(fields []super.Field, name string) (int, bool) {
 	return -1, false
 }
 
-func (s *Schema) mergeAllRecords(types []super.Type) []super.Type {
+// fuseAllRecords preserve the invarient that any union has a single
```
nwt:
```suggestion
// fuseAllRecords preserves the invariant that any union has a single
```

runtime/sam/op/fuse/fuser.go
```diff
@@ -7,106 +7,98 @@ import (
 	"github.com/brimdata/super/runtime/sam/op/spill"
 )
 
-// Fuser buffers records written to it, assembling from them a unified schema of
-// fields and types.  Fuser then transforms those records to the unified schema
-// as they are read back from it.
-type Fuser struct {
+// valueFuser buffers values, computes a supertype over all the values,
+// then upcasts the values to the computed supertype as the values are read.
+type valueFuser struct {
```
nwt: Nit: Rename file too, to valuefuser.go.

mccanne: Yucky file name.  I'll move the code to fuse.go.

runtime/sam/expr/agg/fuser.go
mccanne: It's going to change to optional fields and I think we can get rid of non-optional field fuser then.

PR #6644, opened 2026-02-18
vam.Dequiet: Do not rip unions on record fields

mattnibs:
This commit fixes an issue with vam.Dequiet where record fields would get deunioned causing a mismatch between the record type and field types.


mattnibs:
> Let's add a test for this.

As you wish.

PR #6648, opened 2026-02-19
mdtest: Run tests in both vam and sam runtimes

mattnibs:


book/src/super-sql/operators/values.md
```diff
@@ -124,7 +124,6 @@ _Values is often used to transform records_
 values [a,b],[b,a] | collect(this)
 # input
 {a:1,b:2}
-{a:3,b:4}
```
nwt, line 127: Why remove this?

mattnibs: Because the output is different for vector runtime `[[1,2],[3,4],[2,1],[4,3]]` then sequential `[[1,2],[2,1],[3,4],[4,3]]` by virtue how how vector runtime processes incoming expressions in the values operator. I did not see a simple way to restructure the query and get the same results. I don't think it detracts from the example having only one input record.

mdtest/mdtest.go
```diff
@@ -194,12 +194,18 @@ func parseMarkdown(source []byte) (map[string]string, []*Test, error) {
...
 				switch {
 				case strings.HasPrefix(s, "dir="):
 					commandDir = strings.TrimPrefix(s, "dir=")
 				case s == "fails":
 					commandFails = true
+				case strings.HasPrefix(s, "runtime="):
+					commandRuntime = strings.TrimPrefix(s, "runtime=")
+					if commandRuntime != "sam" && commandRuntime != "vam" {
+						return ast.WalkStop, fcbError(commandFCB, source, fmt.Sprintf("mdtest-command with invalid runtime: %q", commandRuntime))
+					}
```
nwt: Nit: Simplify this.

```suggestion
				case s == "runtime=sam"):
					commandRuntime = "sam"
				case s == "runtime=vam"):
					commandRuntime = "vam"
```

mdtest/test.go
```diff
@@ -20,17 +20,36 @@ type Test struct {
...
+	Runtime   string // "sam", "vam", or "" for both
 
 	// For SPQ tests
 	Input string
 	SPQ   string
 }
 
-// Run runs the test, returning nil on success.
+// Run runs the test in both the sequential and vector runtimes, returning nil
+// on success.
```
nwt: Nit: This isn't really true since Test.Runtime select the runtime so leave this as is.

```suggestion
// Run runs the test, returning nil on success.
```

mdtest/mdtest_test.go
```diff
@@ -196,6 +236,36 @@ expected
...
+			markdown: `
+~~~mdtest-spq runtime=sam
+#
+spq
+#
+input
+#
+expected
+~~~
+`,
```
nwt, line 250: Add a runtiem=vam case too.

mdtest/mdtest.go
```diff
@@ -227,9 +233,18 @@ func parseMarkdown(source []byte) (map[string]string, []*Test, error) {
...
+				switch word {
+				case "fails":
 					fails = true
+				case "runtime=sam":
+					runtime = "sam"
+				case "runtime=vam":
+					runtime = "vam"
+				default:
+					msg := fmt.Sprintf("unknown word in mdtest-spq info string: %q", word)
+					return ast.WalkStop, fcbError(fcb, source, msg)
```
nwt: This isn't compatible with the `{...}` layout hints we use and we've been fine so far without it so maybe just leave it out?

mattnibs: I don't like having the value not checked because if not you could end up running no tests. I'll add back what I had here.

PR #6651, opened 2026-02-20
fuse complex types aggressively

### LINKED ISSUE #6650: No longer able to output fused GHA data as Parquet
We've often shown examples of querying GitHub Archive data with `super`. Running `fuse` on the original JSON, outputting it as Parquet, and working with it as Parquet has been one of the approaches shown to compare/contrast with table-based legacy SQL approaches. Correlated with the changes in #6633, attempts to output the `fuse`'d JSON as Parquet have started failing.

## Details

Repro is with super commit cbb4109 (which is associated with the merge of the changes in #6633) and subsequent commit 4dbc2fe, which return different errors.

To download the JSON GitHub Archive data:

```
$ for num in $(seq 0 23); do file="2023-02-08-${num}.json.gz"; curl -L -O "https://data.gharchive.org/$file"; done
```

The command that worked in the past (such as at commit 1bced44 right before the merge of #6633) that now fails at commit cbb4109:

```
$ super -version && super -f parquet -o gha-super.parquet -c 'fuse' 2023-02-08-*.json.gz
Version: v0.1.0-23-gcbb41094c
parquetio: not a record: error("missing field \"commits\" is not nullable")
```

Then at the very _next_ commit 4dbc2fe, there's a different error.

```
$ super -version && super -f parquet -o gha-super.parquet -c 'fuse' 2023-02-08-*.json.gz
Version: v0.1.0-24-g4dbc2fe29
parquetio: unsupported type: not implemented: support for DENSE_UNION
```

As a user, it's not clear to me if the first error is still present but "hiding" behind the second.

In any case, the symptom remains through to current tip of main, which is currently commit 72522c2.

```
$ super -version && super -f parquet -o gha-super.parquet -c 'fuse' 2023-02-08-*.json.gz
Version: v0.1.0-37-g72522c227
parquetio: unsupported type: not implemented: support for DENSE_UNION
```

nwt:
The type fusion algorithm can create unions containing multiple complex types of the same kind (e.g, multiple array types) and multiple types with the same type ID (because of named types).  Change the algorithm so that unions in fused types never contain unions, have at most one instance of each kind of non-union complex type, and have at most one type with a given type ID.

Fixes #6650

### DIFF
--- a/runtime/sam/expr/agg/fuser.go
+++ b/runtime/sam/expr/agg/fuser.go
@@ -45,7 +45,7 @@ func (f *Fuser) Type() super.Type {
 	return f.typ
 }
 
-func (f *Fuser) fuse(a, b super.Type) (x super.Type) {
+func (f *Fuser) fuse(a, b super.Type) super.Type {
 	if a == b {
 		return a
 	}


mccanne (approved):

runtime/sam/expr/agg/fuser.go
```diff
@@ -45,14 +45,17 @@ func (f *Fuser) Type() super.Type {
 	return f.typ
 }
 
-func (f *Fuser) fuse(a, b super.Type) super.Type {
+func (f *Fuser) fuse(a, b super.Type) (x super.Type) {
```
mattnibs: Is this variable return type being used (didn't look like it to me).

nwt: Thanks for catching that. Just debug garbage I forgot to remove.

nwt: Removed in 75eb9dfda83b1e50703259e6fd26af2332aeae51.

mattnibs (approved):

PR #6655, opened 2026-02-23
sam compareValues: Deunion values before compare

### LINKED ISSUE #5211: Most functions don't support unions

### LINKED ISSUE #6649: Duplicate buckets in aggregation query against persistent DB storage
After loading a large data set into a persistent DB, the aggregation query shown below produces unexpected duplicate buckets.

```
$ super -version &&
super db -db db-cbb4109 init &&
super db -db db-cbb4109 create wrccdc &&
super db -db db-cbb4109 load -use wrccdc conn.log.gz &&
super db -db db-cbb4109 -s -c "from wrccdc | count() by ts:=bucket(ts, 1h)" > count-cbb4109.sup &&
super -c 'count() by ts | sort -r count,ts' count-cbb4109.sup

Version: v0.1.0-23-gcbb41094c
database created: file:///Users/phil/work/newbug/db-cbb4109
pool created: wrccdc 39uKB7tcJ68XI0cIx6UfSbe0wx8
(1/1) 15.58GB/15.58GB 0B/s 100.00%
39uNy1mPfWjD3QHG86v5XD37nsM committed

{ts:2018-03-24T06:00:00Z,count:2}
{ts:2018-03-24T05:00:00Z,count:2}
{ts:2018-03-24T03:00:00Z,count:2}
...
```

## Details

Repro is with super commit cbb4109, which is associated with the merge of #6633 (remove typed nulls). I've also reproduced the problem on tip of main (commit 72522c2 at the moment). The problem was originally surfaced in an Autoperf run that failed due to an unexpected query result.

The test data is unfortunately very large: A 15 GB file of gzip'ed Zeek TSV `conn` logs which can be downloaded from `s3://brim-sampledata/wrccdc/zeek-logs/conn.log.gz`. Alas, I spent a fair amount of time trying to simplify the repro, but I was unsuccessful both when I did just `super -c` against the file and also when I loaded just subsets of the data into the DB. I also happen to be using the original Zeek TSV because the change in treatment of nulls in #6633 makes it unsafe for me to try to use the same BSUP file for repros before/after the merge of that change.

Here's applying the same recipe at commit 1bced44 right before the merge of #6633, where the problem does not repro.

```
$ super -version &&
super db -db db-1bced44 init &&
super db -db db-1bced44 create wrccdc &&
super db -db db-1bced44 load -use wrccdc conn.log.gz &&
super db -db db-1bced44 -s -c "from wrccdc | count() by ts:=bucket(ts, 1h)" > count-1bced44.sup &&
super -c 'count() by ts | sort -r count,ts' count-1bced44.sup

Version: v0.1.0-22-g1bced4402
database created: file:///Users/phil/work/newbug/db-1bced44
pool created: wrccdc 39uDCC1xpqKMPT9GKCaKZOxCqjn
(1/1) 15.58GB/15.58GB 0B/s 100.00%
39uEutSf65TCHugZSDKOtf5ygMH committed

{ts:2018-03-25T01:00:00Z,count:1}
{ts:2018-03-25T00:00:00Z,count:1}
{ts:2018-03-24T23:00:00Z,count:1}
...
```

mattnibs:
Fixes #6649
Partially fixes #5211

### DIFF
--- a/runtime/sam/expr/sort.go
+++ b/runtime/sam/expr/sort.go
@@ -144,6 +144,8 @@ func (c *Comparator) Compare(a, b super.Value) int {
 }
 
 func compareValues(a, b super.Value, nullsMax bool) int {
+	a = a.Under()
+	b = b.Under()
 	// Handle nulls according to nullsMax
 	nullA := a.IsNull()
 	nullB := b.IsNull()
--- a/runtime/ztests/op/sort/union.yaml
+++ b/runtime/ztests/op/sort/union.yaml
@@ -1,0 +1,25 @@
+# Ensure values are deunioned before sort so they do not
+# fallback to bytes.Compare.
+spq: sort this
+
+vector: true
+
+input: |
+  null
+  null::(int64|null)
+  1
+  2::(int64|null)
+  2
+  1::(int64|null)
+  null::(int64|null)
+  null
+
+output: |
+  1
+  1::(int64|null)
+  2::(int64|null)
+  2
+  null
+  null::(int64|null)
+  null::(int64|null)
+  null


nwt (commented):
Test, please.

philrz:
FYI, repeating the repro steps from #6649 on this branch at commit 8da7316, it does appear the fix is effective. 👍

mattnibs:
> Test, please.

~~@nwt this is a fix for #6649 which is intermittently caused by union values causing values to be compared by bytes.Compare. I don't have a reliable way of testing this but am open to suggestions.~~

EDIT

I came up with a way

runtime/ztests/op/sort/deunion.yaml
```diff
@@ -0,0 +1,15 @@
+# Ensure values are deunioned before sort so they do not
+# fallback to bytes.Compare.
+spq: sort this
+
+input: |
+  "a"
+  "z"
+  "aa"
+  "zz"
```
nwt: This doesn't contain any unions. And I'd just call the file union.yaml.

```suggestion
  null
  null::(int64|null)
  1
  2::(int64|null)
  2
  1::(int64|null)
  null::(int64|null)
  null
```

mattnibs: Forgot to add them but we can do your test

nwt (approved):

PR #6659, opened 2026-02-24
fix IS NULL on union containing null

nwt:
I did this wrong in #6633.


mccanne:
Maybe add a test

nwt:
The `null::(int64|null)` in the input in ‎runtime/ztests/expr/is-null.yaml covers this. I just didn't catch that the corresponding output was bad when I did #6633.

PR #6663, opened 2026-02-25
Add support for db vacate command

mattnibs:


db/commits/store.go
```diff
@@ -98,6 +100,35 @@ func (s *Store) Remove(ctx context.Context, o *Object) error {
 	return s.engine.Delete(ctx, s.pathOf(o.Commit))
 }
 
+// DANGER ZONE - commits should only be removed once a new base has been
+// established.
+func (s *Store) DropCommits(ctx context.Context, commits []ksuid.KSUID) error {
```
nwt: Nit: DeleteCommits feels like a better name.

db/pool.go
```diff
@@ -211,6 +213,48 @@ func (p *Pool) ObjectExists(ctx context.Context, id ksuid.KSUID) (bool, error) {
...
 
+func (p *Pool) Vacate(ctx context.Context, commit ksuid.KSUID, dryrun bool) ([]ksuid.KSUID, error) {
+	branches, err := p.branches.All(ctx)
+	if err != nil {
+		return nil, err
+	}
+	var conflicts []string
+	for _, b := range branches {
+		// Find any branches that diverge from a commit in the deletion path
+		// and if any are found fail.
```
nwt: I'm not sure diverge is the right word here.

```suggestion
		// Find any branches whose history does not include commit
		// and if any are found fail.
```

mattnibs:
After talking this through with @mccanne, I'm going to close this and try a different approach that will also cleanup the branch journal.

PR #6667, opened 2026-02-26
add -samplesize flag to super command

nwt:
The -samplesize flag controls the number of values that the semantic analyzer reads from an input file to compute a fused type.  A value less than 1 reads the entire file.  The default value is 1000.


mccanne (approved):
I think 0 should signal max and default should be much smaller like 1000

PR #6669, opened 2026-02-27
add support for optional fields in records

mccanne:
This commit adds support for optional fields in records making fusing of varied JSON data that arises from common sum types much more uniform.  The SUP format is updated with a ? following field names to represent optional fields and _ to represent a field value that is not present.

While the SuperSQL parser can parse SUP literals with optional fields, record expressions do not yet support optional fields as this support will come in a subequent PR.

The vector representation encodes optional field columns by run-length encoding their presence and generates a slot map on demand when serializing to BSUP or when dereferencing a field to in turn generate a vector.Dynamic mixed with Missing error values.

This commit is not backward compatible with previous BSUP versions and we will update the BSUP version nunber in a subsequent PR prior to the RINCON release.

### DIFF
--- a/recordbuilder.go
+++ b/recordbuilder.go
@@ -116,13 +116,11 @@ func NewRecordBuilder(sctx *Context, fields field.List) (*RecordBuilder, error) 
 		fieldInfos[len(fieldInfos)-1].containerEnds = len(currentRecord)
 	}
 
-	r := &RecordBuilder{
+	return &RecordBuilder{
 		fields:  fieldInfos,
 		builder: scode.NewBuilder(),
 		sctx:    sctx,
-	}
-	r.Reset()
-	return r, nil
+	}, nil
 }
 
 // check if fieldname is "in" one of the fields in fis, or if
--- a/sio/arrowio/writer.go
+++ b/sio/arrowio/writer.go
@@ -210,23 +210,14 @@ func (w *Writer) newArrowDataType(typ super.Type) (arrow.DataType, error) {
 		}
 		switch name {
 		case "arrow_day_time_interval":
-			if typ.Opts != 0 {
-				return nil, fmt.Errorf("%w: arrow_day_time_interval cannot have optional fields", ErrUnsupportedType)
-			}
 			if slices.Equal(typ.Fields, dayTimeIntervalFields) {
 				return arrow.FixedWidthTypes.DayTimeInterval, nil
 			}
 		case "arrow_decimal128":
-			if typ.Opts != 0 {
-				return nil, fmt.Errorf("%w: arrow_decimal128 cannot have optional fields", ErrUnsupportedType)
-			}
 			if slices.Equal(typ.Fields, decimal128Fields) {
 				return &arrow.Decimal128Type{}, nil
 			}
 		case "arrow_month_day_nano_interval":
-			if typ.Opts != 0 {
-				return nil, fmt.Errorf("%w: arrow_month_day_nano_interval cannot have optional fields", ErrUnsupportedType)
-			}
 			if slices.Equal(typ.Fields, monthDayNanoIntervalFields) {
 				return arrow.FixedWidthTypes.MonthDayNanoInterval, nil
 			}
--- a/vector/builder.go
+++ b/vector/builder.go
@@ -137,9 +137,9 @@ func (r *recordBuilder) Write(bytes scode.Bytes) {
 }
 
 func (r *recordBuilder) Build() Any {
-	var fields []Field
+	var fields []*Field
 	for k := range r.fields {
-		fields = r.fields[k].build(fields, r.len)
+		fields = append(fields, r.fields[k].build(r.len))
 	}
 	return NewRecordFromFields(r.typ, fields, r.len)
 }
@@ -157,12 +157,12 @@ func (f *fieldBuilder) write(bytes scode.Bytes, off uint32) {
 	f.val.Write(bytes)
 }
 
-func (f *fieldBuilder) build(fields []Field, n uint32) []Field {
+func (f *fieldBuilder) build(n uint32) *Field {
 	var runs []uint32
 	if f.opt {
 		runs = f.runs.End(n)
 	}
-	return append(fields, Field{Val: f.val.Build(), Runs: runs, Len: n})
+	return &Field{Val: f.val.Build(), Runs: runs, Len: n}
 }
 
 type errorBuilder struct {
--- a/vector/record.go
+++ b/vector/record.go
@@ -10,25 +10,21 @@ import (
 
 type Record struct {
 	Typ    *super.TypeRecord
-	fields []Field
+	fields []*Field
 	len    uint32
-}
-
-func (r *Record) RawFields() []Field {
-	return r.fields
 }
 
 var _ Any = (*Record)(nil)
 
 func NewRecord(typ *super.TypeRecord, fields []Any, length uint32) *Record {
-	wrapper := make([]Field, 0, len(fields))
+	wrapper := make([]*Field, 0, len(fields))
 	for _, f := range fields {
-		wrapper = append(wrapper, Field{Val: f, Len: length})
+		wrapper = append(wrapper, &Field{Val: f, Len: length})
 	}
 	return &Record{typ, wrapper, length}
 }
 
-func NewRecordFromFields(typ *super.TypeRecord, fields []Field, length uint32) *Record {
+func NewRecordFromFields(typ *super.TypeRecord, fields []*Field, length uint32) *Record {
 	return &Record{typ, fields, length}
 }
 
@@ -57,10 +53,10 @@ func (r *Record) Fields(sctx *super.Context) []Any {
 }
 
 func (r *Record) Field(i int) *Field {
-	return &r.fields[i]
+	return r.fields[i]
 }
 
-func (r *Record) Slice(from, to int) []Field {
+func (r *Record) Slice(from, to int) []*Field {
 	return r.fields[from:to]
 }
 
@@ -115,16 +111,6 @@ type Field struct {
 	mu    sync.Mutex
 	slots []int32 // map record slot to field slot (-1 for none) (from Nones or builder)
 	dyn   Any
-}
-
-func (f *Field) Copy() Field {
-	return Field{
-		Val:   f.Val,
-		Len:   f.Len,
-		Runs:  f.Runs,
-		slots: f.slots,
-		dyn:   f.dyn,
-	}
 }
 
 func (f *Field) Deref(sctx *super.Context) Any {
--- a/runtime/vam/expr/dropper.go
+++ b/runtime/vam/expr/dropper.go
@@ -48,7 +48,7 @@ func (d *Dropper) eval(vecs ...vector.Any) vector.Any {
 func (d *Dropper) drop(val vector.Any, fm fieldsMap) (vector.Any, bool) {
 	switch val := vector.Under(val).(type) {
 	case *vector.Record:
-		var valFields []vector.Field
+		var valFields []*vector.Field
 		var typFields []super.Field
 		var changed bool
 		for i, f := range super.TypeRecordOf(val.Type()).Fields {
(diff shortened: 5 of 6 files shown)


book/src/formats/bsup.md
```diff
@@ -390,6 +391,17 @@ For sets, the concatenation of elements must be normalized so that the
 sequence of bytes encoding each element's tag-counted value is
 lexicographically greater than that of the preceding element.
 
+For records, when its type has optional fields,
+a bit vector of length NF is encoded as a tag-encoded value of
+floor((NF+7)/8) bytes to indicate the omission of an optional value
+where NF is the number of optional fields in the record.
+The field order of optional fields determines their position
+in the bit vector with bit numbers 0-7 (least significat to most significant)
```
nwt:
```suggestion
in the bit vector with bit numbers 0-7 (least significant to most significant)
```

book/src/tutorials/prs.bsup
nwt, line 1: Do we want this file?

compiler/semantic/sem/expr.go
```diff
@@ -490,12 +498,18 @@ func valueToExpr(loc ast.Node, typ super.Type, bytes scode.Bytes) Expr {
...
-	it := bytes.Iter()
+	it := scode.NewRecordIter(bytes, typ.Opts)
 	for _, f := range typ.Fields {
 		if it.Done() {
 			panic(it)
 		}
-		elem := &FieldElem{Node: loc, Name: f.Name, Value: valueToExpr(loc, f.Type, it.Next())}
+		val, none := it.Next(f.Opt)
+		var elem RecordElem
+		if none {
```
nwt:
```suggestion
		var elem RecordElem
		if al, none := it.Next(f.Opt); none {
```

scode/iter.go
```diff
@@ -50,3 +50,40 @@ func (i *Iter) NextTagAndBody() Bytes {
...
+		off := r.off
+		r.off++
+		if Testbit(r.nones, off) {
+			return nil, true
+		}
+	}
+	return r.elems.Next(), false
+}
+
+func Testbit(b []byte, off int) bool {
```
nwt:
```suggestion
func TestBit(b []byte, off int) bool {
```

sio/arrowio/writer.go
```diff
@@ -97,9 +97,11 @@ func (w *Writer) Write(val super.Value) error {
 	} else if w.typ != recType {
 		return fmt.Errorf("%w: %s and %s", ErrMultipleTypes, sup.FormatType(w.typ), sup.FormatType(recType))
 	}
-	it := val.Iter()
+	it := scode.NewRecordIter(val.Bytes(), recType.Opts)
 	for i, builder := range w.builder.Fields() {
-		w.buildArrowValue(builder, recType.Fields[i].Type, it.Next())
+		// We don't check for none since b will be nil and none is the
```
nwt:
```suggestion
```

complex.go
```diff
@@ -187,16 +187,18 @@ func TypeUnder(typ Type) Type {
 type Field struct {
 	Name string
 	Type Type
+	Opt  bool
 }
 
-func NewField(name string, typ Type) Field {
-	return Field{name, typ}
+func NewField(name string, typ Type, opt bool) Field {
```
nwt: Since this is usually called with opt==false, could we keep its signature unchanged and add a new three-parameter constructor like NewFieldWithOpt? That'll make most calls more concise and also make it easier to find call sites where we're actually creating optional fields.

book/src/tutorials/prs.bsup
mccanne: Yes, the tentative plan is to bring back the jq tutorial for MALIBU.

csup/field.go
```diff
@@ -4,30 +4,54 @@ import (
...
 )
 
 type FieldEncoder struct {
-	name   string
-	values Encoder
+	name    string
+	values  Encoder
+	opt     bool
+	rle     vector.RLE
+	encoder *Uint32Encoder
```
nwt: Nit: Maybe call this optEncoder, presenceEncoder, or something else suggestive?

csup/record.go
```diff
@@ -11,32 +11,42 @@ import (
...
+		if f.Opt {
+			nopt++
+		}
 	}
-	return &RecordEncoder{fields: fields}
+	return &RecordEncoder{fields: fields, nopt: nopt}
 }
 
 func (r *RecordEncoder) Write(body scode.Bytes) {
+	index := r.count
```
nwt: Nit: We usually call this slot.

```suggestion
	slot := r.count
```

csup/field.go
```diff
@@ -4,30 +4,54 @@ import (
...
 }
 
-func (f *FieldEncoder) write(body scode.Bytes) {
+func (f *FieldEncoder) write(body scode.Bytes, index uint32) {
```
nwt: Nit: We usually call this slot.

```suggestion
func (f *FieldEncoder) write(body scode.Bytes, slot uint32) {
```

runtime/sam/expr/function/flatten.go
```diff
@@ -30,25 +30,30 @@ func (n *Flatten) Call(args []super.Value) super.Value {
...
 	n.Reset()
-	n.encode(typ.Fields, inner, field.Path{}, val.Bytes())
+	n.encode(typ, inner, field.Path{}, val.Bytes())
 	return super.NewValue(n.sctx.LookupTypeArray(inner), n.Bytes())
 }
 
-func (n *Flatten) innerTypeOf(fields []super.Field) super.Type {
-	n.types = n.appendTypes(n.types[:0], fields)
+func (n *Flatten) innerTypeOf(b scode.Bytes, typ *super.TypeRecord) super.Type {
+	n.types = n.appendTypes(n.types[:0], b, typ)
```
nwt: Nit: Type precedes bytes in almost all other signatures.

```suggestion
func (n *Flatten) innerTypeOf(typ *super.TypeRecord, b scode.Bytes) super.Type {
	n.types = n.appendTypes(n.types[:0], typ, b)
```

runtime/sam/expr/function/flatten.go
```diff
@@ -79,6 +87,7 @@ func (n *Flatten) encode(fields []super.Field, inner super.Type, base field.Path
 			super.BeginUnion(&n.Builder, union.TagOf(typ))
 		}
 		n.BeginContainer()
+		//n.Append(nil) // Fields aren't optional
```
nwt: Delete this?

```suggestion
```

context.go
```diff
@@ -495,6 +506,9 @@ func (c *Context) DecodeTypeValue(tv scode.Bytes) (Type, scode.Bytes) {
 		for range n {
 			var typ Type
 			typ, tv = c.DecodeTypeValue(tv)
+			if typ == nil {
+				panic(tv)
+			}
```
nwt:
```suggestion
```

book/src/tutorials/prs.bsup
nwt: But when we do we'll generate prs.bsup from prs.json in an mdtest-command, right?

runtime/sam/expr/dequiet.go
```diff
@@ -26,29 +26,47 @@ func (d *Dequiet) Eval(this super.Value) super.Value {
...
+		panic(typ)
+	}
 	var changed bool
 	builder.BeginContainer()
 	var fields []super.Field
-	it := b.Iter()
+	it := scode.NewRecordIter(b, rtyp.Opts)
+	// For building the output record, we don't how many optional fields there
+	// will be until after we make the type.  So we call EndContainerWithBits
+	// to get deal with.
```
nwt:
```suggestion
	// to deal with.
```

runtime/sam/expr/dropper.go
```diff
@@ -1,114 +1,127 @@
...
+		dropMap.Add(f)
+	}
 	return &Dropper{
-		sctx:     sctx,
-		fields:   fields,
-		droppers: make(map[int]*dropper),
+		sctx:      sctx,
+		dropMap:   dropMap,
+		dropType:  make(map[super.Type]super.Type),
+		emptyType: sctx.MustLookupTypeRecord([]super.Field{}),
```
nwt: Nit:

```suggestion
		emptyType: sctx.MustLookupTypeRecord(nil),
```

runtime/sam/expr/eval.go
```diff
@@ -638,6 +638,26 @@ func getNthFromContainer(container scode.Bytes, idx int) (scode.Bytes, int) {
 	return nil, -1
 }
 
+func getNthFromRecord(container scode.Bytes, idx int, typ *super.TypeRecord) (scode.Bytes, int) {
```
nwt: Nit: Type usually precedes bytes.

```suggestion
func getNthFromRecord(typ *super.TypeRecord, container scode.Bytes, idx int) (scode.Bytes, int) {
```

sio/arrowio/writer.go
```diff
@@ -209,21 +210,30 @@ func (w *Writer) newArrowDataType(typ super.Type) (arrow.DataType, error) {
...
+			if typ.Opts != 0 {
+				return nil, fmt.Errorf("%w: arrow_decimal128 cannot have optional fields", ErrUnsupportedType)
+			}
 			if slices.Equal(typ.Fields, decimal128Fields) {
 				return &arrow.Decimal128Type{}, nil
 			}
 		case "arrow_month_day_nano_interval":
+			if typ.Opts != 0 {
+				return nil, fmt.Errorf("%w: arrow_month_day_nano_interval cannot have optional fields", ErrUnsupportedType)
+			}
```
nwt: Nit: I don't think we want to add these. The ethos in here is to return ErrUnsupportedType only when we can't produce a reasonable encoding of the value. If typ.Opts!=0 for any of these, slices.Equal will return false and we'll just encode the value as a record.

vector/builder.go
```diff
@@ -111,32 +111,58 @@ func (n *namedBuilder) Build() Any {
...
+	f.val.Write(bytes)
+}
+
+func (f *fieldBuilder) build(fields []Field, n uint32) []Field {
+	var runs []uint32
+	if f.opt {
+		runs = f.runs.End(n)
 	}
-	return NewRecord(r.typ, vecs, r.len)
+	return append(fields, Field{Val: f.val.Build(), Runs: runs, Len: n})
```
nwt: Nit: Just return a Field and let the caller do the append.

vector/record.go
```diff
@@ -29,10 +44,218 @@ func (r *Record) Len() uint32 {
...
+	dyn   Any
+}
+
+func (f *Field) Copy() Field {
+	return Field{
+		Val:   f.Val,
+		Len:   f.Len,
+		Runs:  f.Runs,
+		slots: f.slots,
+		dyn:   f.dyn,
```
nwt: You need to hold f.mu when reading these. But why not avoid the issue entirely by changing Record.fields to []*Field so you don't need this Copy method in runtime/vam/expr/dropper.go?

book/src/tutorials/prs.bsup
mccanne: The only reason I updated it is because a test failed.  I suppose we can delete and update the markdown to skip any failing tests?  When we re-expose the tutorial, I hope csup will be our default format anyway!

nwt: I didn't realize that tests depend on this. We can fix that later.

recordbuilder.go
```diff
@@ -115,11 +116,13 @@ func NewRecordBuilder(sctx *Context, fields field.List) (*RecordBuilder, error)
...
 
-	return &RecordBuilder{
+	r := &RecordBuilder{
 		fields:  fieldInfos,
 		builder: scode.NewBuilder(),
 		sctx:    sctx,
-	}, nil
+	}
+	r.Reset()
+	return r, nil
```
nwt: Nit: Don't need this change.

vector/record.go
mccanne: Feels gross to have lots of pointers and small allocs when you don't need them but I'll change it

PR #6674, opened 2026-02-27
bump BSUP version number to 1

mccanne:
This commit changes the BSUP version number from 0 to 1 and expands the version number bits from 0 to 7.  There is currently no backward compatibility with version 0 and instead an error will be reported when trying to read old BSUP files.

### DIFF
--- a/sio/bsupio/parser.go
+++ b/sio/bsupio/parser.go
@@ -21,16 +21,15 @@ type parser struct {
 	maxSize int
 }
 
-const BSUP_VERSION = 1
-const BSUP_MAGIC = "BSUP"
+const BSUPVersion = 1
 
 func CheckVersion(code byte) error {
 	var version int
 	if (code & 0x80) != 0 {
 		version = int(code & 0x7f)
 	}
-	if version != BSUP_VERSION {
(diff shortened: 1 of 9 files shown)


sio/bsupio/parser.go
```diff
@@ -21,13 +21,27 @@ type parser struct {
 	maxSize int
 }
 
+const BSUP_VERSION = 1
+const BSUP_MAGIC = "BSUP"
```
nwt: Nit:

```suggestion
const BSUPVersion = 1
const BSUPMagic = "BSUP"
```

sio/bsupio/parser.go
```diff
@@ -21,13 +21,27 @@ type parser struct {
...
+
+func CheckVersion(code byte) error {
+	var version int
+	if (code & 0x80) != 0 {
+		version = int(code & 0x7f)
+	}
+	if version != BSUP_VERSION {
+		return fmt.Errorf("BSUP version mismatch: expected %d found %d", BSUP_VERSION, version)
```
nwt: Nit:

```suggestion
		return fmt.Errorf("BSUP version mismatch: expected %d, found %d", BSUP_VERSION, version)
```

PR #6683, opened 2026-02-28
fix debug deadlock

mccanne:


mccanne:
I'll see if I can finish this

PR #6700, opened 2026-03-06
add support for Optional/None values to vector/vam

mccanne:
This commit changes the fields of vector.Record back to vector.Any and introduces the concept of vector.Optional and vector.None. An Optional is simply a two-typed Dynamic of the original values plus the nones where the nones are represnted by vector.None which wraps a vector.Error("missing").  This means that None expresses itself as Missing when referenced unless asserted to None and treated otherwise.

There is also a performance issue in that we always load tags of Optionals when reading optional fields in vcache and we always compute the tags from the runlens in recordbuidler.  In a future PR, we will compute the tags on demand under a lock.

There are still problems in both sam/vam with put/cut and problems in vam with None turning into Missing in aggregate functions. These issues will be addressed in subsequent PRs.

### DIFF
--- a/vector/record.go
+++ b/vector/record.go
@@ -175,7 +175,7 @@ func NewFieldFromRLE(sctx *super.Context, vec Any, length uint32, nones []uint32
 	return &Optional{NewDynamic(tags, []Any{vec, &None{NewMissing(sctx, noneLen)}})}
 }
 
-// An Optional value is a special Dynamic that has two tags comprosing the
+// An Optional value is a special Dynamic that has two tags comprising the
 // values present and the Nones.
 type Optional struct {
 	*Dynamic


vector/builder.go
```diff
@@ -252,10 +252,10 @@ func (u *unionBuilder) Write(bytes scode.Bytes) {
 	u.tags = append(u.tags, uint32(tag))
 }
 
-func (u *unionBuilder) Build() Any {
+func (u *unionBuilder) Build(sctx *super.Context) Any {
```
mattnibs, line 255: nit: If the argument isn't being used we typically do this but whateves:
```suggestion
func (u *unionBuilder) Build(_ *super.Context) Any {
```

vector/record.go
```diff
@@ -244,3 +141,56 @@ func (r *RLE) End(off uint32) []uint32 {
...
+	}
+	tags, noneLen := buildTags(nones, length)
+	if noneLen == 0 {
+		// This field is optional but everything is here in this instance.
+		return vec
+	}
+	return &Optional{NewDynamic(tags, []Any{vec, &None{NewMissing(sctx, noneLen)}})}
+}
+
+// An Optional value is a special Dynamic that has two tags comprosing the
```
mattnibs:
```suggestion
// An Optional value is a special Dynamic that has two tags comprising the
```

vector/builder.go
mccanne: It's used here in the recursive call...

